PR #2846, opened 2022-06-22
Add jupyterlab to Scooby Report

banesullivan:
This would be useful to include in the Report when helping people debug in-notebook rendering.

@pyvista/developers, any other packages we may want to add while this is open?

### DIFF
--- a/pyvista/utilities/errors.py
+++ b/pyvista/utilities/errors.py
@@ -304,7 +304,6 @@ class Report(scooby.Report):
             'meshio',
             'jupyterlab',
             'pythreejs',
-            'ipyvtklink',
         ]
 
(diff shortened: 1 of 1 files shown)


pyvista/utilities/errors.py
```diff
@@ -302,6 +302,7 @@ def __init__(self, additional=None, ncol=3, text_width=80, sort=False, gpu=True)
             'itkwidgets',
             'tqdm',
             'meshio',
+            'jupyterlab',
```
akaszynski, line 305:
```suggestion
            'jupyterlab',
            'pythreejs',
            'ipyvtklink',
```

akaszynski:
> any other packages we may want to add while this is open?

Two others.

PR #2847, opened 2022-06-22
Add skip_zero_time property in OpenFOAMReader

### LINKED ISSUE #2843: Expose SkipZeroTime property of vtkOpenFOAMReader

I think it would be nice to expose the property named "SkipZeroTime" in the OpenFOAMReader.

vtkOpenFOAMReader : https://vtk.org/doc/nightly/html/classvtkOpenFOAMReader.html#ada0ff6f0dcc570b5f4145dc71944fbde

### Pseudocode or Screenshots

It seems that this property could be implemented the same way as the `case_type` property. It looks like an import settings or something like that.

skip_zero_time_in_paraview

### Code of Conduct

Failxxx:
### Overview

Expose the `SkipZeroTime` property for OpenFOAMReader.

 

Resolves #2843


pyvista/utilities/reader.py
```diff
@@ -808,6 +808,35 @@ def decompose_polyhedra(self, value):
...
+        """
+        return bool(self.reader.GetSkipZeroTime())
+
+    @skip_zero_time.setter
+    def skip_zero_time(self, value):
+        if value:
+            self.reader.SkipZeroTimeOn()
+        else:
+            self.reader.SkipZeroTimeOff()
+
```
MatthewFlamm, line 839: In my testing we need to call `self._update_information` after this.

tests/utilities/test_reader.py
```diff
@@ -532,6 +532,14 @@ def test_openfoam_decompose_polyhedra():
     assert reader.decompose_polyhedra is True
 
 
+def test_openfoam_skip_zero_time():
+    reader = get_cavity_reader()
+    reader.skip_zero_time = False
+    assert reader.skip_zero_time is False
+    reader.skip_zero_time = True
```
MatthewFlamm, line 539: It would be nice to test that `reader.time_values` contains zero when `reader.skip_zero_time is False`, but does not contain zero when `reader.skip_zero_time is True`

MatthewFlamm (commented):
Looking good so far

MatthewFlamm:
> ### Edit
> But now I see that pipeline does not fail... I may have done something wrong in my tests... I am able to reproduce this on another computer.

I do not fully understand what the differences are between some of these update and refresh methods do.  `SetRefresh` doesn't seem to work for me though.  I tested this on [8d0e3c7](https://github.com/pyvista/pyvista/pull/2847/commits/8d0e3c7cc5b81225802ec6bcf589a47f2e5cd4d2), i.e. before you added the call to `self._update_information` to the setter.

```py
import pyvista
from pyvista import examples

filename = pyvista.download_cavity(load=False)
reader = pyvista.get_reader(filename)
reader.time_values  # Has 0.0 in it

reader.skip_zero_time = True
reader.time_values  # Has 0.0 in it

reader.reader.SetRefresh()
reader.time_values  # Has 0.0 in it

reader._update_information()
reader.time_values  # No longer contains 0.0
```

Failxxx:
Here is another test. I am using the same configuration for the OpenFOAMReader as in [8d0e3c7](https://github.com/pyvista/pyvista/pull/2847/commits/8d0e3c7cc5b81225802ec6bcf589a47f2e5cd4d2).
`SetRefresh()` seems to be doing something that `_update_information` does not do.
Could you please confirm that you are also able to reproduce this?

```python
import pyvista

filename = "/home/hnum/OLART/Docs/Data/OpenFOAM/cavity/foam.foam"
reader = pyvista.get_reader(filename)

reader.skip_zero_time = False
reader._update_information()
print(reader.skip_zero_time is False)   # True
print(0.0 in reader.time_values)        # True

reader.skip_zero_time = True
reader._update_information()
print(reader.skip_zero_time is True)    # True
print(0.0 not in reader.time_values)    # True

reader.skip_zero_time = False
reader._update_information()
print(reader.skip_zero_time is False)   # True
print(0.0 in reader.time_values)        # False <--- Should be True here

reader.skip_zero_time = True
reader._update_information()
print(reader.skip_zero_time is True)    # True
print(0.0 not in reader.time_values)    # True

reader.skip_zero_time = False
reader._update_information()
print(reader.skip_zero_time is False)   # True
print(0.0 in reader.time_values)        # False <--- Should be True here

reader.skip_zero_time = True
reader._update_information()
print(reader.skip_zero_time is True)    # True
print(0.0 not in reader.time_values)    # True

reader.reader.SetRefresh()

reader.skip_zero_time = False
reader._update_information()
print(reader.skip_zero_time is False)   # True
print(0.0 in reader.time_values)        # True <--- Added setRefresh(), it seems to be working

reader.skip_zero_time = True
reader._update_information()
print(reader.skip_zero_time is True)    # True
print(0.0 not in reader.time_values)    # True

reader.skip_zero_time = False
reader._update_information()
print(reader.skip_zero_time is False)   # True
print(0.0 in reader.time_values)        # False <--- Should be True here
```

MatthewFlamm:
> Here is another test. I am using the same configuration for the OpenFOAMReader as in [8d0e3c7](https://github.com/pyvista/pyvista/pull/2847/commits/8d0e3c7cc5b81225802ec6bcf589a47f2e5cd4d2).
> `SetRefresh()` seems to be doing something that `_update_information` does not do.
> Could you please confirm that you are also able to reproduce this?

Good investigating!  This is puzzling that one method is required in one direction yet a different method is required in the opposite direction.  Would you recommend adding `SetRefresh` into `_update_information`, or just calling it in addition only in this setter?

Failxxx:
> Would you recommend adding `SetRefresh` into `_update_information`, or just calling it in addition only in this setter?

Hard to tell haha. Since we do not fully understand what these methods do, I think we should only add `SetRefresh()` in this setter to avoid any problem caused by adding it to `_update_information()`.

I am going to do some more testing soon.

Failxxx:
Ok so I've tested with this version (`SetRefresh()` in the setter) and it seems to be working as intended. Do you think we can implement it this way for the moment?

MatthewFlamm (approved):
LGTM

PR #2853, opened 2022-06-24
Set number of labels to release as minimum

tkoyama010:
### Overview

More than two labels are also possible for pull requests with multiple contents.

 

### Details

- None

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -46,7 +46,7 @@ jobs:
         labels: release
     - uses: mheap/github-action-required-labels@v2
       with:
-        mode: exactly
+        mode: minimum
         count: 1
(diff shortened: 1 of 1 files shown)


akaszynski (approved):
LGTM. Thanks for fixing this.

akaszynski:
I'd like to point out that it's a little annoying to see the red check on each and every commit, despite the fact that some of the commits might actually be valid. It makes walking back through looking for valid commits a bit more tedious. Since I like the approach of waiting, is there a way we could still include the labeler workflow but not have this behavior?

tkoyama010:
I agree. I'm happy with the functionality but need a better way to display it than blocking the merge.

akaszynski:
Abandoning. See #2860

PR #2851, opened 2022-06-24
Add community-review label is need to merge

tkoyama010:
### Overview

The PyVista community has grown and also the developer community has grown. Nowadays, pull requests are reviewed from multiple locations around the world. Every time zone in the world should have an equal opportunity to review PyVista's pull request. In other words, reviewers should be granted their chance to review for at least 1 day. However, maintainers, myself included, tend to be aggressive in merging (because we want to improve PyVista quickly!). It is also painful to open an uncommented PullRequest after a day pass even though there is an enable auto-merge.

So I built a system to solve this problem with GitHub Action. This system is divided into two parts.

1. Community Review: Here the label `community-review` is added one day after the last action in the PullRequest. If someone takes some action, it will wait another day. If someone comments, the `community-review` label is removed and a `do-not-merge` label is added. This label prevents merging, so please remove it when the discussion is over. You can also add it when you feel a long time discussion is needed.

1. Pull Request Labels: Here we check the labels required for the merge, blocking the merge if they do not meet the criteria. The label "community-reviewed", which is given if 1. is passed, is a required label. do-not-merge" as a label that must not exist. I also need a required label for release notes. This solves the problem of https://github.com/pyvista/pyvista/pull/2453 .

Please review as there may be use cases that I am not anticipating. Reviews are always important!!!

 
relate to https://github.com/pyvista/pyvista/discussions/2245

resolves #2774

### Details

- https://github.com/marketplace/actions/close-stale-issues
- https://github.com/marketplace/actions/require-labels

PR #2517, opened 2022-04-26
Use threejs for gallery

akaszynski:
Proposal to use threejs as a gallery for our documentation generation.

This PR proposes using threejs to generate interactive documentation for our gallery. It requires https://github.com/jupyter-widgets/pythreejs/pull/374 and a followup release as pythreejs does not currently support auto-resize and manual resizing of our examples requires apriori knowledge of the container or window size (which we don't have).

PR #2857, opened 2022-06-25
Clean up community-reviewed label wording

adeak:
Small cleanup after https://github.com/pyvista/pyvista/pull/2851.

One point though: I've added "approval" to the language. I can't tell from #2851 whether this is actually the case, but it should be: we need the label to only consider _approving_ reviews. In any case @tkoyama010 please check if you agree with the changes here, and whether they make sense :)

OK, now I see that https://github.com/pyvista/pyvista/pull/2517#issuecomment-1166180176 got the comment even though there are no approving reviews there. In fact there are no reviews at all!

### DIFF
--- a/.github/workflows/community-review.yml
+++ b/.github/workflows/community-review.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/stale@v4
         with:
-          stale-pr-message: "This PR is reviewd by community because it has been open 1 day with no activity."
+          stale-pr-message: "This PR has been approved by reviewers for 1 day with no activity."
           days-before-issue-stale: -1
           days-before-issue-close: -1
           days-before-pr-stale: 1
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -433,12 +433,12 @@ The main features of our branching model are:
 Labels
 ~~~~~~
 
-The appropriate label will be added to pull request by CI. This will
+Appropriate labels will be added to pull requests by CI. This will
 be used for future release notes. CI will also verify that the necessary
-labels have been added. If CI doesn't add a label, please it manually.
+labels have been added. If CI doesn't add a label, please add it manually.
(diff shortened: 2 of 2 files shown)


tkoyama010:
Good point. As you say, we currently have 1 day from the last action of the PullRequest to determine that the community review is completed. As Waiting 1 day after Approved Review would be ideal but is technically difficult. We probably need to be technical enough to create a new GitHubAction App. Personally, I see a tradeoff between the development effort and the benefit gained. I believe that if all reviewer actions are not for 1 day, it is enough for a review opportunity.

adeak:
> I believe that if all reviewer actions are not for 1 day, it is enough for a review opportunity.

Hmm, I guess the question is what the goal is with this machinery. My mental model was that we want to prepare auto-merging for when there's an approving review that's at least 1 day old and no other activity on the PR.

Is this meant to do something different? If yes, then I misunderstood, and we should rephrase my rephrasing in this PR :)

tkoyama010:
Your understanding is correct. I am trying to achieve something similar by combining existing packages without developing my own. Perhaps that is what is failing. It is important to reach a goal, so I will explore if it is possible to achieve it. Until then, I'm afraid I will have to ask you to manually remove the "community-reviewed" label if we think it is not correct.

adeak:
Of course, no worries at all. I just wanted to make sure I understand we want the same thing from this feature eventually :)

akaszynski:
The more I think about this, the more I think that this won't work. The community review label will always be added after the last action, regardless if other workflows passed or if anyone else reviewed it. I think we should revert this and continue following the 24 hour rule.

I think, even for gigantic (and important) repositories like numpy, there isn't an explicit process in place for merging. For example, https://github.com/numpy/numpy/pull/21789 didn't even have approvals.

I think our current review standards are solid, and thanks to all the volunteers here, I think we're doing a great job improving our review process.

PR #2872, opened 2022-06-26
Ensure same name arrays do not shallow copy

### LINKED ISSUE #2864: Array segmentation fault
As pointed out by @adeak in [this comment](https://github.com/pyvista/pyvista/pull/1953#issuecomment-1166383011), the following leads to a segmentation fault: 

```py
>>> import numpy as np
>>> puppy = examples.download_puppy()
>>> puppy.clear_data()
>>> puppy['scalars'] = np.arange(puppy.n_points, dtype=float)
>>> puppy['scalars'] *= 2/3
>>> puppy['scalars']
Process Python segmentation fault (core dumped)
```

akaszynski:
Resolves #2864 by ensuring that same named arrays are not shallow copied, but rather have their reference directly returned by `_prepare_array`.

For whatever reason, arrays that have been shallow copied still have their underlying arrays collected by VTK despite that they're used.

You can reproduce the memory issue (from a fresh python shell) with:
```py
import numpy as np
import pyvista

points = np.zeros((10000000, 3))
dataset = pyvista.PointSet(points)
data = np.arange(dataset.n_points, dtype=float)

dataset['scalars'] = data.copy()
dataset['scalars'] *= 0.5
print(dataset['scalars'])
```

### DIFF
--- a/pyvista/core/datasetattributes.py
+++ b/pyvista/core/datasetattributes.py
@@ -774,10 +774,8 @@ class DataSetAttributes(_vtk.VTKObjectWrapper):
                 if data.flags.c_contiguous:
                     # no reason to return a shallow copy if the array and name
                     # are identical, just return the underlying array name
-                    if not deep_copy:
-                        if isinstance(name, str):
-                            if data.VTKObject.GetName() == name:
-                                return data.VTKObject
+                    if not deep_copy and isinstance(name, str) and data.VTKObject.GetName() == name:
+                        return data.VTKObject
 
                     vtk_arr = copy_vtk_array(data.VTKObject, deep=deep_copy)
                     if isinstance(name, str):
--- a/pyvista/core/datasetattributes.py
+++ b/pyvista/core/datasetattributes.py
@@ -772,6 +772,13 @@ class DataSetAttributes(_vtk.VTKObjectWrapper):
                 # VTK doesn't support strides, therefore we can't directly
                 # point to the underlying object
                 if data.flags.c_contiguous:
+                    # no reason to return a shallow copy if the array and name
+                    # are identical, just return the underlying array name
+                    if not deep_copy:
+                        if isinstance(name, str):
+                            if data.VTKObject.GetName() == name:
+                                return data.VTKObject
+
                     vtk_arr = copy_vtk_array(data.VTKObject, deep=deep_copy)
                     if isinstance(name, str):
                         vtk_arr.SetName(name)


adeak:
Hmm, I don't think I fully understand the issue, but I guess if the change makes the segfault go away it's definitely the right call.

But looking at the use case of the fix we can come up with an even more absurd test:
```py
import numpy as np
import pyvista

points = np.zeros((10000000, 3))
dataset = pyvista.PointSet(points)
data = np.arange(dataset.n_points, dtype=float)

dataset['scalars'] = data.copy()
dataset['scalars'] = dataset['scalars']
print(dataset['scalars'])
```

I can't really put my finger on the issue, because it seems to me that in the buggy case execution went like this:
https://github.com/pyvista/pyvista/blob/257f4aafd227ad7a6eeafb43737abe5cff5caeb1/pyvista/core/datasetattributes.py#L223

https://github.com/pyvista/pyvista/blob/257f4aafd227ad7a6eeafb43737abe5cff5caeb1/pyvista/core/datasetattributes.py#L614-L616

And here `vtk_arr` ended up being a shallow copy instead of a reference:
https://github.com/pyvista/pyvista/blob/257f4aafd227ad7a6eeafb43737abe5cff5caeb1/pyvista/core/datasetattributes.py#L775-L778

I would assume what's going on is that `vtkPointData` (through `DataSetAttributes`) sees that key `'scalars'` is being replaced, decreases some internal refcount of the original array, but since it's trying to add the same array back to itself (through the shallow copy) things blow up.

But then why does this not reproduce the segfault:
```py
import numpy as np
import pyvista

points = np.zeros((10000000, 3))
dataset = pyvista.PointSet(points)
data = np.arange(dataset.n_points, dtype=float)
dataset['scalars'] = data.copy()

# pure-pyvsta version
#dataset['scalars'] = dataset['scalars']

# VTK version
shallow = type(dataset['scalars'].VTKObject)()  # create shallow copy
shallow.ShallowCopy(dataset['scalars'].VTKObject)  # initialize shallow copy
shallow.SetName(dataset['scalars'].VTKObject.GetName())  # rename shallow copy
dataset.point_data.VTKObject.AddArray(shallow)  # (re-)add shallow copy
dataset.point_data.VTKObject.Modified()  # set to modified
print(dataset['scalars'])  # no explosion
```

pyvista/core/datasetattributes.py
```diff
@@ -772,6 +772,13 @@ def _prepare_array(
                 # VTK doesn't support strides, therefore we can't directly
                 # point to the underlying object
                 if data.flags.c_contiguous:
+                    # no reason to return a shallow copy if the array and name
+                    # are identical, just return the underlying array name
+                    if not deep_copy:
+                        if isinstance(name, str):
+                            if data.VTKObject.GetName() == name:
+                                return data.VTKObject
+
```
tkoyama010, line 781:
```suggestion
                    # no reason to return a shallow copy if the array and name
                    # are identical, just return the underlying array name
                    if not deep_copy and isinstance(name, str) and data.VTKObject.GetName() == name:
                        return data.VTKObject

```

tkoyama010 (changes requested):

akaszynski:
> Hmm, I don't think I fully understand the issue, but I guess if the change makes the segfault go away it's definitely the right call.

I spent a while trying every alternative to my proposal, and only this one did the trick.

> But then why does this not reproduce the segfault...

I find that perplexing as well, especially as we're effectively following the same steps as in `_prepare_array`.

Recommending that we proceed and debug as later should this ever crop up again.

adeak (approved):
> Recommending that we proceed and debug as later should this ever crop up again.

Good call, let's be pragmatic here. And let's be suspicious if someone mentions "segfault" in the near future.

tkoyama010 (approved):

PR #2877, opened 2022-06-26
Update label.yml

akaszynski:
Remove required labels. This is broken for fresh PRs and should be disabled until https://github.com/mheap/github-action-required-labels/issues/33 is resolved.

PR #2885, opened 2022-06-27
Rotate teapot for flip_z() doctest example

adeak:
The `flip_z()` doctest result looks like this:

two identical-looking teapots, before and after `flip_z()`

The teapots look identical because the xy plane in this configuration is parallel to the symmetry plane of the teapot.

If we first rotate the teapot around `x` we get a more meaningful result:

two teapots, one of them upside down

This makes the test slightly different from `flip_x()` and `flip_y()`, but it's probably more important to have the example be useful in isolation.

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,7 +1,7 @@
 name: "Pull Request Labeler"
 on:
   pull_request:
-    types: [opened, synchronize, reopened, labeled, unlabeled]
+    types: [opened, reopened]
   workflow_dispatch:
 
 jobs:
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -1196,6 +1196,7 @@ class DataSet(DataSetFilters, DataObject):
         >>> mesh2 = mesh1.scale([10.0, 10.0, 10.0], inplace=False)
         >>> _ = pl.add_mesh(mesh2)
         >>> pl.show(cpos="xy")
+
         """
         if isinstance(xyz, (float, int, np.number)):
             xyz = [xyz] * 3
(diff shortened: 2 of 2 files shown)


akaszynski (dismissed):
LGTM. I thought the labeler was fixed...

akaszynski:
> LGTM. I thought the labeler was fixed...

Actually, seems like this fails for forks... @tkoyama010, let's find a way to disable this for forks.

adeak:
At worst I can start using feature branches on the main repo. Vast majority of PRs seems to come from members anyway.

akaszynski:
> At worst I can start using feature branches on the main repo. Vast majority of PRs seems to come from members anyway.

I wouldn't want to discourage forks. Let's get it working for forks.

tkoyama010:
@akaszynski Could you please check that the setting is the follwing?

akaszynski:
> @akaszynski Could you please check that the setting is the follwing?

@banesullivan, are we fine with upping the scope on `GITHUB_TOKEN`? I doubt that we'll run into issues, but it's an additional security risk.

tkoyama010:
I reverted the trigger of labeling which is dropping in https://github.com/pyvista/pyvista/pull/2877.

banesullivan:
> @banesullivan, are we fine with upping the scope on GITHUB_TOKEN? I doubt that we'll run into issues, but it's an additional security risk.

I'm comfortable with this, just so long as we aren't using a Personal Access Token (PAT) which can have elevated permissions

akaszynski (dismissed):
Still LGTM.

tkoyama010 (approved):
LGTM

PR #2889, opened 2022-06-28
Test label action from forks

akaszynski:
This PR simply tests if the label action works for PRs from forks.


akaszynski:
@tkoyama010 and @adeak, this works from forks!

adeak:
I don't see any labels, but as long as it works :smile:

tkoyama010:
Great thanks.

tkoyama010:
> I don't see any labels, but as long as it works 😄

@adeak  It is because this branch didn't hit any rule of labeling. It works fine.

Issue #2832, opened 2022-06-20
All dragon in Plotting Themes section is black


tkoyama010:
### Describe the bug, what's wrong, and what you expected.

All dragon in [Plotting Themes](https://dev.pyvista.org/user-guide/themes.html) section is black.
| [docs.pyvista.org](https://docs.pyvista.org/user-guide/themes.html) | [dev.pyvista.org](https://dev.pyvista.org/user-guide/themes.html) |
| --- | --- |
| ![docs](https://user-images.githubusercontent.com/7513610/174690695-3c79db19-22d7-4d1f-8de1-1f555c13f9e6.png) | ![devs](https://user-images.githubusercontent.com/7513610/174690773-d484fcd9-370b-433e-be9f-caed67b64200.png) |

### Steps to reproduce the bug.

None

### System Information

```shell
None
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adeak:
I wonder if this is due to a `show_edges=True` somehow sneaking in. I wonder why that would only affect these plots... it seems that the theme changes don't affect the plots at all? That can't be right. (Un)fortunately my local build reproduces the issue.

tkoyama010:
It may be a matter of statistical observation, but I have the impression that the documentation often does not keep the release version status. Perhaps a comparison to the release version images needs to be added to the regression test.

PR #2575, opened 2022-05-09
Add split sharp edges to theme

akaszynski:
Add in the ability to set splitting of sharp edges in the global theme. This makes it easier to set default edge splitting for plots with sharp edges (like CAD-like geometry).

```py
>>> import pyvista
>>> pyvista.global_theme.sharp_edges_feature_angle = 45
>>> pyvista.global_theme.split_sharp_edges = True
>>> pyvista.global_theme.split_sharp_edges
True
```

### Bonus (there always is)
- Add a fixture to reset the global theme so it's not modified during docstring testing.
- Remove `doctest:+SKIP` within `theme.py` as it is no longer necessary due to proper theme resetting.

PR #2741, opened 2022-06-03
Correct reference to time values

mcgoldba:
### Overview

`self.reader_time_values` is undefined in the OpenFOAMReader class and was resulting in a `NameError` instead of the intended `ValueError`.

PR #2892, opened 2022-06-28
Skip theme doctests (again) to prevent theme changes leaking into doc build

adeak:
This is a stopgap until we figure out how to fix the underlying problem. (Plus an additional skip of a deprecated reader that keeps emitting warnings during doc building.)

As noted in https://github.com/pyvista/pyvista/issues/2832, our dragons turned black in the devdocs. Similarly, https://docs.pyvista.org/user-guide/data_model.html#cell-data used to have plots like

data_model-7_00_00

But in the devdocs they look like https://dev.pyvista.org/user-guide/data_model.html#cell-data

data_model-7_00_00

I bisected the history and traced the bug back to https://github.com/pyvista/pyvista/pull/2575. Clearly the unskipped doctests are affecting the global theme, e.g. by setting the window size to `[400, 400]` and showing all edges (which makes the dragons black).

---

We have machinery to prevent this from happening:

https://github.com/pyvista/pyvista/blob/6bdedb7fda4937c7709777307777e5e414326ce5/doc/conf.py#L289-L306

https://github.com/pyvista/pyvista/blob/6bdedb7fda4937c7709777307777e5e414326ce5/doc/conf.py#L342-L343

What this _should_ do is run the reset function before and after every example, preventing leakage. However, it turns out that this only runs for literal examples inside `examples/`, but not for figures generated from doctests, nor for things in e.g. `doc/user_guide`. So what happens is that the doctests mess with the global plot theme, which then affects building the `doc/user_guide` and friends.

We also have
https://github.com/pyvista/pyvista/blob/6bdedb7fda4937c7709777307777e5e414326ce5/pyvista/conftest.py#L15-L21
but I confirmed that this doesn't get run during sphinx' doctest runs.

---

My hunch is that this is a bug or missing feature in sphinx-gallery, or there's a switch we could use to enforce resetting for _everything_ being built. In any case I think we should first skip these theme tests for the time being, so that the impending release will have unbroken figures.

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -18,28 +18,35 @@ jobs:
     - uses: actions-ecosystem/action-add-labels@v1
       if: startsWith(github.event.pull_request.head.ref, 'doc/') || startsWith(github.event.pull_request.head.ref, 'docs')
       with:
+        github_token: "${{ secrets.GITHUB_TOKEN }}"
         labels: documentation
     - uses: actions-ecosystem/action-add-labels@v1
       if: startsWith(github.event.pull_request.head.ref, 'docker')
       with:
+        github_token: "${{ secrets.GITHUB_TOKEN }}"
         labels: docker
     - uses: actions-ecosystem/action-add-labels@v1
       if: startsWith(github.event.pull_request.head.ref, 'maint') || startsWith(github.event.pull_request.head.ref, 'no-ci') || startsWith(github.event.pull_request.head.ref, 'ci')
       with:
+        github_token: "${{ secrets.GITHUB_TOKEN }}"
         labels: maintenance
     - uses: actions-ecosystem/action-add-labels@v1
       if: startsWith(github.event.pull_request.head.ref, 'junk')
       with:
+        github_token: "${{ secrets.GITHUB_TOKEN }}"
         labels: ignore-for-release
     - uses: actions-ecosystem/action-add-labels@v1
       if: startsWith(github.event.pull_request.head.ref, 'feat')
       with:
+        github_token: "${{ secrets.GITHUB_TOKEN }}"
         labels: enhancement
     - uses: actions-ecosystem/action-add-labels@v1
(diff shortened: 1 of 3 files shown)


akaszynski:
The labeler, oh the labeler.

tkoyama010:
See https://github.community/t/github-actions-are-severely-limited-on-prs/18179/23

tkoyama010:
It is weird that it is working well in  https://github.com/pyvista/pyvista/pull/2741 even though it is the same setup.

tkoyama010:
@adeak The next time you have the opportunity to create a Pull Request, may I ask you to submit it without labeling it before creating it?

MatthewFlamm:
I'm +1 on fixing the issue quickly here especially if there is a release coming soon.

If possible, we need to place the same reset machinery into sphinx doctest that `pytest --doctest-modules` has.

I have some ideas on how this could be fixed in sphinx directly, but I'm not confident it is a straightforward 1 day kind of fix (although it might be).

MatthewFlamm:
> I'll take this to mean that this is not currently supported (rather than just me not being able to figure out how to do this with sphinx).

It _might_ be possible currently using https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#confval-doctest_global_setup, but I'm not yet clear on the exact granularity of when this functionality executes.

MatthewFlamm:
Ah, It could also be that this only runs in "doctest" mode, but not "documentation building" mode?

adeak:
I think I have a fix in https://github.com/pyvista/pyvista/pull/2896, it's hacky but it uses an already hacky part of our config. Not sure it's correct yet, but it's my best attempt so far.

adeak:
Marking this as draft to put it in limbo until https://github.com/pyvista/pyvista/pull/2896 is merged, will close if that's what's kept.

adeak:
Closing as superseded by https://github.com/pyvista/pyvista/pull/2907

PR #2896, opened 2022-06-28
Add non-included plot directive that resets theme

adeak:
This is an alternative to https://github.com/pyvista/pyvista/pull/2892, resolves https://github.com/pyvista/pyvista/issues/2832.

Using the same directive injection that we use anyway for examples we can add non-rendered setup blocks before and after each API example block that resets the document theme.

In a local test this seemed to fix the issue without breaking anything else, but since then I've also added the same fix _after_ each example block. We'll have to take a close look at the build artifact to see if everything works as intended.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -369,9 +369,19 @@ def _str_examples(self):
     elif re.search(IMPORT_PYVISTA_RE, examples_str) and 'plot-pyvista::' not in examples_str:
         out = []
         out += self._str_header('Examples')
+        # inject hack to reset theme before every example block
+        out += ['.. pyvista-plot::', '   :include-source: False', '']
+        out += self._str_indent(
+            ['>>> import pyvista; pyvista.set_plot_theme("document"); del pyvista', '']
+        )
         out += ['.. pyvista-plot::', '']
         out += self._str_indent(self['Examples'])
         out += ['']
+        # inject same hack after every example block
+        out += ['.. pyvista-plot::', '   :include-source: False', '']
+        out += self._str_indent(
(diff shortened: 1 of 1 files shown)


tkoyama010 (changes requested):
Great idea, but could it be added to the sphinx extension provided by PyVista? I think this is also needed for users of the extension `pyvista.ext.plot_directive`. I think this solution would be the best solution if it takes time to do so since this release is so close.

adeak:
> Great idea, but could it be added to the sphinx extension provided by PyVista? I think this is also needed for users of the extension `pyvista.ext.plot_directive`. I think this solution would be the best solution if it takes time to do so since this release is so close.

I can try to do that, but then we'll have to think about two things:
1. What if someone downstream _wants_ changes to persist across code blocks?
2. Do we have to allow customization of the plot theme to reset to?

For point 2., the best would be to store the state of the theme before running an example block, and restoring the global theme from that after.

tkoyama010 (approved):
I see. I'll deal with the current situation with this solution and leave this matter as an issue. LGTM.

adeak:
OK, I followed your suggestion in https://github.com/pyvista/pyvista/pull/2907 @tkoyama010, thanks for the tip. If that works this should be closed too.

akaszynski:
Is this superseded by #2907?

adeak:
> Is this superseded by https://github.com/pyvista/pyvista/pull/2907?

If we're happy with the kind of solution in #2907 (pending future improvements) then yes, along with the PR this would supersede.

PR #2897, opened 2022-06-28
Add DataSet.cell_point_ids

MatthewFlamm:
### Overview

Adds method to obtain point ids from a cell id.  This addresses #2838.  

### Details

There was some discussion in #2838 on whether we could instead have a more full featured `vtkCell` implementation.  We could deprecate all the `cell_*` methods that implement one-at-a-time functionality from `vtkCell` at that time.

### DIFF
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -2552,3 +2552,33 @@ class DataSet(DataSetFilters, DataObject):
 
         """
         return self.GetCellType(ind)
+
+    def cell_point_ids(self, ind: int) -> List[int]:
+        """Return the point ids in a cell.
+
+        Parameters
+        ----------
+        ind : int
+            Cell ID.
+
+        Returns
+        -------
+        list[int]
+            Point Ids that are associated with the cell.
+
+        Examples
+        --------
+        >>> from pyvista import examples
+        >>> mesh = examples.load_airplane()
+        >>> mesh.cell_type(0)
+        5
+
+        Cell type of 5 is a triangular cell with three points.
+
+        >>> mesh.cell_point_ids(0)
+        [0, 1, 2]
+
+        """
+        cell = self.GetCell(ind)
+        point_ids = cell.GetPointIds()
+        return [point_ids.GetId(i) for i in range(point_ids.GetNumberOfIds())]
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -2573,7 +2573,7 @@ class DataSet(DataSetFilters, DataObject):
         >>> mesh.cell_type(0)
         5
 
-        Cell type of 5 is a triangular cell with three points.
+        Cell type 5 is a triangular cell with three points.
 
         >>> mesh.cell_point_ids(0)
         [0, 1, 2]
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -1107,6 +1107,7 @@ def test_cell_point_ids(grid):
     assert isinstance(point_ids, list)
     assert len(point_ids) == grid.cell_n_points(0)
     assert all([isinstance(id, int) for id in point_ids])
+    assert all([0 <= id < grid.n_points for id in point_ids])
 
 
 def test_cell_bounds(grid):
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -1102,6 +1102,13 @@ def test_cell_points(grid):
     assert points.shape[1] == 3
 
 
+def test_cell_point_ids(grid):
+    point_ids = grid.cell_point_ids(0)
+    assert isinstance(point_ids, list)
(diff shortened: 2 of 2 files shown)


banesullivan (approved):

adeak (approved):
One potential minor thing, LGTM regardless. Thanks!

pyvista/core/dataset.py
```diff
@@ -2552,3 +2552,33 @@ def cell_type(self, ind: int) -> int:
...
+        """Return the point ids in a cell.
+
+        Parameters
+        ----------
+        ind : int
+            Cell ID.
+
+        Returns
+        -------
+        list[int]
```
adeak, line 2566: Actually, one question too: should we extend this to sequence inputs, returning a list of lists? Some cell/point-related filters do that, then again those are probably different (return meshes, probably). I don't know if this would make sense for use cases.

tests/test_dataset.py
```diff
@@ -1102,6 +1102,13 @@ def test_cell_points(grid):
     assert points.shape[1] == 3
 
 
+def test_cell_point_ids(grid):
+    point_ids = grid.cell_point_ids(0)
+    assert isinstance(point_ids, list)
+    assert len(point_ids) == grid.cell_n_points(0)
+    assert all([isinstance(id, int) for id in point_ids])
+
```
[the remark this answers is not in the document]
banesullivan: IMO, this straddles the line of what PyVista should test vs. what PyVista should trust is tested upstream in VTK. Nonetheless, I'm +1 for including this.

pyvista/core/dataset.py
banesullivan: @adeak, are you suggesting a use case like:

(not actually implemented)
```py
>>> mesh.cell_point_ids([0, 3, 5])
[[0, 1, 2], [0, 4, 6], [8, 10, 12]]
```

banesullivan: I'd like to include that here if there's a clean way to vectorize this.

Further, I think it'd be nice to expand many of these methods that accept point/cell IDs to accept slices and even `Elipses` (`...`) -- but this is worthy of a whole separate PR and discussion

tests/test_dataset.py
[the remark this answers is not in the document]
adeak: Fair enough.

pyvista/core/dataset.py
MatthewFlamm: I think we should think through what is the ideal API design before implementing here.  One thought we discussed is to have an API that exposes a wrapper for `vtkCell`.  We could have a property that allows access to the cells themselves, for example `mesh.cells[0]`.  Then, we could do something like `[cell.point_ids for cell in mesh.cells[0:5]]`.  Another thought is, like you suggest, to allow more flexibility in the existing methods.

adeak: > @adeak, are you suggesting a use case like

Yup.

> I'd like to include that here if there's a clean way to vectorize this.

I doubt it, because cells can have different sizes, so thebresult can't be a (decent) 2d array in general.

pyvista/core/dataset.py
```diff
@@ -2552,3 +2552,33 @@ def cell_type(self, ind: int) -> int:
...
+            Point Ids that are associated with the cell.
+
+        Examples
+        --------
+        >>> from pyvista import examples
+        >>> mesh = examples.load_airplane()
+        >>> mesh.cell_type(0)
+        5
+
+        Cell type of 5 is a triangular cell with three points.
```
tkoyama010, line 2576:
```suggestion
        Cell type 5 is a triangular cell with three points.
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM. This is an important method, especially for a CAE person.

pyvista/core/dataset.py
adeak: To close this up, I think this is fine for now. We will either supersede all this with a full-fledged cells attribute (or similar) making this method probably obsolete, or we will keep this around and we can extend it to the sequence-input case later. In other words, I don't think we'd need this method name for something fundamentally different in the future, so we can add this as-is to cover the linked issue and then see where things go.

adeak (approved):

pyvista/core/dataset.py
MatthewFlamm: I agree, for future consideration I plan on submitting a method for checking whether a point is inside a cell.  You could imagine providing either a sequence of cells or a sequence of points.  Does it makes sense to allow both as sequences?

adeak: Hmm, good question. I can't tell how common that use case might be, but when someone _does_ need it, having an implementation for that would be useful. Getting the logistics right for a relatively efficient implementation might make that worthwhile.

PR #2898, opened 2022-06-28
implement gaussian_points representation style

### LINKED ISSUE #2891: Point Guassian representation style
See https://github.com/pyvista/pyvista/discussions/2576

add support for ParaView's Point Gaussian representation style for rendering massive point clouds

banesullivan:
*work in progress*

Resolves #2891 for #2576, cc @Findus23

To do:

- [x] Solicit feedback
- [x] Check mapper changes
- [x] Add an example to the gallery with a full explanation
- [x] Improve auto choice of `gaussian_radius` parameter
- [x] Support more than just `PolyData`

Demo:

```py
import pyvista as pv
import numpy as np

array = np.genfromtxt('halo_zoom.csv', skip_header=1, delimiter=',')
pdata = pv.PolyData(array)

p = pv.Plotter(notebook=0)
p.set_background('k')
p.add_mesh(pdata, color='grey', style='points_gaussian', gaussian_radius=0.002)
p.show()
```
Screen Shot 2022-06-28 at 10 04 15 AM

versus a standard rendering of this point cloud:

```py
p = pv.Plotter(notebook=0)
p.set_background('k')
p.add_mesh(pdata, color='grey')
p.show()
```

Screen Shot 2022-06-28 at 10 04 50 AM

Verifying that it looks the same in ParaView:

Screen Shot 2022-06-28 at 10 12 03 AM


banesullivan:
I'm also thinking that it may be worth implementing this as its own representation style much like how ParaView does it: https://github.com/Kitware/ParaView/blob/master/Remoting/Views/vtkPointGaussianRepresentation.cxx

Right now, my implementation only works for PolyData, but we want this to generally work for all data types

akaszynski:
Worked on this because it's friggen cool:

tmp


```py
import pyvista as pv
import numpy as np

points = np.random.random((1000, 3))
rgba = points - points.min(axis=0)
rgba /= rgba.max(axis=0)

pl = pv.Plotter(shape=(2, 2))

# Standard points
actor = pl.add_points(
   points,
   style='points',
   emissive=False,
   scalars=rgba,
   rgba=True,
   point_size=10,
)

# Gaussian points
pl.subplot(0, 1)
actor = pl.add_points(
   points,
   style='points_gaussian',
   emissive=False,
   scalars=rgba,
   rgba=True,
   opacity=0.99,
   point_size=10,
)

# Gaussian points
pl.subplot(1, 0)
actor = pl.add_points(
   points,
   style='points_gaussian',
   emissive=True,
   scalars=rgba,
   rgba=True,
   # opacity=0.999999,  # does not work !!!
   point_size=10,
)


pl.background_color = 'k'
pl.link_views()
# pl.reset_camera()
pl.camera_position = 'iso'
pl.camera.zoom(1.3)
pl.show()


```

akaszynski:
Broke render points as spheres with `points_gaussian`; I'll clean it up later.

Also, weird triangles show up when `emissive` is `False`, and setting opacity to slightly less than 1.0 cleans it up.

Findus23:
> Also, weird triangles show up when emissive is False, and setting opacity to slightly less than 1.0 cleans it up.

Oh, that's fascinating, because exactly the same happens for me with Paraview. The default opacity of 1 shows each point as a weird triangle (with the blurred disk inside without any transparency). And just changing the setting to any other value (even if it is close to 1) fixes it.

akaszynski:
@Findus23, where does the "halo_high_res.csv" dataset from #2576 come from? I'd like to include it as an example because it's great.

```py
import pyvista
import numpy as np
import vtk

array = np.genfromtxt('/home/akaszyns/Downloads/halo_high_res.csv', skip_header=1, delimiter=',')
pdata = pyvista.PolyData(array)


pdata.plot(
    style='points_gaussian',
    color='#fff7c2',
    opacity=0.1,
    point_size=0.4,
    background='k',
    anti_aliasing='ssao',
    zoom=4,
)
```

tmp

akaszynski:
orbit

akaszynski:
> The particles correspond to (same mass) particles in a cosmological simulation and filtered to only show one resulting dark matter halo.

I've added your datasets in https://github.com/pyvista/vtk-data/tree/master/Data/point-clouds/findus23, please let me know if you have any issue with this. We were hoping to license it according to the MIT license.

jorgepiloto:
The following script may be interesting for this pull-request. It uses the HYG 3.0 catalogue and converts desired information into Astropy objects. This allows to quickly convert those to any desired frame.

<script src="https://gist.github.com/jorgepiloto/847ee0e594b243cc7abc24d6771331b9.js"></script>

akaszynski:
Thanks @jorgepiloto. Turns out space math is hard.
---
tmp

Final source used for the conversion is at https://github.com/pyvista/vtk-data/tree/master/Data/point-clouds/hyg-database.


```py
from pyvista import examples

stars = examples.download_stars_cloud_hyg()

stars.plot(
    style='points_gaussian',
    background='k',
    point_size=0.5,
    scalars='_rgba',
    render_points_as_spheres=False,
    zoom=3.0,
)
```

akaszynski:
@banesullivan, I've made several changes to your original PR, but kept the spirit of it as much as possible. Would appreciate a quick review so we can get this out the door.

Major Changes:
- Removed `gaussian_radius`. Since this effectively is `point_size` + opacity, I decided to keep things simple by removing it and allowing the user to play around with point size and opacity instead.
- Added a variety of "sensible defaults" given the issues encountered with the trifecta combination of `render_points_as_spheres`, `emissivity`, and `opacity`.

pyvista/plotting/plotting.py
```diff
@@ -3101,19 +3127,33 @@ def add_mesh(
...
             line_width=line_width,
             culling=culling,
         )
+
+        if style == 'points_gaussian':
+            self.mapper.emissive = emissive
+            self.mapper.scale_factor = point_size * self.mapper.dataset.length / 1300
+            if not render_points_as_spheres and not emissive:
+                if opacity >= 1.0:
+                    opacity = 0.9999  # otherwise, weird triangles
```
banesullivan, line 3143: Should we report this upstream since it's happening in ParaView as well?

banesullivan (commented):
Really excellent work cleaning this up and taking it the extra mile after finishing it. I'm really liking the way this is done with the mapper classes

I'm not allowed to approve since I created the PR, but happy to rubber stamp your work, @akaszynski

pyvista/plotting/plotting.py
akaszynski: Yep, but I'm scared of gitlab :)

akaszynski (approved):

pyvista/plotting/plotting.py
banesullivan: 🤣

banesullivan: Not worth raising an issue at this time after a bit of reflection

Findus23:
I'm currently not available at the moment, but when I am back, I will try this out and report back.
Many thanks for the amazing work!
Regarding the dataset, I don't see an issue publishing it under a free license. But maybe once I am back, I will create a PR replacing it with a more properly documented one as these files were just random halos I was currently looking at.

PR #2905, opened 2022-06-29
Add method to check if point is inside cell

MatthewFlamm:
### Overview

Add method to check if point is inside a specific cell.  This was motivated partly by the discussion in https://github.com/pyvista/pyvista/discussions/2703#discussioncomment-2828184

### Details

Related to #2897.

Should we allow a Sequence of points (if more than one is provided we can return a numpy ndarray)?  If so, this might complicate any plans to allow for Iterables, Sequences, or similar of cell indices.


pyvista/core/dataset.py
```diff
@@ -2523,3 +2524,49 @@ def cell_type(self, ind: int) -> int:
 
         """
         return self.GetCellType(ind)
+
+    def cell_is_point_inside(self, ind: int, point: Vector) -> bool:
```
tkoyama010, line 2528:
```suggestion
    def point_is_inside_cell(self, ind: int, point: Vector) -> bool:
```
I guess this is more clear what the method is doing.

pyvista/core/dataset.py
```diff
@@ -2523,3 +2524,49 @@ def cell_type(self, ind: int) -> int:
...
+        """Return whether a point is inside a cell.
+
+        Parameters
+        ----------
+        ind : int
+            Cell type ID.
```
tkoyama010, line 2534: Is this Cell ID ?
```suggestion
            Cell ID.
```

tkoyama010 (changes requested):
Thanks for adding a useful method.

pyvista/core/dataset.py
MatthewFlamm: Thanks, this is indeed better.  I was trying to be thematically similar to the `DataSet.cell_*` methods like `DataSet.cell_bounds` as it indicates it is operating on a single cell basis.  But I think you name is the most clear.  We can have a `See Also` section potentially to address my thinking.

I will clean up the usage after committing your suggestion directly.

MatthewFlamm:
Unlike `DataSet.find_containing_cell`, there is only minor efficiency gained by allowing multiple points to be supplied here.  The inefficiency here is multiple calls to `GetCell(ind)` and `GetPoints().GetNumberOfPoints()` (I'm assuming the calls to `vtk.mutable` are cheap), which would be avoided if we allowed multiple points to be supplied to this function.  Should we implement multiple points here?  I'm not sure based on some discussions about also supporting multiple cells here https://github.com/pyvista/pyvista/pull/2897#discussion_r908767542.  If we support multiple points and multiple cells, then it becomes unclear, at least to me, what the 2D output structure should look like.

cc @adeak  and @banesullivan

tkoyama010 (approved):
We need to wait for a reply to the above, otherwise, it is LGTM.

pyvista/core/dataset.py
```diff
@@ -2523,3 +2522,49 @@ def cell_type(self, ind: int) -> int:
...
+        --------
+        >>> from pyvista import examples
+        >>> mesh = examples.load_hexbeam()
+        >>> mesh.cell_bounds(0)
+        (0.0, 0.5, 0.0, 0.5, 0.0, 0.5)
+        >>> mesh.point_is_inside_cell(0, [0.2, 0.2, 0.2])
+        True
+
+        """
+        if ind < 0 or ind >= self.n_cells:
```
adeak, line 2552: Nitpick:
```suggestion
        if not 0 <= ind < self.n_cells:
```
at least I always find it easier to reason about a single interval, despite the negation.

pyvista/core/dataset.py
```diff
@@ -2523,3 +2522,49 @@ def cell_type(self, ind: int) -> int:
...
+            raise ValueError(f"ind must be >= 0 and < {self.n_cells}, got {ind}")
+
+        if not isinstance(point, (collections.abc.Sequence, np.ndarray)):
+            raise ValueError(f"point must be a Sequence or ndarray, got {point}")
+
+        if len(point) != 3:
+            raise ValueError(f"point must have length 3, got {len(point)}")
```
adeak, line 2559: One pro for allowing multiple points: we could use your `_coerce_pointslike_arg()` to do input validation.

pyvista/core/dataset.py
```diff
@@ -2523,3 +2522,49 @@ def cell_type(self, ind: int) -> int:
...
+
+        cell = self.GetCell(ind)
+        npoints = cell.GetPoints().GetNumberOfPoints()
+
+        closest_point = [0.0, 0.0, 0.0]
+        sub_id = _vtk.mutable(0)
+        pcoords = [0.0, 0.0, 0.0]
+        dist2 = _vtk.mutable(0.0)
```
adeak, line 2567: If we don't use these mutables after the call, do they still have to be mutable? Do we get a segfault otherwise?

adeak (commented):
> Unlike `DataSet.find_containing_cell`, there is only minor efficiency gained by allowing multiple points to be supplied here. The inefficiency here is multiple calls to `GetCell(ind)` and `GetPoints().GetNumberOfPoints()` (I'm assuming the calls to `vtk.mutable` are cheap), which would be avoided if we allowed multiple points to be supplied to this function. Should we implement multiple points here? I'm not sure based on some discussions about also supporting multiple cells here [#2897 (comment)](https://github.com/pyvista/pyvista/pull/2897#discussion_r908767542). If we support multiple points and multiple cells, then it becomes unclear, at least to me, what the 2D output structure should look like.

My main reason for contemplating multiple points/cells as input is not that of performance (I don't typically associate VTK with vectorization), but rather convenience. If one has more than one point to check it would be nicer to call the method once with every point, rather than using something like a list comprehension outside the function, point by point.

This ties in with your question: if the API becomes confusing in the "multiple points, multiple cells" case then we should just forbid it. It's not convenient to have to look up the docs to see how it behaves.

That being said, I could imagine that if either points or cells are multiple then we return an array of bools (as large as the number of input points or cells), and if both are sequences then we return a 2d array of shape `(n_points, n_indices)` or `(n_indices, n_points)`. But again, this might be too smart for its own good.

MatthewFlamm:
I agree on the convenience factor when using one or the other.  I find the convenience factor will drop when using both together.  I will never remember which comes first in the indices.  But, maybe this is a rare enough use case that it isn't worth trying to over optimize the API.  I'll implement multiple points

pyvista/core/dataset.py
MatthewFlamm: A bit better than a segfault, but a `TypeError`:

```
E   TypeError: EvaluatePosition argument 3: a vtk.reference() object is required
```

tkoyama010 (approved):

pyvista/core/dataset.py
```diff
@@ -2523,3 +2523,54 @@ def cell_type(self, ind: int) -> int:
...
+        weights = [0.0] * npoints
+
+        in_cell = [
+            bool(cell.EvaluatePosition(node, closest_point, sub_id, pcoords, dist2, weights))
+            for node in point
+        ]
+        if len(in_cell) == 1:
+            return in_cell[0]
```
adeak, line 2575: In the edge case of someone passing in a `(1, 3)`-shaped arraylike, is it OK that this will return `True` instead of `[True]`? If the API says "return a scalar for a scalar", it might be more consistent to respect if the input is `[(0, 0, 0)]` instead of `(0, 0, 0)`. But unpacking the singular case is fine too, just something to consider.

pyvista/core/dataset.py
```diff
@@ -2523,3 +2523,54 @@ def cell_type(self, ind: int) -> int:
...
+        >>> from pyvista import examples
+        >>> mesh = examples.load_hexbeam()
+        >>> mesh.cell_bounds(0)
+        (0.0, 0.5, 0.0, 0.5, 0.0, 0.5)
+        >>> mesh.point_is_inside_cell(0, [0.2, 0.2, 0.2])
+        True
+
+        """
+        if not 0 <= ind < self.n_cells:
+            raise ValueError(f"ind must be >= 0 and < {self.n_cells}, got {ind}")
```
adeak, line 2557: Should we add a `TypeError` for non-integer `ind`? I don't insist on doing that since most weird cases will fail loudly with this check. Only the "passed in a float by accident" case might be relevant for this, assuming that such accidental floats are not int-valued floats.

pyvista/core/dataset.py
MatthewFlamm: Good point.  We currently have this problem in the other functions too: 

https://github.com/pyvista/pyvista/blob/f8208ad25c518d3338e57a821a9bd0772286b08b/pyvista/core/dataset.py#L2285

MatthewFlamm: I think this is important to fix.  A user might be dynamically generating a list of points to use, it will be unexpected that they should have to check for arrays of shape (1, 3).

Since _points_like_arg is a private method, we have free reign to change it.  I'm not sure if we should take extra steps to warn users that the behavior has changed, or not.  I argue that the current behavior is a bug, so we can just change it.

adeak: Although fixing the bug mostly means that some ifs in downstream code will merely be redundant, we should probably annotate the involved public functions with `.. versionadded:`. Otherwise people may never notice.

adeak: If you meant emitting warnings: it could go either way. Not strictly necessary, but this is a very high signal to noise ratio warning.

MatthewFlamm: > If you meant emitting warnings: it could go either way. Not strictly necessary, but this would be a very high signal to noise ratio warning.

One issue is that if we emit a warning in `_coerce_pointslike_arg` for an array of shape `(1, 3)` to warn users of the behavior change, it means that even after the user adjusts their downstream code, it will still emit the warning.  I like the approach of adding to the docstring a clear description of the change, so users can figure out what happened.

adeak: Yeah, I realized that too eventually. It would be useful the first time but useless afterward. Agreed on the doc annotation.

pyvista/core/dataset.py
```diff
@@ -2294,6 +2298,10 @@ def find_containing_cell(
             Index or indices of the cell in this mesh that contains
             the given point.
 
+            .. versionchanged:: 0.35.0
+            Inputs of shape (1, 3) now return a numpy.ndarray
+
+
```
MatthewFlamm, line 2304:
```suggestion
```

pyvista/utilities/common.py
```diff
@@ -9,23 +9,26 @@
...
+            is not ``np.ndarray``.
+
+        Returns
+        -------
+        points : np.ndarray
+            Size (n, 3) array.
+        singular : bool
+            Whether the input array was a single point. This excludes an array with
+            length 1.
+    x
```
MatthewFlamm, line 31:
```suggestion

```

pyvista/core/dataset.py
MatthewFlamm: I used `..versionchanged:: 0.35.0`, but I'm not 100% confident that this is formatted correctly in the right place.

pyvista/utilities/common.py
```diff
@@ -23,9 +23,11 @@ def _coerce_pointslike_arg(
 
     Returns
     -------
-    np.ndarray
+    points : np.ndarray
         Size (n, 3) array.
-
+    singular : bool
+        Whether the input array was a single point. This excludes an array with
+        length 1.
```
adeak, line 30: This is definitely subjective, but I'd prefer to be exact about the distinction, because it should be unambiguous. Something like
```suggestion
        Whether the input was a single point. This excludes a 2d array-like with
        shape ``(1, 3)``.
```

Or may be go even more direct and just say "Whether the input was a single point in an array-like with shape `(3,)`".

pyvista/utilities/common.py
```diff
@@ -23,9 +23,11 @@ def _coerce_pointslike_arg(
 
     Returns
     -------
-    np.ndarray
+    points : np.ndarray
```
adeak, line 26: I don't know how consistent the codebase is in this, but there's precedent of keeping only types here even if there are multiple return values, e.g. https://github.com/pyvista/pyvista/blob/709293784756e7f0fb4f4125a95fafc02eab0da5/pyvista/core/filters/poly_data.py#L3181-L3189

pyvista/core/dataset.py
```diff
@@ -2187,11 +2187,17 @@ def find_closest_cell(
             Index or indices of the cell in this mesh that is/are closest
             to the given point(s).
 
+            .. versionchanged:: 0.35.0
+               Inputs of shape (1, 3) now return a numpy.ndarray
```
adeak, line 2191:
```suggestion
               Inputs of shape ``(1, 3)`` now return a :class:`numpy.ndarray`
               of shape ``(1,)``.
```

pyvista/core/dataset.py
```diff
@@ -2187,11 +2187,17 @@ def find_closest_cell(
...
+
         numpy.ndarray
             Point or points inside a cell of the mesh that is/are closest
             to the given point(s).  Only returned if
             ``return_closest_point=True``.
 
+            .. versionchanged:: 0.35.0
+               Inputs of shape (1, 3) now return a numpy.ndarray
```
adeak, line 2199:
```suggestion
               Inputs of shape ``(1, 3)`` now return a :class:`numpy.ndarray`
               of the same shape.
```

pyvista/core/dataset.py
```diff
@@ -2309,6 +2313,9 @@ def find_containing_cell(
             Index or indices of the cell in this mesh that contains
             the given point.
 
+            .. versionchanged:: 0.35.0
+               Inputs of shape (1, 3) now return a numpy.ndarray
```
adeak, line 2317:
```suggestion
               Inputs of shape ``(1, 3)`` now return a :class:`numpy.ndarray`
               of shape ``(1,)``.
```

pyvista/core/dataset.py
```diff
@@ -2568,3 +2575,61 @@ def cell_point_ids(self, ind: int) -> List[int]:
         cell = self.GetCell(ind)
         point_ids = cell.GetPointIds()
         return [point_ids.GetId(i) for i in range(point_ids.GetNumberOfIds())]
+
+    def point_is_inside_cell(
+        self, ind: int, point: Union[VectorArray, NumericArray]
+    ) -> Union[int, np.ndarray]:
+        """Return whether a point is inside a cell.
```
adeak, line 2582:
```suggestion
        """Return whether one or more points are inside a cell.
```

pyvista/core/dataset.py
```diff
@@ -2568,3 +2575,61 @@ def cell_point_ids(self, ind: int) -> List[int]:
         cell = self.GetCell(ind)
         point_ids = cell.GetPointIds()
         return [point_ids.GetId(i) for i in range(point_ids.GetNumberOfIds())]
+
+    def point_is_inside_cell(
+        self, ind: int, point: Union[VectorArray, NumericArray]
```
adeak, line 2580: How strong feelings do you have about having the cell index first and the point(s) second? I'm wondering if we should flip the order. Or are there other similar filters that this might match?

pyvista/core/dataset.py
```diff
@@ -2568,3 +2575,61 @@ def cell_point_ids(self, ind: int) -> List[int]:
...
+        """Return whether a point is inside a cell.
+
+        .. versionadded:: 0.35.0
+
+        Parameters
+        ----------
+        ind : int
+            Cell ID.
+
+        point : Sequence(float) or np.ndarray
```
adeak, line 2591: I don't really know the specification of these fields, but how standard is `Sequence(float)`? I can only see this in `_coerce_pointslike_arg()` and `find_containing_cell()`, `find_closest_cell()`. If it's like `typing.Sequence` then perhaps it should be `Sequence[float]`?

OK, we have both in https://dev.pyvista.org/api/core/_autosummary/pyvista.DataSet.find_closest_cell.html?highlight=find_closest_cell#pyvista.DataSet.find_closest_cell vs https://dev.pyvista.org/api/core/_autosummary/pyvista.core.common_data.perlin_noise.html?highlight=perlin_noise, and neither get picked up by intersphinx so it probably doesn't really matter either way.

pyvista/core/dataset.py
```diff
@@ -2568,3 +2575,61 @@ def cell_point_ids(self, ind: int) -> List[int]:
...
+            Coordinates of point to query (length 3) or a ``numpy`` array of ``n``
+            points with shape ``(n, 3)``.
+
+        Returns
+        -------
+        bool or numpy.ndarray
+            Whether point(s) is/are inside cell.
```
adeak, line 2598:
```suggestion
            Whether point(s) is/are inside cell. A scalar bool is only returned if
            the input point has shape ``(3,)``.
```

pyvista/core/dataset.py
```diff
@@ -2568,3 +2575,61 @@ def cell_point_ids(self, ind: int) -> List[int]:
...
+        --------
+        >>> from pyvista import examples
+        >>> mesh = examples.load_hexbeam()
+        >>> mesh.cell_bounds(0)
+        (0.0, 0.5, 0.0, 0.5, 0.0, 0.5)
+        >>> mesh.point_is_inside_cell(0, [0.2, 0.2, 0.2])
+        True
+
+        """
+        if not isinstance(ind, int):
```
adeak, line 2610: Sorry to keep bikeshedding this, but numpy ints might pop up their head. What do you think?
```suggestion
        if not isinstance(ind, (int, np.integer)):
```

pyvista/core/dataset.py
```diff
@@ -2568,3 +2575,61 @@ def cell_point_ids(self, ind: int) -> List[int]:
...
+
+        closest_point = [0.0, 0.0, 0.0]
+        sub_id = _vtk.mutable(0)
+        pcoords = [0.0, 0.0, 0.0]
+        dist2 = _vtk.mutable(0.0)
+        weights = [0.0] * npoints
+
+        in_cell = np.array(
+            [
+                bool(cell.EvaluatePosition(node, closest_point, sub_id, pcoords, dist2, weights))
```
adeak, line 2629: [VTK docs](https://vtk.org/doc/nightly/html/classvtkCell.html#a188d8c01ab1cc03e1fd8b04e3ec97faf) says this can return -1 in case of a "computational problem". We might have to pull out the `bool` conversion, and raise/warn-and-set-to-NaN any broken values. Although the NaN setting would be a but problematic if we're aiming for a bool array.

pyvista/core/dataset.py
adeak: I see you've pushed a corresponding chance since. Looks good to me now, resolving.

pyvista/utilities/common.py
akaszynski: That's the precedent, but we can change this if we feel that it needs to be changed. numpydoc specifies that this is fine.

Personally I like how clean it looks without the variable return name, and it also makes maintenance a bit easier.

tests/utilities/test_common.py
```diff
@@ -9,57 +9,68 @@
...
     assert coerced_arg.shape == (1, 3)
     assert np.array_equal(coerced_arg, np.array([[1.0, 2.0, 3.0]]))
+    assert singular
+
+    # Test with (1, 3) np.ndarray
+    point = np.array([[1.0, 2.0, 3.0]])
+    coerced_arg, singular = _coerce_pointslike_arg(point)
+    assert isinstance(coerced_arg, np.ndarray)
+    assert coerced_arg.shape == (1, 3)
+    assert np.array_equal(coerced_arg, np.array([[1.0, 2.0, 3.0]]))
```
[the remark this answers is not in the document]
MatthewFlamm: I don't have a strong opinion on this.  My thought was the above usages cannot use the variable, so I chose to be consistent, but I can see the use of the variable here too.

pyvista/utilities/common.py
MatthewFlamm: Took your later suggestion.  It is the most direct.  This is also a private helper, so direct is best IMO.

pyvista/utilities/common.py
MatthewFlamm: Three non-conforming cases are probably all committed by me.  Sorry.  It is my personal preference, but I should be more careful with the pyvista style.

pyvista/core/dataset.py
```diff
@@ -2568,3 +2578,62 @@ def cell_point_ids(self, ind: int) -> List[int]:
...
+        """Return whether one or more points are inside a cell.
+
+        .. versionadded:: 0.35.0
+
+        Parameters
+        ----------
+        ind : int
+            Cell ID.
+
+        point : Sequence(float) or np.ndarray
```
MatthewFlamm, line 2594:
```suggestion
        point : Sequence[float] or np.ndarray
```

pyvista/core/dataset.py
MatthewFlamm: Good catch, let's check and raise if not 0 or 1.  It makes sense to me to have a bool (array) return for this.

pyvista/core/dataset.py
MatthewFlamm: Went with square brackets.

pyvista/core/dataset.py
adeak: I'd prefer raising too, my only concern is what happens if someone passes 100 points and one of them has a computational problem? They'd have to remove those points somehow and use the rest.

MatthewFlamm: I suppose we should keep the vtk API but it doesn't seem intuitive.    One thought is to put in a Raises section and to suggest to the user to not supply a set of points if they want to catch an error for single points.

adeak: Hmm, yeah, I think that's a ~good~ least bad compromise. I would also love to keep the bool dtype and the raising. Hopefully computational problems are rare enough that this should only affect few people in very rare cases. We could also mention the recommendation in the error message.

In the off-chance that this starts causing issues downstream we could probably switch from bool to float without breaking existing code.

MatthewFlamm: I suspect this will be a very rare corner case, possibly related to parametric cells (I don't know too much about these, but there are several caveats about them in the docs).  vtk docs also mention that this is rare, and there is a very workable solution for those cases, i.e. loop point by point, which shouldn't be that much slower.

tests/utilities/test_common.py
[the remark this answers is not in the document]
adeak: Hmm. If I understand your point correctly, you could use `np.array_equal(coerced_arg, [point])` for the earlier cases. Would that help?

pyvista/core/dataset.py
MatthewFlamm: I chose this since this is a cell wise operation, which feels like it should be the first parameter to me, but I'm open to other opinions.  I like the idea of an eventual API like `dataset.cells[10].is_point_inside(points)`, which would suggest the current order for now.  But there was also the suggestion of being able to provide a Sequence or similar of cells to this type of function.

pyvista/core/dataset.py
MatthewFlamm: Implemented a loop to catch `-1` or any other unexpected value.

tests/utilities/test_common.py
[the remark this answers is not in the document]
MatthewFlamm: On second consideration, I slightly prefer explicitness here.  The data isn't copied sometimes, so we could modify `point` sometimes.

akaszynski:
Since 0.35 will be released soon, I'd like this to wait for the next minor (0.36) release. Best to not add any last second features.

Good news is we won't wait 3 months to get out the next release. There are a few critical features that we need much earlier, so expect 0.36.0 to be out in a month.

adeak:
Oh, I forgot that this was waiting for me, I'm sorry @MatthewFlamm.

@akaszynski I think this is be good to go, if you can give me ten minutes to make sure.

pyvista/core/dataset.py
```diff
@@ -2568,3 +2578,65 @@ def cell_point_ids(self, ind: int) -> List[int]:
...
+        pcoords = [0.0, 0.0, 0.0]
+        dist2 = _vtk.mutable(0.0)
+        weights = [0.0] * npoints
+
+        in_cell = np.empty(shape=co_point.shape[0], dtype=np.bool_)
+        for i, node in enumerate(co_point):
+            is_inside = cell.EvaluatePosition(node, closest_point, sub_id, pcoords, dist2, weights)
+            if not 0 <= is_inside <= 1:
+                raise RuntimeError(
+                    f"Computational difficulty encountered for point {node} in cell {ind}"
```
adeak, line 2636: Two cosmetic remarks here that we could do:
1. skip coverage for the esoteric error case to make coverage happy
2. also add the point index `i` in the error.

I find it much more valuable to get this feature out in v0.35 than to fix this "should almost never happen" edge case and add a doc-build-length delay. We can follow up for the subsequent release if we want to.

Approving, thanks and sorry for forgetting about this.

pyvista/core/dataset.py
adeak: OK, let's leave this as is.

adeak (approved):

PR #2908, opened 2022-06-29
Fix two minor issues causing noise in docs or doc build

adeak:
Found two minor issues while working on other, build-heavy PRs.

The first one (warning from `read_legacy()` doctest) I would've fixed in https://github.com/pyvista/pyvista/pull/2892 but we probably won't merge that. It makes more sense to fix these in a separate PR.

The second one is a subtle issue where our plot directive collects all existing plots, so if there's a dangling plotter it gets rendered too, even if it doesn't have any `show()` calls. (Not sure if we want to try to catch this; sounds too complex and too error-prone).

From https://dev.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.set_background.html in devdocs:
screenshot: two example code bocks followed by a blank black then the intended normal figure

What I think happens is that the first `plotter` is kept open, so when we define a new plotter with name `pl` in the second test the first one is still around. We can either close the first plotter, or make sure to name the second plotter the same as the first one. I went with the former solution for now.

Needs:
- [x] https://github.com/pyvista/pyvista/pull/2907


akaszynski (dismissed):
~~This works. Thanks!~~

Looks like we angered the doc build.

adeak:
So close 😄 I haven't looked at the artifact yet.

adeak:
Oh wow, this one `plotter.close()` line blows up everything. What the heck? Seems like something we should fix in the plot directive... I'll investigate :sweat_smile:

adeak:
OK, one small piece of the issue is that `plotter.close()` doesn't remove the plotter from `pyvista.plotting._ALL_PLOTTERS`, only `pv.close_all()` does. But I guess we can't just remove closed plotters from `_ALL_PLOTTERS` because we need closed movie/gif plotters to stick around to use their gif files.

Side note: we also have a few gifs that are called `.png`. Probably sphinx gallery's doing:
```py
>>> import pathlib
>>> import matplotlib.pyplot as plt
>>> files = dir.glob('**/*.png')
>>> for file in files:
...     try:
...         img = plt.imread(file)
...     except:
...         print(file)
... 
doc/_build/html/_images/sphx_glr_orbit_003.png
doc/_build/html/_images/sphx_glr_reader_002.png
doc/_build/html/_images/sphx_glr_linked_001.png
doc/_build/html/_images/sphx_glr_gif_001.png
doc/_build/html/_images/sphx_glr_collisions_001.png
doc/_build/html/_images/sphx_glr_texture_007.png
doc/_build/html/_images/sphx_glr_orbit_002.png
doc/_build/html/_images/sphx_glr_orbit_001.png
doc/_build/html/_images/sphx_glr_image-fft-perlin-noise_006.png
doc/_build/html/_images/sphx_glr_isovalue_001.png
doc/_build/html/_images/sphx_glr_moving_cmap_001.png
```

pyvista/plotting/renderers.py
```diff
@@ -417,6 +417,7 @@ def set_background(self, color, top=None, all_renderers=True):
         >>> plotter.set_background('black')
         >>> plotter.background_color
         Color(name='black', hex='#000000ff')
+        >>> plotter.close()
```
adeak, line 420: This will work after https://github.com/pyvista/pyvista/pull/2907.

adeak: Output on top of #2907:

akaszynski (approved):
LGTM. Thanks!

PR #2917, opened 2022-06-30
Handle invalid theme on init

akaszynski:
As noted in #2910, if an invalid theme is set in `PYVISTA_PLOT_THEME`, you will simply be unable to `import pyvista` and potentially get a cryptic error.

This PR fixes this behavior by warning rather than raising.

### DIFF
--- a/pyvista/utilities/misc.py
+++ b/pyvista/utilities/misc.py
@@ -1,10 +1,26 @@
 """Miscellaneous pyvista functions."""
 from collections import namedtuple
+import os
 import warnings
 
 import numpy as np
 
 from pyvista import _vtk
+
+
+def _set_plot_theme_from_env():
+    from pyvista.themes import _ALLOWED_THEMES, set_plot_theme
+
+    if 'PYVISTA_PLOT_THEME' in os.environ:
+        try:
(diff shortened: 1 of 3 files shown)


banesullivan (approved):

pyvista/utilities/misc.py
```diff
@@ -1,12 +1,28 @@
...
 from collections import namedtuple
+import os
 import warnings
 
 import numpy as np
 
 from pyvista import _vtk
 
 
+def _set_plot_theme_from_env():
```
tkoyama010, line 11:
```suggestion
def _set_plot_theme_from_env():
    """Set plot theme from an enviroment variable."""
```

tkoyama010 (changes requested):
Thanks for adding this.

pyvista/utilities/misc.py
```diff
@@ -1,12 +1,29 @@
...
+    """Set plot theme from an enviroment variable."""
```
akaszynski, line 12:
```suggestion
    """Set plot theme from an environment variable."""
```

tkoyama010 (approved):
LGTM

PR #2925, opened 2022-07-01
Improve add mesh performance

akaszynski:
This PR improves the performance of [add_mesh](https://docs.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.add_mesh.html) by about 50% on Linux and 60% Windows by cutting out various unnecessary calls.

#### Why?

Fundamentally, a few milliseconds per `add_mesh` isn't so bad. That is, unless you're calling it a thousand times.

Ideally you would merge all these individual datasets and add a single dataset to the plotter, but that means you can't individually pick each dataset via an actor.

Trust me, I don't want to do this, but as it stands, based on an internal project, reducing the bottleneck is the best I can do.

## Test Setup and Timings

Note that this is not the ideal approach to add identical meshes. Use [glyph()](https://docs.pyvista.org/api/core/_autosummary/pyvista.DataSet.glyph.html) for a more efficient implementation.

```py
import numpy as np
import pyvista as pv

meshes = []
for i in range(1000):
    mesh = pv.Sphere(center=np.random.random(3)*10)
    mesh['data'] = np.random.random(mesh.n_points)
    meshes.append(mesh)


def add_meshes():
    pl = pv.Plotter()
    for mesh in meshes:
        pl.add_mesh(mesh)
    pl.close()

```

### Linux: `main`
```py
>>> timeit add_meshes()
2.91 s ± 28.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

### Linux: This Branch

```py
>>> timeit add_meshes()
1.34 s ± 3.76 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### Windows: `main`
```py
>>> timeit add_meshes()
4.6 s ± 180 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### Windows: This Branch
```py
>>> timeit add_meshes()
1.8 s ± 20.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```


pyvista/plotting/_plotting.py
```diff
@@ -5,17 +5,22 @@
...
+            import matplotlib  # noqa
+
+            _has_matplotlib._imported = True
+        except ImportError:  # pragma: no cover
+            _has_matplotlib._imported = False
 
-        return True
-    except ImportError:  # pragma: no cover
-        return False
+    return _has_matplotlib._imported
```
adeak, line 23: I'll be honest, this "slapping on attributes at runtime" thing with a decorator is pretty iffy. Looking at the current use cases in this PR, can't we use caching instead?
```py
from functools import cache

@cache
def _has_matplotlib():
    try:
        import matplotlib  # noqa
        return False
    return True
```

I haven't tested it but I don't see why this wouldn't work.

pyvista/plotting/colors.py
```diff
@@ -945,36 +975,40 @@ def string_to_rgb(string):  # pragma: no cover
...
-            return cmap
+            try:
+                cmap = getattr(cmocean.cm, cmap)
+            except AttributeError:
+                pass
+            else:
+                return cmap
+
+        try:
+            from matplotlib.cm import get_cmap
```
tkoyama010, line 1006: I think it would be better to use `_has_cmap` here as well.

tkoyama010 (changes requested):

pyvista/plotting/colors.py
adeak: I think if `matplotlib.cm` fails then `matplotlib` fails too, so this could be `_has_matplotlib`. Good catch.

pyvista/plotting/_plotting.py
akaszynski: Implemented this in 730f4a548c74d5aa5778b8d7cdc5a9ae3f1a2a92.

I went for functions rather than a module level constant. A module level constant would require `_has_matplotlib` to be evaluated, which completely defeats our lazy import of `matplotlib`.

pyvista/plotting/colors.py
akaszynski: Implemented in 730f4a548c74d5aa5778b8d7cdc5a9ae3f1a2a92.

adeak:
I'm planning to look at this and some other PRs (e.g. https://github.com/pyvista/pyvista/pull/2980, https://github.com/pyvista/pyvista/pull/2964), but I won't have time for that until tomorrow or maybe even Friday. I see there's a varying degree of rush here, I should probably start with the 0.35.2-milestoned PR (assuming they're still open when I get there).

akaszynski:
Breaking this up into:
- [ ] Cache attempted imports
- [ ] Refactor mapper
- [ ] Fix camera reset

Leaving this up for now.

PR #2933, opened 2022-07-03
Add Gitpod dev image

MatthewFlamm:
### Overview

 

Adds a docker image and init script for setting up a quick dev environment.

### Details

I am fumbling my way through setting up a docker container, so please review critically.

You should be able to spin up a version at https://gitpod.io/#https://github.com/MatthewFlamm/pyvista/tree/gitpod

Todo:
- [ ] Document in developers section


.gitpod.yml
```diff
@@ -0,0 +1,10 @@
+image:
+  file: docker/dev-gitpod.Dockerfile
+
+tasks:
+  - init: |
+      pip install -e .
+      pip install -r requirements_test.txt
+      pip install -r requirements_docs.txt
+      pip install pre-commit
+      pre-commit install --install-hooks
```
MatthewFlamm, line 10: This is slow at first load, but will speed up when user does `git commit`.  I think we should remove `--install-hooks` to speed up time to first terminal prompt, but wanted feedback on what is preferred.

MatthewFlamm:
Hopefully, you should be able to load this up and immediately be able to run tests, e.g. `pytest tests/test_dataset.py`, and make the docs.

I'm not sure if it is related to libosmesa, custom vtk wheel, and python 3.8 here, but one test fails and there are a few errors when building the docs.

MatthewFlamm:
test_antialiasing is failing due to use of libosmesa

.gitpod.yml
banesullivan: Why don't we put all of this in the `Dockerfile`?

banesullivan: `pip install -e .` is the only tricky one. You'll want to copy the `setup.py` ONLY to the Dockerfile then do `pip install -e .`

banesullivan:
Does it make sense to add this to the Docker build/publish action for gitpod to pull? or will gitpod handle building and managing the image?

MatthewFlamm:
I'm going to make this draft.  I need to investigate the suggestions above, plus some other things I want to try.

MatthewFlamm:
SciPy seems to have a fairly mature gitpod setup.  See https://scipy.github.io/devdocs/dev/contributor/quickstart_gitpod.html.  Some things I've noticed for consideration:

- SciPy builds in the `init` section, but it is a much more complex process than here.  I haven't thought about the use here, just an observation.
- SciPy uses liveserve for viewing built documentation.

.gitpod.yml
MatthewFlamm: > `pip install -e .` is the only tricky one. You'll want to copy the `setup.py` ONLY to the Dockerfile then do `pip install -e .`

I have to digest this comment a bit.  Do you mean copy only `setup.py` to the dockerfile and then run `pip install -e .` inside the init for gitpod?

banesullivan: I'd have the Docker file looking something like:

```Dockerfile
...
RUN pip install -r requirements_test.txt
RUN pip install -r requirements_docs.txt
RUN pip install pre-commit
COPY setup.py ./setup.py
RUN pip install -e .
...
```

MatthewFlamm:
While I was able to use this PR to remotely work on and finish #2905, I am not a docker expert. I do not feel like I can finish this PR to be high enough quality to merge.  I will create an feature request issue with a link to this PR and to my fork branch, so someone else with more experience in this area can take over.

banesullivan:
I can try to take over this PR - just need to familiarize myself with gitpod

PR #2934, opened 2022-07-03
Fix all-triangles checking for polyfilters

adeak:
We have a few filters that need all-triangle meshes (and this is even documented), but there was no check for this.

The result for non-triangular meshes varies, e.g.
```py
>>> import pyvista as pv
>>> pv.Cube().decimate(0.5)
2022-07-03 17:48:56.522 (   6.404s) [        1C510740]vtkQuadricDecimation.cx:214    ERR| vtkQuadricDecimation (0x2b11330): Can only decimate triangles
ERROR:root:Can only decimate triangles
PolyData (0x7fde1bfa6dc0)
  N Cells:	0
  N Points:	0
  X Bounds:	1.000e+299, -1.000e+299
  Y Bounds:	1.000e+299, -1.000e+299
  Z Bounds:	1.000e+299, -1.000e+299
  N Arrays:	0
```

```py
>>> pv.Cube().decimate_pro(0.5)
2022-07-03 17:49:09.501 (  19.383s) [        1C510740]     vtkDecimatePro.cxx:191    ERR| vtkDecimatePro (0x2b11330): DecimatePro does not accept polygons that are not triangles.
ERROR:root:DecimatePro does not accept polygons that are not triangles.
PolyData (0x7fddfd94fdc0)
  N Cells:	6
  N Points:	8
  X Bounds:	-5.000e-01, 5.000e-01
  Y Bounds:	-5.000e-01, 5.000e-01
  Z Bounds:	-5.000e-01, 5.000e-01
  N Arrays:	3
```

### DIFF
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -2097,7 +2097,7 @@ class PolyDataFilters(DataSetFilters):
         'Rays intersected at (0.499, 0.000, 0.000), (0.000, 0.497, 0.000), (0.000, 0.000, 0.500)'
 
         """
-        if not self.is_all_triangles:
+        if not self.is_all_triangles:  # pragma: no cover
             raise NotAllTrianglesError("Input mesh for multi_ray_trace must be all triangles.")
 
         try:
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -968,6 +968,9 @@ class PolyDataFilters(DataSetFilters):
         See :ref:`decimate_example` for more examples using this filter.
 
         """
+        if not self.is_all_triangles:
+            raise NotAllTrianglesError("Input mesh for decimation must be all triangles.")
+
         alg = _vtk.vtkDecimatePro()
         alg.SetInputData(self)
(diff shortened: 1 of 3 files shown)


banesullivan (approved):

pyvista/core/filters/poly_data.py
```diff
@@ -2086,7 +2098,7 @@ def multi_ray_trace(self, origins, directions, first_point=False, retry=False):
 
         """
         if not self.is_all_triangles:
```
akaszynski, line 2100:
```suggestion
        if not self.is_all_triangles:  # pragma: no cover
```

adeak: I don't think this is the right call. I added a test for `multi_ray_trace` for this branch. The issue is that coverage probably doesn't have the optional dependencies needed for that test, so it skips it. But this relates to the entire `multi_ray_trace` functionality, not just the diff. If we want to skip this for coverage, we should skip the whole method @akaszynski.

akaszynski: Fixed in 051122a780ccd14fc4f547839ae28f71c360338b.

akaszynski (approved):

PR #2941, opened 2022-07-05
Add composite dataset mapper

### LINKED ISSUE #3095: Support vtkCompositePolyDataMapper
Add a composite dataset mapper to `pyvista.Plotter` to improve the performance of `MultiBlock` datasets.

akaszynski:
Add a composite dataset mapper to `pyvista.Plotter` and resolve #3095.

This vastly improves the performance when adding composite (`pyvista.MultiBlock`) datasets to `pyvista.Plotter`, and is absolutely critical for multi-component datasets exceeding 1000 blocks. Given that VTK does not handle a large number of actors well, component datasets are a way to work around that limitation and still be able to individually pick parts of the plotter.

For example:

https://user-images.githubusercontent.com/11981631/177257454-7260fb59-7aa4-424f-8e7f-01f9bbfa7b62.mp4

```py
import random
import pyvista as pv
import numpy as np


def make_poly():
    poly = pv.ParametricSuperEllipsoid(
        n1=random.random(),
        n2=random.random() * 2,
        u_res=50,
        v_res=50,
    )
    poly.points += np.random.random(3) * 20
    return poly


blocks = pv.MultiBlock([make_poly() for _ in range(100)])

pl = pv.Plotter()
actor, mapper = pl.add_composite(blocks, color="w", pbr=True, metallic=True)


def callback(index, *args):
    """Change a block to red if color is unset, and back to the actor color if set."""
    if mapper.block_attr[index].color is None:
        mapper.block_attr[index].color = "r"
    else:
        mapper.block_attr[index].color = None


pl.enable_block_picking(callback, side="left")
pl.background_color = "w"
pl.show()
```


pyvista/core/composite.py
```diff
@@ -690,3 +703,238 @@ def copy(self, deep=True):
...
+        Notes
+        -----
+        The number of components of the data must match.
+
+        """
+        data_assoc: List[Tuple[FieldAssociation, np.ndarray, MultiBlock]] = []
+        for block in self:
+            if block is not None:
+                if isinstance(block, MultiBlock):
+                    field, scalars = block.set_active_scalars(name, preference, allow_missing=True)
```
MatthewFlamm, line 747: We should pass on the value of `allow_missing` here

pyvista/core/composite.py
```diff
@@ -690,3 +703,238 @@ def copy(self, deep=True):
...
+                else:
+                    try:
+                        field, scalars = block.set_active_scalars(name, preference)
+                    except KeyError:
+                        if not allow_missing:
+                            raise
```
MatthewFlamm, line 753: Same comment about reraising

pyvista/core/composite.py
```diff
@@ -690,3 +703,238 @@ def copy(self, deep=True):
...
+                    data_assoc.append((field, scalars, block))
+
+        if name is None:
+            return FieldAssociation.NONE, np.array([])
+
+        if not data_assoc:
+            raise KeyError(f'"{name}" is missing from all the blocks of this composite dataset.')
+
+        # Verify that all the data has identical field association
+        first_field = data_assoc[0][0]
```
MatthewFlamm, line 767: Why is this section of testing necessary for setting active scalars?  It seems we'd need to do this for a particular use of the active scalars, but it isn't required to set them.

pyvista/core/composite.py
```diff
@@ -690,3 +703,238 @@ def copy(self, deep=True):
...
+                    dataset[i] = block.as_polydata()
+                elif not isinstance(block, pyvista.PolyData):
+                    dataset[i] = block.extract_surface()
+                else:
+                    # guarantee that the new instance is a shallow copy
+                    dataset[i] = block.copy(deep=False)
+            else:
+                # must have empty polydata within these datasets to ensure
+                # downstream filters don't work on null pointers
+                dataset[i] = pyvista.PolyData()
```
MatthewFlamm, line 829: It may be worth pointing out that null blocks will be non-null in the docstring

pyvista/core/dataset.py
```diff
@@ -641,31 +641,47 @@ def set_active_scalars(self, name: Optional[str], preference='cell'):
...
-                f'Data field ({name}) with type ({field}) could not be set as the active scalars'
+                f'Data field "{name}" with type ({field}) could not be set as the active scalars'
             )
 
         self._active_scalars_info = ActiveArrayInfo(field, name)
 
+        if field == FieldAssociation.POINT:
+            return field, self.point_data.active_scalars
+        else:  # must be cell
+            return field, self.cell_data.active_scalars
```
MatthewFlamm, line 683: Is this breaking change worth it outside of this PR?

pyvista/core/filters/composite.py
```diff
@@ -154,3 +158,44 @@ def outline_corners(self, factor=0.2, nested=False, progress_bar=False):
...
+        non_manifold_traversal=True,
+        feature_angle=30.0,
+        track_vertices=False,
+        progress_bar=False,
+    ):
+        """Compute point and/or cell normals for a multi-block dataset.
+
+        Warnings
+        --------
+        This will segmentation fault if a single block is non-polydata.
```
MatthewFlamm, line 179: Can we use `is_all_polydata` here to avoid a segfault?

MatthewFlamm (commented):
I haven't gone through the whole thing yet.

pyvista/core/composite.py
```diff
@@ -249,7 +252,13 @@ def get_data_range(self, name: str) -> Tuple[float, float]:  # type: ignore
...
                 continue
             # get the scalars if available - recursive
-            tmi, tma = data.get_data_range(name)
+            try:
+                tmi, tma = data.get_data_range(name)
+            except KeyError:
+                if allow_missing:
+                    continue
+                else:
+                    raise
```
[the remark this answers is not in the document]
akaszynski: These two are equivalent:
```py
try:
    1/0
except ZeroDivisionError:
    raise
```

```
      2 
      3 try:
----> 4     1/0
      5 except ZeroDivisionError:
      6     raise

ZeroDivisionError: division by zero
```

```py
try:
    1/0
except ZeroDivisionError as err:
    raise err
```

```
      3 try:
----> 4     1/0
      5 except ZeroDivisionError as err:
      6     allow_invalid

ZeroDivisionError: division by zero
```

However, I suppose it's a bit cleaner to go with a more explicit `ZeroDivisionError as err`

pyvista/core/composite.py
akaszynski: Done in 416ef55af169ce2da8d4443e3c1d17eeff9d7364.

pyvista/core/filters/composite.py
akaszynski: Implemented in 416ef55af169ce2da8d4443e3c1d17eeff9d7364.

pyvista/core/composite.py
akaszynski: Agreed. I've refactored most of this out in 416ef55af169ce2da8d4443e3c1d17eeff9d7364, but it's still important to know the field association made active when setting the scalars in order to set it in the mapper, so I've kept in a tiny bit of the logic.

pyvista/core/composite.py
akaszynski: Fixed in 416ef55af169ce2da8d4443e3c1d17eeff9d7364.

pyvista/core/composite.py
akaszynski: Good catch! Fixed in 416ef55af169ce2da8d4443e3c1d17eeff9d7364.

pyvista/core/composite.py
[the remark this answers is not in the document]
akaszynski: Fixed in 416ef55af169ce2da8d4443e3c1d17eeff9d7364.

akaszynski:
It's "more ready" for review.

banesullivan (approved):
This is a hefty PR. I do not have the time/bandwidth to give this a full/thorough review, but I have managed to look over much of the code additions here. I am liking the way this is going -- especially with the new `vtkProperty` wrapping!

There are some significant changes to the internals of the plotting mechanisms for `add_mesh` which should always be taken with great consideration as it has historically been a fragile mess (🙋🏻 I hold much of that blame), though the tests seem happy with it and we just had a release, so we have another release cycle to iron out any quirks.

Further, this is *so much better* than what we currently have, and the overall changes here are pushing the library in the right direction. As we continue to refactor and re-architect the plotting mechanisms in PRs like #2964 and subsequent, we can refine this further as needed.

With that all of that and considering the `MultiBlock`  datasets and composite plotting are lesser used features in PyVista, this LGTM! @akaszynski, I'd say merge this when you feel it's ready, though we ought to have someone with a bit more bandwidth review the more impactful changes to at least get a second pair of eyes on this.

akaszynski:
> With that all of that and considering the MultiBlock datasets and composite plotting are lesser used features in PyVista, this LGTM! @akaszynski, I'd say merge this when you feel it's ready, though we ought to have someone with a bit more bandwidth review the more impactful changes to at least get a second pair of eyes on this.

Appreciate the review!

Going to keep improving this patch coverage and then wait 24H if anyone else has anything to say. I'll run through a final verification of the docs. Side note: I would really like an automated sphinx-gallery check, even if it's just local.

banesullivan:
> Side note: I would really like an automated sphinx-gallery check, even if it's just local.

This one's been on my to do list for far too long

pyvista/plotting/plotting.py
```diff
@@ -1807,6 +1804,443 @@ def update(self, stime=1, force_redraw=True):
...
+        roughness=0.5,
+        render=True,
+        component=None,
+        color_missing_with_nan=False,
+        **kwargs,
+    ):
+        """Add a composite dataset to the plotter.
+
+        Parameters
+        ----------
```
MatthewFlamm, line 1860: I called out a few examples, but many parameters are missing defaults.  I assume this also means the same for `add_*`

MatthewFlamm (commented):
Another round of comments.  I can't in detail review the logic of the mapper and properties, but the cleanup of `plotting` is excellent and sorely needed.

pyvista/plotting/_property.py
```diff
@@ -0,0 +1,318 @@
...
+@lru_cache(maxsize=None)
+def _check_supports_pbr():
+    """Check if VTK supports physically based rendering."""
+    if not _vtk.VTK9:  # pragma: no cover
+        from pyvista.core.errors import VTKVersionError
+
+        raise VTKVersionError('Physically based rendering requires VTK 9 or newer.')
+
+
+class Property(_vtk.vtkProperty):
```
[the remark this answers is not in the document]
akaszynski: I've made this class public and added it to the documentation.

pyvista/plotting/_property.py
```diff
@@ -0,0 +1,318 @@
...
+                raise TypeError(
+                    'Expected ``pyvista.themes.DefaultTheme`` for '
+                    f'``theme``, not {type(theme).__name__}.'
+                )
+            self._theme.load_theme(theme)
+
+        self.interpolation = interpolation
+        self.color = color
+        self.style = style
+        if interpolation == 'Physically based rendering':
```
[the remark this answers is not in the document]
akaszynski: Allowing both.

pyvista/plotting/plotting.py
```diff
@@ -1807,6 +1804,443 @@ def update(self, stime=1, force_redraw=True):
...
+        dataset : pyvista.MultiBlock
+            A :class:`pyvista.MultiBlock` dataset.
+
+        color : color_like, optional, defaults to white
```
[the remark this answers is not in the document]
akaszynski: Technically the default is set by the theme. I've updated the docs.

pyvista/plotting/plotting.py
```diff
@@ -1807,6 +1804,443 @@ def update(self, stime=1, force_redraw=True):
...
+        clim : 2 item list, optional
+            Color bar range for scalars.  Defaults to minimum and
+            maximum of scalars array.  Example: ``[-1, 2]``. ``rng``
+            is also an accepted alias for this.
+
+        show_edges : bool, optional
+            Shows the edges of a mesh.  Does not apply to a wireframe
+            representation.
+
+        edge_color : color_like, optional, defaults to black
```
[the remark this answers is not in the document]
akaszynski: Added note regarding theme.

pyvista/plotting/plotting.py
akaszynski: I'm going to leave the rest of the `add_*` alone, but I'll fix this docstring.

pyvista/plotting/plotting.py
```diff
@@ -1807,6 +1804,443 @@ def update(self, stime=1, force_redraw=True):
...
+        show_scalar_bar : bool
+            If ``False``, a scalar bar will not be added to the
+            scene. Defaults to ``True`` unless ``rgba=True``.
+
+        multi_colors : bool, optional
+            Color each block by a solid color using matplotlib's color cycler.
+
+        name : str, optional
+            The name for the added mesh/actor so that it can be easily
+            updated.  If an actor of this name already exists in the
```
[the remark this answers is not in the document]
akaszynski: Default is the memory address, which I'm guessing that they're not going to care about (or check) if they don't set it. Leaving this as "optional".

pyvista/plotting/_property.py
```diff
@@ -0,0 +1,318 @@
...
+
+            self.SetInterpolationToPBR()
+        elif new_interpolation == 'Phong':
+            self.SetInterpolationToPhong()
+        elif new_interpolation == 'Gouraud':
+            self.SetInterpolationToGouraud()
+        elif new_interpolation == 'Flat' or new_interpolation is None:
+            self.SetInterpolationToFlat()
+        else:
+            raise ValueError(f'Invalid interpolation "{new_interpolation}"')
```
adeak: Aside from the `'pbr'` vs `'Physically based rendering'` duality, one might also want to use `'phong'` instead of `'Phong'` etc., or try `'Physically-based rendering'`. This can be a moderate pain in the butt for users. I'd prefer to potentially normalize to lowercase (and converting dashes to spaces). This would actually be the perfect use case for an `Enum`, but I also don't want to overengineer this already 3.7k LOC PR.

akaszynski: > 3.7k LOC PR.

Typical Alex PR.

tests/test_scalar_bars.py
```diff
@@ -85,3 +85,15 @@ def test_add_remove_bar(sphere):
...
     assert len(pl._scalar_bar_slots) == init_slots
+
+
+def test_too_many_scalar_bars(sphere):
+    pl = pv.Plotter()
+
+    with pytest.raises(RuntimeError, match='Maximum number of color'):
+        for i in range(100):
+            mesh = pv.Sphere()
+            mesh[f'{i}'] = range(mesh.n_points)
```
adeak, line 96:
```suggestion
            mesh[str(i)] = range(mesh.n_points)
```

tests/test_scalar_bars.py
```diff
@@ -85,3 +85,15 @@ def test_add_remove_bar(sphere):
...
+            pl.add_mesh(mesh)
+
+    pl.show()
```
adeak, line 99: Do we need to show?

tests/test_picking.py
```diff
@@ -377,3 +377,8 @@ def callback(click_point):
     # ensure callback was called and camera position changes due to "fly"
     assert cpos_before != pl.camera_position
     assert point
+
+
+def test_block_picking(multiblock_poly):
+    pl = pyvista.Plotter()
+    pl.add_composite(multiblock_poly)
```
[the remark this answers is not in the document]
akaszynski: It's just not complete. Added in 9282488c88b6a4f542b330bd226b7bb77e6799d7.

tests/plotting/conftest.py
```diff
@@ -20,10 +20,17 @@ def _is_vtk(obj):
 
 
 @pytest.fixture(autouse=True)
-def check_gc():
+def check_gc(request):
     """Ensure that all VTK objects are garbage-collected by Python."""
     before = set(id(o) for o in gc.get_objects() if _is_vtk(o))
     yield
+
+    # Do not check for collection if the test failed. Tests that fail also fail
```
adeak, line 28: When you say "when the test failed", do you mean "when the test suite failed", i.e. "there was at least one failing test"? Because that's what this seems to do (and I agree with that). Doing it on a per-test basis would be a bit messier, according to https://doc.pytest.org/en/latest/example/simple.html#making-test-result-information-available-in-fixtures and https://github.com/pytest-dev/pytest/issues/230.

pyvista/plotting/_property.py
akaszynski: Done in [498d644](https://github.com/pyvista/pyvista/pull/2941/commits/498d6443c07c7ba843bef026c960c7847028a31a).

pyvista/plotting/_property.py
```diff
@@ -0,0 +1,1144 @@
...
+        >>> prop.interpolation = 'Physically based rendering'
+        >>> prop.plot()
+
+        """
+        return self.GetInterpolationAsString()
+
+    @interpolation.setter
+    def interpolation(self, value: str):
+        if not isinstance(value, str):
+            raise TypeError(f'`interpolation` must be a string, not {type(value)}')
```
adeak, line 752: This seems to clash with "`None` means flat" now.

tests/test_scalar_bars.py
adeak: I meant to submit this with some other comments, but I can only reflect on a recent change by posting this along with that comment, so here goes. We can wait to commit these kinds of changes together later. I'm progressing at snail's pace.

tests/test_scalar_bars.py
akaszynski: Nope. Removed in f76f8f05fd87a400f6b5f03df9f02a7af3df2df6.

tests/plotting/conftest.py
akaszynski: You're correct, and I've updated the comment in f76f8f05fd87a400f6b5f03df9f02a7af3df2df6.

pyvista/plotting/_property.py
akaszynski: Completely removed the mention of `None` in f76f8f05fd87a400f6b5f03df9f02a7af3df2df6.

akaszynski:
> I'd say merge this when you feel it's ready, though we ought to have someone with a bit more bandwidth review the more impactful changes to at least get a second pair of eyes on this.

Looking to merge this fairly soon. I'd like to get major features out early in the dev cycle (apparently we have a dev cycle) and then push patches based on usage.

Coverage is as good as I'm going to get it without some weird tests. I'll leave this up for another 24H and then merge. So comment now or forever hold your peace (or just open another follow up PR :) ).

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,769 @@
...
+
+        Returns
+        -------
+        pyvista.DataObject
+            PyVista dataobject.
+
+        Examples
+        --------
+        Note how the zero index is the entire multiblock. To access individual
+        multiblocks, add one to the index.
```
adeak, line 363: I found this to be super confusing, and I'm sure everyone who comes across this via things like `multi.block_attr[1].color = 'blue'` will be confused too. Is there a VTK convention we are following here?

pyvista/core/composite.py
```diff
@@ -690,3 +703,244 @@ def copy(self, deep=True):
...
+                        field, scalars = FieldAssociation.NONE, np.array([])
+
+                if field != FieldAssociation.NONE:
+                    data_assoc.append((field, scalars, block))
+
+        if name is None:
+            return FieldAssociation.NONE, np.array([])
+
+        if not data_assoc:
+            raise KeyError(f'"{name}" is missing from all the blocks of this composite dataset.')
```
adeak, line 766: This is slightly hacky because normally a `KeyError`'s argument is the offending key, but I think we can get away with this.

pyvista/core/composite.py
```diff
@@ -690,3 +703,244 @@ def copy(self, deep=True):
...
+                        if not allow_missing:
+                            raise err
+                        block.set_active_scalars(None, preference)
+                        field, scalars = FieldAssociation.NONE, np.array([])
+
+                if field != FieldAssociation.NONE:
+                    data_assoc.append((field, scalars, block))
+
+        if name is None:
+            return FieldAssociation.NONE, np.array([])
```
adeak, line 763: If my hunch is correct that scalars are normally `pyvista_ndarray`s then we should consider returning these for type consistency:
```suggestion
                        field, scalars = FieldAssociation.NONE, pyvista.pyvista_ndarray([])

                if field != FieldAssociation.NONE:
                    data_assoc.append((field, scalars, block))

        if name is None:
            return FieldAssociation.NONE, pyvista.pyvista_ndarray([])
```

Might need a `from` import to keep the line under 100 characters, haven't checked.

pyvista/core/composite.py
```diff
@@ -690,3 +703,244 @@ def copy(self, deep=True):
...
+
+        if not data_assoc:
+            raise KeyError(f'"{name}" is missing from all the blocks of this composite dataset.')
+
+        field_asc = data_assoc[0][0]
+        # set the field association to the preference if at least one occurrence
+        # of it exists
+        if field_asc.name.lower() != preference:
```
adeak, line 771: Shouldn't we also normalize the casing of `preference`?

pyvista/core/composite.py
```diff
@@ -690,3 +703,244 @@ def copy(self, deep=True):
...
+            raise ValueError(f'Inconsistent dimensions {dims} in active scalars.')
+
+        # check complex mismatch
+        is_complex = [np.issubdtype(dtype, np.complexfloating) for dtype in dtypes]
+        if any(is_complex) and not all(is_complex):
+            raise ValueError('Inconsistent complex and real data types in active scalars.')
+
+        return field_asc, scalars
+
+    def as_polydata(self):
```
adeak, line 798: The name is slightly confusing, but I guess the docstring is clear enough, and I don't have a better name in mind. Perhaps something `as_polydata_blocks()` or better would make it clear that the encapsulating type won't change, only the contents. Mild suggestion.

pyvista/core/dataset.py
```diff
@@ -641,16 +641,27 @@ def set_active_scalars(self, name: Optional[str], preference='cell'):
...
             self.GetPointData().SetActiveScalars(None)
-            return
+            return FieldAssociation.NONE, np.array([])
         field = get_array_association(self, name, preference=preference)
         if field == FieldAssociation.NONE:
-            raise KeyError(f'Data named ({name}) is a field array which cannot be active.')
+            if name in self.field_data:
+                raise KeyError(f'Data named "{name}" is a field array which cannot be active.')
+            else:
+                raise KeyError(f'Data named "{name}" does not exist in this dataset.')
```
adeak, line 664: Also not sure about these being `KeyError`s. We also have a `ValueError` about "could not be set as active scalars" a few lines below. So it's not even very consistent.

doc/api/plotting/index.rst
```diff
@@ -27,6 +27,23 @@ all plotting functionality in PyVista.
 
    Plotter
    Renderer
+   Property
+
+
+Composite Plotting Reference
+----------------------------
+Composite dataset plotting reference. These classes are used when plotting
+composite datasets like :class:`pyvista.MultiBlock`.
```
adeak, line 36: 1. The first sentence seems redundant after the same heading.
2. What are "composite datasets _like_ [`MultiBlock`]" that aren't `MultiBlock`?

adeak (commented):
Very cursory review. Haven't looked at anything in `_property.py` nor `pyvista/plotting/plotting.py` (yet, but also for the foreseeable future).

I'll also fire up a local doc build just to make sure things build and look OK, and push a few minor changes of my own.

akaszynski:
> I'll also fire up a local doc build just to make sure things build and look OK, and push a few minor changes of my own.

Sounds good. I'll wait til that's done.

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,769 @@
...
+    Add a sphere and a cube as a multiblock dataset to a plotter and then
+    change the visibility and color of the blocks.
+
+    >>> import pyvista as pv
+    >>> dataset = pv.MultiBlock([pv.Cube(), pv.Sphere(center=(0, 0, 1))])
+    >>> pl = pv.Plotter()
+    >>> actor, mapper = pl.add_composite(dataset)
+    >>> mapper.block_attr[1].color = 'b'
+    >>> mapper.block_attr[1].opacity = 0.1
+    >>> mapper.block_attr[1]
```
adeak, line 39: If this 1-based API is here to stay we should definitely explain this too here in the class docs of `BlockAttributes`.

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,769 @@
...
+        return len(self._dataset)
+
+    def __iter__(self):
+        """Return an iterator of all the block attributes."""
+        for ii in range(len(self)):
+            yield self[ii]
+
+
+class CompositePolyDataMapper(_vtk.vtkCompositePolyDataMapper2):
+    """Composite PolyData mapper.
```
adeak, line 416: Currently this class shows up in the docs, but most of the methods only have a single-sentence docstring, no parameters/returns/examples. If this is final then we should open an issue to get these fixed (potentially as good first issues).

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,769 @@
...
+
+    @property
+    def block_attr(self) -> CompositeAttributes:
+        """Return the block attributes.
+
+        Examples
+        --------
+        Add a sphere and a cube as a multiblock dataset to a plotter and then
+        change the visibility and color of the blocks.
+
```
adeak, line 468: One-basedness mentioned here too?

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,769 @@
...
+class BlockAttributes:
+    """Block attributes used to set the attributes of a block.
+
+    Parameters
+    ----------
+    block : pyvista.DataObject
+        PyVista dataobject.
+
+    attr : pyvista.CompositeAttributes
```
adeak, line 25: For some reason in my local build intersphinx is picking up on `pyvista.DataObject` but not `pyvista.CompositeAttributes`. Any idea why?

adeak (commented):
Round of comments after looking at the local doc build. I can't repro the error in CI, [it's probably cache-related](https://shouldiblamecaching.com/).

For a future PR: some of these classes don't have any methods or don't have any attributes, but there's an empty corresponding section in the docs.

`CompositeAttributes`: no attributes (ironic)
Screenshot from 2022-07-27 00-10-47

`BlockAttributes`: no methods
Screenshot from 2022-07-27 00-11-37

If there's a way to remove these empty sections it would probably look neater.

pyvista/plotting/__init__.py
```diff
@@ -14,7 +14,9 @@
 )
 from .export_vtkjs import export_plotter_vtkjs, get_vtkjs_url
 from .helpers import plot, plot_arrows, plot_compare_four, plot_itk
+from .composite_mapper import CompositeAttributes, BlockAttributes, CompositePolyDataMapper
 from .plotting import BasePlotter, Plotter, close_all
+from ._property import Property
```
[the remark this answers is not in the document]
akaszynski: I think property should be public. I'm envisioning one day:
```py
prop = pyvista.Property(<lots-of-kwargs>)
pl = pyvista.Plotter()
actor = pl.add_mesh(mesh)
actor.property = prop
```

tests/plotting/test_plotting.py
```diff
@@ -2452,13 +2444,194 @@ def test_warn_screenshot_notebook():
...
+
+    pl.camera_position = 'xy'
+    pl.show(before_close_callback=verify_cache_image)
+
+
+def test_plot_composite_poly_no_scalars(multiblock_poly):
+    pl = pyvista.Plotter()
+
+    # Note: set the camera position before making the blocks invisible. VTK
+    # 9.0.3 still considers invisible blocks when determining camera bounds.
```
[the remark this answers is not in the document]
akaszynski: Comment clarified.

tests/plotting/test_plotting.py
```diff
@@ -2452,13 +2444,194 @@ def test_warn_screenshot_notebook():
...
+
+
+def test_plot_composite_poly_component_norm(multiblock_poly):
+    for block in multiblock_poly:
+        data = block.compute_normals().point_data['Normals']
+        block['data'] = data
+
+    pl = pyvista.Plotter()
+    pl.add_composite(multiblock_poly, scalars='data', clim=[0.9, 1.1], cmap='bwr')
+    pl.show(before_close_callback=verify_cache_image)
```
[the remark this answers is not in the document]
akaszynski: Good point. I've fixed that.

pyvista/plotting/__init__.py
```diff
@@ -14,7 +14,9 @@
 )
 from .export_vtkjs import export_plotter_vtkjs, get_vtkjs_url
 from .helpers import plot, plot_arrows, plot_compare_four, plot_itk
+from .composite_mapper import CompositeAttributes, BlockAttributes, CompositePolyDataMapper
```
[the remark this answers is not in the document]
akaszynski: I'm fine with dropping these, but the name gets quite long in our docs. Is there a workaround for this?

sc0

pyvista/_vtk.py
```diff
@@ -413,6 +414,8 @@
         vtkCamera,
         vtkCellPicker,
         vtkColorTransferFunction,
+        vtkCompositeDataDisplayAttributes,
+        vtkCompositePolyDataMapper,
```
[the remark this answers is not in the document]
akaszynski: This actually needs to be removed altogether. I was comparing the performance and features of `vtkCompositePolyDataMapper` and `vtkCompositePolyDataMapper2` and forgot to remove the former.

pyvista/core/composite.py
akaszynski: Done in aee6ffa58ee0f26d183b464f761ef056f110cae2

pyvista/core/composite.py
akaszynski: Implemented in aee6ffa58ee0f26d183b464f761ef056f110cae2

pyvista/core/dataset.py
akaszynski: Converted the first to `ValueError` in aee6ffa58ee0f26d183b464f761ef056f110cae2. It's not a key error because the key exists, it just has the wrong field association.

doc/api/plotting/index.rst
akaszynski: We will eventually deprecate Multi-block datasets to composite datasets given the recent changes in VTK's API, but I think we're getting ahead of ourselves since it doesn't exist. Updated in aee6ffa58ee0f26d183b464f761ef056f110cae2

pyvista/core/composite.py
akaszynski: Agreed. `as_polydata` sounds like we're just casting or converting to `pyvista.PolyData`. Fixed this in aee6ffa58ee0f26d183b464f761ef056f110cae2.

pyvista/plotting/composite_mapper.py
akaszynski: Yes, I found it confusing at first and I've since clarified the docstring in aee6ffa58ee0f26d183b464f761ef056f110cae2.

See https://kitware.github.io/vtk-examples/site/Java/CompositeData/CompositePolyDataMapper/

pyvista/plotting/composite_mapper.py
akaszynski: Done in aee6ffa58ee0f26d183b464f761ef056f110cae2.

pyvista/plotting/composite_mapper.py
akaszynski: Mentioned in aee6ffa58ee0f26d183b464f761ef056f110cae2.

pyvista/plotting/composite_mapper.py
akaszynski: Done.

akaszynski:
> For a future PR: some of these classes don't have any methods or don't have any attributes, but there's an empty corresponding section in the docs.

I agree that it shouldn't have these sections. After looking at https://github.com/pyvista/pyvista/blob/main/doc/_templates/custom-class-template.rst, it looks like it simply shouldn't be generating these blocks given `{% if methods %}`, which according to https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html is a list of public methods available. Since there are no public methods/attributes in a few of these classes, I'm surprised these even show up at all.

pyvista/core/composite.py
```diff
@@ -690,3 +704,253 @@ def copy(self, deep=True):
...
+        # we make a shallow copy here to avoid modifying the original dataset
+        dataset = self
+        if copy:
+            dataset = self.copy(deep=False)
+
+        # always convert to polydata
+        for i, block in enumerate(dataset):
+            if block is not None:
+                if isinstance(block, MultiBlock):
+                    dataset[i] = block.as_polydata_blocks()
```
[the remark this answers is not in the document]
akaszynski: You're right, I didn't consider that we might have named blocks. This is where a default copy would have been ideal, but I'll add in the name here.

A replace method would have been useful here.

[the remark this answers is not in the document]
MatthewFlamm: #3031 now adds `replace` method.  If this one is merged first, I will try to remember to add it here.

pyvista/core/composite.py
```diff
@@ -690,3 +704,253 @@ def copy(self, deep=True):
...
+                elif not isinstance(block, pyvista.PolyData):
+                    dataset[i] = block.extract_surface()
+                elif copy:
+                    # create a shallow copy
+                    dataset[i] = block.copy(deep=False)
```
adeak, line 838: Not sure I understand all the logic here.

It seemed to me that `copy=True` on input meant "give me a shallow copy of this MultiBlock". But here we do `block.copy(deep=False)` which copies the PolyData inside (even if this is a shallow copy). This means that `copied_multi[i] is original_multi[i]` doesn't hold, which I'd expect from a shallow copy of `original_multi`. On the other hand the two non-identical meshes are shallow copies, so certain changes in one will show up in the latter...

Not sure how to build a mental model around this, and how `copy` behaves. Plus the docstring says

> Blocks that are already PolyData will not be copied.

But then why do we have a `block.copy()` in here in the first place?

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,1012 @@
...
+        return self.SetLookupTable(table)
+
+    @property
+    def scalar_range(self):
+        """Return or set the scalar range.
+
+        Examples
+        --------
+        Return the scalar range. In this example its set to its default value
+        of ``(0.0, 1.0)``
```
adeak, line 644:
```suggestion
        Return the scalar range. In this example it's set to its default value
        of ``(0.0, 1.0)``.
```

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,1012 @@
...
+
+        >>> import pyvista as pv
+        >>> dataset = pv.MultiBlock([pv.Cube(), pv.Sphere(center=(0, 0, 1))])
+        >>> dataset[0].point_data['data'] = dataset[0].points[:, 2]
+        >>> dataset[1].point_data['data'] = dataset[1].points[:, 2]
+        >>> pl = pv.Plotter()
+        >>> actor, mapper = pl.add_composite(dataset, scalars='data')
+        >>> mapper.scalar_map_mode
+        'point'
+
```
adeak, line 734:
```suggestion
```

tests/plotting/test_plotting.py
```diff
@@ -1254,117 +1272,61 @@ def test_plot_rgb():
...
-    p.add_mesh(data, scalars='vector_values_points', component=0)
+    p.add_mesh(multicomp_poly, scalars='vector_values_points', component=0)
     p.show()
 
-    p = pyvista.Plotter()
-    p.add_mesh(data, scalars='vector_values_cells', component=0)
-    p.show()
+    # p = pyvista.Plotter()
+    # p.add_mesh(multicomp_poly, scalars='vector_values_cells', component=0)
+    # p.show()
```
adeak, line 1293: Did you mean to remove or uncomment this test?

tests/plotting/test_plotting.py
```diff
@@ -1254,117 +1272,61 @@ def test_plot_rgb():
...
-    # does not modify it.
-    p = pyvista.Plotter()
-    p.add_mesh(data, scalars='vector_values_points', component=0)
-    p.show()
-    assert np.array_equal(data['vector_values_points'], copy_vector_values_points)
-
-    p = pyvista.Plotter()
-    p.add_mesh(data, scalars='vector_values_cells', component=0)
-    p.show()
-    assert np.array_equal(data['vector_values_cells'], copy_vector_values_cells)
```
adeak, line 1344: Was this moved or removed altogether for some reason?

adeak:
Build failure seems to be due to this flaky silliness again:
```
/home/runner/work/pyvista/pyvista/doc/api/plotting/_autosummary/pyvista.Property.rst: WARNING: document isn't included in any toctree
/home/runner/work/pyvista/pyvista/doc/api/plotting/_autosummary/pyvista.plotting.composite_mapper.BlockAttributes.rst: WARNING: document isn't included in any toctree
/home/runner/work/pyvista/pyvista/doc/api/plotting/_autosummary/pyvista.plotting.composite_mapper.CompositeAttributes.rst: WARNING: document isn't included in any toctree
/home/runner/work/pyvista/pyvista/doc/api/plotting/_autosummary/pyvista.plotting.composite_mapper.CompositePolyDataMapper.rst: WARNING: document isn't included in any toctree
```

pyvista/core/composite.py
akaszynski: Looking at this with fresh eyes, the behavior of copy should be to control if the `PolyData` datasets are copied or not. When converting to polydata, we should always be returning a copy of the multiblock itself so we don't overwrite any of the original datasets, and instead we should focus on if we're copying `PolyData`, because that's the only dataset type that we aren't already forced to copy.

akaszynski: Fixed in 75c684fa7b8a89eba93dab55ad9c7a9d61715bd0.

examples/02-plot/composite-picking.py
```diff
@@ -0,0 +1,61 @@
...
+# Create a MultiBlock Dataset
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Create 100 super ellipsoid using :func:`pyvista.ParametricSuperEllipsoid`
+
+
+def make_poly():
+    """Create a super ellipsoid in a random location."""
+    poly = pv.ParametricSuperEllipsoid(
+        n1=random.random(),
+        n2=random.random() * 2,
```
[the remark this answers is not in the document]
akaszynski: Fixed in 75c684fa7b8a89eba93dab55ad9c7a9d61715bd0.

pyvista/core/composite.py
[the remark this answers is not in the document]
akaszynski: Fixed both issues in 75c684fa7b8a89eba93dab55ad9c7a9d61715bd0.

pyvista/core/composite.py
akaszynski: Fixed in 75c684fa7b8a89eba93dab55ad9c7a9d61715bd0.

tests/plotting/test_plotting.py
akaszynski: Thanks for catching that. Added back in c619b9f7f173e1bb1bba935599258a17c2ab7557.

tests/plotting/test_plotting.py
akaszynski: Yes, while I was working on the refactor of `setup_multicomponent_data` and forgot about adding the test back in. Fixed in 1e2fe42ac0127f43dc7186daa818c9d7b8c76958.

pyvista/plotting/__init__.py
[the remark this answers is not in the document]
akaszynski: Resolved in 3f1e3ee4e04af0ecad20225723ff34979f92d002.

tests/plotting/test_plotting.py
```diff
@@ -1254,117 +1274,93 @@ def test_plot_rgb():
...
+    pl = pyvista.Plotter()
+    pl.add_mesh(multicomp_poly, scalars='vector_values_points', component=0)
+    pl.show()
 
-    p = pyvista.Plotter()
-    p.add_mesh(data, scalars='vector_values_cells', component=0)
-    p.show()
+    pl = pyvista.Plotter()
+    pl.add_mesh(multicomp_poly, scalars='vector_values_cells', component=0, clim=[3, 5])
+    pl.show(before_close_callback=verify_cache_image)
```
adeak, line 1295: The vecor values in the image cache all seem to be 3, so setting a clim of [3, 5] makes the plot all blue. Can we have more variation for more robust image caching?

I guess this is because the vector data is a constant along the poly...

adeak (commented):
I'd also want to look at the build artifact but something with notebook plotting seems to blow up. Look similar to the "downstream filters don't support `None` blocks".

tests/plotting/test_plotting.py
akaszynski: Fixed in 12a638.

akaszynski:
> I'd also want to look at the build artifact but something with notebook plotting seems to blow up. Look similar to the "downstream filters don't support `None` blocks".

Issue was that the pythreejs converter assumed the mapper was a regular dataset mapper.

tests/plotting/test_plotting.py
```diff
@@ -1254,117 +1279,97 @@ def test_plot_rgb():
...
-    vector_values_cells[:, :] = [3.0, 4.0, 0.0]  # Vector has this value at all cells
-
-    data['vector_values_points'] = vector_values_points
-    data['vector_values_cells'] = vector_values_cells
-
-    return data
-
-
-def test_vector_array_with_cells_and_points():
+def test_vector_array_with_points(multicomp_poly):
```
adeak, line 1282: The corresponding image cache items seem to be missing. (I noticed because the plot I had in mind disappeared.)

akaszynski: Added.

akaszynski:
Merging this shortly. I know it's a huge PR, but it's been well reviewed and even reached tentative approval by bane and has some initial user adoption in https://github.com/pyvista/pyvistaqt/discussions/227.

adeak:
I've pushed two minor changes.

Probably related to #3168 rather than this PR, but only `__init__` seems to be rendered in doc builds for the new helper classes, e.g. this is the entirety of the BlockAttributes doc page (also in the CI artifact):
Screenshot 2022-08-19 at 14-49-51 BlockAttributes — PyVista 0 37 dev0 documentation

adeak:
OK, method autodoc is missing even in a full doc build. I think this needs fixing here.

akaszynski:
> OK, method autodoc is missing even in a full doc build for these three:

Issue fixed in ff73e612c7bedccf8fb94202dd573c3faa9e72ca. Let's (finally) merge this today.

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,1032 @@
...
+        Examples
+        --------
+        Enable coloring missing values with NaN.
+
+        >>> import pyvista as pv
+        >>> dataset = pv.MultiBlock([pv.Cube(), pv.Sphere(center=(0, 0, 1))])
+        >>> pl = pv.Plotter()
+        >>> actor, mapper = pl.add_composite(dataset)
+        >>> mapper.color_missing_with_nan = True
+        >>> mapper.color_missing_with_nan
```
adeak, line 626: This example's output is not very clear:

Screenshot from 2022-08-21 00-15-56

I assume the issue is that there are no "missing values" because there are no scalars chosen at all? Since we already have a full MultiBlock and Plotter set up, can we make this a bit more informative? Or just close the plotter to prevent it from being rendered, perhaps?

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,1032 @@
...
+        Examples
+        --------
+        Disable interpolation before mapping.
+
+        >>> import pyvista as pv
+        >>> dataset = pv.MultiBlock([pv.Cube(), pv.Sphere(center=(0, 0, 1))])
+        >>> pl = pv.Plotter()
+        >>> actor, mapper = pl.add_composite(dataset)
+        >>> mapper.interpolate_before_map = False
+        >>> mapper.interpolate_before_map
```
adeak, line 558: Similar almost-blank example figure. Especially with the link to the Interpolate before map example we can probably skip rendering this.

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,1032 @@
...
+        --------
+        Return the scalar range. In this example it's set to its default value
+        of ``(0.0, 1.0)``.
+
+        >>> import pyvista as pv
+        >>> dataset = pv.MultiBlock([pv.Cube(), pv.Sphere(center=(0, 0, 1))])
+        >>> pl = pv.Plotter()
+        >>> actor, mapper = pl.add_composite(dataset)
+        >>> mapper.scalar_range
+        (0.0, 1.0)
```
adeak, line 672: Another blank example.

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,1032 @@
...
+        >>> pl = pv.Plotter()
+        >>> actor, mapper = pl.add_composite(dataset)
+        >>> mapper.block_attr[1].visible = False
+        >>> mapper.block_attr.reset_visibilities()
+        >>> pl.show()
+
+        """
+        self.RemoveBlockVisibilities()
+
+    def reset_pickability(self):
```
adeak, line 327: This method is named singular ("pickability"), but we also have `reset_colors()`, `reset_opacities()`, `reset_visibilities()`. I think these should all be plural.

pyvista/plotting/composite_mapper.py
```diff
@@ -0,0 +1,1032 @@
...
+
+        Note how the zero index is the entire multiblock, so we have to add 1
+        to our indexing to access the right block.
+
+        >>> import pyvista as pv
+        >>> dataset = pv.MultiBlock([pv.Cube(), pv.Sphere(center=(0, 0, 1))])
+        >>> pl = pv.Plotter()
+        >>> actor, mapper = pl.add_composite(dataset)
+        >>> mapper.block_attr[1].pickable = True
+        >>> mapper.block_attr[2].pickable = False
```
adeak, line 219: This is also a blank-looking figure, but the feature being exemplified here is not visible anyway. Should we perhaps also enable picking in the example with some simple callback (random color? shifting position?), so that interactively running the example as-is will be instructive? In that case blankness would be fine too.

Or perhaps just link to `enable_block_picking()`.

adeak (commented):
Last round of comments, post build fix. I've also pushed some trivial changes.

pyvista/plotting/composite_mapper.py
akaszynski: Fixed in 8835c292f03a212cc953200b58f421e82e67c120.

pyvista/plotting/composite_mapper.py
akaszynski: Linked to `enable_block_picking()` in 8835c292f03a212cc953200b58f421e82e67c120.

pyvista/plotting/composite_mapper.py
akaszynski: Fixed in 8835c292f03a212cc953200b58f421e82e67c120.

pyvista/plotting/composite_mapper.py
akaszynski: Fixed in 8835c292f03a212cc953200b58f421e82e67c120.

pyvista/plotting/composite_mapper.py
akaszynski: Agreed. Example improved in 8835c292f03a212cc953200b58f421e82e67c120.

examples/02-plot/interpolate-before-map.py
```diff
@@ -1,16 +1,20 @@
...
 Interpolate Before Mapping
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The ``add_mesh`` function has an ``interpolate_before_map`` argument - this
-affects the way scalar data is visualized with colors.
-The effect can of this can vary depending on the dataset's topology and the
-chosen colormap.
+The :func:`add_mesh <pyvista.Plotter.add_mesh>` method has an
+``interpolate_before_map`` argument that affects the way scalar data is
+visualized with colors.  The effect can of this can vary depending on the
```
adeak, line 9:
```suggestion
visualized with colors.  The effect of this can vary depending on the
```

adeak (dismissed):
This is as reviewed as I can get a 4600-diff PR. Haven't built the last commit's changes but they look good to me. Approving with fingers crossed.

adeak (approved):
Re-approving...

Issue #137, opened 2019-03-21
layers in the renderer


banesullivan:
We should implement layers in the renderer to have a dedicated layer for holding all the scalar bars, widgets, etc. that are not datasets.

This will be needed for at least the `eye_dome_lighting` mode to prevent the scalar bars from getting shaded like they currently do:

(see [example](http://docs.vtki.org/en/latest/examples/quick/plot/plot-edl.html) to recreate)

![93a52a880ebf39b8782f994e2ef4441e13b88a11](https://user-images.githubusercontent.com/22067021/54766721-65c9b500-4bc1-11e9-9589-b378f812c898.png)

Implementation References:

- https://discourse.vtk.org/t/eye-dome-lighting-vtkedlshading-example-request/508/6
- https://public.kitware.com/pipermail/vtkusers/2008-July/047110.html

@akaszynski: I'm not sure how we should go about this... or if there's an easy way to implement this with the current structure of `vtki`. I'm hoping it's as simple as just setting the active layer when adding scalar bars and widgets but a brief simple hack at this has been unsuccessful for me.

We'll also have to make sure that the widget layer is ignored on the render passes where the eye dome lighting is implemented. Note this would also have a benefit to those using the depth mapping feature (see d5864fd9ab866973e584b45253ce52b5485824d6) @tomtomatron:

```py
import vtki
from vtki import examples
import numpy as np
import matplotlib.pyplot as plt

mesh = examples.load_uniform()

p = vtki.Plotter()
p.add_mesh(mesh)
p.show(auto_close=False) # DO NOT close plotter

# Get the depth imag
img = np.sum(p.image_depth, axis=2)

plt.imshow(img)
plt.colorbar()
plt.show()
```

![download](https://user-images.githubusercontent.com/22067021/54773419-26a26080-4bcf-11e9-94ed-b778f59c5f5b.png)

GuillaumeFavelier:
So far, following your discussion on the [VTK discourse](https://discourse.vtk.org/t/eye-dome-lighting-vtkedlshading-example-request/508/4), I managed to display the scalar bar in a different rendering pass with this pure `vtk` [gist example](https://gist.github.com/GuillaumeFavelier/cfce57ecf911906b638eccb66463c1b5). Maybe it could help?

![image](https://user-images.githubusercontent.com/18143289/59145675-5e0ef500-89e7-11e9-99be-9fa745825ab7.png)

GuillaumeFavelier:
[...]

PR #2948, opened 2022-07-07
Improve pass rendering

### LINKED ISSUE #3096: Improve the pass rendering pipeline
Add the following pass pipelines and ensure they all work when enabled.

- [x] `vtkEDLShading`
- [x] `vtkDepthOfFieldPass`
- [x] `vtkGaussianBlurPass`
- [x] `vtkOpenGLFXAAPass`
- [x] `vtkSSAAPass`
- [ ] `vtkSSAOPass`

akaszynski:
Improve the pass rendering pipeline through the creation of a new class `RenderPasses`. Resolves #3096.

The biggest limitation of our current pass rendering is you can only enable one. This class enables multi-pass rendering as VTK intended, while making it accessible to the user.

sc

```py
from pyvista.plotting.render_passes import RenderPasses


import pyvista
from pyvista import _vtk
import numpy as np
import vtk

pdata = pyvista.Sphere()
pl = pyvista.Plotter(lighting=None)
pl.add_mesh(pdata, show_edges=True, smooth_shading=True)
pl.add_light(pyvista.Light((0, 0, 100)))
pl.add_mesh(pyvista.Plane(center=(0, 0, -1), i_size=4, j_size=4))

# show RenderPasses class usage (this is only for a demo, will be exposed through
# `pyvista.Plotter`
passes = RenderPasses(pl.renderer)
passes.add_blur_pass()
passes.enable_edl_pass()
passes.enable_shadow_pass()
passes.enable_ssaa_pass()
passes.enable_depth_of_field_pass()

# you can also disable and re-enable passes
# passes.disable_shadow_pass()
# passes.disable_edl_pass()

pl.show()
```

One of the critical parts of this class is the ordering of the passes, which simply does not exist in our current pass rendering. Without this pass ordering, it's possible to improperly configure the passes such that their outputs are overwritten by downstream passes.

```py
# ordered
PRE_PASS = [
    'vtkEDLShading',
]

# ordered
POST_PASS = [
    'vtkDepthOfFieldPass',
    'vtkGaussianBlurPass',
    'vtkOpenGLFXAAPass',
    'vtkSSAAPass',  # should really be last
]
```


akaszynski:
Pinging @banesullivan

pyvista/plotting/plotting.py
```diff
@@ -968,16 +968,106 @@ def enable_lightkit(self, only_active=False):
...
+        SSAA and results in comparable image quality.
+
+        FXAA, or Fast Approximate Anti-Aliasing is an Anti-Aliasing technique
+        that is performed entirely in post processing. FXAA operates on the
+        rasterized image rather than the scene geometry. As a consequence,
+        forcing FXAA or using FXAA incorrectly can result in the FXAA filter
+        smoothing out parts of the visual overlay that are usually kept sharp
+        for reasons of clarity as well as smoothing out textures. FXAA is
+        inferior to MSAA but is almost free computationally and is thus
+        desirable on low end platforms.
```
adeak, line 1019: This sounds very different from the Nvidia sales pitch in the anti aliasing example ;) Perhaps we should mention some drawbacks there too?

pyvista/plotting/plotting.py
```diff
@@ -968,16 +968,106 @@ def enable_lightkit(self, only_active=False):
...
+        >>> _ = pl.add_mesh(pyvista.Sphere(), show_edges=True)
+        >>> pl.show()
+
+        """
+        # apply MXAA to entire render window
+        if aa_type == 'mxaa':
+            if hasattr(self, 'ren_win'):
+                self.ren_win.SetMultiSamples(multi_samples)
+            else:
+                self.theme.multi_samples = multi_samples
```
adeak, line 1037: Half of this `if` is not covered according to codecov, and I don't understand the mechanics here to be honest. Is `SetMultiSamples(multi_samples)` exactly MSAA? And if there's no render window (which seems to always be the case according to code cov), just set this for the theme? Why set it for the theme, and why not set it for the first conditional branch?

pyvista/plotting/plotting.py
```diff
@@ -968,16 +968,106 @@ def enable_lightkit(self, only_active=False):
...
+        Examples
+        --------
+        >>> import pyvista
+        >>> pl = pyvista.Plotter()
+        >>> pl.disable_anti_aliasing()
+        >>> _ = pl.add_mesh(pyvista.Sphere(), show_edges=True)
+        >>> pl.show()
+
+        """
+        self.ren_win.SetMultiSamples(0)
```
adeak, line 1064: A-ha! This is why `Renderer.disable_anti_aliasing()` didn't do anything MSAA-y...

pyvista/plotting/render_passes.py
```diff
@@ -0,0 +1,255 @@
...
+        """
+        if self.__pass_collection is None:
+            self._init_passes()
+        return self.__pass_collection
+
+    @property
+    def _seq_pass(self):
+        """Initialize (when necessary the pass collection and return it.
+
+        This lets us lazily generate the pass collection only when we need it
```
adeak: "pass collection" was copied over here. I don't know what "seq_pass" exactly means, so I won't try to fix this.

pyvista/plotting/render_passes.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # Make the sequence the delegate of a camera pass.
+        self.__camera_pass = _vtk.vtkCameraPass()
+        self.__camera_pass.SetDelegatePass(self._seq_pass)
+
+    @property
+    def _renderer(self):
+        """Return the renderer."""
+        if self._renderer_ref is not None:
+            return self._renderer_ref()
```
adeak, line 103: Was this meant to say
```suggestion
        if self._renderer_ref() is not None:
            return self._renderer_ref()
```
?

pyvista/plotting/render_passes.py
```diff
@@ -0,0 +1,255 @@
...
+        """
+        blur_pass = _vtk.vtkGaussianBlurPass()
+        self._add_pass(blur_pass)
+        self._blur_passes.append(blur_pass)
+        return blur_pass
+
+    def remove_blur_pass(self):
+        """Add a vtkGaussianBlurPass pass."""
+        if self._blur_passes:
+            self._remove_pass(self._blur_passes.pop(0))
```
adeak, line 151: When we add blur passes we append. Does it make sense to `pop(0)` when we remove? Shouldn't we pop off the most recent one? I have no idea how this works, sorry...

pyvista/plotting/render_passes.py
```diff
@@ -0,0 +1,255 @@
...
+
+        current_pass = self._camera_pass
+        for class_name in PRE_PASS + POST_PASS:
+            if class_name in self._passes:
+                for render_pass in self._passes[class_name]:
+                    render_pass.SetDelegatePass(current_pass)
+                    current_pass = render_pass
+
+        # reset to the default rendering if no special passes have been added
+        if isinstance(current_pass, _vtk.vtkCameraPass) and self._shadow_map_pass is None:
```
adeak, line 216: Can this first check actually be `if current_pass is self._camera_pass`? Might be more explicit if so. Or can there be other `vtkCameraPass`es involved?

pyvista/plotting/render_passes.py
```diff
@@ -0,0 +1,255 @@
...
+            self._renderer.SetPass(None)
+        else:
+            self._renderer.SetPass(current_pass)
+
+    @staticmethod
+    def _class_name_from_vtk_obj(obj):
+        """Return the class name from a vtk object."""
+        return str(type(obj)).split('.')[-1].split("'")[0]
```
adeak, line 224: I think this should work:
```suggestion
        return type(obj).__name__
```
Of course if this works we might not need a staticmethod for it.

tests/test_render_pass.py
```diff
@@ -0,0 +1,124 @@
+from pyvista import _vtk
+from pyvista.plotting.render_passes import RenderPasses
+
+
+# this ideally would be a fixture, but if it's a fixture renderer collects
+# immediately
```
adeak, line 6: Can you explain this comment? Why can't a fixture return these two? Or what do you mean by "collects"?

tests/test_render_pass.py
```diff
@@ -0,0 +1,124 @@
...
+    assert not passes._passes
+    ssaa_pass = passes.enable_ssaa_pass()
+    assert isinstance(ssaa_pass, _vtk.vtkSSAAPass)
+    assert 'vtkSSAAPass' in passes._passes
+
+    # enabling again should just do nothing
+    ssaa_pass = passes.enable_ssaa_pass()
+    assert 'vtkSSAAPass' in passes._passes
+
+    # disabling again should just do nothing
```
adeak, line 49: This comment seems off, because disabling it again does a lot. But I don't know if you'd prefer to edit the comment or remove it altogether. Or perhaps you meant to disable it once first.

tests/test_render_pass.py
```diff
@@ -0,0 +1,124 @@
...
+def test_ssaa_pass():
+    ren, passes = make_passes()
+    assert not passes._passes
+    ssaa_pass = passes.enable_ssaa_pass()
+    assert isinstance(ssaa_pass, _vtk.vtkSSAAPass)
+    assert 'vtkSSAAPass' in passes._passes
+
+    # enabling again should just do nothing
+    ssaa_pass = passes.enable_ssaa_pass()
+    assert 'vtkSSAAPass' in passes._passes
```
adeak, line 47: Can we/should we check some pass length to show that it "does nothing"?

examples/02-plot/anti-aliasing.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+###############################################################################
+# Fast Approximate Anti-Aliasing (FXAA)
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Also known as Fast Approximate Anti-Aliasing, FXAA was crafted by
+# Nvidia. When it comes to anti-aliasing on low-end devices/PCs, FXAA is the
+# best technique. This is because, in terms of hardware or GPU, FXAA is not
+# that demanding. It directly smoothens the 2D image as soon as it appears on
+# the screen. This reduces the strain over GPU, making it best for low-end PCs.
```
adeak, line 54: I didn't understand "directly smoothens the 2D image as soon as it appears on the screen" at all, first. Having looked at the rest of the PR I think this means that it only blurs a rendered 2d image at the end, but this was not clear to me at all.

pyvista/plotting/plotting.py
```diff
@@ -5406,6 +5516,10 @@ def show(
 
         self.render()
 
+        # initial double render needed for certain passes when offscreen
+        if self.off_screen and 'vtkDepthOfFieldPass' in self.renderer._render_passes._passes:
+            self.render()
```
adeak, line 5521: Codecov said that this branch was uncovered, not sure why. Does that make sense? Or is it just that we have no `plotter.show()` with DOF in our test suite? `test_many_multi_pass()` should cover that case, so I'm not sure why this wasn't entered.

By the way, I reversed the order of the two conditions here, because `self.off_screen` should be `False` most of the time in user code, short-circuiting.

banesullivan:
Sorry for the delay. I will review this asap

examples/02-plot/anti-aliasing.py
```diff
@@ -0,0 +1,95 @@
...
+* ``"fxaa"`` - Fast Approximate Anti-Aliasing
+
+By default, anti-aliasing is disabled, but can be enabled globally with:
+
+.. code:: python
+
+   >>> import pyvista as pv
+   >>> pv.global_theme.antialiasing = 'ssaa'
+
+**Which Anti Aliasing Technique should You use?**
```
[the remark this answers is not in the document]
banesullivan:
```suggestion
**Which anti-aliasing technique should you use?**
```

pyvista/_vtk.py
```diff
@@ -443,14 +443,24 @@
...
         vtkOpenGLHardwareSelector,
         vtkOpenGLRenderer,
         vtkOpenGLTexture,
         vtkRenderPassCollection,
         vtkRenderStepsPass,
         vtkSequencePass,
         vtkShadowMapPass,
+        vtkSSAAPass,
+        vtkTranslucentPass,
+        vtkVolumetricPass,
```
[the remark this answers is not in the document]
banesullivan: +1 for only importing what we need. some imports can actually have overhead... not sure if these in particular do though

pyvista/plotting/render_passes.py
banesullivan: either way, why even have the check? it'll return `None` if the condition evaluates to false (being `None`)

```suggestion
        return self._renderer_ref()
```

pyvista/plotting/render_passes.py
banesullivan: I agree that it should be LIFO

pyvista/plotting/render_passes.py
banesullivan: There is also the `GetClassName()` all all `vtkObject`s

pyvista/plotting/render_passes.py
adeak: Well if the original version is correct, then it's checking for the case of `self._renderer_ref = None` (rather than a stale weakref), in which case trying to call `self._renderer_ref()` would give a `TypeError`.

pyvista/_vtk.py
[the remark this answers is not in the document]
adeak: Importing from the same module should be practically free (dict lookup), but we shouldn't clutter the namespace.

akaszynski:
Note to self, add in `vtkSSAOPass`.

```py
import vtk
import numpy as np
import pyvista as pv

cells = np.array(
    [4, 6, 5, 8, 7, 4, 7, 3, 8, 9, 4, 7, 3, 1, 5, 4, 9, 3, 1, 7, 4, 2,
     6, 5, 8, 4, 2, 6, 0, 4, 4, 6, 2, 0, 8, 4, 5, 2, 8, 3, 4, 5, 3, 8,
     7, 4, 2, 6, 4, 5]
)

celltypes = np.array([10, 10, 10, 10, 10, 10, 10, 10, 10, 10], dtype=np.uint8)
points = np.array([[-0.  ,  0.  , -0.5 ],
       [ 0.  ,  0.  ,  0.5 ],
       [-0.43,  0.  , -0.25],
       [-0.43,  0.  ,  0.25],
       [-0.  ,  0.43, -0.25],
       [ 0.  ,  0.43,  0.25],
       [ 0.43,  0.  , -0.25],
       [ 0.43,  0.  ,  0.25],
       [ 0.  , -0.43, -0.25],
       [ 0.  , -0.43,  0.25]])

# Create and plot the unstructured grid
grid = pv.UnstructuredGrid(cells, celltypes, points)

split_cells = pv.MultiBlock()
for index in range(10):
    single_cell = grid.extract_cells([index])
    single_cell.points += (np.array(single_cell.center) - np.array(grid.center))*0.5
    split_cells.append(single_cell)



pl = pv.Plotter()
pl.add_mesh(split_cells, show_edges=True)

basicPasses = vtk.vtkRenderStepsPass()
sceneSize = 5

ssao = vtk.vtkSSAOPass()
ssao.SetRadius(0.1 * sceneSize)
ssao.SetBias(0.001 * sceneSize)
ssao.SetKernelSize(2**8)
# ssao.BlurOff()
ssao.BlurOn()
ssao.SetDelegatePass(basicPasses)

pl.renderer.SetPass(ssao);

pl.background_color = 'w'
pl.show()
```
tmp

adeak:
I also wonder if https://github.com/pyvista/pyvista/issues/137 could be plugged in here. Or at least built on top of this later.

pyvista/_vtk.py
akaszynski, line 463:
```suggestion
        vtkDepthOfFieldPass,
        vtkEDLShading,
        vtkGaussianBlurPass,
        vtkOpenGLFXAAPass,
        vtkOpenGLHardwareSelector,
        vtkOpenGLRenderer,
        vtkOpenGLTexture,
        vtkRenderPassCollection,
        vtkRenderStepsPass,
        vtkSequencePass,
        vtkShadowMapPass,
        vtkSSAAPass,
```

tests/test_theme.py
```diff
@@ -434,8 +435,28 @@ def test_load_theme(tmpdir, default_theme):
...
-        assert pl.renderer.GetUseFXAA() is value
+        if value == 'fxaa':
+            assert pl.renderer.GetUseFXAA()
+        else:
+            assert not pl.renderer.GetUseFXAA()
+
+        if value == 'ssaa':
+            assert 'vtkSSAAPass' in pl.renderer._render_passes._passes
+        else:
+            assert 'vtkSSAAPass' not in pl.renderer._render_passes._passes
```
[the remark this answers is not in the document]
akaszynski: Test logic cleaned up in c06019a23498510cfbbc9322958cae54575a66fd.

examples/02-plot/anti-aliasing.py
```diff
@@ -0,0 +1,95 @@
+"""
+.. _antialiasing_example:
+
+Anti-Aliasing
+~~~~~~~~~~~~~
+
+PyVista supports three types of anti-aliasing:
+
+* ``"ssaa"`` - Super-Sample Anti-Aliasing
+* ``"mxaa"`` - Multi-Sample Anti-Aliasing
```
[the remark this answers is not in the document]
akaszynski: Renamed in c06019a23498510cfbbc9322958cae54575a66fd. There's still work to be done in the docs.

tests/test_theme.py
```diff
@@ -434,8 +435,28 @@ def test_load_theme(tmpdir, default_theme):
...
+
+    with pytest.warns(PyvistaDeprecationWarning, match='is now a string'):
+        default_theme.antialiasing = True
+        pl = pyvista.Plotter(theme=default_theme)
+        if value == 'fxaa':
+            assert pl.renderer.GetUseFXAA()
```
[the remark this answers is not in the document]
akaszynski: Fixed in c06019a23498510cfbbc9322958cae54575a66fd.

examples/02-plot/anti-aliasing.py
```diff
@@ -0,0 +1,95 @@
...
+* ``"fxaa"`` - Fast Approximate Anti-Aliasing
+
+By default, anti-aliasing is disabled, but can be enabled globally with:
+
+.. code:: python
+
+   >>> import pyvista as pv
+   >>> pv.global_theme.antialiasing = 'ssaa'
```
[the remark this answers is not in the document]
akaszynski: I'd love to, but it's actually harder than it looks.
- `antialiasing` as a property with a setter won't work for the `Plotter` as there are various configuration parameters that we'd have to pass (e.g. `all_renderers`, `multi_samples`).
- `enable_anti_aliasing` is a method, and our class is setup with just attributes. Seems weird to have an all attribute class (similar to [dataclasses](https://docs.python.org/3/library/dataclasses.html)) with a method that changes its state.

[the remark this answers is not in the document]
adeak: To be clear I specifically meant `antialiasing` vs `anti_aliasing`, the underscore. So `enable_antialiasing()` rather than trying to make this a property.

examples/02-plot/anti-aliasing.py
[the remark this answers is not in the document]
akaszynski: Fixed docs in ddce716b89237e4d4631a1f0a62eb575c27b4a65.

examples/02-plot/anti-aliasing.py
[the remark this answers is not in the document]
akaszynski: Oh! Ok, fully agree there. We can depreciate this.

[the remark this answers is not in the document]
adeak: For future consideration: `pyvista/core/dataset.py` vs `pyvista/core/filters/data_set.py` :smile:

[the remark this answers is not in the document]
akaszynski: Done in ee41207f500145104ddae02c1821ffe44159ee42.

pyvista/themes.py
```diff
@@ -1857,23 +1857,45 @@ def title(self, title: str):
         self._title = title
 
     @property
-    def antialiasing(self) -> bool:
+    def antialiasing(self) -> Optional[str]:
         """Enable or disable anti-aliasing.
```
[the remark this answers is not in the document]
akaszynski: Done.

pyvista/themes.py
```diff
@@ -1857,23 +1857,45 @@ def title(self, title: str):
...
         """
         return self._antialiasing
 
     @antialiasing.setter
-    def antialiasing(self, antialiasing: bool):
+    def antialiasing(self, antialiasing: Union[str, None]):
+        if isinstance(antialiasing, bool):
+            warnings.warn(
+                '`antialiasing` is now a string or None and must be either "ssaa", '
+                '"mxaa", "fxaa", or None',
```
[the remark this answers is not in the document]
akaszynski: The `antialiasing` --> `anti_aliasing` refactor mostly accounts for this edge case.

pyvista/themes.py
```diff
@@ -1857,23 +1857,45 @@ def title(self, title: str):
...
+                PyvistaDeprecationWarning,
+            )
+            antialiasing = 'fxaa'
```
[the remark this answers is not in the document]
akaszynski: Addressed in 8a37f5cbb4c0a0a88fb02883c9dc90b209c19d69.

tests/plotting/test_plotting.py
```diff
@@ -47,6 +47,10 @@
 
 skip_9_1_0 = pytest.mark.skipif(pyvista.vtk_version_info < (9, 1, 0), reason="Requires VTK>=9.1.0")
 
+skip_9_0_X = pytest.mark.skipif(
+    pyvista.vtk_version_info < (9, 1) and pyvista.vtk_version_info > (8, 2), reason="Flaky on 9.0.X"
```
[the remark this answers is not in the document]
akaszynski: Reads better. I much prefer the bounded inequalities.

pyvista/plotting/renderer.py
```diff
@@ -437,48 +439,49 @@ def disable_depth_peeling(self):
...
-        self.Modified()
+        else:
+            raise ValueError(f'Invalid `aa_type` "{aa_type}". Should be either "fxaa" or "ssaa"')
 
     def disable_anti_aliasing(self):
-        """Disable anti-aliasing.
+        """Disable all anti-aliasing."""
+        self._render_passes.disable_ssaa_pass()
+        self.SetUseFXAA(False)
+        self.Modified()
```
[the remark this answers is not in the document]
akaszynski: That needs to happen at the render window level, and while I'm tempted to do it here, I don't like having a side effect for a render window.

pyvista/plotting/renderer.py
```diff
@@ -437,48 +439,49 @@ def disable_depth_peeling(self):
...
 
-        Examples
-        --------
-        >>> import pyvista
-        >>> pl = pyvista.Plotter()
-        >>> pl.disable_anti_aliasing()
-        >>> _ = pl.add_mesh(pyvista.Sphere(), show_edges=True)
-        >>> pl.show()
```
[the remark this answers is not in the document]
akaszynski: I don't think it's necessary, but this can be up for debate. Typical use is to have the user interact at the `Plotter` level. This example above even directs the user to `disable_anti_aliasing` from the `Plotter` class.

tests/plotting/test_plotting.py
```diff
@@ -2459,6 +2463,43 @@ def test_add_text():
...
+@skip_windows
+def test_many_multi_pass():
+    pl = pyvista.Plotter(lighting=None)
+    pl.add_mesh(pyvista.Sphere(), show_edges=True)
+    pl.add_light(pyvista.Light(position=(0, 0, 10)))
+    pl.enable_anti_aliasing('ssaa')
+    pl.enable_depth_of_field()
+    pl.add_blurring()
+    pl.enable_shadows()
+    pl.enable_eye_dome_lighting()
```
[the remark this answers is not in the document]
akaszynski: Order matters. I've added note for this in 6347b7dca061df01f565d5825e4c47db4a8c21e8.

examples/02-plot/depth_of_field.py
```diff
@@ -0,0 +1,71 @@
...
+colors /= colors.max(axis=0)
+colors *= 255
+colors = colors.astype(np.uint8)
+
+# obtained camera position with `cpos = pl.show(return_cpos)`
+cpos = [(11.6159, -1.2803, 1.5338), (4.1354, 1.4796, 1.2711), (-0.0352, -0.0004, 1.0)]
+
+# since we're using physically based rendering, let's also download a skybox
+# cubemap
+cubemap = examples.download_sky_box_cube_map()
```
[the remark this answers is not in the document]
akaszynski: Done in 9edba318744b1b81084bc896d4c23b4d8d630d54.

pyvista/plotting/renderer.py
```diff
@@ -437,48 +439,49 @@ def disable_depth_peeling(self):
...
-            # only display the warning when not building documentation
-            if not pyvista.BUILDING_GALLERY:
-                warnings.warn(
-                    "VTK compiled with OSMesa does not properly support anti-aliasing and anti-aliasing will not be enabled."
-                )
-            return
-        self.SetUseFXAA(True)
-        self.Modified()
+        else:
+            raise ValueError(f'Invalid `aa_type` "{aa_type}". Should be either "fxaa" or "ssaa"')
```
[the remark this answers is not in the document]
akaszynski: Resolved in 6347b7dca061df01f565d5825e4c47db4a8c21e8.

tests/test_render_pass.py
akaszynski: Meant to do it twice, thanks for catching that. Fixed in 69243a9ff453f7d292ce9329026b0ef8c43fa355.

pyvista/plotting/renderer.py
```diff
@@ -437,48 +439,49 @@ def disable_depth_peeling(self):
...
+        if aa_type == 'fxaa':
+            if uses_egl():  # pragma: no cover
+                # only display the warning when not building documentation
+                if not pyvista.BUILDING_GALLERY:
+                    warnings.warn(
+                        "VTK compiled with OSMesa does not properly support "
+                        "FXAA anti-aliasing and SXAA will be used instead."
+                    )
+                self._render_passes.enable_ssaa_pass()
+            self._enable_fxaa()
```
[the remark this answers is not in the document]
akaszynski: Good point, and this is a mistake. This would have still resulted in black images. Fixed in 69243a9ff453f7d292ce9329026b0ef8c43fa355

pyvista/plotting/renderer.py
```diff
@@ -2506,6 +2509,95 @@ def enable(self):
...
+        >>> pl.add_blurring()
+        >>> pl.show()
+
+        See :ref:`blur_example` for a full example using this method.
+
+        """
+        self._render_passes.add_blur_pass()
+
+    def remove_blurring(self):
+        """Remove blurring.
```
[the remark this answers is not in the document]
akaszynski: Added note in 69243a9ff453f7d292ce9329026b0ef8c43fa355.

pyvista/plotting/plotting.py
akaszynski: Added in 69243a9ff453f7d292ce9329026b0ef8c43fa355. NVIDIA language removed.

pyvista/plotting/plotting.py
akaszynski: Fix was just to move the set `enable_anti_aliasing` to the end of the Plotter initialization in 69243a9ff453f7d292ce9329026b0ef8c43fa355.

pyvista/plotting/render_passes.py
akaszynski: Fixed in 69243a9ff453f7d292ce9329026b0ef8c43fa355.

pyvista/plotting/render_passes.py
akaszynski: Order of the blur passes doesn't matter. Added note in 69243a9ff453f7d292ce9329026b0ef8c43fa355.

pyvista/plotting/render_passes.py
akaszynski: Good point. Fixed in 69243a9ff453f7d292ce9329026b0ef8c43fa355.

pyvista/plotting/render_passes.py
akaszynski: Went with bane's approach in 69243a9ff453f7d292ce9329026b0ef8c43fa355.

tests/test_render_pass.py
akaszynski: Added in 69243a9ff453f7d292ce9329026b0ef8c43fa355.

pyvista/plotting/render_passes.py
```diff
@@ -0,0 +1,250 @@
...
+        blur_pass = _vtk.vtkGaussianBlurPass()
+        self._add_pass(blur_pass)
+        self._blur_passes.append(blur_pass)
+        return blur_pass
+
+    def remove_blur_pass(self):
+        """Remove a single vtkGaussianBlurPass pass."""
+        if self._blur_passes:
+            # order of the blur passes does not matter
+            self._remove_pass(self._blur_passes.pop(0))
```
adeak, line 151: If order doesn't matter then we should definitely do `self._blur_passes.pop()`: popping from the start of the list has to shift the entire contents, popping from the end is `O(1)`. Of course this is a micro-optimization, because in most cases there are < 5 blur passes, but anyway a good practice.

tests/test_render_pass.py
```diff
@@ -0,0 +1,127 @@
+from pyvista import _vtk
+from pyvista.plotting.render_passes import RenderPasses
+
+
+# this ideally would be a fixture, but if it's a fixture the renderer object
+# collects immediately since RenderPasses only holds a weakref
+def make_passes():
+    ren = _vtk.vtkRenderer()
+    passes = RenderPasses(ren)
+    return ren, passes
```
adeak, line 10: OK, I understand the updated comment then, thanks. So how about returning both the renderer and the passes from the fixture? It would all look the same, just called `passes` (or something like `renderer_with_passes`) instead of `make_passes`. The only downside is that you need to unpack or index into the fixture when you use it. And you also have to delete the name of the fixture (the tuple) in tests where you want to check collection.

akaszynski: I originally went with that approach but I felt it wasn't I. The spirit of fixtures and decided against it.

pyvista/plotting/render_passes.py
akaszynski:
```suggestion
            # order of the blur passes does not matter
            self._remove_pass(self._blur_passes.pop())
```

examples/02-plot/anti-aliasing.py
akaszynski: I’ve cleaned up the docs.

tests/test_render_pass.py
akaszynski: Done.

pyvista/plotting/plotting.py
akaszynski: Covered now.

akaszynski:
@adeak, ready for review.

akaszynski:
Need to merge this to avoid duplication of effort in support of #3216. This is a less scary change than #3216, if there are no further objections, I'd like to merge this by Wednesday.

pyvista/plotting/renderer.py
```diff
@@ -437,48 +439,50 @@ def disable_depth_peeling(self):
         self.SetUseDepthPeeling(False)
         self.Modified()
 
-    def enable_anti_aliasing(self):
-        """Enable anti-aliasing using FXAA.
+    def enable_anti_aliasing(self, aa_type):
```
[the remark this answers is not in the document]
akaszynski: Yes. Sorry about that.

With these big changes I'm considering adding in some optional workflows to verify we work against `mne`.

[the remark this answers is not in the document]
akaszynski: > I'm a CI ignoramus but it would make sense to me to have a dedicated CI run to test downstream. At the moment this would only include MNE, but it's plausible that we'd want to test other compatibility things against downstream, and then we could add that there.

Absolutely, and since their workflow in #3255 takes less time than our longest run, I'd really like for us to test against multiple downstream packages.

PR #2961, opened 2022-07-09
fix vtk version support

### LINKED ISSUE #2867: Supported vtk versions

PyVista should clarify what minimum vtk versions will be supported going forward.

Supporting more versions increases usability for users with previous versions, but decreases maintainability and velocity of new features.

In addition, we should consider whether to have a maximum supported version.  This would prevent issues like #2859.

### Links to source code.

PyVista has many places where specific versions of vtk have to be handled separately.  We are currently testing 8.2, 9.0, 9.1, and soon to be 9.2.  

8.2 in particular requires significant code overhead in the `_vtk` module. Multiple tests are failing for 8.2 and many are skipped entirely.

### Pseudocode or Screenshots

_No response_

### Code of Conduct

akaszynski:
Resolves #2867 by implementing matrix testing for VTK and Python.

- Python 3.7 --> ``vtk==8.1.2``
- Python 3.8 --> ``vtk==9.0.3``
- Python 3.9 --> ``vtk==9.1``
- Python 3.10 --> ``vtk==9.2.0``

Additionally, because we aren't testing Python 3.7 with 8.1.2, we've had some VTK version support regressions only discovered when locally testing 3.7. This PR fixes those regressions.

---

Also, I propose that we support the oldest VTK version with a viable wheel available for [actively supported versions of Python](https://endoflife.date/python)  available on PyPI. For example, the lowest supported version of Python as of this PR is Python 3.7, and the lowest supported version of VTK for 3.7 is [vtk==8.1.2](https://pypi.org/project/vtk/8.1.2/).

This is reflected in the ``install_requires`` in our ``setup.py``.

As soon as support for Python 3.7 is dropped, we can then drop support for both Python 3.7 and VTK 8.1.2 in both our ``python_requires`` and ``install_requires``.

As much as I would prefer to follow NumPy's version support [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) deprecation policy, I know of many who are stuck using older versions of Python and need support for older versions of VTK and cannot upgrade agressively. With this approach we can still depricate support for older versions of VTK and Python, but in a manner that works for all that use this library with any currently supported version of Python.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ install_requires = [
     'pillow',
     'appdirs',
     'scooby>=0.5.1',
-    'vtk',
+    'vtk>=8.1.2,<=9.2',
 ]
 
 readme_file = os.path.join(filepath, 'README.rst')
@@ -54,6 +54,7 @@ setup(
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
     ],
     url='https://github.com/pyvista/pyvista',
     keywords='vtk numpy plotting mesh',
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -138,12 +138,18 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
+
+      # see discussion at https://github.com/pyvista/pyvista/issues/2867
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
-        vtk-version: ['latest']
         include:
-          - python-version: '3.9'
+          - python-version: '3.7'
+            vtk-version: '8.1.2'
+          - python-version: '3.8'
             vtk-version: '9.0.3'
+          - python-version: '3.9'
+            vtk-version: '9.1'
+          - python-version: '3.10'
+            vtk-version: 'latest'
     steps:
       - uses: actions/checkout@v3
         with:
--- a/pyvista/_vtk.py
+++ b/pyvista/_vtk.py
@@ -464,6 +464,18 @@ if VTK9:
     from vtkmodules.vtkViewsContext2D import vtkContextInteractorStyle
 
     # lazy import for some of the less used readers
+    def lazy_vtkVRMLImporter():
+        """Lazy import of the vtkVRMLImporter."""
+        from vtkmodules.vtkIOImport import vtkVRMLImporter
+
+        return vtkVRMLImporter()
+
+    def lazy_vtkVRMLExporter():
+        """Lazy import of the vtkVRMLImporter."""
+        from vtkmodules.vtkIOImport import vtkVRMLExporter
+
+        return vtkVRMLExporter()
+
     def lazy_vtkGL2PSExporter():
         """Lazy import of the vtkGL2PSExporter."""
         from vtkmodules.vtkIOExportGL2PS import vtkGL2PSExporter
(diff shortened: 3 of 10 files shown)


MatthewFlamm (commented):
This, in general, looks good to me.  Maybe for a future PR: we mix skipping tests completely and checking that we raise with `VTKVersionError` in the code base.  In my opinion it would be better to check that it raises to make sure that we handle these cases with useful errors.

setup.py
```diff
@@ -18,7 +18,7 @@
     'pillow',
     'appdirs',
     'scooby>=0.5.1',
-    'vtk',
+    'vtk>=8.1.2,<9.3',
```
MatthewFlamm, line 21: Does `setup.py` support `<=9.2.0.*`?

akaszynski: I was having trouble permitting a release candidates with any version range unless you specifically permit it, as in:
https://github.com/pyvista/pyvista/blob/88a894084bfb93ddad4a8e38f9856f1b182a3c04/setup.py#L21-L22

What I did was explicitly permit the release candidate wheel, but only for Python 3.10. Personally, I don't feel that release candidates should be permitted by default, but only in special circumstances like general unavailability otherwise as in the case of Python 3.10

akaszynski:
> This, in general, looks good to me. Maybe for a future PR: we mix skipping tests completely and checking that we raise with `VTKVersionError` in the code base. In my opinion it would be better to check that it raises to make sure that we handle these cases with useful errors.

Agreed, that seems like a better practice.

setup.py
MatthewFlamm: This makes sense to me.  Ideally the rc part could be removed before the release in pyvista.  If it can't, should we use some form of `==9.2.0.*`, so that later release candidates can be used?  I'm not sure.  Maybe best to pin it and manually keep updating it until the release?  This way there are no surprises.

akaszynski: > If it can't, should we use some form of ==9.2.0.*, so that later release candidates can be used?

If someone knows how to get release candidates working with a version range, please let me know. Otherwise, I'll permit `==9.2.0rc` for PyVista for now.

setup.py
```diff
@@ -18,7 +18,8 @@
     'pillow',
     'appdirs',
     'scooby>=0.5.1',
-    'vtk',
+    'vtk>=8.1.2, <=9.2.0; python_version < "3.10"',
+    'vtk>=8.1.2, <=9.2.0, ==9.2.0rc1; python_version == "3.10"',
```
MatthewFlamm, line 22: This works if we want to allow any `9.2.x` version.  See https://peps.python.org/pep-0440/#compatible-release

Do we need the other parts for python 3.10 as it isn't supported unless a custom wheel is used?

```suggestion
    'vtk~=9.2.0rc1; python_version == "3.10"',
```

setup.py
MatthewFlamm: See my other suggestion, if that doesn't work, I'm in favor of `==9.2.0rc1` and manually bumping it as needed, and hopefully removing it before next release (vtk had a 1 month release cycle last time).

setup.py
akaszynski: I think we should just permit `vtk~=9.2.0rc1` for all python versions. Do you agree?

akaszynski:
> Should we also pin `python_requires` to be <= 3.10?

I'm not sure. I don't see an upper limit for `numpy`, and I think that we should permit those who want to use PyVista with an unsupported version of Python.

setup.py
MatthewFlamm: This would install the rc version even on the other versions. I don't think we should do this.  If vtk fully releases 9.2.0 before the next PyVista release, we can use `vtk>=8.1.2, <=9.2.0` without any python versions.

setup.py
akaszynski: Agreed then. Let's keep to this for now and hopefully 9.2.0 is out before the next release of PyVista.

akaszynski:
This is ready for review/approval. Checking showing up as yellow just means we need to update our branch protection rules once this PR is merged.

MatthewFlamm (approved):
LGTM.

PR #2960, opened 2022-07-09
Remove extra references

akaszynski:
Partial resolves #2577 by removing internal references that are causing `pyvista.Plotter` to not be collected.

Thanks @WesleyTheGeolien for pointing this out.

Memory leak for 1000 plotters goes from ~60 MB to ~7. It's some improvement but it's still annoying that these references aren't being collected.

Issue #2577, opened 2022-05-10
Plotter memory leak due to not clearing up


WesleyTheGeolien:
@banesullivan I saw this was closed in 2019 however I don't believe the whole problem has been addressed.

Just Running and destroying a pv.Plotter Instance in pyvista 0.32.1 (maybe later but untested) still produces a memory hogging for long running process.

Running the following code: (need to install memory_profiler)

```python

from memory_profiler import memory_usage

import matplotlib.pyplot as plt
import pyvista as pv
import numpy as np
import gc

from time import sleep

gc.enable()


def plot_mem(mem, label=None, show=False):
    if label is None:
        label = 'Usage'
    mem = np.array(mem) - mem[0]
    if show:
        plt.close('all')
        plt.plot(np.arange(len(mem)) * 0.1, mem, label=label)
        plt.ylabel('MB')
        plt.xlabel('time (in s)')  # `memory_usage` logs every 100ms
        plt.legend()
        plt.title('memory usage')
        plt.show()
    print(f'Max: {mem.max():.2f} MB\nLast: {mem[-1]:.2f} MB')


def plot_pyvista():
    # sphere = pv.Sphere()
    for i in range(1000):
        plotter = pv.Plotter()
        plotter.close()

    gc.collect()
    sleep(0.01)


if __name__ == '__main__':
    mem_full = memory_usage((plot_pyvista, [], {}))
    plot_mem(mem_full, label='Full', show=True)
```
We can see that 70 Mib is held over time:

![image](https://user-images.githubusercontent.com/18068320/164399547-8a4bd5d0-224b-4bf1-ac15-ea7e4015f91c.png)

Running the following code to try and understand where the memory was being held:

```python
import pyvista as pv

import pandas as pd
from pprint import pprint

from collections import defaultdict
from gc import get_objects
import gc

import sys
before = defaultdict(int)
before_size = defaultdict(int)
after = defaultdict(int)
after_size = defaultdict(int)
for i in get_objects():
    before[type(i)] += 1
    before_size[type(i)] += sys.getsizeof(i)

for i in range(1000):
    plotter = pv.Plotter(off_screen=False)
    plotter.close()

not_collected = gc.collect()


for i in get_objects():
    after[type(i)] += 1
    after_size[type(i)] += sys.getsizeof(i)

df = pd.DataFrame([(str(k), after[k] - before[k], (after_size[k] - before_size[k]))
[...]

PR #2964, opened 2022-07-10
Render window refactor

akaszynski:
This PR is the first phase of pulling out "render window" logic from our `pyvista.Plotter` class into it's own class. It's effecively a better implementation of #2960 and was necessary to work around some of the current limitations of our Plotter class. Namely being:
- It's too big and too much functionality is packed into two classes in module.
- It's brittle and monolithically joined to several other classes. There are cross references everywhere that are leading to objects not collecting.
- It's messy. Everytime we had to patch a bug we ended up adding a bandaid (hence why there's close, deep clean, etc.

The plan is to do this in 3 PRs:

1. Pull out some of the render window logic into `pyvista.RenderWindow` and make the `Renderers`, `Renderer` and `RenderWindowInteractor` independent.
2. Inherit `pyvista.RenderWindow` in the `pyvista.Plotter` class. Reason being:
   - `pyvista.Plotter` is effecively a correlary of `vtkRenderWindow` with lots of bells and whistles. It would be better to separate out the real `vtkRenderWindow` logic into its own class and then leave `pyvista.Plotter` as the glue and logic center to allow users to add datasets.
3. Implement any final cleanup. This means pulling out all of our messy close, clean, deep clean, mr. clean, etc. logic and getting rid of everything that's been built up over the years.


tests/jupyter/test_pythreejs.py
```diff
@@ -202,7 +202,7 @@ def test_export_after_show():
     pl = pyvista.Plotter(shape=(2, 2))
 
     # deleting rather than showing to save time
```
adeak, line 204: Comment seems stale now.

pyvista/plotting/plotting.py
```diff
@@ -262,7 +258,7 @@ def __init__(
 
         # Add self to open plotters
         self._id_name = f"{hex(id(self))}-{len(_ALL_PLOTTERS)}"
-        _ALL_PLOTTERS[self._id_name] = self
+        _ALL_PLOTTERS[self._id_name] = weakref.ref(self)
```
adeak, line 261: I had the same idea for https://github.com/pyvista/pyvista/pull/3037 but then I realised that our plot sphinx extension relies on this to pick out plotters from, and I'm pretty sure we can't let these figures be garbage collected.

If I'm right, we could make these weakrefs real refs for `BUILDING_GALLERY`... all in the name of practicality.

akaszynski: It's quite likely this only worked because the plotter wasn't getting collected.

I think the approach of only tracking plotters when building the gallery is a better approach.

akaszynski:
Note: this PR brings us much closer to solving https://github.com/pyvista/pyvista/issues/2577.

Using the script from #2577 and 30000 calls I'm showing more or less no practical leaks.
tmp

tests/jupyter/test_pythreejs.py
akaszynski: Fixed in 132132a6053b3bfbf7e2045115a31ffedd593bf8.

banesullivan (approved):
This looks really excellent! I feel a thorough review has already taken place, so I do not want to nitpick. This is a serious improvement and is ready to merge IMO

PR #2967, opened 2022-07-10
Fix pyvista.plot kwargs

akaszynski:
At the moment, `pyvista.plot` is ignoring the global theme and using only `scooby` to determine the `notebook` `Plotter` parameter. This is an issue for Google Colab as `notebook=True` cannot be set globally and it's a bit annoying to set this on a per-plot basis. Would appreciate a quick turnaround with this PR for Scipy just in case anyone uses Colab.

Also added another example because it seemed to need it.

tmp


banesullivan (dismissed):

akaszynski:
> previously approved these changes 3 days ago

Merging as changes were approved with only a minor fix.

PR #2976, opened 2022-07-12
[BUG] Fix clearing point and cell data

alexrockhill:
Fixes failed test in https://github.com/mne-tools/mne-python/pull/10913.

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -3369,9 +3369,6 @@ class BasePlotter(PickingHelper, WidgetHelper):
         self.disable_picking()
         if hasattr(self, 'renderers'):
             self.renderers.deep_clean()
-        if getattr(self, 'mesh', None) is not None:
-            self.mesh.clear_point_data()
-            self.mesh.clear_cell_data()
         self.mesh = None
         if getattr(self, 'mapper', None) is not None:
             self.mapper.lookup_table = None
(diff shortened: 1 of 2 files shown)


akaszynski:
Thanks for adding that. Do you need a patch release for this soon?

alexrockhill:
> Thanks for adding that. Do you need a patch release for this soon?

That would be nice, I could just add a `try` `except` in MNE for now but maybe other people will run into this and so it might be better to do a patch sooner but I'm sure you would know much better than I.

pyvista/plotting/plotting.py
```diff
@@ -3370,8 +3370,8 @@ def deep_clean(self):
         if hasattr(self, 'renderers'):
             self.renderers.deep_clean()
         if getattr(self, 'mesh', None) is not None:
-            self.mesh.point_data = None
-            self.mesh.cell_data = None
+            self.mesh.clear_point_data()
+            self.mesh.clear_cell_data()
```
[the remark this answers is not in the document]
akaszynski: This was fixed in f98c3ec1d721e3f02404a2fd8aff7643c103816a.

tests/test_plotter.py
```diff
@@ -253,3 +253,15 @@ def test_add_multiple(sphere):
...
+def test_deep_clean(cube):
+    pl = pyvista.Plotter()
+    cube_orig = cube.copy()
+    pl.add_mesh(cube)
+    pl.deep_clean()
+    assert pl.mesh is None
+    assert pl.mapper is None
+    assert pl.volume is None
+    assert pl.textActor is None
+    assert cube == cube_orig
```
adeak, line 267: This might be a bit paranoid, but I've added a regression test to make sure we're not messing with the mesh. I changed the mesh fixture from the `sphere` to the `cube` because the latter has some point/cell data.

akaszynski: Good point. I agree with the changes.

adeak (approved):
This looks good now. Thanks again @alexrockhill

Issue #2962, opened 2022-07-09
pythreejs Plotter support for point labels


darikg:
### Describe the feature you would like to be added.

The pythreejs plotter currently does not support point labels. It'd be nice to have this since the pythreejs/jupyterlab integration is really nice.

I've poked around a bit and found:

1) Pyvista's pythreejs adapter works by converting the actors in the VTK plotter to their pythreejs equivalents
2) `Plotter.add_point_labels(..., show_points=True)` adds two actors: 
  - an actor whose `GetMapper().GetInputAsDataSet()` returns a PolyData with the corresponding points, and label values stored in `point_data['labels']`
  - an actor whose `GetMapper()` returns a `vtkLabelPlacementMapper` that does not have a `GetInputAsDataSet` method
3) with `Plotter.add_point_labels(..., show_points=False)`, only the latter of the two actors above is added

It seems like it'd be pretty easy to pull points and labels out of that first actor, and represent them as pythreejs `TextGeometry` objects, but if `show_points=False` is used, I don't know how to get the points/labels out of that `vtkLabelPlacementMapper`.

cc @akaszynski, I think you did all the original pythreejs work?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

akaszynski:
> It'd be nice to have this since the pythreejs/jupyterlab integration is really nice.

Fully agree. It's a missing feature I've been meaning to add.

> cc @akaszynski, I think you did all the original pythreejs work?

Yes, and it's a bit of work, not sure how much I'll be able to maintain it without help.

darikg:
> Yes, and it's a bit of work
I'm sure! I'm happy to submit a PR here, just not sure how to use the vtkLabelPlacementMapper. Or do you think it'd be ok to just require the original `add_point_labels` have `use show_points=True`?

akaszynski:
> > Yes, and it's a bit of work
> 
> I'm sure! I'm happy to submit a PR here, just not sure how to use the vtkLabelPlacementMapper. Or do you think it'd be ok to just require the original `add_point_labels` have `use show_points=True`?

I think that's acceptable. Just as long as you raise a helpful `ValueError` to let the user know what the issue is.

darikg:
Ah, looks like pythreejs hasn't implemented TextGeometry yet https://github.com/jupyter-widgets/pythreejs/issues/199

banesullivan:
[...]

PR #2977, opened 2022-07-13
[WIP] pythreejs support for point-labels

darikg:
Add support for `add_point_labels` in the pythreejs plotter adapter (#2962)

image

This uses a `pythreejs.Sprite` with a `pythreejs.TextTexture` to render text on to planes that always face the camera. The main limitation is that the text is rasterized, which means if you zoom in on the text it gets very pixelated. I also can't figure out how to set text background color, the options being a fully opaque white background or completely transparent background.

TODO:
  - [ ] Figure out if it's possible to pull out the user-specified font size, etc. properties from the VTK mapper object
  - [ ] Ditto the "always visible" argument
  - [ ] Support rendering the label location points the same way VTK does
  - [ ] Tests

### DIFF
--- a/pyvista/jupyter/pv_pythreejs.py
+++ b/pyvista/jupyter/pv_pythreejs.py
@@ -289,6 +289,31 @@ def to_surf_mesh(actor, surf, mapper, prop, add_attr=None):
     return tjs.Mesh(geometry=surf_geo, material=material)
 
 
+def point_labels_to_sprites(dataset):
+    for point, label in zip(dataset.points, dataset.point_data['labels']):
+        text = tjs.TextTexture(
+            string=label,
+            color='green',  # face color
+            size=18,
+            fontFace='Arial',
+        )
+        material = tjs.SpriteMaterial(
+            map=text,
+            sizeAttenuation=True,  # Change size when zooming in/out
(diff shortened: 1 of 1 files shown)


akaszynski:
> Hm, this is going to be harder than I thought. The major difficulty is controlling the size of the text objects. Because we're mapping a text texture onto the sprite, the text font size only really controls the resolution of the text rasterization; the onscreen text size is controlled by the sprite size, which is specified in data units. This means we would have to do something like get the bounding box limits of the scene and scale the sprite sizes by that. Furthermore, to have labels of different lengths in the same font size, the sprite widths _also_ need to be scaled by string length, which is non-trivial for non-monospaced fonts. It's all seeming pretty kludgey.

I found the same issue when implementing, and partially that's why I didn't implement it.

Curious, why are using `pythreejs` over `panel`? Since `panal` seems more maintained, do you think it might be easier to implement point-labels there?

darikg:
> Curious, why are using `pythreejs` over `panel`? Since `panal` seems more maintained, do you think it might be easier to implement point-labels there?

Huh, had no idea panel was the more modern one, I'll start playing around with it instead. Thanks! Gonna close this PR as a more-or-less dead-end

PR #2991, opened 2022-07-14
Clear charts when clearing plotters

adeak:
[This Stack Overflow question](https://stackoverflow.com/questions/72985319/pyvista-why-does-the-callback-only-draw-the-3d-entities-and-not-the-2d-one-on-t/) has an excellent minimal example wherein calling `plotter.clear()` doesn't clear the underlying `Charts` object, which prevents replotting from succeeding.

It's still not clear to me what the exact mechanics are that prevent replotting, but it's clear (no pun intended) that
1. we _should_ clear charts on `plotter.clear()`, and
2. doing so fixes the bug.

For others as unfamiliar with these execution paths as I am:
1. `pyvistaqt.QtInteractor` inherits its `.clear()` method from `BasePlotter`
2. `BasePlotter` delegates it to `plotter.renderers.clear()`
3. `Renderers` delegates it to `plotter.renderer.clear()`
4. `Renderer.clear()` is the final source: https://github.com/pyvista/pyvista/blob/8f47b9e479737dcbe1de8bba1eb12c6688b8b9ce/pyvista/plotting/renderer.py#L1862-L1876

Calling `self._charts.deep_clean()` in there will make sure to clear the internal `Charts` object (after creating one from scratch if necessary; this is the cost of not messing with `__charts` directly), then removes the `Charts`' actor from the renderer.

---

A few things stood out to me while I was investigating this:
- There's no `Charts.clear()`, perhaps `deep_clean()` should be called that instead?
- `Charts` doesn't use the renderer's `add_actor()`/`remove_actor()` methods, which doesn't seem nice. E.g. `plotter.renderer._actors` has no knowledge of any chart-related stuff this way. Unless there's a good reason for this, we should fix this (potentially in a different PR; this bugfix would be nice to have in the next patch release).

### DIFF
--- a/pyvista/plotting/charts.py
+++ b/pyvista/plotting/charts.py
@@ -4384,7 +4384,7 @@ class Charts:
     def deep_clean(self):
         """Remove all references to the chart objects and internal objects."""
         if self._scene is not None:
-            charts = [*self._charts]  # Make a copy, as this list will be modified by remove_plot
+            charts = [*self._charts]  # Make a copy, as this list will be modified by remove_chart
             for chart in charts:
                 self.remove_chart(chart)
             self._renderer.RemoveActor(self._actor)
@@ -4458,3 +4458,7 @@ class Charts:
         """Return an iterable of charts."""
         for chart in self._charts:
             yield chart
+
+    def __del__(self):
+        """Clean up before being destroyed."""
+        self.deep_clean()
--- a/pyvista/plotting/charts.py
+++ b/pyvista/plotting/charts.py
@@ -4459,6 +4459,6 @@ class Charts:
         for chart in self._charts:
             yield chart
 
-    def __del__(self):
+    def __del__(self):  # pragma: no cover
         """Clean up before being destroyed."""
         self.deep_clean()
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -1868,6 +1868,7 @@ class Renderer(_vtk.vtkOpenGLRenderer):
                 except KeyError:
                     pass
 
+        self._charts.deep_clean()
         self.remove_all_lights()
         self.RemoveAllViewProps()
         self.Modified()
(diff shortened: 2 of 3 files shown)


adeak:
Welp, I broke the labeler action again :cold_sweat:

pyvista/plotting/renderer.py
```diff
@@ -1868,6 +1868,7 @@ def clear(self):
                 except KeyError:
                     pass
 
+        self._charts.deep_clean()
```
adeak, line 1871: Accessing `self._charts` in `clear()` (even without the `deep_clean()`) leads to GC errors in `test_plotting.py`, due to some `dataset` `vtkWeakReference` not being collecting. Trying to figure this out...

adeak: I guess the issue was related to creating a new `Charts` instance via the `self._charts` property when it was `None` earlier. Adding a less elegant explicit check for `self.__charts is not None` seems to make the GC errors go away, while still fixing the original issue. I'm open to alternate suggestions though.

pyvista/plotting/charts.py
```diff
@@ -4458,3 +4458,7 @@ def __iter__(self):
         """Return an iterable of charts."""
         for chart in self._charts:
             yield chart
+
+    def __del__(self):
```
akaszynski, line 4462:
```suggestion
    def __del__(self):  # pragma: no cover
```

akaszynski (approved):
LGTM. Thanks for looking into this @adeak!

pyvista/plotting/charts.py
adeak: This actually led me down a rabbit hole that I'll open a PR about soon, but I'm still investigating (this is why I haven't yet looked at the other PRs as planned, sorry). I should've pushed this for the time being, thanks for adding it.

tkoyama010 (approved):
LGTM

PR #3001, opened 2022-07-16
Use Set method insted of On and Off method

### LINKED ISSUE #2769: Use Set method insted of On and Off method

Some code uses VTK's `*On()` and `*Off()` methods (ex: `QuadsOn()` and `QuadsOff()` in `vtkTessellatedBoxSource` class). These generally implement a method called `Set*()` (ex: `SetQuads()`), which can be refactored to avoid using if statements.

_Inspired by @adeak in https://github.com/pyvista/pyvista/pull/2767#discussion_r894997566_

### Links to source code.

You can find it by running `grep "On()" ./pyvista/*/*.py` in root path. For example:

https://github.com/pyvista/pyvista/blob/fc9cea791d3d8aa873ba3123ee9d1c0e8d065faa/pyvista/utilities/geometric_objects.py#L675-L678

### Pseudocode or Screenshots

```python 
 src.SetQuads(quads)
```

### Code of Conduct

RobPasMue:
### Overview

Resolves #2769 

### Details

Only conditional statements in which `On()` and `Off()` statements were used have been refactored.

### DIFF
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -2979,10 +2979,10 @@ class Renderer(_vtk.vtkOpenGLRenderer):
             self._legend.SetPosition(x, y)
             self._legend.SetPosition2(size[0], size[1])
 
-        if bcolor is not None:
-            self._legend.SetUseBackground(True)
-        else:
+        if bcolor is None:
             self._legend.SetUseBackground(False)
+        else:
+            self._legend.SetUseBackground(True)
             self._legend.SetBackgroundColor(Color(bcolor).float_rgb)
 
         self._legend.SetBorder(border)
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -2675,8 +2675,8 @@ class Renderer(_vtk.vtkOpenGLRenderer):
         if use_gradient:
             self.SetGradientBackground(True)
             self.SetBackground2(Color(top).float_rgb)
-         else:
-           self.SetGradientBackground(False)
+        else:
+            self.SetGradientBackground(False)
         self.Modified()
 
     def set_environment_texture(self, texture, is_srgb=False):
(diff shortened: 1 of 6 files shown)


tkoyama010 (changes requested):
Good catch. I have some suggestions. It is better to have an explicit code in "else" case.

pyvista/plotting/scalar_bars.py
```diff
@@ -411,10 +411,8 @@ def add_scalar_bar(
         if n_colors is not None:
             scalar_bar.SetMaximumNumberOfColors(n_colors)
 
-        if n_labels < 1:
-            scalar_bar.DrawTickLabelsOff()
-        else:
-            scalar_bar.DrawTickLabelsOn()
+        if not n_labels < 1:
+            scalar_bar.SetDrawTickLabels(True)
```
RobPasMue, line 415:
```suggestion
        if n_labels < 1:
            scalar_bar.SetDrawTickLabels(False)
        else:
            scalar_bar.SetDrawTickLabels(True)
```

tkoyama010 (approved):
LGTM

RobPasMue:
Batch suggestions applied are causing test failures. Looking into issue

RobPasMue:
@tkoyama010 problem solved. Ready to merge.

pyvista/plotting/renderer.py
```diff
@@ -2979,10 +2979,10 @@ def add_legend(
             self._legend.SetPosition(x, y)
             self._legend.SetPosition2(size[0], size[1])
 
-        if bcolor is None:
-            self._legend.UseBackgroundOff()
+        if bcolor is not None:
+            self._legend.SetUseBackground(True)
```
RobPasMue, line 2983:
```suggestion
        if bcolor is None:
            self._legend.SetUseBackground(False)
```

pyvista/plotting/renderer.py
```diff
@@ -2979,10 +2979,10 @@ def add_legend(
...
         else:
-            self._legend.UseBackgroundOn()
+            self._legend.SetUseBackground(False)
```
RobPasMue, line 2985:
```suggestion
            self._legend.SetUseBackground(True)
```

akaszynski (approved):
LGTM. Thanks!

PR #3014, opened 2022-07-17
Fix the flying edges example

akaszynski:
#2980 broke one of our examples. This PR fixes it in preparation for `pyvista==0.35.2` (ideally out today).

### DIFF
--- a/examples/01-filter/flying_edges.py
+++ b/examples/01-filter/flying_edges.py
@@ -108,7 +108,7 @@ pl = pv.Plotter(window_size=[800, 800], off_screen=True)
 
 pl.open_gif('barth_sextic.gif')
 
-for angle in np.linspace(0, np.pi, 20)[:-1]:
+for angle in np.linspace(0, np.pi, 20, endpoint=False):
     # clear the plotter before adding each frame's mesh
     pl.clear()
     pl.enable_lightkit()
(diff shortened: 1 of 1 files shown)


akaszynski:
This is ready for review. Docs build fine locally.

I'd like to merge this to get 0.35.2 out later today.

examples/01-filter/flying_edges.py
```diff
@@ -96,33 +96,33 @@ def barth_sextic(x, y, z):
...
-    cmap="plasma",
-    show_scalar_bar=False,
-)
 pl.open_gif('barth_sextic.gif')
 
-for angle in np.linspace(0, np.pi, 15)[:-1]:
-    new_mesh = grid.contour([angle_to_range(angle)], values, method='flying_edges')
-    mesh.overwrite(new_mesh)
-    pl.update_scalars(np.linalg.norm(new_mesh.points, axis=1), render=False)
+for angle in np.linspace(0, np.pi, 20)[:-1]:
```
adeak, line 111: If this is just here to make the animation periodic, we could use
```suggestion
for angle in np.linspace(0, np.pi, 20, endpoint=False):
```
with the subtle difference that we'd end up with 20 distinct frames insteaad of 19. But I think that would actually fit the example's description better.

adeak (approved):
One small (mostly-)style suggestion, otherwise this looks good to me and runs fine locally (not a doc build, just the example run as a script). Good indicator example ;)

adeak (approved):

adeak:
Build artifact looks good to.

PR #3015, opened 2022-07-17
Legend improvements

germa89:
### Overview
This PR aims to improve the legend plotting, especially when using legends and glyphs.

The implementation is rather simple (code example at the end of this PR).
Every time we use ``mesh.glyph``, we store the generated geometry in a private attribute (``_glyph_geom``) as ``pyvista.PolyData``.
This is later picked by the method ``add_mesh`` and added to the legend (``self._legend``).
Then when the legend is added (``add_legend``), it will make sure to use the PolyData in the legends. 

I think this is a decent way to reuse the specified glyph, but I'm open to discussion.

### Caveats
* So far, if using multiple geometries in the ``mesh.glyph``, the proposed implementation only picks the first (I'm not sure if it is needed/possible more)

* Unit testing is missing. I am still wrapping my head around pyvista unit testing.

### Details
From:
image

To:
image

#### Code example
```py
import pyvista
from pyvista import examples
mesh = examples.load_random_hills()

mesh["Normals2"] = -1*mesh["Normals"].copy()

pl = pyvista.Plotter()

arrows = mesh.glyph(scale="Normals", orient="Normals", tolerance=0.05)
actor = pl.add_mesh(arrows, color="black", label="label 1")

arrows2 = mesh.glyph(scale="Normals", orient="Normals2", tolerance=0.05)
actor = pl.add_mesh(arrows2, color="red", label="label 2")

actor = pl.add_mesh(mesh, scalars="Elevation", cmap="terrain",
                    show_scalar_bar=False)

pl.add_legend()
pl.show()
```


tkoyama010 (commented):
Great contribution. Thank you very much.
1. This change means that when a glyph is used for a plot, the legend will always be that glyph. I would like to discuss if it is better to always use the same glyph for the plot and legend or leave the option to use a different glyph. Personally, I think the glyphs in the plot and legend should be consistent (as should the colors), but let me see if anyone @pyvista/developers  has a different opinion.
1. Could you please add the test for the additional part? We aim to add tests to every line.

adeak:
> This change means that when a glyph is used for a plot, the legend will always be that glyph. I would like to discuss if it is better to always use the same glyph for the plot and legend or leave the option to use a different glyph. Personally, I think the glyphs in the plot and legend should be consistent (as should the colors), but let me see if anyone https://github.com/orgs/pyvista/teams/developers has a different opinion.

I think the default in any plotting scenario should be to have the same marker as that in the plot. However, with 3d glyphs it might easily happen that a given default heuristic doesn't work for the user. Which way does the glyph point in the legend marker? What if, as @germa89 noted, there are multiple kinds of glyphs? What if there's a single call to `glyph` but a glyph table is used, as in https://docs.pyvista.org/examples/01-filter/glyphs_table.html?

So it might be hard to come up with a good default, which means it would be ideal if it's also reasonably easy to customize the markers. (Having said all this without looking at the changes in this PR yet.)

tkoyama010:
> Which way does the glyph point in the legend marker? What if, as @germa89 noted, there are multiple kinds of glyphs? What if there's a single call to glyph but a glyph table is used, as in https://docs.pyvista.org/examples/01-filter/glyphs_table.html?

Good point. It looks like the use cases need to be organized.

germa89:
@tkoyama010  thank you for your comments. The unit tests will be added, for sure. 
Regarding using alternative glyphs, you can use the parameter ``face`` in the ``add_legend`` method. This parameter will overwrite the values set by ``mesh.glyph``. 

@adeak I like the example you posted. I wonder how the label is included in all the process. I'm thinking on the case we have multiple glyphs. Can you supply multiple labels to the parameter ``label`` in ``Plotter.add_mesh``?? If not, does it make sense to plot the legend of multiple glyphs that have only one label?

I think for that case, we should just resource to ``Plotter.add_legend``, there I would made the parameter ``face`` accept iterables of strings or Polydata. The final implementation would look like:
```py
Plotter.add_mesh(
labels=['label 1', 'label 2', 'label 3'], 
face=['triangle', 'square', pyvista.PolyData([0.0, 0.0, 0.0])])
```

Let me know your thoughts so I can proceed further.

MatthewFlamm:
I like the idea of this PR in general.

With respect to edge cases, I think it will be cumbersome to come up with some system to automatically come up with a legend marker shape, aka `face`, that accommodates all scenarios.  I think the suggestion below gives the flexibility to the user in those cases.

> I think for that case, we should just resource to `Plotter.add_legend`, there I would made the parameter `face` accept iterables of strings or Polydata. The final implementation would look like:
> 
> ```python
> Plotter.add_mesh(
> labels=['label 1', 'label 2', 'label 3'], 
> face=['triangle', 'square', pyvista.PolyData([0.0, 0.0, 0.0])])
> ```

This seems like a straightforward, flexible suggestion to me.  I'm +1 on this (probably in another PR).

MatthewFlamm:
> What if there's a single call to glyph but a glyph table is used [...]?

For glyph tables, I'd propose we use the first glyph in the table by default.

> Which way does the glyph point in the legend marker?

This is noted in the docstring of `add_legend`: "A custom face can be created using [pyvista.PolyData](https://docs.pyvista.org/api/core/_autosummary/pyvista.PolyData.html#pyvista.PolyData). This will be rendered from the XY plane."  I assume this is how the glyphs themselves work and is self-consistent, but I haven't delved into the code too deeply.

akaszynski (approved):
Earth was missing, added that.

banesullivan:
@germa89, this is a great addition... sorry this fell off our radars. I'm curious if you'd like to re-open this or if we should close now that it is significantly stale?

germa89:
> @germa89, this is a great addition... sorry this fell off our radars. I'm curious if you'd like to re-open this or if we should close now that it is significantly stale?

I am working on this. It is taking time for me to remember about this PR details, but...

germa89:
> One final question (sorry, I still haven't looked at the PR's code): what if the glyphs are coloured according to scalars, rather than a single colour? Is the behaviour (i.e colour used in the face) well-defined and documented for this case?

@adeak I would say that if there are multiple glyphs (different colours, shapes, etc), Pyvista by default pics the first one as @MatthewFlamm already mentioned:

> For glyph tables, I'd propose we use the first glyph in the table by default.

If that approach does not fit all user cases, we can always use ``add_legend`` as:

```py
pl.add_legend(labels = [
    ["label1", "red", pv.PolyData([1.0, 0, 0])],
    ["label1", "red", pv.PolyData([0.0, 0.1, 0])],
])
```
I am making that ``add_legend`` ``labels`` argument can accept an iterable or a dict (keys: 'text', 'color', 'face'), we can do the above.

**However** it seems we cannot set the symbol color apart... since ``vtkLegendBoxActor::SetEntrySymbol`` accepts a ``vtkPolyData`` but as far as I know, these objects do not container ``color``. It is normally given on the ``add_mesh`` method.

If anyone knows a way to tell the underlying ``vtkmodules`` that this ``vtkPolyData`` must be plotted with a given color, we can then plot the symbols in the legend in different color than the legend text itself. I must say it is desirable that both colors (text and symbol) are sync, but as we mentioned, there might be some cases where we need to customize.

pyvista/core/filters/data_set.py
```diff
@@ -2376,7 +2378,13 @@ def glyph(
         alg.SetScaleFactor(factor)
         alg.SetClamping(clamping)
         _update_alg(alg, progress_bar, 'Computing Glyphs')
-        return _get_output(alg)
+
+        output = _get_output(alg)
+        # breakpoint()
```
germa89, line 2383:
```suggestion
```

pyvista/plotting/renderer.py
```diff
@@ -3278,12 +3278,12 @@ def height(self):  # numpydoc ignore=RT01
...
         self,
         labels=None,
-        bcolor=(0.5, 0.5, 0.5),
+        bcolor=None,
         border=False,
         size=(0.2, 0.2),
         name=None,
         loc='upper right',
-        face='triangle',
+        face="triangle",
```
germa89, line 3286:
```suggestion
        face='triangle',
```

germa89:
Some test are still running, but I think this is ready for review.

germa89:
Do I need to add the image cache manually??

tests/plotting/image_cache/plot_composite_many_options.png
banesullivan, line 1: Will the image comparison pickup on changes to the legend? Can we make a really big legend with different geometries so that the test captures whether this is working?

banesullivan (commented):
I took a cursory look at this and overall it looks excellent, however, I don't have the bandwidth for a full review at this time

tests/plotting/image_cache/plot_composite_many_options.png
germa89: Yeah. I could try to do that. Change the legend box size to something huge.

germa89: Done. I used ``size`` argument to have bigger legends.

MatthewFlamm: I'm confused why this test image needs to be changed.  The background of the legend doesn't look to be touched by this PR. Background opacity is being added in #5509.

tests/plotting/test_renderer.py
```diff
@@ -176,4 +177,102 @@ def test_add_remove_legend(sphere):
...
+    arrows = random_hills.glyph(scale="Normals", orient="Normals", tolerance=0.05)
+    pl.add_mesh(arrows, color="black", label="label 1")
+
+    arrows2 = random_hills.glyph(scale="Normals", orient="Normals2", tolerance=0.05)
+    pl.add_mesh(arrows2, color="red", label="label 2")
+
+    pl.add_mesh(random_hills, scalars="Elevation", cmap="terrain", show_scalar_bar=False)
+
+    pl.add_legend(size=(0.5, 0.5))
+    pl.show()
```
MatthewFlamm, line 212: Should we have some tests that have image regression tests?  You would need to either put the test into `tests/plotting/test_plotting.py` where we automatically use the image regression test, or you need to `opt-in` through using `verify_image_cache` fixture for each test here.

MatthewFlamm (commented):
I need to digest the list and dict options a bit, but some comments to start on testing.

tests/plotting/test_renderer.py
germa89: I don't mind either way.... for the moment I am going to pass the fixture.

germa89: I added the new image cache to my last commit

tests/plotting/image_cache/plot_composite_many_options.png
germa89: In this PR I set the background color to ``None``, which basically means there is not legend background color. I personally find it more aesthetic. But probably this needs an agreement from the maintainers.

germa89:
@banesullivan @MatthewFlamm @tkoyama010 @akaszynski  any updates on this one?

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):
Sorry for late reply. LGTM.

tkoyama010 (approved):

PR #3020, opened 2022-07-17
Use has_module

akaszynski:
In `colors.py` we check for the installation of `colorcet`, `matplotlib`, and `cmocean`. Assuming that the user does not have at least one of those modules installed, there's a signifiant cost to check if a module is installed:

```py
def try_import():
    try:
        import not_a_module
        return True
    except:
        pass

```

```py
>>> timeit try_import()
82.6 µs ± 87.4 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

This PR proposes the use of caching and `importlib` to check if a libary can be imported.

```py
import importlib
from functools import lru_cache


@lru_cache(maxsize=None)
def has_module(module_name):
    """Return if a module can be imported."""
    module_spec = importlib.util.find_spec(module_name)
    return module_spec is not None
```

```py
>>> timeit try_import('not_a_module')
43.7 ns ± 0.164 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

### DIFF
--- a/pyvista/plotting/colors.py
+++ b/pyvista/plotting/colors.py
@@ -178,6 +178,7 @@ import pyvista
 from pyvista import _vtk
 from pyvista._typing import color_like
 from pyvista.utilities import PyvistaDeprecationWarning
+from pyvista.utilities.misc import has_module
 
 IPYGANY_MAP = {
     'reds': 'Reds',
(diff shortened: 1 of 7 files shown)


pyvista/plotting/colors.py
```diff
@@ -945,43 +946,49 @@ def string_to_rgb(string):  # pragma: no cover
...
 
     elif isinstance(cmap, list):
         for item in cmap:
             if not isinstance(item, str):
                 raise TypeError('When inputting a list as a cmap, each item should be a string.')
+
+        if not has_module('matplotlib'):
+            raise ImportError(
+                'The use of custom colormaps requires the installation of matplotlib.'
+            )
```
akaszynski, line 990:
```suggestion
            )  # pragma: no cover
```

pyvista/plotting/colors.py
```diff
@@ -945,43 +946,49 @@ def string_to_rgb(string):  # pragma: no cover
...
+            try:
+                return getattr(cmocean.cm, cmap)
+            except AttributeError:
+                pass
+
         # Else use Matplotlib
+        if not has_module('matplotlib'):
+            raise ImportError(
+                'The use of custom colormaps requires the installation of matplotlib.'
+            )
```
akaszynski, line 976:
```suggestion
            )  # pragma: no cover
```

tkoyama010 (approved):
LGTM. Good refactoring.

MatthewFlamm:
This makes sense to me as well in the cases where one is not available.  But what about the case where they are all available?  We still `import X` each call to `get_cmap_safe`, but maybe the performance hit is negligible in this case.  If it is negligible, then disregard the following statement.  Could we move these checks outside the `get_cmap_safe` function and do the `import X` only once there?

adeak:
> But what about the case where they are all available? We still `import X` each call to `get_cmap_safe`, but maybe the performance hit is negligible in this case.

Indeed it is. Once a module is imported in a given Python session, it is stored in `sys.modules`, and importing it again is as cheap as looking up a key in a global dict. It's practically free. You can test this by seeing how long `import vtk` takes, and then doing that a second time.

PR #3021, opened 2022-07-17
add bool scalars test

akaszynski:
Improve our semi-functional test coverage by adding a bool scalars test.


MatthewFlamm:
I don't use bool scalars, but is it a "feature" that point scalars would be plotted this way?  This looks like it is plotted via cell data. This is bool data so interpolation in the cell probably doesn't happen.  I'm most confused that every other point is True.  Is it some quirk of geometrical meshing that every other stripe has a majority of points that are True?  Or does vtk use some other, more brittle method, to decide which color for which cell if there are equal numbers of True points?  For example, it might use the cell_id in some fashion.  If this latter is true, I'd suggest a more robust test as any subtle change will change this plot.

MatthewFlamm:
It turns out it does interpolate, but the interpolation is still a boolean value which is cool. The points are striped.  This makes sense to me.

image

MatthewFlamm (approved):

PR #3022, opened 2022-07-18
do not reset camera unless plotter has been shown

akaszynski:
## Overview

In `add_mesh` we recalculate the camera bounds each and every time we add a mesh. This greatly increases the amount of time it takes to add many meshes and slows down the plotting significantly.

This PR proposes a simple fix to correct this behavior by moving out the reset cliping range to `reset_camera`.

It also includes a minor performance improvement by not checking if an actor exists when adding a new actor with `add_mesh`. Actors are guaranteed to be new actors as we always create a new mesh, even if it may only be a shallow one.

## Timings and code

```py
import pyvista as pv


cubes = [pv.Cube(center=(i, i, i)) for i in range(1000)]

def many_cubes():
    pl = pv.Plotter(off_screen=True)
    for cube in cubes:
        pl.add_mesh(cube)
    pl.show()
```

### This Branch

```
>>> timeit many_cubes()
501 ms ± 2.83 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### `main`

```
>>> timeit many_cubes()
1.45 s ± 24.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```


MatthewFlamm:
I haven't read the code changes in detail yet, so reacting mostly to the description.  It sounds like the camera will no longer reset everytime a mesh is added, but it will be reset when `plotter.show()` is called.  This behavior should be added to the docstring for arg `reset_camera`, where now it is unclear what the current behavior is.

In case we prefer the current behavior for adding single meshes one at a time, we could accept a list of meshes. For a list of meshes, only update the camera after the last mesh is added.  This might also open up the possibility of getting further efficiencies when using the same set of args for each mesh.

akaszynski:
To dos:
- Update `reset_camera` docs
- Use do not remove existing actor if name is unset in `add_mesh`

banesullivan (approved):
LGTM

banesullivan:
This is a substantial performance increase - good catch

MatthewFlamm (approved):
LGTM too.  I have a slightly better idea of what `_first_time` represents, but it isn't crystal clear to me still.  The comment

https://github.com/pyvista/pyvista/blob/d3e01eecb4622ecb57ad6e6c560955a6a2c69461/pyvista/plotting/plotting.py#L245-L246

indicates it is about the camera. Based on this PR and some minor digging, I think it has more to do with the first render than the camera.  I would think this variable would be better named `_before_first_render`, or something like this.

This is outside this PR.

PR #3023, opened 2022-07-18
Improve make chart tables for docs

akaszynski:
This PR makes the Chart table documentation build step avoid writing new figures and files if there is no new content to write or update. This makes it possible to avoid regenerating chart table files which means we can run our documentation build and have it only update the files that actually need updating:

```
$ make -C doc html
...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 0 changed, 0 removed
```


akaszynski:
@dcbr, could you please give this a quick review? I've added you as a "contributor", you've added a ton to PyVista and we should have added you earlier.

dcbr:
Thanks! This does look good to me, though I haven't been able to rebuild the docs locally as I'm on holiday atm.

dcbr (approved):

akaszynski:
Built docs can be downloaded here:
https://github.com/pyvista/pyvista/suites/7396667533/artifacts/301302840

Issue #1507, opened 2021-07-09
Make the MultiBlock API saner


adeak:
This is just to track our discussion in https://github.com/pyvista/pyvista/pull/1482#discussion_r664009578:

> This is a pretty confused API :(
> 
>   1.  `blocks.keys()` is a thing, and instantiating from a list comes up with string keys, so it looks like a dict.
>   2.  Iterating over the container, however, yields values, which is not like a dict.
>   3.  There's no `blocks.values()` nor `blocks.items()`...
>   4.  `blocks[i]` supports both keys and integer indices?!

Additionally, I've also noticed that `MultiBlock__setitem__` does this:
```py
import pyvista as pv

blocks = pv.MultiBlock()
blocks[3, 'name'] = pv.Sphere()

blocks.keys()
# [None, None, None, 'name']
```

Making a 2-tuple passed to `__setitem__` act neither like a list nor like a dict is especially problematic from an API design standpoint.

I'm still not sure if we can change any of this without breaking too much downstream code, but at least we can track this future work on this issue.

adeak:
Perhaps deprecate this too if we touch this: https://github.com/pyvista/pyvista/blob/e79c03369dcaf57292ae6b7b366410d82debcee7/pyvista/core/composite.py#L434-L443

If we handle the MutiBlock as an iterator then it should just rely on `next(multi)` rather than exposing a `multi.next()` that's synonymous. Having a `.next()` method on an iterator is a python2ism.

akaszynski:
Agreed.  Deprecate `next`.

Would it make sense to have `MultiBlock` look like an OrderedDict with bonus pyvista features?

adeak:
Hmm. It should probably be more like a SortedDict (which is not to say that I know of an implementation of such). We want users to be able to insert in the middle of defined keys, which is different from the behaviour of an OrderedDict (or python 3.7+ dict). Perhaps an OrderedDict plus a private list of names can work. That would have the added benefit that `multi.clean` would just change that one list without touching any of the data.

But I strongly feel that it should be some kind of dict, with list-like bells and whistles added. We'd have to think hard about making the above insertion sane before committing to a new, similarly weird API :) We should start by agreeing on a set of behaviours that we think is consistent (and pythonic). I'll think about this for a bit.

MatthewFlamm:
[...]

PR #3031, opened 2022-07-19
Make MultiBlock meshes MutableSequence with some dict-like features for setting and getting.

### LINKED ISSUE #1823: Setting last item on MultiBlock appends and slicing for `__setitem__` fails
```py
import pyvista
meshes = [pyvista.Sphere() for _ in range(10)]
blocks = pyvista.MultiBlock(meshes)
len(blocks)  # 10
blocks[-1] = pyvista.Cylinder() # Expect to overwrite last mesh
len(blocks) # 11
blocks[-1] = blocks[-1]  # Should do nothing
len(blocks) # 12
```

MatthewFlamm:
### Breaking Change
MultiBlock getting and setting behavior now acts more like a list with ability to use some dict-like features for convenience.  Non-conforming usages were removed.

The following examples will no longer be possible:

- `multi[-1] = data` # appends data
- `multi = MultiBlock(); multi[5] = data` # sets length with an out of bound index
- `multi[1, 'name'] = data` # sets key of data using second item of tuple index in __setitem__
- `multi[['name1', 'name2']]` # use lists of str instead of slices
- `multi[[0, 1]]` # use lists of int instead of slices
- `multi.next()`  # get next block

The following examples should be used instead:

- `multi[-1] = data` # access the last data entry.
- `multi.append(data, 'name')` is the preferred method for appending and now accepts an optional name.
- `multi['name'] = data` will also append if 'name' doesn't yet exist like a dict.  It will overwrite the block 'name' if it exists.
- `multi = MultiBlock(); multi.n_blocks = 6, multi[5] = data`  # don't allow out of bounds indices
- `multi[1] = data; multi.set_block_name(1, 'name')`  # Use two lines to set name. Prefer to use `append`.
- ```
  # need to loop over keys names
  for name in ['name1', 'name2']:
      multi[name]
   ```
- `multi[0:1] = [data1, data2]` # now slicing for setting is possible
- `next(multi)`  # get next block
- `multi.replace(1, data)`  # replace data, but preserve key name

### Overview

This partially addresses #1507 and closes #1823.  This PR proposes that MultiBlock meshes should act like lists, but we also allow dict like get/set behavior, and some additional feature sugar ontop.


### Details
Additional behaviors like iter, next, etc. were not addressed in this PR.

### TODO:

- [x] more testing coverage for unaccepted usage, especially for existing usages that are no longer supported
- [x] if a tuple `('name', data)` is preferred, decide what order should it be in
- [x] check other methods related to getter/setter, e.g. `get`, to more closely follow list-like, then dict-like behavior.


akaszynski:
> This will be a breaking change that will be hard to deprecate. Should we start having a breaking change section in the PR description? It would be nice if we could somehow automatically label them as such and then organize the breaking changes in the release notes.

That's a great idea. There are a few other breaking changes I'd like to incorporate this release as well.

MatthewFlamm:
I considered adding an `update` method like a dict, but it is unclear what we should do in this case when a duplicate key is provided.  Should it behave like a dict, since it is a dict-like method, and replace the value?  Or should it behave like a list and append?  In my proposal, we treat it like a list, but this would be confusing in a dict-like method.  I think the better option is to not include the method in the first place.

pyvista/core/composite.py
```diff
@@ -292,71 +290,94 @@ def __getitem__(self, index: Union[int, str]) -> Optional['MultiBlock']:
...
         if not pyvista.is_pyvista_dataset(dataset):
             dataset = pyvista.wrap(dataset)
+        self.n_blocks += 1
         self[index] = dataset
+        # No overwrite if name is None
+        self.set_block_name(index, name)
 
-    def get(self, index: Union[int, str]) -> Optional['MultiBlock']:
+    def get(self, index: str, default: Optional[DataSet] = None) -> Optional[DataSet]:
         """Get a block by its index or name.
```
akaszynski, line 346:
```suggestion
        """Get a block by its name.
```

akaszynski:
> but our `append` method now has the signature `append(value, key)`.

I think keeping `append` is necessary to make this change at least mostly backwards compatible.

I'm mostly happy that we've eliminated this behavior `blocks[3, 'name'] = pv.Sphere()`

> Allow a tuple for setting name.

This should be so rare as to almost not be needed. I agree with this implementation and you can check this box off.

tests/test_composite.py
```diff
@@ -123,7 +123,7 @@ def test_multi_block_set_get_ers():
     assert multi.n_blocks == 6  # Check pyvista side registered it
     # Add data to the MultiBlock
     data = ex.load_rectilinear()
-    multi[1, 'rect'] = data
+    multi[1] = ('rect', data)
```
akaszynski, line 126: This still seems weird to me, but I don’t know how to address this unless we make MultiBlock either behave as a dict or a list. For the time being this works, and we may not be able to come up with a better solution.

akaszynski (dismissed):
One minor comment, but this is a vast improvement over the status quo, so I’d say this is good to go.

tests/test_composite.py
MatthewFlamm: I'm okay getting rid of this entirely.  I think 95% or more use cases will use `append` which now has a nice API.  This use case is for when someone wants to overwrite in the middle of the multiblock, and set the name simultaneously.  Without this functionality, it only takes one more line to do this, i.e.:

```py
multi[2] = mesh
multi.set_block_name(2, "name")
```

akaszynski: I’d prefer this. From my brief usage of multiblock, I’d agree that most of the time you’re either appending or overwriting.

MatthewFlamm: I took 'this' as removing the tuple assignment.  This feels much better to me, and cements the API as being list-like, but with dict-bells-and-whistles when possible.

adeak:
Since this _is_ a moderately severe breaking change, should we explain the new, more logical design somewhere at e.g. https://dev.pyvista.org/api/core/composite.html? I'm not normally a `MultiBlock` user, but the API was always hazy to me (partly my own fault, but partly due to the design). If someone downstream finds that their multiblock code is broken, it would be nice to point them to a clear demonstration of how the new `MultiBlock` API is meant to be used (emphasizing the listlikeness, pointing out the few dict bells and whistles).

MatthewFlamm:
If we prefer to keep the tuple assignment, we can revert https://github.com/pyvista/pyvista/pull/3031/commits/b853a4c87be5eb141e0c8d8af6b6a4e8b469514e
and https://github.com/pyvista/pyvista/pull/3031/commits/022907310ebb10c00c054b7fbad73bc0c47a4d2b

adeak:
Oh, and we'd definitely want `versionchanged` directives as a stopgap if someone wants to go back to the (most recent) old behaviour until they update their code. Probably enough to put that in the class docstring, that's where I'd be looking if I saw that a method was suddenly gone.

doc/api/core/composite.rst
```diff
@@ -4,29 +4,38 @@ Composite Datasets
...
    Add a dataset to the collection
 
-   >>> blocks.append(pyvista.Sphere())
+   >>> blocks.append(pv.Sphere())
 
-   Or add a named block
+   Add a named block and access it by name like a dict
 
-   >>> blocks["cube"] = pyvista.Cube(center=(0, 0, -1))
+   >>> blocks.append(pv.Cube(center=(0, 0, -1), "cube")
```
MatthewFlamm, line 24:
```suggestion
   >>> blocks.append(pv.Cube(center=(0, 0, -1)), "cube")
```

akaszynski:
> 

[PyVista API Reference - Composite Datasets](https://dev.pyvista.org/api/core/composite.html) is lacking anyway right. @MatthewFlamm, would you like to summarize the new API? I can add it, but I'm bogged down with #2941.

MatthewFlamm:
@MatthewFlamm, would you like to summarize the new API? I can add it, but I'm bogged down with https://github.com/pyvista/pyvista/pull/2941

Yes, I thought of a few more things that need to be added too.  We should support negative indices in methods like `set_block_name`.  I will set to draft until it is ready again.

adeak:
@MatthewFlamm regarding indexing, I was wondering if we could both generalise some functionality and offload some of the work by doing things like
```py
    raw_index = -2  # input
    index = range(self.n_blocks)[raw_index]  # and catch if this raises
```

This would only be slightly helpful for scalar indices (it still removes some potential off-by-one errors), but we could also implement slicing in `__getitem__`, because ranges naturally support that.

MatthewFlamm:
> I was wondering if we could both generalise some functionality and offload some of the work by doing things like

I'm all for offloading logic.  I did some testing and `range(self.n_blocks)[index]` is marginally slower than two sequential if statements for a single index.  In either case, it is very fast to likely not matter.

I'm thinking whether it would be useful for `MultiBlock` to be a formal `collections.abc.Sequence`?  Unfortunately, I don't think it can be a formal `MutableSequence` because it doesn't support `insert` easily without resetting all the block data.  If we use `Sequence`, we we would allow the below, since I think using a slice and a Sequence in `__setitem__` should result in this behavior.

```py
block = pv.MultiBlock([mesh1, mesh2])
block2 = pv.MultiBlock([pv.Sphere() for _ in range(10)])
block2[0:2] = block  # sets the first two indices to be `mesh1` and `mesh2`
```

adeak:
> > I was wondering if we could both generalise some functionality and offload some of the work by doing things like
> 
> I'm all for offloading logic. I did some testing and `range(self.n_blocks)[index]` is marginally slower than two sequential `if` statements for a single index. In either case, it is very fast to likely not matter.

Yeah, that's not surprising. `range` has overhead from figuring out the type that is passed, for instance. I would expect the logic reduction to outweight the performance concerns here (as you also hinted).

> I'm thinking whether it would be useful for `MultiBlock` to be a formal `collections.abc.Sequence`? Unfortunately, I don't think it can be a formal `MutableSequence` because it doesn't support `insert` easily without resetting all the block data.

Hmm, while I don't see the immediate use, it definitely fits the bill, or if it doesn't it should (`__contains__`, `index`, `count` for example might be suspect, but we'd have to check which one of these would be generated automatically by the data model without a performance hit). Perhaps some IDE bells and whistles can hint to users that it's iterable, which might not be trivial.

> If we use `Sequence`, we we would allow the below, since I think using a slice and a Sequence in `__setitem__` should result in this behavior.
> 
> ```python
> block = pv.MultiBlock([mesh1, mesh2])
> block2 = pv.MultiBlock([pv.Sphere() for _ in range(10)])
> block2[0:2] = block  # sets the first two indices to be `mesh1` and `mesh2`
> ```

I'm not sure we _have_ to allow that. the ABC only specifies the methods, but not the exact semantics. For instance lists don't require the replacement to be the same length, in fact any iterable can be used to replace a given slice in `__setitem__`:
```py
>>> lst = list('abcdef')
... lst[2:4] = 'potato'
... lst
['a', 'b', 'p', 'o', 't', 'a', 't', 'o', 'e', 'f']
```
In contrast, `collections.deque` is a `MutableSequence`, and it doesn't even support slices in `__getitem__` (because it's basically a linked list).

So I think we'd have leeway in deciding what makes sense for us here. We could probably even get away with allowing slices in `__getitem__()` (e.g. grab the first 5 meshes for debugging) and forbidding them in `__setitem__()` (how often do you want to replace parts of a `MultiBlock`?).

MatthewFlamm:
I think the biggest upside is

> __contains__, index, count for example might be suspect, but we'd have to check which one of these would be generated automatically by the data model without a performance hit).

I haven't confirmed this, but I _think_ these will default to `__eq__` checks on items. This how we'd have to implement too.  We need testing on this, but I'm hoping that this isn't too slow if it can short circuit early, but this comes down to `numpy.array_equal` (see [NumPy discussion](https://github.com/numpy/numpy/issues/6909)  Most of the time the datasets are very different.  If vtk doesn't loop points to calculate ranges, this would be another good point short-circuit for most cases.  The pathological case is one in which one data array has only one value that has changed that is neither min or max.

MatthewFlamm:
> I'm not sure we have to allow that. the ABC only specifies the methods, but not the exact semantics. 

Up for discussion, I think if we formally declare it to be a sequence then we should treat it like other sequences.  In your example:

> For instance lists don't require the replacement to be the same length, in fact any iterable can be used to replace a given slice in __setitem__:

You say _any iterable_ will have this behavior.  We will either have to remove slicing entirely, which I don't like, or support all Sequences to be applied on slices, including MultiBlock.  If we specifically rule out MultiBlock, this would be an indication we should not formally inherit from `Sequence`.

adeak:
(I'll have to investigate things before I can reply to your first comment about `__contains__` and stuff.)

> > I'm not sure we have to allow that. the ABC only specifies the methods, but not the exact semantics.
> 
> Up for discussion, I think if we formally declare it to be a sequence then we should treat it like other sequences. In your example:
> 
> > For instance lists don't require the replacement to be the same length, in fact any iterable can be used to replace a given slice in **setitem**:
> 
> You say _any iterable_ will have this behavior. We will either ~have to~ should remove slicing entirely, which I don't like, or support all Sequences to be applied on slices, including MultiBlock. If we specifically rule out MultiBlock, this would be an indication we should not formally inherit from `Sequence`.

I think you may have misunderstood me on multiple points.

1. The "iterable" remark was that lists allow any iterable to be _on the right-hand side_ of a slicing assignment, as in my example. I'm mostly concerned about multiblocks _on the left-hand side_ of a slicing assignment. If we allow multiblock slices to be replaced as `multi[2:4] = ...` it would make sense to allow any _iterable of meshes_ on the right-hand side, including other `MultiBlock` instances. I haven't given this too much thought though so if there's some inconsistency here, it's on me.
2. My earlier examples were meant to prove that there's no "this is how `Sequence` behaves" in this regard. Since `lst[2:4] = 'potato'` works for a `list` (this is a `MutableSequence`, subtype of `Sequence`), but `d[2:4] = 'potato'` doesn't work for a `deque` (this is _also_ a `MutableSequence`), because not even `print(d[2:4])` works for a `deque`.

So I still think that we can design the exact slicing behaviour of `__setitem__` in a way we want. My knee-jerk assessment is that allowing iterables of meshes on the right-hand side would be most _`list`-like_, but we might run into technical difficulties around `insert` which you mentioned: this would have to let the size of the `MultiGrid` change in random places. So we probably _don't_ want to do that, but we could restrict the behaviour for equal-length iterables on the right-hand side. (I don't have a standard library example for this, but I think it could make practical sense.)

_Or_ we could go whole hog on "not a _mutable_ sequence" and restrict `__setitem__` to its bare minimum: setting individual (scalar) indices.

MatthewFlamm:
> 1. it would make sense to allow any _iterable of meshes_ on the right-hand side,

This is what I was referring to, but wasnt precise about it.  You bring up a good point that the current implementation allows any Sequence on the right hand side but maybe it should be Iterable instead.  One use case is

I understand your point 2 better now. I did not realize that some subclasses see decide to implement unusable methods for common features of the `abc`. In my view, this greatly dilutes the impact of inheriting from `Sequence` if we can't depend on features in some form.  I need to revaluate my mental model of what these `abc`'s are meant to do.

MatthewFlamm:
> My knee-jerk assessment is that allowing iterables of meshes on the right-hand side would be most _`list`-like_, but we might run into technical difficulties around `insert`

This was my fear from implementing a full `MutableSequence`. This bears some testing as, in name, a MultiBlock is mutable and is a sequence. I'm not sure what kind of inefficiencies will be incurred if we allow insert, maybe not too bad?

adeak:
> I understand your point 2 better now. I did not realize that some subclasses decide to implement unusable methods for common features of the `abc`. In my view, this greatly dilutes the impact of inheriting from `Sequence` if we can't depend on features in some form. I need to revaluate my mental model of what these `abc`'s are meant to do.

I don't think there are any "unusable" methods. It's just that the methods do different things for different subclasses. The protocol/contract/whatever only says that there _is_ a, say, `__setitem__`.



> > My knee-jerk assessment is that allowing iterables of meshes on the right-hand side would be most _`list`-like_, but we might run into technical difficulties around `insert`
> 
> This was my fear from implementing a full `MutableSequence`. This bears some testing as, in name, a MultiBlock is mutable and is a sequence. I'm not sure what kind of inefficiencies will be incurred if we allow insert, maybe not too bad?

As another example (although not stdlib), NumPy ndarrays are `Sequence`s but not `MutableSequence`s, because they don't do `insert`/`pop` etc. due to their fixed size. Yet they are mutable `Sequence`s, allow `arr[2:4] = thing`, but this will always replace those values rather than affecting the size. So we could have a similar kind of "`Sequence` by contract but happens to be mutable" design. My point, still, is that we have a lot of leeway. And still having a `Sequence` makes sense: it guarantees that `iter(multi)` and `len(multi)` and `multi[i] = j` will exist and work.

MatthewFlamm:
I had some sort of mental block on this.  Rereading everything you wrote makes it crystal clear to me.  Thanks for being patient.

I will summarize my thoughts here then:

- I feel strongly that negative indices should work like lists.
- Similarly, we should not allow setting indices that exceed the current length of the MultiBlock, use append for this.  We should also have an extend method.
- The string key API that we have is needed, but it is super inefficient.  Every lookup by string key is O(n) since we loop over all the blocks until it is found. We should reduce the reliance on this API, e.g. no longer allow a list of strings for `__getitem__`, and make this clear in the docs.
- `__contains__` and so on are useful.
- We will replace lists of int for `__getitem__` and use slices only.
- From the above, we should be able to support inheriting from `Sequence`.  I think it makes sense to do so.
- For `__setitem__`, we will remove setting string keys entirely since we are reducing reliance on this API.
- For `__setitem__`, we will allow slices since it could be useful for users who have MultiBlocks with a lot of sub-datasets, and there aren't any large downsides.

Something I'm not sure we want to support, but I'm leaning towards yes.

- To inherit from `MutableSequence`, we need have an `insert` method.  Since vtk doesn't supply this method, we'd have to implement in python.  From preliminary testing, vtk just shuffles around pointers, so the bottleneck will be looping in python.  If there is a way to do this without a huge penalty, it could be useful as well.

adeak:
> * I feel strongly that negative indices should work like lists.
> * Similarly, we should not allow setting indices that exceed the current length of the MultiBlock, use append for this. We should also have an extend method.

Agreed. If multiblocks don't have an `__add__` it might also be a good idea, making it again like `list.__add__`.

> * The string key API that we have is needed, but it is super inefficient. Every lookup by string key is O(n) since we loop over all the blocks until it is found. We should reduce the reliance on this API, e.g. no longer allow a list of strings for __getitem__, and make this clear in the docs.

An alternative is to keep an internal dict of `name -> index` pairs. Then we can still look up string keys with `O(1)` each. This doesn't mean that we have to do this. And I agree list-of-indices (int or not) shout be avoided.

> `__contains__` and so on are useful.

Yeah, but I'd still have to check how `__eq__` works for meshes. Assuming we want to keep Python's habit of being roughly equivalent to
```py
    def __contains__(self, val):
        return any(elem is val or elem == val for elem in self)
```
We might get away with only checking for exact identity (which would make things a lot faster). But if people want to compare copies they wouldn't be able to rely on this. (We had an earlier use case of someone loading in meshes from a file, and wanting to compare to existing meshes, so there might be a use case for comparing copies...) The choice is "pretty slow in most cases" vs "fast but missing a feature for a subset of users".

> * We will replace lists of int for `__getitem__` and use slices only.
> * From the above, we should be able to support inheriting from Sequence. I think it makes sense to do so.

Agreed.

> * For `__setitem__`, we will remove setting string keys entirely since we are reducing reliance on this API.

That can also be justified by string keys potentially being non-unique or missing (now that they are not dict keys). Not sure how it used to behave until now. Although it might be surprising if `__getitem__` allows strings and `__setitem__` does not.

> * For `__setitem__`, we will allow slices since it could be useful for users who have MultiBlocks with a lot of sub-datasets, and there aren't any large downsides.

OK, but the question of the exact semantics remains. Specifically, do we allow the size of the multiblock to change (like in a list)? Or enforce the same length on the right-hand side (like numpy arrays if we forget about broadcasting)? This ties in with your last point:

> * To inherit from `MutableSequence`, we need have an insert method. Since vtk doesn't supply this method, we'd have to implement in python. From preliminary testing, vtk just shuffles around pointers, so the bottleneck will be looping in python. If there is a way to do this without a huge penalty, it could be useful as well.

I expect that having `insert` (and `__delitem__`) is equivalent to allowing `__setitem__` with slices to have unequal lengths on the right-hand side. We should really only do this if there's reasonable performance, e.g. `O(len(multi))` operations that change the size in the middle. "Looping in python" _per se_ is not terrible: it should probably not be noticable. An empty loop with `10_000_000` iterations runs in 0.15 seconds on my laptop. And most multiblocks are probably not gigantic.

MatthewFlamm:
I think we, mainly myself, are getting to the last few points.  I think that we should either implement both 1) changing the size of the list via slicing and 2) insert method, or implement neither of them.

> We should really only do this if there's reasonable performance, e.g. O(len(multi)).

From my preliminary testing, the time effort of inserting objects in MultiBlock does not depend on the size of the datasets.  The limiting factor was the inefficient python code I had cooked up to quickly test.  I think we should implement this and then test the performance to make a decision.

adeak:
> I think we, mainly myself, are getting to the last few points. I think that we should either implement both 1) changing the size of the list via slicing and 2) insert method, or implement neither of them.
> 
> > We should really only do this if there's reasonable performance, e.g. O(len(multi)).
> 
> From my preliminary testing, the time effort of inserting objects in MultiBlock does not depend on the size of the datasets. The limiting factor was the inefficient python code I had cooked up to quickly test. I think we should implement this and then test the performance to make a decision.

All sounds good, thanks! I'll leave you alone now :smile:

MatthewFlamm:
> I'd still have to check how __eq__ works for meshes. Assuming we want to keep Python's habit of being roughly equivalent to
>
>     def __contains__(self, val):
>         return any(elem is val or elem == val for elem in self)
> We might get away with only checking for exact identity (which would make things a lot faster). 

This is how  dataset `__eq__` works, something like, `first check identity, then check _all_ the data`.  I have some ideas that should greatly speed this up when returning `False` by checking summary data first.  Then the only time _all_ the data is checked is when all the summary data is equal.

The Multiblock checks the equality of it's elements as you point out.  We don't need to do anything special since the elements already do the identity check.

adeak:
> The Multiblock checks the equality of it's elements as you point out. We don't need to do anything special since the elements already do the identity check.

My point was partly that if we _don't_ want to check equality (which, then, is what we do now), we'd have to use `any(elem is val for elem in self)` or something similar. Even if `mesh == other_mesh` checks identity first and then equality, we don't _have_ to do that in `__contains__` (although users would probably expect that).

Side note: if we're happy with "identity then equality" testing for `__contains__` we don't even strictly have to define it (https://docs.python.org/3/reference/datamodel.html#object.__contains__):
> For objects that don’t define [`__contains__()`](https://docs.python.org/3/reference/datamodel.html#object.__contains__), the membership test first tries iteration via [`__iter__()`](https://docs.python.org/3/reference/datamodel.html#object.__iter__), then the old sequence iteration protocol via [`__getitem__()`](https://docs.python.org/3/reference/datamodel.html#object.__getitem__), see [this section in the language reference](https://docs.python.org/3/reference/expressions.html#membership-test-details).

Although the docs also recommand that we do implement it ourselves (https://docs.python.org/3/reference/datamodel.html#emulating-container-types):
> It is recommended that both mappings and sequences implement the [`__contains__()`](https://docs.python.org/3/reference/datamodel.html#object.__contains__) method to allow efficient use of the `in` operator

Not sure there would be a huge difference in performance if we implement it ourselves.

MatthewFlamm:
This is definitely needs a critical re-review.  The biggest changes were in formally inheriting from `MutableSequence` including support `insert`.  Since we now allow MultiBlock to become fully Mutable, we have to take care to reorder the keys when we reorder the sequence.  Since this isn't supported directly in vtk, we have to override some of those methods.

I took a stab at updating the documentation with the new data model.

MatthewFlamm:
The documentation is partially failing due to the inherited methods from `MutableSequence` not following numpydoc style.  I'm not sure if there is a way to override just the docstring, or if we should exclude them from the docs, or exclude from the NumPy checks.

MatthewFlamm:
I chose to remove those mixin methods from the numpydoc check to get it to pass.  The documentation for those methods will not be in the same style in PyVista, but consistent with the standard lib documentation.  Assuming this passes the CI doc build too, it will be fully ready for review again.

MatthewFlamm:
A `replace` method was added after reviewing #2941.  This allows an easier way to overwrite the block data while preserving the key name.  This PR aims to make MultiBlock as a MutableSequence, but we also try to keep the key names ordered and preserved.  I haven't done a review of the rest of the code base to see whether we are preserving key names (another PR).

akaszynski (approved):
I've added two examples and made a few tiny changes in the docstrings for consistency, but otherwise LGTM.

PR #3048, opened 2022-07-22
Add check if the path exists before removing it.

turejaku:
### Overview

This is a fix for a minor bug that might happen when trying to delete downloads in a path that doesn't exist.

### Details

- function delete_downloads() tries to remove dirs/files in not existing directory which causes errors:
```
==================================== ERRORS ====================================
__________ ERROR collecting tests/branches/test_branch_operations.py ___________
tests/branches/test_branch_operations.py:5: in <module>
    import pyvista as pv
/opt/conda/lib/python3.7/site-packages/pyvista/__init__.py:104: in <module>
    _verify_cache_integrity()
/opt/conda/lib/python3.7/site-packages/pyvista/examples/downloads.py:64: in _verify_cache_integrity
    delete_downloads()
/opt/conda/lib/python3.7/site-packages/pyvista/examples/downloads.py:87: in delete_downloads
    shutil.rmtree(pyvista.EXAMPLES_PATH)
/opt/conda/lib/python3.7/shutil.py:498: in rmtree
    onerror(os.rmdir, path, sys.exc_info())
/opt/conda/lib/python3.7/shutil.py:496: in rmtree
    os.rmdir(path)
E   FileNotFoundError: [Errno 2] No such file or directory: '/root/.local/share/pyvista/examples'
```

### DIFF
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -84,7 +84,7 @@ def delete_downloads():
 
     """
     _check_examples_path()
-    if os.path.exists(pyvista.EXAMPLES_PATH):
+    if os.path.isdir(pyvista.EXAMPLES_PATH):
         shutil.rmtree(pyvista.EXAMPLES_PATH)
     os.makedirs(pyvista.EXAMPLES_PATH)
     return True
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -84,7 +84,8 @@ def delete_downloads():
 
     """
     _check_examples_path()
-    shutil.rmtree(pyvista.EXAMPLES_PATH)
+    if os.path.exists(pyvista.EXAMPLES_PATH):
+        shutil.rmtree(pyvista.EXAMPLES_PATH)
     os.makedirs(pyvista.EXAMPLES_PATH)
     return True
 
--- a/tests/test_downloads.py
+++ b/tests/test_downloads.py
@@ -106,3 +106,20 @@ def test_delete_downloads(tmpdir):
         assert not os.path.isfile(tmp_file)
     finally:
         pyvista.EXAMPLES_PATH = old_path
+
+
+def test_delete_downloads_does_not_exist(tmpdir):
+    # change the path so we don't delete the examples cache
+    old_path = pyvista.EXAMPLES_PATH
+    new_path = str(tmpdir.join('doesnotexist'))
+
+    try:
+        pyvista.EXAMPLES_PATH = new_path
+        assert not os.path.isdir(pyvista.EXAMPLES_PATH)
+        examples.delete_downloads()
+    except FileNotFoundError as e:
+        # Delete downloads for an empty directory should not fail. Reraise if
+        # it does
+        raise e
+    finally:
+        pyvista.EXAMPLES_PATH = old_path
--- a/tests/test_downloads.py
+++ b/tests/test_downloads.py
@@ -114,12 +114,9 @@ def test_delete_downloads_does_not_exist(tmpdir):
     new_path = str(tmpdir.join('doesnotexist'))
 
     try:
+        # delete_downloads for a missing directory should not fail.
         pyvista.EXAMPLES_PATH = new_path
         assert not os.path.isdir(pyvista.EXAMPLES_PATH)
         examples.delete_downloads()
-    except FileNotFoundError as e:
-        # Delete downloads for an empty directory should not fail. Reraise if
-        # it does
-        raise e
     finally:
         pyvista.EXAMPLES_PATH = old_path


turejaku:
Hello everyone!

This is my first contribution to this project, and although it's very minor fix, I'm very happy to help!
This fix is so small that I've decided not to update related test case, I hope that this is okay.

I'm looking forward to your feedback!

pyvista/examples/downloads.py
```diff
@@ -84,7 +84,8 @@ def delete_downloads():
 
     """
     _check_examples_path()
-    shutil.rmtree(pyvista.EXAMPLES_PATH)
+    if os.path.exists(pyvista.EXAMPLES_PATH):
+        shutil.rmtree(pyvista.EXAMPLES_PATH)
```
adeak, line 88: `shutil` says "_path_ must point to a directory". So this could raise if there's a file where the directory should be (this would be a crazy situation). We _could_ change `os.path.exists` to `os.path.isdir`, but then we wouldn't delete the file if it were there, so we'd get errors down the line anyway. At least this here would be an early place to raise in this esoteric situation. (We _could_ handle the "file is there" case, but I don't think we should.)

adeak (dismissed):
Thanks a lot for finding and fixing this! One musing comment, ultimately I think what you did is the right choice :) Looks good to me.

pyvista/examples/downloads.py
turejaku: Sure thing, changed.

adeak:
> This fix is so small that I've decided not to update related test case, I hope that this is okay.

The size of the fix could still mean that a test is useful :) But it probably wouldn't be easy to set up a test with a dummy `pyvista.EXAMPLES_PATH` to let this break. Especially since the new behaviour is "nothing happens" (although "no error" is a valid test result). And the code that first runs this is in `pyvista/__init__.py`, so it's run when pyvista is first imported.

pyvista/examples/downloads.py
adeak: OK, although my point was that this was good as is :smile:

MatthewFlamm:
I'm wondering if the check inside `_check_examples_path` is the culprit, and this check should instead go there.

pyvista/examples/downloads.py
turejaku: Okay so my colleague told me one thing:

What if we use sub-processes, and one sub-process deletes these files while the second one validates positively its existence, and then tries to remove already removed files?

I should probably add try/except block here to cover this issue (which as it turns out, might've been the root cause of our problem in the first place). Are you okay with that?

adeak:
> I'm wondering if the check inside `_check_examples_path` is the culprit, and this check should instead go there.

If memory serves (from yesterday) that only checks that a path was set. We shouldn't raise there if the directory is missing. We _could_ create the directory there instead. But that might be beyond its responsibility.

MatthewFlamm:
> If memory serves (from yesterday) that only checks that a path was set.

I think you are right.  I was thrown off by the `FileNotFoundError`, which I think is incorrectly used there.

turejaku:
I'm repeating my comment for better visibility:

Okay so my colleague told me one thing:

What if we use sub-processes, and one sub-process deletes these files while the second one validates positively its existence, and then tries to remove already removed files?

I should probably add try/except block here to cover this issue (which as it turns out, might've been the root cause of our problem in the first place). Are you okay with that?

adeak:
> > If memory serves (from yesterday) that only checks that a path was set.
> 
> I think you are right. I was thrown off by the `FileNotFoundError`, which I think is incorrectly used there.

If that's in there, it's definitely incorrect!

> What if we use sub-processes, and one sub-process deletes these files while the second one validates positively its existence, and then tries to remove already removed files, and fails?
> 
> I should probably add try/except block here to cover this issue (which as it turns out, might've been the root cause of our problem in the first place). Are you okay with that?

Using EAFP instead of LBYL is very standard for file operations, yes. But
1. I doubt processes to delete each other's pyvista examples dir, it's just that a fresh install in docker (which I know is involved here) doesn't have the examples directory to begin with, at startup.
2. You could probably just use the `ignore_errors` kwarg of `shutil.rmtree`: https://docs.python.org/3/library/shutil.html#shutil.rmtree . We'd have to check if it also works for missing directories. I didn't originally want to suggest this, because we'd normally want to see errors arising from not being able to delete regular files inside the examples directory.

MatthewFlamm:
> 2\. I didn't originally want to suggest this, because we'd normally want to see errors arising from not being able to delete regular files inside the examples directory.

Based on this reason, I would go with either the current PR state or the `try` block.

Subprocesses deleting the example directory means that we will have a LOT more problems to fix :)

adeak:
> Edit: This function could be tested as well.

@MatthewFlamm I missed your edit originally. Good point: although this code is first run on init, we could manually run it with the `tmpdir` fixture in pytest... passing both existing and non-existing directories within the temp dir. We'd just have to make sure to reset the examples path when the test is over. Do you think we should do this?

MatthewFlamm:
> @MatthewFlamm I missed your edit originally. Good point: although this code is first run on init, we could manually run it with the `tmpdir` fixture in pytest... passing both existing and non-existing directories within the temp dir. We'd just have to make sure to reset the examples path when the test is over. Do you think we should do this?

I agree that we can test the `delete_downloads` function in a folder that doesn't yet exist.  You can disregard my comment about the other function since it doesn't do what I thought

adeak:
@turejaku independent from the choice we make with the implementation, do you feel comfortable adding a test for this? It should probably go in [`test_downloads.py`](https://github.com/pyvista/pyvista/blob/main/tests/test_downloads.py), and you can use the `tmpdir` fixture as in e.g. https://github.com/pyvista/pyvista/blob/main/tests/test_downloads.py#L31, see also https://docs.pytest.org/en/stable/builtin.html

akaszynski:
Test added in d6d7e9b.

akaszynski (dismissed):
LGTM. I'm adding this to the 0.35.3 milestone. @turejaku, do you need a release soon? If not, we can accumulate a few more bugfixes for a 0.35.3 patch release.

turejaku:
Sorry for my late response. Thanks for adding tests for me. I could've added it but you were faster 😃 

@akaszynski, we can wait for bigger release, so don't worry. Thank you for taking care of these tests!

tests/test_downloads.py
```diff
@@ -106,3 +106,20 @@ def test_delete_downloads(tmpdir):
...
+    try:
+        pyvista.EXAMPLES_PATH = new_path
+        assert not os.path.isdir(pyvista.EXAMPLES_PATH)
+        examples.delete_downloads()
+    except FileNotFoundError as e:
+        # Delete downloads for an empty directory should not fail. Reraise if
+        # it does
+        raise e
+    finally:
+        pyvista.EXAMPLES_PATH = old_path
```
adeak, line 125: Other than the comment I think this is the same as just try-finally:
```suggestion
    try:
        # delete_downloads for a missing directory should not fail.
        pyvista.EXAMPLES_PATH = new_path
        assert not os.path.isdir(pyvista.EXAMPLES_PATH)
        examples.delete_downloads()
    finally:
        pyvista.EXAMPLES_PATH = old_path
```

akaszynski (approved):
LGTM. Thanks for adding this @turejaku!

adeak (approved):
LGTM too, thanks!

PR #3049, opened 2022-07-22
Fix for function from_meshio

keurfonluu:
### Overview

- Fixed: function `pyvista.from_meshio` when cells data type is `np.uint64`.

### Details

The following code raises an error:

```python
import pygmsh
import pyvista

with pygmsh.geo.Geometry() as geom:
    geom.add_polygon(
        [
            [0.0, 0.0],
            [1.0, -0.2],
            [1.1, 1.2],
            [0.1, 0.7],
        ],
        mesh_size=0.1,
    )
    mesh = geom.generate_mesh()

mesh = pyvista.from_meshio(mesh)
```

This error is due to `np.hstack` somehow returning an array of type `np.float64` when input array data type is `np.uint64`:

```python
x = np.array([[0, 1], [2, 3]], dtype=np.uint64)
y = np.hstack((np.full((len(x), 1), 0), x))
print(y.dtype)
```

Output:
```
float64
```

### DIFF
--- a/pyvista/utilities/fileio.py
+++ b/pyvista/utilities/fileio.py
@@ -445,14 +445,14 @@ def from_meshio(mesh):
 
     if _vtk.VTK9:
         grid = pyvista.UnstructuredGrid(
-            np.concatenate(cells),
+            np.concatenate(cells).astype(np.int64, copy=False),
             np.array(cell_type),
             np.array(points, np.float64),
         )
     else:
         grid = pyvista.UnstructuredGrid(
             np.array(offset),
-            np.concatenate(cells),
+            np.concatenate(cells).astype(np.int64, copy=False),
             np.array(cell_type),
             np.array(points, np.float64),
         )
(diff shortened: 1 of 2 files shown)


adeak:
Thanks for the PR and finding this!

> This error is due to `np.hstack` somehow returning an array of type `np.float64` when input array data type is `np.uint64`:

This is because of this:
```py
>>> x = np.array([[0, 1], [2, 3]], dtype=np.uint64)

>>> x.dtype, np.full((len(x), 1), 0).dtype
(dtype('uint64'), dtype('int64'))
```

`np.full` sees native python int `0` as the fill value, and creates a numpy default int (`int64` on 64-bit linux/unix, `int32` roughly everywhere else). But you can't reconcile `int64` and `uint64` values in the same int array, because they have different ranges. In these cases NumPy switches to doubles, with the caveat that one might lose precision for very large numbers:

```py
>>> (np.int_(1) + np.uint64(1)).dtype
dtype('float64')
```

I'm wondering if we should also change the `hstack` line, to always match the dtype of the input. That way we can avoid type promotion on the `np.hstack` line.

akaszynski (approved):
Thanks for adding in the 2D test @keurfonluu. This LGTM, and as a side benefit we've marginally improved coverage.

PR #3055, opened 2022-07-23
Only support ASCII characters

tkoyama010:
Co-authored-by: Alex Kaszynski <akascap @ gmail.com>

### Overview

To fix the following PR error @akaszynski fix to only support ASCII characters in pyvista.

https://github.com/pyvista/pyvista/pull/2970

 

### Details

- None

### DIFF
--- a/pyvista/utilities/helpers.py
+++ b/pyvista/utilities/helpers.py
@@ -117,6 +117,12 @@ def convert_string_array(arr, name=None):
 
     """
     if isinstance(arr, np.ndarray):
+        # VTK default fonts only support ASCII. See https://gitlab.kitware.com/vtk/vtk/-/issues/16904
+        if not str(arr).isascii():  # avoids segfault
+            raise ValueError(
+                'String array contains non-ASCII characters that are '
+                'not supported by VTK.'
+            )
         vtkarr = _vtk.vtkStringArray()
         ########### OPTIMIZE ###########
         for val in arr:
--- a/pyvista/utilities/helpers.py
+++ b/pyvista/utilities/helpers.py
@@ -118,7 +118,7 @@ def convert_string_array(arr, name=None):
     """
     if isinstance(arr, np.ndarray):
         # VTK default fonts only support ASCII. See https://gitlab.kitware.com/vtk/vtk/-/issues/16904
-        if not ''.join(arr).isascii():  # avoids segfault
+        if np.issubdtype(arr.dtype, np.str_) and not ''.join(arr).isascii():  # avoids segfault
             raise ValueError(
                 'String array contains non-ASCII characters that are not supported by VTK.'
             )
--- a/pyvista/utilities/helpers.py
+++ b/pyvista/utilities/helpers.py
@@ -118,7 +118,7 @@ def convert_string_array(arr, name=None):
     """
     if isinstance(arr, np.ndarray):
         # VTK default fonts only support ASCII. See https://gitlab.kitware.com/vtk/vtk/-/issues/16904
-        if not str(arr).isascii():  # avoids segfault
+        if not ''.join(arr).isascii():  # avoids segfault
             raise ValueError(
                 'String array contains non-ASCII characters that are not supported by VTK.'
             )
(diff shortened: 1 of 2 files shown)


pyvista/utilities/helpers.py
```diff
@@ -117,6 +117,11 @@ def convert_string_array(arr, name=None):
 
     """
     if isinstance(arr, np.ndarray):
+        # VTK default fonts only support ASCII. See https://gitlab.kitware.com/vtk/vtk/-/issues/16904
+        if not str(arr).isascii():  # avoids segfault
```
adeak, line 121: Slight nitpick here and also in the tests:
```suggestion
        if not ''.join(arr).isascii():  # avoids segfault
```

Because otherwise we're checking a stringified array:
```py
>>> arr = np.array(list('árvíztűrő tükörfúrógép'))

>>> arr
array(['á', 'r', 'v', 'í', 'z', 't', 'ű', 'r', 'ő', ' ', 't', 'ü', 'k',
       'ö', 'r', 'f', 'ú', 'r', 'ó', 'g', 'é', 'p'], dtype='<U1')

>>> str(arr)
"['á' 'r' 'v' 'í' 'z' 't' 'ű' 'r' 'ő' ' ' 't' 'ü' 'k' 'ö' 'r' 'f' 'ú' 'r'\n 'ó' 'g' 'é' 'p']"

>>> ''.join(arr)
'árvíztűrő tükörfúrógép'
```

What we have now is also correct because the fluff in the string representation is ASCII, but it's slightly cleaner to only check the contents.

adeak:
And huge thanks for pulling this out, I hope you don't mind @akaszynski.

akaszynski:
I don't mind at all!

tkoyama010:
@akaszynski Thanks. I worried about it.

tests/test_datasetattributes.py
```diff
@@ -308,6 +308,11 @@ def test_set_array_string_lists_should_equal(arr, hexbeam_field_attributes):
 @settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
 @given(arr=arrays(dtype='U', shape=10))
 def test_set_array_string_array_should_equal(arr, hexbeam_field_attributes):
+    if not str(arr).isascii():
```
akaszynski, line 311:
```suggestion
    if not ''.join(arr).isascii():
```

tests/test_datasetattributes.py
```diff
@@ -457,6 +462,11 @@ def test_active_scalars_setter(hexbeam_point_attributes):
 @settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
 @given(arr=arrays(dtype='U', shape=10))
 def test_preserve_field_data_after_extract_cells(hexbeam, arr):
+    if not str(arr).isascii():
```
akaszynski, line 465:
```suggestion
    if not ''.join(arr).isascii():
```

akaszynski (dismissed):
LGTM

adeak (dismissed):
LGTM too, thanks @akaszynski @tkoyama010 :tada:

adeak:
Test failures show object arrays. I'll try to see what that is about. At worst we could go back to the original version...

pyvista/utilities/helpers.py
```diff
@@ -117,6 +117,11 @@ def convert_string_array(arr, name=None):
 
     """
     if isinstance(arr, np.ndarray):
+        # VTK default fonts only support ASCII. See https://gitlab.kitware.com/vtk/vtk/-/issues/16904
+        if np.issubdtype(arr.dtype, np.str_) and not ''.join(arr).isascii():  # avoids segfault
```
adeak, line 121: The issue was that `convert_array()` dispatches to `convert_string_array()` for both unicode and bytes-like numpy arrays. Only the former can be joined with `''`, but the latter cannot contain non-ASCII characters by definition (I think).

There's still some weirdness: `convert_array()` converts object arrays to bytes-like numpy arrays. Our tests start from `np.arange(3).astype(object)` or something like that, which would contain ints rather than strings, but this also gets converted to strings. This is all unrelated to this fix, just wondering if everything there is as expected...

adeak (approved):

PR #3070, opened 2022-07-25
Remove Sphinx version pin to reduce breakages.

tkoyama010:
### Overview

Sphinx has many extensions, so I suggest using < to specify the version.

 
https://github.com/pyvista/pyvista/pull/3064

### Details

- None

### DIFF
--- a/requirements_docs.txt
+++ b/requirements_docs.txt
@@ -1,4 +1,4 @@
-Sphinx==4.5.0
+Sphinx<4.6.0
 cmocean==2.0
 colorcet==3.0.0
 imageio-ffmpeg==0.4.7
--- a/requirements_docs.txt
+++ b/requirements_docs.txt
@@ -1,4 +1,3 @@
-Sphinx<4.6.0
 cmocean==2.0
 colorcet==3.0.0
 imageio-ffmpeg==0.4.7


akaszynski:
Curious, what's the advantage of using this over a pinned version?

tkoyama010:
In fact, this may not be a good idea. What I would like to do is to resolve the error #3064.

requirements_docs.txt
```diff
@@ -1,4 +1,4 @@
-Sphinx==4.5.0
+Sphinx<4.6.0
```
tkoyama010, line 1:
```suggestion
```

adeak:
https://github.com/pyvista/pyvista/runs/7505539693?check_suite_focus=true does look like a "dependency hell" situation. So the question is what's the _drawback_ of not pinning.

tkoyama010:
If we remove the version pin, all the latest versions will be installed. If a problem occurs, it will be unclear which version caused it. I removed Sphinx from requirements.txt because each library that has a dependency on sphinx specifies the version of sphinx that it supports, and it is obvious that the lowest version of sphinx will be installed.

adeak:
@tkoyama010 I don't think I understand that. Setting `<4.6.0` says "anything under 4.6.0 is OK for us". Removing the dependency altogether says "I don't care about the Sphinx version, anything is fine". Is this not the case? In other words, isn't removing the line even less restrictive than `<4.6.0`?

I also don't understand this part
> If we remove the version pin, all the latest versions will be installed. If a problem occurs, it will be unclear which version caused it.

Wouldn't dependabot bump that version to let us see the breakage when it's new? Or do you mean that when a Sphinx extension bumps its dependency version, it would bump our Sphinx version too, thus introducing issues?

A last option is [a compatible release specification](https://peps.python.org/pep-0440/#compatible-release), e.g. `~=4.5`, hoping that there's a nearby version that's close enough... but this probably has the same issue as `<4.6.0`.

akaszynski:
I think we should keep the sphinx pin, or at the very least `<4.6.0`. Before we had dependabot we would have issues all the time when some library upgrade broke our CI/CD and we spent hours hunting down the issue. Dependabot solves this issue. If we have too many dependencies pinned in our `requirements_docs.txt`, let's remove some of the other smaller ones that are sub-dependencies of sphinx and keep sphinx.

The `hypothesis` upgrade brings to mind how important it is to have dependabot and a bunch of requirements, so let's keep that going provided it's not a maintenance nightmare.

tkoyama010:
@adeak Thank you for asking.

> I don't think I understand that. Setting <4.6.0 says "anything under 4.6.0 is OK for us". Removing the dependency altogether says "I don't care about the Sphinx version, anything is fine". Is this not the case? In other words, isn't removing the line even less restrictive than <4.6.0?

My understanding is that <4.6.0 is the latest version possible to install in the library dependencies under 4.6.0. It is based on my impression and may need evidence.

> Wouldn't dependabot bump that version to let us see the breakage when it's new? Or do you mean that when a Sphinx extension bumps its dependency version, it would bump our Sphinx version too, thus introducing issues?

I said in the meaning of "dependabot bump that version to let us see the breakage when it's new".

tkoyama010:
@akaszynski Thank you. I may have focused too much on resolving the Sphinx dependabot error. We are able to run CI with the current version, so I think it is fine as a policy to ignore Sphinx version upgrades that cause errors.

I will close this PR if there are no other comments.

banesullivan:
This is a *tough* issue to solve... any approach is going to have downsides.

One more way to go about this is to pin all sphinx packages together at known working versions and use [dependabot-bump-together](https://github.com/marcoroth/dependabot-bump-together-action) to bump them all at once as a single PR (contributed to by yours truly 😉) where conflicts can be resolved in one place. This has saved my butt on a few other projects with similar dependency stacks

tkoyama010:
@banesullivan It sounds good. I will register it as a maintenance issue :)

PR #3072, opened 2022-07-25
Add the ability to remove environment texture

akaszynski:
Add the ability to remove an environment texture.

This is useful for QT based apps built off pyvista.


pyvista/plotting/renderer.py
```diff
@@ -2750,6 +2750,26 @@ def set_environment_texture(self, texture, is_srgb=False):
...
+        """Remove the environment texture.
+
+        Examples
+        --------
+        >>> from pyvista import examples
+        >>> import pyvista as pv
+        >>> pl = pv.Plotter(lighting=None)
+        >>> cubemap = examples.download_sky_box_cube_map()
+        >>> _ = pl.add_mesh(pv.Sphere(), pbr=True, metallic=0.9, roughness=0.4)
+        >>> pl.set_environment_texture(cubemap)
```
tkoyama010, line 2763: Can we copy and show the plotter here and then do the following? I think the before and after of the method's operation can be easily explained.

akaszynski: We don't have the ability to copy a plotter. I'd like to be able to show the plot twice with `auto_close=False`, but our scraper doesn't support that.

I think this is the best we can do at this point.

tkoyama010 (dismissed):
LGTM.Thanks.

Issue #2340, opened 2022-03-12
Poly strips are not exposed in our API


adeak:
I noticed something weird:
```py
>>> import pyvista as pv
>>> polygon = pv.Rectangle()
>>> extruded = polygon.extrude((0, 0, 1))
>>> extruded
PolyData (0x7f96e0eccee0)
  N Cells:	4
  N Points:	8
  X Bounds:	0.000e+00, 1.000e+00
  Y Bounds:	0.000e+00, 1.000e+00
  Z Bounds:	0.000e+00, 1.000e+00
  N Arrays:	0

>>> extruded.faces
array([], dtype=int64)
>>> extruded.lines
array([], dtype=int64)
>>> extruded.verts
array([], dtype=int64)
```

After extrusion the box-shaped polydata has 4 cells, but no faces/lines/verts. I would expect to have 8 triangles listed in the `.faces` property. The cell types can be queried:
```py
>>> [extruded.cell_type(n) for n in range(extruded.n_cells)]
[6, 6, 6, 6]
```
This corresponds to `VTK_TRIANGLE_STRIP` which makes sense.

Compare the poly before extrusion:
```py
>>> polygon.faces
array([4, 0, 1, 2, 3])
```

So it turns out that extrusion results in "strips" rather than "polys", and our `faces`/`lines`/`verts` properties don't catch this. Should we expose these too as a property? I don't have a use case for this (other than being surprised a bit), but it might make sense to expose this.

For reference, our implementation of `.faces` looks like this: https://github.com/pyvista/pyvista/blob/dbffcb772811f89dc3e1d2f4fab1d01ebf640d8d/pyvista/core/pointset.py#L646

We could do the same wrapping `GetStrips()`:
```py
>>> pv._vtk.vtk_to_numpy(extruded.GetStrips().GetData())
array([4, 0, 1, 4, 5, 4, 1, 2, 5, 6, 4, 2, 3, 6, 7, 4, 3, 0, 7, 4])
```

banesullivan:
+1 for exposing this. I've never had a need for triangle strips either, but I think it'd be good to expose them like we do for `lines`

akaszynski:
@AlejandroFernandezLuces, please take a look at this and see if you can expose this functionality.

Please see our documentation at https://docs.pyvista.org/ and contributing documentation at [CONTRIBUTING](https://github.com/pyvista/pyvista/blob/main/CONTRIBUTING.rst).

[...]

PR #3087, opened 2022-07-27
Don't copy the mesh if smooth_shading=True when adding it to a plotter

### LINKED ISSUE #3069: Adding a mesh to a plotter with smooth_shading=True leads to a deep copy of the mesh

Hi,

I'm not sure if this behaviour is intended or not, but when a mesh is added to a plotter, if `smooth_shading=True` then a deep copy of the mesh is created.

This is because [point normals are calculated inplace](https://github.com/pyvista/pyvista/blob/2f77c141794de181d1f9cd9a1c39e7a0321d34b5/pyvista/plotting/_plotting.py#L86-L88):
```py
mesh.compute_normals(cell_normals=False, inplace=True)
```

Perhaps calculating normals and adding point data would be preferable? Something like:
```py
normals = mesh.compute_normals(cell_normals=False, inplace=False)['Normals']
mesh.point_data.set_array(normals, 'Normals')
```

Or if this behaviour is intended, perhaps it's worth adding a note to the docstring to clarify a deep copy of the mesh will be made.

This test was added in #2888, but it fails if `smooth_shading=True` when adding the mesh to the plotter.
```py
import numpy as np
import pyvista

sphere = pyvista.Sphere()
sphere.point_data['numbers'] = np.arange(sphere.n_points)
sphere2 = sphere.copy(deep=False)

sphere['numbers'] *= -1  # sphere2 'numbers' are also modified

assert np.array_equal(sphere['numbers'], sphere2['numbers'])

plotter = pyvista.Plotter()
plotter.add_mesh(sphere, scalars=None, smooth_shading=True)

sphere[
    'numbers'
] *= -1  # sphere2 'numbers' are also modified after adding to Plotter.  (See  issue #2461)

assert np.array_equal(sphere['numbers'], sphere2['numbers'])

```

### System Information

```shell
Date: Mon Jul 25 18:06:02 2022 BST

                OS : Darwin
            CPU(s) : 20
           Machine : x86_64
      Architecture : 64bit
       Environment : Python
       GPU Version : 4.1 ATI-4.6.21

           pyvista : 0.36.dev0
               vtk : 9.0.3
             numpy : 1.22.4
           imageio : 2.19.3
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.2
         pyvistaqt : 0.10.dev0
           IPython : 8.3.0
             scipy : 1.8.1
              tqdm : 4.64.0
            meshio : 5.3.4
```

### Screenshots

_No response_

p-j-smith:
### Overview

resolves #3069 

When adding a mesh to a plotter and `smooth_shading=True`, do not create a copy of the mesh.

### Details
- If `plotter.add_mesh` is called with `smooth_shading=True`, the function `_prepare_smooth_shading` no longer calculates normals inplace, meaning a copy of the mesh is not made
- Note, as pointed out in https://github.com/pyvista/pyvista/issues/3069#issuecomment-1194409981, if `split_edges=True` then a new mesh is always returned
- Added tests to check `_prepare_smooth_shading` returns the original input mesh if `split_edges=False`


akaszynski (approved):
Nice work on this one. Agree with changes.

p-j-smith:
> Nice work on this one. Agree with changes.

Thanks for the fast review!

akaszynski (dismissed):

p-j-smith:
Hi, I'm not sure why building the docs failed (the error message is [not very helpful](https://github.com/pyvista/pyvista/runs/7542113956?check_suite_focus=true#step:14:9802)), but building the docs in [this other pr](https://github.com/pyvista/pyvista/pull/3031#issue-1309901687) seems to fail for no reason too

akaszynski:
I'll check this on my end.

adeak:
For the record the actual error (originally a warning) is here: https://github.com/pyvista/pyvista/runs/7542113956?check_suite_focus=true#step:14:3855

```
/home/runner/work/pyvista/pyvista/pyvista/jupyter/pv_pythreejs.py:436: UserWarning: Empty or unsupported dataset attached to actor
  warnings.warn('Empty or unsupported dataset attached to actor')
```

Still not terribly helpful, but at least something.

p-j-smith:
Ah, well spotted! Looks like there was a discussion raised about this recently: https://github.com/pyvista/pyvista/discussions/2747#discussion-4121408

akaszynski:
Weird, I'm getting the same warning locally:
```
make -C doc clean-except-examples phtml > /tmp/out.txt
WARNING: Cell printed to stderr:
/home/alex/python/pyvista/pyvista/jupyter/pv_pythreejs.py:436: UserWarning: Empty or unsupported dataset attached to actor
  warnings.warn('Empty or unsupported dataset attached to actor')

```

However, on `main`, this builds just fine, which is weird because I can't figure out why this PR should affect this in any way.

akaszynski:
Found out what causes this issue, and it comes down to triangle strips.

As pointed out by @adeak in #2340, triangle strips aren't exposed in our API, and I've never thought to expose them... until now.

In our example (the one failing) we have:
```
import numpy as np
import pyvista

# Make the xyz points
theta = np.linspace(-10 * np.pi, 10 * np.pi, 100)
z = np.linspace(-2, 2, 100)
r = z**2 + 1
x = r * np.sin(theta)
y = r * np.cos(theta)
points = np.column_stack((x, y, z))

spline = pyvista.Spline(points, 500).tube(radius=0.1)
```

When we get the `repr` of `spline` we get:
```
>>> spline
PolyData (0x7ff1f5376ee0)
  N Cells:	22
  N Points:	10040
  X Bounds:	-4.680e+00, 4.680e+00
  Y Bounds:	-4.344e+00, 5.097e+00
  Z Bounds:	-2.100e+00, 2.100e+00
  N Arrays:	2
```

First issue (sorta a side note), we should really be listing the number of vertices, lines, faces, and poly strips. In PolyData, not everything is a face, and I think we're not being descriptive enough with our `repr`.

Secondly, before this PR computing normals would actually expand out the triangle strips and then change the spline to:
```py
PolyData (0x7ff219463fa0)
  N Cells:	19996
  N Points:	10040
  X Bounds:	-4.680e+00, 4.680e+00
  Y Bounds:	-4.344e+00, 5.097e+00
  Z Bounds:	-2.100e+00, 2.100e+00
  N Arrays:	3
```
---
Now, the pythreejs logic has no understanding of triangle strips, so the change in this PR, which are valid for just about everything else actually cause a regression with the pythreejs rendering because it doesn't ever check if there are triangle strips.

### Solution

Expose triangle strips into our API and add that to pythreejs and then revisit this PR once that's done. We're going to have to move this PR to the `0.36.0` milestone.

p-j-smith:
Thanks for looking into it! I'll convert this to a draft for now then

banesullivan:
I want to point out that we could work around this issue by using the changes in #3318 to add normal computation as a pipelined filter. We'd still get new mesh objects and perhaps deep copies though.

banesullivan:
See #3865 and https://github.com/pyvista/pyvista/issues/3069#issuecomment-1401167712

PR #3088, opened 2022-07-27
Add copy_mesh argument to plotter.add_mesh

### LINKED ISSUE #3066: [Bug?, Discussion] Setting active scalars of a mesh is ignored by plotters

Hi,

I've run into an issue with setting the active scalars of a mesh.

If a mesh is added to a plotter and then the active scalars of the mesh are changed, the changes are not updated in the plotter.

This issue was introduced in [commit 2c52ffda4a0cf099c2c8950677492542698e4642](https://github.com/pyvista/pyvista/commit/2c52ffda4a0cf099c2c8950677492542698e4642)

```python 
import numpy as np
import pyvista as pv

def toggle_active_scalars(mesh):

    current_scalars = mesh.active_scalars_name
    new_scalars = "Positive" if current_scalars == "Negative" else "Negative"
    mesh.set_active_scalars(new_scalars)

sphere = pv.Sphere()
sphere.point_data['Positive'] = np.arange(sphere.n_points)
sphere.point_data['Negative'] = np.arange(sphere.n_points) * -1

plotter = pv.Plotter()
plotter.iren.add_observer('RightButtonPressEvent', lambda iren, event: toggle_active_scalars(mesh=sphere))
plotter.add_mesh(sphere, clim=[-sphere.n_points, sphere.n_points])

plotter.show()

 ```

### System Information

```shell
Date: Mon Jul 25 13:51:38 2022 BST

                OS : Darwin
            CPU(s) : 20
           Machine : x86_64
      Architecture : 64bit
       Environment : Python
       GPU Version : 4.1 ATI-4.6.21

           pyvista : 0.36.dev0
               vtk : 9.0.3
             numpy : 1.22.4
           imageio : 2.19.3
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.2
         pyvistaqt : 0.10.dev0
           IPython : 8.3.0
             scipy : 1.8.1
              tqdm : 4.64.0
            meshio : 5.3.4
```

### Screenshots

_No response_

### LINKED ISSUE #3079: Clip Plane Filter Doesn't Automatically Update

I am trying to apply the clip plane filter to a mesh. The visualization window appears with the original clipping applied, but when I move the plane around it fails to update.

```python 
import pyvista as pv
from pyvista import examples

vol = examples.download_brain()

p = pv.Plotter()
p.add_mesh_slice_orthogonal(vol)
p.show()

p.plane_clipped_meshes
 ```

### System Information

```shell
  Date: Tue Jul 26 09:20:19 2022 Mountain Daylight Time

                OS : Windows
            CPU(s) : 16
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.7 GiB
       Environment : Python
        GPU Vendor : Intel
       GPU Version : 4.5.0 - Build 27.20.100.9664

  Python 3.8.3 (default, Jul  2 2020, 17:30:36) [MSC v.1916 64 bit (AMD64)]

           pyvista : 0.35.2
             numpy : 1.18.5
           imageio : 2.9.0
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.2.2
             PyQt5 : 5.11.3
           IPython : 7.16.1
          colorcet : 3.0.0
           cmocean : 2.0
             scipy : 1.8.1
              tqdm : 4.64.0
        jupyterlab : 2.1.5

  Intel(R) Math Kernel Library Version 2020.0.1 Product Build 20200208 for
```

### Screenshots

image

p-j-smith:
### Overview
resolves #3079
resolves #3066

Fix use of widgets with a plotter and updating active scalars of a mesh after it has been added to a plotter.

### Details

There is a tension between:
- being able to to add a mesh to a plotter multiple times and visualise different scalars
- being able to use widgets with a plotter

The first requires that a copy of the mesh is made so that when active scalars are set on the added mesh(es), they are not changed for the original mesh.

The second requires that a copy of the mesh is **not** made, so that when changes are made to the original mesh they are updated in the plotter.

I've added a `copy_mesh` argument to `plotter.add_mesh`. By default it is set to `True`. All widgets will need to set it the `False`.

I'm not sure if there's a better solution to this? Would be happy to hear if there is!

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -1861,7 +1861,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         roughness=0.5,
         render=True,
         component=None,
-        copy_mesh=True,
+        copy_mesh=False,
         **kwargs,
     ):
         """Add any PyVista/VTK mesh or dataset that PyVista can wrap to the scene.
@@ -2135,12 +2135,12 @@ class BasePlotter(PickingHelper, WidgetHelper):
             the vector is plotted.
 
         copy_mesh : bool, optional
-            If True, a copy of the mesh will be made before adding it to the plotter.
+            If ``False``, a copy of the mesh will be made before adding it to the plotter.
             This is useful if e.g. you would like to add the same mesh to a plotter multiple
-            times and display different scalars. Setting `copy_mesh` to False is necessary
+            times and display different scalars. Setting ``copy_mesh`` to ``False`` is necessary
             if you would like to update the mesh after adding it to the plotter and have these
             updates rendered, e.g. by changing the active scalars or through an interactive widget.
-            Defaults to True.
+            Defaults to ``False``.
 
         **kwargs : dict, optional
             Optional developer keyword arguments.
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -2135,7 +2135,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
             the vector is plotted.
 
         copy_mesh : bool, optional
-            If ``False``, a copy of the mesh will be made before adding it to the plotter.
+            If ``True``, a copy of the mesh will be made before adding it to the plotter.
             This is useful if e.g. you would like to add the same mesh to a plotter multiple
             times and display different scalars. Setting ``copy_mesh`` to ``False`` is necessary
             if you would like to update the mesh after adding it to the plotter and have these
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -2203,10 +2203,10 @@ class BasePlotter(PickingHelper, WidgetHelper):
         elif isinstance(mesh, pyvista.PointSet):
             # cast to PointSet to PolyData
             mesh = mesh.cast_to_polydata(deep=False)
-        else:
-            # A shallow copy of `mesh` is here so when we set (or add) scalars
-            # active, it doesn't modify the original input mesh.
-            mesh = mesh.copy(deep=False)
+        #else:
+        #    # A shallow copy of `mesh` is here so when we set (or add) scalars
+        #    # active, it doesn't modify the original input mesh.
+        #    mesh = mesh.copy(deep=False)
 
         ##### Parse arguments to be used for all meshes #####
 
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -1861,6 +1861,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         roughness=0.5,
         render=True,
         component=None,
+        copy_mesh=True,
         **kwargs,
     ):
         """Add any PyVista/VTK mesh or dataset that PyVista can wrap to the scene.
@@ -2133,6 +2134,14 @@ class BasePlotter(PickingHelper, WidgetHelper):
             nonnegative, if supplied. If ``None``, the magnitude of
             the vector is plotted.
 
+        copy_mesh : bool, optional
+            If True, a copy of the mesh will be made before adding it to the plotter.
+            This is useful if e.g. you would like to add the same mesh to a plotter multiple
+            times and display different scalars. Setting `copy_mesh` to False is necessary
+            if you would like to update the mesh after adding it to the plotter and have these
+            updates rendered, e.g. by changing the active scalars or through an interactive widget.
+            Defaults to True.
+
         **kwargs : dict, optional
             Optional developer keyword arguments.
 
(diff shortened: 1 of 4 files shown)


p-j-smith:
So I've manually checked that the widgets work by playing around with them:
```py
import numpy as np
import pyvista

mesh = pyvista.ParametricConicSpiral()
pl = pyvista.Plotter()
pl.add_mesh_slice_spline(mesh)
pl.show()

mesh = pyvista.ParametricConicSpiral()
mesh['scalars'] = np.arange(mesh.n_points)
pl = pyvista.Plotter()
pl.add_mesh_isovalue(mesh)
pl.show()

mesh = pyvista.ParametricConicSpiral()
mesh['scalars'] = np.arange(mesh.n_points)
pl = pyvista.Plotter()
pl.add_mesh_threshold(mesh)
pl.show()

mesh = pyvista.ParametricConicSpiral()
pl = pyvista.Plotter()
pl.add_mesh_slice_orthogonal(mesh)
pl.show()

mesh = pyvista.ParametricConicSpiral()
pl = pyvista.Plotter()
pl.add_mesh_slice(mesh)
pl.show()

mesh = pyvista.ParametricConicSpiral()
pl = pyvista.Plotter()
pl.add_mesh_clip_plane(mesh)
pl.show()

mesh = pyvista.ParametricConicSpiral()
pl = pyvista.Plotter()
_ = pl.add_mesh_clip_box(mesh, color='white')
pl.show()
```

However, I'm not sure how best to write tests for these, any ideas would be appreciated!

adeak:
Thanks a lot for tracking this down and opening the PR. This was really helpful for figuring out the issue at https://github.com/pyvista/pyvista/pull/2980#issuecomment-1197355996. To reiterate my point there: we'll definitely want this, but we'll probably want to make `copy_mesh=False` the default (to avoid breaking all widget-like code).

pyvista/plotting/plotting.py
```diff
@@ -1861,6 +1861,7 @@ def add_mesh(
         roughness=0.5,
         render=True,
         component=None,
+        copy_mesh=True,
```
akaszynski, line 1864:
```suggestion
        copy_mesh=False,
```

pyvista/plotting/plotting.py
```diff
@@ -2133,6 +2134,14 @@ def add_mesh(
             nonnegative, if supplied. If ``None``, the magnitude of
             the vector is plotted.
 
+        copy_mesh : bool, optional
+            If True, a copy of the mesh will be made before adding it to the plotter.
+            This is useful if e.g. you would like to add the same mesh to a plotter multiple
+            times and display different scalars. Setting `copy_mesh` to False is necessary
+            if you would like to update the mesh after adding it to the plotter and have these
+            updates rendered, e.g. by changing the active scalars or through an interactive widget.
+            Defaults to True.
```
akaszynski, line 2143:
```suggestion
            If ``False``, a copy of the mesh will be made before adding it to the plotter.
            This is useful if e.g. you would like to add the same mesh to a plotter multiple
            times and display different scalars. Setting ``copy_mesh`` to ``False`` is necessary
            if you would like to update the mesh after adding it to the plotter and have these
            updates rendered, e.g. by changing the active scalars or through an interactive widget.
            Defaults to ``False``.
```

akaszynski:
Ironically, I just encountered a similar issue in #2941. I'll be changing that there as well.

p-j-smith:
I think it makes sense to have the default as don't copy - the only reason I set it to copy by default is so the tests added in #2980 would pass without me needing to modify them

pyvista/plotting/widgets.py
```diff
@@ -269,7 +269,7 @@ def callback(planes):
             interaction_event=interaction_event,
         )
 
-        return self.add_mesh(box_clipped_mesh, reset_camera=False, **kwargs)
+        return self.add_mesh(box_clipped_mesh, reset_camera=False, copy_mesh=False, **kwargs)
```
larsoner, line 272: You no longer need to change these lines because these are the default

larsoner (dismissed):
One tiny nitpick from me, +1 for merge with or without my suggested simplification

I checked and this fixes the problems I'm having with plotting! So also +1 for releasing a patch-bumped 0.35.3 version once merged

pyvista/plotting/widgets.py
p-j-smith: Ah yes, I've reverted the change

p-j-smith:
> I checked and this fixes the problems I'm having with plotting!
That's great!

larsoner (approved):

akaszynski:
> I checked and this fixes the problems I'm having with plotting! So also +1 for releasing a patch-bumped 0.35.3 version once merged

Merging this for immediate merge and release. Thanks all for the combined work.

PR #3101, opened 2022-07-30
Add tetra to the unstructured grid example

akaszynski:
Simply adds a tetrahedral mesh to our unstructured grid example.

image

### DIFF
--- a/examples/00-load/create-unstructured-surface.py
+++ b/examples/00-load/create-unstructured-surface.py
@@ -174,69 +174,33 @@ _ = grid.plot(show_edges=True)
 ###############################################################################
 # Tetrahedral Grid
 # ~~~~~~~~~~~~~~~~
-# Here's an unstructured tetrahedral grid.
+# Here how we can create an unstructured tetrahedral grid.
 
 # There are 10 cells here, each cell is [4, INDEX0, INDEX1, INDEX2, INDEX3]
 # where INDEX is one of the corners of the tetrahedral.
+#
+# Note that the array does not need to be shaped like this, we could have a
+# flat array, but it's easier to make out the structure of the array this way.
 cells = np.array(
     [
-        4,
-        6,
-        5,
-        8,
-        7,
-        4,
-        7,
-        3,
-        8,
-        9,
-        4,
-        7,
-        3,
-        1,
-        5,
-        4,
-        9,
-        3,
-        1,
-        7,
-        4,
-        2,
-        6,
-        5,
-        8,
-        4,
-        2,
-        6,
-        0,
-        4,
-        4,
-        6,
-        2,
-        0,
-        8,
-        4,
-        5,
-        2,
-        8,
-        3,
-        4,
-        5,
-        3,
-        8,
-        7,
-        4,
-        2,
-        6,
-        4,
-        5,
+        [4, 6, 5, 8, 7],
+        [4, 7, 3, 8, 9],
+        [4, 7, 3, 1, 5],
+        [4, 9, 3, 1, 7],
+        [4, 2, 6, 5, 8],
+        [4, 2, 6, 0, 4],
+        [4, 6, 2, 0, 8],
+        [4, 5, 2, 8, 3],
+        [4, 5, 3, 8, 7],
+        [4, 2, 6, 4, 5],
     ]
 )
 
-# This is just vtk.VTK_TETRA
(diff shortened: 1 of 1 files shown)


examples/00-load/create-unstructured-surface.py
```diff
@@ -169,3 +169,66 @@
...
+###############################################################################
+# Tetrahedral Grid
+# ~~~~~~~~~~~~~~~~
+# Here how we can create an unstructured tetrahedral grid.
+
+# There are 10 cells here, each cell is [4, INDEX0, INDEX1, INDEX2, INDEX3]
+# where INDEX is one of the corners of the tetrahedral.
+#
+# Note that the array does not need to be shaped like this, we could have a
+# flat array, but it's easier to make out the structure of the array this way.
```
tkoyama010, line 183:
```suggestion
# Tetrahedral Grid
# ~~~~~~~~~~~~~~~~
# Here is how we can create an unstructured tetrahedral grid.

# There are 10 cells here, each cell is [4, INDEX0, INDEX1, INDEX2, INDEX3]
# where INDEX is one of the corners of the tetrahedral.
#
# Note that the array does not need to be shaped like this, we could have a
# flat array, but it's easier to make out the structure of the array this way.
```

examples/00-load/create-unstructured-surface.py
```diff
@@ -169,3 +169,66 @@
...
+)
+
+# Create and plot the unstructured grid
+grid = pv.UnstructuredGrid(cells, celltypes, points)
+grid.plot(show_edges=True)
+
+
+###############################################################################
+# for fun, lets separate all the cells and plot out the individual cells. shift
+# them a little bit from the center to create an "exploded view".
```
tkoyama010, line 226:
```suggestion
# For fun, let's separate all the cells and plot out the individual cells. shift
# them a little bit from the center to create an "exploded view".
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM. Thanks!

examples/00-load/create-unstructured-surface.py
```diff
@@ -169,3 +169,66 @@
...
 _ = grid.plot(show_edges=True)
+
+
+###############################################################################
+# Tetrahedral Grid
+# ~~~~~~~~~~~~~~~~
+# Here is how we can create an unstructured tetrahedral grid.
+
+# There are 10 cells here, each cell is [4, INDEX0, INDEX1, INDEX2, INDEX3]
+# where INDEX is one of the corners of the tetrahedral.
```
adeak, line 180:
```suggestion
# where INDEX is one of the corners of the tetrahedron.
```

examples/00-load/create-unstructured-surface.py
```diff
@@ -169,3 +169,66 @@
...
+        [4, 2, 6, 5, 8],
+        [4, 2, 6, 0, 4],
+        [4, 6, 2, 0, 8],
+        [4, 5, 2, 8, 3],
+        [4, 5, 3, 8, 7],
+        [4, 2, 6, 4, 5],
+    ]
+)
+
+# 10 is just vtk.VTK_TETRA
```
adeak, line 199: So why not use `vtk.VTK_TETRA`? It might be better practice not to add magic numbers of possible.

examples/00-load/create-unstructured-surface.py
```diff
@@ -169,3 +169,66 @@
...
+    ]
+)
+
+# Create and plot the unstructured grid
+grid = pv.UnstructuredGrid(cells, celltypes, points)
+grid.plot(show_edges=True)
+
+
+###############################################################################
+# For fun, let's separate all the cells and plot out the individual cells. shift
```
adeak, line 225:
```suggestion
# For fun, let's separate all the cells and plot out the individual cells. Shift
```

examples/00-load/create-unstructured-surface.py
tkoyama010: This may be beside the point, but it may be a good idea to wrap these constants as variables in pyvsita. Importing vtk every time is not good for user convenience.

akaszynski: If only we had an enum for this...

tkoyama010: Good. I will do it.

PR #3102, opened 2022-07-30
handle verts in pythreejs

akaszynski:
Address https://github.com/pyvista/pyvista/discussions/2747#discussion-4121408 by handling vertices in `pyvista.PolyData` and `pyvista.PointSet` when using `jupyter_backend='pythreejs'`.

This issue was brought up in #3087.

tmp


akaszynski:
Merging this shortly...

pyvista/jupyter/pv_pythreejs.py
```diff
@@ -422,9 +427,15 @@ def actor_to_mesh(actor, focal_point):
 
         meshes.append(to_surf_mesh(actor, surf, mapper, prop, add_attr))
 
-    elif rep_type == 'Points':
+    if rep_type == 'Points':
```
adeak, line 430: If I understand correctly this change doesn't change anything, because `if rep_type == 'Surface' and has_faces:` and `if rep_type == 'Points':` are mutually exclusive anyway.

tests/jupyter/test_pythreejs.py
```diff
@@ -221,9 +232,8 @@ def test_labels():
     pl = pyvista.Plotter()
     pl.add_point_labels(poly, "My Labels", point_size=20, font_size=36)
 
-    # ensure that we ignore labels
-    with pytest.warns(UserWarning):
-        pv_pythreejs.convert_plotter(pl)
+    # ensure point labels at least don't raise a warning
+    pv_pythreejs.convert_plotter(pl)
```
adeak, line 236: Is this change related to the PR? And do warnings get converted to errors in CI?

adeak (commented):
I can't comfortably approve because I have no idea about pythreejs nor its pyvista machinery, but nothing major sticks out to me. Two questions/remarks.

It did cross my mind that the `poly_points = dataset.extract_cells(dataset.verts[1::2])` dance shows that it's a bit weird to have this dichotomy of "verts but not cells". But I know VTK has a separate notion of verts so it probably makes sense in the grand scheme of things.

pyvista/jupyter/pv_pythreejs.py
akaszynski: Correct. I was probably working through the `has_verts` branch and didn't revert this change.

akaszynski, line 430:
```suggestion
    elif rep_type == 'Points':
```

tests/jupyter/test_pythreejs.py
akaszynski: This is related to this PR and warnings do not get converted into errors. Fixing that...

akaszynski, line 236:
```suggestion
    with pytest.warns(None) as record:
        pv_pythreejs.convert_plotter(pl)
    assert len(record) == 0
```

tests/jupyter/test_pythreejs.py
```diff
@@ -221,9 +232,10 @@ def test_labels():
     pl = pyvista.Plotter()
     pl.add_point_labels(poly, "My Labels", point_size=20, font_size=36)
 
-    # ensure that we ignore labels
-    with pytest.warns(UserWarning):
+    # ensure point labels at least don't raise a warning
+    with pytest.warns(None) as record:
         pv_pythreejs.convert_plotter(pl)
+    assert len(record) == 0
```
adeak, line 238: Slight nitpick: `record` seems to have a `__bool__` that matches the length, so perhaps we could do
```suggestion
    assert not record
```
The current version is fine too, it's explicit.

adeak (commented):
One minor nitpick, but I don't see anything substantial (with the earlier caveat).

banesullivan (approved):
if it works, it works

 I'm not sure how well this will handle meshes with both verts and other cell types... but we don't plan on long term supporting the `pythreejs` backend, so merge away

banesullivan:
Merging this before #3690

banesullivan:
Currently unable to locally reproduce the testing issue on CI

banesullivan:
FYI, `with pytest.warns(None)` doesn't work. Reverted that and merging. It was capturing errors from all kinds of places

banesullivan (approved):

PR #3089, opened 2022-07-27
Add the Grad-Descent-Visualizer package to the external examples.

JacobBumgarner:
### Overview
As discussed in #3078, this PR adds the [grad-descent-visualizer]() package to the external examples.

Also seen in discussion #3074.

PR #3075, opened 2022-07-26
Add space cubemaps

akaszynski:
I figure that we need some cubemaps if we're going to be displaying planets in #2994.

Pinging @jorgepiloto because you love space.

Here's some totally unnecessary (but cool) pictures.

### Space Sphere

```py
import pyvista as pv
from pyvista import examples
import pyvista as pv
cubemap = examples.download_cubemap_space_16k()
pl = pv.Plotter(lighting=None)
_ = pl.add_actor(cubemap.to_skybox())
pl.set_environment_texture(cubemap, True)
pl.camera.zoom(0.4)
pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.24, metallic=1.0)
pl.show()
```

spaceball

### Space Sphere Video

```py
import pyvista as pv
from pyvista import examples
import pyvista as pv
cubemap = examples.download_cubemap_space_16k()
pl = pv.Plotter(lighting=None, off_screen=True)
_ = pl.add_actor(cubemap.to_skybox())
pl.set_environment_texture(cubemap, True)
pl.camera.zoom(0.4)
pl.open_movie('sphere.mp4', quality=5, framerate=60)
pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.24, metallic=1.0)
path = pl.generate_orbital_path(n_points=720, viewup=[0, 0, 1], shift=1.0)
pl.orbit_on_path(path, progress_bar=True, write_frames=True)
```

https://user-images.githubusercontent.com/11981631/180933317-357e8518-0df5-434c-b728-96c52d5f959b.mp4

### Space Stanford Bunny

```py
import pyvista as pv
from pyvista import examples
import pyvista as pv
cubemap = examples.download_cubemap_space_16k()
pl = pv.Plotter(lighting=None, window_size=[1920, 1080], off_screen=True)
_ = pl.add_actor(cubemap.to_skybox())
pl.set_environment_texture(cubemap, True)
pl.camera.zoom(0.4)
mesh = examples.download_bunny()
pl.add_mesh(mesh, pbr=True, roughness=0.0, metallic=0.4)
pl.camera_position = 'xy'
pl.show(screenshot='space_bunny.png')
```

space_bunny

### Space Frog (because I can)

```py
import pyvista as pv
from pyvista import examples
import pyvista as pv
cubemap = examples.download_cubemap_space_16k()
pl = pv.Plotter(lighting=None)
_ = pl.add_actor(cubemap.to_skybox())
[...]

PR #3071, opened 2022-07-25
Add park cubemap

akaszynski:
Found a nice park cubemap and thought it would be nice to add:

```py
from pyvista import examples
import pyvista as pv
pl = pv.Plotter(lighting=None)
dataset = examples.download_cubemap_park()
_ = pl.add_actor(dataset.to_skybox())
pl.set_environment_texture(dataset)
pl.camera_position = 'xy'
pl.camera.zoom(0.4)
pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.1, metallic=0.5)
pl.show()
```
tmp

PR #3053, opened 2022-07-23
Add geemap and GmshModel as external examples

tkoyama010:
### Overview

 

Add [geemap](https://geemap.org/) by @giswqs as external examples.
resolves https://github.com/pyvista/pyvista/issues/3041
Using the [geemap](https://geemap.org/) because [leafmap](https://leafmap.org/) is a spin-off library of [geemap](https://geemap.org/).
geemap
Add GmshModel by @PhilippMetsch as external examples.
resolves https://github.com/pyvista/pyvista/issues/3051
gmsh_model

### Details

- None

PR #3104, opened 2022-07-30
Release/0.35.3

akaszynski:
Release 0.35.3 to address issues in the mne docbuild. Pinging @larsoner for visibility.

Includes:
- #3089
- #3088
- #3071
- #3075
- #3049
- #3048
- #3053
- #3028


larsoner:
Doc build is red

MatthewFlamm:
> Doc build is red

Also getting this same error in #3031.  But not all PRs seem affected.  It might be flaky code/execution in the docs.  Would be good to fix for this release so the docs don't get messed up.

adeak:
>  It might be flaky code/execution in the docs.

I'll check what a local clean build looks like.

> Would be good to fix for this release so the docs don't get messed up.

I thought we only bumped docs.pyvista.org for minor versions (and dev.pyvista.org every n hours).

MatthewFlamm:
> I thought we only bumped docs.pyvista.org for minor versions

If so, we may want to reconsider as this PR includes a feature addition (parameter added to add_mesh) that fixes a bug.

Or require any feature addition to be in a minor release.

adeak:
I'm seeing the same FFT error so far in a local build. I bet this needs something like https://github.com/pyvista/pyvista/pull/2941/files#diff-c32dcf1233c1d39782559141c852a0d0372cb9482b9d1ac3a48a0fc70e2513dcR750-R758 (don't think @akaszynski I didn't notice you sneaking that in there :P)

akaszynski:
Main builds just fine locally, and given that we're adding a feature, how about we abandon this patch release and go with a minor (`0.36`) release instead.

> (don't think @akaszynski I didn't notice you sneaking that in there :P)

Yea, gotta peel that out.

adeak:
For the record in the local build that was the only error, with a warning fixed in https://github.com/pyvista/pyvista/pull/3023/files#diff-355952f3009b3d15d0d85ebb510556c3db59722b98c0d5c7d57abdb2b6ea5363R3192-R3193

PR #3105, opened 2022-07-30
Support complex64 and save complex data

### LINKED ISSUE #2873: Inconsistent complex data representation
Complex data is being improperly represented:

```py
>>> puppy = examples.download_puppy()
>>> fft = puppy.fft()
>>> fft.active_scalars.shape
(1920000,)
>>> fft.point_data.active_scalars.shape
(1920000, 2)
```

The wrapping magic only applies to the active scalars:
```py
>>> fft.active_scalars.dtype
dtype('complex128')
>>> fft.point_data.active_scalars.dtype
dtype('float64')
```

But then I can't get the real dtype when querying the array by name:
```py
>>> fft['JPEGImage'].shape
(1920000,)
>>> fft.point_data['JPEGImage'].shape
(1920000,)
```

The reason why the code in `_check_fft_scalars` works (i.e. it doesn't raise) is because it happens to use `self.point_data.active_scalars`.

_Originally posted by @adeak in https://github.com/pyvista/pyvista/pull/1953#discussion_r906730369_

### LINKED ISSUE #3057: Complex metadata is lost when saving complex arrays
Complex metadata is lost when saving complex arrays.

#### Minimum Reproducible Example

```py
>>> import pyvista as pv
>>> mesh = pv.Sphere()
>>> mesh['data'] = np.zeros(mesh.n_points, dtype=np.complex128)
>>> mesh.save('/tmp/tmp.vtk')
>>> new_mesh = pv.read('/tmp/tmp.vtk')
>>> new_mesh.point_data
pyvista DataSetAttributes
Association     : POINT
Active Scalars  : data
Active Vectors  : None
Active Texture  : None
Active Normals  : Normals
Contains arrays :
    data                    float64    (842, 2)             SCALARS
    Normals                 float32    (842, 3)             NORMALS
```

akaszynski:
Resolves #3057 by storing array type associations as field metadata (as recommended [by Kitware](https://discourse.paraview.org/t/solved-adding-metadata-to-a-vtkinformation-or-a-grid-object-flowing-through-the-pipeline/318). Field data is written on save and restored (and removed) on load.

This PR also resolves #2873.

### DIFF
--- a/examples/01-filter/image-fft.py
+++ b/examples/01-filter/image-fft.py
@@ -64,6 +64,7 @@ fft_image.plot(
     theme=grey_theme,
     log_scale=True,
     text='Moon Landing Image FFT',
+    copy_mesh=True,  # don't overwrite scalars when plotting
 )
 
 
@@ -95,6 +96,7 @@ fft_image.plot(
     theme=grey_theme,
     log_scale=True,
     text='Moon Landing Image FFT with Noise Removed',
+    copy_mesh=True,  # don't overwrite scalars when plotting
 )
 
 
--- a/pyvista/core/dataobject.py
+++ b/pyvista/core/dataobject.py
@@ -127,6 +127,9 @@ class DataObject:
                 f' Must be one of: {self._WRITERS.keys()}'
             )
 
+        # store complex and bitarray types as field data
+        self._store_metadata()
+
         writer = self._WRITERS[file_ext]()
         fileio.set_vtkwriter_mode(vtk_writer=writer, use_binary=binary)
         writer.SetFileName(str(file_path))
@@ -144,6 +147,34 @@ class DataObject:
             if self[array_name].shape[-1] == 4:  # type: ignore
                 writer.SetEnableAlpha(True)
         writer.Write()
+
+    def _store_metadata(self):
+        """Store metadata as field data."""
+        fdata = self.field_data
+        for assoc_name in ('bitarray', 'complex'):
+            for assoc_type in ('POINT', 'CELL'):
+                assoc_data = getattr(self, f'_association_{assoc_name}_names')
+                if assoc_type in assoc_data:
+                    if assoc_data[assoc_type]:
+                        key = f'_PYVISTA_{assoc_name}_{assoc_type}_'.upper()
+                        fdata[key] = list(assoc_data[assoc_type])
+
+    def _restore_metadata(self):
+        """Restore PyVista metadata from field data.
+
+        Metadata is stored using ``_store_metadata`` and contains entries in
+        the format of f'_PYVISTA_{assoc_name}_{assoc_type}_'. These entries are
+        removed when calling this method.
+
+        """
+        fdata = self.field_data
+        for assoc_name in ('bitarray', 'complex'):
+            for assoc_type in ('POINT', 'CELL'):
+                key = f'_PYVISTA_{assoc_name}_{assoc_type}_'.upper()
+                if key in fdata:
+                    assoc_data = getattr(self, f'_association_{assoc_name}_names')
+                    assoc_data[assoc_type] = set(fdata[key])
+                    del fdata[key]
 
     @abstractmethod
     def get_data_range(self):  # pragma: no cover
--- a/pyvista/core/dataobject.py
+++ b/pyvista/core/dataobject.py
@@ -154,10 +154,10 @@ class DataObject:
         for assoc_name in ('bitarray', 'complex'):
             for assoc_type in ('POINT', 'CELL'):
                 assoc_data = getattr(self, f'_association_{assoc_name}_names')
-                if assoc_type in assoc_data:
-                    if assoc_data[assoc_type]:
-                        key = f'_PYVISTA_{assoc_name}_{assoc_type}_'.upper()
-                        fdata[key] = list(assoc_data[assoc_type])
+                array_names = assoc_data.get(assoc_type)
+                if array_names:
+                    key = f'_PYVISTA_{assoc_name}_{assoc_type}_'.upper()
+                    fdata[key] = list(array_names)
 
     def _restore_metadata(self):
         """Restore PyVista metadata from field data.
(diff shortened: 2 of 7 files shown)


akaszynski:
@banesullivan or @adeak, a quick review would be nice so we can get the release out as it's failing.

akaszynski:
Actually, wait a bit on the review, example `image-fft.py` is still failing.

pyvista/core/dataobject.py
```diff
@@ -145,6 +148,34 @@ def save(self, filename: str, binary=True, texture=None):
...
+    def _store_metadata(self):
+        """Store metadata as field data."""
+        fdata = self.field_data
+        for assoc_name in ('bitarray', 'complex'):
+            for assoc_type in ('POINT', 'CELL'):
+                assoc_data = getattr(self, f'_association_{assoc_name}_names')
+                if assoc_type in assoc_data:
+                    if assoc_data[assoc_type]:
+                        key = f'_PYVISTA_{assoc_name}_{assoc_type}_'.upper()
+                        fdata[key] = list(assoc_data[assoc_type])
```
adeak, line 160: Why is this `list` call necessary?

akaszynski:
Based on `git bisect`, f6aa833e06d8a56cd167164e70f201c99504e009 causes the issue.

Based on our conversations, I'm seriously considering buying a decent PC and running the full doc build on a self-hosted runner so we don't end up with these functional regressions.

pyvista/core/filters/uniform_grid.py
```diff
@@ -750,11 +750,9 @@ def _check_fft_scalars(self):
...
-
-        meets_req = scalars.ndim == 2 and scalars.shape[1] == 2
-        if not meets_req:
+        if not np.issubdtype(self.point_data.active_scalars.dtype, np.complexfloating):
             raise ValueError(
                 'Active scalars must be complex data for this filter, represented '
-                'as an array with `dtype=numpy.complex128`.'
+                'as an array with a datatype of `numpy.complex64` or '
+                '`numpy.complex128`.'
             )
```
adeak, line 758: There's a slight mismatch here related to a weird edge case: if someone passes an array with `np.complex256` dtype then they won't trip this error, but downstream operations will probably break (otherwise we'd support `np.complex256` as well).

But I remember noticing and not flagging this before, so it's possible that there's an earlier explicit filter for these two kinds of complex dtypes.

examples/01-filter/image-fft.py
```diff
@@ -64,6 +64,7 @@
     theme=grey_theme,
     log_scale=True,
     text='Moon Landing Image FFT',
+    copy_mesh=True,  # don't overwrite scalars when plotting
```
adeak, line 67: If this fixes the error then we have to rethink how we handle complexes. This would mean that plotting with default `copy_mesh=False` would interfere with RFFT stuff, right?

akaszynski:
Ok 9a16343 fixes the `image-fft.py` example.

Now that we're not removing added scalars and restoring the old ones, `add_mesh` is causing a side-effect of adding and activating scalars again. To me this is a bug, but the logic required to avoid this always seemed like a hack. Always making a shallow copy could have fixed this, but backwards compatibility requires us to make `copy_mesh=True`.

Maybe this is one of those things that we can fix in our `1.0` release (which at this rate of change might just be when people retire Python).

adeak (commented):
Few remarks, let me know when I should continue looking at this (no promises for a fast _and_ thorough review, but I'll see what I can do).

pyvista/core/dataobject.py
akaszynski: We can't assign lists to field data.

Should we enable this? Probably. This PR? Probably not.

akaszynski:
And actually, rather than implementing this last second feature which might have some unintended consequences, I'm opting for creating a PR with just 9a16343.

examples/01-filter/image-fft.py
akaszynski: It's more of the fact that when we add scalars in `add_mesh` we also activate them. Now the active scalars are no longer complex and any `fft` filters are angry about invalid active scalars.

pyvista/core/filters/uniform_grid.py
akaszynski: It's simply not possible to add a `np.complex256` array to our datasets, we check that in:
https://github.com/pyvista/pyvista/blob/9a16343297b70e7d1722540ca0c968ede15f08c8/pyvista/core/datasetattributes.py#L822-L826

akaszynski:
Suggesting we merge this before #2941. Ready for review.

adeak (approved):
I've made a few minor changes, please check. One potential, minor change. Otherwise this LGTM, thanks!

pyvista/core/datasetattributes.py
```diff
@@ -540,17 +541,21 @@ def get_array(self, key: Union[str, int]) -> pyvista_ndarray:
...
+        """Check if array needs to be represented as a different type."""
+        name = narray.VTKObject.GetName()
         if name in self.dataset._association_bitarray_names[self.association.name]:
             narray = narray.view(np.bool_)  # type: ignore
         elif name in self.dataset._association_complex_names[self.association.name]:
-            narray = narray.view(np.complex128)  # type: ignore
+            if narray.dtype == np.float32:
+                narray = narray.view(np.complex64)  # type: ignore
+            if narray.dtype == np.float64:
+                narray = narray.view(np.complex128)  # type: ignore
```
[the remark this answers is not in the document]
akaszynski: That's correct, but the only way it's been added here in `_association_complex_names` is for it to have already had that shape.

We might consider adding in a `assert` here, but I think it's unnecessary.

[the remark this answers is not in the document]
adeak: Sorry, I wasn't clear. What I meant was that `narray` in both cases has shape `(n, 2)`. We know this for a fact. So the views will have shape `(n, 1)`.

_But_ it's just that I missed the next line which squeezes those singleton dimensions away. Sorry, I wasn't paying enough attention. All good.

PR #3109, opened 2022-07-31
Fix rectilinear X coordinates int edge case

akaszynski:
Encountered an edge case where edges are shown as triangles when the x spacing is int for a `pyvista.RectilinearGrid`:

```py

import pyvista
import numpy as np

xrng = np.arange(-10, 10, 5)
yrng = np.arange(-10, 10, 5).astype(float)
zrng = [1]
rec_grid = pyvista.RectilinearGrid(xrng, yrng, zrng)
rec_grid.plot(show_edges=True, cpos='xy')
```

### `main`

tmp

### This branch

image


pyvista/core/grid.py
```diff
@@ -190,6 +190,10 @@ def _from_arrays(
         # Must at least be an x array
         if check_duplicates:
             raise_has_duplicates(x)
+
+        # weird side effects if x is not floating point
```
akaszynski, line 194:
```suggestion
        # edges are shown as triangles if x is not floating point
```

adeak:
This is curious. Do you understand what happens?

The cells are `VTK_PIXEL` as expected, and extracting cell `0` gives me points in a sane (and same across main and this branch) order. And on both branches plotting that single cell with `show_edges=True` shows an ever so faint diagonal, much finer than what would be expected from proper edges. Yet this faint seam seems to be missing with `show_edges=False`. And again, this is the same for both branches, after extracting a single cell with `extract_cells(0)`.

pyvista/core/grid.py
```diff
@@ -190,6 +190,10 @@ def _from_arrays(
         # Must at least be an x array
         if check_duplicates:
             raise_has_duplicates(x)
+
+        # edges are shown as triangles if x is not floating point
+        if not isinstance(x.dtype, np.floating):
```
adeak, line 195:
```suggestion
        if not np.issubdtype(x.dtype, np.floating):
```

adeak (approved):
LGTM, thanks

pyvista/core/grid.py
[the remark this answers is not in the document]
adeak: Come to think of it, even if the bug only affects `x` we should convert every array for consistency...

akaszynski: For whatever reason this has messed up the windows tests...

adeak (dismissed):
LGTM (again).

tests/plotting/test_plotting.py
```diff
@@ -93,19 +93,20 @@
...
+    'test_plot_cell_data',
     'test_plot_complex_value',
-    'test_plot_helper_volume',
     'test_plot_helper_two_volumes',
+    'test_plot_helper_volume',
+    'test_plot_string_array',
+    'test_rectlinear_edge_case',
+    'test_scalars_by_name',
+    'test_user_annotations_scalar_bar_volume',
+    'test_user_annotations_scalar_bar_volume',
```
tkoyama010, line 109: Duplication.
```suggestion
    'test_user_annotations_scalar_bar_volume',
```

tests/plotting/test_plotting.py
```diff
@@ -93,19 +93,20 @@
...
-    'test_enable_stereo_render',  # occurs even without Windows OSMesa
-    'test_plot_add_scalar_bar',
-    'test_plot_cell_data',
-    'test_scalars_by_name',
-    'test_user_annotations_scalar_bar_volume',
-    'test_plot_string_array',
     'test_cmap_list',
     'test_collision_plot',
     'test_enable_stereo_render',
+    'test_enable_stereo_render',
```
tkoyama010, line 99: Duplication.
```suggestion
```

tkoyama010 (changes requested):
Thanks for fixing this. I have small suggestions.

tkoyama010 (approved):
LGTM

PR #3110, opened 2022-07-31
Add various grid filters

akaszynski:
This PR adds various grid filters including:

### `subdivide_tetra`
```py
subdivided = grid.subdivide_tetra()
subdivided.plot(show_edges=True, line_width=2)
```

image

### `partition`
```py
import pyvista as pv
grid = pv.UniformGrid(dims=(5, 5, 5))
out = grid.partition(4, as_composite=True)
out.plot(multi_colors=True, show_edges=True)
```

image

### `explode`

```py
import numpy as np
import pyvista as pv
xrng = np.linspace(0, 1, 3)
yrng = np.linspace(0, 2, 4)
zrng = np.linspace(0, 3, 5)
grid = pv.RectilinearGrid(xrng, yrng, zrng)
exploded = grid.explode()
exploded.plot(show_edges=True)
```
image


pyvista/core/filters/data_set.py
```diff
@@ -4154,15 +4154,20 @@ def plot_over_circular_arc_normal(
         if show:  # pragma: no cover
             plt.show()
 
-    def extract_cells(self, ind, progress_bar=False):
+    def extract_cells(self, ind, assume_unique=False, progress_bar=False):
```
MatthewFlamm, line 4157: This is a breaking change since we do not make progress bar a keyword only arg.  I suspect that it will be very low impact.  Most of the filters I have tried do not really have useful progress bars.

pyvista/core/filters/data_set.py
```diff
@@ -4154,15 +4154,20 @@ def plot_over_circular_arc_normal(
...
         """Return a subset of the grid.
 
         Parameters
         ----------
-        ind : np.ndarray
+        ind : numpy.ndarray
             Numpy array of cell indices to be extracted.
 
-        progress_bar : bool, optional
+        assume_unique : bool, default: False
```
MatthewFlamm, line 4165: This isn't used in this method?

pyvista/core/filters/data_set.py
```diff
@@ -4154,15 +4154,20 @@ def plot_over_circular_arc_normal(
...
+            If the ``ind`` specified are already sorted and unique, then set
+            this to ``True`` to avoid the filter from doing time-consuming
+            sorts and uniquification operations.
+
+        progress_bar : bool, default: False
```
MatthewFlamm, line 4170: Are we using this numpydoc language feature now?  I ask mainly for when I contribute and for reviewing.  My observation is that we use `optional` with default in the description throughout the code.  I personally prefer the style of `default: False`, but it does make the documentation a bit more inconsistent if the default is more complex and it has to be explained in the description.

pyvista/core/filters/data_set.py
```diff
@@ -5213,3 +5218,128 @@ def integrate_data(self, progress_bar=False):
         alg.SetDivideAllCellDataByVolume(False)
         _update_alg(alg, progress_bar, 'Integrating Variables')
         return _get_output(alg)
+
+    def partition(self, n_partitions, generate_global_id=False, as_composite=True):
+        """Break down input dataset into requested number of partitions.
+
+        Cells on boundaries are uniquely to each partition.
+
+        It uses a DIY-based kdtree implementation that builds balances the cell
```
MatthewFlamm, line 5227: To me, DIY implies that the user has to supply the implementation

pyvista/core/filters/data_set.py
```diff
@@ -5213,3 +5218,128 @@ def integrate_data(self, progress_bar=False):
...
+        <https://vtk.org/doc/nightly/html/classvtkRedistributeDataSetFilter.html>`_.
+
+        Parameters
+        ----------
+        n_partitions : int
+            Specify the number of partitions to split the input dataset
+            into. Current implementation results in number of partitions equal
+            to the power of 2 greater than or equal to the chosen value.
+
+        generate_global_id : bool, default: False
```
MatthewFlamm, line 5243: This seems to be unused, and I'm unsure exactly what it does.

pyvista/core/filters/data_set.py
```diff
@@ -5213,3 +5218,128 @@ def integrate_data(self, progress_bar=False):
...
+            Generate global cell ids if ``None`` present in the input.  If
+            global cell ids are present in the input then this flag is
+            ignored.
+
+        as_composite : bool, default: False
+            Return the partitioned dataset as a :class:`pyvista.MultiBlock`.
+
+        Returns
+        -------
+        pyvista.UnStructuredGird or pyvista.MultiBlock
```
MatthewFlamm, line 5253: I would flip the order since `MultiBlock` is the default. Also a typo.

```suggestion
       pyvista.MultiBlock or pyvista.UnstructuredGrid
```

pyvista/core/filters/rectilinear_grid.py
```diff
@@ -0,0 +1,80 @@
...
+        """Create a Tetrahedral mesh from a RectilinearGrid.
+
+        Parameters
+        ----------
+        tetra_per_cell : int, default: 5
+            The number of tetrahedrals to divide each cell in the into. Can be
+            either ``5``, ``6``, or ``12``. If ``mixed=True``, this value is
+            overridden.
+
+        mixed : bool, default: False
```
MatthewFlamm, line 21: Should we also allow `scalars` to be passed in?  Then we would activate them here.  What if there are no active scalars?

pyvista/core/filters/rectilinear_grid.py
```diff
@@ -0,0 +1,80 @@
...
+            Set this to ``True`` to to subdivide some cells into 5 and some
+            cells into 12. Set the active cell scalars of the
+            :class:`pyvista.RectilinearGrid` to be either 5 or 12 to
+            determinine the number of tetrahedra to generate per cell.
+
+        pass_cell_ids : bool, default: False
+            Set if the tetrahedra have scalar data indicating which cell they
```
MatthewFlamm, line 28:
```suggestion
            Set to ``True`` to make the tetrahedra have scalar data indicating which cell they
```

pyvista/core/filters/data_set.py
```diff
@@ -5213,3 +5218,128 @@ def integrate_data(self, progress_bar=False):
...
+        >>> import pyvista as pv
+        >>> xrng = np.linspace(0, 1, 3)
+        >>> yrng = np.linspace(0, 2, 4)
+        >>> zrng = np.linspace(0, 3, 5)
+        >>> grid = pv.RectilinearGrid(xrng, yrng, zrng)
+        >>> exploded = grid.explode()
+        >>> exploded.plot(show_edges=True)
+
+        """
+        split = self.shrink(1.0)  # simply used to isolate each cell
```
MatthewFlamm, line 5333: Depending on merge order of #3112 , you can use `separate` here.

pyvista/core/filters/data_set.py
```diff
@@ -5213,3 +5218,128 @@ def integrate_data(self, progress_bar=False):
         alg.SetDivideAllCellDataByVolume(False)
         _update_alg(alg, progress_bar, 'Integrating Variables')
         return _get_output(alg)
+
+    def partition(self, n_partitions, generate_global_id=False, as_composite=True):
+        """Break down input dataset into requested number of partitions.
+
+        Cells on boundaries are uniquely to each partition.
```
MatthewFlamm, line 5225: Aren't all cells unique to each partition?

In case this stays: `are uniquely` -> `are unique`.

pyvista/core/filters/data_set.py
```diff
@@ -5213,3 +5218,128 @@ def integrate_data(self, progress_bar=False):
...
+        """
+        split = self.shrink(1.0)  # simply used to isolate each cell
+
+        # VTK changed their cell indexing API in 9.0 and
+        if pyvista.vtk_version_info < (9, 0, 0):
+            offset = split.offset.copy()
+            offset -= np.arange(offset.size)
+            offset = np.hstack((offset, split.n_points))
+        else:
+            offset = split.offset
```
MatthewFlamm, line 5341: Attribute `offset` isn't available in PolyData?  Should this be in a different class?

pyvista/core/filters/data_set.py
akaszynski:
```suggestion
    def extract_cells(self, ind, progress_bar=False, assume_unique=False):
```

akaszynski: You're right, kwargs should really be added at the end of a function signature.

pyvista/core/filters/data_set.py
akaszynski: I agree it's inconsistent. If it's alright with you, we can leave it in this PR and have a follow-up before 0.37.0 and fix the docs. I agree that it's way better to use `default: value`

pyvista/core/filters/data_set.py
akaszynski: Yea, this was a copy/pasta from VTK docs, but I think it refers to https://diatomic.github.io/diy/group___algorithms.html

I'll remove it. Reader can still lookup the VTK Docs.

```suggestion
        It uses a kdtree implementation that builds balances the cell
```

pyvista/core/filters/data_set.py
akaszynski:
```suggestion
        split = self.separate()
```

pyvista/core/filters/data_set.py
akaszynski:
```suggestion
        Cells on boundaries are uniquely assigned to each partition without duplication.
```

pyvista/core/filters/data_set.py
```diff
@@ -5213,3 +5218,128 @@ def integrate_data(self, progress_bar=False):
...
+
+        >>> from pyvista import examples
+        >>> mesh = examples.download_bunny()
+        >>> out = mesh.partition(4, as_composite=True)
+        >>> out.plot(multi_colors=True, cpos='xy')
+
+        """
+        # While vtkRedistributeDataSetFilter exists prior to 9.1.0, it doesn't
+        # work correctly, returning the wrong number of partitions.
+        if pyvista.vtk_version_info < (9, 1, 0):
```
akaszynski, line 5280:
```suggestion
        if pyvista.vtk_version_info < (9, 1, 0):  # pragma: no cover
```

pyvista/core/filters/data_set.py
```diff
@@ -5213,3 +5218,128 @@ def integrate_data(self, progress_bar=False):
...
+            raise VTKVersionError('`partition` requires vtk>=9.1.0')
```
akaszynski, line 5281:
```suggestion
            raise VTKVersionError('`partition` requires vtk>=9.1.0')  # pragma: no cover
```

pyvista/core/filters/data_set.py
akaszynski: It's not and it's been removed in 8bcb10af7417fc5ab491b6922f193f7d3fd64ae5.

pyvista/core/filters/data_set.py
akaszynski: Fixed in 8bcb10af7417fc5ab491b6922f193f7d3fd64ae5.

pyvista/core/filters/rectilinear_grid.py
akaszynski: Added in 8bcb10af7417fc5ab491b6922f193f7d3fd64ae5.

pyvista/core/filters/data_set.py
akaszynski: Correct. Fixed in 8bcb10af7417fc5ab491b6922f193f7d3fd64ae5.

pyvista/core/filters/data_set.py
```diff
@@ -5214,6 +5214,137 @@ def integrate_data(self, progress_bar=False):
...
+         Cells on boundaries are uniquely assigned to each partition without duplication.
+
+         It uses a kdtree implementation that builds balances the cell
+         centers among requested number of partitions. Current implementation
+         only supports power-of-2 target partition. If a non-power of two value
+         is specified for ``n_partitions``, then the load balancing simply
+         uses the power-of-two greater than the requested value
+
+         For more details, see `vtkRedistributeDataSetFilter
+         <https://vtk.org/doc/nightly/html/classvtkRedistributeDataSetFilter.html>`_.
```
tkoyama010, line 5229:
```suggestion
        """Break down input dataset into a requested number of partitions.

         Cells on boundaries are uniquely assigned to each partition without duplication.

         It uses a kdtree implementation that builds balances the cell
         centers among a requested number of partitions. The current implementation
         only supports power-of-2 target partition. If a non-power of two value
         is specified for ``n_partitions``, then the load balancing simply
         uses the power-of-two greater than the requested value

         For more details, see `vtkRedistributeDataSetFilter
         <https://vtk.org/doc/nightly/html/classvtkRedistributeDataSetFilter.html>`_.
```

pyvista/core/filters/data_set.py
```diff
@@ -5214,6 +5214,137 @@ def integrate_data(self, progress_bar=False):
...
+
+         Parameters
+         ----------
+         n_partitions : int
+             Specify the number of partitions to split the input dataset
+             into. Current implementation results in number of partitions equal
+             to the power of 2 greater than or equal to the chosen value.
```
tkoyama010, line 5236:
```suggestion
             Specify the number of partitions to split the input dataset
             into. Current implementation results in a number of partitions equal
             to the power of 2 greater than or equal to the chosen value.
```

pyvista/core/filters/data_set.py
```diff
@@ -5214,6 +5214,137 @@ def integrate_data(self, progress_bar=False):
...
+
+         generate_global_id : bool, default: False
+             Generate global cell ids if ``None`` present in the input.  If
+             global cell ids are present in the input then this flag is
+             ignored.
```
tkoyama010, line 5241:
```suggestion
             Generate global cell ids if ``None`` are present in the input.  If
             global cell ids are present in the input then this flag is
             ignored.
```

pyvista/core/filters/data_set.py
```diff
@@ -5214,6 +5214,137 @@ def integrate_data(self, progress_bar=False):
...
+         --------
+         Partition a simple UniformGrid into a :class:`pyvista.MultiBlock`
+         containing each partition.
+
+         >>> import pyvista as pv
+         >>> grid = pv.UniformGrid(dims=(5, 5, 5))
+         >>> out = grid.partition(4, as_composite=True)
+         >>> out.plot(multi_colors=True, show_edges=True)
+
+         Partition the Stanford bunny.
```
tkoyama010, line 5268:
```suggestion
         Partition of the Stanford bunny.
```

pyvista/core/filters/rectilinear_grid.py
```diff
@@ -0,0 +1,101 @@
+"""Filters module with class to manage filters/algorithms for rectilinear grid datasets."""
```
tkoyama010, line 1:
```suggestion
"""Filters module with the class to manage filters/algorithms for rectilinear grid datasets."""
```

pyvista/core/filters/rectilinear_grid.py
```diff
@@ -0,0 +1,101 @@
...
+        progress_bar: bool = False,
+    ):
+        """Create a Tetrahedral mesh from a RectilinearGrid.
+
+        Parameters
+        ----------
+        tetra_per_cell : int, default: 5
+            The number of tetrahedrals to divide each cell in the into. Can be
+            either ``5``, ``6``, or ``12``. If ``mixed=True``, this value is
+            overridden.
```
tkoyama010, line 30:
```suggestion
        tetra_per_cell : int, default: 5
            The number of tetrahedrons to divide each cell into. Can be
            either ``5``, ``6``, or ``12``. If ``mixed=True``, this value is
            overridden.
```

pyvista/core/filters/rectilinear_grid.py
```diff
@@ -0,0 +1,101 @@
...
+
+        mixed : str, bool, sequence, default: False
+            When set, subdivides some cells into 5 and some cells into 12. Set
+            to ``True`` to use the active cell scalars of the
+            :class:`pyvista.RectilinearGrid` to be either 5 or 12 to
+            determining the number of tetrahedra to generate per cell.
+
+            When a sequence, uses these value to subdivide the cells. When a
+            string, uses cell array rather than the active array to determine
+            the number of tetrahedra to generate per cell.
```
tkoyama010, line 40:
```suggestion
            When set, subdivides some cells into 5 and some cells into 12. Set
            to ``True`` to use the active cell scalars of the
            :class:`pyvista.RectilinearGrid` to be either 5 or 12 to
            determining the number of tetrahedra to generate per cell.

            When a sequence, uses these values to subdivide the cells. When a
            string uses a cell array rather than the active array to determine
            the number of tetrahedra to generate per cell.
```

pyvista/core/filters/rectilinear_grid.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Returns
+        -------
+        pyvista.UnstructuredGrid
+            UnstructuredGrid containing the tetrahedral cells.
+
+        Examples
+        --------
+        Divide a rectangular grid into tetrahedrals. Each cell contains by
+        default 5 tetrahedrals.
```
tkoyama010, line 58:
```suggestion
        Divide a rectangular grid into tetrahedrons. Each cell contains by
        default 5 tetrahedrons.
```

pyvista/core/filters/rectilinear_grid.py
```diff
@@ -0,0 +1,101 @@
...
+
+        >>> import numpy as np
+        >>> import pyvista as pv
+        >>> xrng = np.linspace(0, 1, 2)
+        >>> yrng = np.linspace(0, 1, 2)
+        >>> zrng = np.linspace(0, 2, 3)
+        >>> grid = pv.RectilinearGrid(xrng, yrng, zrng)
+        >>> grid.plot()
+
+        Now, generate the tetrahedra plot the exploded view of the cell.
```
tkoyama010, line 70:
```suggestion
        Now, generate the tetrahedra plot in the exploded view of the cell.
```

pyvista/core/filters/rectilinear_grid.py
```diff
@@ -0,0 +1,101 @@
...
+
+        >>> tet_grid = grid.to_tetrahedra()
+        >>> tet_grid.explode(factor=0.5).plot(show_edges=True)
+
+        Take the same grid but divide the first cell into 5 cells and the other
+        cell into 12 tetrahedrals per cell.
```
tkoyama010, line 76:
```suggestion
        Take the same grid but divide the first cell into 5 cells and the other
        cell into 12 tetrahedrons per cell.
```

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
```diff
@@ -5214,6 +5214,137 @@ def integrate_data(self, progress_bar=False):
...
+        >>> exploded = grid.explode()
+        >>> exploded.plot(show_edges=True)
+
+        """
+        split = self.separate_cells()
+        if not isinstance(split, pyvista.UnstructuredGrid):
+            split = split.cast_to_unstructured_grid()
+
+        # VTK changed their cell indexing API in 9.0
+        if pyvista.vtk_version_info < (9, 0, 0):
```
akaszynski, line 5337:
```suggestion
        if pyvista.vtk_version_info < (9, 0, 0):  # pragma: no cover
```

tkoyama010 (approved):
LGTM

tkoyama010 (approved):

PR #3112, opened 2022-08-01
Add separate cells feature

akaszynski:
Adds `separate_cells` as a public method to `pyvista.DataSet` and an example that demonstrates its usage.

While `separate_cells` is a (very) thin wrapper around `shrink(1.0)`, I think we all prefer to be explicit and we don't have an obvious way to separate all the individual cells of a dataset. This is absolutely necessary to be able to assign individual values to each point for each cell, which is needed for finite element results which are continuous within a cell and discontinuous between cells.

Here's a cool plot demonstrating it's usage.

tmp

### DIFF
--- a/examples/02-plot/point-cell-scalars.py
+++ b/examples/02-plot/point-cell-scalars.py
@@ -1,0 +1,79 @@
+"""
+.. _point_cell_scalars_example:
+
+Point Cell Scalars
+~~~~~~~~~~~~~~~~~~
+
+This example demonstrates how to add point scalars for each individual cell to
+a dataset.
+
+
+"""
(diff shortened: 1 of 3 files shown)


adeak (approved):
LGTM, I agree it's better to have dedicated methods instead of obscure tricks.

MatthewFlamm (approved):
I think it is a fine line IMO to add too many convenience functions, but I think this one is very good since it is not intuitive to think about `shrink` when we want to `separate`.

tkoyama010 (approved):
This is interesting. I didn't know we can use shrink in this way.

PR #3117, opened 2022-08-01
Remove vtk requirement

### LINKED ISSUE #3116: BUG: PyVista pins to RC
In MNE-Python we've had issues in CIs with PyVista pinning to a VTK release candidate.

IMO pinning to a release candidate in itself is a bad idea -- it's a candidate, not actually released yet, and could easily have bugs, probably isn't easy to install everywhere yet, etc. -- but also it would be good if pyvista itself had a policy to support as far back in VTK as possible/reasonable. Most scientific Python packages I know (matplotlib, numpy, scipy, scikit-learn) have adopted a policy to try to support ~2 years back.

Maybe that's too tough here, but supporting *at least* 9.1 seems like a minimum to me. Supporting back to 9.0 would be even better.

Only supporting the latest (actual) release is a pretty restrictive condition on libraries that depend on PyVista!

akaszynski:
Resolves #3116 by removing any version requirement for vtk. I would have prefered to keep at least `vtk>=8.1.2`, but having any sort of version requirement for vtk means that the release candidates will not install for Python 3.10, meaning PyVista becomes unusable for Python 3.10 unless you use conda.

We will one day set a minimum version for vtk, but to @larsoner's point, we plan on supporting the oldest version of vtk available on PyPI for the oldest version of currently supported Python.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ install_requires = [
     'pillow',
     'appdirs',
     'scooby>=0.5.1',
-    'vtk>=8.1.2',
+    'vtk',
 ]
 
 readme_file = os.path.join(filepath, 'README.rst')
--- a/setup.py
+++ b/setup.py
@@ -18,8 +18,7 @@ install_requires = [
     'pillow',
     'appdirs',
     'scooby>=0.5.1',
-    'vtk>=8.1.2, <=9.2.0; python_version < "3.10"',
-    'vtk~=9.2.0rc1; python_version == "3.10"',
+    'vtk>=8.1.2',
 ]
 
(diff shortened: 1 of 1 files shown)


larsoner:
In the meantime we could merge https://github.com/conda-forge/pyvista-feedstock/pull/67 if you want. But if you plan to merge this and cut 0.36.1 quickly then we can just close https://github.com/conda-forge/pyvista-feedstock/pull/67

larsoner (approved):
I think it's the best we can do, and we just wait for the kitware folks for 3.10 support to land properly on PyPI!

PR #3118, opened 2022-08-01
Release 0.36.1

akaszynski:
Release 0.36.1

This only includes #3117, which removes vtk version requirement so we can install pyvista on Python 3.10 and support conda.


larsoner (approved):

larsoner:
Thanks for taking care of this so quickly @akaszynski !

Issue #1761, opened 2021-10-28
Meaningless error message when adding the data values to the point data from unflatten array


mkondratyev85:
**Describe the bug, what's wrong, and what you expect:**

When I try to set point data to a mesh from 3D numpy array (which, I know, is wrong, because I should flatten the array beforehand), I get the error message that, at least to me, has no sense at all.

```python
import numpy as np
import pyvista as pv

data = np.array( [[[4., 3.],
                   [4., 1.],
                   [4., 1.],
                   [4., 1.]],
                 
                  [[4., 4.],
                   [4., 4.],
                   [4., 4.],
                   [4., 4.]],
                 
                  [[4., 4.],
                   [4., 4.],
                   [4., 1.],
                   [4., 1.]],
                 
                  [[4., 4.],
                   [4., 4.],
                   [4., 4.],
                   [4., 1.]]]
                )        

x = np.array([0,500,1000,1500])
y = np.array([0,500,1000,1500])
z = np.array([0,1000])
out = pv.RectilinearGrid(x, y, z)
out['data_values'] = data#.flatten(order='F')
```

And here is the error message:

```
Traceback (most recent call last):
  File "pyvt.py", line 29, in <module>
    out['data_values'] = data#.flattten(order='F')
  File "/home/user/.local/lib/python3.7/site-packages/pyvista/core/dataset.py", line 658, in __setitem__
    raise_not_matching(scalars, self)
  File "/home/user/.local/lib/python3.7/site-packages/pyvista/utilities/helpers.py", line 725, in raise_not_matching
    raise ValueError(f'Number of scalars ({scalars.size}) '
ValueError: Number of scalars (32) must match either the number of points (32) or the number of cells (9).
```

The satirical aspect adds to the fact that the number of scalars (32) does indeed match the number of points(32).

Of course, when I flatten the array - everything works as expected. But when random user sees this error message, it is far from obvious that the cause of the problem is actually the shape of the array, not its size.

Is it possible to make this message more descriptive?

adeak:
This was changed in #3124, we now report `scalars.shape[0]` in the error message. If you find this confusing in some situations still, please let us know.

PR #1321, opened 2021-05-17
🚧 Fix error message for scalars of multiple dimensions

MatthewFlamm:
### Overview

 

This gives the error message to indicate that it is the first dimension size that must match.  All current usages of `raise_not_matching` seem to provide a `DataSet`.
 

### Details
Related to #1319 and https://github.com/pyvista/pyvista-support/issues/425

If we can get rid of, or split out, the `vtkTable` piece, we could also consider having this function also check whether the scalars should be applied to points or cells.  This would very minorly simplify the consuming functions.

This refactoring would also provide a central location to do reshaping like suggested in https://github.com/pyvista/pyvista/issues/1319#issuecomment-842699249. #

PR #3124, opened 2022-08-03
Report number of scalars instead of size in error about scalars shape mismatch

adeak:
I ran into this less than helpful error:
```
>>> import numpy as np
>>> import pyvista as pv
>>> 
>>> mesh = pv.PolyData(np.arange(4 * 3.0).reshape(4, 3))
>>> mesh['scalars'] = np.arange(mesh.n_points).reshape(-1, 2)
  File "/home/adeak/pyvista/pyvista/core/dataset.py", line 1907, in __setitem__
    raise_not_matching(scalars, self)
  File "/home/adeak/pyvista/pyvista/utilities/helpers.py", line 1129, in raise_not_matching
    raise ValueError(
ValueError: Number of scalars (4) must match either the number of points (4) or the number of cells (4).
```
The issue is that the checks check `scalars.shape[0]`, but the error reports `scalars.size`. In case the total number of scalars is correct but the shape is wrong (like it was in my case), the error can be confusing.

I didn't thoroughly check every appearance of this helper, I came across it via `DataSet.__setitem__()`. But it should only ever be called for scalars, and I assume the fix is valid for every scalars case.

Closes https://github.com/pyvista/pyvista/issues/1761.

### DIFF
--- a/pyvista/utilities/helpers.py
+++ b/pyvista/utilities/helpers.py
@@ -1124,10 +1124,10 @@ def raise_not_matching(scalars, dataset):
     """
     if isinstance(dataset, _vtk.vtkTable):
         raise ValueError(
-            f'Number of scalars ({scalars.size}) must match number of rows ({dataset.n_rows}).'
+            f'Number of scalars ({scalars.shape[0]}) must match number of rows ({dataset.n_rows}).'
         )
     raise ValueError(
-        f'Number of scalars ({scalars.size}) '
+        f'Number of scalars ({scalars.shape[0]}) '
         f'must match either the number of points ({dataset.n_points}) '
         f'or the number of cells ({dataset.n_cells}).'
     )


MatthewFlamm:
This isn't used in many places.  It shows up in and all of them check shape[0].

https://github.com/pyvista/pyvista/blob/1948096d6e90f802dacc5a25027a9182103c7412/pyvista/core/dataset.py#L1900-L1907

https://github.com/pyvista/pyvista/blob/1948096d6e90f802dacc5a25027a9182103c7412/pyvista/plotting/mapper.py#L271-L290

https://github.com/pyvista/pyvista/blob/1948096d6e90f802dacc5a25027a9182103c7412/pyvista/plotting/plotting.py#L2960-L2967

https://github.com/pyvista/pyvista/blob/1948096d6e90f802dacc5a25027a9182103c7412/pyvista/plotting/plotting.py#L3383-L3388

MatthewFlamm (approved):

adeak:
Thanks for checking @MatthewFlamm!

MatthewFlamm:
Thinking about this more, this PR was very familiar to me :). Have all the complexities from #1321 been resolved?

adeak:
Aw crap :smile: 

Although, wouldn't this issue affect the very checks you linked to? If I understand correctly (which might not be the case; it's 1 AM here again), we always check `scalars.shape[0]` before calling the raising helper, so mentioning the same dimension in the raising helper makes sense. And https://github.com/pyvista/pyvista/pull/1321 would imply changing these checks outside the helper, which would then need changes to the helper to accommodate that.

MatthewFlamm:
One issue is https://github.com/pyvista/pyvista/blob/1948096d6e90f802dacc5a25027a9182103c7412/pyvista/plotting/plotting.py#L2940-L2941

Where the scalars are raveled, so size is the most descriptive to the user in this case.

adeak:
Do you think that we can get away with mentioning the full shape instead? I guess that could also be confusing when transformations like `ravel()` were performed before the raise.

MatthewFlamm:
Although in this case `shape[0]` == `size`, so this PR is still accurate.  I think this ravel issue is a separate bug/feature.

MatthewFlamm:
I'm keeping my approval, since this PR improves things.

MatthewFlamm:
Does this close https://github.com/pyvista/pyvista/issues/1761?  Note that the data that is passed into `__setitem__` fails, but in `add_volume`, if this same data were passed in, it would be fine.  Again, I think this is a problem outside of this PR.  I overcomplicated the thinking in the previous PR.

adeak:
> Does this close #1761? Note that the data that is passed into `__setitem__` fails, but in `add_volume`, if this same data were passed in, it would be fine. Again, I think this is a problem outside of this PR. I overcomplicated the thinking in the previous PR.

Indeed it does, good catch, thanks.

PR #3131, opened 2022-08-04
Distance along spline example

akaszynski:
Add an example based on this https://github.com/pyvista/pyvista/discussions/3123#discussioncomment-3321672 discussion.

### DIFF
--- a/examples/00-load/create-spline.py
+++ b/examples/00-load/create-spline.py
@@ -1,10 +1,11 @@
 """
-.. _ref_create_spline:
+.. _ref_create_spline_example:
 
 Creating a Spline
 ~~~~~~~~~~~~~~~~~
 
-Create a spline/polyline from a numpy array of XYZ vertices
+Create a spline/polyline from a numpy array of XYZ vertices using
+:func:`pyvista.Spline`.
 """
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


examples/02-plot/distance-along-spline.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+###############################################################################
+# Determine the coordinates matching distance along a spline
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Here we write a simple function that gets the closest point matching a distance along a spline and then generate labels for those points.
+
+
+def get_point_along_spline(distance):
+    """Return the closest point on the spline a given a length along a spline."""
```
tkoyama010, line 44:
```suggestion
    """Return the closest point on the spline a given length along a spline."""
```

tkoyama010 (changes requested):

examples/02-plot/distance-along-spline.py
adeak: Good catch, but I think this is supposed to be the other way around:
```suggestion
    """Return the closest point on the spline given a length along the spline."""
```

examples/00-load/create-spline.py
```diff
@@ -1,10 +1,11 @@
 """
-.. _ref_create_spline:
+.. _ref_create_spline_example:
```
akaszynski, line 2:
```suggestion
.. _create_spline_example:
```

tkoyama010 (approved):
LGTM

PR #2994, opened 2022-07-15
3D Celestial Bodies (planet3D)

tkoyama010:
### Overview

See https://github.com/pyvista/pyvista/discussions/1501

 

Requires:
- [x] #3075 

#### 3D Earth and Celestial Bodies
sphx_glr_planets_001

sphx_glr_planets_002

sphx_glr_planets_003

PR #3135, opened 2022-08-07
Add support for backface properties

adeak:
This builds on top of #2941 for its `Property` feature. For now this PR has `feat/add_composite` as the base branch for CI, so before this is unset as draft the base branch has to be changed to `main`! This is otherwise ready for review.

The new feature itself is fairly minor, it just exposes backface properties via calls to `plotter.add_mesh()`. I _may_ have gone a bit overboard and implemented the main results of a differential geometry paper in order to have a nice animation:

sphere eversion animation from the examples

#### TODO:

  - [x] change PR base to `main` once #2941 is merged -> draft until then
  - [x] move eversion example to advanced examples section

#### ~~maybe TODO:~~ future work:

  - [x] check if textures play nicely on the backface
  - [x] check why enabling PBR makes the backfaces disappear; perhaps backface culling is enabled automatically


akaszynski:
Guess we need to merge #2941. I'll work on that this evening.

Been playing around with `cadquery` and pyvista:

tmp

examples/02-plot/sphere_eversion.py
```diff
@@ -0,0 +1,263 @@
...
+    y /= np.cos(theta) ** n
+    z = (
+        lamda * (
+            (w * np.sin(theta) * (np.sin(n * phi) - q * t)) / np.cos(theta)**n
+            - t / n * np.cos(n * phi)
+        )
+        - (1 - lamda) * eta**(1 + kappa) * t * abs(t)**(2 * kappa)
+            * np.sin(theta) / np.cos(theta)**(2 * n)  # noqa
+    )
+    # fmt: on
```
adeak, line 144: Disabling linting is never great form, but complex math equations like these are the epitomy of the kind of code that `black` butchers. Here are the equations:
equation (12) from the paper
equation (15) from the paper

And the `# noqa` directive is necessary in order to have additional indentation on the corresponding line, otherwise flake8 complains.

akaszynski: Agreed. Black is good for 99%, but these long multi-line equations look terrible when less than 100 char.

tkoyama010: @adeak You maybe know it but GitHub is now supporting [TeX rendering](https://github.blog/2022-05-19-math-support-in-markdown/).

$$ x = {-b \pm \sqrt{b^2-4ac} \over 2a} $$

adeak: Thanks @tkoyama010, I did know that. But it would've been tedious to retype those equations myself (and even if the arxiv version has the tex source, it might use custom commands making copy-pasting also tedious), and it would've opened the possibility for typos in my transcription, which could've led to false positive discussions in review here because the code might not match my (wrong) transcription.

adeak:
I didn't expect history to be mucked up this bad by my choice to open this PR on top of a feature branch first. Let me know if I should reopen a new PR with my cherrypicked changes for a less horrible merge commit message at the end.

examples/02-plot/sphere_eversion.py
```diff
@@ -0,0 +1,262 @@
...
+# initial parameters, will be updated
+t = -1 / Q
+q = Q
+p = xi = alpha = 0
+eta = 1
+
+# sphere -> inverted wormhole
+theta, phi = np.mgrid[-np.pi / 2 : np.pi / 2 : 200j, -np.pi : np.pi : 400j]
+h, phi = sphere_to_cylinder(theta, phi)
+for lamda in np.linspace(0, 1, n_steps, endpoint=False):
```
adeak, line 196: Codespell is complaining about my choice of `lamda` here. I want to follow the paper's convention to call this lambda, but of course `lambda` is a statement in Python so we can't use that. Is it OK to add `lamda` to codespell's ignore list? Not sure if that can lead to false negatives later.

akaszynski: I'm fine with that.

MatthewFlamm: It would be ideal if codespell supported line/module/block level exclusion, but it doesn't.

https://github.com/codespell-project/codespell/issues/1212

One suggestion from this thread was to use --ignore-regex.  But "now you have two problems".  I'm also fine with simply adding it to the ignore list since codespell doesn't support this functionality directly.

MatthewFlamm: There is also `--exclude-file/-x` but it looks like it will be brittle to changes in style that may not be obvious to a contributor.

adeak: Thanks for looking into that @MatthewFlamm. I don't think regex would work, because the word appears in all sorts of contexts in the file (otherwise I'd be happy with regex, I'm fairly fluent with it). So I think we're stuck with either ignoring the word or ignoring the file, and the former seems the less harmful option.

I've had some very busy weeks (I'm switching jobs pretty much as we speak) but I'll soon be able to follow up on this PR and others that I've been dragging my feet on.

tkoyama010: How about a name like `.codespellignore` that won't confuse contributors? I just added this at https://github.com/pyvista/pyvista-tutorial/pull/88.

adeak: Thanks @banesullivan, I considered `lambda_` as an alternative, but since "lamda" is actually how many people write it (and how modern Greek people say and write it), I'm leaning towards keeping the current use and ignoring it for codespell.

adeak: Added "lamda" to the codespell ignore list in https://github.com/pyvista/pyvista/pull/3135/commits/c1ff7101fa3369eae1d9a8d5b55e4caab9379065, thank you all for the inputs.

adeak:
Have the sides of our animations always looked this jagged? Or did we change something about anti-aliasing? Or is it just the small image size I used to reduce the size of the GIF?

rendered GIF from the sphere eversion example

MatthewFlamm:
> Have the sides of our animations always looked this jagged? Or did we change something about anti-aliasing? Or is it just the small image size I used to reduce the size of the GIF?

Could this be caused by #3168 changing the default image size for "fast builds"?

akaszynski:
> > Have the sides of our animations always looked this jagged? Or did we change something about anti-aliasing? Or is it just the small image size I used to reduce the size of the GIF?
> 
> Could this be caused by #3168 changing the default image size for "fast builds"?

Yes. Reducing the image size improves the build speed at the expense of quality. Tradeoffs...

adeak:
I've run a full build since, and I see the same behaviour. In fact just running the example in python gives me similar aliasing, although this is harder to notice on the default grey background:

newer sphere eversion GIF with aliasing

Compare with my original gif from the PR comment:

older sphere eversion GIF with no aliasing

Seems like [the render passes PR](https://github.com/pyvista/pyvista/pull/2948/) changed the default meaning of `global_theme.anti_aliasing = None`. If I manually set this to `'ssaa'` it looks nice again.

examples/02-plot/backface_props.py
```diff
@@ -0,0 +1,77 @@
...
+backface_params = dict(color='orangered')
+meshes.plot(color='aquamarine', backface_params=backface_params, smooth_shading=True)
+
+
+###############################################################################
+# A more interesting use case is helping visualize the orientation of complex,
+# self-intersecting surfaces. For instance :func:`Catalan's minimal surface
+# <pyvista.ParametricCatalanMinimal>` has a complex shape, and coloring the
+# front and backfaces differently helps viewers comprehend the intricate
+# structure of the surface.
```
tkoyama010, line 36:
```suggestion
# A more interesting use case is helping visualize the orientation of complex,
# self-intersecting surfaces. For instance :func:`Catalan's minimal surface
# <pyvista.ParametricCatalanMinimal>` has a complex shape, and coloring the
# front and backfaces differently help viewers comprehend the intricate
# structure of the surface.
```

examples/02-plot/backface_props.py
```diff
@@ -0,0 +1,77 @@
...
+
+catalan = pv.ParametricCatalanMinimal()
+backface_params = dict(color='forestgreen', specular=1.0, specular_power=50.0)
+catalan.plot(color='dodgerblue', backface_params=backface_params, smooth_shading=True)
+
+
+###############################################################################
+# In case of non-orientable surfaces adding specific backface properties can
+# make the non-orientable quality very obvious by the emergence of "seams"
+# where the face properties are discontinuous.
```
tkoyama010, line 46:
```suggestion
# In the case of non-orientable surfaces adding specific backface properties can
# make the non-orientable quality very obvious by the emergence of "seams"
# where the face properties are discontinuous.
```

examples/02-plot/sphere_eversion.py
```diff
@@ -0,0 +1,262 @@
+"""
+.. _sphere_eversion_example:
+
+Turning the sphere inside out
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
tkoyama010, line 5:
```suggestion
Turning the sphere inside out
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

examples/02-plot/sphere_eversion.py
```diff
@@ -0,0 +1,262 @@
...
+topology).
+
+The mathematics involved can seem a bit, well, involved. What
+matters is the overall process visible in the animation: first
+the sphere is corrugated and stretched out a bit to allow some
+legroom for the smooth transformation, then the lobes are twisted
+around through each other, and the process is reversed in order
+to unfold the sphere. It's not obvious that the transformation
+is truly smooth; this was proved in the paper by Bednorz and
+Bednorz!
```
tkoyama010, line 23:
```suggestion
The mathematics involved can seem a bit, well, involved. What
matters are the overall process visible in the animation: first
the sphere is corrugated and stretched out a bit to allow some
legroom for the smooth transformation, then the lobes are twisted
around through each other, and the process is reversed in order
to unfold the sphere. It's not obvious that the transformation
is truly smooth; this was proved in the paper by Bednorz and
Bednorz!
```

examples/02-plot/sphere_eversion.py
```diff
@@ -0,0 +1,262 @@
...
+
+# unfolded wormhole -> closed wormhole
+for q in np.linspace(Q, 0, n_steps):
+    p = 1 - abs(q * t)
+    x, y, z = cylinder_to_wormhole(h, phi, t, p, q)
+    x2, y2, z2 = close_wormhole(x, y, z, eta, xi, alpha)
+    save_frame(x2, y2, z2)
+
+# closed wormhole turned inside out (flip sign of time)
+# unfolded wormhole -> closed wormhole
```
tkoyama010, line 217:
```suggestion
# closed wormhole turned inside out (a flip sign of time)
# unfolded wormhole -> closed wormhole
```

tkoyama010 (changes requested):

examples/02-plot/sphere_eversion.py
adeak: "Flip" is a verb here originally, so this doesn't work.

examples/02-plot/backface_props.py
adeak: Hmm, I could see this going both ways, thanks. Not sure if any of the natives can tell which version is more correct.

MatthewFlamm: I actually think both are wrong here.  There is no subject to the sentence. I'll make a separate suggestion.

MatthewFlamm: On second read, I think the addition of a comma will clarify the sentence.  I think either with or without "the" is correct though, it is personal preference.

examples/02-plot/backface_props.py
```diff
@@ -0,0 +1,77 @@
...
+# front and backfaces differently helps viewers comprehend the intricate
+# structure of the surface.
+
+catalan = pv.ParametricCatalanMinimal()
+backface_params = dict(color='forestgreen', specular=1.0, specular_power=50.0)
+catalan.plot(color='dodgerblue', backface_params=backface_params, smooth_shading=True)
+
+
+###############################################################################
+# In case of non-orientable surfaces adding specific backface properties can
```
MatthewFlamm, line 44: Please merge with the other comment if desired.

```suggestion
# In case of non-orientable surfaces, adding specific backface properties can
```

examples/02-plot/backface_props.py
adeak: OK, let me keep both, thanks @MatthewFlamm

examples/02-plot/backface_props.py
adeak: "Outdated suggestions cannot be applied", bah! Could you please suggest it again so that I can credit you in the commit?

MatthewFlamm:
```suggestion
# In case of non-orientable surfaces, adding specific backface properties can
```

adeak: Is it possible that you made the suggestion on a stale state, before the previous commit? The "the" isn't added here yet and github is being weird when I try to commit...

MatthewFlamm: Github is being weird I think, or it is a cache issue (my money is on this).  I'll try again.

examples/02-plot/backface_props.py
```diff
@@ -0,0 +1,77 @@
...
+# front and backfaces differently helps viewers comprehend the intricate
+# structure of the surface.
+
+catalan = pv.ParametricCatalanMinimal()
+backface_params = dict(color='forestgreen', specular=1.0, specular_power=50.0)
+catalan.plot(color='dodgerblue', backface_params=backface_params, smooth_shading=True)
+
+
+###############################################################################
+# In the case of non-orientable surfaces adding specific backface properties can
```
MatthewFlamm, line 44:
```suggestion
# In the case of non-orientable surfaces, adding specific backface properties can
```

examples/02-plot/backface_props.py
adeak: Perfect, thanks. Also: [you seem to be right](https://shouldiblamecaching.com/) ;)

examples/02-plot/sphere_eversion.py
tkoyama010: I see, that makes sense.

adeak:
Considering my workload for the near future I don't think I'll be able to add to this for a while. It would be nice to figure out if we can do anything about PBR here, but these can go in follow-up PRs. So this is final for now in case anyone wants to take another look.

Also thanks @akaszynski for the merge.

akaszynski:
> Also thanks @akaszynski for the merge.

I'll try to take a look at the PBR issue.

MatthewFlamm:
I read through this and I cannot really meaningfully comment on the back face property implementation, although it looks good from a naive read through.  The math/implementation for sphere eversion is impressive!  I also cannot meaningfully comment on it.  It really showcases how complex visualizations can be accomplished with only a little bit of PyVista code.  I wonder if it should go in "advanced" or perhaps a new category "showcase" (or something like this) as the math might scare off users at first glance in the plot section.

adeak:
> It really showcases how complex visualizations can be accomplished with only a little bit of PyVista code. I wonder if it should go in "advanced" or perhaps a new category "showcase" (or something like this) as the math might scare off users at first glance in the plot section.

Very good point, thanks @MatthewFlamm! It did cross my mind at one point when I last looked at https://github.com/pyvista/pyvista/pull/2994, reminding me of the "advanced" section :smile: But then I forgot to come back to this thought. I agree that this is better tucked away in advanced/ for now. Hopefully the basic backface example makes this feature discoverable enough.

adeak:
I've moved the sphere eversion example into the advanced section. I've also updated the simple example a bit to demonstrate the new `Actor` class and its `Actor.backface_prop` property, and ended up doing some non-trivial changes while fixing numpydoc warnings. For instance, I changed all mapper types to pass the `theme` as a keyword parameter as that seems a bit more robust, and I defined a default of `theme=None` in the ABC because mypy complained that `type(self)()` needs a positional arg (and that's right). Please review the new changes carefully.

I'm also still getting the same warning twice:
```
WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.CompositePolyDataMapper.copy':
  RT01: No Returns section found

WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.DataSetMapper.copy':
  RT01: No Returns section found
```
I _think_ I solved this by hinting the return type of `_BaseMapper.copy()`, but I'll have to do a full doc build to make sure of that (and that takes figuratively literally forever for some reason).

adeak:
> I _think_ I solved this by hinting the return type of `_BaseMapper.copy()`, but I'll have to do a full doc build to make sure of that (and that takes figuratively literally forever for some reason).

Well that wasn't it. What I don't get is why all the other methods with no `Returns` section doesn't raise the same validation errors. Almost all of them are properties, could that be it? The one other method that's not a property is
https://github.com/pyvista/pyvista/blob/314e6b1014207b4be9297593e90ec97bbccd67bf/pyvista/plotting/mapper.py#L278-L280

Perhaps the barely existent docstring and the lack of `return` statement makes this other method exempt from validation...

My only issue is that the only thing I can put in the `Returns` section of `_BaseMapper.copy()` is `_BaseMapper`. Is it not an issue to have a private ABC show up in the rendered docs? Or should I wrap the `copy()` method in child classes and copy the docstring?

pyvista/plotting/mapper.py
```diff
@@ -763,28 +770,28 @@ class FixedPointVolumeRayCastMapper(_vtk.vtkFixedPointVolumeRayCastMapper, _Base
...
+        super().__init__(theme=theme)
 
 
 class OpenGLGPUVolumeRayCastMapper(_vtk.vtkOpenGLGPUVolumeRayCastMapper, _BaseVolumeMapper):
     """Wrap _vtk.vtkOpenGLGPUVolumeRayCastMapper."""
 
     def __init__(self, theme=None):
         """Initialize this class."""
-        super().__init__(theme)
+        super().__init__(theme=theme)
```
adeak, line 790: Codecov says these 3 `super().__init__()` calls are uncovered, I assume this is because of OSMesa. Since this is pulling down coverage in this PR: should we no-cover these three classes entirely?

adeak: Actually, do we even need these `__init__` definitions? Looks like they don't do anything compared to the parent super?

akaszynski: > Actually, do we even need these __init__ definitions? Looks like they don't do anything compared to the parent super?

I'll check, but I think you're right, we don't need them.

examples/02-plot/backface_props.py
```diff
@@ -0,0 +1,84 @@
...
+Setting Backface Properties
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default front and backface rendering uses the same properties. In certain
+situations it can be useful to set different properties for backfaces than
+for frontfaces.
+
+One straightforward example is when a closed (or close enough) surface has a
+different color on the inside. Note that the notion of "inside" and "outside"
+depend on the orientation of the surface normals:
```
tkoyama010, line 13:
```suggestion
By default front and backface rendering uses the same properties. In certain
situations it can be useful to set different properties for backfaces than
for frontfaces.

One straightforward example is when a closed (or close enough) surface has a
different color on the inside. Note that the notion of "inside" and "outside"
depend on the orientation of the surface normals:
```

tests/plotting/test_plotting.py
```diff
@@ -2919,6 +2919,28 @@ def test_tight_wide():
     pl.show(before_close_callback=verify_cache_image)
 
 
+def test_backface_params():
+    mesh = pyvista.ParametricCatalanMinimal()
+
+    with pytest.raises(TypeError, match='pyvista.Property or a dict'):
+        mesh.plot(backface_params='invalid')
```
tkoyama010, line 2926:
```suggestion
    with pytest.raises(TypeError, match="pyvista.Property or a dict"):
        mesh.plot(backface_params="invalid")
```

tests/plotting/test_plotting.py
```diff
@@ -2919,6 +2919,28 @@ def test_tight_wide():
...
+
+    params = dict(color='blue', smooth_shading=True)
+    backface_params = dict(color='red', specular=1.0, specular_power=50.0)
+    backface_prop = pyvista.Property(**backface_params)
```
tkoyama010, line 2930:
```suggestion
    params = dict(color="blue", smooth_shading=True)
    backface_params = dict(color="red", specular=1.0, specular_power=50.0)
    backface_prop = pyvista.Property(**backface_params)
```

tkoyama010 (changes requested):

tkoyama010 (dismissed):
LGTM

tkoyama010:
Except for the topics that have yet to be resolved.

tests/plotting/test_plotting.py
tkoyama010: Feel free to revert it as it is my preference :)

pyvista/plotting/mapper.py
adeak: My local tests run fine without them, but I don't have a GPU, and neither will CI catch if anything is wrong here. Could you perhaps fire up the test suite? I'll push a commit to remove these inits and we'll see.

akaszynski:
I've raised a question regarding disappearing backfaces [here](https://discourse.vtk.org/t/backface-culling-with-pbr/9547), but I think this is a fundamental limitation of VTK and have made a note of it in this PR.

I also addressed textures, which I'm happy to report still works with backface properties.

akaszynski (dismissed):

akaszynski:
Thanks for working on this, and I'd love to have this out for the next release.

banesullivan (approved):

examples/02-plot/backface_props.py
```diff
@@ -0,0 +1,128 @@
...
+###############################################################################
+# Apply Backface Properties to Textured Meshes
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Backface textures can also be applied to meshes that have textures applied to
+# them. For this example we load the globe texture with
+# :func:`pyvista.examples.load_globe`, clip it, and then apply a different
+# color to the interior surface.
+#
+# The lighting has been disabled for this example to demonstrate how you can
+# make the interior of the surface appear occuluded without any directional
```
adeak, line 92:
```suggestion
# make the interior of the surface appear occluded without any directional
```

pyvista/plotting/mapper.py
```diff
@@ -25,17 +25,20 @@ class _BaseMapper(_vtk.vtkAbstractMapper):
 
     _theme = None
```
adeak, line 26: This came from `main` but I'm not sure why there's a class attr here when we're shadowing it in `__init__()` anyway.

adeak: Need not be a blocker though, feel free to ignore for now. Thanks @akaszynski for the improvements.

pyvista/plotting/mapper.py
```diff
@@ -25,17 +25,20 @@ class _BaseMapper(_vtk.vtkAbstractMapper):
...
 
-    def copy(self):
-        """Create a deep of this mapper.
+    def copy(self, deep=True) -> '_BaseMapper':
+        """Create a copy of this mapper.
+
+        Parameters
+        ----------
+        deep : bool, optional
+            Whether to return a deep or a shallow copy.
```
tkoyama010, line 38: The deep argument is not used. Will it be added in the future?

pyvista/plotting/mapper.py
akaszynski: That's here because we're using `no_new_attr` and without it we'll get:

```
    def __setattr__(self, name, value):
        """Do not allow setting attributes."""
        if hasattr(self, name):
            object.__setattr__(self, name, value)
        else:
>           raise AttributeError(
                f'Attribute "{name}" does not exist and cannot be added to type '
                f'{self.__class__.__name__}'
            )
E           AttributeError: Attribute "_theme" does not exist and cannot be added to 
type DataSetMapper                                                                  

pyvista/utilities/misc.py:176: AttributeError

```

pyvista/plotting/mapper.py
akaszynski: Good catch @tkoyama010. I've removed this in 48f0f1e5be8bd077481b6a56ea077002bf6599de.

akaszynski (dismissed):

examples/02-plot/backface_props.py
```diff
@@ -0,0 +1,128 @@
...
+# Note that backfaces are automatically culled when physically based rendering
+# is enabled, regardless of the settings of backface parameters.
+
+sphere = pv.Sphere()
+clipped_sphere = sphere.clip(normal='z', value=0.4)
+
+pl = pv.Plotter()
+pl.set_environment_texture(examples.download_sky_box_cube_map())
+pl.add_mesh(
+    clipped,
```
akaszynski, line 117:
```suggestion
    clipped_sphere,
```

akaszynski (approved):

PR #3138, opened 2022-08-08
Correct vtk-data url in _download_file docstring

PuffinDev:
### Overview

This corrects the urls in the `_download_file` function docstring.

### Details

The url supplied in the docstring is invalid because "/tree/" needs to be included to list directories.


akaszynski (approved):
Thanks for fixing this!

PR #3146, opened 2022-08-10
Fix filesystem races caused by pyvista.examples.downloads

### LINKED ISSUE #3145: Filesystem races in pyvista.examples.downloads

The function [pyvista.examples.downloads.delete_downloads](https://github.com/pyvista/pyvista/blob/5348b3dbe973fbd89bad8524bb946ec63ec2da12/pyvista/examples/downloads.py#L69) sometimes results in filesystem races when called from multiple processes (e.g. by pytest), and it's called indirectly [from the main module](https://github.com/pyvista/pyvista/blob/5348b3dbe973fbd89bad8524bb946ec63ec2da12/pyvista/__init__.py#L104), regardless if we actually download any examples.

The problem is that it first checks for the existence of `pyvista.EXAMPLES_PATH` and then tries to delete it, but by that time another process may have already done it. The same goes for creating the directory and then writing a file to it - there's all sorts of races happening.

I don't mind this module not being thread-safe, but I wish it wasn't loaded from the main module. I suggest moving the call to `_verify_cache_integrity` into `pyvista.examples.downloads`, where it's exclusively used.

I don't know how to easily reproduce it, but it makes our CI jobs sporadically fail.

### System Information

```shell
  Date: Wed Aug 10 16:09:18 2022 CEST

                OS : Linux
            CPU(s) : 16
           Machine : x86_64
      Architecture : 64bit
       Environment : Python
       GPU Details : error

  Python 3.8.13 (default, Aug 10 2022, 12:39:58)  [GCC 12.1.1 20220730]

           pyvista : 0.36.1
               vtk : 9.1.0
             numpy : 1.23.1
           imageio : 2.21.1
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.2
             scipy : 1.8.1
```

### Screenshots

_No response_

lukaszmoroz:
### Overview

This PR resolves #3145.

The solution is to move the call to `_verify_cache_integrity` inside `pyvista.examples.downloads`, where it's relevant. As long as the `examples` module (and subsequently `examples.downloads`) is not loaded, we will not interact with the filesystem.

### DIFF
--- a/pyvista/__init__.py
+++ b/pyvista/__init__.py
@@ -98,11 +98,6 @@ except Exception as e:
     )
     EXAMPLES_PATH = ''
 
-# verify example cache integrity
-from pyvista.examples.downloads import _verify_cache_integrity
-
-_verify_cache_integrity()
-
 # Send VTK messages to the logging module:
 send_errors_to_logging()
 
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -4828,3 +4828,7 @@ def download_sea_vase(load=True, progress_bar=False):  # pragma: no cover
     if load:
         return pyvista.read(filename)
     return filename
+
+
+# verify example cache integrity
+_verify_cache_integrity()


adeak:
Thanks for the report and the PR! I don't understand the underlying mechanics enough to review the PR, but here are some thoughts:

1. We should definitely avoid race conditions if possible.
2. Is it normal for your CI to delete the folder concurrently? Or to create the directory concurrently? If I understand correctly the issues caused by this are intermittent, so it's probably hard to figure out what happens exactly.
3. Are you sure this PR fixes your issues? Because as of not long ago when you `import pyvista` you already have access to `pyvista.examples` and through that `pyvista.examples.downloads`. So it looks like the race condition is only moved. I'm pretty sure this wasn't the case until fairly recenly, and you had to `import pyvista.examples` manually. But I haven't had the time to veryify this, and check if/when this changed.
4. The early failure on import might be by design: it might be better to handle issues early. But I wonder if we could instead _ignore_ errors arising from deletion/creation, hoping that this doesn't mask true errors. In any case I don't know the motivation behind this choice.

lukaszmoroz:
See also the associated issue.

> 2. Is it normal for your CI to delete the folder concurrently? Or to create the directory concurrently? If I understand correctly the issues caused by this are intermittent, so it's probably hard to figure out what happens exactly.

I'm running pytest on CI with [pytest-xdist](https://pypi.org/project/pytest-xdist/), which occasionally fails with:
```\
.../pyvista/examples/downloads.py:88: in delete_downloads
    shutil.rmtree(pyvista.EXAMPLES_PATH)
...
E   FileNotFoundError: [Errno 2] No such file or directory: '.../pyvista/examples'
```
So I assumed that the only possibility is that another process deleted the directory just after https://github.com/pyvista/pyvista/blob/5348b3dbe973fbd89bad8524bb946ec63ec2da12/pyvista/examples/downloads.py#L87 but before the actual operation. I haven't been able to reproduce this on my local machine in normal circumstances but maybe there is something with I/O on the CI that makes it more probable.

> 3. Are you sure this PR fixes your issues? Because as of not long ago when you `import pyvista` you already have access to `pyvista.examples` and through that `pyvista.examples.downloads`. So it looks like the race condition is only moved. I'm pretty sure this wasn't the case until fairly recenly, and you had to `import pyvista.examples` manually. But I haven't had the time to veryify this, and check if/when this changed.

The only reference to `pyvista.examples` I found in the code was what this PR removes - the module was loaded here: https://github.com/pyvista/pyvista/pull/3146/files#diff-af06c9c8a5782fc06cd3ec38c9fef4a2b30a7a941d05430871c35ace729bd959L102

With these changes you can no longer access `pyvista.examples` without importing it first. It is true that the race condition is moved to inside that module, but as I understand it, that whole example downloading machinery was not written with concurrency in mind and that's ok at least for me, as I don't use it.

> 4. The early failure on import might be by design: it might be better to handle issues early. But I wonder if we could instead _ignore_ errors arising from deletion/creation, hoping that this doesn't mask true errors. In any case I don't know the motivation behind this choice.

Ignoring errors was my first attempt at fixing this, but I realized it may not too reliable as we then try to write to a file inside the potentially deleted folder: https://github.com/pyvista/pyvista/blob/5348b3dbe973fbd89bad8524bb946ec63ec2da12/pyvista/examples/downloads.py#L65
so I think we'd need some locking mechanism.

If I understand the code correctly, `_verify_cache_integrity` can be run essentially whenever, as it only checks whether the structure of the cache folder hasn't changed between package versions, and removes the folder if it did. So whether it happens when importing `pyvista` or `pyvista.examples.downloads` shouldn't really matter, as we need to import the latter to download the examples anyway (and that's what the cache folder is used for).

adeak:
> > 2. Is it normal for your CI to delete the folder concurrently? Or to create the directory concurrently? If I understand correctly the issues caused by this are intermittent, so it's probably hard to figure out what happens exactly.
> 
> I'm running pytest on CI with [pytest-xdist](https://pypi.org/project/pytest-xdist/), which occasionally fails with:
> 
> ```
> .../pyvista/examples/downloads.py:88: in delete_downloads
>     shutil.rmtree(pyvista.EXAMPLES_PATH)
> ...
> E   FileNotFoundError: [Errno 2] No such file or directory: '.../pyvista/examples'
> ```
> 
> So I assumed that the only possibility is that another process deleted the directory just after
> 
> https://github.com/pyvista/pyvista/blob/5348b3dbe973fbd89bad8524bb946ec63ec2da12/pyvista/examples/downloads.py#L87
> but before the actual operation. I haven't been able to reproduce this on my local machine in normal circumstances but maybe there is something with I/O on the CI that makes it more probable.

I see, thanks. I also don't immediately see how this can happen. Perhaps what's going on is that multiple processes are running in a clean environment (with no `examples` directory), one creates the example directory that is empty, without a cache version file they all try to delete the examples directory, and only the first one succeeds.

> > 3. Are you sure this PR fixes your issues? Because as of not long ago when you `import pyvista` you already have access to `pyvista.examples` and through that `pyvista.examples.downloads`. So it looks like the race condition is only moved. I'm pretty sure this wasn't the case until fairly recenly, and you had to `import pyvista.examples` manually. But I haven't had the time to veryify this, and check if/when this changed.
> 
> The only reference to `pyvista.examples` I found in the code was what this PR removes - the module was loaded here: https://github.com/pyvista/pyvista/pull/3146/files#diff-af06c9c8a5782fc06cd3ec38c9fef4a2b30a7a941d05430871c35ace729bd959L102
> 
> With these changes you can no longer access `pyvista.examples` without importing it first.
> 
> It is true that the race condition is moved to inside that module, but as I understand it, that whole example downloading machinery was not written with concurrency in mind and that's ok at least for me, as I don't use it.

Aaah, I see. This is exactly the change that triggered eager import of `pyvista.examples`! Thanks.

This was added in https://github.com/pyvista/pyvista/pull/2727, and based on this I'm pretty sure this eager importing change was not on purpose. @akaszynski can tell for certain.

> > 4. The early failure on import might be by design: it might be better to handle issues early. But I wonder if we could instead _ignore_ errors arising from deletion/creation, hoping that this doesn't mask true errors. In any case I don't know the motivation behind this choice.
> 
> That code wasn't mean to fail in any case, it's only the concurrent access that causes the failure.

What I meant was that if someone has a broken examples setup. But now I see that "nonsense examples directory" is handled in `__init__.py` anyway, so this point is moot.

> Ignoring errors (and e.g. turning the "if exists then delete" pattern into handling a `FileNotFoundError`) was my first attempt at fixing this, but I realized it may not too reliable as we then try to write to a file inside the potentially deleted folder:
> 
> https://github.com/pyvista/pyvista/blob/5348b3dbe973fbd89bad8524bb946ec63ec2da12/pyvista/examples/downloads.py#L65
> 
> so I think we'd need some locking mechanism anyway.
> 
> If I understand the code correctly, `_verify_cache_integrity` can be run essentially whenever, as it only checks whether the structure of the cache folder hasn't changed between package versions, and removes the folder if it did (which forces the examples to be redownloaded). So whether it happens when importing `pyvista` or `pyvista.examples.downloads` shouldn't really matter, as we need to import the latter to download the examples anyway (and that's where the folder is solely used).

I think you're right. We've also had [a recent issue](https://github.com/pyvista/pyvista/pull/3048) related to this machinery. It would be nice to make this overall more robust. I agree that the current implementation has sequential execution in mind, and it wouldn't be easy to make it thread-safe. It doesn't help that end-users with existing caches aren't affected, these issues mostly arise in CI.

akaszynski:
> This was added in #2727, and based on this I'm pretty sure this eager importing change was not on purpose. @akaszynski can tell for certain.

It's been 6 weeks, but from the looks of it I wanted to explicitly verify the example cache within the `__init__`, but I think it's at least better having this within the examples module.

 > It is true that the race condition is moved to inside that module, but as I understand it, that whole example downloading machinery was not written with concurrency in mind and that's ok at least for me, as I don't use it.

That's something we should resolve at some point in the future. Running the CI the first time with `pytest-xdist` causes a handful of examples to fail on an empty cache due to concurrency.

akaszynski (approved):
Agree with the changes. I know we're kicking the can down the road a bit, but if it helps with `pytest-xdist` on CI and we only verify the image cache on the import of `pytest.examples`, I agree with this change.

PR #3115, opened 2022-08-01
Add CellType enumerate class

tkoyama010:
### Overview

Inspired by https://github.com/pyvista/pyvista/pull/3101#discussion_r934130439
Add CellType enumerate class.

#### main
```python
import vtk


print(vtk.VTK_HEXAHEDRON)
```
#### PR
```python
import pyvista


print(pyvista.CellType.HEXAHEDRON)
```

 

### Details

- [x] Replace vtk cell_type expression to PyVista CellType enum expression.
- [x]  ~Follow the documentation style in #3150.~

PR #1040, opened 2020-11-29
Add Light class pythonizing vtkLight

adeak:
### Overview

This PR adds a `pyvista.Light` class which is a pythonic subclass of `vtk.vtkLight`. Originally suggested in https://github.com/pyvista/pyvista/issues/974.

### Details

Main features:

  1. `Light` is a `vtk.vtkLight` subclass, pyvista objects return `pyvista.Light`s now. A `pyvista.Light.from_vtk` classmethod facilitates converting from base class instances (by copying).
  2. New `LightType` enum (class "constants") with pretty printing.
  3. `add_light()`, `remove_all_lights()` on `Plotter`s and `Renderer`s. For the former class a new kwarg specifies if each renderer should be affected or only the active one. Also new `lighting` `__init__` kwarg for `Plotter`s, in order to enable starting a plotter without any lights.

My example in https://github.com/pyvista/pyvista/issues/974#issuecomment-725972305 now looks like this:
```py
import pyvista as pv
from pyvista.examples import download_nefertiti

data = download_nefertiti()

plotter = pv.Plotter(lighting=None)
plotter.add_mesh(data, color='white', smooth_shading=True)
plotter.set_background('black')

light_positions = [(-2, 2, 0), (2, -2, 0)]
light_colors = [(1, 0, 0), (0, 0, 1)]

for light_pos, light_color in zip(light_positions, light_colors):
    light = pv.Light(position=light_pos, color=light_color)
    light.focal_point = (0, 0, 0)
    
    plotter.add_light(light)

plotter.show(cpos=[(-600, -850, 200), (0, 0, 0), (0, 0, 1)], screenshot='nefertiti_lit_upgrade.png')
```
(although to be fair the cone angle setting was redundant in my original snippet).

---

The lights and the light types have pretty `repr`s:
```py
>>> import pyvista as pv
>>> plotter = pv.Plotter()  # default: `vtkLightKit` with 5 lights, converted to `pyvista.Light`s
>>> plotter.renderer.lights
[<pyvista.Light (Headlight) at 0x7ffa114f96a0>, <pyvista.Light (Camera Light) at 0x7ffa114f98e0>, <pyvista.Light (Camera Light) at 0x7ffa114f9940>, <pyvista.Light (Camera Light) at 0x7ffa114f99a0>, <pyvista.Light (Camera Light) at 0x7ffa114f9a00>]
>>> light = plotter.renderer.lights[0]
>>> light.light_type
<LightType.HEADLIGHT: 1>
[...]

PR #3150, opened 2022-08-11
Expose more options of `MultiBlockPlot3DReader` (`vtkMultiBlockPlot3DReader`)

whophil:
Expose options in `MultiBlockPlot3DReader` related to _functions_ - derived scalar/vector fields computed from the point data living in Plot3D files.

This PR adds an enum to map useful names to their int values required by the `add_function()` method. There may be a better place for this enum to live.

There are still many methods from the `vtkMultiBlockPlot3DReader` available I chose not to expose, mainly relating to the format of the target file. The VTK documentation states that most binary files can be read without calling these methods, so long as "auto-detect" is on - in PyVista, it is by default. Thus, for simplicity, I chose not to expose the many reading flags. A user can still set them directly on the VTK reader instance if necessary.

### DIFF
--- a/pyvista/utilities/reader.py
+++ b/pyvista/utilities/reader.py
@@ -1250,7 +1250,7 @@ class Plot3DMetaReader(BaseReader):
 
 
 class Plot3DFunctionEnum(enum.IntEnum):
-    """An enumeration for the functions used in `MultiBlockPlot3DReader`."""
+    """An enumeration for the functions used in :class:`MultiBlockPlot3DReader`."""
 
     DENSITY = 100
     PRESSURE = 110
@@ -1276,8 +1276,8 @@ class Plot3DFunctionEnum(enum.IntEnum):
 class MultiBlockPlot3DReader(BaseReader):
     """MultiBlock Plot3D Reader.
 
-    The methods :meth:``add_function()`` and :meth:``remove_function()`` accept values from
-    :class:``Plot3DFunctionEnum``. For convenience, the values of that enumeration are available as class variables,
+    The methods :meth:`add_function()` and :meth:`remove_function()` accept values from
+    :class:`Plot3DFunctionEnum`. For convenience, the values of that enumeration are available as class variables,
     as shown below.
 
         - ``MultiBlockPlot3DReader.DENSITY = Plot3DFunctionEnum.DENSITY``
--- a/pyvista/utilities/reader.py
+++ b/pyvista/utilities/reader.py
@@ -1382,7 +1382,9 @@ class MultiBlockPlot3DReader(BaseReader):
         Example
         -------
         >>> import pyvista
-        >>> reader = pyvista.reader.MultiBlockPlot3DReader('grid.x')
+        >>> from pyvista import examples
+        >>> filename, _  = examples.downloads._download_file('multi-bin.xyz')
+        >>> reader = pyvista.reader.MultiBlockPlot3DReader(filename)
         >>> reader.add_function(112)  # add a function by its integer value
         >>> reader.add_function(reader.PRESSURE_COEFFICIENT)  # add a function by enumeration via class variable alias
 
--- a/pyvista/utilities/reader.py
+++ b/pyvista/utilities/reader.py
@@ -1274,9 +1274,54 @@ class Plot3DFunctionEnum(enum.IntEnum):
 
 
 class MultiBlockPlot3DReader(BaseReader):
-    """MultiBlock Plot3D Reader."""
+    """MultiBlock Plot3D Reader.
+
+    The methods ``add_function()`` and ``remove_function()`` accept values from :class:``Plot3DFunctionEnum``.
+    For convenience, the values of that enumeration are available as class variables, as shown below.
+
+    - ``MultiBlockPlot3DReader.DENSITY = Plot3DFunctionEnum.DENSITY``
+    - ``MultiBlockPlot3DReader.PRESSURE = Plot3DFunctionEnum.PRESSURE``
+    - ``MultiBlockPlot3DReader.PRESSURE_COEFFICIENT = Plot3DFunctionEnum.PRESSURE_COEFFICIENT``
+    - ``MultiBlockPlot3DReader.MACH = Plot3DFunctionEnum.MACH``
+    - ``MultiBlockPlot3DReader.SPEED_OF_SOUND = Plot3DFunctionEnum.SPEED_OF_SOUND``
+    - ``MultiBlockPlot3DReader.TEMPERATURE = Plot3DFunctionEnum.TEMPERATURE``
+    - ``MultiBlockPlot3DReader.ENTHALPY = Plot3DFunctionEnum.ENTHALPY``
+    - ``MultiBlockPlot3DReader.INTERNAL_ENERGY = Plot3DFunctionEnum.INTERNAL_ENERGY``
+    - ``MultiBlockPlot3DReader.KINETIC_ENERGY = Plot3DFunctionEnum.KINETIC_ENERGY``
+    - ``MultiBlockPlot3DReader.VELOCITY_MAGNITUDE = Plot3DFunctionEnum.VELOCITY_MAGNITUDE``
+    - ``MultiBlockPlot3DReader.STAGNATION_ENERGY = Plot3DFunctionEnum.STAGNATION_ENERGY``
+    - ``MultiBlockPlot3DReader.ENTROPY = Plot3DFunctionEnum.ENTROPY``
+    - ``MultiBlockPlot3DReader.SWIRL = Plot3DFunctionEnum.SWIRL``
+    - ``MultiBlockPlot3DReader.VELOCITY = Plot3DFunctionEnum.VELOCITY``
+    - ``MultiBlockPlot3DReader.VORTICITY = Plot3DFunctionEnum.VORTICITY``
+    - ``MultiBlockPlot3DReader.MOMENTUM = Plot3DFunctionEnum.MOMENTUM``
+    - ``MultiBlockPlot3DReader.PRESSURE_GRADIENT = Plot3DFunctionEnum.PRESSURE_GRADIENT``
+    - ``MultiBlockPlot3DReader.STRAIN_RATE = Plot3DFunctionEnum.STRAIN_RATE``
+    - ``MultiBlockPlot3DReader.VORTICITY_MAGNITUDE = Plot3DFunctionEnum.VORTICITY_MAGNITUDE``
+    """
 
     _class_reader = staticmethod(_vtk.lazy_vtkMultiBlockPLOT3DReader)
+
+    # pull in function name enum values as class constants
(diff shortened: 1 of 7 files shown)


MatthewFlamm:
PyVista doesn't have many (or any?) `Enum`s.  Now there are two PRs with #3115 .  Most selections for functions, like in this PR, have typically used strings, which isn't necessarily desirable.  But, I'm wary that using an `IntEnum` in this PR is going to break consistency.  We could consider using `StrEnum` for the existing string usage and for the one here, but it doesn't come until Python 3.11!  We can also inherit directly from `str` to get similar features as `StrEnum` in the meantime.

I think we should have some consistent usage across the code base, whether it is plain strings, `Enum`s that are all `StrEnum`-like, or `Enum`s that are various types.  One knock for `Enum` usage is that these objects will start clogging the namespace in PyVista.

If we decide to keep the `Enum` usage, can it be part of this class?  This way it wouldn't pollute the namespace and this one can only be used with this class.

adeak:
> PyVista doesn't have many (or any?) `Enum`s.

For the record, I added `LightType` at https://github.com/pyvista/pyvista/blob/main/pyvista/plotting/lights.py#L18 via https://github.com/pyvista/pyvista/pull/1040.

tkoyama010:
> We could consider using StrEnum for the existing string usage and for the one here, but it doesn't come until Python 3.11!

I think we can wait until 3.11 arrives and move it to `StrEnum` when it comes. I am +100 that namespace should be managed by str. However, I don't see a problem with IntEnum since VTK guarantees that it will not duplicate it. I think the current policy is to inherit VTK's Int as it is.

whophil:
> If we decide to keep the `Enum` usage, can it be part of this class? This way it wouldn't pollute the namespace and this one can only be used with this class.

@MatthewFlamm as a nested class? Or something else?

MatthewFlamm:
> > If we decide to keep the `Enum` usage, can it be part of this class? This way it wouldn't pollute the namespace and this one can only be used with this class.
> 
> @MatthewFlamm as a nested class? Or something else?

I was thinking the class could be stored as a class or instance attribute.  So, something like:

```
reader.add_function(MultiBlockPlot3DReader.FunctionEnum.DENSITY)
# or
reader.add_function(reader.FunctionEnum.DENSITY)
```

But having written it out, I'm not sure anymore.  It was more of a thought for discussion than a request for change.

whophil:
@MatthewFlamm definitely worth the discussion.

> `MultiBlockPlot3DReader.FunctionEnum.DENSITY`

I believe this syntax implies a nested class. I don't know if we do that anywhere else in Pyvista. In general, I don't see it done much in any Python projects, but it does minimize the namespace clutter in this case.

```python
class MultiBlockPlot3DReader(BaseReader):
    """MultiBlock Plot3D Reader."""

        class FunctionEnum(enum.IntEnum):
            PRESSURE = 110
            ...
```

adeak:
Just to plug my `Lights` again, we have the enum
https://github.com/pyvista/pyvista/blob/690acb8311b22943090091c4fe2283a5cab2f9bd/pyvista/plotting/lights.py#L18-L23

But then its values are pulled into the `Lights` class namespace:
https://github.com/pyvista/pyvista/blob/690acb8311b22943090091c4fe2283a5cab2f9bd/pyvista/plotting/lights.py#L122-L125

It would obviously be messier with a dozen enum values, but we could still do this to have `MultiBlockPlot3DReader.DENSITY` etc. Odds are the semantic scope of `MultiBlockPlot3DReader` is small enough that the enum values don't stand out too much.

MatthewFlamm:
Having talked it through I think I agree that nesting it into the reader might be too eager.  I think the current form is fine since it doesn't clog up the top-level namespace.

pyvista/utilities/reader.py
```diff
@@ -1248,6 +1249,30 @@ class Plot3DMetaReader(BaseReader):
     _class_reader = staticmethod(_vtk.lazy_vtkPlot3DMetaReader)
 
 
+class Plot3DFunctionEnum(enum.IntEnum):
```
MatthewFlamm, line 1252: Does this need to be added to the documentation somehow?  I think it will be missing, but I haven't built the docs to check for this PR.

MatthewFlamm, line 1252: Should we rename this to match more closely with the reader name?  Maybe it is too long though?

pyvista/utilities/reader.py
```diff
@@ -1295,6 +1320,68 @@ def auto_detect_format(self):
     def auto_detect_format(self, value):
         self.reader.SetAutoDetectFormat(value)
 
+    def add_function(self, value: Union[int, Plot3DFunctionEnum]):
+        """Specify additional functions to read.
+
+        Parameters
+        ----------
+        value
```
MatthewFlamm, line 1328: I think we need a description of the mapping from int to Enum, or a direct link to the Enum documentation which might suffice.

pyvista/utilities/reader.py
```diff
@@ -1295,6 +1320,68 @@ def auto_detect_format(self):
...
+            An integer or `Plot3DFunctionEnum`.
+        """
```
MatthewFlamm, line 1330: An example would be useful, in particular it won't be obvious that the user has to use `pyvista.reader.Plot3DFunctionEnum`.

pyvista/utilities/reader.py
```diff
@@ -1295,6 +1320,68 @@ def auto_detect_format(self):
...
+            value = value.value
+        self.reader.AddFunction(value)
+
+    def remove_function(self, value: Union[int, Plot3DFunctionEnum]):
+        """Remove one function from list of functions to read.
+
+        Parameters
+        ----------
+        value
+            An integer or `Plot3DFunctionEnum`.
```
MatthewFlamm, line 1341: Same as above.

tests/utilities/test_reader.py
```diff
@@ -384,22 +384,47 @@ def test_multiblockplot3dreader():
...
     reader.add_q_files(q_filename)
+    reader.add_function(112)  # add by int
+    reader.add_function(pyvista.reader.Plot3DFunctionEnum.PRESSURE_GRADIENT)  # add by enum
     mesh = reader.read()
     for m in mesh:
         assert len(m.array_names) > 0
 
+    assert 'MachNumber' in mesh[0].point_data
+    assert 'PressureGradient' in mesh[0].point_data
+
```
MatthewFlamm, line 395: Can we add tests for the remove functionality by int and Enum?

pyvista/utilities/reader.py
adeak: Just in case you're unaware, you can download the build artifact on successful CI doc builds via the checks tab -> build documentation (on the left) -> docs-build (under Artifacts), in this case https://github.com/pyvista/pyvista/suites/7764071910/artifacts/326430707

whophil: The class name would be `MultiBlockPlot3DReaderFunctionEnum`. Happy to change it if that's preferred.

tests/utilities/test_reader.py
whophil: Good catch, now added!

pyvista/utilities/reader.py
MatthewFlamm: Too long I think, let's keep the one you have.

whophil:
Unfortunately, I can't build the docs locally on Mac to understand what is going on with the doc build.

I think it should be clear what I was _trying_ to achieve with the docs - creating a new section for reader enumerations, and then documenting the enumerations using [enum_tools](https://enum-tools.readthedocs.io/en/latest/).

I would appreciate help with the docs, CC @akaszynski @tkoyama010

akaszynski:
> Unfortunately, I can't build the docs locally on Mac to understand what is going on with the doc build.
> 
> I think it should be clear what I was _trying_ to achieve with the docs - creating a new section for reader enumerations, and then documenting the enumerations using [enum_tools](https://enum-tools.readthedocs.io/en/latest/).

Fixed in 912bfc3044ba627abeab1b76e078c9509187448c.

whophil:
Thanks @akaszynski ! After adding some numpydoc excludes, the docs are building now. However, the enum docs are added to the top level readers page, as shown below. Not sure if this is the preferred way of documenting this...

image

tkoyama010:
Great documentation. I should follow it in #3150 .

whophil:
I go back and forth on whether aliasing the enum members as class variables actually makes things clearer. I am open to suggestions.

whophil:
Looking for some final feedback on:
- If we want to keep the enum members inside the reader class
- If the documentation for the enum is satisfactory

Thanks @MatthewFlamm @adeak @akaszynski !

MatthewFlamm:
> Looking for some final feedback on:
> 
> * If we want to keep the enum members inside the reader class
> * If the documentation for the enum is satisfactory
> 
> Thanks @MatthewFlamm @adeak @akaszynski !

I'm not a user of this format, but it seems like these function settings would be an integral part of using this file format.  If that is an accurate representation then I am +1 on having them be aliased inside the reader class.  If it is a niche usage, I would swing slightly the other way.

requirements_docs.txt
```diff
@@ -1,6 +1,7 @@
 Sphinx==4.5.0
 cmocean==2.0
 colorcet==3.0.0
+enum_tools
```
adeak, line 4: Just a drive-by comment:
1. Do we really need to add `enum_tools` and `sphinx-toolbox`? (Asking just because we should have very good reasons for adding additional dependencies, even if just for doc building.)
2. We should pin these dependencies too and let dependabot bump the versions as they are released.

akaszynski: I don't see where `sphinx-toolbox` is used. I'm fine with adding `enum_tools` provided we pin the dependency.

whophil: `sphinx-toolbox` appears to be an unlisted dependency of `enum_tools` (the docs build failed without it). I can pin the version numbers, though.

whophil:
> I'm not a user of this format, but it seems like these function settings would be an integral part of using this file format. If that is an accurate representation then I am +1 on having them be aliased inside the reader class. If it is a niche usage, I would swing slightly the other way.

@MatthewFlamm in my opinion, the functions are likely to be used by anybody who reads in Plot3D files.

requirements_docs.txt
tkoyama010: Now it seems v0.9.0 (What is 0.9.0.post1?).
```suggestion
enum-tools==0.9.0
```

akaszynski (approved):
Made one minor commit to improve coverage, but otherwise this LGTM.

MatthewFlamm (approved):
Also LGTM.

akaszynski:
Thanks for rerunning the CI. I'll merge once it's green. Sorry, been busy with other PRs.

whophil:
Thanks @akaszynski @MatthewFlamm @adeak !

PR #3155, opened 2022-08-11
More descriptive error message when no geodesic path exists between vertices on a mesh

### LINKED ISSUE #1734: More descriptive error message when there is no geodesic path between two vertices
**Describe what feature you would like added.**

Thank you for making pyvista! I'm finding it very useful. I just have a small suggestion.

As discussed in https://github.com/pyvista/pyvista-support/issues/233#issuecomment-676517696, if `start_vertex` and `end_vertex` are in disconnected regions, `PolyData.geodesic` and `PolyData.geodesic_distance` will raise the following exception:

`ValueError: Number of scalars (1) must match either the number of points (0) or the number of cells (0).`

A better message might be something like:

`f"There is no path between vertices {start_vertex} and {end_vertex}."`

Perhaps with a custom exception too, such as `NoPathError`?

--------
Code to reproduce:

```
import pyvista

# Create two disconnected triangles
points = [
    [0, 0, 0],
    [1, 0, 0],
    [0, 1, 0],
    [2, 2, 0],
    [2, 3, 0],
    [3, 2, 0],
]
faces = [
    3, 0, 1, 2,
    3, 3, 4, 5,
]

mesh = pyvista.PolyData(points, faces)

# Find the shortest path between vertices on the disconnected triangles
mesh.geodesic(start_vertex=0, end_vertex=4)
```

p-j-smith:
### Overview

resolves #1734

* If there is no geodesic path between two vertices, `PolyData.geodesic` and `PolyData.geodesic_distance` now raise clearer error messages.
* Added tests to check error message is raised

### Details

Previously, the error raised was:
```
Number of scalars (1) must match either the number of points (0) or the number of cells (0).
```

This was raised by trying to assign the indices of points in the path (i.e. the start index) to the output mesh, which has zero points and cells.

Now the following error message is raised:
```
f"There is no path between vertices {start_vertex} and {end_vertex}."
```

### DIFF
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -1903,7 +1903,13 @@ class PolyDataFilters(DataSetFilters):
         original_ids = vtk_id_list_to_array(dijkstra.GetIdList())
 
         output = _get_output(dijkstra)
-        output["vtkOriginalPointIds"] = original_ids
+
+        try:
+            output["vtkOriginalPointIds"] = original_ids
+        except ValueError as e:
+            if str(e) != 'Number of scalars (1) must match either the number of points (0) or the number of cells (0).':
+                raise
+            raise ValueError(f"There is no path between vertices {start_vertex} and {end_vertex}.")
 
         # Do not copy textures from input
         output.clear_textures()
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -1903,13 +1903,12 @@ class PolyDataFilters(DataSetFilters):
         original_ids = vtk_id_list_to_array(dijkstra.GetIdList())
 
         output = _get_output(dijkstra)
+        if output.n_points == 0:
+            raise ValueError(f"There is no path between vertices {start_vertex} and {end_vertex}. ",
+                "It is likely the vertices belong to disconnected regions."
+            )
 
-        try:
-            output["vtkOriginalPointIds"] = original_ids
-        except ValueError as e:
-            if str(e) != 'Number of scalars (1) must match either the number of points (0) or the number of cells (0).':
-                raise
-            raise ValueError(f"There is no path between vertices {start_vertex} and {end_vertex}.")
+        output["vtkOriginalPointIds"] = original_ids
 
         # Do not copy textures from input
         output.clear_textures()
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -1904,8 +1904,9 @@ class PolyDataFilters(DataSetFilters):
 
         output = _get_output(dijkstra)
         if output.n_points == 0:
-            raise ValueError(f"There is no path between vertices {start_vertex} and {end_vertex}. ",
-                "It is likely the vertices belong to disconnected regions."
+            raise ValueError(
+                f"There is no path between vertices {start_vertex} and {end_vertex}. ",
+                "It is likely the vertices belong to disconnected regions.",
             )
 
         output["vtkOriginalPointIds"] = original_ids
--- a/tests/test_polydata.py
+++ b/tests/test_polydata.py
@@ -7,6 +7,7 @@ import pytest
 
 import pyvista
 from pyvista import examples
+from pyvista._vtk import VTK9
 from pyvista.core.errors import NotAllTrianglesError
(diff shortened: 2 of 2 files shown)


pyvista/core/filters/poly_data.py
```diff
@@ -1903,7 +1903,13 @@ def geodesic(
         original_ids = vtk_id_list_to_array(dijkstra.GetIdList())
 
         output = _get_output(dijkstra)
-        output["vtkOriginalPointIds"] = original_ids
+
+        try:
+            output["vtkOriginalPointIds"] = original_ids
+        except ValueError as e:
+            if str(e) != 'Number of scalars (1) must match either the number of points (0) or the number of cells (0).':
```
adeak, line 1910: Firstly, I think there might be an attribute on exceptions with the message.
Secondly, I think the real issue is `output` being an empty mesh. We should check for that I think, and raise. Maybe phrase the error as "probably not connected"  or similar.

tests/test_polydata.py
```diff
@@ -857,3 +857,19 @@ def test_n_verts():
 def test_n_lines():
     mesh = pyvista.Line()
     assert mesh.n_lines == 1
+
+
+def test_geodesic_error_message(sphere, sphere_shifted):
+
```
adeak, line 863: I'd slightly prefer a semantically oriented name, e.g.
```suggestion
def test_geodesic_disconnected(sphere, sphere_shifted):
```
I'd also remove the empty line (github doesn't seem to show this in this edit suggestion for some reason).

tests/test_polydata.py
```diff
@@ -857,3 +857,19 @@ def test_n_verts():
...
+    combined = (
+        sphere + sphere_shifted
+    )  # the sphere and sphere_shifted are disconnected - no path between them
```
adeak, line 866: It would be nice to get this on one line. Either move the comment before the assignment, or just rephrase to something like
```suggestion
    combined = sphere + sphere_shifted  # disconnected meshes
```

adeak (commented):
Some minor suggestions, but it overall looks good to me. CI is failing on VTK 8.1.2, because that version will happily find a "path":
tmp
this is an obvious bug and we'll have to add a VTK version based skip in the file to ignore this test.

tests/test_polydata.py
p-j-smith: Yeah that's much clearer

tests/test_polydata.py
p-j-smith: I've moved the comment above the assignment

p-j-smith:
> this is an obvious VTK bug and we'll have to add a VTK version based skip in the file to ignore this test.
I've skipped the test if vtk < 9. What do you think about `PolyData.geodesic` raising a warning that the path may connect disconnected regions if the user has vtk < 9?

adeak (approved):
This gave me an idea for something that might need fixing, but that's beyond the scope of this PR anyway. This looks good to me now, thanks!

>  What do you think about `PolyData.geodesic` raising a warning that the path may connect disconnected regions if the user has vtk < 9?

I wondered about that myself. I might even go so far and try to detect buggy cases for VTK < 9, but I'm not sure it's worth the complexity because VTK 9 came out more than two years ago. Just warning might be enough.

But anyway in my experiment from which I posted the screenshot this is what the result looked like:
```py
>>> geod['vtkOriginalPointIds']
pyvista_ndarray([   0,   -1, 1683])
```
So we could probably do an additional check for VTK < 9 if there are any negative indices in there (this would be correct for all VTK versions, but I wouldn't want to add overhead for an upstream bug that's been fixed already).

p-j-smith:
>  This looks good to me now, thanks!

great, thank you!

> Just warning might be enough.

How about adding a note to the docstring about the bug in vtk8? That way there's no overhead of checking the version

> I wouldn't want to add overhead for an upstream bug that's been fixed already

Yeah good point, it's probably unnecessary

adeak:
> That way there's no overhead of checking the version

The version checking is free. The overhead I meant is mostly the runtime of checking every value in `vtkOriginalPointsId`, and the maintenance burden of adding complexity to work around a bug in 8.2. That being said, we can return early in the VTK 9 case, and then whatever is left in the method is for 8.2.

akaszynski:
Merging despite doc failure. Will fix issue with cache reset in a followup PR.

PR #3160, opened 2022-08-13
Add area to DataSet and make pointgrid volume exact

### LINKED ISSUE #3158: Volume property for UnstructuredGrid only returns an estimate of the volume

To determine the volume of UnstructuredGrid grids, pyvista first extracts the outer surface, triangulates it, then uses vtkMassProperties, which provides an _estimate_ of the volume.  See https://vtk.org/doc/nightly/html/classvtkMassProperties.html#details

And

https://vtk.org/doc/nightly/html/classvtkMassProperties.html#details

In some cases the current behavior is fine, but can lead to inaccuracies for complex shapes.  For example:

```py
from pyvista import examples
mesh = examples.download_can_crushed_hdf()
ivol = mesh.integrate_data()["Volume"][0]
print((ivol - mesh.volume)/ivol)
```

Shows more than 3% error.

### System Information

```shell

           pyvista : 0.37.dev0
               vtk : 9.1.0
             numpy : 1.22.4
           imageio : 2.19.3
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.2
           IPython : 8.4.0
          colorcet : 3.0.0
           cmocean : 2.0
        ipyvtklink : 0.2.2
             scipy : 1.8.1
              tqdm : 4.64.0
            meshio : 5.3.4
        jupyterlab : 3.4.4
         pythreejs : Version unknown
```

### Screenshots

_No response_

MatthewFlamm:
### Overview

 

Adds area property to DataSet.  Removes special case for volume for PointGrid (see below)

Original PR:

`PointGrid`s like `UnstructuredGrid` can be 2D and thus have an area rather than a volume.  This PR adds an area property.

This PR also aligns the volume property to use a non approximate method.  My testing is that this is also faster.

### Details

Closes #3158


pyvista/core/pointset.py
```diff
@@ -1280,11 +1280,18 @@ def plot_curvature(self, curv_type='mean', **kwargs):
...
 
-        This extracts the external surface and computes the interior
-        volume.
+        .. versionchanged:: 0.37.0
+           This now calculates the sum of the volume of all cells.
+
         """
-        surf = self.extract_surface().triangulate()
-        return surf.volume
+        vols = self.compute_cell_sizes(length=False, area=False, volume=True)
```
adeak, line 1287: If the inaccuracy comes from `vtkMassProperties`, do you think we could/should replace that for `PolyData` as well?

pyvista/core/pointset.py
```diff
@@ -1280,11 +1280,18 @@ def plot_curvature(self, curv_type='mean', **kwargs):
...
+        return np.sum(vols["Volume"])
+
+    @property
+    def area(self) -> float:
+        """Compute the area of the point grid if 2D."""
```
adeak, line 1292: This method wasn't clear to me on multiple points, but admittedly because I also hadn't realised how `compute_cell_sizes` worked.

1. It wasn't clear what "if 2D" meant: how would it behave if the grid is not 2D? I now know that we just get zero.
2. When talking about "area" I would have assumed (coming from a `PolyData` mindset) that this would also work for space-filling grids, by extracting the surface. I see that this is not the case. Do you think this would make sense? It might make the semantics harder to argue about ("if 2D grid, gives you the total cell area, otherwise the area of the extracted surface"). Especially since for a 2d grid we only count one "side", whereas for a 3d grid we'd have all sides included.
3. It might help clear some of the confusion by adding `Returns` and `Examples` blocks to these docstrings. Maybe show the volume/area of the same 2D and 3D grids. That might prevent some user confusion.

pyvista/core/pointset.py
```diff
@@ -1280,11 +1280,18 @@ def plot_curvature(self, curv_type='mean', **kwargs):
...
+        vols = self.compute_cell_sizes(length=False, area=True, volume=False)
+        return np.sum(vols["Area"])
```
adeak, line 1294: We're talking about microseconds here overall, but invoking the method on the ndarray subclass seems to be 30% faster (on a 2d grid with 10k cells):
```suggestion
        return vols["Area"].sum()
```
(And the same for `volume()`).

tests/test_pointset.py
```diff
@@ -62,3 +62,15 @@ def test_filters_return_pointset(sphere):
     pointset = sphere.cast_to_pointset()
     clipped = pointset.clip()
     assert isinstance(clipped, pyvista.PointSet)
+
+
+def test_volume_area_pointgrid():
+    # 3D grid size 4x4x4
+    grid = pyvista.UniformGrid(dims=(5, 5, 5)).cast_to_unstructured_grid()
```
adeak, line 69: Sorry, one more question: do I understand correctly that these grids are converted to unstuctured grids because `UniformGrid` doesn't inherit from `PointGrid`? But I see the `compute_cell_sizes()` method on `UniformGrid`s. Couldn't we give them the same properties, then? Presumably somewhere higher up the inheritance tree.

pyvista/core/pointset.py
MatthewFlamm: Since PolyData cells never fill a volume, i.e
it is a 2D surface that sometimes encloses a volume, I think this is the only option (although I'm not much of a PolyData user).  `area` for `PolyData` is calculated using the `compute_cell_sizes` as in this PR.

tests/test_pointset.py
MatthewFlamm: Despite grid being in the name of both, the common ancestor is `DataSet`.  But you are right this PR aligns a lot of these methods and we can probably move all these up to `DataSet`.

pyvista/core/pointset.py
adeak: For a moment I got confused between `triangulate()` and Delaunay triangulation! We do the former but not the latter. If we did Delaunay then we could compute exact sizes. But in the general case if we only know the surface is closed, we can't easily compute a 3d triangulation (because the surface could easily be concave). Considering concave surface meshes instantly explains to me why we can only use an approximation.

At best we could leave a note for `PolyData.volume` that this is an approximation, and even that would be outside this PR. I think we're good here, thanks.

tkoyama010:
I will update to solve the documentation build failure fixed by https://github.com/pyvista/pyvista/pull/3162 .

MatthewFlamm:
In standardizing these functionality, I have found that we will provide a volume for PolyData that isn't closed.

```
import pyvista as pv
mesh = pv.Sphere(end_theta=90)
mesh.volume
```

I don't know whether this is expected or not, but I won't be touching this behavior in this PR.

adeak:
> I don't know whether this is expected or not, but I won't be touching this behavior in this PR.

The docs says
> This will throw a VTK error/warning if not a closed surface.

And volume trivially makes no sense for an open surface. I think all bets are off when a surface is open. But it's fine if you don't want to touch this here.

pyvista/core/pointset.py
MatthewFlamm: Please check new docstring.  PolyData is the odd type out here.  I also added a note about how to obtain the outer surface area of a mesh with 3D cells.

pyvista/core/dataset.py
```diff
@@ -1743,23 +1743,78 @@ def center(self) -> Vector:
...
+        Note 5 points in each direction.
+
+        >>> import pyvista as pv
+        >>> mesh = pv.UniformGrid(dims=(5, 5, 1))
+        >>> mesh.area
+        16.0
+
+        A mesh with 3D cells does not have an area.  To get
+        the outer surface area, first extract the surface using
+        :func:`pyvista.DatasetFilters.extract_surface`.
```
MatthewFlamm, line 1803:
```suggestion
        :func:`pyvista.DataSetFilters.extract_surface`.
```

pyvista/core/dataset.py
```diff
@@ -1743,23 +1743,78 @@ def center(self) -> Vector:
...
+
+        Returns
+        -------
+        float
+            Total area of the mesh.
+
+        Examples
+        --------
+        Get the area of a square o size 2x2.
+        Note 5 points in each direction.
```
akaszynski, line 1794:
```suggestion
        Get the area of a square of size 2x2.
        Note 5 points in each direction.
```

tests/test_dataset.py
```diff
@@ -1478,3 +1478,40 @@ def test_separate_cells(hexbeam):
...
+    assert_volume(vol_grid.cast_to_structured_grid())
+    assert_area(surf_grid.cast_to_structured_grid())
+
+    # Rectilinear
+    assert_volume(vol_grid.cast_to_rectilinear_grid())
+    assert_area(surf_grid.cast_to_rectilinear_grid())
+
+    # PolyData
+    # cube of size 4
+    # PolyData ia special baucause it is a 2D surface that can enclose a volume
```
akaszynski, line 1514:
```suggestion
    # PolyData is special because it is a 2D surface that can enclose a volume
```

pyvista/core/dataset.py
```diff
@@ -1743,23 +1743,78 @@ def center(self) -> Vector:
...
         Returns
         -------
         float
             Total volume of the mesh.
 
         Examples
         --------
-        Get the volume of a sphere.
+        Get the volume of a cube of size 4x4x4.
+        Note 5 points in each direction.
```
akaszynski, line 1756:
```suggestion
        Note that there are 5 points in each direction.
```

akaszynski (approved):
Minor edits but otherwise. LGTM.

Thanks!

akaszynski:
Note for any future readers:

Rather than approximating volume by extracting the surface of an `UnstructuredGrid` and computing the volume of it, this PR changes the calculation to instead compute the sum of all the individual cells. This is both faster and more accurate (as much as ~3%).

PR #3162, opened 2022-08-13
disable caching of build directory

akaszynski:
Disable the caching of the build directory as this is causing builds to fail. The cache isn't working correctly anyway, so removing this cache will actually speed up the documentation build.


akaszynski:
Checks pass and I'm merging this as this is making all of our PRs red. Conda test is taking forever and will be fixed by #3163, ignoring.

tkoyama010:
Thanks @akaszynski to fix this quickly!

PR #3163, opened 2022-08-13
Use Mamba to speed up the conda build

akaszynski:
Conda builds take forever to create the environment. This PR proposes we speed that up by using the mamba solver.

After speaking with a few of the guys from anaconda, I think that they're highly considering bringing it from "experimental" to just making it an option. Considering that Anaconda now includes `Libmamba` in their main channel (see (this blog)[https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community]) I think it's safe to say that it's safe that we rely on this for testing.

A few thoughts when going over this PR:
- Is any of this even necessary? We ignore test failures and while I think it's really important that we support anaconda, I don't really think it's critical that we have to test it unless we can get the solution on par with pip's solution speed
- Can we drop testing package dostrings? Part of the reason why we're testing here is to support VTK 8.2, which is only available on conda, and our docstrings are always going to fail here, hence why we ignore test failures.

In short, I think we should either remove or at the very least prune this workflow.

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -101,7 +101,7 @@ jobs:
           conda install -n base conda-libmamba-solver
 
       - name: 'Workaround for mamba-org/mamba#488'
-        run: rm /usr/local/miniconda/pkgs/cache/*.json
+        run: rm /usr/local/miniconda/pkgs/cache/*
 
       - name: Create Anaconda environment
         run: |
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -122,7 +122,6 @@ jobs:
           python -c "import pyvista; print(pyvista.EXAMPLES_PATH)"
 
       - name: Test Core API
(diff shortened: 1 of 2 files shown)


akaszynski:
This PR takes the conda env build time from an abysmal three hours:
tmp

To around 3 minutes:
tmp

akaszynski:
Unit tests pass, but doctests fail.
```
2022-08-13T22:25:54.9084985Z ====== 234 failed, 823 passed, 3 skipped, 1 warning in 213.22s (0:03:33) =======
```

It's for the same reason why we're only testing them on 9.1.0 in Linux, many of our methods will not work with VTK <9.0 and it looks like it's on the order of 20%. I'm actually quite impressed that we have a thousand examples!

I'm removing docstring testing, but keeping the unit testing and having errors flagged rather than silently passed in the workflow by removing `continue-on-error: true`. In my opinion, if we're testing by allowing errors to silently pass, we might as well not test.

akaszynski:
Merging, because life is too short to wait three hours for conda.

PR #3167, opened 2022-08-14
Define pytest markers for skipping tests based on vtk version

adeak:
There's a common occurence in our tests that we want to skip some based on VTK version, most commonly things that require VTK 9. This is also the main source of reliance of tests on `pyvista._vtk`, which we might try to avoid as much as possible (at least according to a comment somewhere by @akaszynski, suggesting that tests should act like third parties to the library, so they shouldn't rely on `pyvista._vtk`).

This PR adds two custom pytest markers:
```
$ pytest --markers |head -3
@pytest.mark.needs_vtk9: skip test unless VTK is 9.0 or newer.

@pytest.mark.needs_vtk_version(version): skip test unless VTK version is at least as specified.
```

And test skips are rewritten to use these. There are a few notable exceptions: ones that have nontrivial skip messages, i.e.
https://github.com/pyvista/pyvista/blob/afe4dc32e5ad9a256c0f395607d03b739469acc6/tests/test_init.py#L16

https://github.com/pyvista/pyvista/blob/afe4dc32e5ad9a256c0f395607d03b739469acc6/tests/test_pointset.py#L9-L11

We could make the markers accept an optional reason, but I figured that would increase their complexity beyond usefulness.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -11,3 +11,6 @@ filterwarnings =
     ignore:.*The NumPy module was reloaded*:UserWarning
 doctest_optionflags = NUMBER ELLIPSIS
 testpaths = tests
+markers =
+    needs_vtk9: skip test unless VTK is 9.0 or newer.
+    needs_vtk_version(version): skip test unless VTK version is at least as specified.
--- a/tests/test_render_window_interactor.py
+++ b/tests/test_render_window_interactor.py
@@ -10,16 +10,12 @@ skip_no_plotting = pytest.mark.skipif(
     not system_supports_plotting(), reason="Requires system to support plotting"
 )
 
-skip_needs_vtk_9 = pytest.mark.skipif(
-    pyvista.vtk_version_info < (9, 1, 0), reason="Requires VTK>=9.1.0"
-)
-
 
 def empty_callback():
     return
 
 
-@skip_needs_vtk_9
+@pytest.mark.needs_vtk_version(9, 1)
 def test_observers():
     pl = pyvista.Plotter()
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,6 @@
 import numpy as np
 from numpy.random import default_rng
-from pytest import fixture
+from pytest import fixture, skip
 
 import pyvista
 from pyvista import examples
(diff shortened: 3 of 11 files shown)


tests/test_render_window_interactor.py
```diff
@@ -10,16 +10,12 @@
     not system_supports_plotting(), reason="Requires system to support plotting"
 )
 
-skip_needs_vtk_9 = pytest.mark.skipif(
-    pyvista.vtk_version_info < (9, 1, 0), reason="Requires VTK>=9.1.0"
-)
-
```
adeak, line 16: This was a weird case where the name is "needs vtk 9" but the check is "needs vtk 9.1". I decided the drop the name and keep the behaviour.

pytest.ini
```diff
@@ -11,3 +11,6 @@ filterwarnings =
     ignore:.*The NumPy module was reloaded*:UserWarning
 doctest_optionflags = NUMBER ELLIPSIS
 testpaths = tests
+markers =
+    needs_vtk9: skip test unless VTK is 9.0 or newer.
+    needs_vtk_version(version): skip test unless VTK version is at least as specified.
```
tkoyama010, line 16: Great idea! Only one point I am wondering is we need to split `nees_vtk9`. In my mind we can express it with `needs_vtk_version(9.0.0)`

tkoyama010 (changes requested):

pytest.ini
adeak: That would certainly work. The only reason why I defined a specific marker is that we have a lot of those cases in the test suite, and this way we don't have to define a `skip_if_vtk9` at the top of a bunch of test modules. After all the point of this PR is to reduce code duplication.

Major version bumps will always introduce a lot of new things. I expect we'll need `needs_vtk10` eventually. So we also can't default `needs_vtk_version` to 9.

tkoyama010 (approved):
LGTM

tests/test_render_window_interactor.py
akaszynski: Agreed. Thanks!

akaszynski:
Agree with the changes. Yes, ideally we should be avoiding use of `_vtk` within our tests as we're effectively treating it (as much as possible) as third party. There will be plenty of cases where we have to access private methods or attributes for unit testing, but if we can avoid those let's do it. It will force us to exercise our public API and improve it as needed.

MatthewFlamm (approved):
I like this much better.  I expect that these distinctions will continue to be important as more minor versions are published in vtk.

PR #3168, opened 2022-08-15
Speed up documentation build

akaszynski:
Our documentation build takes around 40 minutes, and with some local testing I've determined that the main issue is in building individual pages using our customized ``auto-summary`` directive. Personally, I find it quite helpful to have individual methods documented on individual pages, but I dislike how long the build takes.

This PR proposes that we put all the method and attributes for each class on the same page to speed up the build for all but the following conditions:
- `main` branch
- `release` branches
- `*v` tags

This way we get the full documentation build experience on all PRs without having to wait forever. The only thing we'll miss out on is being able to pull down documentation with class method and attributes on individual pages. I think the trade-off is worth it, especially considering that the time savings (at least locally) is on the order of 4x.

#### Documentation build with class documentation on one page.

```
$ time make -C doc clean-except-examples html

...

real    5m45.434s
user    5m4.675s
sys     0m21.619s
```

#### Documentation build with class methods and attributes on individual pages.

```
$ export FULL_DOC_BUILD=TRUE
$ time make -C doc clean-except-examples html

...

real    18m34.060s
user    14m21.593s
sys     0m33.595s
```

### Additional changes

- Reduce image size to `(400, 300)`. This reduces the documentation artifact (locally) from 470 MB to 370 MB and shaves another 30 seconds of the build (locally).
- Only upload `_build/html`. Previously we were including extra unnecessary files when we only need the `_build/html` directory rather than the entire `_build` directory.
- Fix pip caching using [GitHub Actions: setup-python now supports dependency caching](https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/). This should improve build times slightly and reduce the complexity of our CI. Implement this for all builds.
- Conditionally install the OSMesa VTK wheel.


adeak:
I also like having each method a single pahe, but a dump works too as  long as the anchors work reliably even on mobile. I'll try to check.

MatthewFlamm:
This will make developing documentation MUCH easier.  The main change that I see is separate from one page vs. separate pages.  This PR doesn't not duplicate inherited methods.  For example `UnstructuredGrid` documentation no longer contains methods that are inherited from `DataSet`.  This will also significantly reduce re-building the documentation.  Before, if I changed something in `DataSet`, it would rebuild _everything_ inheriting from `DataSet`.

It may be possible to have a middle ground where there are separate pages, but not duplicating inherited methods, if this is preferred.  I'm not sure how it could be accomplished however.

akaszynski:
I'd like to merge this ASAP. Mac OS build time are comparable, Windows reduced by 3 minutes, Linux down by a minute, and doc build is down by 30 minutes to just 15, which is just about the Mac OS time. This means 15 minutes for the total build time.

5 minutes local build (at least for me).

.github/workflows/style_docstr.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: 3.9
+          cache: 'pip'
 
       - name: Install pre-commit
```
MatthewFlamm, line 22: Do we need to use `cache-dependency-path`?  My understanding is that just `pip` will only use `requirements.txt`, but I'm not sure.  This comment applies to all the changes like this.

CONTRIBUTING.rst
```diff
@@ -395,8 +381,76 @@ exists, be sure to add the resulting image with
...
+individual class method and attribute will have its own documentation page, the
+documentation will take much longer to build.
+
+Parallel Documentation Build
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+You can improve your documentation build time on Linux and Mac OS with:
+
+.. code:: bash
+
+   make -C doc html
```
MatthewFlamm, line 446: phtml?

CONTRIBUTING.rst
```diff
@@ -395,8 +381,76 @@ exists, be sure to add the resulting image with
...
+default ``(1024, 768)``. This is to minimize the time it takes to build
+documentation locally. If you wish to generate the full documentation, enable
+it with::
+
+.. code:: bash
+
+   export FULL_DOC_BUILD=TRUE
+   make -C doc html
+
+This will generate the same documentation as rendered online, but as each
```
MatthewFlamm, line 436: I think the one page vs. multi page is important distinction, but it may also be important to point out the difference about inherited methods/attributes here too.  Otherwise it may be surprising to developers when comparing the documentation.

MatthewFlamm (commented):
I have tested this, and it seems good when I build locally.  It would be good to have more eyes as I'm not confident in Sphinx configuration.

adeak:
7.5 minutes clean-except-examples build on my debian with the new version (33 minutes clean build complete with gallery). 19 minutes for the original full build. Things seem to work in both cases.

It took me a while to notice that `CLASS_TABLE_AUTOSUMMARY` was renamed to `FULL_DOC_BUILD`; I recommend updating the PR comment for posterity.

As expected the single-page dumps are worse functionally, but if I understand correctly this only affects local builds, so both docs.pyvista.org and dev.pyvista org will have the original, proper build.  In which case this is indeed perfectly fine.

akaszynski:
> It took me a while to notice that `CLASS_TABLE_AUTOSUMMARY` was renamed to `FULL_DOC_BUILD`; I recommend updating the PR comment for posterity.

Done.

CONTRIBUTING.rst
akaszynski: Addressed in c33039905714aa1917d88431c36e1de855c57920.

CONTRIBUTING.rst
akaszynski: I figured that with the mention of `phtml` below that it's up to the user if they want to spin up all their cores, but I won't advocate it as the default since some people might want to run it in the background and now the vast majority of the time will be in the single threaded examples.

.github/workflows/style_docstr.yml
akaszynski: Good point.  I've added most in c33039905714aa1917d88431c36e1de855c57920, but the documentation style doesn't really need it (nor do we have a requirements file for pre-commit).

akaszynski:
> As expected the single-page dumps are worse functionally, but if I understand correctly this only affects local builds, so both docs.pyvista.org and dev.pyvista org will have the original, proper build. In which case this is indeed perfectly fine.

That's correct. We'll verify after the merge, but both the dev and the production will use the "full build".

CONTRIBUTING.rst
MatthewFlamm: I guess I don't follow this section then.  The code here is exactly the same as the first code block under `Building the documentation`.  Is there a flag missing from this example?  I don't see `phtml` mentioned in this `Building the documentation` section.

akaszynski: Sorry, should have actually referenced the file before commenting. Yes, it should be phtml.

```suggestion
   make -C doc phtml
```

MatthewFlamm (approved):
LGTM based mostly on building the docs successfully locally.

adeak:
CI doc build artifacts have also gone from ~700 MB to 280 MB compressed, 1.2 GB to 370 MB decompressed.

Issue #500, opened 2019-12-13
Crash after closing figure window


alexdaniilidis:
Trying to visualize a 3D dataset with an orthogonal grid. Everything works as expected. Plot appears and is interactive. Closing the plot either from the window or with q gives the error below and kernel becomes unresponsive. Only solution is to restart kernel. After restaring the problem persists

Running on Spyder 4.0.0 with Python 3.7, win10
Sys version: 3.7.4 (default, Aug  9 2019, 18:34:13) [MSC v.1915 64 bit (AMD64)]
Spyder uses the QT5 backend (default) for graphics.

-----

**To Reproduce**

```py
import numpy as np
import discretize
import pyvista


ncx = 15      # number of core mesh cells in x
ncy = 15      # number of core mesh cells in y
ncz = 15      # number of core mesh cells in y
dx = 15       # base cell width x
dy = 15       # base cell width y
dz = 1        # base cell width z
npad_x = 0    # number of padding cells in x
npad_y = 0    # number of padding cells in y
npad_z = 0    # number of padding cells in z
exp_x = 1.25  # expansion rate of padding cells in x
exp_y = 1.25  # expansion rate of padding cells in y
exp_z = 1.25  # expansion rate of padding cells in z

hx = [(dx, npad_x, -exp_x), (dx, ncx), (dx, npad_x, exp_x)]
hy = [(dy, npad_y, -exp_y), (dy, ncy), (dy, npad_y, exp_y)]
hz = [(dz, npad_z, -exp_z), (dz, ncz), (dz, npad_z, exp_z)]

mesh = discretize.TensorMesh([hx, hy, hz])
vals = np.random.randint(1, 1000, mesh.nC)

dataset = mesh.toVTK({'myvals':vals})

p = pyvista.Plotter(notebook=False)
p.show_grid()

xyz = (200,112.5,7.5)
p.add_mesh_slice_orthogonal(dataset)
p.show()
```

**Screenshots**
![PyVista_error](https://user-images.githubusercontent.com/34479636/70791700-de981f00-1d97-11ea-9b3b-07d2770293e2.PNG)

**System Information:**
--------------------------------------------------------------------------------
  Date: Fri Dec 13 10:56:25 2019 W. Europe Standard Time

           Windows : OS
                 8 : CPU(s)
             AMD64 : Machine
             64bit : Architecture
           31.9 GB : RAM
           Jupyter : Environment

[...]

Issue #2463, opened 2022-04-13
Performance Profile Log


adam-grant-hendry:
In attempting how best to tackle testing for [#500](https://github.com/pyvista/pyvista-support/issues/500), I think the best solution would be to add and maintain a performance profile log with the repo, similar to a coverage report.

Tests could be written with `timeit`, `cProfile`, `memory_profile`, etc., and perhaps we could write something like a `pytest` fixture that would generate an `instaviz` report, or similar? I'm not sure how to implement yet (perhaps I would need to create a separate PyPI package or `pytest` plug-in the project would use?). Please let me know your thoughts.

I think a log makes the most sense because:

1. The alternative (in my mind) would be maintaining old copies of code to test against to demonstrate PRs introduce performance improvements, which seems silly and unmaintainable, and

2. It may be desirable in the future to add a PR that reduces performance to some degree for a big benefit in readability or usability, so rather than force future PRs to meet or exceed a certain level of performance, we could reference the log and use it as a guide (similar to how coverage reports are used as a guide, but aren't gospel).

What are your thoughts?

@akaszynski @MatthewFlamm @banesullivan I'm unaware of any such python packages or plug-ins that would implement this. If you have any suggestions, that would be greatly appreciated! I think this could be a great addition!

akaszynski:
I think you're correct @adam-grant-hendry, and we could use [pytest-benchmark](https://pypi.org/project/pytest-benchmark/) for this. Would be nice to have the tests in this repository, though I'm loath to adding any more workflows to our CI/CD.

adam-grant-hendry:
@akaszynski I completely understand. And thanks for introducing me to `pytest-benchmark`! If I can teach myself how to add workflows to our CI/CD, would you be willing to let me add it?

akaszynski:
> If I can teach myself how to add workflows to our CI/CD, would you be willing to let me add it?

We should consider creating a new repository that includes our benchmarks outside this repository. We have enough workflows as it is and it's becoming quite something.

@pyvista/developers, your thoughts on this? Create a new repo with benchmarks (`pyvista-benchmarks`) or add the workflow here in this repo?

adam-grant-hendry:
I'm good with whatever decision is made.

adeak:
> @pyvista/developers, your thoughts on this? Create a new repo with benchmarks (`pyvista-benchmarks`) or add the workflow here in this repo?

[...]

PR #3170, opened 2022-08-15
fix(add_volume): use memory-efficient rescaling

adam-grant-hendry:
### Overview

Inefficient `numpy` copies causes memory leak for large volume
data sets. Use a vectorized rescaling instead.

Fixes `pyvista/support` Issue [#500](https://github.com/pyvista/pyvista-support/issues/500)

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -2934,7 +2934,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
 
         if scalars is None:
             # Make sure scalars components are not vectors/tuples
-            scalars = volume.active_scalars
+            scalars = volume.active_scalars.copy()
             # Don't allow plotting of string arrays by default
             if scalars is not None and np.issubdtype(scalars.dtype, np.number):
                 scalar_bar_args.setdefault('title', volume.active_scalars_info[1])
@@ -2970,8 +2970,10 @@ class BasePlotter(PickingHelper, WidgetHelper):
 
         # Scalars interpolation approach
         if scalars.shape[0] == volume.n_points:
+            volume.point_data.set_array(scalars, title, True)
             self.mapper.SetScalarModeToUsePointData()
         elif scalars.shape[0] == volume.n_cells:
+            volume.cell_data.set_array(scalars, title, True)
             self.mapper.SetScalarModeToUseCellData()
         else:
             raise_not_matching(scalars, volume)
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -2845,6 +2845,11 @@ class BasePlotter(PickingHelper, WidgetHelper):
                     raise TypeError(
                         f'Object type ({type(volume)}) not supported for plotting in PyVista.'
                     )
+        else:
+            # HACK: Make a copy so the original object is not altered.
+            #       Also, place all data on the nodes as issues arise when
+            #       volume rendering on the cells.
+            volume = volume.cell_data_to_point_data()
 
         if name is None:
             name = f'{type(volume).__name__}({volume.memory_address})'
@@ -2925,15 +2930,6 @@ class BasePlotter(PickingHelper, WidgetHelper):
 
         if not isinstance(scalars, np.ndarray):
             scalars = np.asarray(scalars)
-
-        for name in volume.array_names:
-            if scalars is volume[name]:
-                # ``add_volume`` rescales ``scalars`` to 8-bit for color mapping so that
-                # it can be plotted. We should not mutate the original data. We add a
-                # guard here in case any of the previous methods for obtaining ``scalars``
-                # from ``volume`` obtained the original scalars.
-                scalars = volume[name].copy()
-                break
 
         if not np.issubdtype(scalars.dtype, np.number):
             raise TypeError('Non-numeric scalars are currently not supported for volume rendering.')
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -2965,15 +2965,12 @@ class BasePlotter(PickingHelper, WidgetHelper):
         elif isinstance(clim, float) or isinstance(clim, int):
             clim = [-clim, clim]
 
+        # convert the scalars to np.uint8 and scale between 0 and 255 within clim
         clim = np.asarray(clim, dtype=scalars.dtype)
-
         scalars.clip(clim[0], clim[1], out=scalars)
-
         min_ = np.nanmin(scalars)
         max_ = np.nanmax(scalars)
-
         np.true_divide((scalars - min_), (max_ - min_) / 255, out=scalars, casting='unsafe')
-
         volume[title] = np.array(scalars, dtype=np.uint8)
 
         self.mapper.scalar_range = clim
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -2634,8 +2634,6 @@ class BasePlotter(PickingHelper, WidgetHelper):
         specular=0.2,
         specular_power=10.0,
         render=True,
-        rescale_scalars=True,
-        copy_cell_to_point_data=True,
         **kwargs,
     ):
         """Add a volume, rendered using a smart mapper by default.
@@ -2776,22 +2774,6 @@ class BasePlotter(PickingHelper, WidgetHelper):
         render : bool, optional
             Force a render when True.  Default ``True``.
 
-        rescale_scalars : bool, optional
-            Rescale scalar data. This is an expensive memory and time
-            operation, especially for large data. In that case, it is
-            best to set this to ``False``, clip and scale scalar data
-            of ``volume`` beforehand, and pass that to ``add_volume``.
-            Default ``True``.
-
-        copy_cell_to_point_data : bool, optional
-            Make a copy of the original ``volume``, passing cell data
-            to point data. This is an expensive memory and time
-            operation, especially for large data. In that case, it is
-            best to choose ``False``. However, this copy is a current
-            workaround to ensure original object data is not altered
-            and volume rendering on cells exhibits some issues. Use
-            with caution. Default ``True``.
-
         **kwargs : dict, optional
             Optional keyword arguments.
 
@@ -2863,12 +2845,6 @@ class BasePlotter(PickingHelper, WidgetHelper):
                     raise TypeError(
                         f'Object type ({type(volume)}) not supported for plotting in PyVista.'
                     )
-        else:
-            if copy_cell_to_point_data:
-                # HACK: Make a copy so the original object is not altered.
-                #       Also, place all data on the nodes as issues arise when
-                #       volume rendering on the cells.
-                volume = volume.cell_data_to_point_data()
 
         if name is None:
             name = f'{type(volume).__name__}({volume.memory_address})'
@@ -2950,6 +2926,15 @@ class BasePlotter(PickingHelper, WidgetHelper):
         if not isinstance(scalars, np.ndarray):
             scalars = np.asarray(scalars)
 
+        for name in volume.array_names:
+            if scalars is volume[name]:
+                # ``add_volume`` rescales ``scalars`` to 8-bit for color mapping so that
+                # it can be plotted. We should not mutate the original data. We add a
+                # guard here in case any of the previous methods for obtaining ``scalars``
+                # from ``volume`` obtained the original scalars.
+                scalars = volume[name].copy()
+                break
+
         if not np.issubdtype(scalars.dtype, np.number):
             raise TypeError('Non-numeric scalars are currently not supported for volume rendering.')
         if scalars.ndim != 1:
@@ -2984,19 +2969,18 @@ class BasePlotter(PickingHelper, WidgetHelper):
         elif isinstance(clim, float) or isinstance(clim, int):
             clim = [-clim, clim]
 
-        if rescale_scalars:
-            clim = np.asarray(clim, dtype=scalars.dtype)
+        clim = np.asarray(clim, dtype=scalars.dtype)
 
-            scalars.clip(clim[0], clim[1], out=scalars)
+        scalars.clip(clim[0], clim[1], out=scalars)
 
-            min_ = np.nanmin(scalars)
-            max_ = np.nanmax(scalars)
+        min_ = np.nanmin(scalars)
+        max_ = np.nanmax(scalars)
 
-            np.true_divide((scalars - min_), (max_ - min_) / 255, out=scalars, casting='unsafe')
+        np.true_divide((scalars - min_), (max_ - min_) / 255, out=scalars, casting='unsafe')
 
-            volume[title] = np.array(scalars, dtype=np.uint8)
+        volume[title] = np.array(scalars, dtype=np.uint8)
 
-            self.mapper.scalar_range = clim
+        self.mapper.scalar_range = clim
 
         # Set colormap and build lookup table
         table = _vtk.vtkLookupTable()
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -2611,7 +2611,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         scalars=None,
         clim=None,
         resolution=None,
-        opacity="linear",
+        opacity='linear',
         n_colors=256,
         cmap=None,
         flip_scalars=False,
@@ -2621,13 +2621,13 @@ class BasePlotter(PickingHelper, WidgetHelper):
         categories=False,
         culling=False,
         multi_colors=False,
-        blending="composite",
+        blending='composite',
         mapper=None,
         scalar_bar_args=None,
         show_scalar_bar=None,
         annotations=None,
         pickable=True,
-        preference='point',
+        preference="point",
         opacity_unit_distance=None,
         shade=False,
         diffuse=0.7,
@@ -2819,9 +2819,9 @@ class BasePlotter(PickingHelper, WidgetHelper):
         cmap = kwargs.pop('colormap', cmap)
         culling = kwargs.pop('backface_culling', culling)
 
-        if 'scalar' in kwargs:
+        if "scalar" in kwargs:
             raise TypeError(
-                '`scalar` is an invalid keyword argument for `add_mesh`. Perhaps you mean `scalars` with an s?'
+                "`scalar` is an invalid keyword argument for `add_mesh`. Perhaps you mean `scalars` with an s?"
             )
         assert_empty_kwargs(**kwargs)
 
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -2611,7 +2611,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         scalars=None,
         clim=None,
         resolution=None,
-        opacity='linear',
+        opacity="linear",
         n_colors=256,
         cmap=None,
         flip_scalars=False,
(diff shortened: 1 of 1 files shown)


pyvista/plotting/plotting.py
```diff
@@ -2611,7 +2611,7 @@ def add_volume(
         scalars=None,
         clim=None,
         resolution=None,
-        opacity='linear',
+        opacity="linear",
```
adeak, line 2614: Just a drive-by comment for now: I see that a considerable part of the diff comes from string normalization like this. Since we explicitly skip these in `black`, see `skip-string-normalization` in https://github.com/pyvista/pyvista/blob/51fe884c50421e0889add8671a00abcc302f1a6e/pyproject.toml#L10-L14
I suspect this is due to some rogue tooling gone wild again. I would recommend not doing these tangential changes, because they make the PR harder to review, and they bloat the `git blame`. I can think of exceptions, e.g. if you notice a four-line exception message containing a mixture of single and double quotes.

adam-grant-hendry: Thank you, yes. I will switch the quoting back.

adam-grant-hendry:
@akaszynski @adeak My change is causing a noticeable difference, but I can't narrow down why. Below is a sample I ran and the corresponding output:

```python
from pyvista import examples

if __name__ == '__main__':
    uniform = examples.load_uniform()

    uniform.plot(
        volume=True,
        parallel_projection=True,
        show_scalar_bar=False,
        show_grid=True,
    )
```

img_diff

adeak:
> My change is causing a noticeable difference, but I can't narrow down why.

I haven't gotten to the bottom of it yet, but I can only restore the original result if I add back `volume.cell_data.set_array(scalars, title, True)` etc. on lines 2961-2964, _and_ restore old line 2983 `scalars = ((scalars - np.nanmin(scalars)) / (np.nanmax(scalars) - np.nanmin(scalars))) * 255` instead of the call to `np.true_divide()`. I don't see why the latter is wrong. Rewriting it as
```py
scalars -= min_
scalars /= max_ - min_
scalars *= 255
```
doesn't restore the original behaviour. Everything seem to be floats, yet it looks as if some rounding happens in one of the cases. Like I said, not sure what's going on yet.

adam-grant-hendry:
@akaszynski @adeak I found the problem.

`volume['Spatial Point Data']` is getting mutated. The reason is because of [this line](https://github.com/pyvista/pyvista/blob/51fe884c50421e0889add8671a00abcc302f1a6e/pyvista/plotting/plotting.py#L2918):

```python
2918:            scalars = volume.active_scalars
```

This creates a reference to `uniform.active_scalars`, rather than a copy. Hence, we are rescaling it as well as the newly added `Data` points.

@adeak This is also why using inplace methods (`+=`, `-=`, `/=`, `*=`) does not work because the underlying `__imethod__()` methods are designed to modify mutable values whenever possible (unlike immutable objects). It's the same reason why this happens:

```python
>>> a = b = [1, 2, 3]
>>> a += [4, 5]
>>> print(b)
[1, 2, 3, 4, 5]
>>> a = b = [1, 2, 3]
>>> a = a + [4, 5]
>>> print(b)
[1, 2, 3]
```

So the previous incantation for `scalars` was making a copy.

Ned Batchelder gave an excellent explanation of this in his Pycon2015 talk [Facts and Myths about Python names and values](https://www.youtube.com/watch?v=_AEJHKGk9ns).

The solution is to use

```python
2918:            scalars = volume.active_scalars.copy()
```

and then also to add back the removed lines

```python
volume.point_data.set_array(scalars, title, True)
```

and

```python
volume.cell_data.set_array(scalars, title, True)
```

Even though we make a copy of `volume.active_scalars` in both cases, we still win here memory-wise since we are no longer casting `scalars` to a `float64`, increasing its `dtype` from 2 bytes to 8, but instead perform a vectorized float divide with `np.true_divide`.

adam-grant-hendry:
@akaszynski @adeak @tkoyama010 Would you please kindly review this PR (and/or assign an appropriate contributor/developer/maintainer to review)? Thank you!

akaszynski:
Will do tonight or tomorrow morning.

adam-grant-hendry:
@akaszynski @adeak I'm going to put the `HACK` back in and remove the "memory guard" in my last commit.

Iterating over `volume.array_names` causes another mutation: If I run the code from `test_plot_helper_two_volumes`:

```python
import pyvista
from pyvista import examples

if __name__ == '__main__':
    uniform = examples.load_uniform()

    grid = uniform.copy()
    grid.origin = (0, 0, 10)

    pyvista.plot(
        [uniform, grid],
        volume=True,
        show_scalar_bar=False,
    )
```

adding this in `add_volume`:

```python
2929:        for _ in volume.array_names:
2930             pass
```

causes no problems, but doing this:

```python
2929:        for name in volume.array_names:
2930             pass
```

creates an issue:

Scope_Leak

the latter is equivalent to

```python
name = volume.array_names[0]
# Do stuff
name = volume.array_names[1]
# Do stuff
```

and `for` loops leak their scope, so `name` maintains a reference to `volume.array_names[-1]` after the `for` loop ends, which causes only one volume to get plotted for some reason.

One way around this that works is to use a list comprehension, since those do not leak scope (at least in Python 3):

```python
names = [name for name in volume.array_names]
# Now iterate over `names`
```

but I'd rather not play with fire.

I think the occurrence of these mutations is a bit of a concern: they're not a Windows/Linux/MacOS issue; they're a Python usage issue. These kinds of errors may pop up (be popping up?) in other unexpected places throughout the code base, so we might need to investigate this as part of a separate issue with a deeper dive into the implementation.

For now, I'll revert the last changes to what I know works (passing tests, no mutations) so the original issue of the PR can be solved.

akaszynski:
> I think the occurrence of these mutations is a bit of a concern

I find this troubling as well. Let's mention this in the revert and we need to look into this later. I'm not surprised about the scope leaking, I am about the side-effects.

adam-grant-hendry:
> We're failing now due to https://github.com/pyvista/pyvista/commit/f791d0de74e6f1567ddbf2e9ae07771fe93cbe05. I'll wait for this to be fixed.

Yes, this opened another mutation issue. I've reverted the offending changes, but please do see my last comment because it might be something to look into.

> Also, if possible, can you provide a basic performance benchmark for this? I'm curious and will do this on my own if you don't do this regardless.

Yes, I was thinking about that. I'm still not sure how to do it. I know you created [pyvista benchmarks](https://github.com/pyvista/pyvista-benchmarks) for us in response to this issue, but I haven't looked at it in a while.

Do you have any recommendations? I'd appreciate your help with this.

adeak:
I don't have free bandwidth for this this week, alas. But +1 for

> Also, if possible, can you provide a basic performance benchmark for this? I'm curious and will do this on my own if you don't do this regardless.

I'd be curious about the change in memory need, e.g. because you mention "memory leak" but I doubt that's actually the case (that would be pretty bad: unaccounted memory staying allocated after the function's returned). If it's about large intermediary arrays being used, then it would be useful to see how much things are improved (because, as we discussed, working on views alone leads to a lot of subtle bugs).

akaszynski:
Super simple (assuming linux)

### `main` branch
```
time python -c "from pyvista import examples; dataset = examples.download_frog(); dataset.plot(volume=True, off_screen=True)"
```

```
real	0m4.333s
user	0m3.806s
sys	0m0.573s
```

adeak:
Ah, I thought you meant memory profiling @akaszynski. Since this is partly motivated with that in mind.

adam-grant-hendry:
@akaszynski and @adeak Both speed and memory might be good to double-check here. I know [airspeed velocity](https://github.com/pyvista/pyvista/issues/2463#issuecomment-1184572442) was recommended to us (which look's like `cprofile` + `snakeviz` like you said @adeak )

Some questions:

1. Do you want me to add the benchmark test to the `pyvista` test suite, or in [pyvista-benchmarks](https://github.com/pyvista/pyvista-benchmarks)?
2. Do I need to add dependencies to the `requirements.txt`? Ones for consideration are:
    - [airspeed velocity](https://github.com/pyvista/pyvista/issues/2463#issuecomment-1184572442)
    - [pytest-monitor](https://pytest-monitor.readthedocs.io/en/latest/)
    - [pytest-leaks](https://github.com/abalkin/pytest-leaks)
    - [pytest-benchmark](https://pypi.org/project/pytest-benchmark/) -> already part of `pyvista-benchmarks`
Any others?
3. Is `pyvista-benchmarks` ready to go w.r.t. the following items:
    - includes a suite of a benchmark scripts in the repo, which can be run both in this PR's tip and in the base (hopefully pyvista:main) [@adeak 's suggestion](https://github.com/pyvista/pyvista/issues/2463#issuecomment-1100478528)
    - which version of `VTK` should be used? (8 or 9?)
    - hardware isolation?
    - Can I use the [DICOM Stack data set](https://github.com/pyvista/vtk-data/pull/5) I made, or is there still a question of whether we use datasets with [a range of sizes](https://github.com/pyvista/pyvista/issues/2463#issuecomment-1100382675)?

MatthewFlamm:
> I also found https://github.com/pyvista/pyvista-support/issues/500#issuecomment-921108703 and @MatthewFlamm mentioned he https://github.com/pyvista/pyvista-support/issues/500#issuecomment-921106561. However, that profile looked at objects in the same function scope, so I'm not 100% certain that one is accurate.

Let's keep the discussion of this out of this PR since it is off topic.  I created #3178 to track this.

adeak:
I can't speak for what @akaszynski had in mind, but I don't personally think this is about the benchmark repo. At least in my mind that's more about general performance checking. In this case we have a very specific target: `add_volume()`. It's probably a lot easier and faster to come up some simple scripts that call `add_volume` with input of various configurations (different number and dtype and size of data, mostly), and check their runtime/memory need.

We did discuss that one use case for the benchmark repo could be targeted profiling of performance-oriented PRs, but my impression is that coming up with a decent benchmark suite we could add to the benchmark repo would need a lot more care.

adam-grant-hendry:
@adeak @akaszynski 

There might be a better way to do this, but I used [your frog example](https://github.com/pyvista/pyvista/pull/3170#issuecomment-1216977576) and wrote a simple benchmark script:

<details>
<summary><b>benchmark_test.py</b></summary>

```python
import cProfile
import functools
import io
import platform
import pstats
import sys

import GPUtil
from memory_profiler import profile as memory_profile
import psutil
from tabulate import tabulate
import vtkmodules

from pyvista import examples

OS = {
    'win32': 'Windows',
    'darwin': 'MacOS',
    'aix': 'AIX',
    'linux': 'Linux',
    'cygwin': 'Windows with Cygwin',
}


def SI2ibi(mem):
    """Convert from SI bytes to ibibytes.

    Computers use powers of 2, so memory is represented in ibibytes, but
    SI prefixes are powers of 1000)

    kibi (KiB) = (2^10)^1, kilo (KB) = (10^3)^1 (1.024 KiB = 1 KB)
    mebi (MiB) = (2^10)^2, mega (MB) = (10^3)^2 (1.048576 MiB = 1 MB)
    gibi (GiB) = (2^10)^3, giga (GB) = (10^3)^3 (1.073741824 GiB = 1 GB)

    Args:
        mem ([int]): Memory size in bytes
    """
    factor = 1024
    for unit in ["", "K", "M", "G", "T", "P"]:
        if mem < factor:
            break
        mem /= factor

    return mem, f'{unit}iB'


def print_info() -> str:
    header_width = 80

    python_major, python_minor, python_patch, _, _ = sys.version_info
    os_ = OS.get(sys.platform, sys.platform)
    uname = platform.uname()

    print(f"{' System Info ':=^{header_width}}")

    print(f'OS: {os_}')
    if os_ in {'Windows', 'Windows with Cygwin'}:
        windows_major, windows_minor, windows_build, _, _ = sys.getwindowsversion()
        print(f'Version: {windows_major}.{windows_minor}, Build {windows_build}')
    else:
        print(f'Release: {uname.release}')
        print(f'Version: {uname.version}')
    print(f'Machine: {uname.machine}')

    print(f"{' CPU Info ':=^{header_width}}")

    print(f'Processor: {uname.processor}')
    print(f'Physical Cores: {psutil.cpu_count(logical=False)}')
    print(f'Total Cores: {psutil.cpu_count(logical=True)}')

    print(f"{' Virtual Memory ':=^{header_width}}")

    svmem = psutil.virtual_memory()

    total, units = SI2ibi(svmem.total)
    print(f'Total: {total:.2f} {units}')

    avail, units = SI2ibi(svmem.available)
    print(f'Available: {avail:.2f} {units}')

    print(f"{' Swap Memory ':=^{header_width}}")

    swap = psutil.swap_memory()

    total, units = SI2ibi(swap.total)
    print(f'Total: {total:.2f} {units}')

    free, units = SI2ibi(swap.free)
    print(f'Available: {free:.2f} {units}')

    print(f"{' GPU Info ':=^{header_width}}")

    gpus = GPUtil.getGPUs()
    gpu_list = []

    for gpu in gpus:
        name = gpu.name
        total_memory = f'{gpu.memoryTotal} MB'
        free_memory = f'{gpu.memoryFree} MB'

        gpu_list.append([name, total_memory, free_memory])

    print(tabulate(gpu_list, headers=('Name', 'Total Memory', 'Free Memory')))

    print(f"{' Python ':=^{header_width}}")

    print(f'Python: {python_major}.{python_minor}.{python_patch}')

    print(f"{' VTK ':=^{header_width}}")

    print(f'VTK: {vtkmodules.__version__}')


def time_profile(filename):
    def profiler(fnc):
        """Wrapper for cProfile"""
        @functools.wraps(fnc)
        def inner(*args, **kwargs):
            pr = cProfile.Profile()
            pr.enable()

            retval = fnc(*args, **kwargs)

            pr.disable()

            s = io.StringIO()
            sortby = "cumulative"
            ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
            ps.print_stats()

            with open(filename, 'w', encoding='utf8') as w:
                w.write(s.getvalue())

            print(s.getvalue())
            return retval

        return inner
    return profiler


def time_plot_frog(filename, dataset):
    @time_profile(filename)
    def plot_frog(data):
        data.plot(volume=True, off_screen=True)

    plot_frog(dataset)


def memory_plot_frog(filename, dataset):
    with open(filename, 'w+', encoding='utf8') as w:
        @memory_profile(stream=w)
        def plot_frog(data):
            data.plot(volume=True, off_screen=True)

        plot_frog(dataset)

    with open(filename, 'r', encoding='utf8') as r:
        print(r.read())

if __name__ == '__main__':
    header_width = 80

    print_info()
    print()

    print('Downloading dataset...')
    dataset = examples.download_frog()
    print()

    print('Running time benchmark...')
    print()
    print(f"{' Time Profile ':=^{header_width}}")
    time_plot_frog('time_profile.log', dataset)
    print()

    print('Running memory benchmark...')
    print()
    print(f"{' Memory Profile ':=^{header_width}}")
    memory_plot_frog('memory_profile.log', dataset)
    print()

    print('Done!')

```
</details>

On my Windows machine, this is what I get:

<details>
<summary><b>System Info</b></summary>

```powershell
==================================System Info===================================
OS: Windows
Version: 10.0, Build 19042
Machine: AMD64
====================================CPU Info====================================
Processor: Intel64 Family 6 Model 85 Stepping 7, GenuineIntel
Physical Cores: 48
Total Cores: 96
=================================Virtual Memory=================================
Total: 190.65 GB
Available: 133.87 GB
==================================Swap Memory===================================
Total: 217.65 GB
Available: 153.52 GB
====================================GPU Info====================================
Name             Total Memory    Free Memory  
---------------  --------------  -------------
Quadro RTX 8000  46080.0 MB      45662.0 MB   
Quadro RTX 8000  46080.0 MB      43493.0 MB   
Quadro RTX 4000  8192.0 MB       6312.0 MB    
=====================================Python=====================================
Python: 3.8.10
======================================VTK=======================================
VTK: 9.1.0
```
</details>

## `main` branch

[main_time_profile.log](https://github.com/pyvista/pyvista/files/9354796/main_time_profile.log)
[main_memory_profile.log](https://github.com/pyvista/pyvista/files/9354797/main_memory_profile.log)

## `fix/add_volume` branch

[fix_add_volume_time_profile.log](https://github.com/pyvista/pyvista/files/9354818/fix_add_volume_time_profile.log)
[fix_add_volume_memory_profile.log](https://github.com/pyvista/pyvista/files/9354819/fix_add_volume_memory_profile.log)

## Summary

| Branch         | Time (s) | Memory (MiB) |
| -------------- | -------- | ------------ |
| main           | 7.796    | 294.9        |
| fix/add_volume | 7.013    | 78.3     |

adam-grant-hendry:
@adeak @akaszynski Can you let me know if the benchmark script I wrote (and results) are sufficient?

> It's probably a lot easier and faster to come up some simple scripts that call add_volume with input of various configurations (different number and dtype and size of data, mostly), and check their runtime/memory need.

Do I need to add my script as a test to the `pyvista` repo somehow?

adeak:
> Do I need to add my script as a test to the `pyvista` repo somehow?

Nope. Either we should just eyeball it here to get a rough idea about the performance implications, or make it a proper benchmark in the benchmark repo.

adam-grant-hendry:
> I'm not surprised about the scope leaking

Yes, Python leaks its scope in `for` loops by default.

adam-grant-hendry:
> > I think the occurrence of these mutations is a bit of a concern
> 
> I find this troubling as well. Let's mention this in the revert and we need to look into this later. I'm not surprised about the scope leaking, I am about the side-effects.

FYI, it should be noted that in both instances, the side effects (array mutations) happened on attribute access for the dataset, first with `volume.active_scalars`, and then with `volume.array_names`. This might give us a good place to start looking (e.g. in `__getattr__` methods), but we can also dive deeper with a debugger.

pyvista/plotting/plotting.py
```diff
@@ -2972,17 +2965,16 @@ def add_volume(
...
+        clim = np.asarray(clim, dtype=scalars.dtype)
+
+        scalars.clip(clim[0], clim[1], out=scalars)
+
+        min_ = np.nanmin(scalars)
+        max_ = np.nanmax(scalars)
+
+        np.true_divide((scalars - min_), (max_ - min_) / 255, out=scalars, casting='unsafe')
+
+        volume[title] = np.array(scalars, dtype=np.uint8)
```
akaszynski, line 2977:
```suggestion
        # convert the scalars to np.uint8 and scale between 0 and 255 within clim
        clim = np.asarray(clim, dtype=scalars.dtype)
        scalars.clip(clim[0], clim[1], out=scalars)
        min_ = np.nanmin(scalars)
        max_ = np.nanmax(scalars)
        np.true_divide((scalars - min_), (max_ - min_) / 255, out=scalars, casting='unsafe')
        volume[title] = np.array(scalars, dtype=np.uint8)
```

akaszynski (approved):
I made one minor change (comment and removed a few unnecessary lines), but otherwise this is good to go.

Thanks for adding this, and I appreciate the benchmarks. It was mostly for my curiosity as I have limited time to test all of this.

adeak:
@adam-grant-hendry with a newer numpy now we get a warning during doc build from this line https://github.com/pyvista/pyvista/pull/3170/files#diff-355952f3009b3d15d0d85ebb510556c3db59722b98c0d5c7d57abdb2b6ea5363R2974

https://github.com/pyvista/pyvista/blob/68d27ef9d3b519428a7cd521cde495b7bcc2fdf6/pyvista/plotting/plotting.py#L3596

```
/home/adeak/pyvista/pyvista/plotting/plotting.py:3596: RuntimeWarning: invalid value encountered in cast                                                                 
  volume[title] = np.array(scalars, dtype=np.uint8)
```

I think what's going on is that there's one or more `nan`s in there which can't be cast to uint8.

Do you remember by any chance
1. why you replaced `scalars.astype(np.uint8)` with `np.array(scalars, dtype=np.uint8)`? (We could add `casting='unsafe'` to the latter, but of course it will still replace nans and infs with `0`s which might merit a warning note in the docs.)
2. if you considered the case where the input scalars already has `dtype=np.uint8`, in which case in-place operations might be surprising?
3. if the method is supposed to work with uint8 input in the first place? The docstring is not specific, and it would be useful to pin down valid input dtypes.

banesullivan:
We may want to open a new issue for this

PR #3179, opened 2022-08-17
Lessen memory usage for wrapping vtk datasets

### LINKED ISSUE #3178: Reader memory  usage/memory usage when wrapping vtkDataSets
As pointed out in https://github.com/pyvista/pyvista-support/issues/500#issuecomment-921108703 and https://github.com/pyvista/pyvista-support/issues/500#issuecomment-921106561, readers seem to duplicate memory usage.

Using this code shows that the mesh data is copied from the reader to the mesh object.  This is only released when deleting the reader object.

```py
import pyvista as pv
from pyvista import examples

from memory_profiler import profile

@profile
def run():
    filename = examples.download_parched_canal_4k(load=False)
    mesh = reader.read()
    del reader

if __name__ == "__main__":
    run()
```

```txt
$python test.py 
Filename: test.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
     6    110.2 MiB    110.2 MiB           1   @profile
     7                                         def run():
     8    110.2 MiB      0.0 MiB           1       filename = examples.download_parched_canal_4k(load=False)
     9    110.4 MiB      0.2 MiB           1       reader = pv.get_reader(filename)
    10    302.8 MiB    192.3 MiB           1       mesh = reader.read()
    11    206.8 MiB    -96.0 MiB           1       del reader
```

MatthewFlamm:
### Overview

 

This PR adds a `deep` kwarg to all pyvista `DataSets` with default `False`.  This keyword is used when passing in a `vtkDataSet`. Some datasets already did this.  Some only did deep copy.

Closes #3178 

### Details

We could also consider using this kwarg when passing in arrays or other, but I haven't touched that logic here.

TODO:
- [x] add more tests

### Breaking Change

Previously, some pyvista `DataSet`s copied vtk `DataSet`s when wrapping them.  This PR changes the default to not copy for all `DataSet`s.  Any changes to the pyvista object will also change the original vtk object.  A kwarg `deep` was added to obtain the prior behavior, if needed.

### DIFF
--- a/pyvista/core/grid.py
+++ b/pyvista/core/grid.py
@@ -427,7 +427,7 @@ class UniformGrid(_vtk.vtkImageData, Grid, UniformGridFilters):
         Origin of the uniform grid.  Defaults to ``(0.0, 0.0, 0.0)``.
 
     deep : bool, optional
-        Whether to deep copy a vtkImage3Data object.
+        Whether to deep copy a vtkImageData object.
         Default is ``False``.  Keyword only.
 
     Examples
--- a/pyvista/core/grid.py
+++ b/pyvista/core/grid.py
@@ -95,6 +95,10 @@ class RectilinearGrid(_vtk.vtkRectilinearGrid, Grid, RectilinearGridFilters):
         ``False``. If ``True``, an error is raised if there are any duplicate
         values in any of the array-valued input arguments.
 
+    deep : bool, optional
+        Whether to deep copy a vtkRectilinearGrid object.
+        Default is ``False``.  Keyword only.
+
     Examples
     --------
     >>> import pyvista
@@ -122,13 +126,16 @@ class RectilinearGrid(_vtk.vtkRectilinearGrid, Grid, RectilinearGridFilters):
 
     _WRITERS = {'.vtk': _vtk.vtkRectilinearGridWriter, '.vtr': _vtk.vtkXMLRectilinearGridWriter}
 
-    def __init__(self, *args, check_duplicates=False, **kwargs):
+    def __init__(self, *args, check_duplicates=False, deep=False, **kwargs):
         """Initialize the rectilinear grid."""
         super().__init__()
 
         if len(args) == 1:
             if isinstance(args[0], _vtk.vtkRectilinearGrid):
-                self.deep_copy(args[0])
+                if deep:
+                    self.deep_copy(args[0])
+                else:
+                    self.shallow_copy(args[0])
             elif isinstance(args[0], (str, pathlib.Path)):
                 self._from_file(args[0], **kwargs)
             elif isinstance(args[0], np.ndarray):
@@ -419,6 +426,10 @@ class UniformGrid(_vtk.vtkImageData, Grid, UniformGridFilters):
     origin : iterable, optional
         Origin of the uniform grid.  Defaults to ``(0.0, 0.0, 0.0)``.
 
+    deep : bool, optional
+        Whether to deep copy a vtkImage3Data object.
+        Default is ``False``.  Keyword only.
+
     Examples
     --------
     Create an empty UniformGrid.
@@ -468,7 +479,13 @@ class UniformGrid(_vtk.vtkImageData, Grid, UniformGridFilters):
     _WRITERS = {'.vtk': _vtk.vtkDataSetWriter, '.vti': _vtk.vtkXMLImageDataWriter}
 
     def __init__(
-        self, uinput=None, *args, dims=None, spacing=(1.0, 1.0, 1.0), origin=(0.0, 0.0, 0.0)
+        self,
+        uinput=None,
+        *args,
+        dims=None,
+        spacing=(1.0, 1.0, 1.0),
+        origin=(0.0, 0.0, 0.0),
+        deep=False,
     ):
         """Initialize the uniform grid."""
         super().__init__()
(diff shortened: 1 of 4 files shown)


MatthewFlamm:
I also tested this PR with the Reader example in #3178, and the memory usage there is fixed.

pyvista/core/grid.py
```diff
@@ -419,6 +426,10 @@ class UniformGrid(_vtk.vtkImageData, Grid, UniformGridFilters):
     origin : iterable, optional
         Origin of the uniform grid.  Defaults to ``(0.0, 0.0, 0.0)``.
 
+    deep : bool, optional
+        Whether to deep copy a vtkImage3Data object.
```
MatthewFlamm, line 430:
```suggestion
        Whether to deep copy a vtkImageData object.
```

tests/test_grid.py
```diff
@@ -1400,3 +1409,21 @@ def test_ExplicitStructuredGrid_compute_connections():
     grid.compute_connections(inplace=True)
     assert 'number_of_connections' in grid.cell_data
     assert np.array_equal(grid.cell_data['number_of_connections'], connections)
+
+
+def test_copy_no_copy_wrap_object(datasets):
+    for dataset in datasets:
+        # different dataset tyoes have different copy behavior for points
```
MatthewFlamm, line 1416:
```suggestion
        # different dataset types have different copy behavior for points
```

tests/test_grid.py
```diff
@@ -1400,3 +1409,21 @@ def test_ExplicitStructuredGrid_compute_connections():
...
+        # use point data which is common
+        dataset["data"] = np.ones(dataset.n_points)
+        new_dataset = type(dataset)(dataset)
+        new_dataset["data"] += 1
+        assert np.array_equal(new_dataset["data"], dataset["data"])
+
+    for dataset in datasets:
+        # different dataset tyoes have different copy behavior for points
```
MatthewFlamm, line 1424:
```suggestion
        # different dataset types have different copy behavior for points
```

pyvista/core/pointset.py
```diff
@@ -1297,6 +1297,14 @@ class UnstructuredGrid(_vtk.vtkUnstructuredGrid, PointGrid, UnstructuredGridFilt
     - From cell, offset, and node arrays
     - From a file
 
+    Parameters
+    ----------
+    args : various
```
MatthewFlamm, line 1302: I don't like this, but I'm not even sure how the inputs for some of these work.  I wanted to put in a placeholder for further improvement.

pyvista/core/grid.py
```diff
@@ -95,6 +95,10 @@ class RectilinearGrid(_vtk.vtkRectilinearGrid, Grid, RectilinearGridFilters):
         ``False``. If ``True``, an error is raised if there are any duplicate
         values in any of the array-valued input arguments.
 
+    deep : bool, optional
+        Whether to deep copy a vtkRectilinearGrid object.
```
akaszynski, line 99:
```suggestion
        Whether to deep copy a ``vtk.vtkRectilinearGrid`` object.
```

pyvista/core/grid.py
```diff
@@ -419,6 +426,10 @@ class UniformGrid(_vtk.vtkImageData, Grid, UniformGridFilters):
     origin : iterable, optional
         Origin of the uniform grid.  Defaults to ``(0.0, 0.0, 0.0)``.
 
+    deep : bool, optional
+        Whether to deep copy a vtkImageData object.
```
akaszynski, line 430:
```suggestion
        Whether to deep copy a ``vtk.vtkImageData`` object.
```

pyvista/core/pointset.py
akaszynski: Out of scope for this PR, but we can add this later. Better than nothing.

We just need to update this in the same way we did with `UniformGrid`.

pyvista/core/pointset.py
```diff
@@ -2224,6 +2241,14 @@ class ExplicitStructuredGrid(_vtk.vtkExplicitStructuredGrid, PointGrid):
     - From a VTU or VTK file
     - From ``dims`` and ``corners`` arrays
 
+    Parameters
+    ----------
+    args : various
+        See below examples.
+    deep : optional
+        Whether to deep copy a vtkUnstructuredGrid object.
```
akaszynski, line 2249:
```suggestion
        Whether to deep copy a ``vtk.vtkUnstructuredGrid`` object.
```

akaszynski (approved):
Made a few nitpicks, otherwise LGTM.

adam-grant-hendry:
@MatthewFlamm @akaszynski I love this and thanks for doing it!

Quick question: does a shallow copy just return a reference to the object? If so, will users possibly mutate their data with this change when they're not expecting it? Is this a potentially a breaking change for existing code bases that previously relied on certain data sets performing deep copies by default?

MatthewFlamm:
> so, will users possibly mutate their data with this change when they're not expecting it? Is this a potentially breaking change for existing code bases that previously relied on certain data sets performing deep copies by default?

Yes you have it right that any change to the pyvista object will change the original vtk object.  I'm arguing that this is the expected contract for pyvista use with `deep=False`.  This was implemented for a few DataSets, e.g. `PolyData` and `UnstructuredGrid` but not others.  This PR aligns them all.  You are right that is a user was relying on the old behavior, then it will be a breaking change.  I'll tag it and put in the PR description, thanks for pointing this out!

adam-grant-hendry:
> > so, will users possibly mutate their data with this change when they're not expecting it? Is this a potentially breaking change for existing code bases that previously relied on certain data sets performing deep copies by default?
> 
> Yes you have it right that any change to the pyvista object will change the original vtk object. I'm arguing that this is the expected contract for pyvista use with `deep=False`. This was implemented for a few DataSets, e.g. `PolyData` and `UnstructuredGrid` but not others. This PR aligns them all. You are right that is a user was relying on the old behavior, then it will be a breaking change. I'll tag it and put in the PR description, thanks for pointing this out!

No worries! Thanks for aligning things: I agree we should be using `shallow_copy`. Thank you!

PR #3189, opened 2022-08-20
Add use_scalar_weights argument

### LINKED ISSUE #3188: Weight geodesic calculations by edge length

Hi, it would be really nice to be able to weight the shortest path calculations in `PolyData.geodesic` and `PolyData.geodesic_distance` by the edge length between vertices.

I'm not sure how best to incorporate this into pyvista, but I'd be happy to contribute it with a little guidance. Thanks!

It looks like there is an (experimental) option for this in vtk: https://vtk.org/doc/nightly/html/classvtkDijkstraGraphGeodesicPath.html#a7181bcce1675a88b5c8d9e6c3ca84e88

### Pseudocode or Screenshots

_No response_

tkoyama010:
### Overview

Add use_scalar_weights argument

 

resolves #3188

### Details

- @p-j-smith I made a draft PR. Could you test this branch that it works for you?


tests/test_polydata.py
```diff
@@ -222,6 +222,12 @@ def test_geodesic_distance(sphere):
     distance = sphere.geodesic_distance(0, sphere.n_points - 1)
     assert isinstance(distance, float)
 
+    # Use scalar weights
+    distance_use_scalar_weights = sphere.geodesic_distance(
+        0, sphere.n_points - 1, use_scalar_weights=True
+    )
+    assert isinstance(distance_use_scalar_weights, float)
+
```
adeak, line 230: Judging by CI this seems to segfault on VTK 8, might need a conditional skip.

p-j-smith: >  I get a segfault whether or not the scalars are point or cell data

As a side note, looking at the [vtk code](https://github.com/Kitware/VTK/blob/master/Filters/Modeling/vtkDijkstraGraphGeodesicPath.cxx#L158-L176) it seems it will only work with point data. Maybe worth mentioning this in the docstring?

p-j-smith:
Thank you for drafting this so quickly @tkoyama010! It works locally with python=3.9 and vtk=9.1

tests/test_polydata.py
adeak: Yeah, if that's the case then we should definitely mention that, maybe even raise if there aren't any point scalars. We have a few similar filters lying around in the codebase.

tkoyama010: Fix by d34157a801f206ed818698b40d2fcdac665c8800 .

banesullivan: I updated the docstring to include a not about VTK>=9 and point data. Further, I make sure the filter itself skips this option if not VTK9

banesullivan (dismissed):

banesullivan (approved):

PR #3209, opened 2022-08-25
Fix the error of pip option

tkoyama010:
### Overview

See https://pip.pypa.io/en/stable/cli/pip_install/#usage

 

### Details

- None


adeak (approved):
LGTM, it's good practice to put flags first unless we have a very good reason not to do that, for a given utility.

PR #3213, opened 2022-08-25
add reorder-edges example

akaszynski:
Converts https://github.com/pyvista/pyvista/discussions/3211 into an example.

### DIFF
--- a/examples/99-advanced/reorder-edges.py
+++ b/examples/99-advanced/reorder-edges.py
@@ -1,0 +1,97 @@
+"""
+.. _reorder_edges_example:
+
+Reorder Edges
+~~~~~~~~~~~~~
+
+This example demonstrates how to reorder edges based on connectivity.
+
+Boundary edges are extracted from a :class:`pyvista.PolyData` are unordered. We
+can use a simple mapping array based on :attr:`lines <pyvista.PolyData.lines>`
+to reorder the edges to make them sequential.
+
+"""
+
+import numpy as np
+
+import pyvista as pv
+from pyvista import examples
+
+###############################################################################
+# Clip dataset
+# ~~~~~~~~~~~~
+# Download an example dataset and clip it so we have open edges using the
+# :func:`clip <pyvista.DataSet.clip>` filter.
+dataset = examples.download_bunny_coarse()
+mesh = dataset.clip('y', invert=True)
+mesh.plot()
+
+
+###############################################################################
+# Extract and reorder edges
+# ~~~~~~~~~~~~~~~~~~~~~~~~~
+# Extract the edges and show they are unordered.
+#
+
+edge_data = mesh.extract_feature_edges(
+    boundary_edges=True, feature_edges=False, manifold_edges=False
+)
+
+# show the edges are unordered
+pl = pv.Plotter()
+pl.add_mesh(mesh)
+pl.add_mesh(
+    edge_data,
+    scalars=range(edge_data.n_points),
+    line_width=10,
+    show_scalar_bar=False,
+    cmap='plasma',
+)
+pl.show()
+
+
+###############################################################################
+# Reorder edges
+# ~~~~~~~~~~~~~
+# Reorder the edges and plot it.
+
+edges = edge_data.lines.reshape(-1, 3)[:, 1:]  # n_edges x 2
+u, v = edges.T
+
+# create a simple adjcency mapper array
+adj = np.empty(u.shape, dtype=u.dtype)
+adj[u] = v
+
+# Walk through each adjacent edge.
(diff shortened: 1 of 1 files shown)


examples/99-advanced/reorder-edges.py
```diff
@@ -0,0 +1,97 @@
...
+# create a simple adjcency mapper array
+adj = np.empty(u.shape, dtype=u.dtype)
+adj[u] = v
+
+# Walk through each adjacent edge.
+v = edges[0, 0]  # Starting vertex
+vert_idxs = [v]
+for _ in range(len(edges)):
+    v = adj[v]
+    vert_idxs.append(v)
```
adeak, line 70: I don't have much spare bandwidth at the moment, but this looks to me like it assumes that every line is oriented in the same way. In other words, every point along the closed (mangled) curve appears in the first position exactly once and in the second position exactly once.

In other words, it assumes we don't have lines like `[..., 2, 41, 42, 2, 43, 42]` in which case point `42` is never in the first position and the corresponding value in `adj` is left garbage.

The foolproof solution would be something like using [darikg's dict-based solution](https://github.com/pyvista/pyvista/discussions/3211#discussioncomment-3475803), defining adjacency in both directions (both `u -> v` and `v -> u`) and in the adjacency loop keeping track of seen points in a set.

It's plausible that surface orientation somehow leads to edges always being ordered in the same direction, but I don't know if we can take this for granted.

examples/99-advanced/reorder-edges.py
```diff
@@ -0,0 +1,97 @@
...
+.. _reorder_edges_example:
+
+Reorder Edges
+~~~~~~~~~~~~~
+
+This example demonstrates how to reorder edges based on connectivity.
+
+Boundary edges are extracted from a :class:`pyvista.PolyData` are unordered. We
+can use a simple mapping array based on :attr:`lines <pyvista.PolyData.lines>`
+to reorder the edges to make them sequential.
```
adeak, line 11: We should make it clear that this only works if you have a single, closed boundary edge. For surfaces with holes inside you'll get multiple boundary edges and you'd have to call `split_bodies()` or `connectivity()` or similar first in order to get connected components, then running this for a given closed loop.

examples/99-advanced/reorder-edges.py
akaszynski: Quite right. This algorithm breaks down if there isn't a single closed loop. Multiple loops don't work, only the first loop gets stored (and repeated because we iterate based on the number of edges).

akaszynski:
Closing for now until I find time for this.

darikg:
Sorry! just saw this thread now.

@adeak said: 
> It's plausible that surface orientation somehow leads to edges always being ordered in the same direction, but I don't know if we can take this for granted.

it can be taken for granted in the "nice" case, an oriented manifold. AFAIU igl and cgal leave undefined what happens otherwise. I guess if we wanted to implement boundary loop extraction in python it's ok to take a slower, safer approach, where you'd maintain a `visited_vertices: Set[int]` and check for loops

@adeak said:
> We should make it clear that this only works if you have a single, closed boundary edge. For surfaces with holes inside you'll get multiple boundary edges and you'd have to call split_bodies() or connectivity() or similar first in order to get connected components, then running this for a given closed loop.

yeah, [igl's boundary_loop](https://libigl.github.io/libigl-python-bindings/igl_docs/#boundary_loop) simply returns the longest loop; I think it'd make a lot more sense to accept a `start_vertex_idx` argument.

adeak:
I figured we could actually implement this as a helper in pyvista, allowing it to be auto-used at the end of `extract_feature_edges()`.

A rough algo:
1. build a `defaultdict(set)` adjacency graph
2. start from a full set of "needs visiting" point indices and an empty set of "visited" point indices
3. choose an arbitrary "needs visiting" point, keep looping arbitrary "adjacent" points whih haven't been "visited" yet, store them in a list for the new ordered edge. Each visited point goes in the "visited" set.
4. If for a given point we run out of "needs visiting" adjacent points, we either have a closed loop or we hit the end of an open edge. So reverse the order of the work-in-progress edge. If the other end has "needs visiting" adjacent points, continue until we run out of unvisited adjacent points. (For a closed edge this second step will have 0 new points added.)
5. this edge is now ordered, store this list in a list of lists containing all edges
6. goto point 3. until we run out of unvisited points
7. Now all the edge lists we've built are ordered. Generate the corresponding lines of length 2.

It should be straightforward, it should have O(N) memory need due to the reordered edges in sublists, and also O(N) runtime complexity because all lookups are O(1) and we make about as many steps as there are points.

Issue #1801, opened 2021-11-12
Consider refactoring example folder creation


adeak:
As discussed in review comments at https://github.com/pyvista/pyvista/pull/1794, the current strategy of creating example directories in `pyvista/__init__.py` is suboptimal. The issue is that choosing to raise when the example directories can't be created for whatever reason is not an option with the current implementation, because this could affect many users who don't use examples at all.

It seems reasonable from a design standpoint to refactor example directory creation to only happen when the directory would be first used. In that case we could raise when something's off, and let the user try again in the correct directory or with the correct permissions. The drawback is that the implementation would probably be more complicated, with keeping track of whether the example directory has been initialised, and example calls having to check that each time.

cc. @MatthewFlamm

MatthewFlamm:
I think there could be one helper function that is invoked at the beginning of each download function.  This wouldn't be too complex.  It could also be a decorator to keep the syntax out of the function.

I'm also wondering if it can be done when doing: `from pyvista import examples`.  When importing this, it is much more likely that a file will be downloaded.

adeak:
> I'm also wondering if it can be done when doing: `from pyvista import examples`. When importing this, it is much more likely that a file will be downloaded.

Great idea! That would both lead to very few false positives, and at the same time it would likely raise very early for true positives (because imports are normally at the top).

MatthewFlamm:
I'm thinking that in this case, `EXAMPLES_PATH` is ideally moved under the `examples` namespace rather than the top level `pyvista` namespace, because it takes a value only after importing `examples`.  This would be a breaking change, although I'm not sure how many users would be affected.

adeak:
Couldn't we just keep it in the `pyvista` main namespace, initialise it to `None` and check for this in `pyvista.examples`?

banesullivan:
> Couldn't we just keep it in the pyvista main namespace, initialise it to None and check for this in pyvista.examples?

+1. I have relied on `EXAMPLES_PATH` being in the `pyvista` namespace. I dealing, I'd like for it to be created on import if it doesn't exist - I'm in favor of just raising a warning when this fails and have the `examples` check that value and error if the data cannot download to that path

PR #3218, opened 2022-08-27
Use pooch

### LINKED ISSUE #210: Pooch integration for downloading examples
We should change our example downloading utilities to leverage Pooch: https://github.com/fatiando/pooch

Not sure how much effort this will require, but the download utility I made is a bit unpredictable on the CI's - downloads time out a lot... not sure if Pooch would fix that but it's probably worth switching to Pooch to make downloading data files outside of our example data repo a bit easier

akaszynski:
Resolve #210 and #1801 by using [pooch](https://pypi.org/project/pooch/).

Just about all of the download logic is now being accomplished by using `pooch`. This PR also:
- Simplifies testing of the downloads by allowing for ``pytest --test_downloads``.
- Clean up several examples that were broken.
- Improve test coverage (when enabled).


MatthewFlamm:
I'm +1 on delegating logic to another package here.

I haven't gone through in detail, but it looks like the ability to use a local git repo will be removed with this PR.  I am a user of this feature, but I'm not sure if there are others.  I don't see an option for this in pooch, so I will add a feature request there.

adeak:
> I haven't gone through in detail, but it looks like the ability to use a local git repo will be removed with this PR.

@MatthewFlamm could you please explain in a few words what that use case is? Having a local clone of `vtk-data` and using downloads from there? Or something else?

pytest.ini
```diff
@@ -14,3 +14,4 @@ testpaths = tests
 markers =
     needs_vtk9: skip test unless VTK is 9.0 or newer.
     needs_vtk_version(version): skip test unless VTK version is at least as specified.
+    skip_download
```
adeak, line 17: Semantically I think this should be something like 
```suggestion
    needs_download: this test downloads data during execution
```
because the "skip" part comes from the new `--test_downloads` flag. What that should do is skip downloads that are marked as "needs download" (or "has download" or "downloads" or whatever). So my point is that skip should not be in the name of the marker. (And I'd definitely recommend using a description for the marker, not necessarily the one I wrote above).

MatthewFlamm:
> > I haven't gone through in detail, but it looks like the ability to use a local git repo will be removed with this PR.
> 
> @MatthewFlamm could you please explain in a few words what that use case is? Having a local clone of `vtk-data` and using downloads from there? Or something else?

I often use a machine with a lot of memory, etc. that has no outside internet access, but it shares a filesystem with a machine that does. I thought of two solutions: 

1. Give functionality to download all the data and populate the entire cache.  Then share the cache location between the machines.  A big downside is that pyvista versions on each machine would have to approximately match due to the cache versioning.  I'm not sure if using pooch changes this downside.  The big upside is that it is more general.  We don't have a mechanism to do this with our current infrastructure, but `pooch` might.  I need to look into this.
2. Allow for referencing a local path for a git repo.  This has two downsides: duplication of the data in storage and it doesn't work well when you have multiple sources of data.  The upside is that it is more robust to changes in the cache format since we keep an original copy of the data.  I think this is now the inferior option as more outside sources have been added to PyVista.  But it is a current functionality.

MatthewFlamm:
For 1 above, we might be able to add a `download_all` function which will cache everything.

For 2 above, pooch allows for a custom downloader.  This would be a path forward if this method is preferred.

akaszynski:
> I haven't gone through in detail, but it looks like the ability to use a local git repo will be removed with this PR. I am a user of this feature, but I'm not sure if there are others. I don't see an option for this in pooch, so I will add a feature request there.

I’ve kept this functionality, but now the environment variable is `PYVISTA_USERDATA_PATH`. I’ll document this.

pyvista/examples/downloads.py
```diff
@@ -10,304 +10,327 @@
...
+CACHE_VERSION = 3
 
+# If available, a local vtk-data instance will be used for examples
 
-def _http_request(url, progress_bar=False):  # pragma: no cover
-    """Download a file from a url using ``urlretrieve``.
+VTK_DATA_PATH: Optional[str] = None
+if 'PYVISTA_VTK_DATA' in os.environ:  # pragma: no cover
+    warnings.warn(
+        'PYVISTA_VTK_PATH has been deprecated as of 0.37.0\n\n' 'Please use PYVISTA_USERDATA_PATH'
```
MatthewFlamm, line 33: I don't think these environment variables are used the same.  One sets the path of the cache.  The other set a local git repo of `pyvista/vtk-data` used as a source for the cache.

If we keep this, we should warn that this is no longer supported and not set `PYVISTA_USERDATA_PATH`.

akaszynski: Fixed in f6e5506. I've reverted to the previous behavior.

akaszynski:
> 1. Give functionality to download all the data and populate the entire cache.  Then share the cache location between the machines.  A big downside is that pyvista versions on each machine would have to approximately match due to the cache versioning.  I'm not sure if using pooch changes this downside.  The big upside is that it is more general.  We don't have a mechanism to do this with our current infrastructure, but `pooch` might.  I need to look into this.
> 2. Allow for referencing a local path for a git repo.  This has two downsides: duplication of the data in storage and it doesn't work well when you have multiple sources of data.  The upside is that it is more robust to changes in the cache format since we keep an original copy of the data.  I think this is now the inferior option as more outside sources have been added to PyVista.  But it is a current functionality.

I've re-implemented 2. In my opinion, this is the better option as you're treating the `SOURCE` the same in either the remote or local case, leading to a much simpler module. With a remote cache, you simply download from a URL, and in the "file" case, you're simply copying a file using the OS.

> I think this is now the inferior option as more outside sources have been added to PyVista.  But it is a current functionality.

If you really wanted to have a single cache with multiple clients accessing it, you'd simply have to set `PYVISTA_USERDATA_PATH` to that directory (probably a network share). In the current implementation, `CACHE_VERSION` only matters when `PYVISTA_USERDATA_PATH` is unset.

pyvista/examples/downloads.py
```diff
@@ -10,304 +10,364 @@
...
-    return True
-
-
-def _decompress(filename, output_path):
-    _check_examples_path()
-    zip_ref = zipfile.ZipFile(filename, 'r')
-    zip_ref.extractall(output_path)
-    return zip_ref.close()
+    for fname in fnames:
+        if os.path.basename(fname) == target_file:
```
MatthewFlamm, line 277: The OpenFoam tests are failing because there are multiple matches here in a zip file.  We need to match the full path.

banesullivan: Could we use `members` of the pooch's `Unzip` class to work around this? https://www.fatiando.org/pooch/latest/api/generated/pooch.Unzip.html

MatthewFlamm: We need >30 files with preserved folder structure for the OpenFoam tests.  While it is possible to work around it this way, it would require specifying all the files.  IMO, it would be more streamlined to optionally match a full or partial path for the one file that is needed to specify for the reader, but I'm not insisting on it.

If we do keep the current functionality of matching any `basename` or a partial path, then we should raise or warn if multiple files would match.  As we see, this will cause different results on different OS.

MatthewFlamm: The most important thing is that we will want to use the other data in this zip file when new functionality is added to `OpenFoamReader`.  When this happens, we wouldn't be able to exclude anything using `members`.

MatthewFlamm:
> I've re-implemented 2.

It isn't obvious from the pooch documentation that this was possible.  I like this idea in that it doesn't make a large breaking change from current behavior.  I think it should also be possible to implement the first idea, but that can wait for another PR.

akaszynski:
Ready for review. I ran the entire documentation build locally without any issues.

CONTRIBUTING.rst
```diff
@@ -472,6 +472,53 @@ This effectively invokes ``SPHINXOPTS=-j`` and can be especially useful for
...
+^^^^^^^^^^^^^^^^^^^^
+PyVista's examples come in two formats: basic code snippets demonstrating the
+functionality of an individual method, or a full gallery example displaying one
+or more concepts.  Small code samples and snippets are contained in the
+``doc/api`` directory or within our documentation strings, while the full
+gallery examples, meant to be run as individual downloadable scripts, are
+contained in the ``examples`` directory at the root of this repository.
+
+To add a fully fledged, standalone example, add your example to the
+``examples`` directory within one of the applicable subdirectories.  Should
```
banesullivan, line 500: We may want to be more clear that this is the `examples/` directory at the root level of the repository and not the subpackage within `pyvista`

CONTRIBUTING.rst
```diff
@@ -472,6 +472,53 @@ This effectively invokes ``SPHINXOPTS=-j`` and can be especially useful for
...
+<https://sphinx-gallery.github.io/stable/index.html>`_.
+
+For more details see :ref:`add_example_example`.
+
+
+Add a new Example File
+^^^^^^^^^^^^^^^^^^^^^^
+If you have a dataset that you need for your gallery example, add it to
+`pyvista/vtk-data <https://github.com/pyvista/vtk-data/>`_ and follow the
+directions there and in ``downloads.py``. This enables:
```
banesullivan, line 514: we should elaborate that developers may need to add there own `download_*` method in `pyvista.examples.downloads`

examples/99-advanced/add-example.py
```diff
@@ -0,0 +1,184 @@
...
+dataset.plot(text='Example Figure')
+
+
+###############################################################################
+# Caveat - Plotter must be within One Cell
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~---------
+# It's not possible for a single :class:`pyvista.Plotter` object across
+# multiple cells because these are closed out automatically at the end of a
+# cell.
+#
```
banesullivan, line 86: Is this new behavior from a recent refactor??

setup.py
```diff
@@ -17,6 +17,7 @@
     'imageio',
     'pillow',
     'appdirs',
+    'pooch',
```
banesullivan, line 20: Don't we need to pin this in `requirements_test.txt` as well?

tests/examples/test_download_files.py
```diff
@@ -0,0 +1,895 @@
...
+import numpy as np
+import pytest
+
+import pyvista as pv
+from pyvista import examples
+
+if 'TEST_DOWNLOADS' in os.environ:
+    warnings.warn('"TEST_DOWNLOADS" has been deprecated. Use `pytest --test_downloads`')
+
+pytestmark = pytest.mark.needs_download
```
banesullivan, line 21: Doesn't this need to decorate all these functions? or how is this working?

banesullivan: Okay this is good to know - TIL

But is it a problem that we do this here:

https://github.com/pyvista/pyvista/blob/7d65054dff4717338519e73d287c7709caef1b20/tests/utilities/test_reader.py#L12-L14

Are we then skipping this whole module on Mac? Also the fact that it says "Azure" there probably means this is super old and needs to be cleaned up

adeak: Looks like it, yup.

banesullivan: Leaving this for another issue/PR: https://github.com/pyvista/pyvista/issues/3349#issue-1381158342

pyvista/examples/downloads.py
```diff
@@ -4561,7 +4459,7 @@ def download_black_vase(load=True, progress_bar=False):  # pragma: no cover
     Original datasets are under the CC BY 4.0 license.
 
     For more details, see `Ivan Nikolov Datasets
-    <https://github.com/pyvista/vtk-data/tree/master/Data/ivan-nikolov>`_
+    <https://github.com/pyvista/vtk-data/tree/master/ivan-nikolov>`_
```
MatthewFlamm, line 4462: All of these changes removing `Data` from the url lead to 401 errors. The original links were correct.

MatthewFlamm (commented):
I tested this and it works for me with and without local `vtk-data` repo.

pyvista/examples/downloads.py
```diff
@@ -3,311 +3,240 @@
...
-            )
-        retriever = partial(_repo_file_request, pyvista.VTK_DATA_PATH, filename)
-
-    if pyvista.get_ext(filename) == '.zip':
-        return _retrieve_zip(retriever, filename)
-    return _retrieve_file(retriever, filename)
+    if os.path.isdir(PATH):
+        shutil.rmtree(PATH)
+    os.makedirs(PATH)
+    return True
```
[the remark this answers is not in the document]
akaszynski: This was here prior and I agree a return isn't helpful here.

examples/99-advanced/add-example.py
```diff
@@ -0,0 +1,184 @@
...
+###############################################################################
+# Plots and images
+# ~~~~~~~~~~~~~~~~
+# If you use anything that outputs an image (for example,
+# :func:`pyvista.Plotter.show`) the resulting image will be rendered within the
+# output HTML.
+#
+# .. note::
+#    Unless ``sphinx_gallery_thumbnail_number = <int>`` is included at the top
+#    of the example script, first figure (this one) will be used for the
```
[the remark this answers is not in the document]
banesullivan: I think 1-based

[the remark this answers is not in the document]
banesullivan: https://sphinx-gallery.github.io/stable/configuration.html?highlight=sphinx_gallery_thumbnail_number#choosing-the-thumbnail-image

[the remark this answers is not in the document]
MatthewFlamm: It would be helpful to state this explicitly here.  Then contributors don't have to look.

examples/99-advanced/add-example.py
akaszynski: No. Issue raised for this.

pyvista/__init__.py
```diff
@@ -70,39 +61,6 @@
...
-    except Exception as e:
-        warnings.warn(
-            f'Unable to create `PYVISTA_USERDATA_PATH` at "{USER_DATA_PATH}"\n'
-            f'Error: {e}\n\n'
-            'Override the default path by setting the environmental variable '
-            '`PYVISTA_USERDATA_PATH` to a writable path.'
-        )
-        USER_DATA_PATH = ''
-
-EXAMPLES_PATH = os.path.join(USER_DATA_PATH, 'examples')
```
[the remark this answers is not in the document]
akaszynski: Behavior still exists, I've just moved it to `downloads.py`

pyvista/examples/downloads.py
```diff
@@ -3,311 +3,240 @@
 Once downloaded, these datasets are stored locally allowing for the
 rapid reuse of these datasets.
 
+Files are all hosted in https://github.com/pyvista/vtk-data/ and are downloaded
+using the ``download_file`` function. If you add a file to the example data
+repository, you should add a ``download-<dataset>`` method here which will
+rendered on `downloads
+<https://docs.pyvista.org/api/examples/_autosummary/pyvista.examples.downloads.html>`_.
```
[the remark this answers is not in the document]
akaszynski: It's the exact same page, so there's no reason to add a reference other than to state "this page".

pyvista/examples/downloads.py
```diff
@@ -3,311 +3,240 @@
...
 
+CACHE_VERSION = 3
 
-def _decompress(filename, output_path):
-    _check_examples_path()
-    zip_ref = zipfile.ZipFile(filename, 'r')
-    zip_ref.extractall(output_path)
-    return zip_ref.close()
+# If available, a local vtk-data instance will be used for examples
+if 'VTK_DATA_PATH' in os.environ:  # pragma: no cover
```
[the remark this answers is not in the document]
akaszynski: Changed back to `PYVISTA_VTK_DATA`.

pyvista/examples/downloads.py
```diff
@@ -3,311 +3,240 @@
...
+    _path = os.environ['VTK_DATA_PATH']
 
+    if not os.path.basename(_path) == 'Data':
+        # append 'Data' if user does not provide it
+        _path = os.path.join(_path, 'Data')
+    if not os.path.isdir(_path):
+        warnings.warn(f"VTK_DATA_PATH: {_path} is an invalid path and will be ignored")
```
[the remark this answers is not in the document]
akaszynski: Correct. I've deleted this branch.

pyvista/examples/downloads.py
```diff
@@ -3,311 +3,240 @@
...
+
+# allow user to override the local path
+if 'PYVISTA_USERDATA_PATH' in os.environ:  # pragma: no cover
+    if not os.path.isdir(os.environ['PYVISTA_USERDATA_PATH']):
+        warnings.warn('Ignoring invalid {PYVISTA_USERDATA_PATH')
+    else:
+        PATH = os.environ['PYVISTA_USERDATA_PATH']
+else:
+    # use default pooch path
+    PATH = str(pooch.os_cache(f'pyvista_{CACHE_VERSION}'))
```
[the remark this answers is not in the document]
akaszynski: Changed to `USER_DATA_PATH`.

pyvista/examples/downloads.py
```diff
@@ -3,311 +3,240 @@
...
-def _get_vtk_file_url(filename):
-    return f'https://github.com/pyvista/vtk-data/raw/master/Data/{filename}'
+    # provide helpful message if pooch path is inaccessible
+    if not os.path.isdir(PATH):  # pragma: no cover
+        try:
+            os.makedirs(PATH, exist_ok=True)
+            if not os.access(PATH, os.W_OK):
+                raise OSError
+        except (PermissionError, OSError):
+            warnings.warn(
```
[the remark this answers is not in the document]
akaszynski: Normally I'd say this would be a best practice, but in practice this sometimes causes anything depending on PyVista's example module to fail on import because of permission errors.

Regardless I'm adding a note.

setup.py
akaszynski: Added.

pyvista/examples/downloads.py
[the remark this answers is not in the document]
MatthewFlamm: I forgot that we have non downloaded data in this module too.   Ignore these comments about converting to errors.

akaszynski:
@banesullivan, ready for review (again).

MatthewFlamm:
I think `appdirs` is no longer a required dependency for `pyvista`.  And we should add `pooch` to `pv.Report()` and remove `appdirs` from the same.

setup.py
MatthewFlamm, line 20: I am 50/50 on whether it would be good to have this as an `extras.require` instead.  In a significant fraction of use cases, users will never import `pyvista.examples`.  But, it makes getting started more frustrating.  Unfortunately, Python doesn't support a default extras for installation, which would solve this problem.

Thoughts?

banesullivan: Pooch is lightwieght enough that I think it should be included by default - primarily for keeping the barrier to entry low so that users have tons of examples that will work out of the box

banesullivan: +1 to @MatthewFlamm's comment about appdirs. I think it can now be safely removed

banesullivan (dismissed):
I'm quite happy with how this turned out, thanks @akaszynski for the hard work here!

Minor things left
- `appdirs` needs to be removed
- differing to all other feedback from @MatthewFlamm

banesullivan:
FYI: we'll have to update the conda recipe on/before the next release

PR #3230, opened 2022-08-30
Rename method: overwrite -> copy_from

### LINKED ISSUE #3142: Need to rename `overwrite` method
https://github.com/pyvista/pyvista/blob/7ca5c227f9111dd1c280da06a3959e37ec047541/pyvista/core/dataset.py#L2016

The `overwrite` method is a PyVista-specific method that many of our users have found valuable. This previously had no equivalent upstream in VTK, but now, the `overwrite` method is used upstream in VTK after 9.2 for a different use case. Every class in VTK, as of 9.2, now has an `overwrite` method for overwriting the implementation of the Python class. This creates a clash where PyVista might incorrectly change the behavior of VTK and e.g., `vtk.vtkPolyData.overwrite != pyvista.PolyData.overwrite`

I am working on some upstream tooling that will later be inherited by PyVista, where this `overwrite` method is critical.

In summary, the name of this method needs to change because:

1. It clashes with an upstream VTK method as of 9.2
2. It will be used for a different mechanism in PyVista at a later date

### What should this method be renamed to?

I don't have any good ideas for a new name of this method, but we need to change it as soon as possible with a deprecation notice so that it doesn't conflict with VTK 9.2 for too long.

banesullivan:
Resolves #3142 by renaming the `overwrite` method to `copy_from`


banesullivan:
This is actually no longer critical as the upstream method is actually `override` now, not `overwrite`. Still, though, I think this change is warranted to improve clarity and the self-documenting-ness of the API

pyvista/core/dataset.py
```diff
@@ -2115,6 +2115,24 @@ def overwrite(self, mesh: _vtk.vtkDataSet):
...
+        ----------
+        mesh : vtk.vtkDataSet
+            The overwriting mesh.
+
+        """
+        warnings.warn(
+            "Use of `DataSet.overwrite` is deprecated. Use `DataSet.copy_from` instead.",
+            PyvistaDeprecationWarning,
+        )
+        return self.copy_from(mesh)
```
MatthewFlamm, line 2134: numpydoc will require a `Returns` section when you return something.  `copy_from` doesn't return, so we can do that here too.  Or, a section could be added in the docstring.

```suggestion
        self.copy_from(mesh)
```

MatthewFlamm (commented):
LGTM.

banesullivan:
> I would always have to check which object is being overwritten. copy_from is self explanatory.

Self-documenting APIs are the way to go!

adeak:
> This is actually no longer critical as the upstream method is actually `override` now, not `overwrite`.

Great news, by the way.

banesullivan:
Would like to merge this. Reviews would be greatly appreciated!

pyvista/core/dataset.py
```diff
@@ -2117,6 +2117,24 @@ def overwrite(self, mesh: _vtk.vtkDataSet):
         if is_pyvista_dataset(mesh):
             self.copy_meta_from(mesh, deep=True)
 
+    def overwrite(self, mesh: _vtk.vtkDataSet):
+        """Overwrite this dataset inplace with the new dataset's geometries and data.
+
+        .. deprecated:: 0.37.0
+            Use :attr:`DataSet.copy_from` instead.
```
adeak, line 2124: Why not
```suggestion
            Use :func:`DataSet.copy_from` instead.
```
?

tests/test_dataset.py
```diff
@@ -26,7 +26,13 @@ def grid():
 
 def test_invalid_overwrite(grid):
     with pytest.raises(TypeError):
-        grid.overwrite(pyvista.Plane())
+        grid.copy_from(pyvista.Plane())
+
+
+def test_overwirte_deprecation(grid):
```
adeak, line 32:
```suggestion
def test_overwrite_deprecation(grid):
```

examples/02-plot/mesh-picking.py
```diff
@@ -35,7 +35,7 @@
 def callback(mesh):
     """Shrink the mesh each time it's clicked."""
     shrunk = mesh.shrink(0.9)
-    mesh.overwrite(shrunk)  # must operate "in-place" by overwrite
+    mesh.copy_from(shrunk)  # must operate "in-place" by overwrite
```
adeak, line 38: The comment sounded a bit clumsy to begin with, but now it's even more off. How about something like
```suggestion
    mesh.copy_from(shrunk)  # make operation "in-place" by replacing the original mesh
```

adeak (commented):
Few minor comments, nothing serious.

pyvista/core/dataset.py
banesullivan: Thanks!

tests/test_dataset.py
banesullivan: dyslexia is real

examples/02-plot/mesh-picking.py
banesullivan: I like this version, thanks

banesullivan:
Thank you for the edits, @adeak!

adeak (approved):
LGTM, thanks!

PR #3231, opened 2022-08-30
Pyvista -> PyVista

banesullivan:
Many places in the documentation and class names have PyVista improperly capitalized as "Pyvista". This PR bulk changes occurrences of "Pyvista" to "PyVista".

This does affect class names, notably `ResetPyVista`, `PyVistaDeprecationWarning`, and `PyVistaFutureWarning`. I believe these are intended for internal use and so I did not add a deprecation notice.

Note: we should probably add Pyvista -> PyVista upstream in codespell.

### DIFF
--- a/doc/api/plotting/qt_plotting.rst
+++ b/doc/api/plotting/qt_plotting.rst
@@ -1,6 +1,6 @@
 .. _qt_plotting:
 
-Pyvista and PyQt
+PyVista and PyQt
 ----------------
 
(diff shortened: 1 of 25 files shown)


tkoyama010 (approved):
LGTM. Good catch!

adeak:
I think of the class names as capcased versions of `pyvista`, the package name. So I don't really agree with changing those. But in either case it's trivial, just pointing this out.

akaszynski (approved):

MatthewFlamm:
This is a breaking change for any users who are handling `Pyvista*Warning`.  Downstream packages may be using these for CI for example.  I added the breaking-change label for easier user discovery.  I have been starting to use a `Breaking Change` section in PR descriptions to make it further discoverable.   But, I think the description already covers this in some ways, so not neccessary.

banesullivan:
Thank you for adding that label!

PR #3232, opened 2022-08-30
Standardize warnings

### LINKED ISSUE #838: Standardize warnings
We need to standardize how warnings are given in PyVista. This was brought to my attention in https://github.com/pyvista/pyvista/pull/837#discussion_r454749250 and after some investigation, I think we want to use `warnings.warn('message')` everywhere. There are lots of places where we use `logging.warning` that need to be switched. 

Why? Because `warnings.warn` can be understood and captured more easily by code runners, etc. For example, Azure has a nice panel of all these warnings thrown when running our tests:

Screen Shot 2020-07-16 at 7 44 26 AM

I'm a bit weary to use `DeprecationWarning` with this because they are typically ignored by default and we do not want those to be ignored. See https://docs.python.org/3/library/warnings.html#warning-categories

> Base category for warnings about deprecated features (ignored by default).

So perhaps it's good to just stick with `UserWarning`.

Also, I very much dislike how a `warning.warn` is only shown once. IMO, it should be thrown as many times as the situation is encountered to pester the user into updating their code to avoid deprecated features.

@pyvista/developers, any preference on switching how our warnings are thrown and possibly creating a new helper method or class to handle this universally across PyVista?

banesullivan:
resolves #838


pyvista/core/filters/poly_data.py
```diff
@@ -1722,7 +1721,7 @@ def fill_holes(self, hole_size, inplace=False, progress_bar=False):  # pragma: n
         >>> assert edges.n_cells == 0  # doctest:+SKIP
 
         """
-        logging.warning('pyvista.PolyData.fill_holes is known to segfault. Use at your own risk')
+        warnings.warn('pyvista.PolyData.fill_holes is known to segfault. Use at your own risk')
```
MatthewFlamm, line 1724: This type of warning will be annoying as it is thrown every time the filter is called.  Maybe move it into the docstring?

We could also have a special pyvista warning subclass for this kind of usage.  A user may have a use case where they know it is generally robust and would prefer to silence this warning.

pyvista/utilities/helpers.py
```diff
@@ -960,7 +960,7 @@ def wrap(dataset):
         try:
             return pyvista._wrappers[key](dataset)
         except KeyError:
-            logging.warning(f'VTK data type ({key}) is not currently supported by pyvista.')
+            warnings.warn(f'VTK data type ({key}) is not currently supported by pyvista.')
```
MatthewFlamm, line 963: This one seems like we should raise an error instead.  I don't see how a user would be able to proceed after this point without handling it.

MatthewFlamm (commented):
I'm +1, in general, on these changes.

pyvista/utilities/helpers.py
adeak: I was about to say that we rely on this behaviour in `readers.py`, but actually that's just another raise, just like you suggest here
https://github.com/pyvista/pyvista/blob/31f4886cd42742182be56e16b5cb26af5265ffa8/pyvista/utilities/reader.py#L306-L308

pyvista/core/filters/poly_data.py
banesullivan: That's a great point! I like your idea; perhaps a `PyVistaPerformanceWarning`

pyvista/utilities/helpers.py
banesullivan: Thank's for point that out, @adeak! I'll see if I can clean this up within this PR

pyvista/core/filters/poly_data.py
banesullivan: Done

banesullivan:
I think this PR is ready to go. I did notice that `logger.warning` is still used in `pyvista/ext/coverage.py` though... I feel like that module can be an exception but what are your thoughts, @pyvista/developers ?

banesullivan:
Merging this

PR #3237, opened 2022-08-30
🚧 Add add_cell_labels

tkoyama010:
### Overview

Add `add_cell_labels` .

#### main
```python
import pyvista

radius = 0.5
circle = pyvista.Circle(radius)
plotter = pyvista.Plotter()
plotter.add_mesh(circle)
points = [circle.center]
labels = ["1234"]
plotter.add_point_labels(points=points, labels=labels)
plotter.show(cpos="xy", before_close_callback=verify_cache_image)
```

#### PR
```python
import pyvista

radius = 0.5                                                                                                                                                                                                  
circle = pyvista.Circle(radius)                                                                                                                                                                               
plotter = pyvista.Plotter()                                                                                                                                                                                   
plotter.add_mesh(circle)                                                                                                                                                                                      
cells = [circle]                                                                                                                                                                                              
labels = ["1234"]                                                                                                                                                                                             
plotter.add_cell_labels(cells=cells, labels=labels)                                                                                                                                                           
plotter.show(cpos="xy")
```

add_cell_labels
 

### Details

- Inspired by https://github.com/pyvista/pyvista/discussions/3235.

- [x] ~Use [pyvista.MultiBlock](https://docs.pyvista.org/api/core/_autosummary/pyvista.MultiBlock.html?highlight=multiblocks)~


banesullivan (changes requested):
I'm not fully on board with this proposed change and would like to avoid adding any more thin wrappers of existing methods to the plotter. 

IMO, PyVista's data-centric API suggests that the cell centers should be extracted as a new dataset prior to plotting and that the plotting API's `add_point_labels` is agnostic to whether those points are the nodes of a mesh, the cell centers of a mesh, or the centers of the edges of a mesh, etc.

This would be effectively equivalent to:

```py
centers = mesh.cell_centers()
labels = ...

p = pv.Plotter()
p.add_point_labels(centers, labels)
p.show()
```

banesullivan:
Sorry to be discouraging, @tkoyama010! I love the fact that you're seeing ways in which you'd like to use the API and suggesting these changes, though! I'm just one voice here, so definitely solicit feedback from others.

akaszynski:
Agree with @banesullivan, however it may not be immediately apparent how to add labels at cells. @tkoyama010, a basic example of this would be excellent, and we can even explain how to do it at cell, face, and edges.

banesullivan:
Ah yes - adding a comprehensive labelling example (expanding from what already exists) would be highly valuable

tkoyama010:
Thank you for your opinions. It is exactly what I was looking for. I was also 🚧 unsure about this proposal. I like the idea of limiting the changes to adding examples only. I will switch my PR to that.

tkoyama010:
- [ ] Open new Pull Request.

PR #3238, opened 2022-08-30
Bump pydata-sphinx-theme from 0.9.0 to 0.10.1 and fix warnings

tkoyama010:
### Overview

Fix deprecated warning of #3236 .
 

closes #3236 

### Details

- https://github.com/pydata/pydata-sphinx-theme/pull/878

### DIFF
--- a/doc/_templates/sidebar-nav-bs.html
+++ b/doc/_templates/sidebar-nav-bs.html
@@ -1,9 +1,9 @@
 <nav class="bd-links" id="bd-docs-nav" aria-label="Main navigation">
   <div class="bd-toc-item active">
     {% if pagename.startswith("api") %}
-    {{ generate_nav_html("sidebar", maxdepth=4, collapse=True, includehidden=True, titles_only=True) }}
+    {{ generate_toctree_html("sidebar", maxdepth=4, collapse=True, includehidden=True, titles_only=True) }}
     {% else %}
-    {{ generate_nav_html("sidebar", maxdepth=4, collapse=False, includehidden=True, titles_only=True) }}
+    {{ generate_toctree_html("sidebar", maxdepth=4, collapse=False, includehidden=True, titles_only=True) }}
     {% endif %}
   </div>
 </nav>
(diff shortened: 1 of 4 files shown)


banesullivan:
@tkoyama010, can you do this directly in #3236 ?

adeak:
> @tkoyama010, can you do this directly in #3236 ?

@banesullivan I recommend against doing that. I think there's great value in knowing that dependabot PRs are trivial for what they are: version bumps. If changes are necessary to accommodate version bumps, have a dedicated PR that does that. It's great for transparency, it's great for `git bisect`, it's the right call in my opinion.

banesullivan:
Fair enough -- I typically like to see any needed changes included in the dependabot PRs. My typical mode of operation is to let Dependabot bump the deps, if it fails, make the changes in that existing PR, squash and merge to see the bump and the needed changes together. This is just how I, personally, like to manage dependency issues.

Whichever approach we take is unimportant. If you all prefer, we can keep the changes here, we just need to close #3236 then

adeak:
> If you all prefer, we can keep the changes here, we just need to close #3236 then

Why do we have to close that?

banesullivan:
The version of the package *has* to be bumped with these changes (AFAICT) due to the API changing at that version. Also, I think it's cleaner to have the bump and the needed changes for the bump all in the same PR and thus same commit

banesullivan:
> Did the newest version introduce the non-deprecated alternative?

Even if so, IMO, we should track all of this in a single PR since we pin our dependencies for a cleaner history

banesullivan:
Actually, the issue is that we run `make html` with the `-W` flag to turn warnings/deprecations into errors. So if we bump there version, we have to include the needed changes. And we can't just merge the needed changes, because that part of the API doesn't exist in the previous version

banesullivan:
> I'd prefer this PR for my aforementioned reasons

Fair enough! I added a `closes` command to the root description to supersede #3236

banesullivan (approved):
Thank you for handling this, @tkoyama010!!

Issue #3164, opened 2022-08-14
UniformGrid cannot be cast to pointset


MatthewFlamm:
### Describe the bug, what's wrong, and what you expected.

`UniformGrid` has a `cast_to_pointset` method, but it fails when used.

### Steps to reproduce the bug.

```python
import pyvista as pv
mesh = pv.UniformGrid(dims=(5,5,5))
mesh.cast_to_pointset()
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/matt/pyvista/pyvista/core/dataset.py", line 2125, in cast_to_pointset
    pset.SetPoints(self.GetPoints())
  File "/home/matt/pyvista/pyvista/core/dataset.py", line 119, in __getattr__
    return super().__getattribute__(item)
AttributeError: 'UniformGrid' object has no attribute 'GetPoints'
```

### System Information

```shell
GPU Version : 4.5 (Core Profile) Mesa 20.3.5

  Python 3.9.2 (default, Feb 28 2021, 17:03:44)  [GCC 10.2.1 20210110]

           pyvista : 0.37.dev0
               vtk : 9.1.0
             numpy : 1.22.4
           imageio : 2.19.3
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.2
           IPython : 8.4.0
          colorcet : 3.0.0
           cmocean : 2.0
        ipyvtklink : 0.2.2
             scipy : 1.8.1
              tqdm : 4.64.0
            meshio : 5.3.4
        jupyterlab : 3.4.4
         pythreejs : Version unknown
--------------------------------------------------------------------------------
```

banesullivan:
Commenting here, but this is sort of feedback for #3242, cc @MatthewFlamm 

I think we should rewrite `cast_to_pointset` and have it on the `DataSet` class as every dataset type does fundamentally have points that a user may want to extract. IMO, this should use PyVista's `points` property to access the points (avoiding the issue of data types with implicit points like `ImageData` that do not have a `GetPoints()` method).

It should also have an option to transfer the cell data with the `cell_data_to_point_data` filter

banesullivan:
Prototype implementation:

```py
    def cast_to_pointset(self, deep: bool = False, pass_cell_data: bool = False) -> 'pyvista.PointSet':
        """Extract the nodes of this dataset as a :class:`pyvista.PointSet`.

        Parameters
        ----------
        deep : bool, optional
[...]

PR #3242, opened 2022-08-31
Move `cast_to_pointset` to `_PointSet`

MatthewFlamm:
### Overview

 

Not all `DataSet`s can be converted directly to `PointSet`, at least using the current method.  In particular `UniformGrid` and `RectilinearGrid` are either missing the `GetPoints` method entirely, or have a different form of the method.

### Details

It may be possible to implement this method in a different way for `UniformGrid` and `RectilinearGrid`, but I'm not proposing to do that in this PR.  It is possible to do `grid.cast_to_unstructured_grid().cast_to_pointset()` in the interim, albeit with probably more memory/cpu usage then needed.

If this PR is merged, I will update #3164 to be about how the method is missing for others who may want to investigate implementing that method.

### DIFF
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -2144,37 +2144,6 @@ class DataSet(DataSetFilters, DataObject):
         alg.Update()
         return _get_output(alg)
 
-    def cast_to_pointset(self, deep: bool = False) -> 'pyvista.PointSet':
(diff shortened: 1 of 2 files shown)


banesullivan (changes requested):
I'd prefer to move this to the DataSet class. See explanation in #3164

MatthewFlamm:
Based on the comment of #3164, I will close this as there is likely a more universal method for casting to pointset.

PR #3243, opened 2022-08-31
Make `cast_to_pointset` universal and add `cast_to_poly_points`

### LINKED ISSUE #3164: UniformGrid cannot be cast to pointset

banesullivan:
Resolves #3164 by adding a universal implementation of `cast_to_pointset`

Further adds `cast_to_poly_points` for use cases where we may want to extract the nodes of a mesh as `PolyData` and preserve CELL data as CELL data.

cc @MatthewFlamm

### DIFF
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -2142,8 +2142,10 @@ class DataSet(DataSetFilters, DataObject):
         alg.Update()
         return _get_output(alg)
 
-    def cast_to_pointset(self, deep: bool = False) -> 'pyvista.PointSet':
-        """Get a new representation of this object as a :class:`pyvista.PointSet`.
+    def cast_to_pointset(
+        self, deep: bool = False, pass_cell_data: bool = False
+    ) -> 'pyvista.PointSet':
+        """Extract the nodes of this dataset as a :class:`pyvista.PointSet`.
 
         Parameters
         ----------
@@ -2151,6 +2153,10 @@ class DataSet(DataSetFilters, DataObject):
             When ``True`` makes a full copy of the object.  When ``False``,
             performs a shallow copy where the points and data arrays are
             references to the original object.
+
+        pass_cell_data : bool, optional
+            Run the ``cell_data_to_point_data`` filter and pass cell data
+            fields to the new pointset
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


MatthewFlamm:
I have a few thoughts for discussion relating to data copying.

1. When using `pass_cell_data=True`, it may be surprising that the resulting cell data will be a copy of the data while the point data will not be.  But I think the mention of using the `cell_data_to_point_data` filter should be enough for users to figure this out.
2. The points are copied when I try using `sphere = pv.Sphere(); sphere.cast_to_pointset()`, but the point data is not.  That is I can change the points on one mesh without affecting the other.  But when I change the point data on one mesh, it would be updated on the others.  This is unintuitive to me.  I'm not sure if transformations using vector data will be impacted here too.

akaszynski:
> 2\. but the point data is not.  That is I can change the points on one mesh without affecting the other.  But when I change the point data on one mesh, it would be updated on the others.  This is unintuitive to me.  I'm not sure if transformations using vector data will be impacted here too.

IMO either both should be shallow or deep copied. However, this does appear to match VTK's approach when copying datasets.

I'd prefer implementing (2), but if not it's also fine since that's the default for VTK.

MatthewFlamm:
> I'd prefer implementing (2), but if not it's also fine since that's the default for VTK.

What do you mean by this?  Do you mean that you prefer having all data be shallow copied (except for the cell data option?)

If so, this will be dataset dependent.  For example, `UniformGrid` has read only points, so explicitly we cannot shallow copy them.  If we can shallow copy the points for `_PointSet` objects, then it still makes sense to me with some documentation around this.

If we cannot shallow copy the points for `_PointSet` objects, then I think we should either deep copy everything or provide an optional `deep` argument. I'm leaning towards default `deep=True` in this case, so the user can opt-in to the inconsistent usage.

banesullivan:
I updated both of these methods to strictly deep copy all data. This PR is ready for review/merge

pyvista/core/dataset.py
```diff
@@ -2162,35 +2162,98 @@ def cast_to_unstructured_grid(self) -> 'pyvista.UnstructuredGrid':
...
-            return pset.copy(deep=True)
+        pset.points = self.points.copy()
+        if pass_cell_data:
+            self = self.cell_data_to_point_data()
+        pset.GetPointData().DeepCopy(self.GetPointData())
+        pset.active_scalars_name = self.active_scalars_name
+        return pset
+
+    def cast_to_poly_points(self, pass_cell_data: bool = False) -> 'pyvista.PolyData':
+        """Extract the nodes of this dataset as a :class:`pyvista.PolyData`.
```
akaszynski, line 2202: Would prefer moving away from "nodes" to points. I think this comes from FEA and we should generalize PyVista.

```suggestion
        """Extract the points of this dataset and return a :class:`pyvista.PolyData`.
```

pyvista/core/dataset.py
```diff
@@ -2162,35 +2162,98 @@ def cast_to_unstructured_grid(self) -> 'pyvista.UnstructuredGrid':
         alg.Update()
         return _get_output(alg)
 
-    def cast_to_pointset(self, deep: bool = False) -> 'pyvista.PointSet':
-        """Get a new representation of this object as a :class:`pyvista.PointSet`.
+    def cast_to_pointset(self, pass_cell_data: bool = False) -> 'pyvista.PointSet':
+        """Extract the nodes of this dataset as a :class:`pyvista.PointSet`.
```
akaszynski, line 2166:
```suggestion
        """Extract the points of this dataset and return a :class:`pyvista.PointSet`.
```

akaszynski (approved):
This is good to go. Thanks!

PR #3216, opened 2022-08-26
Render window refactor

akaszynski:
Redo of https://github.com/pyvista/pyvista/pull/2964

I'm not entirely sure if the failures on `main` were due to #2964, but I want to be sure since this is a substantial change. I'll run the CI a bunch of times and check for stability.

PR #3245, opened 2022-08-31
Fix Plotter garbage collection

### LINKED ISSUE #2252: Memory Leak Using Plotter and Screenshot
Memory Leak Using Plotter and Screenshot
-----
It seems that `Plotter.screenshot()` produces a memory leak. I am currently running an iterative program implementing the observer pattern, where after each iteration (i.e. when the observer is notified), the observer plots and saves (as a .png) OpenFOAM data.

What I notice in my Activity Monitor is that after each time the observer is called and executed (i.e. PlotVTK.update() is run), my python memory increases by around  ~70MB, which becomes a problem when the observer is called around 300 times, when memory consumption has increased to many many GBs.

I have noticed that when I comment out `p.screenshot(f'plot_{iteration}', window_size=[1000,1000], return_img = False)
`, my memory increase after each observer call is much smaller, i.e. 5MB. Although, when I don't call this observer at all, my memory increase after each observer call in <1MB.

Any help would be appreciated.

**My code**

```py
import Observer
import pyvista as pv

class PlotVTK(Observer):
    
    def update(self, subject) -> None:
        
        # Get current iteration
        iteration = subject.iteration
        
        # Open openfoam file
        reader = pv.OpenFOAMReader(f'test_{iteration}.foam')

        # Obtain data for latest snapshot
        reader.set_active_time_value(reader.time_values[-1])

        # Read data into mesh
        mesh = reader.read()

        # Extract internal mesh data
        internalMesh = mesh["internalMesh"].slice_orthogonal()

        # Create plotting environment
        p = pv.Plotter(off_screen = True)

        # Plot velocity diagram
        p.add_mesh(internalMesh, scalars = 'U')

        # Set viewing angle
        p.view_xy()
        p.camera_position = ((0, 0, 3), (0, 0, 0), (0, 0, 0))
        
        # Take screenshot of image and save in appropriate directory
        p.screenshot(f'plot_{iteration}', window_size=[1000,1000], return_img = False)

        # Clear plot data (i.e. release memory)
        p.clear()
```
-----

**System Information:**
```txt

  Date: Mon Feb 21 16:38:57 2022 GMT

                OS : Darwin
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 16.0 GiB
       Environment : IPython
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M1
       GPU Version : 4.1 Metal - 76.3

  Python 3.9.7 (default, Sep 16 2021, 08:50:36)  [Clang 10.0.0 ]

           pyvista : 0.33.2
               vtk : 9.1.0
             numpy : 1.21.2
           imageio : 2.16.0
           appdirs : 1.4.4
            scooby : 0.5.11
        matplotlib : 3.5.0
           IPython : 7.29.0
        ipyvtklink : 0.2.2
             scipy : 1.7.3
              tqdm : 4.62.3

```

### LINKED ISSUE #2577: Plotter memory leak due to not clearing up

akaszynski:
This is a less ambitious implementation of #3216 that attempts to tackle just the garbage collection of the `Plotter` class to resolve #2577.

Next step is to try to determine what's causing #2252.


pyvista/plotting/render_window_interactor.py
```diff
@@ -24,14 +26,17 @@ class RenderWindowInteractor:
...
         if interactor is None:
             interactor = _vtk.vtkRenderWindowInteractor()
+
         self.interactor = interactor
         self.interactor.SetDesiredUpdateRate(desired_update_rate)
         if not light_follow_camera:
             self.interactor.LightFollowCameraOff()
 
-        # Map of observers to events
+        # # Map of observers to events
```
akaszynski, line 39:
```suggestion
        # Map of observers to events
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -42,10 +47,9 @@ def __init__(self, plotter, desired_update_rate=30, light_follow_camera=True, in
         self._MAX_CLICK_DELAY = 0.8  # seconds
         self._MAX_CLICK_DELTA = 40  # squared => ~6 pixels
 
-        # Set default style
+        # # Set default style
```
akaszynski, line 50:
```suggestion
        # Set default style
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -24,14 +26,17 @@ class RenderWindowInteractor:
 
     def __init__(self, plotter, desired_update_rate=30, light_follow_camera=True, interactor=None):
         """Initialize."""
+        self.__plotter = weakref.ref(plotter)
+
         if interactor is None:
             interactor = _vtk.vtkRenderWindowInteractor()
+
```
akaszynski, line 33:
```suggestion
```

akaszynski:
@banesullivan, this is good for review. Five runs, no weird errors.

pyvista/plotting/plotting.py
```diff
@@ -259,7 +258,7 @@ def __init__(
         elif lighting_normalized != 'none':
             raise ValueError(f'Invalid lighting option "{lighting}".')
 
-        # Add self to open plotters
+        # Track all active plotters
```
akaszynski, line 261:
```suggestion
        # Track all active plotters. This has the side effect of ensuring that plotters are not
        # collected until `close()`. See https://github.com/pyvista/pyvista/pull/3216
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -783,6 +791,9 @@ def initialize(self):
 
     def set_render_window(self, ren_win):
         """Set the render window."""
+        # edge_case: accept pyvista.RenderWindow
+        if hasattr(ren_win, '_ren_win'):
+            ren_win = ren_win._ren_win
```
akaszynski, line 796:
```suggestion
```

pyvista/ext/plot_directive.py
```diff
@@ -373,7 +373,6 @@ def render_figures(
 
             images = []
             figures = pyvista.plotting._ALL_PLOTTERS
```
akaszynski, line 375:
```suggestion
            figures = pyvista.plotting._ALL_PLOTTERS

```

pyvista/plotting/charts.py
```diff
@@ -1364,6 +1364,7 @@ def show(
         window_size=None,
         notebook=None,
         background='w',
+        **kwargs,
```
akaszynski, line 1367:
```suggestion
```

pyvista/plotting/charts.py
```diff
@@ -1404,6 +1405,9 @@ def show(
             ``color='white'``, ``color='w'``, ``color=[1.0, 1.0, 1.0]``, or
             ``color='#FFFFFF'``.  Defaults to ``'w'``.
 
+        **kwargs : dict, optional
+            Optional developer keyword arguments.
+
```
akaszynski, line 1410:
```suggestion
```

pyvista/plotting/charts.py
```diff
@@ -1430,6 +1434,7 @@ def show(
         return pl.show(
             screenshot=screenshot,
             full_screen=full_screen,
+            **kwargs,
```
akaszynski, line 1437:
```suggestion
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -24,6 +26,8 @@ class RenderWindowInteractor:
 
     def __init__(self, plotter, desired_update_rate=30, light_follow_camera=True, interactor=None):
         """Initialize."""
+        self.__plotter = weakref.ref(plotter)
+
```
akaszynski, line 30:
```suggestion
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -45,7 +49,6 @@ def __init__(self, plotter, desired_update_rate=30, light_follow_camera=True, in
         # Set default style
         self._style = 'RubberBandPick'
         self._style_class = None
```
akaszynski, line 51:
```suggestion
        self._style_class = None
        self.__plotter = weakref.ref(plotter)
```

larsoner:
Okay the MNE failures here are reasonable in the sense that we (MNE) modify `_ALL_PLOTTERS` directly, which is a private attribute. IIRC we did this stuff at one point because garbage collection was not always working for us. So we probably need to change how we use `_ALL_PLOTTERS` in concert with this PR.

If the migration is just "don't mess with _ALL_PLOTTERS" then I can open make branch in MNE-Python that does this (or you can) and it can be tested here.

akaszynski:
@larsoner, since this PR effectively fixes the memory leaks with the plotter, I'd add in a backwards compatible try/except or maybe a `pop` so it works for older versions of pyvista.

larsoner (dismissed):
Should come back green with the latest MNE-Python commit -- thanks for working with us on this!

akaszynski:
> Should come back green with the latest MNE-Python commit -- thanks for working with us on this!

More than glad to. I wish we had fixed this a while ago, but late is better than never.

PR #3246, opened 2022-08-31
add more tinypages tests

akaszynski:
Quick PR to improve our tinypages tests to account for edge cases.


akaszynski:
Merging shortly unless there are objections.

PR #3248, opened 2022-08-31
Add additional plotting tests

akaszynski:
Add a handful of plotting tests and improves or splits several of the volume plotting image regression tests.


tests/plotting/test_plotting.py
```diff
@@ -1646,32 +1646,43 @@ def test_image_properties():
...
     data['d'].rename_array('Spatial Point Data', 'd')
-    data.plot(volume=True, multi_colors=True)
+    data.plot(volume=True, multi_colors=True, before_close_callback=verify_cache_image)
+
 
-    # Check that NumPy arrays work
-    arr = vol["Spatial Point Data"].reshape(vol.dimensions)
+def test_array_volume_rendering(uniform):
+    arr = uniform["Spatial Point Data"].reshape(uniform.dimensions)
     pyvista.plot(arr, volume=True, opacity='linear')
```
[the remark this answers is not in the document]
akaszynski: Yes, thanks for that.

pyvista/plotting/charts.py
```diff
@@ -1364,6 +1364,7 @@ def show(
         window_size=None,
         notebook=None,
         background='w',
+        **kwargs,
```
banesullivan, line 1367: Oof, I am really not a fan of unchecked `**kwargs`, these can really come back to bite us. If they are truly for developer use, then could we make this `developer_kwargs=None` instead?

tests/plotting/test_plotting.py
```diff
@@ -1646,33 +1646,44 @@ def test_image_properties():
     p.close()
 
 
-def test_volume_rendering():
-    # Really just making sure no errors are thrown
-    vol = examples.load_uniform()
-    vol.plot(volume=True, opacity='linear')
+def test_volume_rendering_from_helper(uniform):
+    uniform.plot(volume=True, opacity='linear', before_close_callback=verify_cache_image)
```
banesullivan, line 1650: We should really just make a pytest fixture context that sets `verify_cache_image` as the default `before_close_callback`... to do in a later PR though

banesullivan (changes requested):

pyvista/plotting/charts.py
akaszynski: Resolved in 41ae933.

banesullivan (dismissed):

PR #3249, opened 2022-09-01
GC: use deep_clean on close

akaszynski:
Given the complexity of this problem, this PR proposes simply using `deep_clean` on close to clean up resources. The strange VM CI only bugs are challenging to track down and this is an easy fix.


akaszynski:
Finally got it working in #3245

PR #3254, opened 2022-09-01
BUG: Provide default for renderer.enable_anti_aliasing

larsoner:
Maintain backward compat by adding the old default


akaszynski (approved):

akaszynski:
Merging immediately as we're breaking builds and this is a trivial fix.

PR #3255, opened 2022-09-01
CI: Workflow to test downstream packages starting with MNE

larsoner:
Let's see how well this works!


banesullivan:
@larsoner, would you be up for working with me to create a "pyvista-integration-tests" repository with manually triggered GitHub workflows to do this sort of testing?

That way, we could input a branch of PyVista and a branch of a TPL like MNE to test together

banesullivan:
Considering MNE is one of the highest used downstream dependencies and that MNE's success has directly trickled down to become some of PyVista's success, I believe we have some responsibility to do integration tests so that we don't leave you all high and dry on a release

larsoner:
> would you be up for working with me to create a "pyvista-integration-tests" repository with manually triggered GitHub workflows to do this sort of testing?

Sure, we could do that instead if you want. I'm not sure if it brings a lot of benefits compared to just having this CI run, though.

larsoner:
> I'd like to keep the CI in PyVista as streamlined as possible with these sort of integration tests being somewhere else.

FWIW I don't think this would have avoided the issue that we just hit in MNE at least. If the CIs here don't say immediately "hey this breaks stuff!" when people are actively working on things before merge, then what happens instead? The problematic code gets merged to `main`, and tests will break in this other joint-testing repo and in MNE simultaneously, *then* someone needs to come back to this repo and fix the issue, which hopefully happens quickly but there's no guarantee of this. And when they work on a fix, they have no feedback about whether or not it works. (I'm thinking of the recent huge plotter / deep_clean refactor -- the MNE tests might be really useful there!)

So to me the proposed separate-workflow won't avoid the problem of breaking downstream repos as effectively. But I agree such a joint repo is better than nothing, so if the extra 9-minute CI run here is too much (or not sustainable/generalizable enough) I can live with the joint-repo solution.

akaszynski:
> > would you be up for working with me to create a "pyvista-integration-tests" repository with manually triggered GitHub workflows to do this sort of testing?
> 
> Sure, we could do that instead if you want. I'm not sure if it brings a lot of benefits compared to just having this CI run, though.

As long as this isn't the longest running workflow (which it's not) then I'd like this run for each PR.

larsoner:
It looks like the run time was 9 min, which should go down to ~7 once the testing data caching kicks in. Let's see...

akaszynski:
Longest run is 17-20 minutes on our CI. It's one more parallel build, but if it means we catch more things upfront, it's totally worth it.

larsoner:
It was now the first one done, in 6 min :)

banesullivan:
Fair enough to keep here then! Though, I would like this to be a separate workflow, called "Integration Tests" with a job for each downstream package instead of bundling these in the "Unit Testing and Deployment" workflow

Filename: `pyvista/.github/workflows/integration-tests.yml`

Template:

```yml
name: Integration Tests

on:
  pull_request:
  workflow_dispatch:
  schedule:
    - cron: "0 4 * * *"
  push:
    tags:
      - "*"
    branches:
      - main

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

env:
  ALLOW_PLOTTING: true
  SHELLOPTS: 'errexit:pipefail'

jobs:
  mne:
```

MatthewFlamm:
I agree that if we don't have the test run for every PR, the PR author will not be aware of breakages downstream. This has obvious negative consequences downstream and some here too.

But, this proposal also has downsides here.  Bleeding edge downstream packages could be breaking due to unrelated packaging issues that are not present in pyvista.  This will stall progress here that is unrelated to the codebase.

I propose that if this goes forward this check should be optional for merging, unless the failed job is obviously related to the PR.

adeak:
> But, this proposal also has downsides here. Bleeding edge downstream packages could be breaking due to unrelated packaging issues that are not present in pyvista.

What do you think about testing against the latest release of downstream? The upside would be stability across a given downstream version, the downside is false negatives for our changes that break bleeding edge changes downstream.

akaszynski:
> What do you think about testing against the latest release of downstream?

The advantage of the current approach is there's excellent overlap with mne. Their current testing is their latest with our latest. We can test with our latest against their release and then get better version coverage while potentially avoiding issues with things breaking upstream.

> Agreed - this will not be a "Required" check for PRs but it should still run no matter what

Also in agreement there. In fact, I'd like to have this merged soon and make sure #3245 doesn't break anything. It will make me more confident about large changes in the codebase.

akaszynski:
> Their current testing is their latest with our latest. We can test with our latest against their release and then get better version coverage while potentially avoiding issues with things breaking upstream.

Actually, that only works if we checkout a release vs use `main`, which means we'd have to keep modifying the tag. We might be able to scrape something off of PyPI, but for the sake of simplicity, let's keep to main and if it's unstable we'll come up with a better solution.

I'd like to merge this in <24H to test with a few other PRs. Any objections?

banesullivan:
@akaszynski, I'd like this to be a separate workflow before merging. I don't think this PR is enabled for me to edit, but here is the workflow file:

`pyvista/.github/workflows/integration-tests.yml`:

```yml
name: Integration Tests

on:
  pull_request:
  workflow_dispatch:
  schedule:
    - cron: "0 4 * * *"
  push:
    tags:
      - "*"
    branches:
      - main

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

env:
  ALLOW_PLOTTING: true
  SHELLOPTS: 'errexit:pipefail'

jobs:
  # For now this is just MNE-Python, but others could be added
  mne:
    name: MNE-Python
    runs-on: ubuntu-22.04
    env:
      DISPLAY: ':99.0'
      MNE_LOGGING_LEVEL: 'info'
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-python@v4
        with:
        python-version: '3.10'
      - run: git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
      - run: ./mne-python/tools/setup_xvfb.sh
      - run: ./tools/github_actions_dependencies.sh
        working-directory: mne-python
      - run: pip install -ve .  # pyvista
      - run: pip install -ve . && mne sys_info  # mne
        working-directory: mne-python
      - run: ./tools/get_testing_version.sh
        working-directory: mne-python
      - uses: actions/cache@v3
        with:
          key: ${{ env.TESTING_VERSION }}
          path: ~/mne_data
        working-directory: mne-python
      - run: pytest --cov=pyvista --cov-report=xml mne/viz/_brain mne/viz/tests/test_3d.py mne/viz/backends
        working-directory: mne-python
      # In theory this should be fine, too (codecov should combine all reports)
      # if we copied the cov report up one directory
      #- uses: codecov/codecov-action@v3
      #  name: 'Upload coverage to CodeCov'
      #  directory: ./mne-python/

```

akaszynski:
@larsoner, please enable edits; we just want to move the workflow into a separate file for organization.

banesullivan:
Nevermind, I just don't know how to use VSCode in the browser. Made the change

banesullivan:
@akaszynski, before merging, I'd like to audit this a bit to make sure what we expect is getting installed. For example, the `./tools/github_actions_dependencies.sh` step installs PyVista from main:

```
Collecting https://github.com/pyvista/pyvista/zipball/main
  Downloading https://github.com/pyvista/pyvista/zipball/main
```

akaszynski:
> @akaszynski, before merging, I'd like to audit this a bit to make sure what we expect is getting installed. For example, the `./tools/github_actions_dependencies.sh` step installs PyVista from main:
> 
> ```
> Collecting https://github.com/pyvista/pyvista/zipball/main
>   Downloading https://github.com/pyvista/pyvista/zipball/main
> ```

Leaving the merge to you. It's got my approval when you're ready, ping me if you need it GH official.

larsoner:
> For example, the ./tools/github_actions_dependencies.sh step installs PyVista from main:

Yes but then it's immediately installed over in the next line with `- run: pip install -ve .  # pyvista`

larsoner:
> @larsoner, please enable edits; we just want to move the workflow into a separate file for organization.

I already have `Allow edits and access to secrets by maintainers` checked, are you not in the maintainers category for the org?

larsoner:
> Nevermind, I just don't know how to use VSCode in the browser. Made the change

Sorry I'm reading through too slowly today :)

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
+      - run: ./mne-python/tools/setup_xvfb.sh
+      - name: Install PyVista
+        run: pip install -ve .  # pyvista
+      - name: Install MNE dependencies
+        run: ./tools/github_actions_dependencies.sh
```
larsoner, line 40: I had these in the opposite order before intentionally, you should do this first then the `Install PyVista` step

banesullivan: Hm. Okay, I will switch back

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+        working-directory: mne-python
+      - name: Install MNE
+        run: pip install -ve .  # mne
+        working-directory: mne-python
+      - name: MNE Info
+        run: mne sys_info
```
larsoner, line 46: If you just do a `-p` here, it will show the installation dir for `pyvista`, no need fro the `pip show pyvista` below
```suggestion
        run: mne sys_info -p
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+        working-directory: mne-python
+      - name: PyVista Report
+        run: |
+          python -c "import pyvista;print(pyvista.Report())"
+          pip show pyvista
```
larsoner, line 51:
```suggestion
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+    env:
+      DISPLAY: ':99.0'
+      MNE_LOGGING_LEVEL: 'info'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
+      - run: ./mne-python/tools/setup_xvfb.sh
```
banesullivan, line 36: @larsoner, any reason you're using this and not: https://github.com/pyvista/setup-headless-display-action

larsoner: Our `setup_xvfb.sh` is a little bit of a misnomer because it actually also takes care of any `xcb` dependencies that the various PyQt / PySides might need to work properly. If `setup-headless-display-action` also does this, then yes this can be replaced

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Install MNE dependencies
+        run: ./tools/github_actions_dependencies.sh
```
banesullivan, line 38: @larsoner, this scripts concerns me a bit. I'm noticing that it installs a lot of things at their latest development version. I imagine that this might lead to a lot of false flags here if something fails due to another dependency and not PyVista.

larsoner: It does install a bunch of other stuff. But:

1. It actually has been very stable for the last couple of years we've done it this way -- other than PyVista issues, most failures are with mpl deprecations, which we won't hit with the subset of tests in the `pytest` line here
2. There are some dependencies we need to get all visualizations, such as `nibabel` for MRI support, etc.

I'll trim it down to the minimal set of things we need instead of installing everything.

I'll also remove the failing `pyvista.Report` line for now. We can add it back later in a separate PR that fixes scooby or whatever needs to be done there...

banesullivan: I think this just showed up here: https://github.com/pyvista/pyvista/runs/8142770010?check_suite_focus=true#step:10:1

banesullivan: Okay, thanks for clarifying. I think this is close to done now

banesullivan:
Algright! I think this is good to go now. Once the CI comes back green, ready to merge

larsoner:
Okay I think the tests are complete now, including notebook integration

akaszynski (approved):

banesullivan:
For posterity, I want to link to a comment where I justify the presence of integration tests in our CI https://github.com/pyvista/pyvista/pull/4486#discussion_r1214641587

(linking here because it is an inline comment that will likely be lost in the ether)

PR #3256, opened 2022-09-01
Add vtkRectilinearGridGeometryFilter

banesullivan:
Adds `vtkRectilinearGridGeometryFilter` under `extract_geometry_extent` for https://github.com/pyvista/pyvista/discussions/3228

To do:

- [ ] Are we sure this filter should only return points? All other geometry filters return actual geometry
- [ ] Add tests
- [ ] Figure out why this segfaulted for me on a bigger mesh

### DIFF
--- a/pyvista/_vtk.py
+++ b/pyvista/_vtk.py
@@ -243,6 +243,7 @@ if VTK9:
         vtkCompositeDataGeometryFilter,
         vtkDataSetSurfaceFilter,
         vtkGeometryFilter,
+        vtkRectilinearGridGeometryFilter,
         vtkStructuredGridGeometryFilter,
     )
     from vtkmodules.vtkFiltersHybrid import vtkPolyDataSilhouette
--- a/pyvista/core/filters/rectilinear_grid.py
+++ b/pyvista/core/filters/rectilinear_grid.py
@@ -1,7 +1,7 @@
 """Filters module with the class to manage filters/algorithms for rectilinear grid datasets."""
 
 import collections
-from typing import Sequence, Union
+from typing import List, Optional, Sequence, Union
 
 import numpy as np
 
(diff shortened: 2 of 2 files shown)


banesullivan:
I'm inclined not to add this filter as it is not intuitive (likely has a bug) and hasn't been updated since the late 90s

```
lg.SetExtent(0, 0, 0, 0, 0, maxi) # -> Line
alg.SetExtent(0, 0, 0, maxi, 0, maxi) # -> Plane
alg.SetExtent(0, maxi, 0, maxi, 0, maxi) # -> Points
```

the `vtkGeometryFilter` is more modern has a `SetExtent` that should work

banesullivan:
@pyvista/developers, this is ready for review if anyone has a moment

pyvista/core/filters/data_set.py
```diff
@@ -1258,7 +1258,7 @@ def outline_corners(self, factor=0.2, progress_bar=False):
         _update_alg(alg, progress_bar, 'Producing an Outline of the Corners')
         return wrap(alg.GetOutputDataObject(0))
 
-    def extract_geometry(self, progress_bar=False):
+    def extract_geometry(self, extent: Optional[List[int]] = None, progress_bar=False):
```
tkoyama010, line 1261:
```suggestion
    def extract_geometry(self, extent: Optional[Sequence[float]] = None, progress_bar=False):
```

pyvista/core/filters/data_set.py
```diff
@@ -1,6 +1,6 @@
 """Filters module with a class of common filters that can be applied to any vtkDataSet."""
 import collections.abc
-from typing import Union
+from typing import List, Optional, Union
```
tkoyama010, line 3:
```suggestion
from typing import Sequence, Optional, Union
```

tkoyama010 (changes requested):
Thanks. A small suggestion about type hint.

tkoyama010 (approved):
LGTM

PR #3259, opened 2022-09-02
Add linear cells example

akaszynski:
There is an excellent example of [Linear Cells](https://kitware.github.io/vtk-examples/site/Cxx/GeometricObjects/LinearCellDemo/) by kitware that I've ported to PyVista, which was made much easier by #3115.

Depends on #3115.

tmp

- [x] #3266


pyvista/cells/cells.py
```diff
@@ -0,0 +1,823 @@
...
+    >>> pv.cells.plot_cell(grid)
+
+    """
+    pl = pyvista.Plotter()
+    pl.add_mesh(grid, opacity=0.5)
+    edges = grid.extract_all_edges()
+    if edges.n_cells:
+        pl.add_mesh(grid.extract_all_edges(), line_width=5, color='k', render_lines_as_tubes=True)
+    pl.add_points(grid, render_points_as_spheres=True, point_size=30, color='r')
+    pl.add_point_labels(
```
tkoyama010, line 35: Do you know why these line has `Codecov` warning?

tkoyama010 (commented):
Thanks for improving `cells` module. I'm impressed with the example.

pyvista/cells/cells.py
tkoyama010: It seems that it is tested...

tkoyama010:
I will merge the main branch to try the change of https://github.com/pyvista/pyvista/pull/3266.

pyvista/__init__.py
```diff
@@ -22,6 +22,7 @@
 from pyvista.utilities.sphinx_gallery import _get_sg_image_scraper
 
 from pyvista.utilities.wrappers import _wrappers
+from pyvista import cells
```
tkoyama010, line 25:
```suggestion
from pyvista.cells import *
```

pyvista/cells/cells.py
```diff
@@ -0,0 +1,823 @@
+"""Contains a variety of cells to serve as examples."""
```
adeak, line 1: If these are examples, do we really need to pull them into the main pyvista namespace automatically?

pyvista/cells/cells.py
```diff
@@ -0,0 +1,823 @@
...
+    """Create a :class:`pyvista.UnstructuredGrid` containing a single poly line.
+
+    This represents a set of line segments as a single cell.
+
+    This cell corresponds to the :attr:`pyvista.CellType.POLY_LINE` cell type.
+
+    Returns
+    -------
+    pyvista.UnstructuredGrid
+        UnstructuredGrid containing a single poly line.
```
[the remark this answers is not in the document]
banesullivan: I prefer "polyline" as well, but it's worth noting that the VTK class in "vtkPolyLine" with the camel case implying usage should be "poly line".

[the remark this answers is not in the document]
banesullivan: This is similar to "dataset" vs "data set" with `vtkDataSet`. English is a blast

pyvista/cells/cells.py
banesullivan: I agree, I don't like having example code in the main `pyvista` namespace -- that is what the `examples` subpackage should be used for.

However, this is in line with some of the other geometry methods we have like `Cube()` and `Sphere()` or more sophisticated ones like `Wavelet()`.

I wouldn't be opposed to moving all of these and other's like `Cube()` to the `examples` subpackage, but it may create a migration nightmare

pyvista/cells/cells.py
```diff
@@ -0,0 +1,823 @@
+"""Contains a variety of cells to serve as examples."""
+import pyvista
+from pyvista import CellType, UnstructuredGrid
+
+
+def plot_cell(grid, cpos=None, **kwargs):
```
banesullivan, line 6: I feel this method at least better belongs in `examples` somewhere

pyvista/cells/cells.py
adeak: I don't have a _strong_ opinion about this, but in my mind Cube/Sphere etc. are different from these new cells.

1. the former are well-defined geometric objects (see also `pyvista/utilities/geometric_objects.py`)
2. the former have a bunch of parameters that can be used to customize them, which means they can be useful in a bunch of real use cases other than examples or debugging
3. the latter smell very "internals" to me: special cases of unstructured grid cells, something very "VTK". Very abstract. Somewhat arbitrary.

Not sure if the above makes sense to others, but I'd be perfectly fine with keeping the cell example stuff in examples (or at least unimported into the main namespace) and still keeping the geometric objects.

adeak:
I also occasionally look up the linear/nonlinear cells, in my case from https://kitware.github.io/vtk-examples/site/VTKFileFormats/#legacy-file-examples. It would be nice to be able to look these up in our own docs. One thing though: do you think we could squeeze in the cells' int value in the "gallery", similarly to how they are shown in VTK? Considering `.celltypes` and its occasional appearance in our examples, it might be helpful to be able to point at that too in the figures.

This is still just an example so it's not an issue if it doesn't fit or you just don't find that pretty enough.

akaszynski:
> I also occasionally look up the linear/nonlinear cells, in my case from https://kitware.github.io/vtk-examples/site/VTKFileFormats/#legacy-file-examples. It would be nice to be able to look these up in our own docs. One thing though: do you think we could squeeze in the cells' int value in the "gallery", similarly to how they are shown in VTK? Considering `.celltypes` and its occasional appearance in our examples, it might be helpful to be able to point at that too in the figures.
> 
> This is still just an example so it's not an issue if it doesn't fit or you just don't find that pretty enough.

Done in 088c3247c85cea79d4ce7d8e0f016bfb5ea9ef76.

tmp

examples/00-load/linear-cells.py
```diff
@@ -0,0 +1,200 @@
...
+# ~~~~~~~~~~
+# PyVista contains the :class:`pv.CellType` enumerator, which contains all the
+# available VTK cell types mapped to a Python enumerator. These cell types are
+# used when creating cells and also can be used when checking the
+# :attr:`celltypes <pyvista.UnstructuredGrid.celltypes>` attribute. For example
+# ``combined.celltypes`` contains both the ``pv.CellType.HEXAHEDRON`` and
+# ``pv.CellType.HEXAGONAL_PRISM`` cell types.
+
+print(pv.CellType.HEXAHEDRON, pv.CellType.HEXAGONAL_PRISM)
+(pv.CellType.HEXAHEDRON, pv.CellType.HEXAGONAL_PRISM) == combined.celltypes
```
tkoyama010, line 95: The statement seems to have no effect.
```suggestion
assert (pv.CellType.HEXAHEDRON, pv.CellType.HEXAGONAL_PRISM) == combined.celltypes
```

adeak: Doesn't it print `True` via `repr`?

akaszynski: Yep.  It's only just supported to show it's the same.

I think we should keep `assert` statements to our testing.

akaszynski:
Reminder to self to follow-up with the quadratic cells by following [legacy-file-examples](https://kitware.github.io/vtk-examples/site/VTKFileFormats/#legacy-file-examples).

examples/00-load/linear-cells.py
tkoyama010: Sorry. This is an EXAMPLE, not a test. I misunderstood.

akaszynski:
Calling this good to go unless there are any objections.

pyvista/cells/cells.py
banesullivan: I'd like to push for these to be in the `examples` namespace -- primarily so that they do not further crowd the `pyvista` root namespace with is already pretty crowded

banesullivan:
> Calling this good to go unless there are any objections.

My only objects are wanting to move some things into the `examples` module but don't sit on this because of me.

akaszynski:
@banesullivan, I've moved `cells.py` to the examples module. I agree that it shouldn't be importable at the `pyvista` level.

PR #3266, opened 2022-09-04
Prevent warning in polystrips example coming from extrude

adeak:
Full doc builds emit a warning:
```
/home/adeak/pyvista/pyvista/core/filters/poly_data.py:2778: PyVistaFutureWarning: The default value of the ``capping`` keyword argument will change in a future version to ``True`` to match the behavior of VTK. We recommend passing the keyword explicitly to prevent future surprises.
```

This is caused by the doctest of the new poly `strips` not specifying the `capping` kwarg. I decided to keep the old behaviour which will mean that use of the kwarg is still relevant when we flip the switch on the default value of `capping` (from `False` to `True`).

### DIFF
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -802,7 +802,7 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
         --------
         >>> import pyvista as pv
         >>> polygon = pv.Rectangle()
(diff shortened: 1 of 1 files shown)


tkoyama010:
LGTM. Thanks. I think #3259 and (other Pull Request maybe) need this.

tkoyama010 (approved):

PR #3267, opened 2022-09-04
Add dev Gitpod support 

### LINKED ISSUE #2959: Gitpod dev environment

A gitpod dev environment would be very useful to get a quick environment setup.  This would be a very useful thing for new contributors as well.

A start was made in #2933.

### Code of Conduct

MatthewFlamm:
### Overview

 

This is another attempt at #2933.  This PR adds initial gitpod support including prebuilt environment including:
- interactive plotting with browser based vnc
- testing
- documentation
  - prebuilt image
  - live viewer
- preinstalling pre-commit hook environments

### Details

Gitpod handles image building and also prebuilding workspaces.  Pre-builds are turned off for pull requests from fork for now.  I'm not certain on the security implications here.

~A few tests fail when run using this environment.  Future PRs could be used to address this.~  Tests failed when using the OSMesa and the accompanying wheel, but not when using the vnc image.

Closes #2959.

TODO:
- [x] feedback on virtual environment approach
- [x] add documentation

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -14,10 +14,6 @@ vscode:
 
 tasks:
   - init: |
-      # this needs to be a proj var
-      export PRE_COMMIT_HOME=/workspace/.precommit
-      [ ! -d venv ] && python -m venv venv
-      source venv/bin/activate
       pip install --no-cache-dir https://github.com/pyvista/pyvista-wheels/raw/main/vtk-osmesa-9.1.0-cp38-cp38-linux_x86_64.whl
       pip install --no-cache-dir -r requirements_test.txt
       pip install --no-cache-dir -r requirements_docs.txt
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,44 @@
+image:
+  file: docker/dev-gitpod.Dockerfile
+
+# --------------------------------------------------------
+# exposing ports for liveserve
+ports:
+  - port: 5500
+    onOpen: notify
+
+vscode:
+  extensions:
+    - ms-python.python
+    - ritwickdey.liveserver
+
+tasks:
+  - init: |
+      # this needs to be a proj var
+      export PRE_COMMIT_HOME=/workspace/.precommit
+      [ ! -d venv ] && python -m venv venv
+      source venv/bin/activate
+      pip install --no-cache-dir https://github.com/pyvista/pyvista-wheels/raw/main/vtk-osmesa-9.1.0-cp38-cp38-linux_x86_64.whl
+      pip install --no-cache-dir -r requirements_test.txt
+      pip install --no-cache-dir -r requirements_docs.txt
+      pip install --no-cache-dir pre-commit
+      pre-commit install --install-hooks
+      pip install -e .
+      make -C doc html
+
+github:
+  prebuilds:
+    # enable for the default branch (defaults to true)
+    master: true
+    # enable for all branches in this repo (defaults to false)
+    branches: true
+    # enable for pull requests coming from this repo (defaults to true)
+    pullRequests: true
+    # enable for pull requests coming from forks (defaults to false)
+    pullRequestsFromForks: false
+    # add a check to pull requests (defaults to true)
+    addCheck: false
+    # add a "Review in Gitpod" button as a comment to pull requests (defaults to false)
+    addComment: false
+    # add a "Review in Gitpod" button to the pull request's description (defaults to false)
+    addBadge: false
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -6,6 +6,8 @@ image:
 ports:
   - port: 5500
     onOpen: notify
+  - port: 6080
+    onOpen: open-preview
 
 vscode:
   extensions:
@@ -14,7 +16,6 @@ vscode:
 
 tasks:
   - init: |
-      pip install --no-cache-dir https://github.com/pyvista/pyvista-wheels/raw/main/vtk-osmesa-9.1.0-cp38-cp38-linux_x86_64.whl
       pip install --no-cache-dir -r requirements_test.txt
       pip install --no-cache-dir -r requirements_docs.txt
       pip install --no-cache-dir pre-commit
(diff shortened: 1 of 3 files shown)


MatthewFlamm:
cc banesullivan since we discussed this before.  This PR is much more mature and I'm more confident in it

.gitpod.yml
```diff
@@ -0,0 +1,44 @@
...
+
+vscode:
+  extensions:
+    - ms-python.python
+    - ritwickdey.liveserver
+
+tasks:
+  - init: |
+      # this needs to be a proj var
+      export PRE_COMMIT_HOME=/workspace/.precommit
```
MatthewFlamm, line 18: This should be removed before merge, but put into a Gitpod project environment variable.  Kept this here for illustration of how this works.

docker/dev-gitpod.Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM gitpod/workspace-full
+LABEL maintainer="PyVista Developers"
+LABEL repo="https://github.com/pyvista/pyvista"
+
+RUN sudo apt-get install  -yq --no-install-recommends \
+    libosmesa6
+
+ENV PYTHONUSERBASE=/workspace/.pip-modules
```
MatthewFlamm, line 8: This could probably be removed if we keep using a virtual environment.

.gitpod.yml
```diff
@@ -0,0 +1,44 @@
...
+      [ ! -d venv ] && python -m venv venv
+      source venv/bin/activate
```
MatthewFlamm, line 20: I found that a virtual environment is required, or at least greatly simplifies installation, due to problems with one of `enum-tools` dependencies.  See Currently use of `enum_tools`. See https://github.com/sphinx-toolbox/sphinx-jinja2-compat/issues/2.

MatthewFlamm:
You should be able to spin up a version at https://gitpod.io/#https://github.com/MatthewFlamm/pyvista/tree/gitpod

.gitpod.yml
adeak: I'm wondering if we could/should abuse the fact that python executables in a venv know via `site` that they should be using the venv. Even if you don't activate the venv. Probably the same goes for scripts like `pytest`. We could hack something together with symlinks, or just plop `venv/bin` to the front of `$PATH`, and it would _probably_ "just work".

Hmm, or can we edit `/home/gitpod/.bashrc` to activate the venv on login?

adeak: Just came across https://github.com/gitpod-io/gitpod/issues/1555#issuecomment-635911024 saying in 2020

> In theory, most `.bashrc` customizations should be dropped into `/home/gitpod/.bashrc.d/` as separate files.

MatthewFlamm: Good thought.  I will do some testing.  I like the .bashrc path forward the best, if it works.  It would be the most transparent to the user.

adeak: It seems to me that the shell already has a pyenv environment activated, named `gitpod`. Probably has to do with the `60-python` script in `.bashrc.d/`. I'm not sure how we should interact with that.

MatthewFlamm: > It seems to me that the shell already has a pyenv environment activated, named gitpod.

I don't see any pyenv virtual environment named gitpod.  `pyenv virtualenvs` returns nothing.  Or is it something else?  One important thing in gitpod workspaces is that only files in /workspace persist.  Using a regular virtual environment works well for this.  I'm not too familiar with how to use pyenv for virtual environments.

MatthewFlamm: PS the word gitpod is given twice in the prompt.  One is for the username the other for the branch name on my fork.

adeak: > PS the word gitpod is given twice in the prompt. One is for the username the other for the branch name on my fork.

Sorry, that's what confused me. I saw the second "gitpod" in the prompt, saw .pyenv in the default `which python` and I saw pyenv things in `.bashrc.d/60-python`. I've never used pyenv, just jumped to conclusions.

MatthewFlamm: This isn't trivial, gitpod has several stages and these things don't persist or source when I expect them too.  I suspect that the `enum-tools` dependency is creating a headache in combination with the pyenv model in gitpod too.  pyenv does things I don't understand.

MatthewFlamm: Using .bashrc works, thanks for the suggestion! The key was to do it in the docker container.  If done in a gitpod task, it doesn't persist.  If this gets more complex, we could consider having a file that is copied into the container rather than constructing it line by line.  For two lines and a short docker file, I think the current status is fine.

MatthewFlamm:
This PR now includes the ability to get interactive plots through the vnc image all in the browser.  It works better than I thought it would.

This also fixes the problems with tests on this particular python version with OSMesa.

image

adeak:
I've yet to take a good look at this, but this sounds amazing! Great job.

MatthewFlamm:
For documentation, I'm thinking of breaking this into use cases below.  It is hard for me to test some of these use cases since this all exists currently on my fork.

1) For review, the prebuilt image allows a reviewer to open a workspace pointing to the `pyvista/pyvista` repo.
2) For development for users with write access to the pyvista repo, users should also be able to open a workspace pointing to the `pyvista/pyvista` repo and their branch will get pre-builds too.
  - I can't find clear guidelines for usage limits for pre-builds, so we may need to change this policy to only pre-build on `main`.
3) For development for users without write access to the pyvista repo, users will work on their fork.
  - I can't test how this works since the current development happens on my fork.
  - If someone is able to fork my fork, and test this, it would be very valuable.  Or, the documentation could be left until after this is merged and I can test how forks work directly.

MatthewFlamm:
For documentation, I decided to keep it short to get users started quickly.  And point to the official Gitpod documentation.  Otherwise, there are several edge cases to consider.  I added in short pointers to look for the vnc browser and the live documentation viewer.  I'm considering this fully ready for review.

One note for merging is that to manage the prebuilds, an administrator for pyvista probably needs to create a project.  After that more members can be invited for managing the prebuilds and so on.

banesullivan:
This looks really excellent! I haven't used gitpod, but have been using GitHub's codespaces and would love to do something similar there for the interactive plots like you show above

akaszynski (approved):
Nice work @MatthewFlamm. LGTM, thanks!

PR #3285, opened 2022-09-08
Use stable version python for intersphinx links

MatthewFlamm:
### Overview

 

Current documentation links to dev version of python, which is now 3.12.  Use latest stable version instead.


adeak (approved):
Good idea, thanks!

banesullivan (approved):
Oof, thanks for fixing this.

Issue #3269, opened 2022-09-04
Pickle Crashes for Large Data


adam-grant-hendry:
### Describe the bug, what's wrong, and what you expected.

`pickle` crashes for large data.

### Steps to reproduce the bug.

```python
import pickle

import numpy as np
import pyvista as pv
from vtkmodules.vtkIOLegacy import vtkDataSetWriter


def pickle_vtk(mesh, filename):
    writer = vtkDataSetWriter()
    writer.SetInputDataObject(mesh)
    writer.SetWriteToOutputString(True)
    writer.SetFileTypeToBinary()
    writer.Write()
    to_serialize = writer.GetOutputString()

    with open(filename, 'wb') as handle:
        pickle.dump(to_serialize, handle, protocol=pickle.HIGHEST_PROTOCOL)

    return filename


if __name__ == '__main__':
    dims = (2154, 1500, 1167)

    volume = pv.UniformGrid(
        dims=(2154, 1500, 1167),
        spacing=(1, 1, 1),
        origin=(0, 0, 0),
    )

    volume.point_data['scalars'] = np.zeros(
        shape=(dims[0] * dims[1] * dims[2],), dtype=np.uint8
    )

    fname = 'filename.vtkpickle'
    pickle_vtk(volume, fname)  # Crashes
```

However, I have no problem pickling the numpy array by itself to file (this is 3.7 GB).

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Sun Sep 04 11:59:36 2022 Pacific Daylight Time

                OS : Windows
            CPU(s) : 96
           Machine : AMD64
      Architecture : 64bit
               RAM : 190.7 GiB
       Environment : Python
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : Quadro RTX 8000/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 516.40

  Python 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64
  bit (AMD64)]

           pyvista : 0.37.dev0
               vtk : 9.1.0
             numpy : 1.23.2
           imageio : 2.21.2
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.3
         pyvistaqt : 0.9.0
           IPython : 8.4.0
              tqdm : 4.64.0
            meshio : 5.3.4
--------------------------------------------------------------------------------
```

adam-grant-hendry:
[...]

PR #3286, opened 2022-09-08
Pickle DataObjects with XML readers/writers

whophil:
### Overview
Pickle `DataObject` with XML readers/writers, which are more robust but slower than the legacy reader/writer.

This is based on discussion in https://github.com/pyvista/pyvista/discussions/3279 and #3269.

The "legacy" serialization produced by `vtkDataSetWriter` is still available; the choice between XML and "legacy" serialization is controllable via `pyvista.set_pickle_format()`.

This _should_ be backwards compatible with an older pickled `DataObject`. I didn't test it, nor do I think that use case is worth supporting. Pickling support of `DataObject` is primarily intended for multiprocessing situations, and not for disk persistence, where VTK standard data formats are _highly recommended_.

Copying some relevant bits from that discussion:
> The switch-like statements exist because the generic XML readers and writers are [broken](https://gitlab.kitware.com/vtk/vtk/-/issues/18661). The "switches" can be removed in the future when those readers/writers are fixed in VTK.

CC @akaszynski @MatthewFlamm @banesullivan @adeak


pyvista/core/dataobject.py
```diff
@@ -535,28 +535,82 @@ def copy_attributes(self, dataset: _vtk.vtkDataSet):
...
+            elif isinstance(self, _vtk.vtkStructuredGrid):
+                writer = _vtk.vtkXMLStructuredGridWriter()
+            elif isinstance(self, _vtk.vtkRectilinearGrid):
+                writer = _vtk.vtkXMLRectilinearGridWriter()
+            elif isinstance(self, _vtk.vtkUnstructuredGrid):
+                writer = _vtk.vtkXMLUnstructuredGridWriter()
+            elif isinstance(self, _vtk.vtkPolyData):
+                writer = _vtk.vtkXMLPolyDataWriter()
+            elif isinstance(self, _vtk.vtkTable):
+                writer = _vtk.vtkXMLTableWriter()
```
adeak, line 553: I guess we could make this slightly more concise by defining a pair of classes and writers, something like
```py
writers = {
    _vtk.vtkImageData: _vtk.vtkXMLImageDataWriter,
     _vtk.vtkStructuredGrid: _vtk.vtkXMLStructuredGridWriter,
    ...
}
for parent_type, writer_type in writers.items():
    if isinstance(self, parent_type):
        writer = writers[parent_type]
        break
else:
    raise TypeError(f'Invalid mesh type {type(self).__name__}')
```
So basically a dispatch dict with the additional step of allowing subclasses.

pyvista/core/dataobject.py
```diff
@@ -535,28 +535,82 @@ def copy_attributes(self, dataset: _vtk.vtkDataSet):
         self.CopyAttributes(dataset)
 
     def __getstate__(self):
-        """Support pickle. Serialize the VTK object to ASCII string."""
+        """Support pickle. Serialize the VTK object to `bytes` or `str`."""
```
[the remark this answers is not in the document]
whophil: Sorry, you are correct but the docstring is out of date.

whophil:
This should be ready now @MatthewFlamm @adeak

banesullivan (approved):
This is really nicely done! Thanks, @whophil! Great work switching to the XML writers but still having backward compat

tests/test_dataset.py
```diff
@@ -1166,6 +1169,21 @@ def test_serialize_deserialize(datasets):
             assert arr_have == pytest.approx(arr_expected)
 
 
+def n_points(dataset):
+    # used in multiprocessing test
+    return dataset.n_points
+
+
```
adeak, line 1176: Can we stick this inside the test as a nested function, or just use a lambda _in situ_? In library code I'd be tempted to use `operator.attrgetter`, but probably not worth it in a test.

pyvista/utilities/misc.py
```diff
@@ -151,3 +152,14 @@ def can_create_mpl_figure():  # pragma: no cover
 
 
 vtk_version_info = VTKVersionInfo()
+
+
+def set_pickle_format(format: str):
+    """Set the format used to serialize :class:`pyvista.DataObject` when pickled."""
+    supported = {'xml', 'legacy'}
+    format = format.lower()
```
adeak, line 160: Regarding the API:

1. Do we prefer that users use `set_pickle_format()`? Or can they also change  `pyvista.PICKLE_FORMAT` directly?
2. We check stuff in `set_pickle_format()`, but then if this is preferred, shouldn't we consider `pyvista.PICKLE_FORMAT` to be private, and also provide a getter?
3. Can we document somewhere that there are these two format options, and the preferred way of choosing? Especially since we default to legacy, it would be nice to make the opt-in newer feature discoverable. An example perhaps?

adeak (commented):
Looks good to me to overall, just a few remarks/questions.

tests/test_dataset.py
banesullivan: > Can we stick this inside the test as a nested function

I don't think we can. multiprocessing will complain as it needs globals, not locals

adeak: Ah, good shout. Too bad.

pyvista/utilities/misc.py
whophil: On 1. and 2. - The user can set `pyvista.PICKLE_FORMAT` directly, or by using the `set_pickle_format()` method. If it is set manually to an invalid value, an error will be thrown pickling. This error at pickle-time is probably sufficient on its own, so we can probably remove the `set_pickle_format()` method.

On 3. - Yes, I will add a note in the docs on how this works. Clearly it is confusing, because as written, it's currently not opt-in to use the new XML readers/writers :) - the XML readers/writers are used by default. Perhaps you were confused by the comment in the `__setstate__` method; this comment states that a pickled is object is assumed to be pickled in a "legacy" format only if its dict does not contain a `PICKLE_FORMAT` key. This is for backwards compatibility, as old pickled objects will not have this key.

adeak: That's correct, I didn't pay enough attention, sorry 😅 your design is perfect.

Issue #3244, opened 2022-08-31
Wireframe style is not coloring edges


rodrigomologni:
### Describe the bug, what's wrong, and what you expected.

Hi, developers! The parameter `color` in `add_mesh` is not working when `style="wireframe"`. I think the bug was introduced at #2941.

### Steps to reproduce the bug.

```python
import pyvista as pv
plotter = pv.Plotter()
plotter.add_axes()
plotter.add_mesh(pv.Box(), style="wireframe", color="red")
plotter.show()
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Wed Aug 31 14:17:57 2022 Hora oficial do Brasil

                OS : Windows
            CPU(s) : 8
           Machine : AMD64
      Architecture : 64bit
               RAM : 15.9 GiB
       Environment : Python
        GPU Vendor : ATI Technologies Inc.
      GPU Renderer : AMD Radeon (TM) R9 370
       GPU Version : 4.5.13571 Core Profile Context 19.30.02 26.20.13002.133

  Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64
  bit (AMD64)]

           pyvista : 0.37.dev0
               vtk : 9.1.0
             numpy : 1.23.2
           imageio : 2.21.1
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.3
         pyvistaqt : 0.9.0
             PyQt5 : 5.15.7
           IPython : 8.4.0
             scipy : 1.9.0
              tqdm : 4.64.0
            meshio : 4.4.6
        jupyterlab : 3.4.5
--------------------------------------------------------------------------------
[Finished in 10.0s]
```

### Screenshots

![image](https://user-images.githubusercontent.com/1608652/187728449-bee578c3-c4e9-4bd2-a56d-16be2b355e6b.png)

PR #3287, opened 2022-09-08
set active scalars to be visible in collision example

banesullivan:
| `main` | this branch |
|---|---|
|sphx_glr_collisions_001 | collision_movie |

The active scalars behavior has changed since this example was written. This example is supposed to show the active "collisions" scalar, but it wasn't. I bet there are a ton of other examples missing this as well since the behavior of active scalars drastically changed not too long ago.

### DIFF
--- a/examples/01-filter/collisions.py
+++ b/examples/01-filter/collisions.py
@@ -52,7 +52,7 @@ sphere1 = pv.Sphere(radius=0.6, center=(-1, 0, 0))
 
 pl = pv.Plotter()
 pl.enable_hidden_line_removal()
-pl.add_mesh(sphere0, show_scalar_bar=False, cmap='bwr')
+pl.add_mesh(sphere0, scalars='collisions', show_scalar_bar=False, cmap='bwr')
 pl.camera_position = 'xz'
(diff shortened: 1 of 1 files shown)


banesullivan:
Also, I don't know why the wireframe mesh on this branch is black instead of green like the example specifies

banesullivan:
Actually, @pyvista/developers, the difference between the following  concerns me

- (scalars aren't visible) https://docs.pyvista.org/examples/01-filter/collisions.html#sphx-glr-examples-01-filter-collisions-py
- (scalars are visible, but wireframe is black, not green) https://dev.pyvista.org/examples/01-filter/collisions.html#sphx-glr-examples-01-filter-collisions-py

akaszynski:
> Do we not run the examples gallery as a CI check?

We do, but it's cached. We made that tradeoff a while ago when we decided to cache the results of the doc build and I guess this is the price we pay.

akaszynski:
> wireframe is black, not green

This was pointed out in https://github.com/pyvista/pyvista/issues/3244 and should be high priority before the next release. Creating milestone.

akaszynski (approved):
LGTM. Will fix the other issues with follow-up PRs.

PR #3291, opened 2022-09-09
Clean up CI job conditions to prevent PRs on forks

banesullivan:
This PR prevents the CI jobs that interact with the repository (generating comments, issues, or PRs) from running on any forks. Downstream forks are seeing some of these actions creating PRs in their repositories. Besides this being a nuisance, it can lead to problems if someone merges one of those before opening a PR with their new changes (leading to a conflict, albeit a simple one).

I've limited this to actions that generate comments, issues, or PRs but not things like labeling.

I have also deleted the `merge-me` action as I'm not sure what that is and can't remember where it came from and I don't think it is being used. @pyvista/developers, is this needed?


akaszynski:
> I have also deleted the `merge-me` action as I'm not sure what that is and can't remember where it came from and I don't think it is being used. @pyvista/developers, is this needed?

That's mine and it needs to be deleted. It's not used and probably doesn't work.

akaszynski (approved):
LGTM.

Issue #3314, opened 2022-09-13
Migrating helper/utility code out of the `core` subpackage and reorganizing


banesullivan:
I would like us to move in a direction such that the `core` subpackage focuses on strictly wrapping the VTK data object types and implementing filters (with filters being nested in a sub-sub-package as they are now). We've done a great job of this so far, but a few artifacts have slipped by that seem out of place. Namely:

- `pyvista/core/common_data.py`: Why is this in the `core` module? The single method `perlin_noise` seems to me like it should be in a utils file or in the `examples` module.
- `pyvista/core/imaging.py`: Again, this is more of a utility/helper method type thing and does not pertain to the "core" API.

Further, I would like to refactor the organization of classes into more structured files following the hierarchy of the underlying VTK inheritance tree. 

![Screen Shot 2022-09-13 at 12 53 56 PM](https://user-images.githubusercontent.com/22067021/189986391-73e6124c-6dcc-49e6-ac01-80d7344a0007.png)

The file structure should generally follow this tree, for example:

- `core/`
  - `cell.py`
  - `dataset.py`: core `DataSet` wrapping
  - `datasetattributes.py`: field data wrapping
  - `errors.py`
  - `image.py`: `vtkImageData` wrapping (`pyvista.UniformGrid`)
  - `dataobject.py`: `DataObject` wrapping
  - `rectilinear.py`: `vtkRectilinearGrid` wrapping
  - `pointset/`: all `vtkPointSet` subclasses
    - `explicit.py`: `vtkExplicitStructuredGrid` wrapping
    - `points.py`: `vtkPointSet` wrapping
    - `poly.py`: `vtkPolyData` wrapping
    - `structured.py`: `vtkStructuredGrid` wrapping
    - `unstructured.py`: `vtkUnstructuredGridBase` wrapping and subclasses
  - `composite/`
    - `composite.py`: `vtkMultiBlockDataSet` wrapping
    - `partioned.py`: `vtkPartionedDataSet` wrapping

@pyvista/developers, what do you think about this reorganization? This should help make many of the files more manageable (for example, the `pyvista/core/pointset.py` is currently approaching 3,000 lines of code but will be breaking up into multiple files).

adeak:
I see the purity argument behind following VTK's inheritence tree, but then again having a tree like
```
.
└── core
    ├── image.py that is actually UniformGrid
    ├── pointset
    │   ├── explicit.py
    │   ├── points.py for PointSet
    │   ├── poly.py
[...]

PR #3321, opened 2022-09-14
Move utility methods out of core subpackage

banesullivan:
Partially address #3314 by moving some utility methods out of the core sub-package and into the utilities sub-package. I'm open to suggestions for other, better places for these methods, but I would really like to move these out of the `core` sub package.


akaszynski (approved):
I see no issues with this. `common_data` was sorta a placeholder anyway until we figured out where `perlin_noise` was going to end up.

LGTM.

PR #3329, opened 2022-09-15
[maint] update conda environment

banesullivan:
Update the conda environment file to reflect current dependencies (notably bumping the minimum version of Python)

To do: I'd like us to audit this and make sure it only has the needed reqs for where this is used


akaszynski:
Note to add `pooch`.

banesullivan:
I think this is good to go

tkoyama010 (approved):

PR #3332, opened 2022-09-15
Create Mapper and Actor classes

akaszynski:
I've (finally) gotten around to getting rid of `make_mapper` and creating Pythonic wrappers of both the `Mapper` classes and `Actor` class. There are a variety of users on several closed (to be open eventually) projects that I work with that directly modify actors and I think it would be useful to have pythonic attributes and methods for these VTK classes.

In the process of working through the examples for this, I discovered backface properties might be really useful to some users:

```py
import numpy as np
import pyvista as pv
pv.set_plot_theme('document')
plane = pv.Plane(i_size=1.5, j_size=1.5)
mesh = pv.Sphere().clip_surface(plane, invert=False)
pl = pv.Plotter()
actor = pl.add_mesh(mesh, smooth_shading=True)
actor.backface_prop.color = 'lightblue'
_ = pl.add_mesh(
    plane, opacity=0.25, show_edges=True, color='grey', lighting=False,
)
pl.enable_anti_aliasing('ssaa')
pl.show()
```

tmp

Please let me know if you see any attributes or methods that are missing.


banesullivan (approved):
Excellent! Excited about this one!

PR #3336, opened 2022-09-17
Wrap the lookup table

akaszynski:
Wraps the lookup table to enable pythonic usage of the lookup table. Comes with a neat plotting feature:
```py
import pyvista as pv
lut = pv.LookupTable()
lut.values = [
    [0, 0, 0, 255],
    [85, 0, 0, 255],
    [170, 0, 0, 255],
    [255, 0, 0, 255],
]
lut.nan_color = 'grey'
lut.below_range_color = 'blue'
lut.above_range_color = 'lightblue'
lut.plot()
```
tmp

And

```py
from pyvista import examples
import pyvista as pv
lut = pv.LookupTable()
lut.cmap ='Greens'
lut.n_values = 8

pl = pv.Plotter()
actor = pl.add_mesh(mesh, scalars=mesh.points[:, 1], lookup_table=lut)
pl.plot()
```

tmp2

Includes a nice `__repr__`
```
>>> actor.mapper.lookup_table
LookupTable (0x7fe25e5d6ac0)
  Table Range:                (0.0, 1.0)
  N Values:                   8
  Above Range Color:          None
  Below Range Color:          None
  NAN Color:                  Color(name='maroon', hex='#800000ff')
  Log Scale:                  False
  Color Map:                  "Greens"
```


examples/02-plot/lookup-table.py
```diff
@@ -0,0 +1,236 @@
+"""
+.. _lookup_table_example:
+
+Anti-Aliasing
+~~~~~~~~~~~~~
```
banesullivan, line 5:
```suggestion
Lookup Tables
~~~~~~~~~~~~~
```

examples/02-plot/lookup-table.py
```diff
@@ -0,0 +1,236 @@
...
+
+The :class:`pyvista.LookupTable` can be used to have fine-tuned control over
+the mapping between a :class:`pyvista.DataSet`'s scalars and RGBA colors.
+
+"""
+
+import pyvista as pv
+from pyvista import examples
+
+pv.set_plot_theme('document')
```
banesullivan, line 16: Isn't this the default? It should be and we shouldn't need this IMO

examples/02-plot/lookup-table.py
```diff
@@ -0,0 +1,236 @@
...
+
+###############################################################################
+# Plot the Lookup Table
+# ~~~~~~~~~~~~~~~~~~~~~
+# You can also plot lookup table to see the mapping between the scalar values
+# (here, between 0 and 23.3) and RGBA colors.
+
+pl = pv.Plotter()
+actor = pl.add_mesh(bracket)
+actor.mapper.lookup_table.plot()
```
banesullivan, line 44: I really don't like this logic... `Plotter` objects ought to be useable across cells. We should consider fixing this in the SG scraper

examples/02-plot/lookup-table.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+###############################################################################
+# Create a Custom Lookup Table using a matplotlib Color Map
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Here we create a lookup table with a narrow table range (same as ``clim``)
+# and color values above and below the range.
+
+lut = pv.LookupTable(cmap='magma')
+lut.table_range = (5, 15)
```
banesullivan, line 64: `table_range` vs. `value_range` is confusing me...

pyvista/plotting/plotting.py
```diff
@@ -2245,6 +2246,11 @@ def add_composite(
             have these updates rendered, e.g. by changing the active scalars or
             through an interactive widget.  Defaults to ``False``.
 
+        lookup_table : pyvista.LookupTable, optional
+            Use a custom lookup table instead of a color map. If this is set,
+            all parameters controlling the color map like ``cmap`` and
+            ``n_colors`` will be ignored.
```
banesullivan, line 2252: Personally, I'd prefer to pass the `LookupTable` object as the `cmap` parameter. Is this supported? (lots of code to sift through)

pyvista/plotting/plotting.py
```diff
@@ -2765,12 +2773,18 @@ def add_mesh(
...
-            if you would like to update the mesh after adding it to the plotter and have these
-            updates rendered, e.g. by changing the active scalars or through an interactive widget.
-            Defaults to ``False``.
+            If ``True``, a copy of the mesh will be made before adding it to
+            the plotter.  This is useful if e.g. you would like to add the same
+            mesh to a plotter multiple times and display different
+            scalars. Setting ``copy_mesh`` to ``False`` is necessary if you
+            would like to update the mesh after adding it to the plotter and
+            have these updates rendered, e.g. by changing the active scalars or
+            through an interactive widget.  Defaults to ``False``.
```
banesullivan, line 2782: We should more strongly warn against using this option (for another issue or PR though)

pyvista/plotting/plotting.py
```diff
@@ -3500,35 +3516,24 @@ def add_volume(
...
-        # Set colormap and build lookup table
-        table = _vtk.vtkLookupTable()
-        # table.SetNanColor(nan_color) # NaN's are chopped out with current implementation
-        # above/below colors not supported with volume rendering
-
-        if isinstance(annotations, dict):
-            for val, anno in annotations.items():
-                table.SetAnnotation(float(val), str(anno))
+        if not has_module('matplotlib'):
+            raise ImportError('Please install matplotlib for volume rendering.')
```
banesullivan, line 3520: We can't use just a LUT with volume rendering??

pyvista/plotting/lookup_table.py
```diff
@@ -0,0 +1,754 @@
...
+        if self._cmap:
+            return None
+        return self.GetSaturationRange()
+
+    @saturation_range.setter
+    def saturation_range(self, value: tuple):
+        self.SetSaturationRange(value)
+        self._rebuild()
+
+    def _rebuild(self):
```
banesullivan, line 385: I'd be +1 for this not being a private method

pyvista/plotting/lookup_table.py
```diff
@@ -0,0 +1,754 @@
...
+            return Color(self.GetAboveRangeColor())
+        return None
+
+    @above_range_color.setter
+    def above_range_color(self, value):
+        if value is None:
+            self.SetUseAboveRangeColor(False)
+        else:
+            self.SetAboveRangeColor(*Color(value).float_rgba)
+            self.SetUseAboveRangeColor(True)
```
banesullivan, line 506: What happens if I do something logical like: `lut.above_range_color = True` considering that `color=True` is a valid argument for the `add_mesh` method, there is precedent for this. I'd suggest `True` being a special case that either uses the default color or whatever was last set (default color being pulled from the active theme)

banesullivan: Similarly for all like methods

pyvista/plotting/lookup_table.py
```diff
@@ -0,0 +1,754 @@
...
+
+    @below_range_color.setter
+    def below_range_color(self, value):
+        if value is None:
+            self.SetUseBelowRangeColor(False)
+        else:
+            self.SetBelowRangeColor(*Color(value).float_rgba)
+            self.SetUseBelowRangeColor(True)
+
+    def _apply_cmap(self, cmap, n_values=256, flip=False):
```
banesullivan, line 544: +1 for this not being private. LUT are an advanced feature and if a user is dabbling into them, I'd like to give them full control

pyvista/utilities/misc.py
```diff
@@ -163,3 +163,20 @@ def set_pickle_format(format: str):
...
+        if hasattr(self, name):
+            object.__setattr__(self, name, value)
+        else:
+            raise AttributeError(
+                f'Attribute "{name}" does not exist and cannot be added to type '
+                f'{self.__class__.__name__}'
+            )
+
+    setattr(cls, '__setattr__', __setattr__)
+    return cls
```
banesullivan, line 182: Wow is this nice (and needed in Python std libs)

examples/02-plot/lookup-table.py
banesullivan: While these get cleaned up automatically in sphinx gallery, they do not when a a user copy/pastes these snippets into their shell. I think this encourages improper use of the API and this is not intuitive

examples/02-plot/lookup-table.py
```diff
@@ -0,0 +1,236 @@
...
+
+###############################################################################
+# Plot the bracket with the custom colormap
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# You can set assign the lookup table when using ``add_mesh``.
+
+
+pl = pv.Plotter()
+actor = pl.add_mesh(bracket, lookup_table=lut, lighting=False)
+pl.show()
```
banesullivan, line 78: Is this supposed to look like this? Feels off to me...

Screen Shot 2022-09-21 at 9 22 33 AM

pyvista/plotting/lookup_table.py
```diff
@@ -0,0 +1,754 @@
...
+        """Return the lookup type."""
+        if self.cmap:
+            if hasattr(self.cmap, 'name'):
+                return f'{self.cmap.name}'  # type: ignore
+            else:  # pragma: no cover
+                return f'{self.cmap}'
+        elif self._values_manual:
+            return 'From values array'
+        else:
+            return 'VTK lookup table'
```
banesullivan, line 692:
```suggestion
            return 'PyVista Lookup Table'
```

examples/02-plot/lookup-table.py
```diff
@@ -0,0 +1,236 @@
...
+pl = pv.Plotter()
+actor = pl.add_mesh(bracket, lookup_table=lut, lighting=False)
+make_double_slider('alpha_range', 0)
+make_double_slider('hue_range', 1)
+make_double_slider('value_range', 2)
+make_double_slider('saturation_range', 3)
+make_double_slider('table_range', 4)
+
+pl.camera_position = [(9.021, 5.477, 7.780), (-0.679, 1.349, 0.874), (-0.498, -0.228, 0.836)]
+cpos = pl.show(return_cpos=True)
```
banesullivan, line 236: Someone had fun! Looks so awesome!!

Screen Shot 2022-09-21 at 9 26 17 AM

pyvista/plotting/lookup_table.py
```diff
@@ -0,0 +1,754 @@
...
+      Color Map:                  "inferno"
+    >>> lut.plot()
+
+    """
+
+    _nan_color_set = False
+    _cmap = None
+    _values_manual = False
+
+    def __init__(self, cmap=None, n_values=256, flip=False):
```
banesullivan, line 149: Can we support all properties as kwargs in the `__init__`? I'd expect to do something like the following in a single line:

```py
lut = pv.LookupTable(cmap='bwr', table_range=(2,20))
```

banesullivan (changes requested):
This PR is a big deal for me! I'm really excited to see this and appreciate how much thought and effort you put into this, @akaszynski! It's super nice!

I have a few bits of feedback which some are feeling nitpicky, so bare with me


1. I'd like for `cmap`s and `lookup_tables` to be somewhat interchangeable in parts of the API (at least for the `add_mesh` method).

Ideally, I'd like to do something like the following:

```py
lut = pv.LookupTable(cmap='jet')

pl = pv.Plotter()
actor = pl.add_mesh(pv.Wavelet(), cmap=lut)
pl.show()
```

This currently has two issues. a) we get a non-descriptive error when passing the `LookupTable` as the `cmap` argument and b) the value range of the lookup table is (0,1) rather than the active scalar range. 

2. Usage patterns (in comments below) - I'd like the `__init__` method of the `LookupTable` to accept all properties by name.

3. See inline comments


## Some feedback that has me really excited!

This resolves a feature request I've had a for a long time (I don't think it's captured in an issue) where `cmap`'s should be able to be dynamically updated! I'm thrilled to see this.

```py
pl = pv.Plotter()
actor = pl.add_mesh(pv.Wavelet(), cmap='bwr')
pl.show()

...
actor.mapper.lookup_table.cmap='jet'
```

examples/02-plot/lookup-table.py
akaszynski: Yep, anything below a certain stress is opaque. It's not perfect and it's a weird cmap, but it's just a demo.

examples/02-plot/lookup-table.py
akaszynski:
```suggestion
```

examples/02-plot/lookup-table.py
akaszynski: Agreed and this should be fixed in a later PR.

examples/02-plot/lookup-table.py
akaszynski: Agreed, issue is there's `SetRange` would end up being `range`, and this is a python keyword, so I used  [SetTableRange](https://vtk.org/doc/nightly/html/classvtkLookupTable.html#ac5c77be0b309208ba5913f77ccdc8922) instead.

I think `scalar_range` would actually be best. Ping me if you see an issue with that.

banesullivan: +1 for `scalar_range` wrapping `SetRange` instead of `table_range`

pyvista/plotting/plotting.py
akaszynski: I can do that. I feel we have too many keyword arguments.

pyvista/plotting/plotting.py
akaszynski:
```suggestion
            If ``True``, a copy of the mesh will be made before adding it to
            the plotter.  This is useful if you would like to add the same
            mesh to a plotter multiple times and display different
            scalars. Setting ``copy_mesh`` to ``False`` is necessary if you
            would like to update the mesh after adding it to the plotter and
            have these updates rendered, e.g. by changing the active scalars or
            through an interactive widget. This should only be set to ``True``
            with caution. Defaults to ``False``.
```

akaszynski: Using more fighting words.

pyvista/utilities/misc.py
adeak: I don't have context here so sorry if this is a dumb question, but why can't we use `__slots__`?

pyvista/plotting/lookup_table.py
akaszynski: Agreed.

pyvista/plotting/plotting.py
akaszynski: Non-trivial, but it was implemented and had the side effect of pulling more logic out of `pyvista.Plotter`.

akaszynski:
@banesullivan, ready for review.

banesullivan (approved):
So excited for this to merge! Excellent work, @akaszynski!

banesullivan:
Let's merge this, @akaszynski

Issue #1743, opened 2021-10-22
Stray voxels occurring when voxelizing a mesh


anthops:
**Describe the bug, what's wrong, and what you expect:**

When applying pv.voxelize( ) to a mesh, there are voxels occurring outside of the mesh. For example here is a side by side comparison of the mesh and resultant voxels of pv.Sphere()
![image](https://user-images.githubusercontent.com/32993852/138474002-5e5d4a77-8570-4562-8b27-f8a7d6839b55.png)

Here is also a side by side comparison for examples.download_cow()
![image](https://user-images.githubusercontent.com/32993852/138474403-44324c96-c71b-4471-8df0-e1c45dec757a.png)

I expect the voxels to match the shape of the mesh, without any outlier cubes.

-----

**To Reproduce**

Below is the code to replicate the cow voxel. 

```py
import numpy as np

import pyvista as pv
from pyvista import examples

mesh = examples.download_cow()

density = mesh.length/100
voxels = pv.voxelize(mesh, density)

plotter = pv.Plotter()

plotter.add_mesh(voxels, opacity = 0.5)

plotter.show()

```
It is also worth noting that running the exact same code on my friend's computer produced a voxel with way less outlier cubes. There were still outliers present but the shape of the cow could be made out.

I also only began to notice this bug after a few days of using the pyvista library. Objects such as the sphere used to be perfect but then randomly started to display these outliers in the voxel.

**System Information:**
 Please run the following code wherever you are experiencing the bug and paste the output below. This report helps us track down bugs and it is critical to addressing your bug:

```py
# Get system info
import pyvista as pv
print(pv.Report())
```

```txt
--------------------------------------------------------------------------------
  Date: Sat Oct 23 01:18:14 2021 Cen. Australia Summer Time

                OS : Windows
            CPU(s) : 32
           Machine : AMD64
      Architecture : 64bit
               RAM : 63.9 GiB
       Environment : Python
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : NVIDIA GeForce RTX 3090/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 496.13

  Python 3.7.7 (default, May  6 2020, 11:45:54) [MSC v.1916 64 bit (AMD64)]

           pyvista : 0.32.1
               vtk : 9.0.1
             numpy : 1.21.2
           imageio : 2.9.0
           appdirs : 1.4.4
            scooby : 0.5.7
            meshio : 4.4.6
[...]

PR #3361, opened 2022-09-24
Triangulate for voxelization to prevent some artifacts

adeak:
This partly addresses https://github.com/pyvista/pyvista/issues/1743, but it's probably not the whole picture.

The example snippet from that issue thread which has artifacts:
```py
import numpy as np

import pyvista as pv
from pyvista import examples

density = mesh.length/300
voxels = pv.voxelize(mesh, density)

plotter = pv.Plotter()
plotter.add_mesh(voxels, opacity=0.5)
plotter.enable_parallel_projection()
plotter.show_axes()
plotter.show()
```
On `main` I get
voxelized cow with lots of artifact voxels
where there are strings of artifact voxels along the `[1, 1, 1]` direction. Adding `.triangulate()` to `mesh` above (which is the same thing as the patch in this PR) makes the artifacts go away:
voxelized cow with no artifacts

The reason why I don't think this is a complete fix is because the original comment on the issue shows artifacts with a sphere (although I can't reproduce these). Since `pv.Sphere()` is already triangulated, there might be other edge cases still.

I also wondered if we should put the triangulation fix into `select_enclosed_points`, because at the end of the day that seems to be the culprit here.


adeak:
Errors are coming from tests but I'm not sure I'm the one who's wrong here :smile: 

Here's one of the tests:
```py
import pyvista
from pyvista import examples as ex

def test_voxelize():
    mesh = pyvista.PolyData(ex.load_uniform().points)
    vox = pyvista.voxelize(mesh, 0.5)
    assert vox.n_cells
```
Now, what happens here is that we have a point cloud in `mesh`, and originally the function "worked" because
```py
>>> mesh.is_manifold
True
```
Arguably the point cloud has no open edges, yet when I hear "manifold mesh" I think of something different, and I find it hard to argue about the semantics of voxelization for a point cloud (where our strategy is to define a structured grid and select points _enclosed in the manifold mesh_).

So I'd be tempted to break the "voxelize point cloud" test cases and making sure that the input mesh actually has any faces before going forward with triangulation. Thoughts?

akaszynski:
> So I'd be tempted to break the "voxelize point cloud" test cases and making sure that the input mesh actually has any faces before going forward with triangulation.

Agreed. For the purposes of surface geometry, I tend to follow the definition of a manifold surface from [Wolfram - Manifold](https://mathworld.wolfram.com/Manifold.html), where we're specifically describing a manifold with boundary. In this case, manifold in the 3rd dimension.

By definition, for this to be true there must be topology. If it's missing faces, it can't even be considered. In fact, if the number of faces is less than four in 3D, it can't be manifold.

---

Recommend breaking it.

pyvista/utilities/features.py
```diff
@@ -56,6 +56,15 @@ def voxelize(mesh, density=None, check_surface=True):
     if isinstance(density, (list, set, tuple)):
         density_x, density_y, density_z = density
 
+    # check and pre-process input mesh
+    surface = mesh.extract_geometry()  # filter preserves topology
```
adeak, line 60: I replaced `extract_surface()` with `extract_geometry()`. According to https://vtk.org/doc/nightly/html/classvtkGeometryFilter.html#details vs. https://vtk.org/doc/nightly/html/classvtkDataSetSurfaceFilter.html the trade-offs are not obvious in modern VTK. Both filters delegate to the other for linear/non-linear cells, so either case is probably optimal with vanilla arguments.

But crucially `extract_geometry()` preserves the topology, so we end up with manifold meshes when applying it to volumes. For the equivalent behaviour we'd need `extract_surface().clean()`. I suspect that would be overall slower, and the docs mention that `vtkDataSetSurfacefilter` used by `extract_surface()` can also use a lot of memory. If there are potential edge cases where this change might matter, I'd gladly add test cases.

pyvista/utilities/features.py
```diff
@@ -56,6 +56,15 @@ def voxelize(mesh, density=None, check_surface=True):
...
+    if not surface.faces.size:
+        # we have a point cloud or an empty mesh
+        raise ValueError('Input mesh must have faces for voxelization.')
```
adeak, line 63: And now we explicitly check for faces. We could still have isolated points along with a manifold mesh, or multiple disjoint manifold meshes (which might or might not work reliably), but checking for that is probably not worth it in the majority of use cases.

pyvista/utilities/features.py
akaszynski: `clean` is slow primarily because it checks for duplicates (there's some intermediate data structure it creates), and I'm fine with using `extract_geometry`.

As for the edge cases, when those come up we can check for them. For the time being, this PR is created because of an edge case in the first place, so I'm fine with going down the `extract_geometry` route.

pyvista/utilities/features.py
akaszynski: Thanks. This works and allows for a quick fail for a good reason.

akaszynski (approved):
Local testing shows no change in the voxelization time between this branch and `main`.

LGTM. Thanks for fixing this and improving the examples.

pyvista/utilities/features.py
```diff
@@ -36,26 +36,34 @@ def voxelize(mesh, density=None, check_surface=True):
...
     """
     if not pyvista.is_pyvista_dataset(mesh):
         mesh = pyvista.wrap(mesh)
     if density is None:
         density = mesh.length / 100
-    if isinstance(density, (int, float)):
+    if isinstance(density, (int, float, np.number)):
         density_x, density_y, density_z = [density] * 3
-    if isinstance(density, (list, set, tuple)):
+    elif isinstance(density, (collections.abc.Sequence, np.ndarray)):
```
adeak, line 55: Thanks for the review @akaszynski and sorry for invalidating it. I had to fix one more thing I noticed late: we used to allow `set` in there. This is dangerous because sets are not sequences (they don't have well-defined order) but can be unpacked, leading to surprises:
```py
>>> x, y, z = {-2, -3, -4}

>>> x, y, z
(-4, -3, -2)
```
And while I was at it I generalized the types a bit, allowing both sequences and arrays as well as numpy scalars.

akaszynski: Sharp eye @adeak.

Quite right about sets.

akaszynski (approved):

PR #3385, opened 2022-09-29
Trame support and new Jupyter backend

### LINKED ISSUE #2940: JupyterLab : The widget returned by `plotter.show(return_viewer=True)` isn't updated when the plotter is

I am trying to have a widget returned by the method `plotter.show(return_viewer=True)` interact with an ipywidget slider in JupyterLab and pythreejs backend, but the widget scene doesn't get updated when the plotter actors change.

However, I works as expected with the `@interact` decorator.

## Doesn't work as expect
```python 
import pyvista as pv
pv.start_xvfb()
pv.set_jupyter_backend('pythreejs')
from pyvista import examples
from ipywidgets import interact

new_plotter = pv.Plotter()
vol = examples.download_brain()
slice = vol.slice()
new_act = [None]
new_act[0] = new_plotter.add_mesh(slice)

int_range =  widgets.IntSlider(min=0, max=180, step=1, )

widget = new_plotter.show(interactive_update=True, return_viewer=True, auto_close=False)

def update_slice(change):
    x = change["new"]
    new_slice = vol.slice(origin=(x, 0, 0))
    new_plotter.remove_actor(new_act[0], render=False, reset_camera=False)
    new_act[0] = new_plotter.add_mesh(new_slice, render=False, reset_camera=False)
    new_plotter.update()
    new_plotter.show()

int_range.observe(update_slice, names=['value'])

display(int_range, widget)
 
 ```

## Works as expected

```python
import pyvista as pv
pv.start_xvfb()
pv.set_jupyter_backend('pythreejs')
from pyvista import examples
from ipywidgets import interact

vol = examples.download_brain()
p = pv.Plotter()
slice = vol.slice()
actors = [None]
actors[0] = p.add_mesh(slice)
p.show(interactive_update=True, return_viewer=True, auto_close=False)

@interact(x=(0.0,180.0,1.0), continuous_update=False)
def update_slice(x):
    slice = vol.slice(origin=(x, 0, 0))
    p.remove_actor(actors[0], render=False, reset_camera=False)
    actors[0] = p.add_mesh(slice, render=False, reset_camera=False)
    p.show(interactive_update=True, auto_close=False)
```

### System Information

```shell
  Date: Mon Jul 04 16:53:41 2022 UTC

                OS : Linux
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 54.9 GiB
       Environment : Jupyter
        GPU Vendor : VMware, Inc.
      GPU Renderer : llvmpipe (LLVM 10.0.0, 256 bits)
       GPU Version : 3.3 (Core Profile) Mesa 20.0.8

  Python 3.8.0 (default, Dec  9 2021, 17:53:27)  [GCC 8.4.0]

           pyvista : 0.33.2
               vtk : 9.0.1
             numpy : 1.19.5
           imageio : 2.9.0
           appdirs : 1.4.4
            scooby : 0.5.7
        matplotlib : 3.3.4
           IPython : 7.16.1
        ipyvtklink : 0.2.2
             scipy : 1.8.1
              tqdm : 4.60.0
            meshio : 4.4.1
```

### Screenshots

_No response_

### Code of Conduct

### LINKED ISSUE #3091: Display grid with pythreejs backend

Hi,
Is it possible to use the `show_grid` method of the potter in a notebook with pythreejs backend?

https://docs.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.show_grid.html

### LINKED ISSUE #3221: "pythreejs" jupyter backend requires virtual framebuffer, contrary to documentation

Contrary to the documentation, enabling the "pythreejs" backend for jupyter notebooks causes an error suggesting to use a virtual framebuffer when X server is not available:

```
This system does not appear to be running an xserver.
PyVista will likely segfault when rendering.

Try starting a virtual frame buffer with xvfb, or using
  ``pyvista.start_xvfb()``

  warnings.warn(
```

Simply run the following on a headless system in a jupyter notebook:
```
import numpy as np
import pyvista as pv

p = pv.Plotter()
p.add_mesh(pv.PlatonicSolid("cube", radius=0.4))
p.show(jupyter_backend="pythreejs")
```
Sometimes the kernel simply dies, but most of the time this error is printed:
```
This system does not appear to be running an xserver.
PyVista will likely segfault when rendering.

Try starting a virtual frame buffer with xvfb, or using
  ``pyvista.start_xvfb()``

  warnings.warn(
```
Running this instead will not error, but no output is shown:
```
import numpy as np
import pyvista as pv
pv.set_jupyter_backend("pythreejs")
p = pv.Plotter()
p.add_mesh(pv.PlatonicSolid("cube", radius=0.4))
p.show()
```

### System Information

```shell
  Date: Sat Aug 27 17:07:53 2022 UTC

                OS : Linux
            CPU(s) : 64
           Machine : x86_64
      Architecture : 64bit
               RAM : 251.8 GiB
       Environment : Jupyter
       File system : ext4
       GPU Details : error

  Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:56:21)
  [GCC 10.3.0]

           pyvista : 0.36.1
               vtk : 9.1.0
             numpy : 1.22.4
           imageio : 2.21.1
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.4.3
             PyQt5 : 5.15.7
           IPython : 8.4.0
             scipy : 1.9.0
         pythreejs : Version unknown
```

### Screenshots

_No response_

### LINKED ISSUE #3274: `pyvista` (`pythreejs` or `ipygany` backend) incompatible with `ipywidgets>7.7.2.`

The jupyter backend `pythreejs` does not work with the newest release if `ipywidgets`. No plot is produced

```docker
FROM ubuntu:22.04


WORKDIR /tmp/

# pyvista dependencies from apt
RUN apt-get -qq update && \
    apt-get -y install libxrender-dev libgl1-mesa-dev xvfb python3-pip && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/* /var/tmp/*

# Install python dependencies
RUN pip3 install --upgrade pip 


# Install components from dolfinx/lab
RUN pip3 install pyvista pythreejs ipygany pyvista jupyterlab && \
    pip3 cache purge && \
    jupyter nbextension enable --py --sys-prefix ipygany


EXPOSE 8888/tcp

ENV SHELL /bin/bash
ENTRYPOINT ["jupyter", "lab", "--ip", "0.0.0.0", "--no-browser", "--allow-root"]

```

```python
import pyvista as pv

sphere = pv.Sphere(center=(1, 0, 0))
cube = pv.Cube()
pv.set_jupyter_backend("pythreejs")
pl = pv.Plotter()
pl.add_mesh(sphere, color='r')
pl.add_mesh(cube, color='b')
pl.show()
```
Downgrading `ipywidgets` resolves the issue.

### System Information

```shell
  Date: Tue Sep 06 11:23:32 2022 UTC

                OS : Linux
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 15.5 GiB
       Environment : Jupyter
       GPU Details : error

  Python 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]

           pyvista : 0.36.1
               vtk : 9.2.0
             numpy : 1.23.2
           imageio : 2.21.2
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.3
           IPython : 8.5.0
        jupyterlab : 3.4.6
         pythreejs : Version unknown
```

### Screenshots

ipywidgets

### LINKED ISSUE #3482: Coincident topology issues with pythreejs jupyter_backend

Hello,

I have an issue when visualising thin shells. When looking at the normals in the screen recording attached, the values seem incorrect, but the problem disappears when zooming in. Have you encountered this before? Is there a way to fix this?

Try to visualize a mesh with a thin shell.

### System Information

```shell
  Date: Mon Oct 17 14:57:25 2022 UTC

                OS : Linux
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 54.9 GiB
       Environment : Jupyter
        GPU Vendor : Mesa/X.org
      GPU Renderer : llvmpipe (LLVM 12.0.0, 256 bits)
       GPU Version : 4.5 (Core Profile) Mesa 21.2.6

  Python 3.8.10 (default, Mar 15 2022, 12:22:08)  [GCC 9.4.0]

           pyvista : 0.33.2
               vtk : 9.0.1
             numpy : 1.19.5
           imageio : 2.9.0
           appdirs : 1.4.4
            scooby : 0.5.7
        matplotlib : 3.3.4
           IPython : 7.16.1
        ipyvtklink : 0.2.2
             scipy : 1.8.1
              tqdm : 4.60.0
            meshio : 4.4.1
```

### Screenshots

https://user-images.githubusercontent.com/37087741/196212363-5df2c0a6-f7ee-4a28-9976-81cee8e9e692.mov

### LINKED ISSUE #3771: Scalar bar color cannot be changed in trame local view

@banesullivan I'm using branch _feat/trame_ and I found that if I'm using PyVistaLocalView, the color of the scalar bar cannot be changed. I set it to white, but it's still black.
But for PyvistaRemoteView, the _scalar bar arg color_ is working. I think the problem is that the scalar bar widget for the local view is different from the remote view.

Here is the related code.
`plotter.add_mesh(mesh, name='model', scalar_bar_args={'color':'white'})`
It's working for remote view.

### System Information

```shell
OS : Windows
            CPU(s) : 16
           Machine : AMD64
      Architecture : 64bit
       Environment : Python
       GPU Details : error

  Python 3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit
  (AMD64)]

           pyvista : 0.38.dev0
               vtk : 9.1.0
             numpy : 1.23.3
           imageio : 2.22.0
            scooby : 0.5.12
             pooch : v1.6.0
        matplotlib : 3.6.0
             trame : Version unknown
         trame_vtk : Version unknown
```

### Screenshots

Local View - the color of the scalar bar is always black.
image

Remote View - the color of the scalar bar can be controlled.
image

banesullivan:
This adds streamlined support for using PyVista with [Trame](https://kitware.github.io/trame/index.html) and adds a new Trame-based Jupyter backend.

The new Trame-based backend is the backend that we will try to centralize around per #3690 and with it's addition comes an immediate deprecation of the `ipyvtklink` backend.

[Trame](https://kitware.github.io/trame/index.html) is a new framework for building lightweight, reactive web applications in a stateful manner. It has great support for VTK and thus PyVista. These changes add a `trame` module to PyVista that streamlines usage of the `trame-vtk` interface for use with a PyVista `Plotter` object.

While the bulk of this PR is actually focused on the new Jupyter backend, this is also provided streamlined support for using PyVista and Trame to make reactive web apps. There are a few examples web apps included under: `examples_trame/`

#### Basic Trame Web App

https://user-images.githubusercontent.com/22067021/193112815-16df2473-882e-4f35-b479-3b113489031d.mov 

#### Advanced Trame Web App

https://user-images.githubusercontent.com/22067021/213883318-e2bb2918-1fb3-48aa-a66c-adc747d9141a.mov 

####  New Jupyter Backend
https://user-images.githubusercontent.com/22067021/213883433-47edfe99-3e36-4580-ac4c-9b9fa1d7a434.mov

## Jupyter Backend

See usage instructions in docs but simply

```py
import pyvista as pv
from pyvista import examples

await pv.set_jupyter_backend('trame')

mesh = examples.load_random_hills()
arrows = mesh.glyph(scale="Normals", orient="Normals", tolerance=0.05)

p = pv.Plotter()
p.add_mesh(mesh, scalars="Elevation", cmap="terrain", smooth_shading=True)
p.add_mesh(arrows, color="black")
p.show()
```

Screen Shot 2023-01-21 at 12 24 35 PM

## Additional Changes

This PR required a significant amount of work to PyVista and sneaks in a few changes that really could've been their own PRs. These include:

- An an internal refactor of `ren_win` to `render_window` as a `@property`
- The addition of a new `suppress_rendering` `@property` on the `BasePlotter` that will suppress calls to `vtkRenderWindow.Render()` for guarantee use of the PyVista plotter without a virtual frame buffer
- A refactor/change to the `_id_name` attribute of the plotter so that it is a safe string to use as a variable name (added a `P_` prefix)
- Addition of an `actors` `@property` on `BasePlotter` to get all actors on the Plotter
- A new mechanism for adding on-render callbacks. See `add_on_render_callback`. This has two modes (to register with actual VTK render events or be called explicitly by PyVista's `BasePlotter.render()` method (often to be paired with `suppress_rendering`)
- Removal of the already deprecated `use_ipyvtk` argument and `PYVISTA_USE_IPYVTK` environment variable

## Resolves and Closes

This PR renders many issue non-actionable as the new Trame backend handles many issue with the other existing backends

- Closes #3274: since ipyvtklink is deprecated, this is no longer relevant or actionable
- Closes #3091: use the new trame backend for grid lines
- Resolves #3221: `suppress_rendering` should fix this when combined with the `pythreejs` backend
- Closes #2940: this new Trame backend is the only backend that supports dynamic updating
- Closes #3482: Recommend using the new trame backend to resolve coincident topology issues
- Resolves #3771

We will continue to address the remaining issues listed in #3690

## Upstream work

A lot of upstream work was necessary for this new Jupyter backend and much thanks to @jourdain! Here are some (not all) of the relevant work:

- https://github.com/conda-forge/vtk-feedstock/pull/258
- https://github.com/Kitware/trame-vtk/pull/1
- https://github.com/Kitware/trame-vtk/pull/4
- https://github.com/Kitware/trame-vtk/pull/5
- https://github.com/Kitware/trame-vtk/commit/fb94e81ef86c152f207e4fd442747c0d318a8dde
- https://github.com/Kitware/trame-vtk/commit/85ee285f67cf08438d37bae0bfd8d84ffe34db35
- https://github.com/Kitware/trame-vtk/commit/a9e4513e43fea443ee02d2de002270d666515913
- https://github.com/Kitware/trame-vtk/commit/7278d5ed7b8872167a9e9c653792b1b8543ac5ab
- https://github.com/Kitware/trame-vtk/pull/10
- https://github.com/Kitware/trame-vtk/pull/13
- https://github.com/Kitware/trame-server/commit/6b72322c06a5ade065355e2ffbd5d74206a89464
- https://github.com/Kitware/trame-server/pull/8
- https://github.com/Kitware/trame-client/commit/3cad0237582f46fe0f46e726b5fa1d9b71812f6b
- https://github.com/Kitware/trame/commit/b48033261f364fd79f0dcd039b4c04c153eb67b1
- https://github.com/Kitware/trame-client/commit/45a3b2040bead26da83cde8b5bb51cd8423f59ff

### Future Work

- Screenshot button for client-side rendering (currently only works for server-side rendering)
- Support for picking
- Volume rendering support on the client-side viewer (ref https://github.com/Kitware/trame-vtk/issues/14)


pyvista/trame/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+        if not self._plotter.camera_set:
+            self._plotter.view_isometric()
+
+
+class PyVistaRemoteView(VtkRemoteView, _BasePyVistaView):
+    """PyVista wrapping of trame VtkRemoteView."""
+
+    def __init__(self, plotter, interactive_ratio=1, ref='view', **kwargs):
+        """Create a trame remote view from a PyVista Plotter."""
+        _BasePyVistaView.__init__(self, plotter)
```
adeak, line 24: Since multiple inheritance seems like a given, it might be prudent to use `super()` in these examples to show best practices. This might include calling `super().__init__(self)` in `_BasePyvistaView.__init__()`. (Of course the lack of kwargs already decreases flexibility in cooperation, but still.)

banesullivan: yeah... I'm always not sure how best to handle multi-inheritence like this and didn't see a clean/clear way to do this with `super()`. I'll come back to this with a fresh mind soon here and try to clean it up

banesullivan: I don't think there is a way to do this with `super()`

pyvista/plotting/plotting.py
```diff
@@ -5798,6 +5803,10 @@ def on_timer(iren, event_id):
         self._initialized = True
         log.debug('Plotter init stop')
 
+    @property
+    def render_window(self):
+        return self.ren_win
```
banesullivan, line 5808: this probably needs some safety catches, testing, and documentation.

We use `ren_win` and `hassattr(self, 'ren_win') all over the place and should try to clean all of that up into this fully name propery

banesullivan:
MNE integration test is failing as:
```
  File "/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/pyvistaqt/plotting.py", line 309, in _render
    return BasePlotter.render(self, *args, **kwargs)
  File "/home/runner/work/pyvista/pyvista/pyvista/plotting/plotting.py", line 1606, in render
    if self.render_window is not None and not self._first_time:
  File "/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/pyvistaqt/rwi.py", line 4[26](https://github.com/pyvista/pyvista/actions/runs/3246255962/jobs/5324798559#step:13:27), in __getattr__
    raise AttributeError(self.__class__.__name__ +
AttributeError: BackgroundPlotter has no attribute named render_window. Did you mean: '_RenderWindow'?
Fatal Python error: Aborted
```

Trying to decide if I should implement the `render_window` property in `BackgroundPlotter` or if I should just tag this onto `BasePlotter`....

Jimmy-KL:
@banesullivan I tried this trame feature in pyvista and found a strange thing. I have a model with several data arrays, I want to want to change the data array and the data range of the model and scalar bar can be updated. First I remove the old scalar bar, then I add a new one. Here is the code for this part:
```
@state.change("data_array")
mesh = pv.read(os.path.join(file_path, file_anme))
mesh.set_active_scalars(mesh.array_names[0])
plotter = pv.Plotter(off_screen=True)
actor = plotter.add_mesh(mesh)
plotter.set_background('lightgrey')

def update_data_array(data_array=mesh.array_names[0], **kwargs):
    mesh.set_active_scalars(data_array)
    plotter.remove_scalar_bar()
    plotter.add_scalar_bar(title=data_array)
    min_value = mesh[data_array].min()
    max_value = mesh[data_array].max()
    plotter.update_scalar_bar_range([min_value, max_value])
    ctrl.view_update()

with SinglePageLayout(server) as layout:
    with layout.toolbar:
            vuetify.VSpacer()
            vuetify.VSelect(
                label="属性选择",
                v_model=("data_array", mesh.array_names[0]),
                items=("array_list", mesh.array_names),
                hide_details=True,
                dense=True,
                outlined=True,
                classes="pt-1 ml-2",
                style="max-width: 150px",
            )
    with layout.content:
            with vuetify.VContainer(
                fluid=True,
                classes="pa-0 fill-height",
            ):
                view = PyVistaLocalView(plotter)
                ctrl.view_update = view.update
                ctrl.view_reset_camera = view.reset_camera
```
This can work for a simple pyvista application and also for the remote view of trame. But if I use the local view, the old scalar bar cannot be removed unless I refresh the page.
The initial state:
image
After change the data array:
image
Refresh the page and everything is OK:
image
It maybe a problem of trame or I didn't use it properly. But I think I can give some feedback here. Or do you have a solution for this problem?

jourdain:
This is most likely a little issue with the local view. I would think that if you update `trame-vtk` it may partially fix that problem. I'm still in the process on properly fixing that.

Jimmy-KL:
> This is most likely a little issue with the local view. I would think that if you update `trame-vtk` it may partially fix that problem. I'm still in the process on properly fixing that.

I was using `trame-vtk == 2.0.8`. I updated to the latest version 2.0.9, but the model disappeared, and nothing showed. However, for remote view, it's OK.

jourdain:
`trame-vtk == 2.0.10` may fix all the issue you were seeing.

Jimmy-KL:
@banesullivan Is the pyvista remote view for trame can support headless mode? If it is possible, how to do it? Thanks!

banesullivan:
> Is the pyvista remote view for trame can support headless mode? If it is possible, how to do it? Thanks!

Yes! Totally possible. You'll configure it the same way you would for running PyVista in headless mode in any other scenario. Check out this page of the documentation for tips: https://docs.pyvista.org/getting-started/installation.html#running-on-ci-services

Jimmy-KL:
@banesullivan Maybe we can add `PyvistaRemoteLocalView` for pyvista support for trame? It's similar to [VtkRemoteLocalView](https://trame.readthedocs.io/en/latest/trame.widgets.vtk.html#trame.widgets.vtk.VtkRemoteLocalView). I think it's useful if we want to switch between local and remote view dynamically.

banesullivan:
> @banesullivan Maybe we can add `PyvistaRemoteLocalView` for pyvista support for trame? It's similar to [VtkRemoteLocalView](https://trame.readthedocs.io/en/latest/trame.widgets.vtk.html#trame.widgets.vtk.VtkRemoteLocalView). I think it's useful if we want to switch between local and remote view dynamically.

@Jimmy-KL, indeed it is very similar -- infact, it is almost identical, just has a little bit of logic to grab the vtkRenderWindow from PyVista for that same view. @jourdain and I think it is most appropriate to include this class directly in PyVista rather than trame. This way, we as PyVista devs, can maintain the view ourselves. Afterall, the location of this class isn't all that important and the only thing that would change is how the class is imported.

pyvista/trame/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+from pyvista import _vtk
+
+import weakref
+from trame.widgets.vtk import VtkLocalView, VtkRemoteView, VtkRemoteLocalView
+
+CLOSED_PLOTTER_ERROR = "The render window for this plotter has been destroyed. Do not call `show()` for the plotter before passing to trame."
+
+
+class _BasePyVistaView:
```
adeak, line 14: I don't think we should add so much logic to an `__init__.py`. I suspect this follows `jupyter/__init__.py`. I want to strip these down and move the main logic into dedicated modules, and only keep the bare minimum (mostly imports) in `__init__.py`.

banesullivan: Will do -- I will adjust this trame module here since you're cleaning up the imports in another PR. Should I use `__all__`?

adeak: That would be nice, yes, thanks. And it might be worth using `_trame.py` to avoid name clashes, pending discussion on the star import PR.

banesullivan: Take a look at what I have now. `pyvista.trame` needs to be a publically accessible module so I really don't want to use `pyvista._trame`. Unless you were thinking there'd be an `trame.py` file in that module? In which case, yeah that should be avoided

adeak: Yeah, sorry, I confused myself a bit earlier. What you have now is right, thanks.

banesullivan:
@pyvista/developers, this is ready for review! The history is a bit ridiculous/long, so apologies for that, but a lot has gone into this. Testing coverage isn't ideal here, as web apps like this are somewhat tough to test...

doc/getting-started/installation.rst
```diff
@@ -34,7 +34,7 @@ PyVista includes several optional dependencies for visualization and reading a v
 
 * `cmocean <https://pypi.org/project/cmocean/>`_ - Colormaps for Oceanography.
 * `colorcet <https://colorcet.holoviz.org/>`_ - Perceptually accurate 256-color colormaps for use with Python.
-* `ipyvtklink <https://github.com/Kitware/ipyvtklink>`_ - Minimalist ipywidget to interface with any Python vtkRenderWindow.
+* `trame <https://github.com/Kitware/trame>`_ - Used for client- and server-side rendering in Jupyter.
```
akaszynski, line 37:
```suggestion
* `trame <https://github.com/Kitware/trame>`_ - Used for client and server-side rendering in Jupyter.
```

doc/getting-started/installation.rst
```diff
@@ -103,6 +103,8 @@ The following are a list of optional dependencies and their purpose:
 +===================================+=========================================+
 | ``matplotlib``                    | Using Colormaps                         |
 +-----------------------------------+-----------------------------------------+
+| ``trame``                         | Interactive notebook rendering          |
++-----------------------------------+-----------------------------------------+
 | ``ipyvtklink``                    | Interactive notebook rendering          |
 +-----------------------------------+-----------------------------------------+
```
akaszynski, line 109:
```suggestion
```

doc/user-guide/jupyter/index.rst
```diff
@@ -62,8 +66,8 @@ Supported Modules
 The PyVista module supports a variety of backends when plotting within
 a jupyter notebook:
 
-* Server-side rendering with PyVista streaming to the notebook through
-  `ipyvtklink <https://github.com/Kitware/ipyvtklink/>`_
+* Server- and client-side rendering with PyVista streaming to the notebook through
```
akaszynski, line 69:
```suggestion
* Server and client-side rendering with PyVista streaming to the notebook through
```

doc/user-guide/jupyter/trame.rst
```diff
@@ -0,0 +1,171 @@
...
+
+For convenience, you can enable ``trame`` by default with:
+
+.. note::
+    It is critical to ``await`` the call to :func:`set_jupyter_backend() <pyvista.set_jupyter_backend>` when using Trame in Jupyter.
+
+.. code:: python
+
+    import pyvista as pv
+    await pv.set_jupyter_backend('trame')
```
akaszynski, line 40:
```suggestion
.. code:: python

    import pyvista as pv
    await pv.set_jupyter_backend('trame')

.. note::
    It is critical to ``await`` the call to :func:`set_jupyter_backend() <pyvista.set_jupyter_backend>` when using Trame in Jupyter.
```

akaszynski:
Unable to get this started locally:
```
/tmp/ipykernel_22823/3568984180.py in <module>
----> 1 await pyvista.set_jupyter_backend('trame')

~/source/pyvista/pyvista/jupyter/__init__.py in set_jupyter_backend(backend)
    196 
    197         # Returns a future that finalizes when server is ready - should be awaited
--> 198         return launch_server(pyvista.global_theme.trame.jupyter_server_name)
    199     # Return an awaitable future to prevent errors when accidentally awaiting
    200     # other backends

~/source/pyvista/pyvista/trame/jupyter.py in launch_server(server)
     89         )
     90     # else, server is already running
---> 91     return server.ready
     92 
     93 

AttributeError: 'Server' object has no attribute 'ready'
```

After reinstalling with `pip install -e .[jupyter]`, the error went away. We might consider adding a runtime check for the minimum acceptable version of `trame-*` libraries.

Here's the output from `pip`:
```
  Attempting uninstall: trame-client
    Found existing installation: trame-client 2.4.0
    Uninstalling trame-client-2.4.0:
      Successfully uninstalled trame-client-2.4.0
  Attempting uninstall: trame-server
    Found existing installation: trame-server 2.6.1
    Uninstalling trame-server-2.6.1:
      Successfully uninstalled trame-server-2.6.1
  Attempting uninstall: pyvista
    Found existing installation: pyvista 0.38.dev0
    Uninstalling pyvista-0.38.dev0:
      Successfully uninstalled pyvista-0.38.dev0
Successfully installed jupyter-server-proxy-3.2.2 pyvista-0.38.dev0 simpervisor-0.4 trame-client-2.4.2 trame-server-2.8.0
```

Once I got it working I'm quite impressed with its performance (server and client side), but had a few erroneous errors:
```py
import pyvista as pv

await pv.set_jupyter_backend('trame')
sphere = pv.Sphere()
# short example
sphere.plot()
```

```
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkActor2D with id 000000000292e1e0
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkAxesActor with id 00000000028b7540
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkActor2D with id 000000000292e1e0
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkAxesActor with id 00000000028b7540
```

Looks like it's attempting to serialize the scene twice, I'm surprised it's trying to serialize at all since the default appears to be server side rendering when setting the backend to `'trame'``.

Also, one note regarding the naming: do we want to call it just "client" and "server" or should we go with "trame-client" and "trame-server"?

Also, current behavior with `await` is non ideal. I'd really like for this to automagically be started on the first plot in the background. I'd be annoyed as a first time user with:
```
import pyvista as pv

# await pv.set_jupyter_backend('trame')
sphere = pv.Sphere()
sphere.plot()
```

```
TrameJupyterServerDownError: Trame server has not launched.
```

This will be even more noticeable when `pyvista` is used as a dependency.

examples_trame/basic/file_viewer.py
```diff
@@ -0,0 +1,87 @@
...
+        pl.clear_actors()
+        return
+
+    if files and len(files):
+        if not files[0].get("content"):
+            return
+
+        for file in files:
+            print(f'Load {file.get("name")}')
+            bytes = file.get("content")
```
jourdain, line 41: that may not work for large files (multi-piece vs single), you should use the helper class [ClientFile](https://github.com/Kitware/trame/blob/master/examples/validation/core/27_file_upload.py).

doc/getting-started/installation.rst
akaszynski: Assuming we're removing this....

pyvista/jupyter/__init__.py
```diff
@@ -102,11 +128,33 @@ def set_jupyter_backend(backend):
           framebuffer be set up when displaying on a headless server,
           but does not require any additional modules to be installed.
 
+        * ``'client'`` : Export/serialize the scene graph to be rendered
+          with VTK.js client-side through ``trame``. Requires that a virtual
+          framebuffer be set up when displaying on a headless server and
+          requires ``trame`` and ``jupyter-server-proxy`` to be installed.
```
jourdain, line 134: No framebuffer needed for `client`

pyvista/jupyter/__init__.py
```diff
@@ -127,14 +175,29 @@ def set_jupyter_backend(backend):
 
     >>> pv.set_jupyter_backend('ipyvtklink')  # doctest:+SKIP
 
+    Enable the trame Trame backend.
+
+    >>> pv.set_jupyter_backend('trame')  # doctest:+SKIP
```
jourdain, line 180: do you want to put `await`?

doc/user-guide/jupyter/index.rst
```diff
@@ -111,26 +116,29 @@ your deployment environment.
...
+| trame         | Yes          | Client & Server    | vtk.js & vtk  | Optional             |
++---------------+--------------+--------------------+---------------+----------------------+
+| panel         | Yes          | Client             | vtk.js        | Yes                  |
++---------------+--------------+--------------------+---------------+----------------------+
+| pythreejs     | Yes          | Client             | threejs       | No                   |
++---------------+--------------+--------------------+---------------+----------------------+
+| ipygany       | Yes          | Client             | threejs       | No                   |
++---------------+--------------+--------------------+---------------+----------------------+
+| ipyvtklink    | Yes          | Server             | vtk           | Yes                  |
++---------------+--------------+--------------------+---------------+----------------------+
```
akaszynski, line 133:
```suggestion
+---------------+--------------------+---------------+----------------------+
| Jupyter Notebook 3D Modules                                               |
+---------------+--------------------+---------------+----------------------+
|               | Rendering Location | Backend       | Requires Framebuffer |
+---------------+--------------------+---------------+----------------------+
| trame         | Client & Server    | vtk.js & vtk  | Optional             |
+---------------+--------------------+---------------+----------------------+
| panel         | Client             | vtk.js        | Yes                  |
+---------------+--------------------+---------------+----------------------+
| pythreejs     | Client             | threejs       | No                   |
+---------------+--------------------+---------------+----------------------+
| ipygany       | Client             | threejs       | No                   |
+---------------+--------------------+---------------+----------------------+
| ipyvtklink    | Server             | vtk           | Yes                  |
+---------------+--------------------+---------------+----------------------+
```

doc/user-guide/jupyter/trame.rst
```diff
@@ -0,0 +1,171 @@
+.. _trame_jupyter:
+
+Trame Jupyter backend for PyVista
+---------------------------------
+
+PyVista has the ability to display fully featured plots within a
+Jupyter environment using `Trame <https://kitware.github.io/trame/index.html>`_.
+We provide mechanisms to pair PyVista and Trame so that PyVista plotters
+can be used in a web context with both server- and client-side rendering.
```
akaszynski, line 9:
```suggestion
can be used in a web context with both server and client-side rendering.
```

pyvista/trame/jupyter.py
```diff
@@ -0,0 +1,190 @@
...
+        warnings.warn(JUPYTER_SERVER_DOWN_MESSAGE)
+        # and Error
+        super().__init__(JUPYTER_SERVER_DOWN_MESSAGE)
+
+
+def is_server_up(server):
+    """Check if server is running."""
+    if isinstance(server, str):
+        server = get_server(server)
+    return server._running_stage >= 2
```
jourdain, line 64: server already have a property `running`. https://github.com/Kitware/trame-server/blob/master/trame_server/core.py#L332

examples_trame/basic/PyVistaLocalView.py
```diff
@@ -0,0 +1,56 @@
+"""How to use PyVista's ``PyVistaLocalView`` trame view component.
+
+This is a full-fledged example on building your own user interfac
```
akaszynski, line 3:
```suggestion
This is a full-fledged example on building your own user interface
```

pyvista/trame/ui.py
```diff
@@ -0,0 +1,337 @@
...
+    'server',
+]
+
+
+def vuwrap(func):
+    """Call view_update in trame to synchronize changes to a view."""
+
+    def wrapper(self, *args, **kwargs):
+        ret = func(self, *args, **kwargs)
+        self._ctrl.view_update()
```
jourdain, line 31: All the viewers instance are sharing the same controller names?

examples_trame/basic/UITemplate.py
```diff
@@ -0,0 +1,71 @@
+"""How to use PyVista UI template.
+
+This example demonstrates how to use ``ui_container`` to create
+add a PyVista ``Plotter`` to a UI with scene controls and standard
+UI features.
```
akaszynski, line 5:
```suggestion
This example demonstrates how to use `ui_container`` to add a PyVista
``Plotter`` to a UI with scene controls and standard UI features.
```

pyvista/trame/ui.py
```diff
@@ -0,0 +1,337 @@
...
+                    off_icon=icons[1],
+                    dense=True,
+                    hide_details=True,
+                    classes='my-0 py-0 ml-1',
+                )
+        html.Span(tooltip)
+
+
+def ui_container(
+    server, plotter, mode=None, default_server_rendering=True, collapse_menu=False, **kwargs
```
jourdain, line 180: make `trame_server=None` as kwarg and retrieve it from parent element if None.

pyvista/trame/ui.py
```diff
@@ -0,0 +1,337 @@
...
+                **kwargs,
+            )
+            ctrl.view_update_image = view.update_image
+        elif mode == 'server':
+            view = PyVistaRemoteView(plotter, **kwargs)
+        elif mode == 'client':
+            view = PyVistaLocalView(plotter, **kwargs)
+        ctrl.view_update = view.update
+        ctrl.view_reset_camera = view.reset_camera
+        ctrl.view_push_camera = view.push_camera
```
jourdain, line 317: the last instance of the viewer will override the other ctrl. methods...

jourdain (commented):
Looks good but we can improve the `ui_component` element and make sure the viewer don't override methods on the controller across instances...

pyvista/trame/jupyter.py
```diff
@@ -0,0 +1,190 @@
...
+    collapse_menu=False,
+    default_server_rendering=True,
+    **kwargs,
+):
+    """Run and display the trame application in jupyter's event loop.
+
+    plotter : pyvista.BasePlotter
+        The PyVista plotter to show.
+
+    mode : str, default: 'trame'
```
akaszynski, line 132:
```suggestion
    mode : str, optional
```

pyvista/trame/jupyter.py
```diff
@@ -0,0 +1,190 @@
...
+        The UI view mode. This can be set on the global theme. Options are:
+            * ``'trame'``: Uses a view that can switch between client and server
+              rendering modes.
+            * ``'server'``: Uses a view that is purely server rendering.
+            * ``'client'``: Uses a view that is purely client rendering (generally
+              safe without a virtual frame buffer)
+
+    name : str
```
akaszynski, line 140:
```suggestion
    name : str, optional
```

pyvista/trame/jupyter.py
```diff
@@ -0,0 +1,190 @@
...
+        ``jupyter-server-proxy``, often set to ``proxy/``.
+
+    collapse_menu : bool, default: False
+        Collapse the UI menu (camera controls, etc.) on start.
+
+    default_server_rendering : bool, default: True
+        Whether to use server-side or client-side rendering on-start when
+        using the ``'trame'`` mode.
+
+    **kwargs
```
akaszynski, line 158:
```suggestion
    **kwargs : dict, optional
```

doc/user-guide/jupyter/trame.rst
akaszynski: "Notes" seem to make more sense after calling something to explain something than beforehand.

pyvista/trame/ui.py
banesullivan: Oh woah. I'm surprised I didn't notice this...

pyvista/jupyter/__init__.py
banesullivan:
```suggestion
    >>> await pv.set_jupyter_backend('trame')  # doctest:+SKIP
```

examples_trame/basic/multi_views.py
```diff
@@ -0,0 +1,60 @@
+from trame.app import get_server
```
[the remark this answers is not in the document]
banesullivan: Can you elaborate and share screenshots? This is generally working for me

pyvista/trame/ui.py
jourdain: we can chat next week so I can tell you how to better handle it...

banesullivan:
@akaszynski 

> We might consider adding a runtime check for the minimum acceptable version of trame-* libraries.

We could... I suspect this pain point is only going to be for folks that already have trame installed as we will have good support for all future versions after the minimums set in `pyproject.toml`. See also: https://github.com/Kitware/trame/issues/183

> Looks like it's attempting to serialize the scene twice

I'll look into this appearance of serializing twice. 

>  I'm surprised it's trying to serialize at all since the default appears to be server side rendering when setting the backend to `'trame'``.

Well the `'trame'` mode simultaneously uses server- and client-side rendering. So this is expected. Both views are present on the front-end, so the scene will be serialized to the client-side view on each update call. The `'server'` mode uses pure-server-side rendering with no scene serialization.


> Also, one note regarding the naming: do we want to call it just "client" and "server" or should we go with "trame-client" and "trame-server"?

I'm open to suggestions here... I chose these three names:

* ``'trame'``: Uses a view that can switch between client- and server-rendering modes (uses both simultaneously)
* ``'server'``: Uses a view that is purely server-rendering.
* ``'client'``: Uses a view that is purely client-rendering (generally safe without a virtual frame buffer)

We could also support a `trame-` prefix to `server` and `client` as well?

> Also, current behavior with await is non ideal. I'd really like for this to automagically be started on the first plot in the background. I'd be annoyed as a first time user with:

Indeed, not ideal... but we have to `await` the server launch in Jupyter otherwise, the cells will execute before the server is ready and able to produce an output. We can chat offline about this for better clarification. FWIW, this would be resolved by adding a Jupyter Comms "backend" to `wslink` which is on the roadmap

jourdain:
> FWIW, this would be resolved by adding a Jupyter Comms "backend" to wslink which is on the roadmap

Not true. You will still have to wait for the server task to be scheduled.

banesullivan:
> Not true. You will still have to wait for the server task to be scheduled.

Ah, my bad - yes, the server launch would still need to be awaited... it'd simply be the communication mechanisms that changes if using Jupyter Comms. When I wrote that above, I was thinking it would use the Jupyter web server... which would not be the case... but maybe that's a route we could look into?

jourdain:
Either way (now and future) we are using the Jupyter web server. Using the Jupyter Comms has no impact at all.
This is a pure Python async task scheduling problem. So nothing will ever change on that topic. You will always need to know when a task has been schedule.

banesullivan:
@pyvista/developers, this is ready to go! Hoping to merge in ~24ish hours and would greatly appreciate any last minute feedback

akaszynski (dismissed):
Really pleased with this PR. `jupyterlab` backends have been a mess for a while.

I'm fine with the naming. We're going to discover issues moving forward, but we're going to fix this rapidly.

akaszynski (dismissed):

pyvista/jupyter/__init__.py
```diff
@@ -44,6 +58,10 @@ def _validate_jupyter_backend(backend):
             raise ImportError('Please install `pythreejs` to use this feature.')
 
     if backend == 'ipyvtklink':
+        warnings.warn(
+            '`ipyvtklink` backend is deprecated and has been replaced by the `trame` backend.',
+            PyVistaDeprecationWarning,
```
larsoner, line 63:
```suggestion
            PyVistaDeprecationWarning, stacklevel=3,
```
Might not always be correct (?) but at least for MNE this gets me:
```
/Users/larsoner/python/mne-python/mne/viz/backends/_notebook.py:1458: PyVistaDeprecationWarning: `ipyvtklink` backend is deprecated and has been replaced by the `trame` backend.
  viewer = self.plotter.show(
```
instead of the less informative
```
/Users/larsoner/python/pyvista/pyvista/jupyter/notebook.py:55: PyVistaDeprecationWarning: `ipyvtklink` backend is deprecated and has been replaced by the `trame` backend.
  warnings.warn(
```

pyvista/trame/jupyter.py
```diff
@@ -0,0 +1,222 @@
...
+        # server_proxy_prefix assumes trailing slash
+        src = (
+            f"{server_proxy_prefix if server_proxy_prefix else ''}{_server.port}/index.html{params}"
+        )
+    else:
+        src = f'{kwargs.get("protocol", "http")}://{kwargs.get("host", "localhost")}:{_server.port}/index.html{params}'
+    logger.debug(src)
+    width = kwargs.get('width', '99%')
+    height = kwargs.get('height', '600px')
+    return widgets.HTML(f"<iframe src='{src}' style='width: {width}; height: {height};'></iframe>")
```
larsoner, line 108: If I'm following the code correctly, it seems like there is no longer an easy equivalent of `viewer = plotter.show(..., return_viewer=True); viewer.update_image()`. The function including this line is what gets called under the hood and becomes `viewer` in my code snippet (right?) so it now returns a `widgets.HTML` element, which does not seem to have any options for connecting to the server. Maybe it would be better to return a custom object with a `.server` attribute instead, along with a `_repr_html_` that is the iframe source? Then in MNE where we currently call `viewer.update_image()` to say "rerender/update", we could do `viewer.server.<something>`.

larsoner (commented):
It looks like `trame` is not on conda-forge:

https://github.com/conda-forge/trame-feedstock

To me this should be a blocker for deprecating `ipyvtklink` :(

But using this minimal (for MNE anyway!) code:
```
import mne
mne.viz.set_3d_backend('notebook')
data_path = mne.datasets.sample.data_path()
subjects_dir = data_path / 'subjects'
stc = mne.read_source_estimate(data_path / 'MEG/sample/sample_audvis-eeg-lh.stc')
stc.plot(subject='sample', subjects_dir=subjects_dir, initial_time=0.1)
```
When using `'ipyvtklink'` I get:

Screenshot 2023-01-26 at 8 26 07 AM

And when changing the `'ipyvtklink'`s in `mne/viz/backends/_notebook.py` to `'trame'`s on the first run I got a segfault (?), but after that I get something that we could probably fix up at least!

Screenshot 2023-01-26 at 8 36 00 AM

akaszynski:
> To me this should be a blocker for deprecating ipyvtklink

Let's keep `ipyvtklink` around for one (maybe two?) minor releases then. Release schedule is one per three months, this should give users enough time to provide feedback and for Kitware to get trame on conda.

larsoner:
I'm not sure we'll actually need the VTK folks to add support. I'm getting the ball rolling in https://github.com/conda-forge/staged-recipes/pull/21876, let's see if it works...

akaszynski:
Minor issue:

I forgot to install .[jupyter] and would prefer a better error:
```
/home/akaszyns/source/pyvista/pyvista/jupyter/notebook.py:72: UserWarning: Failed to use notebook backend: 

No module named 'vtkmodules.web.errors'

Falling back to a static output.
  warnings.warn(
```

banesullivan:
@larsoner, thanks for jump-starting conda-forge recipes! I'm happy to take this on as well 

> I'm not sure we'll actually need the VTK folks to add support

I want to involve @jourdain and @psavery to make sure we get all of trame on there and handle the dependencies correctly. I feel like we'er going to run into a hurdle with conda and the fact that the core `trame` is a namespace package.

akaszynski:
Minor issue:

I forgot to install .[jupyter] and would prefer a better error:
```
/home/akaszyns/source/pyvista/pyvista/jupyter/notebook.py:72: UserWarning: Failed to use notebook backend: 

No module named 'vtkmodules.web.errors'

Falling back to a static output.
  warnings.warn(
```

akaszynski:
Getting some annoying warnings on Python 3.9
```
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkSmartVolumeMapper with id 000000000377ca00
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkActor2D with id 00000000058160b0
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkAxesActor with id 0000000005765ea0
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkSmartVolumeMapper with id 000000000377ca00
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkActor2D with id 00000000058160b0
ERROR:vtkmodules.web.render_window_serializer:!!!No serializer for vtkAxesActor with id 0000000005765ea0
```

Report:
```
python -c "import pyvista; print(pyvista
.Report())"                                                                         

  Date: Thu Jan 26 07:45:36 2023 MST

                  OS : Linux
              CPU(s) : 6
             Machine : x86_64
        Architecture : 64bit
                 RAM : 62.6 GiB
         Environment : Python
         File system : ext4
        GPU Renderer : Quadro P2000/PCIe/SSE2
         GPU Version : 4.5.0 NVIDIA 470.161.03

  Python 3.9.16 (main, Dec  7 2022, 01:11:51)  [GCC 9.4.0]

             pyvista : 0.38.dev0
                 vtk : 9.2.5
               numpy : 1.23.5
             imageio : 2.23.0
              scooby : 0.7.0
               pooch : v1.6.0
          matplotlib : 3.6.2
             IPython : 7.31.0
            colorcet : 3.0.1
             cmocean : 2.0
          ipyvtklink : 0.2.3
               scipy : 1.10.0
                tqdm : 4.64.1
              meshio : 5.3.4
          jupyterlab : 3.5.2
           pythreejs : Version unknown
      pytest_pyvista : 0.1.3
               trame : Version unknown
        trame_client : Version unknown
        trame_server : Version unknown
           trame_vtk : Version unknown
jupyter_server_proxy : Version unknown
```

Seems to be a scooby issue as well?

akaszynski:
Also, let's add in some sort of warning when you try to plot volumes using `client-trame`. I think it's fine to merge this if the `client` mode doesn't support volumetric rendering, but it should at least have a warning (and a helpful message like switch to server for volumes for now).

larsoner:
> I want to involve @jourdain and @psavery to make sure we get all of trame on there and handle the dependencies correctly. I feel like we'er going to run into a hurdle with conda and the fact that the core trame is a namespace package.

My plan is to add each of the 13 individual `trame-*` packages, then just make `trame` a package that requires the other 13. Adding the first one `import trame` seems to work already, so I think it's on the right track:

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=647911&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1258

```
import: 'trame'
import: 'trame_client'
import: 'trame'
import: 'trame_client'
```

banesullivan:
> Getting some annoying warnings on Python 3.9

Yeah... we can suppress those if desired but they are coming from the VTK serializer.

> Seems to be a scooby issue as well?

Upgrade scooby 😉 ... this was fixed by https://github.com/banesullivan/scooby/pull/101 (see also https://github.com/Kitware/trame/issues/183)

> We also need to add in the docs (somewhere) that you have to return the widget for it to show in jupyter

Oh sheesh. Our other backends call `IPython.display.display`, don't they.... IMO, they should *not* and the widget should be returned and output when at the end of a cell. I see now that `return_viewer=False` by default and we call `IPython.display.display` by default for other backends... Oooof. I'm not huge on this... but I will update this PR to follow the same convention.

In your example, IMO, `plot_orbital` should `return pl.show()` always (same for any demo method that calls `show()` in Jupyter or otherwise) as it could return a camera position, screenshot, etc, depending on the configuration.


> Also, let's add in some sort of warning when you try to plot volumes using client-trame. I think it's fine to merge this if the client mode doesn't support volumetric rendering, but it should at least have a warning (and a helpful message like switch to server for volumes for now).

Actually. The only issue is that the VTK volume serializer does not handle `vtkSmartVolumeMapper`. This should be an easy fix that I can land shortly after this merges. So I'd rather not add a warning. Further, you can use the `'fixed_point'` volume mapper, and all will work in client mode as is.

jourdain:
I don't think adding trame to conda-forge should trigger any hurdle. Thanks @larsoner for taking that task over. Both @psavey and I could be added to those trame packages if you want. 

For the no-serializer issue, a better approach would be to notify the user (in the GUI) that "some content could not be available for local rendering" and then offer the option to switch to remote rendering or ignore.

pyvista/trame/jupyter.py
banesullivan: I fixed this up. I added a new subclass of the `HTML` `ipywidget` that keeps a reference to the `Viewer` instance, which can be used to sync the widget view. Further, I fixed `return_viewer` to work for the `trame` backend the same way as the other backends.

banesullivan:
@larsoner, thanks for all of the feedback! I think I've addressed everything you brought up, but please double-check.

Regarding the `ipyvtklink` deprecation: Hopefully, this is okay now considering the trame conda-forge recipes should be up in no time (thanks to you! 😄 )

I really, really want to merge this today unless something significant comes up -- please let me know what you think.

FYI, I noticed the UI menu bar issue in your screenshot and will try to sneak in a fix for that

banesullivan:
> For the no-serializer issue, a better approach would be to notify the user (in the GUI) that "some content could not be available for local rendering" and then offer the option to switch to remote rendering or ignore.

@jourdain, I feel like that would be best implemented in `trame-vtk`. What do you think?

pyvista/trame/jupyter.py
```diff
@@ -0,0 +1,275 @@
...
+    default_server_rendering : bool, default: True
+        Whether to use server-side or client-side rendering on-start when
+        using the ``'trame'`` mode.
+
+    return_viewer : bool, optional
+        Return the ipywidget.
+
+    **kwargs : dict, optional
+        Mostly ignored, though ``protocol`` and ``host`` can be use to
+        override the iframe src url and ``hieght`` and ``width`` can be
```
larsoner, line 202:
```suggestion
        override the iframe src url and ``height`` and ``width`` can be
```
This also creates an issue. Before you could just set the width and height of the plotter, and the `ipyvtklink` ended up being the same size. So you should probably...

akaszynski:
> Further, you can use the 'fixed_point' volume mapper, and all will work in client mode as is.

Should we just change the default mapper then if it detects that it's in a notebook with either `trame` or `client` enabled? We don't have to fix it, but I also don't like empty plots.

pyvista/trame/jupyter.py
```diff
@@ -0,0 +1,275 @@
...
+    )
+
+    # Show as cell result
+    src = build_url(
+        server,
+        ui=plotter._id_name,
+        server_proxy_enabled=server_proxy_enabled,
+        server_proxy_prefix=server_proxy_prefix,
+        **kwargs,
+    )
```
larsoner, line 236: ... Add this
```suggestion
    )
    for ki, key in enumerate(('width', 'height')):
        if key not in kwargs:
            kwargs[key] = f'{plotter.window_size[ki]}px'
```
Or something similar somewhere. FWIW this code fixes the window sizing for MNE-Python

Screenshot from 2023-01-26 12-33-12

larsoner (approved):
Other than one remaining bug found via my minimal testing, LGTM!

pyvista/trame/jupyter.py
banesullivan: Ahh, nice idea!

banesullivan:
@akaszynski 
>Should we just change the default mapper then if it detects that it's in a notebook with either trame or client enabled? We don't have to fix it, but I also don't like empty plots.

We could... but I'd rather try to just fix this upstream and follow up with this sort of patch if necessary

pyvista/trame/jupyter.py
banesullivan: Ehh... I really want the default to be `100%` rather than matching the plotter size. With these changes it makes the output not so elegant by default:

Screen Shot 2023-01-26 at 11 45 02 AM

Where as with `width: 100%; height: 600px;` its much nicer no matter the window size:

Screen Shot 2023-01-26 at 11 47 51 AM

banesullivan: I reverted this suggestion

larsoner: But didn't it look that way with `ipyvtklink` already, suggesting that this maintains previous behavior?

Okay with me in principle for 100% -- 50% didn't seem to work all that well -- *assuming there is some nice way in the public API to easily specify a fixed size*. What is it with the current code? It seems like the current version completely ignores `Plotter(..., window_size=...)` which seems unexpected/not backward compatible...

One way out of this might be to have `Plotter(..., window_size='auto')` as a backward compat default which means "`(600px, 600px)` for all backends except trame, which will use `('100%', '600px')`" or something.

banesullivan: @larsoner, are you okay with me merging as is and following up in a separate PR (before the release) to expose a better mechanism for controlling the widget size?

larsoner: Okay

banesullivan: Tracking here: https://github.com/pyvista/pyvista/issues/3879

Will try to address ASAP

akaszynski (approved):
Reapproving. Really appreciate everyone's feedback here, and if there are any issues post-merge we'll be quick to fix them in follow-up PRs.

akaszynski:
@banesullivan and @larsoner, let's merge this (after 24 ish hours green) and then follow-up with smaller PRs. I've had my fair share of 2,000+ LOC PRs and I think it will be easier to track issues and fix them in individual PRs.

banesullivan:
> I think it will be easier to track issues and fix them in individual PRs.

Exactly. This PR is touching a bunch of code at this point and it'll be much easier to track and review further changes as new PRs

larsoner (approved):
Sounds good!

jourdain:
It should but it could be a WIP. I know that @banesullivan is working toward streamlining some of those conda recipes.

banesullivan:
The trame recipes just landed on conda-forge a few hours ago.

```
conda install -c conda-forge 'pyvista>=0.38.1' matplotlib ipywidgets trame jupyterlab
```

Should get you everything you need (just confirmed in a fresh env locally)

bilke:
Dear @banesullivan, thanks installation worked so far but I always get a 404 on the trame widget, e.g.:

```
[W 2023-02-07 08:20:40.754 ServerApp] 404 GET /proxy/33025/index.html?ui=P_0x7f8b6409aa10_0&reconnect=auto (3dd9915b860a49639fc672524abd9e7d@141.65.34.101) 0.57ms referer=http://[IP]:8888/lab
```

I tried with and without `PYVISTA_TRAME_SERVER_PROXY_ENABLED="True"`.

banesullivan:
Would you please open a new issue to track this

Issue #3257, opened 2022-09-01
Create pytest fixture for verify_image_cache


akaszynski:
We should really just make a pytest fixture context that sets `verify_cache_image` as the default `before_close_callback`... to do in a later PR though

_Originally posted by @banesullivan in https://github.com/pyvista/pyvista/pull/3248#discussion_r960877967_

PR #3391, opened 2022-09-30
Add other views to `Camera.tight`

MatthewFlamm:
### Overview

This PR allows for filling the render window in views other than `xy`. This brings the selections inline with the `view_xy`, `view_xz`, and so on methods.

### Details

`Camera.zoom` allows a `tight` value which will reorient to the default, `xy` view.  I'm not yet sure if this should be modified to allow these options.  We could deprecate `tight` in this case and switch to `zoom('xy')` and similar.

This PR also refactors the `view_xy` methods to use a common helper.  I also added comprehensive tests that check all combinations.  I added in the ability to modify the image_cache filename according to a parametrized test with a reusable fixture to grab the testname.  It would be great if we could abstract out this logic completely in the future, related to #3257.


adeak:
Just having read the description, this sounds amazing!

pyvista/plotting/_utilities.py
```diff
@@ -0,0 +1,51 @@
...
+    elif view == 'yz':
+        vec = np.array([1, 0, 0])
+        viewup = np.array([0, 0, 1])
+    elif view == 'zy':
+        vec = np.array([-1, 0, 0])
+        viewup = np.array([0, 1, 0])
+    else:
+        raise ValueError(
+            f"Unexpected value for direction {view}\n"
+            "    Expected: 'xy', 'yx', 'xz', 'xz', 'yz', 'zy'"
```
adeak, line 46:
```suggestion
            "    Expected: 'xy', 'yx', 'xz', 'zx', 'yz', 'zy'"
```

MatthewFlamm:
My main concern with deprecating the `Camera.zoom('tight')` usage to be `Camera.zoom('xy')` is that this functionality was just added, so it would be a bit of whiplash.

MatthewFlamm:
I kept the existing functionality to put the focal point at (x, y, 0) and the camera position is set according to a distance that depends on the view angle.  I don't think the view angle in parallel projection mode should matter.  This can also make the camera position be inside the object.  I did some testing, and I couldn't reliably reproduce any problem with this approach, but it seems fishy to me.

pyvista/plotting/camera.py
```diff
@@ -691,22 +699,37 @@ def tight(self, padding=0.0, adjust_render_window=True):
...
+
+        vert_dist = np.abs(np.dot(distance_object, viewup))
+        horiz_dist = np.abs(np.dot(distance_object, horizontal))
+
+        dist = max(horiz_dist / aspect[0], vert_dist) / np.sin(angle / 2) / 2
+
+        focal_point = position.copy()
+        # set focal point to 0 in viewing direction
+        focal_point -= np.abs(direction * focal_point)
+        # offset position from foral point by dist in viewing direction
```
adeak, line 717:
```suggestion
        # offset position from focal point by dist in viewing direction
```

pyvista/plotting/camera.py
```diff
@@ -691,22 +699,37 @@ def tight(self, padding=0.0, adjust_render_window=True):
...
+        position += -1 * np.abs(direction) * position + dist * direction
```
adeak, line 718: Again the logic is a bit unclear to me. `abs(direction)` surprises me, because this will in general mess with your direction. I guess what I'm missing here is that this method is special-cased for vectors where exactly one component of any vector is nonzero, and that nonzero value is +-1.

Judging by the preceding comment I'd expect something like
```py
position = focal_point - dist * direction
```

adeak (commented):
To address your comments:

> `Camera.zoom` allows a `tight` value which will reorient to the default, `xy` view. I'm not yet sure if this should be modified to allow these options. We could deprecate `tight` in this case and switch to `zoom('xy')` and similar.

When I first drove by and saw you mentioning `zoom('xy')` I was preparing to protest, recommending defining the API as `tight(...)` :smile: I think putting this functionality inside `zoom()` would be counterintuitive and confusing, and arguably it shouldn't have ended up there in the first place. I'd much prefer keeping only this PR, with a potential deprecation of the `zoom('tight')` case in favour of the dedicated `tight` machinery. Especially since in the hypothetical future we could allow arbitrary view angles to be passed to `tight()`.

> I kept the existing functionality to put the focal point at (x, y, 0), in the case of using 'xy', and the camera position is set according to a distance that depends on the view angle. I don't think the view angle in parallel projection mode should matter. This can also make the camera position be inside the object. I did some testing, and I couldn't reliably reproduce any problem with this approach, but it seems fishy to me.

Are you sure you meant "view angle" in the second sentence here? What I'd expect not to matter is the _distance_, and that would also fit in context. And if you indeed meant the distance: I also wouldn't be surprised if the distance had no effect with parallel projection. I seem to recall some discussion that interactive zooming works completely differently under the hood when parallel projection is enabled (since dolly doesn't do anything).

MatthewFlamm:
> Are you sure you meant "view angle" in the second sentence here?

Yes, I think there is confusion on the logic here.  I'm not sure I can explain why it is done this way, but it seems to work.  @akaszynski might be able to weigh in.  I generalized the logic to any direction.

The original code does

https://github.com/pyvista/pyvista/blob/16755b72f52143c21b7502c2df62a7971ac650ba/pyvista/plotting/camera.py#L693-L695

Then it sets the focal point to be at the mesh center but at xy plane (z=0), then the camera position is offset by this distance in the z direction (or negative z direction).

MatthewFlamm:
The view angle, I think, is related to the perspective in the rendering which adds to the confusion.  See https://vtk.org/doc/nightly/html/classvtkCamera.html#ab43b787eb19b922cc92f91de239e2a0a

pyvista/plotting/_utilities.py
```diff
@@ -0,0 +1,51 @@
...
+        Plane to return vectors for.
+    negative : bool
+        Whether to view from opposite direction.  Default ``False``.
+
+    Returns
+    -------
+    direction: List
+        [x, y, z] vector that points in the viewing direction
+    viewup: List
+        [x, y, z] vector that points to the viewup direction
```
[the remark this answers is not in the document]
MatthewFlamm: I'll fix List to list after committing your suggestion.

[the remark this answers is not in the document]
MatthewFlamm: I need to do a survey of the pyvista preference for naming returns.  I strongly prefer having a name in the returns particularly when there are multiple values.  Otherwise it is harder to find out at a glance which order the parameters are returned.

pyvista/plotting/renderer.py
```diff
@@ -2385,11 +2386,7 @@ def view_xy(self, negative=False):
         >>> pl.show()
 
         """
-        vec = np.array([0, 0, 1])
-        viewup = np.array([0, 1, 0])
-        if negative:
-            vec *= -1
-        self.view_vector(vec, viewup)
+        self.view_vector(*view_vectors('xy'))
```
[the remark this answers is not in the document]
MatthewFlamm: Good catch!  Also proves that these image regression tests are truly needed.

pyvista/plotting/_utilities.py
```diff
@@ -0,0 +1,51 @@
+"""Utilities for plotting."""
```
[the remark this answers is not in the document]
MatthewFlamm: It didn't seem to fit with the existing functions, but I think you are right that this addition doesn't need it's own module.  I'll move to helpers.

pyvista/plotting/camera.py
MatthewFlamm: > I guess what I'm missing here is that this method is special-cased for vectors where exactly one component of any vector is nonzero, and that nonzero value is +-1.

This is exactly the case here. See my separate comment below for the logic.  I'm open to a more clear implementation, but your comments show at the very least we need to spell out the logic using comments.

pyvista/plotting/_utilities.py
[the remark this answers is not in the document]
adeak: I don't mind naming them myself, although it's often weird because the names don't match things that are returned (often there's no name being returned, but an expression instead), and the name is meaningless when you call the function. I seem to remember @akaszynski actively removing names from return specifications.

For future reference, feel free to edit my suggestion comments if that makes it easier to incorporate further changes.

adeak:
> > Are you sure you meant "view angle" in the second sentence here?
> 
> Yes, I think there is confusion on the logic here. I'm not sure I can explain why it is done this way, but it seems to work. @akaszynski might be able to weigh in. I generalized the logic to any direction.
> 
> The original code does
> 
> https://github.com/pyvista/pyvista/blob/16755b72f52143c21b7502c2df62a7971ac650ba/pyvista/plotting/camera.py#L693-L695
> 
> Then it sets the focal point to be at the mesh center but at xy plane (z=0), then the camera position is offset by this distance in the z direction (or negative z direction).

OK, so based on that I can guess what it does, but I think it's doing it wrong.

If I understand correctly, `dx` and `dy` are the extent of the mesh along `x` and `y` (assuming `xy` view here). Then you consider two right triangles:
<details>
<summary>Click for snippet</summary>

```py
import pyvista as pv

plane = pv.Plane(i_size=1, j_size=2)
poly = pv.PolyData([
    [0, 0, 0],
    [1/2, 0, 0],
    [0, 1, 0],
    [0, 0, 2]
])
poly.lines = [
    4, 0, 1, 3, 0,
    4, 0, 2, 3, 0,
]
poly.point_data['labels'] = ['O', 'A', 'B', 'C']

plotter = pv.Plotter()
plotter.add_mesh(plane, color='lightblue')
plotter.add_mesh(poly, color='red', line_width=5)
plotter.add_point_labels(poly, 'labels', font_size=20, always_visible=True, shape=None, text_color='black')
plotter.enable_anti_aliasing('ssaa')
plotter.show()
```
</details>

plane with two triangles formed with center of plane, position of camera above the plane and either midpoints of a plane edge

The `max()` call with `aspect` inside probably chooses either the AOC or the BOC triangle, whichever is wider on the screen (or something like that). In this plot `dx` and `dy` are the plane lengths along `x` and `y`. So what I'd expect to see, given that the angle OCA (or OCB) is half the view angle, is
```
AO / OC = tan(view_angle/2)
# in other words
(dx/2) / dist = tan(view_angle/2)
# in other words
dist = dx / tan(view_angle/2) / 2
```
which is why I said that I expect the original version to be buggy: we have to divide by the tangent rather than the sine.

But now I also understand that the view angle is not what I thought it was, and indeed the VTK docs you linked makes it explicit that the view angle doesn't matter for parallel projection. I guess what's happening is that a perspective view is used to size up the scene, and then parallel projection is enabled for the tight view (we should make sure that the function works if the view was parallel projected to begin with, unless we do that in the function; I haven't checked).

Anyway, I don't understand why we'd want to mess with the focal point: why not put it in the center, why shift it into `z=0`? Does it work if we keep it naive, with the focal point in the center and just shifting the camera with the distance according to my above snippet?

pyvista/plotting/_utilities.py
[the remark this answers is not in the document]
MatthewFlamm: Now I remember why i did this, I was mostly lazy.  There was a circular import due to `helpers` importing `Plotter` directly.  I fixed this in https://github.com/pyvista/pyvista/pull/3391/commits/df2984051a424d49cf47a2608f22f46826a7437c

pyvista/plotting/_utilities.py
```diff
@@ -0,0 +1,51 @@
...
+
+import numpy as np
+
+
+def view_vectors(view: str, negative: bool = False) -> Tuple[np.ndarray, np.ndarray]:
+    """Given a plane to view, return vectors for setting up camera.
+
+    Parameters
+    ----------
+    view : {'xy', 'yx', 'xz', 'zx', 'yz', 'zy'}
```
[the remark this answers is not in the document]
MatthewFlamm: This is a numpydoc supported format, but I'm not sure if classifies it as conventional.  I tend to like this format since it is immediately obvious that it isn't _any_ string but only one from a set of six possibilities.  But I'm fine if the PyVista preference is to only use the type.

MatthewFlamm:
I think the existing logic in this function is flawed.  I can still pass the tests, which look correct to me, by changing this line

https://github.com/pyvista/pyvista/blob/abe7c0808a6067e91cc232a0c699a8fd47c88937/pyvista/plotting/camera.py#L719

to

```py
dist = 1
```

or even

```
dist = 100
```

`dist~2.23` in the original calculation, but it seems to completely not matter to the final image what the value is.  The only thing that seems to matter is the direction that you are viewing, which makes sense because we are in parallel projection mode.  I propose it would be more understandable to use the mesh center for the focal point, and then offset by 1 unit opposite to the view direction.

adeak:
> I propose it would be more understandable to use the mesh center for the focal point, and then offset by 1 unit opposite to the view direction for camera position.

I agree. I have a strong hunch that doing it right in the way I proposed would make this work for perspective view. If we plan to keep parallel projection this probably doesn't matter.

pyvista/plotting/camera.py
```diff
@@ -690,23 +705,36 @@ def tight(self, padding=0.0, adjust_render_window=True):
...
+        direction, viewup = view_vectors(view, negative)
+        horizontal = np.cross(direction, viewup)
+
+        vert_dist = np.abs(np.dot(objects_size, viewup))
+        horiz_dist = np.abs(np.dot(objects_size, horizontal))
+
+        # set focal point to objects center
+        # offset camera position from objects center by dist in opposite of viewing direction
+        dist = 1
+        camera_position = position + dist * direction
```
MatthewFlamm, line 723: The comment above says opposite direction, but the code here doesn't take the negative.  See #3416 for the reason why.

adeak: Good catch.

pyvista/plotting/helpers.py
```diff
@@ -462,3 +462,52 @@ def plot_itk(mesh, color=None, scalars=None, opacity=1.0, smooth_shading=False):
...
+        vec = np.array([0, 1, 0])
+        viewup = np.array([1, 0, 0])
+    elif view == 'yz':
+        vec = np.array([1, 0, 0])
+        viewup = np.array([0, 0, 1])
+    elif view == 'zy':
+        vec = np.array([-1, 0, 0])
+        viewup = np.array([0, 1, 0])
+    else:
+        raise ValueError(
```
adeak, line 506: Codecov claims this error needs a test case.

pyvista/plotting/camera.py
```diff
@@ -690,23 +705,36 @@ def tight(self, padding=0.0, adjust_render_window=True):
...
+        objects_size = position1 - position0
+        position = position0 + objects_size / 2
+
+        direction, viewup = view_vectors(view, negative)
+        horizontal = np.cross(direction, viewup)
+
+        vert_dist = np.abs(np.dot(objects_size, viewup))
+        horiz_dist = np.abs(np.dot(objects_size, horizontal))
+
+        # set focal point to objects center
```
adeak, line 720: Very tiny nitpick
```suggestion
        # set focal point to objects' center
```
or
```suggestion
        # set focal point to object's center
```
Which brings me to a question: does this also work when there are multiple meshes in the scene? How is this center computed?

pyvista/plotting/camera.py
```diff
@@ -690,23 +705,36 @@ def tight(self, padding=0.0, adjust_render_window=True):
...
+        # offset camera position from objects center by dist in opposite of viewing direction
```
adeak, line 721: Might be worth noting this:
```suggestion
        # offset camera position from objects center by dist in opposite of viewing direction
        # (actual distance doesn't matter due to parallel projection)
```

pyvista/plotting/camera.py
```diff
@@ -645,9 +647,16 @@ def copy(self):
...
-    def tight(self, padding=0.0, adjust_render_window=True):
+    def tight(self, padding=0.0, adjust_render_window=True, view='xy', negative=False):
         """Adjust the camera position so that the actors fill the entire renderer.
 
+        The camera view direction is reoriented to be normal to the ``view``
+        plane. When ``negative=False``, The first letter of ``view`` refers
+        to the axis that points to the right. The second letter of ``view``
+        refers to axis that points up.  When ``negative=True``, the first
+        letter refers to the axis that points left.  The up direction is
+        unchanged.
```
adeak, line 658: Should we also mention that this enables parallel projection? Sorry to keep coming back to this docstring...

pyvista/plotting/_utilities.py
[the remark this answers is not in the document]
adeak: If it's supported by numpydoc then it's good enough for me, thanks. I'll let this unresolved for now to let others chime in.

pyvista/plotting/camera.py
MatthewFlamm: I added a test for this.  It uses the bounds of all the viewable objects in the scene and determines the 'view center'.

pyvista/plotting/helpers.py
MatthewFlamm: Good catch.  We also lacked specific tests for this function.  Added tests.

pyvista/plotting/camera.py
MatthewFlamm: I agree, added.

pyvista/plotting/camera.py
```diff
@@ -691,22 +699,37 @@ def tight(self, padding=0.0, adjust_render_window=True):
...
-        self.SetPosition(x0 + dx / 2, y0 + dy / 2, dist * (1 + padding))
-        self.SetFocalPoint(x0 + dx / 2, y0 + dy / 2, 0)
+        distance_object = np.array([(x1 - x0), (y1 - y0), (z1 - z0)])
+        position = np.array([x0, y0, z0]) + distance_object / 2
+
+        direction, viewup = view_vectors(view, negative)
+        horizontal = np.cross(direction, viewup)
+
+        vert_dist = np.abs(np.dot(distance_object, viewup))
+        horiz_dist = np.abs(np.dot(distance_object, horizontal))
```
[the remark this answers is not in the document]
MatthewFlamm: The logic was changed but I used this methodology.

pyvista/plotting/_utilities.py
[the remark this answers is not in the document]
MatthewFlamm: On a non-representative and non-thorough sampling of pyvista code, return names are not typically provided even with multiple returns.  Other packages like numpy provide names.  For example: https://numpy.org/doc/stable/reference/generated/numpy.ravel.html#numpy.ravel

> image

In this case the actual name of the return is unimportant, but the result can be referenced in the description. To me this is a big positive. In cases like `linspace`, with multiple returns it is immediately obvious what each return object is without having to read the entire description.

https://numpy.org/doc/stable/reference/generated/numpy.linspace.html

This is now way out of scope of this PR.  I will open an issue to gather feedback.  I will remove the names in this PR to not bog down the discussion.

pyvista/plotting/helpers.py
```diff
@@ -462,3 +462,52 @@ def plot_itk(mesh, color=None, scalars=None, opacity=1.0, smooth_shading=False):
...
+    negative : bool
+        Whether to view from opposite direction.  Default ``False``.
+
+    Returns
+    -------
+    numpy.ndarray
+        [x, y, z] vector that points in the viewing direction
+
+    numpy.ndarray
+        [x, y, z] vector that points to the viewup direction
```
akaszynski, line 484: Where there are multiple returns, I think we're good to return the variable names (especially that they're useful). Agree with your points in https://github.com/pyvista/pyvista/issues/3422.

```suggestion
    vec : numpy.ndarray
        [x, y, z] vector that points in the viewing direction

    viewup : numpy.ndarray
        [x, y, z] vector that points to the viewup direction
```

akaszynski (approved):
Made one change based on your comment in https://github.com/pyvista/pyvista/issues/3422, otherwise LGTM.

akaszynski (approved):
Re-approving after merge.

PR #3393, opened 2022-09-30
Use PyVista start headless display action

akaszynski:
Use the pyvista start headless display action within the style and docstring check.

Sometimes our workflows fail when starting xvfb multiple times with `xvfb-run` and this should fix it.

### DIFF
--- a/.github/workflows/style_docstr.yml
+++ b/.github/workflows/style_docstr.yml
@@ -47,19 +47,17 @@ jobs:
       - name: Install pyvista and doctest requirements
         run: pip install . -r requirements_test.txt
 
-      - uses: awalsh128/cache-apt-pkgs-action@v1.1.1
-        with:
-          packages: libgl1-mesa-glx xvfb python-tk
-          version: 3.0
+      - name: Setup headless display
+        uses: pyvista/setup-headless-display-action@v1
 
       - name: Software Report
         run: |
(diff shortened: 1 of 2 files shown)


akaszynski:
Near trivial fix and "Style and Docstring Check" is passing. Failure on CI is due to the greeting failing, not caused by this PR.

.github/workflows/style_docstr.yml
```diff
@@ -47,19 +47,17 @@ jobs:
       - name: Install pyvista and doctest requirements
         run: pip install . -r requirements_test.txt
 
-      - uses: awalsh128/cache-apt-pkgs-action@v1.1.1
```
adeak, line 50: Do we not need the cache action anymore? Or is that somehow incompatible with or included in our setup-headless-display action?

banesullivan: >incompatible with or included in our setup-headless-display action?

Not incompatible! But also not included. See https://github.com/pyvista/setup-headless-display-action/issues/3

banesullivan (approved):

.github/workflows/style_docstr.yml
adeak: @banesullivan so why is it removed in this PR? Or do you mean that your follow-up will fix this once it's merged?

akaszynski: > Or do you mean that your follow-up will fix this once it's merged?

I've removed it since the logic will be moved to the headless action.

PR #3399, opened 2022-10-02
Fix the error of fmt in add_point_scalar_labels method.

tkoyama010:
### Overview

Fixed a bug that prevented the `fmt` argument of add_point_scalar_labels from working.
```python
import pyvista
from pyvista import examples

mesh = examples.load_uniform().slice()
p = pyvista.Plotter()
p.add_mesh(mesh, scalars="Spatial Point Data", show_edges=True)
p.add_point_scalar_labels(
    mesh, "Spatial Point Data", point_size=20, font_size=36, fmt=r"%.3f"
)
p.camera_position = [(7, 4, 5), (4.4, 7.0, 7.2), (0.8, 0.5, 0.25)]
p.show(before_close_callback=verify_cache_image)
```

| main  | PR |
| ------------- | ------------- |
| add_point_scalar_labels_fmt | add_point_scalar_labels_fmt  |

 

### Details

- None


pyvista/plotting/plotting.py
```diff
@@ -4725,12 +4725,10 @@ def add_point_scalar_labels(self, points, labels, fmt=None, preamble='', **kwarg
             raise TypeError(f'input points must be a pyvista dataset, not: {type(points)}')
         if not isinstance(labels, str):
             raise TypeError('labels must be a string name of the scalars array to use')
-        if fmt is None:
```
MatthewFlamm, line 4728: Why was this removed?  It makes sense to me to use the theme format if `fmt` isn't explicitly given.  Is there something wrong with this logic?

pyvista/plotting/plotting.py
```diff
@@ -4725,12 +4725,10 @@ def add_point_scalar_labels(self, points, labels, fmt=None, preamble='', **kwarg
...
-            fmt = self._theme.font.fmt
         if fmt is None:
             fmt = '%.6e'
         scalars = points.point_data[labels]
-        phrase = f'{preamble} %.3e'
+        phrase = f'{preamble} ' + fmt
```
MatthewFlamm, line 4731:
```suggestion
        phrase = f'{preamble} {fmt}'
```

pyvista/plotting/plotting.py
tkoyama010: The reason is that the fmt is then overwritten with "%.6e". If fmt is None, the line is not working.
https://github.com/pyvista/pyvista/blob/c95debf226624195b16abac3b30a01d3676c0715/pyvista/plotting/plotting.py#L4728-L4731
Should we use `self._theme.font.fmt` instead of "%.6e"?

adeak: I read those four lines like this:
1. if `fmt` is `None`, fall back to theme
2. if that is _also_ `None`, fall back to `'%.6e'`.

MatthewFlamm: I think it will only be overwritten if `self._theme.font.fmt` is also `None`.

tkoyama010: Ah! That makes sense. I revert it. Thanks!

pyvista/plotting/plotting.py
```diff
@@ -4725,12 +4725,10 @@ def add_point_scalar_labels(self, points, labels, fmt=None, preamble='', **kwarg
             raise TypeError(f'input points must be a pyvista dataset, not: {type(points)}')
         if not isinstance(labels, str):
             raise TypeError('labels must be a string name of the scalars array to use')
```
tkoyama010, line 4727:
```suggestion
            raise TypeError('labels must be a string name of the scalars array to use')
        if fmt is None:
            fmt = self._theme.font.fmt
```

pyvista/plotting/plotting.py
tkoyama010: Resolved in 675e3cd43f0c6d3a4a6d3cb1f7f6bae5626174ea.

akaszynski (dismissed):
This is good to go. Thanks @tkoyama010.

tests/plotting/test_plotting.py
```diff
@@ -3003,3 +3003,13 @@ def test_wireframe_color(sphere):
...
         lighting=False, color='b', style='wireframe', before_close_callback=verify_cache_image
     )
+
+
+@skip_windows
+def test_add_point_scalar_labels_fmt():
+    mesh = examples.load_uniform().slice()
+    p = pyvista.Plotter()
+    p.add_mesh(mesh, scalars="Spatial Point Data", show_edges=True)
+    p.add_point_scalar_labels(mesh, "Spatial Point Data", point_size=20, font_size=36, fmt=r'%.3f')
```
adeak, line 3013: I don't think the raw string literal is necessary here. Harmless though.
```suggestion
    p.add_point_scalar_labels(mesh, "Spatial Point Data", point_size=20, font_size=36, fmt='%.3f')
```

adeak (approved):

MatthewFlamm (approved):
LGTM

PR #3402, opened 2022-10-03
Minor cleanup of lookup table opacity

adeak:
Minor cleanup. I also have a few questions/remarks, see comments below.

### DIFF
--- a/pyvista/plotting/lookup_table.py
+++ b/pyvista/plotting/lookup_table.py
@@ -720,7 +720,8 @@ class LookupTable(_vtk.vtkLookupTable):
 
         kind : str, default: 'quadratic'
             The interpolation kind if ``interpolate`` is ``True`` and
-            ``scipy`` is available. Options are:
+            ``scipy`` is available. If ``scipy`` is not available, linear
+            interpolation is always used. Options are:
 
             - ``'linear'``
             - ``'nearest'``
--- a/pyvista/plotting/lookup_table.py
+++ b/pyvista/plotting/lookup_table.py
@@ -749,7 +749,7 @@ class LookupTable(_vtk.vtkLookupTable):
 
         """
         if isinstance(opacity, (float, int)):
-            if opacity < 0 or opacity > 1:
+            if not 0 <= opacity <= 1:
                 raise ValueError(f'Opacity must be between 0 and 1, got {opacity}')
             self.values[:, -1] = opacity * 255
         elif len(opacity) == self.n_values:
--- a/pyvista/plotting/lookup_table.py
+++ b/pyvista/plotting/lookup_table.py
@@ -710,15 +710,16 @@ class LookupTable(_vtk.vtkLookupTable):
             values that will be interpolated across the ``n_color`` range for
             user defined mappings. Values must be between 0 and 1.
 
-            If an ``int``, simply applies the same opacity across the entire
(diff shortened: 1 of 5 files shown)


pyvista/plotting/lookup_table.py
```diff
@@ -748,7 +750,7 @@ def apply_opacity(self, opacity, interpolate: bool = True, kind: str = 'quadrati
         """
         if isinstance(opacity, (float, int)):
```
adeak, line 751: It might be a bit surprising that ints are interpreted as floats, but I guess the reason for this is that this is short-hand for passing 1 instead of 1.0. Since we raise in case of ints greated than 1 this is probably fine, because people passing 1 probably don't mean 1/255.

pyvista/plotting/lookup_table.py
```diff
@@ -748,7 +750,7 @@ def apply_opacity(self, opacity, interpolate: bool = True, kind: str = 'quadrati
...
             if opacity < 0 or opacity > 1:
-                raise ValueError(f'Opacity must bet between 0 and 1, got {opacity}')
+                raise ValueError(f'Opacity must be between 0 and 1, got {opacity}')
             self.values[:, -1] = opacity * 255
         elif len(opacity) == self.n_values:
```
adeak, line 755: Right below this line (I can't annotate it, alas) we have
```py
            self.values[:, -1] = np.array(opacity, copy=False).copy()
```
Why is this not just
```py
            self.values[:, -1] = np.array(opacity)
```
Am I missing something here?

akaszynski: You're not missing anything. This is why we have reviewers.

pyvista/plotting/tools.py
```diff
@@ -478,7 +489,10 @@ def opacity_transfer_function(mapping, n_colors, interpolate=True, kind='quadrat
         try:
             return transfer_func[mapping]
         except KeyError:
-            raise KeyError(f'opactiy transfer function ({mapping}) unknown.')
+            raise ValueError(
+                f'Opacity transfer function ({mapping}) unknown. '
+                f'Valid options: {list(transfer_func.keys())}'
+            ) from None
```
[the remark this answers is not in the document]
akaszynski: It's breaking and that's fine, we're still in [ZeroVer](https://0ver.org/).

pyvista/plotting/lookup_table.py
```diff
@@ -710,16 +710,18 @@ def apply_opacity(self, opacity, interpolate: bool = True, kind: str = 'quadrati
...
 
         kind : str, default: 'quadratic'
-            The interepolation kind if ``interpolate`` is ``True`` and
-            ``scipy`` is available. Options are:
+            The interpolation kind if ``interpolate`` is ``True`` and
+            ``scipy`` is available. If ``scipy`` is not available, linear
+            interpolation is always used. Options are:
 
             - ``'linear'``
             - ``'nearest'``
```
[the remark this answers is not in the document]
akaszynski: Done in 3a10d213ce525383686f7f5d339e4a16bad51e68

akaszynski (approved):
LGTM. Thanks for the fixes.

adeak:
Thank you too @akaszynski.

tkoyama010 (approved):
LGTM

PR #3407, opened 2022-10-04
continue on error for the greeter

akaszynski:
I'm implementing the suggestion from [this discussion point](https://github.com/actions/first-interaction/issues/101#issuecomment-1262487501) to simply continue on error until this has been fixed upstream.

### DIFF
--- a/.github/workflows/greetings.yml
+++ b/.github/workflows/greetings.yml
@@ -16,6 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/first-interaction@v1
+        continue-on-error: true
         with:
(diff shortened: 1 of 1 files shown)


tkoyama010 (approved):
Since this CI is originally designed to merge Pull Requests even if an error occurs (i.e., it is not related to testing), there is no problem to keep this flag set to True in the future.

akaszynski:
Continue on error still means these fail. I'm going to simply comment this out until this is fixed.

akaszynski:
Simply disabled via [disabling-a-workflow](https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow#disabling-a-workflow).

PR #3412, opened 2022-10-05
Rename pass_cell_arrays and pass_point_arrays in methods of DataSetFilters

rodrigomologni:
This RP only rename the parameters `pass_cell_arrays` to `pass_cell_data` and `pass_point_arrrays` to `pass_point_data` in methods of `DataSetFilters`.

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -21,7 +21,6 @@ from pyvista.utilities import (
     wrap,
 )
 from pyvista.utilities.cells import numpy_to_idarr
-from pyvista.utilities.misc import PyVistaFutureWarning
 
 
 @abstract_class
@@ -2826,27 +2825,8 @@ class DataSetFilters:
         True
 
         """
-        if pass_cell_data is None:
-            pass_cell_data = False
-            warnings.warn(
-                'The default value of the ``pass_cell_data`` keyword argument will change in ' 
-                'a future version to ``True`` to match the behavior of VTK. We recommend ' 
-                'passing the keyword explicitly to prevent future surprises.',
-                PyVistaFutureWarning
-            )
-
-        if pass_point_data is None:
-            pass_point_data = False
-            warnings.warn(
-                'The default value of the ``pass_point_data`` keyword argument will change in ' 
-                'a future version to ``True`` to match the behavior of VTK. We recommend ' 
-                'passing the keyword explicitly to prevent future surprises.',
-                PyVistaFutureWarning
-            )
-
         if not pyvista.is_pyvista_dataset(points):
             points = pyvista.wrap(points)
-
         alg = _vtk.vtkProbeFilter()
         alg.SetInputData(points)
         alg.SetSourceData(self)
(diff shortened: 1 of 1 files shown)


banesullivan (changes requested):
@rodrigomologni, thank you for making this more consistent with the rest of the API. Can you also implement a deprecation policy where the old kwargs are accepted under `kwargs` and the defaults for `pass_cell_data` and `pass_point_data` are `None`? When `None`, they'll throw a `PyVistaDeprecationWarning` warning.

You can find a nomillay similar example here:

https://github.com/pyvista/pyvista/blob/14cb03918458a8fa86d99dc1db5019838c70c777/pyvista/core/filters/poly_data.py#L2776-L2783

rodrigomologni:
Hi @banesullivan! Done! Check the changes, please.

banesullivan (changes requested):
I'm sorry, I think I may have misdescribed what I was looking for. It should check to see if the user passed the old name of the kwarg. Something like:

```py
def func(pass_cell_data=None, **kwargs):
    if pass_cell_data is None and "pass_cell_arrays" in kwargs:
        pass_cell_data = kwargs.get("pass_cell_arrays")
        warnings.warn("...", PyVistaDeprecationWarning)
    else:
        pass_cell_data = True  # Same default value
    ...  # Same for `pass_point_data`
    assert_empty_kwargs(**kwargs)  # Existing helper method
    ...
```
 Then, the error message should alert them that the old name of the argument is deperecated rather than the default value having changed (as it hasn't)

rodrigomologni:
@banesullivan no problem. Done! Check the changes again, please.

pyvista/core/filters/data_set.py
```diff
@@ -2455,6 +2456,13 @@ def cell_data_to_point_data(self, pass_cell_data=False, progress_bar=False):
         >>> surf.plot(scalars='Area')
 
         """
+        if pass_cell_data is False and 'pass_cell_arrays' in kwargs:
+            pass_cell_data = kwargs.pop('pass_cell_arrays')
+            warnings.warn(
+                'Use of `pass_cell_arrays` is deprecated. Use `pass_cell_data` instead.',
+                PyVistaFutureWarning
```
adeak, line 2463: Arguably if this is a deprecation, it should be a `DeprecationWarning`.

adeak: Although [`DeprecationWarning`s are almost always suppressed](https://docs.python.org/3/library/warnings.html#default-warning-filter). Do we have pyvista-specific deprecation warnings?

banesullivan: Ah yes, It should use `PyVistaDeprecationWarning` as I mentioned in https://github.com/pyvista/pyvista/pull/3412#pullrequestreview-1131535378

banesullivan:
The documentation build is failing due to `numpydoc` issue:

```
WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.CompositeFilters.cell_data_to_point_data':
  PR01: Parameters {'**kwargs'} not documented
```

side note: we need to streamline CI checks like that to not be embedded as a part of the documentation build... this error was buried in *thousands* of lines of output

akaszynski:
> side note: we need to streamline CI checks like that to not be embedded as a part of the documentation build... this error was buried in _thousands_ of lines of output

Yep. Worked on https://github.com/pyvista/numpydoc-validation, but busy at the moment. Would absolutely love for someone to take this up.

PR #3424, opened 2022-10-06
Remove pip cache in docker-package.yml

akaszynski:
Our CI regularly fails on the docker build step due to:
```
Cache folder path is retrieved for pip but doesn't exist on disk: /home/runner/.cache/pip
```
I've yet to see the cache hit and it's not worth the possibility of the docker step failing on main. Let's remove this.


tkoyama010 (approved):

banesullivan (approved):
Yeah this is definitely not needed for the Docker build... not sure how/why was added

PR #2729, opened 2022-06-02
Improve texture implementation and documentation

akaszynski:
We have a lot of great work in `pyvista.Texture` that isn't exposed via our documentation. This PR exposes the class in our documentation and:

- Improves the class documentation
- Adds `repr` to `pyvista.Texture`
- Allow init from 6 images to construct a `cubemap`. The decision was made to add it here since the `pyvista.cubemap` helper is already cluttered.
- Add in a `plot` method that works with cubemap enabled Textures.
- Support greyscale textures.
- Add rotation methods.
- Speed up array access and improve performance:

#### `main` branch

```py

>>> from pyvista import examples
>>> texture = examples.load_globe_texture()

>>> timeit texture.to_array()
659 µs ± 67.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

>>> timeit texture.plot(off_screen=True)
494 ms ± 73.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

#### This branch

```py
>>> timeit texture.image_data
48.9 µs ± 1.06 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

>>> timeit texture.plot(off_screen=True)
421 ms ± 8.88 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

### Bonus

Fixed the `topo-map.py` example.

PR #3431, opened 2022-10-07
Improve Texture implementation and enable 3D data for gridded datasets

akaszynski:
This PR is a redo of #2729 needed to addresses all the comments made by @adeak as well as changes in the codebase in the past few months.

### Implement Multi-Dimension point and cell data for Gridded DataSets
- Completely eliminates the need for `.ravel('F')` when setting point or cell data for `pyvista.Grid` objects, while also removing the need to `reshape` the data when extracting point or cell data from gridded objects.
- Breaking change with full backwards compatibility on the setter, but not on the getter since a `(N, M, O)` dimension `pyvista.Grid` object will return a `(N, M, O)` shaped array of data. In my opinion, it's worth breaking backwards compatibility because:
  - It will be obvious what the error is based on the shape of the new `pyvista_ndarray`.
  - The `shape` polymorphism required for full backwards compatibility would be overly complex and brittle.
  - The `1:1` correspondence between the shape of the `point_data` arrays and the `x`, `y`, `z` arrays is highly desirable, as is the need to eliminate any data reshaping.
- The Gridded dataset point data refactor was necessitated by the desire to clean up the `pyvista.Texture` class, and it became obvious that our current implementation left much to be desired in terms of usability.

### `pyvista.Texture` Class
- Improves the class documentation
- Adds `repr` to `pyvista.Texture`
- Allow init from 6 images to construct a `cubemap`. The decision was made to add it here since the `pyvista.cubemap` helper is already cluttered.
- Add in a `plot` method that works with cubemap enabled Textures.
- Support greyscale textures.
- Add rotation methods.
- Speed up array access
- *Finally* fixed the `topo-map.py` example:

tmp

### Remaining Tasks
- [x] Improve coverage to ~100%


pyvista/core/dataobject.py
```diff
@@ -372,7 +372,11 @@ def add_field_data(self, array: np.ndarray, name: str, deep=True):
         >>> mesh.add_field_data(['I could', 'write', 'notes', 'here'],
         ...                      'my-field-data')
         >>> mesh['my-field-data']
-        pyvista_ndarray(['I could', 'write', 'notes', 'here'], dtype='<U7')
+        pyvista_ndarray([[['I could'],
+                          ['notes']],
+        <BLANKLINE>
+                         [['write'],
+                          ['here']]], dtype='<U7')
```
adeak, line 380: If my understanding is correct, this is buggy now. Field data should not have any shape associated with the grid, so we should keep field data intact. I assume this means having to special-case field data in `DataSetAttributes`.

adeak (commented):
Just one quick remark as I glanced over the PR. Judging by the PR description this sounds amazing. I'll try to review it in a timely manner.

MatthewFlamm:
Firstly, I understand the useful aspects of this PR.  From a personal impact, I'm slightly negative on the changes to the data shape changes, but maybe there are workarounds/modifications that could lessen the negative impacts.  I will try to think about ways we could have 'best of both worlds'.

- This looks really useful and will clear up a lot of confusion for setting and retrieving data when the user is only working with this grid type.
- This will create problems when users want to do operations with the points, still of shape `(n_points, 3)` and the array data, shape `(nx, ny, nz)`.
- This will create problems for my workflows where I could be applying the same workflow on multiple mesh types, e.g. UniformGrid and UnstructuredGrid for example.  The current code state is highly desirable in this case as the everything has the same structure.
- We are now introducing structural changes to the DataSets.  It will be unclear to users which ones fall into which category.  It is even more confusing that we have `UnstructuredGrid` which isn't a `Grid`.  This needs good documentation added.

akaszynski:
> I'm slightly negative on the changes to the data shape changes, but maybe there are workarounds/modifications that could lessen the negative impacts.

I'm starting to think that we might need to add in a new dataset attribute. Right now we have `cell_data` and `point_data`, but perhaps we might want to add `structured_point_data` or `grid_point_data` that returns a view on the data that's organized in this manner. I 100% agree that we shouldn't break backwards compatibility and looking at the failures with MNE, it's clear that we should avoid this.

> We are now introducing structural changes to the DataSets. It will be unclear to users which ones fall into which category. It is even more confusing that we have UnstructuredGrid which isn't a Grid. This needs good documentation added.

Good point. It should really be something like `UnstructuredDataSet`, but I feel like it's a bit late for that. If we implement the new attribute, will that be sufficient to clear it up?

banesullivan:
These changes are very needed. @akaszynski, I'm wondering what's left here and how would you like to approach bringing this stale PR back to life? Should we fix the merge conflict or break this into separate PRs? Ideally, I'd like to land this for 0.38.0

akaszynski:
> These changes are very needed. @akaszynski, I'm wondering what's left here and how would you like to approach bringing this stale PR back to life? Should we fix the merge conflict or break this into separate PRs? Ideally, I'd like to land this for 0.38.0

Will be pushed to v0.39.0

Part of this is implemented in #4179, the rest will be implemented in a second PR, whereby we can finally lay this PR to rest.

banesullivan:
@akaszynski, I'm closing this PR as that merge conflict looks too crazy for any mere mortal to handle (though you've proven to be more superhuman before...). Would be great to open these changes again as smaller, targeted PRs

akaszynski:
Agreed. Lots of it was already pulled out into other PRs. I can work on this again provided I have free time in the next decade.

Issue #3347, opened 2022-09-20
Optional parameters that are not documented as optional


MatthewFlamm:
### Describe what maintenance you would like added.

Optional parameters for a function should be documented in numpydoc style.

### Links to source code.

See below

### Pseudocode or Screenshots

Using the branch from https://github.com/numpy/numpydoc/pull/431 specifically https://github.com/numpy/numpydoc/pull/431/commits/83c4830009362a3be42f1255286bfd0bb364023a

yields a list of parameters that are missing documentation about being optional.

<details>
<summary>Errror codes</summary>

```txt
WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.DataSetMapper.set_scalars':
  PR11: Parameter "nan_color" is optional but not documented, or vice versa

WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.Plotter.add_cursor':
  PR11: Parameter "bounds" is optional but not documented, or vice versa

WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.Plotter.show':
  PR11: Parameter "return_img" is optional but not documented, or vice versa
  PR11: Parameter "cpos" is optional but not documented, or vice versa

WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.Renderer.add_bounding_box':
  PR11: Parameter "outline" is optional but not documented, or vice versa

WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.Renderer.enable_anti_aliasing':
  PR11: Parameter "aa_type" is optional but not documented, or vice versa

WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.Renderer.enable_depth_peeling':
  PR11: Parameter "number_of_peels" is optional but not documented, or vice versa
  PR11: Parameter "occlusion_ratio" is optional but not documented, or vice versa

WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.Renderer.get_default_cam_pos':
  PR11: Parameter "negative" is optional but not documented, or vice versa

WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.Renderer.reset_camera':
  PR11: Parameter "render" is optional but not documented, or vice versa
  PR11: Parameter "bounds" is optional but not documented, or vice versa

WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.Renderer.set_background':
  PR11: Parameter "color" is optional but not documented, or vice versa

[...]

PR #3444, opened 2022-10-11
Specify optional parameters in mapper.py and plotting.py

puripant:
### Overview

Optional parameters in `mapper.py` and `plotting.py` as listed in #3347 are specified in docstring.

### Details

- "nan_color" in `set_scalars()` in `mapper.py`
- "bounds" in `add_cursor()` in `plotting.py`
- "return_img" in `show()` in `plotting.py`
- "cpos" in `show()` in `plotting.py`

### DIFF
--- a/pyvista/plotting/mapper.py
+++ b/pyvista/plotting/mapper.py
@@ -471,7 +471,7 @@ class DataSetMapper(_vtk.vtkDataSetMapper, _BaseMapper):
             than zero are mapped to the smallest representable
             positive float.
 
-        nan_color : color_like
+        nan_color : color_like, optional
             The color to use for all ``NaN`` values in the plotted
             scalar array.
 
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -5894,11 +5894,11 @@ class Plotter(BasePlotter):
             screenshot in a separate call to ``show()`` or
             :func:`Plotter.screenshot`.
 
-        return_img : bool
+        return_img : bool, default: False
             Returns a numpy array representing the last image along
             with the camera position.
 
-        cpos : list(tuple(floats))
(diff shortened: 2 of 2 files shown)


puripant:
I am not sure how this build error happens. I barely changed anything and all changes were in the comments :/

akaszynski:
> I am not sure how this build error happens. I barely changed anything and all changes were in the comments :/

Looks like a simple build failure; these things happen from time to time. I've restarted it.

MatthewFlamm:
This PR mixes use of `optional` and `default`.  See #3347 for discussion on this.  If we want to merge this quickly without too much discussion, I'd suggest using only `optional` here since it is the most prevalent in the code today.

MatthewFlamm:
Looking a little closer, it looks like you've chosen to use `optional` when the default value is `None` and `default` when there is a specific value that is a default.  I actually like this approach.  So disregard my previous comment and we can await further comments on this.

puripant:
> Looking a little closer, it looks like you've chosen to use `optional` when the default value is `None` and `default` when there is a specific value that is a default. I actually like this approach. So disregard my previous comment and we can await further comments on this.

Yes, that's my intention. Glad you like it. Let's see if other people agree on this.

akaszynski:
> Looking a little closer, it looks like you've chosen to use optional when the default value is None and default when there is a specific value that is a default.

I prefer this as well. This needs to be implemented everywhere.

akaszynski:
Doc failures are weird. Rerunning... I'm not seeing this happen anywhere else.

MatthewFlamm (approved):
With some concurrence on the style here, LGTM

akaszynski:
Merging this soon regardless of the doc failure. I'll open a follow-up PR fixing this.

Issue #3339, opened 2022-09-18
export_vtkjs Does not Include PolyData Normals


aidanmolnar:
### Describe the bug, what's wrong, and what you expected.

Line 362 (shown below) of export_vtkjs that dumps the normals is commented out, causing them not to be included in the file. 
```python
    dump_color_array(dataset_dir, data_dir, color_array_info, container, compress)

    # PointData TCoords
    dump_t_coords(dataset_dir, data_dir, dataset, container, compress)
    # dump_normals(dataset_dir, data_dir, dataset, container, compress)

    return root
```

I expected the normals to be included in the output .vtkjs, but they were not.

I think the line should be uncommented by default or an optional argument could be added to the function.  If the feature isn't complete or not working as intended the documentation for the function should be updated to reflect that.

### Steps to reproduce the bug.

```python
import pyvista as pv
p = pv.Plotter()
p.add_mesh(pv.Sphere(radius=1.0))
p.export_vtkjs("pyvista")
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Sat Sep 17 20:28:13 2022 Eastern Daylight Time

                OS : Windows
            CPU(s) : 24
           Machine : AMD64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : NVIDIA GeForce RTX 3080/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 512.15

  Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64
  bit (AMD64)]

           pyvista : 0.36.1
               vtk : 9.1.0
             numpy : 1.23.3
           imageio : 2.21.3
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.6.0
--------------------------------------------------------------------------------
```

### Screenshots

Here is a comparison of the output for a simple sphere in the [vtkjs scene explorer](https://kitware.github.io/vtk-js/examples/SceneExplorer.html).
Line commented as it is currently:
![image](https://user-images.githubusercontent.com/43870861/190880444-a585ba42-efba-4c8e-ac4b-3bc26b88d3c5.png)
Line uncommented:
![image](https://user-images.githubusercontent.com/43870861/190880452-c62b04cc-6246-46af-a5da-6a981ea6562f.png)

banesullivan:
[...]

PR #3452, opened 2022-10-12
line to include PolyData Normals uncommented (line 362 in ./plotting/…

andrewrgarcia:
…export_vtkjs.py)

### Overview

Line 362 from ./plotting/export_vtkjs.py was uncommented to include normals. 

This resolves issue #3339 : export_vtkjs Does not Include PolyData Normals.

### DIFF
--- a/pyvista/plotting/export_vtkjs.py
+++ b/pyvista/plotting/export_vtkjs.py
@@ -359,7 +359,7 @@ def dump_poly_data(dataset_dir, data_dir, dataset, color_array_info, root=None, 
 
     # PointData TCoords
     dump_t_coords(dataset_dir, data_dir, dataset, container, compress)
-    # dump_normals(dataset_dir, data_dir, dataset, container, compress)
+    dump_normals(dataset_dir, data_dir, dataset, container, compress)
 
     return root
 


banesullivan (approved):
I'm happy to approve this PR without knowing the consequences. 

It's worth noting that this script came from VTK.js and ParaView: https://github.com/Kitware/vtk-js/blob/master/Utilities/ParaView/export-scene-macro.py

and it is commented out there too:

https://github.com/Kitware/vtk-js/blob/c05700301537c8c2ccac8b8a2166888f00075042/Utilities/ParaView/export-scene-macro.py#L314

It looks like @jourdain commented this out in 

https://github.com/Kitware/vtk-js/commit/3be2a71ffc9117b36563841fa04e0686b9031129

@andrewrgarcia, if this fixes your issue then I'm happy to merge it

jourdain:
That export is [natively supported](https://github.com/Kitware/VTK/blob/master/IO/Export/vtkJSONSceneExporter.h) and maintained on the C++ side of vtk. ParaView is using the vtk class to export the vtk.js content.

jourdain:
C++ is not as flexible as Python but definitely passing along that it has been ported to mainstream support...

Issue #3324, opened 2022-09-14
Consider implement a pyvista.testing module


jorgepiloto:
🐞 **Problem description**
For libraries depending on PyVista, it would be interesting to have some utility for validating PyVista figures. This would allow to maintain or even increase code coverage on these libraries.

📝 **Background**
So far, PyVista validates its `Plotter` class through the [verify_cache_image](https://github.com/pyvista/pyvista/blob/6aedb562ee8aeebba90fe163b757bd6a355b40a0/tests/plotting/test_plotting.py#L167), which is defined in [tests/plotting/test_plotting.py](https://github.com/pyvista/pyvista/blob/main/tests/plotting/test_plotting.py#L167).

Previous function is passed as a parameter to the `Plotter.plot(before_close_callback=verify_cache_image)` method that looks into the "image_cache/" directory. The `verify_cache_image` looks into the `image_cache/` directory and looks for the associated expected image from the test name.

However, the main issue is that the `verify_cache_image` with all its dependencies are not exposed in PyVista: they are specified in the `tests/` directory. This blocks libraries using PyVista from accessing this useful testing routine.

💡 **Proposed solutions**
Looking at other popular libraries in the Python ecosystem, most of them ship with a `testing` module so all the testing auxiliary routines can be reused:

- Matplotlib testing: https://github.com/matplotlib/matplotlib/tree/main/lib/matplotlib/testing
- NumPy testing: https://github.com/numpy/numpy/tree/main/numpy/testing
- Pandas testing: https://github.com/pandas-dev/pandas/blob/main/pandas/testing.py

This way, third libraries could run:

```
from pyvista.testing import verify_image_cache
```

and use all its functionality in their own test suite.

Additional solutions may include creating a Pytest plugin for comparing PyVista figures, see [pytest-mpl](https://github.com/matplotlib/pytest-mpl), a plugin for testing matplotlib figures.

MatthewFlamm:
I like this idea.  I want to take your suggestion one step further on `verify_cache_image` that it should be a Pytest plugin itself.  Then downstream libs do not even need to import it and could consider wrapping it to make an `autouse=True` fixture.  A start for this is in #3454, which makes this functionality into an internal Pytest fixture.

PR #3454, opened 2022-10-12
Add fixture for verify_cache_image

MatthewFlamm:
### Overview

 

This PR adds in a fixture that is `autouse=True` to automatically cache images.  This means that contributors no longer even need to worry about how to cache images.  It happens automagically.  Contributors only need to interface with this fixture if they need to control its behavior, e.g. skip a test.

### Details

The logic for caching and pytest commandline arg options are now housed in a new class, which avoids the use of global variables.  Now we modify class attributes so that these options are available in each test.  An instance variable `skip` is available so that contributors can selectively skip tests entirely or even skip a specific `show` command and then re-enable later.

This approach enables capturing of multiple calls to `show` in a test by appending `_X` where X is the number of times `show` is called.  To be backwards compatibility, in some sense, we do not append anything for the first time it is called.  Several tests were only capturing the last call to show, so this required modifying the image files.  Other tests are also now captured that weren't captured before, possibly unneccessarily.

Finally, the naming scheme no longer uses the `inspect` module and instead uses the pytest naming.  This allows us to automatically capture parametric tests in the naming style of pytest like `test_abc[True]` and `test_abc[False]`.

This PR also lays the groundwork for exposing this fixture as a pytest plugin for consumption by downstream libraries.  See #3324.


MatthewFlamm:
BTW, I was continually using `verify_image_cache` and had to fix it to be consistent with the current naming `verify_cache_image` when working on this PR.  #3324 also uses this `verify_image_cache` which suggests it is not a 'me' issue.  The naming of `verify_cache_image` just doesn't seem to be natural.  I will swap if others agree.

akaszynski:
This is way better. Nice work @MatthewFlamm. Agree with name change to `verify_image_cache`. This is only used internally, so let's pick a solid name before exposing it as an external plugin.

MatthewFlamm:
The themes don't play well with private attributes.  The leading underscore is stripped off in some functionality.  Instead of trying to fix this, I think it makes sense to promote `before_close_callback` to be a first class option to `Plotter.show` and then we can make it also public in the Theme.

adeak:
> The themes don't play well with private attributes. The leading underscore is stripped off in some functionality. Instead of trying to fix this, I think it makes sense to promote `before_close_callback` to be a first class option to `Plotter.show` and then we can make it also public in the Theme.

I think that makes sense, downstream users or libraries might even make use of the functionality. Or yours truly who keeps feeling too lazy to mess with camera positions for a nice interactive screenshot :smile:

MatthewFlamm:
> > I think it makes sense to promote `before_close_callback` to be a first class option to `Plotter.show`
> 
> I agree that makes sense, downstream users or libraries might even make use of the functionality. Or yours truly who keeps feeling too lazy to mess with camera positions for a nice interactive screenshot 😄

I used this as inspiration for the example in the documentation, which was previously lacking.

MatthewFlamm:
I think I am done cleaning up the `VerifyImageCache` class for this PR.  I think it would be better to pass in list of regex patterns for skipping and high variance tests.  This would be better for parametric tests, or all high variance tests could have similar naming.  regex is one area, one of many(!), that I haven't learned well. 
 I'm not confident for implementing here.  So, this class still isn't 100% ready for exposing downstream.

adeak:
If you have a specific plan with regex I can help with that. We don't have to do that now, just let me know if I can help.

akaszynski (dismissed):
This gets a solid LGTM.

Let's implement any additional changes needed for a plug-in as a follow-up. There's 76 files changed here. Let's merge this and move on.

pyvista/themes.py
```diff
@@ -2197,6 +2209,7 @@ def __repr__(self):
             'Anti-Aliasing': '_anti_aliasing',
             'Split sharp edges': '_split_sharp_edges',
             'Sharp edges feat. angle': '_sharp_edges_feature_angle',
+            'Before close callback': '_before_close_callback',
```
adeak, line 2212: If I understand correctly, this value is either `None` or some not very informative repr of a proper function or lambda. Is the purpose of this in the theme repr just to signal whether a callback is set, hoping for a meaningful function name when it is? I guess it wouldn't be easy with the current mechanism to do anything else.

tests/plotting/test_plotting.py
```diff
@@ -1224,10 +1238,10 @@ def test_scalars_by_name():
     plotter = pyvista.Plotter()
     data = examples.load_uniform()
     plotter.add_mesh(data, scalars='Spatial Cell Data')
-    plotter.show(before_close_callback=verify_cache_image)
+    plotter.show()
 
 
-def test_multi_block_plot():
+def test_multi_block_plot(request):
```
adeak, line 1244: The fixture doesn't seem to be used in the test.
```suggestion
def test_multi_block_plot():
```

adeak (commented):
I skimmed over the PR a bit, very nice work @MatthewFlamm. One minor suggestion and one question.

I also noticed that some of the new parametrised tests are identical when rendered. It would be nice to make this distinguish between parameters for useful regression testing, but I don't want to stall this PR (and the release) so I'll just open a low-hanging-fruit issue for this.

pyvista/themes.py
MatthewFlamm: I think the purpose is what you mentioned that it helps understand that it is set.  It could also help in debugging.  I'm not sure that it makes sense though.  I mainly implemented as everything else is already included, but this doesn't mean it is a good idea.

A more pressing question is that we serialize the values in some way when saving to file.  This probably fails, and this is untested.  We might need to strip the callback when we save to file.

MatthewFlamm:
> I also noticed that some of the new parametrised tests are identical when rendered. It would be nice to make these distinguish between parameters for useful regression testing, but I don't want to stall this PR (and the release) so I'll just open a low-hanging-fruit issue for this.

Thank you for making this issue.  It makes a lot of sense to have more focused PRs for those since. We have so many files added here that it will be hard to review.

pyvista/themes.py
MatthewFlamm: I fixed and added tests for serializing, it now always omits `before_close_callback`.  I'm not sure whether the repr should be modified though.

tests/test_theme.py
```diff
@@ -434,6 +434,25 @@ def test_load_theme(tmpdir, default_theme):
...
+    dark_theme = pyvista.themes.DarkTheme()
+
+    def fun(plotter):
+        pass
+
+    dark_theme.before_close_callback = fun
+    assert dark_theme != pyvista.themes.DarkTheme()
+    dark_theme.save(filename)
+
+    # fun ia stripped from the theme
```
adeak, line 448:
```suggestion
    # fun is stripped from the theme
```

adeak (approved):

pyvista/themes.py
adeak: I think we should keep it for now. It will be None in most cases and it might be informative in other cases. If for some reason it ends up being weird we can change or remove it later.

PR #3462, opened 2022-10-13
Set stereo type before turning on stereo

adeak:
According [to a question on Stack Overflow](https://stackoverflow.com/questions/74054264/how-can-i-show-the-3d-imaging-render-with-stereo) we might end up tripping a spurious VTK warning about trying to use an unsupported kind of stereo type. Setting the stereo type to anaglyph (the only mode we're supporting right now) before turning on stereo should make this warning go away.

As I noted in my answer on that page, I don't see this on my linux system, but it's possible that some VTK version and/or some platform-specific render window subclass has this behaviour.

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -1852,8 +1852,8 @@ class BasePlotter(PickingHelper, WidgetHelper):
 
         """
         if hasattr(self, 'ren_win'):
-            self.ren_win.StereoRenderOn()
             self.ren_win.SetStereoTypeToAnaglyph()
+            self.ren_win.StereoRenderOn()
 
     def disable_stereo_render(self):
         """Disable stereo rendering.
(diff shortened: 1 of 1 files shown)


pyvista/plotting/plotting.py
```diff
@@ -1852,8 +1852,8 @@ def enable_stereo_render(self):
 
         """
         if hasattr(self, 'ren_win'):
-            self.ren_win.StereoRenderOn()
             self.ren_win.SetStereoTypeToAnaglyph()
+            self.ren_win.StereoRenderOn()
 
     def disable_stereo_render(self):
         """Disable stereo rendering.
```
akaszynski, line 1859:
```suggestion
        """Disable anaglyph stereo rendering.
```

akaszynski (approved):
LGTM

adeak:
I intentionally left the disabling method's docstring untouched, because it literally just disables stereo, so even if we end up supporting other stereo modes, the `disable_stereo_render()` method can stay as-is. Not that it's doing any harm with the change.

tkoyama010 (approved):

PR #3457, opened 2022-10-13
Specify all optional parameters in #3347

puripant:
### Overview

All optional parameters in #3347 are specified in docstring. This should resolve #3347 

### Details
 
Per discussion in #3444 `optional` is used when the default value is `None` and `default=...` is specified when the default value is other than `None`.

PR #3472, opened 2022-10-14
Update CONTRIBUTING.rst with docstring guidelines

akaszynski:
Address [comments](https://github.com/pyvista/pyvista/pull/3457#issuecomment-1278021788) made by @adeak:

> > I'll also keep this in mind when adding any new features or updating docs (as well as remind others to do so).
> 
> Documenting this internal convention in our contributing guide would go a long way in replacing the mixed style in the long run, @akaszynski.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -252,6 +252,8 @@ Sample docstring follows:
 
         """
 
+        pass  # implementation goes here
+
 Note the following:
 
 * The parameter definition of ``generate_triangles`` uses ``default: False``,
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -203,16 +203,67 @@ Docstrings
 
 PyVista uses Python docstrings to create reference documentation for our Python
 APIs. Docstrings are read by developers, interactive Python users, and readers
-of our online documentation. This page describes how to write these docstrings
+of our online documentation. This section describes how to write these docstrings
 for PyVista.
 
-* PyVista follows the ``numpydoc`` style for its docstrings. Please follow the
-  `numpydoc Style Guide`_.
+PyVista follows the ``numpydoc`` style for its docstrings. Please follow the
+`numpydoc Style Guide`_ in all ways except for the following:
+
 * Be sure to describe all ``Parameters`` and ``Returns`` for all public
   methods.
 * We strongly encourage you to add an example section. PyVista is a visual
   library, so adding examples that show a plot will really help users figure
   out what individual methods do.
+* With optional parameters, use ``default: <value>`` instead of ``optional``
+  when the parameter has a default value instead of ``None``.
+
+Sample docstring follows:
+
+.. code:: python
+
+    def slice_x(self, x=None, generate_triangles=False):
+        """Create an orthogonal slice through the dataset in the x-direction.
+
+        Yields a MutliBlock dataset of the three slices.
+
+        Parameters
+        ----------
+        x : float, optional
+            The X location of the YZ slice.
+
+        generate_triangles : bool, default: False
+            If this is enabled, the output will be triangles. Otherwise the output
+            will be the intersection polygons.
+
+        Returns
+        -------
+        pyvista.PolyData
+            Sliced dataset.
+
+        Examples
+        --------
+        Slice the random hills dataset with one orthogonal plane.
+
+        >>> from pyvista import examples
+        >>> hills = examples.load_random_hills()
+        >>> slices = hills.slice_orthogonal(5, generate_triangles=False)
+        >>> slices.plot(line_width=5)
+
+        See :ref:`slice_example` for more examples using this filter.
+
+        """
+
+Note the following:
+
+* The parameter definition of ``generate_triangles`` uses ``default: False``,
+  and does not include the default in the docstring's "description" section.
+* There is a newline between each parameter. This is different than
+  ``numpydoc``'s documentation where there are no lines between parameter
+  doc strings.
+* This docstring contains both a returns section and examples section.
+* Returns section does not include the parameter name if it has a single return.
+  Multiple returns (not shown) should have descriptive parameter names.
+* The examples section references the "full example" if it exists.
 
 
 Branch Naming Conventions
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -271,20 +271,21 @@ Note the following:
 Depreciating Features or other Backwards-Breaking Changes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 When implementing backwards-breaking changes within PyVista, care must be taken
-to make the modifications non-breaking in one of the following ways:
+to give users the chance to adjust to any new changes. Any non-backwards
+compatible modifications should proceed through the following steps:
 
-* Retain the old behavior and issue a ``PyVistaDeprecationWarning`` indicating
+# Retain the old behavior and issue a ``PyVistaDeprecationWarning`` indicating
   the new interface you should use.
-* Retain the old behavior but raise a ``pyvista.core.errors.DeprecationError``
+# Retain the old behavior but raise a ``pyvista.core.errors.DeprecationError``
   indicating the new interface you must use.
-* Remove the old behavior.
+# Remove the old behavior.
 
 Whenever possible, PyVista developers should seek to have at least three minor
 versions of backwards compatibility to give users the ability to update their
 software and scripts.
 
-Here's an example of soft depreciation. Note the usage of both
-``PyVistaDeprecationWarning`` warning and the ``.. deprecated`` Sphinx
+Here's an example of a soft depreciation of a function. Note the usage of both
+the ``PyVistaDeprecationWarning`` warning and the ``.. deprecated`` Sphinx
 directive.
 
 .. code:: python
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -222,7 +222,7 @@ Sample docstring follows:
 .. code:: python
 
     def slice_x(self, x=None, generate_triangles=False):
-        """Create an orthogonal slice through the dataset in the x-direction.
+        """Create an orthogonal slice through the dataset in the X direction.
 
         Yields a MutliBlock dataset of the three slices.
 
@@ -232,8 +232,8 @@ Sample docstring follows:
             The X location of the YZ slice.
 
         generate_triangles : bool, default: False
-            If this is enabled, the output will be triangles. Otherwise the output
-            will be the intersection polygons.
+            If this is enabled, the output will be all triangles. Otherwise the output
+            will consist of the intersection polygons.
 
         Returns
         -------
@@ -246,7 +246,7 @@ Sample docstring follows:
 
         >>> from pyvista import examples
         >>> hills = examples.load_random_hills()
-        >>> slices = hills.slice_orthogonal(5, generate_triangles=False)
+        >>> slices = hills.slice_x(5, generate_triangles=False)
         >>> slices.plot(line_width=5)
 
         See :ref:`slice_example` for more examples using this filter.
@@ -258,12 +258,15 @@ Note the following:
 * The parameter definition of ``generate_triangles`` uses ``default: False``,
   and does not include the default in the docstring's "description" section.
 * There is a newline between each parameter. This is different than
-  ``numpydoc``'s documentation where there are no lines between parameter
-  doc strings.
-* This docstring contains both a returns section and examples section.
-* Returns section does not include the parameter name if it has a single return.
-  Multiple returns (not shown) should have descriptive parameter names.
-* The examples section references the "full example" if it exists.
+  ``numpydoc``'s documentation where there are no empty lines between parameter
+  docstrings.
+* This docstring also contains a returns section and an examples section.
+* The returns section does not include the parameter name if the function has 
+  a single return value. Multiple return values (not shown) should have 
+  descriptive parameter names for each returned value, in the same format as
+  the input parameters.
+* The examples section references the "full example" in the gallery if it 
+  exists.
 
 
 Branch Naming Conventions
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -268,7 +268,7 @@ Note the following:
   exists.
 
 
-Depreciating Features or other Backwards-Breaking Changes
+Deprecating Features or other Backwards-Breaking Changes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 When implementing backwards-breaking changes within PyVista, care must be taken
 to give users the chance to adjust to any new changes. Any non-backwards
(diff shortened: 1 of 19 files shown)


CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
-  `numpydoc Style Guide`_.
+PyVista follows the ``numpydoc`` style for its docstrings. Please follow the
+`numpydoc Style Guide`_ in all ways except for the following:
+
 * Be sure to describe all ``Parameters`` and ``Returns`` for all public
   methods.
 * We strongly encourage you to add an example section. PyVista is a visual
   library, so adding examples that show a plot will really help users figure
   out what individual methods do.
+* With optional parameters, use ``default: <value>`` instead of ``optional``
```
adeak, line 217: In https://github.com/pyvista/pyvista/pull/3457/files we seemed to use `default=<value>`. Either works for me, we should just make sure we choose one on purpose.

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+  when the parameter has a default value instead of ``None``.
+
+Sample docstring follows:
```
adeak, line 220: Good call.

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+
+.. code:: python
+
+    def slice_x(self, x=None, generate_triangles=False):
+        """Create an orthogonal slice through the dataset in the x-direction.
```
adeak, line 225: Nitpick, but this example should be exemplary.
```suggestion
        """Create an orthogonal slice through the dataset in the X direction.
```

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+
+        Yields a MutliBlock dataset of the three slices.
```
adeak, line 227:
```suggestion
        Yields a ``MultiBlock`` dataset of the three slices.
```
Or maybe even showcase intersphinx?
```suggestion
        Yields a :class:`MultiBlock <pyvista.MultiBlock>` dataset of the three slices.
```
(assuming this is within the usual length.)

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+
+        Parameters
+        ----------
+        x : float, optional
+            The X location of the YZ slice.
```
adeak, line 232: For optional params we should ideally always explain the default behaviour.

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+
+        generate_triangles : bool, default: False
+            If this is enabled, the output will be triangles. Otherwise the output
+            will be the intersection polygons.
```
adeak, line 236:
```suggestion
            If this is enabled, the output will be all triangles. Otherwise the output
            will consist of the intersection polygons.
```

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+
+        Returns
+        -------
+        pyvista.PolyData
```
adeak, line 240: `PolyData` or `MultiBlock`?

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+        See :ref:`slice_example` for more examples using this filter.
+
+        """
+
+Note the following:
+
+* The parameter definition of ``generate_triangles`` uses ``default: False``,
+  and does not include the default in the docstring's "description" section.
+* There is a newline between each parameter. This is different than
+  ``numpydoc``'s documentation where there are no lines between parameter
```
adeak, line 261:
```suggestion
  ``numpydoc``'s documentation where there are no empty lines between parameter
```

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+  doc strings.
+* This docstring contains both a returns section and examples section.
```
adeak, line 263:
```suggestion
  docstrings.
* This docstring also contains a returns section and an examples section.
```

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+* Returns section does not include the parameter name if it has a single return.
+  Multiple returns (not shown) should have descriptive parameter names.
+* The examples section references the "full example" if it exists.
```
adeak, line 266: Not sure this is the best possible phrasing, feel free to adjust.
```suggestion
* The returns section does not include the parameter name if the function has 
  a single return value. Multiple return values (not shown) should have 
  descriptive parameter names for each returned value, in the same format as
  the input parameters.
* The examples section references the "full example" in the gallery if it 
  exists.
```

adeak (commented):
Thanks for adding this. I've picked many nits because this is something that we expect users to follow, so we should be as precise as possible.

CONTRIBUTING.rst
MatthewFlamm: I think this whole line can be deleted.  This docstring stub is presumably referencing a slice operation in the x-normal direction, i.e. it only yields a single PolyData slice and could be described more in the `Returns` section if needed.

CONTRIBUTING.rst
```diff
@@ -203,16 +203,67 @@ Docstrings
...
+        pyvista.PolyData
+            Sliced dataset.
+
+        Examples
+        --------
+        Slice the random hills dataset with one orthogonal plane.
+
+        >>> from pyvista import examples
+        >>> hills = examples.load_random_hills()
+        >>> slices = hills.slice_orthogonal(5, generate_triangles=False)
```
MatthewFlamm, line 249:
```suggestion
        >>> slices = hills.slice_x(5, generate_triangles=False)
```

CONTRIBUTING.rst
MatthewFlamm: The current code base has more `default: <value>`.  I found 164 hits with this. 
 This includes some false positives, but not as many as I would have suspected.  I found 36 hits for `default=<value>`, with also some false positives.  Before #3457, it would have leaned even heavier towards the colon usage.  Although, I'm fine either way as well.

adeak: I'd also prefer a colon here, ignoring all legacy considerations.

akaszynski: Agree with the colon. Let's go with that and consider a refactor. At least it's low hanging.

CONTRIBUTING.rst
akaszynski: This is just a fake function, so we can just go with PolyData.

CONTRIBUTING.rst
```diff
@@ -203,16 +203,144 @@ Docstrings
...
+  descriptive parameter names for each returned value, in the same format as
+  the input parameters.
+* The examples section references the "full example" in the gallery if it 
+  exists.
+
+
+Depreciating Features or other Backwards-Breaking Changes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+When implementing backwards-breaking changes within PyVista, care must be taken
+to make the modifications non-breaking in one of the following ways:
```
akaszynski, line 274:
```suggestion
When implementing backwards-breaking changes within PyVista, care must be taken
to give users the chance to adjust to any new changes. Any non-backwards
compatible modifications should proceed through the following steps:
```

CONTRIBUTING.rst
```diff
@@ -203,16 +203,144 @@ Docstrings
...
+
+* Retain the old behavior and issue a ``PyVistaDeprecationWarning`` indicating
+  the new interface you should use.
+* Retain the old behavior but raise a ``pyvista.core.errors.DeprecationError``
+  indicating the new interface you must use.
+* Remove the old behavior.
```
akaszynski, line 280:
```suggestion
# Retain the old behavior and issue a ``PyVistaDeprecationWarning`` indicating
  the new interface you should use.
# Retain the old behavior but raise a ``pyvista.core.errors.DeprecationError``
  indicating the new interface you must use.
# Remove the old behavior.
```

CONTRIBUTING.rst
```diff
@@ -203,16 +203,144 @@ Docstrings
...
+
+Whenever possible, PyVista developers should seek to have at least three minor
+versions of backwards compatibility to give users the ability to update their
+software and scripts.
+
+Here's an example of soft depreciation. Note the usage of both
```
akaszynski, line 286:
```suggestion
Here's an example of a soft depreciation of a function. Note the usage of both
```

CONTRIBUTING.rst
```diff
@@ -203,16 +203,144 @@ Docstrings
...
+``PyVistaDeprecationWarning`` warning and the ``.. deprecated`` Sphinx
```
akaszynski, line 287:
```suggestion
Here's an example of a soft depreciation of a function. Note the usage of both
the ``PyVistaDeprecationWarning`` warning and the ``.. deprecated`` Sphinx
```

CONTRIBUTING.rst
```diff
@@ -203,16 +203,145 @@ Docstrings
...
+* This docstring also contains a returns section and an examples section.
+* The returns section does not include the parameter name if the function has 
+  a single return value. Multiple return values (not shown) should have 
+  descriptive parameter names for each returned value, in the same format as
+  the input parameters.
+* The examples section references the "full example" in the gallery if it 
+  exists.
+
+
+Depreciating Features or other Backwards-Breaking Changes
```
adeak, line 271:
```suggestion
Deprecating Features or other Backwards-Breaking Changes
```

CONTRIBUTING.rst
```diff
@@ -203,16 +203,145 @@ Docstrings
...
+   the new interface you should use.
+#. Retain the old behavior but raise a ``pyvista.core.errors.DeprecationError``
+   indicating the new interface you must use.
+#. Remove the old behavior.
+
+Whenever possible, PyVista developers should seek to have at least three minor
+versions of backwards compatibility to give users the ability to update their
+software and scripts.
+
+Here's an example of a soft depreciation of a function. Note the usage of both
```
adeak, line 287:
```suggestion
Here's an example of a soft deprecation of a function. Note the usage of both
```

CONTRIBUTING.rst
```diff
@@ -258,12 +257,91 @@ Note the following:
...
```
adeak, line 287:
```suggestion
Here's an example of a soft deprecation of a function. Note the usage of both
```

CONTRIBUTING.rst
```diff
@@ -258,12 +257,91 @@ Note the following:
...
+        b : float
+            Second term to add.
+
+        Returns
+        -------
+        float
+            Sum of the two inputs.
+
+        """
+        # depreciated 0.37.0, convert to error in 0.40.0, remove 0.41.0
```
adeak, line 313:
```suggestion
        # deprecated 0.37.0, convert to error in 0.40.0, remove 0.41.0
```

CONTRIBUTING.rst
```diff
@@ -258,12 +257,91 @@ Note the following:
...
+        PyVistaDeprecationWarning(
+            '`addition` has been deprecated. Use pyvista.add instead'
+        )
+        add(a, b)
+
+
+    def add(a, b):
+        """Add two numbers.
```
adeak, line 321: I suspect the ellipsis was meant to be a placeholder for the docstring, but we need the ending triple quotes somewhere because this is breaking syntax highlighting at least on github. But if we put the """ after the ellipsis we'd probably have to add another ... or `pass` to get a syntactically valid example... might be simpler to have a docstring stub instead:
```suggestion
        """Add two numbers."""
```

pyvista/utilities/errors.py
```diff
@@ -260,34 +259,73 @@ def __repr__(self):
...
-        text_width : int, optional
-            The text width for non-HTML display modes
+    sort : bool, default: False
+        Alphabetically sort the packages.
 
-        sort : bool, optional
-            Alphabetically sort the packages
+    gpu : bool, default: True
+        Gather information about the GPU. Defaults to ``True`` but if
+        experiencing renderinng issues, pass ``False`` to safely generate a
```
adeak, line 281:
```suggestion
        experiencing rendering issues, pass ``False`` to safely generate a
```

setup.py
```diff
@@ -16,7 +16,6 @@
     'numpy',
     'imageio',
     'pillow',
-    'appdirs',
```
adeak, line 19: Is this removal related?

CONTRIBUTING.rst
```diff
@@ -203,16 +203,145 @@ Docstrings
...
+
+        ...
```
akaszynski, line 323:
```suggestion
        pass  # implementation goes here
```

setup.py
akaszynski: This should have happened in https://github.com/pyvista/pyvista/pull/3218

CONTRIBUTING.rst
```diff
@@ -268,7 +268,7 @@ Note the following:
   exists.
 
 
-Depreciating Features or other Backwards-Breaking Changes
+Deprecating Features or other Backwards-Breaking Changes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
adeak, line 272:
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

pyvista/utilities/helpers.py
```diff
@@ -1306,15 +1306,15 @@ class ProgressMonitor:
...
+    message : str, default: ""
         Message to display in the progress bar.
 
     scaling : float, optional
         Unused keyword argument.
 
     """
 
-    def __init__(self, algorithm, message="", scaling=100):
+    def __init__(self, algorithm, message="", scaling=None):
```
adeak, line 1317: If I'm reading things correctly, `scaling` isn't even used in this function. Why the change then, and can we just dump it? Or is this here for backwards compat?

pyvista/utilities/regression.py
```diff
@@ -64,12 +64,12 @@ def compare_images(im1, im2, threshold=1, use_vtk=True):
...
         window, or ``vtkImageData``.
 
-    threshold : int, optional
+    threshold : int, default: 1
         Threshold tolerance for pixel differences.  This should be
         greater than 0, otherwise it will always return an error, even
         on identical images.
 
-    use_vtk : bool, optional
+    use_vtk : bool, bool: True
```
adeak, line 72:
```suggestion
    use_vtk : bool, default: True
```

pyvista/utilities/helpers.py
adeak: Ah, it's there in the docstring. Sorry. I understand then. Perhaps we should remove it eventually...

akaszynski: It's correct that it's not used. I'm assuming that it's been kept here for backwards compatibility, so probably the right thing to do is to warn then raise during a later release.

It's a bit out of the scope of this PR and I should have left it alone.

CONTRIBUTING.rst
```diff
@@ -203,16 +203,145 @@ Docstrings
...
+        Slice the random hills dataset with one orthogonal plane.
+
+        >>> from pyvista import examples
+        >>> hills = examples.load_random_hills()
+        >>> slices = hills.slice_x(5, generate_triangles=False)
+        >>> slices.plot(line_width=5)
+
+        See :ref:`slice_example` for more examples using this filter.
+
+        """
```
adeak, line 253:
```suggestion
        """

        pass  # implementation goes here
```

adeak (approved):
LGTM now, thanks!

Issue #3268, opened 2022-09-04
Open an Issue when linkcheck fails


laysauchoa:
### Describe the feature you would like to be added.

**Problem**
Recently, we have added linkcheck to run periodically, every week. However, someone has to go and look for the failed build and do the fixes. This could go unseen or unnoticed.

**Proposed solution**
Automatically open an Issue when the ``linkcheck`` fails. So we can better monitor it and fix it. We can also assign to some of the maintainers

I will be happy to work on that

tkoyama010:
It is  a great idea to have auto Pull Requests. We always welcome your contribution.

laysauchoa:
@tkoyama010 this would be more like an issue saying that it failed, not the pull request with the fix.

tkoyama010:
I'm sorry. I misunderstood your suggestion (I was imagining something like dependabot.). In this case, I also think Issue would be better.

tkoyama010:
First of all, we don't know the correct URL automatically :sweat_smile:

banesullivan:
@laysauchoa, I think this is a great idea. Would you be willing to take this on?

I believe there are GitHub Actions that can open an issue like this easily.

I would set up the CI on a cronjob which:

1. runs link check and captures output as text
2. if fails, opens the issue with error text as body of the issue

adeak:
> @laysauchoa, I think this is a great idea. Would you be willing to take this on?
> 
> I believe there are GitHub Actions that can open an issue like this easily.
> 
> I would set up the CI on a cronjob which:
> 
> 1. runs link check and captures output as text
> 2. if fails, opens the issue with error text as body of the issue

@laysauchoa humbly didn't link to it, but we already have https://github.com/pyvista/pyvista/pull/3166 thanks to them :) this machinery could be upgraded to open the issue.

banesullivan:
Ah, thanks for pointing that out @adeak, and thank you @laysauchoa for already implementing most of this!

laysauchoa:
Opening the issue works fine but I tried to catch the output from previous steps to add to the body of the issue and does not work because the GitHub action that I use for the issue does not support such a long output. I am trying to find an alternative for that, but so far nothing worked.

banesullivan:
Do you have a prototype of the code pushed somewhere? I'd be happy to help

laysauchoa:
[...]

PR #3475, opened 2022-10-15
Create Issue if linkcheck fails

laysauchoa:
Linkcheck runs weekly. This creates
an issue if the linkcheck fails.

The issue should contain the broken
links logs.

Fixes https://github.com/pyvista/pyvista/issues/3268

### Details

- < feature1 or bug1 description >
- < feature2 or bug2 description >


.github/ISSUE_TEMPLATE/linkcheck_report.md
```diff
@@ -0,0 +1,12 @@
+---
+title: 🐛 Weekly linkcheck failed
+about: Use this template for checking broken links
+name: Weekly linkcheck failed
+labels: Maintanance, bug
```
tkoyama010, line 5:
```suggestion
labels: maintanance, bug
```

tkoyama010 (changes requested):

.github/ISSUE_TEMPLATE/linkcheck_report.md
laysauchoa: There is a small typo so I had to add in a new commit.

.github/ISSUE_TEMPLATE/linkcheck_report.md
```diff
@@ -0,0 +1,12 @@
+---
+title: 🐛 Weekly linkcheck failed
+about: Use this template for checking broken links
```
adeak, line 3: Will this issue template show up as an option when someone tries to open an issue on github? I think we should avoid that.

.github/ISSUE_TEMPLATE/linkcheck_report.md
```diff
@@ -0,0 +1,12 @@
...
+name: Weekly linkcheck failed
+labels: maintenance, bug
+---
+
+Links are checked weekly to make sure the documentation links areg working. Last check on {{ date | date('dddd, MMM Do YYYY, hh:mm A') }} failed.
```
adeak, line 8:
```suggestion
Links are checked weekly to make sure the documentation links are working. Last check on {{ date | date('dddd, MMM Do YYYY, hh:mm A') }} failed. 
```
And is there documentation for the date field syntax in there? Just so I can check that too.

.github/ISSUE_TEMPLATE/linkcheck_report.md
```diff
@@ -0,0 +1,12 @@
+---
+title: 🐛 Weekly linkcheck failed
+about: Use this template for checking broken links
+name: Weekly linkcheck failed
+labels: maintenance, bug
+---
+
+Links are checked weekly to make sure the documentation links areg working. Last check on {{ date | date('dddd, MMM Do YYYY, hh:mm A') }} failed. 
+
+Partial logs output can be found here:
```
adeak, line 10: I think either "partial log output" or "partial logs" (with no "output") would be a bit better here.
```suggestion
Partial logs can be found here:
```
In what sense are they partial?

.github/workflows/linkcheck.yml
```diff
@@ -41,4 +41,15 @@ jobs:
         run: pip install -r requirements_docs.txt
 
       - name: LinkCheck
-        run: make -C doc linkcheck SPHINXOPTS="-W --keep-going --color"
+        run: |
+          make -C doc linkcheck SPHINXOPTS="-W --keep-going --color" > logs.txt || echo "OUTPUT=$(grep -E "broken" logs.txt | tr '\n' '@' | sed 's/@/<br>/g' | sed 's/broken/*\*`broken`\*\*/g' | sed $'s/\e\\[[0-9;:]*[a-zA-Z]//g')" >> $GITHUB_ENV
```
adeak, line 45: Is there a way to break this huge line with continuation characters?

adeak (commented):
A few comments. I'll also take a look at the log parsing shell script later.

I won't be able to review the CI bits, so if someone CI-savvy has the time it would be nice to get this merged (post review) in October to make it count for hacktoberfest.

.github/workflows/linkcheck.yml
adeak, line 45: Some remarks about this shell command as promised:
1. if I understand correctly, the last bit strips away colour escapes. So why not use `--no-color` instead of `--color` in the `make` call so that we don't have to strip it away later?
2. we don't need to escape the asterisks on the "right" side of `sed 's/broken/*\*`broken`\*\*/g'` (in fact one escape is missing). In the replacement string `*` is not special so these can just be `**`broken`**`.
3. we can probably join two (or more) consecutive `sed` calls with a single `sed` call with two (or more) `-e` switches, though I haven't tested this one.

Do you perhaps have a screenshot of how this will get rendered? Assuming you tested this somewhere.

laysauchoa:
Thanks for the review. I will come back to this soon!

banesullivan:
Closing due to staleness. Apologies we weren't able to help you take this over the finish line @laysauchoa

PR #3512, opened 2022-10-25
Patch default scalars plotting edge case

akaszynski:
Ran into an edge case today where a dataset with default plotting scalars will not be overridden with new data. This is due to a check where we see if an array with scalars already exists in the dataset before adding it.

```py
import numpy as np
import pyvista as pv

mesh = pv.Sphere()
mesh['Data'] = np.zeros(mesh.n_points)
mesh.plot(scalars=range(mesh.n_points))
```

### DIFF
--- a/pyvista/__init__.py
+++ b/pyvista/__init__.py
@@ -73,3 +73,6 @@ FLOAT_FORMAT = "{:.3e}"
 
 # Serialization format to be used when pickling `DataObject`
 PICKLE_FORMAT = 'xml'
+
+# Name used for unnamed scalars
+DEFAULT_SCALARS_NAME = 'Data'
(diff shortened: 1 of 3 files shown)


tkoyama010 (commented):
LGTM for changes. Do we need to add a test for this?

pyvista/__init__.py
```diff
@@ -73,3 +73,6 @@
 
 # Serialization format to be used when pickling `DataObject`
 PICKLE_FORMAT = 'xml'
+
+# Name used for unnamed scalars
```
akaszynski, line 77:
```suggestion
# The name used for unnamed scalars
```

banesullivan (approved):

akaszynski:
> LGTM for changes. Do we need to add a test for this?

Probably not. It's enough of an edge case that we probably don't have to, but if this crops up again I will.

Issue #1012, opened 2020-11-13
Pick cells in Multi-Windows setting


bo-wu:
## Description
Hello authors,

Thanks a lot for the excellent library. I am new to PyVista, and I want to plot objects in multi-windows and pick points or cells. I succeed in plotting, but in the multi-windows setting, picking only works for the last subplot window, since it is the active window. Is there any way to do that for all other windows, not just the last one? How should I activate other windows?

![Screenshot_2020-11-14_19-55-45](https://user-images.githubusercontent.com/8759514/99139075-884ef280-2670-11eb-810d-4fa64d233938.png)

In the image, I picked in the second window, but the picked results are shown in the last window 

```
    mesh = pv.read('./cube.obj')
    p = pv.Plotter(shape='1|2', window_size=(1600, 1200), notebook=False)
    img1 = pv.read('./face1.jpg')
    img2 = pv.read('./face2.jpeg')
    
    p.subplot(0)
    p.add_mesh(mesh, show_edges=False, color='w')
    picker = MeshSurfacePicker(p, mesh, 0)
    p.track_click_position(picker, side='right')
    p.add_text('Press s to save selected result\nPress q to quit')

    p.subplot(1)
    p.add_mesh(img1, rgb=True)
    p.enable_cell_picking(img1, callback=picked_cells)
    p.view_xy()
    # p.enable_parallel_projection()

    p.subplot(2)
    p.add_mesh(img2, rgb=True)
    p.enable_cell_picking(img2)
    p.view_xy()

    p.show()

```

PR #3515, opened 2022-10-26
Fix issues with cell picking

### LINKED ISSUE #3317: Cell Picking Enabled Only for Last Actor Added to Plotter

Following the [Enable Cell Picking Example](https://docs.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.enable_cell_picking.html?highlight=cell+picking#pyvista.Plotter.enable_cell_picking), I found only the last actor added to the plotter is enabled for cell picking, with or without `pickable=True`).

This smells of #3180 ...

```python
import pyvista as pv

mesh = pv.Sphere(center=(1, 0, 0))
cube = pv.Cube()
p = pv.Plotter()

p.add_mesh(mesh, pickable=True)
p.add_mesh(cube, pickable=True)  # Only this actor is pickable
p.enable_cell_picking(color='blue')

p.show()
```

### System Information

```shell
  Date: Tue Sep 13 19:49:34 2022 Pacific Daylight Time

                OS : Windows
            CPU(s) : 96
           Machine : AMD64
      Architecture : 64bit
               RAM : 190.7 GiB
       Environment : Python
      GPU Renderer : Quadro RTX 8000/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 516.40

  Python 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64
  bit (AMD64)]

           pyvista : 0.37.dev0
               vtk : 9.1.0
             numpy : 1.23.3
           imageio : 2.21.2
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.3
         pyvistaqt : 0.9.0
           IPython : 8.5.0
              tqdm : 4.64.1
            meshio : 5.3.4
```

### Screenshots

CellPickingBug

banesullivan:
Resolves #3317, cc @adam-grant-hendry

This fixes an issue where cell picking would only work for the last mesh added before enabling picking.

Further, this changes the API to be consistent with the other `enable_*_picking` methods -- none of the other methods accept a mesh as an argument, so I decided to remove this from `enable_cell_picking`. I've updated the docstring to detail how `enable_cell_picking` should work (by specifying `pickable` when adding a mesh to the plotter).

Further, I looked into an older issue #1012 while doing this to see if we could improve picking across subplots. I think it would take *significant* work to handle subplot picking properly and I'm inclined to warn users against doing any picking at all with subplots at this time.

This also improves the tests for cell picking to capture the issue in #3317

### DIFF
--- a/pyvista/plotting/picking.py
+++ b/pyvista/plotting/picking.py
@@ -392,6 +392,12 @@ class PickingHelper:
                     cids = pyvista.convert_array(selection_node.GetSelectionList())
                     actor = selection_node.GetProperties().Get(_vtk.vtkSelectionNode.PROP())
 
+                    # TODO: this is too hacky - find better way to avoid non-dataset actors
+                    if not actor.GetMapper() or not hasattr(
+                        actor.GetProperty(), 'GetRepresentation'
+                    ):
+                        continue
+
                     # if not a surface
                     if actor.GetProperty().GetRepresentation() != 2:  # pragma: no cover
                         logging.warning(
--- a/pyvista/plotting/picking.py
+++ b/pyvista/plotting/picking.py
@@ -209,7 +209,6 @@ class PickingHelper:
 
     def enable_cell_picking(
         self,
-        mesh=None,
         callback=None,
         through=True,
         show=True,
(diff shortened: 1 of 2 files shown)


akaszynski (approved):
LGTM. Thanks!

Issue #1897, opened 2021-11-29
Active scalars regression


banesullivan:
**Describe the bug, what's wrong, and what you expect:**

It's my opinion that the following is a regression

#1593 introduced a regression where a mesh can have arrays but no active scalars

Take a look at my comment: https://github.com/pyvista/pyvista/pull/1881#issuecomment-981981603

-----

**To Reproduce**

```py
>>> from pyvista import examples
>>> points = examples.download_sparse_points()
>>> points
PolyData (0x7fbc79419fa0)
  N Cells:	1
  N Points:	16
  X Bounds:	-1.500e+01, 1.500e+01
  Y Bounds:	5.000e+00, 3.500e+01
  Z Bounds:	2.327e+00, 1.102e+01
  N Arrays:	1

>>> points.threshold()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/bane/Software/pyvista/pyvista/pyvista/core/filters/data_set.py", line 842, in threshold
    raise ValueError('No arrays present to threshold.')
ValueError: No arrays present to threshold.
>>>
```

Even though there is 1 array.

-----

**System Information:**
  on main and we're seeing it in the docs

banesullivan:
I can understand the opposite view of this quite well, where you should always be explicit about which scalars to be processing/visualizing but one of the core tenants of PyVista's design is to be simple and easy to use.

From the `CONTRIBUTING.md` guide:

> Make it intuitive. PyVista's goal is to create an intuitive and easy to use interface back to the VTK library. ...

IMO, thresholding (or any filter) a mesh with one array should work without specifying the scalars to use as it intuitively, what other scalars array would I want to threshold on?

banesullivan:
Either way, the error message `ValueError: No arrays present to threshold.` is erroneous and I imagine this is present in many other places

adeak:
We can fix part of this. We now have machinery that can add data arrays both setting them active and not. It seems that (even if we don't implement auto-activation of unambiguous arrays) all the meshes created by _us_ (in this case in `download_sparse_points()`) should have their scalars activated.

And a mostly side note: I wonder if the design choice of always activating one interferes with the concept of active scalars/t_coords/normals/etc. which was also recently added (probably in the same PR, in fact).

MatthewFlamm:
I think we need to be consistent with active attributes, for example, if we set default scalars, we should set also default vectors, normals, etc.  Some filters operate on those.
[...]

PR #3535, opened 2022-11-01
Do not set scalars as active when adding

banesullivan:
### Overview

Helps address #1897 and should solve https://github.com/pyvista/pyvista/discussions/3486

### Details

These changes prevent a scenario where a mesh may already have active scalars and adding a new array automatically gets set as active. #1897 and #3486 outline several places where this behavior has unintended consequences

The behavior is as follows when adding a new scalar array (point or cell data):

- If **no** array is currently active, the added array becomes active
- If an array is currently active, the added array is **not** set as active

### Example

```py
import pyvista as pv

# Issue arrises with CELL data
mesh = pv.Wavelet().ptc()

p = pv.Plotter(notebook=0)
p.add_mesh_clip_plane(mesh, crinkle=True, normal=(-1, 1, -1))
p.show()
```

| `main` | this branch |
|---|---|
|Screen Shot 2022-11-01 at 9 12 25 AM | Screen Shot 2022-11-01 at 9 12 40 AM |


banesullivan:
These changes conflict with the `test_active_scalars_remain` in `tests/test_plotter.py` -- seeking feedback from @pyvista/developers

tests/test_plotter.py
```diff
@@ -123,6 +123,8 @@ def test_remove_scalars_single(sphere, hexbeam):
 
 def test_active_scalars_remain(sphere, hexbeam):
     """Ensure active scalars remain active despite plotting different scalars when copy_mesh=True."""
+    sphere.clear_data()
+    hexbeam.clear_data()
```
banesullivan, line 127: Please review

MatthewFlamm:
I think there should be a comprehensive proposal here before moving forward.  Otherwise we might want to change course again with another breaking change.  This should be documented somewhere.  I can think of two possibilities here for discussion, here I just focus on scalars:

1. Never automatically set active scalars when assigning data, try to adhere to this philosophy as much as possible in internal code.  Document when not possible.
2. Automatically set active scalars when none currently exists.  <- current PR.
  a. When filters or internal code assigns data, and no data currently exists, set as active.  This is consistent, but I think will be surprising to users with a side effect.  This happens throughout the entire code base.
  b. When filters or internal code assigns data, and no data currently exists, do not set as active.  Document when not possible.  We could consider a decorator to store active scalars then reset active scalars after a filter or method operation to ensure this.

I like either option 1 or option 2b.  

What about other array types? For example, what about when a (N, 3) array is saved?  Should this be an active scalar when no other scalar exists or should it be active vector similarly, or both?  What about Normals, Tensors, etc?  We don't support all of these the same today, but it is worth thinking about to avoid further breaking changes like this.  I think scalars, vectors, and Normals are the most common usages.  I would vote for implementing this for the most common generic data, scalars and vectors.  Normals are a special data type.  Another generic data type is tensor, but do we support it?

I propose the logic: if there is no active scalar/vector/tensor set and data is saved that is of suitable shape, set it as active.  This would result in:

- If there is no data, and a (N,3) array is saved, it will be set as both active scalar and vector.
- If there is an active scalar, and a (N,3) is saved, it will be set as only the active vector.

akaszynski:
> I like either option 1...

> Never automatically set active scalars when assigning data, try to adhere to this philosophy as much as possible in internal code. Document when not possible.

I genuinely prefer this as this follows "Explicit is better than implicit"  where we minimize side effects as much as possible. There's of course a lack of convenience here, which is why we proposed making them active (and keeping them active) when using `set_array` in something like:

```py
cube = pv.Cube()
cube['scalars'] = cube.points[:, 0]
```

This makes it really convenient when plotting with `cube.plot()`, and I'm happy with this behavior, that users can easily override with `plot(color=<colorlike>)`.

This leaves us with either 2a or 2b

> ... option 2b.
> When filters or internal code assigns data, and no data currently exists, do not set as active. Document when not possible. We could consider a decorator to store active scalars then reset active scalars after a filter or method operation to ensure this.

This seems like a compromise between "assign always" and "assign never". When a user sets data using `__set_item__` it's fairly clear that they're adding data and already expecting some sort of side effect of at least adding another array to the point/cell data. However, this assumption (to me) starts to break down when using filters when it's not clear which arrays are being added and why some are being made active. However, using a filter is still expected to operate on the data (somehow) and filters like `compute_normals` can be reasonably expected to set scalars to active.

That just leads us with 2a:

> if there is no active scalar/vector/tensor set and data is saved that is of suitable shape, set it as active.

I prefer this behavior, the behavior proposed to keep (and really patch) in this PR. However, we need to consider @MatthewFlamm's comments regarding vector data:

> If there is no data, and a (N,3) array is saved, it will be set as both active scalar and vector.
If there is an active scalar, and a (N,3) is saved, it will be set as only the active vector.

Fully agree there. Right now we only address scalars. I recommend adding this in a follow-up PR and keeping this PR for the release since it's a bug fix of (I hope) agreed upon behavior.

larsoner:
Agreed with @akaszynski's summary + plan :+1:

akaszynski (approved):
Given the general consensus, I'm recommending approval for this. Summarizing:
- Agree with @banesullivan's approach of setting active scalars based on the "first" added if there are no active ones. It's a trade-off between usability and having an explicit API.
- We need a follow-up PR that does the same thing for vectors.

Ignoring the MNE CI failure. I'm sure @larsoner is aware and we passed before the jupyter warnings started causing the issue.

MatthewFlamm:
I'm okay with this PR.  Im not convinced that we should greedily activate generated data as much as possible however and this can wait for a future conversation. For example, it doesn't make sense to me to make the data resulting from `compute_normals` active scalars or vectors.

I suppose I'm proposing that we should set active X when the data being generated is the primary expected outcome of an operation.  Here X is whatever type is relevant, e.g. scalars or normals or...  If the data generated is ancillary, then do not make it active.  For example, in some cases the original ID of a point/cell may be stored during a geometrical operation.  Does it makes sense to make these active scalars?

banesullivan:
I think we're all on the same page. This PR can be the first step to improving active scalars (and vectors/tensors to follow). 

IMO, I think this PR is ready to merge as is given @akaszynski's plan above

akaszynski:
Let's merge and open an issue with "the plan".

Issue #3522, opened 2022-10-29
Lights unexpectedly removed after `clear()`


adam-grant-hendry:
### Describe the bug, what's wrong, and what you expected.

I happened upon this rather accidentally while working in `ipython`. After adding the wrong mesh to a plotter, I cleared the plotter and then added the right one. To my surprise, the shadow layer appears to have disappeared.

I am assuming the culprit begins here, but am not sure:

https://github.com/pyvista/pyvista/blob/dd0809210275a5fa1e1da0931a71690f2a7cc929/pyvista/plotting/renderers.py#L362-L367

### Example Plotting before a `clear()`

```python
import pyvista as pv

p = pv.Plotter()
cylinder = pv.Cylinder()

p.add_mesh(cylinder)

p.show()
```

### Before Screenshot

![image](https://user-images.githubusercontent.com/59346180/198754361-4689fe0b-024e-449f-a733-1d51f441cc12.png)

### Example Plotting after a `clear()`

```python
import pyvista as pv

p = pv.Plotter()
cyl1 = pv.Cylinder()
cyl2 = pv.Cylinder()

p.add_mesh(cyl1)
p.clear()
p.add_mesh(cyl2)

p.show()
```

### After Screenshot

![image](https://user-images.githubusercontent.com/59346180/198754488-449285cf-8cc5-4801-ba01-19de7fbe6b64.png)

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Fri Oct 28 17:56:25 2022 Pacific Daylight Time

                OS : Windows
            CPU(s) : 96
           Machine : AMD64
      Architecture : 64bit
               RAM : 190.7 GiB
       Environment : IPython
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : Quadro RTX 8000/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 516.40

  Python 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64    
  bit (AMD64)]

           pyvista : 0.36.1
               vtk : 9.2.2
             numpy : 1.23.4
           imageio : 2.22.2
           appdirs : 1.4.4
            scooby : 0.7.0
        matplotlib : 3.6.1
         pyvistaqt : 0.9.0
           IPython : 8.5.0
             scipy : 1.9.3
              tqdm : 4.64.1
            meshio : 5.3.4
         pythreejs : Version unknown
[...]

PR #3536, opened 2022-11-01
Add `clear_actors` methods to plotter

banesullivan:
This addresses (maybe solves) #3522

Adds a new `clear_actors()` method to only clear actors in a scene -- not lights/view props


akaszynski (approved):
LGTM.

Issue #3313, opened 2022-09-13
Improving consistency of `__init__` parameters / moving towards VTK


banesullivan:
Several of the `__init__` methods in PyVista DataSet classes, while thoughtfully designed, are a little removed from what a typical VTK user might expect and are also inconsistent with the very attributes/properties being set. Take `UniformGrid` for example (actually, UniformGrid, while designed and implemented mostly by me is a hot mess that needs to be dealt with):

The `__init__` for `pyvista.UniformGrid` is the following:

https://github.com/pyvista/pyvista/blob/6aedb562ee8aeebba90fe163b757bd6a355b40a0/pyvista/core/grid.py#L482-L490

The `dims` argument is setting the `dimensions` property. This naming mismatch is not well documented and confusing. The other parameters like `origin` and `spacing` line up one-to-one, and so should the initialization parameter for the `dimensions` property.

Other `__init__` methods have similar issues... notably `PointSet`, which @MatthewFlamm is fixing in #3311 but also `StructuredGrid` where the `__init__` takes X Y and Z coordinates in a meshgrid format which is quite different from how you might declare a `StructuredGrid` without the `__init__`.

I would like to improve these `__init__` methods such that they are a bit more intuitive and line up with what exactly is being set on the data object.

Thoughts and suggestions would be appreciated here!

PR #3537, opened 2022-11-01
Rename dims kwarg for UniformGrid to dimensions

banesullivan:
`UniformGrid` has `dims` as a keyword argument for setting the `dimensions` property - this mismatch is confusing, internally inconsistent, and inconsistent with VTK.

These changes make the API a bit more consistent and intuitive.

Partially addresses #3313

### DIFF
--- a/doc/user-guide/data_model.rst
+++ b/doc/user-guide/data_model.rst
@@ -395,7 +395,7 @@ a simple mesh containing four isometric cells by starting with a
 
 .. jupyter-execute::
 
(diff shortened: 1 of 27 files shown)


akaszynski (dismissed):
LGTM.

Agree that if our public attribute is `dimensions`, then our kwarg should match.

akaszynski (approved):
LGTM. codecov is failing to upload and we need to fix that in a follow-up PR. Just a server load issue, not ours.

PR #3543, opened 2022-11-02
Update install dependencies in installation docs

adeak:
I noticed that `appdirs` was still mentioned in the installation documentation. Taking a closer look I saw that the list was incomplete in other ways. I've tried updating this section to be more up to date; I'm not sure about the blurbs I added to some of the items (pillow and imageio, mainly).

`typing-extensions` is specified as "`python_version < 3.8`", but the package itself is 3.7 or higher, so I only kept 3.7 in the docs.

I also noticed that some of these required dependencies (`imageio` and `PIL`) are present in nested imports. I'm not sure this makes sense now that these are mandatory; it might be cleaner to pull out these imports into the global namespace where they belong? Should we do this? (Even if yes, probably in a separate PR.)

### DIFF
--- a/doc/getting-started/installation.rst
+++ b/doc/getting-started/installation.rst
@@ -21,9 +21,11 @@ the following projects are required dependencies of PyVista:
 
 * `vtk <https://pypi.org/project/vtk/>`_ - PyVista directly inherits types from the VTK library.
 * `NumPy <https://pypi.org/project/numpy/>`_ - NumPy arrays provide a core foundation for PyVista's data array access.
-* `imageio <https://pypi.org/project/imageio/>`_ - This library is used for saving screenshots.
-* `appdirs <https://pypi.org/project/appdirs/>`_ - Data management for our example datasets so users can download tutorials on the fly.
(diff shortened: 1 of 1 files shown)


tkoyama010 (approved):
LGTM

adeak:
> > I also noticed that some of these required dependencies (imageio and PIL) are present in nested imports. I'm not sure this makes sense now that these are mandatory; it might be cleaner to pull out these imports into the global namespace where they belong? Should we do this? (Even if yes, probably in a separate PR.)
> 
> Personally, I think it is better to minimize information about other libraries. For example, if the current nested situation is resolved in the future, it would be easier to maintain if that information is not added.

@tkoyama010 thanks, but I'm not sure I understand. It sounds like you're suggesting that we don't mention nested imports in the docs. Which I agree with. What I meant about the nested imports is merely something I came across when looking at how we use these dependencies. If we do anything in that regard, it would only be an internal refactor.

Or did you mean something else? :)

tkoyama010:
@adeak  Oops. I misunderstood your comment. I thought you were talking about documentation when you talked about nesting. If you are talking about internal refactoring, I think it is fine 👍

adeak:
OK, thanks! Sorry for being confusing.

PR #3552, opened 2022-11-04
Add a badge for the supported Python version.

MaxJPRey:
### Overview

Add a badge to display the Python version supported by PyVista.
Even if this info is already available in the `readme.rst` file, badges are user friendly and expose all the info about the package at the same location.

The target of the badge points to the `python.org` in case a user needs to download the proper Python version.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,11 +51,14 @@ PyVista
   :target: https://github.com/psf/black
   :alt: black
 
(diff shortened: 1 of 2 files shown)


tkoyama010 (approved):
LGTM. It is a nice idea :)

PR #3553, opened 2022-11-04
Add blog post of IvanNik17

tkoyama010:
### Overview

Add blog post of @IvanNik17 as external examples.

 

### Details

- https://github.com/pyvista/pyvista/discussions/3126


banesullivan (dismissed):

doc/make_external_gallery.py
```diff
@@ -150,6 +150,24 @@ def format(self):
...
+        description="How to Voxelize Meshes and Point Clouds in Python",
+        link="https://medium.com/towards-data-science/how-to-voxelize-meshes-and-point-clouds-in-python-ca94d403f81d",
+        image="nikolov2.gif",
+    ),
+    nikolov3=Example(
+        title="Neighborhood Analysis, KD-Trees, and Octrees for Meshes and Point Clouds in Python",
+        description="Neighborhood Analysis, KD-Trees, and Octrees for Meshes and Point Clouds in Python",
+        link="https://medium.com/towards-data-science/neighborhood-analysis-kd-trees-and-octrees-for-meshes-and-point-clouds-in-python-19fa96527b77",
+        image="nikolov3.gif",
+    ),
```
adeak, line 170: Two things come to mind:
1. It would be nice to have a combination of short title + longer description as in the existing examples (instead of copying the same for both fields), but I think the new website doesn't actually use the description field anywhere (@akaszynski  please correct me if I'm wrong)
2. Since the "title" is what gets rendered in the gallery, I'd consider trying to make these shorter. At least look at the built docs (I haven't, yet) an check that the text doesn't look weird or broken in the gallery.

doc/make_external_gallery.py
```diff
@@ -150,6 +150,24 @@ def format(self):
...
+        title="How to Voxelize Meshes and Point Clouds in Python",
+        description="How to Voxelize Meshes and Point Clouds in Python",
+        link="https://medium.com/towards-data-science/how-to-voxelize-meshes-and-point-clouds-in-python-ca94d403f81d",
+        image="nikolov2.gif",
+    ),
+    nikolov3=Example(
+        title="Neighborhood Analysis, KD-Trees, and Octrees for Meshes and Point Clouds in Python",
+        description="Neighborhood Analysis, KD-Trees, and Octrees for Meshes and Point Clouds in Python",
+        link="https://medium.com/towards-data-science/neighborhood-analysis-kd-trees-and-octrees-for-meshes-and-point-clouds-in-python-19fa96527b77",
+        image="nikolov3.gif",
```
adeak, line 169: One more, sorry: it seems to me that this nice yellow bunny gif was made with Vedo, not PyVista :( We should double-check that all 3 gifs were made with PyVista (the first two are probably OK, but I only skimmed the three posts a bit).

tkoyama010: Thank you for the check. Sorry for the mistake. This is due to the fact that the word "PyVista" was on the right and "Vedo" was on the left. I should have looked at the captions more carefully.
image

tkoyama010: Fixed in a1b7aee0a142558d73069415e02d3467a3abcc5f .

doc/make_external_gallery.py
tkoyama010: How about following the format of a scientific paper and titling it "Ivan Nikolov, Apr 26, 2022", "Ivan Nikolov, May 31, 2022" and "Ivan Nikolov, Jun 16, 2022"? (I am not a scientist, so my writing style may be wrong.)

adeak:
One more general question: do we know anything about copyright(/licensing??) of these gifs? Is it OK to just add them to our repo? Should we ask for permission?

I've found https://github.com/IvanNik17/python-3d-analysis-libraries which is very related, but I can't easily find these gifs in there, only source and data. But this is GPL-3.0 licensed. I don't even know if that would apply to images, or only source code... it's probably fine.

doc/make_external_gallery.py
adeak: If we go that route I'd prefer to say something about the contents. So something like
1. Nikolov on Visualization Libraries
2. Nikolov on Voxelization and Point Clouds
3. Nikolov on Neighbourhood Analysis

If "Ivan Nikolov" fits too, that would be even better.

tkoyama010:
@adeak This PR is not a rush, so we can wait for @IvanNik17's response before merging.

doc/make_external_gallery.py
tkoyama010: Fixed in 659f3da288310efe95bd05662b5b5b22187d7311 .

doc/make_external_gallery.py
```diff
@@ -150,6 +150,24 @@ def format(self):
...
         image="grad_descent_visualizer.gif",
     ),
+    nikolov1=Example(
+        title="Ivan Nikolov on Visualization Libraries",
+        description="Python Libraries for Mesh, Point Cloud, and Data Visualization (Part 1)",
+        link="https://medium.com/towards-data-science/python-libraries-for-mesh-and-point-cloud-visualization-part-1-daa2af36de30",
+        image="nikolov1.gif",
+    ),
+    nikolov2=Example(
+        title="Ivan Nikolov on Voxelization and Point Clouds",
```
adeak, line 160: Sorry, I promise this is the last comment for a while :D I misread the original title: it's "voxelize (meshes and point clouds)". So we can just use
```suggestion
        title="Ivan Nikolov on Voxelization",
```

tkoyama010: You can make as many comments as you like.

IvanNik17:
@tkoyama010 Thank you for considering my blog post. I just wanted to add that it's totally fine to add it. All images/gifs are made by me for the article, so copyright should be fine.

tkoyama010:
@IvanNik17 Thanks! We are glad that we can add your great job to our page.

akaszynski (approved):

PR #3556, opened 2022-11-05
Fix nan_opacity issues and above/below annotations

### LINKED ISSUE #3445: Opacity for NaN-values does not work when adding mesh
Dear PyVista-Team,
Dear PyVista-Community,

When creating a mesh from a UniformGrid with some NaN-values, it is not possible to hide those NaNs (opacity=0).
When setting the parameter nan_opacity to 0, the NaNs appear in grey (see screenshot [1]). When setting nan_opacity to 1 (or 0) and nan_color='black' (can be any color) the NaNs are finally hidden in the plot but the opacity of the remaining values also changed (somewhere below 1 - see screenshot [2]). When using this paradoxical workaround for my actual project (same data range), the NaNs in the volume got nearly invisible but the non-NaNs (which I actually want to show) totally diappeared (see screenshot [3]).

NaNs are grey:
```
import pyvista as pv
import numpy as np

noise = pv.perlin_noise(0.1, (1, 1, 1), (0, 0, 0))
grid = pv.sample_function(noise, [0, 1.0, -0, 1.0, 0, 1.0], dim=(20, 20, 20))

minV = min(grid['scalars'])
maxV = max(grid['scalars'])
grid['scalars'] = (grid['scalars'] - minV) / (maxV - minV)
grid['scalars'] *= 100

grid['scalars'] = np.where(grid['scalars'] < 20, np.nan, grid['scalars'])
p = pv.Plotter()

print(grid.get_data_range())
p.add_mesh(grid, nan_opacity=0)
p.show(window_size=[500, 500])
```

NaNs are invisible but the remaining values have a low opacity:
```
import pyvista as pv
import numpy as np

noise = pv.perlin_noise(0.1, (1, 1, 1), (0, 0, 0))
grid = pv.sample_function(noise, [0, 1.0, -0, 1.0, 0, 1.0], dim=(20, 20, 20))

minV = min(grid['scalars'])
maxV = max(grid['scalars'])
grid['scalars'] = (grid['scalars'] - minV) / (maxV - minV)
grid['scalars'] *= 100

grid['scalars'] = np.where(grid['scalars'] < 20, np.nan, grid['scalars'])
p = pv.Plotter()

print(grid.get_data_range())
p.add_mesh(grid, nan_color='black', nan_opacity=1)
p.show(window_size=[500, 500])
```

### System Information

```shell
  Date: Tue Oct 11 15:04:01 2022 CEST

                OS : Linux
            CPU(s) : 16
           Machine : x86_64
      Architecture : 64bit
               RAM : 62.5 GiB
       Environment : Python
       File system : ext4
        GPU Vendor : Intel
      GPU Renderer : Mesa Intel(R) UHD Graphics (TGL GT1)
       GPU Version : 4.6 (Core Profile) Mesa 21.2.6

  Python 3.7.6 (default, Jan  8 2020, 19:59:22)  [GCC 7.3.0]

           pyvista : 0.36.1
               vtk : 9.1.0
             numpy : 1.18.4
           imageio : 2.6.1
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.2.1
             PyQt5 : 5.9.2
           IPython : 7.12.0
             scipy : 1.4.1
              tqdm : 4.46.0
        jupyterlab : 1.2.6

  Intel(R) Math Kernel Library Version 2020.0.0 Product Build 20191122 for
```

### Screenshots

Screenshot [1]:
nan_opacity=0

Screenshot [2]:
pyvista bug

Screenshot [3]:
nearly invisible

banesullivan:
Resolves #3445 (see comments there)

@dcbr, a review would be appreciated.


dcbr (approved):
LGTM if my comment in #3445 is correct, i.e.

> So I believe for this use case (add_mesh) the usage of opacity is correct. You want to override any alpha values of the provided (or theme) nan_color.

banesullivan:
Thanks for the feedback, @dcbr!

I also just added some tests to verify `nan_opacity` works.

@akaszynski, I modified `LoolupTable` to have a `nan_opacity` property, but this feels a bit hacky since `nan_opacity` is really a part of `nan_color`. I feel like my change might have some consequence... would you please review?

tests/plotting/test_plotting.py
```diff
@@ -3031,13 +3031,28 @@ def test_lookup_table(verify_image_cache):
...
     lut.plot()
 
 
+def test_plot_nan_color(uniform):
+    arg = uniform.active_scalars < uniform.active_scalars.mean()
+    uniform.active_scalars[arg] = np.nan
+    # NaN values should be hidden
+    pl = pyvista.Plotter()
+    pl.add_mesh(uniform, nan_opacity=0)
+    pl.enable_depth_peeling()
```
banesullivan, line 3048: Depth peeling makes the opacity of the rest of the mesh correct... is this okay to include for CI tests?

pyvista/plotting/lookup_table.py
```diff
@@ -513,6 +514,31 @@ def nan_color(self, value):
...
+        >>> lut.plot()
+
+        """
+        color = self.nan_color
+        return color.opacity
+
+    @nan_opacity.setter
+    def nan_opacity(self, value):
+        nan_color = self.nan_color.float_rgba[:3]
+        self.nan_color = Color(nan_color, opacity=value).float_rgba
```
dcbr, line 540:
```suggestion
        self.nan_color = Color(self.nan_color, opacity=value)
```

Colors can be created from other `Color` instances, so no need to first extract the RGB components yourself. I also removed the conversion to `float_rgba` as I believe `self.nan_color` should be a `Color` instance?

akaszynski (dismissed):
This is good to go. Feel free to add to it and I'll reapprove.

banesullivan:
Made a few more changes to ensure that if you build a LookupTable with above/below/nan color annotations that the scalar bar would be set up correctly

banesullivan:
NaN must be explicitly requested for the scalar bar though as the NaN color is always set on the LookupTable and we can't really track when it's user set vs. default set without some messy code.

To get a nan-annotaion, users must:

```py
pl.add_mesh(
        ..., nan_opacity=0.5, nan_color='green', scalar_bar_args=dict(nan_annotation=True)
    )
```

pyvista/plotting/colors.py
```diff
@@ -872,6 +872,33 @@ def to_dict(self):
...
+
+        >>> color
+        Color(name='red', hex='#ff000040', opacity=64)
+
+        """
+        return self._opacity
+
+    @opacity.setter
+    def opacity(self, value: Union[int, float, str]):
+        self._opacity = self.convert_color_channel(value)
```
dcbr, line 900: I decided against making Color instances mutable in #2175. The reason being that sometimes colors are only used as "intermediate objects" between the Pyvista user and the underlying VTK methods. See e.g.

https://github.com/pyvista/pyvista/blob/fecf347ca56221c3617a0121ec2cd603b9378439/pyvista/plotting/charts.py#L196-L216

If a user uses the color getter of a `Pen` instance and modifies the opacity, the `Color` instance will be changed, but the `SetColor` VTK method won't be called to actually update the color (as would be the case when using the color setter).
This would also break the hashing of `Color` instances (hash value is calculated from each color channel).

dcbr (changes requested):

pyvista/plotting/colors.py
banesullivan: ah makes sense, thanks for pointing this out. I'll modify this PR to make sure it is not mutating any `Color` objects

pyvista/plotting/scalar_bars.py
```diff
@@ -164,7 +164,7 @@ def add_scalar_bar(
         fmt=None,
         use_opacity=True,
         outline=False,
-        nan_annotation=False,
+        nan_annotation=None,
```
banesullivan, line 167: Maybe this should be kept `False`

akaszynski (approved):
LGTM. Thanks @banesullivan!

Please merge and ignore missing codecov, it's a server-side failure to report, not our fault.

akaszynski:
Merging this immediately as this is a high priority and it's been sitting here for a while.

PR #3579, opened 2022-11-10
Adapt project to pytest-pyvista plugin

### LINKED ISSUE #3324: Consider implement a pyvista.testing module

AlejandroFernandezLuces:
### Overview

In this PR I remove the `verify_image_cache` fixture and `VerifyImageCache` class to [pytest-pyvista plugin](https://github.com/pyvista/pytest-pyvista/pull/3). I also adapt the test to work with the new plugin in necessary cases.

Resolves #3324

### Details

- Removed special case test list, now you have to set a `VerifyImageCache` attribute to set that test with the proper flag

### References

https://github.com/pyvista/pyvista/pull/3494


tests/test_plotter.py
```diff
@@ -170,7 +170,8 @@ def test_plotter_remains_shallow():
     assert np.shares_memory(sphere['numbers'], sphere2['numbers'])
 
 
-def test_add_multiple(sphere):
+def test_add_multiple(sphere, verify_image_cache):
```
MatthewFlamm, line 173: Since `verify_cache_image` is a regular fixture,i.e. not autoused, why is it needed to use it in the function signature and then skip it?

MatthewFlamm (commented):
Where is the cache directory set?  If no where, I don't think that we are testing against our cache.   It would be good if upstream catches this. I think this is related to https://github.com/pyvista/pytest-pyvista/issues/6

We should wrap `verify_image_cache` inside `testing/plotting/test_plotting.py` and make it `autouse=True`.  This would remove a lot of changes here.  This also allows us to set the cache dir

tests/test_plotter.py
AlejandroFernandezLuces: Totally right about the wrapper. 

I missed this file when I removed the autouse flag in the plugin, should be removed.

.github/workflows/style_docstr.yml
```diff
@@ -57,7 +57,10 @@ jobs:
           pip list
 
       - name: Test Package Docstrings
-        run: pytest -v --doctest-modules pyvista
+        run: |
+          pip uninstall --yes pytest-pyvista 
+          pytest -v --doctest-modules pyvista
+          pip install pytest-pyvista
```
akaszynski, line 63: This shouldn't be necessary.

.github/workflows/testing-and-deployment.yml
```diff
@@ -41,7 +41,6 @@ jobs:
         run: |
           pip install .
           python -c "import pyvista;print(pyvista.Report(gpu=False));from pyvista import examples;print('User data path:', examples.USER_DATA_PATH)"
-
```
akaszynski, line 44: These line breaks are outside the scope of this PR. Let's change this in a follow-up and discuss if we want to standardize this. If we do, let's create a pre-commit for it.

.github/workflows/testing-and-deployment.yml
```diff
@@ -160,8 +152,13 @@ jobs:
           restore-keys: |
             Examples-1-
 
+      - name: Install Testing Requirements
+        run: pip install -r requirements_test.txt
```
akaszynski, line 156: Move this back to L178 and you won't have to uninstall and reinstall pyvista.

tests/plotting/test_plotting.py
```diff
@@ -2786,7 +2658,9 @@ def test_plot_composite_poly_component_single(multiblock_poly):
     pl.show()
 
 
-def test_plot_composite_poly_component_nested_multiblock(multiblock_poly):
+def test_plot_composite_poly_component_nested_multiblock(multiblock_poly, verify_image_cache):
```
akaszynski, line 2661: Is there a way we can autouse verify_image_cache for an entire file? Seems like it would be possible and would eliminate a lot of work maintaining function signatures.

AlejandroFernandezLuces: We are doing that in lines 92-94, in this test we need to use an attribute of the plugin. We can move the skip attributes back to the lists, as in lines 81-130, but I though it would be easier to read and maintain this way.

MatthewFlamm: I like the implentation in this PR.  It means that we annotate at the point of use.  When reading a test it is immediately obvious if it is a high variance test for example.  One downside is that you have to do 'ctrl-f' or similar to find all usages like this.  I wonder if there is a way to utilize decorators for this purpose, but this is for another PR.

```py
@skip_windows_cache
def test_plot_1():
    ...
```

tests/conftest.py
```diff
@@ -6,6 +6,7 @@
 from pyvista import examples
 
 pyvista.OFF_SCREEN = True
+# pytest_plugins = ["pyvista"]
```
adeak, line 9: Does this comment here have purpose? Or should it be removed?

tests/plotting/test_plotting.py
```diff
@@ -26,6 +25,17 @@
...
+# TODO: once we have a stable release for VTK, remove these.
+
+# these images vary between Windows when using OSMesa and Linux/MacOS
+# and will not be verified
+
+
+# these images vary between Linux/Windows and MacOS
+# and will not be verified for MacOS
+
+
```
adeak, line 38: These comments seem to be leftovers from the old design, and are somewhat confusing here.

tests/plotting/test_plotting.py
```diff
@@ -78,56 +88,10 @@ def using_mesa():
 )
 skip_mesa = pytest.mark.skipif(using_mesa(), reason='Does not display correctly within OSMesa')
 
-# Image regression warning/error thresholds for releases after 9.0.1
-# TODO: once we have a stable release for VTK, remove these.
-HIGH_VARIANCE_TESTS = {
-    'test_add_title',
-    'test_export_gltf',  # image cache created with 9.0.20210612.dev0
-    'test_import_gltf',  # image cache created with 9.0.20210612.dev0
-    'test_opacity_by_array_direct',  # VTK regression 9.0.1 --> 9.1.0
```
adeak, line 87: Did I see correctly that you didn't carry over these comments to the appropriate skips? I think these can be useful for posterity, e.g. when 9.0.1 is no longer supported then we could hypothetically revisit the flakiness of this test. So I'd copy over these comments next to the appropriate per-test skips in the corresponding tests.

tests/plotting/test_plotting.py
```diff
@@ -2618,12 +2491,7 @@ def test_ssao_pass():
         return
 
     pl.enable_ssao()
-    pl.show(auto_close=False)
-
-    # ensure this fails when ssao disabled
-    pl.disable_ssao()
-    with pytest.raises(RuntimeError):
-        pl.show()
+    pl.show()
```
adeak, line 2494: Why was this part of the test removed?

tests/plotting/test_plotting.py
AlejandroFernandezLuces: I didn't, I'll copy them to the appropiate tests.

tests/plotting/test_plotting.py
AlejandroFernandezLuces: This test wasn't working as intended, it was failing (not throwing a runtime error) on a clean install on main branch. I believe it works on github actions thanks to an old cache, which causes to throw a runtime error due to images dissimilarity.

Since what this test should raise was confusing for me, I asked @akaszynski and he recommended to remove those lines.

tests/plotting/test_plotting.py
adeak: I don't want to sound too nitpicky, but it would be most logical to me to put these comments right before the
```py
verify_image_cache.high_variance_test = True
```
At least if I understand correctly that these comments were explaining the "high variance" part.

AlejandroFernandezLuces: Not nitpicky at all! I really appreciate your comments. I agree, I was not sure when I moved them but it makes sense to be there.

.github/workflows/style_docstr.yml
```diff
@@ -57,7 +59,8 @@ jobs:
           pip list
 
       - name: Test Package Docstrings
-        run: pytest -v --doctest-modules pyvista
+        run: |
+          pytest -v --doctest-modules pyvista
```
akaszynski, line 63:
```suggestion
        run: pytest -v --doctest-modules pyvista
```

tests/plotting/test_plotting.py
akaszynski: Let's leave this for another PR, but I appreciate (and agree) with this.

MaxJPRey (approved):
This looks solid @AlejandroFernandezLuces .

PR #3601, opened 2022-11-15
Hotfix: resolve broken deprecated methods

banesullivan:
These were not deprecated correctly - hotfix until we decide to actually remove the method


tkoyama010 (approved):
LGTM. Thanks for the fix.

banesullivan:
A patch release would be nice. I'm happy to give it a few days or a week+ to let any other patches come in too. We can cherry pick the commits, so no need to release immediately

PR #3602, opened 2022-11-15
Finalize v0.37 deprecations

banesullivan:
This finalizes many deprecations that were targeted for v0.37 (which has already released). IMO, all of these are safe to remove at this point.

For any remaining deprecations, I made sure to add a note of planned removal; most targeting v0.40

### DIFF
--- a/pyvista/core/datasetattributes.py
+++ b/pyvista/core/datasetattributes.py
@@ -1,14 +1,13 @@
 """Implements DataSetAttributes, which represents and manipulates datasets."""
 
 from typing import Any, Dict, Iterator, List, Optional, Sequence, Tuple, Union
-import warnings
 
 import numpy as np
 
 from pyvista import _vtk
 import pyvista.utilities.helpers as helpers
 from pyvista.utilities.helpers import FieldAssociation
-from pyvista.utilities.misc import PyVistaDeprecationWarning, copy_vtk_array
+from pyvista.utilities.misc import copy_vtk_array
 
 from .._typing import Number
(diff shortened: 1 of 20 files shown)


akaszynski (approved):
LGTM. Thanks!

MatthewFlamm (approved):
This should probably be merged after #3601, since that one targets main branch and this one removes the fixed code entirely.

pyvista/core/datasetattributes.py
```diff
@@ -450,21 +401,6 @@ def active_t_coords(self, t_coords: np.ndarray):
...
-        .. deprecated:: 0.32.0
-            Use :attr:`DataSetAttributes.active_t_coords_name` to
-            return the name of the active texture coordinates array.
-
-        """
-        warnings.warn(
-            "Use of `active_texture_name` is deprecated. Use `active_t_coords_name` instead.",
-            PyVistaDeprecationWarning,
-        )
-        return self.active_t_coords_name
```
banesullivan, line 466: I want to point out that is was not supposed to be deprecated... `active_texture_name` is not equivalent to `active_t_coords_name`. `active_texture_name` refers to a PyVista-specific feature I implemented a few years ago that tracks actual `Texture` objects so that if you have an "active texture" that `add_mesh` would pick up on it and render it automatically. This has not equivalent in VTK so I'm fine with it now being broken/not supported but we may want to clean all of this up in a new PR as the `_textures` property this is supposed to control still exists.

cc @pyvista/developers

PR #3606, opened 2022-11-15
Expose BandedPolyDataContourFilter

akaszynski:
Expose `vtk.vtkBandedPolyDataContourFilter`. It's similar to `vtkContourFilter`, but different enough to necessate creating a new method as it's only applicable to `PolyData`.

Example:

```py
Extract the surface from the uniform grid dataset and plot its contours
alongside the output from the banded contour filter.

import pyvista as pv
from pyvista import examples
surf = examples.load_uniform().extract_surface()
output, edges = surf.contour_banded(5, rng=[200, 500])
pl = pv.Plotter()
_ = pl.add_mesh(edges, line_width=5, render_lines_as_tubes=True, color='k')
_ = pl.add_mesh(surf, opacity=0.3, show_scalar_bar=False)
_ = pl.add_mesh(output)
pl.show()
```
tmp


banesullivan (approved):
Indeed a bit different from the standard contour filter so happy to have this as a separate method thought I would've loved to have an `return_edges` argument do this..

PR #3619, opened 2022-11-18
Move project metadata to pyproject.toml; avoid calling setup.py

mwtoews:
### Overview

The PR utilizes pyproject.toml for project metadata and build dependencies, and uses PyPA's [build](https://github.com/pypa/build) tool instead of invoking setup.py directly.

### Details

- Specifies project build dependencies ([PEP 518](https://peps.python.org/pep-0518/)), i.e. it's just setuptools. This means you don't need to `pip install` any thing before installing this package 
- Moves project metadata from setup.py to pyproject.toml, see [setuptools docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) and [PEP 621](https://peps.python.org/pep-0621/)
  - This is close to a 1:1 conversion, with a few small modifications such as alphasorting lists
  - Explicit listing of packages is replaced with setuptools' `find()` to gather everything under "pyvista*"
  - Out in the wild, most folks and docs seem to prefer double quotes, but this PR maintains the current style using mostly single quotes (except keys)
- While setup.py is nearly empty, keep it for now, since some packages depend on it
- Calling setup.py is [discouraged](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html); use PyPA's [build](https://github.com/pypa/build) tool instead
- Use `twine check --strict dist/*` to fail CI check if needed


akaszynski (approved):
Excellent work. Thanks for adding it, and I appreciate barely around `setup.py` so we can support legacy installs until it completely goes away.

I'm also happy we're no longer reading in the `__version__` from file, thanks for refactoring that out.

banesullivan (approved):
Excellent! Thank you for taking the initiative on this as a few of us have wanted/talked about doing this for a while

Issue #3625, opened 2022-11-20
argument "name" in Plotter.update_scalar_bar_range(clim, name=None) is not working


Jimmy-KL:
### Describe the bug, what's wrong, and what you expected.

If I give a value to the argument _name_ in Plotter.update_scalar_bar_range(clim, name=None), there will be an error message as follow ([API Reference](https://docs.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.update_scalar_bar_range.html)):
AttributeError: 'Plotter' object has no attribute '_scalar_bar_mappers'

What I tried to do is that when I change the data array, I want the color map to change its value range and the tile of the color bar automatically. Then I find this method Plotter.update_scalar_bar_range(). Unfortunately, the 2nd argument cannot work.
BTW, if there any other way to do what I expect? Thanks!

### Steps to reproduce the bug.

```python
import pyvista as pv
import numpy as np

mesh = pv.Sphere()
mesh.point_data['data1'] = np.arange(mesh.n_points)
mesh.point_data['data2'] = 10*np.arange(mesh.n_points)

pl = pv.Plotter()
pl.add_mesh(mesh, scalars='data2')
mesh.set_active_scalars('data1')
min_value = min(mesh.get_array('data1'))
max_value = max(mesh.get_array('data1'))
pl.update_scalar_bar_range([min_value, max_value], name='data1')
pl.show()
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Sun Nov 20 22:48:25 2022 中国标准时间

                OS : Windows
            CPU(s) : 16
           Machine : AMD64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : Intel
      GPU Renderer : Intel(R) UHD Graphics
       GPU Version : 4.5.0 - Build 31.0.101.1999

  Python 3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit
  (AMD64)]

           pyvista : 0.36.1
               vtk : 9.1.0
             numpy : 1.22.0
           imageio : 2.22.0
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.1
           IPython : 8.0.1
        ipyvtklink : 0.2.3
             scipy : 1.7.3
            meshio : 5.3.4
         pythreejs : Version unknown
--------------------------------------------------------------------------------
```

adeak:
[...]

PR #3626, opened 2022-11-20
Fix update_scalar_bar_range() with named scalar bar

adeak:
Fixes https://github.com/pyvista/pyvista/issues/3625.

The method was also missing tests which explains why the bug was sticking around. I've added these.

TODO:
- [x] figure out scalar bar tests vs plotting (image regression) tests (awaiting feedback)
- [x] add/remove any new/old image regression images


pyvista/plotting/plotting.py
```diff
@@ -3758,17 +3758,12 @@ def update_scalar_bar_range(self, clim, name=None):
             self.mapper.scalar_range = clim
             return
 
-        # Use the name to find the desired actor
-        def update_mapper(mapper_helper):
-            mapper_helper.scalar_range = clim
-            return
```
adeak, line 3764: This helper function is one line and used in one place, so I've inlined its contents in the loop.

pyvista/plotting/plotting.py
```diff
@@ -3758,17 +3758,12 @@ def update_scalar_bar_range(self, clim, name=None):
...
         try:
-            for mh in self._scalar_bar_mappers[name]:
-                update_mapper(mh)
+            # use the name to find the desired actor
+            for mh in self.scalar_bars._scalar_bar_mappers[name]:
+                mh.scalar_range = clim
         except KeyError:
-            raise KeyError('Name ({}) not valid/not found in this plotter.')
-        return
+            raise ValueError(f'Name ({name!r}) not valid/not found in this plotter.') from None
```
adeak, line 3766: I've changed the error from `KeyError` to `ValueError` for semantics (and the fact that the message of a `KeyError` should traditionally contain the offending key). I've also added `raise ... from None` to suppress the (implementation detail) context of the original exception.

tests/plotting/test_plotting.py
```diff
@@ -2164,17 +2164,6 @@ def test_set_viewup():
...
-    sphere['z'] = sphere.points[:, 2]
-    plotter = pyvista.Plotter()
-    plotter.add_mesh(sphere, show_scalar_bar=False)
-    plotter.add_scalar_bar(interactive=True)
-    assert len(plotter.scalar_bars) == 1
-    plotter.remove_scalar_bar()
-    assert len(plotter.scalar_bars) == 0
-    plotter.show()
-
-
```
adeak, line 2177: This test largely overlapped with a test from `tests/test_scalar_bars.py` that I updated a bit.

But now that I'm writing this I realised that this test had added value via the regression test. Is it worth keeping? Or adding more tests here? If we scrap it then I also have to remove the corresponding image regression test.

On a related note, I was going to move `test/test_scalar_bars.py` to `test/plotting/`, but then I realised that this would need adding image regression testing (`pl.show()` calls for instance) and for those test functions it might not add much value. Perhaps we should do the move after all, and add more regression tests... I'm not really sure what to do with these.

pyvista/plotting/plotting.py
banesullivan: TIL: `raise ... from X`

banesullivan: +1 for the error type change

tests/plotting/test_plotting.py
banesullivan: Hopefully, all of this is made easier with the image regression fixture? I'm +1 for moving `test/test_scalar_bars.py` to `tests/plotting/`

akaszynski: I moved the relevant tests to plotting. Agreed that "seeing" is believing.

akaszynski (approved):
Addressed all of @adeak's remaining issues.

adeak:
I've completely forgotten about this, thanks a lot for cleaning this up @akaszynski. Your changes look good to me (but I can't approve my own PR.)

PR #3628, opened 2022-11-20
Fix test suite warnings

adeak:
I'm seeing a bunch of warnings in local test runs:
```
$ pytest -v tests/test_dataset.py tests/test_helpers.py tests/test_pointset.py tests/test_picking.py tests/plotting/test_widgets.py 
[snip]
tests/test_dataset.py::test_activate_texture_none
  /home/adeak/pyvista/pyvista/core/dataset.py:556: UserWarning: Texture (not a key) not associated with this dataset
    warnings.warn(f'Texture ({name}) not associated with this dataset')

tests/test_dataset.py::test_activate_texture_none
  /home/adeak/pyvista/pyvista/core/dataset.py:550: UserWarning: No textures associated with input mesh.
    warnings.warn('No textures associated with input mesh.')

tests/test_helpers.py: 1 warning
tests/test_pointset.py: 12 warnings
  /home/adeak/pyvista/pyvista/utilities/helpers.py:507: UserWarning: Points is not a float type. This can cause issues when transforming or applying filters. Casting to ``np.float32``. Disable this by passing ``force_float=False``.
    warnings.warn(

tests/test_picking.py::test_point_picking[False]
tests/test_picking.py::test_point_picking[True]
tests/test_picking.py::test_path_picking
tests/test_picking.py::test_geodesic_picking
tests/test_picking.py::test_horizon_picking
  /home/adeak/pyvista/pyvista/utilities/helpers.py:1201: UserWarning: Encountered issue in callback (most recent call last):
  TypeError: <lambda>() takes 0 positional arguments but 1 was given
    warnings.warn(formatted_exception)

tests/test_picking.py::test_point_picking[False]
tests/test_picking.py::test_point_picking[True]
  /home/adeak/pyvista/pyvista/utilities/helpers.py:1201: UserWarning: Encountered issue in callback (most recent call last):
  TypeError: <lambda>() takes 0 positional arguments but 2 were given
    warnings.warn(formatted_exception)

tests/plotting/test_widgets.py::test_widget_slider
  /home/adeak/pyvista/pyvista/utilities/helpers.py:1201: UserWarning: Encountered issue in callback (most recent call last):
  TypeError: <lambda>() missing 1 required positional argument: 'widget'
    warnings.warn(formatted_exception)

tests/plotting/test_widgets.py::test_widget_uniform
  /home/adeak/pyvista/pyvista/utilities/helpers.py:1201: UserWarning: Encountered issue in callback (most recent call last):
  TypeError: <lambda>() takes 1 positional argument but 2 were given
    warnings.warn(formatted_exception)
[snip]
```

I've made some occasionally opinionated decisions in handling these, but I'm open to alternative approaches :)

TODO:
- [x] revert coverage drop due to no failing callback
- [x] fix widgets tests after https://github.com/pyvista/pyvista/pull/3638 is merged, especially important if there's no conflict
- [ ] check for `force_float` warning in `vtk_points` test
- [ ] mock widget callbacks if applicable


pyvista/core/dataset.py
```diff
@@ -553,7 +553,7 @@ def _activate_texture(mesh, name: str) -> _vtk.vtkTexture:
         try:
             texture = mesh.textures[name]
         except KeyError:
-            warnings.warn(f'Texture ({name}) not associated with this dataset')
+            warnings.warn(f'Texture ({name!r}) not associated with this dataset')
```
adeak, line 556: This is just a matter of style, but e.g. in the "invalid" test seeing "`Texture ('not a key') not associated ...`" is better than "``Texture (not a key) not associated ..." in my opinion

tests/plotting/test_widgets.py
```diff
@@ -9,17 +9,20 @@
     pytestmark = pytest.mark.skip
 
 
+def noop_callback(*args, **kwargs):
+    """A callback that doesn't do anything."""
+    pass
```
adeak, line 14: This is the most opinionated choice. Most of the callbacks in this module were fine, but 2 of them weren't. I decided that it's cleaner to use a catch-all callback instead of defining a one-arg, a two-arg etc. separate callbacks. Part of the reason for this was that the existing "OK" callbacks return non-`None` things, which aren't so great callbacks after all.

If we want to keep testing the signature of callbacks I can separate these. But the fact that these warnings have been around means getting the signature wrong is not a very strong signal anyway.

tests/test_pointset.py
```diff
@@ -121,28 +122,29 @@ def test_center_of_mass():
...
+@pytest.mark.filterwarnings("ignore:Points is not a float type")
 def test_points_to_double():
     np_points = np.array([[1, 2, 3]], np.int64)
     pset = pyvista.PointSet(np_points)
     assert pset.points_to_double().points.dtype == np.double
 
 
 def test_translate():
-    np_points = np.array([1, 2, 3], np.int64)
+    np_points = np.array([1, 2, 3], np.float32)
```
adeak, line 133: Since we're testing both values of `force_float` separately, I figured we can just change all the other sources of the warning to use `float32` from the get go. The only thing missing then is a "check that default is `force_float=True`" test, but I'm not sure how useful that is.

adeak:
Coverage drop is due to no longer failing callbacks. I may have removed intentionally-bad calbacks. I'll check.

adeak:
> Coverage drop is due to no longer failing callbacks. I may have removed intentionally-bad calbacks. I'll check and fix.

It didn't seem like any of the callbacks were wrong on purpose, so I've kept the original changes and added a targeted test in `tests/test_helpers.py`. The PR is now complete, pending open comments for feedback.

I've also found a bunch of uncovered lines in `pyvista.utilities.helpers`, so I'll open a separate PR to fix most of those (once this is merged so that I don't have to fix a conflict).

tests/examples/test_download_files.py
```diff
@@ -755,12 +755,14 @@ def test_download_action_figure():
...
 def test_download_mars_jpg():
-    filename = examples.download_mars_jpg()
+    with pytest.warns(pv.PyVistaDeprecationWarning, match=r'Use .+\.download_mars_surface'):
+        filename = examples.download_mars_jpg()
     assert os.path.isfile(filename)
 
 
 def test_download_stars_jpg():
-    filename = examples.download_stars_jpg()
+    with pytest.warns(pv.PyVistaDeprecationWarning, match=r'Use .+\.download_stars_sky_background'):
```
adeak, line 764: These are only shortened from `examples.planets.download_stars_sky_background` in order to fit into the 100-char line length (it didn't seem worth it to break the line with the fully qualified name).

tests/plotting/test_widgets.py
adeak: On second thought, we have cases such as `p.add_spline_widget(callback=noop_callback, pass_widget=True, color=None, show_ribbon=True)` in `test_widgets.py`, and the expected behaviour is exactly that the signature of the callback has to change... so perhaps I should revert (correct) signature-specific callbacks.

MatthewFlamm: Are these callback functions called at least once when initialing the widget?  If so, we could use mock objects in the test and then assert that they are called with the expected signature for that test.

adeak: I'm not sure it's on initialization, but at least some of the callbacks get called some of the time, hence the warning when the signatures are off. I'll look into this, thanks.

pyvista/core/dataset.py
MatthewFlamm: This seems slightly better to me.

tests/test_pointset.py
```diff
@@ -109,6 +109,7 @@ def test_filters_return_pointset(sphere):
 
 
 @pytest.mark.parametrize("force_float,expected_data_type", [(False, np.int64), (True, np.float32)])
+@pytest.mark.filterwarnings("ignore:Points is not a float type")
```
MatthewFlamm, line 112: Do we test anywhere that this warning is spit out? If not, I would think it is better to use the `pytest.warns` context manager for this.  Then you could also add in two more lines to test the default behavior, which would address your comment further down.

adeak: Good point, thanks, it seems we don't check this warning anywhere. I think it's fair to add this in this PR, probably in the test that tests `force_float`, as you mentioned.

adeak:
I guess this is no longer necessary as of  https://github.com/pyvista/pyvista/pull/3740... I can't be bothered to check the diffs.

PR #3638, opened 2022-11-22
Python 3 style using pyupgrade

mwtoews:
### Overview

This upgrades a few Python 3 styles, automated using [pyupgrade](https://github.com/asottile/pyupgrade), using the command:
```shell
$ pyupgrade --py37-plus `find . -name "*.py*"`
```
with a few manual edits for quality control.

### Details

- Use "yield from" where possible
- Use f-strings where possible
- Updates to python modules faulthandler and collections.abc
- No need to inherit classes from object anymore
- IOError is an alias for OSError
- Correction to use a raw string `r'doctest: *\+SKIP')`; I believe "noqa: [W605](https://www.flake8rules.com/rules/W605.html)" was incorrect and should not have been ignored
- Remove unnecessary parentheses
- Use `set` comprehension and literals

If any of these styles are objectionable and the current is preferred, note these and I'll restore them.

### DIFF
--- a/pyvista/core/datasetattributes.py
+++ b/pyvista/core/datasetattributes.py
@@ -250,8 +250,7 @@ class DataSetAttributes(_vtk.VTKObjectWrapper):
 
     def __iter__(self) -> Iterator[str]:
         """Implement for loop iteration."""
-        for array in self.keys():
-            yield array
+        yield from self.keys()
 
     def __len__(self) -> int:
         """Return the number of arrays."""
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -2144,7 +2144,7 @@ class PolyDataFilters(DataSetFilters):
         )
         if retry:
             # gather intersecting rays in lists
-            loc_lst, ray_lst, tri_lst = [arr.tolist() for arr in [locations, index_ray, index_tri]]
+            loc_lst, ray_lst, tri_lst = (arr.tolist() for arr in [locations, index_ray, index_tri])
 
             # find indices that trimesh failed on
             all_ray_indices = np.arange(len(origins))
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -2144,7 +2144,9 @@ class PolyDataFilters(DataSetFilters):
         )
         if retry:
             # gather intersecting rays in lists
-            loc_lst, ray_lst, tri_lst = (arr.tolist() for arr in [locations, index_ray, index_tri])
+            loc_lst = locations.tolist()
+            ray_lst = index_ray.tolist()
+            tri_lst = index_tri.tolist()
 
             # find indices that trimesh failed on
             all_ray_indices = np.arange(len(origins))
--- a/pyvista/plotting/widgets.py
+++ b/pyvista/plotting/widgets.py
@@ -1609,7 +1609,7 @@ class WidgetHelper:
         spline_widget.PlaceWidget(bounds)
         spline_widget.SetResolution(resolution)
         if initial_points is not None:
-            spline_widget.InitializeHandles(pyvista.vtk_points((initial_points)))
+            spline_widget.InitializeHandles(pyvista.vtk_points(initial_points))
         else:
             spline_widget.SetClosed(closed)
         spline_widget.Modified()
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,6 +50,12 @@ repos:
     "typing-extensions==4.1.1",
   ]
 
+- repo: https://github.com/asottile/pyupgrade
+  rev: v3.2.2
+  hooks:
+  - id: pyupgrade
+    args: [--py37-plus, --keep-runtime-typing]
+
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.3.0
   hooks:
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,4 +1,5 @@
 import datetime
+import faulthandler
 import locale
 import os
 import sys
@@ -6,10 +7,7 @@ import sys
 # Otherwise VTK reader issues on some systems, causing sphinx to crash. See also #226.
 locale.setlocale(locale.LC_ALL, "en_US.UTF-8")
 
-if sys.version_info >= (3, 0):
-    import faulthandler
-
-    faulthandler.enable()
+faulthandler.enable()
 
 sys.path.insert(0, os.path.abspath("."))
 import make_external_gallery
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -1,5 +1,5 @@
 """Sub-classes and wrappers for vtk.vtkPointSet."""
-import collections.abc as collections
+import collections.abc
 from functools import wraps
 import numbers
 import os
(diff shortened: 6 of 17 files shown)


pyvista/plotting/widgets.py
```diff
@@ -1609,7 +1609,7 @@ def _the_callback(widget, event_id):
         spline_widget.PlaceWidget(bounds)
         spline_widget.SetResolution(resolution)
         if initial_points is not None:
-            spline_widget.InitializeHandles(pyvista.vtk_points((initial_points)))
+            spline_widget.InitializeHandles(pyvista.vtk_points(initial_points))
```
adeak, line 1612: Perhaps we should add a test since we're at it, if thete's one we can easily extend.

pyvista/core/datasetattributes.py
```diff
@@ -250,8 +250,7 @@ def __contains__(self, name: str) -> bool:
 
     def __iter__(self) -> Iterator[str]:
         """Implement for loop iteration."""
-        for array in self.keys():
-            yield array
+        yield from self.keys()
```
adeak, line 253: I wonder if this could just be `yield from self`.

pyvista/core/filters/poly_data.py
```diff
@@ -2144,7 +2144,7 @@ def multi_ray_trace(
         )
         if retry:
             # gather intersecting rays in lists
-            loc_lst, ray_lst, tri_lst = [arr.tolist() for arr in [locations, index_ray, index_tri]]
+            loc_lst, ray_lst, tri_lst = (arr.tolist() for arr in [locations, index_ray, index_tri])
```
adeak, line 2147: I'm a bit meh on this one. Sure, discarding the list is semantically worse. But it's the same amount of characters and even slightly faster than using the generator expression (214 ns vs 277 ns on my system for a dummy test, so hardly critical either way).

adeak (commented):
Good catch on most of these, thanks. I only have a few remarks.

pyvista/plotting/widgets.py
banesullivan: Could this be added in the pre-commit routine?

pyvista/core/filters/poly_data.py
banesullivan: I would split this up into three lines/statements. Comprehensions are great, but they don't need to be used everywhere as they can reduce the readability of code... like here IMO

pyvista/plotting/widgets.py
adeak: I meant for this line, because codecov said it is uncovered. But good point.

banesullivan: Ah, wasn't sure if that's what you meant or not. Both would be good to add

mwtoews: Do you mean to add a test for with a tuple arg, i.e. `vtk_points((initial_points,))`?

As for the separate thing about [pyupgrade as a pre-commit hook](https://github.com/asottile/pyupgrade#as-a-pre-commit-hook), is that welcome for this PR?

banesullivan: IMO, yes - paring the introduction of that hook with the changes that hook would make/evaluate is most appropriate.

> Do you mean to add a test for with a tuple arg, i.e. vtk_points((initial_points,))?

Not quite. Codecov is showing that the current tests for `add_spline_widget()` leave `initial_points` as `None`. We should expand `test_widget_spline` (in `tests/plotting/test_widgets.py`) to pass an appropriate value for that parameter to get coverage on that line

mwtoews:
I've added a pre-commit for `pyupgrade --py37-plus --keep-runtime-typing`. Another list comprehension to generator conversion was needed. The option `--keep-runtime-typing` was added to keep current typing, as I'm unsure of these changes. Anyone that wants to investigate, it only impacts one file:
```
$ pyupgrade --py37-plus pyvista/plotting/colors.py
```
<details open>
<summary>git diff</summary>

```diff
diff --git a/pyvista/plotting/colors.py b/pyvista/plotting/colors.py
index 6903e34f..86f4713e 100644
--- a/pyvista/plotting/colors.py
+++ b/pyvista/plotting/colors.py
@@ -460,10 +460,10 @@ class Color:
 
     def __init__(
         self,
-        color: Optional[color_like] = None,
-        opacity: Optional[Union[int, float, str]] = None,
-        default_color: Optional[color_like] = None,
-        default_opacity: Union[int, float, str] = 255,
+        color: color_like | None = None,
+        opacity: int | float | str | None = None,
+        default_color: color_like | None = None,
+        default_opacity: int | float | str = 255,
     ):
         """Initialize new instance."""
         self._red, self._green, self._blue, self._opacity = 0, 0, 0, 0
@@ -543,7 +543,7 @@ class Color:
         return h
 
     @staticmethod
-    def convert_color_channel(val: Union[int, np.integer, float, np.floating, str]) -> int:
+    def convert_color_channel(val: int | np.integer | float | np.floating | str) -> int:
         """Convert the given color channel value to the integer representation.
 
         Parameters
@@ -629,7 +629,7 @@ class Color:
                 raise ValueError(f"Invalid color name or hex string: {arg}") from None
 
     @property
-    def int_rgba(self) -> Tuple[int, int, int, int]:
+    def int_rgba(self) -> tuple[int, int, int, int]:
         """Get the color value as an RGBA integer tuple.
 
         Examples
@@ -655,7 +655,7 @@ class Color:
         return self._red, self._green, self._blue, self._opacity
 
     @property
-    def int_rgb(self) -> Tuple[int, int, int]:
+    def int_rgb(self) -> tuple[int, int, int]:
         """Get the color value as an RGB integer tuple.
 
         Examples
@@ -681,7 +681,7 @@ class Color:
         return self.int_rgba[:3]
 
     @property
-    def float_rgba(self) -> Tuple[float, float, float, float]:
+    def float_rgba(self) -> tuple[float, float, float, float]:
         """Get the color value as an RGBA float tuple.
 
         Examples
@@ -707,7 +707,7 @@ class Color:
         return self._red / 255.0, self._green / 255.0, self._blue / 255.0, self._opacity / 255.0
 
     @property
-    def float_rgb(self) -> Tuple[float, float, float]:
+    def float_rgb(self) -> tuple[float, float, float]:
         """Get the color value as an RGB float tuple.
 
         Examples
@@ -787,7 +787,7 @@ class Color:
         return self.hex_rgba[:-2]
 
     @property
-    def name(self) -> Optional[str]:
+    def name(self) -> str | None:
         """Get the color name.
 
         Returns
```
</details>
if these changes are good, then they can be committed and that arg can be removed from the pre-commit.

adeak:
@mwtoews I'm not quite sure how I feel about that diff about runtime typing. If I understand correctly, [`tuple[float, float, float]` needs 3.9](https://peps.python.org/pep-0585/) and [`int | None` needs 3.10](https://peps.python.org/pep-0604/), and the only reason this all works is that the module in question has `from __future__ import annotations` which [is was introduced in 3.7.0b1](https://docs.python.org/3/library/__future__.html). I don't know, for instance, how typing tools such as mypy consider python versions. Do they make assumptions about the python version they are installed in? Because then a 3.7 system (which we support) would not be able to correctly typecheck these annotations. Is this the case, or is this safe?

In other words, what are the exact risks associated with `--keep-runtime-typing`?

mwtoews:
@adeak the way I understand it is that `--keep-runtime-typing` essentially ignores most (all?) typing changes related to [PEP 585](https://peps.python.org/pep-0585/) and [PEP 604](https://peps.python.org/pep-0604/). I've been using Python for a few decades, but have avoided getting involved with annotations as it seems to change each Python release. I wouldn't know which runtime typing is correct. But it seems that pyupgrade is not honouring `--py37-plus` w.r.t. the runtime typing changes, so this is possibly a bug. This is a good reason to keep `--keep-runtime-typing` for now.

pyvista/plotting/widgets.py
adeak: Pre-commit hook is now added. @mwtoews how do you feel about adding the additional test case to make the diff coverage 100% here? Not critical for CI, just "would be nice to have" :)

mwtoews: @adeak how does 4e06e195e7ec29d0d87292f04168bb200ddb2a0b look?

adeak: Works beautifully, thanks!

adeak (approved):
Great work, thanks a lot @mwtoews!

banesullivan (approved):

PR #3642, opened 2022-11-23
Change unresolved BasePlotter intersphinx refs to Plotter

adeak:
I noticed that `BasePlotter` method references don't resolve in the docs:

screenshot from outside plotting.py showing red, unresolved pyvista.BasePlotter.subplot() reference

screenshot from within plotting.py showing red, unresolved BasePlotter.update reference

Switching the references to `Plotter` should fix the issue (existing links referring to `pyvista.Plotter`, or within `plotting.py` to `Plotter`, get resolved correctly).

### DIFF
--- a/examples/02-plot/multi-window.py
+++ b/examples/02-plot/multi-window.py
@@ -12,7 +12,7 @@ from pyvista import examples
 ###############################################################################
 # This example shows how to create a multi-window plotter by specifying the
 # ``shape`` parameter.  The window generated is a two by two window by setting
-# ``shape=(2, 2)``. Use the :func:`pyvista.BasePlotter.subplot` function to
+# ``shape=(2, 2)``. Use the :func:`pyvista.Plotter.subplot` function to
 # select the subplot you wish to be the active subplot.
 
 plotter = pv.Plotter(shape=(2, 2))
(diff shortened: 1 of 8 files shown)


examples/02-plot/multi-window.py
```diff
@@ -116,8 +116,9 @@
 
 plotter = pv.Plotter(shape=shape, row_weights=row_weights, col_weights=col_weights, groups=groups)
 
+###############################################################################
 # A grouped subplot can be activated through any of its composing cells using
-# the :func:`pyvista.BasePlotter.subplot` function.
+# the :func:`pyvista.Plotter.subplot` method.
```
adeak, line 121: This block looked like it was supposed to be outside the code blocks. I've added the break header, but we'll have to check the doc build artifact to make sure the result doesn't look weird in some way (I'm mostly concerned by having a `Plotter` created in the first block but shown in the second).

examples/02-plot/multi-window.py
```diff
@@ -12,7 +12,7 @@
 ###############################################################################
 # This example shows how to create a multi-window plotter by specifying the
 # ``shape`` parameter.  The window generated is a two by two window by setting
-# ``shape=(2, 2)``. Use the :func:`pyvista.BasePlotter.subplot` function to
+# ``shape=(2, 2)``. Use the :func:`pyvista.Plotter.subplot` function to
```
tkoyama010, line 15:
```suggestion
# ``shape=(2, 2)``. Use the :func:`pyvista.Plotter.subplot` method to
```

tkoyama010 (changes requested):

examples/02-plot/multi-window.py
adeak: Good catch, thanks!

examples/02-plot/multi-window.py
adeak: The original change broke the doc build so I've made this into a proper comment with no sphinx markup.

tkoyama010 (approved):
LGTM except for comments about block.

PR #3648, opened 2022-11-25
Don't raise on missing iren key events to clear by default

adeak:
Currently `RenderWindowInteractor.clear_events_for_key()` (wrapped by `Plotter.clear_events_for_key()`) will raise if we try to remove events for a key that has none:
```py
>>> import pyvista as pv
>>> pv.Plotter().clear_events_for_key('a')
  File "/home/adeak/pyvista/pyvista/plotting/plotting.py", line 1620, in clear_events_for_key
    self.iren.clear_events_for_key(key)
  File "/home/adeak/pyvista/pyvista/plotting/render_window_interactor.py", line 175, in clear_events_for_key
    self._key_press_event_callbacks.pop(key)
KeyError: 'a'
```
I think this would be better if it didn't raise by default, and if it did, raised something else. (I'm on the fence on whether this is actually a bug, labelling with "enhancement" for now.)

There are multiple potential approaches here, the first one I took for this PR is to change the default not to raise, and allow the old behaviour using a keyword argument. But still the error type is replaced. Many of these details could be changed, and I'm open for alternatives, this was just the most logical path for me.

After the patch:
```py
>>> import pyvista as pv
>>> pv.Plotter().clear_events_for_key('a')
>>> pv.Plotter().clear_events_for_key('a', raise_on_missing=True)
  File "/home/adeak/pyvista/pyvista/plotting/plotting.py", line 1615, in clear_events_for_key
    self.iren.clear_events_for_key(*args, **kwargs)
  File "/home/adeak/pyvista/pyvista/plotting/render_window_interactor.py", line 189, in clear_events_for_key
    raise ValueError(f'No events found for key {key!r}.') from None
ValueError: No events found for key 'a'.
```

Labelling this as a breaking change, if we decide to change what we do here then we'll have to reconsider whether the label is necessary.

### DIFF
--- a/pyvista/plotting/render_window_interactor.py
+++ b/pyvista/plotting/render_window_interactor.py
@@ -178,7 +178,7 @@ class RenderWindowInteractor:
         key : str
             Key to clear events for.
 
-        raise_on_missing : bool, default=False
+        raise_on_missing : bool, default: False
             Whether to raise a :class:`ValueError` if there are no events
             registered for the given key.
         """
--- a/pyvista/plotting/render_window_interactor.py
+++ b/pyvista/plotting/render_window_interactor.py
@@ -170,9 +170,23 @@ class RenderWindowInteractor:
         for observer in observers:
             self.remove_observer(observer)
 
-    def clear_events_for_key(self, key):
-        """Remove the callbacks associated to the key."""
-        self._key_press_event_callbacks.pop(key)
+    def clear_events_for_key(self, key, raise_on_missing=False):
+        """Remove the callbacks associated to the key.
+
+        Parameters
+        ----------
+        key : str
+            Key to clear events for.
+
+        raise_on_missing : bool, default=False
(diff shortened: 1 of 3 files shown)


pyvista/plotting/render_window_interactor.py
```diff
@@ -170,9 +170,23 @@ def remove_observers(self, event=None):
...
-        self._key_press_event_callbacks.pop(key)
+    def clear_events_for_key(self, key, raise_on_missing=False):
+        """Remove the callbacks associated to the key.
+
+        Parameters
+        ----------
+        key : str
+            Key to clear events for.
+
+        raise_on_missing : bool, default=False
```
adeak, line 181: I forgot that the final decision was the other one
```suggestion
        raise_on_missing : bool, default: False
```

banesullivan (approved):

PR #2904, opened 2022-06-29
Add `py.typed` Marker

adam-grant-hendry:
This will allow static type checkers to recognized the package's type hints per [PEP-561](https://peps.python.org/pep-0561/)

Fixes issue #2903

Issue #1136, opened 2021-02-01
[WIP] Conversion to type hints.


JevinJ:
This issue is to track progress on implementation/overhaul of type hints for all pyvista files.

/core/
- [ ] common.py
- [ ] composite.py
- [ ] datasetattributes.py
- [ ] errors.py
- [ ] filters.py
- [ ] grid.py
- [ ] objects.py
- [ ] pointset.py
- [ ] pyvista_ndarray.py

/plotting/
- [ ] background_renderer.py
- [ ] camera.py
- [ ] colors.py
- [ ] export_vtkjs.py
- [ ] helpers.py
- [ ] itkplotter.py
- [ ] lights.py
- [ ] mapper.py
- [ ] picking.py
- [ ] plotting.py
- [ ] renderer.py
- [ ] theme.py
- [ ] tools.py
- [ ] widgets.py

/utilities/
- [ ] cell_type_helper
- [ ] cells.py
- [ ] errors.py
- [ ] features.py
- [ ] fileio.py
- [ ] geometric_objects.py
- [ ] helpers.py
- [ ] misc.py
- [ ] parametric_objects.py
- [ ] regression.py
- [ ] sphinx_gallery.py
- [ ] transformations.py
- [ ] xvfb.py

banesullivan:
I beleive a lot of this has been implemented!

PR #3649, opened 2022-11-27
Add support to modulate the tube filter's radius in absolute units

jmargeta:
### Overview
This PR adds an argument to the tube filter so that the radius can be specified in absolute units for convenience.

```
spline.tube(scalars="radius", absolute=True)
```
Internally, it uses the [SetVaryRadiusToVaryRadiusByAbsoluteScalar](https://vtk.org/doc/nightly/html/classvtkTubeFilter.html#acdb459567a6d41696f3f29e8cd602639) of vtkTubeFilter.

### Details
 - support to modulate the tube's radius in absolute units, the default behavior remains unchanged
 - added an example to the create_spline section
 
Screenshot from 2022-11-27 01-16-29

### DIFF
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -1019,29 +1019,27 @@ class PolyDataFilters(DataSetFilters):
         scalars : str, optional
             Scalars array by which the radius varies.
 
-        capping : bool, optional
-            Turn on/off whether to cap the ends with polygons. Default
-            ``True``.
+        capping : bool, default: True
+            Turn on/off whether to cap the ends with polygons.
 
-        n_sides : int, optional
+        n_sides : int, default: 20
             Set the number of sides for the tube. Minimum of 3.
 
-        radius_factor : float, optional
+        radius_factor : float, default: 10
             Maximum tube radius in terms of a multiple of the minimum
             radius.
 
-        absolute : bool, optional
+        absolute : bool, default: False
             Vary the radius with values from scalars in absolute units.
-            Default ``False``.
 
-        preference : str, optional
+        preference : str, default: 'point'
             The field preference when searching for the scalars array by
             name.
 
-        inplace : bool, optional
+        inplace : bool, default: False
             Whether to update the mesh in-place.
 
-        progress_bar : bool, optional
+        progress_bar : bool, default: False
             Display a progress bar to indicate progress.
 
         Returns
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -999,7 +999,7 @@ class PolyDataFilters(DataSetFilters):
         scalars=None,
         capping=True,
         n_sides=20,
-        radius_factor=10,
+        radius_factor=10.0,
         absolute=False,
         preference='point',
         inplace=False,
@@ -1025,7 +1025,7 @@ class PolyDataFilters(DataSetFilters):
         n_sides : int, default: 20
             Set the number of sides for the tube. Minimum of 3.
 
-        radius_factor : float, default: 10
+        radius_factor : float, default: 10.0
             Maximum tube radius in terms of a multiple of the minimum
             radius.
 
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -1000,6 +1000,7 @@ class PolyDataFilters(DataSetFilters):
         capping=True,
         n_sides=20,
         radius_factor=10,
+        absolute=False,
         preference='point',
         inplace=False,
         progress_bar=False,
@@ -1028,6 +1029,10 @@ class PolyDataFilters(DataSetFilters):
         radius_factor : float, optional
             Maximum tube radius in terms of a multiple of the minimum
             radius.
+
+        absolute : float, optional
+            Vary the radius with values from scalars in absolute units.
+            Default ``False``.
 
         preference : str, optional
             The field preference when searching for the scalars array by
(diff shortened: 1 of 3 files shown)


pyvista/core/filters/poly_data.py
```diff
@@ -1080,7 +1085,10 @@ def tube(
             field = poly_data.get_array_association(scalars, preference=preference)
             # args: (idx, port, connection, field, name)
             tube.SetInputArrayToProcess(0, 0, 0, field.value, scalars)
-            tube.SetVaryRadiusToVaryRadiusByScalar()
+            if not absolute:
+                tube.SetVaryRadiusToVaryRadiusByScalar()
+            else:
+                tube.SetVaryRadiusToVaryRadiusByAbsoluteScalar()
```
tkoyama010, line 1091: It is better not to use "not" in if sentence.
```suggestion
            if absolute:
                tube.SetVaryRadiusToVaryRadiusByAbsoluteScalar()
            else:
                tube.SetVaryRadiusToVaryRadiusByScalar()
```

pyvista/core/filters/poly_data.py
```diff
@@ -1029,6 +1030,10 @@ def tube(
             Maximum tube radius in terms of a multiple of the minimum
             radius.
 
+        absolute : float, optional
```
tkoyama010, line 1033:
```suggestion
        absolute : bool, optional
```

tkoyama010 (changes requested):
Thanks for this. Great contribution.

tkoyama010 (dismissed):
LGTM

jmargeta:
Thanks a lot for the speedy review, @tkoyama010 !

pyvista/core/filters/poly_data.py
```diff
@@ -1029,6 +1030,10 @@ def tube(
             Maximum tube radius in terms of a multiple of the minimum
             radius.
 
+        absolute : bool, optional
```
adeak, line 1033: Side note: since we're touching this I think it would make sense to update this method's docstring to use the new `optional`/`default` convention described in https://dev.pyvista.org/extras/developer_notes.html#docstrings.

jmargeta: Thanks for the pointer @adeak. I agree with you. Please, see the next commit, which uses the new `default` convention for all parameters of the method.

tkoyama010 (dismissed):

tests/test_polydata.py
```diff
@@ -744,6 +744,10 @@ def test_tube(spline):
     tube = spline.tube(radius=0.5, scalars='arc_length', progress_bar=True)
     assert tube.n_points, tube.n_cells
 
+    # Complicated with absolute radius
+    tube = spline.tube(radius=0.5, scalars='arc_length', absolute=True, progress_bar=True)
+    assert tube.n_points, tube.n_cells
```
adeak, line 749: These asserts along with the existing ones in the test are buggy:
```py
n_points = 1
n_cells = 0
assert n_points, n_cells
# no error
```
I think we can get away with
```suggestion
    assert tube.n_points and tube.n_cells
```
but please fix it in the rest of this test :)

pyvista/core/filters/poly_data.py
```diff
@@ -1018,25 +1019,27 @@ def tube(
...
-            ``True``.
+        capping : bool, default: True
+            Turn on/off whether to cap the ends with polygons.
 
-        n_sides : int, optional
+        n_sides : int, default: 20
             Set the number of sides for the tube. Minimum of 3.
 
-        radius_factor : float, optional
+        radius_factor : float, default: 10
```
adeak, line 1028: And since I'm already asking for revisions: it's a bit dirty to claim this is a `float` and use `10` (if someone were to infer types from the default, they might think this is an int). For purity's sake we should probably change the default to `10.0`.

adeak (commented):
Added things look great overall, just two (final) unrelated issues revealed by your PR (one only a nitpick) :)

tests/test_polydata.py
jmargeta: Great catch, thank you!

adeak (approved):
Looks good to me, thanks a lot @jmargeta!

pyvista/core/filters/poly_data.py
jmargeta: Super, thanks for your feedback. This makes sense.

Out of curiosity, some parts of PyVista use type hints for the arguments (in addition to the types in the docstrings) and some not. Are type hints intentionally avoided, or is it just a matter of convenience?
I do not see this discussed in the Contributing docs.

adeak: As far as I'm concerned typing happens to other people, and I personally find type hints a lot harder to read :sweat_smile: But at the same time I acknowledge that it can help downstream users using an IDE (I don't).

Apparently we have an older tracking issue for adding more type hints: https://github.com/pyvista/pyvista/issues/1136

And @tkoyama010 mentioned wanting to increase the coverage of type hints in https://github.com/pyvista/pyvista/discussions/2133#discussioncomment-2210496.

tkoyama010 (approved):

pyvista/core/filters/poly_data.py
jmargeta: Yeah, I understand, even if I consider myself to be more in the IDE/pro-type-hints camp, haha.
It is so great that both ways can happily co-exist in the codebase :) Thanks for the links to the issues.

adeak: For completeness' sake here's Alex' take on the matter from an issue I overlooked on my first search https://github.com/pyvista/pyvista/pull/2904#issuecomment-1169968918:

> Let it be known that type annotations is still a work in progress and we don't have it in most places. adam-grant-hendry, if you feel up to it, you can feel free to add them. I'd do it slowly across several PRs. Tracking down `mypy` errors usually drives me up a wall.

PR #3656, opened 2022-11-28
Support axes customization

### LINKED ISSUE #3325: Customization limitations when creating a pyvista.Axes actor
🐞 **Problem**
While working on a project, I required to customize and render the axes of a reference frame. From the official documentation, I saw the following functions and classes:

**Plotter.add_axes**

```python
# Returns a vtkmodules.vtkRenderingAnnotation.vtkAxesActor
# PROS: object can be easily customized
# CONS: only acting as a widget in the left-down corner
axes = pv.Plotter().add_axes()
axes.GetXAxisShaftProperty().SetColor(1, 0, 0)
```

**Plotter.add_axes_at_origin**

```python
# Returns a vtkmodules.vtkRenderingAnnotation.vtkAxesActor
# PROS: object can be easily customized
# CONS: already rendered in plotter once created
axes = pv.Plotter().add_axes_at_origin()
axes.GetXAxisShaftProperty().SetColor(1, 0, 0)
```

**pyvista.Axes**

```python
# Returns a pyvista.Axes containing that exposes and actor through pyvsita.Axes.actor
# PROS: intuitive to create
# CONS: cannot different methods for customizing the axes actor
axes = pv.Axes()

# BUG: this line fail as axes.actor does not contain GetXAxisShaftProperty
axes.actor.GetXAxisShaftProperty().SetColor(1, 0, 0)

pv.Plotter().add_actor(axes.actor)
```

**pyvista.plotting.tools.create_axes_marker**
```python
# Returns a vtkmodules.vtkRenderingAnnotation.vtkAxesActor
# PROS: object can be customized directly through the function
# CONS: not a PyVsita object??
axes = create_axes_marker(**axes_customization_options)
pv.Plotter().plot_actor(axes)
```

💡 **Exposing customization options into the pv.Axes class**
From previous examples, it is seen that the `pv.Axes` class does not allow to access most of the customization options. It is a bit surprising as I thought this was the way to go when creating user defined axes for a scene.

AlejandroFernandezLuces:
### Overview

This pull request resolves #3325. I expose the object from VTK that allows the customization of the axes.

### Code example

```py
import pyvista as pv
axes = pv.Axes()
axes.axes_actor.GetZAxisShaftProperty().SetColor(0, 1, 1)
axes.axes_actor.SetShaftTypeToCylinder()
pl = pv.Plotter()
pl.add_actor(axes.axes_actor)
pl.add_mesh(pv.Sphere())
pl.show() 

```
### Result
image


pyvista/plotting/axes.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, show_actor=False, actor_scale=1, line_width=1.0, symmetric=Fa
         self.mapper.SetInputConnection(self.GetOutputPort())
         # Add the axes actor
         self.actor = Actor(mapper=self.actor)
+        self.axes_actor = _vtk.vtkAxesActor()
```
jorgepiloto, line 45: Although this works, why having two actors within the same class, see `self.actor` and `self.axes_actor`.

Ideally, it should be possible to do:

```python
axes = pv.Axes()
axes.actor.GetXAxisShaftProperty().SetColor(1, 0, 0)
```

instead of what the current implementation provides:

```python
axes = pv.Axes()
axes.axes_actor.GetXAxisShaftProperty().SetColor(1, 0, 0)

**Review by jorgepiloto — CHANGES_REQUESTED** [review]

**Comment by jorgepiloto** [comment]
I am not so familiar with the internals of PyVista, but this may affect the `Actor` class to properly expose all the VTK properties for the axes.

File: pyvista/plotting/axes.py (hunk 1 above)
Comment (AlejandroFernandezLuces) [reply pyvista/plotting/axes.py:45]: In VTK, `vtkActor` and `vtkAxesActor` are two different things. If you check on [VTK documentation for `vtkAxesActor`](https://vtk.org/doc/nightly/html/classvtkAxesActor.html), you can see that `vtkActor` is just a dependency of `vtkAxesActor`, not a parent class as you would expect. 

So, despite being able to make `Actor` class inherit from `vtkAxesActor`, I don't think that it is correct, since they are two different and somewhat unrelated classes. It would cause that if you build an `Actor` object, unrelated to the axes, that object could access the axes properties, which is prone to errors when developing.

This is the reason I decided to set two `Actor` attributes. To avoid this, I could create a wrapper for `vtkAxesActor`, which I think that could work to substitute both `Actor` attributes. Anyway, let me know what do you think 😄

File: pyvista/plotting/axes.py [hunk 2]
```diff
@@ -42,6 +42,7 @@ def __init__(self, show_actor=False, actor_scale=1, line_width=1.0, symmetric=Fa
         self.mapper.SetInputConnection(self.GetOutputPort())
         # Add the axes actor
         self.actor = Actor(mapper=self.actor)
```
banesullivan, line 44: Drive by comment but should this line be?:

```suggestion
        self.actor = Actor(mapper=self.mapper)
```

pyvista/plotting/axes.py
banesullivan: Thanks for the thorough investigation, @AlejandroFernandezLuces! I'm curious why this object isn't used anywhere else though?

MaxJPRey:
@AlejandroFernandezLuces If we are using `axes.axes_actor`, could you update the documentation and examples.
Because most people could use `axes.actor` instead of `pl.add_actor(axes.axes_actor)` for instance. The code snippet you mentioned in the description of this PR could help to clarify any doubts from user.

pyvista/plotting/axes.py
jorgepiloto: Thanks for the deep explanation this, @AlejandroFernandezLuces.

pyvista/plotting/actor.py
```diff
@@ -591,33 +591,34 @@ def center(self) -> tuple:
...
+                >>> import pyvista as pv
+                >>> plane = pv.Plane(i_size=1.5, j_size=1.5)
+                >>> mesh = pv.Sphere().clip_surface(plane, invert=False)
+                >>> pl = pv.Plotter()
+                >>> actor = pl.add_mesh(mesh, smooth_shading=True)
+                >>> actor.backface_prop.color = 'lightblue'
+                >>> _ = pl.add_mesh(
+                ...     plane, opacity=0.25, show_edges=True, color='grey', lighting=False,
+                ... )
+                >>> pl.show()
```
MaxJPRey, line 621: Was the indent introduces on purpose?
It should fix the documentation action issue removing it.

pyvista/plotting/axes.py
AlejandroFernandezLuces: `vtkAxesActor` is wrapped in the following function, which is used for processing the [parameters in `Plotter.add_axes()`](https://docs.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.add_axes.html#add-axes). I suppose this customization is not what @jorgepiloto is looking for.


https://github.com/pyvista/pyvista/blob/25a482b4d1abc6c8472b962ba657cc03a376e707/pyvista/plotting/tools.py#L106-L244

akaszynski:
@AlejandroFernandezLuces, can you raise codecov to 90%+?

pyvista/plotting/actor.py
```diff
@@ -592,15 +592,16 @@ def backface_prop(self) -> Optional['pv.Property']:
         """Return or set the backface property.
 
         By default this property matches the frontface property
-        :attr:`Actor.prop`. Once accessed or modified, this backface
+        :attpl.add_mesh(pv.Sphere())
```
akaszynski, line 595: Fix this.

pyvista/plotting/axes_actor.py
```diff
@@ -0,0 +1,52 @@
...
+import pyvista as pv
+
+
+class AxesActor(pv._vtk.vtkAxesActor):
+    """Axes actor wrapper for vtkAxesActor.
+
+    Hybrid 2D/3D actor used to represent 3D axes in a scene. The user
+    can define the geometry to use for the shaft or the tip, and the
+    user can set the text for the three axes. To see full customization
+    options, refer to `VTK axes actor documentation
```
akaszynski, line 11:
```suggestion
    options, refer to `vtkAxesActor Details
```

akaszynski:
Actually, ignore the coverage failure, that's no your issue. However, I have a few other comments regarding the docs. Ping again when these are fixed.

AlejandroFernandezLuces:
Doc issues fixed @akaszynski

pyvista/plotting/axes_actor.py
```diff
@@ -0,0 +1,52 @@
...
+        super().__init__()
+
+    @property
+    def visibility(self) -> bool:
+        """Return or set AxesActor visibility.
+
+        Examples
+        --------
+        Create an Axes object and then change the
+        visibility of its AxesActor.
```
MaxJPRey, line 40: @AlejandroFernandezLuces I don't think you are changing the visibility of the axes but just getting the value.

banesullivan:
open/close cycle to fix CI. @AlejandroFernandezLuces, is this ready for final review and merge?

AlejandroFernandezLuces:
This is ready for a final review on my side @banesullivan .

pyvista/plotting/axes_actor.py
```diff
@@ -0,0 +1,52 @@
...
+    >>> axes.axes_actor.GetZAxisShaftProperty().SetColor(0, 1, 1)
+    >>> axes.axes_actor.SetShaftTypeToCylinder()
+    >>> pl = pv.Plotter()
+    >>> _ = pl.add_actor(axes.axes_actor)
+    >>> _ = pl.add_mesh(pv.Sphere())
+    >>> pl.show()
+
+    """
+
+    def __init__(self, mapper=None, prop=None):
```
banesullivan, line 29: Something needs to be done with these unused arguments

pyvista/plotting/axes_actor.py
```diff
@@ -0,0 +1,52 @@
...
+    <https://vtk.org/doc/nightly/html/classvtkAxesActor.html#details>`
+
+    Examples
+    --------
+    Customize the axis shaft color and shape.
+
+    >>> import pyvista as pv
+    >>> axes = pv.Axes()
+    >>> axes.axes_actor.GetZAxisShaftProperty().SetColor(0, 1, 1)
+    >>> axes.axes_actor.SetShaftTypeToCylinder()
```
banesullivan, line 21: Let's expose these methods on this class

banesullivan: If you're up for exposing the property as well, great but not totally necessary IMO

banesullivan (changes requested):

pyvista/plotting/axes_actor.py
```diff
@@ -0,0 +1,52 @@
...
+    >>> pl.show()
+
+    """
+
+    def __init__(self, mapper=None, prop=None):
+        """Initialize actor."""
+        super().__init__()
+
+    @property
+    def visibility(self) -> bool:
```
MaxJPRey, line 34:
```suggestion
    def is_visible(self) -> bool:
```

banesullivan: This should map to the equivalent VTK property. This is correct as-is.

pyvista/plotting/axes_actor.py
AlejandroFernandezLuces: I exposed the relevant methods of the class. If we want to expose the properties as well, I think we should do it in another PR, since this is getting too big already.

Take a look and let me know what do you think 😄

banesullivan: Great! Thanks! I will give this another pass soon here

pyvista/plotting/axes_actor.py
```diff
@@ -0,0 +1,399 @@
...
+    @shaft_type.setter
+    def shaft_type(self, shaft_type: int):
+        if shaft_type == ShaftType.CYLINDER:
+            self.SetShaftTypeToCylinder()
+        elif shaft_type == ShaftType.LINE:
+            self.SetShaftTypeToLine()
+
+    @property
+    def tip_type(self) -> TipType:
+        """Return or set the shaft type. Can be either a cone(0) or a sphere(1).
```
MaxJPRey, line 327:
```suggestion
        """Return or set the tip type. Can be either a cone(0) or a sphere(1).
```

pyvista/plotting/axes_actor.py
akaszynski: > If we want to expose the properties as well, I think we should do it in another PR

Laughs in 2000 LOC PRs.

I'll cover this. It will be less work in this PR.

akaszynski:
Thanks for implementing the properties @AlejandroFernandezLuces.

.github/workflows/testing-and-deployment.yml
```diff
@@ -191,7 +191,7 @@ jobs:
           pip list
 
       - name: Unit Testing
-        run: xvfb-run coverage run -m --source=pyvista --module pytest --skip_image_cache_vtk8 --fail_extra_image_cache -v
+        run: xvfb-run coverage run -m --source=pyvista --module pytest --skip_image_cache_vtk8 --fail_extra_image_cache -v --add_missing_images
```
tkoyama010, line 194: Could you give an explanation of the reason for adding this for me?

AlejandroFernandezLuces: I'm getting an error when adding a new test that needs image comparison:
>FAILED tests/plotting/test_plotting.py::test_axes_actor_properties - RuntimeError: tests/plotting/image_cache/axes_actor_properties.png does not exist in image cache

This flag should generate the missing images without the need of resetting the whole cache:
> `--add_missing_images` adds any missing images from the test run to the cache.

but is not working in the repository for some reason, I'm trying to find out why.

tkoyama010: I think you need to [add an image for regression tests](https://docs.pyvista.org/extras/developer_notes.html#notes-regarding-image-regression-testing).

AlejandroFernandezLuces: The image was not being generated because the flag `--fail_extra_image` prevents the code from reaching the  `--add_missing_images` if branch. You can see it in this piece of code from `pytest-pyvista`:

```python

if self.add_missing_images and not os.path.isfile(image_filename):
    plotter.screenshot(image_filename)

```

This is probably an undesired behaviour in `pytest-pyvista`, I will open an issue to discuss it there.

banesullivan: `--fail_extra_image_cache` `--add_missing_images` are incompatible together. You can either fail on missing or generate new images. Further, we always want the CI to fail on missing and encourage developers to check in new images with the code

AlejandroFernandezLuces: I think this should be properly documented in any case. Also, maybe `pytest-pyvista` could raise at least a warning regarding you have both flags active.

.github/workflows/testing-and-deployment.yml
```diff
@@ -191,7 +191,7 @@ jobs:
           pip list
 
       - name: Unit Testing
-        run: xvfb-run coverage run -m --source=pyvista --module pytest --skip_image_cache_vtk8 --fail_extra_image_cache -v
+        run: xvfb-run coverage run -m --source=pyvista --module pytest --skip_image_cache_vtk8 --fail_extra_image_cache -v
```
banesullivan, line 194: Pesky trailing space needs to be removed...
```suggestion
        run: xvfb-run coverage run -m --source=pyvista --module pytest --skip_image_cache_vtk8 --fail_extra_image_cache -v
```

tests/plotting/test_plotting.py
```diff
@@ -3371,3 +3375,40 @@ def test_color_cycler_names(name):
...
+    assert axes_actor.x_axis_shaft_properties.color == (1, 1, 1)
+    axes_actor.y_axis_shaft_properties.metallic = 0.2
+    assert axes_actor.y_axis_shaft_properties.metallic == 0.2
+    axes_actor.z_axis_shaft_properties.roughness = 0.3
+    assert axes_actor.z_axis_shaft_properties.roughness == 0.3
+
+    axes_actor.x_axis_tip_properties.anisotropy = 0.4
+    assert axes_actor.x_axis_tip_properties.anisotropy == 0.4
+    axes_actor.x_axis_tip_properties.anisotropy_rotation = 0.4
+    assert axes_actor.x_axis_tip_properties.anisotropy == 0.4
```
tkoyama010, line 3395:
```suggestion
    assert axes_actor.x_axis_tip_properties.anisotropy_rotation == 0.4
```

tests/test_axes.py
```diff
@@ -39,3 +40,109 @@ def test_symmetric():
...
+def test_axes_actor_sphere_rad():
+    axes = pyvista.Axes()
+    axes.axes_actor.sphere_radius = 0.8
+    assert axes.axes_actor.sphere_radius == 0.8
+
+
+def test_axes_actor_cylinder_rad():
+    axes = pyvista.Axes()
+    axes.axes_actor.sphere_radius = 0.03
+    assert axes.axes_actor.sphere_radius == 0.03
```
tkoyama010, line 121:
```suggestion
    axes.axes_actor.cylinder_radius= 0.03
    assert axes.axes_actor.cylinder_radius== 0.03
```

AlejandroFernandezLuces:
I think this PR is done after the addition of the properties. Let me know what do you think!

pyvista/plotting/actor_properties.py
```diff
@@ -0,0 +1,149 @@
...
+    >>> axes.axes_actor.SetShaftTypeToCylinder()
+    >>> pl = pv.Plotter()
+    >>> pl.add_actor(axes.axes_actor)   # doctest:+SKIP
+    >>> pl.add_mesh(pv.Sphere())   # doctest:+SKIP
+    >>> pl.show()   # doctest:+SKIP
+    """
+
+    def __init__(self, properties: pv._vtk.vtkProperty) -> None:
+        super().__init__()
+        self.properties = properties()
```
banesullivan, line 50: This is really odd to me. Why is a bound method often passed to this constructor and not the object itself? The type annotation implies that a `vtkProperty` instance should be passed but this is used where a bound method that produces that instance is passed.

Is this because of being unable to wrap/shallow copy `vtkProperty`?

pyvista/plotting/actor_properties.py
```diff
@@ -0,0 +1,149 @@
...
+
+class ReprModel(Enum):
+    """Types of representations the models can have."""
+
+    POINTS = 0
+    WIREFRAME = 1
+    SURFACE = 2
+
+
+class ActorProperties:
```
banesullivan, line 24: We might want to leverage the default colors for the active theme here. I expect the text color to match that of the current theme by default, but it doesnt

banesullivan (changes requested):
@AlejandroFernandezLuces, this is coming along nicely! I pushed some of my own changes for immediate feedback I had. This included:

- An example of how to use with `add_orientation_widget` as that's what most people probably want to do
- Moving of the `ShaftType` and `TipType` enums because they were a a pain to access as a user. We probably want to do this for `ShaderModel` and `ReprModel` as well.

### Feedback

Overall this is looking excellent! Problem is this is not a simple addition and thus I have some thoughts:

- ActorProperties is really odd (see inline comment)
- We may want different names for `ShaderModel` and `ReprModel`
- We may want to place `ShaderModel` and `ReprModel` enum classes in a better place (these would be nice to use in `add_mesh` as well.
- The current implementation ignores the active theme. We should see defaults for ActorProperties based on the current active theme.
- I'd like to see more image regression testing to make sure the `AxesActor` is actually taking on the properties being set and not just validating that the wrapper class has the set value (this could easily break)

pyvista/plotting/actor_properties.py
AlejandroFernandezLuces: I was having issues with maintaining the information I set in the properties while doing this. Passing the object works fine, it is solved now.

AlejandroFernandezLuces:
> * I'd like to see more image regression testing to make sure the `AxesActor` is actually taking on the properties being set and not just validating that the wrapper class has the set value (this could easily break)

@banesullivan I'm already doing an image regression test for this: 
https://github.com/pyvista/pyvista/blob/225ac8a866dcf52a9a04d6b9cd061e247bb192ba/tests/plotting/test_plotting.py#L3438-L3441

This image regression test captures all of the properties that were changed in the previous code, you can see it in this image:
https://github.com/pyvista/pyvista/blob/feat/axis_customization/tests/plotting/image_cache/axes_actor_properties.png

I think this should be enough for testing the properties, since we test every property in one image. Let me know If I need to add more cases.

pyvista/plotting/actor_properties.py
AlejandroFernandezLuces: I used the active theme color in `AxesActor` class, since each axis should have a different color in the theme. 
https://github.com/pyvista/pyvista/blob/225ac8a866dcf52a9a04d6b9cd061e247bb192ba/pyvista/plotting/axes_actor.py#L69-L89

akaszynski:
> I think this should be enough for testing the properties, since we test every property in one image. Let me know If I need to add more cases.

Provided it's most/all of the properties, I'd prefer this to be in one image. Ideally one set of unit tests that call the individual attributes, and another plotter test for "functional testing".

akaszynski (approved):
LGTM, though @banesullivan should weigh in here.

A few comments regarding testing:
- Use fixtures whenever possible. It reduces code duplication.
- Test everything. I tried writing tests for the setter for `z_axis_shaft_properties` and discvered there's no VTK `SetZAxisShaftProperty`. I ended up removing those setters because it actually was raising a recursive error:

```
axes = <Axes(0x3595fe0) at 0x7f7a38fd0ac0>

    def test_axes_actor_x_prop(axes):
        assert isinstance(axes.axes_actor.x_axis_shaft_properties, ActorProperties)
        prop = ActorProperties(pv._vtk.vtkProperty())
>       axes.axes_actor.x_axis_shaft_properties = prop

tests/test_axes.py:140: 
pyvista/plotting/axes_actor.py:457: in x_axis_shaft_properties
    self.x_axis_shaft_properties = properties
pyvista/plotting/axes_actor.py:457: in x_axis_shaft_properties
    self.x_axis_shaft_properties = properties
E   RecursionError: maximum recursion depth exceeded
!!! Recursion detected (same locals & position)
tests::test_axes.py::test_axes_actor_x_prop   - 1008.0 KB
tests::test_axes.py::test_axes_actor_tip_len  - 504.0 KB
FAILED tests/test_axes.py::test_axes_actor_x_prop - RecursionError: maximum recursion depth exceeded                                  
```

PR #3658, opened 2022-11-28
Add interrogate to .pre-commit-config.yaml

tkoyama010:
### Overview

Add [interrogate](https://interrogate.readthedocs.io/en/latest/): measure docstring coverage to .pre-commit-config.yaml.
<p align="center">
    interrogate logo pink
</p>

 

### Details

- We have already passed coverage so there may not be a need to add it, but I decided to suggest it for future coverage drops.
- An error will occur if coverage of 80% (the recommended value for interrogate) is not met.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,6 +50,13 @@ repos:
     "typing-extensions==4.1.1",
   ]
 
+- repo: https://github.com/econchick/interrogate
+  rev: 1.5.0
+  hooks:
+  - id: interrogate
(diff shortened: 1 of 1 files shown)


banesullivan:
Can you share a bit about how this is different than the existing pydocstyle check? I think this covers the same stuff but I'm not familiar -- hoping for some clarification

banesullivan:
I'm closing for now due to inactivity and lack of clarity. I'm hesitant to add more pre-commit checks as these can add developer friction. Please open an issue outlining the reasoning for adding this to move forward. I'd like for new pre-commit checks to have buy-in/support from multiple developers before opening a PR

PR #3659, opened 2022-11-28
Remove legacy numpy aliases removed in 1.24.0

adeak:
NumPy 1.24.0rc1 is out. We have a handful of test failures arising from the inevitable removal of legacy aliases:
```py
# numpy<1.24.0
print(all([np.int is int, np.float is float, np.bool is bool, np.complex is complex]))
# True; plus a screenful of deprecation warnings
```

We can fix some of these that have snuck back into PyVista since the last purge, but there's also an issue that ITK (at least 0.32.3 which is the newest one matching our requirements for Python < 3.10) breaks due to this:
```py
>>> import itkwidgets
In the future `np.bool` will be defined as the corresponding NumPy scalar.  (This may have returned Python scalars in past versions.
  File "/home/adeak/tmp_test_env_py39/lib/python3.9/site-packages/itkwidgets/__init__.py", line 13, in <module>
    from .widget_viewer import Viewer, view
  File "/home/adeak/tmp_test_env_py39/lib/python3.9/site-packages/itkwidgets/widget_viewer.py", line 14, in <module>
    import itk
  File "/home/adeak/tmp_test_env_py39/lib/python3.9/site-packages/itk/__init__.py", line 28, in <module>
    from itk.support.extras import *
  File "/home/adeak/tmp_test_env_py39/lib/python3.9/site-packages/itk/support/extras.py", line 35, in <module>
    import itk.support.types as itkt
  File "/home/adeak/tmp_test_env_py39/lib/python3.9/site-packages/itk/support/types.py", line 175, in <module>
    ) = itkCType.initialize_c_types_once()
  File "/home/adeak/tmp_test_env_py39/lib/python3.9/site-packages/itk/support/types.py", line 140, in initialize_c_types_once
    _B: "itkCType" = itkCType("bool", "B", np.bool)
  File "/home/adeak/tmp_test_env_py39/lib/python3.9/site-packages/numpy/__init__.py", line 284, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'bool'
```

This breaks our `tests/jupyter/test_itk_plotting.py`, and it breaks `pyvista.Report()` due to its import of `itkwidgets`. We'll have to wait with bumping 1.24 until upstream fixes this, until then we should make sure our own codebase stays clean.

### DIFF
--- a/tests/test_grid.py
+++ b/tests/test_grid.py
@@ -17,9 +17,9 @@ from pyvista.utilities.misc import PyVistaDeprecationWarning
 test_path = os.path.dirname(os.path.abspath(__file__))
 
 # must be manually set until pytest adds parametrize with fixture feature
-HEXBEAM_CELLS_BOOL = np.ones(40, np.bool_)  # matches hexbeam.n_cells == 40
-STRUCTGRID_CELLS_BOOL = np.ones(729, np.bool_)  # struct_grid.n_cells == 729
-STRUCTGRID_POINTS_BOOL = np.ones(1000, np.bool_)  # struct_grid.n_points == 1000
+HEXBEAM_CELLS_BOOL = np.ones(40, dtype=bool)  # matches hexbeam.n_cells == 40
+STRUCTGRID_CELLS_BOOL = np.ones(729, dtype=bool)  # struct_grid.n_cells == 729
+STRUCTGRID_POINTS_BOOL = np.ones(1000, dtype=bool)  # struct_grid.n_points == 1000
 
 pointsetmark = pytest.mark.skipif(
     pyvista.vtk_version_info < (9, 1, 0), reason="Requires VTK>=9.1.0 for a concrete PointSet class"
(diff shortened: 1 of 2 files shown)


tests/test_grid.py
```diff
@@ -1177,7 +1177,7 @@ def test_hide_cells(ind, struct_grid):
     assert out.HasAnyBlankCells()
 
     with pytest.raises(ValueError, match='Boolean array size must match'):
-        struct_grid.hide_cells(np.ones(10, dtype=np.bool), inplace=True)
+        struct_grid.hide_cells(np.ones(10, dtype=bool), inplace=True)
```
adeak, line 1180: This was one of two actually problematic instances in this module, but I think `dtype=bool` is always preferable to `dtype=np.bool_` and especially just `np.bool_` as a positional argument. That's why I changed all the others as well in this module

tkoyama010 (approved):
LGTM

adeak:
For posterity: I've upgraded my environment, and I have no more errors following this PR (I had very stale `scooby` and moderately stale ITK packages). We should be good to go when dependabot bumps to NumPy 1.24.0 once it's released.

PR #3683, opened 2022-12-05
Fix the docstring of demos.py

tkoyama010:
### Overview

Fix the docstring of demos.py.

 

### Details

- None

### DIFF
--- a/pyvista/demos/demos.py
+++ b/pyvista/demos/demos.py
@@ -443,7 +443,7 @@ def plot_datasets(dataset_type=None):
 
     Parameters
     ----------
-    notebook : bool, optional
+    dataset_type : str, optional
         If set, plot just that dataset.  Must be one of the following:
 
         * ``'PolyData'``
--- a/pyvista/demos/demos.py
+++ b/pyvista/demos/demos.py
@@ -225,7 +225,7 @@ def plot_wave(fps=30, frequency=1, wavetime=3, interactive=False, notebook=None)
         Allows the user to set the camera position before the start of the
         wave movement.
 
-    notebook : bool, default: None
+    notebook : bool, optional
         When ``True``, the resulting plot is placed inline a jupyter
         notebook.  Assumes a jupyter console is active.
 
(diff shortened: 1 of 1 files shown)


pyvista/demos/demos.py
```diff
@@ -212,20 +212,20 @@ def plot_wave(fps=30, frequency=1, wavetime=3, interactive=False, notebook=None)
...
+        The desired total display time in seconds.
 
-    interactive : bool, optional
+    interactive : bool, default: False
         Allows the user to set the camera position before the start of the
-        wave movement.  Default ``False``.
+        wave movement.
 
-    notebook : bool, optional
+    notebook : bool, default: None
```
MatthewFlamm, line 228: This should stay optional if default is correct here and it is unclear what happens in that case to me.  Based on the docstring alone and the docstring of Plotter I'm unsure whether we can just just default of `False`, which would be more clear.

This comment applies to 3 more changes too.

banesullivan:
```suggestion
    notebook : bool, optional
```

pyvista/demos/demos.py
```diff
@@ -345,7 +345,7 @@ def plot_ants_plane(notebook=None):
 
     Parameters
     ----------
-    notebook : bool, optional
+    notebook : bool, default: None
```
banesullivan, line 348:
```suggestion
    notebook : bool, optional
```

pyvista/demos/demos.py
```diff
@@ -387,7 +387,7 @@ def plot_beam(notebook=None):
...
```
banesullivan, line 390:
```suggestion
    notebook : bool, optional
```

pyvista/demos/demos.py
```diff
@@ -443,7 +443,7 @@ def plot_datasets(dataset_type=None):
 
     Parameters
     ----------
-    dataset_type : str, optional
+    dataset_type : str, default: None
```
banesullivan, line 446:
```suggestion
    notebook : bool, optional
```

banesullivan (dismissed):

pyvista/demos/demos.py
```diff
@@ -443,7 +443,7 @@ def plot_datasets(dataset_type=None):
 
     Parameters
     ----------
-    dataset_type : str, optional
+    notebook : bool, optional
```
banesullivan, line 446:
```suggestion
    dataset_type : str, optional
```

banesullivan (approved):

PR #3687, opened 2022-12-06
Fix the number error of segment in add_lines

### LINKED ISSUE #3667: add_lines behaves different than documentation suggests

[The documentation of `add_lines` ](https://docs.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.add_lines.html) states that an array of points like `np.array([[0, 0, 0], [1, 0, 0], [1, 0, 0], [1, 1, 0]])` adds two line segments, probably from point 0 to point 1 and from point 2 to point 3.
However, three line segments are added, one between each point in the sequence. It behaves like `lines_from_points` which it should not according to documentation.

This is the example of the docu:
```python
import numpy as np
import pyvista
pl = pyvista.Plotter()
points = np.array([[0, 1, 0], [1, 0, 0], [1, 1, 0], [2, 0, 0]])
actor = pl.add_lines(points, color='yellow', width=3)
pl.camera_position = 'xy'
pl.show()
```

### System Information

```shell
  Date: Thu Dec 01 11:12:36 2022 CET

                OS : Linux
            CPU(s) : 40
           Machine : x86_64
      Architecture : 64bit
               RAM : 62.8 GiB
       Environment : IPython
       File system : ext4
       GPU Details : error

  Python 3.8.10 (default, Jun 22 2022, 20:18:18)  [GCC 9.4.0]

           pyvista : 0.37.0
               vtk : 9.2.2
             numpy : 1.23.5
           imageio : 2.22.4
            scooby : 0.7.0
             pooch : v1.6.0
        matplotlib : 3.6.2
           IPython : 7.13.0
             scipy : 1.9.3
              tqdm : 4.64.1
```

### Screenshots

_No response_

tkoyama010:
### Overview

Fix the number error of the segment in `add_lines`.

 

resolves #3667 . Thanks, @tedalias for reporting this bug.

### Details

- None

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -4498,7 +4498,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         if not isinstance(lines, np.ndarray):
             raise TypeError('Input should be an array of point segments')
 
-        lines = pyvista.lines_from_points(lines)
+        lines = pyvista.line_segments_from_points(lines)
 
         actor = Actor(mapper=DataSetMapper(lines))
         actor.prop.line_width = width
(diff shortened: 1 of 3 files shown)


adeak:
Retroactively tagging this as [breaking-change], see https://github.com/pyvista/pyvista/issues/4065.

PR #3688, opened 2022-12-06
Fix the docstring of camera.py

tkoyama010:
### Overview

Fix the docstring of camera.py

 

### Details

- None

### DIFF
--- a/pyvista/plotting/camera.py
+++ b/pyvista/plotting/camera.py
@@ -14,7 +14,7 @@ class Camera(_vtk.vtkCamera):
 
     Parameters
     ----------
-    renderer : pyvista.Renderer, optional
+    renderer : pyvista.Renderer, default: None
         Renderer to attach the camera to.
 
     Examples
@@ -469,9 +469,8 @@ class Camera(_vtk.vtkCamera):
 
         Parameters
         ----------
-        aspect : float, optional
-            The aspect of the viewport to compute the planes. Defaults
-            to 1.0.
+        aspect : float, default: 1.0
+            The aspect of the viewport to compute the planes.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


pyvista/plotting/camera.py
```diff
@@ -649,19 +648,19 @@ def tight(self, padding=0.0, adjust_render_window=True, view='xy', negative=Fals
...
 
-        adjust_render_window : bool, optional
+        adjust_render_window : bool, default: True
             Adjust the size of the render window as to match the dimensions of
             the visible actors.
 
-        view : {'xy', 'yx', 'xz', 'zx', 'yz', 'zy'}
-            Plane to which the view is oriented. Default 'xy'.
+        view : {'xy', 'yx', 'xz', 'zx', 'yz', 'zy'}, default: 'xy'
+            Plane to which the view is oriented.
```
MatthewFlamm, line 660: numpydoc only suggests putting the default first in the set.  However I don't like this, your suggestion seems okay to me, but interested in others take.

pyvista/plotting/camera.py
```diff
@@ -14,7 +14,7 @@ class Camera(_vtk.vtkCamera):
 
     Parameters
     ----------
-    renderer : pyvista.Renderer, optional
+    renderer : pyvista.Renderer, default: None
```
MatthewFlamm, line 17: This should stay optional since the default is `None`.  What happens when it is `None`?  The current docstring isn't very helpful here.

banesullivan:
```suggestion
    renderer : pyvista.Renderer, optional
```

pyvista/plotting/camera.py
banesullivan: Yeah I think this makes sense

banesullivan (approved):

PR #2881, opened 2022-06-27
Update vtkThreshold API usage

banesullivan:
Resolves #2850 by updating our usage of the `vtkThreshold` filter

PR #732, opened 2020-05-23
VTK v9 Compatibility Fixes

akaszynski:
### VTK v9 Compatibility Fixes

This PR introduces several changes to account for the variations between VTK 8 and VTK 9:
- [x] The `offset` array is no longer required for `vtkUnstructuredGrid`
- [x] Bounds for `MulitBlock` no longer returns another `MultiBlock`
- [x] Offset array is different in VTK9 (decremented by 1 per index)
- [x] Do not trigger mouse clicks until render window is actually active.
- [x] Lines must be raveled for version consistency.
- [x] Set `vtk<=9.0` for Windows Python 3.5
- [x] Needs #730 

It should be noted that the `off_screen` plots look a bit better.  I think they're using anti-aliasing behind the scenes:
image

As compared to VTK 8.1.2:
image

This does seem to significantly increase the render time and our doc-build task will take much much longer and necessitates #720.

PR #3704, opened 2022-12-09
Add CI workflow to test VTK master branch

banesullivan:
This uses VTK's GitLab package index to download and install a wheel for the latest VTK (master branch). To install, run:

```
pip install vtk --pre --no-cache --extra-index-url https://wheels.vtk.org
```

Specifically, this branch uses the OSMesa wheel to prevent the need to xvfb

This runs as a cron job on `main`, has a `workflow_dispatch` trigger to run on request, and has a branch naming pattern for `'maint/vtk-master*'` to be able to open branches/PRs that will verify things are working against VTK master.


banesullivan:
vtkExplicitStructuredGrid DeepCopy is broken on master. Tracking here: https://gitlab.kitware.com/vtk/vtk/-/issues/18758

banesullivan:
## Threshold API changes

```
AttributeError: 'vtkmodules.vtkFiltersCore.vtkThreshold' object has no attribute 'ThresholdBetween'
```

There is a MR to re-add this method: https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9709

There is also a discussion about these changes here: https://discourse.vtk.org/t/unnecessary-vtk-api-change/9929

## Other errors

Two errors we may need to look into on our side:

```
FAILED tests/plotting/test_plotting.py::test_left_button_down - Failed: DID NOT RAISE <class 'ValueError'>
FAILED tests/plotting/test_plotting.py::test_plot_composite_poly_complex - Failed: DID NOT WARN. No warnings of type (<class 'numpy.ComplexWarning'>,) were emitted.
```

banesullivan:
@akaszynski, do you recall why you added this check in #732:

https://github.com/pyvista/pyvista/blob/4bc2064e68b08e9f0374c3f7935d608f2db80d8a/pyvista/plotting/plotting.py#L1774-L1777

This is no longer failing for VTK master branch and thus the `left_button_down` test will be updated here.

pyvista/core/filters/data_set.py
```diff
@@ -1100,7 +1100,10 @@ def threshold(
                 raise ValueError(
                     f'Value range must be length one for a float value or two for min/max; not ({value}).'
                 )
-            alg.ThresholdBetween(value[0], value[1])
+            # alg.ThresholdBetween(value[0], value[1])
+            alg.SetLowerThreshold(value[0])
+            alg.SetUpperThreshold(value[1])
+            alg.SetThresholdFunction(_vtk.vtkThreshold.THRESHOLD_BETWEEN)
```
banesullivan, line 1106: May want to put this in a separate PR to land it sooner

adeak: Does this mean https://github.com/pyvista/pyvista/pull/2881 was an insufficient fix? Or did VTK change further? I can't figure it out, sorry.

banesullivan: I think VTK changed further. This is a confusing one for sure

banesullivan: This was handled in #3750

akaszynski:
> @akaszynski, do you recall why you added this check in #732:
> 
> https://github.com/pyvista/pyvista/blob/4bc2064e68b08e9f0374c3f7935d608f2db80d8a/pyvista/plotting/plotting.py#L1774-L1777
> 
> This is no longer failing for VTK master branch and thus the `left_button_down` test will be updated here.

It's been a while, but I do recall genuine segfaults and having to work around them. I agree with your workaround.

.github/workflows/vtk-pre-test.yml
```diff
@@ -0,0 +1,74 @@
+name: VTK Master Testing
+
+on:
+  # TODO: remove
+  push:
+    branches:
+      - maint/vtk-master-testing
```
banesullivan, line 7:
```suggestion
```

.github/workflows/vtk-pre-test.yml
```diff
@@ -0,0 +1,74 @@
...
+  SHELLOPTS: 'errexit:pipefail'
+
+jobs:
+  Linux:
+    name: Linux VTK Master Testing
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.7, 3.8, 3.9, "3.10"]
```
banesullivan, line 27:
```suggestion
        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
```

.github/workflows/vtk-pre-test.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Set up vtk
+        run: |
+          pip uninstall vtk --yes
+          pip install vtk --pre --no-cache --extra-index-url https://gitlab.kitware.com/api/v4/projects/13/packages/pypi/simple
+
+      - uses: awalsh128/cache-apt-pkgs-action@v1.1.3
+        with:
+          packages: libgl1-mesa-glx xvfb python-tk
+          version: 3.0
```
banesullivan, line 61: Let's use the new OSMesa wheel variants when they get published later this/next week and remove libgl1-mesa-glx

.github/workflows/vtk-pre-test.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Install Testing Requirements
+        run: pip install -r requirements_test.txt
+
+      - name: Software Report
+        run: |
+          xvfb-run python -c "import pyvista; print(pyvista.Report()); from pyvista import examples; print('User data path:', examples.USER_DATA_PATH)"
```
banesullivan, line 68: Note comment above and remove `xvfb-run`

.github/workflows/vtk-pre-test.yml
akaszynski: While we're here, why do we install `python-tk`?

.github/workflows/vtk-pre-test.yml
akaszynski: If we no longer need `xvfb` (and probably not `python-tk`) we can remove all OS packages.

.github/workflows/vtk-pre-test.yml
banesullivan: No idea... had the same thought. This could be an artifice from qt?

akaszynski: Potentially.

Easier to ask forgiveness. Let's remove it and see.

banesullivan: I'll do this in a separate PR that cleans up the other CI jobs and leverages the OSMesa wheel for the 9.2.5 release

(Also upgrading jobs like docs to 3.11 where possible)

banesullivan:
Okay! CI is finally running with the `vtk_osmesa` package and tests are not segfaulting!

Unfortunately we have tests failures with the main branch of VTK

banesullivan:
@pyvista/developers, this is ready for review. IMO, we should merge this without addressing the testing failures, as the PR is to set up a CI routine that will make it easier to debug PyVista+VTK master issues.

After merging, we should follow up with individual PRs that address the testing failures here so that we can have targetted development for specific issues.

.github/workflows/vtk-pre-test.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Install pyvista
+        run: pip install . --no-deps
+
+      - name: Install Testing Requirements
+        run: pip install -r requirements_test.txt
+
+      - name: Set up VTK with OSMesa
+        run: |
+          pip uninstall vtk -y
+          pip install vtk_osmesa --pre --no-cache --extra-index-url https://gitlab.kitware.com/api/v4/projects/13/packages/pypi/simple
```
banesullivan, line 56: @akaszynski, @MatthewFlamm, @AlejandroFernandezLuces: this is the issue I am referring to in https://github.com/pyvista/pytest-pyvista/pull/41

MatthewFlamm: I think this is a separate issue from the issue you linked, but the impact is sometimes encountered in installing that package.  We have a dependency for `vtk` in pyvista, but apparently `vtk_osmesa` can be used without problem.  But, this means that `pyvista` still looks for `vtk`.  The problem is in the dependencies for `pyvista` itself, not in `pytest-pyvista` IMO.  I dont know if there is a solution for an 'either/or' dependency resolution in pip specification.

banesullivan: Indeed the issue is wit `pyvista` having `vtk` in its hard requirements. This is something I think we will have a lot more friction changing. And so a simpler fix is to just remove `pyvista` from `pytest-pyvista`'s hard requirements to work around this.

banesullivan:
@pyvista/developers, I recommend merging this PR as is to help track testing failures against VTK `master`. The workflow here is manually triggered, so will not cause failures on PRs.

.github/workflows/vtk-pre-test.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Install pyvista
+        run: pip install . --no-deps
+
+      - name: Install Testing Requirements
+        run: pip install -r requirements_test.txt
+
+      - name: Set up VTK with OSMesa
+        run: |
+          pip uninstall vtk -y
+          pip install vtk_osmesa --pre --no-cache --extra-index-url https://wheels.vtk.org
```
larsoner, line 56: It would be good to have one run that uses VTK-osmesa and another that uses Xvfb-and-standard-VTK. The failure modes could be different and the latter is a (much more?) widely used test case.

larsoner (commented):
> The workflow here is manually triggered, so will not cause failures on PRs.

Works for me as long as someone sets up to be notified on failures and actually fixes them. In MNE we have had a run for a while now where we install all sorts of stuff either from `main`/`master` including PyVista(Qt) and other stuff from daily/weekly CDNs (NumPy, SciPy, pandas, as of a couple of days ago VTK, etc.), but we *do* run it on all PRs rather than on a schedule. It means that once a month or so some CI breaks in someone's PR for an unrelated reason, but these are easy to ignore during review ("merge anyway ignoring requirements", or just don't add to "Required" list!) but also annoying enough that we fix them. It also catches stuff like using a deprecated APIs which would be missed until after merge by a scheduled run. You might consider trying this approach and seeing how onerous it is, and resorting to a scheduled one only if necessary.

But a scheduled run is better than no run at all so +1 from me either way! Hopefully a follow-up PR soon will make this green 🙂

akaszynski:
> Works for me as long as someone sets up to be notified on failures and actually fixes them

I'm of the opinion that it's better to be notified of failures early on and have this workflow (at least for one version of python) be always on.

However, I'm reminded of [alert fatigue](https://www.atlassian.com/incident-management/on-call/alert-fatigue#:~:text=Alert%20fatigue%E2%80%94also%20known%20as,the%20sheer%20number%20of%20alerts.) If this fails every so often, it's acceptable. If it's all the time and we can't correct it, we should have the option of disabling it.


Since it's failing right now, let's keep this manual for now and consider enabling it by default later.

akaszynski (approved):

.github/workflows/vtk-pre-test.yml
banesullivan: I think I will put that off for another PR -- due to current time constraints and because I want to use the new OSMesa wheel throughout our testing suite.

banesullivan:
> ... You might consider trying this approach and seeing how onerous it is, and resorting to a scheduled one only if necessary.

Thanks for the excellent example. I'd prefer not to have flaky CI jobs on PRs -- it signals that some things are okay to fail and then we as a community are quick to ignore that failure and soon other failures because we're used to merging with a red X. "alert fatigue" as @akaszynski pointed out. I've seen this be a big problem on some projects that now I'm very averse to it.

> Since it's failing right now, let's keep this manual for now and consider enabling it by default later.

Indeed - this is failing right now, and I think we need to raise an issue or two upstream in VTK before these things can be addressed. I suspect many failures moving forward are going to be things are broken upstream and thus we don't have a lot of control here in PyVista -- a further reason I'd prefer this to be a manual job we intentionally run to report issues upstream at once.

One such (serious) issue was discovered in this PR (see https://gitlab.kitware.com/vtk/vtk/-/issues/18758). I need to dig into the logs here and report any current failures

PR #3643, opened 2022-11-23
Feat/add connectivity methods

beroda:
### Overview
This should resolve https://github.com/pyvista/pyvista/issues/3598

PR #3715, opened 2022-12-14
Wrapping vtkCell

beroda:
### Overview

This PR implements a wrapping of **vtkCell**, being inspired from @banesullivan and @MatthewFlamm from [this discussion](https://github.com/pyvista/pyvista/discussions/2838) and [this PR](https://github.com/pyvista/pyvista/pull/3643).

### Details
 
It implements a **cell** attribute to **DataSet** that returns a list of a new **Cell** class.
Consequently, you can access (read-only) the following cell-related information:

```python
>>> import pyvista
>>> mesh = pyvista.Sphere()
>>> mesh.cell[0]
Cell (0x7fd0c064efa0)
  Type:	CellType.TRIANGLE
  Linear:	True
  Dimension:	2
  N Points:	3
  N Faces:	0
  N Edges:	3
  X Bounds:	-1.075e-01, -5.406e-02
  Y Bounds:	-2.235e-02, 0.000e+00
  Z Bounds:	4.883e-01, 4.971e-01
```
Get if the 0-th cell is linear
```python
>>> mesh.cell[0].is_linear
True
```
Get the number of points, faces and edges of the 0-th cell
```python
>>> mesh.cell[0].n_points
3
>>> mesh.cell[0].n_faces
0
>>> mesh.cell[0].n_edges
3
```
Get the point ids of the 0-th cell
```python
>>> mesh.cell[0].point_ids
[840, 29, 28]
```
Get the 0-th edge of the 0-th cell
```python
>>> mesh.cell[0].edges[0]
Cell (0x7f082560dbe0)
  Type:	CellType.LINE
  Linear:	True
  Dimension:	1
  N Points:	2
  N Faces:	0
  N Edges:	0
  X Bounds:	-1.075e-01, -1.051e-01
  Y Bounds:	-2.235e-02, 0.000e+00
  Z Bounds:	4.883e-01, 4.883e-01
```
Get the 0-th face from the 0-th cell of an unstructured grid
```python
>>> from pyvista.examples import load_hexbeam
>>> mesh = load_hexbeam()
>>> mesh.cell[0].faces[0]
Cell (0x7fd0c0247790)
  Type:	CellType.QUAD
  Linear:	True
  Dimension:	2
  N Points:	4
  N Faces:	0
  N Edges:	4
  X Bounds:	5.000e-01, 1.000e+00
  Y Bounds:	5.000e-01, 1.000e+00
  Z Bounds:	5.000e+00, 5.000e+00
```


beroda:
I think this PR might deprecate some **DataSet** methods such as:

- cell_n_points
- cell_bounds
- cell_type
- cell_point_ids

I decided not to put depreciation warnings and notes in the method's docstrings since I think it deserves discussion with maintainers first.

MatthewFlamm:
This is great and look forward to review it!  I am in favor of deprecating.  In a small way, it will help reduce the clutter of methods/filters that are accumulating.

I think this also warrants a mention, or a short section in https://github.com/pyvista/pyvista/blob/main/doc/user-guide/data_model.rst once you feel that this is close enough.

banesullivan:
+1 to @MatthewFlamm's comment! This is looking nice so far. I'm looking forward to reviewing it

pyvista/core/cell.py
```diff
@@ -1,9 +1,213 @@
...
+                fmt += row.format(attr[0], attr[2].format(*attr[1]))
+            except:
+                fmt += row.format(attr[0], attr[2].format(attr[1]))
+        return fmt
+
+    def __repr__(self):
+        """Return the object representation."""
+        return self.str(html=False)
+
+    def __str__(self):
```
MatthewFlamm, line 206: I think we don't need this as it will default to using `__repr__`, which is currently the same.

tests/test_cells.py
```diff
@@ -4,6 +4,134 @@
...
+        assert isinstance(grid.cell, list)
+        assert len(grid.cell) == grid.n_cells
+        assert all(t == Cell for t in map(type, grid.cell[:]))
+
+    @pytest.mark.parametrize("grid", grids, ids=ids)
+    def test_cell_type_is_inside_enum(self, grid):
+        assert grid.cell[0].type in CellType
+
+    types = [
+        CellType.HEXAHEDRON,
```
MatthewFlamm, line 49: nitpick suggestion: I think it would be better to collocate these variables with `grids` since I had to scroll up and down to compare.

MatthewFlamm (commented):
We also try to have Examples for each property in a class.  I think you could move sections of the `Cell` Examples into the properties and reviewers may be able to help finish that too.

pyvista/core/cell.py
```diff
@@ -1,9 +1,213 @@
...
+        """Init the cell object with a _vtk.vtkCell."""
+        if not isinstance(vtkobject, _vtk.vtkCell):
+            msg = f"`vtkobject` must be of vtkCell type (got {type(vtkobject)}) instead"
+            raise TypeError(msg)
+        super().__init__(vtkobject=vtkobject)
+
+    @property
+    def type(self) -> CellType:
+        """Get the cell type from the enum `CellType`."""
+        for t in CellType:
```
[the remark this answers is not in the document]
beroda: Yes I agree. I changed it to use the __call__ method of IntEnum, maybe it is faster than a loop.

pyvista/core/cell.py
beroda: I agree

tests/test_cells.py
beroda: Yes, I changed it

pyvista/core/cell.py
```diff
@@ -1,9 +1,213 @@
...
+        attrs.append(("N Edges", self.n_edges, "{}"))
+        bds = self.bounds
+        fmt = f"{pyvista.FLOAT_FORMAT}, {pyvista.FLOAT_FORMAT}"
+        attrs.append(("X Bounds", (bds[0], bds[1]), fmt))
+        attrs.append(("Y Bounds", (bds[2], bds[3]), fmt))
+        attrs.append(("Z Bounds", (bds[4], bds[5]), fmt))
+
+        return attrs
+
+    def str(self, html=False):
```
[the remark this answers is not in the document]
beroda: You're right, I will make it private. I inteded to implement the `_repr_html_` method and call this `_str(html=True)` inside of it.

beroda:
> We also try to have Examples for each property in a class. I think you could move sections of the `Cell` Examples into the properties and reviewers may be able to help finish that too.

You are completely right, the docstring of `Cell` was getting very long, so I moved a significant portion inside the properties.

pyvista/core/cell.py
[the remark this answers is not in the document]
MatthewFlamm: This is much better and it seems like the pythonic method according to the docs.

For a ~14000 cell mesh, using the `__call__` method is 10ms on my test machine to get all the types.  Returning the ints directly from vtk is 5ms.  For a 1 million cell mesh, it was 695ms vs 350 ms.  So twice as slow, but probably not importantly so.  Given that we created this Enum it makes sense to me to use it even if there is a slight performance hit.  I will leave this comment unresolved for other reviewer comments.

beroda:
@banesullivan @MatthewFlamm I encountered a very weird behavior in my implementation, see this example:

```python
from pyvista.examples import load_hexbeam

mesh = load_hexbeam()
print(mesh.get_cell(0).point_ids)
print(mesh.cell[0].point_ids)
```
[0, 2, 8, 7, 27, 36, 90, 81]
[98, 62, 53, 80, 17, 13, 12, 15]

It seems that successive calls to `get_cell` only retain the last one, see the following:
```python
print(mesh.get_cell(1).point_ids)
cells = [ mesh.get_cell(i) for i in range(2) ]
print(cells[0].point_ids)
```
[2, 1, 4, 8, 36, 18, 54, 90]
[2, 1, 4, 8, 36, 18, 54, 90]

I guess this might have to do with the `VTKObjectWrapper` inheritance, but unfortunately, I think I reached the limits of my knowledge of the vtk-python wrapping. Might need some crucial help here.

MatthewFlamm:
You are using the wrong overload for GetCell, see https://vtk.org/doc/nightly/html/classvtkDataSet.html#a711ed1ebb7bdf4a4e2ed6896081cd1b2 especially the warning:

> Repeat calls to this function for different face ids will change the data stored in the internal member object whose pointer is returned by this function.

We need to use https://vtk.org/doc/nightly/html/classvtkDataSet.html#a6fa0ace8aa3a2de7c16a674f0499bad9

Since vtkGenericCell is a concrete implementation, we might be able to get rid of VTKObjectWrapper.

pyvista/core/cell.py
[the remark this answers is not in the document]
beroda: Thank you for your extensive testing

beroda:
> We need to use https://vtk.org/doc/nightly/html/classvtkDataSet.html#a6fa0ace8aa3a2de7c16a674f0499bad9

You were right, I was able to make it work using **vtkGenericCell**. I also implemented a test that detected this bug.

> Since vtkGenericCell is a concrete implementation, we might be able to get rid of VTKObjectWrapper.

I was not able to get rid of VTKObjectWrapper. 
For some unknown reason, when I inherited from **vtkGenericCell** and deep-copied the input cell, some attributes (but not all) were erroneous (see the following MWE).

```python
from pyvista.examples.cells import Tetrahedron
from pyvista import _vtk

class Cell(_vtk.vtkGenericCell):

    def __init__(self,vtkobject):
        super().__init__()
        self.DeepCopy(vtkobject)

cell = _vtk.vtkGenericCell()
Tetrahedron().GetCell(0,cell)

pvcell = Cell(cell)
print(pvcell.GetCellType())
print(cell.GetCellType())
print(pvcell.GetBounds())
print(cell.GetBounds())
```
0
10
(-1.0, 1.0, -1.0, 1.0, -1.0, 1.0)
(-1.0, 1.0, -1.0, 1.0, -1.0, 1.0)

Something more interesting is that I found the same issue for the  **_edges_** and **_faces_** list attributes of the Cell class.
Unfortunately, I could not find any vtkCell's method which prevent this in vtk's documentation.

However, I managed to overcome this issue by returning a generator instead of a list. It is a mystery to me why this works. See this MWE:
```python
from pyvista.examples.cells import Tetrahedron
from pyvista import _vtk

edges = Tetrahedron().cell[0].edges
for e in edges:
    print(e.point_ids)

print()
edges = list(Tetrahedron().cell[0].edges)
for e in edges:
    print(e.point_ids)
```
[0, 1]
[1, 2]
[2, 0]
[0, 3]
[1, 3]
[2, 3]

[2, 3]
[2, 3]
[2, 3]
[2, 3]
[2, 3]
[2, 3]

If it is fine for you, I will add a warning in the documentation to tell users that the edges and faces generator must never be converted to a list. I also added some tests similar to the one mentioned above to detect this issue.

MatthewFlamm:
> Something more interesting is that I found the same issue for the edges and faces list attributes of the Cell class.

`GetFace` has the same warning, and although `GetEdge` does not, you have confirmed it should. There is a `GetFaces`, but no `GetEdges`.  So we might be able to use for faces but not edges.

However, I like the idea of having a generator, with suitable warning, for all of these as it allows easier looping and keeps a consistent API.  But now I'm wondering if we should also do the same for the cells in the first place, i.e. also make it a generator.

beroda:
If all the vtk-methods were thread-safe, I would have preferred the list-like outputs because it allows both for loops and indexing access. Moreover, it is my personal opinion that generators are relevant when memory usage matters, which does not appear to be the case here.
But it is rather a personal taste and I am opened to discussion with other reviewers :)

MatthewFlamm:
> Moreover, it is my personal opinion that generators are relevant when memory usage matters, which does not appear to be the case here.

Good point.  In my testing the isn't a huge memory hit to generating the list of cells, but it can be very slow for large meshes.  Maybe if we keep the list behavior, we can put in a warning too to prefer the `get_cell` method.

beroda:
You are right. I will add another warning to the `cell` attribute (I already added the warning regarding the edges and faces generators)

MatthewFlamm:
There is one test that consistently fails here, but it doesn't seem related to this PR.  https://github.com/pyvista/pyvista/actions/runs/3761506807/jobs/6476389282#step:11:454

I'm not sure what is the issue, but it doesn't seem to be failing on other PRs.

beroda:
This failed test is very weird. We didn't touch to plotter and it fails only with conda ??
It appeared just after I used the `vtkGenericCell` class with generators for edges and faces.

I can try to investigate on an Ubuntu VM with conda but I am afraid it goes beyond my python's knowledge (but happy to learn new stuff).

adeak:
I'm on debian, I can try what happens with conda.

beroda:
> I'm on debian, I can try what happens with conda.

That would be great, thanks a lot @adeak

adeak:
I can't reproduce the issue with a debian desktop. I even tried following the steps of the conda action (without fluff like html reporting). The setup is not identical, as I get 321 tests during collection whereas in CI there's only 319, but I don't know where the difference is. The most likely reason is the headless display action being used in CI, whereas I have a proper desktop. In any case it seems weirder and weirder that this test is failing in CI.

MatthewFlamm:
It is probably a red herring since all the other CI tests pass and it can't be reproduced locally, but I think `PointSet` is the only `DataSet` without cells.

pyvista/core/dataset.py
```diff
@@ -2479,9 +2480,122 @@ def find_cells_within_bounds(self, bounds: Iterable[float]) -> np.ndarray:
...
+        list[pyvista.Cell]
+            A list of `pyvista.Cell` objects.
+
+        Warnings
+        --------
+        The ``edges`` and ``faces`` properties of the :class:`pyvista.Cell`
+        object are generators that should NEVER be converted to a list
+        since repeated calls to the vtk underlying methods can change the
+        returned data.
+        See https://vtk.org/doc/nightly/html/classvtkCell.html#a11b6ba66e9f7e193b204d478379e32ea.
```
adeak, line 2533: This warning is not entirely clear to me, and at the same time it's kind of terrifying. Consuming a generator into a list is a very natural thing to do. If this is forbidden, do we really want to keep this feature as-is?

What exactly is the risk? What "underlying methods" do get called and what "returned data" change, exactly? I looked at the link and it wasn't clear to me at all, so I don't think an end user will be better off reading this.

Looking at e.g. the `Cell.faces` property: we seem to just be providing a thin wrapper for `get_face()`. So if the way we do this is fragile, can we just not provide the wrapper, or provide it in a way that is safe?

And this warning should be present in a lot more places (everywhere where a `Cell` might be created, I think). This might overlap with what @MatthewFlamm's already said.

MatthewFlamm: It is a vtk limitation.  vtk returns the same object everytime you call get_face() but with new data inside.  So if you make a list, you get a list of length `n_cells` filled with just the last element.  One workaround is to always copy the cell, but this has memory issues.

MatthewFlamm: Although, I could be wrong on this, faces and edges are not typically part of a dataset, and it's cells, and would be generated on-the-fly.  So the memory hit would not be so large and would only become a problem if the user created a nested list of all faces for _all_ cells.

adeak: Thanks, this was my best guess for the situation. Just to make sure we mean the same thing: something like
```py
old_face = cell.get_face(0)
new_face = cell.get_face(1)
new_face is old_face  # oh no
```

But then this means that our generator always returns the same object. And if we did `set(cell.faces)`, we'd end up with a 1-length cell (assuming the object is hashable).

Then my recommendation is not to provide this helper at all. Having a generator that yields the same object each time, given this caveat, is very misleading. Only provide `get_face()` with a huge red warning not to work on two faces at the same time. Or copy, but as you said this is not viable.

adeak: > Although, I could be wrong on this, faces and edges are not typically part of a dataset, and it's cells, and would be generated on-the-fly. 

I was wondering about this. Cells don't have that many faces [citation needed]. So we could return a copy of each. If a user chooses to hang on to a bunch of cell faces that's not our responsibility (assuming there's no _surprising_ amount of memory overhead, from, I don't know, copying cell faces leading to copying the whole dataset or something).

MatthewFlamm: > something like
>```
>old_face = cell.get_face(0)
>new_face = cell.get_face(1)
>new_face is old_face  # oh no
>```

Correct.

> But then this means that our generator always returns the same object. 

Not quite,

```
for face in cell.faces:
    operate_on_face(face)
```

would operate on unique faces.

But I'm tending to agree with you that if we provide the generator option, it will be more natural for users to assume they can make a list.  I'm okay either with explicit clear warnings or removing it.

beroda: Agreeing with the concerns you raised both @adeak and @MatthewFlamm.
I am also fine with copying, but I am afraid it goes beyond my vtk knowledge. What do you mean exactly ? Using the `DeepCopy()` method of the `vtkCell` ?
Might be unrelated, but I encountered a weird behaviour when using `DeepCopy()`, see https://github.com/pyvista/pyvista/pull/3715#issuecomment-1362178873.

beroda: Moreover, the pointset test that consistently failed magically passed when merging with last main ? Weird ...

beroda: > Thanks, this was my best guess for the situation. Just to make sure we mean the same thing: something like
> 
> ```python
> old_face = cell.get_face(0)
> new_face = cell.get_face(1)
> new_face is old_face  # oh no
> ```
> 
> right?

 From this MWE, it seems that the `Cell` faces objects are different:
```python
from pyvista.examples.cells import Tetrahedron

c = Tetrahedron().cell[0]
f1 = c.get_face(0)
f2 = c.get_face(1)
f1 is f2
```
False

adeak: That looks related, thanks. I'd expect your code to work the way you expect, `DeepCopy` should, you know, do a deep copy. The fact that it doesn't is somewhat concerning, but I'm no VTK expert either.

To see how different the original and the copy are, you can compare `print(cell)` and `print(pvcell)`, there are quite a few attributes that are not carried over.

beroda: I was also surprised it did not work as expected.

I'm getting slowly convinced that the our best option would be to remove the `faces` and `edges` generators and only provide the `get_face` and `get_edge` methods to end-users, without any terrifying warnings (for which I apologize for :angel:).

akaszynski:
@beroda, taking a stab at this for a bit to resolve the `GetFace` generator issue.

akaszynski:
> One workaround is to always copy the ~cell~ face

IMO this is the only way. Reason being, users could write their own generator and we've effectively just exposed a potentially dangerous method `get_face` or `get_edge` that could be easily abused. Even if we place a well meaning warning (as in [vtkCell.GetFace](https://vtk.org/doc/nightly/html/classvtkCell.html#a11b6ba66e9f7e193b204d478379e32ea), it's quite easy to run into issues. With my testing, even serial sequential calls to `get_face` seems to corrupt the underlying memory pointer:

```py
>>> import time
>>> from pyvista.examples.cells import Tetrahedron
>>> mesh = Tetrahedron()
>>> cell = mesh.cell[0]
>>> face_a, face_b = cell.get_face(0), cell.get_face(1)
>>> face_a == face_b
True

>>> face_a = cell.get_face(0)
>>> time.sleep(1)
>>> face_b = cell.get_face(1)
>>> face_a == face_b
True
```

To quote Pink Floyd, "This will not do."

>  So the memory hit would not be so large and would only become a problem if the user created a nested list of all faces for all cells.

This new implementation is designed to give the user a new approach for interacting with a `DataSet`, which is both slower and less memory efficient than directly working with the cell arrays as in [cells](https://docs.pyvista.org/api/core/_autosummary/pyvista.UnstructuredGrid.cells.html) attribute. For those who want a quick and dirty solution, I think `Cells` is the way to go, and for those who need speed at the expense of development time/learning curve, accessing the underlying data as in [cells](https://docs.pyvista.org/api/core/_autosummary/pyvista.UnstructuredGrid.cells.html) is preferable.

PyVista has always been about usability, but from looking at the C++ implementation, I'd even be worried about using `GetFaces` in C++ given the bright red warnings. Let's protect the user from going down the rabbit hole we just did.

akaszynski:
This ended up being longer and more in depth than expected...

First, part of the reason for this:

> Something more interesting is that I found the same issue for the edges and faces list attributes of the Cell class.
Unfortunately, I could not find any vtkCell's method which prevent this in vtk's documentation.

Was due to the necessity to set the cell type of a generic cell with [SetCellType](https://vtk.org/doc/nightly/html/classvtkGenericCell.html#ab9d012c98f3d35fe6791f2756eb860d6). With this, I was able to get rid of `VTKObjectWrapper` initially quite quickly with a new `pyvista.Cell` class that subclassed `vtkCell`.

Things started to get challenging when getting the faces and edges from individual cells, to the point where I mistakenly thought it was necessary to subclass every possible vtkCell type. For the sake of maintainability (and @adeak), I went with simply treating every cell as a `vtkGenericCell` (as originally implemented), with a little bit of syntactic sugar.

doc/api/core/cells.rst
```diff
@@ -0,0 +1,58 @@
+Cells
+-----
+
+The cell :class:`pyvista.Cell` class is the PyVista representation of the
+`vtkGenericCell <https://vtk.org/doc/nightly/html/classvtkGenericCell.html>`_ and can be used to inspect a dataset's cells, faces, and edges.
+
+.. note::
+   While methods and classes are quite effective at inspecting and plotting
+   parts of a dataset, they are quite inefficient and should be used only for
```
akaszynski, line 9:
```suggestion
   parts of a dataset, they are inefficient and should be used only for
```

doc/api/core/cells.rst
```diff
@@ -0,0 +1,58 @@
+Cells
+-----
+
+The cell :class:`pyvista.Cell` class is the PyVista representation of the
+`vtkGenericCell <https://vtk.org/doc/nightly/html/classvtkGenericCell.html>`_ and can be used to inspect a dataset's cells, faces, and edges.
```
akaszynski, line 5:
```suggestion
`vtkGenericCell <https://vtk.org/doc/nightly/html/classvtkGenericCell.html>`_ and can be used to inspect a :class:`pyvista.DataSet`'s cells, faces, and edges.
```

doc/api/core/cells.rst
```diff
@@ -0,0 +1,58 @@
...
+
+.. note::
+   While methods and classes are quite effective at inspecting and plotting
+   parts of a dataset, they are quite inefficient and should be used only for
+   interactive exploration and debugging. When working with larger datasets or
+   working with multiple cells it is generally more efficient to use bulk methods
+   like :func:`pyvista.DataSet.extact_cells`.
```
akaszynski, line 12:
```suggestion
   like :func:`pyvista.DataSet.extact_cells`, :attr:`pyvista.pyvista.PolyData.faces`
   or :attr:`pyvista.UnstructuredGrid.cells`.
```

pyvista/core/cell.py
```diff
@@ -1,112 +1,435 @@
...
+    Notes
+    -----
+    Accessing individual cells from a :class:`pyvista.DataSet` using this class
+    will be much slower than accessing bulk data from the
+    :attr:`pyvista.pyvista.PolyData.faces` or
+    :attr:`pyvista.UnstructuredGrid.cells` attributes.
+
+    Also note that the the cell object is a deep copy of the original cell and
+    is unassociated with the original cell. Changing any data of that points of
+    that cell (for example, ``points``) will not change the original dataset.
```
akaszynski, line 42:
```suggestion
    is unassociated with the original cell. Changing any data of
    that cell (for example, :attr:`pyvista.Cell.points`) will not change the original dataset.
```

pyvista/core/cell.py
```diff
@@ -1,112 +1,435 @@
...
-        BEZIER_HEXAHEDRON = _vtk.VTK_BEZIER_HEXAHEDRON
-    if hasattr(_vtk, "VTK_BEZIER_WEDGE"):
-        BEZIER_WEDGE = _vtk.VTK_BEZIER_WEDGE
-    if hasattr(_vtk, "VTK_BEZIER_PYRAMID"):
-        BEZIER_PYRAMID = _vtk.VTK_BEZIER_PYRAMID
+    def __init__(self, vtk_cell=None, cell_type=None, deep=False):
+        """Initialize the cell."""
+        super().__init__()
+        # cell type must be set first before deep or shallow copy
+
```
akaszynski, line 86:
```suggestion
```

akaszynski, line 86:
```suggestion
```

pyvista/core/cell.py
```diff
@@ -1,112 +1,435 @@
...
+        if vtk_cell is not None:
+            if not isinstance(vtk_cell, _vtk.vtkCell):
+                raise TypeError(f'`vtk_cell` must be a vtkCell, not {type(vtk_cell)}')
+            if cell_type is None:
```
akaszynski, line 90:
```suggestion
            # cell type must be set first before deep or shallow copy
            if cell_type is None:
```

pyvista/core/cell.py
```diff
@@ -1,112 +1,435 @@
...
+
+    This class provides the capability to access a given cell topology and can
+    be useful when walking through a cell's individual faces or investigating
+    cell properties.
 
-class CellType(IntEnum):
-    """Define types of cells.
+    Parameters
+    ----------
+    vtk_cell : vtk.vtkCell
```
akaszynski, line 24:
```suggestion
    vtk_cell : vtk.vtkCell, optional
```

pyvista/core/cell.py
```diff
@@ -1,112 +1,435 @@
...
+            args.insert(0, [0])
+
+        return pyvista.UnstructuredGrid(*args)
+
+    @property
+    def dimension(self) -> int:
+        """Return the cell dimension.
+
+        This returns the dimensionality of the cell. For example, 1 for an edge,
+        2 for a triangle and 3 for a tetrahedron.
```
akaszynski, line 186:
```suggestion
        2 for a triangle, and 3 for a tetrahedron.
```

pyvista/core/cell.py
```diff
@@ -1,112 +1,435 @@
...
+        >>> import pyvista
+        >>> mesh = pyvista.Sphere()
+        >>> mesh.cell[0].n_edges
+        3
+        """
+        return self.GetNumberOfEdges()
+
+    @property
+    def point_ids(self) -> List[int]:
+        """Get the point ids composing the cell.
```
akaszynski, line 238:
```suggestion
        """Get the point IDs composing the cell.
```

pyvista/core/cell.py
```diff
@@ -1,112 +1,435 @@
...
+        --------
+        >>> import pyvista
+        >>> mesh = pyvista.Sphere()
+        >>> mesh.cell[0].points
+        array([[-5.40595092e-02,  0.00000000e+00, -4.97068971e-01],
+               [-5.28781787e-02,  1.12396041e-02, -4.97068971e-01],
+               [-5.55111512e-17,  0.00000000e+00, -5.00000000e-01]])
+        """
+        # A copy of the points must be returned to avoid overlapping them since the
+        # `vtk.vtkExplicitStructuredGrid.GetCell` is an override method.
```
akaszynski, line 264:
```suggestion
```

tests/test_grid.py
```diff
@@ -281,7 +281,7 @@ def test_init_polyhedron():
     grid = pyvista.UnstructuredGrid(cells, cell_type, nodes)
 
     assert grid.n_cells == len(cell_type)
-    assert grid.cell_type(0) == pyvista.CellType.POLYHEDRON
+    assert grid.cell[0].type == pyvista.CellType.POLYHEDRON
```
akaszynski, line 284:
```suggestion
    assert grid.get_cell(0).type == pyvista.CellType.POLYHEDRON
```

pyvista/core/cell.py
```diff
@@ -1,112 +1,435 @@
...
+            original cell.
+
+        Returns
+        -------
+        pyvista.Cell
+            Deep or shallow copy of the cell.
+
+        Examples
+        --------
+        Create a deep copy of the cell and demonstrate its deep.
```
akaszynski, line 417:
```suggestion
        Create a deep copy of the cell and demonstrate it is deep.
```

pyvista/core/dataset.py
```diff
@@ -2479,9 +2479,138 @@ def find_cells_within_bounds(self, bounds: Iterable[float]) -> np.ndarray:
...
+        [1, 2, 3]
+        [2, 0, 3]
+        [0, 2, 1]
+        """
+        return [self.get_cell(i) for i in range(self.n_cells)]
+
     def cell_n_points(self, ind: int) -> int:
         """Return the number of points in a cell.
 
+        ..  deprecated:: 0.38.0
```
akaszynski, line 2611:
```suggestion
        .. deprecated:: 0.38.0
```

pyvista/core/dataset.py
```diff
@@ -2479,9 +2479,138 @@ def find_cells_within_bounds(self, bounds: Iterable[float]) -> np.ndarray:
...
+            You can use :py:attr:`DataSet.cell[i].n_points` instead.
```
akaszynski, line 2612:
```suggestion
            Use :attr:`DataSet.Cell.n_points` instead.
```

pyvista/core/dataset.py
```diff
@@ -2500,11 +2629,19 @@ def cell_n_points(self, ind: int) -> int:
...
+            '`cell_n_points` is deprecated. Use `cell[i].n_points` instead',
+            PyVistaDeprecationWarning,
+        )
+        return self.cell[ind].n_points
 
     def cell_points(self, ind: int) -> np.ndarray:
         """Return the points in a cell.
 
+        ..  deprecated:: 0.38.0
+            You can use :py:attr:`DataSet.cell[i].points` instead.
```
akaszynski, line 2643:
```suggestion
        ..  deprecated:: 0.38.0
            Use :attr:`DataSet.Cell.points` instead.
```

pyvista/core/dataset.py
```diff
@@ -2526,15 +2663,18 @@ def cell_points(self, ind: int) -> np.ndarray:
...
+        warnings.warn(
+            '`cell_points` is deprecated. Use `cell[i].points` instead', PyVistaDeprecationWarning
+        )
+        return self.get_cell(ind).points
 
     def cell_bounds(self, ind: int) -> Tuple[float, float, float, float, float, float]:
         """Return the bounding box of a cell.
 
+        ..  deprecated:: 0.38.0
+            You can use :py:attr:`DataSet.cell[i].bounds` instead.
```
akaszynski, line 2676:
```suggestion
        ..  deprecated:: 0.38.0
            You can use :attr:`pyvista.Cell.bounds` instead.
```

pyvista/core/dataset.py
```diff
@@ -2553,11 +2693,18 @@ def cell_bounds(self, ind: int) -> Tuple[float, float, float, float, float, floa
         (896.9940185546875, 907.5390014648438, 48.760101318359375, 55.49020004272461, 80.74520111083984, 83.65809631347656)
 
         """
-        return self.GetCell(ind).GetBounds()
+        # deprecated 0.38.0, convert to error in 0.41.0, remove 0.42.0
+        warnings.warn(
+            '`cell_bounds` is deprecated. Use `cell[i].bounds` instead', PyVistaDeprecationWarning
+        )
+        return self.cell[ind].bounds
```
akaszynski, line 2700:
```suggestion
        return self.get_cell[ind].bounds
```

pyvista/core/dataset.py
```diff
@@ -2500,11 +2629,19 @@ def cell_n_points(self, ind: int) -> int:
         3
 
         """
-        return self.GetCell(ind).GetPoints().GetNumberOfPoints()
+        # deprecated 0.38.0, convert to error in 0.41.0, remove 0.42.0
+        warnings.warn(
+            '`cell_n_points` is deprecated. Use `cell[i].n_points` instead',
+            PyVistaDeprecationWarning,
+        )
+        return self.cell[ind].n_points
```
akaszynski, line 2637:
```suggestion
        return self.get_cell[ind].n_points
```

pyvista/core/dataset.py
```diff
@@ -2573,14 +2720,21 @@ def cell_type(self, ind: int) -> int:
...
-        5
+        <CellType.TRIANGLE: 5>
 
         """
-        return self.GetCellType(ind)
+        # deprecated 0.38.0, convert to error in 0.41.0, remove 0.42.0
+        warnings.warn(
+            '`cell_type` is deprecated. Use `cell[i].type` instead', PyVistaDeprecationWarning
+        )
+        return self.cell[ind].type
```
akaszynski, line 2730:
```suggestion
        return self.get_cell(ind).type
```

pyvista/core/dataset.py
```diff
@@ -2553,11 +2693,18 @@ def cell_bounds(self, ind: int) -> Tuple[float, float, float, float, float, floa
...
+        warnings.warn(
+            '`cell_bounds` is deprecated. Use `cell[i].bounds` instead', PyVistaDeprecationWarning
+        )
+        return self.cell[ind].bounds
 
     def cell_type(self, ind: int) -> int:
         """Return the type of a cell.
 
+        ..  deprecated:: 0.38.0
+            You can use :py:attr:`DataSet.cell[i].type` instead.
```
akaszynski, line 2706:
```suggestion
        .. deprecated:: 0.38.0
            You can use :attr:`DataSet.Cell.type` instead.
```

pyvista/core/dataset.py
```diff
@@ -2573,14 +2720,21 @@ def cell_type(self, ind: int) -> int:
...
+        warnings.warn(
+            '`cell_type` is deprecated. Use `cell[i].type` instead', PyVistaDeprecationWarning
+        )
+        return self.get_cell(ind).type
 
     def cell_point_ids(self, ind: int) -> List[int]:
         """Return the point ids in a cell.
 
+        ..  deprecated:: 0.38.0
+            You can use :py:attr:`DataSet.cell[i].point_ids` instead.
```
akaszynski, line 2736:
```suggestion
        .. deprecated:: 0.38.0
            You can use :attr:`DataSet.Cell.point_ids` instead.
```

akaszynski:
@banesullivan and @MatthewFlamm, this PR uses `addopts` in https://github.com/pyvista/pyvista/pull/3715/commits/79f5a690515f6eb49c5ba83d11d88ff98c57a203. Otherwise, local testing fails and/or gives me the warning:

```
PytestConfigWarning: Unknown config option: image_cache_dir
```

MatthewFlamm:
> @banesullivan and @MatthewFlamm, this PR uses `addopts` in [79f5a69](https://github.com/pyvista/pyvista/commit/79f5a690515f6eb49c5ba83d11d88ff98c57a203). Otherwise, local testing fails and/or gives me the warning:
> 
> ```
> PytestConfigWarning: Unknown config option: image_cache_dir
> ```

Does it only fail locally?  If so, you may have to do `pip install -r requirements_test.txt` again.

akaszynski:
> If so, you may have to do `pip install -r requirements_test.txt` again.

Quite right. Thought I had updated it and it turns out I had just updated the wrong venv.

Reverted in bc4c2b32feaf2fad06ba5bbeb8930cb6a7a27a6e.

akaszynski (dismissed):
LGTM. Thanks @beroda for your hard work and persistence.

beroda:
@akaszynski Thanks for your amazing work too !

akaszynski (approved):
There was one last unresolved question. Reapproving...

PR #3723, opened 2022-12-18
Fix two other closed geometric objects.

dcbr:
### Overview

As per my comment in #3710, there are two other closed geometric objects that require a fix in my opinion.

### Details

<details open>
<summary><b>Ellipse</b></summary>

```py
pyvista.Ellipse(resolution=4).plot()
```
| Before | After |
| ------- | ------ |
| image | image |

Duplicate endpoints and inconsistent edge lengths; similar fix as in #3710.

</details>

<details open>
<summary><b>CylinderStructured</b></summary>

```py
pyvista.CylinderStructured(theta_resolution=4).plot()
```
| Before | After |
| ------- | ------ |
| image | image |

Inconsistent number of edges/faces when compared to `Cylinder`; as this is a `StructuredGrid` duplicate endpoints are needed in this case I believe (correct me if I'm wrong), so the fix is different than for the `Ellipse` above.

</details>

### DIFF
--- a/pyvista/utilities/geometric_objects.py
+++ b/pyvista/utilities/geometric_objects.py
@@ -157,6 +157,12 @@ def CylinderStructured(
     pyvista.StructuredGrid
         Structured cylinder.
 
+    Notes
+    -----
+    .. versionchanged:: 0.38.0
+       Prior to version 0.38, this method had incorrect results, producing
+       inconsistent number of points on the circular face of the cylinder.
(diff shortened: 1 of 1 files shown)


banesullivan (approved):
Thank you for fixing these and adding the `versionchanged` note. I'm going to add the "breaking-change" label to the PR so this gets picked up in the release notes

PR #3731, opened 2022-12-19
Redraw ChartMPL on Plotter render

### LINKED ISSUE #3380: Plotter should redraw chartMPL instances if axes is modified with set_xxxx

In the [chart overlays](https://docs.pyvista.org/examples/02-plot/chart_overlays.html#sphx-glr-examples-02-plot-chart-overlays-py) tutorial, it is shown how to draw standard and chartMPL charts side by side and update them with a `p.add_slider_widget` that is referred to by a user defined function:
```python
# Method and slider to update all visuals based on the time selection
def update_time(time):
    k = np.count_nonzero(t < time)
    h_line.set_xdata(t[: k + 1])
    h_line.set_ydata(h[: k + 1])
    v_line.update(t[: k + 1], v[: k + 1])
    p.add_mesh(pv.Sphere(1, center=(0, 0, h[k])), name="sphere", render=False)
    p.update()


time_slider = p.add_slider_widget(
    update_time, [np.min(t), np.max(t)], 0, "Time", (0.25, 0.9), (0.75, 0.9), event_type='always'
)

# Start incrementing time automatically
for i in range(1, 50):
    ax.set_xlim([0, t[i]])
    time_slider.GetSliderRepresentation().SetValue(t[i])
    update_time(t[i])

```

I think this works as a [chartMPL._redraw](https://docs.pyvista.org/_modules/pyvista/plotting/charts.html#ChartMPL) event is detected when the SliderRepresentation is updated.

When modifying plotted data outside of Widget use --- for example if animating a series of data alongside a mesh object --- with one of matplotlib's `set_xxxx` type functions to modify collections, this update does not occur and the chart is not redrawn, leaving animations not updating. 

Minimum working example:
```python
from time import sleep
import pyvista as pv #meshing
import numpy as np #data
from matplotlib import pyplot as plt #plotting


x_data = np.random.rand(20,1)*100
y_data = np.random.rand(20,1)*100

fig,ax=plt.subplots()
scat = ax.scatter(x=[],y=[])

ax.set_ylim([0, 100])
ax.set_xlim([0, 100])
cb=plt.colorbar(scat,ax=ax)

p = pv.Plotter(shape=(2, 1),window_size=([1200, 800]))
p.show(auto_close=False,interactive=False,interactive_update=True)
#add mesh
sphere = pv.Sphere()
sphere.cell_data["data"] = np.random.random(sphere.n_cells)
p.add_mesh(sphere, scalars="data", clim=[0, 1])

#add chart
p.subplot(1,0)
chart = pv.ChartMPL(fig, size=(0.8, 0.8), loc=(0.1, 0.1))
p.add_chart(chart)

for idx,r in enumerate(x_data):
    p.subplot(1,0)
    scat.set_offsets( np.column_stack(   (x_data[0:idx],y_data[0:idx])  ) )
    
    p.subplot(0,0)
    sphere.cell_data["data"] = np.random.random(sphere.n_cells)

    p.update()
    sleep(1)

```

Here, the plotter will appear but no data in the chart is shown.

Add the hidden `chart._redraw()` function below the set_offsets call:
```python

for idx,r in enumerate(x_data):
    p.subplot(1,0)
    scat.set_offsets( np.column_stack(   (x_data[0:idx],y_data[0:idx])  ) )
    chart._redraw()

    p.subplot(0,0)
    sphere.cell_data["data"] = np.random.random(sphere.n_cells)

    p.update()
    sleep(1)


```

The plotter now appears and the chartMPL object is animated.

## Solutions
This is a fairly expected use case of the chartMPL functionality in my opinion, allowing for much greater control of animated charts than the core PyVista chart capability. It took me a very long time trying to debug and hunting through the source to find the missing `chart._redraw()` call. 
So to that end, either:

1. The chartMPL object is redrawn when p.update is called (this seems most intuitive to me)
2. The chartMPL detection for a draw event is extended to the `set_xxxx` functions from mpl (might be hard to do as I don't think `set_xxxx` functions are an mpl_connect event)
3. At least mention the need for manually using `chart._redraw()` in one of the tutorials or a new tutorial based on this above MWE.

Thanks!

dcbr:
### Overview

As requested in #3380 this PR adds functionality to automatically redraw `ChartMPL` charts whenever the `Plotter.update()` method is called. Furthermore, a bug in the chart's `_resize` method is fixed. Resolves #3380.

### Details

- Adds a `redraw_on_render` property for `ChartMPL` (default `True`), which enables automatic redrawing of these charts when the `Plotter.render()` method is called. To support this feature, extra `on_plotter_render` methods to the `Renderer` and `Renderers` classes are needed to discern between a manual call to `Plotter.render` and manual/automatic calls to `Renderer.Render` (which invoke VTK `StartEvent`, also used by charts).
- Previously the logic in the `_resize` method of charts used exact comparisons of floats, which lead to unnecessary resizings (and redrawings) of charts. This PR converts all current and target dimensions to integers to resolve this bug.


banesullivan (approved):
LGTM... the general updates to the `BasePlotter` look good to me (calling `update()` on the renderers when `BasePlotter.update()` is called.

Otherwise, I trust you, @dcbr, as I'm totally unfamiliar with what's going on with the charting

dcbr:
Alright, I tested locally on Windows, but I'll add a test tomorrow to improve coverage as well :)

dcbr:
I added the test, ready for review (again) if everything turns green :)

akaszynski:
> I added the test, ready for review (again) if everything turns green :)

Failures will be fixed in #3736 and #3737.

pyvista/plotting/renderer.py
```diff
@@ -2891,6 +2891,13 @@ def close(self):
             self._empty_str.SetReferenceCount(0)
             self._empty_str = None
 
+    def update(self):
+        """Update the renderer."""
+        if self.__charts is not None:
+            for chart in self.__charts:
+                # Notify ChartMPLs to redraw themselves when plotter.update() is called
+                chart._render_event(plotter_update=True)
```
banesullivan, line 2899: There isn't an `Update()` method on the `vtkRenderer` class, so I'd like to avoid adding this method here as it introduces an inconsistency with VTK. "update" calls are typically for pipelining whereas, in the rendering flow, the "render" calls trigger updates.

Should this logic be in a new `render()` method instead, since it is forcing a render event? Which would also call `self.Render()` on the `vtkRenderer`?

pyvista/plotting/renderers.py
```diff
@@ -296,6 +296,11 @@ def set_active_renderer(self, index_row, index_column=None):
             raise IndexError(f'Column index is out of range ({self.shape[1]})')
         self._active_index = self.loc_to_index((index_row, index_column))
 
+    def update(self):
+        """Update all renderers."""
+        for renderer in self:
+            renderer.update()
```
banesullivan, line 302: likewise `render()`?

pyvista/plotting/plotting.py
```diff
@@ -1950,11 +1950,13 @@ def update(self, stime=1, force_redraw=True):
                     # Execution resumed after TerminateApp() call in on_timer callback, destroy the timer
                     self.iren.destroy_timer(Plotter.update_timer_id)
                 # Rerender
+                self.renderers.update()
                 self.render()
```
banesullivan, line 1954: If using `render()`, then this added line should be unnecessary as the plotter's `render()` call should propagate

pyvista/plotting/plotting.py
```diff
@@ -1950,11 +1950,13 @@ def update(self, stime=1, force_redraw=True):
...
                 Plotter.last_update_time = curr_time
                 return
 
         if force_redraw:
+            self.renderers.update()
```
banesullivan, line 1959: similarly

pyvista/plotting/renderer.py
banesullivan: I suppose this gets nuanced/confusing with the `update()` method on the `BasePlotter`...

banesullivan (changes requested):
I know I gave this an initial approval but now I have some counter thoughts on adding an `update()` method to the `Renderer` class - IMO, this should be in the `render()` calls

dcbr:
Hmm yea I thought of adding it to the `render()` method instead. The reason I chose not to, is because `render` is already called in various other plotting methods that are unrelated to charts (and thus cause unnecessary overhead of redrawing the `ChartMPL` figures).
Maybe a workable middleground is to split the rendering step (internally) in two parts: `render_viewport` for rendering the 3D parts (and calling `vtkRenderer.Render`) and `render_scene` for rendering the 2D parts (and updating `ChartMPL` figures). The existing calls to `render` from other plotting methods can then be converted to `render_viewport` such that Charts are not impacted by them. And the existing `render` method then just calls both `render_viewport` and `render_scene`. What do you think?

dcbr:
Maybe for now I can just move it to the `render` method and if we ever want to improve performance (or in case there are complaints), we can still do this split later.

pyvista/plotting/renderer.py
dcbr: Renamed and moved everything to `render` calls.

pyvista/plotting/renderer.py
```diff
@@ -2891,6 +2891,13 @@ def close(self):
             self._empty_str.SetReferenceCount(0)
             self._empty_str = None
 
+    def render(self):
+        """Notify renderer components of manual render call."""
+        if self.__charts is not None:
+            for chart in self.__charts:
+                # Notify ChartMPLs to redraw themselves when plotter.render() is called
+                chart._render_event(plotter_render=True)
```
banesullivan, line 2899: Hm, I appreciate you changing this to `render()` but now I'm thinking`self.Render()` should be called here, i.e., `vtk.vtkRenderer.Render`

banesullivan (approved):
Overall, I'm happy with this! but I think `Renderer.render()` needs to call `self.Render()` too

pyvista/plotting/renderer.py
banesullivan: Or if you want to avoid the overhead of calling `Render()` then changing this method to `render_charts()` might be a good option

dcbr: Calling `self.Render()` would lead to duplicate render calls though, so I don't think we should do that. I don't really like the current naming either to be honest, but I couldn't come up with a good alternative yet, suggestions are welcome.
The purpose of the method is to notify charts of an upcoming render event that is manually triggered (by an explicit `plotter.update` or `plotter.render` call), so maybe `on_plotter_render` or `on_explicit_render` (these are quite long), or `notify_render`, `on_render` or `before_render` (the problem I have with these is that the method is not called for all render events which might be confusing)?

dcbr: I went ahead with `on_plotter_render` as the method name.

banesullivan (approved):
`on_plotter_render()` is a good choice. LGTM!

PR #3729, opened 2022-12-19
[HOTFIX] VTK 9.0.3  + NumPy<1.24 CI

banesullivan:
NumPy 1.24.0 was released yesterday which deprecated `np.bool`. VTK 9.0.3 relied on that:

```py
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/vtkmodules/util/numpy_support.py:216: in vtk_to_numpy
    assert typ in get_vtk_to_numpy_typemap().keys(), \
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/vtkmodules/util/numpy_support.py:74: in get_vtk_to_numpy_typemap
    _vtk_np = {vtkConstants.VTK_BIT:numpy.bool,
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/numpy/__init__.py:284: in __getattr__
    raise AttributeError("module {!r} has no attribute "
E   AttributeError: module 'numpy' has no attribute 'bool'
```

These changes limit NumPy for our VTK 9.0.3 CI

PR #3736, opened 2022-12-20
[hotfix] Use latest VTK version for conda tests

banesullivan:
The conda tests were using VTK 8.2... why? This updates conda to use whatever the latest version is which seems most appropriate to me. Other versions of VTK are checked in other CI routines.

Either way, this needs to be updated because VTK8.2.2 is incompatible with the latest NumPy 1.24

@pyvista/developers, I plan on merging this quickly to fix broken CI

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -56,8 +56,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CONDA_ALWAYS_YES: 1
-      VTK_VERSION: 8.2
-      conda_env: pyvista-vtk8.2
+      VTK_VERSION: 9.2.2
+      conda_env: pyvista-vtk9.2.2
       DISPLAY: ':99.0'
       PYVISTA_OFF_SCREEN: True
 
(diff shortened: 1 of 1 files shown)


banesullivan:
This cannot pass until #3737 merges

akaszynski:
We were testing conda using `vtk==8.2` because it was the only way we could actually test that 8.2 works. It's a hard version to support because it was never released on PyPI, but I think at this point we should just drop it and (soon) enforce a hard requirement to use `vtk>=9.0`.

banesullivan:
> should just drop it and (soon) enforce a hard requirement to use vtk>=9.0.

I'm actually thinking `vtk>=9.1` due to the NumPy 1.24 incompatibility for `<9.1`. Ref #3729

MatthewFlamm:
If we drop CI testing, we should also bump the vtk dependency version.

Is there any benefit to having this conda test now?  We still use pip to install this package.

banesullivan:
> Is there any benefit to having this conda test now? We still use pip to install this package.

The conda-forge VTK recipe has a mismatch in enabled modules from the wheel on PyPI. IMO, it makes the most sense to test against the latest conda-forge VTK to ensure PyVista gracefully handles any disabled modules that we have come to expect by default from the PyPI wheel

Issue #3690, opened 2022-12-06
Towards a single, PyVista/VTK-native Jupyter Backend


banesullivan:
I'm working to completely overhaul the state of PyVista's Jupyter backends to provide a new, single "PyVista Backend"

The new backend will be included in #3385 as it will initially be based on `trame`.

In later pull requests, I will deprecate the other backends in the following fashion to move users towards a single backend that we will maintain and support moving forward.

#### Full deprecation and removal (one PR)

- `ipyvtklink` - Full deprecation and removal as it is entirely replaced by the "server" variant of the new backend with much better performance. Usage of this will be automatically replaced with the new backend's server variant.
- `itkwidgets`/`PlotterITK` - Full deprecation and removal (not widely used as far as I can tell)

#### Deprecation and notice of future removal (one PR)

- `panel` - Deprecation notice of no further support and future removal
- `ipygany` - Deprecation notice of no further support and future removal

#### No Action

- `pythreejs` - This backend will remain untouched for now. It's the only backend that does not require a virtual frame buffer which is quite significant for some edge cases. Though we need to look into issues like https://github.com/pyvista/pyvista/issues/3221
- The `static` and `none` backend options will remain as-is.

-------

The new "PyVista Backend" will initially be built on Kitware's `trame` - a new framework for building dynamic web applications with Python with great support for VTK. Under the hood, trame uses features embedded in VTK Python and `wslink` to handle two modes of client-server 3D rendering interaction: 

1. **server-side rendering** with screenshots streaming the client (this is effectively the same as `ipyvtklink` just much improved)
2. **client-side rendering** where a scene graph's state is synchronized between server and client. This is rendered with VTK.js on the client side and requires a `vtkRenderWindow` on the server.

The initial PR will bundle this into a minimally working state. While I'm calling this "minimal" (MVP), it is not minimal in terms of feature completeness but rather minimal in the architecture sense. Initially, this will fully bring in `trame` and all of its dependencies (fairly lightweight). I will later iterate on the architecture to reduce the dependencies and eventually use the Jupyter comms over `wslink` and `jupyter-server-proxy`.

Further details on the new backend will be outlined in #3385 as I wrap up the implementation (in the coming weeks).

## Name of new backend

I'd like for this new backend to have a general name to indicate that it is the main Jupyter backend for PyVista. Considering it has two variants (or modes), I'm thinking of using the names `client` and `server`.

```py
[...]

PR #3737, opened 2022-12-20
Deprecate and remove PlotterITK

banesullivan:
I'm pretty sure no one is actually using this, and it wasn't even registered as a `jupyter_backend`. Maintaining this is more of a burden than a benefit, and there are much better alternatives to this API... like using `itkwidgets` directly which supports PyVista data types!

This is in support of #3690

We're going to deprecate and remove this immediately without a transition plan or warning because it is not a core feature and not a proper "jupyter backend"


banesullivan:
Further justification: `itkwidgets` is broken on import and causing CI failures:

```
tests/jupyter/test_itk_plotting.py:11: in <module>
    import itkwidgets
/usr/share/miniconda/envs/pyvista-vtk/lib/python3.8/site-packages/itkwidgets/__init__.py:13: in <module>
    from .widget_viewer import Viewer, view
/usr/share/miniconda/envs/pyvista-vtk/lib/python3.8/site-packages/itkwidgets/widget_viewer.py:14: in <module>
    import itk
/usr/share/miniconda/envs/pyvista-vtk/lib/python3.8/site-packages/itk/__init__.py:28: in <module>
    from itk.support.extras import *
/usr/share/miniconda/envs/pyvista-vtk/lib/python3.8/site-packages/itk/support/extras.py:35: in <module>
    import itk.support.types as itkt
/usr/share/miniconda/envs/pyvista-vtk/lib/python3.8/site-packages/itk/support/types.py:172: in <module>
    ) = itkCType.initialize_c_types_once()
/usr/share/miniconda/envs/pyvista-vtk/lib/python3.8/site-packages/itk/support/types.py:137: in initialize_c_types_once
    _B: "itkCType" = itkCType("bool", "B", np.bool)
/usr/share/miniconda/envs/pyvista-vtk/lib/python3.8/site-packages/numpy/__init__.py:284: in __getattr__
    raise AttributeError("module {!r} has no attribute "
E   AttributeError: module 'numpy' has no attribute 'bool'
=========================== short test summary info ============================
ERROR tests/jupyter/test_itk_plotting.py - AttributeError: module 'numpy' has no attribute 'bool'
```

Looks like it is incompatoble with numpy 1.24 which gave ample warning about the built-in type aliases going away

With this in mind, I'm treating this as a HOTFIX and will merge expeditiously

banesullivan:
CI failure due to needing #3736

adeak:
I'm surprised by this breakage since in https://github.com/pyvista/pyvista/pull/3659#issuecomment-1331424611 I could get rid of these by upgrading all dependencies...

adeak:
I'm also surprised we don't pin the basic dependencies. If numpy were bumped by dependabot we would've seen this coming.

banesullivan:
> I'm also surprised we don't pin the basic dependencies.

Oh huh. We don't pin numpy, do we. At first, I thought this was just that we don't pin the conda environment but yeah we don't pin the core dependencies. We may want to look into that seperately

akaszynski (approved):
Agree with removing this. This has been broken ever since [JupyterLab 3.0](https://blog.jupyter.org/jupyterlab-3-0-is-out-4f58385e25bb), and we have way too many `juptyer_backend`s and should start to converge on a consistent backend soon.

Recommending immediate merge to fix the build.

PR #3740, opened 2022-12-22
Fix builds

akaszynski:
Fix our builds with:
- [x] Downgrade from `ubuntu-22.04` to `ubuntu-20.04`. The slowdown could be for any number of reasons, but I'm guessing it's due to the increased render time by upgrading from `Mesa 21.2.6` to `Mesa 22.0.5`
- [x] Get rid of a whole mess of test warnings that make debugging pyvista a bit more challenging.
- [x] Fix a memory leak issue due to extra pointers when converting `pyvista_ndarray` to `vtkPoints` with `pyvista.vtk_points` ~that only seems to show up on Python 3.11 (`LinuxConda` build)~.

### More details about the memory leak
When `vtk_points` is provided with a `pyvista_ndarray` and `deep=False`, VTK retails a reference to the `pyvista_ndarray` within `numpy_to_vtk`. Since `pyvista_ndarray` contains a reference to the `vtkDataArray`, this results in a circular reference since the `vtkDataArray` is also referenced by the new `vtkDataArray` contained in the new `vtkPoints` object.

One way to work around this is to simply set the new `vtk.vtkPoints`'s data to point to the data referred to by the `pyvista_ndarray`. This way VTK can track references and wif the `pyvista_ndarray` is deallocated there will still be a reference to the underlying data. However, this doesn't work with strides because (as far as I can tell) VTK doesn't support strides the same way that `numpy` does. I could [WriteVoidPointer](https://vtk.org/doc/nightly/html/classvtkDataArray.html#a0c39dda290f319f8e40fa9c8cc094fad), but we would be back at the same place we'd started since void pointers are just that, pointers to a space in memory rather than an object with reference counting.

Voting for get this merged soon since bad builds mean sad developers.

### DIFF
--- a/pyvista/utilities/helpers.py
+++ b/pyvista/utilities/helpers.py
@@ -1198,6 +1198,7 @@ def try_callback(func, *args):
         formatted_exception = 'Encountered issue in callback (most recent call last):\n' + ''.join(
             traceback.format_list(stack) + traceback.format_exception_only(etype, exc)
         ).rstrip('\n')
+        1 / 0
         warnings.warn(formatted_exception)
 
 
--- a/pyvista/utilities/helpers.py
+++ b/pyvista/utilities/helpers.py
@@ -525,18 +525,17 @@ def vtk_points(points, deep=True, force_float=False):
             f'Shape is {points.shape} and should be (X, 3)'
         )
 
-    # use the underlying vtk data if present to avoid memory leaks
-    if not deep and isinstance(points, pyvista.pyvista_ndarray):
-        if points.VTKObject is not None:
-            vtkpts = _vtk.vtkPoints()
-            vtkpts.SetData(points.VTKObject)
-            return vtkpts
-
     # points must be contiguous
     points = np.require(points, requirements=['C'])
     vtkpts = _vtk.vtkPoints()
     vtk_arr = _vtk.numpy_to_vtk(points, deep=deep)
     vtkpts.SetData(vtk_arr)
+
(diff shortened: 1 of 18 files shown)


tests/plotting/test_widgets.py
```diff
@@ -164,6 +164,7 @@ def test_widget_slider(uniform):
     p.add_mesh_isovalue(uniform)
     p.close()
 
+    func = lambda value: value  # Does nothing
```
banesullivan, line 167: Why?

pyvista/utilities/helpers.py
```diff
@@ -1198,6 +1198,7 @@ def try_callback(func, *args):
         formatted_exception = 'Encountered issue in callback (most recent call last):\n' + ''.join(
             traceback.format_list(stack) + traceback.format_exception_only(etype, exc)
         ).rstrip('\n')
+        1 / 0
```
banesullivan, line 1201: huh?

akaszynski: Debug statement, not quite ready for merge.

tests/plotting/test_widgets.py
akaszynski: We have to change the callback from accepting `value, widget` to just `value` since `pass_widget=False` for the remainder of the test.

pyvista/utilities/helpers.py
akaszynski, line 1201:
```suggestion
```

banesullivan (approved):

PR #3743, opened 2022-12-23
Fix link_views() default camera position

banesullivan:
There was a regression when using `link_views()` that caused the `BasePlotter` to not set the camera position to a default isometric view. This fixes that issue.

When setting the `camera` instance on a `Renderer`, `camera_set` is set to `True`. 

https://github.com/pyvista/pyvista/blob/ff6939af681ec9a1d2d6409ccef0c3782f6f2ed8/pyvista/plotting/renderer.py#L325

This is appropriate. However, `link_views()` will copy a `camera` object on each of the subplot renderers and forcibly set `camera_set` to `True` even though it may not have been set for the source renderer/camera. This patch corrects that behavior by properly setting `camera_set` within `link_views()` to match the original state of `camera_set`

```py
import pyvista as pv
p = pv.Plotter(shape=(1,2))
p.add_mesh(pv.Cone())
p.subplot(0,1)
p.add_mesh(pv.Cube())
p.link_views()
p.show()
```

| `main` | This branch |
|---|---|
| Screen Shot 2022-12-23 at 2 34 46 PM | Screen Shot 2022-12-23 at 2 36 53 PM |


banesullivan:
I'd like for this to make it into the upcoming patch release if possible

banesullivan:
I'm actually wondering if we should simplify all of this and put the `camera_set` property on the `Camera` class directly rather than the `Renderer` class? Perhaps that could come as a followup PR and this could land as an initial patch

MatthewFlamm (commented):
I haven't fully wrapped my head around the different camera setting steps.  But the changes look okay to me.  I think a few of the unrelated tests are intended to be Cartesian aligned views.  `test_vector_array` in particular should be modified to use the plane normal view again.

banesullivan:
> I think a few of the unrelated tests are intended to be Cartesian aligned views. test_vector_array in particular should be modified to use the plane normal view again.

I'm not sure I follow what you're getting at?

akaszynski:
> > I think a few of the unrelated tests are intended to be Cartesian aligned views. test_vector_array in particular should be modified to use the plane normal view again.
> 
> I'm not sure I follow what you're getting at?

@MatthewFlamm's right, `test_vector_array` should use `cpos=xy`. Fixing...

pyproject.toml
```diff
@@ -147,7 +147,7 @@ markers = [
     'needs_vtk_version(version): skip test unless VTK version is at least as specified.',
     'needs_download: this test downloads data during execution',
 ]
-
+addopts = '--image_cache_dir ./tests/plotting/image_cache'
```
banesullivan, line 150: This shouldn't be needed if `main` is merged in. Was handled in #3748

MatthewFlamm: We need to merge it in as there was also a bug with marks for skipping tests and high variance leaking out.

akaszynski: As far as I can tell, both options are equivalent. I merged anyway, but I think we should revert that line change since I prefer `image_cache_dir` over `addopts`.

MatthewFlamm (approved):
LGTM

Issue #2515, opened 2022-04-26
`set_active_*` have inconsistent default `preference`


MatthewFlamm:
**Describe what maintenance you would like added.**

As noted in #2500 in [comment](https://github.com/pyvista/pyvista/pull/2500#discussion_r859040768)

https://github.com/pyvista/pyvista/blob/ce1a3741b541d91fd59fa7cda0984fc43732fa0f/pyvista/core/dataset.py#L622

https://github.com/pyvista/pyvista/blob/ce1a3741b541d91fd59fa7cda0984fc43732fa0f/pyvista/core/dataset.py#L662

https://github.com/pyvista/pyvista/blob/ce1a3741b541d91fd59fa7cda0984fc43732fa0f/pyvista/core/dataset.py#L697

-----

adeak:
To carry over my comment from the PR:

There are 15 instances of `preference='point'` (or with double quotes) and 10 instances of `preference='cell'` (or with double quotes) in `pyvista/`. This excludes docstring cameos, but includes code inside method bodies (so these numbers don't reflect the respective default parameters)...

adeak:
> There are probably use cases where it make sense to change the preference, e.g. if a filter only operates on cell data for example, but this should be the exception and documented as such. This assumes that point data is chosen as the overall default, but this does not imply endorsement of that position.

Good point, thanks. I haven't checked context beyond excluding doctest lines (`>>>` or `...` prefix somewhere along the line) from the output of
```sh
git grep 'preference="point"\|'"preference='point'" pyvista/
# vs
git grep 'preference="cell"\|'"preference='cell'" pyvista/
```

Although now that I'm taking another look, I only see 8 non-doc cell preference instances.

PR #3744, opened 2022-12-23
Make field association preference consistent

banesullivan:
resolves https://github.com/pyvista/pyvista/issues/2515

This moves us towards a consistent field association preference with the default being `POINT`. There are notable exceptions to this standard being the `threshold()` and `glyph()` filters. The docstring of the `threshold()` filter documents the justification from #3750 and the `glyph()` is not addressed in this PR.

In later work, I'd like to move away from hardcoding this and set this value in the theme. This should come later though

## Breaking Change

Since the default associations are changing for `set_active_scalars()`, `rename_array()`, and `get_data_range()` which are core dataset methods, this will likely have a breaking change effect for some users that have similarly named POINT and CELL data.

## To do

- [ ] Make preference an Enum type
- [ ] Control preference through `Theme` (setting default argument to `None`) everywhere


pyvista/core/filters/data_set.py
```diff
@@ -2090,7 +2090,7 @@ def glyph(
                 alg.SetIndexModeToOff()
 
         if isinstance(scale, str):
-            dataset.set_active_scalars(scale, preference='cell')
+            dataset.set_active_scalars(scale)
```
banesullivan, line 2093: I'm curious why this may want to be "cell"-preferenced...

MatthewFlamm: Threshold is a cell-wise operation.  It does allow operating on point data, but it is still cell-wise thresholding, e.g. it keeps the entire cell with all of its points if one point meets the threshold.  My guess is that this filter is probably more common to use for cell centered values, but this could easily be a biased view since I typically work with cell data.

This will be a breaking change.  I know that this will cause problems for anyone using OpenFOAM data as the reader by default makes point and cell data with the same name.  If we change this, we need a strong warning in the documentation about it.

banesullivan: This line/diff is for the `glyph()` filter though. Did you mean to comment above on the change the the `threshold()` method signature? 

> This will be a breaking change. .. will cause problems for anyone using OpenFOAM data as the reader by default makes point and cell data with the same name

This is a good point... (for the threshold and any any other filters). 

I would like to change these to be more consistent though. I'm planning to rework the `threshold()` filter in particular as it's currently yielding different results than paraview and the API upstream has changed a bit. ref #3610

pyvista/core/filters/data_set.py
```diff
@@ -935,7 +935,7 @@ def threshold(
         scalars=None,
         invert=False,
         continuous=False,
-        preference='cell',
+        preference='point',
```
banesullivan, line 938: Noting https://github.com/pyvista/pyvista/pull/3744/files/139e0be1c1bba49cffa28e9d9db6b8e5343ca651#r1056828388

This change is likely a breaking change for many users. We may want to handle this change in the `threshold()` operations in one PR when updating the filter for #3610

pyvista/core/filters/data_set.py
banesullivan: With this in mind, I'm going to revert the changes to the `threshold()` operations for now. I think the remaining changes are all low-impact though. I will try to re-work the threshold filter to have the same results as paraview and will save this `preference` change for then

The question still remains why this line in the `glyph()` filter needs to prefer CELL data though

MatthewFlamm: > This line/diff is for the `glyph()` filter though.

Oops sloppy reading of the diff.  I don't even think glyph filter works with cell data?

pyvista/core/filters/data_set.py
banesullivan: Reverted to deal with later

pyvista/core/filters/data_set.py
MatthewFlamm: I think the glyph filter should accept a preference argument and the default should be `'point'`.  This would be a breaking change, but makes it more consistent and also the glyph filter only operates on point data.  If cell data is selected, then this filter uses `cell_centers` to convert to point data.

MatthewFlamm:
It looks like you have opted to skip the `glyph` filter for now, which is fine.  Although I just commented that I think it should be defaulted to 'point' with an argument provided to modify this choice.

I think this is a breaking change and may not be immediately obvious where it will show up as breaking since there could be internal functions that do not specify a default in calling one of these functions.  A breaking change section in the PR description is minimally needed to describe this so we can track it.  Even better would be a directive in the docstrings to indicate that the behavior changed.  I will mark with the breaking change label.

banesullivan:
You bring up a good point and after digging around a little bit, I noticed that we don't even give users a choice in many scenarios.

Take `clip_scalar()` for example. The user can specify a scalar array for the clipping but we do not expose the `preference` in that API thus using the default for `set_active_scalars()` which changes in this PR. So indeed breaking without a way for the user to go back to the old behavior. 

I think I'd like to re-mark this as a draft until we can sift through the API and make sure `preference` is exposed anywhere `scalars` are being used.

banesullivan:
I've gone through a lot of the API and I think this is much better off now exposing the `preference` choice in more (hopefully all) places where its used

PR #3745, opened 2022-12-23
Add `show_vertices` for vertex rendering

banesullivan:
It's a common request to show the vertices of a mesh, like the edges of a mesh. Unfortunately, VTK is not set up to render the vertices of the mesh with `surface` style representation. There is an option, `actor.GetProperty().SetVertexVisibility(True)`, however, this is only valid if the edges are visible and the display parameters are not well exposed. It's easier to make a separate `Actor` all together to display the vertices of the mesh which can be more easily controlled.

These changes add a new `show_vertices` option to `add_mesh()`, which will show the vertices of a mesh in much the same fashion as showing the edges. This option creates a sub-actor for the vertices and matches the styling of the edges.


tests/plotting/test_plotting.py
```diff
@@ -3089,6 +3089,38 @@ def test_plot_points_gaussian_as_spheres(sphere):
...
+        vertex_style='points',
+        vertex_color='r',
+    )
+
+    multiblock_all.plot(
+        color='w',
+        show_vertices=True,
+        point_size=3,
+        render_points_as_spheres=True,
+    )
```
banesullivan, line 3121: I don't think this image was added to the cache?

banesullivan: Woot! CI is finally picking up on that too!

```
FAILED tests/plotting/test_plotting.py::test_plot_show_vertices - RuntimeError: tests/plotting/image_cache/plot_show_vertices_2.png does not exist in image cache
```

akaszynski (dismissed):
LGTM, @banesullivan, please verify.

banesullivan:
Looks excellent! Thanks for implementing as a `show_vertices` option -- I think this is much needed. I do want to add one more test though to make sure `remove_actor` will remove the vertices actor as well. This should work when the `name` of the actor is used but we may want to add some additional logic if the source actor is passed

banesullivan:
Thanks for implementing this @akaszynski, LGTM - ready to go

akaszynski (approved):

Issue #3702, opened 2022-12-09
2D plots: VTK or Matplotlib?


andrea-bistacchi:
### Discussed in https://github.com/pyvista/pyvista/discussions/3641

<div type='discussions-op-text'>

<sup>Originally posted by **andrea-bistacchi** November 23, 2022</sup>
Hello, I was wondering whether 2D plots in PyVista are based upon VTK or Matplotlib. The reason I ask is we are having poor performance plotting many data in Matplotlib and I am considering using PyVista instead.

Another important point is whether mouse interaction is possible in these plots and if there are examples.

Thanks!</div>

andrea-bistacchi:
At the moment we are using Matplotlib but it is very slow. We have a huge mass of data to plot interactively since we are using it to plot geometries,  meshes, grids etc. in 2D maps and cross sections.

That's why we are looking for alternatives.

We also need interaction to pick objects and draw (we are doing this with Matplotlib mouse interaction at the moment).

Thanks!

dcbr:
I have no experience with very large datasets, so for me personally matplotlib is usually fine for 2D plotting.
But in pyvista you really have the choice:

* If it's simple 2D charts you want to draw, you can have a look at the [Charts API](https://docs.pyvista.org/api/plotting/charts/index.html). These are all VTK wrappers, so I assume they will be fast for large datasets. There's also the special [ChartMPL class](https://docs.pyvista.org/api/plotting/charts/_autosummary/pyvista.ChartMPL.html#pyvista.ChartMPL) which can be used to embed existing matplotlib charts in a pyvista plotting window. Note that this last option is probably not what you want, as you are just using matplotlib (which is already too slow for your purpose) with the extra overhead of converting the figure to an image and embedding it in the plotting window.
* For more complex 2D plots (meshes, grids), I think you can just use the existing [3D plotting capabilities](https://docs.pyvista.org/api/plotting/index.html) of pyvista (leveraging VTK under the hood, so fast for large datasets), while keeping the z-component fixed (e.g. 0) and then use the [view_xy method](https://docs.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.view_xy.html) to obtain a 2D look of the scene?

dcbr:
And mouse interaction is possible with each of these alternatives, except for `ChartMPL` (embedded matplotlib charts).

andrea-bistacchi:
Very good, this clarifies!

[...]

PR #3756, opened 2022-12-28
Fix and improve chart interaction

### LINKED ISSUE #3278: Toggling chart interaction in multi-window plot fails
Slightly modified example from the original discussion thread to reproduce the issue.

```python3
import numpy as np
import pyvista as pv

shape = (4, 1)  # 4 by 1 grid
row_weights = [1, 1, 1, 1]
plotter = pv.Plotter(shape=shape, row_weights=row_weights)
rng = np.random.default_rng(1)

for i in range(4):
    plotter.subplot(i, 0)
    # Dummy
    x = rng.standard_normal(100)
    y = rng.standard_normal(100)
    chart = pv.Chart2D()
    chart.grid = True
    chart.background_color = "white"
    chart.scatter(x, y, size=10, style="o", color="cyan")
    plotter.add_chart(chart)

plotter.show()
            
```

Expected behaviour: all charts should become (non-)interactive after a double click.
Actual behaviour: only the bottom most chart correctly becomes (non-)interactive on a double click. The others remain in their initial state (non-interactive).

_Originally posted by @apaaris in https://github.com/pyvista/pyvista/discussions/3252#discussioncomment-3528435_

dcbr:
### Overview

Improve chart interaction and partially fixes a chart interaction bug.

### Details

- Fix chart interaction bug when there are multiple subplots, resolves #3278; but also requires an upstream (VTK) bug fix, see [vtk#18763](https://gitlab.kitware.com/vtk/vtk/-/issues/18763).
- Support interaction with multiple charts in the same renderer (this was an arbitrary constraint to simplify the initial implementation).
- Additionally, support enabling chart interaction from code, as requested recently in #3702.
- Added render event handling in the ``Renderer`` class.
- Fixed triple click being registered as a second double click.
- Added ``Chart`` type, to simplify doc string parameter types.

### TODO

- [x] Finish documentation.
- [x] Add separate chart border/background properties for when a chart is interactive, to make it visually easier to see which charts are interactive. I'll first finalize #2868 and then come back to this.


pyvista/plotting/render_window_interactor.py
```diff
@@ -305,26 +310,66 @@ def update_style(self):
...
 
+        Parameters
+        ----------
+        scene : vtkContextScene, optional
+            The scene to interact with or ``None`` to stop interaction with any scene.
+
+        """
         # Set scene to interact with or reset it to stop interaction (otherwise crash)
+        if vtk_version_info < (9, 3, 0):
+            # TODO: determine in which VTK version !9811 will be available
```
dcbr, line 366: @banesullivan Do you know whether the next VTK version will be 9.2.6 or 9.3.0? Then I can update this check accordingly :)

banesullivan: I don't think that was backported to the `release` branch, so I'd put my money on `9.3`

pyvista/plotting/render_window_interactor.py
```diff
@@ -305,26 +310,65 @@ def update_style(self):
             self._style_class = _style_factory(self._style)(self)
         self.interactor.SetInteractorStyle(self._style_class)
 
-    def _toggle_context_style(self, mouse_pos):
-        """Toggle the context style for chart interaction.
+    def _toggle_chart_interaction(self, mouse_pos):
+        """
+        Toggle interaction with indicated charts.
```
dcbr, line 315:
```suggestion
        """Toggle interaction with indicated charts.
```

pyvista/plotting/renderer.py
```diff
@@ -293,9 +294,52 @@ def _charts(self):
         # lazy instantiation here to avoid creating the charts object unless needed.
         if self.__charts is None:
             self.__charts = Charts(self)
-            self.AddObserver("StartEvent", partial(try_callback, self._render_event))
+            self.add_render_event(self._before_render_event, "before")
         return self.__charts
 
+    def add_render_event(self, callback, event="before"):
```
dcbr, line 300: I added these methods at an earlier stage in this PR for a workaround that is no longer needed at this point. I think it's still ok to keep these in though?

dcbr (commented):
This is ready for review now!

pyvista/plotting/render_window_interactor.py
```diff
@@ -305,26 +310,65 @@ def update_style(self):
...
+        Set the context style interactor or switch back to previous interactor style.
 
+        Parameters
+        ----------
+        scene : vtkContextScene, optional
+            The scene to interact with or ``None`` to stop interaction with any scene.
+
+        """
         # Set scene to interact with or reset it to stop interaction (otherwise crash)
+        if vtk_version_info < (9, 3, 0):
```
dcbr, line 365:
```suggestion
        if vtk_version_info < (9, 3, 0):  # pragma: no cover
```

pyvista/plotting/renderer.py
banesullivan: I think this method should be `interaction_event`, not `add_render_event` as the VTK event types are `StartEvent` and `EndEvent`. Further, this obfuscates the `add_on_render_callback()` I added in #3385.

banesullivan: Oof, I think I may need to refactor `add_on_render_callback()` to be on the `Renderer`...

pyvista/plotting/renderer.py
```diff
@@ -2939,6 +3002,7 @@ def remove_environment_texture(self):
     def close(self):
         """Close out widgets and sensitive elements."""
         self.RemoveAllObservers()
+        self._render_events.clear()
```
banesullivan, line 3005: Thanks for pointing this out... we should do the same for `add_on_render_callback()`'s callbacks on close

banesullivan (commented):
@dcbr, this is looking good. Some minor concerns about the render event listeners.

Do you want to land this today for `0.38`?

Issue #3423, opened 2022-10-06
[Bug] PVDReader : << KeyError: 'part' >>


MedAdh42:
Hi, I'm trying to use PyVista to read simulation results stored in the .pvd format (GEOSX simulation, see attached zip file).

When i try to read the data using :

`import pyvista as pv 

filename = 'vtkWellPump.pvd'

results = pv.read(filename)`

It throws the following error :

`File ......code/meshing/Python/python3/install/Python-3.9.10/lib/python3.9/site-packages/pyvista/utilities/reader.py:1736, in PVDReader._parse_file(self)
   1731 for element in dataset_elements:
   1732     element_attrib = element.attrib
   1733     datasets.append(
   1734         PVDDataSet(
   1735             float(element_attrib.get('timestep', 0)),
-> 1736             int(element_attrib['part']),
   1737             element_attrib['file'],
   1738             element_attrib.get('group'),
   1739         )
   1740     )
   1741 self._datasets = sorted(datasets)
   1742 self._time_values = sorted(list(set([dataset.time for dataset in self.datasets])))

KeyError: 'part'
`

[PyVista.zip](https://github.com/pyvista/pyvista/files/9726351/PyVista.zip)

PR #3760, opened 2022-12-29
Fix PVD reader, allow no part

beroda:
### Overview
Resolves https://github.com/pyvista/pyvista/issues/3423

### Details

Set the default `part` attribute to 0 when missing from PVD file.


beroda:
Unit tests failed because the PVD file is not yet merged in vtk-data. See https://github.com/pyvista/vtk-data/pull/24

beroda:
Since https://github.com/pyvista/vtk-data/pull/24 has been accepted, how can I relaunch GitHub actions ?

MatthewFlamm:
I reran tests, and some failures look real `tests/utilities/test_reader.py::test_pvdreader_no_part_group - FileNotFoundError: [Errno 2] No such file or directory: 'C:/Users/runneradmin/AppData/Local/pyvista_3/pyvista_3/Cache/PVD/paraview/dualSphereAnimation.zip.unzip\\dualSphereAnimation4NoPart.pvd'` or it might be a cache issue

beroda:
I think it might be a cache issue, since locally on my laptop I had to delete the old unzipped folder inside my vtk-data cache location so that the new zip file (containing the new missing file) was correctly downloaded/unzipped from the vtk-data repo.

I don't really know how to fix this, has it seems to only affect windows but not python 3.8 ? Weird ...

MatthewFlamm:
> I don't really know how to fix this, it is beyond my CI knowledge.
> It seems to only affect windows but not python 3.8 ? Weird ...

It is passing all Linux and MacOs tests, so it is something about windows path or filling naming I suspect.

beroda:
> Edit: however if this is the case, it is curious since you've more or less copied the preceding test. I'll have my fingers crossed that when we can merge with main after the next PR merge that we find out that it is a cache issue.

I guess that the preceding test I copied was working because the zip file that was needed by the test was new to `vtk-data`. Therefore, Pooch downloaded-it and unzipped it for all CIs because it could not find it in its registry.

In the current case, I had to modify the zip file in order to add another pvd file. And because we don't check the hash of files in `examples/downloads.py` (see comments at line 83) , Pooch never updates the zip file (at least for Windows ?).
I might be wrong since I am a Pooch beginner (only read the some small parts of the doc) but it seems logical to me.

Maybe we should delete all downloads when starting a new CI action ? (at least for Windows CI ?). There is a `examples.delete_downloads()` command available in pyvista.

MatthewFlamm:
Im not that familiar of why we have these caches for the downloads, but I'm wondering whether we need to separate the caches according to os, currently they all use the same key.

https://github.com/pyvista/pyvista/blob/72c7d62ea01ec61c73d5970215d4aa5d84d11d8e/.github/workflows/testing-and-deployment.yml#L247

beroda:
I am not familiar with github actions and I have no idea how to deal with these fails.

akaszynski:
Working on this one locally.

akaszynski:
> Im not that familiar of why we have these caches for the downloads, but I'm wondering whether we need to separate the caches according to os, currently they all use the same key.

Potentially, yes. Strangely, the test above this one `test_pvdreader_no_time_group`, doesn't fail.

Skipping this until I get my hands on a windows box.

akaszynski (approved):
LGTM.

MatthewFlamm (approved):
Nice!

Issue #3241, opened 2022-08-30
Plotter.show(return_img) is broken


natter1:
### Describe the bug, what's wrong, and what you expected.

In `plotting.py` `Plotter.show()` is a bug preventing the `return_img` parameter from working when `return_cpos` is `False`:

```
        # If user asked for screenshot, return as numpy array after camera
        # position
        if return_img or screenshot is True:
            if return_cpos:
                return self.camera_position, self.last_image

        if return_cpos:
            return self.camera_position
```
Not sure whats the expected behaviour. Mybe `if return_img=True` and `return_cpos=False` only return the image?

### Steps to reproduce the bug.

```
print(plotter.show(return_img=True))
print(plotter.show(return_img=True, return_cpos=True))
print(plotter.show(return_img=True, return_cpos=False))
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Tue Aug 30 15:23:49 2022 Mitteleurop�ische Sommerzeit
                OS : Windows
            CPU(s) : 12
           Machine : AMD64
      Architecture : 64bit
               RAM : 24.0 GiB
       Environment : Python
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : Quadro 2000/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 353.30
  Python 3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit
  (AMD64)]
           pyvista : 0.34.1
               vtk : 9.0.3
             numpy : 1.22.4
           imageio : 2.19.3
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.2
             scipy : 1.8.1
              tqdm : 4.64.0
--------------------------------------------------------------------------------
```

banesullivan:
Oof yeah, that logic is off. This should be an easy fix! If you're up for it, feel free to take a go at fixing this. Otherwise one of the core devs will get around to this soon

akaszynski:
> Otherwise one of the core devs will get around to this soon

This is a fairly high priority and easy fix. Will work soon. @natter1, is this for PyAnsys?

natter1:
@akaszynski yeah I found it via PyMAPDL. Would be nice if someone with a ready to go workflow would add a fix.

PR #3761, opened 2022-12-29
Resolves return_img for Plotter.show()

beroda:
### Overview

Resolves https://github.com/pyvista/pyvista/issues/3241

### Details

Now, `Plotter.show()` with `return_img = True` works even if `return_cpos = False`.


MatthewFlamm:
Sorry for the poor dev experience, but our image cache regression testing is currently broken.  After we fix it, hopefully in #3748, you will have to add in images to the cache.

We shouldn't merge this until #3748 is merged.

pyvista/plotting/plotting.py
```diff
@@ -6237,6 +6237,8 @@ def fun(plotter):
         if return_img or screenshot is True:
             if return_cpos:
                 return self.camera_position, self.last_image
+            else:
+                return self.last_image
```
banesullivan, line 6241:
```suggestion
            return self.last_image
```

tests/plotting/test_plotting.py
```diff
@@ -328,6 +328,17 @@ def test_plot_return_cpos(sphere):
...
 
+def test_plot_return_img_without_cpos(sphere: pyvista.PolyData):
+    img = sphere.plot(return_cpos=False, return_img=True, screenshot=True)
+    assert isinstance(img, np.ndarray)
+
+
+def test_plot_return_img_with_cpos(sphere: pyvista.PolyData):
+    cpos, img = sphere.plot(return_cpos=True, return_img=True, screenshot=True)
+    assert isinstance(cpos, pyvista.CameraPosition)
+    assert isinstance(img, np.ndarray)
```
banesullivan, line 339: As @MatthewFlamm pointed out, our image regression testing is currently broken, but these tests will likely need to add images for the cache as I believe the default is for all tests in `test_plotting.py` to use the cache... I think...

banesullivan (changes requested):
Awesome! Thank you for fixing this @beroda! 

We can handle adding the image regression testing nonsense here, so if this sounds foreign to you, don't worry about it!

beroda:
Not sure if the tests I implemented for this PR require to be compared to cached images. They only verify that the return statements in Plotter.show works as expected.

banesullivan:
> Not sure if the tests I implemented for this PR require to be compared to cached images. They only verify that the return statements in Plotter.show works as expected.

Your absolutely correct. However since these tests are placed in the `tests/plotting/test_plotting.py` module, image regression testing is automatic. Perhaps a better place for these tests would be in `tests/test_plotter.py`

tests/plotting/test_plotting.py
banesullivan: See PR comment about moving to `tests/test_plotter.py`

pyvista/plotting/plotting.py
banesullivan: I suppose this is purely stylistic

tests/plotting/test_plotting.py
beroda: Sounds better. I will move them asap.

beroda: Done

MatthewFlamm (approved):
I no longer think we need to wait for the other PR first since this no longer needs image cache testing.

banesullivan (approved):

PR #1989, opened 2021-12-31
🎉 HAPPY NEW YEAR 2022 🎉

tkoyama010:
### Overview

🎉 HAPPY NEW YEAR 2022 🎉

 
relate to https://github.com/pyvista/pyvista/pull/1076

### Details

- Updated the year of LICENSE to 2022.

PR #3765, opened 2022-12-31
🎉 HAPPY NEW YEAR 2023 🎉

tkoyama010:
### Overview

🎉 HAPPY NEW YEAR 2023 🎉

 
relate to https://github.com/pyvista/pyvista/pull/1989

### Details

- Updated the year of LICENSE to 2023.

### DIFF
--- a/doc/pyvista-binder-config.yml
+++ b/doc/pyvista-binder-config.yml
@@ -3,4 +3,4 @@ default_context:
     github_account: "pyvista"
     description: "Quick examples demonstrating what PyVista can do! Simply launch this Binder project and test drive PyVista"
     author_name: "PyVista Developers"
-    year: "2022"
+    year: "2022-2023"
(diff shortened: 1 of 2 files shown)


adeak:
Do we also want to update the year in `doc/pyvista-binder-config.yml`? It's not clear to me what that's for exactly... and if we don't touch it then the old date might be more correct.

doc/pyvista-binder-config.yml
```diff
@@ -3,4 +3,4 @@ default_context:
     github_account: "pyvista"
     description: "Quick examples demonstrating what PyVista can do! Simply launch this Binder project and test drive PyVista"
     author_name: "PyVista Developers"
-    year: "2022"
+    year: "2022-2023"
```
banesullivan, line 6:
```suggestion
    year: "2023"
```

I don't think this needs to be a range and the active year is sufficient

banesullivan, line 6:
```suggestion
    year: "2023"
```

I don't think this needs to be a range and the active year is sufficient

banesullivan:
> Do we also want to update the year in `doc/pyvista-binder-config.yml`? It's not clear to me what that's for exactly... and if we don't touch it then the old date might be more correct.

I'm not sure. These days I'm generally of the opinion that none of this needs to be updated. I think This PR is good to go as is

banesullivan (approved):

PR #3776, opened 2023-01-03
Use --fail_extra_image_cache on CI

banesullivan:
`pytest-pyvista` has an option for `fail_extra_image_cache` which will fail if the test is missing its image in the cache. We need to be using this option on CI but NOT locally. This will make sure that new Plotting API tests have an image added to the cache (which is currently not the case and tests might be silently failing)


MatthewFlamm (approved):
LGTM

Issue #83, opened 2019-02-11
enable math-tex for labels


banesullivan:
I know it is somehow possible to use math-tex / LaTeX math formatting for labels in VTK, but how can we enable it in `vtki`? For example:

```py
plotter.add_bounds_axes(xlabel='$F^2$')

plotter.add_mesh(data, stitle='$stuff^2$')
```

akaszynski:
Looks like paraview does it using ``matplotlib`` [Equation Rendering](https://www.paraview.org/Wiki/ParaView/Equation_Rendering).  It's already an optional requirement, and I can implement a module level function for rendering text.

banesullivan:
Awesome, thanks @akaszynski!

akaszynski:
Opened an issue with ``Kitware`` [https://gitlab.kitware.com/vtk/vtk/issues/17528](Issue #17528).

They didn't compile the ``vtk`` wheels with mathtext support as matplotlib might have conflicted with Python somehow.  I can get halfway there, but it's unclear how to take a vtkImageData and have it replace a text actor or text property.

banesullivan:
Hmm, if you have the latest ParaView installed, you could activate a Python 2.7 environment and edit your path to point to ParaView's local VTK Python package to test/debug.

```py
import sys
sys.path.insert(0, '/Path/to/ParaView/contents/bin/site-packages/or/something/')
import vtk
```

```

akaszynski:
Still working on this, but I don't think this will be something we can solve on our end.  It's possible to render equations and load the resulting png into an actor, but getting the text rendered in the same location as the original label would be a nightmare.  It would mean removing the existing label and property and then adding in the equation label by replacing the existing text, which would be quite difficult based on my dive into their source.

I've suggested that future ``vtk`` wheels be built with equation rendering enabled.  If this isn't possible, can make another attempt trying to render using ``matplotlib``

banesullivan:
If VTK were built with equation rendering enabled, rendering mathtex would be as simple as passing the latex string to the label (`p.add_bounds_axes(xlabel='$\frac{1}{x}$')`), correct?

If so, I propose we close this issue and lobby VTK to compile the upcoming 8.2.0 Python release: https://discourse.vtk.org/t/vtk-8-2-0-come-and-get-it/174

banesullivan:
[...]

PR #3781, opened 2023-01-06
Add LaTeX symbol test

banesullivan:
Adding an image regression test to see where LaTeX symbols are supported.

Reference https://github.com/pyvista/pyvista/discussions/2928

LaTeX symbols should be supported for:

- VTK<=9.2.2 and matplotlib<3.6
- VTK>9.2.2 and matplotlib>=3.6

I would like to use this PR as a means of finding out which combinations of VTK and Matplotlib versions support LaTeX symbols and perhaps document this

See also #83

Note that MathTex is currently not enabled for the conda builds of VTK. Reference https://github.com/conda-forge/vtk-feedstock/issues/79


tests/plotting/test_plotting.py
```diff
@@ -2468,6 +2468,20 @@ def test_add_text():
     plotter.show()
 
 
+@pytest.mark.skipif(
+    vtk.vtkMathTextFreeTypeTextRenderer().MathTextIsSupported(),
+    reason='Math text is not supported.',
+)
```
banesullivan, line 2474: This is needed for the conda job due to https://github.com/conda-forge/vtk-feedstock/issues/79

.github/workflows/testing-and-deployment.yml
```diff
@@ -180,6 +180,10 @@ jobs:
         if: ${{ matrix.vtk-version == '9.0.3' }}
         run: pip install 'numpy<1.24'
 
+      - name: Limit Matplotlib for VTK<=9.2.2
+        if: ${{ matrix.vtk-version != 'latest' }}
```
tkoyama010, line 184: I am not familiar with GitHub Actions syntax. Is this the correct wording for small versions?
```suggestion
        if: ${{ matrix.vtk-version <= '9.2.2' }}
```

tkoyama010 (changes requested):
LGTM with 1 small suggestion.

.github/workflows/testing-and-deployment.yml
banesullivan: AFAIK, this won't work as we're doing literal string comparison here

tkoyama010: OK. If you don't mind please use this.
```suggestion
        if: ${{ matrix.vtk-version != '9.2.2' }}
```

tkoyama010: OK. I got As is is the best choice.

tkoyama010 (dismissed):

tests/plotting/test_plotting.py
```diff
@@ -2468,6 +2468,20 @@ def test_add_text():
     plotter.show()
 
 
+@pytest.mark.skipif(
+    vtk.vtkMathTextFreeTypeTextRenderer().MathTextIsSupported(),
+    reason='Math text is not supported.',
```
adeak, line 2473: Is it on purpose that a call to `MathTextIsSupported` gives us whether math text is _not_ supported?

banesullivan: 🤦🏻 Thank you for catching this... this is what happens when I code at night.

.github/workflows/testing-and-deployment.yml
banesullivan: Yeah, its not great but its what works for the current CI configuration. When we update the VTK versions we test against, this will have to be updated.

tests/plotting/test_plotting.py
banesullivan: It looks like the conda build of VTK is giving `True` for `vtk.vtkMathTextFreeTypeTextRenderer().MathTextIsSupported()` even though it isn't supported 😕 

https://github.com/conda-forge/vtk-feedstock/issues/79

akaszynski:
Recommending skip conda for that test.

banesullivan:
Does pytest have a command line option to skip a test by name? As in the opposite of the `-k` option?

My thought being that we could mark that test to skip specifically for our conda CI workflow

Using `pytest.mark.skip` is difficult for this as we need to see if VTK was installed by conda, which I'm unsure of how to do without a bunch of sys calls. Simply seeing if conda is managing the environment isn't sufficient as I use conda to manage my virtual environments yet install VTK with pip

adeak:
> Does pytest have a command line option to skip a test by name? As in the opposite of the `-k` option?

We can do `-k not specific_test_name` if the name doesn't match other tests. I've seen a per-module ignore seitch too but don't know if that can work on the test level.

banesullivan:
That did it. Thanks, @adeak

akaszynski (approved):
This is good to go as-is. We should document somewhere that we need `matplotlib<3.6` for `vtk<=9.2.2` for LaTeX symbols and ideally trigger a runtime error somewhere if that's not the case when adding it in `add_text`

tkoyama010 (approved):

PR #3787, opened 2023-01-06
MAINT: Improve GC message

larsoner:
Resurrecting https://github.com/pyvista/pyvistaqt/pull/66 I realized (I think) that GC was not actually being tested here. Let's try explicitly enabling it via `usefixtures` and see if all tests pass. Also update the fixture to have a nicer output on failure reporting the referrers, like (from same code in `pyvistaqt` PR):
```
_____________________________________________________________ ERROR at teardown of test_background_plotting_menu_bar ______________________________________________________________
tests/conftest.py:87: in check_gc
    assert len(after) == 0, msg
E   AssertionError: Not all objects GCed:
E     vtkEDLShading: ["dict: d['_edl_pass']", '[<vtkmodules.vtkRenderingOpenGL2.vtkEDLShading(0x556489b25b30) at 0x7ff47c7b5d20>]']
E     vtkRenderPassCollection: ["dict: d['_RenderPasses__pass_collection']"]
E     vtkSequencePass: ["dict: d['_RenderPasses__seq_pass']"]
E     vtkCameraPass: ["dict: d['_RenderPasses__camera_pass']"]
E     
E   assert 4 == 0
E    +  where 4 = len([<vtkmodules.vtkRenderingOpenGL2.vtkEDLShading(0x556489b25b30) at 0x7ff47c7b5d20>, <vtkmodules.vtkRenderingOpenGL2.vtk...s(0x55648afaa4a0) at 0x7ff47c7b5a80>, <vtkmodules.vtkRenderingOpenGL2.vtkCameraPass(0x556489aaf530) at 0x7ff47c7b5660>])
```


larsoner:
Okay I think I was wrong and that the `autouse=True` is enough. So this really just ports over the improved `check_gc` fixture now.

Issue #3782, opened 2023-01-06
Documentation build is failing on `main`


banesullivan:
The Documentation CI job is failing on `main` from #3715

cc @akaszynski, @beroda, @MatthewFlamm

I'm not sure why the job passed for the PR but not after merging...

Here are snapshots of the failures/warnings:

```
reading sources... [  2%] api/core/_autosummary/pyvista.DataSet.cell_bounds
/home/runner/work/pyvista/pyvista/pyvista/core/dataset.py:2701: PyVistaDeprecationWarning: `cell_bounds` is deprecated. Use `cell[i].bounds` instead
  warnings.warn(
reading sources... [  2%] api/core/_autosummary/pyvista.DataSet.cell_centers
reading sources... [  2%] api/core/_autosummary/pyvista.DataSet.cell_data
reading sources... [  2%] api/core/_autosummary/pyvista.DataSet.cell_data_to_point_data
reading sources... [  2%] api/core/_autosummary/pyvista.DataSet.cell_n_points
/home/runner/work/pyvista/pyvista/pyvista/core/dataset.py:2637: PyVistaDeprecationWarning: `cell_n_points` is deprecated. Use `cell[i].n_points` instead
  warnings.warn(
reading sources... [  3%] api/core/_autosummary/pyvista.DataSet.cell_point_ids
/home/runner/work/pyvista/pyvista/pyvista/core/dataset.py:2731: PyVistaDeprecationWarning: `cell_type` is deprecated. Use `cell[i].type` instead
  warnings.warn(
/home/runner/work/pyvista/pyvista/pyvista/core/dataset.py:2766: PyVistaDeprecationWarning: `cell_point_ids` is deprecated. Use `cell[i].point_ids` instead
  warnings.warn(
reading sources... [  3%] api/core/_autosummary/pyvista.DataSet.cell_points
reading sources... [  3%] api/core/_autosummary/pyvista.DataSet.cell_type
/home/runner/work/pyvista/pyvista/pyvista/core/dataset.py:2731: PyVistaDeprecationWarning: `cell_type` is deprecated. Use `cell[i].type` instead
  warnings.warn(
```

```
reading sources... [  4%] api/core/_autosummary/pyvista.DataSet.point_is_inside_cell
/home/runner/work/pyvista/pyvista/pyvista/core/dataset.py:2701: PyVistaDeprecationWarning: `cell_bounds` is deprecated. Use `cell[i].bounds` instead
  warnings.warn(
```

[...]

Issue #3783, opened 2023-01-06
Volume np.uint8 casting issues with NumPy 1.24


tkoyama010:
With a newer numpy now we get a warning during doc build from this line 

https://github.com/pyvista/pyvista/pull/3170/files#diff-355952f3009b3d15d0d85ebb510556c3db59722b98c0d5c7d57abdb2b6ea5363R2974

https://github.com/pyvista/pyvista/blob/68d27ef9d3b519428a7cd521cde495b7bcc2fdf6/pyvista/plotting/plotting.py#L3596

```
/home/adeak/pyvista/pyvista/plotting/plotting.py:3596: RuntimeWarning: invalid value encountered in cast                                                                 
  volume[title] = np.array(scalars, dtype=np.uint8)
```

I think what's going on is that there's one or more `nan`s in there which can't be cast to uint8.

Do you remember by any chance
1. why you replaced `scalars.astype(np.uint8)` with `np.array(scalars, dtype=np.uint8)`? (We could add `casting='unsafe'` to the original, but of course it will still replace nans and infs with `0`s which might merit a warning note in the docs.)
2. if you considered the case where the input scalars already has `dtype=np.uint8`, in which case in-place operations might be surprising?
3. if the method is supposed to work with uint8 input in the first place? The docstring is not specific, and it would be useful to pin down valid input dtypes.

_Originally posted by @adeak in https://github.com/pyvista/pyvista/issues/3170#issuecomment-1369009429_

banesullivan:
this casting is no longer present so I feel it can be closed. But I think #7358 would resolve this once and for all

PR #3792, opened 2023-01-07
Fix doc builds for main branch

beroda:
This is an attempt to resolve https://github.com/pyvista/pyvista/issues/3782


banesullivan (approved):

beroda:
From what I can see in doc build logs, 4 warnings seemed to be causing the failure (since we build doc with the `-W` option).
https://github.com/pyvista/pyvista/actions/runs/3858649131/jobs/6577403817#step:11:4302

I think I was able to remove these warnings by implementing docstrings for aforementioned warnings, but I am afraid that we cannot tell the fix is working from this PR's actions, since they appear to behave differently to `main`, (might deserve and investigation/issue ?)

doc/conf.py
```diff
@@ -48,7 +50,7 @@
     category=UserWarning,
     message="Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.",
 )
-
+warnings.filterwarnings("ignore", category=PyVistaDeprecationWarning)
```
beroda, line 53: Added to filter warnings related to Pyvista examples that raise deprecation. I don't think it blocks sphinx, but "pollutes"  logs with tons of `warnings` (see https://github.com/pyvista/pyvista/actions/runs/3858649131/jobs/6577403817#step:11:455)

beroda:
Since we try to remove all warnings in doc builds, I am afraid that [this warning](
https://github.com/pyvista/pyvista/actions/runs/3858649131/jobs/6577403817#step:11:2367) would need to wait for https://github.com/pyvista/pyvista/issues/3783 to be handled.

pyvista/core/dataobject.py
```diff
@@ -252,17 +252,7 @@ def _repr_html_(self):  # pragma: no cover
...
-        ----------
-        ido : pyvista.DataObject
-            Data object to copy the metadata from.
-
-        deep : bool, optional
-            Deep or shallow copy.
-
-        """
+        """Copy pyvista meta data onto this object from another object. Intended to be overridden by subclasses."""
         pass  # called only by the inherited class
```
adeak, line 256: Shouldn't this either `raise NotImplementedError()` or be an abstract method? Or is it OK if this method is a silent no-op?

beroda: It is already an abstract class, right ?

banesullivan: Silent no op is fine for this. Especially since the new Cell class inherits it but doesn't use it. Leaving it no op prevents the need the for a no op implementation in the Cell class (and maybe other classes that don't implement? DataSet might be the only class that uses it? On mobile so not sure)

doc/conf.py
banesullivan: This makes me a little nervous... 

IMO the documentation should not use any deprecated examples. 

I know these deprecations are coming from the examples in the deprecated methods themselves...

Perhaps we should remove the examples entirely from the deprecated methods and make sure the new methods they point to have robust examples

beroda: You convinced me, I will remove examples from deprecated methods and point to the `Cell` ones.

beroda: Done

pyproject.toml
```diff
@@ -139,6 +139,7 @@ filterwarnings = [
     'ignore:.*numpy.ufunc size changed.*:RuntimeWarning',
     'ignore:.*Given trait value dtype "float64":UserWarning',
     'ignore:.*The NumPy module was reloaded*:UserWarning',
+    'error::pyvista.PyVistaDeprecationWarning'
```
beroda, line 142: I added this warning conversion to an error in oder to detect which docstrings use deprecated methods (when using `pytest --doctest-module`), but also that warnings are assessed in tests.

Let me know if it is fine for you.

banesullivan: Ahh this is great, thanks!

banesullivan (approved):

PR #3799, opened 2023-01-09
Fix documentation build main

beroda:
This is an (another and hopefully last) attempt to resolve https://github.com/pyvista/pyvista/issues/3782


MatthewFlamm:
We should merge #3800, if it works, first.  This will allow us to check that this PR indeed fixes the issue.

beroda:
@MatthewFlamm Thank you !

pyvista/plotting/plotting.py
```diff
@@ -3344,7 +3344,7 @@ def add_volume(
             Block resolution. For example ``[1, 1, 1]``. Resolution must be
             non-negative. While VTK accepts negative spacing, this results in
             unexpected behavior. See:
-            `pyvista #1967 <https://github.com/pyvista/pyvista/issues/1967>`_
+            `pyvista #1967 <https://github.com/pyvista/pyvista/issues/1967>`.
```
MatthewFlamm, line 3347: We need the underscore I think  https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages

```suggestion
            `pyvista #1967 <https://github.com/pyvista/pyvista/issues/1967>`_.
```

banesullivan (approved):
Recommend merging to fix the build on `main`.

We can figure out how to handle those warnings with pydicstyle later

MatthewFlamm (approved):
LGTM

MatthewFlamm:
> I think the `should finish with a period` warning at the end of a parameter's description (as reported [here](https://github.com/pyvista/pyvista/pull/3799/checks?check_run_id=10529362632#step:11:4235)) should be captured by `pydocstyle` and not by the full documentation build (which is performed very rarely by devs).
> 
> But unfortunately I was not able to find any error code associated with that http://www.pydocstyle.org/en/stable/error_codes.html 😢

I believe `pydocstyle` is mostly checker for pep 257, which is much less prescriptive than numpydoc.  In fact, pep 257 has examples with parameter descriptions that do not end in a period. See https://peps.python.org/pep-0257/.

For numpydoc, we could try https://numpydoc.readthedocs.io/en/latest/validation.html#docstring-validation-using-python

PR #3800, opened 2023-01-09
Allow building full docs in PR, opt-in

MatthewFlamm:
### Overview

 

Allow adding a label to build full docs.

### Details
This is opt-in from the perspective of the labelling.  PR authors without label write permissions do not have the optionality.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -73,7 +73,10 @@ jobs:
       # configure building documentation with methods on individual pages on
       # main, release, or tags with v*
       - name: Set building full docs
-        if: ${{ github.ref == 'refs/heads/main' ||  contains(github.ref, 'release') || startsWith(github.ref, 'refs/tags/v') }}
+        if: >-
+          ${{ 
+          ( github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'FullDoc') ) ||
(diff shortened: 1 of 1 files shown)


MatthewFlamm:
As of now, this PR will only work if the label pre-exists when a commit is pushed.  I will push a dummy commit to test.  To finish this, we should also run the doc build if a PR is labeled with `full-doc`.

.github/workflows/docs.yml
```diff
@@ -73,7 +73,10 @@ jobs:
       # configure building documentation with methods on individual pages on
       # main, release, or tags with v*
       - name: Set building full docs
-        if: ${{ github.ref == 'refs/heads/main' ||  contains(github.ref, 'release') || startsWith(github.ref, 'refs/tags/v') }}
+        if: >-
+          ${{ 
+          ( github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'full-doc') ) ||
+          github.ref == 'refs/heads/main' ||  contains(github.ref, 'release') || startsWith(github.ref, 'refs/tags/v') }}
```
MatthewFlamm, line 79:
```suggestion
          github.ref == 'refs/heads/main' ||  contains(github.ref, 'release') || startsWith(github.ref, 'refs/tags/v')
          }}
```

MatthewFlamm:
Doc build fails as expected here, as I labeled it before pushing the last commit.

banesullivan (approved):

PR #3803, opened 2023-01-09
pydocstyle using tomli

beroda:
### Overview
pydocstyle [release 6.2.1](http://www.pydocstyle.org/en/stable/release_notes.html#id1) (we currently use 6.2.2 in pre-commit) now uses the [`tomli`](https://pypi.org/project/tomli/) package to read the `pyproject.toml` file.

### Details
- Replacing toml by tomli in pre-commit
- Adding numpy convention to pydocstyle

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,6 +114,8 @@ upload-dir = 'doc/_build/html'
 
 [tool.pydocstyle]
 match = '(?!coverage).*.py'
+convention = "numpy"
+add-ignore = ["D404"]
 
 [tool.codespell]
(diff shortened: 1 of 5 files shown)


banesullivan (approved):

pyproject.toml
```diff
@@ -114,6 +114,8 @@ upload-dir = 'doc/_build/html'
 
 [tool.pydocstyle]
 match = '(?!coverage).*.py'
+convention = "numpy"
+add-ignore = ["D404"]
```
MatthewFlamm, line 118: "This" (sorry couldn't help myself) 
ignore looks weird to me.  You didn't add any docstrings, just reformatted whitespace. Did we need to add this because the numpy style changes the meaning to apply to sections as well as the entire docstring?  I see some sections have "This" as the first word.

A comment here would help us remember why we are ignoring it.

beroda: Indeed, `D404` is raised multiple times in the code base.

However, I found no such rule in [numpydocstyle checks ](https://numpydoc.readthedocs.io/en/latest/validation.html#), so I believe this is a choice from `pydocstyle` devs. Moreover, `D404` is ignored for both `pep257` and `google` conventions, so I don't think there is a consensus on that particular error code.

If you agree, I will add a comment in such a way.

MatthewFlamm: I didn't scroll down far enough on the pydocstyle error codes docs.  I think it is fine as is!  Thanks for the further info for me to understand.

PR #3811, opened 2023-01-11
Add hydrogen 3d orbital example

akaszynski:
Discovered a nifty dataset from https://github.com/pyvista/pyvista/discussions/3810 and through to expose it in `pyvista`:

```py

>>> mesh = examples.download_hydrogen_orbital()
>>> mesh.plot(volume=True, zoom=2)

```
tmp

### DIFF
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -4933,6 +4933,48 @@ def download_dikhololo_night():  # pragma: no cover
     return texture
 
 
+def download_hydrogen_orbital(load=True):
+    """Download a model of the electron cloud of the hydrogen atom.
+
+    Dataset originally hosted at
+    https://www.it.uu.se/edu/course/homepage/vetvis/ht11/ComputerExercises/data/a1/hydrogen.vtk
+
+    Parameters
+    ----------
+    load : bool, default: True
+        Load the dataset after downloading it when ``True``.  Set this
+        to ``False`` and only the filename will be returned.
+
+    Returns
+    -------
+    pyvista.UniformGrid or str
+        DataSet or filename depending on ``load``.
+
+    Examples
+    --------
+    Download and plot the dataset.
+
+    >>> from pyvista import examples
+    >>> mesh = examples.download_hydrogen_orbital()
+    >>> mesh.plot(volume=True, zoom=2)
+
+    Return the statistics of the dataset.
+
+    >>> grid  # doctest:+SKIP
+    UniformGrid (0x7f0e67f3c2e0)
+      N Cells:      250047
+      N Points:     262144
+      X Bounds:     0.000e+00, 6.300e+01
+      Y Bounds:     0.000e+00, 6.300e+01
+      Z Bounds:     0.000e+00, 6.300e+01
+      Dimensions:   64, 64, 64
+      Spacing:      1.000e+00, 1.000e+00, 1.000e+00
+      N Arrays:     1
+
+    """
+    return _download_and_read('hydrogen/hydrogen.vti', load=load)
+
+
 def download_cad_model_case(load=True):  # pragma: no cover
     """Download a CAD model of a Raspberry PI 4 case.
 
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -4945,6 +4945,9 @@ def download_hydrogen_orbital(load=True):  # pragma: no cover
     `Scientific Visualization - Computer exercises and assignments
     <https://www.it.uu.se/edu/course/homepage/vetvis/ht11/ComputerExercises/data/a1/hydrogen.vtk>`_
 
+    For additional information, see `pyvista/pyvista #3811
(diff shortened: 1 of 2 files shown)


pyvista/examples/downloads.py
```diff
@@ -4933,6 +4933,48 @@ def download_dikhololo_night():  # pragma: no cover
     return texture
 
 
+def download_hydrogen_orbital(load=True):  # pragma: no cover
```
adeak, line 4936: This is definitely the physicist talking in me, but can we squeeze in "3d orbital" somewhere? The ground state of hydrogen has sphericslly symmetric density. The 3d orbital is a higher-energy excited state. (You actually see similar electron orbitals on transition-metal atoms such as iron).

akaszynski: This is why we have different reviewers. My understanding of orbitals is limited to high school physics.

Fixing.

akaszynski:
```suggestion
def download_hydrogen_orbital_3d(load=True):  # pragma: no cover
```

tests/examples/test_download_files.py
```diff
@@ -925,6 +925,16 @@ def test_cad_model_case():
     assert dataset.n_points == 7677
 
 
+def test_download_hydrogen_orbital():
+    filename = examples.download_hydrogen_orbital(load=False)
+    assert os.path.isfile(filename)
+    assert filename.endswith('vti')
+
+    dataset = examples.download_hydrogen_orbital()
```
akaszynski, line 933:
```suggestion
def test_download_hydrogen_orbital_3d():
    filename = examples.download_hydrogen_orbital_3d(load=False)
    assert os.path.isfile(filename)
    assert filename.endswith('vti')

    dataset = examples.download_hydrogen_orbital_3d()
```

pyvista/examples/downloads.py
adeak: Sorry, I think I misplaced my comment. I only meant the docstring header line. The function name is fine until we ever add other orbitals.

pyvista/examples/downloads.py
```diff
@@ -4933,6 +4933,48 @@ def download_dikhololo_night():  # pragma: no cover
...
+    """Download a model of the electron cloud of the hydrogen atom.
```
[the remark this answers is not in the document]
akaszynski: @adeak, please verify. My understanding is not at the level of a jumped up high school physics student.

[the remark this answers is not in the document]
adeak: I've made one minor edit from "base state" to "ground state", otherwise this is perfect!

In the meantime I'll try to verify that this is plausibly a 3d orbital, just to be technically correct.

adeak:
Can confirm this looks like a 3d orbital:
```py
import numpy as np
import pyvista as pv
import pyvista.examples

orbital = pv.examples.download_hydrogen_orbital()
contours = orbital.contour([orbital.active_scalars.max() * 0.1], method='marching_cubes').connectivity()
colors = np.array([[0, 128, 128], [255, 215, 0], [255, 215, 0]])[contours.point_data['RegionId']]
contours.plot(scalars=colors, rgb=True, smooth_shading=True)
```
screenshot of a coloured contour of the example dataset

Cf. for instance [this page](https://chem.libretexts.org/Bookshelves/General_Chemistry/Map%3A_Chemistry_-_The_Central_Science_(Brown_et_al.)/06%3A_Electronic_Structure_of_Atoms/6.06%3A_3D_Representation_of_Orbitals):
screenshot from a random online explanatory page plotting all 5 d orbitals, one of them matching the previous plot

I may have forgotten to mention (and consider :P) that there are 5 distinct _d_ orbitals. This one belongs to angular momentum _z_ component 0, so the _x_ axis of the dataset should actually be the _z_ axis (as it often happens with these datasets).

I had to hack the colours manually, because colouring can only come from phase information, but that information is lost when you take the absolute value squared for the probability density (which is always non-negative).

adeak:
And I noticed that the probability doesn't integrate to 1. Instead the scalars span `[0, 1]`. So this is not a probability density _per se_, but it's proportional to it. It's normalised such that the maximum value is 1. (For a probability density the requirement would be that the integral of the density over the entire space is 1, which pretty much never coincides with the maximum of the density being 1.)

akaszynski:
Dataset rotated in https://github.com/pyvista/vtk-data/commit/51f3c75cce5751cf2897f977c7fc59baeb9953ee, added your example in [87be9b7](https://github.com/pyvista/pyvista/pull/3811/commits/87be9b7db81c6f97f60b79d7c4aa10b6bb912faa)

Example now renders as:
tmp

akaszynski:
> (Most of this doesn't have to be mentioned in the docstring :P)

I hope it's sufficient to add a note to this PR. I know it's not the same as adding in the notes, but I'd be remiss to not reference this discussion. Thanks for your research.

akaszynski:
Closing in favor of https://github.com/pyvista/pyvista/pull/3825

PR #3817, opened 2023-01-12
Improve silhouette decimate parameter check

banesullivan:
Ref https://github.com/pyvista/pyvista/pull/3318#discussion_r1066386611

@adeak:

> I know this was here before, but is it not weird to test for `float`ness? Shouldn't we be testing for not `None`ness instead? Or am I missing the point here? If someone passes an int by mistake, we won't raise.

I changed this to check truthy-ness

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -3794,7 +3794,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         if not isinstance(mesh, pyvista.PolyData):
             raise TypeError(f"Expected type is `PolyData` but {type(mesh)} was given.")
 
-        if isinstance(silhouette_params["decimate"], float):
+        if silhouette_params["decimate"]:
             silhouette_mesh = mesh.decimate(silhouette_params["decimate"])
         else:
             silhouette_mesh = mesh


pyvista/plotting/plotting.py
```diff
@@ -3794,7 +3794,7 @@ def add_silhouette(self, mesh, params=None):
         if not isinstance(mesh, pyvista.PolyData):
             raise TypeError(f"Expected type is `PolyData` but {type(mesh)} was given.")
 
-        if isinstance(silhouette_params["decimate"], float):
+        if silhouette_params["decimate"]:
```
adeak, line 3797: I was going to point out how this skips 0 decimation target and we should test for `None` instead (assuming that's the default) and let `decimate()` do what it wants, but I realised that 0 target is almost a no-op anyway. It will remove data arrays, but that can't be something anyone relies on.

Approving.

adeak (approved):

pyvista/plotting/plotting.py
banesullivan: Right, 0 should be a no-op, and thus the filter shouldn't be used in that scenario. As for data array removal, data arrays aren't using with the silhouette plotting

adeak: Good point!

Issue #3546, opened 2022-11-03
Inherited methods are documented too many times


banesullivan:
### Describe what maintenance you would like added.

In the documentation (most notable with the filter classes) inherited methods are documented everywhere they are inherited. This can be overwhelming to find the "right" reference of the method and I suspect the examples for these methods are getting executed and saved for each place they show up.

It's confusing to see the same filter appear in 13+ places for each of the inherited filter classes and actual data classes:

<img width="1038" alt="Screen Shot 2022-11-03 at 10 16 18 AM" src="https://user-images.githubusercontent.com/22067021/199776035-3206c48a-7e7d-486b-accd-465145142746.png">

<img width="1296" alt="Screen Shot 2022-11-03 at 10 16 58 AM" src="https://user-images.githubusercontent.com/22067021/199776098-fc0f4f37-8b2d-4420-8366-2d32292a52c4.png">

akaszynski:
Thanks for pointing this out, agreed that it's quite annoying.

MatthewFlamm:
I also agree with this.  I expect this will reduce build times for the documentation as well.

Is there a way to still list the inherited methods for each downstream class, but with a link back to the original method?  The vtk docs have a collapsed section for these for easier discovering.  I think completely removing the inherited methods will make the documentation harder to parse in other ways.

banesullivan:
I think we want to use `numpydoc_show_inherited_class_members`: https://numpydoc.readthedocs.io/en/latest/install.html#configuration

However, this might make it tough to find common attributes like `n_cells` when looking an `PolyData` as you would have to look at `DataSet`.

-----

Further this is causing examples to run and be saved for the same methods countless times. For instance, I'm pretty sure sphinx/autosummary is running the Python code for each occurrence:

- https://docs.pyvista.org/api/core/_autosummary/pyvista.PointSet.clip.html
- https://docs.pyvista.org/api/core/_autosummary/pyvista.PolyData.clip.html
- https://docs.pyvista.org/api/core/_autosummary/pyvista.UniformGrid.clip.html
- ... etc

Issue #3612, opened 2022-11-17
Verbosity control


PBrockmann:
### Describe the feature you would like to be added.

From a code that extrat edges, I get some messages coming from vtk.
vtkExtractEdges.cxx:435
-->
vtkLog(INFO, "Executing edge extractor: points are renumbered");

Is there a way to control this from pyvista ?

![image](https://user-images.githubusercontent.com/5402758/202430504-f1b187be-79b4-443f-98a3-138f40d39609.png)

banesullivan:
What version of VTK are you using? I think something changed in a more recent VTK where the logging verbosity was set to INFO-level

PBrockmann:
The latest. 9.2.2

I have warnings not errors. So
```
pv.set_warning_output_file("errors.txt") 
```
does not work.
Not very important but just would be nice.

banesullivan:
Did you do the other part with VTK to turn off VTK's output of these warnings and errors:

```py
import vtk
vtk.vtkObject.GlobalWarningDisplayOff()
```

banesullivan:
Huh. Yeah something has changed in VTK 9.2.2 where verbosity has increased and isn't being suppressed in the ways is used to be able to be. I will look into this.

banesullivan:
This is fixable by:

```py
import vtk
vtk.vtkLogger.SetStderrVerbosity(vtk.vtkLogger.VERBOSITY_OFF)
```

banesullivan:
The difference is what you are seeing from this filter is from the `vtkLogger` where as some other filters throw `vtkWarning`. `vtk.vtkLogger.SetStderrVerbosity(vtk.vtkLogger.VERBOSITY_OFF)` turns off all logging (what most people want) and `vtk.vtkObject.GlobalWarningDisplayOff()` turns off all errors and warnings

PBrockmann:
Nice.
Great thanks.

user27182:
> This is fixable by:
> 
> import vtk
> vtk.vtkLogger.SetStderrVerbosity(vtk.vtkLogger.VERBOSITY_OFF)

There is now a getter/setter/context manager for this, added in #7313

``` python
import pyvista as pv
pv.vtk_verbosity('off')
```

See docs:
https://dev.pyvista.org/api/utilities/_autosummary/pyvista.vtk_verbosity.html#pyvista.vtk_verbosity

PR #3822, opened 2023-01-12
Update VTK for docs build and limit core requirements

dcbr:
### Overview

Use the latest VTK version for test coverage and doc building workflows.

This should resolve the failing jobs in #2868. Draft for now to see if there are any failures. We might also immediately upgrade to python 3.11 for testing and doc building, as @akaszynski suggested (requires another custom built osmesa wheel)?

### CI improvements

This also improves the overall docs CI to capture and output all errors/warnings to make it easier to track down build issues moving forward (see the "Dump ... Warnings and Errors" steps)

VTK 9.2.5 is now used from a a custom OSMesa build variant hosted: https://github.com/pyvista/pyvista-wheels/blob/main/vtk_osmesa-9.2.5-cp39-cp39-linux_x86_64.whl

The docs build now ALWAYS uploaded as a CI artifact to improve debugging when errors are arising.

### Improved requirements tracking

Previously, the core requirements list in `pyproject.toml` were not version constrained for CI. This PR introduces a new `requirements.txt` where the versions of these dependencies can be controlled in CI. This will prevent issues where a core dependency update breaks CI (like NumPy did) and will give us a clear indication of the issue when it happens.

CI jobs that install pyvista should now use `pip install -e . --no-deps` and let `requirements_test.txt` or `requirements_docs.txt` handle all dependencies (note their inclusion of `-r requirements.txt`).

`--no-deps` is also needed to make sure `vtk` from PyPI is not installed when using a custom VTK build variant like `vtk_osmesa` introduced here.

### Improved Sphinx Options

The `SPHINXOPTS` were modified to dump all build warnings and errors by default. Further, developers will notice that running `make html` will now fail on warnings locally (by having the `-W` flag included by default).

### Additional fixes

- Changes to the autosummary directive to make sure VTK's `override` method is not picked up by our documentation
- Overrides to `vtkLogger` are used alongside `vtkExtractEdges` to prevent an upstream issue where INFO logs are improperly used.


dcbr:
Hmm, that failed quickly 😅 I'm not sure why there's a complaint about an X server when osmesa vtk is used?

banesullivan:
I think it's because 2 versions of VTK are being installed here: one from PyPI and the wheel we made

See https://github.com/pyvista/pyvista/actions/runs/3905390795/jobs/6672323589#step:6:251 where `vtk` is installed

We will need to be explicit about making sure the right VTK is installed

banesullivan:
I believe the reason this worked before was that the previous wheel we were using had a distribution name of `vtk` and only the wheel filename was edited to `vtk-osmesa` whereas the wheel I made the other day that is now being used has the distribution name changed to `vtk_osmesa`. Thus the `pip install ...`  sees "vtk" and installs from PyPI, not knowing that the already installed `vtk_osmesa` is what we want to use

banesullivan:
We might want to consider changing:

https://github.com/pyvista/pyvista/blob/c3c9f02dc206a3c30a1796d523ead291f5c34005/.github/workflows/docs.yml#L52

to:

```
pip install -e . --no-deps 
pip install -r requirements.txt -r requirements_docs.txt
```

And make sure the core dependencies are pinned in `requirements.txt` except for VTK. @adeak recently brought up that is was odd that our core dependencies (like `matplotlib`) are not pinned and I think this is a good chance to configure that

requirements_docs.txt
```diff
@@ -1,7 +1,7 @@
+-r requirements.txt
```
banesullivan, line 1: Hm... the core deps in `requirements.txt` are limited whereas the rest in `requirements_docs.txt` are pinned. This is better than before as an upgrade to NumPy will no longer break our docs CI, but is it what we want?

banesullivan:
Sorry to hijack this PR, @dcbr! I think what I've pushed should get this working.

After poking around with this... there's a lot more clean up I'd like to do to the Actions... I'll save that for another PR and let this focus on updating where we use the custom VTK wheel variant and adds `requirements.txt`

banesullivan:
I'm seeing CI failures for connection issue when trying to apt install. Should re-trigger tomorrow sometime

```
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/m/mesa/libgl1-mesa-glx_21.2.6-0ubuntu0.1~20.04.2_amd64.deb  Could not connect to azure.archive.ubuntu.com:80 (52.154.174.208), connection timed out [IP: 52.154.174.208 80]
```

dcbr:
No problem at all, I'm glad an expert chimed in and managed to get the build running! I naively thought it would be as simple as using the new wheel and changing the python version and then fixing any failing tests/docs :)
About that last part, it seems numpydoc is complaining a lot about docstrings for the new `override` attributes. That didn't happen during my local build earlier, so I'll see if I can reproduce and figure this out.

doc/conf.py
```diff
@@ -208,6 +208,8 @@
     r'\.MultiBlock\.remove$',
     # Enumerations
     r'\.Plot3DFunctionEnum$',
+    # VTK methods
+    r'\.override$',
```
banesullivan, line 212: I thought this would work... 😕

banesullivan:
>  it seems numpydoc is complaining a lot about docstrings for the new override attributes. That didn't happen during my local build earlier, so I'll see if I can reproduce and figure this out.

Yeah, I'm having trouble reproducing this locally as well

banesullivan:
I downloaded the built docs from CI and indeed, the `override` method from VTK's Python classes is being picked up by our docs. @pyvista/developers, does any remember how we originally implemented this to not pick up on other VTK methods? Does it look for snake case methods and this checks out? 

This method has been available in 9.2 for a bit now so I'm surprised we are only now experiencing this...

Screen Shot 2023-01-13 at 12 28 22 PM

dcbr:
> This method has been available in 9.2 for a bit now so I'm surprised we are only now experiencing this...

I suppose that's because the docs were built with VTK 9.1, which didn't have this `override` method yet? But it's strange that these warnings are not picked up by the docstringcheck (which uses the latest VTK version I believe)? I still can't reproduce it either, will try a clean build on WSL this weekend.

adeak:
> does any remember how we originally implemented this to not pick up on other VTK methods? Does it look for snake case methods and this checks out?

I didn't remember, but this is definitely sus:

https://github.com/pyvista/pyvista/blob/c0d5fc7c498504261aaf3e56c90f8c001ae65d67/doc/_templates_full/autosummary/class.rst#L26-L33

Not sure why it's only applied to attributes (rather than methods) though.

banesullivan:
Aha! Thanks for pointing that out @adeak! I think that's where it's coming from... 

I'll toy around with this and see if I can fix this specific issue and perhaps also fix https://github.com/pyvista/pyvista/issues/3546

dcbr:
So, I was able to reproduce the numpydoc complaints regarding the `override` method locally (I had to set `FULL_DOC_BUILD='true'` in my environment) and fix it in the last commit. Unfortunately I cannot reproduce the remaining warnings `WARNING: Cell printed to stderr:` at the moment...

doc/api/core/grids.rst
```diff
@@ -31,6 +31,7 @@ variable spacing in the three coordinate directions.
 
 .. jupyter-execute::
    :hide-code:
+   :stderr:
```
dcbr, line 34: Alright, I was able to bring the number of `WARNING: Cell printed to stderr:` warnings down from 18 to 12 by adding these `:stderr:` options in `jupyter-execute` blocks that (indirectly) call `extract_all_edges`. I'll see if I can find the remaining 12 cells that cause this. An alternative would be to make `vtkExtractEdges` not print to stderr, but I'm not sure that can be done? I also have no clue why this is only now causing problems...

banesullivan: I think I know of another way to do this globally

banesullivan: I think this diff will suppress the warnings:

```diff
diff --git a/pyvista/_vtk.py b/pyvista/_vtk.py
index 45bcb8d9..94efc2a1 100644
--- a/pyvista/_vtk.py
+++ b/pyvista/_vtk.py
@@ -112,6 +112,7 @@ if VTK9:
         vtkFloatArray,
         vtkIdList,
         vtkIdTypeArray,
+        vtkLogger,
         vtkLookupTable,
         vtkOutputWindow,
         vtkPoints,
@@ -629,7 +630,7 @@ else:  # pragma: no cover
     try:
         from vtk.vtkCommonKitPython import buffer_shared, vtkAbstractArray, vtkWeakReference
     except ImportError:
-        from vtk.vtkCommonCore import buffer_shared, vtkAbstractArray, vtkWeakReference
+        from vtk.vtkCommonCore import buffer_shared, vtkAbstractArray, vtkLogger, vtkWeakReference

     import vtk

diff --git a/pyvista/core/filters/data_set.py b/pyvista/core/filters/data_set.py
index f6def045..b1d36ee7 100644
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -1404,6 +1404,8 @@ class DataSetFilters:
         See :ref:`cell_centers_example` for more examples using this filter.

         """
+        # vtkExtractEdges improperly uses INFO for debugging messages
+        _vtk.vtkLogger.SetStderrVerbosity(_vtk.vtkLogger.VERBOSITY_OFF)
         alg = _vtk.vtkExtractEdges()
         alg.SetInputDataObject(self)
         if use_all_points:
```

This is indeed an upstream issue though. `vtkExtractEdges` is not supposed to use INFO logs for debug messages like this.

I tried having `SetStderrVerbosity` in the `conf.py` but I think it needs to be called in whatever context the `jupyter-execute` is happening

banesullivan: This does have the consequence of turning of logging for all other VTK classes once this is run though...

banesullivan: FYI https://github.com/pyvista/pyvista/discussions/3495#discussioncomment-4170082

banesullivan: And https://github.com/pyvista/pyvista/issues/3612

dcbr: Ok, I will let you implement that fix.
I will revert my previous 'fix', as this prints the warning in the produced docs 😅 
image

doc/conf.py
banesullivan: @dcbr, we probably don't need this, do we?

dcbr: I think this did reduce some `override` related numpydoc warnings, just not all of them?

doc/api/core/grids.rst
dcbr: I quickly applied your vtkLogger fix, let's see what we get :)

dcbr: > build succeeded.

Ah, perfect 😄 Will leave the other changes you wanted to perform to you @banesullivan.

banesullivan:
PR description has been updated to reflect all changes. Pending CI, I think this is ready for review @pyvista/developers 

Thanks for working on this @dcbr!

Please note that I changed this only to modify the `vtkLogger` for VTK>=9 (as `vtkLogger` doesn't exist in 8.1.2 and the issue was introduce in ~9.2)

banesullivan:
CI is happy! I recommend merging this asap

banesullivan (dismissed):
Giving my approval but I implemented a lot here so take this with a grain of salt

pyvista/core/filters/data_set.py
```diff
@@ -1414,7 +1414,13 @@ def extract_all_edges(self, use_all_points=False, progress_bar=False):
                     'This version of VTK does not support `use_all_points=True`. '
                     'VTK v9.1 or newer is required.'
                 )
+        if _vtk.VTK9:
+            # vtkExtractEdges improperly uses INFO for debugging messages
+            _vtk.vtkLogger.SetStderrVerbosity(_vtk.vtkLogger.VERBOSITY_OFF)
```
akaszynski, line 1419: Non blocker: has this been addressed upstream?

requirements.txt
```diff
@@ -0,0 +1,8 @@
+# Core dependencies
+# It is critical the VTK is *not* listed here
```
akaszynski, line 2: Potential non-blocker: Would be good to know why it's not listed here.

requirements_docs.txt
akaszynski: I'm fine with upper limits here. Pinned requirements means a ton of dependabot PRs, which I'm ok for non-core features which make debugging doc failures a pain to track down.

akaszynski (dismissed):
This is good to go. Local doc builds have been frustrating due to edge numbering warnings.

It would be great to investigate using Python3.11 in a follow-up PR to see if any of the claimed [performance improvements](https://docs.python.org/3/whatsnew/3.11.html#:~:text=Python%203.11%20is%20between%2010,See%20Faster%20CPython%20for%20details.) have an effect in real world builds.

pyvista/core/filters/data_set.py
banesullivan: It's a least known about... I will open a proper issue if not tracked/addressed yet

requirements.txt
banesullivan: I added a note in the PR description about this. Should I add a note to this file directly?

> needed to make sure vtk from PyPI is not installed when using a custom VTK build variant like vtk_osmesa introduced here.

akaszynski: > Should I add a note to this file directly?

Either a link to the PR or a note in the file directly. Not everyone will think to look at the blame.

banesullivan:
> It would be great to investigate using Python3.11 in a follow-up PR

Agreed. Let's wait until the "official" VTK OSMesa wheels are available (though I could make a wheel for 3.11 too)

requirements.txt
banesullivan: Done in 578343dc18baf71d66b2d79f185f4a8b305a3bc2

pyvista/core/filters/data_set.py
banesullivan: Tracked in https://gitlab.kitware.com/vtk/vtk/-/issues/18785

banesullivan (approved):

akaszynski (approved):

banesullivan:
Merging ahead of schedule... There will be a follow up PR when VTK publishes it's OSMesa wheels

PR #3824, opened 2023-01-12
Add volume mapper choice notes

banesullivan:
The `'smart'` mapper chooses one of the other available mappers based on rendering parameters and available hardware. I added some notes on how that decision is made.

I also added a waring about the `fixed_point` mapper and how it CPU-based

## Addition notes not included in docs

I noticed on my machine today that when datasets have `NaN` values, the `'fixed_point'` mapper might work best, but the `'smart'` mapper was choosing something different. I added these notes to the docstring to help others. I do not want to change the default behavior as the `'smart'` mapper usually chooses the best mapper for the given hardware and autoselecting `'fixed_point'` in the presence of `NaN` values could fail for some hardware.

I'm seeing this issue on my M2 Mac but *not* my Linux machine with an NVIDIA graphics card

```py
import pyvista as pv
from pyvista import examples
import numpy as np

volume = examples.download_knee_full()

volume['SLCImage'] = volume['SLCImage'].astype(float)
volume['SLCImage'][volume['SLCImage'] < 40] = np.nan

p = pv.Plotter()
p.add_volume(volume, cmap="coolwarm", mapper='fixed_point')
p.show()
```

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -3422,11 +3422,14 @@ class BasePlotter(PickingHelper, WidgetHelper):
             .. note::
                 The ``'smart'`` mapper chooses one of the other listed
                 mappers based on rendering parameters and available
-                hardware.
+                hardware. Most of the time the ``'smart'`` simply checks
+                if a GPU is available and if so, uses the ``'gpu'``
+                mapper, otherwise using the ``'fixed_point'`` mapper.
 
             .. warning::
-                If your volume contains NaN values, you may want to use
-                ``'fixed_point'`` mapper with some graphics hardware.
+                The ``'fixed_point'`` mapper is CPU-based and will have
+                lower performance than the ``'gpu'`` or ``'open_gl'``
+                mappers.
 
         scalar_bar_args : dict, optional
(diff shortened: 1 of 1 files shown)


pyvista/plotting/plotting.py
```diff
@@ -3419,6 +3419,15 @@ def add_volume(
...
             only ``UniformGrid`` types can be used.
 
+            .. note::
+                The ``'smart'`` mapper chooses one of the other listed
+                mappers based on rendering parameters and available
+                hardware.
+
+            .. warning::
+                If your volume contains NaN values, you may want to use
+                ``'fixed_point'`` mapper.
```
banesullivan, line 3429:
```suggestion
                ``'fixed_point'`` mapper with some graphics hardware.
```

akaszynski (dismissed):

banesullivan:
I actually changed this up a bit and removed the original NaN note. That NaN issue seems to be specific to arm-based Macs and the OpenGL implementation on that hardware. I don't really want to point people to using the `fixed_point` mapper since it's performance is quite poor (being CPU-based)

tkoyama010 (approved):
LGTM

PR #3825, opened 2023-01-13
Add hydrogen orbitals example

akaszynski:
Inspired by https://github.com/pyvista/pyvista/pull/3811 and http://staff.ustc.edu.cn/~zqj/posts/Hydrogen-Wavefunction/, this PR adds in an example evaluate the hydrogen wave function for several orbitals.

~As a bonus, it also adds `add_volume_clipper` because to my knowledge we don't current support clipping volumes and it wouldn't be an @akaszynski PR without a bonus.~ volume clipping added in #3834 

Thanks @adeak for your review on #3811. Given how the dataset seemed insufficient, I figured computing the wave function was easier than downloading a dataset. Now we can plot several orbitals.

### Example

```py
import numpy as np

import pyvista as pv
from pyvista import examples

# Create the dataset
grid = examples.load_hydrogen_orbital(3, 2, 0, norm=False)


# Plot the Probability Density of the Orbital
# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

pl = pv.Plotter(theme=pv.themes.DarkTheme())
vol = pl.add_volume(grid, opacity='linear', cmap='magma')
vol.prop.interpolation_type = 'linear'
# pl.camera_position ='yz'
pl.add_volume_clipper(vol, normal='-x', implicit=True)
pl.show_axes()
pl.show()
```
tmp


adeak:
"Anything worth doing is worth overdoing"! I'll take a look in the evening.

akaszynski:
> "Anything worth doing is worth overdoing"! I'll take a look in the evening.

Story of every last of my PRs.

pyvista/plotting/widgets.py
```diff
@@ -622,6 +622,107 @@ def callback(normal, loc):
 
         return self.add_mesh(plane_clipped_mesh, **kwargs)
 
+    def add_volume_clipper(
```
banesullivan, line 625: This name too closely follows the convention of the other methods in this module when the behavior/inputs are very different. 

IMO, this method should be changed to `add_volume_clip_plane()` and take `UniformGrid` or `RectilinearGrid` as input and plot those from here.

pyvista/plotting/widgets.py
```diff
@@ -622,6 +622,107 @@ def callback(normal, loc):
...
+        interaction_event=_vtk.vtkCommand.EndInteractionEvent,
+        origin=None,
+        **kwargs,
+    ):
+        """Clip a volume using a plane widget.
+
+        Parameters
+        ----------
+        volume : pyvista.plotting.volume.Volume
+            The input rendered volume to clip.
```
banesullivan, line 647: Woah... this is quite a bit different from the rest of the methods in this module that take an input dataset to plot and manipulate. This method takes a dataset that is already plotted. IMO, we need to make that explicitly clear.

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
 
     """
     return pyvista.read(os.path.join(dir_path, 'nut.ply'))
+
+
+def _cart2sphe(x, y, z):
```
adeak, line 417: I was about to point out that we have
https://github.com/pyvista/pyvista/blob/c0d5fc7c498504261aaf3e56c90f8c001ae65d67/pyvista/utilities/features.py#L137-L138
but apparently that function doesn't do what I'd expect it to do...

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+    """Convert 3D Cartesian coordinates to spherical coordinates.
+
+    Parameters
+    ----------
+    x, y, z : np.ndarray
+        Coordinates.
```
adeak, line 423: Will numpydoc accept this? Or are we using the fact that this won't have docs built?

pyvista/plotting/widgets.py
adeak: Can confirm; I'd have no idea what a `Volume` is.

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+    Notes
+    -----
+    This examples requires `sympy <https://www.sympy.org/>`_.
+    """
+    try:
+        from sympy import lambdify
+        from sympy.abc import phi, r, theta
+        from sympy.physics.hydrogen import Psi_nlm
+    except ImportError:
+        raise ImportError('Install sympy to run this example.') from None
```
adeak, line 450: Do we plan to test this function? If we do, do we plan to add sympy to CI for this?

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+
+    if n < 1 or n > 5:
+        raise ValueError('`n` must be between 1 and 5')
+    if l not in range(n):
+        raise ValueError(f'`l` must be one of: {list(range(n))}')
+    if m not in range(l + 1):
+        raise ValueError(f'`l` must be one of: {list(range(l+1))}')
```
adeak, line 457: For a given `l` we have `2*l + 1` orbitals, for `m in range(-l, l + 1)`. (Fun fact: this is also true for half-integer angular momentum which can happen with spin: for a half-spin particle such as the electron we have `s = 1/2`, so `m_s in [-1/2, 1/2]` assumes `2*s + 1 == 2` potential values (up and down spin).
```suggestion
    if m not in range(-l, l + 1):
        raise ValueError(f'`m` must be one of: {list(range(-l, l+1))}')
```

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+    xy2 = x**2 + y**2
+    r = np.sqrt(xy2 + z**2)
+    theta = np.arctan2(np.sqrt(xy2), z)  # the polar angle in radian angles
+    phi = np.arctan2(y, x)  # the azimuth angle in radian angles
+    phi[phi < 0] += np.pi * 2  # np.arctan2 returns the angle in the range [-pi, pi]
+
+    return r, theta, phi
+
+
+def load_hydrogen_orbital(n=1, l=0, m=0, norm=True):
```
adeak, line 435: `norm` seems unused.

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+    """Load the hydrogen wave function for a :class:`pyvista.UniformGrid`.
+
+    Ispired by `Hydrogen Wave Function
+    <http://staff.ustc.edu.cn/~zqj/posts/Hydrogen-Wavefunction/>`_
+
+    Notes
+    -----
+    This examples requires `sympy <https://www.sympy.org/>`_.
```
adeak, line 443:
```suggestion
    This example requires `sympy <https://www.sympy.org/>`_.
```

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+    phi[phi < 0] += np.pi * 2  # np.arctan2 returns the angle in the range [-pi, pi]
+
+    return r, theta, phi
+
+
+def load_hydrogen_orbital(n=1, l=0, m=0, norm=True):
+    """Load the hydrogen wave function for a :class:`pyvista.UniformGrid`.
+
+    Ispired by `Hydrogen Wave Function
+    <http://staff.ustc.edu.cn/~zqj/posts/Hydrogen-Wavefunction/>`_
```
adeak, line 439:
```suggestion
    <http://staff.ustc.edu.cn/~zqj/posts/Hydrogen-Wavefunction/>`_.
    
    Parameters
    ----------
    n : int, default: 1
        Principal quantum number. Must be a positive integer.
    
    l : int, default: 0
        Azimuthal quantum number. Must be a non-negative integer strictly smaller than ``n``.
    
    m : int, default: 0
        Magnetic quantum number. Must be an integer ranging from ``-l`` to ``l`` (inclusive).
```
(I haven't checked line lengths, feel free to edit.)

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+    """
+    try:
+        from sympy import lambdify
+        from sympy.abc import phi, r, theta
+        from sympy.physics.hydrogen import Psi_nlm
+    except ImportError:
+        raise ImportError('Install sympy to run this example.') from None
+
+    if n < 1 or n > 5:
+        raise ValueError('`n` must be between 1 and 5')
```
adeak, line 453: Why the upper limit of 5? Sampling heuristics? We should probably also mention this in the docstring.

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+    if l not in range(n):
+        raise ValueError(f'`l` must be one of: {list(range(n))}')
+    if m not in range(l + 1):
+        raise ValueError(f'`l` must be one of: {list(range(l+1))}')
+
+    psi = lambdify((r, phi, theta), Psi_nlm(n, l, m, r, phi, theta), 'numpy')
```
adeak, line 459: We could use strings instead of sympy symbols on this line, but that's probably not worth it.

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+    if n == 1:
+        l, n = 1.5 * n**2 + 1.0, 200
+    else:
+        l, n = 1.5 * n**2 + 10.0, 200
+
+    n = 100
+    s = (l * 2) / (n - 1)
+    grid = pyvista.UniformGrid(dimensions=(n, n, n), spacing=(s, s, s), origin=(-l, -l, -l))
+
+    r0, theta, phi = _cart2sphe(grid.x, grid.y, grid.z)
```
adeak, line 470: We're also shadowing the earlier `theta, phi` symbols here, but this is less of an issue. Might as well double down and rename `r0` to `r`.

adeak (commented):
Only looking at the physics part now.

pyvista/plotting/widgets.py
adeak: Can confirm; I'd have no idea what a `Volume` is.

examples/03-widgets/clip-volume.py
```diff
@@ -0,0 +1,45 @@
+"""
+.. _clip_volume_widget_example:
+
+Clip Volume Widget
+~~~~~~~~~~~~~~~~~~
+If you have a :class:`pyvista.UniformGrid`, you can clip it using the
+:func:`pyvista.Plotter.add_volume_clipper` widget.
+
+
+.. image:: ../../images/gifs/box-clip.gif
```
adeak, line 10: Is this gif (`add_mesh_clip_box` on PolyData Nefertiti) relevant for this example?

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+    ----------
+    x, y, z : np.ndarray
+        Coordinates.
+
+    """
+    xy2 = x**2 + y**2
+    r = np.sqrt(xy2 + z**2)
+    theta = np.arctan2(np.sqrt(xy2), z)  # the polar angle in radian angles
+    phi = np.arctan2(y, x)  # the azimuth angle in radian angles
+    phi[phi < 0] += np.pi * 2  # np.arctan2 returns the angle in the range [-pi, pi]
```
adeak, line 430: I would expect this step to be unnecessary, and running the new example with this removed seems to look the same.
```suggestion
```

pyvista/examples/examples.py
```diff
@@ -412,3 +412,65 @@ def load_nut():
...
+
+    return r, theta, phi
+
+
+def load_hydrogen_orbital(n=1, l=0, m=0, norm=True):
+    """Load the hydrogen wave function for a :class:`pyvista.UniformGrid`.
+
+    Ispired by `Hydrogen Wave Function
```
adeak, line 438:
```suggestion
    Inspired by `Hydrogen Wave Function
```

examples/03-widgets/clip-volume.py
akaszynski: It's not. Sorry for the weak PR. It should have been labeled "draft".

pyvista/examples/examples.py
akaszynski: Upper limit is unnecessary, I was merely duplicating the example.

Seems that it goes out to the 7th shell. I'm all for removing the limit and letting people play around with it.

As for the sampling, it indeed becomes an issue for some of the higher, "non-realistic" orbitals.

pyvista/plotting/widgets.py
akaszynski: How about both? You can either add a clip plane on an existing `volume.Volume` actor, or add a dataset.

Advantage of this approach is you can seamlessly add multiple clip planes to a single actor. For example:

```py
pl = pv.Plotter(theme=pv.themes.DarkTheme())
vol = pl.add_volume(grid, opacity='linear', cmap='magma')
pl.add_volume_clipper(vol, normal='-x', implicit=True)
pl.add_volume_clipper(vol, normal='-y', implicit=True)
pl.show_axes()
pl.show()
```

tmp

pyvista/plotting/widgets.py
akaszynski: Just added this in #3675 as it's quite helpful to wrap [vtkVolume](https://vtk.org/doc/nightly/html/classvtkVolume.html). This really should be called a "vtkVolumeActor". Not sure why it's named this way

I'm of the opinion that we should accept both, as well as expanding our existing widgets to work with `pyvista.DataSet`s as well as `pyvista.Actor`s.

pyvista/examples/examples.py
akaszynski:
```suggestion
def load_hydrogen_orbital(n=1, l=0, m=0):
```

akaszynski:
Splitting off two PRs and will work on this again once we've addressed:
- Volumetric RGBA plotting
- Volume clip widget

pyvista/examples/examples.py
akaszynski: Moving this to utilities where it should be...

akaszynski:
Took longer than expected, but ready for review!
image

akaszynski:
@adeak, would appreciate a quick review as I'm not a physicist.

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,224 @@
...
+    )
+    vol.prop.interpolation_type = 'linear'
+    if clip_plane:
+        pl.add_volume_clip_plane(
+            vol, normal=clip_plane, normal_rotation=False, widget_color=pv.Color(opacity=0.0)
+        )
+    pl.camera_position = cpos
+    pl.camera.zoom(1.5)
+    pl.show_axes()
+    pl.show()
```
banesullivan, line 156:
```suggestion
    return pl.show()
```

😉

pyvista/utilities/features.py
```diff
@@ -172,6 +172,42 @@ def transform_vectors_sph_to_cart(theta, phi, r, u, v, w):
     return u_t, v_t, w_t
 
 
+def cart_to_sphe(x, y, z):
```
banesullivan, line 175: I don't love this name... I'd prefer a more verbose `cartesian_to_spherical`

banesullivan (approved):
Looks really really nice, only minor feedback

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,224 @@
...
+    pl = pv.Plotter()
+    vol = pl.add_volume(
+        orbital,
+        scalars='plot_scalars',
+    )
+    vol.prop.interpolation_type = 'linear'
+    if clip_plane:
+        pl.add_volume_clip_plane(
+            vol, normal=clip_plane, normal_rotation=False, widget_color=pv.Color(opacity=0.0)
+        )
```
banesullivan, line 152: It's kind of tough to use this without a bounding box. I was playing with it and quickly lost my volume

akaszynski: Added bounding box back in 9811bbf.

pyvista/utilities/features.py
akaszynski: Changed to `cartesian_to_spherical` 9811bbf.

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+
+###############################################################################
+# Plot the Orbital Contours as an Isosurface
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Generate the contour plot for the orbital by determining when the orbital
+# equals 10% the maximum value of the orbital. This effectively captures the
+# majority of the "volume" the electron potentially exists in.
+#
+# Note how we use the absolute value of the scalars when evaluating
+# :func:`contour() <pyvista.PolyDataFilters.contour>`. Otherwise, we would only
```
adeak, line 93: Another broken intersphinx link.

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+    scalars=np.abs(grid['real_hwf']),
+    method='marching_cubes',
+)
+contours = contours.interpolate(grid)
+contours.plot(
+    categories=2,
+    smooth_shading=True,
+    annotations={-eval_at: 'down-spin', eval_at: 'up-spin'},
+    scalar_bar_args={'n_labels': 0, 'title': ''},
+)
```
adeak, line 108: Now this is a conceptual bug: there's no spin here. On each of these orbitals (so e.g. on the 3d_{xy} orbital) you could place 2 electrons without violating Pauli's exclusion principle: an up-spin state and a down-spin state are different quantum state so they can coexist. If you've seen orbital energy diagrams in chemistry class (I know I have, but certainly many people don't):

electron configuration of Br

This is the electronic structure of atomic bromine. In the top right you can see 5 horizontal lines above "3d", these are the 5 3d orbitals (ranging from m = -2 to 2). On each of these orbitals you have an up and a down arrow: two electrons!

So my point is that a single orbital (for a single electron) occupies all four lobes of the wave function. What is plotted is the phase of the wave function, you can think of it as a sign (as in +-1). In some places it's negative, in some places it's positive. But this information is lost when you take the square of the wave function in order to obtain a probability density (which is non-negative).

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+    [eval_at],
+    scalars=np.abs(grid['real_hwf']),
+    method='marching_cubes',
+)
+contours = contours.interpolate(grid)
+contours.plot(
+    categories=2,
+    smooth_shading=True,
+    annotations={-eval_at: 'down-spin', eval_at: 'up-spin'},
+    scalar_bar_args={'n_labels': 0, 'title': ''},
```
adeak, line 107: Scalar bar looks wonky:

Screenshot from 2023-01-27 00-08-45

Wouldn't you want a categorical with 2 colours?

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+)
+
+
+###############################################################################
+# Volumetric Plot: Plot the Orbitals using RGBA
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Let's now combine some of the best parts of the two above plots. The
+# volumetric plot is great for showing the probability of the "electron cloud"
+# orbitals, but the colormap doesn't quite match reality as well as the
+# isosurface plot.
```
adeak, line 117: [citation needed] :smile:

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+#
+# For this example we're going to use an RGBA colormap to tightly control way
+# the orbitals are plotted. For this, the opacity will be mapped to the
```
adeak, line 120:
```suggestion
# For this example we're going to use an RGBA colormap to tightly control the
# way the orbitals are plotted. For this, the opacity will be mapped to the
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+# probability of the electron being at a location in the grid, which we can do
+# by taking the absolute value of the orbital's "hydrogen wave function". We
```
adeak, line 122: Absolute value _squared_, but it looks like that's indeed what you're computing here (not suggesting an edit because wrapping would need shifting around).

I'd also omit "hydrogen" and the quotes, you really are looking at the (hydrogen) orbital's wave function. Of course you might mean that "hydrogen wave function" as in `real_hwf`, in which case I'd also recommend renaming `real_hwf`...

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+# can set the color of the orbital based on the phase, which we can get simply
+# by with ``orbital['real_hwf'] < 0``.
+#
+# Let's start with a simple one, the 3pz orbital.
```
adeak, line 126: If we can inline latex this then this should mention `$3p_z$`.

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+prob = np.abs(hydro_orbital['real_hwf']) ** 2
+prob /= prob.sum()
+indices = np.random.choice(hydro_orbital.n_points, 10000, p=prob)
+
+# add a small amount of noise to these coordinates to remove the "grid like"
+# structure present in the underlying UniformGrid
+points = hydro_orbital.points[indices]
+points += np.random.random(points.shape) - 0.5
+
+# Create a point cloud and add the phase of as the active scalars
```
adeak, line 202:
```suggestion
# Create a point cloud and add the phase as the active scalars
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+    smooth_shading=True,
+    show_scalar_bar=False,
+    cmap=['red', 'green'],
+    ambient=0.2,
+)
+pl.enable_ssao(radius=10)
+pl.enable_anti_aliasing()
+pl.camera.zoom(2)
+pl.background_color = 'w'
+pl.show()
```
adeak, line 226: I'm slightly curious how a "Gaussian points" representation of the sampled data would look, but I think the example is already extensive as it is (and the result might be uninteresting anyway).

pyvista/examples/examples.py
```diff
@@ -412,3 +412,95 @@ def load_nut():
 
     """
     return pyvista.read(os.path.join(dir_path, 'nut.ply'))
+
+
+def load_hydrogen_orbital(n=1, l=0, m=0, zoom_fac=1.0):
+    """Load the hydrogen wave function for a :class:`pyvista.UniformGrid`.
+
+    This is the Schrödinger equation evaluated in three dimensions.
```
adeak, line 420:
```suggestion
    This is the solution to the Schrödinger equation for hydrogen
    evaluated in three-dimensional Cartesian space.
```
The "Cartesian space" part is not strictly necessarym but the solution of the equation involves polar coordinates so it might make sense to emphasize that the final result here is in "real space".

pyvista/examples/examples.py
```diff
@@ -412,3 +412,95 @@ def load_nut():
...
+
+    Inspired by `Hydrogen Wave Function
+    <http://staff.ustc.edu.cn/~zqj/posts/Hydrogen-Wavefunction/>`_.
+
+    Parameters
+    ----------
+    n : int, default: 1
+        Principal quantum number. Must be a positive integer. This is often
+        referred to the "energy level" or "shell".
```
adeak, line 429:
```suggestion
        referred to as the "energy level" or "shell".
```

pyvista/examples/examples.py
```diff
@@ -412,3 +412,95 @@ def load_nut():
...
+        Zoom factor for the electron cloud. Increase this value to focus on the
+        center of the electron cloud.
+
+    Returns
+    -------
+    pyvista.UniformGrid
+        UniformGrid containing two ``point_data`` arrays:
+
+        * ``'real_hwf'`` - Real component of the quantum hydrogen wave function.
+        * ``'hwf'`` - Complex quantum hydrogen wave function.
```
adeak, line 451: If it's not too much of a request:
```suggestion
        * ``'real_wf'`` - Real part of the wave function.
        * ``'wf'`` - Complex wave function.
```
I'd drop both "quantum" and "hydrogen", both are strongly implied. I'd also drop the "h" from the names. Might even consider renaming `real_wf` to `wf_real`? Not sure.

pyvista/examples/examples.py
```diff
@@ -412,3 +412,95 @@ def load_nut():
...
+    try:
+        from sympy import lambdify
+        from sympy.abc import phi, r, theta
+        from sympy.physics.hydrogen import Psi_nlm
+    except ImportError:  # pragma: no cover
+        raise ImportError(
+            '\n\nInstall sympy to run this example. Run:\n\n    pip install sympy\n'
+        ) from None
+
+    if l not in range(n):
```
adeak, line 479: Might be worth checking `n` as well.

doc/index.rst
```diff
@@ -166,6 +166,17 @@ Subtract a sphere from a cube mesh.
     boolean.plot(color='darkgrey', smooth_shading=True, split_sharp_edges=True)
 
 
+Plot Volumetric Data
+~~~~~~~~~~~~~~~~~~~~
+Plot the 3dxy orbital of a hydrogen atom.
+
+.. jupyter-execute::
+
+    from pyvista import examples
```
adeak, line 175: Since this is the front page, should we also note here that this needs `sympy`?

pyvista/plotting/plotting.py
```diff
@@ -3715,6 +3721,15 @@ def add_volume(
         elif not isinstance(scalars, np.ndarray):
             scalars = np.asarray(scalars)
 
+        if scalars.ndim != 1:
+            if scalars.ndim != 2:
+                raise ValueError('`add_volume` only supports scalars with 1 or 2 dimensions')
+            if scalars.shape[1] != 4 or scalars.dtype != np.uint8:
+                raise ValueError(
```
adeak, line 3728: I'm half wondering if these should be TypeErrors. No time to look up precedents in the code base right now.

pyvista/plotting/plotting.py
```diff
@@ -3715,6 +3721,15 @@ def add_volume(
...
+                    '`add_volume` only supports scalars with 2 dimensions that have 4 components of datatype np.uint8.\n\n'
+                    f'Scalars have shape ({scalars.shape}) and ({scalars.dtype})'
```
adeak, line 3730:
```suggestion
                    f'Scalars have shape {scalars.shape} and dtype {scalars.dtype.__name__!r}.'
```

pyvista/utilities/features.py
```diff
@@ -172,6 +172,42 @@ def transform_vectors_sph_to_cart(theta, phi, r, u, v, w):
...
+    x, y, z : numpy.ndarray
+        Cartesian coordinates.
+
+    Returns
+    -------
+    r : numpy.ndarray
+        Radial distance.
+
+    theta : numpy.ndarray
+        Angle (radians) with respect to the polar axis.
```
adeak, line 189:
```suggestion
        Angle (radians) with respect to the polar axis. Also known
        as polar angle.
```

pyvista/utilities/features.py
```diff
@@ -172,6 +172,42 @@ def transform_vectors_sph_to_cart(theta, phi, r, u, v, w):
...
+
+    phi : numpy.ndarray
+        Angle (radians) of rotation from the initial meridian plane.
```
adeak, line 192:
```suggestion
        Angle (radians) of rotation from the initial meridian plane.
        Also known as azimuthal angle.
```

doc/index.rst
```diff
@@ -166,6 +166,17 @@ Subtract a sphere from a cube mesh.
     boolean.plot(color='darkgrey', smooth_shading=True, split_sharp_edges=True)
 
 
+Plot Volumetric Data
+~~~~~~~~~~~~~~~~~~~~
+Plot the 3dxy orbital of a hydrogen atom.
```
akaszynski, line 171:
```suggestion
Plot the 3dxy orbital of a hydrogen atom.

.. note::
   This example requires `sympy <https://www.sympy.org/>`_.
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+#        }
+#        e^{-r / na_0}
+#        \left(\frac{2r}{na_0}\right)^l
+#        L_{n-l-1}^{2l+1} \cdot Y_l^m(\theta, \phi)
+#    \end{equation}
+#
+# See `Hydrogen atom <https://en.wikipedia.org/wiki/Hydrogen_atom>`_ for more
+# details.
+#
+# This dataset evaluates this function for the hydrogen orbital 3dxy, with the
```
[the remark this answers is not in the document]
akaszynski:
```suggestion
# This dataset evaluates this function for the hydrogen orbital :math:`$3d_{xy}$`, with the
```
```suggestion
# This dataset evaluates this function for the hydrogen orbital 3dxy, with the
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+# * Magnetic quantum number: ``m=-2``
+
+grid = examples.load_hydrogen_orbital(3, 2, -2)
+grid
+
+
+###############################################################################
+# Plot the Orbital
+# ~~~~~~~~~~~~~~~~
+# Plot the orbital using :func:`add_volume() <pyvista.Plotter.add_volume>` and
```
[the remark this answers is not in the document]
akaszynski: This is an issue with the lean build. I think there's a way to get the normal build to work fast enough so we don't have to rely on two builds...

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,226 @@
...
+#    between ``[-<value>, <value>]``, we cannot use the default opacity
+#    ``opacity='linear'``. Instead, we use ``[1, 0, 1]`` since we would like
+#    the opacity to be proportional to the absolute value of the scalars.
+
+pl = pv.Plotter()
+vol = pl.add_volume(grid, cmap='magma', opacity=[1, 0, 1])
+vol.prop.interpolation_type = 'linear'
+pl.camera.zoom(1.5)
+pl.show_axes()
+pl.show()
```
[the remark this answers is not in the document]
akaszynski: Rendering issue because the black and white parts are deep within the "clouds".

examples/99-advanced/atomic-orbitals.py
akaszynski: I'm using the full docs for verifying this and it works.

pyvista/plotting/plotting.py
akaszynski: We typically use `ValueErrors` for numpy types.

examples/99-advanced/atomic-orbitals.py
akaszynski: Removing the scalar bar altogether.

examples/99-advanced/atomic-orbitals.py
akaszynski: Good idea. Looks great!
tmp

examples/99-advanced/atomic-orbitals.py
akaszynski: Thanks for the explanation!

adeak: My pleasure, but as I said this is a "bug" in the example which we should fix, spin should not be mentioned at all. Unresolving this.

adeak: Ah, sorry, github fooled me by only showing some of the commits. Spin seems gone now, thanks.

doc/index.rst
```diff
@@ -166,6 +166,20 @@ Subtract a sphere from a cube mesh.
...
```
adeak, line 171: Would this also work in an .rst file?
```suggestion
Plot the :math:`3d_{xy}` orbital of a hydrogen atom.
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,240 @@
...
+# Let's now combine some of the best parts of the two above plots. The
+# volumetric plot is great for showing the probability of the "electron cloud"
+# orbitals, but the colormap doesn't quite match reality as well as the
+# isosurface plot.
+#
+# For this example we're going to use an RGBA colormap to tightly control the
+# way the orbitals are plotted. For this, the opacity will be mapped to the
+# probability of the electron being at a location in the grid, which we can do
+# by taking the absolute value squared of the orbital's wave function. We can
+# set the color of the orbital based on the phase, which we can get simply by
```
adeak, line 120:
```suggestion
# set the color of the orbital based on the phase, which we can get simply
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,240 @@
...
+    pl.show_axes()
+    return pl.show()
+
+
+hydro_orbital = examples.load_hydrogen_orbital(3, 1, 0)
+plot_orbital(hydro_orbital, clip_plane='-x')
+
+
+###############################################################################
+# Volumetric Plot: 4dz2 orbital
```
adeak, line 163:
```suggestion
# Volumetric Plot: :math:`4d_{z^2}` orbital
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,240 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+hydro_orbital = examples.load_hydrogen_orbital(4, 2, 0)
+plot_orbital(hydro_orbital, clip_plane='-y')
+
+
+###############################################################################
+# Volumetric Plot: 4dxz orbital
```
adeak, line 170:
```suggestion
# Volumetric Plot: :math:`4d_{xz}` orbital
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,240 @@
...
+
+###############################################################################
+# Plot an Orbital Using a Density Plot
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# We can also plot atomic orbitals using a 3D density plot. For this, we will
+# use :func:`numpy.random.choice` to sample all the points of our
+# :class:`pyvista.UniformGrid` based on the probability of the electron being
+# at that coordinate.
+#
+# Note how we square the probability to better visualize the distribution of
```
adeak, line 184: I don't think we square the probability: looks like we're squaring the wave function, which is itself the probability (density). This sentence might be unnecessary anyway considering the "absolute value squared" explanation in the previous section.

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,240 @@
...
+pl.enable_anti_aliasing()
+pl.camera.zoom(2)
+pl.background_color = 'w'
+pl.show()
+
+
+###############################################################################
+# Density Plot - Gaussian Points Representation
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Finally, let's plot the same data using the "gaussian points" representation.
```
adeak, line 229:
```suggestion
# Finally, let's plot the same data using the "Gaussian points" representation.
```

adeak (commented):
This is looking great! Some final nitpicks, nothing critical.

pyvista/plotting/plotting.py
```diff
@@ -3879,6 +3885,15 @@ def add_volume(
         elif not isinstance(scalars, np.ndarray):
             scalars = np.asarray(scalars)
 
+        if scalars.ndim != 1:
+            if scalars.ndim != 2:
+                raise ValueError('`add_volume` only supports scalars with 1 or 2 dimensions')
+            if scalars.shape[1] != 4 or scalars.dtype != np.uint8:
+                raise ValueError(
+                    '`add_volume` only supports scalars with 2 dimensions that have 4 components of datatype np.uint8.\n\n'
+                    f'Scalars have shape {scalars.shape} and dtype {scalars.dtype.__name__!r}.'
```
adeak, line 3894:
```suggestion
                    f'Scalars have shape {scalars.shape} and dtype {scalars.dtype.name!r}.'
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,240 @@
...
+    return pl.show()
+
+
+hydro_orbital = examples.load_hydrogen_orbital(3, 1, 0)
+plot_orbital(hydro_orbital, clip_plane='-x')
+
+
+###############################################################################
+# Volumetric Plot: :math:`4d_{z^2}` orbital
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
adeak, line 164:
```suggestion
# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

examples/99-advanced/atomic-orbitals.py
```diff
@@ -0,0 +1,240 @@
...
+hydro_orbital = examples.load_hydrogen_orbital(4, 2, 0)
+plot_orbital(hydro_orbital, clip_plane='-y')
+
+
+###############################################################################
+# Volumetric Plot: :math:`4d_{xz}` orbital
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
adeak, line 171:
```suggestion
# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

akaszynski:
> This is looking great! Some final nitpicks, nothing critical.

Thanks again for all your reviews. We'll squeeze this in for the release.

examples/99-advanced/atomic-orbitals.py
akaszynski: Fixed in [7417ab3](https://github.com/pyvista/pyvista/pull/3825/commits/7417ab3e4ea2f340133b3e3716d08cfc99ff543e)

adeak (approved):
Thanks for the great work @akaszynski!

Approving, pending a final check that the built docs have proper math even in section headings (that I added). If you check this yourself and things look OK then this is good to merge as far as I'm concerned.

akaszynski:
> Thanks for the great work @akaszynski!
> 
> Approving, pending a final check that the built docs have proper math even in section headings (that I added). If you check this yourself and things look OK then this is good to merge as far as I'm concerned.

tmp

PR #3830, opened 2023-01-16
add in rgba volume plotting

akaszynski:
VTK supports direct mapping of color values from scalars. See [vtkVolumeProperty.SetIndependentComponents](https://vtk.org/doc/nightly/html/classvtkVolumeProperty.html#a2534a6dc500be42c9a842d405c4ee0bb). This has been added to the ``plot_volume`` API and such that it automatically picks up on plotting RGBA values and enabling independent components when the scalars have four components and ``dtype==np.uint8``.

```py
import pyvista as pv
import numpy as np
grid = pv.UniformGrid(dimensions=(5, 20, 20))
scalars = grid.points - (grid.origin)
scalars /= scalars.max()
opacity = np.linalg.norm(grid.points - grid.center, axis=1).reshape(-1, 1)
opacity /= opacity.max()
scalars = np.hstack((scalars, opacity**3))
scalars *= 255
pl = pv.Plotter()
vol = pl.add_volume(grid, scalars=scalars.astype(np.uint8))
vol.prop.interpolation_type = 'linear'
pl.show()
```
tmp


pyvista/plotting/mapper.py
```diff
@@ -270,7 +270,7 @@ def scalar_map_mode(self) -> str:
     def scalar_map_mode(self, scalar_mode: Union[str, FieldAssociation]):
         if isinstance(scalar_mode, FieldAssociation):
             scalar_mode = scalar_mode.name
-        scalar_mode = scalar_mode.lower()
+        scalar_mode = scalar_mode.lower()  # type: ignore
```
banesullivan, line 273: Why???!!! 

See https://github.com/pyvista/pyvista/pull/3318#discussion_r1067534921

tests/plotting/test_plotting.py
```diff
@@ -2494,6 +2494,7 @@ def test_add_text():
     not vtk.vtkMathTextFreeTypeTextRenderer().MathTextIsSupported(),
     reason='Math text is not supported.',
 )
+@pytest.mark.needs_vtk_version(9, 2, 2)
```
banesullivan, line 2497: This does not require 9.2.2... why are you changing this here?

pyvista/plotting/volume_property.py
```diff
@@ -231,6 +231,32 @@ def shade(self) -> bool:
...
+        first color transfer function and the second component is passed
+        through the first scalar opacity (and gradient opacity) transfer
+        function. Normals will be generated off of the second component. When
+        using gradient based opacity modulation, the gradients are computed off
+        of the second component.
+
+        For 4 component data, the first three will directly represent RGB (no
+        lookup table). The fourth component will be passed through the first
+        scalar opacity transfer function for opacity and first gradient opacity
+        transfer function for gradient based opacity modulation. Normals will
```
banesullivan, line 249: This needs to be clearly stated in `add_volume()` as well. The `opacity` argument still has a strong affect on the visualization of RGBA volumes which IMO is a bit unexpected.

banesullivan (changes requested):
This PR stills needs a bit of work... mainly concerned about some unrelated changes you are sneaking in here that IMO, should not be added.

Further, I don't think RGBA volume plotting is fully implemented here... for example, the opacity transfer function still has a strong impact (seeminly 100% of the impact) on the transparency of the rendered volume where the Alpha component is seeminly ignored?... I'm not sure... but take this example:

```py
import pyvista as pv
import numpy as np
grid = pv.UniformGrid(dimensions=(5, 20, 20))
scalars = grid.points - (grid.origin)
scalars /= scalars.max()
opacity = np.linalg.norm(grid.points - grid.center, axis=1).reshape(-1, 1)
opacity /= opacity.max()
scalars = np.hstack((scalars, opacity**3))
scalars *= 255
```

```py
pl = pv.Plotter(notebook=0)
vol = pl.add_volume(
    grid, 
    scalars=scalars.astype(np.uint8), 
    opacity=[1, 1]
)
vol.prop.interpolation_type = 'linear'
pl.show()
```
Screen Shot 2023-01-17 at 9 55 24 AM

or 

```py
pl = pv.Plotter(notebook=0)
vol = pl.add_volume(
    grid, 
    scalars=scalars.astype(np.uint8), 
    opacity='sigmoid',
)
vol.prop.interpolation_type = 'linear'
pl.show()
```
Screen Shot 2023-01-17 at 9 55 36 AM


These results are a bit unexpected to me as I expect the Alpha component to fully drive the opacity mapping in the volume rendering. Maybe this just needs further clarification in the docs. 

My concerns here are further emphasized by the factor that our default handling of this RGBA volume is different from ParaViews:

Screen Shot 2023-01-17 at 9 58 02 AM


vs our default for this volume:

Screen Shot 2023-01-17 at 9 59 22 AM

pyvista/plotting/mapper.py
akaszynski: Turns out this is a Python 3.11 issue:
```
$ python3.11 -m venv .venv
$ source .venv/bin/activate
$ pip install pre-commit
$ pre-commit run --all-files mypy
```

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

pyvista/plotting/mapper.py:273: error: Item "FieldAssociation" of "Union[str, FieldAs
sociation]" has no attribute "lower"  [union-attr]                                  
pyvista/core/composite.py:589: note: By default the bodies of untyped functions are n
ot checked, consider using --check-untyped-defs  [annotation-unchecked]             
pyvista/core/composite.py:590: note: By default the bodies of untyped functions are n
ot checked, consider using --check-untyped-defs  [annotation-unchecked]             
Found 1 error in 1 file (checked 89 source files)
```

Either we add in `type: ignore`, or I ignore the type failure on Python 3.11 until this is resolved upstream. I agree that it appears to be a red herring, but this also means that anyone else developing on Python 3.11 will encounter this weird type error as well.

tests/plotting/test_plotting.py
akaszynski: https://github.com/pyvista/pyvista/blob/1f78d08fa1a9e254a93b719d95d98030fba7118b/tests/plotting/test_plotting.py#L2501

Yet we don't specify a conditional `matplotlib` requirement in `requirements_test.txt`:
https://github.com/pyvista/pyvista/blob/1f78d08fa1a9e254a93b719d95d98030fba7118b/requirements_test.txt#L13

Fails with the following environment:
```
$ python3.8 -m venv .venv
$ source .venv/bin/activate
$ pip install pip wheel -U
$ pip install -e .
$ pip install -r requirements_test.txt 
$ pip install vtk==9.1.0
$ pytest tests/plotting/ -k latex
```

```
tests/plotting/test_plotting.py F                                             [100%]

===================================== FAILURES ======================================
________________________________ test_add_text_latex ________________________________

    @pytest.mark.skipif(
        not vtk.vtkMathTextFreeTypeTextRenderer().MathTextIsSupported(),
        reason='Math text is not supported.',
    )
    def test_add_text_latex():
        """Test LaTeX symbols.
    
        For VTK<=9.2.2, this requires matplotlib<3.6
        """
        plotter = pyvista.Plotter()
        plotter.add_text(r'$\rho$', position='upper_left', font_size=150, color='blu)
>       plotter.show()

tests/plotting/test_plotting.py:2504: 
pyvista/plotting/plotting.py:6413: in show
    self.close()
pyvista/plotting/plotting.py:4200: in close
    self._before_close_callback(self)

self = <pytest_pyvista.pytest_pyvista.VerifyImageCache object at 0x7f4584573eb0>
plotter = <pyvista.plotting.plotting.Plotter object at 0x7f4584634430>

    def __call__(self, plotter):
        """Either store or validate an image.
    
        Parameters
        ----------
        plotter : pyvista.Plotter
            The Plotter object that is being closed.
    
        """
        if self.skip:
            return
    
        if not VTK9:
            if self.skip_image_cache_vtk8:
                return
            else:
                raise RuntimeError("Image cache is only valid for VTK9+")
    
        if self.ignore_image_cache:
            return
    
        if self.n_calls > 0:
            test_name = f"{self.test_name}_{self.n_calls}"
        else:
            test_name = self.test_name
        self.n_calls += 1
    
        if self.high_variance_test:
            allowed_error = self.var_error_value
            allowed_warning = self.var_warning_value
        else:
            allowed_error = self.error_value
            allowed_warning = self.warning_value
    
        # some tests fail when on Windows with OSMesa
        if os.name == "nt" and self.windows_skip_image_cache:
            return
        # high variation for MacOS
        if platform.system() == "Darwin" and self.macos_skip_image_cache:
            return
    
        # cached image name. We remove the first 5 characters of the function name
        # "test_" to get the name for the image.
    
        # simply save the last screenshot if it doesn't exist or the cache
        # is being reset.
        if self.reset_image_cache or not os.path.isfile(image_filename):
            return plotter.screenshot(image_filename)
    
        # otherwise, compare with the existing cached image
        error = pyvista.compare_images(image_filename, plotter)
        if error > allowed_error:
>           raise RuntimeError(
                f"{test_name} Exceeded image regression error of "
                f"{allowed_error} with an image error equal to: {error}"
            )
E           RuntimeError: test_add_text_latex Exceeded image regression error of 500 
with an image error equal to: 6860.6836601301275                                    

.venv/lib/python3.8/site-packages/pytest_pyvista/pytest_pyvista.py:165: RuntimeError
------------------------------- Captured stderr call --------------------------------
2023-01-17 11:08:41.288 (   1.193s) [        AB69E740]vtkMatplotlibMathTextUt:984    
ERR| vtkMatplotlibMathTextUtilities (0x33f4b50): MaskParser is not initialized!     
... <redacted, line repeats several times>
--------------------------------- Captured log call ---------------------------------
ERROR    root:errors.py:116 MaskParser is not initialized!
... <redacted, line repeats several times>
tests/plotting/test_plotting.py::test_add_text_latex  - 12.0 MB
FAILED tests/plotting/test_plotting.py::test_add_text_latex - RuntimeError: test_add_
text_latex Exceeded image regression error of 500 with an...                        
========================= 1 failed, 341 deselected in 1.32s =========================
```

pyvista/plotting/mapper.py
banesullivan: Aha! Thanks for clarifying. @adeak and I were both wondering why you were adding this. I know this is a trivial change, but could we open a dedicated PR for this to track the exact reason this is being introduced?

akaszynski: > could we open a dedicated PR for this to track the exact reason this is being introduced?

Yep. Easier to track in the blame. Opening...

tests/plotting/test_plotting.py
banesullivan: A follow up PR could add a conditional check to this test to validate VTK and MPL are compatible to run the test

pyvista/plotting/mapper.py
adeak: Can confirm [on the mypy playground](https://mypy-play.net/?mypy=latest&python=3.10) with
```py
import enum
from typing import Union

class FieldAssociation(enum.Enum):
    POINT = 2

def foo(var: Union[str, FieldAssociation]):
    if isinstance(var, FieldAssociation):
        reveal_type(var)
        var = var.name
    reveal_type(var)
```
On mypy newest (0.991) with Python 3.10:
```
main.py:9: note: Revealed type is "__main__.FieldAssociation"
main.py:11: note: Revealed type is "builtins.str"
Success: no issues found in 1 source file
```
On Python 3.11:
```
main.py:9: note: Revealed type is "__main__.FieldAssociation"
main.py:11: note: Revealed type is "Union[builtins.str, __main__.FieldAssociation]"
Success: no issues found in 1 source file
```
This might be covered by some enum issues in https://github.com/python/mypy/issues/12841, e.g. https://github.com/python/mypy/issues/12483...

banesullivan: Just need to remove it from this PR

tests/plotting/test_plotting.py
akaszynski: Addressed in #3838.

akaszynski:
> Probably should have raised them as separate issues. I'm guilty of "sneaking" in changes all the time that are unrelated but annoying to my local dev.

Done now in #3838 and #3837. Thanks for keeping me honest.

> the opacity transfer function still has a strong impact (seeminly 100% of the impact) on the transparency of the rendered volume where the Alpha component is seeminly ignored?

Absolutely correct. According to the [VTK documentation](https://vtk.org/doc/nightly/html/classvtkVolumeProperty.html#a2534a6dc500be42c9a842d405c4ee0bb):

>  The fourth component will be passed through the first scalar opacity transfer function for opacity and first gradient opacity transfer function for gradient based opacity modulation.

To clear things up I propose that instead of having the 4th component go through the opacity transfer function, we instead enforce a linear mapping (`opacity='linear'`) to avoid the unexpected behavior.

banesullivan (approved):
Okay looking good now! All is working as expected.

I do want to note that the default volume rendering for a given RGBA mesh looks different between ParaView and PyVista, and this is something I would like to encourage you to look into if you have time... 

| PyVista | ParaView |
|---|---|
|  Screen Shot 2023-01-20 at 8 50 29 AM | Screen Shot 2023-01-20 at 8 51 40 AM |



with that said, I think what you've done here is more in line with what I'd expect given the opacity array looks like:

Screen Shot 2023-01-20 at 8 54 00 AM

So good to merge by me! 

(I suspect the default opacity transfer function also needs to be disabled in ParaView)

akaszynski:
> (I suspect the default opacity transfer function also needs to be disabled in ParaView)

Probably just change it to linear within Paraview. Is that an option?

banesullivan:
> Probably just change it to linear within Paraview. Is that an option?

Probably! I didn't immediately or obviously see this choice and my ParaView skills are rusty these days. I think this is good to go in PyVista. Merge when ready

PR #3834, opened 2023-01-17
Add volume clipper

akaszynski:
Add a volume clipper widget.

volume-clip-plane-widget

```py
import numpy as np
import pyvista as pv

grid = pv.UniformGrid(dimensions=(200, 200, 200))
grid['scalars'] = np.linalg.norm(grid.center - grid.points, axis=1)
opacity = np.zeros(100)
opacity[::10] = np.geomspace(0.01, 0.75, 10)

pl = pv.Plotter()
pl.add_volume_clip_plane(grid, normal='-x', opacity=opacity[::-1], cmap='magma')
pl.show()
```

Bonus Modifications
- Change `event_type` to `interaction_event` for consistency.
- Consistently accept strings in `interaction_event` everywhere.

### DIFF
--- a/pyvista/plotting/mapper.py
+++ b/pyvista/plotting/mapper.py
@@ -270,7 +270,7 @@ class _BaseMapper(_vtk.vtkAbstractMapper):
     def scalar_map_mode(self, scalar_mode: Union[str, FieldAssociation]):
         if isinstance(scalar_mode, FieldAssociation):
             scalar_mode = scalar_mode.name
-        scalar_mode = scalar_mode.lower()
+        scalar_mode = scalar_mode.lower()  # type: ignore
         if scalar_mode == 'default':
             self.SetScalarModeToDefault()
         elif scalar_mode == 'point':
--- a/pyvista/plotting/mapper.py
+++ b/pyvista/plotting/mapper.py
@@ -270,7 +270,7 @@ class _BaseMapper(_vtk.vtkAbstractMapper):
     def scalar_map_mode(self, scalar_mode: Union[str, FieldAssociation]):
         if isinstance(scalar_mode, FieldAssociation):
             scalar_mode = scalar_mode.name
-        scalar_mode = scalar_mode.lower()  # type: ignore
+        scalar_mode = scalar_mode.lower()
         if scalar_mode == 'default':
             self.SetScalarModeToDefault()
         elif scalar_mode == 'point':
--- a/pyvista/plotting/widgets.py
+++ b/pyvista/plotting/widgets.py
@@ -51,7 +51,7 @@ def _parse_interaction_event(interaction_event):
             "Expected value for `interaction_event` is 'start', "
             f"'end', or 'always'. {interaction_event} was given."
         )
-    else:
+    elif not isinstance(interaction_event, _vtk.vtkCommand.EventIds):
         raise TypeError(
             "Expected type for `interaction_event` is either a str "
             "or an instance of `vtk.vtkCommand.EventIds`."
--- a/pyvista/plotting/widgets.py
+++ b/pyvista/plotting/widgets.py
@@ -23,6 +23,37 @@ from pyvista.utilities.algorithms import (
 from .colors import Color
 
 
+def _parse_interaction_event(interaction_event):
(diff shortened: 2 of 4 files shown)


pyvista/plotting/mapper.py
```diff
@@ -270,7 +270,7 @@ def scalar_map_mode(self) -> str:
     def scalar_map_mode(self, scalar_mode: Union[str, FieldAssociation]):
         if isinstance(scalar_mode, FieldAssociation):
             scalar_mode = scalar_mode.name
-        scalar_mode = scalar_mode.lower()
+        scalar_mode = scalar_mode.lower()  # type: ignore
```
banesullivan, line 273: Stop this!!!

pyvista/plotting/widgets.py
```diff
@@ -85,7 +122,7 @@ def add_box_widget(
 
         rotation_enabled : bool, optional
             If ``False``, the box widget cannot be rotated and is
-            strictly orthogonal to the cartesian axes.
+            strictly orthogonal to the Cartesian axes.
```
banesullivan, line 125: Good catch... this probably needs to be changed in 1000+ other places too

pyvista/plotting/widgets.py
```diff
@@ -651,6 +715,132 @@ def callback(normal, loc):
...
+        normal_rotation=True,
+        interaction_event='end',
+        origin=None,
+        **kwargs,
+    ):
+        """Clip a volume using a plane widget.
+
+        Parameters
+        ----------
+        volume : pyvista.plotting.Volume, pyvista.UniformGrid, pyvista.RectilinearGrid
```
banesullivan, line 739: are comma delimitors supported? we usually use `or` to separate supported types

pyvista/plotting/widgets.py
```diff
@@ -651,6 +715,132 @@ def callback(normal, loc):
...
+            raise TypeError(
+                'The `volume` parameter type must be either pyvista.UniformGrid, '
+                'pyvista.RectangularGrid, or a pyvista.plotting.volume.Volume '
+                'from `Plotter.add_volume`.'
+            )
+        else:
+            assert_empty_kwargs(**kwargs)
+
+        plane = _vtk.vtkPlane()
+        _widget = []
```
banesullivan, line 815: This local reference to track the plane widget seems weird to me... 

What happens if we add multiple volumes clipped by planes or add multiple planes clipping one volume?

banesullivan: Could we use `GetClippingPlanes` somehow from the mapper to track this instead?

banesullivan (changes requested):

pyvista/plotting/mapper.py
akaszynski: The madness continues in https://github.com/pyvista/pyvista/pull/3837

akaszynski: But really, I'll remove it here.

pyvista/plotting/widgets.py
akaszynski: Opening another PR to address this.

pyvista/plotting/widgets.py
akaszynski: According to https://numpydoc.readthedocs.io/en/latest/format.html, it looks like 'or' is prefered.

pyvista/plotting/widgets.py
akaszynski: > This local reference to track the plane widget seems weird to me...
What happens if we add multiple volumes clipped by planes or add multiple planes clipping one volume?

Multi-plane clipping works out, but I agree that it seems a bit weird here. Let me see if there's a better way to do this.

akaszynski: Fixed in 5483bdb9dd6fce17603153e65a63c4a1ec9dd486

Multiple planes work:

https://user-images.githubusercontent.com/11981631/213025985-20b98e57-9fa1-4b52-821c-52d56a0ceb35.mp4

banesullivan (approved):
This looks really nice and I appreciate the clean up to the interaction events here

Issue #3293, opened 2022-09-09
Filters should never modify the source data!!


banesullivan:
### Describe the bug, what's wrong, and what you expected.

A handful of filters are modifying their source/input data which is a violation of the VTK and PyVista design/architecture 

For example, the `glyph` filter set's the active scalars on the input mesh:

https://github.com/pyvista/pyvista/blob/4b3eaffb7c975f929503a2170f1e01c34670bed8/pyvista/core/filters/data_set.py#L2014

This is not how any VTK algorithm or PyVista should operate. Instead, VTK algorithms have a `SetInputArrayToProcess` method that tell the algorithm which array to use. Most filters in PyVista were implemented with `SetInputArrayToProcess` to avoid modifying the active scalars on the source data, but a few other filters have slipped through the cracks:

`streamlines_from_source`:
https://github.com/pyvista/pyvista/blob/4b3eaffb7c975f929503a2170f1e01c34670bed8/pyvista/core/filters/data_set.py#L3292-L3293

`streamlines_evenly_spaced_2D`:

https://github.com/pyvista/pyvista/blob/4b3eaffb7c975f929503a2170f1e01c34670bed8/pyvista/core/filters/data_set.py#L3479

`clip_scalar()`:

https://github.com/pyvista/pyvista/blob/4b3eaffb7c975f929503a2170f1e01c34670bed8/pyvista/core/filters/data_set.py#L417

I suspect there are more. Thank you, @whophil, for helping me find many of these.

## Solution

All filters that need to process an array should use the algorithm's `SetInputArrayToProcess` method much like what is done in the `threshold()` filter:

https://github.com/pyvista/pyvista/blob/4b3eaffb7c975f929503a2170f1e01c34670bed8/pyvista/core/filters/data_set.py#L1036

### Steps to reproduce the bug.

```python
from pyvista import examples
import pyvista as pv

mesh = examples.load_random_hills()

plotter = pv.Plotter()
plotter.add_mesh(mesh, scalars='Elevation')
# PROBLEM: running the `glyph` filter changes the active scalars and messes up the previous `add_mesh` call
arrows = mesh.glyph(scale="Normals", orient="Normals", tolerance=0.05)
plotter.add_mesh(arrows, color='black')
plotter.show()
```

![Screen Shot 2022-09-09 at 3 45 01 PM](https://user-images.githubusercontent.com/22067021/189450131-67d6850e-6cc8-4e03-8f88-9689ecaeaa08.png)

### System Information

[...]

PR #3842, opened 2023-01-18
Fix issues with `contour_banded()` and improve examples

### LINKED ISSUE #3780: `contour_banded()` array renaming logic is incorrect
The logic in `contour_banded()` that renames the output arrays is not robust enough:

https://github.com/pyvista/pyvista/blob/067b597624c1d5c3ec92e7eea2bfd5403f39224b/pyvista/core/filters/poly_data.py#L3471-L3475

This easily fails if the arrays are not the 0-index array, leading to arrays with names like `'Unnamed_1'` or `'Unnamed_2'`

For example:

```py
import pyvista as pv

mesh = pv.Sphere().elevation()
mesh.point_data['foo'] = mesh.points[:, 0]
mesh
```
Screen Shot 2023-01-05 at 4 07 30 PM

```py
mesh.contour_banded(5, generate_contour_edges=False)
```
Screen Shot 2023-01-05 at 4 07 38 PM

banesullivan:
Some bugfixes to `contour_banded()`

- [x] Resolves #3780: array renaming logic is fixed here (note this was also fixed upstream in https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9840)
- [x] Previously, `scalar_mode='index'` incorrectly used `SetScalarModeToValue()`
- [x] The examples have been improved
- [x] Use `SetInputArrayToProcess` per https://github.com/pyvista/pyvista/issues/3293


banesullivan:
these bug fixes are critical for me, and this filter hasn't even made it through a release cycle. merging due to priority

PR #3845, opened 2023-01-19
Add a `recompute_normals` parameter to `PolyData.save`

### LINKED ISSUE #3844: `PolyData.save`: Add a `recompute_normals` parameter when saving STL/PLY meshes

PyVista automatically recomputes the normals when saving a `PolyData` mesh to `stl` or `ply` files ( added in #489 ). The call to `compute_normals` in the `save` method is done with the default parameters. 

Oftentimes this creates unexpected behaviors since the normals already in the `PolyData` object might not have been computed with the default parameters (e.g. with `flip_normals=True` or `auto_orient_normals=True`). Thus, I was wondering if it would make sense to add a `recompute_normals` parameter to the `save` method of `PolyData`.

_No response_

### Pseudocode or Screenshots

```python
mesh.save("file.stl", recompute_normals=False)
```

LouisExtrality:
### Overview

resolves #3844 

### Details

- Add a `recompute_normals` parameter to the `PolyData.save` method
- Update documentation for `recompute_normals`

### DIFF
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -919,9 +919,9 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
 
         recompute_normals : bool, optional
             When ``True``, if ply or stl format is chosen, the face normals
-            are computed in place to ensure the mesh is properly saved. 
-            Default ``True``. Set this to ``False`` to save instead the 
-            already existing normal array in the PolyData. 
+            are computed in place to ensure the mesh is properly saved.
+            Default ``True``. Set this to ``False`` to save instead the
+            already existing normal array in the PolyData.
 
         Notes
         -----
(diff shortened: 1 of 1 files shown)


pyvista/core/pointset.py
```diff
@@ -918,6 +917,12 @@ def save(self, filename, binary=True, texture=None):
             .. note::
                This feature is only available when saving PLY files.
 
+        recompute_normals : bool, optional
+            When ``True``, if ply or stl format is chosen, the face normals
+            are computed in place to ensure the mesh is properly saved.
+            Default ``True``. Set this to ``False`` to save instead the
+            already existing normal array in the PolyData.
```
akaszynski, line 924:
```suggestion
        recompute_normals : bool, default: True
            When ``True``, if ply or stl format is chosen, the face normals
            are computed in place to ensure the mesh is properly saved.
            Set this to ``False`` to save instead the already existing normal
            array in the PolyData.
```

akaszynski (approved):
LGTM. Thanks!

banesullivan (approved):

PR #3855, opened 2023-01-23
Add clean UnstructuredGrid  filter

akaszynski:
Seems that this was added back in `vtk==9.0.3`: [vtkmCleanGrid](https://vtk.org/doc/nightly/html/classvtkmCleanGrid.html)

Works quite well to clean up unmerged points from the parallel vtu reader.

### DIFF
--- a/pyvista/core/filters/__init__.py
+++ b/pyvista/core/filters/__init__.py
@@ -28,7 +28,20 @@ from pyvista.utilities import wrap, ProgressMonitor
 
 
 def _update_alg(alg, progress_bar=False, message=''):
-    """Update an algorithm with or without a progress bar."""
+    """Update an algorithm with or without a progress bar.
+
+    Parameters
+    ----------
+    alg : vtkAlgorithm
+        VTK algorithm.
+
+    progress_bar : bool, default: False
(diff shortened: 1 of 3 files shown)


banesullivan:
The "VTK-m" filters require the `Accelerators/Vtkm` module to be enabled (`VTK_MODULE_ENABLE_VTK_AcceleratorsVtkm`), which is not turned on for the conda-forge build.

The test and error messages here may need to check if "VTK-m" is included in the modules and give a better fail message.

----

In Most cases, I would say that we shouldn't use VTK-m classes directly like this as most are implemented as "accelators" to existing algorithms and the `VTK_ENABLE_VTKM_OVERRIDES` build option makes it so many filters are overridden with these versions of the filter by default. For example, when using `vtkThreshold` and having `VTK_ENABLE_VTKM_OVERRIDES` enabled, we are actually using `vtkmThreshold` under the hood (as far as I understand how all of this works). 

This is a rare case where `vtkmCleanGrid` has no standard counterpart and warrants being used directly.

akaszynski:
Dropping this because of conda support. Also there are issues with this filter dropping arrays with more than three components.

PR #3863, opened 2023-01-23
Support Actor-based picking

banesullivan:
Enables support for picking individual actors. This is useful if you want the picking operation to modify how the actor is displayed.

NOTE: the example requires the new `actors` property on the `BasePlotter` from #3385


banesullivan:
Oof, I'm concerned that Sphinx Gallery did not run the modified example.

akaszynski (approved):
LGTM.

PR #3864, opened 2023-01-23
Track Actor names better and improve add_actor

banesullivan:
PyVista's `BasePlotter` tracks all `Actor`s added to the scene with a unique name. This is a widely used feature and is tested quite well. Prior to this PR, the `Actor`s themselves are unaware of their names, and I think PyVista's `Actor` class should track this value.

I don't like to add attributes that do not map to existing attributes/properties that exist in VTK, but in my opinion, this is a valid exception for a feature we've implemented specifically for PyVista.

Additionally, this adds support for explicit exceptions when using the `@no_new_attr` decorator. `_renderer` was a class attribute on `Actor` previous to this and I don't think it was actually working as expected. Through the use of `_new_attr_exceptions`, now classes can have a few explicit exceptions for new instance attributes like `_renderer` and `_name` for `Actor`

### DIFF
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -754,6 +754,8 @@ class Renderer(_vtk.vtkOpenGLRenderer):
                 # e.g., vtkScalarBarActor
                 name = actor.GetAddressAsString("")
 
+        actor.SetPickable(pickable)
+        self.AddActor(actor)  # must add actor before resetting camera
         self._actors[name] = actor
 
         if reset_camera:
@@ -782,8 +784,6 @@ class Renderer(_vtk.vtkOpenGLRenderer):
             else:
                 raise ValueError(f'Culling option ({culling}) not understood.')
 
-        actor.SetPickable(pickable)
-        self.AddActor(actor)
         self.Modified()
 
         prop = None
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -4,7 +4,6 @@ import collections.abc
 from functools import partial
 from typing import Sequence, cast
 import warnings
-from weakref import proxy
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


banesullivan:
I may refactor the actor tracking such that it's no longer a Key-Value mapping and simply tracks the Actors themselves...

work in progress

pyvista/plotting/renderer.py
```diff
@@ -732,20 +732,27 @@ def add_actor(
...
+            actor = Actor(mapper=actor, name=name)
 
-        self.AddActor(actor)
-        actor.renderer = proxy(self)
+        if isinstance(actor, Actor) and name:
+            # WARNING: this will override the name if already set on Actor
+            actor.name = name
+
+        if isinstance(actor, Actor):
+            actor.renderer = proxy(self)
```
banesullivan, line 747: I'm wondering why the `renderer` might want to be tracked on the `Actor`? Actors are often in multiple renderers. 

cc @akaszynski?

pyvista/utilities/misc.py
```diff
@@ -167,10 +167,12 @@ def set_pickle_format(format: str):
 
 def no_new_attr(cls):
     """Override __setattr__ to not permit new attributes."""
+    if not hasattr(cls, '_new_attr_exceptions'):
+        cls._new_attr_exceptions = []
 
     def __setattr__(self, name, value):
         """Do not allow setting attributes."""
-        if hasattr(self, name):
+        if hasattr(self, name) or name in cls._new_attr_exceptions:
```
banesullivan, line 175: Looking for feedback on this new mechanism

pyvista/plotting/renderer.py
akaszynski: It doesn't. Removed in [89475c0](https://github.com/pyvista/pyvista/pull/3864/commits/89475c045d0c4a3fdd3182f9343693e9030522e8).

pyvista/utilities/misc.py
akaszynski: Agree with this new approach. Thanks!

akaszynski (approved):
Agree with the changes. Thanks!

PR #3865, opened 2023-01-24
Fix inplace filters usage of copy_from

### LINKED ISSUE #3069: Adding a mesh to a plotter with smooth_shading=True leads to a deep copy of the mesh

banesullivan:
Resolves #3069 and closes #3087 

cc @akaszynski and @p-j-smith

`copy_from()` is being used throughout internally as an assumed shallow copy (see its use for the `inplace=True` scenarios of filters). This PR corrects the internal usage of `copy_from()` to be a shallow copy operation where needed by adding a keyword argument to opt-in to a shallow copy if desired

### DIFF
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -1960,13 +1960,22 @@ class DataSet(DataSetFilters, DataObject):
         """Return the object string representation."""
         return self.head(display=False, html=False)
 
-    def copy_from(self, mesh: _vtk.vtkDataSet):
+    def copy_from(self, mesh: _vtk.vtkDataSet, deep=False):
         """Overwrite this dataset inplace with the new dataset's geometries and data.
+
+        .. versionchanged:: 0.38.0
+            This would deep copy by default prior to version ``0.38.0`` and
+            has been modified to shallow copy by default moving forward
+            to fix issues where in-place filters were producing deep copy
+            outputs.
 
         Parameters
         ----------
         mesh : vtk.vtkDataSet
             The overwriting mesh.
+
+        deep : bool, default: False
+            Whether to perform a deep or shallow copy.
 
         Examples
         --------
@@ -1987,7 +1996,10 @@ class DataSet(DataSetFilters, DataObject):
                 f'The Input DataSet type {type(mesh)} must be '
                 f'compatible with the one being overwritten {type(self)}'
             )
-        self.deep_copy(mesh)
+        if deep:
+            self.deep_copy(mesh)
+        else:
+            self.shallow_copy(mesh)
         if is_pyvista_dataset(mesh):
             self.copy_meta_from(mesh, deep=True)
 
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -1960,21 +1960,15 @@ class DataSet(DataSetFilters, DataObject):
         """Return the object string representation."""
         return self.head(display=False, html=False)
 
-    def copy_from(self, mesh: _vtk.vtkDataSet, deep=False):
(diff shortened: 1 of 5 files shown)


akaszynski (dismissed):
Thanks. This was missing in the original implementation as well. Thanks for making it backwards compatible. Checking from the past (for my edification), it was indeed a `DeepCopy`:

https://github.com/pyvista/pyvista/blob/d13ead8bbc6968447cb25cd9614d7c5b853404c2/pyvista/core/dataobject.py#L57-L66

I think it's ok to have this without a depreciation warning since it's only going to be there for one minor release.

banesullivan:
@akaszynski, I actually changed my mind on this. I am keeping `copy_from()` to default as a deep copy for consistency with all other `copy*()` methods (as `deep=True` is the default everywhere else). This PR is now updating where `copy_from()` was used in-place as an assumed shallow copy operation to be explicitly `deep=False`. 

Please re-review if you would

akaszynski (approved):
>  This PR is now updating where copy_from() was used in-place as an assumed shallow copy operation to be explicitly deep=False.

Personally, I'm fine either way, but I agree that consistency is paramount, though there's a performance hit with always having a shallow copy. Either way, as long as we can explain our reasoning, I'm good with it.

banesullivan:
> though there's a performance hit with always having a shallow copy.

Not sure I follow? Do you mean for deep copy? Either way, this method is rarely used

Issue #2008, opened 2022-01-07
Theme updates should update more things


adeak:
## Prologue
I first noticed that the example for `Plotter.theme` doesn't actually work:
![Screenshot from 2022-01-07 22-05-28](https://user-images.githubusercontent.com/17914410/148607397-b589d7db-befa-4f2e-8a8f-f008bb5d3d39.png)
Note the blatant lack of dark theme in the figure.

I first found https://github.com/pyvista/pyvista/pull/2003 but even after fixing that I realised that theme updates don't change a lot of things in plots. In other words, `pl = pv.Plotter(theme=DarkTheme())` works, `pl = pv.Plotter(); pl.theme = DarkTheme()` only partly.

There are bits and pieces in `BasePlotter.__init__()`, e.g.
https://github.com/pyvista/pyvista/blob/1990b018fc82c4fcce2c7d5fcee3dece33a42821/pyvista/plotting/plotting.py#L258-L264

and `Plotter.__init__()`, e.g.
https://github.com/pyvista/pyvista/blob/1990b018fc82c4fcce2c7d5fcee3dece33a42821/pyvista/plotting/plotting.py#L4519-L4536

For example, `self.set_background()` defers to `self.renderers.set_background()` which dispatches to `self.renderer.set_background()`, which uses the `Renderer`'s `_theme` attribute which is a reference to the parent `Plotter`'s theme. However, when we update a `Plotter`'s theme using the setter, https://github.com/pyvista/pyvista/blob/1990b018fc82c4fcce2c7d5fcee3dece33a42821/pyvista/plotting/plotting.py#L284-L290

we're updating the theme but we aren't calling a bunch of methods that were called in the various `__init__()`s.

## Question

Do we agree that updating the theme should change things like renderer background colours? If we do, we need to identify these theme-specific initialization calls, move them to a private method, and call it both in the `__init__()`s and after updating themes. It gets even muddier due to both `BasePlotter` and `Plotter` being involved. (Side note: `pyvistaqt.BackgroundPlotter` doesn't seem to support per-plotter theming at all, alas.)

Any thoughts on how we could tackle this?

MatthewFlamm:
> Do we agree that updating the theme should change things like renderer background colours?

Why do we have a setter here unless this is true?  This is to say that I agree.  Your suggestion on implementation seems good to me and standardizing the internal API as much as possible makes sense to me too.

adeak:
Thanks for the feedback @MatthewFlamm.

[...]

PR #3870, opened 2023-01-25
Custom themes and more theme options

### LINKED ISSUE #3045: enable themes to set themselves

We should make the theme classes callable in some way to make it easier to set a theme with tab-completion.

Right now, setting a theme looks like:

```py
pv.set_plot_theme('name_i_have_to_lookup')
```

or even:

```py
my_theme = pv.themes.DefaultTheme()
pv.global_theme.load_theme(my_theme)
```
but what if instead, it was like this:

```py
pv.themes.Document.load()
```
 working out the instance method and class method might be fun, but all of this should be doable.

n/a

### Pseudocode or Screenshots

_No response_

banesullivan:
I've come across a situation where I want a project- or application-specific theme that is easily re-usable and extensible (for instance, maybe a downstream package wants all of their 3D visualizations to have the same look/feel). These changes make it possible to subclass `DefaultTheme` and create user-defined themes. Further, I am working to expose a lot more rendering options as theme options.

@pyvista/developers, I'd love feedback and any suggestions on additional options to expose in the theme.

resolves #3045, helps with #2008

After this lands, we will want to better group many of these rendering parameters


banesullivan:
I think I've reached my limit on what to expose in this PR. Ready for review!

I think there should be another PR after this for 0.39 that moves a lot of these configurations into more subclasses/subnamespaces. The `DefaultTheme` namespace is getting really crrowded

akaszynski (approved):

banesullivan:
Let's get this out for 0.38 as is and let a few folks try out the custom themes and wait to document it until we have some feedback/reassurance that this all works in a useful way. That way, we can sneakily avoid having to maintain backward compat if needing to change this.

PR #3875, opened 2023-01-25
Avoid awaiting the Trame Jupyter backend

banesullivan:
This is based on #3385 and must merge into `feat/trame`.

The new trame-based Jupyter backend was requiring end-users to `await` the `set_jupyter_backend()` call. @akaszynski and I feel strongly that this introduces too much user overhead and inconvenience by preventing the use of an environment variable to automatically set the Jupyter backend to use trame.

These changes leverage [nest_asyncio](https://github.com/erdewit/nest_asyncio) which patches the standard lib `asyncio` package. This may have unintended consequences for some uses cases,so I've tried to make it opt-in for when `show_trame` is called or the backend is explicitly set the use Trame. I advise strongly to make sure PyVista's Jupyter backend is not set to use Trame when not in a Jupyter environment.

To be transparent, monkey patching (which is what I think `nest_asyncio` is doing) scares the 💩  out of me, and we should really avoid this... This PR makes it so this only happens when explicitly trying to use the Jupyter backend and is easily opt-out-able by using the old await-able API:

```py
import pyvista as pv
from pyvista.trame.jupyter import launch_server

await launch_server()
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,20 +54,20 @@ colormaps = [
 io = ['meshio>=5.2']
 jupyter = [
     'ipyvtklink',
+    'jupyter-server-proxy',
+    'nest_asyncio',
     'panel',
     'pythreejs',
     'trame>=2.2.6',
     'trame-client>=2.4.2',
     'trame-server>=2.8.0',
     'trame-vtk>=2.0.15',
-    'jupyter-server-proxy',
 ]
 trame = [
     'trame>=2.2.6',
     'trame-client>=2.4.2',
     'trame-server>=2.8.0',
     'trame-vtk>=2.0.15',
-    'jupyter-server-proxy',
 ]
 
 [project.urls]
--- a/pyvista/trame/jupyter.py
+++ b/pyvista/trame/jupyter.py
@@ -1,4 +1,5 @@
 """Trame utilities for running in Jupyter."""
+import asyncio
 import logging
 import os
 import warnings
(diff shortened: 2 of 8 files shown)


banesullivan:
To be transparent, monkey patching (which is what I think `nest_asyncio is doing) scares the 💩  out of me, and we should really avoid this... This PR makes it so this only happens when explicitly trying to use the Jupyter backend and is easily opt-out-able by using the old await-able API:

```py
import pyvista as pv
from pyvista.trame.jupyter import launch_server

await launch_server(pv.global_theme.trame.jupyter_server_name)
```

akaszynski:
Agree with this. There are side effects for sure, but the +400 stars for `nest_asyncio` shows that this is indeed something that's desired (and used) by the community.

pyproject.toml
```diff
@@ -54,20 +54,20 @@ colormaps = [
 io = ['meshio>=5.2']
 jupyter = [
     'ipyvtklink',
+    'jupyter-server-proxy',
+    'nest_asyncio',
```
jourdain, line 58: may not need it

pyvista/trame/jupyter.py
```diff
@@ -184,3 +187,37 @@ def show_trame(
...
+"""
+        )
+
+    async def launch_it():
+        await launch_server(server)
+
+    # Basically monkey patches asyncio to support this
+    nest_asyncio.apply()
+
+    return asyncio.run(launch_it())
```
jourdain, line 223: All that method might be able to be replaced by just `asyncio.get_event_loop().run_until_complete(launch_server(server))`

requirements_test.txt
```diff
@@ -12,11 +12,11 @@ ipyvtklink<0.4.0
 ipywidgets<9.0.0
 matplotlib<3.7.0
 meshio<5.4.0
+nest_asyncio<1.5.7
```
jourdain, line 15: might not need it after all.

pyvista/trame/jupyter.py
jourdain: Actually, they are patching the loop which could explain why the above code would actually work.

banesullivan: > they are patching the loop

Exactly. I think this is necessary because we are in Jupyter's already running event loop

PR #3894, opened 2023-01-28
Add BaseVTKReader for custom Reader; Use for PVDReader

### LINKED ISSUE #3893: PVD files cannot be read with `pv.read`

```py
import pyvista as pv
from pyvista import examples
filename = examples.download_dual_sphere_animation(load=False)
pv.read(filename)
```

```
Input In [5], in <cell line: 1>()
----> 1 pv.read(filename)

File ~/pyvista/pyvista/utilities/fileio.py:191, in read(filename, attrs, force_ext,
 file_format, progress_bar)
    189 else:
    190     observer = pyvista.utilities.errors.Observer()
--> 191     observer.observe(reader.reader)
    192     if attrs is not None:
    193         _apply_attrs_to_reader(reader, attrs)

File ~/pyvista/pyvista/utilities/errors.py:165, in Observer.observe(self, algorithm
)
    163 if hasattr(algorithm, 'GetExecutive') and algorithm.GetExecutive() is not N
one:
    164     algorithm.GetExecutive().AddObserver(self.event_type, self)
--> 165 algorithm.AddObserver(self.event_type, self)
    166 self.__observing = True
    167 return

AttributeError: 'PVDReader' object has no attribute 'AddObserver'
```

### System Information

```shell
  Date: Sat Jan 28 08:24:06 2023 EST

                OS : Linux
            CPU(s) : 3
           Machine : x86_64
      Architecture : 64bit
               RAM : 1.9 GiB
       Environment : IPython
       File system : ext4
       GPU Details : error

           pyvista : 0.38.dev0
               vtk : 9.2.2
             numpy : 1.22.4
           imageio : 2.19.3
            scooby : 0.5.12
             pooch : v1.6.0
        matplotlib : 3.5.2
           IPython : 8.4.0
          colorcet : 3.0.0
           cmocean : 2.0
        ipyvtklink : 0.2.2
             scipy : 1.8.1
              tqdm : 4.64.0
            meshio : 5.3.4
        jupyterlab : 3.4.4
         pythreejs : Version unknown
    pytest_pyvista : 0.1.3
```

### Screenshots

_No response_

MatthewFlamm:
### Overview

 

This adds an abstract class that sets the requirements and some defaults for creating a custom reader that exists outside vtk.  We have GIFReader and PVDReader currently.

GIFReader already had the machinery from #3038, this PR refactors to allow reuse and implements for PVDReader.

### Details

Closes #3893


pyvista/utilities/reader.py
```diff
@@ -171,6 +171,51 @@ def get_reader(filename, force_ext=None):
...
+        """Add Observer that can be triggered during Update."""
+        self._observers.append([event_type, callback])
+
+    def RemoveObservers(self, *args):
+        """Remove Observer."""
+        self._observers = []
+
+    def GetProgress(self):
+        """GetProgress."""
+        return 0.0
```
banesullivan, line 199: If subclasses are monitored with `ProgressMonitor`, this hardcoded return value will mess it up

banesullivan (commented):
Looking good to me!

pyvista/utilities/reader.py
MatthewFlamm: I'm not actually sure that this does anything when implemented purely in python.  But, even if I'm wrong, if a subclass can declare progress incrementally, it can override this like GIFReader.

banesullivan: If called from Python, this overrides. It just doesn't affect any calls on the C++ side

MatthewFlamm: I'm thinking the default could check whether `_data_object is None` and return 1 if so.

Probably mostly bike-shedding comments: Since this is a purely sync single thread method, as implented, the entire `Update` method runs before any in progress state could be observed.  We only call `Update` during the `read` method of `BaseReader`, so the user will only see a change _after_ they've already received the mesh, and it will go from 0.0 to 1.0.

banesullivan (approved):

akaszynski:
Merging this so it can be within our release window.

Issue #2924, opened 2022-06-30
Off screen plotter not working, even with off_screen=True


juanilarregui:
### Describe the bug, what's wrong, and what you expected.

Hi guys, I'm trying to render with an off_screen plotter, but still getting:

```
AttributeError: 
This plotter has not yet been set up and rendered with ``show()``.
Consider setting ``off_screen=True`` for off screen rendering.
```

I already created the plotter as:
`plotter = pv.Plotter(off_screen=True)`
but I'm still getting the same error.

I also previously run:
```
sudo apt-get install xvfb
export DISPLAY=:99.0
export PYVISTA_OFF_SCREEN=true
export PYVISTA_USE_IPYVTK=true
Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
sleep 3
```

I'm also setting these environment variables just before the pyvista import in my Python code:
```
import os
os.environ['PYVISTA_OFF_SCREEN'] = 'true'
os.environ['PYVISTA_USE_IPYVTK'] = 'true'
os.environ['PYVISTA_USE_PANEL'] = 'true'
import pyvista as pv
```

**Note**: I don't have `python-qt4` installed since `python-qt4` package is not available over Ubuntu 22.04.

### Steps to reproduce the bug.

Next is the code that reproduces the error. This code was working on `pyvista 0.26` over Ubuntu 16.04, Python 3.5. I just upgraded to Ubuntu 22.04 with `pyvista 0.34.2`, Python 3.10.

```
plotter = pv.Plotter(off_screen=True)
plotter.add_mesh(mesh, texture=texture, lighting=False)
# Get render:
render_as_array = plotter.image.copy()  # This is the line that throws the error.
 ```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Thu Jun 30 17:20:03 2022 -03

                OS : Linux
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 15.5 GiB
       Environment : Jupyter
       File system : ext4
        GPU Vendor : Mesa/X.org
      GPU Renderer : llvmpipe (LLVM 13.0.1, 256 bits)
       GPU Version : 4.5 (Core Profile) Mesa 22.0.1

  Python 3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0]

           pyvista : 0.34.2
               vtk : 9.2.0
             numpy : 1.22.4
           imageio : 2.19.3
[...]

PR #3897, opened 2023-01-29
Scaled screenshots

### LINKED ISSUE #3698: Changing window size requires manual re-render on repeated `Plotter.screenshot()` calls after the first

When changing a plotter's `window_size` and then saving a screenshot, the scene is not re-rendered if it is not the first screenshot. The scene not being automatically rendered after the first screenshot [appears to be intentional](https://github.com/pyvista/pyvista/blob/f5dbce80bc44ff3def8c5eb15db054706fc5f38d/pyvista/plotting/plotting.py#L5090). However, this makes the `window_size` argument of `Plotter.screenshot()` unusable.

```python
import pyvista
sphere = pyvista.Sphere()
plotter = pyvista.Plotter(off_screen=True)
actor = plotter.add_mesh(sphere)
plotter.window_size = (800, 800)
plotter.screenshot('screenshot1.png') # Correct
plotter.window_size = (1000, 1000)
plotter.screenshot('screenshot2.png') # Wrong
plotter.render()
plotter.screenshot('screenshot3.png') # Correct
plotter.screenshot('screenshot4.png', window_size=(600, 600)) # Wrong
```

### System Information

```shell
  Date: Thu Dec 08 18:55:07 2022 Eastern Standard Time

                OS : Windows
            CPU(s) : 20
           Machine : AMD64
      Architecture : 64bit
       Environment : Python
      GPU Renderer : GeForce GTX 1660/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 456.71

  Python 3.10.6 | packaged by conda-forge | (main, Aug 22 2022, 20:30:19) [MSC
  v.1929 64 bit (AMD64)]

           pyvista : 0.36.1
               vtk : 9.1.0
             numpy : 1.23.2
           imageio : 2.21.1
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.3
         pyvistaqt : 0.9.0
             PyQt5 : 5.15.7
          colorcet : 3.0.0
             scipy : 1.9.0

  Intel(R) oneAPI Math Kernel Library Version 2021.4-Product Build 20210904
```

### Screenshots

The saved images:

### LINKED ISSUE #3699: Save images with supersampling

It would be very helpful to be able to capture images from on-screen plotters independently of the window size. I am using PyVista within a QT application where off-screen rendering is not applicable, and the resolution of saved images is limited by the resolution of the screen. This becomes an obstacle when creating things such as high-DPI figures for printing.

This functionality is available in [Mayavi's `savefig()` method](https://docs.enthought.com/mayavi/mayavi/auto/mlab_figure.html?highlight=savefig#mayavi.mlab.savefig) with the `magnification` argument. Mayavi's implementation uses the [VTK vtkWindowToImageFilter class](https://vtk.org/doc/nightly/html/classvtkWindowToImageFilter.html) and its `Scale` attribute to accomplish this, as far as I understand.

_No response_

### Pseudocode or Screenshots

Suppose you are working on a 1920x1080 screen and producing figures "by hand" - interactively. The images need to be very large, for example to be printed on a poster.
```python
# ...
# Set up and use interactive plotter
# ...
plotter.window_size = (800, 800)
plotter.screenshot('screenshot.png', magnification=4)
# -> Saved image is 3200x3200 pixels
```

banesullivan:
Resolves #3699, resolves #3698, Might help with #2924

This makes it easy to scale up the resolution of screenshots -- a long requested feature

Also fixes an issue where setting the `window_size` with screenshots didn't work.

@rileychall, would you be able to review this?

```py
>>> import pyvista as pv
>>> pl = pv.Plotter()
>>> pl.add_mesh(pv.Wavelet())
>>> pl.window_size = (800, 800)
>>> image = pl.screenshot('screenshot.png', scale=4)
>>> image.shape
(3200, 3200, 3)
# Saved PNG is also 3200x3200 pixels
```


pyvista/plotting/plotting.py
```diff
@@ -1627,13 +1637,43 @@ def image(self):
...
 
         # ignore alpha channel
         return data[:, :, :-1]
 
+    @property
+    def image_scale(self) -> int:
+        """Get or set the scale factor when saving a screenshot.
+
+        This will scale up the screenshots taken of the render
+        window so save a higher resolution image than what is rendered
```
banesullivan, line 1652:
```suggestion
        window to save a higher resolution image than what is rendered
```

pyvista/plotting/plotting.py
```diff
@@ -1627,13 +1637,43 @@ def image(self):
...
+        on screen.
+
+        Image sizes will be the :py:attr:`window_size <pyvista.BasePlotter.window_size`
```
banesullivan, line 1655:
```suggestion
        Image sizes will be the :py:attr:`window_size <pyvista.BasePlotter.window_size>`
```

banesullivan:
@pyvista/developers, this is ready for review, please

FYI, I added a single test to check the scaling, but I also want to note that this is a fundamental change to the `screenshot()` mechanism and thus picked up/covered by nearly all of our ~340 plotting image regression tests

tests/plotting/test_plotting.py
```diff
@@ -1060,6 +1060,56 @@ def test_screenshot(tmpdir):
...
+    # plotter.show(auto_close=False)  # for image regression test
+
+    plotter.window_size = (1000, 1000)
+    b = plotter.screenshot()
+    assert b.shape == (1000, 1000, 3)
+    # plotter.show(auto_close=False)  # for image regression test
+
+    d = plotter.screenshot(window_size=(600, 600))
+    assert d.shape == (600, 600, 3)
+    # plotter.show()  # for image regression test
```
banesullivan, line 1108: The image regression fixture is only picking up on the last `show()` call but really really need to make sure the image doesn't have the right size due to padding like

download

banesullivan: This will have to come later as it will likely require a change to `pytest-pyvista`

pyvista/plotting/plotting.py
```diff
@@ -1584,6 +1594,28 @@ def window_size(self):
...
         """Set the render window size."""
         self.render_window.SetSize(window_size[0], window_size[1])
+        self.render()
+
+    @contextmanager
+    def window_size_context(self, window_size=None):
+        """Set the render window size in an isolated context."""
+        # No op if not set
+        if window_size is None:
+            yield
```
banesullivan, line 1604:
```suggetsion
            yield self
```

pyvista/plotting/plotting.py
```diff
@@ -1584,6 +1594,28 @@ def window_size(self):
...
+            return
+        # If render window is not current
+        if self.render_window is None:
+            warnings.warn('Attempting to set window_size on an unavailable render widow.')
+            yield
```
banesullivan, line 1609:
```suggestion
            yield self
```

akaszynski (approved):
LGTM.

Thanks for cleaning this up, and I like the addition of the context manager.

Added a few examples and additional tests since I like testing features at the lowest possible level (i.e. at the individual unit level).

rileychall:
This looks great! I don't know if I'm familiar enough to do a real review, but the functionality is exactly what I was hoping for and more.

akaszynski (dismissed):

banesullivan:
Thanks for taking a look, @rileychall!

akaszynski (approved):

pyvista/plotting/plotting.py
```diff
@@ -1581,6 +1591,47 @@ def window_size(self):
...
+            return
+        # If render window is not current
+        if self.render_window is None:
+            warnings.warn('Attempting to set window_size on an unavailable render widow.')
+            yield self
+            return
+        size_before = self.window_size
+        if window_size is not None:
+            self.window_size = window_size
+        yield self
```
larsoner, line 1630: IIUC note that (IIRC, unlike pytest) `@contextmanager` needs to use this pattern
```
try:
    yield
finally:
    set_whatever = back_to_original_value
```
https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

> At the point where the generator yields, the block nested in the [with](https://docs.python.org/3/reference/compound_stmts.html#with) statement is executed. The generator is then resumed after the block is exited. If an unhandled exception occurs in the block, it is reraised inside the generator at the point where the yield occurred. Thus, you can use a [try](https://docs.python.org/3/reference/compound_stmts.html#try)…[except](https://docs.python.org/3/reference/compound_stmts.html#except)…[finally](https://docs.python.org/3/reference/compound_stmts.html#finally) statement to trap the error (if any), or ensure that some cleanup takes place. If an exception is trapped merely in order to log it or to perform some action (rather than to suppress it entirely), the generator must reraise that exception. Otherwise the generator context manager will indicate to the with statement that the exception has been handled, and execution will resume with the statement immediately following the with statement.

banesullivan: Okay woah. I confidently though that the `@contextmanager` handled this for us. I just tested this and indeed, encountering a exception indeed fails to set back the original value.

Sample code:

```py
from contextlib import contextmanager

class Contextual:
    def __init__(self):
        self.value = 1
        
    @contextmanager
    def context(self, value):
        before = self.value
        self.value = value
        yield
        self.value = before

c = Contextual()

with c.context(5):
    print(f'In context: {c.value}')
    raise RuntimeError()
```

Then after error:
```py
>>> c.value
5
```

I will open up a follow up PR to fix this

Thanks, @larsoner

Issue #3809, opened 2023-01-10
Exposing PyVista's filters as vtkAlgorithms


banesullivan:
#3318 is soon to introduce broad `vtkAlgorithm` support to `BasePlotter.add_mesh()` (and a few other methods). While initially, this is intended to be used by more advanced users, I'd like to move towards a pipeline filtering API that is approachable to the average user.

I'm trying to brainstorm some ways to go about this and would like to solicit community involvement.

## `return_algorithm` keyword argument 

We could introduce a `return_algorithm` keyword argument for all filter methods that will return the `vtkAlgorithm` for that filter. This would allow a user to configure the algorithm to process their data the way they like and connect to a pipeline. In practice, this may look like:

```py
alg = mesh.slice(return_algorithm=True)

def update_mesh(*args):
    mesh['scalars'] = np.random.random(mesh.n_points)

pl = pv.Plotter()
pl.add_mesh(alg)
pl.add_key_event('t', update_mesh)
pl.show()
# User hits `t` key and the rendered slice will update with the new data
```

This gets tricky when we want to chain operations together, e.g., `mesh.threshold().slice()`. I'm not sure of how to go about designing a sane API for chaining algorithms together like this without a full re-wrtie of the the filters.

## Keep in mind

FWIW, VTK recently introduced an `override` mechanism that will make wrapping filters (`vtkAlgorithm`s) much easier. I see a future where all the VTK filter classes are wrapped in a PyVista-like fashion. 

Perhaps the best solution is to go this route... re-wrapping the filters as Pythonic `vtkAlgorithm`s.

For example:

```py
@vtkContourFilter.override
class Contour(vtkContourFilter):
    def __call__(
        self,
        inp,  # Could be mesh or other algorithm
        isosurfaces=10,
        scalars=None,
        compute_normals=False,
        compute_gradients=False,
        compute_scalars=True,
        rng=None,
        preference='point',
        method='contour',
        progress_bar=False,
    ):
        ...  # Set up algorithm

    def __init__(self, *args, **kwargs):
        super().__init__()
        self(*args, **kwargs)

    ...
    @property.setter
    def isosurfaces(self, n):
        # Set number of isosurfaces
        ...

mesh = pv.Wavelet()
alg = Contour(mesh, scalars=mesh.active_scalars_name)

def set_n(n):
    alg.isosurfaces = n

pl = pv.Plotter()
pl.add_mesh(alg)
pl.add_slider_widget(set_n, (2, 20))
[...]

PR #1689, opened 2021-09-23
Refactor silhouette from add_mesh

MatthewFlamm:
### Overview

`add_mesh` is a monster and the logic is very complex depending on combinations of input parameters.  I think it is best if we can extract out as much logic as possible into separable pieces.  Silhouette functionality is completely separate from all the other pieces. This PR separates out the silhouette functionality into it's own method, that can be called on it's own if desired.  `add_mesh` calls this new method.

### Details

My personal style preference would be to have the below for any functionality like this:

```py
plotter.add_mesh(mesh)
plotter.add_silhouette(mesh)
```

This makes it more clear that we have two separate actors being added to the scene.  I'm not sure if the current state has problems. If the mesh is updated, will the silhouette also update?  In any case, separating out the logic is the first step.

TODO:

- [x] Docstring
- [x] Examples for docstring where new method is called directly
- [x] Test where new method is called directly

PR #3901, opened 2023-01-29
Improve add_silhouette with kwargs and remove default decimation

### LINKED ISSUE #3816: Add `vtkAlgorithm` decimation support in silhouette plotting

https://github.com/pyvista/pyvista/pull/3318 is soon to introduce broad vtkAlgorithm support to BasePlotter.add_mesh (and a few other methods). It failed to introduce support decimation with silhouette plotting. 

_No response_

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

### LINKED ISSUE #3900: The default silhouette looks bad

The default parameters for silhouette don't look all that great... Can we somehow improve the default values for this?

```python
import pyvista as pv
pv.set_plot_theme('default')

p = pv.Plotter()
p.add_mesh(pv.Sphere(), silhouette=True)
p.show()
```

### System Information

```shell
`main`
```

### Screenshots

screenshot (1)

banesullivan:
Not sure why this function took a dict instead of actual keyword arguments. These changes make `add_silhouette()` more user-friendly.

Further, this resolves #3900 by decimating by default to make the silhouette lines smoother.

This also adds support for any dataset type as an input instead of just PolyData. Again the power of vtkAlgorithm support is showing up!

Resolves #3816


pyvista/utilities/algorithms.py
```diff
@@ -313,3 +313,20 @@ def point_data_to_cell_data_algorithm(inp, pass_point_data=False):
...
+    return sfilter
+
+
+def triangulate_algorithm(inp):
+    """Triangulate input."""
+    trifilter = _vtk.vtkTriangleFilter()
+    trifilter.PassVertsOff()
+    trifilter.PassLinesOff()
+    set_algorithm_input(trifilter, inp)
+    return trifilter
```
banesullivan, line 332: Considering these exist as filters on PolyDataFilters... we should really address #3809 soon to avoid this module continuing to grow with duplicated code. In the meantime, this will do.

pyvista/themes.py
```diff
@@ -309,7 +309,7 @@ def __init__(self):
         self._line_width = 2
         self._opacity = 1.0
         self._feature_angle = None
-        self._decimate = 0.9
+        self._decimate = None
```
banesullivan, line 312: This resolves #3900

pyvista/plotting/plotting.py
```diff
@@ -3965,6 +3970,14 @@ def add_silhouette(
             mesh, algo = algorithm_to_mesh_handler(algo)
 
         silhouette_params = self._theme.silhouette.to_dict()
+
+        if params is not None:
+            # Deprecated on 0.38.0, estimated removal on v0.40.0
+            warnings.warn(
+                '`params` is depreciated. Set the arguments directly.', PyVistaDeprecationWarning
```
banesullivan, line 3977:
```suggestion
                '`params` is deprecated. Set the arguments directly.', PyVistaDeprecationWarning
```

akaszynski (approved):
LGTM. Thanks for cleaning this up.

MatthewFlamm:
> Not sure why this function took a dict instead of actual keyword arguments. These changes make `add_silhouette()` more user-friendly.

This was added in #1689.  `add_mesh` currently takes a `dict` of params for `silhouette`.  The refactor made this a like-for-like refactor, so a user could use the same inputs.  However, I agree that this method should be made more user friendly.  IMO it would be nice to remove this completely from `add_mesh` after deprecating, but I haven't looked deeply to see if there is some reason to keep it in there.

banesullivan:
This implementation matches what we do for `add_scalar_bar()`

MatthewFlamm:
> IMO it would be nice to remove this completely from `add_mesh` after deprecating, but I haven't looked deeply to see if there is some reason to keep it in there.

From #1689, in reference to not allowing silhouette to be added directly from `add_mesh`:

> This makes it more clear that we have two separate actors being added to the scene.

This would suggest that we should deprecate and then remove from `add_mesh`.

akaszynski:
> This would suggest that we should deprecate and then remove from `add_mesh`

I could go either way. `add_mesh` absolutely a monster but sometimes it's convenient to have a simple parameter there to control the look of the dataset being added to the scene.

banesullivan:
Yeah, with removing the default decimation - a simple `silhouette=True` argument in `add_mesh()` is really nice to have to make a plot look better

banesullivan:
Plus #3870 is making this a theme-settable option to display silhouettes by default is enabled

akaszynski:
Ok. Let's keep this... for now.

PR #3889, opened 2023-01-28
Fix a few issues with Plotter.show

banesullivan:
Fix a few issues with `Plotter.show()`

Resolves #1635, Resolves #3548, Resolves #2953, Resolves #1877, Resolves #1665, Resolves #1621

Some cleanups to improve the return values of `Plotter.show()` so that if `return_*` is set, the requested item will always be returned.

Further, this removes the `store_image` property as the whole point of `last_image` was always to have an image after rendering. I probably need to deprecate this instead of outright removing... 

@akaszynski and @pyvista/developers, I would appreciate a thorough review on this one as changing up the `show()` method nearly always has unexpected consequences

Issue #3879, opened 2023-01-26
Expose mechanism to easily control the Jupyter trame view size


banesullivan:
### Describe the feature you would like to be added.

Reference https://github.com/pyvista/pyvista/pull/3385#discussion_r1088226298

### Pseudocode or Screenshots

In a Jupyter notebook:

```python
import pyvista as pv

mesh = pv.Wavelet()

p = pv.Plotter()
actor = p.add_mesh(mesh)
p.show(jupyter_backend='trame', width='75%')
```
 or something similar so that users can easily control the size of the view

banesullivan:
Hoping to do this after #3897 lands

PR #3902, opened 2023-01-30
Deprecate non-Trame Jupyter backends

### LINKED ISSUE #3690: Towards a single, PyVista/VTK-native Jupyter Backend

banesullivan:
The final step to resolve #3690 -- deprecates all previous notebook backends in favor of the new Trame-based backend.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -48,11 +48,6 @@ warnings.filterwarnings(
     category=UserWarning,
     message="Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.",
 )
-warnings.filterwarnings(
-    "ignore",
-    category=Warning,
-    message=r".*backend is deprecated and is planned for future removal.",
-)
 
 # -- General configuration ------------------------------------------------
 numfig = False
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -48,6 +48,11 @@ warnings.filterwarnings(
     category=UserWarning,
     message="Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.",
 )
+warnings.filterwarnings(
+    "ignore",
+    category=Warning,
+    message=r".*backend is deprecated and is planned for future removal.",
+)
 
 # -- General configuration ------------------------------------------------
 numfig = False
--- a/doc/user-guide/jupyter/ipygany.rst
+++ b/doc/user-guide/jupyter/ipygany.rst
@@ -2,10 +2,10 @@
 
 Using ``ipygany`` with PyVista
 ------------------------------
-.. warning::
-   Currently, this backend has inferior support and features than the
-   ``pythreejs``.  If you would like accurate recreations of VTK
-   scenes in three.js, please see :ref:`pythreejs_ref`.
+.. deprecated:: 0.38.0
+   This backend has been deprecated in favor of :ref:`trame_jupyter` - a new
+   framework for building dynamic web applications with Python with great
+   support for VTK.
 
 The `ipygany <https://github.com/QuantStack/ipygany>`_ jupyterlab
 plotting backend is a powerful module that enables pure plotting that
--- a/doc/user-guide/jupyter/ipyvtk_plotting.rst
+++ b/doc/user-guide/jupyter/ipyvtk_plotting.rst
@@ -3,10 +3,10 @@
 Using ``ipyvtklink`` with PyVista
 ---------------------------------
 
-.. warning::
-
-    The ``ipyvtklink`` backend is deprecated in favor of the new
-    ``server`` backend powered by ``trame``. See :ref:`trame_jupyter`.
+.. deprecated:: 0.38.0
+   This backend has been deprecated in favor of :ref:`trame_jupyter` - a new
+   framework for building dynamic web applications with Python with great
+   support for VTK.
 
 ``pyvista`` has the ability to display fully featured plots within a
 JupyterLab environment using ``ipyvtklink``.  This feature works by
--- a/doc/user-guide/jupyter/panel.rst
+++ b/doc/user-guide/jupyter/panel.rst
@@ -2,6 +2,13 @@
 
 Using ``Panel`` with PyVista
 ----------------------------
+
+.. deprecated:: 0.38.0
+   This backend has been deprecated in favor of :ref:`trame_jupyter` - a new
+   framework for building dynamic web applications with Python with great
+   support for VTK.
+
+
 PyVista supports the usage of the `panel
 <https://github.com/holoviz/panel>`_ module as a ``vtk.js`` jupyterlab
 plotting backend that can be utialized as either a standalone VTK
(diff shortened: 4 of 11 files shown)


banesullivan:
@pyvista/developers, I'd like to land this for `0.38` but if there are reservations about deprecating all other backends so quickly after the introduction of the new Trame-based backend, we can delay this to `0.39`

larsoner:
I think we need https://github.com/pyvista/pyvista/issues/3879 first, and maybe also the code to tell people nicely when a plot isn't expected to work (volume rendering or whatever?)

larsoner:
I think we need https://github.com/pyvista/pyvista/issues/3879 first, and maybe also the code to tell people nicely when a plot isn't expected to work (volume rendering or whatever?)

banesullivan:
> I think we need https://github.com/pyvista/pyvista/issues/3879 first,

On its way, but I need land #3889 and #3897 first. Do you think you could help review those?

> maybe also the code to tell people nicely when a plot isn't expected to work (volume rendering or whatever?)

At this point, the trame-client side backend is better supported and more feature rich than any of the other backends. There are no longer any features (that I'm aware of) that are support by other backends and not the trame backend.

FYI, volume rendering works fine now! (ref https://github.com/Kitware/trame-vtk/pull/19)

akaszynski (approved):
I'm happy to change the backend default, especially now that volume rendering works by default.

We can always submit a patch release should we need to revert.

akaszynski:
Fairly sure that warnings from the docbuild are difficult to suppress: https://github.com/sphinx-doc/sphinx/pull/11024/files

banesullivan:
Erm... Should I just remove where these backends are used in examples?

akaszynski:
> Erm... Should I just remove where these backends are used in examples?

I'd advocate for a soft deprecation here. If we want to, we can include a note or warning directive to indicate that these are being replaced. This will give users a chance to see and compare the new vs. the old.

akaszynski:
Ironically, the brand new jupyter backend doesn't support the coolest feature we have: being able to visualize in static webpages.

I'd really like to make sure we don't block the usage of these alternative backends until we can be feature parity.

banesullivan:
> Ironically, the brand new jupyter backend doesn't support the coolest feature we have: being able to visualize in static webpages.
> 
> I'd really like to make sure we don't block the usage of these alternative backends until we can be feature parity.

We'll have the static output from trame in the next couple weeks... we could just land this PR after 0.38 so that the deprecation comes with that final feature

akaszynski:
> We'll have the static output from trame in the next couple weeks... we could just land this PR after 0.38 so that the deprecation comes with that final feature

Awesome.

akaszynski (approved):

banesullivan:
with #3909 and soon-to-come static embedding, what do you think @akaszynski? Merge for 0.38 or wait for 0.39?

akaszynski:
> with #3909 and soon-to-come static embedding, what do you think @akaszynski? Merge for 0.38 or wait for 0.39?

I'm good with merging this now. Users can still use the older backends, even without warnings since we suppress those.

larsoner (approved):

PR #3906, opened 2023-01-30
Disable conda CI

banesullivan:
Not sure what's going wrong with mamba on the conda CI job, and we don't have the bandwidth to fix it. Temporarily disabling this job to get status checks to be more insightful while on our way to the 0.38 release this week


banesullivan:
I will be sure to test against a conda VTK build before the release locally to make sure we aren't depending on any modules not build there (that's really the only point of this CI job)

akaszynski (approved):
As long as we check pre-release, I'm good with this.

PR #3909, opened 2023-01-31
Use plotter window_size for trame viewer if set

### LINKED ISSUE #3879: Expose mechanism to easily control the Jupyter trame view size

banesullivan:
Resolves #3879

@larsoner, please review

We actually had this same implementation for the `panel` backend checking if `window_size` was ever set and if not, defaulting to something nice for Jupyter.

As a bonus, I added an advanced opt-in mechanism for building your own widget/iframe/whatever by passing a callable that accepts the trame viewer URL

```py
import pyvista as pv

p = pv.Plotter()
p.window_size = 500, 350
p.add_mesh(mesh)
p.show()
```
Screen Shot 2023-01-30 at 5 15 01 PM

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -6251,7 +6251,8 @@ class Plotter(BasePlotter):
         self._window_size_unset = False
         if window_size is None:
             self.window_size = self._theme.window_size
-            self._window_size_unset = True
+            if self.window_size == pyvista.themes.DefaultTheme().window_size:
+                self._window_size_unset = True
         else:
             self.window_size = window_size
 
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -1591,6 +1591,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
     def window_size(self, window_size):
         """Set the render window size."""
         self.render_window.SetSize(window_size[0], window_size[1])
+        self._window_size_unset = False
         self.render()
 
     @contextmanager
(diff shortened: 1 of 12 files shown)


larsoner (dismissed):
Nice, thanks for the quick fix @banesullivan !

Feel free to merge when green unless you want others to look

pyvista/plotting/plotting.py
```diff
@@ -6253,7 +6248,12 @@ def on_timer(iren, event_id):
         self.set_background(self._theme.background)
 
         # Set window size
-        self.window_size = window_size
+        self._window_size_unset = False
+        if window_size is None:
+            self.window_size = self._theme.window_size
+            self._window_size_unset = True
```
banesullivan, line 6254: My only last minute thought is to add a check here that will verify the `window_size` matches that of `DefaultTheme.window_size` this would allow users to set the `window_size` in the theme and have that control the Tram Jupyter widget size.

```suggestion
            if self.window_size == pyvista.themes.DefaultTheme().window_size
                self._window_size_unset = True
```

thoughts, @larsoner?

Though this might mess up the testing as the `TestingTheme` sets a different window size

banesullivan: Done

PR #3918, opened 2023-01-31
Try numpydoc_show_inherited_class_members=False

### LINKED ISSUE #3546: Inherited methods are documented too many times

banesullivan:
An attempt to speed up the docs build and resolve #3546 

See also:

- https://github.com/pyvista/pyvista/issues/3546#issuecomment-1410715396
- https://numpydoc.readthedocs.io/en/latest/install.html#configuration

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -379,6 +379,7 @@ import re
 # -- .. pyvista-plot:: directive ----------------------------------------------
 from numpydoc.docscrape_sphinx import SphinxDocString
 
+# Hide inherited methods to reduce duplicate example execution
 numpydoc_show_inherited_class_members = False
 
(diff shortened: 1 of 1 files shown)


banesullivan:
This had no effect on build time, and I downloaded the artifact to find it does not solve #3546

PR #3919, opened 2023-01-31
Ignore inherited members and attributes and switch to the full doc build

akaszynski:
Get rid of inherited attributes and method in our doc build autosummary.


akaszynski:
Style and Docstring Check / docstringcheck (pull_request)  is failing:
https://askubuntu.com/questions/1265388/problem-with-apt-get-update-on-ubuntu-20-04-lts

Going to look the other way for this.

banesullivan (dismissed):
So needed! Some old links may break but it's worth it to clean up the search interface and not re-run the same examples for every DataSet subclass

banesullivan (commented):
Looks like attributes are still documented for every subclass:

Screen Shot 2023-01-31 at 3 58 38 PM


But OH MY! Does this make finding relevant filters so much better. For example, trying to find ImageData/UniformGrid-specific filters:
Screen Shot 2023-01-31 at 3 59 54 PM

akaszynski:
Trying to fix the attributes issue.

akaszynski:
Attributes no longer duplicated:
tmp

banesullivan (approved):
Awesome!!

PR #3927, opened 2023-02-01
[HOTFIX] Track is_set on Camera rather than Renderer

### LINKED ISSUE #3925: Linked views camera setting regression

The initial camera position in this example isn't right: https://dev.pyvista.org/examples/02-plot/linked.html#sphx-glr-examples-02-plot-linked-py

sphx_glr_linked_001

```python
Insert code here
```

### System Information

```shell
main (works on 0.37)
```

### Screenshots

_No response_

banesullivan:
Resolves #3925, follow up to #3743

This isn't perfect, but it fixes the linking issue. Need to get this in before the 0.38 release.

On this branch, I built the docs locally and used `pv.compare_images` to verify that over 80% sphinx gallery examples are indeed the same as the 0.37 release (a massive boost of confidence in this 0.38 release). That remaining 20% seems to be because some examples changed, and some examples in the published docs are simply broken. I'm continuing to peruse the docs to look for discrepancies, but after this, we should be good

I did this by cloning https://github.com/pyvista/pyvista-docs and comparing the images in the `_images` directory with what is built localling in `doc/_build/html/_images/`

@akaszynski, would you please double-check this? This whole `is_set` API isn't ideal, and there are still remaining issues with `reset_camera()` not being aware of the data bounds within each linked renderer, but this gets us >90% there and is better than being totally broken.


akaszynski (approved):
We’re going to better track regressions to prevent this.

As for the API change, I actually think this PR simplified the API and makes it more explicit.

LGTM. Thanks for fixing it.

PR #3934, opened 2023-02-01
Document BasePlotter

banesullivan:
This is not documented after #3919


pyvista/plotting/plotting.py
```diff
@@ -182,7 +182,19 @@ class BasePlotter(PickingHelper, WidgetHelper):
         Width of the border in pixels when enabled.
 
     title : str, optional
-        Window title of the scalar bar
+        Window title of the scalar bar.
```
banesullivan, line 185:
```suggestion
        Window title.
```

PR #3938, opened 2023-02-02
Drop Python 3.7

mwtoews:
According to [PEP 537](https://peps.python.org/pep-0537/), Python 3.7 will stop maintenance in June 2023. NumPy's [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) dropped Python 3.7 in 2021. So it's time to drop it here too, and replace it with Python 3.11.

This PR does the following:

- Remove `typing-extensions` dependencies
- Adjust python versions in various CI workflows
- Update docs and example code
- Update versions for Trove classifiers
- Also clean-up Python 3.6 logic in pyvista/utilities/cells.py


doc/getting-started/installation.rst
```diff
@@ -3,7 +3,7 @@
 Installation
 ============
 
-PyVista is supported on Python versions 3.7+. Previous versions of Python are
+PyVista is supported on Python versions 3.8+. Previous versions of Python are
 no longer supported as outlined in `this issue`_.
 
 .. _this issue: https://github.com/pyvista/pyvista/issues/164
```
banesullivan, line 9: We should remove the reference to this issue. It's dated.

mwtoews:
should dropping VTK<9 also be included in this PR, or separately? I don't mind either way.

banesullivan:
> should dropping VTK<9 also be included in this PR, or separately? I don't mind either way.

It's impossible (or difficult rather) to use VTK<9 after dropping Python 3.7 support, so yes, IMO, we should remove all VTK8 support in PyVista when dropping support for Python 3.7. This would clean up a lot of "if VTK9: do this else: do that" logic, lazy imports, and a bit of other cruft.

@mwtoews, considering that. I want to encourage you to hold off for a bit as this will be touching a good chunk of the codebase and then may not get support/approval to merge for a while -- leaving you with a stale PR that will be tough to keep up to date.

I'm +1 for doing this months ago, but I want to make sure dropping Python 3.7 and VTK8 support has buy-in. 

I propose opening an issue to discuss this topic and, when a date is agreed upon, revisit this

banesullivan:
FYI, Python 3.11 is currently supported though not tested on CI. I'm going to add these CI workflows with the latest version of VTK (9.2.5) soon

akaszynski:
> It's impossible (or difficult rather) to use VTK<9 after dropping Python 3.7 support, so yes, IMO, we should remove all VTK8 support in PyVista when dropping support for Python 3.7. This would clean up a lot of "if VTK9: do this else: do that" logic, lazy imports, and a bit of other cruft.

I feel that dropping support for Python3.7 just means that users will be forced to use an older version of PyVista, and that's totally ok. It's opt-in and lets us keep up-to-date. If user really, absolutely need a patch fix for an older version, we can backport it, but I'm going to heavily discourage that.

Let's let the general community weigh in, but I'm happy if we want to drop support for Python 3.7 in `pyvista==0.39.0` *and* drop support for VTK<9.0 (in a separate PR but in the same release).

banesullivan:
> drop support for VTK<9.0 (in a separate PR but in the same release).

I feel VTK<9 and Python 3.7 should be dropped simultaneously.

mwtoews:
My personal take is to try and maintain no more than 4 concurrent Python minor versions. These minor versions are generally anticipated [each year](https://en.wikipedia.org/wiki/History_of_Python) (with a few exceptions), which is why the intention of this PR was to keep the number of CI workflow jobs the same (rather than increasing the CI load further). I'll pause on this PR until a consensus has been met, but would be happy to help out further when the timing is right.

MatthewFlamm:
It is starting to get very close to the EOL for 3.7 in terms of # of possible releases here.  I don't think it necessarily makes much sense to support 3.7 if it is only recommended to use for a few months.  Users of 3.7 can still use older versions of pyvista.  By the time of the next release it will easily be within 4 months from EOL, which seems like a very reasonable time frame to me.  I think numpy-style removing support 1.5+ years in advance was a bit quick.  I guess I'm saying that I'm +1 for this change or anytime in the next few months.

I don't think there are significant features in 3.8 that will greatly improve the code here, so there is no big maintenance reason from python to remove support. The biggest gain is the push to remove vtk8.  I'm surprised that we support multiple major versions of vtk here.  I doubt vtk supports changes to vtk 8 at this point, but I haven't checked.  While it currently works, it makes certain parts of the code complicated and overall lowers the velocity of changes.  We currently turn off huge portions of testing for vtk8 so it feels a bit like it already isn't supported in some ways.

I think it will be easier from a PR perspective to make these changes separately but in the same release.  This will also help as continuing the blueprint for future Python version removals.  My thought is that it will be significantly more involved to remove vtk8, so it makes sense to do this first, as it gives us the possibility to postpone it by a release in case it takes time.

banesullivan:
> try and maintain no more than 4 concurrent Python minor versions

I want to share some feedback from @prisae that I agree with:

> Given that Python changed after 3.7 from a ~1.5 year release-cadence to a ~1 year cadence, one might want to test one version more. Hence 3.7-3.11 now, later 3.8-3.12, etc.

banesullivan:
I've updated CI to use Python 3.11 in #3944. @mwtoews, you'll need to rebase this PR and remove any 3.11-related changes

tests/test_filters.py
```diff
@@ -17,10 +16,6 @@
 
 normals = ['x', 'y', '-z', (1, 1, 1), (3.3, 5.4, 0.8)]
 
-skip_py2_nobind = pytest.mark.skipif(
-    int(sys.version[0]) < 3, reason="Python 2 doesn't support binding methods"
-)
-
```
banesullivan, line 23: Oh wow, thank you for finding this... makes me wonder what other Python2 cruft we still have

mwtoews: This should be it for the cases easily identified by various `git grep` patterns to find where version checks are used. More complex try/except blocks are trickier to identify, but these should be harmless.

mwtoews:
If it also makes sense to also remove VTK8 in this PR, I can re-title and add subsequent commits to unravel this logic.

banesullivan:
> If it also makes sense to also remove VTK8 in this PR, I can re-title and add subsequent commits to unravel this logic.

IMO, the Python 3.7 drop and VTK8 drop should come as one PR and thus one commit on main but it looks like others have different views

I say charge ahead with a simultaneous Python3.7 and VTK8 removal, as that'll get my approval to merge but fair warning that others here might want to handle this differently.

I want to re-emphasize that this might not get merged for a while, and after removing VTK8 support, the diff will likely become tough to manage with other development

akaszynski:
> I think it will be easier from a PR perspective to make these changes separately but in the same release. This will also help as continuing the blueprint for future Python version removals. My thought is that it will be significantly more involved to remove vtk8, so it makes sense to do this first, as it gives us the possibility to postpone it by a release in case it takes time.

I'd like to drop 3.7 support first and then VTK 8 secondly. I agree with @MatthewFlamm in regards to providing a blueprint for future PRs. It also makes it easier to follow the blame.

banesullivan:
Fair enough! I'm on board as long as the VTK8 drop comes first

akaszynski:
@mwtoews, does that sound reasonable from your end?

mwtoews:
Good plan from my perspective, I'll hold this PR to only drop Python 3.7 or earlier logic. I don't mind rebasing this after dropping VTK8. However, it seems to me that a core dev should take a lead on that PR, as I'm not as familiar how different VTK releases are integrated with pyvista. Any takers?

akaszynski:
> I'd like to drop 3.7 support first and then VTK 8 secondly. I agree with @MatthewFlamm in regards to providing a blueprint for future PRs. It also makes it easier to follow the blame.

Looks like we did this the other way around with the merge of #4018, but I'm still fine with moving forward with this.

If there are no further issues with dropping Python 3.7 in `main` little early before it's [EOL](https://endoflife.date/python), let's merge this. When release time comes in May/June, we'll be right on time for its EOL.

akaszynski (dismissed):

pyvista/core/dataset.py
```diff
@@ -2,15 +2,9 @@
 
 import collections.abc
 from copy import deepcopy
-import sys
-from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple, Union, cast
+from typing import Any, Dict, Iterable, List, Literal, Optional, Tuple, Union, cast
```
akaszynski, line 5:
```suggestion
from typing import Any, Dict, Iterable, Iterator, List, Literal, Optional, Tuple, Union, cast
```

MatthewFlamm (approved):
LGTM

Issue #3939, opened 2023-02-02
Trame is needed per import statements


RobPasMue:
Getting the following error when installing ``pyvista`` on Python 3.9 and Windows:

```bash
  File "C:\...\.venv\lib\site-packages\pyvista\__init__.py", line 9, in <module>
    from pyvista import trame
  File "C:\...\.venv\lib\site-packages\pyvista\trame\__init__.py", line 2, in <module>
    from pyvista.trame.jupyter import launch_server, show_trame, elegantly_launch
  File "C:\...\.venv\lib\site-packages\pyvista\trame\jupyter.py", line 8, in <module>
    from trame.app import get_server
ModuleNotFoundError: No module named 'trame'
```

Looks like trame is not a required dependency but due to the import statements it is failing. You might want to add it to the pyproject.toml or handle properly those import statements with some kind of try/except.

germa89:
Interestingly ... https://github.com/pyvista/pyvista/issues/3941

florian98765:
Same for me. I would prefer the try/except version in order to avoid additional dependencies.

banesullivan:
0.38.1 should be out in the next 15 minutes -- sorry this slipped through the cracks!

PR #3942, opened 2023-02-02
Fixing missing imports

### LINKED ISSUE #3941: Missing ``ipywidgets`` package

After installing the last version, I get:

```pycon
Python 3.11.1 (main, Dec 23 2022, 09:28:24) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import pyvista
  File "/Users/german.ayuso/trashy/venvtrash/lib/python3.11/site-packages/pyvista/__init__.py", line 9, in <module>
    from pyvista import trame
  File "/Users/german.ayuso/trashy/venvtrash/lib/python3.11/site-packages/pyvista/trame/__init__.py", line 2, in <module>
    from pyvista.trame.jupyter import launch_server, show_trame, elegantly_launch
  File "/Users/german.ayuso/trashy/venvtrash/lib/python3.11/site-packages/pyvista/trame/jupyter.py", line 7, in <module>
ModuleNotFoundError: No module named 'ipywidgets'
```

Using Mac OS (M2-Ventura 13.1)

Just update and import pyvista

```py
import pyvista
```

### System Information

```shell
NA
```

### Screenshots

_No response_

germa89:
### Overview

Adding missing imports.

Resolves #3941 and #3939

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,11 @@ dependencies = [
     'scooby>=0.5.1',
     "typing-extensions; python_version < '3.8'",
     'vtk',
+    'ipywidgets',
+    'trame>=2.2.6',
+    'trame-client>=2.4.2',
+    'trame-server>=2.8.0',
+    'trame-vtk>=2.0.17',
 ]
 dynamic = ['version']
 
@@ -54,7 +59,6 @@ colormaps = [
 io = ['meshio>=5.2']
 jupyter = [
     'ipyvtklink',
-    'ipywidgets',
     'jupyter-server-proxy',
     'nest_asyncio',
     'panel',


pyproject.toml
```diff
@@ -33,6 +33,11 @@ dependencies = [
     'scooby>=0.5.1',
     "typing-extensions; python_version < '3.8'",
     'vtk',
+    'ipywidgets',
+    'trame>=2.2.6',
+    'trame-client>=2.4.2',
+    'trame-server>=2.8.0',
+    'trame-vtk>=2.0.17',
 ]
```
RobPasMue, line 41: The problem with this approach is that you end up increasing the needed dependencies to run PyVista. Do you really need ``ipywidgets`` or ``trame`` to run the typical use cases? The first one is mostly for Jupyter-related views. The same happens with ``trame`` AFAIK. They should really be optional dependencies, and the imports should be handled with caution by checking whether the module is available or not when doing the import statement (try/except approach).

But that's my gut feeling. Pinging @akaszynski @banesullivan and @AlejandroFernandezLuces for their thoughts.

germa89: I totally agree. In fact, I was surprised to see those missing dependencies are in the optional dependencies requirements. Hence fixing this is more complex than I expected. There is something missing in the library structure.

RobPasMue: Something like this...

```py
try:
    import docker
    from docker.errors import APIError, ContainerError, ImageNotFound
    from docker.models.containers import Container


    _HAS_DOCKER = True
except ModuleNotFoundError:  # pragma: no cover
    _HAS_DOCKER = False

```

RobPasMue: But of course... applied to the modules involved in this issue 😄 not ``docker``

banesullivan: We actually want to remove `pyvista.trame` from the top level namespace of the package to avoid this issue rather than adding additional imports or "has package" logic.

See https://github.com/pyvista/pyvista/pull/3943

RobPasMue: Sounds good to me @banesullivan!

PR #3943, opened 2023-02-02
[HOTFIX] Resolve trame import issues

### LINKED ISSUE #3939: Trame is needed per import statements

### LINKED ISSUE #3941: Missing ``ipywidgets`` package

banesullivan:
Close #3942, resolves #3941, resolves #3939

Will backport to `release/0.38`

cc @RobPasMue and @germa89

The trame module is opt-in, yet it was added to the top-level imports due to a numpydoc/autosummary issue in #3909.

This is a hotfix to just leave that API undocumented and fix the import issue. We'll have to look into the documentation issue separately.

Further, I added a CI step to make sure the we verify PyVista does not break on import/plot when only installed with it's core dependencies

### DIFF
--- a/doc/api/plotting/trame.rst
+++ b/doc/api/plotting/trame.rst
@@ -2,7 +2,7 @@
 
 Trame
 -----
-The PyVista :mod:`pyvista.trame` module allows users to access the `Trame
+The PyVista ``pyvista.trame`` module allows users to access the `Trame
 <https://kitware.github.io/trame/index.html>`_ widget from PyVista to create
 web-based 3D visualizations. This allows you to access the VTK pipeline using
 the PyVista API so you can pair PyVista and Trame so that PyVista plotters can
@@ -10,13 +10,8 @@ be used in a web context with both server and client-side rendering.
 
 For the full user guide, see :ref:`trame_jupyter`.
 
-.. currentmodule:: pyvista.trame
 
-.. autosummary::
-   :toctree: _autosummary
-
-   launch_server
-   show_trame
-   elegantly_launch
-   get_or_create_viewer
-   plotter_ui
+.. warning::
+   We are having an issue with Sphinx and numpydoc documenting this module
+   because it is not in the top level namespace of the PyVista package.
+   We will fix this later
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -179,6 +179,11 @@ jobs:
           packages: libgl1-mesa-glx xvfb python-tk
           version: 3.0
 
+      # Make sure PyVista does not break from non-core dependencies
+      - name: Software Report (Core Dependencies)
+        run: |
+          xvfb-run python -c "import pyvista; print(pyvista.Report());
+
       - name: Install Testing Requirements
         run: pip install -r requirements_test.txt
 
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -182,7 +182,7 @@ jobs:
       # Make sure PyVista does not break from non-core dependencies
       - name: Software Report (Core Dependencies)
         run: |
-          xvfb-run python -c "import pyvista; print(pyvista.Report());
+          xvfb-run python -c "import pyvista; print(pyvista.Report());"
 
       - name: Install Testing Requirements
         run: pip install -r requirements_test.txt
--- a/pyvista/__init__.py
+++ b/pyvista/__init__.py
@@ -6,7 +6,6 @@ from typing import Optional
 import warnings
 import os
 
-from pyvista import trame
 from pyvista._version import __version__
 from pyvista.plotting import *
 from pyvista.utilities import *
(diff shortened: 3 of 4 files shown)


RobPasMue (approved):
LGTM @banesullivan thanks for the quick fix!

tkoyama010 (approved):

tkoyama010:
@banesullivan Are we releasing `pyvista==0.38.1` ?

banesullivan:
Yes, 0.38.1 will be on its way right after this merge (so give me 30ish minutes)

larsoner (approved):
Should also fix stuff in `pyvistaqt`, thanks for the quick fix @banesullivan !

larsoner:
... also this makes me think somewhere in the CI matrix there should be a "minimal" run where tests are run with minimal dependencies. In MNE we have this and also an "old" CI run with the oldest supported versions. They routinely catch stuff like this that we would have otherwise missed

banesullivan:
@larsoner, do you have any insight on if this was a numpydoc error:

```
Extension error (sphinx.ext.autosummary):
Handler <function process_generate_options at 0x7f4eca95ddc0> for event 'builder-inited' threw an exception (exception: no module named pyvista.trame)
make: *** [Makefile:50: html] Error 2
```

(see the diff here: after removing `trame` from the top-level namespace, autosummary couldn't find it)

banesullivan:
> ... also this makes me think somewhere in the CI matrix there should be a "minimal" run where tests are run with minimal dependencies. In MNE we have this and also an "old" CI run with the oldest supported versions. They routinely catch stuff like this that we would have otherwise missed

I added a very, very minimal version of this here -- we could do better though. What I added generates a "Report" which actually runs a plot routine under the hood, so it does execute a good chunk of the codebase

larsoner:
Where is the error? The doc CI run I see is green

https://github.com/pyvista/pyvista/actions/runs/4075603126/jobs/7022221690

And I don't see any warnings in the log. Did I miss them?

doc/api/plotting/trame.rst
```diff
@@ -2,21 +2,16 @@
...
-
-   launch_server
-   show_trame
-   elegantly_launch
-   get_or_create_viewer
-   plotter_ui
+.. warning::
+   We are having an issue with Sphinx and numpydoc documenting this module
+   because it is not in the top level namespace of the PyVista package.
+   We will fix this later
```
banesullivan, line 17: @larsoner, this is what I was referring to. This breaks when not having `trame` in the top-level namespace. So I just removed it for now to get a patch release out. Having this would cause the error

banesullivan: Sorry to bug you with this... the line between autosummary and numpydoc is blurry to me and thought you might have seen this before

larsoner: No I haven't, and I know `scipy` does not import most of its namespaces (e.g., `scipy.signal`) when you just do `import scipy` so it must be possible to use numpydoc this way. One possibility is that the import fails on the doc build, but it looks like you didn't change the requirements so that doesn't make a lot of sense.

banesullivan: Exactly, this sort of namespacing is super common. Banged my head on the wall over this and wasn't sure what to do so I guess we'll just leave this API undocumented for now

banesullivan: Thanks for taking a look!

PR #3944, opened 2023-02-02
Test with Python 3.11

banesullivan:
@mwtoews, this handles adding Python 3.11 to CI to supersede some of your changes in #3938

Adds CI testing for Python 3.11 and attempts to speed up the docs build with 3.11

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -36,7 +36,7 @@ jobs:
 
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.9'
+          python-version: '3.11'
           cache: 'pip'
           cache-dependency-path: |
             requirements.txt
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -47,18 +47,21 @@ jobs:
...
-
       - name: Install PyVista and dependencies
         run: |
           pip install -e . --no-deps
           pip install -r requirements_docs.txt
 
+      - name: Install custom OSMesa VTK variant
+        run: |
+          pip uninstall vtk -y
+          pip install https://github.com/pyvista/pyvista-wheels/raw/main/vtk_osmesa-9.2.5-cp311-cp311-linux_x86_64.whl
```
banesullivan, line 58: @akaszynski, @MatthewFlamm, @AlejandroFernandezLuces: this is the issue I am referring to in https://github.com/pyvista/pytest-pyvista/pull/41

.github/workflows/docs.yml
```diff
@@ -174,17 +175,6 @@ jobs:
...
-      - name: Publish notebooks on main
-        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
-        uses: peaceiris/actions-gh-pages@v3
-        with:
-          external_repository: pyvista/pyvista-examples
-          personal_token: ${{ secrets.PYVISTA_BOT_TOKEN }}
-          publish_dir: pyvista-examples
-          publish_branch: develop
-          exclude_assets: ''
-          force_orphan: true
```
banesullivan, line 186: These are pushing to the same repo as the release and thus overwriting the examples from the release. The docker image there uses the `latest` tag so it's very likely that the examples will mismatch the PyVista version there

.github/workflows/docs.yml
MatthewFlamm: I think there might be other issues.  For example here is the relevant lines from the ongoing doc build.  Note that vtk isn't installed by `pip install -e . --no-deps` as expected.  the package from pyvista-wheels is `vtk-osmesa`.

```txt
Run pip uninstall vtk -y
WARNING: Skipping vtk as it is not installed.
Collecting vtk-osmesa==9.2.5
  Downloading https://github.com/pyvista/pyvista-wheels/raw/main/vtk_osmesa-9.2.5-cp311-cp311-linux_x86_64.whl (76.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 76.1/76.1 MB 38.6 MB/s eta 0:00:00
Installing collected packages: vtk-osmesa
Successfully installed vtk-osmesa-9.2.5
```

banesullivan: Sorry. I commented on the wrong spot. `pytest-pyvista` isn't used here 🤦🏻‍♂️

banesullivan: See https://github.com/pyvista/pyvista/pull/3704/files#r1094927082

banesullivan:
There are pending status checks because some of these are set as required in the repo settings. If this merges, I will update the required checks to the new jobs added here

akaszynski (approved):

PR #3945, opened 2023-02-02
Rework render event callbacks

dcbr:
### Overview

Rework of the render event callbacks, merging the functionality added in #3385 and #3756.

cc @banesullivan, your thorough review would be appreciated.

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -299,7 +299,8 @@ class BasePlotter(PickingHelper, WidgetHelper):
         # track if render window has ever been rendered
         self._rendered = False
 
-        self._on_render_callbacks = set()
+        self._render_callbacks = {hook: set() for hook in ("before", "after")}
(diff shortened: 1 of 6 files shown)


pyvista/plotting/plotting.py
```diff
@@ -1781,38 +1782,50 @@ def render(self):
...
         if self.render_window is not None and not self._first_time and not self._suppress_rendering:
             log.debug('Rendering')
+            for callback in self._render_callbacks["before"]:
+                callback(self)
             self.renderers.on_plotter_render()
             self.render_window.Render()
             self._rendered = True
-        for callback in self._on_render_callbacks:
-            callback(self)
+            for callback in self._render_callbacks["after"]:
```
dcbr, line 1790: Just noticed that in the description of this method you wrote the callbacks are always executed, why is this necessary?
I made the `on_plotter_render` method a proper callback in this PR, but it only makes sense to call it before an actual render. That's why I changed this, but maybe it's ok to revert, I'll check.

PR #3946, opened 2023-02-02
Speed up PVDReader for files with many time points

### LINKED ISSUE #3937: PVDReader slow when handling files with large number of timepoints
I have a PVD file with over 10,000 time points and multiple parts in each time point.  So ballpark 50,000 line PVD file.  I iterate over each time point and run an analysis, which was running really slowly.

image

The first list comprehension below takes the majority of the execution time (code run from a released version, github linked from the main branch so the method names and line numbers don't match):

https://github.com/pyvista/pyvista/blob/088305e07f19e1f8e7ca8acdf8bec666b7afc9be/pyvista/utilities/reader.py#L1867-L1875

MatthewFlamm:
### Overview

Closes #3937.  Adds in a mapping with keys of time that stores datasets.  This extremely reduces the time required to iterate over a PVDfile with many time points.


akaszynski (approved):
Sorry for the delayed review @MatthewFlamm, I'll get better about reviewing these.

LGTM. Didn't even see the GIF there in the docstr.

PR #3947, opened 2023-02-03
[WIP] iadd/isub methods

banesullivan:
Would anyone be interested in having `+=` and `-=` operators on the `Plotter` class? For this sort of logic:

```py
import pyvista as pv

pl = pv.Plotter()

pl += pv.Sphere()
pl += pv.Cone()

actor = pl.add_mesh(pv.Cube())
pl -= actor

pl.show()
```

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -6054,6 +6054,28 @@ class BasePlotter(PickingHelper, WidgetHelper):
                 places.append(tuple(self.renderers.index_to_loc(index)))
         return places
 
+    def add(self, to_add, *args, **kwargs):
+        if isinstance(to_add, _vtk.vtkActor):
+            return self.add_actor(to_add, *args, **kwargs)
+        if isinstance(to_add, _vtk.vtkMultiBlockDataSet):
+            return self.add_composite(to_add, *args, **kwargs)
+        if isinstance(to_add, _vtk.vtkDataSet):
+            return self.add_mesh(to_add, *args, **kwargs)
+        if isinstance(to_add, str):
+            return self.add_text(to_add)
+        # TODO: iterables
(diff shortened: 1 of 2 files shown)


adeak:
I have some practical and semantic concerns.

The practical concern is that just "adding" a mesh is never what I do, there's always customisation going into my `add_mesh()` calls. And there's usually enough variation in meshes to plot that theming isn't a solution. So I'd find it rare that a bare `add_mesh(obj)` call would suffice.

The semantic concern is that `__iadd__()` should be analogous to `__add__()`. But here we don't have `__add__()` and for a good reason: `plotter + mesh` is nonsense, so `plotter = plotter + mesh` (which is not the same but semantically close to `plotter += mesh` is dubious. I'd be willing to overlook this issue if there was a good pragmatic reason or common use case for abusing the syntax, but I don't yet see that this would be useful.

(My usual disclaimer of not being a regular PyVista end user applies :P)

MatthewFlamm:
I agree with @adeak here.  You get `__iadd__` for free when you define `__add__`.  This further strengthens the point about this meaning `plotter = plotter + mesh`, which is confusing.

adeak:
> I agree with @adeak here. You get `__iadd__` for free when you define `__add__`.

The subtlety is that the automatic fallback on `__add__()` means that `plotter += mesh` could give you a new plotter depending on the definition of `__add__()`, so there would be value in defining `__iadd__()`. Of course in our case of `plotter + mesh` we'd be mutating the input plotter anyway.

banesullivan:
Thank you both for this input! I agree this is would be a bit confusing to the end user and has too many different scenarios.

I opened this PR out of curiosity and I'm going to close it since it doesn't have any clear benefit but clear cost of as an obfuscated API.

Also agree strongly with the fact that use almost always need to specify some option to `add_mesh()`.

Really the only way I see this making sense is if it were limited to Actors

PR #3966, opened 2023-02-06
Patch contextmanagers to handle exceptions

banesullivan:
Follow up to #3897 to address https://github.com/pyvista/pyvista/pull/3897#discussion_r1097520833

@larsoner, please review

I tried to test this with pytest but had an issue raising an error with `pytests.raises()` and still having the error propagate to the contextmanager.

We should backport this onto `release/0.38` for the `0.38.2` release whenever that needs to come

### DIFF
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -1671,11 +1671,13 @@ class BasePlotter(PickingHelper, WidgetHelper):
         size_before = self.window_size
         if window_size is not None:
             self.window_size = window_size
-        yield self
-        # Sometimes the render window is destroyed within the context
-        # and re-setting will fail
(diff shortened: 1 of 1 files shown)


larsoner (approved):
> I tried to test this with pytest but had an issue raising an error with pytests.raises() and still having the error propagate to the contextmanager.

You could do something like
```
try:
    bad_thing()
except RuntimeError:
    pass
else:
    pytest.fail('did not raise expected exception')
```

But even without this changes LGTM

larsoner (approved):

banesullivan:
FYI, this has been backported to `release/0.38` in 0dce2925b235a164c28d4cf1b4a62263e264e8ad

PR #3967, opened 2023-02-06
MAINT: Add pyvistaqt integration test

larsoner:
Seems like a useful CI run to have

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -65,16 +65,14 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '3.10'
-      - uses: awalsh128/cache-apt-pkgs-action@v1.1.3
-        with:
-          packages: libgl1-mesa-glx xvfb
-          version: 3.0
       - name: Install PyVista
         run: pip install -ve .
       - name: Install Qt and testing dependencies
         run: pip install PyQt6 pytest pytest-qt pytest-cov
       - name: Clone PyVistaQt
         run: git clone --depth=1 https://github.com/pyvista/pyvistaqt.git --branch main --single-branch
+      - name: Setup Xvfb and Qt dependencies
+        run: ./pyvistaqt/.ci/setup_headless_display.sh
       - name: Install PyVistaQt
         run: pip install -ve .
         working-directory: pyvistaqt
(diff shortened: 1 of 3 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -54,8 +54,33 @@ jobs:
...
       - run: pytest --cov=pyvista --cov-report=xml mne/viz/_brain mne/viz/tests/test_3d.py mne/viz/backends
         working-directory: mne-python
-      # In theory this should be fine, too (codecov should combine all reports)
-      # if we copied the cov report up one directory
-      #- uses: codecov/codecov-action@v3
-      #  name: 'Upload coverage to CodeCov'
-      #  directory: ./mne-python/
+      - run: cp -a mne-python/*.xml .
+      - uses: codecov/codecov-action@v3
+        name: 'Upload coverage to CodeCov'
```
banesullivan, line 59: I am confused on why we are uploading coverage and where this goes...

larsoner: See deleted comments above this change. Basically these integration tests still checkout `pyvista` as their first step. As a later step, they checkout some other project in `pyvista/mne-python` for example. Then pytests are rune in `pyvista/mne-python/` with the `--cov pyvista` option, i.e., coverage is tracked for the `pyvista` module, not `mne` (or `pyvistaqt`). So the `.xml` in `pyvista/mne-python/whatever_coverage.xml` should have PyVista code coverage in it. By coping it to the parent `pyvista/` directory, then running codecov (which will occur in `pyvista/` as usual) we should get coverage stats for `pyvista` module added to the reports codecov already aggregates in other CI runs.

larsoner: By example, here's the codecov step for MNE integration:

https://github.com/pyvista/pyvista/actions/runs/4106727616/jobs/7085318565

and it gets the codecov result URL

https://app.codecov.io/github/pyvista/pyvista/commit/05270e33f48103462022ffcaabe02995e42949ad
https://app.codecov.io/github/pyvista/pyvista/commit/05270e33f48103462022ffcaabe02995e42949ad/tree/pyvista

So I think it might be working... let's see what it says once everything comes back green

banesullivan: Ah I missed the `--cov=pyvista` option. Yep this should work

larsoner:
This appears to be headed to green, so I pushed an empty commit to confirm things work here after the merge of https://github.com/pyvista/pyvistaqt/pull/306 (they should, but better safe than sorry!)

banesullivan (approved):

PR #3968, opened 2023-02-06
Fix Plotter docs

### LINKED ISSUE #3956: Missing docs for Plotter inherited classes
We are missing docs for the base classes of `BasePlotter`:

- `PickingHelper`
- `WidgetHelper`

We should consider making an exception to the exclusion of inherited methods specifically for the `Plotter` class.

banesullivan:
~An attempt to resolve~ Resolves #3956 and #4023


akaszynski:
Should we add a check for broken intersphinx docstrings? These are hard to catch manually.

akaszynski:
I'd like to have this merged today for our 0.38.4 release. As this doesn't change the API and really improves the docs, a cursory review would be appreciated but I'll press without it and beg for forgiveness.

akaszynski:
Alright, except for a few exceptions, fixed all the intersphinx and role links.

PR #3992, opened 2023-02-09
Pass texture coordinates when wrapping Trimesh object

Kiord:
Texture coordinates from `trimesh.Trimesh` are not passed to the `pyvista.PolyData` when using `pyvista.wrap(.)`.

This PR adds two lines of code to check if the `Trimesh` object has uv (if `Trimesh.visual.uv` exists) and pass them as `active_t_coords` of the resulting `PolyData`.


akaszynski (approved):
Awesome. Nice PR.

akaszynski:
@banesullivan, please backport this for 0.38.2

PR #4011, opened 2023-02-13
Add .codacy.yml to exclude examples and tests

tkoyama010:
### Overview

image
The CodeQuality in PyVista's Codacy is B, but this is due to judging the sources in the example and test folders. This PR adds a Codacy configuration file to ignore those folders.

 

### Details

- [Codacy configuration file](https://docs.codacy.com/repositories-configure/codacy-configuration-file/)

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,0 +1,7 @@
+---
+exclude_paths:
+  - "examples/**/*"
+  - "examples_flask/**/*"
+  - "examples_trame/**/*"
+  - "tests/*"
+  - "tests/**/*"


.codacy.yml
```diff
@@ -0,0 +1,7 @@
+---
+exclude_paths:
```
banesullivan, line 2: I wish there was an `include_paths` option instead as we'll have to make sure to update this anytime another `examples_*` folder is added. Otherwise, looks great


https://docs.codacy.com/repositories-configure/codacy-configuration-file/

banesullivan (approved):

PR #4012, opened 2023-02-13
Remove unused argument categories

tkoyama010:
### Overview

Remove unused argument categories.

 

### Details

- https://github.com/pyvista/pyvista/pull/2941


banesullivan (commented):
I want to look into this a bit further and make sure its safe to remove. We may have just forgotten to pass it along where it is needed

akaszynski (approved):
This is good to remove as we're not using it. I've added a helpful deprecation warning and added a deprecation note to the docstring.

Thanks @tkoyama010

PR #4018, opened 2023-02-14
Set VTK 9.0 as the minimum supported version

mwtoews:
### Overview

As discussed in #3938 this PR removes pre-VTK 9 logic, setting VTK 9.0 as the minimum supported version.

### Details

- Plenty of chops and simplifications where pre-VTK 9 logic was enabled
- Changes to `pyvista/_vtk.py` are difficult to see due the change of indent level, but are relatively minor
- Keep `pyvista._vtk.VTK9` for now, as this is used externally by pytest-pyvista; remove `pyvista._vtk.VTK91` which had limited use
- From pyvista/utilities/cells.py, remove `generate_cell_offsets_loop()` and `generate_cell_offsets()`, which were only used for pre-VTK9 `UnstructuredGrid()` with four arguments
- With the plotting renderer classess,  remove the `can_process_events` property and other variables like `VERY_FIRST_RENDER` and `update_timer_id`
- Update docstrings and examples, where relevant
- Remove comments, notes or other remarks about previous versions
- Removed the `needs_vtk9` pytest marker


MatthewFlamm:
I started https://github.com/pyvista/pytest-pyvista/pull/44 to enable this PR.

pyproject.toml
```diff
@@ -33,7 +33,7 @@ dependencies = [
     'pooch',
     'scooby>=0.5.1',
     "typing-extensions; python_version < '3.8'",
-    'vtk',
+    'vtk>=9.0.0',
```
banesullivan, line 36: We don't want to add any version constraints on VTK as the VTK version string can be set to all kinds of different things depending on where it came from and we do not want pip to mess with an already installed VTK

pyvista/__init__.py
```diff
@@ -35,11 +35,11 @@
...
 ID_TYPE = _get_vtk_id_type()
 
-# determine if using at least vtk 5.0.0
-if vtk_version_info.major < 5:  # pragma: no cover
+# determine if using at least vtk 9.0.0
+if vtk_version_info.major < 9:  # pragma: no cover
     from pyvista.core.errors import VTKVersionError
 
-    raise VTKVersionError('VTK version must be 5.0 or greater.')
+    raise VTKVersionError('VTK version must be 9.0.0 or greater.')
```
banesullivan, line 42: This is sufficient for the version constraint and why we can not have a constraint in the setup.py/pyproject.toml

banesullivan:
This is fantastic and I am eager to merge this when everyone is on board!

pyvista/utilities/reader.py
```diff
@@ -1392,20 +1392,9 @@ def add_q_files(self, files):
                 files = [files]
         files = [_process_filename(f) for f in files]
 
-        if hasattr(self.reader, 'AddFileName'):
-            # AddFileName was added to vtkMultiBlockPLOT3DReader sometime around
-            # VTK 8.2. This method supports reading multiple q files.
```
MatthewFlamm, line 1397: Given the weakness of the language here, are we sure that it works in vtk versions 9.0+?  Is this in a test?

tests/test_filters.py
```diff
@@ -577,7 +566,6 @@ def test_outline_composite(composite):
     assert isinstance(output, pyvista.PolyData)
     output = composite.outline(nested=True, progress_bar=True)
 
-    # vtk 9.0.0 returns polydata
```
MatthewFlamm, line 580: I can't suggest an edit for some reason, but it looks like this comment might still be relevant.

MatthewFlamm (commented):
Some comments to get started.

pyvista/utilities/reader.py
MatthewFlamm: It looks like there is a test for this, so it should be caught in the tests.

tests/test_filters.py
adeak: (If you hover over the "suggest edit" icon it says "Applying suggestions on deleted lines is not supported.)

adeak: (If you hover over the "suggest edit" icon it says "Applying suggestions on deleted lines is not supported.)

pyproject.toml
mwtoews: good point (https://github.com/conda-forge/vtk-feedstock/issues/234). I'll leave a comment instead.

examples/00-load/create-unstructured-surface.py
```diff
@@ -60,12 +60,7 @@
 points = np.vstack((cell1, cell2)).astype(float)
 
 # create the unstructured grid directly from the numpy arrays
-# The offset is optional and will be either calculated if not given (VTK version < 9),
-# or is not necessary anymore (VTK version >= 9)
-if pv.vtk_version_info < (9,):
-    grid = pv.UnstructuredGrid(offset, cells, cell_type, points)
```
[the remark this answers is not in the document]
mwtoews: Thanks, there were a few other parts of this example to revise w.r.t. offsets.

tests/test_filters.py
mwtoews: I'm not sure if "modern vtk returns polydata" is better, but I'm trying to remove comments that relate only to older versions. Removing version info sort-of helps logically understand that this was older behaviour.

MatthewFlamm: I was assuming that versions >9.0.0 return MultiBlock rather than PolyData, but I don't actually know.  If this indeed is a difference between 8.x and 9.x, we should clean up this test better.  Either way, I think this comment is useful, either to describe why the test is the way it is or as a pointer for future cleanup.

pyvista/plotting/plotting.py
```diff
@@ -6270,10 +6248,6 @@ def on_timer(iren, event_id):
         else:
             self.window_size = window_size
 
-        # add timer event callback to break out of blocking interactive update call (only needed for VTK<9)
-        if not self.iren.can_process_events:
-            self.iren.add_observer(_vtk.vtkCommand.TimerEvent, on_timer)
```
dcbr, line 6275: You can also delete the `on_timer` callback now, as it's not used anywhere else as far as I know.

pyvista/_vtk.py
```diff
@@ -2,772 +2,651 @@
...
 """
 # flake8: noqa: F401
 
-# Checking for VTK9 here even though 8.2 contains vtkmodules.  There
-# are enough idiosyncrasies to VTK 8.2, and supporting it would lead
-# to obscure code.
+from vtkmodules.vtkCommonCore import vtkVersion
 
+# keep legacy variable
+VTK9 = vtkVersion().GetVTKMajorVersion() >= 9
```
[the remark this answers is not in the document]
MatthewFlamm: This was merged, and then when we have a release, we can clean up here.

pyvista/plotting/plotting.py
mwtoews: Thanks. I've left a similar `on_timer` in `tests/test_render_window_interactor.py` alone, as it seems to be different.

MatthewFlamm:
This PR will be increasingly hard to maintain as other PRs are merged.  I recommend merging soon if agreed upon.  I made https://github.com/pyvista/pytest-pyvista/pull/47 for a release then this PR can be finalized.

pyvista/_vtk.py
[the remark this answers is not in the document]
akaszynski: Removed in 2d1a71f99261fa7cfe479e55ceb14ff0db47dad9.

akaszynski (approved):
This is good to go. Thanks @mwtoews.

Let's merge this soon to avoid any potential merge conflicts.

MatthewFlamm (approved):
A forward looking thought: I'm not sure whether it would be now better to `from vtkmodules import ...` where it is used.  `_vtk` now mostly just flattens `vtkmodules`.

akaszynski:
> A forward looking thought: I'm not sure whether it would be now better to from vtkmodules import ... where it is used. _vtk now mostly just flattens vtkmodules.

This is a great point. One issue we've had is that we load everything all upfront despite not using all libraries. If we instead "lazy load" everything, we can improve load time and memory footprint.

PR #4052, opened 2023-02-22
Fix bug of get_data_range method in Table class.

tkoyama010:
### Overview

Fix bug of get_data_range method in Table class.

 

##### main
```python
>>> import pyvista as pv
>>> import numpy as np
>>> nr, nc = 50, 3
>>> arrays = np.random.rand(nr, nc)
>>> table = pv.Table(arrays)
>>> table.get_data_range()
(nan, nan)
```

##### PR
```python
>>> import pyvista as pv
>>> import numpy as np
>>> nr, nc = 50, 3
>>> arrays = np.random.rand(nr, nc)
>>> table = pv.Table(arrays)
>>> table.get_data_range()
(0.00127223358750439, 0.9922766866883922)
```

### Details

- None

### DIFF
--- a/pyvista/core/objects.py
+++ b/pyvista/core/objects.py
@@ -350,7 +350,7 @@ class Table(_vtk.vtkTable, DataObject):
         if isinstance(arr, str):
             arr = get_array(self, arr, preference=preference)
         # If array has no tuples return a NaN range
-        if arr.size == 0 or not np.issubdtype(arr.dtype, np.number):
+        if arr is None or arr.size == 0 or not np.issubdtype(arr.dtype, np.number):
             return (np.nan, np.nan)
         # Use the array range
         return np.nanmin(arr), np.nanmax(arr)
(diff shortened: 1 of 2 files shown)


pyvista/core/objects.py
```diff
@@ -346,11 +346,11 @@ def get_data_range(self, arr=None, preference='row'):
         """
         if arr is None:
             # use the first array in the row data
-            self.GetRowData().GetArrayName(0)
+            arr = self.GetRowData().GetArrayName(0)
         if isinstance(arr, str):
             arr = get_array(self, arr, preference=preference)
         # If array has no tuples return a NaN range
-        if arr is None or arr.size == 0 or not np.issubdtype(arr.dtype, np.number):
+        if arr.size == 0 or not np.issubdtype(arr.dtype, np.number):
```
tkoyama010, line 353: I noticed that this line is correct.
```suggestion
        if arr is None or arr.size == 0 or not np.issubdtype(arr.dtype, np.number):
```

banesullivan (approved):
Good catch, thanks!

PR #4082, opened 2023-03-02
Make DataSet.cell a generator

### LINKED ISSUE #3971: vtkCell wrappings slow down
@bjlittle is experiencing some slow downs from the new wrapping of vtkCell in 0.38+

> performance bottleneck that I was hitting when using pyvista 0.38+. Basically it's a result of the new wrapping of vtkCell and calls to mesh.cell[cid] instead of mesh.get_cell(cid)... Which makes a *massive* difference, particularly for large/huge meshes.

We should look into this

MatthewFlamm:
### Overview

Make `DataSet.cell` a generator to prevent `mesh.cell[0]` usage.  It is recommended in the docstring to use this for `for cell in mesh.cell:` usage and `mesh.get_cell(0)` for single cell access.

### Breaking Change

`DataSet.cell` no long returns a list and now returns a generator.  Forming a list is a very time intensive operation and it is highly discouraged if only a few cells are needed to access. See #3971.  A generator still allows for the usage of `for cell in mesh.cell:`.  For single cell access, if this form was used `mesh.cell[0]`, now use `mesh.get_cell(0)`.

### Details

Closes #3971.

Tries to ameliorate name clash with `UnstructuredGrid.cells` by putting in a See Also section.  I'm still concerned about this name clash however.  It will be more natural to use `for cell in mesh.cells:`, so I expect a lot of problems here.  Should we add a `cell_array` method to match `vtkCellArray` and then deprecate `cells`?  Once the original meaning of `cells` has been removed, we can move `cell` -> `cells`.  While technically avoiding breaking change, it will probably take years to accomplish and will be confusing in its own right.

### DIFF
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -2531,7 +2531,8 @@ class DataSet(DataSetFilters, DataObject):
         >>> cell.edges[0].point_ids
         [0, 1]
 
-        For a Tetrahedoren, get the point ids of the last face
+        For a Tetrahedron, get the point ids of the last face
+
         >>> mesh = examples.cells.Tetrahedron()
         >>> cell = mesh.get_cell(0)
         >>> cell.faces[-1].point_ids
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -3,7 +3,7 @@
 import collections.abc
 from copy import deepcopy
 import sys
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union, cast
+from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple, Union, cast
 import warnings
 
 if sys.version_info >= (3, 8):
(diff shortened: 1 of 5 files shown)


MatthewFlamm:
We should consider whether we should do the same for `Cell.faces` and `Cell.edges`, i.e. make them generators.  It is MUCH less of a time hit for a one off, but I think being consistent has advantages.  Also, there must be someone out there would want to optimize their code by not forming the entire list, e.g. loop over faces, make a check and exit the for loop.  I'm +1 for this, but I think there was some conversation on the original PR that favored the list usage, so this is up for discussion.

darikg:
`mesh.iter_cells()` ?

akaszynski:
> Should we add a cell_array method to match vtkCellArray and then deprecate cells

Yes. Mistakes were made years ago when naming it. I'm fine with doing it this release. It's a pretty big change, and we should give people a few releases to fix it.

akaszynski (dismissed):
Thanks for fixing this.

By the end of the year we can deprecate the old `cells` and use `cell` instead. Let's slowly transition to the new API over this year.

pyvista/core/dataset.py
```diff
@@ -2510,17 +2510,33 @@ def get_cell(self, index: int) -> 'pyvista.Cell':
...
+        array([[897.0,  48.8,  82.3],
+               [906.6,  48.8,  80.7],
+               [907.5,  55.5,  83.7]])
+
+        For the first cell, get the points associated with the first edge
+
+        >>> cell.edges[0].point_ids
+        [0, 1]
+
+        For a Tetrahedoren, get the point ids of the last face
```
tkoyama010, line 2534:
```suggestion
        For a Tetrahedron, get the point ids of the last face

```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM. Thanks.

Issue #3920, opened 2023-01-31
Make point-merging behavior of `PolyDataFilters.merge()` consistent with `MultiBlock.combine()` and `DataSetFilters.merge()`


whophil:
### Describe the feature you would like to be added.

Currently, `MultiBlock.combine()` and `DataSetFilter.merge()` use `vtkAppendFilter`'s built-in point merging when point merging is requested. In contrast, `PolyDataFilters.merge()` delegates the point-merging to `vtkCleanPolyData`, which is considerably slower for large meshes.

Performance differences aside, since all of these are using `vtkAppendFilter` under the hood, I would expect the behavior to be consistent.

See performance numbers below for a simple test case.

```python
import pyvista as pv
import time

s = pv.Sphere(theta_resolution=2000, phi_resolution=2000)
h1 = s.clip(normal='x', invert=False)
h2 = s.clip(normal='x', invert=True)

t0 = time.time()
r = pv.MultiBlock([h1, h2]).combine(merge_points=True)
print('MultiBlock.combine():', time.time() - t0)

t0 = time.time()
r = h1.cast_to_unstructured_grid().merge(h2, merge_points=True)
print('UnstructuredGrid.merge():', time.time() - t0)

t0 = time.time()
r = h1.merge(h2, merge_points=True)
print('PolyData.merge():', time.time() - t0)
```

Outputs:
```
MultiBlock.combine(): 1.3390300273895264
UnstructuredGrid.merge(): 1.6437718868255615
PolyData.merge(): 9.812441110610962
```

whophil:
Correction - 

`PolyDataFilters.merge()` actually does _not_ use `vtkAppendFilter`, it uses `vtkAppendPolyDataFilter`, which does **not** have any point merging built in like `vtkAppendFilter` does.

Given that, I don't know if there's anything left to really change here. It's certainly unintuitive that `PolyDataFilters.merge(merge_points=True)` and `DataSetFilters.merge(merge_points=True)` have such drastically difference performance - an **hour** in my use case - but given the current state of VTK filters, I don't know what else can be done.

For my use case, the workaround is to cast PolyData -> UnstructuredGrid, then `merge()`, and then to extract a PolyData again.

CC @akaszynski @adeak whose names appear in this filter's blame.

banesullivan:
[...]

PR #4085, opened 2023-03-04
Add tolerance to UnstructuredGrid.merge

tkoyama010:
### Overview

Add `tolerance` to `UnstructuredGrid.merge`.

 

### Details

- None.

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -4554,6 +4554,7 @@ class DataSetFilters:
         self,
         grid=None,
         merge_points=True,
+        tolerance=0.0,
         inplace=False,
         main_has_priority=True,
         progress_bar=False,
@@ -4579,6 +4580,10 @@ class DataSetFilters:
         merge_points : bool, optional
             Points in exactly the same location will be merged between
             the two meshes. Warning: this can leave degenerate point data.
+
+        tolerance : float, default: 0.0
+            The absolute tolerance to use to find coincident points when
+            ``merge_points=True``. Note, this was added in VTK v9.
 
         inplace : bool, optional
             Updates grid inplace when True if the input type is an
@@ -4617,6 +4622,8 @@ class DataSetFilters:
         """
         append_filter = _vtk.vtkAppendFilter()
         append_filter.SetMergePoints(merge_points)
+        if pyvista.vtk_version_info >= (9,):
+            append_filter.SetTolerance(tolerance)
 
         if not main_has_priority:
             append_filter.AddInputData(self)
@@ -4627,6 +4634,8 @@ class DataSetFilters:
             grids = grid
             for grid in grids:
                 append_filter.AddInputData(grid)
+        elif grid is None:
+            append_filter.AddInputData(self)
 
         if main_has_priority:
             append_filter.AddInputData(self)
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -4618,8 +4618,6 @@ class DataSetFilters:
             grids = grid
             for grid in grids:
                 append_filter.AddInputData(grid)
-        elif grid is None:
-            append_filter.AddInputData(self)
 
         if main_has_priority:
             append_filter.AddInputData(self)
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -4607,8 +4607,7 @@ class DataSetFilters:
         """
         append_filter = _vtk.vtkAppendFilter()
         append_filter.SetMergePoints(merge_points)
-        if pyvista.vtk_version_info >= (9,):
-            append_filter.SetTolerance(tolerance)
+        append_filter.SetTolerance(tolerance)
 
         if not main_has_priority:
             append_filter.AddInputData(self)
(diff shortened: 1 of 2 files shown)


tests/test_filters.py
```diff
@@ -2718,3 +2718,12 @@ def test_integrate_data():
 def test_subdivide_tetra(tetbeam):
     grid = tetbeam.subdivide_tetra()
     assert grid.n_cells == tetbeam.n_cells * 12
+
+
+def test_merge_points():
```
tkoyama010, line 2723:
```suggestion
@pytest.mark.needs_vtk9
def test_merge_points():
```

pyvista/core/filters/data_set.py
```diff
@@ -4627,6 +4634,8 @@ def merge(
             grids = grid
             for grid in grids:
                 append_filter.AddInputData(grid)
+        elif grid is None:
+            append_filter.AddInputData(self)
```
MatthewFlamm, line 4638: I assume this change is about this part of the docstring:

> Grid is updated in-place by default. Can be used to merge points of adjacent cells when no grids are input.

I would argue that the current code is better since this method also is described as `Join one or many other grids to this grid.`. No grid input would mean no change.  I'm also thinking about this use case:

```py
meshes = [ mesh for mesh in multi_mesh if some_condition(mesh)]
merged = grid.merge(meshes)
```

This would very surprising to find out that `grid` merged with itself if `meshes` is an empty list.

We can still merge a grid with itself in the current code.

MatthewFlamm: Looking closer, my example isn't true to since this PR only merges if `grid is None`.  I'm still concerned about the complexity of logic here and there might be some edge cases that will be unexpected.

tkoyama010: Good point. I will change the docstring without changing the code to convey this information.

MatthewFlamm: I'm not sure whether this behavior is intended for a good reason though, it may not be so obvious how to merge points within a dataset itself.  I mainly wanted to start a discussion on what is ideal since we have a change here.

tkoyama010: Can @pyvista/developers comment on this?

whophil:
Just wanted to add some notes here - this only affects `DataSetFilter.merge()`. For `PolyData` instances, this filter is overridden by `PolyDataFilters.merge()`, which does something entirely different for merging - see https://github.com/pyvista/pyvista/issues/3920.

Further, the other place where `vtkAppendFilter()` is used in this way is in `MultiBlockFilters.combine()`, which already seems to implement the `tolerance` arg in the same way you've done here.

https://github.com/pyvista/pyvista/blob/09e870f2bdbfd2aab1467395cbc7f089b503ae3a/pyvista/core/filters/composite.py#L33-L81

pyvista/core/filters/data_set.py
```diff
@@ -4602,6 +4607,8 @@ def merge(
         """
         append_filter = _vtk.vtkAppendFilter()
         append_filter.SetMergePoints(merge_points)
+        if pyvista.vtk_version_info >= (9,):
+            append_filter.SetTolerance(tolerance)
```
akaszynski, line 4611:
```suggestion
        append_filter.SetTolerance(tolerance)
```

pyvista/core/filters/data_set.py
```diff
@@ -4565,6 +4566,10 @@ def merge(
             Points in exactly the same location will be merged between
             the two meshes. Warning: this can leave degenerate point data.
 
+        tolerance : float, default: 0.0
+            The absolute tolerance to use to find coincident points when
+            ``merge_points=True``. Note, this was added in VTK v9.
```
akaszynski, line 4571:
```suggestion
            ``merge_points=True``.
```

tests/test_filters.py
```diff
@@ -2672,3 +2672,13 @@ def test_integrate_data():
 def test_subdivide_tetra(tetbeam):
     grid = tetbeam.subdivide_tetra()
     assert grid.n_cells == tetbeam.n_cells * 12
+
+
+@pytest.mark.needs_vtk9
```
akaszynski, line 2677:
```suggestion
```

pyvista/core/filters/data_set.py
akaszynski: There are two issues here:
1. `dataset.merge()` with no arguments seems non-intuitive.
2. We have no clear method for merging coincident points within a dataset.

We can resolve this in one of two ways:
1. Create a new method just for merging points (explicit behavior).
2. Modify `merge` as done in this PR (implicit) to allow for the merge of points.

I'm leaning towards (2) because we already specifically call out this use case in: `Can be used to merge points of adjacent cells when no grids are input.` If we add an example to the docstring merging points of a dataset, I think that should make it clear how to use the method in this manner.

akaszynski:
Decided to completely remove:

```py
        elif grid is None:
            append_filter.AddInputData(self)
```

Reason being is that this ends up duplicating cells because `self` is added twice. Even though the points are merged, the side effect of duplicated cells is troubling.

MatthewFlamm:
> Decided to completely remove:
> 
> ```python
>         elif grid is None:
>             append_filter.AddInputData(self)
> ```

And in response to your comment above about providing a mechanism for users to do this, they can still do

```py
mesh.merge(mesh)
```

akaszynski:
Recommending making this reviewable if you're happy with it @tkoyama010.

akaszynski (approved):
Good to go.

For anyone in the future reading this, note that merging with `self` does indeed merge points, but also results in duplicate cells. It's a caveat worth mentioning.

PR #4102, opened 2023-03-09
Stop plotter from reopening

akaszynski:
Resolve #4057 by making ``self.iren.initialize()`` conditional on the VTK version.

This will need verification from mulitple OSes before merging. I've only tested it locally on Ubuntu 22.04.

PR #4108, opened 2023-03-09
Use default matplotlib rcparams during testing

MatthewFlamm:
### Overview

I use a matplotlibrc file for defaults for making better looking figures.  However, this means that I get an image regression failure when running the test suite.  This PR uses the default.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,6 +16,7 @@ def set_mpl():
     except ImportError:
         pass
     else:
+        matplotlib.rcdefaults()
         matplotlib.use('agg', force=True)
 
 


MatthewFlamm:
I'm not sure that I even believe codecov results anymore.  Is this a result of our CI settings?

akaszynski:
> I'm not sure that I even believe codecov results anymore. Is this a result of our CI settings?

It was working on https://github.com/pyvista/pyvista/pull/4102. I think it's a codecov thing.

akaszynski (approved):
LGTM.

PR #4111, opened 2023-03-09
Remove codecov for integration tests

akaszynski:
@MatthewFlamm, found out why `codecov` was giving us erroneous results: turns out we're uploading coverage for the integration tests.

This PR removes the upload and the coverage tests.


akaszynski:
Codecov will still be off until we get this merged. I'm merging this right away because this is ultra annoying.

MatthewFlamm (approved):

MatthewFlamm:
Even if it didn't mess up codecov, I'm +1 for removing these for coverage.  We don't want to rely on external testing for coverage.

PR #4113, opened 2023-03-09
Change source links to github

akaszynski:
The ``mne`` libray uses `sphinx.ext.linkcode` rather than ``sphinx.ext.viewcode`` and they link directly to github:
[mne.io.read_raw](https://mne.tools/stable/generated/mne.io.read_raw.html)

This PR implements the same logic and uses a source link to GitHub. It's much cleaner than having source within our documentation, and it has the (minor) added advantage of decreasing the build time and size of our docs.

### DIFF
--- a/pyvista/utilities/docs.py
+++ b/pyvista/utilities/docs.py
@@ -60,8 +60,6 @@ def linkcode_resolve(domain, info):
             fn = inspect.getsourcefile(sys.modules[obj.__module__])
         except Exception:
             return None
-
-    if not fn:  # pragma: no cover
         return None
 
     fn = op.relpath(fn, start=op.dirname(pyvista.__file__))
--- a/pyvista/utilities/docs.py
+++ b/pyvista/utilities/docs.py
@@ -1,0 +1,85 @@
+"""Supporting functions for documentation build."""
+import inspect
+import os
+import os.path as op
+import sys
+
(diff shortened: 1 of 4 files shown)


akaszynski:
@jorgepiloto and @Revathyvenugopal162, you may consider making the changes implemented here into a general purpose sphinx extension as I think it's quite useful to link to github.

akaszynski:
@MatthewFlamm, seems that https://github.com/pyvista/pyvista/pull/4111 indeed fixed codecov.

pyvista/utilities/docs.py
```diff
@@ -0,0 +1,85 @@
...
+        fn = None
+
+    if not fn:  # pragma: no cover
+        try:
+            fn = inspect.getsourcefile(sys.modules[obj.__module__])
+        except Exception:
+            return None
+
+    if not fn:  # pragma: no cover
+        return None
```
tkoyama010, line 65:
```suggestion
    if not fn:  # pragma: no cover
        try:
            fn = inspect.getsourcefile(sys.modules[obj.__module__])
        except Exception:
            return None
        return None
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

PR #2684, opened 2022-05-23
Add .git-blame-ignore-revs file with black PR commit

adeak:
[GitHub has finally added support for ignoring commits in blames](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)! I've been using a local `.git-blame-ignore-revs` file to ignore The Blackening, but now it makes sense to add it to the repo itself.

Some resources:
  * [Github's explanation](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)
  * [`git blame`'s doc](https://www.git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt)
  * [`git config`'s doc for configuring a local ignore-revs file](https://git-scm.com/docs/git-config#Documentation/git-config.txt-blameignoreRevsFile).
  
GitHub's blame view of https://github.com/pyvista/pyvista/blame/main/examples/00-load/create-point-cloud.py before adding this file:
screenshot with multiple commits shown in blame due to black

and after:
screenshot with black commits hidden

It would be nice if we could somehow add the configuration of the ignore-revs-file to pre-commit or some other setup steps, but I haven't investigated this yet.

PR #4119, opened 2023-03-11
Apply blackdoc to python source

akaszynski:
PyVista has adopted [black](https://github.com/psf/black), but we fail to implement this for our documentation or doc strings.

This PR implements [blackdoc](https://github.com/keewis/blackdoc) to reformat all Python code in our docstrings as a `pre-commit`.

Pinging @jorgepiloto because this is really cool and IMO works better than [blacken-docs](https://pypi.org/project/blacken-docs/)

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -393,7 +393,9 @@ class DataSetFilters:
         >>> import pyvista as pv
         >>> from pyvista import examples
         >>> dataset = examples.load_hexbeam()
-        >>> _below, _above = dataset.clip_scalar(scalars="sample_point_scalars", value=100, both=True)
+        >>> _below, _above = dataset.clip_scalar(
+        ...     scalars="sample_point_scalars", value=100, both=True
+        ... )
 
         Remove the part of the mesh with "sample_point_scalars" below 100.
 
@@ -900,8 +902,7 @@ class DataSetFilters:
 
         >>> pl = pyvista.Plotter()
         >>> _ = pl.add_mesh(hills, smooth_shading=True, style='wireframe')
-        >>> _ = pl.add_mesh(line_slice, line_width=10, render_lines_as_tubes=True,
-        ...                 color='k')
+        >>> _ = pl.add_mesh(line_slice, line_width=10, render_lines_as_tubes=True, color='k')
         >>> _ = pl.add_mesh(arc, line_width=10, color='grey')
         >>> pl.show()
 
@@ -1051,16 +1052,14 @@ class DataSetFilters:
 
         >>> import pyvista
         >>> noise = pyvista.perlin_noise(0.1, (1, 1, 1), (0, 0, 0))
-        >>> grid = pyvista.sample_function(noise, [0, 1.0, -0, 1.0, 0, 1.0],
-        ...                                dim=(20, 20, 20))
+        >>> grid = pyvista.sample_function(noise, [0, 1.0, -0, 1.0, 0, 1.0], dim=(20, 20, 20))
         >>> grid.plot(cmap='gist_earth_r', show_scalar_bar=True, show_edges=False)
 
         Next, apply the threshold.
 
         >>> import pyvista
         >>> noise = pyvista.perlin_noise(0.1, (1, 1, 1), (0, 0, 0))
-        >>> grid = pyvista.sample_function(noise, [0, 1.0, -0, 1.0, 0, 1.0],
-        ...                                dim=(20, 20, 20))
+        >>> grid = pyvista.sample_function(noise, [0, 1.0, -0, 1.0, 0, 1.0], dim=(20, 20, 20))
         >>> threshed = grid.threshold(value=0.02)
         >>> threshed.plot(cmap='gist_earth_r', show_scalar_bar=False, show_edges=True)
 
@@ -1183,8 +1182,7 @@ class DataSetFilters:
 
         >>> import pyvista
         >>> noise = pyvista.perlin_noise(0.1, (2, 2, 2), (0, 0, 0))
-        >>> grid = pyvista.sample_function(noise, [0, 1.0, -0, 1.0, 0, 1.0],
-        ...                                dim=(30, 30, 30))
+        >>> grid = pyvista.sample_function(noise, [0, 1.0, -0, 1.0, 0, 1.0], dim=(30, 30, 30))
         >>> threshed = grid.threshold_percent(0.5)
         >>> threshed.plot(cmap='gist_earth_r', show_scalar_bar=False, show_edges=True)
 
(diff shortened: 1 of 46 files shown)


tkoyama010 (approved):
Excellent. I was always concerned that the docstring code was not formatted by black. This solves that!

adeak:
Any chance that we can separate the black run from any other change (i.e. adding this to pre-commit) in separate PRs? I'd like to be able to skip the style PR's merge commit in the git blame. As usual.

adeak:
To be clear the ideal situation would be to separate all autoformatting. If you changed anything manually and know what that was, that should also be moved along with the other changes. With one big commit that's probably not feasible though.

And along with the pre-commit addition we should add a check for this in CI, otherwise noncomforming commits will startvseeping back quick.

tkoyama010:
> And along with the pre-commit addition we should add a check for this in CI, otherwise noncomforming commits will startvseeping back quick.

I guess the pre-commit error will stop CI, isn't it?
https://github.com/pyvista/pyvista/blob/5ff97bbf64f1a4f1f39627db3ccbb7f6d9e71c4b/.github/workflows/style_docstr.yml#L16-L26

adeak:
Aaah, right! Thanks @tkoyama010 👍

MatthewFlamm:
This would be a good example where a merge commit would be better to preserve the commit history in a PR (one commit for each step here), but we have disabled this (for good reasons).

Another thought is that you could have two PRs, one for adding to our pre-commit (this will fail), the other for the changes.  You would merge the first if the second passes and then merge them in one after the other.

A lot of extra work however.

adeak:
> Another thought is that you could have two PRs, one for adding to our pre-commit (this will fail), the other for the changes. You would merge the first if the second passes and then merge them in one after the other.

That's what I had in mind. First merge the autoformatting changes, then adding the pre-commit hook and any hypothetical manual changes needed.

> A lot of extra work however.

Only if @akaszynski touched formatting beyond running blackdoc. The non-autoformat changes would be quite obvious otherwise. And we can only do this right once, now, or we'll be stuck with crap history.

It's of course possible that I'm the only one looking at git blames, in which case "sucks to be me" is another option :P

akaszynski:
Agree with separating this out. I didn't consider git blame, but when I do need it it's quite valuable for tracking down bugs.

adeak:
You could keep most of the changes here, and just revert the few other changes. Whichever is less of a hassle.

akaszynski:
Ok @adeak. I went ahead and opened #4120. I would like to merge the changes from branch into main with and also add `git blame --ignore-rev`.

After the merge I will add the commit hash to the `. git-blame-ignore-revs` file.

See
https://akrabat.com/ignoring-revisions-with-git-blame/
https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/

akaszynski:
Please note that the changes here are entirely from the auto-formatter. Recommending merging this PR first and adding the ignore rev file to #4120.

adeak:
> Please note that the changes here are entirely from the auto-formatter. Recommending merging this PR first and adding the ignore rev file to #4120.

I'm on mobile for the next few hours still, but if this only has autoformatting only then this is good to go. Any manual corrections would have to be put elsewhere anyway.

Adding the ignore-revs item in the follow-up is fine too. For the original blackening we had https://github.com/pyvista/pyvista/pull/2684.

Thanks a lot @akaszynski.

adeak (approved):

MatthewFlamm:
This sounds like a good plan

MatthewFlamm (approved):

pyvista/core/filters/data_set.py
```diff
@@ -393,7 +393,9 @@ def clip_scalar(
         >>> import pyvista as pv
         >>> from pyvista import examples
         >>> dataset = examples.load_hexbeam()
-        >>> _below, _above = dataset.clip_scalar(scalars="sample_point_scalars", value=100, both=True)
```
adeak, line 396: I wonder how the line length limit exactly works. From [the docs](https://blackdoc.readthedocs.io/en/latest/options.html#black) I assume it uses the same line length as `black`. This doesn't sound ideal though, because I we explicitly strive to have shorter lines for docs [as per CONTRIBUTING.md](https://github.com/pyvista/pyvista/blob/main/CONTRIBUTING.rst#coding-style):

> As for docstrings, follow the guidelines specified in [PEP 8 Maximum Line Length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) of limiting docstrings to 72 characters per line.

Looking at the blackdoc docs it doesn't seem obvious that we can choose a different line length for blackdoc, but [based on the source](https://github.com/keewis/blackdoc/blob/6fa571ad4c905745bf9574bf08248c60529ebe81/blackdoc/blackcompat.py#L106-L109) it seems that the `[blackdoc]` section in `pyproject.toml` takes precedence (as would be logical):
```py
    black_config = pyproject_toml.get("tool", {}).get("black", {})
    blackdoc_config = pyproject_toml.get("tool", {}).get("blackdoc", {})
    config = {**black_config, **blackdoc_config}
```

So do we want to enforce the 72-char length we mention for docstrings?

adeak: Ha, too late :D

pyvista/utilities/reader.py
```diff
@@ -156,7 +156,7 @@ def get_reader(filename, force_ext=None):
     >>> reader  # doctest: +ELLIPSIS
     XMLPolyDataReader('.../Human.vtp')
     >>> mesh = reader.read()
-    >>> mesh # doctest: +ELLIPSIS
+    >>> mesh  # doctest: +ELLIPSIS
```
adeak, line 159: This is not a blackdoc issue, only a pointer to something potentially redundant: in `pyproject.toml` we have `doctest_optionflags = 'NUMBER ELLIPSIS'` so I wonder if the four `+ELLIPSIS` directives we have in `reader.py` are reduntant. To be looked at later if anyone feels like it.

pyvista/utilities/transformations.py
```diff
@@ -240,7 +244,10 @@ def apply_transformation_to_points(transformation, points, inplace=False):
     >>> points_orig = points.copy()
     >>> scale_factor = 2
     >>> tf = scale_factor * np.eye(4)
-    >>> tf[3, 3,] = 1
+    >>> tf[
+    ...     3,
+    ...     3,
+    ... ] = 1
```
adeak, line 250: I looked through the 45 changed files and this was the only terrible change. I suspect this is mostly a bug, and has to do with the trailing comma. I'd consider removing this trailing comma in the follow-up PR.

pyvista/core/filters/data_set.py
adeak: Pinging @akaszynski just to make sure we discuss this (and the other, far less important points) before going forward.

adeak:
Also pointing out for posterity that the changes here were made with an explicit exclusion of `pyvista/plotting/charts.py`, see https://github.com/pyvista/pyvista/pull/4120#discussion_r1133144610.

akaszynski:
I’ll work on the follow up PR and implement the changes before enforcing blackdox. Thanks for the comments, I’ll be less trigger happy…

adeak:
> I’ll be less trigger happy…

It's all good, I wouldn't have approved if I had insisted on taking a close look before merge :) In the worst case of changing the config and rerunning `blackdoc` on the whole codebase we can just have another autoformat PR and add _two_ commits to the ignore-revs file. No harm done.

pyvista/core/filters/data_set.py
adeak: Hmm, yeah, too bad [the problem seems to be a hornet's nest](https://stackoverflow.com/a/71041192/5067311).

pyvista/utilities/transformations.py
akaszynski: Resolved in https://github.com/pyvista/pyvista/pull/4129.

pyvista/core/filters/data_set.py
akaszynski: Discussed in https://github.com/pyvista/pyvista/pull/4129

PR #4123, opened 2023-03-12
Add Auto fixes to ci

tkoyama010:
### Overview

When we modify code in the browser or get a PullRequest from someone else and there is an error in the pre-commit CI, we have to launch a local terminal to do the pre-commit. This is very annoying; since the pre-commit change is being made in CI, let's commit there.

 

### Details

- None

### DIFF
--- a/.github/workflows/style_docstr.yml
+++ b/.github/workflows/style_docstr.yml
@@ -23,6 +23,7 @@ jobs:
         run: pip install pre-commit
 
       - name: Run pre-commit
+        continue-on-error: true
         run: pre-commit run --all-files
 
       - name: Auto fixes
--- a/.github/workflows/style_docstr.yml
+++ b/.github/workflows/style_docstr.yml
@@ -25,6 +25,11 @@ jobs:
       - name: Run pre-commit
         run: pre-commit run --all-files
 
+      - name: Auto fixes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: auto fixes from pre-commit hooks
+
   docstringcheck:
     runs-on: ubuntu-20.04
     steps:
--- a/.github/workflows/style_docstr.yml
+++ b/.github/workflows/style_docstr.yml
@@ -31,6 +31,10 @@ jobs:
         with:
           commit_message: auto fixes from pre-commit hooks
 
+      - name: Run pre-commit
+        continue-on-error: false
+        run: pre-commit run --all-files
+
   docstringcheck:
     runs-on: ubuntu-20.04
     steps:
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -19,6 +19,8 @@ class Cell(_vtk.vtkGenericCell, DataObject):
     be useful when walking through a cell's individual faces or investigating
     cell properties.
 
+    Prameters
+    ----------
     vtk_cell : vtk.vtkCell, optional
         The vtk object to wrap as Cell, that must be of ``vtk.vtkCell`` type.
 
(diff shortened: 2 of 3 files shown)


.github/workflows/style_docstr.yml
```diff
@@ -25,6 +25,11 @@ jobs:
       - name: Run pre-commit
         run: pre-commit run --all-files
```
tkoyama010, line 26:
```suggestion
        continue-on-error: true
        run: pre-commit run --all-files
```

pyvista/core/cell.py
```diff
@@ -19,8 +19,6 @@ class Cell(_vtk.vtkGenericCell, DataObject):
     be useful when walking through a cell's individual faces or investigating
     cell properties.
 
-    Parameters
-    ----------
```
tkoyama010, line 23: Parameters
    ----------

pyvista/core/cell.py
```diff
@@ -19,7 +19,7 @@ class Cell(_vtk.vtkGenericCell, DataObject):
     be useful when walking through a cell's individual faces or investigating
     cell properties.
 
-    Parameters
+    Prameters
```
tkoyama010, line 22:
```suggestion
    Parameters
```

adeak:
I'm a bit conflicted about this. It seeems like really dangerous territory to have CI push commits. On the other hand I acknowledge that this would make web-based development a lot smoother.

Two technical points:
1. Would it be possible to have this functionality as an opt-in feature via a bot? I imagine posting a comment with `@botname autoformat` or something. It would be slightly more work than autofix by CI, but it would be fully transparent and controllable.
2. Will this work on fork branches?

tkoyama010:
Good idea. I will try it that way.

MatthewFlamm:
> 1. Would it be possible to have this functionality as an opt-in feature via a bot?

I agree with this.  We need to balance competing objectives here.  The use case I'm thinking of is for new contributors who are not git-savvy.  I think automated changes might scare them off as they now need to pull from remote and potentially have conflicts.  It is like dumping them in the deep end right away.  If it is opt-in via bot command, we can wait until the PR is ready to merge and then run.

MatthewFlamm:
> we have to launch a local terminal to do the pre-commit

I had used [gitpod](https://docs.pyvista.org/extras/developer_notes.html#using-gitpod-workspace)  for these kind of uses in the past, if the bottleneck is local terminal.  It does take some time to preload everything unfortunately.

MatthewFlamm:
> Will this work on fork branches?

It _should_ work in either method on any fork branch when the "allow for edits by maintainers" (not sure on exact wording) option is enabled.  On PRs without this, we'd have to do the old fashioned way.

akaszynski:
Full agreement that any changes should be opt-in. I've sometimes made mistakes myself (forgot to install pre-commit in a new venv).

banesullivan:
>  if @tkoyama010's workflow often doesn't involve a terminal of some sort then this is worth pursuing.

+1 -- If it will make anyone's workflow for contributing easier, happy to have it added

tkoyama010:
I guess using CodeSpaces #4171 resolves this problem. Closing.

tkoyama010:
I already solved PullRequest, but I had an idea and I'd like to comment on it. If you want to optionally perform a GitHubAction, you can fulfill this need by triggering it with a label. In this case, we only need to execute the GitHubAction when the label "auto format" is added.

adeak:
@tkoyama010 how would that work with multiple requests to auto-format during the lifetime of a PR? If you remove the label and apply it again, will it run auto-format again? Or would auto-format run each time a commit is pushed as long as the label is there?

tkoyama010:
@adeak
Thank you for your comments, I will stop considering this matter as [it seems complicated to deal with TOKEN](https://github.com/pre-commit/action#using-this-action-with-custom-invocations).
As a bonus to my research, I found a [pre-commit official GitHubAction](https://github.com/pre-commit/action) and created #4263.

PR #4125, opened 2023-03-13
Add pydocstringformatter

akaszynski:
In an effort to universally fix wrapping for our docstrings, I came across [pydocstringformatter](https://github.com/DanielNoord/pydocstringformatter). It has the potential to be that solution in the future, but for the time being it's fairly limited.

I still find it quite useful at preventing simple mistakes that would normally take **forever** to show up in the normal documentation build. Since it does basic enforcement of numpydoc standards, let's include this.

I'm on the fence of if we should include or not include this in the blame.

### DIFF
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -1939,7 +1939,6 @@ class StructuredGrid(_vtk.vtkStructuredGrid, PointGrid, StructuredGridFilters):
 
     @dimensions.setter
     def dimensions(self, dims):
-        """Set the dataset dimensions. Pass a length three tuple of integers."""
         nx, ny, nz = dims[0], dims[1], dims[2]
         self.SetDimensions(nx, ny, nz)
         self.Modified()
--- a/pyvista/plotting/plotting.py
+++ b/pyvista/plotting/plotting.py
@@ -785,7 +785,10 @@ class BasePlotter(PickingHelper, WidgetHelper):
 
     @property
     def scalar_bar(self):
-        """First scalar bar.  Kept for backwards compatibility."""
+        """First scalar bar.
+
+        Kept for backwards compatibility.
+        """
         return list(self.scalar_bars.values())[0]
 
     @property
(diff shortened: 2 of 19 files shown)


adeak:
> I'm on the fence of if we should include or not include this in the blame.

Not if we add changes to the pre-commit hooks here :stuck_out_tongue: 

As far as the docs changes go, I'm also on the fence. On the one hand +61/-41 diff means this isn't significant at all, so we can just keep the diff. It's also slightly weird to see the first line of docstrings be broken up, because I thought that this was the basis of the blurb rendered on the page listing methods for a class. But [apparently](https://dev.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.html) that doesn't seem to be the case:

Screenshot from 2023-03-13 23-57-26

("`Plotter.scalar_bar` - First scalar bar", missing the part about backwards compat.)

pyvista/plotting/plotting.py
```diff
@@ -785,7 +785,10 @@ def disable_hidden_line_removal(self, all_renderers=True):
 
     @property
     def scalar_bar(self):
-        """First scalar bar.  Kept for backwards compatibility."""
+        """First scalar bar.
+
+        Kept for backwards compatibility.
+        """
```
akaszynski, line 791:
```suggestion
        """First scalar bar (kept for backwards compatibility)."""
```

banesullivan (approved):

adeak:
> > Not if we add changes to the pre-commit hooks here
> 
> Then we must include it. Thanks.

I managed to write the opposite of what I meant, so thanks for figuring it out anyway.

pyvista/core/pointset.py
```diff
@@ -1939,7 +1939,6 @@ def dimensions(self):
 
     @dimensions.setter
     def dimensions(self, dims):
-        """Set the dataset dimensions. Pass a length three tuple of integers."""
```
adeak, line 1942: I assume this was axed because this is just a setter anyway. But should we add a setting example to the getter docs then?

akaszynski: Turns out there's a bug:

https://github.com/pyvista/pyvista/issues/4135

I'll add to the example there once I fix it.

adeak (approved):

PR #4129, opened 2023-03-13
Use a docstring length of 75

akaszynski:
## Docstrings are limited 75 characters: or how I Learned to Stop Worrying and Love Python

**tl;dr**
75 characters enable both the readers of our online docs and those using the standard Python `help` on 80 character terminals to read our docs without being overly restrictive. This helps us remain as close as possible to PEP8's 72 character guidelines.

### Long Explanation 

First, the standard from [PEP8](https://peps.python.org/pep-0008/) explains: 

> Maximum Line Length
Limit all lines to a maximum of 79 characters.

> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.

The reason for this dates back to an original convention from Fortran and punch cards. as mentioned in [this interesting discussion](https://github.com/python/peps/issues/884#issuecomment-456993432)

The reasoning [ref](http://www.herongyang.com/Computer-History/FORTRAN-Program-Store-on-Punch-Card.html
) is:

> One code line is recorded on one punch card.
> A punch card has 80 columns.
> Columns from 73 to 80 are used to record a sequence number of each card.

For some time I thought that this was just legacy thinking and figured that 99 characters was totally fine since it's all getting rendered via [Sphinx](https://www.sphinx-doc.org/) to html and that's what people typically read, but as a nod to Guido, 79 characters is probably more reasonable.

### Why 75?

First, here's some Python source:
<details>
  <summary>tmp.py</summary>
 
```py
"""This is a class docstring.

#### Same text wrapped to 79 characters  #######################################

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua. Non consectetur a erat nam at
lectus. Arcu felis bibendum ut tristique et egestas. Id velit ut tortor pretium
viverra suspendisse.

"""


def docstr():
    """This function contains a docstring.

    This line is exactly 72 characters #################################
    Here is some lipsum text wrapped to 72 characters ##################

    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
    eiusmod tempor incididunt ut labore et dolore magna aliqua. Non
    consectetur a erat nam at lectus. Arcu felis bibendum ut tristique
    et egestas. Id velit ut tortor pretium viverra suspendisse.

    This line is exactly 79 characters ########################################
    Same text wrapped to 79 characters  #######################################

    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
    tempor incididunt ut labore et dolore magna aliqua. Non consectetur a erat
    nam at lectus. Arcu felis bibendum ut tristique et egestas. Id velit ut
    tortor pretium viverra suspendisse.

    Same text wrapped to 99 characters  ###########################################################

    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
    labore et dolore magna aliqua. Non consectetur a erat nam at lectus. Arcu felis bibendum ut
    tristique et egestas. Id velit ut tortor pretium viverra suspendisse.

    """
    pass

class DocStr():

    def method_docstr():
        """This method contains a docstring.

        This line is exactly 72 characters #############################
        Here is some lipsum text wrapped to 72 characters ##############

        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
        eiusmod tempor incididunt ut labore et dolore magna aliqua. Non
        consectetur a erat nam at lectus. Arcu felis bibendum ut
        tristique et egestas. Id velit ut tortor pretium viverra
        suspendisse.

        This line is exactly 79 characters ####################################
        Same text wrapped to 79 characters  ###################################

        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
        tempor incididunt ut labore et dolore magna aliqua. Non consectetur a
        erat nam at lectus. Arcu felis bibendum ut tristique et egestas. Id
        velit ut tortor pretium viverra suspendisse.

        """
        pass

    class NestedClass():
        """Nested class docstring.

        This line is exactly 72 characters #############################
        Here is some lipsum text wrapped to 72 characters ##############

        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
        eiusmod tempor incididunt ut labore et dolore magna aliqua. Non
        consectetur a erat nam at lectus. Arcu felis bibendum ut
        tristique et egestas. Id velit ut tortor pretium viverra
        suspendisse.

        """
        pass

```
</details>

Let's see how this gets rendered right in the python REPL. For fun, let's use the [cool-retro-term](https://github.com/Swordfish90/cool-retro-term) sized to the [IBM 3270](https://en.wikipedia.org/wiki/IBM_3270) (80x24):

```py
>>> import tmp
>>> help(tmp.docstr)
```

term0

Looks pretty good, right? This seems to make the argument that 79 characters (including the indentation) is just fine.

However, let's try this again when getting the `help` of the module:

```py
>>> import tmp
>>> help(tmp)
```

term1

Ok. So it seems that since everything is indented, why not just use 75 characters. This is actually what numpydoc recommends in https://numpydoc.readthedocs.io/en/latest/format.html:

> The length of docstring lines should be kept to 75 characters to facilitate reading the docstrings in text terminals.

This even works for method docstrings. Behold:

meth

Why then 72? It's to allow a margin to the right of the terminal. As you can see from above, there's no margin at all.

### Should we just go with 72?

The question is: should reduce readability by limiting docstrings to 72 characters due to excessive wrapping using `blackdoc`?

I would argue that we should make the text as wide as possible before it is wrapped on a 80 character terminal. As shown above, this works for 75 characters and has been adopted by the `numpydoc` standard. This stays as close as possible to PEP8 while allowing for an extra four characters of "breathing room" to permit better formatted example sections.

### This PR

This is a refactoring PR and the content will be ignored, we will document our standard in `CONTRIBUTING.md`.


pyvista/utilities/transformations.py
```diff
@@ -244,11 +248,10 @@ def apply_transformation_to_points(transformation, points, inplace=False):
...
     >>> tf = scale_factor * np.eye(4)
-    >>> tf[
-    ...     3,
-    ...     3,
-    ... ] = 1
-    >>> pyvista.transformations.apply_transformation_to_points(tf, points, inplace=True)
+    >>> tf[3, 3] = 1
+    >>> pyvista.transformations.apply_transformation_to_points(
+    ...     tf, points, inplace=True
+    ... )
```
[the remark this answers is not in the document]
adeak: Misfeature? :P Noted, thanks.

pyvista/plotting/_property.py
```diff
@@ -634,7 +634,9 @@ def metallic(self) -> float:
 
         >>> import pyvista as pv
         >>> prop = pv.Property()
-        >>> prop.interpolation = 'pbr'  # requires physically based rendering
+        >>> prop.interpolation = (
+        ...     'pbr'  # requires physically based rendering
+        ... )
```
MatthewFlamm, line 639:
```suggestion
        >>> # requires physically based rendering
        >>> prop.interpolation = 'pbr'
```

pyvista/plotting/_property.py
```diff
@@ -674,7 +676,9 @@ def roughness(self) -> float:
...
```
MatthewFlamm, line 681:
```suggestion
        >>> # requires physically based rendering
        >>> prop.interpolation = 'pbr'
```

pyvista/plotting/_property.py
```diff
@@ -1078,7 +1082,9 @@ def anisotropy(self):
...
```
MatthewFlamm, line 1087:
```suggestion
        >>> # requires physically based rendering
        >>> prop.interpolation = 'pbr'
```

pyvista/plotting/composite_mapper.py
```diff
@@ -341,13 +351,18 @@ def reset_pickabilities(self):
...
+        >>> dataset = pv.MultiBlock(
+        ...     [pv.Cube(), pv.Sphere(center=(0, 0, 1))]
+        ... )
         >>> pl = pv.Plotter()
         >>> actor, mapper = pl.add_composite(dataset)
         >>> mapper.block_attr[1].pickable = True
         >>> mapper.block_attr[2].pickable = False
         >>> mapper.block_attr.reset_pickabilities()
-        >>> [mapper.block_attr[1].pickable, mapper.block_attr[2].pickable]
+        >>> [
```
MatthewFlamm, line 362: Consider making two separate commands, but this is a weak suggestion from me.

pyvista/themes.py
```diff
@@ -1165,9 +1165,17 @@ class _SliderConfig(_ThemeConfig):
     >>> import pyvista as pv
     >>> pv.global_theme.slider_styles.classic.slider_length = 0.02
     >>> pv.global_theme.slider_styles.classic.slider_width = 0.04
-    >>> pv.global_theme.slider_styles.classic.slider_color = (0.5, 0.5, 0.5)
+    >>> pv.global_theme.slider_styles.classic.slider_color = (
```
MatthewFlamm, line 1168: If we do something like the below, it is less messy formatting.

```py
    >>> slider_styles = pv.global_theme.slider_styles
    >>> slider_styles.classic.slider_color = (0.5, 0.5, 0.5)
```

tests/check_doctest_names.py
```diff
@@ -14,8 +14,12 @@
     >>> import vtk
     >>> import pyvista
     >>> offset = np.array([0, 9])
-    >>> cells = np.array([8, 0, 1, 2, 3, 4, 5, 6, 7, 8, 8, 9, 10, 11, 12, 13, 14, 15])
-    >>> cell_type = np.array([vtk.VTK_HEXAHEDRON, vtk.VTK_HEXAHEDRON], np.int8)
+    >>> cells = np.array(
+    ...     [8, 0, 1, 2, 3, 4, 5, 6, 7, 8, 8, 9, 10, 11, 12, 13, 14, 15]
```
MatthewFlamm, line 18: Same comment as previous, now that we are breaking this into multiple lines, the `cells` array could be shown in a human interpretable form with each cell on it's own line.

MatthewFlamm:
My personal thoughts are that usage of examples is in a contradictory state with this phrase:

> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.

I would argue that we might expect to be able to expand our line length when incorporating code into the docstring.  But, given the usefulness of IDE hover hints that also suffer from line-length issues, I think limiting docstring length to some reasonable degree is desirable. 

The below statement is the one that resonates most to me from your explanation for choosing a specific line length.  I am less swayed by the terminal restrictions since I think it is very unlikely someone is using pyvista on such a terminal with specific restrictions in an interactive mode.  Can those systems even install vtk version 9+?  I think modern restrictions tend to be more amorphous (75 better for one user, and 79 better for another user).

> Also, since we're adopting the numpydoc standard, let's align on that length.

MatthewFlamm (commented):
I am disappointed that docstrings are limited to directives with comments inline.  This causes a majority of the poor formatting in my reading, but I'm not sure how to fix those.  I personally don't like when the lines have to be broken only be an inline comments, so I fixed those when I saw them.

adeak:
I don't want to cause additional work to anyone, but we should discuss one thing: manual edits made here vs ignoring this PR in the git blame.

Option 1: we fix up all the uglies here. Pro: no additional work. Con: if we inadvertently break something it will be hidden from the git blame. Large issue in a very unlikely use case.

Option 2: we only keep the autoformat in this PR (assuming we're OK with the diff here, pending manual surgical improvements). Pro: all human intervention will show up in the blame, complete with potential mistakes. Con: considerable amount of additional work of having to apply comments from @MatthewFlamm (thanks!) and potential future comments from others as a separate PR.

Option 3: anything I didn't think of.

What do you all think?

pyvista/core/pointset.py
```diff
@@ -1370,8 +1385,31 @@ def _from_arrays(
         >>> import numpy as np
         >>> from pyvista import CellType
         >>> import pyvista
-        >>> cells = np.array([8, 0, 1, 2, 3, 4, 5, 6, 7, 8, 8, 9, 10, 11, 12, 13, 14, 15])
-        >>> cell_type = np.array([CellType.HEXAHEDRON, CellType.HEXAHEDRON], np.int8)
+        >>> cells = np.array(
+        ...     [
+        ...         8,
```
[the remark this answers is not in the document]
adeak: This is very un-black-like so I doubt this is supported (but I'm only guessing, to be clear). These are the kinds of uglies that are the price to pay for a "dumb" autoformatter. I wonder if it would look a bit less bad (or any different at all) without the "magic trailing comma"...

pyvista/core/filters/poly_data.py
```diff
@@ -2131,10 +2153,15 @@ def multi_ray_trace(
         >>> import pyvista as pv  # doctest:+SKIP
         >>> sphere = pv.Sphere()  # doctest:+SKIP
         >>> points, rays, cells = sphere.multi_ray_trace(
-        ...     [[0, 0, 0]] * 3, [[1, 0, 0], [0, 1, 0], [0, 0, 1]], first_point=True
+        ...     [[0, 0, 0]] * 3,
+        ...     [[1, 0, 0], [0, 1, 0], [0, 0, 1]],
+        ...     first_point=True,
```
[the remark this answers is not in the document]
MatthewFlamm: > Doctest flags are stripped away during doc build:

I'm thinking about inspecting the docstring using `help` or using an IDE hover hint for example.  In these cases, the dev comments show up.

akaszynski:
Voting option 1 with lots of testing.

pyvista/core/filters/data_set.py
```diff
@@ -5088,7 +5176,12 @@ def transform(
         accepted.
 
         >>> transform_matrix = np.array(
-        ...     [[1, 0, 0, 50], [0, 1, 0, 100], [0, 0, 1, 200], [0, 0, 0, 1]]
+        ...     [
+        ...         [1, 0, 0, 50],
+        ...         [0, 1, 0, 100],
+        ...         [0, 0, 1, 200],
+        ...         [0, 0, 0, 1],
+        ...     ]
```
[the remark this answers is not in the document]
akaszynski: Local testing shows that it doesn't matter. I'm guessing it's internally converted to float because I get the same results regardless of the input datatype.

pyvista/core/filters/poly_data.py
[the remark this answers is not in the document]
akaszynski: I'm afraid there's no getting around this unless we want to install `pyembree`, `rtree` and `trimesh` (already included). We have a zillion libraries in our requirements, so I'm of the opinion to continue to ignore it.

MatthewFlamm:
There is no perfect plan.  If the changes are docstring format related, I think the impact is low.  I vote for # 1 too.  Although we have doctesting, they aren't designed primarily to be tests, so it is a soft backstop.

IMO users of git blame are more likely to want to search some functionality change backwards.  So, being on the side of more hidden formatting changes might be beneficial.

It is always good to consider it to make an informed merge rather than regret it later.

pyvista/core/pointset.py
```diff
@@ -929,9 +940,13 @@ def save(self, filename, binary=True, texture=None, recompute_normals=True):
         >>> import numpy as np
         >>> sphere = pyvista.Sphere()
         >>> texture = np.zeros((sphere.n_points, 3), np.uint8)
-        >>> texture[:, 1] = np.arange(sphere.n_points)[::-1]  # just blue channel
+        >>> texture[:, 1] = np.arange(sphere.n_points)[
+        ...     ::-1
+        ... ]  # just blue channel
```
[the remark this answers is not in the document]
akaszynski: I had to think about this myself, and yes, this is not clear.

It's now better written and I think this is sufficient in its new form.

pyvista/core/pointset.py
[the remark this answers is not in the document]
akaszynski: Recommending using `np.hstack`.

pyvista/plotting/composite_mapper.py
akaszynski: Fine as is IMO.

pyvista/core/pointset.py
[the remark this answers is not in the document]
akaszynski: Fixed in 062fe1ab41418178050ae4121a75a2442a22f5f7

pyvista/themes.py
akaszynski: Fixed in 062fe1ab41418178050ae4121a75a2442a22f5f7.

tests/check_doctest_names.py
akaszynski: Fixed in 062fe1ab41418178050ae4121a75a2442a22f5f7.

akaszynski:
Aside from the correction of the color channel wording (blue to green) I think that all these changes are merely style changes and I'm not really concerned about the blame recording these.

MatthewFlamm (approved):
Thanks for addressing my comments.  I like the `np.hstack` usage for forming the cells for `UnstructuredGrid` that you implemented.  This makes the formatting better, and the concept easier to understand, win-win.

banesullivan (approved):

adeak (approved):
I'd have gone with `np.concatenate` instead of `np.hstack`, but I agree with all the changes. Thanks for the discussion, all!

akaszynski:
> I'd have gone with `np.concatenate` instead of `np.hstack`, but I agree with all the changes. Thanks for the discussion, all!

Since we convert to an array regardless, I actually could have gone with just adding the two lists together as well, as that would have been the most concise.

PR #4138, opened 2023-03-15
Add example changing mapped scalars

banesullivan:
Users often request how to change the mapped scalars that have already been plotted. Here is a full fledged example with Trame

For https://github.com/pyvista/pyvista/discussions/4131

Screen Shot 2023-03-15 at 10 21 53 AM

### DIFF
--- a/examples_trame/basic/mesh_scalars.py
+++ b/examples_trame/basic/mesh_scalars.py
@@ -1,0 +1,90 @@
+from trame.app import get_server
+from trame.ui.vuetify import SinglePageLayout
+from trame.widgets import vuetify
+
+import pyvista as pv
+from pyvista import examples
+from pyvista.trame.ui import plotter_ui
+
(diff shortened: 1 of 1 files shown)


akaszynski:
While we can't add this to the gallery, I would like for this to be rendered in our documentation somehow. We could use [literalinclude](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) or just a link, but it needs to have a "copy button". Not everyone will go source diving.

akaszynski:
Also, is there any way we could test these? Even just a smoke test?

akaszynski:
I'm fine with adding this "as-is", but we should have a follow-up with testing and adding to documentation.

akaszynski (approved):

banesullivan:
Merging this as is and will try to make a proper example demonstrating how to modify the mapper, prop, and lut in more ways

PR #4145, opened 2023-03-17
Fix/rectangle to polygon

hippo91:
### Overview

This PR address the problem of having a Rectangle function that behaves as a Quadrilateral factory.

resolves #1604

### Details

- The Rectangle function has been renamed in Quadrilateral.
- A new Rectangle function has been created. 
- It checks that only three points are given. If four points are given it raises a deprecation warning else if more or less than 3 points are given a TypeError is raised.
- The 3 points can be given in arbitrary order but should be different and defines orthogonal vectors otherwise ValueError is raised
- Extensive tests have been added.

PR #1584, opened 2021-08-10
Allow subclassing DataSet using _wrappers dict

MatthewFlamm:
### Overview

This PR closes #1565 and allows for developers to sublass a `DataSet` class and be able to set this as the default object to coerce vtk objects to.

Sample usage:

```py
import pyvista

def Foo(pyvista.PolyData):
    pass

pyvista._wrappers['vtkPolyData'] = Foo

uniform_grid = pyvista.UniformGrid()
surface = uniform_grid.extract_surface()
assert isinstance(surface, Foo)
```
### Details

The test shows that this also works for `inplace=True` usage.  This eliminates the need to change the `overwrite` method.

TODO:

- [x] Find and fix all uses of `isinstance` that checks for `DataSet` classes.
- [x] Wrap all direct uses of a `DataSet` like in geometric objects.
- [x] Add better documentation on different subclassing options that are available.

PR #4146, opened 2023-03-17
pyvista.wrap is not needed for pyvista.PolyData

tkoyama010:
### Overview

pyvsita.wrap is not need for pyvista.PolyData.

 
See https://github.com/pyvista/pyvista/pull/4145#discussion_r1140175163

### Details

- None


banesullivan (dismissed):

banesullivan:
@tkoyama010, would you look into why the tests are now failing for this? Otherwise, thanks for this clean up!

tkoyama010:
The error occurs in the test when defining a wrapper for the subclass added in #1584. @MatthewFlamm I think this test is not necessary if Geometry functions do not use a wrapper. Am I right?
https://github.com/pyvista/pyvista/blob/0a43286293ef8cb922d0ac28428e8ce556f18a86/tests/test_helpers.py#L117-L121

MatthewFlamm:
See https://github.com/pyvista/pyvista/pull/1584#issuecomment-901340448 for the thinking at the time for why it is like this.  I wanted it to be either 1) PyVista aims to make everything the subclass the user specifies, including geometric objects, or 2) the user is in charge of wrapping objects at every step of the way.

I'd welcome ideas on how this could be better (metaclass logic?).

adeak:
> I'd welcome ideas on how this could be better (metaclass logic?).

Toying with an alternative in https://github.com/pyvista/pyvista/pull/4188.

banesullivan:
Closing due to staleness and unclarity

PR #4153, opened 2023-03-17
Link to pytest-pyvista in docs

MatthewFlamm:
### Overview

Add some links to pytest-pyvista documentation.


doc/source/extras/pytest_plugin.rst
```diff
@@ -0,0 +1,6 @@
+.. _ref_pytest_plugin_docs:
+
+PyVista Pytest Plugin
```
adeak, line 3: Super nitpicky, but I wonder about the casing of "Pytest". Looking at https://docs.pytest.org it seems to me that the project styles itself as "pytest":

screenshot from docs.pytest.org with a bunch of lowercase "pytest"s, even outside a coding context, at the start of a sentence

CONTRIBUTING.rst
```diff
@@ -550,6 +550,7 @@ For example, the following writes all images generated by ``pytest`` to
 
    pytest tests/plotting/ -k volume --generated_image_dir debug_images
 
+See `pytest-pyvista <https://pytest.pyvista.org/>` for more details.
```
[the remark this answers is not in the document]
adeak: Also FYI there's a link to the pytest-pyvista github in the preceding paragraph, but I think it's fine to have both links.

doc/source/extras/pytest_plugin.rst
akaszynski: 5896d170fc908f79cb1b77dc830288700d17c749

akaszynski: We're using "title case", so I updated it in the body, but didn't update the title.

akaszynski (dismissed):
Calling this good to go.

doc/source/extras/pytest_plugin.rst
adeak: We could add an exception just like we do for PyVista and others. I don't insist though.

MatthewFlamm: I most admit I was lazy, but not that lazy, and only googled it and saw this.  Their website proper consistently uses lower case.  I think we could use either here.

Screenshot_20230317-201842

adeak: So should we switch or keep titlecase?

tkoyama010 (approved):

doc/source/extras/pytest_plugin.rst
akaszynski: I'm on the fence about this as well. Looking at their docs, they consistently do not capitalize it, even when it starts their sentence case titles:
https://docs.pytest.org/en/7.2.x/contents.html

Honestly, I think it looks weird if it's the only word not capitalized, it's going to look weird. I think an exception would be [gRPC](https://grpc.io/), where GRPC (all caps) or Grpc looks weird. In this case, I think we're fine with "as-is" since it doesn't "look weird".

PR #4155, opened 2023-03-18
Enforce docstring type and default consistency

akaszynski:
This PR is more of a discussion + code suggestion rather than a genuine PR as I'd like to gain traction before implementing this widely.

@akeak, this is particularly for you.

Looking across our docs, I've found that in, for example, [DataSet.find_closest_cell](https://docs.pyvista.org/api/core/_autosummary/pyvista.DataSet.find_closest_cell.html#find-closest-cell), the `Sequence` type does not get rendered properly:

fcc

This isn't a big deal, we have many types that have no links, but still it's annoying. Plus, we're reminded of the "is numpy.ndarray a sequence?" question since we have to accept both a ``Sequence`` and ``np.ndarray`` here.

After some brief research, I found this [SO answer](https://stackoverflow.com/questions/62970581/what-exactly-is-a-sequence#:~:text=A%20numpy%20array%20is%20a,as%20a%20subclass%20of%20Sequence.) stating that a ``numpy.ndarray`` is a [sequence](https://docs.python.org/3/glossary.html#term-sequence) and not a [Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence).

This is actually incredibly helpful because this lets us simplify out documentation because we're fundamentally worried about *if* it is ordered collection of objects that implementes `__len__` and `__get_item__`. We're really concerned about duck typing here, not if it's inherited from [collections.abc.Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence) (capital "s").

Therefore, I propose we change our docstrings to use the ducktyping [sequence](https://docs.python.org/3/glossary.html#term-sequence) where applicable. This is because, according to [PEP 3119](https://www.python.org/dev/peps/pep-3119), ``collections.abc`` check if the object has the expected interface with **without requiring true subclassing**.

In fact, this is appears the same as VTK's rules:

```py
>>> import vtk
>>> vpts = vtk.vtkPoints()
>>> vpts.SetNumberOfPoints(1)
>>> vpts.SetPoint(0, frozenset({1, 2, 3}))
TypeError: SetPoint argument 2: expected a sequence of 3 values, got frozenset

>>> vpts.SetPoint(0, range(1, 4))
>>> vpts.GetPoint(0)
(1.0, 2.0, 3.0)
```

Therefore, to finally clear up the confusion, wherever possible, replace ``iterable``, ``Sequence`` or ``np.ndarray``, or whatever mess of values we have with just [sequence](https://docs.python.org/3/glossary.html#term-sequence):

fcc_fix

---

### Agreement with type hints

We already have a mismatch between our type hints and the docstring, and this suggestion will not improve it. However, I'd argue that our existing typehints, while helpful for `mypy`, are not helpful for humans. Furthermore, in `DataSet.find_closest_cell`, we exclude `range` and other possible types that work. For example:

```py
import numpy as np
import pyvista as pv
from pyvista.utilities.arrays import _coerce_pointslike_arg

# Prepare test sequences
test_sequences = [
    [1, 2, 3],  # list
    (1, 2, 3),  # tuple
    {1, 2, 3},  # set
    frozenset({1, 2, 3}),  # frozenset
    bytearray([1, 2, 3]),  # bytearray
    bytes([1, 2, 3]),  # bytes
    range(1, 4),  # range
]

# Test each sequence type
for seq in test_sequences:
    try:
        result = _coerce_pointslike_arg(seq)
        print(f"Input Type: {type(seq)}")
        print(f"Result Type: {type(result)}")
        print(f"Result: {result}\n")
    except Exception as e:
        print(f"Input Type: {type(seq)}")
        print(f"Error: {e}\n")
```
Output:

```
Input Type: <class 'list'>
Result Type: <class 'tuple'>
Result: (array([[1, 2, 3]]), True)

Input Type: <class 'tuple'>
Result Type: <class 'tuple'>
Result: (array([[1, 2, 3]]), True)

Input Type: <class 'set'>
Error: Given points must be a sequence or an array.

Input Type: <class 'frozenset'>
Error: Given points must be a sequence or an array.

Input Type: <class 'bytearray'>
Result Type: <class 'tuple'>
Result: (array([[1, 2, 3]], dtype=uint8), True)

Input Type: <class 'bytes'>
Error: Given point must have three values

Input Type: <class 'range'>
Result Type: <class 'tuple'>
Result: (array([[1, 2, 3]]), True)
```

A few questions then:
- Can we rewrite our typehints such that they're more helpful?
- Will it become more helpful or more confusing by moving to `sequence[float]` wherever possible even though it disagrees with our type hints?
- Does any of this matter?


adeak:
> Looking across our docs, I've found that in, for example, [DataSet.find_closest_cell](https://docs.pyvista.org/api/core/_autosummary/pyvista.DataSet.find_closest_cell.html#find-closest-cell), the `Sequence` type does not get rendered properly:

I noticed some of these, they are annoying.

> After some brief research, I found this [SO answer](https://stackoverflow.com/questions/62970581/what-exactly-is-a-sequence#:~:text=A%20numpy%20array%20is%20a,as%20a%20subclass%20of%20Sequence.) stating that a `numpy.ndarray` is a [sequence](https://docs.python.org/3/glossary.html#term-sequence) and not a [Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence).

Yes, this makes sense to me.

> This is actually incredibly helpful because this lets us simplify out documentation because we're fundamentally worried about _if_ it is ordered collection of objects that implementes `__len__` and `__get_item__`. We're really concerned about duck typing here, not if it's inherited from [collections.abc.Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence) (capital "s").

Agreed.

> Therefore, I propose we change our docstrings to use the ducktyping [sequence](https://docs.python.org/3/glossary.html#term-sequence) where applicable. This is because, according to [PEP 3119](https://www.python.org/dev/peps/pep-3119), `collections.abc` check if the object has the expected interface with **without requiring true subclassing**.

This definitely makes sense for the docs at least. And I'm pleasantly surprised that `sequence` seems to have an intersphinx mapping; I didn't realize that the glossary was included in that. Assuming that's indeed where the rendered link points.

And I also want to note that the first screenshot you showed is wrong on multiple points. Firstly, `Sequence(float)` is never really correct. It should be `Sequence[float]`. And secondly, we'd probably want to explicitly use [`collections.abc.Sequence`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence) to get the intersphinx link working. Also note the existence of [`typing.Sequence`](https://docs.python.org/3/library/typing.html#typing.Sequence) that we'd need for Python <=3.8 (but type checkers often run on newer Python...).

> ### Agreement with type hints
> 
> We already have a mismatch between our type hints and the docstring, and this suggestion will not improve it. However, I'd argue that our existing typehints, while helpful for `mypy`, are not helpful for humans. Furthermore, in `DataSet.find_closest_cell`, we exclude `range` and other possible types that work. For example:

If what you're saying is "keep type hints correct, let doc types be less correct but more meaningful" then I agree. They serve different purposes so I have no issues letting the two diverge a bit. Of course there should be no blatant contradictions.

> A few questions then:
> 
>   * Can we rewrite our typehints such that they're more helpful?

If by "typehints" you mean the actual type annotations, then I think not. If you mean the ones for the docs, sure. See also my previous point.
 
>   * Will it become more helpful or more confusing by moving to `sequence[float]` wherever possible even though it disagrees with our type hints?

I think it would make sense to use `sequence[float]` where it makes sense. Also using `sequence[float]` when we really and obviously allow integers and NumPy numbers.
 
>   * Does any of this matter?

Absolutely, I'd expect most end users to be looking at the docs more than anything else. Having broken/ugly/confusing types listed there is not very helpful.

And an additional point of my own: you specifically raised `_coerce_pointslike_arg()` as an example. And then you saw that things like `bytes` don't work, even though `issubclass(bytes, collections.abc.Sequence)`. This is because in this helper function (and also a few other functions, I'm sure) we want _array-likes_, i.e. "whatever works with `np.array()`". These are mostly nested "rectangular" (i.e. non-ragged) sequences. But there are some exceptions, such as `bytes`:
```py
>>> np.array(bytes([1, 3, 5])).size
1

>>> np.array(bytes([1, 3, 5]))
array(b'\x01\x03\x05', dtype='|S3')
```
(Although by some definitions this is still an array-like... so I'm using a definition where the resulting array has a numeric type. This might or might not be headcanon.)

Semantically we're not looking for a sequence here, we want an array-like. But in many many cases we really do want a sequence, or actually a fixed-length iterable of any kind (think bounds, min/max limits, light attenuation values etc.). So we should keep an eye out for places where we need array-likes, and perhaps use that instead of `sequence`.

In fact it seems that NumPy has an intersphinx link to [`array_like`](https://numpy.org/devdocs/glossary.html#term-array_like) in its glossary! There's also [`ArrayLike`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike) as of NumPy 1.20 but we probably don't want to use that.

pyvista/core/filters/data_set.py
```diff
@@ -1578,11 +1578,11 @@ def contour(
 
         Parameters
         ----------
-        isosurfaces : int or sequence, optional
+        isosurfaces : int | sequence, optional
             Number of isosurfaces to compute across valid data range or a
             sequence of float values to explicitly use as the isosurfaces.
 
-        scalars : str, collections.abc.Sequence, numpy.ndarray, optional
+        scalars : str | sequence, optional
```
adeak, line 1585: `str` is a sequence though :P

pyvista/utilities/arrays.py
```diff
@@ -33,7 +33,7 @@ def _coerce_pointslike_arg(
         points = np.asarray(points)
 
     if not isinstance(points, np.ndarray):
-        raise TypeError("Given points must be a sequence or an array.")
+        raise TypeError("Given points must be a sequence.")
```
adeak, line 36: Now this is one point where I'd definitely refer to a "numerical array-like" or something similar. Or spell it out as "Given points must be convertible to a numerical array".

pyvista/core/filters/data_set.py
akaszynski:
```suggestion
        scalars : sequence, optional
```

pyvista/utilities/arrays.py
akaszynski:
```suggestion
        raise TypeError("Given points must be convertible to a numerical array.")
```

akaszynski:
> If what you're saying is "keep type hints correct, let doc types be less correct but more meaningful" then I agree.

That's what I'm hoping for.

> Semantically we're not looking for a sequence here, we want an array-like... So we should keep an eye out for places where we need array-likes.

Fully agree, and if we can use [array_like](https://numpy.org/devdocs/glossary.html#term-array_like)[float] and glossary links work, even better.

> > Does any of this matter?
> Absolutely, I'd expect most end users to be looking at the docs more than anything else.

It's always good to ask, especially before going through a refactor. I want the docs to look like one mind wrote them. If we can converge (or at least start to converge) on an idea then it's worth discussing it.

Thanks for the comp-sci discussion.

akaszynski:
In an effort to keep this PR sub 2k LOC, I'm calling it sufficient for the moment.

- Moves from `list or tuple` --> `sequence[float]` whenever possible. Uses `array_like` for `sequence[sequence]` (as in the case of 2D arrays.
- Adds the default whenever available
- Miscellaneous fixes (missing optional, bad default)

Ready for review.

adeak:
Review in progress. I'm still annoyed by the fact that what we end up with is largely hit or miss:
Screenshot from 2023-03-19 17-09-57
(this is from `DataSetAttributes.set_array()`). For some reason `float` and `str` get this border around the link, so do `pyvista_ndarray` and `numpy.ndarray`, but the `bool` and the `array_like` only get a blue link. Due to how wonky this `array_like[float]` (and similarly elsewhere `sequence[float]`) looks, I'd love to be able to disable the whole border thing for intersphinx links. Just make them blue text like normal web pages do.

(In the same sentence `list` and `tuple` aren't resolved at all, but that's because we don't have an explicit `:class:` directive there.)

akaszynski:
> Due to how wonky this array_like[float] (and similarly elsewhere sequence[float]) looks, I'd love to be able to disable the whole border thing for intersphinx links. Just make them blue text like normal web pages do.

I've started by disabling rendering the font as italic. Left is `main`, right is this branch:
tmp

I feel this improves the readability and helps with nested typing `array_like[float]` as the brackets get rendered correctly.

I'm going to try disabling the border to see if that cleans things up more (which it probably will).

akaszynski:
Now, here's without the border and the reduction of the text size to fit said border. Again, `main` left, this branch (`doc/sequence-docstring-types`) right:

no-border

adeak:
> Now, here's without the border and the reduction of the text size to fit said border.

I think that looks a lot better :heart: Here's a screenshot where we have a generic type:

Screenshot from 2023-03-19 23-04-56

It feels a bit unfamiliar that the type links are so big (i.e. as big as the rest of the text). I think I'll get used to the new readable one fast. Assuming you agree this is better and we should keep it.

pyvista/core/pointset.py
```diff
@@ -1359,22 +1358,23 @@ def _from_arrays(
...
-        deep : bool, optional
+        deep : bool, default: True
             When ``True``, makes a copy of the points array.  Default
             ``False``.  Cells and cell types are always copied.
 
-        force_float : bool, optional
+        force_float : bool, default: False
             Casts the datatype to ``float32`` if points datatype is
             non-float.  Default ``True``. Set this to ``False`` to allow
             non-float types, though this may lead to truncation of
```
adeak, line 1380: Actual default is `True`, matching description. Won't suggest and edit here because removing the default from the description will need rewrapping anyway.

pyvista/core/pointset.py
```diff
@@ -2237,11 +2237,11 @@ def _from_arrays(self, dims: Sequence, corners: Sequence) -> None:
...
         ----------
-        dims : Sequence
+        dims : sequence[int]
             A sequence of integers with shape (3,) containing the
             topological dimensions of the grid.
 
-        corners : Sequence
+        corners : sequence[floats]
             A sequence of floats with shape (number of corners, 3)
             containing the coordinates of the corner points.
```
adeak, line 2246: I'd also make anything with "shape" an `array_like`
```suggestion
        corners : array_like[floats]
            A sequence of floats with shape (number of corners, 3)
            containing the coordinates of the corner points.
```

pyvista/core/pointset.py
```diff
@@ -2465,7 +2466,7 @@ def dimensions(self) -> Tuple[int, int, int]:
 
         Returns
         -------
-        tuple(int)
+        tuple[int]
```
adeak, line 2469:
```suggestion
        tuple[int, int, int]
```

pyvista/core/pointset.py
```diff
@@ -2489,7 +2490,7 @@ def visible_bounds(self) -> BoundsLike:
 
         Returns
         -------
-        tuple(float)
+        tuple[float]
```
adeak, line 2493:
```suggestion
        tuple[float, float, float]
```

pyvista/plotting/picking.py
```diff
@@ -264,33 +263,33 @@ def enable_cell_picking(
...
-        style : str, optional
+        style : str, default: "wireframe"
             Visualization style of the selection.  One of the
             following: ``style='surface'``, ``style='wireframe'``, or
-            ``style='points'``.  Defaults to ``'wireframe'``.
+            ``style='points'``.
 
-        line_width : float, optional
-            Thickness of selected mesh edges. Default 5.
+        line_width : float, default: 5
```
adeak, line 283:
```suggestion
        line_width : float, default: 5.0
```

pyvista/plotting/picking.py
```diff
@@ -667,39 +665,38 @@ def enable_point_picking(
...
 
-        point_size : int, optional
-            Size of picked points if ``show_point`` is
-            ``True``. Default 10.
+        point_size : int, default: 10
+            Size of picked points if ``show_point`` is ``True``.
 
-        use_mesh : bool, optional
+        use_mesh : bool, default: False
             If ``True``, the callback will be passed a pointer to the picked
```
adeak, line 682: Unrelated, but
```suggestion
            If ``True``, the callback will be passed the picked
```

pyvista/core/filters/data_set.py
```diff
@@ -874,15 +871,14 @@ def slice_along_line(self, line, generate_triangles=False, contour=False, progre
...
-        generate_triangles : bool, optional
-            If this is enabled (``False`` by default), the output will
-            be triangles. Otherwise the output will be the intersection
-            polygons.
+        generate_triangles : bool, default: False
+            When ``True``, the output will be triangles. Otherwise the output
+            will be the intersection polygons.
 
-        contour : bool, optional
+        contour : bool, default: false
```
adeak, line 878:
```suggestion
        contour : bool, default: False
```

pyvista/core/filters/data_set.py
```diff
@@ -1160,9 +1156,10 @@ def threshold_percent(
 
         Parameters
         ----------
-        percent : float or tuple(float), optional
-            The percentage (0,1) to threshold. If value is out of 0 to 1 range,
-            then it will be divided by 100 and checked to be in that range.
+        percent : float | sequence[float], optional
+            The percentage in the range ``(0, 1)`` to threshold. If value is
+            out of 0 to 1 range, then it will be divided by 100 and checked to
+            be in that range.
```
adeak, line 1162: This is a confusing API with a confusing description. I assumed this only takes percentages, which it does until you try to threshold by 0.5%. The way this is related to this PR is that we should at least try to make

> The percentage in the range ``(0, 1)`` to threshold.

less confusing. Not sure if calling it "[ratio] in the range" would be clearer/correct, for instance.

I'd consider deprecating the ratio feature sometime, potentially with the addition of a dedicated "threshold by ratio/proportion/etc." filter.

pyvista/core/filters/data_set.py
```diff
@@ -1472,33 +1469,33 @@ def elevation(
...
+            center of the dataset. Otherwise pass a length 3 sequence.
 
-        scalar_range : str or tuple(float), optional
+        scalar_range : str | sequence[float], optional
             The scalar range to project to the low and high points on the line
             that will be mapped to the dataset. If None given, the values will
             be computed from the elevation (Z component) range between the
             high and low points. Min and max of a range can be given as a length
-            2 tuple(float). If ``str`` name of scalara array present in the
+            2 sequence. If ``str`` name of scalar array present in the
```
adeak, line 1485:
```suggestion
            2 sequence. If ``str``, name of scalar array present in the
```

pyvista/core/filters/data_set.py
```diff
@@ -1578,40 +1575,40 @@ def contour(
 
         Parameters
         ----------
-        isosurfaces : int or sequence, optional
+        isosurfaces : int | sequence[float], optional
             Number of isosurfaces to compute across valid data range or a
             sequence of float values to explicitly use as the isosurfaces.
 
-        scalars : str, collections.abc.Sequence, numpy.ndarray, optional
+        scalars : sequence[float], optional
```
adeak, line 1582: This can also be the name, right? Perhaps
```suggestion
        scalars : str | sequence[float], optional
```
Or
```suggestion
        scalars : str | array_like[float], optional
```

pyvista/core/filters/data_set.py
```diff
@@ -3872,14 +3866,14 @@ def sample_over_multiple_lines(self, points, tolerance=None, progress_bar=False)
 
         Parameters
         ----------
-        points : np.ndarray or list
+        points : sequence[float]
```
adeak, line 3869: According to the example this can be a list of lists. Perhaps it should be another `array_like`?
```suggestion
        points : array_like[float]
```

pyvista/core/filters/data_set.py
```diff
@@ -4899,37 +4892,37 @@ def compute_derivative(
             derivative quantities.  Defaults to the active scalars in
             the dataset.
 
-        gradient : bool, str, optional
+        gradient : bool, str, default: True
```
adeak, line 4895: Should all these below parameters have `bool | str` instead of `bool, str`?

adeak (commented):
Partial review, 4 more big files to go later (`filters/poly_data.py`, `examples/downloads.py`, `plotting/plotting.py`, `plotting/renderer.py`).

Sorry if a handful of my comments are obsolete upon posting; I haven't checked them against the newest commits you've made.

pyvista/utilities/arrays.py
```diff
@@ -14,8 +14,8 @@ def _coerce_pointslike_arg(
 
     Parameters
     ----------
-    points : Sequence(float) or np.ndarray
-        Argument to coerce into (n, 3) ``np.ndarray``.
+    points : sequence[float]
```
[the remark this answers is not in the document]
MatthewFlamm: Can we use this yet? https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike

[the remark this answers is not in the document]
MatthewFlamm: I'm +1 on that strategy.  ArrayLike also was introduced in numpy 1.20.0, so we'd have to pin numpy version, which probably isn't desirable.

pyvista/core/filters/uniform_grid.py
```diff
@@ -245,22 +244,21 @@ def image_dilate_erode(
...
+        erode_value : float, optional
+            Erode value in the dataset. Default: ``0.0``.
 
-        kernel_size : list(int) or tuple(int), optional
-            Length 3 iterable of ints: ``(xsize, ysize, zsize)``.
-            Determines the size (and center) of the kernel.
-            Default: ``(3, 3, 3)``.
+        kernel_size : sequence[int], default: (3, 3, 3)
+            Determines the size (and center) of the kernel, for example
+            ``(xsize, ysize, zsize)``.
```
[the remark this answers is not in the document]
akaszynski: [Docs](https://vtk.org/doc/nightly/html/classvtkImageDilateErode3D.html
) state:

> This method sets the size of the neighborhood.

> It also sets the default middle of the neighborhood and computes the elliptical foot print.

I feel that we can safely ignore the second statement.

pyvista/core/filters/data_set.py
akaszynski: https://github.com/pyvista/pyvista/issues/4182

pyvista/core/filters/data_set.py
akaszynski: Yes, this is a good candidate for `array_like`.

pyvista/core/dataset.py
```diff
@@ -2208,11 +2208,11 @@ def find_closest_cell(
 
         Parameters
         ----------
-        point : Sequence(float) or np.ndarray
-            Coordinates of point to query (length 3) or a ``numpy`` array of ``n``
-            points with shape ``(n, 3)``.
+        point : sequence[float]
```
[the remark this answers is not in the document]
adeak: For a single point I'd be OK with `sequence`. But here we have `"or an array of shape (n, 3)"`. This sounds like array like including broadcasting (1d array or batch of 2d arrays).

pyvista/core/filters/data_set.py
akaszynski: Resolved in 367d26a111e306d978c835b03f1441d8472fe96d.

pyvista/plotting/_property.py
```diff
@@ -70,10 +70,10 @@ class Property(_vtk.vtkProperty):
         viewer.
 
     diffuse : float, optional
-        The diffuse lighting coefficient. Default 1.0.
+        The diffuse lighting coefficient.
```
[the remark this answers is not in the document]
akaszynski: Turns out the defaults are all set as global themes, so I referenced those in 367d26a111e306d978c835b03f1441d8472fe96d for `Property`.

pyvista/core/composite.py
```diff
@@ -494,7 +494,7 @@ def set_block_name(self, index: int, name: Optional[str]):
         index : int
             Index or the dataset within the multiblock.
 
-        name : str
+        name : str, optional
             Name to assign to the block at ``index``.
```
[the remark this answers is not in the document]
akaszynski: Fixed logic in 367d26a111e306d978c835b03f1441d8472fe96d.

pyvista/core/dataset.py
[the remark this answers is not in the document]
akaszynski: Agreed, didn't read the docstring...

pyvista/core/filters/uniform_grid.py
```diff
@@ -245,22 +244,21 @@ def image_dilate_erode(
...
         ----------
-        dilate_value : int or float, optional
-            Dilate value in the dataset. Default: ``1``.
+        dilate_value : float, optional
+            Dilate value in the dataset. Default: ``1.0``.
 
-        erode_value : int or float, optional
-            Erode value in the dataset. Default: ``0``.
+        erode_value : float, optional
+            Erode value in the dataset. Default: ``0.0``.
```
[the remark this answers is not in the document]
akaszynski: Resolved in 367d26a111e306d978c835b03f1441d8472fe96d.

pyvista/plotting/helpers.py
```diff
@@ -43,7 +43,7 @@ def plot(
 
     Parameters
     ----------
-    var_item : pyvista.DataSet, vtk, or numpy object
+    var_item : pyvista.DataSet | vtk | numpy object
```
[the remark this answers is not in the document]
akaszynski: Docs and acceptable types fixed in 367d26a111e306d978c835b03f1441d8472fe96d.

pyvista/core/pointset.py
akaszynski: Resolved in 367d26a111e306d978c835b03f1441d8472fe96d.

pyvista/core/dataset.py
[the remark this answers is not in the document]
akaszynski: ecc51f4a7a6bfd6260aa0f06d39e36aa1162d25f

akaszynski:
This is everything but `pyvista/plotting/widgets.py` since I know @banesullivan is doing a refactor there.

pyvista/utilities/geometric_objects.py
```diff
@@ -923,7 +923,7 @@ def Wavelet(
 
     Parameters
     ----------
-    extent : sequence[float], default: (-10.0, 10.0, -10.0, 10.0, -10.0, 10.0)
+    extent : sequence[int], default: (-10, 10, -10, 10, -10, 10)
```
tkoyama010, line 926: Good catch ([SetWholeExtent document](https://vtk.org/doc/nightly/html/classvtkRTAnalyticSource.html#ab7e362b69fd0f6cf9b1ad7d9baeba304)).

pyvista/plotting/helpers.py
```diff
@@ -150,17 +150,17 @@ def plot(
         a zoom-in, a value less than 1 is a zoom-out.  Must be greater than 0.
 
     border : bool, optional
```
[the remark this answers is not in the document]
akaszynski: Done.

pyvista/plotting/plotting.py
```diff
@@ -6368,30 +6366,28 @@ def show(
...
-            ``auto_close=False`` to set the scene, then save the
-            screenshot in a separate call to ``show()`` or
-            :func:`Plotter.screenshot`.
+        screenshot : str | pathlib.Path | io.BytesIO | bool, default: False
+            Take a screenshot of the initial state of the plot.  If a string,
+            it specifies the path to which the screenshot is saved. If
+            ``True``, the screenshot is returned as an array. For interactive
+            screenshots it's recommended to first call ``show()`` with
+            ``auto_close=False`` to set the scene, then save the screenshot in
+            a separate call to ``show()`` or :func:`Plotter.screenshot`.
```
adeak, line 6385:
```suggestion
            See also the ``before_close_callback`` parameter for an
            alternative.

```

pyvista/plotting/plotting.py
```diff
@@ -6721,18 +6716,14 @@ def add_cursor(
...
-        bounds : length 6 sequence, default: (-1.0, 1.0, -1.0, 1.0, -1.0, 1.0)
+        bounds : sequence[float], default: (-1.0, 1.0, -1.0, 1.0, -1.0, 1.0)
             Specify the bounds in the format of:
 
             - ``(xmin, xmax, ymin, ymax, zmin, zmax)``
 
-            Defaults to ``(-1.0, 1.0, -1.0, 1.0, -1.0, 1.0)``.
-
-        focal_point : list or tuple, optional
+        focal_point : list | tuple, default: (0.0, 0.0, 0.0)
```
adeak, line 6724:
```suggestion
        focal_point : sequence[float], default: (0.0, 0.0, 0.0)
```

pyvista/plotting/renderer.py
```diff
@@ -94,8 +94,8 @@ def scale_point(camera, point, invert=False):
     camera : Camera
         The camera who's matrix to use.
 
-    point : tuple(float)
-        Length 3 tuple of the point coordinates.
+    point : sequence[float, float, float]
```
adeak, line 97: Bunch of these all across this module as well...
```suggestion
    point : sequence[float]
```

pyvista/plotting/renderer.py
```diff
@@ -736,31 +736,30 @@ def add_actor(
 
         Parameters
         ----------
-        actor : vtk.vtkActor or vtk.vtkMapper
+        actor : vtk.vtkActor | vtk.vtkMapper, pyvista.Actor
```
adeak, line 739:
```suggestion
        actor : vtk.vtkActor | vtk.vtkMapper | pyvista.Actor
```

pyvista/plotting/renderer.py
```diff
@@ -2743,7 +2741,7 @@ def enable_depth_of_field(self, automatic_focal_distance=True):
         See :ref:`depth_of_field_example` for a full example using this method.
 
         """
-        self._render_passes.enable_depth_of_field_pass()
+        self._render_passes.enable_depth_of_field_pass(automatic_focal_distance)
```
adeak, line 2744: Sneaky bugfix.

pyvista/plotting/renderer.py
```diff
@@ -3160,17 +3158,17 @@ def add_legend(
...
             for a white color).  If None, legend background is
             disabled.
 
-        border : bool, optional
+        border : bool, default: False
             Controls if there will be a border around the legend.
             Default False.
 
-        size : sequence, optional
+        size : sequence[float, float], default: (0.2, 0.2)
```
adeak, line 3171:
```suggestion
        size : sequence[float], default: (0.2, 0.2)
```

pyvista/plotting/renderer.py
```diff
@@ -3193,7 +3191,7 @@ def add_legend(
             * ``'upper center'``
             * ``'center'``
 
-        face : str or pyvista.PolyData, optional
+        face : str | pyvista.PolyData, default: "triangle"
```
adeak, line 3194: Technically the valid input `None` is not covered by the current type line, but I don't have a concise way to fix this. Perhaps
```suggestion
        face : str | pyvista.PolyData | NoneType, default: "triangle"
```
Would that be better? I'm not sure.

pyvista/plotting/renderer.py
```diff
@@ -3565,7 +3562,7 @@ def add_legend_scale(
             Factor to scale label size relative to title size.
 
         label_format : str, optional
-            A printf style format for labels, e.g. '%E'.
+            A printf style format for labels, e.g. ``'%E'``.
```
adeak, line 3565: The Python intersphinx mapping has an item for `old-string-formatting` that maps to https://docs.python.org/3/library/stdtypes.html#old-string-formatting. Do you think it would be helpful to link to this for our users?

adeak (commented):
Second half of first pass. I don't expect to have too many further remarks though.

pyvista/utilities/geometric_objects.py
```diff
@@ -76,10 +76,10 @@ def Cylinder(
 
     Parameters
     ----------
-    center : sequence, default: (0.0, 0.0, 0.0)
+    center : sequence[float, float, float], default: (0.0, 0.0, 0.0)
```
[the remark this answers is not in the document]
akaszynski:
```suggestion
    center : sequence[float], default: (0.0, 0.0, 0.0)
```

pyvista/plotting/plotting.py
```diff
@@ -161,7 +161,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
 
     Parameters
     ----------
-    shape : list or tuple, optional
+    shape : sequence[int, int], optional
         Number of sub-render windows inside of the main window.
```
akaszynski, line 165:
```suggestion
        Two item sequence of sub-render windows inside of the main window.
```

pyvista/plotting/renderer.py
akaszynski:
```suggestion
            A printf style format for labels, e.g. ``'%E'``.
            See `old-string-formatting <https://docs.python.org/3/library/stdtypes.html#old-string-formatting>`_.
```

pyvista/plotting/renderer.py
akaszynski: Works for me.

pyvista/plotting/renderer.py
akaszynski: Wouldn't be an @akaszynski PR without one...

pyvista/plotting/renderer.py
```diff
@@ -3565,7 +3562,8 @@ def add_legend_scale(
...
+            See `old-string-formatting <https://docs.python.org/3/library/stdtypes.html#old-string-formatting>`_.
```
adeak, line 3566: Last comment, I promise :smile: I think we should make use of intersphinx if we can, so that if this URL is ever moved in the future we don't have to rely on a permanent redirect.

But I'm not sure how to do that... The intersphinx inventory contains this corresponding line:
```
old-string-formatting std:label -1 library/stdtypes.html#$ printf-style String Formatting
```
As opposed to, say,
```
set py:class 1 library/stdtypes.html#$ -
```
Do you think this means we could reference this as
```suggestion
            See :std:label:`old-string-formatting`.
```
? I don't speak sphinx well enough to make sure based on https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#module-sphinx.ext.intersphinx.

adeak (dismissed):
One future-proofing remark, but this is already fine as is if you don't want to bother with it. Thanks for the quick fixes @akaszynski!

pyvista/plotting/renderer.py
akaszynski: Fixed in 82209fc849bf65f68c199023597adf873692ef0c. Tried a few ways and this was the one that worked.

adeak (approved):

akaszynski:
As always, thanks @adeak!

PR #4161, opened 2023-03-20
Deprecate `inplace` paramter

### LINKED ISSUE #4158: Deprecate `inplace` parameter

It appears that the inplace parameter has been deprecated in pandas 
(see https://github.com/pandas-dev/pandas/issues/16529). 
Personally, I think PyVista should follow suit. Please let me know your opinion.

> Why is this bad?
Using inplace=True encourages mutation rather than immutable data, which is harder to reason about and may cause bugs. It also removes the ability to use the method chaining style for pandas operations.
>
> Further, in many cases, inplace=True does not provide a performance benefit, as pandas will often copy DataFrames in the background.

[Why You Should Probably Never Use pandas inplace=True](https://towardsdatascience.com/why-you-should-probably-never-use-pandas-inplace-true-9f9f211849e4)

### Links to source code.

```python
mercury.translate((0.0, 0.0, 0.0), inplace=True)
```

### Pseudocode or Screenshots

Use instead:
```python
mercury = mercury.translate((0.0, 0.0, 0.0))
```

tkoyama010:
### Overview

resolves #4158

 

### Details

- None


pyvista/core/datasetattributes.py
```diff
@@ -805,7 +805,7 @@ def _prepare_array(
 
         # Swap bytes from big to little endian.
         if data.dtype.byteorder == '>':
-            data = data.byteswap(inplace=True)
+            data = data = data.byteswap()
```
tkoyama010, line 808:
```suggestion
            data = data.byteswap(inplace=True)
```

pyvista/core/filters/poly_data.py
```diff
@@ -351,7 +351,7 @@ def __iadd__(self, dataset):
 
         """
         try:
-            merged = self.merge(dataset, inplace=True)
+            merged = self = self.merge(dataset)
```
tkoyama010, line 354:
```suggestion
            self = self.merge(dataset)
            merged = self
```

pyvista/core/filters/data_set.py
```diff
@@ -1822,18 +1821,25 @@ def texture_map_to_plane(
...
         t_coords.SetName(name)
         otc = self.GetPointData().GetTCoords()
-        self.GetPointData().SetTCoords(t_coords)
-        self.GetPointData().AddArray(t_coords)
-        # CRITICAL:
-        if otc and otc.GetName() != name:
-            # Add old ones back at the end if different name
-            self.GetPointData().AddArray(otc)
-        return self
+        if not inplace:
```
tkoyama010, line 1827:
```suggestion
        if inplace:
```

PR #4115, opened 2023-03-10
Add ruff to .pre-commit-config.yaml

tkoyama010:
### Overview

My friend @jrfk told me about [ruff](https://github.com/charliermarsh/ruff). It seems to be increasingly adopted in large projects such as Scipy and Sphinx. (See [Here](https://github.com/charliermarsh/ruff#whos-using-ruff))
Let's try it in PyVista too!

 

### TODO

- [ ] Fix the errors of ruff. 
   - [x] Using [flake8-to-ruff](https://pypi.org/project/flake8-to-ruff/).)
   - [ ] Check test codes have compatibility. 
   - [ ] Wait `inplace=True` deprecate PR #4161
- [x] Discuss removing the above tools.
- [x] Define explicitly which rule to use.

### Details

> Ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing tens or hundreds of times faster than any individual tool.

- [Python tooling could be much, much faster](https://notes.crmarsh.com/python-tooling-could-be-much-much-faster)
- [Ruff: The First 200 Releases](https://notes.crmarsh.com/ruff-the-first-200-releases)

PR #4170, opened 2023-03-21
Replace isort to ruff

tkoyama010:
### Overview

Replace isort to ruff.

 

### Details

- This is a 1st partial Pull Request of  #4115 .


banesullivan:
The imports should not be changing -- we should keep the same sort rule but just change the mechanism under which they are checked

pyproject.toml
```diff
@@ -172,3 +169,14 @@ image_cache_dir = "tests/plotting/image_cache"
...
+[tool.ruff]
+target-version = "py38"
+select = [
+    # isort
+    "I",
+]
+fixable = [
+    # isort
+    "I",
+]
```
tkoyama010, line 182:
```suggestion
]
[tool.ruff.per-file-ignores]
"pyvista/**/__init__.py,pyvista/__init__.py" = ["I"]
```

pyproject.toml
```diff
@@ -101,14 +101,11 @@ pyvista = [
     'uniform.vtk',
 ]
 
-[tool.isort]
-profile = 'black'
-line_length = 100
```
tkoyama010, line 106: Wee need `line_length` configuration and it is discussed in https://github.com/charliermarsh/ruff/issues/3206 .

MatthewFlamm:
In addition to line length, it looks like `import package` and `from package import thing` are handled differently.  I prefer the original implementation where it keeps these together.  The ruff version splits them into separate sections.

tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

tkoyama010:
It is still not compatible enough to replace ruff with isort. It is possible that ruff will be sufficiently compatible in the future, but until then it is not practical to maintain this PullRequest. We will close this PullRequest once it has been closed. We thank the reviewers for their cooperation.

PR #4171, opened 2023-03-21
Add devcontainer support

MatthewFlamm:
### Overview

 

This PR adds in devcontainer support.  You should be able to go to my fork at https://github.com/MatthewFlamm/pyvista/tree/maint/codespaces and spin up a codespace container for testing.

### Details

This includes pre-installation of requirements and testing requirements + a desktop-lite feature that allows for in browser vnc interaction

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,5 +3,5 @@
   "features": { "ghcr.io/devcontainers/features/desktop-lite:1": {}
   },
   "forwardPorts": [6080],
-  "postStartCommand": "./.devcontainer/postcreatecommand.sh"
+  "postCreateCommand": "./.devcontainer/postcreatecommand.sh"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,5 +2,6 @@
   "image": "mcr.microsoft.com/devcontainers/python:3.11",
   "features": { "ghcr.io/devcontainers/features/desktop-lite:1": {}
   },
-  "forwardPorts": [6080]
+  "forwardPorts": [6080],
+  "postCreateCommand": "./.devcontainer/postcreatecommand.sh"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,5 +3,5 @@
   "features": { "ghcr.io/devcontainers/features/desktop-lite:1": {}
   },
   "forwardPorts": [6080],
-  "postCreateCommand": "./.devcontainer/postcreatecommand.sh"
+  "postStartCommand": "./.devcontainer/postcreatecommand.sh"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,6 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/python:3.11",
+  "features": { "ghcr.io/devcontainers/features/desktop-lite:1": {}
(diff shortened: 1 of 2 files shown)


MatthewFlamm:
Screenshot_20230321-154947

MatthewFlamm:
We are getting errors on `test_ipygany`, `test_pythreejs`, and `test_trame` with the error message of `NotImplementedError: Cannot`, which is not helpful.  When running `pytest tests` inside the container, I only get one additional image regression test for `test_box_axes`.

MatthewFlamm:
There looks to be a workaround suggested in https://github.com/jupyter-widgets/ipywidgets/issues/3729

akaszynski:
> There looks to be a workaround suggested in [jupyter-widgets/ipywidgets#3729](https://github.com/jupyter-widgets/ipywidgets/issues/3729)

I'm on it.

Also, this is a hilarious error.

```
NotImplementedError: Cannot
```

MatthewFlamm:
> > There looks to be a workaround suggested in [jupyter-widgets/ipywidgets#3729](https://github.com/jupyter-widgets/ipywidgets/issues/3729)
> 
> I'm on it.

We might just need to wait.

https://github.com/ipython/comm/pull/13

akaszynski:
> 

I've implemented the fix in https://github.com/jupyter-widgets/ipywidgets/issues/3729#issuecomment-1478152137

It won't help with the integration tests, but it's better than waiting for a patch IMO because we can move on.

akaszynski (approved):
LGTM.

I'd love to see user-guide docs for this at some point.

MatthewFlamm:
> I'd love to see user-guide docs for this at some point.

Other than a sentence or two saying this is possible.  I can only think of two things that need to be mentioned.  1) which button to click to launch it, 2) which button to click inside the workspace to launch the inbrowser vnc and the password to use.  My only concern with detailed directions like this is that it could go stale quickly when GitHub makes any style changes or feature updates.

akaszynski:
> Other than a sentence or two saying this is possible. I can only think of two things that need to be mentioned. 1) which button to click to launch it, 2) which button to click inside the workspace to launch the inbrowser vnc and the password to use. My only concern with detailed directions like this is that it could go stale quickly when GitHub makes any style changes or feature updates.

Let's wait until things stabilize and we have lots of users. It may not happen, so let's not sweat it for the moment.

tkoyama010:
Memo. I finally found the vnc password on this line.
https://github.com/devcontainers/features/blob/0e079baaf7c67c7df9521d1dd5059a293702c5d5/src/desktop-lite/install.sh#L11
This feature is awesome! Thanks, @MatthewFlamm !

MatthewFlamm:
> Memo. I finally found the vnc password on this line.

This is probably worth putting in the documentation.  I'm sure there are instructions for how to override this in different ways for users who want more security, but this is out of scope information for our docs IMO.

MatthewFlamm:
Thinking about next steps:  My reading of https://docs.github.com/en/codespaces/prebuilding-your-codespaces/about-github-codespaces-prebuilds is that enabling prebuilds would consume GitHub action minutes and count against storage.  The container now is snappy enough (although I find the feedback of when it completes a little lacking), so there is no need. The big use case IMO would be prebuilt documentation, but using it with PRs might become complex.  Needs some thought.

tkoyama010:
Yes, I am thinking the same thing.
It might also be a good idea to add a code space badge to the README.rst to express support and make it easier for developers to access the environment.

PR #4179, opened 2023-03-24
Add texture features

### LINKED ISSUE #4178: Texture filtering
### Feature description

As far as know the only texture filtering available by default in PyVista is the nearest filter (equivalent to GL_NEAREST in OpenGL).
Even if it is the fastest filtering, the individual pixels of the texture are visible (image from PyVista doc):
image

It would be nice to add linear filtering or even texture mipmaps.

Such a feature seems to be available in VTK but I don't know in what extent https://vtk.org/doc/nightly/html/classvtkTexture.html
I'm not sure what is currently used in PyVista and how it could be implemented.

akaszynski:
This PR implements various features from #3431 including:
- Plotting textures as flat, "tight" plots.
- Plotting cubemaps
- Allowing for the instantiation of a Texture skymap from a list of UniformGrid.
- Better __repr__
- Resolve #4178 by adding interpolation and minimap control.
- Implement rotation.
- Fixes the broken [topo-map.py](https://docs.pyvista.org/examples/02-plot/topo-map.html#sphx-glr-examples-02-plot-topo-map-py) example:

sphx_glr_topo-map_001
sphx_glr_topo-map_004


banesullivan (dismissed):
Looking much nicer! Glad we finally fixed that topo example

akaszynski:
Merging this as this was approved prior to fixing one small test and merging with main (+24H).

PR #4188, opened 2023-03-25
Implement override mechanics for DataSet subclasses

adeak:
[This comment](https://github.com/pyvista/pyvista/pull/4146#issuecomment-1474338302) and its context (including https://github.com/pyvista/pyvista/pull/1584#issuecomment-901340448) got me thinking. We use a bunch of `pv.wrap(poly_data)` calls in `geometric_objects.py` to allow downstream libraries to swap out the wrapping type and replace types that our helper functions.

I started playing with alternatives a bit, and this is a first vague attempt at having more general mechanics for allowing downstream libraries to swap out types of our own (typically with subclasses of the corresponding types). There are some huge questions:
1. Does this even make sense?
2. Is this kind of black magic acceptable? Should we look for alternative implementations?

Some more specific questions:
<ol start="3">
  <li>What API do we actually want for this? Do we want a dedicated helper that overrides types? Or perhaps a mapping that keeps track of all overridden types? Latter would be messy because we'd have to keep the mapping in sync with the class-level overrides, unless we use an immutable mapping.</li>
  <li>Is it OK to consider <code>DataSet</code> subclasses to be overridden? This excludes <code>MultiBlock</code>. Should we hard-code the overridable types instead? That would mean that future new types might fall between the cracks and accidentally become nonoverridable.</li>
  <li>How do we document this? How private should these mechanics be?</li>
</ol>

---

Here's how this PR works as a user:
```py
#import traceback

import pyvista as pv

sphere = pv.Sphere()
print(type(sphere))

class LoudPolyData(pv.PolyData):
    def __init__(self, *args, **kwargs):
        print(f'Hi, my type is {type(self).__name__}')
        #print('traceback:')
        #traceback.print_stack()
        super().__init__(*args, **kwargs)

pv.PolyData.type_override = LoudPolyData

sphere_overridden = pv.Sphere()
print(type(sphere_overridden))
```

This creates a thin `PolyData` subclass that announces its existence when initialised. It overrides `pv.PolyData` with this type and creates a `pv.Sphere()` both before and after. Here's the output:
```
<class 'pyvista.core.pointset.PolyData'>
Hi, my type is LoudPolyData
Hi, my type is LoudPolyData
Hi, my type is LoudPolyData
<class '__main__.LoudPolyData'>
```
The first line comes from the first print in the script, using usual `PolyData`. The next 3 lines come from `pv.Sphere()`, and the last line is our explicit print. I was a bit surprised to see three `LouPolyData` being initialised, but it turns out that `pv.Sphere()` does `pv.wrap()` -> `rotate_y()` -> `translate()`, and while both filters are `inplace=True`, they first create a new `PolyData` then overwrite the original one to become a copy (which is probably [another nail in `inplace`'s coffin](https://github.com/pyvista/pyvista/issues/4158)). If you uncomment the `traceback`-related comments in the script you can trace where these three meshes are being created.

I'm curious to see what you all think. To be clear I'm not attached to any of the current implementation, this is more of an experiment. I'm OK with any outcome, including scrapping the whole idea. Perhaps it might be easier to come up with saner alternatives if we have this as a baseline. All feedback including design nitpicks are very welcome. Cc. @MatthewFlamm.

### DIFF
--- a/pyvista/__init__.py
+++ b/pyvista/__init__.py
@@ -26,6 +26,12 @@ from pyvista.utilities.sphinx_gallery import _get_sg_image_scraper
 
 from pyvista.utilities.wrappers import _wrappers
 
+# set up overrides once we're done with all the other imports
+from pyvista._overrides import set_up_overrides
+
+set_up_overrides()
+del set_up_overrides  # no longer necessary
+
 global_theme = _GlobalTheme()
 rcParams = _rcParams()  # raises DeprecationError when used
 
--- a/pyvista/_overrides.py
+++ b/pyvista/_overrides.py
@@ -1,0 +1,55 @@
+"""Functionality to override PyVista object creation with other subclasses."""
+from functools import wraps
+
+from pyvista.core import DataSet
+
+
+def set_up_overrides(root=DataSet):
+    """Set up type overriding for all subclasses of a PyVista type.
+
+    Setup entails recursively finding all subclasses of the root
+    type and defining an optional ``type_override`` class
+    attribute on each. __new__ is decorated for each type to
+    return the overriding type if provided.
+
+    Parameters
+    ----------
+    root : type, default: pyvista.DataSet
+        Parent type whose (recursive) subclasses will be set up.
+        Root itself is included.
+    """
+    for cls in OVERRIDABLES:
+        cls.type_override = None
+        old_new = cls.__new__
+
+        @wraps(old_new)
+        def new_new(cls, *args, __old_new=old_new, **kwargs):
+            other_cls = cls.type_override
+            if other_cls is None or other_cls is cls:
+                return __old_new(cls, *args, **kwargs)
+            # non-trivial override is set: return other type
+            instance = other_cls.__new__(other_cls, *args, **kwargs)
+            return instance
+
+        cls.__new__ = new_new
+
+
+def _find_subclasses(root):
+    """Recursively build a set of subclasses for a root type.
+
+    Parameters
+    ----------
+    root : type
+        Root type to find subclasses for.
+
+    Returns
+    -------
+    subclasses : set
+        Set of types containing (recursive) subclasses. Includes
+        the root type.
+    """
+    return {root}.union(*(_find_subclasses(child) for child in root.__subclasses__()))
+
+
+# mainly exposed for ease of resetting:
+OVERRIDABLES = frozenset(_find_subclasses(DataSet))
--- a/pyvista/_overrides.py
+++ b/pyvista/_overrides.py
@@ -32,7 +32,7 @@ def set_up_overrides(root=DataSet):
             instance = other_cls.__new__(other_cls, *args, **kwargs)
             return instance
 
-        new_new.__signature__ = inspect.signature(old_new)
+        new_new.__signature__ = inspect.signature(cls.__init__)
 
         cls.__new__ = new_new
 
--- a/pyvista/_overrides.py
+++ b/pyvista/_overrides.py
@@ -1,5 +1,6 @@
 """Functionality to override PyVista object creation with other subclasses."""
 from functools import wraps
+import inspect
 
 from pyvista.core import DataSet
 
@@ -30,6 +31,8 @@ def set_up_overrides(root=DataSet):
             # non-trivial override is set: return other type
             instance = other_cls.__new__(other_cls, *args, **kwargs)
             return instance
+
+        new_new.__signature__ = inspect.signature(old_new)
 
         cls.__new__ = new_new
 
--- a/pyvista/utilities/helpers.py
+++ b/pyvista/utilities/helpers.py
@@ -1436,11 +1436,18 @@ def abstract_class(cls_):
     Preventing a class from being instantiated similar to abc.ABCMeta
     but does not require an abstract method.
     """
+    orig_new = cls_.__new__
 
     def __new__(cls, *args, **kwargs):
         if cls is cls_:
             raise TypeError(f'{cls.__name__} is an abstract class and may not be instantiated.')
-        return object.__new__(cls)
+        try:
+            instance = orig_new(cls, *args, **kwargs)
+        except TypeError:
+            # pay for the sins of the parents
+            # (someone probably hard-coded object.__new__() upstream)
+            instance = orig_new(cls)
+        return instance
 
     cls_.__new__ = __new__
     return cls_
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -1812,10 +1812,17 @@ def test_point_neighbors_levels(grid: DataSet, i0, n_levels):
 
 def test_type_override(clean_up_type_overrides):
     class NotPolyData(pyvista.PolyData):
-        pass
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.additional_state = 42
 
     poly = pyvista.Sphere()
     assert type(poly) == pyvista.PolyData
+
     pyvista.PolyData.type_override = NotPolyData
     not_poly = pyvista.Sphere()
+    # check correct type
     assert type(not_poly) == NotPolyData
+    # check inherited and additional state
+    assert not_poly == poly
+    assert not_poly.additional_state == 42
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -29,9 +29,17 @@ from pyvista.utilities.misc import PyVistaDeprecationWarning
 HYPOTHESIS_MAX_EXAMPLES = 20
 
 
-@pytest.fixture()
+@pytest.fixture
 def grid():
     return pyvista.UnstructuredGrid(examples.hexbeamfile)
+
+
+@pytest.fixture
+def clean_up_type_overrides():
+    """Reset type overrides on teardown."""
+    yield
+    for cls in pyvista._overrides.OVERRIDABLES:
+        cls.type_override = None
 
 
 def test_invalid_overwrite(grid):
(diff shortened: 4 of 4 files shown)


pyvista/utilities/helpers.py
```diff
@@ -1436,11 +1436,18 @@ def abstract_class(cls_):
...
         if cls is cls_:
             raise TypeError(f'{cls.__name__} is an abstract class and may not be instantiated.')
-        return object.__new__(cls)
+        try:
+            instance = orig_new(cls, *args, **kwargs)
+        except TypeError:
+            # pay for the sins of the parents
+            # (someone probably hard-coded object.__new__() upstream)
+            instance = orig_new(cls)
+        return instance
```
adeak, line 1450: This fix was not strictly necessary, but I came across this while investigating things for this PR. I think we're better off supporting multiple inheritance in general.

pyvista/_overrides.py
```diff
@@ -0,0 +1,55 @@
...
+    -------
+    subclasses : set
+        Set of types containing (recursive) subclasses. Includes
+        the root type.
+    """
+    return {root}.union(*(_find_subclasses(child) for child in root.__subclasses__()))
+
+
+# mainly exposed for ease of resetting:
+OVERRIDABLES = frozenset(_find_subclasses(DataSet))
```
adeak, line 55: I'd have preferred `OVERRIDEABLES` but the linter disagreed.

pyvista/__init__.py
```diff
@@ -26,6 +26,12 @@
 
 from pyvista.utilities.wrappers import _wrappers
 
+# set up overrides once we're done with all the other imports
+from pyvista._overrides import set_up_overrides
+
+set_up_overrides()
+del set_up_overrides  # no longer necessary
```
adeak, line 33: The current API is that we set up all subclasses of `DataSet` to be overrideable, and the only point of entry into this feature is `pv.PolyData.type_override` etc. class attributes. Hence my lack of ideas about documenting this feature beyond a dedicated docs page.

tests/test_dataset.py
```diff
@@ -29,11 +29,19 @@
...
+@pytest.fixture
 def grid():
     return pyvista.UnstructuredGrid(examples.hexbeamfile)
 
 
+@pytest.fixture
+def clean_up_type_overrides():
+    """Reset type overrides on teardown."""
+    yield
+    for cls in pyvista._overrides.OVERRIDABLES:
```
adeak, line 41: It's probably not nice to use a "private" module in the test suite, but I didn't want to expose the module into the public namespace. But the `from pyvista._overrides` import ensures that `pyvista._overrides` is actually there if anyone goes looking.

adeak:
Build failure due to 2 copies of
```
WARNING: [numpydoc] Validation warnings while processing docstring for 'pyvista.PolyData':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'strips', 'n_lines', 'faces', 'n_strips', 'lines', 'n_faces', 'force_float', 'var_inp', 'force_ext', 'deep'}
```

I guess it finds the new wrapped `__new__` and is confused by its signature. Let's see if [ccd7cad](https://github.com/pyvista/pyvista/pull/4188/commits/ccd7cad0d4db39a61bc833042b75a21b71a41825) fixes it...

MatthewFlamm:
My first thought based on the structure, but not looking into the details, is that this seems like a better design than the current state, which is a bit hackish.

One thing Id recommend is that we consider how any strategy fits into using the override method in newer vtk versions. It would be good to have a general plan for this so that we don't need to have another big change for this.

adeak:
Great point, thanks for the quick look @MatthewFlamm. Indeed it makes a lot of sense to make the override mechanics similar to VTK's one. Which I'm completely oblivious to, so I'll look into this :smile:

pyvista/_overrides.py
tkoyama010: How about split like `OVERRIDE_ABLES` ?

tests/test_dataset.py
```diff
@@ -29,11 +29,19 @@
...
-@pytest.fixture()
+@pytest.fixture
 def grid():
     return pyvista.UnstructuredGrid(examples.hexbeamfile)
 
 
+@pytest.fixture
+def clean_up_type_overrides():
+    """Reset type overrides on teardown."""
+    yield
```
tkoyama010, line 40: Do we need this line?

pyvista/_overrides.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Parameters
+    ----------
+    root : type, default: pyvista.DataSet
+        Parent type whose (recursive) subclasses will be set up.
+        Root itself is included.
+    """
+    for cls in OVERRIDABLES:
+        cls.type_override = None
+        old_new = cls.__new__
```
tkoyama010, line 24: I think it would be more readable to use "old" as "base" (naming from the base class) since it does override :)
```suggestion
        base_new = cls.__new__
```

tkoyama010: Or you may use `orig` to match helpers.py.
```suggestion
        orig_new = cls.__new__
```

tkoyama010 (commented):
1. Personally, I could understand it, but that may be because I faced this problem.
2. According to @MatthewFlamm , overrides are supported in newer versions of VTK. This can be considered a temporary workaround until then.
3. Since this is likely to be a feature that will be removed in the future, it would be good to provide the API that is easiest to implement. If it is not removed for a long time and there are requests, we can modify it.
4. I can't think of a solution.
5. As mentioned above, I think this is a short-term feature, so I don't think we should spend a lot of effort on documentation. If a question comes to the community, it should be answered as a hidden feature.

tkoyama010:
Oops, didn't check my earlier comment.

> Indeed it makes a lot of sense to make the API of the override mechanics similar to VTK's one.

+1 to this.

pyvista/_overrides.py
adeak: Thanks, but I don't think that would work because [it's one word](https://en.wiktionary.org/wiki/overrideable). Although the linked wiki page says
>     overridable is much more common than overrideable and nonoverridable; unoverridable is so rare as not to be found in Google Ngram Viewer.

I originally only looked at [ridable](https://en.wiktionary.org/wiki/ridable) where this form seems to be less common. But the specific wiktionary page agrees with the linter, so it's all good.

pyvista/_overrides.py
adeak: I thought it was funny to use this naming :smile: But thanks, you're right, it should be something clearer instead. I'll hold off a bit on choosing a new name, because `orig_new` in `helpers.py` is also something I just made up, so I'm not the best judge. We might want to rename both to something even better.

adeak:
Thanks for your thoughts @tkoyama010.

For two related points:

> 2. According to @MatthewFlamm , overrides are supported in newer versions of VTK. This can be considered a temporary workaround until then.
> 3. Since this is likely to be a feature that will be removed in the future, it would be good to provide the API that is easiest to implement. If it is not removed for a long time and there are requests, we can modify it.

I think the overriding will have a long-term place for us. My understanding is that the VTK override mechanism lets you swap out, say, `vtkPolyData`. But our downstream users will almost all want to swap out our own pure-python subclasses with pure-python subclasses of their own. In other words, I think they'll want to swap out `pyvista.PolyData` directly. And this is exactly that the current wrapper mechanism allows them to do.

So if I'm right about the above, this feature will be here to stay, but still it's very helpful to downstream users to match the user-facing mechanics of VTK if that's reasonably possible. Or at least provide that and additional bells and whistles if we want to. Then downstream users (libraries) don't have to keep track of two kinds of overrides for the same domain.

tests/test_dataset.py
adeak: Yes, otherwise the cleanup would run _before_ the test. Turning the fixture into a generating function splits it into "setup - yield - teardown".

adeak:
Note to self: also look for any stakeholders at https://github.com/pyvista/pyvista/pull/1584 who would use this once we've started to converge on a design.

adeak:
So the VTK API looks like this, quoting the docs of `vtkPolyData.override()`:
```
This method can be used to override a VTK class with a Python subclass.
The class type passed to override will afterwards be instantiated
instead of the type override is called on.
For example,

class foo(vtk.vtkPoints):
  pass
vtk.vtkPoints.override(foo)

will lead to foo being instantied everytime vtkPoints() is called.
The main objective of this functionality is to enable developers to
extend VTK classes with more pythonic subclasses that contain
convenience functionality.
```

So if we want to follow this we'd have to slap on a _method_ to our meshes, which for some reason seems more iffy to me. And the API of `PolyData.override(PolyDataSubclass)` seems less pythonic to me than using a simple attribute with an informative name. And with an attribute we can easily keep track of the current override state, otherwise we'd need some kind of "override getter" to see what a class was overridden with. I suspect VTK went with a method because that's far easier to implement in C++...

Of course we can hack the same API as a helper, but still offer (and recommend to use) something like my current implementation with `PolyData.type_override = PolyDataSubclass`.

tkoyama010:
@adeak
Sorry. Could you give us the link to it? I tried to find it but I cannot. 
However, I found the [experimental feature of vtk](https://vtk.org/doc/nightly/html/md__builds_gitlab_kitware_sciviz_ci_Documentation_Doxygen_PythonWrappers.html).
image

adeak:
> However, I found the [experimental feature of vtk](https://vtk.org/doc/nightly/html/md__builds_gitlab_kitware_sciviz_ci_Documentation_Doxygen_PythonWrappers.html).

Good find @tkoyama010. The example demonstrates that the method approach has the benefit of being used asa decorator. This has convinced me to indeed follow that API (but probably by keeping the current one).

Unfortunately I couldn't find any online docs, but the VTK docs are always hostile territory to me. I just installed VTK 9.2 and checked the interactive help. Then based on that I found [the implementation](https://gitlab.kitware.com/vtk/vtk/-/blob/master/Wrapping/PythonCore/PyVTKObject.cxx#L57-133).

adeak:
Thanks for the link @banesullivan. On a quick glance I can't immediately see the issue, but I've yet to take a closer look.

PR #4202, opened 2023-03-29
devcontainer enhancement

MatthewFlamm:
### Overview

Adds gh cli in the container.

Runs setup in "onCreateCommand" which runs before the codespace pops up.  Using the existing "postCreateCommand" means that the setup is run after the IDE opens up.  This allows you to get to the files quicker, but it could create issues if a user immediately gets a python REPL and does `import pyvista`.  This is a tradeoff for making things more robust at the cost of speed.  Looking for feedback on this.

Adds the most useful defaults vscode settings.  Python extension, and enabling pytest test collection.


.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,22 @@
 {
   "image": "mcr.microsoft.com/devcontainers/python:3.11",
-  "features": { "ghcr.io/devcontainers/features/desktop-lite:1": {}
+  "features": { 
+	"ghcr.io/devcontainers/features/desktop-lite:1": {},
+	"ghcr.io/devcontainers/features/github-cli:1": {}
```
tkoyama010, line 5: Excellent addition, just one question I am concerned about. This PullRequest will use multiple DockerImage, will there be any conflicts? I would catch any information on this.

tkoyama010: Sorry. It's my misunderstanding. It features, not images so there is no conflict.

tkoyama010 (approved):
WFM. Thanks!

.devcontainer/devcontainer.json
MatthewFlamm: Features are a recent addition to devcontainers.  Something that previously required more complex docker container handling, and one reason I was attracted to Gitpod previously.  The ones here are supported by GitHub directly, which is important for maintainability.

Issue #3738, opened 2022-12-21
Pyvista cannot draw and show all edges of the mesh


LXHCSMX:
### Describe the bug, what's wrong, and what you expected.

I found a bug!

When I used Pyvista to display the mesh model in vtk file, I found that Pyvista could not draw and show all edges of the mesh.
The following file shows the data in vtk file.
Delete the. txt suffix and you can use it.
Please help solve this problem. Many thanks.
[test.vtk.txt](https://github.com/pyvista/pyvista/files/10276098/test.vtk.txt)

### Steps to reproduce the bug.

```python
import pyvista as pv

my_mesh = pv.read("D:/cae_test/mesh_file/test.vtk")
plotter = pv.Plotter()
_ = plotter.add_mesh(my_mesh, color = '#BBFFFF', style = 'surface', show_edges = True,
                     edge_color = '#EE8262', line_width = 0.5, opacity = 1.0)
plotter.show()
```

[test.vtk.txt](https://github.com/pyvista/pyvista/files/10276081/test.vtk.txt)

### System Information

```shell
OS : Windows
            CPU(s) : 20
           Machine : AMD64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : Intel
      GPU Renderer : Intel(R) Iris(R) Xe Graphics
       GPU Version : 4.5.0 - Build 31.0.101.3616

  Python 3.8.15 (default, Nov 21 2022, 10:47:13) [MSC v.1916 64 bit (AMD64)]

           pyvista : 0.37.0
               vtk : 9.2.2
             numpy : 1.23.5
           imageio : 2.23.0
            scooby : 0.7.0
             pooch : v1.6.0
        matplotlib : 3.6.2
         pyvistaqt : 0.9.0
             PyQt5 : 5.15.2
            meshio : 5.3.4
```

### Screenshots

![fig3](https://user-images.githubusercontent.com/39549745/208860802-89d58b53-5674-492a-b323-b53fd0c9ab52.png)
![fig2](https://user-images.githubusercontent.com/39549745/208860811-cbe1d687-acf7-4756-a3c1-d6500a6675ad.png)
![fig1](https://user-images.githubusercontent.com/39549745/208860813-775feae1-67fd-48db-a379-f62bf6fa86f8.png)

banesullivan:
[...]

PR #4204, opened 2023-03-30
Add extract_cells_by_type

banesullivan:
This is a filter to help resolve https://github.com/pyvista/pyvista/issues/3738 and issues like it.

`extract_cells_by_type()` is a new dataset filter that wraps `vtkExtractCellsByType` to clean meshes with specific cell types. This is especially useful when needing to remove vertex cells from a mesh that is a surface mesh.

Useability is my last concern, as this requires users to know exactly what cell type to extract and how to look up its integer value. Some work is needed here to streamline access to the cell-type integers.


banesullivan:
Aha! Thanks for pointing me to `CellType`, @adeak! Totally missed that while implementing this before my coffee this morning. Will update to leverage that

pyvista/core/filters/data_set.py
```diff
@@ -5535,6 +5535,56 @@ def separate_cells(self):
...
+
+        .. warning::
+            Unlike the filter vtkExtractCells which always produces
+            vtkUnstructuredGrid output, this filter produces the same output
+            type as input type (i.e., it is a vtkDataSetAlgorithm). Also,
+            vtkExtractCells extracts cells based on their ids.
+
+        Parameters
+        ----------
+        cell_types : bool | int | list[int]
```
MatthewFlamm, line 5565: What is the use case for `cell_types=True` here?  Won't this always return the same data object back if it has cells?  My reading of `vtkExtractCellsByType` is that `AddAllCellTypes` could be useful in conjunction with `RemoveCellType` to select all types but one, for example.  If this is correct, I'd suggest dropping this usage from this method.  Instead, we could use a convenience method for generating a list of all cell types using `pyvista.CellType`.  For example, can we do something like:

```py
cell_types = pv.CellType.all_cells(exclude=pv.CellType.QUADRATIC_EDGE)
mesh.extract_cells_by_type(cell_types)
```

Bonus points if the design in `CellType` could allow us to distinguish between different flavors based on some criteria, e.g. dimensionality  `pv.CellType.all_edge_cells().  This is out of scope for this PR, but I'm trying to provide some argument for the design of this method to stay out of the game of trying to configure cell types via input.

akaszynski: Resolved this in [220f426](https://github.com/pyvista/pyvista/pull/4204/commits/220f4266f267a89d2c6e0b1794fb1f209f7ab598).

akaszynski:
> Useability is my last concern, as this requires users to know exactly what cell type to extract and how to look up its integer value. Some work is needed here to streamline access to the cell-type integers.

I think we've simplified this with an example using [pyvista.CellType](https://docs.pyvista.org/version/stable/api/utilities/_autosummary/pyvista.CellType.html) as suggested by @adeak.

akaszynski (approved):

PR #4211, opened 2023-03-31
Add Trims trailing whitespace in pre-commit-hooks

tkoyama010:
### Overview

Add Trims trailing whitespace in pre-commit-hooks.

 
- [ ] Discuss adding a commit to blame ignore. 

### Details

- None


tkoyama010:
We need to exclude vtk and ply files.

tkoyama010:
@adeak Should I add a commit to blame ignore, even though this change is slight?

adeak:
Thanks for the ping :) Since the diff is small and .pre-commit-config.yaml is changed, I think we shoukd keep this in the blame.

akaszynski (approved):
Agree with @adeak, this does not need to be excluded from the blame.

LGTM, thanks @tkoyama010

PR #4214, opened 2023-04-01
Add 'connected' argument to 'Plotter.add_lines' to create a series of connected lines

### LINKED ISSUE #4065: Add support for previous behaviour of Plotter.add_lines

Hi, I've just been trying out the latest release and have noticed that the behaviour of `Plotter.add_lines` has changed to match the description in the docs (see #3667). Previously, `pyvista.lines_from_points` was used to generate the lines, whereas now `pyvista.line_segments_from_points` is used.

I've been using `Plotter.add_lines` with the previous (incorrect) behaviour and found it super useful. It would be great if this was supported, either as a new function e.g. `Plotter.add_lines_from_points` or as an argument to `Plotter.add_lines` e.g. `Plotter.add_lines(from_points=True)`

Thanks!

_No response_

### Pseudocode or Screenshots

_No response_

p-j-smith:
### Overview
Add support for using `Plotter.add_lines` to create a series of connected lines from an array of points, i.e. the behaviour of this function prior to #3687

 
resolves #4065

### Details

- add a `connected` argument to `Plotter.add_lines`. If `connected=True`, the points passed to `Plotter.add_lines` will be taken to form a series of connected lines. The default is `False` to maintain the current behaviour of the function

- add a test to check the correct number of line segments is created when setting `connected=True`


pyvista/plotting/plotting.py
```diff
@@ -5055,12 +5060,15 @@ def add_lines(self, lines, color='w', width=5, label=None, name=None):
         >>> actor = pl.add_lines(points, color='yellow', width=3)
```
adeak, line 5060: Tangentially related to this PR, but
1. the yellow-on-white is a very bad choice, I recommend changing this while you're at it
2. the doctest here uses a good example, but the docstring above has `np.array([[0, 0, 0], [1, 0, 0], [1, 0, 0], [1, 1, 0]])` as an example which is ambiguous. I recommend changing this to have 4 distinct points.

pyvista/plotting/plotting.py
```diff
@@ -5041,6 +5041,11 @@ def add_lines(self, lines, color='w', width=5, label=None, name=None):
             If an actor of this name already exists in the rendering window, it
             will be replaced by the new actor.
 
+        connected: bool, default: False
+            Treat ``lines`` as points representing a series of *connected* lines.
+            For example, two connected line segments would be represented as
+            ``np.array([[0, 0, 0], [1, 0, 0], [1, 1, 0]])``.
```
adeak, line 5047: For the sake of unambiguity I'd mention that in the `False` case we need an even number of points. This might help drive the message home.

pyvista/plotting/plotting.py
p-j-smith: 1. good point - I've changed the line colour to purple
2. I've changed it to use the same coordinates as the doctest

pyvista/plotting/plotting.py
p-j-smith: I've added a sentence to hopefully make this clearer, but you're right that's it's probably still not super clear for someone that hasn't used this function before

adeak: I think most of the ambiguity comes from the fact that "lines" actually means "2-length/primitive lines" whereas a polyline can be arbitrarily long. But I'm not sure it's worth trying to pin down the language; the doctest complete with picture should be clear enough. Hmm, or perhaps we should add the same function call but with the new parameter, to show how _that_ looks. What do you think?

p-j-smith: > perhaps we should add the same function call but with the new parameter, to show how that looks

I like that! It should clear things up

pyvista/plotting/plotting.py
```diff
@@ -5041,26 +5041,53 @@ def add_lines(self, lines, color='w', width=5, label=None, name=None):
...
         >>> import pyvista
         >>> pl = pyvista.Plotter()
         >>> points = np.array([[0, 1, 0], [1, 0, 0], [1, 1, 0], [2, 0, 0]])
-        >>> actor = pl.add_lines(points, color='yellow', width=3)
+        >>> actor = pl.add_lines(points, color='purple', width=3)
+        >>> pl.camera_position = 'xy'
+        >>> pl.show()
+
+        Adding lines with ``connected=True`` will add a series of connected
+        lines segments.
```
adeak, line 5070:
```suggestion
        line segments.
```

pyvista/plotting/plotting.py
```diff
@@ -5041,26 +5041,53 @@ def add_lines(self, lines, color='w', width=5, label=None, name=None):
...
+
+        >>> import numpy as np
+        >>> import pyvista
+        >>> pl = pyvista.Plotter()
+        >>> points = np.array([[0, 1, 0], [1, 0, 0], [1, 1, 0], [2, 0, 0]])
+        >>> actor = pl.add_lines(
+        ...     points, color='purple', width=3, connected=True
+        ... )
```
adeak, line 5078: We don't have to start from scratch here, I think we only have to create a new plotter. See e.g. [the docs of `collision()`](https://dev.pyvista.org/api/core/_autosummary/pyvista.PolyDataFilters.collision.html).

p-j-smith: that's neater - I'll fix that now, thanks

adeak:
@p-j-smith although we don't have an official policy about this, I recommend merging instead of rebasing and force pushing on PRs unless you have a strong opinion of your own in this regard :) We use squash merge so the local history on PRs is not a concern, and rebasing runs the risk of losing commits pushed by others (not a concern in your case probably, because you know what you're doing) and not being able to use the "changes since you last viewed" button on github that helps reviewing.

pyvista/plotting/plotting.py
```diff
@@ -5041,26 +5041,51 @@ def add_lines(self, lines, color='w', width=5, label=None, name=None):
...
-        >>> actor = pl.add_lines(points, color='yellow', width=3)
+        >>> actor = pl.add_lines(points, color='purple', width=3)
+        >>> pl.camera_position = 'xy'
+        >>> pl.show()
+
+        Adding lines with ``connected=True`` will add a series of connected
+        line segments.
+
+        >>> pl = pyvista.Plotter()
+        >>> points = np.array([[0, 1, 0], [1, 0, 0], [1, 1, 0], [2, 0, 0]])
```
adeak, line 5073: This is just nitpicking but I wouldn't redefine the same array here, emphasizing that it's the same data with just a different kwarg being plotted. But I don't insist if you prefer it this way.

adeak (approved):
Thanks for the contribution @p-j-smith. One comment, works for me with or without that.

p-j-smith:
>  I recommend merging instead of rebasing and force pushing on PRs

sorry about that! I do usually rebase and force push but I'm happy to merge in future pyvista prs

pyvista/plotting/plotting.py
p-j-smith: ah, I was thinking it would be useful to be see the data plotted directly above the plot itself rather than needing to scroll above the first plot. But I'm happy to remove it if anyone else thinks it would be clearer

tkoyama010 (approved):
@p-j-smith LGTM. Thanks for this. It is perfect!

p-j-smith:
Thanks for the fast reviews @adeak and @tkoyama010 ! Would one of you mind merging as I'm not authorised? Thanks!

tkoyama010:
We usually check that there are no objections 24 hours after the initial approval is given. This is to give accounts in all time zones a fair opportunity to comment. Already, I am going to merge because it has been 24 hours since the first approval.

PR #4223, opened 2023-04-03
Improve performance of Trame camera sync

banesullivan:
Some performance improvements to the Trame viewer.

This solves https://github.com/pyvista/pyvista/discussions/4055 to improve the camera controls when there are many actors in the scene

I need to do a bit more validation testing on these changes to make the the `update()` -> `update_camera()` change on `render()` calls doesn't break any behavior


banesullivan:
> I need to do a bit more validation testing on these changes to make the the update() -> update_camera() change on render() calls doesn't break any behavior

This breaks some animation workflows. I will create a validation example and fix

examples_trame/validation/ManyActors.py
akaszynski, line 1: Should we make these file names snakecase?

banesullivan: Yeah.... I should and will.

banesullivan: Okay done. Except I left three files CamelCase as they are examples demonstrating specific classes

akaszynski (approved):
LGTM

PR #4227, opened 2023-04-03
Fast merge PolyData

### LINKED ISSUE #3920: Make point-merging behavior of `PolyDataFilters.merge()` consistent with `MultiBlock.combine()` and `DataSetFilters.merge()`

akaszynski:
Resolves #3920 by implementing @whophil's suggestion always using `vtkAppendFilter` when merging PolyData and then casting back.

The speedup is noticeable where it counts. Seems that the point merging algorithm grows quadraticly  with `vtkCleanPolyData`.

### 1M points example
```py
>>> import pyvista as pv
>>> sphere_a = pv.Sphere(theta_resolution=1000, phi_resolution=1000)
>>> sphere_b = sphere_a.copy()
>>> timeit sphere_a.merge(sphere_b, merge_points=True)
```

#### `main` branch
```
1.48 s ± 6.07 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
#### This branch
```
473 ms ± 3.39 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### 4M points example
```py
>>> import pyvista as pv
>>> sphere_a = pv.Sphere(theta_resolution=2000, phi_resolution=2000)
>>> sphere_b = sphere_a.copy()
>>> timeit sphere_a.merge(sphere_b, merge_points=True)
```

#### `main` branch
```
17.4 s ± 29.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

#### This branch
```
1.97 s ± 21.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -417,42 +417,29 @@ class PolyDataFilters(DataSetFilters):
         """
         # check if dataset or datasets are not polydata
         if isinstance(dataset, (list, tuple, pyvista.MultiBlock)):
-            not_pd = any(not isinstance(data, pyvista.PolyData) for data in dataset)
+            is_pd = all(isinstance(data, pyvista.PolyData) for data in dataset)
         else:
-            not_pd = not isinstance(dataset, pyvista.PolyData)
+            is_pd = isinstance(dataset, pyvista.PolyData)
 
-        # use dataset merge if not polydata
-        if not_pd:
-            return DataSetFilters.merge(
-                self,
-                dataset,
-                merge_points=merge_points,
-                main_has_priority=main_has_priority,
-                inplace=inplace,
-            )
+        if inplace and not is_pd:
+            raise TypeError("In-place merge requires both input datasets to be PolyData.")
 
-        append_filter = pyvista._vtk.vtkAppendPolyData()
+        merged = DataSetFilters.merge(
+            self,
+            dataset,
+            merge_points=merge_points,
+            main_has_priority=main_has_priority,
+            inplace=False,
+            progress_bar=progress_bar,
+        )
 
-        # note: unlike DataSetFilters.merge, we must put the
-        # "to be preserved" dataset last due to the call to clean()
-        if main_has_priority:
-            append_filter.AddInputData(self)
-
-        if isinstance(dataset, pyvista.DataSet):
-            append_filter.AddInputData(dataset)
-        else:
-            for data in dataset:
-                append_filter.AddInputData(data)
-
-        if not main_has_priority:
-            append_filter.AddInputData(self)
-
-        _update_alg(append_filter, progress_bar, 'Merging')
-        merged = _get_output(append_filter)
-        if merge_points:
-            merged = merged.clean(
-                lines_to_points=False, polys_to_lines=False, strips_to_polys=False
-            )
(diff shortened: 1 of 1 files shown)


banesullivan (approved):
That's quite the speedup

whophil:
LGTM, thanks for implementing @akaszynski!

I just want to add that this change is slightly more than a mere performance twiddle. PyVista will no longer use the `vtkAppendPolyData` filter at all, when it used to do so when merging all-polydata sets. Perhaps this has some other implications.

whophil (approved):

MbBrainz:
There is a very important implication of this change which is definitely breaking. I have a piece of code that relies on the order of the points after the merge, which is preserved in Pyvista v0.38.1, but not in v0.39.0. 

I've spent a week to figure this out. Of course this is a unstable release and its part of the usage risk, however it would be great to have this ordering of points as part of the testsuite. 

This change should be explicitly mentioned in the documentation or fixed. Im not familiar with the exact workings of the underneath vtk funtionality, but i can imagine that the changing of the filter usage discussed above may have introduced this different behaviour. In that case, a solution would be to add a flag to the PolyData.merge functionality. Something like "preserve_order"=True

beroda:
@banesullivan @akaszynski I am experiencing the same issue as @MbBrainz.

Switching to pyvista 0.39 in our code base at work caused some tests to crash since we rely on the points order for FEM applications.
Is there a workaround ? If not, do you mind if I open an issue to solve that ?
I could propose a PR if needed.

beroda:
@banesullivan No worries, an edge-case is ... an edge-case :wink:

If it is fine for you, I think I will go for the `PolyData.append_polydata` solution, and add a warning to the `merge` documentation.

banesullivan:
Looks like @whophil's foresight was on the nose here -- sounds beneficial to add back a method to use `vtkAppendPolyData`. @beroda go ahead with that approach.

Issue #1579, opened 2021-08-09
Default point of rotate_* filters


tkoyama010:
Some filters default to the mesh center, and that would actually make sense here. Too bad the rotate_* filters don't do that already.

_Originally posted by @adeak in https://github.com/pyvista/pyvista/pull/1569#discussion_r684923956_

adeak:
Thanks for opening this. However, I'm a bit worried about changing the default here. This would mean that code that used to work still works, but suddenly and silently gives different results, almost always (since, statistically speaking, most meshes aren't centered on the origin).

tkoyama010:
@adeak Thanks for comment. It is a very good point. We should be cautious about the impact on users. What do you think about creating a new method with a slightly different name? (And Deprecate Warning in old method.)

adeak:
Honestly, I have no idea what the best strategy is. Options I can think of:

1. do nothing
2. just change the default
3. add a loud warning telling users that the behaviour will change in `n` (`n=2`?) versions, and they should be explicit about the center in the meantime
4. deprecate the old function and add a new one, as you suggested.

If we go with 1. there's also a side-concern:
    1.b. do we prefer to have the new flip_* methods have sane default, or do we prefer to have the new flip_* methods match the defaults of the rotate_* methods? We can't have both.
    
@akaszynski thoughts?

akaszynski:
Good discussion.

Do nothing with existing methods.

We then add new methods to allow rotation about center (I think `rotate_x_about_center` sounds good).  Same with the flip methods.

Thoughts?

tkoyama010:
@akaszynski Thanks for comment. +1 for adding `rotate_x_about_center` .

MatthewFlamm:
There are already a lot of methods for datasets.  It is always growing too.  I'm not enthusiastic about adding more functions to get around a default parameter issue.  But, the problem with option 3 is that there is no way to future-proof your code without unneeded code logic, i.e. everyone would have to specify the point they want until the deprecation period is over and there is no longer a point in having a default.   So I reluctantly also vote for option 1 or 1a for now.

The only thing I would consider for option 3 is to write down the preferred default behaviors and then identify all the inconsistent methods. The we could have the breaking changes in the first major release, i.e. v1.0.0.  This is a lot of effort required to implement however.

akaszynski:
> There are already a lot of methods for datasets

[...]

PR #4228, opened 2023-04-03
Change development status to stable

### LINKED ISSUE #3975: PyPI Development Status
I just noticed that PyVista is listed as "Beta"... should we update this to `Development Status :: 5 - Production/Stable`? We're still sub version `1.x` (arbitrarily) so I guess "Beta" fits, but IMO does not reflect how PyVista is being used.

https://github.com/pyvista/pyvista/blob/bdb60e1f7ff8872ba9073d194825aa9c1726ced7/pyproject.toml#L16

ref https://pypi.org/classifiers/

akaszynski:
Resolve #3975 by changing our "Development Status" to "Production/Stable".

Knowing how PyVista is being used ([~1.6k connections](https://github.com/pyvista/pyvista/network/dependents)!), I'm happy to take this a little closer to a v1.0 release.

Considering there are likely more changes to be made with the API (and it will probably never stop), I'm fine with delaying it past the end of this year. However, I also think we have enough API coverage as-is to call what we have as a 1.0.

It's for everyone to decide when, but I'm happy either way and regardless, I think we're "stable enough." right now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ requires-python = '>=3.8'
 keywords = ['vtk', 'numpy', 'plotting', 'mesh']
 license = {text = 'MIT'}
 classifiers = [
-    'Development Status :: 4 - Beta',
+    'Development Status :: 5 - Production/Stable',
     'Intended Audience :: Science/Research',
     'Topic :: Scientific/Engineering :: Information Analysis',
     'License :: OSI Approved :: MIT License',


tkoyama010 (approved):
I am sad that the fun times before the 1.0 release are over, but I have to admit that the API is more stable. I agree with this change.
Are there any criteria that other projects use when deciding to make the OSS version 1.0? I would like to know if anyone knows.

adeak:
> I also think we have enough API coverage as-is to call what we have as a 1.0.

I think in terms of API we have too much in places. Before we go 1.0 we should very seriously consider the major changes we still want to do and which will be harder after 1.0. Off the top of my head I can name https://github.com/pyvista/pyvista/pull/3764 (which drowned in "meh" and I ran out of time while I was still invested; woukd have to restart from scratch) and https://github.com/pyvista/pyvista/issues/4158. I'm sure there are other such examples if we dig into issues (e.g. https://github.com/pyvista/pyvista/issues?q=is%3Aissue+is%3Aopen+deprecate).

tkoyama010 (changes requested):
Due to my whims, I had completely forgotten about these discussions lol . I guess anything that breaks backward compatibility like #1579 needs to be carefully considered. In any case, @akaszynski  has said he will release 1.0.0 over the next year, so it would be good to focus the discussion there before that. Why not label it something like `discuss-for-stable` and not release 1.0.0 until it's all closed?

adeak:
To be clear I don't object to the change _per se_. Just want to make sure that we (you) have weighed the trade-offs involved with the version bump. We've told ourselves often enough that breaking things is fine as we're <1.0 :) If the bottom line is that bumping the version is better then it's all good with me.

adeak:
Oh, and one last thought: I don't think the PyPI dev status and the version go strictly hand in hand. Pandas 0.24 was "Production/Stable" mid 2019. I think bumping the status is fine, my remarks are aimed at version 1.0 specifically.

adeak (approved):

akaszynski:
> As for the V1.0 discussions, let's save that for another day as we're not quite there yet -- still LOTS to do before I'd feel comfortable with a v1.0 release

Let's then keep the v1.0 discussion open in issues and merge this. Awaiting approval from @tkoyama010.

tkoyama010 (approved):

PR #4241, opened 2023-04-06
MAINT: Use setup-headless for pyvistaqt

larsoner:
Might as well use `setup-headless-display-action` for pyvistaqt integration. Also remove `--depth=1` which might (?) be necessary for https://github.com/pyvista/pyvistaqt/pull/356


banesullivan (approved):
Nice! Thanks for the follow up here

PR #4248, opened 2023-04-10
Fix tubes example

akaszynski:
Fix the `Plot CFD Data` example by incorporating the following suggestions from @MatthewFlamm:

- Use the inlet with `streamlines_from_source`.
- Return the entire `MultiBlock`.
- `sample` rather than `interpolate`

### DIFF
--- a/examples/99-advanced/openfoam-tubes.py
+++ b/examples/99-advanced/openfoam-tubes.py
@@ -18,8 +18,8 @@ from pyvista import examples
 ###############################################################################
 # Download and load the example dataset.
 
-mesh = examples.download_openfoam_tubes()
-mesh
+block = examples.download_openfoam_tubes()
+block
 
 
 ###############################################################################
@@ -27,12 +27,15 @@ mesh
 # ~~~~~~~~~~~~~~~~~~
 # Plot the outline of the dataset along with a cross section of the flow velocity.
 
+# first, get the first block representing the air within the tube.
+air = block[0]
+
 # generate a slice in the XZ plane
-y_slice = mesh.slice('y')
+y_slice = air.slice('y')
 
 pl = pv.Plotter()
 pl.add_mesh(y_slice, scalars='U', lighting=False, scalar_bar_args={'title': 'Flow Velocity'})
-pl.add_mesh(mesh, color='w', opacity=0.25)
+pl.add_mesh(air, color='w', opacity=0.25)
 pl.enable_anti_aliasing()
 pl.show()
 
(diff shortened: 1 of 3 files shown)


examples/99-advanced/openfoam-tubes.py
```diff
@@ -18,38 +18,54 @@
...
+pl.add_mesh(inlet, color='b', label='inlet')
+pl.add_mesh(air, opacity=0.2, color='w', label='air')
+pl.enable_anti_aliasing()
+pl.add_legend(face=None)
+pl.show()
+
+
+###############################################################################
+# Now, actually generate the streamlines. Since the original inlet contains
+# 1000 points, let's reduce this to around 200 points. First, let's remove the
```
MatthewFlamm, line 61: I'm a bit confused by

> First, let's remove the points on the outer ring

It looks like we are only using every fifth point here.

Also, a weak suggestion: this is a good use case for `decimate` filter.  This could be a commented suggestion.

MatthewFlamm (commented):
A few stylistic comments, but otherwise LGTM

examples/99-advanced/openfoam-tubes.py
akaszynski: You're right. I originally removed the points on the outer ring and then decided against it as it complicated the example without adding value.

I also tried the `decimate` filter, but didn't want to mention that in this case, it improperly samples the input and does not generate a uniform output.

This would be a great use case for [pyvista/pyacvd](https://github.com/pyvista/pyacvd).

MatthewFlamm (approved):

examples/99-advanced/openfoam-tubes.py
MatthewFlamm: Thanks for the note on pyacvd, it looks really useful for this use case.  I will try it in my workflow next time.

MatthewFlamm: We should implement `vtkPolyDataPointSampler` to generate random points on a PolyData to enable a pyvista only option in the future.

MatthewFlamm: After some testing, `vtkPolyDataPointSampler` is clunky and doesn't do what I think it should do.  I will look into it more.

akaszynski: Played with it as well.  It's helpful for generating points on the surface of polydata. I don't know if that's useful or not. Maybe?

MatthewFlamm: For some reason, I could not get it to output a number of random points less than the number of mesh points.  This has been reported in the vtk forums previously by others

https://discourse.vtk.org/t/selecting-random-points-of-a-polydata-stratified-sampling/7003/3

MatthewFlamm: `vtkMaskPoints` with spatially uniform surface or volume mode look promising.

PR #4258, opened 2023-04-11
Add simple aero bracket

akaszynski:
Add the finite element solution of an aero bracket.

Data generated from public SimScale examples at [SimScale Project Library - Turbo](https://www.simscale.com/projects/ayarnoz/turbo/).

Licensing for this dataset is granted to freely and without restriction reproduce, distribute, publish according to the [SimScale Terms and Conditions](https://www.simscale.com/terms-and-conditions).

```py
>>> dataset = examples.download_aero_bracket()
>>> cpos = [
...     (-0.0503, 0.132, -0.179),
...     (0.0505, 0.0185, -0.00201),
...     (0.275, 0.872, 0.405),
... ]
>>> dataset.plot(
...     smooth_shading=True,
...     split_sharp_edges=True,
...     scalars='von Mises stress',
...     cmap='bwr',
...     cpos=cpos,
... )

```

image

### DIFF
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -5264,7 +5264,7 @@ def download_aero_bracket(load=True):  # pragma: no cover
 
     Examples
     --------
-    Download and plot equivalent cell stress.
+    Download the aero bracket.
 
     >>> from pyvista import examples
     >>> dataset = examples.download_aero_bracket()
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -5234,6 +5234,83 @@ def download_cad_model_case(load=True):  # pragma: no cover
     return _download_and_read('cad/4947746/Vented_Rear_Case_With_Pi_Supports.vtp', load=load)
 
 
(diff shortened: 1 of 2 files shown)


akaszynski:
> @akaszynski Thank you for adding the simscale example. This is very useful to validate the FEA software at my hand. simscale examples may be added in the future, so I would like to suggest that you put together an example, [as I did with the planet](https://dev.pyvista.org/api/examples/_autosummary/pyvista.examples.planets.html). How about creating a module using the name `simscale`?

I think the planets module is a great way of organizing the examples! However, I'm opposed to including module that is directly linked to a specific application or software. For example, we can include Ansys, SimScale, or OnScale examples, but we shouldn't tie ourselves to a specific CAE entity.

What we can do instead is add a FEA or FVM module to start collecting all these examples together. We're indeed starting to accumulate quite a few and it will be helpful to have some organization to this moving forward.

tkoyama010 (approved):
@akaszynski Thanks for your thoughts. I'd organize, but that's another PR.

PR #4287, opened 2023-04-17
Fix specular values in examples

tkoyama010:
### Overview

This is an omission of correction in the #4226 changes.

The error occurred when running locally. I think this error should also occur when the document is built, but I would like to know why it did not occur.

### Details

-[ The build of the Japanese page stopped for two weeks because of this error](https://github.com/pyvista/pyvista-docs-dev-ja/actions), so I think it is a CI issue.


banesullivan (approved):

akaszynski:
> The error occurred when running locally. I think this error should also occur when the document is built, but I would like to know why it did not occur.

These errors were not caught during our documentation builds because we cache the examples rather than rebuild them for each documentation build.

We catch these during a release since the documentation cache is reset when the version changes.

tkoyama010:
Ah! I see. Thanks.

PR #4295, opened 2023-04-17
Show scalar bar if scalar_bar_args are passed

banesullivan:
I've been wanting tight control of scalar bars lately and found that if you set `show_scalar_bar=False` on the theme and then pass `scalar_bar_args` to `add_mesh()`, the scalar bar is still hidden. IMO, if the user is explicitly passing `scalar_bar_args`, then they probably want to show the scalar bar.

This PR makes this scenario a bit easier.

For example:

```py
import pyvista as pv

pv.global_theme.show_scalar_bar = False

pl = pv.Plotter()
pl.add_mesh(pv.Wavelet(), scalar_bar_args=dict(title='foo'))  # I expect scalar bar to show
pl.show()
```

### DIFF
--- a/pyvista/plotting/_plotting.py
+++ b/pyvista/plotting/_plotting.py
@@ -224,6 +224,10 @@ def _common_arg_parser(
     elif culling in ['f', 'frontface']:
         culling = 'front'
 
+    if show_scalar_bar is None:
+        # use theme unless plotting RGB
+        _default = theme.show_scalar_bar or scalar_bar_args is not None
+        show_scalar_bar = False if rgb else _default
     # Avoid mutating input
     if scalar_bar_args is None:
         scalar_bar_args = {'n_colors': n_colors}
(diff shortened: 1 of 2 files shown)


akaszynski (dismissed):
Agree with the behavior.

pyvista/plotting/_plotting.py
```diff
@@ -224,6 +224,10 @@ def _common_arg_parser(
     elif culling in ['f', 'frontface']:
         culling = 'front'
 
+    if show_scalar_bar is None:
+        # use theme unless plotting RGB
+        _default = theme.show_scalar_bar or scalar_bar_args is not None
```
MatthewFlamm, line 229: This PR raises some edge cases.  What do we have to happen if the user passes `scalar_bar_args={}`?  I would argue this might mean no scalar bar.  Should we check truthiness instead?

MatthewFlamm (approved):
Not sure that a test is worth it here, LGTM in the current state unless you think a test is warranted.

banesullivan:
> Not sure that a test is worth it here, LGTM in the current state unless you think a test is warranted.

I didn't want to mess with the global theme in a test... I think we have mechanisms for this, but did not seem worth it to me

PR #4299, opened 2023-04-18
Remove flask examples

banesullivan:
These flask examples were rough proof of concept when they were created. They do not scale and are light years behind what's possible with Trame. Considering the adoption and growth we have seen with PyVista+Trame, I think we should outright remove the flask examples as they only mislead and confuse users (as pointed out in https://github.com/pyvista/pyvista/discussions/4297#discussioncomment-5651543)


tkoyama010 (approved):

akaszynski:
Absolutely agree with this. While I think it might be useful in some edge case to save an image offline, I think that sort of workflow is quite straightforward and we don't need to include it out of the concern of confusing the user.

tkoyama010:
> While I think it might be useful in some edge case to save an image offline,

I have learned a bit about Trame and this may be feasible with Trame.

banesullivan:
It is possible with trame 😉  I just need to expose it in the PyVista view

PR #4320, opened 2023-04-23
Fix the edit this page link

akaszynski:
Inspired by https://github.com/pyansys/pymapdl/pull/2012, this PR fixes the "Edit this page" button, which is currently overall because of the move from "doc/" to "doc/source".

Moreover, it also provides correct links for our:
- Gallery examples
- Autosummary examples

These were broken even before the documentation move.

---

@germa89 and @jorgepiloto, I suggest that we implement it this way, it's guarenteed to get to the right source file location rather than implementing a search in GitHub.

### DIFF
--- a/doc/source/_templates/edit-this-page.html
+++ b/doc/source/_templates/edit-this-page.html
@@ -1,0 +1,16 @@
+{% if sourcename is defined and theme_use_edit_page_button and page_source_suffix %}
+  {% set src = sourcename.split('.') %}
+  <div class="tocsection editthispage">
+    <a href="{{ fix_edit_link_button(get_edit_provider_and_url()[1]) }}">
+      <i class="fa-solid fa-pencil"></i>
+      {% set provider = get_edit_provider_and_url()[0] %}
+      {% block edit_this_page_text %}
+        {% if provider %}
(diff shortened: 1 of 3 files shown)


germa89:
To avoid having the search option there are two possibilities:
* Using GitHub API to get where the function is defined. Actually it cannot be done i believe. Github can only provides files, lines numbers and commits. 
* Using python libraries to retrieve where the object is defined. Problem of this, you need to instantiate the object in order to then use something like ``func.__module__`` and ``func.__file__``. That will probably made the whole documentation building very slow. Further more, the object definition you get when building the object (for instance ``ansys.mapdl.core.Mapdl.exit`` might not always match the real object. For instance ``ansys.mapdl.core.launch_mapdl`` defined in ``ansys.mapdl.core.launcher.launch_mapdl``. 

The whole thing needs careful thinking. It should be possible. But it needs more than what I did in https://github.com/pyansys/pymapdl/pull/2012

Temporary alternative, we could maybe link the search to provide directly the first result though. Github search API should provide information enough to directly link the first result of the search.

banesullivan (approved):
Generally, LGTM!

Haven't given this a thorough review but all looks right to me and agree with conclusions of previous discussions

PR #4321, opened 2023-04-24
Disable anti-aliasing on openfoam cooling example

akaszynski:
Disable anti-aliasing on openfoam cooling example to see if this fixes the missing volumetric actor.


MatthewFlamm (approved):
LGTM

PR #4330, opened 2023-04-24
Add icosphere

### LINKED ISSUE #4319: Icosphere Primitive

I am new to pyvista and I am just exploring the different features.

It would be great if pyvista would support a quick way to create an icosphere (an icohedron projected into a sphere). See this image for an example.  The image is taken from https://en.wikipedia.org/wiki/Geodesic_polyhedron

image

akaszynski:
Resolve #4319 by adding ``pyvista.Icosphere``.

```py
import pyvista as pv
icosphere = pv.Icosphere()
icosphere.plot(show_edges=True)
```

image

Thanks @darikg for your code!

### DIFF
--- a/pyvista/utilities/geometric_objects.py
+++ b/pyvista/utilities/geometric_objects.py
@@ -347,6 +347,10 @@ def Sphere(
     pyvista.PolyData
         Sphere mesh.
 
+    See Also
+    --------
+    pyvista.Icosphere
+
     Examples
     --------
     Create a sphere using default parameters.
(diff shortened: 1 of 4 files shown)


pyvista/utilities/geometric_objects.py
```diff
@@ -1730,3 +1734,76 @@ def Icosahedron(radius=1.0, center=(0.0, 0.0, 0.0)):
...
+    >>> icosphere.compute_cell_sizes().plot(scalars='Area')
+
+    """
+    mesh = Icosahedron()
+    mesh.clear_data()
+    mesh = mesh.subdivide(nsub=nsub)
+
+    # scale to desired radius and translate origin
+    dist = np.linalg.norm(mesh.points, axis=1, keepdims=True)  # distance from origin
+    mesh.points[:] = mesh.points * (radius / dist) + center
```
tkoyama010, line 1808: Just out of interest, what is the intention of adding `[:]` here?

tkoyama010 (approved):

pyvista/utilities/geometric_objects.py
akaszynski: I thought it would be faster to update the points in-place, but turns out it's actually way slower!

```
>>> import pyvista as pv
>>> mesh = pv.Sphere(theta_resolution=1000, phi_resolution=1000)
>>> new_points = mesh.points.copy()
>>> timeit mesh.points[:] = new_points
450 µs ± 11.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

timeit mesh.points = new_points
20.2 µs ± 49.5 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Thanks for pointing that out @tkoyama010, I'll change that back.

akaszynski:
```suggestion
    mesh.points = mesh.points * (radius / dist) + center
```

tkoyama010 (approved):

PR #4331, opened 2023-04-24
Deprecate offset in UnstructuredGrid

akaszynski:
Since we now require VTK >= 9.0, we no longer need backwards compatibility when initializing using the `offset` array.

This PR removes the now unnecessary logic to support the `offset` array.


banesullivan (approved):
Finally! The offsets confused me for years and I'm glad we're finally modernizing this API to match the upstream changes

PR #4337, opened 2023-04-25
Lazy reader imports

akaszynski:
We did an OSS scan within [pyansys/pyfluent](https://github.com/pyansys/pyfluent) and discovered a few vulnerabilities. One of those was a TIFF reader from VTK, and it's entirely possible to simply remove the offending library from ``vtk``.

The only issue with this is that we import the library on import, which inspired me to consider if we can just lazy import most, if not all, of our libraries since we now require VTK v9 or greater.

~For now, this PR just lazy imports `vtkmodules.vtkIOImage.vtkTIFFReader`, but is there any reason why we can't just do it for all the modules?~ This has several advantages:

- We only import what we use. If you are only post-procesing and don't actually plot, you shouldn't have to import libGLU. This was brought up in https://github.com/pyvista/pyvista/discussions/3283 and this has even led to people not using PyVista because we import the kitchen sink
- Improve import speed.
- Reduced risk to potential upstream vulnerabilities.

As such, I'd suggest that we rethink our imports and actually get away from using `_vtk` and simply import directly from `vtkmodules`. This was only really implemented to support backwards compatibility between vtk v8 and v9 and we can now get rid of this technical debt.


pyvista/utilities/reader.py
```diff
@@ -2177,7 +2177,13 @@ class TIFFReader(BaseReader):
 
     """
 
-    _class_reader = _vtk.vtkTIFFReader
+    def lazy_vtkTIFFReader(*args):
+        """Lazy import of the vtkTIFFReader."""
+        from vtkmodules.vtkIOImage import vtkTIFFReader
+
+        return vtkTIFFReader()
```
MatthewFlamm, line 2184: I don't think you can instantiate here.

MatthewFlamm (commented):
I'm +1 on overall intent and am fine for this limited implementation.  Possibly future looking comment:  if we are going forward with lazy importing everything, or importing at point of use, then we should think about helper functions that avoid having this boilerplate for each reader.

pyvista/utilities/reader.py
MatthewFlamm: No I think I'm wrong, this isnt called on class definition

pyvista/utilities/reader.py
```diff
@@ -2177,7 +2177,13 @@ class TIFFReader(BaseReader):
 
     """
 
-    _class_reader = _vtk.vtkTIFFReader
+    def lazy_vtkTIFFReader(*args):
```
MatthewFlamm, line 2180: We have used `staticmethod` in other lazy loaded readers to avoid the need for `*args` here.  I think it is cleaner this way.

akaszynski: Check if I'm on the right path with deac1c5.

MatthewFlamm: This is looking better in the reader module, this is a clear cut case where we should not import up front.  I'll look at the more holistic changes later, we need to be careful about not letting lazy importers leak into the pyvista namespace.

pyvista/plotting/plotting.py
```diff
@@ -20,6 +20,7 @@
 import matplotlib
 import numpy as np
 import scooby
+from vtkmodules.vtkRenderingOpenGL2 import vtkOpenGLTexture
```
MatthewFlamm, line 23: If one of the goals of this PR is to allow users to avoid importing from any `vtkmodules.vtkRenderingOpenGL2`, I think these additions will still cause this import during `import pyvista`.

akaszynski: One of the goals of this PR is to start to consider moving imports out of `_vtk` and to import when needed within methods.

While going through this PR, I realize that many of the imports are needed when creating classes, which makes it difficult to lazy import classes like `vtkCharts`, which isn't widely used.

I think I'm going to revert the last two and focus on just reader imports at the moment.

akaszynski: Reverted in [2a3bed3](https://github.com/pyvista/pyvista/pull/4337/commits/2a3bed339553482d883289df885a7e4be8b59bfd).

MatthewFlamm: I'm not against moving the imports like this, but it will require some discussion on best path forward if/whether we want to avoid import unless used.  I'm +1 on your suggested strategy for this PR.

pyvista/utilities/reader.py
```diff
@@ -1115,7 +1143,9 @@ class POpenFOAMReader(OpenFOAMReader):
     parallel reconstructed data, and decomposed data.
     """
 
-    _class_reader = staticmethod(_vtk.lazy_vtkPOpenFOAMReader)
+    _class_reader = staticmethod(lazy_vtkPOpenFOAMReader)
+    _vtk_module_name = ''
+    _vtk_class_name = ''
```
MatthewFlamm, line 1148: Do we need these lines?  They seem to reiterate the default setting.

```suggestion
```

MatthewFlamm (commented):
One small comment, otherwise LGTM

pyvista/utilities/reader.py
akaszynski: We do because we're overriding from the class we're inheriting from `OpenFOAMReader`.

MatthewFlamm (approved):

PR #4357, opened 2023-04-29
Add merge points

akaszynski:
Address #4345 by adding `merge_points`.

```py
>>> import pyvista as pv
>>> cyl = pv.CylinderStructured(radius=[0, 0.1, 0.2])
>>> cyl.n_points
990
>>> cyl_merged = cyl.merge_points()
>>> cyl_merged.n_points
670
```

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -4722,6 +4722,43 @@ class DataSetFilters:
                 raise TypeError(f"Mesh type {type(self)} cannot be overridden by output.")
         return merged
 
+    def merge_points(self, tolerance=0.0, progress_bar=False):
+        """Merge points of this dataset.
+
+        Parameters
+        ----------
+        tolerance : float, default: 0.0
+            The absolute tolerance to use to find coincident points.
+
+        progress_bar : bool, default: False
+            Display a progress bar to indicate progress.
+
+        Returns
+        -------
+        UnstructuredGrid
+            Dataset with merged points.
+
+        See Also
+        --------
+        PolyDataFilters.clean : Preferable with :class:`pyvista.PolyData`.
+
+        Examples
+        --------
+        Create a structured cylinder and merge interior faces. This prevents
+        additional "internal" faces when extracting the surface of the
+        cylinder.
+
+        >>> import pyvista as pv
+        >>> cyl = pv.CylinderStructured(radius=[0, 0.1, 0.2])
+        >>> cyl.n_points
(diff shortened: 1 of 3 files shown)


pyvista/core/filters/data_set.py
```diff
@@ -4722,6 +4722,43 @@ def merge(
...
+        >>> import pyvista as pv
+        >>> cyl = pv.CylinderStructured(radius=[0, 0.1, 0.2])
+        >>> cyl.n_points
+        990
+        >>> cyl_merged = cyl.merge_points()
+        >>> cyl_merged.n_points
+        670
+
+        """
+        return self.merge(tolerance=tolerance, progress_bar=progress_bar)
```
tkoyama010, line 4760: I agree that this name is better, but what about deprecating the `merge` method and replacing it with `merge_points`?
In DataSet, I think only `merge_points` is sufficient.

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
akaszynski: `merge` is still necessary in order to merge multiple datasets. `merge_points` is just for merging points of the dataset itself.

tkoyama010 (approved):
LGTM

pyvista/core/filters/data_set.py
tkoyama010: Got it. Thanks!

banesullivan (changes requested):
I think this should be called `clean()`. It is effectively the  `vtkCleanPolyData` filter but for non-poly data. We have `clean()` as a method specific to `PolyData`, and I think it would be better to have this method be the default `clean()` method for all other data types consider this is missing

akaszynski:
I just learned that there's [vtkStaticCleanUnstructuredGrid](https://vtk.org/doc/nightly/html/classvtkStaticCleanUnstructuredGrid.html), which might be what we want to use instead as a `clean` method as proposed by @banesullivan. However, I'm wary of implementing this as a `clean` method at the `DataSet` level since this will violate the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

banesullivan:
Fair enough!

This meme explains [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle): 
ilxzO

akaszynski:
Dropping in favor of #4367.

PR #4361, opened 2023-04-30
Add iterative closest point

akaszynski:
Implement ``vtkIterativeClosestPointTransform``. Inspired by #4310.

image


tests/test_filters.py
```diff
@@ -2647,6 +2647,22 @@ def test_integrate_data():
     assert np.isclose(integrated["pdata"], 3 * np.pi, rtol=1e-3)
 
 
+def test_iterative_closest_point():
+    # Create a simple mesh
+    source = pyvista.Cylinder(resolution=30).triangulate().subdivide(1)
+    transformed = source.rotate_y(20).rotate_z(25).translate([-0.75, -0.5, 0.5])
+
+    # Perform ICP registration
+    aligned = transformed.iterative_closest_point(source)
```
akaszynski, line 2656:
```suggestion
    aligned = transformed.iterative_closest_point(source)

    _, matrix = transformed.iterative_closest_point(source, return_matrix=True)
    assert isinstance(matrix, np.ndarray)
```

pyvista/core/filters/data_set.py
```diff
@@ -68,6 +68,108 @@ def _clip_with_function(
...
+        return_matrix : bool, default: False
+            Return the transform matrix as well as the aligned mesh.
+
+        Returns
+        -------
+        aligned : pyvista.DataSet
+            The dataset aligned to the target mesh.
+
+        matrix : numpy.ndarray
+            Transform matrix to transform the input dataset to the target dataset.
```
adeak, line 115: The functionality seems great, but the name is super confusing. We have filters like `find_closest_point()`, which finds the closest point (:exploding_head:). This is `iterative_closest_point()` which... returns a mesh and a matrix? Can we call it something more descriptive? `iterative_closest_point_transform()` would be a lot better to me, but of course that's very verbose.

adeak: Maybe cater to clueless users such as myself and call it something like `align_transform()`, and mention the ICP algorithm in the description so that experts looking for it can find the appropriate method?

banesullivan: I'm on the same page here. PR title made me think of something totally different. I think `transform_by_closest_point()` or similar would be a good name (with `transform` first to put it next to other `transform*` methods)

akaszynski: Guess it's my familarity with ICP (iterative closest point).

I'd prefer something like "align", but second place would be `transform_by_alignment`, which we could consider exposing different alignment algorithms.

banesullivan: `align()` would also be a good choice for this and is in line with other methods that require a second dataset, e.g., `sample()`. Leaving this mostly to your discretion, @akaszynski

banesullivan: Though considering `Transform` is in the name of the VTK algorithm, I'm leaning towards `transform_*`... perhaps even `transform_align()`?

akaszynski: Align seems the most intutive to me. I'll wait for others to weigh in here before proceeding.

akaszynski: Alright, going with `align` here.

PR #4365, opened 2023-05-01
Improve pyvista.StructuredGrid example

natsuwater:
DocString of class StructuredGrid is updated to show that `indexing='ij'` option should be applied to prepare `x, y, z = np.meshgrid(xrng, yrng, zrng, indexing='ij')` .

Details on this document modification can be found at https://github.com/pyvista/pyvista/discussions/4360

### Overview

Adding option `indexing='ij'` in np.meshgrid call makes the resulting SturucturedGrid(s, y, z) scans x-y-z order as RectilinearGrid does.
By default, np.meshgrid uses indexing='xy', and this result in y-x-z ordering in StructuredGrid(x, y, z).

### DIFF
--- a/pyvista/utilities/features.py
+++ b/pyvista/utilities/features.py
@@ -69,6 +69,8 @@ def voxelize(mesh, density=None, check_surface=True):
     y = np.arange(y_min, y_max, density_y)
     z = np.arange(z_min, z_max, density_z)
     x, y, z = np.meshgrid(x, y, z, indexing='ij')
+    # indexing='ij' is used here in order to make grid and ugrid with x-y-z ordering, not y-x-z ordering
+    # see https://github.com/pyvista/pyvista/pull/4365
 
     # Create unstructured grid from the structured grid
     grid = pyvista.StructuredGrid(x, y, z)
--- a/pyvista/utilities/features.py
+++ b/pyvista/utilities/features.py
@@ -10,9 +10,6 @@ import pyvista
 def voxelize(mesh, density=None, check_surface=True):
     """Voxelize mesh to UnstructuredGrid.
 
-    Prior to version 0.39.0, this method improperly handled the order of
-    structured coordinates.
-
     Parameters
     ----------
     density : float | array_like[float]
@@ -30,6 +27,11 @@ def voxelize(mesh, density=None, check_surface=True):
     -------
     pyvista.UnstructuredGrid
         Voxelized unstructured grid of the original mesh.
+
+    Notes
+    -----
+    Prior to version 0.39.0, this method improperly handled the order of
+    structured coordinates.
 
     Examples
     --------
--- a/pyvista/utilities/features.py
+++ b/pyvista/utilities/features.py
@@ -68,7 +68,7 @@ def voxelize(mesh, density=None, check_surface=True):
     x = np.arange(x_min, x_max, density_x)
     y = np.arange(y_min, y_max, density_y)
     z = np.arange(z_min, z_max, density_z)
-    x, y, z = np.meshgrid(x, y, z)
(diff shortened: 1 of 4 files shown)


tkoyama010 (dismissed):
LGTM. Thanks for your contribution!

banesullivan (changes requested):
Thank you for this fix and clarification, @natsuwater! Would you please take some extra time to fix this everywhere you can? `np.meshgrid` is used in many examples and docs improperly with `StructuredGrid`, and we may need to fix this in more places.

banesullivan:
Most of the examples have uniform X and Y coordinates, so it hasn't been much of an issue -- we should be more explicit with this though.

natsuwater:
@banesullivan OK, I'll take time to look into them.

natsuwater (commented):
Added `indexing='ij'` where three arguments are passed to `np.meshgrid`. Most of them are docstring and .rst, while `features.py` is an active code, and this change might not be necessary.

Note that I did not touched anything with 'test_' prefix though I find some test includes calls to `np.meshgrid`.

pyvista/utilities/features.py
```diff
@@ -68,7 +68,7 @@ def voxelize(mesh, density=None, check_surface=True):
     x = np.arange(x_min, x_max, density_x)
     y = np.arange(y_min, y_max, density_y)
     z = np.arange(z_min, z_max, density_z)
-    x, y, z = np.meshgrid(x, y, z)
+    x, y, z = np.meshgrid(x, y, z, indexing='ij')
```
tkoyama010, line 71: Thank you for the change. I think we need an explanation (and also add a test if you can) of how the behavior has changed because of the change in the method.

tkoyama010 (changes requested):

pyvista/utilities/features.py
natsuwater: This change is made to make the ordering of the three dimensional ordering of the cell data and point data in the `StructuredGrid` as discussed in https://github.com/pyvista/pyvista/discussions/4360.

Is it suffice to explain here, or do I need to put explanation else where?

And I'm afraid adding test is beyond my capability, for now,

tkoyama010: The following may be helpful in adding explanations to the Docstring. Add it and then ask other maintainers to review it.
I will add the tests later on my end, could you add an explanation about your changes in docstring?
https://github.com/pyvista/pyvista/blob/5c40f80e1f64adbd215ce1589b534956191e19d7/pyvista/core/filters/poly_data.py#L2792-L2798

banesullivan: Thanks for catching this, @natsuwater! This change is warranted, and the previous implementation must be corrected. I suspect this went unnoticed for as long as it did because this helper function isn't too widely used.

I wouldn't worry about testing this change -- we can handle that if it's really warranted

+1 for @tkoyama010's suggestion to add a note in the docstring. Something along the lines of "prior to version 0.39.0, this method improperly handled the structured coordinates"

banesullivan:
> Added indexing='ij' where three arguments are passed to np.meshgrid.

Perfect! Thank you for taking the extra time to do this!

> Note that I did not touched anything with 'test_' prefix though I find some test includes calls to np.meshgrid.

That's fine -- we/I should look over these for a sanity check but if the tests are passing, I wouldn't worry too much about these

akaszynski:
@banesullivan and @tkoyama010, I think this is good to go.

akaszynski (approved):

banesullivan (approved):
Looks great! Thanks, @natsuwater!!

tkoyama010 (approved):
LGTM

PR #4368, opened 2023-05-01
Use fullname for autosummary

akaszynski:
Eliminate duplicate pages within our documentation search by using the ``fullname`` of a class method in the autosummary. For example:

https://docs.pyvista.org/version/dev/search.html?q=clean

## `main` branch
a

## This branch
b

(note that displaying the text requires the documentation to be hosted)


tkoyama010 (approved):
Good! Makes it more visible.

PR #4374, opened 2023-05-02
Add a simple example to project points to a plane

akaszynski:
While working with LlamaIndex, I've discovered that it is incredibly important to have a slew of examples to enable the AI to be able to generate additional examples based on active context.

This PR creates a new directory within out `examples/` directory called `simple`, where we can place simple examples that we might consider pulling as standalone examples from our [discussions](https://github.com/pyvista/pyvista/discussions) or [issues](https://github.com/pyvista/pyvista/issues).

I'm starting this with a simple example from a request raised over email.

tmp

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,6 +23,7 @@ concurrency:
 env:
   USE_CACHE: ${{ (github.event_name == 'workflow_dispatch' && github.event.inputs.cache == 'true') || (github.event_name == 'pull_request') || (github.event_name == 'push') }}
   DOCUMENTATION_CNAME: 'docs.pyvista.org'
+  PYDEVD_DISABLE_FILE_VALIDATION: '1'
 
 jobs:
   doc-check:
--- a/examples/98-common/README.rst
+++ b/examples/98-common/README.rst
@@ -1,0 +1,11 @@
+Common
+------
+
+The following are a collection of examples addressing common tasks and
+frequently asked questions when working with PyVista. These examples, while
+possibly complex, focus on single-purpose solutions to help users understand
+specific functionality.
+
+Feel free to contribute examples or explore the FAQ-like structure, and check
+out our `GitHub Discussions <https://github.com/pyvista/pyvista/discussions>`_
+for more context.
--- a/examples/98-common/project-points-tessellate.py
+++ b/examples/98-common/project-points-tessellate.py
@@ -1,5 +1,6 @@
 """
-Project points to a plane and Tessellate.
+Project points to a plane and Tessellate
+----------------------------------------
 
 Using pyvista and numpy, generate a 3D point cloud, project it to a plane, and
 tessellate it.
--- a/examples/98-simple/README.rst
+++ b/examples/98-simple/README.rst
@@ -1,0 +1,4 @@
+Simple
+------
+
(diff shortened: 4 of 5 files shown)


MatthewFlamm:
One thought before I go in detail to the code. Is what you are sayin that our examples today are "complex" in that they explore several possibilities with explanatory information being printed and plotted.  But, there is also value in having single-purpose built examples that might have explanation around them but the code is single-purpose?  I would say this should have a different name than "simple" as some examples might still be complex, just not in the same way as the examples.  These would be most useful to users if they have a function definition near the top that does the operation (like you have here).  If this is the intent, we could add in some language about welcoming PRs from anyone who has created a purpose built function for performing operations.  This would also lesson the burden to add in convenience features to the core code.

What about "cookbook" similar to the [scipy cookbook](https://github.com/scipy/scipy-cookbook/)? Although I would say the scipy cookbook is typically more similar to our typical examples, so it isn't a great 1:1 comparison.

akaszynski:
You bring up valid points about the naming convention and the intent behind these new examples. I understand your concern about the "simple" label, as it may not accurately represent the nature of these examples. They may still be complex but focus on a single-purpose, as you've mentioned.

@banesullivan and I have mulled over this idea for a while. We have a ton of examples and really need to add more, but our existing repository and documentation build is getting quite long. I'm wary of making it even longer with a variety of examples, so I really like the idea of a `pyvista-cookbook`, but I'm also wary of splitting up the documentation.

In some ways, our existing examples are very much so aligned with the "cookbook" approach, so I'm not sure what to call these examples. I'd like to continue to use the "gallery" approach of showcasing examples, and perhaps this is really a question of scale if we want to start adding a ton of examples.

@banesullivan, maybe you could weigh in here.

akaszynski:
Here's another example I'd like to add. These are less "cookbook" examples with a single purpose, but more solutions of problems. Really, these are questions and answers, which are the best possible input/output (prompt/completion) available for AI learning, and potentially humans as well.

---

Question: There are internal faces within the output from `CylinderStructured`, how can I eliminate them? They're causing my smoothing operation to fail.

Answer:
Use PyVista's `UnstructuredGrid.clean()` method. Note that you have to first cast to an unstructured grid first with `cast_to_unstructured_grid` as this method is unavailable for `pyvista.StructuredGrid`.

```py
>>> import pyvista as pv
>>> cyl = pv.CylinderStructured(radius=[0, 0.1, 0.2])
>>> cyl.n_points
990
>>> cyl_merged = cyl.cast_to_unstructured_grid().clean()
>>> cyl_merged.n_points
670
```

---

The advantage of this approach and having some sort of CI/CD around this is these examples work. In machine learning, it's typically garbage in --> garbage out. We need to have an active set of maintained examples and the problem with just leaving this knowledge on GitHub is it isn't maintainable. Rather, a database of examples, cookbook, or curated and up-to-date Q&A forum.

MatthewFlamm:
So this is more like a curated StackExchange questions and answers?  Naming and description here will be important to keep it focused to the intent.  Simple implies just that, while this second example is short, but quite subtle. You need to understand quite a lot about the underlying data model and what `CylinderStructured` does etc.

akaszynski:
I renamed the section to "Common" per Bane's suggestion with a description that matches the intent.

banesullivan (approved):
I think this is an excellent section to add to the gallery

PR #4389, opened 2023-05-05
v0.40 Deprecations

banesullivan:
Deprecations marked for removal on v0.40

I recommend waiting to merge these until the dust settles as this PR could muck up any dire patch releases.

### DIFF
--- a/pyvista/core/grid.py
+++ b/pyvista/core/grid.py
@@ -2,7 +2,6 @@
 from functools import wraps
 import pathlib
 from typing import Sequence, Tuple, Union
-import warnings
 
 import numpy as np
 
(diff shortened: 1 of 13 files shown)


pyvista/core/grid.py
```diff
@@ -485,70 +484,29 @@ class UniformGrid(_vtk.vtkImageData, Grid, UniformGridFilters):
 
     def __init__(
         self,
-        uinput=None,
+        image=None,
```
banesullivan, line 487: Maybe we shouldn't change the name of this argument?

adeak: On a related note, in half a year when we drop 3.7 support we should consider making all these parameters [positional-only](https://peps.python.org/pep-0570/):
```py
>>> def foo(self, bar=None, /, *args, baz=None):
...     print(bar)
... 
>>> foo(3)
None
>>> foo(3, bar=42)
TypeError: foo() got some positional-only arguments passed as keyword arguments: 'bar'
```
This has the same risk as renaming it today, hence related note.

adeak: I guess we could do the same breaking today by removing the argument and doing `image, *args = args` in the method body, but then we'd have to undo that when we can use positional-only parameters for more readability.

banesullivan: We already dropped 3.7 (ref https://github.com/pyvista/pyvista/pull/3938) and 

https://github.com/pyvista/pyvista/blob/9ee35ba478e573711f54b3fdac5fb04242208c22/pyproject.toml#L12

adeak: Oh, nice, I forgot about that. In that case definitely consider making it positional-only if we rename it.

banesullivan:
Too many conflicts. Recommend new PR

PR #4396, opened 2023-05-09
Automatically label maintenance to .pre-commit-config.yaml update

tkoyama010:
### Overview

Automatically label maintenance to .pre-commit-config.yaml update.

relate to #4395 .

### Details

- None


tkoyama010:
I will merge this because the change does not affect the source.

PR #4399, opened 2023-05-09
Add find_cells_intersecting_line

### LINKED ISSUE #4398: `find_cells_along_line` not working according to description
### Discussed in https://github.com/pyvista/pyvista/discussions/4390

<div type='discussions-op-text'>

<sup>Originally posted by **AsimAl-Sofi** May  6, 2023</sup>
Hi all
I would like to use the function **find_cells_along_line** to determine the cells along a line, but the function does not give me the right answer.
Here is my code example:
```py
out = pv.CylinderStructured(radius=(0,3),
theta_resolution=10,height=1,
z_resolution=2)
box=out.get_cell(5).get_face(3).cast_to_unstructured_grid().extract_surface()
box.compute_normals(inplace=True)
centers = out.get_cell(5).get_face(3).cast_to_unstructured_grid().cell_centers().points
arrows = box['Normals'][0]
pl = pv.Plotter()
pl.add_mesh(out, opacity=0.6, show_edges=True, line_width=1)
pl.add_points(arrows,color="g", point_size=18.0, render_points_as_spheres=True)
point_S = (0, -5, 2)
pl.add_points(np.array([point_S]),color="r", point_size=18.0, render_points_as_spheres=True)
line=pv.Line(point_S, arrows)
pl.add_mesh(line)
pl.show()
```

the function find_cells_along_line gives me the following answer:
```py
>>> out.find_cells_along_line(point_S, arrows,tolerance=0.0)
array([1, 0, 2, 7, 8, 9])
```
As you can see from the example, the line goes through cell 0, 8 and 9 but the function gives 1, 0, 2, 7, 8, 9.
Any suggestions how to tackle this problem?
Thanks</div>

MatthewFlamm:
### Overview

Closes #4398.  This PR first fixes the description of `find_cells_along_line` to clarify that it is finding cells whose bounds intersect the line and not the cells themselves.  The PR adds `find_cells_intersecting_line` to find cells that directly intersect the lines.

### Details

I updated the examples to be better.  The original example put the line in the (0, 0, 1) direction which should go through the vertex at the top of the sphere where many many cells converge together.  It is better to use the (1, 0, 0) direction where it should only intersect 1 cell.  The examples now highlight the difference between the methods where `find_cells_along_line` returns multiple cells, whereas `find_cells_intersecting_line` returns only 1 cell.

### DIFF
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -2419,7 +2419,68 @@ class DataSet(DataSetFilters, DataObject):
         pointb: Iterable[float],
         tolerance=0.0,
     ) -> np.ndarray:
-        """Find the index of cells in this mesh along a line.
+        """Find the index of cells whose bounds intersect a line.
+
+        Line is defined from ``pointa`` to ``pointb``.
+
+        Warnings
+        --------
+        This method returns cells whose bounds intersect the line.
+        This means that the line may not intersect the cell itself.
+        To obtain cells that intersect the line, use
+        :func:`pyvista.DataSet.find_cells_intersecting_line`.
+
+        Parameters
+        ----------
+        pointa : sequence[float]
+            Length 3 coordinate of the start of the line.
+
+        pointb : sequence[float]
+            Length 3 coordinate of the end of the line.
+
+        tolerance : float, default: False
+            The absolute tolerance to use to find cells along line.
+
+        Returns
+        -------
+        numpy.ndarray
+            Index or indices of the cell(s) whose bounds intersect
+            the line.
+
+        See Also
+        --------
+        DataSet.find_closest_point
+        DataSet.find_closest_cell
+        DataSet.find_containing_cell
+        DataSet.find_cells_within_bounds
+        DataSet.find_cells_intersecting_line
+
+        Examples
+        --------
+        >>> import pyvista
+        >>> mesh = pyvista.Sphere()
+        >>> mesh.find_cells_along_line([0, 0, 0], [1.0, 0])
+        array([842, 843, 896, 897])
+
+        """
+        if np.array(pointa).size != 3:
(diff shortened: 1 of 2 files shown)


MatthewFlamm:
cc @AsimAl-Sofi if you want to test this PR

pyvista/core/dataset.py
```diff
@@ -2465,6 +2475,66 @@ def find_cells_along_line(
...
+        >>> mesh = pyvista.Sphere()
+        >>> mesh.find_cells_intersecting_line([0.0, 0, 0], [1.0, 0, 0])
+        array([896])
+
+        """
+        if pyvista.vtk_version_info < (9, 2, 0):
+            raise VTKVersionError("pyvista.PointSet requires VTK >= 9.2.0")
+
+        if np.array(pointa).size != 3:
+            raise TypeError("Point A must be a length three tuple of floats.")
```
MatthewFlamm, line 2526: I followed the other method here, but I think it is probably better as `ValueError`.  If so, we should change both which would be a small breaking change.

akaszynski: Agree that `ValueError` would be correct here. It's out of scope of this PR, but let's do it in a different PR and have it merged for 0.40.0

akaszynski (approved):
Solid LGTM. Thanks @MatthewFlamm.

pyvista/core/dataset.py
adeak: Well with the current wording ("length three tuple of floats") the type error is also correct. Which is just to say that we should probably reword it while we're at it.

akaszynski: We might consider reusing:

https://github.com/pyvista/pyvista/blob/461bb5ed5be8c9dda6e9ed765fab3224149d9c62/pyvista/utilities/errors.py#L414-L421

PR #4419, opened 2023-05-12
Remove SKIP in docstrings

### LINKED ISSUE #4405: Images from docstring examples not always captured in docs

#4383 switched the order of commands to get the plot rendered in the documentation.  I see this in other places of the documentation like https://docs.pyvista.org/version/dev/api/examples/_autosummary/pyvista.examples.downloads.download_electronics_cooling.html#pyvista.examples.downloads.download_electronics_cooling

But in other places like
https://docs.pyvista.org/version/dev/api/examples/_autosummary/pyvista.examples.downloads.download_exodus.html#pyvista.examples.downloads.download_exodus

it renders fine.

### Links to source code.

_No response_

### Pseudocode or Screenshots

_No response_

akaszynski:
Resolves #4405 by removing skips wherever possible in our docstrings. This allows images to be generated, and revealed several incorrect `repr`.

Also removes `"sphinx.ext.doctest"` since we use `pytest doctest-modules`.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -69,7 +69,6 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "sphinx.ext.linkcode",  # This adds the button ``[Source]`` to each Python API site by calling ``linkcode_resolve``
-    "sphinx.ext.doctest",
     "sphinx.ext.extlinks",
     "sphinx.ext.intersphinx",
     "sphinx_copybutton",
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -49,9 +49,12 @@ class PolyDataFilters(DataSetFilters):
         >>> import pyvista
         >>> mesh = pyvista.Cube().triangulate().subdivide(4)
         >>> mask = mesh.edge_mask(45)
+        >>> mesh.plot(scalars=mask)
+
+        Show the array of masked points.
+
         >>> mask  # doctest:+SKIP
         array([ True,  True,  True, ..., False, False, False])
-        >>> mesh.plot(scalars=mask)
 
         """
         poly_data = self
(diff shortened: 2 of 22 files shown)


doc/source/conf.py
```diff
@@ -69,7 +69,6 @@
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "sphinx.ext.linkcode",  # This adds the button ``[Source]`` to each Python API site by calling ``linkcode_resolve``
-    "sphinx.ext.doctest",
```
tkoyama010, line 72: I am concerned that removing this line will not ensure the accuracy of the examples. I am trying to figure out how to address this issue. (May be able to handle this with other PRs.)

tkoyama010: > since we use pytest doctest-modules

I did not read this text. I'm a goofball.

pyvista/core/filters/poly_data.py
```diff
@@ -552,17 +555,17 @@ def curvature(self, curv_type='mean', progress_bar=False):
...
-        >>> curv  # doctest:+SKIP
-        array([0.20587616, 0.06747695, ..., 0.11781171, 0.15988467])
+        >>> hills.plot(scalars=curv)
 
-        Plot it.
+        Show the curvature array.
 
-        >>> hills.plot(scalars=curv)
+        >>> curv  # doctest:+SKIP
+        array([0.20587616, 0.06747695, ..., 0.11781171, 0.15988467])
```
tkoyama010, line 568: Could you show us your reason for leaving this behind?
I assume it is a floating point problem.

akaszynski: Correct. Until `doctest-modules` supports "close enough" natively at least. I know that numpy supports `may vary`, and we can consider adding that later, but it will require us to do additional work.

MatthewFlamm: doctest-modules does support close enough.  We already have it in the configuration here https://github.com/pyvista/pyvista/blob/9f5e989f2bde1c4b20d7ef6dcac039fa0dbf2c54/pyproject.toml#L159

The matching precision is based on the value written in the docstring.

tkoyama010 (approved):

PR #4420, opened 2023-05-14
Add PyHyperbolic3D as an external example

tkoyama010:
### Overview

Add PyHyperbolic3D as an external example.
griddip

### Details

- Thanks @stla reporting in https://github.com/pyvista/pyvista/discussions/4418 .


akaszynski:
Reduced image size down to 2MB. Used:
```
gifsicle --batch --optimize=3 --resize 200x200 --colors 64 griddip.gif
```

Recommend doing the same for other images.

stla:
Thanks.

In case if you want another example, I have some PyVista animations in **PyPlaneGeometry** (also on PyPI). For example:

akaszynski (approved):

PR #4421, opened 2023-05-14
Reduce giant image file size (> 10MB)

tkoyama010:
### Overview

When adding images like #4420, this setting does not accidentally commit a huge file.

### Details

- 10MB is the maximum size of a file that can be submitted for comment on GitHub.


banesullivan:
Has this been an issue? I'm mostly concerned about the local development overhead of adding pre-commit checks that don't have a clear added benefit. 

50Mb is the typical limit for committed files but I agree we shouldn't be committing files over even **1Mb** except in rare cases.

adeak:
I'd expect that the better solution for this is to add git lfs, and use a carefully curated gitattributes file that pushes binaries to lfs.

We could still try to avoid large text files, but those are very very rare and obvious in diffs.

tkoyama010:
Thanks for comments, please give me time to learn about the Git LFS.

banesullivan:
We don't want to enable Git LFS on this main repo. The vtk-data repo has it enabled.

adeak:
To elaborate on my point a bit: large files can be issues from multiple perspectives.

Large files can be an issue for downstream end users when they are part of the main library. They can be an issue for end users when they are part of documentation, because online docs load times and metered connection costs can be impacted. They can be an issue for developers with any of the above. They can be an issue for git, because large _binary_ files have no chance of being compressed after a while, so each commit that touches a 10 MB binary file bumps 10 MB on the size of the repo.

We likely won't have large files as part of the library;  example downloads being outsourced hopefully takes care of that. We can have large files in the gallery, see the example triggering this discussion. The large binary file issue specifically could be tackled with git lfs. We can still want to prevent too many of these files from popping up to keep the online docs as light as reasonably possible.

> We don't want to enable Git LFS on this main repo. The vtk-data repo has it enabled.

@banesullivan what is the relationship between these? The implication that the LFS quota is per-project, so we don't want to burn it in two places? Or something else? Because sometimes we _do_ have large binary files, see again the gallery images. And we _do_ not want to have them in raw git. And I don't imagine that their number would be noticeable next to the example datasets; then again I haven't looked into this, and I also don't know the size of the figures here nor the size of the vtk-data repo. But arguably if the size of the binaries in the main repo are comparable to that in the vtk-data repo then we should use the LFS quota for the main repo rather than the auxiliary one.

banesullivan:
> To elaborate on my point a bit: large files can be issues from multiple perspectives. .......

Yes, yes, yes, and yes to all your points

> We likely won't have large files as part of the library; example downloads being outsourced hopefully takes care of that.

Precisely -- this should prevent the need for any large file storage in this main repo

> We can have large files in the gallery, see the example triggering this discussion.

I totally missed that the example image was for the external gallery (thus commited in this repo) and not the generated gallery -- leading me to think the issue was with the docs hosting repo. This is my mistake.

> The large binary file issue specifically could be tackled with git lfs. We can still want to prevent too many of these files from popping up to keep the online docs as light as reasonably possible.

I agree we want to prevent these from popping up

> The implication that the LFS quota is per-project, so we don't want to burn it in two places? Or something else? Because sometimes we do have large binary files ...

It's been a while since I've looked at the LFS quotas but since PyVista is on the free OS tier I think we should try to keep LFS use to a bare minimum regardless of repo

> ... But arguably if the size of the binaries in the main repo are comparable to that in the vtk-data repo then we should use the LFS quota for the main repo rather than the auxiliary one.

I think we're on the same page here. Correct me if I'm wrong, but I don't think the main repo has many (any?) files near the LFS limit.

-----

Considering that this issue arose from a PR committing a large GIF to this repo, I am now in favor of adding this check. At first I thought this wasn't actually triggered by a strong need, but I see it now.

adeak:
> I think we're on the same page here. Correct me if I'm wrong, but I don't think the main repo has many (any?) files near the LFS limit.

Probably not, yeah, but I haven't checked.
 
> Considering that this issue arose from a PR committing a large GIF to this repo, I am now in favor of adding this check. At first I thought this wasn't actually triggered by a strong need, but I see it now.

But still on the server side (i.e. as part of CI) rather than a pre-commit hook, right? Considering how rare these files arise.

tkoyama010:
This change was suggested because of my own laziness in not wanting to check the file size, but perhaps it is my own habit that should be changed 😅 How about make this change only to resize scalar-bar-interactive.gif and see how often similar problems occur in the future?

adeak (approved):
The new gif has visible jitter and strong dithering, but really it's just about showing that the interactive scalar bar can be moved around, so I think this is fine. It's still 7.5 MB large which is a lot. LGTM.

PR #1602, opened 2021-08-20
Optimize flip_normals and fix is_all_triangles

akaszynski:
This is a simple fix to address @adeak's finding in #1601.  Strangely, I didn't find the same performance improvements, but rather saw a different low hanging fruit:

Originally we were copying the array when assigning it with:
```py
poly_data.faces = f
```

Instead, we can update it in situ with:
https://github.com/pyvista/pyvista/blob/c8ae9ecb42c85d415018cfe886f1b01cf4b7cc35/pyvista/core/filters/poly_data.py#L2096

#### Crude timings:

```py
>>> import pyvista as pv
>>> mesh = pv.Sphere(phi_resolution=1000, theta_resolution=1000)
>>> mesh
PolyData (0x7fcab5e1f340)
  N Cells:	1996000
  N Points:	998002
  X Bounds:	-5.000e-01, 5.000e-01
  Y Bounds:	-5.000e-01, 5.000e-01
  Z Bounds:	-5.000e-01, 5.000e-01
  N Arrays:	1

# poly_data.faces = f

>>> timeit mesh.flip_normals()
70.4 ms ± 342 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

# poly_data.faces[:] = f.ravel()

>>> timeit mesh.flip_normals()
51.5 ms ± 179 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)


##############################################################################
# Once optimized with the above (numpy==1.21.2)

# f[:, 1:] = f[:, 1:][:, ::-1]

>>> timeit mesh.flip_normals()
51.5 ms ± 179 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

# f[:, [1, 2, 3]] = f[:, [3, 2, 1]]

>>> timeit mesh.flip_normals()
54 ms ± 833 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

```

#### Bonus fixes
Convert is_all_triangles to a property.  This was the original problem @adeak pointed out.
- [x] Subclass boolean (really, `int`) and make it callable to allow for backwards compatibility. 
- [x] Add `n_vert` and `n_lines`.  This allows for an early exit of `is_all_triangles` without assembling evaluating the length of these arrays.

PR #1213, opened 2021-03-02
Improve PolyData Import

akaszynski:
### Cleanup `PolyData.__init__`
There are a variety of issues that could be addressed via a refactor and better documentation of our ``pyvista.PolyData`` class.  Additionally, there's a significant performance bottleneck when creating meshes without specifying the number of faces.  For example:

```py
>>> import numpy as np
>>> import pyvista as pv
>>> n = 1000000
>>> points = np.random.random((n, 3))
>>> faces = np.empty((n, 4), dtype=np.int64)
>>> faces[:, 0] = 3
>>> faces[:, 1:] = np.random.randint(n, size=(n, 3))
>>> faces = faces.ravel()
>>> timeit pv.PolyData(points, faces))
397 ms ± 2.23 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Of course, this can be addressed by not using ``faces.ravel``, but if your array is already flat and cannot be reshaped, but you know the number of faces in the array, we need to implement an option for that:

```
>>> timeit pv.PolyData(points, faces, n_faces=n)
7.16 ms ± 73 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Bottleneck can be traced to:
https://github.com/pyvista/pyvista/blob/4eeb7992df03937b3f5c478f78ed072ad86b3d1c/pyvista/utilities/cells.py#L55-L66

There's no way to get around this except for using something like ``Cython``.

---

This PR also provides the ability to create a lines PolyData directly on init with:

```py
>>> lines = np.hstack([[2, 0, 1], [2, 1, 2]])
>>> mesh = pyvista.PolyData(vertices, lines=lines)
```

This seems a bit cleaner than our existing:
```py
>>> lines = np.hstack([[2, 0, 1], [2, 1, 2]])
>>> mesh = pyvista.PolyData(vertices)
>>> mesh.lines = lines
```

PR #4439, opened 2023-05-19
Remove deprecations from pointset

MatthewFlamm:
### Overview

`CallableBool` was added in #1602.

`number of faces` starting raising an error for removal purposes in #1213

This is related to #4389, and might be recommended to merge after some time has passed to avoid any conflicts with patch changes, although the changes here are less likely to cause problems in my opinion.


akaszynski (approved):
LGTM. I'd prefer smaller PRs to make these changes easier to track.

Absolutely love getting rid of `CallableBool`.

PR #4480, opened 2023-05-31
Add Cell center property

MatthewFlamm:
### Overview

Adds property for the center of a cell.  This returns a different type of center compared to `DataSet.center` which is the center of hte bounding box.  This property returns the center of the cell from the parameterized coordinates within the cell.

### Details
See https://vtk.org/doc/nightly/html/classvtkCell.html#ae6d79848c5b56865eadac6bebbe723a1

For consistency, I changed all examples to use `import pyvista as pv` in this class.


pyvista/core/cell.py
```diff
@@ -356,13 +356,34 @@ def bounds(self) -> Tuple[float, float, float, float, float, float]:
...
+        (-0.0356, 0.00375, -0.498)
+
+        """
+        para_center = [0.0, 0.0, 0.0]
+        sub_id = self.GetParametricCenter(para_center)
+        sub_id = _vtk.mutable(0)
+        center = [0.0, 0.0, 0.0]
+        weights = [0.0] * self.n_points
+        self.EvaluateLocation(sub_id, para_center, center, weights)
+        return cast(Tuple[float, float, float], tuple(center))
```
MatthewFlamm, line 385: `DataSet.center` returns a list, however this is return is immutable.  I think I remember there being some discussion about this, but forget where it landed.

pyvista/core/cell.py
```diff
@@ -356,13 +356,34 @@ def bounds(self) -> Tuple[float, float, float, float, float, float]:
...
+        >>> mesh.get_cell(0).center
+        (-0.0356, 0.00375, -0.498)
+
+        """
+        para_center = [0.0, 0.0, 0.0]
+        sub_id = self.GetParametricCenter(para_center)
+        sub_id = _vtk.mutable(0)
+        center = [0.0, 0.0, 0.0]
+        weights = [0.0] * self.n_points
+        self.EvaluateLocation(sub_id, para_center, center, weights)
```
tkoyama010, line 384: Wouldn't this part be a good bonus method if the weight could be specified as the `parametric_center` method?

pyvista/core/cell.py
```diff
@@ -356,13 +356,34 @@ def bounds(self) -> Tuple[float, float, float, float, float, float]:
...
+        --------
+        >>> import pyvista as pv
+        >>> mesh = pv.Sphere()
+        >>> mesh.get_cell(0).center
+        (-0.0356, 0.00375, -0.498)
+
+        """
+        para_center = [0.0, 0.0, 0.0]
+        sub_id = self.GetParametricCenter(para_center)
+        sub_id = _vtk.mutable(0)
```
tkoyama010, line 381: If the variable `sub_id` in the first line is not used, the expression can be simple.
```suggestion
        _ = self.GetParametricCenter(para_center)
        sub_id = _vtk.mutable(0)
```

tkoyama010 (changes requested):

pyvista/core/cell.py
adeak: Last thing I could find was https://github.com/pyvista/pyvista/pull/3747.

MatthewFlamm: Thanks, I think this usage is consistent with that.

pyvista/core/cell.py
MatthewFlamm: `weights` is returned from `EvaluateLocation`.  There are probably other things that could be exposed using these methods, but I'm not sure exactly what you are proposing.

pyvista/core/cell.py
MatthewFlamm: Good catch! I think that we need to pass it into the following line.  I'm not sure why it needs to be mutable however, which is why I originally set it to 0.

MatthewFlamm: https://github.com/pyvista/pyvista/pull/4480/commits/0ca25ab5d9deea52b0a57627f33a9945901435f9 should address this

pyvista/core/cell.py
tkoyama010: Ah, I see, I had mistakenly thought that weight was something the user could specify rather than a return value.

banesullivan (approved):

tkoyama010:
Please hold off merging until we merge #4486 and resolve the conflict with it.

Issue #4445, opened 2023-05-22
Improve VTK imports to avoid GL-dependencies in core module


banesullivan:
Following up on some threads:

- https://github.com/pyvista/pyvista/discussions/3283
- https://github.com/pyvista/pyvista/pull/4337

Since the early days of PyVista, users have requested to use PyVista's `core` module and its robust wrappings of VTK mesh data types without having the needed GL dependencies of PyVista's plotting module. In the linked threads and various others there are compelling reasons why this should be supported.

Over time, we've become increasingly entrenched in cobbling all of the VTK imports together, forcing users to have the necessary libGL dependencies. This has been fine because the overwhelming majority of users also wanted to plot with PyVista. However, the tide is shifting and we're seeing more and more requests to use PyVista's well-documented and feature-rich core API for data processing needs outside of plotting (a testament to how truly powerful this API has become).

There are many ways to go about making this happen, but the approach in mind that keeps coming up as the simplest and likely easiest to maintain would be to *import the used VTK classes directly where they are needed* (rather than having a single `_vtk.py` module that imports everything).

Perhaps the even simpler approach would be to have two `_vtk.py` modules -- one in PyVista's `core` subpackage and one in the `plotting` subpackage (the plotting one would import the core one to keep functionality the same). Then users who want to avoid using the plotting module would change their import to `import pyvista.core as pvc`, keeping `import pyvista as pv` working the same (brings in all plotting capabilities as well).

There will be some nuances to address (like binding `pyvista.plot()` as the `plot()` method on data types) but we can address these fairly easily.

Overall, I think this is the simplest approach that will be generally easy to maintain.

The end goal for me here is to better isolate the `core` and `plotting` modules so that maybe one day, with enough community buy-in and momentum, we could split these into two separate packages (hopefully still in a mono-repo though)

Thoughts, @pyvista/developers?

larsoner:
> users who want to avoid using the plotting module would change their import to import pyvista.core as pvc, keeping import pyvista as pv working the same (brings in all plotting capabilities as well). 

I'm not 100% sure but don't think this bit will work -- if you import a submodule I think you get the higher-level one "for free" / automatically / unavoidably, e.g.:
```
>>> import pyvista.plotting
>>> import sys
>>> sys.modules['pyvista']
<module 'pyvista' from '/home/larsoner/python/pyvista/pyvista/__init__.py'>
```
and just to make sure it's not specific to PyVista at the moment:
```
[...]

PR #4486, opened 2023-06-01
Decouple core and plotting APIs to improve imports and tighten public namespace

banesullivan:
This PR aims to decouple the `pyvista.plotting` subpackage from the `pyvista.core` subpackage of PyVista: enhancing modularity, reducing dependency complexity, and de-cluttering the public namespace. This is a first step in addressing https://github.com/pyvista/pyvista/issues/4445

This is a big PR and while I understand there may be reservations to landing this, I have worked diligently to ensure there are very few breaking changes and absolutely no breaking changes the main, public `pyvista.*` namespace.

As we saw in #3764, reorganizing the imports is an undertaking -- this PR solves most of the issues we've been trying to fix for a while and sets us up to more easily resolve issues with mypy.

While this doesn't solve this issue #3764 is after, it does close #3764 as this is doing much of the same refactoring. Solving the mypy issues with `__all__` statements will come in a later PR.

@pyvista/developers, I'd like to keep momentum on this and merge somewhat soon -- please let me know any concerns you may have so that I can promptly address them.

## Changes

1. **Separate the `pyvista._vtk` module** into `pyvista.core._vtk_core` and `pyvista.plotting._vtk`. This reorganization will improve clarity and provide a clear demarcation between libGL-dependent VTK modules. `pyvista._vtk` no longer exists and `pyvista.plotting._vtk` is the full module now which itself imports all from `pyvista.core._vtk_core`. `pyvista._vtk` is still technically there as we `import *` from the `plotting` but this really shouldn't be used.

2. **Moved `pyvista.core.Texture`** into the `plotting` subpackage -- turns out `vtkTexture` requires importing the rendering modules in VTK. This will still be available from `pyvista.Texture` and is the only "core" class that had to move.

2. **Deprecated the `pyvista.utilities` module** by splitting it into two new modules specifically dedicated to core features (`pyvista.core.utilities`) and plotting features (`pyvista.plotting.utilities`). This structure enhances the segregation of responsibilities and simplifies future maintenance. While migrating these codes, I found that modules like `pyvista.utilities.helpers` had massive scope creep and crowding with a hodgepodge of features. I reorganize much of the contents of this module into several new modules (mostly under `pyvista.core.utilities`) such that array-related things will be in an array-focused module and similarly for other topics. See `pyvista.core.utilities.arrays`, `pyvista.core.utilities.misc`, `pyvista.core.utilities.helpers`, `pyvista.core.utilities.points`.
    - To alleviate any breaking changes, I implemented a `__getattr__` method in `pyvista.utilities` that attempts to find the requested feature from the two new modules. This is not full-proof, but it may help alleviate imports like `from pyvista.utilities import convert_array`. Imports like `from pyvista.utilities.helpers import *` are now broken -- though I don't suspect many downstream projects leveraged the utilities module in this manner.

3. **Reorganized imports:** Imports have been reorganized across the project, favoring explicit over implicit in making the public namespace cleaner and easier to navigate.
    - Previously, we were not running the `flake8` or `isort` pre-commit checks on any of the `__init__.py` files anywhere in PyVista. The work in this PR to finally makes it possible to clean up the imports in these files and now automatically run these checks. ***No longer are the days of circular imports!***
    - Star imports are still used in a few intentional places but have otherwise been reduced.

4. **Rename `pyvista.plotting.plotting` to `pyvista.plotting.plotter`** as this internal `plotting` module was overriding the `pyivsta.plotting` module breaking import paths.

5. **Oh so much cleanup** to the import paths. Many modules were being improperly imported in tests, documentation, examples, and internally, causing many of the previous circular import issues.

6. **Moving the ``plot()`` helper method to the top level.** I've moved the widely used `pyvista.plot()` method to the top level of the package in a new `_plot.py` module so that we can bind the `plot()` method to data types before import the plotting module. This will be most impactful in future work to address #4445, fully decoupling graphics-dependent imports from the core API.

7. **Consistent relative and absolute imports.** Made sure that we are consistently using relative and absolute imports within the PyVista package. The rule I have is relative for "this module and under" and absolute for anything outside or above the current module.

## Impact

Decoupling the `core` and `plotting` APIs makes PyVista more flexible for future growth.

This refactor makes it easier to update or alter these APIs independently, reduces potential points of failure, and simplifies the understanding of the underlying code structure. This change should also help streamline contributions from the community, as developers will find it easier to focus on the aspect they are interested in contributing to.

Further, moving towards a fully de-coupled `core` and `plotting` APIs will help bring the wider VTK community to PyVista.

## Testing

I've made sure all the tests pass locally -- let's see if CI returns happy!

I also added some new tests in `tests/namespace/` where I've saved a TXT file of all the public attributes of PyVista from `main` at this time in `tests/namespace/namespace-pre-refactor-clean.txt`. The "clean" version of this file removes external attributes like `sys`, `pi`, `pathlib`, etc., and also I've commented out several attributes or modules I do not think we intentionally exposed. All this to say that `pyvista.*` on this branch is fully compatible with `pyvista.*` on main, thus not breaking how 99% of users access the PyVista API.

Please provide your valuable feedback, and let's discuss any potential improvements.

## Future Work

I have lots more to come! But this is the heavy lifting that needed to be done first and I'm using this PR to "rip off the bandaid." Next steps include:

- Deprecating texture tracking on datasets: `vtkTexture` requires importing the rendering core module in VTK which we need to separate fully from the core API.
- Testing isolation: move test files into `core/` and `plotting/`
- Finishing what #3764 set out to fix with `__all__` definitions
- I'm not moving the `jupyter` module into plotting because most of the functionality here is already deprecated and will migrate the remaining code in a later PR


tkoyama010:
Thanks for the change! Let's review it carefully as this is significant.

adeak:
I doubt I'll have time to review this, but I recommend holding off merging all other non-dependabot PRs until this is merged to ensure that this PR doesn't aim for a moving target. The guaranteed conflicts can be more easily handled bit by bit on the other, smaller PRs.

banesullivan:
Agreed, thanks for chiming in @adeak -- with something this critical and wide reaching its important to leverage momentum

banesullivan:
Gotta say, I'm impressed with myself. Only a single testing failure and the same failure for the integration tests.

- Testing failure: `tests/test_utilities.py::test_linkcode_resolve`. I may solicit some help here in understanding what `linkcode_resolve` is trying to do and how I broke it.
- Integration tests: `ModuleNotFoundError: No module named 'pyvista.plotting.plotting'` -- this is expected and the only breaking API change. I will see if I can implement a `plotting.plotting` module to access things but we *really* need to remove this module as it messes up the top-level namespace.

larsoner:
>  I will see if I can implement a plotting.plotting module to access things but we really need to remove this module as it messes up the top-level namespace.

IIRC you could add a `pyvista/plotting/plotting.py` (or maybe `pyvista/plotting/plotting/__init__.py`?) with a `__getattr__` that dynamically gets "stuff" from somewhere else (using a nested import) and returns it, possibly even emitting a `warnings.warn` or so. See for example how we deprecated stuff in SciPy

https://github.com/scipy/scipy/blob/62e2879b34f6fc167375f3e37a0e1fc727cc88dd/scipy/interpolate/polyint.py#L24

banesullivan:
Thanks for the suggestion, @larsoner! I just pushed a change with a `__getattr__` override in a new `pyvisat/plotting/plotting/__init__.py` module that implements this and issues a deprecation warning.

I had already made the same mechanism for `pyvista/utilities/__init__.py` and would appreciate feedback on this "hack"

banesullivan:
Hm. It seems the integration tests are now failing because of the warning... perhaps I'll need to set this to be less strict?

banesullivan:
@larsoner, integration tests for PyVistaQt and MNE are now passing, but they ignore `PyVistaDeprecationWarning` -- meaning this PR is deprecating (not breaking) parts of the API used in both these packages

larsoner:
Sounds like a good approach. Then hopefully once this is merged I can update those quickly to accommodate the deprecation, then the warning filter can be removed here

pyvista/core/_vtk_core.py
tkoyama010, line 1: Personally, I think the file name should be `_vtk.py`. It seems redundant as there is already a `core` in the directory name.

banesullivan: This is to prevent `pyvista._vtk` from being limited to only the core module as we still currently have `from pyvista.core import *` in `pyvista/__init__.py`

tkoyama010: Got it. Thanks!

pyvista/__init__.py
```diff
@@ -1,30 +1,30 @@
 """PyVista package for 3D plotting and mesh analysis."""
+# flake8: noqa: F401
```
tkoyama010, line 2: I agree with this policy, which is explicitly stated in the comments, rather than excluding files from the flake8 configuration.

tkoyama010: > Star imports are still used in a few intentional places but have otherwise been reduced.

Also, it is good for reducing star imports.

.github/workflows/integration-tests.yml
```diff
@@ -52,7 +52,7 @@ jobs:
           path: ~/mne_data
       - run: ./tools/github_actions_download.sh
         working-directory: mne-python
-      - run: pytest mne/viz/_brain mne/viz/tests/test_3d.py mne/viz/backends
+      - run: pytest mne/viz/_brain mne/viz/tests/test_3d.py mne/viz/backends -W ignore::pyvista.core.errors.PyVistaDeprecationWarning
```
tkoyama010, line 55: This may not be relevant to this discussion, but what are the reasons for integration testing other repositories? In the past, there have been cases where CI has been stopped due to errors that occurred in repositories other than PyVista. What is the necessity of running CI here?

pyvista/_typing.py
```diff
@@ -27,7 +25,7 @@
     Dict[str, Union[int, float, str]],
     str,
     "Color",
-    _vtk.vtkColor3ub,
+    # _vtk.vtkColor3ub,
```
tkoyama010, line 28:
```suggestion
```

pyvista/plotting/errors.py
```diff
@@ -0,0 +1,30 @@
+"""Core and plotting errors."""
```
tkoyama010, line 1: Only Plotting errors should be here :)
```suggestion
"""Plotting errors."""
```

tkoyama010:
I have checked up to half of the 140 files. Please give me a little more time.

pyvista/_typing.py
banesullivan: Ah thanks for pointing this out! We may actually want to keep that type in there. I commented it out because it required importing the plotting module and I was trying to get the top level to avoid any plotting imports. I will push some changes that clean this up rather than removing this

.github/workflows/integration-tests.yml
banesullivan: These integration tests were added in https://github.com/pyvista/pyvista/pull/3967 and spun out of https://github.com/pyvista/pyvista/pull/3385. 

We are in a scenario where downstream projects heavily depend on PyVista, and these integration tests play a crucial role in ensuring that modifications to PyVista do not inadvertently break or disrupt these projects. 

By conducting integration tests here in PyVista's CI, we can assess how well the library integrates with external software and identify potential issues that may arise due to changes before the issues arise. This approach provides a safety net, allowing us to catch and rectify any compatibility problems early on -- thereby reducing "release pain". By validating that our changes do not break downstream projects through CI tests here, we can instill confidence in the stability and reliability of PyVista, fostering a smooth development process and minimizing the risk of disruptions in downstream projects that may ultimately have more users than PyVista directly. 

Ultimately, integration testing serves as a vital measure to safeguard the overall integrity and functionality of what we are shipping

pyvista/_typing.py
banesullivan: Cleaned this up

banesullivan: Cleaned this up

pyvista/plotting/plotting/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+import inspect
+import warnings
+
+from pyvista.core.errors import PyVistaDeprecationWarning
+
+
+def __getattr__(name):
+    module = importlib.import_module('pyvista.plotting.plotter')
+    try:
+        value = inspect.getattr_static(module, name)
```
adeak, line 12: The logic might be fine here, again just making sure that nothing falls through the cracks.

On `main` these were the contents of `pyvista.plotting.plotting`:
```
>>> names
['Actor', 'BasePlotter', 'BoundsLike', 'Camera', 'Color', 'CompositePolyDataMapper', 'DataSetMapper', 'Dict', 'FONTS', 'FieldAssociation', 'FixedPointVolumeRayCastMapper', 'GPUVolumeRayCastMapper', 'KILL_DISPLAY', 'MissingDataError', 'OpenGLGPUVolumeRayCastMapper', 'Optional', 'PickingHelper', 'Plotter', 'PointGaussianMapper', 'Property', 'PyVistaDeprecationWarning', 'RenderWindowInteractor', 'RenderWindowUnavailable', 'Renderer', 'Renderers', 'SUPPORTED_FORMATS', 'ScalarBars', 'SmartVolumeMapper', 'Thread', 'USE_SCALAR_BAR_ARGS', 'UnstructuredGridVolumeRayCastMapper', 'Volume', 'VolumeProperty', 'WidgetHelper', '_ALL_PLOTTERS', '__annotations__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', '_coerce_pointslike_arg', '_common_arg_parser', '_kill_display', '_vtk', '_warn_xserver', 'abstract_class', 'active_scalars_algorithm', 'algorithm_to_mesh_handler', 'assert_empty_kwargs', 'close_all', 'collections', 'contextmanager', 'convert_array', 'ctypes', 'decimation_algorithm', 'deepcopy', 'export_plotter_vtkjs', 'extract_surface_algorithm', 'get_array', 'get_array_association', 'get_cmap_safe', 'image_from_window', 'io', 'is_pyvista_dataset', 'log', 'logging', 'matplotlib', 'normalize', 'np', 'numpy_to_texture', 'opacity_transfer_function', 'os', 'parse_font_family', 'pathlib', 'platform', 'pointset_to_polydata_algorithm', 'prepare_smooth_shading', 'process_opacity', 'pyvista', 'raise_not_matching', 'run_image_filter', 'scooby', 'set_algorithm_input', 'sys', 'textwrap', 'time', 'triangulate_algorithm', 'uses_egl', 'warnings', 'weakref', 'wrap', 'wraps']
```
All of these are defined on the new `pyvista.plotting.plotter`. Yet from these, here are the ones that are not defined on `pyvista.plotting` of this branch:
```
>>> [name for name in names if not hasattr(pyvista.plotting, name)]
['BoundsLike', 'Dict', 'FieldAssociation', 'FixedPointVolumeRayCastMapper', 'GPUVolumeRayCastMapper', 'KILL_DISPLAY', 'MissingDataError', 'OpenGLGPUVolumeRayCastMapper', 'Optional', 'PickingHelper', 'PointGaussianMapper', 'PyVistaDeprecationWarning', 'RenderWindowInteractor', 'RenderWindowUnavailable', 'Renderers', 'SUPPORTED_FORMATS', 'ScalarBars', 'SmartVolumeMapper', 'Thread', 'USE_SCALAR_BAR_ARGS', 'UnstructuredGridVolumeRayCastMapper', 'Volume', 'VolumeProperty', '_ALL_PLOTTERS', '__annotations__', '_coerce_pointslike_arg', '_common_arg_parser', '_kill_display', '_warn_xserver', 'abstract_class', 'active_scalars_algorithm', 'assert_empty_kwargs', 'collections', 'contextmanager', 'convert_array', 'ctypes', 'decimation_algorithm', 'deepcopy', 'extract_surface_algorithm', 'get_array', 'get_array_association', 'io', 'is_pyvista_dataset', 'log', 'logging', 'matplotlib', 'normalize', 'np', 'os', 'pathlib', 'platform', 'pointset_to_polydata_algorithm', 'prepare_smooth_shading', 'process_opacity', 'pyvista', 'raise_not_matching', 'scooby', 'sys', 'textwrap', 'time', 'triangulate_algorithm', 'uses_egl', 'warnings', 'weakref', 'wrap', 'wraps']
```

It seems to me (but I might have misread something) that all these names are found by `inspect.getattr_static()`, yet the recommended `from pyvista.plotting import Renderers` (or any of the other items) will fail.

Is this the case, and if yes, do we/can we want to change anything about this? Either in the logic here, or about the accessibility of names. And I haven't checked but odds are that the deprecated `utilities` __init__.py` is in a similar situation.

pyvista/plotting/renderer.py
```diff
@@ -355,15 +358,15 @@ def camera_position(self, camera_location):
 
         elif isinstance(camera_location[0], (int, float)):
             if len(camera_location) != 3:
-                raise pyvista.core.errors.InvalidCameraError
+                raise InvalidCameraError
```
adeak, line 361: For another day: these messageless exceptions are weird.

pyvista/plotting/utilities/cubemap.py
```diff
@@ -0,0 +1,135 @@
+"""Cubemap utilities."""
```
adeak, line 1: This is an oddly specific module. Was it so fine-grained before this refactor?

pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -12,6 +12,8 @@ def _get_sg_image_scraper():
     'matplotlib' and 'mayavi'. Details on this implementation can be found in
     `sphinx-gallery/sphinx-gallery/494`_
 
+    This must be at the top level namespace of PyVista.
```
adeak, line 15: Might prevent a double-take if we rephase to something like
```suggestion
    This must be imported into the top level namespace of PyVista.
```

pyvista/report.py
```diff
@@ -0,0 +1,230 @@
...
+        minor = ver.GetVTKMinorVersion()
+        micro = ver.GetVTKBuildVersion()
+    except AttributeError:  # pragma: no cover
+        warnings.warn("Unable to detect VTK version. Defaulting to v4.0.0")
+        major, minor, micro = (4, 0, 0)
+
+    return version_info(major, minor, micro)
+
+
+vtk_version_info = VTKVersionInfo()
```
adeak, line 28: I'd never have thought to look in this module for `vtk_version_info`.

pyvista/utilities/__init__.py
```diff
@@ -1,77 +1,39 @@
...
+"""Deprecated utilities module."""
+import importlib
+import inspect
+import warnings
+
+from pyvista.core.errors import DeprecationError, PyVistaDeprecationWarning
+
+
+def __dir__():
+    raise DeprecationError('The `pyvista.utilities` module has been deprecated.')
```
adeak, line 10: Is it not weird to raise in `__dir__()` when `__getattr__()` works?

pyvista/utilities/__init__.py
```diff
@@ -1,77 +1,39 @@
...
+
+
+def __getattr__(name):
+    # TODO: for some reason this runs twice
```
adeak, line 14: What runs twice and how? Are you saying that the utilities module gets executed twice? And is it twice per `__getattr__()` call? Or twice overall, instead of the one that we'd expect due to module caching?

pyvista/_plot.py
```diff
@@ -0,0 +1,307 @@
+"""PyVista's famous ``plot()`` helper method.
+
+This method is placed at the top-level to allow us to easily bind
+the method to all of the core datatypes before importing the
+``pyvista.plotting`` module and libGL dependent VTK modules.
+This is necessry for future versions of PyVista that will fully
```
adeak, line 6:
```suggestion
This is necessary for future versions of PyVista that will fully
```

tests/test_utilities.py
```diff
@@ -1,4 +1,4 @@
-""" test pyvista.utilities """
+""" test pyvista utilities """
```
adeak, line 1:
```suggestion
"""Test pyvista utilities."""
```
Or since we don't seem to be including other utilities,
```suggestion
"""Test pyvista core utilities."""
```

adeak (commented):
Some comments after a cursory look across the changes, ranging from typo/nitpicky to more high-level. Haven't compared namespaces yet, or checked whether there are any warnings during doc building.

In any case kudos for the mountain of work you've done for this @banesullivan.

pyvista/core/utilities/docs.py
```diff
@@ -41,6 +41,10 @@ def linkcode_resolve(domain, info, edit=False):
     modname = info['module']
     fullname = info['fullname']
 
+    # Little clean up to avoid pyvista.pyvista
+    if fullname.startswith(modname):
+        fullname = fullname.lstrip(f"{modname}.")
```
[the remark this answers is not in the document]
banesullivan: I always forget this is how `lstrip` works 🤦🏻‍♂️  thanks for pointing this out

pyvista/core/_typing_core.py
```diff
@@ -0,0 +1,12 @@
+"""Type aliases for type hints."""
+from __future__ import annotations
```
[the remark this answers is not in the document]
dcbr: IIRC this was added for the `ColorLike` type, indeed related to type aliases for the docs.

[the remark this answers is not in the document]
adeak: Thanks @dcbr, that indeed rings a bell. In that case we might want to check if this still works after the refactor and whether the future import is still necessary. I can try looking up the original discussion.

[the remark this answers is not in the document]
adeak: Cf. https://github.com/pyvista/pyvista/pull/2175#issuecomment-1035338661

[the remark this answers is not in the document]
adeak: OK, it seems the structure hasn't changed. In `pyvista/_typing.py` we still have `ColorLike` with the future import, as before. And the five names moved to `pyvista/core/_typing_core.py` are also the kinds of aliases for which it makes sense to get this working in sphinx.

In a local build I see missing intersphinx linkage in a signature but working alias link in parameter docs:
Screenshot from 2023-06-04 16-40-26

And for e.g. `BoundsLike` here we have
Screenshot from 2023-06-04 16-42-26

Even though the code has
```py
def bounds(self) -> BoundsLike:
```

AAAH, I remember why!
https://github.com/pyvista/pyvista/blob/23ebd3851bcf0f0f6a2f3d109719bab7f067d140/doc/source/conf.py#L129-L131

We'll probably want to add these other aliases here, or if we don't, omit the future import here for now. This all can be handled in a separate PR, or here. Up to @banesullivan.

pyvista/plotting/renderer.py
tkoyama010:
```suggestion
                raise InvalidCameraError('camera location vector's dimension is not equal to 3')
```

adeak: Thanks @tkoyama010 but there's a bunch of others in this module which is why I didn't post a suggestion here.

pyvista/core/utilities/fileio.py
```diff
@@ -428,6 +443,28 @@ def read_plot3d(filename, q_filenames=(), auto_detect=True, attrs=None, progress
...
+    bool
+        ``True`` if ``obj`` is an ``meshio.Mesh``.
+
+    """
+    try:
+        import meshio
+
+        return isinstance(obj, meshio.Mesh)
+    except ImportError:
+        return False
```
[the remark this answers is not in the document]
banesullivan: I don't think this would ever yield a false negative... If `meshio` isn't installed, then the passed object couldn't be a `meshio.Mesh`.

This API isn't intended to be used publically AFAIK. This is mostly an internal helper and I think this should be left as is for this PR due to how it will require code cleanup/usage elsewhere in the internal code

pyvista/demos/logo.py
```diff
@@ -41,7 +42,9 @@ def atomize(grid, shift_fac=0.1, scale=0.9):
 
 def text_3d(string, depth=0.5):
     """Create 3D text."""
-    vec_text = _vtk.vtkVectorText()
+    from vtkmodules.vtkRenderingFreeType import vtkVectorText
```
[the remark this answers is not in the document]
banesullivan: This is because the import at the top was changed to `from pyvista.core import _vtk_core as _vtk` as the rest of the imports in this file from `_vtk` are only using core-features (no plotting/libGL-dependednt modules). Since `vtkVectorText` comes from `vtkRenderingFreeType` it cannot be in the `_vtk_core` module and thus I've changed this in a few places to import at "runtime" within the called function. Another example is `pyvista.Text3D`

pyvista/examples/__init__.py
```diff
@@ -1,7 +1,7 @@
...
-from pyvista.examples.examples import *
-from . import planets
-from . import gltf
-from . import vrml
+# flake8: noqa: F401
+
+from . import gltf, planets, vrml
 from .cells import plot_cell
+from .downloads import *
+from .examples import *
```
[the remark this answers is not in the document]
banesullivan: Absolutely, this PR is already beast enough and I'd like to have a follow up PR that removes *all* star imports and implements `__all__`

pyvista/plotting/_typing.py
```diff
@@ -1,21 +1,17 @@
-"""Type aliases for type hints."""
-
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Dict, List, Sequence, Tuple, Union
+from typing import TYPE_CHECKING, Dict, Sequence, Tuple, Union
 
 import numpy as np
 
+from pyvista.core._typing_core import *  # noqa: F403, F401
```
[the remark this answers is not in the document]
banesullivan: Fair point, will do

pyvista/plotting/lights.py
```diff
@@ -11,7 +11,8 @@
 except ImportError:  # pragma: no cover
     from vtk import vtkLight, vtkLightActor, vtkMatrix4x4
 
-from ..utilities.helpers import vtkmatrix_from_array
+from pyvista.core.utilities.arrays import vtkmatrix_from_array
```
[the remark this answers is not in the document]
banesullivan: I worked to implement a policy for absolute vs. relative imports (let me know what you think of this):

- If "this" module or lower, use relative imports. E.g., `from ._property import Property` in `pyvista/plotting/actor.py` or `from .utilities import blah`
- Otherwise, use absolute imports. E.g., `from pyvista.core.utilities.arrays import `vtkmatrix_from_array` in this file.

Why? 

- To avoid namespace conflicts since there is a `utilities` module in both `pyvista.core` and `pyvista.plotting` now.
- Because I plan to have follow up work that will actually split the `pyvista.core` and `pyvista.plotting` modules into totally separate packages and having the imports already split here will ultimately make this easier.
- For complete clarity. This splits the imports at the top level of the file and developers can easily distinguish what is coming from another part of the API and what is coming from "this" part of the API. Helping clarity what is a "core" utility vs a "plotting" utility

pyvista/plotting/renderer.py
banesullivan: Syntax isn't right above. Let's resolve this in a follow up PR and do them all at once

pyvista/plotting/utilities/cubemap.py
banesullivan: I wasn't too sure where to put these. They were previously in `pyvista.utilities.helpers`... perhaps I could put them in `pyvista.plotting.utilities.helpers`? I made a dedicated module because there were at least 3 "cubemap" related things

pyvista/report.py
banesullivan: Hm. I thought it was intuitive because this module is how/where we generate reports and get all the other versioning info for the system/other packages.

Perhaps this should go in `pyvista/core/_vtk_core.py`?

pyvista/utilities/__init__.py
banesullivan: Honestly, I don't know. I totally cargo-culted and copied this from the scipy example

pyvista/utilities/__init__.py
banesullivan: It seems to be twice per `__getattr__()` call

pyvista/plotting/_vtk.py
```diff
@@ -0,0 +1,108 @@
+"""
+Import from vtk.
+
+These are the modules within vtk that must be loaded across pyvista.
+Here, we attempt to import modules using the ``vtkmodules``
+package, which lets us only have to import from select modules and not
+the entire library.
```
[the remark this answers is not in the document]
banesullivan: Good point, will do

> Actually, I'd go so far as to rename _vtk.py to _vtk_plotting.py. If anything, I'd expect the unqualified _vtk to be the core one...

I generally agree but am reluctant to fo this as this time. This is because of the `*` imports and generally trying to keep the namespace the same as it previously was. This way `pyvista._vtk` is still all VTK imports, GL-dependent and all.

When I eventually split the core and plotting modules into separate packages, both these modules will just be called `_vtk.py` in their respective packages.

pyvista/plotting/plotting/__init__.py
banesullivan: Thanks for bringing this up. I'll make sure the `pyvista.plotting` namespace works the same

.github/workflows/integration-tests.yml
```diff
@@ -70,5 +70,5 @@ jobs:
           pyvista: false
       - run: pip install -ve ./pyvistaqt -r ./pyvistaqt/requirements_test.txt PyQt6
       - run: pip install -ve .
-      - run: pytest ./tests
+      - run: pytest ./tests -W ignore::pyvista.core.errors.PyVistaDeprecationWarning
```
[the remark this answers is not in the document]
banesullivan: Looks like integration test is passing

pyvista/core/utilities/fileio.py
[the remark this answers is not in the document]
MatthewFlamm: This function is only used in one place in my search, in `pyvista.core.utilities.helpers` inside the `wrap` function.  It is used to check if we can wrap the object from meshio.  In this case, if we cannot import `meshio` or it is not a `meshio.Mesh` then it returns `False` and it doesn't wrap.  If we raise, we would just change the code to catch the error in `wrap` and not wrap.  This all smells like a private function to me.  This function might be better named `can_wrap_meshio` or something like this.   Or maybe the description or Returns description could be updated to reflect this nuance.  I'd vote for private usage if it is on-topic for this PR.

It might be better located directly in `pyvista.core.utilities.helpers`, or just inlined inside `wrap` and completely removing it.  This function isn't necessarily directly tied to IO, even though it is likely often (or always?) used that way.  I also do not see the value, other than avoiding a breaking change, in keeping this imported in `pyvista.core.utilities`.

But in the effort to keep this PR somewhat manageable, I think the most actionable decision is where to put it for now and whether to import it into `pyvista.core.utilities`.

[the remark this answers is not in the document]
adeak: > This API isn't intended to be used publically AFAIK. This is mostly an internal helper and I think this should be left as is for this PR due to how it will require code cleanup/usage elsewhere in the internal code

OK.



> This all smells like a private function to me. This function might be better named `can_wrap_meshio` or something like this. Or maybe the description or Returns description could be updated to reflect this nuance. I'd vote for private usage if it is on-topic for this PR.

Sounds good to me (as the eventual state of this helper). Thank you both.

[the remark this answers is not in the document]
adeak: Opened https://github.com/pyvista/pyvista/issues/4493 to track this.

pyvista/plotting/lights.py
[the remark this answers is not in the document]
banesullivan: > sounds like the trump here. No point doing anything else if relative imports would be broken when that split happens.

Indeed the trump card I'll play here

But also, I want to highlight another justification which I'm paraphrasing from someone else I was talking to about this:

> typically an `__init__.py` will import things below it (to re-expose them at the higher level). If some of these child modules want to import a sibling, it will have to go through the parent `__init__.py`, which causes a fragile partial import situation, which often outright breaks (depending on the order of imports within `__init__.py`, which is not normally something you care about, particularly when using isort ). So, the exception is that you should use relative imports for siblings and children sub-modules only.

I'm pretty sure since we weren't following this, its the reason PyVista's imports were so fragile and tended to become circular

pyvista/plotting/renderer.py
adeak: Actually, I was almost done writing an issue for this when I realised that probably all of these are custom exception types _with a default message_. And raising just the type implicitly instantiates the type, so this works. We probably have nothing to do here. Sorry for the noise.

In case you want to look at it, here's what made the penny drop for me:
```
$ git grep '^ *raise [^(]\+$'
pyvista/core/composite.py:                    raise err
pyvista/core/composite.py:        raise StopIteration
pyvista/core/composite.py:                            raise err
pyvista/core/dataset.py:            raise RuntimeError  # this should never be reached with err=True
pyvista/core/filters/poly_data.py:            raise NotAllTrianglesError
pyvista/core/pointset.py:        raise PointSetCellOperationError
pyvista/core/pointset.py:        raise PointSetCellOperationError
pyvista/core/pointset.py:        raise PointSetCellOperationError
pyvista/core/pointset.py:        raise PointSetCellOperationError
pyvista/core/pointset.py:        raise PointSetDimensionReductionError
pyvista/core/pointset.py:        raise PointSetDimensionReductionError
pyvista/core/pointset.py:        raise PointSetDimensionReductionError
pyvista/core/pointset.py:        raise PointSetDimensionReductionError
pyvista/core/pointset.py:        raise PointSetDimensionReductionError
pyvista/core/pointset.py:        raise PointSetCellOperationError
pyvista/core/pointset.py:        raise PointSetCellOperationError
pyvista/core/pointset.py:        raise PointSetCellOperationError
pyvista/core/pointset.py:        raise PointSetCellOperationError
pyvista/examples/downloads.py:                raise OSError
pyvista/plotting/plotting.py:                raise KeyboardInterrupt
pyvista/plotting/renderer.py:                raise pyvista.core.errors.InvalidCameraError
pyvista/plotting/renderer.py:                    raise pyvista.core.errors.InvalidCameraError
pyvista/plotting/renderer.py:                    raise pyvista.core.errors.InvalidCameraError
pyvista/utilities/algorithms.py:            raise e
pyvista/utilities/algorithms.py:            raise e
pyvista/utilities/algorithms.py:            raise e
pyvista/utilities/algorithms.py:            raise e
pyvista/utilities/reader.py:            raise NotImplementedError
pyvista/utilities/reader.py:                raise err
tests/plotting/test_plotting.py:            raise err
tests/tinypages/plot_polygon.py:    raise NotImplementedError
```
The only sus one in there is the `raise RuntimeError` which should probably be an assertion error. Not worth doing more work for, probably.

Consider this resolved indeed.

pyvista/plotting/utilities/cubemap.py
adeak: Well, they smell a bit too fine-grained for me. But this is not a very strong stance on my part. We can also leave as is.

pyvista/report.py
adeak: It might just be me, but yeah, that would be a better place for me. But it's also possible that I'm the one who's weird here. In any case we can look at `type(vtk_version_info)` and that tells us where it's defined...

pyvista/utilities/__init__.py
adeak: The `scipy` example I know returns `__all__` in `__dir__()` https://github.com/scipy/scipy/blob/1305da5985cbb1524aaa58b1296647bb122e6249/scipy/__init__.py#L182-L183

pyvista/utilities/__init__.py
adeak: Can you explain in what situation you see this run twice? I can't trigger that in a quick test. But I do see another issue with two exception contexts, will post a suggestion there separately.

pyvista/utilities/__init__.py
```diff
@@ -1,77 +1,39 @@
...
+        import_path = f'from pyvista.core.utilities import {name}'
+    except AttributeError:
+        utils = importlib.import_module('pyvista.plotting.utilities')
+        try:
+            value = inspect.getattr_static(utils, name)
+            import_path = f'from pyvista.plotting.utilities import {name}'
+        except AttributeError:
+            raise AttributeError(
+                f'Module `pyvista.utilities` has been deprecated and we could not automatically find `{name}` in `pyvista.core.utilities` or `pyvista.plotting.utilities`.'
+            )
```
adeak: I'd suppress the exception context so that users only see this latter exception.
```suggestion
            ) from None
```

adeak: Hmm, I think I posted this to a single commit, let me repost properly. Sorry.

pyvista/utilities/_getattr_.py
```diff
@@ -0,0 +1,41 @@
...
+            new_module_path = 'pyvista'
+            module = importlib.import_module(new_module_path)
+            try:
+                # All else, try top level namespace
+                value = inspect.getattr_static(module, name)
+                import_path = f'from {new_module_path} import {name}'
+            except AttributeError:
+                raise AttributeError(
+                    f'Module `pyvista.utilities` has been deprecated and we could not automatically find `{name}` in `pyvista.core.utilities` or `pyvista.plotting.utilities` or the top level namespace.'
+                )
```
adeak, line 33: Ah, I see my earlier suggestion here was stale... github showed me an earlier state.

Anyway, it's probably still relevant to suppress the exception context here:
```suggestion
                ) from None
```

pyvista/utilities/__init__.py
banesullivan: aha, its not executing twice. What's happening is warnings are being thrown when `__path__` is imported which is done behind the scenes by Python. I need to make sure things like `__path__` and `__file__` are actually resolving correctly

banesullivan: This is proving to give me a headache... this `__getattr__` override may not be the right approach. Will come back to this

pyvista/_plot.py
```diff
@@ -0,0 +1,307 @@
...
+        pl.show_grid()
+    elif show_bounds:
+        pl.show_bounds()
+
+    if cpos is None:
+        cpos = pl.get_default_cam_pos()
+        pl.camera_position = cpos
+        pl.camera_set = False
+    else:
+        pl.camera_position = cpos
```
tkoyama010, line 282:
```suggestion
    if cpos:
        pl.camera_position = cpos
    else:
        cpos = pl.get_default_cam_pos()
        pl.camera_position = cpos
        pl.camera_set = False
```

pyvista/core/utilities/arrays.py
```diff
@@ -0,0 +1,732 @@
...
+        # VTK default fonts only support ASCII. See https://gitlab.kitware.com/vtk/vtk/-/issues/16904
+        if np.issubdtype(arr.dtype, np.str_) and not ''.join(arr).isascii():  # avoids segfault
+            raise ValueError(
+                'String array contains non-ASCII characters that are not supported by VTK.'
+            )
+        vtkarr = _vtk.vtkStringArray()
+        ########### OPTIMIZE ###########
+        for val in arr:
+            vtkarr.InsertNextValue(val)
+        ################################
```
tkoyama010, line 555: What is this optimization? If it is optimization, I think a comment is in order, but want a detailed explanation.

pyvista/core/utilities/arrays.py
```diff
@@ -0,0 +1,732 @@
...
+        if isinstance(name, str):
+            vtkarr.SetName(name)
+        return vtkarr
+    # Otherwise it is a vtk array and needs to be converted back to numpy
+    ############### OPTIMIZE ###############
+    nvalues = arr.GetNumberOfValues()
+    return np.array([arr.GetValue(i) for i in range(nvalues)], dtype='|U')
+    ########################################
```
tkoyama010, line 563: Same with https://github.com/pyvista/pyvista/pull/4486/files#r1218535992 .

pyvista/utilities/__init__.py
adeak: Can you show a use case that's giving you trouble? This seems to work in a casual test, but there might be cases I'm forgetting:
```py
def __getattr__(name):
    if name.startswith('__') and name.endswith('__'):
        # don't proxy dunder names
        raise AttributeError(f'Refusing to proxy {name!r}.')
```
We could probably relax the check and refuse to expose anything that starts with a single underscore.

pyvista/core/utilities/arrays.py
banesullivan: This is pre-existing, the blame is currently messed up. Cannot address in this PR

pyvista/core/utilities/arrays.py
banesullivan: This is pre-existing, the blame is currently messed up. Cannot address in this PR

pyvista/utilities/_getattr_.py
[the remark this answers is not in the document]
banesullivan: I'm going to revert this altogether for this PR and follow up with a deprecated `utilities` module in a separate PR to get the blame right.

pyvista/utilities/__init__.py
banesullivan: This will be addressed in a follow up PR

.github/workflows/integration-tests.yml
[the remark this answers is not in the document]
banesullivan: This PR is merging soon and will follow up in a separate PR with dedicated testing and then stage the removal of this warning ignore

pyvista/core/_typing_core.py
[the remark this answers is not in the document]
banesullivan: removed the future import, follow up work can happen later

.github/workflows/integration-tests.yml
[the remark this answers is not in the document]
larsoner: Downstream PR https://github.com/mne-tools/mne-python/pull/11721 has been merged so feel free to make this change!
```suggestion
      - run: pytest mne/viz/_brain mne/viz/tests/test_3d.py mne/viz/backends
```

banesullivan:
I'm currently running a full documentation build locally -- no further changes on this branch unless that build/test requires it

pyvista/plotting/plotting/__init__.py
banesullivan: Will add validation testing in a follow up PR

pyvista/report.py
akaszynski: Seems like this belongs here or in a utilities module (e.g. `core/utilities`).

PR is so big it's killing the web interface for me, recommend a follow-up PR if needed.

pyvista/plotting/plotting/__init__.py
banesullivan: Yep. The `pyvista.utilities` module is being deprecated. I had changes here that implemented an interface for this import to issue a deprecation warning, but it ruined the history/blame and I will have to implement as a follow on PR

akaszynski (dismissed):
This PR addresses a concern I've had for a while regarding separating out our plotting and data APIs into separate submodules. I really appreciate your work, and while this will mean some refactoring for any libraries dependent on pyvista, I think it's worth it given how much this library has grown over the years.

Local testing reveals no issues, but we should considering the following follow ups.
- Handle `ModuleNotFoundError` better. For example, this import fails without a helpful error message `from pyvista.utilities.cells import numpy_to_idarr`
- Show a diagram of the library layout. This can be helpful for casual users as well as developers/contributors.

adeak (approved):
Approving with the caveat that open points and half-broken states will be fixed in follow-up PRs for ease of transparency. Thanks @banesullivan.

tkoyama010 (approved):
@banesullivan I will make a follow-up PR to pre-existing code fix suggestions that we found in this PR. Thanks for this change. I believe this will accelerate the future development of PyVista.

adeak (approved):

PR #4500, opened 2023-06-07
Reorganize to decouple plotting and core API tests

banesullivan:
This is follow on work to #4486 to organize the testing directory better, separating plotting dependent tests from the core API tests.

The only remaining plotting-dependent feature in the cora API tests are around texture tracking which I will deprecate in a later, separate pull request.

I'd like to have the turn around on this be somewhat quick if possible since it's mostly `git mv` operations and no changes under `pyvista/`

Notable changes

- Texture-related tests are (all but one) moved under `tests/plotting/`)
    - Places where the `globe` fixture are switched to `tetbeam` are due to this.
- The auto-enabled `check_gc` fixture is now only auto-enabled for specific files. We should re-enable this for all of `tests/plotting/` and fix the GC issues in a separate PR.
- Introduction of a new PyTest marker `pytest.mark.skip_plotting` that can be use anywhere in the `tests/plotting/` directory to avoid the duplicated logic with `system_supports_plotting()`


banesullivan:
Realizing now that moving some tests under `tests/plotting/` enables garbage collection validation. The following tests are now failing due to this:

```
ERROR tests/plotting/test_charts.py::test_barplot - AssertionError: Not all objects GCed:
ERROR tests/plotting/test_charts.py::test_axis_tick_offset - AssertionError: Not all objects GCed:
ERROR tests/plotting/test_cube_axes_actor.py::test_labels - AssertionError: Not all objects GCed:
ERROR tests/plotting/test_charts.py::test_axis_init - AssertionError: Not all objects GCed:
ERROR tests/plotting/test_axes.py::test_axes_actor_cone_res - AssertionError: Not all objects GCed:
ERROR tests/plotting/test_texture.py::test_activate_texture_none
ERROR tests/plotting/jupyter/test_ipyvtk.py::test_ipyvtk_warn - AssertionError: Not all objects GCed:
ERROR tests/plotting/jupyter/test_panel.py::test_save - AssertionError: Not all objects GCed:
```

I'm going to see if I can fix most of these. If not, I will disable that fixture for these tests specifically and let a follow up PR address the real issue

banesullivan:
@pyvista/developers, this is ready for review. Like I mentioned above, I'd like to merge this fairly quickly to not impact other work I'm doing and leave any substantial changes for follow up work.

# Proving the decoupling
To prove the value in this, check this out.

Apply the following diff to PyVista which raises a `RuntimeError` if any of `pyvista.plotting` is imported: 

```diff
diff --git a/pyvista/__init__.py b/pyvista/__init__.py
index 89b586a2..0b09ce66 100755
--- a/pyvista/__init__.py
+++ b/pyvista/__init__.py
@@ -12,25 +12,26 @@ from pyvista.core import *
 from pyvista.core.cell import _get_vtk_id_type
 from pyvista.core.utilities.observers import send_errors_to_logging
 from pyvista.core.wrappers import _wrappers
-from pyvista.errors import InvalidCameraError, RenderWindowUnavailable
+# from pyvista.errors import InvalidCameraError, RenderWindowUnavailable
 from pyvista.jupyter import set_jupyter_backend
-from pyvista.plotting import *
-from pyvista.plotting import _typing, _vtk
-from pyvista.plotting.utilities.sphinx_gallery import _get_sg_image_scraper
+# from pyvista.plotting import *
+# from pyvista.plotting import _typing, _vtk
+from pyvista.core import _typing_core as _typing
+# from pyvista.plotting.utilities.sphinx_gallery import _get_sg_image_scraper
 from pyvista.report import GPUInfo, Report, get_gpu_info, vtk_version_info
-from pyvista.themes import (
-    DefaultTheme as _GlobalTheme,
-    _rcParams,
-    _set_plot_theme_from_env,
-    load_theme,
-    set_plot_theme,
-)
-
-global_theme = _GlobalTheme()
-rcParams = _rcParams()  # raises DeprecationError when used
-
-# Set preferred plot theme
-_set_plot_theme_from_env()
+# from pyvista.themes import (
+#     DefaultTheme as _GlobalTheme,
+#     _rcParams,
+#     _set_plot_theme_from_env,
+#     load_theme,
+#     set_plot_theme,
+# )
+
+# global_theme = _GlobalTheme()
+# rcParams = _rcParams()  # raises DeprecationError when used
+
+# # Set preferred plot theme
+# _set_plot_theme_from_env()
 
 # get the int type from vtk
 ID_TYPE = _get_vtk_id_type()
diff --git a/pyvista/errors.py b/pyvista/errors.py
index 620bc941..73717284 100644
--- a/pyvista/errors.py
+++ b/pyvista/errors.py
@@ -16,4 +16,4 @@ from pyvista.core.errors import (
     PyVistaPipelineError,
     VTKVersionError,
 )
-from pyvista.plotting.errors import InvalidCameraError, RenderWindowUnavailable
+# from pyvista.plotting.errors import InvalidCameraError, RenderWindowUnavailable
diff --git a/pyvista/plotting/__init__.py b/pyvista/plotting/__init__.py
index 03eab8de..f5ebd605 100644
--- a/pyvista/plotting/__init__.py
+++ b/pyvista/plotting/__init__.py
@@ -1,5 +1,6 @@
 """Plotting routines."""
 # flake8: noqa: F401
+raise RuntimeError('No plotting allowed!')
 
 from pyvista import MAX_N_COLOR_BARS
 from pyvista._plot import plot
```

Then run the core API tests on this branch:

```bash
$ pytest -v tests/core  
...

====================================== short test summary info ======================================
FAILED tests/core/test_reader.py::test_pvdreader_no_part_group - FileNotFoundError: Path '/home/local/KHQ/bane.sullivan/.cache/pyvista_3/PVD/paraview/dualSphereA...
FAILED tests/core/test_utilities.py::test_report - RuntimeError: No plotting allowed!
ERROR tests/core/test_dataset.py::test_copy_metadata - AttributeError: module 'pyvista' has no attribute 'Texture'
============= 2 failed, 959 passed, 10 skipped, 1 xfailed, 1 error in 95.15s (0:01:35) ==============
```

```bash
$ pytest -v tests/filters
...
================================== 225 passed, 1 xfailed in 49.02s ==================================
```

The core API is almost nearly completely decoupled from GL-dependent imports and any plotting routines. Just a few minor things left with `Texture`

tests/plotting/conftest.py
```diff
@@ -7,11 +7,17 @@
...
 
 import pyvista
+from pyvista.plotting import system_supports_plotting
 
 # these are set here because we only need them for plotting tests
 pyvista.global_theme.load_theme(pyvista.themes._TestingTheme())
 pyvista.OFF_SCREEN = True
 
+# Configure skip_plotting marker
+pytest.mark.skip_plotting = pytest.mark.skipif(
```
adeak, line 17: This (slapping a new name onto `pytest.mark`) seems really hacky, but I'm not sure what else we could do... I suspect if we define a global name here it won't be available without explicitly importing it, which would defeat the purpose of having this in `conftest.py`.

tkoyama010 (approved):
LGTM

tests/core/test_grid.py
```diff
@@ -27,14 +26,6 @@ def test_volume(hexbeam):
     assert hexbeam.volume > 0.0
 
 
-@pytest.mark.skipif(not system_supports_plotting(), reason="Requires system to support plotting")
-def test_struct_example():
-    # create and plot structured grid
-    grid = examples.load_structured()
-    grid.plot(off_screen=True)  # basic plot
-    grid.plot_curvature(off_screen=True)
```
adeak, line 35: According to codecov there are 2 lines being uncovered due to the PR, `PointGrid.plot_curvature()` due to this. Consider restoring this as well in the plotting tests.

tests/core/test_utilities.py
```diff
@@ -24,16 +24,16 @@
...
-from pyvista.report import GPUInfo
 
-skip_no_plotting = pytest.mark.skipif(
-    not system_supports_plotting(), reason="Requires system to support plotting"
-)
+HAS_IMAGEIO = True
+try:
+    import imageio  # noqa: F401
+except ModuleNotFoundError:
+    HAS_IMAGEIO = False
```
adeak, line 36: This seems unused in this module.

tests/plotting/pytest_skip_plotting.py
```diff
@@ -0,0 +1,4 @@
+def pytest_configure(config):
+    config.addinivalue_line(
+        "markers", "skip_plotting: skip the test if it system does not support plotting"
```
adeak, line 3:
```suggestion
        "markers", "skip_plotting: skip the test if the system does not support plotting"
```

tests/plotting/test_collection.py
```diff
@@ -3,12 +3,18 @@
...
+import pytest
 import vtk
 
 import pyvista as pv
 from pyvista.core._vtk_core import vtk_to_numpy
 
 
+@pytest.fixture(autouse=True)
+def validate_gc(check_gc):
+    pass
```
adeak, line 15: Do I understand correctly that the 3 copies of this fixture just serve to trigger autouse on the non-autouse `check_gc` fixture? Might be worth a comment or docstring.

adeak (commented):
Partial review, will try to finish later. But I want to get some comments in now that we have an approval :)

tests/plotting/test_charts.py
```diff
@@ -658,6 +654,7 @@ def test_areaplot(chart_2d, area_plot):
     assert np.allclose(area_plot.y2, y2)
 
 
+@pytest.mark.noautofixt
```
adeak, line 657: This is not a built-in fixture. [At the source](https://stackoverflow.com/a/38763328/5067311) it's implemented with the help of
```py
@pytest.fixture(autouse=True)
def autofixt(request):
    if 'noautofixt' in request.keywords:
        return
    print("patching stuff")
```
Currently `pytest` warns
```
tests/plotting/test_charts.py:657
  /home/adeak/pyvista/tests/plotting/test_charts.py:657: PytestUnknownMarkWarning: Unknown pytest.mark.noautofixt - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.noautofixt
```

adeak (commented):
OK, done looking at the rest. Great work @banesullivan.

tests/plotting/test_charts.py
adeak: And this looks like it only skips the fixture named `autofixt`, rather than all autofixtures. So we'd have to do this `request` dance on whatever fixture(s) we want to be skipped.

tests/plotting/conftest.py
banesullivan: Ah, I mimicked this implementation from elsewhere. Reading up on pytest docs now and will do this properly per https://docs.pytest.org/en/7.1.x/example/markers.html#adding-a-custom-marker-from-a-plugin

tests/core/test_grid.py
banesullivan: Shoot, I really thought I re-added this test under `tests/plotting/`... thanks for catching

tests/plotting/test_charts.py
banesullivan: Ah yes, this is an artifact of when I was trying to disable the GC validation fixture for tests that are failing it. This was an accidental include.

tests/core/test_grid.py
adeak: There's _two_ implementations of `plot_curvatire()`.

banesullivan: Aha! Thank you 🤦🏻‍♂️

larsoner:
> To prove the value in this, check this out.

If you want to turn this into a unit test, this is what we do to make sure `import mne` doesn't import too much stuff in MNE-Python:

https://github.com/mne-tools/mne-python/blob/main/mne/tests/test_import_nesting.py#L8

It could help ensure nobody accidentally imports anything that takes forever, pulls in bare `vtk`, etc.

tests/plotting/test_collection.py
banesullivan: Yes. I've added a docstring to clarify this. These three modules were the original only three under `tests/plotting/` where these GC fixtures were autoused. We should come back to this later to enable the fixture to be used for the entire module, but this requires code fixes to get clean up references.

tests/core/test_composite.py
```diff
@@ -334,8 +334,8 @@ def test_multi_block_repr(ant, sphere, uniform, airplane):
     assert str(multi) is not None
 
 
-def test_multi_block_eq(ant, sphere, uniform, airplane, globe):
-    multi = multi_from_datasets(ant, sphere, uniform, airplane, globe)
```
annehaley, line 338: Should these tests still be performed on `globe` among the other texture-dependent tests? It's unclear whether testing `globe` here actually covered any unique cases (I'm not sure if the presence of a texture affects any of these), but I ask because this seems to be one of the few functional changes made to the tests.

annehaley (approved):
Looks good, since it's mostly moving stuff around. I only have one question (see below).

tests/core/test_composite.py
banesullivan: It *shouldn't* -- I'm hoping that codecov is accurately representing this too by not showing any coverage differences. The texture tracking functionalities should be fully captured by `test_copy_metadata` in `tests/core/test_dataset.py`

Additionally, I am deprecating the texture tracking in a follow up PR so if this does affect coverage, that will be deprecated in the next week

tests/plotting/test_collection.py
MatthewFlamm: Future looking comment, i.e. not to address in this PR, if the goal is to have `check_gc` apply to as many tests as possible, then it will be better to have `check_gc` be `autouse=True` and then provide a mechanism to override this.  This will require a slight rework of `check_gc` to expose a method by which to skip checking in a yielded object, similar to how we do in `validate_image_cache`.  The current state is good for this PR.

adeak (commented):
One last comment, and codecov (contrary to the earlier state) now claims that 8-9 lines are now uncovered, see https://app.codecov.io/gh/pyvista/pyvista/pull/4500/indirect-changes . They mostly or all seem to be VTK version related, so perhaps they are false positives? I can't look into this right now.

tests/core/test_composite.py
adeak: In an earlier state of the PR there were only 2 uncovered lines and those were accounted for by a specific test (discussed in a thread). So I'm fairly sure we're not losing coverage due to this change.

banesullivan:
The codecov changes apper to be due to specific jobs (for VTK <=9.2.2) failing to upload the coverage report: See here on previous CI run. I'm hoping this was a just a network issue with codecov and the next CI run will fix

- FAILED: Linux Unit Testing (3.8, 9.0.3): https://github.com/pyvista/pyvista/actions/runs/5212656466/jobs/9406562964#step:14:41
- FAILED: Linux Unit Testing (3.9, 9.1): https://github.com/pyvista/pyvista/actions/runs/5212656466/jobs/9406563091#step:14:41
- FAILED:  Linux Unit Testing (3.10, 9.2.2): https://github.com/pyvista/pyvista/actions/runs/5212656466/jobs/9406563253#step:14:41
- SUCCEEDED: Linux Unit Testing (3.11, latest): https://github.com/pyvista/pyvista/actions/runs/5212656466/jobs/9406563409#step:14:44

tkoyama010 (approved):

PR #4507, opened 2023-06-09
Deprecate pyvista.utilities

banesullivan:
Follow up to #4486 to properly deprecate the `pyvista.utilities` module and test that this API is still useable (non-breaking) but does issue deprecation warnings.

This whole thing feels hacky, I'm open to suggestions on other ways to do this


tests/namespace/test_utilities_namespace.py
```diff
@@ -0,0 +1,81 @@
+import importlib
+import pathlib
+
+import pytest
+
+from pyvista.core.errors import PyVistaDeprecationWarning
+
+namesapce_data = pathlib.Path(__file__).parent / 'namespace-utilities.txt'
+with open(namesapce_data) as f:
```
annehaley, line 9: `namesapce_data` -> `namespace_data`

tests/namespace/test_public_namespace.py
```diff
@@ -7,7 +7,7 @@
 # Use cleaned data to avoid things like `np`, `os`, etc
 # This prevents testing against things that are not intended
 # to be in the public namespace
-namesapce_data = pathlib.Path(__file__).parent / 'namespace-pre-refactor-clean.txt'
+namesapce_data = pathlib.Path(__file__).parent / 'namespace-top.txt'
 with open(namesapce_data) as f:
```
annehaley, line 11: same typo here

tests/namespace/test_utilities_namespace.py
```diff
@@ -0,0 +1,81 @@
...
+        'regression',
+        'sphinx_gallery',
+        'transformations',
+        'wrappers',
+        'xvfb',
+    ],
+)
+def test_utilities_modules(name):
+    # Smoke test to make sure same modules still exist
+    importlib.import_module(f'pyvista.utilities.{name}')
```
annehaley, line 48: Perhaps you can check that the warning is raised and parse the warning message for the proper path to the module?

pyvista/utilities/_getattr.py
```diff
@@ -0,0 +1,57 @@
...
+        pass
+
+    from pyvista.core.errors import PyVistaDeprecationWarning
+
+    for module in _MODULES:
+        feature, import_path = _try_import(module, name)
+        if feature is not None:
+            break
+    else:
+        raise AttributeError(
```
annehaley, line 46: Codecov is warning that there isn't a test that covers this. You could add a test that attempts to import something like `pyvista.utilities.invalid_module_name`

annehaley (commented):
Most of this looks good, why do you feel it's hacky? I think your loop of `_try_import`s is a better design than the previous implementation.

tests/namespace/test_utilities_namespace.py
adeak: FYI you can leave specific code suggestions on github :) https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-comments-to-a-pull-request

where the code suggestion icon is

tests/namespace/test_public_namespace.py
tkoyama010:
```suggestion
namespace_data = pathlib.Path(__file__).parent / 'namespace-top.txt'
with open(namespace_data) as f:
```

pyvista/utilities/_getattr.py
banesullivan: Added test. Will see if codecov come back happy

tests/namespace/test_utilities_namespace.py
```diff
@@ -0,0 +1,87 @@
...
+        from pyvista.utilities.errors import GPUInfo  # noqa: F401
+        from pyvista.utilities.geometric_objects import Arrow, Cylinder, PlatonicSolid  # noqa: F401
+        from pyvista.utilities.helpers import vtk_id_list_to_array  # noqa: F401
+        from pyvista.utilities.sphinx_gallery import _get_sg_image_scraper  # noqa: F401
+        from pyvista.utilities.xvfb import start_xvfb  # noqa: F401
+
+
+def test_failure_to_find():
+    module = importlib.import_module('pyvista.utilities')
+    with pytest.raises(AttributeError, match='Module `pyvista.utilities` has been deprecated.'):
```
adeak, line 86:
```suggestion
    with pytest.raises(AttributeError, match=r'Module `pyvista.utilities` has been deprecated\.'):
```

tests/namespace/test_utilities_namespace.py
```diff
@@ -0,0 +1,87 @@
...
+with open(namespace_data) as f:
+    namespace = f.read().splitlines()
+    # ignore commented data
+    namespace = [n.split(', ')[0] for n in namespace if not n.startswith('#')]
+
+
+@pytest.mark.parametrize('name', namespace)
+def test_utilities_namespace(name):
+    with pytest.warns(PyVistaDeprecationWarning):
+        utilities = importlib.import_module('pyvista.utilities')
```
adeak, line 18: Do we need importlib in the tests in this module? Why not just
```suggestion
        from pyvista import utilities
```
here, and similarly later?

pyvista/utilities/_getattr.py
```diff
@@ -0,0 +1,57 @@
+import importlib
+import inspect
+import pathlib
+import warnings
+
+# Have to set __path__ to this file's directory so that
+# `import pyvista.utilities` imports this directory
+__path__ = [str(pathlib.Path(__file__).parent)]
```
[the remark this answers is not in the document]
banesullivan: I can't say I truly know what I'm talking about here, but let me try...

The best summary of `__path__` I can find is that the `__path__` attribute of a Python package is used by the import system when trying to import modules of subpackages. So when we run `from pyvista.utilities import convert_array`, the import system tries to get the `__path__` attribute of the `utilities` subpackge. Problem is I override `__getattr__` and this causes it to pull `__path__` from the first place that has it: `pyvista.core.utilities`. This would be fine if we were migrating everything to `pyvista.core.utilities`, but we arent and many of the module names have changed. I haven't *actully* run into issues with this yet, and included this `__path__` attribute here to ensure the import system is pointing to `pyvista.utilities` just in case `from pyvista.utilities.features import <feature>` is used later to ensure it resolves.

pyvista/utilities/_getattr.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Places to look for the utility
+_MODULES = [
+    'pyvista.core.utilities',
+    'pyvista.core.errors',
+    'pyvista.core',
+    'pyvista.plotting.utilities',
+    'pyvista.plotting.errors',
+    'pyvista.plotting.texture',
```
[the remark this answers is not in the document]
banesullivan: I set the order intentionally because everything that is available under `pyvista.core.utilities` and `pyvista.core.errors` is also available under `pyvista.core` (due to `*` imports and wanting to keep the `pyvista.*` namespace the same as previous releases). Having this specific order will provide better deprecaton notice and tell users to import from `pyvista.core.utilities` specifically as opposed to just `pyvista.core` for everything (same goes for `pyvista.plotting`)

In the way I currently have it configured, `pyvista.core` is a fallback if something moved that isn't in `pyvista.core.utilities` or `pyvista.core.errors` which there are several things.

pyvista/utilities/_getattr.py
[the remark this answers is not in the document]
banesullivan: Additionally this is to avoid this warning:

```py
>>> import pyvista.utilities
/home/local/KHQ/bane.sullivan/Software/pyvista/pyvista/pyvista/utilities/_getattr.py:53: PyVistaDeprecationWarning: The `pyvista.utilities` module has been deprecated. `__path__` is now imported as: `from pyvista.core.utilities import __path__`.
  warnings.warn(
```

pyvista/utilities/_getattr.py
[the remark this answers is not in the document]
adeak: Ah, good point, I didn't realise that the order mattered here. Thanks.

pyvista/utilities/_getattr.py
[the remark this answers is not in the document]
banesullivan: This may all be totally wrong. But having `__path__` set here at least prevents an obscure warning for users

tests/namespace/test_utilities_namespace.py
banesullivan: Reverted as this caused errors

tests/namespace/test_utilities_namespace.py
banesullivan: Because of all the `*` imports, `from pyvista import utilities` is NOT equal to `import pyvista.utilities`. `from pyvista import utilities` gives us `pyvista.plotting.utilities` now.

re `importlib` vs `from ... import`, it doesn't matter either way

tests/namespace/test_utilities_namespace.py
adeak: That would mean that the error message is inexact, and the solution would be to fix the error message in `match`. I suggested these changes exactly to make it clear that `match` is a regex, and it shouldn't match accidentally.

tests/namespace/test_utilities_namespace.py
adeak: Ah, I see. Sort of. I don't think star imports matter here. It's just that `import pyvista.utilities` forces a module import, whereas indeed the `from` import looks up the name in `pyvista/__init__.py`'s namespace.

So why not
```suggestion
        import pyvista.utilities as utilities
```
It's really the dynamic import that I'd like to avoid if possible.

tests/namespace/test_utilities_namespace.py
adeak: Yup, the exact error here is
```
AttributeError: Module `pyvista.utilities` has been deprecated and we could not automatically find `this_does_not_exist`. This feature has moved.
```
The `.` in the regex at the end matches the whitespace. So perhaps
```suggestion
    with pytest.raises(AttributeError, match=r'Module `pyvista\.utilities` has been deprecated and we could not automatically find'):
```

pyvista/utilities/_getattr.py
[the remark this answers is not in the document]
adeak: Hmm, thanks, I'll have to think about this. My gut feeling is that the real issue is that in `_getattr.py` when we try to return from `globals()` first, that is always the globals of `_getattr.py`. Not sure if that's intended, or what exactly we're even trying to do with that line to be honest.

I also don't understand why `__path__` is being looked up in `__getattr__()`, because [according to the docs](https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access)

> If an attribute is not found on a module object through the normal lookup, i.e. [`object.__getattribute__()`](https://docs.python.org/3/reference/datamodel.html#object.__getattribute__), then `__getattr__` is searched in the module `__dict__` before raising an [`AttributeError`](https://docs.python.org/3/library/exceptions.html#AttributeError). If found, it is called with the attribute name and the result is returned.

Because this sounds like existing names don't get passed to `__getattr__()`. Unless there's some weirdness about when `__path__` is defined...

Aaaah, I think I'm starting to get it:
```py
>>> import pyvista.core.utilities.features
>>> pyvista.core.utilities.features.__path__
AttributeError: module 'pyvista.core.utilities.features' has no attribute '__path__'
```

Modules don't have a `__path__`, packages do. So the issue might be that we're using a module's `__getattr__()` to drop in for a package's `__getattr__()`. At least that's my best mental model for now. And so `_getattr` doesn't have a `__path__` that we're trying to look up in `globals()`, which is the `globals()` of `_getattr`. Although this still doesn't explain why `pyvista.utilities` doesn't have a `__path__` of its own, which leads to `__getattr__()` being called... I wonder if this has to do with some details of how modules are imported in Python, partially imported packages not having `__path__` or something weird like that. But I can print `__path__` in `pyvista/utilities/__init__.py` before we do `from _getattr import __getattr__`, which seems to rule out this possibility...

Anyway, now we also have (for a deprecated placeholder module)
```py
>>> pyvista.utilities.helpers.__path__
['/home/lou/Documents/stackoverflow/python/pyvista_dev/pyvista/pyvista/utilities']
```
which is a lie, because `pyvista.utilities.helpers` is a module, not a package, and should not have a `__path__`. Also, [the docs says](https://docs.python.org/3/reference/import.html#path__)
> Non-package modules should not have a `__path__` attribute.

but these are just placeholders so it's not that much of an issue. Hopefully no tooling comes back to bite us in the butt.

---

OK, the only alternative I can think of is using a `__getattr__()` _factory_ that is passed the current module globals in each module that uses it. In which case `pyvista.utilities` will naturally have a `globals()` that contains `__path__` and we won't get the spurious warning. I'm not sure this is overall less hacky, just a possible option (unless it has a flaw of its own).

But if we keep the current `__path__` hack I'd like to have a more descriptive comment above it. I'm still not sure what
> `import pyvista.utilities` imports this directory

means to say, because with or without the `__path__` hack we seem to end up with the same result if we do `import pyvista.utilities`.

tests/namespace/test_utilities_namespace.py
```diff
@@ -0,0 +1,87 @@
...
+with open(namespace_data) as f:
+    namespace = f.read().splitlines()
+    # ignore commented data
+    namespace = [n.split(', ')[0] for n in namespace if not n.startswith('#')]
+
+
+@pytest.mark.parametrize('name', namespace)
+def test_utilities_namespace(name):
+    with pytest.warns(PyVistaDeprecationWarning):
+        import pyvista.utilities as utilities
```
adeak, line 18: To make `black` happy:
```suggestion
        import pyvista.utilities as utilities

```

banesullivan: May need to make flake8 happy with a `noqa: F401` too

adeak: `pre-commit` is green. The import is not unused because we use it on the next line :)

pyvista/utilities/_getattr.py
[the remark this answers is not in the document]
banesullivan: Awesome investigative work, @adeak! This is connecting a lot of the dots in my head now... I'm going to continue to experiment with this and see if I can get it right by using the factory approach you outline

tests/namespace/test_utilities_namespace.py
banesullivan: 🤦🏻‍♂️ I wish it would show the following lines like it does the previous lines in these comment views on the "Conversation" page for this reason

pyvista/utilities/_getattr.py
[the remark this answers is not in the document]
banesullivan: The factory approach seems to work really well. Pushing changes in a moment but:

```py

>>> import pyvista.utilities
>>> pyvista.utilities.__path__
['/home/local/KHQ/bane.sullivan/Software/pyvista/pyvista/pyvista/utilities']
>>> pyvista.utilities.arrays.__path__
AttributeError: module 'pyvista.core.utilities.arrays' has no attribute '__path__'
# this attribute error is expected
>>> pyvista.utilities.convert_array
/home/local/KHQ/bane.sullivan/Software/pyvista/pyvista/pyvista/utilities/_getattr.py:52: PyVistaDeprecationWarning: The `pyvista.utilities` module has been deprecated. `convert_array` is now imported as: `from pyvista.core.utilities import convert_array`.
  warnings.warn(
<function convert_array at 0x7f1e17fae310>
>>> from pyvista.utilities import arrays
/home/local/KHQ/bane.sullivan/Software/pyvista/pyvista/pyvista/utilities/_getattr.py:52: PyVistaDeprecationWarning: The `pyvista.utilities` module has been deprecated. `arrays` is now imported as: `from pyvista.core.utilities import arrays`.
  warnings.warn(
>>> arrays
<module 'pyvista.core.utilities.arrays' from '/home/local/KHQ/bane.sullivan/Software/pyvista/pyvista/pyvista/core/utilities/arrays.py'>
>>> 
```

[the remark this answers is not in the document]
banesullivan: Crap, the obscure warning about `__path__` is happening with this:

```py
>>> from pyvista.utilities.errors import GPUInfo
/home/local/KHQ/bane.sullivan/Software/pyvista/pyvista/pyvista/utilities/_getattr.py:52: PyVistaDeprecationWarning: The `pyvista.utilities` module has been deprecated. `__path__` is now imported as: `from pyvista.core.utilities import __path__`.
  warnings.warn(
/home/local/KHQ/bane.sullivan/Software/pyvista/pyvista/pyvista/utilities/_getattr.py:52: PyVistaDeprecationWarning: The `pyvista.utilities` module has been deprecated. `GPUInfo` is now imported as: `from pyvista.report import GPUInfo`.
  warnings.warn(
>>> 
```

I think we'll just have to deal with this and leave it as it stands now as it's the safest solution IMO (not setting `__path__` and letting users see that obscure import warning in some scenarios)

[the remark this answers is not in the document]
akaszynski: Addressed this in [3498b28](https://github.com/pyvista/pyvista/pull/4507/commits/3498b287b3d9c320a17eec658eea488fe70ede51).

This cleans up the warnings to:

```py
>>> from pyvista.utilities.errors import GPUInfo
from pyvista.utilities.errors import GPUInfo
~/source/pyvista/pyvista/utilities/_getattr.py:52: PyVistaDeprecationWarning: The `pyvista.utilities` module has been deprecated. `GPUInfo` is now imported as: `from pyvista.report import GPUInfo`.
  warnings.warn(
```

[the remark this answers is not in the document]
adeak: Uuuh the whole point of the factory was to use the globals :) If you remove the globals you can just use the same `__getattr__()`. And the `_globals` were "unneeded" because instead of `try: return self._globals[name]` you had `try: self._globals[name]`. Baby with the bathwater and all that.

The current state is needlessly overengineered. Either keep a single `__getattr__()` with a manual skip of `__path__` (and other dunders, and even other leading-underscore stuff, probably; I made that suggestion the first time this came up, in another thread). Or use a working factory that actually uses the globals. I don't have a strong preference for either (mildly prefer the factory), but the current state is neither here nor there.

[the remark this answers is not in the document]
adeak: For the sake of unambiguity, what I had in mind is to turn
```py
def __getattr__(name):
    try:
        return globals()[name]
    except KeyError:
        pass

    from pyvista.core.errors import PyVistaDeprecationWarning

    for module in _MODULES:
        feature, import_path = _try_import(module, name)
        if feature is not None:
            break
    else:
        raise AttributeError(
banesullivan marked this conversation as resolved.
            f'Module `pyvista.utilities` has been deprecated and we could not automatically find `{name}`. This feature has moved.'
        ) from None

    message = f'The `pyvista.utilities` module has been deprecated. `{name}` is now imported as: `{import_path}`.'

    warnings.warn(
        message,
        PyVistaDeprecationWarning,
    )

    return feature
```
into
```py
def getattr_factory(globs):
    def __getattr__(name):
        try:
            return globs[name]
        except KeyError:
            pass

        from pyvista.core.errors import PyVistaDeprecationWarning

        for module in _MODULES:
            feature, import_path = _try_import(module, name)
            if feature is not None:
                break
        else:
            raise AttributeError(
    banesullivan marked this conversation as resolved.
                f'Module `pyvista.utilities` has been deprecated and we could not automatically find `{name}`. This feature has moved.'
            ) from None

        message = f'The `pyvista.utilities` module has been deprecated. `{name}` is now imported as: `{import_path}`.'

        warnings.warn(
            message,
            PyVistaDeprecationWarning,
        )

        return feature
    return __getattr__
```
(whether this is a nested function or a callable class is unimportant).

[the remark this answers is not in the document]
akaszynski: @adeak, went with the factory approach in [ab459af](https://github.com/pyvista/pyvista/pull/4507/commits/ab459af2b24135b722c9d4500d25edee2de267c8). Agree with the overengineering, took me a bit to catch up through all the comments.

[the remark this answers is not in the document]
adeak: Thanks @akaszynski. I'm on mobile so I can't check, but I think that's still wrong. You're still using the same function, you could just define that in `utilities/__init__.py` to the same effect. E.g. check `pyvista.utilities.helpers.__path__`, I think you'll see `utilities`' path there.

You need to pass the factory the current globals in each module. That's the point of the factory: all distinct `__getattr__()` instances using different globals.

pyvista/utilities/_getattr.py
```diff
@@ -0,0 +1,81 @@
...
+    except AttributeError:
+        return None, None
+    return feature, import_path
+
+
+def _getattr_factory(globs):
+    """Create and return a custom getattr method.
+
+    The created getattr method tries to dynamically fetch an attribute ``name``
+    from the given dictionary globs, expected to be ``globals``.
```
adeak, line 37:
```suggestion
    from the given dictionary ``globs``, expected to be ``globals()``.
```

pyvista/utilities/_getattr.py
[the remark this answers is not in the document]
akaszynski: > you could just define that in utilities/__init__.py to the same effect
True, and I actually considered that. I was lazy with the copy/pasta when using the factory.

The issue with using `globals` from each module is that the modules don't have a `__path__` attribute because they're not considered a package. The only way to ensure that we avoid the odd `__path__` warning is to use the globals from `utilities/__init__.py`.

If you agree with this approach, we can get rid of `_get_attr.py` altogether and move everything into the `__init__.py`.

BTW, local testing on a closed source library that makes extensive use of `pyvista.utilities` works just fine using this approach with all the right warnings in the right places.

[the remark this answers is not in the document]
adeak: > The issue with using `globals` from each module is that the modules don't have a `__path__` attribute because they're not considered a package.

But that's what I'd expect: not returning `__path__` for the "leaf" modules. Does that somehow break with all this proxying?

> If you agree with this approach, we can get rid of `_get_attr.py` altogether and move everything into the `__init__.py`.

Sure, like I said, I'm mostly OK with just hacking the getattt to ignore dunders or something. This is just for backwards compatibility after all. I just expected the factory approach (not making assumptions about preexisting names) to be purer.

But yes, if we keep the current setup, I agree we should just define the one `__getattr__()` _in situ_ (and use the same imports as now).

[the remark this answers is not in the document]
akaszynski: I created a minimal working example to test to see what's going on here:

```
testpackage/
    __init__.py
    testmodule.py
```

Where `testmodule.py` contains:
```
def __getattr__(name):
    print(f"__getattr__ called with {name}")
    raise AttributeError(f"attribute {name} not found")


def testfunction():
    print("This is a function in testmodule")
```

By simply importing and running `testfunction()`:

```
from testpackage.testmodule import testfunction
testfunction()
```

It seems that it indeed looks for the `__path__` attribute:
```
$ python tmp.py
__getattr__ called with __path__
This is a function in testmodule
```

Interestingly enough, it doesn't care that `None` is returned when requesting the `__path__`.

It's possible that Python is checking to see if a module is a package by checking for the `__path__` attribute. This seems to be the case according to [The import system](https://docs.python.org/3/reference/import.html#packages):

> It’s important to keep in mind that all packages are modules, but not all modules are packages. Or put another way, packages are just a special kind of module. Specifically, any module that contains a __path__ attribute is considered a package.

---

Since this is just for deprecation, I think we should just keep it simple and move everything from `_getattr.py` into the `__init__.py`. We might consider treating the `__path__` attribute differently since Python might think that the leaf modules are packages since they have a `__path__` attribute, but in my (albeit limited) testing, Python doesn't seem to care either way.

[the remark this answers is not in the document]
adeak: Thanks for loooking into that @akaszynski! 

> Since this is just for deprecation, I think we should just keep it simple and move everything from `_getattr.py` into the `__init__.py`.

I concur :)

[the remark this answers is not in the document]
adeak: Aaah, I see. The attribute error on `__path__` is caught when you import. So that's how we could keep telling Python that the proxy is not a package. But our lookup strategy finds a `__path__` from elsewhere, and that's what makes this weird.

So the getattr factory would only work if we also passed it the exact new module that the old one corresponds to, and looked up the names in the `__dict__` of the real module. At that poibt we could just star import everything from the new module into the old proxy one. (And we probably don't have a bijection between old and new modules).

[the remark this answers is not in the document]
banesullivan: After these changes, should we modify the tests to make sure the `pytest.warns()` catch the specific module being imported to make sure warnings are not issued for `__path__` too?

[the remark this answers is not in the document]
akaszynski: > At that point we could just star import everything from the new module into the old proxy one.

Strangely enough, `__getattr__` doesn't seem to be called here except for `__path__`

```py
"""Deprecated module."""
from ._getattr import _getattr_factory

from pyvista.core.utilities.cells import *


def __getattr__(name):
    print(name)

__getattr__ = __getattr__()
```

As clean as it would be, I don't think a star import would work in our case.

Since it's just for deprecation, I'm going ahead with just using a simplified approach in [5690145](https://github.com/pyvista/pyvista/pull/4507/commits/56901459f139abe70b027f05e57ed5ed338b65fb) that just overrides `__getattr__` rather than using a factory.

pyvista/utilities/__init__.py
```diff
@@ -1,38 +1,70 @@
-"""Deprecated utilities module."""
+"""Deprecated utilities subpackage."""
 import importlib
 import inspect
 import warnings
 
-from pyvista.core.errors import DeprecationError, PyVistaDeprecationWarning
+# Places to look for the utility
+_MODULES = [
+    'pyvista.core.utilities.cells',
```
banesullivan, line 8: @akaszynski why is `pyvista.core.utilities.cells` added here? IMO, we should not include this and encourage users to use `from pyvista.core.utilities import <feature>` instead of going down into individual modules in the utilities subpackage.

banesullivan: If something in the `cells.py` module is missing at the top level `pyvista.core.utilities` package we should make sure to include it in `pyvista/core/utilities/__init__.py`

banesullivan:
```suggestion
```

akaszynski: I added this as one of the imports in a closed source project of mine was raising an `ImportError`, specifically `numpy_to_idarr`.

This was the import warning I was hoping to get:
```
/home/alex/source/pyvista/pyvista/utilities/_getattr.py:74: PyVistaDeprecationWarning: The `pyvista.utilities` module has been deprecated. `numpy_to_idarr` is now imported as: `from pyvista.core.utilities.cells import numpy_to_idarr`.
```

banesullivan: IMO, this should be `from pyvista.core.utilities import numpy_to_idarr` to ease the pain if we ever re-organize these modules in `pyvista.core.utilities`

pyvista/utilities/cell_type_helper.py
```diff
@@ -0,0 +1,2 @@
+"""Deprecated module."""
+from . import __getattr__  # noqa: F401  # noqa: F401
```
adeak, line 2:
```suggestion
from . import __getattr__  # noqa: F401
```

pyvista/utilities/__init__.py
```diff
@@ -1,38 +1,69 @@
...
+        import_path = f'from {module} import {name}'
+    except AttributeError:
+        return None, None
+    return feature, import_path
 
 
 def __getattr__(name):
-    utils = importlib.import_module('pyvista.core.utilities')
+    """Fetch an attribute ``name`` from ``globals()`` and warn if it's from a deprecated module.
+
```
adeak, line 33: In case you agree that this helps (so that nobody (like me) thinks later "surely we don't need `globals()` in there!"):
```suggestion

    Note that ``__getattr__()`` only gets called when ``name`` is missing
    from the module's globals. The trick is that we want to import this
    function into other deprecated modules, and we want to carry this
    subpackage's globals along to prevent some spurious warnings.

```
Phrasing open for discussion, of course.

adeak (approved):
Thanks for bearing with me @banesullivan @akaszynski. Looks great to me now.

akaszynski:
Always appreciate your attention to detail @adeak!

PR #4514, opened 2023-06-09
Check garbage collection on all plotting tests, allow specific skipping

MatthewFlamm:
### Overview

After #4500, `check_gc` fixture was only used in three test modules. It would be better to have it check all of the plotting tests, with optional skips.  This PR adds in this functionality by setting `check_gc` to be `autouse=True` in the `tests/plotting/conftest.py` and yielding an object which allows skipping the test.  A new fixture `skip_check_gc` can be used to skip specific tests (or modules) if warranted.

This PR is not aiming to fix any of the gc issues that might be uncovered. But specific uses of `skip_check_gc` will be an easier indication for bug fixing later.

### Details

First, this PR will push the changes that will expose failing tests, then skips will be added.

### DIFF
--- a/tests/plotting/conftest.py
+++ b/tests/plotting/conftest.py
@@ -36,21 +36,30 @@ def _is_vtk(obj):
 
 
 @pytest.fixture()
+def skip_check_gc(check_gc):
+    """Skip check_gc fixture."""
+    check_gc.skip = True
+
+
+@pytest.fixture(autouse=True)
 def check_gc(request):
     """Ensure that all VTK objects are garbage-collected by Python."""
     gc.collect()
     before = {id(o) for o in gc.get_objects() if _is_vtk(o)}
-    yield
 
-    # Do not check for collection if the test session failed. Tests that fail
-    # also fail to cleanup their resources and this makes reading the unit test
-    # output more difficult.
-    #
-    # This applies to the entire session, so it's going to be the most useful
-    # when debugging tests with `pytest -x`
-    pyvista.close_all()
-    if request.session.testsfailed:
+    class GcHandler:
+        def __init__(self) -> None:
+            # if set to True, will entirely skip checking in this fixture
+            self.skip = False
+
+    gc_handler = GcHandler()
+
+    yield gc_handler
+
+    if gc_handler.skip:
         return
+
+    pyvista.close_all()
 
     gc.collect()
     after = [o for o in gc.get_objects() if _is_vtk(o) and id(o) not in before]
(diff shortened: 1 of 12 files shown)


banesullivan:
> After https://github.com/pyvista/pyvista/pull/4500, check_gc fixture was only used in three test modules.

I want to clarify that it was only used in those same three modules before that PR too. Just now there are more testing modules in that directory which didn't all pass the GC validation.

Thanks for implementing this! I couldn't find a clean way to make the future opt-out after setting `autouse=True`

tests/plotting/conftest.py
```diff
@@ -36,22 +36,31 @@ def _is_vtk(obj):
...
-    yield
-
-    # Do not check for collection if the test session failed. Tests that fail
-    # also fail to cleanup their resources and this makes reading the unit test
-    # output more difficult.
-    #
-    # This applies to the entire session, so it's going to be the most useful
-    # when debugging tests with `pytest -x`
-    pyvista.close_all()
-    if request.session.testsfailed:
```
MatthewFlamm, line 52: This removed block to skip gc test using when any previous test failed (using `request.session.testsfailed` was added in #2941.  I needed to turn this off to easily find all the failing tests here, and did not notice any confusing output in doing so.  @akaszynski implemented this and maybe can shed some light on whether it should still be included.

MatthewFlamm: It seems that it my local pytest invocation is somehow different than the CI invocation and it does add confusing output for gc failures on the CI.  We might need to turn it back on.  Will do more digging.

MatthewFlamm: I was misinterpreting my local vs. CI pytest results.  When looking closer, I did not see anything more complicated after removing this if block for the failures.  Note that in #4500 only a small number of tests were identified as failing (see https://github.com/pyvista/pyvista/pull/4500#issuecomment-1581261637), but in reality a large number fail (mostly from within the tests/plotting/jupyter folder, tests/plotting/test_axes, and tests/plotting/test_charts).  The discrepancy is due to this if block.

banesullivan (approved):
Awesome! All looks good to me

MatthewFlamm:
Thanks, even though I'm wary of skipping `check_gc` when any other test has failed, I'd like to wait for comments on this before merging.

adeak (approved):
LGTM. I think it's the right call to always fail unless explicitly skipped, and following up with individual prodding of gc-failing tests. Thanks @MatthewFlamm.

PR #61, opened 2019-01-22
Add a textures property to keep track of associated textures

banesullivan:
@akaszynski: Could you approve this (as its adding new properties that doesn't call back to the VTK object)? I need this new property for the [`omfvtk`](https://github.com/OpenGeoVis/omfvtk) project.

It essentially helps a user keep track of `vtkTextures` that could be associated with that dataset. I need it in `omfvtk` because the data files contain both the dataset and the texture. Having this would enable me to pass textures along with the datasets. 

This also allows for easier texture plotting when a texture is associated with a dataset in the new `textures` dictionary property:

```py
>>> from vtki import examples
>>> globe = examples.load_globe()
>>> globe.textures
{'2k_earth_daymap': (vtkRenderingOpenGL2Python.vtkOpenGLTexture)0x11bf878e8}

>>> globe.plot(texture=True)

```

download

PR #4515, opened 2023-06-09
Deprecate Texture tracking on DataSet

banesullivan:
In order to move towards #4445 and fully decouple the core API from GL-dependent imports/modules in VTK, we have to migrate all `vtkTexture` support to the `pyvista.plotting` module. This was mostly done in #4486 but the remaining texture tracking feature on `DataSet` remains problematic.

The texture tracking feature was introduced (by yours truly) in #61 as a convenience feature. There is no analog to this feature in VTK's data model and this feature has been a quirky PyVista-ism since its introduction. 

With the goal to move PyVista's core data model API closer to VTK and to fully decouple it from GL-dependent imports, we are deprecating then removing support for this feature and requiring users to be explicit with the usage of textures during rendering moving forward.

(running a full doc build locally and will finalize anything needed)


pyvista/examples/examples.py
```diff
@@ -193,11 +193,12 @@ def load_globe():
...
     >>> from pyvista import examples
     >>> dataset = examples.load_globe()
-    >>> dataset.plot()
+    >>> texture = examples.load_globe_texture()
+    >>> dataset.plot(texture=texture)
 
     """
     globe = pyvista.PolyData(globefile)
-    globe.textures['2k_earth_daymap'] = load_globe_texture()
+    globe._textures['2k_earth_daymap'] = load_globe_texture()
```
MatthewFlamm, line 201: The benefit of this switch is to avoid a deprecation warning?  If so, should there be a pointer for this somewhere (maybe in DataSet.textures there can be a source code comment along the lines of `# use _textures internally to avoid PyVistaDeprecationWarning`?)  Not really sure if it is value add, but it confused me for a few minutes anyway.

MatthewFlamm (commented):
Maybe a naive reading, but it looks like the typing is the only thing that requires importing from `pyvista.plotting`.  If this is correct AND this piece of functionality is useful, this change might not be neccessary.  For example, careful use of `typing.TYPE_CHECKING` could be used to import `Texture` only when type checking.

I am not a user of Textures, so I can't comment on how useful this tracking is in the first place.

pyvista/examples/examples.py
banesullivan: Yes, this switch is to avoid the deprecation warning while loading example data. We could, but no new code should use this API now that it is deprecated so I don't think there would be any added value to the note.

banesullivan:
> Maybe a naive reading, but it looks like the typing is the only thing that requires importing from `pyvista.plotting`.

I think this is technically right

> If this is correct AND this piece of functionality is useful, this change might not be neccessary. For example, careful use of `typing.TYPE_CHECKING` could be used to import `Texture` only when type checking.

However, there are two long-term goals at play here:

1. Move PyVista towards the VTK API and remove "quirky PyVista-isms" (my term 😄 ) that do not have an analog in VTK. This is in an effort to make PyVista more familiar to the broader VTK community.
2. Separate the `pyvista.core` subpackage into a new, standalone `pyvista-core` package that will the primary data model API to VTK in Python. (`pyvista.core` would then be a namespace package to `pyvista-core`)

With these goals, we really need to decouple `Texture`s and any GL-dependent code from the core API

> I am not a user of Textures, so I can't comment on how useful this tracking is in the first place.

It's a nice convenience feature but, IMO as the person who implemented it and had clear use cases, it is not useful enough to warrant keeping when it's a clear divergence from the VTK API

pyvista/examples/planets.py
```diff
@@ -72,20 +64,22 @@ def load_sun(radius=1.0, lat_resolution=50, lon_resolution=100):  # pragma: no c
     --------
     >>> import pyvista as pv
     >>> from pyvista import examples
+    >>> mesh = examples.planets.load_sun()
+    >>> texture = examples.planets.download_sun_surface(texture=True)
```
adeak, line 68: The disparity between planet and texture function name is not great, but I guess inevitable. And in any case users trying to use these helpers will look at the examples and see the relevant texture function.

pyvista/examples/planets.py
```diff
@@ -192,22 +190,24 @@ def load_venus(radius=1.0, lat_resolution=50, lon_resolution=100):  # pragma: no
...
     """
     sphere = _sphere_with_texture_map(
         radius=radius, lat_resolution=lat_resolution, lon_resolution=lon_resolution
     )
     surface = download_venus_surface(atmosphere=False, texture=True)
-    sphere.textures["surface"] = surface
+    sphere._textures["surface"] = surface
     atmosphere = download_venus_surface(texture=True)
-    sphere.textures["atmosphere"] = atmosphere
+    sphere._textures["atmosphere"] = atmosphere
```
adeak, line 210: Is there a way we could drop a hint about the atmosphere toggle for the texture download function? Maybe not, because we can only use either texture for plotting, and it's not like it was evident that we had two textures here, based on the docstring example alone.

pyvista/plotting/plotter.py
```diff
@@ -2918,13 +2921,10 @@ def add_mesh(
...
 
-        texture : vtk.vtkTexture or np.ndarray or bool or str, optional
+        texture : vtk.vtkTexture or np.ndarray, optional
             A texture to apply if the input mesh has texture
             coordinates.  This will not work with MultiBlock
-            datasets. If set to ``True``, the first available texture
-            on the object will be used. If a string name is given, it
-            will pull a texture with that name associated to the input
-            mesh.
+            datasets.
```
adeak, line 2927: I'm surprised this didn't work for `MultiBlock` in case the texture was encoded in the mesh. Was it really unsupported? Because if it worked after all then this will be a breaking change once we remove the feature.

pyvista/plotting/plotter.py
```diff
@@ -3337,7 +3337,7 @@ def add_mesh(
         # Try to plot something if no preference given
         if scalars is None and color is None and texture is None:
             # Prefer texture first
-            if len(list(mesh.textures.keys())) > 0:
+            if len(list(mesh._textures.keys())) > 0:
```
adeak, line 3340: Preexisting logic, but
```suggestion
            if len(mesh._textures) > 0:
```
or even
```suggestion
            if mesh._textures:
```

tests/examples/test_download_files.py
```diff
@@ -988,65 +989,76 @@ def test_coil_magnetic_field():
...
 
 
 def test_load_venus():
     mesh = examples.planets.load_venus()
     assert mesh.n_cells
-    assert mesh.textures["surface"]
-    assert mesh.textures["atmosphere"]
+    with pytest.warns(PyVistaDeprecationWarning):
+        assert mesh.textures["surface"]
+        assert mesh.textures["atmosphere"]
```
adeak, line 1015: Another place where we might want to break this up into two `with` blocks. Maybe not, because the texture lookup logic is the same on both lines.

pyvista/core/filters/poly_data.py
```diff
@@ -2054,7 +2054,9 @@ def geodesic(
         output["vtkOriginalPointIds"] = original_ids
 
         # Do not copy textures from input
-        output.clear_textures()
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
adeak, line 2058: I recommend only ignoring our own deprecation warning here. Better not have false negatives.

pyvista/core/dataset.py
```diff
@@ -1400,23 +1403,25 @@ def copy_meta_from(self, ido: DataSet, deep: bool = True):
             Deep or shallow copy.
 
         """
-        self.clear_textures()
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
adeak, line 1407: Same note about ignoring warning subclass.

dcbr:
I don't use textures that often, so maybe this is not a common use case; but I once remember playing around with a "dynamic texture" on a fixed mesh. How would one achieve that after these changes? It seems to me the mesh would have to be completely removed and redrawn with the updated texture?

banesullivan:
> playing around with a "dynamic texture" on a fixed mesh. How would one achieve that after these changes?

I'm not sure I know what you are referring to... 

> It seems to me the mesh would have to be completely removed and redrawn with the updated texture?

This isn't the case. The feature being removed here was simply a convienance method to let PyVista objects know what textures were associated with them so that users wouldn't have to pass texture objects around in `add_mesh(..., texture=...)`

There isn't any difference in visualization capabilities from these changes

pyvista/examples/planets.py
banesullivan: I'm not a huge fan of this `planets` module overall. IMO, it would be better as a standalone `pyvista-planets` package. Otherwise, this seems like a lot of code for examples that only a small minority of users leverage.

I agree the naming here isn't ideal. I do not want this PR to be the place to refine this and recommend opening an issue for follow up work and to discuss striping this module out into a new `pyvista-examples` repo for instance

pyvista/examples/planets.py
banesullivan: I don't want mess with this `planets` module any more than necessary for this deprecation

pyvista/plotting/plotter.py
banesullivan: > Was it really unsupported?

I'm not sure. I noticed it in the docstring and had the thought that you really shouldn't do texture plotting this way with multiblock datasets so I stripped this out.

> Because if it worked after all then this will be a breaking change once we remove the feature.

I'm not sure I follow?

pyvista/examples/planets.py
adeak: cc @tkoyama010

pyvista/examples/planets.py
adeak: I just meant adding another line of example that shows the other texture being used, but fair enough.

pyvista/plotting/plotter.py
adeak: My thinking is that if meshes track their textures, and multiblocks track their constituent meshes, then `texture=True` might work if every mesh of the multiblock has an associated texture, and the result would be intuitive. If we then deprecate and later remove the option to have associated textures on meshes, we'll lose this feature of being able to plot multiblocks with textures, because you can't pass an external texture that makes sense for the whole multiblock.

Is this clearer?

banesullivan: Ah, yes, very clear. Indeed this would be breaking in that way. If I had to guess, there may be one single person out there using the API in this way... if at all, so a break I'm willing to let slide

adeak (approved):
LGMT with potential spin-off work discussed in https://github.com/pyvista/pyvista/pull/4515#discussion_r1224856381.

pyvista/examples/planets.py
tkoyama010: I made https://github.com/pyvista/pyvista/issues/4526 .

PR #4516, opened 2023-06-09
Set DocumentTheme as new default

banesullivan:
It's time we modernize PyVista's default theme.

The previous `DefaultTheme` always looked dated to me. @akaszynski and I have had many discussions offline about this and I think the `DocumentTheme` theme makes the most sense as default for the majority of people. Further, the documentation uses `DocumentTheme` by default, and users currently see a mismatch in when the run any examples locally (IMO, this is a driving reason to switch).

Changes include:

- Renames `DefaultTheme` to `Theme` (deprecates `DefaultTheme` so it is still useable)
- Updates some examples to use `DocumentTheme`
- Set `DocumentTheme` as the default


pyvista/__init__.py
```diff
@@ -19,7 +19,7 @@
 from pyvista.plotting.utilities.sphinx_gallery import _get_sg_image_scraper
 from pyvista.report import GPUInfo, Report, get_gpu_info, vtk_version_info
 from pyvista.themes import (
-    DefaultTheme as _GlobalTheme,
+    DocumentTheme as _GlobalTheme,
```
banesullivan, line 22: This is the only "breaking" change

banesullivan:
Labelling this as a breaking change since visualization results will be different across versions

pyvista/themes.py
```diff
@@ -1490,7 +1493,7 @@ def default_mode(self, mode: str):
         self._default_mode = mode
 
 
-class DefaultTheme(_ThemeConfig):
+class Theme(_ThemeConfig):
     """PyVista default theme.
```
MatthewFlamm, line 1497: This is no longer the default theme?

pyvista/themes.py
```diff
@@ -2989,7 +2990,24 @@ def lighting_params(self, config: _LightingConfig):
...
+    .. deprecated:: 0.40.0
+        Deprecated an renamed to ``Theme``.
+
+    """
+
+    def __init__(self):
+        """Initialize the theme."""
+        super().__init__()
+        warnings.warn(
+            '`DefaultTheme` has been deprecated and renamed `Theme`. Further, it is no longer the default theme.',
```
MatthewFlamm, line 3005: "DefaultTheme is deprecated and rename Theme.  DefaultTheme is no longer the default theme, nor is Theme the default theme."  The warning message here needs work, although I don't have a better suggestion right now.

MatthewFlamm (commented):
A quick review.

I'm +1 for using DocumentTheme as a nicer default theme.  Additionally it is nicer that the documentation no longer points to `DefaultTheme.method_X` but now `Theme.method_X`.

I'm not sure if it would be better to keep `DefaultTheme`, i.e. to not deprecate it, but to have it wrap `DocumentTheme` now.  This would make it clearer in the code what is the DefaultTheme.  Right now it is buried in the pyvista init importing logic.

pyvista/themes.py
banesullivan:
```suggestion
    """Base VTK theme.
```

pyvista/themes.py
banesullivan: Haha, I spent a few minutes thinking on this and I feel it is clear and concise. I'm open to suggestions to improve the message here.

pyvista/themes.py
```diff
@@ -150,12 +150,11 @@ def set_plot_theme(theme):
...
             raise ValueError(f"Theme {theme} not found in PyVista's native themes.")
         pyvista.global_theme.load_theme(new_theme_type())
-    elif isinstance(theme, DefaultTheme):
+    elif isinstance(theme, Theme):
         pyvista.global_theme.load_theme(theme)
     else:
         raise TypeError(
-            f'Expected a ``pyvista.themes.DefaultTheme`` or ``str``, not '
-            f'a {type(theme).__name__}'
+            f'Expected a ``pyvista.themes.Theme`` or ``str``, not ' f'a {type(theme).__name__}'
```
adeak, line 157:
```suggestion
            f'Expected a ``pyvista.themes.Theme`` or ``str``, not {type(theme).__name__}'
```
(Removing the article to make it insensitive to the first letter of the type name.)

pyvista/themes.py
```diff
@@ -197,7 +196,11 @@ def to_dict(self) -> dict:
         return dict_
 
     def __eq__(self, other):
-        if not isinstance(self, type(other)):
+        if not isinstance(other, _ThemeConfig):
+            return False
+
+        if not isinstance(self, Theme) and not isinstance(self, type(other)):
+            # Makes sure subconfigs are same type
```
adeak, line 203: Do we need the subtype check? Wouldn't the different attributes fail equality checking anyway? Or, conversely: if two different subclasses actually have all the same parameters, aren't they equal?

I'm also always wary about `isinstance(self, type(other))` vs `isinstance(other, type(self))` vs `return NotImplemented`. Which is just to say that this is a treacherous piece of logic to touch here.

For instance, it looks like we posit `DocumentTheme() != DarkTheme()`, but we allow `DocumentTheme() == Theme()` (given modified parameters) based on type? I haven't tested this.

pyvista/themes.py
```diff
@@ -2989,7 +2990,24 @@ def lighting_params(self, config: _LightingConfig):
         self._lighting_params = config
 
 
-class DarkTheme(DefaultTheme):
+class DefaultTheme(Theme):
+    """Deprecated default theme.
+
+    .. deprecated:: 0.40.0
+        Deprecated an renamed to ``Theme``.
```
adeak, line 2997:
```suggestion
        Deprecated and renamed to ``Theme``. No longer the default.
```

pyvista/__init__.py
adeak: Should we update `tests/namespace/namespace-pre-refactor*.txt` due to this? The actual type the `_GlobalTheme` name resolves to is different, but we only check for `_GlobalTheme` in the test. This is more of a general question regarding our policy to keep the namespace test inputs up to date.

banesullivan: You bring up a good point. I think we should try to leave those files alone as best as possible so that they represent the state of PyVista 0.39 but I have not thought out a good policy around this

pyvista/themes.py
banesullivan: > Do we need the subtype check?

We don't *need* it, so i removed it

> Wouldn't the different attributes fail equality checking anyway?

This is perfectly sufficient

> conversely: if two different subclasses actually have all the same parameters, aren't they equal?

Theis is what this block disallows but there's no good reason for this

I've removed this check

MatthewFlamm: The original comment might well have been different, but my comment is at least very related. 
 The current code has an awkward architecture, which creates these equality checking subtleties.  This method shows that we should consider two config classes, one `_ThemeConfig` and another is `_ThemeAttrConfig`.  Then the `__eq__` of each class would only check that `isinstance(other, <the appropriate class here>)`.  This _might_ also help clean up the tuple handling below, but I haven't looked closely.

banesullivan: Should I implement `_ThemeAttrConfig` to do this?

MatthewFlamm: I think my comment was aimed mostly at the original implementation, sorry for not checking the new one more closely.  The current status is good with me.  My suggestion would apply if you wanted to put back in the original check that you had.  There _might_ be a problem if someone compares a `Theme` to a `_LightingConfig` object for example, but this issue can be addressed separately by adjusting the attribute checks.

pyvista/themes.py
```diff
@@ -2989,7 +2986,24 @@ def lighting_params(self, config: _LightingConfig):
...
+    .. deprecated:: 0.40.0
+        Deprecated and renamed to ``Theme``. No longer the default.
+
+    """
+
+    def __init__(self):
+        """Initialize the theme."""
+        super().__init__()
+        warnings.warn(
+            '`DefaultTheme` has been deprecated and renamed `Theme`. Further, it is no longer the default theme.',
```
MatthewFlamm, line 3001: My suggestion after sitting on it... I think listing the new default theme helps point users in the right direction in case they want to follow the new recommended default rather than stay on the old default.

```suggestion
            '`DefaultTheme` has been deprecated and renamed `Theme`. Further, `DocumentTheme` is now the PyVista default theme'.,
```

pyvista/themes.py
MatthewFlamm: > there shouldn't be an external usage of `_LightingConfig` or similar config classes.

You have more confidence than I that users won't be doing unintended things :).  Users could be trying to mix and match Themes, for example:

```py
axes = DarkTheme.axes
my_axes_theme.axes = axes
# accidentally use `axes` somewhere where the intention was to use `my_axes_theme`

```

But let's leave it out of this PR IMO.

banesullivan: I mainly think only a small group of users even know about themes in general but I really have no idea... custom theme support was only recently added.

Ineed much left to potentially address that could come later

akaszynski: I feel this is enough of an edge case that we can either not address it or address it in another PR.

akaszynski (approved):
> I think the `DocumentTheme` theme makes the most sense as default

Fully agree here. This also helps us to align to `matplotlib`'s default [style](https://matplotlib.org/stable/users/prev_whats_new/dflt_style_changes.html).

I'm also pleased about the rename of `DefaultTheme` to `Theme`. That was a questionable choice to begin with on my behalf.

adeak (approved):

Issue #3320, opened 2022-09-14
UniformGrid -> ImageData


banesullivan:
A long, long time ago, I made the decision (ref #24) to call PyVista's wrapping of `vtkImageData` `UniformGrid` -- this decision made a lot of sense at the time and was intuitive to me, my colleagues, and the users of PyVista at the time. Unfortunately, this is at odds with the `vtkUniformGrid` class. Further, PyVista is seeing massive growth lately and more adoption from the VTK community, and I am hoping to bring VTK and PyVista closer together. In doing so, I would like to propose that PyVista move more towards VTK class names in its wrappings (this is only an issue for the wrapping of `vtkImageData`).

This will also add consistency to PyVista's core API as every other wrapped VTK class is the same name just without the `vtk` prefix

Related to #3313

I would like to use this issue as a place to develop a transition plan for renaming `pyvista.UniformGrid` to `pyvista.ImageData` with the intention of later wrapping `vtkUniformGrid` and `vtkStructuredPoints` in PyVista as well.

<img width="369" alt="Screen Shot 2022-09-13 at 9 25 56 PM" src="https://user-images.githubusercontent.com/22067021/190052761-40ab9548-cb54-400a-80b5-2878876cc3d0.png">

This ***will not*** be an easy feat as `pyvista.UniformGrid` is used ***all over the place*** and we need to come up with a thoughtful transition/deprecation plan. My initial thoughts on how this will play out:

Phase 1: Rename and issue a deprecation warning

1. Rename `pyvista.UniformGrid` to `pyvista.ImageData`, wrap the new class as a shell `UniformGrid` class and issue a Deprecation warning in its `__init__`
2. Update all internal examples and documentation to use the new `pyvista.ImageData` class
3. Issue a release and let the chaos ensue

Phase 2: Raise an error, but keep `pyvista.UniformGrid` around.

1. Increase the severity or importance of the Deprecation notice. This could be done through a harsher warning of sorts that can't just be ignored or by outright raising a runtime error. This would be much like what we did for the PyVistaQt `BackgroundPlotter` when migrating it out.

Phase 3: Full deprecation and removal

1. at this point, we may be able to introduce a wrapping of `vtkUniformGrid` as `pyvista.UniformGrid` if the community requests it. However, we should delay wrapping this class as long as possible to avoid issues in the transition

adam-grant-hendry:
Love the effort going into this! I definitely want dibs on helping with the impacts to volume rendering on this 😉

PR #4518, opened 2023-06-09
Rename UniformGrid to ImageData

banesullivan:
Deprecate `UniformGrid` to rename it `ImageData` to be better aligned with VTK data model.

Resolves https://github.com/pyvista/pyvista/issues/3320

Please see https://github.com/pyvista/pyvista/issues/3320 for the original proposal and justification. The sooner we do this, the better.


.github/workflows/integration-tests.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           path: ~/mne_data
       - run: ./tools/github_actions_download.sh
         working-directory: mne-python
-      - run: pytest mne/viz/_brain mne/viz/tests/test_3d.py mne/viz/backends
+      - run: pytest mne/viz/_brain mne/viz/tests/test_3d.py mne/viz/backends -W ignore::pyvista.core.errors.PyVistaDeprecationWarning
```
banesullivan, line 57: cc @larsoner

pyvista/core/filters/image_data.py
```diff
@@ -761,14 +761,17 @@ def _check_fft_scalars(self):
...
 
         """
         alg = _vtk.vtkImageFlip()
         alg.SetInputData(self)
         alg.SetFilteredAxes(axis)
         alg.Update()
         return wrap(alg.GetOutput())
+
+
+UniformGridFilters = ImageDataFilters
```
banesullivan, line 777: Leaving in namespace -- since this is an abstract class, not issue deprecation warning

tests/core/test_grid.py
```diff
@@ -811,82 +811,118 @@ def test_cast_rectilinear_grid():
...
+    # uniform grid from a uniform grid
+    with pytest.warns(PyVistaDeprecationWarning):
+        grid = pyvista.UniformGrid(dimensions=dims, spacing=spacing, origin=origin)
+    with pytest.warns(PyVistaDeprecationWarning):
+        grid_from_grid = pyvista.UniformGrid(grid)
+    assert grid == grid_from_grid
+
+    # and is a copy
+    grid.origin = (0, 0, 0)
+    assert grid != grid_from_grid
```
tkoyama010, line 913: In my mind, it is sufficient to test UniformGrid to verify that a deprecated warning is generated when called.
```suggestion
def test_create_uniform_grid():
    with pytest.warns(PyVistaDeprecationWarning):
        grid = pyvista.UniformGrid()
```

tkoyama010 (changes requested):
+100 for this change. As a PyVista user, I felt that the name UniformGrid was confusing.

tests/core/test_grid.py
banesullivan: Sufficient, yes... but... My original implementation could be reduced but I do want to make sure that arguments are passed along safely. Perhaps the `UniformGrid` class's `__init__` should just pass `*args. **kwargs` instead of listing out the same function signature

pyvista/core/composite.py
```diff
@@ -476,7 +476,7 @@ def get(
         --------
         >>> import pyvista as pv
         >>> from pyvista import examples
-        >>> data = {"poly": pv.PolyData(), "uni": pv.UniformGrid()}
+        >>> data = {"poly": pv.PolyData(), "uni": pv.ImageData()}
```
adeak, line 479: Should we also change the key?
```suggestion
        >>> data = {"poly": pv.PolyData(), "img": pv.ImageData()}
```

pyvista/core/grid.py
```diff
@@ -852,3 +852,37 @@ def extent(self, new_extent: Sequence[int]):
...
         return self.cast_to_rectilinear_grid().to_tetrahedra(*args, **kwargs)
+
+
+class UniformGrid(ImageData):
+    """Deprecated ``UniformGrid`` and renamed to ``ImageData``.
+
+    .. deprecated:: 0.40.0
+        ``pyvista.UniformGrid`` has been deprecated and renamed to ``pyvista.ImageData``.
+
+    Future versions of PyVista will implement ``UnformGrid``
```
adeak, line 863:
```suggestion
    Future versions of PyVista will implement ``UniformGrid``
```

pyvista/core/grid.py
```diff
@@ -852,3 +852,37 @@ def extent(self, new_extent: Sequence[int]):
     def to_tetrahedra(self, *args, **kwargs):
         """Cast to a rectangular grid and then convert to tetrahedra."""
         return self.cast_to_rectilinear_grid().to_tetrahedra(*args, **kwargs)
+
+
+class UniformGrid(ImageData):
+    """Deprecated ``UniformGrid`` and renamed to ``ImageData``.
+
+    .. deprecated:: 0.40.0
+        ``pyvista.UniformGrid`` has been deprecated and renamed to ``pyvista.ImageData``.
```
adeak, line 861: Not sure this will get rendered anywhere, but
```suggestion
        ``pyvista.UniformGrid`` has been deprecated and renamed to :class:`pyvista.ImageData`.
```

pyvista/core/grid.py
```diff
@@ -852,3 +852,37 @@ def extent(self, new_extent: Sequence[int]):
...
+            *args,
+            dimensions=dimensions,
+            spacing=spacing,
+            origin=origin,
+            deep=deep,
+            **kwargs,
+        )
+        warnings.warn(
+            '`UniformGrid` is deprecated. Use `ImageData` instead.', PyVistaDeprecationWarning
+        )
```
adeak, line 888: As also noted in another thread, I'd definitely make this a thin wrapper with `def __init__(self, *args, **kwargs)` to push users to the right type.

tests/core/test_grid.py
```diff
@@ -811,82 +811,118 @@ def test_cast_rectilinear_grid():
...
     with pytest.raises(ValueError, match="Spacing must be non-negative"):
-        grid = pyvista.UniformGrid(dimensions=dims, spacing=(-1, 1, 1))
+        grid = pyvista.ImageData(dimensions=dims, spacing=(-1, 1, 1))
 
     # all args (deprecated)
-    with pytest.warns(
-        PyVistaDeprecationWarning, match="Behavior of pyvista.UniformGrid has changed"
-    ):
-        grid = pyvista.UniformGrid(dims, origin, spacing)
+    with pytest.warns(PyVistaDeprecationWarning, match="Behavior of pyvista.ImageData has changed"):
```
adeak, line 845:
```suggestion
    with pytest.warns(PyVistaDeprecationWarning, match=r"Behavior of pyvista\.ImageData has changed"):
```

tests/core/test_grid.py
```diff
@@ -811,82 +811,118 @@ def test_cast_rectilinear_grid():
...
         assert grid.dimensions == dims
         assert grid.origin == origin
         assert grid.spacing == spacing
 
     # just dims (deprecated)
-    with pytest.warns(
-        PyVistaDeprecationWarning, match="Behavior of pyvista.UniformGrid has changed"
-    ):
-        grid = pyvista.UniformGrid(dims)
+    with pytest.warns(PyVistaDeprecationWarning, match="Behavior of pyvista.ImageData has changed"):
```
adeak, line 852:
```suggestion
    with pytest.warns(PyVistaDeprecationWarning, match=r"Behavior of pyvista\.ImageData has changed"):
```

tests/core/test_grid.py
```diff
@@ -811,82 +811,118 @@ def test_cast_rectilinear_grid():
...
+        grid_from_grid = pyvista.UniformGrid(grid)
+    assert grid == grid_from_grid
+
+    # and is a copy
+    grid.origin = (0, 0, 0)
+    assert grid != grid_from_grid
+
+
+def test_image_data_invald_args():
+    with pytest.warns(PyVistaDeprecationWarning, match="Behavior of pyvista.ImageData has changed"):
```
adeak, line 917:
```suggestion
    with pytest.warns(PyVistaDeprecationWarning, match=r"Behavior of pyvista\.ImageData has changed"):
```

pyvista/core/__init__.py
```diff
@@ -24,11 +24,12 @@
 from .filters import (
     CompositeFilters,
     DataSetFilters,
+    ImageDataFilters,
     PolyDataFilters,
     UniformGridFilters,
```
adeak, line 29: Do people rely on `pv.UniformGridFilters` in the wild? Or is it just better to be safe than saying sorry?

adeak (commented):
Same question as with `DefaultTheme` applies: do we want to update the namespace test names? We'd only run into problems if we remove `UniformGrid` later, before restoring it as a `vtkUniformGrid` subclass.

pyvista/core/composite.py
banesullivan: I'm going to here because you pointed it out, but there are a TON of these and I really don't want to waste time digging all of these up. In time we will transition more of these. 

I tried to change the most obvious ones like `uniform = pv.ImageData()` to `image = pv.ImageData()`

pyvista/core/__init__.py
banesullivan: I really, really hope no one is using this. This is more of a better safe than sorry and putting off the namespace cleanup for a follow up PR

.github/workflows/integration-tests.yml
larsoner: You should be able to revert this filter now

pyvista/core/composite.py
adeak: Sure thing. I just didn't have the bandwidth to filter what I came across past 1 AM :D

adeak (approved):
> Same question as with `DefaultTheme` applies: do we want to update the namespace test names? We'd only run into problems if we remove `UniformGrid` later, before restoring it as a `vtkUniformGrid` subclass.

Assuming an answer to leave it alone, in line with https://github.com/pyvista/pyvista/pull/4516#discussion_r1225595002.

Thanks @banesullivan, LGTM.

tkoyama010 (dismissed):

tkoyama010 (approved):

adeak (approved):

PR #4254, opened 2023-04-10
Picking overhaul

banesullivan:
An internal overhaul to the picking interface to improve stability and useability with the following upgrades:

- Picking now works across subplots (see new "poked renderer" APIs)
- Right-click point picking is enabled by default
- A new "element" picking interface to select individual meshes, cells, faces, and edges (not recommended for points)
- Improved setter/getter for picker types

The redesign splits picking into two modes:

1. **Point-based picking** - use the cursor to pick a single point within the scene and use that point to perform an operation (e.g., select the mesh, cell, face, edge at that point)
2. **Rectangle-based picking** - use a selection rectangle on the viewport of the scene to make an area-based selection. This will build a frustum from the viewport area to use to select within the scene

All additional picking methods build on those two interfaces.

Backward compatibility is maintained for the most part (some callbacks did change). Specifically, I need to document what changed:

- `use_mesh` argument for point picking has been removed
- the passed values for some callbacks have changed

### Resolves

Resolves #1012
Resolves #3564
Resolves #2578
Resolves #3027 - makes it easier to choose which picker to use
Resolves #3703
Resolves #3915
Resolves #4381

### To do

- [x] Minor implementation cleanups
- [x] Verify all examples work as expected
- [x] Complete and check documentation
- [x] Improve `picker` setter/getter interface for tab-completion (not just hardcoded strings)
- [x] Additional tests for coverage
- [x] Provide examples for new capabilities

Issue #2520, opened 2022-04-27
Picking doesn't work with render passes


apwebber:
**Describe the bug, what's wrong, and what you expect:**

When shadows are enabled on a plot, picking meshes either doesn't do anything or throws an error. I've encountered two errors, they seem to be different for either pressing p or left clicking.

pressing p:
vtkInteractorStyle.cxx:937   WARN| vtkInteractorStyleTrackballCamera (0000018DBE1A2D80): no current renderer on the interactor style.

left clicking:
2022-04-27 11:21:10.522 (   3.270s) [                ]vtkOpenGLHardwareSelect:203    ERR| vtkOpenGLHardwareSelector (00000216393A90B0): Too many props. Currently only 16777214 props are supported.

-----

**To Reproduce**

Run the dragon shadows example as found [here](https://docs.pyvista.org/examples/04-lights/shadows.html).

Add this code after the `pyvista.Plotter` is defined:

`pl.enable_mesh_picking(style='surface', left_clicking=True)`

run the example with and without the line `pl.enable_shadows()`

-----

**System Information:**

```txt
Date: Wed Apr 27 11:25:08 2022 GMT Summer Time

                OS : Windows
            CPU(s) : 8
           Machine : AMD64
      Architecture : 64bit
               RAM : 15.7 GiB
       Environment : Python
        GPU Vendor : Intel
      GPU Renderer : Intel(R) Iris(R) Xe Graphics
       GPU Version : 4.5.0 - Build 30.0.100.9864

  Python 3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit
  (AMD64)]

           pyvista : 0.34.0
               vtk : 9.1.0
             numpy : 1.21.2
           imageio : 2.16.2
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.4.3
         pyvistaqt : 0.9.0
             PyQt5 : 5.15.4
           IPython : 7.28.0
             scipy : 1.7.1
              tqdm : 4.62.3
```

banesullivan:
I'm able to reproduce this, and it's unclear to me if there's a solution to this... it appears to be an issue with any render passes and not limited to just shadows.

akaszynski:
#4254 fixes this. Minimum working code:

```py
import pyvista as pv

sphere = pv.Sphere(center=(1, 0, 0))
cube = pv.Cube()

pl = pv.Plotter()
[...]

PR #4521, opened 2023-06-10
Change antialiasing default to ssaa

### LINKED ISSUE #4524: test_plot_cell has a very high variance
```
FAILED tests/plotting/test_plotting.py::test_plot_cell - RuntimeError: test_plot_cell Exceeded image regression error of 500.0 with an image error equal to: 2964.9764705882176
```

This test fails on my system (its the only image regression test that fails) due to point size and line size differences with my EGL build of VTK. I tried setting this as a `high_variance_test`, which increases the tolerance to `1000`, but the difference is nearly 3000. Should we increase the tolerance for this test or perhaps skip it on some builds of VTK?

| Saved in image cache | What's produced on my system |
|---|---|
| plot_cell |  plot_cell |

banesullivan:
I use a custom build of VTK for performant off screen rendering with my GPU - an "EGL" build. A handful of PyVista's defaults and tests have been incompatible with this build. These changes rectify that by changing the default anti-aliasing to `ssaa` instead of `fxaa` -- @akaszynski, I'd appreciate your thoughts on this change and if its okay?

Resolves #4524


banesullivan:
> I really feel that having plots look awesome out of the box should be the default. We can then have a theme that's more suited for low power GPUs

Completely agree here -- let's go for this

banesullivan:
NOTE: to validate this is not breaking the docs, we should trigger a "workflow dispatch" job for the documentation build disabling the cache.

banesullivan:
So maybe enabling antialiasing breaks things like picking... might need to rethink enabling it by default

banesullivan:
Ah this has been tracked in https://github.com/pyvista/pyvista/issues/2520

akaszynski:
> Ah this has been tracked in #2520

Sorta. There are indeed issues when enabling SSAA, but it's not in #2520. One I've noticed off-hand is pressing "f" no longer focuses on the mesh. However, enabling picking (for example `enable_mesh_picking`) causes pressing "f" to both focus on the picked point and to trigger picking it. There's a bit to debug there, and I'd like to have this resolved in a separate PR.

banesullivan:
Looks like tests are now passing after #4254

akaszynski:
Interesting... After #4254, pressing "f" no longer focuses on the point under the cursor (regardless of the settings of anti-aliasing).

akaszynski:
> Interesting... After #4254, pressing "f" no longer focuses on the point under the cursor (regardless of the settings of anti-aliasing).

Got it! Simply setting the default picker to the interactor fixes this regression:

```py
import pyvista as pv
from pyvista.plotting.opts import PickerType

sphere = pv.Sphere(center=(1, 0, 0))

pl = pv.Plotter()
pl.add_mesh(sphere, color='lightblue', show_edges=True)
# pl.enable_anti_aliasing('ssaa')  # works with and without anti-aliasing
pl.iren.picker = PickerType.POINT
pl.show()
```

PR #4538, opened 2023-06-13
Prevent deprecation warnings in examples

banesullivan:
This is mostly a follow up to #4515 but it switches the docs build to treat `PyVistaDeprecationWarning` as errors in our examples -- helping to prevent users from adopting deprecated usage patterns.


tkoyama010 (approved):
LGTM. Thanks!

PR #4568, opened 2023-06-23
Add GeoVista integration test

### LINKED ISSUE #4565: GeoVista Integration Testing
@bjlittle has created a truly amazing extension to PyVista for 3D cartographic rendering and I'm using it pretty heavily in my work lately. Considering this project is gaining a good bit of traction, I think we should add it to our integration tests to make sure we account for how it uses PyVista when changing our API.

See https://github.com/bjlittle/geovista

tkoyama010:
### Overview

We should add GeoVista to our integration tests to make sure we account for how it uses PyVista when changing our API.

Ping @bjlittle

relate to  https://github.com/pyvista/pyvista/pull/3967

resolves #4565

### Details

- None


.github/workflows/integration-tests.yml
```diff
@@ -74,3 +74,23 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: git clone https://github.com/bjlittle/geovista.git --single-branch
+      - name: Install PyVista
+        run: pip install -ve . # pyvista
+      - run: sudo apt-get install libgeos-dev
```
bjlittle, line 89: @tkoyama010 Note that `geovista` is using `cartopy` as a minor convenience at the moment, and I'm keen to drop that dependency asap. When that happens there won't be any need to install `libgeos-dev` :+1:

.github/workflows/integration-tests.yml
```diff
@@ -74,3 +74,23 @@ jobs:
...
+      - name: Install GeoVista
+        run: pip install -ve . # geovista
+        working-directory: geovista
+      - run: pip install -r requirements/pypi-optional-test.txt
+        working-directory: geovista
+      - run: python -m pytest
```
banesullivan, line 95: I'm slightly concerned that the warnings aren't causing this to failed

https://github.com/pyvista/pyvista/actions/runs/5361402806/jobs/9727309100?pr=4568#step:9:1633

Note those warnings are unrelated to PyVista, but we want to fail on warnings to make sure deprecations are handled in GeoVista

tkoyama010: Good point. This must raise warnings as errors (see [How to capture warnings](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html)).
```suggestion
      - run: pytest -W error
```

.github/workflows/integration-tests.yml
tkoyama010: @bjlittle The download warning of `cartopy` is now an error due to set alerts to capture warnings. We will wait for `cartopy` installation and testing to become an option in `geovista`.

bjlittle: Didn't realize that you tripped on warnings by default. I'd suggest downloading cartopy assets prior to running the tests, this should avoid such warnings. I'll push a change to your PR later today, if that's okay?

tkoyama010: @bjlittle Yes, please. Thank you for your cotribution.

banesullivan: > Didn't realize that you tripped on warnings by default

I think its generally a good practice to prevent deprecation warnings from becoming deprecation errors.

We can also just specifically ignore the cartopy warnings. For this integration testing, we mainly just want to catch numpy and pyvista warnings

tkoyama010: @bjlittle Thank you for the PR of the solution.

@banesullivan We will adopt @bjlittle 's solution this time. If there are more warnings in the future, we will adopt your solution.

bjlittle:
@tkoyama010 Hmmmm the `codecov/patch` task seems "stuck".

Never seen that before, you?

tkoyama010:
@bjlittle I don't understand why, but waiting is often the answer :)

banesullivan (approved):

banesullivan:
No need to worry above coverage here

tkoyama010:
As this is CI PR, it is considered to be outside the 24-hour rule. Thank you all for your cooperation.

bjlittle:
@tkoyama010 @banesullivan @akaszynski 

Awesome, thanks so much guys for banking this PR! :beers: 

Looking forward to meeting you all at SciPy!

tkoyama010:
@bjlittle Thank you. Let's go drinking during SciPy.

PR #4576, opened 2023-06-25
Set default picker

akaszynski:
Set a default picker when initializing the render window interactor. This fixes the minor regression made in #4254 which removed the convenient ability to press "f" to focus and fly to the point under the cursor.


banesullivan (approved):

akaszynski:
Actually, we’re going to need to go with the `POINT` picker here. Two of the fly two images have minor variations and will need to be regenerated.

banesullivan (approved):

PR #4586, opened 2023-06-27
Fix 3D objects not being displayed in example of  `save_graphic`

tkoyama010:
### Overview

resolves https://github.com/pyvista/pyvista/discussions/4584

| Main  | PR |
| ------------- | ------------- |
| image |  image |

### Details

- None


adeak:
Which change fixes the issue? Is it the render step that was missing before saving? We might consider pointing that out in the examples, or even making it an optional feature of the export.

tkoyama010:
I began to think that it would be user-friendly to follow the screenshot approach. Please give me some time to consider this.
https://github.com/pyvista/pyvista/blob/65042bc9c1f9b0c9b1722c1dca74fcbd862a4b25/pyvista/plotting/plotter.py#L5714-L5720

pyvista/plotting/plotter.py
```diff
@@ -5631,10 +5631,12 @@ def save_graphic(self, filename, title='PyVista Export', raster=True, painter=Tr
         --------
         >>> import pyvista
         >>> from pyvista import examples
-        >>> pl = pyvista.Plotter()
+        >>> pl = pyvista.Plotter(off_screen=True)
```
banesullivan, line 5634: I'm hoping this `off_screen` argument isn't necessary and that we are setting `pyvista.OFF_SCREEN` when running the docstrings.

tkoyama010: Good point. thanks!

tkoyama010:
Rendering during first-time rendering solved this issue.
img

banesullivan (approved):

PR #4591, opened 2023-06-27
Limit Docker image builds on PRs

banesullivan:
We're publishing too many docker images (on for every PR). This changes it to only run on a PR if the branch name starts with `docker`.

Intentionally using `maint/*` here to make sure it skips

### DIFF
--- a/.github/workflows/docker-package.yml
+++ b/.github/workflows/docker-package.yml
@@ -20,15 +20,11 @@ jobs:
   build-and-publish:
     name: Build and Publish Docker Image
     runs-on: ubuntu-20.04
-    # For Pull Requests, only runs from `docker`, `feat`, `fix`, `patch`, or `dependabot`
+    # For Pull Requests, only runs from `docker`
     if: github.event_name == 'push' ||
-        ( github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository &&
-          ( startsWith(github.event.pull_request.head.ref, 'feat') ||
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-package.yml
```diff
@@ -20,15 +20,11 @@ jobs:
   build-and-publish:
     name: Build and Publish Docker Image
     runs-on: ubuntu-20.04
-    # For Pull Requests, only runs from `docker`, `feat`, `fix`, `patch`, or `dependabot`
+    # For Pull Requests, only runs from `docker`
```
adeak, line 23: Only because I don't know these details: this doesn't affect gitpod devcontainers, does it?

banesullivan: That's a good question... I don't think so? Hopefully the pod is based on the main branch image

akaszynski (approved):
Makes sense to not ship on each commit to the main branch.

banesullivan:
as it stands, this still build for every commit on main

.github/workflows/docker-package.yml
adeak: Cc @MatthewFlamm just to be safe.

MatthewFlamm: This build action does not involve any builds for gitpod or devcontainer which are currently built from supported base images from the respective providers and built using their machinery.  I'm not actually sure what these build actions are used for, so I have no real comment on this PR.

tkoyama010 (approved):

PR #4594, opened 2023-06-27
Clarify the difference between n_verts and n_points

darikg:
#4592 was a pretty understandable confusion, this adds some clarification to the n_verts docstring

### DIFF
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -919,7 +919,8 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
     def n_verts(self) -> int:
         """Return the number of vertices.
 
-        A vertex is a single element cell referencing a single point.
+        A vertex is a 0D cell, which is usually a cell that references one point,
+        a vtkVertex.  It can also be a vtkPolyVertex.
         See `pyvista.PolyData.n_points` for the more common measure.
 
         Examples
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -937,7 +937,9 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
         If any other cells are specified, these vertices are not created.
 
         >>> import pyvista
-        >>> mesh = pyvista.PolyData([[1.0, 0.0, 0.0], [1.0, 1.0, 1.0]], lines=[2, 0, 1])
+        >>> mesh = pyvista.PolyData(
+        ...     [[1.0, 0.0, 0.0], [1.0, 1.0, 1.0]], lines=[2, 0, 1]
+        ... )
         >>> mesh.n_points, mesh.n_verts
         (2, 0)
 
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -934,7 +934,7 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
         >>> mesh.n_points, mesh.n_verts
         (2, 2)
 
-        If any other cells are specifed, these vertices are not created.
+        If any other cells are specified, these vertices are not created.
 
         >>> import pyvista
         >>> mesh = pyvista.PolyData([[1.0, 0.0, 0.0], [1.0, 1.0, 1.0]], lines=[2, 0, 1])
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -919,15 +919,26 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
     def n_verts(self) -> int:
         """Return the number of vertices.
 
+        A vertex is a single element cell referencing a single point.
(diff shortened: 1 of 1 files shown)


pyvista/core/pointset.py
```diff
@@ -919,15 +919,26 @@ def n_lines(self) -> int:
     def n_verts(self) -> int:
         """Return the number of vertices.
 
+        A vertex is a single element cell referencing a single point.
```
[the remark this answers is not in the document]
MatthewFlamm: What I mean by this, is that it is possible that a polyvertex is also included, which is a single cell that references a set of points (in a way that I don't fully understand).

pyvista/core/pointset.py
```diff
@@ -919,15 +919,26 @@ def n_lines(self) -> int:
...
+        vertices are automatically created, one per point.
 
         >>> import pyvista
         >>> mesh = pyvista.PolyData([[1.0, 0.0, 0.0], [1.0, 1.0, 1.0]])
-        >>> mesh.n_verts
-        2
+        >>> mesh.n_points, mesh.n_verts
+        (2, 2)
+
+        If any other cells are specifed, these vertices are not created.
```
banesullivan, line 936: Not exactly. It's perfectly okay to have points that are registered to face cells and as their own vertex cell. For example:

```py
import pyvista as pv
from pyvista import examples

mesh = examples.download_bunny_coarse()
mesh.verts = mesh._make_vertex_cells(mesh.n_points)
```
```py
>>> mesh.n_verts
872
>>> mesh.plot()
```
screenshot (28)

There used to be some example datasets with this scenario but I can't remember which

MatthewFlamm: I think that what is being conveyed is pyvista's model when instantiating a `pyvista.PolyData` object, so this could probably just be clarified a bit that this example is chosen to demonstrate the difference.

banesullivan: Ah yes, that is the behavior when using the constructor

darikg: Yup, was trying to clarify the behavior of the PolyData constructor in that specific example, where vertices are constructed only implicitly

banesullivan: In that case, great as is! Sorry for my confusion -- trying to blast through lots of notifications

banesullivan (approved):

pyvista/core/pointset.py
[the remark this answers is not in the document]
banesullivan: This interface on vtkPolyData has always felt odd to me... I'm not sure I have a great answer for your first question. Indeed we're just following the VTK data model here.

> is that it is possible that a polyvertex is also included, which is a single cell that references a set of points (in a way that I don't fully understand).

I'm not sure I follow what you're asking? But a vertex cell can only ever point to one Point ID, not a set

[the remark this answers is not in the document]
MatthewFlamm: A vtkVertex is one cell that has one point.  A vtkPolyVertex is one cell that has more than one point, but no topological structure.  That is both are 0D cells.  I'm not sure if `n_verts` here is the sum of all 0D cells or just the vtkVertex ones.  The API isn't explained well enough in vtk for me to know, but most importantly **the description here would not be correct if it includes all 0D cells**.  I can do some testing soon.

This is similar to the difference of vtkLine and vtkPolyLine.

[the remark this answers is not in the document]
MatthewFlamm: Ok, I tested this.  `n_verts` includes _all_ 0D cells, so it is not true that it is just cells with one point.  Our API is especially awkward because we do not allow to instantiate vertices as a separate input.  We need to overhaul this API a bit.

```py
import pyvista as pv
p = [[0, 0, 0], [1, 0, 0]]
m = pv.PolyData(p)  # no way to control 0D cells
m.n_verts # 2
m.verts = [2, 0, 1]  # make 1 vtkPolyVertex
m.n_verts # 1
m.get_cell(0)  # shows it is POLY_VERTEX
```

MatthewFlamm, line 922:
```suggestion
        A vertex is a 0D cell, which is usually a cell that references one point,
        a vtkVertex.  It can also be a vtkPolyVertex.
```

pyvista/core/pointset.py
adeak: Side note: so many times I wanted to plot points as well as faces for a mesh in a single call (because I'm lazy). TIL I need `verts` for that.

banesullivan (approved):

pyvista/core/pointset.py
```diff
@@ -919,15 +919,27 @@ def n_lines(self) -> int:
...
```
MatthewFlamm, line 937: codespell fix

```suggestion
        If any other cells are specified, these vertices are not created.
```

MatthewFlamm:
pre-commit.ci autofix

banesullivan (approved):

PR #4606, opened 2023-06-29
Move Vale CI job

banesullivan:
I feel the Vae CI job should be in the docstring CI workflow


banesullivan:
Additionally, this prevents it from running on `main` -- this is only needed for PRs

akaszynski (approved):

PR #4629, opened 2023-07-05
Isolate plotting API

### LINKED ISSUE #4445: Improve VTK imports to avoid GL-dependencies in core module

banesullivan:
Isolates the `pyvista.plotting` subpackage such that `import pyvista` and use of the core API does not require GL-dependent VTK modules and does not require any additional system libraries.

This is a follow up to #4486 which completely resolves #4445

We're seeing more and more requests to use PyVista's well-documented and feature-rich core API for data processing needs outside of plotting (a testament to how truly powerful this API has become), and these users are finding that PyVista's tight coupling to GL-dependent modules in VTK is a problem. This PR fixes this.

These changes will go a long way in making it possible to use PyVista's robust wrapping of the VTK data model and file IO in lightweight, headless environments.

CI testing is modified here to validate that the core API tests run without installing mesa or using `xvfb-run`.

At this time, I do not think it is feasible to completely split the core and plotting modules into separate packages (I gave this a fair attempt, and it proved to be a massive undertaking), but perhaps this approach is a 90% solution that addresses the actual needs.

I needed to fully remove the texture tracking, which was deprecated recently in #4515 

Opening as a draft PR to finalize testing on CI, but would appreciate feedback @pyvista/developers 

## Example

Use the `python:3.10` base docker image with no additional system libraries and install this branch. 

On this branch, datasets can be created and filters can be used, but as soon as plotting is attempted, we get the infamous `libGL.so.1: cannot open ...`

```py
>>> import pyvista as pv
>>> mesh = pv.Wavelet()
>>> cntrs = mesh.contour()
>>> cntrs.bounds
(-10.0, 10.0, -10.0, 10.0, -10.0, 10.0)
>>> cntrs.n_arrays
1
>>> cntrs.plot()
  File "/app/pyvista/_plot.py", line 224, in plot
    pl = pyvista.Plotter(
  File "/app/pyvista/__init__.py", line 93, in __getattr__
    _module = importlib.import_module('pyvista.plotting')
  File "/usr/local/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/app/pyvista/plotting/__init__.py", line 7, in <module>
    from . import _vtk
  File "/app/pyvista/plotting/_vtk.py", line 125, in <module>
    from ._vtk_gl import *
  File "/app/pyvista/plotting/_vtk_gl.py", line 20, in <module>
    from vtkmodules.vtkRenderingOpenGL2 import (
ImportError: libGL.so.1: cannot open shared object file: No such file or directory
>>> 
```

Whereas on `main`, this fails immediately on import:

```py
>>> import pyvista as pv
  File "/app/pyvista/__init__.py", line 15, in <module>
    from pyvista.errors import InvalidCameraError, RenderWindowUnavailable
  File "/app/pyvista/errors.py", line 19, in <module>
    from pyvista.plotting.errors import InvalidCameraError, RenderWindowUnavailable
  File "/app/pyvista/plotting/__init__.py", line 7, in <module>
    from . import _vtk
  File "/app/pyvista/plotting/_vtk.py", line 125, in <module>
    from ._vtk_gl import *
  File "/app/pyvista/plotting/_vtk_gl.py", line 20, in <module>
    from vtkmodules.vtkRenderingOpenGL2 import (
ImportError: libGL.so.1: cannot open shared object file: No such file or directory
>>> 
```


pyvista/examples/planets.py
annehaley, line 1: You may want to remove "with texture" from all the descriptions of affected methods in this file.

pyvista/errors.py
```diff
@@ -1,19 +1,67 @@
...
+        return globals()[name]
+    except KeyError:
+        pass
+
+    for module in _MODULES:
+        feature, import_path = _try_import(module, name)
+        if feature is not None:
+            break
+    else:
+        raise AttributeError(
```
annehaley, line 56: It looks like this block is the only section that codecov is flagging. You'll just need to add a test that will catch this error.

annehaley (commented):
This looks good; it seems like mostly refactoring. I tried out the branch and successfully replicated the example you gave above. I tried out the planet examples, too.

pyvista/__init__.py
```diff
@@ -74,3 +54,45 @@
 
 # Name used for unnamed scalars
 DEFAULT_SCALARS_NAME = 'Data'
+
+MAX_N_COLOR_BARS = 10
+
+
+# Lazily import/access the plotting module
+def __getattr__(name):
+    """Fetch an attribute ``name`` from ``globals()``or the ``pyvista.plotting`` module..
```
adeak, line 63:
```suggestion
    """Fetch an attribute ``name`` from ``globals()`` or the ``pyvista.plotting`` module.
```

pyvista/__init__.py
```diff
@@ -74,3 +54,45 @@
...
+    """
+    try:
+        return globals()[name]
+    except KeyError:
+        pass
+
+    import importlib
+    import inspect
+
+    whitelist = [
```
adeak, line 82: Mild recommendation to go with the times and call this `allow_list` or `shortlist` (or something else) instead. And make it a set for slightly neater lookup?

pyvista/__init__.py
```diff
@@ -74,3 +54,45 @@
...
+        'demos',
+        'examples',
+        'ext',
+        'trame',
+        'utilities',
+    ]
+    if name in whitelist:
+        return importlib.import_module(f'pyvista.{name}')
+
+    _module = importlib.import_module('pyvista.plotting')
```
adeak, line 92: I know we discussed something like this recently, but isn't this just `import pyvista.plotting as plotting` (instead of `_module` if we do this)?

pyvista/__init__.py
```diff
@@ -74,3 +54,45 @@
...
+    try:
+        feature = inspect.getattr_static(_module, name)
+    except AttributeError as e:
+        raise AttributeError(e) from None
```
adeak, line 96: Isn't this mostly equivalent to
```suggestion
    try:
        feature = inspect.getattr_static(_module, name)
    except AttributeError:
        raise
```
Which is just
```suggestion
    inspect.getattr_static(_module, name)
```
Or is the point exactly to hide the `getattr_static` call from the traceback?

adeak (commented):
Just a few things that jumped out to me on a cursory glance. I'll only have time for a proper review later.

pyvista/errors.py
banesullivan: I may just `pragma: no cover` this

pyvista/__init__.py
banesullivan: Yes, will change to `import ...` syntax

pyvista/__init__.py
banesullivan: > Or is the point exactly to hide the getattr_static call from the traceback?

This is what I was after.... but I didn't do a good job of this in hindsight. I'm improved the error message here so now this is what happens:

```py
>>> import pyvista as pv
>>> pv.Foo
  File "/home/local/KHQ/bane.sullivan/Software/pyvista/pyvista/pyvista/__init__.py", line 97, in __getattr__
    raise AttributeError(f"module 'pyvista' has no attribute '{name}'") from None
AttributeError: module 'pyvista' has no attribute 'Foo'
>>> 
```

CONTRIBUTING.rst
```diff
@@ -543,7 +543,6 @@ the ``verify_image_cache`` fixture can be utilized:
 .. code:: python
 
 
-       @skip_no_plotting
```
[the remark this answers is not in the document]
banesullivan: This is not a global fixture so I didn't think it should be in the guide

pyvista/plotting/plotter.py
```diff
@@ -3370,18 +3371,13 @@ def add_mesh(
...
+                scalar_bar_args.setdefault('title', scalars)
             else:
-                # Make sure scalars components are not vectors/tuples
-                scalars = mesh.active_scalars_name
-                # Don't allow plotting of string arrays by default
-                if scalars is not None:  # and np.issubdtype(mesh.active_scalars.dtype, np.number):
-                    scalar_bar_args.setdefault('title', scalars)
-                else:
-                    scalars = None
+                scalars = None
```
adeak, line 3380: The inherited `else` seems redundant here:
```py
if scalars is not None:
    ...
else:
    scalars = None
 ```

And I'm a bit wary of the original logic here:
```py
                if scalars is not None:  # and np.issubdtype(mesh.active_scalars.dtype, np.number):
                    scalar_bar_args.setdefault('title', scalars)
                else:
                    scalars = None
                scalars = None
```
This looks to me that even in the `scalars is not None` case we replace the value with `salars=None`. This no longer happens now. Any surprises downstream?

tests/plotting/test_texture.py
```diff
@@ -224,42 +223,13 @@ def test_texture_coordinates():
...
     mesh.texture_map_to_plane(inplace=True, name="tex_b", use_bounds=True)
     assert not np.allclose(mesh["tex_a"], mesh["tex_b"])
-    texture = pv.Texture(examples.mapfile)
-    with pytest.warns(PyVistaDeprecationWarning):
-        mesh.textures["tex_a"] = texture.copy()
-    with pytest.warns(PyVistaDeprecationWarning):
-        mesh.textures["tex_b"] = texture.copy()
-    with pytest.warns(PyVistaDeprecationWarning):
-        mesh._activate_texture("tex_a")
-    assert np.allclose(mesh.active_t_coords, mesh["tex_a"])
```
adeak, line 250: Do we need to throw out the `active_t_coords` checks here? We still have the functionality, so how does one activate them?

tests/examples/test_downloads.py
```diff
@@ -77,8 +77,8 @@ def test_file_copier(tmpdir):
 
 def test_local_file_cache(tmpdir):
     """Ensure that pyvista.examples.downloads can work with a local cache."""
-    basename = os.path.basename(examples.mapfile)
-    dirname = os.path.dirname(examples.mapfile)
+    basename = os.path.basename(examples.planefile)
+    dirname = os.path.dirname(examples.planefile)
```
adeak, line 81: Why the switch from one to the other?

pyvista/errors.py
```diff
@@ -1,19 +1,67 @@
...
+    Raises
+    ------
+    AttributeError
+        If the attribute is not found in ``globals()`` and also could not be
+        imported from the modules in ``_MODULES``.
+
+    PyVistaDeprecationWarning
+        If the attribute has been found via importing from the modules in
+        ``_MODULES``, as this implies that the feature has been moved from
+        pyvista.utilities.
```
adeak, line 41: This might have to go in [`Warns`](https://numpydoc.readthedocs.io/en/latest/format.html#warns).
```suggestion
    Warns
    -----
    PyVistaDeprecationWarning
        If the attribute has been found via importing from the modules in
        ``_MODULES``, as this implies that the feature has been moved from
        ``pyvista.utilities``.
```

pyvista/errors.py
```diff
@@ -1,19 +1,67 @@
...
+    return feature, import_path
+
+
+def __getattr__(name):
+    """Fetch an attribute ``name`` from ``globals()`` and warn if it's from a deprecated module.
+
+    Note that ``__getattr__()`` only gets called when ``name`` is missing
+    from the module's globals. The trick is that we want to import this
+    function into other deprecated modules, and we want to carry this
+    subpackage's globals along to prevent some spurious warnings.
```
adeak, line 30: Does this "import elsewhere" caveat still apply here? If not, the logic can be simplified I think, e.g. `globals()[name]` should never succeed as per this paragraph.

tests/namespace/test_errors_namespace.py
```diff
@@ -0,0 +1,29 @@
...
+        'PyVistaEfficiencyWarning',
+        'PyVistaFutureWarning',
+        'PyVistaPipelineError',
+        'VTKVersionError',
+        'InvalidCameraError',
+        'RenderWindowUnavailable',
+    ],
+)
+def test_utilities_namespace(name):
+    with pytest.warns(PyVistaDeprecationWarning):
```
adeak, line 26: Would it be overkill to separate the core errors from the plotting ones, and also asserting the correct warning message?

pyvista/plotting/themes.py
```diff
@@ -2742,12 +2712,12 @@ def axes(self, config: _AxesConfig):
...
 
     @property
-    def before_close_callback(self) -> Callable[[Plotter], None]:
+    def before_close_callback(self) -> Callable:
         """Return the default before_close_callback function for Plotter."""
         return self._before_close_callback
 
     @before_close_callback.setter
-    def before_close_callback(self, value: Callable[[Plotter], None]):
+    def before_close_callback(self, value: Callable):
```
adeak, line 2720: The return type information is actually pretty important (although I'm not sure how important it is on the theme...). I assume you removed the type signature because you removed the `plotting` dependencies. So would it work to restore these, and put the `Plotter` import inside `if typing.TYPE_CHECKING`? Do people use IDEs or other type checkers on headless installs?

It's also a bit weird if themes are `plotting.themes` and yet with carefully removed plotting dependencies. Just a thought. I also understand that functionally they belong to plotting. But doesn't importing this module already imply importing the GL stuff? I think if you `import pyvista.plotting.themes` you run `pyvista.plotting.__init__` which imports all the renderer stuff and whatnot (as it should). So perhaps we don't have to scrub this module?

.github/workflows/testing-and-deployment.yml
```diff
@@ -195,15 +189,24 @@ jobs:
...
+      - name: Core Testing (no GL)
+        run: python -m pytest --cov=pyvista -v tests/core tests/examples
+
+      - uses: awalsh128/cache-apt-pkgs-action@v1.1.3
+        with:
+          packages: libgl1-mesa-glx xvfb
+          version: 3.0
+
+      - name: Plotting Testing (uses GL)
+        run: xvfb-run python -m pytest --cov=pyvista --cov-append --cov-report=xml --fail_extra_image_cache -v --ignore=tests/core --ignore=tests/examples
```
adeak, line 201: There's no way that installing GL would break non-plotting modules, right? Just wondering whether ignoring the already checked two test subpackages is completely safe.

pyvista/__init__.py
```diff
@@ -74,3 +54,46 @@
 
 # Name used for unnamed scalars
 DEFAULT_SCALARS_NAME = 'Data'
+
+MAX_N_COLOR_BARS = 10
+
+
+# Lazily import/access the plotting module
+def __getattr__(name):
+    """Fetch an attribute ``name`` from ``globals()`` or the ``pyvista.plotting`` module..
```
adeak, line 63:
```suggestion
    """Fetch an attribute ``name`` from ``globals()`` or the ``pyvista.plotting`` module.
```

pyvista/__init__.py
```diff
@@ -74,3 +54,46 @@
...
+
+    allow = {
+        'demos',
+        'examples',
+        'ext',
+        'trame',
+        'utilities',
+    }
+    if name in allow:
+        return importlib.import_module(f'pyvista.{name}')
```
adeak, line 90: Couldn't we just explicitly import these subpackages rather than deferring them to `__getattr__()`? They don't look very plotty to me.

pyvista/__init__.py
```diff
@@ -74,3 +54,46 @@
...
+    Raises
+    ------
+    AttributeError
+        If the attribute is not found.
+
+    """
+    try:
+        return globals()[name]
+    except KeyError:
+        pass
```
adeak, line 77: Same remark as the other `__getattr__()` about this being redundant if this function is never imported into other modules.

tests/namespace/test_errors_namespace.py
adeak: Having thought about it a bit, we could add the new import path to the parameterisation, and also check that this path is what gets imported by the deprecated path. So I think this would be worth the conplexity.

adeak:
Added [breaking-change] due to the fairly sudden texture tracking tear-out.

tests/plotting/test_texture.py
banesullivan: Oh whoops. Good catch

tests/examples/test_downloads.py
banesullivan: Because this was loading the image file as a texture and `pv.Texture` is now in the `plotting` subpackage rather than `core` due to its libGL requirement. This allows us to test this module in the first stage of CI without graphics libs. Texture are well tested otherwise.

pyvista/plotting/themes.py
MatthewFlamm: > The return type information is actually pretty important (although I'm not sure how important it is on the theme...)

I'd say it is as important as any other typing.  If I want to set something on the theme, it is useful to know what I need to set.  It makes type checking more useful if everything is as precise as possible in my experience.  I have issues with type checking in downstream applications due to lax typing in pyvista, mostly in filters.

> Do people use IDEs or other type checkers on headless installs?

I don't, but there could be a vim/emacs user who has linked in mypy or other typing functionality into their workflow on a headless machine.   The Venn diagram starts to shrink quite a bit, but it isn't hard to support this the right way, if needed (pending discussion on why this was removed).

.github/workflows/testing-and-deployment.yml
MatthewFlamm: If this is a concern, one possibility is splitting this into two separate jobs that run concurrently.  That is run a separate job for core only and a separate job for core+plotting.  This would not increase the testing times, just the number of concurrent jobs.

pyvista/plotting/themes.py
adeak: I meant that callbacks are trivially `Callable`. But it's not at all irrelevant what signature that callback must have. Contrast this with e.g. a filter that smells like something that takes another mesh as first argument (after `self`). And you can't just look at the docstring because---well, I guess you could look at the docstring if it were complete, with `Returns`. So yeah, I guess it's similar to other typing issues.

pyvista/plotting/plotter.py
MatthewFlamm: The GitHub diff isn't doing us any favor in reviewing this... I agree on the first point about redundancy, but the change here is simply getting rid of the logic for `mesh._textures` and de-indenting everything that used to be in the `else` block.  I think it is fine to keep the ancillary changes to a minimum, i.e. don't address the redudancy, or clean up here.

I don't understand the second point as the second `scalars=None` line doesn't exist in the original logic, at least not within this section of the code that I can see.

adeak: Ah, you're right, I got confused by the convoluted diff. I agree that just unindenting is the way to go. I thought exactly that this wasn't what was happening.

CONTRIBUTING.rst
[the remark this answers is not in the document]
banesullivan: Nope, it wouldn't work as is

pyvista/__init__.py
banesullivan: They may not look "plotty" but they do require the plotting module. Why?:

- `examples`: Use of `pyvista.Texture`
- `demos`: functions like `plot_glyphs()`
- `ext`: Eh, this could be fine since no modules are imported in the `__init__.py` but the modules each require te plotting module
- `trame`: requires the plotting module (mainly for type annotations)
- `utilities`: this is deprecated but can point to the plotting module if accessed

.github/workflows/testing-and-deployment.yml
banesullivan: I'm not concerned about this. I'm going to advocate that we not add additional CI to test/verify this unless it later proves to be a problem. If installing the GL dependencies breaks the `core` module, I think we have bigger problems in VTK, not PyVista

pyvista/plotting/themes.py
banesullivan: Indeed I removed these annotations to remove the plotting dependency and a circular import. I'll see if I can just add this back in now that `theme.py` is under `pyvista/plotting/`

banesullivan: See 2fe00b4e3328520d94e1bcdda607e58e6a5063e3

tests/namespace/test_errors_namespace.py
banesullivan: Done in 09611295faa4fbb6def1554958274ae979cc9411

pyvista/__init__.py
banesullivan: The other part. These were not previously imported at the top-level namespace. The usage is always: `from pyvista.examples import ...`, never import pyvista; pyvista.examples...`

adeak: I looked into one of the modules when I posted this comment and it seemed to me that we don't need plotting to _import_ the modules. Which concurs with your summary (thanks for that!). So do I understand correctly that you're mostly motivated by the future plan to separate the plotting functionality into a package, which is why you also "axe" functionality that needs GL at runtime?

banesullivan: > mostly motivated by the future plan to separate the plotting functionality into a package

Correct -- this may not happen now or in the next few months but it's important to have this level of decoupling from now on so that we have the option/ability to ship as separate packages down the road.

tests/namespace/test_errors_namespace.py
```diff
@@ -0,0 +1,45 @@
...
+        'PyVistaDeprecationWarning',
+        'PyVistaEfficiencyWarning',
+        'PyVistaFutureWarning',
+        'PyVistaPipelineError',
+        'VTKVersionError',
+    ],
+)
+def test_core_errors_namespace(name):
+    with pytest.warns(
+        PyVistaDeprecationWarning, match='now imported as: `from pyvista.core.errors import'
```
adeak, line 25: Purity nitpick:
```suggestion
        PyVistaDeprecationWarning, match=r'now imported as: `from pyvista\.core\.errors import'
```
And should we not match with the name while we're at it?
```suggestion
        PyVistaDeprecationWarning, match=rf'now imported as: `from pyvista\.core\.errors import {name}`\.'
```

tests/namespace/test_errors_namespace.py
```diff
@@ -0,0 +1,45 @@
...
+@pytest.mark.parametrize(
+    'name',
+    [
+        'InvalidCameraError',
+        'RenderWindowUnavailable',
+    ],
+)
+def test_plotting_errors_namespace(name):
+    with pytest.warns(
+        PyVistaDeprecationWarning, match='now imported as: `from pyvista.plotting.errors import'
```
adeak, line 41:
```suggestion
        PyVistaDeprecationWarning, match=r'now imported as: `from pyvista\.plotting\.errors import'
```

adeak: \*shakes fist at `black`\*

banesullivan: Fixed!

adeak (approved):
LGTM, thanks for the great work @banesullivan.

akaszynski:
In one of my projects I'm getting:

```
    @pytest.mark.skipif(skip_plotting, reason="Requires active X Server")
    def test_plot_sector_numbering(rotor):
>       rotor.plot_sector_numbering()

tests/test_rotor.py:136: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
femorph/rotor.py:484: in plot_sector_numbering
    plotter = pv.Plotter(off_screen=off_screen)
/home/alex/source/pyvista/pyvista/__init__.py:87: in __getattr__
    import pyvista.plotting
shibokensupport/signature/loader.py:61: in feature_imported
    ???
shibokensupport/feature.py:137: in feature_imported
    ???
shibokensupport/feature.py:148: in _mod_uses_pyside
    ???
/usr/lib/python3.10/inspect.py:1147: in getsource
    lines, lnum = getsourcelines(object)
/usr/lib/python3.10/inspect.py:1128: in getsourcelines
    object = unwrap(object)
/usr/lib/python3.10/inspect.py:639: in unwrap
    while _is_wrapper(func):
/usr/lib/python3.10/inspect.py:630: in _is_wrapper
    return hasattr(f, '__wrapped__')
/home/alex/source/pyvista/pyvista/__init__.py:87: in __getattr__
    import pyvista.plotting
shibokensupport/signature/loader.py:61: in feature_imported
    ???
E   RecursionError: maximum recursion depth exceeded
!!! Recursion detected (same locals & position)
```

When running just `test_rotor.py`, the error doesn't occur. I'm still looking into it but I wanted to post initial findings first to halt the merge until I dig into this more.

akaszynski:
Got it!

MRE:

```py
import pyvista as pv

pv.Report()
pv.Plotter()
```

```
~/source/pyvista/pyvista/__init__.py in __getattr__(name)
     85         return importlib.import_module(f'pyvista.{name}')
     86 
---> 87     import pyvista.plotting
     88 
     89     try:

RecursionError: maximum recursion depth exceeded
```

banesullivan:
Ah. Uh-oh. I'm signing off for the evening but will fix tomorrow if you don't beat me to it

adeak:
I can't repro on Debian Python 3.9, I suspect the Python version.

Looks like `inspect.getattr_static` ends up calling `__getattr__()` anyway. I wonder if we can replace the `try: inspect.getattr_static(...)` with a lookup in `vars(pyvista.plotting)` to get around this.

adeak:
> Looks like `inspect.getattr_static` ends up calling `__getattr__()` anyway. I wonder if we can replace the `try: inspect.getattr_static(...)` with a lookup in `vars(pyvista.plotting)` to get around this.

Actually, may be not, I think I misread the traceback. It might be `import pyvista.plotting` triggering `__getattr__`... it's late here so I'll stop guessing :smile:

akaszynski (approved):
Tested offline; LGTM. Thanks @banesullivan!

PR #4637, opened 2023-07-07
More convenient PolyData face interface

### LINKED ISSUE #2330: Transferring PolyData Faces into a Numpy Array
A fast way to transfer the face data, as stored in the Polydata object, into a Numpy object array containing the indexes of all the points of each simplex.

This should be a fairly simple for loop in C, I would add it myself, but I A) don't know C yet, and B) don't know how to integrate C and Python. However, the pseudocode in Python could look something like this:

```python
simplexes = []
index = 0
while index < face_array.shape[0]:
    number_of_points = face_array[index]
    points = face_array[index+1:index+1+number_of_points]
    simplexes.append(points)```

### PR DESCRIPTION
**Pull request by darikg**
# Introduction
Currently, the structure of `PolyData.faces` parallels the original representation of vtkCellArray, where offsets and connectivity were stored in an interleaved fashion. In [November 2019](https://discourse.vtk.org/t/upcoming-changes-to-vtkcellarray/2066) `vtkCellArray` was changed to store offsets and connectivity seperately. These separate arrays are currently exposed only privately via the PolyData properties `_connectivity_array` and `_offset_array`. 

This PR adds a new settable and mutatable property `PolyData.regular_faces`, which, for meshes where all faces are the same size, returns a `n_faces` x `face_size` array. It also adds an alternate `PolyData` constructor in the form of the static method `PolyData.from_regular_faces(points, faces, deep)`. IME, constructing a mesh from a set of points and triangles is by far the most common `PolyData` construction, and using a separate method is cleaner than extending the already heavily-overloaded default constructor. 

# Justification
## Intuition and ease of use
The interleaved format is (IMHO) non-intuitive, leading to a number of beginner issues (e.g. #1097, #966), as well as line noise where either the faces array is written out explicitly with cell sizes, or a column of cell sizes is prepended to the face array.

E.g. IMO `faces = [[0, 1, 2], [1, 3, 2], [0, 2, 3], [0, 3, 1]]` is much easier to read than either `faces = [3, 0, 1, 2, 3, 1, 3, 2, 3, 0, 2, 3, 3, 0, 3, 1]` or

```python
faces = [[0, 1, 2], [1, 3, 2], [0, 2, 3], [0, 3, 1]]
faces = np.stack([np.full(3, (len(faces), 1)), faces], axis=1)
```
  
## Performance

perf

<details>
  <summary>Code for performance profiling</summary>
  
  ```python
    import timeit
    from collections import defaultdict
    from pyvista import PolyData, Sphere
    from matplotlib import pyplot as plt

    def time_stuff(mesh, fn_factories, n_sub, n_times, n_repeat):
        times = defaultdict(list)
        n_faces = []
        for i in range(n_sub):
            n_faces.append(mesh.n_faces)
            for j, (name, fn_factory) in enumerate(fn_factories.items()):
                fn = fn_factory(mesh)
                t = timeit.Timer(fn).repeat(repeat=n_repeat, number=n_times)
                times[name].append(min(t) / n_times * 1000)
            mesh = mesh.subdivide(1)
        return n_faces, times

    def plot_times(title, mesh, fn_factories, n_sub, n_times, n_repeat):
        n_faces, times = time_stuff(Sphere().triangulate(), fn_factories, n_sub=n_sub, n_repeat=n_repeat, n_times=n_times)
        for name, fn_times in times.items():
            plt.plot(n_faces, fn_times, marker='x', label=name)
            plt.title(title)
            plt.xscale('log')
            plt.yscale('log')
            plt.xlabel('# faces')
            plt.legend()
        return times

    mesh = Sphere().triangulate()

    def write_default_faces(mesh):
        points, faces = mesh.points, mesh.faces.reshape(-1, 4)
        return lambda: PolyData(points, faces=faces)

    def write_regular_faces(mesh):
        points, faces = mesh.points, mesh.regular_faces
        return lambda: PolyData.from_regular_faces(points, faces)

    def read_default_faces(mesh):
        return lambda: mesh.faces

    def read_regular_faces(mesh):
        return lambda: mesh.regular_faces

    n_sub = 5
    n_repeat = 7
    n_times = 1000

    plt.figure(figsize=(10, 5))
    fn_factories=dict(faces=write_default_faces, regular_faces=write_regular_faces)
    ax_read = plt.subplot(1, 2, 1)
    read_times = plot_times("Write", mesh, fn_factories, n_sub, n_times, n_repeat)
    plt.ylabel('Time (ms)')

    plt.subplot(1, 2, 2, sharey=ax_read)
    fn_factories=dict(faces=read_default_faces, regular_faces=read_regular_faces)
    write_times = plot_times("Read", mesh, fn_factories, n_sub, n_times, n_repeat)
  ```
</details>

When setting faces from a regular array there is a mild (at best) performance benefit to constructing the underlying CellArray from the two arrays. Note that I'm comparing here against the condition of setting the original faces as a `n_faces` x `face_size + 1` array. Setting faces as a 1-dimensional array instead is exceptionally slow, as already noted in the [comments](https://github.com/pyvista/pyvista/blob/447a0905217dd53aff2ba38bd9a2624d3aa15ce6/pyvista/core/cell.py#L493), which requires scanning the entire array to determine the number of faces. 

The read time when the face size is known is O(1) since it's just a view into the connectivity array. When the face size is unknown, there are two options: 1) greedily check only the size of the first face, and assume the rest are the same size (also O(1)) or 2) scan the the entire offsets array and validate that the faces are all the same size, which gives approximately no benefit over the status quo. I've gone with option 1 here.

## Mutation
I don't think I've ever felt the need for this, but the new face array can be mutated in place, e.g. `mesh.regular_faces[0, 0] = 9`, whereas `PolyData.faces` is set read-only because this isn't possible.

# Implementation details
Most of the actual work was done on `CellArray`, exposing public properties `connectivity_array` and `offset_array`, as well as new alternate constructors `CellArray.from_arrays(offsets, connectivity, deep)` and `CellArray.from_regular_cells(cells, deep)`, in the hopes this will be useful beyond just PolyData faces. I added the `@tkCellArray.override` decorator to `CellArray` so e.g. `PolyData.GetPolys()` returns a `CellArray` instead of a `vtkCellArray`. This greatly simplified the glue code in `PolyData`, but perhaps the override decorator is too new/experimental to rely on.

# Relevant issues
Closes #2330

# Open questions

## Spelling
`regular_faces` vs `face_array`, etc.

## Extensions 
A similar interface could be added to other cases of equisized cells, particularly `UnstructuredGrid.regular_cells` and `PolyData.regular_lines`, as well as `PolyData.regular_strips`.

If my argument for the intuitiveness holds water, it might also be worth generalizing to the more general case where faces aren't equisized, adding a `PolyData.irregular_faces` property returning `list[list[int]]`.
 
# Potential downsides
While I think this is more intuitive for the majority case of triangle faces, this does introduce some potentional for confusion if users think `faces` and `regular_faces` are disjoint properties, whereas they actually modify/replace the same underlying vtkCellArray. In addition this also introduces the question of which interface the examples should use where both are applicable.

### DIFF
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -536,10 +536,7 @@ class CellArray(_vtk.vtkCellArray):
         """Set the offsets and connectivity arrays."""
         offsets = numpy_to_idarr(offsets, deep=deep)
         connectivity = numpy_to_idarr(connectivity, deep=deep)
-        try:
-            self.SetData(offsets, connectivity)
-        except AttributeError:  # pragma: no cover
-            raise VTKVersionError('vtkCellArray.SetData implemented in VTK 9 or newer.')
+        self.SetData(offsets, connectivity)
 
         # Because vtkCellArray doesn't take ownership of the arrays, it's possible for them to get
         # garbage collected. Keep a reference to them for safety
@@ -617,10 +614,7 @@ class CellArray(_vtk.vtkCellArray):
 
 def _get_connectivity_array(cellarr: _vtk.vtkCellArray):
     """Return the array with the point ids that define the cells' connectivity."""
-    try:
-        return _vtk.vtk_to_numpy(cellarr.GetConnectivityArray())
-    except AttributeError:  # pragma: no cover
-        raise VTKVersionError('Connectivity array implemented in VTK 9 or newer.')
+    return _vtk.vtk_to_numpy(cellarr.GetConnectivityArray())
 
 
 def _get_offset_array(cellarr: _vtk.vtkCellArray):
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -6,7 +6,6 @@ from typing import List, Tuple, cast
 import numpy as np
 
 import pyvista
-from pyvista.core.errors import VTKVersionError
 
 from . import _vtk_core as _vtk
 from .celltype import CellType
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -492,6 +492,8 @@ class CellArray(_vtk.vtkCellArray):
 
     def __init__(self, cells=None, n_cells=None, deep=False):
         """Initialize a vtkCellArray."""
+        self.__offsets = None
+        self.__connectivity = None
         if cells is not None:
             self._set_cells(cells, n_cells, deep)
 
@@ -508,6 +510,7 @@ class CellArray(_vtk.vtkCellArray):
                 n_cells = cells.shape[0]
 
         self.SetCells(n_cells, vtk_idarr)
+        self.__offsets = self.__connectivity = None
 
     @property
     def cells(self):
@@ -529,6 +532,23 @@ class CellArray(_vtk.vtkCellArray):
         """Return the array used to store cell offsets."""
         return _get_offset_array(self)
 
+    def _set_data(self, offsets, connectivity, deep=False):
+        """Set the offsets and connectivity arrays."""
+        offsets = numpy_to_idarr(offsets, deep=deep)
+        connectivity = numpy_to_idarr(connectivity, deep=deep)
+        try:
+            self.SetData(offsets, connectivity)
+        except AttributeError:  # pragma: no cover
+            raise VTKVersionError('vtkCellArray.SetData implemented in VTK 9 or newer.')
+
+        if not deep:
+            # Because vtkCellArray doesn't take ownership of the arrays, it's possible for them to get
+            # garbage collected. Keep a reference to them for safety
+            self.__offsets = offsets
+            self.__connectivity = connectivity
+        else:
+            self.__offsets = self.__connectivity = None
+
     @staticmethod
     def from_arrays(offsets, connectivity, deep=False) -> CellArray:
         """Construct a vtkCellArray from offsets and connectivity arrays.
@@ -546,9 +566,7 @@ class CellArray(_vtk.vtkCellArray):
             Default is ``False``.
         """
         cellarr = CellArray()
-        offsets = numpy_to_idarr(offsets, deep=deep)
-        connectivity = numpy_to_idarr(connectivity, deep=deep)
-        cellarr.SetData(offsets, connectivity)
+        cellarr._set_data(offsets, connectivity, deep=deep)
         return cellarr
 
     @property
@@ -569,7 +587,7 @@ class CellArray(_vtk.vtkCellArray):
         return _get_regular_cells(self)
 
     @classmethod
-    def from_regular_cells(cls, cells, deep=True):
+    def from_regular_cells(cls, cells, deep=False):
         """Construct a ``CellArray`` from a (n_cells, cell_size) array of cell indices.
 
         Parameters
@@ -587,18 +605,10 @@ class CellArray(_vtk.vtkCellArray):
         """
         cells = np.asarray(cells, dtype=pyvista.ID_TYPE)
         connectivity = numpy_to_idarr(cells, deep=deep)
-
         n_cells, cell_size = cells.shape
-        offsets = cell_size * np.arange(n_cells + 1, dtype=pyvista.ID_TYPE)
-        offsets = numpy_to_idarr(
-            offsets, deep=True
-        )  # Since we're creating offsets on the fly here, force deep copy
+        offsets = numpy_to_idarr(cell_size * np.arange(n_cells + 1, dtype=pyvista.ID_TYPE))
         cellarr = cls()
-        try:
-            cellarr.SetData(offsets, connectivity)
-        except AttributeError:  # pragma: no cover
-            raise VTKVersionError('vtkCellArray.SetData implemented in VTK 9 or newer.')
-
+        cellarr._set_data(offsets, connectivity, deep=deep)
         return cellarr
 
 
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -478,6 +478,7 @@ class CellArray(_vtk.vtkCellArray):
     Examples
     --------
     Create a cell array containing two triangles from the traditional interleaved format
+
     >>> from pyvista.core.cell import CellArray
     >>> cellarr = CellArray([3, 0, 1, 2, 3, 3, 4, 5])
 
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -533,7 +533,7 @@ class CellArray(_vtk.vtkCellArray):
 
     @staticmethod
     def from_arrays(offsets, connectivity, deep=False) -> 'CellArray':
-        """Construct a vtkCellArray from offsets and connectivity arrays
+        """Construct a vtkCellArray from offsets and connectivity arrays.
 
         Parameters
         ----------
@@ -555,7 +555,7 @@ class CellArray(_vtk.vtkCellArray):
 
     @property
     def regular_cells(self) -> np.ndarray:
-        """Return an array of shape (n_cells, cell_size) of point indices when all faces have the same size
+        """Return an array of shape (n_cells, cell_size) of point indices when all faces have the same size.
 
         Returns
         -------
@@ -563,7 +563,7 @@ class CellArray(_vtk.vtkCellArray):
             (n_cells, cell_size) Array of face indices
 
         Notes
-        --------
+        -----
         This property does not validate that the cells are all
         actually the same size. If they're not, this property may either
         raise a `ValueError` or silently return an incorrect array.
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -462,6 +462,7 @@ class Cell(_vtk.vtkGenericCell, DataObject):
         return type(self)(self, deep=deep)
 
 
+@_vtk.vtkCellArray.override
 class CellArray(_vtk.vtkCellArray):
     """pyvista wrapping of vtkCellArray.
 
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 from typing import List, Tuple, cast
 
 import numpy as np
+from pyvista.core.errors import VTKVersionError
 
 import pyvista
 
@@ -477,10 +478,15 @@ class CellArray(_vtk.vtkCellArray):
 
     Examples
     --------
-    Create a cell array containing two triangles.
-
+    Create a cell array containing two triangles from the traditional interleaved format
     >>> from pyvista.core.cell import CellArray
     >>> cellarr = CellArray([3, 0, 1, 2, 3, 3, 4, 5])
+
+    Create a cell array containing two triangles from separate offsets and connectivity arrays
+    >>> from pyvista.core.cell import CellArray
+    >>> offsets = [0, 3, 6]
+    >>> connectivity = [0, 1, 2, 3, 4, 5]
+    >>> cellarr = CellArray.from_arrays(offsets, connectivity)
     """
 
     def __init__(self, cells=None, n_cells=None, deep=False):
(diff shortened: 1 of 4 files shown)


pyvista/core/pointset.py
```diff
@@ -818,6 +827,76 @@ def faces(self, faces):
...
+        """
+        return _get_regular_cells(self.GetPolys())
+
+    @regular_faces.setter
+    def regular_faces(self, faces: Union[List[List[int]], np.ndarray]):
+        """Set the face cells from an (n_faces, face_size) array."""
+        self.faces = CellArray.from_regular_cells(faces, deep=True)
+
+    @staticmethod
+    def from_regular_faces(points, faces: Union[np.ndarray, Sequence[Sequence[int]]], deep=True) -> 'PolyData':
```
adeak, line 868: I don't have time for a review right now, but this should be a classmethod:
```py
@classmethod
def from regular_faces(cls, ...):
    poly = cls()
    ...
    return poly
```

darikg: you're right, thanks!

darikg:
documentation build fails because of a single sphinx warning:
>>> /home/runner/work/pyvista/pyvista/doc/source/api/core/_autosummary/pyvista.PolyData.regular_faces.rst:8: ERROR: Unexpected indentation.

I cannot for the life of me see the wrong indentation, I must be missing something extremely obvious

pyvista/core/pointset.py
```diff
@@ -818,6 +827,77 @@ def faces(self, faces):
...
+            Whether to deep copy the faces array into vtkCellArray connectivity data. Default `True`.
+
+        Returns
+        -------
+        pyvista.PolyData
+            The newly constructed mesh.
+
+        Examples
+        --------
+        Construct a tetrahedron from four triangles
```
adeak, line 890:
```suggestion
        Construct a tetrahedron from four triangles

```

pyvista/core/pointset.py
```diff
@@ -818,6 +827,77 @@ def faces(self, faces):
...
+
+        Notes
+        -----
+        This property does not validate that the mesh's faces are all
+        actually the same size. If they're not, this property may either
+        raise a `ValueError` or silently return an incorrect array.
+
+        Examples
+        --------
+        Get the face array of a tetrahedron as a 4x3 array
```
adeak, line 851:
```suggestion
        Get the face array of a tetrahedron as a 4x3 array

```

pyvista/core/cell.py
```diff
@@ -476,10 +477,15 @@ class CellArray(_vtk.vtkCellArray):
 
     Examples
     --------
-    Create a cell array containing two triangles.
-
+    Create a cell array containing two triangles from the traditional interleaved format
```
adeak, line 480:
```suggestion
    Create a cell array containing two triangles from the traditional interleaved format

```

pyvista/core/cell.py
```diff
@@ -476,10 +477,15 @@ class CellArray(_vtk.vtkCellArray):
...
     >>> from pyvista.core.cell import CellArray
     >>> cellarr = CellArray([3, 0, 1, 2, 3, 3, 4, 5])
+
+    Create a cell array containing two triangles from separate offsets and connectivity arrays
```
adeak, line 484:
```suggestion
    Create a cell array containing two triangles from separate offsets and connectivity arrays

```

adeak (commented):
Based on this broken render I think this is what's breaking the doc build. I don't understand why `from_regular_faces()` didn't trigger the same error though...

Screenshot from 2023-07-10 00-08-10

pyvista/core/cell.py
darikg: ohhh I see. Is there a way to see the rendered docs from the CI build or did you have to build locally?

adeak: I built it locally to be sure, but you can download successful doc build artifacts via the "checks" tab on the PR -> "build documentation" step in the bottom left -> "docs-build" https://github.com/pyvista/pyvista/suites/14167884129/artifacts/794084378
Come to think of it, I think this artifact is only there when the build is successful...

And FYI you can gather multiple edit suggestions into a single commit with the "add suggestion to batch" button :) https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes

pyvista/core/cell.py
```diff
@@ -510,3 +518,113 @@ def cells(self):
...
+    @classmethod
+    def from_regular_cells(cls, cells, deep=True):
+        """Constract a ``CellArray`` from a (n_cells, cell_size) array of cell indices.
+
+        Parameters
+        ----------
+        cells : numpy.ndarray or list[list[int]]
+            Cell array of shape (n_cells, cell_size) where all cells have the same size `cell_size`
+
+        deep : bool, default: False
```
MatthewFlamm, line 580: Does this work without copied data?  That is if I pass in `cells` and then modify `cells` will it work as intended?  Is there a problem if `cells` changes size, which can happen for a `list[list[int]]`?

pyvista/core/pointset.py
```diff
@@ -818,6 +827,79 @@ def faces(self, faces):
...
+        Parameters
+        ----------
+        points : numpy.ndarray, sequence[sequence[float]]
+            A (n_points, 3) array of points.
+            
+        faces : numpy.ndarray or sequence[sequence[int]]
+            A (n_faces, face_size) array of face indices. For a triangle mesh, face_size = 3.
+
+        deep : bool, optional
+            Whether to deep copy the faces array into vtkCellArray connectivity data. Default `True`.
```
MatthewFlamm, line 882: I think our default for deep copying is usually False, but is there a reason you chose True here?  We should be consistent unless there is a reason.

pyvista/core/pointset.py
```diff
@@ -891,15 +973,12 @@ def __sub__(self, cutting_mesh):
     @property
     def _offset_array(self):
         """Return the array used to store cell offsets."""
-        return _vtk.vtk_to_numpy(self.GetPolys().GetOffsetsArray())
+        return _get_offset_array(self.GetPolys())
```
MatthewFlamm, line 976: It looks like this `_offset_array` and `_connectivity_array` only are for faces.  I'm wondering if they should be `_faces_offset_array` and so on.  This comment is only really relevant if we will support strips or verts in this PR.

MatthewFlamm (commented):
I like this, especially with the performance enhancement.  I think we should also have it for `UnstructuredGrid` in a separate PR.

One comment for `PolyData`, also forward looking. We currently don't support looking at all of the cells together.  I believe this is possible in `vtkPolyData`, but we don't expose it here yet.  If in the future we support this for `regular_cells`, we could consider checking that at most 1 of `n_faces`, `n_verts`, `n_lines`, and `n_strips` are non-zero. We could provide an error or warning otherwise as it would be likely that the user is doing something unintended in that case.

pyvista/core/pointset.py
```diff
@@ -818,6 +827,79 @@ def faces(self, faces):
...
+        array([[0, 1, 2],
+               [1, 3, 2],
+               [0, 2, 3],
+               [0, 3, 1]])
+
+        """
+        return _get_regular_cells(self.GetPolys())
+
+    @regular_faces.setter
+    def regular_faces(self, faces: Union[List[List[int]], np.ndarray]):
```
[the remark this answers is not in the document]
MatthewFlamm: Forgot `regular_lines`, which could be more useful than `regular_strips`.

pyvista/core/cell.py
```diff
@@ -510,3 +518,113 @@ def cells(self):
...
+        -----
+        This property does not validate that the cells are all
+        actually the same size. If they're not, this property may either
+        raise a `ValueError` or silently return an incorrect array.
+        """
+        return _get_regular_cells(self)
+
+    @classmethod
+    def from_regular_cells(cls, cells, deep=True):
+        """Constract a ``CellArray`` from a (n_cells, cell_size) array of cell indices.
```
MatthewFlamm, line 573:
```suggestion
        """Construct a ``CellArray`` from a (n_cells, cell_size) array of cell indices.
```

pyvista/core/pointset.py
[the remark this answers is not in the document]
darikg: I haven't really used strips much. Do you mean that `strips=[4, a, b, c, d]` implies `regular_strips=[[a, b, c], [b, c, d]]` ? That might be less useful just because you lose the benefit of a cheap view into the underlying connectivity array.

Agreed on `regular_lines`, which I think I would use especially for `mesh.extract_all_edges().regular_lines` to return `n_edges x 2` array.

pyvista/core/pointset.py
darikg: The issue here I'm not sure exactly how to handle is that you really can only set deep=False when the cells array is already a C-contiguous numpy array. If you pass in list[list[int]] it gets converted by `numpy.asarray` inside of `numpy_to_idarr`, and since nothing is holding a reference to that array it can get GC'd leaving the vtkCellArray data total garbage. My logic was to default to the safe option, and require deep=False to be opt-in. Maybe better type-checking and warning issuing would be smarter.

pyvista/core/cell.py
darikg: No, see below.

darikg:
Sorry, what does that do, exactly?

MatthewFlamm:
> Sorry, what does that do, exactly?

We have pre-commit setup so that if you install it locally like `pip install pre-commit && pre-commit install` everytime you do `git commit ...` it will run a bunch of tools like black, isort, etc.  and if any of those make changes it will abort the pending commit.  Then you can check and stage those changes and recommit.  

If you do the comment I gave you above in this PR, it will happen automatically in this PR (it will take the changes from black, etc. and commit those here) and you will have to pull those changes onto your local working repo.

If you don't want to do this using pre-commit, you will have to run each command manually, e.g. `black`, to get the CI to be happy.  I personally recommend pre-commit though.

pyvista/core/pointset.py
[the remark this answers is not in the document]
MatthewFlamm: strips would actually be `strips=[4, a, b, c, d]` and `regular_strips=[[a, b, c, d]]`.  More illustrative would be `strips=[4, a, b, c, d, 4, e, f, g, h]` and `regular_strips=[[a, b, c, d]. [e, f, g, h]`.  triangle strips are a weird cell type as they have multiple triangles defined in one cell.  It is sort of similar to `PolyLine` and `PolyVertex`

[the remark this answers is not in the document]
darikg: Oh got it. Yeah that would be an easy extension; I've it for a followup PR once this gets nailed down

darikg:
pre-commit.ci autofix

pyvista/core/pointset.py
MatthewFlamm: Is the vtkCellArray total garbage?  Or only that you no longer have a shallow copied version?  If you deep copy, the object will also be generated inside one of the nested functions.  If it is only a problem that deep copying always occurs for lists, we can put in a note in the docstring that lists will always be deep copied.  We do this conversion to numpy array from list while claiming it is shallow copied in other locations, e.g. when setting point/cell data, so it seems strange to me that it would be a specific problem for vtkCellArray.

MatthewFlamm: This is bad indeed.

One thought is that we could add a `_set_data(offsets, connectivity)` method that stores private attributes of each, so that we maintain the references on our instance object.  I don't think this will lead to duplicating memory.

darikg: oh yeah, that's way easier than what I was thinking

darikg:
pre-commit.ci autofix

darikg:
can't remember if y'all prefer merge or rebase so i'll just leave it here for now

adeak:
We squash merge PRs so there's no need to rebase (I even prefer merging when I review) :)

pyvista/core/pointset.py
```diff
@@ -818,6 +827,81 @@ def faces(self, faces):
...
```
MatthewFlamm, line 865: To be consistent with `from_regular_faces` below:
```suggestion
    def regular_faces(self, faces: Union[np.ndarray, Sequence[Sequence[int]]]):
```

MatthewFlamm (commented):
My last comment.  Otherwise, this LGTM.

pyvista/core/pointset.py
```diff
@@ -5,15 +5,15 @@
 import os
 import pathlib
 from textwrap import dedent
-from typing import Sequence, Tuple, Union
+from typing import List, Sequence, Tuple, Union
```
MatthewFlamm, line 8:
```suggestion
from typing import Sequence, Tuple, Union
```

MatthewFlamm (approved):
This is great.  Thanks for addressing all my comments.

PR #4649, opened 2023-07-12
Add Report class GPU test

tkoyama010:
### Overview

~I notice the system info code in the issue form is not working in pyvista>=0.40 .~
Add Report class test

relate to #4486

### Details

- None


MatthewFlamm:
The else block looks the same as the try block.  Not sure what the difference is.  This is really awkward especially for new users trying to ask a question. Is there a way to instead fix the code to get it to be (at least temporarily) backwards compatible.

tkoyama010:
@MatthewFlamm Thanks. I need more coffee in morning 😅
I was aware that report function is only a debugging function and therefore does not need to be marked as deprecated. Do you think we need deprecation time?

MatthewFlamm:
I also don't want the search for perfect/ideal to stand in the way of fixing an emergent issue.  So I'm fine with updating the issue link if this works while we can discuss how this should work long term.

banesullivan:
The following should always work:

```py
import pyvista as pv
print(pv.Report())
```

If that's not working, we need to fix that directly rather than changing the usage pattern. 

Could someone report the issue they are having with this?

banesullivan:
Ah I see this is coming from https://github.com/bjlittle/geovista/issues/375

tkoyama010:
I switch this branch to bug and revert the form.

tkoyama010:
I checked the development version and found that the following code is working.
```python
import pyvista as pv

print(pv.Report())
```

MatthewFlamm:
Do we have a test for this behavior?

tkoyama010:
@MatthewFlamm Good point. Let's test everything!

tkoyama010:
pre-commit.ci autofix

akaszynski:
Recommend ~copying~ reusing scooby's `test_report`:
https://github.com/banesullivan/scooby/blob/092c890c6525ac85b4c30b9c41679fbc083ea37f/tests/test_scooby.py#L25-L50

tkoyama010:
Thanks! Sorry I missed the test when it existed, I learned a bit about `scooby` and added the results to the test.

tests/core/test_utilities.py
```diff
@@ -330,8 +330,10 @@ def test_voxelize_throws_point_cloud(hexbeam):
 def test_report():
     report = pyvista.Report(gpu=True)
     assert report is not None
+    assert not "GPU Details : None" in report.__repr__()
```
tkoyama010, line 333:
```suggestion
    assert "GPU Details : None" not in report.__repr__()
```

adeak (approved):
LGTM, thanks.

On a not really related note, `pv.Report()` seems to haunt my interactive shell:
```py
>>> import pyvista as pv
>>> pv.Report() is not None
True
>>> pv.Report() is not None
>>> print('wait, where's my output???')
PyVista error in handling VTK error message:
'WARNING:root:WARNING:root:WARNING:root:WARNING:root:WARNING:root:WARNING:root:WARNING:root:WARNING:r' ... (3926 characters)
>>> ':|'
>>> 
```
But the test runs fine, and even if I import the test function and run it manually it succeeds...

Actually, I only noticed the above syntax error due to syntax highlighting here in github. That explains the error lines (for some value of "explains"). If I remove the syntax error I get the output.

But the issue is that this happened originally with no error:
```py
>>> import pyvista as pv
>>> pv.Report() is not None
True
>>> pv.Report() is not None
>>> pv.Report() is None
>>> ':|'
>>> 
```

**Edit**: I see the same error as in the first snippet when I exit the interactive shell in the second snippet... so the root cause is the same, perhaps the second call to `pv.Report()` would raise and this ties something up in knots.

bjlittle:
> LGTM, thanks.
> 
> On a not really related note, `pv.Report()` seems to haunt my interactive shell:
> 
> ```python
> >>> import pyvista as pv
> >>> pv.Report() is not None
> True
> >>> pv.Report() is not None
> >>> print('wait, where's my output???')
> PyVista error in handling VTK error message:
> 'WARNING:root:WARNING:root:WARNING:root:WARNING:root:WARNING:root:WARNING:root:WARNING:root:WARNING:r' ... (3926 characters)
> >>> ':|'
> >>> 
> ```
> 
> But the test runs fine, and even if I import the test function and run it manually it succeeds...
> 
> Actually, I only noticed the above syntax error due to syntax highlighting here in github. That explains the error lines (for some value of "explains"). If I remove the syntax error I get the output.
> 
> But the issue is that this happened originally with no error:
> 
> ```python
> >>> import pyvista as pv
> >>> pv.Report() is not None
> True
> >>> pv.Report() is not None
> >>> pv.Report() is None
> >>> ':|'
> >>> 
> ```
> 
> **Edit**: I see the same error as in the first snippet when I exit the interactive shell in the second snippet... so the root cause is the same, perhaps the second call to `pv.Report()` would raise and this ties something up in knots.

@tkoyama010 and @adeak I had noticed this behaviour myself, and narrowed it down to this [line](https://github.com/pyvista/pyvista/blob/main/pyvista/report.py#L219), but I can't explain why.

Certainly by not doing `from pyvista.plotting.tools import check_math_text_support` and then `extra_meta.append(('MathText Support', check_math_text_support()))` as part of the report makes the problem go away.

I'd love to know why and also if anyone else can replicate this behaviour 🤔

PR #4662, opened 2023-07-16
Move examples modules to pyvista-examples package

### LINKED ISSUE #4526: It would be better as a standalone `pyvista-planets` package
I'm not sure about this `planets` module overall. IMO, it would be better as a standalone `pyvista-planets` or `pyvista-examples` package. As it stands, this seems like a lot of code for examples that only a small minority of users leverage.

I agree the naming here isn't ideal. I do not want this PR to be the place to refine this and recommend opening an issue for follow-up work and discuss stripping this module out into a new `pyvista-examples` repo for instance

_Originally posted by @banesullivan in https://github.com/pyvista/pyvista/pull/4515#discussion_r1225539684_

tkoyama010:
### Overview

Move examples modules to pyvista-examples package.

resolves #4526.

### Details

- `pyvista-examples` is now released at https://pypi.org/project/pyvista-examples .


tkoyama010:
See https://github.com/pyvista/pyvista/issues/4526#issuecomment-1692786072 .

PR #4674, opened 2023-07-19
Fix `Plane` and `Sphere` normal/direction; fix `axis_rotation_angle` numerical error

### LINKED ISSUE #4655: "pv.Plane()" returns a plane with incorrect normal direction

[`pv.Plane`](https://docs.pyvista.org/version/0.40/api/utilities/_autosummary/pyvista.Plane.html) returns a plane whose normal is in the opposite direction from the specified input direction. E.g. setting `direction=(0,0,1)` results in a plane with a normal `(0,0,-1)`

Create plane and plot its normal with the following code:
``` python
import pyvista as pv

# create plane
center = (0,0,0)
expected_normal = (0,0,1)
plane = pv.Plane(center=center, direction=expected_normal)

# get actual plane normal at center
center_point_id = plane.find_closest_point(center)
actual_normal = plane.point_normals[center_point_id]
arrow = pv.Arrow(start=center, direction=actual_normal)

# plot plane and its actual normal
plot = pv.Plotter()
plot.add_mesh(plane, style='wireframe')
plot.add_mesh(arrow)
plot.show_axes()
plot.show()

print(f"Expected normal: {expected_normal}")
print(f"Actual normal: {actual_normal}")
```

Output is below. The normal is pointing in the wrong direction.
```
Expected normal: (0, 0, 1)
Actual normal: [-1.110223e-16  0.000000e+00 -1.000000e+00]
```
image

The issue seems to stem from these [two lines of code](https://github.com/pyvista/pyvista/blob/0519eab351f0f4c65af4ed6b02efaa8a983717bf/pyvista/core/utilities/geometric_objects.py#L437C1-L438C39), which are supposed to correctly translate the mesh to `center` and rotate the plane to align it with `direction`:

```python
surf.rotate_y(-90, inplace=True)
translate(surf, center, direction)
```
Not sure if the bug is with the `translate` method  itself or if the overall approach used for doing this is incorrect.
Either way, a solution is to use the native VTK functions `SetNormal()` and `SetCenter()` instead to apply the necessary translation and rotation as shown in `Plane()` below:

```python
from vtk import vtkPlaneSource
import pyvista as pv

def Plane(
    center=(0.0, 0.0, 0.0),
    direction=(0.0, 0.0, 1.0),
    i_size=1,
    j_size=1,
    i_resolution=10,
    j_resolution=10,
    ):

    planeSource = vtkPlaneSource()
    planeSource.SetCenter(*center)
    planeSource.SetNormal(*direction)
    planeSource.SetXResolution(i_resolution)
    planeSource.SetYResolution(j_resolution)
    planeSource.Update()

    surf = pv.wrap(planeSource.GetOutput())

    surf.points[:, 0] *= i_size
    surf.points[:, 1] *= j_size
    return surf

# create plane
center = (0,0,0)
expected_normal = (0,0,1)
plane = Plane(center=center, direction=expected_normal)

# get actual plane normal at center
center_point_id = plane.find_closest_point(center)
actual_normal = plane.point_normals[center_point_id]
arrow = pv.Arrow(start=center, direction=actual_normal)

# plot plane and its actual normal
plot = pv.Plotter()
plot.add_mesh(plane, style='wireframe')
plot.add_mesh(arrow)
plot.show_axes()
plot.show()

print(f"Expected normal: {expected_normal}")
print(f"Actual normal: {actual_normal}")
```

Running the above code gives the correct output:
```
Expected normal: (0, 0, 1)
Actual normal: [0. 0. 1.]
```

image

### System Information

```shell
OS : Windows
            CPU(s) : 12
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.8 GiB
       Environment : IPython
        GPU Vendor : Intel
  MathText Support : False
  bit (AMD64)]
           pyvista : 0.40.1
               vtk : 9.2.20230323
             numpy : 1.24.2
        matplotlib : 3.7.1
            scooby : 0.7.1
             pooch : v1.7.0
           imageio : 2.26.1
           IPython : 8.12.0
        ipywidgets : 8.0.6
             scipy : 1.10.1
              tqdm : 4.65.0
            meshio : 5.3.4
      nest_asyncio : 1.5.6
```

### Screenshots

_No response_

user27182:
Resolves #4655

- [x] #4702 is merged.


pyvista/core/utilities/geometric_objects.py
```diff
@@ -434,7 +434,7 @@ def Plane(
 
     surf.points[:, 0] *= i_size
     surf.points[:, 1] *= j_size
-    surf.rotate_y(-90, inplace=True)
+    surf.rotate_y(90, inplace=True)
```
tkoyama010, line 437: Thanks for this. I also checked the other functions in the same file and found that the `Sphere` function also had a `-90` degree rotation. Please correct the angle of the Sphere function as well if it is not affected.

tkoyama010 (changes requested):

pyvista/core/utilities/geometric_objects.py
user27182: Ok. I spent some time evaluating your suggestion, and I agree with your proposed changes for `Sphere`. I also think a separate issue should be raised to clarify the meaning of `direction` in the `Sphere` docstring. Details below:

In the `Plane` case it was clear that the normal was pointing in the wrong direction because the meaning of normal or direction for a plane is clear. However, for a sphere, this is less clear. What should direction mean in this case? The docs for [vtkSphereSource](https://vtk.org/doc/nightly/html/classvtkSphereSource.html#ae25246b3c5f43545670efc7f13b10f2d) does not have an explicit `direction` parameter but it does state that "latitude angle 0 is at north pole."  Also note that the docs for [Sphere](https://docs.pyvista.org/version/0.40/api/utilities/_autosummary/pyvista.Sphere.html) define `direction` as "Direction the top of the sphere points to in [x, y, z]." So, in this case, if +Z points north and -Z points south, we can say that the sphere's direction points from its south pole to its north pole.

So, the question that should be asked is, do the current transformations  in `Sphere` correctly orient the sphere so that its direction points from south pole to north pole where the north pole is at latitude 0. To test this, we can run:
``` python
import pyvista
north = pyvista.Sphere(start_phi=0, end_phi=90)
south = pyvista.Sphere(start_phi=90, end_phi=180)
p = pyvista.Plotter()
p.add_mesh(north, show_edges=True, color='red', label='north')
p.add_mesh(south, show_edges=True, color='blue', label='south')
p.add_axes()
p.add_legend()
p.show()
```
Where the northern hemisphere is defined as theta ranging from 0-90, and the southern hemisphere is defined by theta ranging from 90-180.

With `rotate_y(-90)` we get:
image
Which shows us that the southern hemisphere is above the northern hemisphere, and therefore its actual direction is (0, 0, -1). However, the direction should indeed be (0, 0, 1).

With `rotate_y(90)`, we get the correct orientation:
image

The `Sphere` docs could be improved by:

- clarifying that latitude angle 0 is at north pole when setting `phi`
- clarifying that `direction` is defined from south pole to north pole ("Direction the top of the sphere points" is ambiguous in my view)


@tkoyama010 do you still want me to update this PR and change `rotate_y(-90)` to `rotate_y(90)` in `Sphere`? If so, do you also want me to clarify the documentation for `Sphere` as I suggest above? Or, would you rather keep this PR as-is, and keep it restricted to `Plane` fixes only, and then separately raise a new issue to fix the problems with `Sphere`.

tkoyama010: @user27182 Thank you. The relevant part was added by [this commit five years ago](https://github.com/pyvista/pyvista/commit/5254e94f5d87c3549f751e680569c07df5956e28), and there is no rational reason why the NS should not be under the Z-axis, so please add the change. Please clarify the specification by the test code and change the title of the pull request.

user27182: I pushes these changes to `Sphere`, however this caused `test_find_cells_intersecting_line` to fail since that test made an assumption about the cell configuration based on the previous (wrong) orientation of `Sphere`.

Therefore, I pushed an additional commit to fix this test and remove its dependency on `Sphere`.

user27182: @tkoyama010 since `Sphere` is used by so many examples and test cases, fixing the direction bug for `Sphere` seems like a massive undertaking that I am not currently unable to commit to. The changes and effort needed to fix this is likely compounded by #4691.

How would you like to proceed? Should we keep this PR for Plane only just to get that fix merged and open a separate issue specifically to address the `Sphere` direction bug?

tkoyama010: @user27182
Let's merge only `Plane` case and discuss Sphere in an issue. I'm sorry, but could you please revert the Sphere part and change the title back as well?

user27182: Ok. I had a closer look at some of the tests that were failing and was able to resolve a handful of failed doctests and a few regular test cases. So, on second thought I decided to push further and continue working to resolve the problems caused by fixing `Sphere` and will also help resolve #4691 with this PR too.

However, I will need some assistance making sense of the current failed checks, especially plotting / image regression testing related failures. @tkoyama010 if you can provide some guidance on a possible cause or a path forward that would be appreciated.

tkoyama010: @user27182 
There seems to be an error in the regression testing of the output image, could you please refer to [Notes Regarding Image Regression Testing](https://docs.pyvista.org/version/stable/extras/developer_notes.html#notes-regarding-image-regression-testing) to check how much difference is being made to the image?

user27182: Some of the image regression testing errors were caused by the test cases using `Plane` or `Sphere` objects. Since the cached mages were generated using the incorrect orientation of these objects, the new images generated during testing would appear flipped or mirrored when compared to what was expected, and hence the tests would fail. I have pushed new changes to these tests so that the tests now "invert" the direction of any `Plane` or `Sphere` objects (e.g. use `Sphere(direction=(0,0,-1)` instead of `Sphere()`) such that the generated images should now appear identical to the expected cached images. This resolved many failed tests locally on my machine, let's see how the CI tests fare.

tests/core/test_geometric_objects.py
```diff
@@ -35,11 +35,57 @@ def test_sphere():
     assert np.any(surf.points)
     assert np.any(surf.faces)
 
+    # test direction points from south pole to north pole
```
tkoyama010, line 38: Let's define a new test function to make it easier to understand what you are doing by the name of the test.

tests/core/test_geometric_objects.py
```diff
@@ -35,11 +35,57 @@ def test_sphere():
...
+    assert np.all(quadrant4.points[:, 1] <= 0 + atol)  # -Y
+
 
 def test_plane():
     surf = pyvista.Plane()
     assert np.any(surf.points)
     assert np.any(surf.faces)
+    assert np.array_equal(surf.center, (0, 0, 0))
+
+    # test direction
```
tkoyama010, line 77: Same comment on test_sphere.

pyvista/core/utilities/geometric_objects.py
```diff
@@ -332,16 +340,16 @@ def Sphere(
         ``start_phi`` to ``end_phi``).
 
     start_theta : float, default: 0.0
-        Starting longitude angle.
+        Starting longitude angle in degrees ``[0, 360]``.
```
tkoyama010, line 343: Thank you for the addition. Please also add error handling for values other than this.

tkoyama010 (changes requested):

pyvista/core/utilities/geometric_objects.py
user27182: I tried to search for existing conventions on how angles are currently handled/validated in pyvista. E.g. [grid_from_sph_coords](https://github.com/pyvista/pyvista/blob/c7b29f06333d9b3427ba9ede4f30eddeef8e5c9f/pyvista/core/utilities/features.py#L123) notes the valid ranges but does no input validation, and the `prevent_seam` parameter of [texture_map_to_sphere](https://github.com/pyvista/pyvista/blob/c7b29f06333d9b3427ba9ede4f30eddeef8e5c9f/pyvista/core/filters/data_set.py#L1977C13-L1977C13) notes a range but also does no input validation. For what it's worth, `vtkSphereSource` itself will gladly accept values outside of the specified ranges and still instantiate without any issue. E.g. `start_phi=-90` will still work, and so will setting a negative radius for that matter.

So, adding error handling for `Sphere` could make it less consistent with other code and also add restrictions that are not present in the underlying VTK class. Do you still want me to add this ? Or maybe the behaviour outside these ranges could be better explained in the docstring?

tkoyama010: Thank you. The lack of uniformity in input validation is an issue. Let's add it after we made it in our project. I will withdraw this proposal in this Pull Request.

tests/core/test_geometric_objects.py
tkoyama010:
```suggestion
def test_sphere_direction_points():
    # from south pole to north pole
```

tests/core/test_geometric_objects.py
```diff
@@ -35,11 +35,57 @@ def test_sphere():
...
 
+    # test direction points from south pole to north pole
+    directions = np.vstack((np.eye(3), -np.eye(3)))
+    for expected in directions:
+        north_pole = pyvista.Sphere(direction=expected, start_phi=0, end_phi=0).points[0]
+        south_pole = pyvista.Sphere(direction=expected, start_phi=180, end_phi=180).points[0]
+        actual = north_pole - south_pole
+        assert np.array_equal(expected, actual)
+
+    # test phi
```
tkoyama010, line 46:
```suggestion
def test_sphere_phi():
```

tests/core/test_geometric_objects.py
tkoyama010:
```suggestion
def test_plane_direction():
```

tests/core/test_geometric_objects.py
```diff
@@ -35,11 +35,57 @@ def test_sphere():
...
+    assert np.array_equal(surf.center, (0, 0, 0))
+
+    # test direction
+    directions = np.vstack((np.eye(3), -np.eye(3)))
+    for expected in directions:
+        surf = pyvista.Plane(direction=expected)
+        actual = surf.point_normals[0]
+        assert np.array_equal(actual, expected)
+
+    # test size
```
tkoyama010, line 84:
```suggestion
def test_plane_size():
```

tests/core/test_utilities.py
```diff
@@ -598,6 +592,41 @@ def test_axis_angle_rotation():
     with pytest.raises(ValueError):
         transformations.axis_angle_rotation([0, 0, 0], angle)
 
+    # exact transformation cases (90, 180, 270 degrees)
+    # test that applying a rotation many times (up to a multiple of 360)
+    # yields the exact same input
```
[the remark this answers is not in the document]
tkoyama010:
```suggestion
def test_axis_angle_rotation_many_times():
    # yields the exact same input
```

tests/core/test_geometric_objects.py
```diff
@@ -35,11 +35,58 @@ def test_sphere():
...
+
+def test_sphere_phi():
+    atol = 1e-16
+    north_hemisphere = pyvista.Sphere(start_phi=0, end_phi=90)
+    assert np.all(north_hemisphere.points[:, 2] >= 0 - atol)  # north is above XY plane
+    south_hemisphere = pyvista.Sphere(start_phi=90, end_phi=180)
+    assert np.all(south_hemisphere.points[:, 2] <= 0 + atol)  # south is below XY plane
+
+def test_sphere_theta():
+    quadrant1 = pyvista.Sphere(start_theta=0, end_theta=90)
```
tkoyama010, line 55:
```suggestion
    atol = 1e-16

    quadrant1 = pyvista.Sphere(start_theta=0, end_theta=90)
```

tkoyama010:
pre-commit.ci autofix

tests/core/test_utilities.py
```diff
@@ -599,6 +593,41 @@ def test_axis_angle_rotation():
...
+def test_axis_angle_rotation_many_times():
+    # yields the exact same input
+    def _apply_transformation_n_times(points_in, axis_in, angle_in, n):
+        points_out = points_in.copy()
+        trans = transformations.axis_angle_rotation(axis_in, angle_in)
+        for _ in range(n):
+            points_out = transformations.apply_transformation_to_points(trans, points_out)
+        return points_out
+
+    points = np.eye(3)
```
tkoyama010, line 605:
```suggestion
    points = np.eye(3)

```

tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

tests/core/test_utilities.py
```diff
@@ -599,6 +593,17 @@ def test_axis_angle_rotation():
         transformations.axis_angle_rotation([0, 0, 0], angle)
 
 
+@pytest.mark.parametrize("axis,angle,times", [([1, 0, 0], 90, 4), ([1, 0, 0], 180, 2), ([1, 0, 0], 270, 4), ([0, 1, 0], 90, 4), ([0, 1, 0], 180, 2), ([0, 1, 0], 270, 4), ([0, 0, 1], 90, 4), ([0, 0, 1], 180, 2), ([0, 0, 1] , 270, 4)])
+def test_axis_angle_rotation_many_times():
```
tkoyama010, line 597:
```suggestion
def test_axis_angle_rotation_many_times(axis,angle,times):
```

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):
It is good to go after #4702 is merged.

tests/core/test_geometric_objects.py
```diff
@@ -36,10 +36,66 @@ def test_sphere():
...
+    north_pole = pyvista.Sphere(direction=expected, start_phi=0, end_phi=0).points[0]
+    south_pole = pyvista.Sphere(direction=expected, start_phi=180, end_phi=180).points[0]
+    actual = north_pole - south_pole
+    assert np.array_equal(expected, actual)
+
+
+def test_sphere_phi():
+    atol = 1e-16
+    north_hemisphere = pyvista.Sphere(start_phi=0, end_phi=90)
+    assert np.all(north_hemisphere.points[:, 2] >= 0 - atol)  # north is above XY plane
```
tkoyama010, line 53:
```suggestion
    assert np.all(north_hemisphere.points[:, 2] >= -atol)  # north is above XY plane
```

tests/core/test_geometric_objects.py
```diff
@@ -36,10 +36,66 @@ def test_sphere():
...
+    south_hemisphere = pyvista.Sphere(start_phi=90, end_phi=180)
+    assert np.all(south_hemisphere.points[:, 2] <= 0 + atol)  # south is below XY plane
```
tkoyama010, line 55:
```suggestion
    assert np.all(south_hemisphere.points[:, 2] <= atol)  # south is below XY plane
```

tkoyama010 (approved):

PR #4677, opened 2023-07-19
Update implicit_distance docs to include sign info

### LINKED ISSUE #4676: Improve  `compute_implicit_distance` documentation to explain sign

Improve docs for [compute_implicit_distance](https://docs.pyvista.org/version/0.40/api/core/_autosummary/pyvista.DataSetFilters.compute_implicit_distance.html) by explaining the meaning of the distance sign. As stated in [vtkImplicitPolyDataDistance](https://vtk.org/doc/nightly/html/classvtkImplicitPolyDataDistance.html#details): 
- "Points interior to the geometry have a negative distance, points on the exterior have a positive distance, and points on the input [vtkPolyData](https://vtk.org/doc/nightly/html/classvtkPolyData.html) have a distance of zero."

However, this is not mentioned in the docstring.

The lack of documentation on this is made worse by #4655, since the example given shows the following figure:

Which implies that the distance from the plane becomes more negative as we move in the positive Z direction. In reality, however, based on the quoted VTK reference material above, we should expect that the the distance becomes more positive as we move along the positive Z direction since the plane's normals should be pointing up.

As a result, when using `compute_implicit_distance` it's unclear what the sign of the distance is supposed to represent.

Note that when #4655 is resolved, the correct figure for the provided docstring example is:
image

### Links to source code.

_No response_

### Pseudocode or Screenshots

_No response_

user27182:
Resolves #4676

The docstring example was also extended to more clearly demonstrate the meaning of the sign. Note that, due #4655, the wording of the example is currently inconsistent with the figure. I.e. [this line](https://github.com/user27182/pyvista/blob/b30e7b1cc087cba2f4bd17f379dfe42e43fde521/pyvista/core/filters/data_set.py#L441) states "distances above the  plane are positive, and distances below the plane are negative", but the figure implies that it's the other way around.

So, the wording of the docstring is correct but the figure will be wrong until #4655 is resolved.


pyvista/core/filters/data_set.py
```diff
@@ -403,6 +403,11 @@ def compute_implicit_distance(self, surface, inplace=False):
         nodes of this mesh to a given surface. This distance will be
         added as a point array called ``'implicit_distance'``.
 
+        Nodes of this mesh which are interior to the input surface
+        geometry have a negative distance, and nodes on the exterior
+        have a positive distance. Nodes which intersect the input
+        surface have a distance of zero.
```
tkoyama010, line 409:
```suggestion
        Nodes of this mesh which are interior to the input surface
        geometry have a negative distance, and nodes on the exterior
        have a positive distance. Nodes which intersect the input
        surface has a distance of zero.
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

PR #4702, opened 2023-07-25
Fix Vale errors in document

tkoyama010:
### Overview

ref #4674

### Details

- None


tkoyama010:
I will merge immediately after approval as it is a document lint error and has an impact on other PullRequests.

akaszynski (approved):

PR #4710, opened 2023-07-26
Test all cell types in test_cell

MatthewFlamm:
### Overview

While looking into adressing #4707 , I noticed that not all cell types are tested. This PR aims to clean up the testing and a following PR will address #4707.

Issue #4135, opened 2023-03-15
Structured grid dimensions should be immutable


akaszynski:
Attempting to modify the dimensions of a `StructuredGrid` corrupts it:

```py
import pyvista
import numpy as np
xrng = np.arange(-10, 10, 1, dtype=np.float32)
yrng = np.arange(-10, 10, 2, dtype=np.float32)
zrng = np.arange(-10, 10, 5, dtype=np.float32)
x, y, z = np.meshgrid(xrng, yrng, zrng)
grid = pyvista.StructuredGrid(x, y, z)
grid.dimensions = (3, 4, 5)
grid.plot()
```
![tmp3](https://user-images.githubusercontent.com/11981631/225195301-fc9624c4-5684-4b3f-9e6b-06a175910140.png)

Whereas for a `UniformGrid`, it does not:

```py
ugrid = pyvista.UniformGrid(dimensions=(4, 5, 6))
ugrid.dimensions = (3, 3, 3)
ugrid.plot(show_edges=True)
```

![tmp2](https://user-images.githubusercontent.com/11981631/225195235-940e75f0-cb81-4a82-9aa7-0a524ef4662d.png)

banesullivan:
The dimensions should be mutable! These are a property that can change, and there are lots of examples out there where the dimensions get re-set. The problem in your example is that your new dimensions are wrong!

Further, the example with a uniform grid is completely different and not a fair comparison. The uniform grid is defined by its dimensions. Changing the dimensions produces an entirely different mesh. Whereas changing the dimensions of the structured grid means something totally different: how to interpret the structured points.

If this property were made immutable, it will break tests, examples, and lots of downstream code.

One such example is later on in: https://docs.pyvista.org/examples/00-load/create-structured-surface.html

```py
mesh = pv.StructuredGrid()
# Set the coordinates from the numpy array
mesh.points = points
# set the dimensions
mesh.dimensions = [29, 32, 1]
```

And at least one other:

https://docs.pyvista.org/examples/00-load/create-surface-draped.html#sphx-glr-examples-00-load-create-surface-draped-py

Further, I often am given structured points from external software where the dimensions are unknown, and I need to be able to toy with the dimensions (basically guess) until the mesh looks right.

Additionally, there are definitely scenarios where a point set may be valid under multiple dimensions -- I don't have a good example off hand, but I'm pretty sure I've seen this before

adeak:
[...]

PR #4712, opened 2023-07-26
Fix polyhedron cast to unstructured grid

### LINKED ISSUE #4707: Plotting cell with non-planar face throws exception

Trying to plot a cell with a non-planar face throws a `ValueError`:

```
>>> get_cell(cellid).plot()
  File "/Users/dion/codes/sandbox-dion/blocky/test_real.py", line 13, in <module>
    mesh.get_cell(cell).plot()
  File "/Users/dion/.virtualenvs/science/lib/python3.10/site-packages/pyvista/core/cell.py", line 142, in plot
    self.cast_to_unstructured_grid().plot(**kwargs)
  File "/Users/dion/.virtualenvs/science/lib/python3.10/site-packages/pyvista/core/cell.py", line 177, in cast_to_unstructured_grid
    return pyvista.UnstructuredGrid(*args)
  File "/Users/dion/.virtualenvs/science/lib/python3.10/site-packages/pyvista/core/pointset.py", line 1312, in __init__
    self._check_for_consistency()
  File "/Users/dion/.virtualenvs/science/lib/python3.10/site-packages/pyvista/core/pointset.py", line 1464, in _check_for_consistency
    raise ValueError(
ValueError: Number of cell types (1) must match the number of cells 0)
```

```python
cell_type = CellType.POLYHEDRON
cell_points = [[ 1.03791904  0.30976132 -0.34050989]
 [ 1.03685999  0.30017006 -0.32543549]
 [ 1.03750801  0.3124826  -0.32517678]
 [ 1.02476108  0.29915988 -0.32668063]
 [ 1.02512503  0.31213984 -0.32542327]
 [ 1.02444935  0.30510914 -0.33619115]
 [ 1.02858233  0.30887827 -0.34128037]]
cell_point_ids = [0, 1, 2, 3, 4, 5, 6]
face_point_ids = [[3, 1, 2, 4], [6, 0, 1, 3, 5], [3, 4, 6, 5], [2, 1, 0], [6, 4, 2, 0]]

# construct cell from the above, couldn't figure out how to do this with the public API ...
# cell = Cell(...)

cell.plot()
```

### System Information

```shell
  Date: Wed Jul 26 12:23:41 2023 CEST

                OS : Darwin
            CPU(s) : 10
           Machine : arm64
      Architecture : 64bit
               RAM : 16.0 GiB
       Environment : Python
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M1 Pro

  Python 3.10.12 (main, Jun  7 2023, 00:38:32) [Clang 14.0.3
  (clang-1403.0.22.14.1)]

           pyvista : 0.38.5
               vtk : 9.2.6
             numpy : 1.25.0
           imageio : 2.27.0
            scooby : 0.7.1
             pooch : v1.7.0
        matplotlib : 3.7.1
           IPython : 8.8.0
           cmocean : 2.0
        ipywidgets : 7.7.0
             scipy : 1.10.1
              tqdm : 4.64.0
            meshio : 5.0.0
             trame : 2.5.0
      trame_client : 2.9.4
      trame_server : 2.11.4
         trame_vtk : 2.5.4
      nest_asyncio : 1.5.5
```

### Screenshots

Cell plotted with ParaView:

Screenshot 2023-07-26 at 12 24 14

MatthewFlamm:
### Overview

Closes #4707.

Cast to unstructured grid needs to correctly construct the polyhedron cell.

### Details

This PR also includes changes from #4710 .  I opened 2 PRs to keep the changes separate and will keep this one in draft until the other is decided upon.


pyvista/core/cell.py
```diff
@@ -176,8 +176,16 @@ def cast_to_unstructured_grid(self) -> pyvista.UnstructuredGrid:
...
 
         """
+        if self.type == CellType.POLYHEDRON:
+            # construct from faces
+            cell_ids = [self.n_faces]
+            for face in self.faces:
+                cell_ids.extend([len(face.point_ids)] + face.point_ids)
+            cell_ids.insert(0, len(cell_ids))
+        else:
+            cell_ids = [len(self.point_ids)] + list(range(len(self.point_ids)))
```
MatthewFlamm, line 186: I suspect this should also be:
```suggestion
            cell_ids = [len(self.point_ids)] + self.point_ids
```

pyvista/core/cell.py
```diff
@@ -416,6 +424,21 @@ def _get_attrs(self):
...
+        if not isinstance(other, Cell):
+            return False
+        if not self.type == other.type:
+            return False
+        if not np.array_equal(self.points, other.points):
+            return False
+        if not np.array_equal(self.point_ids, other.point_ids):
+            return False
+        if self.type == CellType.POLYHEDRON:
+            return all(zip(self.faces, other.faces))
```
MatthewFlamm, line 439:
```suggestion
            # POLYHEDRON has a nested structure that depends on the faces
            return all(zip(self.faces, other.faces))
```

MatthewFlamm:
There is something wrong with the current implementation of cell that causes problems during copy and is now caught by implementing a better `__eq__` method for `Cell`.  I will remove the addition of `__eq__` here as I'm not sure how to fix the bug yet.  I will test some other way.

pyvista/core/cell.py
adeak: `all(zip(self.faces, other.faces))` looks very surprising here. Did you not mean to compare pairs of values instead? And is it guaranteed at this point that the faces attributes have the same lengths? Otherwise we'd need `itertools.zip_longest()` or an explicit check.

adeak: Actually, yeah, since `zip` always yields 2-tuples, and in the empty case we get a vacuous truth, this line always returned `True` I think.

MatthewFlamm: I think you are right, but this check failed on checking point_ids when shallow copying cells.  This is a rabbit hole I didn't want to go down for this PR

adeak: I just wasn't sure whether you were going to reuse this somewhere else :) All good.

MatthewFlamm:
cc @dionhaefner in case you want to test this branch to see if it fixes your plotting issue.

pyvista/core/cell.py
MatthewFlamm: I _think_ for all other cell types, there should be no difference, but I'm not 100% sure.  @beroda do you have an opinion on this implementation?

beroda: This suggestion does not work, as the connectivity is not correct since the `point_ids` are coming from the original mesh the cell was extracted from. Therefore, it can refers to `point_ids` that do not exist when declaring the `ugrid`.

With your suggestion, this simple example segfaults:
```python
from pyvista.examples import load_hexbeam

cell = load_hexbeam().get_cell(0)
cell.plot()
```

When looking closer at the connectivity array, it becomes clear:

**suggestion:**
```python
ugrid = cell.cast_to_unstructured_grid()
ugrid.cells
```
`array([ 8,  0,  2,  8,  7, 27, 36, 90, 81])`
```python
ugrid.n_points
```
`8`

**current implementation:**
```python
ugrid = cell.cast_to_unstructured_grid()
ugrid.cells
```
`array([8, 0, 1, 2, 3, 4, 5, 6, 7])`
```python
ugrid.n_points
```
`8`

Let me know if I was not clear.

MatthewFlamm: Hmmm this will need some testing with Polyhedron.  Thanks for the insight.  I had to change it to get it to work for that since it has a nested structure.

beroda: Moreover, I am not sure whether the current implementation is correct for `polyhedron`.
This example does not show the cell correctly:
```python
import pyvista as pv
import numpy as np
from pyvista import CellType


ptIds = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
nodes = np.asarray(
    [
        [0.01, 0.01, 0.01],
        [0, 0.01, 0.01],
        [0, 0.0, 0.01],
        [0.01, 0.0, 0.01],
        [0.01, 0.0, 0.0],
        [0, 0.0, 0.0],
        [0, 0.01, 0.0],
        [0.01, 0.01, 0.0],
        [0.01, 0.01, 0.02],
        [0.00, 0.01, 0.02],
    ]
)

# polyhedron format [NElements, NFaces, Face1NPoints, Face1Point1, Face1Point2..., Face1PointN, FaceNNPoints,...]]
# in polygons and polyhedrons, face order/points order matters
polyhedron = np.asarray(
    [24, 5, 4, 2, 3, 4, 5, 4, 5, 6, 7, 4, 3, 4, 7, 3, 3, 5, 6, 2, 4, 6, 7, 3, 2]
)
polygon = np.asarray([4, 1, 9, 8, 0])

cells = np.hstack((polygon, polyhedron))
cell_type = [CellType.POLYGON, CellType.POLYHEDRON]
grid = pv.UnstructuredGrid(cells, cell_type, nodes)
grid.get_cell(1).plot()
```

I guess the implementation might be trickier than expected ...

beroda: This should work IMO (sorry for the screenshot, it is late here :) ).
image

MatthewFlamm: I think you are right.  I will test it.  These Polyhedrons are quite tricky.

MatthewFlamm: This also shows we really need an `__eq__` method for cells, mostly for testing.

beroda: I am not sure whether an `__eq__`would solve this issue.

IMO, the issues we are facing with are related to the fact that you can construct an `UnstructuredGrid` with an erroneous connectivity (ie. the `point_ids` in the connectivity array can be higher than the number of points).

Correct me if I am wrong, but I think that no connectivity checks are performed in the `pyvista.UnstructuredGrid` constructor, nor for any `Dataset` with cells support. I think we should implement connectivity checks to catch such errors as early as possible (currently, it crashes only when plotting, which can confuse users on the failure reasons).

MatthewFlamm: We have several tests that do something like

```py
cell = grid.get_cell(0)
new_cell = cell.some_method()
assert cell == new_cell
```

This includes `cast_to_unstructured_grid`. This means that the current equality is not useful.  I tried to implement this for this PR and the `copy` method tests also started failing.

Failing early would also be good.

MatthewFlamm: But, a cell outside its mesh may no longer be equal to the original cell in the fullest sense in PyVista anyway.  Right now, `Cell` inherits from `DataObject`, which has a suspiciously `DataSet` focused `__eq__` method.

adeak: > I think we should implement connectivity checks to catch such errors as early as possible (currently, it crashes only when plotting, which can confuse users on the failure reasons). If I am correct on this, I think I can find some time to propose a PR.

Not exactly the same context, but similar (structured grid consistency) discussion was [here](https://github.com/pyvista/pyvista/issues/4135#issuecomment-1470354274) with an explicit preference not to be stricter than VTK. Just FYI for "prior art".

beroda: > This includes `cast_to_unstructured_grid`. This means that the current equality is not useful. I tried to implement this for this PR and the `copy` method tests also started failing.
> 
> Failing early would also be good.
I agree, this might need some investigation in another PR.

> Right now, `Cell` inherits from `DataObject`, which has a suspiciously `DataSet` focused `__eq__` method.
I think it confims the fact that `Cell` should have its own `__eq__`.

beroda: > Not exactly the same context, but similar (structured grid consistency) discussion was [here](https://github.com/pyvista/pyvista/issues/4135#issuecomment-1470354274) with an explicit preference not to be stricter than VTK. Just FYI for "prior art".
It makes sense. My point was that for a user who is not a VTK expert and encounters such segfaults, it can be very tricky to understand where the problem comes from.
But I guess that a if a user tries to create the connectivity by himself, he followed some examples and is more susceptible to be aware of such issues as he gained VTK knowledge.

beroda:
> There is something wrong with the current implementation of cell that causes problems during copy and is now caught by implementing a better `__eq__` method for `Cell`. I will remove the addition of `__eq__` here as I'm not sure how to fix the bug yet. I will test some other way.

I tested locally your `__eq__`. The tests fails on the `point_ids` comparison when the copy is `shallow`.
But I am afraid this might be on VTK's side, since the following code shows that the `ShallowCopy` does not work as expected (ie. I would expect a cell shallow copy to have the same point ids, and not zeros).

```python
from pyvista.examples import cells as example_cells
import vtk

m = example_cells.Tetrahedron()
cell = vtk.vtkGenericCell()
m.GetCell(0, cell)

ids = cell.GetPointIds()
ids = [ids.GetId(i) for i in range(ids.GetNumberOfIds())]

cell2 = vtk.vtkGenericCell()
cell2.SetCellType(cell.GetCellType())
cell2.ShallowCopy(cell)
ids2 = cell2.GetPointIds()
ids2 = [ids2.GetId(i) for i in range(ids2.GetNumberOfIds())]

ids, ids2
```
([0, 1, 2, 3], [0, 0, 0, 0])

Maybe we should not allow users to make `shallow` copies, but I agree this might be too strict ...

dionhaefner:
Confirmed fixed, thanks

MatthewFlamm:
I started an issue to keep track the `copy` and `__eq__` discussion. I propose these to be out of scope for this PR, although having a working `__eq__` method would enhance the confidence in the changes of this PR.

beroda:
LGTM. I approve your changes, thank you for this PR.

tkoyama010 (approved):

PR #4726, opened 2023-08-01
For exact checks use `is` / `is not`

tkoyama010:
### Overview

Fix error #4724

### Details

- None

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -4906,7 +4906,7 @@ class DataSetFilters:
         _update_alg(append_filter, progress_bar, 'Merging')
         merged = _get_output(append_filter)
         if inplace:
-            if type(self) == type(merged):
+            if isinstance(self, type(merged)):
                 self.deep_copy(merged)
                 return self
             else:
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -4906,7 +4906,7 @@ class DataSetFilters:
         _update_alg(append_filter, progress_bar, 'Merging')
         merged = _get_output(append_filter)
         if inplace:
-            if isinstance(self, type(merged)):
+            if type(self) is type(merged):
                 self.deep_copy(merged)
                 return self
             else:
--- a/tests/core/test_datasetattributes.py
+++ b/tests/core/test_datasetattributes.py
@@ -436,7 +436,7 @@ def test_length_should_be_0_on_clear(insert_arange_narray):
 def test_keys_should_be_strings(insert_arange_narray):
     dsa, sample_array = insert_arange_narray
     for name in dsa.keys():
-        assert isinstance(name, str)
+        assert type(name) is str
 
 
 def test_key_should_exist(insert_arange_narray):
@@ -447,7 +447,7 @@ def test_key_should_exist(insert_arange_narray):
 def test_values_should_be_pyvista_ndarrays(insert_arange_narray):
     dsa, sample_array = insert_arange_narray
     for arr in dsa.values():
-        assert isinstance(arr, pyvista.pyvista_ndarray)
+        assert type(arr) is pyvista.pyvista_ndarray
 
 
 def test_value_should_exist(insert_arange_narray):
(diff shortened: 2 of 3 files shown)


pyvista/core/filters/data_set.py
```diff
@@ -4906,7 +4906,7 @@ def merge(
         _update_alg(append_filter, progress_bar, 'Merging')
         merged = _get_output(append_filter)
         if inplace:
-            if type(self) == type(merged):
+            if isinstance(self, type(merged)):
                 self.deep_copy(merged)
```
MatthewFlamm, line 4910: The original logic seems to be that the types have to be the same.  The error message in this case says to use `is` comparison in this case.  I haven't checked whether this change makes more sense though.

tkoyama010: Thanks for your comment, I am looking into it as I don't understand what the difference is between the two methods.

tkoyama010: Ah, I see. [isinstance](https://docs.python.org/3/library/functions.html#isinstance) includes the inherited class as well, so a comparison by is is appropriate, as you say.

pyvista/core/filters/data_set.py
```diff
@@ -4906,7 +4906,7 @@ def merge(
         _update_alg(append_filter, progress_bar, 'Merging')
         merged = _get_output(append_filter)
         if inplace:
-            if type(self) == type(merged):
+            if isinstance(self, type(merged)):
```
tkoyama010, line 4909:
```suggestion
            if type(self) is type(merged):
```

tests/core/test_datasetattributes.py
```diff
@@ -436,7 +436,7 @@ def test_length_should_be_0_on_clear(insert_arange_narray):
 def test_keys_should_be_strings(insert_arange_narray):
     dsa, sample_array = insert_arange_narray
     for name in dsa.keys():
-        assert type(name) == str
+        assert isinstance(name, str)
```
tkoyama010, line 439:
```suggestion
        assert type(name) is str
```

tests/core/test_datasetattributes.py
```diff
@@ -447,7 +447,7 @@ def test_key_should_exist(insert_arange_narray):
 def test_values_should_be_pyvista_ndarrays(insert_arange_narray):
     dsa, sample_array = insert_arange_narray
     for arr in dsa.values():
-        assert type(arr) == pyvista.pyvista_ndarray
+        assert isinstance(arr, pyvista.pyvista_ndarray)
```
tkoyama010, line 450:
```suggestion
        assert type(arr) is pyvista.pyvista_ndarray
```

MatthewFlamm (approved):
LGTM

bjlittle:
Wouldn't pyupgrade catch these types of changes in syntax usage?

tkoyama010:
@bjlittle 
None of the [implemented features](https://github.com/asottile/pyupgrade#implemented-features) of pyupgrade seem to apply :)

bjlittle:
> @bjlittle 
> None of the [implemented features](https://github.com/asottile/pyupgrade#implemented-features) of pyupgrade seem to apply :)

And ruff?

MatthewFlamm:
It was _caught_ by `pyupgrade`, but I'm not sure if it would autofix given the likelihood that the code usage wasn't precise.

adeak (approved):
LGTM too, thanks @tkoyama010.

For anyone wondering, it took me a while to understand why this even matters. In most cases there's no difference, because if you do `class Foo: pass` then `Foo.__eq__ is object.__eq__`, in other words equality comparison falls back to object identity.

So you can only make `type(foo) == type(bar)` different from `type(foo) is type(bar)` if you use a metaclass. Here's a silly example in case anyone's interested:
```py
class SillyMeta(type):
    def __eq__(cls, other):
        return True

class Foo:
    pass

class Bar(metaclass=SillyMeta):
    pass

print(Bar == Foo)  # True
```

tkoyama010:
@adeak Thanks. It's an art of bug.

Issue #4632, opened 2023-07-06
DynamicScraper fails to find .vtksz files when doc uses source layout


greschd:
### Describe the bug, what's wrong, and what you expected.

When the documentation uses "source layout" as follows, the `DynamicScraper` fails to find the generated `.vtksz` files.

```
├── doc
│   ├── Makefile
│   ├── make.bat
│   └── source
│       ├── conf.py
│       └── index.rst
```

While building, the following warning appears. The build subsequently does not show the dynamic image.
```
WARNING: Source file examples/images/sphx_glr_clipping_001.vtksz does not exist.
```

<details>
<summary>Expand for full output</summary>

```powershell
 .\make.bat html

Running Sphinx v6.2.1
making output directory... done
Using pandoc version: 2.18 to convert rst text blocks to markdown for .ipynb files
WARNING: html_static_path entry '_static' does not exist
generating gallery...
generating gallery for examples... [100%] clipping.py
computation time summary:
    - ..\..\examples\perlin.py:          1.64 sec   0.0 MB
    - ..\..\examples\clipping.py:        1.43 sec   0.0 MB
    - ..\..\examples\extract-edges.py:   1.24 sec   0.0 MB
building [mo]: targets for 0 po files that are out of date
writing output... 
building [html]: targets for 1 source files that are out of date
updating environment: [new config] 6 added, 0 changed, 0 removed
reading sources... [100%] index
WARNING: Source file examples/images/sphx_glr_clipping_001.vtksz does not exist.
WARNING: Source file examples/images/sphx_glr_clipping_002.vtksz does not exist.
WARNING: Source file examples/images/sphx_glr_clipping_003.vtksz does not exist.
WARNING: Source file examples/images/sphx_glr_clipping_004.vtksz does not exist.
WARNING: Source file examples/images/sphx_glr_clipping_005.vtksz does not exist.
WARNING: Source file examples/images/sphx_glr_extract-edges_001.vtksz does not exist.
WARNING: Source file examples/images/sphx_glr_extract-edges_002.vtksz does not exist.
WARNING: Source file examples/images/sphx_glr_extract-edges_003.vtksz does not exist.
WARNING: Source file examples/images/sphx_glr_perlin_001.vtksz does not exist.
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] index
[...]

PR #4569, opened 2023-06-25
Dynamic Sphinx-Gallery Scraper

banesullivan:
A long-requested feature: saving dynamic, interactive plots in Sphinx-Gallery examples

This implements a mechanism to export a `Plotter` to a new VTK.js standard powered by Trame and viewable with https://kitware.github.io/vtk-js/examples/OfflineLocalView.html

We can then use that viewer combined with exported scenes to embed dynamic renderings in our Sphinx documentation. I've started this by adding a new Sphinx-Gallery scraper specifically for dynamic plots.

This also deprecates and removes the legacy `export_vtkjs`

Requires https://github.com/Kitware/trame-vtk/pull/42

https://github.com/pyvista/pyvista/assets/22067021/c5c1f2b1-838a-4469-83b7-3b3953d06ba6

Resolves #2475
Closes #3632 (subplots are supported in this approach)
Closes #3339
Closes #1234
Closes #3253
Closes #2076
Closes #4181
Closes #4239
Solves #4587

### To Do

- [x] Switch `export_html` to use this to address #3253, #2076, #4181, #4239
- [x] Add some sort of testing
- [x] Remove PyVistaQt's usage of the old VTK.js export: https://github.com/pyvista/pyvistaqt/pull/410
- [x] Fix issue where scraper requires auto-close to be set to false
- [x] Document usage and maybe make an example project in a separate repository

### Future work

- Save jupyter widget state for trame backend with this export mechanism
- Add support for MyST
- Enable the `pyvista-plot` directive to utilize this and switch all `jupyter-execute` blocks in documentation to leverage this
    - This should come at the same time as removing the deprecated jupyter backends

PR #4733, opened 2023-08-03
Use relative paths in the offlineviewer directive

greschd:
### Overview

This is a follow-up to #4632 and its fix in #4636: 

The solution introduced in #4636 uses absolute paths to the `_static` and `_images` directories.

Absolute paths work only if the documentation is served from the root of the domain. This is not the case for example when serving multiple documentation versions from the same domain.
Example deployed documentation: https://composites.dpf.docs.pyansys.com/version/dev/examples/gallery_examples/003_short_fiber_example.html#plot-results

To get the relative path of the '_static' directory to the current HTML document, we assume that the output structure is the same as the source structure.

Use `.as_posix()` on paths used in the HTML to convert them from Windows convention (using backslashes) to using forward-slashes, when built on Windows.

### DIFF
--- a/pyvista/ext/viewer_directive.py
+++ b/pyvista/ext/viewer_directive.py
@@ -4,6 +4,7 @@ import shutil
 
 from docutils import nodes
 from docutils.parsers.rst import Directive
+from docutils.utils import relative_path
 from sphinx.util import logging
 from trame_vtk.tools.vtksz2html import HTML_VIEWER_PATH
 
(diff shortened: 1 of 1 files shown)


tkoyama010:
pre-commit.ci autofix

tkoyama010:
#4569 has not added a test for this extension, causing the coverage error. Comment has been added to skip judgment as a sign to add a test in this area in the future.

tkoyama010 (approved):
I just checked the source because we haven't been able to maintain the test, but LGTM.

tkoyama010:
@greschd Thanks. If you don't mind, could you raise an issue about it. I cannot do it right now.

greschd:
Thanks again for the review @tkoyama010! I've tested `pyvista-docs-dynamic` locally, and opened https://github.com/pyvista/pyvista-docs-dynamic/issues/1 for CI.

PR #4757, opened 2023-08-11
Fix annotation type of `line_width` bool -> float

tkoyama010:
### Overview

Fix annotation type of `line_width` bool -> float.

### Details

- None


adeak (approved):
LGTM, thanks.

PR #4762, opened 2023-08-12
Restore active scalars (fix  #4761).

### LINKED ISSUE #4761: `merge()` modifies `active_scalars_name`

I found a bug!

Merging meshes unexpectedly modifies the currently active scalars.

See the example provided in the [`merge()`](https://docs.pyvista.org/version/0.41/api/core/_autosummary/pyvista.DataSetFilters.merge.html) docstring. The plotted spheres are colored because the merged mesh's `active_scalars_name` is set to `Normals`. However, a regular unmerged sphere does not have any active scalars set and so doing `pv.Sphere().plot()` shows a sphere without any scalar values. The expected behaviour is that a mesh with no active scalars should continue to have no active scalars after merging. So, in the `merge` example, I expected that the merged spheres would appear uniform in color (i.e. no active scalars).

The issue seems to stem from [`DatasetAttributes.update()`](https://github.com/pyvista/pyvista/blob/5b52f342dbb4450b051b55a05286b92952c39233/pyvista/core/datasetattributes.py#L1034). When it copies the data,  [`DatasetAttributes.__setitem__()`](https://github.com/pyvista/pyvista/blob/5b52f342dbb4450b051b55a05286b92952c39233/pyvista/core/datasetattributes.py#L213C15-L213C15) will set the copied array as the active scalars. A fix could be something like modifying `update()` to temporarily store `active_scalars_name`, copy the data, then restore `active_scalars_name`.

```python
import pyvista
sphere_a = pyvista.Sphere(center=(1, 0, 0))
sphere_b = pyvista.Sphere(center=(0, 1, 0))
sphere_c = pyvista.Sphere(center=(0, 0, 1))
merged = sphere_a.merge([sphere_b, sphere_c])
merged.plot()
```

### System Information

```shell
OS : Windows
            CPU(s) : 12
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.8 GiB
       Environment : IPython
        GPU Vendor : Intel
  MathText Support : False
  bit (AMD64)]
           pyvista : 0.42.dev0
               vtk : 9.2.20230323
             numpy : 1.24.2
        matplotlib : 3.7.1
            scooby : 0.7.1
             pooch : v1.7.0
           imageio : 2.26.1
           IPython : 8.12.0
        ipywidgets : 8.0.6
             scipy : 1.10.1
              tqdm : 4.65.0
            meshio : 5.3.4
      nest_asyncio : 1.5.6
```

### Screenshots

Expected:
image

Actual:

user27182:
Resolves #4761

### DIFF
--- a/pyvista/core/datasetattributes.py
+++ b/pyvista/core/datasetattributes.py
@@ -1031,8 +1031,19 @@ class DataSetAttributes(_vtk.VTKObjectWrapper):
             rand                    float64    (1000,)
 
         """
+        # store current active scalars so they can be restored after updating
+        set_active_scalars = False
+        if hasattr(self, 'active_scalars_name'):
+            active_scalars_name = self.active_scalars_name
+            set_active_scalars = True
+
+        # do update
         for name, array in array_dict.items():
             self[name] = array.copy()
+
+        # restore active scalars
+        if set_active_scalars:
+            self.active_scalars_name = active_scalars_name
 
     def _raise_index_out_of_bounds(self, index: Any):
         if isinstance(index, int):


tests/core/test_polydata.py
```diff
@@ -328,50 +328,105 @@ def test_merge(sphere, sphere_shifted, hexbeam):
...
+    mesh = input.copy()
     data_main = np.arange(mesh.n_points, dtype=float)
     mesh.point_data['present_in_both'] = data_main
+    mesh.set_active_scalars('present_in_both')
+
     other = mesh.copy()
     data_other = -data_main
     other.point_data['present_in_both'] = data_other
-    merged = mesh.merge(other, main_has_priority=True)
+    other.set_active_scalars('present_in_both')
```
tkoyama010, line 429: Given the alignment with the name of the test function, in my mind this line should not be changed.

tkoyama010 (changes requested):
Thank you for creating the pull request.

tests/core/test_polydata.py
user27182: Thanks for the feedback. The diff in this case is a bit misleading. The line `merged = mesh.merge(other, main_has_priority=True)` was not changed, it's actually still there (moved from line `374` to `440`. The move was done as part of a general refactoring for that test. This line specifically was originally placed before the `matching_point_data` function was defined, but the variable `merged` is critical to understanding the test being applied. So it was moved to be after the function definition and just before the assertion.

pyvista/core/filters/poly_data.py
```diff
@@ -550,9 +550,19 @@ def merge(
                 polydata_merged = pyvista.PolyData(
                     merged.points, faces=merged.cells, n_faces=merged.n_cells, deep=False
                 )
+                # Calling update() will modify the active scalars in this specific
+                # case. Store values to restore after updating.
```
tkoyama010, line 554: I checked if it is possible to handle this in the update method, but it could not be handled when main_has_priority=True. (This is a memo.)

tkoyama010 (approved):
LGTM. Thanks for this. I have one question. Does this also happen for the following data?
- `active_vectors_name`
- `active_t_coords_name`
- `active_normals_name`
Please send a follow-up PullRequest if it occurs.

PR #4765, opened 2023-08-13
Raise `ValueError` when setting bad data from VTKObject.

### LINKED ISSUE #4764: Setting scalar values with the wrong shape does not raise `ValueError`

I found a bug!

Normally, Pyvista raises a `ValueError` if you try to set scalar data that does not match the shape of its points/cells. e.g. `pv.Sphere()['foo'] = [1,2,3]` is not allowed.

However, this error is not raised if setting scalars from an existing incorrectly-shaped data array. This means that any buggy vtk methods that output incorrectly-sized arrays (e.g. `vtkConnectivityFilter`) may be left unchecked. As such, in specific cases it's possible to set scalar data with an incorrect size. The expected behaviour is that a `ValueError` should always be raised, even if the array was previously set by a vtk method.

```python
from vtk import vtkConnectivityFilter
import pyvista as pv
mesh = pv.Sphere() + pv.Cube()

f = vtkConnectivityFilter()
f.SetInputData(mesh)
f.SetExtractionModeToLargestRegion()
f.ColorRegionsOn()
f.Update()
out = pv.wrap(f.GetOutput())

# output scalar data has the same size as the merged input (sphere + cube)
# however, this is a bug with vtk, as the scalar data should have the same size as the largest region (sphere)

out.cell_data['Foo'] = out.cell_data['RegionId']  # no error, but a ValueError was expected
out.cell_data['Bar'] = out.cell_data['RegionId'].copy() # ValueError correctly raised
```

### System Information

```shell
OS : Windows
            CPU(s) : 12
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.8 GiB
       Environment : IPython
        GPU Vendor : Intel
  MathText Support : False
  bit (AMD64)]
           pyvista : 0.42.dev0
               vtk : 9.2.20230323
             numpy : 1.24.2
        matplotlib : 3.7.1
            scooby : 0.7.1
             pooch : v1.7.0
           imageio : 2.26.1
           IPython : 8.12.0
        ipywidgets : 8.0.6
             scipy : 1.10.1
              tqdm : 4.65.0
            meshio : 5.3.4
      nest_asyncio : 1.5.6
```

### Screenshots

_No response_

user27182:
Resolves #4764 

The attempt to reuse the existing pointer to the underlying VTK data when setting `DatasetAttributes` (and return early) is moved so that the check is made _after_ an initial validation is done on the shape of the data. This is done to ensure that any existing VTK data is always checked for validity before being set.


tkoyama010:
pre-commit.ci autofix

pyvista/core/datasetattributes.py
```diff
@@ -711,6 +711,24 @@ def _prepare_array(
         if data is None:
             raise TypeError('``data`` cannot be None.')
 
+        # convert to numpy type if necessary
```
tkoyama010, line 714: Thanks for the change. Could you please comment on your explanation in the PullRequest summary so that later reviewers of this PullRequest will understand the reason why you moved this code here?

pyvista/core/datasetattributes.py
```diff
@@ -711,6 +711,24 @@ def _prepare_array(
...
+        # Fixup input array length for scalar input
+        if not isinstance(data, np.ndarray) or np.ndim(data) == 0:
+            tmparray = np.empty(array_len)
+            tmparray.fill(data)
+            data = tmparray
+        if data.shape[0] != array_len:
+            raise ValueError(f'data length of ({data.shape[0]}) != required length ({array_len})')
+
         # attempt to reuse the existing pointer to underlying VTK data
         if isinstance(data, pyvista_ndarray):
```
tkoyama010, line 733: The conversion of `data` to `ndarray` makes no sense to have this if statement. In my mind, I want to remove it.

tkoyama010 (changes requested):

pyvista/core/datasetattributes.py
user27182: I edited the initial top-level comment and added the reason there. Is that the right place to add this info?

tkoyama010: Yes. Thanks a lot.

pyvista/core/datasetattributes.py
akaszynski: Checks like `data.shape[0]` would fail if data isn't a numpy array. I'd leave it in.

akaszynski (dismissed):
LGTM. Thanks!

pyvista/core/datasetattributes.py
user27182: I'll note that none of this code was modified, the code block starting on line `732` was simply cut/pasted from its original position at the start of function to down a few lines to be executed after the initial check of the `data` shape is made.

RE: your comment, on line `733`, `isinstance(data, pyvista_ndarray)` makes sense to me here as it's the first time in the function where this check is made and it prevents an `AttributeError` from being raised for non-`pyvista_ndarray` types when accessing its `VTKObject` attribute on the next line.

However, after further review, on line `722`, i.e. `array_len = data.shape[0] if isinstance(data, np.ndarray) else 1`, the check `if isinstance(data, np.ndarray)` looks to me like it will always evaluate to `True` since the data was already cast as `ndarray`-like on line `715` with `data = np.asanyarray(data)`. I believe the intent may be to catch scalar values where `.shape` is not defined, e.g. `np.array(1).shape[0]` will raise an `IndexError`. So, perhaps line `722` should be changed from:

- `array_len = data.shape[0] if isinstance(data, np.ndarray) else 1` to
- `array_len = 1 if np.ndim(data) == 0 else data.shape[0]`

Again I'll stress that I only moved the code around and didn't change any of the lines, so what you see in this PR was already there. But, it does look like there may be a bug.

Indeed, the following code will raise an `IndexError`:
``` python
m = pv.PolyData([0, 0, 0.])
 m.point_data["foo"] = 1   # No error
m.cell_data["bar"] = 1   # No error
m["baz"] = 1  # IndexError raised
```

Interestingly, the error was raised elsewhere in `dataset`, prior to reaching the line in question (i.e. line `722` in `datasetattributes`). Here's the error traceback:
```
  File "\lib\site-packages\IPython\core\interactiveshell.py", line 3505, in run_code
  File "<ipython-input-1-24ef6ab2e5f3>", line 1, in <module>
    m["baz"] = 1
  File "\pyvista\pyvista\core\dataset.py", line 1773, in __setitem__
    if scalars.shape[0] == self.n_points:
IndexError: tuple index out of range
```

tkoyama010: Thanks, both of you for your comments. I have to go to bed in my time zone, so let me think about it again tomorrow.

user27182:
Related to setting scalars with a single value... Is there a desire to support broadcasting? E.g.
``` python
# status quo
m = pv.PolyData([[0, 0, 0.], [1,1,1.]])
m["foo"] = 1  # ValueError: Number of scalars (1) must match either the number of points (2) or the number of cells (2).

# with broadcasting
m = pv.PolyData([[0, 0, 0.], [1,1,1.]])
m["foo"] = 1  # No error
m["foo"]  # pyvista_ndarray([1, 1])
```
This would be a handy shortcut when using labeled data, e.g. labelled components or segmentation labels. Allows for easily setting the entire mesh to be labelled with a single scalar.

Though maybe this should be a separate PR...

MatthewFlamm:
> Is there a desire to support broadcasting?

I think this sounds like a great idea, but needs to be discussed.

> Though maybe this should be a separate PR...

Definitely a separate PR.  There might be some implementation details here that could be tricky.  Does it broadcast to point data or cell data for example?  Maybe it should only broadcast when it is unambiguous like:

```py
m.point_data["foo"] = [1.0, 2.0, 3.0]  # This case can still be ambiguous if there are 3 points.
```

We can discuss in an issue or another PR.  I just wanted to highlight some potential thorns that need to be worked through.

pyvista/core/datasetattributes.py
user27182: The latest commit from yesterday includes a fix for the `IndexError` raised when setting a single scalar value for a mesh with only one point. This was kind of an aside and was not really the intent of this PR, and is a very specific and niche use case. So, to move this PR forward, should I roll back those changes, and go with the changes that were approved by @akaszynski ?

akaszynski:
> so we should also decide whether it makes sense to have any broadcasting.

I think that we should avoid broadcasting and instead rely on `numpy`'s broadcasting rules. There are too many ambiguous cases (e.g. same number of points and cells).

 > The latest commit from yesterday includes a fix for the IndexError raised when setting a single scalar value for a mesh with only one point.
 
 The behavior change you introduced agrees with our underlying API. If this is acceptable:

```py
import pyvista as pv
import numpy as np

m = pv.PolyData([[0, 0, 0.]])
m.point_data["foo"] = 1
```

Then this should also work:

```py
m["foo"] = 1
```

akaszynski (dismissed):
Agree with changes. Thanks for your work!

tkoyama010 (approved):

tests/core/test_dataset.py
```diff
@@ -131,6 +131,16 @@ def test_cell_data_bad_value(grid):
...
 
 
+def test_point_cell_data_single_scalar_no_exception_raised():
+    try:
+        m = pyvista.PolyData([0, 0, 0.0])
+        m.point_data["foo"] = 1
+        m.cell_data["bar"] = 1
+        m["baz"] = 1
+    except Exception as e:
+        pytest.fail(f"Unexpected exception raised: {e}")
```
tkoyama010, line 141:
```suggestion
    try:
        m = pv.PolyData([0, 0, 0.0])
        m.point_data["foo"] = 1
        m.cell_data["bar"] = 1
        m["baz"] = 1
    except Exception as e:
        pytest.fail(f"Unexpected exception raised: {e}")
```

tkoyama010 (approved):

PR #4772, opened 2023-08-15
Add GemGIS as an external example

tkoyama010:
### Overview

GemGIS is a spatial data processing for geomodeling.

gemgis

### Details

- None


AlexanderJuestel:
@tkoyama010 thanks for adding GemGIS as an external example to the Pyvista docs! Many GemGIS methods rely on Pyvista and I also use it personally for my PhD thesis. So thank you and congratulations for this awesome package (@banesullivan) and user support (e.g. @MatthewFlamm very recently).

PR #4779, opened 2023-08-16
Add `Plotter.meshes`

### LINKED ISSUE #4769: Retrieving meshes from the plotter object

I'm not sure if this is a bug or it is the intended behaviour and I'm using pyvista wrong.

But if I add several meshes to a plotter, when I use `pl.mesh` I only retrieve the latest mesh. IMO it should retrieve either a mesh which combine all all the previous meshes, or a list of those meshes.

Maybe am I using it wrong?

```python

import pyvista

sphere = pyvista.Sphere(center=(1, 0, 0))
box = pyvista.Box()

pl = pyvista.Plotter()
pl.add_mesh(sphere, scalars=range(sphere.n_points))
pl.add_mesh(box, scalars=range(box.n_cells))

assert box == pl.mesh
```

### System Information

```shell
  Date: Mon Aug 14 17:58:38 2023 CEST

                OS : Darwin
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
               RAM : 16.0 GiB
       Environment : IPython
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M2
  MathText Support : False

  Python 3.11.3 (main, Apr  7 2023, 20:13:31) [Clang 14.0.0
  (clang-1400.0.29.202)]

           pyvista : 0.41.1
               vtk : 9.2.6
             numpy : 1.25.1
        matplotlib : 3.7.2
            scooby : 0.7.2
             pooch : v1.7.0
           imageio : 2.31.1
           IPython : 8.14.0
        ipywidgets : 8.0.7
             scipy : 1.11.1
              tqdm : 4.65.0
            meshio : 5.3.4
        jupyterlab : 4.0.2
         pythreejs : 2.4.2
             trame : 3.1.0
      trame_client : 2.10.0
      trame_server : 2.11.7
         trame_vtk : 2.5.8
      nest_asyncio : 1.5.6
```

### Screenshots

_No response_

germa89:
### Overview

This PR adds the property `meshes` to the `Plotter` object, so you can use it to retrieve the meshes added to the plotter.

I believe this method is more intuitive and natural than using actors.

Resolves #4769 

### Caveats

- I wonder if eventually in the future `Plotter.mesh` should be deprecated for this method, or maybe renamed to `Plotter.last_mesh` which makes more sense IMO.
- I also believe that `Plotter.mesh` should return a mesh results of merging all the added meshes. But this change will be a breaking change, so I'm not proposing in here.

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -6885,6 +6885,23 @@ class Plotter(BasePlotter):
 
         return actor
 
+    @property
+    def meshes(self):
+        """Return plotter meshes.
+
+        _extended_summary_
+
+        Returns
+        -------
+        List[Union[pv.PolyData, pv.UnstructuredGrid, pv.UniformGrid]
+            List of mesh objects such as pv.PolyData, pv.UnstructuredGrid, etc.
+        """
+        meshes = []
+        for actor in self.actors.values():
+            if hasattr(actor, 'mapper'):
+                meshes.append(actor.mapper.dataset)
+
+        return meshes
 
 # Tracks created plotters.  This is the end of the module as we need to
 # define ``BasePlotter`` before including it in the type definition.
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -6889,11 +6889,9 @@ class Plotter(BasePlotter):
     def meshes(self):
         """Return plotter meshes.
 
-        _extended_summary_
-
         Returns
         -------
-        List[Union[pv.PolyData, pv.UnstructuredGrid, pv.UniformGrid]
+        List[Union[pyvista.DataSet, PyVista.MultiBlock]]
             List of mesh objects such as pv.PolyData, pv.UnstructuredGrid, etc.
         """
         meshes = []
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -6903,6 +6903,7 @@ class Plotter(BasePlotter):
 
         return meshes
 
+
 # Tracks created plotters.  This is the end of the module as we need to
 # define ``BasePlotter`` before including it in the type definition.
 #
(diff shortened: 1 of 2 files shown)


tkoyama010:
pre-commi.ci autofix

pyvista/plotting/plotter.py
```diff
@@ -6885,6 +6885,24 @@ def add_cursor(
 
         return actor
 
+    @property
+    def meshes(self):
+        """Return plotter meshes.
+
+        _extended_summary_
```
tkoyama010, line 6892: Could you please explain as I am not understanding the intent of the `_extended_summary_` symbol?

tkoyama010 (changes requested):

pyvista/plotting/plotter.py
```diff
@@ -6885,6 +6885,24 @@ def add_cursor(
...
+
+        Returns
+        -------
+        List[Union[pv.PolyData, pv.UnstructuredGrid, pv.UniformGrid]
```
MatthewFlamm, line 6896: Instead of an incomplete list, I would use `pyvista.DataSet` instead.  What is returned with a MultiBlock mesh?  I assume that object?  If so, then it would be `List[Union[pyvista.DataSet, PyVista.MultiBlock]]`.

pyvista/plotting/plotter.py
germa89: That shoudn't be there.

pyvista/plotting/plotter.py
germa89: This is something I didn't know at the time and I need help with. I wrote that as my best guess.

I will change it to `pyvista.Dataset`. Thank you!

pyvista/plotting/plotter.py
germa89, line 6892:
```suggestion
```

pyvista/plotting/plotter.py
```diff
@@ -6885,6 +6885,24 @@ def add_cursor(
...
         return actor
 
+    @property
+    def meshes(self):
+        """Return plotter meshes.
+
+        _extended_summary_
+
+        Returns
+        -------
```
germa89, line 6895:
```suggestion
        """Return plotter meshes.

        Returns
        -------
```

pyvista/plotting/plotter.py
germa89, line 6896:
```suggestion
        List[Union[pyvista.DataSet, PyVista.MultiBlock]]
```

tkoyama010 (approved):
LGTM

PR #4782, opened 2023-08-16
Add sample enhancements

### LINKED ISSUE #4658: Enhancements for sample filter

- Enable for composite data sets
- Expose pass partial arrays setting in that case
- Look into whether ComputeTolerance needs to be set when setting tolerance
- SnapToCellWithClosestPoint functionality to enable resampling

Lower Priority
- SetCategoricalData.
- PassFieldArrays. Could be useful to avoid name clashes.
- MarkBlankPointsandCells.  I'm not sure why anyone would ever want to turn this off, but it is an option in vtk

This comes out of the work in #4648

https://vtk.org/doc/nightly/html/classvtkResampleWithDataSet.html#ac433a25e3e096c999bd99b06302f8171

### Pseudocode or Screenshots

_No response_

MatthewFlamm:
### Overview
Closes #4658.

- Enables use with composite datasets.
- Use optionally string for cell locator
- Expose MarkBlankPointsAndCells
- Expose PassFieldArrays

### Details

- SnapToClosestCell is only available on vtk 9.3.0
- I can't figure out what PassPartialArrays does, and it creates seg faults when I implemented except in specific combinations of composite and non composite datasets usage.


MatthewFlamm:
In my testing the default locator, i.e. the dataset `FindCell`, is not robust and often leads to not finding a cell for a given point that is clearly within the dataset.  When using `vtkStaticCellLocator` it gives more robust finding.  But I'm not sure if this the best option for all cases.  I think it would be better to have a default that is more robust, but I'm not as familiar with the intricacies of each locator and how they might interact with all the dataset types.

pyvista/core/filters/data_set.py
```diff
@@ -3156,6 +3158,10 @@ def sample(
         weighting for nearby points.  If there is cell topology, `sample` is
         usually preferred.
 
+        The point data 'vtkValidPointMask' stores whether the point could be sampled
+        with a value of 1 meaning successful sampling and value of 0 meaning
+        unsuccessful.
```
tkoyama010, line 3163: Sorry for nitpick. But I want to split the sentence.
```suggestion
        The point data 'vtkValidPointMask' stores whether the point could be sampled
        with a value of 1 meaning successful sampling. And a value of 0 means
        unsuccessful.
```

tkoyama010 (changes requested):

tkoyama010:
Thanks for this. This is a great addition.

tkoyama010 (approved):
LGTM. Thanks!

PR #4798, opened 2023-08-22
Add user interaction style

akaszynski:
There are times where it's necessary to disable certain events from the interactor without disabling the interactor itself. For example, if you desire to click and "drag" an actor without at the same time modifying the camera position, it's not possible to do this right now with our current wrapped interactor.

This PR adds `enable_user_style` to allow this behavior.


pyvista/plotting/render_window_interactor.py
```diff
@@ -780,6 +781,29 @@ def enable_rubber_band_2d_style(self):
         self._style_class = None
         self.update_style()
 
+    def enable_user_style(self):
+        """Set the interactive style to user.
+
+        This interaction style allows you to customize interaction events. By
+        default, this interaction style does nothing.
```
adeak, line 788: This sounds useful. But it's not at all clear to me how events can be customized. Shouldn't we mention something about this, maybe just link to other parts of the documentation if it's already something that exists? If this is completely novel, I think we should consider adding another docstring example that shows how to configure a single event.

akaszynski: Good point. I mostly added this here just to disable user events to enable a fancy geometry manipulation widget I've been working on for a separate application:
tmp

I'll add the ability to customize at least one event.

tests/plotting/test_render_window_interactor.py
```diff
@@ -220,3 +221,10 @@ def test_poked_subplot_context(verify_image_cache):
         pl.add_mesh(pv.Arrow(), color=True)
 
     pl.show()
+
+
+def test_user_interaction_style():
+    pl = pyvista.Plotter()
+    pl.enable_user_style()
+    assert pl.iren._style == 'User'
+    assert isinstance(pl.iren.interactor.GetInteractorStyle(), vtk.vtkInteractorStyleUser)
```
tkoyama010, line 230:
```suggestion
    pl = pv.Plotter()
    pl.enable_user_style()
    assert pl.iren._style == 'User'
    assert isinstance(pl.iren.interactor.GetInteractorStyle(), vtk.vtkInteractorStyleUser)
```

tkoyama010 (approved):
LGTM

pyvista/plotting/render_window_interactor.py
adeak: @tkoyama010 when @akaszynski said
> I'll add the ability to customize at least one event.

I understood that as a plan to update this PR with that feature. I'd prefer to wait until @akaszynski resolves this thread himself to avoid any misunderstandings.

tkoyama010 (changes requested):
@adeak Thanks. I will wait @akaszynski 's reply.

PR #4801, opened 2023-08-23
Add `download_frog_tissue` example and `'foreground'` opacity mapping

user27182:
### Overview

This PR makes the frog tissue dataset available for download and adds a `'foreground'` opacity mapping to `add_volume` for plotting labelled image data where the background is zero.

There is currently no medical image segmentation labels data available for download with PyVista (that I'm aware of), so this addition is significant in that regard.

### Details
-Add `download_frog_tissue` with a well-tested example in the docstring for viewing the data
-Add `foreground` opacity mapping to `add_volume` and related functions
-Improve `add_volume` documention for `clim` to expose undocumented feature where a single float is used
-Fix incorrect `add_volume`  documentation for the default value of `blending`

Image from the example in `download_frog_tissue`:
image


user27182:
Some notes for future updates:

- There is an [updated vtk example](https://www.kitware.com/froggie-and-the-joy-of-open-source-community/) using the frog and frog tissue dataset. Perhaps a more full-fledged PyVista example can be created based on this.
- A lot of work went into determining the appropriate `add_volume` parameters for visualizing image label data. Perhaps `add_volume` could be updated to allow using preset default values for specific cases like image label data.

pyvista/examples/downloads.py
```diff
@@ -1653,11 +1653,80 @@ def download_frog(load=True):  # pragma: no cover
...
+        to ``False`` and only the filename will be returned.
+
+    Returns
+    -------
+    pyvista.ImageData | str
+        DataSet or filename depending on ``load``.
+
+    Examples
+    --------
+    # Load data
```
tkoyama010, line 1679: We don't need to comment out the comments :)
```suggestion
    Load data

```

pyvista/examples/downloads.py
```diff
@@ -1653,11 +1653,80 @@ def download_frog(load=True):  # pragma: no cover
...
+    >>> from pyvista import examples
+    >>> data = examples.download_frog_tissue()
+
+    # Plot tissue labels as a volume
```
tkoyama010, line 1683:
```suggestion
    Plot tissue labels as a volume

```

pyvista/examples/downloads.py
```diff
@@ -1653,11 +1653,80 @@ def download_frog(load=True):  # pragma: no cover
...
+    >>> # First, define plotting parameters
+    >>> # Configure colors / color bar
```
tkoyama010, line 1685:
```suggestion
    First, define plotting parameters

    >>> # Configure colors / color bar
```

pyvista/examples/downloads.py
```diff
@@ -1653,11 +1653,80 @@ def download_frog(load=True):  # pragma: no cover
...
+    >>> clim = data.get_data_range()  # Set color bar limits to match data
+    >>> cmap = 'glasbey'  # Use a categorical colormap
+    >>> categories = True  # Ensure n_colors matches number of labels
+    >>> opacity = (
+    ...     'foreground'  # Make foreground opaque, background transparent
+    ... )
+    >>> opacity_unit_distance = 1
+
+    >>> # Set plotting resolution to half the image's spacing
```
tkoyama010, line 1694:
```suggestion
    Set plotting resolution to half the image's spacing

```

pyvista/examples/downloads.py
```diff
@@ -1653,11 +1653,80 @@ def download_frog(load=True):  # pragma: no cover
...
+    >>> res = np.array(data.spacing) / 2
+
+    >>> # Define rendering parameters
```
tkoyama010, line 1697:
```suggestion
    Define rendering parameters

```

pyvista/examples/downloads.py
```diff
@@ -1653,11 +1653,80 @@ def download_frog(load=True):  # pragma: no cover
...
+    >>> mapper = 'gpu'
+    >>> shade = True
+    >>> ambient = 0.3
+    >>> diffuse = 0.6
+    >>> specular = 0.5
+    >>> specular_power = 40
+
+    # Make and show plot
```
tkoyama010, line 1705:
```suggestion
    Make and show plot

```

tkoyama010 (changes requested):
@user27182 This is great thanks! I have only a small suggestion in docstring :) Otherwise, LGTM.

user27182:
> Unfortunately this would be a breaking change, but if it makes the method more usable overall, I think it makes sense. But we also need to weigh different uses of `add_volume` for defaults, for example, another use would be to visualize the volumetric density of some variables in a CFD simulation or similar experimental measure.
> 
> I could see a separate method that only provides different defaults for different use cases or we could think about how to have mix and match themes for this sort of usage. I would lean towards towards the latter.

My initial thought was to use something like a theme as you mentioned, but I have no specific preference. However, I concede that making changes to accommodate image label data is likely not a worthwhile effort at this point.

As it stands, `add_volume` is definitely not geared towards visualizing image segmentation labels, and that was made apparent while trying to determine the correct parameters for visualizing the frog tissue, since most of the default parameters had to be changed. 

For example, the default behaviour for `clim` is to set it to `[min(data), max(data)]` in all cases _except_ for when the data type is `uint8`, where `clim` is set to `[0, 255]`. This was probably done to support a specific use case where RGBA scalars are used or something like that. However, this was unexpected behaviour for the image label data, which is almost always type `uint8` for medical images, where the default of `clim=[min(data), max(data)]` would have been preferred; as a result this required manually setting `clim` for the frog tissue example which I don't think should have been necessary (and was not obvious... if the data type was `uint16` I would have expected the same behaviour... so now as a user I am forced to be explicitly aware of the exact data type of my label data).

There were also performance issue with `add_volume` for this type of data, especially when compared to the performance of the VolumeRendering module available in 3DSlicer. Also, trying to visualize separate labels in a `MultiBlock` was pretty much impossible/unusable due to extremely slow speed since a separate mapper is created for each block.

In general I don't think PyVista is particularly well suited to medical image data due to the lack of explicit and consistent use of image orientation with `vtkImageData` (a vtk issue), and there is currently a limited number of `ImageData` specific PyVista filters available.

So, for these reasons, I don't think it's worth making any changes to `add_volume` at this point to improve its use with image label data. But at least there is proof of concept of its possible utility with the frog tissue example provided here.

pyvista/examples/downloads.py
```diff
@@ -1653,11 +1653,86 @@ def download_frog(load=True):  # pragma: no cover
...
+    Returns
+    -------
+    pyvista.ImageData | str
+        DataSet or filename depending on ``load``.
+
+    Examples
+    --------
+    Load data
+
+    >>> from pyvista import examples
```
MatthewFlamm, line 1681:
```suggestion
    >>> import numpy as np
    >>> import pyvista as pv
    >>> from pyvista import examples
```

pyvista/examples/downloads.py
```diff
@@ -1653,11 +1653,86 @@ def download_frog(load=True):  # pragma: no cover
...
+    >>> mapper = 'gpu'
+    >>> shade = True
+    >>> ambient = 0.3
+    >>> diffuse = 0.6
+    >>> specular = 0.5
+    >>> specular_power = 40
+
+    Make and show plot
+
+    >>> p = pyvista.Plotter()
```
MatthewFlamm, line 1712:
```suggestion
    >>> p = pv.Plotter()
```

MatthewFlamm (changes requested):

MatthewFlamm:
I do not like this functionality of having different behavior for clim depending on the type.  This is too confusing.  There instead should be a parameter that turns this on and off.  There is already a bug report/ clarification request for this 'feature' in #4700, and I will try to move this discussion there.

pyvista/examples/downloads.py
```diff
@@ -1653,11 +1653,88 @@ def download_frog(load=True):  # pragma: no cover
...
-    # TODO: there are other files with this
     download_file('froggy/frog.zraw')
     return _download_and_read('froggy/frog.mhd', load=load)
 
 
+def download_frog_tissue(load=True):  # pragma: no cover
+    """Download frog tissue dataset.
+
+    This dataset contains tissue segmentation labels for the frog dataset
+    (see :ref:`download_frog`).
```
MatthewFlamm, line 1664:
```suggestion
    (see :func:`download_frog`).
```

MatthewFlamm (commented):
Also consider putting in a see also in `download_frog`.  Otherwise LGTM.

pyvista/examples/downloads.py
user27182: I'm not very familiar with how the references work for the docs... Is ``:func:`pyvista.examples.downloads.download_frog` `` preferred?

tkoyama010 (approved):

MatthewFlamm (approved):

tests/plotting/test_plotting.py
```diff
@@ -1864,6 +1864,10 @@ def test_opacity_transfer_functions():
     assert len(mapping) == n
     mapping = pv.opacity_transfer_function('sigmoid_10', n)
     assert len(mapping) == n
+    mapping = pyvista.opacity_transfer_function('foreground', n)
+    assert len(mapping) == n
+    mapping = pyvista.opacity_transfer_function('foreground', 5)
+    assert np.array_equal(mapping, [0, 255, 255, 255, 255])
```
tkoyama010, line 1870:
```suggestion
    mapping = pv.opacity_transfer_function('foreground', n)
    assert len(mapping) == n
    mapping = pv.opacity_transfer_function('foreground', 5)
    assert np.array_equal(mapping, [0, 255, 255, 255, 255])
```

tkoyama010 (approved):

PR #4802, opened 2023-08-23
Add `pack_labels` and `sort_labels` filters

user27182:
### Overview

This PR adds two new filters `pack_labels` and `sort_labels` for working with labelled data.
Note that the docstring examples depend on the `frog_tissue` dataset from #4801.

### Details

- `pack_labels` makes use of [`vtkPackLabels` ](https://vtk.org/doc/nightly/html/classvtkPackLabels.html)
- `sort_labels`  is a convenience method which calls `pack_labels` with `sort=True`


user27182:
Note that [vtkPackLabels](https://vtk.org/doc/nightly/html/classvtkPackLabels.html) requires vtk version 9.3, looks like tests are failing from import errors from using older versions. Is there a protocol for managing this with PyVista to add the latest vtk features or are only older versions of vtk supported?

tkoyama010:
@user27182
When there was a mix of less than vtk9 and more than vtk9, we implemented the import as follows This may be helpful to you.
https://github.com/pyvista/pyvista/blob/40c7a57312faf8a482f640f38526019cceec4552/pyvista/_vtk.py#L16-L34

user27182:
I have added conditional skips if vtk version isn't 9.3 or higher, and added (unconditional) doctest skips for the docstrings. However this has reduced the code coverage since I believe any tests with `pack_labels` or `sort_labels` are now skipped altogether. So, the tests may "pass", but that's because they're simply not being run. I'm not sure if any of the CI tests specifically do testing with VTK 9.3? If not then this PR should wait until this is enable/implemented to ensure proper coverage for the `pack_labels` and `sort_labels` filters.

pyvista/core/_vtk_core.py
```diff
@@ -11,6 +11,10 @@
 
 from vtkmodules.vtkCommonCore import vtkVersion
 
+VTK93 = vtkVersion().GetVTKMajorVersion() >= 9 and vtkVersion().GetVTKMinorVersion() >= 3
+if VTK93:
```
tkoyama010, line 15:
```suggestion
if VTK93:  # pragma: no cover 
```

pyvista/core/filters/data_set.py
```diff
@@ -5850,6 +5850,237 @@ def extract_cells_by_type(self, cell_types, progress_bar=False):
...
+        # Determine output scalars
+        if output_scalars is None:
+            output_scalars = scalars
+        elif isinstance(output_scalars, str):
+            output_scalars = output_scalars
+        else:
+            raise TypeError(f"Output scalars must be a string, got {type(output_scalars)} instead.")
+
+        # Do packing
+        if _vtk.VTK93:
```
tkoyama010, line 6038:
```suggestion
        if _vtk.VTK93:  # pragma: no cover 
```

pyvista/core/filters/data_set.py
```diff
@@ -5850,6 +5850,237 @@ def extract_cells_by_type(self, cell_types, progress_bar=False):
...
+        Examples
+        --------
+        Pack segmented image labels.
+
+        >>> # Load non-contiguous image labels
+        >>> from pyvista import examples
+        >>> import numpy as np
+        >>> image_labels = examples.download_frog_tissue()
+        >>>
+        >>> # Show range of labels
```
tkoyama010, line 6000:
```suggestion

        Show range of labels
```

pyvista/core/filters/data_set.py
```diff
@@ -5850,6 +5850,237 @@ def extract_cells_by_type(self, cell_types, progress_bar=False):
...
+        >>> image_labels.get_data_range()
+        (0, 29)
+        >>>
+        >>> # Find 'gaps' in the labels
```
tkoyama010, line 6004:
```suggestion

        Find 'gaps' in the labels
```

pyvista/core/filters/data_set.py
```diff
@@ -5850,6 +5850,237 @@ def extract_cells_by_type(self, cell_types, progress_bar=False):
...
+        >>> label_numbers = np.unique(image_labels.active_scalars)
+        >>> label_max = np.max(label_numbers)
+        >>> missing_labels = set(range(label_max)) - set(label_numbers)
+        >>> len(missing_labels)
+        4
+        >>>
+        >>> # Pack labels to remove gaps
```
tkoyama010, line 6011:
```suggestion

        Pack labels to remove gaps
```

pyvista/core/filters/data_set.py
```diff
@@ -5850,6 +5850,237 @@ def extract_cells_by_type(self, cell_types, progress_bar=False):
...
+        >>> packed_labels = image_labels.pack_labels()
+        >>>
+        >>> # Show range of packed labels
```
tkoyama010, line 6014:
```suggestion

        Show range of packed labels
```

pyvista/core/filters/data_set.py
```diff
@@ -5850,6 +5850,237 @@ def extract_cells_by_type(self, cell_types, progress_bar=False):
...
+        (0, 25)
+
+        """
+        # Determine input scalars
+        if scalars is None:
+            set_default_active_scalars(self)
+            _, scalars = self.active_scalars_info
+        arr = get_array(self, scalars, preference=preference, err=False)
+        if arr is None:
+            raise ValueError('No arrays present to sort.')
```
tkoyama010, line 6025: In my mind, I vote to leave error handling to get_array to reduce maintenance costs, and I don't think users will have much trouble since Python shows which function caused the error when it occurs.
```suggestion
        arr = get_array(self, scalars, preference=preference, err=True)
```

tkoyama010 (changes requested):
This is a great feature to add. vtk9.3 is already in pre-release phase in [vkt's pypi page](https://pypi.org/project/vtk/#history) so we can wait until they release it. Thanks for your diligence in checking out the new features of your vtk.
I will add CI for vtk9.3 in another PR when vtk9.3 is released. Can you please merge that in?
As for coverage, we can skip coverage until VTK < 9.3 is gone adding "pragma: no cover" comment in my mind.

tkoyama010:
pre-commit.ci autofix

pyvista/core/filters/data_set.py
user27182: Agreed. This code block was borrowed from `threshold`. Should `threshold` have this check removed too? Also, `test_sort_labels` has a test case written specifically to check that `ValueError` is raised with no data. Should this be removed as well, since that error should be tested / handled by `get_array`?

user27182:
> I will add CI for vtk9.3 in another PR when vtk9.3 is released. Can you please merge that in?

Thanks for clarifying about the 9.3 CI testing. I am not sure what you me to merge, however. Do you want me to merge changes at a later date once CI testing is added?

tkoyama010:
Yes. Please merge main branch once CI testing is added.

user27182:
Ok will do. So you want to put this PR on hold until 9.3 CI testing is added, then merge main into this branch, then merge this PR after all checks pass at that point?

pyvista/core/filters/data_set.py
tkoyama010: Yes. Please. I guess we should change it too.

akaszynski:
Let's add this feature in once VTK 9.3 is out in the wild or we have testing for 9.3 in our CI. Since Python 3.12 comes out in August and VTK's release should be out sometime in 8 Sept [source](https://discourse.vtk.org/t/vtk-9-3-0-release-cycle/11598/5), we may consider adding CI testing for Python 3.12 + VTK 9.3

pyvista/core/filters/data_set.py
```diff
@@ -5850,6 +5850,235 @@ def extract_cells_by_type(self, cell_types, progress_bar=False):
...
+        >>>
+        >>> # Show sorted label info for the four largest labels. Note
+        >>> # the difference in label size after sorting.
+        >>> sorted_label_number, sorted_label_size = np.unique(
+        ...     sorted_labels["MetaImage"], return_counts=True
+        ... )
+        >>> sorted_label_number[:4]
+        pyvista_ndarray([0, 1, 2, 3], dtype=uint8)
+        >>> sorted_label_size[:4]
+        array([30805713,   438052,   204672,   133880], dtype=int64)
```
tkoyama010, line 5925:
```suggestion
        Load image labels
        >>> from pyvista import examples
        >>> image_labels = examples.download_frog_tissue()

        Show label info for first four labels
        >>> label_number, label_size = np.unique(
        ...     image_labels['MetaImage'], return_counts=True
        ... )
        >>> label_number[:4]
        pyvista_ndarray([0, 1, 2, 3], dtype=uint8)
        >>> label_size[:4]
        array([30805713,    35279,    19172,    38129], dtype=int64)

        Sort labels
        >>> sorted_labels = image_labels.sort_labels()

        Show sorted label info for the four largest labels. Note
        the difference in label size after sorting.
        >>> sorted_label_number, sorted_label_size = np.unique(
        ...     sorted_labels["MetaImage"], return_counts=True
        ... )
        >>> sorted_label_number[:4]
        pyvista_ndarray([0, 1, 2, 3], dtype=uint8)
        >>> sorted_label_size[:4]
        array([30805713,   438052,   204672,   133880], dtype=int64)
```

MatthewFlamm:
vtk 9.3 is now in the CI, please update this branch from main.

pyvista/core/filters/data_set.py
```diff
@@ -6239,6 +6239,236 @@ def extract_cells_by_type(self, cell_types, progress_bar=False):
...
+            raise TypeError(f"Output scalars must be a string, got {type(output_scalars)} instead.")
+
+        # Do packing
+        if _vtk.VTK93:  # pragma: no cover
+            alg = _vtk.vtkPackLabels()
+            alg.SetInputDataObject(self)
+            alg.SetInputArrayToProcess(0, 0, 0, field.value, scalars)
+            if sort:
+                alg.SortByLabelCount()
+            # vtkPackLabels does not always pass data correctly and will sometimes
```
MatthewFlamm, line 6432: Because of this choice to copy later, we have to have three mesh objects when `inplace=False`.  `self`, `output` and `result`.  Is there a way to avoid this?  My guess is that this filter could be used with large datasets.

user27182: This was implemented a while ago and I cannot recall exactly what the issue was or if I wrote a test specifically to check for this issue (probably not). But I believe this was a bug which effectively made the filter act as if `inplace=True` by default, or some of the arrays present in the input would be missing from the output despite setting the `PassData` vars to `On`. I'll have to look into it. Definitely agree copies should be avoid if possible though.

pyvista/core/filters/data_set.py
```diff
@@ -6239,6 +6239,236 @@ def extract_cells_by_type(self, cell_types, progress_bar=False):
...
+        # Determine output scalars
+        if output_scalars is None:
+            output_scalars = scalars
+        elif isinstance(output_scalars, str):
+            output_scalars = output_scalars
+        else:
+            raise TypeError(f"Output scalars must be a string, got {type(output_scalars)} instead.")
+
+        # Do packing
+        if _vtk.VTK93:  # pragma: no cover
```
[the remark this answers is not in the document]
user27182: This should be resolved in https://github.com/pyvista/pyvista/pull/4802/commits/00b654346e85b5edb406855483ddb60cc99b5526

pyvista/core/filters/data_set.py
user27182: After some testing I found that the issue is that `vtkPackLabels` by default does not pass the input array through to the output. The implementation has been updated in [31bc44c](https://github.com/pyvista/pyvista/pull/4802/commits/31bc44cce20ce0ddb1056a1df48f8b22982b25cf) to remove the copying that was done before. Now, only the input array is copied through to the output.

MatthewFlamm (approved):
LGTM.  This will be a great addition to the upcoming release. Thanks!

tkoyama010 (approved):

PR #4813, opened 2023-08-29
Remove comment about interactive

tkoyama010:
### Overview

Example (and many others) is not interactive anymore.

### Details

- None


akaszynski (approved):

akaszynski:
Good catch.

PR #4816, opened 2023-08-29
Fix bug in the documentation of `extract_surface`

chemiskyy:
### Overview

This PR corrects a bug in the documentation of the function "extract_surface" : The array name ("vtkOriginalCellIds" instead of vtkOriginalPointIds), and adds an example to access those arrays


akaszynski (approved):
Thanks for your work!

Issue #4750, opened 2023-08-08
Plotter AttributeError: 'NoneType' object has no attribute 'add_key_event'


gsal:
### Describe the bug, what's wrong, and what you expected.

I think I have found a bug!

Traceback (most recent call last):
  File "/path/to/.../file.py", line 10, in <module>
    p.add_key_event("q",p.close())
  File "/path/to/.../.venv/lib64/python3.9/site-packages/pyvista/plotting/plotter.py", line 1878, in add_key_event
    self.iren.add_key_event(*args, **kwargs)
AttributeError: 'NoneType' object has no attribute 'add_key_event'

### Steps to reproduce the bug.

```python
import pyvista as pv
from pyvista import examples

vol = examples.download_brain()
p = pv.Plotter()
p.add_key_event("q",p.close())
p.show()
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Mon Aug 07 22:25:21 2023 EDT

                  OS : Linux
              CPU(s) : 48
             Machine : x86_64
        Architecture : 64bit
                 RAM : 1511.0 GiB
         Environment : Python
         File system : xfs
          GPU Vendor : NVIDIA Corporation
        GPU Renderer : Quadro RTX 6000/PCIe/SSE2
         GPU Version : 4.5.0 NVIDIA 470.182.03
    MathText Support : False

  Python 3.9.16 (main, May 29 2023, 00:00:00)  [GCC 11.3.1 20221121 (Red Hat
  11.3.1-4)]

             pyvista : 0.41.1
                 vtk : 9.2.6
               numpy : 1.25.2
          matplotlib : 3.7.2
              scooby : 0.7.2
               pooch : v1.7.0
             imageio : 2.31.1
           pyvistaqt : 0.11.0
               PyQt5 : 5.15.9
             IPython : 8.14.0
            colorcet : 3.0.1
             cmocean : v3.0.3
          ipyvtklink : 0.2.3
          ipywidgets : 7.8.0
                tqdm : 4.65.0
              meshio : 5.3.4
          jupyterlab : 4.0.4
           pythreejs : 2.4.2
               trame : 3.1.0
        trame_client : 2.10.0
        trame_server : 2.11.7
           trame_vtk : 2.5.8
jupyter_server_proxy : 4.0.0
        nest_asyncio : 1.5.7
--------------------------------------------------------------------------------
```

tkoyama010:
[...]

PR #4820, opened 2023-08-30
Raise error when callback has empty arguments in `add_key_event`

tkoyama010:
### Overview

This PR is to save us in #4750. We were too lazy to read the documentation and it took us quite some time to solve the problem.

### Details

- None

### DIFF
--- a/pyvista/plotting/render_window_interactor.py
+++ b/pyvista/plotting/render_window_interactor.py
@@ -105,7 +105,7 @@ class RenderWindowInteractor:
             raise TypeError('callback must be callable.')
         for param in signature(callback).parameters.values():
             if param.default is param.empty:
-                TypeError('`callback` must not have any arguments without default values.')
+                raise TypeError('`callback` must not have any arguments without default values.')
         self._key_press_event_callbacks[key].append(callback)
 
     @staticmethod
--- a/pyvista/plotting/render_window_interactor.py
+++ b/pyvista/plotting/render_window_interactor.py
@@ -2,6 +2,7 @@
 import collections.abc
 from contextlib import contextmanager
 from functools import partial
+from inspect import signature
 import logging
 import time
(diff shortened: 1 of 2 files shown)


pyvista/plotting/render_window_interactor.py
```diff
@@ -102,6 +103,8 @@ def add_key_event(self, key, callback):
         """
         if not callable(callback):
             raise TypeError('callback must be callable.')
+        if len(signature(callback).parameters) > 0:
```
dcbr, line 106: Isn't this a bit too restrictive? If the callback provides default values for all arguments, it shouldn't be a problem right? (But I'm not sure how you would check that)

tkoyama010: Thank you for your comment. That is a very good point. I had completely missed it's use case. The following code will accomplish that.
```suggestion
        for param in signature(callback).parameters.values():
            if (param.default is param.empty):
```

dcbr (approved):
LGTM, only a small suggestion for the docs; feel free to ignore.

pyvista/plotting/render_window_interactor.py
```diff
@@ -97,11 +98,14 @@ def add_key_event(self, key, callback):
...
         callback : callable
-            A callable that takes no arguments.
+            A callable that takes no empty arguments.
 
         """
         if not callable(callback):
             raise TypeError('callback must be callable.')
+        for param in signature(callback).parameters.values():
+            if param.default is param.empty:
+                raise TypeError('callback must not have any empty arguments.')
```
akaszynski, line 108:
```suggestion
                TypeError(f'`callback` must not have any arguments without default values.')
```

akaszynski (approved):

dcbr (approved):

PR #4834, opened 2023-09-01
Move version change step before doing test in release

tkoyama010:
### Overview

I moved the version change step before testing and document build verification because of errors due to version changes like #4832.

### Details

- None


akaszynski (approved):
Good point. It's best that we catch these errors prior to pushing the release branch since we check for deprecations.

PR #4610, opened 2023-06-30
Add numpydoc to .pre-commit-config.yaml

tkoyama010:
### Overview

Let's use how it works.
resolves #4608 

### Details

- < feature1 or bug1 description >
- < feature2 or bug2 description >

PR #4840, opened 2023-09-03
Fix numpydoc ignore comment

tkoyama010:
### Overview

Follow-up PR of #4610.

### Details

- None


akaszynski (approved):

tkoyama010:
The 24-hour rule does not apply to amendments with no contents to consider.

PR #4841, opened 2023-09-04
Deploy dev docs on non-tagged push

akaszynski:
Stable documentation fails to push because the GH action runner runs out of space. By disabling the deployment of the development documentation when tagged, we should be able to avoid this issue.

Recommend merging this and cherry picking into the release branch, deleting the tag, and then pushing the tag again to trigger another release build.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -128,7 +128,7 @@ jobs:
           path: .
 
       - name: Deploy Dev Docs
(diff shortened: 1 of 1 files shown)


tkoyama010:
> Recommend merging this and cherry picking into the release branch, deleting the tag, and then pushing the tag again to trigger another release build.

Thanks! I will.

tkoyama010 (approved):

Issue #4854, opened 2023-09-06
Feature request: Possibility to add items to trame toolbar in jupyter environments


FelipeCybis:
### Describe the feature you would like to be added.

Trame has a lot of cool examples using sliders and buttons and stuff (https://kitware.github.io/trame/docs/cheatsheet.html) but in PyVista we are a bit constrained since we are far from the creation of the Vuetifty layout behind the trame plot in a jupyter environment.
By allowing users to add Vuetify widgets on the trame menu could be a good step towards a higher flexibility on jupyter/trame apps.

This would require a new argument `add_menu_items` in the `Viewer` classes. The argument can be a function similar to `Viewer.ui_controls()` that would be called right after `Viewer.ui_controls()`.

With a little tweaking of the arguments on `show_trame` and `initialize` this can be done fairly easily.

### Pseudocode or Screenshots

With my local changes I can do more or less what I wanted:
```python
import asyncio

import pyvista as pv
import vtk

from pyvista.trame.ui.vuetify2 import button
from trame.widgets import html, vuetify as vuetify

def slider(model, tooltip, **kwargs):
    """Create a vuetify slider."""
    with vuetify.VTooltip(bottom=True):
        with vuetify.Template(v_slot_activator='{ on, attrs }'):
            with html.Div(v_on='on', v_bind='attrs'):
                vuetify.VSlider(
                    v_model=model,
                    **kwargs
                )
        html.Span(tooltip)


def text_field(model, tooltip, **kwargs):
    """Create a vuetify text field."""
    with vuetify.VTooltip(bottom=True):
        with vuetify.Template(v_slot_activator='{ on, attrs }'):
            with html.Div(v_on='on', v_bind='attrs'):
                vuetify.VTextField(
                    v_model=model,
                    **kwargs
                )
        html.Span(tooltip)


def select(model, tooltip, **kwargs):
    with vuetify.VTooltip(bottom=True):
        with vuetify.Template(v_slot_activator='{ on, attrs }'):
            with html.Div(v_on='on', v_bind='attrs'):
                vuetify.VSelect(
                v_model=model,
                **kwargs
                )
        html.Span(tooltip)


def btn_play():
    state.play = not state.play
    state.flush()

def custom_tools():
    vuetify.VDivider(vertical=True, classes='mx-1')
    button(
        click=btn_play,
        icon='mdi-play',
        tooltip='Play',
    )
    
    slider(
        model=("resolution", 0),
        tooltip="Resolution slider",
[...]

PR #4839, opened 2023-09-03
Add `add_timer_event` helping make animation.

tkoyama010:
### Overview

Add `add_timer_event` helping make animation.

https://github.com/pyvista/pyvista/blob/6e8d8859b03dc353fcf38f33638fe406cf1f5831/examples/03-widgets/animation.py#L13-L28

https://github.com/pyvista/pyvista/assets/7513610/c571aab2-77df-4ee3-a93c-384c1558dba7

### Details

- This PR is preparing to solve #4402.

Issue #4402, opened 2023-05-10
Add Playback Widgets


tkoyama010:
### Describe the feature you would like to be added.

![PlaybackWidget](https://vtk.org/Wiki/images/4/47/PlaybackWidget.png)

### Links to VTK Documentation, Examples, or Class Definitions.

[classvtkPlaybackWidget](https://vtk.org/doc/release/6.2/html/classvtkPlaybackWidget.html)

### Pseudocode or Screenshots

```python
p.add_playback_widget(callback=simulate, use_vertices=True)
```

banesullivan:
Great idea! I thought I'd add that I use playback ipywidgets often and have an example here: https://github.com/brendanjmeade/kale/blob/173e7b3fb43dff0939ca9954f4aa9ab06372c27b/kale/widgets.py#L8-L34

This works pretty easily with the recently added algorithm support in PyVista. Here is an example in Jupyter:

```py
import pyvista as pv
import numpy as np
import ipywidgets as widgets
import vtk

def time_controls(callback, tmin=3, tmax=20, step=2):

    def set_time(change):
        step = change["new"]
        if step < 0:
            step = 0
        if step >= tmax:
            step = tmax - 1
        callback(step)

    play = widgets.Play(
        value=tmin,
        min=tmin,
        max=tmax,
        step=step,
        description="Time Step",
    )
    play.observe(set_time, "value")

    slider = widgets.IntSlider(min=tmin, max=tmax, step=step, continuous_update=True)
    widgets.jslink((play, "value"), (slider, "value"))
    return widgets.HBox([play, slider])



plotter = pv.Plotter()
algo = vtk.vtkConeSource()
plotter.add_mesh(algo, color="orange")

def update_resolution(time_step):
    algo.SetResolution(int(time_step))
    plotter.render()
    
w = time_controls(update_resolution)
plotter.show()
w
```

https://github.com/pyvista/pyvista/assets/22067021/801f77ef-1f44-4aa6-9cfd-9edd1bbfc01f

PR #4855, opened 2023-09-06
Allow some customization to trame toolbar

FelipeCybis:
### Overview

Resolves https://github.com/pyvista/pyvista/issues/4854. The PR allows adding vuetify items to the trame toolbar in jupyter environments. The idea is to bring more of the power of `trame` to the jupyter view.

Inspired/related by https://github.com/pyvista/pyvista/pull/4839 and https://github.com/pyvista/pyvista/issues/4402.

### Details
Added `add_menu_items` in the vuetify2 and 3 `Viewer` class. This argument should be a function similar to what the `Viewer.ui_controls()` is so it can be passed just after `Viewer.ui_controls()` when building the `VCard` of the trame's toolbar.
Passing `add_menu_items` in the `jupyter_kwargs` of the `show` method and adding `add_menu_items` to `show_trame` is enough so that this argument can reach the `ui` method of the viewers.

### Example
```python
import asyncio

import pyvista as pv
import vtk

from pyvista.trame.ui.vuetify2 import button, slider, text_field, select
from trame.widgets import vuetify

def btn_play():
    state.play = not state.play
    state.flush()

def custom_tools():
    vuetify.VDivider(vertical=True, classes='mx-1')
    button(
        click=btn_play,
        icon='mdi-play',
        tooltip='Play',
    )
    
    slider(
        model=("resolution", 0),
        tooltip="Resolution slider",
        min=3,
        max=20,
        step=1,
        dense=True,
        hide_details=True,
        style="width: 300px",
        classes='my-0 py-0 ml-1 mr-1',
        )
    text_field(
        model=("resolution", 0),
        tooltip="Resolution value",
        readonly=True,
        type="number",
        dense=True,
        hide_details=True,
        style="min-width: 40px; width: 60px",
        classes='my-0 py-0 ml-1 mr-1',
        )
    
    vuetify.VDivider(vertical=True, classes='mx-1')
    select(
        model=("visibility", "Show"),
        tooltip="Toggle visibility",
        items=['Visibility', ["Hide", "Show"]],
        hide_details=True,
        dense=True,
        )
    
pl = pv.Plotter()
algo = vtk.vtkConeSource()
mesh_actor = pl.add_mesh(algo)

viewer = pl.show(jupyter_kwargs=dict(add_menu_items=custom_tools), return_viewer=True)

state, ctrl = viewer.viewer.server.state, viewer.viewer.server.controller
state.play = False
ctrl.view_update = viewer.viewer.update

# trame callbacks
@state.change("play")
async def _play(play, **kwargs):
    while state.play:
        state.resolution += 1
        state.flush()
        if state.resolution >= 20:
            state.play = False
        await asyncio.sleep(0.3)

@state.change("resolution") 
def update_resolution(resolution, **kwargs):
    algo.SetResolution(resolution)
    ctrl.view_update()

@state.change("visibility") 
def set_visibility(visibility, **kwargs):
    toggle = {"Hide": 0, "Show":1}
    mesh_actor.SetVisibility(toggle[visibility])
    ctrl.view_update()


viewer
```

https://github.com/pyvista/pyvista/assets/41338087/d6ba752a-b158-4441-83b0-3e638529c6ef

### DIFF
--- a/examples/99-advanced/customization-trame-toolbar.py
+++ b/examples/99-advanced/customization-trame-toolbar.py
@@ -1,0 +1,84 @@
+import asyncio
+
+import pyvista as pv
+import vtk
+
+from pyvista.trame.ui.vuetify2 import button, slider, text_field, select
+from trame.widgets import vuetify
+
+def btn_play():
+    state.play = not state.play
+    state.flush()
+
+def custom_tools():
+    vuetify.VDivider(vertical=True, classes='mx-1')
+    button(
+        click=btn_play,
+        icon='mdi-play',
+        tooltip='Play',
+    )
+    
+    slider(
+        model=("resolution", 0),
+        tooltip="Resolution slider",
+        min=3,
+        max=20,
+        step=1,
+        dense=True,
+        hide_details=True,
+        style="width: 300px",
+        classes='my-0 py-0 ml-1 mr-1',
+        )
+    text_field(
+        model=("resolution", 0),
+        tooltip="Resolution value",
+        readonly=True,
+        type="number",
+        dense=True,
+        hide_details=True,
+        style="min-width: 40px; width: 60px",
+        classes='my-0 py-0 ml-1 mr-1',
+        )
+    
+    vuetify.VDivider(vertical=True, classes='mx-1')
+    select(
+        model=("visibility", "Show"),
+        tooltip="Toggle visibility",
+        items=['Visibility', ["Hide", "Show"]],
+        hide_details=True,
+        dense=True,
+        )
+    
+pl = pv.Plotter()
+algo = vtk.vtkConeSource()
+mesh_actor = pl.add_mesh(algo)
+
+viewer = pl.show(jupyter_kwargs=dict(add_menu_items=custom_tools), return_viewer=True)
+
+state, ctrl = viewer.viewer.server.state, viewer.viewer.server.controller
+state.play = False
+ctrl.view_update = viewer.viewer.update
+
+# trame callbacks
+@state.change("play")
+async def _play(play, **kwargs):
+    while state.play:
+        state.resolution += 1
+        state.flush()
+        if state.resolution >= 20:
+            state.play = False
+        await asyncio.sleep(0.3)
+
+@state.change("resolution") 
+def update_resolution(resolution, **kwargs):
+    algo.SetResolution(resolution)
+    ctrl.view_update()
+
+@state.change("visibility") 
+def set_visibility(visibility, **kwargs):
+    toggle = {"Hide": 0, "Show":1}
+    mesh_actor.SetVisibility(toggle[visibility])
+    ctrl.view_update()
+
+
+viewer
--- a/examples/99-advanced/customization-trame-toolbar.py
+++ b/examples/99-advanced/customization-trame-toolbar.py
@@ -118,7 +118,7 @@ ctrl.view_update = widget.viewer.update
 # a timer so that an animation can be set to play. To do that with ``trame`` we need
 # to have an asynchronous function so we can continue to do stuff while the
 # "timer" function is running. The ``_play`` function will be called when the ``play``
-# variable is changed (when we click the play button, through the ``btn_play``
+# variable is changed (when we click the play button, through the ``button_play``
 # callback). While ``state.play`` is ``True`` we want to play the animation. We
 # change the ``state.resolution`` value, but to really call the ``update_resolution``
 # function we need to ``flush`` the server and force it to see the change in
--- a/examples/99-advanced/customization-trame-toolbar.py
+++ b/examples/99-advanced/customization-trame-toolbar.py
@@ -1,5 +1,5 @@
 """
-.. _customize_trame_toolbar_example
+.. _customize_trame_toolbar_example:
 
 Customize Trame toolbar
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/examples/99-advanced/customization-trame-toolbar.py
+++ b/examples/99-advanced/customization-trame-toolbar.py
@@ -8,21 +8,24 @@ Bring more of the power of trame to the jupyter view.
 """
 import asyncio
 
-import pyvista as pv
+from trame.widgets import vuetify
 import vtk
 
-from pyvista.trame.ui.vuetify2 import button, slider, text_field, select
-from trame.widgets import vuetify
+import pyvista as pv
+from pyvista.trame.ui.vuetify2 import button, select, slider, text_field
 
 ###############################################################################
 # Here is an explanation of the `btn_play`.
+
 
 def btn_play():
     state.play = not state.play
     state.flush()
 
+
 ###############################################################################
 # Here is an explanation of the custom tools.
+
 
 def custom_tools():
     vuetify.VDivider(vertical=True, classes='mx-1')
@@ -31,7 +34,7 @@ def custom_tools():
         icon='mdi-play',
         tooltip='Play',
     )
-    
+
     slider(
         model=("resolution", 0),
         tooltip="Resolution slider",
@@ -42,7 +45,7 @@ def custom_tools():
         hide_details=True,
         style="width: 300px",
         classes='my-0 py-0 ml-1 mr-1',
-        )
+    )
     text_field(
         model=("resolution", 0),
         tooltip="Resolution value",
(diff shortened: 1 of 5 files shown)


FelipeCybis:
Obs.: I made a mistake in the tests, going to patch it ASAP. Sorry for that.

tkoyama010:
Fantastic! I was trying to do this with Playback Widgets, but your solution is quite cooler!

examples/99-advanced/customization-trame-toolbar.py
```diff
@@ -0,0 +1,106 @@
...
+import asyncio
+
+import pyvista as pv
+import vtk
+
+from pyvista.trame.ui.vuetify2 import button, slider, text_field, select
+from trame.widgets import vuetify
+
+###############################################################################
+# Here is an explanation of the `btn_play`.
```
tkoyama010, line 18: Please add an explanation here.

tkoyama010 (changes requested):
@FelipeCybis Thanks for the great PR. I have added your code to the example because we want to understand it better about your code. Could you please add your detailed description in the comments section of the code? That way we can discuss how we can improve the code, including the UI.

tkoyama010:
pre-commit.ci autofix

tests/plotting/jupyter/test_trame.py
```diff
@@ -3,16 +3,30 @@
 from IPython.display import IFrame
 import numpy as np
 import pytest
+from vtk import vtkConeSource
```
tkoyama010, line 6: It maybe time to add ConeSource class in PyVista, but it is another PR.

pyvista/trame/ui/vuetify3.py
```diff
@@ -43,6 +43,33 @@ def checkbox(model, icons, tooltip):  # numpydoc ignore=PR01
...
+
+def select(model, tooltip, **kwargs):  # numpydoc ignore=PR01
+    """Create a vuetify select menu."""
+    with vuetify.VTooltip(bottom=True):
+        with vuetify.Template(v_slot_activator='{ props }'):
+            with html.Div(v_bind='props'):
+                vuetify.VSelect(v_model=model, **kwargs)
+        html.Span(tooltip)
+
+
```
tkoyama010, line 72: Cool! This is really what we need. It would have been a challenge for me to fully scratch up Trame's Vue wrapper, but this allows us to use a generic UI as a component!

FelipeCybis: yeah, even though vuetify is pretty well documented, there is a lot of trial and error on working with the trame's wrapper :P

FelipeCybis:
> Fantastic! I was trying to do this with Playback Widgets, but your solution is quite cooler! Edit: works fine in my local too.

I would love to chim in in the Timer/Playback Widgets discussion when I have a little bit of time. I am sure you noticed that the vtk `CreateTimer` does not exist on the `iren` used when `trame` is the viewer so that is when I use the `async` in my code to create a kind of `trame` Timer.

examples/99-advanced/customization-trame-toolbar.py
```diff
@@ -0,0 +1,155 @@
...
+from pyvista.trame.ui.vuetify2 import button, select, slider, text_field
+
+###############################################################################
+# Let's first create the menu items we want to add to the trame's toolbar.
+# Here we want a "play" button that will be later connected to a slider
+# through the ``btn_play`` function. The slider itself will represent the
+# "resolution" of the model we will render, a text field where the value of
+# the "resolution" will be displayed.
+# We will also add a dropdown menu to toggle the visibility of the model.
+# The dividers are the same already used to divide and organize the toolbar.
```
tkoyama010, line 24:
```suggestion
# Let's first create the menu items we want to add to the trame's toolbar.
# Here we want a "play" button that will be later connected to a slider
# through the ``btn_play`` function. The slider itself will represent the
# "resolution" of the model we will render, a text field where the value of
# the "resolution" will be displayed.
# We will also add a dropdown menu to toggle the visibility of the model.
# The dividers are the same as already used to divide and organize the toolbar.
```

examples/99-advanced/customization-trame-toolbar.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def custom_tools():
+    vuetify.VDivider(vertical=True, classes='mx-1')
```
tkoyama010, line 28: I agree that VDivider is usable as is. However, the goal of PyVista is to be able to use the divider without the external import of trame. It would be a simple wrapper, but could you add a function like a button or slider?

examples/99-advanced/customization-trame-toolbar.py
```diff
@@ -0,0 +1,155 @@
...
+    button(
+        click=btn_play,
```
tkoyama010, line 30:
```suggestion
        click=button_play,
```

examples/99-advanced/customization-trame-toolbar.py
```diff
@@ -0,0 +1,155 @@
...
+
+###############################################################################
+# The button callback function ``btn_play`` needs to be created before starting
+# the server. This function will toggle the boolean state variable ``play``
+# and flush the server, i.e. "force" the server to see the change.
+# We will see more on the state variables in a bit, but we need to create the
+# function here otherwise the server will complain ``btn_play`` does not exist.
+
+
+def btn_play():
```
tkoyama010, line 75:
```suggestion
def button_play():
```

examples/99-advanced/customization-trame-toolbar.py
```diff
@@ -0,0 +1,155 @@
...
+        dense=True,
+    )
+
+
+###############################################################################
+# The button callback function ``btn_play`` needs to be created before starting
+# the server. This function will toggle the boolean state variable ``play``
+# and flush the server, i.e. "force" the server to see the change.
+# We will see more on the state variables in a bit, but we need to create the
+# function here otherwise the server will complain ``btn_play`` does not exist.
```
tkoyama010, line 72:
```suggestion
# The button callback function ``button_play`` needs to be created before starting
# the server. This function will toggle the boolean state variable ``play``
# and flush the server, i.e. "force" the server to see the change.
# We will see more on the state variables in a bit, but we need to create the
# function here otherwise the server will complain ``button_play`` does not exist.
```

examples/99-advanced/customization-trame-toolbar.py
```diff
@@ -0,0 +1,155 @@
...
+# to have an asynchronous function so we can continue to do stuff while the
+# "timer" function is running. The ``_play`` function will be called when the ``play``
+# variable is changed (when we click the play button, through the ``btn_play``
+# callback). While ``state.play`` is ``True`` we want to play the animation. We
+# change the ``state.resolution`` value, but to really call the ``update_resolution``
+# function we need to ``flush`` the server and force it to see the change in
+# the shared variables. When ``state.play`` changes to ``False``, the animation stops.
+#
+# Note that using ``while play: ...`` would not work here because it is not the
+# actual state variable, but only an argument value passed to the callback function.
```
tkoyama010, line 128:
```suggestion
# Now we can create the callback functions for our menu items.
#
# The functions are decorated with a ``state.change("variable")``. This means
# they will be called when this specific variable has its value changed in the
# server's shared state. When ``resolution`` changes, we want to update the
# resolution of our cone algorithm. When ``visibility`` changes, we want to toggle the
# visibility of our cone.
#
# The ``play`` variable is a little bit trickier. We want to start something like
# a timer so that an animation can be set to play. To do that with ``trame`` we need
# to have an asynchronous function so we can continue to do stuff while the
# "timer" function is running. The ``_play`` function will be called when the ``play``
# variable is changed (when we click the play button, through the ``button_play``
# callback). While ``state.play`` is ``True`` we want to play the animation. We
# change the ``state.resolution`` value, but to really call the ``update_resolution``
# function we need to ``flush`` the server and force it to see the change in
# the shared variables. When ``state.play`` changes to ``False``, the animation stops.
#
# Note that using ``while play: ...`` would not work here because it is not the
# actual state variable, but only an argument value passed to the callback function.
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM. This will change the game of PyVista.

tests/plotting/jupyter/test_trame.py
tkoyama010: See #4860 .

Issue #4849, opened 2023-09-06
`pyvista.DataSetFilters.extract_points` documentation does not represent the features of the method


tkoyama010:
### Describe what maintenance you would like added.

The following image from the [pyvista.DataSetFilters.extract_points](https://docs.pyvista.org/version/dev/api/core/_autosummary/pyvista.DataSetFilters.extract_points.html) documentation does not adequately represent the features of the method. Please modify the method options so that only points are shown.
![pyvista-DataSetFilters-extract_points-1_00_00](https://github.com/pyvista/pyvista/assets/7513610/7f7650d4-f3af-4362-aaca-f5e8415cd404)

### Links to source code.

https://github.com/pyvista/pyvista/blob/f0db627a9b3eecb0ae95a338fc06c5740ea868d6/pyvista/core/filters/data_set.py#L4637-L4645

### Pseudocode or Screenshots

Let's contribute!

PR #4856, opened 2023-09-06
Fix pyvista.DataSetFilters.extract_points documentation does not represent the features of the method

planmi1287:
### Overview
Resolves: #4849. The image from the documentation should be more adequate to the features of the method.

### Details
Switched the `include_cells` parameter from `True` (default) to `False`

### Example
pyvista-DataSetFilters-extract_points-1_00_00


akaszynski:
pre-commit.ci autofix

akaszynski (approved):
Thanks @planmi1287!

tkoyama010 (approved):

PR #4858, opened 2023-09-06
Add step to check tag before git push tag

tkoyama010:
### Overview

I accidentally pushed a tag named **v** today. This was due to an error when I ran the tag creation Python one-liner locally. Now when I copy and run the bash code in the document, it pushes even if the tag creation command is in error. This is very dangerous, so I decided to separate the steps. Let's prepare for the next time human error.

### Details

- None


adeak (approved):
LGTM, better to be safe than sorry.

PR #4865, opened 2023-09-07
Fix typos using `typos`

eggplants:
### Overview

I corrected some typos by following [codespell](https://github.com/codespell-project/codespell) / [typos](https://github.com/crate-ci/typos) suggestions.

### Details

- Ambigious -> Ambiguous
- implementd -> implemented
- Maping -> Mapping
- fom -> from

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,11 +121,6 @@ match = '(?!coverage).*.py'
 convention = "numpy"
 add-ignore = ["D404"]
 
-[tool.codespell]
-skip = '*.pyc,*.txt,*.gif,*.png,*.jpg,*.ply,*.vtk,*.vti,*.vtu,*.js,*.html,*.doctree,*.ttf,*.woff,*.woff2,*.eot,*.mp4,*.inv,*.pickle,*.ipynb,flycheck*,./.git/*,./.hypothesis/*,*.yml,doc/_build/*,./doc/images/*,./dist/*,*~,.hypothesis*,./doc/examples/*,*.mypy_cache/*,*cover,./tests/tinypages/_build/*,*/_autosummary/*'
-ignore-words = "doc/styles/Vocab/pyvista/accept.txt"
-quiet-level = 3
-
 [tool.coverage.run]
 omit = [
     'pyvista/ext/coverage.py',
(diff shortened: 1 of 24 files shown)


tkoyama010 (approved):
Thanks for this. It's a great idea to check typos using typos. I added it in `pre-commit`.

MatthewFlamm:
What is difference between codespell and typos?  Is there a benefit to having both?  The more checks that get added to pre-commit, the slower it is.  The checks themselves may be fast, but I find it annoying to have to install the environments and wait for that.

eggplants:
I prefer to typos than codespell. It's written by Rust and [the benchmark](https://github.com/crate-ci/typos/blob/master/benchsuite/runs/2021-05-21-epage-11562.md) shows about x5 faster than codespell.

tkoyama010:
@eggplants What about accuracy？
Are typos as accurate or more accurate than codespells?

eggplants:
@tkoyama010 I could not google for any precedent for quantitative comparative verification of the accuracy of the correction.
If you have time, it would be good to try and compare the tools with the different approaches listed in [the comparison table](https://github.com/crate-ci/typos/blob/master/docs/comparison.md).

tkoyama010:
Thanks. Personally, I think it would be interesting to remove codespell and try typos. If we are not satisfied with the accuracy of the typos, we can REVERT it after. I would like to publish this as a PR with codespell extracted and see how other developers react.

pyproject.toml
```diff
@@ -121,11 +121,6 @@ match = '(?!coverage).*.py'
 convention = "numpy"
 add-ignore = ["D404"]
 
-[tool.codespell]
-skip = '*.pyc,*.txt,*.gif,*.png,*.jpg,*.ply,*.vtk,*.vti,*.vtu,*.js,*.html,*.doctree,*.ttf,*.woff,*.woff2,*.eot,*.mp4,*.inv,*.pickle,*.ipynb,flycheck*,./.git/*,./.hypothesis/*,*.yml,doc/_build/*,./doc/images/*,./dist/*,*~,.hypothesis*,./doc/examples/*,*.mypy_cache/*,*cover,./tests/tinypages/_build/*,*/_autosummary/*'
-ignore-words = "doc/styles/Vocab/pyvista/accept.txt"
-quiet-level = 3
-
```
tkoyama010, line 128: We need to check if typos require these settings.

tkoyama010:
I found that there are still some things to consider in the transition. Therefore, I will keep it As-Is and merge only the typo found. Let's consider migrating when major projects like Matplotlib and NumPy migrate.
In any case, thank you for your contribution.

tkoyama010 (approved):

PR #4866, opened 2023-09-08
Improve test coverage of `Plotter.add_floor`

tkoyama010:
### Overview

This is the follow-up PR of #4865. The purpose of PR is to improve the test coverage of `Plotter.add_floor`.

### Details

- None

### DIFF
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -1948,7 +1948,7 @@ class Renderer(_vtk.vtkOpenGLRenderer):
             i_size = ranges[2]
             j_size = ranges[1]
         else:
-            raise NotImplementedError(f'Face ({face}) not implementd')
+            raise NotImplementedError(f'Face ({face}) not implemented)
         self._floor = pyvista.Plane(
             center=center,
             direction=normal,
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -1948,7 +1948,7 @@ class Renderer(_vtk.vtkOpenGLRenderer):
             i_size = ranges[2]
             j_size = ranges[1]
         else:
-            raise NotImplementedError(f'Face ({face}) not implemented)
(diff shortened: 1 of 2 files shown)


pyvista/plotting/renderer.py
```diff
@@ -1948,7 +1948,7 @@ def add_floor(
             i_size = ranges[2]
             j_size = ranges[1]
         else:
-            raise NotImplementedError(f'Face ({face}) not implementd')
+            raise NotImplementedError(f'Face ({face}) not implemented)
```
tkoyama010, line 1951:
```suggestion
            raise NotImplementedError(f'Face ({face}) not implemented')
```

tests/plotting/test_plotter.py
```diff
@@ -402,3 +402,22 @@ def test_plotter_meshes(sphere, cube):
...
+        ('+Y', (0, -1, 0)),
+        ('+X', (-1, 0, 0)),
+    ],
+)
+def test_plotter_add_floor(face, normal):
+    pl = pv.Plotter()
+    pl.add_floor(face=face)
+    assert np.allclose(pl.renderer._floor.face_normals[0], normal)
+    with pytest.raises(NotImplementedError, match='not implemented'):
+        pl.add_floor(face='invalid')
```
MatthewFlamm, line 423: small comment:  not sure if there is a better way than testing this in every parameterized test.  A failure here will result in 6 failed tests even though it is logically one failure.

MatthewFlamm (commented):
An image regression test might be useful, but I think the tests you have here are sufficient for now.  One small comment for consideration, but the current PR state is also fine with me.

tests/plotting/test_plotter.py
tkoyama010: Good point. I was too lazy and skipped separating functions, which was not a good thing.

MatthewFlamm (approved):
LGTM

PR #4879, opened 2023-09-14
Make composite mapper dataset optional

akaszynski:
Existing `CompositePolyDataMapper` implementation makes actor copy impossible:

```py
>>> import pyvista as pv
>>> mblock = pv.MultiBlock([pv.Sphere()])
>>> pl = pv.Plotter()
>>> actor, _ = pl.add_composite(mblock)
>>> actor.copy()
...
~/.venv311/lib/python3.11/site-packages/pyvista/plotting/mapper.py in copy(self)
...
TypeError: CompositePolyDataMapper.__init__() missing 1 required positional argument: 'dataset'
```

This PR allows the dataset to be optional to the mapper, which is then added later in `Actor.copy`. Since we don't test copying actors with composite datasets, I added a test as well for good measure even though this doesn't improve coverage.

### DIFF
--- a/pyvista/plotting/composite_mapper.py
+++ b/pyvista/plotting/composite_mapper.py
@@ -521,7 +521,7 @@ class CompositePolyDataMapper(_vtk.vtkCompositePolyDataMapper2, _BaseMapper):
 
     Parameters
     ----------
-    dataset : pyvista.MultiBlock
+    dataset : pyvista.MultiBlock, optional
         Multiblock dataset.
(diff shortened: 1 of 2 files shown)


tkoyama010 (approved):
LGTM

akaszynski:
I'll need this in a patch. Any other PRs we should include?

tkoyama010:
Nothing but this. I will release it immediately.

PR #4860, opened 2023-09-07
Add `ConeSource` class

tkoyama010:
### Overview

This is a follow-up PR of #4855. See https://github.com/pyvista/pyvista/pull/4855#discussion_r1318080773 .

### Details

- None

PR #4890, opened 2023-09-16
Add `CylinderSource` class

tkoyama010:
### Overview

- Follow-up PR of #4860 . Add `CylinderSource` class.
- Move `translate` function from `geometric_objects.py` to `geometric_sources.py` .

### Details

- None

### DIFF
--- a/pyvista/core/utilities/geometric_sources.py
+++ b/pyvista/core/utilities/geometric_sources.py
@@ -5,10 +5,53 @@ Also includes some pure-python helpers.
 """
 from typing import Sequence
 
+import numpy as np
+
 from pyvista.core import _vtk_core as _vtk
 from pyvista.core.utilities.misc import no_new_attr
 
 from .helpers import wrap
+
+
+def translate(surf, center=(0.0, 0.0, 0.0), direction=(1.0, 0.0, 0.0)):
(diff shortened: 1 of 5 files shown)


tkoyama010:
LGTM

pyvista/core/utilities/geometric_sources.py
```diff
@@ -252,3 +295,208 @@ def output(self):
...
+        """Get the output data object for a port on this algorithm.
+
+        Returns
+        -------
+        pyvista.PolyData
+            Cylinder surface.
+        """
+        self.Update()
+        output = wrap(self.GetOutput())
+        output.rotate_z(-90, inplace=True)
```
MatthewFlamm, line 500: Sorry I wasn't able to review this timely.  I have not tested this, but due to the rotation and translation in this method, will could we get different result in these two cases for `cyl_source = pv.CylinderSource()`, `cyl_source.output.plot()` and `pl.add_mesh(cyl_source)`.  I think the latter will not have this rotation and translation?

tkoyama010: We are always happy to review, no matter when. Please review whenever you like.
And very good point. This is a bug and I have created #4956 .

PR #4947, opened 2023-09-21
Add `principal_axes` feature, add `axes_rotation` feature

### LINKED ISSUE #4705: Add `covariance` method to `fit_plane_to_points` and expose `Plane` arguments

This is a follow-up to #4650.

Currently, `fit_plane_to_points` uses SVD. However, calculating fit using covariance may be more "VTK-like" since:
[`vtkPlane`](https://vtk.org/doc/nightly/html/classvtkPlane.html#a7b88495d93a5aa5053b1044965df67bd) has a fitting method which uses covariance, and [`vtkOBBTree`](https://vtk.org/doc/nightly/html/classvtkOBBTree.html) also computes fit using covariance. I provided a graphical example of the different fit methods [here](https://github.com/pyvista/pyvista/issues/555#issuecomment-1646108382).

### Pseudocode or Screenshots

I propose adding `method='covariance'` as the default arg for `fit_plane_to_points` and having `method='svd'` as a non-default alternative.

In addition, sometimes it's useful to extract only the corners of the best-fit plane. In this case, we want to set `i_resolution=1` and `j_resolution=1` in the `Plane` constructor so that the plane mesh only has four points. However, we don't have direct access to these parameters in `fit_plane_to_points`. So I think we should expose these parameters directly.

This would look something like:
``` python 
def fit_plane_to_points(
    points,
    method='covariance',
    return_meta=False,
     i_resolution=10, 
    j_resolution=10):
```

Any thoughts on this or suggestions for an alternative API (e.g. use a `plane_parameters` dict instead)? Should the default resolution be 1 instead of 10?

### LINKED ISSUE #4711: `ImageData().points` raises unexpected ValueError

`ImageData().points` raises an unexpected `ValueError`. I would have expected that an empty array be returned, which is the behaviour with `PolyData().points` and `UnstructuredGrid().points`

Running:
```python
import pyvista
pyvista.ImageData().points
```
Will produce:
```
  File "\pyvista\lib\site-packages\IPython\core\interactiveshell.py", line 3508, in run_code
  File "<ipython-input-5-c3545438d535>", line 1, in <module>
    im.points
  File "c\pyvista\pyvista\core\grid.py", line 629, in points
    x = np.insert(np.cumsum(np.full(nx, dx)), 0, 0.0) + ox
  File "\pyvista\lib\site-packages\numpy\core\numeric.py", line 329, in full
    a = empty(shape, dtype, order)
ValueError: negative dimensions are not allowed
```

Compare this to:
```python
import pyvista
pyvista.PolyData().points  # pyvista_ndarray([], shape=(0, 3), dtype=float64)
pyvista.UnstructuredGrid().points  # pyvista_ndarray([], shape=(0, 3), dtype=float64)
```

### System Information

```shell
OS : Windows
            CPU(s) : 12
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.8 GiB
       Environment : IPython
        GPU Vendor : Intel
  MathText Support : False
  bit (AMD64)]
           pyvista : 0.42.dev0
               vtk : 9.2.6
             numpy : 1.25.1
        matplotlib : 3.7.2
            scooby : 0.7.2
             pooch : v1.7.0
           imageio : 2.31.1
           IPython : 8.14.0
          colorcet : 3.0.1
           cmocean : v3.0.3
        ipyvtklink : 0.2.3
        ipywidgets : 7.7.5
             scipy : 1.11.1
              tqdm : 4.65.0
            meshio : 5.3.4
        jupyterlab : 4.0.3
         pythreejs : 2.4.2
    pytest_pyvista : 0.1.8
             trame : 2.5.2
      trame_client : 2.9.4
      trame_server : 2.11.7
         trame_vtk : 2.5.8
      nest_asyncio : 1.5.6
```

### Screenshots

_No response_

user27182:
### Overview

This PR does:
#### Add `principal_axes` feature
- [x] add a new function `pv.principal_axes_vectors()` to compute the principal axes (aka principal components) of a set of points.
- [x] add the property `pv.DataSet.principal_axes`
- [x] add a convenience function `pv.principal_axes_transform()`
- [x] update `pv.fit_plane_to_points()` to call `pv.principal_axes_vectors()` internally
- [x] add `pv.random_sample_points()`, which is used by  `principal_axes_vectors` internally
- [x] resolves #4705 and exposes the `i_resolution` and `j_resolution` params from `pv.Plane()` to `pv.fit_plane_to_points()`
- [x] resolves #4711
- [ ] add new standalone example

#### Add `axes_rotation` feature
- [x] add `_coerce_transformlike_arg` function to standardize and validate acceptable inputs for transformations
- [x] refactor `data_set.transform`, `transformations.apply_transformation_to_points`, and `plotting.prop3d.user_matrix` to make use of `_coerce_transformlike_arg `
- [x] add `transformations.axes_rotation_matrix` function to create a 4x4 transformation from a 3x3 rotation matrix (or 3x3 axes vectors) which will cause rotation about a specified point
- [x] add `helpers.axes_rotation` as a public API to make use of  `axes_rotation_matrix`
- [x] add `DataSet.rotate_axes` method to make use of  `axes_rotation_matrix`

#### Rework `axes_actor`
- [ ] Merge #5029

#### To-Do
- [ ] Review and/or make examples for all new public methods

### Details
The initial motivation for adding `principal_axes` is described below. The feature has many useful applications, e.g. point cloud registration problems. Adding `axes_rotation` is helpful in cases to change basis vectors, e.g. from XYZ to principal axes. Reworking `axes_actor` is done to add scaling functionality and to enable the direct use of the principal axes transform to the axes actors to orient them in space for plotting purposes.

This PR started off as adding a _new_ method to `pv.fit_plane_to_points()` as in #4705 to use covariance (from PCA) instead of SVD. The thought was that these were actually different methods. However, after implementing this "new" method, the tests revealed that the best-fit plane for the "different" methods was actually the same plane. The key finding is that:
``` python
# SVD
 _, _, axes_vectors = np.linalg.svd(data)  # row vectors, descending order
```
is mathematically the same as:
``` python
# PCA
covariance = np.cov(data, rowvar=False) 
 _, axes_vectors = np.linalg.eigh(covariance)  # column vectors, ascending order 
axes_vectors = axes_vectors.T[::-1]  # row vectors, descending order
```
By the same, I mean that these are equivalent up to a difference in axis sign, difference in axes values due to non-uniqueness of  the vectors, and up to a difference in numerical error. PCA is indeed a special case of SVD, and SVD is preferred for precision/numerical stability since it does not require computing the covariance matrix. 

This [article](https://arxiv.org/pdf/1404.1100.pdf) gives a good technical description of these methods and their relation, and in Section C remarks that "the columns of *V* [from SVD] are the principal components of *X*". Indeed, [`sklearn.decomposition.PCA`](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html) states that it uses SVD to compute the principal components, further demonstrating that these methods are equivalent for this application.

As such, `pv.fit_plane_to_points()` now continues to only have and make use of the single method.


user27182:
For some context, here's a sample output from the `pv.principal_axes_transform()` docstring. The transform is used to align the mesh to the XYZ axes.
image

pyvista/core/utilities/points.py
```diff
@@ -179,11 +182,450 @@ def lines_from_points(points, close=False):
...
+    points,
+    return_meta=False,
+    i_resolution=10,
+    j_resolution=10,
+    normal_direction=None,
+):
+    """Fit a plane to a set of points.
+
+    The plane is fitted to the points using :func:~pyvista.principal_axes_vectors,
+    and is automatically sized to fit the extents of the points.
```
tkoyama010, line 619:
```suggestion
    and is automatically sized to fit the extent of the points.
```

pyvista/core/utilities/points.py
```diff
@@ -209,10 +673,11 @@ def fit_plane_to_points(points, return_meta=False):
     --------
     Fit a plane to a random point cloud.
 
-    >>> import pyvista
+    import numpy.random    >>> import pyvista
```
tkoyama010, line 676:
```suggestion
    >>> import numpy.random
    >>> import pyvista
```

pyvista/core/utilities/points.py
```diff
@@ -209,10 +673,11 @@ def fit_plane_to_points(points, return_meta=False):
...
     >>> import numpy as np
     >>>
     >>> # Create point cloud
+    >>> np.random.seed(42)
```
tkoyama010, line 680: Great. We can fix the output figure by this.

pyvista/core/utilities/points.py
```diff
@@ -179,11 +182,450 @@ def lines_from_points(points, close=False):
...
+        Keyword arguments passed to :func:`~pyvista.principal_axes_vectors`.
+
+    Returns
+    -------
+    numpy.ndarray
+        The 4x4 transformation matrix which aligns the points to the XYZ axes
+        at the origin.
+
+    Examples
+    --------
```
tkoyama010, line 217: This section is informative enough to add a stand-alone example script. It would be better to add a script.

tkoyama010 (changes requested):
Awesome job! I have reviewed it at this time. Please contact us again when ready.

pyvista/core/utilities/points.py
```diff
@@ -179,11 +182,450 @@ def lines_from_points(points, close=False):
...
+        the XYZ axes.
+
+    Returns
+    -------
+    numpy.ndarray
+        A 3x3 array with the principal axes as row vectors or a 4x4
+        transformation matrix if ``as_transform`` is ``True``.
+
+    Examples
+    --------
```
tkoyama010, line 414: Same with https://github.com/pyvista/pyvista/pull/4947#discussion_r1333697020_ .

user27182:
Thanks for reviewing and for the docstring fixes. I will rework the examples. 

I have also found that the API should be reworked because it's sometimes helpful to have both the axes _and_ the transformation returned by `principal_axes_vectors` instead of just one or the other. Also, the inverse transform is helpful to return as well because often the workflow is: transform- > do stuff -> transform back. I will push these changes soon.

user27182:
@tkoyama010 I have added the inverse transform and updated the API (among other changes). I believe this should be the last big change to the code and API. I have not yet reworked the examples, however. Feel free to review current changes if you'd like or wait until the examples are changes.

pyvista/core/utilities/points.py
```diff
@@ -179,11 +182,481 @@ def lines_from_points(points, close=False):
...
+    axis_1_direction = _validate_vector(axis_1_direction, name='axis_1_direction')
+    axis_2_direction = _validate_vector(axis_2_direction, name='axis_2_direction')
+
+    # Compare all direction vectors with each other
+    directions = [axis_0_direction, axis_1_direction, axis_2_direction]
+    for i, vec1 in enumerate(directions):
+        for j, vec2 in enumerate(directions[i + 1 : :]):
+            if vec1 is not None and vec2 is not None:
+                if np.allclose(vec1, vec2):
+                    raise ValueError("Direction vectors must be distinct.")
```
tkoyama010, line 510:
```suggestion
    for i, vec1 in enumerate(directions):
        for j, vec2 in enumerate(directions[i + 1 : :]):
            if vec1 is not None and vec2 is not None and np.allclose(vec1, vec2):
                raise ValueError("Direction vectors must be distinct.")
```

user27182:
@tkoyama010 I'm not quite sure how the integration tests work. Tests with GeoVista are failing, I'm sure why, but perhaps a numpy upgrade can resolve this? see [stackoverflow link](https://stackoverflow.com/questions/66060487/valueerror-numpy-ndarray-size-changed-may-indicate-binary-incompatibility-exp). Is it possible to alter the test config like that?

tkoyama010:
I re-run GitHubActions. Let's see if the error reproduces each time.

tkoyama010:
Ummm. Looks like it will reproduce. @bjlittle Do you have any idea about this?

tkoyama010:
> Edit: geovista [requires numpy>=1.21](https://github.com/bjlittle/geovista/blob/8a9b6429c4ab3480496ac8986f94562033be94b7/requirements/geovista.yml#L16C17-L16C17).
> Some [users reported](https://stackoverflow.com/a/68775561) fixing this issue with numpy ==1.21.1. Is it possible to run the CI test with this version of numpy?

Sure.

user27182:
I did make changes to transform-related code in the recent commits and I see that the GeoVista test fails with  `geovista.bridge import Transform`, so it may be the changes I made which are the cause. I will investigate further tomorrow.

bjlittle:
@tkoyama010 and @user27182 

Interesting 🤔

I've not seen this before, I'll spin up and investigate too 👍

bjlittle:
[numpy 1.26.0](https://github.com/numpy/numpy/releases/tag/v1.26.0) was just release on 16 Sep 2023 ... perhaps this is related ...

bjlittle:
See #4957

user27182 (commented):
Is it possible to modify the pre-commit so that it catches this docstring error before the build fails?

i.e. complain when there is a missing indent following a directive, e.g.:
```
.. directive::
Some text
```
which should be formated as:
```
.. directive::
    Some text
```

tkoyama010:
> Is it possible to modify the pre-commit so that it catches this docstring error before the build fails?

Good point. I looked at `matplotlib` `pre-commit` settings and they seem to be using `check-docstring-first`. I don't know if this will meet your requirements, but we'll give it a try with #4959.

user27182:
> I looked at `matplotlib` `pre-commit` settings and they seem to be using `check-docstring-first`. I don't know if this will meet your requirements, but we'll give it a try with #4959.

Thanks for looking into this. Unfortunately #4959 did not resolve the issue. It is easy to check if this is caught by the pre-commit by simply adding/removing indentation for a directive's text block.

I have opened a new issue with numpydoc https://github.com/numpy/numpydoc/issues/507

user27182:
Most of this PR has been implemented via 
- #6346 
- #6354 
- #6485 

Some items which may be useful but not yet implemented:
- Function pv.random_sample_points() for sampling points
- Fix for https://github.com/pyvista/pyvista/issues/4711
- New standalone example

PR #4688, opened 2023-07-20
Fix VTK 9.3 failures

banesullivan:
The following are a list of failures being captured by our CI that tests against VTK master. Tracking here for the VTK 9.3 release

cc @alesgenova

### Composite core API

Something is going awry with the copying -- I suspect this was broken upstream

- [x] FAILED tests/core/test_composite.py::test_multi_block_copy - assert 139993061330560 == 139993061330944
    - Tracked in https://gitlab.kitware.com/vtk/vtk/-/issues/19143

### Composite plotting

I'm seeing this warning from VTK that might be the cause of many of these:

```
DeprecationWarning('Call to deprecated class vtkCompositePolyDataMapper2. (Please use vtkCompositePolyDataMapper instead) -- Deprecated since version 9.3.0.')
```

- [x] FAILED tests/plotting/test_plotting.py::test_orthographic_slicer - RuntimeError: test_orthographic_slicer Exceeded image regression error of 500.0 with an image error equal to: 3940.07712418308
- [x] FAILED tests/plotting/test_plotting.py::test_plot_composite_poly_complex - Failed: DID NOT WARN. No warnings of type (<class 'numpy.exceptions.ComplexWarning'>,) were emitted.
- [x] FAILED tests/plotting/test_plotting.py::test_plot_composite_bool - RuntimeError: test_plot_composite_bool Exceeded image regression error of 500.0 with an image error equal to: 7256.444444444465
- [x] FAILED tests/plotting/test_plotting.py::test_plot_composite_poly_component_norm - RuntimeError: test_plot_composite_poly_component_norm Exceeded image regression error of 500.0 with an image error equal to: 7393....
- [x] FAILED tests/plotting/test_plotting.py::test_plot_composite_many_options - RuntimeError: test_plot_composite_many_options Exceeded image regression error of 500.0 with an image error equal to: 6737.6326797...
- [x] FAILED tests/plotting/test_plotting.py::test_plot_composite_poly_component_single - RuntimeError: test_plot_composite_poly_component_single Exceeded image regression error of 500.0 with an image error equal to: 472...
- [x] FAILED tests/plotting/test_plotting.py::test_plot_composite_categories - RuntimeError: test_plot_composite_categories Exceeded image regression error of 500.0 with an image error equal to: 3068.726797385632
[...]

PR #4950, opened 2023-09-22
Use ubuntu-latest image in GitHub Actions

tkoyama010:
### Overview

`PyVista` and `vtk` are `Python` tools and are basically unaffected by differences in Ubuntu versions.
This change frees us from future Ubuntu version changes by using ubuntu-latest in GitHub Actions.
_Edit: Changes that will occur by setting the ubuntu version of the current GitHub Actions to the latest is listed below._

### Details

- [x] Do NOT Update test image file for the ubuntu-latest environment.

As shown below, ubuntu-22.04 (ubuntu-latest) is causing some parts of the wireframe to turn black.
| ubuntu 20.04 | ubuntu-latest |
|---|---|
| plot_list | plot_list (1) |


banesullivan:
The rendering differences look like issues with the version of OSMesa. IMO, the new renderings are unexpected and we should not alter our ground-truth images to match. We may need to get to the root of what is causing these rendering differences and report upstream to VTK

tkoyama010:
You are absolutely right about not updating the images. Thank you for setting out the policy to determine the cause.

tkoyama010:
.pre-commit.ci autofix

banesullivan:
@tkoyama010, please refer to #3740 for why `ubuntu-20.04` is used which is due to differences in the versions of Mesa shipped with each version. cc @akaszynski

tests/plotting/test_plotting.py
```diff
@@ -264,6 +264,10 @@ def test_plot_update(sphere):
     pl.close()
 
 
+# https://github.com/pyvista/pyvista/pull/4950
+@pytest.mark.skipif(
+    platform.system() == 'Linux', reason='Does not display correctly within latest ubuntu'
+)
```
banesullivan, line 270: We should avoid these sorts of skips and instead skip via the `verify_image_cache` fixture

banesullivan:
I'm all for using the `ubuntu-latest` image but I want to validate that this isn't significantly slowing down our CI and that we aren't disabling too many tests

tkoyama010:
@banesullivan 100%. Let's consider this a little more carefully.

banesullivan:
Looks like `test_plot_cubemap_alone`. I vaguely recall tracking a failure with cubemaps in VTK with recent versions of Mesa... Perhaps: 

- https://gitlab.kitware.com/vtk/vtk/-/issues/18584
- https://github.com/pyvista/pyvista/pull/4688

banesullivan:
@tkoyama010, would you be okay with closing this PR and tracking this in an issue? I think we should stick with our current CI configurations for the foreseeable future

tkoyama010:
@banesullivan I agree with that. Closing this.

Issue #462, opened 2019-11-21
Standard official VTK Polygonal Data Format


zhaofenqiang:
**Describe the bug, what's wrong, and what you expect:**

I was saving the mesh or surface created by pyvista. I was expecting to obtain the standard official vtk data format as shown [here](https://lorensen.github.io/VTKExamples/site/VTKFileFormats/#dataset-format). But the vtk file I obtained doesn't.

-----

**To Reproduce**

```py
vertices = np.array([[0, 0, 0],
                    [1, 0, 0],
                    [1, 1, 0],
                    [0, 1, 0],
                    [0.5, 0.5, 1]])
faces = np.hstack([[4, 0, 1, 2, 3],
                   [3, 0, 1, 4],
                   [3, 1, 2, 4]])  # one square and two triangles
surf = pyvista.PolyData(vertices, faces)
surf.save('/home/fenqiang/test.vtk', binary=False)
 ```

The vtk file I obtained is:
```
# vtk DataFile Version 4.2
vtk output
ASCII
DATASET POLYDATA
POINTS 5 double
0 0 0 1 0 0 1 1 0 
0 1 0 0.5 0.5 1 
POLYGONS 3 13
4 0 1 2 3 
3 0 1 4 
3 1 2 4 
```
But according to the official format, I was expecting to obtain this:
```
# vtk DataFile Version 4.2
vtk output
ASCII
DATASET POLYDATA
POINTS 5 double
0 0 0 
1 0 0
1 1 0 
0 1 0
0.5 0.5 1 
POLYGONS 3 13
4 0 1 2 3 
3 0 1 4 
3 1 2 4 
```

-----

**System Information:**
```
--------------------------------------------------------------------------------
  Date: Thu Nov 21 11:41:50 2019 EST

             Linux : OS
                12 : CPU(s)
            x86_64 : Machine
             64bit : Architecture
           31.1 GB : RAM
           Jupyter : Environment

  Python 3.6.8 |Anaconda custom (64-bit)| (default, Dec 30 2018, 01:22:34)
  [GCC 7.3.0]

            0.22.4 : pyvista
             8.2.0 : vtk
            1.16.2 : numpy
             2.4.1 : imageio
             1.4.3 : appdirs
             0.4.3 : scooby
             3.1.1 : matplotlib
             5.9.2 : PyQt5
             7.2.0 : IPython
             7.4.2 : ipywidgets

[...]

PR #4957, opened 2023-09-26
Update geovista integration test

bjlittle:
### Overview

This pull-request updates the `geovista` integration test.

On the `main` branch of `geovista` we now have a minimum pin for `cartopy>=0.22` (see [#462](https://github.com/bjlittle/geovista/pull/462)).

This means that there is no longer a need to install GEOS for the integration test (`libgeos-dev`) as geometry handling is now performed by `shapely` within `cartopy>=0.22` instead.


.github/workflows/integration-tests.yml
```diff
@@ -93,11 +93,8 @@ jobs:
       - run: git clone https://github.com/bjlittle/geovista.git --single-branch
       - name: Install PyVista
         run: pip install -ve . # pyvista
-      - run: sudo apt-get install libgeos-dev
       - name: Install GeoVista
-        run: pip install -ve . # geovista
-        working-directory: geovista
-      - run: pip install -r requirements/pypi-optional-test.txt
+        run: pip install -ve .[test] # geovista
```
bjlittle, line 97: Seems to make sense (to me) to install all the PyPI packages for `geovista` at once ... or is there a good reason why this was previously performed in two steps ?

bjlittle:
The `geovista` integration test is failing, akin to #4947.

I have been unable to replicate this locally when following the same install pattern.

image

Investigating further ...

bjlittle:
Okay, I can recreate this error, previously I was running the `geovista` tests with simply `pytest` rather than `pytest -E error`

bjlittle:
The issue is that there seems to be now a new `RuntimeWarning` generated by `numpy=1.26.0`, and we've requested that `pytest` escalates said warning to an error, hence the failure.

What's the wisdom of the crowd here?

Is it acceptable to ignore this specific `RuntimeWarning` from `numpy` ?

.github/workflows/integration-tests.yml
```diff
@@ -93,16 +93,13 @@ jobs:
...
         working-directory: geovista
       - name: Download cartopy assets
         run: |
           wget --quiet ${CARTOPY_FEATURE}
           mkdir -p ${CARTOPY_SHARE_DIR}
           python cartopy_feature_download.py physical --output ${CARTOPY_SHARE_DIR} --no-warn
-      - run: pytest -W error
+      - run: pytest -W error -W "ignore:numpy.ndarray size changed:RuntimeWarning"
```
bjlittle, line 104: @pyvista/developers Is this an acceptable compromise ?

bjlittle: I'll investigate whether I require to recreate all of the `netCDF4` assets of `geovista` using the latest version of `numpy`. I'm not sure what would raise the ABI warning otherwise ...

MatthewFlamm: I'm good with this.

tkoyama010 (approved):
Thanks @bjlittle . It really help us!

PR #4958, opened 2023-09-26
Perform Render after Zoom event in default interactor key events

ChristosT:
### Overview
Based on the [documentation](https://docs.pyvista.org/version/stable/api/plotting/plotting.html) the up/down keys should zoom in/out the view however this does not work as expected. You can verify the issue by pressing "up" for any of the examples.

### Details
Previously the default bindings of the up/down keys wouldn't render the scene and thus the zoomed scene is not visible until another event or interaction causes a render.

The missing part is that the keypress should also trigger a render and this is what this pull request does.

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -2058,11 +2058,20 @@ class BasePlotter(PickingHelper, WidgetHelper):
         b_left_down_callback = lambda: self.iren.add_observer(
             'LeftButtonPressEvent', self.left_button_down
         )
+
+        def zoom_in():  # numpydoc ignore=GL08
+            self.camera.Zoom(1.05)
+            self.render()
+
+        def zoom_out():  # numpydoc ignore=GL08
+            self.camera.Zoom(0.95)
+            self.render()
+
         self.add_key_event('b', b_left_down_callback)
         self.add_key_event('v', lambda: self.isometric_view_interactive())
         self.add_key_event('C', lambda: self.enable_cell_picking())
-        self.add_key_event('Up', lambda: self.camera.Zoom(1.05))
-        self.add_key_event('Down', lambda: self.camera.Zoom(0.95))
+        self.add_key_event('Up', lambda: zoom_in())
+        self.add_key_event('Down', lambda: zoom_out())
         self.add_key_event('plus', lambda: self.increment_point_size_and_line_width(1))
         self.add_key_event('minus', lambda: self.increment_point_size_and_line_width(-1))
 
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -2059,19 +2059,27 @@ class BasePlotter(PickingHelper, WidgetHelper):
             'LeftButtonPressEvent', self.left_button_down
         )
 
-        def zoom_in():  # numpydoc ignore=GL08
-            self.camera.zoom(1.05)
+        def zoom(self, value):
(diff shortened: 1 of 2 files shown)


pyvista/plotting/plotter.py
```diff
@@ -2058,11 +2058,20 @@ def reset_key_events(self):
         b_left_down_callback = lambda: self.iren.add_observer(
             'LeftButtonPressEvent', self.left_button_down
         )
+
+        def zoom_in():  # numpydoc ignore=GL08
+            self.camera.Zoom(1.05)
```
tkoyama010, line 2063: We wrapped `Zoom` method by `zoom`. Let's use it!
```suggestion
            self.camera.zoom(1.05)
```
https://github.com/pyvista/pyvista/blob/a0691dd6687774247ff6e806e801a48c37085023/pyvista/plotting/camera.py#L410-L459

pyvista/plotting/plotter.py
```diff
@@ -2058,11 +2058,20 @@ def reset_key_events(self):
...
+            self.render()
+
+        def zoom_out():  # numpydoc ignore=GL08
+            self.camera.Zoom(0.95)
```
tkoyama010, line 2067:
```suggestion
            self.camera.zoom(0.95)
```

pyvista/plotting/plotter.py
```diff
@@ -2058,11 +2058,20 @@ def reset_key_events(self):
...
             'LeftButtonPressEvent', self.left_button_down
         )
+
+        def zoom_in():  # numpydoc ignore=GL08
+            self.camera.zoom(1.05)
+            self.render()
+
+        def zoom_out():  # numpydoc ignore=GL08
+            self.camera.zoom(0.95)
+            self.render()
```
tkoyama010, line 2068: Let's combine functions into one!
```suggestion
        def zoom(self, value):
        """Set the zoom of the camera and render.

        Parameters
        ----------
        value : float or str
            Zoom of the camera. If a float, must be greater than 0. Otherwise,
            if a string, must be ``"tight"``. If tight, the plot will be zoomed
            such that the actors fill the entire viewport.
            self.camera.zoom(value)
            self.render()

        """
        self.camera.zoom(value)
        self.render()
```

pyvista/plotting/plotter.py
```diff
@@ -2058,11 +2058,20 @@ def reset_key_events(self):
...
+
         self.add_key_event('b', b_left_down_callback)
         self.add_key_event('v', lambda: self.isometric_view_interactive())
         self.add_key_event('C', lambda: self.enable_cell_picking())
-        self.add_key_event('Up', lambda: self.camera.Zoom(1.05))
-        self.add_key_event('Down', lambda: self.camera.Zoom(0.95))
+        self.add_key_event('Up', lambda: zoom_in())
+        self.add_key_event('Down', lambda: zoom_out())
```
tkoyama010, line 2074:
```suggestion
        self.add_key_event('Up', lambda: self.zoom(1.05))
        self.add_key_event('Down', lambda: self.zoom(0.95))
```

tkoyama010:
pre-commit.ci autofix

tkoyama010 (dismissed):
LGTM. Good catch.

tkoyama010 (approved):

tkoyama010 (approved):

ChristosT:
Thanks for the quick review and all the improvements !

PR #5013, opened 2023-10-04
BUG: Give scraper a stable __repr__

larsoner:
In MNE-Python I was getting doc rebuilds no matter what, turns out it's due to the PyVista scraper not having a stable `__repr__` build to build, e.g.  [here](https://sphinx-gallery.github.io/stable/advanced.html#example-1-a-matplotlib-style-scraper).

### DIFF
--- a/pyvista/plotting/utilities/sphinx_gallery.py
+++ b/pyvista/plotting/utilities/sphinx_gallery.py
@@ -66,6 +66,10 @@ class Scraper:
 
     """
 
+    # Give it a stable __repr__
+    def __repr__(self):
+        return f"<{self.__class__.__name__}"
+
     def __call__(self, block, block_vars, gallery_conf):
         """Save the figures generated after running example code.
 


tkoyama010 (dismissed):
LGTM. I am not familiar with `scraper `, but it is an interesting feature.

tkoyama010 (approved):

pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -66,6 +66,10 @@ class Scraper:
 
     """
 
+    def __repr__(self):  # pragma: no cover
+        """Return a stable representation of the class."""
+        return f"<{self.__class__.__name__}"
```
adeak, line 71: Is this missing an ending `>` on purpose? And should this have an explicit `"class"` in its name? I figured this meant to look something like
```py
>>> Foo()
<__main__.Foo object at 0x7f7faf006cd0>
```
except without the "object at ..." part.

adeak: Specifically I'd have expected
```suggestion
        return f"<{self.__class__.__name__}>"
```

tkoyama010: @larsoner Could you have your answer about this?
> Is this missing an ending > on purpose?

larsoner: I just forgot the `>`. Adding `object` is fine too if you want

adeak: OK, thanks for confirming. I'll open a PR later.

PR #5022, opened 2023-10-05
Fix `Scraper` object `repr` and add a test of it

adeak:
Update stable `Scraper` `repr()`, see [discussion](https://github.com/pyvista/pyvista/pull/5013#discussion_r1346535158).

I've also added a test to cover the repr. It's slightly wasteful because it runs for both parameterisations, but it would seem more wasteful to define a dedicated test function just for this tiny thing.


pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -66,9 +66,9 @@ class Scraper:
 
     """
 
-    def __repr__(self):  # pragma: no cover
-        """Return a stable representation of the class."""
-        return f"<{self.__class__.__name__}"
+    def __repr__(self):
+        """Return a stable representation of the class instance."""
+        return f"<{self.__class__.__name__} object>"
```
adeak, line 71: Actually, I'm thinking this could be
```suggestion
        return f"<{type(self).__name__} object>"
```
Thoughts?

tkoyama010: I guess we can and more safe.

tkoyama010 (approved):
LGTM. Sorry for the late reply.

PR #5029, opened 2023-10-06
Unify and extend `AxesActor` and `Property`

### LINKED ISSUE #4980: Unify and extend `AxesActor` API

This is a proposal to unify and extend the API for `AxesActor` and related functions. It is related to #4947. I am documenting the proposal here and would like some feedback on the API before implementing any changes.

### Unify the API
The following axes-related classes/methods:
- [AxesActor](https://docs.pyvista.org/version/stable/api/plotting/_autosummary/pyvista.AxesActor.html)
- [pyvista.create_axes_marker](https://docs.pyvista.org/version/stable/api/plotting/_autosummary/pyvista.create_axes_marker.html#pyvista.create_axes_marker)
- [Plotter.add_axes](https://docs.pyvista.org/version/stable/api/plotting/_autosummary/pyvista.Plotter.add_axes.html)
- [Plotter.add_axes_at_origin](https://docs.pyvista.org/version/stable/api/plotting/_autosummary/pyvista.Plotter.add_axes_at_origin.html)

all have the parameters below in common. The inconsistencies are highlighted. The red cells indicate a missing equivalent parameter, and the blue cells indicate the same parameter but with a different name.

image

Therefore, planned changed would include:
- `AxesActor`: add `line_width`, `label_color`, `x_color`, `y_color`, `z_color` and `labels_off` parameters
- `AxesActor`: rename `x_axis_label` -> `xlabel`, `y_axis_label` -> `ylabel`, `z_axis_label` -> `zlabel`
- `add_axes_at_origin`: add `label_color`
- `add_axes`: rename `color` -> `label_color` 

In addition, it seems like `create_axes_marker` is effectively a constructor for `AxesActor`, but it creates the actor with `_vtk.vtkAxesActor()` instead of `pyvista.AxesActor`. Therefore, I propose to effectively merge the constructor code from `create_axes_marker` to `AxesActor.__init__` so that all `create_axes_marker` does is call `AxesActor(*args, **kwargs)`.

### Extend the API
The main motivation for this change stems from the need to position, rotate, and scale an `AxesActor` in space (not as a widget) so that, for example, it's possible to plot the `AxesActor` centred at the centroid of a mesh/point cloud and orient it to align with its `principal_axes` from #4947. Therefore, proposed additions include:

- Expose the `user_matrix` parameter from `Prop3D`, similar to #4931, to allow using the `principal_axes_transform` from #4947 directly to set the axes' position and orientation
- Expose `position`, `scale`, and `orientation` parameters from `Prop3D` as well
    - For `scale`, a workaround is needed since `vtkAxesActor().SetScale()` has no effect
- Create a higher level API for `AxesActor` by abstracting `tip` and `shaft` related properties, e.g.:
    - add `tip_radius` as an alias/replacement for `cone_radius` or `sphere_radius`
    - add `shaft_width` as an alias/replacement for `line_width`
    - add `shaft_radius` as an alias/replacement for `cylinder_radius`
- Change the default shaft type from `LINE` to `CYLINDER`, since a cylinder scales much better in space (`line_width` units are in pixels, whereas `cylinder` units are physical)
- Add new method `Plotter.add_axes_actor` (or maybe call it `Plotter.add_axes_in_space`) to be able to add axes with any postion, scale, orientation, basically a more generic version of `Plotter.add_axes_to_origin`

### Links to source code.

_No response_

### Pseudocode or Screenshots

_No response_

### LINKED ISSUE #5019: `vtkAxesActor.SetScale()` does not scale and `Plotter.show_bounds()` does not scale with input

This bug is related to #4980.

`vtkAxesActor.SetScale()` (and by extension `pv.AxesActor.scale`) does not have any effect on the actor's scale. As a workaround it's possible to use a transform to apply scaling instead (see example bellow).

I have two issues with this:

1. I have not been able to find a good way to test for this bug programmatically since the transformation matrices, bounds, etc. between the two actors are identical, but the plots differ. So, how can we test that the scaling actually happened? Is image regression testing needed? Which leads to the second issue...
2. Plotting with `add_actor()` shows that the correct scaling is applied to the `UVW` axes actor, but this creates another issue where the scaling isn't picked up by `Plotter.show_bounds()`. This could be due to the bounds being based on world coordinates (i.e. the bounds are defined without the scaling),

```python
import pyvista as pv
import vtk

actor1 = vtk.vtkAxesActor()
actor1.SetScale(2, 2, 2)
print(actor1.GetMatrix())
#  Elements:
#    2 0 0 0
#    0 2 0 0
#    0 0 2 0
#    0 0 0 1
print(actor1.GetBounds()) 
# (-1.0, 1.0, -1.0, 1.0, -1.0, 1.0)


actor2 = vtk.vtkAxesActor()
scale = vtk.vtkTransform()
scale.Scale(2, 2, 2)
actor2.SetUserTransform(scale)
actor2.SetXAxisLabelText('U')
actor2.SetYAxisLabelText('V')
actor2.SetZAxisLabelText('W')
print(actor2.GetMatrix())
#  Elements:
#    2 0 0 0
#    0 2 0 0
#    0 0 2 0
#    0 0 0 1
print(actor2.GetBounds())
 # (-1.0, 1.0, -1.0, 1.0, -1.0, 1.0)


p = pv.Plotter()
p.add_actor(actor1)
p.add_actor(actor2)
#p.show_bounds()
p.show()
```
image

With `show_bounds` uncommented, we see that the bounds do not encompass the scaled `UVW` axes.
image

### System Information

```shell
OS : Windows
            CPU(s) : 12
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.8 GiB
       Environment : IPython
        GPU Vendor : Intel
  MathText Support : False
  bit (AMD64)]
           pyvista : 0.43.dev0
               vtk : 9.2.6
             numpy : 1.25.1
        matplotlib : 3.7.2
            scooby : 0.7.2
             pooch : v1.7.0
            pillow : 10.0.0
           imageio : 2.31.1
           IPython : 8.14.0
          colorcet : 3.0.1
           cmocean : v3.0.3
        ipywidgets : 7.7.5
             scipy : 1.11.1
              tqdm : 4.65.0
            meshio : 5.3.4
        jupyterlab : 4.0.3
    pytest_pyvista : 0.1.8
             trame : 2.5.2
      trame_client : 2.9.4
      trame_server : 2.11.7
         trame_vtk : 2.5.8
     trame_vuetify : 2.3.1
      nest_asyncio : 1.5.6
```

### Screenshots

_No response_

user27182:
### Overview
This PR resolves #4980 and resolves #5019. It also implements the `AxesActor` features and changes needed to close #4947.

### Details

- `ActorProperties`
    - Add `ambient` property
- `Prop3D`
    - Add `origin` property
- `AxesActor`
    - Update `Enum` -> `AnnotatedEnum` to enable string access of `ShaftType` and `TipType` props
    - Change default shaft type from `line` to `cylinder`
    - Add constructor to init any/all properties
    - Inherit from `Prop3D` to expose `orientation`, `scale`, `rotate_xyz`, `user_matrix`,`origin`, `position` properties
    - Rename `xlabel` -> `x_label` and add `xlabel` as alias; deprecate `x_axis_label`
    - Rename `ylabel` -> `y_label` and add `ylabel` as alias; deprecate `y_axis_label`
    - Rename `zlabel` -> `z_label` and add `zlabel` as alias; deprecate `z_axis_label`
    - Remove untested and unusable recursive properties setters, e.g. `x_axis_tip_properties.setter` (there is no `SetProperty()` method available in the underlying class)
    - Add `label_color` property
    - Add `x_color`, `y_color`, `z_color` properties
    - Add `label_size` property
    - Create shaft / tip property abstraction
        - Deprecate `cone_radius`, replace with `tip_radius`
        - Deprecate `sphere_radius`, replace with `tip_radius`
        - Deprecate `cone_resolution`, replace with `tip_resolution`
        - Deprecate `sphere_resolution`, replace with `tip_resolution`
        - Deprecate `cylinder_resolution`, replace with `shaft_resolution`
        - Deprecate `cylinder_radius`, replace with `shaft_radius`
        - Deprecate `line_width`, replace with `shaft_width`
    - Create `properties` init param to handle setting `ActorProperty` kwargs 
        - Deprecate `ambient` param, replace with `properties=dict(ambient=value)`
    - Add workaround for vtk bug where orientation, position, origin, scale, and bounds have no effect or are incorrect (i.e. resolves #5019)
        - The workaround stores the transformations with  `Get/SetUserMatrix`, and implements a mechanism with overrides to ensure the matrix is always up to date
        - Add `_enable_orientation_workaround` property (``True`` by default) to allow for disabling the workaround if needed
- `create_axes_marker`
    -  Deprecate method, merge constructor into `AxesActor()`
- `create_axes_orientation_box`
    - Remove unused `line_width` parameter 
- `add_axes`
    - Remove explicit use of `line_width` from method (parameter is not shared by `create_axes_marker` and `create_axes_orientation_box `)
- `add_axes_at_origin`
    -  Deprecate method, rename to `add_axes_marker`

### TODO

- [x] Resolve failed plotting tests related to API changes (e.g. from new deprecation warnings, failed regression tests due to slight change in how the axes appear)
- [x] Resolve failed docstring tests related to API changes
- [x] Add missing docstrings ~and examples~ for new parameters
- [x] Add plot image testing

For future PRs:
- Fix bug #5051
- Update `Axes` 
- Add `symmetric` param which will add a second `AxesActors` to a scene, with the second actor pointing in -x,-y,-z directions

### Preview
With this PR, we can run the following code to scale, orient, position, etc. axes in space. The bounds from the grid are also correctly sized.
```python
import pyvista as pv
axes_actor1 = pv.AxesActor()
axes_actor2 = pv.AxesActor(
    label_color = 'pink',
    scale=2,
    position=(0.5,-0.5,0.5),
    orientation=(15,30,50),
    x_label='i',
    y_label='j',
    z_label='k'
)

p = pv.Plotter()
p.add_actor(axes_actor1)
p.add_actor(axes_actor2)
p.show_grid()
p.show()
```
Which will result in:
image


tkoyama010:
pre-commit.ci autofix

tkoyama010:
I had no idea there was such a great solution to AxesActor's scale problem. Thank you. This is too good to be.

user27182:
For the docstrings, there will be a need to do a fair bit of copy/pasting, e.g. description of each class property -> roughly copied to AxesActor constructor parameters -> roughly copied to `create_axes_actor` parameters and `add_axes_at_origin` parameters... etc. Is there a way to keep all the type hinting and documentation without doing all the copying?

user27182:
Some notes about the `AxesActor()` workaround:

The `vtkAxesActor`'s complete transformation (returned by `GetMatrix`) is computed with [`ComputeMatrix`](https://github.com/Kitware/VTK/blob/2e9a8beb5cf79c3364bc351f30e133a9a7b04b5d/Rendering/Core/vtkProp3D.cxx#L482) which computes the actor's "complete_transform" as:
- "implicit_transform" -> "user_transform" -> "coordinate_system_transform"

where "implicit_transform" is constructed from the parameters `GetOrientation, GetScale, GetOrigin, GetPosition`, and "user_transform" is accessed with `Set/GetUserMatrix` or `Set/GetUserTransform`.

Calling `GetMatrix` will return the correct "complete_transform" (including a correct "implicit_transform"),  but visually this matrix has no effect when plotted in a scene (see #5019). However, the transform in `Set/GetUserTransform` *does* have an effect on the plotted actor. As such, the implemented workaround effectively re-computes the "implicit_transform" manually, and stores this matrix as part of the "user_transform". 

Effectively, the workaround stores both the "user_transform" *and* the "implicit_transform" parts inside of a single attribute using `Set/GetUserTransform`. This is done without overloading any of the `vtkAxesActor` methods (e.g. `Set/GetUserMatrix` but instead it's done by overloading `pyvista.Prop3D` methods (e.g. `user_matrix`) and storing the user's actual matrix in an internal variable `_user_matrix`.

The workaround now produces correct plots, but it has a few problems:
1. `GetMatrix` now returns an incorrect matrix since it will calculate its own "implicit_transform" and concatenate it with our manually computed "implicit_transform", effectively applying it twice. However, it means that `Set/GetScale`, `Set/GetOrientation`, etc. all continue to work as expected.
2. `GetUserMatrix` now returns a misleading matrix in that this matrix contains both the actual "user_transform" and the re-computed "implicit_transform`, i.e. `user_matrix != GetUserMatrix`
3. This workaround alone does not alter the actor's (incorrect) bounds, however. Therefore, `AxesActor` overrides `bounds` to manually compute the correct bounds. However, this means that downstream methods will need to call `bounds` and not `GetBounds` to get the correct values. E.g. `show_bounds` had to be modified to check for  the `bounds` explicitly, and I found a similar workaround is needed for `camera.tight` which currently calls `renderer.ComputeVisiblePropBounds()`.

To address (1) and (2), we could add an internal `bool` parameter to turn on/off the workaround, so that it's only "on" as required, e.g. "on" for `add_axes_to_origin` (axes in space) vs "off" for `add_axes` (axes as a widget).

I'm thinking to address (3) perhaps `AxesActor` should override `GetBounds` directly instead of `bounds`.

tkoyama010:
> For the docstrings, there will be a need to do a fair bit of copy/pasting, e.g. description of each class property -> roughly copied to AxesActor constructor parameters -> roughly copied to `create_axes_actor` parameters and `add_axes_at_origin` parameters... etc. Is there a way to keep all the type hinting and documentation without doing all the copying?

There is not a particularly good way to do this at the moment now in my mind. 
We can put it on hold in this PR using numpydoc ignore. We can make follow-up after this PR.
Let's concentrate on fixing the bug first.

darikg:
This is fantastic. One question: is there really a need for separate methods `add_axes`, `add_axes_at_origin`, `add_axes_in_space`, etc? It seems to me there could be the sole method `add_axes` with a default position of (0, 0, 0), obviating the need for `add_axes_at_origin`, etc.

user27182:
> is there really a need for separate methods `add_axes`, `add_axes_at_origin`, `add_axes_in_space`, etc?

Great point. I believe there are two main use cases:

1. Axes as a widget (e.g. bottom left-hand corner of plot). This is what `add_axes` does.
2. Axes rendered in the scene. This is what `add_axes_at_origin` does, and what `add_axes_in_space` would do.

I agree here that `add_axes_at_origin` is redundant and should be deprecated. But how to handle the two use cases?
Perhaps this naming is more appropriate:

- `add_axes` -> `add_axes_widget`
- `add_axes_at_origin` -> `add_axes_actor`

Or, more in-line with your comment:
- `add_axes` -> `add_axes(as_widget=True)`
- `add_axes_at_origin` -> `add_axes(as_widget=False)`

Also note that the `Prop3D` parameters for `origin`, `user_transform`, etc. would have no effect (or at least shouldn't) on the widget case. Therefore, separating the two use cases would allow to more clearly separate the applicable `kwargs`, i.e. the widget case would have the `Prop3D` parameters hidden / omitted from the docs

darikg:
Ah, I see now. I know I've been confused by the distinction in the past. My vote's for

> - add_axes -> add_axes_widget
> - add_axes_at_origin -> add_axes_actor

for maximal clarity.

AUTHORS.rst
```diff
@@ -26,3 +26,4 @@ considered "The PyVista Developers":
 - Matthew Flamm, (`@MatthewFlamm <https://github.com/MatthewFlamm>`_)
 - Darik Gamble, (`@darikg <https://github.com/darikg>`_)
 - Bram De Cooman, (`@dcbr <https://github.com/dcbr>`_)
+- `@user27182 <https://github.com/user27182>`_
```
tkoyama010, line 29: @pyvista/maintainers  I have determined that this change is copyrighted and have added @user27182 to `AUTHORS.rst`.
@user27182 This change is my suggestion, so please make it as you wish.

adeak: @tkoyama010 what do you mean by "copyrighted"?

tkoyama010: @adeak Thanks for comment. Meaning they is into this category :)
https://github.com/pyvista/pyvista/blob/75febafe79be9482accc2af931f1978fd4bec7ee/AUTHORS.rst?plain=1#L10-L12

MatthewFlamm: I agree, and I'd like to further emphasize that the change here isn't copyrighted, which might imply that there are license issues.  It is that the cumulative contributions that rise to the level of being added as a mentioned "PyVista Developer".  I'm also +1 on the suggestion if accepted.

user27182: Thanks! I enjoy contributing to PyVista because of the help and support from the community (and the CI testing is fantastic). I'll leave the changes to `AUTHORS.rst` as-is for now.

user27182:
I'm thinking that changing `add_axes` -> `add_axes_widget` is probably better saved for another PR as it will have a big impact, and may need more discussion.

Also, perhaps `add_axes_at_origin` -> `add_axes_actor` is still not clear enough because `add_axes_widget` is also adding an actor to the scene (indeed it returns an `AxesActor`).

Therefore, my current vote is for: `add_axes_at_origin` -> `add_axes_marker`. This avoids the ambiguity mentioned above and also maintains consistency with `create_axes_marker`.

tkoyama010:
Excellent work! Please wait until we complete our review.

pyvista/plotting/axes_actor.py
```diff
@@ -388,111 +745,391 @@ def tip_type(self) -> TipType:  # numpydoc ignore=RT01
...
 
+    @z_label.setter
+    def z_label(self, label: str):  # numpydoc ignore=GL08
+        self.SetZAxisLabelText(label)
+
+    @property
+    def z_axis_label(self) -> str:  # numpydoc ignore=RT01
+        """Text label for the z-axis.
+
+        This parameter is deprecated. Use `z_label` instead.
```
tkoyama010, line 861: For deprecated parameters and classes, please specify your removal policy as follows This will make it easier for us in the future to determine whether it is acceptable to remove (see [Deprecating Features or other Backwards-Breaking Changes](https://docs.pyvista.org/version/dev//extras/developer_notes.html#deprecating-features-or-other-backwards-breaking-changes)).
```suggestion
        .. deprecated:: 0.43.0
           This parameter is deprecated and has been completely removed in version 0.46.0.
           Use `z_label` instead.
```

tests/plotting/test_axes.py
```diff
@@ -127,19 +177,319 @@ def test_axes_actor_shaft_type(axes_actor):
...
+        tip_length=0.1,
+        ambient=0.9,
+        label_size=(0.25, 0.1),
+    )
+    old_args_modified = old_args.copy()
+    with pytest.warns(
+        PyVistaDeprecationWarning,
+        match="Use of `cone_radius` is deprecated. Use `tip_radius` instead.",
+    ):
+        axes_actor = AxesActor(**old_args_modified)
```
tkoyama010, line 449: Lately, we have been adding error codes that work with our upgrades. This will prevent you from forgetting to remove deprecations on version upgrades.
```suggestion
        axes_actor = AxesActor(**old_args_modified)
         if pv._version.version_info >= (0, 46):
            raise RuntimeError('Remove this deprecated class')
```

tkoyama010 (changes requested):

tests/plotting/test_axes.py
tkoyama010: See #5120 .

user27182: This is a good policy, thanks for highlighting it.

This should be resolved by [473fca3](
https://github.com/pyvista/pyvista/pull/5029/commits/473fca38bbe6a029037c20cc383f7cbef487380f)

tkoyama010: Thanks!

pyvista/plotting/axes_actor.py
user27182: This should also be resolved by [473fca3](https://github.com/pyvista/pyvista/pull/5029/commits/473fca38bbe6a029037c20cc383f7cbef487380f)

user27182:
@tkoyama010 Is it possible for contributors to re-run individual failed tests? Or does that require elevated permissions. 

I've noticed this [plotting error](https://github.com/pyvista/pyvista/actions/runs/6656017640/job/18087625151?pr=5029) from the most recent commit 389badf has happened a few times before:
```
ERROR tests/plotting/test_widgets.py::test_affine_widget - AssertionError: Not all objects GCed:
```
But it doesn't seem repeatable since, as you saw, running the test a second time did not raise the same error.

The current [software report error](https://github.com/pyvista/pyvista/actions/runs/6656017640/job/18101866622?pr=5029) also seem to happens somewhat frequently:

```
xvfb-run: error: Xvfb failed to start
```

In fact, this is the [same error](https://github.com/pyvista/pyvista/actions/runs/6657229743/job/18091493402?pr=5100) that caused commit 5162436 from #5100 to also fail its CI testing.

Not sure if these errors should be investigated further or if simply re-running the tests is the solution.

tkoyama010:
@user27182 As you say, I have re-run this in various PRs and it is very annoying :(
I have looked into xvfb, but the software itself is unstable, so investigating could be a waste of time. It would be nice if we could do CI testing without having to run xvfb. But that depends on the vtk project.

MatthewFlamm:
I'm wondering if we need to use `xvfb-run -a`. This is how I would run it locally.  I will create a PR for it.  Another option is to use OSMesa wheel for testing.  One problem is that only recent versions have official wheels.

The affine_widget one needs to be investigated.  It happens too frequently too ignore, and creates a lot of churn for everyone.

user27182:
> The affine_widget one needs to be investigated. It happens too frequently too ignore, and creates a lot of churn for everyone.

To clarify, this is an error that happens with other PRs too, correct? I am asking because I was worried that the `test_affine_widget` error may have been caused/introduced by this PR specifically, since changes are made to `AxesActor` and `user_matrix`, and the `test_affine_widget` uses these directly. However, if this error is common and occurs elsewhere, then hopefully that means this PR isn't the cause.

MatthewFlamm:
> > The affine_widget one needs to be investigated. It happens too frequently too ignore, and creates a lot of churn for everyone.
> 
> To clarify, this is an error that happens with other PRs too, correct? I am asking because I was worried that the `test_affine_widget` error may have been caused/introduced by this PR specifically, since changes are made to `AxesActor` and `user_matrix`, and the `test_affine_widget` uses these directly. However, if this error is common and occurs elsewhere, then hopefully that means this PR isn't the cause.

Here is an example of a failure for test_affine_widget.  I haven't checked that it is an exact same error that might have been seen here.  It isn't showing up right now on this PR.  It also tends to happen on that very specific python version/vtk version, so I suspect there is some vtk specific interaction going on as well.

https://github.com/pyvista/pyvista/actions/runs/6665917251/job/18116465247?pr=5133

pyvista/plotting/prop3d.py
```diff
@@ -213,6 +223,19 @@ def orientation(self) -> tuple:  # numpydoc ignore=RT01
     def orientation(self, value: tuple):  # numpydoc ignore=GL08
```
banesullivan, line 223: Noticed this... should we change this while in here?

```suggestion
    def orientation(self, value: Vector):  # numpydoc ignore=GL08
```

pyvista/plotting/prop3d.py
```diff
@@ -213,6 +223,19 @@ def orientation(self) -> tuple:  # numpydoc ignore=RT01
...
+    def origin(self) -> tuple:  # numpydoc ignore=RT01
+        """Return or set the entity origin.
+
+        This is the point about which all rotations take place.
+
+        """
+        return self.GetOrigin()
+
+    @origin.setter
+    def origin(self, value: tuple):  # numpydoc ignore=GL08
```
banesullivan, line 236:
```suggestion
    def origin(self, value: Vector):  # numpydoc ignore=GL08
```

pyvista/plotting/renderer.py
```diff
@@ -872,48 +873,89 @@ def add_actor(
...
-    def add_axes_at_origin(
+    def add_axes_at_origin(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+        """Add axes actor at the origin.
+
+        This method is deprecated. Use `add_axes_marker` instead.
+        """
+        warnings.warn(
+            "Use of `add_axes_at_origin` is deprecated. Use `add_axes_marker` instead.",
+            PyVistaDeprecationWarning,
+        )
```
banesullivan, line 884: Please add a comment in the form of: `# deprecated 0.43.0, convert to error in 0.45.0, remove 0.46.0`

This will help us grep for and follow through with deprecations during releases

See https://github.com/pyvista/pyvista/blob/main/CONTRIBUTING.rst#deprecating-features-or-other-backwards-breaking-changes

pyvista/plotting/renderer.py
```diff
@@ -923,19 +965,25 @@ def add_axes_at_origin(
...
+            x_label=x_label,
+            y_label=y_label,
+            z_label=z_label,
+            labels_off=labels_off,
             x_color=x_color,
             y_color=y_color,
             z_color=z_color,
-            xlabel=xlabel,
-            ylabel=ylabel,
-            zlabel=zlabel,
```
banesullivan, line 937: Please make sure that you properly deprecate the usage of these keyword arguments. You'll need to capture additional `kwargs`, check for the presence of the old name and issue a deprecation warning if present. 

We have implemented a helper for this too:

https://github.com/pyvista/pyvista/blob/76aa761dac8dddf9995a49872399e4f9a406ecf8/pyvista/core/utilities/misc.py#L12

There are a few examples to model after. One example being:

https://github.com/pyvista/pyvista/blob/76aa761dac8dddf9995a49872399e4f9a406ecf8/pyvista/core/grid.py#L546-L552

pyvista/plotting/renderer.py
```diff
@@ -1018,7 +1066,6 @@ def add_orientation_widget(
     def add_axes(
         self,
         interactive=None,
-        line_width=2,
```
banesullivan, line 1021: Similarly, please make sure you gracefully handle the deprecation of this keyword argument

pyvista/plotting/tools.py
```diff
@@ -1,24 +1,35 @@
...
+from ._vtk import (
+    VTK_ARIAL,
+    VTK_COURIER,
+    VTK_TIMES,
+    vtkAnnotatedCubeActor,
+    vtkMathTextFreeTypeTextRenderer,
+    vtkPolyDataMapper,
+    vtkPropAssembly,
+    vtkRenderWindow,
+)
```
banesullivan, line 23: nit: I have a significant preference for keeping all VTK imports in the `_vtk` module and accessing them directly from the `_vtk` namespace in your module

pyvista/plotting/tools.py
```diff
@@ -102,163 +113,13 @@ def system_supports_plotting():
...
-        label_actor.SetWidth(label_size[0])
-        label_actor.SetHeight(label_size[1])
-
-    _update_axes_label_color(axes_actor, label_color)
-
-    return axes_actor
+@wraps(AxesActor.__init__)
+def create_axes_marker(*args, **kwargs):  # numpydoc ignore=RT01,PR01
+    """Wrap ``AxesActor`` constructor."""
+    return AxesActor(*args, **kwargs)
```
banesullivan, line 119: I think this warrants a deprecation

pyvista/plotting/tools.py
```diff
@@ -102,163 +113,13 @@ def system_supports_plotting():
...
 
 
 def create_axes_orientation_box(
-    line_width=1,
```
banesullivan, line 261: Please gracefully handle keyword argument deprecations

pyvista/plotting/axes_actor.py
```diff
@@ -1,176 +1,416 @@
...
+        position=(0, 0, 0),
+        orientation=(0, 0, 0),
+        origin=(0, 0, 0),
+        scale=(1, 1, 1),
+        user_matrix=np.eye(4),
+        visibility=True,
+        properties=None,
+        **kwargs,
+    ):
+        """Create a hybrid 2D/3D actor to represent 3D axes in a scene.
```
banesullivan, line 75: I think we want the docstrings on the class, not the `__init__`. At least this is how it is done everywhere else in the docs for PyVista. I believe as it stands, your docstring will not be rendered in our documentation.

Note: this is confusing and I recently opened an issue around the topic: #5137

pyvista/plotting/axes_actor.py
```diff
@@ -1,176 +1,416 @@
...
+                setattr(y_shaft, name, value)
+                setattr(z_shaft, name, value)
+                setattr(x_tip, name, value)
+                setattr(y_tip, name, value)
+                setattr(z_tip, name, value)
+
+        if len(kwargs) > 0:
+            raise TypeError(
+                f"AxesActor() got an unexpected keyword argument '{list(kwargs.keys())[0]}'"
+            )
```
banesullivan, line 285: Please use `assert_empty_kwargs` for this. 

https://github.com/pyvista/pyvista/blob/76aa761dac8dddf9995a49872399e4f9a406ecf8/pyvista/core/utilities/misc.py#L12

pyvista/plotting/axes_actor.py
```diff
@@ -1,176 +1,416 @@
...
+    def shaft_length(self, length: Union[float, Sequence[float]]):  # numpydoc ignore=GL08
+        if isinstance(length, Sequence):
             self.SetNormalizedShaftLength(length[0], length[1], length[2])
         else:
             self.SetNormalizedShaftLength(length, length, length)
 
     @property
-    def tip_length(self) -> tuple:  # numpydoc ignore=RT01
-        """Return or set the length of the tip.
+    def tip_length(self) -> Tuple[float, float, float]:  # numpydoc ignore=RT01
```
banesullivan, line 410: We have the `Vector` type for `Tuple[float, float, float]`. See:

https://github.com/pyvista/pyvista/blob/76aa761dac8dddf9995a49872399e4f9a406ecf8/pyvista/core/_typing_core.py#L8

pyvista/plotting/axes_actor.py
```diff
@@ -189,195 +429,351 @@ def tip_length(self) -> tuple:  # numpydoc ignore=RT01
...
 
     @cone_resolution.setter
-    def cone_resolution(self, res: int):  # numpydoc ignore=GL08
-        self.SetConeResolution(res)
+    def cone_resolution(self, resolution: int):  # numpydoc ignore=GL08
+        warnings.warn(
+            "Use of `cone_resolution` is deprecated and will be removed in 0.46.0. "
+            "Use `tip_resolution` instead.",
+            PyVistaDeprecationWarning,
+        )
```
banesullivan, line 518: Please make sure you add the comment to help us grep for deprecations during releases.

Ref https://github.com/pyvista/pyvista/blob/main/CONTRIBUTING.rst#deprecating-features-or-other-backwards-breaking-changes

banesullivan (changes requested):
@user27182, overall, fantastic work!! I'm really excited to see this drastic overhaul of the axes actor. 

I've left a a good bit of comments which are mostly around making sure the *many* deprecations here are handled with care. This was mostly a cursory review and I think this warrants a closer review after this initial feedback is addressed.

I may not watch this thread too closely, so if something I'm bringing up is unclear or off-base, please don't hesitate to ping me on Slack (or ask @tkoyama010 too ping me there if you aren't on the Slack)

Really exciting to have you contributing, @user27182!

pyvista/plotting/prop3d.py
```diff
@@ -1,15 +1,19 @@
...
 from pyvista.core._typing_core import BoundsLike, Vector
-from pyvista.core.utilities.arrays import array_from_vtkmatrix, vtkmatrix_from_array
+from pyvista.core.utilities.arrays import (
+    _coerce_transformlike_arg,
+    array_from_vtkmatrix,
+    vtkmatrix_from_array,
+)
 
-from . import _vtk
+from ._vtk import vtkMatrix3x3, vtkMatrix4x4, vtkProp3D, vtkTransform
```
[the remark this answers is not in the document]
user27182: I have reverted back to use _vtk namespace in ff2b60f3 and f64719bc.

pyvista/plotting/prop3d.py
```diff
@@ -41,8 +45,11 @@ def scale(self) -> Tuple[float, float, float]:  # numpydoc ignore=RT01
         return self.GetScale()
 
     @scale.setter
-    def scale(self, value: Vector):  # numpydoc ignore=GL08
-        return self.SetScale(value)
+    def scale(self, value: Union[float, Sequence[float]]):  # numpydoc ignore=GL08
+        if isinstance(value, Sequence):
+            self.SetScale(*value)
+        else:
+            self.SetScale(value)
```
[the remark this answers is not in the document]
user27182: You are correct, it is not needed. The example code you gave runs just fine. This has been reverted in 95ce9ed9 along with a test added for the case mentioned below.

The condition was added not for the cases you gave, but for the case where only a single number is given, i.e. `actor.SetScale(1)`. It turns out `vtkProp3D` is overloaded for this case, so it is not needed. I must have gotten mixed up with `vtkTransform`, which is _not_ overloaded to handle this:
``` python
import vtk
trans = vtk.vtkTransform()
trans.Scale(1)  # will raise a `TypeError`
```

pyvista/plotting/prop3d.py
```diff
@@ -41,8 +45,11 @@ def scale(self) -> Tuple[float, float, float]:  # numpydoc ignore=RT01
         return self.GetScale()
 
     @scale.setter
-    def scale(self, value: Vector):  # numpydoc ignore=GL08
-        return self.SetScale(value)
+    def scale(self, value: Union[float, Sequence[float]]):  # numpydoc ignore=GL08
```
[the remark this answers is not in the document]
user27182: This change to the typing has been made in 7aaaa1e0

`Vector` itself has `List[float]` in its `Union`, should it be changed to `List[float, float, float]` for the same reason?

pyvista/plotting/prop3d.py
user27182: This change to the typing has been made in 7aaaa1e0

pyvista/plotting/prop3d.py
user27182: This change to the typing has been made in 7aaaa1e0

pyvista/plotting/axes_actor.py
user27182: This is the getter, which will definitely return a `tuple`. Do you want it as `Vector` anyway?

The setter looks like its type can be changed to use `Vector` if you want though (along with all other variables and the class docstring).

pyvista/plotting/tools.py
banesullivan: Thanks for going back to that!

pyvista/plotting/prop3d.py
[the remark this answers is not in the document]
banesullivan: Afaik, we cannot specify defined length list types (ie. `List[float, float, float]` isn't valid), you can only do this for `Tuple`

pyvista/plotting/tools.py
banesullivan: Oh, hm... That's not good. We should really have a lint check to make sure there aren't unused keyword arguments. Thanks for catching that.

pyvista/plotting/axes_actor.py
banesullivan: Ah, good catch. If the returned value is always a tuple, then lest keep it as is.

banesullivan: >The setter looks like its type can be changed to use Vector if you want though

The setter should probably use the `Vector` type

banesullivan:
@user27182, thanks for bearing with me and addressing everything so well! Let me know when you're ready for a final review and I'll check this out locally and give it a go. Would be great to land this for 0.43 if we can and you think we should

pyvista/plotting/axes_actor.py
user27182: The typing for the whole class was updated in 769930c9b98676ae02e83839ff37fe052b1edf8d

pyvista/plotting/tools.py
user27182: create_axes_marker has been deprecated in 8565a4565543f23a87e29d420f38221d2796f295

user27182:
While testing this feature I found that `vtkAxesActor.GetBounds()` is hard-coded to be symmetric about the origin. Overriding `GetBounds()` only works for python calls, but doesn't affect the C++ calls to GetBounds() directly by VTK objects. With the hard-coded bounds, this meant that there was a bug where the camera/renderer would always insist on showing the origin, and not the actual axes if they were positioned elsewhere in space. 

I am working on a resolution to this problem, e.g. with a call to `AxesActor.UseBoundsOff()` so that the camera/renderer ignore the actor's bounds when rendering the scene.

user27182:
Update: the camera issues have largely been resolved and I am generally happy with the current changes. So far in my testing everything works as expected.

A key feature still missing though which I'd like to have is the ability to represent the axes in space to scale. E.g. if `shaft_radius` is set to `0.1`, the axes placed in a scene should have a true radius of `0.1`. This is currently not the case since, `vtkAxesActor` will scale the radius proportional to the length. As a result, if you have shafts with different lengths for each axis, the longer axis will also have a larger radius, despite all axes having the same radius property value.

A way to deal with this is to separately scale each actor. This is what's done in 0b35c0d. However, this doesn't actually work. This is because when the axes are rendered, calls to the `vtkAxesActor` rendering methods (e.g. [`RenderOpaqueGeometry`](https://github.com/Kitware/VTK/blob/92443f814f85ec4e47ab24a92e407a5feae58cc1/Rendering/Annotation/vtkAxesActor.cxx#L210)) will always first call the protected method `UpdateProps` (which cannot be accessed/overridden in Python), and this method will __always__ apply a single `UserMatrix` to all shaft and tip actors. Therefore, it's not possible to apply separate, independent scaling factors to each actor.

In order to allow for a separate `UserMatrix` to be applied to each actor, I am going to write a new `AxesActorAssembly` class. I'm not yet sure about the implementation, but this class will inherit either from `vtkAssembly` or `vtkPropAssembly` and have similar behaviour/API to `AxesActor` in that it will implement the same general methods and properties (e.g. `shaft_cylinder`, `x_color`), but internally it will hold 6 independent objects (probably 6 `AxesActor` objects), one for each axis (+XYZ) and one for each negative axis too (-X -Y -Z) to support symmetric axes.

pyvista/plotting/axes_actor.py
```diff
@@ -59,440 +368,1906 @@ class AxesActor(_vtk.vtkAxesActor):
...
+          and :attr:`tip_radius`, respectively.
+
+        Examples
+        --------
+        >>> import pyvista as pv
+        >>> axes_actor = pv.AxesActor(
+        ...     true_to_scale=True, shaft_radius=0.2, tip_radius=0.2
+        ... )
+        >>> p = pv.Plotter()
+        >>> p.show_grid()
```
tkoyama010, line 675:
```suggestion
        >>> _ = p.show_grid()
```

user27182:
Another update... The new `AxesAssembly` class works as desired in terms of being "true to scale". However I ran into issues implementing text labels. This is because `vtkAxesActor` uses 2D Actors for the text labels but 3D actors for the shaft and tip. The use of 2D actors is an issue because they don't work with `vtkAssembly` since they don't inherit from `vtkProp3D`.

This issue is highlighted in the [MovableAxes](https://examples.vtk.org/site/Cxx/Visualization/MovableAxes/) vtk example. There, `vtkVectorText` is used for the labels and they're used with `vtkFollower`. Since `vtkFollower` is a Prop3D, it can be added to the assembly. So, I am currently working on adding text labels this way.

#5218 adds the text source, and I'll submit another PR soon for `vtkFollower`.

user27182:
@tkoyama010 @banesullivan Please see my recent comment above.

I am not sure what the timeline is for release 0.43, but if you want this PR to make it into that release I can forgo the addition of text labels in the new `AxesAssembly` class for now, and wrap up the current implementation and clean things up and get this PR ready for (another) review. The scope of this PR keep swelling so perhaps this is a good idea regardless of the release schedule.

 Let me know what your preference is as I don't want to hold up the release unnecessarily.

tkoyama010:
> I am not sure what the timeline is for release 0.43, but if you want this PR to make it into that release I can forgo the addition of text labels in the new `AxesAssembly` class for now, and wrap up the current implementation and clean things up and get this PR ready for (another) review. The scope of this PR keep swelling so perhaps this is a good idea regardless of the release schedule.
> 
> Let me know what your preference is as I don't want to hold up the release unnecessarily.

The release will take place on the 30th. If we want to include it in the release, we need to merge the PR into the main branch by then. Our activity is volunteer, so there is no need to force a release in time. Let's take it easy :)

That being said, it is a very good idea to make the scope of PR smaller. It will clarify the purpose of the PR and reduce the burden on the reviewer.

user27182:
Thanks for the feedback. I'll keep working on it and see how it goes. The text label properties have already been abstracted to `AxesProp` so removing them from the implementation at this point is not trivial. I'll create separate PRs as needed, e.g. for `Text3DSource` and  a new `Follower` class to limit the scope of this PR.

tkoyama010:
pre-commit.ci autofix

tkoyama010:
github-actions preview

tkoyama010:
pre-commit.ci autofix

user27182:
I am rethinking this PR altogether.

1. In its current state it is much too large for review.
2. It has high impact since so many things use an axes widget. 
3. It uses some ugly hacks to make the axes orientable.
4. The current implementation of `AxesActor` needs a lot of work to improve the API.

However, the only thing making it high impact is the "unifying" part of this PR, which directly links `plotting.tools.create_axes_marker` to `plotting.AxesActor`, though this is not strictly necessary. This PR is also needlessly trying to improve the `AxesActor` API which is barely used, if at all, in the code base, and requires many changes/deprecations. And even if the `AxesActor` API is improved and unified with `create_axes_marker` and hacked to make it orientable, it is not even possible to correctly scale the shafts and/or tips to make the actor true to scale in a scene.

At its core I think the issue is the link placed between "axes as a widget" and "axes in scene", as if both use cases need to use the same `AxesActor`. But, there is no need to  force the use `AxesActor` to implement the actual desired feature from this PR, which is to be able to add "axes in scene" (and have it true to scale).

So, I am thinking of moving away from using `AxesActor` and `vtkAxesActor`. Instead, just create a new class, give it the desired API without any need for deprecations, and make it work without any need to unify or integrate the new class with any existing axes-related stuff in the code base.

PR #5024, opened 2023-10-05
Allow deploying preview of the forked repository using  label

tkoyama010:
### Overview

Allow deploying preview of the forked repository using label.

### Details

- Currently, deploy preview from forked repositories is disabled. This is for security reasons. However, it is inconvenient that a preview is not available for review, so this PR lets the maintainer preview documents by labeling them. Security is ensured because only the maintainer can apply labels.
- Follo-up of #4934 .
- [x] Test in https://github.com/pyvista/pyvista-tutorial .

PR #5030, opened 2023-10-06
Revert "Allow deploying preview of the forked repository using  label"

tkoyama010:
Reverts pyvista/pyvista#5024
Labeled and checked, but the behavior was unstable.


tkoyama010:
LGTM

PR #5034, opened 2023-10-07
Define alias `pv` when importing `pyvista`

tkoyama010:
### Overview

Define alias when importing pyvista.

### Details

- This is helping PR when reviewing #5029.
- Also inspiring by #5029.
- [ ] Add blame ignore.


tkoyama010:
LGTM

Issue #1629, opened 2021-09-02
Improve example coverage of plotting module


akaszynski:
About 50% of our docstrings in PyVista have examples.  This is fairly good, especially considering that many of these objects are not publicly exposed.

However, for modules like `pyvista.plotting.plotting`, it would be great to see nearly 100% coverage considering how important it is to be able to demonstrate the individual methods of the Plotting class.

This is a great first issue since it doesn't involve changing the code base, just providing an example for some of our docstrings in [plotting](https://github.com/pyvista/pyvista/blob/main/pyvista/plotting/plotting.py)

Feel free to post questions here on how to get started if you're interested.

Note that many of these examples don't need to include `plotter.show`, and if they write files you can feel free to use the `# doctest:+SKIP` directive

```
Name                                         Docstrings     Missed   Covered
----------------------------------------------------------------------------
pyvista.plotting.plotting                           155        110     29.0%
```

Some easy ones:
- [ ] `Plotter.export_vtkjs`
- [ ] `Plotter.fly_to`
- [ ] `Plotter.remove_legend`
- [ ] `Plotter.save_graphic`
- [ ] `Plotter.add_point_labels`
- [ ] `Plotter.add_lines`
- [ ] `Plotter.add_text`

AleenaSA:
Hi, this sounds interesting as a first contribution. Could you give me more pointers on how to begin?

akaszynski:
Add an `Examples` section in one of the methods in the `Plotter` class.  For example, `add_text`:
https://github.com/pyvista/pyvista/blob/598fb020618a0bfa4803039768ce51be4fd5ca68/pyvista/plotting/plotting.py#L2925

See the other example sections like the one for ``add_background_image``:
https://github.com/pyvista/pyvista/blob/598fb020618a0bfa4803039768ce51be4fd5ca68/pyvista/plotting/plotting.py#L4193-L4200

Create a PR with just that new section and you'll be good to go.

AleenaSA:
The [PR](https://github.com/pyvista/pyvista/pull/1640) for `add_text` is merged, shall I look into others mentioned above as well? Individual PRs are required or one consolidated one is fine?

akaszynski:
> The [PR](https://github.com/pyvista/pyvista/pull/1640) for `add_text` is merged, shall I look into others mentioned above as well? Individual PRs are required or one consolidated one is fine?

That would be great!  You can do individual or get several in one depending on how ambitious you'd like.

[...]

PR #5036, opened 2023-10-08
Add examples for widgets methods

abhiramtilakiiit:
### Overview

Add minimal but informative, examples for different functions in `widgets.py` as referenced in #1629.

Also fixed an existing mismatch in examples.

### Details

The examples were added to the following functions in `widgets.py` file:

- `add_mesh_clip_box`
- `add_mesh_clip_plane`
- `add_mesh_clip_isovalue`
- `add_mesh_slice`
- `add_mesh_slice_orthogonal`

The example in function `add_box_widget`, doesn't explicitly use the function, instead uses the `add_mesh_clip_box`, but in my opinion it deserves its own example. So, I have moved the example under `add_box_widget` into `add_mesh_clip_box`.

### DIFF
--- a/pyvista/plotting/widgets.py
+++ b/pyvista/plotting/widgets.py
@@ -662,7 +662,7 @@ class WidgetHelper:
         >>> vol = examples.load_airplane()
         >>> pl = pv.Plotter()
         >>> pl.add_mesh_clip_plane(vol, normal=[0, -1, 0])
-        >>> pl.show(cpos=[-2.1, 0.6 , 1.5])
+        >>> pl.show(cpos=[-2.1, 0.6, 1.5])
         >>> pl.plane_clipped_meshes  # doctest:+SKIP
 
         For a full example see :ref:`plane_widget_example`.
--- a/pyvista/plotting/widgets.py
+++ b/pyvista/plotting/widgets.py
@@ -661,7 +661,7 @@ class WidgetHelper:
         >>> from pyvista import examples
         >>> vol = examples.load_airplane()
         >>> pl = pv.Plotter()
-        >>> pl.add_mesh_clip_plane(vol, normal=[0, -1, 0])
+        >>> _ = pl.add_mesh_clip_plane(vol, normal=[0, -1, 0])
         >>> pl.show(cpos=[-2.1, 0.6, 1.5])
         >>> pl.plane_clipped_meshes  # doctest:+SKIP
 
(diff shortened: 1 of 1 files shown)


pyvista/plotting/widgets.py
```diff
@@ -153,18 +153,6 @@ def add_box_widget(
...
-        Shows an interactive clip box.
-
-        >>> import pyvista as pv
-        >>> mesh = pv.ParametricConicSpiral()
-        >>> pl = pv.Plotter()
-        >>> _ = pl.add_mesh_clip_box(mesh, color='white')
-        >>> pl.show()
-
-        For a full example see :ref:`box_widget_example`.
-
```
tkoyama010, line 167: Thank you for moving the example to another correct place. Could you add the revised example to the deleted section?

tkoyama010 (changes requested):

pyvista/plotting/widgets.py
abhiramtilakiiit: I have more examples for `add_box_widget` and `add_line_widget`, but I'm not sure, if I am
doing it the right way. 
I also want to add `add_plane widget `, but I'm unable to come up with a better example than one given in the [examples](https://docs.pyvista.org/version/stable/examples/03-widgets/plane-widget.html) with carotid.

tkoyama010: @abhiramtilakiiit Thanks for comment.
Please commit any examples you are　considering. We can review and discuss them together. If you don't have any that you can commit, I will commit a draft and you can mentions me.

abhiramtilakiiit: commit [f166a74](https://github.com/pyvista/pyvista/pull/5036/commits/f166a749b4866565a87289c79a9a1d735ed612c2) resolves this.

tkoyama010:
.pre-commit.ci autofix

pyvista/plotting/widgets.py
```diff
@@ -437,6 +467,33 @@ def add_plane_widget(
         vtk.vtkImplicitPlaneWidget or vtk.vtkPlaneWidget
             Plane widget.
 
+        Examples
+        --------
+        Shows an interactive plane moving along x-axis in random-hill example, which is used to mark the max altitude
+        at a particular distance x.
```
tkoyama010, line 473:
```suggestion
        Shows an interactive plane moving along the x-axis in the random-hill example, which is used to mark the max altitude
        at a particular distance x.
```

tkoyama010 (approved):
LGTM. Thanks for your awesome work!
image
image
image
image
image
image
image

PR #5037, opened 2023-10-09
Add define alias `pv` commit to `.git-blame-ignore-revs` file

tkoyama010:
### Overview

Follow-up of #5034 .

### Details

- None


tkoyama010:
LGTM

PR #5039, opened 2023-10-09
Fix ParametricRandomHills parameters names by PEP8

tkoyama010:
### Overview

I fixed the names of parameters of `ParametricRandomHills` because it is hard to recognize the name of the parameter.

### Details

- Following the [Naming Conventions](https://peps.python.org/pep-0008/#naming-conventions) in [PEP8](https://peps.python.org/pep-0008/).

### DIFF
--- a/pyvista/core/utilities/parametric_objects.py
+++ b/pyvista/core/utilities/parametric_objects.py
@@ -1,12 +1,14 @@
 """Module managing parametric objects."""
 
 from math import pi
+import warnings
 
 import numpy as np
 
 import pyvista as pv
 from pyvista.core import _vtk_core as _vtk
 
+from .errors import PyVistaDeprecationWarning
 from .geometric_objects import translate
 from .helpers import wrap
 from .misc import check_valid_vector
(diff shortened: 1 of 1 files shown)


tkoyama010:
pre-commit.ci autofix

tkoyama010:
LGTM

pyvista/core/utilities/parametric_objects.py
```diff
@@ -850,13 +852,21 @@ def ParametricPseudosphere(**kwargs):
 
 def ParametricRandomHills(
     numberofhills=None,
+    number_of_hills=None,
```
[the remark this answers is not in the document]
tkoyama010: You are right. Fixed in 2a2ba00b3588854188fc391dc20e8f5015fc6198 .

tests/core/test_parametric_geometry.py
```diff
@@ -128,6 +129,49 @@ def test_ParametricPseudosphere():
 def test_ParametricRandomHills():
     geom = pv.ParametricRandomHills()
     assert geom.n_points
+    with pytest.warns(PyVistaDeprecationWarning):
+        geom = pv.ParametricRandomHills(numberofhills=30)
+        if pv._version.version_info >= (0, 46):
+            raise RuntimeError('Remove this deprecated parameter')
```
MatthewFlamm, line 135: While these are nice to prevent the deprecation from sticking around too long.  Didn't it create a release problem last time?  Not sure if it is worth fixing here, or kicking down the road.

MatthewFlamm (approved):

tests/core/test_parametric_geometry.py
tkoyama010: As you say, there will raise an error in the release and that is not pleasant. However, we are still unable to come up with a better solution.

PR #5046, opened 2023-10-10
Fix bug with actor removal rendering

larsoner:
`render=render` must be passed when string lookup or iterable is used. Noticed this when I removed a few hundred actors... and it rendered after every one, making it take a few seconds rather than being very quick.


tkoyama010 (approved):
Thanks!

PR #5052, opened 2023-10-11
Add `_coerce_transformlike_arg` function

user27182:
### Overview
`_coerce_transformlike_arg` validates/coerces various input types into a 4x4 `np.ndarray` transformation matrix.

This function is already implemented in #4947, and is being pulled separately so that it can be merged and used by #5029.


pyvista/core/_typing_core.py
```diff
@@ -3,8 +3,11 @@
...
 import numpy as np
 
+from pyvista.core._vtk_core import vtkMatrix3x3, vtkMatrix4x4, vtkTransform
+
 Vector = Union[List[float], Tuple[float, float, float], np.ndarray]
 VectorArray = Union[np.ndarray, Sequence[Vector]]
 Number = Union[float, int, np.number]
 NumericArray = Union[Sequence[Number], np.ndarray]
 BoundsLike = Tuple[Number, Number, Number, Number, Number, Number]
+TransformLike = Union[np.ndarray, vtkMatrix3x3, vtkMatrix4x4, vtkTransform]
```
tkoyama010, line 13: It is more convenient to use `VectorArray` than `np.ndarray` in my mind.
```suggestion
TransformLike = Union[VectorArray, vtkMatrix3x3, vtkMatrix4x4, vtkTransform]
```

tkoyama010 (changes requested):

pyvista/core/_typing_core.py
user27182: Ok 👍. I'm fine with this change.
I have some comments though, maybe for a different PR if you prefer:

1. Is the type hinting for `NumericArray` correct? Currently, as it's defined doing `Sequence[NumericArray]` implies that mixing and matching `Tuples` with `NDArray` is allowed, something like `((1,2,3), np.array([4,5,6]))`. I don't think that's generally expected or handled though... Perhaps it should be something like this instead: 
```python 
NumericArray = Union[Sequence[Number], Sequence[Sequence[Number]], np.ndarray]
```
2. Also, perhaps there should be a `Matrix3x3Like` type and `Matrix4x4Like` type as well, something like:
``` python
NumericArray = Union[Sequence[Number], Sequence[Sequence[Number]], np.ndarray]
Matrix3x3Like = Union[VectorArray, vtkMatrix3x3] 
Matrix4x4Like = Union[NumericArray, vtkMatrix4x4] 
TransformLike = Union[Matrix3x3Like, Matrix4x4Like, vtkTransform]
```

tkoyama010: You seem to be right. Please give me time to think about it.

tkoyama010: Or you are welcome to raise an issue about it.

tkoyama010: Opened #5074 .

PR #5045, opened 2023-10-10
Add a Professional Support section

jourdain:
### Overview (doc)

This contribution aims to list contact information of companies who could provide support around PyVista.
So far, I've only listed Kitware, but ideally, that list could grow.

Adding a __Professional Support__ section at the bottom of the README and doc landing page.

PR #5057, opened 2023-10-12
Move Professional Support section to Previous of Index Section

tkoyama010:
### Overview

Move the Professional Support section to the Previous of Index Section.

### Details

- The section will not appear unless it is in the previous to the index section.
- Follo-up of #5045.


tkoyama010:
LGTM

PR #5040, opened 2023-10-10
Fix the docstring link error which is using an alias

tkoyama010:
### Overview

Fix the docstring link error which is using an alias.

### Details

- Follow up of #5034.

PR #5059, opened 2023-10-12
Add the docstring link error to `.git-blame-ignore-revs` file

tkoyama010:
### Overview

Follow-up of #5040 .

### Details

- I am considering whether to add this change to the coding rules.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -6,5 +6,6 @@ db0691cfb6b033e55664ffba25deb456a386a48c
 a26a52d037a265774be0ca06bb141e3346e73ad0
 635052d1bce8615fcc4879f5f200921a4f78aa1a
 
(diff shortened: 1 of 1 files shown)


adeak:
@tkoyama010 on the other PR you wrote

> Note: I will also add it to git-blame because of the many lines of change. 

I don't think that it's the number of lines that matters. I started pushing for the ignore-revs file when we added lots of _automated_ edits to a diff, and with edits that don't touch the actual contents in meaningful ways.

The point of the git blame is to see where bugs were introduced. As long as a change was done by a human (including a custom search and replace done by a human) I want to see it in the blame.

tkoyama010:
@adeak Thank you. I was only concerned about the number of lines, but it makes sense. I will close this PR and submit a Revert PR for #5037 as well.

PR #5070, opened 2023-10-12
Revert "Add define alias `pv` commit to `.git-blame-ignore-revs` file"

tkoyama010:
Reverts pyvista/pyvista#5037
See the comment in #5059.


adeak (approved):
Thanks a lot, LGTM.

PR #5072, opened 2023-10-13
Remove trame from `all` in pyproject.toml, as it is no longer a variable

jorgensd:
Remove `trame` as optional-dependency class, as it is no longer there, and only part of `jupyter` deps.


tkoyama010 (approved):
LGTM. Thanks for your first contribution!

PR #5079, opened 2023-10-13
Deprecate `t_coords` and use `texture_coordinates`

tkoyama010:
### Overview

I understand that `t_coords` was defined this way to omit the length of the variable name. However, as one user, it is difficult to understand the meaning of this variable.

### Details

- [x] Add Deprecate warning.


tkoyama010:
LGTM

Issue #487, opened 2019-12-02
Double buffering / render=False


jotaf98:
Is there any way to prevent partial results from being displayed, which results in jittery animation?

For instance, `update_coordinates` accepts:

`render : bool, optional
            Forces an update to the render window.  Default True.`

However, setting it to false does not really prevent an update from happening. Is there anything that would achieve a similar effect, or should this be considered a tentative "feature request"?

To be clear, this is not an issue when saving movies/gifs, just when relying on real-time window display.

banesullivan:
Hi @jotaf98! This is an interesting feature request and it's something I've never really thought about needing. Thus I think some hooks were added so that anytime a dataset is added to the scene or even modified that a render call is performed, hence why the `render=False` argument isn't working

we can certainly look into implementing this!

jotaf98:
Thanks for the positive reply!

Yes I agree that the default behavior, of always refreshing changes, is probably what most users want when they start. I guess this would also help with the time-based visualizations mentioned in other feature requests.

banesullivan:
You're exactly on point about time-based visualizations - we have render hooks because so many folks regularly modify the data in the scene and want real-time results to appear in the visualization. Your use case was definitely overlooked. 

This may take a bit for anyone to get around to implementing/fixing

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jotaf98:
Yes that's understandable.

As a shortcut, maybe there's already a function in VTK to "halt" updates globally / for a window, or a facility for double buffering (like many game engines do). Unfortunately I'm not familiar enough with VTK to tell whether that exists.

If true, then it may be simpler to implement (compared to preventing redrawing in each and every pyvista plot function, which would touch a lot of code).

Issue #447, opened 2019-11-16
Pressing V Crashes


akaszynski:
Pressing `v` kills the renderer:

```
~/.local/lib/python3.7/site-packages/pyvista/plotting/plotting.py in <lambda>()
    229         b_left_down_callback = lambda: self.iren.AddObserver('LeftButtonPressEvent', self.left_button_down)
    230         self.add_key_event('b', b_left_down_callback)
--> 231         self.add_key_event('v', lambda: self.isometric_view_interactive())
    232 
    233 

~/.local/lib/python3.7/site-packages/pyvista/plotting/plotting.py in isometric_view_interactive(self)
    420         interactor = self.iren.GetInteractorStyle()
    421         renderer = interactor.GetCurrentRenderer()
--> 422         renderer.view_isometric()
    423 
    424     def update(self, stime=1, force_redraw=True):

AttributeError: 'NoneType' object has no attribute 'view_isometric'
```

banesullivan:
Weird... I don't have this issue. why would the `interactor.GetCurrentRenderer()` call be returning `None`?

GuillaumeFavelier:
I confirm that I can reproduce although my application does not crash. The scenario where I reproduce it is when I don't click or interact at all with the window and just press 'v'. In this case, I get the message once and then I can continue using the application without perturbation.

banesullivan:
I suppose we could add a check if the current renderer returned is None, then amuse the active renderer that the plotter is tracking.  Should be a small patch

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #5081, opened 2023-10-14
Defer to geovista default warning escalation to error with pytest

bjlittle:
### Overview

`geovista` now performs full image tests using `pytest-pyvista` of its example scripts (see [#494](https://github.com/bjlittle/geovista/pull/494)).

However, until [Issue #447](https://github.com/bjlittle/geovista/issues/447) is resolved, `geovista` is pinned to `pyvista<0.42.2`.

This pull-requests skips the image tests to prevent `pyvista` from failing during integration tests.

Also, note that `geovista` now forces `pytest` to escalate all `warnings` to `errors` (see [#487](https://github.com/bjlittle/geovista/pull/487/files)) by default, so it's no longer necessary to include the explicit `-W error` option when running the `geovista` integration test.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -105,5 +105,5 @@ jobs:
           wget --quiet ${CARTOPY_FEATURE}
           mkdir -p ${CARTOPY_SHARE_DIR}
           python cartopy_feature_download.py physical --output ${CARTOPY_SHARE_DIR} --no-warn
-      - run: pytest -m "not image"
+      - run: pytest
         working-directory: geovista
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -105,5 +105,5 @@ jobs:
           wget --quiet ${CARTOPY_FEATURE}
           mkdir -p ${CARTOPY_SHARE_DIR}
           python cartopy_feature_download.py physical --output ${CARTOPY_SHARE_DIR} --no-warn
-      - run: pytest -W error -W "ignore:numpy.ndarray size changed:RuntimeWarning"
+      - run: pytest -m "not image"
```
tkoyama010, line 108: Thanks, I agree with removing the Warning option, but we should not add the "not image" option in my mind. Let's keep CI error and that will be the driving force to resolve https://github.com/bjlittle/geovista/issues/447. (Sorry for the delay in resolving the issue.)
```suggestion
      - run: pytest
```

tkoyama010 (changes requested):

tkoyama010:
I merge this as this is hotfix.

PR #5084, opened 2023-10-15
Fix PyVistaDeprecationWarning not imported error

tkoyama010:
### Overview

Follo-up of #5079 . This PR happens because of forgetting to update the branch before mergeing. Very sorry.

### Details

- None


tkoyama010:
The cause of this failure was that `Require branches to be up to date before merging` was turned off. Turning it on will prevent a recurrence.
image

PR #5088, opened 2023-10-15
Remove deprecations in v0.38.0 to remove at v0.43.0

tkoyama010:
### Overview

Remove deprecations in v0.38.0 to remove at v0.43.0.

### Details

- None


pyvista/plotting/widgets.py
```diff
@@ -1310,10 +1310,6 @@ def add_text_slider_widget(
         **kwargs : dict, optional
             Deprecated keyword arguments.
```
dcbr, line 1311:
```suggestion
```
I guess this can also be removed now?

pyvista/plotting/widgets.py
```diff
@@ -1465,10 +1461,6 @@ def add_slider_widget(
...
```
dcbr, line 1463:
```suggestion
```
I guess this can also be removed now?

pyvista/plotting/widgets.py
tkoyama010: Good catch! I didn't notice that.

pyvista/plotting/widgets.py
tkoyama010: Good catch! I didn't notice that.

dcbr (approved):
LGTM

PR #4938, opened 2023-09-20
Add interactive plots in documentation

ChristosT:
### Overview

An attempt to bring back the interactive plots in the documentation of pyvista .

This pull request  replaces current "pyvista" scraper with the [Dynamic scraper](https://github.com/pyvista/pyvista/pull/4569) in order to allow for interactive examples in the pyvista documentation. 

resolves #4877

### Details

The `pyvista-plot`  directive  will now produce interactive scenes by default.
For cases where this does not work well or a static image is preferred a new option
`force_static`  is introduced. See example [here](https://github.com/pyvista/pyvista/pull/4938/files#diff-1e3ed06e90fe3bf5ecf311ac1bd22215144cfc8c0924f5e05e3394573d630cebR78)

The gallery extension is also updated so that documentation generated from examples contains also interactive scenes.
For this case we provide a fallback option per document (`.py` file): just define
`PYVISTA_GALLERY_FORCE_STATIC_IN_DOCUMENT = True` as a global variable in the script.
More fine-grained control can be given by `PYVISTA_GALLERY_FORCE_STATIC = True/False` which can be defined in block scope [example](https://github.com/pyvista/pyvista/pull/4938/files#diff-bc3a9b424841ec00a624c48e8c9139dd29e3d89712ca6880d97066a301e5b6b1R36).

In both cases the variable can be hidden from the generated documentation via `sphinx_gallery_start_ignore` / `sphinx_gallery_start_ignore` guards.

I went through all the examples and rest of the documentation and turned back to static where the result was sgnificantly different from the current static version. (I may have missed something of course).

### Future work:
- [ ]  global theme options do not seem to work (see here) , not sure why . I can look into this in a separate pull request.
- [ ]  Volume rendering examples do not work. Talking with other developers from Kitware there is a suspicion that it is related with he SmartVolumeMapper that pyvista uses by default for rendering. I will need to investigate this further. For now I left the static images wherever a volume rendering is performed. 
- [ ] labels/text is not working for interactive examples I used static images in this case. From what I understand this is a limitation of vtk-js. 
- [ ] specular lighting is not working correctly and the result is overexposed, again I fallback to static images. No idea where the problem lies needs more investigation.

Note that as [before](https://github.com/pyvista/pyvista/pull/4569#issuecomment-1621994476) to test this locally you need to run a http server `python -m http.server` in the html directory.

PR #5091, opened 2023-10-16
Add Preview the Documentation section in CONTRIBUTING

tkoyama010:
### Overview

Add Preview the Documentation section in CONTRUBUTING.

### Details

- None


MatthewFlamm:
`pull_request_target` runs the action based on the base commit ref, not the PR head ref.  This means that it will build the documentation based on the base commit ref, which is usually the head of the main branch.  I don't think this is useful for this workflow since it will build the documentation without the changes in the PR.  What is the issue with running as just `pull_request`?

MatthewFlamm:
One option is to checkout the PR branch in the `actions/checkout` step, but this would have similar levels of security vulnerability. These options are all recommended against in https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

If we really want this combination (`pull_request_target` but checkout the PR code), I think the `pull_request_target` combination with a _specific_ label would be required.  As I read the current workflow, any label could trigger the execution.  This might be triggered by the auto-label bot for example.  I'm hesitant to recommend this given the explicit discouragement of this practice by GitHub.

tkoyama010:
> `pull_request_target` runs the action based on the base commit ref, not the PR head ref. This means that it will build the documentation based on the base commit ref, which is usually the head of the main branch. I don't think this is useful for this workflow since it will build the documentation without the changes in the PR. What is the issue with running as just `pull_request`?

@MatthewFlamm 

Good point. The reason is that `pull_request` cannot see the following Netlify secret key from the forked repository.
https://github.com/pyvista/pyvista/blob/d8aa00882d1d77f7242dc51ab321f707f3718dd4/.github/workflows/preview-docs.yml#L90-L91

PullRequests by branches created in the repositories of projects that we, the maintainers, create are not a problem, since we can see the secret key in the `pull_request`. However, when a new contributor does this from a forked repository, the secret key cannot be seen and the Preview Documentation is not deployed.

This is inconvenient, but very important for security. For more information, please see the following blog

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

This blog states that the safest way is to label and trigger every time a deployment is needed. This is because it is the maintainer who can label, and they can make sure their PR is safe when labeling.

Note that this tedious process is only necessary for the PR of a forked repository. Pull requests that we, the maintainers, create branch in the project repository are always triggered by pull_request so that the latest documentation is deployed.

Please comment on any areas that need explanation! Thanks for your interest in this issue. I welcome your ideas.

tkoyama010:
> One option is to checkout the PR branch in the `actions/checkout` step, but this would have similar levels of security vulnerability. These options are all recommended against in https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
> 
> If we really want this combination (`pull_request_target` but checkout the PR code), I think the `pull_request_target` combination with a _specific_ label would be required. As I read the current workflow, any label could trigger the execution. This might be triggered by the auto-label bot for example. I'm hesitant to recommend this given the explicit discouragement of this practice by GitHub.



> One option is to checkout the PR branch in the `actions/checkout` step, but this would have similar levels of security vulnerability. These options are all recommended against in https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
> 
> If we really want this combination (`pull_request_target` but checkout the PR code), I think the `pull_request_target` combination with a _specific_ label would be required. As I read the current workflow, any label could trigger the execution. This might be triggered by the auto-label bot for example. I'm hesitant to recommend this given the explicit discouragement of this practice by GitHub.

Yes. Sorry. I had missed this comment. You have it all figured out. I agree with the idea of limiting it to a specific label." What about creating a new label "accepted" and triggering it only then?

.github/workflows/preview-docs.yml
```diff
@@ -0,0 +1,96 @@
...
+    types: [labeled]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  doc:
+    name: Build Documentation
+    runs-on: ubuntu-20.04
```
tkoyama010, line 14: Our idea is to allow this CI to run when labeled like `PreviewDocumentation-accepted`.
```suggestion
    runs-on: ubuntu-20.04
    if: contains(github.event.pull_request.labels.*.name, 'PreviewDocumentation-accepted')
```

tkoyama010:
@pyvista/maintainers  There are several other PRs that need this PR (especially #4938 ), so unless there are any comments, we will merge this PR after 3 days.

.github/workflows/preview-docs.yml
```diff
@@ -0,0 +1,97 @@
+name: Preview Documentation
+on:
+  pull_request:
+  pull_request_target:
```
MatthewFlamm, line 4: I think we can get rid of `pull_request_target` altogether.  I assume we want the `types: [labeled]` line under `pull_request`?

.github/workflows/preview-docs.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Copy ads.txt
+        run: cp doc/source/ads.txt doc/_build/html/
+
+      - name: Dump Sphinx Warnings and Errors
+        if: always()
+        run: if [ -e doc/sphinx_warnings.txt ]; then cat doc/sphinx_warnings.txt; fi
+
+      - name: Dump VTK Warnings and Errors
+        if: always()
+        run: if [ -e doc/errors.txt ]; then cat doc/errors.txt; fi
```
MatthewFlamm, line 82: If we are merging this, it would be great if it is possible to use a reusable workflow for building the documentation. This way we don't have to maintain building the documentation in both the original workflow and this new one.  Maybe a future PR...

tkoyama010: Very good point. I will of course need that refactoring in the future. Currently, we are trying to implement it as quickly as possible. Refactoring is not too late once this system is fully used.

.github/workflows/preview-docs.yml
tkoyama010: GitHub action workflow cannot access secret variables by `pull_request` even if it is triggered by a label in my mind. Please let me know if my understanding is incorrect. You probably know more about this than I do.

tkoyama010:
@MatthewFlamm Sorry, I merge this to help #4938. Let's discuss the point of concern in a new PR or issue.

tkoyama010:
LGTM

MatthewFlamm:
I think this is either not going to work, i.e. by using `pull_request_target` or using `pull_request` without elevating permissions, or is an active security risk, i.e. using `pull_requset` with active security and not restricting it's use to when a certain label is applied.

Issue #4312, opened 2023-04-20
Create a voxelize feature that takes (n_x, n_y, n_z) as parameters instead of just density


josephko91:
### Describe the feature you would like to be added.

Hi! It would be great if PyVista had a feature that allowed for the voxelization of a mesh according to a specified 3d-grid encompassing the object. The current voxelize feature allows for the adjustment of a density parameter to control the voxel size, but there is no way to directly create for example a 256 by 256 by 256 voxel grid centered on bounding box of a given mesh. For example, implementing something similar to the voxelization capabilities binvox or pyntcloud would be super useful in PyVista. It seems like perhaps this is already possible in vtk, but it's not entirely clear to me as I'm not too familiar with the inner workings of vtk. 

binvox: https://www.patrickmin.com/binvox/
pyntcloud: https://pyntcloud.readthedocs.io/en/latest/structures.html highlight=add_structure#pyntcloud.PyntCloud.add_structure

### Links to VTK Documentation, Examples, or Class Definitions.

https://vtk.org/doc/nightly/html/classvtkVoxelModeller.html

banesullivan:
I think we'd just need to modify the `voxelize()` method to use `np.linspace` instead of `arange`. Taking that code, here is an example:

```py
import numpy as np
import pyvista as pv
from pyvista import examples

mesh = examples.download_foot_bones()
surface = mesh.extract_geometry()

nx, ny, nz = 100, 75, 50

x_min, x_max, y_min, y_max, z_min, z_max = mesh.bounds
x = np.linspace(x_min, x_max, nx)
y = np.linspace(y_min, y_max, ny)
z = np.linspace(z_min, z_max, nz)
x, y, z = np.meshgrid(x, y, z)

# Create unstructured grid from the structured grid
grid = pv.StructuredGrid(x, y, z)
ugrid = pv.UnstructuredGrid(grid)

# get part of the mesh within the mesh's bounding surface.
selection = ugrid.select_enclosed_points(surface, tolerance=0.0, check_surface=True, progress_bar=True)
mask = selection.point_data['SelectedPoints'].view(np.bool_)

# extract cells from point indices
vox = ugrid.extract_points(mask, progress_bar=True)
```

-----

Thanks for pointing out `vtkVoxelModeller` -- we should look into using this directly in PyVista.

josephko91:
Thanks @banesullivan! Example worked perfectly and thanks for the response.

josephko91:
[...]

PR #5095, opened 2023-10-16
Improved Voxelization - Volume and Discretised Voxels

bsburnham:
### Overview
Hi `PyVista` team, 

I've been working with @banesullivan to extend the functionality of the `voxelize` feature.  This extension will allow users to construct a voxel volume based on the input mesh. Volume dimensions are equivalent to the input mesh boundaries and is generated from the input mesh origin. The volume is voxelised to the specified `density` value(s). Voxel cells that are _within_ or _intersect_ the input mesh are discretised from volume and added to final output grid as separate element (Fig. 1). Normal voelisation process remains if the `vol` argument is **_not_** set to `True` or `vol=False`. (Fig. 2)

This functionality should help to address and close Issue 248 (https://github.com/pyvista/pyvista-support/issues/248). This is also related to and may help close Issue 4312 (https://github.com/pyvista/pyvista/issues/4312)?

### Details

- If the `vol` argument is set to `True`, then a voxel volume is created. Once voxelised, all cells that _inersect_ and are _within_ the input mesh surfaces are discretised from the volume and added as an element of the final output grid. The discretised cells are labelled `MeshCells` (Fig. 1).
- All cells that are _**not**_ _within_ or _intersect_ the input mesh are added as a separate element to the final output grid. These cells are labelled `BackgroundCells` in the output grid.
- If the `vol` argument is set to `True`, the final output is a `StructuredGrid`. 
- If the `vol` argument is not chosen, or set to `False`, the normal voxelisation process is applied to the input mesh. The final output grid of the normal voxelisation process is still an `UnstructuredGrid`. Default is `vol=False`.
- Docstring is updated to include description of the `vol` agrument addition. Other text updated to clarify some processes. 
- Removed `ugrid = pyvista.UnstructuredGrid(grid)` from normal voxelisation process because `extract_points` method returns an `UnstructuredGrid` by default. 

#### TODO
- Add option, or output, that allows users to populate cells based on the approximate volume of the input mesh in each voxel after the voxelisation process is complete. e.g. if input mesh volume, in a single voxel cell, is >50%, then add a property to that voxel cell. Perhaps this is something that `SelectedPoints` can help with?
- Ideally, the final output grid from the normal voxelisation process is a `StructuredGrid`. This is something that may be resolved if the `extract_points` method is not used as this returns an `UnstructuredGrid`.

**Fig. 1**
image
#### Orthongal slices through voxelised volume if `vol = True`. Returns : `pyvista.StructuredGrid`
----
**Fig. 2**
image
#### Voxelised mesh when `vol=False` or not chosen. Returns : `pyvista.UnstructuredGrid`

P.S. First PR, so hopefully everything is laid out clearly and succinctly.


pyvista/core/utilities/features.py
```diff
@@ -77,25 +101,61 @@ def voxelize(mesh, density=None, check_surface=True):
...
+    
     x_min, x_max, y_min, y_max, z_min, z_max = mesh.bounds
     x = np.arange(x_min, x_max, density_x)
     y = np.arange(y_min, y_max, density_y)
     z = np.arange(z_min, z_max, density_z)
     x, y, z = np.meshgrid(x, y, z, indexing='ij')
     # indexing='ij' is used here in order to make grid and ugrid with x-y-z ordering, not y-x-z ordering
     # see https://github.com/pyvista/pyvista/pull/4365
+    
+    if vol == True:
```
banesullivan, line 113: nit:

```suggestion
    if vol:
```

pyvista/core/utilities/features.py
```diff
@@ -36,14 +44,20 @@ def voxelize(mesh, density=None, check_surface=True):
...
     structured coordinates.
-
+    
+    TODO: Add option to populate voxels with array elements based on
+    the approximate volume of the input mesh in each voxel.
+    e.g., input mesh volume comprises >50% of individual voxel volume
+    
     Examples
-    --------
+    ----------
```
banesullivan, line 60: Did you mean to change this? The number of `-` should match the number of characters in te line above

pyvista/core/utilities/features.py
```diff
@@ -52,10 +66,20 @@ def voxelize(mesh, density=None, check_surface=True):
...
     >>> vox.plot(show_edges=True)
 
+    ----------
     Create a voxelized mesh using unequal density dimensions.
 
     >>> vox = pv.voxelize(mesh, density=[0.01, 0.005, 0.002])
     >>> vox.plot(show_edges=True)
+    
+    ----------
+    Create a voxel volume from input mesh.
```
banesullivan, line 76: We don't want these `----------` separators in the example section

pyvista/core/utilities/features.py
```diff
@@ -77,25 +101,61 @@ def voxelize(mesh, density=None, check_surface=True):
     if not surface.is_all_triangles:
         # reduce chance for artifacts, see gh-1743
         surface.triangulate(inplace=True)
-
+
```
banesullivan, line 104: Make sure to turn on "Trim trailing whitespace" in your IDE. For example in VSCode:

hZrUY

pyvista/core/utilities/features.py
```diff
@@ -13,8 +13,11 @@
 from .helpers import wrap
 
 
-def voxelize(mesh, density=None, check_surface=True):
-    """Voxelize mesh to UnstructuredGrid.
+def voxelize(mesh, density=None, vol=False, check_surface=True):
+    """Voxelize mesh to create an UnstructuredGrid or a StructuredGrid.
+       1. Voxelize the input mesh : UnstructuredGrid
+       2. Create a voxel volume of the input mesh. Volume is constructed to 
+          boundaries of input mesh : StructuredGrid
```
banesullivan, line 20: I suspect our docstrings formatter will complain about this. Might need to separate the list from the first sentence with an empty line

banesullivan (changes requested):
Awesome contribution and improvement! I want to think about whether this should be a separate function and solicit feedback from others as it feels like now `voxelize()` has two separate modes... which typically we implement as two separate functions.

Otherwise, some nitpicky comments.

Thanks for the contribution, @bsburnham!!

tkoyama010:
pre-commit.ci autofix

tkoyama010:
@bsburnham
Great work. Could you also check the **pre-commit.ci - pr** [Details](https://results.pre-commit.ci/run/github/92974124/1697500090.wgQktQi2RgKklCkC6NNNwg)? It says exactly the same thing that @banesullivan said.

pyvista/core/utilities/features.py
bsburnham: I did not mean to change this, apologies.

bsburnham:
Thanks for the suggestions both. I've made the suggested revisions. Hopefully this should pass all checks. 

@banesullivan, I like the idea of separating the `voxelize()` function into two functions. A `volume_voxelzation()` function would be more appropriate for some meshes whilst the current `voxelize()` function is appropriate for others.

tkoyama010:
pre-commit.ci autofix

tkoyama010:
@bsburnham Could you add test code to this PR to resolve test coverage errors? If you don't know how to do this, please mentions me for support.

MatthewFlamm:
Can we use the built-in blanking feature of `vtkStructiredGrid`?  From that documentation:

>We should use [vtkStructuredGrid](https://vtk.org/doc/nightly/html/classvtkStructuredGrid.html) has the ability to blank, or "turn-off" points and cells in the dataset. This is done by setting [vtkDataSetAttributes::HIDDENPOINT](https://vtk.org/doc/nightly/html/classvtkDataSetAttributes.html#a0ce219ff851f79c0a360faef0b36d9eda44f15296518daaa03e404681e4912cf1) or [vtkDataSetAttributes::HIDDENCELL](https://vtk.org/doc/nightly/html/classvtkDataSetAttributes.html#a9881ee5ce6e9cdb8ff94ce98b4815af4a077b78cc13e863df89d96d44de8ddb54) in the ghost array for each point / cell that needs to be blanked


On this point:
> I want to think about whether this should be a separate function and solicit feedback from others as it feels like now voxelize() has two separate modes... which typically we implement as two separate functions.

One observation is that it looks like the current implementation has a completely separate if-else block for the functionality.  More of the code before the if block could also be moved inside the else block too.  This is an indication to me that we should have two separate functions.

pyvista/core/utilities/features.py
```diff
@@ -13,8 +13,11 @@
 from .helpers import wrap
 
 
-def voxelize(mesh, density=None, check_surface=True):
-    """Voxelize mesh to UnstructuredGrid.
+def voxelize(mesh, density=None, vol=False, check_surface=True):
```
MatthewFlamm, line 16: I cannot find `pv.voxelize` in the documentation, can it be added in this PR (in addition to a second function of it is added)?

pyvista/core/utilities/features.py
```diff
@@ -13,8 +13,11 @@
...
+    """Voxelize mesh to create an UnstructuredGrid or a StructuredGrid.
+
+       1. Voxelize the input mesh : UnstructuredGrid.
+       2. Create a voxel volume that encloses the input mesh : StructuredGrid.
```
MatthewFlamm, line 20: Both modes here will not give meshes with `vtkVoxel` cells, but rather `vtkHexahedron` cells.  Why aren't we using `RectilinearGrid` for these? If we use the vtk blanking for this new mode, I think we need `StructuredGrid`.  If we don't, I think we can use `RectilinearGrid` to get true voxels.

pyvista/core/utilities/features.py
```diff
@@ -86,16 +108,52 @@ def voxelize(mesh, density=None, check_surface=True):
     # indexing='ij' is used here in order to make grid and ugrid with x-y-z ordering, not y-x-z ordering
     # see https://github.com/pyvista/pyvista/pull/4365
 
-    # Create unstructured grid from the structured grid
-    grid = pyvista.StructuredGrid(x, y, z)
-    ugrid = pyvista.UnstructuredGrid(grid)
+    if vol:
```
MatthewFlamm, line 111: If we keep this as one function, we should move the above code that is specific to a mode into the else block below.

pyvista/core/utilities/features.py
bsburnham: This is a good point @MatthewFlamm. I've rewritten the function that facilitates generation a voxel volume. I've added an `if-else` statement that creates a `StructuredGrid` if the input `density` values **are** equal, and a `RectilinearGrid` if the `density` values are _not_ equal.

bsburnham:
I agree with @MatthewFlamm and @banesullivan's observations and suggestions that 2 separate functions may be the best way forward. One function that generates a voxel volume and second function that generates a voxelised representation of only the input mesh (similar to the current function). 

Perhaps the `CellGhost` methods of `vtkStructuredGrid` is appropriate as it has [CellGhostTypes](https://vtk.org/doc/nightly/html/classvtkDataSetAttributes.html#a9881ee5ce6e9cdb8ff94ce98b4815af4) like: `EXTERIORCELL` and `HIDDENCELL` that could be useful. However, and correct me if I'm wrong, `remove_cells` function is only currently available for `UnstructuredGrids`? Maybe this is suitable for another PR to add it as a function that my be used for `StructuredGrids`?

@tkoyama010 thanks for the offer to help write some tests. I will probably take you up on that offer as I want to make sure I do it correctly (this is my first attempt at an OSS contribution).

bsburnham:
Hi all, I've aded a new function called `voxelize_volume()`, as suggested. This new function creates either a `StructuredGrid` or `RectilinearGrid`, depending on the `density` values given. If the `density` values are equal, a `StructuredGrid` is created. If the `density` values are unequal, a `RectilinearGrid` is created. 

I've revereted the `voxelize()` function back to the original code. 

@tkoyama010, would you be able to help write some tests for this new function?

tkoyama010:
pre-commit.ci autofix

pyvista/core/utilities/features.py
```diff
@@ -91,14 +96,167 @@ def voxelize(mesh, density=None, check_surface=True):
     ugrid = pyvista.UnstructuredGrid(grid)
```
tkoyama010, line 96: @bsburnham `flake8` says that `ugrid` is not used anymore. Are you sure to remove this?

pyvista/core/utilities/features.py
```diff
@@ -91,14 +96,167 @@ def voxelize(mesh, density=None, check_surface=True):
...
 
 
+def voxelize_volume(mesh, density=None, check_surface=True):
+    """Voxelize mesh to create a StructuredGrid or a RectilinearGrid voxel volume.
+
+    Creates a voxel volume that encloses the input mesh and discretises the cells
+    within the volume that intersect and are contained within the input mesh. A
+    StructuredGrid or RectilinearGrid is created based on density value(s) chosen.
+
+    1. Create a StructuredGrid if ``density`` values have equal dimensions : StructuredGrid.
```
MatthewFlamm, line 114: If we aren't going to use the 'blanking' functionality of `StructuredGrid`, we should use `ImageData` when density is single value for each or return `RectilinearGrid` in both cases.  I would lean towards the latter to keep a consistent return type, but I think either choice is fine.

pyvista/core/utilities/features.py
bsburnham: @tkoyama010 the `extract_points` method returns an `UnstructuredGrid`, so I'm not sure we need to create an empty `ugrid` variable. Unless I've missed something in the documentation?

pyvista/core/utilities/features.py
bsburnham: I take your point of having a consistent return type, however, I'm not sure a `RectilinearGrid` is sufficient for a voxel volume that's created with an **equal** `density` because a `RectilinearGrid` is defined as a structure with variable spacing in all 3 coordinate directions. I also think if we keep the return type as a `StructuredGrid` for cases where the `density` passed are equal, this would ensure that all future implementations of `vtkStructuredGrid` functionality in `PyVista` would be available to the returned data from this function (without having to `cast_to_structured_grid()`). 

For example, implementation of the "blanking" functionality of a `StructuredGrid` is a good idea, and should be a feature. I think other [CellGhostTypes](https://vtk.org/doc/nightly/html/classvtkDataSetAttributes.html#a9881ee5ce6e9cdb8ff94ce98b4815af4) methods of `vtkStructuredGrid` would be great to include in an additional PR like `EXTERIORCELL` and `HIDDENCELL`. These would be a great addition and help the function submitted in this PR.

MatthewFlamm: I think I have confused things a lot.  Sorry.  Let me start over.

`StructuredGrid` can be used for any arbitrary structured mesh including one with a cubical shape with equal number of cells in each direction, one with nonequal number of cells, one with nonuniform cell sizes in each direction, and also a mesh that is curved.  For example, a mesh that is curved along an axis as long as the mesh connectivity is structured.  The points can be in arbitrary positions, but the cell connectivity has a structured i-j-k relationship.

`ImageData` describes a mesh that is a rectangle and also that has a single cell length along a direction, but it can have different cell lengths in each direction.  For example x-length is a constant 1, y-length is a constant 2, z-length is a constant 3.

`RectilinearGrid` is like `ImageData`, but can have nonuniform lengths in each direction.  For example, x points at `(0, 1, 5, 10)` and y points at `(1, 2, 3, 6)`, etc.  Note that this can hold the same data as `ImageData`, but it just has constantly spaced x, y , z points.

`StructuredGrid` is not really needed for this function unless we want to take advantage of the 'blanking' functionality.  If we want to go this route in the future, we should set the groundwork here to return `StructuredGrid`, so it isn't a breaking change in the future.  Or at least think it through.

If we don't want `StructuredGrid`, we can use `ImageData` as of now since we are using a constant mesh size in each direction.  If we allow nonuniform (in a 1D sense) meshes in the future, we need to use `RectilinearGrid`.

pyvista/core/utilities/features.py
tkoyama010: @bsburnham As far as I can see, the changes regarding the voxelize function would be easier to understand if they were reverted in this PR. Note that the function we are trying to add is isolated to voxelize_volume.
By the way, have you checked the file changes at the far right of the tab? It is very useful to know what you are trying to change in this PR.
image

tkoyama010 (changes requested):

pyvista/core/utilities/features.py
bsburnham: No worries, I follow what you're saying now. 

I think having cells with a structured i-j-k relationship and connectivity would be beneficial. A `StructuredGrid` would be good to return here for this reason because the output from this could be used in further processing (e.g., fluid modelling) where other packages require the i-j-k connectivity. I also agree that this would allow future  `vtkStructuredGrid` functionality (e.g., "blanking" or "exteriorcell" etc) to be added without this function breaking. I am not a VTK expert, so I may not be best placed to add these. 

Returning a `StructuredGrid` for all density values, uniform or nonuniform, is cleaner code too.

MatthewFlamm: I think this is getting too complicated if you aren't going to implement the blanking functionality.  And to be clear, all three options give i-j-k cellular conncetivity.  `StructuredGrid` allows arbitrary hexahedron shapes for its cells, i.e. arbitrary point positions, while the other two require rectangular hexahedron shapes.   I'd recommend just returning `RectilinearGrid` in all cases.  A future PR could consider how we allow for nonuniform cell sizes in each direction for both `voxelize` and `voxelize_volume` to fully take advantage of this.

bsburnham: My apologies if I have complicated this. My understanding of grid types was not as thorough as it should be; thanks for the explanations and clarifications @MatthewFlamm. I'm happy to set the return type as a `RectilinearGrid` for `voxelize_volume()` in this PR. 

I agree, it would be good to explore nonuniform cell sizes in a future PR. Voxels have great utility that we should take full advantage of.

bsburnham:
I've changed `voxlize_volume()` to return a `RectilinearGrid` for all cases, as discussed. 

@tkoyama010 I have revereted the `voxelize()` code to the original. 

Please let me know if there is anything else that I should look into.

tkoyama010:
@bsburnham
I have added an image regression test for your method. If you are interested, please see "[Notes Regarding Image Regression Testing](https://docs.pyvista.org/version/dev/extras/developer_notes.html#notes-regarding-image-regression-testing)".
I have decided that the image output is an appropriate test for this method, but if you have a good assert test for checking values, etc., please suggest it to me.

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,121 @@ def voxelize(mesh, density=None, check_surface=True):
...
+    >>> vox.plot(scalars='MeshCells', show_edges=True, cpos=cpos)
+
+    Slice the unequal density voxel volume to view ``MeshCells``.
+
+    >>> slices = vox.slice_orthogonal()
+    >>> slices.plot(scalars='MeshCells', show_edges=True, cpos=cpos)
+
+    """
+    if not pyvista.is_pyvista_dataset(mesh):
+        mesh = wrap(mesh)
```
tkoyama010, line 170:
```suggestion
    if not pyvista.is_pyvista_dataset(mesh):
        mesh = wrap(mesh)
```
We don't need to check `is_pyvista_dataset` as it is checked in `wrap` function.
https://github.com/pyvista/pyvista/blob/35e0063ccbc8053a8edf9d8ac6d5fada339c2a90/pyvista/core/utilities/helpers.py#L102-L104

tkoyama010:
pre-commit.ci autofix

tkoyama010 (dismissed):
@bsburnham Thanks for your contribution! I will merge when other reviewers' requests are met. LGTM

pyvista/core/utilities/features.py
tkoyama010: resolved in #5133.

pyvista/core/utilities/features.py
tkoyama010: @MatthewFlamm @bsburnham Could we resolve this discussion and merge?

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,120 @@ def voxelize(mesh, density=None, check_surface=True):
...
+
+    # Create new element of grid where all cells _within_ mesh boundary are
+    # given new name 'MeshCells' and a discrete value of 1
+    voi['MeshCells'] = np.zeros(voi.n_cells)
+    voi['MeshCells'][cell_ids] = 1
+
+    # Create new element of grid where background cells
+    # given a new name 'BackgroundCells' and a discrete value of 0
+    voi['BackgroundCells'] = np.zeros(voi.n_cells)
+    voi['BackgroundCells'][0] = 0
```
MatthewFlamm, line 211: `'BackgroundCells'` is just 0 at every cell.  I recommend deleting it.  The needed information is all in `'MeshCells'`.

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,120 @@ def voxelize(mesh, density=None, check_surface=True):
...
+        Specify whether to check the surface for closure. If on, then the
+        algorithm first checks to see if the surface is closed and
+        manifold. If the surface is not closed and manifold, a runtime
+        error is raised.
+
+    Returns
+    -------
+    pyvista.RectilinearGrid
+        RectilinearGrid as voxelized volume with discretized cells.
+
```
MatthewFlamm, line 129: We should add a see also link to the voxelize.  Not sure if my suggestion has a whitespace error since GitHub is finicky about multiline suggestions.
```suggestion
        RectilinearGrid as voxelized volume with discretized cells.

    See Also
    --------
    pyvista.voxelize

```

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,120 @@ def voxelize(mesh, density=None, check_surface=True):
...
+    Notes
+    -----
+    Prior to version 0.39.0, this method improperly handled the order of
+    structured coordinates.
+
```
MatthewFlamm, line 134: This function didn't exist in version 0.39.0.
```suggestion
```

MatthewFlamm (commented):
Sorry for the delay in re-reviewing. These are likely my final comments.

pyvista/core/utilities/features.py
MatthewFlamm: Also I don't love the name of 'MeshCells', is there a better name?  I am terrible at naming though.

bsburnham: I've chosen `MeshCells` because they are the cells that selected within the grid. I did have `SelectedCells` as a first iteration, but thought `MeshCells` made more sense becauase they are cells contained within and intersect the input Mesh.

tkoyama010 (approved):

pyvista/core/utilities/features.py
MatthewFlamm: Where I get stuck is that the returned mesh already has cells.  I did a search for similar functions and found https://docs.pyvista.org/version/stable/api/core/_autosummary/pyvista.DataSetFilters.select_enclosed_points.html#pyvista.DataSetFilters.select_enclosed_points

And now that I look at these, I'm not sure what the essential difference is other than this proposed filter presupplies a RectilinearGrid object to find enclosed points/cells.  That is why aren't we using that filter to generate this result?  Is it a difference between points and cells?

That method uses `SelectedPoints` for the name.

bsburnham: This method is intended to return cells in a RectilinearGrid because that is what is commonly used in other software (e.g., fluid flow modelling in the subsurface). But also, and perhaps most importantly, this method generates a volume around the input mesh. I think having the ability to create a voxel volume that _encloses_ the mesh is a powerful feature to have built in. Extracting statistics of the interaction between the `SelectedCells` (if that is a more appropriate and agreed name) and the background cells has important utility. 

The filter you linked is used in this method to _select_ the cells contained within and that intersect the input mesh. This is the only way (at the moment) that I could discretise the voxel volume. As you mentioned before, using `GhostCells` is a good idea that would make this a fundamentally different method. 

I am happy to change the name to `SelectedCells` if that is what we think is clearer.

tkoyama010: @MatthewFlamm Are there any comments about this thread? If you don't have one, let's resolve this thread and merge PR.

MatthewFlamm: I still think `'BackgroundCells'` is unneeded. I don't love either `'MeshCells'` or `'SelectedCells'`.  But I'm also at a loss for a better name.  `'InsideMesh'`?

I'm only a bit concerned on the many seeming ways to do the same thing we are collecting.  This is in reference to understanding difference between this function and `select_enclosed_points`.  It isn't clear to me based on the examples and description how this function is different except that it provides a wrapper for creating a `RectilinearGrid`.  It is very likely these are different in some other way, but it doesn't come out in the description.

bsburnham: I'm happy to remove `'BackgroundCells'`. These data can be extracted from the resultant `RectilinearGrid` the function creates. 

Naming is indeed difficult, but I think `'InsideMesh'` sounds good. It is descriptive enough for what the function does.

The function `select_enclosed_points` returns an `UnstructuredGrid`, which isn't that helpful in further modelling that require at least a structured grid of some kind. A `RectilinearGrid` is sufficient, and as we discussed above, fine to use for this function. You're right, this function essentially provides a wrapper to create a `RectilinearGrid`. I feel this is ok though, especially for people who may not be code saavy but want to use 'PyVista' for their visualisation/analysis work.

If the reviewers are ok with this line of thinking, then I'll make the following changes:
- Remove `BackgroundCells`
- Change the name to `InsideMesh`

tkoyama010:
The release will take place on the 30th. If we want to include it in the release, we need to merge the PR into the main branch by then. Our activity is volunteer, so there is no need to force a release in time. Let's take it easy :)

We have tons of PRs requesting to merge, so we plan to do another new release in December.

bsburnham:
@tkoyama010 thanks for the update. 

If all reviewers (cc @banesullivan and @MatthewFlamm) are ok with the items discussed for this PR, then I'll make the following changes to complete this process:

- Remove `BackgroundCells`
- Change `SelectedCells` or `MeshCells` to something more appropriate like `InsideMesh`

I think what we've discussed in this PR has given some great food for thought for future PRs. (e.g., `HIDDENCELL` implementation).

tkoyama010:
> If all reviewers are ok with the items discussed for this PR, then I'll make the following changes to complete this process:

Could you first push the commit of the change so that other reviewers can easily image it? If there is a reject request, we can revert the commit.

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,114 @@ def voxelize(mesh, density=None, check_surface=True):
     return vox
 
 
+def voxelize_volume(mesh, density=None, check_surface=True):
+    """Voxelize mesh to create a RectilinearGrid voxel volume.
+
+    Creates a voxel volume that encloses the input mesh and discretizes the cells
+    within the volume that intersect and are contained within the input mesh.
+    All discretized cells are named ``MeshCells``. All other cells are named ``BackgroundCells``.
```
MatthewFlamm, line 107:
```suggestion
    ``InsideMesh``, an array in ``cell_data``, is ``1`` for cells inside and ``0`` outside.
```

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,114 @@ def voxelize(mesh, density=None, check_surface=True):
...
+
+    Returns
+    -------
+    pyvista.RectilinearGrid
+        RectilinearGrid as voxelized volume with discretized cells.
+
+    See Also
+    --------
+    pyvista.voxelize
+
```
MatthewFlamm, line 133:
```suggestion
    pyvista.DataSetFilters.select_enclosed_points

```

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,114 @@ def voxelize(mesh, density=None, check_surface=True):
...
+    >>> from pyvista import examples
+    >>> mesh = examples.download_cow()
+
+    Create an equal density voxel volume and plot the result.
+
+    >>> vox = pv.voxelize_volume(mesh, density=0.15)
+    >>> cpos = [(15, 3, 15), (0, 0, 0), (0, 0, 0)]
+    >>> vox.plot(scalars='MeshCells', show_edges=True, cpos=cpos)
+
+    Slice the voxel volume to view ``MeshCells``.
```
MatthewFlamm, line 152:
```suggestion
    Slice the voxel volume to view ``InsideMesh``.
```

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,114 @@ def voxelize(mesh, density=None, check_surface=True):
...
+
+    >>> slices = vox.slice_orthogonal()
+    >>> slices.plot(scalars='MeshCells', show_edges=True)
```
MatthewFlamm, line 155:
```suggestion
    >>> slices.plot(scalars='InsideMesh', show_edges=True)
```

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,114 @@ def voxelize(mesh, density=None, check_surface=True):
...
+
+    Create a voxel volume from unequal density dimensions and plot result.
+
+    >>> vox = pv.voxelize_volume(mesh, density=[0.15, 0.15, 0.5])
+    >>> vox.plot(scalars='MeshCells', show_edges=True, cpos=cpos)
```
MatthewFlamm, line 160:
```suggestion
    >>> vox.plot(scalars='InsideMesh', show_edges=True, cpos=cpos)
```

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,114 @@ def voxelize(mesh, density=None, check_surface=True):
...
+
+    Slice the unequal density voxel volume to view ``MeshCells``.
```
MatthewFlamm, line 162:
```suggestion
    Slice the unequal density voxel volume to view ``InsideMesh``.
```

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,114 @@ def voxelize(mesh, density=None, check_surface=True):
...
+
+    >>> slices = vox.slice_orthogonal()
+    >>> slices.plot(scalars='MeshCells', show_edges=True, cpos=cpos)
```
MatthewFlamm, line 165:
```suggestion
    >>> slices.plot(scalars='InsideMesh', show_edges=True, cpos=cpos)
```

tests/plotting/test_plotting.py
```diff
@@ -3789,3 +3789,17 @@ def test_no_empty_meshes():
...
+
+
+@pytest.mark.skipif(CI_WINDOWS, reason="Windows CI testing fatal exception: access violation")
+def test_voxelize_volume():
+    mesh = examples.download_cow()
+    cpos = [(15, 3, 15), (0, 0, 0), (0, 0, 0)]
+
+    # Create an equal density voxel volume and plot the result.
+    vox = pv.voxelize_volume(mesh, density=0.15)
+    vox.plot(scalars='MeshCells', show_edges=True, cpos=cpos)
```
MatthewFlamm, line 3801:
```suggestion
    vox.plot(scalars='InsideMesh', show_edges=True, cpos=cpos)
```

tests/plotting/test_plotting.py
```diff
@@ -3789,3 +3789,17 @@ def test_no_empty_meshes():
...
+
+    # Create a voxel volume from unequal density dimensions and plot result.
+    vox = pv.voxelize_volume(mesh, density=[0.15, 0.15, 0.5])
+    vox.plot(scalars='MeshCells', show_edges=True, cpos=cpos)
```
MatthewFlamm, line 3805:
```suggestion
    vox.plot(scalars='InsideMesh', show_edges=True, cpos=cpos)
```

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,114 @@ def voxelize(mesh, density=None, check_surface=True):
     return vox
 
 
+def voxelize_volume(mesh, density=None, check_surface=True):
+    """Voxelize mesh to create a RectilinearGrid voxel volume.
+
+    Creates a voxel volume that encloses the input mesh and discretizes the cells
+    within the volume that intersect and are contained within the input mesh.
```
MatthewFlamm, line 106: My understanding of this function is that it selects all cells if any one point is selected inside the mesh.  If so, the below would work, or consider wording it like the previous sentence.

```suggestion
    within the volume that intersect or are contained within the input mesh.
```

MatthewFlamm (commented):
Looking better to me.  This makes more sense to me now, although I think an example would be great for a future PR.  This function could be highlighted as a convenient wrapper, and then a more complex example could be given using `select_enclose_points` as the more flexible method.

bsburnham:
Thanks for catching the `MeshCells` in the examples that I forgot to change @MatthewFlamm! 

A galltery example of this function would be good - nice idea.

pyvista/core/utilities/features.py
```diff
@@ -99,6 +99,115 @@ def voxelize(mesh, density=None, check_surface=True):
...
+    Load file from PyVista examples.
+
+    >>> from pyvista import examples
+    >>> mesh = examples.download_cow()
+
+    Create an equal density voxel volume and plot the result.
+
+    >>> vox = pv.voxelize_volume(mesh, density=0.15)
+    >>> cpos = [(15, 3, 15), (0, 0, 0), (0, 0, 0)]
+    >>> vox.plot(scalars='MeshCells', show_edges=True, cpos=cpos)
```
MatthewFlamm, line 151:
```suggestion
    >>> vox.plot(scalars='InsideMesh', show_edges=True, cpos=cpos)
```

tkoyama010 (approved):

tkoyama010:
@banesullivan This is a friendly reminder. Could you approve this Pull Request to merge?

banesullivan (approved):
Awesome work!! Generally, this all looks good to me and I'm happy to sneak it in right before the 0.43 release

Issue #4977, opened 2023-09-29
Move input validation functions to a single location


user27182:
### Requested maintenance

- Move, consolidate, refactor, etc. input validation functions to a single location, e.g. `core.utilities.input_validation.py`
- Update the "Contributing to PyVista" section of the docs to explicitly highlight that this is a thing that exists and recommend making use of these validation functions

### Rationale
PyVista's methods have a consistent need to validate inputs, e.g. 

- a single point/vector with shape (3,)
- point/vector arrays with shape (N,3)
- transformation matrix with shape (4,4)
 
Points, vectors, and transformations are used everywhere. However, I have noticed that many methods tend to re-write the same or similar validation routines locally, and while these checks are often sufficient, they tend to:

- have some inconsistency with the type hinting, e.g. `np.ndarray` vs `sequence[float]`
- create an overall inconsistent API across the code base (e.g. one function may only accept a 4x4 `np.ndarray` for a transformation, whereas another might also accept `vtkMatrix4x4`)
- tend not to capture edge cases, e.g.:
    -  do `if x.shape[1] == 3` without first checking `if isintance(x, np.ndarray)` for the `shape` property or that `ndim > 0` which is needed to catch single `int` or `float` input cases
    - assume an input `Sequence` is numeric, when `"abc"` is also a valid `Sequence` and would make it past validation
    - do checks like `len(x) == 3` on an `np.ndarray` or `Sequence` to validate an input has 3 elements (x,y,z), even though a 3x3 `np.ndarray` also has `len(x) == 3` (the check should instead do `x.size == 3` for `np.ndarray`)

Moreover, this tends to result in the constant need to also write `def test_foo_raises()` test cases; can result in unexpected bugs for the edge cases listed above; reduced code coverage; increase maintenance, etc. It is also tedious/annoying to do for development, as developers would rather spend time implementing new features than writing input validation code.

[...]

Issue #5074, opened 2023-10-13
Type hinting `NumericArray`, `Matrix3x3Like` and `Matrix4x4Like`


tkoyama010:
### Describe what maintenance you would like added.

1. Is the type hinting for `NumericArray` correct? Currently, as it's defined doing `Sequence[NumericArray]` implies that mixing and matching `Tuples` with `NDArray` is allowed, something like `((1,2,3), np.array([4,5,6]))`. I don't think that's generally expected or handled though... Perhaps it should be something like this instead: 
```python 
NumericArray = Union[Sequence[Number], Sequence[Sequence[Number]], np.ndarray]
```
2. Also, perhaps there should be a `Matrix3x3Like` type and `Matrix4x4Like` type as well, something like:
``` python
NumericArray = Union[Sequence[Number], Sequence[Sequence[Number]], np.ndarray]
Matrix3x3Like = Union[VectorArray, vtkMatrix3x3] 
Matrix4x4Like = Union[NumericArray, vtkMatrix4x4, vtkTransform] 
TransformLike = Union[Matrix3x3Like, Matrix4x4Like]
```

_Originally posted by @user27182 in https://github.com/pyvista/pyvista/pull/5052#discussion_r1355901047_

darikg:
> Currently, as it's defined doing Sequence[NumericArray] implies that mixing and matching Tuples with NDArray is allowed, something like ((1,2,3), np.array([4,5,6])). I don't think that's generally expected or handled though... 

Seems like that would be fine if it's just going to be passed into `np.asarray` -- `np.asarray(((1,2,3), np.array([4,5,6])))` correctly returns a 2x3 array

user27182:
I'd like to clarify what the issue is. Currently, the following type definition:
https://github.com/pyvista/pyvista/blob/3f8b06f0ee7b05dacf177a10f6219b77e57ac8ff/pyvista/core/_typing_core.py#L11
is missing the `Sequence[Sequence[Number]]` case, i.e. it fails to capture 2D arrays created using nested sequences. I would consider  ``[[1,2],[3,4]]`` to be a numeric array, but it's not a `NumericArray` based on the current definition.

One possible solution proposed by @tkoyama010 [here](https://github.com/pyvista/pyvista/pull/5052#discussion_r1354066563) is to use `Sequence[NumericArray]` to capture this missing case. This is what my comment (reposted above) is responding to.

[...]

PR #4965, opened 2023-09-27
Add 3D SolidSphere and SolidSphereGeneric as geometric objects

MatthewFlamm:
### Overview

closes #4943 .  This PR adds in a geometric object that returns a filled, 3D sphere in an `UnstructuredGrid`.  

It has two kinds of API.  1) similar to `Sphere`, e.g. `SolidSphere(outer_radius=0.7, theta_resolution=15)`.  2) specifying arbitrary point locations in spherical coordinates, e.g. `SolidSphereGeneric(radius=np.geomspace(0.01, 10, 15))`

### Details

TODO:

- [ ] Align on naming.  The naming currently follows `Cylinder` and `CylinderStructured` although I'm not sure it is as descriptive as `FilledSphere` or similar.  For example, we could convert `pyvista.Sphere` to an `UnstructuredGrid` and it would still be 2D.
- [x] Switch phi and theta definition to be consistent with `Sphere`.
- [x] Enable starting and ending points that differ from defaults.
- [x] Add test for starting and ending points.
- [x] Check edge cases when only 2 or 1  values are supplied for each parameter type
- [x] Add more examples
- [x] Add translate and orient
- [x] Check that orientation of this object matches `Sphere`
- [x] Fix description of phi and theta and use in `Sphere`
- [x] Split into two functions, one for simple usage, one for flexible usage.
- [x] Radians flag for phi/theta

PR #5100, opened 2023-10-17
Add `input_validation` package

user27182:
### Overview

This PR is related to #4977  and adds a new `pyvista.core.input_validation` package. The package can be used to check, or validate many common input types and array shapes used by PyVista. Use of this package removes the need to do manual input validation and write associated test cases 🥳.

### TODO
- [x] settle on function naming conventions. Currently, convention is:
    - Two main functions categories: `validate_` and `check_`
    - Do not use `coerce_`; instead, use an explicit term such as `reshape`
        -  e.g. `validate_arrayNx3(arr, reshape=True)` will allow `(3,)` or `(N,3)` input if `reshape=True`
- [x] finalize API and add a few more helper functions
- [x] add some missing test cases / improve coverage
- [x] add docstrings and examples
- [x] resolve numerous numpydoc and mypy precommit errors
- [x] remove Numpy typing `ArrayLike`, `NDArray`, `DTypeLike` (typing requires version bump to 1.20)
- [x] evaluate performance with `timeit` 
- [x] update 'contributing ' docs

### Details

A key contribution is the following function:
``` python
ef validate_array(
    arr,
    /,
    *,
    must_have_shape=None,
    must_have_dtype=None,
    must_have_length=None,
    must_have_min_length=None,
    must_have_max_length=None,
    must_be_nonnegative=False,
    must_be_finite=False,
    must_be_real=True,
    must_be_integer_like=False,
    must_be_sorted=False,
    must_be_in_range=None,
    strict_lower_bound=False,
    strict_upper_bound=False,
    reshape=None,
    broadcast_to=None,
    dtype_out=None,
    as_any=True,
    copy=False,
    to_list=False,
    to_tuple=False,
    name="Array",
) 
```
Which is basically a swiss-army knife validation function for arrays... one function to rule them all!

The function works fundamentally by casting everything as a numpy array and then doing operations or checks from there. Any constraint can be added using keywords (kwargs with ``must_`` prefix), and an error will be raised if the input doesn't meet the requirements. Many common numpy functions are consolidated and integrated as well for convenience, such as `dtype_out`, `copy`, `as_any`, `reshape`, and `broadcast_to`.

### Example: Validate Nx3 arrays
Replicate the functionality of [`_coerce_pointslike_arg`](https://github.com/pyvista/pyvista/blob/5492c71c6423c534dc133a62b4d8e4e3b3d769b8/pyvista/core/utilities/arrays.py#L58C1-L61C50), and do all the checks and reshaping needed to validate points-like or vector-like Nx3 arrays with one line:
```python
validate_array(arr, must_have_shape=[3, (-1, 3)], reshape=(-1, 3))
```
 Since this is a common use case, it is "wrapped" as a convenience function
```python
validate_arrayNx3(arr)
```
### Example: Validate a data range
Validate a data range, which must be numeric and have 2 sorted elements. Format the output as tuple.
```python
validate_array(rng, must_have_shape=2, must_be_sorted=True, to_tuple=True)
```
This function is also wrapped for convenience as:
```python
validate_data_range(rng)
```
### Example: Validate point or cell IDs
Validate arrays representing IDs. These inputs may be integer-like (but have type `float`), can be scalar (`shape=()`), or 1-dimentional of any length (`shape=(-1)`), and must be non-negative. The output data type is typically required to be `int` for further processing. All checks and formatting can be done with one line:
```python
validate_array(
    ids,
    must_have_shape=[(),(-1)],
    must_be_integer_like=True,
    must_be_nonnegative=True,
    dtype_out=int, 
    reshape=(-1,)
)
```
This function is also wrapped for convenience as:
```python
validate_uintlike_arrayN(ids)
```

### Example: Validate 1D 3-element arrays
Replicate the functionality of [`check_valid_vector`](https://github.com/pyvista/pyvista/blob/5492c71c6423c534dc133a62b4d8e4e3b3d769b8/pyvista/core/utilities/misc.py#L47) with one line:
```python
validate_array(arr, must_have_shape=[3, (1, 3)], reshape=(3,))
```
 Since this is a common use case, it is "wrapped" as a convenience function. The option to broadcast a scalar to shape (3,) is also added, which is useful for working with overloaded VTK functions that accept scalars or 3-element arrays.
```python
validate_array3(arr, broadcast=True)
```

### Package functions
Many standalone `check_` or `validate_` functions are added as part of the package. Many of the functions are used directly under-the-hood by `validate_array`, but they may also be used generally. E.g. the function `check_string_is_in_iterable`, can be used to validate user input used to specify a method from a list of acceptable methods.


user27182:
Looking for some feedback:
- naming: more explicit? e.g. `def must_be_sorted()` -> `def must_be_sorted_array()`
- naming: how to express shape in function name? e.g. is `validate_arrayNx3` clear?
- typing: Is there a desire to rework the types in `core._typing_core` (e.g. #5074) to make use of them explicitly with these validation functions?

darikg:
Love the big picture. #5102 would be caught by the validation here right?

I have some performance suspicions related to unnecessary allocations -- e.g. `check_is_ArrayLike` calls `cast_array_to_NDArray` (discarding its output), which in turn calls `numpy.asanyarray`, which, if the original input is e.g. `list[list[float]]` would allocate the array needlessly. Multiple unnecessary allocations can occur in `validate_numeric_array` since ` check_is_subdtype`, `check_is_real`, etc. can each allocate in turn and then discard the temporary array. `coerce_shapelike_as_shape` calls `np.zeros(shape=shape_arr)` to ensure a shape is valid, which seems wildly unnecessary to me. I suppose ultimately we could do performance testing to see how much this stuff matters. 

A much more minor quibble, but it looks like the check methods consistently call `check_is_string(name, name="Name")` -- the name variable would only be used internally, not by the end-user, correct? Seems like overkill to me.

Regarding naming, I like the `check_` and `validate_` separation, but even after reading the naming conventions section I'm still a little confused by the difference between `coerce_` and `_validate`. I guess the important distinction is the shape possibly changing in e.g. a 3x3 transform converting to 4x4, but I wonder if it's not just as valid to call that a validation as well.

user27182:
> #5102 would be caught by the validation here right?

Quite possibly, yes. I do see lots of shape and type checking done at various points in the pipeline in `dataset.py` and `datasetattributes.py` when doing something like `mesh["data"] = data`, so there is plenty of opportunity to make use of `validate_numeric_array` or related methods to do more extensive checking / validation to ensure we have the exact input shape(s) and/or dtype needed. Any errors raised could be cast as a warning if needed.

> I have some performance suspicions related to unnecessary allocations e.g. `check_is_ArrayLike` calls `cast_array_to_NDArray` (discarding its output), which in turn calls `numpy.asanyarray`, which, if the original input is e.g. `list[list[float]]` would allocate the array needlessly.

Some comments:
- `check_is_ArrayLike` is more of a primitive convenience method to perform this check if you want to in case you don't actually want to check anything else. It's not used much in this package, and should not be used if you actually want to do other checks or have an `np.ndarray` output. This could be made clear in the docstring or this function could be made internal to discourage its use/
- In general the more typical use case is to use `cast_array_to_NDArray` directly, which will only do the allocation once, and you can grab the output for futher processing without needing to cast multiple time unnecessarily.
- If your input was `list[list[float]]`, and you wanted `list[list[float]]` for an output, then yes, this allocation would be unnecessary. However, I don't think this is a big deal because use cases where we actually want `list` outputs typically involve small dimensionality, e.g. a sequence of 3 values. Even for cases with large dimensionality (e.g. a large 1D list of point IDs), I think the status quo approach to validating that the inputs have the correct type (e.g. integers) usually involve casting the input as `np.ndarray` anyway in order to do these checks using `dtype` properties or using `==` syntax. So, I don't think it's any more wasteful of an allocation than existing validation methods currently employed in PyVista which also do this kind of intermediate allocation.
- I think it would be good to a do a `timeit` comparisons, e.g. between `_coerce_pointslike_args` and its analogue through `validate_numeric_array` to get an idea of the true performance hit by using this function.

> Multiple unnecessary allocations can occur in `validate_numeric_array` since ` check_is_subdtype`, `check_is_real`, etc. can each allocate in turn and then discard the temporary array.

-  Yes, `check_is_subdtype`, `check_is_real`, etc. make a call to `cast_array_to_NDArray` internally. However, `cast_array_to_NDArray` is called once already at the start of `validate_numerica_array`, and therefore the array input to `check_is_subdtype` at the time these are called is already of type `np.ndarray`. 
- The call to `cast_array_to_NDArray` inside of `check_is_subdtype` is made with default arguments where `copy=False`. My understanding is that the allocation to a new array by NumPy is only made if the input is not already an `np.ndarray` (e.g. the `np.asarray` documention states "No copy is performed if the input is already an ndarray with matching dtype and order"). Therefore, I believe that multiple calls to  `cast_array_to_NDArray`  will _not_ result in multiple unnecessary allocations. E.g. I _think_ the following will only allocate memory once:
```python
arr = np.asarray([1,2,3])
for i in range(100):
    arr = np.asarray(arr)
```

> `coerce_shapelike_as_shape` calls `np.zeros(shape=shape_arr)` to ensure a shape is valid, which seems wildly unnecessary to me. I suppose ultimately we could do performance testing to see how much this stuff matters.

- This is a good point and yes I agree this isn't the most performative way of doing this check. This was done to allow for generally accepting any `ShapeLike` input, i.e. allow any shape that NumPy allows. Similar to `ArrayLike`, however, I couldn't find a clear definition of what `ShapeLike` means for NumPy, i.e. couldn't determine exactly what type of input NumPy was willing to accept as a valid `shape` input (e.g. when calling `np.zeros(shape=shape_arr)`, which cases of `shape_arr` would fail?). Further complicating this is the fact the `(-1)` syntax for shape isn't typically accepted by numpy as a valid shape (but is valid when calling `np.reshape(shape=shape)`).
- Some options to address this: (1) simply use `np.empty` instead of `np.zeros`, or (2) use a stricter definition of accepted input types for `shape`, e.g. only allow `tuple[int,...]` or `int` inputs and reject any other input types (even if NumPy would otherwise allow it), thereby eliminating the need to validate through NumPy.

> A much more minor quibble, but it looks like the check methods consistently call `check_is_string(name, name="Name")` -- the name variable would only be used internally, not by the end-user, correct? Seems like overkill to me.

The `name` parameter is exposed to the user in `validate_numeric_array`, so the parameter is external in this sense but then yes passed internally to other functions like `check_is_subdtype`, etc. Similarly, `check_is_string` is exposed as its own external method (with a `name` parameter) in case a user would like to validate some other non-array string input. So, both functions as they are designed have public APIs and therefore I think both inputs could be called with bad input for `name` (and therefore should checked it first).

In fact, I didn't initially have the `check_is_string` method called by _every_ method (which, yes, is redundant in most cases) for the reason you stated. But then during testing I got a weird bug at some point, and it turns out it's because I had accidentally passed an incorrect argument to one of the `check` methods and passed a non-string as an arg (not a kwarg) which Python happily interpreted as a positional argument for the `name` parameter, even though I intended it to be an arg, not a kwarg. After that, I made the decision that even savvy users can still pass bad input to these methods and introduce bugs, so despite the performance hit felt it was better to be safe than sorry.

Some possible solutions:

- Do nothing and wait until performance checks are done to determine that this is actually a problem
- Change the function signature so that `name` is _only_ accepted as a kwarg, i.e. change it to `check_is_string(some_string,*,name="Name")` with the asterix `*` added.
- Relegate any low-level `check` methods to internal methods only (e.g. as `_check`) and skip any redundant validations
- Add an `unsafe` parameter to enable or disable certain (possible redundant) checks

> Regarding naming, I like the `check_` and `validate_` separation, but even after reading the naming conventions section I'm still a little confused by the difference between `coerce_` and `_validate`. I guess the important distinction is the shape possibly changing in e.g. a 3x3 transform converting to 4x4, but I wonder if it's not just as valid to call that a validation as well.

Yes the main distinction is change of shape, but I agree that it's not very clear and these could /should be considered the same thing. To your point, I gave the following two functions `coerce` names:
```
coerce_transformlike_as_array3x3
 coerce_transformlike_as_array4x4
```
But, technically, using the existing naming convention,`coerce_transformlike_as_array3x3` should actually be named `validate_transformlike_as_array3x3` since we have 3x3 in and 3x3 out... This illustrates that even I am not following the convention, so it's easy to miss.

If we did follow convention, however, and named it as `validate_transformlike_as_array3x3` instead, this would result in two very similar functions having different names, since ` coerce_transformlike_as_array4x4` would remain as `coerce` since it may change the input's shape. Having very similar methods with different names isn't ideal, however...

I do believe the distinction is important in some cases, though, since it's good to know whether any reshaping will be done if desired (and not as a bug like #5102).

Since the key difference is shape, perhaps we should do the following:
- stop using `coerce`, only use `validate` and `check`
- use the term `reshape` explicitly where `np.reshape` is used
- for transformations, e.g. 3x3 being coerced to 4x4, we can keep using `validate` since `np.reshape` isn't the source of the change in shape, and the function name has `_as_array4x4` which I think makes the output shape explicit
e.g. The name for something like `_coerce_pointslike_arg` could be :
1. `validate_and_reshape_to_arrayNx3`
Or, since validation may be implied by the fact it's in the `input_validation` namespace, perhaps just call it:
2. `reshape_as_arrayNx3`
In case (2) perhaps the term `validate` should be dropped altogether for all package functions? Depends on how explicit we want to be with the function names.

We could also add a new `shape_out` param as as `tuple` and call `np.reshape(shape=shape_out)` inside of `valindate_numeric_array` to do the reshaping automatically (and do error handling as appropriate). `shape` param would be renamed to `shape_in` as well to make that more explicit also.

In this case, two package functions for validating Nx3 arrays, one of which allows coercion, would look like:

```python
# function allows 1D 3-element vector or 2D Nx3 array
# name uses `reshape` instead of `coerce`
def reshape_as_arrayNx3(arr):  
    validate_numeric_array(arr, shape_in=[3, (-1, 3)], shape_out=(-1,3))  

# same function as above but does not reshape and strictly enforces Nx3 shape
# name does not include the term `validate`(implied)
def as_arrayNx3(arr): 
    validate_numeric_array(arr, shape=(-1, 3))

**Comment by MatthewFlamm** [comment]
This is quite nice and looks very in depth!  I unfortunately only have time for straightforward reviews at the moment, so I won't be able to take a deep dive on it for awhile.  My first thought is that this looks useful to more than just for PyVista and could be considered as a standalone package.

**Comment by MatthewFlamm** [comment]
There is this package, but it is only for typing: https://github.com/ramonhagenaars/nptyping/tree/master

**Comment by tkoyama010** [comment]
Let's try embedding it in PyVista first, and then isolate the repository once it has been refined.

**Comment by user27182** [comment]
There are still some outstanding items on the TODO list but I think the main workings of the package are complete and ready for review. I have updated the top comment, as well as my earlier response to @darikg to address some of the early concerns with the code.

If everything looks good I'll work on finishing up the TODO list.

File: pyvista/core/input_validation/validate.py [hunk 1]
```diff
@@ -0,0 +1,730 @@
...
+
+    must_be_nonnegative : bool, default: False
+        Check that all elements of the array are nonnegative.
+
+    must_be_finite : bool, default: False
+        Check that all elements of the array are finite, i.e. not
+        infinity and not Not a Number (NaN).
+
+    must_be_real : bool, default: True
+        Check that the array's has real numbers, i.e. its data type is
```
tkoyama010, line 117:
```suggestion
        Check that the arrays has real numbers, i.e. its data type is
```

tkoyama010:
@user27182 Great job! It would be more useful to add the content of the description as API documentation. Do you want to address this in this PR or should I set up another PR later?

user27182:
> @user27182 Great job! It would be more useful to add the content of the description as API documentation. Do you want to address this in this PR or should I set up another PR later?

I was thinking that uses of existing functions `_coerce_pointslike_arg` or `check_valid_vector` in PyVista would be replaced by a separate PR. But otherwise the examples could be made into a more formal document.... do you want a standalone `Input Validation` Example file in the docs? Or where/how do you would want to include this documentation?

tkoyama010:
> > @user27182 Great job! It would be more useful to add the content of the description as API documentation. Do you want to address this in this PR or should I set up another PR later?
> 
> I was thinking that uses of existing functions `_coerce_pointslike_arg` or `check_valid_vector` in PyVista would be replaced by a separate PR. But otherwise the examples could be made into a more formal document.... do you want a standalone `Input Validation` Example file in the docs? Or where/how do you would want to include this documentation?

Creating a standalone example is overkill in my mind. Creating API documentation from docstring is sufficient in this case. Just add the `rst` file under the api directory and sphinx will automatically create the documentation from docstring.
https://github.com/pyvista/pyvista/tree/main/doc/source/api/core

user27182:
There are some definite performance concerns, e.g. `validate_arrayNx3` is approx 40x slower for a shape (3,) input than `_coerce_pointslike_arg`:
``` python
import timeit
from pyvista.core.input_validation.validate import validate_arrayNx3
from pyvista.core.utilities.arrays import _coerce_pointslike_arg
import numpy as np
print(timeit.timeit(lambda: validate_arrayNx3(np.random.random((3,))), number=10000))
print(timeit.timeit(lambda: _coerce_pointslike_arg(np.random.random((3,))), number=10000))
0.8408326003700495
0.022614598274230957
```

The effect is less noticeable for larger array inputs, approx 2x slower.
``` python
import timeit
from pyvista.core.input_validation.validate import validate_arrayNx3
from pyvista.core.utilities.arrays import _coerce_pointslike_arg
import numpy as np
print(timeit.timeit(lambda: validate_arrayNx3(np.random.random((10000,3))), number=10000))
print(timeit.timeit(lambda: _coerce_pointslike_arg(np.random.random((10000,3))), number=10000))
3.2129379007965326
1.6613970007747412
```

To be fair the new validation methods are meant to be very general and are doing additional checks / validations which are not done otherwise, which  will inherently penalize performance.

I'll look into what the potential bottleneck is (e.g. calls to `isinstance`?) and perhaps add skips / shortcuts where possible to avoid any redundant/unnecessary checks.

darikg:
> I'll look into what the potential bottleneck is (e.g. calls to `isinstance`?) and perhaps add skips / shortcuts where possible to avoid any redundant/unnecessary checks.

Thanks for checking this! If asanyarray turns out to be the bottleneck I was thinking about a wrapper class that could be passed around internally, like this:

```python
from dataclasses import dataclass
from functools import cached_property

@dataclass
class _ToBeChecked:
    orig: list | ndarray
    
    @cached_property
    def array:
        return np.asanyarray(self.orig)

    @property
    def dtype:
        return self.array.dtype
```

user27182:
> Thanks for checking this! If asanyarray turns out to be the bottleneck I was thinking about a wrapper class that could be passed around internally, like this:
> 
> ```python
> from dataclasses import dataclass
> from functools import cached_property
> 
> @dataclass
> class _ToBeChecked:
>     orig: list | ndarray
>     
>     @cached_property
>     def array:
>         return np.asanyarray(self.orig)
> 
>     @property
>     def dtype:
>         return self.array.dtype
> ```

That's a great idea! Maybe `name` could be stored there too and passed around as well. There is lots of opportunity to use OOP here... when designing the structure of the package I wasn't sure how simple and/or independent to make each function and whether to link the separate functions together somehow.

user27182:
Without any major design changes, for now I was able to find and fix a few key bottlenecks.
I ran a `cProfile` test, and found that `cast_to_ndarray` (which calls `np.asanyarray`) was one of the bottlenecks, but was relatively minor compared to some apparently very inefficient code in `_validate_shape_value`. 

I added a short-circuit return condition to `cast_to_ndarray` to exit early without doing anything in the most common default case. This cut the `timeit` time for `validate_arrayNx3` in half. But the bigger change came from mostly reworking `_validate_shape_value`.

New times from `timeit` show the validation function  is now only 3-4 times slower (down from 40).

``` python
from pyvista.core.input_validation.validate import validate_arrayNx3
from pyvista.core.utilities.arrays import _coerce_pointslike_arg
import numpy as np
print(timeit.timeit(lambda: validate_arrayNx3(np.random.random((3,))), number=10000))
print(timeit.timeit(lambda: _coerce_pointslike_arg(np.random.random((3,))), number=10000))
0.07509090006351471
0.02376769855618477

```

and for the larger array case, it's now about 7% slower (down from 100%):

``` python
import timeit
from pyvista.core.input_validation.validate import validate_arrayNx3
from pyvista.core.utilities.arrays import _coerce_pointslike_arg
import numpy as np
print(timeit.timeit(lambda: validate_arrayNx3(np.random.random((10000,3))), number=10000))
print(timeit.timeit(lambda: _coerce_pointslike_arg(np.random.random((10000,3))), number=10000))
1.7628600988537073
1.655470598489046
```

CONTRIBUTING.rst
```diff
@@ -411,6 +411,55 @@ encouraged to use the ``.. versionadded`` sphinx directive. For example:
         """
 
 
+Input Validation
+^^^^^^^^^^^^^^^^
```
tkoyama010, line 415: +100 for adding this. Thanks!

user27182:
I made a few more optimizations. Turns out calls to `np.issubdtype` are expensive, and so should generally be avoided.

Final `timeit` numbers.
`validate_arrayNx3` is now approx 80% slower than `_coerce_pointslike_arg` (down from 300-400%).
``` python
print(timeit.timeit(lambda: validate_arrayNx3(np.random.random((3,))), number=10000))
print(timeit.timeit(lambda: _coerce_pointslike_arg(np.random.random((3,))), number=10000))
0.04088430106639862
0.02282389998435974
```

And with the larger array case, which is now less than 2% slower.
``` python
print(timeit.timeit(lambda: validate_arrayNx3(np.random.random((10000,3))), number=10000))
print(timeit.timeit(lambda: _coerce_pointslike_arg(np.random.random((10000,3))), number=10000))
1.692161701619625
1.6698403991758823
```

user27182:
I think this level of optimization is probably enough for now... for completeness here's the report from `cProfile` showing the performance bottlenecks:
``` python
import cProfile
 from pyvista.core.input_validation.validate import validate_arrayNx3 
def f(): 
    for _ in range(10000): 
        validate_arrayNx3(np.random.random((10000,3))) 
cProfile.run("f()",sort='tottime')
```
```
         280004 function calls in 1.779 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    1.657    0.000    1.657    0.000 {method 'random' of 'numpy.random.mtrand.RandomState' objects}
    10000    0.018    0.000    0.088    0.000 validate.py:36(validate_array)
    10000    0.017    0.000    0.051    0.000 check.py:366(check_has_shape)
    10000    0.014    0.000    0.112    0.000 validate.py:476(validate_arrayNx3)
    20000    0.011    0.000    0.021    0.000 check.py:398(_shape_is_allowed)
        1    0.010    0.010    1.779    1.779 test_cProfile.py:126(f)
    20000    0.009    0.000    0.012    0.000 check.py:825(_validate_shape_value)
    20000    0.009    0.000    0.011    0.000 validate.py:723(_set_default_kwarg_mandatory)
    10000    0.008    0.000    0.009    0.000 check.py:109(check_is_real)
    10000    0.004    0.000    0.007    0.000 {built-in method builtins.all}
    10000    0.004    0.000    0.005    0.000 arrays.py:834(cast_to_ndarray)
```

tkoyama010 (changes requested):
Quick review of documentation. nits

CONTRIBUTING.rst
```diff
@@ -411,6 +411,55 @@ encouraged to use the ``.. versionadded`` sphinx directive. For example:
...
+        # An error is automatically raised for bad input.
+        arr = valid.validate_arrayNx3(points_array)
+
+        # Validate input method. An error is automatically
+        # raised if the method is not valid.
+        possible_methods = ["method_a", "method_b"]
+        check_is_string_in_list(method, possible_methods)
+
+When using the array validation methods, consider setting the input
+type in the docstring to ``array_like`` (if appropriate), as the validation
```
darikg, line 459: is `array_like` defined somewhere, or it's just a convention?

user27182: It's not formally defined as a type, no. The term `array_like` is just a [NumPy glossary term](https://numpy.org/devdocs/glossary.html#term-array_like). In NumPy 1.20 a proper [ArrayLike type](https://numpy.org/devdocs/reference/typing.html) was defined, however. I did initially make used of `ArrayLike` typing but removed it since (1) it doesn't (yet) support hinting array shape or dtype, and (2) it requires a bump to NumPy 1.20.

I'll note the term `array_like` is still used in the NumPy docs, and is [already in use in PyVista](https://github.com/search?q=org%3Apyvista%20array_like&type=code)

Do you think we should include a link to the glossary definition for reference?

darikg: I like having a link somewhere, especially if it's only going to get more common in the pyvista codebase

user27182:
I haven't been able to resolve this documentation build error...  not sure how to view these files online but when I build the docs locally the rst file ``validate_arrayNx3.rst`` only has 6 lines in it, so not sure what's going on with lines 7 and 18. The docstrings appear to have correct indentation. Any help resolving this is appreciated.

```
/home/runner/work/pyvista/pyvista/doc/source/api/core/_autosummary/
pyvista.core.input_validation.validate.validate_arrayNx3.rst:7: ERROR: Unexpected indentation.

/home/runner/work/pyvista/pyvista/doc/source/api/core/_autosummary/
pyvista.core.input_validation.validate.validate_arrayNx3.rst:18: ERROR: Unexpected indentation.
```

CONTRIBUTING.rst
user27182: The additional info and link has been aded in [7c8a237](https://github.com/pyvista/pyvista/pull/5100/commits/7c8a237eb46dfa18e7953ced6ee05d11ae5a86d7)

banesullivan:
This is awesome!

user27182:
The docs LGTM, this is ready for review.

user27182:
I think the naming convention used is pretty good as it is, but there are few minor inconsistencies, e.g.

-  `reshape` param of `validate_array`, should it be renamed to `reshape_to` to be more explicitly and consistent with `broadcast_to` ? Or maybe rename `broadcast_to` to `broadcast` instead?
- `dtype_out` param of `validate_array`, should it just be `dtype`?
- there is `check_is_integerlike`, but in `validate_array` the relevant param is called `must_be_integer_like` which has an extra underscore between `integer` and `like`. There are possibly other similar "as one word" vs "as two words" cases, eg. dtype vs dtype-like vs dtype_like.

darikg:
My prefs:
- reshape_to and broadcast_to are slightly clearer. It's a fairly verbose api (in a good way), and these fit in well
- I like dtype_out, it makes it clearer that the function is willing to change the dtype if necessary, as opposed to just checking and throwing an error on the wrong dtype
- both integerlike and integer_like seem slightly off, my brain really wants a hyphen there. intlike doesn't seem good either. total toincoss

user27182:
I had a look at the input validation in #4965, and noticed that the need for "monotonically increasing" sorted values require the array be "strictly" increasing. However, `check_is_sorted` did not support the "strict" option. Commit f7d8de4 addresses this and generalizes the sorting check to add a `strict` param (as well as `axis`param and option for descending, too).

pyvista/core/input_validation/validate.py
```diff
@@ -0,0 +1,1022 @@
...
+    Examples
+    --------
+    Validate an axes array.
+
+    >>> import numpy as np
+    >>> import pyvista.core.input_validation as valid
+    >>> valid.validate_axes(np.eye(3))
+    array([[1., 0., 0.],
+           [0., 1., 0.],
+           [0., 0., 1.]])
```
tkoyama010, line 380:
```suggestion
    array([[ 1., 0., 0.],
           [ 0., 1., 0.],
           [ 0., 0., 1.]])
```

pyvista/core/input_validation/validate.py
```diff
@@ -0,0 +1,1022 @@
...
+
+    Validate individual axes vectors as a 3x3 array.
+
+    >>> valid.validate_axes([1, 0, 0], [0, 1, 0], [0, 0, 1])
+    array([[1., 0., 0.],
+           [0., 1., 0.],
+           [0., 0., 1.]])
```
tkoyama010, line 387:
```suggestion
    array([[ 1., 0., 0.],
           [ 0., 1., 0.],
           [ 0., 0., 1.]])
```

pyvista/core/input_validation/validate.py
```diff
@@ -0,0 +1,1022 @@
...
+
+    Create a validated left-handed axes array from two vectors.
+
+    >>> valid.validate_axes(
+    ...     [1, 0, 0], [0, 1, 0], must_have_orientation='left'
+    ... )
+    array([[1., 0., 0.],
+           [0., 1., 0.],
+           [0., 0., -1.]])
```
tkoyama010, line 396:
```suggestion
    array([[ 1., 0., 0.],
           [ 0., 1., 0.],
           [ 0., 0., -1.]])
```

tkoyama010 (approved):
LGTM. let's merge this and use how it works.

user27182:
This should probably be updated to make use of updated typing now that #5171 has been merged.

tkoyama010:
> This should probably be updated to make use of updated typing now that #5171 has been merged.

Would you like to be added to this PR or would you like to be added to another PR later?

user27182:
> @user27182 Would you like to be added to this PR or would you like to be added to another PR later?

We merge this as-is if you like and update the typing as another PR later.

user27182:
@tkoyama010 After some work on #5226, I believe this package should be renamed and restructured to simplify the package namespace and separate array checking from type checking. I also think the function names can be made less verbose, e.g. `check_is_length` -> `check_length`. These are substantial changes. Therefore. if possible, can this PR be held back from the 0.43 release?

tkoyama010:
> @tkoyama010 After some work on #5226, I believe this package should be renamed and restructured to simplify the package namespace and separate array checking from type checking. I also think the function names can be made less verbose, e.g. `check_is_length` -> `check_length`. These are substantial changes. Therefore. if possible, can this PR be held back from the 0.43 release?

Thank you for your comment. Please let us revert this PR once. It is possible to add it after release.

MatthewFlamm:
It might be tricky to revert if other PRs have started using it.  Is it possible to make it a private package in the meantime?  Then we can make changes at will.

user27182:
Thank you

PR #5103, opened 2023-10-19
Support user logos on the viewport

### LINKED ISSUE #4236: Adding a personal logo
I'm creating a PyVista based python package that allows researchers in a specific field to generate videos of their earth data.
The only part I'm missing is the ability for the user to add a logo of their institute to their video that will always be present e.g. in the right bottom corner.

Thus, I want to add to the plotter `p` an image that is always in the front of the scene and does not move with the camera - just like `p.add_text` or `p.add_legend` have an absolute location.

I understand that I can hack it so that at every scene I move the "logo mesh" based on the camera position to achieve this, but it doesn't seem like the proper way to go about it and I'm hoping I missed something in the documentation.

PyVista is amazing! Thank you so much for creating it and your support!

banesullivan:
Resolves #4236 to support user logos on the viewport.

By default if a logo is not specified, the PyVista logo is used:

```py
import pyvista as pv
pl = pv.Plotter()
pl.add_mesh(pv.Sphere(), show_edges=True)
pl.add_logo_widget()
pl.show()
```

Screenshot 2023-10-18 at 8 24 40 PM


pyvista/plotting/themes.py
```diff
@@ -33,10 +33,12 @@
 from enum import Enum
 import json
 import os
+import pathlib
 from typing import Callable, List, Optional, Union
 import warnings
 
 import pyvista
+from pyvista import examples
```
banesullivan, line 41: Hm. This might have implications... would prefer to not have this cross import

tkoyama010: Setting the default value to `None` is maybe enough because of the complexity of cross-importing.

pyvista/plotting/widgets.py
```diff
@@ -2629,6 +2633,77 @@ def clear_button_widgets(self):
         """Remove all of the button widgets."""
         self.button_widgets.clear()
 
+    def add_logo_widget(
+        self,
+        logo: Optional[Union[pyvista.ImageData, str, pathlib.Path]] = None,
+        position: Tuple[float, float] = (0.75, 0.8),
```
MatthewFlamm, line 2639: I assume that `position` and `size` can actually be any `Sequence` of length 2 or `ndarray` of length 2?  Same comment for the descriptions below.

tests/plotting/test_widgets.py
```diff
@@ -582,3 +582,15 @@ def release_callback(transform):
...
+
+def test_logo_widget(verify_image_cache):
+    pl = pv.Plotter()
+    pl.add_mesh(pv.Sphere())
+    pl.add_logo_widget()
+    pl.show()
+
+    pl = pv.Plotter()
+    pl.add_mesh(pv.Sphere())
+    pl.add_logo_widget(position=(0.075, 0.05), size=(0.2, 0.1))
```
MatthewFlamm, line 595: Can you make the size difference more noticable?  I can't tell a difference in size between the two snapshots.

pyvista/plotting/themes.py
```diff
@@ -2914,6 +2918,28 @@ def lighting_params(self, config: _LightingConfig):  # numpydoc ignore=GL08
...
         self._lighting_params = config
 
+    @property
+    def logo_file(self) -> Optional[pathlib.Path]:  # numpydoc ignore=RT01
+        """Return or set the logo file.
+
+        Examples
+        --------
+        >>> import pyvista as pv
+        >>> pv.global_theme.silhouette.opacity = 1.0
```
MatthewFlamm, line 2928: Fix the example here.

pyvista/plotting/themes.py
```diff
@@ -2914,6 +2918,28 @@ def lighting_params(self, config: _LightingConfig):  # numpydoc ignore=GL08
             raise TypeError('Configuration type must be `_LightingConfig`.')
         self._lighting_params = config
 
+    @property
+    def logo_file(self) -> Optional[pathlib.Path]:  # numpydoc ignore=RT01
```
MatthewFlamm, line 2922: do we also need to allow for a `str` type?  I see this is used in `add_logo_widget`.

pyvista/plotting/themes.py
```diff
@@ -2914,6 +2918,28 @@ def lighting_params(self, config: _LightingConfig):  # numpydoc ignore=GL08
...
+        Examples
+        --------
+        >>> import pyvista as pv
+        >>> pv.global_theme.silhouette.opacity = 1.0
+
+        """
+        return self._logo_file
+
+    @logo_file.setter
+    def logo_file(self, logo_file: Optional[pathlib.Path]):  # numpydoc ignore=GL08
```
MatthewFlamm, line 2934: same comment about `str` type as above.

pyvista/plotting/themes.py
```diff
@@ -2914,6 +2918,28 @@ def lighting_params(self, config: _LightingConfig):  # numpydoc ignore=GL08
             raise TypeError('Configuration type must be `_LightingConfig`.')
         self._lighting_params = config
 
+    @property
+    def logo_file(self) -> Optional[pathlib.Path]:  # numpydoc ignore=RT01
+        """Return or set the logo file.
```
MatthewFlamm, line 2923: Can we state the default behavior is the PyVista logo.  Probably too complex to describe why it is a default of `None`.

MatthewFlamm (commented):
There are also some untested parts here.  It would be good to test with a user supplied ImageData. Also should be straightforward to test with an invalid input for the logo.

pyvista/plotting/themes.py
banesullivan: The setter will always make it a `pathlib.Path` object or `None`

banesullivan: The setter will always make it a pathlib.Path object or None

pyvista/plotting/widgets.py
banesullivan: My understanding is that when type hinting for structures that have a defined/fixed length, you need to use `List` or `Tuple` and that `Sequence` is for variable length structure

pyvista/plotting/themes.py
banesullivan: I opted to remove the example as the example would need to specify a path and wouldn't be copy/pastable for end users.

pyvista/plotting/themes.py
banesullivan: I updated all of this to set the value as a `str` on the theme object to ensure its serializable

pyvista/plotting/themes.py
banesullivan: Add a note

pyvista/plotting/widgets.py
MatthewFlamm: List is a variable length container too.  What if the user gets these values from some other calculation in a different type?  This will be very annoying to change their code because we have opinionatedly required only Tuple.  This assumes they are type checking.  We should use the broadest type possible.

MatthewFlamm: > `List` is a variable length container too. What if the user gets these values from some other calculation in a different type? This will be very annoying to change their code because we have opinionatedly required only `Tuple`. This assumes they are type checking. We should use the broadest type possible.

pyvista/plotting/themes.py
MatthewFlamm: Good point.  One thought is to make it clear that it is not copy/pastable like `pv.global_theme.logo_file = "path/to/image_file.txt". # Use your own path`.

pyvista/plotting/widgets.py
```diff
@@ -2629,6 +2633,77 @@ def clear_button_widgets(self):
...
+        size: Tuple[float, float] = (0.2, 0.2),
+        opacity: float = 1.0,
+    ):
+        """Add a logo widget to the top of the viewport.
+
+        Parameters
+        ----------
+        logo : pyvista.ImageData or pathlib.Path
```
MatthewFlamm, line 2647:
```suggestion
        logo : pyvista.ImageData or pathlib.Path, optional
```

pyvista/plotting/widgets.py
```diff
@@ -2629,6 +2633,77 @@ def clear_button_widgets(self):
...
 
+    def add_logo_widget(
+        self,
+        logo: Optional[Union[pyvista.ImageData, str, pathlib.Path]] = None,
+        position: Tuple[float, float] = (0.75, 0.8),
+        size: Tuple[float, float] = (0.2, 0.2),
+        opacity: float = 1.0,
+    ):
+        """Add a logo widget to the top of the viewport.
+
```
MatthewFlamm, line 2644: We should also add a note about default behavior, either in the description here like in the theme, or in the parameter description.

pyvista/plotting/themes.py
banesullivan: Done in 538533bb5dc722ed08f9329444a29a2ebd4eb7e9

pyvista/plotting/widgets.py
banesullivan: Done in ba04807cd2f02ff8d485612484c0061aa969913a

pyvista/plotting/widgets.py
banesullivan: After a little digging, it seems using `Tuple[...]` is the correct type annotation when it is fixed length. Otherwise, I would need to add a union with `List[float]`, which would then suggest that any variable length list of floats can be passed.

banesullivan: My stance is to leave this as is. If you feel strongly, please feel free to implement a correction here

banesullivan: Done

MatthewFlamm (approved):
This LGTM as is.  if you want to improve this, there are a few untested branches.

Thanks for changing the typing. The type `Union[Tuple[float, float], Sequence[float]]`, is redundant, i.e. `Tuple[float, float]` is already a `Sequence[float]`, but it does give the user a hint that it is a two length sequence that is wanted.  I'm fine with the implementation.

tkoyama010 (changes requested):
Please give me a time to improve test coverage.

tkoyama010 (dismissed):
LGTM

MatthewFlamm (dismissed):

banesullivan:
Thanks for improving the tests, @tkoyama010!

Thanks for the review, @MatthewFlamm!

tkoyama010 (approved):

PR #5109, opened 2023-10-20
Pin Vale version to version 2.29.5

tkoyama010:
### Overview

This confirms Vale's version upgrade with dependabot's version-up PR. Currently, the latest version of Vale is always used. It is annoying to get Vale's warning outputs unrelated to the respective PRs because of it.

### Details

- None

### DIFF
--- a/.github/workflows/style_docstr.yml
+++ b/.github/workflows/style_docstr.yml
@@ -54,6 +54,17 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.11'
+        cache: 'pip'
+        cache-dependency-path:  |
+          requirements.txt
+          requirements_docs.txt
+    - name: Vale Version
+      run: |
+        echo "VALE_VERSION=$(vale --version | sed -e 's/vale version //g')" >> $GITHUB_ENV
     # this is equivalent to
     # vale --config doc/.vale.ini doc pyvista examples
     - name: "Run Vale"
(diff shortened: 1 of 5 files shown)


tkoyama010:
LGTM

MatthewFlamm:
vale isn't a pypi package as far as I can tell.  I don't think this will work for auto updating.  We might as well just hard code the version in the action unless there is a way to get it to auto update from whatever package system vale uses.

tkoyama010:
> vale isn't a pypi package as far as I can tell. I don't think this will work for auto updating. We might as well just hard code the version in the action unless there is a way to get it to auto update from whatever package system vale uses.

@MatthewFlamm 
Thanks for your comment. I found the following side.
https://pypi.org/project/vale/
The site seems actively updated. So dependabot autoupdate requirements_docs.tx in my mind.

tkoyama010:
It seems that what I have presented is for providing a Python package independent of the vale project. And I found a helpful GitHubActions called [check-vale-update.yml](https://github.com/daniperez/vale-python-package/blob/main/.github/workflows/check-vale-update.yml) in the [vale-python-package](https://github.com/daniperez/vale-python-package). Isn't it a good idea to add this to our project?

MatthewFlamm:
This could work, but it could become out of sync.  For example, if this third party provided python package publishes a version that is subsequently yanked from the official repository.  Not sure what the risk is here, but pinning it in some form is important to avoid all the noise in PRs.  I don't have a great suggestion here, so maybe we can pin it directly in the action until we find a good path forward?

MatthewFlamm (approved):
LGTM.  Thanks!

PR #5112, opened 2023-10-20
Ensure Theme subclasses cannot receive arbitrary attributes after instantiation

### LINKED ISSUE #5106: Themes allow setting of non-existent properties

I found a bug (that I introduced in #4516). The themes are supposed to prevent setting of properties that don't exist. #4516 (found via a git bisect) introduced a bug where this is no longer handled and users can set theme options that don't exist. Making typos silently fail:

The following should error because the property is `show_scalar_bar` not `show_scalar_bars`:

```py
import pyvista as pv
pv.global_theme.show_scalar_bars = False
```

Current version of PyVista

```python
>>> import pyvista as pv
>>> pv.global_theme.foo = 0  # expect an AttributeError
>>>
```

Previous versions of PyVista: 

```py
>>> import pyvista as pv
>>> pv.global_theme.foo = 0
AttributeError: 'DefaultTheme' object has no attribute 'foo'
>>>
```

### System Information

```shell
main branch
```

### Screenshots

_No response_

MatthewFlamm:
### Overview

Closes #5106.  This adds an empty slots object to every subclass to ensure that they all cannot get new attributes.


tests/plotting/test_theme.py
```diff
@@ -388,6 +388,21 @@ def test_theme_slots(default_theme):
...
+
+    # Repeat with a subclass
+    theme = DarkTheme()
+    with pytest.raises(AttributeError, match='has no attribute'):
+        theme.new_attr = 1
+
+    # verify we can't create an arbitrary attribute on an attribute
+    assert default_theme.lighting_params
+    with pytest.raises(AttributeError, match='has no attribute'):
+        default_theme.lighting_params.new_attr = 1
```
tkoyama010, line 404: We need a comment to why this test is being run again after defining the subclass.

tkoyama010 (changes requested):

MatthewFlamm:
When adding in empty `__slots__` to all subclasses `pv.global_theme.load_theme(new_theme)` seems to be broken.  This is why all the image regression tests are currently failing because the testing theme doesn't get set correctly.

MatthewFlamm:
These lines are the cause.  Not sure how this worked before:

https://github.com/pyvista/pyvista/blob/1430f0cff6f96d0532a76cb2eb3bc9e57e3ed294/pyvista/plotting/themes.py#L2825-L2826

MatthewFlamm:
There are a lot of peculiarities about `Theme`.  For example, `pv.set_plot_theme('dark')` does not make `pv.global_theme` an instance of `DarkTheme`.  It might be better to fix these peculiarities in addition to the issue of this PR.

tests/plotting/test_theme.py
MatthewFlamm: Good catch, this was supposed to test the subclass.  Now fixed.

pyvista/plotting/themes.py
```diff
@@ -161,7 +173,7 @@ def to_dict(self) -> dict:
         """
         # remove the first underscore in each entry
         dict_ = {}
-        for key in self.__slots__:
+        for key in self._all__slots__():
```
MatthewFlamm, line 176: This logic only really works because only one class in the hierarchy has non-empty `__slots__`.  It is needed because subclasses of `Theme` have no `__slots__`, we copy the attributes from the parent `Theme`.  A better solution in my opinion would be to get rid of these subclasses, i.e. make them objects of `Theme`, but this is an API change.

pyvista/plotting/themes.py
```diff
@@ -2822,7 +2841,7 @@ def load_theme(self, theme):
                 '``theme`` must be a pyvista theme like ``pyvista.plotting.themes.Theme``.'
             )
 
-        for attr_name in theme.__slots__:
+        for attr_name in Theme.__slots__:
```
MatthewFlamm, line 2844: Since we know in this case it is the `Theme` `__slots__` we want, we do not need to access `_all__slots__`

MatthewFlamm:
This is the fix I propose for #5112 that keeps the same API.  I can make a separate PR in the case we'd rather change the API.

banesullivan (dismissed):
LGTM, thanks @MatthewFlamm

banesullivan (approved):

tkoyama010 (approved):

PR #5115, opened 2023-10-21
Fix the error of docstring example code in `open_movie`

tkoyama010:
### Overview

Fix the error of docstring example code in `open_movie`.

### Details

- None


tkoyama010:
LGTM

PR #5119, opened 2023-10-24
Make built-in themes class methods, deprecate built-in classes; simplify to `slider_style`; allow registering default themes.

### LINKED ISSUE #5549: A way to 'register' new `Theme` classes

I would like to be able to add to the list of `Theme`s that can be selected via `set_plot_theme()` / `$PYVISTA_THEME` environment variable / etcetera.

It seems the simplest solution might be to make [`themes._NATIVE_THEMES`](https://github.com/pyvista/pyvista/blob/0b89a69b04b76af1052e876ce408b084cdc6eecb/pyvista/plotting/themes.py#L3233-L3242) a **public**, **modifiable** object, but you might prefer a different solution.

This change would make it easier for packages that wrap PyVista to offer their own themes without needing their own copy of `set_plot_theme()` and surrounding architecture. For example my work in bjlittle/geovista#668.

I would be happy to raise a pull request if you like the idea.

_No response_

### Pseudocode or Screenshots

```python
import pyvista as pv
from pyvista.plotting import themes

class MyTheme(themes.Theme):
    self.name = "my_theme"
    self.color = "lightgray"

themes.THEME_LIST["my_theme"] = MyTheme

pv.set_plot_theme("my_theme")

```

MatthewFlamm:
### Overview

There are several changes in this PR, major changes listed here

1. Deprecate subthemes and only use classmethods for set themes, and these have a structured way to define defaults.  All sub-configs can have defaults too, for now only `slider_style` uses this.
3. `Theme.slider_styles` is removed to have a single `Theme.slider_style`, and the multiple styles are now default styles accessible like `Theme.slider_style.classic()`.
4. Add `register_default` to allow for downstream packages to hook into PyVista `Theme` to provide defaults that can be selected in `pyvista.set_plot_theme` and more.
5. Add ability to use kwargs to define Theme.  For example, `Theme(show_edges=True)`.
6. Add `update_from_dict` method that allows an existing `Theme` instance to be updated by a dictionary rather than created from scratch.

### Problem with subclassing Theme

This PR deprecates the class usage of themes, e.g. `DarkTheme`.  There are a few problems with this approach.

1) we currently use `__slots__` to prevent users from accidentally setting a value to a non-existing attribute, e.g. a typo.  We also use `__slots__` for serializing and doing `__eq__` checks etc.  But subclasses either do not have `__slots__` or have empty `__slots__`.  This caused #5106.  #5112 is proposed to fix this using the class methodology.
2) The concept of a class for a defined theme creates problems if users want to use the class to infer something about the type of theme.  For example:

```py
import pyvista as pv
from pyvista.plotting.themes import Theme, DarkTheme, DocumentTheme

theme = DarkTheme()
theme.load_theme(DocumentTheme())
isinstance(theme, DarkTheme)  # True
assert theme == DarkTheme() # False
assert theme == DocumentTheme()  # True
```

3) The subclasses only exist to provide starting initializations of the `Theme` class.

This PR instead makes the theme class methods and deprecates the subclasses.  While we still need the changes in #5112 since the classes are still present (just deprecated), this has the potential to remove the need for the `_all__slots__` functionality when those classes are fully removed.

1) `_all__slots__` will no longer be needed when the classes are removed.
2) The user won't be tempted to use `isinstance` checks
3) It makes it clear that the methods return a `Theme` object with settings of interest.

### Problem with `Theme.slider_styles`

`Theme.slider_styles` contain two separate slider styles, i.e. these are like default themes.  `add_slider_widget` allows users to select from these styles.  However if no style is selected in `add_slider_widget` then the user cannot use a `Theme` to style the widget.

This PR uses a single `Theme.slider_style` that is now the default style when no style is chosen in `add_slider_widget`.  The user can still pass in a string to choose from a separate default style in `add_slider_widget` so there is no breaking change.  There may be subtle differences due to the new default style not exactly matching the defaults in `add_slider_widget`.

### Details

I recommend addressing #5112 first.  Note that this PR includes the same changes as #5112.

closes #5549

TODO:

- [x] Add deprecation comments for version to remove


banesullivan:
@MatthewFlamm, i haven't looked at this in depth at all but I want to mak sure you are aware of a usecase where downstream projects subclass `Theme` to create their own themes. I've seen this adopted in a handful of closed and open source projects.

MatthewFlamm:
> @MatthewFlamm, I haven't looked at this in depth at all but I want to make sure you are aware of a use case where downstream projects subclass Theme to create their own themes. I've seen this adopted in a handful of closed and open-source projects.

Good point.  I don't recommend this strategy for all the reasons I wrote, but I think we can keep support for user subclasses.  Then they can deal with the weirdness.  If/When we remove these classes here, we'd have to add a test for this use case. I will add to the TODO.

Also, I'm open to opinions on why this PR isn't a good idea.

MatthewFlamm:
One potentially painful change however is if these users subclass one of our subclasses.  `MyTheme(Dark theme)` for example.  This PR will still be a problem for that use case.

banesullivan:
@MatthewFlamm, I generally like what you've done here. Upon reflection, only a few projects are using custom theme subclasses and I'm pretty sure it is only me doing this. If you want to issue a harsh deprecation for that in favor of this, it's good by me. I just want to make sure we capture how a user would define their own custom theme as that general concept is incredibly useful to at least me

MatthewFlamm:
This PR does _not_ prevent maintaining a custom Theme.  It will continue to support maintaining a custom subclass of Theme but not a built-in subclass of Theme.  That is users can still do `class MyTheme(Theme):` but not `class MyTheme(DarkTheme):`.  This PR is suggesting a better alternative is the following.

```py
my_theme = Theme()
my_theme.setting1 = value1

# or

my_theme = Theme.dark_theme()
my_theme.setting1 = value1

# or

my_theme = Theme.from_dict(my_theme_dict)
```

What do you think about extending this PR to also include this usage?  We would accept a theme_dict or kwargs...

```py

my_theme = Theme(base_theme="dark", setting1=value1, ...)
```

banesullivan:
I really like the `from_dict` approach and definitely would love to have the keyword argument support for:

```
my_theme = Theme(base_theme="dark", setting1=value1, ...)
```

MatthewFlamm:
Making this PR draft to address some of the discussion here.

1) More explicit documentation around custom theme maintenance.
2) Allow instantiation via `Theme.__init__` via
  - `**kwargs`
  - `theme_dict` <- this one is questionable as we already have `Theme.from_dict`

MatthewFlamm:
There is one more related issue that causes problems with the proposals of having dictionary or kwarg cascade of information through the configurations. `_SliderConfig` works differently than all the other classes in that it stores two separate possible configs that are each configurable.   Right now you can tinker with each part separately, e.g. `pv.global_theme.slider_styles.modern.slider_length = 0.02` and `pv.global_theme.slider_styles.classic.slider_length = 0.02`.  It would be more consistent with how all the other parts of `Theme` works to have a single configuration and set the property directly, e.g. `pv.global_theme.slider_style.slider_length = 0.02`. We could have preset slider_styles that could loaded and then modified, similar to what I proposed for the `Theme` itself.  I'm now thinking that we could have the following type of usage for both Themes and sub-parts:

- `pv.global_theme.set("dark") # would override to dark theme preset`  Or `pv.global_theme = Theme.dark_theme() # same effect`
- `pv.global_theme.slider_style.set("classic")` or `pv.global_theme.slider_style = _SliderConfig.classic()`  # Note that the last example would be better if we make public the sub-themes.  Which I would advocate for as we already have them in the documentation.


One attractive upside of this approach is that we could could enable presets that could be mixed-matched within an overall theme. For example, multiple configs for `slider_style` that is separate from multiple configs for `lighting_params` and so on.  An example is you could choose the `dark_theme` but with a particular preset for `slider_style` that is typically found in a different theme as an example.

tkoyama010:
pre-commit.ci autofix

MatthewFlamm:
Coming back to this again.  `slider_styles` is only consumed in `add_slider_widget`.  Slider config styles works backwards compared to all the other configuration in the themes.  That is the default behavior of `add_slider_widget` is _not_ set by the global theme, but rather set inside `add_slider_widget`.  See

https://github.com/pyvista/pyvista/blob/0b89a69b04b76af1052e876ce408b084cdc6eecb/pyvista/plotting/widgets.py#L1546-L1563

Only if the user supplies `style` does the behavior change to some set in the theme.  This is backwards in my opinion.  I am going to move forward with trying to fix this, but keep as many backwards compatibility options as possible.  It is inevitable that some breaking changes or subtle differences will result however.

MatthewFlamm:
This is nearing completion for review.  I have made several changes.

1) deprecate subthemes and only use classmethods for themes
2) use an private attribute to keep track of available themes instead of _NATIVE_THEMES
3) rework the way slider styles work by having a single theme, but use the classmethods to define the different pre-built themes
4) add `register_default` to allow for adding to the available themes
5) add ability to use kwargs to define Theme

Still TODO:

- [ ] add comment deprecation version to remove
- [ ] some docstrings still missing
- [ ] add to existing example documentation if needed
- [ ] add example documentation for adding default themes
- [ ] write deprecation note in this PR description

MatthewFlamm:
This is now ready for review.  I updated the PR description with a summary of the high level changes to the code.  I think it makes for a more sane and consistent Theme API.

@bjlittle - this PR has changes which will necessitate a small change in testing for geovista.

Some other notes that I'd like to address in future PRs (or in this PR if reviewers feel any piece warrants addressing here).  This one is already very long.

- `Theme` should be easier to access in the API, maybe moved into the `pyvista` namespace.
- The sub-configs like `_SliderStyleConfig` are private classes, yet they are in the documentation as they are needed for users to inspect attributes.  We should make them public.  This PR also potentially makes them more useful to users by adding the possibility to add defaults for each sub config class.

MatthewFlamm:
github-actions preview

MatthewFlamm:
@tkoyama010 I see you have self assigned this.  Do you think it is worth cleaning up the merge conflicts here to get a review?

MatthewFlamm:
precommit.ci autofix

MatthewFlamm:
pre-commit.ci autofix

MatthewFlamm:
@tkoyama010 this should now be ready for review. 

@user27182 I am baffled by some of the type errors that occurred with this PR.  Specifically I had to sprinkle a bunch of `type: ignore` on a few lines that don't appear to be directly related to these changes.  See the changes in 89b5ca2.  mypy cannot determine the type, but I don't know how to fix this.  Any ideas?

pyvista/plotting/utilities/regression.py
```diff
@@ -217,7 +217,7 @@ def to_img(img):
         elif isinstance(img, np.ndarray):
             return wrap_image_array(img)
         elif isinstance(img, Plotter):
-            if img._first_time:  # must be rendered first else segfault
+            if img._first_time:  # type: ignore[has-type]
```
MatthewFlamm, line 220:
```suggestion
            # must be rendered first else segfault
            if img._first_time:  # type: ignore[has-type]
```

user27182:
> @user27182 I am baffled by some of the type errors that occurred with this PR.

The plotting module is super messy type-wise. For attributes I find `mypy` will sometimes infer the type from anywhere it sees the first assignment to that attribute. This usually happens if the attribute isn't a formal `property`. So, if it sees an assignment in `__init__` or even in some misc. function to, say `int` type, then it will assume that this is the required type for that attribute everywhere. But then elsewhere a `str` type may be assigned, so it complains because it thinks it should be `int`, not `str.

I think the proper resolution is to identify everywhere in the code where the attribute is assigned a value, and then try to work backward to figure out what the type union should be (e.g. `int | str` for this example), and then hint that type to mypy.

But, the interactions are so messy between classes that this could turn into a game of whack-a-mole. Sometimes this may require minor structural changes to fix. I think it's OK to ignore these challenging errors for now. Doing separate, dedicated PR trying to fully type a single class or its methods is probably necessary to fully fix these.

examples/02-plot/themes.py
```diff
@@ -168,6 +179,44 @@ def plot_example():
 pl.add_mesh(pv.Cube())
 cpos = pl.show()
 
+###############################################################################
+# An alternative to setting theme configurations via attributes is to use
```
MatthewFlamm, line 183: This section is redundant to below and it needs a section header

MatthewFlamm:
Before this PR, the testing theme was private although GeoVista is using it, related to the integration test failure.  This PR makes it public, as the GeoVista use shows that downstream packages want to use it.

This could be problematic if we ever want to adjust the testing theme in the future for PyVista.  My thought is if this happens, we can introduce a private functionality to themes and use a `_pyvista_testing_theme` while keeping the existing `testing_theme` unchanged.  THe only thing needed to enable private themes is to remove any leading underscore themes returned from the `defaults` class method.

I'm also okay if we want to make it private here and think about public testing themes in a future PR.  Thoughts?

examples/02-plot/themes.py
MatthewFlamm: Also cleaned up for the newer cell-like commenting.

pyvista/plotting/themes.py
```diff
@@ -230,6 +252,71 @@ def _all__slots__(cls):
...
+        --------
+        Theme.defaults
+
+        """
+        theme = getattr(cls, cls._defaults[name])
+        return theme()
+
+    @classmethod
+    def defaults(self):
+        """Return list of default themes.
```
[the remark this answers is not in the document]
user27182: Maybe this can be used instead? Can be probably me moved into core utilities somewhere for re-use
https://github.com/pyvista/pyvista/blob/ed4178eefeb1553ffe2d902150285e648ae1c4c4/doc/source/make_tables.py#L85-L94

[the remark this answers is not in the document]
MatthewFlamm: Let's do this if possible!

examples/02-plot/themes.py
```diff
@@ -169,6 +180,36 @@ def plot_example():
...
+my_theme = Theme(color='blue', show_edges=True)
+
+pv.global_theme.load_theme(my_theme)
+cpos = pv.Sphere().plot()
+
+# %%
+# # A theme can also be updated using a dictionary.  Here the global theme is
+# directly updated.
+
+pv.global_theme.update_from_dict({'show_edges': False})
```
user27182, line 210: If we want to update from another theme, do we need to do `pv.global_theme.update_from_dict(other_theme.to_dict())`? Is it better to maybe simply call it `update` and allow either another theme or a dict as input?

pyvista/plotting/__init__.py
```diff
@@ -62,7 +62,7 @@
 from .texture import Texture as Texture
 from .texture import image_to_texture as image_to_texture
 from .texture import numpy_to_texture as numpy_to_texture
-from .themes import DocumentTheme as _GlobalTheme
+from .themes import Theme as _Theme
```
user27182, line 65: Is the inclusion of the underscore intended to communicate that the import should not be made from `plotting` directly, but importing from `themes` directly is ok?

user27182 (commented):
I am not too familiar with themes so can't really comment on the motivation for this PR but the changes here all look good and are a nice improvement.

My main general comment would be to add `.. versionadded::` and `.. versionchanged::` directives as needed. Otherwise I just have a few minor comments.

pyvista/plotting/themes.py
```diff
@@ -174,6 +188,15 @@ def from_dict(cls, dict_):
                 setattr(inst, key, value)
         return inst
 
+    def update_from_dict(self, dict_):
+        """Update theme in place with dictionary."""
```
MatthewFlamm, line 192: Needs documentation

pyvista/plotting/themes.py
```diff
@@ -89,8 +93,8 @@ def load_theme(filename):
     Examples
     --------
     >>> import pyvista as pv
-    >>> from pyvista.plotting.themes import DocumentTheme
-    >>> theme = DocumentTheme()
+    >>> from pyvista.plotting.themes import Theme
+    >>> theme = Theme.document_theme()
```
MatthewFlamm, line 97: `load_theme` currently also accepts a `Theme` object.  This existing before this PR.  This either needs to be documented or split out.

examples/02-plot/themes.py
MatthewFlamm: I want to flesh out my thinking here.  Sorry for the long winded response just to agree with you, but your comment has raised a question that I don't have a position on yet.

Using current PR naming, the `update_from_dict` uses the current `pv.global_theme` (in this line) and only overrides the values on the current `pv.global_theme`.  That is, all the other values are unchanged.

When using a `theme` to update the global theme, we use `pv.global_theme.load_theme(new_theme)` in the current PR.  This completely replaces the old settings of `pv.global_theme` with _all_ the settings from `new_theme`.  This method existed before this PR, but the usage of passing in a `Theme` instance object was undocumented.

Related is `from_dict` (existed before this PR) which starts with the base settings from `Theme()` and updates the dict from that.  On naming, I chose to disambiguate the naming with `update_from_dict`, which updates in place rather than `from_dict`, which starts fresh.

I see what you are saying.  `pv.global_theme.update_from_dict(other_theme.to_dict())` would be more convenient to be ``pv.global_theme.update(other_theme)`.  I'm +1 for making it `update` and allowing either a `dict` or a `Theme` instance object.  We will document using a `Theme` instance object will override the entire theme with the new theme.

My question: `load_theme` becomes nonideal in this proposal because it now has overlapping functionality with `update`.  Should we deprecate the undocumented use of passing in a `Theme` object, or document it as a first class usage, or keep it in the undocumented state but allow it?

pyvista/plotting/__init__.py
MatthewFlamm: I admit that I didn't think about this and simply used the precedence of the existing code.  I think we should avoid making more namespace changes in this PR as it is already long and complicated.

I collected some thoughts on additional things that could be done including naming and namespace in https://github.com/pyvista/pyvista/pull/5119#issuecomment-1962027567., It is now likely buried inside a "load more" button in the PR due to length.  I don't have a strong opinion on whether the namespace location for `Theme` should be changed.  I do think that the sub-config classes should become publicly named, but these should stay in `pyvista.plotting.themes`.

user27182: Sounds good, I do see that the convention was simply followed here. I was more just curious because I've seen the underscores in the documentation for themes before which sort of confused me about whether it's private or public.

Keeping this as-is is OK.

examples/02-plot/themes.py
user27182: > Using current PR naming, the `update_from_dict` uses the current `pv.global_theme` (in this line) and only overrides the values on the current `pv.global_theme`. That is, all the other values are unchanged.

This behavior seems to be consistent with `dict.update()`, so if we consider `Theme` to be kind of a like a `dict` duck type then using `Theme.update()` is consistent with this behavior (and would not be surprising IMO).

> When using a `theme` to update the global theme, we use `pv.global_theme.load_theme(new_theme)` in the current PR. This completely replaces the old settings of `pv.global_theme` with _all_ the settings from `new_theme`. This method existed before this PR, but the usage of passing in a `Theme` instance object was undocumented.

This behavior makes sense to me. And allowing a `Theme` object seems natural, and _not_ allowing a `Theme` object here seems odd. +1 for documenting this.

> Related is `from_dict` (existed before this PR) which starts with the base settings from `Theme()` and updates the dict from that. On naming, I chose to disambiguate the naming with `update_from_dict`, which updates in place rather than `from_dict`, which starts fresh.

> I see what you are saying. `pv.global_theme.update_from_dict(other_theme.to_dict())` would be more convenient to be `pv.global_theme.update(other_theme)`. I'm +1 for making it `update` and allowing either a `dict` or a `Theme` instance object. We will document using a `Theme` instance object will override the entire theme with the new theme.

I see the distinction now. Though to me `update()` should be relatively familiar to python users from its use with actual `dict` objects, so the in-place behavior should not be surpising. And I think `from_dict` similarly implies _not_ in place.

Side note: `from_dict` for themes does not appear to be documented...? https://dev.pyvista.org/search.html?q=from_dict

> My question: `load_theme` becomes nonideal in this proposal because it now has overlapping functionality with `update`. Should we deprecate the undocumented use of passing in a `Theme` object, or document it as a first class usage, or keep it in the undocumented state but allow it?

As a non-user and non-expert of themes it sounds like there are a lot of subtleties to this. So I think good documentation explaining these subtleties (similar to the explanations in your comment) would be helpful regardless of the final API. Maybe include cross-refs in `See Also` sections to the related methods?

From a practical point of view, if I have a theme (either Theme object or dict) and I want to load it... I would probably use `load_theme`. This seems intuitive. If I then want to update it from another theme, I would use `update`, which also seems intuitive. And if I want to create a new theme altogether (but not load it?) then I should use `from_dict`, is that correct?

Even if there is some overlap, as long as the methods are not 100% identical and as long as there is good documentation (with cross refs and examples) then it seems to me that this kind of API would make sense.

user27182: As a middle ground between `update` and `update_from_dict`, you could also call it `update_from`. We have something similar with [rotation_from](https://dev.pyvista.org/api/plotting/_autosummary/pyvista.prop3d.rotation_from#pyvista.Prop3D.rotation_from)

MatthewFlamm:
Given #8568, this PR is no longer viable in it's current form.  Closing.

PR #5122, opened 2023-10-25
Add theme option to allow plotting empty meshes

banesullivan:
It is increasingly common to initialize a scene with empty meshes and later populate those meshes in-place. This is currently supported via `vtkAlgorithm`, but I also want the ability to do this in my own routines.

This sets a global theme option to allow adding empty meshes to the scene for use cases like:

```py
import pyvista as pv
pv.global_theme.allow_empty_mesh = True

mesh = pv.PolyData()

pl = pv.Plotter()
pl.add_mesh(mesh, show_edges=True)
pl.show()
```

Then in some custom callback:

```py
mesh.points = [[0, 0, 0], [1, 0, 0], [1, 1, 0], [0, 1, 0]]
mesh.faces = [3, 0, 1, 2, 3, 0, 2, 3, 3, 0, 2, 3]
pl.reset_camera()
```


tkoyama010 (changes requested):
I love this. Could you explain the reason `allow_empty_mesh` is False by default? (Is it to avoid confusion when a user is presented with an empty plotter maybe?) If we don't fear user confusion, we could simply remove the error handling itself in my mind.

banesullivan:
It's false by default as that was already the default set a long time ago in https://github.com/pyvista/pyvista/commit/63f3f98656ffa3045012903dd0cb959613646e84

> Is it to avoid confusion when a user is presented with an empty plotter maybe

This is exactly why we originally raised an error! Users would often get empty plots after running different filters and this would be a major point of confusion. The idea is/was that this helps users immediately understand why no plot is generated, instead of leaving them to wonder why they are presented with an empty plotter. 

I'm not sure how effective this really was. I think changing this default warrants a dedicated issue and discussion where as this PR allows users to opt-out of this behavior

tkoyama010 (dismissed):
Got it. Thanks!

pyvista/plotting/themes.py
```diff
@@ -2709,6 +2711,29 @@ def before_close_callback(
...
+        --------
+        Enable plotting of empty meshes.
+
+        >>> import pyvista as pv
+        >>> pv.global_theme.allow_empty_mesh = True
+
+        Now add an empty mesh to a plotter
+
+        >>> pl = pv.Plotter()
+        >>> pl.add_mesh(pv.PolyData())
```
tkoyama010, line 2728:
```suggestion
        >>> _ = pl.add_mesh(pv.PolyData())
```

tkoyama010 (approved):

tkoyama010 (dismissed):

banesullivan:
@MatthewFlamm, I think this would be useful, but I too, hesitate to add yet another argument to `add_mesh()`. The usage patterns/scenarios where I was inspired to make this change are very often doing this in callbacks managed by the VTK event loop where errors are sometimes steamrolled/not caught and it would be best to just enable this by default

pyvista/plotting/themes.py
```diff
@@ -2709,6 +2711,29 @@ def before_close_callback(
...
+
+        >>> import pyvista as pv
+        >>> pv.global_theme.allow_empty_mesh = True
+
+        Now add an empty mesh to a plotter
+
+        >>> pl = pv.Plotter()
+        >>> _ = pl.add_mesh(pv.PolyData())
+        >>> pl.show()  # doctest: +SKIP
+        """
```
MatthewFlamm, line 2730: Pyvista standard is to have an empty line at the end.

```suggestion

        """
```

MatthewFlamm (commented):
One nitpick comment, otherwise LGTM

pyvista/plotting/themes.py
banesullivan: Thanks! I always forget which we do

MatthewFlamm (approved):

pyvista/plotting/plotter.py
```diff
@@ -3475,7 +3475,7 @@ def add_mesh(
             if scalars.ndim != 2 or scalars.shape[1] < 3 or scalars.shape[1] > 4:
                 raise ValueError('RGB array must be n_points/n_cells by 3/4 in shape.')
 
-        if algo is None and not mesh.n_points:
+        if algo is None and not self.theme.allow_empty_mesh and not mesh.n_points:
             # Algorithms may initialize with an empty mesh
             raise ValueError('Empty meshes cannot be plotted. Input mesh has zero points.')
```
user27182, line 3480: Feel free to ignore this comment or reject the suggestion, I don't mean to hold anything up, but I think many PyVista users (myself included) may not know much about the theme settings or that it even exists. If I got this error (and actually intended to plot an empty mesh) it would not be clear to me that there is a setting to work around the error.
```suggestion
            raise ValueError('Empty meshes cannot be plotted. Input mesh has zero points. To allow plotting empty meshes, set `pv.global_theme.allow_empty_mesh = True`')
```

banesullivan: Great suggestion!

MatthewFlamm:
pre-commit.ci autofix

MatthewFlamm (approved):

tkoyama010 (approved):

tkoyama010:
@user27182
I invite you as a member of the [developer in our project](https://github.com/orgs/pyvista/teams/developers/members). You can request changes / approve Pull Requests if you approve the invitation.

PR #5123, opened 2023-10-25
Add GitHub Codespaces badge in README.rst and CONTRIBUTING.rst

tkoyama010:
### Overview

See [One Click Into GitHub Codespaces](https://github.blog/changelog/2023-04-24-one-click-into-github-codespaces/).

### Details

- [x] ~~Now it is two click experience. We need to make it one click experience.~~


MatthewFlamm:
See #5128.  We will need to modify this PR to get a truly one-click experience.  As it stands you need to do two-clicks with how it is implemented here.  I recommend waiting until #5128 is merged, then we can add in the one-click badges here.

tkoyama010:
> See #5128. We will need to modify this PR to get a truly one-click experience. As it stands you need to do two-clicks with how it is implemented here. I recommend waiting until #5128 is merged, then we can add in the one-click badges here.

Yes. I was questioning my own policy of just adding batches without explaining anything about CodeSpace. Integrating it into your documentation would be more helpful to newcomers. Thanks for adding the documentation.

banesullivan:
I'm curious what's needed to now make this "one click"

MatthewFlamm:
I was mistaken.  Every option is really "two-click".

The current PR gives uses the link https://codespaces.new/pyvista/pyvista  gives a window like this when clicked:
image

You must then click again to get a codespace spun up.  Note that "Dev container configuration" allows to select "Osmesa" for an off screen version.  We should look into getting better name for the default.

tkoyama010:
I renamed it to `offscreen`. Feel free to suggest other names.

MatthewFlamm (approved):
Thanks for adding more to this!

banesullivan (approved):
+1 of `offscreen`

PR #5128, opened 2023-10-25
Add off screen devcontainer option

### LINKED ISSUE #5125: GitHub Dev convatiner needs `OFF_SCREEN`
We need to make sure `OFF_SCREEN=True` is set by default in the codespaces dev container so that pyvista will work out of the box in this environment

MatthewFlamm:
### Overview

Closes #5125.  This PR adds in a second devcontainer option to do off screen rendering.  This is needed in case users cannot connect via vnc for some reason, or they want to test osmesa vtk flavor etc.

### Details
We could also consider other flavors with other vtk variants, but I'm less familiar.

TODO:

- [x] add documentation about the options including the live viewer for the default devcontainer.


MatthewFlamm:
To test the offscreen container you can click "\<code\>" button -> "..." -> "New with options..." -> change "Dev container configuration" to "Osmesa".  The default is currently "desktop-lite".  It is working for me currently both ways.

offscreen tested with 
```py
import pyvista as pv
pv.Sphere().plot()
pv.Sphere().plot(screenshot="test.png")
```
image

MatthewFlamm:
I have added a section to the CONTRIBUTING.rst on this.  Not sure if it should also go somewhere else in addition...  

I have also removed the documentation on the GitPod container as I'm no longer using it, and am using codespaces primarily.  If we don't have a champion to keep it current, we should not recommend it to users.  We could consider deleting it, but I don't see harm in keeping it in case some users prefer this (and then they can help us maintain it).

tkoyama010:
Awesome. Thanks for sharing your knowledge!

CONTRIBUTING.rst
```diff
@@ -33,6 +33,19 @@ running:
...
 
+Quick Start Development with Codespaces
+---------------------------------------
+A dev container is provided to quickly get started. The default container
+comes with the repository code checked out on a branch of your choice
+and all pyvista dependencies including test dependencies pre-installed.
+In addition, it uses the
+`desktop-lite feature <https://github.com/devcontainers/features/tree/main/src/desktop-lite>`_
+to provide live interaction windows.  Follow directions
+`here <https://github.com/devcontainers/features/tree/main/src/desktop-lite#connecting-to-the-desktop>`_
```
tkoyama010, line 44: A UI design expert commented that it is helpful for readers to explicitly indicate the title of the link when add a link ([Jimenez Gacha & Presedo-Floyd - Accessibility best practices for authoring Jupyter notebooks](https://youtu.be/onjRLjveGxk?si=CkG_sZ_sSTIwB36X)). I didn't know that either, so I will try to do it from now on.
```suggestion
`Connecting to the desktop <https://github.com/devcontainers/features/tree/main/src/desktop-lite#connecting-to-the-desktop>`_
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM. Thanks!

banesullivan (approved):
Thanks, @MatthewFlamm!

tkoyama010 (approved):

PR #5131, opened 2023-10-26
Add ssort to .pre-commit-config.yaml

tkoyama010:
### Overview

This is a proposal to add [ssort](https://github.com/bwhmather/ssort) to avoid the reduced bikeshedding discussion about where to put properties and methods in the code. The idea of the order of properties and methods varies from code implementer to code implementer, and I have my own principles. However, it is against our spirit to impose our principles on others when reviewing code. However, we are always concerned about differences in principles during a review, so if we can come to an agreement on the linter, we can avoid discussions of bikeshedding in our mind.

One concern is that [ssort](https://github.com/bwhmather/ssort) is not a very common tool, so it is difficult to reach a consensus. Of course, if you all cannot agree on this, I am prepared to withdraw this proposal.

### Details

- None


tkoyama010:
.pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

tkoyama010:
I withdraw my proposal because it has no track record of being implemented in major libraries such as NumPy and Matplotlib. That is our most important criterion. I am sorry for the disturbance.

PR #5144, opened 2023-10-29
Remove set_plot_theme from examples

banesullivan:
As of #4516, the document theme is the new default. We do not need to set the theme to document in every example


tkoyama010 (approved):

tkoyama010:
Auto merge for non-controversial changes.

tkoyama010 (approved):

MatthewFlamm:
This was needed to prevent leakage from _other_ themes from messing up the theme for later examples.  I recommend keeping it.

MatthewFlamm:
An example is that a bug or regression in one example could poison all other examples making it very difficult to debug.

PR #5147, opened 2023-10-30
Clean up unused arguments

banesullivan:
Some clean up to outright remove unused arguments and make the usage of unused arguments more consistent in cases like VTK callbacks.

Locally I used `flake8-unused-arguments` to find these occurrences. Unfortunately, I do not think we should add `flake8-unused-arguments` to our pre-commit checks at this time as it doesn't seem widely used, and it doesn't handle ignoring testing fixtures well at this time. 

For cases where we need to capture arguments positionally and might ignore preceding arguments, I opted for the leading underscore convention

For the case of many VTK event callbacks, I opted to typically capture the positional arguments but with leading underscores (as opposed to just `*args`). Why typically? Because in some of these places, I foresee us wanting to tap into the passed arguments. Otherwise, I did use `*args`.

While it might not be ideal, I opted not to issue deprecation warnings for the few public-facing APIs that had unused keyword arguments. In my mind, it feels weird to deprecate a non-existent feature... but then again, people could have been using these arguments thinking they were doing something. I am happy to change this and issue full deprecations for the few public-facing scenarios if others feel it is warranted


tkoyama010 (approved):
It is my opinion that deprecated additions to this are not necessary because the effort is not worth the benefit. I'd rather spend the time to add it to implementing new features to make users happy.

Please let us know what @MatthewFlamm  opinion is this before merging.  I apologize for merging before I heard your opinion last time,

MatthewFlamm:
I haven't gone through this in detail but scanned the changes generally.  I think it is good to clean up parameters that are unused.  But, I think this sort of linting is not that important for this code base given use cases like default callbacks.

I find the usage of `_parameter` in the function signature a bit unorthodox. These are still unused parameters, but I assume this is so that the linter no longer complains.  I find myself going into the codebase to find examples of usage.  This change will make it odd to read.    The cases where all parameters are swallowed by an `*args, **kwargs` will make this not as useful to users looking for examples.  In the cases that we know there are specific `args and kwargs`, this obfuscates more than helps in my opinion.

If the reason is that we do not currently use the parameters, does this mean that they will become non-privatized at that time?  If so, this is ongoing churn.

banesullivan:
> but I assume this is so that the linter no longer complains.

Exactly and this using either a leading `_` or just an `_` is a convention for "this isn't used"

> In the cases that we know there are specific args and kwargs, this obfuscates more than helps in my opinion.

Fair point... I may think about this a bit more

> If the reason is that we do not currently use the parameters, does this mean that they will become non-privatized at that time? If so, this is ongoing churn.

Yes, as arguments are used, we'd non-privatize them at that time and yes... this would mean that its technically ongoing, though rare

MatthewFlamm:
Thanks for the further explanation.  For the internal code uses, it doesn't matter too much.  So I'm okay if this is wanted.

larsoner:
With the latest PyVista release on MNE-Python I now get because of this PR:
```
--> 159 plotter = self._plotter_class(**self.store)
    160 plotter.background_color = self.background_color
    161 self._plotter = plotter

TypeError: Plotter.__init__() got an unexpected keyword argument 'multi_samples'
```
So even though this kwarg did nothing, it's a backward compat break with no deprecation warning. In the future, would it be possible to have a deprecation cycle (i.e., with a DeprecationWarning) for stuff like this so that code doesn't immediately break?

banesullivan:
Sorry about that @larsoner, we'll do better in the future to issue deprecation warnings. I made a conscious decision not to issue one for this argument (above) based on the incorrect assumption no one was using it

banesullivan:
Though I'm curious why this didn't happen in our integration tests.... this should have appeared, right?

larsoner:
Looks like it did but only in an example maybe?

https://github.com/pyvista/pyvista/pull/5185

PR #5160, opened 2023-11-01
Support button release events on interactor

### LINKED ISSUE #4976: Button Release Events

I'm trying to play around with button release events, but not getting any traction. I've included a minimal working example below.

If I change the event to just `LeftButtonPressEvent`, I can get the callback to work. But otherwise, I'm not seeing what I would expect. In fact, there appear to be numerous events that are not handled. Is there a way to do this?

https://vtk.org/doc/nightly/html/classvtkCommand.html#a59a8690330ebcb1af6b66b0f3121f8fea9f5bb78d69ea168492a544c1ebc5248c

### Pseudocode or Screenshots

```python

import pyvista as pv
p = pv.Plotter()
s = pv.Sphere()

def cb(s,e):
    print(e,flush=True)

p.iren.add_observer('LeftButtonReleaseEvent',cb)
p.show()

```
Or drag your screenshot here!

banesullivan:
Resolves #4976 to tap into the interactor style for button release events

cc @emielkemmm


tests/plotting/test_render_window_interactor.py
```diff
@@ -252,3 +252,24 @@ def test_add_pick_observer():
...
+
+    cb = CallBack()
+    pl = pv.Plotter()
+    pl.iren.add_observer(event, cb)
+
+    pl.iren.interactor.GetInteractorStyle().InvokeEvent(event)
+    assert cb._i == 1
+
+    pl.iren.interactor.GetInteractorStyle().InvokeEvent(event)
+    assert cb._i == 2
```
banesullivan, line 275: Test this is actually pretty tough but I think this is sufficient

emielkemmm:
@banesullivan Awesome! thanks for digging into this!

tkoyama010 (approved):
LGTM

PR #5167, opened 2023-11-01
Update .pre-commit-config.yaml to exclude Sphinx objects.inv files

tkoyama010:
### Overview

Follow up of #5161 .

### Details

- None


tkoyama010:
LGTM

Issue #3851, opened 2023-01-22
Trame Jupyter support on Google Colab


francesco-ballarin:
Hi @banesullivan,
as mentioned in #3690 I am interested in testing the new trame backend on online jupyter notebook services.
Since #3385 is now ready, I have started looking into it.

My first try was with Google Colab, see
https://colab.research.google.com/drive/1608zRoubwfN2KgTRcEk5hOq3tQJSF0XQ?usp=sharing

I currently get an error at
```
await pv.set_jupyter_backend('client')
```
namely
```
ERROR:root:Exception raised by task = <Task finished name='Task-2' coro=<AiohttpWslinkServer.start() done, defined at /usr/local/lib/python3.8/dist-packages/wslink/backends/aiohttp/__init__.py:101> exception=OSError(99, "error while attempting to bind on address ('::1', 0, 0, 0): cannot assign requested address")>
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/trame_server/core.py", line 533, in on_done
    task.result()
  File "/usr/local/lib/python3.8/dist-packages/wslink/backends/aiohttp/__init__.py", line 123, in start
    await my_site.start()
  File "/usr/local/lib/python3.8/dist-packages/aiohttp/web_runner.py", line 121, in start
    self._server = await loop.create_server(
  File "/usr/lib/python3.8/asyncio/base_events.py", line 1463, in create_server
    raise OSError(err.errno, 'error while attempting '
OSError: [Errno 99] error while attempting to bind on address ('::1', 0, 0, 0): cannot assign requested address
```
Do you have any idea what could be the issue here? Thanks!

banesullivan:
hi @francesco-ballarin, I'm excited to hear you're trying out the new jupyter backend! The problem you're facing is a common one which I may address directly in #3385. The default `host` for `wslink` (the server mechanism behind Trame) is `'localhost'` and some containerized environments do not particularly like that. You will need to switch the host to `'127.0.0.1'`. This can be done through the `TRAME_DEFAULT_HOST` environment variable.

I added some notes in #3385 about needing this in Docker but didn't realize it's probably the case for a lot of containerized environments (Colab included). 

@jourdain, I'm thinking about making the `host` default to `'127.0.0.1'` in PyVista when launching the Trame server to avoid this issue, what do you think? This may have issues on Windows, so perhaps I could have this logic:

```py
# Defualt host to "127.0.0.1":
[...]

PR #5138, opened 2023-10-28
🚀 Release 0.43

tkoyama010:
PyVista December 2023 release

Release v0.43.0

---
## Highlights

Feel free to suggest highlights you want to add.

- #4855 by @FelipeCybis

https://github.com/pyvista/pyvista/assets/41338087/d6ba752a-b158-4441-83b0-3e638529c6ef

- #4965 by @MatthewFlamm

- #4974 by @tkoyama010 

- #5103 by @banesullivan 

Screenshot 2023-10-18 at 8 24 40 PM

- #5171 by @akaszynski 

- #5176 by @jmargeta

Screenshot 2023-10-18 at 8 24 40 PM

## Waiting on the following Issues or PRs:

- [x] #4688
- [x] #5091
- [x] #5095
- [x] #5158
- [x] #5160
- [x] #5112 
- [x] #5128                                                                     
- [x] #5133
- [x] #5140
- [x] #5143
- [x] #5176
- [x] #5168 
- [x] #5247
- [x] #5240
- [x] #5242
- [x] #5246
- [x] #5271

Potentially:

Next release:

- #4938
- #4942
- #5029
- #5119
- #5188

------

## Steps to release

- [x] Sync this branch with `main`
- [x] Grep for deprecations slated to be removed by `v0.43` and remove
- [x] Bump `main` to `0.44.dev0` (this PR)
- [ ] Keep this branch (`release/0.43`) and update the version to `0.43.0`. And commit the change to this branch.
- [ ] Tag `v0.43.0` and push the tag.
- [ ] Verify [docs](https://docs.pyvista.org/) and [PyPI](https://pypi.org/project/pyvista/) have updated.
- [ ] Update the [conda-forge feedstock](https://github.com/conda-forge/pyvista-feedstock).
- [ ] Generate [release documentation](https://github.com/pyvista/pyvista/releases/tag/v0.43.0)

------

PR #5168, opened 2023-11-02
Add embeddable jupyter widget

banesullivan:
Adds an embeddable jupyter backend that exports the scene using Trame-VTK: `pv.set_jupyter_backend("html")`

cc @ChristosT 

This mostly does it, but I cannot quite get the widget state to save and rerender when reopening a notebook


ChristosT:
I don't see any issues. What exactly does not work ?

[Here](https://github.com/ChristosT/pyvista/tree/feat/embeddable-widget) a patch that add a working tinypages test-cases that validates the code.


https://github.com/pyvista/pyvista/assets/6725596/017a525d-bb6d-462c-9241-6a04f2b6040d

banesullivan:
I'm trying to get this to work with tools like nbviewer where the widget state is embedded in a notebook https://nbviewer.org

ChristosT:
Even though this does not work as is in nbviewer it resolves 
- https://github.com/Kitware/trame-jupyter-extension/issues/5
- https://github.com/ansys/pyansys-geometry/pull/826

Can we merge it as is ? I believe that it is still a significant improvement over the previous state. Let me know if you need any help.

banesullivan:
I agree we should merge as is. Though we should probably review and cleanup the code here as needed

banesullivan:
Turns out this requires a setting in JupyterLab and it works!

Screenshot 2023-11-17 at 8 36 54 PM

ChristosT:
Should we also add a tinypages test ? I have one here https://github.com/pyvista/pyvista/compare/feat/embeddable-widget...ChristosT:pyvista:feat/embeddable-widget .

banesullivan:
@ChristosT, thanks for that! I went ahead and merged your diff into this PR. Further, I added you to `@pyvista/developers` so you can push those suggestions directly to these branches/PRs in pyvista moving forward

francesco-ballarin:
I tested this by applying the diff in this PR on top of the most recent release (rather than main), and I am happy to report that this works with both Google Colab and Kaggle.

Test notebooks (Colab first, Kaggle second)
https://colab.research.google.com/github/fem-on-colab/fem-on-colab.github.io/blob/gh-pages/tests/vtk/test-pyvista.ipynb
https://kaggle.com/kernels/welcome?src=https://github.com/fem-on-kaggle/fem-on-kaggle.github.io/blob/gh-pages/tests/vtk/test-pyvista.ipynb


Screenshots (Colab first, Kaggle second)
Screenshot from 2023-11-21 11-03-05
Screenshot from 2023-11-21 11-11-44

A few questions, apologies if they are not fully related to the content of this PR:
1. scrollbars appear in both screenshots above. Can the default width and height be tweaked a bit to avoid this?
2. I noticed that, upon interacting with the plot, the style of the axes widget changes, see screenshots below
Screenshot from 2023-11-21 11-03-28
Screenshot from 2023-11-21 11-12-00
3. is starting `xvfb` still needed with trame, or not? You do not see it in the notebooks because I hardcode it in my build scripts https://github.com/fem-on-colab/fem-on-colab/blob/main/vtk/patches/01-start-xvfb


Overall, especially for my basic needs,  this PR  would be enough to close #3851

francesco-ballarin:
It would be great if this could make it into #5138.

francesco-ballarin:
> I tested this by applying the diff in this PR on top of the most recent release (rather than main), and I am happy to report that this works with both Google Colab and Kaggle.
> 
> Test notebooks (Colab first, Kaggle second)
> https://colab.research.google.com/github/fem-on-colab/fem-on-colab.github.io/blob/gh-pages/tests/vtk/test-pyvista.ipynb
> https://kaggle.com/kernels/welcome?src=https://github.com/fem-on-kaggle/fem-on-kaggle.github.io/blob/gh-pages/tests/vtk/test-pyvista.ipynb

I now downloaded one of those two test notebooks, and played with it locally.

Is
```
jupyter nbconvert --to html --execute test_pyvista.ipynb
```
supposed to work?

I get the following screenshot
Screenshot from 2023-11-21 13-33-08
and a warning which is probably relevant
```
/usr/local/lib/python3.11/dist-packages/nbconvert/filters/widgetsdatatypefilter.py:72: UserWarning: Your element with mimetype(s) dict_keys([]) is not able to be represented.
```

banesullivan:
> scrollbars appear in both screenshots above. Can the default width and height be tweaked a bit to avoid this?

I'll see if I can reproduce this and fix

> I noticed that, upon interacting with the plot, the style of the axes widget changes, see screenshots below

That's odd but an artifact coming from vtk.js/trame-vtk. Will not be fixed here

> is starting xvfb still needed with trame, or not? You do not see it in the notebooks because I hardcode it in my build scripts https://github.com/fem-on-colab/fem-on-colab/blob/main/vtk/patches/01-start-xvfb

It shouldn't be required but I will confirm in a headless docker container with this.

> It would be great if this could make it into https://github.com/pyvista/pyvista/pull/5138.

Added!

> nbconvert ... supposed to work?

Yes! I didn't test this but will now because it should work

francesco-ballarin:
@mobigroup can you test this on your notebooks on Colab?

The patch in https://patch-diff.githubusercontent.com/raw/pyvista/pyvista/pull/5168.diff applies cleanly on top of the most recent `pyvista` release. Just `pip install pyvista` in your notebooks, `cd` to the installation directory, apply the patch. Just make sure that you apply the patch before the first `import pyvista` in the notebook, otherwise the unpatched version will get imported.

banesullivan:
> scrollbars appear in both screenshots above. Can the default width and height be tweaked a bit to avoid this?

I cannot fix the styling in our widget to fix this. It seems like an issue coming from the upstream VTK.js viewer. 

> nbconvert ... supposed to work?

I can't get this to work, actually. Considering that the goal of this PR is just to get something working for easily embedding in the documentation and perhaps colab, I'm going to call it on this PR and let any other fixes be followup PRs.

Going to fix the tests then merge for the upcoming release

banesullivan:
@ChristosT, there was all kinds of weirdness happening in the tinypages test that I do not have time to debug, so I removed those tests. I'm going to consider the smoketest in `tests/plotting/jupyter/test_trame.py` sufficient for this PR so that we don't delay the release

banesullivan:
> do you need us to open issues on these tasks as a reminder?

Up to you! I personally don't think we need to take this any further and I think letting folks open issues as they arise is okay. If you're keen on fixing some of the points you brought up, feel free to track these in new issues

tkoyama010 (approved):
LGTM

ekwessel:
I'm running on a notebook system provided by an HPC cluster at my university. They do not allow notebook extensions, so being able to use these little embedded html scenes is very helpful.

However, I am encountering an unusual issue with the 'html' backend I haven't seen discussed elsewhere. When I call:
`static_plotter.show(jupyter_backend="html", return_viewer=True)`
Nothing appears in the notebook.
If I print this object, I can see it is an EmbeddableWidget object with the value parameter set to an HTML scene, but for some reason IPython's display() function, which is implicitly called, doesn't seem know what to do with this.

I figured out that I can get around this by creating a wrapper that has the _repr_html_ property expected by display:
`scene = static_plotter.show(jupyter_backend="html", return_viewer=True)
scene_wrapper = types.SimpleNamespace(_repr_html_=lambda : scene.value)`
Then, if I execute:
`static_wrapper`
The interactive HTML does appear!

As happy as I am to have figured out this work-around, this is surprisingly hack-y for something so simple. I am wondering why the display() routine doesn't just work on the pure EmbeddableWidget object like it normally does. Could this have something to do with the version of jupyter used by my HPC center? Or is there some magic configuration that needs to be run?

jourdain:
The html export is an asynchronous task which probably through the system off here... Try if you ensure that the trame server is actually running before you trigger the show. (may still not work since the export is still async and may not properly trigger a display call.

```python
from pyvista.trame.views import get_server

server = get_server(name=pyvista.global_theme.trame.jupyter_server_name)

def ready(server):
    if not server.running:
        # NoOp if already start[ing/ed]
        server.start(
            exec_mode="task",
            port=0,
            open_browser=False,
            show_connection_info=False,
            disable_logging=True,
            timeout=0,
        )
    return server.ready

await ready(server)
```

ekwessel:
I just gave that a try and it returns "True", which I assume means the trame server is running. However, two odd things happen. If I try to show a plot with the html backend before I run the above cell, which presumably starts the trame server, then there is no difference: the EmbeddableWidget still isn't rendered, but I can extract the HTML string and manually get it to be rendered in the notebook with the approach above.
However, if I run the above code that starts the server *before* I attempt to show the plot with the html backend, I get an error I've never seen before!
The show() function fails, and within it what seems to fail is the EmbeddableWidget initializer, which fails because when export_html is called on the plotter, it tries to call write_html(data, buffer), but apparently the "data" is a None.
Here's the full traceback I get:
```
TypeError                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 static_plotter.show(jupyter_backend="html", return_viewer=True)

File ~/virtual_environments/simple_pyvista/lib/python3.11/site-packages/pyvista/plotting/plotter.py:6937, in Plotter.show(self, title, window_size, interactive, auto_close, interactive_update, full_screen, screenshot, return_img, cpos, jupyter_backend, return_viewer, return_cpos, before_close_callback, **kwargs)
   6934         jupyter_backend = self._theme.jupyter_backend
   6936     if jupyter_backend.lower() != 'none':
-> 6937         jupyter_disp = handle_plotter(self, backend=jupyter_backend, **jupyter_kwargs)
   6939 self.render()
   6941 # initial double render needed for certain passes when offscreen

File ~/virtual_environments/simple_pyvista/lib/python3.11/site-packages/pyvista/jupyter/notebook.py:34, in handle_plotter(plotter, backend, screenshot, **kwargs)
     31     if backend in ['server', 'client', 'trame', 'html']:
     32         from pyvista.trame.jupyter import show_trame
---> 34         return show_trame(plotter, mode=backend, **kwargs)
     36 except ImportError as e:
     37     warnings.warn(
     38         f'Failed to use notebook backend: \n\n{e}\n\nFalling back to a static output.',
     39     )

File ~/virtual_environments/simple_pyvista/lib/python3.11/site-packages/pyvista/trame/jupyter.py:372, in show_trame(plotter, mode, name, server_proxy_enabled, server_proxy_prefix, jupyter_extension_enabled, collapse_menu, add_menu, add_menu_items, default_server_rendering, handler, **kwargs)
    369 kwargs.setdefault('height', dh)
    371 if mode == 'html':
--> 372     return EmbeddableWidget(plotter, **kwargs)
    374 if jupyter_extension_enabled is None:
    375     jupyter_extension_enabled = pyvista.global_theme.trame.jupyter_extension_enabled

File ~/virtual_environments/simple_pyvista/lib/python3.11/site-packages/pyvista/trame/jupyter.py:120, in EmbeddableWidget.__init__(self, plotter, width, height, **kwargs)
    118 if HTML is object:
    119     raise ImportError('Please install `ipywidgets`.')
--> 120 scene = plotter.export_html(filename=None)
    121 src = scene.getvalue().replace('"', '&quot;')
    122 # eventually we could maybe expose this, but for now make sure we're at least
    123 # consistent with matplotlib's color (light gray)

File ~/virtual_environments/simple_pyvista/lib/python3.11/site-packages/pyvista/plotting/plotter.py:640, in BasePlotter.export_html(self, filename)
    638 data = self.export_vtksz(filename=None)
    639 buffer = io.StringIO()
--> 640 write_html(data, buffer)
    641 buffer.seek(0)
    643 if filename is None:

File ~/virtual_environments/simple_pyvista/lib/python3.11/site-packages/trame_vtk/tools/vtksz2html.py:16, in write_html(data, output)
     15 def write_html(data: bytes, output: TextIO):
---> 16     base64Content = data_to_base64(data)
     17     with open(HTML_VIEWER_PATH, mode="r", encoding="utf-8") as srcHtml:
     18         for line in srcHtml:

File ~/virtual_environments/simple_pyvista/lib/python3.11/site-packages/trame_vtk/tools/vtksz2html.py:11, in data_to_base64(data)
     10 def data_to_base64(data: bytes):
---> 11     base64Content = base64.b64encode(data)
     12     return base64Content.decode().replace("\n", "")

File /opt/ohpc/pub/apps/python/3.11.4/lib/python3.11/base64.py:58, in b64encode(s, altchars)
     51 def b64encode(s, altchars=None):
     52     """Encode the bytes-like object s using Base64 and return a bytes object.
     53 
     54     Optional altchars should be a byte string of length 2 which specifies an
     55     alternative alphabet for the '+' and '/' characters.  This allows an
     56     application to e.g. generate url or filesystem safe Base64 strings.
     57     """
---> 58     encoded = binascii.b2a_base64(s, newline=False)
     59     if altchars is not None:
     60         assert len(altchars) == 2, repr(altchars)

TypeError: a bytes-like object is required, not 'NoneType'
```

PR #3004, opened 2022-07-16
Specify in CONTRIBUTING.rst use of NumPy docstrings

ENIAC-6:
Explicitly state use of the NumPy style for docstrings.

### Overview

Specify use of NumPy style for docstrings

Issue #2913

Resolves issue #2913

### Details

Added documentation from: [Docstrings](https://numpy.org/doc/stable/dev/howto-docs.html#docstring-intro) section in NumPy project.

PR #5171, opened 2023-11-02
Add more typehints

akaszynski:
This PR adds a variety of additional typehints throughout the library where they're needed. One day, it would be great if we were able to use `--disallow-untyped-calls`, but that might be a long type off...


MatthewFlamm:
This PR is specifically for typing and aiming at having a complete typing for the package, so I'd like to bring up my unease about continuing with the status quo in typing.  The typing is very heterogeneous in pyvista currently.  If we continue this heterogeneous approach, it will be hard to achieve this goal if we do not try to align on best practices as we add type hints like this.

For one reference on recommendations for typing see https://typing.readthedocs.io/en/latest/source/libraries.html#best-practices-for-inlined-types.  This section starts with the recommendation that function parameter inputs should be typed as wide as possible.  For example, this PR uses `List[int]` or similar in lots of places, but this should probably be `Sequence[int]` in most cases.  Also in most of those cases, it would also be acceptable to get `np.ndarray`.

We have a typing module, and although it has gotten better, it still isn't widely enough typed to be able to use everywhere in the codebase. https://github.com/pyvista/pyvista/blob/main/pyvista/core/_typing_core.py.  For example `Vector` is limited to 3-size `Tuple`.  So it cannot be used for 2-size `Tuple`.

I would prefer we simplify to the widest possible types, which would also greatly simplify the maintenance burden.  It seems we are trying to often use the type hints to also demonstrate the size of the sequence, which creates lots of problems.  For example, np.ndarray does not support this and List and Tuple are different in this regard.  I propose that we align the typing in https://github.com/pyvista/pyvista/blob/main/pyvista/core/_typing_core.py module so that it can be used in the most number of functions possible.  Right now it is problematic to use broadly becauase of not typing in the widest possible manner.  I'd propose these widest possible types as a starting point.  I think a vast majority of the typing here could be replaced by one of these below.  Even though I included `IntVector` and `FloatVector` because in some cases there might be a specific need, we need to be careful not to use them when `Vector` could be used instead.

```py
Number = Union[int, float]
Vector = Union[np.ndarray, Sequence[Number]
IntVector = Union[np.ndarray, Sequence[int]]
FloatVector = Union[np.ndarray, Sequence[float]]
Matrix = Union[np.ndarray, Sequence[Sequence[Number], np.ndarray]]
IntMatrix = Union[np.ndarray, Sequence[Sequence[int], np.ndarray]]
FloatMatrix = Union[np.ndarray, Sequence[Sequence[float], np.ndarray]]
```

I know @user27182 has recently worked on the typing in this module, so I think their thoughts would be relevant here.

akaszynski:
>  I propose that we align the typing in https://github.com/pyvista/pyvista/blob/main/pyvista/core/_typing_core.py module

Fully agree. When going through and adding type hints, I discovered that we indeed are inconsistent in our typehints (`List[float]` vs `Sequence[float] where `Vector[float]` would be ideal). In fact, as you suggested, we should actually be using:
```
Number = Union[int, float]
Vector = Union[np.ndarray, Sequence[Number]
```

for most cases that accept a sequence of `int` or `float` values.

---

I don't think I'll be able to refactor it all at once in this PR, but I do agree that typing needs to be overhauled. I'll be able to work on a handful of files, potentially the ones I've already touched in this PR.

akaszynski:
Encountering a severe issue where we use different types in our setters and getters. There's a conversation about this in:
https://github.com/python/mypy/issues/3004

I can see the point on both sides. Whenever we set points using a sequence (for example, a list of lists):
```
pd.points = [[1,2,3], [1,2,3]]
```
We of course convert this to an array out of convenience for the user:

```
>>> pd.points
pyvista_ndarray([[0, 0, 0],
                 [1, 2, 3]])
```

However, `mypy` sees this as an an improper assignment:

```
Incompatible types in assignment (expression has type "list[list[int]]", variable has type "pyvista_ndarray")  [assignment]
```

At the moment, there isn't a workaround for this, and [mypy #3004](https://github.com/python/mypy/issues/3004) is 6 years old. However, given that typescript 5.1 had this restriction removed, there might be hope for this yet.

MatthewFlamm:
I like your proposal of adding in a specific ignore flag.  Completely ignoring type is not a good solution.  But it might be what we have to do.

pyvista/core/_typing_core.py
```diff
@@ -1,13 +1,26 @@
...
 import numpy as np
+import numpy.typing as npt
 
 from pyvista.core._vtk_core import vtkMatrix3x3, vtkMatrix4x4, vtkTransform
 
-Vector = Union[List[float], Tuple[float, float, float], np.ndarray]
-VectorArray = Union[np.ndarray, Sequence[Vector]]
-Number = Union[float, int, np.number]
-NumericArray = Union[Sequence[Number], np.ndarray]
+NumpyNumArray = npt.NDArray[np.number]
```
MatthewFlamm, line 9: Unless we want to pin numpy version >=1.21, we cannot use `NDArray`. See https://numpy.org/doc/stable/reference/typing.html#numpy.typing.NDArray

pyvista/core/utilities/helpers.py
```diff
@@ -10,7 +15,9 @@
 from .fileio import from_meshio, is_meshio_mesh
 
 
-def wrap(dataset):
+def wrap(
+    dataset: Union[np.ndarray, _vtk.vtkDataSet, 'Trimesh', 'Mesh']
+) -> Optional[Union['pyvista.DataSet', 'pyvista.pyvista_ndarray']]:
```
MatthewFlamm, line 20: Can we use an overload to distinguish between `DataSet` return and `pyvista_ndarray` return?

pyvista/core/_typing_core.py
akaszynski: That will have to be for everyone to decide, but I'd be in favor of bumping it.

pyvista/core/utilities/helpers.py
akaszynski: I can't for the life of me get overloading to work.

This seems like it should work fine:

```py
@overload
def wrap(dataset: np.ndarray) -> Optional['pyvista.pyvista_ndarray']:
    ...

@overload
def wrap(dataset: Union[_vtk.vtkDataSet, 'Trimesh', 'Mesh']) -> Optional['pyvista.DataSet']:
    ...


def wrap(
    dataset: Union[np.ndarray, _vtk.vtkDataSet, 'Trimesh', 'Mesh']
) -> Union[Optional['pyvista.DataSet'], Optional['pyvista.pyvista_ndarray']]:
```

And yet:
```
pyvista/core/utilities/helpers.py:18: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]        
```

darikg: Maybe the `dataset` in the actual implementation should be left untyped?

Also, you could get rid of the `Optional` return type as well with another overload, right?

```python
@overload
def wrap(dataset: None) -> None: ...
```

MatthewFlamm: One issue is that you shouldn't type the actual function, this causes the duplicate.  For example:

```py
@overload
def wrap(dataset: np.ndarray) -> Optional['pyvista.pyvista_ndarray']:
    ...

@overload
def wrap(dataset: Union[_vtk.vtkDataSet, 'Trimesh', 'Mesh']) -> Optional['pyvista.DataSet']:
    ...


def wrap(dataset):
    ...
```

MatthewFlamm: See a few examples down from here https://peps.python.org/pep-0484/#function-method-overloading

pyvista/core/_typing_core.py
MatthewFlamm: I'm +1 to pin to >=1.21.

pyvista/core/utilities/helpers.py
MatthewFlamm: I tried @darikg suggestion too and that didn't work.

Also we are missing the possibility that `None` is passed into the function.  I'm surprised that the type checker doesn't complain about this now and this seems like a clue to this problem.

MatthewFlamm: Found it. The type of `_vtk.vtkDataSet` is `Any`.  vtk doesn't ship with type hints, although they can be autogenerated.

pyvista/core/utilities/helpers.py
```diff
@@ -10,7 +15,9 @@
 from .fileio import from_meshio, is_meshio_mesh
 
 
-def wrap(dataset):
+def wrap(
+    dataset: Union[np.ndarray, _vtk.vtkDataSet, 'Trimesh', 'Mesh']
```
MatthewFlamm, line 19:
```suggestion
    dataset: Optional[Union[np.ndarray, _vtk.vtkDataSet, 'Trimesh', 'Mesh']]
```

pyvista/core/utilities/helpers.py
MatthewFlamm: We are not going to be able to use overload unless this can be addressed.

darikg: Interesting! I gave it a shot. `try_vtk_mypy.py`:

```python
from vtkmodules.vtkCommonDataModel import vtkDataSet

reveal_type(vtkDataSet())
```

At the command line, `mypy try_vtk_mypy.py` says `try_mypy.py:3: note: Revealed type is "Any"`

Autogenerate the stubs: `stubgen --package vtkmodules --output stubs` and `set MYPYPATH=./stubs`, then  `mypy try_vtk_mypy.py` again:

```sh
stubs\vtkmodules\vtkCommonDataModel.pyi:4494: error: invalid syntax  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

Ok, manually fix some minor syntax errors in `vtkCommonDataModel.pyi`, and after the second and third tries some duplicate function argument names in `vtkCommonCore.pyi` and `vtkCommonMath.pyi`. Then run `mypy try_vtk_mypy.py` again:

```sh
(Thousands of warnings and errors omitted)
try_mypy.py:3: note: Revealed type is "vtkmodules.vtkCommonDataModel.vtkDataSet"
Found 6123 errors in 4 files (checked 1 source file)
```

It works! But there's other 6123 errors to patch up, haha

MatthewFlamm: I don't think we can ship typing that requires the user to modify their vtk site-packages.  I think it would require vtk to ship the type stubs.

darikg: yeah, or manage a separate stubs repo required as a dependency, which seems like a nightmare

akaszynski: > yeah, or manage a separate stubs repo required as a dependency, which seems like a nightmare

I think it should be the responsibility of Kitware to do that. Typing for pyvista alone is enough work.

akaszynski:
@pyvista/developers, we're going to pin `numpy>=1.21` to support type hints using [numpy.typing.NDArray](https://numpy.org/doc/stable/reference/typing.html#numpy.typing.NDArray).

Please let me know if there are any objections. Given that 1.21 it supports the oldest maintained version of Python (actually, Python 3.7 as well), I think it's safe to pin it.

pyvista/core/composite.py
```diff
@@ -432,8 +432,11 @@ def append(self, dataset: Optional[_TypeMultiBlockLeaf], name: Optional[str] = N
             raise ValueError("Cannot nest a composite dataset in itself.")
 
         index = self.n_blocks  # note off by one so use as index
-        # always wrap since we may need to reference the VTK memory address
-        dataset = wrap(dataset)
+        # always wrap since we may need to reference the VTK memory address]
```
darikg, line 435: extraneous square bracket added here

adeak:
I agree with @MatthewFlamm's design remarks.

NumPy 1.21 was tagged more than 2 years ago, I think it's safe to pin that (whoever is stuck with old NumPy will be stuck with 2 years younger PyVista which I think is fair).

pyvista/core/dataset.py
```diff
@@ -1315,16 +1345,16 @@ def flip_z(self, point=None, transform_all_input_vectors=False, inplace=False):
 
         Parameters
         ----------
-        point : list, optional
+        point : sequence[float], optional
```
darikg, line 1348: numpydoc edits earlier in the file used point: FloatVector, here it's sequence[float]

pyvista/core/dataset.py
```diff
@@ -1198,12 +1213,17 @@ def scale(
...
-        transform.Scale(xyz)
+        transform.Scale(xyz)  # type: ignore
         return self.transform(
             transform, transform_all_input_vectors=transform_all_input_vectors, inplace=inplace
         )
 
-    def flip_x(self, point=None, transform_all_input_vectors=False, inplace=False):
+    def flip_x(
+        self,
+        point: Optional[Vector] = None,
```
darikg, line 1223: methods earlier in the file with a `point` argument took FloatVector, here's it's just Vector

pyvista/core/dataset.py
```diff
@@ -3208,9 +3259,7 @@ def point_cell_ids(self, ind: int) -> List[int]:
         self.GetPointCells(ind, ids)
         return [ids.GetId(i) for i in range(ids.GetNumberOfIds())]
 
-    def point_is_inside_cell(
-        self, ind: int, point: Union[VectorArray, NumericArray]
-    ) -> Union[int, np.ndarray]:
+    def point_is_inside_cell(self, ind: int, point: FloatMatrix) -> Union[int, np.ndarray]:
```
darikg, line 3262: point: FloatVector | FloatMatrix

pyvista/core/pointset.py
```diff
@@ -159,21 +170,19 @@ def points_to_double(self):
...
 
         Parameters
         ----------
-        xyz : array_like[float]
-            Cartesian values to displace with. Length 3 array-like.
+        xyz : Vector
+            A vector of three floats of cartesian values to translate the mesh with.
 
-        transform_all_input_vectors : bool, default: False
+        Transform_all_input_vectors : bool, default: False
```
darikg, line 185: shouldn't be capitalized

darikg (commented):
I wonder if the FloatMatrix, FloatVector etc. are really the right type. Most methods that want them will happily accept arrays of int.

pyvista/core/dataset.py
```diff
@@ -2595,15 +2639,22 @@ def find_cells_intersecting_line(
...
+            cast(Sequence[float], pointb),
+            tolerance,
+            points,
+            id_list,
+            cell,
+        )
         return vtk_id_list_to_array(id_list)
 
-    def find_cells_within_bounds(self, bounds: Iterable[float]) -> np.ndarray:
+    def find_cells_within_bounds(self, bounds: Vector) -> np.ndarray:
```
darikg, line 2657: bounds: BoundsLike

pyvista/core/dataset.py
```diff
@@ -2636,7 +2687,7 @@ def find_cells_within_bounds(self, bounds: Iterable[float]) -> np.ndarray:
         if np.array(bounds).size != 6:
             raise TypeError("Bounds must be a length three tuple of floats.")
```
darikg, line 2688: must be a length six

akaszynski:
```suggestion
            raise TypeError("Bounds must be a length six tuple of floats.")
```

akaszynski: Great suggestions/review @darikg! Feel free to make inline comments if you want. You'll get "credit" on the PR if you care about that.

pyvista/core/dataset.py
akaszynski:
```suggestion
    def find_cells_within_bounds(self, bounds: BoundsLike) -> np.ndarray:
```

pyvista/core/pointset.py
akaszynski:
```suggestion
        transform_all_input_vectors : bool, default: False
```

pyvista/core/dataset.py
akaszynski:
```suggestion
    def point_is_inside_cell(self, ind: int, point: FloatVector | FloatMatrix) -> Union[int, np.ndarray]:
```

pyvista/core/dataset.py
akaszynski:
```suggestion
        point: Optional[FloatVector] = None,
```

pyvista/core/dataset.py
akaszynski:
```suggestion
        point : FloatVector, optional
```

pyvista/core/composite.py
akaszynski:
```suggestion
        # always wrap since we may need to reference the VTK memory address
```

akaszynski:
> I wonder if the FloatMatrix, FloatVector etc. are really the right type. Most methods that want them will happily accept arrays of int.

For some of these methods, that's correct, but not for all methods. For example in [vtkTransform](https://vtk.org/doc/nightly/html/classvtkTransform.html), most of the methods here are typed for floats, though they will (at least within python) accept ints. In fact, for C++, provided you're not passing by reference, variables will implicitly be converted to the correct type (sometimes with a warning).

I'd prefer to stay aligned a much as possible with VTK's types, which, aside from generating types on our own, means manually interpreting the docs. However, I'm not dead set on this and I'm up for whatever solution makes the most sense.

pyvista/core/datasetattributes.py
```diff
@@ -1,16 +1,19 @@
...
 import numpy as np
 
 from . import _vtk_core as _vtk
-from ._typing_core import Number
+from ._typing_core import Array, Matrix
 from .errors import PyVistaDeprecationWarning
 from .pyvista_ndarray import pyvista_ndarray
 from .utilities.arrays import FieldAssociation, convert_array, copy_vtk_array
 
+if TYPE_CHECKING:
```
akaszynski, line 14:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

pyvista/core/utilities/helpers.py
```diff
@@ -1,5 +1,10 @@
 """Core helper utilities."""
 import collections
+from typing import TYPE_CHECKING, Optional, Union, cast
+
+if TYPE_CHECKING:
```
akaszynski, line 5:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

MatthewFlamm:
> > I wonder if the FloatMatrix, FloatVector etc. are really the right type. Most methods that want them will happily accept arrays of int.
> 
> For some of these methods, that's correct, but not for all methods. For example in [vtkTransform](https://vtk.org/doc/nightly/html/classvtkTransform.html), most of the methods here are typed for floats, though they will (at least within python) accept ints. In fact, for C++, provided you're not passing by reference, variables will implicitly be converted to the correct type (sometimes with a warning).
> 
> I'd prefer to stay aligned a much as possible with VTK's types, which, aside from generating types on our own, means manually interpreting the docs. However, I'm not dead set on this and I'm up for whatever solution makes the most sense.

Tough one here as it is near impossible to get this right everywhere without having the typing from vtk.  Based on my comment above that we should use widest type, we should use `Vector` instead of `FloatVector` whenever we know it can auto cast the value.  But this is easier said in theory than practice.  I'm +1 for making a best guess based on vtk docs.  If anyone runs into typing problems when trying to use int, we can relax typing.

akaszynski:
> I'm +1 for making a best guess based on vtk docs

Would appreciate if anyone from Kitware would like to weigh in. @ChristosT?

akaszynski:
> we should use Vector instead of FloatVector whenever we know it can auto cast the value

I'm going with this unless there are any objections. This will avoid errors with simple operations like:

```py
mesh.translate([1,1,1])
```

It's fine in C++, it will be fine here.

darikg:
Another minor concern: do type aliases like `Matrix` work correctly in numpydocs? I feel like things like 

`points : Union[Matrix, _vtk.vtkPoints]` might need to be changed to `points : Union[pyvista.core._typing_core.Matrix, _vtk.vtkPoints]`  which becomes unfortunately verbose

akaszynski:
> Another minor concern: do type aliases like `Matrix` work correctly in numpydocs? I feel like things like
> 
> `points : Union[Matrix, _vtk.vtkPoints]` might need to be changed to `points : Union[pyvista.core._typing_core.Matrix, _vtk.vtkPoints]` which becomes unfortunately verbose

Let's address this in a followup PR.

I'd like to merge this (any any other easier PRs) before the virtual hackathon.

banesullivan (approved):
LGTM! This is tedious yet important work! Thanks for being meticulous here!

ChristosT:
> Would appreciate if anyone from Kitware would like to weigh in. @ChristosT?

It looks great overall ! Regarding the VTK objects and types I don't see any obvious issue.

banesullivan (approved):

Issue #5170, opened 2023-11-02
SurfaceNets to visualize 3D label maps


jmargeta:
### Describe the feature you would like to be added.

Hi folks, I often a need to visualize semantic label maps from 3D medical images.

These are regular grids, where each voxel of the label map is a discrete label that usually represents some tissue.  
Image thresholding followed by surface extraction (as discussed [here](https://github.com/pyvista/pyvista/discussions/2434)) works ok, but is extremely slow for many labels. Moreover, as each label is handled separately, there are no guarantees for consistency of the generated contours.

![image](https://github.com/pyvista/pyvista/assets/1239359/4f227944-5589-4b0a-984f-cede2b2cb764)

[SurfaceNets by S. Frisken](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9623606/) solve these issues. They are superfast (easily a few orders of magnitude faster for my use case), look better and are much more accurate.

Not long ago, they have been implemented in VTK and are now available in a pre-release Python version. Although wrapping this is not too hard and having an extra function is fine too, supporting this in PyVista out of the box would be fantastic.

I think that this fits the best into uniform_grid filters. However, I am not sure about keeping the original name, SurfaceNets.

Everyone should be able to visualize segmentation maps, parcelations, multi-label machine learning ground-truth and outputs, etc.

What do you think?

### Links to VTK Documentation, Examples, or Class Definitions.

[Blog post about this being in VTK](https://www.kitware.com/really-fast-isocontouring/)
[My gist with a simple functional wrapper](https://gist.github.com/jmargeta/1e8d383fe121458a1c5ecf6a109c82cc/) and [X post with some more examples](https://twitter.com/jmargeta/status/1719489527541100931)
[Slicer issue about integrating this](https://github.com/Slicer/Slicer/issues/7103)
[vktSurfaceNets3d class definition](https://vtk.org/doc/nightly/html/classvtkSurfaceNets3D.html)
[Frisken SF. SurfaceNets for Multi-Label Segmentations with Preservation of Sharp Boundaries. J Comput Graph Tech. 2022 Jan-Mar;11(1):34-54. Epub 2022 Feb 28. PMID: 36325473; PMCID: PMC9623606.](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9623606/)

### Pseudocode or Screenshots

[...]

PR #5176, opened 2023-11-05
Add conversion of 3D label maps to labeled meshes with SurfaceNets.

jmargeta:
### Overview
This PR adds support for conversion of 3D label maps to labeled surfaces for colored visualization using vtkSurfaceNets3D algorithm from VTK. This is particularly useful for visualization of anatomical structures in medical images from their labels or the outputs of multi-label segmentation maps.

Resolves https://github.com/pyvista/pyvista/issues/5170

### Details
- added contour_labeled to ImageData -  contours can be extracted with `label_map.contour_labeled()`
- added example into the gallery with the frog tissue labels and a simple test for the number of extracted labels in the output mesh
image
- Note, vtk 9.3+ must be used which is currently in the pre-release stage on PyPI. Otherwise, the support with throw an exception. The example will, however, not be generated, and the test will be blank.

As mentioned in the issue, another appealing example could be made using the great [Neuroparc project](https://github.com/neurodata/neuroparc). Is it acceptable to make an example with an external dataset?
```python
import pooch
import pyvista as pv
file_name = pooch.retrieve(
    "https://github.com/neurodata/neuroparc/raw/master/atlases/label/Human/Schaefer1000_space-MNI152NLin6_res-1x1x1.nii.gz",
    known_hash="c3efe797aab3b3d9e705645bf29fac4e932c88dbe54ccaeb03982f11e66b3249",
)
label_map = pv.read(file_name)
label_map.contour_labeled(smoothing=True).plot(cmap="glasbey_bw")
```
image

Does the name `contour_labeled` fit well with PyVista, or would you suggest a better one?

Also, I added types to the function, and had to silence these Mypy errors (as in https://github.com/pyvista/pyvista/pull/3837): 
```
pyvista/core/filters/image_data.py:853: error: Argument 1 to "set_default_active_scalars" has incompatible type "ImageDataFilters"; expected "DataSet"  [arg-type]
pyvista/core/filters/image_data.py:854: error: "ImageDataFilters" has no attribute "active_scalars_info"  [attr-defined]
pyvista/core/filters/image_data.py:858: error: "ImageDataFilters" has no attribute "get_array_association"  [attr-defined]
```

More info about the algorithm: Sarah F. Frisken, SurfaceNets for Multi-Label Segmentations with Preservation of Sharp Boundaries, Journal of Computer Graphics Techniques (JCGT), vol. 11, no. 1, 34-54, 2022. Available online http://jcgt.org/published/0011/01/03/

### DIFF
--- a/examples/01-filter/contouring.py
+++ b/examples/01-filter/contouring.py
@@ -74,3 +74,15 @@ pl.add_mesh(edges, line_width=5, render_lines_as_tubes=True, color='k')
 pl.add_mesh(contours, **dargs)
 pl.add_mesh(arrows, **dargs)
 pl.show()
+
+###############################################################################
+# Contours from a label map
+# +++++++++++++++++++++++++
+#
+# Create labeled surfaces from 3D label maps (e.f. multi-label image segmentation)
+# using :func:`contour_labeled() <pyvista.PolyDataFilters.contour_labeled>`.
+# Requires VTK version 9.3
+if pv.vtk_version_info >= (9, 3):
+    label_map = pv.examples.download_frog_tissue()
+    mesh = label_map.contour_labeled()
+    mesh.plot(cmap="glasbey_warm", cpos="yx", show_scalar_bar=False)
--- a/examples/01-filter/contouring.py
+++ b/examples/01-filter/contouring.py
@@ -80,7 +80,7 @@ pl.show()
 # +++++++++++++++++++++++++
 #
 # Create labeled surfaces from 3D label maps (e.f. multi-label image segmentation)
-# using :func:`contour_labeled() <pyvista.PolyDataFilters.contour_labeled>`.
+# using :func:`contour_labeled() <pyvista.ImageDataFilters.contour_labeled>`.
 # Requires VTK version 9.3
 if pv.vtk_version_info >= (9, 3):
     label_map = pv.examples.download_frog_tissue()
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -796,7 +796,6 @@ class ImageDataFilters(DataSetFilters):
         This filter requires that the :class:`ImageData` has integer point
         scalars, such as multi-label maps generated from image segmentation.
 
-
         Parameters
         ----------
         n_labels : int, optional
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -842,6 +842,7 @@ class ImageDataFilters(DataSetFilters):
         Examples
         --------
         See :ref:`contouring_example` for a full example using this filter.
+
         """
         if not hasattr(_vtk, 'vtkSurfaceNets3D'):  # pragma: no cover
             from pyvista.core.errors import VTKVersionError
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -847,7 +847,7 @@ class ImageDataFilters(DataSetFilters):
         if not hasattr(_vtk, 'vtkSurfaceNets3D'):  # pragma: no cover
             from pyvista.core.errors import VTKVersionError
 
-            raise VTKVersionError('Surface nets 3D require VTK 9.3.20230807rc0 or newer.')
+            raise VTKVersionError('Surface nets 3D require VTK 9.3.0 or newer.')
 
         alg = _vtk.vtkSurfaceNets3D()
         if scalars is None:
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -841,7 +841,7 @@ class ImageDataFilters(DataSetFilters):
 
         Examples
         --------
-        See :ref:`surface_nets_example` for a full example using this filter.
+        See :ref:`contouring_example` for a full example using this filter.
         """
         if not hasattr(_vtk, 'vtkSurfaceNets3D'):  # pragma: no cover
             from pyvista.core.errors import VTKVersionError
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -852,11 +852,11 @@ class ImageDataFilters(DataSetFilters):
         if scalars is None:
             set_default_active_scalars(self)  # type: ignore
             field, scalars = self.active_scalars_info  # type: ignore
-            if field.value != FieldAssociation.POINT:
+            if field != FieldAssociation.POINT:
                 raise ValueError('If `scalars` not given, active scalars must be point array.')
         else:
             field = self.get_array_association(scalars, preference='point')  # type: ignore
-            if field.value != FieldAssociation.POINT:
+            if field != FieldAssociation.POINT:
                 raise ValueError(
                     f'Can only process point data, given `scalars` are {field.name.lower()} data.'
                 )
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -1,6 +1,6 @@
 """Filters with a class to manage filters/algorithms for uniform grid datasets."""
 import collections.abc
-from typing import Optional
+from typing import Literal, Optional
 
 import numpy as np
 
@@ -781,8 +781,8 @@ class ImageDataFilters(DataSetFilters):
         smoothing_num_iterations: int = 50,
         smoothing_relaxation_factor: float = 0.5,
         smoothing_constraint_distance: float = 1,
-        output_mesh_type: str = 'quads',
-        output_style: str = 'default',
+        output_mesh_type: Literal['quads', 'triangles'] = 'quads',
+        output_style: Literal['default', 'boundary'] = 'default',
         scalars: Optional[str] = None,
         progress_bar: bool = False,
     ) -> 'pyvista.PolyData':
@@ -795,6 +795,9 @@ class ImageDataFilters(DataSetFilters):
 
         This filter requires that the :class:`ImageData` has integer point
         scalars, such as multi-label maps generated from image segmentation.
+
+        .. note::
+           Requires ``vtk>=9.3.0``.
 
         Parameters
         ----------
@@ -814,10 +817,14 @@ class ImageDataFilters(DataSetFilters):
             Constraint distance of the smoothing.
 
         output_mesh_type : str, default: 'quads'
-            Mesh type - triangles or quads.
+            Type of the output mesh. Must be either ``'quads'``, or ``'triangles'``.
 
         output_style : str, default: 'default'
-            Output style (default, boundary). "selected" is currently not implemented.
+            Style of the output mesh. Must be either ``'default'`` or ``'boundary'``.
+            When ``'default'`` is specified, the filter produces a mesh with both
+            interior and exterior polygons. When ``'boundary'`` is selected, only
+            polygons on the border with the background are produced (without interior
+            polygons). Note that style ``'selected'`` is currently not implemented.
 
         scalars : str, optional
             Name of scalars to process. Defaults to currently active scalars.
@@ -874,9 +881,10 @@ class ImageDataFilters(DataSetFilters):
             alg.SetOutputStyleToDefault()
         elif output_style == 'boundary':
             alg.SetOutputStyleToBoundary()
-        else:
-            alg.SetOutputStyleToSelected()
+        elif output_style == 'selected':
             raise NotImplementedError(f'Output style "{output_style}" is not implemented')
+        else:
+            raise ValueError(f'Invalud output style "{output_style}", use "default" or "boundary"')
         if smoothing:
             alg.SmoothingOn()
             alg.GetSmoother().SetNumberOfIterations(smoothing_num_iterations)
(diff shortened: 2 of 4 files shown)


user27182:
Awesome feature! Thanks for your contribution. This should be linked with #4802 in terms of PyVista VTK 9.3 support. `PackLabels` is used internally by `SurfaceNets`.

jmargeta:
Thanks, @user27182, for the pointer and for linking the two as it should be! Great to see the support for `PackLabels`.
Once your PR gets merged, I will be happy to adopt your vtk 9.3 detection logic.
Cannot wait to see the latest capabilities of vtk in PyVista.

MatthewFlamm:
vtk 9.3 is now in the CI, so it can be tested directly

pyvista/core/filters/image_data.py
```diff
@@ -772,3 +773,114 @@ def _flip_uniform(self, axis) -> 'pyvista.ImageData':
...
+        Sarah F. Frisken, SurfaceNets for Multi-Label Segmentations with Preservation
+        of Sharp Boundaries, Journal of Computer Graphics Techniques (JCGT), vol. 11,
+        no. 1, 34-54, 2022. Available online http://jcgt.org/published/0011/01/03/
+
+        https://www.kitware.com/really-fast-isocontouring/
+
+        Examples
+        --------
+        See :ref:`contouring_example` for a full example using this filter.
+        """
```
MatthewFlamm, line 845: Our standard is to have a space at the end.
```suggestion

        """
```

pyvista/core/filters/image_data.py
```diff
@@ -772,3 +773,114 @@ def _flip_uniform(self, axis) -> 'pyvista.ImageData':
...
+        if not hasattr(_vtk, 'vtkSurfaceNets3D'):  # pragma: no cover
+            from pyvista.core.errors import VTKVersionError
+
+            raise VTKVersionError('Surface nets 3D require VTK 9.3.20230807rc0 or newer.')
```
MatthewFlamm, line 849:
```suggestion
            raise VTKVersionError('Surface nets 3D require VTK 9.3.0 or newer.')
```

examples/01-filter/contouring.py
```diff
@@ -74,3 +74,15 @@
 pl.add_mesh(contours, **dargs)
 pl.add_mesh(arrows, **dargs)
 pl.show()
+
+###############################################################################
+# Contours from a label map
+# +++++++++++++++++++++++++
+#
+# Create labeled surfaces from 3D label maps (e.f. multi-label image segmentation)
+# using :func:`contour_labeled() <pyvista.PolyDataFilters.contour_labeled>`.
```
MatthewFlamm, line 83:
```suggestion
# using :func:`contour_labeled() <pyvista.ImageDataFilters.contour_labeled>`.
```

pyvista/core/filters/image_data.py
```diff
@@ -772,3 +773,114 @@ def _flip_uniform(self, axis) -> 'pyvista.ImageData':
...
+        """Generate labeled contours from 3D label maps.
+
+        SurfaceNets algorithm is used to extract contours preserving sharp
+        boundaries for the selected labels from the label maps.
+        Optionally, the boundaries can be smoothened to reduce the staircase
+        appearance in case of low resolution input label maps.
+
+        This filter requires that the :class:`ImageData` has integer point
+        scalars, such as multi-label maps generated from image segmentation.
+
```
MatthewFlamm, line 798:
```suggestion
```

pyvista/core/filters/image_data.py
```diff
@@ -772,3 +773,114 @@ def _flip_uniform(self, axis) -> 'pyvista.ImageData':
...
+        if not hasattr(_vtk, 'vtkSurfaceNets3D'):  # pragma: no cover
+            from pyvista.core.errors import VTKVersionError
+
+            raise VTKVersionError('Surface nets 3D require VTK 9.3.20230807rc0 or newer.')
+
+        alg = _vtk.vtkSurfaceNets3D()
+        if scalars is None:
+            set_default_active_scalars(self)  # type: ignore
+            field, scalars = self.active_scalars_info  # type: ignore
+            if field.value == 1:
```
MatthewFlamm, line 855: Use the enums from 

https://github.com/pyvista/pyvista/blob/94c484c6a6a54b86704168a54666323234d7fdf5/pyvista/core/utilities/arrays.py#L15-L21

pyvista/core/filters/image_data.py
```diff
@@ -772,3 +773,114 @@ def _flip_uniform(self, axis) -> 'pyvista.ImageData':
...
+                raise ValueError('If `scalars` not given, active scalars must be point array.')
+        else:
+            field = self.get_array_association(scalars, preference='point')  # type: ignore
+            if field.value == 1:
```
MatthewFlamm, line 859: same as above

pyvista/core/filters/image_data.py
MatthewFlamm: We should also use `field != FieldAssociation.POINT`, or similar.  It is probably certain that the only other possible value is cells here, but better to be safe and it is more understandable in the code.

jmargeta:
@MatthewFlamm, thanks for the update, that is great to hear!

I added extra tests to improve the coverage and benefit from vtk 9.3 in the CI; and consider this wrapper complete functionally-wise.

Any feedback on the PR and what else I can do here to make the API better, please?
```python
def contour_labeled(
    self,
    n_labels: Optional[int] = None,
    smoothing: bool = False,
    smoothing_num_iterations: int = 50,
    smoothing_relaxation_factor: float = 0.5,
    smoothing_constraint_distance: float = 1,
    output_mesh_type: str = 'quads',
    output_style: str = 'default',
    scalars: Optional[str] = None,
    progress_bar: bool = False,
) -> 'pyvista.PolyData'
```

In particular, does the `contour_labeled` name align well with the spirit of PyVista? I chose it primarily for easier discoverability under the contour«TAB»-completion.

More explicit alternatives could be something like  `segmentaton`, `label_contours`, `label_map`, `surface_nets`, `label`, `label_surfaces`, `extract_label_map`, `segmentation_contours`... 

@tkoyama010, any chance for this to get into the November Release (#5138)?

pyvista/core/filters/image_data.py
jmargeta: Thanks for highligting this enum for me, fully agree!

I copied the logic from gaussian_smooth and image_dilate_erode in the same file. Shall I update these two functions as well with the comparison against the enum for consistency (despite it being unrelated to this PR)?

pyvista/core/filters/image_data.py
jmargeta: Cool, I also updated the error to give the detail about the given type (rather than saying "cell data" no matter what is provided):
`f'Can only process point data, given `scalars` are {field.name.lower()} data.'`

jmargeta:
@MatthewFlamm, thanks a lot for the update on the CI and your review!
I believe to have addressed all your comments and consider the  wrapper complete functionally-wise.

Any feedback on what else I can do here to make the API better, please? In particular, does the `contour_labeled` name align well with the spirit of PyVista? I chose it primarily for easier discoverability under the contour«TAB»-completion.
```python
def contour_labeled(
    self,
    n_labels: Optional[int] = None,
    smoothing: bool = False,
    smoothing_num_iterations: int = 50,
    smoothing_relaxation_factor: float = 0.5,
    smoothing_constraint_distance: float = 1,
    output_mesh_type: str = 'quads',
    output_style: str = 'default',
    scalars: Optional[str] = None,
    progress_bar: bool = False,
) -> 'pyvista.PolyData'
```

More explicit alternatives could be something like  `segmentaton`, `label_contours`, `label_map`, `surface_nets`, `label`, `label_surfaces`, `extract_label_map`, `segmentation_contours`... 

@tkoyama010, any chance for this to get into the November Release (#5138)?

pyvista/core/filters/image_data.py
jmargeta: This is indeed ballooning quite a bit (9 test failures). I will not add it to this PR.

tkoyama010:
@jmargeta I added this PR to the waiting list in the release. Let's add this to the next release!

jmargeta:
Awesome, thanks @tkoyama010 !

pyvista/core/filters/image_data.py
```diff
@@ -772,3 +773,116 @@ def _flip_uniform(self, axis) -> 'pyvista.ImageData':
...
+            Number of smoothing iterations.
+
+        smoothing_relaxation_factor : float, default: 0.5
+            Relaxation factor of the smoothing.
+
+        smoothing_constraint_distance : float, default: 1
+            Constraint distance of the smoothing.
+
+        output_mesh_type : str, default: 'quads'
+            Mesh type - triangles or quads.
```
banesullivan, line 817: I'm wondering if this should be an enum instead of a string

pyvista/core/filters/image_data.py
```diff
@@ -772,3 +773,116 @@ def _flip_uniform(self, axis) -> 'pyvista.ImageData':
...
+
+        Examples
+        --------
+        See :ref:`contouring_example` for a full example using this filter.
+
+        """
+        if not hasattr(_vtk, 'vtkSurfaceNets3D'):  # pragma: no cover
+            from pyvista.core.errors import VTKVersionError
+
+            raise VTKVersionError('Surface nets 3D require VTK 9.3.0 or newer.')
```
banesullivan, line 849: Can you add an `.. note::` in the docstring that this is only available for VTK 9.3+?

banesullivan (commented):
Really awesome work! Happy to approve this, but I'm wondering if you can bump up the code coverage in the tests since some of the conditional options are not currently covered.

Otherwise small comments

pyvista/core/filters/image_data.py
jmargeta: Done.

jmargeta:
How about the name of the function itself? `labels.contour_labeled()` works well.
However, after using this internally for a while, it feels that `labels.surfaces_labeled()` or just `labels.surfaces()` would be slightly more ergonomic alternatives.

'contour' suggests computation of an iso-contour. Here, 'labels'  are a label map as an output from an image segmentation algorithm (3D image with discrete labels per voxel) we want to extract a surface "per label" rather than an iso-contour.
What do you think? Or is the original name fine?

pyvista/core/filters/image_data.py
banesullivan: Sounds good to me! Nice suggestion!

banesullivan:
> How about the name of the function itself? labels.contour_labeled() works well.

+1 for `contour_labeled` as that is consistent with the few other contouring methods and makes it a bit more discovered with tab-complete. see also: `contour_banded()`

In my mind "contouring" is simply a geometry reducing operation: volume -> surface, surface -> line, line -> point. Since this is taking a volume (ImageData) and producing surfaces (PolyData), this naming feels appropriate to me.

Otherwise, since this is a direct wrapping of `vtkSurfaceNets3D` I don't think anyone could argue against `surface_nets()`

jmargeta:
@banesullivan, thank you very much for your thoughts. `contour_labeled` it is, and I am fairly happy with the PR now.
Please, do not hesitate to ping me if you see anything else to add/change.

banesullivan (approved):
LGTM! Thanks for your contributions, @jmargeta!

eigenvivek:
Thank you for this PR, it is very very helpful!

When I use `contour_labeled`, it prints some internal messages from VTK. Is there a way to suppress this by default? I don't want this showing up on a docs page I'm building. Also, messages like this don't show up when you use marching cubes, for example.

Thank you very much for your work, have been looking forward to the possibility of using SurfaceNets in PyVista for a few months now!

Screen Shot 2023-11-25 at 11 33 28

banesullivan:
@eigenvivek, can you share a `pv.Report()`? I mainly want to know what your VTK version is

jmargeta:
Hi @eigenvivek,
Great to hear you are using this functionality!

You can try to change the vtk logging level before the contour_labeled execution. For example, this mutes the messages for me:
```python
import vtk
vtk.vtkLogger.SetStderrVerbosity(vtk.vtkLogger.ConvertToVerbosity(-1))
```
Disclaimer, I am not sure, this is the official way to do it :)

jmargeta:
All this is happening within the compiled part of VTK.
[  vtkLog(INFO, "Executing Surface Nets 3D");](https://github.com/Kitware/VTK/blob/2638973711fc9284ef71cadfb8468cb78ca427a6/Filters/Core/vtkSurfaceNets3D.cxx#L2322C1-L2322C1)

eigenvivek:
@jmargeta thank you for the suggestion, that worked perfectly :)

@banesullivan using `vtk==9.3.0`

Screen Shot 2023-11-25 at 14 10 40

banesullivan:
Glad you got it fixed! FYI, this discussion also has some tips for suppressing these logs: https://github.com/pyvista/pyvista/discussions/3495

@jmargeta and @eigenvivek, if either of you think we should perform this suppression in this filter directly, don't hesitate to add that

jmargeta:
Hi @banesullivan, thanks for the pointer, `vtk.vtkLogger.SetStderrVerbosity(vtk.vtkLogger.VERBOSITY_OFF)` from the discussion you mentioned looks good. PR incoming.

PR #5188, opened 2023-11-12
Deprecate direction in `translate` function

tkoyama010:
### Overview

Resolves https://github.com/bjlittle/geovista/issues/447

The `transform` function determines the direction of an object by specifying its x-direction vector. However, the y-direction is determined by the function and set to a value. However, this is an implicit decision that can cause unintended results for the user(like https://github.com/bjlittle/geovista/issues/447). Therefore, I moved the direction parameter to deprecated, and normx and normy are specified directly.

### Details

- None

### DIFF
--- a/pyvista/core/utilities/geometric_sources.py
+++ b/pyvista/core/utilities/geometric_sources.py
@@ -15,7 +15,7 @@ from .arrays import _coerce_pointslike_arg
 from .helpers import wrap
 
 
-def translate(surf, center=(0.0, 0.0, 0.0), direction=(1.0, 0.0, 0.0)):
+def translate(surf, center=(0.0, 0.0, 0.0), direction=None, normx=(1.0, 0.0, 0.0), normy=(0.0, 1.0, 0.0)):
     """Translate and orient a mesh to a new center and direction.
 
     By default, the input mesh is considered centered at the origin
@@ -27,23 +27,36 @@ def translate(surf, center=(0.0, 0.0, 0.0), direction=(1.0, 0.0, 0.0)):
         Mesh to be translated and oriented.
     center : tuple, optional, default: (0.0, 0.0, 0.0)
         Center point to which the mesh should be translated.
-    direction : tuple, optional, default: (1.0, 0.0, 0.0)
+    direction : tuple, optional, default: None
         Direction vector along which the mesh should be oriented.
+    normx : tuple, optional, default: (1.0, 0.0, 0.0)
+        Norm x vector along which the mesh should be oriented.
+    normy : tuple, optional, default: (0.0, 1.0, 0.0)
+        Norm y vector along which the mesh should be oriented.
 
     """
-    normx = np.array(direction) / np.linalg.norm(direction)
-    normy_temp = [0.0, 1.0, 0.0]
+    if direction is not None:
+        normx = np.array(direction) / np.linalg.norm(direction)
+        normy_temp = [0.0, 1.0, 0.0]
 
-    # Adjust normy if collinear with normx since cross-product will
-    # be zero otherwise
-    if np.allclose(normx, [0, 1, 0]):
-        normy_temp = [-1.0, 0.0, 0.0]
-    elif np.allclose(normx, [0, -1, 0]):
-        normy_temp = [1.0, 0.0, 0.0]
+        # Adjust normy if collinear with normx since cross-product will
+        # be zero otherwise
+        if np.allclose(normx, [0, 1, 0]):
+            normy_temp = [-1.0, 0.0, 0.0]
+        elif np.allclose(normx, [0, -1, 0]):
+            normy_temp = [1.0, 0.0, 0.0]
 
-    normz = np.cross(normx, normy_temp)
-    normz /= np.linalg.norm(normz)
-    normy = np.cross(normz, normx)
+        normz = np.cross(normx, normy_temp)
+        normz /= np.linalg.norm(normz)
+        normy = np.cross(normz, normx)
+
+        # Deprecated on v0.43.0, estimated removal on v0.46.0
+        warnings.warn(
+            '`direction` argument is deprecated. Please use `normx` and `normy`.',
+            PyVistaDeprecationWarning,
+        )
+    else:
+        normz = np.cross(normx, normy)
 
     trans = np.zeros((4, 4))
     trans[:3, 0] = normx


tkoyama010:
pre-commit.ci autofix

tkoyama010:
> There’s always going to be a tradeoff between convenience and explicitness so it’s worth pointing out an alternative solution is for the user to simply specify `direction=None` and perform the desired rotation explicitly.

This is an excellent idea! I would like to withdraw the deprecation and adopt this proposal. In this case, we only want to set the direction to explicit in the Plane. Therefore, I would like to make a small change to accommodate this.

I am also considering how to address the other suggestions.

tkoyama010:
pre-commit.ci autofix

tkoyama010:
> So `translate` doesn’t just translate, it also rotates. (This was surprising to me but it’s used throughout geometric object construction so whatever.) It does this by accepting a `direction` vector as the image of the x-axis under rotation. Because this underspecifies a 3D rotation, the y-image is chosen as an arbitrary vector orthogonal to the x-image. The z-image is then fully specified by assuming the rotation to be proper so z=x×y. This arbitrariness of the y-image can lead to bugs or confusing behavior.
> 
> There’s always going to be a tradeoff between convenience and explicitness so it’s worth pointing out an alternative solution is for the user to simply specify `direction=None` and perform the desired rotation explicitly.
> 
> A couple issues with this PR:
> 
> * Non-orthogonality / collinearity
>   
>   * The current implementation doesn’t check for non-orthogonality or co-linearity of the two direction vectors. This is justifiable but also could lead to confusing bugs.
> * Naming
>   
>   * The argument names `normx` and `normy` were taken from the `translate` function where they represent normalized vectors.     Their usage in constructors implies (to me at least) that they’re normal vectors, which is incorrect. `translate` should probably take `x_dir` / `y_dir` or `x_image` / `y_image`, and I think it’d be far clearer to customize the names to each object, so `Cone` would have `axial_dir` and `radial_dir` and `Plane` would have `i_dir` and `j_dir`, etc.
> * Default behavior
>   
>   * If we have defaults `Cone(axial_dir=(1, 0, 0), radial_dir=(0, 1, 0))` a user might think it reasonable to call `Cone(axial_dir=(0, 1, 0))` which would lead to a degenerate surface. For a radially symmetric object it’s often convenient to only specify one direction, so `Cone(axial_dir=(1, 0, 0), radial_dir=None)` might be a better default, then choosing a `radial_dir` arbitrarily if it’s not specified, which presumably `translate` could handle by allowing `y_dir=None`.
> * Repetition
>   
>   * There’s a ton of copy-pasted code in `parametric_objects.py`, that could all be refactored out into an internal helper method
> * Improper rotations
>   
>   * I don't feel very strongly about this but for completeness' sake: is it ever possible the user wants  z=−x×y in the rotation?

Thanks for the great discussion!
- Non-orthogonality / collinearity
    I added error checking.
- Naming
    I named the `i_direction` and `j_direction` parameters in the `Plane` function.
- Default behavior
    Since I have decided to support only the `Plane` function in this PR, this issue will be considered in another PR.
- Repetition
    Same as "Default behavior".
- Improper rotations
    At this time, it is unlikely that it will be requested and it will not be considered.

pyvista/core/utilities/geometric_objects.py
```diff
@@ -942,6 +944,18 @@ def Plane(
     j_resolution : int, default: 10
         Number of points on the plane in the j direction.
 
+    i_direction : sequence[float], default: (0.0, 0.0, 1.0)
```
tkoyama010, line 947:
```suggestion
    i_direction : sequence[float], default: (1.0, 0.0, 0.0)
```

user27182:
I like the idea of supporting the use of explicit direction vectors. I'm linking this PR to #4947 since `fit_plane_to_points` specifically avoids using the plane's current `direction` parameter precisely because it's under-defined, so making use of two explicit direction vectors is useful in that case.

I think support for "improper rotations" (which I believe is also referring to left-handed coordinate system) is a good consideration in general, especially for medical applications, but I also think it's fine to generally adopt a right-handed convention in PyVista. In cases where a right-handed convention is assumed, simply adding an explicit `Note` stating the assumption is probably good enough.

user27182:
Some more general comments (not intended to block merging of this PR):

- I'll add that #5100 has a `validate_axes` method which could be used to compute a 3x3 axes matrix when the first two axes are given. The method also supports left-handed axes, which could be used to support left-hand convention rotations for `translate` in a future PR.

- The function `axes_rotation` is also added in #4947 , which could be leveraged by `translate` to actually perform the transformations.

darikg:
@user27182 
> I think support for "improper rotations" (which I believe is also referring to left-handed coordinate system) is a good consideration in general, especially for medical applications, but I also think it's fine to generally adopt a right-handed convention in PyVista. In cases where a right-handed convention is assumed, simply adding an explicit Note stating the assumption is probably good enough.

Here's a case where I think it might be important. If you have `Plane(i_dir=(1, 0, 0), j_dir=(0, 1, 0))` and `Plane(i_dir=(-1, 0, 0), j_dir=(0, 1, 0))` should the two planes have the same or opposite normal?

user27182:
> should the two planes have the same or opposite normal?

I think the normal direction should be calculated as `k = i x j`, so the normals should have opposite sign. I think if it's made clear that the plane is initially defined in the xy plane with initial i direction along +x axis, initial j direction along +y axis, and normals pointing along +z axis, and then rotated by the matrix defined as `[[i_dir, j_dir,  np.cross(i_dir, j_dir)]]`, then the expected result should be unambiguous I think.

tkoyama010:
> > should the two planes have the same or opposite normal?
> 
> I think the normal direction should be calculated as `k = i x j`, so the normals should have opposite sign. I think if it's made clear that the plane is initially defined in the xy plane with initial i direction along +x axis, initial j direction along +y axis, and normals pointing along +z axis, and then rotated by the matrix defined as `[[i_dir, j_dir, np.cross(i_dir, j_dir)]]`, then the expected result should be unambiguous I think.

@user27182
Thank you for your explanation.
@darikg
PR is implemented with this specification. Is OK  to merge this PR?

MatthewFlamm:
This function keeps getting poked at recently since it is embedded for use inside the geometric object generation, and it isn't so well defined. 

That is to say, I agree with the approach to give this time to be considered unless it fixes a critical problem for some other PR.

banesullivan:
As far as I can tell this doesn't fix https://github.com/pyvista/pyvista/issues/5405 or https://github.com/bjlittle/geovista/issues/447 ?

tkoyama010:
> As far as I can tell this doesn't fix #5405 or [bjlittle/geovista#447](https://github.com/bjlittle/geovista/issues/447) ?

Yes. This is different from backward compatibility.
This PR is proposed to define explicit specification of direction.

user27182:
Some more comments:

- I generally support this PR as-is.
- More broadly, from the discussion it seems like it makes sense to add additional direction parameters, but only where necessary (e.g. it's ok for some objects to use a single `direction` vector, and for others to use multiple `direction` vectors).
- In #6283 a new geometric source is added. It uses a `direction_vectors` parameter. But maybe to be more consistent with this PR we should use separate `x_direction`, `y_direction`, z_direction` parameters instead. Or maybe have both options?

PR #5194, opened 2023-11-14
Add option to get `figure_path` from environment variables

AlejandroFernandezLuces:
### Overview

This PR adds the option to get `FIGURE_PATH` global `pyvista` variable from the environment variables of the OS.

PR #5197, opened 2023-11-15
Add option to always get screenshots from plotter shows

AlejandroFernandezLuces:
### Overview

This PR adds an option to always produce screenshots instead of showing the plotter when performing a `pl.show()`. We need this feature in some PyAnsys projects, where we need to have the screenshots of several plotters without modifying the underlaying code. This is related to #5194 , which we also needed for the project.

Let me know your thoughts about this feature 🙂 

### Details

I'm using UUID library to generate unique identifiers for the screenshot filenames. This may be substituted by some attribute from PyVista, but not sure if any would ensure that the name is unique.

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -6761,7 +6761,7 @@ class Plotter(BasePlotter):
                     "A screenshot is unable to be taken as the render window is not current or rendering is suppressed."
                 )
         if _is_current:
-            if pyvista.SCREENSHOTS_ONLY:
+            if pyvista.ON_SCREENSHOT:
                 filename = uuid.uuid4().hex
                 self.last_image = self.screenshot(filename, return_img=True)
             else:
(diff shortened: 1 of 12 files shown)


pyvista/plotting/plotter.py
```diff
@@ -6404,7 +6405,7 @@ def __init__(
                 notebook = scooby.in_ipykernel()
 
         self.notebook = notebook
-        if self.notebook:
+        if self.notebook or pyvista.SCREENSHOTS_ONLY:
```
tkoyama010, line 6408: This is a great feature. Thank you. I will follow the `OFF_SCREEN` naming convention and suggest the variable name `ON_SCREENSHOT`.
```suggestion
        if self.notebook or pyvista.ON_SCREENSHOT:
```

tkoyama010 (changes requested):

tkoyama010 (dismissed):
LGTM

AlejandroFernandezLuces:
Changed to pytest `tmpfolder` for image cleanup. Also, added a fixture that resets the global variables to the values it had before running the tests, I think it's cleaner and can be reused in other tests.

MatthewFlamm (approved):
I can foresee users also wanting sequentially numbered filenames to keep track of which image is which, but a future PR to add in a configurable mode.

tkoyama010 (approved):

Issue #5216, opened 2023-11-23
Use `@no_new_attr` decorator for data model classes


banesullivan:
I'm wondering if we should use the `@no_new_attr` decorator for data model classes to prevent users from continuing to misuse deprecated/removed properties:

https://github.com/pyvista/pyvista/blob/09b9d512d2245901fa719da7c5aae446e7b3b6d9/pyvista/core/utilities/misc.py#L267-L283

Take a look at the scenario in https://github.com/pyvista/pyvista/issues/5209#issuecomment-1823857944 for some context. The gist is that a user lagged behind on PyVista versions for some time and missed the deprecation cycle for `t_coords` -> `texture_coordinates` later finding that after upgrading (jumping many PyVista versions), their use of `t_coords` silently failed.

Leveraging `@no_new_attr` may have unintended consequences for any downstream libraries that further subclass our classes, though I'm not sure. Either way, I think this is significant enough that I'd like to solicit a few folk's opinions from @pyvista/developers

Utilizing `@no_new_attr` would prevent silent failures for things like typos and continued use for deprecated properties:

```py
>>> mesh.foo = 6
>>> mesh.t_coords = array
```

adeak:
> Leveraging `@no_new_attr` may have unintended consequences for any downstream libraries that further subclass our classes, though I'm not sure.

It should be possible to lift the constraint from subclasses. Not sure how, and we might have to refactor a bit, but we could probably do that. And then users who hack our own types directly can use a subclass (as they should).

MatthewFlamm:
Another possibility with less conflict with downstream packages is to maintain a list of forbidden attributes.

user27182:
I'm currently working on some features where I encountered two use cases where the current implementation of @no_new_attr was somewhat incompatible:

- case where a class wants to override __setattr__ for other purposes and still keep the override functionality of __setattr__ provided by @no_new_attr. This is not currently possibly (or at least requires a tricky workaround of some sort) since the decorator will define __setattr__ and not allow the class to define its own override for it
- I've also had problems with subclasses, where I've found @no_new_attr complains about adding new class properties where the property variable is stored locally/internally (which will create two attributes, not just one). I think something doesn't work quite right with how the class property _new_attr_exceptions is defined or used by subclasses. 

I think maybe refactoring @no_new_attr as a mixin parent class instead of as a decorator could address these cases.

adeak:
[...]

PR #5218, opened 2023-11-23
Add Text3DSource class

user27182:
Add `Text3DSource` class.


user27182:
Note: I tried using the `@no_new_attr` decorator with this class but was unable to find a way to make it work since `Text3DSource` overrides `__setattr__`. See [comment](https://github.com/pyvista/pyvista/issues/5216#issuecomment-1824558766) in #5216

pyvista/core/utilities/geometric_objects.py
```diff
@@ -1389,16 +1396,52 @@ def Disc(center=(0.0, 0.0, 0.0), inner=0.25, outer=0.5, normal=(0.0, 0.0, 1.0),
...
+    depth : float, optional
+        Depth of the text. If ``None``, the depth is set to half
+        the :attr:`height` by default. Set to ``0.0`` for planar
+        text.
+
+        .. versionchanged:: 0.43
+
+            The default depth is now calculated dynamcically as
+            half the height. Previously, the default depth had
+            a fixed value of ``0.5``.
```
tkoyama010, line 1430: To ensure backward compatibility, could you add the new argument after the old argument `depth`? Specifically, we need to prevent users from assigning the value of `depth` to `width`.

tkoyama010 (changes requested):

pyvista/core/utilities/geometric_objects.py
user27182: Good point. This is addressed in [200235d](https://github.com/pyvista/pyvista/pull/5218/commits/200235dd0fa4d21578b2acf45fe71e4d97bbefdc)

tests/core/test_geometric_sources.py
```diff
@@ -74,3 +74,158 @@ def test_translate_direction_collinear(is_negative, delta, bunny):
...
+    assert src._output is out1
+    assert src._output.GetPoints() is out2_points
+    assert len(src._output.split_bodies()) == len(string_val)
+
+    # Test calling update modifies the output
+    src.update()
+    assert not src._modified
+    assert src._output is out1
+    assert src._output.GetPoints() is not out2_points
+    assert len(src._output.split_bodies()) == len(src.string)
```
tkoyama010, line 231: In my mind this test is perfect. However, the long function lines make it difficult for the reviewer to understand the scope of the variables. I am also concerned that the previous test might affect the test behind it. Could you please make the Text3DSource definition a fixture and split the test function into separate sections?

tkoyama010 (changes requested):

tests/core/test_geometric_sources.py
```diff
@@ -74,3 +75,199 @@ def test_translate_direction_collinear(is_negative, delta, bunny):
...
+
+def test_text3d_source_output(text3d_source_with_text):
+    # Store initial object references
+    out1 = text3d_source_with_text._output
+    out1_points = out1.GetPoints()
+    assert out1.n_points == 0
+
+    # Test getting output triggers an update
+    out2 = text3d_source_with_text.output
+    assert not text3d_source_with_text._modified
```
tkoyama010, line 215:
```suggestion
    assert text3d_source_with_text._modified
    out2 = text3d_source_with_text.output
    assert not text3d_source_with_text._modified
```

tkoyama010 (approved):
LGTM. Thanks!

Issue #5201, opened 2023-11-17
`n_faces` does not update after `.clean()`


Louis-Pujol:
### Describe the bug, what's wrong, and what you expected.

Hello,

Just found something that looks like a bug in the example provided in [the `clean()` documentation](https://docs.pyvista.org/version/stable/api/core/_autosummary/pyvista.PolyDataFilters.clean.html). Actually the code :

```python
import pyvista as pv
import numpy as np

points = np.array(
    [[0, 0, 0], [0, 1, 0], [1, 0, 0]], dtype=np.float32
)
faces = np.array([3, 0, 1, 2, 3, 0, 2, 2])
mesh = pv.PolyData(points, faces)
mout = mesh.clean()
mout.faces

print(f"faces: {mout.faces}")
print(f"n_faces: {mout.n_faces}")
```

returns

```
faces: [3 0 1 2]
n_faces: 2
```
I was expected `n_faces: 1` instead. Inspecting the `_get_output` function, it seems that VTK does not update the number of cells when cleaning a PolyData (but it updates n points as I expteced). It looks weird to me, but perhaps there is a reason for this behavior. Do you think it should be kept as it is in PyVista or the `n cells` attribute must be updated in this situation ?

Thanks,

### Steps to reproduce the bug.

```python
import pyvista as pv
import numpy as np

points = np.array(
    [[0, 0, 0], [0, 1, 0], [1, 0, 0]], dtype=np.float32
)
faces = np.array([3, 0, 1, 2, 3, 0, 2, 2])
mesh = pv.PolyData(points, faces)
mout = mesh.clean()
mout.faces

print(f"faces: {mout.faces}")
print(f"n_faces: {mout.n_faces}")
```

### System Information

```shell
Date: Fri Nov 17 14:20:47 2023 CET

                OS : Linux
            CPU(s) : 20
           Machine : x86_64
      Architecture : 64bit
               RAM : 31.0 GiB
       Environment : Jupyter
       File system : ext4
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : NVIDIA RTX A2000 8GB Laptop GPU/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 545.23.06
  MathText Support : False

  Python 3.9.18 (main, Aug 25 2023, 13:20:04)  [GCC 9.4.0]

           pyvista : 0.42.3
               vtk : 9.3.0
             numpy : 1.26.2
        matplotlib : 3.8.1
[...]

PR #5229, opened 2023-11-26
deprecate/fix PolyData.n_faces

darikg:
Addresses #5201.

Currently, `PolyData.n_faces` is an alias for `DataSet.n_cells`, i.e. it returns the total number of cells as in the sum of the number of vertices, lines, triangle strips, and polygonal faces. This is surprising because `PolyData.n_lines` returns the number of lines in `PolyData.lines`, `PolyData.n_strips` returns the number of strips in `PolyData.strips`, etc., but `PolyData.n_faces` does not return the number of faces in `PolyData.faces`. Personally, because I almost exclusively work with meshes whose cells are only polygonal faces, I never noticed the distinction.  

This PR introduces `PolyData.n_faces_strict` that returns the number of faces in `PolyData.faces`, and deprecates the current behavior of `PolyData.n_faces`, advising users to use `PolyData.n_cells` or `PolyData.n_faces_strict` as appropriate. 

Because the ultimate goal is to eventually reintroduce `PolyData.n_faces` to return only the number of polygonal faces, and this two-step procedure to switch to `PolyData.n_faces_strict` only to later switch back to `PolyData.n_faces` seems a little tedious for what I suspect is the most common use-case, I also introduce `PolyData.use_strict_n_faces(True)` which would allow downstream projects to globally opt-in to the future behavior early, limiting the necessary changes to a single line at import time rather than every single usage of `PolyData.n_faces`.


pyvista/core/pointset.py
```diff
@@ -1062,11 +1068,36 @@ def n_strips(self) -> int:  # numpydoc ignore=RT01
...
+        Parameters
+        ----------
+        mode : bool
+            If true, all future calls to :attr:`n_faces <pyvista.PolyData.n_faces>`
+            will return the same thing as :attr:`n_faces_strict <pyvista.PolyData.n_faces_strict>`.
+
+        """
+        global _USE_STRICT_N_FACES
+        _USE_STRICT_N_FACES = mode
+
```
tkoyama010, line 1084: In my mind `use_strict_n_faces` should be a function. This is because this operation affects all objects in the library.

pyvista/core/pointset.py
```diff
@@ -1062,11 +1068,36 @@ def n_strips(self) -> int:  # numpydoc ignore=RT01
...
+        return the total number of cells, i.e. the sum of the number of
+        vertices, lines, triangle strips, and polygonal faces.
+        In the future, this will change to return only the number of
+        polygonal faces, i.e. those cells represented in the
+        `pv.PolyData.faces` array. If you want the total number of cells,
+        use `pv.PolyData.n_cells`. If you want only the number of polygonal faces,
+        use `pv.PolyData.n_faces_strict`. Alternatively, you can opt into the
+        future behavior globally by calling `pv.PolyData.use_strict_n_faces(True)`,
+        in which case `pv.PolyData.n_faces` will return the same thing as
+        `pv.PolyData.n_faces_strict`.
```
tkoyama010, line 1100:
```suggestion
        .. deprecated:: 0.43.0
            The current (deprecated) behavior of this property is to
            return the total number of cells, i.e. the sum of the number of
            vertices, lines, triangle strips, and polygonal faces.
            In the future, this will change to return only the number of
            polygonal faces, i.e. those cells represented in the
            `pv.PolyData.faces` array. If you want the total number of cells,
            use `pv.PolyData.n_cells`. If you want only the number of polygonal faces,
            use `pv.PolyData.n_faces_strict`. Alternatively, you can opt into the
            future behavior globally by calling `pv.PolyData.use_strict_n_faces(True)`,
            in which case `pv.PolyData.n_faces` will return the same thing as
            `pv.PolyData.n_faces_strict`.
```

pyvista/core/pointset.py
```diff
@@ -1076,8 +1107,47 @@ def n_faces(self) -> int:  # numpydoc ignore=RT01
...
+        if not _WARNED_DEPRECATED_NONSTRICT_N_FACES:
+            _WARNED_DEPRECATED_NONSTRICT_N_FACES = True
+
+            # deprecated 0.43.0, convert to error in 0.46.0, remove 0.49.0
+            warnings.warn(
+                """The current behavior of `pv.PolyData.n_faces` has been deprecated.
+                Use `pv.PolyData.n_cells` or `pv.PolyData.n_faces_strict` instead.
+                See the documentation in '`pv.PolyData.n_faces` for more information.""",
+                PyVistaDeprecationWarning,
+            )
```
tkoyama010, line 1124: is recommended to make it a function as well as `use_strict_n_faces` in my mind. It is prefer to avoid using global within a class as much as possible.

tkoyama010 (changes requested):
Thank you! This is a nice PR.

darikg:
@tkoyama010 I moved the module-level globals to class variables, is that what you meant? Or did you mean use_strict_n_faces should operate distinctly on each individual `PolyData` instance?

tkoyama010 (approved):
> @tkoyama010 I moved the module-level globals to class variables, is that what you meant? Or did you mean use_strict_n_faces should operate distinctly on each individual `PolyData` instance?

Sorry for the poor explanation. The suggestion was to define `use_strict_n_faces` as an independent function rather than a PolyData method. However, your suggestion changed my mind since defining it as a method makes it clear that it is an operation on PolyData (I am also pleased that `global` has been removed.). I am fine with it as it is.

darikg:
> > @tkoyama010 I moved the module-level globals to class variables, is that what you meant? Or did you mean use_strict_n_faces should operate distinctly on each individual `PolyData` instance?
> 
> Sorry for the poor explanation. The suggestion was to define `use_strict_n_faces` as an independent function rather than a PolyData method. However, your suggestion changed my mind since defining it as a method makes it clear that it is an operation on PolyData (I am also pleased that `global` has been removed.). I am fine with it as it is.

Ah I understand now, thanks for clarifying. Either way I agree it's better without `global`.

PR #4811, opened 2023-08-28
Refactor Trame widget - Vue 2 and Vue 3 versions

annehaley:
### Overview

This PR is intended to make the Trame widget compatible with both Vue 2 and Vue 3 client types within a Trame application. Resolves #4776. 
This PR refactors `pyvista/trame/ui.py` into several new files: 

- `trame/pyvista/ui/__init__.py`
- `trame/pyvista/ui/base_viewer.py`
- `trame/pyvista/ui/vuetify2.py`
- `trame/pyvista/ui/vuetify3.py`

### Details
Previously, `ui.py` defined a class `Viewer` and several top-level functions and variables. The following changes were made:

- Most top-level functions and variables were moved to `ui/__init__.py`. 
- Two top-level functions, `button` and `checkbox` required different implementations for the Vuetify 2 and Vuetify 3 API, so these were moved to `ui/vuetify2.py` and `ui/vuetify3.py`, respective to the implementations.
- The `Viewer` class was renamed to `BaseViewer` and moved to `ui/base_viewer.py`. Two class methods, `ui_controls` and `ui`, required different implementations for Vuetify 2 and Vuetify 3. The `ui` method for `BaseViewer` now raises a `NotImplementedError` and the `ui_controls` method was only ever referenced within the `ui` method, so it was removed entirely from `BaseViewer`
- Two new `Viewer` classes were written in `ui/vuetify2.py` and `ui/vuetify3.py`, both derived from the `BaseViewer` class. Both derived classes implement the `ui` and `ui_controls` methods, using the Vuetify 2 and Vuetify 3 APIs respectively.
- The top-level function `get_or_create_viewer` (used within the top-level function `plotter_ui`) now has an extra bit of logic to determine which implementation of `Viewer` to use in the case of new viewer creation. See below code. 

```
CLIENT_TYPE = get_server().client_type
if CLIENT_TYPE == 'vue2':
    return Vue2Viewer(plotter, suppress_rendering=suppress_rendering)
else:
    return Vue3Viewer(plotter, suppress_rendering=suppress_rendering)
```

This solves the problem I mentioned in the linked issue about needing to set the Trame server's client type before the import of `plotter_ui` within the Trame application. Now the client type is not fetched until the `plotter_ui` function calls `get_or_create_viewer`. This PR does not change the API for using the `plotter_ui` function.

Issue #4708, opened 2023-07-26
Trame app crashes with large meshes


leilakanync:
### Describe the bug, what's wrong, and what you expected.

Hi,

I was trying to display a large mesh with the new trame backend (~5M vertices) in a jupyter notebook, but the app crashes with this error.

```
ERROR:root:Exception raised
ERROR:root:ConnectionResetError(104, 'Connection reset by peer')
ERROR:root:Traceback (most recent call last):
  File "/home/container_user/.local/lib/python3.10/site-packages/wslink/protocol.py", line 329, in onMessage
    await self.sendWrappedMessage(
  File "/home/container_user/.local/lib/python3.10/site-packages/wslink/protocol.py", line 466, in sendWrappedMessage
    await ws.send_bytes(attachments[key])
  File "/usr/local/lib/python3.10/dist-packages/aiohttp/web_ws.py", line 315, in send_bytes
    await self._writer.send(data, binary=True, compress=compress)
  File "/usr/local/lib/python3.10/dist-packages/aiohttp/http_websocket.py", line 688, in send
    await self._send_frame(message, WSMsgType.BINARY, compress)
  File "/usr/local/lib/python3.10/dist-packages/aiohttp/http_websocket.py", line 659, in _send_frame
    await self.protocol._drain_helper()
  File "/usr/local/lib/python3.10/dist-packages/aiohttp/base_protocol.py", line 90, in _drain_helper
    await asyncio.shield(waiter)
  File "/usr/lib/python3.10/asyncio/futures.py", line 285, in __await__
    yield self  # This tells Task to wait for completion.
  File "/usr/lib/python3.10/asyncio/tasks.py", line 304, in __wakeup
    future.result()
  File "/usr/lib/python3.10/asyncio/futures.py", line 201, in result
    raise self._exception.with_traceback(self._exception_tb)
  File "/usr/lib/python3.10/asyncio/selector_events.py", line 949, in _write_ready
    n = self._sock.send(self._buffer)
ConnectionResetError: [Errno 104] Connection reset by peer
```

However it works well with pythreejs.
I even have the issue with 1.5M vertices

### Steps to reproduce the bug.

Call the plot function on a pyvista object having around 5 millions vertices.

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Wed Jul 26 15:56:10 2023 Europe

                  OS : Linux
              CPU(s) : 8
             Machine : x86_64
        Architecture : 64bit
                 RAM : 54.9 GiB
[...]

Issue #4652, opened 2023-07-13
Trame client mode JupyterLab: connection closed error


dennissergeev:
### Describe the bug, what's wrong, and what you expected.

I have recently updated my `pyvista` environment, and I can't get the trame widget working in Jupyter Lab, when I use it both on a remote server _and_ on my local PC.

The error I'm getting is "Connection closed", exactly the same as described by @dannymcculloch in #3979. This happens if I plot the global topographical map example. A "simpler" example such as the cone object does get plotted ok (albeit with a warning: `WARNING:aiohttp.websocket:Client protocols [''] don’t overlap server-known ones ()`).

I primarily use `pyvista` on a remote server via an SSH tunnel. However, I discovered that exactly the same behaviour occurs when I try to do it on my local PC too.

I'm using the latest versions of the key packages (see below).

Thanks in advance.

P.S. On a related note, I typically use pyvista in a separate conda environment (and no, it doesn't help if I install it in the `base` environment either). What is the correct way to setup `trame` and `jupyter-server-proxy` if I use multiple environments? I launch Jupyter Lab from the base environment where I have `nb_conda_kernels` installed.

### Steps to reproduce the bug.

```python
import pyvista as pv
import numpy as np
from pyvista import examples

pv.global_theme.trame.server_proxy_enabled = True
pv.set_jupyter_backend("client")

pv.Cone().plot()  # works

examples.download_topo_global().plot()  # fails
```

I've also tried using the code suggested as a workaround in #3979:
<details>
<summary>host_address workaround</summary>

```python
import os
os.environ['TRAME_DEFAULT_HOST'] = host_address

import pyvista as pv
from pyvista.trame.jupyter import elegantly_launch, launch_server

elegantly_launch(host_address)
# or:
# await launch_server(host_address).ready

pv.global_theme.trame.server_proxy_enabled = True
pv.global_theme.trame.server_proxy_prefix = f"{host_address}:"

pv.set_jupyter_backend("client")
```
</details>

but alas it doesn't help.

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Thu Jul 13 11:26:21 2023 BST

                OS : Linux
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 31.2 GiB
       Environment : Jupyter
       File system : ext4
        GPU Vendor : Intel
      GPU Renderer : Mesa Intel(R) UHD Graphics 620 (KBL GT2)
[...]

PR #5240, opened 2023-11-28
Add support for trame_jupyter_extension (trame over jupyter comm)

alesgenova:
### Overview

Add support in pyvista for the new [`trame_jupyter_extension`](https://github.com/Kitware/trame-jupyter-extension)

When this extension is installed (and the flag added in this PR is `True`), the trame server and client will communicate through the existing jupyter comm, instead of creating a new websocket connection.

### Notes
- I have removed the launching of the server from the `set_jupyter_backend()` function, relying on the fact that the server will be started in `show_trame()` if it's not running. The reason why I did it (other than the fact that it seemed redundant) is because if the server is prematurely started with `aiohttp` wslink backend, it will not be possible to make it work with the `jupyter` backend, and vice versa. So it seemed like a good idea to delay the starting of the server as much as possible until we actually need to display something. While this seemed like a reasonable change to me, let me know if there were deeper reasons for it to be there.
- It seemed that both the vue2 and vue3 viewers were being embedded in a vue2 `VAppLayout`. According to @jourdain this shouldn't be happening, so I changed the `BaseViewer` class so that each subclass is in charge of creating the appropriate layout.


pyvista/trame/ui/base_viewer.py
```diff
@@ -258,3 +259,21 @@ def export(self):  # numpydoc ignore=RT01
     def ui(self):
         """Implement in derived classes."""
         raise NotImplementedError()
+
+    def make_layout(self, *args, **kwargs) -> AbstractLayout:
```
banesullivan, line 263:
```suggestion
    def make_layout(self, *args, **kwargs) -> AbstractLayout:  # pragma: no cover
```

banesullivan (changes requested):
Looking great and working excellently! Some minor feedback left inline. Would you also please update `doc/source/user-guide/jupyter/trame.rst` with usage instruction for this and any potential tradeoffs or benefits for using this?

Also, would you please comment on how this extension may or may not help with #4708, and #4652? I think this will resolve those issues.

banesullivan:
Re the server launching changes you mentioned:

> While this seemed like a reasonable change to me, let me know if there were deeper reasons for it to be there.

I think you took the right approach here, thanks for cleaning this up.

> It seemed that both the vue2 and vue3 viewers were being embedded in a vue2 VAppLayout

Thanks for cleaning this up. Seems like it was an artifact of #4811

alesgenova:
> @alesgenova, let me know how I can support you as this is the final PR we'd like to sneak in for the `0.43` release in the coming days (see #5138)

Thanks for the review, it won't be a problem to address the minor issues :)
I'm planning on having it done on Monday, unless it's too late for the new release?

banesullivan:
That should work! Thanks!

pyvista/plotting/themes.py
```diff
@@ -1469,6 +1479,33 @@ def server_proxy_prefix(self):  # numpydoc ignore=RT01
...
+    def jupyter_extension_available(self) -> bool:  # numpydoc ignore=RT01
+        """Return whether the trame_jupyter_extension is detected."""
+        return self._jupyter_extension_available
+
+    @jupyter_extension_available.setter
+    def jupyter_extension_available(self, _available: bool):  # numpydoc ignore=GL08
+        warnings.warn(
+            "The jupyter_extension_available flag is read only and is automatically detected."
+        )
+
```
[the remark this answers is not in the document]
alesgenova: Originally I omitted the setter like you suggest here, however one of the unit tests was failing because the setter was missing, and that's why I ended up adding a noop setter.

Agreed that the theme might not be the best place for available flag. If you have a better place in mind, I'll be happy to move it there

alesgenova:
@banesullivan 
I addressed all the comments from your review (minus the `extension_available` setter, see my comment above), you can find the changes in the last commit added to the PR.

I also tested out the example reported in #4652 , and can confirm that at least locally it works when using the new extension, while it's broken (as expected) when using the server proxy. Not definitive but seems promising.

pyvista/plotting/themes.py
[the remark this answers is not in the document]
banesullivan: Ah, I didn't realize a test was complaining about this. I think let's just leave this as is then. Thanks for looking into this!

banesullivan:
Looks fantastic! Thank you for addressing all of my feedback. One remaining question, though: is there any disadvantage to using this jupyter comms infrastructure here? I believe at some point @jourdain and I thought it may be limited compared to the capabilities of wslink

banesullivan:
Not sure if this needs to be included in the docs, but going to add it as a comment here for posterity and future readers: if running on a _Jupyter**Hub**_, you will need to install either of these (`jupyter_server_proxy` or `trame_jupyter_extension`) in the base Python environment where the JupyterHub server is running.

banesullivan (dismissed):
LGTM! Happy to give the approval and merge regardless of the testing coverage check. IMO, testing is sufficient here.

pyvista/trame/ui/base_viewer.py
```diff
@@ -258,3 +259,21 @@ def export(self):  # numpydoc ignore=RT01
     def ui(self):
         """Implement in derived classes."""
         raise NotImplementedError()
+
+    def make_layout(self, *args, **kwargs) -> AbstractLayout:  # pragma: no cover
+        """Create instance of an AbstractLayout which is appropriate for a concrete viewer.
```
tkoyama010, line 264:
```suggestion
        """Create an instance of an AbstractLayout which is appropriate for a concrete viewer.
```

pyvista/trame/jupyter.py
```diff
@@ -149,6 +168,8 @@ def launch_server(server=None, port=None, host=None, **kwargs):
     if host is None:
         # Default to `127.0.0.1` unless user sets TRAME_DEFAULT_HOST
         host = os.environ.get('TRAME_DEFAULT_HOST', '127.0.0.1')
+    if wslink_backend is None and pyvista.global_theme.trame.jupyter_extension_enabled:
+        wslink_backend = "jupyter"
```
tkoyama010, line 172:
```suggestion
    if wslink_backend is None and pyvista.global_theme.trame.jupyter_extension_enabled:  # pragma: no cover
        wslink_backend = "jupyter"
```

pyvista/trame/jupyter.py
```diff
@@ -334,12 +361,19 @@ def handler(viewer, src, **kwargs):
...
         return EmbeddableWidget(plotter, **kwargs)
 
+    if jupyter_extension_enabled is None:
+        jupyter_extension_enabled = pyvista.global_theme.trame.jupyter_extension_enabled
+
     if name is None:
         server = get_server(name=pyvista.global_theme.trame.jupyter_server_name)
     else:
         server = get_server(name=name)
     if name is None and not server.running:
```
tkoyama010, line 371:
```suggestion
    if name is None and not server.running:  # pragma: no cover
```

pyvista/trame/jupyter.py
```diff
@@ -356,19 +390,27 @@ def handler(viewer, src, **kwargs):
...
-    # Show as cell result
-    src = build_url(
-        server,
-        ui=plotter._id_name,
-        server_proxy_enabled=server_proxy_enabled,
-        server_proxy_prefix=server_proxy_prefix,
-        host=kwargs.get('host', 'localhost'),
-        protocol=kwargs.get('protocol', 'http'),
-    )
+    if jupyter_extension_enabled:
```
tkoyama010, line 393:
```suggestion
    if jupyter_extension_enabled:  # pragma: no cover
```

tkoyama010 (dismissed):
LGTM. Great contribution!

pyvista/trame/jupyter.py
```diff
@@ -334,12 +361,19 @@ def handler(viewer, src, **kwargs):
...
+    if jupyter_extension_enabled is None:
+        jupyter_extension_enabled = pyvista.global_theme.trame.jupyter_extension_enabled
+
     if name is None:
         server = get_server(name=pyvista.global_theme.trame.jupyter_server_name)
     else:
         server = get_server(name=name)
-    if name is None and not server.running:
-        elegantly_launch(server)
+    if name is None and not server.running:  # pragma: no cover
```
tkoyama010, line 371:
```suggestion
    if name is None and not server.running:
```

pyvista/trame/jupyter.py
```diff
@@ -334,12 +361,19 @@ def handler(viewer, src, **kwargs):
...
+        wslink_backend = "aiohttp"
+        if jupyter_extension_enabled:
```
tkoyama010, line 373:
```suggestion
        if jupyter_extension_enabled: # pragma: no cover
```

tkoyama010 (approved):

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

banesullivan:
Sorry I jumped the gun a bit there, @tkoyama010! I think we're ready to start shipping the release so I merged this

tkoyama010:
No problem. We waited enough.

jourdain:
> I believe at some point @jourdain and I thought it may be limited compared to the capabilities of wslink

The binary messages remain binary, so we should be good. The overhead is most likely less than the server-proxy.

jourdain:
yes for jupyter definitely

PR #5242, opened 2023-11-28
Add cast_to_polydata for Cell

### LINKED ISSUE #5238: Add a `cast_to_polydata` method for `Cell`

There is a `cast_to_unstructured_grid` method, but sometimes it is needed to have a `PolyData` object instead.  We can check if the cell is:
- 0D -> create PolyData with vertex 
- 1D -> create PolyData with line 
- 2D -> create PolyData with faces (in pyvista terminology)

For example, in #5227 it was requested to get a single streamline from a `PolyData` object with each cell being a single streamline.  When `mesh.get_cell(i)` is used to get a single streamline, it would be better to have the option to cast to `PolyData` so that it has the same options as the original streamlines object.

_No response_

### Pseudocode or Screenshots

_No response_

MatthewFlamm:
### Overview

Closes #5238.  Adds ability to cast to PolyData for non 3D cells

### Details

Added more cell types to testing in general as those are important to test when casting to PolyData.


MatthewFlamm:
As far as I can tell there is no way to construct a Poly Vertex cell through the `PyVista.PolyData` constructor method.  I think we need to add a `vertices` input like `faces` so that it is possible to do this.

MatthewFlamm:
> As far as I can tell there is no way to construct a Poly Vertex cell through the `PyVista.PolyData` constructor method. I think we need to add a `vertices` input like `faces` so that it is possible to do this.
> 
> This feels like a separate PR.

Started in #5245

pyvista/core/cell.py
```diff
@@ -163,6 +163,45 @@ def plot(self, **kwargs):
...
+        """
+        cells = [len(self.point_ids)] + list(range(len(self.point_ids)))
+        if self.dimension == 0:
+            return pyvista.PolyData(self.points.copy(), verts=cells)
+        if self.dimension == 1:
+            return pyvista.PolyData(self.points.copy(), lines=cells)
+        if self.dimension == 2:
+            if self.type == CellType.TRIANGLE_STRIP:
+                return pyvista.PolyData(self.points.copy(), strips=cells)
+            return pyvista.PolyData(self.points.copy(), faces=cells)
```
tkoyama010, line 201:
```suggestion
            else:
                return pyvista.PolyData(self.points.copy(), faces=cells)
```

tkoyama010 (approved):
LGTM. Thanks!

PR #5245, opened 2023-11-28
Add ability to control vertices for `PolyData` creation; allow all cell types to be used independently

MatthewFlamm:
### Overview

Enable doing `pv.PolyData(points, verts=[1, 2]) # single vertex cell`  or `pv.PolyData(points, verts=[2, 0, 1] # single poly vertex cell`.  This PR also allows fully indepedent definition of all the different cell types.  For example `pv.PolyData(points, verts=[1, 0], lines=[2, 1, 2], faces=[3, 3, 4, 5], strips=[4, 6, 7, 8, 9])` 

### Details

This is needed to enable being able to cast to polydata for cells in #5242


pyvista/core/pointset.py
```diff
@@ -665,19 +685,19 @@ def __init__(
             raise TypeError(dedent(msg.strip('\n')))
 
         # At this point, points have been setup, add faces and/or lines
-        if faces is None and lines is None and strips is None:
+        if faces is None and lines is None and strips is None and verts is None:
```
tkoyama010, line 688: nits
```suggestion
        if faces and lines and strips and verts is None:
```

tkoyama010 (changes requested):

pyvista/core/pointset.py
MatthewFlamm: I've never seen this type of excursion before.  I tested it and it works but it is unexpected to me.  When all are `None`, I would expect `verts is None` to evaluate to `True` and then `strips is verts is None` to evaluate to `False` because it compares `None` to `True`.  What am I missing?

tkoyama010: I knew this for knowledge but had never checked the Python documentation.

In the [Comparisons section of Python document](https://docs.python.org/3/reference/expressions.html#comparisons), we will find the following sentence. We can recognize `aces is None and lines is None and strips is None and verts is None` is equal `faces is lines is strips is verts is None` by this.

> Comparisons can be chained arbitrarily, e.g., x < y <= z is equivalent to x < y and y <= z, except that y is evaluated only once (but in both cases z is not evaluated at all when x < y is found to be false).

MatthewFlamm: This makes a lot more sense to me thanks!  I think the last example is very clear.  I think this is a bit more obscure to think about, whereas `x < y < z` is more natural.  I'm fine with the suggestion though, it is certainly shorter and compact.

tkoyama010 (approved):
LGTM

PR #5246, opened 2023-11-29
Prevent unnecessary copy in DataSet.rename_array

### LINKED ISSUE #5244: Wrong result - Rename the "implicit_distance" after compute_implicit_distance filter

After the _implicit distance_ of a dataset is computed, I tried to rename it using the method `rename_array`.  But I got a different data array result from the original one which means not only the name was changed.
I noticed the `rename_array` uses the `pop` method of` point_data` or `cell_data`. Then I did more research and found it seems there is a bug in the `pop` method.
I did a simple example (check the code in the next block) and got the same bug in my project. However, the strange thing is that in my data array can work well with the method `rename_array`, only the _implicit distance_ data array cannot work.

Here is a simple example
```python
import pyvista as pv

sphere = pv.Sphere(radius=5)
plane = pv.Plane()

new_sphere = sphere.compute_implicit_distance(plane)
new_sphere["pd"] = range(new_sphere.n_points)
new_sphere.plot(scalars="implicit_distance")

new_sphere.rename_array("implicit_distance", "im")
new_sphere.rename_array("pd", "pdc")
new_sphere.plot(scalars="im")
```
The bug is obvious when running the above example.

### System Information

```shell
Date: Tue Nov 28 19:39:51 2023

                OS : Windows
            CPU(s) : 16
           Machine : AMD64
      Architecture : 64bit
               RAM : 15.8 GiB
       Environment : Python
        GPU Vendor : Intel
       GPU Version : 4.5.0 - Build 31.0.101.1999
  MathText Support : False

  Python 3.11.5 | packaged by conda-forge | (main, Aug 27 2023, 03:23:48) [MSC
  v.1936 64 bit (AMD64)]

           pyvista : 0.42.3
               vtk : 9.2.6
             numpy : 1.24.3
        matplotlib : 3.7.1
            scooby : 0.7.2
             pooch : v1.7.0
            pillow : 10.0.0
           imageio : 2.31.2
           IPython : 8.14.0
             scipy : 1.11.1
            meshio : 5.3.4
      nest_asyncio : 1.5.6

  Intel(R) oneAPI Math Kernel Library Version 2023.1-Product Build 20230303
```

### Screenshots

Original Data
original_data
Renamed Data
renamed_data

darikg:
Fixes #5244


tkoyama010 (dismissed):
LGTM

tests/core/test_dataset.py
```diff
@@ -649,31 +649,50 @@ def test_set_active_scalars_name(grid):
...
+    # Regression test for issue #5244
+    def make_mesh():
+        m = pv.Sphere()
+        m.point_data['orig'] = np.ones(m.n_points)
+        return m
+
+    for _ in range(20):  # Depends on garbage collection
+        mesh = make_mesh()
+        mesh.rename_array('orig', 'renamed')
+        assert (mesh['renamed'] == 1).all()
```
dcbr, line 695: Can't we force garbage collection like [here](https://github.com/pyvista/pyvista/blob/19896236df625885e453f95677476ec588434354/tests/plotting/conftest.py#L45)?
E.g. something like
```suggestion
    mesh = make_mesh()
    mesh.rename_array('orig', 'renamed')
    gc.collect()
    assert (mesh['renamed'] == 1).all()
```

darikg: You're right, it's not a great test. I think the issue is not whether gc runs, but whether the freed memory is overwritten so we can detect the change. I just changed the test to test directly that the array isn't deleted with a VTKObject.AddObserver('DeleteEvent') callback.

tests/core/test_dataset.py
```diff
@@ -649,31 +649,59 @@ def test_set_active_scalars_name(grid):
...
+        m = pv.Sphere()
+        m.point_data['orig'] = np.ones(m.n_points)
+        return m
+
+    mesh = make_mesh()
+    was_deleted = [False]
+
+    def on_delete(*_):
+        # Would be easier to throw an exception here but even though the exception gets printed to stderr
+        # pytest reports the test passing -- I guess because of how it's thrown inside the VTK callback?
```
tkoyama010, line 697: Good guess, but I would like to comment only on facts. Let's leave a pull request number so that we know what we thought
```suggestion
        # pytest reports the test passing. See #5246 .
```

dcbr (approved):

PR #5251, opened 2023-11-30
Default to matplotlib-base for conda

bjlittle:
### Overview

Opt to use `matplotlib-base` instead of `matplotlib` in the `environment.yml` for `conda`.

This will potentially resolve with fewer dependencies i.e., `matplotlib` additionally depends on `pyqt` and `tornado`, see [here](https://github.com/conda-forge/matplotlib-feedstock/blob/main/recipe/meta.yaml#L88-L89).

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - vtk
   - scooby>=0.5.1
   - meshio
-  - matplotlib
+  - matplotlib-base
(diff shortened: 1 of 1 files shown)


bjlittle:
Or are those dependencies explicitly required?

bjlittle:
Okay, the CI has answered my question ... worth a punt.

PR #5262, opened 2023-12-02
Fix test code to bump trame-vtk from 2.5.9 to 2.6.2

tkoyama010:
### Overview

I found that `test_trame_views` affect other test function in #5255. So I moved it to the end of the code.

@banesullivan Would you give me your thoughts?

closes #5255

### Details

- None


banesullivan:
Hm this is odd behavior to me. I'm wondering if our usage pattern here is incorrect or if this should be accounted for/handled upstream. @jourdain, would you please take a look at this? It seems that the presence of this test (which is just `get_server('foo')`) causes later usage of the export functionality in trame-vtk on different/the original server to cause the error `TypeError: a bytes-like object is required, not 'NoneType'`. I think this happened with the 2.6 release of trame-vtk

jourdain:
Actually, having it on the HTMLElement instance might actually fix it... But I really need to think deeper on that.

banesullivan:
@tkoyama010, let's keep this test where it is, but mark it as `pytest.mark.skip` (unconditional skip) for now as I this is a very niche use case and link to this pull request

banesullivan (approved):

tkoyama010:
pre-commit.ci autofix

tkoyama010:
LGTM

banesullivan (approved):

Issue #5277, opened 2023-12-04
[BUG] Scraper doesn't let code execution finish before taking screenshot


alexrockhill:
### Describe the bug, what's wrong, and what you expected.

The `mne.viz.Brain` inherits from `PyVistaFigure` and is scraped in `sphinx-example-gallery` examples but `PyVistaFigure.set_camera` calls don't take effect before the image is snapped.

### Steps to reproduce the bug.

Here is an example artifact, the brain plots at the end are way zoomed in and they should have the camera zoomed out after the `show_view` which calls `set_camera` call: https://output.circle-artifacts.com/output/job/7614b75f-9bc0-4d8e-a59b-983dfdbead6d/artifacts/0/dev/auto_examples/ieeg_locate.html

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Mon Dec 04 10:00:48 2023 PST

                OS : Linux
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 31.2 GiB
       Environment : IPython
       File system : ext4
        GPU Vendor : Intel
      GPU Renderer : Mesa Intel(R) Xe Graphics (TGL GT2)
       GPU Version : 4.6 (Core Profile) Mesa 23.2.1-1pop0~1699535617~22.04~6cd9a72~dev
  MathText Support : False

  Python 3.10.10 | packaged by conda-forge | (main, Mar 24 2023, 20:08:06)
  [GCC 11.3.0]

           pyvista : 0.41.1
               vtk : 9.2.6
             numpy : 1.23.5
        matplotlib : 3.7.2
            scooby : 0.7.1
             pooch : v1.7.0
           imageio : 2.27.0
         pyvistaqt : 0.11.0
             PyQt5 : 5.15.7
           IPython : 8.14.0
        ipyvtklink : 0.2.2
        ipywidgets : 8.0.6
             scipy : 1.10.1
              tqdm : 4.65.0
      nest_asyncio : 1.5.6
--------------------------------------------------------------------------------
```

banesullivan:
I see you're using `pyvistaqt` here... perhaps the threading involved with that is the problem? Otherwise this may be tough for us to track down... Would you be able to try to create some sort of minimal, reproducible example? I suspect in that process you may tease out a usage issue with pyvistaqt, but otherwise its going to be tough for us to dive into the MNE code to track this down

alexrockhill:
[...]

PR #5283, opened 2023-12-05
Update plots before scraping

alexrockhill:
Fixes https://github.com/pyvista/pyvista/issues/5277

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -167,8 +167,8 @@ jobs:
             Examples-1-
 
       - name: Setup xvfb dependencies
-        - run: git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
-        - run: ./mne-python/tools/setup_xvfb.sh
+        run: git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
+        run: ./mne-python/tools/setup_xvfb.sh
 
       - name: Build wheel and install pyvista
         run: |
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -31,7 +31,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
@@ -151,7 +151,7 @@ jobs:
           fetch-depth: 2
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
@@ -240,7 +240,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -166,6 +166,10 @@ jobs:
           restore-keys: |
             Examples-1-
 
+      - name: Setup xvfb dependencies
+        - run: git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
+        - run: ./mne-python/tools/setup_xvfb.sh
+
       - name: Build wheel and install pyvista
         run: |
           pip install build
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -167,8 +167,9 @@ jobs:
             Examples-1-
 
       - name: Setup xvfb dependencies
-        run: git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
-        run: ./mne-python/tools/setup_xvfb.sh
+        run: |
+          git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
+          ./mne-python/tools/setup_xvfb.sh
 
       - name: Build wheel and install pyvista
         run: |
--- a/pyvista/plotting/utilities/sphinx_gallery.py
+++ b/pyvista/plotting/utilities/sphinx_gallery.py
@@ -85,7 +85,10 @@ class Scraper:
         image_path_iterator = block_vars["image_path_iterator"]
         figures = pyvista.plotting.plotter._ALL_PLOTTERS
         for _, plotter in figures.items():
-            plotter.update()
+            if plotter.iren.initialized:
+                plotter.update()
+                if hasattr(plotter, "app"):
+                    plotter.app.processEvents()
             fname = next(image_path_iterator)
             if hasattr(plotter, "_gif_filename"):
                 # move gif to fname
--- a/pyvista/plotting/utilities/sphinx_gallery.py
+++ b/pyvista/plotting/utilities/sphinx_gallery.py
@@ -85,6 +85,7 @@ class Scraper:
         image_path_iterator = block_vars["image_path_iterator"]
         figures = pyvista.plotting.plotter._ALL_PLOTTERS
         for _, plotter in figures.items():
+            plotter.update()
             fname = next(image_path_iterator)
             if hasattr(plotter, "_gif_filename"):
                 # move gif to fname
--- a/pyvista/plotting/utilities/sphinx_gallery.py
+++ b/pyvista/plotting/utilities/sphinx_gallery.py
@@ -85,7 +85,7 @@ class Scraper:
         image_path_iterator = block_vars["image_path_iterator"]
         figures = pyvista.plotting.plotter._ALL_PLOTTERS
         for _, plotter in figures.items():
-            if plotter.iren.initialized:
+            if plotter.iren is not None and plotter.iren.initialized:
                 plotter.update()
                 if hasattr(plotter, "app"):
                     plotter.app.processEvents()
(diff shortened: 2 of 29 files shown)


alexrockhill:
Can you check if the `plotter` is initialized before calling update? That would fix the errors:

```
FAILED tests/plotting/test_scraper.py::test_scraper[1] - RuntimeError: Render window interactor must be initialized before processing events.
```

Not sure how to do that if anyone has a second.

alexrockhill:
I ran this locally and it fixes my issue.

tkoyama010:
To pass test coverage, we need a sample for adding a test. Please provide us with a sample that you have run locally.

tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

requirements_test.txt
```diff
@@ -12,11 +12,14 @@ meshio<5.4.0
 nest_asyncio<1.5.9
 numpydoc==1.6.0
 param<2.1.0
+PyQt6!=6.6.1
+PyQt6-Qt6!=6.6.1
```
tkoyama010, line 16:
```suggestion
PyQt6!=6.6.1,<6.7.0
PyQt6-Qt6!=6.6.1,<6.7.0
```

alexrockhill:
Sorry about not figuring out the pre-commit protocol, got it working now and it's very nice.

I'm not sure if it's easiest to mark this to test it only on the CIs with the MNE dependencies. I'm not sure of a way to do that in this case though... `pytest.mark` I don't think would work. Maybe just some logic about trying and failing dependencies. I don't want it to be overly complicated and not look good though.

requirements_test.txt
```diff
@@ -12,11 +12,14 @@ meshio<5.4.0
...
 pytest<7.5.0
 pytest-cov<4.2.0
 pytest-memprof<0.3.0
 pytest-xdist<3.4.0
 pytest_pyvista==0.1.8
+qtpy
```
tkoyama010, line 22:
```suggestion
qtpy<2.5.0
```

tkoyama010:
> Sorry about not figuring out the pre-commit protocol, got it working now and it's very nice.

Never mind. We are happy to show that our CI is excellent.

> I'm not sure if it's easiest to mark this to test it only on the CIs with the MNE dependencies. I'm not sure of a way to do that in this case though... `pytest.mark` I don't think would work. Maybe just some logic about trying and failing dependencies. I don't want it to be overly complicated and not look good though.

I am not as familiar with this bug as you are, but this is a common problem in all environments in my mind. So I'm fine as is.

tests/test_tinypages.py
```diff
@@ -83,3 +84,9 @@ def plot_file(plt, num, subnum):
 
     assert b'you should not be reading this right now' not in html_contents
     assert b'should be printed: include-source with no args' in html_contents
+
+    # test that the plot has the camera position updated with a checksum
+    assert (
+        imread(html_dir / 'plot_cone_00_00.png').sum()
+        != imread(html_dir / 'plot_cone_00_01.png').sum()
```
tkoyama010, line 91:
```suggestion
        imread(html_dir / 'plot_cone_00_01.png').sum()
        != imread(html_dir / 'plot_cone_00_02.png').sum()
```

tests/tinypages/plot_cone.py
```diff
@@ -1,3 +1,12 @@
...
+import pyvista as pv
+
+pl = pv.Plotter()
+pl.iren.initialize()
+pl.app = QApplication.instance()
+pl.add_mesh(pv.Cone())
+pl.camera_position = 'xy'
+
+pl2 = pv.Plotter()
+pl2.add_mesh(pv.Cone())
```
tkoyama010, line 12:
```suggestion
pv.Cone().plot()

pl = pv.Plotter()
pl.iren.initialize()
pl.app = QApplication.instance()
pl.add_mesh(pv.Cone())
pl.camera_position = 'xy'

pl2 = pv.Plotter()
pl2.add_mesh(pv.Cone())
```

tests/test_tinypages.py
```diff
@@ -83,3 +84,9 @@ def plot_file(plt, num, subnum):
 
     assert b'you should not be reading this right now' not in html_contents
     assert b'should be printed: include-source with no args' in html_contents
+
+    # test that the plot has the camera position updated with a checksum
```
tkoyama010, line 88:
```suggestion
    # test that the plot has the camera position updated with a checksum when the Plotter has an app instance
```

tkoyama010 (changes requested):
Would you mind the output of the image to three pieces in order to keep the existing test intact?

.github/workflows/testing-and-deployment.yml
```diff
@@ -166,6 +166,10 @@ jobs:
           restore-keys: |
             Examples-1-
 
+      - name: Setup xvfb dependencies
+        run: git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
+        run: ./mne-python/tools/setup_xvfb.sh
```
tkoyama010, line 171:
```suggestion
        run: |
          git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
          ./mne-python/tools/setup_xvfb.sh
```

alexrockhill:
Thanks for the review, yes the changes look good

tkoyama010 (approved):
LGTM. Thanks!

tests/plotting/test_scraper.py
```diff
@@ -16,6 +18,15 @@ def test_scraper(tmpdir, monkeypatch, n_win):
...
     plotters = [pv.Plotter(off_screen=True) for _ in range(n_win)]
+    if n_win == 2:
+        # add cone, change view to test that it takes effect
+        plotters[0].iren.initialize()
+        plotters[0].app = QApplication.instance()
+        plotters[0].add_mesh(pv.Cone())
+        plotters[0].camera_position = 'xy'
+
+        plotters[1].add_mesh(pv.Cone())
+
```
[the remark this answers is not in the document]
alexrockhill: Sure

alexrockhill:
I got down a bit of a rabbit hole trying to add coverage for the dynamic scraper but it looks like it doesn't have any coverage in general so maybe not the PR to do it in. I was close to adding it, just with `plotters[n].show(interactive=False)` there were 7 vtk objects that we not garbage collected. Might be something to look into as I guess that may cause memory leaks if you use the dynamic scraper?

Should be good to go after implementing the change you asked for with a new test function without branching logic.

tests/plotting/test_scraper.py
```diff
@@ -10,12 +12,61 @@
...
+        plotter.close()
+
+
 @pytest.mark.parametrize('n_win', [1, 2])
 def test_scraper(tmpdir, monkeypatch, n_win):
     pytest.importorskip('sphinx_gallery')
     monkeypatch.setattr(pv, 'BUILDING_GALLERY', True)
     pv.close_all()
     plotters = [pv.Plotter(off_screen=True) for _ in range(n_win)]
+
```
tkoyama010, line 69:
```suggestion
```

tests/plotting/test_scraper.py
```diff
@@ -54,6 +105,7 @@ def test_scraper(tmpdir, monkeypatch, n_win):
     scraper(block, block_vars, gallery_conf)
     for img_fname in img_fnames:
         assert os.path.isfile(img_fname)
+
```
tkoyama010, line 108:
```suggestion
```

tkoyama010 (approved):
LGTM. Let's make a test of `DynamicScraper `in the future.

.github/workflows/testing-and-deployment.yml
```diff
@@ -166,6 +166,11 @@ jobs:
           restore-keys: |
             Examples-1-
 
+      - name: Setup xvfb dependencies
+        run: |
+          git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
+          ./mne-python/tools/setup_xvfb.sh
+
```
banesullivan, line 173: Woah... why was this added here? This needs to be removed ASAP as this section of the tests is strictly non-graphics.

requirements_test.txt
```diff
@@ -12,11 +12,14 @@ meshio<5.4.0
...
 numpydoc==1.6.0
 param<2.1.0
+PyQt6!=6.6.1,<6.7.0
+PyQt6-Qt6!=6.6.1,<6.7.0
 pytest<7.5.0
 pytest-cov<4.2.0
 pytest-memprof<0.3.0
 pytest-xdist<3.4.0
 pytest_pyvista==0.1.8
+qtpy<2.5.0
```
banesullivan, line 22: We do not test Qt dependencies in PyVista. This was a part of the entire motivation for making PyVistaQt a separate repository

banesullivan:
Apologies @alexrockhill and @tkoyama010  but these changes need to be reverted as they are too impactful and need to undergo a more thorough review

pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -110,6 +119,10 @@ class DynamicScraper:
 
     """
 
+    def __repr__(self):
+        """Return a stable representation of the class instance."""
+        return f"<{type(self).__name__} object>"
+
```
banesullivan, line 125: @alexrockhill, why was this needed? We do not intend for this class to be used or displayed to users as far as I understand? 

Asking because I'm going to see if I can re-land this PR with less impact

.github/workflows/testing-and-deployment.yml
banesullivan: Further I want to discourage the use of custom scripts from external projects in our core CI routines. We should keep any configuration and set up limited to well-established GitHub Actions or scripts managed by PyVista.

Adding and executing arbitrary shell scripts is a major red flag and security risk albeit this is from a well-trusted downstream project.

pyvista/plotting/utilities/sphinx_gallery.py
alexrockhill: It was added when I was trying to add coverage of the dynamic scraper but then not removed when adding that coverage exposed a bug where the dynamic scraper failed to close plots and so that was punted to a future PR.

.github/workflows/testing-and-deployment.yml
alexrockhill: That sounds reasonable. They were added for PyQtx.QWidgets.QApplication.process_events testing. Not sure there is a way to fake that for non-graphics tests, I just don't know there might be though. Otherwise since it was tested it could be no covered.

PR #5285, opened 2023-12-05
Add custom `mypy` plugin to support generic array-like types

### LINKED ISSUE #5250: PyVista types are not compatible with `ArrayLike`

Some of PyVista's new type definitions from #5171  are not compatible with numpy's `ArrayLike` type. This is surprising because `ArrayLike` handles nested sequences and should be an all-encompasing type. That is, I expected the pyvista type `FloatMatrix` to be compatible with `ArrayLike`.

Instead, `mypy` complains when doing the following:

``` python
import numpy.typing as npt
from pyvista.core._typing_core import FloatMatrix

def foo(obj: npt.ArrayLike): 
    return

x: FloatMatrix = np.eye(3)
foo(x)  # mypy error here
```

The (formatted) mypy error that is generated is:
``` python
error: Argument 1 to "foo" has incompatible type
  "
      ndarray[Any, dtype[floating[Any]]] | 
      Sequence[ndarray[Any, dtype[floating[Any]]] | 
      Sequence[float]]
  "
expected 
  "
      _SupportsArray[dtype[Any]] | 
      _NestedSequence[_SupportsArray[dtype[Any]]] |
       bool | int | float | complex | str | bytes | 
      _NestedSequence[bool | int | float | complex | str | bytes]
  "
[arg-type]
```

Not sure if types are incompatible due to PyVista's types supporting finite sequences, whereas numpy's nested sequences are generic.  This may be a `mypy` issue, since in [`numpy._typing._array_like.py`](https://github.com/numpy/numpy/blob/8216139d9200aea5f93e5d775a011526387ffb47/numpy/_typing/_array_like.py#L56C6-L56C6) they define finite sequences but not use them, citing ```# TODO: Wait until mypy supports recursive objects in combination with typevars```

This issue was encountered while trying to incorporate the new types as part of #5226. I intended to use `npt.ArrayLike` as the input array type for the `validate_array` function, but I think I'll set the type to `Any` for now to avoid these unexpected type errors.

### Links to source code.

_No response_

### Pseudocode or Screenshots

_No response_

user27182:
### Overview

Currently, non-generic types are specified, e.g. `IntVector`, `FloatVector`, in `_typing_core`. Ideally, these would be generic, and would instead be expressed as `Vector[int]`, `Vector[float]`, etc. in the same way  `List[float]` and `Tuple[int]` may be used.

This PR replaces the non-generic type aliases with generic ones.

To support these generic types, this PR:
- Adds `pyvista.core._typing._array_like` type aliases (similar to `numpy.typing._array_like`)
- Replaces existing type aliases `Vector`, `Matrix`, `Array`, etc. with generic versions
- Adds a custom `mypy` plugin ([https://github.com/pyvista/npt-promote](https://github.com/pyvista/npt-promote)) to enable two-way type promotion between:
    -  `float` and `numpy.floating`
    - `int` and `numpy.signedinteger`
  __Note__: the type promotions fix/silence `mypy` errors that would otherwise be generated when using `npt.NDArray[float]` or `npt.NDArray[int]`

Since changes are made to the file structure, a new `pyvista.core._typing` package is added first (this PR) and the `pyvista.core._typing_core.py` module will be removed in a separate PR.

This PR also resolves #5250.

### Details

The current pyvista types, e.g. `Vector`, are array-like in the sense that they allow either `Sequence`s or `numpy.ndarray` types. Ideally, numpy's `numpy.typing.ArrayLike` would be generic and support doing something like `ArrayLike[Shape, DType]` such that `Vector` could be defined as an alias like `Vector = ArrayLike[(1,), DType]`. Unfortunately,  `ArrayLike` is not made to be generic like this and static type hints are not yet supported by numpy/mypy for hinting the array shape.  

To support generic type hints, this PR defines new `_array_like` definitions. These are similar to NumPy's `_array_like` definitions, except they
-  are generic with one typevar
-  use finite sequences (NumPy's definitions use recursive nested sequences which are not yet compatible with mypy, see #5250)
-  exclude scalars (NumPy's definitions include scalar types by default, though an actual array is often expected instead)

Furthermore, type hints for the array's  `DType` only support native numpy types, e.g. `np.int_`,  and do not support hints for builtin types, e.g. `int`. This means that mypy will complain if you try to annotate a type as `NDArray[int]`, for example:
```
error: Type argument "int" of "dtype" must be a subtype of "generic"  [type-var]
```

This PR overcomes this limitation by implementing a custom mypy plugin which applies the following two-way type promotions (for static type checking only):
-`np.floating` is promoted as a subtype of `float` and vice-versa.
-`np.signedinteger` is promoted as a subtype of `int` and vice-versa
The mypy plugin was made to be a separate repository since pre-commit does not allow installing from a module under test.


.pre-commit-config.yaml
```diff
@@ -74,9 +74,10 @@ repos:
   - id: mypy
     exclude: ^(doc/|tests/|examples/|pyvista/ext/|examples_trame/)
     additional_dependencies: [
-    "mypy-extensions==1.0.0",
-    "toml==0.10.2",
-    "numpy",
+      "mypy-extensions==1.0.0",
+      "toml==0.10.2",
+      "numpy>=1.21",
+      "https://github.com/pyvista/npt-promote/tarball/main"
```
user27182, line 80: `npt-promote` is a new pyvista repo. It has not yet been added to PyPI. If added, this line could be changed to "npt-promote".

pyvista/core/_typing_core.py
```diff
@@ -1,26 +1,10 @@
 """Type aliases for type hints."""
+# flake8: noqa: F401
```
tkoyama010, line 2:
```suggestion
# noqa: F401
```

pyvista/core/cell.py
```diff
@@ -614,7 +614,7 @@ class CellArray(_vtk.vtkCellArray):
 
     def __init__(
         self,
-        cells: Optional[Union[IntMatrix, IntVector]] = None,
+        cells: Optional[Union[Matrix[int], Vector[int]]] = None,
```
tkoyama010, line 617: Great! This expression is elegant.

tkoyama010 (changes requested):

tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

tkoyama010:
@darikg @MatthewFlamm To me this PR looks very attractive. However, I am not familiar with the NumPy-type check situation, so I would like to hear what you two have to say.

.pre-commit-config.yaml
tkoyama010: If possible, I would like to merge this after releasing version 0.1. Is it OK to register it with PyPI?

user27182: Yes please register it if you can. I wasn't able to a few weeks ago since there was a block on making new PyPI accounts.

user27182: Looks like registration for PyPI is open again, I'll make some minor changes to `pyvista/npt-promote` register it, and push a new commit here with a fixed version of the plugin.

MatthewFlamm:
I must confess that these niche typing issues are not natural to me either.  I'm surprised that numpy doesn't allow this already, or is this a version thing?  But on the face the changes here look fine to me.

.pre-commit-config.yaml
user27182: I have updated `npt-promote` with CI testing. I think's ready for a first publishing (e.g. as version 0.1). I went to publish on `PyPI` but it seems fairly involved, requiring several builds to be submitted for each platform and python version  (I have no prior experience with publishing to PyPI). How is this process typically done within `pyvista` ?

Also, I'm not sure if the publishing account matters? Is it better to publish it from the same account from which `pyvista` is also published?

tkoyama010: Below is where we upload to PyPI. It is triggered when released with the tag. Let's start by releasing using Python 3.9 version.

https://github.com/pyvista/pyvista/blob/f37b5da5c9de3bff2e69abebc147155eab8e25db/.github/workflows/testing-and-deployment.yml#L221-L228

Since `npt-promote` is your project, you can publish it under your account :)

user27182: Thanks, that was helpful. `npt-promote==0.1` has been published.

tkoyama010 (approved):

PR #5288, opened 2023-12-06
Use trame-jupyter-extension by default

banesullivan:
Follow up to #5240 to set the default comms backend to leverage `trame-jupyter-extension`. This should mean that `pip install "pyvista[jupyter]"` will _just work_ for most JupyterLab environments and deployments.

@alesgenova and @jourdain, would you please double check what I'm doing here?

### DIFF
--- a/doc/source/user-guide/jupyter/trame.rst
+++ b/doc/source/user-guide/jupyter/trame.rst
@@ -148,7 +148,7 @@ On MyBinder, the ``JUPYTERHUB_SERVICE_PREFIX`` string often needs to prefix
 In PyVista, we automatically check for the presence of this variable and
 prepend it to the ``server_proxy_prefix``.
 
-You may also need to install ``nest_asyncio`` to seemlessly use launch the
+You may also need to install ``nest_asyncio`` to seamlessly use launch the
 Trame server in the background.
 
 
--- a/doc/source/user-guide/jupyter/trame.rst
+++ b/doc/source/user-guide/jupyter/trame.rst
@@ -83,7 +83,36 @@ Remote Jupyter Host
 
 When using PyVista in Jupyter that is hosted remotely (docker, cloud JupyterHub,
 binder, or otherwise), you will need to pair the Trame backend with either
-``jupyter_server_proxy`` or ``trame_jupyter_extension``.
+``trame_jupyter_extension`` or ``jupyter_server_proxy``.
+
+Trame Jupyter Extension
+#######################
+
+`Trame Jupyter Extension <https://github.com/Kitware/trame-jupyter-extension/>`_
+enables the trame server and client to communicate over the existing
+`Jupyter Comms <https://jupyter-notebook.readthedocs.io/en/stable/comms.html>`_
+infrastructure, instead of creating a separate WebSocket connection.
+
+Using this extension removes the need for a secondary web server and thus
+``jupyter_server_proxy``.
+
+Using pip, you can install the extension:
+
+.. code::
+
+    pip install trame_jupyter_extension
+
+If using Jupyter Lab 3.x, make sure to install the version 1.x of the extension:
+
+.. code::
+
+    pip install "trame_jupyter_extension<2"
+
+Once the extension is installed, you can select whether PyVista will use it by
+setting the following flag to ``True`` or ``False``:
+
+* :py:attr:`pyvista.global_theme.trame.jupyter_extension_enabled
+  <pyvista.plotting.themes._TrameConfig.jupyter_extension_enabled>`
 
 
 Jupyter Server Proxy
@@ -94,7 +123,7 @@ lets you access the Trame server hosting the views of the PyVista plotters
 alongside your notebook, and provide authenticated web access to them directly
 through Jupyter.
 
-To configure PyVista and Trame to work with ``jupyter-server-proxy`` in a remote
+To configure PyVista and Trame to work with ``jupyter_server_proxy`` in a remote
 environment, you will need to set some options on the global PyVista theme:
 
 * :py:attr:`pyvista.global_theme.trame.server_proxy_enabled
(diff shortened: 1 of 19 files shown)


banesullivan:
@tkoyama010, FYI we may want to sneak this in for the release pending feedback from @alesgenova. I'm working to confirm this all works

docker/requirements.txt
```diff
@@ -3,10 +3,13 @@ colorcet
 imageio
 imageio-ffmpeg
 ipywidgets<9.0.0
-jupyter-server-proxy
 jupyterlab<5.0.0
 meshio
+nest_asyncio
```
jourdain, line 8: do we still need it?

pyproject.toml
```diff
@@ -48,10 +48,10 @@ io = [
 ]
 jupyter = [
     'ipywidgets',
-    'jupyter-server-proxy',
     'nest_asyncio',
     'trame>=2.5.2',
     'trame-client>=2.12.7',
+    'trame-jupyter-extension>=2.0.1',
```
jourdain, line 54: I think the name is `trame_jupyter_extension`

docker/requirements.txt
banesullivan: Good catch! Thanks!

jourdain:
> Hm. I'm seeing some console errors now when trying to use this in our docker image:
> 
> ```
> trame::jupyter-comm::close FIXME
> chunk-vendors.74a8bb6a.js:1 Missing header for received binary message
> ```

Can you elaborate on what you are doing for running into that issue. That error comes from wslink when a binary message is received but without a matching text message for its registration.

jourdain:
Everything seems reasonable otherwise.

docker/requirements.txt
banesullivan: Ah turns out we do still need this due to the following:

https://github.com/pyvista/pyvista/blob/9273ed67f6fa721cfb75492f3f0ef541009227c6/pyvista/trame/jupyter.py#L373C5-L382

jourdain: I thought @alesgenova tried and thought it was not needed anymore.

banesullivan:
> Can you elaborate on what you are doing for running into that issue. 

I'm able to reproduce this minimally. I'll open a new issue under https://github.com/Kitware/trame-jupyter-extension

docker/requirements.txt
```diff
@@ -5,11 +5,10 @@ imageio-ffmpeg
...
 meshio
-nest_asyncio
 trame>=2.5.2
-trame-client>=2.12.7
-trame-jupyter-extension>=2.0.1
-trame-server>=2.11.7
-trame-vtk>=2.5.8
-trame-vuetify>=2.3.1
+trame_client>=2.12.7
+trame_jupyter_extension>=2.0.1
```
jourdain, line 10: `trame_jupyter_extension>=2.0.1` is the only one with `_`

docker/requirements.txt
```diff
@@ -5,11 +5,10 @@ imageio-ffmpeg
...
+trame_server>=2.11.7
```
jourdain, line 11: revert

docker/requirements.txt
```diff
@@ -5,11 +5,10 @@ imageio-ffmpeg
...
 jupyterlab<5.0.0
 meshio
-nest_asyncio
 trame>=2.5.2
-trame-client>=2.12.7
-trame-jupyter-extension>=2.0.1
-trame-server>=2.11.7
-trame-vtk>=2.5.8
-trame-vuetify>=2.3.1
+trame_client>=2.12.7
```
jourdain, line 9: revert

docker/requirements.txt
```diff
@@ -5,11 +5,10 @@ imageio-ffmpeg
 ipywidgets<9.0.0
 jupyterlab<5.0.0
 meshio
-nest_asyncio
 trame>=2.5.2
-trame-client>=2.12.7
```
jourdain: revert

docker/requirements.txt
banesullivan: I think pip resolves these either way?

docker/requirements.txt
banesullivan: @alesgenova's work from #5240 is currently still dependent on nest_asyncio. @alesgenova, any ideas on how to work around this?

doc/source/user-guide/jupyter/trame.rst
```diff
@@ -119,35 +148,8 @@ On MyBinder, the ``JUPYTERHUB_SERVICE_PREFIX`` string often needs to prefix
...
-.. code::
-
-    pip install "trame_jupyter_extension<2"
-
-Once the extension is installed, you can select whether PyVista will use it by
-setting the following flag to ``True`` or ``False``:
-
-* :py:attr:`pyvista.global_theme.trame.jupyter_extension_enabled
-  <pyvista.plotting.themes._TrameConfig.jupyter_extension_enabled>`
+You may also need to install ``nest_asyncio`` to seemlessly use launch the
```
tkoyama010, line 151:
```suggestion
You may also need to install ``nest_asyncio`` to seamlessly use launch the
```

docker/requirements.txt
jourdain: I guess I got it from the generated readme... https://pypi.org/project/trame-jupyter-extension/

I should update the readme to match the actual package name from pypi.

docker/requirements.txt
jourdain: I see. Thanks for the clarification.

alesgenova:
> Hm. I'm seeing some console errors now when trying to use this in our docker image:
> 
> ```
> trame::jupyter-comm::close FIXME
> chunk-vendors.74a8bb6a.js:1 Missing header for received binary message
> ```

I need to look into that more, but this warning could be due to the fact that we are displaying the output of a cell from a previous kernel. As in, the iframe is displayed but tries to connect to a kernel that doesn't exist anymore (a connection closed message should appear in the UI). Re-executing the cell would refresh it correctly.

To verify this hypothesis you can clear all the cells outputs, and restart jupyter lab. The warning then shouldn't appear once you re-run the pyvista cell.

jourdain:
That error is on the JS side, were a binary message is received but no matching ID for that attachment is found.

alesgenova:
@jourdain 
No, that warning comes from the FakeWebsocked `close()` method (https://github.com/Kitware/trame-jupyter-extension/blob/d4a2fa676e8e9379584f3b215091077c572c966c/src/websocket.ts#L107)

This method gets invoked wither when the (fake) websocket is closed, or when the underlying jupyter comm is closed

jourdain:
I was talking about `Missing header for received binary message`

alesgenova:
> Considering that the following issues do not look to be easily fixable, I'm going to push that we move forward with our release and leave jupyter-server-proxy as the preferred default comms infrastructure for this backend:
> 
>     * [break ipywidgets Kitware/trame-jupyter-extension#1](https://github.com/Kitware/trame-jupyter-extension/issues/1)
> 
>     * [Broken with the jupyter base-botebook Docker image Kitware/trame-jupyter-extension#6](https://github.com/Kitware/trame-jupyter-extension/issues/6)
> 
> 
> The ipywidgets issue concerns me greatly but I did not experience any issues myself.
> 
> Marking as draft and considering this PR blocked

The ipywidget issue should have been marked as resolved, I fixed it with this PR: https://github.com/Kitware/trame-jupyter-extension/pull/3

banesullivan:
Ah glad to hear it is resolved! I think this PR is going to still have to wait on https://github.com/Kitware/trame-jupyter-extension/issues/6 though

No rush, but I think it'd be awesome to make this the default so that all of this will _just work_ in nearly all Jupyter environments

PR #5290, opened 2023-12-06
Update data_set.py to skip doctest of  `get_data_range`

tkoyama010:
### Overview

The following error occurs about once every two times in CI's doctest.
This is a new feature added in vtk 9.3 and is most likely a bug.
Please let us skip this test temporarily.

```bash
=================================== FAILURES ===================================
______ [doctest] pyvista.core.filters.data_set.DataSetFilters.pack_labels ______
6410 
6411         Load non-contiguous image labels
6412 
6413         >>> from pyvista import examples
6414         >>> import numpy as np
6415         >>> image_labels = examples.download_frog_tissue()
6416 
6417         Show range of labels
6418 
6419         >>> image_labels.get_data_range()
Expected:
    (0, 29)
Got:
    (0, 255)
```

### Details

- None

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -6416,7 +6416,7 @@ class DataSetFilters:
 
         Show range of labels
 
-        >>> image_labels.get_data_range()
+        >>> image_labels.get_data_range()  # doctest:+SKIP
         (0, 29)
 
         Find 'gaps' in the labels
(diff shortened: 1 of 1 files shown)


tkoyama010:
It may be better to investigate this as a bug.

MatthewFlamm:
Since this is also showing up in the document example, this feels like it is better to fix the bug then continue on. However if this is holding up other unrelated PRs then it might be worth silencing it for now.

user27182:
The flaky tests and other issues with frog tissue were resolved by #6110, this pr can be closed.

PR #5293, opened 2023-12-06
Fix docstring of `DataSet.cell_neighbors_levels`

dionhaefner:
`cell_neighbors_levels` refers to `point_neighbors` when it should be `cell_neighbors`. Probably a copy-paste artifact.


tkoyama010:
Good catch thanks!

tkoyama010:
github-actions preview

PR #5299, opened 2023-12-06
Integrate `flake8` settings into `pyproject.toml`

tkoyama010:
### Overview

Integrate `flake8` settings into `pyproject.toml`.

### Details

- None


MatthewFlamm:
I think it would be great if `flake8` would support this directly, but the devs seem quite strict about not supporting it.  Unfortunately, the package that is added here is noted to be unstable and of Alpha quality in the sense that the methodology is liable to break often.  I think we should avoid it.

https://github.com/csachs/pyproject-flake8#:~:text=Caveat,stability%20between%20versions.

tkoyama010:
> I think it would be great if `flake8` would support this directly, but the devs seem quite strict about not supporting it. Unfortunately, the package that is added here is noted to be unstable and of Alpha quality in the sense that the methodology is liable to break often. I think we should avoid it.
> 
> Better link: https://github.com/csachs/pyproject-flake8#caveat ~https://github.com/csachs/pyproject-flake8#:~:text=Caveat,stability%20between%20versions~.

Thank you for pointing this out. Your point has changed my mind that we should choose the stability of the tool over the comfort of the tool's configuration. I will withdraw this proposal.

PR #5304, opened 2023-12-07
Add `ruff` linter config to `pyproject.toml`

tkoyama010:
### Overview

We previously considered introducing ruff (https://github.com/astral-sh/ruff) and dismissed it (https://github.com/pyvista/pyvista/pull/4115). On the other hand, the `matplotlib` project has introduced ruff (https://github.com/matplotlib/matplotlib/pull/25147) and `numpy` project is also considering using it. ruff has a linter and a formatter, Last time I tried to introduce both of those, I failed. So this time I would like to propose only the linter which has less impact.

### Details

- None

### DIFF
--- a/pyvista/trame/ui/base_viewer.py
+++ b/pyvista/trame/ui/base_viewer.py
@@ -8,7 +8,6 @@ See `pyvista.trame.ui.vuetify2` and ``pyvista.trame.ui.vuetify3` for its derived
 import io
 
 from trame.app import get_server
-from trame.widgets import html
 from trame_client.ui.core import AbstractLayout
 
 import pyvista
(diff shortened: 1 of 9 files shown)


pyvista/trame/ui/base_viewer.py
```diff
@@ -8,7 +8,6 @@
 import io
 
 from trame.app import get_server
-from trame.widgets import html
```
tkoyama010, line 11: I am not familiar with trame viewer so I don't know if this import makes sense.

banesullivan: cc @annehaley or @alesgenova

annehaley: This is fine to remove, this import is unused.

banesullivan (approved):
Lgtm, thanks for making this happen! Excited to be using ruff 🐶

PR #5314, opened 2023-12-09
Update auto-approve.yml to fix syntax error

tkoyama010:
### Overview

Update auto-approve.yml to fix syntax error. Sorry I didn't notice before merging.

### Details

- None


tkoyama010:
I close this, as there is no assurance that this fix is correct. See https://github.com/ansys/pymapdl/pull/2229#discussion_r1422197662 .

PR #5310, opened 2023-12-09
Revert "Update plots before scraping"

banesullivan:
Reverts pyvista/pyvista#5283

cc @alexrockhill and @tkoyama010

PR #5321, opened 2023-12-11
Update plots before scraping

alexrockhill:
Reimplements #5310.

@banesullivan I think these changes are much more modest and within the scope of the `pyvista` as opposed to the `pyvistaqt` library. I'm glad we tested it with `qtpy` compatibility so that we know this actually works in the original PR but we don't need to keep that all in the PR just to test one line so I think this should be okay now.

### DIFF
--- a/pyvista/plotting/utilities/sphinx_gallery.py
+++ b/pyvista/plotting/utilities/sphinx_gallery.py
@@ -85,7 +85,10 @@ class Scraper:
         image_path_iterator = block_vars["image_path_iterator"]
         figures = pyvista.plotting.plotter._ALL_PLOTTERS
         for _, plotter in figures.items():
-            plotter.update()
+            if plotter.iren.initialized:
+                plotter.update()
+                if hasattr(plotter, "app"):
+                    plotter.app.processEvents()
             fname = next(image_path_iterator)
             if hasattr(plotter, "_gif_filename"):
                 # move gif to fname
--- a/pyvista/plotting/utilities/sphinx_gallery.py
+++ b/pyvista/plotting/utilities/sphinx_gallery.py
@@ -85,6 +85,7 @@ class Scraper:
         image_path_iterator = block_vars["image_path_iterator"]
         figures = pyvista.plotting.plotter._ALL_PLOTTERS
         for _, plotter in figures.items():
+            plotter.update()
             fname = next(image_path_iterator)
             if hasattr(plotter, "_gif_filename"):
                 # move gif to fname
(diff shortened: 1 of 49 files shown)


banesullivan:
github-actions preview

banesullivan:
@ChristosT, since you're the most versed on the scraper at this time (being one of the only other people to modify it), would you please take a quick look at these changes?

pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -53,6 +53,14 @@ def html_rst(
     return images_rst
 
 
+def _process_events_before_scraping(plotter):
+    """Process events such as changing the camera or object before scraping."""
+    if plotter.iren is not None and plotter.iren.initialized:
+        if hasattr(plotter, "app") and plotter.app is not None:
```
banesullivan, line 59: Nitpicky, but would you please add a detailed inline comment here about how `app` is specific to pyvistaqt and this is specific to handling that type of plotter (this will help us in the future)

pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -53,6 +53,14 @@ def html_rst(
...
+            plotter.app.processEvents()
+        plotter.update()
```
banesullivan, line 61: I want to double check that this `.update()` call is necessary or if it should just be in the nest if-block

banesullivan (changes requested):
Just some nitpicky comments, otherwise this looks good! I'd like @ChristosT's feedback before merging since they are also modifying the scraper significantly in #4938

pyvista/plotting/utilities/sphinx_gallery.py
alexrockhill: If it were nested in the if block then it would still solve my issue but if there were an `iren` that isn't associated with `pyvistaqt` then the `update` call would still be helpful. Not sure if that's a common use case but it seems to be logically consistent with the structure of `pyvista` as I understand it.

ChristosT:
It looks good from my point of view. 
It actually makes sense to want to drain the interaction queues before taking a screenshot!

I merged this branch with my changes and built the documentation locally, as far as I checked everything looks good. We just need to be careful when merging because there is a (trivial) merge conflict in `sphinx._gallery.py` but since I already handled it, we can go ahead and merge this first if every other comment is resolved.

ChristosT (approved):

tkoyama010:
github-actions preview

banesullivan (approved):

tkoyama010:
@alexrockhill Thank you for your flexibility in responding to our comments. It is a great contribution.

alexrockhill:
Thanks for the reviews everyone

PR #5329, opened 2023-12-12
Deploy preview document only from command

tkoyama010:
### Overview

Now we have `github-actions preview` command to deploy preview. So we don't need the deployment of `docs.yml` anymore.

### Details

- None


banesullivan (approved):
Thanks for cleaning this up, @tkoyama010!

PR #5332, opened 2023-12-13
Inherit `XdmfReader` class from `TimeReader` to improve time properties

### LINKED ISSUE #5331: Timestamp extraction from XdmfReader

Hey, guys!

I needed to extract the timestamps from a .xdmf file, but I couldn't find a clear and easy way to do that using pyvista's XdmfReader. I noticed that the XdmfReader class does not inherit from the TimeReader class, so that is probably the reason. 

In the following code snippet, I found a workaround to this issue, which is to call the methods from the vtkXdmfReader original class inside the pyvista reader object. However, this is not an ideal solution. Is there anyway this could be integrated into pyvista in future updates?

```python
import pyvista as pv
import numpy as np
import vtk

file_path = "file.xdmf"

# Create an Xdmf reader
pv_reader = pv.XdmfReader(path=file_path)

# Get the field names
field_names = pv_reader.point_array_names

# Get the output information
info = pv_reader.reader.GetOutputInformation(0)

# Get the time steps
timestamps = info.Get(vtk.vtkCompositeDataPipeline.TIME_STEPS())
```

An example can be found in the code snippet I provided.

### Pseudocode or Screenshots

```python
import pyvista as pv
import numpy as np
import vtk

file_path = "file.xdmf"

# Create an Xdmf reader
pv_reader = pv.XdmfReader(path=file_path)

# Get the field names
field_names = pv_reader.point_array_names

# Get the output information
info = pv_reader.reader.GetOutputInformation(0)

# Get the time steps
timestamps = info.Get(vtk.vtkCompositeDataPipeline.TIME_STEPS())
```

tkoyama010:
### Overview

resolves #5331 .

### Details

- None


tkoyama010:
github-actions preview

MatthewFlamm:
If you can get the time steps in a list you can probably get this reader to inherit from `TimeReader`.

MatthewFlamm:
We should discourage using the pipeline until we figure out whether we will use it for something in Pyvista, but I think this just gets the needed key to get the info.  So it seems okay.

tkoyama010:
github-actions preview

pyvista/core/utilities/reader.py
```diff
@@ -2434,6 +2435,11 @@ class XdmfReader(BaseReader, PointCellDataSelection):
     _vtk_module_name = "vtkIOXdmf2"
     _vtk_class_name = "vtkXdmfReader"
 
+    def __init__(self, path):
+        super().__init__(path)
+        self._active_time_value = 0.0
```
MatthewFlamm, line 2440: There is no guarantee in most file formats that there will be a time point at 0.0.  I assume it is the same here.  We might have to set it after the information is read using `_set_defaults_post`.  To be backwards compatible we will need to know what the current behavior is, usually the first time point.

tkoyama010: Thanks. You are right that the first step is not zero. And I also understood that _set_defaults_post is the way to solve this problem. This abstract class is very well done!
The first step is active by default, so the latest changes should be fine.

MatthewFlamm (approved):
The only very small question is whether the `_active_time_value` should be initialized as `None` in an `__init__`.  I don't think we need to since it is set in `_set_defaults_post`, which is already called from the inherited `__init__`.  But a consideration if you prefer explicitly initializing it in this classes `__init__`.  Some linters would require you to.

PR #872, opened 2020-08-19
Add RubberBand2D interactor style

GuillaumeFavelier:
This PR adds the 'RubberBand2D' interactor style.

- [x] Update testing

Closes #871

PR #5336, opened 2023-12-13
Add Paraview-like 2D interactor style

### LINKED ISSUE #5335: Add Paraview-like 2D interactor
### Discussed in https://github.com/pyvista/pyvista/discussions/5322

<div type='discussions-op-text'>

<sup>Originally posted by **FrankFrank9** December 11, 2023</sup>
Hello,

what is the pyvista equivalent of a traced state command in paraview `renderView1.InteractionMode = '2D'`?

I have not been able to find anything suitable

Best regards</div>

MatthewFlamm:
### Overview

This PR adds in a 2D interactor style that is similar to the Paraview interactor.  Closes #5335 and was discussed in #5322.

### Details

When selecting 2D interaction in Paraview, it also sets parallel projection in addition to changing the mouse interaction.  #872 started a precedent to not automatically use `enable_parallel_projection` when selecting 2D styles.  I followed that precedent here.  It is called out explicitly in the docstring and the example.

I'm not sure if the keyboard bindings are also important here, the current PR inherits from `vtkInteractorStyleTrackballCamera` and modifies the three mouse button events to preserve any other key bindings that might be expected.


MatthewFlamm:
I don't see tests for any of the other interactors, or I can't find them in a quick search.  I will continue looking.  I think a functional test is very important for this PR.  @FrankFrank9 are you able to test out this PR?

MatthewFlamm:
Another thought I had while developing this was that we could make a generic mouse binding helper interactor that would enable something like this:

```py
import pyvista as pv
pl = pv.Plotter()
pl.enable_custom_style(left_button="pan", middle_button="rotate", right_button="dolly", wheel="dolly")
pl.add_mesh(pv.Cube())
pl.show()
```

tkoyama010:
> Another thought I had while developing this was that we could make a generic mouse binding helper interactor that would enable something like this:
> 
> ```python
> import pyvista as pv
> pl = pv.Plotter()
> pl.enable_custom_style(left_button="pan", middle_button="rotate", right_button="dolly", wheel="dolly")
> pl.add_mesh(pv.Cube())
> pl.show()
> ```

+1 to this idea; if we implement it, we can refactor it so that all styles are defined by that function.

MatthewFlamm:
> +1 to this idea; if we implement it, we can refactor it so that all styles are defined by that function.

Right now, I think this might only clean up 1-2 styles.  Most styles are just wrappers around an existing style.  But this is a good point, we could also consider adding this as a `**kwargs` to all styles as an override to customize the styles.  I think this particular style warrants a built-in style still, so this can wait for a separate PR.

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,6 +547,109 @@ def enable_trackball_style(self):
         self._style_class = None
         self.update_style()
 
+    def enable_paraview_2D_style(self):
+        """Set the interactive style to Paraview-like 2D style.
```
MatthewFlamm, line 551:
```suggestion
        """Set the interactive style to ParaView-like 2D style.
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,6 +547,109 @@ def enable_trackball_style(self):
...
+
+        For a 3-button mouse, the left button pans, the
+        right button dollys, the middle button spins, and the wheel
+        dollys.
+        ctrl + left button spins, shift + left button zooms,
+        ctrl + middle button pans, shift + middle button dollys,
+        ctrl + right button rotates in 3D, and shift + right button
+        dollys.
+
+        Shift + right currently varies from Paraview and may be changed
```
MatthewFlamm, line 561:
```suggestion
        Shift + right currently varies from ParaView and may be changed
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,6 +547,109 @@ def enable_trackball_style(self):
...
+        in the future.
+
+        To get a view most like Paraview 2D style, also use
```
MatthewFlamm, line 564:
```suggestion
        To get a view most like ParaView 2D style, also use
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,6 +547,109 @@ def enable_trackball_style(self):
...
+        :func:`pyvista.Plotter.enable_parallel_projection`.
+
+        Examples
+        --------
+        Create a simple scene with a plotter that has the
+        Paraview-like 2D style:
```
MatthewFlamm, line 570:
```suggestion
        ParaView-like 2D style:
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,6 +547,109 @@ def enable_trackball_style(self):
...
+                self._style_class.StartRotate()
+            elif self.interactor.GetShiftKey():
+                self._style_class.StartDolly()  # Paraview does a different type of zooming
+                self._style_class.StartDolly()
+            self._style_class.OnRightButtonDown()
+
+        def _right_button_release_callback(_obj, event):  # pragma: no cover
+            """Turn off all right button behavior."""
+            self._style_class.EndRotate()
+            self._style_class.EndUniformScale()
```
MatthewFlamm, line 631:
```suggestion
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,6 +547,109 @@ def enable_trackball_style(self):
...
+            self._style_class.EndDolly()
+            self._style_class.OnMiddleButtonUp()
+
+        def _right_button_press_callback(_obj, event):  # pragma: no cover
+            """Right button press behavior."""
+            if self.interactor.GetControlKey():
+                self._style_class.StartRotate()
+            elif self.interactor.GetShiftKey():
+                self._style_class.StartDolly()  # Paraview does a different type of zooming
+                self._style_class.StartDolly()
```
MatthewFlamm, line 625:
```suggestion
            else:
                self._style_class.StartDolly()
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,6 +547,109 @@ def enable_trackball_style(self):
...
+            self._style_class.EndPan()
+            self._style_class.EndDolly()
+            self._style_class.OnMiddleButtonUp()
+
+        def _right_button_press_callback(_obj, event):  # pragma: no cover
+            """Right button press behavior."""
+            if self.interactor.GetControlKey():
+                self._style_class.StartRotate()
+            elif self.interactor.GetShiftKey():
+                self._style_class.StartDolly()  # Paraview does a different type of zooming
```
MatthewFlamm, line 624:
```suggestion
                self._style_class.StartDolly()  # ParaView does a different type of zooming
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,6 +547,109 @@ def enable_trackball_style(self):
...
+            self._style_class.EndPan()
+            self._style_class.EndDolly()
+            self._style_class.OnMiddleButtonUp()
+
+        def _right_button_press_callback(_obj, event):  # pragma: no cover
+            """Right button press behavior."""
+            if self.interactor.GetControlKey():
+                self._style_class.StartRotate()
+            elif self.interactor.GetShiftKey():
+                self._style_class.StartDolly()  # ParaView does a different type of zooming
```
MatthewFlamm, line 624: ParaView 2D style does some sort of different type of zooming that I cannot figure out. I tried some of the other `Start*` modes available but some did nothing me and the documentation on these methods is sparse in vtk.

I put in a note that this behavior could change in the future

tkoyama010: Let's add the PR number so we can see this comment later.
```suggestion
                self._style_class.StartDolly()  # ParaView does a different type of zooming (#5336)
```

MatthewFlamm: Will this link appropriately, should we put in a direct link instead?

tkoyama010: Sure. If you don't mind the length of it.
```suggestion
                self._style_class.StartDolly()  # ParaView does a different type of zooming (https://github.com/pyvista/pyvista/pull/5336#discussion_r1426060132)
```

tkoyama010: Or...
```suggestion
                # ParaView does a different type of zooming
                # https://github.com/pyvista/pyvista/pull/5336#discussion_r1426060132
                self._style_class.StartDolly()
```

MatthewFlamm:
Moving to draft until I can figure out testing.  I have some ideas.

MatthewFlamm:
I'm not sure why it says that the lines controlling ctrl- or shift- click are covered by tests, but they are not.

In trying to add coverage, I expected that these lines in the test would result in the ctrl-left click behavior, but it does not

```py
    pl = setup_plot()
    pl.show(auto_close=False)
    pl.iren.interactor.SetControlKey(1)
    pl.iren._mouse_left_button_press(100, 100)
    pl.iren._mouse_left_button_release(100, 150)
    pl.iren.interactor.SetControlKey(0)
    pl.close()
```

The control key appears to be pressed when debugging using `pl.iren.interactor.GetControlKey()`, but it disappears once the `_mouse_left_button_press` function is called when debugged.  Needs investigation.

pyvista/plotting/render_window_interactor.py
```diff
@@ -912,6 +1016,32 @@ def _mouse_left_button_click(self, x=None, y=None, count=1):
...
+
+    def _mouse_middle_button_release(
+        self, x=None, y=None
+    ):  # pragma: no cover # numpydoc ignore=PR01,RT01
+        """Simulate a middle mouse button release."""
+        if x is not None and y is not None:
+            self._mouse_move(x, y)
+        self.interactor.MiddleButtonReleaseEvent()
+
+    def _mouse_middle_button_click(self, x=None, y=None, count=1):
```
MatthewFlamm, line 1040: This function is not needed for this code, and it is only used for testing purposes I think.  I propose that we don't need to have coverage for these lines at this time.

MatthewFlamm:
> The control key appears to be pressed when debugging using `pl.iren.interactor.GetControlKey()`, but it disappears once the `_mouse_left_button_press` function is called when debugged. Needs investigation.

https://github.com/pyvista/pyvista/blob/b960c2257dd03eaa18b9686f7a48ee30fa779b69/pyvista/plotting/render_window_interactor.py#L941-L944

`SetEventInformation` has an optional parameters, and by default, sets all keys to be off.  We should either expose the ability to manage the other parameters in the `_mouse_*_button_*` methods, or switch to another method like `SetEventPosition`.

pyvista/plotting/render_window_interactor.py
```diff
@@ -940,7 +1086,7 @@ def _mouse_right_button_click(self, x=None, y=None, count=1):
 
     def _mouse_move(self, x, y):  # pragma:
         """Simulate moving the mouse to ``(x, y)`` screen coordinates."""
-        self.interactor.SetEventInformation(x, y)
+        self.interactor.SetEventPosition(x, y)
```
MatthewFlamm, line 1089: This change was because `SetEventInformation` resets all the controlling key presses, while `SetEventPosition` does not.

tests/plotting/test_plotting.py
```diff
@@ -3811,3 +3811,98 @@ def test_voxelize_volume():
...
+    pl.close()
+
+    # ctrl left click spins, image 4
+    pl = setup_plot()
+    pl.show(auto_close=False)
+    pl.iren._control_key_press()
+    pl.iren._mouse_left_button_press(*start)
+    pl.iren._mouse_left_button_release(*spin)
+    pl.iren._control_key_release()
+    pl.close()  # image 4
```
tkoyama010, line 3863:
```suggestion
    pl.close()
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -912,6 +1032,32 @@ def _mouse_left_button_click(self, x=None, y=None, count=1):
...
+
+    def _mouse_middle_button_release(
+        self, x=None, y=None
+    ):  # pragma: no cover # numpydoc ignore=PR01,RT01
+        """Simulate a middle mouse button release."""
+        if x is not None and y is not None:
+            self._mouse_move(x, y)
+        self.interactor.MiddleButtonReleaseEvent()
+
+    def _mouse_middle_button_click(self, x=None, y=None, count=1):
```
tkoyama010, line 1056:
```suggestion
    def _mouse_middle_button_click(self, x=None, y=None, count=1):  # pragma: no cover
```

tests/plotting/test_plotting.py
```diff
@@ -3811,3 +3811,98 @@ def test_voxelize_volume():
...
+    # baseline, image
+    pl = setup_plot()
+    pl.show()
+
+    start = (100, 100)
+    pan = rotate = (150, 150)
+    spin = (100, 150)
+    dolly = (100, 50)
+
+    # left click pans, image 1
```
tkoyama010, line 3835: Checking if the image is in 2D view in the test is a great idea. Please explain in your comments what images you expect to be displayed by each operation. I can generally describe what we check in the images, but comments would give me more confidence.

tkoyama010 (changes requested):
Thanks for this! Your changes are the best.

pyvista/plotting/render_window_interactor.py
MatthewFlamm: I'm okay with this. Do we have a general policy about when to use this?  I'm unsure when we would prefer this over leaving it uncovered so that future PRs can raise coverage.  We shouldn't add this just to see a higher number.

tkoyama010: I was going to put this part on to make the reader aware that it was not tested, but then I realized that it can be confirmed by going to the codedev page. It is fine as it is.

I would like to make an issue to organize a project-wide policy on skipping coverage.

MatthewFlamm: I don't think it is necessarily bad, but it isn't clear to me what it means when we use this in this package.  Some people might expect that it means that there is a reason we aren't testing it.

tests/plotting/test_plotting.py
MatthewFlamm: I added some comments, hopefully it makes sense.  I also exaggerated more the dolly images since those were more subtle.

tkoyama010 (dismissed):
LGTM

MatthewFlamm:
I'm wondering if we should add this to the see also of `enable_parallel_projection` and vice versa.

tkoyama010 (approved):
Thanks!

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,6 +547,110 @@ def enable_trackball_style(self):
         self._style_class = None
         self.update_style()
 
+    def enable_paraview_2d_style(self):
```
banesullivan, line 550: I'm wondering if we should just call this `enable_2d_style()`?

pyvista/plotting/render_window_interactor.py
banesullivan: Good catch!

tests/plotting/test_plotting.py
```diff
@@ -3811,3 +3811,104 @@ def test_voxelize_volume():
     # Create a voxel volume from unequal density dimensions and plot result.
     vox = pv.voxelize_volume(mesh, density=[0.15, 0.15, 0.5])
     vox.plot(scalars='InsideMesh', show_edges=True, cpos=cpos)
+
+
+def test_paraview_2d_style():
```
banesullivan, line 3816: I'm wondering if it makes sense to set `show_scalar_bar=False` for all these tests. I've been thinking lately that we out to make the image cache focus on the key feature we are testing and leave out some things like scalar bars that might be noise and overpower subtle changes that we may want to detect in the future. I don't know 🤷🏻

banesullivan (commented):
This overall looks great! Minor feedback that I'll leave up to you if you want to address. 

Though I do sort of feel strongly that this should be "enable_2d_style" rather than "enable_paraview_2d_style"... let me know what you think about this and I'll be happy to rubber stamp

pyvista/plotting/render_window_interactor.py
MatthewFlamm: I'm +1 on this suggestion.  This also frees us from matching ParaView behavior 100%.  We can still mention that it is similar to the behavior in ParaView.

tests/plotting/test_plotting.py
MatthewFlamm: I agree.  The scalar bar provides a point of reference for size and position here.  But it is more of an accident than on purpose.  It also means any subtle change to scalar bar will break these tests, which is not good.  I will check that the images are still understandable for review of behavior without them.

MatthewFlamm: Done in https://github.com/pyvista/pyvista/pull/5336/commits/b51baa89bc238caefb94c83376e135de82e8b1ab

pyvista/plotting/render_window_interactor.py
MatthewFlamm: Using this now in https://github.com/pyvista/pyvista/pull/5336/commits/4e38f0de087521ada6c7901e89caf27a72be353c

banesullivan (approved):
Thanks for making those changes!

tkoyama010 (approved):

PR #5337, opened 2023-12-13
`CODEOWNERS` assignments requires admin approval

tkoyama010:
### Overview

`CODEOWNERS` assignments requires admin approval.

### Details

- None


banesullivan (approved):
Good call

Issue #5334, opened 2023-12-13
Fix `vtk` segmentation errors in integration tests or only support Trame for VTK 9.2+


tkoyama010:
### Describe what maintenance you would like added.

Our CI frequently gets `vtk` segmentation errors during unit test runs. We have to re-run with "re-run failed job" to make the test pass, but this is manual and very tedious. We need to fix `vtk` segmentation errors in integration tests or only support Trame for VTK 9.2+.

This is an ongoing issue. Let's report in this thread about CI environment etc when we have reruns error.

### Links to source code.

https://github.com/pyvista/pyvista/blob/main/.github/workflows/integration-tests.yml

### Pseudocode or Screenshots

None

beroda:
To help diagnose flaky tests, the [`pytest-repeat`](https://github.com/pytest-dev/pytest-repeat) plugin might be useful.
```python
import pytest
pytestmark = pytest.mark.repeat(3)

def test_one():
    ...

@pytest.mark.repeat(6)
def test_two():
    ...
```
(My two cents)

tkoyama010:
> To help diagnose flaky tests, the [`pytest-repeat`](https://github.com/pytest-dev/pytest-repeat) plugin might be useful.
> 
> ```python
> import pytest
> pytestmark = pytest.mark.repeat(3)
> 
> def test_one():
>     ...
> 
> @pytest.mark.repeat(6)
> def test_two():
>     ...
> ```
> 
> Moreover, [`pytest` doc](https://docs.pytest.org/en/7.1.x/explanation/flaky.html) has good resources concerning flaky tests
> 
> (My two cents)

Thanks for the comment. I would like to +1 this idea. This is because it fulfills my following request.
1. It can be used for local unstable testing.
1. No need to make changes to GitHub Actions.
1. No need to make changes to the test target by the reason of flaky test. I may only adopt Trame for VTK 9.2+ in the future, but I want to separate that discussion.

tkoyama010:
`tests/plotting/test_widgets.py::test_affine_widget` in Linux Unit Testing (3.10, 9.2.2) raise vtk error happen in #5336. We should list it as flaky.

MatthewFlamm:
> `tests/plotting/test_widgets.py::test_affine_widget` in Linux Unit Testing (3.10, 9.2.2) raise vtk error happen in #5336. We should list it as flaky.

[...]

PR #5345, opened 2023-12-15
Fix Error: Not all objects GCed in `test_affine_widget` function

tkoyama010:
### Overview

Ad reported in #5334, we sometimes have a GC error in `test_affine_widget` function.
I suspect it is because `interact_calls` and `release_calls` is not removed.
I hope this change will reduce errors in the `test_affine_widget` function.

### Details

- None

### DIFF
--- a/tests/plotting/test_widgets.py
+++ b/tests/plotting/test_widgets.py
@@ -583,6 +583,9 @@ def test_affine_widget(sphere):
     assert not widget._circles
     assert not widget._arrows
 
+    interact_calls = []
+    release_calls = []
+
 
 def test_logo_widget(verify_image_cache):
     pl = pv.Plotter()


MatthewFlamm:
The root cause error is not the GC error.  There is a previous error in the test that breaks the GC of vtk objects, so there is an additional error during teardown.


https://github.com/pyvista/pyvista/actions/runs/7204624222/job/19626361240#step:13:1702
```txt
# test Z axis translation
        pl.iren._mouse_left_button_press(width // 2, height // 2 + 5)
>       assert widget._selected_actor is widget._arrows[2]
E       assert Actor (0x7f153ea559c0)\n  Center:                     (-0.5, 37.246177673339844, 0.0)\n  Pick
```

tkoyama010:
> The root cause error is not the GC error. There is a previous error in the test that breaks the GC of vtk objects, so there is an additional error during teardown.
> 
> https://github.com/pyvista/pyvista/actions/runs/7204624222/job/19626361240#step:13:1702
> 
> ```
> # test Z axis translation
>         pl.iren._mouse_left_button_press(width // 2, height // 2 + 5)
> >       assert widget._selected_actor is widget._arrows[2]
> E       assert Actor (0x7f153ea559c0)\n  Center:                     (-0.5, 37.246177673339844, 0.0)\n  Pick
> ```

Thanks for pointing this out. Looks like we need to look at this a bit more closely.

Issue #5338, opened 2023-12-14
Charts may not be fully deleted


Rmelikyan:
### Describe the bug, what's wrong, and what you expected.

tl;dr: Calling remove chart does not actually delete the chart and all references to it. This led to unintended behavior where if a new chart was created and placed on top of old (not fully deleted chart) then the new chart would default to bottom lefthand position rather than input position. Attached code shows minimum working example of behavior.

I have been working on a pyvista app which allows a user to flick through time series point-cloud data using key bindings. I've been progressively upgrading the program to include more functionality such as changing scalar-data and colormap with button widgets and most recently displaying an overlaying chart that should update along with the updating point cloud. My naive first approach was, at each time step (call of `plotter.update()`) I would generate the figure of interest with matplotlib, delete the old pyvista chart and create and add a new chart with `pyvista.ChartMPL` and `plotter.add_chart`. Oddly, the first plot always looked great, but the second would never show up in the specified location, instead would default to the bottom lefthand corner. I got the suspicion that maybe the old chart wasn't being deleted properly (`del(chart)` throws errors and `plotter.remove_chart()` didn't do the trick). So I started digging into this and found that even after calling the remove_chart function, that the renderer still thought it had a chart (`renderer.has_charts` gave True).

Looking into the source, I could find no way to actually entirely delete a chart from a plotter instance.

Luckily, I realized that the chart `redraw_on_render` attribute, which I had turned off for increased performance when the chart was not visible (my chart button is also the toggle), was actually the solution I was looking for. So now I just update the matplotlib figure on each render which updates the chart.

That being said, I still wanted to provide the below example and inquire if this is the intended behavior.

Thanks!

### Steps to reproduce the bug.

```
import pyvista as pv
pl = pv.Plotter()
assert not pl.renderer.has_charts
chart = pv.Chart2D()
chart.plot(range(10), range(10))
pl.add_chart(chart)
assert pl.renderer.has_charts
print("A chart correctly exists and can be accessed")
print(pl.renderer._charts)
print(pl.renderer._charts[0])
pl.remove_chart(chart)
try:
    assert not pl.renderer.has_charts
except Exception as error:
    print(type(error).__name__)
    print(f"Called pl.remove_chart but still renderer.has_charts == {pl.renderer.has_charts}")
    print(pl.renderer._charts)
print("But the chart cannot be accessed due to index error")
try:
    print(pl.renderer._charts[0])
except Exception as error:
    print(type(error).__name__)
```

### System Information

[...]

PR #5356, opened 2023-12-16
Simplify charts bookkeeping and update `has_charts`

dcbr:
### Details

Simplify the charts bookkeeping in the `Renderer` class. The `_charts` attribute is now either `None` (no charts have been added yet) or a valid `Charts` object (at least one chart has been added so far, but could have been removed again). The lazy instantiation is still there, but made explicit in the `add_charts` method and no longer hidden behind the `_charts` property proxying the `__charts` attribute.
The `has_charts` property was updated such that it makes more intuitive sense to end users, see #5338. It is now `True` only when there is at least one chart in the `_charts` collection.


tkoyama010:
@dcbr
Thank you for this change.
I am not sure whether to label this change as [enhancement](https://github.com/pyvista/pyvista/labels/enhancement) or [bug](https://github.com/pyvista/pyvista/labels/bug) or [maintenance](https://github.com/pyvista/pyvista/labels/maintenance). What is your perception as the author? (Sorry for the noise comment. But, it is needed to create release notes for the next version.)

dcbr:
@tkoyama010 I just added the maintenance label, I think that best fits these changes (low impact, mostly "under the hood" changes).

tkoyama010 (approved):
LGTM

PR #5360, opened 2023-12-18
Deploy docs from GitHub Actions

banesullivan:
Work to automate the documentation deployment for #5344 

This re-enables deployment of our development docs from the main branch and will generate docs previews for all PRs.

Work to deploy release docs will come later

PR #5363, opened 2023-12-18
More force static docs additions

MatthewFlamm:
### Overview

I looked through some of the examples with the longest execution times and found these examples are either not rendering correctly in interactive mode or don't add anything, so I have turned them off.

### Details

In [backface properties](https://docs.pyvista.org/version/stable/examples/02-plot/backface_props,) the back faces do not have different colors in interactive mode.

In [depth of field](https://docs.pyvista.org/version/stable/examples/02-plot/depth_of_field), there is nothing wrong with the one plot that is rendered with interactive mode, but that plot is only useful when contrasted with the second plot that does not work.  We should just make the whole page static.


tkoyama010:
github-actions preview

banesullivan (approved):
Good catches, thank you!

banesullivan:
> ~github-actions preview~ I had forgotten that the build was disabled.

I manually disabled this for now. Unfortunately all documentation deployments must go through me, manually, until I finish #5360

PR #5369, opened 2023-12-20
Add support to `meshio`'s `polyhedronX` cell types

keurfonluu:
### Overview

Adding support to `meshio`'s `polyhedronX` cell types. 

### Details

- Fixed: `pyvista.from_meshio` and `pyvista.save_meshio` now supports polyhedral cells.


tkoyama010:
pre-commit.ci autofix

MatthewFlamm:
Looks great so far.  Is it possible for you to add a test for this?

pyvista/core/utilities/fileio.py
```diff
@@ -552,27 +565,34 @@ def save_meshio(filename, mesh, file_format=None, **kwargs):
...
-            if cell_type not in pixel_voxel
-            else cell[[0, 1, 3, 2]]
-            if cell_type == 8
-            else cell[[0, 1, 3, 2, 4, 5, 7, 6]]
-        )
-        cell_type = cell_type if cell_type not in pixel_voxel else cell_type + 1
-        cell_type = vtk_to_meshio_type[cell_type] if cell_type != 7 else f"polygon{numnodes}"
+    for i, (offset, cell_type) in enumerate(zip(vtk_offset, vtk_cell_type)):
+        if cell_type == 42:
+            cell_ = mesh.get_cell(i)
```
keurfonluu, line 570: Not sure if there is a more efficient way to do that without having to deal with a deep copy. I don't see any grid attribute that exposes the faces of a cell.

pyvista/core/utilities/fileio.py
```diff
@@ -552,27 +565,34 @@ def save_meshio(filename, mesh, file_format=None, **kwargs):
...
 
         if len(cells) > 0 and cells[-1][0] == cell_type:
             cells[-1][1].append(cell)
         else:
             cells.append((cell_type, [cell]))
 
     for k, c in enumerate(cells):
-        cells[k] = (c[0], np.array(c[1]))
+        if not c[0].startswith("polyhedron"):
+            cells[k] = (c[0], np.array(c[1]))
```
tkoyama010, line 595: One small question. Does this mean the same as `cell_type == 42` in the previous for loop? If so, it would be easier for the reviewer to read if the statement of if is branched in one place instead of two places in my mind :)

keurfonluu: I checked ``meshio`` source code, and these lines are redundant and can be removed as it is already done when initiating a ``meshio.Mesh``.

tkoyama010 (approved):
LGTM. Thanks for this!

PR #5374, opened 2023-12-21
Add `flake8-unused-arguments` check using `ruff`

### LINKED ISSUE #3751: Unused arguments are not being caught by flake8 check

I just noticed there are a handful of filters in `pyvista/core/filters/data_set.py` that do not use their `progress_bar` keyword argument. IMO, there should never be unused arguments. The flake8 check should be picking up on this, but it isn't.

### Links to source code.

See `extract_largest` and `split_bodies` for examples.

https://github.com/pyvista/pyvista/blob/3db44d72070ccd6e8fc5c7f708500424011bcc47/pyvista/core/filters/data_set.py#L2235

https://github.com/pyvista/pyvista/blob/3db44d72070ccd6e8fc5c7f708500424011bcc47/pyvista/core/filters/data_set.py#L2275

### Pseudocode or Screenshots

_No response_

tkoyama010:
### Overview

resolve #3751

### Details

- None

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -286,6 +286,7 @@ indent-width = 4
 target-version = 'py39'
 
 [tool.ruff.lint]
+select = ["E", "F", "ARG"]
 ignore = [
     # whitespace before ':'
     "E203",


pyproject.toml
```diff
@@ -280,6 +280,8 @@ exclude = [
     'dist',
     'doc/examples',
     'doc/_build',
+    'examples',
+    'tests',
```
tkoyama010, line 284: Added to exclude flake8-unused-arguments check, but this setting also excludes other ruff checks which are not good. But there is no other way.

tkoyama010:
```suggestion
    # https://github.com/pyvista/pyvista/pull/5374/files#r1433366947
    'examples',
    'tests',
```

banesullivan (commented):
I'm a little nervous about how many files are being modified here... Perhaps we should ignore the entire `tests/` directory for these `# noqa: ARG005` etc?

tkoyama010:
> I'm a little nervous about how many files are being modified here... Perhaps we should ignore the entire `tests/` directory for these `# noqa: ARG005` etc?

Good point.
But, the ruff is not currently designed to exclude files at each rule. So, I'm thinking about excluding ARG005 from the check.
https://docs.astral.sh/ruff/rules/unused-lambda-argument/
Even if we don't check this, it doesn't affect our project quality in my mind.

pyvista/ext/plot_directive.py
```diff
@@ -345,7 +345,7 @@ class PlotError(RuntimeError):
     pass
 
 
-def _run_code(code, code_path, ns=None, function_name=None):
+def _run_code(code, code_path, ns=None, function_name=None):  # noqa ARG001
```
tkoyama010, line 348:
```suggestion
def _run_code(code, code_path, ns=None, function_name=None):  # noqa: ARG001
```

pyvista/plotting/plotter.py
```diff
@@ -3674,6 +3674,7 @@ def add_mesh(
                 lighting=lighting,
                 render=False,
                 show_vertices=False,
+                edge_opacity=edge_opacity,
```
tkoyama010, line 3677:
```suggestion
```

PR #5375, opened 2023-12-21
Add `from_irregular_faces` to `PolyData`

darikg:
cc @tkoyama010 

This adds from_irregular_cells to `CellArray`, and the (settable) property `irregular_faces` and corresponding alternate constructor `from_irregular_faces` to `PolyData`.

PR #5376, opened 2023-12-21
Check that arrays with padding are valid

### LINKED ISSUE #5217: Check that arrays with padding are valid

Hi !

I realized that `PolyData` can be initialized with `strips`, `faces` or `lines` that are invalid, as in the following examples :

```python
import pyvista as pv
import numpy as np

faces = np.array([3, 0, 1, 2, 3, 0])
points = np.random.rand(4, 3)
mesh = pv.PolyData(points, faces=faces)
print(mesh.faces)
>>> [   3    0    1    2    3    0   64 1009]

strips = np.array([5, 0, 1, 2])
points = np.random.rand(4, 3)
print(mesh.strips)
>>> [   5    0    1    2   48 2529]

lines = np.array([5, 0, 1, 2])
points = np.random.rand(4, 3)
mesh = pv.PolyData(points, lines=lines)
print(mesh.lines)
>>> [   5    0    1    2   48 4609]
```

It seems that the memory is read outside the numpy arrays. I wrote a checker to assess that an array with padding is valid :

```python
def is_valid(list_padding):
    """Determines if a list with padding is valid."""
    if len(list_padding) == 0:
        return True
    
    offset = list_padding[0]
    if len(list_padding) < offset + 1:
        return False

    return is_valid(list_padding[offset + 1:])
```

Do you think it could be a good idea to integrate it somewhere ? (in [`validation`](https://github.com/pyvista/pyvista/pull/5100) ?) Let me know, I can work on the integration if requested.

_No response_

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

Louis-Pujol:
### Overview

Resolves #5217 . When attributes `strips`, `verts`, `lines` or `faces` were set on a `PolyData` object, no check were made about the validity of the length of the array with respect to the padding information (expected size of successive chunks of data). No error was raised and memory was read outside of the scope, possibly leading to memory leaks. This PR solves this by adding checks and run them when `PolyData` is initialized or modified

### Details

In the module `pyvista.core.input_validation`
- add a function `check_padding` to check an array's padding information validity and raise an error if needed
- add an argument `must_be_padding` (default `False`) to `validate_array`

In the module `pyvista.core.pointset`
- call `validate_array` with `must_be_padding=True` to check the validity of `strips`, `verts`, `lines` or `faces`. Calls are made in `__init__()` or in appropriate setters.

Added two tests
- One in `test_input_validation` to validate the function `check_padding`
- One in `test_polydata` to validate that it is no longer possible to pass arrays with invalid padding to `PolyData`


darikg:
I like having the validation available, but I'm concerned about the performance implications for constructing large meshes

Louis-Pujol:
@darikg thanks for your remark.

I've done a little test to have data on running time with/without padding check (prior to that I added a `check_padding` boolean option to `PolyData` constructor in order to be able to disable the test).

I check the running time for creating a PolyData
- with `1,000,000` points
- with `1,000,000` points + `n` triangles without checking the padding
- with `1,000,000` points + `n` triangles + padding check

`n` varying from `1` to `10,000,000`. Here is the code :

```python
import numpy as np
import pyvista as pv

n_points = int(1e6)
n_triangles_list = np.logspace(1, 7, 20).astype(int)

n_triangles_list[-1]

times_only_points = []
times_points_triangles = []
times_points_triangles_check_padding = []

points = np.random.rand(n_points, 3)

import time
for n_triangles in n_triangles_list:

    points = np.random.rand(n_points, 3)
    faces = np.tile([3, 0, 1, 2], reps=n_triangles)
    start = time.time()
    mesh = pv.PolyData(points)
    end = time.time()
    times_only_points.append(end - start)

    points = np.random.rand(n_points, 3)
    start = time.time()
    mesh = pv.PolyData(points, faces=faces, check_padding=False)
    end = time.time()
    times_points_triangles.append(end - start)

    points = np.random.rand(n_points, 3)
    start = time.time()
    mesh = pv.PolyData(points, faces=faces, check_padding=True)
    end = time.time()
    times_points_triangles_check_padding.append(end - start)

plt.plot(n_triangles_list, times_only_points, label="Points")
plt.plot(n_triangles_list, times_points_triangles, label="Points + Triangles")
plt.plot(n_triangles_list, times_points_triangles_check_padding, label="Points + Triangles + Check padding")

plt.legend()
plt.xlabel("Number of triangles")
plt.ylabel("Time (s)")
#x in log scale
plt.xscale("log")
plt.show()
```

And the output :

output

The ratio of initialization time without/with padding check is close to `0.5`:

```python
for i, n in enumerate(n_triangles_list):
    print(f"Number of triangles : {n}, ratio: {times_points_triangles[i] / times_points_triangles_check_padding[i]}")
```

Output:
```
Number of triangles : 10, ratio: 0.57
Number of triangles : 20, ratio: 0.70
Number of triangles : 42, ratio: 0.71
Number of triangles : 88, ratio: 0.72
Number of triangles : 183, ratio: 0.73
Number of triangles : 379, ratio: 0.63
Number of triangles : 784, ratio: 0.58
Number of triangles : 1623, ratio: 0.61
Number of triangles : 3359, ratio: 0.53
Number of triangles : 6951, ratio: 0.52
Number of triangles : 14384, ratio: 0.52
Number of triangles : 29763, ratio: 0.51
Number of triangles : 61584, ratio: 0.51
Number of triangles : 127427, ratio: 0.51
Number of triangles : 263665, ratio: 0.51
Number of triangles : 545559, ratio: 0.51
Number of triangles : 1128837, ratio: 0.50
Number of triangles : 2335721, ratio: 0.51
Number of triangles : 4832930, ratio: 0.51
Number of triangles : 10000000, ratio: 0.51
```

I don't know if this can be considered as a problematic overhead, but if it is the case I see two solutions:
- Make the padding check optional (via a global variable or an argument in PolyData constructor)
- Improve the padding checker by compiling a cython module (or a JIT-compilation solution) instead of using a python while loop

darikg:
Oh yeah, that reminds me of this old [bottleneck](https://github.com/pyvista/pyvista/blob/6d1583bb51aecea964bde64107d8b2641c1f456b/pyvista/core/cell.py#L630) that I think is scanning the array in more or less the exact same way? Wonder if they could be combined/merged

Louis-Pujol:
Thanks @darikg for pointing this ! Your right that my code is redundant with

https://github.com/pyvista/pyvista/blob/6d1583bb51aecea964bde64107d8b2641c1f456b/pyvista/core/utilities/cells.py#L14-L35

(actually this implementation is slightly faster than mine thanks to the use of clever data structures).

So in the current implementation of this PR :
- if none of `n_faces`, `n_strips`, `n_faces`, or `n_verts` are provided: the `check_padding` will run but not `ncells_from_cells`, then the total running time will be approximately 1x the python loop
- if one of `n_faces`, `n_strips`, `n_faces`, or `n_verts` is provided: the `check_padding` will run as well as `ncells_from_cells`, then the total running time will be approximately 2x the python loop. There is clearly a redudancy here as we run through the array twice

One option is to remove function `check_padding` and uses `ncells_from_cells` to raise an error if the dimension of the array is not correct. The limitation of this solution is that padding will not be checked anymore if `n_faces` is specified. If we call `ncells_from_cells` even with `n_faces`, there will be no speedup anymore.

I made a cython version of `n_cells_from_cells` (with a `ValueError` if size is incorrect). It is available [here](https://github.com/Louis-Pujol/cython_ncells_from_cells). It is significantly faster than the python implementation. It seems reasonable to me to use the cython version in pyvista, what are your thoughts about that ?

darikg:
> One option is to remove function check_padding and to use ncells_from_cells to raise an error if the dimension of the array is not correct. The limitation of this solution is that padding will not be checked anymore if n_faces is specified.

One of the ways n_faces is specified is by passing faces as an (n_faces, face_size + 1) array, right? In which case I suspect if it's possible to take advantage of that structure for padding validation somehow since you wouldn't have to iterate over each face. But that still leaves the case where the user supplies faces as a 1D list *and* specifies n_faces.

> I made a cython version of n_cells_from_cells (with a ValueError if size is incorrect). It is available [here](https://github.com/Louis-Pujol/cython_ncells_from_cells). It is significantly faster than the python implementation. It seems reasonable to me to use the cython version in pyvista, what are your thoughts about that ?

> ratio: 5.776

wow. In this benchmark you're only comparing n_cells_from_cells, and in the earlier profiling in this PR you were measuring overall PolyData construction, but am I right in thinking that n_cells_from_cells dominates the PolyData construction, and that this would improve PolyData construction times *and* provide better user QoL?

tests/core/test_input_validation.py
```diff
@@ -949,3 +950,27 @@ def test_validate_axes_orthogonal(bias_index):
...
+    with pytest.raises(ValueError):
+        check_padding(invalid_padding)
+
+    invalid_padding = [3, 0]
+    with pytest.raises(ValueError):
+        check_padding(invalid_padding)
+
+    invalid_padding = [3]
+    with pytest.raises(ValueError):
+        check_padding(invalid_padding)
```
tkoyama010, line 976: `@pytest.mark.parametrize` makes the expression of the test code more concise.
```suggestion


@pytest.mark.parametrize("padding", [[3, 0, 1, 2, 2, 1, 2, 4, 1, 2, 3, 8], [], [1, 0]])
def test_check_valid_padding(padding):
    check_padding(padding)


@pytest.mark.parametrize("padding", [[8, 0, 1, 2, 3, 4], [3, 0, 1], [3, 0]])
def test_check_invalid_padding(padding):
    with pytest.raises(ValueError):
        check_padding(padding)
```

tkoyama010 (changes requested):
Thanks for this!

tkoyama010:
pre-commit.ci autofix

tests/core/test_polydata.py
```diff
@@ -1107,3 +1107,34 @@ def test_regular_faces_mutable():
...
+)
+def test_init_with_invalid_padding(arr):
+    """This test checks that an error is raised when the padding is invalid.
+
+    More precisely, if a PointSet is created with ``verts``, ``lines``, ``faces``
+    or ``strips`` we are expecting a ValueError to be raised if the array length
+    is not compatible with the padding structure.
+
+    """
+
```
tkoyama010, line 1129: The function name is sufficient to read the contents, so let's remove the docstring as you would any other test function.
```suggestion
```

pyvista/core/pointset.py
```diff
@@ -689,15 +690,29 @@ def __init__(
             # one cell per point (point cloud case)
             verts = self._make_vertex_cells(self.n_points)
             n_verts = self.n_points
+        elif verts is not None:
+            # If verts are specified, then we check the padding
```
tkoyama010, line 694:
```suggestion
```

pyvista/core/pointset.py
```diff
@@ -689,15 +690,29 @@ def __init__(
...
+            verts = np.asarray(verts)
+            validate_array(verts.flatten(), must_be_padding=True)
 
         # here we use CellArray since we must specify deep and n_faces, etc.
         if verts is not None:
+            # do not check padding here since we already did it above
```
tkoyama010, line 700:
```suggestion
```

pyvista/core/input_validation/validate.py
```diff
@@ -321,6 +333,14 @@ def validate_array(
         else:
             check_is_sorted(arr_out, name=name)
 
+    if must_be_padding:
+        # check that the array is integer-like and nonnegative
+        check_is_integerlike(arr_out, strict=True, name=name)
+        check_is_nonnegative(arr_out, name=name)
+
+        # check the validity of the padding
+        check_padding(arr_out, name=name)
```
tkoyama010, line 342:
```suggestion
        check_is_integerlike(arr_out, strict=True, name=name)
        check_is_nonnegative(arr_out, name=name)

        check_padding(arr_out, name=name)
```

tkoyama010 (changes requested):

Louis-Pujol:
@darikg for the case where faces are provided as an `(n_faces, face_size + 1)`, the padding check can be written in numpy :
```python
padding = array_2d[0, 0]
assert np.all(array_2d[:, 0] == padding)
assert array_2d.shape[1] == padding + 1
```
It seems to me that `PolyData.from_regular_faces` with faces `(n_faces, face_size)` is the only situation where checking can be safely discarded. Otherwise we need to define what is the best behavior : should we check padding by default, should we propose an "unsafe" but faster mode for reading faces if n_faces is provided ?

I did more simulations, and in particular I compared the "base" time (time for constructing the polydata without any `n_cells_from_cells` or `check_padding` calls) versus the "check" time (time for `n_cells_from_cells` with check) in term of percentage of the total time for creating the PolyData for an increasing number of triangles. There are three situations :
- python version of `n_cells_from_cells` 
- cython version of `n_cells_from_cells` 
- numpy version when faces is (n_faces, face_size + 1)

```
Python:
1000000 points, 100000 triangle faces: total: 0.051 s
 - base: 0.001 (2.1%)
 - padding check: 0.050 (97.9%)
1000000 points, 464158 triangle faces: total: 0.245 s
 - base: 0.005 (2.2%)
 - padding check: 0.240 (97.8%)
1000000 points, 2154434 triangle faces: total: 1.132 s
 - base: 0.036 (3.2%)
 - padding check: 1.096 (96.8%)
1000000 points, 10000000 triangle faces: total: 5.174 s
 - base: 0.183 (3.5%)
 - padding check: 4.991 (96.5%)

Cython:
1000000 points, 100000 triangle faces: total: 0.010 s
 - base: 0.001 (11.1%)
 - padding check: 0.008 (88.9%)
1000000 points, 464158 triangle faces: total: 0.042 s
 - base: 0.005 (12.9%)
 - padding check: 0.036 (87.1%)
1000000 points, 2154434 triangle faces: total: 0.215 s
 - base: 0.036 (16.7%)
 - padding check: 0.180 (83.3%)
1000000 points, 10000000 triangle faces: total: 1.014 s
 - base: 0.183 (18.1%)
 - padding check: 0.830 (81.9%)

2D Numpy:
1000000 points, 100000 triangle faces: total: 0.002 s
 - base: 0.001 (77.8%)
 - padding check: 0.000 (22.2%)
1000000 points, 464158 triangle faces: total: 0.005 s
 - base: 0.004 (80.2%)
 - padding check: 0.001 (19.8%)
1000000 points, 2154434 triangle faces: total: 0.040 s
 - base: 0.035 (87.8%)
 - padding check: 0.005 (12.2%)
1000000 points, 10000000 triangle faces: total: 0.212 s
 - base: 0.190 (89.5%)
 - padding check: 0.022 (10.5%)
```

I hope it helps clarify the situation. We can also to detect if the 1D faces array has regular faces and take advantage of the speed of numpy checking if so:
```python
padding = array[0]
try:
    array_2d = array.reshape(-1, padding + 1)
    assert np.all(array_2d[:, 0] == padding)
    n_faces = array_2d.shape[0]
except:
    n_faces = cython_ncells_from_cells(array)
```

Louis-Pujol:
+ Just had a look at  #5375.

If `from_irregular_faces` is implemented, the situation will be different.  Every combination of faces' length would be possible in a fast and safe mode using `from_regular_faces` or `from_irregular_faces`. (No need to check padding information and `n_faces` is simply the length of the sequence of faces)

Then the easy solution would be to test the validity of the connectivity arrays for `faces`, `strips`, `verts` and `lines` when provided by the user (using fastest cython/numpy implementations of `ncells_from_cells` for performance) while encouraging using `from_regular_faces` or `from_irregular_faces` for large meshes.

In the same spirit, we can imagine having `from_regular_strips`, `from_irregular_strips`, `from_regular_lines`...

Now, I agree with @darikg in #4637 (thank you for the explanations provided in this PR), for me the interleaved format is not the most intuitive for initializing faces. I'm not a VTK user, so I don't know how people accustomed with VTK perceive it. But I imagine that changing the default interface would be a great deal of work and possibly lead to compatibility issues

darikg:
I agree with your thorough analysis!

> In the same spirit, we can imagine having from_regular_strips, from_irregular_strips, from_regular_lines

I'm wondering now if this leads to a too-large api surface and maybe #4637 should've encouraged something like

```python
mesh = PolyData(points, faces=CellArray.from_regular_cells(faces), lines=CellArray.from_regular_cells(lines))
```

but that obviously doesn't need to be addressed here

tkoyama010 (approved):
LGTM. Thank you both for your excellent insights.

Louis-Pujol:
Thanks @tkoyama010 and @darikg  for the feedback !

Just did a last commit for this PR that do not affect the structure/behavior but :
- speed-up the checking in the special case where the connectivity array represents regular faces
- make the documentation more consistent with the one of PolyData by writing "connectivity array" instead of "padding array"

I'll continue to think about the best way to initialize safely and fastly data structures, IMO there is room for improvement in the future.

pyvista/core/pointset.py
```diff
@@ -689,15 +690,27 @@ def __init__(
...
             self.verts = CellArray(verts, n_verts, deep)  # type: ignore
         if strips is not None:
+            if not isinstance(strips, CellArray):
+                strips = np.asarray(strips)
+                validate_array(strips.flatten(), must_be_padded=True)
             self.strips = CellArray(strips, n_strips, deep)  # type: ignore
         if faces is not None:
+            if not isinstance(faces, CellArray):
+                faces = np.asarray(faces)
+                validate_array(faces.flatten(), must_be_padded=True)
```
darikg, line 708: Here and elsewhere you're calling `flatten`, which throws out the shape information, no?

pyvista/core/pointset.py
```diff
@@ -689,15 +690,27 @@ def __init__(
...
             self.faces = CellArray(faces, n_faces, deep)  # type: ignore
```
darikg, line 709: Here and elsewhere, if `isinstance(faces, CellArray)`, no need to construct a new `CellArray`

darikg (changes requested):
Are you intending on adding the Cython stuff in this PR, or a followup? And currently you're still scanning the array in the validation? I thought that was going to be merged into n_cells_from_cells. (My apologies if I've missed something)

Louis-Pujol:
Hi @darikg sorry for being unclear about that.

My intention was to finish this PR with the current implementation and move forward to address the rest (cython + improve the interface) on a further PR. The reason was that I was not sure about the way to integrate cython code in pyvista + as the PolyData interface has changed in the meantime (notably with `from_irregular_faces`), I thought that it could be a temporary solution to let the checker as it was as a first patch to the initial issue.

- I opened a discussion about the integration of cython in pyvista : https://github.com/pyvista/pyvista/discussions/5393
- Also, I thought about the best way to integrate the checker, my main concern was the interaction with the `n_faces` argument: if we check the validity of the connectivity array, the time saved for not computing the number of cells would be lost.


But yesterday I came up with a way more simpler solution that the one proposed in this PR and that does not implies any additional running time:

In `CellArray._set_cells`:
https://github.com/pyvista/pyvista/blob/b594aac86953875b90b91200a586679d0b15468b/pyvista/core/cell.py#L626-L641C20

We can check after `.SetCells` that `vtk` has not read outside of the memory bounded by the numpy array `cells` by asserting that `cells.size == self.cells.size` just after `.SetCells`. It does not change anything in the way that PolyData are initialized and does not implies any extra running time.

What is your opinion about this solution ? If you agree with me the factit is a better idea than the `check_padding` function, I see two things to do:
- Implement this new solution (maybe closing this PR without merging and opening a new one as everything commited here should be reverted)
- Think about the integration of cython `n_cells_from_cells` following the indications given by @banesullivan in https://github.com/pyvista/pyvista/discussions/5393. But we need to address pros and cons: the time saved only concern large connectivity arrays when `n_faces` is not provided, alternatives : `.from_regular_faces`, `.from_irregular_faces` and `(..., n_faces=...)` already avoid a call to `n_cells_from_cells`. The maintenance cost of integrating a cython module seems high to me in comparison to the expected performance gain. Maybe adding a Notes on PolyData docstring to precise the more efficient ways to initialize large polydatas would be sufficient ?

pyvista/core/pointset.py
Louis-Pujol: I agree with that

pyvista/core/pointset.py
Louis-Pujol: I call `.flatten` because the checker must be called with 1D array, then it tries to reshape as a 2D array to quickly check if the connectivity array is regular. It sounds like going round in circles but I think that numpy does not copy data and the performance is not impacted. Anyway, it would be safer to call `.view` instead of `.flatten`/`.reshape`.

Louis-Pujol:
I just opened a new PR with the lighter solution to the initial issue. Thanks a lot @darikg, @tkoyama010 for your precious help !

PR #5383, opened 2023-12-22
Remove `add_volume_clip_plane` unused argument

tkoyama010:
### Overview

This is the follow-up PR of #5374.
Remove `add_volume_clip_plane` unused argument.

### Details

- None


banesullivan (commented):
My gut reaction is that we should support these options for this method. Perhaps we should look into using these arguments rather than removing them

PR #5404, opened 2023-12-28
Check the validity of padding information when initializing cells of a CellArray

### LINKED ISSUE #5217: Check that arrays with padding are valid

Louis-Pujol:
Resolves #5217 (follows #5376 where a suboptimal fix was proposed)

When a `CellArray` is inititialized, the vtk method `SetCells` is called. 

https://github.com/pyvista/pyvista/blob/276b50b1cfff04ab7912a3e4e6acf72d28d1916f/pyvista/core/cell.py#L626-L641

At this point, cells is supposed to be a connectivity array with padding information. However if the padding structure is incorrect, `SetCells` can read outside the memory bounded by the array `cells`, causing possible memory leak.

This PR fixes this issue by adding a check of the expected/actual size of `cells` by comparing size of `cells` and `self.cells`.

Tests were added to ensure that
- initialize a `PolyData` with invalid connectivity arrays for `faces`, `lines`, `strips` or `verts`
- intialize a `UnstructuredGrid` with invalid connectivity arrays for `cells`

leads to ValueError.

### DIFF
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -11,6 +11,7 @@ from . import _vtk_core as _vtk
 from ._typing_core import IntMatrix, IntVector, NumpyIntArray
 from .celltype import CellType
 from .dataset import DataObject
+from .errors import CellSizeError
 from .utilities.cells import ncells_from_cells, numpy_to_idarr
 
 
@@ -643,10 +644,12 @@ class CellArray(_vtk.vtkCellArray):
         # read memory outside of the array bounds if the padding is incorrect.
         # See https://github.com/pyvista/pyvista/issues/5217
         if self.cells.size != cells.size:
-            raise ValueError(
-                f"Cells array size ({cells.size}) does not match expected size"
-                f" ({self.cells.size}). This is likely due to an invalid"
-                " connectivity array."
+            raise CellSizeError(
+                message=(
+                    f"Cell array size is invalid. Size ({cells.size}) does not"
+                    f" match expected size ({self.cells.size}). This is likely"
+                    " due to invalid connectivity array."
+                )
             )
 
         self.__offsets = self.__connectivity = None
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -639,10 +639,7 @@ class CellArray(_vtk.vtkCellArray):
 
         self.SetCells(n_cells, vtk_idarr)
 
-        # Check that the size of cells array matches expectation.  Errors here
-        # are likely due to an invalid connectivity array. VTK SetCells can
-        # read memory outside of the array bounds if the padding is incorrect.
-        # See https://github.com/pyvista/pyvista/issues/5217
+        # https://github.com/pyvista/pyvista/pull/5404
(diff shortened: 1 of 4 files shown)


darikg:
👍 Really appreciate all the time & effort that went into #5376. Very cool you found a zero-downside alternative. Trivial suggestion: maybe raise some sort of custom `CellSizeError` (or however you want to spell it) subclassing from `ValueError`? There's a bunch of prior art [here](https://github.com/pyvista/pyvista/blob/main/pyvista/core/errors.py).

Louis-Pujol:
@darikg  I did the modification following your advice. Thanks

pyvista/core/cell.py
```diff
@@ -637,6 +638,20 @@ def _set_cells(self, cells: NumpyIntArray, n_cells: Optional[int], deep: bool) -
                 n_cells = cells.shape[0]
 
         self.SetCells(n_cells, vtk_idarr)
+
+        # Check that the size of cells array matches expectation.  Errors here
+        # are likely due to an invalid connectivity array. VTK SetCells can
+        # read memory outside of the array bounds if the padding is incorrect.
+        # See https://github.com/pyvista/pyvista/issues/5217
```
tkoyama010, line 645: I would encourage the code reader to read the PR description of the change rather than reading the code description. However, this is a personal policy and can be ignored.
```suggestion
        # https://github.com/pyvista/pyvista/pull/5404
```

tkoyama010 (changes requested):

tkoyama010 (approved):

tkoyama010:
Awesome job! It is ready to go after if @darikg approve this.

darikg:
lgtm

PR #5406, opened 2023-12-29
[PATCH] Fix geometry orientation with Y-normal vector

### LINKED ISSUE #5405: Plane (and other geometries) orientation is incorrect as of 0.43

@tkoyama010 teased out that #4853 is causing the issue in https://github.com/bjlittle/geovista/issues/447, and I dug into this a bit and realized that now many of our geometric operations are producing entirely different results across versions.

I'm going to propose a PR that reverts #4853 for now and issue a patch release if that merges... though I'm worried that since a number of changes have happened to the underlying `translate()` method, this may not be too clean. 

As far as I can tell this is only isolated to geometries that have a user-specified normal of `(0, 1, 0)`. 

```python
import pyvista as pv

plane = pv.Plane(
    center=(0.5, 0, 0),
    i_resolution=1,
    j_resolution=1,
    i_size=1.0,
    j_size=2.0,
    direction=(0, 1, 0),
)

plotter = pv.Plotter()
plotter.add_mesh(plane, color='r', show_edges=True)
plotter.view_xz()
plotter.show()
```

### System Information

```shell
main and 0.43
```

### Screenshots

| 0.42 and prior | 0.43 (`main`) |
|---|---|
| plane_orientation-0 42 | plane_orientation-0 43 |

banesullivan:
Fix #5405 

I honestly don't understand what's going on in this method, but this appears to fix the issue that https://github.com/pyvista/pyvista/pull/4853 introduced.

If this is working and correct, I'd like to land asap and issue a patch release. We can handle the implementation of tests to prevent a regression in follow-up work.

@pyvista/developers, I'd really appreciate some extra eyes on this as I feel like my solution here is too simple and perhaps not accounting for other scenarios.

### DIFF
--- a/pyvista/core/utilities/geometric_sources.py
+++ b/pyvista/core/utilities/geometric_sources.py
@@ -39,9 +39,9 @@ def translate(surf, center=(0.0, 0.0, 0.0), direction=(1.0, 0.0, 0.0)):
     # Adjust normy if collinear with normx since cross-product will
     # be zero otherwise
     if np.allclose(normx, [0, 1, 0]):
-        normy_temp = [-1.0, 0.0, 0.0]
+        normy_temp = [0.0, 0.0, -1.0]
     elif np.allclose(normx, [0, -1, 0]):
-        normy_temp = [1.0, 0.0, 0.0]
+        normy_temp = [0.0, 0.0, 1.0]
 
     normz = np.cross(normx, normy_temp)
     normz /= np.linalg.norm(normz)


tkoyama010:
Thank you for your fix for this.

I, too, am having trouble grasping the process in the modified area because it is too implicit. In the meantime, let's fix it so that we can respond to user reports of backward compatibility breakage.

As Zen of Python says, "Exlicit is better than implicit". Therefore, it is my hope to encourage users to specify explicitly via #5188 in the future after releasing this patch.

darikg:
`translate` is kinda strange. it came up before [here](https://github.com/pyvista/pyvista/pull/5188#issuecomment-1810419519). Fundamentally the rotation it's supposed to apply is underspecified, and [discontinuous](https://math.stackexchange.com/a/1863091).

tkoyama010:
I wrote to maintain backward compatibility, but the effort may cost a lot. It may be better to raise warnings that the implicitly defined method is unstable and unsupported.
The `translate` specification should also be clarified.

user27182:
I'm not sure that reverting an old bugfix is the solution... what about simply copy/pasting the current (v0.43) `translate` method as a new custom, nested `_translate` method inside of `pv.Plane`, and modifying it to restore the v0.42 functionality? This will fix the current problem for `Plane` only and leave the fix from #4853 intact.

This would be a temporary measure until the translations / rotations are made more explicit. For the proper fix moving forward, I'm a fan of explicitly specifying two direction vectors, similar to what's proposed in #5188 (but generalized for all geometric objects).

PR #5428, opened 2024-01-02
Add custom trackball style and refactoring `enable_2d_style`

MatthewFlamm:
### Overview

Adds ability to customize trackball style with user defined options for mouse bindings.

### Details

`enable_2d_style` is now a convenience wrapper with some preset values.

### TODO

- [x] Add in guard checks for valid string options
- [x] Should `None` remove all functionality rather then use default?
- [x] Add EnvironmentRotate for `shift_right`
- [x] Add tests


MatthewFlamm:
> Should `None` remove all functionality rather then use default?

Since we inherit from `vtkInteractorStyleTrackballCamera` there are already observers defined in vtk that define the functionality for each button press.   Since we add observers later in python, our observers execute later and thus override the behavior.  This means it is not easy to set the behavior to nothing, e.g. shift-left click could do nothing when the mouse moves.  We could consider clearing all observers for the button press events, but then I'm not sure if there are unintended consequences of this.  I'm leaning towards leaving this as is, i.e. you can only change the default behavior from `vtkInteractorStyleTrackballCamera` to something else.

pyvista/plotting/render_window_interactor.py
```diff
@@ -547,107 +547,196 @@ def enable_trackball_style(self):
...
+    def enable_custom_trackball_style(
+        self,
+        left="rotate",
+        shift_left="pan",
+        control_left="spin",
+        middle="pan",
+        shift_middle="pan",
+        control_middle="pan",
+        right="dolly",
+        shift_right=None,
```
MatthewFlamm, line 559: This is actually 'environment rotate', see https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6566.  All the other options have named defaults, so it is a bit weird to keep this as `None`.  If we do implement this, we can also consider getting rid of the `None` functionality to keep defaults.

MatthewFlamm: I didn't implement this yet, as I don't use this feature.  I need to investigate how it works for testing.

MatthewFlamm:
> We could consider clearing all observers for the button press events, but then I'm not sure if there are unintended consequences of this.

After thinking this through, I think the potential downsides and the complexity of making this change means we should not do this.  In the case that we want a fully customizable style, we would want to inherit from `[vtkInteractorStyleUser](https://vtk.org/doc/nightly/html/classvtkInteractorStyleUser.html).  This class however would require us to implement our own routines for rotating the camera, and so on.

tests/plotting/test_plotting.py
```diff
@@ -3937,11 +3949,12 @@ def setup_plot():
     pl.iren._control_key_release()
     pl.close()
 
-    # shift right click dollys, image 9
+    # shift right click environment rotate, image 9
+    # does nothing here
```
MatthewFlamm, line 3953: I couldn't figure out how to best test this functionality.  In my testing, it only rotates the reflection in an object using `pbr` rendering and using a skybox like in our examples.  I can't get it to also rotate the background like in https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6566.  There is no reproducing code in that MR, so I'm not sure what the difference is.  But I do not think this code is the issue as the reflection does seem to rotate in a manual test.

tests/plotting/test_plotting.py
```diff
@@ -3937,11 +3949,12 @@ def setup_plot():
     pl.iren._control_key_release()
     pl.close()
 
-    # shift right click dollys, image 9
+    # shift right click environment rotate, image 9
```
MatthewFlamm, line 3952: We have a bunch of duplicate images in the image cache.  I'm thinking that it would be useful to add an option to `pytest_pyvista` to specify which image name to compare against for each image comparison.  This way, we could store only the unique images for this test.

MatthewFlamm: Tracked in https://github.com/pyvista/pytest-pyvista/issues/109

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM. This will change the game of mouse control :) Thanks!

PR #5140, opened 2023-10-28
Fix issue with +/-Y face floors

banesullivan:
`add_floor()` has been broken for +/- Y faces. This fixes that and adds regression tests

Previous `add_floor('-y')`:
Screenshot 2023-10-28 at 3 24 29 PM

PR #5435, opened 2024-01-04
[PATCH] Restore geometric and parametric object orientation from `v0.42.1`

### LINKED ISSUE #5405: Plane (and other geometries) orientation is incorrect as of 0.43

user27182:
Fix #5405.

This PR is similar to #5406. It was initially a fix for `Plane` only but is now a more general fix for any geometric or parametric objects which use `translate` under-the-hood.

The fix is implemented by:
- Re-introducing a shift in axes direction vectors to `translate` in a manner similar to `np.roll`, a behaviour which was first introduced in `v0.42` by #4674. 
  - NOTE: This behaviour was later changed by #4853 and patched into `v0.42.2`, so most objects have a different default `Y` direction in thereafter.
  - This PR (i.e. [#5435](https://github.com/pyvista/pyvista/pull/5435)) effectively re-introduces bug https://github.com/pyvista/pyvista/pull/4850 as a result.
- Reverting the changes to `add_floor` which were made by https://github.com/pyvista/pyvista/pull/5140.
- Adding a 180 degree rotation to `Cylinder` to fix the orientation error introduced by https://github.com/pyvista/pyvista/pull/4890 and https://github.com/pyvista/pyvista/pull/5004
- Adding regression tests with images based on the output from `v0.42.1`.

EDIT: The regression tests (without fixes) have been merged separately in #5614.


pyvista/core/utilities/geometric_objects.py
```diff
@@ -972,8 +972,15 @@ def Plane(
 
     surf.points[:, 0] *= i_size
     surf.points[:, 1] *= j_size
+
+    # Rotate so the plane's normal points in the +x direction
     surf.rotate_y(90, inplace=True)
+    # The plane's i and j directions are under-defined, so we rotate again
+    # for special cases to maintain compatibility with older versions, see GH#5405
```
tkoyama010, line 988:
```suggestion
    # https://github.com/pyvista/pyvista/pull/5435
```

tests/core/test_geometric_objects.py
```diff
@@ -333,6 +333,21 @@ def test_plane_direction(expected):
     assert np.array_equal(actual, expected)
 
 
+def test_plane_with_y_direction():
+    # test plane orientation when direction is along y axis is consistent
+    # with older versions of pyvista
```
tkoyama010, line 338:
```suggestion
    # https://github.com/pyvista/pyvista/pull/5435
```

tests/core/test_geometric_objects.py
```diff
@@ -333,6 +333,21 @@ def test_plane_direction(expected):
...
+    surf = pv.Plane(
+        i_resolution=1,
+        j_resolution=1,
+        i_size=1.0,
+        j_size=2.0,
+        direction=(0, 1, 0),
+    )
+    # long axis should be parallel to z-axis
+    # short axis should be parallel to x-axis
```
tkoyama010, line 347:
```suggestion
```

pyvista/core/utilities/geometric_objects.py
```diff
@@ -972,8 +981,15 @@ def Plane(
 
     surf.points[:, 0] *= i_size
     surf.points[:, 1] *= j_size
+
+    # Rotate so the plane's normal points in the +x direction
```
tkoyama010, line 985:
```suggestion
```

pyvista/core/utilities/geometric_objects.py
```diff
@@ -972,8 +981,15 @@ def Plane(
...
     surf.rotate_y(90, inplace=True)
+    # The plane's i and j directions are under-defined, so we rotate again
+    # for special cases to maintain compatibility with older versions, see GH#5405
+    if np.allclose(np.abs(direction), (0, 1, 0)):
+        surf.rotate_x(90, inplace=True)
     translate(surf, center, direction)
+
```
tkoyama010, line 992:
```suggestion
```

pyvista/core/utilities/geometric_objects.py
```diff
@@ -972,7 +981,11 @@ def Plane(
 
     surf.points[:, 0] *= i_size
     surf.points[:, 1] *= j_size
+
```
tkoyama010, line 984:
```suggestion
```

tkoyama010 (approved):
Thank you for the clear explanation in the comment. I have moved the explanation to the docstring to make this clear to users.

tkoyama010:
@bjlittle Could you confirm that this patch will resolve https://github.com/bjlittle/geovista/issues/447? If it resolves the issue, we will release a minor version immediately.

bjlittle:
Awesome effort guys! 🍻

I'll make time to test this evening and let you know asap, thanks 👍

pyvista/core/utilities/geometric_objects.py
```diff
@@ -949,6 +949,15 @@ def Plane(
...
 
+    Notes
+    -----
+    Rotate the original plane surface 90 degrees with the y-axis,
+    so the plane's normal points in the +x direction.
+    The plane's i and j directions are under-defined,
+    so we rotate again when the direction is (0, 1, 0)
+    to maintain compatibility with older versions, see
+    https://github.com/pyvista/pyvista/issues/5405.
+
```
user27182, line 960:
```suggestion
    .. warning::
    
        :func:`~pyvista.Plane` is known to have an incorrect orientation 
        when ``direction`` is exactly parallel to the y-axis, i.e. 
        ``direction=(0, 1, 0)`` or `direction=(0, -1, 0)``. This behavior is 
        maintained for compatibility with older versions, see 
        `pyvista/#5405 <https://github.com/pyvista/pyvista/issues/5405>`_.
    
        To obtain the correct orientation for these special cases, it's necessary
        to rotate the plane an additional 90 degrees along its normal axis after 
        instantiation, e.g.
    
        .. code-block:: python
        
            mesh = pv.Plane(direction=(0, 1, 0))
            mesh.rotate_y(90, inplace=True)
 
```

user27182:
I'll note that I don't think the reason for this bug is that the orientation of `Plane` is under-determined. I gave my reasons here https://github.com/pyvista/pyvista/issues/5405#issuecomment-1878307637 for why I think it's possible to fully determine the expected output for this case, and why I believe that the orientation of plane is currently correct.

As such, this PR isn't fixing a bug with `Plane`, but rather is deliberately re-introducing a bug that was previously fixed (for compatibility reasons).

Long term, I think the true fix is to keep the current implementation for Plane (i.e. revert this PR in the future) and update downstream projects that rely on this behavior accordingly, e.g. implement the suggested changes such as this one https://github.com/bjlittle/geovista/issues/447#issuecomment-1879565162

tkoyama010 (dismissed):

user27182:
pre-commit.ci autofix

bjlittle:
@user27182 @tkoyama010 

Okay, I can confirm that this patch fixes the issue in `geovista`. Amazing. Thanks so much 🍻 

Given the suggestion of @user27182 above, I'll try this change within `geovista` to see how it behaves and report back here ASAP.

I agree with @user27182, it would make more sense for `geovista` to change, rather than regress the behaviour in `pyvista`.

tkoyama010 (approved):

tkoyama010:
Once @banesullivan approves this PR, we will do a patch release. He seems to be busy so please wait a bit.

bjlittle:
@user27182 @tkoyama010 

Okay, great news. You can close this PR!

We've found a really sweet compromise here.

I can confirm that we can fully contain the fix within `geovista`, so there is no need for `pyvista` to regress behaviour by applying this patch. I think merging this patch would compromise `pyvista` and just cause you guys headaches, as already mentioned by @user27182.

So, given that a plane is (unexpectedly) rotated when created with `direction=(0, 1, 0)`, thanks to @banesullivan for spotting this, I've decided to completely avoid this behaviour.

I realised that we can get the same end result in `geovista` by adopting the default `Plane(..., direction=(0, 0, 1))` and then applying a `rotate_x(90, inplace=True)` on the resultant plane instead. 

The idea of performing the additional rotation was suggested by @user27182, so thanks for that 🍻 

I can instead release a patch on `geovista` that includes a minimum pin of `pyvista>=0.43.1` ... I simply need to remove new deprecations warnings being generated by the latest version of `pyvista`, but otherwise I'm now unblocked.

Thanks guys for the awesome support. Totally first class 💯🥳🎖️

bjlittle:
Although, a warning being added to the `Plane` docstring is still of great value to the community and developers, as it documents the behaviour when using `direction=(0, 1, 0)`.

Perhaps this could be the focus of this pull-request now ?

user27182:
I'm glad this could be resolved. I agree a warning should be added and possibly patched for both 0.42 and 0.43. The warning could be added inside of `translate`.

banesullivan:
+1 for adding a warning here! Great work everyone digging into this!

user27182:
Do we want a warning in the docs or a warning in the code (e.g. in translate)? Or both?

Is the plan to patch 0.42, 0.43, or both ? (The behaviour changed in 0.42.2)

tkoyama010:
> Do we want a warning in the docs or a warning in the code (e.g. in translate)? Or both?

Both please.

> Is the plan to patch 0.42, 0.43, or both ? (The behaviour changed in 0.42.2)

Yes. We will.

tkoyama010 (changes requested):
Set Request changes to prevent merging before changing the above.

user27182:
I encountered some (still) unexpected behaviour while adding a few edge cases to codify what I believed to be the current expected behaviour of `Plane`.

So, I'm working on adding tests to better understand the broader impact the patch from 0.42.2 had on the orientation for other geometric objects. I'm leaning towards adding regression tests for all objects which have a `direction` or `normal` parameter. If we include all parametric objects, there are approx. 33 objects with these parameters which may have been affected by the changes to `translate`. So, I think I'll generate 33 orientation test images based on the output from 0.42.1, and then compare the output from 0.43.1.

user27182:
pre-commit.ci autofix

user27182:
Not sure what we want the scope of this PR to be now? What do we want to patch, exactly? 

We could do something like what was originally proposed in #5406 and try to restore the 0.42.1 orientation behaviour for all objects.

At the same time, we could add a general warning to `translate` to inform users that `direction` is unstable and should be avoided, and that objects should instead be rotated manually using `rotate_x`, etc. ? (similar to the fix implemented in geovista).

Thoughts?

banesullivan:
Excellent investigative work here, @user27182! This problem is the worst case scenario in my mind when this started... 

>  try to restore the 0.42.1 orientation behaviour for all objects. .. At the same time, we could add a general warning to translate to inform users that direction is unstable and should be avoided ...

+1 for this from me. I think the safest thing to do here is restore previous behavior and issue warnings.

user27182:
Looks like #4890 and/or #5004 may have also introduced changes to the orientation geometry for `Cylinder`.

Here is the v0.42.1 output:
direction_objects object_function3
and v0.43.1 output:
direction_objects object_function3

If we look at `direction=(1,0,0)`, for example, we see that the geometry has been rotated 180 degrees. Visually the cylinder appears to have the same general shape and orientation, but the point IDs / scalars differ between versions.

user27182:
In [3181fab](https://github.com/pyvista/pyvista/pull/5435/files/4de6bda435f740ba47aaceb372809ed914269730..3181fabb6c5d54b369716ba914e81673bf550526), the new regression images (and minor changes to core tests) for all direction/orientation objects are all consistent with the output from `v0.42.1`.

Since #4674 is what first introduced this kind of axis-shifting behaviour (but without any regression tests), I was also concerned that the outputs from `v0.42.1` might also differ from `v0.41.1`. And so I have also double-checked that the images in In https://github.com/pyvista/pyvista/pull/5435/commits/3181fabb6c5d54b369716ba914e81673bf550526 are consistent with `v0.41.1`. And, much to my relief, they are indeed the same, except for `Sphere` and `Plane`, which had incorrect directions in `v0.41` and prior (this was fixed by #4674 in `v0.42`).

user27182:
I believe the technical part of this patch is now finished: the object orientations are now consistent with `v0.42` and prior.

Note that this patch effectively re-introduces bug #4853 as a result. Do we want to include a 'fix' for this as part of this patch? Perhaps locally in `ParametricEllipsoid` with a rotation after calling `translate`? Though I'm not convinced this is easily fixed since I think the fundamental issue stems from `translate` being discontinuous and under-defined, and the bug is not specific to `ParametricEllipsoid`.

If anyone cares to review the changes, test things out locally, etc. feel free to do so now.

I believe adding warnings to code and/or docs is all that remains otherwise.

user27182:
Also, the regression tests exposed a new bug: `CylinderStructured` seems to completely disappear from the plot with `direction=(0, 0, -1)`
direction_objects object_function4-False

I don't have time to investigate this any further at this point and will leave it for now. Otherwise, all the other images / objects look fine though.

user27182:
@banesullivan @tkoyama010 Is there still a desire to get this patch merged? In terms of the technical changes and fixes listed in the top comment?

tkoyama010:
Yes. Of course. Let's set this as "ready for review" and start reviewing.

user27182:
Ok thanks. The technical changes are ready but the documentation and any warnings in code have not yet been updated. Please make suggestions for this as I'm not sure what kind of warnings or information we want to provide with respect to the instability of `translate` or how to avoid it.

user27182:
"pre-commit.ci autofix"

user27182:
pre-commit.ci autofix

tkoyama010:
github-actions deploy

tkoyama010:
github-actions preview

user27182:
Closing since the initial motivation for this PR was to resolve downstream issues which were resolved early on, see https://github.com/pyvista/pyvista/pull/5435#issuecomment-1879907332.

Also, the tests from this PR have already been merged in #5614.

PR #5445, opened 2024-01-09
Add a `generate_images` function to sphinx-gallery module

### LINKED ISSUE #5433: Make the `sphinx-gallery` scraper more flexible

Hi !

A few weeks ago I've done some [work](https://github.com/smarie/mkdocs-gallery/pull/91) on integrating pyvista into [mkdocs-gallery](https://github.com/smarie/mkdocs-gallery). [Here](https://smarie.github.io/mkdocs-gallery/generated/gallery/plot_11_pyvista/) is the example.

The work did there was to copy/paste the scraper from `pyvista.plotting.utilities.Scraper` and adapt it in two ways:
- modify the structure of the inputs (`mkdocs` has its own versions of `block_vars` and `gallery_conf` sphinx's variables)
- at the end, instead of calling `figure_rst`, call an equivalent that convert the list of image names + the source path to a markdown text

The code is mainly duplicated, which is bad, I was wondering if the code inside the class `Scraper` could be reorganized to add more flexibility to let the user the freedom to reuse the generation of pyvista images outside sphinx. The `__call__()` method has 3 steps:

https://github.com/pyvista/pyvista/blob/84ac2a3990a5e429b30383e988f5065c907c071f/pyvista/plotting/utilities/sphinx_gallery.py#L94-L119

- 1/ Iterate over `pyvista.plotting.plotter._ALL_PLOTTERS` to identify active plotters
- 2/ save static/animated/dynamic scenes into files and save name fils in `image_names`
- 3/ call `html_rst` which generate the viewer in reST

Perhaps moving steps 1 and 2 into another function (something like `generate_images_names(image_path_iterator, dynamic= True or False)`) will increase flexibility. Then
- the `sphinx-gallery` scraper will use `generate_image_names` + `figure_rst()`
- user can combine `generate_image_names` with its own generator of `html`/`markdown`/... to integrate pyvista views wherever he wants

If you think it is a good idea, i can start a PR.

Louis-Pujol:
This PR resolve #5433 

A `generate-images` function is added to the module `plotting.utilities.sphinx-gallery`. This function takes as input an iterator that automatically generates images filename sequentially and a boolean argument ``dynamic`` to determined if images must be saved as static ``.png`` images or dynamic ``vtksz`` scenes.

`Scraper` and `DynamicScraper` has been rewritten to use `generate-images`  internally, while the behavior of these objects remains the same.

The goal of this modification is to add flexibility to gallery generation letting the user generates images/scenes and use them with other tools than `sphinx-gallery`.


Louis-Pujol:
I added a test for `DynamicScraper`, previously this class was not tested and was excluded for coverage.

I think the tests are passing well; the recent failure on macOS seems to be related to GHA. But perhaps there was another reason for not testing this class before?

pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -74,6 +76,56 @@ def _process_events_before_scraping(plotter):
...
+        if hasattr(plotter, "_gif_filename"):
+            # move gif to fname
+            fname = fname[:-3] + "gif"
+            shutil.move(plotter._gif_filename, fname)
+            image_names.append(fname)
+        else:
+            plotter.screenshot(fname)
+            if not dynamic or plotter.last_vtksz is None:
+                image_names.append(fname)
+            else:
```
tkoyama010, line 119:
```suggestion
            else:  # pragma: no cover
```

tkoyama010:
> I added a test for `DynamicScraper`, previously this class was not tested and was excluded for coverage.
> 
> Tests are passing well. But perhaps there was another reason for not testing this class before?

I don't think there was any particular reason. Thank you for adding the test.

pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -74,6 +76,56 @@ def _process_events_before_scraping(plotter):
...
+    If a plotter has a ``_gif_filename`` attribute, a gif will be created.
+    Else, depending on the value of ``dynamic``, either a ``.png`` static image
+    or a ``.vtksz`` file will be created.
+
+    Parameters
+    ----------
+    image_path_iterator : Iterator[str]
+        An iterator that yields the path to the next image to be saved.
+    dynamic : bool, default: False
+        Whether to save a static ``.png`` image or a ``.vtksz`` (interactive)
```
tkoyama010, line 93:
```suggestion
    image_path_iterator : Iterator[str]
        An iterator that yields the path to the next image to be saved.

    dynamic : bool, default: False
        Whether to save a static ``.png`` image or a ``.vtksz`` (interactive)
```

pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -74,6 +76,56 @@ def _process_events_before_scraping(plotter):
...
 
 
+def generate_images(image_path_iterator: Iterator[str], dynamic: bool = False) -> List[str]:
+    """Generate images from the current plotters.
+
+    The file names are taken from the ``image_path_iterator`` iterator.
+
+    If a plotter has a ``_gif_filename`` attribute, a gif will be created.
+    Else, depending on the value of ``dynamic``, either a ``.png`` static image
+    or a ``.vtksz`` file will be created.
```
tkoyama010, line 86:
```suggestion
    A gif will be created if a plotter has a ``_gif_filename`` attribute.
    Otherwise, depending on the value of ``dynamic``, either a ``.png`` static image
    or a ``.vtksz`` file will be created.
```

tkoyama010 (dismissed):
LGTM

pyvista/plotting/utilities/sphinx_gallery.py
```diff
@@ -167,21 +205,15 @@ def __call__(self, block, block_vars, gallery_conf):  # pragma: no cover
...
-                        image_names.append(fname)
-        pyvista.close_all()  # close and clear all plotters
+
+        if force_static is None:
+            # Just in case force_static is None at this point
+            force_static = False
+
+        dynamic = False if force_static else True
+
+        image_names = []
```
MatthewFlamm, line 215: Haven't done a full review, but this line looks unneeded.

tkoyama010 (approved):

MatthewFlamm (approved):
This looks good to me, even if only for the reduced duplication of logic in the different Scrapers.

banesullivan (approved):
LGTM! All seems to be working the same when I build locally

PR #5450, opened 2024-01-10
Add development CI for NumPy 2.0 beta testing

MatthewFlamm:
### Overview

This PR and branch serve as the development branch for numpy 2.0 testing~fixes.  Any fixes for numpy 2.0 should be use branch `maint/add-tests-for-numpy2.0` as the base in a PR.~

### Details

https://numpy.org/devdocs/dev/depending_on_numpy.html#testing-against-the-numpy-main-branch-or-pre-releases


MatthewFlamm:
I'm not sure that we will need to test 2.0 against all versions of vtk+python in our CI in an ongoing manner, but I think it would be good to test all of them up front.

MatthewFlamm:
I set this up so that the jobs still pass so it isn't confusing for other PRs.  This means that to tell if changes work against numpy v 2.0, you have to click "Details" and look for failures in the logs themselves.

MatthewFlamm:
Currently tests for pyvista/core have 48 failures and many look like simple variable name substitutions even for vtk == 9.1.   Tests for pyvista/plotting  (and `pv.Report()` for that matter) don't run because at least one plotting depedency, `cmocean` is not compatible.

banesullivan:
> because at least one plotting depedency, cmocean is not compatible.

If its just `cmocean`, I'll advocate just dropping that dependency here

banesullivan (commented):
Generally this looks well implemented though I'm concerned that this will significantly increase our GitHub actions runner usage and we should look into our current usage vs the quota before merging.

I'm  wondering if instead of merging this soon here, if we should have a `numpy-2.0` development branch until we are ready to support numpy2? Then we'd land changes, fixes, etc for numpy2 compatibility as separable PRs against this branch until we are ready to support numpy2 at which point we could merge this branch into `main`.

banesullivan:
^ is just a suggestion/thought. @MatthewFlamm, happy to go with what you think will provide the least friction since this sort of upgrade is likely to be wrought with friction already. Let me know and I'd happy to approve this PR as code-owner so as not to block the merge

banesullivan:
> we should look into our current usage vs the quota before merging.

I took a quick look and think well under our quota but let's not leave these on forever

MatthewFlamm:
> If its just `cmocean`, I'll advocate just dropping that dependency here

Let's give packages some time.  We just started preparing for this  and I put in a PR in `cmocean`.  I'm more worried about `meshio`.

MatthewFlamm:
> I'm wondering if instead of merging this soon here, if we should have a numpy-2.0 development branch until we are ready to support numpy2? 

I'm subconsciously already treating this branch as the development branch. #5453 points here. So let's go with that idea. I will update the description here and change the CI here to show the fails.  Any 2.0 fix PRs should use this branch as the base, and additionally we will keep this branch up to date with main.  Before we merge this PR with main we will prune the CI to be more manageable, but I think testing on all versions would be useful to do while working out the initial bugs.

MatthewFlamm:
Locally testing #5453 with fixed versions of cmocean and meshio, I don't get any errors in our tests.  So, I think we can merge #5453 and then wait for cmocean and meshio.

MatthewFlamm:
#5460 shows that downstream packages also need us to fix this. So, it might be good to prune the numpy 2.0 tests to something reasonable and merge to main.  We will have to ignore failing numpy 2.0 tests in every PR though.

banesullivan:
The most practical way to go here might be to keep this branch for testing and validation and merge small changes like #5460 to `main` as they come

MatthewFlamm:
I added in installing `cmocean` and `meshio` from PR branches in the CI that fix those packages.  We are now having more errors.  I must have messed up the install order previously on my local test that resulted in passing tests.

MatthewFlamm:
We are now getting errors that seem to be related to https://github.com/numpy/numpy/pull/25553

MatthewFlamm:
#5493 and #5496 should fix all errors except one in our tests.  An example error message below:

```txt
    def to_opacity_tf(self):
        """Return the opacity transfer function of this table.
    
        Returns
        -------
        vtk.vtkPiecewiseFunction
            Piecewise function of the opacity of this color table.
    
        Examples
        --------
        >>> import pyvista as pv
        >>> lut = pv.LookupTable()
        >>> tf = lut.to_opacity_tf()
        >>> tf
        <vtkmodules.vtkCommonDataModel.vtkPiecewiseFunction(...) at ...>
    
        """
        opacity_tf = _vtk.vtkPiecewiseFunction()
        for ii, value in enumerate(self.values[:, 3]):
>           opacity_tf.AddPoint(ii, value / self.n_values)
E           OverflowError: Python integer 256 out of bounds for uint8
```

In the cases I've checked, `self.n_values` is 256 of type python `int` and value is a type `np.uint8` (Ranging from 0-255).  It seems in numpy 2.0, it first tries to coerce `256` into a `uint8` before dividing?  This could be a bug in numpy, but it also smells like there is a bug here.  There are two possible fixes I can think of if this isn't changed in numpy. One is to cast `value` to a float type.  The other is to change the denominator to `(self.n_values - 1)` since the largest value I've seen in the tests is 256.  I'm not sure if `self.n_values` can ever be larger than 256 however.  I can't tell what the logic is in this part of the code.

MatthewFlamm:
From discussion in the numpy issue, it seems likely that this will be a bug that is fixed in numpy (hopefully before the release!).  I cannot figure out what the correct logic is here for PyVista as the current code smells like a bug, and in my investigations found that this part of the code, and probably in conjunction with others, is likely causing issues.  This is tracked in #5505.

If the numpy issue isn't resolved timely and we cannot resolve the internal code issue, we can wrap the `self.n_values` with `np.float64(self.n_values)` to remove the error.

.github/workflows/testing-and-deployment.yml
```diff
@@ -190,7 +207,15 @@ jobs:
         if: ${{ matrix.vtk-version == '9.1' || matrix.vtk-version == '9.0.3' }}
         run: pip install 'matplotlib<3.6'
 
+      - name: Set up numpy 2.0
+        if: ${{ matrix.numpy-version != 'default' }}
+        run: |
+          pip uninstall numpy matplotlib -y
+          pip install --pre --no-deps -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple matplotlib
+          pip install --pre --no-deps -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy
```
MatthewFlamm, line 215:
```suggestion
          pip install --pre --no-deps numpy==2.0.0b1
```

MatthewFlamm:
@user27182 pinging you so you are aware of the errors in the tests for the validation functionality with numpy 2.  The tracebacks are extremely long, but this is an example:

```
# Test raises correct error with invalid input
        with pytest.raises(case.error_type, match=case.error_msg):
>           validate_array(invalid_array, **common_kwargs)
E           AssertionError: Regex pattern did not match.
E            Regex: 'must have finite values'
E            Input: "Input cannot be cast as <class 'numpy.ndarray'>."
```

user27182:
Thanks for the heads up. What do you recommend for fixing these? E.g. should I install numpy 2.0 locally with the main pyvista branch, and make a separate PR to merge with main with the fixes?

MatthewFlamm:
Thanks for the heads up. What do you recommend for fixing these? E.g. should I install numpy 2.0 locally with the main pyvista branch, and make a separate PR to merge with main with the fixes?

Good question, maybe now that numpy 2.0 has a beta release, it is time to put the testing into our regular CI?  @tkoyama010 thoughts?

Then yes, you would install numpy 2 locally make fixes and have a separate PR.

user27182:
> Good question, maybe now that numpy 2.0 has a beta release, it is time to put the testing into our regular CI? 

In case it's still too soon to add the CI to the main branch, maybe I can just make a new branch from this one ([maint/add-tests-for-numpy2.0](https://github.com/pyvista/pyvista/tree/maint/add-tests-for-numpy2.0)), add the fixes, make a PR, and then @MatthewFlamm you can merge that branch into this one directly ? And then the new PR I made can be closed without (yet) merging into main.

user27182:
Also, #5571 is still pending review, not sure if we should wait for that to merge first? This would be to avoid possible duplicate work of fixing the errors with this branch (i.e. for the old version of validation packing), only to have the fixes regress with the changes from #5571?

MatthewFlamm:
Reworked this PR to simplify things and allow it to be merged to main.  However, it is currently using `continue-on-error: true` for numpy 2.0 tests.  This will make it show up as successful, so it would be required to always check the logs when dealing with a PR that addresses this.

tkoyama010 (dismissed):
> > Thanks for the heads up. What do you recommend for fixing these? E.g. should I install numpy 2.0 locally with the main pyvista branch, and make a separate PR to merge with main with the fixes?
> 
> Good question, maybe now that numpy 2.0 has a beta release, it is time to put the testing into our regular CI? @tkoyama010 thoughts?

I'm sorry for not getting back to you sooner. I am +1 about putting the testing into our regular CI.
It's a great idea to be able to check the logs without disturbing other CI.

.github/workflows/testing-and-deployment.yml
```diff
@@ -190,7 +198,15 @@ jobs:
         if: ${{ matrix.vtk-version == '9.1' || matrix.vtk-version == '9.0.3' }}
         run: pip install 'matplotlib<3.6'
 
+      - name: Set up numpy 2.0
+        if: ${{ matrix.numpy-version != 'default' }}
```
banesullivan, line 202: This may be nit-picky, feel free to ignore:

```suggestion
      - name: Install latest numpy 2.0 beta/rc
        if: ${{ matrix.numpy-version == 'latest' }}
```

banesullivan: Similarly for the other `continue-on-error` conditions, I'd make the choice of explicitly using `${{ matrix.numpy-version == 'latest' }}`

banesullivan: Mostly just thinking of a scenario where we specify the version of numpy being installed down the road

.github/workflows/testing-and-deployment.yml
```diff
@@ -190,7 +198,15 @@ jobs:
...
+        run: |
+          pip uninstall numpy matplotlib -y
+          pip install --pre --no-deps -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple matplotlib
+          pip install --pre --no-deps numpy==2.0.0b1
```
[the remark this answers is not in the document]
MatthewFlamm: You are right, we didn't need the `--pre` when directly pinning.  I removed the pin however so that we get the latest rc candidate.  We now need the `--pre`.

Using `>` doesn't work with pre releases.  Since pip does not require a hierarchy here, it doesn't know whether `2.0.0b1` is higher or lower than `2.0.0rc1`.  I'd argue it should be able to compare  within a single kind of pre release, but this doesn't seem possible.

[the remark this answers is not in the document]
MatthewFlamm: This machinery needs to be updated when 2.0 is fully released in any case here.  Given the current failures with our CI with no clear way to fix them, we will have problems at that point.

.github/workflows/testing-and-deployment.yml
```diff
@@ -199,6 +215,7 @@ jobs:
           version: 3.0
 
       - name: Plotting Testing (uses GL)
+        continue-on-error: ${{ matrix.numpy-version != 'latest' }}
```
MatthewFlamm, line 218:
```suggestion
        continue-on-error: ${{ matrix.numpy-version == 'latest' }}
```

2.0.0b1
banesullivan, line 1: @MatthewFlamm, what do these files do?

.github/workflows/testing-and-deployment.yml
```diff
@@ -136,14 +136,22 @@ jobs:
...
+            numpy-version: "default"
           - python-version: "3.11"
             vtk-version: "9.2.6"
+            numpy-version: "default"
           - python-version: "3.12"
             vtk-version: "latest"
+            numpy-version: "default"
+          - python-version: "3.12"
+            vtk-version: "latest"
+            numpy-version: "latest"
```
banesullivan, line 154: Hrm... I had to iterate on this one more time but I'm a little concerned about the difference in meaning of "latest" for `vtk-version` and `numpy-version`. "latest" for VTK means the latest stable release where as "latest" for numpy here means pre-release. Should we change the terms here so its consistent for numpy?

banesullivan: Sorry for my onslaught of nitpicky feedback here! 😄

MatthewFlamm: I'm open to feedback, but we will have to change this once numpy fully releases in any cases here.  So whatever we choose will be short lived.  At that point and assuming we can fix the errors, we should generally test with latest numpy going forward, and targeted older versions.

But I take the point that the way GitHub displays the making makes it hard to distinguish which test is what.  Let's use 'pre' for numpy, or something like this, for now.

MatthewFlamm, line 154:
```suggestion
            numpy-version: "pre"
```

.github/workflows/testing-and-deployment.yml
```diff
@@ -190,7 +198,15 @@ jobs:
         if: ${{ matrix.vtk-version == '9.1' || matrix.vtk-version == '9.0.3' }}
         run: pip install 'matplotlib<3.6'
 
+      - name: Install latest numpy 2.0 beta/rc
+        if: ${{ matrix.numpy-version == 'latest' }}
```
MatthewFlamm, line 202:
```suggestion
        if: ${{ matrix.numpy-version == 'pre' }}
```

.github/workflows/testing-and-deployment.yml
```diff
@@ -190,7 +198,15 @@ jobs:
...
+        run: |
+          pip uninstall numpy matplotlib -y
+          pip install --pre --no-deps -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple matplotlib
+          pip install --pre --no-deps numpy
+
       - name: Core Testing (no GL)
+        continue-on-error: ${{ matrix.numpy-version == 'latest' }}
```
MatthewFlamm, line 209:
```suggestion
        continue-on-error: ${{ matrix.numpy-version == 'pre' }}
```

.github/workflows/testing-and-deployment.yml
```diff
@@ -199,6 +215,7 @@ jobs:
           version: 3.0
 
       - name: Plotting Testing (uses GL)
+        continue-on-error: ${{ matrix.numpy-version == 'latest' }}
```
MatthewFlamm, line 218:
```suggestion
        continue-on-error: ${{ matrix.numpy-version == 'pre' }}
```

tkoyama010 (approved):

banesullivan (approved):

PR #5453, opened 2024-01-11
Fixes for numpy 2.0 in core

MatthewFlamm:
### Overview

Fixes for numpy 2.0 in core.

### Details
~Note the base branch is not main, it is the one from #5450.~

The remaining errors in the tests for core should all be related to meshio.  The remaining errors in the tests for plotting should all be related to cmocean.


tests/core/test_geometric_objects.py
```diff
@@ -490,10 +490,10 @@ def test_text_3d():
         bnds[3] - bnds[2],
         bnds[5] - bnds[4],
     )
-    assert actual_width == 2
-    assert actual_height == 3
-    assert actual_depth == 0.5
-    assert mesh.center == [1.0, 2.0, 3.0]
+    assert np.isclose(actual_width, 2)
+    assert np.isclose(actual_height, 3)
```
tkoyama010, line 494: Let's make explicit that it is float.
```suggestion
    assert np.isclose(actual_width, 2.0)
    assert np.isclose(actual_height, 3.0)
```

MatthewFlamm:
This PR shows that numpy 2.0 changes break vtk 9.0.3 because it requires too old of a numpy version.  We either need to put in `try, except` blocks for `numpy.exceptions` or drop that vtk version.

MatthewFlamm:
I have changed this PR to use `main` as the base branch, so it can be merged now.

pyvista/core/utilities/arrays.py
```diff
@@ -866,6 +866,15 @@ def cast_to_ndarray(arr, /, *, as_any=True, dtype=None, copy=False):
     """
     if as_any and not copy and dtype is None and isinstance(arr, np.ndarray):
         return arr
+
+    # needed to support numpy <1.25
```
tkoyama010, line 870:
```suggestion
    # needed to support numpy <1.24 and vtk 9.0.3
```
See:
https://github.com/pyvista/pyvista/blob/e84354e72acf6d69277954cdc0f16c739aab6897/.github/workflows/testing-and-deployment.yml#L186-L188

tests/core/test_input_validation.py
```diff
@@ -767,13 +767,21 @@ def _check_is_sorted_params(arr):
     num_elements = np.prod(shape)
     arr_strict_ascending = np.arange(num_elements).reshape(shape)
 
+    # needed to support numpy <1.25
```
tkoyama010, line 770:
```suggestion
    # needed to support numpy <1.24 and vtk 9.0.3
```
See:
https://github.com/pyvista/pyvista/blob/e84354e72acf6d69277954cdc0f16c739aab6897/.github/workflows/testing-and-deployment.yml#L186-L188

pyvista/core/utilities/arrays.py
```diff
@@ -866,6 +866,15 @@ def cast_to_ndarray(arr, /, *, as_any=True, dtype=None, copy=False):
...
+    # needed to support vtk 9.0.3
```
tkoyama010, line 871:
```suggestion
```

tests/core/test_input_validation.py
```diff
@@ -767,13 +767,21 @@ def _check_is_sorted_params(arr):
...
+    # needed to support vtk 9.0.3
```
tkoyama010, line 771:
```suggestion
```

tkoyama010 (changes requested):

pyvista/core/utilities/arrays.py
MatthewFlamm: This is true for this vtk version.  If someone tries to use numpy 1.24 however with a newer version, it also will fail.  It just happens we only test <1.24

tkoyama010: OK. Thanks!

tkoyama010 (approved):
LGTM

MatthewFlamm:
For reference and future breadcrumbs, the `numpy.exceptions` module was added in numpy 1.25, see https://github.com/numpy/numpy/blob/d35cd07ea997f033b2d89d349734c61f5de54b0d/numpy/exceptions.py#L8

banesullivan (approved):
🚀

PR #5226, opened 2023-11-25
Update `validation` package

user27182:
### Overview
Update `validation` package. This is a follow-up PR from #5100.

- Rename package: #5558
- Simplify function names, e.g. `check_has_shape` -> `check_shape`
- Add array wrappers to process flat sequences and nested 2D sequences separately from all other array-like inputs. This allows sequence inputs to be processed _without_ needing to cast to a numpy array first.
- Change and update definitions for checking `scalar` and `number` types
- Add `check_numeric` function
- Update package typing:
   - use numpy types `ArrayLike`, `DTypeLike`, etc. 
   - use updated generic PyVista types, e.g.`Vector[float]` from #5171 and #5285.
- Rework `check_string_in_iterable` as `check_contains` for general use with any input type and any container.
- Remove functions `check_is_dytpe`, `validate_dytpe` and `check_is_arraylike` since their use cases are very limited and type errors are already handled by NumPy.
- Remove `check_iterable_of_strings`;  the more general `check_iterable_items` function should be used instead.

Package function names may also be renamed accordingly to better match the new pyvista type names (as applicable).

Enhacements for follow-up PRs:
- Add `optional` param to checkers to do nothing if input is `None`
- Add `check_ndim` to check array dimensionality
- Remove use of `inspect` module to improve performance

PR #5454, opened 2024-01-11
Add list data type support to wrap dataset helpers

tkoyama010:
### Overview

Add list data type support to wrap dataset helpers.

### Details

- None


tests/plotting/test_plotter.py
```diff
@@ -265,15 +265,15 @@ def test_deep_clean(cube):
...
+def test_disable_depth_of_field():
     pl = pv.Plotter()
     pl.enable_depth_of_field()
     assert pl.renderer.GetPass() is not None
     pl.disable_depth_of_field()
     assert pl.renderer.GetPass() is None
 
 
-def test_remove_blurring(sphere):
+def test_remove_blurring():
```
tkoyama010, line 276:
```suggestion
def test_remove_blurring(sphere):
```

tests/plotting/test_plotter.py
```diff
@@ -305,7 +305,7 @@ def test_clear_actors(cube, sphere):
     assert len(pl.renderer.actors) == 0
 
 
-def test_anti_aliasing_multiplot(sphere):
+def test_anti_aliasing_multiplot():
```
tkoyama010, line 308:
```suggestion
def test_anti_aliasing_multiplot(sphere):
```

tests/plotting/test_plotting.py
```diff
@@ -163,7 +163,7 @@ def test_import_vrml():
     pl.show()
 
 
-def test_export_vrml(tmpdir, sphere, airplane, hexbeam):
+def test_export_vrml(tmpdir, sphere):
```
tkoyama010, line 166:
```suggestion
def test_export_vrml(tmpdir, sphere, airplane, hexbeam):
```

tests/plotting/test_plotting.py
```diff
@@ -1607,7 +1618,7 @@ def test_link_views(sphere):
 
 
 @skip_windows
-def test_link_views_camera_set(sphere, verify_image_cache):
+def test_link_views_camera_set(verify_image_cache):
```
tkoyama010, line 1621:
```suggestion
def test_link_views_camera_set(sphere, verify_image_cache):
```

tests/plotting/test_plotting.py
```diff
@@ -3564,7 +3575,7 @@ def test_plot_volume_rgba(uniform):
     pl.show()
 
 
-def test_plot_window_size_context(sphere):
+def test_plot_window_size_context():
```
tkoyama010, line 3578:
```suggestion
def test_plot_window_size_context(sphere):
```

tests/plotting/test_scalar_bars.py
```diff
@@ -63,7 +63,7 @@ def test_clear(scalar_bars):
     assert len(scalar_bars) == 0
 
 
-def test_too_many_scalar_bars(sphere):
+def test_too_many_scalar_bars():
```
tkoyama010, line 66:
```suggestion
def test_too_many_scalar_bars(sphere):
```

tests/plotting/test_theme.py
```diff
@@ -56,7 +56,7 @@ def test_depth_silhouette_eq(default_theme):
     assert my_theme.silhouette != 1
 
 
-def test_depth_silhouette_opacity_outside_clamp(default_theme):
+def test_depth_silhouette_opacity_outside_clamp():
```
tkoyama010, line 59:
```suggestion
def test_depth_silhouette_opacity_outside_clamp(default_theme):
```

tests/plotting/test_widgets.py
```diff
@@ -291,7 +291,7 @@ def __call__(self, *args, **kwargs):
     p.close()
 
 
-def test_widget_sphere(uniform):
+def test_widget_sphere():
```
tkoyama010, line 294:
```suggestion
def test_widget_sphere(uniform):
```

tests/plotting/test_widgets.py
```diff
@@ -322,7 +322,7 @@ def test_widget_closed(uniform):
 
 
 @pytest.mark.needs_vtk_version(9, 1)
-def test_add_camera_orientation_widget(uniform):
+def test_add_camera_orientation_widget():
```
tkoyama010, line 325:
```suggestion
def test_add_camera_orientation_widget(uniform):
```

pyvista/core/utilities/helpers.py
```diff
@@ -16,7 +16,7 @@
 
 
 def wrap(
-    dataset: Optional[Union[np.ndarray, _vtk.vtkDataSet, 'Trimesh', 'Mesh']]
+    dataset: Optional[Union[np.ndarray, _vtk.vtkDataSet, 'Trimesh', 'Mesh', list]]
```
MatthewFlamm, line 19: I think we can use `Matrix` from our `_typing_core`?

pyvista/core/utilities/helpers.py
```diff
@@ -34,7 +34,7 @@ def wrap(
 
     Parameters
     ----------
-    dataset : :class:`numpy.ndarray` | :class:`trimesh.Trimesh` | vtk.DataSet
+    dataset : :class:`numpy.ndarray` | :class:`trimesh.Trimesh` | vtk.DataSet | list
```
MatthewFlamm, line 37: Same as above.

pyvista/core/utilities/helpers.py
```diff
@@ -110,6 +110,9 @@ def wrap(
         # Return object if it is already wrapped
         return dataset  # type: ignore
 
+    if isinstance(dataset, list):
```
MatthewFlamm, line 113: This should probably be any `Sequence`?

pyvista/core/utilities/helpers.py
```diff
@@ -110,6 +110,9 @@ def wrap(
...
+        dataset = np.array(dataset)
```
MatthewFlamm, line 114: We could have a list of `DataSets` that we should maybe coerce into a `MultiBlock`.  Also if the `isinstance` check is changed to `Sequence` a `MultiBlock` could be here.  We need to handle all these edge cases here.

pyvista/core/utilities/helpers.py
MatthewFlamm: If we allow Sequences of DataSets to be wrapped to MultiBlock then we'd also need to add that as I don't think `Matrix` includes this.

tkoyama010: MultiBlock support sounds appealing, but I would like to carefully consider the specifications and would prefer to do so in another PR.
```suggestion
    dataset: Optional[Union[np.ndarray, _vtk.vtkDataSet, 'Trimesh', 'Mesh', Matrix]]
```

pyvista/core/utilities/helpers.py
tkoyama010:
```suggestion
    dataset : :class:`numpy.ndarray` | :class:`trimesh.Trimesh` | vtk.DataSet | Matrix
```

pyvista/core/utilities/helpers.py
tkoyama010:
```suggestion
    if isinstance(dataset, ollections.abc.Sequence):
```

pyvista/core/utilities/helpers.py
tkoyama010: Good point. I am considering whether we can use `validate_array`.
https://github.com/pyvista/pyvista/blob/1d27d8f331c75929a89e444770c21d81f031a25b/pyvista/core/input_validation/validate.py#L37

tkoyama010: It seems to work fine! But maybe we need to wait for #5226.

pyvista/core/utilities/helpers.py
```diff
@@ -110,6 +113,13 @@ def wrap(
         # Return object if it is already wrapped
         return dataset  # type: ignore
 
+    if isinstance(dataset, collections.abc.Sequence):
+        dataset = validate_array(
+            dataset,
+            must_have_shape=(-1, 3),
+            must_have_dtype=[np.int32, np.int64, np.float32, np.float64],
```
tkoyama010, line 120:
```suggestion
            must_have_dtype=np.number,
```

pyvista/core/utilities/helpers.py
```diff
@@ -10,13 +12,14 @@
...
 from pyvista.core import _vtk_core as _vtk
+from pyvista.core.input_validation.validate import validate_array
 
 from . import transformations
 from .fileio import from_meshio, is_meshio_mesh
 
 
 def wrap(
-    dataset: Optional[Union[np.ndarray, _vtk.vtkDataSet, 'Trimesh', 'Mesh']]
+    dataset: Optional[Union[np.ndarray, _vtk.vtkDataSet, 'Trimesh', 'Mesh', Matrix[float]]]
```
MatthewFlamm, line 22: Could this technically be of type `Matrix[int]` too? I'm not sure.

pyvista/core/utilities/helpers.py
```diff
@@ -110,6 +113,13 @@ def wrap(
         # Return object if it is already wrapped
         return dataset  # type: ignore
 
+    if isinstance(dataset, collections.abc.Sequence):
+        dataset = validate_array(
+            dataset,
+            must_have_shape=(-1, 3),
```
MatthewFlamm, line 119: The logic below this allows for arrays that are not (N, 3) to be wrapped.

pyvista/core/utilities/helpers.py
MatthewFlamm: The scenario in which this creates a problem is when the `pyvista._wrappers` has been overwritten for `vtkMultiBlockDataSet`.  Then when any piece of internal or external code calls `pv.wrap(multi_block_mesh)`, where `multi_block_mesh` is type `pv.MultiBlock`, it will fail on this line.  We need to handle this edge case here in my opinion, but we don't necessarily need to add in support for something like `[mesh_1, mesh_2, mesh_3]` to be wrapped to a `pv.MultiBlock` at this time.

MatthewFlamm: The reason is because `pv.MultiBlock` inherits from `collections.abc.Sequence`.

pyvista/core/utilities/helpers.py
user27182: `int` is allowed by type checkers anywhere `float` is used. So I think `int` is implied here, and doing something like `Union[Matrix[float], Matrix[int]]` is not necessary. (see [PEP484](https://peps.python.org/pep-0484/#the-numeric-tower)). 

As an aside, the mypy plugin from #5285 could mess with this a bit, since that plugin maps `float <-> np.floating` and `int <-> np.integer`... so `np.integer` may also be allowed, though I'm not sure.

But, this `float` and `int` interchangeability for type hints has its quirks, since type checkers won't complain if you do `x: Sequence[float] = [1.0, 2]` since both float and int are allowed.

If you really only want floats and no ints, then  it's best to use a `TypeVar` instead, since they can be bound to `float` explicitly which will cause a type checker to complain, e.g:

``` python
FloatType = TypeVar('FloatType', bound=float)
`x: Sequence[FloatType] = [1.0, 2]`  # not allowed
```

A general `NumberType` TypeVar is already part of `_typing_core`, so perhaps you can just use that, i.e. `Matrix[NumberType]`. This is what's used in #5571 for annotating all array-like inputs.

tkoyama010:
pre-commit.ci autofix

PR #5460, opened 2024-01-12
MAINT: Replace np.float_ with np.float64

larsoner:
See https://numpy.org/devdocs/release/2.0.0-notes.html#numpy-2-0-python-api-removals and [related failure in MNE-Python NumPy 2.0 run](https://github.com/mne-tools/mne-python/actions/runs/7500390514/job/20419001265?pr=12354#step:15:213)


MatthewFlamm:
I changed the base branch to point to the pyvista branch with numpy 2.0 testing to make sure it all passes here too.  The recent #5285 added this additional incompatible line, which means we should try to get the testing moved over to include numpy 2.0 as soon as is possible to avoid further regressions.

larsoner:
If you want to update CIs feel free to take inspiration from here where we use prerelease NumPy 2.0 plus prerelease VTK wheels

https://github.com/mne-tools/mne-python/blob/eefd179b767fea456fd1eab3700c5299310b5b8e/tools/github_actions_dependencies.sh#L31-L44

MatthewFlamm:
You may want to weigh in on #5450 on the preferred method for downstream packages on how we test for and merge in changes here for numpy 2.0.  Currently, we were planning to maintain a development branch for testing and changes, but maybe this isn't ideal for downstream?

banesullivan:
I think this is small and targeted enough to immediately land on `main` rather than our numpy-2.0 dev branch

larsoner:
Up to you where to merge it, if you go with some dev branch not named `main` then I can switch MNE-Python to use that branch instead in our pip-pre job that uses the NumPy 2.0 dev wheel. FWIW in principle you should be able to write code that is 2.0 and 1.x compatible so I'm not sure there is too much value added in splitting 🤷 

Failure should be fixed by https://github.com/matplotlib/cmocean/pull/100 assuming you use the dev version of `cmocean` in your NumPy 2.0 CIs currently (haven't checked but can implement if not once that PR is merged)

user27182:
> I think this is small and targeted enough to immediately land on main rather than our numpy-2.0 dev branch

I agree. Perhaps we should merge this ASAP as duplicates are popping up (e.g. #5473).

banesullivan (dismissed):
Reverted to the original commit against `main`. Let's get this in!

banesullivan (approved):

PR #5464, opened 2024-01-13
Merge `core._typing` package into `core._typing_core` namespace

user27182:
Follow-up from #5285 to remove the old `_typing_core.py` file and rename the new `_typing` package to retain the `_typing_core` namespace.


tkoyama010 (approved):
LGTM

PR #5465, opened 2024-01-13
Update Sphinx PyVista Plot Directive information

tkoyama010:
### Overview

Update Sphinx PyVista Plot Directive information.

### Details

- Interactive scene information is added.

### DIFF
--- a/doc/source/extras/plot_directive.rst
+++ b/doc/source/extras/plot_directive.rst
@@ -2,15 +2,16 @@
 
 Sphinx PyVista Plot Directive
 =============================
-You can generate static images of pyvista plots using the
+You can generate static and interactive scene of pyvista plots using the
 ``.. pyvista-plot::`` directive by adding the following to your
 ``conf.py`` when building your documentation using Sphinx.
 
 .. code:: python
 
    extensions = [
-       "sphinx.ext.napoleon",
        "pyvista.ext.plot_directive",
+       "pyvista.ext.viewer_directive",
+       "sphinx_design",
    ]
 
 You can then issue the plotting directive within your sphinx
@@ -34,6 +35,23 @@ Which will be rendered as:
    >>> sphere = pyvista.Sphere()
    >>> out = sphere.plot()
 
+.. note::
+
+   You need to install following packages to build interactive scene.
+
+   * `jupyter_sphinx>=0.5.3`
+   * `sphinx-design>=0.5.0`
+   * `trame>=3.5.0`
+   * `trame-vtk>=2.6.3`
+   * `trame-vuetify>=2.3.1`
+
+.. note::
+
+   You need to spin up a local http server to view the interactive scene in documentation.
+
+   .. code:: bash
+
+      python -m http.server 11000 --directory _build/html
 
 .. automodule::
    pyvista.ext.plot_directive


doc/source/extras/plot_directive.rst
```diff
@@ -34,6 +35,23 @@ Which will be rendered as:
...
+
+   * `jupyter_sphinx>=0.5.3`
+   * `sphinx-design>=0.5.0`
+   * `trame>=3.5.0`
+   * `trame-vtk>=2.6.3`
+   * `trame-vuetify>=2.3.1`
+
+.. note::
+
+   You need to spin up a local http server to view the interactive scene in documentation.
```
tkoyama010, line 50:
```suggestion
   You need to spin up a local server to view the interactive scene in the documentation.
```

doc/source/extras/plot_directive.rst
```diff
@@ -2,15 +2,16 @@
 
 Sphinx PyVista Plot Directive
 =============================
-You can generate static images of pyvista plots using the
+You can generate static and interactive scene of pyvista plots using the
```
ChristosT, line 5: scene -> scenes

doc/source/extras/plot_directive.rst
```diff
@@ -34,6 +35,25 @@ Which will be rendered as:
    >>> sphere = pyvista.Sphere()
    >>> out = sphere.plot()
 
+.. note::
+
+   You need to install the following packages to build the interactive scene.
+
+   * `jupyter_sphinx>=0.5.3`
+   * `jupyterlab>=4.0.10`
+   * `osmnx>=1.8.1`
```
ChristosT, line 44: why `osmnx` ? It is not seem directly relevant to the interactive infrastructure. 
Same with `jupyterlab` I am not sure it is strictly required. Although it does not hurt.

Do we really want to have explicit packages here ? I am sure we will forget to update them at some point. Why not point to the `requirements_docs.txt` file ? OR maybe we can add some comments in that file saying which packages are required in general and which for pyvista specifically.

ChristosT (commented):
Nice improvements overall !

doc/source/extras/plot_directive.rst
tkoyama010:
```suggestion
You can generate static and interactive scenes of pyvista plots using the
```

doc/source/extras/plot_directive.rst
tkoyama010: Sorry, I was wrong about needing osmx. We can confirm required package by looking at the following repository I will add a note to the CONTRIBUTION so that when the dynamic plots are updated, this repository will also be updated.
https://github.com/pyvista/pyvista-docs-dynamic

MatthewFlamm: I agree with this comment, it will be hard to maintain here.  I like the idea of pointing to the requirements.  Can we instead link to the https://github.com/pyvista/pyvista-docs-dynamic/blob/main/requirements.txt by somehow embedding it into the docs there and providing logical commenting inside?

tkoyama010: Thanks for your comments. Resolved in 6dfd959a981cc7403d817fa0b8f2313a0e38bada . Please check to see if we have reflected your intentions.

tkoyama010:
@pyvista-bot preview

ChristosT (approved):

PR #5484, opened 2024-01-17
Fix plotting of ambiguous cell data: set preference based on actual field association

### LINKED ISSUE #5461: `plot` does not show cell scalars by default when only cell scalars exist

Calling `add_mesh` or `mesh.plot()` does not plot the scalars when `n_points == n_cells` and when only cell scalars exist.

```python
import pyvista as pv
c = pv.Cone()
c.cell_data['ids'] = list(range(c.n_cells))

# These are the conditions needed to reproduce the bug
assert(c.n_cells == c.n_points)
assert(len(c.array_names) == len(c.cell_data.keys())) # only one scalar array is defined

c.plot()
```
Actual result
image

Expected result
image

As a workaround, the expected result is obtained using `c.plot(preference='cell')`, but this should not be necessary since only cell scalars exist.

### System Information

```shell
Date: Fri Jan 12
                OS : Darwin
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
               RAM : 16.0 GiB
       Environment : IPython
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M2
  MathText Support : True
  Python 3.12.1 (v3.12.1:2305ca5144, Dec  7 2023, 17:23:38) [Clang 13.0.0
  (clang-1300.0.29.30)]
           pyvista : 0.44.dev0
               vtk : 9.3.0
             numpy : 1.26.3
        matplotlib : 3.8.2
            scooby : 0.9.2
             pooch : 1.8.0
            pillow : 10.2.0
           imageio : 2.33.1
           IPython : 8.20.0
          colorcet : 3.0.1
           cmocean : 3.0.3
        ipywidgets : 8.1.1
              tqdm : 4.66.1
            meshio : 5.3.4
    pytest_pyvista : 0.1.8
             trame : 3.5.1
      trame_client : 2.14.2
      trame_server : 2.15.0
         trame_vtk : 2.6.3
     trame_vuetify : 2.3.1
      nest_asyncio : 1.5.8
```

### Screenshots

_No response_

MatthewFlamm:
### Overview

Fixes #5461 

### Details

Reset preference based on the actual field association value for a string based scalars input to `add_mesh`.  Note the existing test uses `rgb=True`, which uses a different branch of logic.


tkoyama010 (approved):
LGTM

PR #5489, opened 2024-01-18
Lock closed issues and pull requests after a period of inactivity

tkoyama010:
### Overview

See: https://github.com/sphinx-doc/sphinx/pull/9417.

### Details

- None

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -1,0 +1,20 @@
+name: Lock old threads
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  action:
+    if: github.repository_owner == 'pyvista'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v3
+        with:
+          github-token: ${{ github.token }}
+          issue-inactive-days: "30"
+          pr-inactive-days: "30"


MatthewFlamm:
I don't have a strong opinion on whether to implement or not implement, but my personal preference is to keep the current state. Having issues and PR's locked feels exclusionary.  If we are observing that people are adding off-topic comments to old issues, then this would outweigh this feeling to me.  If we implement, it would also feel less exclusionary to me if the time period is longer than 30 days.  6 months? 12 months?

tkoyama010:
Thank you for your comment. I suggested this because there was a comment in a previous PR, which was an example of a comment that might have been missed. However, I may have been closed-minded in trying to respond perfectly. I changed my mind want to do this in a way that I could get people to comment more comfortably.

banesullivan:
> If we are observing that people are adding off-topic comments to old issues, then this would outweigh this feeling to me

Unfortunately I have seen this more than a handful of times. It's not often, but it does happen and when it does, I try to comment with encouragement to start a new issue then lock the old thread manually. 

> If we implement, it would also feel less exclusionary to me if the time period is longer than 30 days. 6 months? 12 months?

I agree that something like 12 months or even 24 months would be most appropriate as we have often re-opened issues we thought were closed after a release.

Overall, I'm indifferent to this. I think it would benefit my notification inbox to enable this, but I don't think it is necessary (at this time at least).

tkoyama010:
Changing my mindset is better.
I will be open mind for comment to old issue. Thank you all for your comments.

PR #5495, opened 2024-01-19
Update CellArray to use ImportLegacyFormat

darikg:
### Overview

@Louis-Pujol [showed](https://github.com/pyvista/pyvista/discussions/5393#discussioncomment-8116593) that `vtkCellArray.ImportLegacyFormat` is faster than both the current `SetData` method and their earlier proposed cython version. This implements that. This has the pleasant side effect of no longer requiring `n_cells` as a perfomance optimization. 

### Details

- `pv.CellArray` uses `ImportLegacyFormat`
- Deprecate `n_cells` in the `pv.CellArray` constructor
- Also deprecate `deep` in the `pv.CellArray` constructor. (Since `vtkCellArray` needs to copy the data in the legacy format into separate connectivity and offsets arrays, the `deep` parameter was useless even before this PR)
- Deprecate `n_faces`, `n_lines`, `n_strips`, and `n_verts` in the `pv.PolyData` constructor since these were just relayed to the `pv.CellArray` constructor
- Allow the `pv.PolyData` constructor params `faces`, `lines`, `strips`, and `verts` to be ~~`pv.CellArray`~~ `vtkCellArray` objects in addition to the legacy format specifications. This allows some conveniences like `pv.PolyData(pts, lines=pv.CellArray.from_regular_cells(lines))` and `pv.PolyData(pts, faces=other_mesh.GetPolys())`
- Add documentation and widen type definitions on `faces`, `lines`, `strips`, and `verts` `@property.setter`s to make it clear that they can be assigned `pv.CellArray`s in addition to the legacy format. It was already possible to do this, but it was undocumented.


Louis-Pujol (commented):
Thanks @darikg. Everything looks good to me, I love the way this new implementation simplifies the PolyData creation ! I just feel that the possibility of creating PolyDatas with `faces`, `lines`, `strips` or `verts` being `vtk.CellArray` should be emphasized more in the docstring of the class.

pyvista/core/pointset.py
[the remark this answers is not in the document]
darikg: Good call!

darikg:
@bjlittle changing this pr to a draft because the geovista integration tests are failing due to `n_faces` deprecation warnings in PolyData(faces=..., n_faces=...) constructions. Easy to remove but one potential issue I could foresee is if users update geovista without updating pyvista, everything would still work but there could be performance regressions. Not sure how you synchronize geovista/pyvista version updates.

bjlittle:
@darikg Thanks for the heads-up :+1: 

Makes for an interesting situation ... you guys are testing against the `main` branch of `geovista`, but for stability I only test against a tagged release of `pyvista`. 

This is why I was slightly surprised to see this issue happen on `pyvista`. But I understand what's happening now.

As a workflow, it seems to me that in such situations like this where a new deprecation warning is introduced in the pull-request that will cause downstream packages (such as `geovista`) to raise the deprecation warning is unavoidable.

I can't see anything I can do on my side to prevent the integration test failure on `pyvista` other than to be responsive to the warning and simply register the deprecation warning in the `filterwarnings` of `pytest` in the `geovista` [pyproject.toml](https://github.com/bjlittle/geovista/blob/main/pyproject.toml#L84-L90).

Then in the next release of `pyvista`:

- remove the entry in the `filterwarnings`
- change `geovsita` to avoid the deprecation
- bump the min pin of `pyvista`

Does this seem reasonable? Am I missing a better approach? 

Let me know what you think @darikg and @pyvista/developers :thinking:

bjlittle:
> @bjlittle changing this pr to a draft because the geovista integration tests are failing due to `n_faces` deprecation warnings in PolyData(faces=..., n_faces=...) constructions. Easy to remove but one potential issue I could foresee is if users update geovista without updating pyvista, everything would still work but there could be performance regressions. Not sure how you synchronize geovista/pyvista version updates.

Hmmm I guess we simply introduce a minimum pin for `pyvista` in `geovista` to ensure that on release the packages stay in step with each other. I don't have any other suggestions up my sleeve, but happy to roll with this approach and see how things work out 👍

bjlittle:
@darikg You can now take this pull-request out of draft, as I've banked the appropriate changes over on `geovista` for the `main` branch 👍

darikg:
Makes sense to me @bjlittle. Thanks!

Louis-Pujol (approved):
LGTM

tkoyama010:
@Louis-Pujol Thanks for your review! I have just invited you to join our team. Next time, your approval will make the merge possible.

pyvista/core/_typing_core/_aliases.py
```diff
@@ -18,3 +18,5 @@
 
 TransformLike = Union[Matrix[float], vtkMatrix3x3, vtkMatrix4x4, vtkTransform]
 BoundsLike = Tuple[Number, Number, Number, Number, Number, Number]
+CellsLike = Union[Matrix[int], Vector[int]]
+CellArrayLike = Union[CellsLike, vtkCellArray]
```
user27182, line 22: This is perhaps a bit counter-intuitive, but I think the TypeVar needs to be explicitly specified here. Otherwise, the type will default to `Any` as in
`CellArrayLike = Union[CellsLike[Any], vtkCellArray]`
```suggestion
CellArrayLike = Union[CellsLike[int], vtkCellArray]
```

pyvista/core/cell.py
```diff
@@ -614,54 +618,54 @@ class CellArray(_vtk.vtkCellArray):
 
     def __init__(
         self,
-        cells: Optional[Union[Matrix[int], Vector[int]]] = None,
+        cells: Optional[CellsLike] = None,
```
user27182, line 621:
```suggestion
        cells: Optional[CellsLike[int]] = None,
```

pyvista/core/pointset.py
```diff
@@ -617,16 +688,16 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
     def __init__(
         self,
         var_inp: Union[_vtk.vtkPolyData, str, Matrix] = None,
-        faces: Optional[Vector[int]] = None,
+        faces: Optional[CellArrayLike] = None,
```
user27182, line 691:
```suggestion
        faces: Optional[CellArrayLike[int]] = None,
```

pyvista/core/pointset.py
```diff
@@ -617,16 +688,16 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
...
         n_faces: Optional[int] = None,
-        lines: Optional[Vector[int]] = None,
+        lines: Optional[CellArrayLike] = None,
```
user27182, line 693:
```suggestion
        lines: Optional[CellArrayLike[int]] = None,
```

pyvista/core/pointset.py
```diff
@@ -617,16 +688,16 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
...
         n_lines: Optional[int] = None,
-        strips: Optional[Vector[int]] = None,
+        strips: Optional[CellArrayLike] = None,
```
user27182, line 695:
```suggestion
        strips: Optional[CellArrayLike[int]] = None,
```

pyvista/core/pointset.py
```diff
@@ -617,16 +688,16 @@ class PolyData(_vtk.vtkPolyData, _PointSet, PolyDataFilters):
...
         n_strips: Optional[int] = None,
         deep: bool = False,
         force_ext: Optional[str] = None,
         force_float: Optional[bool] = True,
-        verts: Optional[Vector[int]] = None,
+        verts: Optional[CellArrayLike] = None,
```
user27182, line 700:
```suggestion
        verts: Optional[CellArrayLike[int]] = None,
```

pyvista/core/pointset.py
```diff
@@ -756,19 +842,26 @@ def verts(self) -> NumpyArray[int]:  # numpydoc ignore=RT01
...
+        >>> mesh.verts = pv.CellArray.from_regular_cells(
+        ...     np.arange(mesh.n_points).reshape((-1, 1))
+        ... )
         """
         return _vtk.vtk_to_numpy(self.GetVerts().GetData())
 
     @verts.setter
-    def verts(self, verts: Vector[int]):  # numpydoc ignore=GL08
-        if isinstance(verts, CellArray):
+    def verts(self, verts: CellArrayLike):  # numpydoc ignore=GL08
```
user27182, line 854:
```suggestion
    def verts(self, verts: CellArrayLike[int]):  # numpydoc ignore=GL08
```

pyvista/core/pointset.py
```diff
@@ -785,8 +878,8 @@ def lines(self) -> NumpyArray[int]:  # numpydoc ignore=RT01
         return _vtk.vtk_to_numpy(self.GetLines().GetData()).ravel()
 
     @lines.setter
-    def lines(self, lines: Vector[int]):  # numpydoc ignore=GL08
-        if isinstance(lines, CellArray):
+    def lines(self, lines: CellArrayLike):  # numpydoc ignore=GL08
```
user27182, line 881:
```suggestion
    def lines(self, lines: CellArrayLike[int]:  # numpydoc ignore=GL08
```

pyvista/core/pointset.py
```diff
@@ -856,8 +952,8 @@ def faces(self) -> NumpyArray[int]:  # numpydoc ignore=RT01
         return array
 
     @faces.setter
-    def faces(self, faces: NumpyArray[int]):  # numpydoc ignore=GL08
-        if isinstance(faces, CellArray):
+    def faces(self, faces: CellArrayLike):  # numpydoc ignore=GL08
```
user27182, line 955:
```suggestion
    def faces(self, faces: CellArrayLike[int]):  # numpydoc ignore=GL08
```

pyvista/core/pointset.py
```diff
@@ -1042,8 +1132,8 @@ def strips(self) -> np.ndarray:  # numpydoc ignore=RT01
         return _vtk.vtk_to_numpy(self.GetStrips().GetData())
 
     @strips.setter
-    def strips(self, strips):  # numpydoc ignore=GL08
-        if isinstance(strips, CellArray):
+    def strips(self, strips: CellArrayLike):  # numpydoc ignore=GL08
```
user27182, line 1135:
```suggestion
    def strips(self, strips: CellArrayLike[int]):  # numpydoc ignore=GL08
```

user27182 (commented):
Suggestion to make type annotations more explicit

tkoyama010 (changes requested):
Block the merge until the above discussion is complete.

pyvista/core/_typing_core/_aliases.py
darikg: *what?!* 

actually if I write `check_cellslike.py`:

```python
from typing import reveal_type
from core._typing_core import CellsLike

x: CellsLike = [3, 0, 1, 2]
reveal_type(x)

y: CellsLike[int] = [3, 0, 1, 2]
reveal_type(y)
```

and mypy 1.7.1 says:
```
check_cellslike.py:6: note: Revealed type is "Any"
check_cellslike.py:9: note: Revealed type is "Any"
```

and now i'm even more confused. trying to reduce the issue, `check_cellslike_minimal.py`:

```python
from typing import reveal_type, TypeVar, List, Union

T = TypeVar('T')
Vector = List[T]
Matrix = List[List[T]]
CellsLike = Union[Vector[int], Matrix[int]]

x: CellsLike = [3, 0, 1, 2]
reveal_type(x)

y: CellsLike[int] = [3, 0, 1, 2]
reveal_type(y)
```

gives the (expected):

```
check_cellslike_minimal.py:9: note: Revealed type is "Union[builtins.list[builtins.int], builtins.list[builtins.list[builtins.int]]]"
check_cellslike_minimal.py:11: error: Bad number of arguments for type alias, expected: 0, given: 1  [type-arg]
check_cellslike_minimal.py:12: note: Revealed type is "Union[builtins.list[builtins.int], builtins.list[builtins.list[builtins.int]]]"
```

is this related to the custom mypy plugin from #5285? Or some other mypy black magic?

user27182: My mistake, I believe the implementation as-is is correct since  `Matrix` in `CellsLike` is bound  to `int` explicitly, and not to a TypeVar.  No need to make any of the suggested changes.

The concern I had would only apply in the case where `CellsLike` would be generic, e.g. (slightly modified version of your second example):

``` python
from typing import reveal_type, TypeVar, List, Union

T = TypeVar('T')
Vector = List[T]
Matrix = List[List[T]]
IntType = TypeVar('IntType', bound=int) 
CellsLike = Union[Vector[IntType], Matrix[IntType]]

x: CellsLike = [3, 0, 1, 2]
reveal_type(x)

y: CellsLike[int] = [3, 0, 1, 2]
reveal_type(y)
```

which yields the following, with an assumed `Any` type.
```
<string>:10: note: Revealed type is "Union[builtins.list[Any], builtins.list[builtins.list[Any]]]"
<string>:13: note: Revealed type is "Union[builtins.list[builtins.int], builtins.list[builtins.list[builtins.int]]]"
```

For your first example, how are you calling `mypy`? It could be that mypy can't find pyvista, so it doesn't know the type, and therefore spits out `Any`. E.g. doing:
``` bash
mypy -c 'from pyvista.core._typing_core import Vector; x: Vector[int]; reveal_type(x)' --follow-imports skip
```
will yield
``` python
note: Revealed type is "Any"
```
because of the `--follow-imports skip` flag. But running without the flag will yield the expected result.

tkoyama010 (approved):

darikg:
thanks @user27182, starting a new comment thread here outside of the closed review prior

i just created a new environment to double check.

```shell
cd pyvista
conda create -n pvtest python=3.11
conda activate pvtest
pip install -e .
pip install mypy npt_promote
python -c "import pyvista; print ('ok')"  # prints ok
```

 `mypy -c "import pyvista"` starts with

```
pyvista\core\utilities\cell_type_helper.py:6: error: Name "vtkCommonDataModel" already defined (by an import)  [no-redef]
pyvista\plotting\_vtk_gl.py:17: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]
pyvista\core\_vtk_core.py:19: error: Module "vtkmodules.vtkFiltersExtraction" has no attribute "vtkExtractEdges"; maybe "vtkExtractGrid"?  [attr-defined]
```

and goes on for 100 lines or so. and `mypy check_cellslike.py` still says

```
check_cellslike.py:5: note: Revealed type is "Any"
check_cellslike.py:8: note: Revealed type is "Any"
```

user27182:
> and goes on for 100 lines or so.

Yes the output will also spit out a bunch of unrelated mypy errors. These mypy errors should likely be resolved separately (i.e. these errors are a bug). If you look closely through the output though, you should be able to find the only case where the type is revealed.

You can grep the output to isolate the revealed type, e.g.:
``` bash
mypy -c 'from pyvista.core._typing_core import Vector; x: Vector; reveal_type(x)' | grep --line-buffered 'Revealed type'
```

 > `mypy check_cellslike.py` still says `check_cellslike.py:5: note: Revealed type is "Any"`

I recall I had many frustrating import issues while developing `npt-promote`. Have a look at this [mypy documention](https://mypy.readthedocs.io/en/stable/running_mypy.html#how-imports-are-found).

Also, I just noticed that the `check_cellslike.py` code has this line:
``` python
from core._typing_core import CellsLike
```

which is missing the root package `pyvista`, i.e. it should be:
```
from pyvista.core._typing_core import CellsLike
```

 So that may be the problem with the import there.

darikg:
ahhhh thank you for your patience, I understand now, it was a bewildering combination of me 

  - not importing `CellsLike` properly
  - not grepping for 'Revealed type' in the mypy output
  - not realizing `mypy: error: May only specify one of: module/package, files, or command.` when copy/pasting your command line suggestions stemmed from the differences of Windows and Linux handling of single vs double quotes (this one's a little embarassing to admit but it's here for the record)

so the definitions of `CellsLike` and `CellArrayLike` are fine as is, right?

user27182:
> ahhhh thank you for your patience

You're welcome, glad it could be resolved.

> so the definitions of `CellsLike` and `CellArrayLike` are fine as is, right?

Yes, I believe they are correct as-is. Sorry for the false alarm!

PR #5496, opened 2024-01-19
More numpy 2.0 fixes: ndarray.ptp and np.ComplexWarning

MatthewFlamm:
### Overview

More fixes for numpy2.0.  numpy2.0 devdocs and migration guide is either out of date or is missing key information.

### Details

ptp error example:

```txt
>       ranges = np.array(self.bounds).reshape(-1, 2).ptp(axis=1)
E       AttributeError: `ptp` was removed from the ndarray class in NumPy 2.0. Use np.ptp(arr, ...) instead.
```

`ComplexWarning` was put into `np.exceptions` in v2.0, but we also need older numpy support for vtk 9.0.3 which do not have this module.  So we have to do a dance with these warning until we drop support for vtk 9.0.3.

### DIFF
--- a/examples/01-filter/image-fft-perlin-noise.py
+++ b/examples/01-filter/image-fft-perlin-noise.py
@@ -160,7 +160,7 @@ def warp_low_pass_noise(cfreq, scalar_ptp=np.ptp(sampled['scalars'])):
 
     # on the right: scale to fixed warped height
     output_scaled = output.translate((-11, 11, 0), inplace=False)
-    output_scaled['scalars_warp'] = output['scalars'] / np.ptp(output['scalars'].ptp) * scalar_ptp
+    output_scaled['scalars_warp'] = output['scalars'] / np.ptp(output['scalars']) * scalar_ptp
     warped_scaled = output_scaled.warp_by_scalar('scalars_warp')
     warped_scaled.active_scalars_name = 'scalars'
     # push center back to xy plane due to peaks near 0 frequency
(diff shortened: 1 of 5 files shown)


examples/01-filter/image-fft-perlin-noise.py
```diff
@@ -160,7 +160,7 @@ def warp_low_pass_noise(cfreq, scalar_ptp=sampled['scalars'].ptp()):
 
     # on the right: scale to fixed warped height
     output_scaled = output.translate((-11, 11, 0), inplace=False)
-    output_scaled['scalars_warp'] = output['scalars'] / output['scalars'].ptp() * scalar_ptp
+    output_scaled['scalars_warp'] = output['scalars'] / np.ptp(output['scalars'].ptp) * scalar_ptp
```
MatthewFlamm, line 163:
```suggestion
    output_scaled['scalars_warp'] = output['scalars'] / np.ptp(output['scalars']) * scalar_ptp
```

pyvista/plotting/lookup_table.py
```diff
@@ -1058,6 +1058,8 @@ def to_opacity_tf(self):
         """
         opacity_tf = _vtk.vtkPiecewiseFunction()
         for ii, value in enumerate(self.values[:, 3]):
+            if not isinstance(value, np.uint8):
+                raise ValueError("not uint8 here")
```
MatthewFlamm, line 1062:
```suggestion
```

tkoyama010 (approved):
LGTM

PR #5500, opened 2024-01-20
WIP : Working towards fixing issues arising from VTK changes.

berkgeveci:
VTK is adding support for properties in its wrapping. This is causing issues with pyvista, especially where subclassing and keyword args are used in the constructors. This is a first attempt at fixing some of the issues.

### Overview

The VTK team is working on improving the VTK wrapping by adding support for Python properties
and constructor keyword arguments to set these properties. See:

https://gitlab.kitware.com/vtk/vtk/-/merge_requests/10820

These changes are causing issues in PyVista. Mainly, with classes that inherit from VTK classes using
multiple inheritance where the first superclass is a VTK class and one of the other superclasses has
properties with the same name as a (new) VTK property. Example: point_data in dataset classes. With
the change in VTK, the PyVista classes now inherit the VTK property causing API issues. The second
problem occurs when keyword arguments are used in the constructor of VTK subclass. These keyword
arguments are now interpreted by VTK and used to assign values to properties, which is different
than current behavior. Note that VTK ends up calling the PyVista property setters with default values
such as None, which are normally ignored (or lead to the setting of default values) before these changes.


pyvista/plotting/_property.py
berkgeveci, line 1: This whole block is unnecessary. There is nothing wrong with having default arguments in the constructor. They don't lead to anything being called on the VTK side. It is calling the constructor with keyword arguments with "invalid" arguments that cause the issue. Such as:
```
Property(metallic=None)
```
So the fix needs to be where `Property` is constructed.

darikg: Since `pv.Property.__init__` never calls `vtkProperty.__init__` why does anything need to change at all? Is it the case that 
```
class Property(vtkProperty)
    @property
    def metallic(self) -> ...

    @metallic.setter
    def metallic(self) -> ...
```

doesn't actually override the new VTK python metallic property?

berkgeveci: Thank you for your comment. It led me to investigate the issue more and I found out that this is a bug in the new wrapper code. The keyword argument parsing code is currently happening in the new method of the type (equivalent of `__new__()`) and it is always executing whether or not `__init__()` of `Property` calls `super`'s `__init__()`. Moving the parsing to the `init` method fixes the issue. With this change, there will have to be only minimal changes to PyVista - just the ones that override the properties of the VTK superclass. Phew. I am very relieved.

tkoyama010:
pre-commit.ci autofix

berkgeveci:
Here is an updated commit against the latest version of the VTK wrapper. Fewer changes and much cleaner. Probably not 100% as I have likely missed some other subclasses of VTK but it should give a good idea about what needs to be fixed and how.

tkoyama010:
pre-commit.ci autofix

darikg:
I still find the the `point_data = Grid.point_data` class variable reassignments a little distasteful, but I fully admit to not having a better solution. 

Just thinking out loud here: I guess the issue is that for *methods* in multiple inheritance it's easy to specify `super(self, PyvistaClass).some_method()` vs `super(self, VtkClass).some_method()` but there's no such equivalent for properties. Which is why changing the order of the superclass list is doomed either way -- at some point you're going to want to specify the superclass property from a non-leftmost-parent. And furthermore surely the natural position of the VTK class is leftmost anyway, as is the status-quo.

I think ideally pyvista's mixin classes would be pure method-only and only the final concrete classes could do something like `point_data = PointData()` to define properties so there's never any reassignment. But that would obviously be a massive sprawling architectural overhaul.

berkgeveci:
> I think ideally pyvista's mixin classes would be pure method-only and only the final concrete classes could do something like point_data = PointData() to define properties so there's never any reassignment. But that would obviously be a massive sprawling architectural overhaul.

Actually, I don't think that this would be too hard to do. You could have the PointData() method defined in pv.Dataset that returns the point_data property. I guess that it would look like this:

class DataSet(...):
     @property
      def point_data(self):
     ....
     def PointData(self):
          return point_data

darikg:
Hmm, here's another approach

```python
class VtkClass:
    @property
    def prop(self):
        return 'vtk'

class PyvistaSuperClass:
    @property
    def prop(self):
        return 'pyvista'

    def __init_subclass__(cls):
        # Make sure all subclasses use the pyvista property instead of the vtk one
        cls.prop = __class__.prop

class PyvistaSubclass1(VtkClass, PyvistaSuperClass):
    pass

class PyvistaSubclass2(VtkClass, PyvistaSuperClass):
    pass

PyvistaSubclass1().prop, PyvistaSubclass2().prop  # Both return 'pyvista'
```

the advantage is only the superclasses need to specify the override, instead of every subclass

banesullivan:
Big +1 to the use of `__init_subclass__` for this. That could be the elegant solution we're looking for

banesullivan:
Before this merges, I'd like to set up a CI job to run our full test suit against the new version of VTK (we might be able to just hardcode the url of the CI artifacts form VTK to pip install it). I'll happily take this on, but its something I'll have to make time for on a weekend. 

If we are thinking this is looking good / is compatible and VTK merges [the MR](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/10820) to `master`, then we can use an existing workflow with a manual dispatcher to test against VTK's nightly wheels: https://github.com/pyvista/pyvista/actions/workflows/vtk-pre-test.yml

berkgeveci:
Sounds great @banesullivan. Thanks! We'll hold off merging until this is sorted.

PR #5509, opened 2024-01-24
Add `background_opacity` argument to `add_legend`

tkoyama010:
### Overview

Add `background_opacity` argument to `add_legend`.

### Details

- None


tkoyama010:
@pyvista-bot LGTM

tests/plotting/test_renderer.py
```diff
@@ -177,3 +178,11 @@ def test_legend_face(sphere, face):
     pl = pv.Plotter()
     pl.add_mesh(sphere, label='sphere')
     pl.add_legend(face=face)
+
+
+def test_add_legend_background_opacity(sphere):
+    background_opacity = default_rng(0).random()
```
MatthewFlamm, line 184: Small point: Is there a reason not to hard code a value here?  Seeding a RNG will result in the same value each time as well, we might as well make it explicit for easier human understanding.

pyvista/plotting/renderer.py
```diff
@@ -3357,6 +3358,9 @@ def add_legend(
             or ``'arial'``. Defaults to :attr:`pyvista.global_theme.font.family
             <pyvista.plotting.themes._Font.family>`.
 
+        background_opacity : int | float, default: 1.0
```
MatthewFlamm, line 3361: I'm assuming that it should be a float input, but an int can be promoted to a float in python/vtk?

```suggestion
        background_opacity : float, default: 1.0
```

MatthewFlamm:
#3015 is still being kept current.  I haven't checked.  But if there will be a conflict, it might be better to hold off here until that one is merged?

pyvista/plotting/renderer.py
tkoyama010: Good catch. Thanks.

tests/plotting/test_renderer.py
tkoyama010: Only my mistake. I want to test each time with a random value between 0 and 1. Thank you for catching it.

```suggestion
    background_opacity = default_rng().random()
```

tkoyama010: I see, that is easier to understand.

tests/plotting/test_renderer.py
```diff
@@ -177,3 +178,11 @@ def test_legend_face(sphere, face):
     pl = pv.Plotter()
     pl.add_mesh(sphere, label='sphere')
     pl.add_legend(face=face)
+
+
+def test_add_legend_background_opacity(sphere):
+    background_opacity = default_rng().random()
```
tkoyama010, line 184:
```suggestion
    background_opacity = 0.4
```

tkoyama010:
pre-commit.ci autofix

tests/plotting/test_renderer.py
```diff
@@ -1,3 +1,4 @@
+from numpy.random import default_rng
```
MatthewFlamm: Last bit

```suggestion
```

tkoyama010: Thanks!

MatthewFlamm (approved):
LGTM but worth looking at #3015 before merging.

tkoyama010:
Thanks! I will address the conflicts in #5509, as changing by fixing the conflicts here would require another approval.

Issue #5514, opened 2024-01-25
Testing against VTK master branch is failing


banesullivan:
Looks like our routine tests against VTK's master branch are failing for 5 different tests:

```
FAILED tests/core/test_reader.py::test_openfoam_decompose_polyhedra - AttributeError: 'vtkmodules.vtkIOParallel.vtkPOpenFOAMReader' object has no attribute 'SetDecomposePolyhedra'
FAILED tests/plotting/test_picking.py::test_single_cell_picking - assert False
 +  where False = <test_picking.test_single_cell_picking.<locals>.PickCallback object at 0x7f41412f1d90>.called
FAILED tests/plotting/test_picking.py::test_cell_picking_interactive - IndexError: list index out of range
FAILED tests/plotting/test_picking.py::test_cell_picking_interactive_subplot - IndexError: list index out of range
FAILED tests/plotting/test_picking.py::test_switch_picking_type - assert []
```

See jobs under https://github.com/pyvista/pyvista/actions/workflows/vtk-pre-test.yml

cc @berkgeveci @pyvista/kitware

banesullivan:
Last successful run was on Nov 17, 2023 in 18m 15s

berkgeveci:
How does one run individual tests? Can it be done on Mac? (I read something about image tests running on Linux only?) I can try to do a bisect.

MatthewFlamm:
95+% of the tests that use image regression are in `tests/plotting/test_plotting.py`.  None of the failing tests above are in there, so it is likely OK if that is the only issue with Mac testing.  We also decorate tests to skip on Mac, so you will know.

You can run individual test by doing this `pytest tests/plotting/test_picking.py::test_cell_picking_interactive`.  Note that you can copy and paste directly from the lines above to make your life easier.

Some of these changes will require not trivial changes to the code.  Do you have a recommended way we should tag or name branches to get the testing right?  A way to run the tests via tag would be the most flexible in my opinion.

The one related to OpenFoamReader is a deprecation/removal cycle in vtk.

https://github.com/Kitware/VTK/blob/5b8982f91c2b24e66629c9d6188647659ed5e4b3/Documentation/release/9.3.md?plain=1#L973-L974

We can do as `hasattr` and guard tests appropriately.

berkgeveci:
I tracked the interaction test failures to several commits around this one: 1da4867f7a26530ce8ca60ad41eeb4af3b8ff6ae
I'll ask the author of those commits to take a look.

[...]

PR #5527, opened 2024-01-26
Use labels for vtk-master testing

MatthewFlamm:
### Overview

Should help PRs addressing #5514.  Right now, only if the branch follows some naming, will the test run.  But to be easier for contributors, we should add the possibility of adding a label.
### Details

Will test with label here by closing and reopening a few times.


.github/workflows/vtk-pre-test.yml
```diff
@@ -19,6 +20,8 @@ env:
 
 jobs:
   VTK-master:
+    # only for pull requests, run only when 'vtk_master' label exists
+    if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'vtk_master') }}
```
MatthewFlamm, line 24:
```suggestion
    if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'vtk-master') }}
```

MatthewFlamm:
This worked to run the tests with label, and skip test without.  If we want to have passing CI before merge, we can remove label and close/reopen before merging.

banesullivan (approved):
Awesome, thanks for adding this!

Issue #5533, opened 2024-01-29
Numpy 2.0 compatibility


MatthewFlamm:
### Describe what maintenance you would like added.

numpy 2.0 testing is currently be tracked in #5450.  Several fixes have already been merged. This issue is for tracking remaining issues (and add any that pop up).  Details of each in further comments.

- [ ] garbage collection errors
- [x] meshio ~incompatibility and~ errors
- [ ] array casting errors (as of release candidate 2.0.0rc2)

MatthewFlamm:
garbage collection errors (~150 tests failing)

Example error from `tests/plotting/test_actor.py::test_actor_copy_deep`.  It may be a useful comparison that `tests/plotting/test_actor.py::test_actor_from_plotter` does not have this error.

```txt
>       assert len(after) == 0, msg
E       AssertionError: Not all objects GCed:
E         vtkWeakReference: ["dict: d['dataset']", "dict: d['dataset']"]
E         vtkWeakReference: ["dict: d['dataset']", "dict: d['dataset']"]
E         vtkFloatArray: ["dict: d['VTKObject']", "dict: d['VTKObject']", '<managedbuffer object at 0x7fe309925d40>', '<managedbuffer object at 0x7fe309924740>', "dict: d['VTKObject']", "dict: d['VTKObject']"]
E         
E       assert 3 == 0
E        +  where 3 = len([<vtkmodules.vtkCommonCore.vtkWeakReference(0x55a25c494080) at 0x7fe309978ac0>, <vtkmodules.vtkCommonCore.vtkWeakReference(0x55a25bfc2030) at 0x7fe309978ee0>, <vtkmodules.vtkCommonCore.vtkFloatArray(0x55a25d0d9b10) at 0x7fe309978a60>])
```

MatthewFlamm:
meshio.  The biggest issue here is the recent lack of maintenance in `meshio`. Without this changing, we need to consider either fully dropping `meshio` as a dependency or firewalling it to be an optional dependency (and only used for numpy <2).

https://github.com/nschloe/meshio/pull/1451 fixes import problems.

We have other errors related to `pyvista.save_meshio`, snippet below.  It is unclear whether this is an internal or meshio issue at this time.  An example is from `tests/test_meshio.py::test_meshio` and the error occurs inside the `pyvista.save_meshio` code.

```txt
            for k in range(len(data)):
>               data[k] = np.asarray(data[k])
E               TypeError: 'tuple' object does not support item assignment
```

[...]

PR #5551, opened 2024-02-01
Fix `save_meshio` with Numpy v2

keurfonluu:
### Overview

- Fixed: `np.split` returns a tuple of arrays in v2 instead of a list (see discussion #5533).

### DIFF
--- a/pyvista/core/utilities/fileio.py
+++ b/pyvista/core/utilities/fileio.py
@@ -597,7 +597,10 @@ def save_meshio(filename, mesh, file_format=None, **kwargs):
     vtk_cell_data = mesh.cell_data
     indices = np.insert(np.cumsum([len(c[1]) for c in cells]), 0, 0)
     cell_data = (
-        {k.replace(" ", "_"): [v[i1 : i2] for i1, i2 in zip(indices[:-1], indices[1:])] for k, v in vtk_cell_data.items()}
+        {
+            k.replace(" ", "_"): [v[i1:i2] for i1, i2 in zip(indices[:-1], indices[1:])]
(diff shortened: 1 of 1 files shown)


MatthewFlamm (approved):
This change keeps the status quo, so it is fine.  I still suspect some sort of error or inefficiency here, but I am not well aware of the meshio format to offer detailed suggestions.

keurfonluu:
> This change keeps the status quo, so it is fine. I still suspect some sort of ~error or~ inefficiency here, but I am not well aware of the meshio format to offer detailed suggestions.

I replaced `np.split` with a simple comprehension list. Looks faster.

```python
import numpy as np

def split(x, indices):
    return [x[i1 : i2] for i1, i2 in zip(np.insert(indices, 0, 0), np.append(indices, len(x)))]

np.random.seed(0)
n = 10000
x = np.random.rand(n)
indices = np.sort(np.random.choice(n, size=n // 100, replace=False))

xn = list(np.split(x, indices))
xc = split(x, indices)

assert len(xn) == len(xc)
for x1, x2 in zip(xn, xc):
    assert np.allclose(x1, x2)

%timeit list(np.split(x, indices))
# 91.4 µs ± 6.81 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
%timeit split(x, indices)
# 43.6 µs ± 2.45 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

MatthewFlamm (approved):
Quite a minor speedup, but thank you for looking into it.  I think I was making a mountain over a molehill.

PR #5558, opened 2024-02-01
Rename `input_validation` package to `validation`

user27182:
Part of #5226.

It's necessary to separate the package rename from other changes in #5226 so that the diff there can properly reflect line-by-line changes instead of entire file deletions and additions.


tkoyama010 (approved):
LGTM

PR #5266, opened 2023-12-03
Fix redirect link to fix LinkCheck CI error

tkoyama010:
### Overview

Fix redirect link to fix LinkCheck CI error.

relate to #3166

### Details

- None

Issue #265, opened 2019-06-18
Support Pandas structures and anything "array-like"


banesullivan:
I've seen a few users get tripped up when using PyVista with Pandas as we explicitly check if something is a NumPy array. Often, a user might have an array-like structure like a Pandas `DataFrame` or `Series` and get hard to decipher errors.

Take this for example:

```py
import pandas as pd
import pyvista as pv

# Load Excel sheet using Pandas
# Note - you may need to `pip install xlrd`
df = pd.read_excel('/Users/bane/Downloads/WFA.xlsx')
coords = df[['X', 'Y', 'Z']]

# Preview the coordinates to see what we are dealing with
poly = pv.PolyData(coords)
```

which yields:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-66-f94ed2aa81cf> in <module>
      1 # Preview the coordinates to see what we are dealing with
----> 2 poly = pv.PolyData(coords)

~/Documents/OpenGeoVis/Software/pyvista/pyvista/pointset.py in __init__(self, *args, **kwargs)
    112                 self._from_arrays(points, cells, deep, verts=True)
    113             else:
--> 114                 raise TypeError('Invalid input type')
    115 
    116         elif len(args) == 2:

TypeError: Invalid input type
```

A simple `.values` call on the coords data frame would fix this but not all users may remember this.

How should we go about making PyVista compatible with array-like datastructures?

akaszynski:
Would a try/except work with ``np.asarray``?

banesullivan:
Possibly... that way anything array-like would get casted as a NumPy array

banesullivan:
I started to implement this but realized `np.asarray` can be called on any object:

```py
>>> import numpy as np
>>> import pyvista as pv
>>> np.asarray({'foo':'3'})
array({'foo': '3'}, dtype=object)
>>> np.asarray(6)
array(6)
>>> len(np.asarray(6))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-49-9cfac8a4e0fa> in <module>
----> 1 len(np.asarray(6))

TypeError: len() of unsized object
>>> np.asarray(pv.PolyData)
array(<class 'pyvista.pointset.PolyData'>, dtype=object)
```

This could lead to some wacky errors that might confuse new (and even experienced) users.

We'd probably have to implement lots of `dtpye` checking as well

AndrewAnnex:
[...]

PR #5567, opened 2024-02-02
Add support for casting `polars` array-like input to `numpy`

### LINKED ISSUE #5516: Add support for pandas and polars dataframes

Polars and especially Pandas are very core Python data processing libraries by now. It would be nice to be able to create a mesh and/or add scalars directly from these datasets instead of the (very costly) casting to Numpy arrays.
Now the Polydata creation raises: 
```
TypeError: Invalid Input type:

Expected first argument to be either a:
- vtk.PolyData
- pyvista.PolyData
- numeric numpy.ndarray (1 or 2 dimensions)
- List (flat or nested with 3 points per vertex)
- vtk.vtkDataArray

Instead got: <class 'polars.series.series.Series'>
```

And the scalar assigning takes ridiculously long time (did not manage to see if it works in the end)

_No response_

### Pseudocode or Screenshots

```python
import pyvista as pv
import polars as pl
import numpy as np

N = int(1e6)
data = pl.DataFrame(dict(
    points = np.random.sample((N, 3)),
    color = np.random.sample((N, 4))
    ))

cloud = pv.PolyData(data["points"])
cloud["colors"] = data["color"]
```

user27182:
This PR modifies `cast_to_ndarray` to attempt to gracefully deal with array-like inputs from `polars`. It is a possible initial step to fix #5516.


tkoyama010:
I haven't followed discussing #5516 well, but can we support pandas too?

user27182:
The TLDR of #5516 is that `pandas` and `polars` have array-like data structures which are implicitly supported anywhere `np.asarray` is used. Not only that, but calling `np.asarray` likely does not result in any copies of the data being made through the entire `pandas` (or `polars`) -> `numpy` -> `vtk` chain since they all make use of numpy's array APIs. The catch, though, is that the data must be contiguous.  And `polars` does not store data in its data frames contiguously by default, so calling `np.asarray` on a `polars` data frame will raise an error.

This PR is effectively a workaround for the `polars` error that is generated.

Another possible issue with supporting `pandas` was mentioned in #265, where using `.values` was needed to make use of `pandas` data. However, this is no longer the case, and using `.values` is [not recommended](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.values.html). Calling `np.asarray` will implicitly call `to_numpy` on the data and just work.

The other issue mentioned in #265 is that calling `np.asarray` will happily create object arrays. However, this case is already dealt with by `cast_to_ndarray`. The function will still allow string arrays, though, so I suppose we could add a dtype check for `np.number` types only (if we want to support complex numbers) or `np.floating` and `np.integer`. If we want to make this vtk-safe we should not allow complex numbers since [`numpy_to_vtk`](https://github.com/Kitware/VTK/blob/1df5e2c0629462569ac491a54729514f26db8256/Wrapping/Python/vtkmodules/util/numpy_support.py#L104) does not handle complex arrays.

**If we add the dtype check then I think this PR would effectively support both `polars` and `pandas`.**

The next steps (future PRs) to support these types would be to either:
1. replace `isinstace(x, np.array)` checks in pyvista code with `isinstace(x, _SupportsArray)` (see [#5516 (comment)](https://github.com/pyvista/pyvista/issues/5516#issuecomment-1915608738)) and then call `cast_to_ndarray(x)`, or
2. remove all instance checks and simply call `cast_to_ndarray(x)` on array-like inputs.

`cast_to_ndarray(x)` is also used by pyvista's validation package, so calling `validate_array(x)` would also work in case other validation checks are needed.

user27182:
Another quirk with `polars` is that after casting, the ndarray has its writable flag set to False by default. This is fine for plotting but could pose an issue for any pyvista methods that mutate arrays directly. 

This could just be a safe default choice by `polars` and may be ok to change this to True. `polars` is written in Rust so this kind of safe practice may stem more from a design philosophy rather than a technical requirement. Even with the flag, it may be possible for vtk to mutate the array anyway after calling `numpy_to_vtk` since vtk will access the data at the c level. 

The impact of keeping the `writable=False` has not been tested.

darikg:
Just a big picture comment here, but I wonder if it might be worth waiting for #5266 and do more integration of the validation package and an overall review of locations that check input arrays?

- `PolyData.__init__`
Checks `isinstance(var_inp, (np.ndarray, list, _vtk.vtkDataArray))` and calls `vtk_points` which does a bunch of custom array validation that should be entirely replaced by by one the validation functions

- `CellArray.__init__`
Simply calls `np.asarray` and then `pv.core.utilities.cells.numpy_to_idarr`, which could be replaced by one of the validation methods

- `DataSetAttributes.set_array` 
Calls DataSetAttributes.`_prepare_array` which does a lot of custom validation

- `core.utilities.arrays._coerce_pointslike_arg` 
Used in a bunch of locations, could be removed and replaced by `validate_arrayNx3` I think

etc. I'd feel a lot better about adding more complexity in the form of e.g. Polars support if it was all isolated to the validation package and the rest of pyvista used the validation more uniformly. E.g. when are we supposed to use `cast_to_ndarray` from the `core.utilities.arrays` package instead of `validate_array` in the validation package?

darikg:
> And polars does not store data in its data frames contiguously by default, so calling np.asarray on a series which is part of polars data frame will raise an error.

I thought it just silently copied the data?

user27182:
> it might be worth waiting for #5266 and do more integration of the validation package and an overall review of locations that check input arrays?

Yes I agree. Clean up the input validation across pyvista first, then add support for more general array-like inputs later.

> when are we supposed to use `cast_to_ndarray` from the `core.utilities.arrays` package instead of `validate_array` in the validation package?

It does seem a bit unclear. `cast_to_ndarray` was added specifically for input validation as part of #5100. It was added to the `pyvista.core.utilities.arrays` namespace at the time since it seemed to have general use. However, it may be best to move it directly into the `pyvista.core.validation` namespace, e.g. as part of #5226, and make it private and mostly just for internal use as part of `validate_array`.

darikg:
> However, it may be best to move it directly into the pyvista.core.validation namespace, e.g. as part of https://github.com/pyvista/pyvista/pull/5226, and make it private and mostly just for internal use as part of validate_array.

That sounds good! Right now it's not possible for pyvista.core.utilities.arrays to import from validation because of the circular import

user27182:
> > And polars does not store data in its data frames contiguously by default, so calling np.asarray on a series which is part of polars data frame will raise an error.
> 
> I thought it just silently copied the data?

In my view, `polars` _should_ silently copy the data. Unfortunately it does not. This code will fail:

``` python
import polars as pl
import numpy as np
data = pl.DataFrame(dict(points=[[1, 2, 3], [4, 5, 6]]))
array_in = np.asarray(data['points'])
```
``` polars.exceptions.ComputeError: 'to_numpy' not supported for dtype: List(Int64) ```

The workaround in this PR does exactly that though and will create a copy instead, although not silently (a warning is issued).

This case is covered by this PR here.
https://github.com/pyvista/pyvista/blob/8fc55b5e80b983eb1af97b11d765d4a9a66218ed/tests/core/test_utilities.py#L983-L991

IMHO I think this is surprising behaviour and may be worth raising an issue on the `polars` github. `polars` should probably be catching the exception and printing a warning directly.

darikg:
Oh! Thanks for explaining. I agree that does seem like a miss on polars' part.

tkoyama010 (dismissed):

tkoyama010 (approved):

tkoyama010 (changes requested):
Thanks. Let's wait!

Issue #5563, opened 2024-02-02
Running `mypy pyvista` reports 144 type errors which are not caught by `pre-commit run mypy`


user27182:
### Describe the bug, what's wrong, and what you expected.

There are many mypy errors when running `mypy pyvista` instead of `pre-commit run mypy`. See next section for a full list. 

Pre-commit only checks modified files, so that makes sense that they aren't caught there, but otherwise I'm not sure how these errors got there in the first place.

Running `pre-commit run mypy` uses the [pre-commit mypy mirror](https://github.com/pre-commit/mirrors-mypy) and uses the pyvista config:

https://github.com/pyvista/pyvista/blob/74d9476507d88d501261e830e5e1a80b44c44921/.pre-commit-config.yaml#L64-L75

Whereas running `mypy pyvista` uses this config:
https://github.com/pyvista/pyvista/blob/74d9476507d88d501261e830e5e1a80b44c44921/pyproject.toml#L160-L162

Both seem to have the same options,  namely just `--ignore-missing-imports=True`, but perhaps these two configs are still different? 

Either way some of these errors should be fixed or ignored. In particular, this one:
``` python
pyvista/core/filters/rectilinear_grid.py:112: error: "RectilinearGridFilters" has no attribute "dimensions"  [attr-defined]
```
actually _does_ pop up often when running `pre-commit` locally, causing me to need to use `--no-verify` to force the commit through (this error then disappears for the CI pre-commit checks, so the CI checks will still pass).

Having all these errors here also affects debugging type errors, e.g. see https://github.com/pyvista/pyvista/pull/5495#issuecomment-1905165871.

### Steps to reproduce the bug.

```bash
mypy pyvista
```
<details>
  <summary>Click to see all 144 errors</summary>

``` python
pyvista/core/utilities/cell_type_helper.py:6: error: Name "vtkCommonDataModel" already defined (by an import)  [no-redef]
pyvista/plotting/_vtk_gl.py:17: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]
pyvista/core/_vtk_core.py:19: error: Module "vtkmodules.vtkFiltersExtraction" has no attribute "vtkExtractEdges"; maybe "vtkExtractGrid"?  [attr-defined]
pyvista/core/_vtk_core.py:19: error: Name "vtkExtractEdges" already defined (possibly by an import)  [no-redef]
[...]

PR #5571, opened 2024-02-03
Rename and Simplify `validation` package

user27182:
## Overview
Update `validation` package. This PR was originally opened as #5226 as a follow up to #5100.

- ~Rename package~ (#5558)
- ~Simplify package~ (#5586)
- Add generic `ArrayLike[T]` -> `ArrayLike[T]` functionality
  - Array wrapper classes are added to process scalars, flat sequences and nested 2D sequences separately from numpy arrays
  - This removes the need to always cast inputs to a numpy array. `validate_array` will now return the input as-is by default.
  - Remove `to_list` and `to_tuple` params; replace with `return_type`param with option to specify `numpy` as a return type
  -  Most `check` functions were modified to process sequences and numpy arrays with distinct implementations
- Add overloads to `validate_array` to hint return types
  - Add new test suite for testing static type hints with mypy
- Add `get_flags` param to return info about changes to the array 
- Update definitions for `check_number`
- ~Rename `validate_arrayN_uintlike` -> `validate_arrayN_unsigned`~
- Add `check_ndim` to check array dimensionality
- Update package typing:
   - Use updated generic PyVista types, e.g.`Vector[float]` from #5171 and #5285.
   - Annotations and function overloads were added to `validate_array` and related functions 
Enhacements for follow-up PRs:
- Add `optional` param to checkers to do nothing if input is `None`

## Codecov
The `Codecov` report got lost among the commits, see it here: https://github.com/pyvista/pyvista/pull/5571#issuecomment-1930715619


user27182:
pre-commit.ci autofix

user27182:
This PR is ready for a draft review. It still needs work but the generic and array-wrapping mechanics are pretty much fully implemented. CC: @tkoyama010, @darikg

user27182:
Not sure how to fix the docs... the overloads and type annotations make a mess of them

image

user27182:
[polars.repeat](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.repeat.html) is an overloaded function but the docs don't show the overloads. The kwargs are also each displayed on a new line. I think both of these are desirable features for the `validation` docs to make them more readable.

I'm not at all familiar with sphinx configurations, but maybe there are config options specified somewhere in the `polars` source code which could help with fixing the docs here.

user27182:
#5607 seems to have fixed the main issue with the docs. The overloads are gone and the kwargs are each on a a new line.

But, the `ArrayLike` type alias expansion takes up 8 lines alone. This is not good. I think there may be a way to tell sphinx not to expand the aliases and simply show `Array[float]` for example.

image

darikg:
A couple driveby comments:

- [_coerce_pointslike_args](https://github.com/pyvista/pyvista/blob/ce6d08cbcbd6d525ea3eafe27f4e0b016fec204a/pyvista/core/utilities/arrays.py#L59) could now be replaced by something like `validate_arrayNx3(points, must_be_real=True, reshape=True, return_type='numpy')`, *except* that it has an extra return argument to disambiguate whether the original array was shape (3,) or (n, 3). Is the solution just an extra wrapper function?
- `validate_axes(must_be_orthogonal=True)` seems like it needs a tolerance value to allow vectors to be orthogonal within some precision

tkoyama010:
pre-commit.ci autofix

user27182:
> A couple driveby comments:
> 
> * [_coerce_pointslike_args] could now be replaced by something like `validate_arrayNx3(points, must_be_real=True, reshape=True, return_type='numpy')`, _except_ that it has an extra return argument to disambiguate whether the original array was shape (3,) or (n, 3). Is the solution just an extra wrapper function?

Maybe just add another kwarg to capture that functionality? Either generally in `validate_array` with a `return_reshaped_flag`, which if True, will return True/False along with the array. Or we could add something similar to `validate_array3` only?

> * `validate_axes(must_be_orthogonal=True)` seems like it needs a tolerance value to allow vectors to be orthogonal within some precision

Good idea 👍

tests/core/test_validation.py
```diff
@@ -983,3 +1094,259 @@ def test_array_from_vtkmatrix(cls, shape):
...
+
+
+from enum import Enum
+
+
+class arraylike_types(Enum):
+    Number = 0
+    NumpyArraySequence = 1
+    NumberSequence1D = 2
+    NumberSequence2D = 2
```
tkoyama010, line 1134:
```suggestion
    NumberSequence2D = 2 # noqa: PIE796
```

tkoyama010:
pre-commit.ci autofix

tests/core/test_validation.py
tkoyama010, line 1: Good change. Thanks!

pyvista/core/_validation/check.py
```diff
@@ -546,75 +652,215 @@ def check_shape(
...
+            f"Got {array_ndim}, expected {expected}."
+        )
+        raise ValueError(msg)
+
+
+def check_number(
+    num: Union[float, int, complex, np.number, Number],  # type: ignore[type-arg]
+    /,
+    *,
+    definition: Literal['abstract', 'builtin', 'numpy'] = 'abstract',
```
user27182, line 751: At one point it wasn't clear to me if support for complex numbers was needed. This function was generalized accordingly. But, the must_be_real check is a bit redundant... just use `check_real` if you need to, no? Having `must_be_real` here with default True (same as validate_array) is to make it explicit to the user though. Not sure if it should be changed.

Also, I went a bit mad because of the differences in definitions of what a number is in python. Hence the `definition` kwarg. But, it kind of complicates things unnecesarily. 

I think maybe aligning the definition of `number` with the `_NumberType` typevar is perhaps the way to go, i.e. python float or int, or numpy float or int. We could directly use `_is_Number` from `_type_guards.py` if that's the case and everything will be self-consistent.

pyvista/core/_validation/check.py
```diff
@@ -921,7 +1175,7 @@ def check_iterable_items(
         raise
 
 
-def check_contains(*, item, container, name='Input'):
+def check_contains(*, item: Any, container: Any, name: str = 'Input'):
```
user27182, line 1178: I'm not happy with this API, forcing all kwargs is awkward and inconsistent with the rest of the package.

consider changing to:
`check_contains(container: Any, *, contains: Any, name: str = 'Input')`

tests/core/test_validation.py
```diff
@@ -983,3 +1095,256 @@ def test_array_from_vtkmatrix(cls, shape):
...
+        result = mypy_api.run(['-c', code_snippet])
+        assert 'usage: mypy' not in result[1]
+        assert 'Cannot find implementation' not in result[0]
+        stdout = str(result[0]).replace('Tuple', 'tuple')
+        match = findall(r'note: Revealed type is "([^"]+)"', stdout)
+        assert match is not None
+        return match
+
+    finally:
+        os.chdir(cur)
```
user27182, line 1331: Running mypy tests like this is good and allows full flexibility and customization over how we run our tests.

One downside though is that it's somewhat non intuitive, and requires creating a separate script file for analysis.

An alternative option is to use `pytest-mypy-plugins` package. This integrates directly with pytest, and is what's used to test the `npt-promote` package.

But, I tried to use it initially and found two main hurdles: (1) specifying the mypy config and path was a bit challenging, and (2) it's not easy to isolate the tests to specific files. So, all 144+ mypy errors that are currently unresolved from #5563 get dumped into each test case and need to be filtered.

Long term though, I think using `pytest-mypy-plugins` is a more elegant solution rather than calling `mypy.api.run`

tests/core/test_validation_reveal_type.txt
```diff
@@ -0,0 +1,77 @@
+from typing import TYPE_CHECKING
```
user27182, line 1: this file is valid python code, but is a .txt file instead of .py file because the linters mess up the formatting. Not sure how to make the linters ignore an entire file, but it's an option if we want to keep it as a .py file.

pyvista/core/_validation/validate.py
```diff
@@ -33,44 +43,173 @@
...
+    array: _ArrayLikeOrScalar[_NumberType],
+    *,
+    dtype_out=None,
+    return_type=None,
+    **kwargs: Unpack[_TypedKwargs],
+) -> NumpyArray[_NumberType]: ...
+
+
+def validate_array(
+    array: _ArrayLikeOrScalar[_NumberType],
```
user27182, line 164: Still need to resolve the docs issues with type alias expansion

pyvista/core/_validation/validate.py
```diff
@@ -91,69 +233,82 @@ def validate_array(
...
-        :func:`Check <pyvista.core.validation.check.check_has_length>`
-        if the array's length is this value or greater.
+        :func:`Check <pyvista.core._validation.check.check_length>`
+        if the array's length is this value or greater. See note in
+        ``must_have_length`` for details.
 
     must_have_max_length : int, optional
-        :func:`Check <pyvista.core.validation.check.check_has_length>`
-        if the array' length is this value or less.
+        :func:`Check <pyvista.core._validation.check.check_length>`
```
user27182, line 276: All the cross-refs in the docs still need to be reviewed to make sure the links work with no typos

pyvista/core/_validation/validate.py
```diff
@@ -260,33 +436,46 @@ def validate_array(
...
-        except TypeError as e:
-            raise TypeError(f"{name} must be numeric.") from e
-
+        check_real(wrapped(), name=name)
     if must_have_dtype is not None:
-        check_subdtype(arr_out, must_have_dtype, name=name)
+        check_subdtype(wrapped(), base_dtype=must_have_dtype, name=name)
+
+    if dtype_out not in (None, float, int, bool):
+        return_type = "numpy"
```
user27182, line 451: should probably raise an error here if `dtype` is specified as a numpy type and if `return_type` is not set to `numpy`

pyvista/core/_validation/validate.py
```diff
@@ -221,28 +392,33 @@ def validate_array(
...
         validation checks fail.
 
     Returns
     -------
-    array_like
+    float | int | bool | Array[float] | Array[int] | Array[bool]
         Validated array. Returned object is:
 
         * an instance of ``np.ndarray`` (default), or
         * a nested ``list`` (if ``to_list=True``), or
```
user27182, line 419: Text needs to be updated to match new generic ArrayLike[T] -> ArrayLike[T] behaviour.

Also should add a new `shape_changed` or `return_reshaped_flag` (or similar) kwarg.
Initially suggested by @darikg to provide similar functionality provided by `_coerce_pointslike_arg` and document the return behaviour here.

pyvista/core/_validation/_array_wrapper.py
```diff
@@ -0,0 +1,369 @@
...
+            raise ValueError(f"The following array is not valid:\n\t{reprlib.repr(_array)}")
+
+    def __getattr__(self, item):
+        try:
+            return self.__getattribute__(item)
+        except AttributeError:
+            return getattr(self._array, item)
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}({self._array.__repr__()})'
```
user27182, line 125: A few lines still missing test coverage (like this one)

pyvista/core/_validation/check.py
```diff
@@ -10,36 +10,69 @@
...
+    name: str = 'Input',
+):
+    """Check if an input's data-type is a subtype of another data-type(s).
 
     Parameters
     ----------
-    arg1 : dtype_like | array_like
-        ``dtype`` or object coercible to one. If ``array_like``, the dtype
-        of the array is used.
+    input_obj : DType | Array | Scalar
```
user27182, line 60: The docstring arg types for arrays, scalars, dtypes, shapes, etc. of all `validate` and `check` functions need to be reviewed and made consistent. Since the function signatures are annotated with actual type hints, the docstring types don't necessarily need to be exact, e.g. can use `ArrayLike`, `array-like`, or `Array[float]`... I'm not sure what the convention should be.

Either way, help with reviewing these to make them fairly consistent across the package would be appreciated.

pyvista/core/_validation/check.py
```diff
@@ -118,27 +158,28 @@ def check_real(arr, /, *, name="Array"):
...
-        check_subdtype(arr, (np.floating, np.integer), name=name)
+        check_subdtype(dtype, (np.floating, np.integer), name=name)
     except TypeError as e:
         raise TypeError(f"{name} must have real numbers.") from e
 
 
-def check_sorted(arr, /, *, ascending=True, strict=False, axis=-1, name="Array"):
+def check_sorted(
+    array: _ArrayLikeOrScalar[_NumberType],
+    /,
```
user27182, line 170: We could drop the `\` from all functions, there's no technical need to have it like this (I think its use stems from compatibility with C). Just having `*` for kwargs only is OK I think.

The [example from scientific python](https://learn.scientific-python.org/development/principles/design/#complexity-is-always-conserved) has `kwargs` only as a recommendation.

user27182 (commented):
I've added a few of my own thoughts, possible changes for review. Feel free to comment and/or suggest changes

pyvista/core/_typing_core/_type_guards.py
```diff
@@ -0,0 +1,47 @@
...
+        and all(_is_NumberSequence(subarray, N=N) for subarray in array)
+    ):
+        # We have the correct type, now check all subarray shapes are equal
+        array = cast(Sequence[Sequence[_NumberType]], array)
+        sub_shape = len(array[0])
+        return all(len(sub_array) == sub_shape for sub_array in array)
+    return False
+
+
+def _has_element_types(array: Iterable[Any], types: Tuple[Type[Any], ...], N=None) -> bool:
```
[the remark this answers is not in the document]
user27182: This unused param is removed in 67d34a352a9ab151b64abca00c7f328a60bd9f17

pyvista/core/_typing_core/_type_guards.py
```diff
@@ -0,0 +1,47 @@
...
+from itertools import islice
+from typing import Any, Iterable, Sequence, Tuple, Type, TypeVar, cast
+
+import numpy as np
+
+from ._array_like import _ArrayLikeOrScalar, _NumberType
+
+_BuiltinNumberType = TypeVar('_BuiltinNumberType', float, int, bool, covariant=True)
+
+# TODO: import and return TypeGuard instead of bool for python >= 3.10
```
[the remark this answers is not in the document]
user27182: TypeGuard is added in 67d34a352a9ab151b64abca00c7f328a60bd9f17

pyvista/core/_validation/check.py
user27182: > consider changing to: `check_contains(container: Any, *, contains: Any, name: str = 'Input')`

This order (container, item) is consistent with python's [operator.contains](https://docs.python.org/3/library/operator.html#operator.__contains__) function API.

The API has been updated to `check_contains(container: Any, /, *, must_contain: Any, name: str = 'Input')` in 9ca27dfab49c230e665fc2a4318408a0fa73acf1

pyvista/core/_validation/check.py
darikg: > The docstring arg types for arrays, scalars, dtypes, shapes, etc. of all validate and check functions need to be reviewed and made consistent. Since the function signatures are annotated with actual type hints, the docstring types don't necessarily need to be exact, e.g. can use ArrayLike, array-like, or Array[float]... I'm not sure what the convention should be.

That's something I've struggled with before. Really it's kinda silly to have to duplicate the types from the signature to the docstr, since the signature is actually going to be checked by mypy. I wish numpydoc could just do it automatically.

user27182: With #5621 I'm thinking that the annotations and docstrings should match, e.g. `Array[float]` in both annotation and docstrings. Since #5621 makes the aliases linkable, at least this way it will be clear.

I can see use cases where we may not necessarily want these to match, e.g. the annotations are complex but we want to simplify the docstring for readability. 

While working on #5285 I also found that `pandas` does this somewhat, where their docs would say `npt.NDArray[float]` but the annotations would say `npt.NDArray[np.float64]` since mypy would complain otherwise about using `float` for a numpy array (which is exactly the issue that #5285 fixed for us here).

I believe there may also be options within sphinx to display types in annotations-only, docstrings-only, or both. Not sure though

pyvista/core/_validation/validate.py
user27182: This has been added in 5cf5deb1b338b50c56e1e735533079e5883b62ce

tests/core/test_validation_reveal_type.txt
user27182: updated as .py file in https://github.com/pyvista/pyvista/pull/5571/commits/e81d3774ddba8c85f77b5d8170a7b45694ea2a7e

user27182:
> * [_coerce_pointslike_args](https://github.com/pyvista/pyvista/blob/ce6d08cbcbd6d525ea3eafe27f4e0b016fec204a/pyvista/core/utilities/arrays.py#L59) could now be replaced by something like `validate_arrayNx3(points, must_be_real=True, reshape=True, return_type='numpy')`, _except_ that it has an extra return argument to disambiguate whether the original array was shape (3,) or (n, 3). Is the solution just an extra wrapper function?

a new param, `get_flags` is added in https://github.com/pyvista/pyvista/pull/5571/commits/fd8e90544e19f72c55939386c8a69e3dadd94924, so the call would be something like:

``` python
array, flags = validate_arrayNx3(points, must_be_real=True, reshape=True, return_type='numpy', get_flags=True)
is_reshaped = not flags.same_shape
```

user27182:
> * `validate_axes(must_be_orthogonal=True)` seems like it needs a tolerance value to allow vectors to be orthogonal within some precision

I reviewed the `validate_axes` code and it already uses `isclose` and `allclose` for checking orthogonality. I added a non-identity  test case in https://github.com/pyvista/pyvista/pull/5571/commits/9175c7ee792311998e710d811a84a7a26c4e9bc2 to be sure. So, I think this comment is resolved.

tests/core/test_validation.py
user27182: The typing tests have been updated extensively. I think the current test suite works great, no need to use `pytest-mypy-plugins`

pyvista/core/_validation/validate.py
user27182: Pending, see #5621

pyvista/core/_validation/_array_wrapper.py
user27182: Test coverage is now nearly 100%

pyvista/core/_validation/check.py
user27182: I think the current use of `/` and `*` is fine.

pyvista/core/_validation/validate.py
user27182: The returns section has been uddated, and a new `get_flags` param has been added with info about the change in shape

CONTRIBUTING.rst
```diff
@@ -450,6 +450,67 @@ encouraged to use the ``.. versionadded`` sphinx directive. For example:
         """
 
 
+Input Validation
```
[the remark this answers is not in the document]
user27182: Moved as part of #5664

pyvista/core/_validation/check.py
user27182: It's probably fine as is, can be changed later if needed.

pyvista/core/_validation/validate.py
user27182: The docs can be checked as part of #5664

pyvista/core/_validation/check.py
user27182: This issue can be dealt with separately as part of #5665 I think

user27182:
I believe all issues are now resolved. A few outstanding issues with the docs can be dealt with separately, I think, via:
- #5621 
- #5664 
- #5665 

Otherwise, this PR LGTM. Ready for additional review.

darikg:
Haven't done anything thorough, but started poking at `validate_arrayNx3`. First, if I pass in `list[list[number]]` I get `list[list[number]]` back (unless I specify return_type='numpy'). Is that expected? It's typed as returning `NumpyArray[_NumberType]` with no overloads.

Second, I'm concerned about `_is_NumberSequence` and `_is_NestedNumberSequence`. 

```python
from pyvista.core._validation import validate_arrayNx3
from pycallgraph2 import PyCallGraph
from pycallgraph2.config import Config

arr = [[0, 0, 0] for _ in range(10_000)]
g = PyCallGraph(config=Config(output='graphviz'))
g.start()
val, flags = validate_arrayNx3(
    arr,
    reshape=True,
    must_be_real=True,
    get_flags=True,
    return_type='numpy',
)
g.done()
```

pycallgraph

When the requested output type is numpy, it should be possible to cast to numpy immediately and not bother with the linear scans of the array. I suppose if the output is list/tuple, it might be okay to assume the user is okay with slow validations.

user27182:
> Haven't done anything thorough, but started poking at `validate_arrayNx3`. First, if I pass in `list[list[number]]` I get `list[list[number]]` back (unless I specify return_type='numpy'). Is that expected? It's typed as returning `NumpyArray[_NumberType]` with no overloads.

The typing has not been updated for functions which call `validate_array` internally. So, generic `ArrayLike[T]` ->  `ArrayLike[T]` for `validate_arrayNx3` is technically expected (the type hint is wrong). But, we can also make `to_numpy` the default for this function. This could also be the default for other functions. As you may have seen from the large number of overloads for `validate_array`, it takes a lot of overloads (and associated testing) to support the type hints generically. What do you think?

> Second, I'm concerned about `_is_NumberSequence` and `_is_NestedNumberSequence`.
> When the requested output type is numpy, it should be possible to cast to numpy immediately and not bother with the linear scans of the array. I suppose if the output is list/tuple, it might be okay to assume the user is okay with slow validations.

Cool graph! Never heard of `PyCallGraph`. We can probably pass the `return_type` param to `_ArrayLikeWrapper.__new__` to bypass these checks. To speed up the `_is_NumberSequence` checks I previously had an `N` parameter to limit the checks to `N` values in the array. This was removed in 67d34a352a9ab151b64abca00c7f328a60bd9f17, but could be added back. Or, keep it as-is and add a note to the docs to recommend using `to_numpy` to speed things up. Any preference, or other ideas?

darikg:
> The typing has not been updated for functions which call validate_array internally. So, generic ArrayLike[T] -> ArrayLike[T] for validate_arrayNx3 is technically expected (the type hint is wrong). But, we can also make to_numpy the default for this function. This could also be the default for other functions. As you may have seen from the large number of overloads for validate_array, it takes a lot of overloads (and associated testing) to support the type hints generically. What do you think?

haha yeah I appreciate the effort that went into all of the overloads there and agree it's not worth doing it for all of the specialized `validate_arrayXXX` versions. I think it makes total sense for them to be forced return_type=numpy, since it's always possible to drop down to the bare `validate_array` call if a different return type is required. (I briefly considered but discarded as completely insane the idea of converting the validation functions to classes implementing `__call__` and trying to do all the overloading once in the superclass)

> We can probably pass the return_type param to _ArrayLikeWrapper.__new__ to bypass these checks.

Perfect!

> To speed up the _is_NumberSequence checks I previously had an N parameter to limit the checks to N values in the array. This was removed in https://github.com/pyvista/pyvista/commit/67d34a352a9ab151b64abca00c7f328a60bd9f17, but could be added back. Or, keep it as-is and add a note to the docs to recommend using to_numpy to speed things up. Any preference, or other ideas?

As long as there's a fast escape hatch for numpy outputs I think either choice here is perfectly reasonable.

user27182:
> it's not worth doing it for all of the specialized `validate_arrayXXX` versions. I think it makes total sense for them to be forced return_type=numpy, since it's always possible to drop down to the bare `validate_array` call if a different return type is required. 

Sounds good. I'll update the other validate functions. `validate_number` has already been updated but I'll need some time to update the others. 

> (I briefly considered but discarded as completely insane the idea of converting the validation functions to classes implementing `__call__` and trying to do all the overloading once in the superclass)

I like this, and I actually attempted this briefly yesterday, but still ran into the same overloading issue we have now: mypy always requires explicit annotations for every function. Annoyingly, this means that adding a simple wrapper completely removes all previous annotations, e.g: `def wrapper(array): return validate_array(array)` will result in `reveal_type` reporting `Any` even though the function does nothing at all. So, using `__call__` instead as a wrapper function and calling `validate_array` internally still mean that we would lose the type hints from all the overloads. I couldn't find a way around this... 

> > We can probably pass the return_type param to _ArrayLikeWrapper.__new__ to bypass these checks.


Bypass added in 12fd90baed0a044f6453224ffe5fcee9e5f5a480

user27182:
> > it's not worth doing it for all of the specialized `validate_arrayXXX` versions. I think it makes total sense for them to be forced return_type=numpy, since it's always possible to drop down to the bare `validate_array` call if a different return type is required.
> 
> Sounds good. I'll update the other validate functions.

The docstrings and type hints have all been updated with the most recent commit

user27182:
pre-commit.ci autofix

tkoyama010:
@pyvista-bot preview

tests/core/typing/test_validation_typing.py
```diff
@@ -0,0 +1,221 @@
...
+the format:
+
+    reveal_type(arg)  # EXPECTED_TYPE: "<T>"
+
+where `arg` is any argument you want mypy to analyze, and <T> is the expected
+revealed type returned by mypy. Note: the output types from mypy are truncated
+with the module names removed, e.g. `typing.Sequence` -> `Sequence`,
+`builtins.float` -> `float`, etc.
+
+"""
```
tkoyama010, line 14: It took me a while, but I finally figured out how to create this great test. :)
I recommend adding a section to [Contributing](https://docs.pyvista.org/version/stable/extras/developer_notes) to understand how to add tests.

tkoyama010 (changes requested):

tkoyama010:
pre-commit.ci autofix

pyvista/core/utilities/misc.py
```diff
@@ -6,15 +6,16 @@
...
 import threading
 import traceback
-from typing import TYPE_CHECKING, Type, TypeVar, Union
+from typing import TYPE_CHECKING, Any, Dict, Type, TypeVar, Union
 import warnings
 
 import numpy as np
 
-if TYPE_CHECKING:  # pragma: no cover
+if TYPE_CHECKING:
```
tkoyama010, line 18:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

tests/core/typing/test_validation_typing.py
user27182: This should be resolved with https://github.com/pyvista/pyvista/pull/5571/commits/88e4aff93a1c415a3f5cec9aa0d259cd4ec3ea07

user27182:
pre-commit.ci autofix

pyvista/core/utilities/misc.py
```diff
@@ -311,3 +312,16 @@ def _reciprocal(x, tol=1e-8):
     x[~zero] = np.reciprocal(x[~zero])
     x[zero] = 0
     return x
+
```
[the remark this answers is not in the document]
tkoyama010: If we are not using it, let's delete it. We can later search the repository for the history of this implementation.

tkoyama010:
@pyvista-bot preview

CONTRIBUTING.rst
```diff
@@ -639,6 +639,81 @@ For example, the following writes all images generated by ``pytest`` to
...
+
+#. Regular unit tests in ``tests/core/test_validation.py``
+#. Customized unit tests in ``tests/core/typing`` for testing type hints
+
+The custom unit tests check that the type hints for the validation package are
+correct both statically and dynamically. This is mainly used to check complex and
+overloaded function signatures, such as the type hints for ``validate_array``
+or related functions.
+
+Individual test cases are written as a single line of python code with the format:
```
tkoyama010, line 655:
```suggestion
Individual test cases are written as a single line of Python code with the format:
```

CONTRIBUTING.rst
```diff
@@ -639,6 +639,81 @@ For example, the following writes all images generated by ``pytest`` to
...
+
+    reveal_type(arg)  # EXPECTED_TYPE: "<T>"
+
+where ``arg`` is any argument you want mypy to analyze, and ``"<T>"`` is the
+expected revealed type returned by ``Mypy``.
+
+For example, the ``validate_array`` function by default returns a list of floats
+when a list of floats is provided at the input. The type hint should reflect this.
+To test this, we can write a test case as for the function call ``validate_array([1.0])``
+as follows:
```
tkoyama010, line 667:
```suggestion
For example, the ``validate_array`` function, by default, returns a list of floats
when a list of floats is provided at the input. The type hint should reflect this.
To test this, we can write a test case for the function call ``validate_array([1.0])``
as follows:
```

CONTRIBUTING.rst
```diff
@@ -639,6 +639,81 @@ For example, the following writes all images generated by ``pytest`` to
...
+.. code:: python
+
+    "builtins.list[builtins.float]"
+
+Notice that the revealed type is fully qualified, i.e. includes ``builtins``. For
+brevity, the custom test suite omits this, and requires that only ``list`` be
+included in the expected type. Therefore, for this test case the ``EXPECTED_TYPE``
+type is ``"list[float]"``, not ``"builtins.list[builtins.float]"``. (Similarly the
+package name ``numpy`` should also be omitted for tests where a ``numpy.ndarray`` is
+expected.)
```
tkoyama010, line 693:
```suggestion
Notice that the revealed type is fully qualified, i.e. includes ``builtins``. For
brevity, the custom test suite omits this and requires that only ``list`` be
included in the expected type. Therefore, for this test case, the ``EXPECTED_TYPE``
type is ``"list[float]"``, not ``"builtins.list[builtins.float]"``. (Similarly, the
package name ``numpy`` should also be omitted for tests where a ``numpy.ndarray`` is
expected.)
```

CONTRIBUTING.rst
```diff
@@ -639,6 +639,81 @@ For example, the following writes all images generated by ``pytest`` to
...
+
+The tests can be executed with:
+
+.. code:: python
+
+    pytest tests/core/typing
+
+When executed, a single instance of ``Mypy`` will statically analyze all of the
+test cases. The actual revealed types by ``Mypy`` are compared against the
+``EXPECTED_TYPE`` defined by each test case.
```
tkoyama010, line 707:
```suggestion
When executed, a single instance of ``Mypy`` will statistically analyze all the
test cases. The actual revealed types by ``Mypy`` are compared against the
``EXPECTED_TYPE``is defined by each test case.
```

CONTRIBUTING.rst
```diff
@@ -639,6 +639,81 @@ For example, the following writes all images generated by ``pytest`` to
...
+
+In addition, the ``pyanalyze`` package is used to test the actual returned
+type at runtime matches the statically-revealed type. The
+`pyanalyze.runtime.get_compatibility_error <https://pyanalyze.readthedocs.io/en/latest/reference/runtime.html#pyanalyze.runtime.get_compatibility_error>`_
+method is used for this. If new typing test cases are added for a new
+validation function, the new function will need to be added to the list of
+imports in ``tests/core/typing/test_validation_typing.py`` so that the
+runtime test is able to call the function.
```
tkoyama010, line 715:
```suggestion
In addition, the ``pyanalyze`` package tests the actual returned
type at runtime to match the statically-revealed type. The
`pyanalyze.runtime.get_compatibility_error <https://pyanalyze.readthedocs.io/en/latest/reference/runtime.html#pyanalyze.runtime.get_compatibility_error>`_
method is used for this. If new typing test cases are added for a new
validation function, the new function must be added to the list of
imports in ``tests/core/typing/test_validation_typing.py`` so that the
runtime test can call the function.
```

tkoyama010 (changes requested):

tkoyama010:
This title correction was incorrect as rename part has already been merged :(

tkoyama010:
Please wait as we merge them little by little as I understand them :bow:

pyvista/core/utilities/misc.py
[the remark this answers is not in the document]
user27182: Removed in https://github.com/pyvista/pyvista/pull/5571/commits/29cad6977fa53a28c0be816d88802bcec1830f2b

pyvista/core/_typing_core/_array_like.py
```diff
@@ -21,29 +21,65 @@
...
+import typing
+from typing import TYPE_CHECKING, Any, List, Sequence, Tuple, Type, TypeVar, Union
 
 import numpy as np
 import numpy.typing as npt
 
-# Create alias of npt.NDArray bound to numeric types only
-NumberType = TypeVar('NumberType', bool, int, float, np.bool_, np.int_, np.float64, np.uint8)
+# Define numeric types
+# TODO: remove # type: ignore once support for 3.8 is dropped
```
tkoyama010, line 32:
```suggestion
# TODO: remove # type: ignore once support for Python3.8 is dropped
```

pyvista/core/_typing_core/_array_like.py
```diff
@@ -21,29 +21,65 @@
...
+
+
 _T = TypeVar('_T')
-_FiniteNestedSequence = Union[  # Note: scalar types are excluded
-    Sequence[_T],
-    Sequence[Sequence[_T]],
-    Sequence[Sequence[Sequence[_T]]],
-    Sequence[Sequence[Sequence[Sequence[_T]]]],
+if not TYPE_CHECKING and sys.version_info < (3, 9, 0):
+    # TODO: Remove this conditional block once support for 3.8 is dropped
```
tkoyama010, line 51:
```suggestion
    # TODO: Remove this conditional block once support for Python3.8 is dropped
```

pyvista/core/dataset.py
```diff
@@ -1416,7 +1416,7 @@ def flip_z(
         """
         if point is None:
             point = self.center
-        check_valid_vector(point, 'point')
+        check_valid_vector(cast(VectorLike[float], point), 'point')
```
tkoyama010, line 1419: @user27182 Is there any reason why this cast is needed? If needed, there are other places in this file where a cast is needed as well.

user27182: I'll have to look into it. These changes were not directly related to this PR IIRC. If it's redundant this mypy option should fix this kind of thing

https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_redundant_casts

user27182: The cast is not needed. Both casts added for this file were removed in https://github.com/pyvista/pyvista/pull/5571/commits/9420512fa39a38203a166f7f0944ebffb54db92e.

user27182:
@tkoyama010 if you want to break this PR up further, the typing overloads and all the typing tests could be removed from here and made into a separate PR, e.g. "Add typing overloads for validation package"

pyvista/core/_validation/__init__.py
```diff
@@ -2,32 +2,33 @@
...
+from .validate import validate_array  # noqa: F401  # noqa: F401
+from .validate import validate_array3  # noqa: F401  # noqa: F401
+from .validate import validate_arrayN  # noqa: F401  # noqa: F401
+from .validate import validate_arrayN_unsigned  # noqa: F401  # noqa: F401
+from .validate import validate_arrayNx3  # noqa: F401  # noqa: F401
+from .validate import validate_axes  # noqa: F401  # noqa: F401
+from .validate import validate_data_range  # noqa: F401  # noqa: F401
+from .validate import validate_number  # noqa: F401  # noqa: F401
+from .validate import validate_transform3x3  # noqa: F401  # noqa: F401
+from .validate import validate_transform4x4  # noqa: F401  # noqa: F401
```
tkoyama010, line 34: I recommend to summarize.
```suggestion
# flake8: noqa: F401
from .check import check_contains
from .check import check_finite
from .check import check_greater_than
from .check import check_instance
from .check import check_integer
from .check import check_iterable
from .check import check_iterable_items
from .check import check_length
from .check import check_less_than
from .check import check_ndim
from .check import check_nonnegative
from .check import check_number
from .check import check_range
from .check import check_real
from .check import check_sequence
from .check import check_shape
from .check import check_sorted
from .check import check_string
from .check import check_subdtype
from .check import check_type
from .validate import validate_array
from .validate import validate_array3
from .validate import validate_arrayN
from .validate import validate_arrayN_unsigned
from .validate import validate_arrayNx3
from .validate import validate_axes
from .validate import validate_data_range
from .validate import validate_number
from .validate import validate_transform3x3
from .validate import validate_transform4x4
```

user27182:
I might try making a separate PR to further reduce the scope of this one. It might include all updates to `validation` _except_ for the changes that make `validate_array` generic, i.e. where the original behavior of always returning a numpy array is preserved.

user27182:
Unmerged changes are tracked in #6751 instead.

PR #5573, opened 2024-02-03
Move array casting functions into `validation` package

user27182:
Related to #5567 and #5571. 

The following functions are renamed:
- `cast_to_ndarray` -> `_cast_to_numpy`
- `cast_to_tuple_array` -> `_cast_to_tuple`
- `cast_to_list_array` -> `_cast_to_list`

Dependency on the `core.utilities.arrays` module is removed to avoid circular imports:

- The above functions are moved inside the `validation` package for private use. Using `validate_array` directly is preferred,  see https://github.com/pyvista/pyvista/pull/5567#issuecomment-1925357548.
- A localized version of `array_from_vtkmatrix` is added

A `must_be_real` kwarg is also added to `_cast_to_numpy` to check the `dtype`  is real numbers. Inputs should be real if the array will be used by vtk. See https://github.com/pyvista/pyvista/pull/5567#issuecomment-1925098974.


tkoyama010 (dismissed):
I agree with moving these functions to the validation package.
Note that since these functions are not published in the documentation, I have decided not to respond to deprecation notices due to API changes.

darikg:
There's another circular import I found I'd like to figure out here

`core.validation.validate` imports `pyvista.core.utilities.arrays.array_from_vtkmatrix`
`core.utilities.__init__` imports from most of its subpackages to re-export them. 

This means that e.g. `core.utilities.geometric_objects` can't import any of the validation methods.

So I think either `array_from_vtkmatrix` needs to move into the validation package, which seems weird, or `validate_transform3x3` and `validate_transform4x4` need to move outside of the validation package, which also seems weird.

user27182:
> So I think either `array_from_vtkmatrix` needs to move into the validation package, which seems weird, or `validate_transform3x3` and `validate_transform4x4` need to move outside of the validation package, which also seems weird.

We could also import `array_from_vtkmatrix` locally inside of `validate_transform3x3` and `validate_transform4x4` instead of at the package level. 

Or, add (copy) the key bits of code from those `array_from_vtkmatrix` functions into a standalone lightweight version of the function. Do we maybe want `_cast_to_numpy` to handle vtkMatrix objects too? And then just call  `_cast_to_numpy` instead of importing and using `array_from_vtkmatrix`?

darikg:
> Or, add (copy) the key bits of code from those array_from_vtkmatrix functions into a standalone lightweight version of the function. 

The array_from_vtkmatrix function is so simple I think this might be the way to go.

> Do we maybe want _cast_to_numpy to handle vtkMatrix objects too? And then just call _cast_to_numpy instead of importing and using array_from_vtkmatrix?

Long term I could see this validation package being widely useful outside of pyvista, it might be better to keep the vtk stuff isolated.

tkoyama010 (dismissed):

tkoyama010:
@darikg Could we get your approval?

pyvista/core/validation/validate.py
```diff
@@ -554,6 +556,17 @@ def validate_transform3x3(transform, /, *, name="Transform"):
     return arr
 
 
+def _array_from_vtkmatrix(
+    matrix: Union[vtkMatrix3x3, vtkMatrix4x4],
+    shape: Union[Tuple[Literal[3], Literal[3]], Tuple[Literal[4], Literal[4]]],
```
darikg, line 561: I really like this `shape` typing

tests/core/test_validation.py
```diff
@@ -955,3 +956,48 @@ def test_validate_axes_orthogonal(bias_index):
     assert np.array_equal(axes, axes_left)
     with pytest.raises(ValueError, match=msg):
         validate_axes(axes_left, must_be_orthogonal=True)
+
+
+@pytest.mark.parametrize('as_any', [True, False])
+@pytest.mark.parametrize('copy', [True, False])
+@pytest.mark.parametrize('dtype', [None, float])
+def test__cast_to_numpy(as_any, copy, dtype):
```
darikg, line 964:
```suggestion
def test_cast_to_numpy(as_any, copy, dtype):
```

tests/core/test_validation.py
```diff
@@ -955,3 +956,48 @@ def test_validate_axes_orthogonal(bias_index):
...
+        assert array_out.dtype.type is array_in.dtype.type
+    else:
+        assert array_out.dtype.type is np.dtype(dtype).type
+
+
+def test_cast_to_numpy_raises():
+    msg = "Input cannot be cast as <class 'numpy.ndarray'>."
+    with pytest.raises(ValueError, match=msg):
+        _cast_to_numpy([[1], [2, 3]])
+
```
darikg, line 986:
```suggestion
def test_cast_to_numpy_must_be_real():
    _ = _cast_to_numpy([0, 1], must_be_real=True)

    with pytest.raises(ValueError):
        _ = _cast_to_numpy([0, 1 + 1j], must_be_real=True)


```

tests/core/test_validation.py
```diff
@@ -955,3 +956,48 @@ def test_validate_axes_orthogonal(bias_index):
...
+    with pytest.raises(ValueError):
+        _cast_to_tuple([[1, [2, 3]]])
+
+
+def test_cast_to_list():
+    array_in = np.zeros(shape=(3, 4, 5))
+    array_list = _cast_to_list(array_in)
+    assert np.array_equal(array_in, array_list)
+    with pytest.raises(ValueError):
+        _cast_to_tuple([[1, [2, 3]]])
```
darikg, line 1003:
```suggestion

@pytest.mark.parametrize(['cls', 'shape'], [
    (vtkMatrix3x3, (3, 3)),
    (vtkMatrix4x4, (4, 4)),
])
def test_vtkmatrix_from_array(cls, shape):
    orig = np.random.default_rng().random(shape)
    mat = cls()
    for i, j in itertools.product(range(shape[0]), range(shape[1])):
        mat.SetElement(i, j, orig[i, j])

    assert np.array_equal(orig, _array_from_vtkmatrix(orig, shape=shape))
```

tests/core/test_validation.py
```diff
@@ -45,6 +45,7 @@
     validate_transform3x3,
     validate_transform4x4,
 )
+from pyvista.core.validation._cast_array import _cast_to_list, _cast_to_numpy, _cast_to_tuple
 from pyvista.core.validation.check import _validate_shape_value
 from pyvista.core.validation.validate import _set_default_kwarg_mandatory
```
darikg, line 51:
```suggestion
from pyvista.core._vtk_core import vtkMatrix3x3, vtkMatrix4x4
```

darikg (changes requested):
Suggested two extra tests and one type fix but lgtm

tests/core/test_validation.py
```diff
@@ -955,3 +958,80 @@ def test_validate_axes_orthogonal(bias_index):
...
+    with pytest.raises(ValueError, match=msg):
+        _cast_to_numpy([[1], [2, 3]])
+
+
+def test_cast_to_numpy_must_be_real():
+    _ = _cast_to_numpy([0, 1], must_be_real=True)
+    _ = _cast_to_numpy("abc", must_be_real=False)
+
+    msg = "Array must have real numbers. Got dtype <class 'numpy.complex128'>"
+    with pytest.raises(TypeError, match=msg):
```
user27182, line 995: Should dtype errors raise `ValueError` or `TypeError` ?

user27182:
@darikg this should be ready for review again

pyvista/core/validation/_cast_array.py
```diff
@@ -0,0 +1,116 @@
...
+
+    must_be_real : bool, default: True
+        Raise a TypeError if the array does not have real numbers, i.e.
+        its data type is not integer or floating.
+
+    Raises
+    ------
+    ValueError
+        If input cannot be cast as a NumPy ndarray.
+        ValueError
```
darikg, line 82: `ValueError` duplicated here

tests/core/test_validation.py
darikg: Didn't think about that. Either seems justifiable. I guess I have a slight preference for TypeError.

user27182: Type of object vs type of the object's values... not always clear. But yea here I think TypeError is appropriate for wrong dtype.

pyvista/core/validation/_cast_array.py
user27182: I see, that part of the docs is from the original function. This emulates NumPy's behaviour for ragged arrays, e.g. a ValueError is raised for `np.array([[1,2],[4,5,6]])`.

Just below the reviewed code block is new docs added for the TypeError generated if `must_be_real=True`.

But, the original function also forbids object dtype and also raises a ValueError there.

I think any dtype errors, either object or not real numbers, should be TypeErrors.

darikg: I just mean the term `ValueError` is listed twice instead of once.

> I think any dtype errors, either object or not real numbers, should be TypeErrors.

Yeah, that makes sense to me.

user27182: Oh I see, sorry I thought it was related to your other comment. I'll fix that too.

user27182:
LGTM now

darikg:
ditto!

tkoyama010 (approved):

PR #5575, opened 2024-02-04
Add `pv.OCCTReader` to read step file using OCC

tkoyama010:
### Overview

Add `pv.OCCTReader` to read the step file using OCC.

### Details

- None


tkoyama010:
Seems that we need to build OCC: https://discourse.vtk.org/t/difficulty-accessing-the-vtkocctreader-through-the-python-modules/12588/2

MatthewFlamm:
This reader also supports IGES.  Do we need to handle the file type in the constructor of the reader?  In particular for the `SetFileFormat` method, does it need to be set before we read the file information, which we do when instantiating the reader in pyvista?

MatthewFlamm:
> Seems that we need to build OCC: https://discourse.vtk.org/t/difficulty-accessing-the-vtkocctreader-through-the-python-modules/12588/2

Missed this, I think this will be hard to support here.  It seems like a good add-on package to pyvista if it requires a special vtk build.

PR #5580, opened 2024-02-05
Fix uint8 error for numpy2

MatthewFlamm:
### Overview

To resolve error from #5450

```txt
    def test_check_is_real():
        check_is_real(1)
        check_is_real(-2.0)
>       check_is_real(np.array(-2.0, dtype="uint8"))
E       OverflowError: Python integer -2 out of bounds for uint8
```


user27182 (approved):
Good catch, LGTM

PR #5583, opened 2024-02-06
Remove unused notes of XdmfReader class about time

tkoyama010:
### Overview

We don't need notes about time value in `XdmfReader` so we defined the time value in #5332.

### Details

- None


MatthewFlamm (approved):
LGTM

PR #5585, opened 2024-02-07
Add `flake8-return` check using ruff

tkoyama010:
### Overview

See https://docs.astral.sh/ruff/rules/#flake8-return-ret.

### Details

- None


tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

PR #5586, opened 2024-02-07
Rename, remove, rework `validation` functions

user27182:
Part of #5571.
- Simplify function names, e.g. `check_has_shape` -> `check_shape`
- Rework `check_string_in_iterable` as `check_contains` for general use with any input type and any container.
- Remove functions `check_is_dytpe`, `validate_dytpe` and `check_is_arraylike` since their use cases are very limited and type errors are already handled by NumPy.
- Remove `check_iterable_of_strings` the more general `check_iterable_items` function should be used instead.
- Remove `check_scalar` to discourage use of 0-dimensional numpy arrays; `check_number` and/or `check_shape` (or `check_ndim` from #5571) should be used instead.
- Update package import and make it private: `pyvista.core.validation` -> `pyvista._validation`


pyvista/core/validation/check.py
```diff
@@ -967,90 +890,48 @@ def check_is_iterable_of_some_type(
...
-    """
-    try:
-        check_is_iterable_of_some_type(iterable_obj, str, allow_subclass=allow_subclass, name=name)
     except TypeError:
         raise
 
 
-def check_is_string_in_iterable(string_in, /, string_iterable, *, name='String'):
+def check_contains(obj, /, container, *, name='Input'):
     """Check if a given string is in an iterable of strings.
```
darikg, line 924: Outdated docstr. `Check if a given object is in an iterable of objects' I guess?

user27182: Thanks, fixed by 4db3a488

pyvista/core/validation/check.py
```diff
@@ -967,90 +890,48 @@ def check_is_iterable_of_some_type(
...
-    """
-    try:
-        check_is_iterable_of_some_type(iterable_obj, str, allow_subclass=allow_subclass, name=name)
     except TypeError:
         raise
 
 
-def check_is_string_in_iterable(string_in, /, string_iterable, *, name='String'):
-    """Check if a given string is in an iterable of strings.
+def check_contains(item, /, container, *, name='Input'):
```
user27182, line 923: Should the order of `item` and `container` be swapped here? The current order is consisntent with `item in containter` syntax, but every other `check` or `validate` function has the "object being tested" as the first arg. So maybe `containter` should be the first arg here?

user27182:
The changes here include most if not all of the breaking changes I wanted to implement as part of #5571. The rest of that PR is mostly an implementation rework and adds support for typing. (The main API changes from there that we could also try to incorporate here is possibly the updates to `check_number`, and the addition of `check_ndim` from #5571.)

So, we could possibly make the docs public as part of this PR if we want, and finally 'release' the package for use.

user27182:
Here's the diff between #5571 and this PR.:
https://github.com/user27182/pyvista/compare/maint/simplify_validation..maint/update_validation
Not sure if there are any changes there worth integrating here first if we do want to publish the package now.

user27182:
> The changes here include most if not all of the breaking changes I wanted to implement as part of #5571. 

I have introduced a new breaking API change in #5571 to merge the `validate_array` kwargs `to_tuple`, `to_list`, into a single parameter `output_type` (a new `to_numpy` option is also consolidated into this param).

And so, I think we can leave this PR as-is without adding to the docs yet. 

Ready for review.

pyvista/core/validation/check.py
darikg: no strong opinion but if the signature was `def check_contains(*, item, container,  name='String`) the caller would be forced to explicit about which arg is which

darikg:
LGTM. 

> So, we could possibly make the docs public as part of this PR if we want, and finally 'release' the package for use.

My recommendation would be to leave this package undocumented for now until it's more heavily used throughout the codebase so we have more flexibility in changes to the API until we're confident it's stable. Possibly even renaming the package to `_validation`.

tkoyama010 (dismissed):

user27182:
Package name updated, should be good, hopefully checks pass 🤞

tkoyama010 (approved):

PR #5589, opened 2024-02-07
Add `--warn-unused-ignores` to `mypy` config

user27182:
This config option helps catch `# type: ignore` comments which are no longer used or needed. See [mypy docs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-unused-ignores).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,6 +65,7 @@ repos:
     rev: v1.8.0
     hooks:
       - id: mypy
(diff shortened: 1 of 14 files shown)


user27182:
pre-commit.ci autofix

tkoyama010 (dismissed):
Excellent. Automatic deletion of unused items is always important.

tkoyama010 (approved):

PR #5592, opened 2024-02-07
Ignore `flake8` E704 to fix conflict with `black`

### LINKED ISSUE #5591: `black` and `flake8` disagree on code style

Running pre-commit on something like:

``` python
def foo():
    ...
```

will cause `black` to modify it to one-line:
``` python
def foo(): ...
```

after which, `flake8` will complain with `E704 multiple statements on one line (def)`, to which the solution is:
``` python
def foo():
    ...
```

and the cycle continues...

Not sure if this is introduced by #5581 ...

See above.

### System Information

```shell
See pre-commit config from #5581
```

### Screenshots

_No response_

user27182:
Fix #5591.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -20,6 +20,6 @@ ignore =
     # bare excepts (temporary)
     B001, E722
     # we already check black
-    BLK100
+    BLK100,
     # let black auto-format statements on one line
     E704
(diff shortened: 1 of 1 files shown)


tkoyama010 (approved):
Sorry to bother you. We will drop flake8 once ruff is stable, but we are in a transition period right now.

PR #5593, opened 2024-02-08
Add `--disallow-any-generics` to `mypy` config

user27182:
## Overview

Add [`--disallow-any-generics`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-any-generics) option.

In addition, this PR fixes roughly 130 `mypy` errors that were initially generated by using this option.

## Details

This option is a follow-up from #5285. Without this option, using array-like annotations such as `Vector` without specifying a typevar will result in a default assumption of `Vector[Any]` by `mypy`. This is generally not safe since it implies sequences or numpy arrays of strings, complex numbers, or even object types are allowed, when typically only real numbers are expected.

This option forces developers to be explicit and do something like `Vector[float]` instead of `Vector`. In cases where `Vector[Any]` is the correct annotation, this must be made explicit (or use `# type: ignore` ).


tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM. This is a change of annotation, but there are source changes, so we will only wait a day for other reviewers' reactions before merging.

tkoyama010 (dismissed):

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

tkoyama010 (dismissed):

tkoyama010:
Please help us correct the error again.

user27182:
@tkoyama010 should be fixed now 👍

tkoyama010 (approved):

PR #5598, opened 2024-02-11
Add FElupe package as external example of PyVista

tkoyama010:
### Overview

Add the [FElupe](https://github.com/adtzlr/felupe) package by @adtzlr as an external example of PyVista.
image

### Details

- None


tkoyama010:
@pyvista-bot LGTM

PR #5599, opened 2024-02-11
Fix top-level linter settings deprecated in ruff

tkoyama010:
### Overview

Fix top-level linter settings deprecated in ruff.

### Details

- None


tkoyama010:
@pyvista-bot LGTM

PR #5600, opened 2024-02-11
Fix str.join calls can be replaced with f-strings

tkoyama010:
### Overview

Fix str.join calls can be replaced with f-strings.

### Details

- None


tkoyama010:
@pyvista-bot LGTM

PR #5602, opened 2024-02-11
Use iterable unpacking instead of concatenation

tkoyama010:
### Overview

Use iterable unpacking instead of concatenation.

### Details

- See https://docs.astral.sh/ruff/rules/collection-literal-concatenation/


tkoyama010:
@pyvista-bot LGTM

tkoyama010:
@pyvista-bot LGTM

PR #5615, opened 2024-02-13
Fix `CylinderStructured` orientation

user27182:
### Overview

This PR modifies how `CylinderStructured` is oriented, and aligns its orientation behaviour with all other geometric and parametric objects, i.e. it now makes use of `translate`. Previously, it had its own custom version of `translate` for centering and rotating the object. This lead to an unexpected output from the regression images, which was caused by a `RuntimeError` due to a division by zero bug in the code.

It is effectively a fix for the bug shown in the image from here: https://github.com/pyvista/pyvista/pull/5435#issuecomment-1892851901

Note: the base of this branch is set to the branch from #5614, not main.

Compare this branch to that one to see the changes for this PR:
https://github.com/user27182/pyvista/compare/testing/direction_objects..fix/cylinder_orientation


tkoyama010 (approved):
LGTM. I was impressed by how efficiently you tested so many objects.

user27182:
Thanks! This PR includes all the changes from #5614 as well as the bug fix. If you want the diff of the bugfix to show what the regression images looked like before and after the fix, I recommend merging that one first.

tkoyama010 (approved):
Nice!
image

PR #5617, opened 2024-02-13
Use source class and filter in Tube function

tkoyama010:
### Overview

Use source class and filter in the Tube function.

### Details

- None


tkoyama010:
@pyvista-bot LGTM

PR #5618, opened 2024-02-13
Fix UnboundLocalError: cannot access local variable 'iren'

darikg:
Came up while investigating #5604.

Minimal example:

```python
import time
from pyvista import plotting

plotter = plotting.Plotter()

def callback() -> None:
    plotter.iren.clear_events_for_key("j")
    time.sleep(1)
    print("done")
    plotter.add_timer_event(max_steps=1, duration=0, callback=add_callback)

def add_callback(*_):
    plotter.add_key_event(key="j", callback=callback)

add_callback()
plotter.show()
```

Press j to trigger:

> iren.DestroyTimer(self.id)
^^^^
UnboundLocalError: cannot access local variable 'iren' where it is not associated with a value
  warnings.warn(formatted_exception)

I don't understand why this wasn't caught by the add_timer_event tests -- not sure they work


tkoyama010:
Do we need test code?

darikg:
If I run the `test_add_timer_event` code as a standalone file:

```python
import pyvista as pv

sphere = pv.Sphere()

pl = pv.Plotter()
actor = pl.add_mesh(sphere)


def callback(step):
    actor.position = [step / 100.0, step / 100.0, 0]
    print(step)

pl.add_timer_event(max_steps=200, duration=500, callback=callback)

cpos = [(0.0, 0.0, 10.0), (0.0, 0.0, 0.0), (0.0, 1.0, 0.0)]
pl.show(cpos=cpos)
```

Nothing happens and the callback is never executed. I don't really understand the timer callback stuff

dcbr:
I think this test should call `iren.process_events` until the `TimerEvent` is fired, similarly to what is done in [`test_timer`](https://github.com/pyvista/pyvista/blob/e6ebf94920a02a54e6c04bb583b9d4f98b3d88f1/tests/plotting/test_render_window_interactor.py#L149) (not sure why we have different methods to deal with `TimerEvent` though).

pyvista/plotting/render_window_interactor.py
```diff
@@ -45,9 +45,10 @@ def __init__(self, max_steps, callback):
 
     def execute(self, obj, _event):  # pragma: no cover # numpydoc ignore=PR01,RT01
         """Execute Timer."""
+        iren = obj
```
tkoyama010, line 48:
```suggestion
        # https://github.com/pyvista/pyvista/pull/5618
        iren = obj
```

tkoyama010 (approved):
I added the link to this PR. If we need to discuss this change again, let's discuss it in another PR.

PR #5623, opened 2024-02-15
Drop Python3.8 support which EOL is 2024-10

tkoyama010:
### Overview

Several CI errors are currently occurring with Python 3.8 errors. This is unlikely to be of PyVista origin. As Python 3.8 will be EOL this October, it is better to end support than spend time investigating it.

### Details

- None

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -134,8 +134,6 @@ jobs:
       # see discussion at https://github.com/pyvista/pyvista/issues/2867
       matrix:
         include:
-          - python-version: "3.8"
-            vtk-version: "9.0.3" # Requires numpy<1.24
           - python-version: "3.9"
             vtk-version: "9.1"
           - python-version: "3.10"
@@ -236,7 +234,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v4
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -85,7 +85,7 @@ repos:
     rev: v3.15.0
     hooks:
       - id: pyupgrade
-        args: [--py38-plus, --keep-runtime-typing]
+        args: [--py39-plus, --keep-runtime-typing]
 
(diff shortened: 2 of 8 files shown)


tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

banesullivan (changes requested):
This is a bit too soon in my opinion. Would you please point me to the failures related to Python 3.8?

tkoyama010:
> This is a bit too soon in my opinion. Would you please point me to the failures related to Python 3.8?

OK. Please give me time.

MatthewFlamm:
I suspect this is a `trame_vtk` change that is incompatible with python 3.8.  Trying to see if this is fixed in #5628.

tkoyama010:
> I suspect this is a `trame_vtk` change that is incompatible with python 3.8. Trying to see if this is fixed in #5628.

Thank you. I will check and merge it soon.

hakostra:
Just want to support this and give my opinions:

Big, important packages like matplotlib, scipy and numpy no longer support Python 3.8 (at least they do not publish packages for it - I have no idea if you still can compile form source against Python 3.8 - but I guess not many do that anyway).

Matplotlib 3.7.5 and Numpy 1.24.4 is the currently last published versions supporting Python 3.8. Matplotlib >= 3.8 and numpy >= 1.25 publish packages for Python version >= 3.9 only.

For scipy you have to go back to 1.10.1, released 19. February 2023, one year ago, to find the last release for Python 3.8: https://pypi.org/project/scipy/1.10.1/#files

So it seems that the ecosystem for numerical computations in Python have already ditched Python 3.8...

MatthewFlamm:
It was previously decided to support all python versions https://github.com/pyvista/pyvista/pull/2961 and vtk versions that support those python versions.  Maybe this, or some other policy, should be officially written somewhere.  A big issue here is that we are more tied to vtk than numpy, etc.  Although, I'm generally in favor of getting closer to the scientific ecosystem at large.

Scientific python packages are encouraged to support python only 3.10+ today. 3.10 will be dropped in October of 2024. I think this velocity will be more difficult here.  Our package is less stable so users need to upgrade often and users may have special vtk build requirements they have to maintain.  At least I do.
https://scientific-python.org/specs/spec-0000/

tkoyama010:
>Maybe this, or some other policy, should be officially written somewhere.

I have an idea about this and will create a Pull Request later.

tkoyama010:
I raised #5713. Thank you all for discussing this matter.

Issue #5619, opened 2024-02-14
[Feature Request, Camera] enabling parallel projection, roll, elevation, and azimuth in global/custom theme


MarkusZimmerDLR:
Dear pyvista-folks,

It would be nice if some more camera attributes would be settable over a custom theme. 
Namely:
- parallel projection
- azimuth
- elevation
- roll

Thanks a ton!
Cheers,
Markus

### Discussed in https://github.com/pyvista/pyvista/discussions/5612

<div type='discussions-op-text'>

<sup>Originally posted by **MarkusZimmerDLR** February 13, 2024</sup>
Dear pyvista folks,

I am trying to activate parallel projection for a complete project by setting it in the global or a custom theme.

However, setting doesn't do anything

```python
>>> from pyvista import themes
>>> import pyvista as pv
>>> t = themes.DocumentTheme()
>>> t.camera['parallel_projection']=True
>>> t.camera['_parallel_projection']=True
>>> p = pv.Plotter(theme=t)
>>> p.camera.parallel_projection
False
```

A quick look into the source code of the Camera class reveals:
```python
class Camera(_vtk.vtkCamera):
    def __init__(self, renderer=None):
        """Initialize a new camera descriptor."""
        self._parallel_projection = False
        self._elevation = 0.0
        self._azimuth = 0.0
        self._is_set = False
```
And I cannot find a reference to the theme anywhere.

Am I missing something? The camera key actually exists in some themes.

Cheers
Markus</div>

PR #5624, opened 2024-02-15
Add parallel_projection and parallel_scale to Theme; make `camera` setting a first class `_ThemeConfig`

MatthewFlamm:
### Overview

Addresses part of #5619.

### Breaking Change

The `Theme.camera` attribute is now a `_ThemeConfig` object.  It is still supported to set the `position`, `viewup`, etc. via a `dict`, for example `theme.camera = {'position': [1.0, 1.0, 1.0]}`, but the returned value will no longer be a dict, i.e. `camera = theme.camera` will not be a `dict` itself.


MatthewFlamm:
The current PR is a naive implementation with no deprecations.  This puts `parallel_projection` and `parallel_scale` as direct `Theme` attributes.  These are camera options however.  I want to propose deprecating setting `Theme.camera` as a `dict` and instead make it a `_ThemConfig`.  This might also be possible to do without deprecation.

MatthewFlamm:
This PR has converted the `Theme.camera` into using a `_CameraConfig` class object.  But, for backwards compatibility, we convert any dict passed into a `_CameraConfig`.  This is not a breaking change for setting the config via dict, but is a breaking change for anyone inspecting the `camera` attribute and expecting a `dict`.

This sort of casting dict into a `_ThemeConfig` was also suggested in #5119, and might be worth a separate PR to expose for all `_ThemeConfig`.

MatthewFlamm:
@MarkusZimmerDLR if you want to test

MatthewFlamm:
I made a quick scan of the other attributes of `Theme` and did not see any other candidates for this conversion.  For consistency and flexibility in the future, we should discourage using a `dict` for any theme setting, particularly on the top level `Theme`.

tkoyama010 (approved):
LGTM. As far as breaking backward compatibility, I guess it is not a problem in this case. Let's only notify on the release note.

PR #5628, opened 2024-02-15
Fix python 3.8 errors in CI by pinning trame-vtk

MatthewFlamm:
### Overview

trame-vtk recent release broke compatibility with python 3.8

PR #5629, opened 2024-02-15
Specify the version of test library at patch level

tkoyama010:
### Overview

Suggestion based on discussion in #5628. I propose to specify the version of the test library at the patch level so that errors due to patch version releases are noticed by updating `dependabot`.

### Details

- None


tkoyama010:
@pyvista-bot LGTM

PR #5637, opened 2024-02-16
Add style rule derived from flake8-pytest-style

tkoyama010:
### Overview

Add style rule derived from [`flake8-pytest-style`](https://pypi.org/project/flake8-pytest-style/).

### Details

- [PT011](https://docs.astral.sh/ruff/rules/pytest-raises-too-broad/) which needs to add a `match` argument add `noqa: PT011` to hope that it will be added in the future. Let's add `match` argument in new tests.


tkoyama010:
@pyvista-bot LGTM

PR #5643, opened 2024-02-18
Modify `test_direction_objects` to give meaningful names to test cases

user27182:
Follow up from #5614.
The test cases from that PR have generic names, which makes it hard to know which object the tests are failing for. This PR modifies how the test is parametrized to give a meaningful name id.


tkoyama010:
This is awesome. Please give me time to study about `pytest_generate_tests example`.
Note: [Basic pytest_generate_tests example](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html#pytest-generate-tests)

user27182:
`pytest_generate_tests` is also used here in #5571: 
https://github.com/pyvista/pyvista/blob/b562ee4b8d48b47860564ded7d2a472778b305bc/tests/core/typing/test_validation_typing.py#L143

For this PR and that one, this kind of parameterization is useful whenever test cases are generated dynamically.

MatthewFlamm (dismissed):
This is great!  We should do more naming of parametrized tests in general.  One small thing is that most images are being picked up as renamed without changes. Some images are showing up as being deleted and then recreated.  If these could be fixed it would keep the image size churn in the git repo smaller.  But this can be tricky to get to happen with git in my experience with this number of files.  I approve even if this isn't addressed.

user27182:
> Some images are showing up as being deleted and then recreated. If these could be fixed it would keep the image size churn in the git repo smaller. But this can be tricky to get to happen with git in my experience with this number of files. I approve even if this isn't addressed.

Yes I agree renamed is preferred to deleted and added. I did do a lazy copy/paste from `debug_images` instead of manually renaming each file, I could try to manually rename the files that are being flagged as deleted to fix this.

MatthewFlamm:
> > Some images are showing up as being deleted and then recreated. If these could be fixed it would keep the image size churn in the git repo smaller. But this can be tricky to get to happen with git in my experience with this number of files. I approve even if this isn't addressed.
> 
> Yes I agree renamed is preferred to deleted and added. I did do a lazy copy/paste from `debug_images` instead of manually renaming each file, I could try to manually rename the files that are being flagged as deleted to fix this.

It will be annoying for this number of files (so feel free to decide this is too much work), but I have had most success with `git-mv`.

user27182:
Another thing to consider is that `test_direction_objects` is doing (at least) two things: testing the images, but also testing to ensure that new `direction` objects have new regression images added. This can kind of complicate things...

So to de-couple these, one option would be to split `test_direction_objects` into two separate tests:

1. Static test which simply plots and tests the images using pre-determined test cases, e.g.:
``` python
TEST_CASES = [ 
    ('Arrow', pyvista.Arrow, 'direction'), 
    ('Plane', pyvista.Plane, 'normal'), 
    ... 
]
def test_direction_objects(TEST_CASES): ...
```

2. Dynamic test which generates a list of direction objects, and the test passes if there is a static test case for every direction object, but the test passes if a new object without a test cases is added, e.g.:
``` python 
def test_direction_objects_coverage():
    expected_cases = TEST_CASES
    actual_cases = _generate_dynamic_test_cases
    if not (case in expected_cases for case in actual_cases):
        pytest.fail()

user27182:
>  Some images are showing up as being deleted and then recreated.

Fixed in [2b978e8](https://github.com/pyvista/pyvista/pull/5643/commits/2b978e8e60a0d361bc441e4d3ae1556a06654377)

I also removed the conditional for pyvista 0.43 in b99ecd3, this was only needed local testing as part of #5435 

> Another thing to consider is that `test_direction_objects` is doing (at least) two things: testing the images, but also testing to ensure that new `direction` objects have new regression images added.

On second thought, I think keeping these together is fine. It's pretty clear as-is with the explicit list of expected objects. With the named images any failed tests due to missing images should also be clear.

LGTM now

tkoyama010 (approved):
LGTM

PR #5647, opened 2024-02-18
Implement capsule mesh creation

namheegordonkim:
### Overview

Capsule objects are very commonly used in rigid body simulators. Adding support for capsules will increase the viability of PyVista as a renderer for physics simulators that have no mature headless rendering solution, e.g. Brax and Isaac.

### Details

- `utilities` module to load and register `Capsule` class
- `geometric_objects.py` to create capsules by deforming hemisphere vertex positions (test cases are included in the docstring)

PR #5655, opened 2024-02-20
Add capsule geometry use vtkCapsuleSource and vtkCylinderSource with `SetCapsuleCap`

kmarchais:
### Overview

Implementation of the capsule geometry following #5647 .
VTK 9.0 introduced `vtkCapsuleSource` which was deprecated in 9.3 and replaced by the hemispherical capping on the `vtkCylinderSource`.

I tried to implement this while taking care about the different versions but I am not sure how to manage this.
I also added corresponding tests.

Tested with VTK 9.3 and VTK 9.2.

| VTK 9.3 | 9.0 <= VTK <= 9.2 |
|----------|---------------------|
| image | image |

Edit: close #5647


user27182:
Since the new `Capsule` object has a `direction`, the `direction_objects` tests will fail unless a new regression image is added for `Capsule`.

The regression tests were added very recently in #5614. However, since the images are generated in alphabetical order (but with numerical values`, it's likely that all direction object images which come after `Capsule` alphabetically will need to be renumbered. This means that many of the test images will show up as renamed and/or deleted. PR #5643 should fix this issue, such that only the new `direction_objects[Capsule-pos].png` and `direction_objects[Capsule-neg].png` will need to be added.

user27182:
(Note: to generate the images locally you can run `pytest tests/plotting/ -k direction --generated_image_dir debug_images`, then copy the new images from `./debug_images` to `./tests/plotting/image_cache`)

kmarchais:
Alright, thanks a lot for this @user27182 !
I've generated the images, copied and replaced the ones in `./tests/plotting/image_cache`. Now the test seems to pass, at least locally.
Should I commit this or wait for PR #5643 ?

user27182:
@kmarchais no problem, yes please commit any changes to the image cache, it will be good to run all the CI tests to make sure everything else is still good.

user27182:
Ok yes so as I thought it looks like pretty much all images had to be renumbered, which is not ideal since this PR otherwise should have no impact on the other images. I think #5643 should be merged first, but that will be up do the main devs to decide.

Also, it looks some other `direction` images were modifed in the cache, unrelated to the `direction_objects` test. These should probably be deleted / left unchanged in the image cache.

This command should be more specific and generate only the images from that test (`direction` is replaced with `direction_objects`):

`pytest tests/plotting/ -k direction_objects --generated_image_dir debug_images`

kmarchais:
So it looks good with Python 3.12 which has VTK 9.3 but not with VTK < 9.3. 
I guess it's because of the difference between the capsules shown on the first message of this PR.

user27182:
> So it looks good with Python 3.12 which has VTK 9.3 but not with VTK < 9.3.
> I guess it's because of the difference between the capsules shown on the first message of this PR.

One option may be to use `pytest.xfail`condtionally inside `test_direction_objects` if `name==Capusle` and `vtk_version < (9, 3)`. Perhaps better would be to increase the regression error tolerance conditionally just for that test, though I'm not sure if that's possible.

If the geometries are substantially different across VTK versions, is it worth adding a warning to the docs, to warn users about this?

kmarchais:
Here is the difference caused by the way the resolution is managed in both versions.

image

In VTK 9.3, the resolution follows the one of a Cylinder while before 9.3 it has a `phi_resolution` and a `theta_resolution` like a Sphere.

In the new `Capsule` function in `pyvista/core/utilities/geometric_objects.py` both `phi` and `theta` resolutions are given but only `theta_resolution` is used as `resolution` in the VTK 9.3 case.
To follow VTK's structure, maybe it would be better to have the `Capsule` function only for 9.0 <= VTK <= 9.2 since the capsule is managed with the `vtkCylinderSource` from 9.3.
But I am not sure how to manage this correctly, especially with regression tests.

user27182:
> In the new `Capsule` function in `pyvista/core/utilities/geometric_objects.py` both `phi` and `theta` resolutions are given but only `theta_resolution` is used as `resolution` in the VTK 9.3 case. 

I'm not familiar with the use cases of `Capsule` objects, but would it be acceptable in this case to just have a single `resolution` parameter, which internally would set `phi_resolution = theta_resolution = resolution` for VTK < 9.3 ?

> To follow VTK's structure, maybe it would be better to have the `Capsule` function only for 9.0 <= VTK <= 9.2 since the capsule is managed with the `vtkCylinderSource` from 9.3. 

If capsules have a distinct use cases, I think having a unified `Capsule` API is good, regardless of `vtk` version. Even if it uses `vtkCylinderSource` for 9.3,  having rounded hemispheres by default would at least make `Capsule` a convenience/wrapper function.

> But I am not sure how to manage this correctly, especially with regression tests.

Given that the geometries do differ substantially in terms of their mesh construction, we should probably test these separately. I think adding a conditional `pytest.xfail` in `test_direction_objects` for the < 9.3 cases should be OK. We can test the deprecated < 9.3 case separately.

kmarchais:
Trying with `verify_image_cache.skip = name == 'Capsule' and pv.vtk_version_info < (9, 3, 0)` since `verify_image_cache` is used in other tests.
Is this similar to using `xfail` @user27182 ?

tests/plotting/test_plotting.py
```diff
@@ -4102,14 +4102,19 @@ def _generate_direction_object_functions() -> List[Tuple[str, FunctionType]]:
...
+        ]
     assert sorted(actual_names) == sorted(expected_names)
     return list(functions.items())
 
 
 @pytest.mark.parametrize('positive_dir', [True, False])
 @pytest.mark.parametrize('object_function', _generate_direction_object_functions())
-def test_direction_objects(object_function, positive_dir):
+def test_direction_objects(object_function, positive_dir, verify_image_cache):
     name, func = object_function
```
user27182, line 4119:
```suggestion

    # Test Capsule separately based on vtk version
    if 'Capsule' in name:
       legacy_vtk = pv.vtk_version_info.major < 9 and pv.vtk_version_info.minor < 3
       if legacy_vtk and 'legacy' not in name or not legacy_vtk and 'legacy' in name:
            pytest.xfail('Test capsule separately for different vtk versions. Expected to fail if testing with wrong version.')
            
```

tests/plotting/test_plotting.py
```diff
@@ -4102,14 +4102,19 @@ def _generate_direction_object_functions() -> List[Tuple[str, FunctionType]]:
...
+            'Text3D',
+        ]
     assert sorted(actual_names) == sorted(expected_names)
     return list(functions.items())
 
 
 @pytest.mark.parametrize('positive_dir', [True, False])
 @pytest.mark.parametrize('object_function', _generate_direction_object_functions())
-def test_direction_objects(object_function, positive_dir):
+def test_direction_objects(object_function, positive_dir, verify_image_cache):
```
user27182, line 4117:
```suggestion
def test_direction_objects(object_function, positive_dir):
```

tests/plotting/test_plotting.py
```diff
@@ -4119,6 +4124,8 @@ def test_direction_objects(object_function, positive_dir):
     elif name == 'Text3D':
         kwargs['string'] = 'Text3D'
 
+    verify_image_cache.skip = name == 'Capsule' and pv.vtk_version_info < (9, 3, 0)
```
user27182, line 4127:
```suggestion

```

tests/plotting/test_plotting.py
```diff
@@ -4102,14 +4102,19 @@ def _generate_direction_object_functions() -> List[Tuple[str, FunctionType]]:
...
     major, minor, patch = pv._version.version_info
     if major == 0 and minor >= 43:
-        expected_names += ['SolidSphere', 'SolidSphereGeneric', 'Text3D']
+        expected_names += [
+            'Capsule',
+            'SolidSphere',
+            'SolidSphereGeneric',
+            'Text3D',
+        ]
     assert sorted(actual_names) == sorted(expected_names)
```
user27182, line 4111:
```suggestion

        # Add a separate test for vtk < 9.3
        expected_names.append('Capsule_legacy')
        actual_names.append('Capsule_legacy')
    
    assert sorted(actual_names) == sorted(expected_names)
```

user27182 (commented):
Here is an alternative where an extra test case is added. This way, regression images are generated for both new and legacy versions of Capsule. No skips are used, instead the tests will still run for mis-matched vtk versions but are marked as expected to fail.

user27182:
> Trying with `verify_image_cache.skip = name == 'Capsule' and pv.vtk_version_info < (9, 3, 0)` since `verify_image_cache` is used in other tests.
> Is this similar to using `xfail` @user27182 ?

This would have similar effect yes. I think there are subtle differences, such as skipping altogether means the test isn't executed, whereas `xfail` will still run the test. Also, using `xfail` would show up in test summaries reported by pytest, wherase `verify_image_cache.skip` may not, though this may not be a problem.

user27182:
I think the main thing is that we may not actually want to completely skip the tests for the legacy Capsule, so we can try to make a separate regression test altogether just for that case, or manually add a new case to the `test_direction_objects` test

user27182:
> image

With the additional test cases, tests with vtk version < 9.3 will test against the legacy image. So, two regression images (positive and negative direction) need to be added for both cases shown in these screenshots.

Ideally, with #5643, these would be:

- `direction_objects[Capsule-pos].png`
- `direction_objects[Capsule-neg].png`
- `direction_objects[Capsule_legacy-pos].png`
- `direction_objects[Capsule_legacy-neg].png`

kmarchais:
Ok great! 
With the current naming is there a way to make it pass?
I think it will be named `object_function1` in both cases.

user27182:
> I think it will be named `object_function1` in both cases.

These lines should be generating both a `Capsule` and `Capsule_legacy` test case, regardless of vtk version. So in theory `Capsule` -> `object_function1` and `Capsule_legacy` -> `object_function2`... but it seems like this isn't the case?https://github.com/pyvista/pyvista/blob/64b409a0db6ed1eda636730546a75ad18ec1a221/tests/plotting/test_plotting.py#L4104-L4114

I would expect that the generated images with vtk 9.3 would result in images for `object_function1` and `object_function2` being identical. Running again with vtk < 9.3 would generate the legacy images, but again `object_function1` and `object_function2` would be identical. 

And then copy the 9.3 version images for `object_function1` to the cache, and copy the <9.3 images for `object_function2` to the cache. After copying, running `pytest` locally  vtk 9.3 the should result in `object_function1` tests passing, and `object_function2` tests failing. But running the tests again with vtk <9.3 should result in the opposite cases passing/failing. 

Also, I would also expect that all downstream images would (again) have to be renumbered and re-commited.

tests/plotting/test_plotting.py
```diff
@@ -4102,7 +4102,17 @@ def _generate_direction_object_functions() -> List[Tuple[str, FunctionType]]:
...
+            'Capsule',
+            'SolidSphere',
+            'SolidSphereGeneric',
+            'Text3D',
+        ]
+
+        # Add a separate test for vtk < 9.3
+        expected_names.append('Capsule_legacy')
+        actual_names.append('Capsule_legacy')
+
```
user27182, line 4115:
```suggestion

```

tests/plotting/test_plotting.py
```diff
@@ -4102,7 +4102,17 @@ def _generate_direction_object_functions() -> List[Tuple[str, FunctionType]]:
 
     major, minor, patch = pv._version.version_info
     if major == 0 and minor >= 43:
```
user27182, line 4104:
```suggestion

```

tests/plotting/test_plotting.py
```diff
@@ -4066,7 +4066,7 @@ def _generate_direction_object_functions() -> List[Tuple[str, FunctionType]]:
         if name[0].isupper() and (_has_param(func, 'direction') or _has_param(func, 'normal'))
     }
 
-    actual_names = functions.keys()
+    actual_names = list(functions.keys())
     expected_names = [
         'Arrow',
         'CircularArcFromNormal',
```
user27182, line 4072:
```suggestion
        'Capsule',
        'Capsule_legacy',
        'CircularArcFromNormal',
```

user27182 (commented):
The issue was that the legacy case was not added to the actual test case dictionary.

Also, the conditional for pyvista 0.43 can be safely removed (it was added only for testing part of #5435 )

tests/plotting/test_plotting.py
```diff
@@ -4066,7 +4066,7 @@ def _generate_direction_object_functions() -> List[Tuple[str, FunctionType]]:
         if name[0].isupper() and (_has_param(func, 'direction') or _has_param(func, 'normal'))
     }
```
user27182, line 4068: This is the key line needed to add the new test case
```suggestion
functions['Capsule_legacy'] = functions['Capsule']
```

MatthewFlamm (commented):
I left some comments for the PR as is.  I haven't fully thought this one through, but would it better to simply have this option only be available for vtk >9.3 and expose this option in Cylinder as a parameter? This would reduce maintenance burden, but reduce availability for vtk <9.3.  Given 9.3 is just released, I'd probably lean towards the current PR state.

I think we should wait for the changes to the test images for geometric objects with direction for this one.

pyvista/core/utilities/geometric_objects.py
```diff
@@ -56,6 +62,79 @@
...
+    Examples
+    --------
+    Create a capsule using default parameters.
+
+    >>> import pyvista as pv
+    >>> capsule = pv.Capsule()
+    >>> capsule.plot(show_edges=True)
+
+    """
+    if pyvista.vtk_version_info.major >= 9:
```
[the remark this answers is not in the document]
kmarchais: Thanks!

tests/plotting/test_plotting.py
```diff
@@ -4100,9 +4103,6 @@ def _generate_direction_object_functions() -> List[Tuple[str, FunctionType]]:
         'Sphere',
```
user27182, line 4103:
```suggestion
        'SolidSphere', 
        'SolidSphereGeneric',
        'Sphere',
        'Text3D'
```

I should have re-added these here when I suggested removing them from the conditional 0.43 block

Thanks for your patience dealing with these changes!

pyvista/core/utilities/geometric_sources.py
```diff
@@ -61,6 +61,212 @@ def translate(surf, center=(0.0, 0.0, 0.0), direction=(1.0, 0.0, 0.0)):
         surf.points += np.array(center, dtype=surf.points.dtype)
 
 
+if _vtk.vtkVersion.GetVTKMajorVersion() >= 9 and 0 <= _vtk.vtkVersion.GetVTKMinorVersion() <= 2:
```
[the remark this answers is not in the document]
kmarchais: This seems to be causing a circular import 

```
if pyvista.vtk_version_info <= (9, 2):
AttributeError: partially initialized module 'pyvista' has no attribute 'vtk_version_info' (most likely due to a circular import)
```

user27182:
Looks like #5643 will be approved, sorry for all the trouble @kmarchais dealing with the test images. Once that PR is merged the image cache here should be easier to work with, and only the 4 named images mentioned earlier will need to be added.

kmarchais:
Ok perfect! Thanks for the help!

kmarchais:
> > I think it will be named `object_function1` in both cases.
> 
> These lines should be generating both a `Capsule` and `Capsule_legacy` test case, regardless of vtk version. So in theory `Capsule` -> `object_function1` and `Capsule_legacy` -> `object_function2`... but it seems like this isn't the case?
> 
> https://github.com/pyvista/pyvista/blob/64b409a0db6ed1eda636730546a75ad18ec1a221/tests/plotting/test_plotting.py#L4104-L4114
> 
> I would expect that the generated images with vtk 9.3 would result in images for `object_function1` and `object_function2` being identical. Running again with vtk < 9.3 would generate the legacy images, but again `object_function1` and `object_function2` would be identical.
> 
> And then copy the 9.3 version images for `object_function1` to the cache, and copy the <9.3 images for `object_function2` to the cache. After copying, running `pytest` locally vtk 9.3 the should result in `object_function1` tests passing, and `object_function2` tests failing. But running the tests again with vtk <9.3 should result in the opposite cases passing/failing. All tests should now pass locally for any vtk version since the conditional should correctly mark if a test is expected to fail
> 
> Also, I would also expect that all downstream images would (again) have to be renumbered and re-commited.

I keep having only one version generated (`function1`), `function2` being `CircularArcFromNormal`. I may be missing something though.

user27182:
> I keep having only one version generated (`function1`), `function2` being `CircularArcFromNormal`. I may be missing something though.

The `Capsule_legacy` case is added to the end of the test cases dict. And the dict itself isn't sorted before the test cases are processed by pytest. So, I think `Capsule_legacy` may actually be `object_function32` or something like that instead of `object_function2`. So, you could sort the dict as part of the `return` statement, or look for item 32 in the image cache, or wait until #5643 is merged, which should fix all these issues with the numbered test cases.

pyvista/core/utilities/geometric_sources.py
MatthewFlamm, line 64: A different suggestion to avoid the circular import.  We only support vtk >=9.

```suggestion
if _vtk.vtkVersion.GetVTKMajorVersion() == 9 and _vtk.vtkVersion.GetVTKMinorVersion() <= 2:
```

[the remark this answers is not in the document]
MatthewFlamm: Hmmm.  I wonder if we need to move `vtk_version_info` instead.  I'd suggest inside `_vtk_core`.  But that isn't for this PR.  Let another suggestion.

kmarchais:
I have just added `Capsule` and `Capsule_legacy` to `image_cache` but I also had to add all the `direction_objects` images to pass the tests locally.
Also, running the command `pytest tests/plotting/ -k direction_objects --generated_image_dir debug_images` generated 69 `direction_objects[Capsule-pos]_*.png` extra files (70 being the total number of `direction_objects` images)  and I don(t know why.

user27182:
> I have just added `Capsule` and `Capsule_legacy` to `image_cache` but I also had to add all the `direction_objects` images to pass the tests locally.

You can try running 
`git checkout --patch main tests/plotting/image_cache`

This should replace all images in your cache locally with the original ones from main for all direction objects. Do the tests still fail locally then?

> Also, running the command `pytest tests/plotting/ -k direction_objects --generated_image_dir debug_images` generated 69 `direction_objects[Capsule-pos]_*.png` extra files (70 being the total number of `direction_objects` images)  and I don't know why.

When you run the plotting tests and there is a missing image in the cache, pytest will sometimes re-try the test automatically. This can mean it will generate many images in `debug_images` (one for each test attempt). The solution is to copy just one of these images to `image_cache`. The next time you run the test it should be OK since now there is no missing image and pytest won't retry the test.

kmarchais:
> You can try running 
> `git checkout --patch main tests/plotting/image_cache`
This tells me `cannot apply binary patch` ans `patch does not apply`.

When looking at the differences between images using swipe mode, the positions of some axes labels differ.

user27182:
> > You can try running 
> 
> > `git checkout --patch main tests/plotting/image_cache`
> 
> 
> This tells me `cannot apply binary patch` ans `patch does not apply`.

You can try deleting all images in your local image_cache first (except for the 4 new capsule images), and then try the command again.

> When looking at the differences between images using swipe mode, the positions of some axes labels differ.

I have not reviewed the diff between main and this branch for the direction object images, but sometimes there are minor differences between images generated on Windows vs MacOS vs Linux. There is a small tolerance in the tests to allow for this.

You can try committing the original images and push the changes anyway, see how the CI tests do

kmarchais:
> You can try deleting all images in your local image_cache first (except for the 4 new capsule images), and then try the command again.

I tried this also but didn't change. I probably don't do it the right way.
So I put back the original images and I think it is good now.

pyvista/core/utilities/__init__.py
```diff
@@ -100,6 +101,12 @@
     SuperquadricSource,
     Text3DSource,
 )
+
+try:
+    from .geometric_sources import CapsuleSource
+except ImportError:
```
tkoyama010, line 107:
```suggestion
except ImportError:  # pragma: no cover
```

pyvista/core/utilities/geometric_objects.py
```diff
@@ -44,6 +44,12 @@
     Text3DSource,
     translate,
 )
+
+try:
+    from .geometric_sources import CapsuleSource
+except ImportError:
```
tkoyama010, line 50:
```suggestion
except ImportError:  # pragma: no cover
```

pyvista/core/utilities/geometric_objects.py
```diff
@@ -57,6 +63,76 @@
...
+    Examples
+    --------
+    Create a capsule using default parameters.
+
+    >>> import pyvista as pv
+    >>> capsule = pv.Capsule()
+    >>> capsule.plot(show_edges=True)
+
+    """
+    if pyvista.vtk_version_info >= (9, 3):
```
tkoyama010, line 111:
```suggestion
    if pyvista.vtk_version_info >= (9, 3):  # pragma: no cover
```

tests/core/test_geometric_sources.py
```diff
@@ -37,6 +51,13 @@ def test_cylinder_source():
     assert np.array_equal(algo.direction, direction)
 
 
+def test_cylinder_capsule_cap():
+    if pv.vtk_version_info >= (9, 3):
+        algo = pv.CylinderSource()
+        algo.capsule_cap = True
+        assert algo.capsule_cap
```
tkoyama010, line 58:
```suggestion
@pytest.mark.needs_vtk_version(9, 3, 0)
def test_cylinder_capsule_cap():
    algo = pv.CylinderSource()
    algo.capsule_cap = True
    assert algo.capsule_cap
```

tests/core/test_geometric_sources.py
```diff
@@ -7,6 +7,20 @@
...
+        algo = pv.CapsuleSource()
+        assert np.array_equal(algo.center, (0.0, 0.0, 0.0))
+        assert np.array_equal(algo.direction, (1.0, 0.0, 0.0))
+        assert algo.radius == 0.5
+        assert algo.cylinder_length == 1.0
+        assert algo.theta_resolution == 30
+        assert algo.phi_resolution == 30
+        direction = np.random.default_rng().random(3)
+        algo.direction = direction
+        assert np.array_equal(algo.direction, direction)
```
tkoyama010, line 21:
```suggestion
@pytest.mark.needs_vtk_version(9, 3, 0)
def test_capsule_source():
    algo = pv.CapsuleSource()
    assert np.array_equal(algo.center, (0.0, 0.0, 0.0))
    assert np.array_equal(algo.direction, (1.0, 0.0, 0.0))
    assert algo.radius == 0.5
    assert algo.cylinder_length == 1.0
    assert algo.theta_resolution == 30
    assert algo.phi_resolution == 30
    direction = np.random.default_rng().random(3)
    algo.direction = direction
    assert np.array_equal(algo.direction, direction)
```

tests/core/test_geometric_sources.py
```diff
@@ -390,3 +411,4 @@ def test_box_source():
     assert np.array_equal(algo.bounds, [-1.0, 1.0, -1.0, 1.0, -1.0, 1.0])
     assert algo.level == 0
     assert algo.quads
+    assert algo.quads
```
tkoyama010, line 414:
```suggestion
```

tests/plotting/test_plotting.py
```diff
@@ -4142,6 +4145,14 @@ def test_direction_objects(direction_obj_test_case):
     elif name == 'Text3D':
         kwargs['string'] = 'Text3D'
 
+    # Test Capsule separately based on vtk version
+    if 'Capsule' in name:
+        legacy_vtk = pv.vtk_version_info.major == 9 and pv.vtk_version_info.minor < 3
```
tkoyama010, line 4150:
```suggestion
        legacy_vtk = pv.vtk_version_info < (9, 3)
```

tkoyama010 (changes requested):
Awesome job!

kmarchais:
Thanks a lot for all the help!

tkoyama010:
github-actions preview

tkoyama010 (approved):
LGTM
<p align="center">

</p>

tkoyama010:
@namheegordonkim Could you confirm that this will also meet your needs in #5647?

pyvista/core/utilities/geometric_sources.py
```diff
@@ -64,6 +64,212 @@ def translate(surf, center=(0.0, 0.0, 0.0), direction=(1.0, 0.0, 0.0)):
         surf.points += np.array(center, dtype=surf.points.dtype)
 
 
+if _vtk.vtkVersion.GetVTKMajorVersion() == 9 and _vtk.vtkVersion.GetVTKMinorVersion() <= 2:
+
+    @no_new_attr
+    class CapsuleSource(_vtk.vtkCapsuleSource):
+        """Capsule source algorithm class.
+
```
user27182, line 72:
```suggestion

.. versionadded:: 0.44.0

```

pyvista/core/utilities/geometric_sources.py
```diff
@@ -519,6 +725,28 @@ def capping(self, capping: bool):
         """
         self.SetCapping(capping)
 
+    @property
+    def capsule_cap(self) -> bool:
+        """Get whether the capping should make the cylinder a capsule.
+
```
user27182, line 731:
```suggestion

.. versionadded:: 0.44.0

```

user27182 (commented):
LGTM, some minor suggestions for docstrings

tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

namheegordonkim:
Haven't pulled these to check with the rest of my visualizer code, but from the looks of it, it definitely fits my bill! Thank you for the awesome work.

PR #5656, opened 2024-02-20
Add `get_charts` method to `Renderer`

kmarchais:
### Overview

The aim of this PR is to add properties to `Renderer` and `Charts` classes to be able to access matplotlib plots from the plotter without having to use protected attributes.

- Add `mpl_charts` property to `Renderer`: returns a list of charts. It should probably be renamed but I wanted it to not be too close to `_charts` which is a `Charts` object.
- Add `charts` property to `Charts`: returns `self._charts` but I guess there is a good reason why it is a protected variable.

With this change, the following code:
```python
axes = plotter.renderer._charts._charts[0].figure.axes
```

becomes:

```python
axes = plotter.renderer.mpl_charts[0].figure.axes
```

Example:
```python
import pyvista as pv

fig, ax = plt.subplots()
ax.plot([1, 2, 3], [4, 5, 6])

chart = pv.ChartMPL(fig)

pl = pv.Plotter(shape=(1, 2))

pl.subplot(0, 0)
pl.add_mesh(pv.Sphere())

pl.subplot(0, 1)
pl.add_chart(chart)

ax = pl.renderer.mpl_charts[0].figure.axes[0]
ax.get_lines()[0].set_ydata([6, 5, 4])

pl.show()
``` 

image

### DIFF
--- a/pyvista/plotting/charts.py
+++ b/pyvista/plotting/charts.py
@@ -4742,6 +4742,11 @@ class Charts:
         """Return the weakly dereferenced renderer, maybe None."""
         return self.__renderer()
 
+    @property
+    def charts(self):  # numpydoc ignore=RT01
+        """Return the list of charts in this collection."""
+        return self._charts
+
     def _setup_scene(self):
         """Set up a new context scene and actor for these charts."""
         self._scene = _vtk.vtkContextScene()
--- a/pyvista/plotting/charts.py
+++ b/pyvista/plotting/charts.py
@@ -4742,11 +4742,6 @@ class Charts:
         """Return the weakly dereferenced renderer, maybe None."""
         return self.__renderer()
 
-    @property
-    def charts(self):  # numpydoc ignore=RT01
-        """Return the list of charts in this collection."""
-        return self._charts
-
     def _setup_scene(self):
         """Set up a new context scene and actor for these charts."""
         self._scene = _vtk.vtkContextScene()
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -697,7 +697,7 @@ class Renderer(_vtk.vtkOpenGLRenderer):
 
     @property
     def charts(self):  # numpydoc ignore=RT01
-        """Return a list of all Matplotlib charts in this renderer.
+        """Return a list of all charts in this renderer.
 
         Examples
         --------
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -705,10 +705,10 @@ class Renderer(_vtk.vtkOpenGLRenderer):
 
            >>> import pyvista as pv
            >>> chart = pv.Chart2D()
-           >>> chart.line([1, 2, 3], [0, 1, 0])
+           >>> _ = chart.line([1, 2, 3], [0, 1, 0])
            >>> pl = pv.Plotter()
            >>> pl.add_chart(chart)
-           >>> print(chart is pl.renderer.get_charts()[0])
+           >>> chart is pl.renderer.get_charts()[0]
            True
 
         """
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -703,19 +703,12 @@ class Renderer(_vtk.vtkOpenGLRenderer):
         .. pyvista-plot::
            :force_static:
 
-           >>> import matplotlib.pyplot as plt
            >>> import pyvista as pv
-           >>> fig, ax = plt.subplots()
-           >>> _ = ax.plot([1, 2, 3], [4, 5, 6])
-           >>> chart = pv.ChartMPL(fig)
-           >>> pl = pv.Plotter(shape=(1, 2))
-           >>> pl.subplot(0, 0)
-           >>> _ = pl.add_mesh(pv.Sphere())
-           >>> pl.subplot(0, 1)
-           >>> _ = pl.add_chart(chart)
-           >>> ax = pl.renderer.get_charts()[0].figure.axes[0]
-           >>> ax.get_lines()[0].set_ydata([6, 5, 4])
-           >>> pl.show()
+           >>> chart = pv.Chart2D()
+           >>> chart.line([1, 2, 3], [0, 1, 0])
+           >>> pl = pv.Plotter()
+           >>> pl.add_chart(chart)
+           >>> print(chart is pl.renderer.get_charts()[0])  # True
 
         """
         return [*self._charts] if self.has_charts else []
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -703,8 +703,7 @@ class Renderer(_vtk.vtkOpenGLRenderer):
         --------
         >>> import matplotlib.pyplot as plt
         >>> import pyvista as pv
-        >>> fig, ax = plt.subplots()
-        >>> ax.plot([1, 2, 3], [4, 5, 6])
+        >>> fig, _ = plt.subplots()
         >>> chart = pv.ChartMPL(fig)
         >>> pl = pv.Plotter(shape=(1, 2))
         >>> pl.subplot(0, 0)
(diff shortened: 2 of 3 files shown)


tkoyama010:
github-actions preview

tkoyama010:
github-actions preview

tkoyama010:
github-actions preview

kmarchais:
I think the name `mpl_charts` is not well chosen because if I am right, it will return a list of all kind of charts, not only matplotlib charts.
I didn't want to call it `charts` though because there is `_charts` that is a `Charts` object.

pyvista/plotting/charts.py
```diff
@@ -4742,6 +4742,11 @@ def _renderer(self):
         """Return the weakly dereferenced renderer, maybe None."""
         return self.__renderer()
 
+    @property
+    def charts(self):  # numpydoc ignore=RT01
```
dcbr, line 4746: This shouldn't be needed as far as I know. The `Charts` object implements the `__getitem__`, `__len__` and `__iter__` methods, so it should be usable as any other list. Can you provide an example on why access to the explicit list would be needed?

pyvista/plotting/renderer.py
```diff
@@ -695,6 +695,32 @@ def has_charts(self):  # numpydoc ignore=RT01
         """Return whether this renderer has charts."""
         return self._charts is not None and len(self._charts) > 0
 
+    @property
+    def mpl_charts(self):  # numpydoc ignore=RT01
```
dcbr, line 699: The name is best changed indeed.
I don't remember exactly why `_charts` was made a private attribute of `Renderer`, but it might be a good idea to add a `get_charts` method that can filter on specific chart types (e.g. `ChartMPL` for matplotlib charts only).

tkoyama010:
> I think the name `mpl_charts` is not well chosen because if I am right, it will return a list of all kind of charts, not only matplotlib charts. I didn't want to call it `charts` though because there is `_charts` that is a `Charts` object.

I want to hear @dcbr 's opinion.

dcbr (changes requested):
Thank you for this contribution.
First of all I would like to ask what the specific use case is for these changes? In a typical setup (like your example), you first create the charts and then add them to the plotter. So to access those charts, you typically don't need to go through the plotter and renderer, hence why a `get_charts` method hasn't been implemented.
E.g. this line
```
axes = plotter.renderer._charts._charts[0].figure.axes
```
could just be changed by `chart.figure.axes`. I assume your actual use case might be more complicated though, so that's why I would like to learn more about it.

That being said, it might be a useful addition to have a `get_charts` method on `Renderer` that filters on specific chart types. This would require some changes to the current implementation in this PR, so let me know if you want to work towards this or have further questions.

pyvista/plotting/charts.py
kmarchais: I missed that!

pyvista/plotting/renderer.py
kmarchais: In my case I don't really need to filter specific chart types but to get the list of charts.
Would it be fine to replace `mpl_charts` by `charts`?

kmarchais:
> Thank you for this contribution. First of all I would like to ask what the specific use case is for these changes? In a typical setup (like your example), you first create the charts and then add them to the plotter. So to access those charts, you typically don't need to go through the plotter and renderer, hence why a `get_charts` method hasn't been implemented. E.g. this line
> 
> ```
> axes = plotter.renderer._charts._charts[0].figure.axes
> ```
> 
> could just be changed by `chart.figure.axes`. I assume your actual use case might be more complicated though, so that's why I would like to learn more about it.
> 
> That being said, it might be a useful addition to have a `get_charts` method on `Renderer` that filters on specific chart types. This would require some changes to the current implementation in this PR, so let me know if you want to work towards this or have further questions.

I would like to be able to modify the chart from different scopes/functions without having to bring the chart object everywhere since it is already in the plotter. Maybe there is a better way though.

tkoyama010:
> what the specific use case is for these changes?

Good point. As more and more interactive plots like #4855 are developed in the future, the ability to change the target parameters after definition makes a lot of sense in my mind.

tkoyama010:
github-actions preview

pyvista/plotting/renderer.py
dcbr: Alright, I would still make it a method (`get_charts`) instead of a property, such that filters could be added in the future then.

pyvista/plotting/renderer.py
```diff
@@ -695,6 +695,32 @@ def has_charts(self):  # numpydoc ignore=RT01
...
+    @property
+    def charts(self):  # numpydoc ignore=RT01
+        """Return a list of all charts in this renderer.
+
+        Examples
+        --------
+        .. pyvista-plot::
+           :force_static:
+
+           >>> import matplotlib.pyplot as plt
```
dcbr, line 706: I would simplify this example, to focus on the `get_charts` method.
E.g. a simple `Chart2D` with a line plot, adding the chart to the plotter, and showing that `chart is pl.renderer.get_charts()[0]` evaluates to `True`?

pyvista/plotting/renderer.py
kmarchais: Done, thanks!

pyvista/plotting/renderer.py
```diff
@@ -695,6 +695,32 @@ def has_charts(self):  # numpydoc ignore=RT01
...
+           >>> pl.subplot(0, 0)
+           >>> _ = pl.add_mesh(pv.Sphere())
+           >>> pl.subplot(0, 1)
+           >>> _ = pl.add_chart(chart)
+           >>> ax = pl.renderer.charts[0].figure.axes[0]
+           >>> ax.get_lines()[0].set_ydata([6, 5, 4])
+           >>> pl.show()
+
+        """
+        return self._charts
```
dcbr, line 721: We need to check that the renderer has charts and otherwise return an empty list. I would also just return a (new) list of charts instead of the whole `Charts` object, which was meant for internal usage.
E.g.
```suggestion
        return [*self._charts] if self.has_charts else []
```

tests/plotting/test_plotting.py
```diff
@@ -2494,6 +2494,17 @@ def test_chart_matplotlib_plot(verify_image_cache):
     pl.show()
 
 
+def test_get_charts_property():
```
dcbr, line 2497: Remove the `_property` and I would add an extra assert to verify that `chart is pl.renderer.get_charts()[0]`

dcbr (commented):
Thanks for the swift reply and changes. Here's some final comments from my part. Otherwise this looks good to me!

tests/plotting/test_plotting.py
kmarchais: Done, thanks

pyvista/plotting/renderer.py
kmarchais: Done, thanks

tkoyama010:
github-actions preview

pyvista/plotting/renderer.py
```diff
@@ -695,6 +695,24 @@ def has_charts(self):  # numpydoc ignore=RT01
...
+        --------
+        .. pyvista-plot::
+           :force_static:
+
+           >>> import pyvista as pv
+           >>> chart = pv.Chart2D()
+           >>> chart.line([1, 2, 3], [0, 1, 0])
+           >>> pl = pv.Plotter()
+           >>> pl.add_chart(chart)
+           >>> print(chart is pl.renderer.get_charts()[0])  # True
```
tkoyama010, line 711: We need to write output.
```suggestion
           >>> print(chart is pl.renderer.get_charts()[0])
           True
```

dcbr: I think the `print` is not necessary either

dcbr (approved):
If the tests pass, this is good for me. Thanks for the contribution @kmarchais!

tkoyama010:
github-actions preview

dcbr (approved):

PR #5621, opened 2024-02-14
Add core typing to documentation

user27182:
### Overview

Add documentation for the type definitions in `core._typing_core`. 

Specifically:

1. Publish the types in the docs in `Core API -> Typing`
2. Make the `_typing_core` types in docstrings and annotations automatically linkable.
    - Example issue: the docs for [set_array](https://docs.pyvista.org/version/0.43/api/core/_autosummary/pyvista.DataSetAttributes.set_array.html) have a `data : Array` parameter in the docstring, but it is not linkable. This makes it unclear what is meant by `Array`.
    - This is accomplished by adding core types to the `pyvista` namespace. This is so that adding the types to the docs with just the type name (e.g. `Vector` will mean that sphinx will implictly find and link it automatically (no need to explicitly reference `pyvista.core._typing_core.Vector`)

3. Make the `_typing_core` annotations in function signatures more readable.
    - Example issue: since aliases are automatically expanded by sphinx, the annotations for array-like types can take up to 8 lines for a single parameter. See screenshot in https://github.com/pyvista/pyvista/pull/5571#issuecomment-1939197128

The additions in this PR are similar to the configuration for [ColorLike](https://docs.pyvista.org/version/stable/api/utilities/_autosummary/pyvista.ColorLike.html), which is a pyvista type alias that is linkable and is defined formally the docs. (See the `color` param for [ChartBox](https://docs.pyvista.org/version/stable/api/plotting/charts/_autosummary/pyvista.chartbox#pyvista.ChartBox) as an example of its use in docstrings).

PR #5661, opened 2024-02-21
Append 'Like' suffix to core array-like type aliases

user27182:
### Overview

Related to #5621. In that, linkable references to the core array-like type aliases are added to the docs. In order to make it seamless and easy when adding these types to docstrings, that PR adds the type aliases to the `pyvista` namespace so that `Sphinx` can implicitly reference the type aliases. This is similar to what is already done with `pyvista.ColorLike`.

However, one potential issue with adding these types to the `pyvista` namespace is that users may access `pyvista.Array` thinking that this creates an actual `Array` object, and not a type alias object.

This PR appends `Like` to address this:
- `Vector` -> `VectorLike`
- `Matrix` -> `MatrixLike`
- `Array` -> `ArrayLike`

In addition, appending `Like` to the types will make `pyvista` more consistent with other scientific libraries such as NumPy and Pandas which also use the term `ArrayLike`. 
However, it may also create ambiguity since our definitions will differ slightly (e.g. PyVista's ArrayLike doesn't include scalars, Numpy's does), see relevant issue https://github.com/pandas-dev/pandas/issues/41807. Perhaps we should also add a `pv` prefix, e.g. `pvArrayLike`?

### DIFF
--- a/pyvista/core/_typing_core/__init__.py
+++ b/pyvista/core/_typing_core/__init__.py
@@ -1,13 +1,12 @@
 """Type aliases for type hints."""
 
 from ._aliases import (  # noqa: F401
-    Array,
     BoundsLike,
     CellArrayLike,
     CellsLike,
-    Matrix,
+    MatrixLike,
     Number,
     TransformLike,
-    Vector,
+    VectorLike,
 )
-from ._array_like import NumpyArray  # noqa: F401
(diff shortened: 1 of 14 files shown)


tkoyama010:
github-actions preview

tkoyama010 (approved):
I think this is a good suggestion. As for the differ with other libraries, let's think about that after the problem arises, since it doesn't affect the API spec.

PR #5664, opened 2024-02-21
Publish `validation` package docs

user27182:
### Overview

This PR will publish the docs for the `validation` package. Publication should be considered sometime after #5571 is merged. Originally suggested by @darikg here https://github.com/pyvista/pyvista/pull/5586#issuecomment-1935781239.


user27182:
@pyvista-bot preview

user27182:
I think we should publish the docs now. This will enable better review of changes to the docstrings as updates to the package as pushed as part of #5571.

We can un-publish the docs again in case there is a release if we want to. Alternatively, this PR adds a warning the docs about the API being unstable, so we could also leave it there and publish the docs with a release if desired.

user27182:
Another option is to include a 'PyVistaFutureWarning' to warn directly in the code when imported. But I don't think that's necessary though. The package is already marked as private with `pyvista.core._validation`, so I think it should be OK.

tkoyama010 (approved):
LGTM. Thanks!

PR #5673, opened 2024-02-23
Allow subclasses to set new attributes with `no_new_attr` decorator

user27182:
Subclasses of classes with the `no_new_attr` decorator are not currently allowed to set new attributes. E.g.:

``` python
from pyvista.core.utilities.misc import no_new_attr

@no_new_attr
class A:
    ...

class B(A):
    _new_attr_exceptions = ['eggs']
    def __init__(self):
        self.eggs = 'ham'

x = B()
```

will error with `AttributeError: Attribute "eggs" does not exist and cannot be added to type B`

This PR fixes this, allowing subclasses to define new attributes (only if they're in defined in `_new_attr_exceptions`).

This issue was first raised here https://github.com/pyvista/pyvista/issues/5216#issuecomment-1824558766. I'm running into this problem trying to subclass `Actor` as part of #5029


tkoyama010 (approved):
LGTM

PR #5674, opened 2024-02-23
Ignore flake8 E701 and E702 to fix conflict with black

user27182:
Follow-up from #5592. The same issue is popping up in #5673, causing unresolvable `pre-commit` failures


tkoyama010 (approved):
Good catch! Thanks.

PR #5685, opened 2024-02-25
Add `ContourFilter` class to support trame example

tkoyama010:
### Overview

I have recently been very interested in Trame for prototyping applications. The Trame application in the PyVista repository relies on vtk for the filter part of its implementation. I propose to make a class for it in PyVista. Let's have fun pythonic implementing applications together!

### Details

- This proposal also has purpose to discuss if we can add a class for other filters as well.


tkoyama010:
This is not elegant. I have another idea.

Issue #5694, opened 2024-02-27
Expose all DataSet methods in documentation


MatthewFlamm:
### Describe what maintenance you would like added.

A while ago a change was made to not include super class methods in the documentation.  This greatly reduced build time and clutter on the documentation.  However, this means that we need to document all super classes to get the documentation for methods, which isn't currently done.  An example is `_PointSet`, which has a bunch of methods including `remove_cells` that cannot be accessed in the documentation.

One difficulty is that it isn't obvious which super classes are inherited by which final DataSet classes.  We explicitly call this out for Filters: https://docs.pyvista.org/version/stable/api/core/filters.  However, if I go to https://docs.pyvista.org/version/stable/api/core/_autosummary/pyvista.unstructuredgrid, I don't know which other classes methods will be included in `UnstructuredGrid`.  Is there a Sphinx setting to show the inheritance with methods in an understandable way?

MatthewFlamm:
One low tech but effective solution here could be to manually add in links to the relevant filters for each DataSet.  We would add in a standard paragraph to each DataSet docstring like:  `"Filter methods from DataSetFilters, ImageDataFilters are available"`.  We could also do the reverse for each Filter class to reference which DataSets inherit.  This would require maintenance, but this would only happen if a new structure is introduced or a new DataSet is added.

Something like this is in this part of the documentation, but I suspect not many people see this:
https://docs.pyvista.org/version/stable/api/core/filters

njneeteson:
Is there a high tech way to automatically add links? Ideally any class that inherits methods from another class would have a link on its documentation page that indicates what other classes it inherits from and links to its documentation so you can easily see what other methods are available. Something like the dependency graphs in the VTK documentation maybe?

The low tech route seems like a lot of manual effort and moving forward stuff is likely to get missed.

MatthewFlamm:
#5702 is currently adding in this inheritance linking as well as adding in every inherited method onto every class. The inheritance linking is currently problematic because of the way that PyVista has internally structured inheritance.  There could be several levels of inheritance needed to be added.  We'd have to document all the internally used classes, which might create even more confusion.

The manual method would only be work if we change the structure of inheritance, which we could address at that time.  But it certainly increases friction of making changes.  Maybe the best solution is to  include all the inherited methods in the documentation for the DataSets. As long as it doesn't significantly prolong the documentation build times or size.

Issue #5693, opened 2024-02-27
Add Ghost cells handling API to PyVista


MatthewFlamm:
### Describe the feature you would like to be added.

vtk allows for ghosting and blanking cells without modifying the mesh.  We have `_PointSet.remove_cells`, but it changes the mesh.  In fact the entire [example](https://docs.pyvista.org/version/stable/examples/02-plot/ghost-cells.html) is misleading as there should be small, if any, performance benefit to using `remove_cells` over some other method.

We should expose methodology to handle blanking cells the vtk way, e.g. https://www.kitware.com/ghost-and-blanking-visibility-changes/.  This is done through adding a `vtkGhostType` point or cell array.  For now, for expert users, they can simply add this array, but it isn't obvious to PyVista users how this works.  See #5689.  There should be a PyVista way to do this.

### Pseudocode or Screenshots

On possible API for managing this.  The first is more similar to how vtk operates IMO.  It is more flexible if users also want to access the other ghost cell types.
```py
mesh.ghost_cells  # mostly an alias for mesh["vtkGhostType"]
mesh.ghost_cells[0] = pyvista.module.path.GhostCellType.HIDDEN_CELL
```

Another possible API that is more PyVista-like IMO.  This would be way more complex if users want to access other ghost cell types.
```
mesh.mark_hidden_cells([0, 1, 5, 10])  # adds HIDDENCELL entry for cell #'s
mesh.mark_hidden_cells([1, 5], remove=True)  # removes HIDDENCELL entry for cell #'s
```

It could also be a combination of both, or some other API entirely.

MatthewFlamm:
A manual level example showing the differences here:

```py
import pyvista as pv
import numpy as np
from pyvista import examples

vol = examples.load_channels()
mesh = vol.cast_to_unstructured_grid()
print(mesh.n_cells)

pl = pv.Plotter()
pl.add_mesh(mesh.outline())
ghost_array = np.zeros(mesh.n_cells, dtype=np.uint8)
np.place(ghost_array, mesh.cell_centers().points[:, 0] > 120, 32)
mesh["vtkGhostType"] = ghost_array

pl.add_mesh(mesh, clim=[0, 4])
pl.show()
print(mesh.n_cells)  # same number of cells
mesh.RemoveGhostCells()  # this is what `remove_cells` does
print(mesh.n_cells)  # reduced number of cells
```

prints

```txt
6250000
6250000
3000000
```

and figure

[...]

PR #5695, opened 2024-02-27
Fix ghost-cells example

MatthewFlamm:
### Overview

The `remove_cells` filter probably used to default to `inplace=True`, but the documentation was never updated. See https://docs.pyvista.org/version/stable/examples/02-plot/ghost-cells.html

### Details

#5693 is opened to track whether there should be a better API for Ghosting cells.  Also `remove_cells` filter isn't in the documentation. This is tracked in #5694.


tkoyama010:
github-actions preview

MatthewFlamm:
Looks good on generated docs 
image

tkoyama010 (approved):

PR #5702, opened 2024-02-28
Expose `DataSet` docs

### LINKED ISSUE #5694: Expose all DataSet methods in documentation

user27182:
Possiblly fix #5694 

The method of adding the class to `class.rst` worked previously as part of #5029 to expose the abstract docs of `Prop3D` so it could be referenced for both `Actor` and `AxesActor` child classes.

Let's see what this does for `DataSet`.

### DIFF
--- a/doc/source/_templates/autosummary/class.rst
+++ b/doc/source/_templates/autosummary/class.rst
@@ -3,6 +3,7 @@
 .. currentmodule:: {{ module }}
 
 .. autoclass:: {{ objname }}
+   :show-inheritance:
 
    {% block methods %}
 
--- a/doc/source/_templates/autosummary/class.rst
+++ b/doc/source/_templates/autosummary/class.rst
@@ -13,7 +13,7 @@
       :toctree:
    {% for item in methods %}
       {% if not item in skipmethods %}
-        {% if name == 'Plotter' or item not in inherited_members %}
+        {% if name == 'Plotter' or item in inherited_members %}
           {{ name }}.{{ item }}
         {% endif %}
       {% endif %}
@@ -28,7 +28,7 @@
    .. autosummary::
       :toctree:
    {% for item in attributes %}
-      {% if name == 'Plotter' or item not in inherited_members %}
+      {% if name == 'Plotter' or item in inherited_members %}
         {% if item.0 != item.upper().0 %}
           {{ name }}.{{ item }}
         {% endif %}
--- a/doc/source/_templates/autosummary/class.rst
+++ b/doc/source/_templates/autosummary/class.rst
@@ -13,7 +13,7 @@
       :toctree:
    {% for item in methods %}
       {% if not item in skipmethods %}
-        {% if name == 'Plotter' or item not in inherited_members %}
+        {% if name in ['Plotter', 'DataSet'] or item not in inherited_members %}
           {{ name }}.{{ item }}
         {% endif %}
       {% endif %}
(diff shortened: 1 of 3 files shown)


user27182:
github-actions preview

user27182:
CC @MatthewFlamm

user27182:
I don't think this worked. No new methods showing.

0.43
image


This PR
image

MatthewFlamm:
It worked for `DataSet` https://65deeb78a493a1ba242fe891--meek-duckanoo-d2f5ec.netlify.app/api/core/_autosummary/pyvista.dataset#pyvista.DataSet

Screenshot_20240228-065624

MatthewFlamm:
This is the method that the documentation used previously, it used all inherited methods in all DataSets, but it led to very long build times and messy documentation.

For example if you searched for `threshold` you would find `UnstructuredGrid.threshold` and `PolyData.threshold` and ...

I don't think this idea is possible in Sphinx, but maybe there is a great plugin.  I'd prefer that the methods are all associated with their own class, but inherited methods were linked in sub classes.

user27182:
> I'd prefer that the methods are all associated with their own class, but inherited methods were linked in sub classes.

What about exposing `:inherited-members:` from `autodoc` ?
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

[621e6b0](https://github.com/pyvista/pyvista/pull/5702/commits/621e6b0e779289a764655e93c20e0820012e5b62) tries to add this

user27182:
github-actions preview

MatthewFlamm:
Not sure what the solution here is, but another suggestion: https://www.sphinx-doc.org/en/master/usage/extensions/inheritance.html

user27182:
github-actions preview

user27182:
Marked as draft since it's a bit of a trial and error

user27182:
Looks like this worked, many methods listed now
https://65dfb1b1016abe7c427345fb--meek-duckanoo-d2f5ec.netlify.app/api/core/_autosummary/pyvista.unstructuredgrid
image

This change impacts pretty much all docs though, not just `DataSet`.

user27182:
Also, build time was 1h4m, doesn't seem any longer than usual, so not really any impact there with this change

user27182:
> Not sure what the solution here is, but another suggestion: https://www.sphinx-doc.org/en/master/usage/extensions/inheritance.html
> 
> 
> 
> Edit:  A simpler suggestion is `:show-inheritance:`

We could try using this, and also give the inherited methods their own sections. Something like:

```
UnstructuredGrid
-Inheritance Diagram
-Methods
-Inherited Methods
```

user27182:
github-actions preview

user27182:
> Edit: A simpler suggestion is `:show-inheritance:`

Example from Deployed on https://65dff7e140f26caa45a20e47--meek-duckanoo-d2f5ec.netlify.app
image

This is helpful, but will also show inheritance from undocumented classes and vtk classes, and I think only shows immediate parents (e.g. may not show `DataSet` explicitly for some classes)

I think maybe just separating class-specific methods and properties from inherited ones, e.g. with different sections headings, may be good enough.

PR #5712, opened 2024-02-29
Dropping isort and move to lint.isort rule in ruff

tkoyama010:
### Overview

Dropping `isort` and moving to `lint.isort` rule in `ruff`.

Close #6137

### Details

- See https://docs.astral.sh/ruff/rules/#isort-i.


tkoyama010:
pre-commit.ci autofix

user27182:
I'm not sure how much overlap this tool has with `isort`, but this seems worth considering :

https://github.com/asottile/reorder-python-imports

Its main goal is to reduce merge conflicts.

user27182:
> I'm not sure how much overlap this tool has with `isort`, but this seems worth considering :
> 
> https://github.com/asottile/reorder-python-imports
> 
> Its main goal is to reduce merge conflicts.

For context, I stumbled upon `reorder-python-imports` trying to look for a solution to https://github.com/pyvista/pyvista/pull/5621#issuecomment-1970082659 . In this case, `reorder-python-imports` could be used to automatically add `from future import __annotations__` to every file in the code base to support the type aliases for the docs. The config would be similar to this: https://github.com/asottile/pyupgrade/blob/df17dfa3911b81b4a27190b0eea5b1debc7ffa0a/.pre-commit-config.yaml#L18-L20

I plan on opening a new PR with `reorder-python-imports` to support this. And so if this works and the PR is accepted, then perhaps we may want to also use `reorder-python-imports` instead of `isort`. Just an idea though, I am not very familiar with these linting tools.

tkoyama010:
> Just an idea though, I am not very familiar with these linting tools.

No problem. We will review it.

user27182:
> > I'm not sure how much overlap this tool has with `isort`, but this seems worth considering :
> > https://github.com/asottile/reorder-python-imports
> > Its main goal is to reduce merge conflicts.
> 
> For context, I stumbled upon `reorder-python-imports` trying to look for a solution to [#5621 (comment)](https://github.com/pyvista/pyvista/pull/5621#issuecomment-1970082659) . In this case, `reorder-python-imports` could be used to automatically add `from future import __annotations__` to every file in the code base to support the type aliases for the docs. The config would be similar to this: https://github.com/asottile/pyupgrade/blob/df17dfa3911b81b4a27190b0eea5b1debc7ffa0a/.pre-commit-config.yaml#L18-L20
> 
> I plan on opening a new PR with `reorder-python-imports` to support this. And so if this works and the PR is accepted, then perhaps we may want to also use `reorder-python-imports` instead of `isort`. Just an idea though, I am not very familiar with these linting tools.

I looked into using `reorder-python-imports`. It is fundamentally incompatible with black (`black` adds a new line after docstrings, `reorder-python-imports` insists on removing it), see summary about this issue here: https://github.com/psf/black/issues/4175#issuecomment-2036987251

The main goal of my earlier suggestion was to add `from __future__ import annotations` to every file. This can be accomplished with `ruff` and the proposed changes from this PR by adding this [option](https://docs.astral.sh/ruff/settings/#lint_isort_required-imports):

```
[tool.ruff.lint.isort]
required-imports = ["from __future__ import annotations"]
```

As an aside, in case there is a desire to reduce merge conflicts (which is another reason to use `reorder-python-imports`),  the `force-single-line` [option](https://docs.astral.sh/ruff/settings/#lint_isort_force-single-line) may also be used.

So, there is no need to make use of  `reorder-python-imports` as a pre-commit hook. 

+1 for this PR and dropping `isort` and using `ruff` instead.

tkoyama010:
pre-commit.ci autofix

pyproject.toml
```diff
@@ -404,3 +405,9 @@ allow-dict-calls-with-keyword-arguments = true
     # https://github.com/pyvista/pyvista/pull/6019
     "B018",
 ]
+
+[tool.ruff.lint.isort]
+# Sort by name, don't cluster "from" vs "import"
+force-sort-within-sections = true
+# Combines "as" imports on the same line
+combine-as-imports = true
```
tkoyama010, line 413:
```suggestion
combine-as-imports = true
required-imports = ["from __future__ import annotations"]
```

pyproject.toml
```diff
@@ -404,3 +405,10 @@ allow-dict-calls-with-keyword-arguments = true
...
+required-imports = ["from __future__ import annotations"]
```
tkoyama010, line 414:
```suggestion
# https://github.com/pyvista/pyvista/pull/5712
required-imports = ["from __future__ import annotations"]
```

pyproject.toml
```diff
@@ -404,3 +405,11 @@ allow-dict-calls-with-keyword-arguments = true
...
     "B018",
 ]
+
+[tool.ruff.lint.isort]
+# Sort by name, don't cluster "from" vs "import"
+force-sort-within-sections = true
+# Combines "as" imports on the same line
+combine-as-imports = true
+# https://github.com/pyvista/pyvista/pull/5712
+required-imports = ["from __future__ import annotations"]
```
tkoyama010, line 415:
```suggestion
required-imports = ["from __future__ import annotations"]
# https://github.com/pyvista/pyvista/pull/5712
force-single-line = true
```

tkoyama010:
pre-commit.ci autofix

user27182:
The [documentation build](https://github.com/pyvista/pyvista/actions/runs/9278295702) for https://github.com/pyvista/pyvista/pull/5712/commits/1fc8cfcfb9f9e76f5dab3e5847372afbec3a469b took about 2h20 mins. Usually it's about 1h10 mins. Not sure how the caching works for the build, maybe it took extra long because pretty much every file in the repo got modified, so had to be completely re-built from scratch? Hopefully it's not a permanent increase in build time.

tkoyama010:
> The [documentation build](https://github.com/pyvista/pyvista/actions/runs/9278295702) for [1fc8cfc](https://github.com/pyvista/pyvista/commit/1fc8cfcfb9f9e76f5dab3e5847372afbec3a469b) took about 2h20 mins. Usually it's about 1h10 mins. Not sure how the caching works for the build, maybe it took extra long because pretty much every file in the repo got modified, so had to be completely re-built from scratch? Hopefully it's not a permanent increase in build time.

The cause is that the directory under the `examples` directory was rerun. After the merge, a new cache will be created and we expect the execution time to return to normal.

tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

tkoyama010:
The modification of `pyvista/trame` is an automatic modification by ruff, so the code-owner check is skipped and merged.

PR #5715, opened 2024-03-01
Import `pyvista` plotting modules statically to enable auto-complete for type checkers

### LINKED ISSUE #5708: Help static code analysis tools find submodule information

Someone recently posted a question in the Python Discord server asking why they were unable to see tooltip information (type hints, docstings, etc.) for the `plotting` submodule in their text editor (Visual Studio Code).

For example, in VSC, if you hover over `pv.Plotter`, with the control key depressed, you should see information about the class, such as its parameters and docstring:
```py
import pyvista as pv
pv.Plotter()
```

I believe the lazy import mechanism is confusing the text editor's static analysis tool. This seemed to be the case with both PyLance and MyPy.

Could you consider adding the following code (or something similar) to the `__init__.py` file to help the type checkers find the information they need:
```py
import typing
if typing.TYPE_CHECKING:
    from pyvista.plotting import *
```

The constant [`typing.TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) is always assumed to be `True` by type checkers, but is always `False` at runtime.

If there's a solution to this problem on the user's side, let me know so I can pass that information on to the person who asked the question.

_No response_

### Pseudocode or Screenshots

_No response_

user27182:
- Fix #5708
- ~~Possibly fix issues with type alias expansions in the docs for `ArrayLike` aliases as well, see https://github.com/pyvista/pyvista/pull/5621#issuecomment-1970082659~~  This issue will be addressed separately.


user27182:
github-actions preview

user27182:
github-actions preview

user27182:
github-actions preview

pyvista/__init__.py
```diff
@@ -60,6 +63,17 @@
 MAX_N_COLOR_BARS = 10
 
 
+# Import all modules for type checkers and linters
+if TYPE_CHECKING:  # pragma: no cover
+    import demos
+    import ext
+    from plotting import *
```
MatthewFlamm, line 70: Does this mean that anyone using an IDE with auto complete will now need to import GL libraries?

user27182: No, the type checkers do not actually execute any code. So this will not force any imports to actually be imported.

If the libraries are not available, then linters may flag it as an error similar to this one in PyVista's `tools.opacity_transfer_function`, where `scipy` is imported by is not installed:
image

But this error has no impact on any actual code execution. 

Similarly, if `mypy` is analyzing the code and cannot find the import, then the default assumption will be that the output type for these imports will be `Any`. I suppose this could be an issue if a downstream package had a very strict `mypy` config that disallowed missing imports or `Any` types, but if this was the case then there would also be many other errors generated anyway because most of the `pyvista` code base does not have annotations, and so already defaults to `Any`.

At most, if this actually does pose a problem for non-GL libraries somehow, I think adding a `type: ignore` for these lines somewhere would likely resolve it. Alternatively, would we could encase the imports in a `try-except` block, similar to the screenshot above for the `scipy, which I think may have a similar effect (I think `mypy` cannot fully resolve exceptions statically, and will assume variables have `Any` type).

user27182:
LGTM.

For testing, I tried this with PyCharm locally as well as with CodeSpaces. Writing something like `pv.Actor` previously would not auto-complete. But, now it does.
@lxnn do you mind testing the changes from this PR locally to see that it works for you too?

If we want to include an actual test case for this, we could do something like this:

``` python
import pyvista as pv

reveal_type(pv.Actor)  
# Without this PR: Revealed type is "Any"
# With this PR: Revealed type is "def (mapper: Any =, prop: Any =, name: Any =) -> pyvista.plotting.actor.Actor"
```

We don't currently have any testing mechanisms for static testing with mypy, but #5226 adds a custom testing suite for checking `reveal_type` with mypy which we could leverage.

lxnn:
@user27182 

I don't mind. It appears to be working.

PyLance:
PyLance

MyPy:
MyPy


Thanks for fixing this issue.

tkoyama010 (approved):

PR #5719, opened 2024-03-02
Update `mypy` config to use error codes in `type: ignore` comments

user27182:
See [mypy docs ](https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-type-ignore-include-an-error-code-ignore-without-code) for details


tkoyama010 (approved):
LGTM

PR #5747, opened 2024-03-08
Allow setting and getting string scalar field data

user27182:
### Overview

Setting string scalars for field data is not currently allowed:
``` python
import pyvista as pv
mesh = pv.Sphere()
mesh.field_data['name'] = 'sphere'
```

Will error with:
```
File "~/pyvista/core/datasetattributes.py", line 736, in _prepare_array
  tmparray.fill(data)
ValueError: could not convert string to float: 'sphere'
```

This PR fixes this and implements a `str` -> `str` scalar mapping.

This allows `DataSet` objects to be used like a `dict[str, str]` object, which is useful for storing string metadata directly with the dataset.

### Details

Normally, scalar inputs are always reshaped to 1-d arrays, e.g.:
``` python
mesh.field_data['number'] = 1
mesh['number']  # output is: pyvista_ndarray([1.])
```

However, in this PR, string scalars are not reshaped. This allows for a `str` -> `str` scalar mapping, e.g.:

``` python
mesh.field_data['name'] = 'sphere'
mesh['name']  # output is: 'sphere', not `pyvista_ndarray(['sphere'])`
```

This PR also allows setting string field data implicitly, e.g. the following is allowed:

``` python
mesh['name'] = 'sphere'
```

This is notable because by comparison, setting field data implicitly is not usually allowed, e.g. trying to set a scalar number:
``` python
mesh = pv.Sphere()
mesh['number'] = 1  
```
Will error with:
```
ValueError: Number of scalars (1) must match either the number of points (842) or the number of cells (1680).
```


tkoyama010 (approved):
LGTM

PR #5757, opened 2024-03-11
Add Path type hint for PolyData

kmarchais:
### Overview

This PR adds the `pathlib.Path` type hint to the `var_inp` parameter of the `PolyData` class.

#### Details

It is already compatible since, the first type check is already taking it into account with `if isinstance(var_inp, (str, pathlib.Path)):`.

However, in the following code, mypy throws an error when a path is provided because the type hint is missing. With the type hint added, there is no more error.

```python
import pyvista as pv

pd = pv.PolyData(Path("mesh.vtk"))
```


tkoyama010 (approved):

tkoyama010:
Thanks!

PR #5776, opened 2024-03-15
Allow reading DICOM stack from directory with `pv.read()`

user27182:
Reading a DICOM stack from a directory is not currently allowed:

``` python
import pyvista as pv
from pyvista import examples as ex
directory = ex.download_dicom_stack(load=False)
pv.read(directory)  # OSError: This file was not able to be automatically read by pyvista.
```

This PR fixes this.

### DIFF
--- a/tests/core/test_reader.py
+++ b/tests/core/test_reader.py
@@ -25,7 +25,7 @@ skip_windows = pytest.mark.skipif(os.name == 'nt', reason='Test fails on Windows
 def test_get_reader_fail(tmp_path):
     with pytest.raises(ValueError):  # noqa: PT011
         pv.get_reader("not_a_supported_file.no_data")
-    match= "`pyvista.get_reader` does not support reading from directory:\n\t"
+    match = "`pyvista.get_reader` does not support reading from directory:\n\t"
     with pytest.raises(ValueError, match=match):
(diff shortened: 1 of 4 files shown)


tests/core/test_reader.py
```diff
@@ -22,9 +22,12 @@
 skip_windows = pytest.mark.skipif(os.name == 'nt', reason='Test fails on Windows')
 
 
-def test_get_reader_fail():
+def test_get_reader_fail(tmp_path):
     with pytest.raises(ValueError):  # noqa: PT011
         pv.get_reader("not_a_supported_file.no_data")
+    msg = "`pyvista.get_reader` does not support reading from directory:\n\t"
+    with pytest.raises(ValueError, match=msg):
+        pv.get_reader(str(tmp_path))
```
tkoyama010, line 30: nit: It is better to match the name of the argument to be assigned with the name of the variable.
```suggestion
    match= "`pyvista.get_reader` does not support reading from directory:\n\t"
    with pytest.raises(ValueError, match=match):
        pv.get_reader(str(tmp_path))
```

tkoyama010 (changes requested):

user27182:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM

PR #5782, opened 2024-03-16
Fix Actor error of prop argument in initialization

tkoyama010:
### Overview

Fix Actor error of prop argument in initialization.

### Details

- None


tkoyama010:
@pyvista-bot LGTM

PR #5786, opened 2024-03-17
Remove tests for deprecated examples

user27182:
Follow-up from #5785

These tests fail locally, e.g. when running `` pytest --test_downloads -v -k test_download_mars`` , but were somehow missed by the CI tests.


tkoyama010 (approved):
Thank you. This test is not to be run by CI because it requires a download data.

Issue #5784, opened 2024-03-16
Add file and instance metadata to documentation for meshes in `pyvista.examples`


user27182:
### Describe the feature you would like to be added.

There are hundreds of meshes available for use and/or download in `pyvista.examples`. However, when trying to find a specific kind of mesh to use for testing, generating documentation, or experimentation, it's difficult to know which mesh may be suitable for the task. E.g. perhaps a point cloud or surface mesh is needed, either with many points/cells or with few. Maybe several image volumes are needed, or maybe 2D images for textures. It's also helpful to know if the example includes array data. Since some downloadable meshes may be hundreds of MB in size, it would be helpful to know the file size beforehand.

Ideally, this feature would:

- Add file info to the documentation (e.g. file extension, file size)
- Add mesh instance info to the documentation (e.g. mesh type, num cells, num points, num arrays)
- Summarize this info in a single table for easy lookup/reference (instead of inside individual docstrings)
- BONUS: Add a small preview image of the example in the table

### Links to VTK Documentation, Examples, or Class Definitions.

A custom table could be generated using similar code in [make_tables.py](https://github.com/pyvista/pyvista/blob/3068f5c54de10da55fa16824939666c60bf7f844/doc/source/make_tables.py#L5), for example, which is already used to generate the [Color scheme table](https://docs.pyvista.org/version/stable/api/plotting/charts/plot_color_schemes.html#id1)

Issue #5775, opened 2024-03-15
Add missing `load` parameter to some example downloads


user27182:
### Describe the feature you would like to be added.

The following downloads are all missing a `load` parameter.

1. [download_tetra_dc_mesh](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_tetra_dc_mesh)
2. [download_sky_box_cube_map](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_sky_box_cube_map)
3. [download_cubemap_park](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_cubemap_park)
4. [download_cubemap_space_4k](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_cubemap_space_4k)
5. [download_cubemap_space_16k](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_cubemap_space_16k)
6. [download_osmnx_graph](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_osmnx_graph)
7. [download_dikhololo_night](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_dikhololo_night.html)

Since all the other downloads have this option to allow returning just the file(s), it is surprising that the listed downloads are missing this feature. Adding this option would make the example download API more consistent.

Issue #5774, opened 2024-03-15
Missing file reader for NumPy file types (e.g. `.npy`)


user27182:
### Describe the bug, what's wrong, and what you expected.

There are pyvista examples, e.g. [`download_cloud_dark_matter`](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_cloud_dark_matter.html#pyvista.examples.downloads.download_cloud_dark_matter )which return a `.npy` file. Using `pv.read` on these files will fail:
``` python
import pyvista as pv
from pyvista import examples as ex
file = ex.download_cloud_dark_matter(load=False)
pv.read(file)  # OSError: This file was not able to be automatically read by pyvista.
```
I would have expected that using `pv.read` on the returned file would succeed.

In the example's source code, the files are loaded with:
https://github.com/pyvista/pyvista/blob/3d6e327a7c43d9035ad54de936228f3ac7c5a88a/pyvista/examples/downloads.py#L4807

Perhaps a new reader class should be made to handle loading `.npy` and `npz` files?

### Steps to reproduce the bug.

See above.

### System Information

```shell
--------------------------------------------------------------------------------
                OS : Darwin
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
               RAM : 16.0 GiB
       Environment : Python
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M2
       GPU Version : 4.1 Metal - 83.1
  MathText Support : True

  Python 3.12.2 (v3.12.2:6abddd9f6a, Feb  6 2024, 17:02:06) [Clang 13.0.0
  (clang-1300.0.29.30)]

           pyvista : 0.44.dev0
               vtk : 9.3.0
             numpy : 1.26.4
        matplotlib : 3.8.2
            scooby : 0.9.2
             pooch : 1.8.0
            pillow : 10.2.0
           imageio : 2.33.1
           IPython : 8.21.0
          colorcet : 3.0.1
           cmocean : 3.1.3
        ipywidgets : 8.1.2
              tqdm : 4.66.2
            meshio : 5.3.5
    pytest_pyvista : 0.1.8
             trame : 3.5.2
      trame_client : 2.16.1
      trame_server : 2.17.2
         trame_vtk : 2.8.5
     trame_vuetify : 2.4.2
      nest_asyncio : 1.6.0
--------------------------------------------------------------------------------
```

PR #5787, opened 2024-03-17
Refactor and standardize loading examples from `downloads.py`

user27182:
### Overview

This PR refactors and standardizes the downloading and loading mechanisms for the examples in `downloads.py`. It is part of #5784.

No public API changes are introduced. A few `See Also` sections in some docstrings are updated (mostly limited to images which also have a corresponding `texture` download).

### Details

Currently, the downloads API is inconsistent:

1. some examples do not have a `load` parameter (see #5775), and so the filepaths are not available for these examples
2. some examples return a file which cannot be directly read by pyvista (#5774)
3. the list of file(s) returned when `load=False` does not include metafiles (e.g. only the `.raw` data file is returned for examples which also use a `.mhd` header file)

These inconsistencies make it difficult or impossible to analyze the file metadata and instance metadata needed to implement #5784. For example, how should file size be computed if not all downloaded files are returned? Also, how can a preview image of the file(s) be generated if it's not possible to read or load the file directly?

With this PR, these issues are resolved by exposing and compartmentalizing all of the downloading, reading, and loading functions used by all examples with a new, private API in `_example_file.py` E.g.:

Specialized reading (e.g. for `.npy` files) and post-processing functions (e.g. setting scalars) are fully encapsulated, e.g. for the `download_cloud_dark_matter` example:
``` python
from pyvista.examples import downloads
ex = downloads._example_cloud_dark_matter
ex.download()
# '/Users/user/Library/Caches/pyvista_3/point-clouds/findus23/halo_low_res.npy'

ex.load()
# PointSet (0x15e335420)
#  N Cells:    0
#  N Points:   32314
#  X Bounds:   7.451e+01, 7.892e+01
#  Y Bounds:   1.616e+01, 2.275e+01
#  Z Bounds:   8.900e+01, 9.319e+01
#  N Arrays:   0
```

All metafiles are also available for private access, e.g. for the `download_frog` example, two files are downloaded:
``` python
ex = downloads._example_frog
ex.download()
# ('/Users/user/Library/Caches/pyvista_3/froggy/frog.mhd',
#  '/Users/user/Library/Caches/pyvista_3/froggy/frog.zraw')
```

Calling `download_frog` publically, however, only one file is returned (i.e. the current public API remains the same):
``` python
downloads.download_frog(load=False)
# '/Users/user/Library/Caches/pyvista_3/froggy/frog.mhd'
```

This standardized representation gives full (private) access to all downloaded files and loading methods needed to properly analyze the file and instance metadata to implement #5784  

EDIT: Some metadata features are also added:
``` python
ex = downloads._example_tetra_dc_mesh
ex.download()
# ('/Users/user/Library/Caches/pyvista/dc-inversion.zip.unzip/dc-inversion/mesh-forward.vtu',
#  '/Users/user/Library/Caches/pyvista/dc-inversion.zip.unzip/dc-inversion/mesh-inverse.vtu')

# get unique file extensions
ex.extension
# '.vtu

# sum of all file sizes on disk
ex.total_size
# '58.0 MiB'

# get vtk readers for all loadable files
ex.reader
(XMLUnstructuredGridReader('/Users/user/Library/Caches/pyvista_3/dc-inversion.zip.unzip/dc-inversion/mesh-forward.vtu'),
 XMLUnstructuredGridReader('/Users/user/Library/Caches/pyvista_3/dc-inversion.zip.unzip/dc-inversion/mesh-inverse.vtu'))

### CODE REVIEW

**Comment by user27182** [comment]
Might be worth considering merging #5718 before this one.

There are approx. 160 `download` examples affected by this PR. So, it has a potential high impact to introduce a regression in some some of the examples. During local testing I noticed many downloads have explicit unit tests in `test_download_files.py`, but some do not. Some examples may only have indirect coverage through `test_plotting`, or through docstrings in the codebase which use them.

Most examples were very straightforward to refactor, and are very unlikely to regress from these changes, e.g.:

``` diff
def download_cow(load=True):
-     return _download_and_read('cow.vtp', load=load)
+     return _download_example(_example_cow, load=load)
+ _example_cow = _SingleFileDownloadableLoadable('cow.vtp')
```

However, some examples with multiple files do some post-processing, e.g. to remove or add scalars. I tried my best to make as few changes as possible, but for the complex cases there is a risk that the refactoring may change the example if there are no tests ensuring that the post-processing modifications were applied correctly.

pyvista/examples/_example_loader.py
```diff
@@ -0,0 +1,307 @@
...
+
+    def __init__(
+        self,
+        filename: str,
+        read_func: Optional[
+            Callable[[str], pyvista.DataSet | pyvista.Texture | NumpyArray[Any]]
+        ] = None,
+        load_func: Optional[Callable[[pyvista.DataSet], Any]] = None,
+    ):
+        # TODO: Refactor filename stuff into parent class
```
tkoyama010, line 63: Are you going to resolve TODO in this PR or keep it?

user27182: Was just working on that, should be resolved by https://github.com/pyvista/pyvista/pull/5787/commits/9dcfa421f4c8dd4ce5445e42cfd7cde4b1b63dde

user27182:
The refactoring part of this PR is complete as of https://github.com/pyvista/pyvista/pull/5787/commits/d19f59c92004e78d9c8c9a4baa048b420a2e20a7

The 3 latest commits after that are technically adding new (private) features to extract the file meta data. 

I'll update the main PR comment soon with these additional details. But otherwise, this is ready for review. (I won't add any other changes/features)

tkoyama010:
@pyvista-bot preview

pyvista/examples/_example_loader.py
```diff
@@ -0,0 +1,504 @@
...
+        """...."""
+
+
+@runtime_checkable
+class _Loadable(Protocol):
+    """...."""
+
+    @abstractmethod
+    def load(self) -> Any:
+        """...."""
```
tkoyama010, line 78: @user27182
To understand the code, I added a docstring("""....""") to the section I want you to add. Please add there the intent of adding the class and method for our understanding.

tkoyama010 (changes requested):

pyvista/examples/_example_loader.py
```diff
@@ -0,0 +1,538 @@
...
+#   1. classes for single file inputs: T -> T
+#   2. classes for multi-file inputs: (T, ...) -> (T, ...)
+_FilePropStrType_co = TypeVar('_FilePropStrType_co', str, Tuple[str, ...], covariant=True)
+_FilePropIntType_co = TypeVar('_FilePropIntType_co', int, Tuple[int, ...], covariant=True)
+
+
+class _FileProps(Protocol[_FilePropStrType_co, _FilePropIntType_co]):
+    @property
+    @abstractmethod
+    def filename(self) -> _FilePropStrType_co:
```
user27182, line 62: I think the name `filename` is a bit misleading as it implies just the name, but the full path is actually returned. Also, the path can be a folder path. This should probably  be renamed to `path`, which is consistent with the property name for the reader class 
https://docs.pyvista.org/version/stable/api/readers/_autosummary/pyvista.basereader.path#pyvista.BaseReader.path

pyvista/examples/_example_loader.py
```diff
@@ -0,0 +1,538 @@
...
+        read function (e.g. :func:`pyvista.read_texture` for textures). The function
+        must have the filename as the first argument and should return a dataset.
+        If default arguments are required by your desired read function, consider
+        using :class:`functools.partial` to pre-set the arguments before passing it
+        as an argument to the loader.
+
+    load_func
+        Specify the function used to load the file. Defaults to `None`. This is typically
+        used to specify any processing of the dataset after reading. The load function
+        typically will accept a dataset as an input and return a dataset.
```
user27182, line 167: `read_func` should be changed to `reader`, and only allow setting to a `pyvista.BaseReader` class. Any custom file reader functions should be lumped into the `load_func` param instead. This way, instead of simply calling `pyvista.read` on the file, we can instead do `pyvista.get_reader` first in `__init__` and also set the reader as the loader's `reader` property. The default `read_func` would then become `self.reader.read()`.

user27182 (commented):
A few comments about key things I think could still be improved. I'll leave it up to reviewers to decide if these should be implemented prior to approval, or if making the changes in separate PR is fine.

user27182:
The last 3 commits were added because I ran into an issue loading the [download_can](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_can.html) example locally. There is no existing test coverage for it.

The commits include a new 'load_and_merge' feature and tests to make sure the merging option works. But, I am still not able to load the example locally because this example is deprecated for VTK > 9.1 which I can't install.

The docstring example is also missing the image from the call to `plot` (likely due to the deprecated VTK issue)  and  has been replaced with [download_can_crushed_hdf](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_can_crushed_hdf). So, I think this example should be deprecated.

pyvista/examples/_example_loader.py
tkoyama010: +1 to `path`.

tkoyama010 (changes requested):

pyvista/examples/_example_loader.py
user27182: Great! This change is implemented already in  https://github.com/pyvista/pyvista/pull/5814, if it's ok, we can leave it as `filename` here and address this as part of that PR.

pyvista/examples/_example_loader.py
user27182: This is a non-trivial fix, and I'm not convinced the effort is entirely worth, especially if the current `_example_loader.py` api is kept fully private long-term.

The main motivation for having the `reader` property is actually to extract the reader's `type` as part of the metadata extraction. So, it doesn't really matter much if the `reader` instance isn't the actual reader used when `read_func` is called. So, perhaps we can simply make the property private (i.e. rename to  `_reader` and only implement the API change from `reader_func` -> `reader`  if there is a need for this feature in the future.

I can make the change of `reader` -> `_reader` in #5814.

tkoyama010 (dismissed):
@user27182 Thanks for writing the docstring, it makes it much easier to understand. Can we merge this?

user27182:
>  Can we merge this?

LGTM

tkoyama010 (approved):

PR #5819, opened 2024-03-25
Add VisualPIC as external example of PyVista

tkoyama010:
### Overview

Add the [VisualPIC](https://github.com/AngelFP/VisualPIC) by @AngelFP as an external example of PyVista.

image

### Details

- None


tkoyama010:
@pyvista-bot preview

PR #1478, opened 2021-07-01
Do not return cpos by default

akaszynski:
This PR resolves #1467 by addressing the discussion in #1457 by not returning the camera position by default.

Also, fixes an issue that has been bothering me for our docs.  The sidebar is offset and removing ``html_show_sourcelink = False`` fixes it.  This is important as we make our API documentation one method per page and the sidebar gets clobbered without this fix.

PR #5831, opened 2024-03-28
Drop `param` installation as `panel` was dropped

tkoyama010:
### Overview

Drop `param` installation as `panel` was dropped.

### Details

- See #1478 .


tkoyama010:
@pyvista-bot LGTM

PR #5836, opened 2024-03-31
Checks for `if` branches with identical arm bodies

tkoyama010:
### Overview

Checks for `if` branches with identical arm bodies.

### Details

- https://docs.astral.sh/ruff/rules/if-with-same-arms/


tkoyama010:
@pyvista-bot LGTM

PR #5874, opened 2024-04-07
Fix ambiguous unicode character docstring

tkoyama010:
### Overview

Fix ambiguous unicode character docstring.

### Details

- See https://docs.astral.sh/ruff/rules/ambiguous-unicode-character-docstring/ .


tkoyama010:
@pyvista-bot LGTM

PR #5875, opened 2024-04-07
Fix unnecessary return None of flake8-return lint

tkoyama010:
### Overview

Fix unnecessary return None of flake8-return lint.

### Details

- None


tkoyama010:
@pyvista-bot LGTM

PR #5879, opened 2024-04-07
Fix suppressible exception from flake8-simplify

tkoyama010:
### Overview

Fix suppressible exception from flake8-simplify.

### Details

- See https://docs.astral.sh/ruff/rules/suppressible-exception/.


tkoyama010:
@pyvista-bot LGTM

PR #5880, opened 2024-04-07
Fix verbose raise derived from tryceratops linter

tkoyama010:
### Overview

Fix verbose raise derived from tryceratops linter.

### Details

- See https://docs.astral.sh/ruff/rules/verbose-raise/.


tkoyama010:
@pyvista-bot LGTM

PR #5881, opened 2024-04-07
Fix if else block instead of if exp (SIM108 rule)

tkoyama010:
### Overview

Fix if else block instead of if exp (SIM108 rule).

### Details

- See https://docs.astral.sh/ruff/rules/if-else-block-instead-of-if-exp/.


tkoyama010:
@pyvista-bot LGTM

PR #5882, opened 2024-04-07
Trame UI: add parallel projection toggle

banesullivan:
When rendering server-side, we may want to quickly toggle on/off parallel projection. These changes add a button toggle to the trame UI bar when the mode is server-rendering.

@jourdain, is it possible for VTK.js and trame-vtk to sync the parallel projection mode so that we can also do this with client-side rendering?

*edit*: it appears parallel projection is indeed supported in VTK.js [here](https://kitware.github.io/vtk-js/api/Rendering_Core_Camera.html#getParallelProjection), so this may just be a matter of making sure trame-vtk serializes this part of the camera state

I struggled to find good material design icons for this... suggestions for better icons would be appreciated!!


jourdain:
We usually don't serialize the camera to keep some freedom client side. So we'll have to be careful about how we want to implement that correctly. On the other hand, the WASM local rendering is getting close to being usable. Once that's available, there will be little interest in keeping vtk.js for the local rendering in trame.

jourdain:
Otherwise, the code looks good to me and for the icon, I don't find anything better. But if you have svg/png, we can use those too.

jourdain:
For the camera projection , we should look at the code in push_camera and update it...

jourdain:
> > On the other hand, the WASM local rendering is getting close to being usable.
> 
> Awesome! It will be a game changer.

It will, indeed. The major push and CI automation are done. Now we are in the details. Hopefully, we can tackle those quickly without much trouble. I'm hoping to switch the `trame-vtklocal` to public by the end of the week... 🤞 And fix any remaining issues before vtk 9.4 is fully released...

annehaley (approved):

tkoyama010:
@annehaley Thanks for the cooperative review, including the other PRs. It helps us a lot.

PR #5883, opened 2024-04-08
Fix string literal concatenation using `ISC` rules

tkoyama010:
### Overview

Fix string literal concatenation using `ISC` rules.

### Details

- See https://docs.astral.sh/ruff/rules/#flake8-implicit-str-concat-isc .


tkoyama010:
@pyvista-bot LGTM

PR #5884, opened 2024-04-08
Fix unnecessary iterable allocation for first elem

tkoyama010:
### Overview

Fix unnecessary iterable allocation for first element.

### Details

- See https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/.


tkoyama010:
@pyvista-bot LGTM

PR #5885, opened 2024-04-08
Fix loops that can be replaced with a builtin func

tkoyama010:
### Overview

Fix loops that can be replaced with a builtin function.

### Details

- See https://docs.astral.sh/ruff/rules/reimplemented-builtin/.


tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

tkoyama010:
@pyvista-bot LGTM

PR #5886, opened 2024-04-08
Fix if else block instead of dict lookup (SIM116)

tkoyama010:
### Overview

Fix if else block instead of dict lookup (SIM116).

### Details

-See https://docs.astral.sh/ruff/rules/if-else-block-instead-of-dict-lookup/.


tkoyama010:
@pyvista-bot LGTM

PR #5887, opened 2024-04-08
Fix if expressions that can be replaced with bool

tkoyama010:
### Overview

Fix if expressions that can be replaced with bool.

### Details

- See https://docs.astral.sh/ruff/rules/if-expr-with-true-false/.


tkoyama010:
@pyvista-bot LGTM

PR #5888, opened 2024-04-08
Use all SIM rules at ruff check configuration

tkoyama010:
### Overview

Use all SIM rules at ruff check configuration.

### Details

- See https://docs.astral.sh/ruff/rules/#flake8-simplify-sim.
- Ignore [SIM117](https://docs.astral.sh/ruff/rules/multiple-with-statements/)  not to change Trame's code.


tkoyama010:
@pyvista-bot LGTM

PR #5890, opened 2024-04-09
Fix repeated equality comparison from the Pylint

tkoyama010:
### Overview

Fix repeated equality comparison from the Pylint.

### Details

- See https://docs.astral.sh/ruff/rules/repeated-equality-comparison/.


tkoyama010:
@pyvista-bot LGTM

PR #5891, opened 2024-04-09
Fix collapsible else if derived from the Pylint

tkoyama010:
### Overview

Fix collapsible else if derived from the Pylint.

### Details

- See https://docs.astral.sh/ruff/rules/collapsible-else-if/.

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -2778,16 +2778,11 @@ class DataSetFilters:
             # PolyData with 'largest' mode generates bad output with unreferenced points
             output_needs_fixing = True
 
-        else:
-            # All other extraction modes / cases may generate incorrect scalar arrays
-            # e.g. 'largest' may output scalars with shape that does not match output mesh
-            # e.g. 'seed' method scalars may have one RegionId, yet may contain many
-            # disconnected regions. Therefore, check for correct scalars size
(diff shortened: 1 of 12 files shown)


tkoyama010:
@pyvista-bot LGTM

pyvista/core/filters/data_set.py
```diff
@@ -2778,16 +2778,11 @@ def _extract_points(_input, points):
             # PolyData with 'largest' mode generates bad output with unreferenced points
             output_needs_fixing = True
 
-        else:
-            # All other extraction modes / cases may generate incorrect scalar arrays
-            # e.g. 'largest' may output scalars with shape that does not match output mesh
-            # e.g. 'seed' method scalars may have one RegionId, yet may contain many
-            # disconnected regions. Therefore, check for correct scalars size
```
dcbr, line 2785: @tkoyama010 I haven't kept track of all the formatting changes as of lately, but I don't like that this one removed some comments for explaining the `else` case (not only here, as can be seen in the diff). Personally, I also find a nested `else`-`if` tree sometimes more clear than this flattened `elif` result, but that's more a stylistic preference I guess.

tkoyama010: Good point. I will revert this once and reconsider.

PR #5903, opened 2024-04-11
Add check for mutable default values in class attr

tkoyama010:
### Overview

Add check for mutable default values in class attr.

### Details

- See https://docs.astral.sh/ruff/rules/mutable-class-default/.


tkoyama010 (commented):
@pyvista-bot LGTM

tkoyama010:
@pyvista-bot LGTM

PR #5905, opened 2024-04-11
Fix getattr takes a constant attr value as an arg

tkoyama010:
### Overview

Checks for uses of `getattr` that take a constant attribute value as an argument (e.g., `getattr(obj, "foo")`).

> [!NOTE]
> https://docs.astral.sh/ruff/rules/get-attr-with-constant/


tkoyama010:
@pyvista-bot LGTM

PR #5909, opened 2024-04-11
Check imports that are using a common convention

tkoyama010:
### Overview

Check imports that are using a common convention.

> [!NOTE]
> https://docs.astral.sh/ruff/rules/unconventional-import-alias/


tkoyama010:
@pyvista-bot LGTM

PR #5822, opened 2024-03-26
Add new `Dataset Gallery` docs page to summarize metadata for downloadable datasets

user27182:
### Overview

Continuation of #5814. Resolves #5784.

This PR adds a new Dataset Gallery page to summarize dataset instance and file metadata for datasets. Datasets are also grouped into separate galleries based on their metadata to allow for some filtering and enable the discovery of datasets with specific properties.

In addition, the 'See Also' sections of all datasets from `downloads.py` are updated with cross-reference links to each respective entry in the Dataset Gallery.

Check out a recent preview of the new Dataset Gallery [here](https://661039f500c5bf28c4091d12--meek-duckanoo-d2f5ec.netlify.app/api/examples/dataset_gallery).

### Details
The galleries are created with the following steps.

- The metadata is scrapped using the `dataset_loader` class from #5787. Many new props and bug fixes are added to this class in this PR.
- RST of the metadata fields is then generated for each dataset and stuffed into a [sphinx-design card](https://sphinx-design.readthedocs.io/en/latest/cards.html)
- Cards are collected together to create a [`card-carousel`](https://sphinx-design.readthedocs.io/en/latest/cards.html#card-carousels)
- The rst for each carousel has a similar structure to a table where each row is a card, and so a `DatasetGalleryCarousel` class is added to `make_tables.py` which inherits from the `DocTable` class.
- Datasets are grouped together using filters (e.g. filter dataset type) to create many card carousels
- A `<name>_carousel.rst` file is automatically generated for each carousel. These rst files can be included anywhere in the docs. 
- Card carousels are then grouped together using [tabs](https://sphinx-design.readthedocs.io/en/latest/tabs.html) and [dropdowns](https://sphinx-design.readthedocs.io/en/latest/dropdowns.html)
- The dropdowns and tab groups are defined in a new `dataset_gallery.rst`
- Since there are hundreds of datasets and many dataset categories, using dropdowns, tabs, and card carousels helps keep things organized and compact without the need to scroll vertically past all the datasets. 

Some notable features of the gallery:
- The card carousels were built to accommodate both small screens and large screens, so that the gallery is viewable from mobile or desktop. (If on desktop, try this out by shrinking your browser's width).
- The structure, content, and language used for the dataset types featured in the Dataset Gallery (e.g. PointSets, Gridded Data) aims to closely match the [core API docs](https://docs.pyvista.org/version/stable/api/core/) so that no confusing terminology is introduced.
[...]

PR #5921, opened 2024-04-12
Add coverage for code in `doc/source/make_tables.py` 

user27182:
Add a new doc test to be executed during the `Build Documentation` work flow to ensure there is coverage for some of the documentation source code. Specifically, tests are added for the `make_tables.py` module .

This does two things:
- Adds test coverage for the Dataset Gallery (see #5822) rst generation from the dataset loaders. Currently, there are no tests checking the rst outputs.
- Since `make_tables` loads all datasets from `downloads.py`, this should eliminate the need to use codecov skips for that module, and give a more accurate representation of the actual test coverage


user27182:
pre-commit.ci autofix

PR #5922, opened 2024-04-12
Fix typing in `doc/source/make_tables.py`

user27182:
Mypy is currently excluded from checking `doc/source/make_tables.py`. This PR fixes that to include that module, and resolves the type errors.


tkoyama010 (approved):
LGTM. I have verified that the `mypy` exclude is dropped and CI has passed.

PR #5818, opened 2024-03-24
Add `flake8-use-pathlib` lint check using ruff

tkoyama010:
### Overview

Add `flake8-use-pathlib` lint check using ruff.

### Details

- Follow up of #5718 .

PR #5926, opened 2024-04-13
[patch] fix jupyter-server-proxy prefix path issue from #5818

banesullivan:
fix jupyter-server-proxy prefix path issue from #5818

See https://github.com/pyvista/pyvista/pull/5818#discussion_r1563753299


tkoyama010 (approved):
Thanks!

Issue #5848, opened 2024-04-02
Interactive plot cannot be shown in Jupyter


EricHUANG970823:
### Describe the bug, what's wrong, and what you expected.

I found a bug!

UserWarning: Failed to use notebook backend: cannot import name 'vtk' from 'trame.widgets'

pyvista: 0.43.4
vtk: 9.3.0

MacOS Sonoma 14.4.1

### Steps to reproduce the bug.

```python
# examples
from pyvista import examples
mesh = examples.download_dragon()
mesh['scalars'] = mesh.points[:, 1]
mesh.plot(cpos='xy', cmap='plasma')
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Tue Apr 02 18:54:52 2024 CST

                OS : Darwin
            CPU(s) : 10
           Machine : arm64
      Architecture : 64bit
               RAM : 16.0 GiB
       Environment : Jupyter
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M2 Pro
       GPU Version : 4.1 Metal - 88
  MathText Support : True

  Python 3.11.0 | packaged by conda-forge | (main, Jan 14 2023, 12:26:40)
  [Clang 14.0.6 ]

           pyvista : 0.43.4
               vtk : 9.3.0
             numpy : 1.26.4
        matplotlib : 3.8.3
            scooby : 0.9.2
             pooch : 1.8.1
            pillow : 10.2.0
           imageio : 2.34.0
           IPython : 8.22.2
        ipywidgets : 8.1.2
             scipy : 1.12.0
              tqdm : 4.66.2
        jupyterlab : 4.1.5
             trame : 3.4.0
      trame_client : 2.16.5
      trame_server : 2.17.2
      nest_asyncio : 1.6.0
--------------------------------------------------------------------------------
```

jxfruit:
I got the similar problem, which shown in the following picture.
![Screenshot from 2024-04-02 19-06-21](https://github.com/pyvista/pyvista/assets/11846449/679d58f8-1ec7-4713-a471-663da0afa2c6)
If I install trame-vtk and trame-vuetify, a new problem will be here:
![3](https://github.com/pyvista/pyvista/assets/11846449/83ea6cf1-ac95-4ec2-86fa-29b88f9c1664)

my environment running with jupyter-notebook in docker container:
`pip list| grep jupyter
imjoy-jupyter-extension        0.3.0 
jupyter-client                 8.6.1
[...]

PR #5930, opened 2024-04-15
Update installation command on WSL at document

dveni:
### Overview

Install pyvista with jupyter option to avoid error in interactive rendering at jupyter lab, following https://github.com/pyvista/pyvista/issues/5848#issuecomment-2032334641


tkoyama010 (approved):
LGTM. Thanks!

PR #5933, opened 2024-04-16
Add `user_dict` property to `DataSet` objects for storing user-specified data

user27182:
### Overview

This PR adds a new `user_dict` property to enable easily storing/accessing user data in a dict-like fashion. It's intended to:

1. Make it easy to store scalar data alongside a mesh, since using regular field data requires storing data in arrays.
2. Better separate user data from mesh data, since all user data can now be nested in a single field data key instead of requiring a new field data key for every item.

If the mesh is saved in a compatible format (e.g. `.vtk`) and then read again, the JSON dict is automatically de-serialized and processed so that it is useable again as a dict object. This is very useful when batch-processing files, allowing for processing a mesh with a script and adding timestamps, parameters, results, etc. and saving everything to a single `.vtk` file.

### Details

Internally, the user dict is first serialized to JSON and stored as a `vtkStringArray`. The array is then stored as a `_PYVISTA_USER_DICT` field data array.

This PR is loosely related to #5747.

### DIFF
--- a/pyvista/core/utilities/arrays.py
+++ b/pyvista/core/utilities/arrays.py
@@ -1,9 +1,11 @@
 """Internal array utilities."""
 
+from collections import UserDict
 import collections.abc
 import enum
 from itertools import product
-from typing import Optional, Tuple, Union
+import json
+from typing import Optional, Tuple, Type, Union
 
 import numpy as np
 
@@ -548,6 +550,15 @@ def vtk_id_list_to_array(vtk_id_list):
     return np.array([vtk_id_list.GetId(i) for i in range(vtk_id_list.GetNumberOfIds())])
 
 
+def _set_string_scalar_object_name(vtkarr: _vtk.vtkStringArray):
+    """Set object name for scalar string arrays."""
+    # This is used as a flag so that scalar arrays can be reshaped later.
+    try:
+        vtkarr.SetObjectName('scalar')
+    except AttributeError:
+        vtkarr.GetObjectName = lambda: 'scalar'
+
+
 def convert_string_array(arr, name=None):
     """Convert a numpy array of strings to a vtkStringArray or vice versa.
 
@@ -583,13 +594,10 @@ def convert_string_array(arr, name=None):
             )
         vtkarr = _vtk.vtkStringArray()
         if arr.ndim == 0:
+            arr = arr.reshape((1,))
             # distinguish scalar inputs from array inputs by
             # setting the object name
-            arr = arr.reshape((1,))
-            try:
-                vtkarr.SetObjectName('scalar')
-            except AttributeError:
-                vtkarr.GetObjectName = lambda: 'scalar'
+            _set_string_scalar_object_name(vtkarr)
 
         ########### OPTIMIZE ###########
         for val in arr:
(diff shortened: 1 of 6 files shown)


tkoyama010:
@pyvista-bot preview

tkoyama010:
@pyvista-bot preview

pyvista/core/utilities/arrays.py
```diff
@@ -811,3 +819,109 @@ def _coerce_transformlike_arg(transform_like: TransformLike) -> NumpyArray[float
...
+    Type[str],
+    Type[int],
+    Type[float],
+    Type[bool],
+    Type[None],
+]
+
+
+# TODO: add generic type annotations 'UserDict[str, _JSONValueType]'
+#  once support for Python 3.8 is dropped
```
tkoyama010, line 837: Please see https://github.com/pyvista/pyvista/pull/5623 about Python 3.8 dropping schedule.

darikg:
that's a cool idea & implementation. Do field data values get preserved through VTK filters?

user27182:
> Do field data values get preserved through VTK filters?

I believe existing point_data, cell_data, or field_data is generally preserved by the filters, but I'm not 100% sure. Sometimes filters may need to be configured as such (e.g. IIRC the `vtkPackLabels` filters has a flag to explicitly pass-through this data). Or sometimes filters may need to clear data as part of a workaround, which may inadvertently clear the field data.

In cases where data _is not_ preserved by a filter though, perhaps this should be considered a bug?

Do you think I should add some tests with a few filters just to make sure?

darikg:
> > Do field data values get preserved through VTK filters?
> 
> I believe existing point_data, cell_data, or field_data is generally preserved by the filters, but I'm not 100% sure. Sometimes filters may need to be configured as such (e.g. IIRC the `vtkPackLabels` filters has a flag to explicitly pass-through this data). Or sometimes filters may need to clear data as part of a workaround, which may inadvertently clear the field data.
> 
> In cases where data _is not_ preserved by a filter though, perhaps this should be considered a bug?
> 
> Do you think I should add some tests with a few filters just to make sure?

I was just going to suggest adding a note to the documentation explaining what's expected to happen with filters but I forgot some have those `pass_data` options. I guess just leave it up to the filters to do whatever they want then.

user27182:
> > Do you think I should add some tests with a few filters just to make sure?
> 
> I was just going to suggest adding a note to the documentation explaining what's expected to happen with filters but I forgot some have those `pass_data` options. I guess just leave it up to the filters to do whatever they want then.

I added a few explicit tests with `merge`, `threshold` and `pack_labels`. Good thing too, since it caught a bug caused by copying field data for a string when an array was expected.

These three filters all pass-through the field data.

I have added a warning to the docstring though indicating that this is not guaranteed though.

user27182:
@pyvista-bot preview

tests/core/test_dataset.py
```diff
@@ -192,6 +194,93 @@ def test_field_data_string(grid):
...
+
+    ant_read = pv.PolyData(filepath)
+    assert ant_read.user_dict == dict_data
+
+    dict_field_repr = repr(ant.user_dict)
+    field_data_repr = repr(ant.field_data)
+    assert dict_field_repr in field_data_repr
+
+
+def test_user_dict_persists_with_merge_filter(uniform):
```
tkoyama010, line 259:
```suggestion
def test_user_dict_persists_with_merge_filter():
```

tests/core/test_dataset.py
```diff
@@ -192,6 +194,93 @@ def test_field_data_string(grid):
...
+    assert merged.user_dict['name'] == 'sphere2'
+
+
+def test_user_dict_persists_with_threshold_filter(uniform):
+    uniform.user_dict['name'] = 'uniform'
+    uniform = uniform.threshold(0.5)
+    assert uniform.user_dict['name'] == 'uniform'
+
+
+def test_user_dict_persists_with_pack_labels_filter(uniform):
```
tkoyama010, line 276:
```suggestion
def test_user_dict_persists_with_pack_labels_filter():
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

PR #5943, opened 2024-04-18
Add medical labels examples featuring `whole_body_ct` dataset

user27182:
Add new full example featuring the recently added `download_whole_body_ct_male` dataset from #5910.

PR #5944, opened 2024-04-19
Add `sigmoid_15` and `sigmoid_20` opacity transfer functions

user27182:
Sometimes, more sigmoid is needed. This PR adds two new sigmoids for plotting volumes. ``sigmoid_15`` is used by #5943. I found that for plotting high-intensity values only, e.g. showing only bones without any soft tissue, ``sigmoid_10`` wasn't enough.


tkoyama010 (approved):
LGTM. Next time we need further value, let's be able to specify the sigmoid values dynamically.

user27182:
Thanks. Yes I agree dynamic would be better.

PR #5950, opened 2024-04-20
Add `split_labels` filter and update `split_bodies`

### LINKED ISSUE #5613: split_bodies() generates a body that has 0 volume

After running `split_bodies` on a mesh that has a volume I expect returned bodies to have volume as well.

See [Colab](https://colab.research.google.com/drive/1P2xhRVPtCOCmpHWMbvbAD8u4o3tIQBMC?usp=sharing) for a reproduction. I have a single-component data that has volume == 0 in currentt pyvista.

See [Colab for pyvista 0.36.1](https://colab.research.google.com/drive/1g6RlzoXuupf8uGBqM_1IISqvAHLBTKtj?usp=sharing)  where it used to work 

```python

import pyvista
from pyvista import examples
dataset = examples.download_lucy()
print([(len(x.points), x.volume) for x in dataset.split_bodies()])
print(dataset.volume)
```

### System Information

```shell
  Date: Tue Feb 13 11:14:02 2024 UTC

                OS : Linux
            CPU(s) : 2
           Machine : x86_64
      Architecture : 64bit
               RAM : 12.7 GiB
       Environment : IPython
        GPU Vendor : Mesa
      GPU Renderer : llvmpipe (LLVM 15.0.7, 256 bits)
       GPU Version : 4.5 (Core Profile) Mesa 23.0.4-0ubuntu1~22.04.1
  MathText Support : False

  Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]

           pyvista : 0.43.3
               vtk : 9.3.0
             numpy : 1.23.5
        matplotlib : 3.7.1
            scooby : 0.9.2
             pooch : 1.8.0
            pillow : 9.4.0
           imageio : 2.31.6
           IPython : 7.34.0
          colorcet : 3.0.1
        ipywidgets : 7.7.1
             scipy : 1.11.4
              tqdm : 4.66.1
      nest_asyncio : 1.6.0
```

### Screenshots

_No response_

user27182:
Add new `split_labels` `DataSet` filter. This function is like `split_bodies`, but without the initial call to the `connectivity` filter. I.e., it assumes the input is already labeled data.  As such, this PR can be viewed as a refactoring of `split_bodies`.

In addition, this PR adds a new `method` param to `split_bodies`, allowing to split with either `threshold` or `connectivity`. This resolves #5613.

The  `split_labels` method pairs nicely with other filters which produce labeled data, e.g. `connectivity`, `contour_labeled`,  `pack_labels`.

### DIFF
--- a/tests/core/test_dataset_filters.py
+++ b/tests/core/test_dataset_filters.py
@@ -1424,20 +1424,32 @@ def labeled_data_Labels(labeled_data_RegionId):
     [pyvista.DataSetFilters.split_labels, pyvista.DataSetFilters.split_bodies],
 )
 @pytest.mark.parametrize('method', ['threshold', 'connectivity'])
+@pytest.mark.parametrize('test_with_RegionId', [True, False])
+@pytest.mark.parametrize('keep_RegionId_labels', [True, False])
 @pytest.mark.needs_vtk_version(9, 1, 0)
 def test_split_labels_split_bodies(
     dataset_filter,
     method,
+    test_with_RegionId,
+    keep_RegionId_labels,
     labeled_data_Labels,
     labeled_data_RegionId,
 ):
-    if True:
-        sphere, box, labeled = labeled_data_Labels
-        input_array_names = labeled.array_names
-    else:
+    if test_with_RegionId:
         sphere, box, labeled = labeled_data_RegionId
+    else:
+        sphere, box, labeled = labeled_data_Labels
+    input_arrays = labeled.array_names
+    expected_output_arrays = list(input_arrays)
 
-    split = dataset_filter(labeled, method=method, progress_bar=True)
+    kwargs = dict(method=method, progress_bar=True)
+    if dataset_filter is pyvista.DataSetFilters.split_bodies:
+        kwargs['label'] = keep_RegionId_labels
+        if test_with_RegionId and keep_RegionId_labels:
+            expected_output_arrays.append('RegionId')
+            expected_output_arrays.append('RegionId')
+
+    split = dataset_filter(labeled, **kwargs)
 
     assert len(split) == 2
     if method == 'threshold':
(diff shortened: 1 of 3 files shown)


tests/core/test_dataset_filters.py
```diff
@@ -1400,6 +1401,41 @@ def test_split_bodies():
...
+    assert isinstance(labeled, pyvista.PolyData)
+    return sphere, box, labeled
+
+
+@pytest.mark.parametrize(
+    'dataset_filter',
+    [pyvista.DataSetFilters.split_labeled, pyvista.DataSetFilters.split_bodies],
+)
+@pytest.mark.parametrize('method', ['threshold', 'connectivity'])
+def test_split_labeled_split_bodies(dataset_filter, labeled_data, method):
```
tkoyama010, line 1419:
```suggestion
@pytest.mark.needs_vtk_version(9, 1, 0)
def test_split_labeled_split_bodies(dataset_filter, labeled_data, method):
```

user27182:
There are some inconsistencies in the filter naming conventions for labeled data filters, e.g. `pack_labels` vs. `contour_labeled`.  I've updated the filter name to follow the `_labels` convention rather than `_labeled`.

user27182:
There is an issue with RegionId scalars being added to the output when they should not be. I'll mark as draft until this is fixed.

user27182:
Closing this in favor of #5963. The `extract_values` filter there is more general, and calling `extract_values` without arguments is effectively the same as calling `split_labels`. 

That PR doesn't resolve #5613 though. But, I am rethinking the casting from Unstructured grid to PolyData. I think that #5613  may best be resolved by adding notes to the docs rather than modifying methods to address the PolyData -> PolyData mapping.

PR #5955, opened 2024-04-22
Improve detection of embree and update installation instructions in ImportError of multi_ray_trace

Keou0007:
### Overview

The `embree` package in anaconda/conda-forge has moved up to version 4.x, but the `embree` based ray tracing in `trimesh` only supports `embree=2.x`. If `embree >= 3` is installed, `trimesh` will silently fail to use the `embree` library and fallback on the `rtree` based approach that is much much slower, often to the point of being problematic.

This change is a little bit smarter about how it checks for the availability of `embree` based ray tracing through `trimesh`, and will raise an `ImportError` if `embree` isn't working, rather than allowing the silent fallback to the slow method. It also provides up to date instructions on installing the right packages to make it work, via either `conda` or `pip`.

### Details

- Will no longer raise `ImportError` if `rtree` is not installed, which is not actually required for the `embree` based ray tracing.
- Will explicitly check that `trimesh` has successfully imported `embree` and raise `ImportError` if not.
- Updated the `conda` installation instructions to install `embree=2` and not install `rtree`
- Added installation instructions using `pip`, which will install the `embreex` package (preferred by `trimesh`) instead of `pyembree`.
- have edited `test_multi_ray_trace` so the pytest skips match the checks for `embree` in `multi_ray_trace`
- have added `embreex` to `requirements_test.txt` so `test_multi_ray_trace` will run instead of being skipped


Keou0007:
pre-commit.ci autofix

tkoyama010 (dismissed):
LGTM

tkoyama010 (approved):

tkoyama010:
pre-commit.ci autofix

Keou0007:
pre-commit.ci autofix

Keou0007:
This is ready to go now. There was a mistake in the updated code from #5957 that wasn't getting caught because the test was being skipped. Now that this PR enables the test by adding `embreex` to the requirements, the test was failing. But I've found the problem and fixed it.

tkoyama010 (approved):

tkoyama010:
> This is ready to go now. There was a mistake in the updated code from #5957 that wasn't getting caught because the test was being skipped. Now that this PR enables the test by adding `embreex` to the requirements, the test was failing. But I've found the problem and fixed it.

Thanks for your effort!

PR #5956, opened 2024-04-22
Add `PyElastica` as an external example of PyVista

tkoyama010:
### Overview

Add `PyElastica` as an external example of PyVista.

### Details

- https://www.cosseratrods.org/software/pyelastica/


tkoyama010:
@pyvista-bot LGTM

PR #5957, opened 2024-04-22
Fix missed ray traces in `multi_ray_trace` when origin points are outside bounding box

### LINKED ISSUE #5587: multi_ray_trace Enhancement Proposal

The `multi_ray_trace` function in `PolyDataFilters` (located in `filters/poly_data.py`) currently handles retrying ray trace when Embree’s ray tracing fails. However, there is an issue with the current approach.

When retrying, the function creates a second point and invokes `ray_trace` from PyVista, using `origins_retry` and `second_points` as arguments. The calculation for `second_points` is as follows:

```
second_points = origins_retry + unit_directions * self.length
```

Here, `self.length` represents the size of the poly_data.

However, if `origins_retry` is significantly distant from the `poly_data` (beyond self.length), the line segments between `origins
_retry` and`second_points` do not necessarily intersect with `poly_data`. As a result, `ray_trace` may fail to find intersections.

To increase the likelihood of finding intersections, I propose changing the calculation for the second points as follows:

```
poly_data = self

direct = (
    (poly_data.bounds[0] + poly_data.bounds[1]) / 2.  - origins_retry[0],
    (poly_data.bounds[2] + poly_data.bounds[3]) / 2.  - origins_retry[1],
    (poly_data.bounds[4] + poly_data.bounds[5]) / 2.  - origins_retry[2]
)
second_points = origins_retry + unit_directions * (
    np.sqrt((direct[0] ** 2 + direct[1] ** 2 + direct[2] ** 2)) + poly_data.length)
    
```

The code here is meant to show the idea. I'm not sure `bounds` is suited for this purpose.

This change should improve the function’s ability to find intersections in a wider range of cases. 

_No response_

### Pseudocode or Screenshots

The following code is based on `multi_ray_trace` to show that the choice of the second_points is import fo find intersections.

```python
import pyvista as pv
import numpy as np

sphere = pv.Sphere()

src = np.array([(10., 0., 0.)])
directions = np.array([(-1., 0.01, 0.01)])

origins_retry = src
directions_retry = directions

unit_directions = directions_retry / np.linalg.norm(
    directions_retry, axis=1, keepdims=True
)

second_points = origins_retry + unit_directions * sphere.length

for origin, second_point in zip(origins_retry, second_points):
    locs, indices = sphere.ray_trace(origin, second_point, first_point=True)
    print(locs) ##  "[]"

direct = (
    (sphere.bounds[0] + sphere.bounds[1]) / 2.  - origins_retry[0][0],
    (sphere.bounds[2] + sphere.bounds[3]) / 2.  - origins_retry[0][1],
    (sphere.bounds[4] + sphere.bounds[5]) / 2.  - origins_retry[0][2]
)
second_points = origins_retry + unit_directions * (
    np.sqrt((direct[0] ** 2 + direct[1] ** 2 + direct[2] ** 2)) + sphere.length)
                                                   
for origin, second_point in zip(origins_retry, second_points):
    locs, indices = sphere.ray_trace(origin, second_point, first_point=True)
    print(locs) ## "[0.48025852 0.09519742 0.09519742]"
```
Or drag your screenshot here!

Keou0007:
Implements @natsuwater's proposed enhancement to ensure the ray trace vector length is always long enough to reach any intersection points with the dataset.

Resolves #5587


tkoyama010:
pre-commit.ci autofix

tkoyama010:
Thanks! I'm not sure if we should consider this a bug or a feature addition.

Keou0007:
> Thanks! I'm not sure if we should consider this a bug or a feature addition.

I think probably bug, but squashed now! :)

Keou0007:
I figured out how to do it by patching the RayMeshIntersector in Trimesh. The test works as expected now and correctly tests that the retry function works when there is an origin point well outside the bounding box. 

Following my comment above, the test only runs when embree is installed. Currently this test is skipped in all the CI builds.
I have made a different change to the test in #5955 that resolves this by adding embreex to the test requirements.

Keou0007:
pre-commit.ci autofix

tests/core/test_polydata.py
```diff
@@ -336,14 +337,26 @@ def test_ray_trace_origin():
...
     assert np.any(ind_t)
 
+    # patch embree to test retry
+    with patch.object(
+        trimesh.ray.ray_pyembree.RayMeshIntersector,
+        'intersects_location',
+        return_value=[np.array([])] * 3,
+    ):
+        points, ind_r, ind_t = sphere.multi_ray_trace(origins, directions, retry=True)
+        print(points, ind_r, ind_t)
```
tkoyama010, line 355: Thanks for adding test. Can we remove this?
```suggestion
```

tkoyama010 (changes requested):

tests/core/test_polydata.py
Keou0007: Yes of course, sorry. Was late when I finished this and I thought I'd deleted all my print statements from testing, but I clearly missed one!

Keou0007:
I'm not sure why the python39 macOS test is failing, but I don't think it has to do with anything I've changed here.

tkoyama010:
> I'm not sure why the python39 macOS test is failing, but I don't think it has to do with anything I've changed here.

Please wait until #5960 is merged.

tests/core/test_polydata.py
```diff
@@ -336,14 +337,25 @@ def test_ray_trace_origin():
...
     assert np.any(points)
     assert np.any(ind_r)
     assert np.any(ind_t)
 
+    # patch embree to test retry
+    with patch.object(
+        trimesh.ray.ray_pyembree.RayMeshIntersector,
+        'intersects_location',
+        return_value=[np.array([])] * 3,
+    ):
```
tkoyama010, line 353: I'm sorry. I am new to Python and do not understand the intent of this syntax 🤦 Could you please share a URL that will help me understand?

Keou0007: https://docs.python.org/3/library/unittest.mock.html

It patches the `intersects_location` method in the Trimesh object (which it inherits from the RayMeshIntersector), so that when that method is called, instead of it actually running it just returns three empty numpy arrays. 

This is the behaviour that would occur if the ray trace ran and found no intersection points, so here it's being used to simulate the embree ray trace failing and returning nothing, such that the retry code in `multi_ray_trace` can then be tested. It also means that the `intersects_location` function doesn't actually run during the test, which makes the test a tiny bit quicker.

tkoyama010: Oh I see, you are mocking up the `trimesh` method in the following section. I understand now. Thanks for explaining :)
https://github.com/pyvista/pyvista/blob/2a59ac183fb7553a71faf5fd7f212b0ef115da46/pyvista/core/filters/poly_data.py#L2312-L2317

tkoyama010 (approved):
LGTM

PR #5960, opened 2024-04-22
Pin `macos-12` in testing and deployment actions

tkoyama010:
### Overview

The following error is occurring in CI. Let's pin the Mac version to 12 until we drop support of Python 3.9.
```
Error: The version '3.9' with architecture 'arm64' was not found for macOS 14.4.1.
```

### Details

- None

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -21,7 +21,7 @@ env:
 
 jobs:
   macOS:
-    runs-on: macos-latest
+    runs-on: macos-12
     name: Mac OS Unit Testing
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/testing-and-deployment.yml
```diff
@@ -21,7 +21,7 @@ env:
 
 jobs:
   macOS:
-    runs-on: macos-latest
+    runs-on: macos-12
```
tkoyama010, line 24:
```suggestion
    # https://github.com/pyvista/pyvista/pull/5960
    runs-on: macos-12
```

banesullivan (approved):

PR #5966, opened 2024-04-24
Add anatomical groups example

user27182:
### Overview

Add full example showing how to plot different anatomical groups from the `whole_body_ct` datasets. Preview image showing a plot of the cardiovascular system:

image

#### TODO

Dependencies:
- [ ] Merge `extract_values` filter from #5963 
- [ ] Merge updates to `contour_labeled` filter from #5968 
- [ ] Upload color dict to `pyvista/vtk-data` repo mapping labels to colors for the `whole_body_ct` datasets

Update `whole_body_ct` datasets:
- [ ] Add color dict mapping to the `user_dict` for these datasets
- [ ] Add new kwargs to enable loading only part of the dataset, i.e. integrate the utility functions from the full example into the loading functions
    e.g. `download_whole_body_ct(['heart'])` to load only the heart label for this dataset.

PR #5968, opened 2024-04-24
Update `contour_labeled` filter

user27182:
### Overview
Add new params and update a few existing ones for the `contour_labeled` filter.

Some of these features are used directly by #5966.

### Details

- Add `selected` output style and associated `select_outputs` param to filter output labels
- Add `select_inputs` param to filter input labels before contouring
- Fix `smoothing_constraint_distance`, it currently has no effect since automatic smoothing constraints are on by default which override this value (This is fixed by setting `alg.AutomaticSmoothingConstraintsOff()`)
- Add `smoothing_constraint_scale` to allow setting a relative smoothing constraint distance
- Modify default behavior for `output_mesh_type`. The default for `vtkSurfaceNets` is to use quads when smoothing is off, and triangles when smoothing is on. I think this dynamic default makes sense instead of forcing `quads` to be the default since the quads are likely non-planar for smoothed meshes.


user27182:
@pyvista-bot preview

user27182:
@pyvista-bot preview

user27182:
@pyvista-bot preview

user27182:
@pyvista-bot preview

user27182:
@pyvista-bot preview

user27182:
IIRC this feature is technically complete, but still need to handle/define the behavior for edge cases with empty meshes

user27182:
Closing in favor of #6732

PR #5969, opened 2024-04-24
Fix datetime call to use timezone-aware objects

tkoyama010:
### Overview

Fix datetime call to use timezone-aware objects.

### Details

- https://docs.astral.sh/ruff/rules/call-date-today/


tkoyama010:
@pyvista-bot LGTM

PR #5970, opened 2024-04-25
Add ruff rules which doesn't raise errors now

tkoyama010:
### Overview

Add ruff rules which doesn't raise errors now.

### Details

- None


tkoyama010:
@pyvista-bot LGTM

PR #5977, opened 2024-04-25
Bump black to 24.4.1 to support 3.12 f-strings

user27182:
### Overview

Bump black to [latest version](https://github.com/psf/black/releases/tag/24.4.1) which supports new f-string syntax for 3.12. 

This is relevant because the docs are built using python 3.12.

#5966 uses this syntax to format and print a dict. For that PR I can format it differently and work around this but thought I'd try to bump black instead.


tkoyama010 (approved):
LGTM

PR #5993, opened 2024-04-29
Add `pygrep-hooks` rule which raise now error now

tkoyama010:
### Overview

Add `pygrep-hooks` rule which raise now error now.

### Details

- See https://github.com/pre-commit/pygrep-hooks


tkoyama010:
@pyvista-bot LGTM

PR #5997, opened 2024-04-30
Use `ArrowSource` insted of using `vtkArrowSource`

tkoyama010:
### Overview

Use `ArrowSource` insted of using `vtkArrowSource`.

### Details

- None

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -2307,9 +2307,8 @@ class DataSetFilters:
 
         # Make glyphing geometry if necessary
         if geom is None:
-            arrow = _vtk.vtkArrowSource()
-            _update_alg(arrow, progress_bar, 'Making Arrow')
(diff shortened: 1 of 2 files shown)


pyvista/core/filters/data_set.py
```diff
@@ -2307,9 +2307,8 @@ def glyph(
 
         # Make glyphing geometry if necessary
         if geom is None:
-            arrow = _vtk.vtkArrowSource()
-            _update_alg(arrow, progress_bar, 'Making Arrow')
-            geom = arrow.GetOutput()
+            arrow = pyvista.ArrowSource()
+            geom = arrow.output
```
tkoyama010, line 2311:
```suggestion
            geom = arrow.output
            del arrow
```

banesullivan:
I'm actually wondering is we could just use `pv.Arrow()` instead of the source at all

PR #6014, opened 2024-05-02
Fix useless comparisons which have no effect

tkoyama010:
### Overview

Fix useless comparisons which have no effect.

### Details

- https://docs.astral.sh/ruff/rules/useless-comparison/


tkoyama010:
@pyvista-bot LGTM

examples/00-load/create-polydata-strips.py
```diff
@@ -101,4 +101,4 @@
 ###############################################################################
 # The output from the ``strip`` filter is, as expected, identical to the
 # original ``mesh``.
-restripped == mesh
+assert restripped == mesh
```
MatthewFlamm, line 104: These comparisons in the examples will show `True` as an output in the documentation.  I prefer it like this.  By changing to asset it isn't 100% clear whether it is True or False.  It could be asserted but raise an error for example.  In some cases we will want it to print `False` rather than raise an error.

tkoyama010: Thanks! You are right. I decide not to change these under the examples directory at  51ee4e81909c471a1ab665eb529a6a3fc6a63ce2.

MatthewFlamm (approved):

PR #6018, opened 2024-05-03
Fix function calls in default function arguments

tkoyama010:
### Overview

Fix function calls in default function arguments.

### Details

- https://docs.astral.sh/ruff/rules/function-call-in-default-argument/


tkoyama010:
@pyvista-bot LGTM

PR #6020, opened 2024-05-03
Fix loop control variable that overrides iterable

tkoyama010:
### Overview

Fix loop control variable that overrides iterable.

### Details

- https://docs.astral.sh/ruff/rules/loop-variable-overrides-iterator/


tkoyama010:
@pyvista-bot LGTM

PR #6024, opened 2024-05-03
Add sshd in `devcontainer.json` to ssh access

tkoyama010:
### Overview

When working with devcontainer, I want to access the environment via ssh to do some quick work. This PR adds sshd to devcontainer.json

### Details

- https://docs.github.com/en/codespaces/developing-in-a-codespace/using-github-codespaces-with-github-cli#ssh-into-a-codespace


MatthewFlamm (approved):
Seems fine to me.  Although we should avoid bloating this with too many things to keep build times low.  If you find that you don't end up using it much, we should remove it again.

PR #6037, opened 2024-05-05
Fix `dict.items` that discard either key or value

tkoyama010:
### Overview

Fix `dict.items` that discard either key or value.

### Details

- Fixing [PERF102](https://docs.astral.sh/ruff/rules/incorrect-dict-iterator/)
- [PERF203](https://docs.astral.sh/ruff/rules/try-except-in-loop/) is excluded from the rule because we use a loop iterator in except.


pyproject.toml
```diff
@@ -329,6 +329,7 @@ ignore = [
     # "BLK100",
     # 'from module import *' used; unable to detect undefined names
     "F403",
+    "PERF203",
```
tkoyama010, line 332:
```suggestion
    # https://github.com/pyvista/pyvista/pull/6037
    "PERF203",
```

annehaley (approved):

PR #6048, opened 2024-05-07
Checks for self-assignment of variable from Pylint

tkoyama010:
### Overview

Checks for self-assignment of variable from Pylint.

### Details

- https://docs.astral.sh/ruff/rules/self-assigning-variable/


tkoyama010:
@pyvista-bot LGTM

PR #2316, opened 2022-03-09
Crinkle clip

banesullivan:
Resolve #2313 cc @liangwang0734

```py
import pyvista as pv
import numpy as np

# Input mesh
mesh = pv.Wavelet()

# Perform the clip
normal = (1,1,1)
plane = pv.Plane(i_size=30, j_size=30, direction=normal)

clipped = mesh.clip(normal=normal)

# Crinkle the clip
crinkled = mesh.clip(normal=normal, crinkle=True)

# Plot comparison
p = pv.Plotter(notebook=0, shape=(1,2))
p.add_mesh(clipped, show_edges=True)
p.add_mesh(plane.extract_feature_edges(), color='r')
p.subplot(0,1)
p.add_mesh(crinkled, show_edges=True)
p.add_mesh(plane.extract_feature_edges(), color='r')
p.link_views()
p.show()
```
Screen Shot 2022-03-09 at 12 43 34 PM

```py
import pyvista as pv
import numpy as np

# Input mesh
mesh = pv.Wavelet()

p = pv.Plotter(notebook=0)
p.add_mesh_clip_plane(mesh, crinkle=True)
p.show()
```
Screen Shot 2022-03-09 at 12 58 32 PM

PR #6060, opened 2024-05-09
Ensure no overlap with crinkle clip

### LINKED ISSUE #5667: Crinkle clip duplicates cells in the unclipped surface

When I implement the crinkle clip feature in pyvista.DataSetFilters.clip and set crinkle=True and return_clipped=True there are common cells between the clipped and unclipped datasets. I would expect that the unclipped surface would be effectively the original surface, without the clipped part. 

It would be great if this could be changed because I would like to continue clipping the unclipped surface again but without the duplicated cells at interface.

Thanks! 

```python
import pyvista as pv
import numpy as np

dataset = pv.examples.download_bunny_coarse()
clipped, unclipped = dataset.clip('y', return_clipped=True, crinkle=True)

clipped.save(".../bunny_top.vtk")
unclipped.save(".../bunny_bottom.vtk")
```

### System Information

```shell
  Date: Thu Feb 22 15:43:29 2024 GMT Standard Time
            
                OS : Windows
            CPU(s) : 36
           Machine : AMD64
      Architecture : 64bit
               RAM : 255.7 GiB
       Environment : IPython
      GPU Renderer : Quadro RTX 4000/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 536.67
  MathText Support : False
  Python 3.8.8 (default, Apr 13 2021, 15:08:03) [MSC v.1916 64 bit (AMD64)]
           pyvista : 0.42.2
               vtk : 9.2.6
             numpy : 1.20.1
        matplotlib : 3.3.4
            scooby : 0.7.2
             pooch : v1.7.0
            pillow : 8.2.0
           imageio : 2.9.0
             PyQt5 : 5.9.2
           IPython : 7.22.0
        ipywidgets : 7.6.3
             scipy : 1.6.2
              tqdm : 4.59.0
        jupyterlab : 3.0.14
      nest_asyncio : 1.5.1
  Intel(R) oneAPI Math Kernel Library Version 2021.2-Product Build 20210312
```

### Screenshots

image

banesullivan:
### Overview

Resolves #5667 to ensure that when crinkle clipping and returning the clipped out mesh, the clipped out mesh does not have any of the same cells as the primary.

### Details

```py
import pyvista as pv
from pyvista import examples

dataset = examples.download_bunny_coarse()
clipped, unclipped = dataset.clip('y', return_clipped=True, crinkle=True)

pl = pv.Plotter()
pl.add_mesh(clipped, color='blue')
pl.add_mesh(unclipped, color='red')
pl.show()
```
screenshot-2


MatthewFlamm:
I am +1 for including this feature, and it makes sense based on the description of the parameters. But, are there also any use cases for returning the crinkle cut from both directions here (current behavior)?  I can't think of any obvious ones, but it seems probable that someone would want this behavior.

banesullivan:
> are there also any use cases for returning the crinkle cut from both directions here (current behavior)? I can't think of any obvious ones, but it seems probable that someone would want this behavior.

I also cannot think of any and I think that when calling the clip function with crinkle enabled and wanting both sides of the clip, there should not be any overlap.

If someone wants to crinkle both sides on the overlap, then thats still possible, its just two seperate clips and can be done two ways:

1. By flipping the normal:

```py
a = dataset.clip('y', crinkle=True)
b = dataset.clip('-y', crinkle=True)
```

2. Using the `invert` option:

```py
a = dataset.clip('y', crinkle=True, invert=True)
b = dataset.clip('y', crinkle=True, invert=False)
```


```py
import pyvista as pv
from pyvista import examples

dataset = examples.download_bunny_coarse()
clipped, unclipped = dataset.clip('y', return_clipped=True, crinkle=True)

a = dataset.clip('y', crinkle=True)
b = dataset.clip('-y', crinkle=True)
# a = dataset.clip('y', crinkle=True, invert=True)
# b = dataset.clip('y', crinkle=True, invert=False)

assert (a == clipped)
assert not (b == unclipped)

pl = pv.Plotter(shape=(1,2))
pl.add_mesh(clipped, color='blue')
pl.add_mesh(unclipped, color='red', style='wireframe', line_width=5)
pl.subplot(0, 1)
pl.add_mesh(a, color='blue')
pl.add_mesh(b, color='red', style='wireframe', line_width=5)

pl.link_views()
pl.show()
```
screenshot-3

MatthewFlamm:
I think it is fine as the PR has it.  There could be an unexpected breaking change for someone using the current behavior.  We might argue it was unintended behavior, so they should not have been using it?  I'm mostly trying to think through whether a 'breaking change' label should be applied, or this is just a bug fix.  Not trying to change the PR itself.

I thought of one case, two crinkle cuts that are offset in space are used in opposing directions to obtain a sliver of cells to compare cell size distribution (common in CFD for example).  Currently, a user could achieve this by either inverting one clip or using `return_clip=True` from one clip.

tests/core/test_dataset_filters.py
```diff
@@ -96,6 +96,9 @@ def test_clip_filter(datasets):
     clp1, clp2 = pv.Wavelet().clip(normal=(1, 1, 1), return_clipped=True, crinkle=True)
     assert clp1 is not None
     assert clp2 is not None
+    set_a = set(clp1.cell_data['cell_ids'])
+    set_b = set(clp2.cell_data['cell_ids'])
+    assert set_a.isdisjoint(set_b)
```
MatthewFlamm, line 101: Does it make sense to also test that `set_a + set_b == set(range(ncells))`?  This would require saving the original mesh before clipping.

MatthewFlamm (dismissed):
Small suggestion to consider.  Also consider labeling breaking change. Im only 50/50 on both.

tests/core/test_dataset_filters.py
banesullivan: Nice suggestion, done

MatthewFlamm (approved):

banesullivan:
While in my opinion this is a bug and unintended behavior I think you provide a great example and I want to note that this bug has been around for 2 years (ref #2316). I think a breaking change label to denote that the output mesh is indeed different is appropriate here.

PR #6062, opened 2024-05-09
Fix active normals for properties `point_normals`, `cell_normals`, and method `plot_normals` with `smooth_shading`

user27182:
Resolves #6061
Resolves #2921
Resolves #6058

PR #6069, opened 2024-05-10
Update docs for `point_normals`, `cell_normals`, `face_normals`

user27182:
Fix normals in docs which would have changed from the fix in #4674.

This is a follow up to #6062 to make sure that changes from that PR are not the cause of the changes to the sphere's normals.


tkoyama010:
@pyvista-bot preview

tkoyama010 (approved):
`# doctest:+SKIP` removal confirms this correctness. LGTM

PR #6082, opened 2024-05-14
MAINT: Fix bug with VTK dev

larsoner:
This PR should fail with the added test on the vtk `dev` job. This is because when you install that you'll see:
```
(Pdb) p PolyData.point_data
<attribute 'point_data' of 'vtkmodules.vtkCommonDataModel.vtkDataSet' objects>
```
the MRO of PolyData puts vtkPolyData first, and on the `dev` branch of VTK this exists for that class.

Once it fails I'll push what I think might constitute a fix, namely adding a `@property` wrapper for `point_data` and `cell_data` directly in `PolyData` itself which wraps to `PointSet.point_data`.

### DIFF
--- a/.github/workflows/vtk-pre-test.yml
+++ b/.github/workflows/vtk-pre-test.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Set up VTK with OSMesa
         run: |
           pip uninstall vtk -y
-          pip install vtk_osmesa --pre --no-cache --extra-index-url https://wheels.vtk.org
+          pip install vtk_osmesa --pre --no-cache --index-url https://wheels.vtk.org
 
       - name: Software Report
         run: |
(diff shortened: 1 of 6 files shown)


larsoner:
Successfully failing in 857de8f, trying my fix

larsoner:
Can't seem to get it to work right after spending a couple of hours messing with classes. I have to move on to other stuff, someone feel free to take over or come up with a better solution!

jourdain:
That is tacking care of in this [PR](https://github.com/pyvista/pyvista/pull/5650)

larsoner:
Okay I'll close this then!

banesullivan:
Yep, I got started fixing this and just haven't come back to it. I'll try to jump back on my PR soon here but we are closely tracking this and making sure we get this fixed before VTK's next release

PR #6094, opened 2024-05-16
Cache build html for preview in each pull request

tkoyama010:
### Overview

Cache build `html` for preview in each pull request.  Delete `vtksz` files so that the Cache does not grow too large.

Close #6094
Close #5859

### Details

- None

PR #6095, opened 2024-05-17
Revert "Cache build html for preview in each pull request"

tkoyama010:
Reverts pyvista/pyvista#6094
I have confirmed that this change has no effect length of built time because the api-doc will be rebuilt. I will look into making the plot-directive static in PR.

### DIFF
--- a/.github/workflows/preview_pull_request2.yml
+++ b/.github/workflows/preview_pull_request2.yml
@@ -61,13 +61,6 @@ jobs:
           echo PYVISTA_EXAMPLE_DATA_PATH=$(python -c "from pyvista import examples; print(examples.USER_DATA_PATH)") >> $GITHUB_ENV
           pip list
 
-      - name: Cache build html
-        uses: actions/cache@v4
-        if: env.USE_CACHE == 'true' && !startsWith(github.ref, 'refs/heads/release/') && !startsWith(github.ref, 'refs/tags/v')
-        with:
-          path: doc/_build/html/
-          key: doc-html-${{ github.event.issue.number }}
-
       - name: Cache Sphinx-Gallery Examples
         uses: actions/cache@v4
         if: env.USE_CACHE == 'true' && !startsWith(github.ref, 'refs/heads/release/') && !startsWith(github.ref, 'refs/tags/v')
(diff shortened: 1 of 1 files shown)


tkoyama010:
@pyvista-bot LGTM

user27182:
#6094 had immediate effect, right? As in once that was merged, any PRs could trigger the actions (no merge main required)?

Because the preview command issued here:
 https://github.com/pyvista/pyvista/pull/6071#issuecomment-2116377975
got a response in just over an hour (~70 minutes) instead of the 2+ hours it usually takes.

The actual docs build is also reporting 1h12 minutes here:
https://github.com/pyvista/pyvista/actions/runs/9120187514

So, I think it #6094 may have had a positive effect.

user27182:
The previous successful build for that PR from yesterday took about 1hr 28mins (not 2+ hrs)
https://github.com/pyvista/pyvista/actions/runs/9105572470

See also the preview comment for that build
https://github.com/pyvista/pyvista/pull/6071#issuecomment-2113915406

Is there a way to more explicitly report the build times? E.g. have the bot post a comment with the time? Just to help keep track and make sure it's not increasing longer

tkoyama010:
> #6094 had immediate effect, right? As in once that was merged, any PRs could trigger the actions (no merge main required)?
> 
> Because the preview command issued here: [#6071 (comment)](https://github.com/pyvista/pyvista/pull/6071#issuecomment-2116377975) got a response in just over an hour (~70 minutes) instead of the 2+ hours it usually takes.
> 
> The actual docs build is also reporting 1h12 minutes here: https://github.com/pyvista/pyvista/actions/runs/9120187514
> 
> So, I think it #6094 may have had a positive effect.

Speed-up you say is affected by the “Cache Sphinx-Gallery Examples” and “Cache Examples Data”. It was working and saving time before #6094 was merged.
1. Before #6094 was merged, the job is like this. https://github.com/pyvista/pyvista/actions/runs/9120082952/job/25076752627 
Note that “Cache Sphinx-Gallery Examples” and “Cache Examples Data” succeeded. This makes the total time one hour. Previously, the build time was 2 hours without Cache, but now that Cache is saved, the build time is 1 hour.
2. After #6094 was merged , the job is like this. https://github.com/pyvista/pyvista/actions/runs/9122100007/job/25082349529 "Cache build html" step download succeeded but the api-doc Sphinx build which takes the longest is need. This is likely due to the fact that api-doc relies on Python code.
3. Since api-doc seem to be always executed in Sphinx, we need to speed up the api-doc build itself. Therefore, we plan to suppress the dynamic output of api-doc to speed up the build time.

tkoyama010:
> The previous successful build for that PR from yesterday took about 1hr 28mins (not 2+ hrs) https://github.com/pyvista/pyvista/actions/runs/9105572470
> 
> See also the preview comment for that build [#6071 (comment)](https://github.com/pyvista/pyvista/pull/6071#issuecomment-2113915406)
> 
> Is there a way to more explicitly report the build times? E.g. have the bot post a comment with the time? Just to help keep track and make sure it's not increasing longer

It's a good way to use the bot to report. I would like to plan to add it.

user27182:
Ah Ok, thanks for the info! I was happily surprised by the relatively quick 70 mins build time, glad you're working this!

PR #6100, opened 2024-05-18
Fix `image_threshold` bug with integer arrays on some systems

user27182:
Fix issue with `image_threshold` unexpectedly returning all zeros for integer scalar arrays. This seems to be build-specific. I am not able to reproduce this locally but the bug shows up in the docs.

See images of the bug here https://github.com/pyvista/pyvista/pull/6071#issuecomment-2118468493

### DIFF
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -422,16 +422,15 @@ class ImageDataFilters(DataSetFilters):
             scalars,
         )  # args: (idx, port, connection, field, name)
         # set the threshold(s) and mode
-        if isinstance(threshold, (np.ndarray, collections.abc.Sequence)):
-            if len(threshold) != 2:
-                raise ValueError(
-                    f'Threshold must be length one for a float value or two for min/max; not ({threshold}).',
-                )
-            alg.ThresholdBetween(threshold[0], threshold[1])
-        elif isinstance(threshold, collections.abc.Iterable):
-            raise TypeError('Threshold must either be a single scalar or a sequence.')
+        threshold_val = np.atleast_1d(threshold).astype(float)
+        if (size := threshold_val.size) not in (1, 2):
+            raise ValueError(
+                f'Threshold must have one or two values, got {size}.',
+            )
+        if size == 2:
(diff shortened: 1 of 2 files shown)


user27182:
From the failed tests it looks like it's a Linux and/or Python < 3.12 issue. The tests fail for some cases with integer inputs.

pyvista/core/filters/image_data.py
```diff
@@ -407,13 +408,28 @@ def image_threshold(
...
-        alg = _vtk.vtkImageThreshold()
-        alg.SetInputDataObject(self)
         if scalars is None:
             set_default_active_scalars(self)
             field, scalars = self.active_scalars_info
         else:
             field = self.get_array_association(scalars, preference=preference)
+
+        # For some systems and/or versions of numpy, integer scalars won't threshold
+        # correctly. Cast to float in these cases.
```
tkoyama010, line 418:
```suggestion
        # For some systems and Numpy < 2.0, integer scalars won't threshold
        # correctly. Cast to float in these cases.
```

pyvista/core/filters/image_data.py
```diff
@@ -407,13 +408,28 @@ def image_threshold(
...
+            array_dtype := self.active_scalars.dtype,
+            int,
+        ) and array_dtype != np.dtype(np.uint8)
+        cast_dtype = (
+            has_int_dtype
+            and int(np.__version__.split('.')[0]) < 2
+            and platform.system() in ['Darwin', 'Linux']
+        )
+        if cast_dtype:
+            self[scalars] = self[scalars].astype(float, casting='safe')
```
tkoyama010, line 429:
```suggestion
        self[scalars] = self[scalars].astype(float, casting='safe') if cast_dtype
```

tkoyama010 (changes requested):

pyvista/core/filters/image_data.py
tkoyama010: This doesn't work closing.

tkoyama010 (approved):
LGTM. Thanks!

PR #6101, opened 2024-05-19
Add environment var to force static plot directive

tkoyama010:
### Overview

Add environment var to force static plot directive.

### Details

- None


tkoyama010:
@pyvista-bot preview

PR #6105, opened 2024-05-19
Check immediate raise within exception handlers

tkoyama010:
### Overview

Check immediate raise within exception handlers.

### Details

- See https://docs.astral.sh/ruff/rules/useless-try-except/


tkoyama010:
@pyvista-bot LGTM

PR #5718, opened 2024-03-02
Add testing for doc images generated by sphinx build

user27182:
### Overview

Add initial testing of images in the documentation. This is related to #5410, but does not (yet?) involve `pytest-pyvista`. This is a first proof of concept to experiment with how this kind of testing may be achieved.

### Details
The test works by collecting all images from `pyvista/doc/_build/_images` (after building the documentation) and comparing them to a collection of cached images.

Unlike the testing in `test/plotting`, this test suite does not use `plot`. Instead, `pv.compare_images` is used directly.

The number of `png` images collected from the build is approx. 2000, and their collective size is approx 250MB. To address concerns about file size, the build images are first pre-processed and resized to have a maximum size of 400x400 pixels (the same size as images in `tests/plotting/image_cache`. This reduces the size from 250MB to approx 70MB, which may be a reasonably small enough size to store the images directly as-is in the `pyvista` repository.

The tests will not be triggered by calling `pytest` generally. Instead, they must be triggered explicitly, e.g.  with 
```pytest tests/doc/tst_doc_images.py```, and the build images must already be in `pyvista/doc/_build/_images` 

#### TODO:

- [ ] Integrate tests into a new `Test Documentation` workflow

PR #6106, opened 2024-05-20
Make PyVista examples reproducible

user27182:
Follow up from #5718 

Many examples in the docs use random number generators. These are not seeded and therefore produce different plots every time they execute. This makes it so that the images they generate are not static, and makes it impossible to test as part of #5718.

This PR fixes this and seeds the rngs so the generated plots won't change every time.


examples/00-load/create-point-cloud.py
```diff
@@ -26,14 +26,17 @@
...
+    ids = rng.integers(
         low=0,
         high=dataset.n_points - 1,
         size=int(dataset.n_points * subset),
     )
     return dataset.points[ids]
 
 
+# Seed the random number generator for generating data
+rng = np.random.default_rng(seed=0)
```
MatthewFlamm, line 38: It is confusing to me to define this after the function that references it, although it is valid code.  I would recommend putting this up top.

user27182: Moved it up in https://github.com/pyvista/pyvista/pull/6106/commits/f155577032d248d99e9f0c5abfa7af9fc4152e2f, hopefully that's clearer now

user27182:
General comment - we could also add a hook with auto doc to execute code to always seed the rng when processing docstrings. 

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-process-docstring

Not sure if that would work, but if it does it would remove the need to seed the rng in the actual docstrings, which can distract from the example. 

It would also prevent future failures caused by #5718 if new examples are added with rng that fail to seed it first

tkoyama010:
> General comment - we could also add a hook with auto doc to execute code to always seed the rng when processing docstrings.
> 
> https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-process-docstring
> 
> Not sure if that would work, but if it does it would remove the need to seed the rng in the actual docstrings, which can distract from the example.
> 
> It would also prevent future failures caused by #5718 if new examples are added with rng that fail to seed it first

Let's try it!

user27182:
> > General comment - we could also add a hook with auto doc to execute code to always seed the rng when processing docstrings.
> > https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-process-docstring
> > Not sure if that would work, but if it does it would remove the need to seed the rng in the actual docstrings, which can distract from the example.
> > It would also prevent future failures caused by #5718 if new examples are added with rng that fail to seed it first
> 
> Let's try it!

I don't think this will work after all. We could search every line of all docstrings and try to replace `default_rng()` with `default_rng(seed=0)`, but this would fail for the create-point-cloud example since there `default_rng()` is used inside a function to generate random points multiple times. Adding a static seed for that case would break the example. So the fix would need to be a bit more sophisticated.

Plus, this would only work with autodoc, I don't think the gallery examples would be covered. So, might be better to add the seed explicitly after all

tkoyama010:
> I don't think the gallery examples would be covered.

Ah that makes sense.

MatthewFlamm:
I think that making it explicitly set in each case is good because then users can also reproduce the exact figure.

> I don't think this will work after all. We could search every line of all docstrings and try to replace `default_rng()` with `default_rng(seed=0)`, but this would fail for the create-point-cloud example since there `default_rng()` is used inside a function to generate random points multiple times. Adding a static seed for that case would break the example. So the fix would need to be a bit more sophisticated.

Are you referring to?
https://github.com/pyvista/pyvista/blob/73425063a2d06749c919b55a9821f324a6920c59/examples/00-load/create-point-cloud.py#L29

I think we should move the rng instantation outside and above that function either way.

user27182:
> Are you referring to?
> 
> https://github.com/pyvista/pyvista/blob/73425063a2d06749c919b55a9821f324a6920c59/examples/00-load/create-point-cloud.py#L29
> 
> I think we should move the rng instantation outside and above that function either way.

Yes I was. But, I also (incorrectly) thought the function was called multiple times to generate different random data multiple times. But, it is only called once. So the rng can be seeded directly inside the function after all.

The seed was moved back inside in https://github.com/pyvista/pyvista/pull/6106/commits/4433b61204810df0bfd085fdf59e94b8b07dbfac

Moving it outside initially is what seemed to make it confusing. I don't really mind where it's seeded either way.

@MatthewFlamm if you still prefer to have the rng seeded outside we can move it back.

MatthewFlamm:
Not an important comment.  It is weird to have this function that only is executed once, but that is not the point of this PR.  It is fine as it is.

pyvista/plotting/plotter.py
```diff
@@ -5748,7 +5750,9 @@ def add_points(self, points, style='points', **kwargs):
 
         >>> import numpy as np
         >>> import pyvista as pv
-        >>> points = np.random.default_rng().random((10, 3))
+        >>>
```
tkoyama010, line 5753:
```suggestion
```

pyvista/plotting/plotter.py
```diff
@@ -5799,8 +5803,10 @@ def add_arrows(self, cent, direction, mag=1, **kwargs):
 
         >>> import numpy as np
         >>> import pyvista as pv
-        >>> cent = np.random.default_rng().random((10, 3))
-        >>> direction = np.random.default_rng().random((10, 3))
+
```
tkoyama010, line 5806:
```suggestion
```

pyvista/plotting/plotter.py
```diff
@@ -5799,8 +5803,10 @@ def add_arrows(self, cent, direction, mag=1, **kwargs):
...
+        >>> rng = np.random.default_rng(seed=0)
+        >>> cent = rng.random((10, 3))
+        >>> direction = rng.random((10, 3))
```
tkoyama010, line 5809: Let's change seed here.
```suggestion
        >>> rng = np.random.default_rng(seed=1)
        >>> direction = rng.random((10, 3))
```

tkoyama010 (changes requested):

pyvista/plotting/plotter.py
dcbr: There's no need for that (the `cent` and `direction` parameters will already have different numbers with a single seed value).

examples/00-load/create-point-cloud.py
```diff
@@ -90,7 +91,7 @@ def generate_points(subset=0.02):
 # This time, we're going to create a totally new, random point cloud.
 
 # Create random XYZ points
-points = np.random.default_rng().random((100, 3))
+points = np.random.default_rng(seed=1).random((100, 3))
```
dcbr, line 94: I personally think it would be more clear if all examples and docstrings that need random numbers follow the same structure, i.e. importing the required packages, seeding an `rng` object, and continuing with the actual demo code (as is done in most other places in this PR). So no single line `rng` creation and usage as is done here. But that's just my 2 cents :)

pyvista/plotting/plotter.py
tkoyama010: I didn't know that. Thanks!

examples/00-load/create-point-cloud.py
user27182: This is also the recommended by scientific Python:

https://blog.scientific-python.org/numpy/numpy-rng/

"You should create an RNG at the beginning of your script (with your own seed if you want reproducibility) and use this RNG in the rest of the script."

E.g.

``` python 
import numpy as np

rng = np.random.default_rng()
rng.random()  # generate a floating point number between 0 and 1
```

user27182:
Not sure if there's a real risk for confusing terminology here but ``rng`` is also used in pyvista as an alias for ``clim`` , e.g. for ``add_mesh``:
https://github.com/pyvista/pyvista/blob/e6d69731d166776b267c7c0146c82cd3a8b9208a/pyvista/plotting/plotter.py#L2994-L2997

tkoyama010 (approved):
LGTM

CONTRIBUTING.rst
```diff
@@ -789,6 +792,27 @@ General usage and API descriptions should be placed within ``doc/api`` and
...
+When using NumPy's random number generator (RNG) you should create an RNG at
+the beginning of your script and use this RNG in the rest of the script. For
+example:
+
+.. code:: python
+
+    import numpy as np
+
+    rng = np.random.default_rng()
+    rng.random()  # generate a floating point number between 0 and 1
```
user27182, line 811:
```suggestion
When using NumPy's random number generator (RNG) you should create an RNG at
the beginning of your script and use this RNG in the rest of the script. Be 
sure to include a seed value. For example:

.. code:: python

    import numpy as np

    rng = np.random.default_rng(seed=0)
    rng.random()  # generate a floating point number between 0 and 1
```

tkoyama010:
@pyvista-bot preview

tkoyama010:
pre-commit.ci autofix

MatthewFlamm (approved):
Nice!

tkoyama010:
@user27182 Ready to merge?

user27182:
@tkoyama010 LGTM

Issue #5291, opened 2023-12-06
Wrong number of labels in `download_frog_tissue` example.


user27182:
### Describe the bug, what's wrong, and what you expected.

There is an issue with the number of labels or data range for the [download_frog_tissue](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_frog_tissue.html#pyvista-examples-downloads-download-frog-tissue) example.

This has caused intermittent doctest failures (see #5290) and also results in an incorrect image shown in the documentation.

Current image from the 0.42 release docs:
![](https://docs.pyvista.org/version/stable/_images/pyvista-examples-downloads-download_frog_tissue-1_00_00.png)

Expected image from when the example was originally added (#4801):
![image](https://github.com/pyvista/pyvista/assets/89109579/6f020c38-6ec1-405e-9068-29c3f3db8357)

### Steps to reproduce the bug.

See #5290.

### System Information

```shell
See #5290.
```

MatthewFlamm:
I haven't figured this out completely, but if I run one doctest for only `pack_labels`, then it passes.  If I run all doctests in that module, i.e. `pyvista.core.filters.data_set` then it fails.

MatthewFlamm:
Sorry for the noise, it seems to be more random than this, even though I could have sworn that it was consistent over several tries at first.  This behavior is reproducible on the default codespaces environment.

user27182:
> If I run `pytest --doctest-modules pyvista/core/filters/data_set.py::pyvista.core.filters.data_set.DataSetFilters.cell_data_to_point_data pyvista/core/filters/data_set.py::pyvista.core.filters.data_set.DataSetFilters.pack_labels -v` it _always_ fails.

This seems suspicious. `cell_data_to_point_data` is called internally by the `connectivity` filter. And the docstring for `download_frog_tissue` uses both the `connectivity` filter and `sort_labels`.  Since the order of tests clearly matters, is it possible that one test is corrupting the namespace or setup for the others?

MatthewFlamm:
Something is corrupting something else for sure.  I cannot figure out why.  This has now been a few months, I remember that I set breakpoints and saw differences that would explain the change in behavior.  But I couldn't make heads or tails of what was causing this, but you may want to do something similar to pinpoint the exact operation that gets corrupted.

MatthewFlamm:
Just went back and now I _cannot_ reproduce the failures using:

[...]

PR #6110, opened 2024-05-22
Make `frog_tissue` a built-in example and fix flaky tests

user27182:
### Overview
Fix https://github.com/pyvista/pyvista/issues/5291 and related flaky tests for pack_labels / sort_labels from https://github.com/pyvista/pyvista/issues/5334.
This PR also eliminates the need for https://github.com/pyvista/pyvista/pull/6173 and https://github.com/pyvista/pyvista/pull/6171 which are workarounds for the flaky tests.

Close #6171
Close #6173

### Details
- The dataset is saved as a `.vti` file instead of the original `.mhd` and `.zraw` files
- The `.vti` is saved directly with the repository instead of hosted on `pyvista/vtk-data`
- `load_frog_tissues` is added to load the new `.vti`
- `download_frog_tissue` is deprecated


user27182:
@pyvista-bot preview

user27182:
@tkoyama010 I believe we can sort out all of the documentation and testing issues with `frog_tissue` by saving and loading the image as `.vti` instead of `.mhd`. This is done in https://github.com/pyvista/pyvista/pull/6110/commits/1ff26b0ddc109926c7edf2f6008e6bc464a033f5
pyvista-examples-downloads-download_frog_tissue-1_00_00
sphx_glr_contouring_004, here are the generated build images, they are no longer buggy

If you prefer this solution, then #6173 and #6171 won't be necessary. Let me know and I can upload the `.vti` file to `vtk-data` and fix up this PR

tkoyama010:
> @tkoyama010 I believe we can sort out all of the documentation and testing issues with `frog_tissue` by saving and loading the image as `.vti` instead of `.mhd`. This is done in [1ff26b0](https://github.com/pyvista/pyvista/commit/1ff26b0ddc109926c7edf2f6008e6bc464a033f5) pyvista-examples-downloads-download_frog_tissue-1_00_00 sphx_glr_contouring_004, here are the generated build images, they are no longer buggy
> 
> If you prefer this solution, then #6173 and #6171 won't be necessary. Let me know and I can upload the `.vti` file to `vtk-data` and fix up this PR

I am +1 to this solution in this PR. Let's go this way :) I think `frog_tissue.vti` is a good example to add to pyvista itself.

user27182:
> I think `frog_tissue.vti` is a good example to add to pyvista itself.

Good idea, that will also make it easier to deprecate `download_frog_tissue`, use `load_frog_tissue` instead kind of thing. 

It's late for me locally, but I can update this tomorrow. Or feel free to make changes directly or open a separate PR with the vti file if you want this sorted asap

tkoyama010:
> > I think `frog_tissue.vti` is a good example to add to pyvista itself.
> 
> Good idea, that will also make it easier to deprecate `download_frog_tissue`, use `load_frog_tissue` instead kind of thing.
> 
> It's late for me locally, but I can update this tomorrow. Or feel free to make changes directly or open a separate PR with the vti file if you want this sorted asap

Awesome. Thanks!

user27182:
Still need to wait for the build tests to pass but this should be good to go otherwise.

user27182:
@pyvista-bot preview

examples/01-filter/contouring.py
```diff
@@ -84,8 +84,6 @@
...
 # Create labeled surfaces from 3D label maps (e.f. multi-label image segmentation)
 # using :func:`contour_labeled() <pyvista.ImageDataFilters.contour_labeled>`.
-# Requires VTK version 9.3
-if pv.vtk_version_info >= (9, 3):
-    label_map = pv.examples.download_frog_tissue()
-    mesh = label_map.contour_labeled()
-    mesh.plot(cmap="glasbey_warm", cpos="yx", show_scalar_bar=False)
+label_map = pv.examples.load_frog_tissue()
+mesh = label_map.contour_labeled(smoothing=True)
+mesh.plot(cmap="glasbey", cpos="yx", show_scalar_bar=False, clim=mesh.get_data_range())
```
tkoyama010, line 89: This is still necessary because there are cases where users run vtk versions earlier than v9.3.
```suggestion
if pv.vtk_version_info >= (9, 3):
    label_map = pv.examples.load_frog_tissue()
    mesh = label_map.contour_labeled(smoothing=True)
    mesh.plot(cmap="glasbey", cpos="yx", show_scalar_bar=False, clim=mesh.get_data_range())
```

tkoyama010 (changes requested):

examples/01-filter/contouring.py
user27182: Sounds good. This was the first time I had seem something like this in the docs. 

So this line (from a separate example in the docs) is considered a bug or bad practice since it can only execute with  vtk >= 9.1?
https://github.com/pyvista/pyvista/blob/0007ba4c536b5b795e85a381595def5238bf522c/examples/99-advanced/openfoam-tubes.py#L69

tkoyama010: Yes, it would be helpful to have a separate PR for that one.

tkoyama010:
@pyvista-bot preview

user27182:
The dataset gallery, and tests for the dataset loaders, make some strong assumptions about dataset names being unique. I think this is the cause of the failure here https://github.com/pyvista/pyvista/pull/6110/checks#step:12:29.

Might be easiest to change the name of the new `load_frog_tissue` function to something else., e.g. `load` +  `frog_tissues` (with an 's'), or maybe `frog_tissue_labels`, `frog_labels`, `frog_label_map`, `frog_segments`, etc.

Any preference?

tkoyama010:
> The dataset gallery, and tests for the dataset loaders, make some strong assumptions about dataset names being unique. I think this is the cause of the failure here https://github.com/pyvista/pyvista/pull/6110/checks#step:12:29.
> 
> Might be easiest to change the name of the new `load_frog_tissue` function to something else., e.g. `load` + `frog_tissues` (with an 's'), or maybe `frog_tissue_labels`, `frog_labels`, `frog_label_map`, `frog_segments`, etc.
> 
> Any preference?

+1 for `load_frog_tissues`.

user27182:
@pyvista-bot preview

user27182:
The build is finally succeeding 🎉

tkoyama010 (approved):
LGTM. Thanks for your works!

PR #6121, opened 2024-05-25
Remove walrus operator to read code in linear flow

tkoyama010:
### Overview

Fixed a section that was obviously difficult to read due to assignment expressions

@adeak Thanks for your advice.

### DIFF
--- a/doc/source/make_tables.py
+++ b/doc/source/make_tables.py
@@ -406,7 +406,8 @@ class ColorTable(DocTable):
 
 def _get_doc(func: Callable[[], Any]) -> Optional[str]:
     """Return the first line of the callable's docstring."""
-    return doc.splitlines()[0] if (doc := func.__doc__) else None
+    doc = func.__doc__
+    return doc.splitlines()[0] if (doc) else None
 
 
 def _get_fullname(typ: Type[Any]) -> str:
@@ -1173,8 +1174,10 @@ class DatasetPropsGenerator:
         if not isinstance(loader, _Downloadable):
             return None
         # Collect url names and links as sequences
-        names = [name] if isinstance(name := loader.source_name, str) else name
-        urls = [url] if isinstance(url := loader.source_url_blob, str) else url
+        name = loader.source_name
+        names = [name] if isinstance(name, str) else name
+        url = loader.source_url_blob
+        urls = [url] if isinstance(url, str) else url
 
         # Use dict to create an ordered set to make sure links are unique
         url_dict = {}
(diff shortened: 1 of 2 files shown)


user27182:
Probably a good style rule to avoid use of the walrus operator with ternary if-else blocks. They are indeed a bit confusing to read since the variable must be assigned in the middle of the expression. 

I tried looking for a rule with ruff or similar to enforce this but couldn't find anything.

user27182 (dismissed):

doc/source/make_tables.py
```diff
@@ -406,7 +406,8 @@ def get_row(cls, i, row_data):
 
 def _get_doc(func: Callable[[], Any]) -> Optional[str]:
     """Return the first line of the callable's docstring."""
-    return doc.splitlines()[0] if (doc := func.__doc__) else None
+    doc = func.__doc__
+    return doc.splitlines()[0] if (doc) else None
```
adeak, line 410: Parentheses were only necessary for the assignment expression:
```suggestion
    return doc.splitlines()[0] if doc else None
```

pyvista/examples/_dataset_loader.py
```diff
@@ -81,8 +81,9 @@ def num_files(self) -> int:
 
         If a path is a folder, the number of files contained in the folder is returned.
         """
-        path = [path] if isinstance(path := self.path, str) else path
-        return sum([1 if os.path.isfile(p) else len(_get_all_nested_filepaths(p)) for p in path])
+        path = self.path
+        paths = [path] if isinstance(path, str) else path
+        return sum([1 if os.path.isfile(p) else len(_get_all_nested_filepaths(p)) for p in paths])
```
adeak, line 86: Unrelated, but we can use a generator comprehension here
```suggestion
        return sum(1 if os.path.isfile(p) else len(_get_all_nested_filepaths(p)) for p in paths)
```

adeak (changes requested):

adeak:
> Probably a good style rule to avoid use of the walrus operator with ternary if-else blocks. They are indeed a bit confusing to read since the variable must be assigned in the middle of the expression.
> 
> I tried looking for a rule with ruff or similar to enforce this but couldn't find anything.

My recommendation would be to restrict the (mis)feature to the header line of conditional blocks or loops, as a general rule. If there's some consensus about this we could add it to our contributing guide.

adeak (approved):

user27182 (approved):

PR #6124, opened 2024-05-26
Add `warn_redundant_casts` to mypy config

user27182:
### Overview

See https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-cast-is-not-redundant-redundant-cast

I ran mypy locally with this option and there are no errors like this currently in the repo. But it may help prevent these cases in the future, e.g. https://github.com/pyvista/pyvista/pull/5571#discussion_r1615272845


tkoyama010 (approved):
LGTM

PR #6137, opened 2024-05-28
Add required import `from __future__ import annotations` 

user27182:
### Overview

Follow up from #5712. See comment https://github.com/pyvista/pyvista/pull/5712#issuecomment-2135776216.

In short, adding `from __future__ import annotations` is useful to prevent sphinx from auto-expanding type aliases in the docs. 

It can also be used to update the typing in the code base, e.g. to make use of Python 3.10 features such as [PEP 604](https://peps.python.org/pep-0604/).


user27182:
pre-commit.ci autofix

user27182:
pre-commit.ci autofix

user27182:
pre-commit.ci autofix

PR #6139, opened 2024-05-28
Upgrade type annotations to Python 3.10

user27182:
### Overview

Follow-up from #6137 

If that PR is accepted, it should be possible to upgrade all type annotations to make use of syntax from [PEP 585](https://peps.python.org/pep-0585/) and [PEP 604](https://peps.python.org/pep-0604/), for example.

This PR updates the config for `pyupgrade` to implement the changes.


user27182:
Closing in favor of #6158. Just easier than fixing all the merge conflicts for hundreds of files.

PR #6153, opened 2024-05-30
Add `decimate_polyline` filter

MatthewFlamm:
### Overview

Adds `decimate_polyline` filter.  This was requested in https://github.com/pyvista/pyvista/discussions/6144.

image


MatthewFlamm:
Looking for feedback on naming of the filter.  Not sure if this shoudl be `decimate_polylines` or `decimate_polyline` or something else.

tkoyama010:
> Looking for feedback on naming of the filter. Not sure if this should be `decimate_polylines` or `decimate_polyline` or something else.

+1 `decimate_polylines`.  It sounds natural.

tkoyama010:
@pyvista-bot preview

MatthewFlamm:
I added the `point_dtype` argument in the same manner as Cube 
https://github.com/pyvista/pyvista/blob/b8c4dc4e8c780b1dd6cb2be4e8f9bb60d860541b/pyvista/core/utilities/geometric_objects.py#L1175

but I'm a bit unsure that we should be setting the default to be 'float32' rather than 'float64'.  I think we discussed making this standardized in some manner.  I think maybe we should not implement it here, i.e. remove it from this PR, to avoid breaking changes if we decide to go a different way here.  Thoughts?

MatthewFlamm:
pre-commit.ci autofix

MatthewFlamm:
@zixiao-yin  if you want to test this, it implements your request directly in PyVista.

tkoyama010:
> I added the `point_dtype` argument in the same manner as Cube
> 
> https://github.com/pyvista/pyvista/blob/b8c4dc4e8c780b1dd6cb2be4e8f9bb60d860541b/pyvista/core/utilities/geometric_objects.py#L1175
> 
> but I'm a bit unsure that we should be setting the default to be 'float32' rather than 'float64'. I think we discussed making this standardized in some manner. I think maybe we should not implement it here, i.e. remove it from this PR, to avoid breaking changes if we decide to go a different way here. Thoughts?

It is my opinion that PyVista data is required to follow with numpy's array. I have looked  https://github.com/numpy/numpy/issues/9464 and it seems to be platform dependent which is reported as issue. I note that apparently there is no conclusion to this (7 years!).  I am reading this and trying to figure out what is the best way to go about it. It may be inconclusive...

tkoyama010:
This may be an int-only issue, though. I don't know about float array.

MatthewFlamm:
> It is my opinion that PyVista data is required to follow with numpy's array.

Another perspective is that this is determined through vtk. This is illustrated by the fact we set this parameter through vtk method `SetOutputPointsPrecision`.  For this PR, the vtk default is DOUBLE_PRECISION.

https://github.com/Kitware/VTK/blob/ebdd2031210c66dc150538da49344c2b82f65a79/Filters/Core/vtkDecimatePolylineFilter.cxx#L81

However, this PR used the `Cube` default which changes the vtk default.  I'm not sure here, so I think we should remove this from the PR, just to use the vtk default, which is how the rest of PyVista works today, other than `Cube`.

user27182:
> This may be an int-only issue, though. I don't know about float array.

FWIW, NumPy 2.0 fixes the platform-dependent issue for int. It will be 64bit by default

From https://numpy.org/devdocs/release/2.0.0-notes.html :

"The default integer type on Windows is now int64 rather than int32, matching the behavior on other platforms,"

tkoyama010:
@pyvista-bot preview

tkoyama010:
pre-commit.ci autofix

tkoyama010:
pre-commit.ci autofix

tkoyama010:
@pyvista-bot preview

tkoyama010:
> Edit: For future discussion, one option is to use the vtk default always, e.g. use default arg of None, but allow the user to specify single or double precision through the point_dtype arg. This would be a clear policy that avoids the numpy dependence on platform type problem.

OK. I am fine as is in this PR :)

examples/01-filter/decimate.py
```diff
@@ -67,3 +69,69 @@
 pl.reset_camera()
 pl.link_views()
 pl.show()
+
+###############################################################################
+# Decimate Polyline Mesh
+# ----------------------
+
+# Generate a fairly slow spiral polyline mesh.
```
tkoyama010, line 77:
```suggestion
#
# Generate a fairly slow spiral polyline mesh.
```

tkoyama010:
@pyvista-bot preview

tkoyama010 (approved):
LGTM

tkoyama010:
@pyvista-bot preview

MatthewFlamm:
Sorry @tkoyama010 for pushing late changes.  I just made a tweak to the gallery example to make the number of point comparison a little easier to find.  The documentation preview is super helpful for this, but does result in some last minute tweaks.  Let's deploy again, and then I'm done making changes.

tkoyama010:
> Sorry @tkoyama010 for pushing late changes. I just made a tweak to the gallery example to make the number of points comparison a little easier to find. The documentation preview is super helpful for this, but does result in some last minute tweaks. Let's deploy again, and then I'm done making changes.

Please don't worry. It's fun to look at documents that are changing.

tests/core/test_polydata_filters.py
```diff
@@ -58,6 +58,33 @@ def test_identical_boolean(sphere):
...
+
+    decimated = poly_circle.decimate_polylines(0.5)
+    # Allow some leeway for approximtely 50%
+    assert decimated.n_points >= 14
+    assert decimated.n_points <= 16
+
+    # low maximum error will prevent decimation.
+    # Since this is a regular shape, no decimation occurs at all with suitable choice
+    decimated = poly_circle.decimate_polylines(0.5, maximum_error=0.0001)
+    assert decimated.n_points == 30
```
tkoyama010, line 78:
```suggestion
def test_decimate_polylines(poly_circle):
    assert poly_circle.n_points == 30
    decimated = poly_circle.decimate_polylines(0.5)
    # Allow some leeway for approximtely 50%
    assert decimated.n_points >= 14
    assert decimated.n_points <= 16

def test_decimate_polylines_maximum_error(poly_circle):
    assert poly_circle.n_points == 30
    # low maximum error will prevent decimation.
    # Since this is a regular shape, no decimation occurs at all with suitable choice
    decimated = poly_circle.decimate_polylines(0.5, maximum_error=0.0001)
    assert decimated.n_points == 30
```

tkoyama010 (changes requested):
Let's check only one thing in one test function so that it is easy to see where errors occur!

tkoyama010:
pre-commit.ci autofix

tkoyama010:
@pyvista-bot preview

adeak:
> Looking for feedback on naming of the filter. Not sure if this should be `decimate_polylines` or `decimate_polyline` or something else.

Late to the party, but one potential consideration is that [the VTK filter](https://vtk.org/doc/nightly/html/classvtkDecimatePolylineFilter.html#details) uses singular `DecimatePolyline`, and has a warning that says
>     This algorithm is a very simple implementation that overlooks some potential complexities. For example, if a vertex is multiply connected, meaning that it is used by multiple distinct polylines, then the extra topological constraints are ignored. This can produce less than optimal results.

which sounds like it's only safe to use when there's a single polyline in the `PolyData`. But I don't have a personal stance on the naming here.

---

> This may be an int-only issue, though. I don't know about float array.

For the record, as @user27182 hinted this is (in a few years to be "was") an int-only issue. The history is that the ancestor of NumPy used the C `long` type, which happens to be 64-bit on 64-bit systems _except Windows where it's 32 bits_. By the time they realised the issue they couldn't affort a breaking change to bump up default NumPy ints to 64 bits on Windows. They decided they could only make this breaking change in NumPy 2.0. Fortunately `float` (in C) and `double` are well-defined so they don't share the same ambiguity.

MatthewFlamm:
Sorry for dropping this for a long time once it was very close to the finish line.

> which sounds like it's only safe to use when there's a single polyline in the `PolyData`. But I don't have a personal stance on the naming here.

After the time off, I am now aligned to this way of thinking.  The filter is mainly designed to decimate a single polyline. Or polyline cell by polyline cell at the very most.  We should also be deferential to vtk naming when it makes sense.

> which sounds like it's only safe to use when there's a single polyline in the PolyData

I added this warning into the docstring.

tkoyama010:
pre-commit.ci autofix

MatthewFlamm:
@tkoyama010 the last thing that is needed (unless there are other reviews) is to change the version addedd.  With #6849, I'm not sure if this will be 0.45 or 0.46 for this PR.  Can you advise?

tkoyama010:
> @tkoyama010 the last thing that is needed (unless there are other reviews) is to change the versionadded.  With #6849, I'm not sure if this will be 0.45 or 0.46 for this PR.  Can you advise?

Let's add this to v0.45!

MatthewFlamm:
@user27182 it looks like we are in an inbetween spot for typing annotations in the filters.  I also don't want to create more merge complications for other PRs you have.  Is ignoring the typing issue as in https://github.com/pyvista/pyvista/pull/6153/commits/33f3ef73f6fb1a29d89d7eea8ba8ca28358e0e39 suggested, or typing the `self` better at this point?  I followed the pattern from `decimate_pro` here.

pyvista/core/filters/poly_data.py
```diff
@@ -1138,6 +1145,123 @@ def decimate_pro(
...
+
+        .. versionadded:: 0.45.0
+
+        Parameters
+        ----------
+        reduction : float
+            Reduction factor. A value of 0.9 will leave 10% of the
+            original number of vertices.
+
+        maximum_error : float, default: 1.0
```
MatthewFlamm, line 1167:
```suggestion
        maximum_error : float, default: 10.
```

pyvista/core/filters/poly_data.py
```diff
@@ -1138,6 +1145,123 @@ def decimate_pro(
 
         return mesh
 
+    def decimate_polyline(
+        self,
+        reduction,
+        maximum_error=10,
+        inplace=False,
+        progress_bar=False,
```
MatthewFlamm, line 1153:
```suggestion
        reduction: float,
        maximum_error: float=10.,
        inplace: bool=False,
        progress_bar: bool=False,
```

adeak:
Side note: wouldn't most of these "selves" be annotatable with [`Self`](https://docs.python.org/3/library/typing.html#typing.Self)? (New in 3.11, not sure what typing version constraints we have.)

user27182:
> Side note: wouldn't most of these "selves" be annotatable with [`Self`](https://docs.python.org/3/library/typing.html#typing.Self)? (New in 3.11, not sure what typing version constraints we have.)

`Self` is probably what we _should_ use, and usually all the new typing constructs are back-ported through `typing-extensions` (so likely could use it with Python 3.9).

But for the filters we have to annotate `self` explicitly as a dataset to resolve a bunch of type errors. See https://github.com/pyvista/pyvista/pull/6870

MatthewFlamm:
Typing as `Self` is not correct here as the actual type will be `PolyData` (similar for other filter methods), which has methods that don't exist on the filters classes.  Those methods that don't exist are heavily used.  A Protocol would be one solution, or redoing the inheritance model.

MatthewFlamm:
> But doing `# type: ignore` as in [33f3ef7](https://github.com/pyvista/pyvista/commit/33f3ef73f6fb1a29d89d7eea8ba8ca28358e0e39) is also fine.

I'm going to leave it is as, as it is consistent with the current code in this file.  It is better to address the self typing issue 'change' in a single PR IMO.

user27182:
> Typing as `Self` is not correct here as the actual type will be `PolyData` (similar for other filter methods), which has methods that don't exist on the filters classes. Those methods that don't exist are heavily used. A Protocol would be one solution, or redoing the inheritance model (or a combo of both).
> 
> Edit: I should have said that I believe `Self` is typically used for purposes of the return type and not for specifying the type of `self` itself.

Seems like using `Self` is ideal for something like our `DataObject` class with the `copy` method. If you do `mesh.copy()` you want the return type to be the same as the input type, e.g. `PolyData -> PolyData`). Otherwise, without using `Self`, `copy` would return a more general `DataObject` type,  (and mypy would complain if you try using a PolyData-specific method thereafter).

Notable, is that `Self` is just a wrapper for a bound `TypeVar`. And the good news is that #6918 does this exactly: it defines bound `TypeVar`s to annotate self for filters explicitly. So, for any filters which return the same type as the input, using our pyvista-specific types (e.g. `self: ConcreteDataSetType`) instead of `self: Self` should achieve the same result.

pyvista/core/filters/poly_data.py
```diff
@@ -1138,6 +1145,123 @@ def decimate_pro(
 
         return mesh
 
+    def decimate_polyline(
+        self,
```
tkoyama010, line 1149: Let's prepare for additional arguments before the `progress_bar`.
```suggestion
        self,
        *,
```

tkoyama010 (changes requested):

pyvista/core/filters/poly_data.py
MatthewFlamm: Even though I'm +1 for this style in general, I think we should have a standard style for the whole project. It will be surprising to users if only some filters have it. What do you think about just making `progress_bar` keyword only?  This has lower downsides.

user27182: Using `*` for kwargs is already implemented in a lot of new code, so there is already an inconsistency there. I don't think using it (or not) here will make-or-break this though, so I don't think using `*` (or not) should block this PR.

But, IMO the benefits of keyword-only args far outweigh this inconsistency. I would even consider the use of any positional `bool` arg as a bad practice in general which should be avoided. It's not clear if or when keyword-only args will be enforced for legacy code, but in the mean time I don't think we should go out of way to avoid using it in new code.

user27182 (dismissed):
LGTM, though I do think using `*` for all kwargs is a good idea.

pyvista/core/filters/poly_data.py
MatthewFlamm: Ok I'm +1, but the suggestion needs to be modified as `reduction` should not be keyword only.  Will do this soon.

tkoyama010 (approved):
LGTM

pyvista/core/filters/poly_data.py
banesullivan: +1 from me and I would love to standardize this across the entire code base

PR #6156, opened 2024-05-30
Ensure  compatibility with NumPy 2.0

adeak:
Handle remaining NumPy 2.0 compatibility issues.

- [x] fix garbage collection errors (done upstream after report)
- [x] fix array copy flag errors
- [x] check if `pyvista_ndarray.__array_wrap__()` scalar return logic should be different -> [not yet](https://github.com/pyvista/pyvista/pull/6156#discussion_r1625060601)
- [x] check `ImageData.image_threshold()` special casing and related errors
- [x] check image regression warnings -> [unrelated to NumPy version, due to minor variation across OSes](https://github.com/pyvista/pyvista/pull/6156#issuecomment-2146169193)

Closes https://github.com/pyvista/pyvista/issues/5533.


pyvista/core/pyvista_ndarray.py
```diff
@@ -109,14 +109,14 @@ def __setitem__(self, key: Union[int, NumpyArray[int]], value):
         if dataset is not None and dataset.Get():
             dataset.Get().Modified()
 
-    def __array_wrap__(self, out_arr, context=None):
+    def __array_wrap__(self, out_arr, context=None, return_scalar=False):
```
adeak, line 112: I need to look into the exact behaviour of the new `return_scalar` kwarg, because I suspect it might have something to do with our existing "return scalar" hack at the end.

pyvista/core/_validation/_cast_array.py
```diff
@@ -103,10 +103,11 @@ def _cast_to_numpy(arr, /, *, as_any=True, dtype=None, copy=False, must_be_real=
             if copy:
                 out = out.copy()
         else:
-            out = np.array(arr, dtype=dtype, copy=copy)
+            # handle copy flag manually because numpy>=2.0 raises if np.array(..., copy=False) must copy
+            out = np.array(arr, dtype=dtype) if copy else np.asarray(arr, dtype=dtype)
```
adeak, line 107: This is too ugly for my taste, but I didn't find a more elegant solution:
1. `np.array(..., copy=False)` raises on old NumPy if copying fails
2. `np.array(..., copy=None)` which doesn't raise doesn't exist on old NumPy
3. `np.asarray(..., copy=False)` doesn't exist on old NumPy.

So I think there's no one function call we can use here.

adeak:
The two remaining test failures as of opening this PR come from `ImageData.image_threshold()`. And it has to do with this block in the filter:
https://github.com/pyvista/pyvista/blob/b8c4dc4e8c780b1dd6cb2be4e8f9bb60d860541b/pyvista/core/filters/image_data.py#L421-L434

Now, it would be easy to tweak that filter (probably by removing the NumPy version condition, depending on how our CI matrix looks on various OSes), but we seem to have this logic: "linux/mac doesn't threshold correctly on NumPy < 2.0", but now it also doesn't threshold on NumPy 2.0 (at least on my Linux). I have a suspicion that what's really going on is "it only works on Windows, NumPy < 2.0" which would likely have to do with the fact that "Windows NumPy < 2.0" is the only system where the default int size has 32 bits (this was fiiiinally fixed for NumPy 2.0).

So I want to take a closer look at this thresholding logic to see what's really going on here. I suspect it's Windows with NumPy < 2.0 that's somehow broken and needs to be special cased...

adeak:
Ping @MatthewFlamm.

pyvista/core/_validation/_cast_array.py
user27182: `np.asarray` only copies if needed for both NumPy 1.26 and 2.0. So, you could probably refactor the whole function instead, and only copy explicitly at the end if needed. Something like:

``` python
# Do casting
out = np.asanyarray(arr, dtype=dtype) if as_any else np.asarray(arr, dtype=dtype)

# Array may have been copied by the casting. Only copy if necessary.
if copy and out is arr:
    out = out.copy()
```

user27182:
> The two remaining test failures as of opening this PR come from `ImageData.image_threshold()`. And it has to do with this block in the filter:
> 
> https://github.com/pyvista/pyvista/blob/b8c4dc4e8c780b1dd6cb2be4e8f9bb60d860541b/pyvista/core/filters/image_data.py#L421-L434
> 
> Now, it would be easy to tweak that filter (probably by removing the NumPy version condition, depending on how our CI matrix looks on various OSes), but we seem to have this logic: "linux/mac doesn't threshold correctly on NumPy < 2.0", but now it also doesn't threshold on NumPy 2.0 (at least on my Linux). I have a suspicion that what's really going on is "it only works on Windows, NumPy < 2.0" which would likely have to do with the fact that "Windows NumPy < 2.0" is the only system where the default int size has 32 bits (this was fiiiinally fixed for NumPy 2.0).
> 
> So I want to take a closer look at this thresholding logic to see what's really going on here. I suspect it's Windows with NumPy < 2.0 that's somehow broken and needs to be special cased...

See #6100.

The issue with `image_threshold` was initially identified as an issue with the docs build , i.e. linux. But the tests were added with that PR were also found to fail with some MacOS builds.

This may not be NumPy 2.0 specific. I think I added this condition `int(np.__version__.split('.')[0]) < 2 ` because I saw that the NumPy 2.0 CI checks were passing. But in hindsight, I think these checks always passed. So, this condition can probably be removed.

adeak:
Thanks for the tips @user27182.

user27182:
> This may not be NumPy 2.0 specific. I think I added this condition `int(np.__version__.split('.')[0]) < 2 ` because I saw that the NumPy 2.0 CI checks were passing. But in hindsight, I think these checks always passed. So, this condition can probably be removed.
> 
> EDIT: By "I think these checks always passed" , I mean that there is always green checkmark. So, the actual 2.0 tests may have still been failing for these builds, but the workflow was marked nevertheless marked as "succeeded"

I had a look at the CI logs for #6100. The checks from the first few commits show failing tests (i.e. the bug is present) for all macos/linux workflows. And if we look at the NumPy2.0 tests, we can see they also failed here:
https://github.com/pyvista/pyvista/actions/runs/9144044806/job/25141449296?pr=6100#step:12:1137

This confirms that using NumPy 2.0 did not fix this issue, so the condition can be removed. So, the logic could be changed to something like:

``` python
      # Integer scalars won't threshold correctly on some systems. Cast to float in these cases.
      has_int_dtype = np.issubdtype(
          array_dtype := self.active_scalars.dtype,
          int,
      ) and array_dtype != np.dtype(np.uint8)
      if has_int_dtype and platform.system() in ['Darwin', 'Linux']:
          self[scalars] = self[scalars].astype(float, casting='safe')
```

tests/core/test_dataset_filters.py
```diff
@@ -1029,7 +1029,7 @@ def test_glyph_color_mode(sphere, color_mode):
 
 def test_glyph_raises(sphere):
     with pytest.raises(ValueError, match="Invalid color mode 'foo'"):
-        sphere.glyph(color_mode="foo")
+        sphere.glyph(color_mode="foo", scale=False, orient=False)
```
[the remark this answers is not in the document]
adeak: Correct, these are only here to reduce the number of warnings I saw during a local build. [The commit list](https://github.com/pyvista/pyvista/pull/6156/commits) shows the marginally or not at all relevant commits. Sorry, I forgot to leave a comment about these when I rushed to open this PR.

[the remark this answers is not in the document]
MatthewFlamm: Thanks, should be fine here if it helps sleuthing.  This code definitely throws warnings, so it all checks out.

pyvista/core/pyvista_ndarray.py
adeak: In NumPy 1.x with something like `ndarray.sum()` the default `False` is kept, whereas on NumPy 2.0 `return_scalar=True` is passed, and in something like `arr * 1` (where the result is not an array scalar) we get `return_scalar=False` passed. Despite this, calling `super().__array_wrap__(self, out_arr, context, return_scalar)` doesn't squeeze out the result of `arr.sum()` from a minimal subclass that tries to just use `super()`.

So for NumPy 2.x we could slightly change the method to do
```python
if return_scalar:
    return out_arr[()]
...  # fall back on super()
```
but as long as we support NumPy 1.x we have to do what we do now: investigating the dimensions of the result array. Nothing to do here for now.

tests/core/test_dataset_filters.py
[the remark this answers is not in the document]
adeak: To be honest it's more of a "came across these warnings that we really don't need to emit". Ideally we'd have zero warnings, and even more ideally, warnings turned into errors during testing. But that's always risky due to downstream warnings. We have plenty of our own warnings, quite a few from trame stuff. But I'm not touching those :smile:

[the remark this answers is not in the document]
adeak: Speaking of which, there are two other warnings of the same ilk that we might want to fix... :smile: 
```
tests/core/test_dataset_filters.py::test_glyph_color_mode[scale]
tests/core/test_dataset_filters.py::test_glyph_color_mode[scalar]
tests/core/test_dataset_filters.py::test_glyph_color_mode[vector]
  /home/adeak/pyvista_dev/pyvista/pyvista/core/filters/data_set.py:2356: UserWarning: No data to use for scale. scale will be set to False.
    warnings.warn("No data to use for scale. scale will be set to False.")

tests/core/test_dataset_filters.py::test_glyph_color_mode[scale]
tests/core/test_dataset_filters.py::test_glyph_color_mode[scalar]
tests/core/test_dataset_filters.py::test_glyph_color_mode[vector]
  /home/adeak/pyvista_dev/pyvista/pyvista/core/filters/data_set.py:2383: UserWarning: No vector-like data to use for orient. orient will be set to False.
    warnings.warn("No vector-like data to use for orient. orient will be set to False.")
```
But again completely off-topic here...

adeak:
I've now checked a local build with NumPy 1.26.4 and nightly, and I get the exact same warnings (so all the image regression warnings I see are unrelated). Although some warnings could/should be fixed, they don't concern the NumPy version bump, I'll only have to look at the errors (which we've already discussed).

adeak:
What's the general feeling about fixing unrelated warnings [such as these](https://github.com/pyvista/pyvista/pull/6156#discussion_r1625072472) here? Of course only ones that can be fixed trivially. E.g. there's a warning from `load_theme()` I looked at, and it seemed like a nightmare to fix (short of explicitly silencing it).

MatthewFlamm:
> What's the general feeling about fixing unrelated warnings [such as these](https://github.com/pyvista/pyvista/pull/6156#discussion_r1625072472) here? Of course only ones that can be fixed trivially. E.g. there's a warning from `load_theme()` I looked at, and it seemed like a nightmare to fix (short of explicitly silencing it).

The warnings from `load_theme` are indicative that this logic is problematic.  I had also noticed this in https://github.com/pyvista/pyvista/pull/5119#issuecomment-1962027567.  Let's not touch that here.  Since you've already touched the code for suppressing warnings for `glyph` it makes sense if you want to for easier testing here.

adeak:
I've opened https://github.com/pyvista/pyvista/pull/6210 to temporarily give an upper version pin for our NumPy dependency. Cross-posting my plan here because it's also relevant to this PR:

> My plan/proposal is:
> 1. merge this PR quickly to protect `main` from the NumPy 2.0 release until we're compatible
> 2. remove the version restriction in https://github.com/pyvista/pyvista/pull/6156
> 3. probably make the NumPy nightly build in our CI matrix mandatory, at least until 2.0 is released.

adeak:
Thanks @MatthewFlamm, that was also precisely my plan :)

MatthewFlamm:
I tested removing the check for numpy version as in 
 https://github.com/pyvista/pyvista/pull/6156#issuecomment-2140961850 and it works for me.

@adeak do you have time to close this out soon?  If not, I can make the changes needed to get this green and mandatory in the CI.  To make mandatory in the CI, we need to remove the two `continue-on-error` lines, one is below.  This is important, as it is understandable that others are assuming the current green check mark means "Passing".

https://github.com/pyvista/pyvista/blob/6698923f73a1032bb3a42131d265b98b891e4836/.github/workflows/testing-and-deployment.yml#L224

adeak:
Thanks @MatthewFlamm, I originally made the test mandatory on my branch, I just reverted that so as not to conflict with your numpy nightly PR.

I also know the version check removal "works", I just wanted to take a closer look at how it's broken without the check (but I prpbably won't do that).

I can wrap this up on the weekend if that's an acceptable time frame.

.github/workflows/testing-and-deployment.yml
```diff
@@ -166,7 +166,7 @@ jobs:
             numpy-version: "default"
           - python-version: "3.12"
             vtk-version: "latest"
-            numpy-version: "pre"
+            numpy-version: "nightly"
```
adeak, line 169: Renamed this for clarity. We can name it back when we revert to pre-releases once 2.0 is released (if we want to do that).

tests/core/test_dataset_filters.py
```diff
@@ -1030,12 +1030,14 @@ def test_glyph_orient_and_scale():
 
 @pytest.mark.parametrize('color_mode', ['scale', 'scalar', 'vector'])
 def test_glyph_color_mode(sphere, color_mode):
-    assert sphere.glyph(color_mode=color_mode)
+    # define vector data
+    sphere.point_data['velocity'] = sphere.points[:, [1, 0, 2]] * [-1, 1, 0]
+    sphere.glyph(color_mode=color_mode)
```
adeak, line 1035: This was the source of 2 remaining fixable warnings. Two remarks:
1. I removed the `assert` because even empty `PolyData` are truthy
2. the real test of this feature would be to do an image regression test... I don't know if we could squeeze it in somewhere in `tests/plotting`. Would that make sense? Should I open an issue for it?

pyvista/core/_validation/_cast_array.py
adeak: Done in https://github.com/pyvista/pyvista/pull/6156/commits/62d915a2a0f5227a4bc518eedcbf5b5a0354753c.

adeak:
This is ready for review.

adeak:
[3.12 with NumPy nightly](https://github.com/pyvista/pyvista/actions/runs/9423414205/job/25961737337?pr=6156):
```
===== 1691 passed, 5 skipped, 2 xfailed, 13 warnings in 1064.07s (0:17:44) =====
```

Warnings:
```
=============================== warnings summary ===============================
tests/plotting/jupyter/test_trame.py::test_trame_plotter_ui[vue2]
tests/plotting/jupyter/test_trame.py::test_trame_plotter_ui[vue3]
  /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/wslink/backends/aiohttp/__init__.py:89: NotAppKeyWarning: It is recommended to use web.AppKey instances for keys.
  https://docs.aiohttp.org/en/stable/web_advanced.html#application-s-config
    self.app["state"] = {}

tests/plotting/jupyter/test_trame.py::test_trame_plotter_ui[vue2]
tests/plotting/jupyter/test_trame.py::test_trame_plotter_ui[vue3]
  /home/runner/work/pyvista/pyvista/pyvista/trame/ui/__init__.py:54: UserWarning: Suppress rendering on the plotter is changed to True
    warnings.warn(

tests/plotting/jupyter/test_trame.py::test_trame_plotter_ui[vue2]
tests/plotting/jupyter/test_trame.py::test_trame_plotter_ui[vue3]
  /home/runner/work/pyvista/pyvista/pyvista/trame/ui/__init__.py:54: UserWarning: Suppress rendering on the plotter is changed to False
    warnings.warn(

tests/plotting/test_charts.py::test_chart_common[chart_mpl]
  /home/runner/work/pyvista/pyvista/pyvista/plotting/charts.py:4797: UserWarning: No artists with labels found to put in legend.  Note that artists whose label start with an underscore are ignored when legend() is called with no argument.
    legend = self._fig.axes[0].legend()

tests/plotting/test_picking.py::test_fly_to_right_click
  /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pytest_pyvista/pytest_pyvista.py:137: UserWarning: pyvista test generated image dir: debug_images does not yet exist.  Creating dir.
    warnings.warn(

tests/plotting/test_plotting.py::test_chart_plot
  /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/pytest_pyvista/pytest_pyvista.py:218: UserWarning: test_chart_plot Exceeded image regression warning of 200.0 with an image error of 464.9084967320269
    warnings.warn(

tests/plotting/test_theme.py::test_load_theme
tests/plotting/test_theme.py::test_save_before_close_callback
  /home/runner/work/pyvista/pyvista/pyvista/plotting/themes.py:1552: UserWarning: The jupyter_extension_available flag is read only and is automatically detected.
    warnings.warn(

tests/plotting/test_theme.py::test_trame_config
  /home/runner/work/pyvista/pyvista/pyvista/plotting/themes.py:1531: UserWarning: Enabling server_proxy will disable jupyter_extension
    warnings.warn("Enabling server_proxy will disable jupyter_extension")

tests/plotting/test_theme.py::test_trame_config
  /home/runner/work/pyvista/pyvista/pyvista/plotting/themes.py:1567: UserWarning: Enabling jupyter_extension will disable server_proxy
    warnings.warn("Enabling jupyter_extension will disable server_proxy")
```

I'm not sure why I don't see some of the chart warnings in a local build. But the remaining warnings seem to be
1. `trame` x8
2. `load_theme()` that we already discussed is a pain in the ass to fix
3. something about a generated image dir not yet existing (i.e. only affects CI, not an actual runtime warning)
4. chart plot large image regression (bit _too_ large, but I don't see the same issue locally)
5. the aforementioned warning in `tests/plotting/test_charts.py::test_chart_common[chart_mpl]` that I don't see locally (on Debian linux on Python 3.11, although CI has the same warning for 3.11).

adeak:
Last message for now: the image regression warning comes from this image
tmp_chart_plot

Locally I see the yellow orbit dots in the bottom right subfigure shifting upward a bit. I'm sure this is what leads to the warning, and I guess the same offset is more pronounced in CI. I don't know why this shifts, but it seems harmless enough (and hopefully unrelated to our chart functionality).

tests/core/test_dataset_filters.py
MatthewFlamm: This just tests calling the vtk method.  While a test is nice to be in the very safe side, I don't think it is necessary.  Definitely not for this PR.

MatthewFlamm (approved):
Nice!

tests/core/test_polydata.py
```diff
@@ -1234,7 +1234,7 @@ def test_empty_regular_faces():
 
 
 def test_regular_faces_mutable():
-    points = [[1, 1, 1], [-1, 1, -1], [1, -1, -1], [-1, -1, 1]]
+    points = [[1.0, 1, 1], [-1, 1, -1], [1, -1, -1], [-1, -1, 1]]
```
tkoyama010, line 1237: If this is not what intended, it is better to revert.
```suggestion
    points = [[1, 1, 1], [-1, 1, -1], [1, -1, -1], [-1, -1, 1]]
```

tkoyama010 (changes requested):

pyvista/core/_validation/_cast_array.py
```diff
@@ -99,7 +99,7 @@ def _cast_to_numpy(
             * is a subclass of ``np.ndarray`` and ``as_any`` is ``False``.
 
     must_be_real : bool, default: True
-        Raise a TypeError if the array does not have real numbers, i.e.
+        Raise a ``TypeError`` if the array does not have real numbers, i.e.
```
tkoyama010, line 102: Since the docstring modification has nothing to do with this PR, I have separated it to #6227 and #6228.  Maybe I am being too particular, but I want to keep the PR differential beautiful.

tests/core/test_dataset_filters.py
adeak: Thanks @MatthewFlamm, I won't open an issue then. (I had figured we could test that we're dispatching to the correct VTK method.)

tests/core/test_polydata.py
adeak: This is intended, that's why it's there. This causes the resulting numpy array to be float rather than int, suppressing another warning.

This is admittedly lazy, but there was already precedent for this in the module and I wanted to make minimal changes to resolve trivial warmings.

If you'd like I can make these hacky cases nicer, at the cost of a larger unrelated diff.

pyvista/core/_validation/_cast_array.py
adeak: I don't think this is a fruitful way of spending time, and it's a pretty bad experience for a contributor when PRs get micro-managed like this. Even if you feel so strongly about this we can discuss and you can let me handle my own work. Same issue why I was bothered by you pulling out a random comment into a new issue (without editing the issue to at least give it appropriate context).

Practicality beats purity and for sure we've historically mixed in unrelated but importat stuff in PRs. For instance, it would've been much simpler and more transparent to fix the recent broken main CI in a single PR rather than ~5 parallel ones, each with red builds of their own.

You do know I care about purity where it matters: I kept badgering others to keep auto-format PRs clean to preserve our git blame.

tests/core/test_polydata.py
tkoyama010: Thank you. Can we add our intent in the comments?

pyvista/core/_validation/_cast_array.py
tkoyama010: I see. Perhaps I am being a bit dogmatic. Your pointed comments are very helpful. Please keep the advice coming.

tests/core/test_polydata.py
adeak: It felt too lazy to slap a comment on what's really just a lazy kludge, so I decided to clean these up properly in https://github.com/pyvista/pyvista/pull/6156/commits/bf0c73eb8237c63d7d9c067addea8236210b049d.

tkoyama010 (approved):
LGTM

banesullivan (approved):

RobPasMue:
Hi @tkoyama010 @banesullivan @MatthewFlamm! Following the merge of this PR, are you expecting to release a new minor version of PyVista that supports Numpy 2.x? Is there any tentative date for that?

tkoyama010:
> Hi @tkoyama010 @banesullivan @MatthewFlamm! Following the merge of this PR, are you expecting to release a new minor version of PyVista that supports Numpy 2.x? Is there any tentative date for that?

Thanks for pointing this. This is labelled as "bug". So I will make patch release for this :)

PR #6158, opened 2024-05-30
Update type annotations in `/pyvista/trame`

user27182:
### Overview

With the `from __future__ import annotations` import added in  #5712, it should be possible to upgrade all type annotations to make use of syntax from [PEP 585](https://peps.python.org/pep-0585/) and [PEP 604](https://peps.python.org/pep-0604/), for example.

This PR updates the config for ~`pyupgrade`~ `ruff` to implement the changes.

EDIT: The upgrades for most of the repository have been merged in #6235. The scope of this PR is now reduced to files in `/pyvista/trame`


.pre-commit-config.yaml
```diff
@@ -83,7 +83,7 @@ repos:
     rev: v3.15.2
     hooks:
       - id: pyupgrade
-        args: [--py38-plus, --keep-runtime-typing]
+        args: [--py310-plus, --keep-runtime-typing]
```
tkoyama010, line 86: `pyupgrade` also affects syntax other than typing. Therefore, I was concerned about possible syntax errors in python<3.10. However, I decide it is not too late to take care of it after something happen.

user27182: We could drop `pyupgrade` altogether and use ruff instead, see with https://docs.astral.sh/ruff/rules/#pyupgrade-up. This way it may be easier to control different options rather than having it as a kind of all-or-nothing with `pyupgrade`. This would be similar to dropping isort from #5712.

user27182: This has been reverted to `--py38-plus`. With the `--keep-runtime-typing` option, `ruff` will treat fixes such as [UP006](https://docs.astral.sh/ruff/rules/non-pep585-annotation/) as "unsafe" and won't automatically apply the fixes. This is a good thing, since in some (very few/relatively uncommon) cases we need the runtime behavior (e.g. for some aliases we need to use `Tuple`, and not `tuple`). With the `--py38-plus` oftion ruff will simply complain about using `Tuple` but won't autofix this.

tkoyama010:
Thank you. This makes the type annotation easier to read.

tkoyama010 (dismissed):

pyvista/core/dataset.py
```diff
@@ -49,6 +42,12 @@
 from .utilities.misc import check_valid_vector
 from .utilities.points import vtk_points
 
+if TYPE_CHECKING:
```
user27182, line 45:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

tkoyama010:
We will merge this after merge #6159.

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

tkoyama010 (dismissed):

tkoyama010:
pre-commit.ci autofix

pyvista/core/grid.py
```diff
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Sequence
```
tkoyama010, line 5:
```suggestion
```

pyvista/examples/_dataset_loader.py
```diff
@@ -32,6 +32,8 @@
 from __future__ import annotations
 
 from abc import abstractmethod
+from collections.abc import Callable
+from collections.abc import Sequence
```
tkoyama010, line 36:
```suggestion
```

pyvista/plotting/charts.py
```diff
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Sequence
```
tkoyama010, line 5:
```suggestion
```

tkoyama010 (dismissed):

pyproject.toml
```diff
@@ -94,7 +94,7 @@ pyvista = [
 [tool.black]
 line-length = 100
 skip-string-normalization = true
-target-version = ['py39']
+target-version = ['py38']
```
user27182, line 97: This can be reverted, I don't think it was intentional and I may have confused it with the target version for `ruff`.
```suggestion
target-version = ['py39']
```

banesullivan (approved):

PR #6159, opened 2024-05-31
Dropping pyupgrade and move to pyupgrade-up rule in ruff

tkoyama010:
### Overview

Dropping `pyupgrade ` and moving to `pyupgrade-up` rule in `ruff`.

Close #6137

### Details

- See https://docs.astral.sh/ruff/rules/#pyupgrade-up.


user27182 (approved):
LGTM. I also tested this config with #6158 using `target-version = 'py310'` and it's working great.

PR #6161, opened 2024-05-31
Add `show-fixes` arg to `ruff` to make it clear which rule(s) are being applied

user27182:
### Overview
The `ruff` config uses `--fix` which auto-fixes code. In theory this should only be done for "safe" fixes, but sometimes the fix needs to be reverted and ignored. However, this is not currently easy to do since `ruff` will simply print something like:

```
Found 1 error (1 fixed, 0 remaining).
```

This leaves you wondering which code to use to ignore the fix. 

This PR adds the `--show-fixes`  option to make the fix more verbose, e.g.:

```
Fixed 1 error:
- pyvista/core/_typing_core/_array_like.py:
    1 × UP035 (deprecated-import)
```

Since the rule applied for the fix is now shown, it makes it easy to ignore the fix, e.g. using `# noqa: UP035` for this example.


tkoyama010 (approved):
LGTM

PR #6162, opened 2024-05-31
Remove `import annotations` to speed up documentation build time

user27182:
### Overview

The build times seem to be a bit longer after https://github.com/pyvista/pyvista/pull/5712 was merged. I wonder if removing the `from __future__ import annotations` can speed up the build.


tkoyama010:
The reason for the longer build time is cache doesn't exist due to a failed build in main. This is due to an error in a newer version of Matplotlib, which I am trying to fix with #6163.

tkoyama010:
Do you mean longer apart from that?

user27182:
Ah ok I didn't realize the CI checks on the main branch were currently broken. I (incorrectly) assumed that the cache issues were resolved before #5712 was merged, so I was expecting the build times to go back to normal.

PR #6165, opened 2024-05-31
Make imports from `collections` and `collections.abc` consistent

user27182:
### Overview

In #6158, many imports from the `typing` module are deprecated and replaced with imports from `collections.abc` instead. E.g. `from typing import Sequence` -> `from collections.abc import Sequence`.

However, some modules do not import `Sequence` directly into the local namespace, and instead simply do `import collections.abc`. This makes for an inconsistent import scheme across modules.

This PR fixes this and imports from `collections` and `collections.abc` directly into the local namespace to make these imports consistent.


tkoyama010 (approved):
LGTM

pyvista/core/composite.py
```diff
@@ -889,7 +889,7 @@ def reverse(self):
         ['sphere', 'cube']
 
         """
-        # Taken from implementation in collections.abc.MutableSequence
+        # Taken from implementation in MutableSequence
```
user27182, line 892: Revert this change
```suggestion
        # Taken from implementation in collections.abc.MutableSequence
```

pyvista/core/composite.py
```diff
@@ -57,7 +57,7 @@ class MultiBlock(
 
     .. versionchanged:: 0.36.0
        ``MultiBlock`` adheres more closely to being list like, and inherits
-       from :class:`collections.abc.MutableSequence`.  Multiple nonconforming
+       from :class:`MutableSequence`.  Multiple nonconforming
```
user27182, line 60: Revert this change
```suggestion
       from :class:`collections.abc.MutableSequence`.  Multiple nonconforming
```

pyvista/core/partitioned.py
```diff
@@ -15,8 +13,11 @@
 from .utilities.helpers import is_pyvista_dataset
 from .utilities.helpers import wrap
 
+if TYPE_CHECKING:
```
user27182, line 16:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

PR #6186, opened 2024-06-03
Use central vtk data in CI tests

MatthewFlamm:
### Overview

Sometimes we get random timeouts retrieving data from pyvista/vtk-data.  This is compounded by the fact that we download data for every test type separately.  This PR checkouts the vtk-data repo once and uses an artifact, and I assume this will be robust since it is a GitHub maintained action.

Not sure if this has an unintended downside due to downloading all the data in vtk-data, but I intend to compare overall execution times.


MatthewFlamm:
Closing as this has unexpected issues on different platforms and also not that much faster.

PR #6192, opened 2024-06-04
Unify `Property` and `ActorProperties` classes

user27182:
### Overview

Follow up from #5029.

- Deprecate the `ActorProperties`  class
- Update the `Property` class to include attributes from `ActorProperties`
- Update `vtkAxesActor` to use `Property`  instead of `ActorProperties`

### Details
The `ActorProperties`  class seems to only exist to wrap the property objects of `AxesActor`. The wrapping is done with a 'has-a' relationship, where a `ActorProperties` object stores a reference to the underlying `vtkProperty` object. This contrasts with the 'is-a' wrapping used by most `PyVista` wrappers, including the `Property` class. This may have been done because the `vtkAxesActor` class does not provide a direct API for wrapping and setting the actor's properties. However, this leads to redundancy and inconsistency, since there are now two APIs for wrapping `vtkProperty` objects.

This PR uses `vtkPropCollection` and the `GetActors` method to enable direct access to the individual shaft and tip actors of a `vtkAxesActor` instance. With references to the actors, we can use wrapped PyVista `Property` objects and set them using the `SetProperty` method of the actors.


tkoyama010:
pre-commit.ci autofix

user27182:
The macOS tests are still expected to fail (axes properties image regression error of 535, 500 is allowed) but everything else should pass. I'll look into the failure after #6204 is merged.

user27182:
@pyvista-bot preview

pyvista/plotting/_property.py
```diff
@@ -257,6 +264,8 @@ def __init__(
             edge_opacity = self._theme.edge_opacity
         self.edge_opacity = edge_opacity
 
+        self.shading = shading
```
user27182, line 267: Should probably add `index_of_refraction` and `anisotropy_rotation` here as well

PR #6204, opened 2024-06-04
Upload artifacts with generated images for failed plotting tests

user27182:
### Overview

Sometimes there are failures with the plotting tests that are hard to debug since the failures may be build-specific. This PR uploads the ~failed~ generated images so failed tests can be more easily debugged.

This uses a similar config to the testing added in #5718


.github/workflows/testing-and-deployment.yml
```diff
@@ -53,7 +53,14 @@ jobs:
...
 
       - name: Test Plotting
-        run: pytest -v tests/plotting --fail_extra_image_cache
+        run: pytest -v tests/plotting --fail_extra_image_cache --generated_image_dir _debug_images
+
+      - name: Upload Images for Failed Tests
+        if: ${{ failure() && hashFiles('_debug_images/*.png') != '' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: debug-images
```
MatthewFlamm, line 62: I think we should have different naming for each job here.  Otherwise, if multiple jobs try to upload, it will throw an error.

user27182: yes good idea. I'll see if there's a way to get the name of the job and insert it

.github/workflows/testing-and-deployment.yml
```diff
@@ -53,7 +53,14 @@ jobs:
         run: pytest -v --ignore=tests/plotting
 
       - name: Test Plotting
-        run: pytest -v tests/plotting --fail_extra_image_cache
+        run: pytest -v tests/plotting --fail_extra_image_cache --generated_image_dir _debug_images
+
+      - name: Upload Images for Failed Tests
+        if: ${{ failure() && hashFiles('_debug_images/*.png') != '' }}
```
MatthewFlamm, line 59: What is the reason for the hashFiles check?  I assume to not upload when no files are present?  Why not just upload an empty directory in this case?  That would be my preference.

user27182: Yes it's for no upload when empty. This was left over from the doc image testing. For that test there is a specific dir created for failed tests only. The plotting tests are different, and all images (failed or not) are generated. So yea we can always upload this

tkoyama010:
pre-commit.ci autofix

.github/workflows/testing-and-deployment.yml
user27182: The artifact names now follow the convention:
`debug_images-PR-number_system-name_python-version_vtk-version`

.github/workflows/testing-and-deployment.yml
user27182: Resolved in https://github.com/pyvista/pyvista/pull/6204/commits/fd251e7f520544ca24b8123df71eb5f498f34245

.github/workflows/testing-and-deployment.yml
MatthewFlamm: What happens if this is executed not from a PR event, e.g. via cron?  Since every action is already unique to a specific event, I don't think it is necessary to have the PR part, and might cause more problems.

user27182: We can remove the 'PR' text and just keep the event number? I imagined working across multiple PRs and suddenly having a bunch of downloads with the same name. Or maybe needing to look for a generated image that you recall downloading a few commits ago and don't want to wade through the CI checks history. The PR number in this case would help find a particular debug image locally I think.

But no other artifacts in other workflows include the PR number, and it's probably best to actually review the full CI logs directly to make sure you are reviewing the correct set of images... So we can revert https://github.com/pyvista/pyvista/pull/6204/commits/9fcc69666c91b5d7c9da2f526f45e1549d38efaa

MatthewFlamm: I'm +1 on everything but the PR#.  I don't think the other events have a `github.event.number` and would be an error.

user27182: We need to add the numpy version too for the 2.0 pre-release testing. This is added in https://github.com/pyvista/pyvista/pull/6204/commits/23f82b6d33c772e14e8f731004188c68c13d56a2 for the linux tests only. 

So the linux test artifact names now follow this convention:
`debug-images_system-name_python-version_vtk-version_numpy-version`

And windows/macos/linux-conda follow this convention:
`debug-images_system-name_python-version_vtk-version`

MatthewFlamm: Looked into this more.  I think we should use {{github.job}}.  This ensures these stay aligned with the job name and we don't have to maintain it 

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

tkoyama010 (dismissed):
LGTM

tkoyama010 (approved):

.github/workflows/testing-and-deployment.yml
user27182: > Looked into this more. I think we should use {{github.job}}.

By itself unfortunately this doesn't give the name we want. By itself this evaluates to `windows` or `linux`, which is not a unique name:
Conflict: an artifact with this name already exists on the workflow run

I'm looking for a way to easily grab the job name as seen in the sidebar of the CI checks, e.g. the full name `Linux Unit Testing (3.12, latest, pre)`

.github/workflows/testing-and-deployment.yml
```diff
@@ -53,7 +53,14 @@ jobs:
...
 
       - name: Test Plotting
-        run: pytest -v tests/plotting --fail_extra_image_cache
+        run: pytest -v tests/plotting --fail_extra_image_cache --generated_image_dir _debug_images
+
+      - name: Upload Generated Images
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ github.job }}
```
tkoyama010, line 62:
```suggestion
          name: ${{ github.job }}-${{ matrix.python-version }}
```

.github/workflows/testing-and-deployment.yml
```diff
@@ -217,7 +231,14 @@ jobs:
...
       - name: Plotting Testing (uses GL)
         continue-on-error: ${{ matrix.numpy-version == 'pre' }}
-        run: xvfb-run -a python -m pytest --cov=pyvista --cov-append --cov-report=xml --fail_extra_image_cache -v --ignore=tests/core --ignore=tests/examples
+        run: xvfb-run -a python -m pytest --cov=pyvista --cov-append --cov-report=xml --fail_extra_image_cache -v --ignore=tests/core --ignore=tests/examples --generated_image_dir _debug_images
+
+      - name: Upload Generated Images
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ github.job }}
```
tkoyama010, line 240:
```suggestion
          name: ${{ github.job }}-${{ matrix.python-version }}-${{ matrix.vtk-version }}-${{ matrix.numpy-version }}
```

.github/workflows/testing-and-deployment.yml
```diff
@@ -296,4 +317,11 @@ jobs:
...
 
       - name: Test Plotting
-        run: python -m pytest -v tests/plotting --fail_extra_image_cache
+        run: python -m pytest -v tests/plotting --fail_extra_image_cache --generated_image_dir _debug_images
+
+      - name: Upload Generated Images
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ github.job }}
```
tkoyama010, line 326:
```suggestion
          name: ${{ github.job }}-${{ matrix.python-version }}
```

tkoyama010 (changes requested):

.github/workflows/testing-and-deployment.yml
user27182: This suggestion is better than what was there before. I'm happy with this (and the other) suggestions.

Ideally there would be a single magic variable that returns the full test name instead of needing to concatenate the relevant vars, e.g. if the numpy pre-release tests are removed in the future, it would be preferred not to also need to remove the refs to `matrix.numpy-version`. Not sure if this is possible though.

MatthewFlamm: > Ideally there would be a single magic variable that returns the full test name instead of needing to concatenate the relevant vars, e.g. if the numpy pre-release tests are removed in the future, it would be preferred not to also need to remove the refs to `matrix.numpy-version`. Not sure if this is possible though.

This is what I was going for in my comments, thanks for understanding my terse statements.  I fear that this method will yield a situation where it becomes a maintenance burden for CI.  I'm also looking into this, otherwise, this type of suggestion looks like a good path forward.

user27182: That's too bad there isn't a simple way to do this. Maybe it's possible to simply use `matrix` generically?
``` diff
+ ${{ github.job }}-${{ matrix }}
- ${{ github.job }}-${{ matrix.python-version }}-${{ matrix.vtk-version }}-${{ matrix.numpy-version }}
```
Or maybe a general for loop to loop over the different matrix options.

user27182: > Maybe it's possible to simply use `matrix` generically?

So it's possible to use `${{ toJSON(matrix) }}` to get the matrix vars as a string. But then this needs to be parsed to create a valid file name.

E.g. the JSON might be:
``` json
{                                                                                   
  "python-version": "3.8",
  "vtk-version": "9.0.3",
  "numpy-version": "default"
}
```

Surprisingly, github actions don't have any string replacement expressions. Instead, we can use a `bash` function to format this. E.g. define a function `format` that takes `JSON` input, formats it as hyphen-separate name-value pairs, and replaces newline characters with underscores:
``` bash
format() { echo $1  | jq -r 'to_entries[] | "\(.key)-\(.value)"' | tr '\n' '_' | sed 's/_$//' }
```

The problem with this is that I haven't found a way to easily define a bash function  with github actions for re-use by different jobs. Either a separate `.sh` script file needs to be written then called, or encapsulate the formatting as its own separate action that is then called by each job. 

Instead of using a function, we can call this directly, E.g.
``` bash
 echo '{
  "python-version": "3.8",
  "vtk-version": "9.0.3",
  "numpy-version": "default"
}' | jq -r 'to_entries[] | "\(.key)-\(.value)"' | tr '\n' '_' | sed 's/_$//'
```
will output:
```
python-version-3.8_vtk-version-9.0.3_numpy-version-default
```

But then the script needs to be copy/pasted. Not sure if having a more complex bash script copied multiple times is better maintenance-wise than referencing the vars explicitly with
``` bash
${{ github.job }}-${{ matrix.python-version }}-${{ matrix.vtk-version }}-${{ matrix.numpy-version }}
```

user27182: Right so, the latest commit https://github.com/pyvista/pyvista/pull/6204/commits/e1a54d81743db004533d8dcf602813bd2c81e94b didn't work, the bash needs to be run as its own step to store the filename as a variable first.

This is definitely becoming more complicated than necessary, I think the original suggestions from @tkoyama010 are probably good enough

.github/workflows/testing-and-deployment.yml
```diff
@@ -59,7 +59,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: ${{ github.job }}-${{ echo toJSON(matrix) | jq -r 'to_entries[] | "\(.key)-\(.value)"' | tr '\n' '_' | sed 's/_$//' }}
+          name: ${{ github.job }}-${{ join(matrix.* , '-/') }}
```
MatthewFlamm, line 62: looks like the slash doesn't work, just use '-'?

user27182: yea I don't know how the '/' got there in the first place

.github/workflows/testing-and-deployment.yml
user27182: > I found some other questions related to this:
> 
> https://stackoverflow.com/questions/71593861/how-to-get-github-actions-matrix-job-name-within-job

Turns out this answer from the linked question had the winning solution for us
https://stackoverflow.com/a/71971218

MatthewFlamm:
Going back to an original comment, there was unrelated segfault in a test and no artifact is uploaded and a nice warning is given for context.  This is a nice result.

> Warning: No files were found with the provided path: debug_images. No artifacts will be uploaded.

user27182:
> Going back to an original comment, there was unrelated segfault in a test and no artifact is uploaded and a nice warning is given for context. This is a nice result.
> 
> > Warning: No files were found with the provided path: debug_images. No artifacts will be uploaded.

Thanks for pointing this out. It is a nice result. Might be good to consider removing the conditionals from the artifact upload from the doc image testing in the docs workflow.

.github/workflows/testing-and-deployment.yml
```diff
@@ -53,7 +53,14 @@ jobs:
...
 
       - name: Test Plotting
-        run: pytest -v tests/plotting --fail_extra_image_cache
+        run: pytest -v tests/plotting --fail_extra_image_cache --generated_image_dir debug_images
+
+      - name: Upload Generated Images
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: debug_images-${{ github.job }}-${{ join(matrix.* , '-') }}
```
tkoyama010, line 62: Awesome expression!

tkoyama010 (approved):

user27182:
Trame tests keep failing due to request timeouts. This is caused by Pooch when downloading mesh files. Turns out the default timeout is hard-corded to 5 seconds. A very recent PR to Pooch was pushed to address this. https://github.com/fatiando/pooch/pull/418

Not sure if it's worth trying to install from the main branch for Pooch temporarily until a patch is issued.

user27182:
> Not sure if it's worth trying to install from the main branch for Pooch temporarily until a patch is issued.

Shouldn't be necessary, looks like a new version is scheduled for release in the next day or two:
https://github.com/fatiando/pooch/issues/419

MatthewFlamm:
> Trame tests keep failing due to request timeouts. This is caused by Pooch when downloading mesh files. Turns out the default timeout is hard-corded to 5 seconds. A very recent PR to Pooch was pushed to address this. [fatiando/pooch#418](https://github.com/fatiando/pooch/pull/418)
> 
> Not sure if it's worth trying to install from the main branch for Pooch temporarily until a patch is issued.

I had also noticed these problems recently. This happens periodically in generally.  My thought was to check out `pyvista/vtk-data` and then reuse it in each test.  I had problems with environment variables in windows and macos.

See https://github.com/pyvista/pyvista/pull/6186

leouieda:
> Shouldn't be necessary, looks like a new version is scheduled for release in the next day or two: https://github.com/fatiando/pooch/issues/419

The patch (v1.8.2) is out on PyPI already and should be on conda-forge soon as well.

user27182:
> The patch (v1.8.2) is out on PyPI already and should be on conda-forge soon as well.

Thank you! 🙏

banesullivan (approved):

PR #6203, opened 2024-06-04
Add `nitpicky` option to doc config to prevent broken links

user27182:
### Overview

This config option should help prevent introducing broken links in the docs.

See: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-nitpicky

Currently a bunch of nitpick ignores are already defined here:
https://github.com/pyvista/pyvista/blob/334d852a0c46a598c036f79230d5bbc6db666170/doc/source/conf.py#L206-L225

But these have no effect unless `nitpicky` is set to `True`.

PR #6206, opened 2024-06-04
Fix internal links to core typing

user27182:
### Overview

The internal core typing links are currently broken. (See the build errors generated when setting `nitpicky=True` in #6203 for a full list of broken links).

The issue seems to stem from aliases not being documented as `py:class:` (aliases are `py:data:`). See https://github.com/sphinx-doc/sphinx/issues/10974#issuecomment-1315385037

The `pyvista.ColorLike` aliases work, however.  And @dcbr pointed out in https://github.com/pyvista/pyvista/pull/5621#discussion_r1532044361 that this could be because `ColorLike` uses `.. autosummary::`.

It would be preferable not to use autosummary, however, because it creates a separate entry (link and page) for the class. But for aliases I think it would be better if they were all shown in a single flat level, glossary-style.

I found this rst stub generated locally in the pyvista repo:
``` rst
pyvista.ColorLike
=================

.. currentmodule:: pyvista

.. autodata:: ColorLike
```

So, it seems like this is what `autosummary` generates.  This formatting was copied and applied to `typing.rst`. Let's see if it works.

### DIFF
--- a/doc/source/api/core/typing.rst
+++ b/doc/source/api/core/typing.rst
@@ -1,6 +1,5 @@
 Typing
 ======
-.. currentmodule:: pyvista
 
 Type aliases and type variable used by PyVista.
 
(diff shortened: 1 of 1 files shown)


user27182:
@pyvista-bot preview

user27182:
It worked! E.g. see https://665fa649d4c47e72ddbd754d--meek-duckanoo-d2f5ec.netlify.app/api/core/_autosummary/pyvista.datasetattributes.set_array#pyvista.DataSetAttributes.set_array

tkoyama010 (approved):
Nice!

dcbr:
Great, glad to see it finally work! Crazy how such subtleties can cause things to break or work.

user27182:
> Crazy how such subtleties can cause things to break or work.

Yes definitely not an intuitive feature to implement! 

Unfortunately even this fix is only a partial fix. The aliases in the docstrings are linked, but not in the function signatures, e.g. see the `ArrayLike` alias used for `data`:
image.

Also, after this PR was merged, the number of broken links did not get reduced in https://github.com/pyvista/pyvista/pull/6203, so, this did not fix the broken links in that sense.

This seems to be a known bug, and may not be fixed any time soon. See https://github.com/sphinx-doc/sphinx/issues/10455#issuecomment-1872516960

There are apparently some hacky workarounds involving html mods, redefinitions, or custom extensions. But this doesn't seem worth the effort. For the `nitpicky` option in #6203, I think any missing ref warnings for type aliases will just need to be explicitly ignored.

PR #6210, opened 2024-06-05
Temporarily restrict NumPy version in pyproject.toml

adeak:
We hope to fix all compatibility issues in https://github.com/pyvista/pyvista/pull/6156, but NumPy 2.0 is expected in 10 days and we should probably make sure that `main` doesn't break when NumPy 2.0 is released. Even though we'll only release when compatibility issues are fixed, excluding a dependency version that's known broken (just a bit) is good practice.

My plan/proposal is:
1. merge this PR quickly to protect `main` from the NumPy 2.0 release until we're compatible
2. remove the version restriction in https://github.com/pyvista/pyvista/pull/6156
3. probably make the NumPy nightly build in our CI matrix mandatory, at least until 2.0 is released.


tkoyama010 (approved):
LGTM. I will merge this since we need to release as soon as possible.

tkoyama010:
I will create a patch release on this.

adeak:
Thanks, @tkoyama010!

PR #6214, opened 2024-06-06
Split up `validation` PR

user27182:
### Overview

Follow-up from #5571. Trying to split that that thing up to make reviews easier. I think it can be roughly split up as:

1. changes  to `_core_typing` (#6216)
2. changes to `check` functions
3. changes to `validate` functions
4. add array wrappers
5. add typing overloads (#6191)

This PR currently only includes changes from items 1 and 2. (Code from #5571 was patched in and all validate/array-wrapper stuff was removed).  I'll create a separate PR from this to only include item 1.


user27182:
pre-commit.ci autofix

PR #6216, opened 2024-06-06
Update `ArrayLike` and `NumberType` definitions in core typing

user27182:
### Overview

Follow up from #5571 generally, and #6214 specifically (this PR is item 1 defined there).

### Details
- Remove `FiniteNestedSequence` (no longer used)
- Redefine `ArrayLike` as a union of 1D, 2D, 3D, and 4D array-like types
- Generalize `NumberType` definition as any python or numpy float, int, or bool type
- Add private `_ArrayLikeOrScalar` alias to make it easier to add typing for inputs where scalars are allowed


tkoyama010 (approved):
LGTM

PR #6220, opened 2024-06-07
Split `test_axes_actor_properties` into two separate tests

user27182:
### Overview

Follow-up from #6192

This test is a bit confusing. It sets shaft properties in two separate locations in the test code. It also sets actor properties which require PBR, but does not enable PBR, so these properties actually do nothing (and therefore aren't actually being tested in terms of the plotting output). This makes it hard to debug.

This PR fixes this.

### Details

The tests are split into two:

1. Testing setting/getting the `ActorProperties` attributes independently of  `AxesActor` (no plot)
2. Testing updating the `ActorProperties` of `AxesActor` (with plot)

For (2), the main test here is checking that the underlying `vtkProperty` objects are modified. Therefore, changing one attribute of `vtkProperty` (e.g. color) is sufficient for this test. It is therefore not necessary to check the plotting output from setting PBR-specific properties. This avoids any possible confusion about whether the property was applied correctly (e.g. no need to guess whether PBR must be enable).


user27182:
I think I finally nailed down the original bug that motivated changing/splitting this test up in the first place. On some macOS builds the opacity doesn't render correctly.

Actual:
image

Expected:
image

We could skip macOS or just not use opacity. May be best to avoid opacity.

user27182:
I'll add that I don't believe the changes from this PR is causing this issue with opacity.

E.g. the generated image for this test (from a separate and unrelated PR, macOS python 3.9) https://github.com/pyvista/pyvista/actions/runs/9436982764/job/25992298774?pr=6230#step:9:28

looks like this:

image

tkoyama010:
@user27182  Please request my review after you resolve conflict.

tkoyama010 (approved):
LGTM

PR #6221, opened 2024-06-07
Fix duplicate Ruff's target-version and project.requires-python

tkoyama010:
### Overview

We have both Ruff's target-version and project.requires-python. We only need the latter.

### Details

- https://learn.scientific-python.org/development/guides/style/#RF002


tkoyama010:
@pyvista-bot LGTM

PR #6225, opened 2024-06-07
Enable `redundant-expr` and `truthy-bool` in MyPy

tkoyama010:
### Overview

- [MY105](https://learn.scientific-python.org/development/guides/style#MY105): MyPy enables redundant-expr
- [MY106](https://learn.scientific-python.org/development/guides/style#MY106): MyPy enables truthy-bool

### Details

- None


user27182:
I think `truthy-bool` can probably be enabled on its own here without requiring any changes to code

PR #6229, opened 2024-06-09
Fix `AxesActor` default colors

user27182:
### Overview

Follow up from #5029.

Initializing `AxesActor` does not correctly apply the axes colors.

``` python
import pyvista as pv
axes = pv.AxesActor()
plot = pv.Plotter()
plot.add_actor(axes)
plot.show()
```
image

This is caused by using `int_rgb` instead of `float_rgb` in ``__init__``. This PR fixes this.


tests/plotting/test_plotting.py
```diff
@@ -3906,6 +3906,16 @@ def test_add_remove_scalar_bar(sphere):
     pl.show()
 
 
+def test_axes_actor_colors():
```
user27182, line 3909:
```suggestion
def test_axes_actor_default_colors():
```

tkoyama010 (approved):
Awesome! This PR is very clear for me :)

PR #6230, opened 2024-06-09
Update `CONTRIBUTING.rst` for input validation package

user27182:
### Overview

Follow-up from #5664. This update should be made once the input validation API stabilizes.


banesullivan:
@user27182 should we add this now?

PR #6231, opened 2024-06-09
Add `labels` property to `AxesActor` for convenience

user27182:
### Overview

Follow-up from #5029.

New property to make it more convenient to set the labels, e.g. 
``` python
axes.labels = 'UVW'
```
instead of
``` python
axes.x_axis_label = 'U'
axes.y_axis_label = 'V'
axes.z_axis_label = 'W'
```


pyvista/plotting/axes_actor.py
```diff
@@ -402,6 +403,43 @@ def tip_type(self, tip_type: Union[TipType, int]):  # numpydoc ignore=GL08
...
+
+        Examples
+        --------
+        Set single-character axes labels.
+
+        >>> import pyvista as pv
+        >>> axes_actor = pv.AxesActor()
+        >>> axes_actor.labels = 'UVW'
+        >>> axes_actor.labels
+        ('U', 'V', 'W')
```
tkoyama010, line 425: The latter expression feels convenient, but this one feels confusing when viewed as a user. A short-sighted user like myself might specify “X AxisY AxisZ Axis” and so on. How about remove this and use only latter expression?

tkoyama010 (changes requested):
Thanks for this.

pyvista/plotting/axes_actor.py
user27182: Do you mean allow `Sequence[str]` but do not allow `str`? I do agree it is perhaps a bit too convenient. We can remove it.

tkoyama010: Yes. Let's remove it.

tkoyama010 (approved):
LGTM

PR #6237, opened 2024-06-10
Add type annotations to `check_real` and update implementation

user27182:
### Overview

Follow up from #5571 / #6214


tkoyama010 (approved):
LGTM

PR #5044, opened 2023-10-10
Add `adjusted_edge_curvature` method

tkoyama010:
### Overview

resolves #4933
image

- [ ] Compare value with [CurvaturesAdjustEdges](https://examples.vtk.org/site/Python/PolyData/CurvaturesAdjustEdges/).

### Details

- None

PR #6253, opened 2024-06-13
Add `compute_line_length` to `PolyData` class

tkoyama010:
### Overview

Follow up in #5044.

### Details

- None

### DIFF
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -2766,6 +2766,38 @@ class PolyDataFilters(DataSetFilters):
         _update_alg(alg, progress_bar, 'Computing the Arc Length')
         return _get_output(alg)
 
+    def compute_points_length(self, point_id_a, point_id_b):
+        """Compute the length between two points given their ids.
+
+        .. versionadded:: 0.44.0
+
+        Parameters
+        ----------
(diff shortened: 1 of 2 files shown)


tkoyama010:
@pyvista-bot preview

tkoyama010:
pre-commit.ci autofix

tkoyama010:
@pyvista-bot preview

MatthewFlamm:
What is the use case here?  A `DataSet` already has `length` which is the length of the bounding box.  We can get the length of each 1D cell using `compute_cell_sizes`. This isn't actually a `line_length` but the distance between two points. This feels more like a user routine if they need to determine the distance between various points, which could be an example.

tkoyama010:
> What is the use case here? A `DataSet` already has `length` which is the length of the bounding box. We can get the length of each 1D cell using `compute_cell_sizes`. This PR isn't actually a `line_length` but the distance between two points. This feels more like a user routine if they need to determine the distance between various points, which could be an example.

Thanks for the feedback. After some cool thinking, I found that this method can be easily written as follows.
```python
np.linalg.norm(self.points(start_vertex) - self.points(end_vertex))
```
I added this method for the calculations in #5044, but this method is redundant clearly. I will close this proposal.

Issue #6252, opened 2024-06-13
number_labels in add_ruler not working as expected; similar for show_bounds


heril96:
### Describe the bug, what's wrong, and what you expected.

I found a bug!

My application required adding multiple ticks and the data set where i am using pyvista is multiblock data set with x axis range from -1.4m to 1.4. I am adding two rulers one in +ve x dirn and another in -ve x dirn,  the issue is  I can only add total 5 labels from 0 to 1.4m even if i want to add 20. Moreover when i use ruler in negative x-dirn which is 0 to -1.4m the ticks are above the x axis and total label number is fixed to 5. Attached image!

Below is different example pasted for which increasing the number of labels is not changing the number of labels (but the numbers do change, please try) but reducing the number_labels reduces the ticks. If i want to produce 15 ticks it is not possible here but produce 3 is possible in below example.

### Steps to reproduce the bug.

```python
import pyvista as pv
data = [
    pv.Sphere(center=(2, 0, 0)),
    pv.Cube(center=(0, 2, 0)),
    pv.Cone(),
]
blocks = pv.MultiBlock(data)

#blocks.center=[2,2,2]
translation_vector = (2, 1, 2)
for i in range(len(blocks)):
    blocks[i].translate(translation_vector, inplace=True)
print(blocks.center) 
plotter = pv.Plotter()
plotter.add_mesh(blocks, smooth_shading=True)
plotter.show_bounds(n_xlabels=10,n_ylabels=20) 
plotter.view_xy()
plotter.show()
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Thu Jun 13 10:28:02 2024 India Standard Time

                  OS : Windows (10 10.0.19045 SP0 Multiprocessor Free)
              CPU(s) : 16
             Machine : AMD64
        Architecture : 64bit
                 RAM : 31.7 GiB
         Environment : IPython
         GPU Details : error
    MathText Support : False

  Python 3.11.9 | packaged by Anaconda, Inc. | (main, Apr 19 2024, 16:40:41)
  [MSC v.1916 64 bit (AMD64)]

             pyvista : 0.43.8
                 vtk : 9.3.0
               numpy : 1.26.4
          matplotlib : 3.8.4
              scooby : 0.10.0
               pooch : 1.8.1
              pillow : 10.3.0
             imageio : 2.33.1
               PyQt5 : 5.15.10
             IPython : 8.20.0
            colorcet : 3.1.0
             cmocean : 4.0.3
[...]

PR #6258, opened 2024-06-14
Fix `add_ruler number_labels` to turn off the auto

MatthewFlamm:
### Overview

This adds in logic to turn off the automatic "niceness" behavior when selecting number of labels.

### Details

- Addresses 1/2 of #6252


MatthewFlamm:
A future enhancement could be to let the user switch between the different "niceness" modes that vtkAxisActor2D exposes.

tkoyama010 (approved):
LGTM

PR #6259, opened 2024-06-14
Add `Actor` to `add_mesh` return type

### LINKED ISSUE #6257: Return type hint in plotting/plotter.py

Hello, I would like to have a return type hint for methods in `plotting/plotter.py`.

For example with `add_mesh()` I would like to have a return type `Actor` so that my IDE (VS code) understand that the variable is an `Actor` and make it easier to navigate into the code.

In short, I would like to go from the first picture to the second:
image
image

I wanted to make a PR but it seems I need to fix some additional ruff rules and mypy errors shown by pre-commit so I am a bit stuck. 
Is there a way to implement the return type hints? Or maybe a different way to solve my problem?

Thanks

_No response_

### Pseudocode or Screenshots

_No response_

MatthewFlamm:
### Overview

closes #6257


pyvista/plotting/plotter.py
```diff
@@ -3728,7 +3728,7 @@ def add_mesh(
         prop = Property(**prop_kwargs)
         actor.SetProperty(prop)
 
-        if style == 'points_gaussian':
+        if style == 'points_gaussian' and self.mapper.dataset is not None:
```
MatthewFlamm, line 3731: `self.mapper.dataset` type is `DataSet | None`.  This is probably the right fix from a typing standpoint.  It is _possible_ that it should be only `DataSet` here and a different fix is needed, but given the complexity of `add_mesh` I'm not confident that it is true.

MatthewFlamm:
@kmarchais

pyvista/plotting/plotter.py
MatthewFlamm: If it is `None`, then we just stop setting the parameters for `point_gaussian` which seems like an okay result to me.

kmarchais:
Thank you ! 

Maybe we can take this opportunity to add more return type hints on this file in this PR?

kmarchais:
On this branch on my fork [kmarchais/pyvista/pyvista/plotting/plotter.py](https://github.com/kmarchais/pyvista/blob/maint/add-mesh-return-type/pyvista/plotting/plotter.py), I have added return types for almost all other functions except for `show` if you want to use it.

tkoyama010:
> On this branch on my fork [kmarchais/pyvista/pyvista/plotting/plotter.py](https://github.com/kmarchais/pyvista/blob/maint/add-mesh-return-type/pyvista/plotting/plotter.py), I have added return types for almost all other functions except for `show` if you want to use it.

@kmarchais 
Sorry to interrupt, but you can also create a Pull Request for `maint/add-mesh-return-type`. That would work better as your contribution would be clearer in my mind.

kmarchais:
Oh right! Thanks, I have just done it

pyvista/plotting/plotter.py
adeak: Yup, I'm not at all aware of all the mechanics here, but the fix seems to err on the safe side.

adeak (approved):
LGTM, thanks.

tkoyama010 (approved):
LGTM

PR #6260, opened 2024-06-14
Fix "none" for legend

akaszynski:
There appears to have been a regression in the treatment of `None` for `face` for `add_legend`:

```py
pl = Plotter()
pl.add_mesh(
    pv.Icosphere(center=(3, 0, 0), radius=1), color="r", label="Sphere", show_edges=True
)
pl.add_mesh(pv.Box(), color="w", label="Box")
pl.add_legend(face=None, bcolor="k")
pl.show()
```
main

Upon investigation, it turns out that the current default, `face=None` does not match the documented default of "triangle" and setting it to "triangle" breaks a variety of our tests because we reuse the vtk object when displaying the legend. Therefore, we have to keep `None` and it's necessary to now specify `"none"` to remove the label.

Corrected this becomes:
```py
pl = Plotter()
pl.add_mesh(
    pv.Icosphere(center=(3, 0, 0), radius=1), color="r", label="Sphere", show_edges=True
)
pl.add_mesh(pv.Box(), color="w", label="Box")
pl.add_legend(face="none", bcolor="k")
pl.show()
```
none-corrected

Notice the single pixel in place of the legend face. This is because our `PolyData` filler is actually a single point (done automatically when inputting just points). We can address this by setting the legend face to have no faces with:

```py
legendface = pyvista.PolyData([0.0, 0.0, 0.0], faces=np.empty(0, dtype=int))
``` 
This finally results in:
final


pyvista/plotting/renderer.py
```diff
@@ -3676,17 +3676,21 @@ def add_legend(
             * ``'upper center'``
             * ``'center'``
 
-        face : str | pyvista.PolyData | NoneType, default: "triangle"
-            Face shape of legend face.  One of the following:
+        face : str | pyvista.PolyData, optional
```
tkoyama010, line 3679:
```suggestion
        face : str | pyvista.PolyData, optional, default: "triangle"
```
This expression is still need.

tkoyama010 (changes requested):

pyvista/plotting/renderer.py
akaszynski: The default isn't triangle though. If glyphs are input (as in our glyph legend test):

```py
def test_legend_using_add_legend_only_labels(random_hills, verify_image_cache):
    pl = pv.Plotter()

    arrows = random_hills.glyph(scale="Normals", orient="Normals", tolerance=0.05)
    pl.add_mesh(arrows, color="black", label="label 1")

    pl.add_mesh(random_hills, scalars="Elevation", cmap="terrain", show_scalar_bar=False)

    legend_entries = ["label 1", "label 2"]

    pl.add_legend(legend_entries, size=(0.5, 0.5))
    pl.show()
```
image


The default isn't triangle (nor is the default text).

tkoyama010: Thanks for explanation. Got it :)

tkoyama010 (approved):
LGTM

tkoyama010 (changes requested):
I will add a test for this. Please wait for the merge.

akaszynski:
> I will add a test for this. Please wait for the merge.

Added a test in https://github.com/pyvista/pyvista/pull/6260/commits/53adccecc512469e676ce6c5dc7001f4fe104ea9.

tkoyama010 (approved):
LGTM. Thanks!

MatthewFlamm (commented):
Looks like I never submitted this review. Sorry about that. My comments were not critical either way.

pyvista/plotting/renderer.py
```diff
@@ -135,7 +135,7 @@ def make_legend_face(face):
             '\t"triangle"\n'
             '\t"circle"\n'
             '\t"rectangle"\n'
-            '\tNone'
+            '\t"none"\n'
```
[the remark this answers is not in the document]
akaszynski: Good point. I'll resolve that (and the other comment).

Issue #5665, opened 2024-02-22
Add `autodoc-typehints` to docs.


user27182:
### Describe what maintenance you would like added.

Currently, there is redundancy when using annotations since there is a need to add the argument types to the docstring as well.

Using [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) should make it so that type hints from annotations can be used directly without the need to specify both.

This issue was originally pointed out by @darikg here https://github.com/pyvista/pyvista/pull/5571#discussion_r1491545746.

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

PR #6262, opened 2024-06-15
Following PR#6259 add return type for other functions in plotter.py

kmarchais:
This PR follows #6259 and brings more return types in the `pyvista/plotting/plotter.py` file.

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -1409,7 +1409,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         return self.renderer.parallel_scale
 
     @parallel_scale.setter
-    def parallel_scale(self, value) -> None:  # numpydoc ignore=GL08
+    def parallel_scale(self, value: float) -> None:  # numpydoc ignore=GL08
         self.renderer.parallel_scale = value
 
     @wraps(Renderer.add_axes_at_origin)
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -1618,7 +1618,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         return self.renderer.camera.is_set
 
     @camera_set.setter
-    def camera_set(self, is_set) -> None:  # numpydoc ignore=GL08
+    def camera_set(self, is_set: bool) -> None:  # numpydoc ignore=GL08
         self.renderer.camera.is_set = is_set
 
     @property
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -1050,7 +1050,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         return self.renderer.remove_legend(*args, **kwargs)
 
     @property
-    def legend(self) -> _vtk.vtkLegendBoxActor:  # numpydoc ignore=RT01
+    def legend(self) -> _vtk.vtkLegendBoxActor | None:  # numpydoc ignore=RT01
         """Legend actor.
 
         There can only be one legend actor per renderer.  If
@@ -1415,7 +1415,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         self.renderer.parallel_scale = value
 
     @wraps(Renderer.add_axes_at_origin)
-    def add_axes_at_origin(self, *args, **kwargs) -> AxesActor:  # numpydoc ignore=PR01,RT01
+    def add_axes_at_origin(self, *args, **kwargs) -> _vtk.vtkAxesActor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_axes_at_origin``."""
         return self.renderer.add_axes_at_origin(*args, **kwargs)
 
@@ -1541,7 +1541,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         self.renderer.enable(*args, **kwarg)
 
     @wraps(Renderer.enable_depth_peeling)
-    def enable_depth_peeling(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT01
+    def enable_depth_peeling(self, *args, **kwargs) -> bool | None:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.enable_depth_peeling``."""
         if self.render_window is not None:
             result = self.renderer.enable_depth_peeling(*args, **kwargs)
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -90,6 +90,8 @@ from .widgets import WidgetHelper
 
 if TYPE_CHECKING:  # pragma: no cover
     from pyvista.core._typing_core import BoundsLike
+    from pyvista.plotting.axes_actor import AxesActor
+    from pyvista.plotting.cube_axes_actor import CubeAxesActor
 
 SUPPORTED_FORMATS = [".png", ".jpeg", ".jpg", ".bmp", ".tif", ".tiff"]
 
@@ -944,7 +946,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         return self._scalar_bars
 
     @property
-    def _before_close_callback(self) -> None:
+    def _before_close_callback(self):
         """Return the cached function (expecting a reference)."""
         if self.__before_close_callback is not None:
             return self.__before_close_callback()
@@ -1026,12 +1028,12 @@ class BasePlotter(PickingHelper, WidgetHelper):
         self.renderers.set_active_renderer(index_row, index_column)
 
     @wraps(Renderer.add_ruler)
-    def add_ruler(self, *args, **kwargs) -> _vtk.vtkActor:  # numpydoc ignore=PR01,RT01
+    def add_ruler(self, *args, **kwargs) -> Actor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_ruler``."""
         return self.renderer.add_ruler(*args, **kwargs)
 
     @wraps(Renderer.add_legend_scale)
-    def add_legend_scale(self, *args, **kwargs) -> _vtk.vtkActor:  # numpydoc ignore=PR01,RT01
+    def add_legend_scale(self, *args, **kwargs) -> Actor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_legend_scale``."""
         return self.renderer.add_legend_scale(*args, **kwargs)
 
@@ -1061,7 +1063,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         return self.renderer.legend
 
     @wraps(Renderer.add_floor)
-    def add_floor(self, *args, **kwargs) -> _vtk.vtkActor:  # numpydoc ignore=PR01,RT01
+    def add_floor(self, *args, **kwargs) -> Actor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_floor``."""
         return self.renderer.add_floor(*args, **kwargs)
 
@@ -1311,12 +1313,12 @@ class BasePlotter(PickingHelper, WidgetHelper):
         return self.renderer.add_orientation_widget(*args, **kwargs)
 
     @wraps(Renderer.add_axes)
-    def add_axes(self, *args, **kwargs) -> _vtk.vtkAxesActor:  # numpydoc ignore=PR01,RT01
+    def add_axes(self, *args, **kwargs) -> AxesActor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_axes``."""
         return self.renderer.add_axes(*args, **kwargs)
 
     @wraps(Renderer.add_box_axes)
-    def add_box_axes(self, *args, **kwargs) -> _vtk.vtkAxesActor:  # numpydoc ignore=PR01,RT01
+    def add_box_axes(self, *args, **kwargs) -> AxesActor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_box_axes``."""
         return self.renderer.add_box_axes(*args, **kwargs)
 
@@ -1411,17 +1413,17 @@ class BasePlotter(PickingHelper, WidgetHelper):
         self.renderer.parallel_scale = value
 
     @wraps(Renderer.add_axes_at_origin)
-    def add_axes_at_origin(self, *args, **kwargs) -> _vtk.vtkAxesActor:  # numpydoc ignore=PR01,RT01
+    def add_axes_at_origin(self, *args, **kwargs) -> AxesActor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_axes_at_origin``."""
         return self.renderer.add_axes_at_origin(*args, **kwargs)
 
     @wraps(Renderer.show_bounds)
-    def show_bounds(self, *args, **kwargs) -> _vtk.vtkCubeAxesActor:  # numpydoc ignore=PR01,RT01
+    def show_bounds(self, *args, **kwargs) -> CubeAxesActor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.show_bounds``."""
         return self.renderer.show_bounds(*args, **kwargs)
 
     @wraps(Renderer.add_bounding_box)
-    def add_bounding_box(self, *args, **kwargs) -> _vtk.vtkActor:  # numpydoc ignore=PR01,RT01
+    def add_bounding_box(self, *args, **kwargs) -> Actor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_bounding_box``."""
         return self.renderer.add_bounding_box(*args, **kwargs)
 
@@ -1436,7 +1438,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         return self.renderer.remove_bounds_axes(*args, **kwargs)
 
     @wraps(Renderer.show_grid)
-    def show_grid(self, *args, **kwargs) -> _vtk.vtkCubeAxesActor:  # numpydoc ignore=PR01,RT01
+    def show_grid(self, *args, **kwargs) -> CubeAxesActor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.show_grid``."""
         return self.renderer.show_grid(*args, **kwargs)
 
@@ -1579,7 +1581,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
     #### Properties from Renderer ####
 
     @property
-    def actors(self) -> dict[str, _vtk.vtkActor]:  # numpydoc ignore=RT01
+    def actors(self) -> dict[str, Actor]:  # numpydoc ignore=RT01
         """Return the actors of the active renderer.
 
         Returns
@@ -2102,10 +2104,10 @@ class BasePlotter(PickingHelper, WidgetHelper):
     @pickable_actors.setter
     def pickable_actors(self, actors=None) -> None:  # numpydoc ignore=GL08
         actors = [] if actors is None else actors
-        if isinstance(actors, _vtk.vtkActor):
+        if isinstance(actors, Actor):
             actors = [actors]
 
-        if not all(isinstance(actor, _vtk.vtkActor) for actor in actors):
+        if not all(isinstance(actor, Actor) for actor in actors):
             raise TypeError(
                 f'Expected a vtkActor instance or a list of vtkActors, got '
                 f'{[type(actor) for actor in actors]} instead.',
@@ -5298,9 +5300,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
 
         return zval
 
-    def add_lines(
-        self, lines, color='w', width=5, label=None, name=None, connected=False
-    ) -> _vtk.vtkActor:
+    def add_lines(self, lines, color='w', width=5, label=None, name=None, connected=False) -> Actor:
         """Add lines to the plotting object.
 
         Parameters
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -1028,12 +1028,14 @@ class BasePlotter(PickingHelper, WidgetHelper):
         self.renderers.set_active_renderer(index_row, index_column)
 
     @wraps(Renderer.add_ruler)
-    def add_ruler(self, *args, **kwargs) -> Actor:  # numpydoc ignore=PR01,RT01
+    def add_ruler(self, *args, **kwargs) -> _vtk.vtkAxisActor2D:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_ruler``."""
         return self.renderer.add_ruler(*args, **kwargs)
 
     @wraps(Renderer.add_legend_scale)
-    def add_legend_scale(self, *args, **kwargs) -> Actor:  # numpydoc ignore=PR01,RT01
+    def add_legend_scale(
+        self, *args, **kwargs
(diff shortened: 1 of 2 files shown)


MatthewFlamm:
I recommend that if this is complex, which it seems it is, we do it in smaller chunks.  Otherwise in my experience, it is hard to track down whether a specific mypy error is directly caused by the type you added or will be solved by fixing a different error.

kmarchais:
I have removed all the return types where were mypy errors. It is still a lot of changes in the same PR though, what do you think?

pyvista/plotting/plotter.py
```diff
@@ -104,7 +105,7 @@
         KILL_DISPLAY = False
 
 
-def close_all():
+def close_all() -> bool:
```
MatthewFlamm, line 108: Weirdly, this function ONLY returns True.  But the way the documentation is written makes it seem like the intent is that it might return False too.  Let's keep it with the type hint you have to avoid any surprises later.

pyvista/plotting/plotter.py
```diff
@@ -366,11 +367,11 @@ def suppress_rendering(self):  # numpydoc ignore=RT01
         return self._suppress_rendering
 
     @suppress_rendering.setter
-    def suppress_rendering(self, value):  # numpydoc ignore=GL08
+    def suppress_rendering(self, value) -> None:  # numpydoc ignore=GL08
         self._suppress_rendering = bool(value)
 
     @property
-    def render_window(self):  # numpydoc ignore=RT01
+    def render_window(self) -> _vtk.vtkRenderWindow | None:  # numpydoc ignore=RT01
```
MatthewFlamm, line 374: The `|` Union is a Python 3.10 feature.  Will we need to import from `__future__`?

MatthewFlamm: Otherwise anyone doing type checking in 3.9, it might not work?  This isn't my area of expertise, just questions from me.

pyvista/plotting/plotter.py
```diff
@@ -943,22 +944,22 @@ def scalar_bars(self):  # numpydoc ignore=RT01
         return self._scalar_bars
 
     @property
-    def _before_close_callback(self):
+    def _before_close_callback(self) -> None:
```
MatthewFlamm, line 947: This one might return something based on the return type of `__before_close_callback`?  I recommend removing this one unless you want a more complex type investigation.

pyvista/plotting/plotter.py
```diff
@@ -1358,181 +1363,181 @@ def add_actor(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
...
 
     @property
-    def parallel_projection(self):  # numpydoc ignore=RT01
+    def parallel_projection(self) -> bool:  # numpydoc ignore=RT01
         """Return or set parallel projection state of active render window."""
         return self.renderer.parallel_projection
 
     @parallel_projection.setter
-    def parallel_projection(self, state):  # numpydoc ignore=GL08
+    def parallel_projection(self, state) -> None:  # numpydoc ignore=GL08
```
MatthewFlamm, line 1401:
```suggestion
    def parallel_projection(self, state: bool) -> None:  # numpydoc ignore=GL08
```

pyvista/plotting/plotter.py
```diff
@@ -1358,181 +1363,181 @@ def add_actor(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
...
 
     @property
-    def parallel_scale(self):  # numpydoc ignore=RT01
+    def parallel_scale(self) -> float:  # numpydoc ignore=RT01
         """Return or set parallel scale of active render window."""
         return self.renderer.parallel_scale
 
     @parallel_scale.setter
-    def parallel_scale(self, value):  # numpydoc ignore=GL08
+    def parallel_scale(self, value) -> None:  # numpydoc ignore=GL08
```
MatthewFlamm, line 1410:
```suggestion
    def parallel_scale(self, value: float) -> None:  # numpydoc ignore=GL08
```

pyvista/plotting/plotter.py
```diff
@@ -1602,16 +1607,16 @@ def camera(self):  # numpydoc ignore=RT01
...
 
     @property
-    def camera_set(self):  # numpydoc ignore=RT01
+    def camera_set(self) -> bool:  # numpydoc ignore=RT01
         """Return or set if the camera of the active renderer has been set."""
         return self.renderer.camera.is_set
 
     @camera_set.setter
-    def camera_set(self, is_set):  # numpydoc ignore=GL08
+    def camera_set(self, is_set) -> None:  # numpydoc ignore=GL08
```
MatthewFlamm, line 1619:
```suggestion
    def camera_set(self, is_set: bool) -> None:  # numpydoc ignore=GL08
```

pyvista/plotting/plotter.py
```diff
@@ -1025,27 +1026,27 @@ def subplot(self, index_row, index_column=None):
         self.renderers.set_active_renderer(index_row, index_column)
 
     @wraps(Renderer.add_ruler)
-    def add_ruler(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def add_ruler(self, *args, **kwargs) -> _vtk.vtkActor:  # numpydoc ignore=PR01,RT01
```
MatthewFlamm, line 1029: Should be `Actor`?

pyvista/plotting/plotter.py
```diff
@@ -1025,27 +1026,27 @@ def subplot(self, index_row, index_column=None):
...
         """Wrap ``Renderer.add_ruler``."""
         return self.renderer.add_ruler(*args, **kwargs)
 
     @wraps(Renderer.add_legend_scale)
-    def add_legend_scale(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def add_legend_scale(self, *args, **kwargs) -> _vtk.vtkActor:  # numpydoc ignore=PR01,RT01
```
MatthewFlamm, line 1034: Should be `Actor`?

pyvista/plotting/plotter.py
```diff
@@ -1060,16 +1061,16 @@ def legend(self):  # numpydoc ignore=RT01
         return self.renderer.legend
 
     @wraps(Renderer.add_floor)
-    def add_floor(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def add_floor(self, *args, **kwargs) -> _vtk.vtkActor:  # numpydoc ignore=PR01,RT01
```
MatthewFlamm, line 1064: Should be `Actor`?

pyvista/plotting/plotter.py
```diff
@@ -1358,181 +1363,181 @@ def add_actor(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
...
 
     @wraps(Renderer.show_bounds)
-    def show_bounds(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def show_bounds(self, *args, **kwargs) -> _vtk.vtkCubeAxesActor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.show_bounds``."""
         return self.renderer.show_bounds(*args, **kwargs)
 
     @wraps(Renderer.add_bounding_box)
-    def add_bounding_box(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def add_bounding_box(self, *args, **kwargs) -> _vtk.vtkActor:  # numpydoc ignore=PR01,RT01
```
MatthewFlamm, line 1424: Should be `Actor`?

pyvista/plotting/plotter.py
```diff
@@ -1555,26 +1560,26 @@ def get_default_cam_pos(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
...
-    def remove_environment_texture(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def remove_environment_texture(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.remove_environment_texture``."""
         return self.renderer.remove_environment_texture(*args, **kwargs)
 
     #### Properties from Renderer ####
 
     @property
-    def actors(self):  # numpydoc ignore=RT01
+    def actors(self) -> dict[str, _vtk.vtkActor]:  # numpydoc ignore=RT01
```
MatthewFlamm, line 1582: should be `Actor`?

pyvista/plotting/plotter.py
```diff
@@ -5289,7 +5298,9 @@ def get_image_depth(self, fill_value=np.nan, reset_camera_clipping_range=True):
 
         return zval
 
-    def add_lines(self, lines, color='w', width=5, label=None, name=None, connected=False):
+    def add_lines(
+        self, lines, color='w', width=5, label=None, name=None, connected=False
+    ) -> _vtk.vtkActor:
```
MatthewFlamm, line 5303: Should be `Actor`?

MatthewFlamm (commented):
Haven't gone through everything, but some comments.  I would also look to ensure whether we should be returning a pyvista class rather than a vtk class in some cases (`Actor` vs. `vtkActor` for example).  It would pass mypy, but it will be best to use the pyvista class if it exists.

pyvista/plotting/plotter.py
kmarchais: `from __future__ import annotations` is already at the top of the file.

pyvista/plotting/plotter.py
kmarchais: Indeed, I removed it.

pyvista/plotting/plotter.py
```diff
@@ -6216,7 +6231,7 @@ def orbit():
                 else:
                     self.render()
                 sleep_time = step - (time.time() - tstart)
-                if sleep_time > 0 and not self.off_screen:
+                if sleep_time > 0 and (hasattr(self, "off_screen") and not self.off_screen):
```
MatthewFlamm, line 6234: Given this needed change, this method should probably be in `Plotter`.  But let's not do that for a typing fix.

pyvista/plotting/plotter.py
```diff
@@ -6026,7 +6039,7 @@ def screenshot(
                 # Plotter hasn't been rendered or was improperly closed
                 raise RuntimeError('This plotter is closed and unable to save a screenshot.')
 
-            if self._first_time and not self.off_screen:
+            if self._first_time and (hasattr(self, "off_screen") and not self.off_screen):
```
MatthewFlamm, line 6042: Similar comment

pyvista/plotting/plotter.py
```diff
@@ -2095,12 +2102,12 @@ def pickable_actors(self):  # numpydoc ignore=RT01
         ]
 
     @pickable_actors.setter
-    def pickable_actors(self, actors=None):  # numpydoc ignore=GL08
+    def pickable_actors(self, actors=None) -> None:  # numpydoc ignore=GL08
         actors = [] if actors is None else actors
-        if isinstance(actors, _vtk.vtkActor):
+        if isinstance(actors, Actor):
```
MatthewFlamm, line 2107: Hmmmm, maybe the suggestions I made were not ideal.  I think I see that we allow users to pass in a `_vtk.vtkActor` directly, so the typing was correct before.  Sorry about that.  We shouldn't make this change.

pyvista/plotting/plotter.py
```diff
@@ -2033,22 +2040,22 @@ def track_mouse_position(self):
...
 
     @wraps(RenderWindowInteractor.untrack_click_position)
-    def untrack_click_position(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def untrack_click_position(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT01
         """Stop tracking the click position."""
         self.iren.untrack_click_position(*args, **kwargs)
 
     @property
-    def pickable_actors(self):  # numpydoc ignore=RT01
+    def pickable_actors(self) -> list[Actor]:  # numpydoc ignore=RT01
```
MatthewFlamm, line 2058: revert back to `_vtk.vtkActor`?

MatthewFlamm (commented):
I might have made this more complicated.  It seems the Plotter methods allow a _vtk.vtkActor to be directly supplied in some cases.  I think the decisions should be:
- Does pyvista create the actor? If yes, and pyvista returns a `pv.Actor`, we should return a `pyvista.Actor` type.  For example, `add_floor` creates the actor completely internally, so we know this to true.
- If it interrogates something that pyvista does not have full control over, it might be best to leave as the vtk type.  For example `pickable_actors`.

pyvista/plotting/plotter.py
```diff
@@ -1783,7 +1790,7 @@ def window_size(self, window_size):  # numpydoc ignore=GL08
         self.render()
 
     @contextmanager
-    def window_size_context(self, window_size=None):
+    def window_size_context(self, window_size=None) -> Generator[BasePlotter, None, None]:
```
user27182, line 1793:
```suggestion
    def window_size_context(self, window_size=None) -> Iterator[BasePlotter]:
```

pyvista/plotting/plotter.py
```diff
@@ -20,6 +20,7 @@
 from threading import Thread
 import time
 from typing import TYPE_CHECKING
+from typing import Generator
```
user27182, line 23:
```suggestion
from typing import Iterator
```

pyvista/plotting/plotter.py
```diff
@@ -1924,14 +1931,14 @@ def image_scale(self) -> int:  # numpydoc ignore=RT01
...
-    def image_scale(self, value: int):  # numpydoc ignore=GL08
+    def image_scale(self, value: int) -> None:  # numpydoc ignore=GL08
         value = int(value)
         if value < 1:
             raise ValueError('Scale factor must be a positive integer.')
         self._image_scale = value
 
     @contextmanager
-    def image_scale_context(self, scale: int | None = None):
+    def image_scale_context(self, scale: int | None = None) -> Generator[BasePlotter, None, None]:
```
user27182, line 1941:
```suggestion
    def image_scale_context(self, scale: int | None = None) -> Iterator[BasePlotter]:
```

user27182 (commented):
Thanks for these great updates! The annotations help a lot.

Small suggestion: for generators that only yield values, using `Iterator` here is a bit cleaner I think. See https://docs.python.org/3/library/typing.html#typing.Generator

MatthewFlamm:
I want to make sure that credit is apportioned fairly with these PRs.  If this is merged into the branch from #6259, it might seem on first glance as if I were the main author of the changes from this PR once #6259 is merged into main.  For example, anyone using `blame` in GitHub would see #6259 as the PR that committed the code.

I also hesitate to suggest it as changing the parent branch on GitHub can create a very messy situation in git history and potential loss of comments here.

I would leave it up to @kmarchais to make the call.  I do not have any hesitation on the current plan, but want to open the door for you to decide another route if you prefer.

pyvista/plotting/plotter.py
kmarchais: To make sure I understood correctly, are we keeping this change here?

MatthewFlamm: Given that this functionality only sleeps when `not self.off_screen`, it seems reasonable to me that we would just ignore this when it doesn't have this attribute.

Although the code organization smells bad here, but we shouldn't fix that in a typing PR.

I would vote to keep your change, maybe add a comment if you feel it is needed.

kmarchais:
I don't have much experience with this. I think the most important is to make the history as clear as possible. Also, this PR is mentioned in #6259.

Thank you for initiating the changes and thanks for the help!

pyvista/plotting/plotter.py
```diff
@@ -1555,26 +1564,26 @@ def get_default_cam_pos(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
...
-    def remove_environment_texture(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def remove_environment_texture(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.remove_environment_texture``."""
         return self.renderer.remove_environment_texture(*args, **kwargs)
 
     #### Properties from Renderer ####
 
     @property
-    def actors(self):  # numpydoc ignore=RT01
+    def actors(self) -> dict[str, Actor]:  # numpydoc ignore=RT01
```
MatthewFlamm, line 1586: Went through one more time.  This is the only one that jumped out at me as questionable.  Should it be `vtkActor`?  Good work on trying to separate the two usages, despite my conflicting comments earlier.

kmarchais: Indeed, thanks

pyvista/plotting/plotter.py
```diff
@@ -1281,69 +1284,73 @@ def disable_anti_aliasing(self, all_renderers=True):
...
 
     @wraps(Renderer.add_axes)
-    def add_axes(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def add_axes(self, *args, **kwargs) -> AxesActor:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_axes``."""
         return self.renderer.add_axes(*args, **kwargs)
 
     @wraps(Renderer.add_box_axes)
-    def add_box_axes(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def add_box_axes(self, *args, **kwargs) -> AxesActor:  # numpydoc ignore=PR01,RT01
```
user27182, line 1323: This can't be right, `AxesActor` is the regular axes with arrows, not the box.  I this return type may be coming from the docstring for this method, which also says `AxesActor` is returned (and which also isn't right). 

`vtk.vtkAnnotatedCubeActor` should probably be the return type here.

I'm surprised `mypy` is even allowing this...

pyvista/plotting/plotter.py
```diff
@@ -1281,69 +1284,73 @@ def disable_anti_aliasing(self, all_renderers=True):
...
-    def add_orientation_widget(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def add_orientation_widget(
+        self, *args, **kwargs
+    ) -> _vtk.vtkOrientationMarkerWidget:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.add_orientation_widget``."""
         return self.renderer.add_orientation_widget(*args, **kwargs)
 
     @wraps(Renderer.add_axes)
-    def add_axes(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def add_axes(self, *args, **kwargs) -> AxesActor:  # numpydoc ignore=PR01,RT01
```
user27182, line 1318: I also don't think is quite right, the only place `pv.AxesActor` is returned in the code base is with `pv.Axes.axes_actor`. I believe `vtk.vtkAxesActor` should be the return type here.

pyvista/plotting/plotter.py
```diff
@@ -1358,181 +1365,181 @@ def add_actor(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
...
 
     @wraps(Renderer.remove_bounds_axes)
-    def remove_bounds_axes(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def remove_bounds_axes(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT01
         """Wrap ``Renderer.remove_bounds_axes``."""
         return self.renderer.remove_bounds_axes(*args, **kwargs)
 
     @wraps(Renderer.show_grid)
-    def show_grid(self, *args, **kwargs):  # numpydoc ignore=PR01,RT01
+    def show_grid(self, *args, **kwargs) -> CubeAxesActor:  # numpydoc ignore=PR01,RT01
```
user27182, line 1443: This return type differs from the docstring, which says `vtk.vtkAxesActor` is returned (which is incorrect). I've noticed a few other docstrings which have the incorrect types too. Unfortunately until #5665 is resolved I think all docstrings will probably need to be double-checked to make sure they're in sync.

pyvista/plotting/plotter.py
kmarchais: Taking a look at Renderer's `add_axes`, `self.axes_actor` is either created using `create_axes_orientation_box` (which seems to return either `_vtk.vtkPropAssembly` or `_vtk.vtkAnnotatedCubeActor`) or using `create_axes_marker` (which returns `_vtk.vtkAxesActor`).

So I put `_vtk.vtkAxesActor | _vtk.vtkPropAssembly | _vtk.vtkAnnotatedCubeActor`

pyvista/plotting/plotter.py
kmarchais: I have just fixed some docstring return types.

pyvista/plotting/plotter.py
kmarchais: Fixed, thanks

MatthewFlamm: > I'm surprised mypy is even allowing this..

I think the wrapped renderer method also needs to be typed for mypy to check the internal code here. Otherwise it will be typed as `Any` and always pass.

MatthewFlamm:
I looked into some of this typing weirdness.  I believe that our mypy testing is flawed, so we need to be careful on the types we add.  Using [mirrors-mypy](https://github.com/pre-commit/mirrors-mypy), means that:

> Because pre-commit runs mypy from an isolated virtualenv (without your dependencies) you may also find it useful to add the typed dependencies to additional_dependencies so mypy can better perform dynamic analysis:

Our pre-commit config only adds numpy to the dependencies (ignoring the mypy specific dependencies):

https://github.com/pyvista/pyvista/blob/11d39721f364cc06c689b769f10158a83fde70bd/.pre-commit-config.yaml#L64-L70

We are missing importantly vtk, but also matplotlib, pillow, scooby...  vtk now ships with type hints in the wheels, so we are missing that in our typing.  Enabling it results in a huge amount of errors.

One consequence of this is that running `pre-commit run mypy --all` is different than `mypy pyvista [all the same args from pre-commit config]`.

pyvista/plotting/plotter.py
adeak: And potential simplification to
```suggestion
                if sleep_time > 0 and not getattr(self, "off_screen", True):
```
but the counter-intuitiveness of the `True` default paired with `None` might mean we're better off with the original plus a comment.

> Although the code organization smells bad here, but we shouldn't fix that in a typing PR.

Wholeheartedly agree.

kmarchais: I have just added the following comment `# 'off_screen' attribute is specific to Plotter objects.`

pyvista/plotting/plotter.py
kmarchais: I have just added the following comment `# 'off_screen' attribute is specific to Plotter objects.`

kmarchais:
Sorry this review request was not intended, I don't understand why all these commits and the review request have been added.

adeak:
> Sorry this review request was not intended, I don't understand why all these commits and the review request have been added.

Looks like you merged the main branch of the wrong upstream repo into your feature branch, or something similar. The review request is automatic because the bogus diff (including all the commits merged from main) touches CI code whose code owner is Bane.

adeak:
If you don't have added changes other than the merge then you can probably reset your local feature branch to the commit before the merge, and force push it. That will undo the push of the merge commit. I always recommend setting up temporary local branches before messing with force pushes and rewriting history. In that case it's easy to undo any fix you might have to fix later :)

kmarchais:
Thank you @adeak, it looks better now. I'm not sure to have managed well the temporary branch though.

adeak:
> I'm not sure to have managed well the temporary branch though.

That was just a safety tip to cover yourself if anything goes awry with the forve push. If this branch looks the way it should now then you're all set :)

kmarchais:
> I could remove the bogus review request, everything back to normal here.

Thank you for that and thanks a lot for the help, I will not change anything else!

tkoyama010:
Thanks @kmarchais

Issue #796, opened 2020-06-13
Support for vtkAssembly


tinebryder:
First of all, thank you for the awesome work! 

**Describe the bug, what's wrong, and what you expect:**
I have some pre-existing code generating a rather complicated vtkAssemby and would like to show it using the pyvista `Plotter`, but there doesn’t seem to be a method for this (or at least, I haven’t been able to find it). 

However, if I use `add_actor` and modify the source code slightly it seems to be working. The only alternation needed is replacing the return statement of the `add_actor` method of `Renderer` with something like
```
        if hasattr(actor, 'GetProperty'):
            prop = actor.GetProperty()
        else:
            prop = None

        return actor, prop
```

-----

**To Reproduce**

The following toy example fails: 

```py
import vtk
import pyvista as pv

# Create assembly
actor = vtk.vtkActor()
sphere = vtk.vtkSphereSource()
sphere.Update()
mapper = vtk.vtkPolyDataMapper()
mapper.SetInputData(sphere.GetOutput())
actor.SetMapper(mapper)

vtkAssembly = vtk.vtkAssembly()
vtkAssembly.AddPart(actor)

# Add assembly to Plotter
plotter = pv.Plotter()
plotter.add_actor(vtkAssembly)
plotter.show()
```

-----

**System Information:**
 Please run the following code wherever you are experiencing the bug and paste the output below. This report helps us track down bugs and it is critical to addressing your bug:

```py
# Get system info
import pyvista as pv
print(pv.Report())
```

```txt
--------------------------------------------------------------------------------
  Date: Sat Jun 13 11:52:38 2020 Romance Daylight Time
                OS : Windows
            CPU(s) : 8
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.8 GB
       Environment : Python
        GPU Vendor : Intel
      GPU Renderer : Intel(R) UHD Graphics 620
       GPU Version : 4.5.0 - Build 26.20.100.7756
  Python 3.6.10 (default, Mar  5 2020, 10:17:47) [MSC v.1900 64 bit (AMD64)]
           pyvista : 0.25.1
               vtk : 8.2.0
             numpy : 1.18.4
           imageio : 2.8.0
           appdirs : 1.4.3
            scooby : 0.5.4
            meshio : 4.0.13
        matplotlib : 3.0.3
         pyvistaqt : 0.1.0
             PyQt5 : 5.12.3
             scipy : 1.3.1
[...]

PR #6263, opened 2024-06-15
Add `AxesAssembly` class to position and orient axes in a scene

user27182:
### Overview

Related to #5029. However, unlike that PR, which uses the existing `AxesActor` (with some ugly hacks) to create orientable axes, this PR instead adds a totally new class that is independent of `AxesActor`. This PR also does not do any "unifying". See recording for a preview of this feature.

https://github.com/pyvista/pyvista/assets/89109579/0a6e6d73-2cc9-4c04-a5f2-41bb4a311707

### Details
- Add `AxesGeometry` class. This outputs a colored mesh (`pyvista.DataSet`) with axes shafts and tips only (no labels)
- Add `AxesAssembly` class. This extends `AxesGeometry` to make the separate shaft and tip datasets into individual actors, and collects them into a `vtkPropAssembly`. It also add label actors.
- Add `TextFollower` class. This is a subclass of `Actor` which uses `Text3DSource` to add 3D (non-raster) text labels to a scene which follow the camera.

See also #796 for a discussion on supporting `vtkAssembly` and/or `vtkPropAssembly`


user27182:
The use of text labels is by far the most challenging aspect of this feature. The use of `vtkFollower` with `vtkVectorText` was initially considered but the text for `vtkVectorText` is not easily configured (it has no TextProperty attribute), and `vtkFollower` has complications, e.g. requires adding a camera. In contrast, it's possible to use `vtkTextActor` instead, which is apparently made to replace the follower/vector text classes, e.g. see the title of this discussion.
https://public.kitware.com/pipermail/vtkusers/2004-July/025514.html

So, a new `TextLabel` class is added instead of a `TextFollower` class

user27182:
This pretty much all merged via other PRs. The only 'feature' not yet merged is the deprecation of "add_axes_at_origin" and replacement with "add_axes_marker" (which can be positioned anywhere, not just at the origin)

Issue #5611, opened 2024-02-13
pre-commit `mypy` config is not catching errors reliably


user27182:
### Describe the bug, what's wrong, and what you expected.

Random `mypy` errors seem to be popping up here and there with `pre-commit`.

#5589 added ``--warn-unused-ignores``. In theory, adding this option would have removed all unused `# type: ignore` comments in the code at the time this option was added.

However, some errors were not caught, and one popping up recently in #5609 where mypy complained about an unused `# type: ignore`; in that same PR, it also caught a second, unrelated error, which I believe is one of the errors reported in #5563.

Until this problem is tracked down, some PRs will likely be forced to fix seemingly random mypy errors which are unrelated to the PR code changes.

### Steps to reproduce the bug.

Does not seem easily reproducible. But, generally this seems to occur after changes are made to a file which wasn't previously analyzed by `pre-commit mypy` with the current config.

### System Information

```shell
'
```

MatthewFlamm:
One issue is that mypy is run through `pre-commit` that has a different dependency pinning scheme compared to our CI.  It would be more stable to use a traditional `requirements-typing.txt` and then run mypy that way.  This would also fix the problem of having mypy behave differently in pre-commit vs in the dev environment.

MatthewFlamm:
However, it is my experience that we have a LOT of errors when running mypy in a typical dev environment.  This is because the mypy pre-commit config does not include `vtk` and other dependencies in the typing.  It will be a lot of work in one PR to do this.  I suggest that we work towards this goal by adding in dependencies to mypy one at a time.  I started this with #6275

user27182:
> One issue is that mypy is run through `pre-commit` that has a different dependency pinning scheme compared to our CI. It would be more stable to use a traditional `requirements-typing.txt` and then run mypy that way. This would also fix the problem of having mypy behave differently in pre-commit vs in the dev environment.

+1 for using the more traditional setup. In addition to solving the issue of working with two configs, it should also help us actually run `mypy` on the entire project and not just the few modified files that are analyzed with `pre-commit`.

> However, it is my experience that we have a LOT of errors when running mypy in a typical dev environment. This is because the mypy pre-commit config does not include `vtk` and other dependencies in the typing. It will be a lot of work in one PR to do this. I suggest that we work towards this goal by adding in dependencies to mypy one at a time. I started this with #6275

Hopefully adding those dependencies brings the numbers of errors down. And yea I agree it will take a fair amout of work to properly resolve these.

[...]

PR #6268, opened 2024-06-17
🚀 Patch Release 0.43.11

tkoyama010:
### Overview

Fix `pre-commit.ci` errors in `release/0.43` branch.

### Details

- [ ] Fix errors in document building.
- [ ] Fix test errors with `numpy` > 2.0
- [ ] #6275

### DIFF
--- a/pyvista/core/grid.py
+++ b/pyvista/core/grid.py
@@ -243,7 +243,7 @@ class RectilinearGrid(_vtk.vtkRectilinearGrid, Grid, RectilinearGridFilters):
             List of numpy arrays representing the points of this mesh.
 
         """
-        return np.meshgrid(self.x, self.y, self.z, indexing='ij')
+        return np.meshgrid(self.x, self.y, self.z, indexing='ij')  # type: ignore
 
     @property  # type: ignore
     def points(self) -> np.ndarray:  # type: ignore  # numpydoc ignore=RT01
--- a/pyvista/plotting/charts.py
+++ b/pyvista/plotting/charts.py
@@ -3843,9 +3843,7 @@ class BoxPlot(_vtk.vtkPlotBox, _MultiCompPlot):
     def __init__(self, chart, data, colors=None, labels=None):
         """Initialize a new box plot instance."""
         super().__init__(chart)
-        self._table = pyvista.Table(
-            {f"data_{i}": np.asarray(d) for i, d in enumerate(data)}
-        )
+        self._table = pyvista.Table({f"data_{i}": np.asarray(d) for i, d in enumerate(data)})
         self._quartiles = _vtk.vtkComputeQuartiles()
         self._quartiles.SetInputData(self._table)
         self.SetInputData(self._quartiles.GetOutput())


pyvista/core/grid.py
```diff
@@ -243,7 +243,7 @@ def meshgrid(self) -> list:  # numpydoc ignore=RT01
             List of numpy arrays representing the points of this mesh.
 
         """
-        return np.meshgrid(self.x, self.y, self.z, indexing='ij')
+        return np.meshgrid(self.x, self.y, self.z, indexing='ij')  # type: ignore
```
MatthewFlamm, line 246: This isn't a problem only for the release.  It exists in the main branch too.  I'm not sure why it is only showing up in the typing here though.  I saw it in #6275 as well.

MatthewFlamm: The return type is wrong for numpy 2.0.  We need to come up with a decision on how we will type going forward.

larsoner: Maybe a naive suggestion, but you could `list(np.meshgrid(...))` or (probably better / more future-compatible) `tuple(np.meshgrid(...))`, no?

MatthewFlamm: The return type is the same the same as `np.meshgrid`.  In v 1.26, the return type of `np.meshgrid` is `builtins.list[numpy.ndarray[Any, numpy.dtype[Any]]]`.  In v 2.0, the return type is `builtins.tuple[numpy.ndarray[Any, numpy.dtype[Any]], ...]`.

My comment is that we shoudl decide whether we type on numpy version 1.26 or version 2.x.  If we ignore like this, implicitly we are saying that we are going to type numpy version 1.26.  We need to decide this.  I would vote for numpy 2.x.  But this needs to be decided on a more holistic level with the CI testing.

Also our pre-commit mypy testing using different numpy pinning than our CI testing, e.g. for pytest.  This is creating these sorts of problems.

MatthewFlamm: Will mypy be happy with typing it with both list and tuple?

larsoner: Yeah I see that the return type changes for 1.26 vs 2.0. I'm suggesting that if you change the code to:
```
        return tuple(np.meshgrid(self.x, self.y, self.z, indexing='ij'))
```
then it doesn't matter if you type on 1.26 or 2.0, the return type of this `meshgrid` method will always be `-> tuple[np.ndarray]`. But maybe `mypy` wouldn't like this, or you prefer your return type to mirror NumPy's 🤷

MatthewFlamm: I was preferring not to let the type-tail wag the code-dog here, given that this method is a thin wrapper around `np.meshgrid`. But both are valid options.  My main comment is that ignoring the type seems like the wrong solution long term, but may be viable for this release.

user27182: > This isn't a problem only for the release. It exists in the main branch too. I'm not sure why it is only showing up in the typing here though. I saw it in #6275 as well.

This may be the same issue as #5611. `mypy` can be very finicky with imports. If not done correctly `mypy` will treat missing imports as having type `Any`, see https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports. Since `pre-commit` runs `mypy` in an isolated environment and only checks modified files, it may be that `mypy` is unable to properly import from any unmodified files, causing `mypy` to miss imports and treat objects as `Any`. Then, when `mypy` runs again later with _different_ modified files, the imports are successful and an error is raised.

This is a huge problem in my opinion. We should seriously consider dropping `mypy` from pre-commit, or keeping it but using a customized hook to make sure it does what we want. The main thing that's needed I think is to make `mypy` has full project scope for the imports, not just one or two modified files. One option to consider is creating a local pre-commit hook (see https://stackoverflow.com/a/67796237)  to run `mypy` in the local environment, e.g. by calling `mypy pyvista` in the shell instead of using the `pre-commit` environment.

MatthewFlamm: I thought about this, and I think the best solution for this particular use case is to type the return for both versions, i.e. `list` and `tuple`.  Since we are just providing a thin wrapper around `np.meshgrid` it makes sense to type based on all the versions we support.

tkoyama010:
Close by #6348.

PR #6272, opened 2024-06-17
Fix `rgb` option for `add_composite`

### LINKED ISSUE #6012: `rgb` parameter for `add_composite`  not working

If I plot a point cloud with the `rgb` option like this:
``` python
import pyvista as pv
import numpy as np

BLACK = [0.0, 0.0, 0.0]
WHITE = [1.0, 1.0, 1.0]
RED = [1.0, 0.0, 0.0]
GREEN = [0.0, 1.0, 0.0]
BLUE = [0.0, 0.0, 1.0]

rng = np.random.default_rng(seed=1)
points = rng.random((5, 3))

point_cloud = pv.PointSet(points)
point_cloud['colors'] = [WHITE, BLACK, RED, GREEN, BLUE]
point_cloud.plot(render_points_as_spheres = True, point_size = 100, rgb = True)
```

I get this result:
image

But if I try to plot each point as a multiblock like this:

``` python
white_point = pv.PointSet(points[0])
white_point['colors'] = [WHITE]
black_point = pv.PointSet(points[1])
black_point['colors'] = [BLACK]
red_point = pv.PointSet(points[2])
red_point['colors'] = [RED]
green_point = pv.PointSet(points[3])
green_point['colors'] = [GREEN]
blue_point = pv.PointSet(points[4])
blue_point['colors'] = [BLUE]

multiblock = pv.MultiBlock([white_point,black_point,red_point,green_point,blue_point])
multiblock.plot(render_points_as_spheres = True, point_size = 100, rgb = True)
```

I get this result:
image

I expected the second image to match the first. Not sure why they're colored red and blue.

The scalars in the multiblock seem to be ok
``` python
print(multiblock.combine()['colors'])

pyvista_ndarray([[0., 0., 0.],
                 [1., 1., 1.],
                 [1., 0., 0.],
                 [0., 1., 0.],
                 [0., 0., 1.]])
```

For reference, setting `rgb=False` with the multiblock produces this result:

image

See above.

### System Information

```shell
OS : Darwin
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
               RAM : 16.0 GiB
       Environment : IPython
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M2
       GPU Version : 4.1 Metal - 88
  MathText Support : True
  Python 3.12.2 (v3.12.2:6abddd9f6a, Feb  6 2024, 17:02:06) [Clang 13.0.0
  (clang-1300.0.29.30)]
           pyvista : 0.44.dev0
               vtk : 9.3.0
             numpy : 1.26.4
        matplotlib : 3.8.2
            scooby : 0.9.2
             pooch : 1.8.0
            pillow : 10.2.0
           imageio : 2.34.0
           IPython : 8.21.0
          colorcet : 3.1.0
           cmocean : 3.1.3
        ipywidgets : 8.1.2
             scipy : 1.12.0
              tqdm : 4.66.2
            meshio : 5.3.5
        jupyterlab : 4.1.4
    pytest_pyvista : 0.1.8
             trame : 3.5.3
      trame_client : 2.16.1
      trame_server : 2.17.2
         trame_vtk : 2.8.5
     trame_vuetify : 2.4.3
      nest_asyncio : 1.6.0
```

### Screenshots

_No response_

user27182:
Resolves #6012.


pyvista/plotting/plotter.py
```diff
@@ -2846,6 +2846,12 @@ def add_composite(
                 )
 
             if scalars is not None:
+                # enable rgb if the scalars name ends with rgb or rgba
+                if rgb is None:
+                    if scalars.endswith(('_rgb', '_rgba')):
+                        rgb = True
+                        show_scalar_bar = False
```
tkoyama010, line 2853:
```suggestion
                if rgb is None and scalars.endswith(('_rgb', '_rgba')):
                    rgb = True
                    show_scalar_bar = False
```

tests/plotting/test_plotting.py
```diff
@@ -1401,27 +1400,92 @@ def test_read_texture_from_numpy():
...
+            def as_dtype(color: tuple[float, float, float]):
+                return pv.Color(color).int_rgb
+        elif dtype == 'uint8':
+
+            def as_dtype(color: tuple[float, float, float]):
+                return np.array(pv.Color(color).int_rgb, dtype=np.uint8)
+        else:
+            raise NotImplementedError
+
+        return as_dtype
```
tkoyama010, line 1421:
```suggestion
        # Convert color to the specified dtype
        if dtype == 'float':
            def as_dtype(color: tuple[float, float, float]):
                return pv.Color(color).float_rgb
        elif dtype == 'int':
            def as_dtype(color: tuple[float, float, float]):
                return pv.Color(color).int_rgb
        elif dtype == 'uint8':
            def as_dtype(color: tuple[float, float, float]):
                return np.array(pv.Color(color).int_rgb, dtype=np.uint8)
        else:
            raise NotImplementedError
        return as_dtype
```

tkoyama010 (changes requested):

tests/plotting/test_plotting.py
user27182: I'm fine with this, but the formatters (black? Maybe ruff?) want to see a blank line before function definitions, so this will probably be reverted with ci.auto-fix. We could try using lambda functions instead, though ruff may complain about that too and suggest using a function def instead.

tkoyama010: I'm fine if formatter say so. Let's keep as is :)

tkoyama010 (approved):
LGTM. Thanks!

pyvista/core/composite.py
```diff
@@ -1236,6 +1237,10 @@ def _activate_plotting_scalars(self, scalars_name, preference, component, rgb):
         if rgb:
             if scalars.ndim != 2 or scalars.shape[1] not in (3, 4):
                 raise ValueError('RGB array must be n_points/n_cells by 3/4 in shape.')
+            if dtype != np.uint8:
+                # uint8 is required by the mapper to display correctly
+                _validation.check_subdtype(scalars, (np.floating, np.integer), name='rgb scalars')
+                scalars_name = self._convert_to_uint8_rgb_scalars(data_attr, scalars_name)
```
tkoyama010, line 1243: Sorry after I approved it. I was just wondering, is there any vtk documentation or discussion on the basis that uint_8 is required as a data type?

tkoyama010 (changes requested):

pyvista/core/composite.py
user27182: The pyvista mappers use `color_mode='direct'` for plotting rgb scalars:
https://github.com/pyvista/pyvista/blob/693569ad478d4905ab28e5c3dc08001b784ce660/pyvista/plotting/mapper.py#L186-L188
The docs for [vtkMapper.SetColorModeToDirectScalars()](https://vtk.org/doc/nightly/html/classvtkMapper.html#a6e34c4582491a60b4621f018ff6416d8) says that float and int types should just work, and uint8 is not specifically required.

This is consistent with how `pv.DataSetMapper` / `vtkDataSetMapper` works. But  #6012 shows it's not working for `pv.CompositePolyDataMapper` / `vtk.vtkCompositePolyDataMapper`. Not sure exactly why that is though.

tkoyama010: Thank you. I will report this to the vtk project later. If there are no other comments, I will merge it in 24 hours.

tkoyama010 (approved):

PR #6273, opened 2024-06-17
Add a Dependabot config to auto-update GitHub action versions

kurtmckee:
### Overview

Currently, deprecation warnings are being thrown due to actions using out-of-date Node versions ([recent example](https://github.com/pyvista/pyvista/actions/runs/9555140364)). Rather than submitting a PR to update them manually, this PR configures Dependabot to regularly submit PRs to update GitHub action versions as needed.

If this PR merges, you can expect Dependabot to immediately submit a PR to update actions to newer versions.

### Details

- Add a Dependabot config to auto-update GitHub action versions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -30,6 +30,9 @@ updates:
         patterns:
           - "actions/upload-artifact"
           - "actions/download-artifact"
+      github-actions:
+        patterns:
+          - "*"
     labels:
       - "maintenance"
       - "dependencies"


tkoyama010:
Thanks for your suggestion. We already have [.github/dependabot.yml](https://github.com/pyvista/pyvista/blob/main/.github/dependabot.yml) in main branch. To be clarify, do we need extra file?

kurtmckee:
Goodness gracious, how did I miss that?! I'll update this as soon as I'm back in front of the right computer.

kurtmckee:
@tkoyama010 Thanks for reviewing this PR! I've amended the commit and force-pushed. I *think* that this will now have the desired effect to auto-update all GitHub actions, with `*-artifact` actions specifically grouped together.

MatthewFlamm:
Our current setup already bumps action versions for all actions.  Or is there a different meaning here?  See https://github.com/pyvista/pyvista/pull/6269 as an example.  You don't have to add '"*" to a group to get this behavior.

This PR groups all actions version updates together in a single PR rather than individual PRs.  I'd probably keep the current state unless we are getting too many action version bump PRs.

kurtmckee:
@MatthewFlamm Hmm, I think I understand what's happening here.

A draft PR you opened a few hours ago is [showing a number of deprecation warnings](https://github.com/pyvista/pyvista/pull/6275/checks) for `actions/cache@v3`...but now that I'm digging into that, I see that `actions/cache` has been updated to `actions/cache@v4` throughout those workflows.

However, `awalsh128/cache-apt-pkgs-action@v1.1.3`, in `docs.yml` is approaching two years old:

https://github.com/pyvista/pyvista/blob/11d39721f364cc06c689b769f10158a83fde70bd/.github/workflows/docs.yml#L47-L50

The latest version is v1.4.2, but Dependabot's PR to update that, #5740, was closed instead of merging, perhaps due to an incompatibility between 1.1.3 and 1.4.2?

It appears that this PR can be closed; I misinterpreted the warnings as if Dependabot was not configured properly. Sorry for the confusion!

MatthewFlamm:
Thank you for the help!  Sounds like we need to investigate this action.

kurtmckee:
You're welcome! :grinning:

PR #6280, opened 2024-06-20
Numpy updates 2.0 to testing, aligning return types

MatthewFlamm:
### Overview

We had implemented numpy2.0 check only for pytest using Linux there are several other places that are now showing errors when updating to numpy2.0.

### Details

- [x] docstring examples
  - A lot of numpy functions now return numpy scalars that have a repr that doesn't match our doctests
- [x] mypy
- [x] windows testing
- [x] linux python 3.9 testing (numpy version mismatch?)


pyvista/core/_validation/_cast_array.py
```diff
@@ -122,7 +122,8 @@ def _cast_to_numpy(
     try:
         VisibleDeprecationWarning = np.exceptions.VisibleDeprecationWarning
     except AttributeError:
-        VisibleDeprecationWarning = np.VisibleDeprecationWarning
+        # we only type for newer numpy, and this branch only touched in older numpy
```
MatthewFlamm, line 125: As far as I can tell, we can only type based on a single version of numpy.  Should we declare the newest supported version of any dependency to be the version of truth here?

This is a particular problem supporting such wide ranges of dependency versions.  We could remove this pain if we narrow support to less versions of vtk and numpy.  Something to consider to avoid this maintenance headache.

pyvista/core/grid.py
```diff
@@ -255,7 +255,7 @@ def _from_arrays(
         self._update_dimensions()
 
     @property
-    def meshgrid(self) -> list[NumpyArray[float]]:
+    def meshgrid(self) -> tuple[NumpyArray[float], ...]:
```
MatthewFlamm, line 258: Similar comment about typing based on numpy versions here.  It is a little more impactful to users however, as it is user facing and can be wrong depending on numpy version.

I'm not sure that we want to override the type to always be `tuple` however, as suggested in https://github.com/pyvista/pyvista/pull/6268#discussion_r1644906529, since this function is declared to be a simple wrapper of `np.meshgrid`.

MatthewFlamm: This is now typed with _both_ types so it is correct for both versions of numpy.

ant.vtk
MatthewFlamm, line 1: This is generated by running doctests locally.  We either need to change the doctest or add it to .gitignore.

pyvista/core/filters/image_data.py
```diff
@@ -430,7 +430,7 @@ def image_threshold(
             array_dtype := self.active_scalars.dtype,
             int,
         ) and array_dtype != np.dtype(np.uint8)
-        cast_dtype = has_int_dtype and platform.system() in ['Darwin', 'Linux']
+        cast_dtype = has_int_dtype and platform.system() in ['Darwin', 'Linux', 'Windows']
```
MatthewFlamm, line 433: @user27182 I am neither a Windows user nor a a user of this functionality.  Windows fails on this test with numpy 2.0.  Can we get rid of this platform check altogether (I would get rid of the whole check in this case, but did this just to get the CI green).  It seems like it is safest to do the type conversion no matter the platform, and then the behavior will also be consistent.  Thoughts?

MatthewFlamm:
cc-@adeak since you spearheaded the prior sprint to get numpy 2.0 working and in case you have time.  This PR is more edge cases.

adeak:
Thanks for the ping, I'll try to take a look.

>  windows testing

I hope this doesn't mean actual user-facing errors. I thought the test matrix with NumPy nightly also applied to Windows, with no errors.

pyvista/core/_validation/_cast_array.py
adeak: If I understand the issue it might also be an option to use `if not typing.TYPE_CHECKING: ...`.

.github/workflows/testing-and-deployment.yml
```diff
@@ -149,21 +149,22 @@ jobs:
...
         include:
+          # numeric numpy versions are < conditions, e.g. "1.24" means "numpy<1.24"
           - python-version: "3.8"
             vtk-version: "9.0.3" # Requires numpy<1.24
-            numpy-version: "default"
+            numpy-version: "1.24"
           - python-version: "3.9"
             vtk-version: "9.1"
-            numpy-version: "default"
+            numpy-version: "2.0" # matplotlib and numpy incomptability for python 3.9
```
adeak, line 158:
```suggestion
            numpy-version: "2.0" # matplotlib and numpy incompatibility for python 3.9
```

adeak, line 158: This indeed sounds weird. We might want to exclude this combination in our `pyproject.toml` unless it's a huge pain in the ass to specify.

pyvista/core/filters/image_data.py
user27182: Sure let's get rid of it this check and always convert to be safe. The checks were conservative and made just to pass our specific CI builds for that PR at the time. But clearly the issue is broader in scope.

.github/workflows/testing-and-deployment.yml
```diff
@@ -205,9 +206,9 @@ jobs:
       - name: Install Testing Requirements
         run: pip install -r requirements_test.txt
 
-      - name: Limit NumPy for VTK 9.0.3
-        if: ${{ matrix.vtk-version == '9.0.3' }}
-        run: pip install 'numpy<1.24'
+      - name: Set up numpy
+        if: ${{ matrix.numpy-version != 'latest' && matrix.numpy-version != 'nightly' }}
+        run: pip install 'numpy<${{ matrix.numpy-version }}'
```
adeak, line 211: Having `numpy < matrix.numpy-version` is a but surprising to me, even though I see that the result preserves `numpy<1.24`. So how about setting the highest known `1.23` version instead, and using `<=` or a similar operator?

adeak (commented):
First quick pass, overall looks great, thanks @MatthewFlamm. I'm leaning toward changing some of our more cluttery-repred API, see comments below. I haven't looked at any local test outputs yet.

pyvista/core/grid.py
adeak: Ah, I missed this thread, sorry. Independent discovery of always converting to `tuple` :P

.github/workflows/testing-and-deployment.yml
MatthewFlamm: I think I might get rid of this comment altogether, as it is also nice to have testing for numpy 1.26.  With the comment, someone might figure out the versioning and remove this pin.

pyvista/examples/downloads.py
```diff
@@ -7472,7 +7472,7 @@ def download_whole_body_ct_male(load=True):  # pragma: no cover
 
     >>> label_map = dataset['label_map']
     >>> label_map.get_data_range()
-    (0, 117)
+    (np.uint8(0), np.uint8(117))
```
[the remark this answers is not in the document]
user27182: I guess the issue here is that `get_data_range` returns an array with the same dtype as the array from which the range is computed.

instead of `tuple(array)` we could do `tuple(array.tolist())` and let numpy convert to the nearest built-in type. See https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tolist.html.

For 1-d arrays this is fine, but for multi-dimensional arrays we need to be careful since a nested list will be generated, but `tuple` won't be recursively applied to the nested lists. For these cases `_validation.validate_array(array, to_tuple=True)` can be used,  or maybe even just `_validation._cast_array._cast_to_tuple` which is used by that function internally.

pyvista/core/grid.py
```diff
@@ -254,8 +254,9 @@ def _from_arrays(
         # Ensure dimensions are properly set
         self._update_dimensions()
 
+    # typing depends on numpy version
```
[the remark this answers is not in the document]
MatthewFlamm: This method is declared as a simple wrapper for `np.meshgrid`, if we do this, it is no longer technically true.  It is a wrapper that enforces a return type. I think the practical importance is pretty low in either case.  I've heard two votes for converting to tuple always, I'll wait for more feedback before implementing it though.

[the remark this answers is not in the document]
adeak: To me the wrapper reference is just about semantics rather than strict API adherence. I'd expect any user actually affected by this to benefit from removing this moving target. We could rephrase the docs as "implemented as a thin wrapper for meshgrid".

pyvista/core/composite.py
```diff
@@ -190,7 +190,7 @@ def bounds(self) -> BoundsLike:
         ... ]
         >>> blocks = pv.MultiBlock(data)
         >>> blocks.bounds
-        (-0.5, 2.5, -0.5, 2.5, -0.5, 0.5)
+        (np.float64(-0.5), np.float64(2.5), np.float64(-0.5), np.float64(2.5), np.float64(-0.5), np.float64(0.5))
```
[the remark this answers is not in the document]
tkoyama010: In #6271, I supported following the numpy data types, but your point made me realize that the more important thing here is that the data type is a tuple of floats. I created a patch for this PR at #6282, so please check it out.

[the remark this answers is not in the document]
MatthewFlamm: Let's keep the work here as it is likely that one decision for one part will trickle over into other parts. 

> How annoying do you think this repr change is? We're going out of our way to return a tuple instead of a numpy array in `.bounds`, so we could just explicitly return a tuple of native floats. 

I think this whole function could be simplified to directly form a tuple rather than doing this to `numpy.stack` and `ravel` magic here.


> And fix the regular mesh .bounds as well.

This needs more digging.  This typing `cast` makes me suspicious about the `DataSet.bounds` method too.  It might just be a length of container thing, but it is worth aligning these methods.

https://github.com/pyvista/pyvista/blob/c6ee1a6cc9d56774f2140007455e70e7b1f75318/pyvista/core/dataset.py#L1752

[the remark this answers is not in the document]
tkoyama010: I separate #6282 as other PR. Please go on as is here.

MatthewFlamm:
I still have a bunch of methods to go back through and see if it makes sense to go back to enforce Python scalars. But, if this desire for Python scalars is driven mainly by documentation, we should look at what numpy does for scalars, as they do not have this issue in their docs.  See: 

https://numpy.org/doc/stable/reference/generated/numpy.argmin.html

.github/workflows/testing-and-deployment.yml
MatthewFlamm: Now using ~= checks.

.github/workflows/testing-and-deployment.yml
MatthewFlamm: I'm not super confident this is true, but our CI couldn't resolve compatible dependencies.  This pin is now for testing numpy 1.26 rather than specifically to avoid incompatibility.

If someone wants to put in the time to figure this out, it makes sense to me to put into `pyproject.toml`, although I'm not sure that python dependency specification allows for conditions on dependency versions.  I suppose we could do it by Python version, but it is a bit of an incomplete situation.  What if it is possible for a user on Python 3.10 to also get the problematic combination of the two packages, for example?

It should be up to matplotlib to handle this in an ideal world, but they run on a (much) narrower support window than we do.

pyvista/examples/downloads.py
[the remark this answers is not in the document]
MatthewFlamm: `get_data_range` in all cases does a `return value1, value2`, so it was most straightforward to do something like `return value1.item(), value2.item()`.  We now also use `math.nan` to avoid returning the `np.nan`, which messes up the typing.

pyvista/core/grid.py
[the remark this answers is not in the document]
MatthewFlamm: I think the practical importance of the change is not important.  I took the advice to be opinionated and return `tuple` type.

pyvista/core/composite.py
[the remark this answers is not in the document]
MatthewFlamm: I am using something different than #6282 now.  Take a look please @tkoyama010

.github/workflows/testing-and-deployment.yml
MatthewFlamm: I learned the hard way that `numpy~=1.23` will install version 1.26.4.  It needs to be `numpy~=1.23.0` which will give the latest patch release from 1.23.x.  This is essentially the same as `numpy<1.24`.  But the ~= usage is nicer for the CI, so the name of the job matrix is accurate.

pyvista/examples/downloads.py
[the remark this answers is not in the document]
adeak: > We now also use `math.nan` to avoid returning the `np.nan`, which messes up the typing.

Can you please elaborate on this? Because
```
>>> type(np.nan)
<class 'float'>
```

adeak:
> I still have a bunch of methods to go back through and see if it makes sense to go back to enforce Python scalars. But, if this desire for Python scalars is driven mainly by documentation, we should look at what numpy does for scalars, as they do not have this issue in their docs. See:
> 
> https://numpy.org/doc/stable/reference/generated/numpy.argmin.html

[Sounds like an option of `np.set_printoptions()`](https://github.com/numpy/numpy/issues/26730#issuecomment-2174254475).

For what it's worth documentation is only part of the issue. I also expect that most of our users work interactively, and this bloated repr is in most cases not helpful, and very distracting. It _might_ make sense for the integers (because you have to be aware of possible integer overflow compared to native ints), but e.g. that's not the case for things like `Plotter` subfigure indices, and it's never the case for `np.float64` because native Python floats are "the same" doubles.

pyvista/core/grid.py
adeak: Isn't this always a 3-length tuple?
```suggestion
    def meshgrid(self) -> tuple[NumpyArray[float], NumpyArray[float], NumpyArray[float]]:
```
(modulo autoformatting)

pyvista/core/dataset.py
```diff
@@ -1862,15 +1864,16 @@ def area(self) -> float:
...
 
-        Get the area of a sphere.
+        Get the area of a sphere. Discretization error results
+        in slight difference from ``pi``.
 
         >>> mesh = pv.Sphere()
-        >>> mesh.volume
-        0.51825
+        >>> mesh.area
+        3.13
```
adeak, line 1872: Good catch.

MatthewFlamm:
Converting this to draft as I work on aligning your usage.  I am going to summarize my proposal here while I chip away at it.  I think it will be easier to discuss in this format now rather than interleaved in the code.

I think it is important to note that we have always returned numpy scalars in these cases, but the inconsistency in our returns with all the other methods is now laid bare with the change in `repr` for numpy scalars.  This PR aims to tighten consistency while also providing full CI testing for numpy 2.0.

- [ ] `MultiBlock.bounds`: return `tuple[python scalar]` to be consistent with `DataSet.bounds`
- [ ] `DataSet.get_data_range`: return `tuple[numpy scalar]`.  This is because the result is directly from a mesh data array.  It is a thin wrapper of `ndarray.min` and `ndarray.max`.  Note here that we always returned numpy types, it is just the numpy scalar repr that changes.  I want feedback on this one before I implement as I suspect there could be other opinions. 
- [ ] `DataSet.area` and `DataSet.volume`, while under the hood each are a sum of a mesh data array.  It is not a 1:1 correspondence to an existing data array in most cases.  Return Python scalar.
- [ ] `DataSetFilters.point_is_inside_cell`.  Returns either single bool or `ndarray[bool].  I cannot think of a reason to return `np.bool_` scalar other than slight consistency within the `ndarray` return.
- [ ] `meshgrid`. Return `tuple` to be consistent with new numpy. Minor point, look into specifying exact length of tuple.
- [ ] `PolyData.is_all_triangles`. return Python True.  Same reason as above. 
- [ ] `ExplicitStructuredGrid.cell_id`.  Weirdly returns either tuple or ndarray.  Align to always return ndarray.
- [ ] `ExplicitStructuredGrid.coords`.  Can return either scalar or ndarray.  Keep as numpy scalar to be consistent. This is also an integer where the impact could be more important.
- [ ] `Plotter.shape`.  While an sequence of integers, it is returned as a tuple.  Make a tuple of python scalars.
- [ ] `Plotter.bounds`.  align with other `bounds`, use tuple[python scalar]
- [ ] `Plotter.center`.  this one is alkin to `Plotter.bounds`.  It is currently a list, but probably should be a tuple for consistency.  Return `tuple[python scalar]`.
- [ ] `Plotter.where_is` should mirror `Plotter.shape`.  return `list[tuple[python scalar]]`
- [ ] `Renderers.index_to_loc` feels like an internal method for the most part.  It returns either `ndarray` or a scalar.  Return a numpy scalar in that case. Again also an integer.

adeak:
I fully agree with all of the above (as of revision 2 just in case there are any fundamental edits later).

> Minor point, look into specifying exact length of tuple.

Do you just mean how to specify a 3-length tuple? Because that would be [this](https://github.com/pyvista/pyvista/pull/6280#discussion_r1649430956). Did you mean something else?

> `Renderers.index_to_loc` feels like an internal method for the most part.

I was going to comment that this should be safe to assume to be unused downstream, but [I've easily found external use](https://github.com/SimBe-hub/PySubdiv/blob/3cb898c73c5b085673024d16c1c5a8689e3af191/PySubdiv/create_control_cage/vtk_custom_widget.py#L369) and [there are probably other examples hidden in the query that gave me this one](https://github.com/search?q=%2F%5C.index_to_loc%5C%28%2F+language%3APython&type=code). But yeah, "for the most part" this probably doesn't affect anyone, and even those who use it shouldn't be hindered by us cleaning this up a bit.

MatthewFlamm:
> Do you just mean how to specify a 3-length tuple? Because that would be [this](https://github.com/pyvista/pyvista/pull/6280#discussion_r1649430956). Did you mean something else?

I believe so, I was being lazy and not wanting to look up whether it is always 3 length, etc.

MatthewFlamm:
> But yeah, "for the most part" this probably doesn't affect anyone, and even those who use it shouldn't be hindered by us cleaning this up a bit.

In reference to `Renderers.index_to_loc`. To be clear, the clean up I'm proposing is to return ndarray or a numpy scalar.

adeak:
> > But yeah, "for the most part" this probably doesn't affect anyone, and even those who use it shouldn't be hindered by us cleaning this up a bit.
> 
> In reference to `Renderers.index_to_loc`. To be clear, the clean up I'm proposing is to return ndarray or a numpy scalar.

You were clear, I just wasn't paying attention because it doesn't matter either way.

bjlittle:
@MatthewFlamm Apologies. I'll spin-up on unblocking `geovista`.

I'm hoping to make a 0.5.0 release soon, so I can fold the changes into that release to make it `numpy` 2.x compliant 👍

adeak:
> It was not simple to switch the Plotter.shape to be a tuple of python int. There is some convoluted logic used that compares shape and/or length of `Plotter.shape`, `Renderers.shape` and/or `Renderers._shape` in multiple methods including but not limited to `Renderers.index_to_loc`. I don't have time to look into this piece any more in the short term. Given that this isn't just a user facing detail, it is pushing me to keep as numpy integers. If we are okay with this path forward, I will also revert the changes to `Plotter.where_is` for consistency. Someone else could take up the torch to update these methods in a future PR.

Sure, the conversion decisions here are just a nice addition but the real point is completing NumPy compatibility. Don't waste too much time on these.

I'll try to look into this. Can you say a few words about what kind of issues you've encountered? Any specific tests that kept failing? Or a specific scenario where this ended up being a loose yarn end in a sweater that started pulling apart the whole thing?

MatthewFlamm:
> I'll try to look into this. Can you say a few words about what kind of issues you've encountered? Any specific tests that kept failing? Or a specific scenario where this ended up being a loose yarn end in a sweater that started pulling apart the whole thing?
Also concerning is that there is no direct test for Renderers.shape and Plotter.shape that I could easily find. This makes me less comfortable touching this in this PR.

See workflow run before I reverted the changes in https://github.com/pyvista/pyvista/pull/6280/commits/115ed96e318e9b116b8377af0216c59bc218ac04

https://github.com/pyvista/pyvista/actions/runs/9649372560

An example of a failing test is
> FAILED tests/plotting/test_plotting.py::test_multi_renderers_subplot_ind_3x1 - AttributeError: 'int' object has no attribute 'item'

I had started by adding a test that looks something like this to reproduce it

```py
def test_plotter_shape():
    pl = pv.Plotter(shape=(1, 2))
    assert pl.shape == (1, 2)
    pl.subplot(0)  # this shouldn't really part of this test, but will cause a failure
```

adeak:
It's late and I'm tired so I might be missing something, but it seems to me that you were very close.

The issue I see in `115ed96^` (i.e. fc656cd) is that `Renderers.shape` is set to a tuple-of-numpy-scalars on only one of three conditional branches. Applying the following patch to fc656cd:
```diff
diff --git a/pyvista/plotting/renderers.py b/pyvista/plotting/renderers.py
index 49164fd9f..8044262de 100644
--- a/pyvista/plotting/renderers.py
+++ b/pyvista/plotting/renderers.py
@@ -117,8 +117,8 @@ class Renderers:
             shape = np.asarray(shape)
             if not np.issubdtype(shape.dtype, np.integer) or (shape <= 0).any():
                 raise ValueError('"shape" must contain only positive integers.')
-            # always assign shape as a tuple
-            self._shape = tuple(shape)
+            # always assign shape as a tuple of native ints
+            self._shape = tuple(size.item() for size in shape)
             self._render_idxs = np.empty(self._shape, dtype=int)
             # Check if row and col weights correspond to given shape,
             # or initialize them to defaults (equally weighted).
@@ -344,7 +344,7 @@ class Renderers:
         tuple[int, int]
             Shape of the renderers.
         """
-        return tuple(s.item() for s in self._shape)
+        return self._shape
 
     def set_active_renderer(self, index_row, index_column=None):
         """Set the index of the active renderer.
```
makes all plotting tests pass locally.

Of course your points about not wanting to touch this stands. Just FYI in case this makes you change your mind.

adeak:
> Thanks! If it is different in other branches, we need to consider those too to make sure it is consistent. Let me take a look at how complex the other branches are...

The other two branches explicitly construct tuples of ints, this is why things started breaking a lot when you assumed they were always numpy scalars in there.

MatthewFlamm:
> The other two branches explicitly construct tuples of ints, this is why things started breaking a lot when you assumed they were always numpy scalars in there.

This makes it higher priority to align the types here.

MatthewFlamm:
I got myself turned around because the test I provided above was flawed.  The error message it gives is unhelpful to figure it why it was flawed.

pyvista/examples/downloads.py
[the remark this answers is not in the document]
MatthewFlamm: I believe you are right, but now this is moot as we are using numpy scalars here.

pyvista/core/grid.py
```diff
@@ -264,11 +265,18 @@ def meshgrid(self) -> list[NumpyArray[float]]:
...
+            Tuple of numpy arrays representing the points of this mesh.
 
         """
-        return np.meshgrid(self.x, self.y, self.z, indexing='ij')
+        # Converting to tuple needed to be consistent type across numpy version
+        # Remove when support is dropped for numpy 1.x
+        # We also know this is 3-length so make it so in typing
+        out = tuple(np.meshgrid(self.x, self.y, self.z, indexing='ij'))
+        # Python 3.8 does not allow subscripting tuple, but only used for type checking
+        if TYPE_CHECKING:
```
MatthewFlamm, line 277:
```suggestion
        if TYPE_CHECKING:  # pragma: no cover
```

MatthewFlamm:
This is ready for full review.  Pinging @banesullivan for changes to CI approval (also any other comments).

pyvista/plotting/renderers.py
```diff
@@ -312,13 +312,13 @@ def index_to_loc(self, index):
...
         -------
-        numpy.ndarray
+        numpy.ndarray or numpy.int64
             2D location on the plotting grid.
         """
         if not isinstance(index, (int, np.integer)):
             raise TypeError('"index" must be a scalar integer.')
         if len(self.shape) == 1:
-            return index
+            return np.int64(index)
```
adeak, line 321: I'm wondering if we should change this to
```suggestion
            return np.intp(index)
```
(the return type annotation is probably better left as `int64` for end users).

[To quote Ralf Gommers](https://github.com/numpy/numpy/issues/9464#issuecomment-2185930655):
> This has been addressed in NumPy 2.0 by changing the default integer type to `intp`, which has the practical consequence of changing the default on 64-bit Windows from `int32` to `int64`.

[To quote the 2.0 release notes](https://numpy.org/doc/stable/release/2.0.0-notes.html#redefinition-of-np-intp-np-uintp-almost-never-a-change):
> Due to the actual use of [`np.intp`/`np.uintp`] almost always matching the use of `size_t`/`Py_ssize_t` this is now the definition in C. Previously, it matched `intptr_t` and `uintptr_t` which would often have been subtly incorrect. This has no effect on the vast majority of machines since the size of these types only differ on extremely niche platforms.

"Extremely niche platforms" are unlikely to be our userbase, right?

It can also make sense to go with explicit `int64`, but then we must ensure that all code paths return that, and that the array case has the same dtype.

pyvista/plotting/renderers.py
```diff
@@ -336,12 +336,12 @@ def active_renderer(self):  # numpydoc ignore=RT01
         return self._renderers[self._active_index]
 
     @property
-    def shape(self):  # numpydoc ignore=RT01
+    def shape(self) -> tuple[int, int]:  # numpydoc ignore=RT01
```
adeak, line 339: On the first code path we have `self._shape = (n + m,)` so I think we need
```suggestion
    def shape(self) -> tuple[int] | tuple[int, int]:  # numpydoc ignore=RT01
```
(and same in the docstring).

And why do we need `RT01`? [Isn't that "No returns section found"](https://numpydoc.readthedocs.io/en/v1.6.0rc1/validation.html#built-in-validation-checks)?

pyvista/plotting/plotter.py
```diff
@@ -982,12 +982,12 @@ def _before_close_callback(self, func):
             self.__before_close_callback = None
 
     @property
-    def shape(self):  # numpydoc ignore=RT01
+    def shape(self) -> tuple[int, int]:  # numpydoc ignore=RT01
```
adeak, line 985: Same remark as `Renderers`:
```python
>>> pv.Plotter(shape='2/3').shape
(5,)
```

pyvista/core/filters/image_data.py
```diff
@@ -424,13 +423,12 @@ def image_threshold(
         else:
             field = self.get_array_association(scalars, preference=preference)
 
-        # For some systems and/or Numpy < 2.0, integer scalars won't threshold
-        # correctly. Cast to float in these cases.
-        has_int_dtype = np.issubdtype(
+        # For some systems integer scalars won't threshold
+        # correctly. Cast to float in these cases to be robust.
```
adeak, line 427: Since we no longer have a platform check:
```suggestion
        # correctly. Cast to float to be robust.
```

pyvista/core/grid.py
```diff
@@ -264,11 +265,18 @@ def meshgrid(self) -> list[NumpyArray[float]]:
 
         Returns
         -------
-        list[numpy.ndarray]
-            List of numpy arrays representing the points of this mesh.
+        tuple[numpy.ndarray]
+            Tuple of numpy arrays representing the points of this mesh.
```
adeak, line 269: Should we mention the length 3 either on the type line or the short description?

adeak (commented):
Few probably-last minor comments.

pyvista/plotting/renderers.py
adeak: Although the "size of these types only differ" might have referred to `size_t` vs `intptr_t` rather than `int64` vs `intp` so that quote might be a weaker indication for my point.

MatthewFlamm: Given your quote it sounds like the preference should be `np.intp`, although I'm not fully grokking the practical difference between `np.int_` and `np.intp`.  Given `np.intp` has something to do with pointer sizes, and this parameter is an index value, it makes logical sense here.

I highly suspect that there will be no impact to any of the choices here.

pyvista/core/grid.py
MatthewFlamm: Put in length three explicitly into the tuple here.

pyvista/plotting/renderers.py
MatthewFlamm: I was just leaving things as they were, but since we are touching these, I cleanup up.  I removed any ones I could find touched.

MatthewFlamm: Also added to the returns sections

adeak (approved):
Looks good to me, thanks a lot @MatthewFlamm.

Looks like our docstring checking segfaults, but that has to be unrelated to changed here.

banesullivan (approved):

PR #6281, opened 2024-06-20
Fix `Text3DSource` garbage collection

user27182:
### Overview

This class stores references to additional filters in `__init__` for internal use by the class. Turns out that this causes the plotting gc checks to fail when a parent class which stores its own references to `Text3DSource` is used as a pytest fixture (the internal filters are not properly gc'd). This PR fixes this by moving the filters out of init.

No tests are added with this PR to verify that the gc is working. But a separate PR will be created later with a new `Text3DFollower` class (from #6263), and tests for that class will implicitly verify that the gc is working.


tkoyama010 (approved):
Good catch. Thanks!

PR #6282, opened 2024-06-20
Change type of `bounds` element from np.float to float

tkoyama010:
### Overview

resolve https://github.com/pyvista/pyvista/pull/6280/files#r1648119718

### Details

- None

### DIFF
--- a/pyvista/core/composite.py
+++ b/pyvista/core/composite.py
@@ -190,7 +190,7 @@ class MultiBlock(
         ... ]
         >>> blocks = pv.MultiBlock(data)
         >>> blocks.bounds
-        (np.float64(-0.5), np.float64(2.5), np.float64(-0.5), np.float64(2.5), np.float64(-0.5), np.float64(0.5))
+        (-0.5, 2.5, -0.5, 2.5, -0.5, 0.5)
 
         """
         # apply reduction of min and max over each block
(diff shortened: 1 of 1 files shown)


pyvista/core/composite.py
```diff
@@ -205,7 +205,7 @@ def bounds(self) -> BoundsLike:
             maxima = np.maximum.reduce(all_bounds)[1::2]
 
         # interleave minima and maxima for bounds
-        the_bounds = np.stack([minima, maxima]).ravel('F')
+        the_bounds = np.stack([minima, maxima]).ravel('F').tolist()
```
MatthewFlamm, line 208: I feel like it might be more efficient to explicitly form the tuple directly rather than creating a new ndarray, convert to list, then convert to tuple.  But I'm not sure.  There is another method, `item` that can convert a numpy scalar from an array to a python scalar.  I think it would come down to whether this being optimized within `tolist` outweighs the multiple container conversions.  Since `tolist` documents that it uses `item` under the hood, I have my doubts there is benefit here.

Either way, I think it is likely we will be making this decision as a whole in the other PR so it makes sense to do there.

tkoyama010:
Switching merge target branch as treat is as other PR.

adeak (changes requested):
After the target branch change this now has a ton of unrelated changes from @MatthewFlamm. Please fix.

MatthewFlamm:
If those changes are removed then we won't be testing the changes here.  This strategy of opening up a parallel PR creates a lot of friction for both PRs.

tkoyama010:
I can wait until #6280 merged.

PR #6283, opened 2024-06-21
Add `AxesGeometrySource` for generating axes datasets

user27182:
### Overview

Follow-up from #6263. This PR adds a new class for generating the core geometric components of arrow-style axes. 

### Details

The class is similar to `pyvista.AxesActor` but has some key differences:

- `AxesActor` is an actor defined in `plotting`; this class is defined in `core` and returns a `MultiBlock` 
- `AxesActor` uses normalized lengths and is not dimensionally "true-to-scale"; the output from this class is "true-to-scale", and optionally supports normalized lengths as well using a `normalized_mode` property
- `AxesActor` only supports line or cylinder shafts and cone or sphere tips; this class does not support line shafts, but otherwise supports more geometries (e.g. cube or pyramid) and even supports arbitrary datasets.

Another key feature supported by the `AxesGeometrySource` class is the use of auto-updating features for dependent length properties. The axes lengths are defined with three properties: `shaft_length`, `tip_length`, and `total_length`, but in reality only two properties are needed to fully define the lengths.  As such, when `normalized_mode` is enabled, the `shaft_length` and `tip_length` are auto-updated such that their sum is always 1. Or, when `normalized_mode` is disabled, the `shaft_length` and `total_length` are auto-updated such that `shaft_length` + `tip_length` = `total_length`.

Other differences between these classes exist and will be supported in future PRs:
- `AxesActor` cannot be oriented arbitrarily in a scene; since `AxesGeometrySource` outputs a regular dataset, it is inherently orientable with transformations (`position` and `direction_vectors` properties will be added in a separate PR)
- `AxesActor` cannot be made symmetric;  a `symmetric` property will be added to `AxesGeometrySource` in a separate PR
- `AxesActor` supports setting axes colors via actor properties; `AxesGeometrySource` can support colors using `rgb` arrays (color properties will be added in a separate PR)
- `AxesActor` has labels; `AxesGeometrySource` does not support labels (labels will be added by `AxesAssembly` in [#6263](#6263))


pyvista/core/utilities/geometric_sources.py
```diff
@@ -2812,3 +2819,1043 @@ def output(self):
...
+                    # Move tip to end of shaft
+                    part.points[:, axis] += shaft_length[axis]
+
+                if self.symmetric:
+                    # Flip and append to part
+                    origin = [0, 0, 0]
+                    normal = [0, 0, 0]
+                    normal[axis] = 1
+                    flipped = part.flip_normal(normal=normal, point=origin)
+                    part.append_polydata(flipped, inplace=True)
```
tkoyama010, line 3727: I recommend to use [itertools.product](https://docs.python.org/3/library/itertools.html#itertools.product) in this kind of situation.
```suggestion
        for part_type, axis in itertools.product(_PartEnum, _AxisEnum):
            for axis in _AxisEnum:
            # Reset geometry
            part = AxesGeometrySource._normalize_part(nested_datasets[part_type][axis])

            # Offset so axis bounds are [0, 1]
            part.points[:, axis] += 0.5

            # Scale by length along axis, scale by radius off-axis
            radius, length = (
                (shaft_radius, shaft_length)
                if part_type == _PartEnum.shaft
                else (tip_radius, tip_length)
            )
            diameter = radius * 2
            scale = [diameter] * 3
            scale[axis] = length[axis]
            part.scale(scale, inplace=True)

            if part_type == _PartEnum.tip:
                # Move tip to end of shaft
                part.points[:, axis] += shaft_length[axis]

            if self.symmetric:
                # Flip and append to part
                origin = [0, 0, 0]
                normal = [0, 0, 0]
                normal[axis] = 1
                flipped = part.flip_normal(normal=normal, point=origin)
                part.append_polydata(flipped, inplace=True)
```

tkoyama010 (changes requested):

user27182:
@tkoyama010 looks like there may be some issues with the docs build. Not sure if it's related to caching. 

But this PR adds new docstrings with plots. I was expecting the first build to fail so I could add the failed images to the doc_image_cache. But the build did not fail and the debug images do not contain anything from `AxesGeometrySource`.

I was also expecting a build failure from this rst error, fixed in https://github.com/pyvista/pyvista/pull/6283/commits/a66d133d14ae21f4ff1819e23d0f39f139186c57. But again no error, and the build is passing.

user27182:
I've removed non-core features from this class for now (such as colors and transformations), these can be added in a separate PR. Hopefully this makes it easier for review.

user27182:
@pyvista-bot preview

user27182:
@pyvista-bot preview

user27182:
@pyvista-bot preview

user27182:
I think the `normalized_mode` option is a bit complex. I'll try to also remove that option and leave it for a separate PR

user27182:
Perhaps we shouldn't even include a `normalized_mode`and `total_length` paramater at all, i.e. let's not make a separate Pr to add the removed changes from https://github.com/pyvista/pyvista/pull/6283/commits/2ba9fc298fceb9d68a40dba8c32e5226a6186e11. I didn't realize how unnecessarily complicated this option was until I removed it.

`total_length` is a concept borrowed from `AxesActor`, and I guess it makes sense there since those axes lengths are normalized, but it is still a bit unintuitive and exhibits strange behavior. However, since `AxesGeometrySource` is intended to be true-to-scale, there's no need for any normalization or for specifying a redundant 3rd `total_length` parameter.

user27182:
@pyvista-bot preview

tkoyama010 (approved):
LGTM. Thanks!

PR #6287, opened 2024-06-22
Run `pre-commit` `mypy` on project directories instead of only modified files

user27182:
### Overview

This *should* resolve #5611 and put us one step closer to #6279. See https://github.com/pyvista/pyvista/pull/6268#discussion_r1644972791 for additional context. 

After updating the config locally, `mypy` immediately reported some type errors that should have previously been addressed. This suggests the config is now working as it should.

Some of these errors are currently being resolved by other PRs, e.g. the numpy depreaction warnings and meshgrid return type in https://github.com/pyvista/pyvista/pull/6280, so I've just `# type: ignore`d these for so that CI checks pass. Some `docutils` errors were addressed by adding the type stubs as a dependency.

### Details

By default, pre-commit runs on _modified_ files only by passing the modified file names as command-line arguments. There is an escape-hatch though for this behavior by setting `pass_filenames: false` and specifying paths explicitly via `entry`. See https://github.com/pre-commit/pre-commit/issues/1519.

Any concerns about this being a slow and not recommended way of running this hook are not warranted in my view because `mypy` maintains its own cache, so once the cache is populated it runs quickly between successive calls to `pre-commit` despite running mypy on most of the project files.

EDIT: Corrected the resolved issue

PR #6291, opened 2024-06-22
DO NOT MERGE: Test pre-commit mypy config-only change

user27182:
### Overview

Similar to #6290 and other, see https://github.com/pyvista/pyvista/pull/6287#issuecomment-2184067944 for context.

pre-commit is expected to pass despite the config change.


user27182:
Pre-commit is indeed failing, so the CI checks are working as expected

PR #6292, opened 2024-06-22
Pin numpy<2.0.0 in GeoVista integration test

tkoyama010:
### Overview

- Currently, PyVista has not completed support for NumPy v2 (See #6280). Until support is completed, we will limit the version of NumPy used by GeoVista to <2.0.
- The GeoVista side will also need to be modified to support NumPy v2.0, but that should be done after the PyVista side has been modified.
- Even after PyVista's support is complete, it will take a while to make changes on the GeoVista side, so I decided to pin with this CI.

### Details

- None

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -106,7 +106,7 @@ jobs:
           mkdir -p ${CARTOPY_SHARE_DIR}
           python cartopy_feature_download.py physical --output ${CARTOPY_SHARE_DIR} --no-warn
       - name: Pin NumPy version
-        run: pip install numpy<2.0.0
+        run: pip install "numpy<2.0.0"
       - run: pytest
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -105,6 +105,8 @@ jobs:
           wget --quiet ${CARTOPY_FEATURE}
           mkdir -p ${CARTOPY_SHARE_DIR}
           python cartopy_feature_download.py physical --output ${CARTOPY_SHARE_DIR} --no-warn
+      - name: Pin NumPy version
+        run: pip install numpy<2.0.0
```
tkoyama010, line 109:
```suggestion
        run: pip install "numpy<2.0.0"
```

tkoyama010:
@pyvista-bot LGTM

PR #5699, opened 2024-02-27
Add `point_dtype` to `Line` and `LineSource` class

tkoyama010:
### Overview

Add `point_dtype` to `Line` and `LineSource` classes.

### Details

- See #5502 .

PR #5697, opened 2024-02-27
Add `point_dtype` to `Sphere` and `SphereSource` class

tkoyama010:
### Overview

Add `point_dtype` to `Sphere` and `SphereSource` classes.

### Details

- See #5502 .

PR #5696, opened 2024-02-27
Add `point_dtype` to `Cone` and `ConeSource` class

tkoyama010:
### Overview

Add `point_dtype` to `Cone` and `ConeSource` class.

### Details

- See #5502 .

PR #5502, opened 2024-01-22
Add `point_dtype` argument to function ans methods

tkoyama010:
### Overview

- [ ] vtk3DLinearGridCrinkleExtractor
- [ ] vtk3DLinearGridPlaneCutter
- [ ] vtkAdaptiveSubdivisionFilter
- [ ] vtkAppendDataSets
- [ ] vtkAppendFilter
- [ ] vtkAppendPoints
- [ ] vtkAppendPolyData
- [ ] vtkArcSource
- [ ] vtkBoundedPointSource
- [ ] vtkCapsuleSource
- [ ] vtkCleanPolyData
- [ ] vtkCleanUnstructuredGrid
- [ ] vtkClipDataSet
- [ ] vtkClipPolyData
- [ ] vtkConeSource
- [ ] vtkConnectivityFilter
- [ ] vtkConstrainedSmoothingFilter
- [ ] vtkContour3DLinearGrid
- [ ] vtkContourFilter
- [ ] vtkContourGrid
- [ ] vtkCubeSource
- [ ] vtkCutter
- [ ] vtkCylinderSource
- [ ] vtkDecimatePolylineFilter
- [ ] vtkDecimatePro
- [ ] vtkDelaunay3D
- [ ] vtkDepthImageToPointCloud
- [ ] vtkDiskSource
- [ ] vtkEllipseArcSource
- [ ] vtkEllipticalButtonSource
- [ ] vtkExtractCells
- [ ] vtkExtractCellsAlongPolyLine
- [ ] vtkFeatureEdges
- [ ] vtkFrustumSource
- [ ] vtkGeometryFilter
- [ ] vtkGlyph3D
- [ ] vtkGlyphSource2D
- [ ] vtkGridSynchronizedTemplates3D
- [ ] vtkHedgeHog
- [ ] vtkImageDataOutlineFilter
- [ ] vtkImageToPoints
- [ ] vtkLinearToQuadraticCellsFilter
- [ ] vtkLineSource
- [ ] vtkMaskPoints
- [ ] vtkMergeCells
- [ ] vtkOutlineFilter
- [ ] vtkOutlineSource
- [ ] vtkParametricFunctionSource
- [ ] vtkPlaneCutter
- [ ] vtkPlaneSource
- [ ] vtkPlatonicSolidSource
- [ ] vtkPointSource
- [ ] vtkPolyDataConnectivityFilter
- [ ] vtkPolyDataEdgeConnectivityFilter
- [ ] vtkPolyDataNormals
[...]

PR #6307, opened 2024-06-26
Raise error when only one index is given in subplot

### LINKED ISSUE #6306: Unhelfpul error message for `Plotter.subplot`

Running the code

```py
import pyvista as pv
pl = pv.Plotter(shape=(1, 2))
pl.subplot(0)
``` 

is invalid, but the error message is not helpful:

```txt
[shortened]
File /workspaces/pyvista/pyvista/plotting/renderers.py:367, in Renderers.set_active_renderer(self, index_row, index_column)
    365 if index_row < 0 or index_row >= self.shape[0]:
    366     raise IndexError(f'Row index is out of range ({self.shape[0]})')
--> 367 if index_column < 0 or index_column >= self.shape[1]:
    368     raise IndexError(f'Column index is out of range ({self.shape[1]})')
    369 self._active_index = self.loc_to_index((index_row, index_column))

TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

See above

### System Information

```shell
  Date: Wed Jun 26 13:59:45 2024 UTC

                OS : Linux
            CPU(s) : 2
           Machine : x86_64
      Architecture : 64bit
               RAM : 7.7 GiB
       Environment : IPython
        GPU Vendor : Mesa/X.org
      GPU Renderer : llvmpipe (LLVM 15.0.6, 256 bits)
       GPU Version : 4.5 (Core Profile) Mesa 22.3.6
  MathText Support : True

  Python 3.11.8 (main, Feb  7 2024, 22:38:59) [GCC 12.2.0]

           pyvista : 0.44.dev0
               vtk : 9.3.0
             numpy : 2.1.0.dev0
        matplotlib : 3.10.0.dev252+g7ccfd3813b
            scooby : 0.9.2
             pooch : 1.8.0
            pillow : 10.2.0
           imageio : 2.33.1
           IPython : 8.21.0
          colorcet : 3.0.1
           cmocean : 3.1.3
        ipywidgets : 8.1.2
              tqdm : 4.66.2
            meshio : 5.3.5
    pytest_pyvista : 0.1.8
             trame : 3.5.2
      trame_client : 2.16.0
      trame_server : 2.16.1
         trame_vtk : 2.8.4
     trame_vuetify : 2.4.2
      nest_asyncio : 1.6.0
```

### Screenshots

_No response_

tkoyama010:
### Overview

resolve #6306.

### Details

- None

### DIFF
--- a/pyvista/plotting/renderers.py
+++ b/pyvista/plotting/renderers.py
@@ -362,6 +362,8 @@ class Renderers:
             self._active_index = index_row
             return
 
+        if index_column is None:
+            raise IndexError(f'Only one index is given for ({self.shape}).')
         if index_row < 0 or index_row >= self.shape[0]:
             raise IndexError(f'Row index is out of range ({self.shape[0]})')
         if index_column < 0 or index_column >= self.shape[1]:
--- a/tests/plotting/test_plotter.py
+++ b/tests/plotting/test_plotter.py
@@ -547,3 +547,9 @@ def test_add_ruler_scale():
     min_, max_ = ruler.GetRange()
     assert min_ == 0.6
     assert max_ == 0.0
+
+
+def test_subplot_with_one_index():
+    pl = pv.Plotter(shape=(1, 2))
+    with pytest.raises(IndexError, match='Only one index is given'):
+        pl.subplot(0)


MatthewFlamm:
It isn't clear to me that this PR handles this case correctly: https://github.com/pyvista/pyvista/blob/7ae6b9a7e08efebd499b125c93e28b4aa58a064a/examples/00-load/reader.py#L122

But I haven't looked deeply.  It seems to me that the logic assumes done coherence in state.  We should at least put these conditions into this test.  That would make me more confident.

pyvista/plotting/renderers.py
```diff
@@ -362,6 +362,8 @@ def set_active_renderer(self, index_row, index_column=None):
             self._active_index = index_row
             return
 
+        if index_column is None:
+            raise IndexError(f'Only one index is given for ({self.shape}).')
```
user27182, line 366: Allowing `index_column=None` suggests that `index_column` is optional, but raising an error here suggests that it's actually required. So why even allow it to be optional / have a default value of `None`? Maybe it's better to do:
``` diff
- def set_active_renderer(self, index_row, index_column=None):
+ def set_active_renderer(self, index_row, index_column):
```
That way a `TypeError` is automatically raised, e.g. 
```TypeError: set_active_renderer() missing 1 required positional argument: 'index_column```'

Even better might be to add type annotations:

``` diff
- def set_active_renderer(self, index_row, index_column=None):
+ def set_active_renderer(self, index_row: int, index_column: int):
```

And let `mypy` find any issues with the code, e.g. invalid less-than comparisons with `None`.

user27182: I guess it's optional since the shape may be one-dimensional... But in the 1-d case no checks are made to ensure that (1) the `row_index` is not out of range and (2) the `column_index` is actually `None`.

So maybe overall the validation logic needs to be updated, maybe something like:

``` python
from pyvista.core import _validation

# validate row index
_validation.check_instance(index_row, int)
if index_row < 0 or index_row >= self.shape[0]:
    raise IndexError(f'Row index is out of range ({self.shape[0]})')

# validate column index and set active index
if len(self.shape) == 1:
    _validation.check_type(index_column, type(None))
    self._active_index = index_row
else:
    _validation.check_instance(index_column, int)
    if index_column < 0 or index_column >= self.shape[1]:
        raise IndexError(f'Column index is out of range ({self.shape[1]})')
    self._active_index = self.loc_to_index((index_row, index_column))


```

MatthewFlamm: > I guess it's optional since the shape may be one-dimensional...

This is my understanding too. However, I'm not fully confident because there aren't any failing tests with the current PR. The logic is very opaque, so to make my original comment more precise.  I suggest that we figure out the branches of logic that result in different kinds of shape for the Plotter object, and see if there are any tests for it.  Maybe there aren't and we need to add some.

I suspect your suggested code is in the right direction to fix this.

tests/plotting/test_plotter.py
```diff
@@ -547,3 +547,11 @@ def test_add_ruler_scale():
     min_, max_ = ruler.GetRange()
     assert min_ == 0.6
     assert max_ == 0.0
+
+
+def test_subplot_with_one_index():
+    pl = pv.Plotter(shape='2|1')
```
MatthewFlamm, line 553: Why do we not get an error with test now, it is curious?

banesullivan:
These points dtype issues need to be resolved. I keep running into it as well and will propose a solution soon.

Closing this due to staleness, unresolved discussions and in the same themes as:
https://github.com/pyvista/pyvista/pull/5502, https://github.com/pyvista/pyvista/pull/5696, https://github.com/pyvista/pyvista/pull/5697, https://github.com/pyvista/pyvista/pull/5698, https://github.com/pyvista/pyvista/pull/5699, https://github.com/pyvista/pyvista/pull/5700, https://github.com/pyvista/pyvista/pull/5701

user27182:
> These points dtype issues need to be resolved. I keep running into it as well and will propose a solution 

@banesullivan consider looking at #8368 for a global configurable and enforceable solution

banesullivan:
Shoot! Commented on the wrong issue. May have closed this one by mistake. Sorry y'all, been... busy going through PRs and I messed up here

PR #6311, opened 2024-06-27
Add functions to convert orientation angles to/from a 3x3 rotation matrix

user27182:
### Overview

Follow-up from #6263. Add conversion methods to facilitate transforming actors (or even datasets). 

`Prop3D` supports `orientation` to specify rotations, but users may wish to specify rotations with a rotation matrix instead. This PR adds converters to enable this.


user27182:
@pyvista-bot preview

tkoyama010 (changes requested):
Thank you. This is a very important feature. My personal wish is to have a `from_...` method in Prop3D like [Rotation class in SciPy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html). Let me know your thoughts on this design proposal. (Edit: By making it resemble major libraries like SciPy, Pandas, and Matplotlib, it will please more people in my mind.)

user27182:
> Thank you. This is a very important feature. My personal wish is to have a `from_...` method in Prop3D like [Rotation class in SciPy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html). Let me know your thoughts on this design proposal. (Edit: By making it resemble major libraries like SciPy, Pandas, and Matplotlib, it will please more people in my mind.)

I was intending to have a follow-up PR to add a new property to `Prop3D`. I had initially thought of adding `rotation` or `rotation_matrix` as a new property, and it would be a direct alternative to `orientation`, e.g. see the draft code from #6263 here:
https://github.com/pyvista/pyvista/blob/1c5ac2f46fd883e191b152ec4767700c24ee0a9d/pyvista/plotting/prop3d.py#L261-L267)

But, I wasn't sure about any of the naming or level of support to include for other transformation formats so thanks for pointing out the SciPy library. It would be best not to re-invent the wheel with these conversions, so I think it would be great to support the use of SciPy's `Rotation` objects and `from...` methods directly (with optional dependency on SciPy).

> My personal wish is to have a `from_...` method in Prop3D

What about adding a new method `rotation_from` instead of adding the `rotation_matrix` property? Something like:

``` python
from scipy.spatial.transform import Rotation
import pyvista as pv
import numpy as np


# This would be a new `Prop3D` method
# Users may specify any scipy Rotation object
def rotation_from(self, rotation: Rotation): 
    self.orientation = rotation.as_euler('yxz', degrees=True)

# Example usage of the method 
actor = pv.Actor()
matrix = [[0, -1, 0],
          [1, 0, 0],
          [0, 0, 1]]
rotation = Rotation.from_matrix(matrix)

# Note: actual usage would have method syntax: `actor.rotation_from(rotation)`
rotation_from(actor, rotation)
actor.orientation
# Out: (0.0, -0.0, 89.99999999999999)
```

This would support any arbitrary rotations, and the user would be 100% responsible for specifying the initial `Rotation.from...` method.

Using the VTK conversion methods from this PR, we could also at least support the `from_matrix` case natively without requiring the SciPy dependency.
E.g. the method could be modified as:
``` python
def rotation_from(self, rotation: Rotation | np.ndarray): 
    if isinstance(rotation, np.ndarray):
        euler_yxz = pv.rotation_matrix_to_orientation(rotation)
    else:
        euler_yxz = rotation.as_euler('yxz', degrees=True)
    self.orientation = euler_yxz
```

@tkoyama010 what are your thoughts on this kind of interface?

user27182:
Another, perhaps more general option , is to modify this PR and make `orientation_from_rotation_matrix` a private function, and instead make a new public function `rotation_as_euler_yxz` which is similar to `rotation_from` suggested above, but independent of `Prop3D`:
``` python
def rotation_as_euler_yxz(self, rotation: Rotation | np.ndarray): 
    if isinstance(rotation, np.ndarray):
        # Assume input is a 3x3 matrix by default
        # Convert with VTK
        return _rotation_matrix_to_orientation(rotation)
    else:
        # Convert with SciPy
        return rotation.as_euler('yxz', degrees=True)
```

Here, `as_euler` is fixed to use the `yxz` sequence since that's what seems to be the default and only directly supported sequence in VTK, e.g. it's what's used by `Prop3D.orientation` and `vtkTransform.GetOrientation`.

Similarly, we can make `rotation_matrix_to_orientation` from this PR private to provide the inverse function:

``` python
def rotation_as_matrix(self, rotation: Rotation | np.ndarray): 
    if isinstance(rotation, np.ndarray):
        # Assume input is euler-yxz orientation angles by default
        # Convert with VTK
        return _orientation_to_rotation_matrix(rotation)
    else:
        # Convert with SciPy
        return rotation.as_matrix()
```

Then, a separate PR would address how to integerate these directly with `Prop3D`, e.g. adding a new `rotation_from` method or `rotation_matrix` property to that class.

tkoyama010:
@user27182 
Thank you. I am satisfied if `rotation_from` is defined as a method of actor and it supports SciPy Rotation.
I will respect your wishes regarding VTK rotation support and independent function support for transformations. I am happy to review including these.

user27182:
> @user27182 Thank you. I am satisfied if `rotation_from` is defined as a method of actor and it supports SciPy Rotation. I will respect your wishes regarding VTK rotation support and independent function support for transformations. I am happy to review including these.

Great! Would you prefer to add `Prop3D.rotation_from`  to this PR or leave it for a follow-up PR?

tkoyama010:
> > @user27182 Thank you. I am satisfied if `rotation_from` is defined as a method of actor and it supports SciPy Rotation. I will respect your wishes regarding VTK rotation support and independent function support for transformations. I am happy to review including these.
> 
> Great! Would you prefer to add `Prop3D.rotation_from` to this PR or leave it for a follow-up PR?

Thanks. Please open new PR. I will review this PR if you want to merge this too.

user27182:
> I will review this PR if you want to merge this too.

The conversion functions are used by `AxesAssembly` to go from 3x3 axes vectors -> `Prop3D.orientation` and back, so let's keep them. One of the two functions is used by `rotation_from` in #6318 as well.

tkoyama010 (approved):
LGTM

PR #6317, opened 2024-06-28
Add initial implementation of `AxesAssembly`

user27182:
### Overview

Follow-up from #6263. Adds the initial (simplified) implementation of this class. It builds upon the `AxesGeometrySource` to create a `vtkPropAssembly` of actors (one actor for each xyz shaft and tip).

This initial implementation only adds color attributes. A future PR will add the labels and other features to support orientating the axes in space (similar to `Prop3D`).


user27182:
@pyvista-bot preview

tkoyama010 (approved):
LGTM. There is a problem with the display of interactive scenes, but we expect this to be updated by vtk-asm in the future.
image

PR #6318, opened 2024-06-29
Add  `Prop3D.rotation_from` method to support arbitrary rotations

user27182:
### Overview

Follow-up from #6311.


user27182:
I guess we don't have scipy as a test dependency

user27182:
Looks like np.asarray is directly supported by scipy rotation objects. It's equivalent to calling `as_matrix`. 

See https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html

But the rotations can be 3D. So, we also need to check the shape. 

This explains why there is no coverage reported for the scipy import branch. So, `validate_transform3x3` already supports scipy Rotation objects implicitly. 

We can therefore remove all scipy-specific code and `rotation_from` will just work. The only downside is that there may be a small loss of precision from the double conversion - to matrix from scipy, then to orientation by VTK. But the code will be much simpler. 

If the double conversion really is an issue, then `as_euler` can be used to set the orientation directly

user27182:
@pyvista-bot preview

tkoyama010 (approved):
LGTM. Thank you for fulfilling my wishes.

tests/plotting/test_actor.py
```diff
@@ -233,6 +234,21 @@ def test_vol_actor_prop(vol_actor):
...
 
+@pytest.mark.parametrize('func', [np.ndarray, scipy.spatial.transform.Rotation.from_matrix])
+def test_rotation_from(actor, func):
+    array = [
+        [0.78410209, -0.49240388, 0.37778609],
+        [0.52128058, 0.85286853, 0.02969559],
+        [-0.33682409, 0.17364818, 0.92541658],
+    ]
+
+    actor.rotation_from(array)
```
user27182, line 245: Whoops, this is a key part of the test and `scipy` is not used/tested otherwise.
```suggestion
    rotation = func(array)
    actor.rotation_from(rotation)
```

user27182:
> LGTM. Thank you for fulfilling my wishes.

No problem! It's a great suggestion. The use of 'yxz' euler angles specifically seems like a bit of an odd choice for `vtkProp3D` orientation, using a 3x3 matrix or any other rotation input makes more sense.

user27182:
> Looks like np.asarray is directly supported by scipy rotation objects. It's equivalent to calling `as_matrix`.
> 
> See https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html

`np.asarray` is only supported for converting SciPy `Rotation` objects to a numpy array of `Rotation` objects. So, it doesn't actually call `as_matrix`.

> This explains why there is no coverage reported for the scipy import branch. So, `validate_transform3x3` already supports scipy Rotation objects implicitly.

This is not correct. There was no coverage because the test was not set up correctly.

> We can therefore remove all scipy-specific code and `rotation_from` will just work.

Also not correct. SciPy-specific code will need to be re-introduced to make this work.

tkoyama010 (approved):

PR #6329, opened 2024-07-01
Add `Prop3D` features to `AxesAssembly` to make it orientable in space

user27182:
### Overview

Follow-up from #6263.

### Details

`AxesAssembly` has both 2D actors (labels) and 3D actors (axes shafts/tips).  This makes it challenging to position and orient it in space.

To make the assembly orientable, one option would normally be for the assembly to inherit from `vtkAssembly` directly, which itself inherits from `vtkProp3D`. But since the label actors not 3D props themselves, this is not possible. Instead, `AxesAssembly` inherits from `vtkPropAssembly` to group both 2D labels and 3D parts. This means that custom `Prop3D` features need to be implemented since they are not inherited.

This PR implements these features by using as "has-as" relationship to the `Prop3D` class instead of an "is-a" relationship. A "dummy" Actor/Prop3D instance is stored internally, and new Prop3D attributes are created which wrap this dummy actor.

When setting the Prop3D attributes for the assembly:

- The 3D actors (axes shafts/tips) are all updated by also setting their Prop3D attributes
- The 2D actors are updated by applying the transformation matrix generated by the Prop3D to the positions of the actors


user27182:
@pyvista-bot preview

tkoyama010 (approved):
LGTM

PR #6339, opened 2024-07-02
Add `symmetric_bounds` parameter to `AxesGeometrySource`

user27182:
### Overview

Follow-up from #6263. This option enables `AxesAssembly` to be used as a widget.

One of the example uses `orientation`, which is added in #6329 .

EDIT:
Implementing this feature also exposed a bug with how the parts are normalized by `AxesGeometrySource`. In the symmetric cases, the symmetry is added by mutating the part directly. But then it's not possible to undo the mutation, e.g. turn of symmetry. It also meant that updating a part multiple times would recursively make the axes parts symmetric.

So, this PR addresses this too and stores a normalized version of the part, then makes a copy when updating the geometry. This way we can always generate the output from the original normalized parts.


user27182:
@pyvista-bot preview

pyvista/core/utilities/geometric_sources.py
```diff
@@ -2957,6 +2969,68 @@ def symmetric(self) -> bool:  # numpydoc ignore=RT01
...
+        >>> axes_geometry_source.output.bounds
+        (-0.10000000149011612, 1.0, -0.10000000149011612, 1.0, -0.10000000149011612, 1.0)
+
+        >>> axes_geometry_source.output.center
+        array([0.45, 0.45, 0.45])
+
+        Show the difference in camera positioning with and without
+        symmetric bounds. Orientation is added for visualization.
+
+        >>> # Create actors
```
tkoyama010, line 3010:
```suggestion
        Create actors.
```

pyvista/core/utilities/geometric_sources.py
```diff
@@ -2957,6 +2969,68 @@ def symmetric(self) -> bool:  # numpydoc ignore=RT01
...
+        >>> axes_sym = pv.AxesAssembly(
+        ...     orientation=(90, 0, 0), symmetric_bounds=True
+        ... )
+        >>> axes_asym = pv.AxesAssembly(
+        ...     orientation=(90, 0, 0), symmetric_bounds=False
+        ... )
+        >>>
+        >>> # Show multi-window plot
```
tkoyama010, line 3018:
```suggestion

        Show multi-window plot.
```

tkoyama010 (changes requested):

tkoyama010:
@pyvista-bot preview

tkoyama010 (approved):
LGTM. Thanks!

PR #6341, opened 2024-07-03
Add `AxesAssemblySymmetric` class to allow the use of orientable symmetric axes

user27182:
### Overview

This is an extension of #6263. A new `AxesAssemblySymmetric` class is subclassed from `AxesAssembly`. The new symmetric class accepts all the same arguments as its parent, except it overrides the text label setters and getters. Specifically, it allows setting two labels for each axis instead of one. Some of the mechanics for how the transformations are applied to the text labels had to be refactored to make this work.

This new class can be viewed as a possible replacement for https://docs.pyvista.org/version/stable/api/plotting/_autosummary/pyvista.Axes.html#pyvista.Axes which has very limited features.

### Details

Initially this PR started off by extending `AxesAssembly` and simply adding a new `symmetric` property and passing that to `AxesGeometrySource` (which also has a `symmetric` property). But, the use of labels in `AxesAssembly` complicates things, because we would need to:

- add separate params for setting/getting the negative labels
- conditionally adding a `'+'` character to the positive labels if `symmetric is True`
- conditionally setting `show_labels` to turn on/off the symmetric labels, etc

In the end it was easier to implement this by making `AxesAssembly` exclusively asymmetric and making `AxesAssemblySymmetric` exclusively symmetric.


pyvista/plotting/axes_assembly.py
```diff
@@ -303,9 +307,8 @@ def show_labels(self) -> bool:  # numpydoc ignore=RT01
     @show_labels.setter
     def show_labels(self, value: bool):  # numpydoc ignore=GL08
         self._show_labels = value
-        self._label_actors[0].SetVisibility(value)
-        self._label_actors[1].SetVisibility(value)
-        self._label_actors[2].SetVisibility(value)
+        for label in self._label_actor_iterator:
+            label.SetVisibility(value)
```
tkoyama010, line 311: This may be a different PR, but is it possible to set the value for each direction using a list?

pyvista/plotting/axes_assembly.py
```diff
@@ -372,9 +374,8 @@ def label_color(self) -> Color:  # numpydoc ignore=RT01
     def label_color(self, color: ColorLike):  # numpydoc ignore=GL08
         valid_color = Color(color)
         self._label_color = valid_color
-        self._label_actors[0].prop.color = valid_color
-        self._label_actors[1].prop.color = valid_color
-        self._label_actors[2].prop.color = valid_color
+        for label in self._label_actor_iterator:
+            label.prop.color = valid_color
```
tkoyama010, line 378: Same with `value` in `show_labels` .

pyvista/plotting/axes_assembly.py
```diff
@@ -430,15 +431,27 @@ def _get_transformed_label_positions(self):
...
-        self._label_actors[0].position = x_pos
-        self._label_actors[1].position = y_pos
-        self._label_actors[2].position = z_pos
+    def _apply_transformation_to_labels(
+        self, position_scalars: tuple[float, float, float], labels: tuple[Label, Label, Label]
+    ):
+        x_vector, y_vector, z_vector = self._transform_label_position(position_scalars)
+        labels[0].position = x_vector
+        labels[1].position = y_vector
+        labels[2].position = z_vector
```
tkoyama010, line 440: Let's use for loop.
```suggestion
        vectors = self._transform_label_position(position_scalars)
        for label, vector in zip(labels, vectors):
            label.position = vector
```

pyvista/plotting/axes_assembly.py
```diff
@@ -318,9 +321,8 @@ def label_size(self) -> int:  # numpydoc ignore=RT01
     @label_size.setter
     def label_size(self, size: int):  # numpydoc ignore=GL08
         self._label_size = size
-        self._label_actors[0].size = size
-        self._label_actors[1].size = size
-        self._label_actors[2].size = size
+        for label in self._label_actor_iterator:
+            label.size = size
```
tkoyama010, line 325: Same with `value` in `show_labels` .

tkoyama010 (changes requested):

pyvista/plotting/axes_assembly.py
user27182: Do you mean you want to be able to control visibility of each label separately? Instead of all having the same visibility?

We can make the change. Probably best for a separate PR for the full feature. But, the changes in this PR make use of `self._label_actor_iterator` to abstract some of the label setter methods to the superclass, so that the method would apply to any/all labels, regardless if there are 3 (AxesAssembly) or 6 (AxesAssemblySymmetric). Because of this abstraction we lose fine-tuned control over each label actor.

Some options:

1. We could revert this abstraction, add a concrete implementation for the symmetric subclass, and modify the separate implementations of `show_labels`, `label_size`, etc. to allow setting the individual text properties.

2. An alternative, which is currently implemented in `AxesActor`, is to expose the separate actor props to the user, e.g. `AxesAssembly.x_label_prop`, `AxesAssembly.y_label_prop`, etc. But I'm personally not a fan of this, especially for the symmetric case since you need 6 lines of code to set all 6 labels. Plus there are `Text` attributes which we may to modify (e.g. visibility), but also `TextProperty` attributes, so this solution doesn't make it easy to set both kinds of attributes.

3. In the current implementation for #5029, I had added generic `set_prop` and `get_prop` methods (see the [code](https://github.com/pyvista/pyvista/blob/930fcfbf578ec3b45ac497199c94a21ea4ee1327/pyvista/plotting/axes_actor.py#L935-L1033)), allowing setting/getting actor properties on a part-basis (e.g. all shafts or all tips) or an axis basis (all x parts, all y parts, or all z parts). This was for the 3D actor properties, but we could add something like for the text label properties too, and do `try... except` on both the `Text` and `TextProperty` to make it simple to use.

user27182: Just to add to this, options 2 and 3 aim to be generic to allow setting _any_ actor attribute, not just `size` or `visibility`. Here are some concrete examples... If we want to make the positive labels visibly but not the negative ones, for option 1 (your suggestion) this would maybe look like:
```python
assembly.show_labels = [True, False, True, False, True, False]
```
For option 3, we could keep the current implementation for simplicity, e.g. 
```python
assembly.show_labels = True
```
But then allow something like:
```python
assembly.set_label_prop('visibility', True, axis='positive')
```
This would also allow setting the justification, bold/italics, text shadow... etc., e.g. to center-justify the text for all labels:
```python
assembly.set_label_prop('justification_horizontal', 'center')
```

tkoyama010: Thank you. It was just a passing thought, so if it has that much of an impact, there's no need to do it. Let's leave it as it is and respond if there is a strong request.

tkoyama010 (approved):
LGTM

pyvista/plotting/axes_assembly.py
user27182: I now see the value in having this feature. If plotting with any plane view, it would be nice to hide the in-plane and out-of plane labels, e.g. hide `z` labels for `view_xy`. See the unsightly and unreadable stacked labels in the center of this widget:

Screenshot 2024-08-15 at 12 52 39 AM

PR #6342, opened 2024-07-04
Add `set_actor_prop` and `get_actor_prop` to `AxesAssembly` for setting/getting part properties

user27182:
### Overview

Convenience methods for setting `Property` attributes of the axes parts. 

See the discussion here https://github.com/pyvista/pyvista/pull/6341#discussion_r1664920288 for context. That discussion initially suggests adding a similar method but for the axes labels, e.g. `set_label_prop` and `get_label_prop`. But here, the methods are first implemented for the parts, and a separate PR can be made for setting/getting label properties.


user27182:
@pyvista-bot preview

pyvista/plotting/axes_assembly.py
```diff
@@ -465,43 +478,212 @@ def label_color(self, color: ColorLike):  # numpydoc ignore=GL08
...
+        Examples
+        --------
+        Set :attr:`~pyvista.Property.ambient` for all axes shafts and tips to a
+        single value.
+
+        >>> import pyvista as pv
+        >>> axes_actor = pv.AxesAssembly()
+        >>> axes_actor.set_part_prop('ambient', 0.7)
+        >>> axes_actor.get_part_prop('ambient')
+        _AxesPropTuple(x_shaft=0.7, y_shaft=0.7, z_shaft=0.7, x_tip=0.7, y_tip=0.7, z_tip=0.7)
```
user27182, line 556: I am not sure I like the NamedTuple here. On the one hand, the named values are helpful since it's not obvious that the order should be `shaft-xyz` then `tip-xyz`. The order could be especially confusing since the symmetric `labels` param has the axes "grouped"  together as `(+x. -x, +y, -y, +z, -z)`, so both of these 6-tuples use different ordering conventions.

But, the named tuple  also makes it hard to read the values and compare them, since the text gets in the way and is cluttering things a bit.

Should we get rid of it?

pyvista/plotting/axes_assembly.py
```diff
@@ -465,43 +478,212 @@ def label_color(self, color: ColorLike):  # numpydoc ignore=GL08
...
+        if len(values) != len(actors):
+            raise ValueError(
+                f"Number of values ({len(values)}) in {value} must match the number of parts ({len(actors)}) for axis '{axis}' and part '{part}'"
+            )
+
+        # Sequence is valid, now set values
+        for actor, val in zip(actors, values):
+            setattr(actor.prop, name, val)
+
+    def get_part_prop(self, name: str):
```
user27182, line 632: I've toyed with adding an `'axis'` and `'part'` parameter to the getter too. This would make this use case a bit nicer:
```  diff
- self.get_part_prop('color')[_AxisEnum.z.value :: 3]
+ self.get_part_prop('color', axis='z')

 But then the output could be a scalar, or a length 2, 3, or 6 tuple, which could result in index errors if user's aren't careful with how they use the arguments/outputs. This could maybe be addressed by overloading the typing to hint the output type for all cases...

Should we add these parameters, or keep the status quo and always return a 6-tuple?

File: pyvista/plotting/axes_assembly.py [hunk 3]
```diff
@@ -465,43 +478,212 @@ def label_color(self, color: ColorLike):  # numpydoc ignore=GL08
...
+        _AxesPropTuple(x_shaft=0.0, y_shaft=0.0, z_shaft=0.0, x_tip=0.0, y_tip=0.0, z_tip=0.0)
+
+        """
+        prop_values = [getattr(actor.prop, name) for actor in self._shaft_and_tip_actors]
+        return _AxesPropTuple(*prop_values)
+
+    def _filter_part_actors(
+        self,
+        axis: Literal['x', 'y', 'z', 0, 1, 2, 'all'] = 'all',
+        part: Literal['shaft', 'tip', 0, 1, 'all'] = 'all',
```
user27182, line 665: I'm thinking we should remove the use of integers from the type annotations and the docstring. They are used interally, so we can support it unofficially, but otherwise they are like magic numbers. And I don't think we should expose the enums either.

user27182:
@pyvista-bot preview

pyvista/plotting/axes_assembly.py
tkoyama010: As a reviewer, I didn't find it difficult to read. `NamedTuple` is a good way to organize information, so I'm in favor of using it. +1 keep it as it is. If users complain about any inconveniences they experience when using the site, let's reconsider it.

tkoyama010 (approved):
LGTM. If there are no additional comments, it will be merged in 24 hours.

pyvista/plotting/axes_assembly.py
tkoyama010: When specifying parameters, I think the user will be aware of the size, so I don't think usability will be affected even if we change the size.
We are looking forward to your new PR.

Issue #555, opened 2020-01-21
Refactor ObbTree and add examples


banesullivan:
Obb trees are super useful and interesting. We should refactor the `PolyData.obbTree` property to `obb_tree` and add some helper methods to actually be able to use that obb tree for intersection testing, cool visualizations, and more. Here's an example that isn't very intuitive as the user have to pass an empty PolyData method to an obscure method on the `vtkObbTree` object:

```py
import pyvista as pv
from pyvista import examples

shark = examples.download_shark()

boxes = pv.PolyData()
tree = shark.obbTree
tree.GenerateRepresentation(3, boxes)

p = pv.Plotter()
p.add_mesh(shark,)
p.add_mesh(boxes, show_edges=True, opacity=0.25, color="green")
p.camera_position = [(-117.43896518622428, 76.33874007163854, 235.18626454074305),
 (1.6917457580566406, -1.3803672790527344, -0.004845619201660156),
 (0.18971104936049324, 0.9568522451433392, -0.2200988385128178)]
p.show()
```

![download](https://user-images.githubusercontent.com/22067021/72826531-44e54c80-3c36-11ea-9b16-c7de3de0bed8.png)

user27182:
It's often helpful to get a 0-level 3D oriented bounding box without the recursion (i.e. `tree.GenerateRepresentation(0, boxes)`. So as part this refactor perhaps we can add a method, e.g. `obb`, (without `_tree`) as a convenience.

In the examples, we can compare the `obb` output to the output from `fit_plane_to_points` once #4650 is resolved.
Since `fit_plane_to_points` uses SVD and `obbTree` uses covariance/PCA, the "fit" or orientation can differ.

For example, the alignment between the two methods is better for shark than cow (see code below):
![image](https://github.com/pyvista/pyvista/assets/89109579/1ed6d411-ccef-4473-8af0-73dce9c8e54d)
![image](https://github.com/pyvista/pyvista/assets/89109579/2dad6bfc-68ce-404f-a391-5324d6c30f00)

``` python
import numpy as np
import pyvista as pv
from pyvista import examples
def fit_plane_to_points(points, return_meta=False):
    # Apply SVD to get orthogonal basis vectors to define the plane
    data = np.array(points)
    data_center = data.mean(axis=0)
    U, S, Vh = np.linalg.svd(data - data_center)
    i_vector = Vh[0]
    j_vector = Vh[1]
    normal = np.cross(i_vector, j_vector)

[...]

PR #6346, opened 2024-07-05
Add `principal_axes` function to compute best-fit axes vectors from points

user27182:
### Overview

Follow-up from #4947. This PR adds an initial simplified version of this function from that PR.


user27182:
@pyvista-bot preview

pyvista/core/utilities/points.py
```diff
@@ -459,3 +465,112 @@ def vector_poly_data(orig, vec):
...
+    def _has_nan(array: NumpyArray[float]) -> np.bool_:
+        return np.any(np.isnan(array))
+
+    # Compute OBB
+    corner = _init_vector()
+    max_ = _init_vector()
+    mid_ = _init_vector()
+    min_ = _init_vector()
+    size = _init_vector()
+    _vtk.vtkOBBTree().ComputeOBB(vtk_points(points, deep=False), corner, max_, mid_, min_, size)
```
adeak, line 552: I was a bit skeptical at first because it's not documented that OBB uses a proper PCA, and the axes of the OBB (naively, in my head) could differ from the principal ellipsoid.

So I threw together a quick check with a manual NumPy transformation. On one hand I get _the exact same_ axes as your helper (to all digits of the printed array, so I assume it's the exact same numbers modulo sign). On the other hand, the NumPy version is about 2x faster.

Here's the implementation, which of course doesn't handle degenerate cases and nans and whatnot:
```python
import numpy as np

def principal_axes_numpy(points):
    points_centered = points - points.mean(0)
    np_eigvals, np_axes = np.linalg.eigh(points_centered.T @ points_centered)
    np_sizes = np.sqrt(np_eigvals)
    np_axes = np_axes.T  # columns -> rows
    sort_inds = (-np_sizes).argsort()
    np_axes = np_axes[sort_inds, :]
    return np_axes
```

So, do you think we're better off with offloading the work to VTK? My hunch is that for the non-singular case we'll already have right-handed orthonormal vectors. Less sure about singular cases, so we'd probably have to do similar cross product tricks. Haven't thought that much ahead.

pyvista/core/utilities/points.py
```diff
@@ -459,3 +465,112 @@ def vector_poly_data(orig, vec):
...
+
+    Parameters
+    ----------
+    points : MatrixLike[float]
+        Nx3 array of points. At least two points are required.
+
+    Returns
+    -------
+    numpy.ndarray
+        3x3 orthonormal array with the principal axes as row vectors.
```
adeak, line 502: Do you think it would also be interesting to (optionally??) return the sizes? We already have them.

tests/core/test_utilities.py
```diff
@@ -913,6 +914,53 @@ def test_fit_plane_to_points():
...
+    ],
+)
+PLANAR_CASE = (
+    [[0, 0, 0], [1, 1, 1]],
+    [
+        [0.57735027, 0.57735027, 0.57735027],
+        [0.0, -0.70710678, 0.70710678],
+        [0.81649658, -0.40824829, -0.40824829],
+    ],
+)
```
adeak, line 932: This is a rank 1, _linear_ case, right? So I'd call it `LINEAR_CASE`, and use a 3-point `PLANAR_CASE`. Assuming I understand the intent here correctly.

pyvista/core/utilities/points.py
adeak: Ah, it _is_ documented, but [in the main section of the documentation](https://vtk.org/doc/nightly/html/classvtkOBBTree.html#details) rather than in `ComputeOBB()` (which makes sense in retrospect):

> First, the root OBB is constructed by finding the mean and covariance matrix of the cells (and their points) that define the dataset. The eigenvectors of the covariance matrix are extracted, giving a set of three orthogonal vectors that define the tightest-fitting OBB.

tests/core/test_utilities.py
user27182: Yes you are right it's two points (line) not a plane

pyvista/core/utilities/points.py
user27182: This OBB implementation with VTK is a refactored version from what was originally implemented in #4947, which used numpy and SVD originally (and the SVD uses LAPACK under the hood) 
In that implementation the SVD values were also extracted. In my testing with this VTK version I was finding there wasn't a good correlation between the SVD values from numpy and the eigen values (sizes) from this VTK version (im not an expert on PCA and SVD but I thought they would correlate by a sqrt factor).

I looked at the VTK source code and it uses vtkMathJacobi for the computation, which returns the vectors and size. But those are returned as column vectors. So I'm just not sure if the sizes actually correlate correctly with the vectors. Like, what do they mean? Are they the eigen values? So I don't have confidence about what the `size` represents

adeak: Good point. The documentation says
> Also return a sorted list of relative "sizes" of axes for comparison purposes. 

which doesn't spark confidence.

pyvista/core/utilities/points.py
user27182: Good find with the documentation, I hadn't actually read it, but saw that it computes the eigenvalues from the covariance matrix of mean-centred points in the source code. 

From a performance perspective the main issue I saw with the original SVD implementation in #4947 is that for even not too large meshes (e.g. points numbering in the thousands) I keep running into MemoryErrors with SVD. This requires adding a sampling support function to reduce the memory requirements. I also found the numpy SVD implementation was hard coded to use doubles instead of singles, making the memory footprint worse (I even had a hacky single version implemented which used private numpy functions). 

So, the memory issue was a big issue. I'm not sure if this would still be an issue using an eigen vector implementation instead, as you suggest. So, maybe needs more testing? I can probably add the same memory test I had previously as part of #4947. 

We could also try using numpy first, and use VTK as backup if there a memory error is raised.

adeak: With the above code we always compute a `(3 x N) * (N x 3)` matrix product, which gives us a `(3, 3)`-shaped real symmetric (and consequently Hermitian) matrix. And the first matrix is the transpose of the second matrix, so I think we have a memory footprint of only literally dozens of float values. The cost is matrix multiplication (CPU time). For small matrices we might be able to be faster by doing some smart copies of transposes, to help caching on the CPU. Probably not worth the effort.

Oh, and the fact that we can use a Hermitian solver might actually mean that we readily have an orthonormal system! Of course the orientation of the vectors in rank-deficient cases is probably a different kind of arbitrary than what we have right now...

adeak: Minor correction: we need to center the input array so we double the memory used.

adeak: For `100_000_000` points (2 GB of floats) it takes 2 seconds to compute the centered array and 0.5 more seconds to compute the 3x3 matrix:
```
In [1]: import numpy as np
   ...: 
   ...: arr = np.random.rand(100_000_000, 3) * [2, 3, 4]
   ...: %timeit arr_centered = arr - arr.mean(0)
2.02 s ± 3.59 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [2]: import numpy as np
   ...: 
   ...: arr = np.random.rand(100_000_000, 3) * [2, 3, 4]
   ...: %timeit arr_centered = arr - arr.mean(0); X = arr_centered.T @ arr_centered
2.49 s ± 8.38 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

adeak: Sorry, last message for now.

We can avoid computing the centered array (that doubles the memory footprint) by breaking out the `(points - points_mean).T @ (points - points_mean)` formula:
```
In [1]: import numpy as np
   ...: 
   ...: arr = np.random.rand(100_000_000, 3) * [2, 3, 4]
   ...: %timeit arr_sum = arr.sum(0); arr_mean = arr_sum / arr.shape[0]; X = arr_mean[:, None] * arr_mean * arr.shape[0] - arr_mean[:, None] * arr_sum - arr_sum[:, None] * arr_mean + arr.T @ arr
1.75 s ± 2.45 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Sorry for the one-liner, that's how I could quickly squeeze this into a `%timeit` (`%%timeit` wouldn't work and no time to figure this out now). We just compute 4 product terms by expanding the parantheses involved in centering the array. We can probably just compute the middle terms once and take the transpose to make this even more efficient. But this is already both faster _and_ more memory efficient than the naive approach.

So I think this is viable.

user27182: > I was a bit skeptical at first because it's not documented that OBB uses a proper PCA, and the axes of the OBB (naively, in my head) could differ from the principal ellipsoid.
> 
> So I threw together a quick check with a manual NumPy transformation. On one hand I get _the exact same_ axes as your helper (to all digits of the printed array, so I assume it's the exact same numbers modulo sign). On the other hand, the NumPy version is about 2x faster.
> 
> Here's the implementation, which of course doesn't handle degenerate cases and nans and whatnot:
> 
> ```python
> import numpy as np
> 
> def principal_axes_numpy(points):
>     points_centered = points - points.mean(0)
>     np_eigvals, np_axes = np.linalg.eigh(points_centered.T @ points_centered)
>     np_sizes = np.sqrt(np_eigvals)  # sqrt not actually needed unless we want to return these too
>     np_axes = np_axes.T  # columns -> rows
>     sort_inds = (-np_sizes).argsort()
>     np_axes = np_axes[sort_inds, :]
>     return np_axes
> ```
> 
> So, do you think we're better off with offloading the work to VTK? My hunch is that for the non-singular case we'll already have right-handed orthonormal vectors. Less sure about singular cases, so we'd probably have to do similar cross product tricks. Haven't thought that much ahead.

For completeness here's what was used previously in #4947 which was removed in https://github.com/pyvista/pyvista/pull/4947/commits/55a98b75ea64344cf8645e066224eb9715ebc786 (and replaced with the VTK OBB method instead):
``` python
# Center data
centroid = np.mean(data, axis=0)
data -= centroid

# Use SVD as it's numerically more stable than using PCA (below)
_, axes_values, axes_vectors = np.linalg.svd(data)  # row vectors, descending order

## Equivalently (up to a difference in axis sign, non-uniqueness of
## vectors, and numerical error), the axes may also be computed using
## PCA, i.e. using covariance and eigenvalue decomposition.
# covariance = np.cov(data, rowvar=False)
# _, axes_vectors = np.linalg.eigh(covariance)  # column vectors, ascending order
# axes_vectors = axes_vectors.T[::-1]  # row vectors, descending order
```

Other than small differences with the covariance computation and sorting, the commented out PCA method from that PR and your suggested PCA method is the same. (I also had a note about SVD being more numerically stable than PCA but that's kind of irrelevant if the SVD version is inefficient memory-wise.)

So, it looks like we have three ways to do the calc:

1. NumPy SVD
2. NumPy PCA
3. VTK PCA via vtkOBBTree

But NumPy SVD appears to have a large memory footprint, and vtkOBBTree is slow compared to Numpy PCA (and has a questionable `sizes` output). So, shall we go ahead and use the PCA implementation then? And also use the suggest points centering computation?

> We can avoid computing the centered array (that doubles the memory footprint) by breaking out the `(points - points_mean).T @ (points - points_mean)` formula:
> 
> ```
> In [1]: import numpy as np
>    ...: 
>    ...: arr = np.random.rand(100_000_000, 3) * [2, 3, 4]
>    ...: %timeit arr_sum = arr.sum(0); arr_mean = arr_sum / arr.shape[0]; X = arr_mean[:, None] * arr_mean * arr.shape[0] - arr_mean[:, None] * arr_sum - arr_sum[:, None] * arr_mean + arr.T @ arr
> 1.75 s ± 2.45 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
> ```

user27182: > I was a bit skeptical at first because it's not documented that OBB uses a proper PCA, and the axes of the OBB (naively, in my head) could differ from the principal ellipsoid.

There is good reason to suspect `vtkOBBTree` will produce different results. Most of the methods in that class seem to use cells, not points, for the computation, which can affect the output. See the generated best-fit plane (using numpy SVD in `fit_plane_to_points`) vs the generated OBB in this discussion https://github.com/pyvista/pyvista/issues/555#issuecomment-1646108382 for examples. I had to specifically go looking for a points-only implementation in `vtkOBBTree`.

In a future PR I am considering adding an `oriented_bound_box` feature, so this will be relevant as we'll actually want to _avoid_ using vtkOBBTree for this task, and instead use this `principal_axes` function under-the-hood.

user27182: > We can avoid computing the centered array (that doubles the memory footprint) by breaking out the `(points - points_mean).T @ (points - points_mean)` formula:
> 
> ```
> In [1]: import numpy as np
>    ...: 
>    ...: arr = np.random.rand(100_000_000, 3) * [2, 3, 4]
>    ...: %timeit arr_sum = arr.sum(0); arr_mean = arr_sum / arr.shape[0]; X = arr_mean[:, None] * arr_mean * arr.shape[0] - arr_mean[:, None] * arr_sum - arr_sum[:, None] * arr_mean + arr.T @ arr
> 1.75 s ± 2.45 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
> ```

@adeak I'm not entirely sure how to modify the naive implementation to make use of this calculation... can you please make a suggested change to incorporate this (or post a code snippet with more context code, e.g. use this output with `np.linalg.eigh`) ?

pyvista/core/utilities/points.py
user27182: The implementation has changed to use numpy, and so the `sizes` are now optionally returned as suggested in https://github.com/pyvista/pyvista/pull/6346/commits/566c6ba1a403a6f7ef8d33a20ea66371b04b21ae

pyvista/core/utilities/points.py
adeak: I'll probably be able to do this Europe-tonight. But in the meantime I've realised that the above alternative can't be used with huge means compared to the origin. I'll discuss this too.

tests/core/test_utilities.py
```diff
@@ -913,6 +914,81 @@ def test_fit_plane_to_points():
...
+
+    axes = principal_axes(points)
+    assert np.allclose(axes, expected_axes, atol=1e-7)
+
+    assert np.array_equal(np.cross(axes[0], axes[1]), axes[2])
+    assert np.allclose(np.linalg.norm(axes[0]), 1)
+    assert np.allclose(np.linalg.norm(axes[1]), 1)
+    assert np.allclose(np.linalg.norm(axes[2]), 1)
+
+    assert type(axes) is np.ndarray
```
adeak, line 966: Are you doing exact type checks on purpose? Or can we use
```suggestion
    assert isinstance(axes, np.ndarray)
```
? I see that `np.linalg.eigh()` nicely returns a `pyvista_ndarray()` for input of the same type, so I assume this is on purpose, but then a comment might help future readers.

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,91 @@ def vector_poly_data(orig, vec):
...
+    eig_vals, eig_vectors = np.linalg.eigh(points_centered.T @ points_centered)
+    sizes = np.sqrt(eig_vals)[::-1]  # ascending order -> descending order
+    axes = eig_vectors.T[::-1]  # columns, ascending order -> rows, descending order
+
+    # Normalize and ensure output forms right-handed coordinate frame
+    axes[0] /= np.linalg.norm(axes[0])
+    axes[1] /= np.linalg.norm(axes[1])
+    axes[2] = np.cross(axes[0], axes[1])
+    if type(axes) is not np.ndarray:
+        axes = np.array(axes, dtype=axes.dtype)
```
adeak, line 550: Why not
```suggestion
    axes = np.asarray(axes)
```
?

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,91 @@ def vector_poly_data(orig, vec):
...
+
+    points_centered = points - np.mean(points, axis=0)
+    eig_vals, eig_vectors = np.linalg.eigh(points_centered.T @ points_centered)
+    sizes = np.sqrt(eig_vals)[::-1]  # ascending order -> descending order
+    axes = eig_vectors.T[::-1]  # columns, ascending order -> rows, descending order
+
+    # Normalize and ensure output forms right-handed coordinate frame
+    axes[0] /= np.linalg.norm(axes[0])
+    axes[1] /= np.linalg.norm(axes[1])
+    axes[2] = np.cross(axes[0], axes[1])
```
adeak, line 548: According [to the docs](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigh.html#numpy.linalg.eigh) the vectors are orthonormal. With that in mind I think we can skip normalization. Unfortunately neither the NumPy docs [nor that of LAPACK](https://netlib.org/lapack/explore-html/d8/d30/group__heevd_ga3cac18a55c8b7b87a42cfa26c1f3499a.html) mentions handedness. But I'd assume it's always right-handed, and if we still want to verify this we could also compute `np.linalg.det(axes)`. If it's negative, we have a left-handed system and we should flip the sign of one of the vectors (last one?). In that case we don't touch the vectors in most expected (right-handed) cases, and when we do touch them we only introduce sign changes rather than new floating-point errors (not that I'd expect these to be significant in any way).

tests/core/test_utilities.py
```diff
@@ -913,6 +914,81 @@ def test_fit_plane_to_points():
...
+
+    _, sizes = principal_axes(points, return_sizes=True)
+    assert sizes[0] >= sizes[1]
+    assert sizes[1] >= sizes[2]
+    assert type(sizes) is np.ndarray
```
adeak, line 971: I recommend also adding a dedicated `return_sizes` test where the input are points from an ellipsoid (shifted away from the origin for good measure). If our code works then the sizes should be `np.allclose()` to the ellipsoid's corresponding parameters.

tests/core/test_utilities.py
```diff
@@ -913,6 +914,81 @@ def test_fit_plane_to_points():
     )
 
 
+DEFAULT_PRINCIPAL_AXES = [[0.0, 0.0, 1.0], [0.0, 1.0, 0.0], [-1.0, 0.0, 0.0]]
```
adeak, line 917: Why is `[z, y, -x]` the default? Is this just what `eigh()` does?

tests/core/test_utilities.py
```diff
@@ -913,6 +914,81 @@ def test_fit_plane_to_points():
...
+    axes = principal_axes([1, 2, 3])
+    assert np.allclose(axes, DEFAULT_PRINCIPAL_AXES)
+
+
+def test_principal_axes_vectors_many_points():
+    n_points = 2_000_000
+    points = np.random.default_rng().random((n_points, 3))
+    axes = pv.principal_axes(points)
+    assert np.any(axes)
+    assert np.all(axes != 0)
```
adeak, line 989: 1. What does this test exactly? Just that there's no error for "many" points? Note that `np.all(axes != 0)` implies `np.any(axes)`. And for any kind of random float data `!= 0` will almost certainly be true. This is why I'm not sure what you're confirming here.
2. Mostly a style note, but since we ensure we have `ndarray`s here, I'd go with `axes.any()`/`axes.all()` instance methods.

tests/core/test_utilities.py
user27182: Yes this is just what it puts out by default. I can update the name to make this clear?

I do have intentions for a future PR to flip / project / reorder the vectors as needed or where possible to make them better line up with the positive XYZ axes by default though. E.g. such that the default output would be the identity matrix in this case.

tests/core/test_utilities.py
user27182: We can keep the default behaviour and keep the `T` -> `T` type mapping  (and thus return `pyvista_ndarray` if we have `pyvista_ndarray` in.

tests/core/test_utilities.py
adeak: > Yes this is just what it puts out by default. I can update the name to make this clear?

Since this is just a test this is not too critical to me (comment can also work before the definition, but again, not critical).

> I do have intentions for a future PR to flip / project / reorder the vectors as needed or where possible to make them better line up with the positive XYZ axes by default though. E.g. such that the default output would be the identity matrix in this case.

OK, but that will interact with the current API where vectors are sorted by decreasing size, right? You don't have the freedom to change anything beyond changing 2 vector signs, unless there are degenerate axes, but we're talking about floats so telling "degenerate" is a bit messy issue.

tests/core/test_utilities.py
adeak: > We can keep the default behaviour and keep the `T` -> `T` type mapping (and thus return `pyvista_ndarray` if we have `pyvista_ndarray` in.

But returning `ndarray` for `list` or other compatible sequence input might throw a wrench in that...

pyvista/core/utilities/points.py
user27182: > According [to the docs](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigh.html#numpy.linalg.eigh) the vectors are orthonormal. With that in mind I think we can skip normalization. 

Might have been left over from the vtk implementation. The unit vector tests were failing at one point but I'll remove the normalization. Tests are passing without it.

> Unfortunately neither the NumPy docs [nor that of LAPACK](https://netlib.org/lapack/explore-html/d8/d30/group__heevd_ga3cac18a55c8b7b87a42cfa26c1f3499a.html) mentions handedness. But I'd assume it's always right-handed, 

Several of the tests fail when I comment out the cross-product, so this assumption can't be made.

> and if we still want to verify this we could also compute `np.linalg.det(axes)`. If it's negative, we have a left-handed system and we should flip the sign of one of the vectors (last one?). In that case we don't touch the vectors in most expected (right-handed) cases, and when we do touch them we only introduce sign changes rather than new floating-point errors (not that I'd expect these to be significant in any way).

I'm fine with this. Related to an earlier note about the numerical stability of SVD vs PCA, it does seem like there are precision issues with this method, Not huge, but I did have to tweak the tests to use "nicer" values, otherwise in at least one case I had to use `atol=1e-2` to make it pass. This was mostly an issue with values close to zero, and for low-rank inputs, and the error was very machine dependent. But you will still see `atol=1e-7` in the tests for this reason.

With that in mind, re-computing the third vector may help reduce the error in some cases. But, I think it's ok without this, and only change the sign if needed.

pyvista/core/utilities/points.py
user27182: Yup way better, I even went looking at the numpy docs to find this.

But, I think we can probably remove this altogether, and return `pyvista_ndarray` if we have `pyvista_ndarray` in as suggested elsewhere.

tests/core/test_utilities.py
user27182: > 1. What does this test exactly? Just that there's no error for "many" points? 

Yes, this is because a similar test using `np.linalg.svd` instead of the PCA method fails with numpy raising array memory errors. So I wanted to make sure this is no longer the case.

> Note that `np.all(axes != 0)` implies `np.any(axes)`. And for any kind of random float data `!= 0` will almost certainly be true. This is why I'm not sure what you're confirming here.

This part isn't really necessary, but I was intending to make sure that the output wasn't simply the default output, i.e. checking that the computation didn't just silently fail and return zeros and ones.

pyvista/core/utilities/points.py
user27182: This kind of optimization seems helpful, I haven't done any real performance testing though so I am not entirely sure how big of a issue this is. Should we leave this kind of change for a future PR then? There shouldn't be another release soon, so changes near-term shouldn't require any deprecations, and we're free to revert to the OBB method if needed (though its `sizes` output is still a concern..).

tests/core/test_utilities.py
user27182: Ah yes, I guess I meant we can always return an instance of `np.ndarray`, and therefore allow for things like `pyvista_ndarray` -> `pyvista_ndarray`. Basically just remove any type casting from the output.

pyvista/core/utilities/points.py
adeak: > This kind of optimization seems helpful, I haven't done any real performance testing though so I am not entirely sure how big of a issue this is. Should we leave this kind of change for a future PR then? There shouldn't be another release soon, so changes near-term shouldn't require any deprecations, and we're free to revert to the OBB method if needed (though its `sizes` output is still a concern..).

Good point, we can also go with the naive approach for now. I recommend noting in the docstring that the implementation uses a temporary array of the same size as the input array. Then supporting high-memory cases is a future feature.

tests/core/test_utilities.py
adeak: Hmm, then perhaps you could use this test to check the `size`s? Not sure if it's an issue to mix test semantics. You could use a dense ellipsoid with "many" points and check the resulting sizes while you're at it. Might be a bad idea, I'm not sure.

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,90 @@ def vector_poly_data(orig, vec):
...
+    .. versionadded:: 0.45.0
+
+    Parameters
+    ----------
+    points : MatrixLike[float]
+        Nx3 array of points.
+
+    return_sizes : bool, default: False
+        If ``True``, also returns the axes sizes. The sizes are computed as the square
+        root of the eigen vectors of the mean-centered covariance matrix.
```
adeak, line 499:
```suggestion
        If ``True``, also returns the axis sizes. The sizes are computed as the square
        root of the eigenvectors of the mean-centered covariance matrix.
```

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,90 @@ def vector_poly_data(orig, vec):
...
+    Compute its principal axes.
+
+    >>> principal_axes = pv.principal_axes(mesh.points)
+    >>> principal_axes
+    array([[-1.0000000e+00,  1.6606515e-10,  1.9055901e-10],
+           [-1.6606515e-10, -1.0000000e+00, -5.0257620e-10],
+           [ 1.9055901e-10, -5.0257620e-10,  1.0000000e+00]], dtype=float32)
+
+    Note that the principal axes have ones along the diagonal and zeros
+    in the off diagonals. This indicates that the first principal axis is
```
adeak, line 530: I'm less sure about this one but I usually see this spelled either `offdiagonal` or `off-diagonal`, never separately.

adeak (commented):
Some docstring nitpicks.

tests/core/test_utilities.py
user27182: > OK, but that will interact with the current API where vectors are sorted by decreasing size, right? You don't have the freedom to change anything beyond changing 2 vector signs, unless there are degenerate axes, but we're talking about floats so telling "degenerate" is a bit messy issue.

The order must be decreasing, yes. But for symmetric datasets two eigenvalues may be almost equal (up to some tolerance). So in these cases I don't see why you can't swap/reorder two vectors if they both explain the same amount of variance.

adeak: Yes, that is "degeneracy". In fact you can use any two perpendicular vectors from that 2d subspace (plane). As I said, choosing when we have degeneracy can be a bit tricky.

pyvista/core/utilities/points.py
user27182:
```suggestion
    in the off-diagonal. This indicates that the first principal axis is
```

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,90 @@ def vector_poly_data(orig, vec):
...
+    numpy.ndarray
+        Three-item array of the axes sizes.
+
+    Examples
+    --------
+    Create a mesh with points that have the largest variation in ``X``,
+    followed by ``Y``, then ``Z``.
+
+    >>> import pyvista as pv
+    >>> mesh = pv.ParametricEllipsoid(xradius=10, yradius=5, zradius=1)
```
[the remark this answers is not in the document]
user27182: The example has been updated with the sizes returned.

tests/core/test_utilities.py
user27182: > Why is `[z, y, -x]` the default? Is this just what `eigh()` does?

I've added a comment to make this a bit more clear.

Further discussion about degeneracy or swapping vectors can be saved for another PR if/when that feature comes up.

tests/core/test_utilities.py
user27182: It's a good idea, I like this better. I've updated the test to make the intent more clear, and use a high-resolution ellipsoid and check the returned sizes.

tests/core/test_utilities.py
user27182: This has been updated to remove any type casting.

pyvista/core/utilities/points.py
user27182: The handedness of the axes was the important thing, so let's leave the vectors alone, the determinant check and direction swap has been added.

No more normalization or cross products now.

tests/core/test_utilities.py
user27182: > I recommend also adding a dedicated `return_sizes` test where the input are points from an ellipsoid (shifted away from the origin for good measure). If our code works then the sizes should be `np.allclose()` to the ellipsoid's corresponding parameters.

Done, added in https://github.com/pyvista/pyvista/pull/6346/commits/8264409e45fb6fd3ca4a65bfd7c6122fb6c6500f

Though I used a simplified dataset with 6 points instead of the ellipsoid. I noticed  that `sizes` correlates with the mesh density. So, more points -> larger sizes. Using the ellipsoid at first gave me much larger values than the input xradius, yradius, zradius... and simplifying to 6 points reduced the values of `sizes`

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,90 @@ def vector_poly_data(orig, vec):
...
+
+    The axes are computed as the eigenvectors of the covariance matrix from the
+    mean-centered points, and are processed to ensure that they form a right-handed
+    coordinate frame.
+
+    The axes explain the total variance of the points. The first axis explains the
+    largest percentage of variance, followed by the second axis, followed again by
+    the third axis which explains the smallest percentage of variance. The axes may
+    be used to build an oriented bounding box or to align the points to another set
+    of axes (e.g. the world XYZ axes).
```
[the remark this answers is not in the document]
user27182:
```suggestion
    The axes explain the total variance of the points. The first axis explains the
    largest percentage of variance, followed by the second axis, followed again by
    the third axis which explains the smallest percentage of variance. Geometrically,
    they can be interpreted as the semi-major axes of the best-fitting ellipsoid to
    the data (for some value of "best-fitting").

    The axes may be used to build an oriented bounding box or to align the points to
    another set of axes (e.g. the world XYZ axes).
```

tests/core/test_utilities.py
user27182: I was able to make sense of the returned sizes. You have to convert them to ratios/percentages for it to make sense.

The test was updated in https://github.com/pyvista/pyvista/pull/6346/commits/c9bf952c3d676679cb32ea5ab91f2f759db51566

and the docstring example updated too in https://github.com/pyvista/pyvista/pull/6346/commits/c0664d71f3bf0006b970e6533842fd515fec2c54

Other than formatting or minor edits this should now be resolved.

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,119 @@ def vector_poly_data(orig, vec):
...
+    and 10% of the total variance in the points, respectively.
+
+    Let's compare this to the ratios of the known radii of the ellipsoid.
+
+    >>> radii / sum(radii)
+    array([0.6, 0.3, 0.1])
+
+    From this result, we can observe that the relative sizes of the ellipsoid radii
+    correspond to the relative sizes of the principal axes. Hence, the principal axes
+    can be interpreted as the semi-major axes of the best-fitting ellipsoid to the data.
```
user27182, line 566:
```suggestion
    Note how the two ratios are nearly identical. From this result, we interpret the 
    principal axes as the semi-major axes of the best-fitting ellipsoid to the data.
```

user27182:
@pyvista-bot preview

adeak:
@user27182 first of all, apologies for the late reply, life's been too busy.

Second of all, even more apologies, only now did I realise (after seeing your excellent sizes update) how completely wrong I've been about the geometric interpretation of the sizes :disappointed: I was certain that the sizes indeed corresponded to ellipsoid semi-major axes, perfectly when the input points define an ellipsoid. Well that's blatantly wrong if we look at some simple cases.
```python
import pyvista as pv

# sphere: fitting ellipsoid is also a sphere, right?
print(pv.principal_axes(pv.Sphere().points, return_sizes=True)[1])
# [10.087122   7.373941   7.3739405]
```
Scaling issues notwithstanding, one of three axes is much longer. That makes no sense with my mental model. At this point I started to realise that my mental model would only have a chance to work if the points are uniformly distributed. The reason why the sizes are "prolate" is that points are denser near the poles.

But then surely a uniform sphere should work:
```python
import numpy as np

arr = np.random.default_rng().normal(size=(100000, 3))
sphere = arr / np.linalg.norm(arr, axis=1, keepdims=True) * 2  # uniform spherical surface with radius 2
print(pv.principal_axes(sphere, return_sizes=True)[1])
# [366.58908927 364.9330855  363.91548878]
```
At this point I realised what you mentioned [in a comment](https://github.com/pyvista/pyvista/pull/6346#discussion_r1671316196):
> I noticed that sizes correlates with the mesh density. So, more points -> larger sizes. 

Ah, yes. Because we indeed need covariances, and the covariance is a _mean_, so we need to divide the eigenvalues by the number of points! Since we currently have a square root over eigenvalues, we can correct with the square root of the number of points:
```python
print(pv.principal_axes(sphere, return_sizes=True)[1] / np.sqrt(sphere.shape[0]))
# [1.15925649 1.15401974 1.15080182]
```
And, while uniform, this is very much not radius 2.

And then I finally remembered that PCA is not about ellipsoidal point clouds. It's about statistics (as you explained nicely in the docstring example). In particular, the (properly normalized) sizes give us the normal distribution that leads to the same (co)variance of the point cloud.

So what this means is that good agreement is expected for _normally distributed_ points, not ellipsoids! (Bah!) Proof:
```python
# arr is still an array of standard normal numbers
print(pv.principal_axes(arr, return_sizes=True)[1] / np.sqrt(arr.shape[0]))
# [1.00136082 0.99787885 0.99428009]

# normal numbers with scale [1, 2, 3] along [x, y, z]:
print(pv.principal_axes(arr * [1, 2, 3], return_sizes=True)[1] / np.sqrt(arr.shape[0]))
# [2.99160155 1.99490081 0.99885707]
```

What this also implies is that
1. sizes aren't at all as important as I'd thought (because the geometric meaning I thought was there is much weaker, and the statistical meaning is not very easy to understand for our end users). We could use normal distributions to check the sizes, but this is probably more useful for testing rather than our users.
2. Eigenvalues should be divided by the number of points before taking the square root, and I think this won't affect your existing geometric demonstration about relative variances.
3. You can ignore most of my geometry-motivated tips from earlier, because they were mostly wrong. I'm very happy that you found a way to make sense of the sizes despite my "help", sorry again. If based on the above you want to remove the sizes, or remove/rewrite some examples in order for them to make more sense, please do. 

I'll still have to look at your updates in detail, but I wanted to clear this up first.

adeak:
I've taken the liberty to remove my bullshit in https://github.com/pyvista/pyvista/pull/6346/commits/0cdf7a2f64aa7e051dda3a677e7eeba9716f5c84. There's another similar mention at the end of the docstring example, but that's intertwined with your statistical reasoning, so I don't want to touch that. But please consider rephrasing that as well as you see fit.

tests/core/test_utilities.py
```diff
@@ -913,6 +914,103 @@ def test_fit_plane_to_points():
...
+)
+CASE_RANK3 = (
+    3,
+    pv.examples.load_airplane().points,
+    [
+        [8.1159601e-07, -9.9255711e-01, -1.2178054e-01],
+        [-1.0000000e00, -8.0911440e-07, -6.9828459e-08],
+        [-2.9225653e-08, 1.2178054e-01, -9.9255711e-01],
+    ],
+)
```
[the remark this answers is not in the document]
user27182: > Do you know why/how rank was important in the original implementation?

Between the different implementations (NumPy SVD, NumPy PCA, and and VTK OBB), some of them fail for certain edge cases, e.g. one or two points,  many points but they're planar, etc. This required special handling for some of the implementations. So the main intent was to have coverage for these cases, especially if the implementation may change in the future.

> The ranks got me a bit confused here. I expected the two-point cases to always have rank 1, because two points always define a line.

> I think what's going on here is that we're checking the rank of the points array, which is a bit misleading. I think we should be checking the rank of the centered points array, in which case all 2 point cases will be rank1 unless the two points coincide, see the current rank 0 test case.

"Rank" is the term I settled on to describe the edge cases but maybe we can call it something else. I wasn't so much thinking about the actual meaning or definition of it but rather just "what value does `np.linalg.matrix_rank` return". Maybe let's just call them "TEST_CASES" generally instead? And forgo any matrix rank checks.

tests/core/test_utilities.py
```diff
@@ -913,6 +914,103 @@ def test_fit_plane_to_points():
...
+
+def test_principal_axes_single_point():
+    axes = principal_axes([1, 2, 3])
+    assert np.allclose(axes, DEFAULT_PRINCIPAL_AXES)
+
+
+def test_principal_axes_vectors_success_with_many_points():
+    # Use large mesh to verify no memory errors are raised
+    res = 4000
+    ellipsoid = pv.ParametricSuperEllipsoid(
```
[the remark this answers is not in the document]
user27182: Let's use a regular ellipsoid. I used super because... for some reason I thought there wasn't a regular ellipsoid available to use.

tests/core/test_utilities.py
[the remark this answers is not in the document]
adeak: OK, I think your "rank 1" and "rank 2" are the same edge case. I can't check now but I'd expect both to have 2 zero sizes, the long axis along `points[1] - points[0]`, and two other indeterminate axes in the perpendicular plane.

And you're missing the rank 2 case where you have 3 points: one zeeo size, two nonzero sizes, and the zero axis is perpendicular to the plane spanned by the three points.

user27182:
> @user27182 first of all, apologies for the late reply, life's been too busy.

Not a problem! Thanks for reviewing.

> 1. sizes aren't at all as important as I'd thought (because the geometric meaning I thought was there is much weaker, and the statistical meaning is not very easy to understand for our end users). We could use normal distributions to check the sizes, but this is probably more useful for testing rather than our users.
> 3. You can ignore most of my geometry-motivated tips from earlier, because they were mostly wrong. I'm very happy that you found a way to make sense of the sizes despite my "help", sorry again. If based on the above you want to remove the sizes, or remove/rewrite some examples in order for them to make more sense, please do.

I didn't really have a good intuition for what these values or "sizes" really meant until you suggested the geometry link, so it was very helpful! I had to go looking at some articles before getting a hint to divide them by the sum.

> 2. Eigenvalues should be divided by the number of points before taking the square root

Will this normalize the values so that they sum to one? Because I'm thinking that the sizes output would be more useful or at least easier to interpret if the output is always in range [0,1].

Also, should be even still call them `sizes`? This implies the geometric link, but maybe `proportions` is better? I.e. proportion of total variance explained by each axis.

adeak:
> > 2\. Eigenvalues should be divided by the number of points before taking the square root
> 
> Will this normalize the values so that they sum to one? Because I'm thinking that the sizes output would be more useful or at least easier to interpret if the output is always in range [0,1].

No, this will turn the values into standard deviations in data units. I believe this would mean the rough width of normally distributed blobs that have the same variance as the input. We can still choose to normalize these to return ratios instead, and I suspect that's what OBB does, based on its docs. Or just tell users to look at relative sizes if they want.

> Also, should be even still call them `sizes`? This implies the geometric link, but maybe `proportions` is better? I.e. proportion of total variance explained by each axis.

Yes, this is indeed a misnomer. They are standard deviations unless we normalize them to 1. See e.g. the second figure [here](https://en.wikipedia.org/wiki/Covariance_matrix) with a 2d normally distributed blob.

user27182:
> > > 2. Eigenvalues should be divided by the number of points before taking the square root
> > 
> > 
> > Will this normalize the values so that they sum to one? Because I'm thinking that the sizes output would be more useful or at least easier to interpret if the output is always in range [0,1].
> 
> No, this will turn the values into standard deviations in data units. I believe this would mean the rough width of normally distributed blobs that have the same variance as the input. We can still choose to normalize these to return ratios instead, and I suspect that's what OBB does, based on its docs. Or just tell users to look at relative sizes if they want.
> 
> > Also, should be even still call them `sizes`? This implies the geometric link, but maybe `proportions` is better? I.e. proportion of total variance explained by each axis.
> 
> Yes, this is indeed a misnomer. They are standard deviations unless we normalize them to 1. See e.g. the second figure [here](https://en.wikipedia.org/wiki/Covariance_matrix) with a 2d normally distributed blob.

the `sizes` parameter has been renamed to `std` in https://github.com/pyvista/pyvista/pull/6346/commits/76fe72d04bd1ec2e70f2faae2e546a99fbd1319d

Since `std` this has a clear statistical meaning I think we can return it as-is without normalizing them as proportions which sum to 1. It's easy enough to get the proportions if desired, and they're computed in the example to demonstrate this. 

I've also expanded the example to show the difference between the points of an ellipsoid and normally distributed points (this is based of a previous example you gave earlier in the comments)

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,156 @@ def vector_poly_data(orig, vec):
...
+
+    Parameters
+    ----------
+    points : MatrixLike[float]
+        Nx3 array of points.
+
+    return_std : bool, default: False
+        If ``True``, also returns the standard deviation of the points along each axis.
+        Standard deviation is computed as the square root of the eigenvalues of the
+        mean-centered covariance matrix, divided by the number of points.
```
adeak, line 506: Nitpick: dividing by the number of points is what makes it a covariance matrix, then you just take the square root of the eigenvalues:
```suggestion
        mean-centered covariance matrix.
```

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,156 @@ def vector_poly_data(orig, vec):
...
+
+    Returns
+    -------
+    numpy.ndarray
+        3x3 orthonormal array with the principal axes as row vectors.
+
+    numpy.ndarray
+        Three-item array of the axis sizes.
```
adeak, line 514:
```suggestion
        Three-item array of the standard deviations along each axis.
```

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,156 @@ def vector_poly_data(orig, vec):
...
+
+    Plot the mesh and highlight its points in black.
+
+    >>> p = pv.Plotter()
+    >>> _ = p.add_mesh(mesh)
+    >>> _ = p.add_points(mesh, color='black')
+    >>> _ = p.show_grid()
+    >>> p.show()
+
+    Compute its principal axes and return the standard deviation.
```
adeak, line 538: Still nitpick
```suggestion
    Compute its principal axes and return the standard deviations.
```

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,156 @@ def vector_poly_data(orig, vec):
...
+    array([[-0.99997578,  0.00682346,  0.00136972],
+           [ 0.00681368,  0.99995213, -0.00702282],
+           [-0.00141757, -0.00701331, -0.9999744 ]])
+
+    Once again, the axes have ones along the diagonal as expected since the
+    points are already axis-aligned. Now let's examine the standard deviation
+    and compare the relative proportions.
+
+    >>> std
+    array([0.18798689, 0.0915765 , 0.03228786])
```
adeak, line 594: This should be close to `[6, 3, 1]`, see code comment later. If we do PCA with a normal blob then we should get back the same stds, give or take randomness.

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,156 @@ def vector_poly_data(orig, vec):
...
+
+    Let's compare this to the proportions of the known radii of the ellipsoid.
+
+    >>> radii / sum(radii)
+    array([0.6, 0.3, 0.1])
+
+    Note how the two ratios are similar, but do not match exactly. This is
+    because the points of the ellipsoid are prolate and are denser near the
+    poles. If the points were normally distributed, however, the proportions
+    would match exactly.
```
adeak, line 576: Your ellipsoid is _oblate_, so I think there's a competition between "smaller axis along z" and "denser points along z". It might not be obvious which one wins. But this is very handwavy, based on my earlier surprises with `pv.Sphere()`...

Further musing: I'm thinking that (according to my current newly found understanding) PCA fits the covariance of a distribution with normals, a large issue here is that the distribution is very "not normal". My hunch is that we'd get better agreement with a space-filling ellipsoid cloud. But I don't think it would be worth anyone's time looking into that :smile:

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,156 @@ def vector_poly_data(orig, vec):
...
+        3x3 orthonormal array with the principal axes as row vectors.
+
+    numpy.ndarray
+        Three-item array of the axis sizes.
+
+    Examples
+    --------
+    >>> import pyvista as pv
+    >>> import numpy as np
+    >>> rng = np.random.default_rng(seed=0)
```
adeak, line 520: Might be better to protect end users with something like:
```suggestion
    >>> rng = np.random.default_rng(seed=0)  # only seeding for the example!
```
or just removing the seeding and allowing variation in the doctest results.

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,156 @@ def vector_poly_data(orig, vec):
...
+    >>> scaled_points = normal_points * radii
+    >>> axes, std = pv.principal_axes(scaled_points, return_std=True)
+    >>> axes
+    array([[-0.99997578,  0.00682346,  0.00136972],
+           [ 0.00681368,  0.99995213, -0.00702282],
+           [-0.00141757, -0.00701331, -0.9999744 ]])
+
+    Once again, the axes have ones along the diagonal as expected since the
+    points are already axis-aligned. Now let's examine the standard deviation
+    and compare the relative proportions.
```
adeak, line 591: And we probably have a larger difference compared to a unit matrix because we have only "a few" points and randomness. The ellipsoid example had structured points. But this again might not be worth mentioning here.

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,156 @@ def vector_poly_data(orig, vec):
...
+    eig_vals, eig_vectors = np.linalg.eigh(points_centered.T @ points_centered)
+    axes = eig_vectors.T[::-1]  # columns, ascending order -> rows, descending order
+
+    # Ensure axes form a right-handed coordinate frame
+    if np.linalg.det(axes) < 0:
+        axes[2] *= -1
+
+    if return_std:
+        # Compute standard deviation and swap order from ascending -> descending
+        std = (np.sqrt(eig_vals) / len(points))[::-1]
```
adeak, line 617: Division should be on the inside, because the covariance matrix is `points_centered.T @ points_centered / points.shape[0]`:
```suggestion
        std = (np.sqrt(eig_vals / len(points)))[::-1]
```
This should fix the doctest std results.

Perhaps we could put the division right up into the argument of `eigh()`, maybe even explicitly defining `covariance = ...`. It's not clear to me right now what is better for numerical stability. I suspect `eigh()` might be happier if its argument isn't potentially huge.

So perhaps worth putting the factor into the matrix. Perhaps we should define `points_centered = (points - points.mean(0)) / np.sqrt(points.shape[0])`, because then we still use the same memory footprint.

user27182: > So perhaps worth putting the factor into the matrix. Perhaps we should define `points_centered = (points - 
points.mean(0)) / np.sqrt(points.shape[0])`, because then we still use the same memory footprint.

If I make this change (https://github.com/pyvista/pyvista/pull/6346/commits/9841419c4cf60308cb7157ab6b6f7e4cdd12cc40) the output ratios no longer match in `test_principal_axes_return_std`:
```
ratios_in
Out[1]: array([0.5       , 0.33333333, 0.16666667])
ratios_out
Out[2]: array([0.64285714, 0.28571429, 0.07142857])
```
So, this change is reverted. Maybe this falls in the realm of pre-mature optimization, so perhaps can be addressed in a separate PR only if/when its determined that numerical stability and/or memory usage are found to be an issue

> Division should be on the inside, because the covariance matrix is points_centered.T @ points_centered / points.shape[0]

Thanks, fixed in https://github.com/pyvista/pyvista/pull/6346/commits/4766baa1c07381329e0fae19f442cc12d5a7563e

pyvista/core/utilities/points.py
user27182: I'm not really familiar with this technical nuance, so I'll leave it up to you, feel free to modify the wording or the example itself, or even remove it altogether if you don't think it adds value.

TBH I feel the real value of this method will come through with examples using real datasets and applying transformations with the principal axes. I plan on adding a full example in a separate PR for this.

pyvista/core/utilities/points.py
user27182: Similar to a previous comment, I'm not very familiar with these details, I'll leave it up to you if you want o keep/modify/remove any of this.

tests/core/test_utilities.py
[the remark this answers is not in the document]
user27182: > As far as my semi-naive picture goes, I'd expect something like this:
> 
> * rank 0: e.g. what you have now, or a single point
> * rank 1: any 2 non-coincidental points, e.g. what you have now
> * rank 2: any 3 non-collinear points, e.g. `[[1, 0, 0], [0, 1, 0], [0, 0, 1]]`
> * rank 3: any 4 non-coplanar points, e.g. `[[1, 0, 0], [0, 1, 0], [0, 0, 1], [-1, -1, -1]]`

Tests are updated in https://github.com/pyvista/pyvista/pull/6346/commits/a521e83cae901c0f3bb9f06ab949773a3e0bff09

The rank 2 case suggested above has a `np.linalg.matrix_rank()` equal to 3, so using term rank may be confusing, so the cases are renamed and the rank assertion has been remove. The intent is communicated in comments instead. 
Plus, this rank 2 case spat out `nan` as its 3rd std value. Not sure if we like that, or should force it zero? This behaviour doesn't seem consistent with other cases, e.g. the single point cases doesn't have nans, but maybe I would have expected it to based on the rank2 behaviour.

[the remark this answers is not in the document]
user27182: > * rank 3: any 4 non-coplanar points, e.g. `[[1, 0, 0], [0, 1, 0], [0, 0, 1], [-1, -1, -1]]`

The axes for this test in python 3.8 are:
```
[[ 0.57735027, 0.57735027, 0.57735027],
 [-0.44225427, -0.37327035, 0.81552462],
 [ 0.68635109, -0.72617898, 0.03982789]],
```
But it's expected to be:    
```
[[-0.57735027, -0.57735027, -0.57735027],
 [0.0, -0.70710678, 0.70710678],
 [-0.81649658, 0.40824829, 0.40824829]]
```
The axes are still a valid result, but this makes it hard annoying to test since we need to special-case it. So a skip is added in https://github.com/pyvista/pyvista/pull/6346/commits/4b807de1d0b44382c4b2f675986d699426ca58da

user27182:
I believe the technical aspects of this PR are complete. The only outstanding issue seems to be related to the contents/wording of the docstring example here https://github.com/pyvista/pyvista/pull/6346#discussion_r1677200190

adeak:
Sorry again for dropping the ball on this. I'm not too worried about documentation. My main concern is [this thread](https://github.com/pyvista/pyvista/pull/6346#discussion_r1677201222). I'm highly convinced that my suggestion would lead to the proper behaviour, which should be tested against normal blobs. If fixing the code means the simple example breaks then the simple example has to be changed.

But since resolving this needs me to get some time to play with the actual code and my suggestion, I haven't been able to come back to this...

user27182:
> Sorry again for dropping the ball on this. I'm not too worried about documentation. My main concern is [this thread](https://github.com/pyvista/pyvista/pull/6346#discussion_r1677201222). I'm highly convinced that my suggestion would lead to the proper behaviour, which should be tested against normal blobs. 

No problem, thanks for highlighting your concerns. Can you be more specific about which suggestion you are referring  to? This code suggestion:
``` diff
- std = (np.sqrt(eig_vals) / len(points))[::-1]
+ std = (np.sqrt(eig_vals / len(points)))[::-1]
```
was implemented in https://github.com/pyvista/pyvista/commit/4766baa1c07381329e0fae19f442cc12d5a7563e

Is the main concern that the current implementation has a sub-optimimal memory footprint and/or sub-optimal numerical precision? If so, we can make these improvements as part of a separate PR, or do you think it should block this one? Having this kind of optimization as a separate PR would also help better evaluate and compare performance before / after since both PRs will have separate commits to main in that case with a clear diff. Especially if you believe the tests may need to be changed, it would be good to have a clear record of this. This kind of change might otherwise get lost in the review comments here.

adeak:
> This code suggestion:
> 
> ```diff
> - std = (np.sqrt(eig_vals) / len(points))[::-1]
> + std = (np.sqrt(eig_vals / len(points)))[::-1]
> ```
> 
> was implemented in [4766baa](https://github.com/pyvista/pyvista/commit/4766baa1c07381329e0fae19f442cc12d5a7563e)

My main concern was that the `std` results are wrong, because I missed that this part was fixed. You said as much in your thread reply at the end, I just didn't read carefully enough (or if I did, I've forgotten since), sorry.

So yes, the remaning part of that comment of mine is about numerical stability. If we can move the division with the number of points to before `eigh()` then the eigensolver can work with a much better scaled matrix (otherwise for 1M points we'll get 1M times the covariance matrix in there, which is obviously worse for numerical precision). But then we have a memory issue unless we sneak the division into the vector centering step as I suggested. Now, what's still fishy to me is that this changes the results as per your reply. I don't think that should happen unless the results were very inaccurate to begin with. This is also what I wanted to investigate before responding...

>  If so, we can make these improvements as part of a separate PR, or do you think it should block this one? 

That's very fair, sorry for holding up this one. There's no telling when I'll have a couple of hours to get my hands dirty with this one, so if the basic functionality is correct then this shouldn't wait anymore. I hope there's some kind of test (doctest or unit, doesn't matter) which proves that for a _normal_ blob we indeed get the "same" standard deviations back. Ideally I'd use a unit test which generates a large enough normal blob (probably with a seeded NumPy generator to avoid any flakes in future test runs), because that's the "right" test to numerically verify the `std` results. But again, this is not critical to get this out the door if the results make sense (as they are, not just when converted to ratios!).

adeak (approved):
Approving so as not to block this any further. Would be nice to have a follow-up with some open comments in the above thread.

user27182:
@adeak thanks for your review and approval. 

> My main concern was that the `std` results are wrong

Added explicit tests for this in https://github.com/pyvista/pyvista/pull/6346/commits/8d14cce9abcd562e4dc9d53b81c0b68cab1fdf9c, this should be reviewed.
I used the code from this comment for creating the fixture. https://github.com/pyvista/pyvista/pull/6346#issuecomment-2226308993.

> So yes, the remaning part of that comment of mine is about numerical stability. If we can move the division with the number of points to before `eigh()` then the eigensolver can work with a much better scaled matrix (otherwise for 1M points we'll get 1M times the covariance matrix in there, which is obviously worse for numerical precision). But then we have a memory issue unless we sneak the division into the vector centering step as I suggested. Now, what's still fishy to me is that this changes the results as per your reply. I don't think that should happen unless the results were very inaccurate to begin with. This is also what I wanted to investigate before responding...

There are indeed issues with numerical stability which would be nice to address. It's very possible the tests were failing because I either didn't implement your suggestion correctly or the `atol` wasn't set up correctly. This was also pre-fix for the std calc so that may have also been the cause. Definitely worth investigating further if possible.

>  I hope there's some kind of test (doctest or unit, doesn't matter) which proves that for a _normal_ blob we indeed get the "same" standard deviations back. Ideally I'd use a unit test which generates a large enough normal blob (probably with a seeded NumPy generator to avoid any flakes in future test runs), because that's the "right" test to numerically verify the `std` results. 

There weren't any explicit tests for this, so tests were added in https://github.com/pyvista/pyvista/pull/6346/commits/8d14cce9abcd562e4dc9d53b81c0b68cab1fdf9c, see above.

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,161 @@ def vector_poly_data(orig, vec):
...
+    eig_vals, eig_vectors = np.linalg.eigh(points_centered.T @ points_centered)
+    axes = eig_vectors.T[::-1]  # columns, ascending order -> rows, descending order
+
+    # Ensure axes form a right-handed coordinate frame
+    if np.linalg.det(axes) < 0:
+        axes[2] *= -1
+
+    if return_std:
+        # Compute standard deviation and swap order from ascending -> descending
+        std = np.sqrt(eig_vals / len(points), out=eig_vals)[::-1]
```
user27182, line 622: This can be reverted, left over from investigating the array type of the eigenvalues.
```suggestion
        std = np.sqrt(eig_vals / len(points))[::-1]
```

adeak:
> Added explicit tests for this in [8d14cce](https://github.com/pyvista/pyvista/commit/8d14cce9abcd562e4dc9d53b81c0b68cab1fdf9c), this should be reviewed.
> I used the code from this comment for creating the fixture. [#6346 (comment)](https://github.com/pyvista/pyvista/pull/6346#issuecomment-2226308993).

Sorry that I keep making you run in circles, but this is not what I had in mind. Now the test in https://github.com/pyvista/pyvista/commit/8d14cce9abcd562e4dc9d53b81c0b68cab1fdf9c has 1.15 in there as a magic number. Instead what I had in mind is the _last_ block [in my same comment](https://github.com/pyvista/pyvista/pull/6346#issuecomment-2226308993):

> So what this means is that good agreement is expected for _normally distributed_ points, not ellipsoids! (Bah!) Proof:
> 
> ```python
> # arr is still an array of standard normal numbers
> print(pv.principal_axes(arr, return_sizes=True)[1] / np.sqrt(arr.shape[0]))
> # [1.00136082 0.99787885 0.99428009]
> 
> # normal numbers with scale [1, 2, 3] along [x, y, z]:
> print(pv.principal_axes(arr * [1, 2, 3], return_sizes=True)[1] / np.sqrt(arr.shape[0]))
> # [2.99160155 1.99490081 0.99885707]
> ```

In this case we have a normal blob, and the `std` values we get are "the same" as the parameters we use to generate the normal distributions. There's no magic numbers.

To be clear in the test I have in mind you'd generate `rng.normal((n, 3)) * [1, 2, 3]`, and then confirm that the `std` directly from `principal_axes()` is roughly `[3, 2, 1`.

user27182:
> To be clear in the test I have in mind you'd generate `rng.normal((n, 3)) * [1, 2, 3]`, and then confirm that the `std` directly from `principal_axes()` is roughly `[3, 2, 1]`.

This makes a lot more sense as a test, thanks for clarifying. I have updated the test in https://github.com/pyvista/pyvista/pull/6346/commits/5e425f2ab69a59d51cbf387d6687eef322e2294c

pyvista/core/utilities/points.py
```diff
@@ -459,3 +464,161 @@ def vector_poly_data(orig, vec):
...
+
+    >>> std
+    array([3.014956 , 1.507478 , 0.7035637], dtype=float32)
+
+    Compare this to using :meth:`numpy.std` for the computation.
+
+    >>> np.std(mesh.points, axis=0)
+    pyvista_ndarray([3.0149572, 1.5074761, 0.7035699], dtype=float32)
+
+    Convert the values to proportions for analysis.
```
user27182, line 565:
```suggestion
    Since the points are axis-aligned, the two results agree in this case. In general, 
    however, these two methods differ in that :meth:`numpy.std` with `axis=0` computes
    the standard deviation along the `x-y-z` axes, whereas the standard deviation 
    returned by :meth:`principal_axes` is computed along the principal axes.

    Convert the values to proportions for analysis.
```

user27182:
pre-commit.ci autofix

adeak (approved):
Thanks for the good work and the patience @user27182.

pyvista/core/utilities/points.py
adeak: For what it's worth I tried moving the division by number of points into the argument of `eigh()` (so numerical errors would still be relevant in the calculation of the covariance matrix) or into the definition of `points_centered` (with which we'd "probably" minimize numerical issues). In all cases I get the exact same `std_out` for the `100_000``-sized normal blob in the `std` test (also the exact same axes).

PR #6353, opened 2024-07-08
Add `relative_position` and `Prop3D`-like features to `Label` class

user27182:
### Overview

The `Label` class allows text to be positioned in 3D space. However, the labels are often used to label _something_, e.g. an actor (or dataset), and we may want the position of the label to be fixed relative to the thing being labeled. But, currently if we move, re-orient, or otherwise transform the thing being labeled, then a custom transformation must also be applied to the label to keep its relative positioning the same. 

This PR simplifies this process and abstracts away the need to manually apply these transformations. It does this by adding a new `relative_position` property and a `Prop3DMixin`, allowing the `Label` object  to internally compute and apply the necessary transformation to keep its relative positioning the same.

The `AxesAssembly` class is also refactored to make use of this feature. (The class no longer needs to apply custom transformations to update the relative positioning of the labels.)

### Details

Since `Label` is a 2D actor, not 3D, it does not inherit from `Prop3D`. Therefore, a  `Prop3DMixin` class is added to add the necessary `position`, `orientation`, `origin`, and `user_matrix` properties used by `Prop3D` for actor transformations. This enables `Prop3D` transformations to be applied to non-Prop3d classes such as `Label`.


user27182:
@pyvista-bot preview

pyvista/plotting/text.py
```diff
@@ -253,44 +262,98 @@ class Label(Text):
...
+    >>> label.relative_position
+    (0.5, 0.0, 0.0)
+
+    Plot the results.
+
+    >>> pl = pv.Plotter()
+    >>> _ = pl.add_actor(cone_actor)
+    >>> _ = pl.add_actor(label)
+    >>> pl.show()
+
```
tkoyama010, line 318: Usefulness of this function will become apparent after you move the cone in my mind. Can you add an explanation and code of that as well?

tkoyama010 (changes requested):

tkoyama010:
pre-commit.ci autofix

pyvista/plotting/text.py
user27182: The example is updated in https://github.com/pyvista/pyvista/pull/6353/commits/b8eaaeae34c9874f9c37ec5615a36d14335ebe5c, hopefully the utility of `relative_position` is clearer now.

tkoyama010:
@pyvista-bot preview

user27182:
Looks like the docs aren't deployable

> Netlify credentials not provided, not deployable

user27182:
So, I'm not sure how to proceed with review for this PR (or any others which I have open where the documentation should be reviewed). Since the docs can't be previewed, we can close this and continue review in #6373 . But the review comments from this PR no longer apply to #6373, which isn't ideal.

Alternatively, we continue all review here and push any changes here, but then also sync #6373 solely for the purpose of deploying and reviewing the docs?

user27182:
This PR was merged as #6373

PR #6354, opened 2024-07-09
Add `transformations.rotation()` and `DataSet.rotate()` for rotating datasets

user27182:
### Overview
The function `transformations.rotation` generates a 4x4 transformation matrix, and `DataSet.rotate` applies it to a dataset. The methods are similar to `Prop3D.rotation_from` from #6318 , except they are for datasets, and also allow specifying a rotation point.

These methods will help make it easy to apply coordinate system transformations, e.g. transforming a mesh from its principal axes from #6346 to origin-centered world XYZ axes.


pyvista/core/dataset.py
```diff
@@ -47,6 +47,8 @@
     from collections.abc import Generator
     from collections.abc import Iterator
 
+    import scipy
+
```
tkoyama010, line 51: Since `scipy` is not currently included in our dependencies, we need to check whether it can be imported. We also need to set up a process for when it cannot be imported.
https://github.com/pyvista/pyvista/blob/9a88c38b87969899709289abc0b45c42cfc1e5a0/pyproject.toml#L29C1-L37

pyvista/core/dataset.py
```diff
@@ -1139,6 +1141,74 @@ def rotate_vector(
             inplace=inplace,
         )
 
+    def rotate(
```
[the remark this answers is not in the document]
tkoyama010: `rotate` +1. In my mind, users can understand that the matrix is given as an argument rotation.

tkoyama010 (changes requested):

pyvista/core/dataset.py
user27182: Good catch this should be for type-checking only, as validate_transform3x3 otherwise will handle the import at runtime. I wonder why ruff didn't suggest moving it to a type-checking block

pyvista/core/dataset.py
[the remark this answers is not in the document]
user27182: Ok let's keep it as-is

pyvista/core/dataset.py
user27182: So, it's already in a type checking block. If type-checking anything, the code will still be able to continue type-checking when it hits this line, but missing imports may be imported as `Any`. 

A similar import is already in `main`, merged from #6318:
https://github.com/pyvista/pyvista/blob/9a88c38b87969899709289abc0b45c42cfc1e5a0/pyvista/plotting/prop3d.py#L15-L16

So if this is an issue we should fix it there too

user27182: Maybe we should add `scipy` as a type-checking dependency first, in a separate PR?

tkoyama010: > Maybe we should add `scipy` as a type-checking dependency first, in a separate PR?

Yes let's do it. Thanks :)

MatthewFlamm: Given that we cannot use scipy in pre-commit mypy, I think that the current state is fine.  We can add in scipy to our type checking once we get it out of pre-commit.

user27182 (commented):
Suggest using `RotationLike` from #6431. This would also remove the need to import scipy locally since this is handled by the import of `RotationLike`

pyvista/core/dataset.py
user27182: This should be resolved now by using the `RotationLike` alias as part of https://github.com/pyvista/pyvista/pull/6354/commits/48aa6e6d335638dba04731bb03d6d9330e080558.

There is still the outstanding issue of not having `SciPy` as a dependency as part of the `mypy` config (see #6368), but that's a separate issue from this PR.

tkoyama010 (approved):
LGTM

PR #6356, opened 2024-07-09
BUG: Fix bug with reference cycle

larsoner:
https://github.com/pyvista/pyvista/pull/3015 introduced a reference cycle that Python's garbage collector can't resolve properly, see for example this MNE-Python CI run where we make sure there aren't any lingering `vtkPolyData` after running each example, and following 0.44 release there were some:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/24519/workflows/d7b1e478-c51f-46e3-96c4-259b5be3b627/jobs/67052

The proposed fix uses `weakref.ref` to resolve this, storing just a weakref to the first geom (which is all that's used later). I confirmed locally that it fixes the problem. Not sure a great way to test for this at the `pyvista` level.

### DIFF
--- a/pyvista/core/_validation/check.py
+++ b/pyvista/core/_validation/check.py
@@ -259,13 +259,13 @@ def check_sorted(
         )
 
 
-def check_finite(arr, /, *, name="Array"):
+def check_finite(array: _ArrayLikeOrScalar[NumberType], /, *, name: str = "Array"):
     """Check if an array has finite values, i.e. no NaN or Inf values.
 
     Parameters
     ----------
-    arr : array_like
-        Array to check.
+    array : float | ArrayLike[float]
+        Number or array to check.
 
     name : str, default: "Array"
         Variable name to use in the error messages if any are raised.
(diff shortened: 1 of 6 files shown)


larsoner:
Realized `weakref` won't work here, and instead a better approach anyway is to fix `PolyData.__del__` to set `self._glyph_geom = None`. The CIs should now hopefully come back green :crossed_fingers:

MatthewFlamm:
This LGTM.  It looks like the downstream MNE-Python CI is passing on this branch, but was that with the original commit (https://github.com/pyvista/pyvista/pull/6356/commits/34752db7f0fe01a0cc6390323878be65cabc6652) or the new commit (https://github.com/pyvista/pyvista/pull/6356/commits/2dfac25c57e802d4a9b16716791515a8f288ed7d)?  Just want to make sure that this does indeed fix the issue.

larsoner:
Yep I tested locally and also just restarted the CircleCI run which [still came back green](https://app.circleci.com/pipelines/github/mne-tools/mne-python/24525/workflows/f68e4bce-7669-4a86-b40f-0452e7bdc271/jobs/67079) on 313bf2f90655910d483a691e7c7c863b2b48e6d4!

MatthewFlamm (approved):

PR #6361, opened 2024-07-10
RFC: Deploy VTK distribution agnostic PyVista package

banesullivan:
**Request for Comment**

We get a number of requests to provide better compatibility between PyVista and some of the different build variants of VTK deployed by Kitware at https://wheels.vtk.org where the distribution name of the package might differ from `vtk` (e.g., `vtk-egl` and `vtk-osmesa`).

Until setuptools solves https://github.com/pypa/setuptools/issues/1139 this might be our best route forward to provide a version of PyVista that end-users can depend on while specifying the exact distribution of VTK they need.

The crux of the problem is that the main `pyvista` package depends on `vtk` to make installing PyVista as easy as possible for the average user, but other builds of VTK may have different distribution names like `vtk-osmesa` and `vtk-egl` which setuptools/pip cannot properly resolve. Pip sees `vtk-osmesa` as an entirely seperate package from `vtk`, because they are - they are different distributions, though they have the same `import` name of `vtk`. This leads to namespace collisions if installing both, which is incredibly easy to accidentally do.

In this pull request, I'm proposing that we deploy a second version of `pyvista`, named ~`pyvista-agnostic`~  `pyvista-no-vtk` on PyPI that doesn't have a dependency on `vtk` at all so that end users can specify the exact VTK distribution they need.

This will allow downstream packages to depend on PyVista via the `pyvista-no-vtk` package and point the the exact package index and distribution name of VTK that they want.

Better naming suggestions welcomed!

### What will this look like

Before:

```bash
pip install pyvista
pip uninstall vtk
pip install --extra-index-url https://wheels.vtk.org vtk-osmesa
```

After:

```bash
pip install --extra-index-url https://wheels.vtk.org vtk-osmesa pyvista-no-vtk
```


banesullivan:
cc @pyvista/kitware

akaszynski:
I've considered this as well as sometimes I have to ship my own custom compiled VTK wheels and `pyvista` complains when we don't have VTK.

`pyvista-no-vtk` would be my vote because it's perfectly clear what it's doing. Ideally there would be an option to:

```
pip install pyvista[no-vtk]
```

But optional requirements are additive and not subtractive.

banesullivan:
>Ideally there would be an option to `pip install pyvista[no-vtk]` But optional requirements are additive and not subtractive.

Exactly. https://github.com/pypa/setuptools/issues/1139 would make this work, but until that is implemented upstream, shipping a separate package might be our best bet

adeak:
It would be nice if we could log some message after installation of the special-cased package, to raise awareness to users installing it by mistake. From a quick search this doesn't seem feasible.

What we _can_ do is modify the agnostic package to check somewhere early if `vtk` import fails, and if it does, raise an informative error along the lines of "did you forget to install VTK?".

banesullivan:
With VTK 9.4 this shouldn't be needed. Closing in favor of #6879

PR #6366, opened 2024-07-11
Add `OrthogonalPlanesSource` for generating orthogonal planes

user27182:
### Overview

New class for generating orthogonal planes. The output is a `MultiBlock` with the three planes. The planes can be sized to fit any `bounds` input. The orientations of the planes are well-defined and the direction of their normals can be controlled. 

The planes can be used for downstream computations, e.g. distance from plane queries. They are also useful for anatomical orientation, see the vtk example: https://examples.vtk.org/site/Cxx/VisualizationAlgorithms/AnatomicalOrientation/ 

``` python
import pyvista as pv
from pyvista import examples
human = examples.download_human()
planes = pv.OrthogonalPlanesSource(human.bounds).output
pl = pv.Plotter()
pl.add_mesh(human)
pl.add_mesh(planes, opacity=0.3, show_edges=True)
pl.show() 
```


user27182:
@pyvista-bot preview

user27182:
@pyvista-bot preview

pyvista/core/utilities/geometric_sources.py
```diff
@@ -3384,3 +3386,192 @@ def _make_axes_parts(
...
+        yz_source.SetCenter(center)
+        yz_source.Update()
+
+        zx_source.SetXResolution(z_res)
+        zx_source.SetYResolution(x_res)
+        zx_source.SetPoint1(0.0, 0.0, z_size)
+        zx_source.SetPoint2(x_size, 0.0, 0.0)
+        zx_source.SetOrigin(ORIGIN)
+        zx_source.SetCenter(center)
+        zx_source.Update()
```
tkoyama010, line 3549: I plan to update this after merging #6372.

tkoyama010 (changes requested):

pyvista/core/utilities/geometric_sources.py
```diff
@@ -3384,3 +3386,192 @@ def _make_axes_parts(
...
+
+        # Update the output
+        output = self._output
+        for index, name, source, plane, sign in zip(
+            range(3), self.names, self._plane_sources, output, self.normal_sign
+        ):
+            plane.copy_from(source.GetOutput())
```
user27182, line 3556: Related to https://github.com/pyvista/pyvista/pull/6366/files#r1676027612

We can remove this line if we also modify `pv.PlaneSource` to make use of `copy_from` mechanics instead of `pv.wrap` in the output.

Most of the pyvista sources seem to use `pv.wrap` in the output. But updating the output this way creates a new output instance every time. This differs from vtk sources, where the output always refers to the same instance.
e.g.
``` python
import pyvista as pv
import vtk as vtk
vtk_source = vtk.vtkPlaneSource()
assert vtk_source.GetOutput() is vtk_source.GetOutput() # success
pv_source = pv.PlaneSource()
assert pv_source.output is pv_source.output  # fail
```

This is an issue because this class  (`OrthogonalPlanesSource`)  stores the outputs in `__init__`, and any updates thereafter therefore needs to only update the outputs without creating new instances.

pyvista/core/utilities/geometric_sources.py
user27182: Thanks!

I think having an explicit `update` method is good to have too, otherwise we need to get the `output` property without actually storing its value to implicitly do the update.

``` python
zx_source.output
```

which `ruff` will complain about:
```
B018 Found useless expression. Either assign it to a variable or remove it.
```

so then we need to do this instead which isn't a great solution IMO 
```python
_ = zx_source.output
```

user27182: @tkoyama010 if you're OK with it I think we can update this PR with changes from https://github.com/pyvista/pyvista/pull/6372, but keep the existing `Update()` and `copy_from(GetOutput())` mechanics for now, as I think those should probably be batch-updated for all sources that use `pv.wrap` (with tests for the whole module  to verify that `pv_source.output is pv_source.output` is always `True`), e.g.:
``` diff
-        xy_source.SetXResolution(x_res)
-        xy_source.SetYResolution(y_res)
-        xy_source.SetPoint1(x_size, 0.0, 0.0)
-        xy_source.SetPoint2(0.0, y_size, 0.0)
-        xy_source.SetOrigin(ORIGIN)
-        xy_source.SetCenter(center)
+        xy_source.i_resolution = x_res
+        xy_source.j_resolution = y_res
+        xy_source.point_a = x_size, 0.0, 0.0
+        xy_source.point_b = 0.0, y_size, 0.0
+        xy_source.origin = ORIGIN
+        xy_source.center = center
         xy_source.Update()
```

user27182: Resolved by https://github.com/pyvista/pyvista/pull/6366/commits/0854c1c4d512ac99749abe0f6ff2e34be68e08ac

user27182:
pre-commit.ci autofix

pyvista/core/utilities/geometric_sources.py
```diff
@@ -3500,3 +3502,195 @@ def _make_axes_parts(
...
+        zx_source.point_b = x_size, 0.0, 0.0
+        zx_source.origin = ORIGIN
+        zx_source.center = center
+        zx_source.Update()
+
+        # Update the output
+        output = self._output
+        for index, name, source, plane, sign in zip(
+            range(3), self.names, self._plane_sources, output, self.normal_sign
+        ):
```
tkoyama010, line 3672:
```suggestion
        for index, (name, source, plane, sign) in enumerate(zip(
            self.names, self._plane_sources, output, self.normal_sign
        )):
```

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM. Strictly speaking, we should deploy the document preview in the `pyvista` repository branch, but the effect we will get is limited. We will not run it here.

user27182:
Looks like we another documentation cache issue, similar to https://github.com/pyvista/pyvista/pull/6393#issuecomment-2237573358
This time the issue is with images removed from the docs in #6389.

PR #6368, opened 2024-07-11
Add `scipy` dependency to `mypy` pre-commit config

user27182:
### Overview

Follow-up from https://github.com/pyvista/pyvista/pull/6354#discussion_r1674059803

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,6 +68,7 @@ repos:
             "numpy>=2.0",
             "npt-promote==0.1",
             "matplotlib<3.9.1",
+            "scipy<1.14.1",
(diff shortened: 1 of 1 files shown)


user27182:
Looks like we won't be able to add `scipy` as a dependency, or any other relative-large packages to pre-commit configs. pre-commit.ci allows max 250MB

```
build of https://github.com/pre-commit/mirrors-mypy:mypy-extensions==1.0.0,toml==0.10.2,numpy>=2.0,npt-promote==0.1,matplotlib<3.9.1,scipy<1.14.1@v1.10.1 for python@python3 exceeds tier max size 250MiB: 348.4MiB
```

PR #6373, opened 2024-07-12
Add `relative_position` and `Prop3D`-like features to `Label` class

user27182:
Duplicate of #6353, but branched from the `pyvista` repo directly instead of a fork. This is done to enable previewing the docs since this no longer appears to be possible from forks after #5360 was merged.


tkoyama010 (approved):
LGTM. Thanks for this!

PR #6375, opened 2024-07-13
Add `sphinx-tags` to classify `sphinx-gallery` examples

tkoyama010:
### Overview

As you can see on the following page, the examples in the sphinx-gallery are categorized by tag in the Matplotlib documentation. We will apply this to the PyVista documentation as well.

- https://matplotlib.org/stable/_tags/tagsindex.html

### Details

- This is an idea of @story645.
- At present, categories are defined as directory name, but we will be adding more useful tags as we use them in the future.

PR #6378, opened 2024-07-13
Fix block splitter syntax to `# %%` in `sphinx-gallery`

tkoyama010:
### Overview

This change was made because I noticed a difference between PyVista and Matplotlib script file when making the #6375 change.  

The following statement from the `sphinx-gallery` documentation is a good reason for making this change.

https://matplotlib.org/stable/gallery/

> The #%% and # %% syntax is consistent with the ‘code block’ (or ‘code cell’) separator syntax in [Visual Studio Code Python extension](https://code.visualstudio.com/docs/python/jupyter-support-py#_jupyter-code-cells), [Visual Studio Python Tools](https://learn.microsoft.com/en-us/visualstudio/python/python-interactive-repl-in-visual-studio?view=vs-2019#work-with-code-cells), [Jupytext](https://jupytext.readthedocs.io/en/latest/formats.html#the-percent-format), [Pycharm Professional](https://www.jetbrains.com/help/pycharm/running-jupyter-notebook-cells.html), [Hydrogen plugin (for Atom)](https://nteract.gitbooks.io/hydrogen/content/docs/Usage/Cells.html#example-definitions) and [Spyder](https://docs.spyder-ide.org/current/panes/editor.html#defining-code-cells).

### Details

- https://sphinx-gallery.github.io/stable/syntax.html#embed-rest-in-your-example-python-files


tkoyama010:
@pyvista-bot LGTM

adeak:
Late to the party, but good catch, thanks @tkoyama010!

MatthewFlamm:
Not that this PR missed something, but I noticed that https://docs.pyvista.org/version/stable/examples/99-advanced/add-example is missing what this block does.  Should it be added there?

adeak:
Sounds good to me @MatthewFlamm. I didn't even know we had such a page, very nice.

It also looks like the last section has broken syntax, making the last section a code block (probably a missing empty line?):
image

Would be nice to fix that in the same go.

Issue #6167, opened 2024-05-31
Fix failing doc image tests


user27182:
### Overview
Issue to track and investigate failing tests with the documentation images.
The failing tests are initially reported in https://github.com/pyvista/pyvista/pull/5712, see the build failure [here](https://github.com/pyvista/pyvista/actions/runs/9278295702).

### Image Diff
Here's a diff of 54 failing test images:
https://github.com/pyvista/pyvista/compare/main...user27182:doc/failed_images

There appears to be two main categories of failed tests: 
1. failures from random numbers, i.e. examples that were not captured as part of #6106 but should have been, and
2. failures with MPLCharts, see #6141 

### Software Report Diff
Here's a diff between the Software Report from the successful build in https://github.com/pyvista/pyvista/pull/5718 (when the build images in the `doc_image_cache` were first generated) and the failed build linked above:
https://www.diffchecker.com/WGONBfLU/
From this diff it doesn't appear as though updates to any packages is the cause of the failures.

### Steps to reproduce the bug.

n/a

### System Information

```shell
n/a
```

user27182:
One of the failures is with `ImageNoiseSource`. According to the docs for  [vtkImageNoiseSource](https://vtk.org/doc/nightly/html/classvtkImageNoiseSource.html#details):
"There is one major problem with this source. Every time it executes, it will output different pixel values."

So, this test cannot be made reproducible. It will need to be special-cased, e.g. always converted to a warning.

user27182:
The three issues linked above should resolve all failed tests except for:
1. MPLChart images (approx 19 images)
2. test case `sphx_glr_contouring_004` (example which uses `frog_tissue` dataset)

For 1. I don't know what the cause is. See #6141. Perhaps for now we assume the tests are reproducible and simply commit the new (incorrect, zoomed-in) images. If the tests are found to be flaky, we can add these as flaky test cases explicitly later.

For 2. Same thing, don't know the cause. See #5291. This one is probably flaky, but for now maybe just commit the new image from a build and see if it still fails, and add these as flaky test cases explicitly later. Alternatively, we can use a different dataset in the example.

user27182:
In summary, these five PRs should resolve the 54 failed tests:

- 1 failure (rng) fixed by https://github.com/pyvista/pyvista/pull/6168
- 32 failures (rng) fixed by https://github.com/pyvista/pyvista/pull/6169
[...]

PR #6380, opened 2024-07-14
Fix missing `read_func` in  openfoam download

MatthewFlamm:
### Overview

Fix problem noted in https://github.com/pyvista/pyvista/discussions/6376#discussioncomment-10040234


MatthewFlamm:
image

MatthewFlamm:
@user27182 The doc image regression should be very useful to prevent these sort of regressions in the future.  Also, great work on the image regression, it is really well thought out and convenient to use.

MatthewFlamm:
I suspect the sphinx gallery cache is causing this to fail as the failed build images are still the old ones in the artifacts.

user27182:
> @user27182 The doc image regression should be very useful to prevent these sort of regressions in the future. Also, great work on the image regression, it is really well thought out and convenient to use.

Glad it's working out well! Pretty sure I caused this bug in the first place as part of #5822 , so hopefully adding the doc image tests redeems me. In hindsight it would have been better to merge these tests _before_ that PR since the downloads otherwise have very little test coverage.

> I suspect the sphinx gallery cache is causing this to fail as the failed build images are still the old ones in the artifacts.

We ran into this issue before, e.g. a bunch of tests suddenly failed in #6167 when the build cache was updated, those errors should have been caught earlier but weren't due to the re-use of an out-of-date build cache. I wonder if we could run a workflow, maybe daily, which fully updates the build cache along with all the gallery examples.

MatthewFlamm:
> > I suspect the sphinx gallery cache is causing this to fail as the failed build images are still the old ones in the artifacts.
> 
> We ran into this issue before, e.g. a bunch of tests suddenly failed in #6167 when the build cache was updated, those errors should have been caught earlier but weren't due to the re-use of an out-of-date build cache. I wonder if we could run a workflow, maybe daily, which fully updates the build cache along with all the gallery examples.

I looked into this more, and I'm not so sure what is going on now.  The `tests/doc/doc_image_cache` are not cached in the CI.  Are the images cached in the gallery examples used as the `from_build` output in the artifact?  If so, then I think my original assessment is correct.

I propose a label to avoid the cache, since every time we build on main, we will always get the incorrect images in the cache.

user27182:
> I looked into this more, and I'm not so sure what is going on now. The `tests/doc/doc_image_cache` are not cached in the CI. 

Correct, for the generated artifacts, images in `from_cache` are copied directly from images in the repo in `tests/doc/doc_image_cache`. 

> Are the images cached in the gallery examples used as the `from_build` output in the artifact? If so, then I think my original assessment is correct.

Cached gallery examples are not used directly, in the sense that the doc image tests do not go out of their way to make use of them.  Rather, images in `from_build` are copied from the locally-generated build, however it was generated. So, if the build is generated using cached gallery examples, then those cached images will part of the build output, and then yes the doc image tests will use those cached images for the tests (and will be part of any artifact uploads from the doc image tests).

> If so, then I think my original assessment is correct.

Yes I think you are right, i.e. that:

 > the sphinx gallery cache is causing this to fail as the failed build images are still the old ones in the artifacts.

tkoyama010 (approved):
LGTM

PR #6383, opened 2024-07-15
Enable invalidating caches for documentation build with label from PR

MatthewFlamm:
### Overview

See discussion in https://github.com/pyvista/pyvista/pull/6380#issuecomment-2228812354.  Sometimes we need to invalidate the cache on a case by case basis.  With image regression testing for documentation, this has become even more important.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  USE_CACHE: ${{ (github.event_name == 'workflow_dispatch' && github.event.inputs.cache == 'true') || (github.event_name == 'pull_request') || (github.event_name == 'push') }}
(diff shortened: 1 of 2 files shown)


MatthewFlamm:
This skipped the sphinx gallery cache with the label `no-gallery-cache` in https://github.com/pyvista/pyvista/actions/runs/9943759501/job/27468317690?pr=6383.  Will now remove label and close/reopen to test without label.

MatthewFlamm:
The run without the label in https://github.com/pyvista/pyvista/actions/runs/9943898342/job/27468732878?pr=6383 did run the cache check, but confusingly it is no longer finding the cache key that was found in #14680.  I don't understand the actions cache logic here, but this if statement check seems to work as intended.

MatthewFlamm:
This PR now allows to invalidate any particular cache with one (or more) of three labels.  Is this too niche?  Should it be one label that invalidates all of them?

MatthewFlamm:
I'm not so sure that this PR solves all problems with the caching however.  I think it still will require admin's to keep the cache up to date through various means.  However, this PR will allow a future PR to be insulated from the madness of the cache problems, if needed.

banesullivan (approved):

tkoyama010:
This is really useful ! Thanks @MatthewFlamm 🚀

PR #6391, opened 2024-07-17
pytest examples gallery

banesullivan:
A pytest module for running the examples gallery. I'm not sure if we want to include this in CI since we are already running this in the docs build but perhaps?

I use this locally to validate that our examples are not failing with `pytest -v -n 10 test.py`

It would be awesome to use image regression testing for these too.

Sharing as a draft PR, not sure if I'll make the time to finish this


MatthewFlamm:
This approach needs to be compared to and reconciled with #5718.  That approach does the comparison after the doc build.  This is not fool proof, as the documentation build can cache the results.  This was an issue for #6380.  So I see value in having a pytest job.  Best would be if we can use the same image cache for both.

PR #6379, opened 2024-07-14
Add `PlanesAssembly` class adding orthogonal planes to a scene

user27182:
### Overview

This class uses the `OrthogonalPlanesSource` from #6366 to create a new assembly class.

It inherits from `XYZAssembly` and is similar to the [`AxesAssembly`](https://docs.pyvista.org/version/stable/api/plotting/_autosummary/pyvista.AxesAssembly.html#pyvista-axesassembly) class with the following main differences:

1. `AxesAssembly` uses [`AxesGeometrySource`](https://docs.pyvista.org/version/stable/api/utilities/_autosummary/pyvista.AxesGeometrySource.html#pyvista-axesgeometrysource) for its geometry, whereas `PlanesAssembly` uses `OrthogonalPlanesSource`
2. `AxesAssembly` uses the [`Label` ](https://docs.pyvista.org/version/stable/api/plotting/_autosummary/pyvista.Label.html#pyvista.Label) class for its labels, whereas `PlanesAssembly` uses a customized `vtkAxisActor`

Here's a preview of this feature using similar a similar example to the [AnatomicalOrientation](https://examples.vtk.org/site/Cxx/VisualizationAlgorithms/AnatomicalOrientation/) vtk example. Unlike that example, which uses `vtkVectorText` for the plane labels which remain static and will appear "backwards" if viewed from behind the plane, the labels for this new class instead follow the camera.

``` python
import pyvista as pv
from pyvista import examples
human = examples.download_human()
labels = 'Transverse', 'Sagittal', 'Coronal'
planes_assembly = pv.PlanesAssembly(bounds=human.bounds, labels=labels, label_size=25)
plot = pv.Plotter() 
plot.add_mesh(human, color='yellow')
plot.add_actor(planes_assembly) 
planes_assembly.camera = plot.camera 
plot.show()
```

https://github.com/user-attachments/assets/c7631fea-79c1-4515-a794-7e412412e1f0

PR #6394, opened 2024-07-18
Refactor `AxesAssembly` using abstract `_XYZAssembly` class

user27182:
### Overview

Create new abstract class to better define the assembly interface and standardize its initialization. No API changes are added and no tests are modified.

This abstraction is done to support new assembly classes, such as #6379 , which uses a different class/actor for the labels and thus needs to override many of the methods for its implementation.


tkoyama010 (approved):
LGTM

PR #6395, opened 2024-07-19
Preview the Documentation using preview command

tkoyama010:
### Overview

The following features missing from the #5360 have been added.

https://github.com/pyvista/pyvista/blob/142bfd2739ac25bca7f5353f94ce63239313ae18/CONTRIBUTING.rst?plain=1#L1116-L1125

### Details

- https://github.com/pyvista/pyvista/pull/5360#issuecomment-2226159808


tkoyama010:
pre-commit.ci autofix

MatthewFlamm:
How does this work with #6387?

banesullivan:
#6387 handles this

MatthewFlamm:
Is this PR about deploying the documentation to the dev docs or previewing in the PR.  The description makes it sound like the former while the code makes it sound like that latter.  As in my comment https://github.com/pyvista/pyvista/pull/6387#discussion_r1681247086, this naming convention problem is rooted in our CI, where "preview" means both of these things.

banesullivan (changes requested):
In my opinion we should not enable this for 2 reasons:

1. This is a significant security vulnerability. At least in this current implementation anyone could comment `@pyvista-bot preview` and execute a CI routine that has access to our secrets. We could mitigate this by limiting who is allowed to issue that comment, but it's an attack vector I don't want us to be any closer to. (Honestly, I'm starting to think we should reconsider `.github/workflows/auto-approve.yml`... at this time, its limited to a select few individuals and isn't allowed to change without admin approval)
2. We ought to minimize which CI jobs use the `issue_comment` trigger to avoid CI noise since these conditions will evaluate on every singe comment on any issue or PR in the entire repository.

If a contributor is working off of a fork, they can still access the documentation build via a CI artifact. This artifact can simply be downloaded and inspected locally

tkoyama010:
> In my opinion we should not enable this for 2 reasons:
> 
> 1. This is a significant security vulnerability. At least in this current implementation anyone could comment `@pyvista-bot preview` and execute a CI routine that has access to our secrets. We could mitigate this by limiting who is allowed to issue that comment, but it's an attack vector I don't want us to be any closer to. (Honestly, I'm starting to think we should reconsider `.github/workflows/auto-approve.yml`... at this time, its limited to a select few individuals and isn't allowed to change without admin approval)
> 2. We ought to minimize which CI jobs use the `issue_comment` trigger to avoid CI noise since these conditions will evaluate on every singe comment on any issue or PR in the entire repository.
> 
> If a contributor is working off of a fork, they can still access the documentation build via a CI artifact. This artifact can simply be downloaded and inspected locally

Thank you for your comment. Thanks to the api-doc image regression test that @user27182 has added, we now have less need to check the document previews of new contributors. I would like to withdraw this change.

PR #6397, opened 2024-07-20
Define a base class for all geometric sourcesd

tkoyama010:
### Overview

See https://github.com/pyvista/pyvista/pull/5696#issuecomment-1969366087

### Details

- None

### DIFF
--- a/pyvista/core/utilities/geometric_sources.py
+++ b/pyvista/core/utilities/geometric_sources.py
@@ -6,7 +6,6 @@ Also includes some pure-python helpers.
 
 from __future__ import annotations
 
-from abc import ABC
 from enum import IntEnum
 import itertools
 from typing import TYPE_CHECKING
@@ -26,6 +25,9 @@ from pyvista.core.utilities.misc import _check_range
 from pyvista.core.utilities.misc import _reciprocal
 from pyvista.core.utilities.misc import no_new_attr
 
+from .misc import _lazy_vtk_instantiation
+from .misc import abstract_class
+
 if TYPE_CHECKING:  # pragma: no cover
     from typing import Sequence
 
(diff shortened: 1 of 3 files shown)


pyvista/core/utilities/geometric_sources.py
```diff
@@ -40,6 +42,59 @@
...
+class BaseSource(_vtk.vtkAlgorithm):
+    """Superclass for all sources.
+
+    Parameters
+    ----------
+    point_dtype : str, default: 'float32'
+        Set the desired output point types. It must be either 'float32' or 'float64'.
+    """
+
+    def __init__(self, point_dtype='float32'):
```
MatthewFlamm, line 55: Is the default `point_dtype`, `SINGLE_PRECISION` for all sources here?  It seems like in a cursory check, a lot of them are, but if there are any that are `DOUBLE_PRECISION` this choice would change the vtk default, perhaps with poor outcomes.  In that case, it might be better to make the default `None` and defer to the vtk default.

Another possibility is to find all the cases that have `DOUBLE_PRECISION` and override the `__init__` method.  This is more maintenance burden, but is more clean for the documentation and user perspective.

pyvista/core/utilities/geometric_sources.py
```diff
@@ -40,6 +42,59 @@
...
+        """Initialize the geometric source class."""
+        super().__init__()
```
MatthewFlamm, line 57: This class inherits from `vtkAlgorithm` and so does `vtkConeSource`.  Will calling the `__init__` method here clash with the `vtkConeSource` `__init__`, which presumably will also call the `__init__` of `vtkAlgorithm` usage?  This class seems more like a mixin to me rather than needing to inherit from `vtkAlgorithm`.  Unfortunately this makes the typing much harder for python.

PR #6399, opened 2024-07-20
[do not merge] test forked PR docs preview

banesullivan:
do not merge

This is a test to see if #6387 enabled documentation previews from forked PRs

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -317,3 +317,6 @@ Project Index
 *************
 
 * :ref:`genindex`
+
+
+Test!


banesullivan:
Turns out it didn't work. 

https://github.com/pyvista/pyvista/actions/runs/10017494668/job/27692665292?pr=6399#step:3:15

banesullivan:
@tkoyama010, I'm confused how this ever worked in the first place? 

https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

banesullivan:
Ah I see now, this worked because the event was a PR comment, thus coming from this repository.

PR #6400, opened 2024-07-20
Speed up trame examples tests

banesullivan:
The Trame examples tests were previously taking over an hour. Using the `--timeout` argument properly will speed them up to ~1 minute

Additionally some cleanup to make testing of new example automatic moving forward


tkoyama010 (approved):
LGTM

Issue #5506, opened 2024-01-22
It would have been better to make `progress_bar` a keyword only argument everywhere


tkoyama010:
### Describe what maintenance you would like added.

If considering adding new arguments, it would have been better to make `progress_bar` a keyword only argument everywhere.

_Originally posted by @MatthewFlamm in https://github.com/pyvista/pyvista/pull/5502#discussion_r1462236499_

### Links to source code.

None

### Pseudocode or Screenshots

None

user27182:
I agree with enforcing keyword only args, especially for something like `progress_bar`.

I think this applies more broadly to a lot of the code in PyVista. There are many keyword args which are likely never used in practice as positional args, but are technically positional (and therefore would introduce breaking changes if new kwargs are introduced anywhere but at the end of the kwargs dict). Many of these probably should be enforced as keyword only anyway since it can be easy to specify args for a different keyword than intended (and introduce a bug as a result).

For example, all the kwargs in `plot` are positional. This means it's possible to do something like:
``` python
pv.PolyData([0.0,0.0,0.0]).plot(False, False, False, False)
```
But using positional args here is not very helpful here since there are many boolean parameters for `plot`. In addition to eliminating backwards-compatibility issues, enforcing keyword-only args would result in much more explicit code:

``` python
pv.PolyData([0.0,0.0,0.0]).plot(off_screen=False, full_screen=False, screenshot=False, interactive=False)
```

darikg:
I think a number of scientific python libraries went mostly keyword only, [sklearn](https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep009/proposal.html) for example. They have a [decorator](https://github.com/thomasjpfan/scikit-learn/blob/a48f0eafd13ec6cbc4ab6a8b08c92588f9cc8973/sklearn/utils/validation.py#L39) to issue positional deprecation warnings/errors.

MatthewFlamm:
The major point that gives me pause for enforcing this for *all* function args, or a majority, is the breaking change required.  If it can be done with warnings and a deprecation like the example given, then I'm +1 for implementing it.  If we do it, then we should align on what should be positional vs. keyword only and do it within one release cycle.

PR #6407, opened 2024-07-24
Add `categorical` parameter to `point_data_to_cell_data()`

nicolascedilnik:
### Overview

Exposes the underlying option of the VTK filter.

### Details

Not many details to add, maybe forcing point data to be treated as categorical is possible some other way and I haven't found it?


pyvista/core/filters/data_set.py
```diff
@@ -3250,7 +3250,12 @@ def ctp(self, pass_cell_data=False, progress_bar=False, **kwargs):
             **kwargs,
         )
 
-    def point_data_to_cell_data(self, pass_point_data=False, progress_bar=False):
+    def point_data_to_cell_data(
+        self,
+        pass_point_data=False,
+        progress_bar=False,
+        categorical=False,
+    ):
```
tkoyama010, line 3258:
```suggestion
    def point_data_to_cell_data(
        self,
        pass_point_data=False,
        categorical=False,
        progress_bar=False,
    ):
```

pyvista/core/filters/data_set.py
```diff
@@ -3264,6 +3269,11 @@ def point_data_to_cell_data(self, pass_point_data=False, progress_bar=False):
         progress_bar : bool, default: False
             Display a progress bar to indicate progress.
 
+        categorical : bool, default: False
+            Control whether the source point data is to be treated as
+            categorical. If the data is categorical, histograming will be used
+            to assign the cell data.
+
```
tkoyama010, line 3276:
```suggestion
        categorical : bool, default: False
            Control whether the source point data is to be treated as
            categorical. If the data is categorical, histograming will be used
            to assign the cell data.

        progress_bar : bool, default: False
            Display a progress bar to indicate progress.

```

tkoyama010 (changes requested):
Thank you. Please move the `progress_bar`, as it is consistent that it is defined at the end of all methods :)

nicolascedilnik:
> Thank you. Please move the `progress_bar`, as it is consistent that it is defined at the end of all methods :)

Well, looks like you've been faster than me to do that.

tkoyama010:
> > Thank you. Please move the `progress_bar`, as it is consistent that it is defined at the end of all methods :)
> 
> Well, looks like you've been faster than me to do that.

Sorry, I applied it so that we could merge this quickly 🙇

tkoyama010 (dismissed):

user27182:
> Thank you. Please move the `progress_bar`, as it is consistent that it is defined at the end of all methods :)

Isn't this the same issue as #5506? I.e. since `progress_bar` is a positional argument (not keyword-only), this is technically a breaking change.

To be clear I think `progress_bar` _should_ be the last keyword, just trying to understand what the policy is for making this change consistently

pyvista/core/filters/data_set.py
```diff
@@ -3261,6 +3266,11 @@ def point_data_to_cell_data(self, pass_point_data=False, progress_bar=False):
         pass_point_data : bool, default: False
             If enabled, pass the input point data through to the output.
 
+        categorical : bool, default: False
+            Control whether the source point data is to be treated as
+            categorical. If the data is categorical, histograming will be used
```
MatthewFlamm, line 3271: I don't know what "histograming" means here without more context.  In the vtk docs it says "If the data is categorical, then the resultant cell data will be determined by a "majority rules" vote (using a histogram of the point data scalar values), with ties going to the smaller point data value." which is much more understandable to me.  Can we says something like `If ``True``, then the cell data will be determined by "majority rules" vote, with ties going to smaller values.`?

nicolascedilnik: This is how they frame it in the VTK docs:

> [vtkPointDataToCellData](https://vtk.org/doc/nightly/html/classvtkPointDataToCellData.html) is a filter that transforms point data (i.e., data specified per point) into cell data (i.e., data specified per cell). By default, the method of transformation is based on averaging the data values of all the points defining a particular cell. Optionally (by enabling CategoricalData), histograming can be used to assign the cell data. For large datasets with several cell data arrays, the filter optionally supports selective processing to speed up processing. Optionally, the input point data can be passed through to the output as well.

MatthewFlamm: Just a suggestion from me.  Not a blocker.  I personally find the vtk definition in that section not very descriptive.

user27182: > I personally find the vtk definition in that section not very descriptive.

This is also my personal assessment. When I think of histograms I think of *variable* bin sizes, e.g. there are 100 categories and use 5 bins with values [0, 20], [21-40], [41-60], etc. or maybe use 4 bins instead [0, 25], [26-50], etc. to group and visualize the data. But this is not what the categorical option is doing in this case, it just does a frequency count of the point scalars for each cell and keeps the value with the largest count. I found this comment in the vtk [source code]( https://github.com/Kitware/VTK/blob/3866322a8bea79a5dc6c99ffe312a8cb76326073/Filters/Core/vtkPointDataToCellData.cxx#L279-L280) more helpful than the docs for describing what "histograming" means:
> // Populate a histogram from the scalar values at each
> // point, and then select the bin with the most elements.

But, I still think it's best if we avoid using the term "histograming". I liked @MatthewFlamm's suggestion which is clear and succinct:

> If ``True``, then the cell data will be determined by "majority rules" vote, with ties going to smaller values.

user27182: > When I think of histograms I think of _variable_ bin sizes, e.g. there are 100 categories and use 5 bins with values [0, 20], [21-40], [41-60], etc. or maybe use 4 bins instead [0, 25], [26-50], etc. to group and visualize the data. But this is not what the categorical option is doing in this case, it just does a frequency count of the point scalars for each cell and keeps the value with the largest count. 

As a follow-up, this method *does* seem to group values together, and not just do a simple frequency count ([source code](
https://github.com/Kitware/VTK/blob/3866322a8bea79a5dc6c99ffe312a8cb76326073/Filters/Core/vtkPointDataToCellData.cxx#L192-L193)):
``` c++
// If the adjacent bins are close enough to be merged...
if (std::abs((*it1).Value - (*it2).Value) < VTK_EPSILON)
```

with 
``` c++
#define VTK_EPSILON 1.e-6
```

But, this makes for a very narrow bin, and only really applies to continuous scalars. For actual categorical data, integers are typically used, in which case this is effectively a majority rules without any real "binning" or grouping in the histogram sense I described earlier.

MatthewFlamm: > As a follow-up, this method _does_ seem to group values together, and not just do a simple frequency count ([source code](https://github.com/Kitware/VTK/blob/3866322a8bea79a5dc6c99ffe312a8cb76326073/Filters/Core/vtkPointDataToCellData.cxx#L192-L193)):

Thanks for the in depth analysis.  I always thought the "histograming" comment was odd, but from the source code, it is the most accurate when including float-valued data.  I still like my original suggestion, as users should be avoid using this flag with float data.  It also makes the process more clear what happens for the more suitable integer-valued data.  But, I also don't feel strongly about this, since the original wording is technically correct and is what vtk uses.

pyvista/core/filters/data_set.py
```diff
@@ -3261,6 +3266,11 @@ def point_data_to_cell_data(self, pass_point_data=False, progress_bar=False):
...
+            to assign the cell data.
+
```
user27182, line 3273: In response to this thread https://github.com/pyvista/pyvista/pull/6407#discussion_r1691956369, how about adding this explanation? Any thoughts on this change @nicolascedilnik ?

```suggestion
        categorical : bool, default: False
            Control whether the source point data is to be treated as
            categorical. If ``True``,  histograming is used to assign the 
            cell data. Specifically, a histogram is populated for each cell
            from the scalar values at each point, and the bin with the most
            elements is selected. In case of a tie, the smaller value is selected.
            
            .. note::
             
                If the point data is continuous, values that are almost equal (within 
                ``1e-6``) are merged into a single bin. Otherwise, for discrete data
                the number of bins equals the number of unique values.

```

user27182 (dismissed):
LGTM, thanks for the contribution!

Changing the arg order of `progress_bar` is a very low-risk breaking change which need not block this.

And I'm also fine with or without changes to the parameter description.

pyvista/core/filters/data_set.py
truenicoco: It's fine with me, of course!

tkoyama010: @user27182 Could you commit this proposal before we forget?

user27182 (approved):

MatthewFlamm:
pre-commit.ci autofix

MatthewFlamm (approved):

tkoyama010 (approved):
LGTM

PR #6389, opened 2024-07-17
Add `Prop3DMixin` to `AxesAssembly` class

user27182:
### Overview

Make `AxesAssembly` inherit from  `Prop3DMixin` to remove redundant methods and documentation.

PR #6412, opened 2024-07-25
Add `bounds`, `center`, and `length` to `Prop3DMixin`

user27182:
### Overview

Follow-up to #6394 and #6389 to further abstract prop3d properties from `AxesAssembly` into `Prop3DMixin`.


tkoyama010 (approved):
LGTM

PR #6414, opened 2024-07-25
Use the term `assembly` instead of `actor`  in `AxesAssembly` docs and tests

user27182:
### Overview

Parts of the code for the `AxesAssembly` class refer to axes actor class which should be updated.


tkoyama010 (approved):
LGTM

PR #6415, opened 2024-07-25
Use rgb scalars in `download_human` example

user27182:
### Overview

This dataset has built-in rgb scalars (see this [vtk example](https://examples.vtk.org/site/Cxx/VisualizationAlgorithms/AnatomicalOrientation/)). This PR makes use of them for the example's plot.


tkoyama010 (approved):
LGTM

PR #6418, opened 2024-07-26
Make `Prop3DMixin.bounds` abstract

user27182:
### Overview

#6412 added `bounds` to `_Prop3DMixin` with a concrete implementation, but this property should be abstract since the bounds rely on a transformation applied by the abstract method `_post_set_update`.

The current implementation for `bounds` was actually  based on two separate concrete implementations for the `_XYZAssembly` and `Label` classes, so this PR splits and moves the implementation to those classes. To avoid the need for subclasses to re-write or re-wrap the docs for `bounds` , a new `_get_bounds` method is used for the abstraction and called internally instead of making the `bounds`  property itself abstract.


tests/plotting/test_text.py
```diff
@@ -72,6 +72,12 @@ def test_label_prop3d():
     assert label.center == position
     assert label.length == 0.0
 
+    # Test correct bounds with more complex transformations
+    label.relative_position = (100, 0, 0)
+    label.orientation = (0, 0, 90)
+    bounds = (1.0, 1.0, 102.0, 102.0, 3.0, 3.0)
+    assert np.allclose(label.bounds, bounds)
```
tkoyama010, line 79: Please add a comment explaining why the `bounds` have this value.

tkoyama010 (changes requested):

tests/plotting/test_text.py
user27182, line 79: Is this better?

```suggestion
    # Test correct bounds with more complex transformations
    # Add offset along x-axis
    offset = 100
    label.relative_position = (offset, 0, 0)
    # Rotate about z-axis
    label.orientation = (0, 0, 90)
    # Expect offset to be applied along y-axis (due to the rotation) 
    bounds = (position[0], position[0], position[1] + offset, position[1] + offset, position[2], position[2])
    assert np.allclose(label.bounds, bounds)
```

tkoyama010 (approved):
LGTM

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

PR #6417, opened 2024-07-26
Make attribute arguments consistent for `Prop3D` and `Prop3DMixin`

user27182:
### Overview

Update the typing and testing to ensure these two classes exhibit the same behavior.

PR #6420, opened 2024-07-26
Remove `_coerce_transformlike_arg` and use `validate_transform4x4` instead

user27182:
### Overview
Follow-up to #6417.

Replace `_coerce_transformlike_arg` with `validate_transform4x4`. It  was only used by `Prop3D.user_matrix`. This change broadens the kinds of inputs allowed when setting `user_matrix`. In particular, it allows setting with nested lists.


pyvista/plotting/prop3d.py
```diff
@@ -370,8 +366,10 @@ def user_matrix(self) -> NumpyArray[float]:  # numpydoc ignore=RT01
         return array_from_vtkmatrix(self.GetUserMatrix())
 
     @user_matrix.setter
-    def user_matrix(self, value: TransformLike):  # numpydoc ignore=GL08
-        array = np.eye(4) if value is None else _coerce_transformlike_arg(value)
+    def user_matrix(
+        self, value: TransformLike | scipy.spatial.transform.Rotation
```
tkoyama010, line 370: I'd like to make a comment that is a little off-topic from this PR discussion. Should `scipy.spatial.transform.Rotation` be integrated into `TransformLike`? Or does it complicate the code?

tkoyama010 (approved):
LGTM

pyvista/plotting/prop3d.py
user27182: It probably should be part of `TransformLike`.

A related issue is how to deal with the import, similar to https://github.com/pyvista/pyvista/pull/6354#discussion_r1673855824.
Adding it to the aliases in `typing_core` would make it a runtime import, which would also fix the static import issue for `scipy`.

Should we define a new `RotationLike` alias for the 3x3 inputs, and make it a subset of `TransformLike`?

tkoyama010: +1 for adding `RotationLike` alias for the 3x3 inputs. If creating a PR not takes your time, then let's create one and see how other developers react.

PR #6421, opened 2024-07-26
Add `push()` to `PlaneSource` and `OrthogonalPlanesSource` to translate plane along normal

user27182:
### Overview

The planes generated by `OrthogonalPlanesSource` are geometrically centered by default. But this may not coincide with some other definition of center. `move` makes it easy to adjust the planes accordingly to better "center" them on a mesh, for example.

EDIT: Note that `vtkPlaneSource` has `Push` and `Pull` methods to accomplish something similar, but during tests this was found not to work properly (e.g. `source.Push(1)` did not translate the plane by `1`).  These methods also translate in the direction of the plane's normal, whereas `move` does not (it's always w.r.t. positive x-y-z axes). And so the term `move`  is used instead of `push` to avoid confusion since it has different behaviour from `Push`.


pyvista/core/utilities/geometric_sources.py
```diff
@@ -3667,13 +3718,15 @@ def update(self):
...
-            zip(self.names, self._plane_sources, output, self.normal_sign)
+        move_vectors = np.diag(self.move)
+        for index, (name, source, plane, sign, move) in enumerate(
+            zip(self.names, self._plane_sources, output, self.normal_sign, move_vectors)
         ):
             plane.copy_from(source.GetOutput())
             if sign == '-':
                 plane['Normals'] *= -1
             output.set_block_name(index, name)
+            plane.points += move
```
tkoyama010, line 3729: How about make `_plane_sources` accessible so that the points can be moved? It is more intuitive in my mind.

tkoyama010 (changes requested):

pyvista/core/utilities/geometric_sources.py
user27182: I don't mind exposing the `plane_sources`, but they don't have a `points` property (only the output does), so we can't modify the points of a `PlaneSource` directly. We _can_ add `PlaneSource.push` and/or `PlaneSource.pull` methods (e.g. similar to [vtkPlaneSource.Push()](https://vtk.org/doc/nightly/html/classvtkPlaneSource.html#a2af1869e467d1be22d4f43141c69315e)) but I tried using `Push()` in earlier implementations but I found this didn't actually work as expected. Maybe we can "fix" it as part of the `PlaneSource.push` implementation? This would effectively move the translation to `PlaneSource.update()` rather than `OrthogonalPlaneSource.update()`. Should I try adding this feature to `PlaneSource` ?

tkoyama010: I'm sorry. That was my misunderstanding. You are right, source and output are two different things. I agree that there needs to be a method here. One more point I would like to confirm is your opinion about renaming move to translate. The method name [`translate`](https://docs.pyvista.org/api/core/_autosummary/pyvista.dataset.translate) is used in other parts of PyVista. If what this method does is the same as it, it would be better to rename the method to `translate`.

user27182: > I'm sorry. That was my misunderstanding. You are right, source and output are two different things. I agree that there needs to be a method here. I'll One more point I would like to confirm is your opinion about renaming move to translate. The method name [`translate`](https://docs.pyvista.org/api/core/_autosummary/pyvista.dataset.translate) is used in other parts of PyVista. If what this method does is the same as it, it would be better to rename the method to `translate`.

It's similar to `translate`, in that a translation is applied, except with `move` (or `push`/`pull`) only the magnitude is given. The direction is implied by the normal of the plane(s). `move(1,2,3)` moves the first plane a distance of 1, the second plane a distance of 2, the third plane a distance of 3. I think calling it `translate` would be misleading. But `move` isn't great either.

Perhaps implementing both `PlaneSource.push` and `PlaneSource.pull` is best. These names match the vtk methods.

user27182: This should be resolved. `push` is now implemented instead of `move`.

tkoyama010 (approved):
LGTM

PR #6422, opened 2024-07-27
Add type annotations to `check_less_than` and update variable names

user27182:
### Overview
Follow-up from https://github.com/pyvista/pyvista/pull/5571.

The `value` is also validated in the same manner as done in `check_greater_than`.


tkoyama010 (dismissed):
LGTM

pyvista/core/_validation/check.py
```diff
@@ -374,6 +374,13 @@ def check_nonnegative(array: _ArrayLikeOrScalar[NumberType], /, *, name: str = "
     check_greater_than(array, 0, strict=False, name=name)
 
 
+def _validate_real_finite_value(scalar, name='Value'):
```
user27182, line 377: This used to `check_finite` but was removed, the name should reflect this
```suggestion
def _validate_real_value(scalar, name='Value'):
```

tkoyama010 (dismissed):

tkoyama010 (approved):

PR #6423, opened 2024-07-27
Add type annotations to `check_range` and update variable names

user27182:
### Overview

Follow-up from https://github.com/pyvista/pyvista/pull/5571.

A duplicate `TypeVar` is also added to support multiple numeric type args.


pyvista/core/_typing_core/_array_like.py
```diff
@@ -39,6 +39,13 @@
 )
 NumberType.__doc__ = """Type variable for numeric data types."""
 
+# Create a copy of the typevar which can be used for annotating a second variable.
+# Its definition should be identical to `NumberType`
+_NumberType = TypeVar(  # noqa: PYI018
+    '_NumberType',
+    bound=Union[np.floating, np.integer, np.bool_, float, int, bool],  # type: ignore[type-arg]
+)
+
```
tkoyama010, line 48: Is `bool` necessary for this type?  I feel a little uncomfortable with the name `_NumberType.`

tkoyama010 (changes requested):

pyvista/core/_typing_core/_array_like.py
user27182: Yea this is where things get a bit messy. `bool` inherits from `int` but `np.bool_` does not inherit from `np.integer` (not even `np.number`). So a bool is both a number and also not a number. 

I recall having issues trying to set the `TypeVar` correctly. It's currently defined to allow subtypes of the union members, e.g. `np.uint8` is allowed since `np.integer` is included in the definition. But then if we have `np.integer` and `int` without any bool types this means that `bool` is allowed (subclass of `int`), but `np.bool_` is not (not a subclass of `np.integer`). Maybe we're OK with that? I.e. allowing `bool` but not `np.bool_` since anything typed as a single `int` already implies that `bool` inputs are ok.  

Alternatively, one way around this is to make the `TypeVar` invariant and only allow the specific types listed, and do not allow subclasses e.g. 
``` python 
NumberType = TypeVar('NumberType', np.float64, np.float32, np.int_, np.uint8, float, int).
```
But then we need to explicitly include all of the `floating` and `integer` types defined by NumPy that we want to include, maybe that's all of them (large list) or maybe we limit the types to small subset (e.g. as above).

user27182: The `TypeVar` was recently updated in #6216.
``` diff

- NumberType = TypeVar('NumberType', bool, int, float, np.bool_, np.int_, np.float64, np.uint8)
+ NumberType = TypeVar('NumberType', bound=Union[np.floating, np.integer, np.bool_, float, int, bool])
```
I can't recall exactly why but I remember having problems (i.e. `mypy` errors) with getting this definition right while adding all the typing to the `validation` package.

Can we come back to this issue later once more of the typing updates from #5571 are merged? I'll note that this PR simply duplicates the existing definition and otherwise isn't making any changes to the `TypeVar`.

tkoyama010: Great explanation thanks! I just want to confirm, can we merge this PR as is or you need to add your change in this PR?

user27182: This is good to go as-is 👍

tkoyama010 (approved):

PR #6431, opened 2024-07-28
Add `RotationLike` type alias to the docs

user27182:
### Overview

Add new type alias to the docs and update `TransformLike` as suggested in https://github.com/pyvista/pyvista/pull/6420#discussion_r1693894103.

EDIT: Note that the branch name should be `doc/rotationlike_alias` (not `boundslike`)


pyvista/core/_validation/validate.py
```diff
@@ -544,7 +544,7 @@ def validate_transform3x3(transform, /, *, name="Transform"):
 
     Parameters
     ----------
-    transform : array_like | vtkMatrix3x3 | scipy.spatial.transform.Rotation
+    transform : RotationLike
```
user27182, line 547: `RotationLike` here implies that only 3x3 rotation matrices are supported. But any 3x3 matrix is supported... not sure how to communicate this, maybe just a note in the docs?

user27182:
pre-commit.ci autofix

user27182:
The new alias is properly configured 👍. It is automatically linked in docstrings, e.g.: https://66a6ca76975f4c9c9c977a9a--pyvista-dev.netlify.app/api/plotting/_autosummary/pyvista.prop3d.rotation_from

user27182:
pre-commit.ci autofix

pyvista/core/_validation/validate.py
tkoyama010: Yes, I think that will get the message to users.

tkoyama010 (changes requested):

user27182:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM

PR #6435, opened 2024-07-29
Add `CubeFacesSource` for generating a `MultiBlock` with named faces

user27182:
### Overview

This class inherits from `CubeSource` but modifies the output to generate a `MultiBlock` instead of `PolyData`. 

This class will be used to create a new `LabeledCubeAssembly` class, where each face is labeled.


pyvista/core/utilities/geometric_sources.py
```diff
@@ -3694,3 +3694,473 @@ def output(self) -> pyvista.MultiBlock:
...
+        z_len = np.linalg.norm(bnds[5] - bnds[4])
+        lengths = np.array((x_len, y_len, z_len))
+        min_length = np.min(lengths)
+
+        # Store vars for updating the output
+        shrink_factor, explode_factor, frame_width = (
+            self.shrink_factor,
+            self.explode_factor,
+            self.frame_width,
+        )
```
tkoyama010, line 4114:
```suggestion
        shrink_factor = self.shrink_factor
        explode_factor = self.explode_factor
        frame_width = self.frame_width
```

tkoyama010 (changes requested):

tkoyama010 (dismissed):
LGTM. Thanks!

tests/typing/test_bounds_tuple.py
```diff
@@ -66,7 +67,9 @@ def pytest_generate_tests(metafunc):
             'UnstructuredGridVolumeRayCastMapper',
             'Volume',
         ]
-        assert sorted(class_names) == sorted(expected_names)
+        diff = set(class_names) - set(expected_names)
+        if diff:
+            pytest.fail(f"The `bounds` property for class {diff}\n must return a {pv.BoundsTuple}.")
```
user27182, line 72: This change is not correct. The aim was to improve the error message, it should say something like:
``` python
f"Class {diff} has a `bounds` property but is not listed as a test case." 
```
I suggest reverting this and leaving this change for another PR.

```suggestion
        assert sorted(class_names) == sorted(expected_names)
```

user27182:
The merge with main https://github.com/pyvista/pyvista/pull/6435/commits/23119c254dd6db434d2b0ec75dc86273a4b4a437 failed because `CubeFacesSource` has bounds which is now checked by the tests from #6454. So, `'CubeFacesSource'` just needed to be added to the list of test cases (`CubeFacesSource.bounds` already returns a `BoundsTuple` since it inherits from `CubeSource`).

user27182:
> The merge with main [23119c2](https://github.com/pyvista/pyvista/commit/23119c254dd6db434d2b0ec75dc86273a4b4a437) failed because `CubeFacesSource` has bounds which is now checked by the tests from #6454. So, `'CubeFacesSource'` just needed to be added to the list of test cases (`CubeFacesSource.bounds` already returns a `BoundsTuple` since it inherits from `CubeSource`).

No need to add `CubeFacesSource` as a test case anymore since #6470 was merged.

tkoyama010 (approved):

Issue #4736, opened 2023-08-03
Definitions of 'closed', 'manifold', and 'boundary' relating to `vtkFeatureEdges` are used inconsistently


user27182:
### Describe the feature you would like to be added.

This is kind of a bug, feature request, and maintenance request all in one! The exact classification will depend on the interpretation / definition of what a manifold or closed surface is what its boundaries are. I mainly want to focus on the manifold/closed stuff, but I touch on some of the boundary stuff too since it's related (though secondary).

### Bug
It appears as though `is_manifold` is currently doing more than checking for manifoldness (at least in the vtk sense), and is actually checking whether a surface is closed (see code/explanation below).

### Feature Request
Add a new property, `is_closed`, which performs this check. `is_manifold` will need to be redefined accordingly. Code for `select_enclosed_points` can also be updated to explicitly use the `is_closed` property.

### Maintenance

- Improve the docs for [`is_manifold`](https://docs.pyvista.org/version/0.41/api/core/_autosummary/pyvista.PolyData.is_manifold.html) so that it very clearly states what is meant, and not meant, by the term 'manifold'. For example, the definition of manifoldness can vary depending on the field or software being used (e.g. meshes with self-intersections may be considered non-manifold, there is also vertex- and edge-manifoldness), so being very clear about the specific pyvista/vtk definition is very helpful.
- Improve the docs for [`select_enclosed_points`](https://docs.pyvista.org/version/0.41/api/core/_autosummary/pyvista.DataSetFilters.select_enclosed_points.html), as it seems to use the terms 'manifold' and 'closed' interchangeably / inconsistently. Also, add a suggestion for the user to use the `is_closed` property before using the method so they can perform a check manually before calling instead of doing `try`/`except`

### Related boundary stuff
The term 'boundary' is also used inconsistently and it is sometimes unclear what is meant by this.
- E.g. the [extract edges example](https://docs.pyvista.org/version/0.41/examples/01-filter/extract-edges.html) with the bunny has the function call
```python
edges = mesh.extract_feature_edges(boundary_edges=True, feature_edges=False, manifold_edges=False)
```
- This is misleading because this makes it look like only boundary edges are being plotted. However, since `non_manifold_edges=True` by default, the viewer is actually being shown all open edges, including non-manifold ones and not just boundary edges (though in this case it's possible that there aren't any non-manifold edges).
[...]

PR #6450, opened 2024-07-31
Add Example of creating mesh multiple ways

MatthewFlamm:
### Overview

There have been several requests that could be solved by users forming their own mesh from scratch.  This PR adds in a more complex example to show how to form a mesh 'by hand' with some useful tips on how to visualize the connectivity ordering along the way using `add_point_labels`.

### Details

This is one example https://github.com/pyvista/pyvista/discussions/6416, here the user already has code to generate the points but is looking for how to make it into a mesh with cells (I think this is the request).  There was another request on Slack.

We could consider making this a separate 'sphere' geometric object, but I would suggest this is a separate PR as it would require a lot of careful edge case handling like in `SolidSphere` from #4965 .


examples/00-load/create-poly.py
```diff
@@ -80,5 +80,148 @@
...
+
+points_to_label = tuple(range(ntheta + 1))
+mesh = pv.PolyData(points, faces=faces)
+pl = pv.Plotter(off_screen=True)
+pl.add_mesh(mesh, show_edges=True)
+pl.add_point_labels(
+    mesh.points[points_to_label, :], points_to_label, font_size=30, fill_shape=False
+)
+pl.view_xy()
+pl.show(screenshot="tmp.png")
```
tkoyama010, line 161:
```suggestion
pl.show()
```

examples/00-load/create-poly.py
```diff
@@ -80,5 +80,148 @@
...
+pl.add_mesh(mesh, show_edges=True)
+pl.add_point_labels(
+    mesh.points[points_to_label, :],
+    points_to_label,
+    font_size=30,
+    fill_shape=False,
+    always_visible=True,
+)
+pl.view_xy()
+pl.show(screenshot="tmp2.png")
```
tkoyama010, line 191:
```suggestion
pl.show()
```

examples/00-load/create-poly.py
```diff
@@ -80,5 +80,148 @@
...
+
+faces.extend([3, nphi * ntheta + 1, nphi * ntheta, (nphi - 1) * ntheta + 1])
+
+# %%
+# All the point labels are messy when plotted, so don't add to the final plot.
+
+mesh = pv.PolyData(points, faces=faces)
+pl = pv.Plotter()
+pl.add_mesh(mesh, show_edges=True)
+pl.show(screenshot="tmp3.png")
```
tkoyama010, line 224:
```suggestion
pl.show()
```

tkoyama010 (changes requested):
Thanks!

examples/00-load/create-poly.py
```diff
@@ -80,5 +80,148 @@
...
+    faces.extend([3, 0, i, i + 1])
+
+faces.extend([3, 0, ntheta, 1])
+
+# %%
+# Demonstrate the connectivity of the mesh so far.
+
+points_to_label = tuple(range(ntheta + 1))
+mesh = pv.PolyData(points, faces=faces)
+pl = pv.Plotter(off_screen=True)
```
MatthewFlamm, line 155:
```suggestion
pl = pv.Plotter()
```

examples/00-load/create-poly.py
```diff
@@ -80,5 +80,148 @@
...
+    faces.extend([4, i, i + 1, i + ntheta + 1, i + ntheta])
+
+faces.extend([4, ntheta, 1, ntheta + 1, ntheta * 2])
+
+# %%
+# Demonstrate the connectivity of the mesh with first quad layer.
+
+points_to_label = tuple(range(ntheta * 2 + 1))
+mesh = pv.PolyData(points, faces=faces)
+pl = pv.Plotter(off_screen=True)
```
MatthewFlamm, line 181:
```suggestion
pl = pv.Plotter()
```

tkoyama010 (dismissed):
LGTM

MatthewFlamm:
One thought I had is whether this would be a better as part of a standalone example contrasting multiple ways to construct different kinds of a sphere mesh, e.g. by using `pv.Sphere`  by using `pv.SolidSphere`, by using `pv.RectilinearGrid`, etc.

tkoyama010:
> One thought I had is whether this would be a better as part of a standalone example contrasting multiple ways to construct different kinds of a sphere mesh, e.g. by using `pv.Sphere` by using `pv.SolidSphere`, by using `pv.RectilinearGrid`, etc.

Good idea. Would you like to add it to this Pull Request?

MatthewFlamm:
I'm going to make this draft while I think about it.  Plus there was this recent discussion https://github.com/pyvista/pyvista/discussions/6453.

examples/00-load/create-poly.py
```diff
@@ -80,5 +80,7 @@
 
 mesh = pv.PolyData(points, faces)
 mesh.plot(show_edges=True, line_width=5)
+
```
MatthewFlamm, line 83:
```suggestion
```

examples/00-load/create-poly.py
```diff
@@ -80,5 +80,6 @@
...
```
MatthewFlamm, line 83:
```suggestion
```

MatthewFlamm:
I have revamped this PR to be a standalone example that shows how to create a Sphere mesh multiple ways.

1. using builtin `pv.Sphere`
2. making a `StructuredGrid` that has degenerate points/cells/seams
3. making a `PolyData` mesh that is seamless without degenerate points/cells

MatthewFlamm:
Thanks for uploading the final thumbnail for this.  I think it is ready for final review.

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+phi = np.linspace(0, np.pi, nphi)
+
+r_, phi_, theta_ = np.meshgrid([radius], phi, theta, indexing='ij')
+x, y, z = pv.spherical_to_cartesian(r_, phi_, theta_)
+mesh = pv.StructuredGrid(x, y, z)
+
+# %%
+# The mesh has quadrilateral cells. The cells that look triangular
+# at the poles are actually degenerate quadrilaterals, i.e. two
+# points are coincident the pole, as will be shown later.
```
MatthewFlamm, line 64:
```suggestion
# points are coincident at the pole, as will be shown later.
```

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+
+"""
+# sphinx_gallery_thumbnail_number = 5
+
+# %%
+from __future__ import annotations
+
+import numpy as np
+
+import pyvista as pv
```
user27182, line 17: This creates an empty block, not sure if there's a way to format it differently. 
image

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+# Simple Sphere
+# ~~~~~~~~~~~~~
+# The quickest method to get a Sphere mesh is to use :func:`pyvista.Sphere`.
+
+mesh = pv.Sphere()
+mesh.plot(show_edges=True)
+
+# %%
+# This gives an :class:`pyvista.PolyData` mesh, i.e. a 2D surface. In this case, it is watertight
+# and encloses a volume.
```
user27182, line 29: Is watertight the same as `is_manifold` here? https://dev.pyvista.org/api/core/_autosummary/pyvista.polydata.is_manifold#pyvista.PolyData.is_manifold
Not sure if it's helpful to cross-ref this, though this property itself is a bit ambiguously defined, see
https://github.com/pyvista/pyvista/issues/4736

Is it overkill to also show that its `volume` is zero (since it's a 2D surface)?

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+
+mesh
+
+# %%
+# The cells are triangles. For example, the first cell
```
user27182, line 34: Add cross-ref
```suggestion
# The cells are :attr:`~pyvista.CellType.TRIANGLE` cells.. For example, the first cell
```

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+nphi = 12
+theta = np.linspace(0, 2 * np.pi, ntheta)
+phi = np.linspace(0, np.pi, nphi)
+
+r_, phi_, theta_ = np.meshgrid([radius], phi, theta, indexing='ij')
+x, y, z = pv.spherical_to_cartesian(r_, phi_, theta_)
+mesh = pv.StructuredGrid(x, y, z)
+
+# %%
+# The mesh has quadrilateral cells. The cells that look triangular
```
user27182, line 62: Add cross-ref.
```suggestion
# The mesh has :attr:`~pyvista.CellType.QUAD` cells. cells. The cells that look triangular
```

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+
+mesh.plot(show_edges=True)
+
+# %%
+# The mesh is of type :class:`pyvista.StructuredGrid`.
+
+mesh
+
+# %%
+# The first cell is at the top pole, and it is a Quadrilateral.
```
user27182, line 74: Maybe add the quad cross-ref here instead of above (or keep both?)
```suggestion
# The first cell is at the top pole, and it is a :attr:`~pyvista.CellType.QUAD` cell.
```

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+
+# %%
+# The first cell has two degenerate points.
+
+cell.points
+
+# %%
+# The cells on either side of the 'seam' along the start and end of
+# the azimuthal component are not connected. Visualize this by plotting
+# the boundary edges of the mesh.
```
user27182, line 87: In addition to the figure, if showing the `is_manifold` property from `Sphere` above (expected True), maybe show `is_manifold` here too (expected False)

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+pl.show()
+
+# %%
+# Generate quadrilateral mesh of Sphere
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# This example shows how a more complicated mesh can be defined.
+#
+# In contrast to the example above, this example generates a mesh
+# that does not have degenerate points at the poles. Triangle cells
+# will be used at the poles.  First, regenerate the structured data.
```
user27182, line 105: Though linked earlier, sometimes it's helpful to add links in multiple locations in case only this section is linked/referenced.

```suggestion
# that does not have degenerate points at the poles. :attr:`~pyvista.CellType.TRIANGLE` cells
# will be used at the poles.  First, regenerate the structured data.
```

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+radius = 0.5
+ntheta = 9
+nphi = 12
+theta = np.linspace(0, 2 * np.pi, ntheta)
+phi = np.linspace(0, np.pi, nphi)
+
+# %%
+# We do not want duplicate points, so remove the duplicate in theta, which
+# results in 8 unique points in theta. Similarly, the poles at `phi=0` and
+# `phi=pi` will be handled separately to avoid duplicate points, which
```
user27182, line 116: Format as literals instead of italics
```suggestion
# results in 8 unique points in theta. Similarly, the poles at ``phi=0`` and
# ``phi=pi`` will be handled separately to avoid duplicate points, which
```

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+# results in 10 unique points in phi.  Remove these from the grid in spherical
+# coordinates.
+
+theta = theta[:-1]
+ntheta -= 1
+phi = phi[1:-1]
+nphi -= 2
+
+# %%
+# Generate cartesian coordinates for points in the ``(N, 3)``
```
user27182, line 126: Add cross-ref

```suggestion
# Use :func:`pyvista.spherical_to_cartesian` to generate cartesian coordinates for points in the ``(N, 3)``
```

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,239 @@
...
+
+# %%
+# Structured quadrilateral mesh of Sphere
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# The structure of the mesh can be important. Instead of a
+# triangulated mesh, it can be useful to have a structured
+# mesh that has an i-j-k ordering that allows for simplified
+# cell connectivity.
+#
+# The points are generated as a regular grid in spherical coordinates.
```
user27182, line 46: Add cross-ref 
```suggestion
# The points are generated as a regular grid in spherical coordinates using  :func:`pyvista.spherical_to_cartesian`.
```

user27182 (commented):
Looks good, I have some minor edits and suggestions. Might also be helpful to add a link to this example in the See Also section of `Sphere`

examples/00-load/create-sphere.py
MatthewFlamm: I think it does mean the same thing. Good point, I will update the language. Linking is a good idea too, but I'm on the fence based on the following:

The only 2D surface that has a valid volume (in vtk/PyVista) is a manifold PolyData mesh. `pv.Sphere()` has a volume, but `pv.Sphere().cast_to_unstructured_grid()` does not have a volume.  The `StructuredGrid` example that is later also does not have a volume.  `is_manifold` also only exists for PolyData meshes.  These subtleties are important but are better IMO in an example that compares the different mesh types rather than an example that shows different ways to form those meshes.

user27182: Ok, I'm fine with not including any links or comparing/discussing volume. These were partly just my own curiosities.

examples/00-load/create-sphere.py
user27182: It's fine to omit this too.

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,261 @@
...
+
+r_, phi_, theta_ = np.meshgrid([radius], phi, theta, indexing='ij')
+x, y, z = pv.spherical_to_cartesian(r_, phi_, theta_)
+points = np.vstack((x.ravel(), y.ravel(), z.ravel())).transpose()
+
+# %%
+# The first and last points are the poles.
+
+points = np.insert(points, 0, [0.0, 0.0, radius])
+points = np.append(points, [0.0, 0.0, -radius])
```
user27182, line 150: This change is needed to ensure `points` keeps its shape `(N, 3)`. Otherwise, these methods flatten the input.
```suggestion
points = np.insert(points, 0, [0.0, 0.0, radius], axis=0)
points = np.append(points, [[0.0, 0.0, -radius]], axis=0)
```

Passing a 1D flat array was previously allowed, and this flattened array was reshaped to `(-1,3)` automatically by `pyvista.vtk_points`. This was changed in https://github.com/pyvista/pyvista/pull/6876. And so, this example now fails and `vtk_points` complains that the input is not a single point or a 2D array of points.

From reading this example (and not knowing that `insert` and `append` will flatten by default), I _thought_ this code was trying to keep the array as `(N,3)`. In that case, the suggestion above is good. But, if the expected use-case for `vtk_points` is to pass it a flat array of points and have it reshape it, then we should probably modify `vtk_points` to allow this again.

My understanding of this, however, is that `vtk_points` should only allow either a _single_ point (1D array) or multiple points (2D array). Indeed, there are no tests in the code checking that flattened point arrays are acceptable input for this function.

MatthewFlamm: Your suggestion is a good one. I find the differences for `np.insert` and `np.append` to be confusing in this case, but maybe it is my mental block.  On first thought, I'm +1 for requiring points to be (N,3).  If it is flat, then it is ambiguous which order it should be reshaped.

examples/00-load/create-sphere.py
user27182: fixed! 
image

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,261 @@
...
+# encloses a volume. To demonstrate this, there are no boundaries on the mesh
+# as indicated by no points/cells being extracted.
+
+boundaries = mesh.extract_feature_edges(
+    non_manifold_edges=True, feature_edges=False, manifold_edges=False
+)
+boundaries
+
+# %%
+# The cells are :attr:`~pyvista.CellType.TRIANGLE` cells.. For example, the first cell
```
user27182, line 42:
```suggestion
# The cells are :attr:`~pyvista.CellType.TRIANGLE` cells. For example, the first cell
```

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,261 @@
...
+nphi = 12
+theta = np.linspace(0, 2 * np.pi, ntheta)
+phi = np.linspace(0, np.pi, nphi)
+
+r_, phi_, theta_ = np.meshgrid([radius], phi, theta, indexing='ij')
+x, y, z = pv.spherical_to_cartesian(r_, phi_, theta_)
+mesh = pv.StructuredGrid(x, y, z)
+
+# %%
+# The mesh has :attr:`~pyvista.CellType.QUAD` cells. cells. The cells that look triangular
```
user27182, line 70:
```suggestion
# The mesh has :attr:`~pyvista.CellType.QUAD` cells. The cells that look triangular
```

examples/00-load/create-sphere.py
```diff
@@ -0,0 +1,261 @@
...
+    mesh.points[points_to_label, :], points_to_label, font_size=30, fill_shape=False
+)
+pl.view_xy()
+pl.show()
+
+# %%
+# Next form the quadrilaterals. This process is the same except
+# by connecting points across two levels of phi.  For point 1
+# and point 2, these are connected to point 9 and point 10. Note
+# for quadrilaterals it must be defined in a consistent direction.
```
user27182, line 189: Format as literals (which are similarly used above)
```suggestion
# by connecting points across two levels of ``phi``.  For point ``1``
# and point ``2``, these are connected to point ``9`` and point ``10``. Note
# for quadrilaterals it must be defined in a consistent direction.
```

user27182 (approved):
LGTM, minor formatting/typos

tkoyama010 (approved):

tkoyama010:
Thanks @MatthewFlamm !

MatthewFlamm:
Thanks for the great reviews.  This example is much better as a result.

PR #6454, opened 2024-08-02
Replace type alias `BoundsLike` with named tuple `BoundsTuple`

user27182:
### Overview

Use a named tuple to make accessing bounds more explicit. This makes code easier to read and understand.

This PR also standardizes the bounds representation as `x_min` rather than `xmin` or `xMin`, etc.

The tests are dynamically generated for any class with a `bounds` attribute. This will enforce future classes to also return a `BoundsTuple`.

PR #6455, opened 2024-08-02
Fix `CubeSource` get bounds when not initialized

user27182:
### Overview

This code unexpectedly raises an attribute error. The bounds should be returned instead.
``` python
import pyvista as pv
source = pv.CubeSource()
source.bounds  # AttributeError: 'CubeSource' object has no attribute '_bounds'
```
This PR fixes this. This bug was discovered as part of #6454.

It also updates the `bounds` getter for `BoxSource` to directly use `GetBounds`.


tkoyama010 (approved):
Good catch. Thanks!

PR #6456, opened 2024-08-03
Add merge_points to PolyData class to merge points

tkoyama010:
### Overview

When merging duplicate points in a `PolyData`, it is necessary to create a `PolyData` from the points and merge it. Since this is time-consuming, we define it as one of the filters.

### Details

- None

### DIFF
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -3886,3 +3886,45 @@ class PolyDataFilters(DataSetFilters):
         alg.SetInputData(self)
         _update_alg(alg, progress_bar, "Generating Protein Ribbons")
         return _get_output(alg)
+
+    def merge_points(self, tolerance=0.0, inplace=False, progress_bar=False):
+        """Merge duplicate points in this mesh.
+
+        Parameters
+        ----------
+        tolerance : float, optional
+            Specify a tolerance to use when comparing points. Points within
+            this tolerance will be merged.
+
+        inplace : bool, default: False
+            Overwrite the original mesh with the result.
+
+        progress_bar : bool, default: False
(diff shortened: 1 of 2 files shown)


pyvista/core/filters/poly_data.py
```diff
@@ -3886,3 +3886,45 @@ def protein_ribbon(self, progress_bar=False):
...
+
+        Examples
+        --------
+        Merge duplicate points in a mesh.
+
+        >>> import pyvista as pv
+        >>> mesh = pv.Cylinder(resolution=4)
+        >>> mesh.n_points
+        16
+        >>> mesh.merge_points(inplace=True)
```
user27182, line 3918: The doctest is failing here because a mesh is returned but shouldn't be with `inplace=True`. I think maybe this is a bug with `merge_points`, and this line should be changed:
https://github.com/pyvista/pyvista/blob/6a1cb4c68ff7dd4461646d4186076f92f2c51df3/pyvista/core/filters/poly_data.py#L559-L561

``` diff
-     return self 
+     return None 
```

tkoyama010: Thanks!

tkoyama010: I have checked other methods. PyVista's `inplace` is designed to modify the object itself and return the object. I do not like this specification. Let's discuss in new issue.

tkoyama010:
```suggestion
        >>> _ = mesh.merge_points(inplace=True)
```

tkoyama010 (commented):
@pyvista-bot LGTM

tkoyama010:
@pyvista-bot LGTM

PR #6458, opened 2024-08-04
Fix import error of matplotlib in Windows Unit Testing

tkoyama010:
### Overview

Currently, the main branch Windows Unit Testing CI raise the following error.

```bash
  File "D:\a\pyvista\pyvista\pyvista\core\filters\data_set.py", line 13, in <module>
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\matplotlib\__init__.py", line [15](https://github.com/pyvista/pyvista/actions/runs/10232901666/job/28310593442#step:7:16)9, in <module>
    from . import _api, _version, cbook, _docstring, rcsetup
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\matplotlib\cbook.py", line 32, in <module>
    from matplotlib import _api, _c_internal_utils
ImportError: DLL load failed while importing _c_internal_utils: The specified module could not be found.
```
 Comparing the logs before and after the error, I found that the matplotlib installation changed from a binary to a source build. ~Although I have not been able to determine the reason for the change,~(Edit: See https://github.com/pyvista/pyvista/pull/6458#issuecomment-2267451811) I will prohibit source-built matplotlib installation in Windows CI as an emergency measure so as not to stop development.

The left is the log before the error occurred and the right is the log after the error occurred.
image

### Details

- https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-no-binary

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -293,13 +293,12 @@ jobs:
       - name: Install PyVista
         run: pip install . --no-deps
 
+      - name: Install msvc-runtime
+        run: |
+          pip install msvc-runtime
+
       - name: Install test dependencies
         run: pip install vtk -r requirements_test.txt
-
-      - name: Reinstall matplotlib
-        run: |
-          pip uninstall matplotlib
-          pip install matplotlib
 
       - name: Report
         run: python -c "import pyvista; print(pyvista.Report(gpu=False)); from pyvista import examples; print('User data path:', examples.USER_DATA_PATH)"
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -296,6 +296,9 @@ jobs:
       - name: Install test dependencies
         run: pip install vtk -r requirements_test.txt
 
+      - name: Install msvc-runtime
+        run: pip install msvc-runtime
+
       - name: Report
         run: python -c "import pyvista; print(pyvista.Report(gpu=False)); from pyvista import examples; print('User data path:', examples.USER_DATA_PATH)"
 
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -294,6 +294,7 @@ jobs:
         run: pip install . --no-deps
 
       - name: Install test dependencies
+        # https://github.com/pyvista/pyvista/pull/6458
         run: pip install vtk --only-binary "matplotlib" -r requirements_test.txt
 
       - name: Report
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -296,8 +296,10 @@ jobs:
       - name: Install test dependencies
         run: pip install vtk -r requirements_test.txt
 
-      - name: Install msvc-runtime
-        run: pip install msvc-runtime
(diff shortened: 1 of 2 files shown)


tkoyama010:
pre-commit.ci autofix

.github/workflows/testing-and-deployment.yml
```diff
@@ -294,7 +294,7 @@ jobs:
         run: pip install . --no-deps
 
       - name: Install test dependencies
-        run: pip install vtk -r requirements_test.txt
+        run: pip install vtk --only-binary "matplotlib" -r requirements_test.txt
```
tkoyama010, line 297:
```suggestion
        # https://github.com/pyvista/pyvista/pull/6458
        run: pip install vtk --only-binary "matplotlib" -r requirements_test.txt
```

adeak:
Looks like there were issues with the windows wheels so they were pulled from PyPI, see https://github.com/matplotlib/matplotlib/issues/28551#issuecomment-2267223102

From the comment thread it looks like they are aiming for a quick 3.9.2 release to fix this.

.github/workflows/testing-and-deployment.yml
```diff
@@ -294,7 +294,8 @@ jobs:
         run: pip install . --no-deps
 
       - name: Install test dependencies
-        run: pip install vtk -r requirements_test.txt
+        # https://github.com/pyvista/pyvista/pull/6458
+        run: pip install vtk --only-binary "matplotlib" -r requirements_test.txt
```
adeak, line 298: Does `--only-binary` work like this, in the middle? These switches usually apply globally to the whole install incantation.

tkoyama010:
> Looks like there were issues with the windows wheels so they were pulled from PyPI, see [matplotlib/matplotlib#28551 (comment)](https://github.com/matplotlib/matplotlib/issues/28551#issuecomment-2267223102)
Thank you. I had not noticed that.

> From the comment thread it looks like they are aiming for a quick 3.9.2 release to fix this.
Awesome, we will not need this PR once that is released.

.github/workflows/testing-and-deployment.yml
tkoyama010: Yes. We can specify the package individually.
https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-no-binary
>  or one or more package names with commas between them (no colons).

adeak (approved):
LGTM, thanks @tkoyama010.

tacaswell:
I think you can also stick the `--only-binary` in the requiremnets.txt file.

The 3.9.1 wheels on windows would, depending on import order, cause segfaults, see https://github.com/matplotlib/matplotlib/issues/28551#issuecomment-2266699619.

We deleted the bad wheels to avoid segfaults (which seemed like a good idea) but failed to anticipate that this would cause downstream projects to start trying to build mpl from source instead of pulling the 3.9.0 wheel.

Sorry for the trouble.

tkoyama010:
> I think you can also stick the --only-binary in the requiremnets.txt file.

Thanks! Now patch is much easier. Thank you for maintaining Matplotlib.

tkoyama010:
@pyvista-bot LGTM

banesullivan (approved):

adeak (approved):

PR #5872, opened 2024-04-06
Add built-in datasets to the Dataset Gallery

user27182:
Follow-up from #5822. The changes from that branch were first patched into this one before adding new commits for this PR.

This PR:

- Refactors the dataset loading mechanism for built-in datasets in `pyvista.examples.examples`. This is similar to what was done in #5787.
- Adds a new 'Built-in' carousel to the Dataset Gallery for browsing the 18 built-in datasets
- Adds a new explicit structured grid carousel (made possible since there is exactly one such dataset built-in)
- Updates the `dataset_loader` module and classes. Most notably, most classes are renamed for clarity and there is some restructuring to accommodate a new use-case for built-in datasets where a dataset is loaded without a file.
- Uses the build images directly, eliminating the need to store local copies in the repo

PR #6459, opened 2024-08-04
Fix broken alphabet buttons for the Dataset Gallery docs

user27182:
### Overview

https://github.com/pyvista/pyvista/pull/5872 added alphabet buttons to the Dataset Gallery to easily navigate to datasets by name. See the "A B C D..." buttons at the top of this screenshot:
https://github.com/pyvista/pyvista/pull/5872#pullrequestreview-1995660369

But this is now broken, it's been replaced with `<property object at (...)>`

image

I suspect this was broken by https://github.com/pyvista/pyvista/pull/5922 with this change:
``` diff
-    def doc(cls):
-        return DatasetCardFetcher.generate_alphabet_index(cls.dataset_names)
+    def doc(self):
+        return DatasetCardFetcher.generate_alphabet_index(self.dataset_names)
```

This PR attempts to fix this.

### DIFF
--- a/doc/source/make_tables.py
+++ b/doc/source/make_tables.py
@@ -1647,9 +1647,7 @@ class AllDatasetsCarousel(DatasetGalleryCarousel):
 
     name = "all_datasets_carousel"
 
-    @property
-    def doc(self):
-        return DatasetCardFetcher.generate_alphabet_index(self.dataset_names)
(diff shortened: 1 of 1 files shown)


user27182:
The docs are fixed. Only problem now is that `@classmethod` chained with `@property` is deprecated:
https://docs.python.org/3.11/library/functions.html#classmethod
image

user27182:
It's fixed now
image

tkoyama010 (approved):
LGTM

PR #6470, opened 2024-08-07
Fix `test_bounds_tuple` to enable code re-use by other tests

user27182:
### Overview

Follow-up from #6469. The tests are refactored to make it easy to add tests for other properties (like `center`).

The use of explicit named test cases is also removed since it's not necessary and leads to unnecessary failures (e.g. https://github.com/pyvista/pyvista/pull/6435#issuecomment-2270155070).


tkoyama010 (dismissed):
LGTM

tkoyama010 (approved):

PR #6473, opened 2024-08-07
Expose underlying `PlaneSource`s for `OrthogonalPlanesSource`

user27182:
### Overview

Suggested in https://github.com/pyvista/pyvista/pull/6421#discussion_r1697578036

This is done by modifying the sources when setting/getting properties instead only when `update` is called.


tkoyama010 (approved):
LGTM

PR #6474, opened 2024-08-07
Add `normal` and `flip_normal` to `PlanesSource`

user27182:
### Overview

Follow-up to #6421. The `normal` property is used to implement `PlaneSource.push()` to translate the plane along its normal.


tests/core/test_geometric_sources.py
```diff
@@ -394,10 +394,20 @@ def test_plane_source():
...
+    point_a = (0.5, -0.5, 0.0)
+    point_b = (-0.5, 0.5, 0.0)
+    normal = (0.0, 0.0, 1.0)
+    assert algo.point_a == point_a
+    assert algo.point_b == point_b
+    assert algo.normal == normal
+    algo.flip_normal()
+    assert algo.point_a == point_b
+    assert algo.point_b == point_a
+    assert algo.normal == tuple((np.array(normal) * -1).tolist())
```
tkoyama010, line 410: Is there an intention to use `array_equal` instead of `==`? `array_equal` is more appropriate for float comparisons.

tkoyama010 (changes requested):

tests/core/test_geometric_sources.py
user27182: Using `==` is better for `tuple` outputs, in my opinion because it fails for non-tuples, whereas `array_equal` does not. So, the tests are both checking value _and_ type, which removes the need to do `isinstance(x, tuple)` checks. But it also isn't obvious that the type is also being checked. `array_equal` only checks the values.

Using `==` instead of `array_equal` is what led me to propose #6469, because I tried to do `algo.center == algo.output.center` which failed because the former is a tuple and the latter is a list. This would not have been detected with `array_equal`.

But I'm not tied to this, we can change `array_equal` if that's more standard.

user27182: The tests were updated in https://github.com/pyvista/pyvista/pull/6474/commits/c491259065147c4da62edd32c8482365fe8c8383 to use `array_equal`

tkoyama010 (approved):
LGTM

PR #6477, opened 2024-08-08
Allow calling `mesh.plot()` with `background` set to an image

user27182:
### Overview

Currently only a color is allowed for `background`. This PR enables the use of file inputs as well.


MatthewFlamm (approved):
LGTM

MatthewFlamm:
Maybe a future PR, but we have the opposite problem in https://docs.pyvista.org/api/plotting/_autosummary/pyvista.plotter.add_background_image.  It accepts a path but not a color.

MatthewFlamm:
One question also maybe for another PR, should we accept any "ColorLike" as an input?  Unfortunately, this clashes with the string based Path usage.  It could be used as a nested fallback if the Path based usage fails is one thought.

user27182:
> Maybe a future PR, but we have the opposite problem in https://docs.pyvista.org/api/plotting/_autosummary/pyvista.plotter.add_background_image. It accepts a path but not a color.

I agree this should be considered for a future PR . It seems there are two methods: `add_background_image` for images and [`set_background`](https://dev.pyvista.org/api/plotting/_autosummary/pyvista.plotter.set_background#pyvista.Plotter.set_background) for colors. 

I considered maybe deprecating `add_background_image` and moving the `try..except` block from this PR into `set_background` to combine both methods into one. But both methods have a few options, and it might be confusing to have half the options work for colors but not images, and vice versa. Something to consider though.

At the very least the naming of these two methods could probably be better synchronized.

tkoyama010 (approved):

PR #6478, opened 2024-08-08
Make docstrings in `examples.planets` less verbose

user27182:
### Overview

Use feature #6477 to update the `planets` docstrings.


tkoyama010 (changes requested):
We will merge after #6477 is merged :)

tkoyama010 (approved):

PR #6482, opened 2024-08-09
Add `Transform` class for generating transformations and their inverses

user27182:
### Overview

Add new class which inherits from `vtkTransform`. This class can be used to generate more complex compound transformations. A key feature of this class is the computation of the inverse transformation, which is something I find to be relatively absent from PyVista.


pyvista/core/_validation/validate.py
```diff
@@ -521,7 +521,6 @@ def validate_transform4x4(transform, /, *, name="Transform"):
                 arr = validate_array(
                     transform,
                     must_have_shape=(4, 4),
-                    must_be_finite=True,
```
user27182, line 524: This removal is needed because this test:
https://github.com/pyvista/pyvista/blob/2930f1f7ffafb454e99b3aa51714ea9a82ecd1b9/tests/core/test_dataset.py#L281
Checks that inputs with `inf` and `nan` values do _not_ raise any errors. This is done since the vtk methods support `inf` and `nan` inputs. But is this what we want? Are there real-world use cases where `nan` and `inf` are desirable to have in a transformation matrix?
If not I think we should keep `must_be_finite=True` here and modify the test instead to _not_ allow `inf` and `nan` in a transformation.

user27182: This should be addressed as part of #6484.

pyvista/core/utilities/transform.py
```diff
@@ -0,0 +1,676 @@
...
+from pyvista.core.utilities.arrays import array_from_vtkmatrix
+from pyvista.core.utilities.arrays import vtkmatrix_from_array
+
+if TYPE_CHECKING:  # pragma: no cover
+    from pyvista.core._typing_core import NumpyArray
+    from pyvista.core._typing_core import RotationLike
+    from pyvista.core._typing_core import TransformLike
+
+
+class Transform(_vtk.vtkTransform):
```
tkoyama010, line 19: Thanks for adding great class! It is possible to make the operation of the Transform class even more intuitive to use with ["operator — Standard operators as functions"](https://docs.python.org/3/library/operator.html#operator) in my mind :)

tkoyama010 (changes requested):

pyvista/core/utilities/transform.py
user27182: Great idea! Hadn't thought of that. I was thinking this PR is already a bit big with approx 800-900 LOC, maybe we can add that as a separate PR?

This class can also be extended to incorporate `reflect`, `rotate_x`, etc. transformations. And I was thinking of adding a `point` kwarg too for rotation about a point, reflection about a point, etc. 

Lot's of possible features to add.

tkoyama010 (approved):
LGTM

pyvista/core/utilities/transform.py
user27182: > Thanks for adding great class! It is possible to make the operation of the Transform class even more intuitive to use with ["operator — Standard operators as functions"](https://docs.python.org/3/library/operator.html#operator) in my mind :)

To add to this, we could even make this class comply with numpy's array protocols, such that you can do `np.array(transform)` or `np.eye(4)` @ `transform` which should return a numpy array. Or even implement `__getitem__` to do `element = transform[0,1]`.

As an implementation note (for future PRs), this means that the `matrix_list` should return a list of `Transform` instances instead of currently doing `array_from_vtkmatrix(transform.GetMatrix())`

PR #6480, opened 2024-08-08
Improve test coverage of `set_focus` and `set_viewup`

tkoyama010:
### Overview

Follow up of #6465

### Details

- None

PR #6479, opened 2024-08-08
Increase test coverage of `plot_datasets`

tkoyama010:
### Overview

Follow up of #6465

### Details

- None

PR #6465, opened 2024-08-06
Remove unnecessary indent by integrate if block

tkoyama010:
### Overview

Remove unnecessary indent by integrate if block.

### Details

- None

PR #6483, opened 2024-08-09
`codecov.yml` coverage goal change requires admin approval

tkoyama010:
### Overview

#6479 , #6480 and #6465 made me decide that we should be cautious about changing the coverage target.

### Details

- None


tkoyama010:
@pyvista-bot LGTM

PR #6484, opened 2024-08-10
Use `Transform` to generate dataset transformations

user27182:
### Overview

Integrate the `Transform` class from #6482 into the code base.


tests/core/test_dataset.py
```diff
@@ -53,7 +54,9 @@ def test_invalid_copy_from(grid):
 def n_numbers(draw, n):
     numbers = []
     for _ in range(n):
-        number = draw(one_of(floats(), integers()))
+        number = draw(
+            one_of(floats(), integers(max_value=np.iinfo(int).max, min_value=np.iinfo(int).min))
+        )
```
user27182, line 59: Using `integers()` will test integer over/underflow, e.g. values such as `np.iinfo(int).min-1`. This is great for test coverage, but it also revealed that `numpy` will actually create an `object` array instead of an `int` array in some cases. E.g.:
``` python
import numpy as np
np.array((np.iinfo(int).min-1))
# array(-9223372036854775809, dtype=object)
```

And so because it's an object array, the `_cast_to_numpy` function for the array validation raises a TypeError. This causes the tests to fail. And I don't know how to add an exception to this, e.g. how to add something like `pytest.xfail` using `hypothesis`. Instead, max and min and for integers is restricted so there is no over/underflow.

user27182:
Draft since this should probably be merged after #6501 so that all changes to all transformations can be done in this PR

tkoyama010 (approved):
LGTM

tests/core/test_dataset.py
```diff
@@ -1258,31 +1274,34 @@ def test_rotate_y():
     mesh = examples.load_uniform()
     out = mesh.rotate_y(30)
     assert isinstance(out, pv.StructuredGrid)
-    with pytest.raises(TypeError):
-        out = mesh.rotate_y(30, point=5)
-    with pytest.raises(ValueError):  # noqa: PT011
-        out = mesh.rotate_y(30, point=[1, 3])
+    match = 'Shape must be one of [(3,), (1, 3), (3, 1)]'
+    with pytest.raises(ValueError, match=re.escape(match)):
+        out = mesh.rotate_x(30, point=5)
```
user27182, line 1279:
```suggestion
        out = mesh.rotate_y(30, point=5)
```

tests/core/test_dataset.py
```diff
@@ -1258,31 +1274,34 @@ def test_rotate_y():
...
+    with pytest.raises(ValueError, match=re.escape(match)):
+        out = mesh.rotate_x(30, point=[1, 3])
```
user27182, line 1281:
```suggestion
        out = mesh.rotate_y(30, point=[1, 3])
```

tests/core/test_dataset.py
```diff
@@ -1258,31 +1274,34 @@ def test_rotate_y():
...
     mesh = examples.load_uniform()
     out = mesh.rotate_z(30)
     assert isinstance(out, pv.StructuredGrid)
-    with pytest.raises(TypeError):
-        out = mesh.rotate_z(30, point=5)
-    with pytest.raises(ValueError):  # noqa: PT011
-        out = mesh.rotate_z(30, point=[1, 3])
+    match = 'Shape must be one of [(3,), (1, 3), (3, 1)]'
+    with pytest.raises(ValueError, match=re.escape(match)):
+        out = mesh.rotate_x(30, point=5)
```
user27182, line 1291:
```suggestion
        out = mesh.rotate_z(30, point=5)
```

tests/core/test_dataset.py
```diff
@@ -1258,31 +1274,34 @@ def test_rotate_y():
...
+    with pytest.raises(ValueError, match=re.escape(match)):
+        out = mesh.rotate_x(30, point=[1, 3])
```
user27182, line 1293:
```suggestion
        out = mesh.rotate_z(30, point=[1, 3])
```

tkoyama010 (approved):

PR #6485, opened 2024-08-10
Add `align_xyz` filter to align a mesh to the x-y-z axes

user27182:
### Overview

Align a mesh to the x-y-z axes. This uses `principal_axes` from #6346.


pyvista/core/filters/data_set.py
```diff
@@ -193,6 +200,248 @@ def align(
...
+            axes[2] *= -1
+
+        # Maybe flip direction of third axis
+        if axis_2_direction is not None:
+            has_correct_sign = np.dot(axes[2], axis_2_direction) >= 0
+            other_directions_are_specified = (
+                axis_0_direction is not None and axis_1_direction is not None
+            )
+            if has_correct_sign or other_directions_are_specified:
+                pass
```
tkoyama010, line 423: Is this the change you intend?
```suggestion
            if not (has_correct_sign or other_directions_are_specified):
                raise ValueError("message of error.")
```

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
user27182: Should be fixed with https://github.com/pyvista/pyvista/pull/6485/commits/9cc91aadfed05b01a44a3fde069a62168a2926ba

PR #6486, opened 2024-08-10
Add helpful error message when reading VRML file

user27182:
### Overview
I tried to read a file with:
``` python
import pyvista as pv
pv.read('file.wrl')
```
and got this error:
`OSError: This file was not able to be automatically read by pyvista.`

which led me to think this file format wasn't supported. It is, but requires a special import. This PR adds a helpful error message to communicate this.


tkoyama010 (approved):
LGTM

PR #6487, opened 2024-08-10
Fix Plotter bounds when importing VRML file

user27182:
### Overview

The Plotter bounds in PyVista are manually computed from the list of actors that have explicitly been added. This means that any actors added to a renderer by any other means are not "seen" or "known" by the Plotter, and thus the bounds computation will be incorrect.

As a result, using the `import_vrml` method to add actors to a scene does not report the correct bounds. E.g.
``` python
import pyvista as pv
from pyvista import examples
pl = pv.Plotter()
pl.import_vrml(examples.vrml.download_sextant())
pl.bounds  # BoundsTuple(x_min=-1.0, x_max=1.0, y_min=-1.0, y_max=1.0, z_min=-1.0, z_max=1.0)
```

This PR fixes this, e.g. the correct bounds are:
``` python
pl.bounds  # BoundsTuple(x_min=0.48927752354865817, x_max=3.3604576885597908, y_min=-0.8170784576890213, y_max=1.2712147237834017, z_min=-2.109999952837825, z_max=-1.0)
```

This is more of a workaround than an actual fix. In my opinion I think the Plotter's `bounds` computation is problematic and should be updated. See https://github.com/pyvista/pyvista/pull/5182 for a separate issue with the bounds computation.

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -513,8 +513,18 @@ class BasePlotter(PickingHelper, WidgetHelper):
             importer.SetFileName(str(filename))
         else:
             importer.SetFileName(filename)
-        importer.SetRenderWindow(self.render_window)
+
+        # Import to temporary plotter first
+        temp_plotter = pyvista.Plotter()
+        importer.SetRenderWindow(temp_plotter.render_window)
         importer.Update()
+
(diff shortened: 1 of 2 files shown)


user27182:
I do not work with this file format but if anyone wants to pick this up and fix this feel free to use this PR as a starting point.

PR #6491, opened 2024-08-11
Add `check_finite` property to `Transform` class

user27182:
### Overview

Add `check_finite` property to `Transform` class. To implement this feature,  the `must_be_finite` option for `validate_transform3x3` and `validate_transform4x4` is also exposed.

In #6484, the `Transform` class was found to generate matrices with `nan` values for some inputs (with the `RotateWXYZ` method in particular). This PR enables checks to make it known to the user if the computed transformations are invalid.


tkoyama010 (approved):
LGTM

pyvista/core/_validation/validate.py
```diff
@@ -588,7 +600,9 @@ def validate_transform3x3(transform, /, *, name="Transform"):
             else:
                 if isinstance(transform, Rotation):
                     # Get matrix output and try validating again
-                    return validate_transform3x3(transform.as_matrix())
+                    return validate_transform3x3(
+                        transform.as_matrix(), must_be_finite=must_be_finite
```
user27182, line 604:
```suggestion
                        transform.as_matrix(), must_be_finite=must_be_finite, name=name
```

tkoyama010 (approved):

PR #6501, opened 2024-08-12
Add `point` parameter to `Transform`

user27182:
### Overview

Allow `scale`, `reflect`, `rotate`, `rotate_x`, `rotate_y`, `rotate_z`, `rotate_vector` to transform about a `point`.

~The `DataSet.scale()`  method is also updated to allow scaling from a point.~ Changes to dataset methods are made in #6484 instead.

### DIFF
--- a/tests/core/test_utilities.py
+++ b/tests/core/test_utilities.py
@@ -1261,12 +1261,9 @@ def transformed_actor():
 
 @pytest.mark.parametrize('override_mode', ['pre', 'post'])
 @pytest.mark.parametrize('object_mode', ['pre', 'post'])
-def test_transform_override_multiply_mode_and_point(
-    transform, transformed_actor, object_mode, override_mode, object_point, override_point
-):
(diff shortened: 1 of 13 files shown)


tests/core/test_utilities.py
```diff
@@ -1173,18 +1172,55 @@ def test_transform_reflect(transform, reflect_args):
...
+        ('rotate', (ROTATION,)),
+        ('rotate_x', (ANGLE,)),
+        ('rotate_y', (ANGLE,)),
+        ('rotate_z', (ANGLE,)),
+        ('rotate_vector', (VECTOR, ANGLE)),
+    ],
+)
+def test_transform_with_point(transform, multiply_mode, method, args):
+    func = getattr(Transform, method)
+    vector = np.array((1, 2, 3))
```
tkoyama010, line 1200:
```suggestion
    vector = np.array(NORMAL)
```

tkoyama010 (changes requested):

PR #6502, opened 2024-08-12
Cast `center` to list in `project-plane.py`

user27182:
### Overview

Follow-up to #6469. This example got missed, presumably due to caching from the docs. It is now causing failures in other PRs.


tkoyama010 (approved):
Thanks!

PR #6503, opened 2024-08-12
Add `reflect` method to `Transform` class

user27182:
### Overview

Add `reflect` to `Transform`.

Note: this transformation can make use of `point` from #6501, so either this PR or that one should be modified (which ever is merged second)


tkoyama010 (approved):
LGTM

PR #6504, opened 2024-08-13
Add operator overloads to `Transform`

user27182:
### Overview

Originally suggested here: https://github.com/pyvista/pyvista/pull/6482#discussion_r1712838332

`__add__` implements post-multiply, and `__matmul__` implements pre-multiply.

`copy` method is also added (needed to avoid self-mutation).

The main class examples are also reworked to show the use of these operators.


pyvista/core/utilities/transform.py
```diff
@@ -51,93 +51,123 @@ class Transform(_vtk.vtkTransform):
...
+           [ 0. ,  0. ,  1. ,  2.1],
+           [ 0. ,  0. ,  0. ,  1. ]])
+
+    >>> scale_factor = 2.0
+    >>> scaling = Transform().scale(scale_factor)
+    >>> scaling.matrix
+    array([[2., 0., 0., 0.],
+           [0., 2., 0., 0.],
+           [0., 0., 2., 0.],
+           [0., 0., 0., 1.]])
```
tkoyama010, line 72: Awesome job! Is it possible to assign `scale` to [`operator.__mul__`](https://docs.python.org/ja/3.10/library/operator.html#operator.__mul__)?

tkoyama010: ref [Python : Operator Overloading a specific type](https://stackoverflow.com/questions/3188666/python-operator-overloading-a-specific-type)

user27182: ok 👍

We could probably overload `__add__` too to allow for translations if doing addition where `other` is array-like

user27182: For things like `__mul__` and `__add__`, should we use those for post-multiply? e.g. `Transform() * 5`
But then use pre-multiply with `__rmul__`, `__radd__` ?  e.g. `5 * Transform()`?

user27182:
Draft - will implement more operators

pyvista/core/utilities/transform.py
tkoyama010: How about making it behave like NumPy's Array? I believe it could also use right multiplication as a scale. I do not fully understand the disadvantages of this, so if you have any additional information, please share :)

user27182: The main thing to consider is that the transformations are not commutative, which is why there is pre- and post- multiply. The transformations use matrix operations by default, whereas numpy uses element-wise operations and broadcasting by default.

For matmul with `@`, the transform behaves exactly like a numpy array using pre-multiply. But there is no equivalent operator for post-multiply. So, with this PR as it is, `__add__` is used, but it is actually doing matrix multiplication for concatenating the transformations, not actual addition. This is very different from how a regular numpy array behaves. And overloading this to allow `__add__` where one array is 4x4 (the transform) and another array is 3x1 (position vector) also is very different from how a numpy array behaves, which would raise an error due to shape mis-match.

So I guess we should be cautious about making this class look too much like a numpy array to avoid unintended operations. For example, if we add a 4x4 numpy array to a Transform, should we expect regular numpy addition? Or should the 4x4 array be concatenated with the transform?

user27182:
Ready for review, most of the sensible operators have been implemented (`__mul__`, `__add__`, `__matmul__`). Notably, no in-place operators are included, e.g. `__iadd__`. This is kind of extra, and can be part of a separate PR.

pyvista/core/utilities/transform.py
tkoyama010: > So I guess we should be cautious about making this class look too much like a numpy array to avoid unintended operations. For example, if we add a 4x4 numpy array to a Transform, should we expect regular numpy addition? Or should the 4x4 array be concatenated with the transform?

I can't provide a clear answer to this. I think it would be best to prohibit operations with objects other than Transform objects at the moment. If there is a request for operations with objects of other types at a later date, let's respond to it.

tkoyama010 (approved):
LGTM

pyvista/core/utilities/transform.py
user27182: > I think it would be best to prohibit operations with objects other than Transform objects at the moment.

To be clear, this PR _does_ allow operations with non-Transform objects, e.g. translation with a length-3 vector, scaling with a float or length-3 vector, or concatenation with a 3x3 or 4x4 NumPy array or vtkMatrix (i.e. any TransformLike object). E.g. see the parameters for this test:
https://github.com/pyvista/pyvista/blob/c9429d272a894c899efc24f1e0d697c29518e0df/tests/core/test_utilities.py#L1419-L1422

But, unlike with #6508, _only_ the `Transform` overloads are used for the operation, not `np.array` overloads. So, if a `Transform` object is either on the right or left of the operator, then a `Transform` operation is always performed (not a numpy array operation)

tkoyama010 (approved):

PR #6507, opened 2024-08-13
Add custom `repr` to `Transform`

user27182:
### Overview

The repr mainly formats the `matrix` array to have a standard shape and precision to make it easier to see and compare its elements.


pyvista/core/utilities/transform.py
```diff
@@ -185,6 +187,24 @@ def __init__(self, trans: TransformLike | None = None):
         if trans is not None:
             self.matrix = trans  # type: ignore[assignment]
 
+    def __repr__(self):
+        """Representation of the transform."""
+
+        def _matrix_repr():
+            repr_ = np.array_repr(self.matrix, precision=4, suppress_small=True)
```
user27182, line 194: Maybe it's best not to truncate the precision and/or suppress small values?
```suggestion
            repr_ = np.array_repr(self.matrix)
```

tkoyama010: +1 not to

tkoyama010 (approved):
LGTM

PR #6508, opened 2024-08-13
Add `__array__` method to `Transform` for casting to a NumPy array

user27182:
### Overview

Make this code valid and return a `Transform`'s `matrix`: 
``` python
np.array(Transform())
```


tkoyama010 (approved):
LGTM

tkoyama010:
OK. Let me think about it for a bit.

user27182:
Closing as I'm not sure this is a good idea, it would create ambiguity between operations, e.g. is `+` concatenation or addition? is `*` concatenation with a scaling matrix or is it multiplication?

PR #6509, opened 2024-08-14
Add annotations to  `check_length` and update variable names

user27182:
### Overview

Follow-up to #5571.


pyvista/core/_validation/check.py
```diff
@@ -1071,48 +1073,45 @@ def check_length(
...
+            raise ValueError(
+                f"{name} must have a minimum length of {min_length}. "
+                f"Got length {array_len} instead.",
+            )
+    if max_length is not None:
+        if array_len > max_length:
+            raise ValueError(
+                f"{name} must have a maximum length of {max_length}. "
+                f"Got length {array_len} instead.",
+            )
```
tkoyama010, line 1114:
```suggestion
    if min_length is not None and array_len < min_length:
        raise ValueError(
            f"{name} must have a minimum length of {min_length}. "
            f"Got length {array_len} instead.",
        )
    if max_length is not None and array_len > max_length:
        raise ValueError(
            f"{name} must have a maximum length of {max_length}. "
            f"Got length {array_len} instead.",
        )
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

PR #6514, opened 2024-08-15
Pad `voxelize` and `voxelize_volume` to contain input mesh

aeverallpx:
### Overview

Padding added to pyvista voxelisation. 

### Details

Currently pyvista voxelize and voxelize volume do not cover the entire mesh because the np.arange method finds the nearest bin within the range. 

This commit ensures that the bounds of the voxelization completely cover the input mesh.

Example from the docs - https://docs.pyvista.org/api/utilities/_autosummary/pyvista.voxelize_volume.html

The following code generates a voxelization with surface vertices which lie within the input mesh - therefore the input mesh isn't fully covered.
```python
import numpy as np
from pyvista import examples

import pyvista as pv
pv.start_xvfb(wait=0.0)

vox = pv.voxelize_volume(mesh, density=0.15)
vox = vox.select_enclosed_points(mesh, tolerance=0.0)
vox.plot(scalars='SelectedPoints', show_edges=True, jupyter_backend='static')
```
image

The horns, nose and side of the cow extend outside the voxelization. Using the `enclosed` option, the mesh is fully contained and the boundaries of the mesh lie within the last voxel at the edges.

```
vox = pv.voxelize_volume(mesh, density=0.15, enclosed=True)
vox = vox.select_enclosed_points(mesh, tolerance=0.0)
vox.plot(scalars='SelectedPoints', show_edges=True, jupyter_backend='static')
```

image


tkoyama010 (changes requested):
@aeverallpx Thank you for your contribution. This patch is awesome. In order to ensure backward compatibility, I sugget that this be made an option. If you need any help, please let us know :)

aeverallpx:
Hi @tkoyama010, I've made the suggested change so there is now an `enclosed` kwarg to enable this which is `False` by default. I've also added a quick couple of tests. I haven't managed to get purest to run on this repo locally as I get a core dump but let me know if this is necessary or if you need any other changes.

pyvista/core/utilities/features.py
```diff
@@ -32,6 +62,10 @@ def voxelize(mesh, density=None, check_surface=True):
         algorithm first checks to see if the surface is closed and
         manifold. If the surface is not closed and manifold, a runtime
         error is raised.
+        
+    enclosed : bool, default: False
```
user27182, line 66: Can we add an example to the docstring with `enclosed=True`? It would be nice to have two figures one after the other to easily see / compare the difference. Something similar to what's shown in the this PR.

pyvista/core/utilities/features.py
```diff
@@ -121,6 +160,10 @@ def voxelize_volume(mesh, density=None, check_surface=True):
...
```
user27182, line 164: Same as above - let's add an example.

pyvista/core/utilities/features.py
```diff
@@ -14,7 +14,37 @@
 from .helpers import wrap
 
 
-def voxelize(mesh, density=None, check_surface=True):
+def padded_bins(mesh, density):
```
user27182, line 17: Is this intended for general use, or should we mark it as private?
```suggestion
def _padded_bins(mesh, density):
```

user27182 (commented):
Thanks for your contribution! Here's a few thoughts and suggestions.

tkoyama010:
pre-commit.ci autofix

tests/core/test_utilities.py
```diff
@@ -320,6 +320,23 @@ def test_voxelize_volume_no_face_mesh(rectilinear):
...
+    for i in range(6):
+        assert vox_bounds[i] * (-1)**i <= mesh_bounds[i] * (-1)**i
+        
+def test_voxelize_enclosed_bounds():
+    mesh = pv.Sphere()
+    vox = pv.voxelize(mesh, density=0.15, enclosed=True)
+    mesh_bounds = mesh.bounds
+    vox_bounds = vox.bounds
+    for i in range(6):
+        assert vox_bounds[i] * (-1)**i <= mesh_bounds[i] * (-1)**i
```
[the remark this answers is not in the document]
aeverallpx: Thanks for the suggestion. I've largely implemented what you suggested. A couple of things I couldn't implement:
2. Not being enclosed does not mean the voxelization exceeds all bounds on all axes, it just means they're not guaranteed to fall inside. In practice you get the lower bound matching the input mesh lower bound and the upper bound falling inside.
4. Airplane is unfortunately not manifold so pyvista voxelization doesn't work as it looks for points contained within the mesh.
Also with this test - I realised it isn't a good test for voxelize as depending on the shape of the mesh, the outer cells can be discarded. This is true for the ant example. The test is good for voxelize_volume though and since they both use the same underlying binning function the base grid generated is the same.

pyvista/core/utilities/features.py
aeverallpx: Also added here

pyvista/core/utilities/features.py
aeverallpx: Thank you, good change.

tests/core/test_utilities.py
```diff
@@ -320,6 +320,20 @@ def test_voxelize_volume_no_face_mesh(rectilinear):
...
+    mesh = ex.load_ant()
+    vox = function(mesh, density=1, enclosed=True)
+
+    assert (vox.bounds.x_min <= mesh.bounds.x_min)
+    assert (vox.bounds.y_min <= mesh.bounds.y_min)
+    assert (vox.bounds.z_min <= mesh.bounds.z_min)
+
+    assert (vox.bounds.x_max >= mesh.bounds.x_max)
+    assert (vox.bounds.y_max >= mesh.bounds.y_max)
+    assert (vox.bounds.z_max >= mesh.bounds.z_max)
```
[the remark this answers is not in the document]
user27182: This change was made because there was missing test coverage for `pv.voxelize`. But now the tests are failing:
https://github.com/pyvista/pyvista/actions/runs/10548641482/job/29222787144#step:12:6076

This must be related to this https://github.com/pyvista/pyvista/pull/6514#discussion_r1726604581:

> Also with this test - I realised it isn't a good test for voxelize as depending on the shape of the mesh, the outer cells can be discarded. This is true for the ant example. The test is good for voxelize_volume though and since they both use the same underlying binning function the base grid generated is the same.

@aeverallpx can you elaborate on this? What do you mean by outer cells can be discarded,  and why is this OK/expected?

We can use a different fixture just to make the tests pass but I think the term `enclosed` should imply that this test should always pass. Is there something I'm missing about how this option works?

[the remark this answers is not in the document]
aeverallpx: Sorry it has taken a while to get back to this. 

The problem was that voxelize choses vertices to include if they are contained in the input mesh surface. 
```selection = ugrid.select_enclosed_points(surface, tolerance=0.0, check_surface=check_surface)```

I've  changed this to include cells which are within one cell of the surface which means that the input mesh is enclosed by the produced voxelization. Below are screenshots for voxelization of the ant with `enclosed=False` and `enclosed=True`.
image

Tests now all pass. I hope this resolves the issue.

aeverallpx:
pre-commit.ci autofix

tkoyama010:
For information about document build errors, [Documentation Image Regression Testing](https://dev.pyvista.org/extras/developer_notes#documentation-image-regression-testing) is a good reference :)

user27182:
The documentation tests are failing because the examples that were added generate new images which were not previously part of the documentation. To resolve this you should add the new images to `tests/doc/doc_image_cache`.

The new images can be downloaded from this artifact:
https://github.com/pyvista/pyvista/actions/runs/10634006676/job/29480547685?pr=6514#step:15:33

tkoyama010:
pre-commit.ci autofix

aeverallpx:
Thanks for the pointers. I've been trying to run the build docs locally but couldn't get it to work due to rendering issues. I've also realised that I'd set the density wrong in the voxelize docstring so I've changed that to be correct for the bunny. Once the docs have built here I'll add the new figures to the test directory then everything should be complete.

aeverallpx:
Should all be fixed now. Waiting on docs to finish building but they passed before I had to merge in main so should be fine. @tkoyama010 happy with changes?

tkoyama010 (approved):
LGTM. Thanks for your great contribution!

PR #6517, opened 2024-08-16
Simplify code for `fit_plane_to_points`, fix crash with many points, fix returned center

### LINKED ISSUE #6460: `fit_plane_to_points` crashes python with many points

This code crashes python:
```python
import pyvista as pv
points = pv.examples.download_armadillo().points
len(points)  # 172974
pv.fit_plane_to_points(points)  # crashes
# Process finished with exit code 137 (interrupted by signal 9: SIGKILL)
```
I believe the issue stems from using `np.linalg.svd` to fit the points, which has a high memory footprint. Instead, PCA may be used for the computation. PCA is used in the `principal_axes` method from #6346, so that method can be used internally to replace the call to `np.linalg.svd`

See above

### System Information

```shell
OS : Darwin (macOS 14.4.1)
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : Apple
      GPU Renderer : Apple M2
       GPU Version : 4.1 Metal - 88
  MathText Support : True

  Python 3.9.13 (v3.9.13:6de2ca5339, May 17 2022, 11:37:23)  [Clang 13.0.0
  (clang-1300.0.29.30)]

           pyvista : 0.45.dev0
               vtk : 9.3.1
             numpy : 2.0.0
        matplotlib : 3.9.0
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 10.4.0
           IPython : 8.18.1
          colorcet : 3.1.0
             scipy : 1.13.1
              tqdm : 4.66.4
    pytest_pyvista : 0.1.9
```

### Screenshots

_No response_

user27182:
### Overview

Maintenance:
- Simplify code by using `align_xyz` to fit the plane instead of using `np.linalg.svd`
- Update tests to remove the use of magic numbers

Bug fix:
- Fix #6460 (`align_xyz` uses `principal_axes` instead of svd)
- Fix bug with incorrect center returned


tests/core/test_utilities.py
```diff
@@ -901,23 +901,49 @@ def test_has_module():
     assert not has_module('not_a_module')
 
 
-def test_fit_plane_to_points():
-    points = ex.load_airplane().points
-    plane, center, normal = fit_plane_to_points(points, return_meta=True)
-
-    assert np.allclose(normal, [-2.5999512e-08, 0.121780515, -0.99255705])
-    assert np.allclose(center, [896.9954860028446, 686.6470205328502, 78.13187948615939])
```
user27182, line 909: Note that the in the new tests, the returned center does NOT have this value. This means that there was previously a bug with how center was calculated. This is now fixed by this PR (and verified  as correct by the tests)

tkoyama010 (changes requested):
Need to wait #6521

user27182:
I think I'll split this up to keep it maintenance only,  and add features in a separate PR

user27182:
I think this may be a breaking change in the sense that the direction of the normal may be flipped between the old implementation and the new one. Any code fitting a plane, then relying on this plane for signed distance calcs or for clipping a dataset would now silently return a different result. 

Do we need a warning in the docs, or just a versionchanged directive?

tkoyama010:
Please add a warning in the docs.

user27182:
Keeping this as draft based on this comment: https://github.com/pyvista/pyvista/issues/6460#issuecomment-2299735699
I want to explore keeping SVD as an option for computing the principal axes

user27182:
> I want to explore keeping SVD as an option for computing the principal axes

This is explored here: https://github.com/pyvista/pyvista/issues/6460#issuecomment-2302993313. We should not use SVD. This is ready for review as-is.

tkoyama010 (approved):
LGTM

PR #6520, opened 2024-08-16
Speed up and simplify `principal_axes` test with many points

### LINKED ISSUE #6460: `fit_plane_to_points` crashes python with many points

user27182:
### Overview

The current test to check that principal axes succeeds with many points is slow and hangs up the testing (have to wait 5~10 seconds for this one test to execute). This is because it generates an ellipsoid for the test. The ellipsoid was initially used to check the correct std output. But, this requires magic numbers for the test, and is redundant since the correct std output is tested elsewhere (and more correctly without magic numbers).

This PR uses a numpy array instead the ellipsoid for the tests. This speeds up the test considerably. In addition, I'd like to add a similar to test for `fit_plane_to_points` as part of #6517 to test that it resolves #6460. The fixture from this PR can be used for that test as well.

### DIFF
--- a/tests/core/test_utilities.py
+++ b/tests/core/test_utilities.py
@@ -1004,19 +1004,15 @@ def test_principal_axes_single_point():
     assert np.allclose(axes, DEFAULT_PRINCIPAL_AXES)
 
 
-def test_principal_axes_vectors_success_with_many_points():
+@pytest.fixture()
+def one_million_points():
+    return np.random.default_rng().random((1_000_000, 3))
+
+
+def test_principal_axes_success_with_many_points(one_million_points):
     # Use large mesh to verify no memory errors are raised
-    res = 1000
-    ellipsoid = pv.ParametricEllipsoid(
-        xradius=3, yradius=2, zradius=1, u_res=res, v_res=res, w_res=res
-    )
-    assert ellipsoid.n_points == 997_004
-
-    axes, std = pv.principal_axes(ellipsoid.points, return_std=True)
-
-    # Check std to verify the computed output is valid
-    # Need large atol due to loss of numerical precision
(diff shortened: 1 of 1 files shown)


tkoyama010 (approved):
Good catch. Thanks!

user27182:
No code cov yet, will close and re-open

user27182:
Still no codecov 🤷‍♂️

PR #6521, opened 2024-08-16
Do not initialize `axis_2_direction` for `align_xyz`

user27182:
### Overview

Only two axis directions can be set for `align_xyz`. Setting the third by default has no effect and can result in an unexpected error. This PR fixes this.

Discovered as part of #6517.


tkoyama010 (approved):
LGTM

PR #6523, opened 2024-08-16
Fix default value in docs for `partition` filter `as_composite` option

user27182:
### Overview

Function signature says True, doc says False. Change this to True.


tkoyama010 (approved):
LGTM

PR #6525, opened 2024-08-17
Allow empty `ImageData` init

### LINKED ISSUE #4711: `ImageData().points` raises unexpected ValueError

user27182:
### Overview

Fix #4711.


tests/core/test_grid.py
```diff
@@ -864,6 +864,18 @@ def test_create_image_data_from_specs():
...
+@pytest.mark.parametrize('dims', [None, (0, 0, 0), (1, 0, 0), (0, 1, 0), (0, 0, 1)])
+def test_image_data_empty_init(dims):
+    image = pv.ImageData(dimensions=dims)
+    assert image.n_points == 0
+    assert image.n_cells == 0
+    assert image.area == 0
+    assert image.volume == 0
+
+    points = image.points
+    assert np.array_equal(points, np.empty((0, 3)))
```
MatthewFlamm, line 876: The fact that `np.empty` gives zeros here is not guaranteed.  See notes section in [`numpy.empty` docs](https://numpy.org/doc/stable/reference/generated/numpy.empty.html#numpy-empty)

```suggestion
    assert np.array_equal(points, np.zeros((0, 3)))
```

MatthewFlamm (commented):
Otherwise, LGTM

pyvista/core/grid.py
```diff
@@ -652,6 +652,10 @@ def points(self) -> NumpyArray[float]:
                [1., 1., 1.]])
 
         """
+        # Handle empty case
+        if not all(self.dimensions):
+            return np.empty((0, 3))
```
tkoyama010, line 657: Do we also need to initialize this to zero?
```suggestion
            return np.zeros((0, 3))
```

tkoyama010 (changes requested):

pyvista/core/grid.py
user27182: Seems like they produce the same result? Both are arrays are empty. Since the dimensions are `(0,3)` there are no actual values in the output
``` python
import numpy as np; np.zeros((0,3))
# Out[3]: array([], shape=(0, 3), dtype=float64)
import numpy as np; np.empty((0,3))
# Out[4]: array([], shape=(0, 3), dtype=float64)
```

But maybe it's better to change it to `np.zeros` to match the test

tests/core/test_grid.py
user27182: See this [comment](https://github.com/pyvista/pyvista/pull/6525#discussion_r1720987894). Seems like it doesn't matter which method is used to create the empty array

pyvista/core/grid.py
tkoyama010: Ah! If the size is zero, it's the same. I'll change it to make it same with test code. Thanks for the comment.

tkoyama010 (approved):
LGTM

pyvista/core/grid.py
MatthewFlamm: I missed that it was used twice.  The fact that `np.empty` returns zeros is not guaranteed.  It _could_ return nonzeros on some systems.  Thanks for fixing.

PR #6526, opened 2024-08-17
Add flag to force all static plots for documentation builds

### LINKED ISSUE #6515: Reduce size of documentation build or modify actions to use less disk space

Documentation builds are failing due to a lack of disk space. This is happening during post-build caching:
no-space

This is preventing new PRs from merging, e.g. see the checks for #6504, the build has failed multiple times for this reason.

Apparently the linux runners only have 7GB of space: https://dev.to/n3wt0n/github-larger-runners-explained-no-more-limits-419p. And the builds are now 4GB + in size so depending on what's being cached and how temp files are managed the runner storage requirements could easily require 8GB.

Maybe it would be good to download a build locally and sort the files by size, and remove the "interactive" plotting option for the largest files, as I suspect it's the interactive plots that are inflating the build size.

Alternatively (or in addition), maybe it's possible to modify how caching is done, e.g. first upload the build files (this is already done for the build preview action), and perform caching as part of a separate action.

user27182:
### Overview

Attempt to resolve #6515.

The idea behind this PR is to add a global option to force all plots to be static for the documentation build. This should hopefully speed up build times considerably and make it so that the html is much smaller, enabling for faster turnaround between commits / CI checks.

This option could maybe be toggled using labels (still need to figure out how to do this).


user27182:
Build size is now 1.3GB instead of 4.2GB.
Build time is still about the same though 1h40m

tkoyama010:
Thanks for considering this!!! +100 for outputting static images when adding labels.

user27182:
> Thanks for considering this!!! +100 for outputting static images when adding labels.

Great! The PR currently works for disabling the interactive plots,  but I'm unfamiliar with how to toggle this with labels. I think some environment variables may need to be set? Might need some help getting this part of the feature to work.

user27182:
pre-commit.ci autofix

user27182:
This flag does seem to be working to reduce the build size. That is good, but what I was really hoping for is a reduction in build _time_, which this doesn't seem to do.

Feel free to reopen if anyone else can help with adding labels to toggle this open or if anyone knows about other hidden options for `plot_directive` which, if enabled, can help to reduce the build time.

user27182:
> Thanks for considering this!!! +100 for outputting static images when adding labels.

Just realized that you opened a previous PR for this option in #6101

> This flag does seem to be working to reduce the build size. That is good, but what I was really hoping for is a reduction in build _time_, which this doesn't seem to do.

Will try to measure processing times with `'sphinx.ext.duration'` and compare to #6993. Though I'm still doubtful that build times will be reduced. Perhaps it's not generating the interactive plot that takes time, but that simply executing the docstring examples is what is taking so long.

PR #6528, opened 2024-08-17
Add `resolution` parameter to `fit_plane_to_points` and update examples

### LINKED ISSUE #4705: Add `covariance` method to `fit_plane_to_points` and expose `Plane` arguments

user27182:
### Overview

Split from #6517.

Enhancement:
- Add new `resolution` parameter (fix #4705)

Documentation:
- Update docstring to include an example using the returned meta data


tkoyama010 (approved):
LGTM

PR #6530, opened 2024-08-18
Add `apply` method to `Transform` to apply transformations to points or datasets

user27182:
### Overview

New method to make it easy to apply transformations.

EDIT: ~This method could generalize to include MultiBlock too if #6543 is merged first~ Transforming MultiBlock is now supported


pyvista/core/utilities/transform.py
```diff
@@ -1243,6 +1247,105 @@ def n_transformations(self) -> int:  # numpydoc ignore: RT01
...
+
+            Points with integer values are cast to a float type before the
+            transformation is applied.
+
+        Parameters
+        ----------
+        obj : VectorLike[float] | MatrixLike[float] | pyvista.DataSet
+            Object to apply the transformation to.
+
+        inplace : bool, default: False
```
user27182, line 1285: I'll rename `inplace` to `copy`

tkoyama010 (approved):
Thanks. I tried the docstring example and it's very useful!

PR #6542, opened 2024-08-19
Add `merge_points` and `cell_centers` options to `align_xyz` filter

user27182:
### Overview

New options to better control which points are used for the alignment. 

~Note: This is technically a breaking change since `merge_points=True` by default. But this filter is new to 0.45 and has not yet been released so this should not be an issue.~ `merge_points=False` is now the default, and we do `cell_centers=True` instead.


pyvista/core/filters/data_set.py
```diff
@@ -205,6 +205,8 @@ def align_xyz(
         axis_0_direction: VectorLike[float] | str | None = None,
         axis_1_direction: VectorLike[float] | str | None = None,
         axis_2_direction: VectorLike[float] | str | None = None,
+        merge_points: bool = True,
```
user27182, line 208: We can probably keep this option off by default

pyvista/core/filters/data_set.py
```diff
@@ -384,7 +401,17 @@ def _validate_vector(vector, name):
...
 
-        axes, std = pyvista.principal_axes(self.points, return_std=True)
+        # Get points and compute principal axes
+        input_mesh = self.cell_centers() if cell_centers else self
+        if merge_points:
+            append_filter = _vtk.vtkAppendFilter()
+            append_filter.MergePointsOn()
+            append_filter.AddInputData(input_mesh)
+            append_filter.Update()
+            points = pyvista.wrap(append_filter.GetOutput()).points  # type: ignore[union-attr]
```
user27182, line 411: Once #6551 is merged:
```suggestion
            points = input_mesh.merge_points().points
```

pyvista/core/filters/data_set.py
```diff
@@ -384,7 +401,17 @@ def _validate_vector(vector, name):
...
+        else:
+            points = input_mesh.points
```
user27182, line 413: Once #6551 is merged:
```suggestion
        points = input_mesh.merge_points().points if merge_points else input_mesh.points
```

pyvista/core/filters/data_set.py
```diff
@@ -384,7 +401,10 @@ def _validate_vector(vector, name):
                 vector = _validation.validate_array3(vector, dtype_out=float, name=name)
             return vector
 
-        axes, std = pyvista.principal_axes(self.points, return_std=True)
+        # Get points and compute principal axes
+        input_mesh = self.cell_centers() if cell_centers else self
+        points = input_mesh.merge_points().points if merge_points else input_mesh.points
+        axes, std = pyvista.principal_axes(points, return_std=True)
```
tkoyama010, line 407: It seems natural to me that use `cell_centers` defaults in my mind. I leave the final decision to you. I would like to add that since this method is not yet released, it is not a problem to change the default behavior.

tkoyama010 (dismissed):

pyvista/core/filters/data_set.py
user27182: Yes that does seem natural. It should also implicitly remove any real need to use `merge_points=True` for these cases since that's meant to address duplicates caused 
by having independent points for the cells. 

I'll make this option `None` by default and use cells if available, use points otherwise (e.g. PointSet)

banesullivan:
this looks like it can land? anything holding it back?

user27182:
IIRC there may have been flaky issues with the tests…? Probably worth reviving otherwise and see if tests are stable.

banesullivan (dismissed):

banesullivan (approved):

PR #6552, opened 2024-08-20
Add `fit_line_to_points` function

user27182:
### Overview

Similar to `fit_plane_to_points` but for a line.


tkoyama010 (approved):
LGTM. Awesome!

pyvista/core/utilities/points.py
```diff
@@ -346,6 +355,136 @@ def fit_plane_to_points(points, return_meta=False, resolution=10):
...
+        Fit a plane using the first two principal axes of the points.
+
+    principal_axes
+        Compute axes vectors which best fit a set of points.
+
+    Returns
+    -------
+    pyvista.PolyData
+        Line mesh.
+
```
user27182, line 400:
```suggestion

    numpy.ndarray
        Line length if ``return_meta=True``.

    numpy.ndarray
        Line direction (unit vector) if ``return_meta=True``.

```

tkoyama010 (approved):

pyvista/core/utilities/points.py
```diff
@@ -398,6 +398,12 @@ def fit_line_to_points(
     pyvista.PolyData
         Line mesh.
 
+    numpy.ndarray
```
user27182, line 401: Whoops, this should be a float, sorry about that!

tkoyama010 (dismissed):
I will enable auto merge after 24 hours if there is no other comment :)

tkoyama010 (approved):

PR #6553, opened 2024-08-21
Fix `cell_centers` with `PointSet`

user27182:
### Overview

`cell_centers` doesn't work with `PointSet`s. This PR fixes this. Discovered as part of #6542.


tests/core/test_dataset_filters.py
```diff
@@ -780,8 +780,8 @@ def test_compute_cell_sizes_composite(composite):
     assert output.n_blocks == composite.n_blocks
 
 
-def test_cell_centers(datasets):
-    for dataset in datasets:
+def test_cell_centers(datasets, airplane):
```
user27182, line 783: The `datasets` fixture should probably be updated to include a `PointSet` dataset. This can be a separate PR as it will likely cause failures (i.e. discover bugs) with other filters.

tkoyama010 (approved):
LGTM

PR #6561, opened 2024-08-24
Avoid sqrt of negative values for principal axes `std` computation

user27182:
### Overview

The eigenvalues can be negative. So we must use `abs` with `sqrt`.


tkoyama010 (approved):
LGTM

PR #6565, opened 2024-08-24
Document `CellType` enum members with figures

user27182:
### Overview

- [x] Allow documenting individual enum members (see https://enum-tools.readthedocs.io/en/latest/api/autoenum.html)
- [x] Add cell info
- [x] Add figures for linear cells
- [x] Format table 

Resolve #6549

The docstring text for the cells is adapted from each of the vtk docstrings for the respective cell, as well as information from the VTK book: https://book.vtk.org/en/latest/VTKBook/05Chapter5.html#

PR #6567, opened 2024-08-25
Make cell examples more visible

user27182:
### Overview

The [Line](https://dev.pyvista.org/api/examples/_autosummary/pyvista.examples.cells.line#pyvista.examples.cells.Line) and [PolyLine](https://dev.pyvista.org/api/examples/_autosummary/pyvista.examples.cells.polyline) examples are not very visible.  And I find the font size, line width, and sphere size are all a bit small relative to the size of the plot. 

This PR fixes this and makes these examples more visible. 
The intent is to include the images from these examples in the `CellType` docs as part of #6565 , so the content of the plots should ideally be sized to be easily understandable as smaller thumbnail images.


tkoyama010 (approved):
LGTM. Since this is a design of output demo images change, I decided no mention of the change in the documentation is required.

PR #6576, opened 2024-08-27
Add option to specify a plane mesh with plane-based filters

user27182:
### Overview
Add new `plane` parameter to the following filters:
- `clip`
- `slice`
- `clip_closed_surface`
- `project_points_to_plane`

This allows using an actual plane to implicitly define the `normal` and `origin` parameters instead of passing these explicitly.


tests/core/test_dataset_filters.py
```diff
@@ -285,6 +303,24 @@ def test_slice_filter(datasets):
     assert result.n_points < 1
 
 
+def test_slice_filter_plane(ant):
+    origin = (1, 2, 3)
+    normal = (4, 5, 6)
+    plane = pv.Plane(center=origin, direction=normal)
+    output_no_plane = ant.slice(origin=origin, normal=normal)
+    output_with_plane = ant.slice(plane=plane)
+    assert np.allclose(output_no_plane.bounds, output_with_plane.bounds)
```
user27182, line 312: This test is exactly the same for the four filters which have this feature. I can parametrize it instead into a single test if that's preferred.

user27182:
I'll add examples to the docstrings

user27182:
Replaced by #8151

PR #6579, opened 2024-08-28
Add operators to `PolyData` for boolean operations

tkoyama010:
### Overview

Add operators to `PolyData` for boolean operations.

```python
result = sphere_a | sphere_b
```
<p align="center">

</p>

```python
result = sphere_a & sphere_b
```
<p align="center">

### Details

- https://docs.python.org/3/library/operator.html


MatthewFlamm:
Before this is implemented, we may also want to be able to ask whether a mesh is falsey line so:

```py
if mesh1 and mesh2:  # if mesh1 and mesh2 both have cells/points
    # do something
```

This would conflict with your proposal.

user27182:
> Before this is implemented, we may also want to be able to ask whether a mesh is falsey like so:
> 
> ```python
> if mesh1 and mesh2:  # if mesh1 and mesh2 both have cells/points
>     # do something
> ```
> 
> This would conflict with your proposal.

I may be wrong, but I think `__and__` overloads `&` and `__or__` overloads `|` (bitwise operators).  So this PR implements boolean union with `sphere_a | sphere_b` and boolean intersection with `sphere_a & sphere_b`. The logical use of `and`  and `or` (truthiness of a mesh) should remain unaffected. 

I tested this with a debugger and adding a print statement inside `__or__`. `__or__` isn't called with `pv.Sphere() or pv.Sphere()` but is called with `pv.Sphere() | pv.Sphere()`. Not sure how the tests are passing though, might need to be tweaked, maybe add a plotting test as a visual sanity check

examples/01-filter/boolean-operations.py
```diff
@@ -28,9 +28,10 @@
 .. note::
    For merging, the ``+`` operator can be used between any two meshes
    in PyVista which simply calls the ``.merge()`` filter to combine
-   any two meshes.  This is different from ``boolean_union`` as it
-   simply superimposes the two meshes without performing additional
-   calculations on the result.
+   any two meshes.  This is different from the operator ``or`` in PyVista
+   which simply calls the ``boolean_union`` filter as it simply superimposes
+   the two meshes without performing additional calculations on the result.
+   The ``and`` operator in PyVista simply calls the ``boolean_intersection``.
```
user27182, line 34:
```suggestion
   any two meshes.  This is different from the operator ``|`` in PyVista
   which simply calls the ``boolean_union`` filter as it simply superimposes
   the two meshes without performing additional calculations on the result.
   The ``&`` operator in PyVista simply calls the ``boolean_intersection``.
```

examples/01-filter/boolean-operations.py
```diff
@@ -63,7 +64,7 @@
 # Order of operands does not matter for boolean union (the operation is
 # commutative).
 
-result = sphere_a.boolean_union(sphere_b)
+result = sphere_a or sphere_b
```
user27182, line 67:
```suggestion
result = sphere_a | sphere_b
```

examples/01-filter/boolean-operations.py
```diff
@@ -107,7 +108,7 @@
 # Order of operands does not matter for boolean intersection (the
 # operation is commutative).
 
-result = sphere_a.boolean_intersection(sphere_b)
+result = sphere_a and sphere_b
```
user27182, line 111:
```suggestion
result = sphere_a & sphere_b
```

pyvista/core/filters/poly_data.py
```diff
@@ -139,6 +139,9 @@ def boolean_union(self, other_mesh, tolerance=1e-5, progress_bar=False):
         .. versionchanged:: 0.32.0
            Behavior changed to match default VTK behavior.
 
+        .. versionchanged:: 0.45.0
+           Define ``or`` operator to perform boolean union.
```
user27182, line 143:
```suggestion
           Define ``|`` operator to perform boolean union.
```

pyvista/core/filters/poly_data.py
```diff
@@ -164,7 +167,7 @@ def boolean_union(self, other_mesh, tolerance=1e-5, progress_bar=False):
         >>> import pyvista as pv
         >>> sphere_a = pv.Sphere()
         >>> sphere_b = pv.Sphere(center=(0.5, 0, 0))
-        >>> result = sphere_a.boolean_union(sphere_b)
+        >>> result = sphere_a or sphere_b
```
user27182, line 170:
```suggestion
        >>> result = sphere_a | sphere_b
```

pyvista/core/filters/poly_data.py
```diff
@@ -212,6 +215,9 @@ def boolean_intersection(self, other_mesh, tolerance=1e-5, progress_bar=False):
 
         .. versionadded:: 0.32.0
 
+        .. versionchanged:: 0.45.0
+           Define ``and`` operator to perform boolean intersection.
```
user27182, line 219:
```suggestion
           Define ``&`` operator to perform boolean intersection.
```

pyvista/core/filters/poly_data.py
```diff
@@ -237,7 +243,7 @@ def boolean_intersection(self, other_mesh, tolerance=1e-5, progress_bar=False):
         >>> import pyvista as pv
         >>> sphere_a = pv.Sphere()
         >>> sphere_b = pv.Sphere(center=(0.5, 0, 0))
-        >>> result = sphere_a.boolean_intersection(sphere_b)
+        >>> result = sphere_a and sphere_b
```
user27182, line 246:
```suggestion
        >>> result = sphere_a & sphere_b
```

user27182 (commented):
Replace `and` with `&` and `or` with `|`

user27182:
> When I did the `and` and `or` operations, it looked like this. 

I am unable to reproduce those plots locally. Working from main dev branch  I get:
``` python
(pv.Sphere() or pv.Cone()).plot()  # only Sphere is shown
```
```python
(pv.Sphere() and pv.Cone()).plot()  # only Cone is shown
```


Using pure boolean logic I would have expected these expressions to fail with an `AttributeError` from doing `True.plot()`. But in the python truthiness sense I guess this makes sense for the `or` case since the logic short-circuits with `pv.Sphere()` evaluating to `True`, so `pv.Cone()` never executes (and we're left with a value of `pv.Sphere()`, not `True`). Not too sure how to make sense of the `and` case, maybe python returns the right argument by convention for this?

@tkoyama010 Is this what you mean by "unnatural behavior", or are you referring to something else?

tkoyama010:
I call what you are describing in your comment “unnatural behavior”. I am unable to explain this behavior well.

> I would have expected these expressions to fail with an AttributeError from doing True.plot()

This is what I am looking for in the results.

user27182:
> Using pure boolean logic I would have expected these expressions to fail with an `AttributeError` from doing `True.plot()`. But in the python truthiness sense I guess this makes sense for the `or` case since the logic short-circuits with `pv.Sphere()` evaluating to `True`, so `pv.Cone()` never executes (and we're left with a value of `pv.Sphere()`, not `True`). Not too sure how to make sense of the `and` case, maybe python returns the right argument by convention for this?


As a follow-up, this appears to be consistent with the general truthiness of objects in python. E.g. we get similar behavior with a numpy scalar:
``` python
import numpy as np
(np.array((1)) and np.array(2) and np.array(3))
# Out[2]: array(3)
(np.array((1)) or np.array(2) or np.array(3))
# Out[3]: array(1)
```

tkoyama010:
I see. It seems that this is not a issue with PyVista, but with Python. I am wondering what happens when you use and and or with values other than bool. I'll ask with a friend who knows a lot about Python.

tkoyama010:
OK. I found it.
> The expression x and y first evaluates x; if x is false, its value is returned; otherwise, y is evaluated and the resulting value is returned.
> The expression x or y first evaluates x; if x is true, its value is returned; otherwise, y is evaluated and the resulting value is returned.

(https://docs.python.org/3/reference/expressions.html#boolean-operations)

user27182:
Weird. Doesn't seem like we can override this either. E.g. even if we implement this for PolyData:
``` python
def __bool__(self):
    return False
```
we still get this behavior:
``` python
bool(pv.Sphere() or pv.Cone())  # False
(pv.Sphere() or pv.Cone())  # we get Cone()
```

A safe option to guard against this (at least for the PyVista codebase) is to enable `truthy-bool` error codes with `mypy`:
https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-expression-is-not-implicitly-true-in-boolean-context-truthy-bool

tests/plotting/test_plotting.py
```diff
@@ -911,7 +911,7 @@ def test_legend_circle_face(sphere):
     plotter.add_mesh(sphere)
     legend_labels = [['sphere', 'r']]
     face = "circle"
-    _ = plotter.add_legend(
+    plotter.add_legend(
```
tkoyama010, line 914:
```suggestion
    _ = plotter.add_legend(
```

tests/plotting/test_plotting.py
```diff
@@ -926,7 +926,7 @@ def test_legend_rectangle_face(sphere):
     plotter.add_mesh(sphere)
     legend_labels = [['sphere', 'r']]
     face = "rectangle"
-    _ = plotter.add_legend(
+    plotter.add_legend(
```
tkoyama010, line 929:
```suggestion
    _ = plotter.add_legend(
```

tests/plotting/test_plotting.py
```diff
@@ -1078,7 +1078,7 @@ def test_set_background():
     plotter.set_background('k')
     plotter.background_color = "yellow"
     plotter.set_background([0, 0, 0], top=[1, 1, 1])  # Gradient
-    _ = plotter.background_color
+    plotter.background_color
```
tkoyama010, line 1081:
```suggestion
    _ = plotter.background_color
```

tests/plotting/test_plotting.py
```diff
@@ -4352,7 +4352,7 @@ def test_create_axes_orientation_box():
         show_text_edges=True,
     )
     plotter = pv.Plotter()
-    _ = plotter.add_actor(actor)
+    plotter.add_actor(actor)
```
tkoyama010, line 4355:
```suggestion
    _ = plotter.add_actor(actor)
```

tkoyama010:
@pyvista-bot LGTM

PR #6580, opened 2024-08-28
Fix names when using `MultiBlock.transform`

user27182:
### Overview

`transform` for `MultiBlock` uses integer indexing with `enumerate` to update the blocks. It turns out that indexing this way does not preserve the block name, e.g.:
``` python
import pyvista as pv
poly = pv.PolyData()
m = pv.MultiBlock(dict(data=poly)) 

# Block name is 'data'
m.keys() # ['data'] 

# Block name is modified
m[0] = poly
m.keys() # ['Block-00']
```

And so the filter unexpectedly mutates the block names as a result. This PR fixes this by indexing by name instead so that output names (keys) match the input names.

### DIFF
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -450,7 +450,7 @@ def test_transform_filter(ant, sphere, airplane, tetbeam, inplace):
     nested = multi_from_datasets(airplane, tetbeam)
     multi.append(nested)
     multi.append(None)
-    for i in enumerate(multi):
+    for i, _ in enumerate(multi):
         multi.set_block_name(i, str(i))
 
     NUMBER = 42
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -450,7 +450,7 @@ def test_transform_filter(ant, sphere, airplane, tetbeam, inplace):
     nested = multi_from_datasets(airplane, tetbeam)
     multi.append(nested)
     multi.append(None)
-    for i in enuerate(multi):
+    for i in enumerate(multi):
         multi.set_block_name(i, str(i))
 
     NUMBER = 42
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -450,11 +450,14 @@ def test_transform_filter(ant, sphere, airplane, tetbeam, inplace):
     nested = multi_from_datasets(airplane, tetbeam)
     multi.append(nested)
     multi.append(None)
+    for i in range(len(multi)):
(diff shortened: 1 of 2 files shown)


tests/core/test_composite.py
```diff
@@ -450,23 +450,28 @@ def test_transform_filter(ant, sphere, airplane, tetbeam, inplace):
     nested = multi_from_datasets(airplane, tetbeam)
     multi.append(nested)
     multi.append(None)
+    for i in range(len(multi)):
+        multi.set_block_name(i, str(i))
```
tkoyama010, line 454:
```suggestion
    for i in enuerate(multi):
        multi.set_block_name(i, str(i))
```

tkoyama010 (changes requested):

tkoyama010 (dismissed):
LGTM

tkoyama010 (approved):

tests/core/test_composite.py
```diff
@@ -450,23 +450,28 @@ def test_transform_filter(ant, sphere, airplane, tetbeam, inplace):
...
     # Do test
     output = multi.transform(
         transform, inplace=inplace, transform_all_input_vectors=False, progress_bar=False
     )
     bounds_after = np.array(output.bounds)
     n_blocks_after = output.n_blocks
+    keys_after = output.keys()
 
     assert (output is multi) == inplace
     assert [(block_in is block_out) == inplace for block_in, block_out in zip(multi, output)]
```
user27182, line 471: Not directly related to this PR, but this looks like a truthy-object (always asserts True). Missing `all`
```suggestion
    assert all((block_in is block_out) == inplace for block_in, block_out in zip(multi, output))
```

tkoyama010: Good catch. It seems that we have more. Could you open new PR for this?
```bash
$ grep "assert \[" ./tests/**/*.py
./tests/core/test_composite.py:    assert [(block_in is block_out) == inplace for block_in, block_out in zip(multi, output)]
./tests/examples/test_dataset_loader.py:    assert [os.path.isabs(file) for file in path]
./tests/examples/test_dataset_loader.py:    assert [os.path.isabs(file) for file in path_loadable]
./tests/examples/test_dataset_loader.py:    assert [os.path.isfile(file) for file in path_download]
./tests/examples/test_dataset_loader.py:    assert [
./tests/examples/test_dataset_loader.py:    assert [
```

tkoyama010: I fixed it in #6588. I added you as co-author.

PR #6588, opened 2024-08-29
Fix assert list line which is always raise true

tkoyama010:
### Overview

Fix assert list line which is always raise true.

### Details

- https://github.com/pyvista/pyvista/pull/6580#discussion_r1735473852


tests/core/test_composite.py
```diff
@@ -468,7 +468,7 @@ def test_transform_filter(ant, sphere, airplane, tetbeam, inplace):
     keys_after = output.keys()
 
     assert (output is multi) == inplace
-    assert [(block_in is block_out) == inplace for block_in, block_out in zip(multi, output)]
+    assert all((block_in is block_out) == inplace for block_in, block_out in zip(multi, output))
```
user27182, line 471: This test is failing. `MultiBlock.transform` will need to be updated. I explored using `copy` as part of the solution but there are issues with `copy`, see #6590

user27182: Another issue with `__next__` was found which was causing this to fail by using `zip`. See #6600

tkoyama010:
@pyvista-bot LGTM

PR #6590, opened 2024-08-30
Fix `MultiBlock.copy()` for `None` and nested datasets and use common test fixture

user27182:
### Overview

Fix bug with deep copy where the names of `None` blocks are not copied
``` python
import pyvista as pv
multi = pv.MultiBlock([None])
print(multi.keys())  # ['Block-00']
copied = multi.copy(deep=True)
print(copied.keys())  # [None]
```

Fix bug with shallow copy where nested multiblocks are copied as new instances 
``` python
import pyvista as pv
nested = pv.MultiBlock()
multi = pv.MultiBlock([nested])
copied = multi.copy(deep=False)
print(copied[0] is nested)  # False, but should be True
```

The composite tests are also generally updated to use a common multiblock fixture (where it makes sense to do so). Test coverage is also improved since this fixture includes `None` and nested blocks whereas most tests did not include these previously


user27182:
I can narrow the scope to only include the bug fixes first, and do a follow up PR to update the fixtures if that's preferred

tests/conftest.py
```diff
@@ -213,6 +213,14 @@ def multiblock_all(datasets):
     return pyvista.MultiBlock(datasets)
 
 
+@pytest.fixture()
+def multiblock_all_with_nested_and_none(datasets, multiblock_all):
+    """Return datasets fixture combined in a pyvista multiblock."""
+    # nested = multiblock_all
+    # nested.append(None)
+    return pyvista.MultiBlock([*datasets, None, multiblock_all])
```
tkoyama010, line 221:
```suggestion
    return pyvista.MultiBlock([*datasets, None, multiblock_all])
```

tkoyama010 (changes requested):

tests/conftest.py
user27182: This has been resolved with [b4b71c0](https://github.com/pyvista/pyvista/pull/6590/commits/b4b71c0718e8dff75fc007f840d756d0e41191e3)

tkoyama010: Thank you. I should have checked with you before making the suggestion.

user27182: I probably should have marked it as draft initially. I appreciate the quick turnaround though, so thanks for the initial comments

tkoyama010 (approved):
LGTM

PR #6592, opened 2024-08-30
Add colors from `vtkNamedColors`

user27182:
### Overview

Resolve #6591.

Add approx 60 new colors from https://vtk.org/doc/nightly/html/classvtkNamedColors.html#details.

Only new colors from this class are added.`vtkNamedColors` which have the same name or value as existing colors are excluded.

PR #6593, opened 2024-08-30
Add tests for conformance of pyvista colors to the web color standard

user27182:
### Overview

Follow-up to #6592. 

- Make sure colors match the web standard: https://www.w3.org/TR/css-color-4/#named-colors 
- `brown` was found to be non-conformant and is modified by this PR.
- Tests for color synonyms and color uniqueness are also added.

EDIT: These are the same colors referred to as CSS4 in Matplotlib:
https://matplotlib.org/stable/gallery/color/named_colors.html#css-colors


tests/plotting/test_colors.py
```diff
@@ -121,3 +122,187 @@ def test_color():
 def test_color_opacity():
     color = pv.Color(opacity=0.5)
     assert color.opacity == 128
+
+
+# Define standardized names and hex values of web colors defined here:
+# https://www.w3.org/TR/css-color-4/#named-colors
+WEB_COLORS: final = {
+    'aliceblue': '#F0F8FF',
```
tkoyama010, line 130: It's hard to check if these color codes are correct, so it's a good idea to use the Matplotlib function. However, we need to check that Matplotlib covers all of the codes here.
```python
>>> from matplotlib import colors
>>> print(colors.cnames['aliceblue'])
#F0F8FF
```

tkoyama010 (changes requested):

tests/plotting/test_colors.py
user27182: Fixed in https://github.com/pyvista/pyvista/pull/6593/commits/fe593cf7249433a96b275ba5fc3be392fbb6f98e

tests/plotting/test_colors.py
```diff
@@ -121,3 +122,33 @@ def test_color():
...
+
+
+def test_css4_colors(css4_color):
+    name, value = css4_color
+    assert pv.Color(name).hex_rgb.lower() == value.lower()
+
+
+def test_color_synonyms(color_synonym):
+    color = pv.Color(color_synonym)
+    assert color.name
```
tkoyama010, line 148:
```suggestion
    assert isinstance(color, pv.Color)
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

tkoyama010 (approved):

PR #6369, opened 2024-07-11
Change documentation theme to sphinx-book-theme

banesullivan:
These changes switch us to sphinx-book-theme which may work a bit better for our content and sections. Feedback welcome here! Which theme do you like better?

| Before | After |
|---|---|
| PyVista _ PyVista 0_44_1 documentation | PyVista _ PyVista 0_45_dev0 documentation · 10 17pm · 07-22 | 
| Examples _ PyVista 0_44_1 documentation | Examples _ PyVista 0_45_dev0 documentation · 10 17pm · 07-22 |

PR #6595, opened 2024-08-31
Move color table to its own page and use dynamically-generated rst files

user27182:
### Overview

- Remove redundant repo copy of `color_table.rst` and use the dynamically generated copy instead.
- Remove redundant list of colors in module docstring
- Make standalone `Colors` section in docs
- Add cross-refs to `color_table` in docs
- ~Add dropdown back to `Color` docs which was removed by https://github.com/pyvista/pyvista/pull/6369. WIthout a dropdown, users have to scroll through 150+ rows of the table to get to the API docs at the bottom.~ Table is moved to a separate page which resolves this issue


tkoyama010 (approved):
Awesome! We can now generate the page of color table :)
image

PR #6570, opened 2024-08-26
Add quadratic cell examples

user27182:
### Overview

Related to #6549.

Add cell examples for:

- QuadraticEdge
- QuadraticTriangle
- QuadraticQuadrilateral
- QuadraticTerahedron
- QuadraticHexahedron
- QuadraticWedge

See:
https://examples.vtk.org/site/Cxx/GeometricObjects/IsoparametricCellsDemo/

~This PR also includes changes from #6567 for the plots.~  #6567 is merged

PR #6596, opened 2024-08-31
Add more quadratic cell examples

user27182:
### Overview

More examples. Follow-up to #6570. 

- QuadraticPolygon 
- QuadraticPyramid 
- QuadraticLinearQuadrilateral 
- QuadraticLinearWedge
- BiQuadraticQuadraticWedge


pyvista/examples/cells.py
```diff
@@ -986,6 +986,89 @@ def QuadraticQuadrilateral() -> UnstructuredGrid:
...
+    #     [0.0, 2.0, 0.0],
+    #     [1.0, 0.0, 0.0],
+    #     [2.0, 1.0, 0.0],
+    #     [1.0, 2.0, 0.0],
+    #     [0.0, 1.0, 0.0],
+    #     [3.0, 1.0, 0.0],
+    # ]
+    #
+    # cells = [len(points), *list(range(len(points)))]
+    # return UnstructuredGrid(cells, [CellType.QUADRATIC_POLYGON], points)
```
tkoyama010, line 1041: Let's move this to an issue and add a link to the issue in the comments. I want to create an issue here, but you can create issues about this better than me in my mind.

tkoyama010 (changes requested):

pyvista/examples/cells.py
user27182: I found a bug with the example. They set 9 points but the cell only has 8. This resulted in 9 points incorrectly defined in the pyvista version. Fixed in https://github.com/pyvista/pyvista/pull/6596/commits/5cdda9dc797e09588e173edb4e50e1e2b72718c3

tkoyama010 (approved):
LGTM

PR #6599, opened 2024-08-31
Fix MultiBlock transform inplace

user27182:
Fix to resolve failed tests from https://github.com/pyvista/pyvista/pull/6588#discussion_r1737609116


tests/core/test_composite.py
```diff
@@ -439,7 +440,8 @@ def test_transform_filter(ant, sphere, airplane, tetbeam, inplace):
     keys_after = output.keys()
 
     assert (output is multi) == inplace
-    assert [(block_in is block_out) == inplace for block_in, block_out in zip(multi, output)]
+    for i, _ in enumerate(multi):
+        assert (multi[i] is output[i]) == inplace or (multi[i] is None)
```
tkoyama010, line 444:
```suggestion
    for block_in, block_out in zip(multi, output):
        assert (block_in is block_out) == inplace or (block_in is None)
```

tkoyama010 (approved):
Thanks a lot!

tkoyama010:
Sorry I  missed https://github.com/pyvista/pyvista/issues/6600#issuecomment-2322980261. Reverting.

tests/core/test_composite.py
user27182: Using zip like this won't work for MultiBlock because of #6600. So we have to iterate without zip, e.g. with `enumerate` so that __getitem__ is used internally instead of __next__

tkoyama010 (approved):

tests/core/test_composite.py
tkoyama010:
```suggestion
    # https://github.com/pyvista/pyvista/pull/6599/files#r1739931261
    for i, _ in enumerate(multi):
        assert (multi[i] is output[i]) == inplace or (multi[i] is None)
```

tkoyama010 (approved):

tests/core/test_composite.py
tkoyama010: ref #6644

PR #6607, opened 2024-09-02
Fix pytest fixture incorrect parentheses style

tkoyama010:
### Overview

- follow-up https://github.com/pyvista/pyvista/pull/6606
- see https://docs.astral.sh/ruff/rules/pytest-fixture-incorrect-parentheses-style/

### Details

- None

PR #6608, opened 2024-09-02
Fix unconventional import alias of ElementTree

tkoyama010:
### Overview

- follow-up #6607
- see https://docs.astral.sh/ruff/rules/unconventional-import-alias/

### Details

- None


tkoyama010:
@pyvista-bot LGTM

Issue #6609, opened 2024-09-03
Include end bound in `pv.voxelize()` and `pv.voxelize_volume()` to avoid cropping the mesh


GuillaumeBroggi:
### Describe the feature you would like to be added.

Voxelizing a mesh with `pv.voxelize()` or `pv.voxelize_volume()` will crop the mesh. 

For instance, for a simple cube with the default parameters, the mesh bounds are `[-0.5, 0.5]`: 

```python
import pyvista as pv
cube = pv.Cube()
cube
```
<table style='width: 100%;'><tr><th>Header</th><th>Data Arrays</th></tr><tr><td>
<table style='width: 100%;'>
<tr><th>PolyData</th><th>Information</th></tr>
<tr><td>N Cells</td><td>6</td></tr>
<tr><td>N Points</td><td>8</td></tr>
<tr><td>N Strips</td><td>0</td></tr>
<tr><td>X Bounds</td><td>-5.000e-01, 5.000e-01</td></tr>
<tr><td>Y Bounds</td><td>-5.000e-01, 5.000e-01</td></tr>
<tr><td>Z Bounds</td><td>-5.000e-01, 5.000e-01</td></tr>
<tr><td>N Arrays</td><td>3</td></tr>
</table>

</td><td>
<table style='width: 100%;'>
<tr><th>Name</th><th>Field</th><th>Type</th><th>N Comp</th><th>Min</th><th>Max</th></tr>
<tr><td>Normals</td><td>Points</td><td>float32</td><td>3</td><td>-1.000e+00</td><td>1.000e+00</td></tr>
<tr><td>TCoords</td><td>Points</td><td>float32</td><td>2</td><td>-1.000e+00</td><td>1.000e+00</td></tr>
<tr><td>FaceIndex</td><td>Cells</td><td>int64</td><td>1</td><td>0.000e+00</td><td>5.000e+00</td></tr>
</table>

</td></tr> </table>

After voxelization, the mesh bounds are `[-0.5, 0.3]`, the missing `0.2` corresponding to a missing row of cells: 

```python
pv.voxelize(cube, density=0.2)
```
<table style='width: 100%;'><tr><th>Header</th><th>Data Arrays</th></tr><tr><td>
<table style='width: 100%;'>
<tr><th>UnstructuredGrid</th><th>Information</th></tr>
<tr><td>N Cells</td><td>64</td></tr>
<tr><td>N Points</td><td>125</td></tr>
[...]

Issue #6255, opened 2024-06-14
voxelize_volume


MosGeo:
### Describe the feature you would like to be added.

It would be great to have the ability to voxelize_volume and voxelize to create labeled volume rather than a binary volume. I see this being used in conjection with MultiBlock. I want each block to have its own number.

If there is a workaround to do this right now, that would be very helpful.

### Pseudocode or Screenshots

I see the api as something like this:
```python
mesh = pv.MultiBlock(...)
labeled_volume = pv.voxelize_volume(mesh, label=True)
binary_volume = pv.voxelize_volume(mesh, label=False)
```
Obviously, the default for label parameter is False.

user27182:
Great idea! Here's a possible way to implement this for `voxelize` using the `sample` and `split_values` filters. See code/images below. 

A few notes about this method:

- `sample` seems to only operate on point data, not sure how to make it work directly with cell data. Using `ptc` seems is a bit hacky because it means the data will be interpolated twice.
- the tolerance for `sample` must be sufficiently large, (maybe set it to `inf`? or just really large)
- `split_values` is a new filter (`pyvista==0.44.0`), currently only available from the `main` (dev) branch.


``` python
import pyvista as pv
from pyvista import examples
import numpy as np

# 1. Load mesh and add cell data labels
mesh = examples.download_bunny_coarse().clean()
labels = np.zeros((mesh.n_cells,))
labels[:700] = 1
mesh['labels'] = labels
mesh.plot()

# 2. Voxelize and sample
vox = pv.voxelize(mesh, density=0.01)
sampled = vox.sample(target=mesh, tolerance=1)
sampled.set_active_scalars('labels')
sampled.plot(show_edges=True)

# 3. We want cell data not point data
sampled = sampled.ptc()
sampled['labels'] = np.round(sampled['labels'])
sampled.plot(show_edges=True)

# 4. Split into multiblock
multiblock = sampled.split_values()

multiblock[0].plot()
# Also multiblock[1].plot(show_edges=True)
```
Output:
1. Inital labeled mesh:
<img width="500" alt="image" src="https://github.com/pyvista/pyvista/assets/89109579/3c8ce49c-bdf4-4186-b441-824f0ce68f37">

2. Sampled voxelized mesh. The cell data is interpolated as point data (bad), but we want cell data
<img width="500" alt="image" src="https://github.com/pyvista/pyvista/assets/89109579/9e81a1d5-f022-4154-a38f-cfda0bdec4af">

[...]

PR #6611, opened 2024-09-04
Add `voxelize_binary_mask` filter to generate `ImageData` from `PolyData`

### LINKED ISSUE #4312: Create a voxelize feature that takes (n_x, n_y, n_z) as parameters instead of just density

user27182:
### Overview

In medical image processing, it's often necessary to go from image (e.g. labelmap or binary segmentation mask) to 3D surface (e.g. using marching cubes or surface nets), do some processing, and then convert the surface back to an image. A key detail is that we want the output image to be identical in every way to the original image but with the labelmap/mask updated to reflect any changes made to the polydata surface. I.e. we want to preserve the input image extent, spacing, and orientation.

In PyVista we have `contour_labeled` to go from image to surface, but are missing methods to convert back to an image. This PR adds this new method.

TODO:
- [x] Improve API
- ~[ ] Support for multiple labels~ Will be a separate filter / PR
- [x] Better test coverage

### Details

This method is similar to `voxelize_volume` with a few key differences:
- `voxelize_volume` returns a `RectilinearGrid`; this new method returns `ImageData`
- `voxelize_volume` assumes the input image data is axis-aligned; this new method takes into account image orientation
- `voxelize_volume` only allows specifying density; this new method also allows setting `dimensions` (i.e. closes #4312)
- `voxelize_volume` uses a default spacing (density) of `mesh.length/100`; this new method uses the 10th percentile of the input's cell lengths
- `voxelize_volume` does not consider labeled input data (see #6255); ~this new method aims to support labeled inputs~ separate PR for this
- `voxelize_volume` uses `select_enclosed_points` internally which limits its used to closed surfaces only; this new method does not have this limitation
- `voxelize_volume` uses a custom implementation with `np.arange` or `np.linspace` which can cause issues with generating the correct bounds (e.g. see #6609); this new method uses vtk methods instead (using image stencils)
- The implementation is based on the "ModelToLabelMap" function from 3D Slicer, which is the gold standard (IMO) for processing medical data with VTK:
https://github.com/Slicer/Slicer/blob/7727475e9a771e6d509c994a4b1e4daa08cf0f8c/Modules/CLI/ModelToLabelMap/ModelToLabelMap.cxx


GuillaumeBroggi:
I tried this new method to voxelize a mesh. As mentioned in #1743, it seems superior to the existing approaches as it does not produce stray voxels and is significantly faster.

Instead of raising an error when `reference_volume` is `None`, what about automatically generating a reference volume based on the bounds of the input mesh? I can write a commit if this seems like a good idea.

tkoyama010:
> I tried this new method to voxelize a mesh. As mentioned in #1743, it seems superior to the existing approaches as it does not produce stray voxels and is significantly faster.
> 
> Instead of raising an error when `reference_volume` is `None`, what about automatically generating a reference volume based on the bounds of the input mesh? I can write a commit if this seems like a good idea.

@GuillaumeBroggi Thanks for your thought. You can create Pull Request for this branch if you want :)

user27182:
That's a good suggestion for improving the API. The current error is based on the Slicer implementation that was used as an initial reference. Still a few artifacts left over from porting that over. 

The hold up for this PR at the moment is with how images are transformed in PyVista , see #6616. Once resolved I'll work on improving this PR

user27182:
@GuillaumeBroggi The filter has been updated to no longer require a reference volume, e.g.:

``` python
from pyvista import examples
import pyvista as pv
poly = examples.load_ant()
labelmap = poly.generate_labelmap()

pl = pv.Plotter()
pl.add_mesh(labelmap.points_to_cells().threshold(0.5), color='blue')
pl.add_mesh(poly, color='lime')
pl.show()
```
Voxelized is blue, input poly data is green:
ant

Feel free to review the code and suggest changes or improvements. (Also note that it's still a draft and I may also push changes)

pyvista/core/filters/poly_data.py
```diff
@@ -3892,3 +3897,111 @@ def protein_ribbon(self, progress_bar=False):
         alg.SetInputData(self)
         _update_alg(alg, progress_bar, "Generating Protein Ribbons")
         return _get_output(alg)
+
+    def generate_labelmap(
```
user27182, line 3901: Looking for feedback on the name. The term `labelmap` is kind of domain specific to medical images, maybe something more general is better?

This method is basically the inverse surface nets from `contour_labeled`.  So we could call it something like `contour_to_labeled` to better relate the two? 

Or, with #5968, we also have the option to name _both_ filters, e.g. rename `contour_labeled` -> `image_labels_to_polydata`, and name the filter from this PR `polydata_to_image_labels`

Or, even still, we could better align the name with `voxelize` and `voxelize_volume`, and maybe call this filter `voxelize_labels`?

user27182:
With [d4eedba](https://github.com/pyvista/pyvista/pull/6611/commits/d4eedba1b83587e559cd517d295da341b6df3f8e), the filter now dynamically sets the spacing based on the input cell length. This differs from `voxelize_volume` which uses `mesh.length/100` by default.
 
E.g. Using the same code snippet above, but with `download_bunny_coarse`:
image
And again but with full-res `download_bunny`:
image

And once again with the ant. This one is a good test because it has many large cells for the body, and tiny cells for the antennae:
image

pyvista/core/filters/poly_data.py
GuillaumeBroggi: Coming from material sciences, `labelmap` is not the usual terminology, although some literature uses `labelling`.

`voxelize` is more common in this field but does not carry the idea that voxels are labeled. Another one could be `segmentation`.

user27182:
> @user27182 I reviewed the code, and it looks great, much better than I would have done.

Thanks for the feedback!

> In my implementation, I had chosen to use `**kwargs` in case the user wants to use a custom spacing or custom origin. I think your proposition brings more functionality.

Great, so the current option to specify `spacing` or `dimensions` separately is good? I originally considered adding `origin` explicitly too, but it gets too complicated since it also needs to be rounded for the grid to match the bounds correctly. If desired, users can specify `origin` implicitly using a reference volume instead.

> `spacing_bound` is a bit misleading IMO since the code rounds `dimensions`. See #6647 for a different implementation where `rounding_func` replaces `spacing_bound` and accepts any callable. This would provide more flexibility and is less verbose.

I agree with changing `spacing_bounds`.  #6647 looks good to me.

> Also, there is a floating point error when casting the results to a `RectilinearGrid`, but I am not sure it is possible to do anything about it

It appears the error stems from the casting to rectilinear grid, perhaps this function can be adjusted (in a separate PR) to remove the errror?
https://github.com/pyvista/pyvista/blob/e1b3f1559c1a95add0f9013d48005f79fb9d3905/pyvista/core/grid.py#L845-L849

GuillaumeBroggi:
> Great, so the current option to specify `spacing` or `dimensions` separately is good?

Yes, it works nicely like that IMO.

> It appears the error stems from the casting to rectilinear grid, perhaps this function can be adjusted (in a separate PR) to remove the error?

Thanks for pointing it out, I'll try to correct the error.

pyvista/core/filters/poly_data.py
user27182: Yes I think the term `label` is perhaps a bit vague/non-obvious outside of specific fields. Maybe `voxelize_binary_mask`  is good? The method currently only works for binary masks anyway, I'm thinking a multi-label version is probably better suited for a separate filter (and separate PR), maybe call that one `voxelize_label_mask`

user27182:
I'd say this is ready for review. Docs _might_ still fail from minor RST formatting, but otherwise is feature complete.

pyvista/core/filters/poly_data.py
```diff
@@ -3892,3 +3898,403 @@ def protein_ribbon(self, progress_bar=False):
...
+        If no inputs are provided, the 10th ``cell_length_percentile`` is used by
+        default on systems with VTK >= 9.2. For older versions of VTK, the default
+        spacing is ``mesh_length_fraction=1/100``.
+
+        .. note::
+            For best results, ensure the input surface is a closed surface. The
+            surface is considered closed if it has zero :attr:`pyvista._PointSet.n_open_edges`.
+
+        .. note::
+            This filter returns voxels represented as point data, not :attr:`pyvista.CellTypes.VOXEL` cells..
```
user27182, line 3949:
```suggestion
            This filter returns voxels represented as point data, not :attr:`pyvista.CellTypes.VOXEL` cells.
```

pyvista/core/filters/poly_data.py
```diff
@@ -3892,3 +3898,403 @@ def protein_ribbon(self, progress_bar=False):
...
+        pyvista.ImageData
+            Generated binary mask with a ``'mask'``  point data array. The data array
+            has dtype :class:`numpy.uint8` if the foreground and background values are
+            unsigned and less than 256.
+
+        See Also
+        --------
+        pyvista.voxelize
+            Similar function that returns a :class:`pyvista.UnstructuredGrid` of
+            :attr:`pyvista.CellType.VOXEL` cells.
```
user27182, line 4034:
```suggestion
            :attr:`~pyvista.CellType.VOXEL` cells.
```

pyvista/core/filters/poly_data.py
```diff
@@ -3892,3 +3898,403 @@ def protein_ribbon(self, progress_bar=False):
...
+
+        pyvista.voxelize_volume
+            Similar function that returns a :class:`pyvista.RectilinearGrid` with cell data.
+
+        pyvista.ImageDataFilters.contour_labeled
+            Filter that generates surface contours from labeled image data. Can be
+            loosely considered as the inverse of this filter.
+
+        pyvista.ImageDataFilters.points_to_cells
+            Convert voxels represented as points to :attr:`pyvista.CellType.VOXEL`
```
user27182, line 4044:
```suggestion
            Convert voxels represented as points to :attr:`~pyvista.CellType.VOXEL`
```

pyvista/core/filters/poly_data.py
```diff
@@ -3892,3 +3898,403 @@ def protein_ribbon(self, progress_bar=False):
...
+            # Get size of poly data for computing dimensions
+            bnds = self.bounds  # type: ignore[attr-defined]
+            x_size = bnds.x_max - bnds.x_min
+            y_size = bnds.y_max - bnds.y_min
+            z_size = bnds.z_max - bnds.z_min
+            sizes = np.array((x_size, y_size, z_size))
+
+            if dimensions is None:
+                dimensions = np.array(rounding_func(sizes / initial_spacing), dtype=int)
+
```
user27182, line 4227:
```suggestion
            # Get initial spacing (will be adjusted later)
            initial_spacing = _validation.validate_array3(spacing, broadcast=True)

            # Get size of poly data for computing dimensions
            bnds = self.bounds  # type: ignore[attr-defined]
            x_size = bnds.x_max - bnds.x_min
            y_size = bnds.y_max - bnds.y_min
            z_size = bnds.z_max - bnds.z_min
            sizes = np.array((x_size, y_size, z_size))

            if dimensions is None:
                dimensions = np.array(rounding_func(sizes / initial_spacing), dtype=int)

            reference_volume = pyvista.ImageData()
```

pyvista/core/utilities/features.py
```diff
@@ -83,6 +83,18 @@ def voxelize(mesh, density=None, check_surface=True, enclosed=False, fit_bounds=
     Prior to version 0.39.0, this method improperly handled the order of
     structured coordinates.
 
+    See Also
+    --------
+    pyvista.voxelize
+        Similar function that returns a :class:`pyvista.UnstructuredGrid` of
+        :attr:`pyvista.CellType.VOXEL` cells.
```
user27182, line 90:
```suggestion
        :attr:`~pyvista.CellType.VOXEL` cells.
```

pyvista/core/utilities/features.py
```diff
@@ -246,6 +258,12 @@ def voxelize_volume(mesh, density=None, check_surface=True, enclosed=False, fit_
     See Also
     --------
     pyvista.voxelize
+        Similar function that returns a :class:`pyvista.UnstructuredGrid` of
+        :attr:`pyvista.CellType.VOXEL` cells.
```
user27182, line 262:
```suggestion
        :attr:`~pyvista.CellType.VOXEL` cells.
```

GuillaumeBroggi (approved):
Looks good to me, that's a handy new feature.

If I am not mistaken there is a circular reference in docs, see #6651

tkoyama010 (approved):
LGTM

GuillaumeBroggi:
> It appears the error stems from the casting to rectilinear grid, perhaps this function can be adjusted (in a separate PR) to remove the errror?
> 
> https://github.com/pyvista/pyvista/blob/e1b3f1559c1a95add0f9013d48005f79fb9d3905/pyvista/core/grid.py#L845-L849

Fix proposed in #6656

Issue #6213, opened 2024-06-06
Fix `ImageData` transformations


user27182:
### Describe the feature you would like to be added.

`ImageData` transformations don't seem to be working well, e.g. #6017, and https://github.com/pyvista/pyvista/issues/6207#issuecomment-2151490805. I'm not sure why using `transform` with `ImageData` should return a `StructuredGrid`

The issue may stem from the use of `vtkTransform` to apply the transformations in `transform`. But `ImageData` can be special-cased, e.g. `translate` can simply modify the image's `origin` to apply the translation, and `transform` can simply update the `ImageData` `DirectionMatrix` with the transformation matrix.

So this fix would:
- add a `direction_matrix` property to `ImageData`
- overload or special-case `translate` for `ImageData`
- overload or special-case `transform` for `ImageData`
- `ImageData.points` may also need to be updated to apply the transformation from `direction_matrix` to the points

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

Issue #6017, opened 2024-05-02
Translate the volume doesn't work well


tkoyama010:
### Describe the bug, what's wrong, and what you expected.

See https://github.com/pyvista/pyvista/discussions/6016 . I want to deprecate using the `translate` method in `ImageData` and recommend using the `origin` property.

### Steps to reproduce the bug.

```python
import numpy as np
import pyvista as pv
from pyvista import examples


# Download a volumetric dataset
vol = examples.download_knee_full()
print(vol.center)
print(vol.origin)
vol = vol.translate(-np.array(vol.center).astype(np.float64))

cpos = [(-381.74, -46.02, 216.54), (74.8305, 89.2905, 100.0), (0.23, 0.072, 0.97)]
pl = pv.Plotter()
pl.add_volume(vol, cmap="bone", opacity="sigmoid")
pl.add_point_labels(vol.center, ['Center'], point_color='blue', point_size=20, text_color='black', font_size=20,)
pl.add_point_labels(vol.origin, ['Origin'], point_color='red', point_size=20, text_color='black', font_size=20,)
pl.camera_position = cpos
pl.show()
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Thu May 02 12:04:26 2024 SE Asia Standard Time
                OS : Windows
            CPU(s) : 12
           Machine : AMD64
      Architecture : 64bit
       Environment : IPython
        GPU Vendor : ATI Technologies Inc.
      GPU Renderer : AMD Radeon(TM) Graphics
       GPU Version : 4.5.14802 Core Profile Context 21.40.52 30.0.14052.5
  MathText Support : False
  Python 3.9.18 (main, Sep 11 2023, 14:09:26) [MSC v.1916 64 bit (AMD64)]
           pyvista : 0.43.6
               vtk : 9.3.0
             numpy : 1.26.4
        matplotlib : 3.8.2
            scooby : 0.9.2
             pooch : 1.8.1
            pillow : 9.4.0
           imageio : 2.34.1
           IPython : 8.18.1
             scipy : 1.11.4
      nest_asyncio : 1.5.8
--------------------------------------------------------------------------------
```

user27182:
Fixed by #6616. `translate` should now work

PR #6612, opened 2024-09-05
Add `direction_matrix` to `ImageData`

user27182:
### Overview

- Add `direction_matrix` property
- Use it in `points` to fix #5066
- ~Update `bounds` to support direction as well~ `GetBounds` already considers direction matrix internally 
- ~Add `index_to_physical_matrix` and `physical_to_index_matrix` to move between index and physical space~ split into #6614 instead

PR #6616, opened 2024-09-05
Return `ImageData` when transforming `ImageData`

user27182:
### Overview
Fix https://github.com/pyvista/pyvista/issues/6213, and possibly close the suggestion from #6017 too.

EDIT: Here's some background on the subject of adding orientation to `vtkImageData`. Oriented images are supported as of VTK 9.
https://discourse.vtk.org/t/proposal-to-add-orientation-to-vtkimagedata-feedback-wanted/120


user27182:
This is ready but waiting on #6612 and #6628

pyvista/core/grid.py
```diff
@@ -918,6 +918,20 @@ def direction_matrix(self):
...
+        T, R, S, K = transform.decompose()
+        if not np.allclose(K, 0):
+            # Input has shear, which is not directly supported
+            # Approximate rotation and scale by normalizing direction vectors instead
+            matrix3x3 = transform.matrix[:3, :3]
+            S = np.linalg.norm(matrix3x3, axis=1)
+            R = matrix3x3 / S[:, np.newaxis]
+        self.origin = T
+        self.direction_matrix = R
+        self.spacing = S
```
user27182, line 933: Note: There is a VTK implementation of this function in VTK main, but it's not yet part of a release:
https://github.com/Kitware/VTK/blob/cb6d87f81f9f7fcac15551e74531251aa917f3ee/Common/DataModel/vtkImageData.cxx#L2408

The VTK  implementation does not do a formal decomposition, but simply normalizes the direction matrix to get the scaling.  For the PyVista implementation, this is only done conditionally when there is shear in the transformation. The decomposition is more expensive but is more robust, e.g. some of the "direction_in == direction_out" tests fail otherwise.

user27182:
pre-commit.ci autofix

user27182:
The merge generated a bunch of typing errors since we are properly type-checking with vtk. Annotating the `output: DataSet` creates a bunch of other type errors, and same with doing `output: vtkDataSet`. Trying to resolve this nicely without adding a bunch of `type: ignore`s is what led to #6790.

user27182:
The changes in https://github.com/pyvista/pyvista/pull/6616/commits/c0ca2ed55853de879242f2ffc1f08bacdfd95e3f include a type change from `vtkDataSet` to `DataSet`. Since it has somewhat broad typing implications this single change is made into a separate PR #6788

user27182:
Is it better if I remove the breaking change and create a new filter instead, e.g. `ImageData.transform_image` ?

pyvista/core/filters/data_set.py
```diff
@@ -7193,16 +7202,41 @@ def transform(  # type: ignore[misc]
...
+            input_.point_data.active_scalars_name = active_point_scalars_name
+            input_.cell_data.active_scalars_name = active_cell_scalars_name
+
+            # Only update output if necessary
+            if input_ is not output_:
+                output_.point_data.active_scalars_name = active_point_scalars_name
+                output_.cell_data.active_scalars_name = active_cell_scalars_name
+
+        if isinstance(self, pyvista.RectilinearGrid):
+            output: ConcreteDataSetType = pyvista.StructuredGrid()  # type: ignore[assignment]
```
user27182, line 7216: This error should not be ignored. The TypeVar assumes a `T -> T` mapping and therefore expects a `RectilinearGrid`  -> `RectilinearGrid`  mapping, which is not the case. We should use the Union from #6918 instead.
```suggestion
            output: ConcreteDataSetAlias = pyvista.StructuredGrid()
```

tkoyama010 (approved):
LGTM

user27182:
Thanks so much @tkoyama010 for all your reviews! 🙏

PR #6618, opened 2024-09-05
Consolidate `inplace` logic for `DataSetFilter.transform`

user27182:
### Overview

Follow-up to #6616. These updates allow for a single `return` and a single call to `copy_from`.


pyvista/core/filters/data_set.py
```diff
@@ -7057,18 +7057,17 @@ def transform(
...
-            self.copy_from(res, deep=False)
-            return self
-
+        output = (
+            pyvista.StructuredGrid()
+            if isinstance(self, pyvista.Grid)
+            else self
+            if inplace
+            else self.__class__()
+        )
```
tkoyama010, line 7066:
```suggestion
        self_output = (
            self
            if inplace
            else self.__class__()
        )
        output = (
            pyvista.StructuredGrid()
            if isinstance(self, pyvista.Grid)
            else self_output
        )
```

pyvista/core/filters/data_set.py
```diff
@@ -7057,18 +7057,17 @@ def transform(
...
         # The output from the transform filter contains a shallow copy
         # of the original dataset except for the point arrays.  Here
         # we perform a copy so the two are completely unlinked.
-        if isinstance(self, pyvista.Grid):
-            output: _vtk.vtkDataSet = pyvista.StructuredGrid()
-        else:
-            output = self.__class__()
-        output.copy_from(res, deep=True)
+        output.copy_from(res, deep=not inplace)
```
tkoyama010, line 7070:
```suggestion
        if inplace:
            output.copy_from(res)
        else:
            output.copy_from(res, deep=True)
```

tkoyama010 (changes requested):
+100 unifying the return part. However, the parts of the code that have to expand to read in one's mind are burdensome to the code reader. I suggest use the code that has already expanded.

user27182:
pre-commit.ci autofix

tkoyama010 (dismissed):
LGTM

pyvista/core/filters/data_set.py
```diff
@@ -7057,18 +7057,15 @@ def transform(
...
-
+        self_output = self if inplace else self.__class__()
+        output = pyvista.StructuredGrid() if isinstance(self, pyvista.Grid) else self_output
         # The output from the transform filter contains a shallow copy
         # of the original dataset except for the point arrays.  Here
         # we perform a copy so the two are completely unlinked.
-        if isinstance(self, pyvista.Grid):
-            output: _vtk.vtkDataSet = pyvista.StructuredGrid()
+        if inplace:
+            output.copy_from(res)
```
user27182, line 7066:
```suggestion
            output.copy_from(res, deep=False)
```

tkoyama010 (approved):

pyvista/core/filters/data_set.py
```diff
@@ -7057,18 +7057,15 @@ def transform(
...
         # of the original dataset except for the point arrays.  Here
         # we perform a copy so the two are completely unlinked.
-        if isinstance(self, pyvista.Grid):
-            output: _vtk.vtkDataSet = pyvista.StructuredGrid()
+        if inplace:
+            output.copy_from(res, deep=False)
         else:
-            output = self.__class__()
-        output.copy_from(res, deep=True)
+            output.copy_from(res, deep=True)
```
tkoyama010, line 7068: Just to confirm, are the behavior will change if `inplace` when self is `pyvista.Grid`?

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
user27182: No the logic should be the same: shallow copy if inplace, deep copy otherwise. And output SteucturedGrid if input is a Grid (ImageData or RectilinearGrid)

user27182: inplace must always be false for Grid inputs, there is logic on the first line of this method to raise an error for these types if inplace is True

tkoyama010: Ah. That makes sense. Thanks!

tkoyama010 (approved):

PR #6627, opened 2024-09-06
Fix copy structure with self returning empty mesh

user27182:
### Overview
Copying `ImageData` structure from itself returns an empty mesh:
``` python
import pyvista as pv
image = pv.ImageData(dimensions=(10, 10, 5))
image.copy_structure(image)
print(image.dimensions)  # (0, 0, 0)
```
This PR fixes this. Discovered as part of #6616

### DIFF
--- a/pyvista/core/dataobject.py
+++ b/pyvista/core/dataobject.py
@@ -676,7 +676,7 @@ class DataObject:
         >>> target.plot(show_edges=True)
 
         """
-        self.CopyStructure(dataset)
+        self.CopyStructure(dataset) if dataset is not self else None
 
     def copy_attributes(self, dataset: _vtk.vtkDataSet) -> None:
         """Copy the data attributes of the input dataset object.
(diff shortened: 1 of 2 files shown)


pyvista/core/dataobject.py
```diff
@@ -676,7 +676,7 @@ def copy_structure(self, dataset: _vtk.vtkDataSet) -> None:
         >>> target.plot(show_edges=True)
 
         """
-        self.CopyStructure(dataset)
+        self.CopyStructure(dataset) if dataset is not self else None
```
tkoyama010, line 679:
```suggestion
        self.CopyStructure(dataset) if dataset is not self
```

tkoyama010 (changes requested):

pyvista/core/dataobject.py
user27182: I would also prefer this but it's not valid syntax, `else` is always required unfortunately

tkoyama010: Too bad Python is a bit redundant.

tkoyama010 (approved):
LGTM

PR #6628, opened 2024-09-06
Add `Transform.decompose()` to extract translation, rotation, reflection, scale, and shear components

user27182:
### Overview

Breakdown a transform into its parts. This is used by #6616 to transform `ImageData`.

Also update API to enable transform re-composition:
- Allow setting `multiply_mode` with init
- Allow sequence of transform-like inputs


user27182:
I'm going to try and decompose this further, e.g. reflections and shear

user27182:
> I'm going to try and decompose this further, e.g. reflections and shear

Turns out that shear can be represented by a rotation and scaling. This is the case when using polar decomposition. The implementation has been updated accordingly.  And, reflections can be lumped either into the scale factors (i.e. negative scale factors) or by multiplying one of the rotation rows by -1. For this implementation, scale factors are always positive, and reflections (if any) are lumped into the rotation.

Some useful resources I found on polar decomposition relevant to this PR:

https://www.cs.cornell.edu/courses/cs4620/2014fa/lectures/polarnotes.pdf
http://buzzard.ups.edu/courses/2014spring/420projects/math420-UPS-spring-2014-buffington-polar-decomposition.pdf
http://3map.snu.ac.kr/courses/engmath_materials/lecture_notes_new/LN2015-Polar.pdf

user27182:
> > I'm going to try and decompose this further, e.g. reflections and shear
> 
> Turns out that shear can be represented by a rotation and scaling. This is the case when using polar decomposition.

I abandoned the use of polar decomposition. When shear was added, the scaling term was represented as a 3x3 matrix instead of a vector, and I could not separate it or lump it into the rotation.
Instead, I found a 3rd party implementation to decompose into `TRSK` which includes shear explicitly and seems to work exactly as I want/expect.

pyvista/core/utilities/transformations.py
```diff
@@ -392,3 +394,180 @@ def rotation(
...
+        return T4, R4, S4, K4
+
+    return T, R, S, K
+
+
+def _decompose(matrix, allow_negative_scale=False):
+    """Decompose a matrix into its parts.
+
+    Copied from:
+    https://github.com/matthew-brett/transforms3d/blob/6a43a98e3659d198ff6ce2c90d52ddef50fcf770/transforms3d/affines.py#L156
```
user27182, line 514: Not sure if this is allowed or if it poses a copyright/licensing issue. Repo uses BSD-2 https://github.com/matthew-brett/transforms3d?tab=License-1-ov-file

Probably best not to add `transforms3d` as another dependency either...

tkoyama010: Could you explain why it is better not to add `transforms3d` as a new dependency?

tkoyama010 (changes requested):

pyvista/core/utilities/transformations.py
user27182: I _could_ try to use polar decomposition again from https://github.com/pyvista/pyvista/pull/6628/commits/1b66c0ae0ca2ba42ef6e24fddfc33443ca5c5412 with this code:
``` python

def _polar_decomposition(a):
      # Decompose `a=up` where u is orthonormal and p is positive semi-definite
      # See scipy.linalg.polar for details
      w, s, vh = np.linalg.svd(a, full_matrices=False)
      u = w.dot(vh)
      p = (vh.T.conj() * s).dot(vh)
      return u, p
```
Where `u` is the rotation `R`, and `p` is scaling and shear `SK`. I had problems separating `SK` earlier but they may be separable. The problem is that this is also copying code, but from `scipy` library instead of `transforms3d`. And, polar decomp uses svd, which I think makes it more computationally expensive than the other method (though I haven't tested this).

user27182: > Could you explain why it is better not to add `transforms3d` as a new dependency?

One reason is just based on avoiding "yet another dependency". We are only using a single function that only depends on numpy, so why add the entire package?

Another reason may be testing/maintenance, their [home page](https://github.com/matthew-brett/transforms3d) states: "We use Github actions to test the code automatically under Pythons 3.7 through 3.10." So it's not officially tested for later python. But, the maintainer has pushed recent commits, and we have code coverage for the key function with our testing, so maybe it's OK.

user27182: But, if we do keep it as a dependency, the package does have other transformation methods that could be useful to wrap in a PyVista friendly way. E.g. some of the shear methods: http://matthew-brett.github.io/transforms3d/reference/transforms3d.shears.html#transforms3d.shears.sadn2aff

E.g. similar to `Transform.rotate_vector`, we could add `Transform.shear_plane` method or something like that. I don't really use shear so not familiar with its use cases but maybe that is good to have.

tkoyama010: I see. How about installing it optionally, for example `pip install pyvista[transford]`. That way, there is no default Python version support or impact on load times.

user27182: > I see. How about installing it optionally, for example `pip install pyvista[transform]`?

This is a good suggestion, however this PR is supporting #6616. And so the decomposition is needed to transform `ImageData` here:
https://github.com/pyvista/pyvista/blob/f921caf110f9d633e8784723f0ebfa90bf78e131/pyvista/core/grid.py#L921-L933

As noted in https://github.com/pyvista/pyvista/pull/6616#discussion_r1755860304, however, we don't _need_ to do a formal decomposition. We could simply estimate scale and rotation by normalizing the 3x3 part of the transformation.

It's probably best to only support one or the other, i.e. either always do formal decomposition (with `transforms3d`) as a dependency), or always do informal decomp and simply estimate it. Otherwise, different transformations would be applied based on wether the optional package is installed or not.

tkoyama010: It is better to always use the same calculation with and without options. It would be better to output an error message if this calculation is called, and transforms3d is not installed. Please see options such as jupyter.

pyvista/core/utilities/transformations.py
```diff
@@ -392,3 +394,176 @@ def rotation(
         translate_from_origin[:3, 3] = valid_point
 
         return translate_from_origin @ rotate @ translate_to_origin
+
+
+def decompose(
```
user27182, line 399: Could rename this to `decomposition` to make it more consistent with other names in this module

pyvista/core/utilities/transformations.py
user27182: After further testing I found that this third party decomposition with `transfoms3d` package did not correctly recover some values, e.g. using a matrix with scaling (1,2,3) returned (2,1,3) in cases with a rotation (i.e. the decomposed scale was dependent on the rotation).

I have reverted to using polar decomposition, initially used earlier in https://github.com/pyvista/pyvista/pull/6628/commits/1b66c0ae0ca2ba42ef6e24fddfc33443ca5c5412.

As such, additional dependencies are no longer required.

user27182:
The documentation still needs to be reviewed but otherwise I am happy with this PR's current implementation. It has also been tested as part of #6213 and seems to be working well for transforming ImageData.

pyvista/core/utilities/transformations.py
```diff
@@ -392,3 +394,176 @@ def rotation(
...
+        Array or transform to decompose.
+
+    homogeneous : bool, default: False
+        If ``True``, return the components (translation, rotation, etc.) as
+        4x4 matrices. By default, translation and scaling components are returned as
+        length-3 vectors, and rotation and shear components are returned as 3x3 matrices.
+
+    allow_negative_scale : bool, default: False
+        If ``True``, the first scaling term may be negative. By default, the
+        scaling factors are always positive.
```
user27182, line 449: This is kind of left over from the previous ``transforms3d`` implementation. I think it might be an unnecessary complication. We can probably remove it, such that scaling factors returned by this decomposition are _always_ positive, and reflections (if any) are _always_ included with the rotation. This is also how the current polar decomposition works naturally by default, so this would also remove a processing step.

pyvista/core/utilities/transformations.py
user27182: Renamed in https://github.com/pyvista/pyvista/pull/6628/commits/fbba4ebe240dbb99fef10577d226ac1a94a5c7ef

pyvista/core/utilities/transformations.py
user27182: The main thing this option does is multiply the reflection with either the rotation term (`allow_negative_scale=False`) or the scaling term (`allow_negative_scale=True`).

But, the naming is awkward, and makes it harder to check for reflections. Instead, I think it's best to simply decompose the reflection term explicitly. This was done in https://github.com/pyvista/pyvista/pull/6628/commits/ded9567ee37196a0e2a4f896b33d1e6d7baf3cfa.

So the decomposition is now `TRNSK` instead of `TRSK` with `N` as the reflection.

user27182:
Docs for `Transfom.decompose` look good. The docs for `decomposition` are not published, but none of the `transformations` module documentation is (I think they are more low-level functions, intended mainly for internal use, maybe?)

tkoyama010 (dismissed):
LGTM

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

PR #6622, opened 2024-09-05
Add `pickle` support for `MultiBlock` and use official VTK format

user27182:
### Overview

- Use the official pickle format supported by VTK:
https://github.com/Kitware/VTK/blob/master/Wrapping/Python/vtkmodules/util/pickle_support.py
    - This utility is new to VTK 9.3
- Make this format the new default for `pyvista.PICKLE_FORMAT`
- Fix #6419
- Support serializing object with `user_dict` set

PR #6631, opened 2024-09-07
Move `pickle` tests from `test_dataset` to `test_dataobject`

user27182:
### Overview

These tests are in the wrong place. Pickling applies to all data objects, not just datasets. I couldn't find these tests initially when working through #6622 partly because of this. This PR also adds "pickle" to the tests names, since that was also contributing to me not being able to find them.


tkoyama010 (approved):
LGTM

PR #6632, opened 2024-09-08
Use `pydocstyle` rules with `ruff`

user27182:
### Overview

See rules: https://docs.astral.sh/ruff/rules/#pydocstyle-d

### DIFF
--- a/examples/01-filter/distance-between-surfaces.py
+++ b/examples/01-filter/distance-between-surfaces.py
@@ -1,5 +1,4 @@
-"""
-Distance Between Two Surfaces
+"""Distance Between Two Surfaces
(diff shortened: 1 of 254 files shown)


examples/01-filter/distance-between-surfaces.py
```diff
@@ -1,5 +1,4 @@
-"""
-Distance Between Two Surfaces
+"""Distance Between Two Surfaces
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
user27182, line 2: This change is not great. Only seems to apply to `examples`

user27182:
pre-commit.ci autofix

banesullivan (approved):

PR #6639, opened 2024-09-10
Compare sample and interpolate in example

MatthewFlamm:
### Overview

Stemming from a discussion on slack, there is ongoing confusion about when to use `interpolate` and when to use `sample`.  (added because it was part of the original discussion: there used to be `probe` but this has been removed.)

I think most of the confusion arises due to naming, i.e. `interpolate` is generic. However, `interpolate` semi-follows vtk naming, `vtkPointInterpolator`.  We could consider renaming it to `interpolate_points`, but this is a breaking change and this filter is likely to be heavily utilized.

This PR adds in an example that directly compares and contrasts the two approaches.  Including one contrived example to show when ``interpolate`` fails and when ``sample`` succeeds.  And vice versa.

I think this PR would still be useful even if a rename is decided upone.

## Details
Will add in doc image cache for testing after the CI runs.


examples/01-filter/resample.py
```diff
@@ -4,22 +4,9 @@
 Resampling
 ~~~~~~~~~~
```
MatthewFlamm, line 5:
```suggestion
Detailed Resampling
~~~~~~~~~~~~~~~~~~~
```

MatthewFlamm:
@banesullivan cc since we discussed this on slack.

tkoyama010 (approved):
LGTM

examples/01-filter/interpolate_sample.py
```diff
@@ -0,0 +1,199 @@
...
+# of the grid.
+#
+# Make subset (0.7, 0.7, 0) units in dimension and then rotate by 45 degrees around
+# its center.
+subset = pv.ImageData(dimensions=(8, 8, 1), spacing=[0.1, 0.1, 0], origin=(0.15, 0.15, 0))
+rotated_subset = subset.rotate_vector(vector=(0, 0, 1), angle=45, point=(0.5, 0.5, 0))
+output = rotated_subset.sample(grid)
+output
+
+# The data in the sampled region looks identical to the original grid
```
MatthewFlamm, line 172:
```suggestion
# %%
# The data in the sampled region looks identical to the original grid
```

banesullivan (dismissed):
I made a cursory read through the examples on the documentation preview and this looks excellent! Great work adding this detailed comparison - going to be frequently referenced!

tkoyama010 (approved):

MatthewFlamm:
@tkoyama010 thanks again for cleaning up my mistakes in the doc image testing.

PR #6646, opened 2024-09-12
Catch runtime errors in doc image testing

user27182:
### Overview

Catch runtime errors when comparing images so that the failed test images artifact is populated. This should make resolving this error more easily:
https://github.com/pyvista/pyvista/actions/runs/10815577469/job/30004788411#step:14:207

### DIFF
--- a/tests/doc/tst_doc_images.py
+++ b/tests/doc/tst_doc_images.py
@@ -185,37 +185,35 @@ def _test_both_images_exist(filename, docs_image_path, cached_image_path):
 
 
 def _test_compare_images(test_name, docs_image_path, cached_image_path):
-    docs_image = pv.read(docs_image_path)
-    cached_image = pv.read(cached_image_path)
-
-    # Check if test should fail or warn
     try:
+        docs_image = pv.read(docs_image_path)
+        cached_image = pv.read(cached_image_path)
+
+        # Check if test should fail or warn
         error = pv.compare_images(docs_image, cached_image)
-    except RuntimeError as e:
-        fail_msg = repr(e)
-        warn_msg = None
-    else:
         fail_msg = _check_compare_fail(test_name, error)
         warn_msg = _check_compare_warn(test_name, error)
-
-    if fail_msg:
-        # Check if test case is flaky test
-        if test_name in FLAKY_TEST_CASES:
-            # Compare build image to other known valid versions
-            success_path = _is_false_positive(test_name, docs_image)
-            if success_path:
-                # Convert failure into a warning
-                warn_msg = fail_msg + (
-                    '\nTHIS IS A FLAKY TEST. It initially failed (as above) but passed when '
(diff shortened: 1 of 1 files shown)


tests/doc/tst_doc_images.py
```diff
@@ -145,7 +145,11 @@ def test_docs(test_case: _TestCaseTuple):
         _save_failed_test_image(fail_source)
         pytest.fail(fail_msg)
 
-    warn_msg, fail_msg = _test_compare_images(*test_case)
+    try:
+        warn_msg, fail_msg = _test_compare_images(*test_case)
```
MatthewFlamm, line 149: To me, it would better to have the try block within `_test_compare_images`, but it is a matter of taste.

tests/doc/tst_doc_images.py
```diff
@@ -145,7 +145,11 @@ def test_docs(test_case: _TestCaseTuple):
         _save_failed_test_image(fail_source)
         pytest.fail(fail_msg)
```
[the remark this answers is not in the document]
user27182: Sure, we can put the whole thing in a try..except block to be safe.

tests/doc/tst_doc_images.py
user27182: Ok, I also think is cleaner

tests/doc/tst_doc_images.py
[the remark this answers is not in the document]
user27182: This suggestion is kind of incompatible with this one https://github.com/pyvista/pyvista/pull/6646#discussion_r1756970039

We _could_ add a second try block inside of `_test_both_images_exist`, but the code there is just checking `None` and assigning variables... not really much to try..except for that test

tests/doc/tst_doc_images.py
```diff
@@ -185,31 +185,35 @@ def _test_both_images_exist(filename, docs_image_path, cached_image_path):
...
-                fail_msg += (
-                    '\nTHIS IS A FLAKY TEST. It initially failed (as above) and failed again for '
-                    f'all images in \n\t{Path(FLAKY_IMAGE_DIR, test_name)!s}.'
-                )
+    try:
+        docs_image = pv.read(docs_image_path)
+        cached_image = pv.read(cached_image_path)
+
+        # Check if test should fail or warn
+        error = pv.compare_images(docs_image, cached_image)
```
MatthewFlamm, line 193: Assuming we don't want to necessarily catch this error in other locations, it would be more precise to only wrap the call to `pv.compare_images` in the `try` block, keep the `except` as is and then put the rest of this in an `else:` block.

user27182: There is another call to `compare_images`  later on in `_is_false_positive`.  So, we need to wrap both calls somehow

MatthewFlamm (approved):

tkoyama010 (approved):

PR #6651, opened 2024-09-15
Fix self-reference in docs

GuillaumeBroggi:
`pyvista.voxelize` is self-referencing itself in `See Also` section.


user27182 (approved):
LGTM, thanks!

PR #6656, opened 2024-09-16
Fix `cast_to_rectilinear_grid` floating-point error

GuillaumeBroggi:
The current implementation can introduce a floating-point error when the `spacing` is a float.

For instance, the following would fail:
```python
import pyvista as pv

grid = pv.ImageData(
    dimensions=(10, 10, 10),
    spacing=(27.88888888888889, 28.11111111111111, 28.22222222222222),
    origin=(-126.0, -127.0, -127.0),
)
rectilinear = grid.cast_to_rectilinear_grid()
assert rectilinear.bounds == grid.bounds
```

The proposed PR uses `np.linspace` to generate the rectilinear coordinates so floating-point errors do not accumulate.

### DIFF
--- a/tests/core/test_grid.py
+++ b/tests/core/test_grid.py
@@ -938,6 +938,7 @@ def test_cast_uniform_to_rectilinear():
 
 
 def test_cast_image_data_with_float_spacing_to_rectilinear():
+    # https://github.com/pyvista/pyvista/pull/6656
     grid = pv.ImageData(
         dimensions=(10, 10, 10),
         spacing=(27.88888888888889, 28.11111111111111, 28.22222222222222),
--- a/tests/core/test_grid.py
+++ b/tests/core/test_grid.py
@@ -937,6 +937,18 @@ def test_cast_uniform_to_rectilinear():
     assert rectilinear.bounds == grid.bounds
 
 
+def test_cast_image_data_with_float_spacing_to_rectilinear():
+    grid = pv.ImageData(
+        dimensions=(10, 10, 10),
+        spacing=(27.88888888888889, 28.11111111111111, 28.22222222222222),
+        origin=(-126.0, -127.0, -127.0),
+    )
+    rectilinear = grid.cast_to_rectilinear_grid()
+    assert rectilinear.n_points == grid.n_points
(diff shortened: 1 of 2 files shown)


tests/core/test_grid.py
```diff
@@ -937,6 +937,18 @@ def test_cast_uniform_to_rectilinear():
     assert rectilinear.bounds == grid.bounds
 
 
+def test_cast_image_data_with_float_spacing_to_rectilinear():
+    grid = pv.ImageData(
```
tkoyama010, line 941:
```suggestion
    # https://github.com/pyvista/pyvista/pull/6656
    grid = pv.ImageData(
```

tkoyama010 (approved):
LGTM

GuillaumeBroggi:
As a side note, the proposed PR does not degrade performances.

```python
%%timeit
grid = pv.ImageData(
    dimensions=(10000, 10000, 10000),
    spacing=(27.88888888888889, 28.11111111111111, 28.22222222222222),
    origin=(-126.0, -127.0, -127.0),
)
rectilinear = grid.cast_to_rectilinear_grid()
```
Current implementation:
> `446 μs ± 10.5 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)`
Proposed PR:
> `304 μs ± 2.37 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)`

tkoyama010:
Glad to hear it :)

MatthewFlamm (approved):
Thanks!  This would be better without the unrelated formatting changes however.  If the formatting needs to be changed in this file it would be better in a separate PR.

GuillaumeBroggi:
@MatthewFlamm Good point; I did not notice the formatting changed; that's a mistake. I'll correct it. Thanks!

user27182:
Looks like the use of single quotes for the formatting may be preferred for the project. See #6657.

tkoyama010 (changes requested):
OK. Let's wait for it.

GuillaumeBroggi:
It should be fixed now. I had unintentionally applied black formatter.

tkoyama010 (approved):

MatthewFlamm (approved):

PR #2180, opened 2022-02-10
Style/black and resolve many flake8 issues

banesullivan:
For #2178 Uses Black to format code and flake8 to lint. See `.flake8` for which patterns we are not checking

PR #6657, opened 2024-09-17
Use single quotes formatting with `ruff`

user27182:
### Overview

Enforce the use of single quotes instead of double quotes. I don't have any opinion on which is better, but I think having a consistent format is better than an inconsistent one.

EDIT: Use single quotes instead of double quotes. Double was originally suggested since that's what's [preferred by Black](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings).


banesullivan:
I was under the impression that we converged on single quotes.

There is some history here... see https://github.com/pyvista/pyvista/pull/2180#issuecomment-1034383251 and https://github.com/pyvista/pyvista/discussions/2178#discussioncomment-2145873 but I can't dig up where I thought we had converged on single quotes (might have been on Slack)

user27182:
> I was under the impression that we converged on single quotes.
> 
> There is some history here... see [#2180 (comment)](https://github.com/pyvista/pyvista/pull/2180#issuecomment-1034383251) and [#2178 (comment)](https://github.com/pyvista/pyvista/discussions/2178#discussioncomment-2145873) but I can't dig up where I thought we had converged on single quotes (might have been on Slack)

Indeed the diff is much smaller when using `quote-style = "single"`, I'll change it to single

banesullivan:
I'm +1 for single quotes. I think that diff shows there has been at lease some level of historical preference for single quotes. 

As long as the formatter will automatically fix these for me, I'm indifferent

user27182:
> I'm +1 for single quotes. I think that diff shows there has been at lease some level of historical preference for single quotes.
> 
> As long as the formatter will automatically fix these for me, I'm indifferent

Apparently there are a lot of passionate discussions and opinions about this. Singles is fine with me, but I am also indifferent, mostly care about consistency and avoiding introducing formatting changes unnecessarily. Use whichever you prefer and the formatter will fix it.

.github/workflows/style_docstr.yml
```diff
@@ -66,5 +66,6 @@ jobs:
           filter_mode: nofilter
           fail_on_error: true
           version: 2.29.5
+          reviewdog_url: https://github.com/reviewdog/reviewdog/releases/download/v0.17.5/reviewdog_0.17.5_Linux_x86_64.tar.gz
```
MatthewFlamm, line 69: This change is unrelated to this PR?  Need to wait for #6658?

MatthewFlamm:
I have no preference on which to use as long as we don't have to spend time in PR reviews dealing with it.  It should be handled by the formatter which is what is done in this PR, so I'm +1.

After this is merged, the commit ref should be added to `.git-blame-ignore-revs` file.

user27182:
> On a similar note, when my `black` breaks a line to enforce the length limit, `ruff` (I suppose) does not necessarily reverts it when running the `pre-commit` hooks.

This could be due to the use of a [trailing comma]( 
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#trailing-commas) by black, which are presumably respected by `ruff` (but which can be manually removed to collapse a line that was previously expanded)

tkoyama010 (approved):
LGTM

banesullivan (approved):

PR #6658, opened 2024-09-17
Use updated version of `reviewdog` with `Vale` config

user27182:
### Overview

Use `reviewdog_url` in the Vale workflow to force using an updated version of `reviewdog` (`17.5` instead of `17.0`). 

Without this, we get a `reviewdog` error like this one from #6657 for large diffs https://github.com/pyvista/pyvista/actions/runs/10907071242/job/30269840434?pr=6657#step:3:40


tkoyama010 (approved):
LGTM

Issue #6649, opened 2024-09-13
Add a `connectivity` filter based on shared faces


GuillaumeBroggi:
As detailed in the [pyvista.datasetfilters.connectivity documentation](https://docs.pyvista.org/api/core/_autosummary/pyvista.datasetfilters.connectivity):
> In general, cells are considered to be connected if they share a point.

Some use cases require cells to be connected if and only if they share faces. I am not sure how such a filter could be implemented.

user27182:
I believe that limitation exists because the `connectivity` filter is general-purpose. If we look at connectivity for structured meshes like ImageData, 6-connectivity can be computed with this filter:
https://vtk.org/doc/nightly/html/classvtkImageConnectivityFilter.html

But, like most ImageData filters, this operates on point data, not cell data, and so still isn't quite what you may be after. But, if you have a cell-based input, you could use `cells_to_points` first to make it compatible with the point-based filter, use the filter, then do `points_to_cells` to return a cell-based representation. I believe this would generate the expected result of connected faces.

GuillaumeBroggi:
Thanks for your useful tips! See #6659 for PR that implements such a connectivity filter.

GuillaumeBroggi:
Close as #6659 was merged.

PR #6659, opened 2024-09-17
Add `label_connectivity` method to `ImageDataFilters`

GuillaumeBroggi:
# Overview

PyVista can be used as a powerful tool to manipulate meshes, for instance, for numerical modeling in mechanics, computational fluid dynamics, etc. Sometimes, a numerical model requires a mesh continuity. 

[pyvista.datasetfilters.connectivity](https://docs.pyvista.org/api/core/_autosummary/pyvista.datasetfilters.connectivity) can be used to filter out unconnected regions but leverages a 1-connectivity, i.e., any two cells that share a common point are considered as connected. This criterion is not strict enough for some applications where the connectivity requires shared faces.

# Details
- As pointed out by @user27182 in #6649, [classvtkImageConnectivityFilter](https://vtk.org/doc/nightly/html/classvtkImageConnectivityFilter.html) can be leveraged to implement this feature.
- This PR wraps `classvtkImageConnectivityFilter` and proposes it as an `ImageDataFilters.`
- The implementation is not restricted to face connectivity. Instead, it exposes the complete functionality of `classvtkImageConnectivityFilter`. Face connectivity is only one of the possible applications.

# Examples

```python
segmented_grid = pv.ImageData(dimensions=(4, 3, 3))
segmented_grid.cell_data["Data"] = [0, 0, 0, 1, 0, 1, 1, 2, 0, 0, 0, 0]

connected = segmented_grid.label_connectivity()

pl = pv.Plotter()
_ = pl.add_mesh(connected.threshold(0.5), show_edges=True)
_ = pl.add_mesh(connected.threshold(0.5, invert=True), show_edges=True, opacity=0.5)
pl.show(jupyter_backend="notebook")
```
image


pyvista/core/_vtk_core.py
```diff
@@ -432,6 +432,7 @@
 from vtkmodules.vtkIOXML import vtkXMLWriter
 
 with contextlib.suppress(ImportError):
+    from vtkmodules.vtkImagingMorphological import vtkImageConnectivityFilter
```
user27182, line 435: Suppression is only needed if the import fails for older supported vtk versions (e.g. fails CI testing), but this filter was introduced before VTK 9.
```suggestion
from vtkmodules.vtkImagingMorphological import vtkImageConnectivityFilter

with contextlib.suppress(ImportError):
```

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1606,287 @@ def _update_and_get_output():
         # Restore active scalars
         self.set_active_scalars(scalars, preference='point')  # type: ignore[attr-defined]
         return output
+
+    def label_connectivity(
+        self,
+        *,
+        scalars: str | None = None,
+        scalar_range: str | VectorLike[int] | VectorLike[float] | None = 'auto',
```
user27182, line 1614: With type hints, `float` also implies `int`, and so specifying both is redundant. See 
https://peps.python.org/pep-0484/#the-numeric-tower

Also suggest using Literal if `auto` is the only acceptable string
```suggestion
        scalar_range: Literal['auto'] | VectorLike[float] | None = 'auto',
```

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1606,287 @@ def _update_and_get_output():
...
+        if scalars is None:
+            set_default_active_scalars(input_mesh)
+        else:
+            input_mesh.set_active_scalars(scalars)
+        # Make sure we have point data (required by the filter)
+        field, scalars = input_mesh.active_scalars_info
+
+        if field == FieldAssociation.CELL:
+            # Convert to point data
+            input_mesh = input_mesh.cells_to_points(scalars=scalars)
```
user27182, line 1786: Minor optimization, a copy was already made earlier
```suggestion
            input_mesh = input_mesh.cells_to_points(scalars=scalars, copy=False)
```

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1606,287 @@ def _update_and_get_output():
...
+                raise TypeError('Scalar range must be a numpy array, a sequence, "auto" or None.')
+
+            if len(scalar_range) != 2:
+                raise ValueError('Scalar range must have two elements defining the min and max.')
+            if scalar_range[0] > scalar_range[1]:  # type: ignore[operator]
+                raise ValueError(
+                    f'Lower value of scalar range {scalar_range[0]}'
+                    f' cannot be greater than the upper value {scalar_range[0]}'
+                )
+            alg.SetScalarRange(*scalar_range)
```
user27182, line 1819: Check for string first since equality checks for numpy arrays raises ambiguous logic error.
Also suggest using validation methods for scalar range. (May also need to add `from pyvista.core import _validation` to file)
```suggestion
            if isinstance(scalar_range, str) and scalar_range == 'auto':
                unique_scalars = np.unique(input_mesh.point_data[scalars])
                scalar_range = [unique_scalars[1], unique_scalars[-1]]
            else:
                scalar_range = _validation.validate_data_range(scalar_range)
            alg.SetScalarRange(*scalar_range)
```

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1606,287 @@ def _update_and_get_output():
...
+
+        if extraction_mode == 'all':
+            alg.SetExtractionModeToAllRegions()
+        elif extraction_mode == 'largest':
+            alg.SetExtractionModeToLargestRegion()
+        elif extraction_mode == 'point_seeds':
```
user27182, line 1825: `point_seed` without the `s` is the name used by the `connectivity` filter, perhaps using the same spelling would be better? Or is its usage by this filter substantially different? (I think xyz points may be used by this filter, whereas point ids are used by `connectivity`)

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1606,287 @@ def _update_and_get_output():
...
+            if point_seeds is None:
+                raise ValueError(
+                    '`point_seeds` must be specified when `extraction_mode="point_seeds"`.',
+                )
+            elif isinstance(point_seeds, (Sequence, np.ndarray)):
+                # Ensure points are floats, see pyvista.core.utilities.points
+                point_seeds = np.array(point_seeds, dtype=float)
```
user27182, line 1832: Small optimization to use `asarray` to avoid making a copy if input is already an array
```suggestion
                point_seeds = np.asarray(point_seeds, dtype=float)
```

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1606,287 @@ def _update_and_get_output():
...
+                # PointSet requires vtk >= 9.1.0
+                # See https://docs.pyvista.org/api/core/_autosummary/pyvista.pointset#pyvista.PointSet
+                if pyvista.vtk_version_info >= (9, 1, 0):
+                    point_seeds = pyvista.PointSet(point_seeds)
+                else:
+                    point_seeds = pyvista.PolyData(point_seeds)
+            elif not isinstance(point_seeds, _vtk.vtkDataSet):
+                raise ValueError(
```
user27182, line 1840:
```suggestion
                raise TypeError(
```

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1606,287 @@ def _update_and_get_output():
...
+                f'Invalid `label_mode` "{label_mode}", use "size", "constant", or "seeds".'
+            )
+
+        _update_alg(alg, progress_bar, 'Identifying and Labelling Connected Regions')
+
+        output = _get_output(alg)
+
+        if field == FieldAssociation.CELL:
+            # Convert back to cell data
+            output = output.points_to_cells()
```
user27182, line 1879:
```suggestion
            output = output.points_to_cells(copy=False)
```

tests/core/test_imagedata_filters.py
```diff
@@ -468,3 +468,80 @@ def test_pad_image_raises(single_point_image, uniform, logo):
...
+
+def test_label_connectivity_inplace(segmented_grid):
+    connected = segmented_grid.label_connectivity(inplace=True)
+    assert connected == segmented_grid
+    assert 'RegionId' in connected.cell_data
+    assert 'Data' in connected.cell_data
+
+
+def test_label_connectivity_invalid_parameters(segmented_grid):
+    with pytest.raises(ValueError):  # noqa: PT011
```
user27182, line 538: Nit: I think it's better to include explicit matches for the expected error message using `match=` instead of ignoring with `# noqa: PT011`. This is a good sanity check for cases with complex validation logic because sometimes the error is emitted from an unexpected branch (with a different error message emitted than may be expected).

```suggestion
    with pytest.raises(ValueError, match='Invalid extraction mode'): 
```

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1606,287 @@ def _update_and_get_output():
...
         self.set_active_scalars(scalars, preference='point')  # type: ignore[attr-defined]
         return output
+
+    def label_connectivity(
+        self,
+        *,
+        scalars: str | None = None,
+        scalar_range: str | VectorLike[int] | VectorLike[float] | None = 'auto',
+        extraction_mode: Literal['all', 'largest', 'point_seeds'] = 'all',
+        point_seeds: MatrixLike[float] | Sequence[float] | _vtk.vtkDataSet | None = None,
```
user27182, line 1616: Is `Sequence` meant to capture the single-vector case?
```suggestion
        point_seeds: MatrixLike[float] | VectorLike[float] | _vtk.vtkDataSet | None = None,
```

user27182 (commented):
Thanks for implementing this! I have a few comments/suggestions, some are more subjective, feel free to push back on any of these

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1607,281 @@ def _update_and_get_output():
...
+            field.value,
+            scalars,
+        )  # args: (idx, port, connection, field, name)
+
+        # Set the scalar range considered for connectivity
+        if scalar_range is None:
+            # Default to vtk default 0.5 to VTK_DOUBLE_MAX, nothing to do
+            # See https://vtk.org/doc/nightly/html/classvtkImageConnectivityFilter.html
+            pass
+        else:
```
tkoyama010, line 1805:
```suggestion
        # Default to vtk default 0.5 to VTK_DOUBLE_MAX, nothing to do
        # See https://vtk.org/doc/nightly/html/classvtkImageConnectivityFilter.html
        if scalar_range is not None:
```

tkoyama010 (changes requested):

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1607,281 @@ def _update_and_get_output():
...
+
+        .. versionadded:: 0.44.2
+
+        Notes
+        -----
+        This filter implements `vtkImageConnectivityFilter
+        <https://vtk.org/doc/nightly/html/classvtkImageConnectivityFilter.html>`_.
+
+        Using a dissimilar type for `scalars` values and `scalar_range` might produce
+        unexpected results.
```
GuillaumeBroggi, line 1643: @user27182 @tkoyama010  PyVista could issue a warning when the types are dissimilar. Would it be suitable? There are very few warnings in PyVista code at the moment.

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1607,280 @@ def _update_and_get_output():
...
+        connectivity. A 4-connectivity is used for 2D images or a 6-connectivity for 3D
+        images. This filter operates on point-based data. If cell=based data are provided,
+        they are re-meshed to a point-based representation using
+        :func:`~pyvista.ImageDataFilters.cells_to_points` and the output is meshed back
+        to a cell-based representation with :func:`~pyvista.ImageDataFilters.points_to_cells`,
+        effectively filtering based on face connectivity. The connected regions are
+        extracted and labelled according to the strategy defined by `extraction_mode`
+        and `label_mode`, respectively. Unconnected regions are labelled with `0` value.
+
+        .. versionadded:: 0.44.2
```
GuillaumeBroggi, line 1635: Version to edit if required.

pyvista/core/filters/image_data.py
tkoyama010: It is appropriate to raise an error if the type is different. Would that address the issue?

GuillaumeBroggi: Since vtk is not exiting with an error in this situation but silently accepts to filter, for instance, int cell data with a double range, raising an error would be more restrictive than the original vtk class.

At the same time, accepting silently this situation leads to unexpected rounding.

For instance:
```python
segmented_grid = pv.ImageData(dimensions=(4, 3, 3))
segmented_grid.cell_data["Data"] = [0, 0, 0, 1, 0, 1, 1, 1, 0, 0, 0, 0]
connected = segmented_grid.label_connectivity(scalar_range=[0.7, 2])

pl = pv.Plotter()
_ = pl.add_mesh(connected.threshold(0.5), show_edges=True)
pl.show(jupyter_backend="notebook")
```
leads to only one labeled region because (probably) `0.7` is rounded to `0` by vtk which is not the expected behavior from a pythonic point of view.

image

user27182: There are many unexpected results and/or bugs that arise from implicit dtype conversions with vtk. I think this kind of thing should be caught and prevented from happening. One idea is to use numpy's dtype casting rules to do the check and raise an error, e.g. do `scalar_range.astype(scalars.dtype)`

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1607,280 @@ def _update_and_get_output():
...
+            alg.SetScalarRange(*scalar_range)
+
+        if extraction_mode == 'all':
+            alg.SetExtractionModeToAllRegions()
+        elif extraction_mode == 'largest':
+            alg.SetExtractionModeToLargestRegion()
+        elif extraction_mode == 'seeded':
+            if point_seeds is None:
+                raise ValueError(
+                    '`point_seeds` must be specified when `extraction_mode="point_seeds"`.',
```
user27182, line 1819: Here is an example of where it's helpful to include the full error message in the tests to catch changes like this
```suggestion
                    '`point_seeds` must be specified when `extraction_mode="seeded"`.',
```

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1607,280 @@ def _update_and_get_output():
...
+                )
+            elif isinstance(point_seeds, (Sequence, np.ndarray)):
+                # Ensure points are floats, see pyvista.core.utilities.points
+                point_seeds = np.asanyarray(point_seeds, dtype=float)
+                # PointSet requires vtk >= 9.1.0
+                # See https://docs.pyvista.org/api/core/_autosummary/pyvista.pointset#pyvista.PointSet
+                if pyvista.vtk_version_info >= (9, 1, 0):
+                    point_seeds = pyvista.PointSet(point_seeds)
+                else:
+                    point_seeds = pyvista.PolyData(point_seeds)
```
user27182, line 1829: The DataSet constructors automatically cast to float32 and emit a warning if points are integers, maybe just let the constructor handle this conversion? And if PointSet is not specifically required by the filter, maybe just use PolyData (one less branch to test)?
```suggestion
                point_seeds = pyvista.PolyData(point_seeds)
```

pyvista/core/filters/image_data.py
GuillaumeBroggi: I ended up adding an explicit `force_float` parameter with the commit I pushed a few minutes ago.

When `True` (default), it performs a numpy cast as you just described. When `False`, casting is not performed, which saves a bit of resources and works fine in many cases. The docstrings warn users about possible unexpected results.

pyvista/core/filters/image_data.py
GuillaumeBroggi: I simplified the logic as suggested but did not remove PointSet for now according to the above (incomplete) data.

user27182: Makes sense, yes it seems `PointSet` is better to use generally if cells are not required. As a side note, it's possible to avoid creating cells with `PolyData` by assigning points outside of the constructer, e.g. `p = pv.PolyData(); p.points = points`. Running `timeit` as above but with this change yields similar performance to using `PointSet`. Either way, using `PointSet` is fine, I think I saw missing code coverage here so that was part of the motivation for this change. I'll mark this as resolved.

pyvista/core/filters/image_data.py
user27182: That could work, I think this is fine. Though the main issue was mainly catching a dtype mismatch when a scalar range is specified (of floats) is specified for int inputs, right? So we don't have to _always_ force floats, but just in these specific cases? I think `scalar_range.astype(scalars.dtype)` with `scalars.dtype` instead of `float` would only cast in these specific cases, or maybe even just raise a `TypeError` as suggested by @tkoyama010.

Though like I said this is probably fine as-is. Feel free to mark it as resolved.

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1608,293 @@ def _update_and_get_output():
...
+            scalars,
+        )  # args: (idx, port, connection, field, name)
+
+        # Set the scalar range considered for connectivity
+        # Default to vtk default 0.5 to VTK_DOUBLE_MAX, nothing to do
+        # See https://vtk.org/doc/nightly/html/classvtkImageConnectivityFilter.html
+        if scalar_range is not None:
+            if isinstance(scalar_range, str) and scalar_range == 'auto':
+                unique_scalars = np.unique(input_mesh.point_data[scalars])
+                scalar_range = [unique_scalars[1], unique_scalars[-1]]
```
user27182, line 1824: Use get data range instead
```suggestion
                scalar_range = input_mesh.get_data_range(scalars, preference='point') 
```

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1608,293 @@ def _update_and_get_output():
...
+            field.value,
+            scalars,
+        )  # args: (idx, port, connection, field, name)
+
+        # Set the scalar range considered for connectivity
+        # Default to vtk default 0.5 to VTK_DOUBLE_MAX, nothing to do
+        # See https://vtk.org/doc/nightly/html/classvtkImageConnectivityFilter.html
+        if scalar_range is not None:
+            if isinstance(scalar_range, str) and scalar_range == 'auto':
+                unique_scalars = np.unique(input_mesh.point_data[scalars])
```
user27182, line 1823: I like your default of using the full data range. This matches how the `connectivity` filter also works by default. I think setting to `None` for the vtk default a bit odd in term of API, since actively setting to `None` is actually setting a range, not remove it.

What about using `None` to mean `'auto'` instead? I.e. use `None` to use the full data range (by default). And then use the term `foreground` to use the VTK default? extracting foreground connectivity is a common use case in image processing and is effectively what the VTK default is doing with `[0.5, VTK_DOUBLE_MAX]`. 

Feel free to disagree though and simply mark this as resolved and keep using `'auto'`.

pyvista/core/filters/image_data.py
```diff
@@ -1604,3 +1608,293 @@ def _update_and_get_output():
...
+        label_mode: Literal['size', 'constant', 'seeds'] = 'size',
+        constant_value: int | None = None,
+        inplace: bool = False,
+        progress_bar: bool = False,
+    ) -> tuple[pyvista.ImageData, NDArray[int], NDArray[int]]:
+        """Find and label connected regions in a :class:`~pyvista.ImageData`.
+
+        Only points whose `scalar` value is within the `scalar_range` are considered for
+        connectivity. A 4-connectivity is used for 2D images or a 6-connectivity for 3D
+        images. This filter operates on point-based data. If cell=based data are provided,
```
user27182, line 1629:
```suggestion
        images. This filter operates on point-based data. If cell-based data are provided,
```

user27182 (commented):
A few final comments/suggestions

pyvista/core/filters/image_data.py
GuillaumeBroggi: Agreed, explicit is better.

So now there are 3 different keywords:
- `auto`: uses the full range, with the same behaviour as `connectivity` filter
- `foreground`: automatically determine the foreground using the full range minus the smallest value. This is very similar to the default vtk value but more flexible, allowing to not cast the data to floats.
- `vtk_default`

pyvista/core/filters/image_data.py
GuillaumeBroggi: IMO, the main issue is ensuring that the user gets the expected behavior. For instance, when using a `scalar_range=[0.6, 10]`, one would expect the `0` values in `[0, 0, 1,  4, 2, 0]` to be excluded. This is not the case. In short, we need to cast to floats when the `scalar_range` is not made of whole numbers.

Raising a `TypeError` in this case would serve the purpose but prevent the use of integer data. 

Casting systematically to float is also fine but requires additional resources and is not required in some situations, for instance, `scalar_range=[1, 10]`. Note that vtk expects doubles for `scalar_range,` so it always gets cast to floats.

GuillaumeBroggi: Removed `force_float`. Cast to floats only if the data are integers and the bounds are not whole numbers.

GuillaumeBroggi:
Thanks for your detailed feedback!
The reviews have been addressed and all checks are successful, so the PR is ready for your approval IMO.

user27182:
LGTM, thanks! Though I haven't downloaded and reviewed the docs. I saw some single backticks but I think RST requires two backticks and was going to have a look. But, we can fix formatting if needed after it gets deployed to https://dev.pyvista.org

tkoyama010 (approved):

user27182:
The single backticks are indeed italics in RST, and not code literals. The docstring should be updated to use double backticks. https://dev.pyvista.org/api/core/_autosummary/pyvista.imagedatafilters.label_connectivity

PR #6664, opened 2024-09-19
Add ruff formating to .git-blame-ignore-revs

MatthewFlamm:
### Overview

Added commit from #6607, #6632, and #6657 to `.git-blame-ignore-revs`

### Details

Also added PR link to each existing commit listed for easier cross reference.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -11,5 +11,9 @@ a26a52d037a265774be0ca06bb141e3346e73ad0
 635052d1bce8615fcc4879f5f200921a4f78aa1a
 
 # ruff formatting
+# https://github.com/pyvista/pyvista/pull/6607
+7c4166477b875d40e4650b102f67a29a40c2d8f2
+# https://github.com/pyvista/pyvista/pull/6632
+7f8ba86bbe07f85380937bc8d7561775e6a1d4d8
 # https://github.com/pyvista/pyvista/pull/6657
 c0396f73936e1772bc86bde7438e10b9ad19547b
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,7 +1,15 @@
 # enforce unified code style with black
+# https://github.com/pyvista/pyvista/pull/2180
 28914fc2236cd5850f536b151d0ef383855d8519
+# https://github.com/pyvista/pyvista/pull/2212
 db0691cfb6b033e55664ffba25deb456a386a48c
 
 # enforce unified code style in our documentation with blackdoc
+# https://github.com/pyvista/pyvista/pull/4119
 a26a52d037a265774be0ca06bb141e3346e73ad0
+# https://github.com/pyvista/pyvista/pull/4129
 635052d1bce8615fcc4879f5f200921a4f78aa1a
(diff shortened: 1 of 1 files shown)


MatthewFlamm:
There are more ruff related commits, but some were either small number of lines or had more substantial changes than formatting.  The pytest parantheses commit change is innocuous enough to include here IMO.

.git-blame-ignore-revs
```diff
@@ -1,7 +1,19 @@
...
+# https://github.com/pyvista/pyvista/pull/4119
 a26a52d037a265774be0ca06bb141e3346e73ad0
+# https://github.com/pyvista/pyvista/pull/4129
 635052d1bce8615fcc4879f5f200921a4f78aa1a
+
+# ruff formatting
+# https://github.com/pyvista/pyvista/pull/6607
+7c4166477b875d40e4650b102f67a29a40c2d8f2
+# https://github.com/pyvista/pyvista/pull/6632
+7f8ba86bbe07f85380937bc8d7561775e6a1d4d8
```
tkoyama010, line 17: I learned from https://github.com/pyvista/pyvista/pull/5059#issuecomment-1760334221 that the criterion for adding to .git-blame-ignore-revs is whether or not it is automatically formatted, rather than the number of lines. For that reason, I think it would be better to delete this commit as it is human handed change.
```suggestion
```

.git-blame-ignore-revs
```diff
@@ -1,7 +1,19 @@
 # enforce unified code style with black
+# https://github.com/pyvista/pyvista/pull/2180
```
tkoyama010, line 2: It's good idea to add link of PR.

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

adeak:
Thanks for doing this, and being careful, and thanks @tkoyama010 for the heads-up.

I'm also a bit wary of the remaining PRs. In #6657 we have https://github.com/pyvista/pyvista/pull/6657/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R290 and a bit too many commits, e.g. https://github.com/pyvista/pyvista/pull/6657/commits/d0692aadb393f43517609b6c72125ab301fba903 (but such changes seem to have been reverted).

Similarly, in #6607 we have https://github.com/pyvista/pyvista/pull/6607/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R98. Losing the respective config change in `git blame` might not be ideal.

Unfortunately with the squash merge model we use it's hard to end up with a pure-formatting commit accidentally. One has to go out of their way to craft PRs such that they are autoformatting-only, specifically with ignore-revs files in mind. While I love the feature, it's purely for convenience, and non-autoformat (non-auto or non-format) changes being caught up and hidden in the blame is a much larger risk in my opinion.

MatthewFlamm:
Hmmm I see what you mean about the config change.  The alternative is also annoying to have these formatting commits clog up the blame.  Neither is great with the merge model.

I will close this PR if you agree @tkoyama010

PR #6676, opened 2024-09-24
Improve workflow when manipulating `ImageData` (and other grids) dimensionality

GuillaumeBroggi:
### Overview

> [!NOTE]  
> This PR rationale was significantly updated since the initial post. See @user27182 comments below for details and discussions in #6678 and #6679 for a full overview.

`ImageData` filters do not modify singleton dimensions. For instance,
```python
tube = pv.Circle().extrude((0, 0, 0.1), capping=False)
binary_tube = tube.voxelize_binary_mask(dimensions=(10, 10, 1)).points_to_cells()
binary_tube.plot(jupyter_backend="notebook")
```
image

The rationale behind this behavior is that preserving dimensionality is useful when working with images. However, the `ImageData` structure is also useful to work with "image-like" data structures, such as regular meshes., that can benefit from a better handling of dimensionality.

For instance, obtaining a one-layer mesh from the previous example is possible with this PR. 

```python
tube = pv.Circle().extrude((0, 0, 0.1), capping=False)
binary_tube = tube.voxelize_binary_mask(dimensions=(10, 10, 1)).points_to_cells(dimensionality='3D')
binary_tube.plot(jupyter_backend="notebook")
```
image

```python
binary_tube.dimensionality
```
> 3

### Details

- A new helper internal method `_validate_dim_operation` is defined for `ImageDataFilters`. It allows for a flexible definition of dimensional operation through the parameter `dimensionality`.
-  For instance, `dimensionality=True` will apply the operation on all dimensions while `dimensionality=[True, True, False]` will apply only on the first two. 
- Additionally, it provides many convenience aliases to automatically select the dimensions to modify based on the desired outputs. For instance, if possible, `dimensionality='1D'` will ensure that the output is 1D. An error is raised if the resulting operation is impossible.
- A brute force approach is used to select the dimensions to modify. There are only 9 cases to test; doing so has (probably) no significant impact on performance but simplifies the code logic significantly.
- Finally, a `'preserve'` alias is set as default to reproduce the current behavior by default: singleton dimensions are not modified. Hence, the dimensionality is preserved. 
----
- `Grid` (and related `PointSet`) implement a `dimensionality` property that conveniently returns the data dimensionality. This is intended to make image / structured mesh-related workflow more friendly.
----
This is probably out of the scope of this PR, but I wonder if offering a "NumPy-like" indexing for `ImageData` would be possible.

For instance, let's assume we are working with a 3D mesh:
```python
mesh = pv.ImageData(dimensions=(11, 11, 11))
mesh.cell_data['data'] = np.arange(image.n_cells)
mesh.plot(jupyter_backend="notebook")
```
image

Manipulating it as a 3D NumPy array requires a reshape:
```python
mesh.cell_data['data'].reshape((10,10,10))[:,:,0]=2000
mesh.plot(jupyter_backend="notebook")
```
image

Being able to directly index the data would simplify this workflow.

### TODO
- [x] Agree on naming
- [x] Agree on default behavior
- [x] ~Discuss indexing~ out of scope
- [x] Document version and depreciation if suitable
- [x] Review


user27182:
This feature to increase singleton dimensions was actually considered to be a bug when `points_to_cells` was initially written, see https://github.com/pyvista/pyvista/pull/6071#issuecomment-2116331027. When working with 2D images it's preferable for the image to remain 2D after remeshing.

I suggest adding a new keyword to control this behaviour. For example, the [pad_image](https://dev.pyvista.org/api/core/_autosummary/pyvista.imagedatafilters.pad_image#pyvista.ImageDataFilters.pad_image) filter has a `pad_singleton_dims` option to enable similar functionality. Maybe call the new keyword `remesh_singleton_dims` or similar?

GuillaumeBroggi:
Thanks for the rationale, that makes sense.

I will implement your suggestion then.

user27182:
Thanks, for more of a rationale (which as you pointed out is not documented with the filter), the intent of points_to_cells and cells_to_points is to be able to represent voxel points as voxels cells and vice versa. And so by extension we also want pixel points to be represented as pixel cells. And so that's where the reasoning behind keeping 2D points as 2D cells comes from. 

Also, `remesh_singleton_dims` seems a bit verbose, if you have a better name for it...

pyvista/core/filters/image_data.py
```diff
@@ -1331,17 +1504,28 @@ def pad_image(
...
+                dimension basis.
+            - ``'0D'``: convenience alias to output a 0D ImageData, if possible, i.e.,
+                dims will be similar to (1, 1, 1).
+            - ``'1D'``: convenience alias to output a 1D ImageData, if possible, i.e.,
+                dims will be similar to (>1, 1, 1).
+            - ``'2D'``: convenience alias to output a 2D ImageData, if possible, i.e.,
+                dims will be similar to (>1, >1, 1).
+            - ``'3D'``: convenience alias to output a 3D ImageData, if possible, i.e.,
+                dims will be similar to (>1, >1, >1).
+            - ``'preserve'``: convenience alias to not modify singleton dimensions.
```
user27182, line 1522: Suggest removing `True` and `False` options as suggested earlier, and using padding-specific language. Also, the `0D` case is only valid here if the filter does nothing (i.e. no padding is applied). Is it worth it to include this case in the docs? 

```suggestion
            Control which dimensions will be padded by the filter.

            - Can be specified as a sequence of 3 boolean apply padding on a per
                dimension basis.
            - ``'1D'``: apply padding such that the output is a 1D ImageData where exactly one dimension is greater than one, e.g. ``(>1, 1, 1)``. Only valid for 0D or 1D inputs.
            - ``'2D'``: apply padding such that the output is a 2D ImageData where exactly two dimensions are greater than one, e.g. ``(>1, >1, 1)``. Only valid for 0D, 1D, or 2D inputs.
            - ``'3D'``: apply padding such that the output is a 3D ImageData, where all three dimensions are greater than one, e.g. ``(>1, >1, >1)``. Valid for any 0D, 1D, 2D, or 3D inputs.
```

pyvista/core/filters/image_data.py
```diff
@@ -1915,3 +2102,143 @@ def label_connectivity(
...
+        (array([ True, False,  True]), array([6, 1, 6]))
+
+        """
+        dims = np.asarray(self.dimensions)  # type: ignore[attr-defined]
+        # Build an array of the operation size
+        operation_size = _validation.validate_array3(operation_size, reshape=True, broadcast=True)
+
+        if not isinstance(operation_mask, str):
+            # Build a bool array of the mask, cast to int first to be a valid input for validate_array3
+            dims_mask = np.asarray(operation_mask).astype(int)
```
user27182, line 2179: We should allow `bool` types to be used directly. I'll create a separate PR to remove this restriction. 
```suggestion
```

pyvista/core/filters/image_data.py
```diff
@@ -1915,3 +2102,143 @@ def label_connectivity(
...
+            dims_mask = _validation.validate_array3(
+                dims_mask, reshape=True, broadcast=True, dtype_out=bool
```
user27182, line 2181: We should require the input dtype to be bool rather than casting the output since any object can be converted to a bool.
```suggestion
                dims_mask, reshape=True, broadcast=True, must_have_dtype=bool
```

pyvista/core/filters/image_data.py
```diff
@@ -1915,3 +2102,143 @@ def label_connectivity(
...
+                raise ValueError(
+                    f'The operation requires to {operator.__name__} at least {operation_size} dimension(s) to {self.dimensions}.'  # type: ignore[attr-defined]
+                    f' A {operation_mask} ImageData with dims {desired_dims} cannot be obtained.'
+                )
+
+        # Check that the result dim is feasible
+        dims_result = operator(dims, operation_size * dims_mask)
+
+        if not (dims_result >= 1).all():
+            raise ValueError(
```
user27182, line 2239: We should add a test case for this

tests/core/test_grid.py
```diff
@@ -1611,3 +1611,15 @@ def test_copy_no_copy_wrap_object_vtk9(datasets_vtk9):
...
+@pytest.mark.parametrize('grid_class', [pv.RectilinearGrid, pv.ImageData])
+@pytest.mark.parametrize('dim', [0, 1, 2, 3])
+def test_grid_dimensionality(grid_class, dim):
+    dims = {0: (1, 1, 1), 1: (1, 42, 1), 2: (42, 1, 142), 3: (2, 42, 142)}[dim]
+    if grid_class == pv.ImageData:
+        grid = grid_class(dimensions=dims)
+    elif grid_class == pv.RectilinearGrid:
+        grid = grid_class(range(dims[0]), range(dims[1]), range(dims[2]))
+
+    assert grid.dimensionality == dim
```
user27182, line 1625: Suggest parametrize dims and dim together instead of using dict, and use the full terms `dimensions` and `dimensionality` instead of `dims` and `dim` for clarity. Also add check to verify dimensionality matches the cell dimensionality.
```suggestion
@pytest.mark.parametrize(('dimensionality','dimensions'), [(0, (1, 1, 1)), (1, (1, 42, 1)), (2, (42, 1, 142)), (3,(2, 42, 142))])
def test_grid_dimensionality(grid_class, dimensionality, dimensions):
    if grid_class == pv.ImageData:
        grid = grid_class(dimensions=dimensions)
    elif grid_class == pv.RectilinearGrid:
        grid = grid_class(range(dimensions[0]), range(dimensions[1]), range(dimensions[2]))

    assert grid.dimensionality == dimensionality
    assert grid.dimensionality == grid.get_cell(0).GetCellDimension()
```

tests/core/test_pointset.py
```diff
@@ -352,3 +352,30 @@ def test_rotate():
...
+    [
+        (pv.ExplicitStructuredGrid, 3),
+        (pv.StructuredGrid, 0),
+        (pv.StructuredGrid, 1),
+        (pv.StructuredGrid, 2),
+        (pv.StructuredGrid, 3),
+    ],
+)
+def test_pointgrid_dimensionality(grid_class, dim):
+    dims = {0: (1, 1, 1), 1: (1, 42, 1), 2: (42, 1, 142), 3: (2, 42, 142)}[dim]
```
user27182, line 368: suggest parametrizing dims and dim together instead of using a dict (similar to the other grid test)

tests/core/test_pointset.py
```diff
@@ -352,3 +352,30 @@ def test_rotate():
...
+    elif grid_class == pv.StructuredGrid:
+        x, y, z = np.meshgrid(
+            np.arange(dims[0], dtype=np.float32),
+            np.arange(dims[1], dtype=np.float32),
+            np.arange(dims[2], dtype=np.float32),
+            indexing='ij',
+        )
+        grid = grid_class(x, y, z)
+
+    assert grid.dimensionality == dim
```
user27182, line 381: Make sure this matches the cell dimension
```suggestion
    assert grid.dimensionality == dim
    assert grid.dimensionality == grid.get_cell(0).GetCellDimension()
```

tests/core/test_imagedata_filters.py
```diff
@@ -232,7 +233,25 @@ def test_points_to_cells_and_cells_to_points_dimensions(uniform, logo):
 
     assert logo.dimensions == (1920, 718, 1)
     assert logo.points_to_cells().dimensions == (1921, 719, 1)
+    assert logo.points_to_cells(dimensionality='2D').dimensions == (1921, 719, 1)
+    assert logo.points_to_cells(dimensionality=(True, True, False)).dimensions == (1921, 719, 1)
     assert logo.cells_to_points().dimensions == (1919, 717, 1)
+    assert logo.points_to_cells(dimensionality=True).dimensions == (1921, 719, 2)
+    assert logo.points_to_cells(dimensionality='3D').dimensions == (1921, 719, 2)
+
```
user27182, line 241: Should we also add 0D and 1D test cases here?

pyvista/core/filters/image_data.py
```diff
@@ -1022,6 +1039,27 @@ def points_to_cells(self, scalars: str | None = None, *, copy: bool = True):
             By default, all point data arrays at the input are passed through as cell
             data at the output.
 
+        dimensionality : bool, VectorLike[bool], Literal['0D', '1D', '2D', '3D', 'preserve'], default: 'preserve'
+            Control which dimensions will be modified by the filter.
+
+            - ``True``: all dimensions are modified.
+            - ``False``: no dimensions are modified.
```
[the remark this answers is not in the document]
GuillaumeBroggi: > I'm not sure if these options should be included... isn't `True` equivalent to `'3D'`?

It's correct that `True` is equivalent to `'3D'`. However, I consider `'1D'`, `'2D'`, `'3D'`, etc. as aliases for boolean vectors.

> [At] least one dimension _must_ change for 1D inputs, at least two for 2D, and all three for 3D, the conversion to be valid?

Indeed, there is no filter for which `False` makes sense except `cells_to_points` and `points_to_cells` for 0D input, but I considered it for the sack of generability (a future filter could require it?). For the 2D and 3D inputs, it is, for instance, possible to pad only one dimension and get a valid output.

pyvista/core/filters/image_data.py
GuillaumeBroggi: I'll keep this unresolved in the meantime.

pyvista/core/filters/image_data.py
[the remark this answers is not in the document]
user27182: > > I'm not sure if these options should be included... isn't `True` equivalent to `'3D'`?
> 
> If we have boolean vectors, I think it is nice to automatically broadcast `True` and `False` but it might not be required to document it.

I like your general implementation to use boolean vectors. My main concern is with the broadcasting specifically and allowing `True` and `False` scalars. This is because I find `dimensionality=True` unclear from a plain-language perspective... unless you look at the docs it's not immediately clear what this means. `dimensionality='3D'` is much clearer I think, so why not just use that instead? And, forcing a vector in this case (e.g. `dimensionality=(True,True,True)` hints that a mask is being specified, otherwise using simply `True` or `False` could be (incorrectly) interpreted as an option that is being enabled.

> > [At] least one dimension _must_ change for 1D inputs, at least two for 2D, and all three for 3D, the conversion to be valid?
> 
> Indeed, there is no filter for which `False` makes sense except `cells_to_points` and `points_to_cells` for 0D input, but I considered it for the sack of generability (a future filter could require it?). 

This would still be supported by specifying `dimensionality=(False, False, False)` explicitly. Again, I like the generalization, but not sure if we should be supporting the broadcasting.

> For the 2D and 3D inputs, it is, for instance, possible to pad only one dimension and get a valid output.

In these cases, isn't it necessary to provide the full length-3 mask? e.g. `(True, True, False)`? These cases are still supported if the broadcasting option is removed.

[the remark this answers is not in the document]
GuillaumeBroggi: That makes sense, thanks for your input.

GuillaumeBroggi:
@user27182  I addressed the comments and implemented the suggested improvements.

Is the naming and default behavior OK?

The tests should succeed once #6687 is merged.

pyvista/core/pointset.py
```diff
@@ -2497,6 +2497,31 @@ def dimensions(self, dims):  # numpydoc ignore=GL08
...
+        >>> xrng = np.arange(-10, 10, 1, dtype=np.float32)
+        >>> yrng = np.arange(-10, 10, 2, dtype=np.float32)
+        >>> zrng = np.arange(-10, 10, 5, dtype=np.float32)
+        >>> x, y, z = np.meshgrid(xrng, yrng, zrng, indexing='ij')
+        >>> grid = pv.StructuredGrid(x, y, z)
+        >>> grid.dimensionality
+        3
+
+        """
+        dims = np.asarray(self.dimensions)
```
GuillaumeBroggi, line 2522: Codecov produces a missing coverage but I don´t  know why since the lines are covered, see https://app.codecov.io/gh/pyvista/pyvista/pull/6676

tkoyama010: I checked the codecov link, and it seems that the tests are covered. Is it okay to merge this?

GuillaumeBroggi: For me, it's okay to merge it, since:
- I believe I have addressed all @user27182 comments
- I have been using the feature introduced by this PR for 2 weeks without issues

But ofc I have a conflict of interest there ;)

tkoyama010: We have waited long enough for this and will merge. Please open a new PR if someone wish to add something.

PR #6678, opened 2024-09-25
Add `remesh_singleton_dims` option to `points_to_cells` to remesh 2D inputs as 3D outputs

user27182:
### Overview

Add option from https://github.com/pyvista/pyvista/pull/6676#issuecomment-2372544523

EDIT: I created a new branch instead of branching from #6676 so that we can preview the docs (preview is only available for branches from the `pyvista` repo)

### DIFF
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -978,7 +978,9 @@ class ImageDataFilters(DataSetFilters):
         _vtk.vtkLogger.SetStderrVerbosity(verbosity)
         return cast(pyvista.PolyData, wrap(alg.GetOutput()))
 
-    def points_to_cells(self, scalars: str | None = None, *, copy: bool = True):
+    def points_to_cells(
+        self, scalars: str | None = None, *, remesh_singleton_dims: bool = False, copy: bool = True
+    ):
         """Re-mesh image data from a point-based to a cell-based representation.
 
         This filter changes how image data is represented. Data represented as points
@@ -1021,6 +1023,12 @@ class ImageDataFilters(DataSetFilters):
             this parameter to restrict the output to only include the specified array.
             By default, all point data arrays at the input are passed through as cell
             data at the output.
+
+        remesh_singleton_dims : bool, default: False
+            Control if singleton dimensions are re-meshed. By default, singleton
+            dimensions are not re-meshed, which means that 1D or 2D inputs will remain
+            1D or 2D at the output. Set this to ``True`` to allow singleton dimensions
+            to increase.
 
         copy : bool, default: True
             Copy the input point data before associating it with the output cell data.
@@ -1080,7 +1088,36 @@ class ImageDataFilters(DataSetFilters):
         - The output has one less array (the input cell data is ignored)
         - The dimensions have increased by one
         - The bounds have increased by half the spacing
-        - The output N Cells equals the input N Points
+        - The output ``N Cells`` equals the input ``N Points``
+
+        Since the input points are 3D (i.e. there are no singleton dimensions), the
+        output cells are 3D :attr:`~pyvista.CellType.VOXEL` cells.
+
+        >>> cells_image.get_cell(0).type
+        <CellType.VOXEL: 11>
+
+        If the input points are 2D (i.e. one dimension is singleton), the
+        output cells are 2D :attr:`~pyvista.CellType.PIXEL` cells.
+
+        >>> image2D = examples.load_logo()
+        >>> image2D.dimensions
+        (1920, 718, 1)
+
+        >>> pixel_cells_image = image2D.points_to_cells()
+        >>> pixel_cells_image.dimensions
+        (1921, 719, 1)
+        >>> pixel_cells_image.get_cell(0).type
+        <CellType.PIXEL: 8>
+
+        Use ``remesh_singleton_dims`` to remesh 2D points as 3D cells.
+
+        >>> voxel_cells_image = image2D.points_to_cells(
+        ...     remesh_singleton_dims=True
+        ... )
+        >>> voxel_cells_image.dimensions
+        (1921, 719, 2)
+        >>> voxel_cells_image.get_cell(0).type
+        <CellType.VOXEL: 11>
 
         See :ref:`image_representations_example` for more examples using this filter.
 
@@ -1091,7 +1128,12 @@ class ImageDataFilters(DataSetFilters):
                 raise ValueError(
                     f"Scalars '{scalars}' must be associated with point data. Got {field.name.lower()} data instead.",
                 )
-        return self._remesh_points_cells(points_to_cells=True, scalars=scalars, copy=copy)
+        return self._remesh_points_cells(
+            points_to_cells=True,
+            scalars=scalars,
+            remesh_singleton_dims=remesh_singleton_dims,
+            copy=copy,
+        )
 
     def cells_to_points(self, scalars: str | None = None, *, copy: bool = True):
         """Re-mesh image data from a cell-based to a point-based representation.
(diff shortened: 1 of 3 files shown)


user27182:
@GuillaumeBroggi I tried adding you as co-author in the commit message but it didn't get recognized. Feel free to make any suggestion to get co-author credit.
Also, it seems I can't request your review officially, but please take a look if you have time.

MatthewFlamm:
I like this version better, but how does the spacing work?  

Is there already a 3rd spacing value specified and it uses that value?  If so, then maybe documenting this behavior would be appropriate.

user27182:
Good question. Spacing is passed through unmodified by the filter, regardless of 2D or 3D inputs. Spacing basically has no effect when one of the dimensions is one and can be any value. I can update the docs to make this clear

user27182:
> @GuillaumeBroggi I tried adding you as co-author in the commit message but it didn't get recognized. Feel free to make any suggestion to get co-author credit. Also, it seems I can't request your review officially, but please take a look if you have time.

I was able to revert the initial commit and fix the co-author attribution with https://github.com/pyvista/pyvista/pull/6678/commits/823f6afa1550cd5c3f6f152f57abb9406c4c849c

user27182:
> I like this version better, but how does the spacing work?
> 
> Is there already a 3rd spacing value specified and it uses that value? If so, then maybe documenting this behavior would be appropriate.

Docs have been updated in https://github.com/pyvista/pyvista/pull/6678/commits/16ce8c6513eacb43736b761badbb04beeff7836e, hopefully this makes it clear.

GuillaumeBroggi:
Thanks @user27182!

One use case is not covered (correct me if I am wrong). Let's assume we have a 1D mesh: we can only re-mesh to 1D or 3D, not 2D, right?

Could it make sense to enable explicitly and per dimension increase and decrease in that light?
For instance, `modify_dims = (True, True, False)` to get a 2D mesh from a 1D mesh.

user27182:
I believe you are correct, this is a limitation of the current implementation. I like your proposal, it's more flexible instead of having a simple Boolean flag.
By default it could be `modify_dims=None` and keep the current behaviour of `remesh_singleton_dims=False`, but otherwise use `modify_dims` to effectively set the internal `dims_mask` variable directly

GuillaumeBroggi:
To be more specific, what about `'non_singleton'` instead of `None?

If not for backward compatibility, would you maintain this default behavior?

user27182:
> To be more specific, what about `'non_singleton'` instead of `None?

This is fine, so `None` would not be accepted as an input then?

> If not for backward compatibility, would you maintain this default behavior?

I am fine with changing the default to operate on all dimensions. We can probably address backwards compatibility with deprecations. But my main concern is having consistent behaviour across filters where possible. E.g. `pad_image` does not pad singleton dimensions by default, so changing the default for `points_to_cells` would make these two filters behave differently for these cases. 

I'm not sure if any other filters handle this specifically though. In general I'd probably guess that most filters don't consider this, though I'd like to see better support for 2D images in the library (e.g. it would be nice to have something like `plot_texture` but for 2D images). In that regard the current behaviour is more 2D-image friendly. Even if working with 3D volumes, the current behaviour is such that filtering all 2D slices of a volume separately is equivalent to filtering the entire volume as a whole, which I think is nice.

My background is in image processing though, so I have a bias where I use ImageData with actual images and not as a more general purpose data structure, so am otherwise unfamiliar with other use cases.

tkoyama010:
I'm currently unwell and unable to review GitHub (It should recover in about a week). If someone else comments LGTM on this pull request, I will consider that as approval. Thank you.

GuillaumeBroggi:
> I am fine with changing the default to operate on all dimensions. We can probably address backwards compatibility with deprecations. But my main concern is having consistent behaviour across filters where possible. E.g. `pad_image` does not pad singleton dimensions by default, so changing the default for `points_to_cells` would make these two filters behave differently for these cases.
> 
> I'm not sure if any other filters handle this specifically though. In general I'd probably guess that most filters don't consider this, though I'd like to see better support for 2D images in the library (e.g. it would be nice to have something like `plot_texture` but for 2D images). In that regard the current behaviour is more 2D-image friendly. Even if working with 3D volumes, the current behaviour is such that filtering all 2D slices of a volume separately is equivalent to filtering the entire volume as a whole, which I think is nice.

In that case, we could have a helping function with a signature similar to:
```python
_get_dims_mask(dims: NDArray[int], resize_dims: bool| VectorLike[bool] | Literal['1D', '2D', '3D'], dims_operator)
```
that we could use in any filter (only `points_to_cells`, `cells_to_points`, and `pad_image`) to consistently indicate which dimensions can be increased or decreased. The litterals  would be helpers to automatically obtained the desired output dimensions based on the input:
- `'1D'` -> output dims are `[N, 1, 1]` or `[1, N, 1]` or `[1, 1, N]`
- `'2D'` -> output dims are `[N, N, 1]` or `[1, N, N]` or `[1, N, N]`
- `'3D'` -> output dims are `[N, N, N]`

I am drafting code to implement this idea, let me know if you like it.

user27182:
> I am drafting code to implement this idea (see #6679), let me know if you like it.

Sounds great! Thanks for working on this

pyvista/core/filters/image_data.py
```diff
@@ -978,7 +979,13 @@ def contour_labeled(
         _vtk.vtkLogger.SetStderrVerbosity(verbosity)
         return cast(pyvista.PolyData, wrap(alg.GetOutput()))
 
-    def points_to_cells(self, scalars: str | None = None, *, copy: bool = True):
+    def points_to_cells(
+        self,
+        scalars: str | None = None,
+        *,
+        resize_dims: bool | VectorLike[bool] | Literal['1D', '2D', '3D'] = True,
```
user27182, line 986: @MatthewFlamm Looking for some additional feedback on the default behaviour here. With this line it's currently set to resize inputs, i.e. 1D and 2D inputs become 3D by default. But, this breaks backwards compatibility, since `points_to_cells` (and `pad_image`) were both originally written _not_ resize inputs. So is it preferred to:

1. break backwards compatibility and make this filter resize all dimensions by default
2. keep the original behaviour and only resize non-singleton dimensions

The original "only resize non-singleton" dimensions behaviour makes sense to me for working with actual 2D or 3D images. But I don't think most filters in pyvista were written with 2D image cases specifically in mind with special handling like this. So, what to do? I am in favour of whichever API produces the more consistent and expected "pyvista-like" experience. Based on the historical naming of `ImageData` as `UniformGrid` this class may have been less geared towards working with images specifically...

GuillaumeBroggi: For context, I set the default to `True` because it's the less likely-to-fail option.

The original "only resize non-singleton" dimensions are not exactly achievable with the current state of the PR, but the same intended functionality (i.e., working with 1D and 2D images)  can be obtained, IMO, with the literals `'1D'` and `'2D'`. 

We could add a `'non-singleton'` literal as a fourth option to maintain the previous behaviour, but I am biased against it for two reasons:
- the expected result is not explicit.
- as long the user knows what are the input dims and what are the required output dims, it does not bring extra functionality.

But I would still be happy to have it if you think it's useful :wink:

user27182: I like specifying the dimensionality with the literals, and am fine with the changes generally. Since these are breaking changes though and I was thinking it might be good to get another opinion on the matter.

user27182:
@GuillaumeBroggi  I just got around to updating the code to move this further along but it's clear the solution for the fix is a bit more involved/complex than I thought. I also don't have much time in the next few weeks to work on this. Perhaps it would be best if you re-open #6676 and continue implementing this feature on that branch directly rather than through this PR. It's probably easier for you to resolve some of the intermediate bugs this way too (e.g. with setting 0 dimensions).

GuillaumeBroggi:
No problem, thanks for your help! I have re-opened #6676 and I will continue development there.

PR #6700, opened 2024-10-11
Add fine grain execution control for pyvista-plot directive

bjlittle:
### Overview

This pull-request adds some fine grain options to control the execution of the `pyvista-plot` directive.

#### Motivation

Building documentation can be expensive, particularly if it's image rich, which is pretty much the accepted norm here. However, sometimes as a developer there is a need for a slicker workflow experience when changing the documentation i.e., change the docs, build the docs, review the rendered docs ... typically this can involve several rinse'n repeat iterations, but the major bottleneck is always the documentation build time.

I'm proposing a couple of minor changes that will hopefully give developers the ability to improve the turnaround of their documentation changes by rendering only images of their choosing, which may be none, all or some.

##### Directive Options

The `pyvista-plot` directive has the following **new** options:

- `:skip:` which has an optional boolean argument i.e., `yes`, `1`, `true`, `no`, `0`, `false`. If no argument is provided, then it is assumed to be enabled e.g., `:skip:` is equivalent to `:skip: yes`. The purpose of this option is to override the behaviour of the `plot_skip` configuration variable, and enforce the execution of the directive or not
- `:optional:` which is a flag (no argument). The purpose of this option is to mark a directive for conditional execution, which is controlled by the state of the `plot_skip_optional` configuration variable.

##### Configuration Variables (`conf.py`)

The `pyvista-plot` directive has the following **new** configuration variables:

- `plot_skip` is a boolean. Default is `False`. This variable is used to seed the value of any directive that doesn't explicitly specify a `:skip:` option
- `plot_skip_optional` is a boolean. Default is `False`. This variable is used to control whether any directive marked with an `:optional:` flag is executed or not

#### Workflow

Note that, given the default values of the **new** `conf.py` configuration variables there is no change in the existing behaviour of the `pyvista-plot` directive.

Here are some example workflows that are now easily possible:

- disable all `pyvista-plot` directives with `plot_skip = True` in the `conf.py`
- disable all `pyvista-plot` directives with `plot_skip = True` apart from one (or more) marked with `:skip: no`
- execute all `pyvista-plot` directives with `plot_skip = False` and `plot_skip_optional = False` (default behaviour)
- execute all `pyvista-plot` directives with `plot_skip = False` apart from one (or more) marked with `:optional:` because `plot_skip_optional = True`

Note that, skipped `pyvista-plot` directives will still have their code blocks rendered but with no image and no caption.

### DIFF
--- a/pyvista/ext/plot_directive.py
+++ b/pyvista/ext/plot_directive.py
@@ -55,8 +55,8 @@ The ``pyvista-plot`` directive supports the following options:
         directives for which the ``:context:`` option was specified.  This only
         applies to inline code plot directives, not those run from files.
 
-    nofigs : bool, default None
-        If specified, the code block will be run, but no figures will be
+    nofigs : None
+        When setting this flag, the code block will be run but no figures will be
         inserted.  This is usually useful with the ``:context:`` option.
 
     caption : str
@@ -64,10 +64,11 @@ The ``pyvista-plot`` directive supports the following options:
         figure. This overwrites the caption given in the content, when the plot
         is generated from a file.
 
-    force_static : bool, default None
-        If specified, static images will be used instead of an interactive scene.
+    force_static : None
+        When setting this flag, static images will be used instead of an
+        interactive scene.
 
-    skip : bool
+    skip : bool, default: True
         Whether to skip execution of this directive. If no argument is provided
         i.e., ``:skip:``, then it defaults to ``:skip: true``.  Default
         behaviour is controlled by the ``plot_skip`` boolean variable in
@@ -87,7 +88,7 @@ include *alt*, *height*, *width*, *scale*, *align*.
 **Configuration options**
 The plot directive has the following configuration options:
 
-    plot_include_source : bool
+    plot_include_source : bool, default: True
         Default value for the ``include-source`` directive option.
         Default is ``True``.
 
@@ -96,7 +97,7 @@ The plot directive has the following configuration options:
         to.  If ``None`` or unset, file names are relative to the
         directory where the file containing the directive is.
 
-    plot_html_show_formats : bool
+    plot_html_show_formats : bool, default: True
         Whether to show links to the files in HTML. Default ``True``.
 
     plot_template : str
@@ -108,11 +109,12 @@ The plot directive has the following configuration options:
     plot_cleanup : str
         Python code to be run after every plot directive block.
 
-    plot_skip : bool, default False
+    plot_skip : bool, default: False
         Default value for the ``skip`` directive option.
 
-    plot_skip_optional : bool, default False
+    plot_skip_optional : bool, default: False
         Whether to skip execution of ``optional`` directives.
+
 These options can be set by defining global variables of the same name in
 :file:`conf.py`.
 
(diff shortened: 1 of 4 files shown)


bjlittle:
pre-commit.ci autofix

pyvista/ext/plot_directive.py
```diff
@@ -55,7 +55,7 @@
         directives for which the ``:context:`` option was specified.  This only
         applies to inline code plot directives, not those run from files.
 
-    nofigs : bool
+    nofigs : None
```
tkoyama010, line 58: ref https://numpydoc.readthedocs.io/en/stable/format.html#parameters
```suggestion
    nofigs : bool, default None
```

pyvista/ext/plot_directive.py
```diff
@@ -64,10 +64,22 @@
         figure. This overwrites the caption given in the content, when the plot
         is generated from a file.
 
-    force_static : bool
+    force_static : None
```
tkoyama010, line 67:
```suggestion
    force_static : bool, default None
```

pyvista/ext/plot_directive.py
```diff
@@ -95,6 +108,14 @@
     plot_cleanup : str
         Python code to be run after every plot directive block.
 
+    plot_skip : bool
+        Default value for the ``skip`` directive option.  Default is
+        ``False``.
```
tkoyama010, line 113:
```suggestion
    plot_skip : bool, default False
        Default value for the ``skip`` directive option.
```

pyvista/ext/plot_directive.py
```diff
@@ -95,6 +108,14 @@
...
+
+    plot_skip_optional : bool
+        Whether to skip execution of ``optional`` directives.  Default is
+        ``False``.
+
```
tkoyama010, line 118:
```suggestion

    plot_skip_optional : bool, default False
        Whether to skip execution of ``optional`` directives.

```

tkoyama010 (changes requested):

tkoyama010 (changes requested):
Thank you for the change. It would be grateful if you could add a test to the `tests/test_tinypages.py` so that we can understand the specific use of this change.

bjlittle:
@tkoyama010 `uber/h3-py` have just release a breaking major 4.x release ... patching `geovista` now :+1:

bjlittle:
pre-commit.ci autofix

bjlittle:
@tkoyama010 Back to you :+1: 

I've updated the test coverage as you requested. Hopefully this helps.

If you have any further questions then please don't hesitate to ask.

Also, I appear to now have a failing `Vale` CI tests ... not quite sure how to fix that. Can you advise please :pray:

bjlittle:
Ahhh the same `vale` barf appears to be happening with other pull-requests e.g., #6707

So there seems to be a systemic failure here, but since I don't use and understand `vale` not quite sure what to do ...

tkoyama010:
> Ahhh the same `vale` barf appears to be happening with other pull-requests e.g., #6707
> 
> So there seems to be a systemic failure here, but since I don't use and understand `vale` not quite sure what to do ...

It is resolved in #6708 .

bjlittle:
@tkoyama010 Okay, back to you as the `vale` CI job seems to be behaving itself again.

Thanks :+1:

tkoyama010 (approved):
LGTM

bjlittle:
@tkoyama010 Awesome! Thanks! Much appreciated :beers: :partying_face: :tada:

PR #6701, opened 2024-10-12
Drop Python 3.8 support which is EOL in 07 Oct 2024

tkoyama010:
### Overview

This PR is a duplicate of #5623. However, #5623 has a more significant divergence from the main branch, which is why this PR was opened.

Close #5623.

### Details

- None


MatthewFlamm:
I am +1 on this.  It is unfortunate that we squash commits here as the formatting changes would be better as a separate commit.

MatthewFlamm:
There are probably other places to clean up on regards to dropping 3.8 but probably better for separate PRs IMO.

.github/workflows/testing-and-deployment.yml
```diff
@@ -150,9 +150,6 @@ jobs:
       matrix:
         include:
           # numeric numpy versions are ~= conditions, e.g. "1.23" means "numpy~=1.23.0"
-          - python-version: "3.8"
-            vtk-version: "9.0.3" # Requires numpy~=1.23
```
MatthewFlamm, line 154: Are we also dropping support for this vtk version?  It looks like can be installed with putting  Python 3.9, so we should probably keep it in the matrix somehow. Either add a second Python 3.9 or reorder the vtk-versions in the matrix.

banesullivan: We really ought to continue testing VTK 9.0.3. Good call on having two Py3.9 entries for VTK 9.0.3 and VTK 9.1

tkoyama010:
pre-commit.ci autofix

MatthewFlamm:
> It might take some more digging, but the current test failure is for this test which is currently skipped for python version <3.9. It seems likely it should have been skipped for vtk<9.1?
> 
> https://github.com/pyvista/pyvista/blob/cde0ae04e4a99506679e1372c6133b6649e8eda3/tests/core/test_utilities.py#L1011-L1017

I'm no longer sure.  This test doesn't interact with vtk as far as I can tell.  It was added in https://github.com/pyvista/pyvista/pull/6346/.  Maybe @user27182 can shed light here?

user27182:
> > It might take some more digging, but the current test failure is for this test which is currently skipped for python version <3.9. It seems likely it should have been skipped for vtk<9.1?

I'm not seeing the current test failure for this, but I believe the differences in test outputs here may stem from using different numpy versions. Looks like NumPy 1.23 is used with the vtk 9.0.3 test suite. So, we can skip for vtk<9.1 as suggested, or perhaps more directly skip numpy < 1.26.

MatthewFlamm:
> I'm not seeing the current test failure for this

The PR was changed to skip the test based on vtk version.

> perhaps more directly skip numpy < 1.26.

Ah yes, something like this is the right fix.  Thanks!

tkoyama010:
pre-commit.ci autofix

.github/workflows/testing-and-deployment.yml
tkoyama010: Now this is resolved thanks all for your help!

MatthewFlamm (approved):

MatthewFlamm:
I presume all previous changes to merging #6713 in this PR were ruff rules changes.  Merging #6713 in here complicates the history in main.  We will now be mixing the dropping of python 3.8 (and forced ruff rules changes) with manual changes.  I think we should avoid more of this in this PR.  But the changes look fine to me.

An aside is that we aren't type checking on multiple versions, so it isn't clear to me that type checking will work on all python versions.  This is related to #5611.

MatthewFlamm (approved):

user27182:
> I presume all previous changes to merging #6713 in this PR were ruff rules changes. 

This is not quite true as a README and workflow file were also changed here, and possibly other manual changes related to the project config.

> Merging #6713 in here complicates the history in main. We will now be mixing the dropping of python 3.8 (and forced ruff rules changes) with manual changes. I think we should avoid more of this in this PR. But the changes look fine to me.

Agreed, fewer changes is best. It would have been better to rebase #6713 to main and merge it after this PR.

> An aside is that we aren't type checking on multiple versions, so it isn't clear to me that type checking will work on all python versions. This is related to #5611.

Type checking does not pass for me locally on MacOS with python 3.9 and the current dev branch. This has been the case for many weeks now. mypy with pre-commit routinely complains with 4-8 errors for me (depending on the files being modified), and I just have to ignore them and commit anyway.

MatthewFlamm:
> Type checking does not pass for me locally on MacOS with python 3.9 and the current dev branch. This has been the case for many weeks now. mypy with pre-commit routinely complains with 4-8 errors for me (depending on the files being modified), and I just have to ignore them and commit anyway.

I am having this problem too.  This mypy checking is becoming a royal pain recently.

banesullivan (approved):
LGTM! While I'm hesitant to drop support for versions of Python that VTK may still support, Python 3.8 users can continue to use a recent release which has been quite stable for some time

Issue #6705, opened 2024-10-16
arial front is not arial


chapochn:
### Describe the bug, what's wrong, and what you expected.

When I put the labels, with the font 'arial', the font that shows up is not arial! Also looking through the documentation, it is not arial:
https://docs.pyvista.org/user-guide/

### Steps to reproduce the bug.

```
import pyvista as pv
import numpy as np
import matplotlib.pyplot as plt

pl = pv.Plotter()

n = 10
points = np.zeros((n, 3))
points[:, 0] = -np.arange(n)

pl = pv.Plotter()
labels = [str(i) for i in range(n)]
pl.add_point_labels(points ,labels, point_color='k', point_size=30, always_visible=True, font_family='arial')

pl.render()
pl.show()
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Wed Oct 16 16:24:32 2024 EDT

                OS : Darwin (macOS 14.6.1)
            CPU(s) : 12
           Machine : arm64
      Architecture : 64bit
               RAM : 32.0 GiB
       Environment : Jupyter
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M2 Pro
       GPU Version : 4.1 Metal - 88.1
  MathText Support : True

  Python 3.12.7 | packaged by conda-forge | (main, Oct  4 2024, 15:57:01)
  [Clang 17.0.6 ]

           pyvista : 0.44.1
               vtk : 9.3.1
             numpy : 1.26.4
        matplotlib : 3.9.2
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 11.0.0
           imageio : 2.36.0
         pyvistaqt : 0.11.1
             PyQt5 : 5.15.9
           IPython : 8.28.0
        ipywidgets : 8.1.5
             scipy : 1.14.1
              tqdm : 4.66.5
        jupyterlab : 4.2.5
             trame : 3.6.5
      trame_client : 3.3.2
      trame_server : 3.2.3
         trame_vtk : 2.8.10
     trame_vuetify : 2.7.1
      nest_asyncio : 1.6.0
--------------------------------------------------------------------------------
```

### Screenshots

This is the output vs arial

![Screenshot 2024-10-16 at 4 16 19 PM](https://github.com/user-attachments/assets/463e1126-aa69-4f31-8cd9-e42b6346f97d)
[...]

PR #4639, opened 2023-07-09
Add `Text` class for `font_file` as a parameter to multiple language

tkoyama010:
### Overview

resolves #4548

```python
import os
import pyvista


plotter = pyvista.Plotter()
font_file = "fonts/Mplus2-Regular.ttf"
plotter.add_text("左上", position='upper_left', font_size=25, color='blue', font_file=font_file)
plotter.add_text("中央", position=(0.5, 0.5), viewport=True, orientation=-90, font_file=font_file)
plotter.show()
```

add_text_font_file

### Details

- [x] Corresponds to a local path.
- [x] Support for errors when a file does not exist

PR #6707, opened 2024-10-16
added font_file property to add_point_labels

chapochn:
### Overview

added the option to specify the font_file for the function add_point_labels

This is a similar change to:
https://github.com/pyvista/pyvista/pull/4639
which was done for the add_text function

gives a workaround (but maybe does not fully resolve the issue) for this:
https://github.com/pyvista/pyvista/issues/6705

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -5448,6 +5448,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
         font_size=None,
         text_color=None,
         font_family=None,
+        font_file=None,
         shadow=False,
         show_points=True,
         point_color=None,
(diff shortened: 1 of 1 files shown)


tkoyama010 (approved):
@MatthewFlamm Should we add a test for this change? I don't have a strong opinion about it, but I'd like to hear what other people think.

MatthewFlamm (approved):
We have similar tests for TextProperty

https://github.com/pyvista/pyvista/blob/6624cdedf0416d7dfb4c300328a43f2f98f07c3a/tests/plotting/test_text.py#L185-L190

And another with plotting 

https://github.com/pyvista/pyvista/blob/6624cdedf0416d7dfb4c300328a43f2f98f07c3a/tests/plotting/test_plotting.py#L2983-L2995

I don't see any tests for modifying these properties inside `add_point_labels`.  I think it should be fine to continue this way.  [Added: What I mean is that this PR is fine as is.]  If we want to test, we should add a test that modifies all of them and checks the TextProperty somehow.

chapochn:
Do I need to write about this change anywhere in some change file?

MatthewFlamm:
I can let @tkoyama010 chime in too, but I'm fairly sure that the release notes are auto-generated via label.  So this will automatically get included in the "New Features" section of the next release, e.g. https://github.com/pyvista/pyvista/releases/tag/v0.44.0.  And of course, if it somehow isn't, please drop a line somewhere and we can fix it.

PR #6714, opened 2024-10-23
Update `cache-apt-pkgs-action` to `v1.4.2` for Linux Core testing workflow

user27182:
### Overview

The linux unit testing workflows are generating this warning, e.g.:
> Linux Unit Testing (3.10, 9.2.2, latest)
> The following actions use a deprecated Node.js version and will be forced to run on node20: actions/cache@v3.

I believe this may be caused by `awalsh128/cache-apt-pkgs-action@v1.1.3`, which uses `actions/cache@v3`:
https://github.com/awalsh128/cache-apt-pkgs-action/blob/d93c95e39c0450c48c3e0c2b2ad2993f33da786a/action.yml#L48

This PR updates this action to the latest version to fix the warning.

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -227,7 +227,7 @@ jobs:
       - uses: awalsh128/cache-apt-pkgs-action@v1.4.2
         with:
           packages: libgl1-mesa-glx xvfb
-          version: 3.0
+          version: 4.0
 
       - name: Plotting Testing (uses GL)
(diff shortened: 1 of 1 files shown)


user27182:
For reference, it is necessary to also bump the workflow `version` to make the update work.
See https://github.com/awalsh128/cache-apt-pkgs-action/issues/56

user27182:
Looks like some more investigation is needed to update this action. Marking as draft, but if anyone else wants to investigate this further...

user27182:
I'm not seeing these warnings anymore, so I guess this action is now "forced to run on node20". It would probably still be good to update this action though

PR #6724, opened 2024-10-24
Add `validate_rotation` and use it to validate rotation inputs

user27182:
### Overview

Rotation matrices are not currently validated by `transformations.rotation` and `prop3d.rotation_from` to ensure inputs are actually rotations (e.g. they can be any 3x3 matrix). This PR adds a new validation method for checking this.


pyvista/core/_validation/validate.py
```diff
@@ -472,6 +473,77 @@ def validate_axes(
...
+    By default, left-handed rotations (which include reflections) are allowed.
+
+    >>> rotation *= -1  # Add reflections
+    >>> _validation.validate_rotation(rotation)
+    array([[-1., -0., -0.],
+           [-0., -1., -0.],
+           [-0., -0., -1.]])
+
+    """
+    check_contains(item=must_have_handedness, container=['right', 'left', None], name=name)
```
user27182, line 528:
```suggestion
    check_contains(item=must_have_handedness, container=['right', 'left', None], name='must_have_handedness')
```

banesullivan (approved):

user27182:
Might need to add a tolerance parameter to `validate_rotation`, this test is giving a false negative:
https://github.com/pyvista/pyvista/actions/runs/12215994150/job/34078629983?pr=6946#step:6:6983
```
FAILED tests/core/test_dataset_filters.py::test_bounding_box_return_meta[False-True] - ValueError: Rotation is not valid. Its inverse must equal its transpose.
```

PR #6725, opened 2024-10-24
Remove redundant `transformations.rotation` function

user27182:
### Overview

This function was originally added in #6354 in version 0.45dev and has not yet been released.

It has been made redundant by the new `Transform` class with https://github.com/pyvista/pyvista/pull/6482 and https://github.com/pyvista/pyvista/pull/6501 and there is no longer test coverage for it either.


tkoyama010 (approved):
LGTM

PR #6726, opened 2024-10-24
Simplify code in `ImageData.__init__` and remove `_from_specs` method

user27182:
### Overview

Remove `ImageData._from_specs` and move its code directly into `__init__`. The `_from_specs` method appears outdated and is not used elsewhere. It makes mention of init directly from numpy arrays (this is no longer allowed) and makes reference to uniform grid (it has been renamed to ImageData)


tkoyama010 (dismissed):
LGTM

pyvista/core/grid.py
```diff
@@ -608,8 +608,11 @@ def __init__(
                     '    ...     origin=(10, 35, 50),\n'
                     '    ... )\n',
                 )
-        elif dimensions is not None:
-            self._from_specs(dimensions, spacing, origin)
+        else:
+            if dimensions is not None:
+                self.dimensions = dimensions
+            self.origin = origin
+            self.spacing = spacing
```
user27182, line 615: Note that the logic here has also changed a bit. Previously, there was a bug with init where `spacing` and `origin` were only set if `dimensions` was also set. Now these values can all be initialized independently. Tests have been added for this in https://github.com/pyvista/pyvista/pull/6726/commits/7c3c3174449fe88341a22d41b6c100a3fedc5e45

user27182:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM

PR #6727, opened 2024-10-24
Add `direction_matrix` kwarg to `ImageData.__init__`

user27182:
### Overview

Follow-up to https://github.com/pyvista/pyvista/pull/6612. `direction_matrix` is used alongside `spacing` and `origin` to define the ImageData orientation and so I think it makes sense to include this in `__init__`.


user27182:
This should be good to go but should wait until #6726 is merged first

pyvista/core/grid.py
```diff
@@ -583,6 +583,7 @@ def __init__(
         dimensions=None,
         spacing=(1.0, 1.0, 1.0),
         origin=(0.0, 0.0, 0.0),
+        direction_matrix=None,
         deep=False,
```
user27182, line 587: This is technically a breaking change, might be best to put the new keyword at the end.
```suggestion
        deep=False,
        direction_matrix=None,
```

tkoyama010: +1 not breaking change.

pyvista/core/grid.py
```diff
@@ -584,6 +584,7 @@ def __init__(
         spacing=(1.0, 1.0, 1.0),
         origin=(0.0, 0.0, 0.0),
         deep=False,
+        direction_matrix=None,
```
tkoyama010, line 587: We should add explanations for each parameter in the docstring :)

tkoyama010 (changes requested):

pyvista/core/grid.py
user27182: whoops! 😅 

Odd that there are no complaints from pre-commit about this (numpydoc or ruff should be linting this kind of thing...)

tkoyama010: ref #6755

user27182: > Let's verify if #6755 can catch this error after merging it.

Still not caught by pre-commit, unfortunately :/

user27182: This might be the issue. This line should be removed:
https://github.com/pyvista/pyvista/blob/ea732e95990b1f3ad56ca1ee3e514142f6d66b59/pyproject.toml#L309

user27182: Hmm... this isn't it either. That doesn't check for missing members. Plus we define init params in the class docstring, not the __init__ docstring.

banesullivan (dismissed):

banesullivan:
@user27182, would you please handle the missing type annotation here and update this branch?

tkoyama010 (approved):
LGTM

pyvista/core/grid.py
```diff
@@ -15,6 +15,7 @@
 
 if TYPE_CHECKING:
```
tkoyama010, line 16:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

tkoyama010 (changes requested):

tkoyama010 (approved):

PR #6728, opened 2024-10-24
Increase flaky test tries for `voxelize_binary_mask` sampling test

user27182:
### Overview

This test is known to be flaky but the default number of tries (3) doesn't quite seem to be enough, and sometimes still produces false negatives. Change this to 5 tries as a quick fix.

### DIFF
--- a/tests/core/test_polydata_filters.py
+++ b/tests/core/test_polydata_filters.py
@@ -143,7 +143,7 @@ def test_voxelize_binary_mask_auto_spacing(ant):
 
 # This test is flaky because of random sampling that cannot be controlled.
 # Sometimes the sampling produces the same output.
-@flaky_test
+@flaky_test(times=5)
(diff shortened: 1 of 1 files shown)


tests/core/test_polydata_filters.py
```diff
@@ -143,7 +143,7 @@ def test_voxelize_binary_mask_auto_spacing(ant):
 
 # This test is flaky because of random sampling that cannot be controlled.
 # Sometimes the sampling produces the same output.
-@flaky_test
+@flaky_test(times=5)
```
tkoyama010, line 146:
```suggestion
# https://github.com/pyvista/pyvista/pull/6728
@flaky_test(times=5)
```

tkoyama010 (changes requested):

tkoyama010 (approved):

Issue #6279, opened 2024-06-19
Roadmap to `strict` typing with Mypy


user27182:
### Describe what maintenance you would like added.

Type annotations make code easier to read and help prevent bugs. Having 100% annotations would also make it easy to switch to using `autodoc-typehints` (see #5665) and simplify the documentation. To this extent, I propose working towards using the `--strict` option with `mypy`. This issue can be used to track progress towards this goal. 

[This section](https://mypy.readthedocs.io/en/stable/existing_code.html#introduce-stricter-options) in the Mypy docs shows how this can be worked up to by enabling some options one at a time. The options for these are shown in "Add config options" below.

In addition there are some other outstanding issues with running `mypy` CI checks for the project. I believe these should be addressed first before enabling more options.

## Fix outstanding issues
- [ ] Use a single, common `mypy` environment for CI checks (we currently have two configs, one  in `pyproject.toml` and another in `pre-commit-config.yaml`), see  [#6257 (commet)](https://github.com/pyvista/pyvista/issues/6257#issuecomment-2168742132)
- [ ] Make sure `mypy` imports project files correctly (e.g. resolve #5611)
- [ ] Resolve known existing type errors, e.g. resolve #5563
- [ ] Add dependencies to config, e.g. #6275 

### Start off with these

- [ ] `warn_unused_configs = True`
- [ ] `warn_redundant_casts = True`
- [ ] `warn_unused_ignores = True`

### Getting these passing should be easy

- [ ] `strict_equality = True`
- [ ] `strict_concatenate = True`

### Strongly recommend enabling this one as soon as you can

- [ ] `check_untyped_defs = True`

### These shouldn't be too much additional work, but may be tricky to get passing if you use a lot of untyped libraries

- [ ] `disallow_subclassing_any = True`
- [ ] `disallow_untyped_decorators = True`
- [ ] `disallow_any_generics = True`

### These next few are various gradations of forcing use of type annotations

- [ ] `disallow_untyped_calls = True`
- [ ] `disallow_incomplete_defs = True`
- [ ] `disallow_untyped_defs = True`

### This one isn't too hard to get passing, but return on investment is lower

- [ ] `no_implicit_reexport = True`

### This one can be tricky to get passing if you use a lot of untyped libraries

- [ ] `warn_return_any = True`

user27182:
[...]

PR #6738, opened 2024-10-25
Add static type checking workflow with `mypy`

user27182:
### Overview

Part of #6279.

Run `mypy` checks using a standard testing environment. We need to use a custom workflow instead of pre-commit.ci in part because of the 250MB pre-commit.ci limitations, see https://github.com/pyvista/pyvista/pull/6368

We also should be testing with different python versions, see https://github.com/pyvista/pyvista/pull/6701#issuecomment-2432395713


user27182:
Running `mypy pyvista` (with the current `mypy` config in `pyproject.toml`) is successful on python 3.12 with no errors when the 19 fixes in #6739 are applied. See this test: https://github.com/pyvista/pyvista/actions/runs/11524077765/job/32083456680#step:8:1

But, as expected, this fails with 6 errors with python 3.9:
https://github.com/pyvista/pyvista/actions/runs/11524077765/job/32083455705#step:8:1

user27182:
The mypy checks are currently passing on python 3.12 with all project test requirements installed, including the latest VTK. See https://github.com/pyvista/pyvista/actions/runs/11525834793/job/32088870009#step:8:12

So, assuming we simply ignore all the type errors in this PR, we can begin to use this workflow for type checking instead of the pre-commit mypy mirror.

But, this config directly clashes with pre-commit since pre-commit thinks the ignores are not needed. Not sure how we should transition away from the pre-commit `mypy` mirror hook.... It would be nice to still use pre-commit but maybe use the local env instead a separate isolated venv (i.e. do not use the auto-managed pre-commit venv) ? Any thoughts on this @MatthewFlamm ?

user27182:
> But, this config directly clashes with pre-commit since pre-commit thinks the ignores are not needed. Not sure how we should transition away from the pre-commit `mypy` mirror hook.... It would be nice to still use pre-commit but maybe use the local env instead a separate isolated venv (i.e. do not use the auto-managed pre-commit venv) ?

It might be tricky to do everything all at once. I think we should drop it from `pre-commit` entirely for now, and consider adding back to pre-commit as a local env later. See proposal in #6741

MatthewFlamm:
> > But, this config directly clashes with pre-commit since pre-commit thinks the ignores are not needed. Not sure how we should transition away from the pre-commit `mypy` mirror hook.... It would be nice to still use pre-commit but maybe use the local env instead a separate isolated venv (i.e. do not use the auto-managed pre-commit venv) ?
> 
> It might be tricky to do everything all at once. I think we should drop it from `pre-commit` entirely for now, and consider adding back to pre-commit as a local env later. See proposal in #6741

I'm +1 for adding this and dropping pre-commit.  If we add pre-commit later, it is only to make dev life easier.  So we can assess pros and cons at that time.  There are a few pitfalls with using mypy in the local environment too, see https://github.com/python/mypy/issues/13916.

pyvista/ext/coverage.py
```diff
@@ -38,12 +38,12 @@
 
 
 # utility
-def write_header(f: IO, text: str, char: str = '-') -> None:
```
MatthewFlamm, line 41: This file was excluded from pre-commit checks.  Let's exclude it from local too, unless you think there is value add here.  It looks like a lot of ignores needed to be added.

pyvista/ext/plot_directive.py
```diff
@@ -181,7 +181,7 @@ class PlotDirective(Directive):
     required_arguments = 0
     optional_arguments = 2
     final_argument_whitespace = False
-    option_spec: ClassVar[dict[str, Callable]] = {
```
MatthewFlamm, line 184: Same comment about this file being originally excluded.

requirements_test.txt
```diff
@@ -31,4 +31,5 @@ trame>=2.5.2,<3.6.6
 trame-vtk>=2.5.8,<2.8.11
 trame-vuetify>=2.3.1,<2.7.2
 trimesh<4.5.0
+types-docutils
```
MatthewFlamm, line 34: We pin all the versions in this file.

.github/workflows/type-checking.yml
```diff
@@ -0,0 +1,62 @@
...
+  mypy:
+    name: Static Type Checking
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+
+      matrix:
+        include:
+          # TODO: Add testing for older python and/or numpy versions
+          # - python-version: "3.9"
```
MatthewFlamm, line 19: While I think that this is a great goal, and I might have suggested it.  Python's type checking system is not the best set up to support this.  For example, if we need different typing for vtk 9.1 and 9.4 because the function signature changes in vtk, I don't think type checking can handle this dependency level detail, or even if it can, creates even more complexity.  I like the idea of this PR to implement a 'like-for-like' switch first, which it is.

.pre-commit-config.yaml
```diff
@@ -47,7 +47,6 @@ repos:
             --disallow-any-generics,
             --pretty,
             --show-error-context,
-            --warn-unused-ignores,
```
MatthewFlamm, line 50: Why this change here, if it will be wholesale removed?  Seems like this will cause merge conflicts.

MatthewFlamm: This line is in the project toml too

https://github.com/pyvista/pyvista/blob/b516e91275eda3260c20a2e1cf0438a7997a6fce/pyproject.toml#L143-L153

requirements_test.txt
MatthewFlamm: I'm also puzzled why we need this one as it isn't in pre-commit.  Is it due to not excluding the same files?  

Also do we know whether we need `mypy-extensions`?  Unless we know it is likely safer to keep this as like-for-like as possible.
```yaml
# list of dependencies in pre-commit mypy
            "mypy-extensions==1.0.0",
            "toml==0.10.2",
            "numpy>=2.0",
            "npt-promote==0.1",
            "matplotlib<3.9.1",
```

pyvista/ext/coverage.py
MatthewFlamm: Since the CI check explicitly passes `pyvista` folder, we only will need to exclude the subfolder `ext` from this list to make it the same as before.

https://github.com/pyvista/pyvista/blob/b516e91275eda3260c20a2e1cf0438a7997a6fce/.pre-commit-config.yaml#L55

MatthewFlamm (commented):
This is great work!  I tried a few times to do this, but always left defeated with so many errors to track down from vtk.

requirements_test.txt
user27182: This might have something to do with how `mypy` follows imports. With this workflow we run `mypy pyvista` and let `mypy` find everything it needs. But with pre-commit `mypy` is run separately on a list of files, not a directory. (Note, pre-commit will also run on files from `doc` and `tests` as well, which are technically excluded/not covered by the  new workflow).

I think including `types-docutils` is necessary here since there are a couple of errors raised otherwise.

The ignores needed for this to work are included with #6739, which is passing pre-commit CI. So, if we want this to be like-for-like as much as possible, we can create a separate PR first where we add `types-docutils` to the pre-commit dependencies along with the few ignores needed.

This example also highlights why this new workflow is needed: pre-commit does not appear to be catching errors like this.

pyvista/ext/coverage.py
user27182: Ok. I'll look into file-level excludes. Note that the current workflow excludes tests, axamples, and examples_trame by default since it's only checking the `pyvista` module/directory (i.e. it runs `mypy pyvista`, not `mypy examples`, etc.)

.github/workflows/type-checking.yml
user27182: Yea... it will be challenging to support across NumPy and VTK versions. I'm thinking we could add a mechanism to simply ignore this type check conditionally, e.g. if `TYPE_CHECKING and pv.vtk_version > (9.3)`. This is more a nice-to-have though. TBD if we will support this in the future.

.pre-commit-config.yaml
user27182: Yes this should be removed. It was an attempt to get CI to pass, since this is the only error raised by the pre-commit config. But, it seems pre-commit.ci is not catching this removal and is still warning about unused type ignores for some reason.

requirements_test.txt
user27182: Figured it out. `types-docutils` this is only needed for `pyvista/ext` which should otherwise be excluded. See https://github.com/pyvista/pyvista/actions/runs/11533261412/job/32105976694?pr=6738#step:8:1

So we can remove this dependency.

.github/workflows/type-checking.yml
user27182: Maybe we should remove the TODO and the use of `matrix`? I'm thinking that in the future, if we really want to support this properly, we should probably test using the same matrix used for the unit testing.  And in that case, to avoid having to duplicate the matrix, it might be better to run `mypy` as part of that workflow instead. But that's for a later conversion

user27182: I removed the matrix in [67015d7](https://github.com/pyvista/pyvista/pull/6738/commits/67015d74d3754a6abd9ed313844229efdf0aa7d9)

user27182:
All comments/issues should be addressed now

pyproject.toml
```diff
@@ -151,6 +151,7 @@ plugins = ['numpy.typing.mypy_plugin','npt_promote']
 enable_error_code = [
     "ignore-without-code",
 ]
+exclude = ['pyvista/ext/']
```
user27182, line 154: Include packages here. This way we can simply run `mypy`  instead of `mypy pyvista`
Should probably include `doc` here too since coverage for that was added in https://github.com/pyvista/pyvista/pull/5922
```suggestion
exclude = ['pyvista/ext/']
packages = ['pyvista']
```

user27182:
Looks like #6741 has been approved and will be merged soon. Since the workflow from this PR is intended as a replacement, it might be a good idea to review it soon so that we don't merge too many other PRs in between as we will temporarily lose type-checking CI checks until this PR is merged.

pyvista/ext/plot_directive.py
```diff
@@ -344,7 +344,7 @@ def _split_code_at_show(text):
 
 # the context of the plot for all directives specified with the
 # :context: option
-plot_context = {}
+plot_context = {}  # type: ignore[var-annotated]
```
MatthewFlamm, line 347: Is this needed now?

MatthewFlamm (commented):
Only have one comment, and I think this should be merged quickly to avoid typing failure creep.

pyvista/ext/plot_directive.py
user27182: No, we can remove this, good catch
```suggestion
plot_context = {}
```

MatthewFlamm (approved):
Thanks for doing this!

PR #6739, opened 2024-10-25
Fix `dtype` type error in `voxelize_binary_mask`

user27182:
### Overview

Follow-up to #6738. If we want to use the `mypy` config there, the errors in this PR need to be resolved/ignored first.


tkoyama010 (approved):
LGTM

PR #6740, opened 2024-10-25
Ignore `mypy` errors caused by including VTK as a dependency

user27182:
### Overview

Follow-up to https://github.com/pyvista/pyvista/pull/6738. If we want to use the mypy config there, the errors in this PR need to be resolved/ignored first.

This PR is similar to #6739, except the errors in this PR are exclusively caused by including VTK as a dependency.

See all 219 errors ignored by this PR:
<details>

```
pyvista/core/utilities/cell_type_helper.py:8: error: Name "vtkCommonDataModel" already defined (by an import)  [no-redef]
pyvista/plotting/_vtk_gl.py:18: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]
pyvista/core/_vtk_core.py:32: error: Module "vtkmodules.vtkFiltersExtraction" has no attribute "vtkExtractEdges"; maybe "vtkExtractGrid"?  [attr-defined]
pyvista/core/_vtk_core.py:32: error: Name "vtkExtractEdges" already defined (possibly by an import)  [no-redef]
pyvista/core/_vtk_core.py:39: error: Module "vtkmodules.vtkFiltersGeneral" has no attribute "vtkCellTreeLocator"  [attr-defined]
pyvista/core/_vtk_core.py:39: error: Name "vtkCellTreeLocator" already defined (possibly by an import)  [no-redef]
pyvista/core/_vtk_core.py:82: error: Module "vtkmodules.vtkCommonCore" has no attribute "buffer_shared"  [attr-defined]
pyvista/plotting/_vtk.py:69: error: Cannot assign to a type  [misc]
pyvista/plotting/_vtk.py:69: error: Incompatible types in assignment (expression has type "None", variable has type "type[vtkHardwarePicker]")  [assignment]
pyvista/core/celltype.py: note: In member "__new__" of class "CellType":
pyvista/core/celltype.py:235: error: Argument 1 to "_generate_linear_badge" has incompatible type "int"; expected "bool"  [arg-type]
pyvista/core/celltype.py:236: error: Argument 1 to "_generate_primary_badge" has incompatible type "int"; expected "bool"  [arg-type]
pyvista/core/celltype.py:240: error: Argument 1 to "_generate_points_badge" has incompatible type "Literal['variable', 'n/a'] | int"; expected "int"  [arg-type]
pyvista/core/celltype.py:243: error: Argument 1 to "_generate_edges_badge" has incompatible type "Literal['variable', 'n/a'] | int"; expected "int"  [arg-type]
[...]

PR #6741, opened 2024-10-25
Remove `mypy` from pre-commit

user27182:
### Overview

Part of #6279. We have two mypy configs and the pre-commit version is causing problems since we can't install the necessary dependencies (e.g. VTK), see https://github.com/pyvista/pyvista/issues/6279#issuecomment-2181326986 and #6368.

I propose:
1. Dropping `mypy` from `pre-commit` entirely (this PR)
2. Use a GitHub workflow for running `mypy` instead (#6738)
3. ~Since the workflow generates a number of dependency-related errors, we can ignore these errors first (for now) i.e. merge #6739 and #6740, in order to separate config changes from the ignores.~

This should allow us to complete the changeover now and prevent more type errors from being introduced. This means devs will need to run `mypy pyvista` manually for local checks.

In the future, it would be nice to:

1. Revisit ~#6740~ #6738 and fix some or many of the type errors that were ignored.
2. Add `mypy` back as a `pre-commit` hook, but use a local environment instead of one managed by `pre-commit`


user27182:
The new type checking job in #6738 is passing, so the change-over from `pre-commit` to using the new workflow is ready to go. The suggested merge order for this is: merge this PR first, then #6739, then #6740, then #6738.

akaszynski (approved):
LGTM. Thanks for your work!

user27182:
> 2. Add `mypy` back as a `pre-commit` hook, but use a local environment instead of one managed by `pre-commit`

Instead of using a local environment, we can keep the pre-commit venv for mypy if we can somehow point to the `requirements_test.txt` file in the pre-commit config for the mypy dependencies instead of listing them explicitly.


Then, we can prevent mypy from running in pre-commit.ci (which can't run due to the 250MB dependency limit) with
```
ci:
    skip: [mypy]
```

That way we can continue to use `pre-commit` locally but run our own workflow for CI, and both should have the same dependencies and mypy config

MatthewFlamm:
I agree with the plan, I would order the merging this way:

1) merge any needed fixes
2) drop  pre-commit
3) immediately after 2 add the mypy CI action

MatthewFlamm:
If we can't merge the fix PRs before dropping pre-commit, which is likely, then we need to follow your order.  It will take a concerted effort to not merge in any other PR after dropping pre-commit until the sequence is complete. 

Or, we put all the fixes into the PR that adds the CI job.

The complexity is driven by the squash merge model, otherwise we could have separate commits in one PR.

user27182:
> If we can't merge the fix PRs before dropping pre-commit, which is likely, then we need to follow your order. It will take a concerted effort to not merge in any other PR after dropping pre-commit until the sequence is complete.
> 
> Or, we put all the fixes into the PR that adds the CI job.
> 
> The complexity is driven by the squash merge model, otherwise we could have separate commits in one PR.

I think the main block preventing us from merging the new workflow before dropping it from pre-commit is that the pre-commit config complains about unused `type: ignore`s. So, we can probably remove this line first (as a new PR):
https://github.com/pyvista/pyvista/blob/636a578208e9600ecd280f3681fdaa90011b341c/.pre-commit-config.yaml#L50

and then we should be able to merge the VTK ignores and new workflow without issue since the CI should pass without that line. And then drop `mypy` from pre-commit. That way we'll have uninterrupted mypy coverage without any need to worry about merging other PRs in between.

user27182:
> I agree with the plan, I would order the merging this way:
> 
> 1. merge any needed fixes
> 2. drop  pre-commit
> 3. immediately after 2 add the mypy CI action

Ignore my previous response, let's just do this. Specifically, this would look like:

1. Merge #6739. After review, it now only has a single type error fix. And pre-commit.ci is passing. So we can merge that first without issue.
2. Merge this PR
3. Merge #6738.

We can then close #6740 (these changes are included as part of #6738). We can also close #6742 (not needed).

user27182:
I was able to get pre-commit to work properly with `mypy` in #6744. That PR is currently an all-in-one solution: it splits requirements into `test` and `typing`, adds the project dependencies (e.g. vtk), adds the type ignores, and adds the new workflow. New tests are also added to ensure that the `mypy` pre-commit dependencies are always in sync with those in `requirements_typing.txt`.

So, we don't really need to drop `mypy` from pre-commit after all.

I'll leave it up to reviewers how or if we want to break up that PR. We could, for example, close all these other PRs and only merge #6744. Or, we continue as above, and merge the PRs separately in stages.

tkoyama010 (approved):
For the sake of understanding, I will merge these separately.

PR #6744, opened 2024-10-27
Add custom `pre-commit` `mypy` hook

user27182:
### Overview

Add custom `mypy` hook to replace the one removed in #6741.

Instead of using https://github.com/pre-commit/mirrors-mypy, a customized fork (https://github.com/pyvista/pre-commit-mypy) is used instead. 

Parametrized tests are also included to ensure the hook's virtual env dependencies are synchronized with new `typing` dependencies in pyproject.toml. This ensures that the pre-commit dependencies match those used in the new workflow in #6738.


.pre-commit-config.yaml
```diff
@@ -38,29 +39,22 @@ repos:
       - id: numpydoc-validation
         files: ^pyvista/
 
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.11.2
+  - repo: https://github.com/pyvista/pre-commit-mypy
+    rev: v0.1dev6
```
user27182, line 43: A `v0.1` release will be need to be created once the review is complete.

The main thing this fork does is:

1. Remove the "mirror" part. The repo's version is no longer sync'd with the `mypy` version. Instead, we control the `mypy` version via `additional_dependencies`, which is controlled here in the `pyvista` repo instead.
2. Remove the [default args](https://github.com/pre-commit/mirrors-mypy/blob/main/.pre-commit-hooks.yaml#L7): `["--ignore-missing-imports", "--scripts-are-modules"]` so that we have full control over what is passed to `mypy`
4. Use `pass_filenames: false`, as recommended by the `mypy` devs here: https://github.com/python/mypy/issues/13916 (see https://github.com/pyvista/pyvista/pull/6738#issuecomment-2439471798 for context)

user27182:
Ready for review, see also the new pre-commit hook used by this PR https://github.com/pyvista/pre-commit-mypy

Not sure if we want to take a staged approach for merging these changes or merge this PR as-is, see https://github.com/pyvista/pyvista/pull/6741#issuecomment-2443245396.

.pre-commit-config.yaml
```diff
@@ -38,6 +39,25 @@ repos:
...
 
+  - repo: https://github.com/pyvista/pre-commit-mypy
+    rev: v0.1.0
+    hooks:
+      - id: mypy
+        args: [] # Args should be blank. Set args via 'pyproject.toml' instead.
+        additional_dependencies:
+          # These should match `pyvista[typing]` dependencies in 'pyproject.toml'
+          [
+            "vtk<9.4.0",
```
MatthewFlamm, line 50: How will the versions here be kept up to date with those in pyproject?  I think dependabot will do one, and precommit.ci will do the other.  These will always be out of sync.

We can also consider the second suggestion in https://github.com/python/mypy/issues/13916 to

>Use language: system (and maybe a helper script) to run something from outside a pre-commit environment

user27182: There are new parametrized tests included with this PR (see tests/typing). These tests compare dependencies in pyproject.toml with those in pre-commit-config.yaml. The tests fail if the dependencies are out of sync. 

So, when dependabot bumps a dependency in pyproject.toml, there will initially be a test failure, and the pre-commit config will need to be manually bumped to resolve the failure. This should keep everything in sync.

user27182: > We can also consider the second suggestion in https://github.com/python/mypy/issues/13916 to
> 
> 
> 
> >Use language: system (and maybe a helper script) to run something from outside a pre-commit environment

I considered this option, but I think it's good to keep typing dependencies separate from other dependencies (i.e. test, docs, and dev). And so the problem with using system is that now devs will need to make sure they do `pip install -e .[typing]` before running pre-commit or else they risk having the wrong dependencies installed. 

For example, I found that doing `pip install -e .[dev]` will install numpy 1.26, because some dependencies (e.g. osmnx) have numpy pinned less than 2.0. This is not compatible with the typing environment where we want numpy>=2.0. 

So, to avoid this kind of complexity (and possible frustration for devs), I like the idea of letting pre-commit configure the venv.

user27182:
Note that #6746 is included as part of this PR. If preferred that one may be reviewed / merged first.

user27182:
Example of new tests from this PR, one for each `pyvista[typing]` dependency
https://github.com/pyvista/pyvista/actions/runs/11781538183/job/32814587650?pr=6744#step:6:6657
> tests/typing/test_mypy_precommit_dependencies.py::test_dependency[matplotlib] PASSED [ 98%]
tests/typing/test_mypy_precommit_dependencies.py::test_dependency[mypy] PASSED [ 98%]
tests/typing/test_mypy_precommit_dependencies.py::test_dependency[npt-promote] PASSED [ 98%]
tests/typing/test_mypy_precommit_dependencies.py::test_dependency[numpy] PASSED [ 98%]
tests/typing/test_mypy_precommit_dependencies.py::test_dependency[pillow] PASSED [ 98%]
tests/typing/test_mypy_precommit_dependencies.py::test_dependency[pooch] PASSED [ 98%]
tests/typing/test_mypy_precommit_dependencies.py::test_dependency[scooby] PASSED [ 98%]
tests/typing/test_mypy_precommit_dependencies.py::test_dependency[typing-extensions] PASSED [ 98%]
tests/typing/test_mypy_precommit_dependencies.py::test_dependency[vtk] PASSED [ 98%]

user27182:
And here is an example failure caused by simulating a dependency update in https://github.com/pyvista/pyvista/pull/6744/commits/77b17119b29f6571fe45388cb08d387d15d39ddd (e.g. from a dependabot bump)
https://github.com/pyvista/pyvista/actions/runs/11781808762/job/32815455151?pr=6744#step:6:6737
```
_________________________ test_dependency[matplotlib] __________________________

dependency_test_case = _TestCaseTuple(package_name='matplotlib', project_specifier={'<3.9.2', '>=3.0.1'}, precommit_specifier={'<3.9.3', '>=3.0.1'})

    def test_dependency(dependency_test_case):
        package_name, project_specifier, precommit_specifier = dependency_test_case
    
        msg = _test_both_dependencies_exist(*dependency_test_case)
        if msg:
            pytest.fail(msg)
        if project_specifier != precommit_specifier:
            project_specifier_str = ','.join(str(s) for s in project_specifier)
            precommit_specifier_str = ','.join(str(s) for s in precommit_specifier)
>           pytest.fail(
                f"The `mypy` pre-commit dependency '{package_name+precommit_specifier_str}' in '{PRE_COMMIT_CONFIG_PATH.name}'\n'"
                f"must match the `pyvista[typing]` dependency '{package_name+project_specifier_str}' in '{PYPROJECT_TOML_PATH.name}'."
            )
E           Failed: The `mypy` pre-commit dependency 'matplotlib<3.9.3,>=3.0.1' in '.pre-commit-config.yaml'
E           'must match the `pyvista[typing]` dependency 'matplotlib<3.9.2,>=3.0.1' in 'pyproject.toml'.
```

We can see from:
```
_TestCaseTuple(package_name='matplotlib', project_specifier={}, precommit_specifier={'<3.9.3', '>=3.0.1'})
```
that the test detects:
`matplotlib<3.9.2,>=3.0.1` in `pyproject.toml`
`matplotlib<3.9.3,>=3.0.1` in `pre-commit-config.yaml`

.pre-commit-config.yaml
user27182: > We can also consider the second suggestion in [python/mypy#13916](https://github.com/python/mypy/issues/13916) to
> 
> > Use language: system (and maybe a helper script) to run something from outside a pre-commit environment

If there is a strong preference for using `system` let me know and I will rework this PR accordingly

MatthewFlamm: I do not have a strong preference here. I would ask @tkoyama010 about the maintenance burden as they have been taking the lead on version updates recently.

While the addition of the test, which I haven't thoroughly reviewed, should enhance the robustness of keeping the requirements in sync, this is now code that will also need to be maintained, also adding several new test dependencies, etc..

user27182: The extra maintenance burden is an important consideration. There doesn't seem to be a perfect solution. 

If we use `system`, the burden of updating the environment is now placed on the developer. In this case it could lead to local errors not showing up in CI or vice-versa. This could be unexpected and may be difficult to track down, especially if a typing dependency is bumped mid-way through PR changes.

Perhaps it's better not to use `mypy` in pre-commit at all, and rely solely on CI for this. Devs may then optionally choose to run `mypy` locally if desired between commits. We could add a note in `CONTRIBUTING.md` about this.

MatthewFlamm: > Perhaps it's better not to use mypy in pre-commit at all, and rely solely on CI for this. Devs may then optionally choose to run mypy locally if desired between commits. We could add a note in CONTRIBUTING.md about this.

This is my personal preference, but it is only that.  I think other devs should weigh in on it.  There are certainly upsides to the approach in this PR.

user27182:
This PR seems to add a lot of overhead with limited benefit. If other parties are interested in this though I'm open to further discussion and updating this to make CI green again.

PR #6746, opened 2024-10-28
Split `test` requirements into separate `typing` and `test` requirements

user27182:
### Overview

See https://github.com/pyvista/pyvista/issues/5611#issuecomment-2176855411

Splitting the two will enable us to more carefully control which packages are included for type-checking. And it allows us to specify `numpy>=2.0`, which was part of the pre-commit config before #6741 was merged.


requirements_typing.txt
```diff
@@ -0,0 +1,5 @@
+# Note: using < here to reduce breakages
+# These should be automatically kept up to date by dependabot
+-r requirements.txt
+mypy<1.14.0
+npt-promote==0.1
```
user27182, line 5: Longer term, if #6738 is accepted, we should be able to move some packages out of `test` and into `typing` requirements (e.g. `scipy`, etc.). And remove the `--ignore-missing-imports` flag from `mypy` to properly type-check with these dependencies.

user27182, line 5:
```suggestion
mypy<1.14.0
npt-promote==0.1
numpy>=2.0,
```

These are just args to `pip`, and doing `pip install numpy>1.21 numpy>2.0` is valid, so we should keep this after all

MatthewFlamm (approved):
LGTM

tkoyama010 (approved):

Issue #5871, opened 2024-04-06
`GeoVista` integration tests are failing


user27182:
### Describe what maintenance you would like added.

See failure here:
https://github.com/pyvista/pyvista/actions/runs/8576853557/job/23508600334?pr=5822#step:9:51

Looks like it has to do with `cmocean`. And a new version was just released 2 hrs ago on `PyPI`
https://pypi.org/project/cmocean/#history

bjlittle:
@user27182 I've rerun the failing pyvista integration test for GeoVista on https://github.com/pyvista/pyvista/pull/5822 successfully.

It pulled in cmocean 4.0.3, which is the latest patch release on PyPI, and thankfully this works 😅

Seems like the released some bad builds over on cmocean, but they've now fixed that blip.

Happy to close this pull-request now?

user27182:
Good to know, thanks. It was version 4.0.2 when I opened this issue. If everything is working again as-is then yes this and the GeoVista PR can be closed 👍

PR #6753, opened 2024-10-30
Move requirements into pyproject.toml

user27182:
### Overview

Move requirements to keep all project dependencies in a single location.

EDIT:

This PR is mostly a like-for-like replacement, except for

- ~The main `pyvista` install dependencies are now pinned to the same upper-bound versions that were specified in `requirements.txt`.~ 
- `requirements.txt` is moved to a new `[pinned]` optional dependency (VTK is now included)
- A new `[dev]` optional dependency list is added


user27182:
Relevant PR: https://github.com/pyvista/pyvista/pull/6361

It would be nice to have `pyvista-no-vtk` or `pyvista[no-vtk]` for better define all the test environments in `pyproject.toml`

user27182:
pre-commit.ci autofix

.github/workflows/linkcheck.yml
```diff
@@ -23,9 +23,6 @@ jobs:
         with:
           python-version: "3.12"
           cache: "pip"
-          cache-dependency-path: |
```
user27182, line 26: Note: `cache-dependency-path` is not required. It will search `pyproject.toml` by default. https://github.com/actions/setup-python#caching-packages-dependencies

user27182:
TODO / follow-up: move dependencies out of `docker/requirements.txt` into `pyproject.toml`, and use `pip install pyvista[docker]` in the `docker/jupyter.Dockerfile` script

requirements.txt
```diff
@@ -1,12 +0,0 @@
-# Core dependencies
-# `vtk` must be explicitly installed
-# It is critical that VTK is *not* listed here.
-#  This is to prevent installing `vtk` from PyPI
-#  when using a custom variant like `vtk_osmesa`
-#  which we do in some of our CI jobs.
```
user27182, line 6: We can keep this file for this purpose if necessary. Though, at least for our CI jobs, in practice it seems that running 
```
pip install pyvista
pip uninstall vtk
pip install vtk-osmesa
```
is a more typical way to do this, and I think should be preferred over:
```
pip install --no-deps
pip install -r requirements.txt vtk-osmesa
```

The former is more explicit, though the latter is a bit wasteful with the unnecessary vtk install

If kept, it should probably be renamed (otherwise it may be ambigious if `pyproject.toml` or `requirements.txt` should be used)
Also, keeping this file would require bumping versions in two files instead of one...

Maybe `pip install pyvista-no-vtk` is best? https://github.com/pyvista/pyvista/pull/6361

requirements_test.txt
```diff
@@ -1,34 +0,0 @@
-# Note: using < here to reduce breakages
```
user27182, line 1: Should probably keep this note, move it into `pyproject.toml`

requirements_test.txt
```diff
@@ -1,34 +0,0 @@
...
-# These should be automatically kept up to date by dependabot
```
user27182, line 2: Not sure if this is still the case with `pyproject.toml`? maybe `dependabot` will just work without any config change?

banesullivan (changes requested):
This approach presents some challenges, and I’d like to understand more about the motivation behind it to ensure I fully understand the benefits you’re aiming for.

Historically, we’ve taken care to avoid pinning dependencies in released versions of PyVista. My primary concern is that strict versioning could introduce unnecessary constraints in users’ environments, potentially creating conflicts that would be difficult to resolve, especially with patch version constraints as currently outlined in the PR.

One relevant example is Pillow. Since it’s only used in specific contexts within PyVista (such as screenshots, texturing, or image generation) and isn’t essential to the core API, PyVista is compatible with a broad range of Pillow versions. However, other packages in a user’s environment might rely on more recent versions of Pillow, and additional constraints on this package could lead to avoidable compatibility issues.

banesullivan (changes requested):
This approach presents some challenges, and I’d like to understand more about the motivation behind it to ensure I fully understand the benefits you’re aiming for.

Historically, we’ve taken care to avoid pinning dependencies in released versions of PyVista. My primary concern is that strict versioning could introduce unnecessary constraints in users’ environments, potentially creating conflicts that would be difficult to resolve, especially with patch version constraints as currently outlined in the PR.

One relevant example is Pillow. Since it’s only used in specific contexts within PyVista (such as screenshots, texturing, or image generation) and isn’t essential to the core API, PyVista is compatible with a broad range of Pillow versions. However, other packages in a user’s environment might rely on more recent versions of Pillow, and additional constraints on this package could lead to avoidable compatibility issues.

MatthewFlamm:
> This approach presents some challenges, and I’d like to understand more about the motivation behind it to ensure I fully understand the benefits you’re aiming for.
> 
> Historically, we’ve taken care to avoid pinning dependencies in released versions of PyVista. My primary concern is that strict versioning could introduce unnecessary constraints in users’ environments, potentially creating conflicts that would be difficult to resolve, especially with patch version constraints as currently outlined in the PR.
> 
> One relevant example is Pillow. Since it’s only used in specific contexts within PyVista (such as screenshots, texturing, or image generation) and isn’t essential to the core API, PyVista is compatible with a broad range of Pillow versions. However, other packages in a user’s environment might rely on more recent versions of Pillow, and additional constraints on this package could lead to avoidable compatibility issues.

You bring up good points here.  As a counter point, there are advantages to pinning maximum versions.  A good example is #2859, where when a pre release of vtk was released for a new python version.  This could also manifest with breaking changes in a key dependency like numpy.  The pros/cons of both approaches need to be considered.  See also #2867.

A second issue is downsides to downstream packages that are currently utilizing requirements.txt-like files from here, if there are any. They probably shouldn't be, but it is possible.

user27182:
> This approach presents some challenges, and I’d like to understand more about the motivation behind it to ensure I fully understand the benefits you’re aiming for.

There are few reasons:

1. Dependencies fall out of sync. We have lower bounds for `matplotlib` and `numpy` in `pyproject.toml`
https://github.com/pyvista/pyvista/blob/ecd0dcf2b26430b29376f8605dfb2ce5b085f325/pyproject.toml#L29-L30
which are missing from `requirements.txt`
https://github.com/pyvista/pyvista/blob/ecd0dcf2b26430b29376f8605dfb2ce5b085f325/requirements.txt#L7-L8
If they are defined in exactly one location this would not happen.
2. Ambiguity. I honestly didn't even realize we had dependencies listed in _both_ locations until I started this PR. I saw `requirements.txt` and (wrongfully) assumed that those were the versions used by PyVista. I didn't know that `pip install pyvista` actually used unpinned versions of these dependencies.

2. Convenience. Doing something like `pip install pyvista[dev]` is so much nicer than doing:
```
pip install pyvista pre-commit
pip install requirements_test.txt
pip install requirenments_docs.txt
```
(though we _could_ also have `requirenments_dev.txt`, but this is still less convenient than above).

3. Reliability. Pinning things reduces the chances of things breaking unexpectedly. It's a feature, not a bug (see examples below).

4. Consistency. Encouraging users to use `requirements.txt` to install PyVista in some cases but then simply doing `pip install pyvista` in others means that users will likely have different and inconsistent environments depending on how it was installed.

> Historically, we’ve taken care to avoid pinning dependencies in released versions of PyVista. My primary concern is that strict versioning could introduce unnecessary constraints in users’ environments, potentially creating conflicts that would be difficult to resolve, especially with patch version constraints as currently outlined in the PR.
> 
> One relevant example is Pillow. Since it’s only used in specific contexts within PyVista (such as screenshots, texturing, or image generation) and isn’t essential to the core API, PyVista is compatible with a broad range of Pillow versions. However, other packages in a user’s environment might rely on more recent versions of Pillow, and additional constraints on this package could lead to avoidable compatibility issues.

For newer versions of PyVista, this should not be an issue since dependabot is used to ensure dependencies are always pinned to the latest versions. In that sense I don't believe the constraints are overly strict, since they are an upper bound of the latest versions available at the time of release.

But if users are trying to install older versions of PyVista with newer versions of dependencies, I would argue that the constraint is actually _necessary_ to ensure PyVista works as expected. We are only testing for the pinned versions, so we have no guarantee that our methods won't break with newer versions of dependencies. 

There are a few issues in the PyVista repo that were caused by unpinned versions of a dependency suddenly being updated (e.g. https://github.com/pyvista/pyvista/issues/5451, https://github.com/pyvista/pyvista/issues/5812, https://github.com/pyvista/pyvista/issues/5871). The most recent high-impact example was the release of NumPy 2.0. This broke many things, causing an emergency pin to be issued for GeoVista (https://github.com/pyvista/pyvista/pull/6292) and PyVista (https://github.com/pyvista/pyvista/pull/6210) just before its release. Unfortunately, since these pins are not present for older releases of PyVista, a user installing PyVista with `pip install pyvista==0.42` will have PyVista installed with NumPy 2.0. Does it work? Maybe, though it likely broke in the same way the PyVista `dev` branch did at the time NumPy 2.0 was released. We'll never know though, since we haven't tested this combination.

If you really need to use the latest version of NumPy with an older PyVista release, you probably know what you're doing, and you will now have to do something like:
```
pip install pyvista==0.42
pip uninstall numpy
pip install numpy>2.0
``` 
If anything breaks, at least you'll likely know what caused it because you went out of your way to install an untested combination of software.

user27182:
Packages like numpy, matplotlib, and scipy all seem not to apply upper bounds on dependencies for their releases. I do think there are issues with this approach, but we can drop that aspect from this PR. 

Pinning release dependencies is a bit of a distraction and more of a side effect of this PR, and not the main intention. The main point of this PR was more for maintenance and dev consistency, e.g. installing pyvista in a consistent and reliable manner for CI and dev without things breaking randomly on us and without running tests with slightly different environments. Having everything centralized in pyproject.toml helps with this. 

Part of what motivated these changes to begin with is all the issues we've been having with type checking CI which is in large part caused by having multiple and inconsistent environments being used.

Scipy keeps everything centralized in pyproject.toml and uses a script to auto-generate requirements files with pinned versions. Maybe we can do something similar to keep everything in sync.

akaszynski:
> Pinning release dependencies is a bit of a distraction and more of a side effect of this PR, and not the main intention. The main point of this PR was more for maintenance and dev consistency, e.g. installing pyvista in a consistent and reliable manner for CI and dev without things breaking randomly on us and without running tests with slightly different environments. Having everything centralized in pyproject.toml helps with this.

I fully agree that having everything centralized in `pyproject.toml` is ideal. Using `pip install -e .[test]`  instead of requirements files and `pip install -r ...` really simplifies installation and testing; it also aligns us to (what I believe) the community is gravitating towards as a best practice.

As for "Pinning release dependencies..." I agree with @banesullivan that this might present issues for several users who like to play fast and dirty with the latest and greatest and don't want pyvista holding them back. That being said, it 100% should be done for testing and docs, so the only change I'd ask for in this PR is loosening the direct dependencies.

user27182:
Release dependencies are no longer pinned

.github/workflows/linkcheck.yml
akaszynski: BTW, didn't realize that pyproject is now supported by the Python GitHub action. That's great, and one my previous reasons for holding out over `pyproject.toml` vs. requirements files (though I use `pyproject.toml` internally.

akaszynski (dismissed):
This gets a solid LGTM, and I like how you have a `pinned` section to help with CI.

.github/workflows/testing-and-deployment.yml
```diff
@@ -205,15 +196,19 @@ jobs:
...
-        run: pip install 'numpy~=${{ matrix.numpy-version }}.0'
+        run: |
+          pip uninstall numpy -y
+          pip install 'numpy~=${{ matrix.numpy-version }}.0'
 
       - name: Limit Matplotlib for VTK<9.2.2
         if: ${{ matrix.vtk-version == '9.1' || matrix.vtk-version == '9.0.3' }}
-        run: pip install 'matplotlib<3.6'
+        run: |
+          pip uninstall matplotlib -y
```
banesullivan, line 210: I'm curious, is this uninstall necessary? I would think/hope the following `install <x` line would handle uninstalling it

banesullivan (approved):
This is looking really good now and I'm really liking how it centralizes around pyproject.toml. One thing I'd like you to double check before merging is that the Dependabot config is compatible with pyproject.toml. 

I appreciate you leaving the release dependencies unpinned here for now and providing such well thought out explanations of the proposal. Perhaps as an isolated, follow-on PR we can start limiting the release dependencies in a way the improved guarantees of stability.

Thanks @user27182! Also a solid LGTM from me on this now

.github/workflows/testing-and-deployment.yml
user27182: I tried 
```
install matplotlib
install 'matplotlib<3.6'
```
and you are right, it works, so uninstall is not needed. This convention was copied from the `numpy` install a few lines above. Mayne that uninstall should also be removed.

user27182:
>  One thing I'd like you to double check before merging is that the Dependabot config is compatible with pyproject.toml.

Confirmed, it should work https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pip-and-pip-compile

> Dependabot supports updates to pyproject.toml files if they follow the PEP 621 standard.

(Our dependencies are defined in `dependencies`, and `[project.optional-dependencies]`, so we are following [PEP 621](https://peps.python.org/pep-0621/))

> Perhaps as an isolated, follow-on PR we can start limiting the release dependencies in a way the improved guarantees of stability.

To add to this discussion, we now have `pip install pyvista[pinned]`, and so maybe a good compromise is to keep the status quo (i.e. do not pin dependencies), and simply add a bit of additional documentation wherever the main "How-to Install PyVista" instructions are, and give users two options:

1. Use `pip install pyvista` for the general case,
2. Use `pip install pyvista[pinned]` for increased reliability

As a bonus, if there are any issues reported here in the future relating to installation problems with older versions of PyVista, we can now suggest running `pip install pyvista[pinned]` as a first troubleshooting step.

.github/workflows/testing-and-deployment.yml
```diff
@@ -205,15 +196,19 @@ jobs:
...
+          pip install 'matplotlib<3.6'
```
user27182, line 211:
```suggestion
        run: pip install 'matplotlib<3.6'
```

.github/workflows/testing-and-deployment.yml
```diff
@@ -205,15 +196,19 @@ jobs:
...
       - name: Install Testing Requirements
-        run: pip install -r requirements_test.txt
+        run: pip install .[test]
 
       - name: Set up numpy
         if: ${{ matrix.numpy-version != 'latest' && matrix.numpy-version != 'nightly' }}
-        run: pip install 'numpy~=${{ matrix.numpy-version }}.0'
+        run: |
+          pip uninstall numpy -y
+          pip install 'numpy~=${{ matrix.numpy-version }}.0'
```
user27182, line 205:
```suggestion
        run: pip install 'numpy~=${{ matrix.numpy-version }}.0'
```

user27182 (commented):
Revert unnecessary changes to numpy and matplotlib install

tkoyama010 (approved):

tkoyama010:
@user27182 Are you ready to proceed with the merge?

banesullivan (approved):

user27182:
Happy to report dependabot is still working https://github.com/pyvista/pyvista/pull/6769

akaszynski:
Thanks for all your work @user27182!

PR #6756, opened 2024-10-31
Define convention of pydocstyle explicitly for NumPy-style

tkoyama010:
### Overview

Define convention of pydocstyle explicitly for NumPy-style.

### Details

- [convention](https://docs.astral.sh/ruff/settings/#lint_pydocstyle_convention)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -418,3 +418,6 @@ force-single-line = true
 [tool.ruff.lint.pyupgrade]
 # Preserve types, even if a file imports `from __future__ import annotations`.
 keep-runtime-typing = true
+
+[tool.ruff.lint.pydocstyle]
+convention = "numpy"


user27182:
IIRC this convention with ruff will also _disable_ some rules as well. But, some of these disabled rules are good to include (e.g. I think the blank line at end of a docstring rule is excluded, though I think we should include it). Because of the implicit exclusion of rules I'm not sure I'm a fan of using the convention this way. 

This convention was previously enabled in the pydocstyle config but was dropped when it was moved to ruff. See the diff for pyproject.toml in #6632. In that PR a number of options were (explicitly) enabled which may differ from the numpy convention. Might be worth reviewing if we perhaps want to remove some of the options in case it's important to be more strictly aligned with the convention.

user27182:
For details about the rules excluded by the convention see 

https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings

> The NumPy convention includes all D errors apart from: D107, D203, D212, D213, D402, D413, D415, D416, and D417.

Not sure how that compares with our config, I think we have a few of these excluded rules enabled explicitly

user27182:
In particular I think we should keep D413 https://docs.astral.sh/ruff/rules/blank-line-after-last-section/

pyproject.toml
```diff
@@ -418,3 +418,6 @@ force-single-line = true
 [tool.ruff.lint.pyupgrade]
 # Preserve types, even if a file imports `from __future__ import annotations`.
 keep-runtime-typing = true
+
+[tool.ruff.lint.pydocstyle]
+convention = "numpy"
```
user27182, line 423:
```suggestion
[tool.ruff.lint.pydocstyle]
# The NumPy convention is already effectively enabled using `extend-select` and 
# `ignore`. Use those fields to configure project `pydocstyle` rules instead.
# See https://github.com/pyvista/pyvista/pull/6756/
# convention = "numpy"
```

I believe we are effectively already using this convention in the sense that most of its rules are already in force. Though we do enforce a few additional rules, and we ignore maybe one rule from the convention. But, I don't think we should enforce it using `convention = "numpy"`

> Because of the implicit exclusion of rules I'm not sure I'm a fan of using the convention this way.

After trying to integrate this option (partly through #6759), I still feel this way.

Our convention with the `ruff` config seems to be mostly subtractive, i.e. include the entire group in `extend-select` then ignore what you don't want. This is true for the `pydocstyle` rules we have now: we have `"D"` in `extend-select` with a few ignores.

However, using `convention = "numpy"` means we'll need to switch to an additive style to "re-add" the rules we want to use which are otherwise ignored by this style by default. This is not great because it breaks from convention, plus we'll need to both select _and_ ignore rules to make it work, which is a bit confusing. I also don't like how the selected and ignored rules are now implicit. This means we have to rely on the ruff documentation to understand which rules are being applied. And it's not obvious that it's implicit either, so one might only see "D413" in `extend-select` and think only one rule is being applied.

I think we should keep it in the config but not use it so that's it's clear we are intentionally avoid it.

banesullivan:
Closing due to staleness and that we mostly rely on ruff for this already. I'm up for evolving our style but I think these kinds of style changes should first be discussed and agreed up in an issue

Issue #6731, opened 2024-10-25
VTK 9.4 Testing Failures


banesullivan:
### Describe what maintenance you would like added.

VTK 9.4 appears to be on the horizon with release candidates and we have begun testing PyVista against this version. As tradition there are a slew of testing failures to address!

This issue will be where we tract the progress of these failures from a meta level and we will create other issues/PRs referencing this meta issue as needed.

cc @pyvista/kitware 

### Core Testing failures

Seeing ~277 failures with the overwhelming majority (~256) coming from the `test_text3d_source_parameters` test and a check in our `copy_from` method which is showing up in other tests (and also why are there 256 parameterized runs of `test_text3d_source_parameters`?!?! 🤯 )

The remaining ~21 failures are listed in the following section (captured from `pytest -v tests/core -k 'not test_text3d_source_parameters'`

<details>

```
================================================== short test summary info ===================================================
FAILED tests/core/test_dataobject.py::test_pickle_serialize_deserialize[vtk] - AssertionError: assert <vtkmodules.v...x742809ec4580> == vtkStructured...able: (none)\n\n
FAILED tests/core/test_dataobject.py::test_pickle_serialize_deserialize[xml] - AssertionError: assert <vtkmodules.v...x742809ddf7c0> == vtkStructured...able: (none)\n\n
FAILED tests/core/test_dataobject.py::test_pickle_serialize_deserialize[legacy] - AssertionError: assert <vtkmodules.v...x742809ddfdc0> == vtkStructured...able: (none)\n\n
FAILED tests/core/test_composite.py::test_multi_block_eq - AssertionError: assert MultiBlock (0...0.000, 282.130 == MultiBlock (0...0.000, 282.130
FAILED tests/core/test_dataobject.py::test_pickle_multiblock[vtk] - AssertionError: assert MultiBlock (0...0.000, 282.130 == MultiBlock (0...0.000, 282.130
FAILED tests/core/test_dataobject.py::test_uniform_eq - AssertionError: assert ImageData (0x... Arrays:     2 == ImageData (0x... Arrays:     2
FAILED tests/core/test_geometric_sources.py::test_text3d_source_modified[kwarg_tuple5] - TypeError: The Input DataSet type <class 'vtkmodules.util.data_model.vtkPolyData'> must be compatible with the one being ...
[...]

PR #6758, opened 2024-11-01
Support Python 3.13

user27182:
### Overview

Support Python 3.13

This is loosely a follow-up to #6701 (3.13 was released and 3.8 was dropped on the same day).


user27182:
Python 3.13 is not supported by VTK. And there doesn't seem to be support for it with vtk 9.4 either (at least not based on the release candidate wheels available at https://vtk.org/download/)

bjlittle:
This is pretty interesting.

The `conda-forge` ecosystem can already resolve and build the `py313` environment, and all the tests are passing over on `geovista`.

However, the dependencies don't install for PyPI.

See https://github.com/bjlittle/geovista/pull/1180

user27182:
So I guess vtk does work, but we'll just need to wait for the wheels to be made available?

bjlittle:
@user27182 It kinda seems that way, I guess.

However, I couldn't conclusively say that the `conda` offering 100% works ... only the parts we've got test coverage for 🫤

I'd love to know when the PyPI wheels will be available

banesullivan:
9.4 is released as of today with wheels for Python 3.13

bjlittle:
@user27182 I just re-ran my stalled PR over on `geovista` that attempts to extend support for `py313` ... and I'm glad to report that I can now create an environment (using `pip` rather than `conda`) that pulls in the latest `vtk` 9.4 and all my tests are passing :smile: 

Worth re-spinning the wheels here? :thinking:

banesullivan:
This is pretty much block by #6731

user27182:
yea, though I did see some 3.13-specific test failures earlier from `test_center_tuple`, I think some API for annotations changed, `inspect.getmembers` is behaving differently

PR #6759, opened 2024-11-01
Add explicit ruff pydocstyle codes

user27182:
### Overview

Follow-up to #6756

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -297,11 +297,6 @@ ignore = [
     "B904",
     "D105", # Missing docstring in magic method
     "D107", # Missing docstring in `__init__`
-    "D203", # May conflict with the formatter
-    "D211", # Incompatible with D203
-    "D213", # Incompatible with D212
-    "D402", # First line should not be the function's signature
-    "D416", # Section name ends in colon
     # whitespace before ':'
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -297,11 +297,6 @@ ignore = [
     "B904",
     "D105", # Missing docstring in magic method
     "D107", # Missing docstring in `__init__`
```
tkoyama010, line 299: D107 is also included in extend-select. Do you intend to remove it from the IGNORE list?
```suggestion
```

tkoyama010 (changes requested):

pyproject.toml
user27182: It should be removed from extend-select then. If we include it we get about 23 errors that need to be fixed. And I think we put the docs in the class docstring instead of in __init__ by convention

user27182: Also, I never really finished implementing this PR, so the lists are not fully complete

user27182:
Closing this as this was an attempt to see how we could support #6756, but I find it a bit contrived to use the `numpy` doc option with ruff since it implicitly extends and ignores options which we then need to override

PR #6767, opened 2024-11-05
Fix ruff lint error: open-file-with-context-handler (SIM115)

tkoyama010:
### Overview

Fix ruff lint error: open-file-with-context-handler (SIM115).

### Details

- [open-file-with-context-handler (SIM115)](https://docs.astral.sh/ruff/rules/open-file-with-context-handler/#open-file-with-context-handler-sim115)


tkoyama010:
@pyvista-bot LGTM

PR #6770, opened 2024-11-06
Update `macOS` testing strategy to run on latest `arm` architecture

user27182:
### Overview

Add testing with `macos-latest`.

### Details
macOS was pinned to `macos-12` in https://github.com/pyvista/pyvista/pull/5960

However, this means that we are not testing for compatibility with newer macbooks which use ARM processors instead of intel processors. The new architecture breaks compatibility with some of pyvista's optional dependencies, e.g. `embreex`, but this is undetected without testing on the newer architecture.

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -21,21 +21,22 @@ env:
 
 jobs:
   macOS:
-    # https://github.com/pyvista/pyvista/pull/5960
-    runs-on: macos-12
     name: Mac OS Unit Testing
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.11"]
+        config:
+          - { os: macos-12, python-version: 3.9 }  # intel processor
+          - { os: macos-latest, python-version: 3.12 }  # arm processor
+    runs-on: ${{ matrix.config.os }}
 
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


.github/workflows/testing-and-deployment.yml
```diff
@@ -21,21 +21,22 @@ env:
...
-    runs-on: macos-12
     name: Mac OS Unit Testing
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.11"]
+        config:
+          - { os: macos-12, python-version: 3.9 } # intel processor
+          - { os: macos-latest, python-version: 3.12 } # arm processor
+    runs-on: ${{ matrix.config.os }}
```
user27182, line 31: Not sure if there's a reason we only have two configs in the matrix, should we add more? E.g. windows testing has a runner for every supported python version
```suggestion
        config:
          - { os: macos-12, python-version: 3.9 } # intel processor
          - { os: macos-13, python-version: 3.10 } # intel processor
          - { os: macos-14, python-version: 3.11 } # arm processor
          - { os: macos-latest, python-version: 3.12 } # arm processor
    runs-on: ${{ matrix.config.os }}
```

banesullivan (approved):
This is excellent and well overdue! Thanks for updating these and setting up the matrix from cross platform testing on Mac

banesullivan:
It appears I may need to edit the "required checks" for CI

tkoyama010 (approved):

tkoyama010:
I added "macos-*" requirements

PR #6771, opened 2024-11-06
Restrict `embree` in test dependencies to intel macs

user27182:
### Overview

`embreex` does not work with newer mac hardware which use ARM processors. This PR makes it so that this package is not installed on this hardware.


tkoyama010 (approved):
LGTM

PR #6781, opened 2024-11-11
Add `taplo` pre-commit hook to format and sort `toml` files

user27182:
### Overview

`toml` files are not currently formatted or sorted. This enables things like:
https://github.com/pyvista/pyvista/blob/a71e027637a704648af4fbdac9befa425fd503ec/pyproject.toml#L69-L74

Where spaces around `=` are not enforced and where alphabetical sorting is not enforced (we have `n`->`m`->`n` order in the example)

This PR uses ~https://github.com/pappasam/toml-sort~ https://github.com/tamasfe/taplo to fix and enforce the formatting and sorting.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -93,5 +93,5 @@ repos:
     rev: v0.23.1
     hooks:
       - id: toml-sort
-        args: [--in-place]
+        args: [] # Args should be blank. Set args in `pyproject.toml` instead
         types_or: [toml]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -89,9 +89,9 @@ repos:
     hooks:
       - id: shfmt
 
-  - repo: https://github.com/pappasam/toml-sort
-    rev: v0.23.1
+  - repo: https://github.com/ComPWA/taplo-pre-commit
+    rev: v0.9.3
     hooks:
-      - id: toml-sort
-        args: [] # Args should be blank. Set args in `pyproject.toml` instead
-        types_or: [toml]
+      - id: taplo-format
+        # See options: https://taplo.tamasfe.dev/configuration/formatter-options.html
+        args: [--option, "reorder_arrays=true"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -88,3 +88,10 @@ repos:
     rev: v3.10.0-1
     hooks:
       - id: shfmt
+
+  - repo: https://github.com/pappasam/toml-sort
+    rev: v0.23.1
+    hooks:
+      - id: toml-sort
+        args: [--in-place]
+        types_or: [toml]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,8 +56,8 @@ jupyter = [
     'trame-vuetify>=2.3.1',
     'trame>=2.5.2',
 ]
+# Pinned versions of core dependencies
 pinned = [
-    # Pinned versions of core dependencies
     'matplotlib<3.9.3',
     'numpy<2.2.0',
     'pillow<11.1.0',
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -427,3 +400,14 @@ force-single-line = true
 [tool.ruff.lint.pyupgrade]
 # Preserve types, even if a file imports `from __future__ import annotations`.
 keep-runtime-typing = true
+
+[tool.tomlsort]
+in_place = true
+# Sorting options:
+no_sort_tables = true
+sort_inline_arrays = true
```
user27182, line 408: If it's desirable to have separate PRs, one for formatting and another for sorting, we can remove `sort_inline_arrays = true` and add this line in a separate PR

user27182:
> +100 sorting TOML files. However, could you consider whether using [Taplo](https://github.com/tamasfe/taplo) might be a better option compared to toml-sort?

I have no preference, `taplo` seems to have similar options https://taplo.tamasfe.dev/configuration/formatter-options.html, so we could probably use it instead

.pre-commit-config.yaml
```diff
@@ -88,3 +88,10 @@ repos:
     rev: v3.10.0-1
     hooks:
       - id: shfmt
+
+  - repo: https://github.com/ComPWA/taplo-pre-commit
+    rev: v0.9.3
+    hooks:
+      - id: taplo-format
+        # See options: https://taplo.tamasfe.dev/configuration/formatter-options.html
+        args: [--option, "reorder_arrays=true"]
```
user27182, line 97: With [fb0df95](https://github.com/pyvista/pyvista/pull/6781/commits/fb0df959950e6d2b6a2c1bbb02a24e99307f2ecd) I opted not to use `indent_string`, so the default indent of 2 is used. This is also the default for `toml-sort`. I think it's okay to use the default 2 spaces, but we can use 4 if preferred
```suggestion
        args: [--option, "reorder_arrays=true", --option, "indent_string=4"]
```

Have a look at the other [options](https://taplo.tamasfe.dev/configuration/formatter-options.html), are there any else we should include (can also add options in a separate PR)

user27182: The diff _is_ better with `"indent_string=    "` enabled if that's desired (i.e. fewer formatting changes). It's easier to see the sorting that way.

pyproject.toml
```diff
@@ -5,179 +5,158 @@ build-backend = 'setuptools.build_meta'
...
-    'sphinx_design<0.7.0',
-    'sympy<1.14.0',
-    'tqdm<4.67.0',
-    'trame>=2.5.2,<3.7.1',
-    'trame-vtk>=2.5.8,<2.8.12',
-    'trame-vuetify>=2.3.1,<2.7.2',
-    'trimesh<4.6.0',
-    'typing-extensions<4.13.0',
+  # Embreex does not work with arm-based macs
+  "embreex<2.17.8; sys_platform != 'darwin' or platform_machine != 'arm64'",
```
user27182, line 64: Flip the quote style so that this is properly sorted
```suggestion
  'embreex<2.17.8; sys_platform != "darwin" or platform_machine != "arm64"',
```

.pre-commit-config.yaml
tkoyama010: +1 using default.

tkoyama010 (approved):
LGTM

tkoyama010 (approved):

pyproject.toml
```diff
@@ -5,179 +5,158 @@ build-backend = 'setuptools.build_meta'
...
-    'trame>=2.5.2,<3.7.1',
-    'trame-vtk>=2.5.8,<2.8.12',
-    'trame-vuetify>=2.3.1,<2.7.2',
-    'trimesh<4.6.0',
-    'typing-extensions<4.13.0',
+  'cmocean<4.0.4',
+  'colorcet<3.2.0',
+  'pyvista[pinned]',
+  # Embreex does not work with arm-based macs
+  'embreex<2.17.8; sys_platform != "darwin" or platform_machine != "arm64"',
```
user27182, line 67: Is this a bug with the sorting? We have `p` before `e`, not a huge deal but not expected either.

tkoyama010: Following #6782, let's move comment inline. Additionally, it may be helpful to consider submitting an issue to Taplo regarding why the sorting functionality is not working as expected.

tkoyama010 (changes requested):

pyproject.toml
user27182: > Following #6782, let's move comment inline. 

Fixed in https://github.com/pyvista/pyvista/pull/6781/commits/d0eb514b2c0f87c06e2eea3ff2ed18d3a43e5559

> Additionally, it may be helpful to consider submitting an issue to Taplo regarding why the sorting functionality is not working as expected.

Apparently having full-line comments split up the sorting is the expected behaviour, see https://github.com/tamasfe/taplo/issues/675

So we should generally avoid using any full-line comments in any `toml` files with this tool.

tkoyama010 (approved):

MatthewFlamm:
This PR messed up the sorting of python version in the classifiers, see

https://github.com/pyvista/pyvista/blob/c39a649a0bd8e213125d11ad745009004cf4584d/pyproject.toml#L14-L17

Not sure if there is a way to fix this.

PR #6784, opened 2024-11-12
Support reading and saving pickled meshes

user27182:
### Overview

Allow using `mesh.save()` and `pv.read()` with `.pkl` or `.pickle` extension.

### DIFF
--- a/pyvista/core/dataobject.py
+++ b/pyvista/core/dataobject.py
@@ -20,7 +20,9 @@ from .pyvista_ndarray import pyvista_ndarray
 from .utilities.arrays import FieldAssociation
 from .utilities.arrays import _JSONValueType
 from .utilities.arrays import _SerializedDictArray
+from .utilities.fileio import PICKLE_EXT
 from .utilities.fileio import read
+from .utilities.fileio import save_pickle
 from .utilities.fileio import set_vtkwriter_mode
 from .utilities.helpers import wrap
 from .utilities.misc import abstract_class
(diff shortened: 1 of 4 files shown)


user27182:
One reason to support this PR and allow pickling directly from file is to better test existing pickling support. Pickling datasets is already supported (just not to/from file), but we aren't currently testing if changes to the PyVista API break pickling between versions. E.g. if I pickled a mesh with version `0.42`, can I still unpickle it with `0.45`? There is some dataset metadata stored with mesh when pickled/unpickled, and it's possible the format of this metadata could change and break pickling across versions.

With this PR though, we could more easily save a `.pkl` mesh file for a specific release version, and use it as input test data to ensure this format can still be read in the future.

pyvista/core/dataobject.py
```diff
@@ -149,32 +179,20 @@ def save(
...
+        writer_ext = self._WRITERS.keys()
+        if file_ext in writer_ext:
+            _write_vtk(self)
+        elif file_ext in PICKLE_EXT:
+            save_pickle(filename, self)
+        else:
+            raise ValueError(
+                'Invalid file extension for this data type.'
+                f' Must be one of: {list(writer_ext) + list(PICKLE_EXT)}',
+            )
```
tkoyama010, line 195:
```suggestion
        writer_exts = self._WRITERS.keys()
        if file_ext in writer_exts:
            _write_vtk(self)
        elif file_ext in PICKLE_EXT:
            save_pickle(filename, self)
        else:
            raise ValueError(
                'Invalid file extension for this data type.'
                f' Must be one of: {list(writer_exts) + list(PICKLE_EXT)}',
            )
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM. It is useful. Thanks!

PR #6786, opened 2024-11-12
Add recursive shallow copy for `MultiBlock`

user27182:
### Overview

#6590 fixed `MultiBlock.shallow_copy` so that only the root MultiBlock is copied by default. But sometimes it may be desirable to shallow-copy nested `MultiBlock`s recursively (this was the default behaviour prior to #6590). This PR adds the option to use the previous recursive behavior.

### DIFF
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -484,6 +484,21 @@ def test_multi_block_copy(deep, multiblock_all_with_nested_and_none):
         assert (multi[i] is multi_copy[i]) != deep or (multi[i] is None)
 
 
+@pytest.mark.parametrize('recursive', [True, False])
+def test_multi_block_shallow_copy(recursive, multiblock_all_with_nested_and_none):
+    multi = multiblock_all_with_nested_and_none
+    multi_copy = MultiBlock()
+    multi_copy.shallow_copy(multi, recursive=recursive)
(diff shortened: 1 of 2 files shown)


tests/core/test_composite.py
```diff
@@ -484,6 +484,21 @@ def test_multi_block_copy(deep, multiblock_all_with_nested_and_none):
...
+    multi_copy = MultiBlock()
+    multi_copy.shallow_copy(multi, recursive=recursive)
+    assert multi.n_blocks == multi_copy.n_blocks
+    for i in range(multi_copy.n_blocks):
+        block = multi_copy.GetBlock(i)
+        assert pv.is_pyvista_dataset(block) or block is None
+        if isinstance(multi[i], MultiBlock):
+            assert (multi[i] is multi_copy[i]) != recursive
+        else:
+            assert multi_copy[i] is multi[i]
```
tkoyama010, line 499:
```suggestion
    for i, block in enumerate(multi_copy):
        assert pv.is_pyvista_dataset(block) or block is None
        if isinstance(multi[i], MultiBlock):
            assert (multi[i] is multi_copy[i]) != recursive
        else:
            assert multi_copy[i] is multi[i]
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

Issue #4774, opened 2023-08-15
Decouple filters: Make sure PyVista’s filtering API is entirely isolated


alesgenova:
### Describe what maintenance you would like added.

Good morning,

I an an engineer at Kitware, and talking to @banesullivan before he left the company he mentioned a possible improvement to the code separation between data objects and their filters.

It appears that in some methods within a data object a method in its corresponding filter class is invoked. For example:

```python
@abstract_class
class DataSet(DataSetFilters, DataObject):
    @property
    def volume(self) -> float:
        sizes = self.compute_cell_sizes(length=False, area=False, volume=True)
        return sizes.cell_data['Volume'].sum()
```

`self.compute_cell_sizes()` is a method defined in the `DataSetFilters` class.

```python
@abstract_class
class DataSetFilters:
def compute_cell_sizes(self, length=True, area=True, volume=True, progress_bar=False):
        alg = _vtk.vtkCellSizeFilter()
        # ...
        return _get_output(alg)
```

There are many similar examples spread across the codebase.

Bane's suggestion was that if a method in a `DataObject` depends on something defined in the `Filter`, that method should actually be in the filter. This shouldn't be a breaking change since the DataObject class is a mixin.

I would like to get some input from the maintainers on if you think this is a worthwhile change. If so I'll be happy to open a PR in the next few days.

akaszynski:
> if a method in a DataObject depends on something defined in the Filter, that method should actually be in the filter.

Fully agree. In this case, `volume` should be in `DataSetFilters`.

> I would like to get some input from the maintainers on if you think this is a worthwhile change. If so I'll be happy to open a PR in the next few days.

I'd be happy to review it.

MatthewFlamm:
It might be worth considering the inheritance model, as it prevents catching this automatically.  Another piece of evidence is that we cannot type check the Filters because it does not inherit from the DataSet classes.

alesgenova:
@MatthewFlamm @akaszynski  I agree about the lack of type checking.

I briefly looked into how to get the mixin pattern used in pyvista to play nice with type checking, and this is a proof of concept that I think would work with the existing approach.

The central point is that the filter needs to be able to access attributes of both the `DataObject` super class and `DataSet` subclass, without inheriting from them.

Here is a possible working solution:

```python
# typing_mixins.py

from typing import Protocol

# Abstract protocols that define attributes and methods present of the base DataObject class
# and DataSet subclass

class DataObjectProtocol(Protocol):
[...]

PR #6788, opened 2024-11-13
Use `self: pyvista.DataSet` typing for `DataSet.transform` filter

user27182:
### Overview

I made one typing change: 
``` diff
     def transform(  # type: ignore[misc]
-        self: _vtk.vtkDataSet,
+        self: pyvista.DataSet,
```
And then addressed all subsequent type errors that were generated by this change.


user27182:
Superseded by #6790

tkoyama010 (approved):
LGTM

user27182:
After #6795 this is no longer a simple fix since `mypy` now tries to apply the `self: pyvista.DataSet` type to _all_ filters. I think perhaps a more complete solution is needed. Related issue: #4774

PR #6789, opened 2024-11-13
Make `DataSet` inherit from `vtkDataSet`

user27182:
### Overview

I made one change:

``` diff
- class DataSet(DataSetFilters, DataObject):
+ class DataSet(DataSetFilters, DataObject, _vtk.vtkDataSet):
```

And addressed all type errors generated by this change.

### DIFF
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -132,7 +132,7 @@ class ActiveArrayInfo:
 
 
 @abstract_class
-class DataSet(DataSetFilters, DataObject):
+class DataSet(DataSetFilters, DataObject, _vtk.vtkDataSet):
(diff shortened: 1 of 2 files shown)


user27182:
This won't work due to issues with the MRO. It seems that super classes should not inherit from vtk classes directly, and only concrete, child classes should inherit from vtk classes.

So, to fix type errors where both `DataSet` and `vtkDataSet` methods are used, we cannot specify the abstract `DataSet` type since it isn't a subclass of `vtkDataSet`. Instead, we must specify any/all concrete implementations explicitly, since those _are_ subclasses of `vtkDataSet`. E.g. we need to type things with `Union[PolyData, UnstructuredGrid,... etc]` instead of using `DataSet`.

PR #6791, opened 2024-11-13
Fix missing plotting typing imports

user27182:
### Overview

Fix bad typing imports and resolve related typing errors that were generated after fixing the imports.


tkoyama010 (approved):
LGTM

PR #6799, opened 2024-11-14
Move transformations from `DataSet` to `DataSetFilters`

user27182:
### Overview

We have `DataSetFilters.transform` and `DataSetFilters.reflect`, but all other transformation methods are `DataSet` methods. This is inconsistent and confusing, especially since these methods pretty much all call `DataSetFilters.transform` internally. 

This PR moves all methods to `DataSetFilters`. Related tests are also moved.


pyvista/core/dataset.py
```diff
@@ -35,17 +35,14 @@
...
 from .utilities.points import vtk_points
-from .utilities.transform import Transform
 
 if TYPE_CHECKING:  # pragma: no cover
     from collections.abc import Callable
     from collections.abc import Generator
     from collections.abc import Iterator
 
     from ._typing_core import MatrixLike
-    from ._typing_core import Number
```
user27182, line 46: `scale` was the only method to use this type. This was changed to `float` as part of the move and so `Number` is no longer imported. I can revert this and keep `Number` if desired.

user27182: Reverted in [72bc225](https://github.com/pyvista/pyvista/pull/6799/commits/72bc2256fd937e1f80c86693e50417ab577d3530)

user27182:
Related issue on decoupling filters from datasets:
https://github.com/pyvista/pyvista/issues/4774

akaszynski (approved):
Agree with the change. LGTM.

tkoyama010 (approved):

user27182:
Just a heads up that since a few doc build images were removed by this PR, this messes up the build cache. So, all Documentation Build checks will now fail unless they use the `no-sphinx-build-cache` label.  The build cache should be fixed once the next PR (with the `no-sphinx-build-cache` label) gets merged into main.

PR #6801, opened 2024-11-15
Add Koch snowflake example

user27182:
### Overview

Add new loadable example based on https://examples.vtk.org/site/Python/Visualization/KochSnowflake/


pyvista/examples/examples.py
```diff
@@ -672,6 +673,183 @@ def _hydrogen_orbital_load_func(n=1, l=0, m=0, zoom_fac=1.0):
...
+
+def _snowflake_load_func(level=6, as_triangles=False):
+    from math import cos
+    from math import pi
+    from math import sin
+    from math import sqrt
+
+    def make_polyline(points, level_):
+        """Koch Snowflake as a vtkPolyLine."""
+        # Use the points from the previous iteration to create the points of the next
```
MatthewFlamm, line 735: This looks to be a significant copy and paste from the vtk-examples code.  Do we need to have a license associated with it then?  If we want to avoid this, we could recast the example to use PyVista code

MatthewFlamm (commented):
I think it is a good addition.

Issue #6589, opened 2024-08-29
Type checking failing with usage of wraps


banesullivan:
Our usage of `wraps()` is causing us to see this issue in MyPy https://github.com/python/mypy/issues/17166 when typechecking usage of pyvista's API

(can post reproducible error and message later, but wanted to track this while on my mind)

user27182:
Here's an example of this bug occurring with `Actor.plot`
https://github.com/pyvista/pyvista/blob/3f0fad3f42d9b491679e6aa50e52d93c1a81c042/pyvista/plotting/actor.py#L281-L283
``` python
pyvista/plotting/actor.py: note: In member "plot" of class "Actor":
pyvista/plotting/actor.py:282: error: Argument 1 to "__call__" of "_Wrapped" has incompatible type "Actor"; expected "BasePlotter"  [arg-type]
            pl.add_actor(self)
                         ^~~~
```

fosskers:
This is also happening to me. `mypy` really doesn't like `_Wrapped`, whatever it is.

PR #6809, opened 2024-11-18
Add `None` return annotations to `actor.py`

user27182:
### Overview

I ran `python -m autotyping --none-return pyvista/plotting/actor.py`. One error was reported which had to be ignored due to a bug with mypy (see #6589 )


tkoyama010 (approved):
LGTM

PR #6820, opened 2024-11-19
Add `None` return type annotations to most modules in `core`

user27182:
### Overview

I ran `python -m autotyping pyvista/core --none-return` to automatically add None return annotations and then rolled back changes to modules which generated errors. Omitted from this PR:

- reader.py
- observers.py
- geometric_source.py

All changes are automatic.


user27182:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM

PR #6822, opened 2024-11-20
Update typing for active array info and field association in `dataset.py` and `arrays.py`

user27182:
### Overview

Most of the typing updates relate to `ActiveFieldInfo` and `FieldAssociation`, though there are some other minor unrelated typing changes as well (e.g. magic methods).

Also:

- The override for `ActiveArrayInfo.__setitem__` is  removed (has no effect)
- The override for `ActiveArrayInfo.__getattr__` is removed. This was an unnecessary "middle-man" that prevented `mypy` from being able to infer the type of `association` and `name` attributes.


tkoyama010 (approved):
LGTM

PR #6824, opened 2024-11-20
Adds option to `DataSetFilters.glyph` to orient using vector or normal mode

### LINKED ISSUE #6823: Add vector mode options to glyph filter

Currently when using `DataSetFilters. glyph` the glyph is aligned using "VTK_USE_VECTOR". In some cases, it is handy to be able to orient the glyph using "VTK_USE_NORMAL". 

This is possible by adding an extra argument to glyph `vector_mode` and allowing this to be either vector or normal, similar to how color_mode works.  

Link to c++ code https://vtk.org/doc/nightly/html/classvtkGlyph3D.html#a11eb951daf212698f21fcfcf80c76a96

### Pseudocode or Screenshots

Using normal mode:
image
Using vector mode:
image

lachlangrose:
### Overview

Add the ability to choose whether to orient with vector or normal model when using the DataSetFilters.glyph method. Default is vector mode which is consistent with the existing implementation.

resolve #6823

### Details

Similar to recently added color_mode, as a flag allowing the user to specify the vector mode.

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -2536,6 +2536,7 @@ class DataSetFilters:
             If ``'normal'``, the glyphs are aligned using vtk normal mode.
 
             .. versionadded:: 0.45
+
         progress_bar : bool, default: False
             Display a progress bar to indicate progress.
 
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -2699,14 +2699,14 @@ class DataSetFilters:
             alg.SetColorModeToColorByVector()
         else:
             raise ValueError(f"Invalid color mode '{color_mode}'")
-        
+
         if vector_mode == 'vector':
             alg.SetVectorModeToUseVector()
         elif vector_mode == 'normal':
             alg.SetVectorModeToUseNormal()
         else:
             raise ValueError(f"Invalid vector mode '{vector_mode}'")
-        
+
         if rng is not None:
             alg.SetRange(rng)
         alg.SetOrient(orient)
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -2707,7 +2707,7 @@ class DataSetFilters:
         elif vector_mode == 'normal':
             alg.SetVectorModeToUseNormal()
         else:
-            raise ValueError('``vector_mode`` must be ``'vector'`` or ``'normal'``.')
+            raise ValueError(f"Invalid vector mode '{vector_mode}'")
 
         if rng is not None:
             alg.SetRange(rng)
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -2466,6 +2466,7 @@ class DataSetFilters:
         clamping=False,
         rng=None,
         color_mode='scale',
+        vector_mode='vector',
         progress_bar=False,
     ):
         """Copy a geometric representation (called a glyph) to the input dataset.
@@ -2529,6 +2530,10 @@ class DataSetFilters:
             If ``'vector'`` , color the glyphs by vector.
 
             .. versionadded:: 0.44
+
+        vector_mode : str, optional, default: ``'vector'``
+            If ``'vector'``, the glyphs are aligned using vtk vector mode.
+            If ``'normal'``, the glyphs are aligned using vtk normal mode.
 
         progress_bar : bool, default: False
             Display a progress bar to indicate progress.
@@ -2694,12 +2699,18 @@ class DataSetFilters:
             alg.SetColorModeToColorByVector()
         else:
             raise ValueError(f"Invalid color mode '{color_mode}'")
-
+        
+        if vector_mode == 'vector':
+            alg.SetVectorModeToUseVector()
+        elif vector_mode == 'normal':
+            alg.SetVectorModeToUseNormal()
+        else:
+            raise ValueError(f"Invalid vector mode '{vector_mode}'")
+        
         if rng is not None:
             alg.SetRange(rng)
         alg.SetOrient(orient)
         alg.SetInputData(source_data)
-        alg.SetVectorModeToUseVector()
         alg.SetScaleFactor(factor)
         alg.SetClamping(clamping)
         _update_alg(alg, progress_bar, 'Computing Glyphs')
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -2535,6 +2535,7 @@ class DataSetFilters:
             If ``'vector'``, the glyphs are aligned using vtk vector mode.
             If ``'normal'``, the glyphs are aligned using vtk normal mode.
 
+            .. versionadded:: 0.45
         progress_bar : bool, default: False
             Display a progress bar to indicate progress.
 
@@ -2705,7 +2706,7 @@ class DataSetFilters:
         elif vector_mode == 'normal':
             alg.SetVectorModeToUseNormal()
         else:
-            raise ValueError(f"Invalid vector mode '{vector_mode}'")
+            raise ValueError('``vector_mode`` must be ``'vector'`` or ``'normal'``.')
 
         if rng is not None:
             alg.SetRange(rng)


tkoyama010:
pre-commit.ci autofix

pyvista/core/filters/data_set.py
```diff
@@ -2694,12 +2699,18 @@ def glyph(
...
         else:
             raise ValueError(f"Invalid color mode '{color_mode}'")
-
+        
+        if vector_mode == 'vector':
+            alg.SetVectorModeToUseVector()
+        elif vector_mode == 'normal':
+            alg.SetVectorModeToUseNormal()
+        else:
+            raise ValueError(f"Invalid vector mode '{vector_mode}'")
```
tkoyama010, line 2708:
```suggestion
            raise ValueError('``vector_mode`` must be ``'vector'`` or ``'normal'``.')
```

pyvista/core/filters/data_set.py
```diff
@@ -2530,6 +2531,10 @@ def glyph(
 
             .. versionadded:: 0.44
 
+        vector_mode : str, optional, default: ``'vector'``
+            If ``'vector'``, the glyphs are aligned using vtk vector mode.
+            If ``'normal'``, the glyphs are aligned using vtk normal mode.
+
```
tkoyama010, line 2537:
```suggestion

            .. versionadded:: 0.45
```

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
```diff
@@ -2695,11 +2702,17 @@ def glyph(
         else:
             raise ValueError(f"Invalid color mode '{color_mode}'")
 
+        if vector_mode == 'vector':
+            alg.SetVectorModeToUseVector()
+        elif vector_mode == 'normal':
+            alg.SetVectorModeToUseNormal()
+        else:
+            raise ValueError(f"Invalid vector mode '{vector_mode}'")
```
tkoyama010, line 2710: Thank you for your contribution! Could you add some test code for this code? Other than that, it's perfect.

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
lachlangrose: @tkoyama010 I was a little bit quick writing this fix... when I started making tests it is not as simple as I thought. 

Looking at the vtk code in more detail it the `SetVectorModeToUseNormal` doesn't change the way the glyphs are interpreted rather it changes to using an attribute on the vtk object inNormals... See [here](https://github.com/Kitware/VTK/blob/1fe1cd1f467d432aa18b45f0b8a8ef278dfb8b5d/Filters/Generic/vtkGenericGlyph3DFilter.cxx#L176).

What confused me is the different behaviour between how normal is interpreted as a property of a pyvista geometry class and how that normal is used by the glyph filter...

`import pyvista as pv

p1 = pv.PolyData(np.array([[0, 0, 0]]))
p1["norm"] = np.array([[1, 0, 0]])
p2 = pv.PolyData(np.array([[1, 1, 1]]))
p2["norm"] = np.array([[0, 1, 0]])
p3 = pv.PolyData(np.array([[2, 2, 2]]))
p3["norm"] = np.array([[0, 0, 1]])

disk1 = pv.Disc(center=[0, 0, 0], normal=[1, 0, 0], inner=0)
disk2 = pv.Disc(center=[1, 1, 1], normal=[0, 1, 0], inner=0)
disk3 = pv.Disc(center=[2, 2, 2], normal=[0, 0, 1], inner=0)
geom = pv.Disc()
p1 = p1.glyph(
    geom=geom,
    orient="norm",
)
p2 = p2.glyph(
    geom=geom,
    orient="norm",
)
p3 = p3.glyph(
    geom=geom,
    orient="norm",
)
view = pv.Plotter()
view.add_mesh(disk1, name="disk1", color="red")
view.add_mesh(disk2, name="disk2", color="green")
view.add_mesh(disk3, name="disk3", color="blue")
view.add_mesh(p1, name="p1", color="red")
view.add_mesh(p2, name="p2", color="green")
view.add_mesh(p3, name="p3", color="blue")
view.show()`


image


If we use a glyph like arrow
import pyvista as pv

p1 = pv.PolyData(np.array([[0, 0, 0]]))
p1["norm"] = np.array([[1, 0, 0]])
p2 = pv.PolyData(np.array([[1, 1, 1]]))
p2["norm"] = np.array([[0, 1, 0]])
p3 = pv.PolyData(np.array([[2, 2, 2]]))
p3["norm"] = np.array([[0, 0, 1]])

disk1 = pv.Arrow(start=[0, 0, 0], direction=[1, 0, 0], )
disk2 = pv.Arrow(start=[1, 1, 1], direction=[0, 1, 0], )
disk3 = pv.Arrow(start=[2, 2, 2], direction=[0, 0, 1], )
geom = pv.Arrow()
p1 = p1.glyph(
    geom=geom,
    orient="norm",
)
p2 = p2.glyph(
    geom=geom,
    orient="norm",
)
p3 = p3.glyph(
    geom=geom,
    orient="norm",
)
view = Loop3DView()
view.add_mesh(disk1, name="disk1", color="red")
view.add_mesh(disk2, name="disk2", color="green")
view.add_mesh(disk3, name="disk3", color="blue")
view.add_mesh(p1, name="p1", color="red")
view.add_mesh(p2, name="p2", color="green")
view.add_mesh(p3, name="p3", color="blue")
view.show()
image

I found that if I rotate the geometry by 90 degrees before passing it to the glyph method it will give me the results I expect. 
`import pyvista as pv

p1 = pv.PolyData(np.array([[0, 0, 0]]))
p1["norm"] = np.array([[1, 0, 0]])
p2 = pv.PolyData(np.array([[1, 1, 1]]))
p2["norm"] = np.array([[0, 1, 0]])
p3 = pv.PolyData(np.array([[2, 2, 2]]))
p3["norm"] = np.array([[0, 0, 1]])

disk1 = pv.Disc(center=[0, 0, 0], normal=[1, 0, 0], inner=0)
disk2 = pv.Disc(center=[1, 1, 1], normal=[0, 1, 0], inner=0)
disk3 = pv.Disc(center=[2, 2, 2], normal=[0, 0, 1], inner=0)
geom = pv.Disc()
geom = geom.rotate_y(90)
p1 = p1.glyph(
    geom=geom,
    orient="norm",
)
p2 = p2.glyph(
    geom=geom,
    orient="norm",
)
p3 = p3.glyph(
    geom=geom,
    orient="norm",
)
view = pv.Plotter()
view.add_mesh(disk1, name="disk1", color="red")
view.add_mesh(disk2, name="disk2", color="green")
view.add_mesh(disk3, name="disk3", color="blue")
view.add_mesh(p1, name="p1", color="red")
view.add_mesh(p2, name="p2", color="green")
view.add_mesh(p3, name="p3", color="blue")
view.add_arrows(np.array([[0,0,0],[1,1,1],[2,2,2]]),np.array([[1,0,0],[0,1,0],[0,0,1]]),color='black')

view.show()`

I noticed the same issue with the rectangle glyph and parametric ellipsoid depending on the aspect ratio of the ellipsoid.  I can modify this PR to just rotate the geometry by 90 degrees in the y direction if "normals" is chosen but I am not sure if it would be better to use a different name.

lachlangrose:
I don't think it is worthwhile adding this change anymore, unless its worth adding a parameter to rotate the initial glyph orientation. But this can just be done to the glyph before passing it to the filter. Happy to update or for it to be closed.

MatthewFlamm:
My understanding is that this option in the vtk filter changes the active attribute that is used to orient the object.  The `orient` parameter in pyvista currently inherently always uses the active vector.  That is, if 'orient' is a string it sets that array to be the active vectors, e.g. `orient='velocity'`.

You can do `orient='Normals'` or whatever the normals vector is named, and it should work just like the vtk filter option, but it will also set that array as the active vector.  Since we aren't operating in a multistep pipeline mode for PyVista filters this is generally OK.  But, adding this option would, in theory, enable running the glyph filter without resetting the active vectors and using the active normals instead.   It doesn't sound like you need this functionality, and it would take some extra work in handling the difference in active normals versus active vectors to get the current PR merged.

Up to you on whether to continue on this or not.  But either way, thank you for getting involved here!

PR #6829, opened 2024-11-21
Add note to `glyph` filter about `geom` direction assumption

user27182:
### Overview

See https://github.com/pyvista/pyvista/issues/6823#issuecomment-2489700375


tkoyama010 (approved):
LGTM
Screenshot_20241122-053040

pyvista/core/filters/data_set.py
```diff
@@ -2511,6 +2511,14 @@ def glyph(
             ``indices``. The values of the range (see ``rng``) affect lookup
             in the table.
 
+            .. note::
+
+                The provided geometry should initially be "pointing" in the
```
MatthewFlamm, line 2516: Should we say that the `The reference direction is relative to (1, 0, 0). It is recommended that the provided geometry should be "pointing"...` as the first sentence?  It can be valid for the provided geometry to not be "pointing" in the positive "+x" direction if the vectors are also transformed in the same way.  This is complicated, but it can be done.

MatthewFlamm: Another thought: I'm not sure if there is a better way to explain that there is not a true "pointing" towards `(1, 0, 0)` for a mesh unless it is 1D (even then it is pointing in two directions).  It is all about the reference direction you want to transform via the vector.  Suggestion for wording could be `The reference direction is relative to (1, 0, 0) on the provided geometry. That is, the provided geometry will be rotated from (1, 0, 0) to the direction of the orient vector at each point.`

tkoyama010 (changes requested):
Waiting until finish review.

pyvista/core/filters/data_set.py
```diff
@@ -2511,6 +2511,14 @@ def glyph(
...
+                positive ``+x`` direction. E.g. use ``direction=(1, 0, 0)``
+                when using some :ref:`Geometric Objects <geometry_api>` or
+                rotate the geometry manually, e.g. with
+                :func:`~pyvista.DataSetFilters.rotate_vector`.
```
user27182, line 2520: From https://github.com/pyvista/pyvista/pull/6829#discussion_r1854511065
```suggestion
                The reference direction is relative to ``(1, 0, 0)`` on the
                provided geometry. That is, the provided geometry will be rotated
                from ``(1, 0, 0)`` to the direction of the ``orient`` vector at
                each point.
```

user27182: The new text is a better, more accurate description. I couldn't find good wording to also add the note for direction objects. So, I think the changes arei good as suggested.

Leaving this open for @MatthewFlamm to commit for co-author credit

PR #6835, opened 2024-11-21
Deprecate `ActiveArrayInfo` and use `ActiveArrayInfoTuple` directly

user27182:
### Overview

The `ActiveArrayInfo` class is just a wrapper for a named tuple. Its repr, getitem, and iter methods all wrap the named tuple methods directly. This is problematic for type checking since unpacking the tuple this way will result in all items of the tuple having the same unionized type. So, the type of  the unpacked `association` and `name` fields are ambiguous.

It's better to just use the `NamedTuple` class directly. In this PR the copy method is moved to `ActiveArrayInfoTuple` but otherwise none of the other methods are needed and can later be removed.


bjlittle:
@user27182 Re-ran the failing `Integration Tests / GeoVista` after merging [PR#1219](https://github.com/bjlittle/geovista/pull/1219) on `geovista`, and you're unblocked, thanks :+1:

tests/core/test_dataset.py
```diff
@@ -1393,3 +1393,11 @@ def test_active_t_coords_deprecated(mesh):
         mesh.active_t_coords = t_coords
         if pv._version.version_info[:2] > (0, 46):
             raise RuntimeError('Remove this deprecated property')
+
+
+def test_active_array_info_deprecated(mesh):
+    match = 'ActiveArrayInfo is deprecated. Use ActiveArrayInfoTuple instead.'
+    with pytest.warns(PyVistaDeprecationWarning, match=match):
+        pv.core.dataset.ActiveArrayInfo(association=pv.FieldAssociation.POINT, name='name')
+        if pv._version.version_info >= (0, 48):
```
tkoyama010, line 1402: See #6850 and #6862.
```suggestion
        if pv._version.version_info[:2] > (0, 48):
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

tkoyama010 (approved):

tkoyama010 (approved):

PR #6842, opened 2024-11-22
Remove redundant "missing docstring" `GL08` from `numpydoc-validation`

user27182:
### Overview

Do not use numpydoc-validation to check for missing docstrings. This is redundant and is covered by `ruff` [D102](https://docs.astral.sh/ruff/rules/undocumented-public-method/) and [D103](https://docs.astral.sh/ruff/rules/undocumented-public-function/).

In particular this is nice because whereas numpydoc will complain for overloaded type signatures, requiring multiple ignores (one for each overload), `ruff` does not complain about missing docstrings for overloads. `ruff` also does not complain about missing docstrings in setters, which is consistent with `pyvista` practice of allowing docstrings in the property getters only. `ruff` also allows for finer control, e.g. there is a separate rule [D105](https://docs.astral.sh/ruff/rules/undocumented-magic-method/) for magic methods.


pyproject.toml
```diff
@@ -227,13 +227,16 @@ warn_unused_ignores = true
...
+  "all",
+  # all but the following:
   "ES01", # Not all docstrings need an extend summary.
   "EX01", # Examples: Will eventually enforce
   "GL01", # Contradicts numpydoc examples
   "GL02", # Permit a blank line after the end of our docstring
   "GL03", # Considering enforcing
   "GL06", # Found unknown section
   "GL07", # "Sections are in the wrong order. Correct order is: {correct_sections}",
+  "GL08", # The object does not have a docstring
```
user27182, line 239: All changes in this PR _except_ for the ones here in `pyproject.toml` are automatic. So, if desired, we can create a separate PR with only the rule change, and merge that first. That way, all changes here will be fully automatic, and we can then add this PR (with auto-changes only) to the git blame ignore.

MatthewFlamm (commented):
This is great.  The numpydoc ignores were very ugly.

Given the merge model, I would vote for merging as is.  If we allowed another merge type, for example rebasing the commits onto main, we could do it all in one step.  Without that, it adds too much churn for reviewing and merging IMO.  I'm sympathetic to having a single merge type for the policy, as a slip of the finger can accidentally use an unintended type.

pyproject.toml
```diff
@@ -227,13 +227,16 @@ warn_unused_ignores = true
 
 [tool.numpydoc_validation]
 checks = [
+  "all",
+  # all but the following:
```
tkoyama010, line 231: Inline comments are considered better for compatibility with the formatter in my mind.
```suggestion
  "all", # all but the following:
```

tkoyama010 (changes requested):

pyproject.toml
user27182: I noticed this comment has a colon, but it was at the bottom of the list. It seems to make more sense for `all` (and its comment) to be the first entry. But, it's not a necessary change, we can revert it.

Making the comment _not_ inline in this case breaks the sorting into two (above and below are sorted independently). This was considered a bug here https://github.com/pyvista/pyvista/pull/6781#discussion_r1838729385, but for this PR it's a feature. If we keep the comment inline, we'll need a blank comment after if we want `all` to be first.

```suggestion
  "all", # all but the following:
  #
```

pyvista/core/utilities/geometric_sources.py
```diff
@@ -1045,16 +1045,17 @@ def normal(self) -> tuple[float, float, float]:  # numpydoc ignore=RT01
         return self._normal
 
     @normal.setter
-    def normal(self, normal: Sequence[float]):
-        self._normal = float(normal[0]), float(normal[1]), float(normal[2])
+    def normal(self, normal: VectorLike[float]) -> None:
+        valid_normal = _validation.validate_array3(normal, dtype_out=float, to_tuple=True)
```
tkoyama010, line 1049:
```suggestion
        normal_ = _validation.validate_array3(normal, dtype_out=float, to_tuple=True)
```

pyvista/core/utilities/geometric_sources.py
```diff
@@ -1045,16 +1045,17 @@ def normal(self) -> tuple[float, float, float]:  # numpydoc ignore=RT01
...
+        self._normal = cast(tuple[float, float, float], valid_normal)
```
tkoyama010, line 1050:
```suggestion
        self._normal = cast(tuple[float, float, float], normal_)
```

tkoyama010 (approved):
LGTM

banesullivan (approved):

CONTRIBUTING.rst
```diff
@@ -348,17 +348,15 @@ These standards will be enforced using ``pre-commit`` using
 If for whatever reason you feel that your function should have an exception to
 any of the rules, add an exception to the function either in the
 ``[tool.numpydoc_validation]`` section in ``pyproject.toml`` or add an inline
```
adeak, line 350: The changes in this module also don't look auto-generated (concerning https://github.com/pyvista/pyvista/pull/6842#discussion_r1854344488).

user27182: Correct. These changes were added after that comment was made.

adeak: OK. I'm resolving that thread in that case.

tkoyama010 (approved):

banesullivan (approved):

banesullivan (approved):

tkoyama010 (approved):

PR #6840, opened 2024-11-22
Add typing overloads to `wrap`

user27182:
### Overview

Add typing overloads for wrapping meshes.

This fixes approx 60 type errors that were ignored.

PR #6851, opened 2024-11-23
Add typing overloads to `is_pyvista_dataset`

user27182:
### Overview

Follow-up to #6840. The overloads here are included as part of that PR.

### DIFF
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -1609,7 +1609,7 @@ class DataSet(DataSetFilters, DataObject):
         else:
             self.shallow_copy(mesh)
         if is_pyvista_dataset(mesh):
-            self.copy_meta_from(mesh, deep=deep)  # type: ignore[arg-type]
+            self.copy_meta_from(mesh, deep=deep)
 
     def cast_to_unstructured_grid(self) -> pyvista.UnstructuredGrid:
         """Get a new representation of this object as a :class:`pyvista.UnstructuredGrid`.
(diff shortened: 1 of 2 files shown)


tkoyama010 (approved):
Thanks! Great to see using overload.

tkoyama010 (changes requested):
@user27182 I talked to a friend in the Python community and they told me that the following pep might be the feature you are looking for. Could you please share your thoughts on this before merging?
https://peps.python.org/pep-0647/
https://peps.python.org/pep-0742/

user27182:
> @user27182 I talked to a friend in the Python community and they told me that the following pep might be the feature you are looking for. Could you please share your thoughts on this before merging? https://peps.python.org/pep-0647/ https://peps.python.org/pep-0742/

Thanks for looking into this. Is this for solving the `DataSet` is not a `vtkDataSet` type issue? `TypeIs` looks to be useful, though I'm not sure if it can help here. `TypeIs` does type-narrowing, here I think we want something a bit different, more like registering `DataSet` as a subclass of `vtkDataSet` even though it isn't. [ABCMeta.register](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register) could work but it is [not supported by mypy](https://github.com/python/mypy/issues/2922).

user27182:
Or are you suggesting using `TypeIs` for this PR? Instead of or in addition to the overloads?

tkoyama010 (approved):
Thank you for your comments. I just wanted to get your opinion on this for my own future inspirations. Not being implemented in mypy, etc. is indeed a major barrier. Let's keep this PR as it is.

PR #6852, opened 2024-11-23
Include `None` type with `TypeMultiBlockLeaf`

user27182:
### Overview

Split from #6840, this change was found to be necessary to resolve typing errors there. `None` is a valid leaf, so I think it makes sense to include it with the Union directly.


tkoyama010 (dismissed):
Awesome!

tkoyama010 (approved):

PR #6853, opened 2024-11-24
Fully annotate `transform.py` and `transformations.py`

user27182:
### Overview

I ran `ruff check --select ANN` on both files and added all missing annotations.

PR #6854, opened 2024-11-24
Add DataSet TypeVar to typing core

user27182:
### Overview

Add new TypeVar for datasets. It can be used in place of a Union (`PolyData | ImageData | ...`) or wherever a `T - > T` mapping is needed. E.g.  in #6853 it's used for overloading `Transform.apply` so that transforming `PolyData` will properly hint that `PolyData` is returned, etc., without needing a separate overload for each dataset subclass. A similar TypeVar is also used in #6840. 

And I think it can also be used to solve the typing issue we have where using `DataSet` as a type hint generates a type error if `vtkDataSet` is expected (since a `DataSet` is not a subclass of `vtkDataSet`), see https://github.com/pyvista/pyvista/pull/6828#discussion_r1854604818. This is solved by specifying all subclasses explicitly in the TypeVar since the subclasses _do_ inherit from `vtkDataSet`.  So, we can simply use `DataSetType_co` instead of `DataSet` as type hint.

### DIFF
--- a/doc/source/api/core/typing.rst
+++ b/doc/source/api/core/typing.rst
@@ -40,19 +40,12 @@ pyvista.VectorLike
 VTK Related Types
 -----------------
 
-pyvista.DataSetType
-~~~~~~~~~~~~~~~~~~~
+pyvista.ConcreteDataSetType
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. currentmodule:: pyvista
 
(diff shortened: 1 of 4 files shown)


user27182:
New type is documented:
https://67437e5c969e4000f32ea444--pyvista-dev.netlify.app/api/core/typing#pyvista-concretedatasettype
screenshot

tkoyama010 (approved):
LGTM

PR #6839, opened 2024-11-22
Add parameter annotations to `geometric_sources.py`

user27182:
### Overview

Add annotations to geometric sources.

Annotations and validation methods for a number of parameters (e.g. `center` and `direction`) were also standardized to make the overall API more consistent (i.e. allow setting with type `VectorLike[float]`, and return type `tuple[float, float, float]`)

PR #6855, opened 2024-11-24
Fully annotate `geometric_sources.py`

user27182:
### Overview

Follow-up to #6839. 

I ran `ruff check --select ANN pyvista/core/utilities/geometric_sources.py` then fixed all annotation errors that were reported. See https://docs.astral.sh/ruff/rules/#flake8-annotations-ann for the rules.

The main additions are: `self` is now annotated, `output` return types are now annotated, and any parameter annotations missed by #6839 are added.


tkoyama010 (approved):
LGTM

PR #6856, opened 2024-11-24
Fully annotate `arrays.py`

user27182:
### Overview

I ran `ruff check --select ANN  pyvista/core/utilities/arrays.py  --ignore ANN002 --ignore ANN003 --ignore ANN401` and added annotations and resolved any type errors. Some typing overloads are included. Validation logic was updated for `Lights.transform_matrix`.


pyvista/core/filters/data_set.py
```diff
@@ -1533,11 +1533,11 @@ def threshold(
         if scalars is None:
             set_default_active_scalars(self)  # type: ignore[arg-type]
             _, scalars = self.active_scalars_info  # type: ignore[attr-defined]
-        arr = get_array(self, scalars, preference=preference, err=False)
+        arr = get_array(self, scalars, preference=preference, err=False)  # type: ignore[arg-type]
```
user27182, line 1536: These new ignores are similar to the other `# type: ignore[attr-defined]` ignores used everywhere in DataSetFilters. DataSet is expected for `self` but `DataSetFilters` is given. #6870 should fix both the `# type: ignore[attr-defined]` and the new `# type: ignore[arg-type]` errors.

tkoyama010 (approved):
LGTM

PR #6861, opened 2024-11-25
[PROPOSAL] Add support for saving DataObjects directly as .vtkhdf files

krishanbhasin-px:
### Overview
Now that VTK 9.4.0 is out, with a `vtkHDFWriter`, lets add support for pyvista objects to write out to it

### Details

- I've gone through a handful of places where we are saving objects out to disk and, where they are supported by the `vtkHDFWriter`, added it as an allowable writer class.

I will go through again and add VTK version checks to make sure it doesn't try to make use of the new writer when the user has an older version of VTK


pyproject.toml
```diff
@@ -56,7 +56,7 @@ pinned = [ # Pinned versions of core dependencies
   'pooch<1.9.0',
   'scooby<0.11.0',
   'typing-extensions<4.13.0',
-  'vtk<9.4.0',
+  'vtk<=9.4.0',
```
banesullivan, line 59: let's hold off on this until #6731 is fully resolved

krishanbhasin-px: sounds good! Are you okay if I continue to work on the PR in the meantime, to get it ready to ship?
for eg: I need to do some VTK version checks, to avoid import errors if the user happens to have an old version of VTK

banesullivan: Yeah, totally fine to to keep working on this, I may just ask we block merging until VTK 9.4 support is ironed out

krishanbhasin-px: hey @banesullivan; I've rebased the PR over https://github.com/pyvista/pyvista/pull/6879 to benefit from the fixes you've put in, and to make sure the CI failures I would see are related to the change this PR is trying to introduce.

I've tried to put in some logic around handling different vtk versions but its really not fun trying to run the full CI locally; would it be possible for you to approve a CI run for this branch as it stands, to help me see where any remaining issues are?

user27182: I approved the workflows, this approval process seems new (they used to run automatically), will try to approve when more pushes are made

pyvista/core/pointset.py
```diff
@@ -51,6 +52,41 @@
     from ._typing_core import NumpyArray
     from ._typing_core import VectorLike
 
+if TYPE_CHECKING:
```
user27182, line 55: We can ignore missing test coverage for type-checks
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

pyvista/core/utilities/fileio.py
```diff
@@ -38,7 +38,13 @@
     from pyvista.core.utilities.reader import BaseReader
     from pyvista.plotting.texture import Texture
 
-_VTKWriterAlias = Union[_vtk.vtkXMLWriter, _vtk.vtkDataWriter]
+if TYPE_CHECKING:
+    _VTKWriterAlias = Union[_vtk.vtkXMLWriter, _vtk.vtkDataWriter, _vtk.vtkHDFWriter]
+else:
+    _VTKWriterAlias = Union[_vtk.vtkXMLWriter, _vtk.vtkDataWriter]
+    if hasattr(_vtk, 'vtkHDFWriter'):
+        _VTKWriterAlias = Union[_VTKWriterAlias, _vtk.vtkHDFWriter]
```
user27182, line 46: This else branch is only needed if the types are used at runtime. This is usually needed if using things like `typing.cast`. But, if the type is strictly used for type hints, then we can avoid runtime definitions and avoid the need to handle different vtk versions as a result. Unlike the runtime tests, the type-checking CI only uses the latest VTK version, so if `mypy` is successful on VTK 9.4 then we shouldn't need to bother with supported older versions for type-checking.

Suggest removing branch if runtime behavior is not required
```suggestion
```

pyvista/core/utilities/fileio.py
```diff
@@ -38,7 +38,13 @@
     from pyvista.core.utilities.reader import BaseReader
     from pyvista.plotting.texture import Texture
 
-_VTKWriterAlias = Union[_vtk.vtkXMLWriter, _vtk.vtkDataWriter]
+if TYPE_CHECKING:
```
user27182, line 41:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

pyvista/core/utilities/fileio.py
user27182: This requires using `from __future__ import annotations`, which will convert all annotations to strings which can then simply be read and interpreted by type-checkers instead importing the actual objects. All code in PyVista has this import statement.

``` python

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing import Sequence

def foo(x: Sequence) -> Sequence: ...

foo([42])
```

Our CI only does static type checking with the latest VTK version. This means that any conditional type-checking branches for older versions will _not_ be checked by CI. This is a maintenance concern and these conditional checks may break in the future without our knowledge. Unless there are compelling reasons to keep these then I think we should remove them.

If you really want to keep this conditional type definition then I suggest moving it fully outside of the `TYPE_CHECKING` block. That way, the runtime unit-tests will provide the necessary coverage (since those test using different VTK versions). You can add a `# noqa` comment in case `ruff` complains about this.

user27182: To clarify, I think this code as-is would have full test coverage, since there is only one static definition (for vtk9.4) and two runtime definitions (9.4 and <9.4). But, what I'm suggesting is removing the runtime definitions completely. 

Unless we _need_ them at runtime (e.g. if using `typing.cast`), then these definitions can be fully static (i.e. only defined inside the `TYPE_CHECKING` block). And, if we _did_ need them at runtime, then in this case we could remove the static definition (since the type-checking will still pick up all other runtime definitions). So, there is no need to define the type both statically _and_ at runtime.

krishanbhasin-px:
@tkoyama010 may I ask why this was closed? I was tracking https://github.com/pyvista/pyvista/pull/6879 so that I could get this in after it was merged

krishanbhasin-px:
ah was it auto-closed because the target branch was deleted? Normally I see GitHub transition the target over to the main branch of the repo when that happens 🤔

krishanbhasin-px:
No worries! I've opened https://github.com/pyvista/pyvista/pull/7396 which has the substantive changes from this PR 👍

Issue #5434, opened 2024-01-04
Typing for `pyvista_ndarray` is not compatible with `numpy.ndarray`


user27182:
### Describe the bug, what's wrong, and what you expected.

``` python
import pyvista as pv
import numpy as np
import numpy.typing as npt

x: npt.NDArray[np.float_]= np.array((2.0), dtype=np.float_)
reveal_type(x)  # Revealed type is "numpy.ndarray[Any, numpy.dtype[numpy.floating[Any]]]"


y = pv.pyvista_ndarray(x)
reveal_type(y)  # Revealed type is "pyvista.core.pyvista_ndarray.pyvista_ndarray"

y += x  # mypy error: Result type of + incompatible in assignment  [misc]


```

This code is valid at runtime, but mypy complains about the `+` operation. I believe the issue that whereas the `.pyi` stub for `ndarray` defines all the overloads for the different allowed types for all array magic methods, these static type hint overloads aren't necessarily inherited or defined by subclasses such as `pyvista_ndarray` (i.e. type hints from stub files are not inherited).

This also means that using generics like `pyvista_ndarray[Any, np.dtype[np.float_]]` for type hints, in the same manner as one might do `np.ndarray[Any, np.dtype[np.float_]]` also does not work.

A workaround for now is to simply tell mypy to ignore the error, but support for explicit `pyvista_ndarray` dtype hints may be useful in the future since `pyvista_ndarray` is tied to things like `dataset.points`, which is a commonly used array for operations. Therefore, this may become a larger issue as dtype hints for `np.ndarray` are added to the codebase.

Another solution may be to customize a mypy plugin to explicitly tell mypy to treat `pyvista_ndarray` in (almost) the exact same manner as `np.ndarray`, e.g. using type promotion similar to what's done in https://github.com/pyvista/npt-promote.

### Steps to reproduce the bug.

see above

### System Information

```shell
--------------------------------------------------------------------------------

                OS : Darwin
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : Apple
      GPU Renderer : Apple M2
       GPU Version : 4.1 Metal - 83.1
  MathText Support : True
  Python 3.10.11 (v3.10.11:7d4cc5aa85, Apr  4 2023, 19:05:19) [Clang 13.0.0
  (clang-1300.0.29.30)]
           pyvista : 0.43.1
               vtk : 9.3.0
             numpy : 1.26.3
[...]

PR #6870, opened 2024-11-26
Add `self` annotations to dataset filters

user27182:
### Overview

There is a known typing issue with the dataset filters. It's a mixin class,  so `self` should be a dataset, but a dataset filter is expected. This generates a lot of type errors for undefined attributes or invalid args. See also https://github.com/pyvista/pyvista/issues/6279#issuecomment-2183191197.

One solution from #4774 is to use protocols. In fact, using a protocol class is _required_ by mypy for mixin classes to be typed correctly. See https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes. But, this requires a lot of restructuring. 

Instead, this PR proposes annotating `self` as needed, e.g. as vtkDataSet, MultiBlock, ImageData, whatever. This resolves all of the type errors. The caveat is that, since we aren't using protocols,  this annotation is itself a type error. So, this requires `# type: ignore[misc]`. It's not perfect, but this way we only need to ignore a single type error in the function signature, rather than a bunch of type errors in the function body.


user27182:
Looking for some input from @MatthewFlamm and @akaszynski on this. Do you like the idea of annotating `self` as a dataset type to fix the attribute/arg type errors and use a single `# type: ignore[misc]` in the function signature instead?

tkoyama010 (approved):
LGTM

tkoyama010 (changes requested):
Waiting review of others.

pyvista/core/filters/data_set.py
```diff
@@ -47,8 +48,10 @@
 class DataSetFilters:
     """A set of common filters that can be applied to any vtkDataSet."""
 
-    def _clip_with_function(
-        self,
+    points: pyvista.pyvista_ndarray
```
user27182, line 51: This line is needed to resolve a bunch of type errors caused by #5434.

If we remove this line, it generates errors like:
``` python
pyvista/core/filters/poly_data.py:2838: error: Cannot determine type of "points"  [has-type]
            np.apply_along_axis(f, 1, mesh.points)
```
``` python
pyvista/core/dataset.py:439: error: Signature of "points" incompatible with supertype "DataSetFilters"  [override]
        @property
         ^
pyvista/core/dataset.py:439: note:      Superclass:
pyvista/core/dataset.py:439: note:          ndarray[Any, dtype[floating[_32Bit]]]
pyvista/core/dataset.py:439: note:      Subclass:
pyvista/core/dataset.py:439: note:          pyvista_ndarray
```

It is a bit odd to have this type hint here though. Alternatively, we can remove it and simply `# type: ignore` the individual errors on a per-line basis.

user27182: The hint was removed in [d5ab232](https://github.com/pyvista/pyvista/pull/6870/commits/d5ab232c147dad2696251ad422fb2cf012a8fc3f) and type ignores added instead. This is better, because if https://github.com/pyvista/pyvista/issues/5434 ever gets fixed, then mypy should auto-detect that the ignores are no longer needed at that time. By contrast, if we keep the property's type hint we won't get notified by mypy that it's redundant (and we won't know to remove it later).

akaszynski:
Tentative yes to your question, @user27182, but give me a bit to review. Cooking a turkey for the next few hours.

MatthewFlamm (approved):
I think it would be best long term to either make a protocol or redo the inheritance model (this would be my preference).  But in the meantime, it makes sense to me to type what the intention is, so I'm +1.

banesullivan (approved):
+1 for looking into better long term solutions to this inheritance model or moving towards Protocols. In the meantime, I think these type annotations are a great benefit

pyvista/core/filters/data_set.py
```diff
@@ -1755,7 +1762,9 @@ def outline(self, generate_faces: bool = False, progress_bar: bool = False):
         _update_alg(alg, progress_bar, 'Producing an outline')
         return wrap(alg.GetOutputDataObject(0))
 
-    def outline_corners(self, factor=0.2, progress_bar: bool = False):
+    def outline_corners(  # type: ignore[misc]
+        self: _vtk.vtkDataSet | _vtk.vtkMultiBlockDataSet, factor=0.2, progress_bar: bool = False
```
user27182, line 1766: Adding a note here for completeness (not intended to block approval).

This is a tricky filter to type since the internal vtk filters allows vtkDataObject, not just vtkDataSet.
It would probably be better to use a new `ConcreteDataObjectType` type here (similar to `ConcreteDataSetType` added in https://github.com/pyvista/pyvista/pull/6854). Otherwise, if a `DataObject` is passed to this function a false positive type error will be generated (since `DataObject` does not inherit from `vtkDataObject`).

But, this would also (perhaps unfavorably) narrow the type to pyvista DataObjects only, even though in this case any vtkDataObject is allowed, resulting in false positive type errors for unwrapped vtk classes.

Perhaps it may be best to do `self: DataObject | vtkDataObject` and then wrap `self` in the filter. This general idea of typing `vtk | pyvista` inputs and wrapping `self` could apply to all filters. And it would help remove the use of `ConcreteDataSetType` or `ConcreteDataObjectType` as a type hint, which is good because this is more of an internal typing construct rather than something that users should have to think about.

These subtle typing changes can be made in a follow-up PR though. The main goal for this PR is to add initial annotations and get approval for the `type: ignore[misc]` idea generally.

banesullivan (approved):

tkoyama010 (approved):

Issue #6858, opened 2024-11-25
Dependabot not updateing [dev] install option: test and docs conflict


banesullivan:
### Describe what maintenance you would like added.

I just ran `pip install -e '.[dev]'` and ran into the error:

```
INFO: pip is looking at multiple versions of pyvista[docs,test] to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install pyvista[docs,test]==0.45.dev0 because these package versions have conflicting dependencies.

The conflict is caused by:
    pyvista[docs,test] 0.45.dev0 depends on trame-vtk<2.8.12 and >=2.5.8; extra == "test"
    pyvista[docs,test] 0.45.dev0 depends on trame-vtk==2.8.12; extra == "docs"

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

This seems to have been introduced by #6845 which updated the `test` section but not the constraint set in the `test` section. We need to make sure that dependabot is updating all of these sections together.

cc @user27182 

### Links to source code.

https://github.com/pyvista/pyvista/blob/c4abcdb201bc4359389202f3b608105eedd17484/pyproject.toml#L96

https://github.com/pyvista/pyvista/blob/c4abcdb201bc4359389202f3b608105eedd17484/pyproject.toml#L135

### Pseudocode or Screenshots

n/a

banesullivan:
Further, we should make sure the CI somewhere runs `pip install -e '.[dev]'` to validate any automated changes from dependabot

banesullivan:
Or maybe the upper constraint should be on the next major/minor version not a patch release

MatthewFlamm:
I at first thought that the less than pin was the problem, but this slightly [older update](https://github.com/pyvista/pyvista/commit/636a578208e9600ecd280f3681fdaa90011b341c) did exactly what was expected.  This smells a bit like a dependabot bug?



Either way your comment about making sure we test each install group would prevent bugs in dependabot from creeping in here.

MatthewFlamm:
I also checked that we didn't have some unintended 'ignore' condition for dependabot, so this likely wasn't a factor

https://github.com/pyvista/pyvista/pull/6845#issuecomment-2498158963

user27182:
[...]

PR #6872, opened 2024-11-26
Limit VTK<9.4.0

banesullivan:
Upper limit of the version of VTK that is supported. We are working on full support for VTK 9.4 and have some base level of support at present but overall VTK 9.4 brought a lot of breaking changes that we still need to resolve.

for #6731

To slow down the reports of this incompatibility, I will cherry pick this commit on to the release branch and issue a patch

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
   'pooch',
   'scooby>=0.5.1',
   'typing-extensions',
-  'vtk',               # keep without version constraints
+  'vtk<9.4.0',
 ]
 description = 'Easier Pythonic interface to VTK'
 dynamic = ['version']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,6 @@ pinned = [ # Pinned versions of core dependencies
   'pooch<1.9.0',
   'scooby<0.11.0',
   'typing-extensions<4.13.0',
-  'vtk<9.4.0',
 ]
 
 typing = [


MatthewFlamm (approved):
I'm +1 for this, and I also think that we should keep this pin going forward.  The pin will be upgraded accordingly when new vtk versions are released and we support them.

banesullivan:
While I've been hesitant and vocal about that hesitation in the past to prevent any pins on VTK, I think we are at a point where we just need to in order to maintain a level of stability and reliability. Especially since we are getting more and more adoption with last month being over 350,000 downloads (ref https://pypistats.org/packages/pyvista)

akaszynski:
I think we should enforce an upper limit with vtk always, even for patches. VTK has backwards breaking changes for their patch releases, and it would help with bug reports to only support what we’ve tested. We can always push a patch to support the latest version of VTK.

For what it’s worth, the lack of an upper limit broke my closed source builds yesterday and it took me half an hour to figure out it was VTK 9.4 causing the issue.

user27182:
If VTK is permanently pinned in the required dependencies then it's redundant to also pin it in `pinned` (and may cause issues if dependabot does not update both, e.g. #6858)

suggestion:
``` diff
- pinned = [ # Pinned versions of core dependencies
+ pinned = [ # Pinned versions of core dependencies (which are not already pinned)
   'matplotlib<3.9.3',
   'numpy<2.2.0',
   'pillow<11.1.0',
   'pooch<1.9.0',
   'scooby<0.11.0',
   'typing-extensions<4.13.0',
-  'vtk<9.4.0',
  ]
```

banesullivan:
Ah good catch @user27182. I just removed vtk from the `pinned` group

pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
   'pooch',
   'scooby>=0.5.1',
   'typing-extensions',
-  'vtk',               # keep without version constraints
+  'vtk<9.4.0',
```
user27182, line 27: Based on https://github.com/pyvista/pyvista/pull/6872#issuecomment-2501547036, maybe it's better to do 
```suggestion
  'vtk<9.3.2',
```

MatthewFlamm (approved):

user27182 (commented):
LGTM, though it may be good to restrict patch updates too. However, pinning the minor version only (i.e. 'vtk<9.4.0'), may be a good compromise.

pyproject.toml
banesullivan: I think restricting the minor version is best for now. Limiting the patch release feels a little too restrictive to me, but I'm open to do this.

In the meantime, I vote we release this limit on 9.4 and come back to this

MatthewFlamm: Since the current cycle is a minor release it is moot for the near term use, the next update to this line will either be `<9.4.1` or `<9.5.0`.  We will have to decide at that point.  If vtk has breaking changes, Id also lean towards pinning down to the patch level.

banesullivan: While VTK regularly issues breaking changes and regressions across minor releases, I feel this doesn't happen so often during patch releases (though I don't have the data on hand to support this feeling)

user27182: Maybe keep vtk in `pinned` then and do `'vtk<9.3.2'` there? I.e. adopt a convention of pinning vtk to minor releases generally, but pin vtk to patch releases for testing. Something to consider.

MatthewFlamm: If we are unsure,  then I'd vote to pin minor versions only.  If major breaking changes are found in patch releases, then we pin patch releases going forward too.

bjlittle:
I think adopting this controlled pinning strategy for vtk makes a lot of sense, both for PyPI and conda 👍

Since we're in this space, are there any other key packages in the PyVista stack, akin to vtk, that merit the same treatment?

akaszynski:
> I think adopting this controlled pinning strategy for vtk makes a lot of sense, both for PyPI and conda 👍
> 
> Since we're in this space, are there any other key packages in the PyVista stack, akin to vtk, that merit the same treatment?

I don't think so. VTK is our most important dependency and it also is the also a hard moving target since they add so many features between releases and often their bug fixes have unintended consequences.

I'm also in favor of limiting only when absolutely necessary. Something like `numpy<3` is fine for me because it's going to likely be another 5-10 years before they upgrade, and when they do we will have to update our library.

PR #6873, opened 2024-11-26
Fully annotate `dataset.py`

user27182:
### Overview

Add annotations such that `ruff check pyvista/core/dataset.py --select ANN --ignore ANN401 --ignore ANN003 --ignore ANN002` passes.


tkoyama010 (approved):
LGTM

PR #6874, opened 2024-11-26
Fully annotate `composite.py`

user27182:
### Overview

Add annotations. Running `ruff check pyvista/core/composite.py --select ANN --ignore ANN002 --ignore ANN003` passes.


tkoyama010 (approved):
LGTM

PR #6879, opened 2024-11-27
VTK 9.4 compatibility

banesullivan:
Work in progress for #6731

- VTK decided to enable their [new Python wrappings](https://gitlab.kitware.com/vtk/vtk/-/blob/master/Wrapping/Python/vtkmodules/util/data_model.py?ref_type=heads) (their version of PyVista's core API) by default and override no matter what. This is what caused https://github.com/pyvista/pyvista/issues/6731#issuecomment-2452644747 and many cascading errors as VTK would unexpectedly return their new data_model classes. To resolve this, we have to disable the override of these VTK classes altogether until some of the issue with the mechanism are resolved upstream. I wish VTK didn't enable this by default... but they did [seemingly intentionally here I believe](https://gitlab.kitware.com/vtk/vtk/-/blob/ed64ffc627e08c2c43a19fc87665941996159576/Wrapping/Python/vtkmodules/__init__.py.in#L113-117).
- ~We have to provide a dummy wrapping of `ImageData.cells` since VTK is now returning `vtkStructuredCellArray` which doesn't seem too useable from my inspection so far. I'm just not sure how to productively use this object so for now, just return an empty array since it has to be used in our comparison logic... alternative would be to reimplement the `__eq__` logic~ `DataObject.__eq__` has been modified as part of #7360 
- Fix issues with Charts wrapping in #7189
- ~Increase maxclients for xserver for our CI testing in #7214 (see comments below for explanation). This is tracked upstream in https://gitlab.kitware.com/vtk/vtk/-/issues/19618~ No longer needed since https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11985 was fixed with vtk 9.4.2
- New overrides for `faces`, `polyhedron_faces`, `face_locations`, and `polyhedron_face_locations` for  polyhedral `UnstructuredGrid` meshes
- Ensure our wrapping of `AxesActor`, `CubeAxesActor`, and `_BaseMapper` implement `bounds` and `center` properties to avoid new upstream wrapped properties for these
- Work around an upstream issue with the axes widget (tracked in https://gitlab.kitware.com/vtk/vtk/-/issues/19592) where turning on/off results in visual artifacts. See the refactoring of `delete_axes_widget()` and any usage of `axes_widget.EnabledOff()`
- New `is_vtk_attribute` function, useful for checking if a `snake_case` attribute is defined by a vtk class or not
- Set default render window for OSMesa with VTK 9.4 (#7373)
- Fix`charts.Axis` pen attribute documentation for VTK 9.4 #7374
- Fix pointset and subdivide examples broken by VTK 9.4 (#7371)
- Update doc image cache for VTK 9.4 (#7376)

Issues to address later:

- ~We are skipping the `test_flip_normals` test due to a failure in VTK 9.4. Not sure why~
- Ensure `to_meshio` supports polyhedral unstructured grids with VTK 9.4 (#7130)
- `test_camera3d_widget` is failing on MacOS CI with a seemingly different camera position (see https://github.com/pyvista/pyvista/pull/6879#issuecomment-2692519662)
- Remove the axes widget hack (issue was fixed in vtk 9.4.2)
- Need to update the test matrix to include testing for VTK 9.3.1 (this is done along with python 3.13 support in #7366)
- PyVistQt integration tests are flaky and should be reviewed and/or updated
----

FYI, a good chunk of the initial compatibility was implemented in #5650


pyvista/core/_vtk_core.py
```diff
@@ -507,3 +507,46 @@ def VTKVersionInfo():
...
+
+
+class vtkPyVistaOverride:
+    """Base class to automatically override VTK classes with PyVista classes."""
+
+    # TODO: `MutableSequence` base class is breaking the VTK override mechanism
+    VTK_OVERRIDE_EXCLUSIONS = (
+        'vtkMultiBlockDataSet',
+        'vtkPartitionedDataSet',
+    )
```
banesullivan, line 537: I think we need to raise an upstream issue for when `MutableSequence` is in the base classes it breaks the override mechanism in https://github.com/Kitware/VTK/blob/master/Wrapping/PythonCore/PyVTKObject.cxx

banesullivan: Posted upstream in https://gitlab.kitware.com/vtk/vtk/-/issues/19545

banesullivan: For these we should just call `.override(None)` to unregister the parent VTK classes

banesullivan: A fix is inbound here: https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11698

jourdain:
Side note: `ImageData.cells` should not be part of `__eq__` as the cells are implicit and defined by the extent. So a comparing the extent is enough.

banesullivan:
> Side note: ImageData.cells should not be part of __eq__ as the cells are implicit and defined by the extent. So a comparing the extent is enough.

Absolutely, this `__eq__` operator needs to be rewritten more robustly and/or provide unique comparisons for each data type. For now, having `.cells` return a VTK reference object doesn't make sense here and thus the empty list. This needs to be cleaned up later but shouldn't block this effort to provide compatibility with VTK 9.4

user27182:
> Side note: `ImageData.cells` should not be part of `__eq__` as the cells are implicit and defined by the extent. So a comparing the extent is enough. 

Just a heads up that this has been updated in #7075 in case this affects this PR

cquammen:
Hi @banesullivan is there anything we can do to help push this over the finish line? Thanks in advance!

pyvista/plotting/renderer.py
```diff
@@ -974,6 +974,18 @@ def add_axes_at_origin(
...
+        """Remove and delete the current axes widget."""
+        if hasattr(self, 'axes_widget'):
+            self.axes_actor = None
+            # HACK: set the viewport to a tiny value to hide the widget first
+            # This is due to an issue with a blue box appearing after removal
+            # Tracked in https://gitlab.kitware.com/vtk/vtk/-/issues/19592
+            self.axes_widget.SetViewport(0.0, 0.0, 0.0001, 0.0001)
+            self.axes_widget.EnabledOff()
+            self.Modified()
+            del self.axes_widget
```
jspanchu, line 987: I propose a less hacky workaround until https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11874 lands.

```python
            self.axes_actor = None
            # HACK: do not disable the widget before deleting it.
            # This is due to an issue with the older widget Z axis shaft
            # appearing after removal
            # Tracked in https://gitlab.kitware.com/vtk/vtk/-/issues/19592
            self.axes_widget.SetViewport(0.0, 0.0, 0.0001, 0.0001)
            self.Modified()
            del self.axes_widget
```

pyvista/core/pointset.py
```diff
@@ -2047,6 +2048,42 @@ def cells(self, cells) -> None:
...
+            Array of face locations.
+
+        """
+        if hasattr(self, 'GetPolyhedronFaceLocations'):
+            faces = self.GetPolyhedronFaceLocations()  # vtkCellArray
+            if faces is not None:
+                faces = faces.GetData()  # vtkCellArray -> vtkIdTypeArray
+        else:
+            faces = self.GetFaceLocations()  # vtkCellArray
+        return convert_array(faces)
```
banesullivan, line 2085: These return totally different results between <9.4 and 9.4+. Not sure how to handle this and its proving incompatible with the meshio conversion code from #7079

@jspanchu or @cquammen, are you familiar with the polyhedral cell storage layout changes and how we can correct the usage in the `to_meshio` conversion utility?

cc @keurfonluu

tests/typing/test_return_type.py
```diff
@@ -25,10 +26,11 @@ def _get_classes_from_module(module: ModuleType) -> list[type]:
...
             except TypeError:
-                pass  # not a class
-            else:
-                if hasattr(module_attr, attr):
-                    class_types.append(module_attr)
+                continue  # not a class
+            if issubclass(module_attr, (_vtk.vtkAxesActor, _vtk.vtkTextActor)):
+                return  # Skip these classes
+            if hasattr(module_attr, attr):
+                class_types.append(module_attr)
```
banesullivan, line 33: @user27182, this was proving to be troublesome this would now yield what I might consider false positives. For example, `vtkAxesActor` and `vtkTextActor` both have `GetBounds()` and `GetCenter()` methods that return `None` and for that reason we don't wrap these methods in PyVista. However with VTK 9.4's new properties, `.bounds` and `.center` get mapped to thee methods and the test here is not valid for those wrappings. Do you think this is an appropriate workaround?

banesullivan:
We're getting close!! roughly ~11-15 tests remaining failing out of the over 8,000 tests

tests/typing/test_return_type.py
user27182: I'm not sure I understand. What is the desired or expected behavior?



> `vtkAxesActor` and `vtkTextActor` both have `GetBounds()` and `GetCenter()` methods that return `None`

is this vtk 9.4 behavior? on 9.3 I get:
``` python
import vtk 
vtk.vtkAxesActor().GetBounds()
# (-1.0, 1.0, -1.0, 1.0, -1.0, 1.0)
```

But in pyvista we don't define `bounds`

``` python
import pyvista as pv
pv.AxesActor().bounds
# AttributeError: 'AxesActor' object has no attribute 'bounds'
```

Without any knowledge of the failures you're seeing with tests or an understanding of the new vtk 9.4 behavior, it seems to me that we should probably define a `AxesActor.bounds` property, or am I missing something?

banesullivan:
```py
>>> import vtk
>>> vtk.vtkTextActor().GetBounds()
>>> vtk.vtkAxesActor().GetBounds()
(-1.0, 1.0, -1.0, 1.0, -1.0, 1.0)
>>> 
```

I mistakenly only checked `vtkTextActor` and thought `vtkAxesActor` was also returning None but I may be getting different failures there.

> But in pyvista we don't define bounds

Correct, but VTK 9.4 implements a new wrapping mechanism that creates new properties like `bounds` which wrap `GetBounds()`. So even though we don't implement `bounds` it is now a property in VTK 9.4 and picked up by this dynamic parameterization. I'm going to just implement `bounds` and `center` properties for our wrapping of `vtkAxesActor` but this will still be an issue for `vtkTextActor`

banesullivan: Okay ignore `vtkAxesActor`. That is now fixed by implementing our own `bounds` and `center` properties.

user27182: I see. So there is now a confusing (and unavoidable?) blending of `vtk`'s API with our API. Our API previously promised that `bounds` returns a tuple (else raise `AttributeError`), but now it seems there is a risk that the return type for some objects is `tuple | None` if we haven't explicitly defined it ourselves...

Maybe we should override `vtkTextActor.bounds` to raise `NotImplementedError` instead of returning None?

pyvista/core/pointset.py
jspanchu: the changes to storage are documented in https://docs.vtk.org/en/latest/release_details/9.4/vtk-polyhedral-cells-storage.html. Since it's only a deprecation, you could address it later by adapting the meshio conversion to handle new design.

keurfonluu:
HI @banesullivan,

Sorry to bother you with my polyhedral meshes (I am dealing with Voronoi meshes in my work). It seems that VTU files with polyhedral cells written with VTK 9.4 do not work properly when opened in ParaView.

Taking the example from here: https://docs.pyvista.org/examples/00-load/create-polyhedron.html, as you can see, the polyhedron is missing when saved with VTK 9.4 (it works when saved with VTK 9.3).

```python
import pyvista as pv

quad_points = [
    [0.0, 0.0, 0.0],  # 0
    [0.0, 0.01, 0.0],  # 1
    [0.01, 0.01, 0.0],  # 2
    [0.01, 0.0, 0.0],  # 3
]
polygon_points = [
    [0.02, 0.0, 0.0],  # 4
    [0.02, 0.01, 0.0],  # 5
    [0.03, 0.01, 0.0],  # 6
    [0.035, 0.005, 0.0],  # 7
    [0.03, 0.0, 0.0],  # 8
]
hexa_points = [
    [0.0, 0.0, 0.02],  # 9
    [0.0, 0.01, 0.02],  # 10
    [0.01, 0.01, 0.02],  # 11
    [0.01, 0.0, 0.02],  # 12
    [0.0, 0.0, 0.03],  # 13
    [0.0, 0.01, 0.03],  # 14
    [0.01, 0.01, 0.03],  # 15
    [0.01, 0.0, 0.03],  # 16
]
polyhedron_points = [
    [0.02, 0.0, 0.02],  # 17
    [0.02, 0.01, 0.02],  # 18
    [0.03, 0.01, 0.02],  # 19
    [0.035, 0.005, 0.02],  # 20
    [0.03, 0.0, 0.02],  # 21
    [0.02, 0.0, 0.03],  # 22
    [0.02, 0.01, 0.03],  # 23
    [0.03, 0.01, 0.03],  # 24
    [0.035, 0.005, 0.03],  # 25
    [0.03, 0.0, 0.03],  # 26
]
points = quad_points + polygon_points + hexa_points + polyhedron_points
quad = [4, 0, 1, 2, 3]
polygon = [5, 4, 5, 6, 7, 8]
hexa = [8, 9, 10, 11, 12, 13, 14, 15, 16]
polyhedron_connectivity = [
    # NItems will go here
    7,  # number of faces
    5,  # number of points in face0
    17,  # point index 0
    18,  # point index 1
    19,  # point index 2
    20,  # point index 3
    21,  # point index 4
    4,  # number of points in face1
    17,  # point index ...
    18,
    23,
    22,
    4,
    17,
    21,
    26,
    22,
    4,
    21,
    26,
    25,
    20,
    4,
    20,
    25,
    24,
    19,
    4,
    19,
    24,
    23,
    18,
    5,
    22,
    23,
    24,
    25,
    26,
]
polyhedron = [len(polyhedron_connectivity), *polyhedron_connectivity]
cells = quad + polygon + hexa + polyhedron
celltypes = [pv.CellType.QUAD, pv.CellType.POLYGON, pv.CellType.HEXAHEDRON, pv.CellType.POLYHEDRON]
grid = pv.UnstructuredGrid(cells, celltypes, points)
grid.save('mesh.vtu')
```

With VTK 9.4
image

With VTK 9.3
image

user27182:
> Sorry to bother you with my polyhedral meshes (I am dealing with Voronoi meshes in my work). It seems that VTU files with polyhedral cells written with VTK 9.4 do not work properly when opened in ParaView.
> 
> Taking the example from here: https://docs.pyvista.org/examples/00-load/create-polyhedron.html, as you can see, the polyhedron is missing when saved with VTK 9.4 (it works when saved with VTK 9.3). For other polyhedral meshes, ParaView simply crashes.

I just tried this out by:

- Checking out this branch [maint/vtk-override](https://github.com/pyvista/pyvista/tree/maint/vtk-override)
- Running the example code provided
- Opening the saved mesh `mesh.vtu` in Paraview 5.13.2

And I'm getting the expected output:

paraview_

What version of Paraview are you using? Looking at the latest release notes https://www.kitware.com/paraview-5-13-2-release-notes/ it seems some issues with polyhedra were fixed

keurfonluu:
> What version of Paraview are you using? Looking at the latest release notes https://www.kitware.com/paraview-5-13-2-release-notes/ it seems some issues with polyhedra were fixed

I am running ParaView 5.13.2 as well (on Windows). I also tried the branch maint/vtk-override and it still does not work on two different computers.

Attached the file generated (zipped), which can be read back with PyVista and correctly plotted.
[mesh.zip](https://github.com/user-attachments/files/18562790/mesh.zip)

user27182:
> I just tried this out by:
> 
> * Checking out this branch [maint/vtk-override](https://github.com/pyvista/pyvista/tree/maint/vtk-override)
> * Running the example code provided
> * Opening the saved mesh `mesh.vtu` in Paraview 5.13.2
> 
> And I'm getting the expected output:

I forgot to actually install vtk 9.4 here 😅. Installing vtk 9.4.1 before saving the mesh does indeed reproduce this issue, and the mesh does not load correctly in ParaView 5.13.2.

tests/typing/test_return_type.py
banesullivan: > So there is now a confusing (and unavoidable?) blending of vtk's API with our API.

Yes, except if we override these properties then ours will always be used (well after this PR since I disable VTK's override mechanism in this PR). Most of this was handled in my original 9.4 compatibility work to nail down the method resolution order in #5650

> ... if we haven't explicitly defined it ourselves...
exaclty, thats the key here.

> Or maybe return uninitialized bounds like (1.0, -1.0, 1.0, -1.0, 1.0, -1.0) where min > max?

I think this is the most appropriate.

banesullivan:
@keurfonluu and @user27182, I've reverted this code to use the legacy GetFaces and GetFaceLocations APIs and indeed it is unable to load into ParaView for me too. I'm going to mark the to_meshio utility as incompatible with VTK 9.4 and we'll have to resolve this later so as to not block the remaining upgrade to 9.4

keurfonluu:
> @keurfonluu and @user27182, I've reverted this code to use the legacy GetFaces and GetFaceLocations APIs and indeed it is unable to load into ParaView for me too. I'm going to mark the to_meshio utility as incompatible with VTK 9.4 and we'll have to resolve this later so as to not block the remaining upgrade to 9.4

It's a VTK issue, not PyVista. I opened a thread on VTK's Discourse: https://discourse.vtk.org/t/polyhedra-saved-in-a-vtu-file-written-with-vtk-9-4-are-not-visible-in-paraview/15220

keurfonluu:
> It's a VTK issue, not PyVista. I opened a thread on VTK's Discourse: https://discourse.vtk.org/t/polyhedra-saved-in-a-vtu-file-written-with-vtk-9-4-are-not-visible-in-paraview/15220

Not a VTK issue, but a ParaView issue :D it works with the nightly version of ParaView 5.13.
The VTU files produced with VTK 9.3 is versioned 0.1, while that of VTK 9.4 is versioned 2.3.

beroda:
> In the final stretch! All that remains are:
> 
> 
>     * Dealing with a segfault on linux coming from `test_plot_volume_ugrid` I think
> 
FWIW, I tried to look into the segfault for Linux. Here are my findings:

- It only depends on vtk version. I managed to get the segfault on both python 3.12 and 3.11.
- It is not attached to a specific test. I tried to randomly order tests using `pytest-random-order`, and it systematically fails for different tests.
- The failure seems to occur when instantiating multiple  `vtkRenderWindow` during the tests. The issue seems to be coming from vtk's side since this small test is systematically failing at the 215-th test
```python
from vtk import vtkRenderWindow
import pytest

@pytest.mark.parametrize('i', range(300))
def test_foo(i):
    r = vtkRenderWindow()
```

There might be some kind of memory leak or something else, but I am afraid this is beyond my current knowledge.
I plan to run git bisect by building local vtk wheels, but it might take some time.

user27182:
> * VTK decided to enable their [new Python wrappings](https://gitlab.kitware.com/vtk/vtk/-/blob/master/Wrapping/Python/vtkmodules/util/data_model.py?ref_type=heads) (their version of PyVista's core API) by default and override no matter what. This is what caused [VTK 9.4 Incompatibility Tracker #6731 (comment)](https://github.com/pyvista/pyvista/issues/6731#issuecomment-2452644747) and many cascading errors as VTK would unexpectedly return their new data_model classes. To resolve this, we have to disable the override of these VTK classes altogether until some of the issue with the mechanism are resolved upstream.

IMO the overrides should be intentionally disabled for all classes, and not just for a few classes with known issues or failing tests. And possibly disabled permanently by design. When using `PyVista`, or even for internal PyVista code, there has always been a clear communication to users or developers that anything with `snake_case` is executing PyVista code, and anything with `CamelCase` is executing VTK code directly. Having both use `snake_case` not only makes this ambiguous but it could also make things confusing in cases where our API diverges from VTK's. E.g. the `pyvista.Transform` class  deliberately uses `invert()` instead of vtk's naming of `vtkTransform.Inverse()`. But with the overrides enabled auto-complete will now suggest both `invert` and `inverse` to users. This kind of mixing should really be avoided.

pyvista/core/_vtk_core.py
```diff
@@ -584,3 +584,27 @@ def VTKVersionInfo():
...
+                    and base.__module__.startswith('vtkmodules.')
+                    and hasattr(base, 'override')
+                ):
+                    # For now, just remove any overrides for these classes
+                    # There are clear issues with the current implementation
+                    # of overriding these classes upstream and until they are
+                    # resolved, we will entirely remove the overrides.
+                    # See https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11698
+                    # See https://gitlab.kitware.com/vtk/vtk/-/issues/19550#note_1598883
+                    base.override(None)
```
user27182, line 607: This seems to only be a partial fix, and may only be disabling overrides for child classes. E.g. we can still access things like `vtkObject.global_warning_display`:

``` python
import pyvista as pv
poly = pv.PolyData()
poly.global_warning_display  # Returns `1`
```

I've tried calling `base.override(None)` recursively up the chain but that also still doesn't get rid of these attributes.

The [vtk docs on overrides](https://docs.vtk.org/en/latest/advanced/PythonWrappers.html#experimental-features) state:
> The override can be reversed by setting an override of None, but this will not impact instantiations that have already occurred.

So maybe the parent classes are already initialized at this point, which is why `override(None)` doesn't work for all attributes. Maybe we need to override `__getattribute__` and dynamically intercept these override attributes? Or maybe we can delete them all with `delattr` ?

banesullivan: The override mechanism is separate from the new Python wrappings -- these snake case properties are now a part of VTK's Python wrapping and as far as I am aware there is no way to opt-out of these. 

This is why getting the method resolution order in https://github.com/pyvista/pyvista/pull/5650 was so important

This blog has some details: https://www.kitware.com/vtk-9-4-a-step-closer-to-the-ways-of-python/

(sorry for the quick/short comment, pressed for time lately but I'm happy to explain further)

user27182: > The override mechanism is separate from the new Python wrappings -- these snake case properties are now a part of VTK's Python wrapping and as far as I am aware there is no way to opt-out of these.

Thanks for clarifying. An attempt to deal with the VTK snake_case API issue is given in #7198

beroda:
> > In the final stretch! All that remains are:
> > ```
> > * Dealing with a segfault on linux coming from `test_plot_volume_ugrid` I think
> > ```
> 
> FWIW, I tried to look into the segfault for Linux. Here are my findings:
> 
>     * It only depends on vtk version. I managed to get the segfault on both python 3.12 and 3.11.
> 
>     * It is not attached to a specific test. I tried to randomly order tests using `pytest-random-order`, and it systematically fails for different tests.
> 
>     * The failure seems to occur when instantiating multiple  `vtkRenderWindow` during the tests. The issue seems to be coming from vtk's side since this small test is systematically failing at the 215-th test
> 
> 
> ```python
> from vtk import vtkRenderWindow
> import pytest
> 
> @pytest.mark.parametrize('i', range(300))
> def test_foo(i):
>     r = vtkRenderWindow()
> ```
> 
> There might be some kind of memory leak or something else, but I am afraid this is beyond my current knowledge. I plan to run git bisect by building local vtk wheels, but it might take some time.

git bisect found this commit to be the origin of the issue: https://gitlab.kitware.com/vtk/vtk/-/commit/04661f4bcdf84195a75e4e91a542b9e6e8bd8e37

It seems that the maximum of clients for the xserver is reached when running the tests, thereby causing the segfault. Interestingly, increasing the maximum number of clients following [this post](https://www.antixforum.com/forums/topic/please-increase-x-server-maxclients/#post-50154) helps but the segfault is only delayed but not removed.

If you agree, I plan to open an issue in vtk.

banesullivan:
Should we just bum the max clients to 2048 like that post suggests to at least merge this PR and start supporting VTK 9.4? I agree this is just a bandaid and delays this issue.

An alternative solution to explore would be to do some sort of test splitting with CI so that we spin up several individual nodes to handle a chunk of tests individually -- out of the scope of this PR in my opinion and should come later

jspanchu:
> Should we just bum the max clients to 2048 like that post suggests to at least merge this PR and start supporting VTK 9.4? I agree this is just a bandaid and delays this issue.
> 

Yes, for now this seems a good idea. I will look into a fix for the max no. of x clients in VTK.

jspanchu:
Tracked upstream at https://gitlab.kitware.com/vtk/vtk/-/issues/19618. We will aim to fix it by 9.4.2 (Feb 28)

user27182:
> Should we just bum the max clients to 2048 like that post suggests to at least merge this PR and start supporting VTK 9.4? I agree this is just a bandaid and delays this issue.

Running the plotting tests with 
``` shell
xvfb-run -s '-maxclients 2048'
```
_and_ ignoring the charts tests with
``` shell
--ignore=tests/plotting/test_charts.py
```
will cause the Linux CI tests to pass with VTK 9.4. This is just kicking the can down the road but at least helps us isolate the issue.

Full command as part of #7214 
https://github.com/pyvista/pyvista/blob/54b1d8e35e8ce2fc8a5dcbbcfd41d0000ad8947e/.github/workflows/testing-and-deployment.yml#L233-L244

pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
   'pooch',
   'scooby>=0.5.1',
   'typing-extensions',
-  'vtk<9.4.0',
+  'vtk<9.5.0',
```
MatthewFlamm, line 27: Assuming there are no surprises, I think we should add tests for Python 9.13 and add to classifiers in pyproject.toml here.  vtk 9.4.x is the thing holding back support for this version.

banesullivan:
Very close here, just now seeing a number of image regression failures on Windows due to issues with scalar bars. I suspect this is really from the upgrade to `pyvista/setup-headless-display-action@v3` than the VTK 9.4 upgrade itself.

opacity_by_array_preference
bool_scalars

```py
FAILED tests/plotting/test_plotting.py::test_vector_array_with_cells - RuntimeError: test_vector_array_with_cells Exceeded image regression error of 500.0 with an image error equal to: 5783.806535947132
FAILED tests/plotting/test_plotting.py::test_vector_plotting_doesnt_modify_data - RuntimeError: test_vector_plotting_doesnt_modify_data_1 Exceeded image regression error of 500.0 with an image error equal to: 2388.3816993463965
FAILED tests/plotting/test_plotting.py::test_opacity_by_array - RuntimeError: test_opacity_by_array Exceeded image regression error of 500.0 with an image error equal to: 562.5973856209164
FAILED tests/plotting/test_plotting.py::test_opacity_by_array_uncertainty - RuntimeError: test_opacity_by_array_uncertainty Exceeded image regression error of 500.0 with an image error equal to: 570.1673202614394
FAILED tests/plotting/test_plotting.py::test_opacity_by_array_preference - RuntimeError: test_opacity_by_array_preference Exceeded image regression error of 500.0 with an image error equal to: 586.0313725490211
FAILED tests/plotting/test_plotting.py::test_plot_categories_int - RuntimeError: test_plot_categories_int Exceeded image regression error of 500.0 with an image error equal to: 550.8183006535929
FAILED tests/plotting/test_plotting.py::test_plot_categories_true - RuntimeError: test_plot_categories_true Exceeded image regression error of 500.0 with an image error equal to: 528.9215686274407
FAILED tests/plotting/test_plotting.py::test_plot_composite_poly_component_norm - RuntimeError: test_plot_composite_poly_component_norm Exceeded image regression error of 500.0 with an image error equal to: 732.3986928104576
FAILED tests/plotting/test_plotting.py::test_bool_scalars - RuntimeError: test_bool_scalars Exceeded image regression error of 500.0 with an image error equal to: 704.9281045751536
FAILED tests/plotting/test_plotting.py::test_plot_nan_color - RuntimeError: test_plot_nan_color Exceeded image regression error of 500.0 with an image error equal to: 556.2601307189526
```

banesullivan:
`test_camera3d_widget` is failing on the Mac OS CI with a concerning result. The resulting image all looks good however it seems the camera position is in a different location that the baseline image

I'm marking this as a skip for now and noting that this needs investigation later

```
FAILED tests/plotting/test_widgets.py::test_camera3d_widget - RuntimeError: test_camera3d_widget Exceeded image regression error of 500.0 with an image error equal to: 1497.8366013072205
```

| truth | Mac OS CI image|
|---|---|
| camera3d_widget | camera3d_widget |

user27182:
> `test_camera3d_widget` is failing on the Mac OS CI with a concerning result. The resulting image all looks good however it seems the camera position is in a different location that the baseline image
> 
> I'm marking this as a skip for now and noting that this needs investigation later
> 
> ```
> FAILED tests/plotting/test_widgets.py::test_camera3d_widget - RuntimeError: test_camera3d_widget Exceeded image regression error of 500.0 with an image error equal to: 1497.8366013072205
> ```

Not sure if this is related but curiously this image was recently updated as part of https://github.com/pyvista/pyvista/pull/7184/files

banesullivan:
Aaaand segfaults under the docstrings test

banesullivan:
Seeing several failures in the doctests too where it seems GetPointCells is returning in reverse order now:

Pre-9.4:

```py
>>> import pyvista as pv
>>> mesh = pv.Sphere(theta_resolution=10)
>>> mesh.point_cell_ids(0)
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

On 9.4

```py
>>> import pyvista as pv
>>> mesh = pv.Sphere(theta_resolution=10)
>>> mesh.point_cell_ids(0)
[9, 8, 7, 6, 5, 4, 3, 2, 1, 0]
```

user27182:
> Very close here, just now seeing a number of image regression failures on Windows due to issues with scalar bars. I suspect this is really from the upgrade to `pyvista/setup-headless-display-action@v3` than the VTK 9.4 upgrade itself.
> 
> opacity_by_array_preference 



This may not be strictly related to this PR. I am getting a similar error as part of #7266 (with VTK < 9.4). Here's an example failure on windows. This failure is specific to `MultiBlock` meshes (these are images of a  crinkle-clipped MultiBlock).

| expected | windows |
|---|---|
| clip_multiblock_crinkle False-True | clip_multiblock_crinkle False-True |

banesullivan:
At least with the onese I've checked so far, the undelying mapping is correct its just the scalar bar that has these errors. It looks like in your example the underlying mapper has something going wrong too 😕

cquammen:
@banesullivan @user27182 I'm eager to see this pushed through to the end. How can I contribute as an experienced VTK developer but PyVista newbie? Would working to debug the Windows regressions help?

user27182:
My understanding is that the bottleneck is with our CI seg faults, in particular with our docstring checks. As far as I can tell we have not been able to find a workaround to what we believe is likely to be the root cause of the failures (see https://github.com/pyvista/pyvista/pull/6879#issuecomment-2663545111).

I don't know much about the VTK release cycle but if there is a VTK 9.4.2 release planned with a fix for https://gitlab.kitware.com/vtk/vtk/-/issues/19618 this _could_ help push this PR forward if it resolves our seg faults.

@banesullivan has a much better understanding of our plotting CI and docstring checks though so I'll defer to them in case there are other solutions in the works or other critical issues to address.

larsoner:
Just a drive-by comment/idea -- would it be possible to use `pytest-xdist` to spawn multiple jobs? Those would run in separate processes IIRC and thus might make things work.

user27182:
> The fix for maximum X11 clients error went in https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11985

@jspanchu thanks for this! Looks like this fixed our CI seg faults, this is no longer an issue with vtk 9.4.2

user27182:
@banesullivan I took the liberty to update this PR:

- I reverted the workaround for the plotting tests (https://github.com/pyvista/pyvista/pull/7214) since this is not needed with vtk 9.4.2.
-  I renamed `polyhedral_faces` and `polyhedral_face_locations` to use the term `polyhedron` instead. This explicitly overrides the equivalent vtk method (otherwise, both snake-case variants are suggested by IDEs).

The main PR comment has also been updated accordingly to reflect these changes.

Also, `DataSet.point_cell_ids` was updated to reverse the id ordering since the ids returned by `vtkDataSet.GetPointCells` are reversed in vtk 9.4

banesullivan:
Fantastic! Thanks for stepping up @user27182! It's been a busy time for me lately and I let this fall wayside. 

Am I right in thinking that with VTK 9.4.2, this PR is ready to ship? Or what else is needed? I'm trying to catch myself up again right now

user27182:
> Very close here, just now seeing a number of image regression failures on Windows due to issues with scalar bars. I suspect this is really from the upgrade to `pyvista/setup-headless-display-action@v3` than the VTK 9.4 upgrade itself.
> 
> opacity_by_array_preference bool_scalars
> 
> ```python
> FAILED tests/plotting/test_plotting.py::test_vector_array_with_cells - RuntimeError: test_vector_array_with_cells Exceeded image regression error of 500.0 with an image error equal to: 5783.806535947132
> FAILED tests/plotting/test_plotting.py::test_vector_plotting_doesnt_modify_data - RuntimeError: test_vector_plotting_doesnt_modify_data_1 Exceeded image regression error of 500.0 with an image error equal to: 2388.3816993463965
> FAILED tests/plotting/test_plotting.py::test_opacity_by_array - RuntimeError: test_opacity_by_array Exceeded image regression error of 500.0 with an image error equal to: 562.5973856209164
> FAILED tests/plotting/test_plotting.py::test_opacity_by_array_uncertainty - RuntimeError: test_opacity_by_array_uncertainty Exceeded image regression error of 500.0 with an image error equal to: 570.1673202614394
> FAILED tests/plotting/test_plotting.py::test_opacity_by_array_preference - RuntimeError: test_opacity_by_array_preference Exceeded image regression error of 500.0 with an image error equal to: 586.0313725490211
> FAILED tests/plotting/test_plotting.py::test_plot_categories_int - RuntimeError: test_plot_categories_int Exceeded image regression error of 500.0 with an image error equal to: 550.8183006535929
> FAILED tests/plotting/test_plotting.py::test_plot_categories_true - RuntimeError: test_plot_categories_true Exceeded image regression error of 500.0 with an image error equal to: 528.9215686274407
> FAILED tests/plotting/test_plotting.py::test_plot_composite_poly_component_norm - RuntimeError: test_plot_composite_poly_component_norm Exceeded image regression error of 500.0 with an image error equal to: 732.3986928104576
> FAILED tests/plotting/test_plotting.py::test_bool_scalars - RuntimeError: test_bool_scalars Exceeded image regression error of 500.0 with an image error equal to: 704.9281045751536
> FAILED tests/plotting/test_plotting.py::test_plot_nan_color - RuntimeError: test_plot_nan_color Exceeded image regression error of 500.0 with an image error equal to: 556.2601307189526
> ```

In addition to the scalar bar issue,  we seem to have a more substantial difference for the first two tests above. The coloring for the first two rows is different, and it looks like the differently-colored cells are also triangulated
| Expected | Actual |
|----------|--------|
| vector_array_with_cells | vector_array_with_cells |

banesullivan:
ergh, the scalar bars... I've been struggling with them so badly that on my own projects I've just implemented https://github.com/pyvista/pyvista/issues/7192

user27182:
> Am I right in thinking that with VTK 9.4.2, this PR is ready to ship?

Yes, pretty much. Just need to sort out the last windows plotting errors from https://github.com/pyvista/pyvista/pull/6879#issuecomment-2764240660

> ergh, the scalar bars... I've been struggling with them so badly that on my own projects I've just implemented #7192

I'm +1 for #7192. Do you know if this issue breaks _all_ scalar bars in Windows or just for some cases? If it's just for some cases we can probably just highlight it as a known bug and put out a new release anyway? Or should we try to work through #7192 before releasing?

Some other considerations for this PR:

- Should we consider rebasing #7130 and include those changes with this PR?
- This PR creates test coverage gaps for VTK==9.3.1, we should update the test matrix
- Should we restrict vtk version to _exclude_ `9.4.0` and `9.4.1` ? These _probably_ work, the main reason for not allowing these versions is mainly due to our CI seg faults, so for users of these VTK versions it's probably not an issue.

banesullivan:
I just ran the tests locally on this branch against VTK 9.4.2 (on linux)

- ✅ All core tests pass: `pytest tests/core`
- ✅ All other non-plotting tests pass: `pytest --ignore=tests/plotting --ignore=tests/core`
- ✅ All plotting tests pass: `pytest tests/plotting`

I think those failures are only happening on windows and are a result of the GL librarty version installed by `pyvista/setup-headless-display-action`

If so, I'm tempted to just skip the current failures (assuming they only happen on Windows) to get this out.

Let's see how the CI on the above commit turns out ⬆️

banesullivan:
> I'm +1 for https://github.com/pyvista/pyvista/issues/7192. Do you know if this issue breaks all scalar bars in Windows or just for some cases? 

My impression is that it is just some scalar bars... something to do with opacity maybe? I can break out my windows machine if need be

> If it's just for some cases we can probably just highlight it as a known bug and put out a new release anyway? Or should we try to work through https://github.com/pyvista/pyvista/issues/7192 before releasing?

We definitely should not block a release for 9.4 support based on this. It's unfortunate that things are breaking on Windows but considering none of us are funded to support this, there is only so much we can do right now.

banesullivan:
>Should we consider rebasing https://github.com/pyvista/pyvista/pull/7130 and include those changes with this PR?

Let's plan on just merging that shortly after merging this PR if that sounds good? 

> This PR creates test coverage gaps for VTK==9.3.1, we should update the test matrix

Good catch! Yes, let's update that here before merging

> Should we restrict vtk version to exclude 9.4.0 and 9.4.1 ? These probably work, the main reason for not allowing these versions is mainly due to our CI seg faults, so for users of these VTK versions it's probably not an issue.

I think we should be explicit that those versions are not supported - we don't want downstream projects to run into these issues in their CI either

banesullivan:
Proposing #7356

Just rant the tests locally on a Windows machine that I have, and indeed all is good with the scalar bars and opacity for those testsso it must be due to the version of Mesa installed by `pyvista/setup-headless-display-action`

Screenshot 2025-03-29 at 2 55 50 PM


That said... I had a handful of other failures in the plotting module locally. These seem unrelated though.

banesullivan:
Off actually, `test_vector_array_with_cells` doesn't work at all for me locally and renders a completely blank scene

user27182 (changes requested):

user27182:
CI is all green except for the integration test issue introduced by #7356

banesullivan:
> CI is all green except for the integration test issue introduced by https://github.com/pyvista/pyvista/pull/7356

i don't think that #7356 introduced that issue in the integration test... rather I bet something with pyvistaqt needs to be upgraded to support VTK 9.4 that is flakily segfaulting. CI is green now. Let's keep an eye on that

pyvista/plotting/mapper.py
```diff
@@ -61,6 +61,18 @@ def bounds(self) -> BoundsTuple:  # numpydoc ignore=RT01
...
+        """Return the center of mapper.
+
+        Returns
+        -------
+        tuple[float, float, float]
+            Center of the active renderer.
+
+        """
+        return self.GetCenter()  # type: ignore[attr-defined]
+
```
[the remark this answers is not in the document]
banesullivan: I think raising an error or returning `[0, 0, 0]` would be the way to go

pyvista/plotting/renderer.py
```diff
@@ -1072,6 +1072,19 @@ def add_axes_at_origin(
         self.Modified()
         return self._marker_actor
 
+    def delete_axes_widget(self):
+        """Remove and delete the current axes widget."""
+        if hasattr(self, 'axes_widget'):
+            self.axes_actor = None
+            # HACK: set the viewport to a tiny value to hide the widget first
+            # This is due to an issue with a blue box appearing after removal
+            # Tracked in https://gitlab.kitware.com/vtk/vtk/-/issues/19592
```
user27182, line 1081: Looks like this was fixed and backported to 9.4.2. Can probably remove any hacks for the axes widget

pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
...
```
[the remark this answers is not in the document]
ConnorNoddin: Does having multiple version requirements "AND" them or "OR" them? My research seems to indicate it is an "AND" statement so this doesn't seem logical.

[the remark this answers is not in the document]
user27182: > Does having multiple version requirements "AND" them or "OR" them? My research seems to indicate it is an "AND" statement so this doesn't seem logical.

Thanks, I modified my suggestion:

``` diff
- 'vtk<9.4.0',
- 'vtk>=9.4.2',
+ 'vtk!=9.4.0',
+ 'vtk!=9.4.1',
  'vtk<9.5.0',
```

keurfonluu:
Sorry, I am not sure what I just did, tried to rebase another branch from this one. Is it possible to undo what I've done? Sincere apologies.

user27182:
> Sorry, I am not sure what I just did, tried to rebase another branch from this one. Is it possible to undo what I've done? Sincere apologies.

@keurfonluu  All good, I am not sure I understand what happened here either but CI should be green again.

user27182 (approved):
LGTM. Some follow-up items:
- I think we should remove the axes widget hack (issue was fixed in vtk 9.4.2)
- Need to update the test matrix to include testing for VTK 9.3.1 (I think this can be done along with python 3.13 support in a separate PR)
- Need to fix polyhedron support with meshio (#7130)
- PyVistQt integration tests are flaky and should be reviewed and/or updated

user27182:
Looks like there's still work to do. 

### Issue 1
We're getting empty meshes for some examples:

https://github.com/pyvista/pyvista/actions/runs/14227645421/job/39870895969#step:10:302
``` python
WARNING: 
../../examples/00-load/create_pointset.py unexpectedly failed to execute correctly:

      File "/home/runner/work/pyvista/pyvista/examples/00-load/create_pointset.py", line 56, in <module>
        clipped_pset.plot(show_scalar_bar=False)
      File "/home/runner/work/pyvista/pyvista/pyvista/core/pointset.py", line 393, in plot
        return pdata.plot(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/work/pyvista/pyvista/pyvista/_plot.py", line 280, in plot
        pl.add_mesh(var_item, **kwargs)
      File "/home/runner/work/pyvista/pyvista/pyvista/plotting/plotter.py", line 3695, in add_mesh
        raise ValueError(msg)
    ValueError: Empty meshes cannot be plotted. Input mesh has zero points. To allow plotting empty meshes, set `pv.global_theme.allow_empty_mesh = True`
```



https://github.com/pyvista/pyvista/actions/runs/14227645421/job/39870895969#step:10:394


``` python
WARNING: 
../../examples/01-filter/subdivide.py unexpectedly failed to execute correctly:

      File "/home/runner/work/pyvista/pyvista/examples/01-filter/subdivide.py", line 65, in <module>
        plotter = plot_subdivisions(mesh, 1, 3)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/work/pyvista/pyvista/examples/01-filter/subdivide.py", line 54, in plot_subdivisions
        row_plot(1, 'butterfly')
      File "/home/runner/work/pyvista/pyvista/examples/01-filter/subdivide.py", line 50, in row_plot
        p.add_mesh(mesh.subdivide(subs[i], subfilter=subfilter), **display_args)
      File "/home/runner/work/pyvista/pyvista/pyvista/plotting/plotter.py", line 3695, in add_mesh
        raise ValueError(msg)
    ValueError: Empty meshes cannot be plotted. Input mesh has zero points. To allow plotting empty meshes, set `pv.global_theme.allow_empty_mesh = True`
```

### Issue 2
Plus, we're getting some X Server warnings, not yet sure if this is issue:

https://github.com/pyvista/pyvista/actions/runs/14227645421/job/39870895969#step:10:1375
``` python
2025-04-02 20:45:29.310 (   1.070s) [    7F1AAB315B80]vtkXOpenGLRenderWindow.:1416  WARN| bad X server connection. DISPLAY=
```

### Issue 3
And on top of that the build was unexpectedly cancelled. Looks like it may have crashed...?  Seems to have happened with the charts examples, maybe these are still broken and the fix from #7189 didn't actually work?
https://github.com/pyvista/pyvista/actions/runs/14227645421/job/39870895969#step:10:2633
``` python
Error: The operation was canceled.
```

user27182 (changes requested):
The fact that the build is crashing is a sign this PR should not yet be merged. If anyone here has time to contribute to help resolve any of the issues outlined in https://github.com/pyvista/pyvista/pull/6879#issuecomment-2773862865 that would go a long way.

I'm thinking this would entail [building the documentation](https://dev.pyvista.org/extras/developer_notes#building-the-documentation) locally and trying to get some sort of traceback or more complete information about what the cause of the crash is and work from there.

user27182:
Re: https://github.com/pyvista/pyvista/pull/6879#issuecomment-2773862865
Fixes are on the way via #7371 and #7373.

user27182:
> image

I have identified the source of this change. With vtk < 9.4, `vtkPolyDataNormals` seems to always compute normals, regardless if normals already exist. But with vtk 9.4, normals are only computed if there aren't already pre-existing normals, and so any existing normals are preserved.

And so the new image is the (incorrect) default point normals that come pre-computed by `vtkCubeSource` instead of the  re-computed (correct) point normals.

Show default cube normals (these are wrong, IMO):
``` python
import pyvista as pv; 
normals1 = pv.Cube().point_data['Normals']
print(normals1)
# pyvista_ndarray([[-1.,  0.,  0.],
#                  [-1.,  0.,  0.],
#                  [-1.,  0.,  0.],
#                  [-1.,  0.,  0.],
#                  [ 1.,  0.,  0.],
#                  [ 1.,  0.,  0.],
#                  [ 1.,  0.,  0.],
#                  [ 1.,  0.,  0.]], dtype=float32)
```

Re-compute the normals.

``` python
normals2 = pv.Cube().compute_normals().point_data['Normals']
```
With vtk < 9.4, this is the output:
``` python
print(normals2)
# pyvista_ndarray([[-0.57735026, -0.57735026, -0.57735026],
#                  [-0.57735026, -0.57735026,  0.57735026],
#                  [-0.57735026,  0.57735026,  0.57735026],
#                  [-0.57735026,  0.57735026, -0.57735026],
#                  [ 0.57735026, -0.57735026, -0.57735026],
#                  [ 0.57735026,  0.57735026, -0.57735026],
#                  [ 0.57735026,  0.57735026,  0.57735026],
#                  [ 0.57735026, -0.57735026,  0.57735026]], dtype=float32)
```

But with vtk 9.4, the original normals are returned.

Maybe be we should add `recompute=True` as a new kwarg for `compute_normals` so that normals are re-computed by default?
Plus, unrelated to vtk 9.4, we should also fix the default point `Normals` array returned by `vtkCubeSource`

user27182 (dismissed):
LGTM.
The header comment has also been updated to include the recent patch fixes and follow-up items.

tkoyama010:
Thank you all!

adeak:
Kudos, all!

banesullivan:
Thank you to everyone who helped make this happen!!

Major shoutout to @user27182 for taking the lead and getting this over the finish line!

pyvista/plotting/renderer.py
```diff
@@ -1537,9 +1541,7 @@ def hide_axes(self) -> None:
         >>> pl.hide_axes()
 
         """
-        if hasattr(self, 'axes_widget') and self.axes_widget.GetEnabled():
-            self.axes_widget.EnabledOff()
-            self.Modified()
+        self.delete_axes_widget()
```
larsoner, line 1544: This substitution is incorrect I think... don't want to delete the axes in this case just hide them so that `show_axes` can be called later

PR #6881, opened 2024-11-27
Raise warning when some points could not be sampled in `sample` method

tkoyama010:
### Overview

Raise warning when some points could not be sampled in `sample` method.

### Details

- None


banesullivan:
This could be a valuable opt-in addition but I think its too aggressive to always warn as the majority of datasets I work with always have missed/masked points. Closing for now, but i do think having an option flag to warn in this scenario would be helpful

PR #6884, opened 2024-11-27
Fully annotate `fileio.py`

user27182:
### Overview

I ran ` ruff check --select ANN pyvista/core/utilities/fileio.py --ignore ANN003` and added all annotations such that this check passes.


tkoyama010 (approved):
LGTM

PR #6889, opened 2024-11-27
Fully annotate `points.py`

user27182:
### Overview

I ran  `ruff check --select ANN pyvista/core/utilities/points.py` and resolved all errors until the check passed.


tkoyama010 (approved):
LGTM

PR #6890, opened 2024-11-27
Add `ANN` rule to `ruff` for checking annotations

user27182:
### Overview

In #6279, it is suggested to use [`disallow_untyped_defs = True`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-untyped-defs) to check for the use of annotations. As an alternative, I propose using `ruff` with the [`ANN` rules](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann). Using `ruff` instead of `mypy` is superior for a few reasons:

- `ruff` is very fast and runs in pre-commit, whereas mypy is slower and needs to be run separately from pre-commit.
- `ruff` is more configurable, e.g. it allows for only enforcing annotations in public or private functions, for example
- `ruff`'s per-file ignores allows us to see a central "to-do" list of all files which still require annotations. This will enable us to create an open issue and point contributors to these ignores as a starting point for files which need attention.
- If using per-line ignores, then using `# noqa: ANN001` instead of `# type: ignore[no-untyped-def]` is a more helpful ignore because the rule is more specific, and it keeps these kind of errors separate from "true" type errors

By default, all `ANN` rules are enabled in this PR, except that annotations for `*args` and `**kwargs` are not enforced and the use of `Any` is allowed.

As new annotation PRs are created (e.g. #6889, #6856, etc.), the newly annotated files can be removed from the ignores list.


akaszynski (approved):
Fully agree with moving as much as possible from `mypy` to `ruff`. Who knows, maybe one day they'll be able to typecheck as well!

akaszynski:
> Fully agree with moving as much as possible from `mypy` to `ruff`. Who knows, maybe one day they'll be able to typecheck as well!

Along that line, there's this: https://github.com/mtshiba/pylyzer

PR #6898, opened 2024-11-29
Use Trimesh.contains for select_enclosed_points

banesullivan:
Improves the implementation of `select_enclosed_points` so that we can leverage Trimesh.contains which I've noticed to be ***much more reliable*** than VTK's `vtkSelectEnclosedPoints`

See https://trimesh.org/trimesh.base.html#trimesh.base.Trimesh.contains

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -3761,6 +3761,7 @@ class DataSetFilters:
         inside_out: bool = False,
         check_surface: bool = True,
         progress_bar: bool = False,
+        use_trimesh: bool | None = None,
     ):
         """Mark points as to whether they are inside a closed surface.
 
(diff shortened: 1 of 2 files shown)


derek-kinakin:
Maybe late to the discussion, so feel free to ignore, but have you looked at IGL's "fast winding number" calculations? https://libigl.github.io/libigl-python-bindings/igl_docs/#fast_winding_number_for_meshes

A collegue first introduced it to me as a way to test for points inside or outside of a mesh. It is now our default approach for this type of testing.

banesullivan:
Thanks for chiming in @derek-kinakin! I hadn't heard of IGL and it looks like it has a ton of useful functionality we could leverage when our filters fail or are missing functionality

banesullivan:
I'm closing due to staleness and because some other work has been done in https://github.com/pyvista/pyvista/pull/8176 to provide better interoperability with trimesh. Perhaps a better solution is just to convert to trimesh then use trimesh directly for this operation

user27182:
> I'm closing due to staleness and because some other work has been done in #8176 to provide better interoperability with trimesh. Perhaps a better solution is just to convert to trimesh then use trimesh directly for this operation

I think the new `select_interior_points` filter proposed in #8181 should enable native use of pyvista filters without the need to use Trimesh.

PR #6895, opened 2024-11-29
Add `nitpicky` option to doc config to prevent broken links

user27182:
### Overview

Continued from #6203

PR #6904, opened 2024-12-01
Remove 'of', 'or', and 'various' from nitpick ignore

user27182:
### Overview

Disallow using these terms for documenting types in docstrings. Part of #6895.


tkoyama010 (approved):
LGTM

PR #6912, opened 2024-12-02
Fully annotate `check.py`

user27182:
### Overview

Add annotations such that this passes:
`ruff check --select ANN --ignore ANN401 pyvista/core/_validation/check.py`


tkoyama010 (approved):
LGTM

PR #6913, opened 2024-12-02
Add annotations to `validate.py`

user27182:
### Overview

Add annotations except for `*args`, `**kwargs`, and return types. Return types require overloads which will be addressed separately.

This passes 
`ruff check --select ANN --ignore ANN401 --ignore ANN003 --ignore ANN202 pyvista/core/_validation/validate.py `


tkoyama010 (dismissed):
LGTM

tkoyama010:
@user27182 Could you please resolve the conflict?

tkoyama010 (approved):

pyvista/core/_validation/validate.py
```diff
@@ -708,7 +721,10 @@ def _array_from_vtkmatrix(
     return array
 
 
-def validate_number(num, /, *, reshape: bool = True, **kwargs):
+_T = TypeVar('_T', float, int)
+
+
+def validate_number(num: _T, /, *, reshape: bool = True, **kwargs) -> _T:
```
tkoyama010, line 727: Sphinx raises warnng that we cannot reference `_T` in document.
```suggestion
def validate_number(num: float | int, /, *, reshape: bool = True, **kwargs) -> float | int:
```

tkoyama010 (changes requested):

pyvista/core/_validation/validate.py
user27182: Hmm. This is because of #6895.  TypeVar is better because it handles the `int -> int` mapping but we can use overloads instead (and add these in a separate PR)

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

tkoyama010:
I got the knowledge from error of ruff that the representation of int | float is redundant. (see [PYI041](https://docs.astral.sh/ruff/rules/redundant-numeric-union/#redundant-numeric-union-pyi041))

tkoyama010:
Oops, this make it inconsistent with mypy and give an error... I guess I'll have to skip the ruff error with .noqa.

tkoyama010 (approved):

user27182:
Yes the typing for `float` is special since it includes `int`. Let's add return types separately

tkoyama010 (approved):

PR #6915, opened 2024-12-02
Fix ruff lint error: Bump version to 0.8.1

tkoyama010:
### Overview

Fix ruff lint error: Bump version to 0.8.1

### Details

- ref #6914


pyvista/plotting/renderer.py
```diff
@@ -874,7 +874,7 @@ def add_actor(
         self.AddActor(actor)  # must add actor before resetting camera
         self._actors[name] = actor
 
-        if reset_camera or not self.camera_set and reset_camera is None and not rv:
+        if reset_camera or (not self.camera_set and reset_camera is None and not rv):
```
banesullivan, line 877: This needs careful review, I don't think the assumption here by ruff is correct

pyvista/plotting/renderer.py
```diff
@@ -2669,7 +2669,7 @@ def remove_actor(self, actor, reset_camera=False, render=True):
         if name is not None:
             self._actors.pop(name, None)
         self.update_bounds_axes()
-        if reset_camera or not self.camera_set and reset_camera is None:
+        if reset_camera or (not self.camera_set and reset_camera is None):
```
banesullivan, line 2672: This needs careful review, I don't think the assumption here by ruff is correct

tests/examples/test_dataset_loader.py
```diff
@@ -69,10 +69,8 @@ def add_to_dict(func: str, dataset_function: Callable[[], Any]):
 
     # Collect all `download_<name> or `load_<name>` functions
     def _is_dataset_function(name, item):
-        return (
-            isinstance(item, FunctionType)
-            and name.startswith('download_')
-            or name.startswith('load_')
+        return (isinstance(item, FunctionType) and name.startswith('download_')) or name.startswith(
+            'load_'
         )
```
banesullivan, line 74: This one looks to be blatantly incorrect. The check is first evaluating `isinstance(item, FunctionType)` and if so see if the name starts with download or load.

This makes me wearing of all changes to conditional checks in this PR

banesullivan (changes requested):
Ruff is incorrectly changing these conditional statements and they need to be handled manually with care

tkoyama010:
The points that were pointed out are thought to be automatic changes by ruff. Please give me some time as I will carefully check them (Note: Documents that need to be verified [Operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence)). I'll also check to see if this automatic change has been reported as a bug at ruff project.

user27182:
If we don't like the fixes for a specific rule we can do:
``` toml
[tool.ruff.lint]
unfixable = ["Rule1", "Rule2", "etc."]
```

That way we can still upgrade ruff but just disable auto-fixes for the problem rule(s). See https://docs.astral.sh/ruff/configuration/

MatthewFlamm:
We do need to do this carefully, but even if ruff is wrong here, the lines highlighted are very confusing as written.  The addition of parentheses to whatever the correct order is should be done in either case.

tests/examples/test_dataset_loader.py
MatthewFlamm: In my reading of https://docs.python.org/3/reference/expressions.html#operator-precedence ruff is making the order of python precedence explicit.  That is, it evaluates the `and` first then the `or` second.  So the ruff version is giving us what is currently being done, but explicitly showing it.  Which means that the code was incorrect and ruff found a bug.

pyvista/core/composite.py
```diff
@@ -907,7 +907,7 @@ def clean(self: MultiBlock, empty: bool = True) -> None:
                 self[i].clean()  # type: ignore[union-attr]
                 if self[i].n_blocks < 1:  # type: ignore[union-attr]
                     null_blocks.append(i)
-            elif self[i] is None or empty and self[i].n_points < 1:  # type: ignore[union-attr]
+            elif self[i] is None or (empty and self[i].n_points < 1):  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
MatthewFlamm: This one seems correct to me in that ruff is making the python precedence order explicit.  If the logic is incorrect, then it was a bug.

MatthewFlamm:
My reading is that all the highlighted examples are _not_ ruff bugs, but are either bugs in pyvista or misunderstanding of the precedence order in python.  If these are indeed bugs in pyvista, it shows the power of keeping this rule in ruff turned on.

pyvista/core/composite.py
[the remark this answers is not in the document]
MatthewFlamm: This logic looks correct to me as ruff has it.

pyvista/plotting/renderer.py
MatthewFlamm: While ruff has not changed the python logic, it highlights the underlying issue.  The correct logic here is dense and not immediately obvious to me.  Could be a bug here....

MatthewFlamm: `reset_camera` is declared to be of type `bool`, so I'm not sure when it can be `None`.

tests/examples/test_dataset_loader.py
MatthewFlamm: This one is a bug.

MatthewFlamm, line 74:
```suggestion
        return isinstance(item, FunctionType) and (
            name.startswith('download_') or name.startswith('load_')
        )
```

Not sure if this is correctly formatted with tabs.

doc/source/make_tables.py
```diff
@@ -1488,16 +1488,9 @@ def fetch_multiblock(cls, kind: Literal['hetero', 'homo', 'single']):
...
-                        num_datasets >= 2
-                        and len(set(types_list)) == 1
-                        and kind == 'homo'
-                        or len(set(types_list)) > 1
-                        and kind == 'hetero'
-                    )
+                if (num_datasets == 1 and kind == 'single') or (
+                    (num_datasets >= 2 and len(set(types_list)) == 1 and kind == 'homo')
+                    or (len(set(types_list)) > 1 and kind == 'hetero')
                 ):
```
user27182, line 1494: Make the logic easier to follow. I ran this locally and checked that the generated tables add up correctly in the dataset gallery (i.e. n_homo + n_hetero + n_single == n_multiblock)
```suggestion
                num_types = len(set(types_list))

                is_single = num_datasets == 1
                is_homo = num_datasets >= 2 and num_types == 1
                is_hetero = num_datasets >= 2 and num_types > 1
                if (is_single and kind == 'single') or (is_homo and kind == 'homo') or (is_hetero and kind == 'hetero'):
```

pyvista/core/composite.py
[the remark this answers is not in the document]
user27182: This also looks good to me as formatted by ruff

pyvista/plotting/renderer.py
user27182: We now have redundant logic and not reachable checks enabled with mypy, merge with main might highlight this?

tests/examples/test_dataset_loader.py
```diff
@@ -69,10 +69,8 @@ def add_to_dict(func: str, dataset_function: Callable[[], Any]):
 
     # Collect all `download_<name> or `load_<name>` functions
     def _is_dataset_function(name, item):
-        return (
-            isinstance(item, FunctionType)
-            and name.startswith('download_')
-            or name.startswith('load_')
+        return (isinstance(item, FunctionType) and name.startswith('download_')) or name.startswith(
+            'load_'
```
user27182, line 73: I ran this locally and double checked that test_examples and test_downloads are working as expected w.r.t. this function
```suggestion
        return isinstance(item, FunctionType) and name.startswith(('download_', 'load_'))
```

tests/examples/test_dataset_loader.py
user27182: name.startswith can accept a tuple input and ruff will complain about thist, see my suggestion above

user27182 (commented):
Still need to work through the `reset_camera` example but these other fixes look good to me

tests/examples/test_dataset_loader.py
user27182: This was fixed in https://github.com/pyvista/pyvista/pull/6915/commits/8e574e338d91b91ffc4d237baabfba5c4337b098

user27182: This was fixed in https://github.com/pyvista/pyvista/pull/6915/commits/8e574e338d91b91ffc4d237baabfba5c4337b098

pyvista/plotting/renderer.py
user27182: If we change `reset_camera=False` to `reset_camera: bool=False` in the function signature, mypy indeed warns that `reset_camera is None` is unreachable. We can remove this logic altogether is it must be `False` here.

And `rv` is initialized earlier as `None` and will only be set if an actor of the same name previously existed and was removed. So this logic will reset the camera if it hasn't already been set and if no actors were removed.

I think this is pretty much what the logic was already doing before this PR:
```suggestion
        if reset_camera or (not self.camera_set and rv is None):
```

But, is this correct...? We only want to reset the camera if explicitly requested OR if it hasn't been set and no actor was removed? Shouldn't we reset the camera if the camera _is_ set and an actor _is_ removed?

I'm thinking that this may be the correct logic here, since we probably need to reset the camera if an actor was removed:
```suggestion
        if reset_camera or (self.camera_set and rv is not None):
```

pyvista/plotting/renderer.py
```diff
@@ -2667,7 +2667,7 @@ def remove_actor(self, actor, reset_camera=False, render=True):
...
```
user27182, line 2670: Re: this thread https://github.com/pyvista/pyvista/pull/6915/files#r1866776558 and this thread https://github.com/pyvista/pyvista/pull/6915/files#r1866776493, I think this may be the correct logic here:

```suggestion
        if reset_camera or not self.camera_set:
```
When removing an actor should we reset the camera if it hasn't been manually set by the user...? I am not sure how the `camera_set` flag works, is it set automatically internally whenever it's set by the renderer or is this strictly controlled by the user (to avoid resetting manual camera changes)

user27182 (approved):
LGTM. There are still unresolved questions regarding if the logic with reset_camera is correct, but this logic is unchanged by this PR - ruff merely made it more explicit. 

If any changes to the logic _do_ need to be made then I think this can be made in a follow-up PR and should not necessarily hold this one up since it is otherwise preventing us from updating ruff.

banesullivan (approved):
I agree that a lot of the logic highlighted in this diff needs to be revisted/audited as I believe some are making incorrect assumptions to begin with and are likely bugs in our code. 

Otherwise, from a quick glance over and trusting the thorough discussions above, approving

PR #6917, opened 2024-12-03
Add `slice_along_plane` method for faster slicing

tkoyama010:
### Overview

ref https://vtk.org/doc/nightly/html/classvtkPlaneCutter.html

### Details

- None


banesullivan (changes requested):
This is a great find! Though we should implement this as a part of the `slice_implicit` or `slice` method directly to choose which kind of cutter to use. `vtkPlaneCutter` and also `vtk3DLinearGridPlaneCutter` should be supported https://vtk.org/doc/nightly/html/classvtk3DLinearGridPlaneCutter.html

PR #6918, opened 2024-12-03
Expand concrete dataset type definitions

user27182:
### Overview

Follow-up from #6854. This PR expands the types so that all abstract dataset classes have a new "concrete" typevar.
These typevars are needed in #6828 to resolve type errors, and are used similarly to #6870.

Instead of using invariant typevars (as originally defined in #6854), this PR uses `bound` typevars. Using `bound` allows us to use Unions (type aliases) instead which can be re-used in multiple definitions. This avoids the need to duplicate definitions and should make it easier to modify and maintain.


tkoyama010 (approved):
LGTM

PR #6919, opened 2024-12-03
Add return info to `set_default_active_scalars` and  `set_default_active_vectors`

user27182:
### Overview

These functions currently return `None`. But, a common pattern is to first set an active array, and then get its info on the next line.

From a typing perspective, this is problematic because the array name has type `str | None`, but the `set_default` functions guarantee that an array will be set as active (else raise an error), so the type should be narrowed to `str` only. With the current implementation, we now need to do `cast(str, active_scalars_name)` every time.

Instead, this PR proposes to return the array info directly and have these functions do the casting internally. The data set filters are refactored to make use of this new feature.

This change will help with the typing in #6828.


user27182:
Should add type hints to the `scalars` parameters to "turn on" type checking for this to make sure the type-narrowing is working.

pyvista/core/dataset.py
```diff
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
-from collections.abc import Iterator
```
user27182, line 6: Side note: this is an unused import that was introduced in #6835 and not linted (removed) by `ruff` automatically

tkoyama010 (approved):
LGTM

PR #6935, opened 2024-12-05
Add parameter annotations to `extract` dataset filters

user27182:
### Overview

Part of #6828. Also includes changes from #6834.


pyvista/core/filters/data_set.py
```diff
@@ -1801,7 +1801,9 @@ def outline_corners(  # type: ignore[misc]
         return wrap(alg.GetOutputDataObject(0))
 
     def extract_geometry(  # type: ignore[misc]
-        self: ConcreteDataSetType, extent: Sequence[float] | None = None, progress_bar: bool = False
+        self: ConcreteDataSetType,
+        extent: Sequence[float] | None = None,
```
tkoyama010, line 1805: Is the policy is to use VectorLike instead of Sequence?
```suggestion
        extent: VectorLike[float] | None = None,
```

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
user27182: VectorLike is more general and is consistent with other changes, so I think this is good practice. Will likely need some additional minor changes to satisfy `mypy` though.

tkoyama010 (approved):
LGTM

PR #6939, opened 2024-12-05
Make `check_contains` API easier to use and understand

user27182:
### Overview

I find this check particularly confusing to use.  There is ambiguity in the order of operands because typically we do `a in b` in code but when using arguments it's less clear, and the [`operator` module](https://docs.python.org/3/library/operator.html#operator.contains) even reverses the order of the operands.

To address this, `check_contains` currently enforces keyword args only so that `item` and `container` are explicit. I still find this a bit awkward though since it leads to a different order of arguments in code. Every time I see this check I find I have to read it carefulyl and think about whether the arguments are set correctly.

Instead, this PR modifies the API to align it more with the `operator` module, such that the container is the first arg. It also can no longer be used as a keyword, so it _must_ always be the first arg. And the item being checked is renamed to `must_contain` and remains keyword only. Now, when reviewing code, it reads more like "x must contain y" which I think is much more understandable at first glance.

EDIT: This is a breaking change. But the module is private and the documentation states that the API is unstable and subject to change without notice.

### DIFF
--- a/pyvista/core/_validation/check.py
+++ b/pyvista/core/_validation/check.py
@@ -1036,16 +1036,16 @@ def check_iterable_items(
     )
 
 
-def check_contains(*, item: Any, container: Container[Any], name: str = 'Input') -> None:
+def check_contains(container: Container[Any], /, must_contain: Any, *, name: str = 'Input') -> None:
     """Check if an item is in a container.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


tkoyama010 (approved):
If this has already been released, please label `breaking-change`.

PR #6940, opened 2024-12-06
Move common dataset/composite filters into a new `DataObjectFilters` class

user27182:
### Overview

`MultiBlock` is valid input for a number of `DataSetFilters`. However, instead of using inheritance to support this, these methods are "inherited" through simple variable assignment, e.g. `CompositeFilters.clip = DataSetFilters.clip`. I find this to be very confusing and is not consistent with the inheritance model used for all other filters. It also makes it extremely difficult to add proper type hints to filters since it's no longer a valid assumption that a `DataSetFilter` should only accept a `DataSet` as input for `self` (pretty much all filters were annotated with `self: ConcreteDataSetType` in https://github.com/pyvista/pyvista/pull/6870, but this is incorrect for a number of filters that also accept `MultiBlock`).

This PR adds a new `DataObjectFilters` class, and `DataSetFilters` and `CompositeFilters` now inherit from it. All filters common to `DataSetFilters` and `CompositeFilters` are moved to this new class and relevant tests are also moved to a new test file. 

Any internal references to `DataSetFilters.method(self)` in the moved filters were updated to use `self.method` instead. No changes are otherwise made to the code.

Note: some tests were moved to `test_dataobject_filters` without also moving the filter: e.g. `extract_geometry` and `outline` have their own separate implementations for `DataSet` and `MultiBlock`, but tests for these inputs are often grouped together anyway so these were moved for consistency. Also, tests for `transform` are _not_ moved even though this filter is also common to both `DataSetFilters` and `CompositeFilters` since all the related filters `rotate_x`, `scale`, etc. are only implemented for `DataSetFilters`.

As a follow-up it might be good to consider unifying some of the common filters like `extract_geometry` into `DataObjectFitlers` so that all code is in a central location. And a follow-up is also needed to fix the `self` annotations to include `MultiBlock`.

See list of filters moved to this new class:
<details> 

```
DataSetFilters._clip_with_function
DataSetFilters.clip 
DataSetFilters.clip_box 
DataSetFilters.slice
DataSetFilters.slice_implicit
DataSetFilters.slice_orthogonal
DataSetFilters.slice_along_axis
DataSetFilters.slice_along_line
DataSetFilters.extract_all_edges
DataSetFilters.elevation
DataSetFilters.compute_cell_sizes
DataSetFilters.cell_centers
DataSetFilters.cell_data_to_point_data
DataSetFilters.ctp
DataSetFilters.point_data_to_cell_data
DataSetFilters.ptc
DataSetFilters.sample
DataSetFilters.triangulate
```


user27182:
There are currently lots of changes being made to the filters with adding annotations, so this PR will wait until those are finalized first. Otherwise, the copy/pasting here to new files will likely be stale/outdated and the new changes may be lost.

PR #6943, opened 2024-12-06
Fully annotate `grid.py`

user27182:
### Overview

Add annotations.

### DIFF
--- a/pyvista/core/grid.py
+++ b/pyvista/core/grid.py
@@ -15,7 +15,6 @@ import numpy as np
 import pyvista
 
 if TYPE_CHECKING:  # pragma: no cover
-    from typing_extensions import Never
     from typing_extensions import Self
 
     from pyvista import StructuredGrid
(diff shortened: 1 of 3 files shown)


pyvista/core/grid.py
```diff
@@ -685,7 +694,7 @@ def points(self) -> NumpyArray[float]:
         return points
 
     @points.setter
-    def points(self, points):  # numpydoc ignore=PR01
+    def points(self: ImageData, points: Never) -> Never:  # numpydoc ignore=PR01
```
tkoyama010, line 697: I did not know about the type `Never` . Please wait as I am learning its usage.

tkoyama010 (changes requested):

pyvista/core/grid.py
```diff
@@ -214,20 +221,20 @@ def __init__(
             else:
                 raise TypeError('Arguments not understood by `RectilinearGrid`.')
 
-    def __repr__(self):
+    def __repr__(self: RectilinearGrid) -> str:
         """Return the default representation."""
         return DataSet.__repr__(self)
 
-    def __str__(self):
+    def __str__(self: RectilinearGrid) -> str:
```
user27182, line 228: Maybe use Self here

pyvista/core/grid.py
user27182: I think it has an alias `NoReturn` which may be more explicit.

We can also remove it and hint the actual types, i.e. make them have the same signature as `PointSet.points` setter

tkoyama010 (approved):
LGTM

PR #6944, opened 2024-12-06
Fully annotate `datasetattributes.py`

user27182:
### Overview

`_patch_type` was also modified since it previously assumed the input had a `VTKObject` property, but this is not always true, and is only the case when `pyvista_ndarray` is created from a `vtkAbstractArray`


tkoyama010 (approved):
LGTM

Issue #6867, opened 2024-11-25
Test collection fails due to non-unique names


beroda:
### Describe the bug, what's wrong, and what you expected.

After a fresh editable install of `pyvista[dev]` following the contributing guide, pytest complains to collect tests due to non-unique test module names.

![image](https://github.com/user-attachments/assets/79d3b9e9-73b4-4800-971e-168a1e1f0e06)

Tried to remove `__pycache__` and `.pyc` files as suggested without any success.
However, following pytest `guidelines` [here](https://docs.pytest.org/en/latest/explanation/goodpractices.html#choosing-an-import-mode), adding an `__init__.py` inside `tests/core` fixes the issue.

Apparently, switching pytest import mode to [`importlib`](https://docs.pytest.org/en/latest/explanation/pythonpath.html#import-modes) could also help. Might be overkill though.

### Steps to reproduce the bug.

```shell
pytest --collect-only
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Mon Nov 25 23:58:11 2024 CET

                OS : Linux (Ubuntu 20.04)
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 15.5 GiB
       Environment : Python
       File system : ext4
        GPU Vendor : Intel
      GPU Renderer : Mesa Intel(R) HD Graphics 630 (KBL GT2)
       GPU Version : 4.6 (Core Profile) Mesa 21.2.6
  MathText Support : True

  Python 3.11.7 (main, Feb 23 2024, 11:15:59) [GCC 8.4.0]

           pyvista : 0.45.dev0
               vtk : 9.3.1
             numpy : 1.26.4
        matplotlib : 3.9.2
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 11.0.0
           imageio : 2.36.0
           IPython : 8.29.0
          colorcet : 3.1.0
           cmocean : 4.0.3
        ipywidgets : 8.1.5
             scipy : 1.14.1
              tqdm : 4.66.6
            meshio : 5.3.5
        jupyterlab : 4.2.5
    pytest_pyvista : 0.1.8
             trame : 3.7.0
      trame_client : 3.5.0
      trame_server : 3.2.3
         trame_vtk : 2.8.11
     trame_vuetify : 2.7.1
      nest_asyncio : 1.6.0
--------------------------------------------------------------------------------
```

beroda:
[...]

PR #6948, opened 2024-12-06
Fix non-unique test names during collection

beroda:
### Overview

Fixes https://github.com/pyvista/pyvista/issues/6867

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -217,6 +217,9 @@ jobs:
           pip install --pre --no-deps -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple matplotlib
           pip install --pre --no-deps -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy
 
+      - name: Test collection (temporary)
+        run: python -m pytest --collect-only
+
       - name: Core Testing (no GL)
         run: python -m pytest --cov=pyvista -v tests/core tests/examples
 


.github/workflows/testing-and-deployment.yml
```diff
@@ -217,6 +217,9 @@ jobs:
           pip install --pre --no-deps -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple matplotlib
           pip install --pre --no-deps -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy
 
+      - name: Test collection (temporary)
+        run: python -m pytest --collect-only
+
```
beroda, line 222: Added temporary only to see if it fails on CI. Will revert later.

beroda: CI has failed as expected https://github.com/pyvista/pyvista/actions/runs/12206271755/job/34055394148

beroda: Although the failure is different from what's been observed locally ...

beroda: Differences on CI was caused by OpenGL. Correctly fails [here](https://github.com/pyvista/pyvista/actions/runs/12213181187/job/34072507064?pr=6948#step:14:8321) .

beroda:
The fix is working correctly, all tests are collected (see [here for example](https://github.com/pyvista/pyvista/actions/runs/12213253599/job/34072661062?pr=6948#step:14:8323)).
Test collection step can be removed from CI.

user27182:
Looks like the macOS tests are hanging up - auto-cancelled after running for 6hrs
https://github.com/pyvista/pyvista/actions/runs/12214255536/job/34074801179?pr=6948

beroda:
For some unknown reason, the MacOS tests are failing with `import-mode=importlib` (see [here](https://github.com/pyvista/pyvista/actions/runs/12214141366/job/34074545934?pr=6943#step:6:431) )
Trying to switch to test modules with `__init__.py` files

user27182 (approved):

banesullivan:
It looks like a force push reverted the changes?

user27182:
Looks good to me as-is. Two new empty init files are added. This was one of the possible fixes. The original fix that was reverted used a pytest arg instead of init files. Both fixes seemed to work locally but CI fails with the pytest option.

PR #6949, opened 2024-12-06
Fully annotate `pyvista_ndarray.py`

user27182:
### Overview

Add annotations.

### DIFF
--- a/pyvista/core/pyvista_ndarray.py
+++ b/pyvista/core/pyvista_ndarray.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from collections.abc import Iterable
 from typing import TYPE_CHECKING
+from typing import cast
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


pyvista/core/pyvista_ndarray.py
```diff
@@ -111,7 +118,7 @@ def __setitem__(self, key: int | NumpyArray[int], value) -> None:
         if dataset is not None and dataset.Get():
             dataset.Get().Modified()
 
-    def __array_wrap__(self, out_arr, context=None, return_scalar: bool = False):
+    def __array_wrap__(self: pyvista_ndarray, out_arr, context=None, return_scalar: bool = False):  # noqa: ANN001, ANN204
```
tkoyama010, line 121: Could you explain us why we don't define the type of `out_arr` ?

tkoyama010 (changes requested):

pyvista/core/pyvista_ndarray.py
user27182: My IDE (PyCharm) gives this type hint by default:
array_wrap

It looked a bit too complex to reproduce, and I'm not sure these annotations would be much benefit in terms of external API type hints for users.

tkoyama010: Got it. Thanks!

tkoyama010 (approved):
LGTM

PR #6953, opened 2024-12-07
Add annotations to `interpolate`, `sample` and `plot` dataset filters

user27182:
### Overview

Part of #6828. Note that the `target` types have been generalized to accept either `DataSet` or `vtkDataSet` inputs. See https://github.com/pyvista/pyvista/pull/6828#discussion_r1854545645


tkoyama010 (approved):
LGTM

PR #6955, opened 2024-12-08
Classify named colors in documentation based on hue, lightness, and saturation 

user27182:
### Overview

This PR is similar to #6594. However, general Hue Saturation Value (HSV) sorting is not ideal since light and dark colors get mixed with saturated colors. Instead, this PR automatically classifies colors based on Hue, Lightness, Saturation (HLS). The aim is to generate tables similar to the `vtkNamedColors` tables here:
https://htmlpreview.github.io/?https://github.com/Kitware/vtk-examples/blob/gh-pages/VTKNamedColorPatches.html

This PR is related to #6592. The classifier from this PR will work on any color, and the tables will automatically expand if new colors are added.

Issue #6591, opened 2024-08-30
Support VTK named colors


user27182:
### Describe the feature you would like to be added.

Enable use of these colors: [VTK Named Colors](https://htmlpreview.github.io/?https://github.com/Kitware/vtk-examples/blob/gh-pages/VTKNamedColorPatches.html)

e.g.
```python
import pyvista as pv
pv.Sphere().plot(color='ultramarine')
```

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

PR #6957, opened 2024-12-08
Add source information to named colors documentation

user27182:
### Overview

Add an extra column to the [Color Tables](https://dev.pyvista.org/api/utilities/color_table.html) to identify the source of the color, e.g. `CSS` web color, `TABLEAU` color, etc.

This allows us to globally sort all colors by name or by color classification (#6955) while making it easy to identify the source of the color.

This should help with #6592, where a new `VTK` source label would be added to those new colors.


user27182:
Update docs with new source info:
image

user27182:
One notable change is that all colors are now sorted together by name (for the first table). So, the tab colors are mixed in with the other colors. With #6591 the VTK colors would also be mixed in.

When using the TAB palette on its own, however, it's usually ordered as:
image

And so we lose this separate TAB ordering with this PR.

We could also make separate the tables for each group? e.g. Have a CSS table, TAB table, and VTK table with #6591.

So we would have 6 tables total when all is done:

1. All colors sorted by name
2. All colors sorted by HSV/HLS
3. CSS colors
4. Tableau colors
5. Paraview colors
6. VTK colors

tkoyama010 (approved):
LGTM

PR #6960, opened 2024-12-08
InteractionEvent consistency

beroda:
### Overview

Not all methods related to widgets use the `_parse_interaction_event` function [here](https://github.com/pyvista/pyvista/blob/3b959865d1493d0f0616790b0628e0a361ddd0e5/pyvista/plotting/widgets.py#L34).

This PR tries to bring consistency across all widgets methods.


pyvista/core/_typing_core/_aliases.py
```diff
@@ -79,3 +80,5 @@ class BoundsTuple(NamedTuple):
 
 # Undocumented alias - should be expanded in docs
 _ArrayLikeOrScalar = Union[NumberType, ArrayLike[NumberType]]
+
+InteractionEventType = Literal['end', 'start', 'always'] | _vtk.vtkCommand.EventIds
```
user27182, line 84: Use old syntax for types used at runtime 
```suggestion
InteractionEventType = Union[Literal['end', 'start', 'always'], _vtk.vtkCommand.EventIds]
```

pyvista/plotting/widgets.py
```diff
@@ -28,15 +28,16 @@
...
 
 
-def _parse_interaction_event(interaction_event):
+def _parse_interaction_event(interaction_event: InteractionEventType):
     """Parse the interaction event.
 
     Parameters
     ----------
-    interaction_event : vtk.vtkCommand.EventIds, str, optional
+    interaction_event : InteractionEventType
```
user27182, line 40: Union-types are auto-expanded by Sphinx. So, in the function signature this will appear as
`Literal['end', 'start', 'always'], _vtk.vtkCommand.EventIds]`, not `InteractionEventType`.

But it won't be auto-expanded in the docstring. So, Sphinx will try to find a ref for `InteractionEventType` which doesn't exist and will cause the build to fail (broken link).

In the docstrings, the type should either be expanded manually, e.g. 
```suggestion
    interaction_event : Literal['end', 'start', 'always'], _vtk.vtkCommand.EventIds
```

Or, you can add the type to the [typing docs](https://dev.pyvista.org/api/core/typing) by doing:
1. Add the type to the "autodoc_type_aliases" dictionary in /doc/source/conf.py
2. Add the type to /doc/core/typing.rst
3. (Optionally) add a short description by defining `InteractionEventType.__doc__` in _`_aliases.py`

This way, Sphinx will be able to find a ref for `InteractionEventType` automatically and auto-link it to the core typing docs.

user27182 (commented):
Some initial comments about the typing which might save a few headaches

pyvista/plotting/widgets.py
beroda: Thanks for the detailed feedback. I added the type in the docs here https://github.com/pyvista/pyvista/pull/6960/commits/f915a4b4be8f6173740bdb37c3caf5182e0af815

tests/plotting/test_widgets.py
```diff
@@ -835,3 +841,163 @@ def test_clear_camera3d_widget(verify_image_cache):
     pl.add_camera3d_widget()
     pl.clear_camera3d_widgets()
     pl.show(cpos='xy')
+
+
+class Test_event_parser:
+    """Class to regroup tests for widgets that use the  `_parse_interaction_event()` function"""
```
user27182, line 847: A lot these function look very similar. Is it possible to parametrize the tests instead? Though I can see there are small differences in the fixtures, mock setup, kwargs passed to widget method, etc., and parametrizing could be a bit complex....

I think it's ok as-is if you think it's not really feasible to parametrize.

user27182 (dismissed):
LGTM. I'm not familiar with the widgets but the changes are sensible. Minor comment about the tests but those changes are optional. 

I haven't reviewed the docs either to check that the new `InteractionEventType` is properly linked but it looks like it's correctly defined and should be working.

beroda:
> I haven't reviewed the docs either to check that the new `InteractionEventType` is properly linked but it looks like it's correctly defined and should be working.

@tkoyama010 could you kindly deploy the doc with `@pyvista-bot preview` on this PR to look if the documentation is built correctly ?
Thanks in advance

tests/plotting/test_widgets.py
beroda: Thank you for the feedback.
Done in https://github.com/pyvista/pyvista/commit/69a1458a216c77abaf98b4f128f823c211c29b1d with `pytest parametrize`, although some readability was lost on the road :angel: .

FWIW, when it comes to write complex test parametrization, I really like the plugin [pytest_cases](https://smarie.github.io/python-pytest-cases/). I managed to do without it here but it is really powerful.

user27182: > Done in [69a1458](https://github.com/pyvista/pyvista/commit/69a1458a216c77abaf98b4f128f823c211c29b1d) with `pytest parametrize`, although some readability was lost on the road 👼 .

Thanks, yes it's hard to find a balance with readability and generalization, I think it reads OK here.

> FWIW, when it comes to writing complex test parametrization, I really like the plugin [pytest_cases](https://smarie.github.io/python-pytest-cases/). I managed to do without it here but it is really powerful.

Good to know, thanks for the tip!

user27182 (approved):

PR #6956, opened 2024-12-08
Mark `test_dataset_loader_source_url_blob` as flaky

user27182:
### Overview

This failure is common:
https://github.com/pyvista/pyvista/actions/runs/12218087918/job/34083120888?pr=6947#step:12:6690
```
FAILED tests/examples/test_downloads.py::test_dataset_loader_source_url_blob[doorman] - AttributeError: type object 'DatasetLoaderTestCase' has no attribute 'dataset_name'
```

PR #6961, opened 2024-12-08
Update exceptions for dataset loader url blob test

user27182:
### Overview

Fix the exceptions caught and re-tried for this test #6956.


tkoyama010 (approved):
Thanks!

PR #6962, opened 2024-12-09
Ignore `F401` rule for all `__init__.py` files

beroda:
### Overview

Move the `unused-import` ruff rule `F401` for all `__init__.py` files


tkoyama010 (approved):
LGTM. Nice fix :)

Issue #6891, opened 2024-11-28
Support for Exodus II files


mattfalcone1997:
### Describe the feature you would like to be added.

It would be very useful if Pyvista could have the ability to read Exodus II files. The particular applications in mind are for post-processing data from the MOOSE framework, which uses exodus II by default.

### Links to VTK Documentation, Examples, or Class Definitions.

Documentation for the [vtkExodusIIReader](https://vtk.org/doc/nightly/html/classvtkExodusIIReader.html). There are some aspects of the vtkExodusIIReader class that probably should be exposed at high level. Other pyvista reader classes often use methods inherited from PointCellDataSelection class to choose which arrays are to be read. However, vtkExodusIIReader does not have the SetPointArrayStatus functions, which the PointCellDataSelection class wraps. Instead, vtkExodusIIReader has many Set<value>ArrayStatus functions, such as SetObjectArrayStatus. It may be useful to expose at least some of these at high level and that its interface is similar to the PointCellDataSelection class.

A basic implementation is given below, in which (I think) the arrays within the element blocks are selected when the filename is set.

```python
import vtk
import pyvista as pv

from pyvista.core.utilities import (BaseReader, 
                                    TimeReader)
                                    
class ExodusIIReader(BaseReader, TimeReader):
    """
    Class that reads the Exodus II format used by MOOSE
    """
    _vtk_module_name = "vtkIOExodus"
    _vtk_class_name = "vtkExodusIIReader"

    def _set_filename(self, filename):
        super()._set_filename(filename)
        # Activate the arrays and by default use the latest time
        
        types = [vtk.vtkExodusIIReader.NODAL,
                 vtk.vtkExodusIIReader.ELEM_BLOCK,
                 vtk.vtkExodusIIReader.GLOBAL]
        
        for type_ in types:
            n = self.reader.GetNumberOfObjectArrays(type_)
            for i in range(n):
                self.reader.SetObjectArrayStatus(type_, i, 1)
                # initialise time step value on last step

    @property
    def number_time_points(self):  # noqa: D102
        """Return number of time points or iterations available to read.

        Returns
        -------
        int

        """ 
        return self.reader.GetNumberOfTimeSteps()
    
    @property
    def time_values(self):
        """All time or iteration values.

        Returns
        -------
        list[float]

        """
        key = vtk.vtkStreamingDemandDrivenPipeline.TIME_STEPS()
        vtkinfo = self.reader.GetExecutive().GetOutputInformation(0)
[...]

PR #6963, opened 2024-12-09
Implement class to read Exodus II files 

mattfalcone1997:
### Overview

Implement class to read Exodus II files with time traversal and enabling point and cell arrays. This will resolve issue #6891.

### Details
- The PointCellDataSelection class is overridden so that the point arrays are enabled through `SetPointResultArrayStatus` and cell arrays arrays through `SetElementResultArrayStatus`.
- Times can be traversed easily unlike `read_exodus`

### To do's before merge with main
- Allow enabling/disabling nodeset/sideset arrays (Not clear currently how to test, example data doesn't include these).
- Deprecation of `read_exodus`?
- Documentation improvements


MatthewFlamm:
Haven't looked deeply, but looks great on a skim.  On your comment in the issue

> Also there seems to be an CI issue with Ruff failing with a numpy2deprecation warning

I don't know what is going on with this ruff error related to numpy.  It only seems to show up when you have other ruff errors.  Once those are cleaned up the numpy error goes away.  You can either fix the other ruff errors manually or comment `pre-commit.ci autofix` here to automatically do it and then pull changes to your local environment.  Reach out if you need help here.

pyvista/core/utilities/reader.py
```diff
@@ -2831,6 +2839,255 @@ class ProStarReader(BaseReader):
...
+    >>> from pyvista import examples
+    >>> filename = examples.download_mug(load=False)
+    >>> reader = pv.get_reader(filename)
+    >>> mesh = reader.read()
+    >>> mesh.plot()
+    """
+    _vtk_module_name = 'vtkIOExodus'
+    _vtk_class_name = 'vtkExodusIIReader'
+
+    def _set_filename(self, filename):
```
MatthewFlamm, line 2861: Better to use either `_set_defaults` or `_set_defaults_post` than overriding `_set_filename`.

pyvista/core/utilities/reader.py
```diff
@@ -2831,6 +2839,255 @@ class ProStarReader(BaseReader):
...
+    def time_values(self):
+        """All time or iteration values.
+
+        Returns
+        -------
+        list[float]
+
+        """
+
+        vtkStreaming = _lazy_vtk_instantiation("vtkCommonExecutionModel",
```
MatthewFlamm, line 3029: Instead of this method, can the methods like `GetNumberOfTimeSteps` and `Get[Set]TimeStep` be used from `vtkExodusIIReader`?

MatthewFlamm (commented):
Some initial comments

pyvista/core/utilities/reader.py
mattfalcone1997: In this case we are getting the time values, whereas I believe GetTimeStep would give you the integer step

MatthewFlamm: `TimeReader` already has this implementation for `time_values`

https://github.com/pyvista/pyvista/blob/fadc7e40d860444d73f3e7d026d77457dabb7660/pyvista/core/utilities/reader.py#L658-L666

If `time_point_value` can leverage `GetTimeStep`, then we can get rid of the pipeline here.

pyvista/core/utilities/reader.py
```diff
@@ -2831,6 +2839,261 @@ class ProStarReader(BaseReader):
...
+    >>> mesh = reader.read()
+    >>> mesh.plot()
+
+    """
+
+    _vtk_module_name = 'vtkIOExodus'
+    _vtk_class_name = 'vtkExodusIIReader'
+
+    def _set_defaults_post(self, filename):
+        super()._set_defaults_post(filename)
```
MatthewFlamm, line 2864: This method doesn't accept a file name and the `super()` is a no-op, so we don't need to call it.  It is only available for Readers to modify the settings before reading the mesh.  `_set_defaults_post` sets the settings after reading the meta data from the file, which is appropriate here.

```suggestion
    def _set_defaults_post(self):
```

mattfalcone1997:
I have an idea of how to deal with the nodeset/sidesets, but I think it will involve a bit of reworking the reader class. The basic idea is to have a separate `ExodusIIObject` to abstract the `SetObjectArrayStatus` etc. methods. This can then be used for all the block and set arrays, with the PointCellDataSelection routines a convenience for the element block arrays. I can provide more info here, but it may be better to implement in a separate pull request.

mattfalcone1997:
I believe the current unit test failure relates to another test that relies on the .exo file extension not having a reader shown below.

https://github.com/pyvista/pyvista/blob/4a915c7125abca321fee64260c175d6bdf8e2fbd/tests/examples/test_dataset_loader.py#L620

With a superficial check, there is another test that checks if the dataset has no reader, so I believe that I can just edit the asserts to account for the `ExodusIIReader` type. Before I push the fix, it will be useful for this to be confirmed.

user27182:
Yes that is a bug. The dataset loaders try to show the reader in the docs. But that only works Reader classes, so is broken with the current `read_exodus` function. This PR will fix this. 

See https://dev.pyvista.org/api/examples/dataset_gallery#exodus-dataset

image

tests/core/test_reader.py
```diff
@@ -1213,3 +1213,110 @@ def test_read_write_pickle(tmp_path, data_object, ext, datasets):
...
+
+    e_reader = pv.get_reader(fname_e)
+    exo_reader = pv.get_reader(fname_exo)
+
+    assert isinstance(e_reader, pv.core.utilities.reader.ExodusIIReader)
+
+    assert isinstance(exo_reader, pv.core.utilities.reader.ExodusIIReader)
+
+    # Focus remainder on mug because it has multiple times
+    # and block arrays
```
user27182, line 1233: Maybe split this into two tests?

```suggestion
def test_exodus_reader_ext():
    # test against mug and exodus to check different valid file
    # extensions: .e and .exo

    fname_e = examples.download_mug(load=False)
    fname_exo = examples.download_exodus(load=False)

    e_reader = pv.get_reader(fname_e)
    exo_reader = pv.get_reader(fname_exo)

    assert isinstance(e_reader, pv.core.utilities.reader.ExodusIIReader)
    assert isinstance(exo_reader, pv.core.utilities.reader.ExodusIIReader)

```suggestion
def test_exodus_reader():
    # Use mug because it has multiple times and block arrays
    e_reader = examples.download_exodus(load=False).get_reader(fname_e)
```

user27182:
To fix some of the build errors the new reader needs to be added to the documentation:
`doc/source/api/readers/index.rst`

The errors are from new links that the dataset loader is trying to add to replace `None` with a link to the new reader in the screenshot I shared above.

MatthewFlamm:
> I have an idea of how to deal with the nodeset/sidesets, but I think it will involve a bit of reworking the reader class. The basic idea is to have a separate `ExodusIIObject` to abstract the `SetObjectArrayStatus` etc. methods. This can then be used for all the block and set arrays, with the PointCellDataSelection routines a convenience for the element block arrays. I can provide more info here, but it may be better to implement in a separate pull request.

It sounds like whatever implementation will be significant for nodeset/sidesteps.  In my opinion, it is best as a separate PR.

doc/source/api/readers/index.rst
```diff
@@ -31,6 +31,7 @@ Reader Classes
     DICOMReader
     EnSightReader
     FacetReader
+    ExodusIIReader
```
user27182, line 34: Sort alphabetically
```suggestion
    ExodusIIReader
    FacetReader
```

pyvista/core/utilities/reader.py
MatthewFlamm: I looked into this, `GetTimeStep` gets the active time step for reading, so it is not applicable here.

pyvista/core/utilities/reader.py
```diff
@@ -2831,6 +2839,259 @@ class ProStarReader(BaseReader):
...
+    @property
+    def active_time_value(self):
+        """Active time or iteration value.
+
+        Returns
+        -------
+        float
+
+        """
+        return self.time_values[self._active_time_point]
```
MatthewFlamm, line 3061: Can we use `GetTimeStep` here instead of having a private attribute?  It is always better to use the vtk method when possible because someone could directly modify the vtk object, and it wouldn't be updated on the PyVista class private attribute.

MatthewFlamm:
I'm wondering if `TimeReader` should have `active_time_point` as a property.  Apparently vtk reader classes are not consistent on whether readers use a time point or a time value.  All the underlying vtk classes whose PyVista Reader inherits from `TimeReader` currently use a value, while the reader in this PR uses the index.  This is probably best as a separate PR, because it would have to be implemented for all the readers that inherit from `TimeReader`.

MatthewFlamm (approved):
This is great!  Thanks!

PR #6964, opened 2024-12-09
Add the EM ruff rule

beroda:
### Overview

Add the error message `EM` rule from ruff (see [here](https://docs.astral.sh/ruff/rules/#flake8-errmsg-em)).


pyvista/core/filters/structured_grid.py
```diff
@@ -125,34 +125,39 @@ def concatenate(self, other, axis, tolerance=0.0):
...
+                    'are not compatible.'
                 )
+                raise ValueError(msg)
 
         # check point/cell variables are the same
-        if not set(self.point_data.keys()) == set(other.point_data.keys()):  # type: ignore[attr-defined]
-            raise RuntimeError('Grid to concatenate has different point array names.')
-        if not set(self.cell_data.keys()) == set(other.cell_data.keys()):  # type: ignore[attr-defined]
-            raise RuntimeError('Grid to concatenate has different cell array names.')
+        if set(self.point_data.keys()) != set(other.point_data.keys()):  # type: ignore[attr-defined]
```
beroda, line 143: Weird, applying the fix `EM` on the code below the `if` triggered the `SIM201` [rule](https://docs.astral.sh/ruff/rules/negate-equal-op/#negate-equal-op-sim201) . Maybe a ruff issue ...
I saw this behavior in other places.

user27182:
+1 for this change. Looks like a bunch of these error branches are untested though. Currently 318 lines missing coverage, I'm guessing we may need to add coverage for a bunch of these or remove the branches altogether to get this PR to pass the 90% coverage requirement.

At a quick glance it looks like a number of these branches could be eliminated altogether by using validation methods instead of doing array length checks, instance checks, etc.

user27182:
> We might need to another PR(s) to increase coverage on raises before using this rule.

I think it's going to be hard to fix the coverage issues indirectly. What about adding some file ignores for now so that we can add this rule, and create a separate issue with a TODO list to remove the ignores? This would be similar to the "ANN" ignores that are currently there.

Looking at the codecov report, these files can probably ignored for now (and reverted). They all have a low patch percentage and all have 8 or more missed lines. Together, this should reduce the missed line count by about half (approx 150)

- pyvista/core/utilities/helpers.py
- pyvista/plotting/utilities/regression.py
- pyvista/plotting/widgets.py
- pyvista/plotting/render_window_interactor.py
- pyvista/plotting/plotter.py
- pyvista/core/dataobject.py
- pyvista/plotting/renderers.py
- pyvista/core/dataset.py

beroda:
Good idea. I'll try to revert those files to reach the 90% coverage threshold.

user27182:
Not sure what the other options are to work around this... it might be possible for admins to push this through anyway, even with codecov failing CI? Or have a separate PR to first lower the coverage threshold so this PR passes, and then revert it so that it's back at 90%?

beroda:
> Not sure what the other options are to work around this... it might be possible for admins to push this through anyway, even with codecov failing CI? Or have a separate PR to first lower the coverage threshold so this PR passes, and then revert it so that it's back at 90%?

Or split this PR in smaller ones to be under the 90% threshold

user27182 (approved):
LGTM. Although there are many untested branches affected by this change, the fact that all changes are automatic and that the branches that _do_ have coverage are all passing gives confidence that the untested branches will not be broken by this PR. 

I also think it would be a lot of additional work to try to work around the codecov failures and may lead to inconsistent application of this rule. So, if it's possible for admins to merge despite having a CI failure, I think it's would be OK to merge this as-is.

doc/source/make_tables.py
```diff
@@ -155,12 +155,14 @@ def generate(cls):
     @classmethod
     def fetch_data(cls):
         """Get a list of row_data used to generate the table."""
-        raise NotImplementedError('Subclasses should specify a fetch_data method.')
+        msg = 'Subclasses should specify a fetch_data method.'
+        raise NotImplementedError(msg)
```
banesullivan, line 159: What's the justification behind all of these changes? This feels wrong to me and unnecessarily changing 1000+ lines just to break up the message and raising the exception?

banesullivan (changes requested):

doc/source/make_tables.py
user27182: This PR will make errors easier to understand and trace back. 
See https://docs.astral.sh/ruff/rules/f-string-in-exception/

pyproject.toml
```diff
@@ -279,6 +279,7 @@ docstring-code-format = false
 quote-style = 'single'
 
 [tool.ruff.lint]
+extend-safe-fixes = ["EM"]
```
user27182, line 282: Following #6987 
```suggestion
extend-safe-fixes = ['EM']
```

doc/source/make_tables.py
banesullivan: Oooh TIL -- okay this has actually botherwed me in many other places. +1 from me now

beroda:
This PR is ready for review, even though the codecov threshold is not reached. See comment here https://github.com/pyvista/pyvista/pull/6964#pullrequestreview-2492992281

pyvista/core/utilities/geometric_sources.py
```diff
@@ -3682,9 +3688,8 @@ def _make_default_part(geometry: str) -> PolyData:
                 must_contain=geometry,
                 name='Geometry',
             )
-            raise NotImplementedError(
-                f"Geometry '{geometry}' is not implemented"
-            )  # pragma: no cover
+            msg = f"Geometry '{geometry}' is not implemented"
+            raise NotImplementedError(msg)  # pragma: no cover
```
user27182, line 3692: Some of the unreported coverage was previously ignored but now since it's split across multiple lines we have one line skipped and the other is not skipped. I've seen this in a few spots, I think we should address this. Either add the skip to both lines, or maybe move the comment up to the closest branch (e.g. one line above, if applicable)
```suggestion
            msg = f"Geometry '{geometry}' is not implemented"  # pragma: no cover
            raise NotImplementedError(msg)  # pragma: no cover
```

pyvista/core/filters/data_set.py
```diff
@@ -8117,12 +8139,14 @@ def partition(  # type: ignore[misc]
         # While vtkRedistributeDataSetFilter exists prior to 9.1.0, it doesn't
         # work correctly, returning the wrong number of partitions.
         if pyvista.vtk_version_info < (9, 1, 0):
-            raise VTKVersionError('`partition` requires vtk>=9.1.0')  # pragma: no cover
+            msg = '`partition` requires vtk>=9.1.0'
+            raise VTKVersionError(msg)  # pragma: no cover
         if not hasattr(_vtk, 'vtkRedistributeDataSetFilter'):
```
user27182, line 8144:
```suggestion
        if not hasattr(_vtk, 'vtkRedistributeDataSetFilter'):  # pragma: no cover
```

pyvista/core/filters/data_set.py
```diff
@@ -8117,12 +8139,14 @@ def partition(  # type: ignore[misc]
...
-            raise VTKVersionError(
+            msg = (
                 '`partition` requires vtkRedistributeDataSetFilter, but it '
-                f'was not found in VTK {pyvista.vtk_version_info}',
-            )  # pragma: no cover
+                f'was not found in VTK {pyvista.vtk_version_info}'
+            )
+            raise VTKVersionError(msg)  # pragma: no cover
```
user27182, line 8149:
```suggestion
            raise VTKVersionError(msg)
```

user27182 (changes requested):
More suggestions based on https://github.com/pyvista/pyvista/pull/6964/files#r1955349805

These changes should only preserve ignored coverage if it was previously ignored. It does not ignore lines that are otherwise still missing coverage.

pyvista/core/filters/data_set.py
```diff
@@ -2968,9 +2982,8 @@ def _post_process_extract_values(before_extraction, extracted):
                 ids_to_remove = list(all_ids - ids_to_keep)
                 if len(ids_to_remove) != 0:
                     if pyvista.vtk_version_info < (9, 1, 0):
```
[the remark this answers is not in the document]
beroda: See https://github.com/pyvista/pyvista/pull/6964#discussion_r1957327946

user27182 (approved):
Not sure if this is ready... but this LGTM. Can't say I've looked closely at every change but I did skim over most of it. 

Might be good to merge this one sooner than later since it's likely that other commits to merge will create merge conflicts here.

tkoyama010 (approved):
LGTM

beroda:
Waiting for https://github.com/pyvista/pyvista/pull/7252 to land before merging this one.
We should not have anymore codecov patch failures after.

user27182 (approved):

tkoyama010:
pre-commit.ci autofix

banesullivan (approved):

PR #6947, opened 2024-12-06
Fully annotate `dataobject.py`

user27182:
### Overview

Add annotations.

PR #6965, opened 2024-12-09
Bump `npt-promote` to resolve `bool` and `np.bool_` type promotions

user27182:
### Overview

See https://github.com/pyvista/pyvista/pull/6947#discussion_r1874940932

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ pinned = [ # Pinned versions of core dependencies
 
 typing = [
   'mypy<1.14.0',
-  'npt-promote==0.2rc1',
+  'npt-promote==0.2',
   'numpy>=2.0.0',
   'pyvista[pinned]',
   'trimesh<4.6.0',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ pinned = [ # Pinned versions of core dependencies
 
 typing = [
   'mypy<1.14.0',
-  'npt-promote==0.1',
+  'npt-promote==0.2rc1',
   'numpy>=2.0.0',
   'pyvista[pinned]',
   'trimesh<4.6.0',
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -60,7 +60,7 @@ pinned = [ # Pinned versions of core dependencies
 
 typing = [
   'mypy<1.14.0',
-  'npt-promote==0.1',
+  'npt-promote==0.2rc1',
```
user27182, line 63: Use `v0.2`
```suggestion
  'npt-promote==0.2',
```

pyvista/core/utilities/geometric_objects.py
```diff
@@ -1979,9 +1979,9 @@ def Rectangle(points: MatrixLike[float] | None = None) -> PolyData:
     if np.isclose(mag_01, 0) or np.isclose(mag_02, 0) or np.isclose(mag_12, 0):
         raise ValueError('Unable to build a rectangle with less than three different points')
 
-    scalar_pdct_01_02 = np.dot(vec_01, vec_02) / min(mag_01, mag_02) ** 2  # type: ignore[call-overload]
-    scalar_pdct_01_12 = np.dot(vec_01, vec_12) / min(mag_01, mag_12) ** 2  # type: ignore[call-overload]
-    scalar_pdct_02_12 = np.dot(vec_02, vec_12) / min(mag_02, mag_12) ** 2  # type: ignore[call-overload]
+    scalar_pdct_01_02 = np.dot(vec_01, vec_02) / min(mag_01, mag_02) ** 2
```
user27182, line 1982: This is an interesting side-effect of the new `bool` <->`np.bool_` type promotions.

The previous type error reported was:
`error: No overload variant of "min" matches argument types "floating[Any]", "floating[Any]"  [call-overload]`
which suggests  that the built-in `min` function doesn't support numpy floats.

I'm not sure why this type error with floats is now no longer considered an error with the new `bool` type promotions in place.

tkoyama010 (approved):
LGTM

PR #6371, opened 2024-07-11
Fix active normals when merging polydata

user27182:
### Overview

Resolves #6365.

This PR also implements a new `_active_normals_name` property in `DataSetAttributes`. It is currently made private so that this PR could be released as a part of a bug fix without adding new API changes.

The private methods could be made public as a new feature in a separate PR for a future release.

PR #6966, opened 2024-12-10
Deprecate `flip_normals` and replace with `flip_faces` and `flip_normal_vectors`

### LINKED ISSUE #6958: Generalize and/or rename `PolyData.flip_normals` using `vtkReverseSense`

[`PolyData.flip_normals`](https://dev.pyvista.org/api/core/_autosummary/pyvista.polydatafilters.flip_normals#pyvista.PolyDataFilters.flip_normals) has a misleading name. It does not operate on normals, but instead reverses the ordering of points. I would have expected this filter to do something more like `mesh.active_normals *= -1`. It also only works with triangles.

Instead, I think the filter should be called something like `reverse_sense` and  use [vtkReverseSense](https://vtk.org/doc/nightly/html/classvtkReverseSense.html#details) to reverse the point order for any polygon, not just triangles. This filter also handles the `normals *= -1` case (though it's simple enough to implement this without a vtk filter).

_No response_

### Pseudocode or Screenshots

_No response_

### LINKED ISSUE #6971: `flip_normals` does not flip normal vectors

`flip_normals` does not flip normal vectors. Instead, it only reverses the cell ordering.

Instead, IMO it should both flip the normal vector direction _and_ reverse cell ordering, which is exactly what the `flip_normals` option does for the `compute_normals` filter.

https://github.com/pyvista/pyvista/blob/695edc4f4acf34c3f66dd6255439af81f2c2d1b5/pyvista/core/filters/poly_data.py#L1635-L1638 

Only reversing cell ordering by default is bad because it makes the implicit definition of the surface normal (implied by the cell's point ordering) inherently inconsistent with any existing normals which may already be defined (and many geometric/paramatric objects have active normals defined by default).

E.g. the current [`Boolean Operations` ](https://dev.pyvista.org/examples/01-filter/boolean-operations.html) example demonstrates the effects of having "flipped normals". But, what it's actually doing is showing the effects of reversing cell ordering, since the normal vectors are currently unaffected by `flip_normals`. And so what's not obvious about this example is that the boolean operations only seem to care about the cell's ordering, and seem to ignore any normal vectors that are defined.

```python
import pyvista as pv 
sphere = pv.Sphere() 
sphere.point_data['Normals'][0]  # pyvista_ndarray([0., 0., 1.], dtype=float32)
sphere.regular_faces[0]  # array([ 2, 30,  0])

sphere.flip_normals() 
sphere.point_data['Normals'][0]  # pyvista_ndarray([0., 0., 1.], dtype=float32)
sphere.regular_faces[0]  # array([ 0, 30,  2])
```

### System Information

```shell
Main branch
```

### Screenshots

_No response_

user27182:
### Overview

Resolve #6958 and fix #6971.

This PR adds a new internal `_reverse_sense` filter to reverse cell ordering and reverse normals direction, and uses it to create two new filters as a replacement for `flip_normals`.


MatthewFlamm:
I'm +1 that this filter needs to be changed/fixed.  I find it a weird that the filter name is `flip_normals` yet there is also a `reverse_normals` parameter that negates the meaning of the filter name.  One thought is that we could add a `reverse_sense` filter [or whatever naming] that has this option and then `flip_normals` becomes a thin wrapper around `reverse_filter` that doesn't allow this parameter to be varied.  I'm not tied to this idea, just up for discussion.

user27182:
> I'm +1 that this filter needs to be changed/fixed. I find it a weird that the filter name is `flip_normals` yet there is also a `reverse_normals` parameter that potentially negates the meaning of the filter name. One thought is that we could add a `reverse_sense` filter [or whatever naming] that has this option and then `flip_normals` becomes a thin wrapper around `reverse_filter` that doesn't allow this parameter to be varied. I'm not tied to this idea, just up for discussion.

A core issue for me is the name of the filter is ambiguous. IMO it's unclear what this operation will do.  So, having both parameter options with this filter sort of gets around this because the parameter names themselves do the disambiguation. This was partly done to limit the scope of this change and avoid deprecations. But, I agree the parameter names are now inconsistent with the filter name.

I like the idea of having a wrapper around an internal implementation of `vtkSenseFilter`, but in this case I'd rather rename the filter altogether and fully deprecate `flip_normals`. So, we could create two filters, e.g. `flip_normal_vectors` and `reverse_cell_ordering`. These names should be self-evident, but the former would use  `reverse_normals=True` and the latter would use `reverse_cells=True`. I can't think of any use cases where both options would be needed with a single method call, but if there are use cases, then it's probably better to make two separate calls to two filters anyway.

MatthewFlamm:
> I like the idea of having a wrapper around an internal implementation of `vtkSenseFilter`, but in this case I'd rather rename the filter altogether and fully deprecate `flip_normals`. So, we could create two filters, e.g. `flip_normal_vectors` and `reverse_cell_ordering`. These names should be self-evident, but the former would use `reverse_normals=True` and the latter would use `reverse_cells=True`. I can't think of any use cases where both options would be needed with a single method call, but if there are use cases, then it's probably better to make two separate calls to two filters anyway.

I'm +1 for this overall suggestion, unless it is common to do both of these most of the time together.  The examples I've seen are related to mirroring datasets where both are used.  This isn't convincing enough to me to keep them as a single method.  Given that the behavior of `flip_normals` feels very strongly like a bug, I'm not sure that we need to have `flip_normal_vectors` unless you think this an improvement in naming.

darikg:
`flip_normal_vectors` seems almost useless since it's so easy to write `mesh.point_data['normals'] *= -1` or whatever, and possibly even easier to read.

user27182:
> > `flip_normal_vectors` seems almost useless since it's so easy to write `mesh.point_data['normals'] *= -1` or whatever, and possibly even easier to read.

> The advantage of `flip_normal_vectors`, or whatever name, is that it will work on whatever array is the active normals. It need not be `'normals'`.

It does seem to have limited benefit, though it is nice not to have to think about the name of the array. And if the name happens to be `"my_vectors"`, then `mesh.flip_normal_vectors()` seems to communicate the behavior better than `mesh.point_data['my_vectors'] *= -1`. The filter should also also flip both point and cell normals.

Also, having two filters (with See Also cross-refs) helps make it clear that the operations are separate and distinct, which helps solve the current ambiguity issue with `flip_normals`.

Related: It may also be good to make `active_normals_name` part of the public API. It's currently private, see https://github.com/pyvista/pyvista/pull/6371

user27182:
`flip_normals` is deprecated, replaced with `reverse_cell_ordering` and `flip_normal_vectors` in [96de207](https://github.com/pyvista/pyvista/pull/6966/commits/96de2076c2a42bb68692e3f7a7515ff71d213236). I'm not tied to these names, looking for feedback. 

Technically the cell ordering itself isn't being reversed, but rather the point ordering, so `reverse_cell_ordering`  is also a bit ambiguous. I have also seen the term "winding" and "orientation" used, not sure what would be clearest and most consistent with other filters.

user27182:
Another option still is to keep `flip_normals` as a single stand-alone filter, much like this PR's earlier implementation, but change the default to match `compute_normals`. 

And maybe use better parameter names, e.g. `flip_direction` and `reverse_ordering`

I'll hold off on making changes until there's consensus on how to name things and separate the functionality.

user27182:
> Another option still is to keep `flip_normals` as a single stand-alone filter, much like this PR's earlier implementation, but change the default to match `compute_normals`.
> 
> And maybe use better parameter names, e.g. `flip_direction` and `reverse_ordering`

This is now my preferred solution. I think we can treat the previous behavior as a bug (see #6971) and simply update the filter without the need for any deprecations.

user27182:
> I'm +1 that this filter needs to be changed/fixed. I find it a weird that the filter name is `flip_normals` yet there is also a `reverse_normals` parameter that potentially negates the meaning of the filter name. One thought is that we could add a `reverse_sense` filter [or whatever naming] that has this option and then `flip_normals` becomes a thin wrapper around `reverse_filter` that doesn't allow this parameter to be varied. I'm not tied to this idea, just up for discussion.

This is done in https://github.com/pyvista/pyvista/pull/6966/commits/34446fc1ed9a9c9d1fbf478358bf7ca6e233ac09. I think this works well if we also modify `flip_normals` to also flip the normal vectors (i.e. fix #6966).

user27182:
Meant to request a review from both of you, not assign it. Looks like I can't request a formal review from @darikg , but your feedback on the changes would be nice to have.

user27182:
pre-commit.ci autofix

darikg:
Looks fine to me, although I have weak preferences for 

1) deprecating flip_normals, mostly just because of the whole "There should be one-- and preferably only one --obvious way to do it" thing.

2) 'flip_faces' or 'reverse_face_orientation' -- I suspect orientation is the more common term than sense -- just poking around a bit to see what other libs use:

- CGAL [reverse_face_orientations](https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__orientation__grp.html#ga21119757bf5835496ae1c9a6d4f5fa0a)
- MeshLib [flipOrientation](https://meshlib.io/documentation/classMR_1_1MeshTopology.html#a66254e26554618bd3b494548ac9a94f0)
- PMPlib [flipFaces](https://www.pmp-library.org/group__algorithms.html#gaab69d401c211502d183849e3778d79a4)
- bmesh [flip_face](https://docs.blender.org/api/current/bmesh.utils.html#bmesh.utils.face_flip)

(of course if all these terms are in the documentation so they can be found by searching it doesn't really matter)

user27182:
> 1. deprecating flip_normals

The deprecation is nice because we have the option to set `inplace=False` by default if we want. I think returning a copy by default is more common with pyvista filters. It also avoids potential silent breakage in case anyone previously relied on the normal vectors _not_ being also flipped for whatever reason.

---
> "There should be one-- and preferably only one --obvious way to do it"

So we should only keep `reverse_sense` (or renamed to `flip_faces`, etc.), and with default options `reverse_cells=True` and `reverse_normals=True`?  And we should  _not_ have a separate wrapper filter? Other than the deprecation and different name, this was effectively what was initially proposed (with CI passing after https://github.com/pyvista/pyvista/pull/6966/commits/4413d43cc56bf143179a9b10c25565c8ca7987ec). But there were concerns about readability if using any kwargs:

>  I find it a weird that the filter name is flip_normals yet there is also a reverse_normals parameter that potentially negates the meaning of the filter name.

Maybe we can truncate the names to address this? E.g. `flip_faces(faces=True, normals=True)`

---
> 2. 'flip_faces' or 'reverse_face_orientation' -- I suspect orientation is the more common term than sense -- just poking around a bit to see what other libs use:

 +1 for `flip_faces`

---

Thanks for the feedback, based on this I'm leaning towards deprecating `flip_normals` and having a single new filter: 
``` python
flip_faces(faces=True, normals=True, inplace=False, progress_bar=False)
```

darikg:
> So we should only keep `reverse_sense` (or renamed to `flip_faces`, etc.), and with default options `reverse_cells=True` and `reverse_normals=True`? And we should _not_ have a separate wrapper filter? Other than the deprecation and different name, this was effectively what was initially proposed (with CI passing after [4413d43](https://github.com/pyvista/pyvista/commit/4413d43cc56bf143179a9b10c25565c8ca7987ec)). But there were concerns about readability if using any kwargs:

Oh I just meant a flip_normal_vectors method which only changes vector data, and a flip_orientations filter that only reverses cells, and if you want to do both you have to call both filters. But again, I don't feel super strongly about it.

user27182:
Thanks for clarifying! I am also fine with having two single-purpose filters. This was initially done in https://github.com/pyvista/pyvista/pull/6966/commits/aedad6d71aed97895699976a4224aea31625b8eb

I like `flip_faces` over `flip_orientation` because it directly implicates the `faces` property, whereas `flip_orientation` is maybe less clear (is it a spatial transformation like `DataSetFilters.reflect` or `DataSetFilters.flip_normal` ?)

One potential downside to have two separate filters like this though is that I don't think all users may understand that the point ordering implicitly defines the normals, or that objects like `pv.Sphere` and `pv.Plane` include active normals by default.

So calling `pv.Sphere().flip_faces()` where only the point ordering is reversed will result in the pre-computed normals having a direction that is inconsistent with the new normal direction implied by the point ordering. This might lead to unexpected results with the shading (which uses the normal vectors), or when using boolean filters (which use the face orientation).

So in that sense, having a single filter do both by default _may_ be more user-friendly. For advanced users, separate kwargs are available to flip one and not the other.

MatthewFlamm:
> The deprecation is nice because we have the option to set inplace=False by default if we want.

I didn't notice this before.  All of filters except for transform look to have a default of False.  This definitely gives extra weight to deprecating this method name and starting afresh.  I have no strong notion whether it should be 1 or 2 filters.

user27182:
Thanks for the feedback. For optimal clarity I think it would be best to have two filters after all. We can address any inconsistency that may result between face orientation and vector orientation with good documentation. 

I am partial towards
`flip_faces` and `flip_normal_vectors`

darikg:
> I didn't notice this before. All of filters except for transform look to have a default of False. This definitely gives extra weight to deprecating this method name and starting afresh. I have no strong notion whether it should be 1 or 2 filters.

+100 on deprecating this now.

> One potential downside to have two separate filters like this though is that I don't think all users may understand that the point ordering implicitly defines the normals, or that objects like pv.Sphere and pv.Plane include active normals by default.

I'm starting to think the greatest strength of this pr is in improving the documentation -- it's the perfect place to explain the difference between flipping normal vectors and flipping face orientations. As long as both docstrings include the right keywords and cross-reference eachother the user in question is going to stumble on one or the other and I think having two distinct methods helps delineate the difference 🙏

darikg:
> I am partial towards `flip_faces` and `flip_normal_vectors`

`flip_normal_vectors` seems unarguably correct. the name of the other one seems impossibly bikesheddable but `flip_faces` is clear and concise. my brief experience earlier looking into what naming conventions other libs used just wants to stress that all of the keywords should be in the docstr -- I think that's 'orientation', 'sense', 'flip', 'reverse', and 'face'.

PR #6969, opened 2024-12-11
Add custom `pyvista-mypy-plugin` to resolve `pyvista` and `vtk` abstract class compatibility issue

user27182:
### Overview

In #6854 and #6918 a bunch of new types were added to resolve a typing problem we have where `DataSet` cannot be used as a valid type wherever `vtkDataSet` is expected since `DataSet` does not inherit from `vtkDataSet` (and cannot inherit, see #6789).  These types were initially integrated in the code in #6870.

However, the new `Concrete...Type` classes are pretty ugly to use, and worse is that sometimes the `Union` is required instead of the `TypeVar`, which makes things really complicated and unclear. It also exposes an internal typing problem to the public typing API .  Things will only get more complex when we want to support more generic typing for mixed `DataSet | MultiBlock` filters, since these cases will likely require a new  TypeVar.

Instead, this PR proposes using a new [`pyvista-mypy-plugin`](https://github.com/pyvista/pyvista-mypy-plugin). The plugin effectively tells `mypy` to pretend that `DataSet` inherits from `vtkDataSet`, `DataSetObject` inherits from `vtkDataObject`, and that `_PointSet` inherits from `vtkPointSet`. These can be customized and expanded as needed.

This plugin has very similar functionality to [`npt-promote`](https://github.com/pyvista/npt-promote), but the two plugins are currently separate since `npt-promote` can be used generically for any `numpy` project without a `vtk` dependency.


doc/source/api/core/typing.rst
```diff
@@ -76,41 +76,34 @@ pyvista.TransformLike
...
-Concrete Data Object Types
---------------------------
-
-pyvista.ConcreteGridType
-~~~~~~~~~~~~~~~~~~~~~~~~
-
-.. currentmodule:: pyvista
-
-.. autotypevar:: ConcreteGridType
+Data Object Types
```
user27182, line 79: Now that the type definitions are much simpler I'm not so sure if these types should be published anymore. 

The main reason to include types here is to make them auto-linked. But this only works in docstrings, not function signatures (the only place these types are likely to be used)

Plus, the names I think are now too similar to the actual classes, and so having `pyvista.DataSetType` in the namespace alongside `pyvista.DataSet` is not ideal.

Suggest removing all types from the docs and removing from the `pyvista` namespace. Can probably move the types to their respective modules now, e.g. move `DataSetType` to `core.dataset.py`

darikg:
really interesting approach to an incredibly thorny problem i've tried to solve before and gave up on. do you see this as a permanent fixture or a stopgap solution? (not a pointed question! genuinely curious)

user27182:
I'm hopeful that this can be a permanent fix. I think it's a really clean solution, especially compared to using the Concrete types. I've already tested the plugin with #6972 and seems to be working well and as expected.

Once the typing stabilizes a bit I'm thinking more long term we would publish the plugin for use by downstream projects. In that sense it would be a fairly permanent part of the project. `npt-promote` could be merged into it as well, just install the one plugin and you're good to go for typing with PyVista

user27182:
Turns out that typing for wrapper classes is still an open issue in `mypy` and is a popular feature request:
https://github.com/python/mypy/issues/5523

In that thread, a possible solution from the devs is to use `_promote`. That's what the `pyvista-mypy-plugin` uses. So, the plugin is consistent with what the mypy developers suggest. However, this _is_ technically an undocumented feature.

As such, this solution may still be considered as a workaround since it's not (yet) supported by `mypy` officially. No solution seems perfect, but I think this plugin is the most promising way to support this for now.

user27182:
One downside to the current plugin implementation is that the types are hard-coded. If we need to add new promotions or modify existing ones we need to modify the plugin, issue a new release, bump it in pyvista... etc. 

Instead, I think I found a way to make the plugin generic with a decorator, e.g.:

``` python
from mypy_plugin import promote_type
from vtk import vtkDataSet

@promote_type(vtkDataSet)
class DataSet: ...

```

If there is an interest in this kind of plugin I will modify this PR accordingly. Since this behavior is much more generic I'll probably rename the plugin as well, maybe `ducktype-plugin` or `mypy-ducktype`

doc/source/api/core/typing.rst
```diff
@@ -81,43 +81,3 @@ pyvista.InteractionEventType
 .. currentmodule:: pyvista
 
 .. autodata:: InteractionEventType
-
-
-Concrete Data Object Types
---------------------------
-
```
user27182, line 88: Maybe we keep the docs after all, but rename the TypeVars to `_DataSetType`, etc. to avoid polluting the `pyvista` namespace

pyvista/core/dataset.py
```diff
@@ -170,6 +170,7 @@ def __eq__(self: ActiveArrayInfo, other: object) -> bool:
         return False
 
 
+@promote_type(_vtk.vtkDataSet)
 @abstract_class
 class DataSet(DataSetFilters, DataObject):
```
user27182, line 175: Highlighting this to demonstrate how the plugin works. Use the decorator to promote `DataSet` as a `vtkDataSet`
This is similarly done for `DataObject` and `vtkDataObject`

user27182:
The plugin now makes use of a decorator for type promotion. This makes it easy to use and extend to other classes later without changing the plugin.

The downside to this though is it adds a run-time dependency. This would have required adding the plugin as a new core dependency. Instead, the plugin is now directly integrated into pyvista.

user27182:
@tkoyama010 looking for some feedback with this PR. Are we interested in using a mypy pluging to solve this typing issue? IMO this is a nice solution because the current `Concrete` types are not easy to work with. Using a plug-in (or not) is currently holding up other PRs for adding return types (see #6972) since the return types depend on the `TypeVar` we are using.

If we do want a plugin, do we want it directly integrated with pyvista, or do we want to keep it as an independent mypy plugin? If we keep the plugin independent, this will require adding it as a new core dependency since it uses decorators that need to work at runtime. If we prefer to integrate it directly with pyvista, there is no new dependency requirement, but there are some challenges with the testing (mypy path related) that I'll need to work through.

Please let me know if you have any preferences for moving this PR forward, or if it's better to close it in case the plugin is not desired.

tkoyama010:
Please let me open this so I can put this on my wish list.

pyproject.toml
```diff
@@ -241,7 +242,11 @@ extra_checks = true
 ignore_missing_imports = true
 no_implicit_reexport = true
 packages = ['pyvista']
-plugins = ['npt_promote', 'numpy.typing.mypy_plugin']
+plugins = [
+  'npt_promote',
+  'numpy.typing.mypy_plugin',
+  'pyvista.typing.mypy_plugin',
```
user27182, line 248: Note the similarity in how both the numpy and pyvista plugins are added. This consistency is nice and is only possible by including the plug-in directly as part of the pyvista repo.

user27182:
> Do we want this plugin directly integrated with pyvista, or do we want to keep it as an independent mypy plugin?


+1 for keeping it directly with the repo. It has some nice benefits like the one mentioned in https://github.com/pyvista/pyvista/pull/6969#discussion_r1956571492.

pyproject.toml
```diff
@@ -81,6 +81,7 @@ test = [
   'pyanalyze<=0.13.1',
   'pytest-cov<6.1.0',
   'pytest-memprof<0.3.0',
+  'pytest-mypy-plugins<3.2.0',
```
user27182, line 84: This package is not really needed. It has its own separate test mechanism that was previously used but these tests have since been migrated into regular `pytest` test. 

But, if we remove the package then our pytest tests fail. I think it's because this plugin modifies the mypy paths, which allows our pytest tests to work.

So we can keep this as a dependency for that purpose, or we can try to track down the mypy path issue so that we don't need this as a dependency.

user27182: > This package is not really needed ... we can try to track down the mypy path issue so that we don't need this as a dependency.

I've tried to fix this a few times and haven't been able to fix the path issue. We're stuck with keeping this as a dependency.

tkoyama010 (approved):
LGTM

user27182:
It seems now NumPy's plugin is deprecated
https://github.com/pyvista/pyvista/actions/runs/13706466015/job/38332835618?pr=7276#step:15:70
``` python
ERROR tests/typing/test_typing.py - DeprecationWarning: `numpy.typing.mypy_plugin` is deprecated, and will be removed in a future release. Please remove `plugins = numpy.typing.mypy_plugin` in your mypy config.(deprecated in NumPy 2.3)
```

Issue #5401, opened 2023-12-27
Follow Scientific Python Library Development Guide


tkoyama010:
### Describe what maintenance you would like added.

> I've adopted their `pre-commit` hook over on `geovista`, see [here](https://github.com/bjlittle/geovista/blob/main/.pre-commit-config.yaml#L97-L111) ... and documented (for my future self and sanity 😉) the sections that I still require to comply with.
> 
> _Originally posted by @bjlittle in https://github.com/pyvista/pyvista/issues/5397#issuecomment-1869836631_

Let's follow [Scientific Python Library Development Guide](https://learn.scientific-python.org/development/) as [geovista](https://github.com/bjlittle/geovista) did.

### Links to source code.

None

### Pseudocode or Screenshots

None

PR #6970, opened 2024-12-11
Add pre commit hook sp-repo

beroda:
### Overview

Closes https://github.com/pyvista/pyvista/issues/5401


pyproject.toml
```diff
@@ -431,7 +431,9 @@ required-imports = [
...
 [tool.repo-review]
-# MY101: to be removed once https://github.com/pyvista/pyvista/issues/6279 is closed
-# PC140: mypy deliberately removed from pre-commit in https://github.com/pyvista/pyvista/pull/6741
-# RTD: the documentation is not hosted by ReadTheDocs
-ignore = ["MY101", "PC140", "RTD"]
+ignore = [
+  "MY101", # to be removed once https://github.com/pyvista/pyvista/issues/6279 is closed
+  "PC111", # ruff is taking care of docstring formatting https://github.com/pyvista/pyvista/pull/6991
+  "PC140", # mypy deliberately removed from pre-commit in https://github.com/pyvista/pyvista/pull/6741
+  "RTD",   # RTD: the documentation is not hosted by ReadTheDocs
```
user27182:
```suggestion
  'MY101', # to be removed once https://github.com/pyvista/pyvista/issues/6279 is closed
  'PC111', # ruff is taking care of docstring formatting https://github.com/pyvista/pyvista/pull/6991
  'PC140', # mypy deliberately removed from pre-commit in https://github.com/pyvista/pyvista/pull/6741
  'RTD',   # RTD: the documentation is not hosted by ReadTheDocs
```

user27182:
`ERROR: Unknown config option: image_cache_dir`
This already ignored here:
https://github.com/pyvista/pyvista/blob/339d0904add5c67a4b323b3a17cd40c500561056/tests/conftest.py#L275
But it looks like the new strict options don't like it. Does this also need to be added to `filterwarnings` ? Or maybe it's just easier to add `pytest-pyvista` as a docs dependency to resolve this?

beroda:
> `ERROR: Unknown config option: image_cache_dir` This already ignored here:
> 
> https://github.com/pyvista/pyvista/blob/339d0904add5c67a4b323b3a17cd40c500561056/tests/conftest.py#L275
> 
> But it looks like the new strict options don't like it. Does this also need to be added to `filterwarnings` ? Or maybe it's just easier to add `pytest-pyvista` as a docs dependency to resolve this?

Indeed, the `--config-strict` option directly raises an error. No other option than adding `pytest-pvvista` to `docs` extra deps. See https://github.com/pyvista/pyvista/pull/6970/commits/519e9a4dbb7a08324abdf179e1577c4cbc05ee6d

user27182:
> No other option than adding `pytest-pvvista` to `docs` extra deps.

I suppose there is another option to pass the `image_cache_dir` arg explicitly in the unit testing workflows when invoking `pytest` instead of putting it in `pyproject.toml`, though this would force devs to explicitly set this every time locally.

Or we could set an environment variable in `conf.py` called `BUILDING_DOCUMENTATION` that could work similarly to `typing.TYPE_CHECKING`  and do something like `if not BUILDING_DOCUMENTATION: image_cache_dir = ...` so that `image_cache_dir` is not set when building the docs.

These workarounds seem a bit too complex for what could otherwise be a simple fix, seems harmless enough to just add `pytest-pyvista` as a dependency

tkoyama010 (approved):
LGTM

PR #6973, opened 2024-12-12
Switch to trigger for PR label action

banesullivan:
Switch from `pull_request_target` to `pull_request` as the even trigger for the labelling GitHub action.

Using `pull_request_target` puts us in a position where someone could open a malicious PR modifying that action to run any arbitrary code with full access to our GitHub token and other secrets

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,6 +1,6 @@
 name: "Pull Request Labeler"
 on:
-  pull_request_target:
+  pull_request:
     types: [opened, reopened]
 
 jobs:


tkoyama010 (approved):
~Thanks!~

tkoyama010 (changes requested):
Please wait until I can remember about this.

MatthewFlamm:
According to https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

Labeling PRs is the exact reason this trigger is available.  Has something changed that makes it risky?

PR #6975, opened 2024-12-14
Replace `DataSet` with `Self` in `dataset.py`

user27182:
### Overview

Using `Self` is best practice.

https://docs.python.org/3/library/typing.html#typing.Self


tkoyama010 (approved):
Thanks!

MatthewFlamm:
I'm okay with this, but I don't think 'recommended' is accurate.  This is only true when returning the same type of `self`.  https://peps.python.org/pep-0673/ says 

> Note that specifying self: Self is harmless, so some users may find it more readable to write the above as:
> ```
>  class Shape:
>     def difference(self: Self, other: Self) -> float: ...
> ```

MatthewFlamm:
Since `typing.Self` was introduced in Python 3.11 does this addition mean that downstream users cannot use typing in prior versions?

pyvista/core/dataset.py
```diff
@@ -45,6 +45,8 @@
     from collections.abc import Generator
     from collections.abc import Iterator
 
+    from typing_extensions import Self
```
MatthewFlamm, line 48: I see this now is for backwards compatibility.  Should we be testing for backwards compatibility?  I started this #6980

user27182:
> I'm okay with this, but I don't think 'recommended' is accurate.  This is only true when returning the same type of `self`.  https://peps.python.org/pep-0673/ says 
> 
> 
> 
> > Note that specifying self: Self is harmless, so some users may find it more readable to write the above as:
> 
> > ```
> 
> >  class Shape:
> 
> >     def difference(self: Self, other: Self) -> float: ...
> 
> > ```

Fair enough. The typing library does not explicitly recommend `Self` everywhere. 

I'll reword it as _I_ recommend using `Self`. 

I recommend using it for consistency (users don't need to specifically think about whether anything is returned). 

Plus, it's not _only_ useful for methods that return something. It's useful anywhere a `TypeVar` is useful. This includes methods like `copy_from(self: Self, other: Self)`. We want to type hint that self and the object being copied from should have the same type.

Since these subtleties are difficult to communicate across code, I think it's just better to consistently use it without having to technically think about whether a `TypeVar` is specifically needed. 

Also, that first example in the docs for `Self` show using it _only_ as a return type. According to mypy this a type error: a type var should only be returned if it's also used to annotate one of the inputs. So I wouldn't necessarily take the python docs as gospel either.

tkoyama010:
My apologies. The lock was my operational error.

PR #6979, opened 2024-12-14
Allow using delimiters with named colors

user27182:
### Overview

Allow spaces, hyphens, and underscores as input when creating a `Color` from a string. E.g. allow `dark blue`. Part of #6592.


tkoyama010 (approved):
LGTM

PR #6980, opened 2024-12-14
Type check on multiple Python versions

MatthewFlamm:
### Overview

Run type checking in supported versions.

### DIFF
--- a/.github/workflows/type-checking.yml
+++ b/.github/workflows/type-checking.yml
@@ -14,6 +14,7 @@ jobs:
   mypy:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: [3.9, "3.10", "3.11", "3.12"]
     steps:
--- a/.github/workflows/type-checking.yml
+++ b/.github/workflows/type-checking.yml
@@ -13,13 +13,16 @@ env:
 jobs:
   mypy:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.9, "3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v4
 
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.12
+          python-version: ${{ matrix.python-version }}
           cache: "pip"
 
       - name: Install PyVista test requirements


MatthewFlamm:
Less failures than I expected...  Unused type ignores seems like a hard one to handle across Python versions.  Maybe there is a way we can turn off the mypy checks depending on python version?

user27182:
This would be nice to have. It's been discussed before (https://github.com/pyvista/pyvista/pull/6738#issue-2614739247).

> Unused type ignores seems like a hard one to handle across Python versions. Maybe there is a way we can turn off the mypy checks depending on python version?

Unfortunately I don't think it's possible to override a config option via command line, i.e. I was hoping we could do something like `mypy --warn-unused-ignores=false` but this does not seem possible.

If unused ignores is the only issue, we could remove it from the config, and only enable it for CI with the latest python. i.e. for 3.12 we do `mypy --warn-unused-ignores` but only run `mypy` otherwise for older python. This would also have the effect of disabling this check locally as well (not a big deal, the CI will still catch it).

I also think the effectiveness of these checks is limited by our not-so-strict config. We currently have `ignore_missing_imports = true`. So if `mypy` tries to import `Self`, but can't, it'll just ignore it, treat it as `Any`, and move on. So IMO if that's the kind of thing we want to check, then we should first remove `ignore_missing_imports = true`.

The cause of the unused ignores we're seeing may in fact be that something could not be imported, so was treated as `Any`, and so any ignores related to this import would no longer apply.

user27182:
Can probably use `disable-error-code` to disable unused imports code for older versions (if needed):
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disable-error-code

Also, instead of using a matrix with four runners, we can run everything on 3.12 with a single runner and use `--python-version` instead
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version

Would probably speed things up since everything would already be installed and mypy analysis would be cached after the first run

MatthewFlamm:
> Can probably use disable-error-code to disable unused imports code for older versions (if needed):
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disable-error-code

Another thought is to use a controlling config file stored in the `.github` folder that is used for CI only.  Then we can use the `--config-file` in the CI runs to override the config.  This would require keeping the CI files in sync, but so will using the command line method.

Another wrinkle is that our typing is likely vtk (and maybe numpy) version specific (could be related to some of the errors we are already seeing).  These are not easy problems to solve in duck-typed Python unless we narrow our supported versions significantly.

My current thought is that we document that typing is enforced on "Python version X, and pinned dependencies from `pip install pyista[typing]`.  But, we type check using other python versions with a scaled down version to make sure that there is no catastrophic import error.

Maybe I'm overthinking this though.  What is the consequence of having `from typing import Self`? in the code, if it is wrapped into a `if TYPE_CHECKING:` block? Would mypy treat it as `Any` and thus not cause a big problem?

user27182:
> Another thought is to use a controlling config file stored in the `.github` folder that is used for CI only. Then we can use the `--config-file` in the CI runs to override the config. This would require keeping the CI files in sync, but so will using the command line method.

Keeping things in sync seems to be a big pain and could be a source of confusion. I think it would generally be preferred if we can just run `mypy` without special considerations. Maybe for ease of maintenance we rule out any options that require doing otherwise?

> Another wrinkle is that our typing is likely vtk (and maybe numpy) version specific (could be related to some of the errors we are already seeing). These are not easy problems to solve in duck-typed Python unless we narrow our supported versions significantly.

Yup, this could also be a source of (hard to track) type errors and would be difficult to maintain.

> My current thought is that we document that typing is enforced on "Python version X, and pinned dependencies from `pip install pyista[typing]`. 

Yes I think this is best. Our CI runs within a single environment, and that's our gold-standard for type-checking.

> But, we type check using other python versions with a scaled down ~version~ config to make sure that there is no catastrophic import error.

Does this require a completely separate config? Might be easy enough to use the existing config but append ``--python-version 3.X`` in the CLI invocation, see
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version. Not sure if this is a good enough alternative to testing with the real thing.

I quickly tried this locally and `mypy` seems to be happy with `3.12`, `3.11`, and `3.10`, but 325 errors with `3.9` (might be a TypeVar issue with our `NumpyArray` definitions).

> Maybe I'm overthinking this though. What is the consequence of having `from typing import Self`? in the code, if it is wrapped into a `if TYPE_CHECKING:` block? Would mypy treat it as `Any` and thus not cause a big problem?

This just depends on the level of strictness we want for our config. In general, importing as `Any` is a bad thing for type-checking, and we definitely want to catch this and avoid it (and/or flag it explicitly as an error that is ignored). So in that regard we want `mypy` to let us know when it can't import something. In terms of our "gold standard" CI this is what should strive for. For any other environment (e.g. different python version, missing dependencies) maybe importing `Any` is fine though, in which case options like `--ignore-missing-imports` are helpful because they can get rid of nuisance errors which otherwise may not be resolvable.

For things like `Self`, if we strictly import it inside a `TYPE_CHECKING` block then we're probably OK importing from `typing` and not from `typing_extensions` (depending on how strictly we want to support older python). But for runtime type definitions, it will be necessary to import from `typing_extensions`, e.g. if using `cast(Self, var)`. But this would also be picked up by our unit tests since it would be an import error. So for simplicity, we could adopt a convention of: "Always import from `typing` unless necessary for runtime type-checks". That being said, I don't think there's much harm in doing the opposite either, and adopting a convention of always importing from `typing_extensions`. That way we never really have to think about compatibility issues.


> Maybe I'm overthinking this though.

I think perhaps we are over thinking it. We should strive to first and foremost get our typing sorted out with our current CI set-up. Strict typing support for any other environment is secondary. The type-checking won't fundamentally break for any other environment, but it's possible that `mypy` will complain about missing imports and/or unused ignores.

MatthewFlamm:
I'm going to close this.  It isn't clear to me what the strategy should be, but it was useful to see that there are no showstopping issues with the current state.  Once the typing has settled down a bit, we could consider a `workflow_dispatch` trigger to test the other python versions for information only.

PR #6982, opened 2024-12-14
Add mypy warn unreachable

beroda:
### Overview

As recommended by the scientific python development guide [here](https://learn.scientific-python.org/development/guides/style/#MY103), [unreachable warnings ](https://mypy.readthedocs.io/en/stable/common_issues.html#unreachable-code)should be enabled for mypy.

Part of https://github.com/pyvista/pyvista/pull/6970


pyvista/core/utilities/fileio.py
```diff
@@ -239,12 +239,13 @@ def read(
     if file_format is not None and force_ext is not None:
         raise ValueError('Only one of `file_format` and `force_ext` may be specified.')
 
-    if isinstance(filename, (list, tuple)):
+    if isinstance(filename, Sequence) and not isinstance(filename, str):
```
beroda, line 242: Slightly changing the condition because
```python
isinstance("", Sequence) == True
```

pyvista/core/filters/data_set.py
```diff
@@ -2170,7 +2170,7 @@ def contour(  # type: ignore[misc]
         if isinstance(scalars, str):
             scalars_name = scalars
         elif isinstance(scalars, (Sequence, np.ndarray)):
-            scalars_name = 'Contour Data'
+            scalars_name = 'Contour Data'  # type: ignore[unreachable]
```
user27182, line 2173: scalars is typed `str | None` here. The docstring says array scalars are supported. Might just need to update the allowed type, e.g.   `str | VectorLike[float] | None` (assuming only 1D scalars are supported)

pyvista/core/filters/data_set.py
```diff
@@ -2044,7 +2044,7 @@ def elevation(  # type: ignore[misc]
 
     def contour(  # type: ignore[misc]
         self: ConcreteDataSetType,
-        isosurfaces: int = 10,
+        isosurfaces: int | Sequence[float] = 10,
```
user27182, line 2047: Can we use `VectorLike[float]` here? (to match typing for other vector-like inputs)
```suggestion
        isosurfaces: int | VectorLike[float] = 10,
```

pyvista/core/filters/rectilinear_grid.py
```diff
@@ -94,7 +94,7 @@ def to_tetrahedra(
         alg.SetRememberVoxelId(pass_cell_ids or pass_data)
         if mixed is not False:
             if isinstance(mixed, str):
-                self.cell_data.active_scalars_name = mixed
+                self.cell_data.active_scalars_name = mixed  # type: ignore[attr-defined]
```
user27182, line 97: typing `self: RectilinearGrid` here might fix this error, though a separate "sweep" of this file still need to be done to add these annotations to all filters It's fine if you leave it / don't change it

pyvista/core/pointset.py
```diff
@@ -3374,7 +3374,7 @@ def cell_coords(
         else:
             if isinstance(coords[0], np.ndarray):
                 return np.stack(coords, axis=1)
-            return np.asanyarray(coords)
+            return np.asanyarray(coords)  # type: ignore[unreachable]
```
user27182, line 3377: This logic seems odd. `https://numpy.org/doc/stable/reference/generated/numpy.unravel_index.html` seems to always return a tuple of arrays. Perhaps what was intended was to check if the array was scalar or not?
``` diff
- if isinstance(coords[0], np.ndarray):
+ if coords[0].ndim > 0:
```
To check if it's a scalar or array (and stack it if it is an array)

pyvista/core/utilities/arrays.py
```diff
@@ -248,12 +248,19 @@ def convert_array(
 def convert_array(
     arr: npt.ArrayLike, name: str | None = ..., deep: bool = ..., array_type: int | None = None
 ) -> _vtk.vtkAbstractArray: ...
+@overload
+def convert_array(
+    arr: None,
+    name: str | None = None,
+    deep: bool = False,
+    array_type: int | None = None,
```
user27182, line 256: The overloads are tricky, I _think_ using ellipses here will make the overload work in cases like `deep=True` or something like that.

```suggestion
    name: str | None = ...,
    deep: bool = ...,
    array_type: int | None = ...,
```

pyvista/core/dataobject.py
```diff
@@ -628,7 +629,7 @@ def _config_user_dict(self: Self) -> None:
                 # When loaded from file, field will be cast as pyvista ndarray
                 # Convert to string and initialize new user dict object from it
                 self._user_dict = _SerializedDictArray(''.join(array))
-            elif isinstance(array, str) and repr(self._user_dict) != array:
+            elif isinstance(array, str) and repr(self._user_dict) != array:  # type: ignore[unreachable]
```
user27182, line 632: This is fine to ignore here. I suspect some of these issues come from the `convert_array` overloads not allow a `str` return type (which is only used for things like `user_dict`

pyvista/core/composite.py
```diff
@@ -39,7 +39,7 @@
 
     from ._typing_core import NumpyArray
 
-_TypeMultiBlockLeaf = Union['MultiBlock', DataSet, None]
+_TypeMultiBlockLeaf = Union['MultiBlock', DataSet]
```
user27182, line 42: This change is a bit concerning since `None` was only recently added here. https://github.com/pyvista/pyvista/pull/6852
`None` is a valid leaf, so I think we should always expect to handle `None` blocks (and have `mypy` warn us when we're not handling it properly)

pyvista/core/dataobject.py
```diff
@@ -161,8 +161,8 @@ def _write_vtk(mesh_: DataObject) -> None:
             set_vtkwriter_mode(vtk_writer=writer, use_binary=binary)
             writer.SetFileName(str(file_path))
             writer.SetInputData(mesh_)
-            if isinstance(writer, _vtk.vtkPLYWriter) and texture is not None:
-                mesh_ = cast(pyvista.DataSet, mesh_)
+            if isinstance(writer, _vtk.vtkPLYWriter) and texture is not None:  # type: ignore[unreachable]
+                mesh_ = cast(pyvista.DataSet, mesh_)  # type: ignore[unreachable]
```
user27182, line 165: This is OK. These annotations were added recently and I recall the typing here was challenging (hence the cast, too)

pyvista/core/dataset.py
```diff
@@ -1655,7 +1655,7 @@ def copy_from(self: Self, mesh: _vtk.vtkDataSet, deep: bool = True) -> None:
                 f'The Input DataSet type {type(mesh)} must be '
                 f'compatible with the one being overwritten {type(self)}',
             )
-        if deep:
+        if deep:  # type: ignore[unreachable]
```
user27182, line 1658: This looks like `mypy` is assuming the previous line `if not isinstance(self, type(mesh)):` is always true. Maybe there's a way to modify the check? E.g. maybe `issubclass` instead of `isinstance`?

user27182 (commented):
+1 for adding this config option. It's good to add it early too, since a lot of the reported errors are with type hints that were just recently added. Better to catch these now before the problem gets worse with more type hints.

I have a few comments / suggestions. No need to act on all of them, it's fine to keep ignores if the required fix is otherwise out of scope. Removing `None` from the multiblock leaf type is a bit problematic though.

pyvista/core/_validation/check.py
```diff
@@ -234,9 +234,9 @@ def check_sorted(
         except ValueError:
             raise ValueError(f'Axis {axis} is out of bounds for ndim {ndim}.')
 
-    if axis is None and ndim >= 1:
+    if axis is None and ndim >= 1:  # type: ignore[unreachable]
```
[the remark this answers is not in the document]
beroda: Tried but it produced more typing errors in downstream functions. Might be a bit too much for this PR

pyvista/core/composite.py
```diff
@@ -904,10 +902,10 @@ def clean(self: MultiBlock, empty: bool = True) -> None:
         for i in range(self.n_blocks):
             if isinstance(self[i], MultiBlock):
                 # Recursively move through nested structures
-                self[i].clean()  # type: ignore[union-attr]
-                if self[i].n_blocks < 1:  # type: ignore[union-attr]
+                self[i].clean()
+                if self[i].n_blocks < 1:
                     null_blocks.append(i)
-            elif self[i] is None or empty and self[i].n_points < 1:  # type: ignore[union-attr]
+            elif self[i] is None or empty and self[i].n_points < 1:
```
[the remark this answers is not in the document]
beroda: You were right, mypy was not inferring the indexed type. See 6afa9f51b18117fc5a30337892158427fabb9601

pyvista/core/dataset.py
user27182: this is ok, we can fix later

pyvista/core/_validation/check.py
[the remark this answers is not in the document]
user27182: ok, we can fix later

pyvista/core/filters/rectilinear_grid.py
user27182: this is ok to fix later

pyvista/core/pointset.py
user27182: This is also ok to fix later

user27182 (dismissed):
LGTM, thanks!

pyvista/plotting/lookup_table.py
```diff
@@ -199,7 +201,7 @@ class LookupTable(_vtk.vtkLookupTable):
     """
 
     _nan_color_set = False
-    _cmap = None
+    _cmap: colors.Colormap | colors.ListedColormap | None = None
```
user27182, line 204: There's a broken link for `colors.ListedColormap` :
```
Build error: `/pyvista/plotting/lookup_table.py:docstring of pyvista.LookupTable.cmap:1: WARNING: py:class reference target not found: colors.ListedColormap [ref.class]
```

Might need to add it to `nitpick_regex_ignore` in the docs `conf.py`

beroda: Thanks , done in https://github.com/pyvista/pyvista/commit/bcdbac94fa277e1e647d3d9432034ae5f4aafbe9

beroda:
> LGTM, thanks!

Thank you very much for you thorough review

PR #6983, opened 2024-12-15
Do not ignore unused imports `F401` (ignore useless imports `PLC0414` instead)

user27182:
### Overview

Unused imports [F401](https://docs.astral.sh/ruff/rules/unused-import/) should not be ignored. If they are not used, they should be removed. However, we currently ignore this rule in a few places, mostly in `init` files and for `_vtk.py` imports where the "unused imports" are actually being used for re-export. This PR explicitly re-exports these imports so that they are no longer considered unused, allowing the `F401` ignores to be removed.

However, re-exporting like this is considered "useless" by [PLC0414](https://docs.astral.sh/ruff/rules/useless-import-alias/) and is inherently incompatible with this change. This effectively makes it such that `F401` ignores were replaced with `PLC0414` ignores in many places by this PR. However, we now get all the benefits of removing unused imports everywhere while also preventing redundant imports outside of `init` and `_vtk` imports.

EDIT: 
This is indirectly related to #6279 to enable  mypy's `no_implicit_reexport=True` option.

From main, `mypy --no-implicit-reexport` returns:
```Found 1807 errors in 87 files```

From this branch, it returns 
```Found 50 errors in 28 files```

### CODE REVIEW

File: pyvista/plotting/plotter.py [hunk 1]
```diff
@@ -72,9 +72,6 @@
 from .text import TextProperty
 from .texture import numpy_to_texture
 from .themes import Theme
-from .tools import normalize  # noqa: F401
-from .tools import opacity_transfer_function  # noqa: F401
-from .tools import parse_font_family  # noqa: F401
```
user27182, line 77: I think these were added in #4639. Not sure if or why the import needed though...

user27182: Actually, it was added in #1962 for an example, not sure if it's still needed, will add back if CI fails

user27182:
pre-commit.ci autofix

pyvista/jupyter/__init__.py
```diff
@@ -42,7 +40,7 @@ def _validate_jupyter_backend(backend):
 
     if backend in ['server', 'client', 'trame', 'html']:
         try:
-            from pyvista.trame.jupyter import show_trame
+            from pyvista.trame.jupyter import show_trame as show_trame
```
tkoyama010, line 43: Is it possible to use `importlib.util.find_spec` here too?

tkoyama010 (changes requested):

pyproject.toml
```diff
@@ -358,13 +358,13 @@ ignore = [
   "SIM117",  # https://github.com/pyvista/pyvista/pull/5888
   "SIM118",  # https://github.com/pyvista/pyvista/pull/5837
 ]
-unfixable = []
+unfixable = ['F401']
```
user27182, line 361: I want to highlight this change. Unused imports will no longer be auto-fixed (removed). The auto-fix is nice in many instances, but I find it to be a bit aggressive in others. When I'm mid-development (and have unused imports that I plan to use later) I like to use pre-commit to clean up the formatting even though I'm not yet ready to commit. But this always removes the imports when I don't want them removed.

pyvista/jupyter/__init__.py
user27182: Maybe, though unlike the other cases I'm not sure if `show_trame` is intended to be re-exported as part of the API.  The other cases were importing only to check if a module was installed.

If you don't think `show_trame` is used here as part of the API then I can probably remove it and use importlib

tkoyama010: OK. Let's keep as is.

tkoyama010 (approved):
LGTM

PR #6888, opened 2024-11-27
Add Reader `Nek5000Reader` to read Nek5000 files

tkoyama010:
### Overview

resolve #6885

### Details

- https://vtk.org/doc/nightly/html/classvtkNek5000Reader.html

PR #6990, opened 2024-12-16
Support reading nek5000 files

### LINKED ISSUE #6885: Support for Nek5000 (.nek5000) files

It would be very useful if pyvista could have the ability to read nek5000 files. The particular applications in mind are for post-processing Nek5000/ NekRS data.

The documentation for the [vtkNek5000Reader](https://vtk.org/doc/nightly/html/classvtkNek5000Reader.html). I have also attached a basic implementation below. The resulting code should behave similarly to other time readers.

The Nek5000 Reader inherits from `BaseReader`, `TimeReader` and `PointCellDataSelection` classes. Nek5000 only produces point data. As a result, the `vtkNek5000Reader` class does not have methods for enabling cell arrays as far as I can tell. It therefore could be useful to disable the cell array routines inherited from `PointCellDataSelection` in the implementation. I should add the caveat that I am no expert in VTK, so there are almost certainly better ways to implement this.

```python
import vtk

from pyvista.core.utilities import (BaseReader, 
                                                    PointCellDataSelection,
                                                    TimeReader)

class Nek5000Reader(BaseReader, PointCellDataSelection, TimeReader):
    """
    Class for reading .nek5000 files produced by Nek5000 and NekRS
    """
    _vtk_module_name = "vtkIOParallel"
    _vtk_class_name = "vtkNek5000Reader"

    @property
    def number_time_points(self):  # noqa: D102
        """Return number of time points or iterations available to read.

        Returns
        -------
        int

        """ 
        return self.reader.GetNumberOfTimeSteps()
    
    @property
    def time_values(self):
        """All time or iteration values.

        Returns
        -------
        list[float]

        """
        key = vtk.vtkStreamingDemandDrivenPipeline.TIME_STEPS()
        vtkinfo = self.reader.GetExecutive().GetOutputInformation(0)
        return [vtkinfo.Get(key, i) for i in range(self.number_time_points)]

    def time_point_value(self, time_point):  # noqa: D102i
        """Value of time point or iteration by index.

        Parameters
        ----------
        time_point : int
            Time point index.

        Returns
        -------
        float

        """
        return self.time_values[time_point]

    @property
    def active_time_value(self):  # noqa: D102
        """Active time or iteration value.

        Returns
        -------
        float

        """
        return self.time_values[self._active_time_point]

    def set_active_time_value(self, time_value):  # noqa: D102
        """Set active time or iteration value.

        Parameters
        ----------
        time_value : float
            Time or iteration value to set as active.

        """
        try:
            index = self.time_values.index(time_value)
        except ValueError:
            raise ValueError(f"Time {time_value} not present") from None

        self.set_active_time_point(index)
    def set_active_time_point(self, time_point):  # noqa: D102
        """Set active time or iteration by index.

        Parameters
        ----------
        time_point : int
            Time or iteration point index for setting active time.

        """
        self._active_time_point = time_point
        self.reader.SetTimeStepRange(time_point,time_point)
        self.reader.Update()
```

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

mattfalcone1997:
### Overview
Implements Nek5000Reader class based on the PointCellDataSelection and TimeReader classes with the cell array routines disabled on the former as nek5000 data does not produce cell data.

This pull request resolves #6885.


mattfalcone1997:
I was wondering whether this should be re-forked, as it was originally forked some time ago and there are many commits that have already been merged here, which may make the review confusing.

user27182:
> I was wondering whether this should be re-forked, as it was originally forked some time ago and there are many commits that have already been merged here, which may make the review confusing.

Since this branch is based from #6888, updating that branch and this one seems to sync everything up nicely and the diff now looks correct/ easy to review.

Though, it may be easier to rebase this PR to main instead of indirectly working to merge with that other branch

user27182:
I removed the review requests since those were automatically triggered because of changes to `trame` files introduced from other PRs. Updating the branch removed those erroneous changes though, so the auto-review request no longer applies to the files modified by this PR

pyvista/core/utilities/reader.py
```diff
@@ -2085,6 +2087,168 @@ def set_active_time_point(self, time_point) -> None:  # noqa: D102
...
+        """
+        if time_point < 0 or time_point >= self.number_time_points:
+            raise ValueError(f'Time point out of range [0, {self.number_time_points-1}]')
+
+        self.set_active_time_value(self.time_values[time_point])
+
+    @property
+    def number_cell_arrays(self):  # noqa: D102
+        raise AttributeError(
+            'Nek 5000 data does not contain cell arrays, ' 'this method cannot be used'
```
MatthewFlamm, line 2217: This looks like a ruff reformatting issue?  Can it be a single string?  Applies to all below . It would be cleaner to have a private variable defined once for these strings, but not critical.

MatthewFlamm (approved):
LGTM overall, nitpicky comment only.

MatthewFlamm:
The author in the GitHub history on the main branch will ultimately be the author of #6888 in the current approach.  This won't reflect the true contribution here.  The changes in #6888 look to be a placeholder only.

If #6888 updates from main and then this PR updates from #6888, it might be possible to switch back to basing this PR on main.  But, I'm never sure in these situations with GitHub.  I usually take the route of starting a new branch based off of main.

mattfalcone1997:
> The author in the GitHub history on the main branch will ultimately be the author of #6888 in the current approach. This won't reflect the true contribution here. The changes in #6888 look to be a placeholder only.
> 
> If #6888 updates from main and then this PR updates from #6888, it might be possible to switch back to basing this PR on main. But, I'm never sure in these situations with GitHub. I usually take the route of starting a new branch based off of main.

It would be nice if the contribution is reflected, but if it isn't the end of the world.

mattfalcone1997:
Some of the unit tests on Linux appear to fail with an odd issue where it says that vtkNek5000Reader is not an attribute of vtkmodules.vtkIOParallel.

user27182:
> > The author in the GitHub history on the main branch will ultimately be the author of #6888 in the current approach. This won't reflect the true contribution here. The changes in #6888 look to be a placeholder only.
> > If #6888 updates from main and then this PR updates from #6888, it might be possible to switch back to basing this PR on main. But, I'm never sure in these situations with GitHub. I usually take the route of starting a new branch based off of main.
> 
> It would be nice if the contribution is reflected, but if it isn't the end of the world.

You can try rebasing this PR to main as suggested. Hit "Edit" next to the title and you can change the base to `main` from there. This _may_ work, though you can also create a new branch and separate PR if not.

> Some of the unit tests on Linux appear to fail with an odd issue where it says that vtkNek5000Reader is not an attribute of vtkmodules.vtkIOParallel.

This seems like the reader is not supported for older VTK, or if it is supported the module name may have changed across versions. You can maybe follow convention from other readers/vtk imports and do something like a `try ... except ImportError`

pyvista/examples/downloads.py
```diff
@@ -8185,3 +8185,38 @@ def _full_head_files_func():  # pragma: no cover
...
+    -------
+    pyvista.Nek5000Reader | str
+        DataSet or filename depending on ``load``.
+
+    Examples
+    --------
+    >>> from pyvista import examples
+    >>> dataset = examples.download_nek5000()
+    >>> dataset.plot()
+
```
user27182, line 8211: This will add a cross-ref to the dataset's entry in the gallery

```suggestion

    .. seealso::
    
        :ref:`Nek5000 Dataset <nek5000_dataset>`
            See this dataset in the Dataset Gallery for more info.

```

pyvista/examples/downloads.py
```diff
@@ -8185,3 +8185,38 @@ def _full_head_files_func():  # pragma: no cover
...
+    Parameters
+    ----------
+    load : bool, optional
+        Load the dataset after downloading it when ``True``.  Set this
+        to ``False`` and only the filename will be returned.
+
+    Returns
+    -------
+    pyvista.Nek5000Reader | str
```
user27182, line 8203: This should return a dataset type loaded by the reader, not the reader itself e.g. `pyvista.PolyData`

pyvista/core/utilities/reader.py
```diff
@@ -176,7 +178,7 @@ def get_reader(filename, force_ext=None):
     | ``.xdmf``      | :class:`pyvista.XdmfReader`                 |
     +----------------+---------------------------------------------+
     | ``.vtpd``      | :class:`pyvista.XMLPartitionedDataSetReader`|
-    +----------------+---------------------------------------------+
+    +--------------------------------------------------------------+
```
tkoyama010, line 181:
```suggestion
    +----------------+---------------------------------------------+
```

tkoyama010 (changes requested):

mattfalcone1997:
> > > The author in the GitHub history on the main branch will ultimately be the author of #6888 in the current approach. This won't reflect the true contribution here. The changes in #6888 look to be a placeholder only.
> > > If #6888 updates from main and then this PR updates from #6888, it might be possible to switch back to basing this PR on main. But, I'm never sure in these situations with GitHub. I usually take the route of starting a new branch based off of main.
> > 
> > 
> > It would be nice if the contribution is reflected, but if it isn't the end of the world.
> 
> You can try rebasing this PR to main as suggested. Hit "Edit" next to the title and you can change the base to `main` from there. This _may_ work, though you can also create a new branch and separate PR if not.
> 
> > Some of the unit tests on Linux appear to fail with an odd issue where it says that vtkNek5000Reader is not an attribute of vtkmodules.vtkIOParallel.
> 
> This seems like the reader is not supported for older VTK, or if it is supported the module name may have changed across versions. You can maybe follow convention from other readers/vtk imports and do something like a `try ... except ImportError`
> 
> EDIT: Looks like it was introduced Jan 2023 [Kitware/VTK@7cc3156](https://github.com/Kitware/VTK/commit/7cc31568dcff59de412f3e6f269905be0ebbca9f), so probably this is available for vtk >= 9.3 only. `ImportError` handling should be added to the reader with a message indicating the reader requires vtk>=9.3.

Changing the base seemed to work. I have also modified to raise a VTKVersionError if version of vtk insufficient

pyvista/core/utilities/reader.py
mattfalcone1997: I have made them private

pyvista/core/utilities/reader.py
mattfalcone1997: I believe this is done

user27182 (approved):
LGTM, thanks!

tkoyama010 (approved):
LGTM

user27182:
New dataset info 👍
nek5000

PR #6467, opened 2024-08-07
Drop blackdoc and use blacken-docs instead

tkoyama010:
### Overview

- ~https://docs.astral.sh/ruff/formatter/#docstring-formatting~
- https://learn.scientific-python.org/development/guides/style/#format

### Details

- https://github.com/scientific-python/cookie/issues/343

PR #6991, opened 2024-12-16
Format docstring with ruff

beroda:
### Overview

Guess what ... ruff can ~make coffee :coffee:~ format docstrings https://docs.astral.sh/ruff/formatter/#docstring-formatting

Let's try out ...

Close #6467


pyvista/_plot.py
```diff
@@ -200,9 +200,7 @@ def plot(
     ImageData. Note ``volume=True`` is passed.
 
     >>> import numpy as np
-    >>> grid = pv.ImageData(
-    ...     dimensions=(32, 32, 32), spacing=(0.5, 0.5, 0.5)
-    ... )
+    >>> grid = pv.ImageData(dimensions=(32, 32, 32), spacing=(0.5, 0.5, 0.5))
```
beroda, line 203: Lines are wrapped more consistently than with blackdoc (length is 69 < limit of 75)

user27182:
+1 for using `ruff` generally.
See also https://github.com/pyvista/pyvista/pull/6467#issuecomment-2272536196
There were previously concerns with `ruff` not formatting `rst` and `md` files

user27182:
The ruff [configuration](https://docs.astral.sh/ruff/configuration/) states:

> Enable auto-formatting of code examples in docstrings. Markdown,
reStructuredText code/literal blocks and doctests are all supported.

Maybe I'm missing something but if markdown is supported, shouldn't it format this code block to use single quotes?
https://github.com/pyvista/pyvista/blob/6e59ba2115416e82c5994fce5abb6c5be271bbfc/CONTRIBUTING.rst?plain=1#L430-L437

But I'm not seeing `CONTRIBUTING.md` in the diff for this PR... There are also no `rst` files listed either, not sure if there are any codeblocks anywhere that _should_ be picked up, but still... makes me think `rst` files aren't covered either. 

Also, the current `blackdoc` config seems not to be working? Because if I try running `blackdoc CONTRIBUTING.rst`, this _does_ seem to work, but it errors out (it tries formatting bash code blocks and can't). But this clearly isn't the case with our current blackdoc config.


So, I'm +1 with this PR and dropping blackdoc and using ruff. But we should still figure out how to fill the gap and format code in `rst` and `md` files

beroda:
> The ruff [configuration](https://docs.astral.sh/ruff/configuration/) states:
> 
> > Enable auto-formatting of code examples in docstrings. Markdown,
> > reStructuredText code/literal blocks and doctests are all supported.
> 
> Maybe I'm missing something but if markdown is supported, shouldn't it format this code block to use single quotes?

I agree that the ruff documentation is a bit misleading on this topic. I think they mean that ruff can format `Markdown` and `rst` python code blocks that are **inside the docstring of functions**. The examples in the configuration speak for themselves.
https://docs.astral.sh/ruff/settings/#format_docstring-code-format

> So, I'm +1 with this PR and dropping blackdoc and using ruff. But we should still figure out how to fill the gap and format code in `rst` and `md` files

I guess this gap will be filled once https://github.com/astral-sh/ruff/issues/8237 and https://github.com/astral-sh/ruff/issues/3792 are solved.

beroda:
This PR should be merged after https://github.com/pyvista/pyvista/pull/6994 to be compatible.
Move to draft.

pyproject.toml
```diff
@@ -175,6 +175,7 @@ pyvista = ['py.typed']
 # reading the docstrings in text terminals.
 line-length = 75
 
+
```
tkoyama010, line 178:
```suggestion
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

PR #6993, opened 2024-12-17
Add sphinx-duration extension to measure sphinx processing times

user27182:
### Overview

See https://www.sphinx-doc.org/en/master/usage/extensions/duration.html#module-sphinx.ext.duration

Not sure if it shows total build times or per-file processing times or even  just the slowest files, but could help highlight some build-time bottlenecks


user27182:
Output from this option. Apparently over 30 minutes is spent processing cube maps?!

image

tkoyama010:
Awesome analysis. What about switching to a static display for those that run too slowly to save time?

user27182:
> Awesome analysis. What about switching to a static display for those that run too slowly to save time?

Tried static display with #6526. Results from running that PR with `sphinx.ext.duration`:
times

Times haven't changed.

Maybe we can use `sphinx.ext.autodoc.between` to edit all docstrings and remove any/all the examples? Then we could have a "fast-build" version of the current build process which doesn't run any code. https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-process-signature

Either way I think extension is valuable and will help with monitoring build times.

MatthewFlamm (approved):
Looks extremely useful to help reduce build times on CI for PRs.

PR #6995, opened 2024-12-17
Fix type annotation for `pyvista.core.filters.data_set.contour`

finsberg:
### Overview

According to the docstring of [`pyvista.core.filters.data_set.contour`](https://github.com/pyvista/pyvista/blob/main/pyvista/core/filters/data_set.py#L2045-L2073), the argument  `isosurfaces` can take an int or a sequence of floats. However, the type annotation only sais `int`, so adding `Sequence[float]`.
Similarly, the argument `scalars` can take an optional `str` or an `array_like` float, but type annotation only say optional `str`, so adding `VectorLike[float]` here as well.

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -2044,8 +2044,8 @@ class DataSetFilters:
 
     def contour(  # type: ignore[misc]
         self: ConcreteDataSetType,
-        isosurfaces: int = 10,
-        scalars: str | None = None,
+        isosurfaces: int | Sequence[float] = 10,
+        scalars: str | VectorLike[float] | None = None,
         compute_normals: bool = False,
         compute_gradients: bool = False,
         compute_scalars: bool = True,


beroda:
Good catch. mypy `unreachable` warning also found the issue (see [here](https://github.com/pyvista/pyvista/pull/6982/files#r1887641411))

finsberg:
> Good catch. mypy `unreachable` warning also found the issue (see [here](https://github.com/pyvista/pyvista/pull/6982/files#r1887641411) and [here](https://github.com/pyvista/pyvista/pull/6982/files#r1887639223))

Nice. I also had to add an additional check for scalars not being a Sequence. The reason being that the following example 
```python
from collections.abc import Sequence

x: str | Sequence[float] = 'hello'
reveal_type(x)
if isinstance(x, str):
    reveal_type(x)
elif isinstance(x, Sequence):
    reveal_type(x)
```
produces the following output
```
$  python3 -m mypy mwe.py
mwe.py:4: note: Revealed type is "Union[builtins.str, typing.Sequence[builtins.float]]"
mwe.py:6: note: Revealed type is "builtins.str"
mwe.py:8: note: Revealed type is "Union[builtins.str, typing.Sequence[builtins.float]]"
Success: no issues found in 1 source file
```
when run through mypy.

pyvista/core/filters/data_set.py
```diff
@@ -2044,8 +2044,8 @@ def elevation(  # type: ignore[misc]
 
     def contour(  # type: ignore[misc]
         self: ConcreteDataSetType,
-        isosurfaces: int = 10,
-        scalars: str | None = None,
+        isosurfaces: int | Sequence[float] = 10,
+        scalars: str | NumpyArray[float] | None = None,
```
user27182, line 2048: This looks correct in terms of minimal typing changes. Though, we are making an effort to standardize inputs and accept more general inputs like `VectorLike[float]`. I see this was tried initially but it seems would require modifying the validation logic a bit to accommodate, so we can leave that for a separate PR.

pyvista/core/filters/data_set.py
```diff
@@ -2169,7 +2169,7 @@ def contour(  # type: ignore[misc]
 
         if isinstance(scalars, str):
             scalars_name = scalars
-        elif isinstance(scalars, (Sequence, np.ndarray)):
+        elif isinstance(scalars, (Sequence, np.ndarray)) and not isinstance(scalars, str):
```
user27182, line 2172: In general this is a good approach but in this case it seems redundant and would probably be considered as such if we had `redundant-expr` enabled since `isinstance(scalars, str)` will always evaluate false here (it's already checked above)
```suggestion
        elif isinstance(scalars, (Sequence, np.ndarray))
```

We can leave it though and fix this if/when mypy complains later after adding this option (either via #6225 or #6970

user27182 (approved):
LGTM, thanks! I have some minor nits (for future PRs)

user27182:
Side note, there is a nice protocol definition that can be used precisely for cases where non-string sequences are desired:
https://github.com/hauntsaninja/useful_types/blob/305c66f860117ea41e0dd1f951f2fc3b4786c57e/useful_types/__init__.py#L285-L295

PR #6999, opened 2024-12-18
Use `mathjax` extension to render math equations in Sphinx documents

tkoyama010:
### Overview

TBD

### Details

- None


banesullivan:
I don't think we have any mathjax equations at present so I'll let any new contributions adding those add this extension at that time

PR #7001, opened 2024-12-18
Fix empty input behavior for `extract_values` and remove redundant scalar validation

user27182:
### Overview

`extract_values` returns `self.copy()` for empty inputs. But the return type should be `MultiBlock` or `UnstructuredGrid`. This PR fixes this.

The scalar array validation is also redundant and is simplified - it raises the same errors which are otherwise raised automatically by `set_default_active_scalars` and `get_array`.


tkoyama010 (approved):
LGTM. Just a minor suggestion: The variable `split` is accessed globally from outside the function, which made the code slightly harder to follow. It might be a good idea to modularize the internal functions for better readability. It is another PR anyway.

user27182:
> LGTM. Just a minor suggestion: The variable `split` is accessed globally from outside the function, which made the code slightly harder to follow. It might be a good idea to modularize the internal functions for better readability. It is another PR anyway.

I'm not sure I follow,  `extract_values` has a keyword argument `split`, do you mean this is defined elsewhere? Should currently be defined at the root function level:

``` python
def extract_values(split):
    ...
    if split: 
        ...

tkoyama010:
> > LGTM. Just a minor suggestion: The variable `split` is accessed globally from outside the function, which made the code slightly harder to follow. It might be a good idea to modularize the internal functions for better readability. It is another PR anyway.
> 
> I'm not sure I follow, `extract_values` has a keyword argument `split`, do you mean this is defined elsewhere? Should currently be defined at the root function level:
> 
> ```python
> def extract_values(split):
>     ...
>     if split: 
>         ...
> ```

I meant my comment in the sense that I was a bit confused because `_validate_values_and_ranges` does not have `split` as an argument. But this is a minor issue since, as you say, it is immediately obvious that it is in the `extract_values`.

user27182:
Oh I understand now, the variable scope is not local. Yes that can be confusing (and buggy).

PR #7003, opened 2024-12-19
Do not run docstring examples with long processing times

user27182:
### Overview

Some examples take a long time to process (see https://github.com/pyvista/pyvista/pull/6993#issuecomment-2548444061). Not yet sure what to do about it but I want to see if removing these completely speeds things up.


user27182:
Build time is now reduced to 45 mins, all doctest code from all docstrings in `pyvista` package has been disabled

Processing times summary:
times_

user27182:
The entire workflow now takes about 40-45 minutes run (compared to ~2hrs before)

- ~7 minutes to install deps, load cache, and read all sources
- ~30 minutes to write output
- ~5 minutes to run tests,

PR #7005, opened 2024-12-19
Move classproperty decorator to misc utilities

user27182:
### Overview

See https://github.com/pyvista/pyvista/pull/5119#discussion_r1892375446


pyvista/core/utilities/misc.py
```diff
@@ -317,3 +319,27 @@ def _reciprocal(x: ArrayLike[float], tol: float = 1e-8) -> NumpyArray[float]:
     x[~zero] = np.reciprocal(x[~zero])
     x[zero] = 0
     return x
+
+
+class classproperty(property):
```
MatthewFlamm, line 324: I think we should make this private.  We don't want downstream packages using these things.  If we dont like the `@_classproperty` usage, it would be better if we had a `_misc` module.

MatthewFlamm (dismissed):
Thanks!

tests/core/test_utilities.py
```diff
@@ -1850,3 +1851,13 @@ def test_parse_interaction_event_raises_wrong_type():
...
+
+
+def test_classproperty():
+    class Foo:
+        @_classproperty
+        def prop(cls):
+            return 42
+
+    assert Foo.prop == 42
+    assert Foo().prop == 42
```
MatthewFlamm, line 1863: Should we also test that `Foo.prop()` and the instance variation both fail?

user27182: This implementation doesn't off any guards against setting the property. You can do `Foo.prop = 99` without error... not a perfect workaround

MatthewFlamm: Hmmmmm.... This is a decently big downside.  Without the callable nature and restricting setting, someone may be tempted to do `Theme.defaults = {`my_default`}` to register a default theme, in regards to #5119.

There is a suggestion that a metaclass is required https://stackoverflow.com/a/76047344

I'm not that familiar with metaclasses, but I'm pretty sure that this would be a one at a time type of usage rather than an easy to use decorator like here.  Maybe best just to keep this PR as is, and I will add in a sentence about registering themes into the `default` docstring.

MatthewFlamm (dismissed):
LGTM.  We should wait to merge this one until someone else can also review in case they have a stronger opinion about the downsides.

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

PR #7006, opened 2024-12-19
Add zizmor security linter to pre-commit

tkoyama010:
### Overview

Recently, there was an attack on the Ultralytics package. For more information, please see [Supply-chain attack analysis: Ultralytics](https://blog.pypi.org/posts/2024-12-11-ultralytics-attack-analysis/). In this article, it is suggested that checking actions with [zizmor](https://github.com/woodruffw/zizmor). Let's consider using zizmor checks in pre-commit.

### Details

- None


.github/workflows/label.yml
```diff
@@ -1,6 +1,6 @@
 name: "Pull Request Labeler"
 on:
-  pull_request_target:
+  pull_request:
```
tkoyama010, line 3: We will need to manually add labels to pull requests from external repositories. However, this should be acceptable for safety reasons.

.pre-commit-config.yaml
```diff
@@ -94,3 +94,8 @@ repos:
       - id: taplo-format
         # See options: https://taplo.tamasfe.dev/configuration/formatter-options.html
         args: [--option, "reorder_arrays=true", --option, "reorder_keys=true"]
+
+  - repo: https://github.com/woodruffw/zizmor-pre-commit
+    rev: v0.8.0
```
user27182, line 99: Should we use latest with this PR or bump in a follow-up?
```suggestion
    rev: v1.0.0
```

.github/workflows/label.yml
user27182: How does this tool work for enforcing policies? E.g. does it error if there are critical flaws, and only warn if mid-risk flaws are found?

e.g. if this change is reverted, will pre-commit fail?

user27182 (commented):
+1 for security, that `ultralytics` case sounds like a nightmare which we want to avoid.

.github/workflows/label.yml
tkoyama010: If this change is reverted, the check will fail. Currently, errors occur at all stages of risk assessment. If needed, the configuration can be reviewed for potential adjustments.

user27182 (approved):

user27182:
I guess there are new errors after updating to `v1.0.0` but in principal this LGTM otherwise

banesullivan (approved):
Thanks for adding zizmor!

user27182:
For the new cache-poisoning errors, I _think_ [ignoring](https://woodruffw.github.io/zizmor/usage/#ignoring-results) the errors for any testing workflows is _mostly_ ok. The main [remediation](https://woodruffw.github.io/zizmor/audits/#remediation_13) seems to be to remove any caching from workflows that publish build artifacts, which I would think only applies to actions that check ` if startsWith(github.ref, 'refs/tags/v')`. So maybe we just need to double check the caching logic to make sure that nothing is cached for releases, and ignore everything else.

.pre-commit-config.yaml
```diff
@@ -94,3 +94,8 @@ repos:
       - id: taplo-format
         # See options: https://taplo.tamasfe.dev/configuration/formatter-options.html
         args: [--option, "reorder_arrays=true", --option, "reorder_keys=true"]
+
+  - repo: https://github.com/woodruffw/zizmor-pre-commit
+    rev: v1.0.0
```
user27182, line 99:
```suggestion
    rev: v0.8
```

In the interest of moving this PR forward (and improving this repo's security) I think it's fine to revert to an older version. The cache-poisoning checks from the newer version can be addressed in a separate PR.

user27182 (approved):

banesullivan (approved):

PR #7007, opened 2024-12-19
Add redundant expr mypy

beroda:
### Overview

Part of https://github.com/pyvista/pyvista/pull/6970.
Add `redundant-expr` to mypy ([MY105](https://learn.scientific-python.org/development/guides/style/#MY105))


pyvista/core/utilities/fileio.py
```diff
@@ -242,7 +242,7 @@ def read(
     if isinstance(filename, Sequence) and not isinstance(filename, str):
         multi = pyvista.MultiBlock()
         for each in filename:
-            name = Path(each).name if isinstance(each, (str, Path)) else None
+            name = Path(each).name if isinstance(each, (str, Path)) else None  # type: ignore[redundant-expr]
```
beroda, line 245: Added because `filename` can be nested. Not in the docstring but tested.
https://github.com/pyvista/pyvista/blob/f15eff8e27c35402fa16699f7b441ab3e9178cbc/tests/core/test_utilities.py#L134-L136
I don't know how to annotate nested, I am open to suggestions/help

pyvista/examples/_dataset_loader.py
```diff
@@ -29,6 +29,7 @@
 """
 
 # ruff: noqa: PTH102,PTH103,PTH107,PTH112,PTH113,PTH117,PTH118,PTH119,PTH122,PTH123,PTH202
+# mypy: disable-error-code="redundant-expr"
```
beroda, line 32: Disabled because of a very high number of `redundant-expr` and I am not very familiar to this new module. Might need some help if someone's interested, you are welcome :pray:

pyvista/core/utilities/fileio.py
user27182: can it be arbitrarily deeply nested (just like a MultiBlock nested with MultiBlock)? e.g. 
``` python
Path | str | Sequence[Path | str | Sequence[ Path | str | Sequence[...]]]
```

If so it's possible to type recursively, I think the `_SerializedDictArray` in `arrays.py` uses a recursive type, can look there for an example

beroda: Done in https://github.com/pyvista/pyvista/pull/7007/commits/ad65307fd284c66d3f757c5a092036cfc8c9a4b9

tkoyama010 (approved):
LGTM

PR #7010, opened 2024-12-21
Add custom sphinx duration extension

user27182:
### Overview

Follow-up to #6993. That extension is useful, but it is hard-coded to only show 5 entries. For a large project with hundreds of files this is way too few.

This PR duplicates [`duration.py`](https://github.com/sphinx-doc/sphinx/blob/df3d94ffdad09cc2592caccd179004e31aa63227/sphinx/ext/duration.py) from sphinx and modifies it to show 100 entries instead.


MatthewFlamm:
This is useful but a decent amount of maintenance burden to merge in.  IMO it would be better to make a change in `sphinx` directly with a configuration option.

user27182:
Ok. It was mainly used in #7003 but has already served its purpose there and can also easily be added temporarily to any PR for investigative purposes in the future.

MatthewFlamm:
> Ok. It was mainly used in #7003 but has already served its purpose there and can also easily be added temporarily to any PR for investigative purposes in the future.

It makes sense to me if you wanted to keep this as a draft PR that is periodically updated to check progress.

user27182:
Checking periodically is nice but is unfortunately a reactive measure post-merge of other PRs. I think it is more useful if we could use this extension to instead _prevent_ the addition of new code which has a substantial sphinx reading time to prevent build times from ballooning. E.g. it would have been nice to catch the 6minute read time required for the cube maps before whatever change caused that. 

A follow-up PR could `grep` the output and report a diff from baseline values and/or emit an error for long read times. These kinds of changes may require further customizing the extension though.

PR #6894, opened 2024-11-29
Add color output to `pytest` and `mypy` CI

user27182:
### Overview

Enable colored console output.

- pytest, see https://github.com/pytest-dev/pytest/issues/7443
- mypy, see https://github.com/pytorch/pytorch/pull/61391

PR #7018, opened 2024-12-22
Enable color output for sphinx build in CI

user27182:
### Overview

See sphinx docs for this option here https://www.sphinx-doc.org/en/master/man/sphinx-build.html#environment-variables.
This PR is similar to #6976 and #6894.


user27182:
It looks like `FORCE_COLOR` is somewhat standard and `pip` is also using this env var:
pip color

Maybe we could add `FORCE_COLOR` to all workflows, not just docs?

user27182:
Color is working for `sphinx` 👍
sphinx color

banesullivan (approved):

PR #7021, opened 2024-12-27
Add smoothing parameter to `contour_labels` filter

user27182:
### Overview

Split from #6732.


pyvista/core/filters/image_data.py
```diff
@@ -1226,6 +1258,12 @@ def contour_labels(  # type: ignore[misc]
         using ``select_inputs`` converts previously-internal boundaries into external
         ones.
 
+        Disable smoothing to generate staircase-like surface. Without smoothing, the
```
tkoyama010, line 1261: It would be easier to understand if there is also an example of when the smoothing parameters are changed.

tkoyama010 (changes requested):

pyvista/core/filters/image_data.py
user27182: An example with `smoothing_scale` is added in [17aa6fa](https://github.com/pyvista/pyvista/pull/7021/commits/17aa6fa9a1c4f3d21ebac2071ad7ade3f468b856).

user27182: For the example with `smoothing_scale=0.5`, this is what the final docs will look like (with `closed_surface=True`)



Compared to the default with `smoothing_scale=1.0`




These images require `closed_surface=True` from #7028 which is not yet merged as an option, so the examples as they currently are in this PR don't yet appear like this.

tests/plotting/test_plotting.py
```diff
@@ -4548,6 +4549,159 @@ def _create_object(_direction=None):
...
+
+
+@pytest.mark.usefixtures('_allow_empty_mesh', '_show_edges')
+@pytest.mark.parametrize(
+    ('select_inputs', 'select_outputs'),
+    [(None, None), (None, 2), (2, 2)],
+    ids=['in_None-out_None', 'in_None-out_2', 'in_2-out_2'],
+)
+@pytest.mark.needs_vtk_version(9, 3, 0)
+def test_contour_labels_boundary_style(
```
user27182, line 4579: This test is made possible by this PR (it uses `smoothing=False`), but is otherwise not really testing the smoothing parameter. This test provides additional coverage for boundary style.

We could split this as a separate PR if desired. There are other few misc test coverage changes from #6732 that could probably be lumped in as well.

tkoyama010: I'm fine as is, but if you want to do it, I'm happy to review new PR.

user27182: Ok let's keep it as-is

tkoyama010 (dismissed):
LGTM

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

tests/plotting/test_plotting.py
```diff
@@ -4557,6 +4557,160 @@ def test_contour_labels_compute_normals(labeled_image, compute_normals):  # noqa
...
+        )
+        # Shrink mesh to help reveal cells hidden behind other cells
+        return mesh.shrink(0.7)
+
+    # Remove one foreground point from the fixture to simplify plots
+    labeled_image.active_scalars[19] = 0
+
+    fixed_kwargs = dict(
+        smoothing_distance=0.3,
+        output_mesh_type='quads',
```
user27182, line 4618:
```suggestion
        output_mesh_type='quads',
        compute_normals=False,
```

tests/plotting/test_plotting.py
```diff
@@ -4557,6 +4557,160 @@ def test_contour_labels_compute_normals(labeled_image, compute_normals):  # noqa
...
+@pytest.mark.parametrize('smoothing', [True, False])
+@pytest.mark.needs_vtk_version(9, 3, 0)
+def test_contour_labels_compare_select_inputs_select_outputs(
+    labeled_image,  # noqa: F811
+    smoothing,
+):
+    common_kwargs = dict(
+        smoothing=smoothing,
+        smoothing_distance=0.8,
+        output_mesh_type='quads',
```
user27182, line 4702:
```suggestion
        output_mesh_type='quads',
        compute_normals=False,
```

user27182 (commented):
These changes _might_ be needed if tests fails caused by #7034 being merged.

tkoyama010 (approved):

PR #7022, opened 2024-12-27
Add background value parameter to `contour_labels` filter

user27182:
### Overview

Split from https://github.com/pyvista/pyvista/pull/6732.

### DIFF
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -989,6 +989,7 @@ class ImageDataFilters(DataSetFilters):
         self: ImageData,
         boundary_style: Literal['external', 'internal', 'all'] = 'external',
         *,
+        background_value: int = 0,
         select_inputs: int | VectorLike[int] | None = None,
         select_outputs: int | VectorLike[int] | None = None,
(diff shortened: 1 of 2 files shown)


pyvista/core/filters/image_data.py
```diff
@@ -1038,6 +1039,10 @@ def contour_labels(  # type: ignore[misc]
             generated between foreground background. ``'all'``  includes both internal
             and external boundary polygons.
 
+        background_value : int, default: 0
+            Background value of the input image. All other values are considered
+            as foreground.
+
```
tkoyama010, line 1045:
```suggestion
            .. versionadded:: 0.45.0

```

tkoyama010 (changes requested):

pyvista/core/filters/image_data.py
user27182: This filter is not yet released, so all parameters are new to 0.45. Versionadded is used in the main description, is that sufficient or it's better to also include it with the parameters?

tkoyama010: Oops didn't realised that. Thanks!

tkoyama010 (approved):

PR #7023, opened 2024-12-27
Improve initial import time

darikg:
Currently it takes 2.14 seconds to import pyvista on my machine. This PR reduces it to 0.85s.  I used `python -X importtime -c "import pyvista" 2> import_times.log` to record the times and [tuna](https://github.com/nschloe/tuna) (`tuna import_times.log`) to visualize them.

The two main culprits were:

1. `core.filters.data_set` imported `matplotlib.pyplot` at module-level. Since matplotlib is only used for three filters there, 
(`plot_over_line`, `plot_over_circular_arc`, `plot_over_circular_arc_normal`), I just moved them to function-level imports. This is responsible for ~0.7 s 

2. `core._typing_core._aliases` imported `scipy.spatial.Rotation` to use in the `RotationLike`/`TransformLike` type alias. I simply put that import behind a `if TYPE_CHECKING` guard. This is responsible for another ~0.7 s.

There's a couple of smaller things that could be trimmed off with more sophisticated lazy imports but I'm happy to get this low-hanging fruit.


user27182:
Is there a way to test this to prevent regressions? Two recent PRs mess with color imports in `core` (#7024, #7025), not 100% sure if these imports are all lazy

darikg:
> Is there a way to test this to prevent regressions? Two recent PRs mess with color imports in `core` (#7024, #7025), not 100% sure if these imports are all lazy

I can look into it. Some sort importlib hook to make sure matplotlib and scipy don't get imported by `import pyvista`? Or were you thinking of a more general performance test to prevent any regression in import time?

user27182:
I meant the timing, like if scipy or matplotlib.plt are imported in core during init at runtime in the future (imports which are avoided by this PR), how can we detect this and avoid doing it in the future to keep import times low?

darikg:
I don't think we have any performance profiling in place, do we? It's hard because it's non-deterministic, varying heavily on what nodes the tests run on. I think you really need dedicated machines for that kind of stuff.

I kinda like just having a hard and fast rule that prevents big libs like matplotlib and scipy from being imported at root, more culprits can be added to the exclusion list as they're discovered.

darikg:
Added tests to prohibit matplotlib and scipy imports

pyvista/core/_typing_core/_aliases.py
```diff
@@ -14,8 +14,13 @@
 from ._array_like import _ArrayLike1D
 from ._array_like import _ArrayLike2D
 
-with contextlib.suppress(ImportError):
-    from scipy.spatial.transform import Rotation
+if TYPE_CHECKING:
```
user27182, line 17:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

user27182 (approved):
LGTM

user27182:
I just thought of something to check, the docs for the type aliases rely on runtime definitions, I wonder if the ref to `Rotation` is lost in the docs with these changes?
i.e. will `Rotation` be replaced with `None`?
https://dev.pyvista.org/api/core/typing#pyvista.RotationLike
rotationlike

I am unable to download the docs artifact to check, maybe someone with better internet can check?

darikg:
@user27182 good catch! let's see if this fixes it

darikg:
image
it correctly links to [Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html#scipy.spatial.transform.Rotation) now

user27182 (approved):

Issue #5994, opened 2024-04-29
Incorrect color mapping using `categories=True` for non-linear categories


user27182:
### Describe the bug, what's wrong, and what you expected.

When plotting categorical data with a non-linear or non-sequential integer mapping, the categories are not colored correctly.

Edit: For context, this bug means that the use of `categories=True` should probably be avoided until this is fixed, unless it's certain that the categories in the data being plotted are sequential, contiguous, 0-indexed integers. E.g any examples in the docs which currently use the `download_frog_tissue` or `download_whole_body_ct` and plot `categorical=True` are likely showing the segmentation labels incorrectly.

### Steps to reproduce the bug.

E.g. Plot mesh with three categories for cell data: [0, 2, 8]
``` python

import pyvista as pv
import numpy as np
sphere = pv.Sphere()
sphere['labels'] = np.zeros(sphere.n_cells)      # category: '0'
sphere['labels'][range(sphere.n_cells//2)] = 2   # category: '2'
sphere['labels'][range(sphere.n_cells//4)] = 8   # category: '8'
sphere.plot(cmap='glasbey', categories=True)
```
Will produce this, which shows 3 colors in the scalar bar for 3 categories, but visually only colors the mesh with two.
![image](https://github.com/pyvista/pyvista/assets/89109579/dc774d99-9b6f-464d-9970-5be644fb0d41)

Setting `categories=False`, we can see the "correct" coloring for the sphere (with non-categorical scalar bar). I would have expected this to as the coloring with `categories=True`, however.
![image](https://github.com/pyvista/pyvista/assets/89109579/0e086a06-3908-42bd-ac35-79b5fa9509b6)

I suspect the issue has something to do with this line:
https://github.com/pyvista/pyvista/blob/3accbf0ec6f7429a4b4a0e6500a709ab81b65573/pyvista/plotting/lookup_table.py#L762

which applies a linear mapping for the colors. In this case, the `apply_cmap` function has limited information about the data, as it only receives an `n_values` parameter. To resolve this issue, I think `apply_cmap` needs to know the actual unique values in the data so that it can create a non-linear mapping based on the actual data values, rather than just the number of values in the data.

For reference, the value of `n_values` which is passed to `apply_cmap`, is set here when `categories=True`:
https://github.com/pyvista/pyvista/blob/3accbf0ec6f7429a4b4a0e6500a709ab81b65573/pyvista/plotting/plotter.py#L4269-L4270

[...]

PR #7024, opened 2024-12-27
Add new `color_labels` filter to add color scalars to labeled data

user27182:
### Overview

This filter can be used for coloring non-contiguous labeled data. It is well-suited to segmentation applications where labelmaps are commonly used. In part, this filter overcomes the limitations of using `categorical=True` with the plotter, see #5994.

As a follow-up, this can be used to improve the examples for the `contour_labels` filter, see https://github.com/pyvista/pyvista/pull/6732#discussion_r1869899480


pyvista/core/filters/data_set.py
```diff
@@ -8942,6 +8945,296 @@ def pack_labels(  # type: ignore[misc]
...
+
+            items = zip(keys, values)
+
+        colors_out = np.full((len(array), num_components), default_channel_value, dtype=color_dtype)
+        for label, color in items:
+            if isinstance(color, dict):
+                color = color['color']  # type: ignore[unreachable]
+            colors_out[array == label, :] = color
+
+        colors_name = name + '_rgba' if output_scalars is None else output_scalars
```
user27182, line 9232: if `color_type` is `float_rgb` or `int_rgb`, should `_rgb` be appended to the array name in these cases instead of `_rgba` ?

user27182: fixed in [cd6a05f](https://github.com/pyvista/pyvista/pull/7024/commits/cd6a05f2fbc4fc2df26d9b4f419d7033fbc67fe5)

pyvista/core/filters/data_set.py
```diff
@@ -8942,6 +8945,302 @@ def pack_labels(  # type: ignore[misc]
...
+            RGB integers to reduce memory usage.
+
+            .. note::
+                The color type affects the default value for unspecified colors when
+                a dictionary is used. See ``colors`` for details.
+
+        scalars : str, optional
+            Name of scalars with labels. Defaults to currently active scalars.
+
+        preference : str, default: "point"
```
user27182, line 9039: Might make more sense for `'cell'` to be the default.
```suggestion
        preference : str, default: "cell"
```

pyvista/core/filters/data_set.py
```diff
@@ -8942,6 +8945,302 @@ def pack_labels(  # type: ignore[misc]
...
+                    '  - sequence of color-like values,\n'
+                    '  - dict with color-like values,\n'
+                    '  - named colormap string.\n'
+                    f'Got: {seq}'
+                )
+
+        def _is_index_like(array_, max_value):
+            if np.issubdtype(array_.dtype, np.integer) or np.array_equal(array, np.floor(array_)):
+                min_, max_ = output_mesh.get_data_range(name)
+                if min_ >= 0 and max_ <= max_value:
```
tkoyama010, line 9145:
```suggestion
                if 0 <= min and max_ <= max_value:
```

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
user27182: This is linted by ruff to have the constant on the right:
https://docs.astral.sh/ruff/rules/yoda-conditions/

We can `noqa` it though, it does make sense here to have consistent operators

pyvista/core/filters/data_set.py
user27182: cell is now the default [0ea8d97](https://github.com/pyvista/pyvista/pull/7024/commits/0ea8d973098d8b883962619edc753b2ef112f8ef)

pyvista/core/filters/data_set.py
```diff
@@ -8942,6 +8945,302 @@ def pack_labels(  # type: ignore[misc]
...
+            Dataset with RGB(A) scalars. Output type matches input type.
+
+        Examples
+        --------
+        Load labeled data and crop it to simplify the data.
+
+        >>> from pyvista import examples
+        >>> import numpy as np
+        >>> image_labels = examples.load_channels()
+        >>> image_labels = image_labels.extract_subset(voi=(75, 109, 75, 109, 85, 100))
```
tkoyama010, line 9064: Would it be possible to plot the dataset at this stage to provide a clearer understanding of the filter's behavior?

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
user27182: An example with the default coloring applied by the plotter is added in [d4f9e44](https://github.com/pyvista/pyvista/pull/7024/commits/d4f9e4496e4256ff2f514623fa46d914540c4110)

Still need to update the doc image cache but should be good to go.

tkoyama010 (approved):
LGTM

PR #7025, opened 2024-12-28
Move `_validate_color_sequence` to validation module for reuse

user27182:
### Overview

Move method since it's is also used by #7024


tkoyama010 (approved):
LGTM

PR #1860, opened 2021-11-26
Make sure inplace never defaults to True

banesullivan:
### Overview

The default value of inplace should never be `True`. This cleans up a few methods that slipped by us.

Addresses issues in #1855

### Details

See my explanation in https://github.com/pyvista/pyvista/issues/1855#issuecomment-979774232

The established convention for PyVista filters is: 

```
result = source.filter(options)
```

And `inplace` should always be an opt-in feature and never imposed on a user because of the fact that it would then unintentionally alter the source data.

@akaszynski, please review

PR #7027, opened 2024-12-28
Fix default  param value in docs for `transform` `inplace`

dad616610:
### Overview

The default value of the function parameter  `inplace` in `transform` function is `True`: https://github.com/pyvista/pyvista/blob/57f40202a8d4bb2315a79d088746dc1a04676cf1/pyvista/core/filters/data_set.py#L7022
 while the docs say it's `False`: https://github.com/pyvista/pyvista/blob/57f40202a8d4bb2315a79d088746dc1a04676cf1/pyvista/core/filters/data_set.py#L7066

This PR makes the docs mirror the code's behaviour:
```diff
-       inplace : bool, default: False 
+       inplace : bool, default: True
```

But maybe, the default value of the function parameter should be changed to `False` instead?
```diff
- inplace: bool = True, 
+ inplace: bool = False,
```


tkoyama010 (changes requested):
The documentation is correct on this. In other functions, when we want to inplace, we specify it explicitly.
Could you also add code to test for this bug?

darikg:
> The documentation is correct on this. In other functions, when we want to inplace, we specify it explicitly. Could you also add code to test for this bug?

@tkoyama010 could you clarify this? I think @dad616610 is simply proposing to to update the documentation to reflect actual behavior, are you suggesting changing the behavior to reflect the documentation? I support that but it's a more significant PR involving deprecation cycles, etc.

tkoyama010:
> > The documentation is correct on this. In other functions, when we want to inplace, we specify it explicitly. Could you also add code to test for this bug?
> 
> @tkoyama010 could you clarify this? I think @dad616610 is simply proposing to to update the documentation to reflect actual behavior, are you suggesting changing the behavior to reflect the documentation? I support that but it's a more significant PR involving deprecation cycles, etc.

Thank you for your comment. Regarding this, I am thinking that it is a simple bug and just FIX the behavior. If additional action is needed, I would appreciate your input on that.

MatthewFlamm:
I thought we aligned on having `inplace` always default to False`, but I can't find that.  There is a discussion on removing `inplace` entirely in #4158, but there was no consensus.

I'm a believer in consistency in the API being a high priority.  Nearly every other place uses a default of `False`.  The decision in my mind is whether to do a hard breaking change or a softer one where we transition the default over multiple versions.  While the documentation is what I/we want, there are likely users currently using the existing code.

Thanks for finding this!

user27182:
> I thought we aligned on having `inplace` always default to False`, but I can't find that.

Maybe this? https://github.com/pyvista/pyvista/pull/1860

Haven't read through it but at first glance this seems to suggest that `inplace=True` is a bug (not a breaking change) and that all transformations should _not_ be inplace.

MatthewFlamm:
I'm +1 for merging this one.  Since we are likely deprecating the behavior, the behavior is currently default `True`.  It makes sense that the documentation reflects this now.  @darikg will likely have to fix a small merge conflict if we do it.  As long as they are ok with that,  I think merging this makes sense.

user27182 (approved):

MatthewFlamm (approved):

tkoyama010 (approved):

PR #7033, opened 2024-12-30
Add `simplify_output` option to `contour_labels` to return 1D labels array

user27182:
### Overview

This filter outputs 2-component scalars by default. But this is only useful when internal boundaries are generated (which are not by default) and can be problematic for plotting since the `norm` is shown by default (and therefore requires care to plot these scalars correctly).

This PR adds the option to output _either_ a one- or two-component array. The default behavior is to output a single-component array for `external` boundaries, and a two-component array for the `all` and `internal` cases. This greatly simplifies the default behavior for generating external contours, since the input image and output surface of this filter will both have matching single-component labels.


pyvista/core/filters/image_data.py
```diff
@@ -1155,6 +1156,22 @@ def contour_labels(  # type: ignore[misc]
                 boundaries are generated. Do not assume the normals will point outward
                 in all cases.
 
+        multi_component_output : bool, optional
```
user27182, line 1159: Could rename this to `simplify_output` to make it a bit clearer that this is what's happening.

pyvista/core/filters/image_data.py
```diff
@@ -1155,6 +1156,22 @@ def contour_labels(  # type: ignore[misc]
...
+            - When ``boundary_type`` is ``'external'``, only the first component is
+              returned. Since external polygons may only share a boundary with the
+              background, the second component is always ``background_value`` in this
+              case and can therefore be dropped without loss of information.
+
+            - When ``boundary_type`` is ``'internal'`` or ``'all'``, the array is
+              simplified by computing the vector norm of the values.
+
+            By default, ``multi_component_output`` is ``False`` when ``boundary_type``
+            is ``'external'``, and ``True`` otherwise.
```
user27182, line 1173: We can change the default to always output a 2-component array. The option to simplify to 1D array would then need to be explicitly set as a kwarg in all cases.

user27182 (commented):
Possible improvements.

pyvista/core/filters/image_data.py
```diff
@@ -1155,6 +1156,22 @@ def contour_labels(  # type: ignore[misc]
...
+            2D array. If ``False``, the ``'boundary_labels'`` array is simplified as a
+            single-component 1D array. The simplification is as follows:
+
+            - When ``boundary_type`` is ``'external'``, only the first component is
+              returned. Since external polygons may only share a boundary with the
+              background, the second component is always ``background_value`` in this
+              case and can therefore be dropped without loss of information.
+
+            - When ``boundary_type`` is ``'internal'`` or ``'all'``, the array is
+              simplified by computing the vector norm of the values.
```
user27182, line 1170: This simplification is not ideal.
1. I think any simplification should probably generate integer values since we are dealing with categorical data. 
2. using `norm` is bad because of perfect triangles, e.g. the boundary label `[3 4]` is mapped to `5`, but this value could already be assigned to a separate foreground region `5`.

Maybe simply labelling internal boundaries sequentially with negative values is better. e.g. `[1 2] -> -1`, `[1 3] -> -2`, etc. We still lose information about _which_ regions each cell is bound to, but this way the values are unique integers and won't clash with existing foreground regions which are positive

pyvista/core/filters/image_data.py
tkoyama010: This is exactly what I was thinking. With `multi_component_output` it was unclear to me.

tkoyama010 (changes requested):

pyvista/core/filters/image_data.py
user27182: This new negative integer mapping is implemented in [5b37bbb](https://github.com/pyvista/pyvista/pull/7033/commits/5b37bbb3f6b63234fb036e552fdd9382e0021b41)

pyvista/core/filters/image_data.py
user27182: I do think simplifying by default is a good thing to keep. So this change is not currently planned.

pyvista/core/filters/image_data.py
```diff
@@ -1298,10 +1328,27 @@ def contour_labels(  # type: ignore[misc]
...
+
+        >>> contours = image.contour_labels('internal', simplify_output=True)
+        >>> contours['boundary_labels'].ndim
+        1
+        >>> np.unique(contours['boundary_labels'])
+        pyvista_ndarray([-5, -4, -3, -2, -1])
+
+        Make the values positive for plotting.
+
+        >>> contours['boundary_labels'] += 20
```
user27182, line 1342: This is only needed temporarily for this PR. This seemingly arbitrary assignment is removed with #7051.

tests/core/test_imagedata_filters.py
```diff
@@ -345,6 +346,37 @@ def test_contour_labels_pad_background(labeled_image):
...
+    poly = labeled_image.contour_labels('external', simplify_output=True)
+    assert poly[BOUNDARY_LABELS].ndim == 1
+    poly = labeled_image.contour_labels('internal', simplify_output=True)
+    assert poly[BOUNDARY_LABELS].ndim == 1
+
+    # Test `False`
+    poly = labeled_image.contour_labels('external', simplify_output=False)
+    assert poly[BOUNDARY_LABELS].ndim == 2
+    poly = labeled_image.contour_labels('internal', simplify_output=False)
+    assert poly[BOUNDARY_LABELS].ndim == 2
```
tkoyama010, line 377: I think it will be easier to understand if we parametrize this with `simplify_output`.

pyvista/core/filters/image_data.py
```diff
@@ -1536,6 +1583,26 @@ def _is_small_number(num) -> bool | np.bool_:
...
+            simplify_output = boundary_style == 'external'
+        if simplify_output:
+            # Simplify scalars to a single component
+            if boundary_style != 'external':
+                # Replace internal boundary values with negative integers
+                labels_array = output.cell_data[PV_NAME]
+                if boundary_style == 'internal':
+                    is_internal = np.full((output.n_cells,), True)
+                else:
+                    is_internal = np.all(labels_array != background_value, axis=1)
```
tkoyama010, line 1596:
```suggestion
                is_internal = np.full((output.n_cells,), True) if boundary_style == 'internal' else np.all(labels_array != background_value, axis=1)
```

pyvista/core/filters/image_data.py
```diff
@@ -1155,6 +1156,27 @@ def contour_labels(  # type: ignore[misc]
...
+            - External boundaries are simplified by keeping the first component and
+              removing the second. Since external polygons may only share a boundary
+              with the background, the second component is always ``background_value``
+              and therefore can be dropped without loss of information. The values
+              of external boundaries always match the foreground values of the input.
+
+            - Internal boundaries are simplified by assigning them unique negative
+              values sequentially. E.g. the boundary label ``[1, 2]`` is replaced with
+              ``-1``, ``[1, 3]`` is replaced with ``-2``, etc. The mapping to negative
+              values is not fixed, and can change depending on the input.
```
tkoyama010, line 1175: It would be more helpful to clearly specify the benefits or implications of modifying the value to a negative number. I understand the advantage of being able to clear the internal value by displaying only values greater than zero, but this wasn’t immediately clear to me at first.

tkoyama010 (changes requested):
All reviews have been completed. I will merge them after resolving the comments.

pyvista/core/filters/image_data.py
user27182: Above the bullet points this statement is made:
> This simplification is useful when only external boundaries
>  are generated and/or when visuallizing internal boundaries.

Maybe we can move this around to the bullet points?

user27182: Added info about benefits of negative values in [1f98c47](https://github.com/pyvista/pyvista/pull/7033/commits/1f98c47b5b35e9e825ba63c25b4c1a4af60ba280)

The benefits will also be made clear with #7051, since the negative values are used by `color_labels` for the plots

tests/core/test_imagedata_filters.py
user27182: Should be fixed [2c8363d](https://github.com/pyvista/pyvista/pull/7033/commits/2c8363d3b80d6bd1668977e197b0aa6a792235d5)

tkoyama010 (approved):
LGTM

tkoyama010:
@user27182 I haven't been able to keep track of old pull requests. If you have one that needs review and merging, please send me a review request.

PR #7028, opened 2024-12-29
Add `pad_background` parameter to `contour_labels` filter

user27182:
### Overview

New parameter to pad the image before contouring to effectively "close" the surface by default.

PR #7034, opened 2024-12-30
Add option to auto-orient normals of `contour_labels` filter

user27182:
### Overview

Split from #6732.

The default orientation of the cells generated by this filter is inconsistent, such that the external surface often points inwards instead of out. I think it's desirable to auto-orient the generated contours by default so that the normals point outward.


tkoyama010 (approved):
LGTM

tkoyama010:
It has not yet been released, so I removed a breaking-change label.

pyvista/core/filters/image_data.py
```diff
@@ -1103,6 +1104,22 @@ def contour_labels(  # type: ignore[misc]
...
+            Compute point and cell normals for the contoured output using
+            :meth:`~pyvista.PolyDataFilters.compute_normals` with ``auto_orient_normals``
+            enabled by default. If ``False``, the generated polygons may have
+            inconsistent ordering and orientation (and may negatively impact
+            the shading used for rendering).
+
+            .. warning::
+
+                Enabling this option is likely to generate surfaces with normals
+                pointing outward when ``closed_surface`` is ``True`` and
```
user27182, line 1117: Whoops, this term is left over from before #7028 and should be updated
``` diff
- ``closed_surface``
+ ``pad_background``
```

user27182: This change is captured in #7039

Issue #7036, opened 2024-12-30
Lossy voxelization 


01baftb:
### Describe the bug, what's wrong, and what you expected.

This issue is regarding [voxelize](https://docs.pyvista.org/api/utilities/_autosummary/pyvista.voxelize#pyvista-voxelize) and [voxelize_volume](https://docs.pyvista.org/api/utilities/_autosummary/pyvista.voxelize_volume) function. 

I have a 1mm x 1mm x 1mm cube, and I use _density=0.25_ for voxelize, I expect the results to have a 4x4x4=64 cells. However, what I getting is a 3x3x3=27 cells. This results in lossy-voxelization even though the specified density results in whole integer sub-divisions. 

I think the issue is related to [np.arange](https://numpy.org/doc/2.1/reference/generated/numpy.arange.html) in the [source code](https://github.com/pyvista/pyvista/blob/2fa4f7ecd54c113ef3b21e5a68febb9be2627068/pyvista/core/utilities/features.py#L17-L99) not including the end of the interval. This results in volelize to always be lossy even when the provided _density_ (i.e. voxel size) does not result in rounding error. 

I think the voxelize function should first check if the specified _density_ results in whole integer sub-division. If so, then it should make the end of the interval in the np.arange inclusive. If the sub-division does not result in whole integer, then the end interval should not be inclusive. For example, if I use _density_=0.30, I expect the results to have 3x3x3=27. 

### Steps to reproduce the bug.

```python
import numpy as np
import pyvista as pv

mesh = pv.Cube()

my_voxel_density = 0.25
voxels = pv.voxelize(mesh, density=my_voxel_density)
voxels

voxels.plot()
```

### System Information

```shell
--------------------------------------------------------------------------------

  Python 3.12.7 (tags/v3.12.7:0b05ead, Oct  1 2024, 03:06:41) [MSC v.1941 64
  bit (AMD64)]

             pyvista : 0.44.2
                 vtk : 9.3.1
               numpy : 2.1.3

--------------------------------------------------------------------------------
```

user27182:
[...]

Issue #7035, opened 2024-12-30
Voxelize results in artifacts 


01baftb:
### Describe the bug, what's wrong, and what you expected.

I am trying to import an STL file and [voxelize ](https://docs.pyvista.org/api/utilities/_autosummary/pyvista.voxelize) it. I am noticing weird artifacts (e.g. holes) in the voxelized result depending on what value is chosen for the _density_ parameter. 

Please see [voxel_test_stacked_cubes.zip](https://github.com/user-attachments/files/18276913/voxel_test_stacked_cubes.zip) which contains sample STL file for which I am having issue with. Below screenshot show the geometry of the STL. 
![image](https://github.com/user-attachments/assets/a84027dd-7a2b-4c97-b9c3-277fd4160e7c)
![image](https://github.com/user-attachments/assets/c91f42cb-10d1-4f05-b599-577d076c6937)

When I use _density=0.05_ I get weird artifacts as shown in screenshot below. Noticed bunch of missing voxels or areas in the geometry. I have noticed this issue with other chosen _density_ values, e.g. _density=0.04_
![image](https://github.com/user-attachments/assets/8c165610-ae78-4d1c-926a-91b27d60d39b)

If I use _voxelize_ without specifying the _density_, it is my understanding, the _density_ will default to _mesh.length/100.0=0.04898979485566356_. In this case, the voxelized results looks fine and does not have artifacts. 

![image](https://github.com/user-attachments/assets/cf6fb172-3a5f-4a8d-86a3-d628f2d0836b)

### Steps to reproduce the bug.

```python
import numpy as np
import pyvista as pv


mesh = pv.read('voxel_test_stacked_cubes.stl')
mesh

print(f'{mesh.length=} {mesh.length/100.0=}')
my_voxel_density = 0.05
voxels = pv.voxelize(mesh, density=my_voxel_density)
voxels

voxels.plot()

print(f'{mesh.length=} {mesh.length/100.0=}')
voxels = pv.voxelize(mesh)
voxels

voxels.plot()
```

### System Information

```shell
--------------------------------------------------------------------------------
                  OS : Windows (10 10.0.19045 SP0 Multiprocessor Free)
              CPU(s) : 8
             Machine : AMD64
        Architecture : 64bit
                 RAM : 31.4 GiB
         Environment : Jupyter
          GPU Vendor : Intel
        GPU Renderer : Intel(R) Iris(R) Xe Graphics
[...]

PR #7037, opened 2024-12-30
Deprecate `voxelize` and replace with `voxelize_unstructured_grid`

### LINKED ISSUE #1743: Stray voxels occurring when voxelizing a mesh

user27182:
### Overview

Fix #1743 #7035 #7036, maybe others too

This feature has a number of issues. Most (all?) have been solved by using a completely different and VTK-based implementation with a new filter `voxelize_binary_mask` in #6611. This implementation is also much more performant, see analysis here https://github.com/pyvista/pyvista/issues/1743#issuecomment-2343492121

I propose using `voxelize_binary_mask` as the back-end for the `voxelize` function like this:

``` python
def voxelize_unstructured_grid(
    mesh: DataSet | MultiBlock | _vtk.vtkDataSet | _vtk.vtkMultiBlockDataSet,
    *,
    dimensions: VectorLike[int] | None = None,
    spacing: float | VectorLike[float] | None = None,
    rounding_func: Callable[[VectorLike[float]], VectorLike[int]] | None = None,
    cell_length_percentile: float | None = None,
    cell_length_sample_size: int | None = None,
    mesh_length_fraction: float | None = None,
    progress_bar: bool = False,
) -> UnstructuredGrid:

    surface = wrap(mesh).extract_geometry()
    binary_mask = surface.voxelize_binary_mask(
        dimensions=dimensions,
        spacing=spacing,
        rounding_func=rounding_func,
        cell_length_percentile=cell_length_percentile,
        cell_length_sample_size=cell_length_sample_size,
        mesh_length_fraction=mesh_length_fraction,
        progress_bar=progress_bar,
    )
    voxel_cells = binary_mask.points_to_cells(copy=False).threshold(0.5)
    del voxel_cells.cell_data['mask']
    return voxel_cells
```

A similar change can be made for `voxelize_volume`.

Currently this PR deprecates `voxelize` and fully replaces it with `voxelize_unstructured_grid`. Similarly `voxelize_volume` would be replaced with `voxelize_rectilinear_grid`.

It may be possible to deprecate keywords only and keep the function names unchanged. There are differences in terminology though, so should `density` be replaced with a new `spacing` parameter?

Maybe @banesullivan can comment on this preference here: do we try to keep the `voxelize` and `voxelize_volume` functions and only update their back-end, or do we create two entirely new functions? Maybe it's better to make new functions since the back-end change is so significant?


banesullivan:
> Maybe @banesullivan can comment on this preference here: do we try to keep the voxelize and voxelize_volume functions and only update their back-end, or do we create two entirely new functions? Maybe it's better to make new functions since the back-end change is so significant?

Sorry for my delay here! My preference would be to keep the `pv.voxelize` method. The name is short and sweet and this has been around for a good while. If it isn't too bad to just swap out the backend here, I say we ought to do that (adding notes about anything that may have changed). Ultimately, i'll leave this to your judgement though

user27182:
> My preference would be to keep the `pv.voxelize` method.

What about making these filters instead? E.g. `pv.DataSetFilters.voxelize` and `pv.DataSetFilters.voxelize_volume`, but still deprecate `pv.voxelize` and `pv.voxelize_volume` ? This is done in [f9e937e](https://github.com/pyvista/pyvista/pull/7037/commits/f9e937e96e46a05243efa387262add6fa6c98e4e). Or are you partial to having these available at the pyvista namespace level?

banesullivan:
Great idea! +1

user27182:
I think we'll have to keep the old voxelize method just for creating the logo. There are some quirks (bugs) with the original implementation which gave the "V" in PyVista an asymmetric look (e.g. an `enclosed` (https://github.com/pyvista/pyvista/pull/6514) and `fit_bounds` kwargs (https://github.com/pyvista/pyvista/pull/6610) were recently added to address this). It will be difficult to recreate this intentionally with the new method


| `pv.DataSetFilters.voxelize` | `pv.voxelize` |
|---|---|
| image | image |

pyvista/core/filters/data_set.py
```diff
@@ -9350,6 +9353,774 @@ def _is_index_like(array_, max_value):
...
+        ----------
+        background_value : int, default: 0
+            Background value of the generated mask.
+
+        foreground_value : int, default: 1
+            Foreground value of the generated mask.
+
+        reference_volume : pyvista.ImageData, optional
+            Volume to use as a reference. The output will have the same ``dimensions``,
+            ``origin``, and ``spacing`` as the reference.
```
user27182, line 9867: Does it make sense to use `ImageData` here as a reference for generating a `RectilinearGrid`?

pyvista/core/filters/data_set.py
```diff
@@ -9350,6 +9353,774 @@ def _is_index_like(array_, max_value):
...
+
+        dimensions : VectorLike[int], optional
+            Dimensions of the generated mask image. Set this value to control the
+            dimensions explicitly. If unset, the dimensions are defined implicitly
+            through other parameter. See summary and examples for details.
```
user27182, line 9872: This needs to be carefully documented. ``dimensions`` here does not refer to actual `ImageData.dimensions` or `RectilinearGrid.dimensions`, which is the number of points. Instead, this refers to the number of cells.

We should either

1. update the implementation so that ``dimensions`` matches the number of points
2. rename this to ``cell_dimensions`` to make it explicit
3. keep it as ``dimensions`` but document that it means cell dimensions, not point dimensions

I'm leaning towards option 1. This would be similar to how [`resample`](https://dev.pyvista.org/api/core/_autosummary/pyvista.imagedatafilters.resample#pyvista.ImageDataFilters.resample) handles `dimensions` when resampling cell data.

pyvista/core/filters/data_set.py
```diff
@@ -9350,6 +9353,774 @@ def _is_index_like(array_, max_value):
...
+        progress_bar: bool = False,
+    ) -> UnstructuredGrid:
+        """Voxelize mesh to UnstructuredGrid.
+
+        Parameters
+        ----------
+        dimensions : VectorLike[int], optional
+            Dimensions of the generated mask image. Set this value to control the
+            dimensions explicitly. If unset, the dimensions are defined implicitly
+            through other parameter. See summary and examples for details.
```
user27182, line 10009: Similar to `voxelize_volume`, dimensions here is the number of voxel cells to use, not points.

But, `UnstructuredGrid` does not have a `dimensions` property, so it doesn't make sense for users to specify dimensions in terms of points here. I think here we should maybe rename this to `cell_dimensions` and specify the number of cells to use along each axis, not points.

user27182 (commented):
Looking for some input / feedback on some of the parameters. In particular, how do we address possible confusion about point vs. cell dimensions

pyvista/core/filters/data_set.py
```diff
@@ -9350,6 +9353,774 @@ def _is_index_like(array_, max_value):
...
+        >>> vox = mesh.voxelize(spacing=(0.01, 0.005, 0.002))
+        >>> vox.plot(show_edges=True)
+
+        The bounds of the voxelized mesh always match the bounds of the input.
+
+        >>> mesh.bounds
+        BoundsTuple(x_min=-0.13155962526798248, x_max=0.18016336858272552, y_min=-0.12048563361167908, y_max=0.18769524991512299, z_min=-0.14300920069217682, z_max=0.09850578755140305)
+
+        >>> vox.bounds
+        BoundsTuple(x_min=-0.13155962526798248, x_max=0.18016336858272552, y_min=-0.12048563361167908, y_max=0.18769524991512299, z_min=-0.14300920069217682, z_max=0.09650979936122894)
```
user27182, line 10090: This example highlighted a bug: `z_max` does not match between `mesh` and `vox`, but they should be equal. This is a bug with `voxelize_binary_mask` and will need to be fixed in a separate PR

pyvista/core/filters/data_set.py
```diff
@@ -9350,6 +9353,774 @@ def _is_index_like(array_, max_value):
...
+        stencil.SetBackgroundValue(foreground_value)
+        _update_alg(stencil, progress_bar, 'Generating binary mask')
+        output_volume = _get_output(stencil)
+
+        # Set the orientation of the output
+        output_volume.direction_matrix = reference_volume.direction_matrix
+
+        return pyvista.wrap(output_volume)
+
+    def voxelize_volume(  # type: ignore[misc]
```
user27182, line 9835: Should we keep the name `voxelize_volume`? It's a bit ambiguous since `voxelize_binary_mask` also returns a volume, but as `ImageData`.

Should we rename these, maybe `voxelize_image` and `voxelize_rectilinear`?

pyvista/core/filters/data_set.py
```diff
@@ -9350,6 +9353,774 @@ def _is_index_like(array_, max_value):
...
+        *,
+        background_value: int | float = 0,  # noqa: PYI041
+        foreground_value: int | float = 1,  # noqa: PYI041
+        reference_volume: pyvista.ImageData | None = None,
+        dimensions: VectorLike[int] | None = None,
+        spacing: float | VectorLike[float] | None = None,
+        rounding_func: Callable[[VectorLike[float]], VectorLike[int]] | None = None,
+        cell_length_percentile: float | None = None,
+        cell_length_sample_size: int | None = None,
+        mesh_length_fraction: float | None = None,
```
user27182, line 9367: Should we even keep this option?

``` python
mesh.voxelize(mesh_length_fraction=1/200)
```
is equivalent to 
``` python
mesh.voxelize(spacing=mesh.length=1/200)
```

pyvista/core/filters/data_set.py
tkoyama010: +1 `voxelize_rectilinear`

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
tkoyama010: Let's deprecate it!

banesullivan: Yeah lets just remove it if it isn't released

pyvista/core/filters/data_set.py
user27182: I think for consistency across different voxelize filters, we should just use `dimensions` and note that it's the number of points, not cells. I'll update the filter docs and implementations.

pyvista/core/filters/data_set.py
user27182: All filters now use `dimensions` defined as points and are fully consistent with `ImageData` and `RectilinearGrid` `dimensions`.

pyvista/core/filters/data_set.py
user27182: All methods now allow using `ImageData` as a reference. If needed a future PR could add RectilinearGrid

user27182:
All comments should now be resolved

user27182:
The APIs are all pretty much the same for `voxelize_binary_mask`, `voxelize_rectilinear`, and `voxelize`. Maybe it would be better to consolidate the different methods into one? And specify new keywords, e.g. `output_type`, with values `['image', 'rectilinear', 'unstructured']`, and possibly add the option `voxel_type` as either `'points'` or `'cells'`. This way, we can return either  `ImageData`,  `RectilinearGrid` , or `UnstructuredGrid`  with either voxels as points or as cells. And all with a single `voxelize` filter.

I mention this because I am considering how to resolve https://github.com/pyvista/pyvista/issues/6255. For that, we would probably need a similar  `CompositeFilters.voxelize` filter specialized for `MultiBlock`. But for that it would also be nice to have the option to choose the output type. But if we follow the current pattern from this PR, it means we would also need `CompositeFilters.voxelize_rectilinear` and `CompositeFilters.voxelize_labels`. This requires a lot of duplication, and there would be a mis-match in filter names for `ImageData` (`voxelize_binary_mask` vs `voxelize_labels`)

user27182:
Maybe it would be better to consolidate the different methods into one? And specify new keywords, e.g. `output_type`, with values `['image', 'rectilinear', 'unstructured']`, and possibly add the option `voxel_type` as either `'points'` or `'cells'`. This way, we can return either `ImageData`, `RectilinearGrid` , or `UnstructuredGrid` with either voxels as points or as cells. And all with a single `voxelize` filter.

I am going to try this out. I will make changes here until CI is green, then will create a new PR with the new filter. This PR will stay for deprecation only, for possible merge _after_ creating the new `DataSetFilters.voxelize` filter.

user27182:
> > Maybe it would be better to consolidate the different methods into one? And specify new keywords, e.g. `output_type`, with values `['image', 'rectilinear', 'unstructured']`, and possibly add the option `voxel_type` as either `'points'` or `'cells'`. This way, we can return either `ImageData`, `RectilinearGrid` , or `UnstructuredGrid` with either voxels as points or as cells. And all with a single `voxelize` filter.
> 
> I am going to try this out. I will make changes here until CI is green, then will create a new PR with the new filter. This PR will stay for deprecation only, for possible merge _after_ creating the new `DataSetFilters.voxelize` filter.

Ideally, this would be the case, and we would have a single voxelize function that can generate any output type. I did try this initially in #7239, but this adds a lot of complexity to what is already a fairly complex deprecation as part of this PR. Plus, it is more of an under-the-hood change, and filters like `voxelize`, `voxelize_rectiliear`, and `voxelize_binary_mask` would still exist and remain largely as-is from the changes in this PR.

Therefore, I have reverted recent changes back to when CI was green in https://github.com/pyvista/pyvista/pull/7037/commits/9135b82ce5b8aa64c7316d5057e915e30d847c9a, and I think this PR is ready to review as-is.

banesullivan (approved):

PR #7048, opened 2025-01-02
Allow ExodusIIReader to handle all the blocks and sets in the Exodus II file

mattfalcone1997:
### Overview

Allows all the important blocks and sets to be read alongside their arrays. This is implemented through the `ExodusIIBlockSet` class.

### Details

1. Implementation of `ExodusIIBlockSet` which handles each of the blocks and sets and their arrays. Both the blocks/sets and arrays are enabled similarly to the `PointCellDataSelection` with enable, disable and status methods. 
  * The blocks and sets implemented are
    * Element blocks
    * Face blocks
    * Edge blocks
    * Side sets
    * Node sets
    * Face sets
    * Element sets
  * At the moment, these are stored in the `ExodusIIReader` as `_element_blocks` and accessed through the `element_blocks` property. This could be changed so that the `ExodusIIBlockSet` is created and returned by the property and not permanently stored within the `ExodusIIReader` object.
2. Global result arrays can now also be accessed similarly to the `PointCellDataSelection` with `enable_global_array` etc. These can be read using the `read_global` method, which returns a `Table`.
  * The global data could also be added to the field_data attribute of the `MultiBlock` instance returned by `read` instead.
3. download_biplane example added
  * Added to test the `enable_array` method in the side_sets `ExodusIIBlockSet` class.

Other considerations include
* Deprecation or reimplementation of `read_exodus` to use the `ExodusIIReader` class
* Not sure where to put the `ExodusIIBlockSet` class in the documentation. At the moment, it is next to the `ExodusIIReader`, but it is not a reader!


tests/core/test_reader.py
```diff
@@ -1365,6 +1365,27 @@ def test_exodus_reader():
         e_reader.enable_cell_array(name)
         assert e_reader.cell_array_status(name)
 
+    ## check global arrays handling
```
tkoyama010, line 1368:
```suggestion
    # check global arrays handling
```

tests/core/test_reader.py
```diff
@@ -1414,3 +1435,87 @@ def test_exodus_reader():
...
+    assert block.names == names
+
+    block.enable_all()
+
+    for name in names:
+        assert block.status(name)
+
+    block.disable_all()
+    for name in names:
+        assert not block.status(name)
```
tkoyama010, line 1451:
```suggestion
    block.disable_all()

    for name in names:
        assert not block.status(name)
```

tests/core/test_reader.py
```diff
@@ -1414,3 +1435,87 @@ def test_exodus_reader():
...
+
+    _test_block_arrays(e_reader.element_blocks, ['aux_elem'])
+
+    # Test example with set arrays
+    fname_e = examples.download_biplane(load=False)
+    e_reader = pv.get_reader(fname_e)
+    _test_block_arrays(e_reader.side_sets, ['PressureRMS'])
+
+    # check construct_result_array for those that do not have blocks
+    # in the example
```
tkoyama010, line 1510:
```suggestion
    # check construct_result_array for those that do not have blocks in the example
```

tkoyama010 (changes requested):
Thanks!

tkoyama010 (approved):
LGTM

PR #6543, opened 2024-08-19
Add `transform` filter for `MultiBlock`

user27182:
### Overview

Recursively transform all datasets in a multiblock.

PR #7053, opened 2025-01-04
Deprecate dataset_filters.transform defaulting to inplace=True

darikg:
Cf #7027

Draft PR, just curious how much churn a soft-deprecation would be. 

1. Change the actual default of `inplace` to `None`. If called with `inplace=None`, raise a deprecation warning asking the user to explicitly supply a True/False value for `inplace`, but continue on with `inplace=True` so backwards compatibility is maintained.

2. Change the documentation to correctly indicate the default value is actually `True`, but add a deprecation note confirming the default will change in the future to `inplace=False`.

3. Try to find all the places `transform` is called in the codebase and make sure inplace is specified explicitly.

TODO:
- [ ] Didn't run all the tests locally, assume a bunch more will fail
- [ ] `CompositeFilters.transform` also defaults to True, need to add deprecations there too
- [ ] Add required version checks in the tests so the deprecation warning becomes an error, etc

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -7024,7 +7024,7 @@ class DataSetFilters:
         self: ConcreteDataSetType,
         trans: TransformLike,
         transform_all_input_vectors: bool = False,
-        inplace: bool = True,
+        inplace: bool | None = None,
         progress_bar: bool = False,
     ):
         """Transform this mesh with a 4x4 transform.
@@ -7068,8 +7068,12 @@ class DataSetFilters:
             transformed. Otherwise, only the normals and vectors are
             transformed.  See the warning for more details.
 
-        inplace : bool, default: False
+        inplace : bool, default: True
             When ``True``, modifies the dataset inplace.
+
+            .. deprecated:: 0.45.0
+            `inplace` was previously defaulted to `True`. For API consistency this will be
+            changed to `False`.
 
         progress_bar : bool, default: False
             Display a progress bar to indicate progress.
@@ -7105,10 +7109,20 @@ class DataSetFilters:
         ...         [0, 0, 0, 1],
         ...     ]
         ... )
-        >>> transformed = mesh.transform(transform_matrix)
+        >>> transformed = mesh.transform(transform_matrix, inplace=False)
         >>> transformed.plot(show_edges=True)
 
         """
+        if inplace is None:
+            warnings.warn(
+                """
+The default value of `inplace` for the `dataset.transform` filter will change in the future.
+ Previously it defaulted to `True`, but will change to `False`.
+ Explictly set `inplace` to `True` or `False` to silence this warning. """,
+                PyVistaDeprecationWarning,
+            )
+            inplace = True  # The old default behavior
+
         if inplace and isinstance(self, pyvista.RectilinearGrid):
             raise TypeError(f'Cannot transform a {self.__class__} inplace')
 
@@ -8492,7 +8506,7 @@ class DataSetFilters:
             if box_style == 'outline':
                 face.copy_from(pyvista.lines_from_points(face.points))
             if oriented:
-                face.transform(inverse_matrix)
+                face.transform(inverse_matrix, inplace=True)
 
         # Get output
         alg_output = box if as_composite else _multiblock_to_polydata(box)
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -196,7 +196,7 @@ class _PointSet(DataSet):
         self: ConcretePointSetType,
         xyz: VectorLike[float],
         transform_all_input_vectors: bool = False,
-        inplace=None,
+        inplace=False,
     ):
         """Translate the mesh.
 
--- a/pyvista/core/filters/composite.py
+++ b/pyvista/core/filters/composite.py
@@ -227,7 +227,7 @@ class CompositeFilters:
         self,
         trans: TransformLike,
         transform_all_input_vectors: bool = False,
-        inplace: bool = True,
+        inplace: bool = True,  # TODO also needs deprecation warning
         progress_bar: bool = False,
     ):
         """Transform all blocks in this composite dataset.
@@ -272,7 +272,7 @@ class CompositeFilters:
         >>> import pyvista as pv
         >>> mesh = pv.MultiBlock([pv.Sphere(), pv.Plane()])
         >>> transform = pv.Transform().translate(50, 100, 200)
-        >>> transformed = mesh.transform(transform)
+        >>> transformed = mesh.transform(transform, inplace=False)
         >>> transformed.plot(show_edges=True)
 
         """
--- a/pyvista/core/utilities/geometric_sources.py
+++ b/pyvista/core/utilities/geometric_sources.py
@@ -84,7 +84,7 @@ def translate(
     trans[:3, 2] = normz
     trans[3, 3] = 1
 
-    surf.transform(trans)
+    surf.transform(trans, inplace=True)
     if not np.allclose(center, [0.0, 0.0, 0.0]):
         surf.points += np.array(center, dtype=surf.points.dtype)
 
--- a/pyvista/core/utilities/transform.py
+++ b/pyvista/core/utilities/transform.py
@@ -211,7 +211,7 @@ class Transform(_vtk.vtkTransform):
     produce different transformations.
 
     >>> mesh_post = pv.Sphere().transform(transform_post)
-    >>> mesh_pre = pv.Cone().transform(transform_pre)
+    >>> mesh_pre = pv.Cone().transform(transform_pre, inplace=False)
     >>> pl = pv.Plotter()
     >>> _ = pl.add_mesh(mesh_post, color='goldenrod')
     >>> _ = pl.add_mesh(mesh_pre, color='teal')
@@ -245,7 +245,7 @@ class Transform(_vtk.vtkTransform):
     Transform the mesh by its inverse to restore it to its original un-scaled state
     and positioning at the origin.
 
-    >>> mesh_pre_inverted = mesh_pre.transform(inverse_matrix)
+    >>> mesh_pre_inverted = mesh_pre.transform(inverse_matrix, inplace=False)
     >>> pl = pv.Plotter()
     >>> _ = pl.add_mesh(mesh_pre_inverted, color='teal')
     >>> _ = pl.add_axes_at_origin()
--- a/tests/core/test_dataset_filters.py
+++ b/tests/core/test_dataset_filters.py
@@ -3682,7 +3682,13 @@ def test_transform_imagedata_warns_with_shear(uniform):
         match='The transformation matrix has a shear component which has been removed. \n'
         'Shear is not supported when setting `ImageData` `index_to_physical_matrix`.',
     ):
-        uniform.transform(shear)
+        uniform.transform(shear, inplace=True)
+
+
+def test_transform_filter_inplace_default_warns(cube):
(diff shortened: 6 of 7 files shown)


pyvista/core/pointset.py
```diff
@@ -196,7 +196,7 @@ def translate(  # type: ignore[misc]
         self: ConcretePointSetType,
         xyz: VectorLike[float],
         transform_all_input_vectors: bool = False,
-        inplace=None,
```
darikg, line 199: does anyone know why `translate` has its `inplace` default to `None` instead of `False`?

user27182:
> * [ ]  `CompositeFilters.transform` also defaults to True, need to add deprecations there too

This is added in #6543 and is not yet released. No deprecation needed here to make `inplace=False` by default.

pyvista/core/pointset.py
user27182: Looking at the blame, this change appears to be made in https://github.com/pyvista/pyvista/pull/1860

I suspect it was made `None` for the same reason `None` is used by this PR: for deprecation.

darikg: ok great, 2021 seems long enough ago that it's safe to update that to `False` now

darikg:
> EDIT: Maybe we should still deprecate it, so that it's consistent with `dataset.transform`? I think the reason it was made `True` by default was for the same reason: consistency. Otherwise doing `object.transform` is ambiguous w.r.t. `inplace` if you're working with both `MultiBlock` and regular datasets.

Sure, if we go through with this PR I think it makes sense to enforce explicit opt-in/opt-out `inplace` behavior on all transform calls through the deprecation cycles

pyvista/core/filters/data_set.py
```diff
@@ -7105,10 +7109,20 @@ def transform(  # type: ignore[misc]
...
         >>> transformed.plot(show_edges=True)
 
         """
+        if inplace is None:
+            warnings.warn(
+                """
+The default value of `inplace` for the `dataset.transform` filter will change in the future.
+ Previously it defaulted to `True`, but will change to `False`.
+ Explictly set `inplace` to `True` or `False` to silence this warning. """,
+                PyVistaDeprecationWarning,
```
user27182: Is a `PyVistaFutureWarning` more appropriate? 

```suggestion
                PyVistaFutureWarning,
```

Though I don't see reference to this warning in CONTRIBUTING, not sure what the policy should be here. #1860 also used `PyVistaDeprecationWarning` for the same purpose, though it's also an older PR...



A future warning is used by `extrude` though for the same kind of default value change:
https://github.com/pyvista/pyvista/blob/f9fc22d304b37a0e8f91207dee55b0c6ee75f92f/pyvista/core/filters/poly_data.py#L3086-L3090

darikg: Yeah I always lean pretty heavily on CONTRIBUTING for this kinda stuff because I can never remember what's exactly expected. If PyVistaFutureWarning is preferred would like CONTRIBUTING updated / clarified

darikg:
A single test failure in [Linux Unit Testing (3.10, 9.2.2, latest)](https://github.com/pyvista/pyvista/actions/runs/12620370823/job/35165990116?pr=7053#step:12:6201) is almost certainly a numerical issue in validate_rotation checking a random rotation matrix:

```
__________________ test_bounding_box_return_meta[False-True] ___________________

oriented = True, as_composite = False

    @pytest.mark.parametrize('oriented', [True, False])
    @pytest.mark.parametrize('as_composite', [True, False])
    def test_bounding_box_return_meta(oriented, as_composite):
        # Generate a random rotation matrix
        vector = np.random.default_rng().random((3,))
        angle = np.random.default_rng().random((1,)) * 360
        rotation = pv.transformations.axis_angle_rotation(vector, angle)
    
        # Transform a box manually and get its OBB
        box_mesh = pv.Cube(x_length=1, y_length=2, z_length=3)
        box_mesh.transform(rotation, inplace=True)
>       obb, point, axes = box_mesh.bounding_box(
            oriented=oriented, return_meta=True, as_composite=as_composite
        )

tests/core/test_dataset_filters.py:4461: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pyvista/core/filters/data_set.py:8453: in bounding_box
    return self.oriented_bounding_box(
pyvista/core/filters/data_set.py:8298: in oriented_bounding_box
    alg_input, matrix = self.align_xyz(
pyvista/core/filters/data_set.py:431: in align_xyz
    rotation = Transform().rotate(axes)
pyvista/core/utilities/transform.py:1003: in rotate
    valid_rotation = _validation.validate_rotation(rotation)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

rotation = pyvista_ndarray([[ 6.1753786e-01,  4.8325321e-01, -6.2057501e-01],
                 [-2.1952267e-04,  7.8909826e-01,  6.1426693e-01],
                 [ 7.8654110e-01, -3.7919688e-01,  4.8740414e-01]],
                dtype=float32)
must_have_handedness = None, name = 'Rotation'
```

Two things I want to note before I forget. 

1. It's suspicious that the rotation matrix is dtype=float32. That seems like a loss of precision originating from a wrapped VTK algorithm somewhere that should be defaulting to float64.
2. I think `np.allclose(np.linalg.inv(rotation_matrix), rotation_matrix.T)` seems fishy. Probably more robust with something like 
```
diff = np.linalg.inv(rotation_matrix) - rotation_matrix.T
norm_diff = np.linalg.norm(diff, 'fro')
assert norm_diff < TOLERANCE
```
this particular matrix has a norm_diff of 6.6e-8.

user27182:
This test is known to be flaky. See also https://github.com/pyvista/pyvista/pull/6724#issuecomment-2525364619

I think the 32 bit dtype is probably OK, the rotation matrix is constructed from `principal_axes` of the mesh and our geometric source meshes have 32 bit points by default. 

My thought on this is that the randomness is to blame, I think every so often this generates a rotation matrix that is numerically unstable causing the inverse calc to be out of tolerance.

Not 100% sure though. Would be good to investigate further and properly address this. For now re-running the test should pass.

darikg:
> I think the 32 bit dtype is probably OK, the rotation matrix is constructed from `principal_axes` of the mesh and our geometric source meshes have 32 bit points by default.

Oh right, so if the source is 64-bit the output is as well. That seems fine. Curiously a 32-bit `point` does actually manage to sneak in somehow:

```python
import pyvista as pv
import numpy as np
mesh = pv.Cube(point_dtype='float64')
obb, point, axes = mesh.bounding_box(oriented=True, return_meta=True, as_composite=True)
point.dtype, axes.dtype
```
> (dtype('float32'), dtype('float64'))

But `point` is float64 for the other combinations of `oriented` and `as_composite`. Not a very big deal, obviously.


> This test is known to be flaky. See also [#6724 (comment)](https://github.com/pyvista/pyvista/pull/6724#issuecomment-2525364619)
>
> My thought on this is that the randomness is to blame, I think every so often this generates a rotation matrix that is numerically unstable causing the inverse calc to be out of tolerance.
> 
> Not 100% sure though. Would be good to investigate further and properly address this. For now re-running the test should pass.

That and only 32-bits is making the instability of the inverse calc worse. More generally I think a validate_rotation function that incorrectly raises on a valid rotation matrix is wrong, not flaky. Probably need a tolerance argument (like you suggested in that other thread) defaulting to something pretty generous, like 1e-3.

user27182:
> > I think the 32 bit dtype is probably OK, the rotation matrix is constructed from `principal_axes` of the mesh and our geometric source meshes have 32 bit points by default.
> 
> Oh right, so if the source is 64-bit the output is as well. That seems fine. Curiously a 32-bit `point` does actually manage to sneak in somehow:
> 
> ```python
> import pyvista as pv
> import numpy as np
> mesh = pv.Cube(point_dtype='float64')
> obb, point, axes = mesh.bounding_box(oriented=True, return_meta=True, as_composite=True)
> point.dtype, axes.dtype
> ```
> 
>  > (dtype('float32'), dtype('float64'))
> 
> But `point` is float64 for the other combinations of `oriented` and `as_composite`. Not a very big deal, obviously.

Curious, this dtype mismatch does seem like a bug.

> More generally I think a validate_rotation function that incorrectly raises on a valid rotation matrix is wrong, not flaky. Probably need a tolerance argument (like you suggested in that other thread) defaulting to something pretty generous, like 1e-3.

Good point, this is also a bug. I liked your idea of using the Frobenius norm instead. And for the tolerance we can set it dynamically as a fraction of the precision of the dtype used.

darikg:
> Good point, this is also a bug. I liked your idea of using the Frobenius norm instead. 

Nice. Actually I think `norm_diff = np.linalg.norm(rotation_matrix @ rotation_matrix.T - eye(3), ord='fro')` might be slightly more sound since it doesn't have to explicitly calculate the inverse.

> And for the tolerance we can set it dynamically as a fraction of the precision of the dtype used.

Oh yeah, that seems wise.

pyvista/core/filters/data_set.py
user27182: Let's keep deprecation then.

pyvista/core/filters/_deprecate_transform_inplace_default_true.py
```diff
@@ -0,0 +1,39 @@
...
+def check_inplace(cls: type, inplace: bool | None = None) -> bool:  # numpydoc ignore=RT01
+    """Check if user explicitly opted into inplace behavior."""
+    # Deprecated v0.45, convert to error in v0.48, remove v0.51
+    if inplace is None:
+        # if inplace is None user has not explicitly opted into inplace behavior
+
+        if version_info >= (0, 48):
+            raise RuntimeError("""
+Convert this deprecation warning into an error.
+Also update docstrs for `DataSetFilters.transform` and `CompositeFilters.transform`
```
user27182, line 24: Do we need to use `"""` here? Something about the non-alignment to the code is irksome, can we use concatenated regular strings instead so that the white space is aligned with the code?

user27182 (changes requested):

pyvista/core/filters/_deprecate_transform_inplace_default_true.py
```diff
@@ -0,0 +1,40 @@
...
+from pyvista._version import version_info
+
+
+def check_inplace(cls: type, inplace: bool | None = None) -> bool:  # numpydoc ignore=RT01
+    """Check if user explicitly opted into inplace behavior."""
+    # Deprecated v0.45, convert to error in v0.48, remove v0.51
+    if inplace is None:
+        # if inplace is None user has not explicitly opted into inplace behavior
+
+        if version_info >= (0, 48):
```
user27182, line 21:
```suggestion
        if version_info >= (0, 48):  # pragma: no cover
```

pyvista/core/filters/_deprecate_transform_inplace_default_true.py
```diff
@@ -0,0 +1,40 @@
...
+            raise RuntimeError(
+                'Convert this deprecation warning into an error. '
+                'Also update docstrs for `DataSetFilters.transform` and `CompositeFilters.transform`.'
+            )
+        if version_info >= (0, 51):
```
user27182, line 26:
```suggestion
        if version_info >= (0, 51):  # pragma: no cover
```

user27182 (dismissed):

user27182 (approved):

user27182:
waiting for #7027 to merge first

PR #7055, opened 2025-01-05
Fix undefined variable exception in conditional check for `Rotation` import

darikg:
I made a small mistake in #7023 

the conditional check prior to that was `if 'Rotation' in locals()`, which became `if Rotation is not None`, but since the `ImportError` was suppressed, it would be possible to get an undefined variable exception if importing during `TYPE_CHECKING` or a documentation build where `scipy` is not importable


darikg:
@user27182 could you have a look when you get a chance?

user27182 (approved):
LGTM

tkoyama010:
@darikg I revised the title to make the release notes clearer. Please feel free to make any necessary corrections if this doesn't align with your intentions.

PR #7057, opened 2025-01-05
Add `threshold_otsu` filter to `ImageDataFilters`

tkoyama010:
### Overview

See [Otsu thresholding in Scientific Python Lectures](https://lectures.scientific-python.org/packages/scikit-image/auto_examples/plot_threshold.html)

<p align="center">
  Otsu thresholding in Scientific Python Lectures
</p>

### Details

- [Separate Foreround and Background Using Otsu Method](https://examples.itk.org/src/filtering/thresholding/separategroundusingotsu/documentation)
- [Threshold an Image Using Otsu](https://examples.itk.org/src/filtering/thresholding/thresholdanimageusingotsu/documentation)


MatthewFlamm:
Itk is another c++ library like vtk with Python bindings.  Are there graphics related functionality to consider for including this?

I feel strongly that this should at a minimum be an optional dependency and possibly a plugin package like `pyvista-image`.

user27182:
Tools like 3DSlicer use both ITK and VTK. I believe ITK is used mainly for filters and VTK is used for visualization. 

There is a really annoying but very important consideration when working with these two packages: image orientation can differ. See https://www.slicer.org/w/index.php/Slicer:VTK_ITK_Coordinates

If we do use ITK I feel that this orientation difference should be correctly handled (and tested) internally so that this issue is abstracted away from the user.

tkoyama010:
Thank you for your comment. Personally, I believe this functionality should be added as an option to the `[image]` package rather than being developed as a separate package. Implementing it as part of the ImageData method would make it more convenient for users.

banesullivan:
@tkoyama010 makes a great point in
> I believe this functionality should be added as an option to the [image] package rather than being developed as a separate package. Implementing it as part of the ImageData method would make it more convenient for users.

I think we should weight the pros/cons of letting ITK be an optional dependency. I agree this would make it more streamlined and give us access to a great suite of ImageData filters. However, I'm concerned about the complexity of maintaining an optional dependency that could easily balloon the methods on the ImageData class.

If we go the route of an optional dependency, then one way to do this is to have all of the ITK methods isolated in their own module (`.py` file) in their own mixin class (e.g., `ITKImageDataFilters`) to keep a healthy separation

user27182:
> I think we should evaluate the pros/cons of letting ITK be an optional dependency. I agree this would make it more streamlined and give us access to a great suite of ImageData filters. However, I'm concerned about the complexity of maintaining an optional dependency that could easily balloon the methods on the ImageData class. And I don't want us to be on the hook for maintaining support across ITK releases
> 
> If we go the route of an optional dependency, then one way to do this is to have all of the ITK methods isolated in their own module (`.py` file) in their own mixin class (e.g., `ITKImageDataFilters`) to keep a healthy separation

I would also be concerned about supporting ITK optionally, could be a huge can of worms requiring similar levels of support and testing as VTK (though maybe `SimpleITK` is not as onerous)

One of the main reasons to use VTK and/or ITK at all is for their support of spatial images (as opposed to simply representing an image as a multidimensional array). So if there's an ITK filter that's needed because of this spatial support that is otherwise absent from any other similar filter from another package, then maybe we should consider ITK. But for `otsu` thresholding specifically, this filter only really operates on the scalar values, so in this case another library like `scikit-image`'s [threshold_multiotsu](https://scikit-image.org/docs/stable/api/skimage.filters.html#skimage.filters.threshold_multiotsu) or [threshold_otsu](https://scikit-image.org/docs/stable/api/skimage.filters.html#skimage.filters.threshold_otsu) may be used. Might be easier to have this as an optional dependency instead of ITK.

If we go this route then the main issue for us would probably have to do with reshaping the data to a multidimensional array expected by `scikit-image`. To get the reshaping right, we might need a new API (e.g. see #7100). Or, if we'd rather be more hands off, instead of creating a new pyvista filter layered on top of a `scikit-image` filter we could instead just show a bunch of examples of how _any_ array-based image filter package can be used in combination with pyvista, and simply show how to reshape the data to make this work.

banesullivan:
In my opinion, any ITK wrappings in PyVista would be better suited in a new `pyvista-itk` standalone package

banesullivan (changes requested):
Formalizing my disapproval of adding ITK as a dependency to PyVista

tkoyama010:
My intuition to add this as an option was wrong. Let's create a new repository. Your opinions helped me realize that. Thanks all.

banesullivan:
Sorry to be discouraging here! Otherwise I think this is a valuable effort, I'm just concerned the main pyvista repo/project isn;t the best place for it

Issue #6222, opened 2024-06-07
Connectivity filter crashes kernel, works on computers with older package/ python versions


NKoerth:
### Describe the bug, what's wrong, and what you expected.

Hello- I hope this is the correct place to post this, but this might be a vtk related issue and not a pyvista one. 
We have a script that used to work on all machines, but now it only seems to run on computers with older versions of python/ pyvista/ vtk, but not with updated versions. 
It crashes my kernel when trying to apply the connectivity filter to the pyvista grid. When testing with a lower resolution grid, I can sometimes run this without it crashing. My coworker (with older package versions) can run this code (with our higher or lower resolution grids) without problem. 

The message I get in the Jupyter log when it crashes is:
[error] Disposing session as kernel process died ExitCode: 3221225477, Reason: 

### Steps to reproduce the bug.

``` python
import numpy as np
import pyvista as pv

# Dataset extent
xmin = -55213.82
xmax = 37514.74
ymin = -165885.97
ymax = -85496.65 
zmin = 0.3048
zmax = 121.92

#Example grid
xres = (xmax- xmin) / 250
yres = (ymax - ymin) / 250
zres = (zmax - zmin) / 26
xx, yy, zz = np.mgrid[xmin:xmax:xres, ymin:ymax:yres, zmin:zmax:zres]

grid = pv.StructuredGrid(xx, yy, zz)

# Fake dataset for grid 
grid_vals = np.sin(xx**2 + yy**2) / (xx**2 + yy**2)
grid_vals = (grid_vals - np.min(grid_vals)) / (np.max(grid_vals) - np.min(grid_vals)) 

grid['value'] = grid_vals.ravel(order='F').astype(float)

# Filtering out values below user specific threshold 
filter_grid = grid.ptc().threshold([0.7, 1.1], 'value')

# Find connected bodies
filter_grid['Volume'] = np.ones((filter_grid.points.shape[0], 0))
filter_grid = filter_grid.connectivity()  # Crashes here
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Fri Jun 07 11:35:12 2024 Pacific Daylight Time

                  OS : Windows
              CPU(s) : 12
             Machine : AMD64
        Architecture : 64bit
                 RAM : 15.8 GiB
         Environment : Jupyter
          GPU Vendor : Intel
        GPU Renderer : Intel(R) Iris(R) Xe Graphics
         GPU Version : 4.5.0 - Build 31.0.101.4502
[...]

PR #7060, opened 2025-01-06
Do not allow setting empty arrays for non-empty meshes

user27182:
### Overview
Partial fix for #6222. See https://github.com/pyvista/pyvista/issues/6222#issuecomment-2573659530.

### Details
It is currently possible to set an empty array for a non-empty mesh:
``` python
import pyvista as pv
import numpy as np
mesh = pv.Sphere() 
mesh['array'] = np.zeros((mesh.n_points, 0))

print(mesh['array'])  # pyvista_ndarray([], dtype=float64)
```

But this should not be possible. If we set the array to itself we get the expected error:
``` python
mesh['array'] = mesh['array']
# ValueError: Number of scalars (0) must match either the number of points (842) or the number of cells (1680).
```


pyvista/core/datasetattributes.py
```diff
@@ -788,7 +788,10 @@ def _prepare_array(
                 raise ValueError(
                     f'data length of ({data.shape[0]}) != required length ({array_len})',
                 )
-
+            if array_len > 0 and data.size == 0:
+                raise ValueError(
+                    f"Empty arrays are not allowed. Array '{name}' cannot have a size of 0."
```
darikg, line 793: Maybe put `{data.shape}` in the error message too? I could see this message being a little confusing if the offending line is something like `mesh['foo'] = np.zeros((mesh.n_points, n_dims))` if `n_dims` has been set to zero a non-trivial number of lines of code away.

pyvista/core/datasetattributes.py
```diff
@@ -788,7 +788,10 @@ def _prepare_array(
                 raise ValueError(
                     f'data length of ({data.shape[0]}) != required length ({array_len})',
                 )
-
+            if array_len > 0 and data.size == 0:
```
user27182, line 791: This is a better condition since `len` only check the first axis. This change can handle cases like `(0, N)` shape
```suggestion
            if not all(data.shape):
```

MatthewFlamm:
How does this interact with field data?  Does it make sense that field data could be an empty array?

user27182:
> How does this interact with field data? 

This impacts field data in the same way. So empty arrays are not allowed with this PR. Tests confirm this in [338c2ab](https://github.com/pyvista/pyvista/pull/7060/commits/338c2ab512250222113201b7579eaa90f42685d4)

> Does it make sense that field data could be an empty array?
 
We don't have the same constraints here. I don't think it would be the same source of bugs as it could be with point/cell data. So we could allow it. But something weird seems to be happening with setting/getting empty arrays. See original example where the empty array is initially valid, but then setting second time it's no longer valid.

So currently it seems empty arrays aren't really handled properly anyway. I think it would be wise to be conservative here and not allow any empty arrays, even for field data, unless we explicitly support it and test that the correct empty shape is preserved when setting/getting the array.

pyvista/core/datasetattributes.py
```diff
@@ -788,7 +788,10 @@ def _prepare_array(
                 raise ValueError(
                     f'data length of ({data.shape[0]}) != required length ({array_len})',
                 )
-
+            if any(data.shape) and data.size == 0:
```
MatthewFlamm, line 791: This line only evaluates true when there is a nonzero entry in `shape`.  It will skip this when there a shape like `(0,)`.  This seems valid as a field data even for a nonempty mesh.  Can we test that?

MatthewFlamm: Or do you intend that this use case is also invalid?

user27182: The `(0,)` case is already covered in the check above with the line:
`if data.shape[0] != array_len:`
where `array_len` is either `n_points` for point data, `n_cells` for cell data, or the actual length of the input array for field data.

This means that the `(0,)` case is not allowed for non-empty meshes for point and cell data, but is allowed for field data.

And, as you pointed out, this `(0,)` case is not caught by the new logic from this PR. This means that `(0,)` sized empty meshes are allowed for field data only with the current logic. Any other empty shapes are not allowed though (e.g. `(0,1)`). 

The tests have been updated in [9b2e3e9](https://github.com/pyvista/pyvista/pull/7060/commits/9b2e3e97bd5682322ee0110d21fd5fffc81a099b) to confirm this.

(I think other special cases like `(0, 0)` may also valid for field data with non-empty meshes, but this is untested.)

Is this the behavior you are expecting?

user27182: `(0, 0)` case is tested in [fef56f1](https://github.com/pyvista/pyvista/pull/7060/commits/fef56f11a385c7844633921603e1d04fe1b82f61)

This means that currently;

- Empty point, cell, or field data arrays are allowed for empty meshes
- Empty point or cell data arrays (i.e. any dimension with shape 0) are not allowed for non-empty meshes
- Empty field data arrays where any shape is non-zero are not allowed
- Empty field data arrays with only zero shape value(s) are allowed

darikg: might also need yet another special case for `np.zeros(())` which has shape `()`

user27182: > might also need yet another special case for `np.zeros(())` which has shape `()`

This is a good test case to check. When I tried adding this though it exposed a bug where the array is unexpectedly broadcasted.

``` python
import numpy as np
poly = pv.examples.load_uniform()
array = np.ones(())
print(array.size) # size is 1 
poly.cell_data['data'] = array
print(poly['data'].size) # size is 729
```

So, although I agree we should check this case, I think this bug should be fixed first. The tests in this PR are already a bit tricky to read and follow, and adding logic to make the tests pass for this bug will make it messy (and will need to be updated along with the bug fix).

I'm thinking we can exclude this case for now to reduce the scope of this PR. Maybe add a `#TODO` comment to the tests?

user27182: Also, since the size of an array with shape `()` is `1`, not `zero` it's not really an empty array, so perhaps would be best tested separately anyway (and in a separate PR) and not included as a new parameter with the current tests

MatthewFlamm: I'm +1 for that plan, and your proposal is a good one.  Thanks for considering many edge cases here.

user27182: This bug is tracked in #7076 and any fixes should handle the `()` case separately with its own tests. So I won't add a `TODO` note to the tests here.

tkoyama010 (approved):
LGTM

PR #7064, opened 2025-01-07
Fix `color_labels` color cycling mode

user27182:
### Overview

`color_labels` `cycler` mode should cycle through colors by default when there are fewer colors than values. But currently it doesn't.
``` python
import pyvista as pv
mesh = pv.examples.download_foot_bones()
mesh.connectivity().color_labels(['red','lime','blue'], coloring_mode='cycler').plot()
```
The plot should show repeating colors, with no black.
image

This PR fixes this. The fix includes a change from using `cycler` to `itertools.cycle`, so the mode is also renamed to `'cycle'` which I think is a better name.

The examples are also extended and updated for clarity.


tkoyama010 (approved):
LGTM

PR #7068, opened 2025-01-08
Add image resampling filter

### LINKED ISSUE #5136: Add vtkImageResample  filter for upscaling
## Description

How would we go about properly upscaling a mesh much like the image from [this SO post](https://stackoverflow.com/questions/40988864/creating-a-grid-using-vtk-activiz-c-sharp?rq=1):

tuVWd

## Example Data

I started this using the `sample` filter, but this isn't exactly what I'm looking for as the sample filter can only sample onto the point data of the output grid. I'd like a way to properly upscale a uniform/ImageData grid on the cell data. Also, the coarse grid that I created does not perfectly align with the fine grid which we should fix.

```py
import pyvista as pv
from pyvista import examples

fine = examples.load_channels()
# Perfrom an upscaling of sorts
coarse = pv.create_grid(fine, dimensions=(50,50,45)).sample(fine)

# Show it
dp = dict(cmap='Reds', show_edges=True, lighting=False)

p = pv.Plotter(shape=(1,2), notebook=False)
p.add_mesh(fine, **dp)

p.subplot(0,1)
p.add_mesh(coarse, **dp)

p.link_views()
p.show()
```

Screen Shot 2019-07-22 at 12 10 47 PM

user27182:
### Overview

Resolve #5136.

The `resample` filter uses `vtkImageResamle` and is generalized and works with both point and cell data. Resampling can be done by setting a sampling rate, specifying the output dimensions, or specifying a reference image.

~The implementation is based on 3DSlicer's implementation which handles oriented images and uses `vtkImageReslice` under the hood.
https://github.com/Slicer/Slicer/blob/d2fca7d34fd0b29214385c08519434ca74ed1d08/Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx#L262~


pyvista/core/filters/image_data.py
```diff
@@ -2923,3 +2923,485 @@ def _validate_dimensional_operation(
...
 
         return dimensions_mask, dimensions_result
+
+    def resample(  # type: ignore[misc]
+        self: ImageData,
+        interpolation: Literal['linear', 'nearest', 'cubic'] | None = None,
+        *,
+        reference_image: ImageData | None = None,
+        dimensions: VectorLike[int] | None = None,
+        sample_rate: float | VectorLike[float] | None = None,
```
user27182, line 2933: Only `interpolation` can be used as an arg. This isn't great, I am thinking that doing something like `resample(2.0)` or `resample((2, 3, 4))` is more useful but this is ambiguous if we allow it to be a variable arg (is it a sampling rate or is it the resampled dimensions?), so `interpolation` seems like the only sensible kwarg to also allow as an arg. Maybe we could _only_ allow specifying the sampling rate, and keep dimensions as keyword only?

user27182: > Maybe we could _only_ allow specifying the sampling rate, and keep dimensions as keyword only?

This is now allowed in https://github.com/pyvista/pyvista/pull/7068/commits/c66de622fb26e1eb30e72cc325415f72bbdf16e8

pyvista/core/filters/image_data.py
```diff
@@ -2923,3 +2923,513 @@ def _validate_dimensional_operation(
...
+                new_origin = np.array(input_image.origin)
+                new_origin[~singleton_dims] += shift_new - shift_old
+
+                output_image.origin = new_origin
+            else:
+                # Compute spacing to match bounds of input and dimensions of output
+                bnds = input_image.bounds
+                size = np.array(
+                    (bnds.x_max - bnds.x_min, bnds.y_max - bnds.y_min, bnds.z_max - bnds.z_min)
+                )
```
user27182, line 3403:
```suggestion
                size = np.array(input_image.size)
```
Use size property from #7092

user27182: This change can also be made later as part of https://github.com/pyvista/pyvista/pull/7092 in case we want to merge this PR before that one

akaszynski (approved):

banesullivan (approved):

tkoyama010 (approved):

PR #7075, opened 2025-01-08
Use extent and image geometry for checking image equality

user27182:
### Overview

See https://github.com/pyvista/pyvista/pull/6879#issuecomment-2521771142


pyvista/core/dataobject.py
```diff
@@ -373,14 +373,17 @@ def __eq__(self: Self, other: object) -> bool:
...
         # these attrs use numpy.array_equal
-        equal_attrs = [
-            'verts',  # DataObject
-            'points',  # DataObject
-            'lines',  # DataObject
-            'faces',  # DataObject
-            'cells',  # UnstructuredGrid
-            'celltypes',
-        ]  # UnstructuredGrid
+        if isinstance(self, pyvista.ImageData):
```
tkoyama010, line 376: Would it be possible to overwrite it with the inherited `ImageData`?

user27182: Yes we can, but this test also tests equality of all the arrays too which we want to keep the same. So we would need to refactor this a bit into two parts , e.g. `_eq_structure` and `_eq_arrays`. Then `DataObject.__eq__` would check both `_eq_structure` and `_eq_arrays` by default internally, but `ImageData.__eq__` would override `_eq_structure` only.

IMO since it's a special case this kind of makes the equality check more opaque and harder to understand, because it won't be obvious which subclasses override this.

Keeping this in `DataObject` would be similar to how the `DataSetFilters` handle special cases all inside a single implementation.

tkoyama010 (approved):
LGTM

pyvista/core/dataobject.py
```diff
@@ -373,14 +373,17 @@ def __eq__(self: Self, other: object) -> bool:
...
+            equal_attrs = ['extent', 'index_to_physical_matrix']
```
jourdain, line 377: You should add `spacing` and `origin`

jourdain: comparing `extent` is just validating the cells part of the dataset.

user27182: This is implied by `index_to_physical_matrix` which is computed from `direction_matrix`, `spacing`, and `origin`. So my understanding is that this property alone fully defines the geometry.

https://github.com/Kitware/VTK/blob/b97c978f3f3752987db65344cd063c70f3729c9a/Common/DataModel/vtkImageData.cxx#L2353-L2371

We could check these all individually / separately if you think that's better?

user27182: See also this code from 3DSlicer where two images are considered equal if both the image extents and the geometries match.
https://github.com/Slicer/Slicer/blob/d2fca7d34fd0b29214385c08519434ca74ed1d08/Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx#L317-L327

And where `DoGeometriesMatch` is checking equality of the index to physical matrix, much like this PR does.
https://github.com/Slicer/Slicer/blob/d2fca7d34fd0b29214385c08519434ca74ed1d08/Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx#L681-L695

jourdain: `equal_attrs = ['extent', 'spacing', 'origin', 'direction_matrix']` is the input needed to define fully a vtkImageData.

PR #7077, opened 2025-01-09
Fix `points_to_cells` and `cells_to_points` extent value

user27182:
### Overview
These functions should generally return the same input when moving between representations. But the `extent` information is currently lost:

``` python
import pyvista as pv 
im = pv.ImageData()
im.extent = (0,1,0,1,0,1)
im = im.points_to_cells().cells_to_points()
print(im.extent)  # (0,1,0,1,0,1)

im2 = pv.ImageData()
im2.extent = (1,2,1,2,1,2)
im2 = im2.points_to_cells().cells_to_points()
print(im2.extent)  # (0,1,0,1,0,1), does not match original!
```

The issue stems from only considering `dimensions` in the implementation. But `extent` should be preferred over `dimensions` as it implies dimensions _and_ encodes an initial offset. This PR fixes this.


tkoyama010 (approved):
LGTM

PR #7078, opened 2025-01-09
Enable ruff security (S) linting

user27182:
### Overview

See https://docs.astral.sh/ruff/rules/#flake8-bandit-s

### DIFF
--- a/examples_trame/tests/test_trame.py
+++ b/examples_trame/tests/test_trame.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 from pathlib import Path
 import subprocess
+import sys
 
 import pytest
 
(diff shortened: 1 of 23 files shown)


examples_trame/tests/test_trame.py
```diff
@@ -27,7 +28,7 @@ def collect_example_files():
 
 @pytest.mark.parametrize('test_file', collect_example_files())
 def test_serve(test_file):
-    returncode = subprocess.run(
-        ['python', EXAMPLES_DIR / test_file, '--serve', '--timeout', '1', '--port', '0'],
+    returncode = subprocess.run(  # noqa: S603
```
user27182, line 31: With this linting option the errors provide helpful hints to make the code safer. But often I am finding the safest recommended option is to use `subprocess.run`, which has some inherent risk and cannot be made safer, so no matter what you are always left with this error
``` python
 S603 `subprocess` call: check for execution of untrusted input
```
which must then always be ignored. This is helpful to at least prompt the user to check, but is maybe not clear that it _must_ be ignored. We could alternatively ignore `S103` globally, which would be less annoying.

PR #7079, opened 2025-01-09
Optimize conversion to `meshio` format and add new `pyvista.to_meshio()` function 

keurfonluu:
### Overview

Small improvements for ``meshio``:

 - Add function ``pyvista.to_meshio()`` as its representation of cells is, sometimes, more convenient to use. The old code in ``pyvista_save_meshio()`` has been extracted and optimized along the way (e.g., no deep copies for polyhedral cells, no loop for mesh with single cell type).
 - Convert cell sets to cell data, for instance useful when using ``pygmsh`` to visualize groups, e.g.:

```python
import pygmsh
import pyvista as pv

with pygmsh.occ.Geometry() as geo:
    disk1 = geo.add_disk([0.0, 0.0, 0.0], 1.0)
    disk2 = geo.add_disk([2.0, 0.0, 0.0], 1.0)
    geo.add_physical(disk1, "Disk 1")
    geo.add_physical(disk2, "Disk 2")
    mesh = geo.generate_mesh()

pv.from_meshio(mesh).plot(show_edges=True)
```
image


pyvista/core/utilities/fileio.py
```diff
@@ -993,6 +997,140 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
...
+            polyhedral_cell_faces.append(faces)
+
+    # Single cell type (except POLYGON and POLYHEDRON)
+    if vtk_celltypes.min() == vtk_celltypes.max() and vtk_celltypes[0] not in {7, 42}:
+        vtk_celltype = vtk_celltypes[0]
+        cells = connectivity.reshape((mesh.n_cells, connectivity.size // mesh.n_cells))
+
+        if vtk_celltype == 8:
+            cells = cells[:, [0, 1, 3, 2]]
+            celltype = 'quad'
```
user27182, line 1079:
```suggestion
        if vtk_celltype == pv.CellType.PIXEL:
```
Can we use the [CellType](https://dev.pyvista.org/api/utilities/_autosummary/pyvista.celltype#pyvista.CellType) enum here instead?

pyvista/core/utilities/fileio.py
```diff
@@ -993,6 +997,140 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
...
+
+            while len(faces) < n_faces:
+                n_vertices = polyhedral_cells[i]
+                faces.append(polyhedral_cells[i + 1 : i + 1 + n_vertices])
+                i += n_vertices + 1
+
+            polyhedral_cell_faces.append(faces)
+
+    # Single cell type (except POLYGON and POLYHEDRON)
+    if vtk_celltypes.min() == vtk_celltypes.max() and vtk_celltypes[0] not in {7, 42}:
```
user27182, line 1073: Use CellType enum
```suggestion
    if vtk_celltypes.min() == vtk_celltypes.max() and vtk_celltypes[0] not in {pv.CellType.POLYGON, pv.CellType.POLYHEDRON}:
```

pyvista/core/utilities/fileio.py
```diff
@@ -993,6 +997,140 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
...
+        vtk_celltype = vtk_celltypes[0]
+        cells = connectivity.reshape((mesh.n_cells, connectivity.size // mesh.n_cells))
+
+        if vtk_celltype == 8:
+            cells = cells[:, [0, 1, 3, 2]]
+            celltype = 'quad'
+
+        elif vtk_celltype == 11:
+            cells = cells[:, [0, 1, 3, 2, 4, 5, 7, 6]]
+            celltype = 'hexahedron'
```
user27182, line 1083:
```suggestion
        elif vtk_celltype == pv.CellType.VOXEL:
            cells = cells[:, [0, 1, 3, 2, 4, 5, 7, 6]]
            celltype = 'hexahedron'
```

pyvista/core/utilities/fileio.py
```diff
@@ -993,6 +997,140 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
...
+    # Mixed cell types
+    else:
+        cells = []
+        offset = mesh.offset
+        polyhedron_count = 0
+
+        for i1, i2, vtk_celltype in zip(offset[:-1], offset[1:], vtk_celltypes):
+            cell = connectivity[i1:i2]
+
+            if vtk_celltype == 42:
```
user27182, line 1099:
```suggestion
            if vtk_celltype == pv.CellType.POLYHEDRON:
```

pyvista/core/utilities/fileio.py
```diff
@@ -993,6 +997,140 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
...
+                celltype = f'polyhedron{len(cell)}'
+                cell = polyhedral_cell_faces[polyhedron_count]
+                polyhedron_count += 1
+
+            else:
+                celltype = (
+                    f'polygon{len(cell)}' if vtk_celltype == 7 else vtk_to_meshio_type[vtk_celltype]
```
user27182, line 1106:
```suggestion
                    f'polygon{len(cell)}' if vtk_celltype == pv.CellType.POLYGON else vtk_to_meshio_type[vtk_celltype]
```

pyvista/core/utilities/fileio.py
```diff
@@ -1058,87 +1196,11 @@ def save_meshio(
...
-    meshio.write_points_cells(
-        filename=filename,
-        points=np.array(mesh.points),
-        cells=cells,
-        point_data=point_data,
-        cell_data=cell_data,
-        file_format=file_format,
-        **kwargs,
-    )
+    to_meshio(mesh).write(filename, file_format=file_format, **kwargs)
```
user27182, line 1203: This looks like `read_meshio` should be deprecated

user27182 (changes requested):
Thanks for the updates! Suggest using enum where possible instead of integers and deprecating `save_meshio`. Plus it would be good to have test coverage for some of the (new?) code.

pyvista/core/utilities/fileio.py
user27182: Whoops this suggestion accidentally removed the code block, sorry!

pyvista/core/utilities/fileio.py
user27182: Nevermind, I think I understand now, `to_meshio` creates an instance, `save_meshio` saves to file. This is OK, no deprecation needed.

Since there is no fileio per-se with `to_meshio`, I wonder if it should be a `DataSetFilter` instead?

keurfonluu: A method ``pyvista.DataSet.to_meshio()`` sounds good to me as well. Should I move the function to ``pyvista.DataSetFilters``?

user27182: Maybe hold off to get the opinion of other reviewers before making this change? There may be a deliberate effort or undocumented "policy" to keep 3rd party mesh conversion stuff separate for some reason. 

Though, I think integrating 3rd party meshes directly would be a good thing. A more complete API to support 3rd party meshes might look like:

- Allow `pv.PolyData(mesh)` and add a new `DataSetFilters.to_trimesh()` method for `trimesh.TriMesh` objects
- Allow `pv.UnstructuredGrid(mesh)` and add new `DataSetFilters.to_trimesh()` method for `meshio.Mesh` objects

user27182: [pv.from_meshio](https://dev.pyvista.org/api/utilities/_autosummary/pyvista.from_meshio.html#pyvista.from_meshio) is a thing, so it makes sense to keep `to_meshio` in `fileio.py` for now.

Better integrations using pyvista dataset constructors or filters can be a separate effort / PR

pyvista/core/utilities/__init__.py
```diff
@@ -50,6 +50,7 @@
 from .fileio import save_pickle as save_pickle
 from .fileio import set_pickle_format as set_pickle_format
 from .fileio import set_vtkwriter_mode as set_vtkwriter_mode
+from .fileio import to_meshio as to_meshio
```
user27182, line 53: New function should also be added to docs via `doc/source/api/utilities/utilities.rst`

pyvista/core/utilities/fileio.py
```diff
@@ -993,6 +997,145 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
...
+    Returns
+    -------
+    meshio.Mesh
+        A mesh instance from the ``meshio`` library.
+
+    Raises
+    ------
+    ImportError
+        If the meshio package is not installed.
+
```
user27182, line 1017: Add cross-refs
```suggestion

    See Also
    --------
    from_meshio, read_meshio, save_meshio, :func:`~pyvista.wrap`

```

Would also be good to add a similar cross-refs to these other meshio methods in `fileio.py` but it's OK if not

user27182 (commented):
Minor changes to documentation

pyvista/core/utilities/fileio.py
```diff
@@ -1058,87 +1207,11 @@ def save_meshio(
...
-    meshio.write_points_cells(
-        filename=filename,
-        points=np.array(mesh.points),
-        cells=cells,
-        point_data=point_data,
-        cell_data=cell_data,
-        file_format=file_format,
-        **kwargs,
-    )
+    to_meshio(mesh).write(filename, file_format=file_format, **kwargs)
```
user27182, line 1214: The kwargs were originally passed to ``meshio.write_points_cells``. Are they still valid for `meshio.Mesh.write`?

If not we should probably add a ``.. versionchanged:: 0.45`` directive to communicate the change. Either way the docs for `save_meshio` should be updated to say the kwargs are passed to ``meshio.Mesh.write``

keurfonluu: ``kwargs`` are the same. ``meshio.Mesh(points, cells, point_data, cell_data).write(**kwargs)`` is the same as ``meshio.write_points_cells(points, cells, point_data, cell_data, **kwargs)``.

pyvista/core/utilities/fileio.py
```diff
@@ -993,6 +997,151 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
...
+    # Generate polyhedral cell faces if any
+    polyhedral_cells = pyvista.convert_array(mesh.GetFaces())
+
+    if polyhedral_cells is not None:
+        locations = pyvista.convert_array(mesh.GetFaceLocations())
+        polyhedral_cell_faces = []
+
+        for location in locations:
+            if location == -1:
+                continue
```
user27182, line 1065: This does not have test coverage. I guess this covers this case from the VTK code? Is that right?
https://github.com/Kitware/VTK/blob/b97c978f3f3752987db65344cd063c70f3729c9a/Common/DataModel/vtkUnstructuredGrid.cxx#L840-L844

user27182 (approved):
LGTM, missing coverage for one branch but it seems niche and is _probably_ ok based on VTK code

banesullivan:
Hey @keurfonluu, I'm running into some issues with this code in #6879 as the polyhedral faces storage layout has changed in VTK 9.4 reference these release notes: https://docs.vtk.org/en/latest/release_details/9.4/vtk-polyhedral-cells-storage.html

I'm trying to get this right but getting awfully confused. Could you help update the new `to_meshio` method so that it is compatible with VTK's latest changes to the polyhedral cell storage?

keurfonluu:
Hi @banesullivan,

> Another case are the GetFaces and GetFaceLocations methods. They are still available but an internal cache with the old internal layout is used. It means that they should be considered as read only methods and not be relied on to change internal data of vtkUnstructuredGrid. The caching process may impact a bit performance.

My understanding is that the old API should still work, but it relies on internal copies thus impacting the performance. I tried the current development version of PyVista with VTK 9.4, and the function to_meshio() seems to work correctly with a polyhedral mesh.

```python
import numpy as np
import pyvista as pv

points = np.array(
    [
        [0.3568221, -0.49112344, 0.79465446],
        [-0.3568221, -0.49112344, 0.79465446],
        [0.3568221, 0.49112344, -0.79465446],
        [-0.3568221, 0.49112344, -0.79465446],
        [0.0, 0.98224693, 0.18759243],
        [0.0, 0.60706198, 0.79465446],
        [0.0, -0.60706198, -0.79465446],
        [0.0, -0.98224693, -0.18759243],
        [0.93417233, 0.30353101, 0.18759247],
        [0.93417233, -0.30353101, -0.18759247],
        [-0.93417233, 0.30353101, 0.18759247],
        [-0.93417233, -0.30353101, -0.18759247],
        [-0.57735026, 0.18759249, 0.79465446],
        [0.57735026, -0.79465446, 0.18759249],
        [-0.57735026, -0.18759249, -0.79465446],
        [0.57735026, 0.79465446, -0.18759249],
        [0.57735026, 0.18759249, 0.79465446],
        [-0.57735026, 0.79465446, -0.18759249],
        [-0.57735026, -0.79465446, 0.18759249],
        [0.57735026, -0.18759249, -0.79465446],
        [0.3568221, 0.49112344, -1.0],
        [0.57735026, -0.18759249, -1.0],
        [0.0, -0.60706198, -1.0],
        [-0.57735026, -0.18759249, -1.0],
        [-0.3568221, 0.49112344, -1.0],
        [0.3568221, -0.49112344, 1.0],
        [0.57735026, 0.18759249, 1.0],
        [0.0, 0.60706198, 1.0],
        [-0.57735026, 0.18759249, 1.0],
        [-0.3568221, -0.49112344, 1.0],
    ]
)
cells = np.array(
    [
        73, 12,  5,  0, 16,  5, 12,  1,  5,  1, 18,  7, 13,  0,  5,  2, 19,
        6, 14,  3,  5,  3, 17,  4, 15,  2,  5,  4,  5, 16,  8, 15,  5,  5,
        4, 17, 10, 12,  5,  6,  7, 18, 11, 14,  5,  7,  6, 19,  9, 13,  5,
        8, 16,  0, 13,  9,  5,  9, 19,  2, 15,  8,  5, 10, 17,  3, 14, 11,
        5, 11, 18,  1, 12, 10, 38,  7,  5,  2, 19,  6, 14,  3,  4, 20, 21,
        19,  2,  4, 21, 22,  6, 19,  4, 22, 23, 14,  6,  4, 23, 24,  3, 14,
        4, 24, 20,  2,  3,  5, 20, 21, 22, 23, 24, 38,  7,  5,  0, 16,  5,
        12,  1,  4,  0, 16, 26, 25,  4, 16,  5, 27, 26,  4,  5, 12, 28, 27,
        4, 12,  1, 29, 28,  4,  1,  0, 25, 29,  5, 25, 26, 27, 28, 29,
    ]
)
celltypes = [pv.CellType.POLYHEDRON] * 3

mesh = pv.UnstructuredGrid(cells, celltypes, points)
pv.to_meshio(mesh).write("mesh.vtu")
mesh2 = pv.read("mesh.vtu")
assert np.allclose(mesh.points, mesh2.points)
assert np.allclose(mesh.cell_connectivity, mesh2.cell_connectivity)
assert np.allclose(mesh.compute_cell_sizes()["Volume"], mesh2.compute_cell_sizes()["Volume"])
```

Or do you mean that you want to use ``GetPolyhedronFaces()`` and ``GetPolyhedronFaceLocations()`` instead? Here is a quick and dirty snippet:

```python
import numpy as np
import vtk

def to_numpy(cell_array):
    cell_array.InitTraversal()
    id_list = vtk.vtkIdList()
    cells = []
    
    while cell_array.GetNextCell(id_list):
        cell = [id_list.GetId(i) for i in range(id_list.GetNumberOfIds())]
        cells.append(cell)
    
    return cells

faces = to_numpy(mesh.GetPolyhedronFaces())
locations = to_numpy(mesh.GetPolyhedronFaceLocations())

polyhedral_cell_faces = [[faces[face] for face in cell] for cell in locations]
```

PR #7059, opened 2025-01-05
Add new standalone `anatomical-groups` example

user27182:
### Overview

Continuation from #5966

Edit: the groups in this example are loosely based on this image:
https://github.com/wasserth/TotalSegmentator/blob/master/resources/imgs/overview_classes_v2.png

PR #7083, opened 2025-01-11
Update formatting for anatomical groups example

user27182:
### Overview

Follow-up to #7059 .

- Commas are added to the printed color mapping to make it valid python code (so it can be easily copy/pasted and re-used).
- Some cross refs are not working (they are just comments) and are moved elsewhere so that it's valid RST and properly linked


tkoyama010 (approved):
LGTM

PR #7090, opened 2025-01-15
Remove normal flipping in logo

banesullivan:
https://github.com/pyvista/pyvista/pull/6062 caused the PyVista logo to plot incorrectly as the cell normals were pointing inwards. I suspect that the `flip_normals()` call originally in the logo script was a bandaid on a wider issue with cell normals which #6062 fixed. This PR removes the flip_normals call in that script.

@user27182, would you be able to take a look at this since you submitted #6062 and this is related to that?

| `main` | this branch |
|---|---|
|  screenshot | pyvista_high_res |


user27182:
I can look into it. The logo looks fine in the current dev docs (built from main), so not sure if I will be able to reproduce this locally. Are you able to post a report here to list dependencies?

user27182:
> I suspect that the `flip_normals()` call originally in the logo script was a bandaid on a wider issue with cell normals which #6062 fixed.

Also note that `flip_normals()`  doesn't actually flip normals, it reverses cell ordering... See the open PR #6966 for a fix (it's ready for review, just waiting on approval)

banesullivan:
> The logo looks fine in the current dev docs (built from main), so not sure if I will be able to reproduce this locally. Are you able to post a report here to list dependencies?
> 
> EDIT: See https://dev.pyvista.org/api/examples/_autosummary/pyvista.examples.examples.load_logo.html

This is actually the wrong example. `load_logo` is loading a pre-saved PNG of the logo asset generated long ago and plotting it as `ImageData`. We need to look at `plot_logo` in the `demos` module which isn't built in the documentation

banesullivan:
> Also note that flip_normals() doesn't actually flip normals, it reverses cell ordering... See the open PR https://github.com/pyvista/pyvista/pull/6966 for a fix (it's ready for review, just waiting on approval)

Oh, huh. this indeed may be the real issue

user27182 (dismissed):

MatthewFlamm:
Something like this ought to have an image regression test.  Should the demos be in the docs?

user27182:
> Something like this ought to have an image regression test. Should the demos be in the docs?

At the very least for the logo we could add this to `tests/plotting/test_plotting.py` for regression testing
``` python
def test_plot_logo():
    logo_plotter = plot_logo(just_return_plotter=True)
    logo_plotter.show()
```

And, if we want to make sure that the `logofile` loaded with `load_logo` actually matches the logo from `plot_logo`, we could add something like this:

``` python
@pytest.mark.skip_plotting
def test_plot_logo_matches_load_logo(tmp_path):
    loaded_logo = pv.examples.load_logo()
    size = loaded_logo.dimensions[:2]

    screenshot_file = Path(tmp_path / 'screenshot.png')
    plot_logo(screenshot=screenshot_file, window_size=size)
    plotted_logo = pv.read(screenshot_file)
    pv.compare_images(plotted_logo, loaded_logo, threshold=500)
```

Though this test fails at the moment because `loaded_logo` has a different resolution ``(1920, 718)`` than the default plotted logo ``(960, 400)``. This difference can be sort of fixed by using `window_size`, but this adds whitespace around the logo. So, not really sure how to fully reproduce the image from `logofile`.

MatthewFlamm:
> And, if we want to make sure that the `logofile` loaded with `load_logo` actually matches the logo from `plot_logo`, we could add something like this:

This would also be nice, but not critical IMO.  We can manually inspect for differences and then any regression would be caught by the test you proposed.

MatthewFlamm (approved):

tkoyama010 (approved):

PR #7091, opened 2025-01-16
Fix validate dimensionality with integer values

user27182:
### Overview

This works to convert 2D point data into 3D cell data:
``` python
from pyvista import examples
mesh = examples.load_logo()
mesh.points_to_cells(dimensionality='3D')
```

But this fails unexpectedly:
``` python
mesh.points_to_cells(dimensionality=3) 
# TypeError: Array has incorrect dtype of 'int64'. The dtype must be a subtype of <class 'bool'>.
```

This PR fixes this.

cc @GuillaumeBroggi


pyvista/core/filters/image_data.py
```diff
@@ -2848,7 +2848,7 @@ def _validate_dimensional_operation(
         # Build an array of the operation size
         operation_size = _validation.validate_array3(operation_size, reshape=True, broadcast=True)
 
-        if not isinstance(operation_mask, str):
+        if not isinstance(operation_mask, str) and operation_mask not in [0, 1, 2, 3]:
```
tkoyama010, line 2851: This might come down to personal preference. If it doesn't resonate with you, feel free to disregard it.
```suggestion
        if not isinstance(operation_mask, str) and operation_mask not in range(4):
```

tkoyama010 (changes requested):

pyvista/core/filters/image_data.py
user27182: This change is more compact but I have a slight preference for explicitly listing out the values in this case

tkoyama010 (approved):
LGTM

PR #7092, opened 2025-01-16
Add `bounds_size` property to datasets and actors

user27182:
### Overview

Computing the difference between the x-y-z max/min bounds is a common pattern. This PR adds a new ~`size`~ `bounds_size` property to compute this more easily.

~The `Label` class currently defines its own `size` property which is inconsistent with the new `size` property for Prop3D classes. This PR also deprecates this property so that in a future version, `Label.size` will be consistent with `Actor.size`~


tkoyama010 (dismissed):
LGTM

darikg:
Not saying this is better but could alternatively put the `size` method directly on the `BoundsTuple` class, as well as `@property`s `x_size`, `y_size`, `z_size`.

user27182:
> Not saying this is better but could alternatively put the `size` method directly on the `BoundsTuple` class, as well as `@property`s `x_size`, `y_size`, `z_size`.

Is the intent to avoid indexing like `mesh.size[0]`? e.g. use `mesh.x_size` instead?

In this regard doing `mesh.size[0]` is not too different from indexing other 3-tuple properties like `PlaneSource.position[0]`, and my experience with classes like `AxesActor` is that using `axes.x_label` `axes.y_label` `axes.z_label` separately every time is cumbersome. So in that sense my preference is for having a single `size` property that returns a tuple.

We could separately add a new `XYZTuple` class to any/all methods that return a 3-tuple like this. This would enable doing `mesh.size.x` instead of `mesh.size[0]`

tests/core/test_geometric_sources.py
```diff
@@ -196,8 +191,7 @@ def test_text3d_source_parameters(string, center, height, width, depth, normal):
         # For width and height, create an unscaled version for reference
         src_not_scaled = pv.Text3DSource(string=string, center=center)
         out_not_scaled = src_not_scaled.output
-        bnds = out_not_scaled.bounds
-        unscaled_width, unscaled_height = bnds.x_max - bnds.x_min, bnds.y_max - bnds.y_min
+        unscaled_width, unscaled_height = out_not_scaled.size[0:2]
```
user27182, line 194:
```suggestion
        unscaled_width, unscaled_height, _ = out_not_scaled.size
```

tests/plotting/test_actor.py
```diff
@@ -218,9 +218,20 @@ def test_actor_length(klass, actor, dummy_actor):
     initial_length = 2**0.5  # sqrt(2)
     scale_factor = 2
 
-    assert actor.length == initial_length
+    assert np.allclose(actor.length, initial_length)
```
user27182, line 221: Scalar comparison (note: this change is not related to the new size parameter)
```suggestion
    assert np.isclose(actor.length, initial_length)
```

pyvista/core/utilities/points.py
```diff
@@ -358,9 +358,7 @@ def fit_plane_to_points(
     )
 
     # Fit plane to xyz-aligned mesh
-    aligned_bnds = aligned.bounds
-    i_size = aligned_bnds.x_max - aligned_bnds.x_min
-    j_size = aligned_bnds.y_max - aligned_bnds.y_min
+    i_size, j_size = aligned.size[0:2]
```
user27182, line 361:
```suggestion
    i_size, j_size, _ = aligned.size
```

pyvista/plotting/text.py
```diff
@@ -404,12 +407,33 @@ def size(self) -> int:  # numpydoc ignore=RT01
...
+        if version_info >= (0, 48):  # pragma: no cover
+            raise RuntimeError('Convert this deprecation warning into an error.')
+        if version_info >= (0, 49):  # pragma: no cover
+            raise RuntimeError('Remove this property.')
+
+        msg = (
+            'The behavior of `size` will change in a future version. It currently returns a float with\n'
+            "the label's font size, but will return a tuple with its physical geometric size in the future.\n"
+            'Use `font_size` instead.'
+        )
```
user27182, line 419: Should probably add note to the docstring too. The use of `size` in the `Label` constructor may also need to be deprecated.

pyvista/core/dataset.py
```diff
@@ -1232,6 +1232,31 @@ def length(self: Self) -> float:
         """
         return self.GetLength()
 
+    @property
+    def size(self: Self) -> tuple[float, float, float]:
```
user27182, line 1236: This term is heavily used, e.g. the `Label.size` property deprecated in this PR, but also many others:
https://dev.pyvista.org/search.html?q=size

This makes the term ambiguous, maybe call it `bounds_size` instead? Or maybe `physical_size`... or something else? With this change we could also remove the deprecation of `Label.size`.

```suggestion
    def bounds_size(self: Self) -> tuple[float, float, float]:
```

darikg:
> > Not saying this is better but could alternatively put the `size` method directly on the `BoundsTuple` class, as well as `@property`s `x_size`, `y_size`, `z_size`.
> 
> Is the intent to avoid indexing like `mesh.size[0]`? e.g. use `mesh.x_size` instead?

More that you'd write `mesh.bounds.size` instead of `mesh.size` which makes the `BoundsTuple` a little more useful on its own, and also potentially avoid some naming conflicts by having the `size` method belong to the bounds instead of the mesh.

user27182:
> More that you'd write `mesh.bounds.size` instead of `mesh.size` which makes the `BoundsTuple` a little more useful on its own, and also potentially avoid some naming conflicts by having the `size` method belong to the bounds instead of the mesh.

Thanks for clarifying, I didn't catch this. I like this and would be cleaner. The issue I think would be that with named tuples, the field names are reserved for indexing (`.x_min` is the same as `[0]`). Having a `size` property would violate this since it's not indexed. Not sure if it's actually prohibited by the class though. I think methods are supported though, so we could have `get_size()` instead.

darikg:
> Thanks for clarifying, I didn't catch this. I like this and would be cleaner. The issue I think would be that with named tuples, the field names are reserved for indexing (`.x_min` is the same as `[0]`). Having a `size` property would violate this since it's not indexed. Not sure if it's actually prohibited by the class though. I think methods are supported though, so we could have `get_size()` instead.

I believe `NamedTuple`s can have dynamic `@property`s just fine, but yeah a method makes sense too

user27182:
`bounds.size` is now used instead of `mesh.size` or `actor.size`. This is a much cleaner change without any deprecations.
There are a few downsides but I think they're not deal-breakers:

- `size` is not indexable, unlike all other BoundsTuple properties (`x_min`, `x_max`, etc), which is inconsistent
- `size` is a bounds property, but `length` is a mesh/actor property. Arguably, `length` should also be a property of the bounds just like `size` is. So there is some inconsistency there too.

user27182:
This PR is once again ready, just need to get consensus on the naming and parent class(es) for the new property:

1. Use `DataSet.size` and `Prop3D.size`. This was originally proposed but is maybe ambiguous and requires deprecations 
2. Use` DataSet.bounds_size`, this is like (1) but more explicit and does not require deprecations
3. Use `bounds.size`, this is the current implementation, but also has drawbacks (see above)

user27182:
> This PR is once again ready, just need to get consensus on the naming and parent class(es) for the new property:
> 
> 1. Use `DataSet.size` and `Prop3D.size`. This was originally proposed but is maybe ambiguous and requires deprecations

I have reverted to this option. Despite the deprecations I think it's better to have it as a dataset property to be consistent with other properties like `length`.

user27182:
For some more context, Paraview also computes the `size` as the delta between bounds values:

size_imagedata
size_polydata

user27182:
> Attention: Patch coverage is 90.90909% with 4 lines in your changes missing coverage. Please review.

Will need to add tests for `Plotter.size` and `Prop3D.size`

pyvista/core/dataset.py
tkoyama010: Since the values are obtained from bounds internally, the name `bounds_size` feels more natural.

tkoyama010 (changes requested):

pyvista/core/dataset.py
```diff
@@ -1242,6 +1242,31 @@ def length(self: Self) -> float:
...
+        by default.
+
+        >>> import pyvista as pv
+        >>> mesh = pv.Cube()
+        >>> mesh.size
+        (1.0, 1.0, 1.0)
+
+        """
+        bnds = self.bounds
+        return bnds.x_max - bnds.x_min, bnds.y_max - bnds.y_min, bnds.z_max - bnds.z_min
```
tkoyama010:
```suggestion
        bounds = self.bounds
        return bounds.x_max - bounds.x_min, bounds.y_max - bounds.y_min, bounds.z_max - bounds.z_min
```

pyvista/plotting/plotter.py
```diff
@@ -1788,6 +1788,32 @@ def bounds(self) -> BoundsTuple:  # numpydoc ignore=RT01
...
+
+        >>> import pyvista as pv
+        >>> pl = pv.Plotter()
+        >>> _ = pl.add_mesh(pv.Cube())
+        >>> pl.size
+        (1.0, 1.0, 1.0)
+
+        """
+        bnds = self.bounds
+        return bnds.x_max - bnds.x_min, bnds.y_max - bnds.y_min, bnds.z_max - bnds.z_min
```
tkoyama010:
```suggestion
        bounds = self.bounds
        return bounds.x_max - bounds.x_min, bounds.y_max - bounds.y_min, bounds.z_max - bounds.z_min
```

pyvista/plotting/prop3d.py
```diff
@@ -475,6 +475,32 @@ def length(self) -> float:  # numpydoc ignore=RT01
...
+
+        >>> import pyvista as pv
+        >>> pl = pv.Plotter()
+        >>> actor = pl.add_mesh(pv.Cube())
+        >>> actor.size
+        (1.0, 1.0, 1.0)
+
+        """
+        bnds = self.bounds
+        return bnds.x_max - bnds.x_min, bnds.y_max - bnds.y_min, bnds.z_max - bnds.z_min
```
tkoyama010:
```suggestion
        bounds = self.bounds
        return bounds.x_max - bounds.x_min, bounds.y_max - bounds.y_min, bounds.z_max - bounds.z_min
```

pyvista/plotting/prop3d.py
```diff
@@ -681,3 +707,10 @@ def length(self) -> float:  # numpydoc ignore=RT01
         return np.linalg.norm(
             (bnds.x_max - bnds.x_min, bnds.y_max - bnds.y_min, bnds.z_max - bnds.z_min)
         ).tolist()
+
+    @property
+    @wraps(Prop3D.size.fget)  # type: ignore[attr-defined]
+    def size(self) -> tuple[float, float, float]:  # numpydoc ignore=RT01
+        """Wrap :class:`pyvista.Prop3D.size."""
+        bnds = self.bounds
+        return bnds.x_max - bnds.x_min, bnds.y_max - bnds.y_min, bnds.z_max - bnds.z_min
```
tkoyama010:
```suggestion
        bounds = self.bounds
        return bounds.x_max - bounds.x_min, bounds.y_max - bounds.y_min, bounds.z_max - bounds.z_min
```

tkoyama010 (approved):
LGTM

PR #7100, opened 2025-01-18
Add `reshaped_point_data` and `reshaped_cell_data` properties for auto-reshaping data

### LINKED ISSUE #4862: ImageData DataArrays Flipped

Plotting or saving ImageData changes how the data is stored. For example, if I plot a YZ slice of my dataset, pyvista transposes the data array such that it no longer matches the original ordering. I saw the same behavior using a RectilinearGrid.

```python
import pyvista
import numpy as np

def main():
    # create grid
    dimension = (1,5,3)
    origin=(0,0,0)
    spacing=(0,1,1)
    grid = pyvista.ImageData(
        dimensions=dimension,
        spacing=spacing,
        origin=origin,
    )
    # initialize dummy data - vary in Y, constant in Z
    arr = np.expand_dims(np.array([1,2,3,4,5]),axis=1)
    V = np.zeros(dimension)
    V[:] = np.tile(arr, dimension[2])
    grid["V"] = V.ravel()
    # create meshgrid for matplotlib
    y = np.linspace(0,spacing[1]*(dimension[1]-1), dimension[1])
    z = np.linspace(0,spacing[2]*(dimension[2]-1), dimension[2])
    Y,Z = np.meshgrid(y,z,indexing="ij")
    # plot via matplotlib
    plt.figure()
    plt.contourf(Y,Z,grid["V"].reshape(dimension[1],dimension[2]))
    plt.axis("equal")
    # plt.tight_layout()
    plt.xlim([0,4])
    plt.xlabel("y")
    plt.ylim([0,2])
    plt.ylabel("z")
    plt.title("Matplotlib")
    plt.savefig("grid_imagedata_mplotlib.png")
    # save image data mesh
    grid.save("grid_imagedata.vti")

if __name__=="__main__":
    main()
```

### System Information

```shell
  Date: Thu Sep 07 10:37:19 2023 CDT

                OS : Linux
            CPU(s) : 12
           Machine : x86_64
      Architecture : 64bit
       Environment : Python
       GPU Details : error
  MathText Support : False

  Python 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:17) [GCC
  12.2.0]

           pyvista : 0.41.1
               vtk : 9.2.6
             numpy : 1.25.2
        matplotlib : 3.7.2
            scooby : 0.7.2
             pooch : v1.7.0
             scipy : 1.11.2
            meshio : 5.3.4
```

### Screenshots

Expected: plotting variable "V" as a contourf plot in matplotlib. The data is varying in Y and is constant in Z.
matplotlib
Actual: plotting variable "V" in Paraview (saw same behavior with pyvista.Plotter). The Y and Z position data is correct, but dataset "V" has been transposed or otherwise jumbled.
pyvista_render_paraview

### LINKED ISSUE #5102: Unexpected (?) behavior with multidimensional point_data
Hello,

In a program, I need to define a point_data with more than 2 dimensions to a PolyData. Here is a simple example where I want to associate a 3x3 matrix to each point :
```python
import pyvista

mesh = pyvista.Sphere()
n_points = mesh.n_points

data = np.random.rand(n_points, 3, 3)
mesh["data"] = data
```

This code doesn't raise any error but it seems that my data was silently reshaped to be 2-dimensional
```python
print(mesh["data"].shape)
>>> (842, 9)
```

After a few searches, it seems that vtk is not happy at all with more than 2-dimensional arrays : 
```python
from vtkmodules.util.numpy_support import numpy_to_vtk
import numpy as np

data = np.random.rand(10, 3, 3)
data_vtk = numpy_to_vtk(data, deep=True)
>>> AssertionError: Only arrays of dimensionality 2 or lower are allowed!
```

With that in mind, I understand perfectly the choice to reshape point_data in pyvista. However, I think that it would be good to add a warning when doing so to inform the user. What are you thoughts about it ?

Thanks in advance,

### LINKED ISSUE #7098: Better handling of higher dimensionality data

While the dimensionality of appended data to `PolyData` and `UnstructuredGrid` fundamentally currently limited to 2D arrays, I feel there could be features in place that help support the storing and loading of higher dimensionality data. 

This could include a helper that records and stores the original shape of the data along side the flattened array, allowing it to fetch and received when saving data, and then correctly reshaping when loading it out of the VTK format.

Additionally the refinement of the data handling within the backend can allow for more consistent behaviour.

Would appreciate thoughts and some clarification on current behaviour.

3D arrays are currently quietly flattened behind the scene, [link](https://github.com/pyvista/pyvista/blob/5deb52aad5ed8c7271867d28e673734105416f3b/pyvista/core/datasetattributes.py#L859)

And larger arrays currently throw VTK error within the numpy_to_vtk function [link](https://github.com/Kitware/VTK/blob/041af59c711ad95b9cc4355cc3d8966de19da27f/Wrapping/Python/vtkmodules/util/numpy_support.py#L135)

user27182:
### Overview

Fix https://github.com/pyvista/pyvista/issues/4862 fix https://github.com/pyvista/pyvista/issues/5102 fix #7098

This PR adds `DataSet.reshaped_point_data` and `DataSet.reshaped_cell_data` for automatically reshaping point and cell data. These properties are similar in spirit to the `PolyData.regular_faces` API (which makes accessing `faces` easier) and aims to make accessing multi-dimensional arrays easier.

The API is fully opt-in, and no reshaping is done by default. Both a flattening method (for setting arrays) and a reshaping method (for getting arrays) must first be provided since reshaping is application-specific and  the array order (c or fortran) may differ. Whereas `point_data` and `cell_data` should expect _no_ array reshaping to be done, the `reshaped_point_data` and `reshaped_cell_data` properties are exclusively used when reshaping _is_ desired/expected.

``` python
import pyvista as pv
import numpy as np

shape = (3,3,3)
data_in = np.zeros(shape=shape)
data_in.shape  # data in has desired shape (3, 3, 3)
image = pv.ImageData(dimensions=shape)

# Must define auto-reshaping behavior
image.data_flattening_method = 'ravel_c'
image.data_reshaping_method = 'dimensions'

# No longer need to reshape arrays manually
image.reshaped_point_data['data'] = data_in
data_out = image.reshaped_point_data['data']

data_out.shape # data out has desired shape (3, 3, 3)
```

The reshaping methods in general can be any custom function, though a predefined preset method can also be chosen for convenience. The plan is to include a few simple preset reshaping methods at first, and more complex reshaping methods be added with follow-up PRs.


banesullivan:
Sorry there was never any follow up on this @user27182. Should we revive this and try to close out those issues?

Issue #7096, opened 2025-01-16
Reduce doc build size


user27182:
### Describe what maintenance you would like added.

The doc build artefact is huge at 4GB+. This is mainly caused by the interactive plot feature from https://github.com/pyvista/pyvista/pull/4938 which included a bunch of very large `vtksz` files (docs build size used to be 500MB before that IIRC), but this issue is also compounded by plotting very large datasets, e.g. the `whole_body_ct` datasets are approx 60MB each on disk.

This is an issue for collaborators since I am personally unable to download such a large artefact at all (I think my ISP throttles github downloads) and there is demand from others for it to be smaller, e.g. https://github.com/pyvista/pyvista/pull/7095#issuecomment-2596929153. This also contributes to slow build times and may be reaching a github runner size limit (https://github.com/pyvista/pyvista/issues/6515). 

There are a few options.

1. Disable generating all interactive plots. This was tried in #6526 and seemed to work.
2. Disable interactive plots on a per-plot basis using the plot directive
3. Create a second build artefact with all the `vtksz` interactive plot files omitted
4. Do not run doctests during build, this is tried in #7003
5. Use filters like `resample` from #7068 and [decimate](https://dev.pyvista.org/api/core/_autosummary/pyvista.polydatafilters.decimate#pyvista.PolyDataFilters.decimate) to downsample datasets. Maybe add a new `low_resolution` kwarg to any loadable dataset >10 MB in size and set `low_resolution=True` wherever these are used in the documentation

I'm leaning towards options 5 and 3. This should keep the size of interactive plots inherently low so that the full build artefact size is reduced, but then also generate a second and much more minimal artefact. To enforce this it would also be nice to cap the allowed file size for any individual interactive plot, e.g. make the build fail if there is any single file > 50MB

user27182:
> Indeed, having a light documentation build artifacts for new contributors who wish to modify documentation is highly welcomed 🙏
> 
> I think I can propose a PR for option 3. Do you mind if I give it a try ?

That would be awesome! Please go ahead.

user27182:
Keeping this open to address item 5 and reduce the size of the full build artifact

user27182:
Item 5 is largely complete. The build is now 2GB in size. 

> To enforce this it would also be nice to cap the allowed file size for any individual interactive plot, e.g. make the build fail if there is any single file > 50MB

This is done in #7114

user27182:
before this issue was raised: 4.15GB
[...]

PR #7101, opened 2025-01-18
Add non-interactive doc artifacts

beroda:
### Overview

Fixes https://github.com/pyvista/pyvista/issues/7096
More specifically, option **3** from https://github.com/pyvista/pyvista/issues/7096 is tested, ie. creating a doc artifacts without `*.vtksz`  files which are storing interactive plots.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -69,7 +69,7 @@ jobs:
           name: docs-build-light
           path: |
             doc/
-            !doc/**.png
+            !doc/**/*.png
 
       - name: Cache Sphinx-Gallery Examples
         uses: actions/cache@v4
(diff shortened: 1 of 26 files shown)


user27182:
new build artifact seems to work, 450MB and I was able to download it, the interactive plots also "fail" nicely (they just show a blank white image).

beroda:
Here are screenshots of html page downloaded from the new artifact [here](https://github.com/pyvista/pyvista/actions/runs/12853915757/artifacts/2452970223).
As @user27182 correctly pointed out, the artifact is much smaller (450 Mb compared to 4 Gb for the interactive one).

image
image

tkoyama010 (approved):
LGTM

user27182 (approved):
LGTM. We could also move the new action to upload the light artifact _before_ the full 4GB docs gets uploaded, this could save waiting 6 mins in case someone is eagerly waiting to review the docs. But not a big deal.

banesullivan (approved):

PR #7104, opened 2025-01-20
Branch coverage for tests

beroda:
### Overview

Branches coverage can robustify the test suite since it shows if we tested all the potential branches.
See [the doc](https://coverage.readthedocs.io/en/7.6.10/branch.html) for more details.


user27182 (changes requested):
+1 for having more detailed information about the coverage.
 
And I agree with the changes to the coverage report that if a branch is only partially tested then it should not count as a hit, but as a partial hit instead. And so it seems our reported coverage was previously being artificially inflated.

There _could_ be false negatives with this option though which may need to be ignored:

> But there are many ways in your own code to write intentionally partial branches, and you don’t want coverage.py pestering you about them. You can tell coverage.py that you don’t want them flagged by marking them with  `# pragma: no branch`

This new pragma is likely unfamiliar to contributors. It would be good to add a blurb to CONTRIBUTING.rst (maybe under Unit Testing?) about how to ignore coverage (either with `no cover` or `no branch`) and include a link directly to the coverage docs for more info.

beroda:
> +1 for having more detailed information about the coverage.
> 
> And I agree with the changes to the coverage report that if a branch is only partially tested then it should not count as a hit, but as a partial hit instead. And so it seems our reported coverage was previously being artificially inflated.
> 
> There _could_ be false negatives with this option though which may need to be ignored:
> 
> > But there are many ways in your own code to write intentionally partial branches, and you don’t want coverage.py pestering you about them. You can tell coverage.py that you don’t want them flagged by marking them with  `# pragma: no branch`
> 
> This new pragma is likely unfamiliar to contributors. It would be good to add a blurb to CONTRIBUTING.rst (maybe under Unit Testing?) about how to ignore coverage (either with `no cover` or `no branch`) and include a link directly to the coverage docs for more info.

@user27182 Done in https://github.com/pyvista/pyvista/pull/7104/commits/6b570f7cc22c57787610f6e12d75cd17821e6831

user27182 (approved):
Thanks!

user27182 (dismissed):

banesullivan (approved):

CONTRIBUTING.rst
```diff
@@ -520,6 +520,19 @@ leverage multiple processes. Example usage:
...
 
+When submitting a PR, it is highly recommended that all modifications are thoroughly tested.
+This is further enforced in the CI by the `codecov GitHub action <https://app.codecov.io/gh/pyvista/pyvista>`_
+which has a 90% target, ie. it ensures that 90% of the code modified in the PR is tested.
+It should be mentioned that branch coverage is measured on the CI, meaning for examples that both
+values of an ``if`` clause must be tested to ensure full coverage. For more details on branch
+coverage, please refer to the `coverage documentation <https://coverage.readthedocs.io/en/latest/branch.html>`_.
+
+If needed, code coverage can be deactivated for specific lines by adding the ``# pragma: no cover`` or
+``# pragma: no branch`` comments. See the documentation `here <https://coverage.readthedocs.io/en/latest/branch.html#excluding-code>`__
```
tkoyama010, line 531: It is better practice to clearly state where a link will direct the user.
```suggestion
``# pragma: no branch`` comments. See the documentation `excluding code <https://coverage.readthedocs.io/en/latest/branch.html#excluding-code>`__
```

CONTRIBUTING.rst
```diff
@@ -520,6 +520,19 @@ leverage multiple processes. Example usage:
...
+and `here <https://coverage.readthedocs.io/en/latest/excluding.html#excluding>`__ for more details.
```
tkoyama010, line 532: Seems it it same with the first one.
```suggestion
for more details.
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

user27182:
One down side to this change is that functions with `@typing.overload` decorators are flagged as partial branches. So PRs which add typing overloads will get false positives for missing coverage on those overloads, necessitating the use of `# pragma: no branch` on every overload. E.g.

``` python
@overload
def foo(bar: int) -> int:  # pragma: no branch
    ... 
@overload
def foo(bar: str) -> str:  # pragma: no branch
    ...
def foo(bar: int | str) -> int | str:
    ...
```

And since the missing coverage is on the line _after_ the `@overload` decorator, we can't simply exclude `@overload` lines from coverage like we did for `if TYPE_CHECKING:` branches in https://github.com/pyvista/pyvista/pull/7126. I can't seem to find any workaround to address this other than ignore every line manually.

beroda:
> One down side to this change is that functions with `@typing.overload` decorators are flagged as partial branches. So PRs which add typing overloads will get false positives for missing coverage on those overloads, necessitating the use of `# pragma: no branch` on every overload. E.g.
> 
> ``` python
> @overload
> def foo(bar: int) -> int:  # pragma: no branch
>     ... 
> @overload
> def foo(bar: str) -> str:  # pragma: no branch
>     ...
> def foo(bar: int | str) -> int | str:
>     ...
> ```
> 
> And since the missing coverage is on the line _after_ the `@overload` decorator, we can't simply exclude `@overload` lines from coverage like we did for `if TYPE_CHECKING:` branches in https://github.com/pyvista/pyvista/pull/7126. I can't seem to find any workaround to address this other than ignore every line manually.
> 

pytest has an overload exclude [here](https://github.com/pytest-dev/pytest/blob/main/.coveragerc#L31)

user27182:
Thanks, will try that. So many ways (too many?) to configure code coverage it seems (codecov.yml, pyproject.toml, .coveragerc)

Issue #7097, opened 2025-01-17
Improve `Transform` API naming conventions


user27182:
### Describe what maintenance you would like added.

The [`Transform`](https://dev.pyvista.org/api/utilities/_autosummary/pyvista.transform#pyvista.Transform) class is new and is not yet part of a released. This means it's a good time to make API changes if needed since we can avoid deprecations.

It might be good idea to update the naming to better match some of the names used by the new `scipy.spatial.transform.RigidTransformation` class. See https://discuss.scientific-python.org/t/enh-scipy-spatial-transform-rigidtransformation-for-representing-3d-coordinate-frames/1673

Some possible changes:

- Rename the class to `RigidTransformation` or `RigidTransform`. VTK supports non-linear transforms as well so this would avoid future ambiguity. Currently our transforms technically support shear though, so we could explicitly check for this and forbid it? Maybe call it `AffineTransform` instead and keep the shear?
- Use `__call__` instead of `apply`? See https://github.com/scipy/scipy/issues/19254#issuecomment-1729733349, though the current[ scipy implementation ](https://github.com/scipy/scipy/pull/22267) also uses `apply`.
- The scipy implementation uses `inv()`, which I think is similar to our `Transform.invert()`
- We use `Transform.matrix` property to return the matrix, looks like scipy uses `as_matrix()` method. We also have an `inverse_matrix` property though, not sure what the name would be here.... maybe remove it and use `inv().as_matrix()` explicitly instead?
- Scipy has a `rotation` and `translation` property. I like these. We could add something similar, along with `reflection`, `scale`, and `shear`... basically one property each for `TRNSK` returned by `Transform.decompose()`.
- There may be other differences/changes/improvements...

### Links to source code.

https://github.com/scipy/scipy/pull/22267

scottshambaugh:
Popping in from the scipy side! A couple thoughts after a quick review:

* On naming, I think that the path we'll end up taking is:
  - Rotations & Translations: RigidTransformation (maybe confusingly does not include reflections)
  - Rotations, Translations, Reflection & Scale: SimilarTransformation (may skip)
  - Rotations, Translations, Reflection, Scale & Shear: AffineTransformation

  If you are jumping straight to affine transformations, I think it makes sense to keep the `Transform` naming as-is.

[...]

PR #7109, opened 2025-01-21
Remove `__matmul__` operator from `Transform` class

user27182:
### Overview

See #7097. Using `__matmul__` can be confusing and is not consistent with the scipy API for transformations.

cc @darikg


tkoyama010 (approved):
LGTM

PR #7110, opened 2025-01-21
Use `__mul__` instead of `__add__` for concatenating `Transform`

user27182:
### Overview

See #7097. Using `__mul__` instead of `__add__` for post-multiplication is consistent with scipy's transformations API.

### DIFF
--- a/tests/core/test_utilities.py
+++ b/tests/core/test_utilities.py
@@ -1564,7 +1564,7 @@ def test_transform_chain_methods():
     assert np.array_equal(matrix, eye4)
 
 
-def test_transform_mulll():
+def test_transform_add():
     scale = Transform().scale(SCALE)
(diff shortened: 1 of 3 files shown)


tkoyama010 (dismissed):
LGTM

tests/core/test_utilities.py
```diff
@@ -1564,31 +1564,28 @@ def test_transform_chain_methods():
     assert np.array_equal(matrix, eye4)
 
 
-def test_transform_add():
+def test_transform_mulll():
```
user27182, line 1567: Revert name
```suggestion
def test_transform_add():
```

tkoyama010 (dismissed):

tkoyama010 (approved):

PR #7115, opened 2025-01-23
New vector graphic PyVista logo

banesullivan:
pyvista

We have a new vectorized PyVista logo!! I recently worked with a graphic designer to vector trace the PyVista logo so that we could have a vector format (SVG) that is a bit more polished and able to scale to different resolutions easily.

This PR updates the documentation to use the new SVG logo and places the new logo in a directory at the top level of the repository to make it easier to find for use (I want to encourage people to tag the PyVista logo anywhere PyVista is showcased!)

Coming soon, I will create a variant of the PyVista logo that is more cutout-friendly (has a little cloud-like outline around the logo) so we can hand out stickers at conferences and events like SciPy.


banesullivan:
hrm, I didn't realize the SVG would be such a large diff. I'll look into ignoring this addition

.gitattributes
```diff
@@ -1 +1,2 @@
 *.ipynb linguist-detectable=false
+*.svg binary
```
banesullivan, line 2: I thought this would help reduce the diff for the SVG files. Perhaps this needs to be on `main` first?

pyvista/examples/pyvista_logo.png
user27182, line 1: This is a breaking change due to the change in dimensions. I think it's ok but a few examples will need to be updated, docs will need a .. versionchanged:: directive, etc. Is that your preference? 

Could optionally keep the resolution the same as the old file, and add a note in the docs to look at the repo root for other variations of the logo file.

banesullivan: Can you point me to an example of where this resolution change has an impact that may be considered breaking? I'm wondering if its the examples that load the logo as `vtkImageData` since the dimensions of those image data would change (aside, I'm wondering why we have those examples at all)

banesullivan: Actually, I guess the `tests/core/test_imagedata_filters.py::test_points_to_cells_and_cells_to_points_dimensions` testing failure here is telling

user27182: Check out the docstring examples for pad image. You can download the test documentation images artifact and review the differences. 100 pixels of padding looks different on a 1000x1000 image compared to the same padding on a 5000x5000 image. 

If anyone downstream is using the image and embedding it may be sized or resized based on an assumption about the input image resolution. That resizing would now break with a larger image.

user27182: I think it's fine if it's a breaking change, I'm +1 for updating the logo and making improvements and higher resolution can be nicer to work with, but just saying that it does break things so we should document it as such.

user27182: > I suggest using/copying the `logo/pyvista_sm.png` file which is 1389 x 592 pixels to be the default `pyvista/examples/pyvista_logo.png` instead. Might be good to also add a note suggesting to look in the `logo` directory for other size options.

This is implemented in #7132 (along with fixes for failed tests)

user27182 (commented):
LGTM

user27182 (approved):

.gitattributes
user27182: Maybe this will work (based on https://stackoverflow.com/a/44389165) ?
```suggestion
*.svg -text diff
```

banesullivan: My understanding is that it is equivalent, but worth a try... tested here https://github.com/banesullivan/test-svg-diff/pull/1 and seems to make no difference

user27182: ok thanks for checking, seems odd that it's not really working as expected... 🤷‍♂️

banesullivan:
Thanks for your additions @user27182 helping me get this over the finish line and prevent it from being a breaking/intrusive change!

banesullivan:
Do we still consider this a breaking change though?

user27182:
> Do we still consider this a breaking change though?

I'm kind of meh on this now. It's like a soft break and isn't really likely to cause issues downstream unless the image is being used for testing purposes. I'm fine with removing the tag.

PR #7117, opened 2025-01-23
Add complete start-to-finish plot directive example

user27182:
### Overview

I wanted to investigate / debug something relating to the plot directive but didn't see any examples I could easily run -> check -> modify -> re-run. I also found some of the existing instructions a bit vague.

So I created this all-in-one example which can be easily copy/pasted without needing to manually create any files and fully fleshes out all of the details for getting interactive plots to work.


doc/source/extras/plot_directive.rst
```diff
@@ -47,5 +47,90 @@ Which will be rendered as:
...
+at the address ``http://localhost:11000``.
+
+.. code:: bash
+
+    # Setup new virtual environment and activate it
+    python -m venv .venv
+    emulate bash -c '. .venv/bin/activate'
+
+    # Install dependencies for the build
+    pip install sphinx pyvista trame-vtk sphinx_design ipywidgets jupyter-server-proxy nest_asyncio trame-client trame-server trame-vuetify trame
```
user27182, line 75: Can probably simplify this:
```suggestion
    pip install pyvista[jupyter] sphinx sphinx_design
```
But this doesn't actually work yet since I believe `pyvistapjupyter]` is new

user27182: Figured it out. It does work, but with some shells (like zsh on macos) you need to surround it with apostrophes

tkoyama010: It seems to be common practice not to mention in the documentation that options for Python libraries do not work with zsh, as this is considered a zsh-specific issue.

user27182: Is your preference to omit the apostrophes from the docs then? Seems like a headache for macos users we can easily avoid just by including them

tkoyama010: First, let's focus on ensuring you're happy as the primary user. Please leave this as is for now. If anyone requests a correction later, we can address it then. As a Linux user, this change does not affect me.

user27182: OK. I would generally also prefer to avoid the quotes, but as a mac user this behavior is a bit surprising and the error messages are not too helpful if the quotes are omitted, so it's more user-friendly this way.

Also in case there is a repo-level preference to omit the quotes they were also included here as part of #6753 
https://github.com/pyvista/pyvista/blob/2e31b11787b7f0545d3bb3ca248873b21f506797/CONTRIBUTING.rst?plain=1#L48

user27182:
Link to the new example

https://67936ed8934a5e0ed760be37--pyvista-dev.netlify.app/extras/plot_directive.html#complete-example

tkoyama010 (approved):
LGTM

PR #7118, opened 2025-01-24
Replace code directive with code-block

user27182:
### Overview

Both `.. code::` and `.. code-block::` perform a similar function. But `.. code-block::` seems to be better supported, e.g.

-  `ruff` seems to auto-format `.. code-block:: python` but not `.. code:: python` (it auto-formatted double quotes to single quotes in this PR)
-  PyCharm provides correct syntax highlighting for `.. code-block:: python` but not `.. code:: python`


banesullivan (approved):

tkoyama010 (approved):
LGTM

PR #7121, opened 2025-01-24
Use 'static_cell' as default for `sample` filter

MatthewFlamm:
### Breaking Change

The default locator for the `sample` filter is now `'static_cell'`.  To get the previous behavior use `locator=None`.

### Overview

This PR changes the default locator so that the filter is more robust.  See https://stackoverflow.com/a/79382404/11423283.

### Details

I considered a softer deprecation here, but an issue is that the `None` input signifies a specific type of locator, i.e. the `FindCell` method of the dataset.  We could consider adding a specific string for this and switching `None` to be deprecated.  But I don't know if any large downsides to the breaking change here.


MatthewFlamm:
It could be that there are some dataset types where the original default is faster, e.e. `ImageData`, but the lack of robustness for types like `UnstructuredGrid` is a bigger problem.

tkoyama010 (approved):
LGTM

PR #7122, opened 2025-01-25
Fix removal of `user_dict`

user27182:
### Overview
Setting `user_dict = None` should clear it completely. But it does not:

``` python
import pyvista as pv

mesh = pv.Sphere()
mesh.user_dict = dict(a=2)
mesh.user_dict = None

# BUG: dict gets re-initialized with old value
mesh.user_dict 
# {"a": 2}
```
This PR fixes this. A retroactive versionadded directive is also added to the docstring.


tkoyama010 (approved):
LGTM

PR #7114, opened 2025-01-22
Add tests to limit the size of interactive plot files

user27182:
### Overview

Resolves #6515 resolves #7096. The new tests fail if any single `.vtksz` file is too large.

PR #7123, opened 2025-01-25
Downsample parched canal example before plotting

user27182:
### Overview

Part of #7096 

The interactive plot for this example is over 100MB (see https://github.com/pyvista/pyvista/pull/7114#issuecomment-2608745447). This PR removes the plot completely from the HDRReader example and down-samples the plot from the example in `downloads`.


user27182:
The build artifact is now < 4GB

tkoyama010 (approved):
Good catch. Thanks!

PR #7124, opened 2025-01-25
Add `MultiBlock` recursive iterator for iterating through all datasets

user27182:
### Overview

New method for iterating a `MultiBlock` recursively.

This can be used for querying all dataset properties or in-place filtering of datasets.
The iterator could be used to add a new `MultiBlock.flatten()` method in a follow-up PR.


tkoyama010 (approved):
LGTM

PR #7127, opened 2025-01-25
Add isolated tests for core and plotting conftest files using pytester plugin

beroda:
### Overview

`pytest` ships with a disabled by default `pytester` plugin to run isolated tests for plugins (see doc [here](https://docs.pytest.org/en/stable/reference/reference.html#pytest.Pytester)).
It notably enables to test that custom hooks, markers ... defined in conftest work as expected (ie. `needs_vtk_version`).

This PR add some tests for core conftest and plotting conftest files.


beroda:
This PR is ready. It currently tests that:

- the custom `skip_plotting` and `needs_download` markers work as expected
- the syntax of the custom `needs_vtk_version` marker is correctly skipping tests depending on vtk version
- warnings are turned to errors

tkoyama010 (approved):
LGTM

PR #7131, opened 2025-01-27
Use downsampled versions of whole body ct datasets

user27182:
### Overview

Part of #7096 

Use downsampled datasets from https://github.com/pyvista/vtk-data/pull/52 by default. These datasets are just too big as-is and take a long time to load, process, and plot, and contribute to longer build times and a larger build size artifact.

This change in the datasets is a breaking change. We could deprecate it instead (e.g. use `high_resolution=None`) or just keep `high_resolution=True` by default and update the examples to explicitly use `high_resolution=False`.

But, I think the negative impact to CI of using high-res datasets is too great to have this as the default. And downsteam users will also greatly benefit from the improved loading and processing times.

Note: the loading functions have also been refactored and placed into a separate `WholeBodyCtUtilities` class to keep things better organized


user27182:
Docs artifact is now 3.35 GB (compared to 3.81 GB from the latest main) and the relevant sphinx gallery examples run much more quickly now (e.g. 10 seconds instead of 1 minute +)

user27182:
There are some visible artifacts in the CT image:
image

The downsample in https://github.com/pyvista/vtk-data/pull/52 used `nearest` for all images, which is appropriate for the segmentation masks but not the CT image which has continuous data. I suspect some anti-aliasing and/or a different interpolator is needed. There are added in  #7144

tkoyama010 (approved):
LGTM

PR #10, opened 2018-05-22
Added functionality for filling mesh holes, scrap removal, mesh (plane) clipping and normal generation

supersubscript:
Added functionality for filling mesh holes (PolyData.FillHoles), extracting the largest domain / scrap removal (PolyData.ExtractLargest; rename?), clipping meshes with planes (PolyData.ClipPlane; rename?), generating normals (PolyData.GenerateNormals; can't currently handle vertex splitting). 

I'm not sure whether GenerateNormals should be contained in the PolyData class and just add the array to the input mesh, or whether it should return an array, similar to Curvature. At the moment it's not possible to pass the output of the PolyDataNormals filter to PolyData.Overwrite as with other functions in the PolyData class, as this doesn't preserve the PointData object. 

Also amended the PolyData.Decimate function to include all user-definable parameters (such as volume preservation).

I added some documentation for the things I'm PRing which should be complete AFAICT. Most is copied from the corresponding VTK functions, with some minor modifications.

Thanks for maintaining this really awesome package!

PR #7134, opened 2025-01-28
Expose the quadric decimation filters and its parameters

CharlesGueunet:
### Overview

Add the `decimate_quadric` for triangulated polydata

### Details

Expose the `vtkQuadricDecimation` filter as a new decimate method for polydata, including its specific parameters.


tkoyama010:
pre-commit.ci autofix

user27182:
Thanks for your contribution! The [`decimate`](https://docs.pyvista.org/api/core/_autosummary/pyvista.polydatafilters.decimate) filter uses `vtkQuadricDecimation` internally, so unfortunately it seems this PR is a duplicate of an existing filter.

user27182:
> @user27182 I will indeed update the exinsting one instead. I do not know how I missed it in the first place ...
> 
> In my version, I had no `attribute_error` as I find it redundant with the 5 distinct boolean just after. Can I deprecate the direct use of this attribute and put the default value for `scalars`, `vectors`, `normals`, `tcoords`, and `tensors` to be **False** ? (this may be breaking though) The `attribute_error` is handled internally with a `or` operator on all those..

That sounds reasonable to me!
Though I guess the current API allows for `decimate(attribute_error=True)` which is shorthand for `decimate(scalars = True, vectors = True, normals = False, tcoords = True, tensors = True)`. Might be annoying to have to type all of those out every time... Though it's also equally annoying to have to set these to False if you want to only enable a single attribute.

Not sure why `normals` is False by default though when all the others are True. The [`vtkQuadricDecimation`](https://vtk.org/doc/nightly/html/classvtkQuadricDecimation.html#details) sets _all_ to True by default, including normals.

I looked through the git blame and it looks like these options were originally added in https://github.com/pyvista/pyvista/pull/10
Though it doesn't really provide any insight into why the API is as it is. It is a very old PR though so the API can likely benefit from some updates.

What about adding a new `enable_all_attribute_error` option which will enable all (including normals)? And then set vectors, normals, tcoords, etc. as `None` by default. If `enable_all_attribute_error=True`, then these all get set to True, otherwise they are all unset (and attribute error is disabled), but allow setting individual errors to true,

e.g. currently if you want to to only use normals for attribute error, you have to do:

``` python
decimate(attribute_error = True, scalars = False, vectors = False, normals = True, tcoords = False, tensors = False)
```

Instead it would be better to do:
``` python
decimate(normals = True)
```

And to replicate the existing behavior you could do:
``` python
decimate(enable_all_attribute_error=True, normals = False)
```
OR, equivalently:
``` python
decimate(scalars = True, vectors = True, tcoords = True, tensors = True)
```
This would require deprecating `attribute_error`.

@CharlesGueunet what do you think?

CharlesGueunet:
@user27182 , I have updated the code in this regard

pyvista/core/filters/poly_data.py
```diff
@@ -1593,28 +1597,32 @@ def decimate(
...
-            If attribute errors are to be included in the metric (i.e.,
-            ``attribute_error`` is ``True``), then these flags control
-            which attributes are to be included in the error
-            calculation.
+        boundary_weight: float, default: 1.0
+           A floating point factor to weigh the boundary quadric constraints
+           by: higher factors further constrain the boundary.
 
-        vectors : bool, default: True
+        enable_all_attribute_error: bool, default: False
```
user27182, line 1608: For backwards compatibility, this should be set to `True` to keep existing behavior the same. OR, we need to somehow communicate that the future behavior will change as part of the deprecation of `attribute_error`

Also, if we keep it as `True`, we should also communicate that `normals` is now `True` (and was previously `False`) with a `.. versionchanged::` directive in the docstring

pyvista/core/filters/poly_data.py
```diff
@@ -1673,20 +1681,49 @@ def decimate(
...
+
         if not self.is_all_triangles:  # type: ignore[attr-defined]
             raise NotAllTrianglesError('Input mesh for decimation must be all triangles.')
 
+        has_attribute_error = kwargs.pop('attribute_error', False)
+        if has_attribute_error:  # pragma: no cover
+            warnings.warn(
+                "Use of `attribute_error=True` is deprecated."
+                "Use 'enable_all_attribute_error' instead.",
+                PyVistaDeprecationWarning,
```
user27182, line 1702: Should also include a `.. deprecated:: 0.45` directive in the docstring to communicate this.

pyvista/core/filters/poly_data.py
```diff
@@ -1593,28 +1597,32 @@ def decimate(
...
             is active, these errors can be large.
 
-        attribute_error : bool, default: False
-            Decide whether to include data attributes in the error metric. If
-            ``False``, then only geometric error is used to control the
-            decimation. If ``True``, the following flags are used to specify
-            which attributes are to be included in the error calculation.
+        boundary_constraints: bool, default: False
+            Use the legacy weighting by boundary_edge_length instead of by
+            boundary_edge_length^2 for backwards compatibility
```
user27182, line 1602: Include a `.. versionadded:: 0.45` directive for new keywords
```suggestion
        boundary_constraints: bool, default: False
            Use the legacy weighting by boundary_edge_length instead of by
            boundary_edge_length^2 for backwards compatibility
            
           .. versionadded:: 0.45
```

pyvista/core/filters/poly_data.py
```diff
@@ -1593,28 +1597,32 @@ def decimate(
...
 
-        scalars : bool, default: True
-            If attribute errors are to be included in the metric (i.e.,
-            ``attribute_error`` is ``True``), then these flags control
-            which attributes are to be included in the error
-            calculation.
+        boundary_weight: float, default: 1.0
+           A floating point factor to weigh the boundary quadric constraints
+           by: higher factors further constrain the boundary.
```
user27182, line 1606: Include a `.. versionadded:: 0.45` directive
```suggestion
        boundary_weight: float, default: 1.0
           A floating point factor to weigh the boundary quadric constraints
           by: higher factors further constrain the boundary.
           
           .. versionadded:: 0.45
```

pyvista/core/filters/poly_data.py
```diff
@@ -1563,19 +1564,22 @@ def decimate(
...
+        # tcoords: bool,  # kwargs for default value
+        # tensors: bool,  # kwargs for default value
         scalars_weight=0.1,
         vectors_weight=0.1,
         normals_weight=0.1,
         tcoords_weight=0.1,
         tensors_weight=0.1,
         inplace: bool = False,
         progress_bar: bool = False,
+        **kwargs
```
user27182, line 1582: In the docsstring, `**kwargs` will need to be documented, e.g.:

``` python
        **kwargs : dict, optional
            Used for handling deprecated parameters.
```

pyvista/core/filters/poly_data.py
CharlesGueunet: Added in the sentence directly, tell me if you would rather have the specific label

pyvista/core/filters/poly_data.py
CharlesGueunet: I think my comment line [1584](https://github.com/pyvista/pyvista/pull/7134/files#diff-cc6362aa4f128e0c10e9c44265623056dfa33b382a5f33cbb56c05a02cacf06aR1586) provides more insight than before. Please tell me if I need to add more details

pyvista/core/filters/poly_data.py
```diff
@@ -1673,20 +1681,49 @@ def decimate(
...
+            )
+            enable_all_attribute_error = True
+
+        # Get each attributes if defined, otherwise fallback to the
+        # enable_all_attribute_error value
+        use_scalars = kwargs.pop('scalars', enable_all_attribute_error)
+        use_vectors = kwargs.pop('vectors', enable_all_attribute_error)
+        use_normals = kwargs.pop('normals', enable_all_attribute_error)
+        use_tcoords = kwargs.pop('tcoords', enable_all_attribute_error)
+        use_tensors = kwargs.pop('tensors', enable_all_attribute_error)
```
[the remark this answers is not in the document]
user27182: You can set them all to be `None` by default, and if `None`, then set their value to `enable_all_attribute_error`

pyvista/core/filters/poly_data.py
user27182: Looks good I'll mark this as resolved, if it needs further revision I'll comment there directly

pyvista/core/filters/poly_data.py
```diff
@@ -1634,12 +1642,27 @@ def decimate(
...
+            Use the legacy weighting by boundary_edge_length instead of by
+            boundary_edge_length^2 for backwards compatibility
+
+            .. versionadded:: 0.45.0
+
+        boundary_weight: float, default: 1.0
+           A floating point factor to weigh the boundary quadric constraints
+           by: higher factors further constrain the boundary.
+
+            .. versionadded:: 0.45.0
```
user27182, line 1655: The docs may fail here, `..` should align with the text
```suggestion
           .. versionadded:: 0.45.0
```

pyvista/core/filters/poly_data.py
```diff
@@ -16,6 +16,7 @@
 from pyvista.core._typing_core import NumpyArray
 from pyvista.core.errors import MissingDataError
 from pyvista.core.errors import NotAllTrianglesError
+from pyvista.core.errors import PyVistaDeprecationWarning
```
user27182, line 19: Need to import `assert_empty_kwargs`
```suggestion
from pyvista.core.errors import PyVistaDeprecationWarning
from pyvista.core.utilities.misc import assert_empty_kwargs
```

pyvista/core/filters/poly_data.py
```diff
@@ -1563,22 +1564,30 @@ def decimate(
...
-        scalars: bool = True,
-        vectors: bool = True,
-        normals: bool = False,
-        tcoords: bool = True,
-        tensors: bool = True,
+        # scalars: bool,  # kwargs for default value
+        # vectors: bool,  # kwargs for default value
+        # normals: bool,  # kwargs for default value
+        # tcoords: bool,  # kwargs for default value
+        # tensors: bool,  # kwargs for default value
```
user27182, line 1571: Use `None` and set value dynamically based on `enable_all_attribute_error`
```suggestion
        scalars: bool | None = None,
        vectors: bool | None = None,
        normals: bool | None = None,
        tcoords: bool | None = None,
        tensors: bool | None = None,
```

pyvista/core/filters/poly_data.py
```diff
@@ -1671,22 +1694,56 @@ def decimate(
...
             raise NotAllTrianglesError('Input mesh for decimation must be all triangles.')
 
+        has_attribute_error = kwargs.pop('attribute_error', False)
+        if has_attribute_error:  # pragma: no cover
+            warnings.warn(
+                "Since 0.45, use of `attribute_error=True` is deprecated."
+                "Use 'enable_all_attribute_error' instead.",
+                PyVistaDeprecationWarning,
+            )
+            enable_all_attribute_error = True
```
user27182, line 1719: For backwards compatibility maybe we should also set normals=False here
```suggestion
            enable_all_attribute_error = True
            if normals is None:
                normals = False
```

pyvista/core/filters/poly_data.py
```diff
@@ -1563,22 +1564,30 @@ def decimate(
...
+        boundary_constraints: bool = False,
+        boundary_weight: float = 1.0,
+        enable_all_attribute_error: bool = False,
+        **kwargs
     ):
         """Reduce the number of triangles in a triangular mesh using ``vtkQuadricDecimation``.
 
+        .. versionchanged:: 0.45
+          scalars, vectors, normals, tcoords and tensors are now disabled by default.
+          They can be enabled alltogether using `enable_all_attribute_error`.
```
user27182, line 1588: Format parameter names as code literals with double backtick (single backtick is italics in rst)
```suggestion
          ``scalars``, ``vectors``, ``normals``, ``tcoords`` and ``tensors`` are now disabled by default.
          They can be enabled alltogether using ``enable_all_attribute_error``.
```

pyvista/core/filters/poly_data.py
```diff
@@ -1671,22 +1694,56 @@ def decimate(
...
+        use_vectors = kwargs.pop('vectors', enable_all_attribute_error)
+        use_normals = kwargs.pop('normals', enable_all_attribute_error)
+        use_tcoords = kwargs.pop('tcoords', enable_all_attribute_error)
+        use_tensors = kwargs.pop('tensors', enable_all_attribute_error)
+        use_attribute = use_scalars or use_vectors or use_normals or use_tcoords or use_tensors
+
+        if not assert_empty_kwargs():
+            warnings.warn(
+                "Unused arguments in decimater: ", kwargs
+            )
```
user27182, line 1733: Should be able to do
```suggestion
        assert_empty_kwargs(**kwargs):
```
immediately after using `pop`

pyvista/core/filters/poly_data.py
CharlesGueunet: not needed, I removed `kwargs`

pyvista/core/filters/poly_data.py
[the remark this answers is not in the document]
CharlesGueunet: thanks to this trick, I removed the whole `kwargs`

pyvista/core/filters/poly_data.py
CharlesGueunet: no more `kwargs`

CharlesGueunet:
Ready on my side, tell me if you need anything else

user27182:
Unit tests for vtk < 9.3 are failing. The new options will need to check vtk version and, if enabled, raise an error if vtk < 9.3.

A note about requires vtk >= 9.3 should also be added to the docstring for these parameters

tkoyama010:
pre-commit.ci autofix

pyvista/core/filters/poly_data.py
```diff
@@ -1562,22 +1562,30 @@ def decimate(
...
         progress_bar: bool = False,
+        boundary_constraints: bool = False,
+        boundary_weight: float = 1.0,
+        enable_all_attribute_error: bool = False,
     ):
         """Reduce the number of triangles in a triangular mesh using ``vtkQuadricDecimation``.
 
+        .. versionchanged:: 0.45
+          ``scalars``, ``vectors``, ``normals``, ``tcoords`` and ``tensors`` are now disabled by default.
+          They can be enabled alltogether using ``enable_all_attribute_error``.
```
tkoyama010, line 1586:
```suggestion
          They can be enabled all together using ``enable_all_attribute_error``.
```

tkoyama010 (changes requested):

pyvista/core/filters/poly_data.py
```diff
@@ -1592,28 +1600,35 @@ def decimate(
             volume preservation is disabled and if ``attribute_error``
             is active, these errors can be large.
 
+        enable_all_attribute_error: bool, default: False
```
tkoyama010, line 1603: nits The argument order in the function definition does not match the order in the docstring. Please correct this discrepancy.

tkoyama010 (changes requested):

tkoyama010:
Everything else looks good to me!

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM!

PR #7144, opened 2025-01-28
Add new `resample` interpolation modes and `anti_aliasing` option

user27182:
### Overview

Follow-up to #7068.

In that PR [`vtkImageResample`](https://vtk.org/doc/nightly/html/classvtkImageResample.html#details) is used but [`vtkImageResize`](https://vtk.org/doc/nightly/html/classvtkImageResize.html#details) is apparently newer and more performant. In addition, `vtkImageResize` supports more general interpolators, including `vtkImageSincInterpolator`. 

Changes in this PR:
- Use `vtkImageResize` instead of  `vtkImageResample`
- Add interpolators `'hamming'`, `'blackman'`, and `'lanczos'`. This makes `resample` more similar to other libraries such as [`Pillow.resize`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.resize) which supports some of these interpolators as well.
- Add `anti_aliasing` option which is often necessary for down-sampling.
- Always return image with the same dtype as input. Previously, `float` scalars were always returned when using linear or cubic interpolation.

There's some good background on the the interpolators used by `vtkImageResize` here:
https://vtk.org/Wiki/VTK/Image_Interpolators


tkoyama010 (approved):
LGTM

PR #7146, opened 2025-01-30
Load decimated versions of large downloaded datasets by default

user27182:
### Overview

Part of #7096.

Load decimated datasets from https://github.com/pyvista/vtk-data/pull/54 and https://github.com/pyvista/vtk-data/pull/55 by default.

- download_bird_bath
- download_black_vase
- download_ivan_angel
- download_owl
- download_plastic_vase
- download_sea_vase
- download_action_figure

### DIFF
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -4831,16 +4831,29 @@ def download_drill(load=True):  # pragma: no cover
 _dataset_drill = _SingleFileDownloadableDatasetLoader('drill.obj')
 
 
-def download_action_figure(load=True):  # pragma: no cover
+def download_action_figure(load=True, *, high_resolution=False):  # pragma: no cover
     """Download scan of an action figure.
 
     Originally obtained from Laser Design.
+
+    .. versionchanged:: 0.45
+
+        A decimated version of this dataset with # points is now returned.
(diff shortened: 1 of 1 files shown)


pyvista/examples/downloads.py
```diff
@@ -4831,17 +4831,30 @@ def download_drill(load=True):  # pragma: no cover
 _dataset_drill = _SingleFileDownloadableDatasetLoader('drill.obj')
 
 
-def download_action_figure(load=True):  # pragma: no cover
+def download_action_figure(load=True, *, high_resolution=False):  # pragma: no cover
```
user27182, line 4834: Can also call this `decimate`... or some other term (coarse=True) ?

```suggestion
def download_action_figure(load=True, *, decimate=True):  # pragma: no cover
```

Note that a similar change is introduced in #7131 which also uses the term `high_resolution`, but this dataset is ImageData so `resample` is used instead of `decimate`. So a similar suggestion there may be to use the term `downsample` instead `high_resolution`.

tkoyama010 (approved):
LGTM

user27182:
Small note: this PR also removes the images from #7149. I think this is OK but maybe if you want a cleaner diff merge that PR first?

tkoyama010:
Thanks!

PR #7148, opened 2025-01-30
Move `voxelize_binary_mask` to DataSetFilters

user27182:
### Overview

Part of #7037. All voxelize methods are being moved to `DataSetFilters`.

The filter is moved as-is without modifications _except_ `extract_geometry` is now used on the input and validation is modified slightly to check that the input has faces.


tkoyama010 (approved):
LGTM

PR #7151, opened 2025-01-30
Rename `Transform.concatenate` to `compose`

user27182:
### Overview

Part of #7097

Scipy has a `Rotation.concatenate` method which is completely different to `Transform.concatenate`.

This PR renames `Transform.concatenate` to avoid confusion if working between the two libraries, and also leaves the option open to implement `Transform.concatenate` in the future with a more similar behavior to `Rotation.concatenate`


tkoyama010 (approved):
LGTM

PR #7154, opened 2025-01-31
Exclude redundant `vtksz` files from doc upload

### LINKED ISSUE #7147: Plot directive creates redundant copies of interactive plot files

[#7114](https://github.com/pyvista/pyvista/pull/7114) revealed that duplicate copies of `vtksz` files are generated and included as part of the `html` output.

These files can be quite large and currently contribute more than 3GB collectively to the doc artifact size. Removing the copies can potentially reduce the build size by more than 1.5GB.

build docs

### System Information

```shell
CI docs build
```

### Screenshots

_No response_

user27182:
### Overview

Related to #7096, Fix #7147 

This is more of a workaround, it would be better to prevent redundant copies in the first place though I suspect this is required to make the plot directive work.


user27182:
Build artifact is now 2.58 GB instead of 3.81. And interactive plots still work because they use images from `build/_images`, so any other vtksz files can be omitted

user27182:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM

banesullivan (approved):
I don't totally understand why there are redundancies, but trusting you here

user27182:
> I don't totally understand why there are redundancies, but trusting you here

It may be unavoidable, looks like sphinx is doing the copying for PNG files. E.g. See build logs:
https://github.com/pyvista/pyvista/actions/runs/13077346437/job/36492675465#step:11:6553
```
copying images... [  0%] ../_build/plot_directive/api/core/_autosummary/pyvista-Camera-1_00_00.png
```

Maybe the plot directive does something similar?

user27182:
> > I don't totally understand why there are redundancies, but trusting you here
> 
> It may be unavoidable, looks like sphinx is doing the copying for PNG files. E.g. See build logs: https://github.com/pyvista/pyvista/actions/runs/13077346437/job/36492675465#step:11:6553
> 
> ```
> copying images... [  0%] ../_build/plot_directive/api/core/_autosummary/pyvista-Camera-1_00_00.png
> ```
> 
> Maybe the plot directive does something similar?

The issue is with this code:
https://github.com/pyvista/pyvista/blob/3ffbba5c2602b9105c1dddfdcad92d62f7f9b181/pyvista/ext/plot_directive.py#L794-L801

It generates a copy unnecessarily and pollutes the `html` build dir with a bunch of pyvista plot directive output files. This code can be safely removed, which is done in #8834. The `vtksz` excludes in the docs build workflow can also be removed accordingly.

PR #7155, opened 2025-01-31
fix: use color cycler if color=True

### LINKED ISSUE #7153: Color cycler is not used with `color=True`

If a color cycler is set to be used, I expect that specifying a generic `color=True` in `Plotter.add_mesh` to use the next color in the cycler rather than the default color. At present, using `color=True` ignores the cycler.

This is particularly troublesome if you have multiple datasets which each have scalar arrays you don't want to use and you just want to plot them with different colors

```py
import pyvista as pv

pv.global_theme.color_cycler = 'default'

a = pv.Wavelet().clip(invert=True)
b = pv.Wavelet().clip(invert=False)

pl = pv.Plotter()
pl.add_mesh(a, color=True)
pl.add_mesh(b, color=True)
pl.show()
```

### System Information

```shell
latest version of pyvista
```

### Screenshots

| Current output | expect output |
|---|---|
|  Image |  Image |

banesullivan:
Resolves #7153 by checking if `color is True` when fetching the next color.

This removes `color` from the `_common_arg_parser` since it is no longer needed to be parsed there

This also adds a new fixture to `tests/plotting` that is automatically used for every test and it resets the global theme for every test to prevent theme modifications from leaking between tests


tests/plotting/test_plotting.py
```diff
@@ -3919,6 +3919,20 @@ def test_color_cycler():
         pl.set_color_cycler(5)
 
 
+def test_color_cycler_true():
+    pv.global_theme.color_cycler = 'default'
```
user27182, line 3923: This test is modifies the theme globally and affects other tests. Use a fixture to reset the value after testing.

```suggestion
@pytest.fixture
def _modifies_color_cycler():
    before = pv.global_theme.color_cycler
    yield
    pv.global_theme.color_cycler = before


@pytest.mark.usefixtures('_modifies_color_cycler')
def test_color_cycler_true():
    pv.global_theme.color_cycler = 'default'
```

MatthewFlamm: or maybe better is to use a per-plotter theme?

banesullivan: Huh, I thought we already had an auto fixture that autoresets the theme for every single test to avoid this

banesullivan: Making a new fixture:

```py
@pytest.fixture(autouse=True)
def set_default_theme():
    """Reset the testing theme for every test."""
    pv.global_theme.load_theme(pv.plotting.themes._TestingTheme())
    yield
    pv.global_theme.load_theme(pv.plotting.themes._TestingTheme())
```

user27182:
This fix is nice though I wonder if this behavior is documented anywhere. In particular I don't think it's obvious that setting `color=True` is needed to forcibly ignore scalar coloring.

Should something be added to the docs in [`Plotter.set_color_cycler`](https://dev.pyvista.org/api/plotting/_autosummary/pyvista.plotter.set_color_cycler#pyvista.Plotter.set_color_cycler) ?

E.g. in rst:
``` rst
If a mesh has scalar data, set ``color=True`` in the call to :meth:`add_mesh` to 
color the mesh with the next color in the cycler. Otherwise the mesh's scalars are
used to color the mesh by default.
```

And maybe also include the example used in #7153 in the docstring ?

``` rst
Load a mesh with active scalars and split it into two separate meshes.

>>> import pyvista as pv
>>> a = pv.Wavelet().clip(invert=True)
>>> b = pv.Wavelet().clip(invert=False)

Enable color cycling and set ``color=True`` to force the meshes to be colored with the 
cycler's colors.

>>> pv.global_theme.color_cycler = 'default'
>>> pl = pv.Plotter()
>>> pl.add_mesh(a, color=True)
>>> pl.add_mesh(b, color=True)
>>> pl.show()
```

tests/plotting/conftest.py
```diff
@@ -102,6 +102,14 @@ def colorful_tetrahedron():
     return mesh
 
 
+@pytest.fixture(autouse=True)
+def set_default_theme():
+    """Reset the testing theme for every test."""
```
MatthewFlamm, line 107: Can we get rid of the global theme handling at the very top of this file now?  It is better to have a single location.

banesullivan: Totally agree, will do this

banesullivan:
> In particular I don't think it's obvious that setting color=True is needed to forcibly ignore scalar coloring.

I agree not obvious but a definite feature since the early days! I'll add an example for this

user27182:
> > In particular I don't think it's obvious that setting color=True is needed to forcibly ignore scalar coloring.
> 
> I agree not obvious but a definite feature since the early days! I'll add an example for this

Thanks, yes sorry you're right `color=True` would apply the default theme color instead of coloring with scalar values. I meant more in the context of using the color cycler, like if I enable color cycling I might expect that by default this should color meshes _without_ needing to set `color=True`. But the way it's implemented setting `color=True` is required to make this work as expected in all cases, so that's the non-obvious thing I'm referring to.

banesullivan:
Ah I think I see, I'm inferring the confusing behaviour to be the fact that scalars are **always** preferred/used by default when plotting

user27182:
> Ah I think I see, I'm inferring the confusing behavior to be the fact that scalars are **always** preferred/used by default when plotting

Yes I think that's the source of possible confusion. Something like a note reminding users of this behavior, even with the color cycler enabled, I think would be helpful here (and so use `color=True` to override this)

pyvista/plotting/renderer.py
```diff
@@ -350,6 +350,11 @@ def set_color_cycler(self, color_cycler) -> None:
         Setting to ``None`` will disable the use of the color cycler on this
         renderer.
 
+        .. note::
+            If a mesh has scalar data, set ``color=True`` in the call to :meth:`add_mesh`
```
user27182, line 354:
```suggestion
            If a mesh has scalar data, set ``color=True`` in the call to :meth:`~pyvista.Plotter.add_mesh`
```

pyvista/plotting/renderer.py
```diff
@@ -368,6 +373,21 @@ def set_color_cycler(self, color_cycler) -> None:
...
+        >>> b = pv.Wavelet().clip(invert=False)
+
+        Enable color cycling and set ``color=True`` to force the meshes to be colored with the
+        cycler's colors.
+
+        >>> pv.global_theme.color_cycler = 'default'
+        >>> pl = pv.Plotter()
+        >>> pl.add_mesh(a, color=True)
+        >>> pl.add_mesh(b, color=True)
+        >>> pl.show()
```
user27182, line 389:
```suggestion
        Load a mesh with active scalars and split it into two separate meshes.
        
        >>> mesh = pv.Wavelet()
        >>> mesh.active_scalars_name
        'RTData'
        
        >>> a = mesh.clip(invert=True)
        >>> b = mesh.clip(invert=False)

        Enable color cycling and set ``color=True`` to force the meshes to be colored with the
        cycler's colors.

        >>> pv.global_theme.color_cycler = 'default'
        >>> pl = pv.Plotter()
        >>> _ = pl.add_mesh(a, color=True)
        >>> _ = pl.add_mesh(b, color=True)
        >>> pl.show()
```

user27182:
pre-commit.ci autofix

user27182 (approved):

banesullivan:
Thanks for getting this over the finish line @user27182!

PR #7156, opened 2025-01-31
Add planets module to dataset gallery

user27182:
### Overview

Update `planets.py` to use the dataset loaders and update docs to include the datasets in the dataset gallery.

Note: `load` functions are not included since these all return the same blank (untextured) sphere by default and `load_saturn_rings` conflicts with `download_saturn_rings`. See #7165


tkoyama010 (approved):
LGTM

PR #7158, opened 2025-01-31
Check download urls with `urlparse` instead of `requests`

user27182:
### Overview

Using `requests` to check download urls is slow (approx 1 second per request, and we have almost 200 downloads). Instead we can use `urlparse` to simply check the structure of the URL. This will reduce test times by approx 3 minutes.


MatthewFlamm:
I think this is a good idea overall, but `urlparse` does not validate that the url is valid, according to the docstring that was added.  See https://docs.python.org/3/library/urllib.parse.html#url-parsing-security

> The [urlsplit()](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit) and [urlparse()](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) APIs do not perform validation of inputs. They may not raise errors on inputs that other applications consider invalid. They may also succeed on some inputs that might not be considered URLs elsewhere. Their purpose is for practical functionality rather than purity.

I do not know enough whether this is more from a security standpoint and for purposes here it is okay?

user27182:
> I think this is a good idea overall, but `urlparse` does not validate that the url is valid, according to the docstring that was added. See https://docs.python.org/3/library/urllib.parse.html#url-parsing-security
> 
> > The [urlsplit()](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit) and [urlparse()](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) APIs do not perform validation of inputs. They may not raise errors on inputs that other applications consider invalid. They may also succeed on some inputs that might not be considered URLs elsewhere. Their purpose is for practical functionality rather than purity.
> 
> I do not know enough whether this is more from a security standpoint and for purposes here it is okay?

This change actually stems from #7078. Enabling `'S'` with `ruff` will make the suggestion to avoid using `requests`

From a testing perspective, the actual URL validation happens when the dataset is downloaded. If the URL is bad the download fails and the build fails. So we still have coverage that the links actually work.

tests/examples/test_downloads.py
```diff
@@ -31,12 +31,12 @@ def test_dataset_loader_name_matches_download_name(test_case: DatasetLoaderTestC
 
 
 def _is_valid_url(url):
+    """Check if a URL has a valid format."""
     try:
-        requests.get(url)
-    except requests.RequestException:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc])
```
MatthewFlamm, line 37: Is there also always a path expected?  I would think so?

user27182: Yes you're right currently it's only validating `https://github.com` for all downloads.

I'm not sure this overall change is good after all, the tests seem kind of useless now...

I'll propose some optimizations to the use of requests instead

MatthewFlamm: Is there a common path up to some point?  We could test that all paths are correct up to the common part.  This seems reasonable to me, and the speedup is very welcome.

user27182:
#7825 addresses this by checking `is_file()` using a local cache of `vtk-data` and avoiding the http requests altogether.

PR #7166, opened 2025-02-02
Fix doc build error on main by updating images

tkoyama010:
### Overview

Fix doc build error on main by updating images.

### Details

- None


user27182:
These changes are not correct. The failures appear to be caused by a stale CI cache being incorrectly used. Not sure why it's not being invalidated / refreshed.

The files currently in `doc_image_cache` in main are correct, e.g. https://github.com/pyvista/pyvista/pull/7148 moved `voxelize_binary_mask` images to `DataSetFilters`, so these should _not_ have `PolyDataFilters` in the name. And the other images causing failures are from #7131 but are also currently correct (and should not be modified/reverted).

user27182:
E.g. the most recent commit to main is failing CI:
https://github.com/pyvista/pyvista/actions/runs/13095571810/job/36537482837

The build time though is under 2 hours, indicating the cache is being used. But it should not be, these should be invalidated and regenerated, and the build time should be closer to 3 hrs.

tkoyama010:
OK. I removed the cache on the main branch and re-ran the document build. Let's check it.

user27182:
> OK. I removed the cache on the main branch and re-ran the document build. Let's check it.

CI on main is now passing ✅🎉
https://github.com/pyvista/pyvista/actions/runs/13096733108/job/36545502005

user27182:
This issue seems to persist and new commits are not refreshing the cache. Could there be an error in the logic for how/when the cache is applied? It's my understanding that the cache should _not_ be used for new commits to main, or am I missing something?

PR #7167, opened 2025-02-02
Add `MultiBlock.flatten` and `MultiBlock.recursive_iterator` options

user27182:
### Overview

Use #7124 to flatten `MultiBlock`.


pyvista/core/composite.py
```diff
@@ -181,25 +188,41 @@ def recursive_iterator(self: MultiBlock, *, skip_none: bool = True) -> Iterator[
         skip_none : bool, default: True
             Do not include ``None`` blocks in the iterator.
 
+        contents : 'keys', 'blocks', 'items', default: 'blocks'
```
MatthewFlamm, line 191: Drive by comment here.  I like the direction so far of this PR, but haven't read in super detail.

I see an inconsistency between `keys` and `names` in various sections here.  Since `MultiBlock` has some `dict` features and has a `keys` method, using `keys` as the string value here makes sense to me.  To be consistent with `dict`-like features, we could consider `keys` and `values`, but I think `blocks` is more vtk-like.  So I would vote for `keys` and `blocks`.

user27182: Thanks for the feedback, yes I kind of waded back and forth with the naming as I was not sure which to use.

> So I would vote for `keys` and `blocks`.

Ok I'll standardize the naming to use these.

MatthewFlamm: On a contrary note however, there is a `set_block_name` method. 
 Inside the vtk machinery uses `NAME` ID.  With this I am more neutral as long as it is consitent. 
 https://github.com/pyvista/pyvista/blob/05a8c71ac528fa787b338ef52c47aa44b931c469/pyvista/core/composite.py#L692

user27182: Yes there is also a `name_mode` as a kwarg in the new `flatten` method, and renaming this to `key_mode` seems less clear... and keeping it as `name_mode` but using `keys` in the recursive iterator makes this self-inconsistent

> With this I am more neutral as long as it is consitent.

Sounds good. I think I started with `keys` initially but gravitated to `names` part-way through the implementation. Maybe let's go with `names`, `blocks`, `items` then.

MatthewFlamm: I'm +1 for your suggestion.

pyvista/core/composite.py
```diff
@@ -171,55 +173,362 @@ def wrap_nested(self: MultiBlock) -> None:
...
+        order : 'breadth', 'depth', 'hybrid', default: 'hybrid'
+            Order in which to iterate through nested blocks.
+
+            - ``'breadth'``: Iterate breadth-first. This option reorders the blocks to
+              iterate through nested ``MultiBlock`` blocks last.
+            - ``'depth'``: Iterate depth-first. This option reorders the blocks to
+              iterate through nested ``MultiBlock`` blocks first.
+            - ``'hybrid'``: Iterate breadth-first initially but iterate through
+              any nested blocks when they are encountered. This iterates through the
+              ``MultiBlock`` recursively as-is without reordering.
```
user27182, line 213: Not sure if these option names `'breadth'` , `'depth'` , `'hybrid'` are clear. Maybe something like `'last'`, `'first'`, `'as-is'` is better? Open to suggestions

tkoyama010: > EDIT: Or even 'last', 'first', None where None is used instead of as-is, i.e. no-reordering
+1

user27182: Options are now `nested_first`, `nested_list`, and `None`

tkoyama010 (approved):
LGTM

tests/core/test_composite.py
```diff
@@ -902,16 +902,125 @@ def test_transform_filter_inplace_default_warns(multiblock_poly):
...
+    # Test breadth, expect nested dataset last
+    iterator = list(nested.recursive_iterator('items', order='nested_last', skip_empty=False))
+    assert iterator[0][0] == 'image'
+    assert iterator[1][0] == 'poly'
+    assert iterator[2][0] == 'grid'
+    assert isinstance(iterator[0][1], pv.ImageData)
+    assert isinstance(iterator[1][1], pv.PolyData)
+    assert isinstance(iterator[2][1], pv.UnstructuredGrid)
+
+    # Test depth, expect nested dataset first
```
user27182, line 986:
```suggestion
    # Expect nested dataset first
```

tests/core/test_composite.py
```diff
@@ -902,16 +902,125 @@ def test_transform_filter_inplace_default_warns(multiblock_poly):
...
+    # Test adaptive, expect nested dataset in center
+    iterator = list(nested.recursive_iterator('items', order=None, skip_empty=False))
+    assert iterator[0][0] == 'image'
+    assert iterator[1][0] == 'grid'
+    assert iterator[2][0] == 'poly'
+    assert isinstance(iterator[0][1], pv.ImageData)
+    assert isinstance(iterator[1][1], pv.UnstructuredGrid)
+    assert isinstance(iterator[2][1], pv.PolyData)
+
+    # Test breadth, expect nested dataset last
```
user27182, line 977:
```suggestion
    # Expect nested dataset last
```

tests/core/test_composite.py
```diff
@@ -902,16 +902,125 @@ def test_transform_filter_inplace_default_warns(multiblock_poly):
...
+def test_recursive_iterator_order():
+    nested = pv.MultiBlock(dict(image=pv.ImageData(), poly=pv.PolyData()))
+    multi = pv.MultiBlock(dict(grid=pv.UnstructuredGrid()))
+    nested.insert(1, multi)
+    assert isinstance(nested[0], pv.ImageData)
+    assert isinstance(nested[1], pv.MultiBlock)
+    assert isinstance(nested[1][0], pv.UnstructuredGrid)
+    assert isinstance(nested[2], pv.PolyData)
+
+    # Test adaptive, expect nested dataset in center
```
user27182, line 968:
```suggestion
    # Expect nested dataset in center
```

pyvista/core/composite.py
```diff
@@ -171,55 +173,360 @@ def wrap_nested(self: MultiBlock) -> None:
...
+                if isinstance(block, MultiBlock):
+                    multi_names.append(name)
+                    multi_blocks.append(block)
+                else:
+                    other_names.append(name)
+                    other_blocks.append(block)
+            if order == 'nested_last':
+                names = [*other_names, *multi_names]
+                blocks = [*other_blocks, *multi_blocks]
+            else:  # 'depth'
```
user27182, line 347:
```suggestion
            else:
```

user27182 (commented):
Update comments to reflect new order names

tkoyama010 (dismissed):

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

PR #7169, opened 2025-02-03
Rename examples to use `snake_case` with underscores

user27182:
### Overview

The guide for adding new examples suggests using `-` not `_` for example file names. A few files are non-conformant.

https://dev.pyvista.org/examples/99-advanced/add-example

EDIT: 
This PR updates the examples to use `snake_case` with underscores instead of dashes


tkoyama010 (changes requested):
In the Python community, it is a common convention to use snake_case for file names. I suggest standardizing by using underscores (_) instead of hyphens (-). Since I haven't provided evidence to support this convention, I will conduct additional research later.

user27182:
> In the Python community, it is a common convention to use snake_case for file names. I suggest standardizing by using underscores (_) instead of hyphens (-). Since I haven't provided evidence to support this convention, I will conduct additional research later.

All of our `.py` files everywhere else in the repo use underscores (e.g. `data_set.py`). So we have a convention established in that sense. I think following this convention in `examples` also makes sense. 

So +1 for using underscores instead. The text in "[add example](https://dev.pyvista.org/examples/99-advanced/add-example)" will also need to be updated accordingly

PR #7175, opened 2025-02-05
Remove `mesh_length_fraction` from `voxelize_binary_mask`

user27182:
### Overview

See https://github.com/pyvista/pyvista/pull/7037#discussion_r1937636156

This filter is not yet released so I don't think we need to deprecate this and can simply remove it.

Note: PolyData filter tests were also moved to DataSet filter tests. These were missed by #7148


tkoyama010 (approved):
LGTM

PR #7176, opened 2025-02-05
Remove `check_subdtype` test of deprecated NumPy behavior

user27182:
### Overview

This test was originally included because of surprising behavior when using `np.number` with `check_subdtype`. But this usage emits a deprecation warning (so users can know to avoid it) and this deprecation is now an error with NumPy 2.3 and causing CI to fail.

We can simply remove this test.


tkoyama010 (approved):
LGTM

PR #7177, opened 2025-02-05
Use elif instead of else then if, to reduce indentation

tkoyama010:
### Overview

If an `else` block contains a single `if` statement, it can be collapsed into an `elif`, thus reducing the indentation level. I've come across this anti-pattern several times during reviews. Let's check for it in advance using ruff.

### Details

- [collapsible-else-if (PLR5501)](https://docs.astral.sh/ruff/rules/collapsible-else-if/#collapsible-else-if-plr5501)


user27182 (approved):
LGTM. If these changes are all automatic (they should be), then we should not worry about missing coverage for some of the changes. I trust `ruff`'s safe auto-fixes.

user27182:
pre-commit.ci autofix

user27182 (approved):

user27182 (approved):

PR #7179, opened 2025-02-06
Add PLR1730 `ruff` rule for min/max statements

user27182:
### Overview

See https://docs.astral.sh/ruff/rules/if-stmt-min-max/#if-stmt-min-max-plr1730


tkoyama010 (approved):
LGTM

Issue #5179, opened 2023-11-06
`renderer.bounds` includes invisible actors


user27182:
### Describe the bug, what's wrong, and what you expected.

The renderer bounds are currently calculated for all actors regardless of their visibility. This means that actors with `VisibilityOff()` or `UseBoundsOff()` will have their bounds included in the renderer's bounds when they should not be. The expected behaviour is that the bounds should match the output from [`vtkRenderer.ComputeVisiblePropBounds()`](https://github.com/Kitware/VTK/blob/b8817fd385de2d6dc8119b06c438a762e6e6cede/Rendering/Core/vtkRenderer.cxx#L956C36-L956C36), which is what [`vtkRenderer.ResetCamera()`](https://github.com/Kitware/VTK/blob/b8817fd385de2d6dc8119b06c438a762e6e6cede/Rendering/Core/vtkRenderer.cxx#L1027C22-L1027C22) calls internally.

I think PyVista's docs, e.g. for [`Plotter.bounds`](https://docs.pyvista.org/version/stable/api/plotting/_autosummary/pyvista.Plotter.bounds.html) and [`Camera.tight`](https://docs.pyvista.org/version/0.42/api/core/_autosummary/pyvista.Camera.tight.html) could be more explicit as well and use the term "visible actors" where possible to make it clear that actor visibility is taken into consideration.

If there is a need for getting all bounds regardless of visibility, maybe this should be parameterized, e.g. `include_invisible=True`?

### Steps to reproduce the bug.

``` python
import pyvista as pv
a=pv.Actor(mapper=pv.DataSetMapper(pv.Sphere()))
a.VisibilityOff()
p=pv.Plotter(off_screen=True)
p.add_actor(a)
p.show()

# Actual bounds:
print(p.renderer.bounds)
# (-0.49926671385765076, 0.49926671385765076, -0.4965316653251648, 0.4965316653251648, -0.5, 0.5)

# Expected bounds:
print(p.renderer.ComputeVisiblePropBounds())
# (1.0, -1.0, 1.0, -1.0, 1.0, -1.0)
```

### System Information

```shell
.
```

banesullivan:
Hm. You bring up a great point. I think we implemented it this way because I often use and invisible box/outline in the scene to manage the camera extent while viewing contours or thresholds.

Feel free to propose this change as a pull request!

user27182:
[...]

PR #7184, opened 2025-02-07
Use `PropCollection` to access `renderer.actors` internally

### LINKED ISSUE #6932: Using `import_vrml` to add actors to a scene does not report the correct bounds

See #6487 for details and for a possible initial implementation working towards a fix.

See #6487

### System Information

```shell
See CI in #6487
```

### Screenshots

_No response_

user27182:
### Overview

The `Renderer` currently stores its own `actors` internally as a dict. But this requires it to be manually sync'd with the `vtkRenderer`. This is bad from a maintenance perspective (have to remember to sync all the time), but also creates bugs when actors are added through other means, e.g.  `import_vrml`, `import_3ds`, `import_gltf` all add actors through vtk methods, meaning that `renderer.actors` is empty after the import (but should not be).

`vtkRenderer` already stores all props internally. This PR adds a new `PropCollection` class to provide a sequence/dict-like interface to the renderer's actors

This fixes #6932 and also brings us one step closer to fixing #5179. With this PR we can now always access all actors in the scene and therefore correctly compute the plotter's bounds using all the actors.


pyvista/plotting/renderer.py
```diff
@@ -880,21 +881,18 @@ def add_actor(
...
-
         if name is None:
-            # Fallback for non-wrapped actors
-            # e.g., vtkScalarBarActor
-            name = actor.name if isinstance(actor, Actor) else actor.GetAddressAsString('')
-
+            name = (
+                actor.name
+                if (hasattr(actor, 'name') and actor.name)
+                else actor.GetAddressAsString('')
```
user27182, line 888: Can change this to better match the naming used by `Actor` (and related) classes
```suggestion
                else f'{type(actor).__name__}({actor.GetAddressAsString('')})'
```

Naming used by actors:

https://github.com/pyvista/pyvista/blob/d1c369dd79b6ab4d92a10bdbd123b426aac0fa0c/pyvista/core/utilities/misc.py#L362-L367

tkoyama010: +1

tkoyama010 (approved):
LGTM

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

PR #7186, opened 2025-02-07
Add `NameMixin` class to give `name` attribute to actor-like classes

user27182:
### Overview

Part of #7184. Move `name` prop from `Actor` to its own class and add it to `Text`, `CornerAnnotation`, and `XYZAssembly` classes


tkoyama010 (approved):
LGTM

PR #7187, opened 2025-02-08
Disable caching for pushes to main branch and consolidate caching logic

user27182:
### Overview

The docs caching logic currently uses the cache for pushes to main and tags. This is bad for a few reasons:

- CI may fail when changes to doc images are made since a stale cache is used for pushes to main. See https://github.com/pyvista/pyvista/pull/7166. This effectively makes the caching useless, requiring all PRs to disable the cache manually else CI will fail
- A bad actor could potentially poison the cache in a PR, and if pushed to main or a tag ref this cache would be reused. This is a security issue (see https://github.com/pyvista/pyvista/pull/7006#issuecomment-2581761039)

EDIT: Caching for tags and releases was already disabled later on in workflow. This PR now moves that logic to a single location _and_ disables caching for pushes to main


user27182:
pre-commit.ci autofix

banesullivan (approved):

PR #7193, opened 2025-02-10
Fix zooming in terrain style with parallel projection

banesullivan:
I use terrain style interaction pretty much 100% of the time and had this code lying around to fix a few issues with zoom interactions under terrain style

- This enables scroll wheel zoom by default (this feels more intuitive to me)
- The scrolling factor was too sensitive/jumpy so I reduced it and changed it to actually dolly the camera instead of using the zoom function
- This adds support for zooming when parallel projection is enabled by using `parallel_scale`


pyvista/plotting/render_window_interactor.py
```diff
@@ -1013,16 +1017,29 @@ def enable_terrain_style(self, mouse_wheel_zooms=False, shift_pans=False):
...
 
         if mouse_wheel_zooms:
+            factor = 1.05 if isinstance(mouse_wheel_zooms, bool) else mouse_wheel_zooms
 
             def wheel_zoom_callback(_obj, event):  # pragma: no cover
                 """Zoom in or out on mouse wheel roll."""
                 if event == 'MouseWheelForwardEvent':
                     # zoom in
-                    zoom_factor = 1.1
+                    zoom_factor = 1 / factor
```
tkoyama010, line 1026:
```suggestion
                    zoom_factor = 1.0 / factor
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -975,11 +977,13 @@ def enable_terrain_style(self, mouse_wheel_zooms=False, shift_pans=False):
...
         ----------
-        mouse_wheel_zooms : bool, default: False
+        mouse_wheel_zooms : bool, default: True
             Whether to use the mouse wheel for zooming. By default
             zooming can be performed with right click and drag.
+            Pass a float value for to control the zoom factor,
+            default is 1.05.
 
-        shift_pans : bool, default: False
+        shift_pans : bool, default: True
```
tkoyama010, line 986: Could you add a `.. versionchanged::` directive? It would help clarify when this change was introduced.

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

PR #7197, opened 2025-02-11
Fix zizmore error: Bump version to v1.3.0

tkoyama010:
### Overview

Fix zizmore error: Bump version to v1.3.0

### Details

- ref #7170

### DIFF
--- a/.github/workflows/type-checking.yml
+++ b/.github/workflows/type-checking.yml
@@ -12,6 +12,8 @@ env:
 
 jobs:
   mypy:
+    permissions:
+      contents: read
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/update-pr-branch.yml
+++ b/.github/workflows/update-pr-branch.yml
@@ -6,6 +6,9 @@ on:
       - "main"
 jobs:
   autoupdate:
+    permissions:
+      contents: write
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - name: Automatically update PR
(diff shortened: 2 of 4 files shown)


user27182:
All cache-poisoning errors are ignored. These were triggered because caching is used for pushes to main, which should be avoided. To address this, a `USE_CACHE` env var is used to disable caching for some cases. This env var was copied from the docs workflow.

user27182 (approved):
LGTM. 
Note that this PR adds a few new untested `if` branches that will only evaluate to False after merge (e.g. to avoid using any cache for pushes to main). So there is some risk that this could break a workflow temporarily.

banesullivan (approved):

user27182:
> Note that this PR adds a few new untested `if` branches that will only evaluate to False after merge (e.g. to avoid using any cache for pushes to main). So there is some risk that this could break a workflow temporarily.

Looks like the docker workflow needs write permissions
https://github.com/pyvista/pyvista/actions/runs/13795556560/job/38586212822
`ERROR: denied: installation not allowed to Write organization package`

Fixed in #7299

banesullivan:
Wait, why did this PR add specific permisions for the `id-token` scope? This is for:

> Fetch an OpenID Connect (OIDC) token. This requires id-token: write

Why is this needed at all? ref https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#overview

banesullivan:
Oh. I see now. Zizmor wants these to be explicitly limited or it triggers an excessive permissions error

PR #7198, opened 2025-02-11
Disable VTK 9.4 snake_case API

user27182:
### Overview

VTK 9.4 includes a new pythonic API with snake_case attributes. This can't be disabled and the vtk package includes `pyi` files with type hints for the new snake_case API as well. This means dev tools will suggest both PyVista and VTK snake_case attributes/methods which can lead to users mistakenly using VTK methods when they may have assumed that these are PyVista attributes.

This PR disables VTK's API by raising an error if these are accessed.

~For now an error is always raised but we can add a global config option to convert this to a warning instead or disable it completely.~
`pv.vtk_snake_case` is also added for managing the global state


tkoyama010:
I love this idea. This will reduce our burden.

user27182:
> I love this idea. This will reduce our burden.

Great! This PR is ready as-is and was tested locally with vtk 9.4 and all the tests pass.

I could rebase it with the branch from #6879 to get CI passing here, or we can wait for that one to merge first, then patch this in afterwards

banesullivan:
This is a great solution to what I figured was unavoidable! Taking a Quick Look on my phone but I'll give this some testing soon here

banesullivan:
Great solution, considering this will error, we ought to allow end-users to control that with a global flag. 

In the meantime, I'd like for #6879 to merge before this

user27182:
> Great solution, considering this will error, we ought to allow end-users to control that with a global flag.
> 
> In the meantime, I'd like for #6879 to merge before this

Ok. I've rebased this PR to the branch from #6879. Will add the option to remove the error later after that PR has merged"

user27182:
This has a similar issue with rebasing as https://github.com/pyvista/pyvista/pull/7366#issuecomment-2794098069, it would be nice to remove the commits from #6879

banesullivan (approved):
+1 for this

Let's merge this for the same release that'll bring VTK 9.4 support

tkoyama010 (approved):

PR #7200, opened 2025-02-12
Add `'ids'` and `'all'` options to `MultiBlock.recursive_iterator`

user27182:
Add new options for iterating with block ids. This is useful for replacing blocks. 

See also #7202.

PR #7202, opened 2025-02-13
Add `MultiBlock.get_block` and extend `replace` to allow indexing nested blocks

user27182:
### Overview

Split from #7200.

The nested ids returned by `recursive_iterator` can be used with the nested replace features from this PR to replace nested blocks.

### DIFF
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -268,15 +268,6 @@ def test_replace_nested(nested_fixture, replace_indices):
     assert nested.flatten().keys() == expected_flat_keys
 
 
-def test_get_block(nested_fixture):
-    index = (1, 0)
-    name = 'grid'
-    block_by_index = nested_fixture[index[0]].get_block(index[1])
-    block_by_nested_index = nested_fixture.get_block(index)
-    block_by_name = nested_fixture[index[0]].get_block(name)
-    assert block_by_name is block_by_index is block_by_nested_index
-
-
 @pytest.mark.parametrize(
     'invalid_indices',
     [
@@ -289,6 +280,15 @@ def test_replace_nested_invalid_indices(nested_fixture, invalid_indices):
     match = re.escape(invalid_indices[1])
     with pytest.raises(IndexError, match=match):
         nested.replace(invalid_indices[0], None)
+
+
+def test_get_block(nested_fixture):
+    index = (1, 0)
+    name = 'grid'
+    block_by_index = nested_fixture[index[0]].get_block(index[1])
+    block_by_nested_index = nested_fixture.get_block(index)
+    block_by_name = nested_fixture[index[0]].get_block(name)
+    assert block_by_name is block_by_index is block_by_nested_index
 
 
 def test_reverse(sphere):
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -230,49 +230,49 @@ def test_slicing_multiple_in_setitem(sphere):
     assert len(multi) == 9
 
 
-def test_replace_nested():
+@pytest.fixture
+def nested_fixture():
     image = pv.ImageData()
     poly = pv.PolyData()
     grid = pv.UnstructuredGrid()
     nested = pv.MultiBlock(dict(image=image, poly=poly))
     multi = pv.MultiBlock(dict(grid=grid))
     nested.insert(1, multi, 'multi')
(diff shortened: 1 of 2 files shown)


tests/core/test_composite.py
```diff
@@ -230,6 +230,51 @@ def test_slicing_multiple_in_setitem(sphere):
...
+    assert nested.flatten().keys() == expected_flat_keys
+
+    match = re.escape('Invalid indices (0, 0, 0).')
+    with pytest.raises(IndexError, match=match):
+        nested.replace((0, 0, 0), None)
+    match = re.escape('Invalid indices (0, 0).')
+    with pytest.raises(IndexError, match=match):
+        nested.replace((0, 0), None)
+
+
```
tkoyama010, line 277: Thank you always for your amazing patches!
I rewrote the test code using `fixture` and `parametrize` to better understand your test intentions.
```suggestion
@pytest.fixture
def nested_fixture():
    image = pv.ImageData()
    poly = pv.PolyData()
    grid = pv.UnstructuredGrid()
    nested = pv.MultiBlock(dict(image=image, poly=poly))
    multi = pv.MultiBlock(dict(grid=grid))
    nested.insert(1, multi, 'multi')
    return nested

@pytest.mark.parametrize(
    "replace_indices",
    [
        ((0,)),
        ((1, 0),),
        ((2,),),
    ]
)
def test_replace_nested(nested_fixture, replace_indices):
    nested = nested_fixture
    expected_keys = ['image', 'multi', 'poly']
    expected_flat_keys = ['image', 'grid', 'poly']
    
    original_values = [nested[0], nested[1][0], nested[2]]
    nested.replace(replace_indices, None)
    
    assert all(
        (nested[idx] is None if idx == replace_indices else nested[idx] is original_values[i])
        for i, idx in enumerate([(0,), (1, 0), (2,)])
    )
    
    assert nested.keys() == expected_keys
    assert nested.flatten().keys() == expected_flat_keys

@pytest.mark.parametrize("invalid_indices", [
    ((0, 0, 0), 'Invalid indices (0, 0, 0).'),
    ((0, 0), 'Invalid indices (0, 0).'),
])
def test_replace_nested_invalid_indices(nested_fixture, invalid_indices):
    nested = nested_fixture
    match = re.escape(invalid_indices[1])
    with pytest.raises(IndexError, match=match):
        nested.replace(invalid_indices[0], None)


```

tkoyama010 (changes requested):

tests/core/test_composite.py
user27182: Thanks!

user27182:
pre-commit.ci autofix

tests/core/test_composite.py
```diff
@@ -230,6 +230,56 @@ def test_slicing_multiple_in_setitem(sphere):
...
+def test_replace_nested(nested_fixture, replace_indices):
+    nested = nested_fixture
+    expected_keys = ['image', 'multi', 'poly']
+    expected_flat_keys = ['image', 'grid', 'poly']
+
+    original_values = [nested[0], nested[1][0], nested[2]]
+    nested.replace(replace_indices, None)
+
+    assert all(
+        (nested[idx] is None if idx == replace_indices else nested[idx] is original_values[i])
```
user27182, line 261: This relies on `__getitem__` which does not allow nested indices. And this kind of indexing is not allowed by packages like NumPy for ragged arrays, so I don't think we should allow it either. But we can probably overload `get` to allow getting by index or nested index (it current only gets by name).

user27182: The tests were needlessly complex and asserting that a single block was changed while also asserting that all other blocks were _not_ changed. If we only check the changed block everything is easier to read and understand.

This is done in [ab44de4](https://github.com/pyvista/pyvista/pull/7202/commits/ab44de41e60e5bbbb4472b5228f87648b5974186). `get_block`

> we can probably overload `get` to allow getting by index or nested index (it current only gets by name).

Overloading get isn't great because it returns a default value. But here we want to raise an IndexError and not return a default. So I've added a new `get_block` method to get a block from a nested index.

user27182: Tests should be passing now, this should be resolved. We can split this into a separate PR for `get_block` or rename title to better reflect the changes.

tkoyama010 (approved):
LGTM

PR #7203, opened 2025-02-13
Add generic composite filter to apply any filter to `MultiBlock` blocks

user27182:
### Overview

Use the new `MultiBlock.recursive_iterator` to apply filters generically.

~The filter is currently private as the API may change. E.g. there are some options specific to `CompositeFilters` such as `nested` which would be good to include, but this may require updates to the iterator or a new helper method.~
This filter is public.

Note: this PR depends on #7200  and #7202

### DIFF
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -1149,27 +1149,46 @@ def test_flatten_copy(multiblock_all, copy):
     assert shares_memory == (not copy)
 
 
-def test_generic_filter(multiblock_all_with_nested_and_none):
+@pytest.mark.parametrize(
+    'function', [lambda x: x.cast_to_unstructured_grid(), 'cast_to_unstructured_grid']
+)
+def test_generic_filter(multiblock_all_with_nested_and_none, function):
+    # Include empty mesh
+    empty_mesh = pv.PolyData()
+    multiblock_all_with_nested_and_none.append(empty_mesh)
+
+    output = multiblock_all_with_nested_and_none.generic_filter(function)
+    flat_outputs = output.flatten()
+    # Make sure no `None` blocks were removed
+    assert None in flat_outputs
+    # Check output
+    for block in flat_outputs:
+        assert isinstance(block, pv.UnstructuredGrid) or block is None
+
+
+@pytest.mark.parametrize('inplace', [True, False])
+def test_generic_filter_inplace(multiblock_all_with_nested_and_none, inplace):
     # Include empty mesh
     empty_mesh = pv.PolyData()
     multiblock_all_with_nested_and_none.append(empty_mesh)
     flat_inputs = multiblock_all_with_nested_and_none.flatten()
-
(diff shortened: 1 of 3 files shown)


tests/core/test_composite.py
```diff
@@ -1017,41 +1029,66 @@ def test_recursive_iterator_prepend_names(separator, prepend_names):
...
+
+    # Test names
+    assert names0 == expected_names[0]
+    assert names1 == expected_names[1]
+    assert names2 == expected_names[2]
+
+    # Test blocks
+    assert nested.get_block(ids0) is locals()[names0]
+    assert nested.get_block(ids1) is locals()[names1]
+    assert nested.get_block(ids2) is locals()[names2]
```
tkoyama010, line 1091: I modified it so that the `for` loop can be used without explicitly calling `next(iterator)`.
```suggestion
def test_recursive_iterator_order(nested_fixture, order, expected_ids, expected_names):
    nested = nested_fixture
    common_kwargs = dict(skip_empty=False, nested_ids=True, contents='all')

    iterator = nested.recursive_iterator(order=order, **common_kwargs)

    for i, (ids, names, block) in enumerate(iterator):
        # Test ids
        assert ids == expected_ids[i]

        # Test names
        assert names == expected_names[i]

        # Test blocks
        assert nested.get_block(ids) is locals()[names]
```

tests/core/test_composite.py
```diff
@@ -1017,41 +1029,66 @@ def test_recursive_iterator_prepend_names(separator, prepend_names):
...
+        ('nested_first', [(1, 0), (0,), (2,)], ['grid', 'image', 'poly']),
+        ('nested_last', [(0,), (2,), (1, 0)], ['image', 'poly', 'grid']),
+        (None, [(0,), (1, 0), (2,)], ['image', 'grid', 'poly']),
+    ],
+)
+def test_recursive_iterator_order(nested_fixture, order, expected_ids, expected_names):
+    nested = nested_fixture
+    image = nested_fixture['image']
+    poly = nested_fixture['poly']
+    grid = nested_fixture['multi']['grid']
```
tkoyama010, line 1070: Could you explain the intention behind defining these variables?

tkoyama010 (changes requested):

tests/core/test_composite.py
user27182: This is used later for testing the blocks with `locals()`. This works because the name of the local variable matches the block name. We could alternatively add the expected instance type as part of the parametrization and remove these definitions (and remove the use of locals), though checking the object directly with `is` is a stricter test.

tests/core/test_composite.py
user27182: Thanks, this is nicer.

Also, note that these tests are duplicated from #7200. They're included here because this PR depends on that one. So we can probably update those tests directly instead as part of that PR (and I'll work through any merge conflicts for this PR's tests)

The tests for `generic_filter` explicitly are the only new tests that are specific to this PR

tkoyama010 (approved):
LGTM!

PR #7206, opened 2025-02-16
Compute visible bounds by default

### LINKED ISSUE #5179: `renderer.bounds` includes invisible actors

user27182:
### Overview

Fix #5179 

- add `use_bounds` property to `Actor`
- `bounds` is now computed directly with `renderer.ComputeVisiblePropBounds`
- The new `compute_bounds` function uses the old code to manually compute the bounds. But all special cases for vtkLightActor, vtkCubeAxesActor and bounding_box_actor have been removed. `use_bounds` and `visibility` properties can be overriden and specific actors can be removed from bounds explicitly.


pyvista/plotting/renderer.py
```diff
@@ -485,8 +485,75 @@ def camera(self, source) -> None:
...
+        BoundsTuple
+            Bounds of all visible actors in the active renderer.
+
+        See Also
+        --------
+        compute_bounds
+            Compute the bounds with options to enable or disable actor visibility.
+
+        """
+        return self.compute_bounds()
```
user27182, line 509: Both `bounds` and `compute_bounds` return the same bounds by default. Should we compute the bounds of ALL actors (not just visible ones) by default for `compute_bounds` just to differentiate them?

banesullivan:
Big +1 for this from me! Thanks for doing this!

I'll take a deeper look and provide a review shortly

pyvista/plotting/renderer.py
```diff
@@ -485,8 +485,76 @@ def camera(self, source) -> None:
...
+        """
+        bounds = self.ComputeVisiblePropBounds()
+        return _fixup_bounds(bounds)
+
+    def compute_bounds(
+        self,
+        *,
+        force_visibility: bool = False,
+        force_use_bounds: bool = False,
+        ignore_actors: Sequence[str | _vtk.vtkProp | type[_vtk.vtkProp]] | None = None,
```
banesullivan, line 517: Should we make sure to set use_bounds to False on specific internal actors like the axes actors and perhaps certain widgets? One reason for me putting the `ignored_actors` in the example in https://github.com/pyvista/pyvista/issues/5179#issuecomment-2647287991 was to make sure that the bounds were ignoring a plane widget I have in my scene

The other thing I did in my code was to separate the notion of "dataset actors" -- basically the check goes:

```py
def is_likely_dataset_actor(actor: pv.Actor | Any):
    """Try to determine if the actor has an associated dataset without triggering a pipeline execution."""
    if not isinstance(actor, pv.Actor):
        # ignore anything that is not wrapped as a PyVista Actor
        return False
    if not hasattr(actor, "mapper"):
        return False
    if not isinstance(actor.mapper, pv.DataSetMapper) or not hasattr(actor.mapper, "GetInputAsDataSet"):
        # The most surefire way to do this is with `hasattr(actor.mapper, "dataset")`
        # but that will trigger a pipeline execution, or we could check the input producer is not none
        return False
    return True  # likely a dataset actor
```

user27182: > Should we make sure to set use_bounds to False on specific internal actors like the axes actors (and perhaps certain widgets... or maybe widgets don't get actors I forget right now)?

My understanding is that this is not necessary. I am not sure if there are specific tests for this, but based on the fact that all image regression testing on plotting tests and all the docs are passing after changing the bounds calculation, this indicates to me that the plots are generated as expected/desired. There is one change to a single image which involves CubeAxesActor, but this change to me seems like an improvement (the axes are now tighter to the teapot).

Also, setting `use_bounds` to off for something like CubeAxesActor seems unintuitive... if I add cube axes, don't I want to be sure that all the axes of the cube are visible? Disabling the bounds would risk them only being partly visible, not sure what use cases would make this desirable. And having these bounds included doesn't seem to break any plots for our tests currently.

> The other thing I did in my code was to separate the notion of "dataset actors" so that bounds were only computed from visible datasets -- basically the check goes:
> 
> ```python
> def is_likely_dataset_actor(actor: pv.Actor | Any):
>     """Try to determine if the actor has an associated dataset without triggering a pipeline execution."""
>     if not isinstance(actor, pv.Actor):
>         # ignore anything that is not wrapped as a PyVista Actor
>         return False
>     if not hasattr(actor, "mapper"):
>         return False
>     if not isinstance(actor.mapper, pv.DataSetMapper) or not hasattr(actor.mapper, "GetInputAsDataSet"):
>         # The most surefire way to do this is with `hasattr(actor.mapper, "dataset")`
>         # but that will trigger a pipeline execution, or we could check the input producer is not none
>         return False
>     return True  # likely a dataset actor
> ```

I am not sure what you mean by `visible` datasets, the new code for `compute_bounds` checks `visibility` explicitly.

And I disagree with this line:
 >         # ignore anything that is not wrapped as a PyVista Actor
There are some methods which add a valid vtkActor what we want to show that isn't necessarily wrapped. See #6932 which was recently fixed by #7184.

IMO if you want to show pyvista-wrapped actors only then we should add the option to manually specify this. E.g. something like `ignore_actors` but inverted, so you can do `compute_bounds(only_show=[pv.Actor])` and allow listing by name, type, or instance.

Basically I don't think any special cases should be baked into `compute_bounds` directly. If you don't want it computed, use `visbility`, `use_bounds`, or the `ignore_actors` options explicitly.

user27182: Another thing to note from #5179 is that ResetCamera and ResetClippingRange call ComputeVisiblePropBounds internally. So if we want a specialized default computation for computing bounds that deviates from ComputeVisiblePropBounds, then we need to also ensure that any method that resets the camera also recomputes the bounds in our specialized way to ensure that any special cases are properly considered there too. 

Otherwise, the bounds may be unexpectedly computed differently for some cases where camera resets are involved but not others

user27182: > if you want to show pyvista-wrapped actors only then we should add the option to manually specify this. E.g. something like `ignore_actors` but inverted, so you can do `compute_bounds(only_show=[pv.Actor])` and allow listing by name, type, or instance.

If this kind of feature is helpful for moving this PR forward let me know and I can try adding it

banesullivan: > If this kind of feature is helpful for moving this PR forward let me know and I can try adding it

I don't think we should do this just yet... the code I have for this is rather niche where I specifically want to ignore anything that isn't a pv.Actor (feels application specific to me)

> Another thing to note from https://github.com/pyvista/pyvista/issues/5179 is that ResetCamera and ResetClippingRange call ComputeVisiblePropBounds internally. So if we want a specialized default computation for computing bounds that deviates from ComputeVisiblePropBounds, then we need to also ensure that any method that resets the camera also recomputes the bounds in our specialized way to ensure that any special cases are properly considered there too.

This is a good point and a good reason to just stay consistent with ComputeVisiblePropBounds

I think having the `use_bounds` property alleviates my specific concerns and glad to have learned about this property

banesullivan (approved):
I'm really happy with the way this is done! Thanks and nice work here, @user27182!

PR #7210, opened 2025-02-18
Move common dataset/composite filters to `DataObjectFilters` class

user27182:
### Overview

Continuation from #6940. 

The scope of this PR is limited to filters in `composite.py` which are directly mapped to `DataSetFilters`. Other common filters (e.g. `outline`, `transform`) will be moved/addressed separately.

These filters were explicitly defined in `CompositeFilters` with mappings to `DatSetFilters`:
- clip
- clip_box
- slice
- slice_orthogonal
- slice_along_axis
- slice_along_line
- extract_all_edges
- elevation
- compute_cell_sizes
- cell_centers
- cell_data_to_point_data
- point_data_to_cell_data
- sample
- triangulate

These filters are moved from `DataSetFilters` because they're used internally by the clip/slice filters:
- _clip_with_function
- slice_implicit

These are also moved from `DataSetFilters` because they're aliases:
- ctp
- ptc


user27182:
Both this PR and #7211 add a new `DataObjectFilters` class, and either PR can be merged first. And for this PR #7217 should be reviewed/merged first.

tkoyama010 (changes requested):
https://github.com/pyvista/pyvista/pull/7210#issuecomment-2670051548

PR #7211, opened 2025-02-18
Move all transformation filters to `DataObjectFilters` class

user27182:
### Overview

Related to #7210.

This is partly maintenance but is also a feature because it enables all transformation methods to be used with `MultiBlock` (currently, only `MultiBlock.transform` is available). This PR:

- Removes the current `CompositeFilters.transform` filter and its implementation
- Moves all `DataSetFilters` transformations to `DataObjectFilters`
- Re-implements the functionality of `MultiBlock.transform` using the new generic filter #7203 as part of `DataObjectFilters.transform`


tkoyama010 (approved):
LGTM

tkoyama010 (changes requested):
https://github.com/pyvista/pyvista/pull/7210#issuecomment-2670051548

user27182:
> LGTM

Thanks for the reviews! 🙏
This PR also includes the refactoring of MultiBlock.transform from #7222, perhaps we should merge that one first

tkoyama010 (approved):

PR #7213, opened 2025-02-18
Remove unnecessary Updates in nek5000Reader and ExodusIIReader

mattfalcone1997:
### Overview
This PR is a small update to remove some `Update()` calls from the Nek5000Reader and ExodusIIReader. I believe that these are unnecessary and caused some extra overhead particularly in the former where it was called when the time was set, which is done on class initialisation. This meant that if the time was changed before `read()` then `Update() would be called twice when only required once, unless I have misunderstood the purpose of update in reader classes.

### DIFF
--- a/pyvista/core/utilities/reader.py
+++ b/pyvista/core/utilities/reader.py
@@ -2205,8 +2205,6 @@ class Nek5000Reader(BaseReader, PointCellDataSelection, TimeReader):
         vtkinfo = self.reader.GetExecutive().GetOutputInformation(0)
         vtkinfo.Set(key, time_value)
 
-        self.reader.Update()
-
     def set_active_time_point(self, time_point):
         """Set active time or iteration by index.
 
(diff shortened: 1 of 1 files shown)


tkoyama010 (approved):
Thank you! As long as the test passes, it's good to go!

mattfalcone1997:
> Thank you! As long as the test passes, it's good to go!

The rest are failing on the triage with the error 'Error: Parameter token or opts.auth is required'. I'm not really sure how to fix this. Do the tests just need to be re run now a label has been applied?

tkoyama010 (approved):

tkoyama010:
> > Thank you! As long as the test passes, it's good to go!
> 
> The rest are failing on the triage with the error 'Error: Parameter token or opts.auth is required'. I'm not really sure how to fix this. Do the tests just need to be re run now a label has been applied?

It's a problem of CI. Please see #7215.

PR #7214, opened 2025-02-19
Run `test_charts.py` in a separate process and increase `xvfb` max clients

user27182:
### Overview

~Skip tests for charts since these are causing issues.~
Run tests for charts in a separate process and increase xvfb maxclients

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -231,9 +231,8 @@ jobs:
 
       - name: Plotting Testing (uses GL)
         run: |
-          PYTEST_CMD="xvfb-run -s '-maxclients 2048' -a python -m pytest"
-
-          COMMON_ARGS="
+          PYTEST_CMD=(python -m pytest)
+          COMMON_ARGS=(
             -v
             --cov=pyvista
             --cov-append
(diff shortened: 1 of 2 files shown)


.github/workflows/testing-and-deployment.yml
```diff
@@ -230,7 +230,29 @@ jobs:
...
+          xvfb-run -s '-maxclients 2048' -a "${PYTEST_CMD[@]}" \
+            --ignore=tests/core \
+            --ignore=tests/examples \
+            --ignore=tests/plotting/test_charts.py \
+            "${COMMON_ARGS[@]}"
+          
+          # Run charts tests
+          xvfb-run -s '-maxclients 2048' -a "${PYTEST_CMD[@]}" \
+            tests/plotting/test_charts.py \
+            "${COMMON_ARGS[@]}"
```
banesullivan, line 255: My preference would be for us to split these into separate steps so that it is more clear on GitHub Actions which tests are failing. And we could set this so that the chart tests still run if the above plotting tests fail to get a comprehensive picture of testing status

user27182: I tried this in https://github.com/pyvista/pyvista/pull/7214/commits/822922b78501751fee0bbf2bb8519afc79f3d84f, let's see if it works.

Do you want based from main or maint/vtk-override? VTK 9.4 is not strictly needed for this change

user27182: New separate steps are working, and the seg fault is gone. The Linux charts tests are now only failing the same way the MacOS tests are failing.

PR #7220, opened 2025-02-20
Add `MultiBlock.as_unstructured_grid_blocks`

user27182:
### Overview

Similar to `MultiBlock.as_polydata_blocks`


pyvista/core/composite.py
```diff
@@ -1772,6 +1774,49 @@ def block_filter(block: DataSet | None) -> PolyData:
...
+    ) -> MultiBlock:
+        """Convert all the datasets within this MultiBlock to :class:`~pyvista.UnstructuredGrid`.
+
+        Parameters
+        ----------
+        copy : bool | 'deep' | 'shallow', default: False
+            Option to create a deep or shallow copy of any datasets that are already a
+            :class:`~pyvista.UnstructuredGrid`. When ``False``, any datasets that are
+            already UnstructuredGrid will not be copied. If ``True`` or ``'deep'``, a
+            deep copy is made. Use ``'shallow'`` for a shallow copy.
```
user27182, line 1788: This API differs from `as_polydata_blocks`. That method uses `True` for a shallow copy. It does not do deep copy.

Not sure if we should align and only allow shallow copies here, or optionally allow deep copies there.

pyvista/core/composite.py
```diff
@@ -1853,6 +1855,46 @@ def block_filter(block: DataSet | None) -> PolyData:
 
         return self.generic_filter(block_filter, _skip_none=False)
 
+    def as_unstructured_grid_blocks(
+        self: MultiBlock, copy: bool | Literal['deep', 'shallow'] = False
```
user27182, line 1859:
```suggestion
        self: MultiBlock, copy: bool = False
```

tkoyama010 (dismissed):
LGTM

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):

PR #7221, opened 2025-02-20
Do not skip ``None`` or empty blocks by default with `recursive_iterator`

user27182:
### Overview

The default to skip ``None`` and ``empty`` blocks can be problematic. Sometimes it's helpful to exclude them but in general cases it's desirable to include all blocks. I think it's best to always include these, and opt-out explicitly.

`generic_filter` is also updated to allow `None` blocks, but these are still skipped by default.
Allowing None blocks here is needed for general cases, e.g. see #7223 and #7220

### DIFF
--- a/pyvista/core/composite.py
+++ b/pyvista/core/composite.py
@@ -181,8 +181,8 @@ class MultiBlock(
         contents: Literal['ids', 'names', 'blocks', 'items', 'all'] = 'blocks',
         order: Literal['nested_first', 'nested_last'] | None = None,
         *,
-        skip_none: bool = True,
-        skip_empty: bool = True,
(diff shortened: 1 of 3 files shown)


user27182:
This should be ready now. I found a few minor bugs with the function name used in the error message for `generic_filter`. Fixes for that are also included here, but I can create a separate PR to fix that if needed

user27182:
>  I found a few minor bugs with the function name used in the error message for `generic_filter`. Fixes for that are also included here, but I can create a separate PR to fix that if needed

These changes have been reverted. The fix is applied in #7227 instead.

tkoyama010 (approved):
LGTM

PR #7222, opened 2025-02-20
Refactor `MultiBlock.transform` to use `generic_filter`

user27182:
### Overview

~The transform filter currently uses `__setitem__` instead of `replace` to replace blocks. This means that block names are lost when transforming.~
EDIT: After review I'm not sure this is true as a copy currently is made and blocks are replaced in-place. But I still think it's good to refactor this.

Use the `generic_filter` instead which handles block replacements correctly.


tkoyama010 (approved):
LGTM

PR #7223, opened 2025-02-20
Refactor `as_polydata_blocks` to use `generic_filter`

user27182:
### Overview

Use the `generic_filter` internally to simplify the code.

Cross-refs are also added for related functions.


user27182:
This should be good to go but depends on #7221 so we should wait for that first

tkoyama010:
Is this ready to go?

tkoyama010 (approved):
LGTM

Issue #7224, opened 2025-02-20
Performance regression of `contour_labeled`


user27182:
### Describe what maintenance you would like added.

`contour_labeled` is deprecated and replaced with `contour_labels`. The new implementation works better but is much slower. This was partly suspected (see https://github.com/pyvista/pyvista/pull/6732#discussion_r1869872264) but I have now profiled the differences:

``` python
from pyinstrument import Profiler
from pyvista import examples
tissues = examples.load_frog_tissues()
profiler = Profiler()
profiler.start()
for _ in range(10):
    tissues.contour_labels()
profiler.stop()
print(profiler.output_text(unicode=True, color=True))
```

Result with `contour_labels` is approx 11 seconds, with `compute_normals` and `np.unique` as bottlenecks
``` python
11.048 <module>  <ipython-input-3-fa1d9619d8a8>:1
└─ 11.023 ImageData.contour_labels  pyvista/core/filters/image_data.py:1013
   ├─ 5.314 PolyData.compute_normals  pyvista/core/filters/poly_data.py:1770
   │  └─ 5.302 _update_alg  pyvista/core/filters/__init__.py:39
   │     └─ 5.302 vtkPolyDataNormals.Update  <built-in>
   ├─ 2.483 _update_alg  pyvista/core/filters/__init__.py:39
   │  └─ 2.483 vtkSurfaceNets3D.Update  <built-in>
   ├─ 2.435 _configure_boundaries  pyvista/core/filters/image_data.py:1439
   │  └─ 2.435 _get_unique_labels_no_background  pyvista/core/filters/image_data.py:1392
   │     └─ 2.435 unique  numpy/lib/_arraysetops_impl.py:145
   │           [2 frames hidden]  numpy
   │              2.292 pyvista_ndarray.sort  <built-in>
   └─ 0.654 PolyData.remove_cells  pyvista/core/pointset.py:117
      └─ 0.651 PolyData.RemoveGhostCells  <built-in>
```

With `contour_labeled` this is 10x faster:
``` python
0.923 <module>  <ipython-input-2-43ca39a3cc1b>:1
└─ 0.921 ImageData.contour_labeled  pyvista/core/filters/image_data.py:846
   └─ 0.921 _update_alg  pyvista/core/filters/__init__.py:39
      └─ 0.921 vtkSurfaceNets3D.Update  <built-in>
```

PR #7225, opened 2025-02-21
Add `'strict_external'` boundary style to `contour_labels`

user27182:
### Overview

Add ~`fast_mode`~ `'strict_external'` boundary style to avoid expensive calls to `np.unique` and `remove_cells`.

This is a partial fix for #7224. When smoothing and normals are also disabled, we get:
``` python
from pyinstrument import Profiler
from pyvista import examples

tissues = examples.load_frog_tissues()
profiler = Profiler()
profiler.start()

for _ in range(10):
    tissues.contour_labels('strict_external', smoothing=False, compute_normals=False)

profiler.stop()
print(profiler.output_text(unicode=True, color=True))
```

Output:
``` python
1.026 <module>  <ipython-input-4-f08a2dddc178>:1
├─ 1.002 ImageData.contour_labels  pyvista/core/filters/image_data.py:1013
│  ├─ 0.953 _update_alg  pyvista/core/filters/__init__.py:39
│  │  └─ 0.953 vtkSurfaceNets3D.Update  <built-in>
│  └─ 0.047 ImageData.pad_image  pyvista/core/filters/image_data.py:2144
│     └─ 0.046 _get_padded_output  pyvista/core/filters/image_data.py:2449
│        └─ 0.045 _update_and_get_output  pyvista/core/filters/image_data.py:2455
│           └─ 0.045 _update_alg  pyvista/core/filters/__init__.py:39
│              └─ 0.045 vtkImageConstantPad.Update  <built-in>
└─ 0.024 [self]  <ipython-input-4-f08a2dddc178>
```

This computation takes approx 1 second which is on par with `contour_labeled` timing reported in #7224.
The use of `compute_normals=True` by default will be re-evaluated in a separate PR.


pyvista/core/filters/image_data.py
```diff
@@ -1019,6 +1020,7 @@ def contour_labels(  # type: ignore[misc]
         select_outputs: int | VectorLike[int] | None = None,
         pad_background: bool = True,
         output_mesh_type: Literal['quads', 'triangles'] | None = None,
+        fast_mode: bool = False,
```
tkoyama010, line 1023: This is a great idea, and I completely agree that it contributes to speeding things up. However, from a user's perspective, rather than emphasizing speed, it’s more beneficial for code clarity to clearly show what’s happening internally. Therefore, I would like to propose the variable name `remove_cell`.
```suggestion
        remove_cell: bool = False,
```

tkoyama010 (changes requested):

pyvista/core/filters/image_data.py
user27182: What about `ignore_internal_boundary_effects` ?

It's more complicated than just removing cells. It's including internal cells and then removing them for the purpose of improving how the external cells are generated.

With the new option there are effectively two versions of `'external'` boundaries. So how can we easily differentiate the two? My concern with using language that refers to implementation details is that it might appear contradictory. E.g. doing 

`mesh.contour_labels('external', ignore_internal_boundary_effects=True)` 

May seem odd because both the term internal and external appear.

user27182: What about removing the new `fast_mode` keyword and adding a new `'strict_external'` boundary style instead?

tkoyama010: +100 strict_external

pyvista/core/filters/image_data.py
```diff
@@ -1545,12 +1574,17 @@ def _is_small_number(num) -> bool | np.bool_:
...
-        _configure_boundaries(
-            alg,
-            cast(pyvista.pyvista_ndarray, alg_input.active_scalars),
-            select_inputs,
-            select_outputs,
-        )
+        if boundary_style == 'strict_external':
+            # Use default alg parameters
+            if select_inputs is not None or select_outputs is not None:
+                raise TypeError('Selecting inputs and/or outputs is not supported by `fast_mode`.')
```
tkoyama010, line 1580:
```suggestion
                raise TypeError('Selecting inputs and/or outputs is not supported by `strict_external`.')
```

tests/core/test_imagedata_filters.py
```diff
@@ -371,6 +372,28 @@ def test_contour_labels_cell_data(channels):
...
+    start = time.perf_counter()
+    contours = channels.contour_labels('strict_external', compute_normals=False)
+    time_fast = time.perf_counter() - start
+    assert time_fast < time_slow / 1.5
+
+    # Test output is simplified correctly
+    assert contours.active_scalars.ndim == 1
+    assert np.all(contours.active_scalars > 0)
+
+    match = 'Selecting inputs and/or outputs is not supported by `fast_mode`.'
```
tkoyama010, line 390:
```suggestion
    match = 'Selecting inputs and/or outputs is not supported by `strict_external`.'
```

tkoyama010 (changes requested):

user27182:
pre-commit.ci autofix

user27182:
Looks like one of the tests from this PR is flaky:

https://github.com/pyvista/pyvista/actions/runs/13549252342/job/37868611724?pr=7248#step:6:7300
``` python
FAILED tests/core/test_imagedata_filters.py::test_contour_labels_strict_external - assert 0.6273909660000072 < (0.7543209680000018 / 1.5)
```

PR #7227, opened 2025-02-21
Fix function name in error messages for `MultiBlock.generic_filter`

user27182:
### Overview

The error message for `MultiBlock.generic_filter` isn't helpful when a generic function is used since it currently says `'function'` could not be applied instead of using the function's name. This PR fixes this.


tkoyama010 (approved):
LGTM

PR #7228, opened 2025-02-21
Rework `transform_all_input_vectors` as `mode` parameter for `Transform.apply`

user27182:
### Overview

The `transform_all_input_vectors` keyword is dataset-specific, which is not ideal, and `apply` is also missing an option to transform vectors. See https://github.com/pyvista/pyvista/issues/7097#issuecomment-2600913235 for context.

This PR removes `transform_all_input_vectors` and adds `mode` instead. `'points'` mode transforms points, `'vectors'` mode transforms vectors, and `'all_vectors'` enables `transform_all_input_vectors` for datasets.

### DIFF
--- a/pyvista/core/utilities/transform.py
+++ b/pyvista/core/utilities/transform.py
@@ -1435,30 +1435,30 @@ class Transform(_vtk.vtkTransform):
         self: Transform,
         obj: VectorLike[float] | MatrixLike[float],
         /,
+        mode: Literal['points', 'vectors'] | None = ...,
         *,
         inverse: bool = ...,
         copy: bool = ...,
-        transform_all_input_vectors: bool = ...,
     ) -> NumpyArray[float]: ...
     @overload
     def apply(
         self: Transform,
         obj: ConcreteDataSetType,
         /,
+        mode: Literal['all_vectors'] | None = ...,
         *,
         inverse: bool = ...,
         copy: bool = ...,
-        transform_all_input_vectors: bool = ...,
     ) -> ConcreteDataSetType: ...
     @overload
     def apply(
         self: Transform,
         obj: MultiBlock,
         /,
+        mode: Literal['all_vectors'] | None = ...,
         *,
         inverse: bool = ...,
         copy: bool = ...,
-        transform_all_input_vectors: bool = ...,
     ) -> MultiBlock: ...
     def apply(
         self: Transform,
--- a/pyvista/core/utilities/transform.py
+++ b/pyvista/core/utilities/transform.py
@@ -1464,12 +1464,12 @@ class Transform(_vtk.vtkTransform):
         self: Transform,
         obj: VectorLike[float] | MatrixLike[float] | ConcreteDataSetType | MultiBlock,
         /,
+        mode: Literal['points', 'vectors', 'all_vectors'] | None = None,
         *,
         inverse: bool = False,
         copy: bool = True,
-        transform_all_input_vectors: bool = False,
     ):
-        """Apply the current transformation :attr:`matrix` to points or a dataset.
+        """Apply the current transformation :attr:`matrix` to points, vectors, or a dataset.
 
         .. note::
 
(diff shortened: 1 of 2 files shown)


user27182:
@darikg your feedback on these changes would be nice to have.

darikg:
There's four possible calling conventions, right?

```python
transform.apply(array_of_points, mode='points')
transform.apply(array_of_vectors, mode='vectors')
transform.apply(dataset)
transform.apply(dataset, mode='all_vectors')
```
The fact that the other combinations like `transform.apply(array_of_vectors, mode='all_vectors')` or `transform.apply(dataset, mode='points')` aren't meaningful still seems awkward to me, but I see that this is captured in the `@overload` typing so mypy should be able to flag those as incorrect. I guess the other approach would be something like 

```python
transform.transform_points(array_of_points)
transform.transform_vectors(array_of_vectors)
transform.transform_dataset(dataset, all_vectors: bool = False)
```

I think I prefer this direction more but it's pretty subjective

user27182:
> There's four possible calling conventions, right?
> 
> ```python
> transform.apply(array_of_points, mode='points')
> transform.apply(array_of_vectors, mode='vectors')
> transform.apply(dataset)
> transform.apply(dataset, mode='all_vectors')
> ```

Yes and the keyword `mode` is optional, and for arrays it defaults to points.
So these are equivalent:
``` python
transform.apply(array_of_points)
transform.apply(array_of_points, 'points')
```

> The fact that the other combinations like `transform.apply(array_of_vectors, mode='all_vectors')` or `transform.apply(dataset, mode='points')` aren't meaningful still seems awkward to me

I agree this is not ideal... though this kind of pattern where one parameter depends on another and is not valid or used otherwise is common with our broader API and with VTK in general. At least this way the `mode` parameter is used by all input types, unlike before where `transform_all_input_vectors` was definitively dataset-specific.

> but I see that this is captured in the `@overload` typing so mypy should be able to flag those as incorrect. 

A `ValueError` is also raised at runtime if you use the wrong mode for the wrong data type.

I guess the other approach would be something like
> 
> ```python
> transform.transform_points(array_of_points)
> transform.transform_vectors(array_of_vectors)
> transform.transform_dataset(dataset, all_vectors: bool = False)
> ```
> 
> I think I prefer this direction more but it's pretty subjective

This is what [`vtkAbstractTransform`](https://vtk.org/doc/nightly/html/classvtkAbstractTransform.html) does.

In the context of #7097 , the scipy objects use `apply`. Though, in the `RigidTransform` case, their version of apply does not handle points/vectors differently, and of course they don't have scipy dataset objects like we do...

So is it better to be more vtk-like and explicit and have three separate methods? Or align more with scipy and use `apply` but overload the behavior more implicitly?

user27182:
What about having both methods as aliases? I think it's beneficial to align with `scipy` and keep `apply`. But explicit and more VTK-like can also be useful. We don't necessarily have to keep only one or the other.

user27182:
> ```python
> transform.transform_points(array_of_points)
> transform.transform_vectors(array_of_vectors)
> transform.transform_dataset(dataset, all_vectors: bool = False)
> ```
> 
> I think I prefer this direction more but it's pretty subjective

These are now included with this PR:

- `apply_to_points`
- `apply_to_vectors`
- `apply_to_dataset`

The term "apply" is used to be consistent with the more general `apply` method, plus I find `Transform.transform ...` to be a bit redundant anyway.

pyvista/core/utilities/transform.py
```diff
@@ -1576,6 +1601,147 @@ def apply(
...
+        """
+        return self.apply(vectors, 'vectors', inverse=inverse, copy=copy)
+
+    def apply_to_dataset(
+        self,
+        dataset: ConcreteDataSetOrMultiBlockType,
+        /,
+        copy: bool = True,
+        inverse: bool = False,
+        transform_all_input_vectors: bool = False,
```
user27182, line 1698: Should we rename this  to `all_vectors` to be self-consistent with `apply(mesh, 'all_vectors')` and with this comment https://github.com/pyvista/pyvista/pull/7228#issuecomment-2675543652?

user27182:
I'm now leaning towards removing apply in favor of the specialized methods. 

I'm working through #7279 and finding that transforming an actor is more complex because it already has its own user matrix / user transform. So could benefit from having modes like `pre-multiply`, `post-multiply` to modify the transform, or even a `set` mode to set the transform to a new one. 

In this case it would probably be best to have a dedicated method, and avoid combining it with apply. By extension , maybe it's best to do that with all input types.

user27182:
> I'm now leaning towards removing apply in favor of the specialized methods.
> 
> I'm working through #7279 and finding that transforming an actor is more complex because it already has its own user matrix / user transform. So could benefit from having modes like `pre-multiply`, `post-multiply` to modify the transform, or even a `set` mode to set the transform to a new one.
> 
> In this case it would probably be best to have a dedicated method, and avoid combining it with apply. By extension , maybe it's best to do that with all input types.

The API for transforming an actor has been fleshed out in #7279, and is identical to transforming datasets and arrays _except_ that different modes are allowed depending on the type. Overall, I think this works quite nicely. Although all of the modes are input-specific without overlap, this is clearly documented and helpful errors are otherwise raised if an invalid mode is used.

And so I think we should keep the general `apply` method after all. I'll update the main comment accordingly.

Waiting for green CI but should be ready for review.

tkoyama010 (approved):
LGTM

PR #7230, opened 2025-02-23
Add `select_values` filter for `ImageData`

user27182:
### Overview

This filter is very similar to `DataSetFilters.extract_values` except it's specialized for `ImageData` inputs and returns `ImageData`

PR #7231, opened 2025-02-23
Move `extract_values` validation and splitting code as instance methods

user27182:
### Overview

- Move input validation code into a separate `_validate_extract_values` method 
- Move the value-splitting code into a new `_split_values` method

This is needed to support #7230 (the methods are re-used by that filter) .


tkoyama010 (approved):
LGTM

PR #7233, opened 2025-02-24
Move deprecation warning for `transform` filter  `inplace` into `DataObjectFilters`

user27182:
### Overview

`transform(inplace=True)` was deprecated in #7053.

At the time this required a separate package since there was both `DataSetFilters.transform` and `CompositeFilters.transform`. But now with #7211 there is only `DataObjectFilters.transform` and so a separate package is not needed for this.

This moves the warning message and removes the temporary package.

cc @darikg


user27182:
pre-commit.ci autofi

user27182:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM

PR #7234, opened 2025-02-24
Updates to expose merging points on spectral element boundaries

mattfalcone1997:
A minor update to the Nek5000 reader to allow the `CleanGridOn` and `CleanGridOff` methods to be exposed from the vtkNek5000Reader class. By default, with `CleanGridOff` there is no connectivity across spectral element boundaries with points on the boundary of each element duplicated. This can cause issues with filters such as `extract_surface`, if one desires the external surface mesh to be displayed rather than the spectral element boundaries. This can now be changed at high level with
```python3
import pyvista
filename = examples.download_nek5000()
reader.enable_merge_points()
data = reader.read()
```

To disable this, `reader.disable_merge_points()` can be called.


mattfalcone1997:
I decided, while here, to also expose the methods associated with visualising the spectral element IDs. Possible that the method name `enable_spectral_element_ids` is a bit verbose. A potential option is `spectral_element_ids_on` which is a bit shorter, although using `enable_` fits better with the syntax of the reader classes.

tkoyama010 (approved):
LGTM

PR #7240, opened 2025-02-26
Move `DataObjectFilters` tests to new test file

user27182:
### Overview

Move tests for all filters moved in #7210 and #7211

This should help disentangle the DataSet-only filter tests from the more general ones that also work with MultiBlock

Some test names are also changed for clarity (e.g. `clip_scalar` tests, which only applies to DataSet)
~A `composite` fixture was also removed since it's the same as `multiblock_all` (otherwise it would have needed to be re-defined or imported)~ This is split to #7241


tkoyama010 (approved):
LGTM

PR #7237, opened 2025-02-25
Add `node_type` option to `recursive_iterator`

user27182:
### Overview

Iterate though parent nodes (i.e. the actual MultiBlock blocks themselves) instead of child nodes (i.e. DataSet or None blocks).

PR #7251, opened 2025-02-26
Add `move_nested_field_data_to_root` method for moving field data from nested `MultiBlock` blocks

user27182:
### Overview

This a first step toward a workaround for https://github.com/pyvista/pyvista/issues/6408
Nested `MultiBlock` field data cannot be saved, so this new method can be used to move all the field data to the root block before saving.

This method can also be used to extend `MultiBlock.flatten` so that nested field data can be moved as part of the flattening. This will serve to address/remove this warning:

https://github.com/pyvista/pyvista/blob/d44b3ec4e4a5eee483cc2808edeb48a098a91c8e/pyvista/core/composite.py#L468-L471

Note: this PR depends on #7237


user27182:
Ready but #7253 should be merged first.

tkoyama010 (changes requested):
https://github.com/pyvista/pyvista/pull/7251#issuecomment-2689081685

pyvista/core/composite.py
```diff
@@ -505,6 +516,256 @@ def _recursive_iterator(
...
+                        duplicate_key: str, block_name_: str, index_: tuple[int, ...]
+                    ) -> NoReturn:
+                        index_fmt = _format_nested_index(index_)
+                        msg = (
+                            f"The root user dict cannot be updated with data from nested MultiBlock at index {index_fmt} with name '{block_name_}'.\n"
+                            f"The key '{duplicate_key}' already exists in the root user dict and would be overwritten."
+                        )
+                        raise ValueError(msg)
+
+                    if user_dict_mode in 'preserve':
```
user27182, line 719: This doesn't seem right, should probably be:
```suggestion
                    if user_dict_mode == 'preserve':
```
Will need to double check the tests for this

tests/core/test_composite.py
```diff
@@ -1114,6 +1122,152 @@ def test_recursive_iterator_order(nested_fixture, order, expected_ids, expected_
...
+    if nested1_block_name:
+        root.set_block_name(0, nested1_block_name)
+    if root_field_data:
+        root.field_data.update(root_field_data)
+    if root_user_dict:
+        root.user_dict.update(root_user_dict)
+    return root
+
+
+def test_move_nested_field_data_to_root_duplicate_key_errors():
```
user27182, line 1185:
```suggestion
def test_move_nested_field_data_to_root_safe_update():
```

pyvista/core/composite.py
```diff
@@ -505,6 +516,256 @@ def _recursive_iterator(
...
+
+        separator : str, default: '::'
+            String separator to use when ``prepend_names`` is enabled. The separator
+            is inserted between parent and child block names.
+
+        safe_update : bool, default: True
+            Update the root data safely without overwriting existing data. If ``True``,
+            an error is raised if any nested keys match the root block's keys. If
+            ``False``, nested data is moved without checking if a key already exists,
+            and data may be overwritten.
```
user27182, line 595: We could add a third `None` option:
``True`` : raise error
``False`` : warn if overwriting
``None`` : no error and no warning

pyvista/core/composite.py
```diff
@@ -505,6 +516,257 @@ def _recursive_iterator(
...
+        All field data and user-dict keys are directly updated from the nested blocks
+        by default. Optionally, different modes are provided to control the key names
+        and/or how the data is moved. The move is done safely such that no key values
+        will be overwritten and no data is lost. If any nested keys are duplicates
+        of the root keys, an error is raised.
+
+        .. note::
+            This operation only applies to nested :class:`MultiBlock` blocks. Field data
+            associated with :class:`~pyvista.DataSet` blocks is `not` affected.
+
```
user27182, line 547:
```suggestion

        .. versionadded:: 0.45

```

pyvista/core/composite.py
```diff
@@ -505,6 +516,259 @@ def _recursive_iterator(
...
+        If any nested :class:`MultiBlock` blocks define a :attr:`~pyvista.DataObject.user_dict`,
+        the root user-dict is also updated to include the nested block's user-dict
+        data.
+
+        All field data and user-dict keys are directly updated from the nested blocks
+        by default. Optionally, different modes are provided to control the key names
+        and/or how the data is moved. The move is done safely such that no key values
+        will be overwritten and no data is lost. If any nested keys are duplicates
+        of the root keys, an error is raised.
```
tkoyama010, line 542: 1. Are we planning to remove this once the issue mentioned in https://github.com/pyvista/pyvista/issues/6408#issuecomment-2248597273 is resolved in the new version of VTK?

2. Providing a description that explains this feature exists due to a bug in VTK would make it easier for users to understand.

tkoyama010 (changes requested):

pyvista/core/composite.py
user27182: Although it does help to address the MultiBlock save bug, this method can also be used with `MultiBlock.flatten` to flatten the field data of any nested MultiBlock

I'm not sure if this method is of much use otherwise though. Maybe it's useful if there are lots of blocks and you just want to consolidate the field data (without flattening the blocks)

We can add a note in the docs though with a link to the issue if that's helpful. 

Can even cross-ref this method directly and add it to see also section of `DataObject.save`
Though this can also be added as part of #7258

tests/core/test_composite.py
```diff
@@ -1114,6 +1122,152 @@ def test_recursive_iterator_order(nested_fixture, order, expected_ids, expected_
...
+    root_dict = {'root': 0}
+    nested_dict1 = {'nested1': 1}
+    nested_dict2 = {'nested2': 2}
+    separator = ';'
+
+    expected_user_dict = root_dict.copy()
+    if user_dict_mode == 'flat':
+        expected_user_dict[block_name1] = nested_dict1
+        expected_user_dict[block_name2] = nested_dict2
+    if user_dict_mode == 'prepend':
```
tkoyama010, line 1242:
```suggestion
    elif user_dict_mode == 'prepend':
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

PR #7253, opened 2025-02-27
Fix clearing field data with user dict

user27182:
### Overview

Similar to #7122.
The `user_dict` is proving to be stubborn and doesn't clear properly when clearing field data.

``` python
import pyvista as pv
mesh = pv.Sphere()
mesh.user_dict = dict(a=0)
mesh.field_data.clear()  # *should* clear the user_dict
mesh.user_dict  # not empty
# {"a": 0}
```

This PR fixes this.
The tests are updated to provided better coverage of different removal methods.
The key `'_PYVISTA_USER_DICT` is replaced with a named variable


tkoyama010 (approved):
LGTM

PR #7266, opened 2025-03-02
Fix `crinkle=True` with `clip` filter for `MultiBlock`

### LINKED ISSUE #7218: `clip` with crinkle option does not work with `MultiBlock`

Crinkle doesn't work with `MultiBlock` with the clip filters. This is because crinkle relies on using `extract_cells` which is not available for `MultiBlock`.

This bug was highlighted when refactoring code as part of #7210.

Related issue with crinkle which is also caused by the use of `extract_cells` https://github.com/pyvista/pyvista/issues/6098

``` python
import pyvista as pv
multi = pv.MultiBlock([pv.Sphere()])
multi.clip()  # works
multi.clip(crinkle=True)  # error
```

### System Information

```shell
`
```

### Screenshots

_No response_

user27182:
### Overview

Fix #7218

The following undocumented arrays are also removed from the output as these are implementation-specific and are only used by `crinkle`:
- `'vtkOriginalPointIds'`
- `'vtkOriginalCellIds'`

The `'cell_ids'` array is no longer set as the active scalars.


pyvista/core/filters/data_object.py
```diff
@@ -1036,9 +1037,64 @@ def _clip_with_function(
         crinkle: bool = False,
     ):
         """Clip using an implicit function (internal helper)."""
+        CELL_IDS_KEY = 'cell_ids'
```
user27182, line 1040: It isn't documented that an array with this name is added to the output. This array may just be an implementation detail... is it better to perhaps remove it from the output?

Similarly, I think extract_cells might also be adding a `vtkOriginalCellIds` array, which we may want to remove

banesullivan:
While you're in here, could you fix two other issues with crinkle clipping that I ran into today?

- If the clip produces an empty output then this filter fails as the `"cell_ids"` array is missing
- If the input dataset had no active arrays, then this filter sets `"cell_ids"` as the active array since the `mesh[name] = arr` API does this by default if there isn't an active array (intended behavior) but in my opinion this filter should not be setting an active array

banesullivan:
Only quickly glanced at this code and maybe you already have? If so, could we just add tests for these scenarios at least? 😄

user27182:
> While you're in here, could you fix two other issues with crinkle clipping that I ran into today?
> 
> * If the clip produces an empty output then this filter fails as the `"cell_ids"` array is missing
> * If the input dataset had no active arrays, then this filter sets `"cell_ids"` as the active array since the `mesh[name] = arr` API does this by default if there isn't an active array (intended behavior) but in my opinion this filter should not be setting an active array

Ok, I think I ran into something similar already with empty inputs (this PR casts the array dtype to `int` to fix this, since extract_cells complains with empty float arrays). Will add tests for this though

And I can fix the active array setting issue.

To be clear, are you OK with removing `cell_ids` from the output and keeping this as an implementation detail?

user27182:
> To be clear, are you OK with removing `cell_ids` from the output and keeping this as an implementation detail?

On second thought, I think we may have to keep these. I thought these were not documented, but the docs for `clip` and `clip_box` both state `'cell_ids'` array is added. I'll update this PR to fix this and keep this array.

user27182:
@banesullivan  the active scalars are no longer set to `'cell_ids'` and I added tests for empty meshes in [8588e41](https://github.com/pyvista/pyvista/pull/7266/commits/8588e4180ad5935089bdb53b9b7c28e0afc245fd)

banesullivan:
Awesome, thank you for doing that! And I agree that `"cell_ids"` should probably stay at this point -- I know I have some very old code that assumed `"cell_ids"` was added which makes me think others have done that too

banesullivan (approved):
Looks excellent! Thanks for fixing this!

PR #7271, opened 2025-03-04
Fix invalid f-string and make code more readable in anatomical_groups.py

banesullivan:
This example is failing any time I run the pre-commit checks locally and I found this code unreadable:

```
debug statements (python)................................................Failed
- hook id: debug-statements
- exit code: 1

examples/99-advanced/anatomical_groups.py - Could not parse ast

          File "/home/bane/.cache/pre-commit/repogagxs5o5/py_env-python3.11/lib/python3.11/site-packages/pre_commit_hooks/debug_statement_hook.py", line 57, in check_file
            ast_obj = ast.parse(f.read(), filename=filename)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "/home/bane/.pyenv/versions/3.11.6/lib/python3.11/ast.py", line 50, in parse
            return compile(source, filename, mode, flags,
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "examples/99-advanced/anatomical_groups.py", line 63
            for name, (R, G, B) in names_to_colors.items()
            ^^^
        SyntaxError: f-string: unterminated string
```

This is an attempt to make the code a bit more readable and be a valid f-string for Python 3.11-

### DIFF
--- a/examples/99-advanced/anatomical_groups.py
+++ b/examples/99-advanced/anatomical_groups.py
@@ -58,8 +58,7 @@ ids_to_colors = dataset.user_dict['ids_to_colors']
 
 print('{')
 for name, (R, G, B) in names_to_colors.items():
-    key = f"'{name}'"
-    print(f'{key:<32}: ({R:>3}, {G:>3}, {B:>3}),')
+    print(f'{repr(name):<32}: ({R:>3}, {G:>3}, {B:>3}),')
 print('}')
 
 # %%
--- a/examples/99-advanced/anatomical_groups.py
+++ b/examples/99-advanced/anatomical_groups.py
@@ -58,7 +58,7 @@ ids_to_colors = dataset.user_dict['ids_to_colors']
 
 print('{')
 for name, (R, G, B) in names_to_colors.items():
-    print(f'{repr(name):<32}: ({R:>3}, {G:>3}, {B:>3}),')
+    print(f'{name!r:<32}: ({R:>3}, {G:>3}, {B:>3}),')
 print('}')
 
 # %%
--- a/examples/99-advanced/anatomical_groups.py
+++ b/examples/99-advanced/anatomical_groups.py
@@ -59,7 +59,7 @@ ids_to_colors = dataset.user_dict['ids_to_colors']
 print('{')
 for name, (R, G, B) in names_to_colors.items():
     key = f"'{name}'"
-    print(f'{key:>32}: ({R:>3}, {G:>3}, {B:>3}),')
+    print(f'{key:<32}: ({R:>3}, {G:>3}, {B:>3}),')
 print('}')
 
 # %%
--- a/examples/99-advanced/anatomical_groups.py
+++ b/examples/99-advanced/anatomical_groups.py
@@ -56,14 +56,11 @@ ids_to_colors = dataset.user_dict['ids_to_colors']
 #     This mapping is specific to the PyVista datasets and is not part of the
 #     TotalSegmentator data.
 
-print(
-    '{\n'
-    + '\n'.join(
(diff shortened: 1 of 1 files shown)


examples/99-advanced/anatomical_groups.py
```diff
@@ -56,14 +56,11 @@
...
-    + '\n'.join(
-        f'    {"'" + name + "':":<32} ({R:>3}, {G:>3}, {B:>3}),'
-        for name, (R, G, B) in names_to_colors.items()
-    )
-    + '\n}',
-)
+print('{')
+for name, (R, G, B) in names_to_colors.items():
+    key = f"'{name}'"
+    print(f'{key:>32}: ({R:>3}, {G:>3}, {B:>3}),')
```
user27182, line 62: Thanks, this is much better. This may be a style thing but I find it hard to read the label names with the strings right-aligned
```suggestion
    print(f'{key:<32}: ({R:>3}, {G:>3}, {B:>3}),')
```

examples/99-advanced/anatomical_groups.py
```diff
@@ -56,14 +56,11 @@
...
-    '{\n'
-    + '\n'.join(
-        f'    {"'" + name + "':":<32} ({R:>3}, {G:>3}, {B:>3}),'
-        for name, (R, G, B) in names_to_colors.items()
-    )
-    + '\n}',
-)
+print('{')
+for name, (R, G, B) in names_to_colors.items():
+    key = f"'{name}'"
```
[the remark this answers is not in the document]
banesullivan: Oh neat, but with that then maybe I can combine to the one line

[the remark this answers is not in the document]
banesullivan: Done in 5abfb19

examples/99-advanced/anatomical_groups.py
```diff
@@ -56,14 +56,10 @@
...
-    '{\n'
-    + '\n'.join(
-        f'    {"'" + name + "':":<32} ({R:>3}, {G:>3}, {B:>3}),'
-        for name, (R, G, B) in names_to_colors.items()
-    )
-    + '\n}',
-)
+print('{')
+for name, (R, G, B) in names_to_colors.items():
+    print(f'{repr(name):<32}: ({R:>3}, {G:>3}, {B:>3}),')
```
adeak, line 61: Sorry, I brainfarted my comment from earlier. I meant
```suggestion
    print(f'{name!r:<32}: ({R:>3}, {G:>3}, {B:>3}),')
```
This is what this format specifier was made for. But if you think it's more obscure than explicit repr than I don't insist.

user27182 (approved):

examples/99-advanced/anatomical_groups.py
banesullivan: That indeed feels a little obscure to me

user27182: The important thing here for me is that the printed dict looks good. I wanted users to be able to easily see and copy the color mapping without needing to download the dataset. So from that perspective, I'm indifferent to how this code is formatted. Feel free to revert to use `repr`.

banesullivan: Ag didn't realize it was committed, this works for me as just having this in a traditional for loop makes it readable enough

user27182: The suggestion by @adeak is also consistent with [RUF010](https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/), which we have enabled. Pre-commit was failing previously when using `repr`:

https://results.pre-commit.ci/run/github/92974124/1741124058.AyC2Vz0oRcW7U2xvhkSxQA

So I think this is the right choice.

PR #7272, opened 2025-03-04
Refactor `clip` filter tests and add checks for `MultiBlock` inputs

user27182:
### Overview

Split from #7266

- use a fixture which includes `MultiBlock` inputs
- add instance checks to test `MultiBlock` inputs
- parametrize `return_clipped` so only a single for loop is needed (currently the test uses two separate for-loops, one for each value of `return_clipped`)
- split test for various `normal` inputs into its own test
- define a `'cell_ids'` array explicitly for the crinkle test

### DIFF
--- a/tests/core/test_dataobject_filters.py
+++ b/tests/core/test_dataobject_filters.py
@@ -38,7 +38,7 @@ def test_clip_filter(multiblock_all_with_nested_and_none, return_clipped):
     assert None in multi.recursive_iterator()
 
     for dataset in multi:
-        clp = dataset.clip(normal='x', invert=True, return_clipped=return_clipped)
+        clips = dataset.clip(normal='x', invert=True, return_clipped=return_clipped)
         assert clp is not None
 
         if return_clipped:
--- a/tests/core/test_dataobject_filters.py
+++ b/tests/core/test_dataobject_filters.py
@@ -35,6 +35,7 @@ def test_clip_filter(multiblock_all_with_nested_and_none, return_clipped):
         if block is None:
             del multi[i]
     assert None not in multi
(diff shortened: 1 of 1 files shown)


tests/core/test_dataobject_filters.py
```diff
@@ -26,34 +26,57 @@
...
+    # Remove None blocks in the root block but keep the none block in the nested MultiBlock
+    multi = multiblock_all_with_nested_and_none
+    for i, block in enumerate(multi):
+        if block is None:
+            del multi[i]
+    assert None not in multi
+    assert None in multi.recursive_iterator()
+
+    for dataset in multi:
+        clp = dataset.clip(normal='x', invert=True, return_clipped=return_clipped)
```
tkoyama010, line 41:
```suggestion
        clips = dataset.clip(normal='x', invert=True, return_clipped=return_clipped)
```

tests/core/test_dataobject_filters.py
```diff
@@ -26,34 +26,57 @@
...
-            assert isinstance(clp, pv.UnstructuredGrid)
+            assert isinstance(clp, pv.DataObject)
+            # Make dataset iterable
+            clp = [clp]
 
-    # clip with get_clipped=True
-    for i, dataset in enumerate(datasets):
-        clp1, clp2 = dataset.clip(normal=normals[i], invert=True, return_clipped=True)
-        for clp in (clp1, clp2):
+        for clipped in clp:
```
tkoyama010, line 52:
```suggestion
        for clip in clips:
```

tkoyama010 (changes requested):
nits

tkoyama010 (approved):
LGTM

Issue #7254, opened 2025-02-27
Unexpected rotation direction in Affine Widget


dad616610:
### Describe the bug, what's wrong, and what you expected.

When the Affine Widget's arrow points in the negative direction of the global axis, the rotation is reversed: if the mouse moves counterclockwise, the mesh rotates clockwise.

![Image](https://github.com/user-attachments/assets/3981f675-97a9-4f1b-a344-97270fa5cd52)

### Steps to reproduce the bug.

The mesh rotates clockwise when the mouse moves counterclockwise around the Z (blue) and X (red) axes.

```python
import pyvista as pv
import numpy as np

pl = pv.Plotter()
actor = pl.add_mesh(pv.Cube())
widget = pl.add_affine_transform_widget(
    actor,
    axes=np.array(
        (
            (-1, 0, 0),
            (0, 1, 0),
            (0, 0, -1),
        ),
    ),
)
pl.show()
```

### System Information

```shell
Python 3.12.0 (tags/v3.12.0:0fb18b0, Oct  2 2023, 13:03:39) [MSC v.1935 64
  bit (AMD64)]

           pyvista : 0.44.2
               vtk : 9.3.1
             numpy : 1.26.4
        matplotlib : 3.9.2
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 11.1.0
           imageio : 2.37.0
         pyvistaqt : 0.11.1
             scipy : 1.13.1
              tqdm : 4.66.5
--------------------------------------------------------------------------------
```

user27182:
This issue seems to stem from how the rotation angle is calculated. Replacing the `angle` here with `-angle` fixes this for this case (but also breaks it for the `np.eye(3)` case).
https://github.com/pyvista/pyvista/blob/349737c18a82fe96f948032cc59cba1de6c2d011/pyvista/plotting/affine_widget.py#L369

dad616610:
I believe I fixed it by modifying this condition https://github.com/pyvista/pyvista/blob/349737c18a82fe96f948032cc59cba1de6c2d011/pyvista/plotting/affine_widget.py#L363
to 
```diff
- if (cross[index] < 0):
+ if (cross[index] < 0) ^ (np.dot(-GLOBAL_AXES[index], normal) > 0):
```
So, we flip the angle just as we did when pointing towards the positive global axis, but we "flip the flip" when pointing towards the negative global axis.

This approach works for both the axes in the example and for `np.eye(3)`.

[...]

PR #7273, opened 2025-03-04
Fix affine widget last event position

dad616610:
### Overview

Previously, translation and rotation positions in the affine widget were taken from past events. This PR makes use of positions from the current event instead.

Partially retaled to #7254 

### Details

In real usage, relying on the previous event’s position is almost unnoticeable since mouse move events occur frequently, resulting in minimal differences. However, in tests—where mouse movement is strictly controlled—this issue becomes significant.

Using outdated positions leads to inaccurate and unreliable tests. As shown in the test output below:

- The mouse position in `move_callback` differs from the position obtained in `_get_world_coord_trans`.
- The "Y translation" test incorrectly gets the "press" position from the previous event, which differs from the current mouse position.

The data was obtained using 1st commit in this PR

<details><summary>Last Event test output, abridged </summary>
<p>

```
X translation
move cbck (click_x = 199, click_y = 199)
Trans coord (x = 0, y = 0), when = 'press'
move cbck (click_x = 400, click_y = 200)
Trans coord (x = 199, y = 199), when = 'move cbck'
move cbck (click_x = 400, click_y = 200)
Trans coord (x = 400, y = 200), when = 'move cbck'

Y translation
move cbck (click_x = 201, click_y = 199)
Trans coord (x = 400, y = 200), when = 'press'
move cbck (click_x = 400, click_y = 200)
Trans coord (x = 201, y = 199), when = 'move cbck'
```

</p>
</details> 

This data for the fixed version was obtained using 4th commit. Notice, that every position from `_get_world_coord_trans` matches the current position of the mouse

<details><summary>Event test output, abridged </summary>
<p>

```
X translation
move cbck (click_x = 196, click_y = 196)
Trans coord (x = 196, y = 196), when = 'press'
move cbck (click_x = 400, click_y = 200)
Trans coord (x = 400, y = 200), when = 'move cbck'
move cbck (click_x = 400, click_y = 200)
Trans coord (x = 400, y = 200), when = 'move cbck'
actor.user_matrix[0, 3] = np.float64(-171.0686161560028)

Y translation
move cbck (click_x = 202, click_y = 197)
Trans coord (x = 202, y = 197), when = 'press'
move cbck (click_x = 400, click_y = 200)
Trans coord (x = 400, y = 200), when = 'move cbck'
actor.user_matrix[1, 3] = np.float64(162.74239704488312)
```

</p>
</details> 

___

I also changed almost all the press and move positions to ensure test correctness and reduced redundancy with `test_translation` and `test_rotation` inner functions


user27182:
Awesome! I wonder if this could have also been the root cause of why this test needed to be marked as flaky. Do you think we should also remove the `@flaky_test` decorator from `test_affine_widget`?

user27182:
> > Do you think we should also remove the `@flaky_test` decorator from `test_affine_widget`?
> 
> I'm not sure why it was added in the first place
> 
> The test was deterministic regarding mouse positions even before the fix: if the test passed once, it shouldn't fail in subsequent runs (since mouse positions remain consistent across tests). The fix shouldn’t have affected this aspect of the test
> 
> Another possible factor could be `GetEventPosition`, `np.linalg.inv` and similar functions. It's a wild guess, but could they behave differently on different platforms or software versions? During my local tests on Windows, I didn't encounter the flaky behaviour. There might have been a press position close to another actor, but I’m not sure if that actor was sometimes chosen over the intended one
> 
> I believe the fix didn't attribute much to the flaky behaviour (if there was any) for the reasons mentioned above. We could remove the decorator and observe over time whether the test is actually flaky

Ok let's keep the decorator for now. I believe this test was flaky for Linux only with older VTK versions, perhaps not likely to be related to the changes in this PR after all.

> Side note: The current test suite only checks individual translations and rotations—applying one transformation at a time and then resetting the widget. Should we also test a sequence of transformations (multiple translations, rotations, and/or other transformations), or is the current approach sufficient? Would testing a series be overkill?

I think https://github.com/pyvista/pyvista/issues/7254 shows that this widget could probably benefit from some additional tests, so probably not a bad idea. With the test actions refactored from this PR it should hopefully be easier to write and debug. If the tests can be parametrized without much duplication of code then I think this is a good idea.

user27182 (approved):
I have not closely reviewed the changes to the tests (e.g. check for modified values) but the refactoring is nice and these changes LGTM overall. I think any additional tests can be added in a separate PR

tkoyama010 (approved):
LGTM

PR #7275, opened 2025-03-06
Add `DataObject.is_empty` property

user27182:


pyvista/core/dataobject.py
```diff
@@ -909,3 +909,7 @@ def _unserialize_pyvista_pickle_format(self: Self, state: dict[str, Any]) -> Non
         # copy data
         self.copy_structure(mesh)  # type: ignore[arg-type]
         self.copy_attributes(mesh)  # type: ignore[arg-type]
+
+    @property
+    def is_empty(self) -> bool:
+        """Return ``True`` if the object is empty."""
```
MatthewFlamm, line 915: Looks like mypy errors.  I'm assuming we cannot use ABC here for the class to use abstract method to force overwriting on the subclasses.  Should we `raise NotImplementedError` instead?  This _might_ make mypy happy.

pyvista/core/objects.py
```diff
@@ -380,3 +380,25 @@ def get_data_range(  # type: ignore[override]
             return (np.nan, np.nan)
         # Use the array range
         return np.nanmin(arr), np.nanmax(arr)
+
+    @property
+    def is_empty(self) -> bool:  # numpydoc ignore=RT01
+        """Return ``True`` if the table has no rows or columns.
```
MatthewFlamm, line 386: is it possible to have rows but no columns or vice versa?  The text here has `or` but the conditional in the property has `and`.

tkoyama010 (approved):
LGTM

MatthewFlamm (approved):

PR #7276, opened 2025-03-06
Add typing cases for `wrap` and `recursive_iterator`

user27182:
### Overview

Mypy only sees the annotations, so an overloaded signature can resolve to the wrong overload with nothing failing. These cases pin both halves at once: `assert_types(expr, T)` is `typing_extensions.assert_type` to Mypy and a real check when the line runs, so a case passes only if the two agree. They cover the `wrap` overloads from #7284 and the `recursive_iterator` overloads from #7301. The framework started in the validation work of #6751 and now lives in [type-assert](https://github.com/user27182/type-assert), pinned exactly because its API is still unstable.

It found one straight away: `recursive_iterator('blocks', prepend_names=True)` type-checked and then raised `ValueError`, so the `'ids'` and `'blocks'` overloads no longer accept `prepend_names` or `separator`.

- Drops `pytest-mypy-plugins`, which nothing uses. It was where the test environment got Mypy, which the `typing` group now supplies directly.
- Ignores an unraisable `ResourceWarning` for the launcher's own `pytest.EXE` handle, which the reference leak check finalizes mid-run on Windows.

Changes drafted by Claude Opus 5 but fully understood by me.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tests/typing/test_typing_cases/transform_filters.py
```diff
@@ -0,0 +1,84 @@
...
+
+# fmt: off
+
+# test transform with all mesh types
+reveal_type(RectilinearGrid().transform(np.eye(4), inplace=False))           # EXPECTED_TYPE: "StructuredGrid"
+reveal_type(ImageData().transform(np.eye(4), inplace=False))                 # EXPECTED_TYPE: "ImageData"
+reveal_type(PointSet().transform(np.eye(4), inplace=False))                  # EXPECTED_TYPE: "PointSet"
+reveal_type(StructuredGrid().transform(np.eye(4), inplace=False))            # EXPECTED_TYPE: "StructuredGrid"
+reveal_type(PolyData().transform(np.eye(4), inplace=False))                  # EXPECTED_TYPE: "PolyData"
+# reveal_type(ExplicitStructuredGrid().transform(np.eye(4), inplace=False))    # EXPECTED_TYPE: "ExplicitStructuredGrid"
```
user27182, line 22: This is commented out because an empty init with `ExplicitStructuredGrid()` generates seg faults sporadically on Windows. But I think we should leave the test in, even if commented out, since it should be tested.

pyvista/core/partitioned.py
```diff
@@ -248,3 +251,9 @@ def append(self, dataset) -> None:
         index = self.n_partitions
         self.n_partitions += 1
         self[index] = dataset
+
+    def get_data_range(  # numpydoc ignore=RT01
+        self: Self, name: str | None, preference: FieldAssociation | str
+    ) -> tuple[float, float]:  # pragma: no cover
+        """Get the non-NaN min and max of a named array."""
+        return DataObject.get_data_range(self, name=name, preference=preference)
```
user27182, line 259: This is needed because the overload tests for `wrap` use an empty init of `pv.PartitionedDataSet()` and mypy complains about this because `get_data_range` is an abstract property. It should be properly defined (probably similar to MultiBlock) but is not. Here, a workaround is to call super just to make mypy happy, though this will still raise `NotImplementedError` at runtime.

tkoyama010:
I'm sorry. I may have merged in the wrong order. Was this merge intended?

user27182:
Could have merged this PR before #7301 but it's fine to also have that one merged as part of this PR. Just depends on preference for having more or less changes for the squash and merge with main. 

If you're fine with the scope of this PR (which now includes the recursive iterator overloads) then it's also fine with me. Otherwise we can revert the latest commit and open a new PR for the iterator overloads if there's a preference to merge those separately.

user27182:
We can also split the overloads separately _without_ adding the tests, like #7284

tkoyama010 (dismissed):
LGTM

pyvista/core/partitioned.py
```diff
@@ -244,6 +247,26 @@ def n_partitions(self, n) -> None:
...
+        >>> mesh = pv.MultiBlock()
+        >>> mesh.is_empty
+        True
+
+        >>> mesh.append(pv.Sphere())
+        >>> mesh.is_empty
+        False
+
+        """
+        return self.n_partitions == 0
```
user27182, line 275: mypy complains about this being abstract, so it's defined here. But it's missing test coverage.... this is added in #7753

user27182:
New updates from [bc28246](https://github.com/pyvista/pyvista/pull/7276/commits/bc2824608e83e6a38a14420af64f39f28b62c229) seem to be causing runtime tests to be flaky, will look into this

akaszynski (approved):

PR #7278, opened 2025-03-06
Add `Transform.as_rotation` method to generate rotation representations

user27182:
### Overview

Convenience method to convert a transform's rotation component into a SciPy ``Rotation`` object or any of its representations.


tkoyama010 (approved):
LGTM

PR #7279, opened 2025-03-07
Allow applying a `Transform` to actors

user27182:

PR #7283, opened 2025-03-07
Skip `@overload` test coverage

user27182:
### Overview

Fix false positives from partial branches from typing overloads. See https://github.com/pyvista/pyvista/pull/7104#issuecomment-2707251680

This fix was tested in #7279 and resolves the overload coverage issues experienced there.


tkoyama010 (approved):
LGTM

PR #7284, opened 2025-03-08
Add return type annotations for transformation filters

user27182:
### Overview

Split from #7276. The typing tests will be added separately.


tkoyama010 (approved):
LGTM

PR #7285, opened 2025-03-08
Add `Prop3D.transform` method

user27182:
### Overview

Related to #7279. 
This modifies the object's `user_matrix` to apply the transformation.

PR #7287, opened 2025-03-08
Add `Prop3D.copy` method

user27182:
### Overview

Add new abstract property and implement it for the `pv.Volume` class.

A bug with `Actor.copy()` is also fixed in cases where there is no mapper to copy.

This is related to #7285 since `Prop3D.transform` has an `inplace` keyword which requires making a copy.


tkoyama010 (approved):
LGTM

banesullivan (approved):

PR #7288, opened 2025-03-08
Flatten field data as part of `MultiBlock.flatten`

user27182:
### Overview

Address issue with `flatten` identified in https://github.com/pyvista/pyvista/pull/7251#issue-2882547962


pyvista/core/composite.py
```diff
@@ -809,14 +813,47 @@ def flatten(
...
+                and ``'prepend'`` modes are all equivalent. They only differ when there
+                is at least two levels of nesting.
+
         separator : str, default: '::'
-            String separator to use when ``name_mode='prepend'`` is used. The separator
-            is inserted between parent and child block names.
+            String separator to use when ``'prepend'`` mode is used. The separator
+            is inserted between parent and child block name or field data array names.
+
+        safe_update : bool, default: True
```
user27182, line 850: This name matches the keyword name from `move_nested_field_data_to_root`, but I'm not sure it's a meaningful description for `flatten`. Should this perhaps be called `check_duplicate_keys` or something similar? Should the keyword in `move_nested_field_data_to_root` also be updated to match?
```suggestion
        check_duplicate_keys : bool, default: True
```

tkoyama010: +1 `check_duplicate_keys`. I am always a big fan of explicit expression!

tkoyama010 (changes requested):
https://github.com/pyvista/pyvista/pull/7288/files#r1986254733

tkoyama010 (approved):
LGTM

PR #7291, opened 2025-03-10
Only run plotting tests with Plotting Linux CI

user27182:
### Overview

This PR modifies the Linux CI tests to match CI testing for Windows and macOS. The `tinypages` tests are also moved to the plotting module since these fail otherwise.

### Details

CI for macOS and windows tests currently run core and plotting tests as:
https://github.com/pyvista/pyvista/blob/bbd8060b5655e313c68a90808528b9f6af048f0e/.github/workflows/testing-and-deployment.yml#L53-L57

Note how `Test Core` ignores plotting, and `Test Plotting` only tests plotting.

For Linux, however, this is not the case:
https://github.com/pyvista/pyvista/blob/bbd8060b5655e313c68a90808528b9f6af048f0e/.github/workflows/testing-and-deployment.yml#L224-L225
https://github.com/pyvista/pyvista/blob/bbd8060b5655e313c68a90808528b9f6af048f0e/.github/workflows/testing-and-deployment.yml#L235-L236

This makes the testing inconsistent across different OS, and also means that many non-plotting tests are executed as part of the plotting tests.

This is important to note because these tests use `xvfb-run` which slow down the tests. E.g.  #7276 adds typing tests which have nothing to do with plotting, but because they run as part of the `Test Plotting` Linux CI testing, these add approx 1hr of test time when run with `xvfb`, but take < 1 minute when run as part of normal `Core` testing.


banesullivan:
> typing tests which have nothing to do with plotting, but because they run as part of the Test Plotting Linux CI testing, these add approx 1hr of test time when run with xvfb, but take < 1 minute when run as part of normal Core testing.

👀 👀 👀 👀  for real?? I'm really wondering why this is

banesullivan (approved):
This is a much needed change! Thanks for taking the initiative here and fixing this mismatch in how the core vs plotting tests run. 

All looks good here ✅

banesullivan:
I'm going to go ahead and merge this, as I want to work on some CI things tonight and try to split up the docstring tests similarly since they are now taking ~an hour to run

user27182:
> > typing tests which have nothing to do with plotting, but because they run as part of the Test Plotting Linux CI testing, these add approx 1hr of test time when run with xvfb, but take < 1 minute when run as part of normal Core testing.
> 
> 👀 👀 👀 👀 for real?? I'm really wondering why this is

Yes this is very odd.

See green Linux CI from #7276, 
https://github.com/pyvista/pyvista/actions/runs/13766179219/job/38493097453?pr=7276
Linux tests took ~1hr 40min, but was expecting closer to 40min. The only change from main here is that a `test_typing` test file is added, which runs `mypy` once and generates 88 test cases from that. Not really sure why this would have such a big effect on the plotting tests

But after modifying the workflow to remove the typing tests from the Test Plotting action, we're down to 43 mins
https://github.com/pyvista/pyvista/actions/runs/13771028390/job/38512929697?pr=7276

PR #7295, opened 2025-03-11
Remove conda CI testing and environment files

banesullivan:
The conda-forge CI job hasn't run in over 2 years and we are not fully maintaining the `environment.yml` for conda.

I propose outright removing this CI job since it is very out of date and to remove the conda environment spec to avoid potentially confusing users


tkoyama010 (approved):
LGTM

PR #7299, opened 2025-03-11
Fix docker workflow permissions

user27182:
Follow-up to #7197

### DIFF
--- a/.github/workflows/docker-package.yml
+++ b/.github/workflows/docker-package.yml
@@ -17,7 +17,7 @@ env:
   IMAGE_NAME: ${{ github.repository }}
 
 permissions:
-  id-token: none
+  id-token: write
 
 jobs:
   build-and-publish:


banesullivan (approved):

tkoyama010 (approved):

user27182:
This fix didn't work, we could try adding more permissions, e.g.
``` diff
permissions:
  id-token: write
+ contents: read
+ packages: write
```

Though I'm not sure if this will work.
Or, we could revert the addition of permissions and ignore the zizmor error instead.

banesullivan:
What's the problem exactly?

user27182:
> What's the problem exactly?

CI triggered from pushes to main is failing:
https://github.com/pyvista/pyvista/actions/runs/14003619005/job/39214344590#step:7:684
``` 
Error: buildx failed with: ERROR: denied: installation not allowed to Write organization package
```

But this PR gave write permissions. So I'm not really sure what the issue is, so maybe we either revert adding permissions altogether or try giving more permissions.

banesullivan:
We want `packages: write` not `id-token: write` per https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#overview

PR #7302, opened 2025-03-12
Fix `to_meshio` with empty mesh

user27182:
### Overview

This is broken:

``` python
import pyvista as pv
pv.to_meshio(pv.UnstructuredGrid())  # Error
```

This PR fixes this.

### DIFF
--- a/pyvista/core/utilities/fileio.py
+++ b/pyvista/core/utilities/fileio.py
@@ -939,6 +939,13 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
         from meshio._vtk_common import meshio_to_vtk_type
         from meshio.vtk._vtk_42 import vtk_type_to_numnodes
 
+    if not mesh.cells:
+        # Empty mesh
+        grid = pyvista.UnstructuredGrid()
+        if mesh.points.size > 0:
+            grid.points = mesh.points
+        return grid
+
     # Extract cells from meshio.Mesh object
     cells = []
     cell_type = []
@@ -981,13 +988,8 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
         zero_points = np.zeros((len(points), 1), dtype=points.dtype)
         points = np.hstack((points, zero_points))
 
-    cat_cells = (
-        np.array((), dtype=np.int64)
-        if len(cells) == 0
-        else np.concatenate(cells).astype(np.int64, copy=False)
-    )
     grid = pyvista.UnstructuredGrid(
-        cat_cells,
+        np.concatenate(cells).astype(np.int64, copy=False),
         np.array(cell_type),
         np.array(points, np.float64),
     )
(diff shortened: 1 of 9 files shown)


pyvista/core/utilities/fileio.py
```diff
@@ -981,8 +981,13 @@ def from_meshio(mesh: meshio.Mesh) -> UnstructuredGrid:
         zero_points = np.zeros((len(points), 1), dtype=points.dtype)
         points = np.hstack((points, zero_points))
 
+    cat_cells = (
+        np.array((), dtype=np.int64)
+        if len(cells) == 0
+        else np.concatenate(cells).astype(np.int64, copy=False)
+    )
```
keurfonluu, line 988: Why not simply return an empty unstructured grid at the beginning of ``from_meshio``? e.g.

```python
if not mesh.cells:
    return pv.UnstructuredGrid()
```

user27182: Can do, wasn't sure if any other properties could exist that were handled here. Is empty defined as no cells? What if there are points but no cells? And bool logic fails with numpy arrays, maybe this instead?:

``` python
if mesh.points.size == 0:
    return pv.UnstructuredGrid()
```

keurfonluu: ``meshio.Mesh.cells`` is always a list (of ``meshio.CellBlock``). Maybe something like this to handle case where there are points but no cells:

```python
if not mesh.cells:
    grid = pv.UnstructuredGrid()
    
    if len(mesh.points):
        grid.points = mesh.points

    return grid
```

tests/test_meshio.py
```diff
@@ -99,15 +100,25 @@
 )
 
 
-@pytest.mark.parametrize('mesh_in', [beam, airplane, uniform, uniform2d, mesh2d, polyhedron, cow])
+@pytest.mark.parametrize(
+    'mesh_in', [beam, airplane, uniform, uniform2d, mesh2d, polyhedron, cow, empty]
+)
 def test_meshio(mesh_in, tmpdir):
     if isinstance(mesh_in, meshio.Mesh):
```
keurfonluu, line 110: Maybe better to simply not use ``read_meshio`` and ``save_meshio`` which are wrappers for ``from_meshio`` and ``to_meshio`` anyway and already tested in the other tests.

```python
if isinstance(mesh_in, meshio.Mesh):
    mesh_in = pv.from_meshio(mesh_in)

mesh = pv.from_meshio(pv.to_meshio(mesh_in))
```

user27182: The fileio was removed from the test in https://github.com/pyvista/pyvista/pull/7302/commits/0ca2327454fb221e87ca6b61aab54f428ecbd71d, I think that's what you were suggesting? (To avoid having to handle the meshio read errors with empty meshes)

keurfonluu (approved):

PR #7305, opened 2025-03-12
Fix empty `UnstructuredGrid` cell properties when cast from `PolyData`

user27182:
### Overview

Casting empty `PolyData` to `UnstructuredGrid` (with vtkAppendFilter internally) breaks fundamental cell properties:

``` python
import pyvista as pv
pv.PolyData().cast_to_unstructured_grid().cells  # Error
```


tkoyama010 (approved):
LGTM

PR #7306, opened 2025-03-13
Deprecate `compute_cell_quality` and replace with a more general `cell_quality` filter

user27182:
### Overview
New filter `cell_quality` with the following changes/differences:

API changes
- Add option to specify multiple quality measures instead of just one
- Add option to compute all measures and/or all valid measures
- Return cell quality array with the same name as the measure instead of using `'CellQuality'` as the name

Under-the-hood changes
- Add type hints for quality measures
- Generate the quality measures dict dynamically instead of via a static dict
- Add tests to ensure the type hints match the actual measures
- Generalize so work with `MultiBlock` inputs

Documentation changes
- Update `mesh_quality` example with new plots for comparing measures.
- Add table to filter docs showing valid measures and cell types:
image


pyvista/core/filters/data_set.py
```diff
@@ -5319,14 +5353,32 @@ def compute_cell_quality(  # type: ignore[misc]
...
+            array is created for each measure.
+
+            .. versionadded:: 0.45
+
+                Add option to compute multiple measures or all measures.
+
+            .. deprecated:: 0.45
+
+                The default quality measure will change from ``'scaled_jacobian'`` to
+                ``'all'`` in a future version.
```
MatthewFlamm, line 5381: `all` will result in a very large memory requirement as there will be a lot of data.  Additionally, some of the measures are meaningless for certain cell types in addition.  What about allowing a sequence of measures to be passed in?

user27182: Yes I have also noticed this while generated an example. I will revert this, and `'all'` will not be the default. And your point about memory is absolutely valid since there are maybe close to 30 measures.

And this PR already allows a sequence of measures as input, e.g.;
``` python
measures = ['volume', 'collapse_ratio', 'jacobian', 'scaled_jacobian' ]
qual = mesh.compute_cell_quality(measures)
```

MatthewFlamm: > And this PR already allows a sequence of measures as input, e.g.;

Didn't read close enough, this aspect is a great addition!

examples/01-filter/mesh_quality.py
```diff
@@ -6,29 +6,72 @@
...
+plot_kwargs = dict(cmap='bwr', show_edges=True, cpos='xy', zoom=1.3)
+qual.plot(scalars='scaled_jacobian', **plot_kwargs)
+
+# %%
+# Compute additional measures and plot them all for comparison.
+
+measures = ['area', 'max_angle', 'min_angle', 'shape']
+qual = mesh.compute_cell_quality(measures)
+for measure in measures:
+    qual.plot(scalars=measure, **plot_kwargs)
```
user27182, line 50: I think it would be better to use four subplots here in single plotter. But this is avoided here because of #7307.

MatthewFlamm:
I am +2 on using a sequence of measures as an allowable input.  I don't think we should remove the ability to provide a single string.  However, it looks like this mechanism is being used as a deprecation for the output data array name change, which I'm also in favor of.  It isn't straightforward how to deprecate the array name change while also allowing for a single string to be provided.

One thought is to have the deprecation warning only say that the data array name will change in the future when using a single string.  Until that time, use a sequence of strings.  This might be confusing to users?  Another thought is to keep the current PR strategy, but then "change our minds" and just swap the data array name.

user27182:
> I don't think we should remove the ability to provide a single string.  However, it looks like this mechanism is being used as a deprecation for the output data array name change, which I'm also in favor of.  It isn't straightforward how to deprecate the array name change while also allowing for a single string to be provided.

Yes this is correct. Strings are allowed, but during the deprecation period a warning is currently emitted for string inputs because users may currently depend on using the CellQuality array. 
> 
> One thought is to have the deprecation warning only say that the data array name will change in the future when using a single string.  Until that time, use a sequence of strings.  This might be confusing to users?  Another thought is to keep the current PR strategy, but then "change our minds" and just swap the data array name.

I'm not sure I understand the difference between these two? The former is the current intention for this PR, but maybe the way it's written isn't clear

We could also 
- use None as default, users could specify the string instead of None to remove the warning
- move the warning to `get_array` and only warn if users try to access the array

Using None as the new (temporary) default may be the most straightforward?

MatthewFlamm:
> The former is the current intention for this PR, 

You are right, my comprehension isn't very good today.

> use None as default, users could specify the string instead of None to remove the warning

This one isn't good as users are almost always already specifying a parameter here and will be using the old data array name. 

> move the warning to get_array and only warn if users try to access the array

This is an interesting suggestion. It would 'break' if anyone is creating this array outside this workflow.  But seems unlikely?


My feeling is to keep the current PR strategy.

user27182:
An `'all_valid'` option was added in 3833666 to filter out irrelevant measures.

With this option, I was able to run an analysis to determine which cells and measures `compute_cell_quality` is actually valid for. (This is limited to cells which are documented in the `examples.cells` modules)

``` python
import inspect
from pyvista.examples import cells

# Get all functions from the module that start with a capital letter
cell_funcs = [
    name for name, obj in inspect.getmembers(cells, inspect.isfunction)
    if name[0].isupper()
]

has_measures = {}
for func in cell_funcs:
    mesh = getattr(cells, func)()
    cell_name = pv.CellType(mesh.celltypes[0]).name
    mesh = mesh.compute_cell_quality('all_valid')
    measures = mesh.array_names
    if len(measures) > 0:
        has_measures[cell_name] = measures

formatted_dict = "{\n"
for key, value in has_measures.items():
    formatted_dict += f"    '{key}': {value},\n"
formatted_dict += "}"
print(formatted_dict)
``` 

output:
``` python
{
    'HEXAHEDRON': ['condition', 'diagonal', 'dimension', 'distortion', 'jacobian', 'max_aspect_frobenius', 'max_edge_ratio', 'med_aspect_frobenius', 'oddy', 'relative_size_squared', 'scaled_jacobian', 'shape', 'shape_and_size', 'shear', 'shear_and_size', 'skew', 'stretch', 'taper', 'volume'],
    'PIXEL': ['area'],
    'PYRAMID': ['jacobian', 'scaled_jacobian', 'shape', 'volume'],
    'QUAD': ['area', 'aspect_ratio', 'condition', 'distortion', 'jacobian', 'max_angle', 'max_aspect_frobenius', 'max_edge_ratio', 'med_aspect_frobenius', 'min_angle', 'oddy', 'radius_ratio', 'relative_size_squared', 'scaled_jacobian', 'shape', 'shape_and_size', 'shear', 'shear_and_size', 'skew', 'stretch', 'taper', 'warpage'],
    'TETRA': ['aspect_frobenius', 'aspect_gamma', 'aspect_ratio', 'collapse_ratio', 'distortion', 'jacobian', 'min_angle', 'radius_ratio', 'relative_size_squared', 'shape', 'shape_and_size', 'volume'],
    'TRIANGLE': ['area', 'aspect_frobenius', 'aspect_ratio', 'condition', 'distortion', 'max_angle', 'min_angle', 'radius_ratio', 'relative_size_squared', 'scaled_jacobian', 'shape', 'shape_and_size'],
    'TRIANGLE_STRIP': ['area'],
    'WEDGE': ['condition', 'distortion', 'jacobian', 'max_aspect_frobenius', 'scaled_jacobian', 'shape', 'volume'],
}
```

We could add this to the docstring in this PR, or maybe document it in a follow-up PR?

user27182:
> ```python
> {
>     'HEXAHEDRON': ['condition', 'diagonal', 'dimension', 'distortion', 'jacobian', 'max_aspect_frobenius', 'max_edge_ratio', 'med_aspect_frobenius', 'oddy', 'relative_size_squared', 'scaled_jacobian', 'shape', 'shape_and_size', 'shear', 'shear_and_size', 'skew', 'stretch', 'taper', 'volume'],
>     'PIXEL': ['area'],
>     'PYRAMID': ['jacobian', 'scaled_jacobian', 'shape', 'volume'],
>     'QUAD': ['area', 'aspect_ratio', 'condition', 'distortion', 'jacobian', 'max_angle', 'max_aspect_frobenius', 'max_edge_ratio', 'med_aspect_frobenius', 'min_angle', 'oddy', 'radius_ratio', 'relative_size_squared', 'scaled_jacobian', 'shape', 'shape_and_size', 'shear', 'shear_and_size', 'skew', 'stretch', 'taper', 'warpage'],
>     'TETRA': ['aspect_frobenius', 'aspect_gamma', 'aspect_ratio', 'collapse_ratio', 'distortion', 'jacobian', 'min_angle', 'radius_ratio', 'relative_size_squared', 'shape', 'shape_and_size', 'volume'],
>     'TRIANGLE': ['area', 'aspect_frobenius', 'aspect_ratio', 'condition', 'distortion', 'max_angle', 'min_angle', 'radius_ratio', 'relative_size_squared', 'scaled_jacobian', 'shape', 'shape_and_size'],
>     'TRIANGLE_STRIP': ['area'],
>     'WEDGE': ['condition', 'distortion', 'jacobian', 'max_aspect_frobenius', 'scaled_jacobian', 'shape', 'volume'],
> }
> ```

From this output we can also see that this filter is only really useful for eight cell types. And `PIXEL` and `TRIANGLE_TRIP` are very limited.

If we get the intersection of all other cell type measures, we get:
``` python
common_measures = set(has_measures['HEXAHEDRON'])
common_measures.intersection_update(has_measures['PYRAMID'])
common_measures.intersection_update(has_measures['QUAD'])
common_measures.intersection_update(has_measures['TETRA'])
common_measures.intersection_update(has_measures['TRIANGLE'])
common_measures.intersection_update(has_measures['WEDGE'])
print(common_measures) # {'shape'}
```

This shows that the `'shape'` measure is the most general. And so maybe there is an argument that this should be the default, instead of `'scaled_jacobian'` ?

user27182:
> We could add this to the docstring in this PR, or maybe document it in a follow-up PR?

I am working on a separate PR to address this.

> This shows that the `'shape'` measure is the most general. And so maybe there is an argument that this should be the default, instead of `'scaled_jacobian'` ?

We can address this in a separate PR.

user27182:
> > use None as default, users could specify the string instead of None to remove the warning
> 
> This one isn't good as users are almost always already specifying a parameter here and will be using the old data array name.
> 
> > move the warning to get_array and only warn if users try to access the array
> 
> This is an interesting suggestion. It would 'break' if anyone is creating this array outside this workflow. But seems unlikely?
> 
> My feeling is to keep the current PR strategy.

@MatthewFlamm Here's another option: deprecate `compute_cell_quality` and create a new `cell_quality` filter. That way, we can keep `str` inputs. And with #7311 , we can also generate categorical data, e.g. "good quality" and "bad quality" cells based on a particular measure's acceptable range. 
And so we can add a `acceptable_range` option to effectively auto-threshold the values. Set to:
- `None` by default (no threshold)
- `True`  to look up acceptable range using `pv.cell_quality_info`
- sequence to set the range manually

MatthewFlamm:
> Here's another option: deprecate compute_cell_quality and create a new cell_quality filter. 

I think this is the best approach.  We can get the best functionality without breaking change.

examples/00-load/read_dolfin.py
```diff
@@ -32,7 +32,7 @@
 
 # %%
 # Now we can work on and plot that Dolfin mesh.
-qual = dolfin.compute_cell_quality()
+qual = dolfin.cell_quality()
```
user27182, line 35:
```suggestion
qual = dolfin.cell_quality('scaled_jacobian')
```

The generated images from this example have changes because the new default measure is `'shape'`. We can add `'scaled_jacobia'` explicitly to keep the previous images.

pyvista/core/filters/data_object.py
```diff
@@ -2751,3 +2787,174 @@ def sample(  # type: ignore[misc]
                 raise VTKVersionError('`snap_to_closest_point=True` requires vtk 9.3.0 or newer')
         _update_alg(alg, progress_bar, 'Resampling array Data from a Passed Mesh onto Mesh')
         return _get_output(alg)
+
+    def cell_quality(  # type: ignore[misc]
+        self: _DataSetOrMultiBlockType,
+        quality_measure: Literal['all', 'all_valid']
+        | _CellQualityLiteral
+        | Sequence[_CellQualityLiteral] = 'shape',
```
user27182, line 2795: Note the new default to `'shape'` instead of `'scaled_jacobian'`. This was done based on this analysis: https://github.com/pyvista/pyvista/pull/7306#issuecomment-2725663835, which showed that this is the only valid measure for the supported cell types.

However, I'll also note that `'scaled_jacobian'` _should_ be valid  for the same 6 cell types for which `'shape'` is also valid. It appears that `'scaled_jacobian'`  is not valid for tetrahedral cells:

``` python
from pyvista import examples
tetra = examples.cells.Tetrahedron() 
qual = tetra.cell_quality('scaled_jacobian') 
qual.get_data_range()
# (np.float64(-1.0), np.float64(-1.0))
```
But, this measure _is_ explicitly listed for tetrahedral cells in https://public.kitware.com/Wiki/images/6/6b/VerdictManual-revA.pdf

And so I'm not sure why it's not working with `vtkCellQuality` for this cell type.

I do not have a deep understanding of these measures though so we can keep `'scaled_jacobian'` if that's understood to be a superior general-purpose measure.

MatthewFlamm: Something is weird here for this Tetrahedral cell.  The `shape` is `0` which is the bottom of the range from the pdf you linked.  Using the existing filter:

```python
>>> tetra = examples.cells.Tetrahedron() 
>>> tetra.compute_cell_quality('scaled_jacobian')['CellQuality']
pyvista_ndarray([-1.])
>>> tetra.compute_cell_quality('shape')['CellQuality']
pyvista_ndarray([0.])
```

MatthewFlamm: Also telling is this.  This example is cell is not defined correctly:

```py
>>> tetra.volume
-2.6666666666666665
```

MatthewFlamm: https://github.com/pyvista/pyvista/blob/fed059f6ca1cf778213447c2700c20a989fdcb3d/pyvista/examples/cells.py#L556-L561

The first three points form a base whose orthogonal vector points away from the fourth using right hand rule.  See https://vtk.org/doc/nightly/html/classvtkTetra.html#details

> The tetrahedron is defined by the four points (0-3); where (0,1,2) is the base of the tetrahedron which, using the right hand rule, forms a triangle whose normal points in the direction of the fourth point.

user27182: Swapping the first two points fixes this, and `'scaled_jacobian'` is 1.0 (and volume is positive)
I have reverted back to use this measure as default.

user27182: This issue with the cell examples is fixed by #7326
Until that PR merges, the auto-generated table in the `cell_quality` filter docs will incorrectly show that `scaled_jacobian` is not valid for tetra cells. But this issue should not block this PR.

akaszynski:
@MatthewFlamm, I'll defer to you to approve this, but IMO this is good to go.

pyvista/core/filters/data_object.py
```diff
@@ -2753,3 +2789,169 @@ def sample(  # type: ignore[misc]
...
+    ) -> _DataSetOrMultiBlockType:
+        r"""Compute a function of (geometric) quality for each cell of a mesh.
+
+        The per-cell quality is added to the mesh's cell data, in an array with
+        the same name as the quality measure. Cell types not supported by this
+        filter or undefined quality of supported cell types will have an
+        entry of ``-1``. See table below for all measures and the
+        :class:`~pyvista.CellType` supported by each one.
+
+        Defaults to computing the scaled jacobian quality measure.
```
MatthewFlamm, line 2810:
```suggestion
        entry of ``-1``.

        See table below for all measures and the
        :class:`~pyvista.CellType` supported by each one.
        Defaults to computing the scaled jacobian quality measure.
```

MatthewFlamm: I think we need to emphasis the table more.  Moving this sentence to start the paragraph below helps a bit in my opinion.

pyvista/core/filters/data_object.py
```diff
@@ -2753,3 +2789,169 @@ def sample(  # type: ignore[misc]
...
+
+        .. dropdown:: Cell Quality Measures
```
MatthewFlamm, line 2812: This table is essential to the documentation of this filter, I think we should use the `open` option to have it be 'dropped down' on open.

pyvista/core/filters/data_object.py
```diff
@@ -2753,3 +2789,169 @@ def sample(  # type: ignore[misc]
...
+            for low-level technical information about how each metric is computed,
+            which :class:`~pyvista.CellType` it applies to, as well as the metric's
+            full, normal, and acceptable range of values.
+
+        .. versionadded:: 0.45
+
+        Parameters
+        ----------
+        quality_measure : str | sequence[str], default: 'scaled_jacobian'
+            The cell quality measure to use. Specify a single measure or a sequence of
```
MatthewFlamm, line 2828: Is it possible to link to the dropdown or the table itself?

examples/01-filter/mesh_quality.py
```diff
@@ -6,29 +6,93 @@
 
 Leverage powerful VTK algorithms for computing mesh quality.
 
-Here we will use the :func:`pyvista.DataSetFilters.compute_cell_quality` filter
+Here we will use the :func:`~pyvista.DataObjectFilters.cell_quality` filter
 to compute the cell qualities. For a full list of the various quality metrics
 available, please refer to the documentation for that filter.
+
+Note that many of the measures may only apply to 2D cells or 3D cells. Examples for
```
MatthewFlamm, line 13: Can we add the Cell Quality Measures table into this example?  It is such a great table and it demonstrates this paragraph at a glance.  I'm fine having it either be dropdown closed or initialized upon open here.

MatthewFlamm (commented):
This is great overall!  The gallery examples for this are a really nice and showcase PyVista really well.

My only comment is that I find the list of cell quality measures hard to find in the documentation on first look through.  You have to expand this table to find the strings you can use.
image

Comments related to this. Otherwise it looks good.

user27182:
Docs have been updated. @MatthewFlamm I believe I addressed all your comments. Just need to wait for docs to build as I'm not sure if sphinx will be happy with adding a reference target to a dropdown.

MatthewFlamm (approved):
LGTM thanks!

tkoyama010 (approved):

PR #7308, opened 2025-03-14
Make `compute_cell_quality` a data object filter

user27182:
### Overview

Support `MultiBlock` inputs. Depends on #7306.


user27182:
This is now part of #7306

PR #7311, opened 2025-03-15
Add `pv.cell_quality_info` to retrieve info about measures used by cell quality filter

user27182:
### Overview

The technical reference from #7309 includes information about cell quality measures. For each measure, there is info about the:

- acceptable range
- normal range
- full_range
- reference value for the unit cell

This PR extracts all of this information from https://public.kitware.com/Wiki/images/6/6b/VerdictManual-revA.pdf and stores in it `CellQualityInfo` dataclass objects. The info can then be retrieved using:

``` python
import pyvista as pv
pv.cell_quality_info(pv.CellType.TRIANGLE, 'area')
```
Output:
``` python
CellQualityInfo(
  cell_type=<CellType.TRIANGLE: 5>, 
  measure='area', 
  acceptable_range=(0, inf), 
  normal_range=(0, inf), 
  full_range=(0, inf), 
  unit_cell_value=0.4330127018922193
)
```

For measures where the acceptable range is well defined (e.g. `(0.5, 1.0)`), this will enable users to easily identify and extract "good quality" cells (within the acceptable range) and "bad quality" cells (outside the acceptable range), and also compare cell quality values against a reference unit cell (i.e. compare to `unit_cell_value`)

PR #7312, opened 2025-03-15
Expose metadata from `CellType` enum

user27182:
### Overview

When the `CellType` enum is created, it generates RST for  each for each enum value from a number of fields. These fields are currently "hidden" inside the enum, and are not accessible otherwise. E.g. it is not possible to do `pv.CellType.VERTEX.cell_class`

This PR refactors the enum to move all the info into a separate `_CELL_TYPE_INFO` dict class, and then initializes the enum from the dict. This exposes all of the metadata so that we can do:

`_CELL_TYPE_INFO['VERTEX'].cell_class` to access the metadata associated with each cell type.

This is used by #7311 to validate cell quality measures.

### DIFF
--- a/pyvista/core/celltype.py
+++ b/pyvista/core/celltype.py
@@ -92,11 +92,10 @@ class _CellTypeTuple(NamedTuple):
     edges_override: Literal['variable', 'n/a'] | None = None
     faces_override: Literal['variable', 'n/a'] | None = None
 
-    ####################################################################################
-    # Linear cells
-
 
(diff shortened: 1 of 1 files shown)


user27182:
Docs are unchanged:

https://67d5ffd69fa679986b01ad6b--pyvista-dev.netlify.app/api/utilities/_autosummary/pyvista.celltype#pyvista.CellType

tkoyama010 (approved):
LGTM

PR #7313, opened 2025-03-15
Add `pv.vtk_verbosity` context manager for suppressing vtk `stderr` output

user27182:
### Overview

Sometimes it's helpful or necessary to modify stderr output from VTK. This PR adds a context manager to do this easily.

This can used by `extract_edges` and `contour_labels`, and can also be extend to `compute_cell_quality` (e.g. #7311 and #7306)

### DIFF
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -632,7 +632,7 @@ class _VTKVerbosity(contextlib.AbstractContextManager[None]):
 
     >>> mesh = pv.Sphere()
     >>> with pv.vtk_verbosity('off'):
-    ...     mesh.compute_cell_quality('volume')
+    ...     mesh = mesh.compute_cell_quality('volume')
 
     """
 
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -671,7 +671,7 @@ class _VTKVerbosity(contextlib.AbstractContextManager):
         if self._original_verbosity is not None:
             self._verbosity = self._original_verbosity
 
-    def __call__(self, verbosity: _VerbosityOptions):  # type:ignore[override]
+    def __call__(self, verbosity: _VerbosityOptions):
         """Call the context manager."""
         # Set the verbosity permanently
         self._original_verbosity = self._verbosity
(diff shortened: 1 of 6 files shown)


pyvista/core/_vtk_core.py
```diff
@@ -584,3 +585,124 @@ def VTKVersionInfo():
...
+
+    @staticmethod
+    def _validate_verbosity(
+        verbosity: _VerbosityOptions,
+    ) -> vtkLogger.Verbosity:
+        if isinstance(verbosity, vtkLogger.Verbosity):
+            return verbosity
+        else:
+            try:
+                return getattr(vtkLogger, f'VERBOSITY_{str(verbosity).upper()}')
```
user27182, line 647: Todo: do not allow "invalid" verbosity

tests/core/test_utilities.py
```diff
@@ -2002,3 +2003,51 @@ def prop(cls):
...
+        'warning',
+        'info',
+        'trace',
+        'max',
+        *range(10),
+        '0',
+        _vtk.vtkLogger.VERBOSITY_OFF,
+    ],
+)
+def test_vtk_verbosity_context(verbosity, modifies_verbosity):
```
tkoyama010, line 2030: The `modifies_verbosity` argument does not appear to be used. Can it be removed?

tests/core/test_utilities.py
```diff
@@ -2002,3 +2003,51 @@ def prop(cls):
...
+    initial_verbosity = vtk.vtkLogger.VERBOSITY_4
+    _vtk.vtkLogger.SetStderrVerbosity(initial_verbosity)
+    with pv.vtk_verbosity(verbosity):
+        ...
+    assert _vtk.vtkLogger.GetCurrentVerbosityCutoff() == initial_verbosity
+
+
+@pytest.mark.parametrize('verbosity', ['off', _vtk.vtkLogger.VERBOSITY_OFF])
+def test_vtk_verbosity_set_get(verbosity, modifies_verbosity):
```
tkoyama010, line 2039: Same with previous comment.

tkoyama010 (changes requested):

tests/core/test_utilities.py
user27182: This is a fixture used to store and restore the vtkLogger state since modifying it will modify it globally between tests. Can probably it move to a decorator instead? Something like `@pytest.mark.usefixtures("modifies_verbosity")`

Or enable autouse on the fixture?

tkoyama010 (approved):
LGTM

PR #7316, opened 2025-03-17
Allow pyvista to be used with -OO flag

ConnorNoddin:
### Overview

This change allows the `-OO` Python flag to be used with PyVista without errors. Previously, when running Python in optimized mode (`-OO`), PyVista would break due to missing docstrings (`NoneType` errors). This PR ensures that docstrings are handled safely when they are stripped.

### Details
- Fixes `doc_subs` to safely handle missing docstrings by replacing `None` with an empty string (`""`).
- Updates `__init_subclass__` to check for `None` before performing `.format(**subs)`, preventing `TypeError`.
- Ensures PyVista can run in optimized mode (`-OO`) without breaking.


user27182:
Thanks! Can we add a test for this? Maybe in `tests/test_init.py`?

ConnorNoddin:
> Thanks! Can we add a test for this? Maybe in `tests/test_init.py`?

Take a gander

user27182:
We're missing code coverage for the case where pyvista fails due to missing docstrings. Can we add another test for this? Maybe using `with pytest.raises(ImportError):`?

ConnorNoddin:
> We're missing code coverage for the case where pyvista fails due to missing docstrings. Can we add another test for this? Maybe using `with pytest.raises(ImportError):`?

Okay, is this better?

user27182:
> > We're missing code coverage for the case where pyvista fails due to missing docstrings. Can we add another test for this? Maybe using `with pytest.raises(ImportError):`?
> 
> Okay, is this better?

This did not resolve it. The coverage report is showing that the branches are covered for the `if True` case but not the `False` case. 

We can probably collapse the if branches into a single statement to get around this (because then it's considered a single line with no branching)

ConnorNoddin:
> > > We're missing code coverage for the case where pyvista fails due to missing docstrings. Can we add another test for this? Maybe using `with pytest.raises(ImportError):`?
> > 
> > 
> > Okay, is this better?
> 
> This did not resolve it. The coverage report is showing that the branches are covered for the `if True` case but not the `False` case.
> 
> We can probably collapse the if branches into a single statement to get around this (because then it's considered a single line with no branching)

That is what I had before (an inline if-else) https://github.com/pyvista/pyvista/pull/7316/commits/09bcdddd988085c2f13fe3a20785fd2ec7247599

user27182:
Now there are two missed lines. Before there was still one missed line. Both times code coverage was reporting failure. 

Either:
- Collapse the other branch as well
- Add test coverage for branches by simulating a `None` docstring and using `with pytest.raises`
- Use `# pragma: ignore` to ignore the missing test coverage

ConnorNoddin:
> Now there are two missed lines. Before there was still one missed line. Both times code coverage was reporting failure.
> 
> Either:
> 
> * Collapse the other branch as well
> * Add test coverage for branches by simulating a `None` docstring and using `with pytest.raises`
> * Use `# pragma: ignore` to ignore the missing test coverage

I collapsed the branches with an or statement

tests/test_init.py
```diff
@@ -36,3 +37,15 @@ def test_large_dependencies_not_imported(module: str):
...
     """
     assert not _module_is_loaded(module), error_msg
+
+
+def test_pyvista_oo_flag():
+    """Test that PyVista works correctly with the -OO optimization flag."""
+    code = """
+import pyvista as pv
+mesh = pv.Sphere()
+""".strip()
```
user27182, line 47: This code executes just fine using `-OO` _without_ this PR. The issue was identified with charts, so let's test that instead. This code _does_ currently fail:
```suggestion
    code = "from pyvista import Charts2D"
```

user27182: "import pyvista" works because there's a lazy importer that only imports modules when requested.
So a better test would probable be:

``` suggestion
    # Bypass lazy importer and directly import everything from core and plotting
    code = 'from pyvista.core import *; from pyvista.plotting import *'
```

user27182:
pre-commit.ci autofix

user27182 (approved):
Thanks! LGTM

tests/test_init.py
```diff
@@ -36,3 +37,12 @@ def test_large_dependencies_not_imported(module: str):
     https://github.com/pyvista/pyvista/pull/7023
     """
     assert not _module_is_loaded(module), error_msg
+
+
+def test_pyvista_oo_flag():
+    """Test that PyVista works correctly with the -OO optimization flag."""
+    code = 'from pyvista import Charts2D'
```
user27182, line 44:
```suggestion
    code = 'from pyvista import Chart2D'
```

user27182 (approved):

PR #7317, opened 2025-03-17
Add parameters to Polydata.triangulate filter

beroda:
### Overview

This PR adds `pass_verts` and `pass_lines` boolean parameters to `PolyData.triangulate`

A `preserve_polys` bool only available for `vtk >= 9.4` could be added once https://github.com/pyvista/pyvista/pull/6879 lands.
See the [vtk doc](https://vtk.org/doc/nightly/html/classvtkTriangleFilter.html#a36b585ab8fefc4d1dc36695a466ec0e6)


tkoyama010 (approved):
LGTM

pyvista/core/filters/poly_data.py
```diff
@@ -748,13 +748,29 @@ def plot_curvature(self, curv_type='mean', **kwargs):
         kwargs.setdefault('scalar_bar_args', {'title': f'{curv_type.capitalize()} Curvature'})
         return self.plot(scalars=self.curvature(curv_type), **kwargs)  # type: ignore[attr-defined]
 
-    def triangulate(self, inplace: bool = False, progress_bar: bool = False):
+    def triangulate(  # type: ignore[override]
+        self,
```
user27182, line 752:
```suggestion
        self,
        *,
```

user27182 (commented):
This is technically a breaking change since `inplace` and `progress_bar` are positional. I think this is an OK change though (and there is precedent for allowing this kind of breakage for progress bar). But then we might as well make it keyword only, especially if another keyword (preserve_polys) will be added.

pyvista/core/filters/poly_data.py
beroda: I agree, I also tend to prefer keyword-only since it is more explicit.

tkoyama010 (approved):

PR #7318, opened 2025-03-17
Add codecov test analytics

beroda:
### Overview

Add codecov new feature to display test results in the codecov PR comment.
See the corresponding documentation [here](https://docs.codecov.com/docs/test-analytics)

It could greatly help to diagnose test failures without needing to inspect testing jobs logs for various os (macOS, Linux, Windows) and python/vtk/numpy versions.


tests/test_init.py
```diff
@@ -18,6 +18,20 @@ def _module_is_loaded(module: str) -> bool:
...
+@pytest.mark.skipif(os.name == 'nt', reason='Test skipped on Windows')
+def test_failed_windows():
+    pytest.fail('test')
+
+
+@pytest.mark.skipif(sys.version_info >= (3, 11), reason='Test skipped for python version >= 3.11')
+def test_failed_python_version():
+    pytest.fail('test')
+
+
```
beroda, line 34: Adding some test that fail to see how the codecov test reports display them depending on python version and os.

PR #7319, opened 2025-03-18
Fix nested `vtk_verbosity` context

user27182:
### Overview

Follow-up to #7313. A small bug was identified when nesting contexts.


tkoyama010 (approved):
LGTM

Issue #7322, opened 2025-03-18
Make `to_meshio` compatible with mixed cell types that include `VOXEL`


GuillaumeBroggi:
### Describe the feature you would like to be added.

While the `to_meshio` method introduced in https://github.com/pyvista/pyvista/pull/7079 is compatible with voxels as in:

```python
import pyvista as pv
from pyvista import examples
mesh = examples.cells.Voxel()
pv.to_meshio(mesh)
```
```text
<meshio mesh object>
  Number of points: 8
  Number of cells:
    hexahedron: 1
```
it will fail if the cell types are mixes, e.g.:

```python
import pyvista as pv
from pyvista import examples
mesh = examples.cells.Voxel() + examples.cells.Quadrilateral()
pv.to_meshio(mesh)
```
```text
KeyError: np.uint8(11)
```
The error is related to the line 1135 below:
https://github.com/pyvista/pyvista/blob/807dd8a3011e5ca55804b929477f591846b1fc5c/pyvista/core/utilities/fileio.py#L1118-L1136

Inspecting `vtk_to_meshio_type` shows that there is no meshio type corresponding to voxels (might be related to https://github.com/nschloe/meshio/issues/690#issuecomment-586002489). Hence, the voxels should be processed as in the single cell type.

PR #7323, opened 2025-03-18
Fix `to_meshio` with mixed cell types that include `VOXEL`

GuillaumeBroggi:
### Overview

See #7322


keurfonluu:
Thanks for fixing that. The same issue for PIXEL, may you fix it in this PR as well?

GuillaumeBroggi:
> Thanks for fixing that. The same issue for PIXEL, may you fix it in this PR as well?

`vtk_to_meshio_type` contains `8: 'pixel'` so I assumed pixels were already supported by meshio, but I did not test it. Let me check if ti requires special care.

user27182 (commented):
nits

GuillaumeBroggi:
Thanks, @user27182; I applied your suggestions.

tkoyama010 (approved):
LGTM

PR #7326, opened 2025-03-19
Fix cell examples with negative volume

user27182:
### Overview

See https://github.com/pyvista/pyvista/pull/7306#discussion_r2004121010.

I added general tests for all cells and found that these had negative volume:

- Tetrahedron
- HexagonalPrism
- QuadraticWedge

This PR fixes this so computed volumes of all 3D cell examples are all positive.


user27182:
pre-commit.ci autofix

tkoyama010 (dismissed):
LGTM

tests/examples/test_cell_examples.py
```diff
@@ -1,10 +1,50 @@
...
+    assert isinstance(mesh, pyvista.UnstructuredGrid)
+    assert mesh.n_cells == 1
+
+    if mesh.celltypes[0] in [
+        pyvista.CellType.BIQUADRATIC_QUADRATIC_HEXAHEDRON,
+        pyvista.CellType.TRIQUADRATIC_HEXAHEDRON,
+    ]:
+        # Special cases, these cells have bugs, see https://gitlab.kitware.com/vtk/vtk/-/issues/19639
+        # Volume should be positive but returns zero
+        assert np.isclose(mesh.volume, 0.0)
```
MatthewFlamm, line 32: This is fine, but could consider marking them xfail. The nice thing about xfail is that if vtk fixes that bug we more quickly realize the intent.  But your comment does the same.  I have no opinion on which is preferred.

MatthewFlamm (dismissed):
One comment for consideration only.

tests/examples/test_cell_examples.py
user27182: xfail is better. And I noticed the vtk_needs_version marker can be updated to (9,1) instead of (9,3). Will push changes soon

tests/examples/test_cell_examples.py
```diff
@@ -1,10 +1,51 @@
...
+    assert mesh.n_cells == 1
+
+    if mesh.celltypes[0] in [
+        pyvista.CellType.BIQUADRATIC_QUADRATIC_HEXAHEDRON,
+        pyvista.CellType.TRIQUADRATIC_HEXAHEDRON,
+    ]:
+        pytest.xfail(
+            'Volume should be positive but returns zero, see https://gitlab.kitware.com/vtk/vtk/-/issues/19639'
+        )
+        assert np.isclose(mesh.volume, 0.0)
```
MatthewFlamm, line 33: We need to remove this line if we use `pytest.xfail` because this line would then fail, i.e. it would 'pass' the xfail check.
```suggestion
```

MatthewFlamm (approved):

PR #7327, opened 2025-03-19
Add codecov test analytics

beroda:
### Overview
(redo from https://github.com/pyvista/pyvista/pull/7318 and https://github.com/pyvista/pyvista/pull/7325 since the codecov comment was not updated)

Add codecov new feature to display test results in the codecov PR comment.
See the corresponding documentation [here](https://docs.codecov.com/docs/test-analytics)

It could greatly help to diagnose test failures without needing to inspect testing jobs logs for various os (macOS, Linux, Windows) and python/vtk/numpy versions.

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -158,7 +158,7 @@ jobs:
           pip install --pre --no-deps -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy
 
       - name: Core Testing (no GL)
-        run: python -m pytest --cov=pyvista --cov-branch -v --ignore=tests/plotting
+        run: python -m pytest tests/test_init.py --cov=pyvista --cov-branch --cov-report=xml -v --ignore=tests/plotting
 
       - uses: awalsh128/cache-apt-pkgs-action@v1.4.3
         with:
@@ -187,6 +187,7 @@ jobs:
 
       - uses: codecov/codecov-action@v4
         name: "Upload coverage to CodeCov"
+        if: always()
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


user27182:
> With that in mind, do you see any benefit of implementing the test reports feature as-is or should we wait for another version of the action ?

The current codecov comment as-is seems limited in usefulness. First, the hits/misses and actual code coverage seems to be missing (we should keep this), and second as you mentioned it does not currently show each combination of os/python/vtk/numpy versions.

A dashboard with a more detailed summary seems better IMO. In particular, I think total test time(s) for each OS/vtk version as well as the slowest tests is important to have. The vtk>=9.3 Linux tests in particular are taking much longer to execute lately than the other runs, and the docstring tests have also ballooned in time. Having these times summarized would help catch this and help prevent unexpected large increases to test times.

beroda:
> > With that in mind, do you see any benefit of implementing the test reports feature as-is or should we wait for another version of the action ?
> 
> The current codecov comment as-is seems limited in usefulness. First, the hits/misses and actual code coverage seems to be missing (we should keep this), and second as you mentioned it does not currently show each combination of os/python/vtk/numpy versions.
> 
> A dashboard with a more detailed summary seems better IMO. In particular, I think total test time(s) for each OS/vtk version as well as the slowest tests is important to have. The vtk>=9.3 Linux tests in particular are taking much longer to execute lately than the other runs, and the docstring tests have also ballooned in time. Having these times summarized would help catch this and help prevent unexpected large increases to test times.

You are right. Closing this PR while waiting for another version of Codecov feature.
Thanks @user27182 for your feedback.

user27182:
> Similar actions such as `Test reporter` (see [here](https://github.com/dorny/test-reporter/blob/main/assets/fluent-validation-report.png)) support such visualization.

I just came across Test Reporter again as I was looking into running tests against the sphinx gallery example computation times. The times are reported in a JUnit xml format:
https://sphinx-gallery.github.io/stable/configuration.html#using-junit-xml-files
and I think these can be read by Test reporter ? CircleCI is also mentioned in the provided link.

Might be worth getting something like this integrated into our workflows.

PR #7331, opened 2025-03-20
Fix `strips` equality for DataObject

beroda:
### Overview

`DataObject` equality does not take into account the `strips` attribute, which is a bug IMHO.
See the corresponding example below:

```python
import numpy as np

import pyvista as pv

# Create an array of points
points = np.array(
    [
        [1.0, 0.0, 0.0],
        [0.0, 0.0, 0.0],
        [1.0, 1.0, 0.0],
        [0.0, 1.0, 0.0],
        [1.0, 2.0, 0.0],
        [0.0, 2.0, 0.0],
        [1.0, 3.0, 0.0],
        [0.0, 3.0, 0.0],
    ],
)

strips = np.array([8, 0, 1, 2, 3, 4, 5, 6, 7])

strips = strips.copy()
strips[1:] = strips[:0:-1]
mesh2 = pv.PolyData(points, strips=strips)

strips = strips.copy()
strips[0] = 4
mesh3 = pv.PolyData(points, strips=strips[0:5])
````

Before this PR:
```python
mesh2 == mesh, mesh3 == mesh, mesh2 == mesh3
(True, True, True)
```

On this PR:
```python
mesh2 == mesh, mesh3 == mesh, mesh2 == mesh3
(False, False, False)
```


tkoyama010 (approved):
LGTM

PR #7332, opened 2025-03-20
Always run plotting tests

beroda:
### Overview

Currently, plotting tests are skipped when core tests are failing.
Having both run should help diagnose problems more quickly.

One downside is that CI running time is increased, but I think it is worth it IMHO.

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -69,6 +69,7 @@ jobs:
         run: pytest -v --ignore=tests/plotting
 
       - name: Test Plotting
+        if: always()
         run: pytest -v tests/plotting --fail_extra_image_cache --generated_image_dir debug_images
 
       - name: Upload Generated Images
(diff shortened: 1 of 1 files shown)


tests/test_init.py
```diff
@@ -18,6 +18,10 @@ def _module_is_loaded(module: str) -> bool:
     return os.system(f'{sys.executable} -c "{exe_str}"') != 0
 
 
+def test_failed():
+    pytest.fail('foo')
+
+
```
beroda, line 24: Adding failed test in core to demonstrate this PR impact.
To be removed when the PR is ready.

beroda:
Plotting tests are executed after core ones:
- mac-os: https://github.com/pyvista/pyvista/actions/runs/13980065310/job/39143145571?pr=7332#step:7:1
- linux: https://github.com/pyvista/pyvista/actions/runs/13980065310/job/39143145574?pr=7332#step:15:1
- windows: https://github.com/pyvista/pyvista/actions/runs/13980065310/job/39143145581?pr=7332#step:10:1

banesullivan (approved):
This makes a lot of sense, especially since "Upload Generated Images" already has an `always()` condition

PR #7333, opened 2025-03-20
Fix newly windows data path

beroda:
### Overview

In a very special scenario, I encountered an issue with `downloads` module on windows:

1. Fresh new Windows laptop with `pyvista` never installed at all
2. Running tests on a closed source project that imports the `pv.examples.download` module
3. Encountered the warning raised [here](https://github.com/pyvista/pyvista/blob/66b6b3c9d15b061637988567342663d2d358ec7f/pyvista/examples/downloads.py#L102)
4. Warning treated as error -> tests failed

I found the root cause being `pooch.os_cache` returning a nested path on Windows that would need `Path().mkdir(parents=True)` to work correctly.


.github/workflows/testing-and-deployment.yml
```diff
@@ -231,6 +231,9 @@ jobs:
       - name: Install PyVista test dependencies
         run: pip install .[test]
 
+      - name: Test user data path
+        run: pytest tests/test_user_path.py
+
```
beroda, line 236: Temporary to test encountered error

banesullivan:
Is this an issue that we should raise upstream in pooch?

beroda:
> Is this an issue that we should raise upstream in pooch?

Don't think so since from what I've seen in `pooch`, the user folder is correctly created when downloading the first file.
In our case, we try to create the folder before `pooch` does.

Once I manage to reproduce on CI, I'll set the `parents=True` at folder creation and it should be fine.

beroda:
The encountered behavior is observed in https://github.com/pyvista/pyvista/actions/runs/14006358501/job/39220725492?pr=7333

akaszynski (approved):
LGTM. This is a PyVista and not a `pooch` issue since we're creating the pyvista user data directory rather than relying on `pooch`.

PR #7335, opened 2025-03-21
Modify cell examples to define unit cells

user27182:
### Overview

Modify the following cell examples to make them unit cells:
- define a unit `Triangle` as equilateral with unit edge length
- define a unit `Tetrahedron` with unit edge lengths
- define a unit `Pyramid` as a pyramid whose triangular faces are equilateral triangles with unit edge length
- define a unit `Wedge` as a wedge whose quadrilateral faces are unit squares with the 2 triangular faces as equilateral triangles with unit edge length

Tests for `Quad` and `Hexahedron` are also added to ensure they are also unit cells.

In the context of measuring cell quality with `vtkCellQuality`, these six cell types have an expected "unit cell value" as reported by https://public.kitware.com/Wiki/images/6/6b/VerdictManual-revA.pdf. As such, it is useful to have a reference cell to use and compare to when assessing cell quality.


tkoyama010 (approved):
LGTM

PR #7338, opened 2025-03-24
Add `mask_points` filter to extract points within an image/volume mask

tkoyama010:
### Overview

Add `mask_points` filter to extract points within an image/volume mask.

```python
        >>> import pyvista as pv
        >>> sphere = pv.Sphere()
        >>> mask = sphere.points[:, 0] > 0
        >>> masked = sphere.mask_points(mask)
        >>> masked.plot()
```

### Details

- None


tkoyama010:
pre-commit.ci autofix

PR #7343, opened 2025-03-25
Remove `None` from `MultiBlock.keys()` typing

user27182:
### Overview

The `MultiBlock` keys are always strings and should hint the type as `list[str]`. But the typing hints `list[str | None]` instead because `get_block_name` returns `str | None`. But the `None` case is only valid here when `GetMetaData` returns `None`, which is only the case when a MultiBlock is empty. However, if it's empty then `get_block_name` raises an `IndexError` before we even call `GetMetaData`, and so the `None` case is not valid and `None` will never be returned.

This PR therefore removes `None` from the return type of `get_block_name` and `keys`.


tkoyama010 (approved):
LGTM

PR #7356, opened 2025-03-29
Skip failing tests on Windows for VTK 9.4

banesullivan:
Skip failing tests on Windows for VTK 9.4 per https://github.com/pyvista/pyvista/pull/6879#issuecomment-2692483666

### DIFF
--- a/tests/plotting/test_plotting.py
+++ b/tests/plotting/test_plotting.py
@@ -1578,6 +1578,7 @@ def test_vector_array_with_points(multicomp_poly):
     pl.show()
 
 
+@skip_windows_mesa
 def test_vector_array_with_cells(multicomp_poly):
     """Test using vector valued data with and without component arg."""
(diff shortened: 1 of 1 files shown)


user27182 (approved):
I think it's fine to skip these. The failures seem to be environment and/or machine specific and may not impact all windows users. Plus our hands may be tied anyway in terms of supporting any fixes since the root cause of these failures is with VTK.

banesullivan:
No idea why the pyqt integration test is segfaulting here 😬 ... going to merge this into #6879 anyways

user27182:
> No idea why the pyqt integration test is segfaulting here 😬 ... going to merge this into #6879 anyways

This is very odd indeed. Why would adding a decorator to a pyvista test to skip on Windows affect a pyvistaqt test running on Linux ? 🤔

We might be stuck having to fix this...

Issue #5813, opened 2024-03-21
Plotting the `download_parched_canal_4k` dataset shows the texture as a red square


user27182:
### Describe the bug, what's wrong, and what you expected.

Running the docstring example for [download_parched_canal_4k](https://docs.pyvista.org/version/stable/api/examples/_autosummary/pyvista.examples.downloads.download_parched_canal_4k.html#pyvista.examples.downloads.download_parched_canal_4k) shows a red square:

![](https://d33wubrfki0l68.cloudfront.net/00d2e4017d9135f586fab69ae02e1dd14b0b724a/e2e07/_images/pyvista-examples-downloads-download_parched_canal_4k-1_00_00.png)

This red square is showing up in the docs and I also get a red square running the example locally on my machine as well.

Interestingly, the "Interactive Scene" in the docs appears to show the texture correctly.

Note: 
This bug was found as part of local testing for #5787.
Cross ref to #5718 , which could help catch any future regressions in the docs images.

### Steps to reproduce the bug.

```python
from pyvista import examples
dataset = examples.download_parched_canal_4k()
dataset.plot(cpos="xy")
```

### System Information

```shell
pyvista 0.44.dev0
```

PR #7357, opened 2025-03-30
Skip plotting examples which use `set_environment_texture`

user27182:
### Overview

This method seems to hang in CI and increases docstring tests times and doc build times considerable. See https://github.com/pyvista/pyvista/pull/6993#issuecomment-2548444061

### DIFF
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -3540,12 +3540,14 @@ class Renderer(_vtk.vtkOpenGLRenderer):
 
         >>> from pyvista import examples
         >>> import pyvista as pv
-        >>> pl = pv.Plotter(lighting=None)
-        >>> cubemap = examples.download_sky_box_cube_map()
-        >>> _ = pl.add_mesh(pv.Sphere(), pbr=True, metallic=0.9, roughness=0.4)
-        >>> pl.set_environment_texture(cubemap)
-        >>> pl.camera_position = 'xy'
-        >>> pl.show()
+        >>> pl = pv.Plotter(lighting=None)  # doctest:+SKIP
+        >>> cubemap = examples.download_sky_box_cube_map()  # doctest:+SKIP
+        >>> _ = pl.add_mesh(
+        ...     pv.Sphere(), pbr=True, metallic=0.9, roughness=0.4
+        ... )  # doctest:+SKIP
+        >>> pl.set_environment_texture(cubemap)  # doctest:+SKIP
+        >>> pl.camera_position = 'xy'  # doctest:+SKIP
+        >>> pl.show()  # doctest:+SKIP
 
         """
         # cube_map textures cannot use spherical harmonics
@@ -3567,12 +3569,14 @@ class Renderer(_vtk.vtkOpenGLRenderer):
         >>> from pyvista import examples
         >>> import pyvista as pv
         >>> pl = pv.Plotter(lighting=None)
-        >>> cubemap = examples.download_sky_box_cube_map()
-        >>> _ = pl.add_mesh(pv.Sphere(), pbr=True, metallic=0.9, roughness=0.4)
-        >>> pl.set_environment_texture(cubemap)
-        >>> pl.remove_environment_texture()
-        >>> pl.camera_position = 'xy'
-        >>> pl.show()
+        >>> cubemap = examples.download_sky_box_cube_map()  # doctest:+SKIP
+        >>> _ = pl.add_mesh(
+        ...     pv.Sphere(), pbr=True, metallic=0.9, roughness=0.4
+        ... )  # doctest:+SKIP
+        >>> pl.set_environment_texture(cubemap)  # doctest:+SKIP
+        >>> pl.remove_environment_texture()  # doctest:+SKIP
+        >>> pl.camera_position = 'xy'  # doctest:+SKIP
+        >>> pl.show()  # doctest:+SKIP
 
         """
         self.UseImageBasedLightingOff()
--- a/examples/00-load/load_gltf.py
+++ b/examples/00-load/load_gltf.py
@@ -30,11 +30,11 @@ texture = examples.download_dikhololo_night()
 # sphinx_gallery_start_ignore
 PYVISTA_GALLERY_FORCE_STATIC = True
 # sphinx_gallery_end_ignore
-pl = pyvista.Plotter()
-pl.import_gltf(helmet_file)
-pl.set_environment_texture(texture)
-pl.camera.zoom(1.7)
-pl.show()
+# pl = pyvista.Plotter()
+# pl.import_gltf(helmet_file)
+# pl.set_environment_texture(texture)
+# pl.camera.zoom(1.7)
+# pl.show()
 
 
 # %%
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -4277,11 +4277,11 @@ def download_sky_box_cube_map(load=True):  # pragma: no cover
     --------
     >>> from pyvista import examples
     >>> import pyvista as pv
-    >>> pl = pv.Plotter()
-    >>> dataset = examples.download_sky_box_cube_map()
-    >>> _ = pl.add_actor(dataset.to_skybox())
-    >>> pl.set_environment_texture(dataset)
-    >>> pl.show()
+    >>> pl = pv.Plotter()  # doctest:+SKIP
+    >>> dataset = examples.download_sky_box_cube_map()  # doctest:+SKIP
+    >>> _ = pl.add_actor(dataset.to_skybox())  # doctest:+SKIP
+    >>> pl.set_environment_texture(dataset)  # doctest:+SKIP
+    >>> pl.show()  # doctest:+SKIP
 
     .. seealso::
 
@@ -4345,14 +4345,17 @@ def download_cubemap_park(load=True):  # pragma: no cover
     --------
     >>> from pyvista import examples
     >>> import pyvista as pv
-    >>> pl = pv.Plotter(lighting=None)
     >>> dataset = examples.download_cubemap_park()
-    >>> _ = pl.add_actor(dataset.to_skybox())
-    >>> pl.set_environment_texture(dataset, True)
-    >>> pl.camera_position = 'xy'
-    >>> pl.camera.zoom(0.4)
-    >>> _ = pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.1, metallic=0.5)
-    >>> pl.show()
+    >>>
+    >>> pl = pv.Plotter(lighting=None)  # doctest:+SKIP
+    >>> _ = pl.add_actor(dataset.to_skybox())  # doctest:+SKIP
+    >>> pl.set_environment_texture(dataset, True)  # doctest:+SKIP
+    >>> pl.camera_position = 'xy'  # doctest:+SKIP
+    >>> pl.camera.zoom(0.4)  # doctest:+SKIP
+    >>> _ = pl.add_mesh(
+    ...     pv.Sphere(), pbr=True, roughness=0.1, metallic=0.5
+    ... )  # doctest:+SKIP
+    >>> pl.show()  # doctest:+SKIP
 
     .. seealso::
 
@@ -4405,12 +4408,14 @@ def download_cubemap_space_4k(load=True):  # pragma: no cover
     >>> import pyvista as pv
     >>> from pyvista import examples
     >>> cubemap = examples.download_cubemap_space_4k()
-    >>> pl = pv.Plotter(lighting=None)
-    >>> _ = pl.add_actor(cubemap.to_skybox())
-    >>> pl.set_environment_texture(cubemap, True)
-    >>> pl.camera.zoom(0.4)
-    >>> _ = pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.24, metallic=1.0)
-    >>> pl.show()
+    >>> pl = pv.Plotter(lighting=None)  # doctest:+SKIP
+    >>> _ = pl.add_actor(cubemap.to_skybox())  # doctest:+SKIP
+    >>> pl.set_environment_texture(cubemap, True)  # doctest:+SKIP
+    >>> pl.camera.zoom(0.4)  # doctest:+SKIP
+    >>> _ = pl.add_mesh(
+    ...     pv.Sphere(), pbr=True, roughness=0.24, metallic=1.0
+    ... )  # doctest:+SKIP
+    >>> pl.show()  # doctest:+SKIP
 
     .. seealso::
 
@@ -4469,12 +4474,14 @@ def download_cubemap_space_16k(load=True):  # pragma: no cover
     >>> import pyvista as pv
     >>> from pyvista import examples
     >>> cubemap = examples.download_cubemap_space_4k()
-    >>> pl = pv.Plotter(lighting=None)
-    >>> _ = pl.add_actor(cubemap.to_skybox())
-    >>> pl.set_environment_texture(cubemap, True)
-    >>> pl.camera.zoom(0.4)
-    >>> _ = pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.24, metallic=1.0)
-    >>> pl.show()
+    >>> pl = pv.Plotter(lighting=None)  # doctest:+SKIP
+    >>> _ = pl.add_actor(cubemap.to_skybox())  # doctest:+SKIP
+    >>> pl.set_environment_texture(cubemap, True)  # doctest:+SKIP
+    >>> pl.camera.zoom(0.4)  # doctest:+SKIP
+    >>> _ = pl.add_mesh(
+    ...     pv.Sphere(), pbr=True, roughness=0.24, metallic=1.0
+    ... )  # doctest:+SKIP
+    >>> pl.show()  # doctest:+SKIP
 
     .. seealso::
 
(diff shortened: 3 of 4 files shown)


user27182:
doctests run in ~19 minutes now instead of ~55 minutes. This confirms that `set_environment_texture` is the issue.
https://github.com/pyvista/pyvista/actions/runs/14151693064/job/39645683775?pr=7357

user27182:
I have a working fix. I am not sure how we want to roll this out. Some options:

1. Enable this fix with an environment variable (e.g. `os.environ['PYVISTA_DOWNSAMPLE_ENVIRONMENT_TEXTURE'] = 'true')`. This fix is likely specific to our CI environment, so with this option we _always_ downsample for CI purposes. We could add a github label to disable this (e.g. when building from main or for releases) but otherwise have this set true by default.
2. Add this as a keyword option for `set_environment_texture`, e.g. `image_lighting_sampling_rate=None`. This is better because it avoids managing env variables but is not ideal because then we always need to remember to set this value (e.g. `image_lighting_sampling_rate=0.1` when writing new example. This may also have limited use since it's only for specific environments.

pyvista/plotting/renderer.py
```diff
@@ -3556,7 +3556,21 @@ def set_environment_texture(self, texture, is_srgb=False) -> None:
                 self.UseSphericalHarmonicsOff()
 
         self.UseImageBasedLightingOn()
-        self.SetEnvironmentTexture(texture, is_srgb)
+        # if os.environ.get('PYVISTA_DOCUMENTATION_BULKY_IMPORTS_ALLOWED'):
+        if True:
```
user27182, line 3560: This is temporary. Will be updated once there's feedback on how to manage this fix, see
https://github.com/pyvista/pyvista/pull/7357#issuecomment-2767685837

user27182:
For reference, with this PR:
- docstring tests now take 25 mins (down from 55 mins)
- doc build takes 1h50 mins, down from 3hrs

akaszynski:
> I have a working fix. I am not sure how we want to roll this out. Some options:
> 
> 1. Enable this fix with an environment variable (e.g. `os.environ['PYVISTA_DOWNSAMPLE_ENVIRONMENT_TEXTURE'] = 'true')`. This fix is likely specific to our CI environment, so with this option we _always_ downsample for CI purposes. We could add a github label to disable this (e.g. when building from main or for releases) but otherwise have this set true by default.
> 2. Add this as a keyword option for `set_environment_texture`, e.g. `image_lighting_sampling_rate=None`. This is better because it avoids managing env variables but is not ideal because then we always need to remember to set this value (e.g. `image_lighting_sampling_rate=0.1` when writing new example. This may also have limited use since it's only for specific environments.

Leaning towards 2. It's verbose. To avoid forgetting, let's warn (and not fail) when within CI. In fact, let's define a new warning like `PyVistaTestingWarning`.

---

Really appreciate reducing the CI run times. This has been a big hindrance to development and I'd like to keep driving those times down.

user27182:
> Leaning towards 2. It's verbose and explicit.

Thanks for the feedback, will work on making this change. Will also need to update `Texture.plot` kwargs for this to make this option available with that API.

> To avoid forgetting, let's warn (and not fail) when within CI. In fact, let's define a new warning like `PyVistaTestingWarning`.

Good idea, will add this too. Might need to add a new env var or global var, e.g. `PYVISTA_BUILDING_DOCUMENTATION`

This testing will enforce the downsampling, which is good, but is not robust in terms of actually enforcing computation times. Longer-term, as a follow up PR we can consider enforcing build times as follows:

- for the sphinx gallery, we can parse the auto-generated computation times table and generate test cases for each entry, and compare times to a threshold limit, and fail the test if computation time for any example exceeds the threshold 

- for regular docstring build times, we can enforce this using the sphinx duration extension. Similar to above, we parse the durations, build test cases, and fail if any docstring example takes too long to run (this would be limited to the top 5 longest run times, unless we want to re-consider merging #7010)

user27182:
pre-commit.ci autofix

akaszynski (dismissed):
This is fantastic. I have a closed source project that sometimes uses PBR and there are times in which it's running on weak GPUs (or no GPU).

Thanks, LGTM

user27182:
Also, these examples are still the slowest to run in CI.

new durations from https://github.com/pyvista/pyvista/actions/runs/14184108844/job/39736183963?pr=7357#step:11:8745

```
====================== slowest reading durations =======================
63.921 api/examples/_autosummary/pyvista.examples.downloads.download_cubemap_space_16k
63.870 api/examples/_autosummary/pyvista.examples.downloads.download_cubemap_space_4k
62.112 api/examples/_autosummary/pyvista.examples.gltf.download_damaged_helmet
61.169 api/examples/_autosummary/pyvista.examples.downloads.download_sky_box_cube_map
60.977 api/plotting/_autosummary/pyvista.Plotter.set_environment_texture
```

This is much better than originally reported from https://github.com/pyvista/pyvista/pull/6993#issuecomment-2548444061, but still, I wonder if we should try `resample=0.05` by default instead of `0.1` ?

MatthewFlamm:
I have not read these changes in detail, so a bit of a drive by comment.  Would this be a good candidate for being a theme config?  We could have a `_PyVistaDocTheme` too if needed.

MatthewFlamm:
Almost 33% reduction in doc build is phenomenal!  Thanks for finding this and fixing it.

MatthewFlamm:
> I have not read these changes in detail, so a bit of a drive by comment. Would this be a good candidate for being a theme config? We could have a `_PyVistaDocTheme` too if needed.

After reading the changes in more detail, I do think that a theme config would largely solve most of these problems including having a consistent default state for the documentation and also allow downstream packages to choose their own values using a theme.  However I also don't want to slow down this PR, we could consider making the env var private looking `_PYVISTA_DOCUMENTATION_CI` or something like this, and then implement the theme config later.

akaszynski:
>  I wonder if we should try resample=0.05 by default instead of 0.1 ?

Absolutely fine with 0.05. Also, we have a testing theme, let's just set some global testing defaults and use those. We could also continue the env var approach, I'm fine with either one, just as long as we document it in contributing.

.github/workflows/docs.yml
```diff
@@ -36,6 +36,7 @@ env:
   PYDEVD_DISABLE_FILE_VALIDATION: "1"
   PYTEST_ADDOPTS: "--color=yes"
   FORCE_COLOR: "True"
+  PYVISTA_DOCUMENTATION_CI: "true" # Flag for running CI-specific code
```
user27182, line 39:
```suggestion
  _PYVISTA_DOCUMENTATION_CI: "true" # Flag for running CI-specific code
```

.github/workflows/style_docstr.yml
```diff
@@ -9,6 +9,7 @@ concurrency:
 
 env:
   PYTEST_ADDOPTS: "--color=yes"
+  PYVISTA_DOCUMENTATION_CI: "true" # Flag for running CI-specific code
```
user27182, line 12:
```suggestion
  _PYVISTA_DOCUMENTATION_CI: "true" # Flag for running CI-specific code
```

pyvista/plotting/renderer.py
```diff
@@ -3556,7 +3575,31 @@ def set_environment_texture(self, texture, is_srgb=False) -> None:
                 self.UseSphericalHarmonicsOff()
 
         self.UseImageBasedLightingOn()
-        self.SetEnvironmentTexture(texture, is_srgb)
+        running_ci = os.environ.get('PYVISTA_DOCUMENTATION_CI', 'false').lower() == 'true'
```
user27182, line 3578:
```suggestion
        running_ci = os.environ.get('_PYVISTA_DOCUMENTATION_CI', 'false').lower() == 'true'
```

pyvista/plotting/renderer.py
```diff
@@ -3531,6 +3535,21 @@ def set_environment_texture(self, texture, is_srgb=False) -> None:
             texture or set this parameter to ``True``. Textures are assumed
             to be in linear color space by default.
 
+        resample : bool | float, default: False
+            Resample the environment texture. Set this to a float to set the
+            sampling rate explicitly or set to ``True`` to downsample the
+            texture to 1/10th of its original resolution. Downsampling the
+            texture can substantially improve performance for some environments
+            (e.g. headless setups or with if GPU support is limited).
```
user27182, line 3543:
```suggestion
            (e.g. headless setups or if GPU support is limited).
```

tests/plotting/test_plotting.py
```diff
@@ -321,12 +323,31 @@ def test_remove_environment_texture_cubemap(sphere):
...
+def documentation_ci_env_var():
+    var = 'PYVISTA_DOCUMENTATION_CI'
+    os.environ[var] = 'true'
+    yield
+    del os.environ[var]
+
+
+@pytest.mark.usefixtures('documentation_ci_env_var')
+def test_set_environment_texture_warns_ci(sphere):
+    """Test remove_environment_texture with a cubemap."""
```
user27182, line 342:
```suggestion
```

user27182 (commented):
Make env var private and fixup some of the doc

user27182:
> >  I wonder if we should try resample=0.05 by default instead of 0.1 ?
> 
> 
> 
> Absolutely fine with 0.05. Also, we have a testing theme, let's just set some global testing defaults and use those. We could also continue the env var approach, I'm fine with either one, just as long as we document it in contributing.

Didn't see this earlier. The theme approach is nice and easy to configure, so let's just use that after all. I'll update this PR accordingly. And experiment with smaller sampling rate. 


The current implementation currently forces `resample=True` to be made explicit, else emit a PyVistaTestWarning. Do we still want those changes to be explicit (e.g. so users are aware of it)? Or do we just set resample in the background to whatever is defined by the theme?

MatthewFlamm:
> The current implementation currently forces `resample=True` to be made explicit, else emit a PyVistaTestWarning. Do we still want those changes to be explicit (e.g. so users are aware of it)? Or do we just set resample in the background to whatever is defined by the theme?

@akaszynski should weight in, but from my perspective, if we make it opt-in (as originally planned), we will forget to opt-in in a future test.  If it is part of the testing theme, we will have to explicitly opt-out during testing.  This serves the same purpose IMO.  The env var or the testing theme approach will enable this.

We could expose our testing theme as a nonprivate variable if others want to use it.  This was a part of #5119 but I have been losing steam to finish it.

akaszynski:
Would prefer a testing theme (or doc theme), not an opt-in or env var if at all possible.

MatthewFlamm:
> Would prefer a testing theme (or doc theme), not an opt-in or env var if at all possible.

While we currently modify the global theme after being set to `document` theme for building documentation, we should instead have a `pyvista_document` theme that we use and share with other downstream users.  It would also make it easier for newer contributors to find it too.  This is outside the scope of this PR, so I captured it in #7389.

user27182:
Something isn't right, the docstring check is taking one hour. I'm thinking the theme isn't being applied correctly for this

MatthewFlamm:
> Something isn't right, the docstring check is taking one hour. I'm thinking the theme isn't being applied correctly for this

It turns out the gallery is using a _different_ theme right now.

https://github.com/pyvista/pyvista/blob/26909177f14bc262bc7f5488295e9ad09b353ea2/doc/source/conf.py#L455

user27182:
The theme is now working for docstring tests and sphinx gallery, but not for the docstring build.

user27182:
> The theme is now working for docstring tests and sphinx gallery, but not for the docstring build.

I think perhaps lines like this
https://github.com/pyvista/pyvista/blob/8f82547c4c4a3a62eccac11c9e1a9272f3ef1638/examples/99-advanced/extending_pyvista.py#L32

may be resetting the theme globally and setting `resample_environment_texture=False` as a result, disabling it for the rest of the build

MatthewFlamm:
It seems likely to me that the theme set in `doc/source/conf.py` never persisted throughout the whole doc build and is now being exposed in this PR.  How do the themes themselves get reset during docstring building?  There is a fixture in `conftest.py` in `pyvista/` but this will only run during doctest.

user27182:
> > may be resetting the theme globally and setting resample_environment_texture=False as a result,
> 
> Weird, the addition from [b0657c0](https://github.com/pyvista/pyvista/commit/b0657c00dbd59f4e7717069d0ec763dff5abec76) should be be run before and after every example.

Yes you're right, the gallery has before and after events which run `reset_pyvista` so the gallery is all good.

I had another look through `conf.py` and think maybe the plot directive is resetting the theme here:

https://github.com/pyvista/pyvista/blob/a43b10fcf846b339d5475404fb8f3cafe63fcf32/doc/source/conf.py#L503-L508

This is a strong signal that a new documentation-build theme from https://github.com/pyvista/pyvista/issues/7389 may be needed here so that we can do:
``` diff
- __s_p_t('document')
+ __s_p_t('doc_build')
``` 

(assuming we call the theme `'doc_build'`)

MatthewFlamm:
> I had another look through `conf.py` and think maybe the plot directive is resetting the theme here:

So many places that the theme has to be reset.  Good find.

> This is a strong signal that a new documentation-build theme from #7389 may be needed here so that we can do:

Even if there is a way to do it without the documentation-building theme, it is turning out to be a nightmare to keep all these pieces in sync, so I agree that #7389 is becoming higher priority.

pyvista/conftest.py
```diff
@@ -19,5 +19,7 @@ def reset_gloal_theme():
     """Reset global_theme."""
     # this stops any doctest-module tests from overriding the global theme and
     # creating test side effects
+    pyvista.set_plot_theme('document')
+    pyvista.global_theme.resample_environment_texture = True  # Speed up CI
```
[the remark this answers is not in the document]
MatthewFlamm: The original `restore_defaults` makes it hard to know what the theme is prior to this PR.  I assume it is currently `document`.  If so, I would keep the documentation building and doctest versions the same theme, i.e. `document_build`. Having different themes for those two cases will introduce cases where it works on one but not the other.

pyvista/plotting/themes.py
```diff
@@ -3384,6 +3425,21 @@ def __init__(self):
...
+
+    def __init__(self):
+        """Initialize the theme."""
+        super().__init__()
+        self.name = 'document_build'
+        self.window_size = [1024, 768]
+        self.font.size = 22
+        self.font.label_size = 22
+        self.font.title_size = 22
+        self.return_cpos = False
```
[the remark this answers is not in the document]
user27182: Yes I'm thinking these settings (originally from conf.py) were never applied, since `reset_pyvista` would have reset this to vanilla `'document'` for the gallery, and it seems the plot directive was also using `'document'` for the rest of the docs. So I'm thinking all images in the cache currently reflect the `'document'` style, without these other settings. Maybe `'document_build'` should only have `resample_environment_texture = True` as the only difference? 

But maybe the docs will look better with these changes. Like you said, let's wait for CI.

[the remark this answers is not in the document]
user27182: The change worked, and the new theme was applied to all the docs. The larger fonts are probably good to have, but creates many failures with the cache, so I'm going to leave this alone for this PR.

Let's keep the new `'document_build'` theme, but only add `'resample_environment_texture'` as the main modification for now.

user27182:
This is once again ready for review. The theme is now applied consistently for all document testing and building, and build time is < 2hrs, docstring tests about 25 mins.

The build is expected to fail once more to generate new cache images but otherwise this PR should be good to go.

user27182:
> Also, these examples are still the slowest to run in CI.
> 
> new durations from https://github.com/pyvista/pyvista/actions/runs/14184108844/job/39736183963?pr=7357#step:11:8745
> 
> ```
> ====================== slowest reading durations =======================
> 63.921 api/examples/_autosummary/pyvista.examples.downloads.download_cubemap_space_16k
> 63.870 api/examples/_autosummary/pyvista.examples.downloads.download_cubemap_space_4k
> 62.112 api/examples/_autosummary/pyvista.examples.gltf.download_damaged_helmet
> 61.169 api/examples/_autosummary/pyvista.examples.downloads.download_sky_box_cube_map
> 60.977 api/plotting/_autosummary/pyvista.Plotter.set_environment_texture
> ```
> 
> This is much better than originally reported from [#6993 (comment)](https://github.com/pyvista/pyvista/pull/6993#issuecomment-2548444061), but still, I wonder if we should try `resample=0.05` by default instead of `0.1` ?

new sampling rate is now 1/16 (`0.0625`), with these new build times:

``` 
====================== slowest reading durations =======================
47.300 api/examples/dataset_gallery
45.293 api/core/_autosummary/pyvista.Texture.plot
44.820 api/examples/_autosummary/pyvista.examples.downloads.download_cubemap_space_16k
44.706 api/examples/_autosummary/pyvista.examples.downloads.download_cubemap_space_4k
43.235 api/examples/_autosummary/pyvista.examples.gltf.download_damaged_helmet
```

These are still among the slowest examples, but I think this is OK for now. This can also be tuned later if needed once we enable testing of the durations to enforce an upper limit.

tests/doc/doc_image_cache/sphx_glr_load_gltf_thumb.jpg
MatthewFlamm, line 1: Why are these images changed?  They look much better, but it looks unrelated to resampling?

user27182: Don't really know, but the implementation for `set_environment_texture` now uses `SetBackground`, perhaps it's related to that change? This could be related to https://github.com/pyvista/pyvista/issues/5813 as well, where perhaps some textures only appear red under specific circumstances.

For this particular example, `download_dikhololo_night()` texture is used for the environment texture. If we download that and do a simple `texture.plot()`, it too appears as a red rectangle. And so previously, a pure red image was being used for rendering the reflections, but now maybe we fixed this, and the environment texture is being rendered correctly with `SetBackground`.

MatthewFlamm (approved):

PR #7358, opened 2025-03-30
Fix usage of LookupTable with Volume Rendering

banesullivan:
Volume rendering has been more or less broken up until now in that it did not properly map scalars values. Several hacks were implemented to work around what was ultimately just a bug in our understanding of how to generate the needed `vtkColorTransferFunction` and `vtkPiecewiseFunction` for volume rendering.

Fix `LookupTable.to_color_tf` and `LookupTable.to_opacity_tf` to properly build their `vtkColorTransferFunction` and `vtkPiecewiseFunction`. We were incorrectly setting the X points on these functions to an `(0, n_colors)` range. This is incorrect and the X coordinate is supposed to be in the space of the mapped values such that if you are mapping values from 0.005 to 0.0333, a linear space between those points in that space should be used. This is outlined well in countless examples of volume rendering in VTK however confusing because so many of those examples are using medical image datasets whose value range is typically 0-255. Further more, we were scaling the opacity values by `n_values`, again incorrect as these need to be between 0-1 for the `vtkPiecewiseFunction` and the alpha values are originally 0-255 values. This happened to work because our default `n_colors` was 256, but it would yield wildly incorrect results if using a different `n_colors` (see some of the image regression test change like `tests/plotting/image_cache/volume_rendering_from_plotter.png`. 

Some of these issues were highlighted in https://github.com/pyvista/pyvista/discussions/4700 and this pull requests now solves the issue there by outright removing the section of `add_volume` code that would clip the scalars and remap them from 0-255 to workaround the bugs outlined above.

Volume rendering is entirely different from surface rendering and how it uses the lookup table. With volume rendering, we have to convert the LookupTable to a `vtkColorTransferFunction` and `vtkPiecewiseFunction` and set those on the `vtkVolumeProperty` (very different from surface rendering where the lookup table is set on the mapper -- in volume rendering, the lookup table on the mapper is not used). However, we might expect that modifying the `LookupTable` we have with volume rendering should update scene. To do this, I've added a weakref to the `LookupTable` used on the `VolumeProperty` and track any changes to the `LookupTable` with a `ModifiedEvent` observer to re-apply the `LookupTable` to the `VolumeProperty`. This lets us modify the scalar mapping after adding a volume to a plot as we would for surface rendering:

```py
import pyvista as pv
from pyvista import examples

vol = examples.download_knee_full()

pl = pv.Plotter()
actor = pl.add_volume(vol, cmap="viridis")
actor.mapper.lookup_table.cmap = "coolwarm"
pl.show()
```


banesullivan:
A few image regression failures from the doc build:

```
FAILED tests/doc/tst_doc_images.py::test_static_images[pyvista-examples-examples-load_frog_tissues-1_00_00] - Failed: pyvista-examples-examples-load_frog_tissues-1_00_00 Exceeded image regression error of 500.0 with an image error equal to: 1178.522875817018
FAILED tests/doc/tst_doc_images.py::test_static_images[pyvista-plotting-volume_property-VolumeProperty-1_00_00] - Failed: pyvista-plotting-volume_property-VolumeProperty-1_00_00 Exceeded image regression error of 500.0 with an image error equal to: 16921.179084967716
FAILED tests/doc/tst_doc_images.py::test_static_images[pyvista-plotting-volume_property-VolumeProperty-apply_lookup_table-1_00_00] - Failed: pyvista-plotting-volume_property-VolumeProperty-apply_lookup_table-1_00_00 Exceeded image regression error of 500.0 with an image error equal to: 16921.179084967716
FAILED tests/doc/tst_doc_images.py::test_static_images[sphx_glr_gaussian_smoothing_002] - Failed: sphx_glr_gaussian_smoothing_002 Exceeded image regression error of 500.0 with an image error equal to: 956.4862745098651
FAILED tests/doc/tst_doc_images.py::test_static_images[sphx_glr_gaussian_smoothing_thumb] - Failed: sphx_glr_gaussian_smoothing_thumb Exceeded image regression error of 500.0 with an image error equal to: 829.7189542484141
FAILED tests/doc/tst_doc_images.py::test_static_images[sphx_glr_volume_010] - Failed: sphx_glr_volume_010 Exceeded image regression error of 500.0 with an image error equal to: 7477.0078431375405
```

I'll make sure there are no issues with these before resetting the cache

banesullivan:
Fixed the example in `VolumeProperty` as it was performing a 0-255 remapping workaround. Otherwise all volume rendering plots look correct whith only one image providing significant difference from `examples/02-plot/volume.py`

| `main` | this branch |
|---|---|
| sphx_glr_volume_010 | screenshot (1) |

banesullivan:
The best ways I'm finding to handle NaN values are

1. Simply mask the NaNs yourself to some value outside of the colormapped range:

```py
volume["data"][np.isnan(volume["data"])] = -1
```

2. Use ghost cells to hide cells that have NaN values:

```py
volume["cell_ids"] = np.arange(_voi.n_cells)
_nan = volume.threshold(scalars="data", invert=True)  # this returns an all NaN mesh
nan_cids = _nan["cell_ids"]
ghost_array = np.zeros(np.prod([d - 1 for d in voi.dimensions]), dtype=np.uint8)
ghost_array[nan_cids] = pv._vtk.vtkDataSetAttributes.HIDDENCELL
volume.cell_data[pv._vtk.vtkDataSetAttributes.GhostArrayName()] = ghost_array
```


Considering this, I'm going to opt to not solve this issue in this PR and try to land something later that makes it seamless to mask ghost cells based on the NaN values in either POINT or CELL data fields.

banesullivan:
I actually hadn't noticed those black artifacts... I'll take a look at that. 

But generally I think every thing looks good here and suppose we just need to reset the image cache for the docs

I have a lot more I'm doing with this volume rendering so there's more improvements to come. Hoping we can have a consensus that these changes while not perfect are a big step in the right direction for volume rendering for the next imminent release. With that are you comfortable merging this if I can't figure out these black artifacts when I get back to coding tonight?

user27182:
> I actually hadn't noticed those black artifacts... I'll take a look at that.
> 
> But generally I think every thing looks good here and suppose we just need to reset the image cache for the docs
> 
> I have a lot more I'm doing with this volume rendering so there's more improvements to come. Hoping we can have a consensus that these changes while not perfect are a big step in the right direction for volume rendering for the next imminent release. With that are you comfortable merging this if I can't figure out these black artifacts when I get back to coding tonight?

Black artifacts are fine with me, may be macOS specific. And the label colors are different for some reason but I know there's lots of issues with discontinuous color mapping with `n_labels`... I think these changes are OK.

banesullivan:
I'm seeing these black artifacts on my Linux machine for both this branch and main and I can subtly see them in the documentation example here https://docs.pyvista.org/api/examples/_autosummary/pyvista.examples.examples.load_frog_tissues

banesullivan:
> And the label colors are different for some reason but I know there's lots of issues with discontinuous color mapping with n_labels... I think these changes are OK.

Yeah, this was my main concern. The colors exactly match so this is purely an issue with how we chop up a discontinous colormap with n_colors. IMO, something we should address seperately

pyvista/plotting/lookup_table.py
```diff
@@ -1081,8 +1090,14 @@ def to_opacity_tf(self, clamping: bool = True) -> _vtk.vtkPiecewiseFunction:
...
         opacity_tf.SetClamping(clamping)
-        for ii, value in enumerate(self.values[:, 3]):
-            opacity_tf.AddPoint(ii, value / self.n_values)
+        mn, mx = self.scalar_range
+        for ii, value in enumerate(np.linspace(mn, mx, self.n_values)):
+            alpha = self.values[ii, 3]
+            # vtkPiecewiseFunction expects alphas between 0 and 1
+            # our lookup table is between 0 and 255
+            alpha = alpha / 255
+            alpha = min(alpha, max_clip)
```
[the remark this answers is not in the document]
banesullivan: This is added to the docstring.

user27182:
> > And the label colors are different for some reason but I know there's lots of issues with discontinuous color mapping with n_labels... I think these changes are OK.
> 
> Yeah, this was my main concern. The colors exactly match when plotting isobodies so this is purely an issue with how we chop up a discontinous colormap with n_colors. IMO, something we should address seperately

Relevant issue: https://github.com/pyvista/pyvista/issues/5994
And we have [color_labels](https://dev.pyvista.org/api/core/_autosummary/pyvista.datasetfilters.color_labels#pyvista.DataSetFilters.color_labels) now as a workaround
So yea we can fix this separately.

user27182 (commented):
All changes look good except this change is a bit concerning to the small toe
voxelize

banesullivan:
That test should be unaffected by this. It's coming from https://dev.pyvista.org/examples/01-filter/voxelize.html, right? I don't think that is using any volume rendering

tests/doc/doc_image_cache/sphx_glr_voxelize_004.jpg
banesullivan, line 1: This is coming from `examples/01-filter/voxelize.py` and it is not volume rendering but rather transparent contours. See 

https://github.com/pyvista/pyvista/blob/b14b7d679232a66b2b246a9eae0347047fc804e8/examples/01-filter/voxelize.py#L65-L70

tests/doc/doc_image_cache/sphx_glr_voxelize_003.jpg
banesullivan, line 1: This also is not volume rendering and should be unrelated to the changes in this PR

user27182: Ok. Will need to look into where this change originated. There should not be warnings for any of the images in `doc_image_cache` (warnings were all reset with #7398), but these images emitted in [4fbd169](https://github.com/pyvista/pyvista/pull/7358/commits/4fbd16929c535fdf0f60ab6b02f53f515b0396fc)) all emitted warnings.

user27182: It seems these could be "flaky" artifacts from the build, the warnings appear again in the latest build showing the small toe as it was originally/ as expected

user27182 (commented):
Changes here look good. I'm not familiar with volume rendering on a technical level but the changes appear to be an improvement.

Looks like some test coverage is missing but can probably merge this once that's addressed.

banesullivan:
I'll see what I can do to bump that coverage tonight. It's time we cut a release!

pyvista/plotting/volume_property.py
```diff
@@ -131,23 +171,29 @@ def apply_lookup_table(self, lookup_table: pyvista.LookupTable):
...
         >>> lut.apply_opacity([1.0, 0.0, 0.0, 0.3, 0.0, 0.0, 0.0, 0.3])
-        >>> actor.prop.apply_lookup_table(lut)
         >>> pl.show()
 
         """
         if not isinstance(lookup_table, pyvista.LookupTable):
             msg = '`lookup_table` must be a `pyvista.LookupTable`'  # type: ignore[unreachable]
             raise TypeError(msg)
-
+        if self._lookup_table != lookup_table:
```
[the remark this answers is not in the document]
tkoyama010: Don't need this.

tkoyama010 (approved):
LGTM

pyvista/plotting/volume_property.py
```diff
@@ -131,23 +171,29 @@ def apply_lookup_table(self, lookup_table: pyvista.LookupTable):
...
+            self._lookup_table = lookup_table
         self.SetColor(lookup_table.to_color_tf())
         self.SetScalarOpacity(lookup_table.to_opacity_tf())
 
+    def __del__(self):
+        """Clean up the lookup table observer when the object is deleted."""
+        if self._lookup_table_observer_id is not None and self._lookup_table is not None:
+            self._lookup_table.RemoveObserver(self._lookup_table_observer_id)
+            self._lookup_table_observer_id = None
+            self._lookup_table = None
```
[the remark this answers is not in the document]
banesullivan: This suggestion was invalid. Pushing a fix

banesullivan:
I'm struggling to get the coverage up any higher here... the code that is missing is definitely being hit during the tests which we can see in the failures from the changes copilot made there which broke the tests. Not sure what to do here... maybe I make a test that explicitly deletes the property?

tkoyama010:
I added test in 8078d75d86d760017d5f8ff176d43299d15c28a2 . Let's see if it raise coverage.

pyvista/plotting/volume_property.py
```diff
@@ -131,23 +171,29 @@ def apply_lookup_table(self, lookup_table: pyvista.LookupTable):
...
         if not isinstance(lookup_table, pyvista.LookupTable):
             msg = '`lookup_table` must be a `pyvista.LookupTable`'  # type: ignore[unreachable]
             raise TypeError(msg)
-
+        if self._lookup_table != lookup_table:
+            self._lookup_table = lookup_table
         self.SetColor(lookup_table.to_color_tf())
         self.SetScalarOpacity(lookup_table.to_opacity_tf())
 
+    def __del__(self):
```
tkoyama010, line 190: If that doesn't work, let's temporarily skip coverage.
```suggestion
    def __del__(self):  # pragma: no cover
```

tkoyama010 (approved):

user27182 (approved):

PR #7365, opened 2025-04-01
Prepend pyvista to plot directive configuration options

akaszynski:
### Overview

Prepend `pyvista_` to the configuration options in `pyvista.ext.plot_directive`. The resolves the critical issue where both the `matplotlib.sphinxext.plot_directive` and `pyvista.ext.plot_directive` cannot both be used at the same time.

### Details

See:

- https://github.com/pyvista/pyvista/discussions/5951

#### Backwards compatibility

I'm of the opinion of not making this backwards compatible for the following reasons:
- There's only a handful of (propbably) power users using this.
- It's completely in conflict with `matplotlib`'s extension.
- One more thing to remove at a later date.


akaszynski:
Our tests might be long, but they really do catch so much. Thanks for everyone who continues to maintain these tests.

user27182:
Might need to update these lines if the plot directive directory name is modified
https://github.com/pyvista/pyvista/blob/895475d9255e07e6c3e92e9d647706a349fd5fed/doc/source/make_tables.py#L74
https://github.com/pyvista/pyvista/blob/895475d9255e07e6c3e92e9d647706a349fd5fed/pyvista/core/celltype.py#L41

user27182:
`plot_directive.rst` docs should also be updated
https://dev.pyvista.org/extras/plot_directive.html#module-pyvista.ext.plot_directive
https://github.com/pyvista/pyvista/blob/main/doc/source/extras/plot_directive.rst

akaszynski:
Rather than making a new counter, I've modified this PR to simply reuse the mpl counter if it exists. This helps with tracking figure order. Even if we didn't reuse their counter, it's unlikely that there would be a conflict as we append two additional numbers (num and subnum), but it's best not to rely on `matplotlib`'s behavior.

> `plot_directive.rst` docs should also be updated...

Thanks for pointing that out @user27182. Those docs rely on the autosummary and it will be reflected when the docs (eventually) build.

user27182 (approved):

adtzlr:
Thank you all , this is great ❤️

PR #7370, opened 2025-04-02
Refactor `vtk_verbosity` to use abstract state manager class

user27182:
### Overview

Managing the state of a global variable for general setting/getting but also with a context manager has multiple use-cases. This PR refactors `vtk_verbosity` to make it possible to do this easily and generically. 

An abstract `_StateContextManager` is added to manage the state. Subclasses only need to define a getter and setter for the state variable. E.g. #7198 uses this class to manage the use of vtk's snake_case api.

This PR also simplifies the interface of `vtk_verbosity` to only allow string inputs. The use of integer inputs and `vtkLogger.Verbosity` inputs are no longer allowed. This is done to allow specifying a `Literal` type to define `_StateContextManager` subclasses so that the parent class can auto-validate inputs (using the Literal arguments).


pyvista/core/_vtk_core.py
```diff
@@ -589,30 +596,144 @@ def VTKVersionInfo():
...
-            'max',
-        ]
-        | int
-        | vtkLogger.Verbosity
-    )
 
+T = TypeVar('T')
+
+
+class _StateManager(contextlib.AbstractContextManager[None], Generic[T], ABC):
```
user27182, line 603: This new class and the refactored VTKVerbosity should probably live in its own module in `utilities`. They are kept in `_vtk_core` for now so that the diff is easier to review.

I can make the move as part of this PR though if desired.

akaszynski: > I can make the move as part of this PR though if desired.

Recommend moving it out. Prefer one class per module if possible.

user27182: Do you want it moved as part of this PR or as a follow up? (The diff is not so great sometimes when moving across files)

tkoyama010:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM

Issue #1180, opened 2021-02-15
Segmentation fault: Maximum number of clients reached


hakostra:
**Describe the bug, what's wrong, and what you expect:**

When creating a sequence of visualizations, the memory usage by the Python process keep rising, indicating that there are some resources not being deallocated properly, before it eventually segfault. The number of plots you can make before it segfaults seems to be around the same number each time, just above 180... The complexity of the images does not seem to affect, i.e. I can create the same number of figures before the segfault, independent on the complexity of the figure.

The segfault message is:
```
Maximum number of clients reachedFatal Python error: Segmentation fault

Current thread 0x00007f7b2539d740 (most recent call first):
  File "/home/hakostra/tmp/vtktest/lib/python3.8/site-packages/pyvista/plotting/plotting.py", line 841 in render
  File "/home/hakostra/tmp/vtktest/lib/python3.8/site-packages/pyvista/plotting/plotting.py", line 4266 in show
  File "pyvista-leak.py", line 17 in slice_once
  File "pyvista-leak.py", line 42 in <module>
Segmentation fault (core dumped)
```

-----

**To Reproduce**

Please include a code snippet to reproduce the bug in the block below:

```py
def slice_once(grid, x0, nvec):
    myslice = grid.slice(origin=x0, normal=nvec)

    plotter = pv.Plotter(off_screen=True, window_size=[1920, 1200],
                         lighting='none')
    plotter.add_mesh(myslice)
    plotter.enable_parallel_projection()
    plotter.camera_position = "yz"
    plotter.show(screenshot='demo.png')


print(pv.Report())

rng = np.random.default_rng()
size = (14, 14, 14,)

# Grid 1
x = np.linspace(-0.2, 1.2, 15)
y = np.linspace(-0.2, 1.2, 15)
z = np.linspace(-0.2, 1.2, 15)
X, Y, Z = np.meshgrid(x, y, z, indexing='ij')
grid1 = pv.StructuredGrid(X, Y, Z)

# Associate some sinteresting data
P = rng.random(size, dtype=np.float32)
grid1.cell_arrays["P"] = P.flatten()

N = 250
origin = np.array([0.0, 0.5, 0.5])
nvec = np.array([1.0, 0.0, 0.0])
for i in range(N):
    x0 = origin + (i/(N-1))*nvec
    print(f"Generating slice {i}")
    slice_once(grid1, x0, nvec)

```
The above program manage to generate 183 screenshots before it segfaults on my system.

[...]

PR #7373, opened 2025-04-03
Set default render window for OSMesa with VTK 9.4

user27182:
Configure the `vtkOpenGLRenderWindow`, see https://docs.vtk.org/en/latest/advanced/runtime_settings.html#opengl and https://github.com/pyvista/pyvista/issues/1180#issuecomment-2776917657


user27182:
Build no longer seg faults, now we're getting these build errors (can deal with this in a separate PR)
```
/home/runner/work/pyvista/pyvista/pyvista/plotting/charts.py:docstring of pyvista.plotting.charts.Axis:22: WARNING: py:obj reference target not found: pen [ref.obj]
/home/runner/work/pyvista/pyvista/pyvista/plotting/charts.py:docstring of pyvista.plotting.charts.Axis:25: WARNING: py:obj reference target not found: grid_pen [ref.obj]
```

user27182:
Might be good to also add this info to `pv.Report` for debugging:
``` python
f'OpenGL backend: {vtk.vtkRenderWindow().GetClassName()}'
```

user27182:
> Might be good to also add this info to `pv.Report` for debugging:
> 
> ```python
> f'OpenGL backend: {vtk.vtkRenderWindow().GetClassName()}'
> ```

We should also consider generating a warning and means to set this variable easily, see https://github.com/pyvista/pyvista/issues/1180#issuecomment-2779492631. Both of these changes can be done in a follow-up PR. 
For the immediate purpose of getting #6879 merged, I think this PR is fine as-is

banesullivan (changes requested):
I don't like having this value set as an export in the `Makefile` like this as it would make any developer running on linux locally use OSMesa instead of GPU-based render. For example:

```bash
$ python -c "import pyvista;print(pyvista.GPUInfo())"                      

      GPU Renderer : NVIDIA GeForce RTX 3080 Ti/PCIe/SSE2
       GPU Version : 4.6.0 NVIDIA 550.120
```

```bash
$ VTK_DEFAULT_OPENGL_WINDOW=vtkOSOpenGLRenderWindow python -c "import pyvista;print(pyvista.GPUInfo())"

        GPU Vendor : Mesa
      GPU Renderer : llvmpipe (LLVM 19.1.1, 256 bits)
       GPU Version : 4.5 (Core Profile) Mesa 24.2.8-1ubuntu1~24.04.1

```

Instead, I think this ought to be set specifically on the docs gtihub action

banesullivan:
Does this need to be set on all of our CI jobs that plot on linux actually?

user27182:
> Does this need to be set on all of our CI jobs that plot on linux actually?

Good question. I think technically no in the sense that all jobs are passing. But maybe we should? I don't think the docstring check workflow installs the osmesa packages like this workflow does, so maybe that's creating an issue somehow?

TBH I don't know much about any of the linux build stuff for off-screen rendering. Was more just focused on resolving issue 2 from https://github.com/pyvista/pyvista/pull/6879#issuecomment-2773862865. This issue doesn't seem to occur with the other workflows so didn't bother looking into it.

banesullivan:
Okay, let's not worry about it since builds are passing but I'll add this to my own list to come back and resolve any of these differences later

banesullivan (approved):

PR #7374, opened 2025-04-04
Fix `charts.Axis` pen attribute documentation for VTK 9.4

user27182:
Fix https://github.com/pyvista/pyvista/pull/7373#issuecomment-2777228428


user27182:
This fix was previously patched as part of #7376 in order for that PR's images to be generated in the first place. So approving that PR in some sense implicitly approves this one.

I think these changes are low-risk and can be further reviewed anyway before merging #6879 into main, and so I'm going to merge this into #6879 to move it forward.

PR #7377, opened 2025-04-04
Fix using cell vectors with `DataSet.arrows`

user27182:
### Overview

`arrows` currently assumes point vectors are used. This breaks when trying to use cell vectors. This PR fixes this.


MatthewFlamm (commented):
This LGTM although there is some ruff error related to quoting the cast type, which I personally find annoying in this case.  I mean the suggested pattern by ruff is annoying to me.

PR #7379, opened 2025-04-05
Fix default cube normals

user27182:
### Overview

Default point normals are not correct (see https://github.com/pyvista/pyvista/pull/6879#issuecomment-2779658209). This PR fixes this.


tkoyama010 (approved):
LGTM

PR #7380, opened 2025-04-05
Add __dict__ method to Theme class and include a test

tkoyama010:
### Overview

Add `__dict__` method to Theme class and include a test

### Details

- None


tkoyama010:
I realized that modifying `__dict__` is dangerous.

Issue #7378, opened 2025-04-05
Docs examples are guaranteed to be compatible with minimum supported version of Python


banesullivan:
### Describe what maintenance you would like added.

Looks like our docs build isn't compatible with Python 3.11

```
/home/bane/Software/pyvista/pyvista/pyvista/core/filters/data_object.py:docstring of pyvista.core.filters.data_object.DataObjectFilters.cell_quality:72: WARNING: Exception occurred in plotting pyvista-DataObjectFilters-cell_quality-1
 from /home/bane/Software/pyvista/pyvista/doc/source/api/core/_autosummary/pyvista.DataObjectFilters.cell_quality.rst:
Traceback (most recent call last):
  File "/home/bane/Software/pyvista/pyvista/pyvista/ext/plot_directive.py", line 400, in _run_code
    exec(code, ns)
  File "<string>", line 14
    print(f'[{",\n ".join(repr(measure) for measure in valid_measures)}]')
                                                                         ^
SyntaxError: f-string expression part cannot include a backslash [docutils]
```

What's out general thoughts on this? CI is running the docs build on 3.12 but should we make sure all of our doc examples are compatible for the minimum supported Python version

### Links to source code.

https://github.com/pyvista/pyvista/blob/a377dcafc156609c94723ebce2afa5b30d0f08cf/pyvista/core/filters/data_object.py#L2848

cc @user27182 since you've been leverage some of the 3.12 f-string syntax more and more lately

banesullivan:
I feel that we ought to change the docs CI to run on the minimum version of Python we support that can run the latest version of VTK we support

user27182:
> I feel that we ought to change the docs CI to run on the minimum version of Python we support that can run the latest version of VTK we support

+1

Is there a way to lint this with pre-commit? 
If we can't, then a concern I have is that this will increase dev time, e.g. needing to wait 3hrs for the docs to build only to find out you need to use `Union[float, str]` instead of `float | str`. I would prefer to avoid this situation if possible.

user27182:
Though I guess maybe we can also run the docstring check with the older Python version? This should catch any Python syntax errors in the docstrings fairly quickly (much sooner than the docs build)

And the sphinx gallery examples are the first to run in the build.... so these should also be caught fairly early in the build

So, may not be too bad after all in case we can't lint it.

MatthewFlamm:
[...]

PR #7381, opened 2025-04-06
Make `cell_quality` docstring compatible with older python

user27182:
### Overview

Fix syntax error from https://github.com/pyvista/pyvista/issues/7378#issue-2973750465

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -2845,7 +2845,8 @@ class DataObjectFilters:
 
         >>> cqual = sphere.cell_quality('all_valid')
         >>> valid_measures = cqual.cell_data.keys()
-        >>> print(f'[{",\n ".join(repr(measure) for measure in valid_measures)}]')
+        >>> joined = ',\n '.join(repr(measure) for measure in valid_measures)
+        >>> print(f'[{joined}]')
         ['area',
          'aspect_frobenius',
          'aspect_ratio',


tkoyama010 (approved):
LGTM

pyvista/core/filters/data_object.py
```diff
@@ -2845,7 +2845,8 @@ def cell_quality(  # type: ignore[misc]
 
         >>> cqual = sphere.cell_quality('all_valid')
         >>> valid_measures = cqual.cell_data.keys()
-        >>> print(f'[{",\n ".join(repr(measure) for measure in valid_measures)}]')
+        >>> joined = ',\n '.join(repr(measure) for measure in valid_measures)
+        >>> print(f'[{joined}]')
```
MatthewFlamm, line 2849: Cannot do a suggestion now that it is merged, but we do we even need a print statement with an f string here.  Can we just do:

```py
        >>> ',\n '.join(repr(measure) for measure in valid_measures)
```

user27182: yes that would work, could probably omit the space after `\n` since we also no longer have the brackets

This formatting all stems from differences in CI vs `IPython`, since typing `valid_measures` + ENTER locally for me automatically formats the output as

``` python
['area',
 'aspect_frobenius',
 'aspect_ratio',
 'condition',
 'distortion',
 'max_angle',
 'min_angle',
 'radius_ratio',
 'relative_size_squared',
 'scaled_jacobian',
 'shape',
 'shape_and_size']
```

But when I tried that, the docstring CI failed since it does not apply this formatting and prints it flat instead:

``` python
['area', 'aspect_frobenius', 'aspect_ratio', 'condition', 'distortion', 'max_angle', 'min_angle', 'radius_ratio', 'relative_size_squared', 'scaled_jacobian', 'shape', 'shape_and_size']
```

Hence why a custom f-string was used in the first place.
If we can configure the doctest to apply this kind of formatting automatically that would probably be best

MatthewFlamm: You could try `# doctest: +NORMALIZE_WHITESPACE`

user27182: > You could try `# doctest: +NORMALIZE_WHITESPACE`

This seems to be exactly what is needed here, thanks. Done in #7385

PR #7386, opened 2025-04-08
Add tests to limit sphinx gallery example execution times

user27182:
### Overview

Sphinx gallery generates an xml file with computation times, see https://sphinx-gallery.github.io/stable/configuration.html#junit-xml

This PR parses this xml file and generates pytest cases. The tests fail if they exceed the specified maximum allowed time.


user27182:
new tests are working, see https://github.com/pyvista/pyvista/actions/runs/14337838724/job/40189412437?pr=7386#step:14:328

``` python

FAILED tests/doc/tst_doc_images.py::test_gallery_example_execution_time[backface_props] - AssertionError: Sphinx gallery example 'Setting Backface Properties' from file
  	../../examples/02-plot/backface_props.py
  has an execution time: 385.58251428604126 seconds
  which exceeds the maximum allowed: 60.0 seconds.
assert 385.58251428604126 < 60.0
FAILED tests/doc/tst_doc_images.py::test_gallery_example_execution_time[depth_of_field] - AssertionError: Sphinx gallery example 'Depth of Field Plotting' from file
  	../../examples/02-plot/depth_of_field.py
  has an execution time: 792.1083476543427 seconds
  which exceeds the maximum allowed: 60.0 seconds.
assert 792.1083476543427 < 60.0
FAILED tests/doc/tst_doc_images.py::test_gallery_example_execution_time[edl] - AssertionError: Sphinx gallery example 'Eye Dome Lighting' from file
  	../../examples/02-plot/edl.py
  has an execution time: 70.12101650238037 seconds
  which exceeds the maximum allowed: 60.0 seconds.
assert 70.12101650238037 < 60.0
FAILED tests/doc/tst_doc_images.py::test_gallery_example_execution_time[pbr] - AssertionError: Sphinx gallery example 'Physically Based Rendering' from file
  	../../examples/02-plot/pbr.py
  has an execution time: 765.3375141620636 seconds
  which exceeds the maximum allowed: 60.0 seconds.
assert 765.3375141620636 < 60.0
FAILED tests/doc/tst_doc_images.py::test_gallery_example_execution_time[planets] - AssertionError: Sphinx gallery example '3D Earth and Celestial Bodies' from file
  	../../examples/99-advanced/planets.py
  has an execution time: 620.1809523105621 seconds
  which exceeds the maximum allowed: 60.0 seconds.
assert 620.1809523105621 < 60.0
```

Thinking of setting the limit to `900` seconds (15 minutes) for now, though we can probably make it `360` (6 minutes) once #7357 lands.

tests/doc/tst_doc_images.py
```diff
@@ -268,3 +274,46 @@ def test_interactive_plot_file_size(vtksz_file: str):
...
+        )
+    return cases
+
+
+TEST_CASES = load_test_cases_from_xml(HTML_DIR / SPHINX_GALLERY_CONF_JUNIT)
+IDS = [case['classname'] for case in TEST_CASES]
+
+
+@pytest.mark.parametrize('testcase', TEST_CASES, ids=IDS)
+def test_gallery_example_execution_time(testcase):
```
user27182, line 309: This test doesn't quite match the module name `tst_doc_images`. But`pytest tests/doc/tst_doc_images.py` is hard-coded into the CI workflow, so it's easier to lump the test into this module.  We can consider renaming this module or creating a second test module though if that's preferred

MatthewFlamm: You could consider renaming this to `tst_doc.py` to reflect the increased scope and update any hardcoded values.  I assume it is named like this to avoid autocollection for `pytest`.

tests/doc/tst_doc.py
```diff
@@ -25,6 +27,12 @@
 
 MAX_VTKSZ_FILE_SIZE_MB = 50
 
+# Same value as `sphinx_gallery_conf['junit']` in `conf.py`
+SPHINX_GALLERY_CONF_JUNIT = Path('sphinx-gallery') / 'junit-results.xml'
+SPHINX_GALLERY_EXAMPLE_MAX_TIME = 900.0  # Measured in seconds
```
user27182, line 32: This needs to be set to a reasonable value. From #7397 , the slowest example takes about 2 minutes:
https://67f84843a3147b162189bc78--pyvista-dev.netlify.app/sg_execution_times.html

But I think this is too long. Should we aim to make this 30 seconds? (Plus fudge-factor so maybe warning after 30, fail after 45 seconds?). We can allow exceptions for the few examples that currently exceed this, but put a hard cap on new ones

MatthewFlamm: I like the idea of enforcing a default time that is quite fast and then a list of slow execution examples that is opt-in via list.  I don't think we should exempt them completely, otherwise we may get regressions that creep into those examples making them take longer.  This way new examples are by default checked to be fast.  If they aren't we can assess whether it should be trimmed down for execution speed or added to the list.

user27182: > I like the idea of enforcing a default time that is quite fast and then a list of slow execution examples that is opt-in via list. 

If we define "quite fast" to be 10 seconds, I think we can shave about 20 minutes from the build time by skipping any examples that take longer than this to run.

> I don't think we should exempt them completely, otherwise we may get regressions that creep into those examples making them take longer. 

We can maybe add a new CI env var and corresponding GitHub label `SKIP_SLOW` to opt-in to skip these? We can couple this with #7392, where the package `pytest-fail-slow` can be used to mark tests as slow (and optionally skip them). And maybe mirror how the cache labels works, and have `skip-slow-doctest`, `skip-slow-unittest`, `skip-slow-gallery`, etc. labels for controlling this.

> This way new examples are by default checked to be fast. If they aren't we can assess whether it should be trimmed down for execution speed or added to the list.

I agree, slow examples should first fail and require adjustments of some kind to make them pass.

user27182: The suggestions above are good and I think more needs to be done to make this all work as we'd like it to. But for this PR, is it OK if we at least enable the tests as-is for now, and simply place a hard limit of maybe 150 seconds (30 seconds slower than the slowest example) ?

MatthewFlamm: Coming back to this, it seems fine to me to get this in to avoid future problems and then refine later.  Will do another review.

MatthewFlamm (approved):
LGTM.

IMO it would be better to have the functionality to limit ~test~ gallery example execution time with fine grained control (exclusions or time overrides for specific examples) in sphinx-gallery directly.

user27182:
> LGTM.
> 
> 
> 
> IMO it would be better to have the functionality to limit ~test~ gallery example execution time with fine grained control (exclusions or time overrides for specific examples) in sphinx-gallery directly.

That's a good idea. Would also be nice if it worked with [abort_build_on_first_fail](https://sphinx-gallery.github.io/stable/configuration.html#abort-build-on-first-fail) such that the build fails early instead of having to wait until after the build completes to find that an example took too long.

banesullivan (approved):

PR #7384, opened 2025-04-07
Bump `ruff` to 11.4

user27182:
### Overview

Split from https://github.com/pyvista/pyvista/pull/7382

Added a few ignores and changes to TYPE_CHECKING blocks

PR #7390, opened 2025-04-08
Fix path handling  to use pathlib instead of os

tkoyama010:
### Overview

Fix path handling  to use pathlib instead of os.

### Details

- None


pyvista/examples/_dataset_loader.py
```diff
@@ -395,8 +395,8 @@ def load(self):
                 # Re-define read function to read all files in a directory as a multiblock
                 read_func = lambda path: _load_as_multiblock(  # type: ignore[assignment]
                     [
-                        _SingleFileDatasetLoader(str(Path(path, fname)))
-                        for fname in sorted(os.listdir(path))  # noqa: PTH208
+                        _SingleFileDatasetLoader(str(Path(path) / fname))
+                        for fname in sorted(Path(path).iterdir())
```
user27182, line 399: I considered some of these changes (suggested by ruff) as part of #7384, but found that these two are not directly interchangeable. `os.listdir` returns a list of filenames only (as strings), whereas `Path.iterdir()` returns a generator of full paths (as Path objects).

so a 1:1 replacement for `os.listdir` is something like `[p.name for p in path.iterdir()]`. Due to these significant differences in output, this rule must be applied very carefully. For the particular change here, `fname` is now a full Path object and not just a string filename, and is very likely to break this code.

tkoyama010:
```suggestion
                        _SingleFileDatasetLoader(str(Path(path, fname)))
                        for fname in sorted(os.listdir(path))  # noqa: PTH208
```

tkoyama010:
@pyvista-bot LGTM

banesullivan:
This broke handling of `_server_proxy_prefix` on the theme

user27182:
> This broke handling of `_server_proxy_prefix` on the theme

This seems to be the second time this happens... https://github.com/pyvista/pyvista/pull/5818#discussion_r1563752955

is there a test we should add?

banesullivan:
Yeah probably, I'll add to my todo list to test this (since I seem to be the only developer affected)

PR #7394, opened 2025-04-10
Save images from test warnings for doc image cache

user27182:
### Overview

Some changes to the repo can have a substantial impact on the documentation images, e.g. 30 or more images with test errors or warnings.

In such cases, it may be preferred to update the image cache's failed images _and_ images with warnings. But currently, the tests and artifacts only save/upload the failures from errors. This PR changes this and also uploads images with warnings.


tkoyama010 (approved):
LGTM

tests/doc/tst_doc_images.py
```diff
@@ -134,14 +135,16 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize('vtksz_file', files, ids=ids)
 
 
-def _save_failed_test_image(source_path):
+def _save_failed_test_image(source_path, parent_dir: Literal['warnings', 'errors', 'flaky']):
```
tkoyama010, line 138:
```suggestion
def _save_failed_test_image(source_path, category: Literal['warnings', 'errors', 'flaky']):
```

tests/doc/tst_doc_images.py
```diff
@@ -134,14 +135,16 @@ def pytest_generate_tests(metafunc):
...
     """Save test image from cache or build to the failed image dir."""
+    parent_dir = Path(parent_dir)
```
tkoyama010, line 140:
```suggestion
    parent_dir = Path(category)
```

tkoyama010 (changes requested):

user27182:
This is mostly working (I was able to generate warning images for #7398), but need to fix saving to flaky dir

user27182:
> This is mostly working (I was able to generate warning images for #7398), but need to fix saving to flaky dir

I double checked this locally, and indeed everything is working as expected, and a flaky dir is being properly created. But the issue is that this workflow only uploads the artifact when there are failures

https://github.com/pyvista/pyvista/blob/98b6e8117a8c48cae2582b138c6a3e6ab5a14256/.github/workflows/docs.yml#L112-L113

Doing the hasfiles check was highlighted as an issue in https://github.com/pyvista/pyvista/pull/6204#discussion_r1626451025, and can be fixed in a separate PR I think. For now, the main thing is that flaky images are separated from warnings/errors when there are many failures to review and update

tkoyama010 (approved):
LGTM

PR #7397, opened 2025-04-10
Speed up rendering of planets example

user27182:
### Overview

The planets example takes 4 minutes to render, even after #7357 was implemented.

This PR lowers the sampling resolution even more to speed it up.


MatthewFlamm (approved):
If anything the still images look even better than the originals.

PR #7398, opened 2025-04-10
Fix theme settings applied to document build

### LINKED ISSUE #7389: Make documentation theme available as a theme

The PyVista documentation theme might be useful as a starting point for downstream projects to use.  Even if just used as a reference for useful settings, particularly `resample_environment_texture` from #7357.

Currently the theme is set in `doc/source/conf.py`, but we could expose it as a direct theme in `pyvista` and include in the documentation itself.

https://github.com/pyvista/pyvista/blob/26909177f14bc262bc7f5488295e9ad09b353ea2/doc/source/conf.py#L43-L48

user27182:
### Overview

Fix #7389


MatthewFlamm (approved):
I went through many, but not all of the image diffs. In 90+% of the cases the updated font size is a positive.  Of the remaining fraction where the updated font size was not good, e.g. tick labels clashing togeher, 80-90% of the original images were not ideal either.  I'm inclined to approve/merge this one as is, as the intention was to have these font sizes as the default.  If there are egregious font issues, that could be a separate PR to clean up.

PR #7407, opened 2025-04-11
Make `bool` parameters of `Transform.apply_to` methods keyword-only

user27182:
### Overview

Bool args should be keyword-only. This is not a breaking change since this is not yet part of a release.


tkoyama010 (approved):
LGTM

PR #7420, opened 2025-04-12
Add __str__ method to GPUInfo and update related tests

tkoyama010:
### Overview

- Implemented __str__ method in GPUInfo class to provide a string representation of GPU information.
- Updated test_gpuinfo to include tests for the __str__ method.

### Details

- None

### DIFF
--- a/pyvista/report.py
+++ b/pyvista/report.py
@@ -98,6 +98,10 @@ class GPUInfo:
         content += '\n'
         return content
 
+    def __str__(self):
+        """Return string representation of GPUInfo."""
+        return f'GPUInfo(vendor={self.vendor}, renderer={self.renderer}, version={self.version})'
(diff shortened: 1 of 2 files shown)


pyvista/report.py
```diff
@@ -98,6 +98,10 @@ def __repr__(self):
         content += '\n'
         return content
 
+    def __str__(self):
+        """Return string representation of GPUInfo."""
+        return f'GPUInfo(vendor={self.vendor}, renderer={self.renderer}, version={self.version})'
```
banesullivan, line 103: In my mind, these sorts of representations (e.g., `Class(a=1, b=2)`) infer that the object can be instantiated but we never want anyone to populate these properties this themselves

Curious what the motivation here is?

Issue #7422, opened 2025-04-13
Add type hints for function arguments and returns to plotter.py


maxdswain:
### Describe the feature you would like to be added.

As a user of both pyvista and a type checker, it would be useful to have type hints for [plotter.py](https://github.com/pyvista/pyvista/blob/main/pyvista/plotting/plotter.py) as this is commonly used interface. I'd be happy to implement this myself.

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

maxdswain:
I haven't done any inlay type hints, *args or **kwargs so it didn't end being that bad to type hint a

user27182:
Ok thanks that's good these can be handled separately then

user27182:
Completed by #7423

PR #7423, opened 2025-04-13
Add type hints to plotter API

maxdswain:
### Overview

Add function argument and function return type hints to `plotter.py` and some related functions.

This fully addressed #7422.


user27182:
I'll have a closer look later, but we have the `ANN` rule from `ruff` currently disabled for all plotting modules. Can you also modify `pyproject.toml` to enable this rule for `plotter.py`? If not I can make a separate PR to enable this. Would require following what's currently done with the `core` module for that rule.

user27182:
I see there are a lot of `mypy` errors... some may be important to resolve but many of these may be OK to simply ignore. Same goes for the `ruff` `ANN` errors, e.g. feel free to ignore some error groups completely (e.g. ANN202).

We use typing both for communicating input/return types for the API, but also to help prevent type-related bugs related to the implementation. I think this PR aims to do the former, so I think it's OK if internal type issues are not addressed directly.

maxdswain:
@user27182 mypy, pre-commit checks and pytest tests are all now passing (at least locally). I _think_ everything should be fixed.

maxdswain:
I think that should have fixed the unit tests and some of the sphinx warnings? I'm not sure about the sphinx stuff and don't have much experience with sphinx so I may need a pointer.

maxdswain:
Should be done!

pyvista/plotting/plotter.py
```diff
@@ -3511,6 +3559,12 @@ def add_mesh(
             if algo is not None:
                 msg = 'Algorithms with `MultiBlock` output type are not supported by `add_mesh` at this time.'
                 raise TypeError(msg)
+            if not isinstance(opacity, (float, int, type(None))):
+                msg = f'`opacity` must be a float, int or None type, not ({type(opacity)})'
+                raise TypeError(msg)
```
user27182, line 3564:
```suggestion
            _validation.check_instance(opacity, (float, int, type(None)), name='opacity')
```

pyvista/plotting/plotter.py
```diff
@@ -3511,6 +3559,12 @@ def add_mesh(
...
+            if not isinstance(scalars, (str, type(None))):
+                msg = f'`scalars` must be a string or None type for MultiBlock mesh, not ({type(scalars)})'
+                raise TypeError(msg)
```
user27182, line 3567:
```suggestion
            _validation.check_instance(opacity, (str, type(None)), name='scalars')
```

pyvista/plotting/plotter.py
```diff
@@ -3564,6 +3618,7 @@ def add_mesh(
             # active, it doesn't modify the original input mesh.
             # We ignore `copy_mesh` if the input is an algorithm
             mesh = mesh.copy(deep=False)
+        assert isinstance(mesh, (pyvista.DataSet, pyvista.PointSet))
```
user27182, line 3621: We avoid using `assert` statements in code where users may encounter this error.

If this is needed to satisfy `mypy`, use `cast` instead for type-casting only, (or modify the logic above to ensure `mypy` can "see" that the type must be `DataSet` at this point).

If this is an important runtime check for users, use `raise TypeError` or similar, or use `_validation.check_instance`.

Also, `PointSet` is a `DataSet`, so `DataSet` alone should be sufficient here.

pyvista/plotting/plotter.py
```diff
@@ -3622,6 +3677,7 @@ def add_mesh(
                 silhouette_actor = self.add_silhouette(algo or mesh)
             silhouette_actor.user_matrix = user_matrix
 
+        assert isinstance(scalar_bar_args, dict)
```
user27182, line 3680: similar to above, `assert` should be avoided

pyvista/plotting/plotter.py
```diff
@@ -3686,10 +3742,12 @@ def add_mesh(
 
         if rgb:
             show_scalar_bar = False
+            assert isinstance(scalars, np.ndarray)
             if scalars.ndim != 2 or scalars.shape[1] < 3 or scalars.shape[1] > 4:
                 msg = 'RGB array must be n_points/n_cells by 3/4 in shape.'
                 raise ValueError(msg)
 
+        assert isinstance(mesh, (pyvista.DataSet, pyvista.PointSet, pyvista.PolyData))
```
user27182, line 3750: similar to above, `assert` should be avoided

pyvista/plotting/plotter.py
```diff
@@ -3895,36 +3959,56 @@ def _add_legend_label(self, actor, label, scalars, color: Color):
...
+        ]
+        | NumpyArray[float] = 'linear',
+        n_colors: int = 256,
+        cmap: str | list[str] | pyvista.LookupTable | None = None,
+        flip_scalars: bool = False,
+        reset_camera: bool | None = None,
+        name: str | None = None,
+        ambient: float | None = None,
+        categories: bool | int = False,
+        culling: Literal['front', 'back'] | bool = False,
```
user27182, line 3993: `frontface` and `backface` are also supported.
```suggestion
        culling: Literal['front', 'back', 'frontface', 'backface'] | bool = False,
```

Consider creating an alias since this option is re-used, e.g. with `add_volume`

```suggestion
        culling: `CullingOptions` | bool = False,
```

pyvista/plotting/plotter.py
```diff
@@ -4218,17 +4302,17 @@ def add_volume(
...
         if show_scalar_bar is None:
-            show_scalar_bar = self._theme.show_scalar_bar or scalar_bar_args
+            show_scalar_bar = self._theme.show_scalar_bar or scalar_bar_args  # type: ignore[assignment]
 
         # Avoid mutating input
         scalar_bar_args = {} if scalar_bar_args is None else scalar_bar_args.copy()
 
         if culling is True:
-            culling = 'backface'
+            culling = 'backface'  # type: ignore[assignment]
```
user27182, line 4311: Ignore can probably be removed once `backface` is added to the hinted `Literal`
```suggestion
            culling = 'backface'
```

pyvista/plotting/plotter.py
```diff
@@ -4243,19 +4327,20 @@ def add_volume(
                 elif len(resolution) != 3:
                     msg = 'Invalid resolution dimensions.'
                     raise ValueError(msg)
-                volume.spacing = resolution
+                volume.spacing = resolution  # type: ignore[assignment]
             else:
                 # TODO: implement `is_pyvista_dataset` with `typing_extensions.TypeIs`
                 #   to remove the need to cast the type here
                 volume = cast('Union[pyvista.MultiBlock, pyvista.DataSet]', wrap(volume))
```
user27182, line 4334: A `TypeGuard` is defined by this PR for `is_pyvista_dataset`. Do we know why this cast is still needed? Can we use `TypeIs` instead of `TypeGuard`?

pyvista/plotting/plotter.py
```diff
@@ -3895,36 +3959,56 @@ def _add_legend_label(self, actor, label, scalars, color: Color):
...
+        shade: bool = False,
+        diffuse: float = 0.7,  # TODO: different default for volumes
+        specular: float = 0.2,  # TODO: different default for volumes
+        specular_power: float = 10.0,  # TODO: different default for volumes
+        render: bool | None = True,
+        user_matrix: NumpyArray[float] | _vtk.vtkMatrix4x4 | None = None,
+        log_scale: bool = False,
         **kwargs,
-    ):
+    ) -> Actor | _vtk.vtkActor | list[Actor]:
```
user27182, line 4011: Similar to `add_mesh`, why is `_vtk.vtkActor` needed here?

pyvista/plotting/plotter.py
```diff
@@ -2975,62 +3016,69 @@ def add_composite(
...
+        color: ColorLike | None = None,
+        style: Literal['surface', 'wireframe', 'points', 'points_gaussian'] | None = None,
+        scalars: str | pyvista.pyvista_ndarray | None = None,
+        clim: Sequence[float] | None = None,
+        show_edges: bool | None = None,
+        edge_color: ColorLike | None = None,
+        point_size: float | None = None,
+        line_width: float | None = None,
+        opacity: float
+        | Literal['linear', 'linear_r', 'geom', 'geom_r']
```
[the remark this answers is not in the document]
user27182: EDIT: I see that the docs are probably out of date for `add_mesh`. `add_volume` includes these options in the docstring. I think the `Literal` hint here should be the same for both `add_mesh` and `add_volume`.

pyvista/plotting/plotter.py
```diff
@@ -4567,13 +4658,15 @@ def add_silhouette(
...
-        prop.SetColor(Color(color).float_rgb)
+        prop.SetColor(Color(color).float_rgb)  # type: ignore[call-overload]
         prop.SetOpacity(opacity)
         prop.SetLineWidth(line_width)
 
         return actor
 
-    def update_scalar_bar_range(self, clim, name=None) -> None:
+    def update_scalar_bar_range(
+        self, clim: float | tuple[float, float], name: str | None = None
```
user27182, line 4668: Is there a reason why sequence can't be used here?
```suggestion
        self, clim: float | Sequence[float, float], name: str | None = None
```

Note that we can also use `_validation.validate_data_range(clim)` to do all the validation work of checking `clim[0] < clim[1]`, checking `len(clim)==2`, and returning a `tuple[float, float]`

If we do that, then we can hint `VectorLike[float]`:

```suggestion
        self, clim: float | VectorLike[float], name: str | None = None
```

pyvista/plotting/plotter.py
```diff
@@ -5433,35 +5552,35 @@ def remove_scalar_bar(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT
...
-        tolerance=0.001,
-        reset_camera=None,
-        always_visible=False,
-        render=True,
-        justification_horizontal=None,
-        justification_vertical=None,
-        background_color=None,
-        background_opacity=None,
-    ):
+        points: Sequence[float] | NumpyArray[float] | pyvista.DataSet | _vtk.vtkAlgorithm,
```
user27182, line 5555: Can we use `MatrixLike[float]` here?
```suggestion
        points: MatrixLike[float] | pyvista.DataSet | _vtk.vtkAlgorithm,
```

pyvista/plotting/plotter.py
```diff
@@ -5433,35 +5552,35 @@ def remove_scalar_bar(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT
...
+        labels: list[str | int] | str,
+        italic: bool = False,
+        bold: bool = True,
+        font_size: float | None = None,
+        text_color: ColorLike | None = None,
+        font_family: Literal['courier', 'times', 'arial'] | None = None,
```
user27182, line 5561: This `Literal` is repeated for `add_text`, can we use an alias instead so both use the same type hint, e.g. `FontFamilyOptions  = Literal['courier', 'times', 'arial']`
```suggestion
        font_family: FontFamilyOptions | None = None,
```

pyvista/plotting/plotter.py
```diff
@@ -5433,35 +5552,35 @@ def remove_scalar_bar(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT
...
+        margin: int = 3,
+        shape_opacity: float = 1.0,
+        pickable: bool = False,
+        render_points_as_spheres: bool = False,
+        tolerance: float = 0.001,
+        reset_camera: bool | None = None,
+        always_visible: bool = False,
+        render: bool = True,
+        justification_horizontal: Literal['left', 'center', 'right'] | None = None,
+        justification_vertical: Literal['bottom', 'center', 'top'] | None = None,
```
user27182, line 5580: These stem from `TextProperty`... I wonder if we should define those options there instead, and import the `Literal` from there instead?

pyvista/plotting/themes.py
```diff
@@ -1838,7 +1839,9 @@ def __init__(self):
         # Grab system flag for auto-closing
         self._auto_close = os.environ.get('PYVISTA_AUTO_CLOSE', '').lower() != 'false'
 
-        self._jupyter_backend = os.environ.get('PYVISTA_JUPYTER_BACKEND', 'trame')
+        self._jupyter_backend: Literal['static', 'client', 'server', 'trame', 'html', 'none'] = (
```
user27182, line 1842: This `Literal` is repeated, can we create an alias for this and re-use it instead?

tests/plotting/jupyter/test_static.py
```diff
@@ -37,7 +37,7 @@ def test_validate_jupyter_backend_raises(mocker: MockerFixture):
 @pytest.mark.parametrize('backend', [None, 'none'])
 def test_set_jupyter_backend_none(backend):
     pv.set_jupyter_backend(backend)
-    assert pv.global_theme.jupyter_backend is None
+    assert pv.global_theme.jupyter_backend == 'none'
```
user27182, line 40: This change seems odd, why is `None` no longer returned here?

user27182 (commented):
Thanks for adding these! Here's an initial review. Feel free to reject or push back on some of them.

The changes mostly look good, a key concern is re-using the same `Literal` values multiple times, I think these should be defined exactly once and re-used multiple times as needed. Otherwise, the type hints can easily fall out of sync when they should be the same.

Also, should all type aliases have a suffix? E.g. use an `Options`, `Literal`, or `Type` suffix for type hints to make them more easily distinguishable from pure class types

pyvista/plotting/plotter.py
maxdswain: Sequence only takes 1 type argument so that cannot be done. With regards to VectorLike[float], I think that we would like the user to know the number of elements inputted via the type checker and this doesn't allow for that.

pyvista/plotting/plotter.py
maxdswain: I think using `TypeIs` as you suggest is best.

tests/plotting/jupyter/test_static.py
maxdswain: I did this because on lines 7125-7129 in `plotter.py`, the None case is handled so if a None is then again returned from the setter, there will be an error as None does not support `.lower()`. To fix this I modified the _validate_jupyter_backend function in `pyvista/jupyter/__init__.py` to only return strings and change this test accordingly.

maxdswain:
Everything should be addressed in 4f72feffbf38a9d53ede25cdc10d27b9b2830bc2.

pyvista/plotting/plotter.py
```diff
@@ -2981,60 +3028,60 @@ def add_composite(
...
+        above_color: ColorLike | None = None,
+        annotations: dict[float, str] | None = None,
+        pickable: bool = True,
+        preference: Literal['point', 'cell'] = 'point',
+        log_scale: bool = False,
+        pbr: bool | None = None,
+        metallic: float | None = None,
+        roughness: float | None = None,
+        render: bool = True,
+        user_matrix: NumpyArray[float] | _vtk.vtkMatrix4x4 | None = None,
```
user27182, line 3078:
```suggestion
        user_matrix: TransformLike | None = None,
```

pyvista/plotting/plotter.py
```diff
@@ -2981,60 +3028,60 @@ def add_composite(
...
+        render_lines_as_tubes: bool | None = None,
+        smooth_shading: bool | None = None,
+        split_sharp_edges: bool | None = None,
+        ambient: float | None = None,
+        diffuse: float | None = None,
+        specular: float | None = None,
+        specular_power: float | None = None,
+        nan_color: ColorLike | None = None,
+        nan_opacity: float = 1.0,
+        culling: Literal['front', 'back'] | None = None,
```
user27182, line 3063:
```suggestion
        culling: CullingOptions | None = None,
```

pyvista/plotting/plotter.py
```diff
@@ -2981,60 +3028,60 @@ def add_composite(
...
+        rgb: bool | None = None,
+        categories: bool = False,
+        silhouette: Silhouette | bool | None = None,
+        use_transparency: bool = False,
+        below_color: ColorLike | None = None,
+        above_color: ColorLike | None = None,
+        annotations: dict[float, str] | None = None,
+        pickable: bool = True,
+        preference: Literal['point', 'cell'] = 'point',
```
user27182, line 3072: import from `core.utilities.arrays`
```suggestion
        preference: PointLiteral | CellLiteral = 'point',
```

pyvista/plotting/plotter.py
```diff
@@ -3904,36 +3963,36 @@ def _add_legend_label(self, actor, label, scalars, color: Color):
...
+        opacity_unit_distance: float | None = None,
+        shade: bool = False,
+        diffuse: float = 0.7,  # TODO: different default for volumes
+        specular: float = 0.2,  # TODO: different default for volumes
+        specular_power: float = 10.0,  # TODO: different default for volumes
+        render: bool | None = True,
+        user_matrix: NumpyArray[float] | _vtk.vtkMatrix4x4 | None = None,
```
user27182, line 3992:
```suggestion
        user_matrix: TransformLike | None = None,
```

pyvista/plotting/plotter.py
```diff
@@ -3904,36 +3963,36 @@ def _add_legend_label(self, actor, label, scalars, color: Color):
...
+        categories: bool | int = False,
+        culling: CullingOptions | bool = False,
+        multi_colors: bool = False,
+        blending: Literal['additive', 'maximum', 'minimum', 'composite', 'average'] = 'composite',
+        mapper: Literal['fixed_point', 'gpu', 'open_gl', 'smart'] | None = None,
+        scalar_bar_args: ScalarBarArgs | None = None,
+        show_scalar_bar: bool | None = None,
+        annotations: dict[float, str] | None = None,
+        pickable: bool = True,
+        preference: Literal['point', 'cell'] = 'point',
```
user27182, line 3985:
```suggestion
        preference: PointLiteral | CellLiteral = 'point',
```

pyvista/plotting/plotter.py
user27182: > Sequence only takes 1 type argument so that cannot be done. With regards to VectorLike[float], I think that we would like the user to know the number of elements inputted via the type checker and this doesn't allow for that.

We would also like to allow flexibility with the input type and not be overly constrained to only allow tuples when a list can also work... for array-like inputs we typically try to be very general with inputs and narrow the type with input validation.

pyvista/plotting/plotter.py
```diff
@@ -4864,7 +4926,12 @@ def update_scalars(self, scalars, mesh=None, render=True) -> None:
         if render:
             self.render()
 
-    def update_coordinates(self, points, mesh=None, render=True) -> None:
+    def update_coordinates(
+        self,
+        points: pyvista.pyvista_ndarray,
```
user27182, line 4931:
```suggestion
        points: MatrixLike[float],
```

pyvista/plotting/plotter.py
```diff
@@ -4969,31 +5036,42 @@ def close(self):
...
+            'upper_left',
+            'upper_right',
+            'lower_edge',
+            'upper_edge',
+            'right_edge',
+            'left_edge',
+        ]
+        | Sequence[float]
+        | None = 'upper_left',
+        font_size: float | None = 18,
```
user27182, line 5064:
```suggestion
        font_size: int | None = 18,
```

pyvista/plotting/plotter.py
```diff
@@ -5456,35 +5544,35 @@ def remove_scalar_bar(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT
...
-        justification_horizontal=None,
-        justification_vertical=None,
-        background_color=None,
-        background_opacity=None,
-    ):
+        points: MatrixLike[float] | pyvista.DataSet | _vtk.vtkAlgorithm,
+        labels: list[str | int] | str,
+        italic: bool = False,
+        bold: bool = True,
+        font_size: float | None = None,
```
user27182, line 5551:
```suggestion
        font_size: int | None = None,
```

pyvista/plotting/plotter.py
```diff
@@ -5456,35 +5544,35 @@ def remove_scalar_bar(self, *args, **kwargs) -> None:  # numpydoc ignore=PR01,RT
...
+        pickable: bool = False,
+        render_points_as_spheres: bool = False,
+        tolerance: float = 0.001,
+        reset_camera: bool | None = None,
+        always_visible: bool = False,
+        render: bool = True,
+        justification_horizontal: HorizontalOptions | None = None,
+        justification_vertical: VerticalOptions | None = None,
+        background_color: Color | None = None,
+        background_opacity: Color | None = None,
```
user27182, line 5574:
```suggestion
        background_color: ColorLike | None = None,
        background_opacity: float | None = None,
```

pyvista/plotting/plotter.py
```diff
@@ -5803,21 +5896,26 @@ def add_point_scalar_labels(
...
-            scalars = points.point_data[labels]
+            scalars = points.point_data[labels]  # type: ignore[assignment, index]
         phrase = f'{preamble} {fmt}'
         labels = [phrase % val for val in scalars]
         return self.add_point_labels(points, labels, **kwargs)
 
-    def add_points(self, points, style='points', **kwargs) -> Actor:
+    def add_points(
+        self,
+        points: NumpyArray[float] | pyvista.DataSet,
```
user27182, line 5915: All points (here and elsewhere in this PR) should technically allow a single vector or vector array (type should match `DataSet.points`)
```suggestion
        points: VectorLike[float] | MatrixLike[float] | pyvista.DataSet,
```

pyvista/plotting/plotter.py
```diff
@@ -6473,7 +6593,7 @@ def reset_camera_clipping_range(self) -> None:
         """Reset camera clipping planes."""
         self.renderer.ResetCameraClippingRange()
 
-    def add_light(self, light, only_active=False) -> None:
+    def add_light(self, light: Light | _vtk.vtkLight, only_active: bool = False) -> None:
```
user27182, line 6596:
```suggestion
    def add_light(self, light: _vtk.vtkLight, only_active: bool = False) -> None:
```

pyvista/plotting/plotter.py
```diff
@@ -6652,25 +6772,25 @@ class Plotter(BasePlotter):
...
+        border: bool | None = None,
+        border_color: ColorLike = 'k',
+        border_width: float = 2.0,
+        window_size: list[int] | None = None,
+        line_smoothing: bool = False,
+        point_smoothing: bool = False,
+        polygon_smoothing: bool = False,
+        splitting_position: float | None = None,
+        title: str | None = None,
+        lighting: Literal['light kit', 'three lights', 'none'] = 'light kit',
```
user27182, line 6790: This is repeated elsewhere
```suggestion
        lighting: LightingOptions = 'light kit',
```

pyvista/plotting/plotter.py
```diff
@@ -1797,7 +1840,7 @@ def camera_position(self):  # numpydoc ignore=RT01
         return self.renderer.camera_position
 
     @camera_position.setter
-    def camera_position(self, camera_location) -> None:
+    def camera_position(self, camera_location: CameraPosition) -> None:
```
user27182, line 1843: This should be more general and include `xy`, etc. , check out `Renderer.CAMERA_STR_ATTR_MAP.keys()`
This type hint appears elsewhere too (and should also be updated)

tests/plotting/jupyter/test_static.py
user27182: Ok thanks. We'll need to be careful with this. I think `none` is used to mean "Has been set (i.e. is not `None`), but is set a value(i.e. `'none'`)  that conveys that no value is used". So as long as we're staying consistent with that meaning...

The assert `is None` here then might mean that the value has not yet been set by the user, which may be correct.

tests/plotting/test_plotter.py
```diff
@@ -704,7 +704,7 @@ def test_plotter_add_volume_raises(uniform: pv.ImageData, sphere: pv.PolyData):
     with pytest.raises(TypeError, match='not supported for volume rendering'):
         pl.add_volume(sphere)
 
-    with pytest.raises(TypeError, match='not supported for plotting in PyVista'):
+    with pytest.raises(TypeError, match='volume must be an instance of any type'):
```
user27182, line 707: This message doesn't really make sense here... could be a bug with the messaging in `_validation.check_instance`...?

user27182 (commented):
Thanks for all the updates! Round two....

tests/plotting/jupyter/test_static.py
user27182: After review I think this change is OK. Everywhere in the docs it states this should be a string.

maxdswain:
Sorry, I have been busy with work and other stuff. Will get to this by Monday evening latest.

user27182:
@maxdswain I've submitted a PR https://github.com/maxdswain/pyvista/pull/1 with changes for this branch based on the review comments.

If you have limited time, you can simply merge that PR into your branch, CI should be green and we can do a final review and get this merged into PyVista

user27182:
Would love to get this PR merged, just need https://github.com/maxdswain/pyvista/pull/1 to be merged downstream and everything should be good to go (all that is needed is for the "Squash and Merge" button to be pressed).

Alternatively, we can merge #7524 instead of this PR. That PR has all the same commits/contributions from @maxdswain, just without any merge conflicts and with all review comments addressed.

Ideally, it would be better to merge this PR (#7423) instead of #7524, as this would properly attribute credit for the commit to main to @maxdswain. But, without any further action here, perhaps we should consider following through with #7524 instead (@maxdswain will still receive co-author credit for the squashed commit to main).

PR #7425, opened 2025-04-14
Parametrize merge filter test for easier debugging

user27182:
### Overview

This test currently relies on the output from `main_has_priority=True` for testing `main_has_priority=False`. This dependency makes it hard to debug.

Also, using `Plane` with resolution=1 also makes it easier to debug since this mesh only has four scalars to compare instead of hundreds with Sphere.


tkoyama010 (approved):
LGTM

Issue #7428, opened 2025-04-15
`merge` filter merges in wrong order and is not compatible with latest VTK


user27182:
### Describe the bug, what's wrong, and what you expected.

The way the `merge` filter is implemented is problematic.

Currently, it appends the "main" mesh last by default, instead of appending it first. This means doing something like merging a quad cell with a pixel cell will result in the pixel cell being first in the result. But, I would expect the main mesh to be the first mesh, with other meshes appended thereafter.

``` python
import pyvista as pv
from pyvista import examples
quad = examples.cells.Quadrilateral()
quad.point_data['data'] = [0,0,0,0]
pixel = examples.cells.Pixel()
pixel.point_data['data'] = [1,1,1,1]

out = quad.merge(pixel)
```

The `PIXEL` cell is first, but expected QUAD first

``` python
pv.CellType(out.celltypes[0])  # PIXEL
pv.CellType(out.celltypes[1])  # QUAD
```

It seems to me that the reason for doing this has to do with the `main_has_priority` keyword, which aims to ensure that, under special circumstances where the meshes have the same `n_points` and same scalar array name and `merge_points` is enabled, the "main" mesh's _scalars_ have priority.

E.g. here the scalars of the "main" quad mesh are retained (as desired).

``` python
out.point_data.active_scalars  # pyvista_ndarray([0, 0, 0, 0])
```

But, appending the "main" mesh last seems to be more of a workaround to ensure that the scalars are correct, but it has the side-effect of breaking the merge order. So this keyword fixes one problem but creates another.

With the latest vtk-master wheel `vtk-9.4.20250405.dev0`, the merging issue with the scalars has been fixed for `vtkAppendFilter`, see https://gitlab.kitware.com/vtk/vtk/-/issues/19656.

With the latest VTK, we no longer need to append the "main" mesh last to make the scalars merge correctly. Instead, we can simply append all inputs to `vtkAppendFilter` in the order they're given, starting with the main mesh. This is a good thing, and fundamentally fixes the core issue that "main_has_priority" tries to address. However, it now means that the `merge` filter introduces a breaking change, since appending the "main" mesh first will result in a different cell ordering for output meshes.

So what to do? Do we simply introduce a breaking change for "vtk>9.4.2" but leave code for older versions unchanged? Or do we try to patch this behavior to make it consistent across legacy vtk (i.e. ensure correct merge order _and_ correct merging of scalars).

[...]

PR #7430, opened 2025-04-16
Move vtk master testing into main testing workflow

### LINKED ISSUE #7429: Improve VTK master testing

The current test strategy for testing PyVista against the latest VTK dev wheels is not ideal.

Currently, a dedicated workflow is used for this testing, introduced in https://github.com/pyvista/pyvista/pull/3704. Alternatively, `maint/vtk-master*` branch naming may be also be used to enable these for a specific PR.

This is problematic for a few reasons:
- The use of `maint/vtk-master*` branch naming for enabling these tests in PRs is undocumented
- Maintainers have to go out of their way to check for failures (e.g. manually inspect the workflow runs). This means if there are failures or problems with the workflow, they're easily missed. I personally was unaware this workflow even existed for a very long time because it's relatively hidden.
- This workflow was broken for more than 5 months by #6753 and went unnoticed until #7328 was merged, and was mistakenly seen as progress towards VTK 9.4 compatibility (see https://github.com/pyvista/pyvista/issues/6731#issuecomment-2671356371)
- The workflow is independent of the regular testing workflows. This means it gets easily overlooked for CI changes, e.g.
  - #6204 added image uploads, but forgot to add it to the VTK master workflow. 
  - #6894 added color, but forgot to add it to the VTK master workflow
  - #7291 fixed inconsistencies with the `pytest` invocation, but forgot to add it to the VTK master workflow

Due to a lack of "in-your-face" day-to-day vtk compatibility checks and failures, these incompatibilities compounded over time and ultimately culminated in a very drawn-out, complex, and time-consuming endeavor to make PyVista compatible with VTK 9.4 (see #6731 and #6879). Instead of us being able to detect these and fix them quickly, instead it seems we are relying on downstream issues to surface before we take action (e.g. #7405).

I personally would rather avoid these headaches and address any vtk compatibility failures as soon as they're identified. This would make it much easier to maintain, and could possibly lead to patches being fixed as part of a VTK release right away so that we don't have to create workarounds (e.g. the workaround needed in #7371 for the clip filter could have been avoided if we simply caught and reported the underlying VTK bug sooner).

Based on early testing for the latest VTK compatibility (see https://github.com/pyvista/pyvista/pull/7410), there are already core failures with the `merge` filter as well as a number of plotting failures, including a segmentation fault. These fixes are already creating headaches and it seems once again we may be on the road towards complex and time-consuming fixes to make PyVista compatible again with the latest VTK.

I would personally like to see vtk-master testing directly integrated with our regular CI testing (i.e. add it to the testing matrix). Perhaps these can be enabled by default, but optionally skipped with a `skip-vtk-master` label, and always skipped for pushes to `main`. This way, if there are vtk-master failures outside of our control, we can temporarily suppress them with a label, but still be reminded of this issue for the next PR. This should also ensure we always have fully green CI,  and should avoid failure fatigue which was reported as a concern in https://github.com/pyvista/pyvista/pull/3704#issuecomment-1512445765.

### Links to source code.

_No response_

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

user27182:
### Overview

Fix #7429

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -113,6 +113,9 @@ jobs:
           - python-version: "3.13"
             vtk-version: "latest"
             numpy-version: "nightly"
+          - python-version: "3.13"
+            vtk-version: "dev"
+            numpy-version: "latest"
     steps:
       - uses: actions/checkout@v4
         with:
@@ -139,8 +142,13 @@ jobs:
           pip install dist/pyvista*.whl
 
       - name: Set up vtk
-        if: ${{ matrix.vtk-version != 'latest' }}
+        if: ${{ matrix.vtk-version != 'latest'  && matrix.vtk-version != 'dev'  }}
         run: pip install vtk==${{ matrix.vtk-version }}
+
+      - name: Install latest VTK dev wheel
+        if: ${{ matrix.vtk-version == 'dev' }}
+        run: |
+          pip install --upgrade vtk --pre --no-cache --extra-index-url https://wheels.vtk.org
 
       # Make sure PyVista does not break from non-core dependencies
       - name: Software Report (Core Dependencies)
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -113,9 +113,6 @@ jobs:
           - python-version: "3.13"
             vtk-version: "latest"
             numpy-version: "nightly"
-          - python-version: "3.13"
-            vtk-version: "dev"
-            numpy-version: "latest"
     steps:
       - uses: actions/checkout@v4
         with:
@@ -142,13 +139,8 @@ jobs:
           pip install dist/pyvista*.whl
 
       - name: Set up vtk
-        if: ${{ matrix.vtk-version != 'latest'  && matrix.vtk-version != 'dev'  }}
+        if: ${{ matrix.vtk-version != 'latest' }}
         run: pip install vtk==${{ matrix.vtk-version }}
-
-      - name: Install latest VTK dev wheel
-        if: ${{ matrix.vtk-version == 'dev' }}
-        run: |
-          pip install --upgrade vtk --pre --no-cache --extra-index-url https://wheels.vtk.org
 
       # Make sure PyVista does not break from non-core dependencies
       - name: Software Report (Core Dependencies)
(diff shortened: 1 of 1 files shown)


user27182:
With the new label applied, testing is run against the latest vtk wheels:
https://github.com/pyvista/pyvista/actions/runs/14483232798/job/40624180614?pr=7430

user27182:
And with the label removed, the testing is skipped:
https://github.com/pyvista/pyvista/actions/runs/14483938655/job/40625989614?pr=7430

.github/workflows/testing-and-deployment.yml
```diff
@@ -215,6 +215,77 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
           TWINE_REPOSITORY_URL: "https://upload.pypi.org/legacy/"
 
+  VTK-dev:
+    # For PRs, only run this job if the 'vtk-dev-testing' label is applied
+    if: |
+      (github.event_name != 'pull_request' && github.event_name != 'push') ||
+      contains(github.event.pull_request.labels.*.name, 'vtk-dev-testing')
+    name: Linux VTK Dev Testing
```
user27182, line 223: The term "VTK dev" is used here instead of "VTK master". We can change this if "master" is preferred.

I think "master" is misleading because it implies that we are directly installing from the master branch https://gitlab.kitware.com/vtk/vtk/-/tree/master, which is something we _can_ do for python packages. But for VTK we are installing from pre-built wheels, which seem to only be published every so often (maybe once every few weeks...? not sure exactly).

user27182: Also, I think this logic _should_ allow for us to manually run this job independent of a PR, though I'm not 100% sure. We should consider removing `vtk-pre-test.yml` and using this new job exclusively for VTK dev testing.

CONTRIBUTING.rst
```diff
@@ -533,6 +533,23 @@ for more details.
...
+^^^^^^^^^^^^^^^^^^^^^
+Most unit testing is run with stable VTK releases. However, it is sometimes useful to
+run tests with the latest VTK dev wheels. To install these locally, run
+
+.. code-block:: shell
+
+    pip install vtk --upgrade --pre --extra-index-url https://wheels.vtk.org
+
+For CI on GitHub, the ``vtk-dev-testing`` label can be used to enable unit testing with
+the VTK dev wheels. The tests only run when the label is applied.
```
user27182, line 546: We could also invert this, and always run these tests _unless_ a `skip-vtk-dev-testing` label is applied

In the default case, we should probably always be running these tests, and only skip in rare cases (e.g. a breaking change was introduced by VTK and we need time to address it)

MatthewFlamm: I am inclined to agree with your suggestion to invert this.  The last vtk release cycle uncovered that we are very slow to respond to failures due to changes in vtk and/or numpy. This is not a good situation.  The downside to this suggestion is that we will have to apply labels to PRs when we start having issues, but this will likely be less maintenance headache overall than continuing to hold vtk issues in the back burner.

.github/workflows/testing-and-deployment.yml
MatthewFlamm: I'm +1 for the naming here.  I think we have two thumbs up, so resolving it.

MatthewFlamm:
This one is becoming more important with 9.5 release coming in just weeks. 
 https://discourse.vtk.org/t/vtk-9-5-0-release/15018/6

Let's get this in one way or another so we can test better.  I like the idea of having it be tested by default, but turned off with a label flag.  If so, we will have to add this label to all non-related PRs right now.

user27182:
> we will have to add this label to all non-related PRs right now.

There is a known issue / breaking change with the `merge` filter we will need to address first #7428. Then after that there are (or were, maybe this has been fixed upstream) seg faults with the plotting tests. Until then we'll need to apply the label. 

IMO these issues can't be properly addressed without first integrating VTK master testing as part of every PR. So yes we should move this PR forward

banesullivan (approved):

MatthewFlamm:
> We could also invert this, and always run these tests unless a skip-vtk-dev-testing label is applied

I'm fine to merge this one as is too.  Whatever is your plan @user27182.

user27182:
> > We could also invert this, and always run these tests unless a skip-vtk-dev-testing label is applied
> 
> I'm fine to merge this one as is too. Whatever is your plan @user27182.

Let's merge as-is for now, we can invert the logic post vtk 9.5 support once VTK dev testing is green

user27182:
> I think this logic should allow for us to manually run this job independent of a PR, though I'm not 100% sure. We should consider removing vtk-pre-test.yml and using this new job exclusively for VTK dev testing.

Confirmed that this the VTK Dev test runs automatically with the scheduled job:

https://github.com/pyvista/pyvista/actions/runs/15482866640

Should we remove the `vtk-pre-test.yml` workflow then?

PR #7433, opened 2025-04-16
Add mapper utilities to handle deprecated calls for VTK 9.5

user27182:
### Overview

VTK pre-release is issuing this warning:

``` shell
 "Call to deprecated method GetInputAsDataSet. (Use GetDataSetInput() instead.) -- Deprecated since version 9.5.0."
```
This PR adds utility functions to address this


user27182:
Ideally it would also be nice to add the `vtk-dev-testing` label from #7430 to test against the vtk pre-release wheels

tkoyama010 (approved):
LGTM

PR #7441, opened 2025-04-18
Use the `tox` package for CI and CD

beroda:
### Overview

Attempt to introduce tox as CI/CD orchestrator.


user27182:
Looks like some of the test times have increased substantially, e.g. windows now takes 1hr instead of 25 mins. I'm thinking this is caused by display drivers no longer being available...  Maybe the tox environment is too isolated, such that the environment changes from the headless display action https://github.com/pyvista/setup-headless-display-action have no effect in the tox env?

beroda:
> Looks like some of the test times have increased substantially, e.g. windows now takes 1hr instead of 25 mins. I'm thinking this is caused by display drivers no longer being available... Maybe the tox environment is too isolated, such that the environment changes from the headless display action https://github.com/pyvista/setup-headless-display-action have no effect in the tox env?

Nice catch. Added env variables from action in https://github.com/pyvista/pyvista/commit/98929061aab4f9e765a80747deb60b2e49f0dc8f

beroda:
> > Looks like some of the test times have increased substantially, e.g. windows now takes 1hr instead of 25 mins. I'm thinking this is caused by display drivers no longer being available... Maybe the tox environment is too isolated, such that the environment changes from the headless display action https://github.com/pyvista/setup-headless-display-action have no effect in the tox env?
> 
> Nice catch. Added env variables from action in [9892906](https://github.com/pyvista/pyvista/commit/98929061aab4f9e765a80747deb60b2e49f0dc8f)

No effect. In fact, all test times have increased (os, linux, windows). Might be due to coverage flags ... Hard to track.

beroda:
> > > Looks like some of the test times have increased substantially, e.g. windows now takes 1hr instead of 25 mins. I'm thinking this is caused by display drivers no longer being available... Maybe the tox environment is too isolated, such that the environment changes from the headless display action https://github.com/pyvista/setup-headless-display-action have no effect in the tox env?
> > 
> > 
> > Nice catch. Added env variables from action in [9892906](https://github.com/pyvista/pyvista/commit/98929061aab4f9e765a80747deb60b2e49f0dc8f)
> 
> Passing all env variables had no effect. In fact, all test times have increased (os, linux, windows). Might be due to coverage flags ... Hard to track.

Was coverage indeed. The test times are now correct (see https://github.com/pyvista/pyvista/actions/runs/14695356868/job/41236215770?pr=7441 for macOS)

beroda:
> > > Looks like some of the test times have increased substantially, e.g. windows now takes 1hr instead of 25 mins. I'm thinking this is caused by display drivers no longer being available... Maybe the tox environment is too isolated, such that the environment changes from the headless display action https://github.com/pyvista/setup-headless-display-action have no effect in the tox env?
> > 
> > 
> > Nice catch. Added env variables from action in [9892906](https://github.com/pyvista/pyvista/commit/98929061aab4f9e765a80747deb60b2e49f0dc8f)
> 
> Passing all env variables had no effect. In fact, all test times have increased (os, linux, windows). Might be due to coverage flags ... Hard to track.

The increase of test time was caused by coverage which was enabled for all os.
When using coverage for Linux only (as it was before), the test times are more consistent with before.

I guess the Linux test times have increased due to `xvfb-run` being used for both core and plotting tests (was only plotting tests on main).

See below for more details.
| This PR                                                                                                                                                                             | main                                                                                     |
|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------|
| image image | image image |

.github/workflows/vtk-pre-test.yml
```diff
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       # To resolve issues on VTK master and test, use this branch name pattern
-      - "maint/vtk-master*"
+      - "maint/tox*"
```
beroda, line 8: Changed temporarily to test that vtk master job is correctly running

beroda: The vtk-master jobs run as expected, seeing the same errors as in another PR that runs the same job https://github.com/pyvista/pyvista/actions/runs/14483232798/job/40624180614?pr=7430

See https://github.com/pyvista/pyvista/actions/runs/14763559674

.github/workflows/linkcheck.yml
```diff
@@ -1,6 +1,7 @@
 name: LinkCheck - All files 🔗
 
 on:
+  pull_request:
```
beroda, line 4: Added temporary to check if job runs fine

beroda: The job output [here](https://github.com/pyvista/pyvista/actions/runs/14801773525/job/41561958259) is the same as in master branch [here](https://github.com/pyvista/pyvista/actions/runs/14688722200/job/41220938057)

user27182:
Haven't looked through the changes, but it seems like the test times have increased substantially, which isn't ideal. Any idea why?

| main                              | tox                              |
|---------------------------------------|---------------------------------------|
|  |  |
|  |  |
|  |  |

edabor:
@user27182 I'm investigating test time increase using `pytest-durations` in [08e2d4a](https://github.com/pyvista/pyvista/pull/7441/commits/08e2d4a28fe1ab74021b49d371377cdad5c9e25c)
This is weird indeed that test times have increased ...

edabor:
The increase in test time is caused by `core` and `plotting` tests being regrouped in one `pytest` command, not by tox.
See the comparison with https://github.com/pyvista/pyvista/pull/7601 where all tests are regrouped:

| https://github.com/pyvista/pyvista/pull/7601                                              | This PR                                                                                   |
|-------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|
| image | image |
| image | image |


It seems that the `check_gc` fixture takes an important amount of time, around 12 min see https://github.com/pyvista/pyvista/actions/runs/15560275642/job/43810575203#step:8:10198.
I guess this is because many objects are created internally by `pytest` when running all tests and running `gc.collect()` might inflate because of that.

Will investigate deeper by splitting the fixture to get in-depth insights ...

user27182:
> The increase in test time is caused by `core` and `plotting` tests being regrouped in one `pytest` command, not by tox. 

Is there a technical reason for doing this? The plotting tests also run with `xvfb-run` on Linux which will also slow things down I think. IMO it's nice to have the tests split up like this (i.e. separate core and plotting) because within 5 minutes it's possible to easily see if core tests have failed and see the full test log. Otherwise, the logs are truncated in GitHub actions (only the last 1000 lines or so are shown), which means that you typically can't see _which_ test failed until the action completes, which can require waiting 45 minutes for some of the Linux tests with coverage enabled.

edabor:
> Is there a technical reason for doing this? The plotting tests also run with `xvfb-run` on Linux which will also slow things down I think. IMO it's nice to have the tests split up like this (i.e. separate core and plotting) because within 5 minutes it's possible to easily see if core tests have failed and see the full test log. Otherwise, the logs are truncated in GitHub actions (only the last 1000 lines or so are shown), which means that you typically can't see _which_ test failed until the action completes, which can require waiting 45 minutes for some of the Linux tests with coverage enabled.

There no specific reason other than reducing the CI complexity. But I agree that it can have some advantages to split the test suite.
Done in [51414d9](https://github.com/pyvista/pyvista/pull/7441/commits/51414d9a2b48fb7d53e13f2305935cd0ab87d52c)

edabor:
Test times are back to previous.

.github/workflows/testing-and-deployment.yml
```diff
@@ -284,64 +224,49 @@ jobs:
...
-    name: Windows Unit Testing
     runs-on: windows-latest
     env:
       CI_WINDOWS: true
+      PYVISTA_USERDATA_PATH: ${{ github.workspace }}\.pyvista_examples
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+    name: Windows Unit Testing - Python ${{ matrix.python-version }}
```
user27182: If the name is being changed, can we shorten it a bit? Maybe remove "Unit Testing"? With the longer names, the important bits are now hidden in the desktop gui:
ss

The Linux names will be extra long... maybe we need to also shorten "Python 3.9" to "py3.9", etc. ?

edabor: Done in 1465bbfcefaf934c4ea0f9e5789c9551acbdca81

user27182: Looks good, thanks!
ci

user27182 (approved):
LGTM. 
Using `tox` does add complexity and introduces a bit of a learning curve for maintainers to figure out how to modify this in the future, but IMO this change is very much needed to unify and consolidate all of the environments and commands.

edabor:
@banesullivan Since you're code owner of most CI/CD related files, do you want ownership of the `tox.ini` file ?

.github/workflows/docs.yml
```diff
@@ -97,18 +87,7 @@ jobs:
...
-      - name: Dump Sphinx Warnings and Errors
-        if: always()
-        run: if [ -e doc/sphinx_warnings.txt ]; then cat doc/sphinx_warnings.txt; fi
-
-      - name: Dump VTK Warnings and Errors
-        if: always()
-        run: if [ -e doc/errors.txt ]; then cat doc/errors.txt; fi
-
-      - name: Test Documentation
-        run: pytest tests/doc/tst_doc_build.py
```
user27182: Can we keep these as separate jobs? It helps isolate the logs for easier review. It might be OK to keep the two "dump" jobs together with the build, but if so then I think the names will need to be more descriptive instead of "commands_post[0]" and "commands_post[1]", because it's not clear what's going on here:
```
  docs-build: commands_post[0] /home/runner/work/pyvista/pyvista/doc> python -c 'from pathlib import Path;print(file.open('"'"'r'"'"').read() if (file:=Path('"'"'sphinx_warnings.txt'"'"')).exists() else '"'"''"'"');'
  
  docs-build: commands_post[1] /home/runner/work/pyvista/pyvista/doc> python -c 'from pathlib import Path;print(file.open('"'"'r'"'"').read() if (file:=Path('"'"'errors.txt'"'"')).exists() else '"'"''"'"');'
  
  docs-build: OK (5471.29=setup[9.23]+cmd[3.45,5399.72,58.76,0.06,0.06] seconds)
  congratulations :) (5471.45 seconds)
```

But I do think we should at least keep the test logs separate from the build logs.

edabor: Modified accordingly in https://github.com/pyvista/pyvista/pull/7441/commits/477033fad92093692f682bfd4331fa9a0a4dda56

edabor:
https://github.com/pyvista/pyvista/pull/7441/commits/dbff254286c26511fc62978c2d1ee5e27491e1fc is anticipating https://github.com/pyvista/pyvista/pull/7627 to fix geovista integration tests

bjlittle:
@beroda Rebase to include the fix #7627 now available in `main` 👍

tox.ini
```diff
@@ -0,0 +1,270 @@
...
+    vtk_main: uv pip install --upgrade vtk --pre --no-cache --extra-index-url https://wheels.vtk.org
+    {[testenv]software_report_cmdline}
+
+setenv =
+    TOX_ROOT = {tox_root}
+
+commands =
+    {[testenv]all_testing_cmdline}
+
+cov_flags =
```
user27182, line 80: Code coverage is failing, I think this may be caused by #7610 which used `--cov-append`, but this option is now missing, meaning only coverage for charts tests is being uploaded.

We can add it here, or wait until #7612 is merged which reverts this change from #7610 (i.e. splitting plotting tests is no longer required)

edabor: I think it is more subtle.
Codedov is already combining reports, so the `--cov-append` is not needed and was already working before.

However, it turns out that the 2 plotting tests are producing the same coverage xml file, meaning that only charts tests (which is in second position) are covered.

I think we can wait for #7612

user27182 (approved):

user27182:
@beroda do you think you could resolve the merge conflicts? And @banesullivan are you able to review this?

I'd like to update CI to include another docstring job which runs with the oldest supported python to help address #7378, and it'll be easier to do that after this PR lands. I'd also like to extend the tests with the `vtk-dev-testing` label applied so that the docstrings and docs build are also run using the latest vtk dev wheels (currently only the Linux unit tests install these). Again, it would be easier to do after this PR lands.

tox.ini
```diff
@@ -0,0 +1,271 @@
...
+    test \
+    --cov-report \
+    term:skip-covered \
+    --cov-report \
+    xml:coverage_{env_name}.xml \
+    --cov-report \
+    html:coverage_{env_name}
+
+common_flags = --color=yes -v
+image_flags = --fail_extra_image_cache --generated_image_dir debug_images
```
user27182: These options were removed in #7659

edabor: Done in https://github.com/pyvista/pyvista/pull/7441/commits/07d98376c69cd687d55bdffe2b2e863b4e6430ba

.github/workflows/integration-tests.yml
```diff
@@ -34,119 +34,50 @@ permissions:
...
-        run: |
-          pip install --upgrade pip
-          pip install -ve . --group test
+          python-version: "3.12"
+          activate-environment: true
+
+      - name: Install tox-uv
+        run: uv pip install tox-uv
+
+      - name: Run tests (no geovista)
```
user27182: I'm curious to know why this is needed to have (no geovista) and (geovista), I guess it's because of the need for `xvfb-run`? While reviewing the logs, it seemed odd to see `geovista` mentioned in the `trame` job. Would it be better to mention `xvfb` instead?
```suggestion
      - name: Run tests (no xvfb)
```

tox.ini
```diff
@@ -0,0 +1,270 @@
...
+    mne: uv pip install git+https://github.com/pyvista/pyvistaqt.git
+    mne: uv pip install -e {env:MNE_HOME}
+    mne: mne sys_info -p
+    mne: bash -c "cd {env:MNE_HOME} && ./tools/get_testing_version.sh"
+    mne: bash -c "cd {env:MNE_HOME} && ./tools/github_actions_download.sh"
+    mne: python -c "import pyvista; assert not pyvista.OFF_SCREEN, f'{pyvista.OFF_SCREEN=} should be False'"
+
+    # PYVISTAQT SETUP
+    pyvistaqt: git clone https://github.com/pyvista/pyvistaqt.git {env:PYVISTAQT_HOME} --single-branch
+    pyvistaqt: uv pip install -e {env:PYVISTAQT_HOME} -r {env:PYVISTAQT_HOME}{/}requirements_test.txt
```
user27182: Is this what the original test did? It seems to be uninstalling the pyvista dev version:

https://github.com/pyvista/pyvista/actions/runs/16340976849/job/46163267386?pr=7441#step:6:62
``` diff
- pyvista==0.46.dev0 (from file:///home/runner/work/pyvista/pyvista/.tox/.tmp/package/1/pyvista-0.46.dev0-py3-none-any.whl)
+ pyvista==0.45.2
```

edabor: Good catch ! Indeed, the `pyvista dev` package is installed before `pyvistaqt`, which caused unfortunate package removal.

Thanks to your review, I checked the previous behavior and I found that the `geovista` integration tests were also erroneously removing the `dev pyvista`. See https://github.com/pyvista/pyvista/actions/runs/16416742646/job/46384438671#step:7:373

I fixed everything in https://github.com/pyvista/pyvista/pull/7441/commits/6727a7c76170302e5edaba2692e9a2ab24044bc2 

See the correct version of `dev pyvista` in the geovista integration test https://github.com/pyvista/pyvista/actions/runs/16417880417/job/46388245697?pr=7441#step:7:241

user27182: thank you!

akaszynski:
I'm conflicted about the use of tox. On the plus side, you can run workflows locally and avoid the headache of debugging on GitHub. On the otherhand as @beroda pointed out:

> Needs for dev to learn `tox` ini file syntax

There's a bit of a learning curve with this.

I'm leaning towards using tox, especially after the challenge of refactoring the parallel doc build #7800.

edabor:
> I'm conflicted about the use of tox. On the plus side, you can run workflows locally and avoid the headache of debugging on GitHub. On the otherhand as @beroda pointed out:
> 
> > Needs for dev to learn `tox` ini file syntax
> 
> There's a bit of a learning curve with this.
> 
> I'm leaning towards using tox, especially after the challenge of refactoring the parallel doc build #7800.

@akaszynski Thanks for considering this PR 😄. I agree that the learning curve is an issue.

Given the fact that this PR regularly conflicts with main developments, it starts to become tedious to fix those conflicts.
I fully acknowledge that reviewing such PR takes time due to its size since it covers most CI/CD aspects, namely:

- test suite
- documentation building and tests
- doctests
- type checking
- integration tests

I leaning towards splitting this PR for each aforementioned aspect to keep changes easier to review. What are your thoughts about it ?

akaszynski:
Fully agree regarding splitting up the PR. Changes are easier to swallow when they're smaller.

akaszynski:
Also, special request: can we run the CI locally using our new self hosted docker images for the Linux runners?

edabor:
> Also, special request: can we run the CI locally using our new self hosted docker images for the Linux runners?

Could you detail your request ? I don't have a lot of skills regarding docker images.

akaszynski:
> Could you detail your request ? I don't have a lot of skills regarding docker images.

Yes. For our self-hosted Linux runners we're going to use [ghcr.io/pyvista/arc-runners:ubuntu22.04-latest](https://github.com/pyvista/arc-runners/pkgs/container/arc-runners/491943527?tag=ubuntu22.04-latest). This is effectively the same environment as GitHub's self-hosted runners except with a few VTK and doc build dependencies.

A cursory search reveals that there is a plug-in to use docker with tox ([tox-docker](https://github.com/tox-dev/tox-docker)), but I've not reviewed it. It's not entirely necessary, just curious if it's possible for us to do this.

Issue #7307, opened 2025-03-13
`Plotter` scalars do not match `subplot` scalars


user27182:
### Describe the bug, what's wrong, and what you expected.

I am working on an example for computing cell quality as part of #7306.
I have found that if I plot four meshes separately, their scalars look very different when compared to plotting them with `subplot`. This seems like a pretty significant bug.

### Steps to reproduce the bug.

This will plot four quality measures using subplot.
``` python
import pyvista as pv
from pyvista import examples

def add_mesh(plotter, mesh, scalars=None, cmap='bwr', show_edges=True, zoom=1.3):
    plotter.add_mesh(mesh, scalars=scalars, cmap=cmap, show_edges=show_edges)
    plotter.view_xy()
    plotter.camera.zoom(zoom)
    
mesh = examples.download_cow().triangulate().decimate(0.7)
measures = ['area', 'max_angle', 'min_angle', 'shape']
for measure in measures:
    mesh[measure] = mesh.compute_cell_quality(measure)['CellQuality']

pl = pv.Plotter(shape=(2,2))
pl.link_views()
pl.subplot(0,0)
add_mesh(pl, mesh,scalars=measures[0])
pl.subplot(0,1)
add_mesh(pl, mesh,scalars=measures[1])
pl.subplot(1,0)
add_mesh(pl, mesh,scalars=measures[2])
pl.subplot(1,1)
add_mesh(pl, mesh,scalars=measures[3])
pl.show()
```
This is the generated image, but it's not correct. The first three plots are wrong.

![Image](https://github.com/user-attachments/assets/33a319cd-59af-4e67-bba1-0835b83bc1d1)

These are the expected plots:
``` python
pl = pv.Plotter()
add_mesh(pl, mesh,scalars=measures[0])
pl.show()
pl = pv.Plotter()
add_mesh(pl, mesh,scalars=measures[1])
pl.show()
pl = pv.Plotter()
add_mesh(pl, mesh,scalars=measures[2])
pl.show()
pl = pv.Plotter()
add_mesh(pl, mesh,scalars=measures[3])
pl.show()
```

| ![Image](https://github.com/user-attachments/assets/c299dd55-bda5-4450-bb05-9a1dcd5b64c3) | ![Image](https://github.com/user-attachments/assets/d79e7458-14fd-43f2-aa1c-fdaee1d5ee72) |
|----------------------------------------------------------|----------------------------------------------------------|
[...]

Issue #542, opened 2020-01-15
Unexpected colormapping result using subplots


MatthewFlamm:
**Describe the bug, what's wrong, and what you expect:**

See possibly related issue #539.

I'm trying implement the `component` functionality for vector plotting, and am almost finished.  I intended to use a very simple subplot example to show the different components, but it doesn't behave as expected.  Specifically, when you do not use subplot, the `clim` argument changes the resulting colors plotted.  However, when you use subplot, the `clim` argument _does not_ change the resulting colors plotted and it instead plots based on max and min values, but the scalar bar _does_ change.

-----

**To Reproduce**

Please include a code snippet to reproduce the bug in the block below:

```py
import pyvista as pv
import numpy as np
values_small = np.linspace(0, 1, 1000)
values_large = np.linspace(0, 1000, 1000)

shape = (20, 5, 10)
grid = pv.UniformGrid()
grid.dimensions = np.array(shape) + 1
grid.origin = (100, 33, 55.6)  # The bottom left corner of the data set
grid.spacing = (1, 5, 2)  # These are the cell sizes along each axis
grid["values_small"] = values_small
grid["values_large"] = values_large

p1 = pv.Plotter()
p1.add_mesh(grid, scalars="values_small", clim=[0, 1000])
p1.show()

p2 = pv.Plotter()
p2.add_mesh(grid, scalars="values_large", clim=[0, 1000])
p2.show()

p3 = pv.Plotter(shape=(1,2))
p3.subplot(0,0)
p3.add_mesh(grid, scalars="values_small", clim=[0, 1000], show_scalar_bar=True)
p3.subplot(0,1)
p3.add_mesh(grid, scalars="values_large", clim=[0, 1000], show_scalar_bar=True)
p3.show()
```

**Screenshots**
![image](https://user-images.githubusercontent.com/39341281/72473771-661ce780-37b5-11ea-89e9-5c3edef36065.png)
![image](https://user-images.githubusercontent.com/39341281/72473820-7cc33e80-37b5-11ea-97e3-226ea71e5b29.png)
![image](https://user-images.githubusercontent.com/39341281/72473840-88af0080-37b5-11ea-97b7-8f0f59963db5.png)

-----

**System Information:**
 Please run the following code wherever you are experiencing the bug and paste the output below. This report helps us track down bugs and it is critical to addressing your bug:

```py
# Get system info
import pyvista as pv
print(pv.Report())
```

```txt
--------------------------------------------------------------------------------
[...]

PR #7444, opened 2025-04-20
Add example on sharing scalar bars

navlalli:
### Overview

This adds an example that shows how to share scalar bars among plotted arrays or use a unique scalar bar for
each plotted array. It took me a bit of time to figure out this behaviour when I first started using pyvista, so I thought this example could be helpful to others.

### Additional details

I was unable to build the docs to test this example on both Linux (Ubuntu 20.04.2) and MacOS (12.6.3). I followed the instructions on the [contribution page](https://docs.pyvista.org/extras/developer_notes):

I installed the required packages with `python -m pip install -e '.[dev]'` and then tried to build the docs with `make -C doc html`. There were then module not found errors. After installing those modules using
```
pip install networkx shapely enum-tools[sphinx]
```
I then got a segmentation fault when I tried to build the docs
```
Fatal Python error: Segmentation fault
```
I assume this isn't an issue provided no one else is experiencing this.


examples/02-plot/sharing_scalar_bars.py
```diff
@@ -0,0 +1,98 @@
...
+pl.subplot(0, 1)
+pl.add_mesh(cube, scalars="scalars2", cmap='coolwarm')
+pl.camera.zoom(0.75)
+pl.subplot(1, 0)
+pl.add_mesh(cube, scalars="scalars3", cmap='coolwarm')
+pl.camera.zoom(0.75)
+pl.subplot(1, 1)
+pl.add_mesh(cube, scalars="scalars4", cmap='coolwarm')
+pl.camera.zoom(0.75)
+pl.show()
```
user27182, line 73: This part of the example likely also requires (shallow) copies of the mesh to be passed with `add_mesh` to avoid issue #542 / #7307

Might be good to include a note about this since this is non-obvious and not really documented.

examples/02-plot/sharing_scalar_bars.py
```diff
@@ -0,0 +1,98 @@
...
+# By default, a unique scalar bar is used for all plotted scalars with unique names:
+
+# Create a cube mesh with 4 scalar arrays
+cube = pv.Cube()
+cube.cell_data['scalars1'] = range(0, 6, 1)
+cube.cell_data['scalars2'] = range(0, 12, 2)
+cube.cell_data['scalars3'] = range(0, 18, 3)
+cube.cell_data['scalars4'] = range(0, 24, 4)
+
+# Plot the 4 scalar arrays - not how there is 1 scalar bar for each plotted scalar
```
user27182, line 60:
```suggestion
# Plot the 4 scalar arrays - note how there is 1 scalar bar for each plotted scalar
```

examples/02-plot/sharing_scalar_bars.py
navlalli: Thanks for the info, I've added this and a note on why this is done.

navlalli:
> The contributing page should be updated to indicate that `pip install -e '.[docs]'` should be used instead of dev.
> 
> It may be easier to simply rely on CI on GitHub to build the docs instead of building it locally.

Good to know, thank you!

user27182:
See also https://docs.pyvista.org/extras/developer_notes.html#documentation-image-regression-testing
to resolve the errors from the build

user27182:
pre-commit.ci autofix

tkoyama010 (approved):
LGTM

Issue #5111, opened 2023-10-20
More complete linking of API methods in Gallery Examples


MatthewFlamm:
### Describe what maintenance you would like added.

The gallery examples are extremely useful to help users get started.  But, often I want to jump from the example to the detailed documentation for the particular method that is used in the example.  An example of a gallery example that does this well is https://docs.pyvista.org/version/stable/examples/01-filter/clipping.html, where links to `clip`, `clip_box` and so on are provided.  An example of a gallery example that does not currently do this well is https://docs.pyvista.org/version/stable/examples/01-filter/project-plane.html where there is no link to `project_points_to_plane`.

### Links to source code.

Example that links to detailed documentation:
https://github.com/pyvista/pyvista/blob/v0.42.3/examples/01-filter/clipping.py

Example that doesn't link to detailed documentation:
https://github.com/pyvista/pyvista/blob/v0.42.3/examples/01-filter/project-plane.py

RequieMa:
I'd like to help this while learning the examples. 
Which branch should I merge if I create a pull request?

MatthewFlamm:
You can use #5811 as an example.  Fork this repo, create a branch off `main`, make changes on that branch, then make PR here requesting to merge to main.

PR #7451, opened 2025-04-23
Add sphinx gallery reference tests

user27182:
### Overview

A more comprehensive fix for https://github.com/pyvista/pyvista/issues/5111 with tests to ensure compliance.

This PR adds tests to ensure that each gallery example file:

- has a reference anchor `.. _<example_name>_example:` at the top of the file
- the filename matches the reference anchor name
- includes at least one reference to the API methods
- has at least one reference from the API methods to the example

All non-compliant gallery examples are modified to ensure compliance with the above, including renaming files, renaming the reference tag, and adding new references to/from each example as required.


MatthewFlamm (approved):
I am in favor of the strategy and overall implementation here.  My weakness is regex, which is a key part of this implementation.  Hopefully someone else can review that portion.

PR #7452, opened 2025-04-24
Add links to PyVista methods inside Sphinx Gallery code blocks

user27182:
### Overview

Related to #5111

The sphinx gallery uses intersphinx to add hyperlinks to docs from  `python`, `numpy`, `matplotlib`, `scipy`, etc. The links are added inside of code blocks, e.g. see this example:
https://dev.pyvista.org/examples/99-advanced/warp_by_vector_eigenmodes.html

Links to PyVista methods are currently missing. This PR adds them.

See https://sphinx-gallery.github.io/stable/configuration.html#add-intersphinx-links-to-your-examples


user27182:
Looks like the PyVista methods are all being linked, which is nice. Some erroneous links still exist though.

MatthewFlamm (approved):
This works overall great.  Some missing bits like `pyvista.examples`, but this simple addition gets the majority in my poking around.

user27182:
> Some missing bits like `pyvista.examples`

I think something like #7453 is needed to fix this. Sphinx only knows `pyvista.examples.examples` and `pyvista.examples.downloads`, we have to somehow make those both resolve to `pyvista.examples` in the docs...

PR #7453, opened 2025-04-24
Make `pyvista.examples` referenceable in the docs

user27182:
### Overview

The docs do not define `pyvista.examples` such that it's possible to do `pyvista.examples.load_uniform` in code but in the docs this is not recognized. Instead, it's necessary to do `pyvista.examples.examples.load_uniform` for cross-referencing. This PR fixes this by adding this entry explicitly.


user27182:
This was also tried in #8883 and it's too complex with only minimal benefit

PR #7459, opened 2025-04-26
Use dependency groups instead of extras for installing dev dependencies

user27182:
### Overview

Pip now supports dependency groups, see:
https://ichard26.github.io/blog/2025/04/whats-new-in-pip-25.1/#dependency-groups-pep-735

This PR moves the `dev`, `test`, `typing`, and `docs` extras from `[project.optional-dependencies]` into `[dependency-groups]`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ jupyter = [
   'trame>=2.5.2',
 ]
 
+[dependency-groups]
 pinned = [ # Pinned versions of core dependencies
   'matplotlib<3.10.2',
   'numpy<2.2.0',
@@ -64,8 +65,8 @@ typing = [
   'mypy<1.14.0',
   'npt-promote==0.2',
   'numpy>=2.0.0',
-  'pyvista[pinned]',
   'trimesh<4.6.0',
+  { include-group = 'pinned' },
 ]
 
 test = [
@@ -89,7 +90,6 @@ test = [
   'pytest<8.4.0',
   'pytest_cases<3.8.7',
   'pytest_mock<3.15.0',
-  'pyvista[pinned]',
   'scipy<1.14.2',
   'sphinx-book-theme<1.2.0',
   'sphinx-gallery<0.19.0',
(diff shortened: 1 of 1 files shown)


banesullivan:
So this requires anyone installing PyVistas to have the latest `pip`? This feels a little too soon to me

user27182:
pip reminds you to update it every time it runs... and also the longer we wait on this, the longer the current `pyvista[test|dev|docs|typing]` installation will be exposed as a public install method, which we never really intended to make public, but was only public due to the lack of support for something like PEP 735 until now.

So I guess we have two problems: 

1. If we update to using `--group` now, the install may break if you don't have the latest pip
2. If we update to `--group` later, the install will break unless you change `pyvista[test]` to `pyvista --group test`

Neither is ideal. But since `pyvista[test]` was only released as part of `0.45`, it might be best to nip this in the bud now before it gets widely adopted downstream. There doesn't seem to be a deprecation mechanism to move between these two.

E.g. sphinx has already adopted `--group` for their CI, but removing the install from extras will break hundreds of downsteam packages that use them:
https://github.com/search?q=%2Fsphinx%5C%5B%28docs%7Clint%7Ctest%29%2F&type=code

Compare this to downsteam projects using the pyvista extras:
https://github.com/search?q=%2Fpyvista%5C%5B%28docs%7Ctest%7Cdev%7Ctyping%29%2F&type=code

user27182:
> So this requires anyone installing PyVista to have the latest `pip`? This feels a little too soon to me

Yes and no. Installing from `main` will still work. The new `[dependency-groups]` table is simply ignored by older versions of pip and will run just fine. This will only break if you want to install `dev|docs|typing|test` dependencies, in which case yes, the latest pip is required.

But not earlier releases. How long is appropriate to wait otherwise? If 6 months is OK, then you can plan on holding off for the 0.46 release until 6 months from now

banesullivan (approved):
Ah okay, I understand, thanks for the explanation here! I was originally thinking that this would affect optional deps like the `jupyter` section but it doesn't.

+1 for this

banesullivan (approved):

PR #7462, opened 2025-04-26
Enable `abort_on_example_error` for sphinx gallery

user27182:
### Overview

If there is an error with the sphinx gallery, it's better to fail fast so we can see the error quickly and rectify it. This will help avoid having to wait the full 2hrs for the build, only to see an error that could have been caught in the first 10 minutes.

See https://sphinx-gallery.github.io/stable/configuration.html#abort-build-on-first-fail


tkoyama010 (approved):
LGTM

PR #7463, opened 2025-04-26
Add command line interface with `--version` and `report` options

user27182:
### Overview

Enable `python -m pyvista --version` and `python -m pyvista report`.

This is easier than doing `python -c "import pyvista; print(pyvista.Report())"`
The CLI can be extended in future PRs with a `pyvista convert` option (similar to `meshio convert`), and a quick/simple `pyvista plot` option for quickly reading and plotting a file.

### DIFF
--- a/pyvista/__main__.py
+++ b/pyvista/__main__.py
@@ -1,0 +1,127 @@
+"""PyVista's command-line interface."""
+
+from __future__ import annotations
+
+import argparse
+import inspect
+import re
+
+import pyvista as pv
+
+
+def _parse_numpy_doc(doc: str) -> dict[str, str]:
+    """Parse a NumPy-style docstring to get parameter names and help strings."""
+    help_map: dict[str, str] = {}  # Store parameter name to description mapping
+    if not doc:
+        return help_map
+
+    lines = doc.splitlines()
+    in_params = False  # Flag to indicate whether we are inside the 'Parameters' section
+    current_param = None  # Temporary variable
+
+    # Iterate through lines of the docstring
+    for line in lines:
+        stripped = line.strip()
+
+        # Detect the start of the 'Parameters' section
+        if re.match(r'^Parameters$', stripped):
+            in_params = True
+            continue
+
+        # Skip the underline section of 'Parameters'
+        if in_params and re.match(r'^-{3,}$', stripped):
+            continue
+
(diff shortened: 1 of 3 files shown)


pyvista/__main__.py
```diff
@@ -0,0 +1,99 @@
...
+    if argv is None:
+        argv = sys.argv[1:]
+
+    # Create top-level parser with --version option
+    parser = argparse.ArgumentParser(prog='pyvista')
+    parser.add_argument(
+        '--version',
+        '-v',
+        action='version',
+        version=f'PyVista {pyvista.__version__}',
```
user27182: Minor detail, but `pytest --version` returns with lower-case name, e.g. `pytest 8.4.2`, whereas `python --version` returns with upper-case "P", e.g. `Python 3.13.5`, so we could optionally use all lower-case:
```suggestion
        version=f'pyvista {pyvista.__version__}',
```

pyvista/__main__.py
```diff
@@ -0,0 +1,99 @@
...
+    """PyVista Command-Line Interface entry point."""
+    # Use sys.argv if no arguments were explicitly passed
+    if argv is None:
+        argv = sys.argv[1:]
+
+    # Create top-level parser with --version option
+    parser = argparse.ArgumentParser(prog='pyvista')
+    parser.add_argument(
+        '--version',
+        '-v',
```
user27182: We could remove this and possibly reserve it for a future `--verbose` flag? `-v` is short for `--verbose` for Python and pytest, but `git -v` prints the version. Not exactly sure what "verbose" means for PyVista though...

```suggestion
```

akaszynski: Verbose doesn't really make sense here unless we're running a report, but I can understand the -v flag confusion. Recommend not including `-v` and keep it explicit.

pyvista/__main__.py
akaszynski: I'd recommend lowercase since it matches `pip show | grep pyvista` or `pip show pyvista`, but I could go either way.

akaszynski (approved):
This is great!

We might also want to think about incorporating `vedo`'s awesome examples (see https://vedo.embl.es/) with:

```py
pyvista examples --list
pyvista examples create-spline
```

PR #7473, opened 2025-04-27
Limit dependabot PRs to two and update weekly instead of monthly

user27182:
### Overview

Dependabot chokes up the CI in this repo when it opens more than 2-3 PRs at once. I propose limiting this to two so it does not overwhelm our CI, and update more frequently instead.


tkoyama010 (approved):
LGTM

Issue #7448, opened 2025-04-23
GLTF exports have buggy label coloring


Jorissss:
### Describe the bug, what's wrong, and what you expected.

Following e.g. the example at https://dev.pyvista.org/api/core/_autosummary/pyvista.imagedatafilters.contour_labels, the colors should look like:

![Image](https://github.com/user-attachments/assets/ca267fda-e431-4085-bfbb-a34602096c30)

But they end up looking like:

![Image](https://github.com/user-attachments/assets/ea6d1828-28e3-4b85-88bb-e6cf3d2f1fda)

when viewed at e.g. https://gltf-viewer.donmccurdy.com/

### Steps to reproduce the bug.

```python
import pyvista as pv
import numpy as np
from pyvista import examples
image = examples.load_channels()
label_ids = np.unique(image.active_scalars)
image = image.extract_subset(voi=(75, 109, 75, 109, 85, 100))

def labels_plotter(mesh, zoom=None):
    colored_mesh = mesh.color_labels(negative_indexing=True)
    plotter = pv.Plotter()
    plotter.add_mesh(colored_mesh, show_edges=True)
    if zoom:
        plotter.camera.zoom(zoom)
    return plotter

labels_plotter(image).export_gltf("labels_plotter.gltf")
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Wed Apr 23 08:54:06 2025 UTC

                  OS : Linux (Ubuntu 22.04)
              CPU(s) : 96
             Machine : x86_64
        Architecture : 64bit
                 RAM : 377.6 GiB
         Environment : Jupyter
          GPU Vendor : Mesa
        GPU Renderer : llvmpipe (LLVM 15.0.7, 256 bits)
         GPU Version : 4.5 (Core Profile) Mesa 23.2.1-1ubuntu3.1~22.04.3
       Render Window : vtkXOpenGLRenderWindow
    MathText Support : True

  Python 3.11.0rc1 (main, Aug 12 2022, 10:02:14) [GCC 11.2.0]

             pyvista : 0.45.0
                 vtk : 9.4.2
               numpy : 1.26.4
          matplotlib : 3.10.1
              scooby : 0.10.0
               pooch : 1.8.2
              pillow : 11.2.1
             imageio : 2.37.0
             IPython : 9.1.0
            colorcet : 3.1.0
             cmocean : 4.0.3
          ipywidgets : 8.1.6
               scipy : 1.15.2
                tqdm : 4.67.1
[...]

PR #7475, opened 2025-04-29
Add note to `color_labels` indicating that `colorcet` is required by default

user27182:
### Overview

See https://github.com/pyvista/pyvista/issues/7448#issuecomment-2837450543

~This was tested locally but no tests are included because it requires not having `colorcet` installed.~
EDIT: Error handling is addressed in #7477 , this PR mostly just addresses the docs issue


pyvista/core/filters/data_set.py
```diff
@@ -7020,8 +7025,12 @@ def _is_index_like(array_, max_value):
...
                 try:
                     cmap = get_cmap_safe(colors)
-                except ValueError:
-                    pass
+                except ValueError as e:
+                    if (  # pragma: no cover
+                        'colorcet' in (cmap_repr := repr(e)) or 'cmocean' in cmap_repr
+                    ):
+                        # Valid cmap but package is not installed.
+                        raise
```
tkoyama010, line 7033: We should raise `ModuleNotFoundError` in my mind.

tkoyama010 (changes requested):

pyvista/core/filters/data_set.py
user27182: I'm ok with that. This code is basically re-raising this error:

https://github.com/pyvista/pyvista/blob/main/pyvista/plotting/colors.py#L1225-L1236

Should we change that to an ImportError too? Or just here with the re-raise?

user27182: Making the original source raise an import error would simplify the re-raise, because then we don't need to check the repr here, which we are only doing because the ValueError there is currently "masking" the true cause which is really an import error

So yes probably best to raise an ImportError inside get_cmap_safe

user27182:
The issue is not specific to `color_labels` and is more general to how cmaps are validated. PR #7477 addresses this problem more generally.

user27182:
> The issue is not specific to `color_labels` and is more general to how cmaps are validated. PR #7477 addresses this problem more generally.

This PR still has valid changes to the docs so we can keep this open for those.

tests/core/test_validation.py
```diff
@@ -1114,3 +1115,14 @@ def test_validate_dimensionality(dimensionality, reshape, expected_dimensionalit
...
+
+def test_validate_color_sequence_raises():
+    match = 'Input must be a single ColorLike color or a sequence of ColorLike colors.'
+    with pytest.raises(ValueError, match=match):
+        _validate_color_sequence('foo', n_colors=None)
+
+    n_colors = 42
+    match = f'Input must be a single ColorLike color or a sequence of {n_colors} ColorLike colors.'
+    with pytest.raises(ValueError, match=match):
+        _validate_color_sequence('foo', n_colors=n_colors)
```
tkoyama010, line 1128:
```suggestion


@pytest.mark.parametrize(["n_colors", "match"], [(None, 'Input must be a single ColorLike color or a sequence of ColorLike colors.'), (42, 'Input must be a single ColorLike color or a sequence of 42 ColorLike colors.')])
def test_validate_color_sequence_raises(match):
    with pytest.raises(ValueError, match=match):
        _validate_color_sequence('foo', n_colors=n_colors)
```

tkoyama010 (changes requested):

tkoyama010 (approved):
LGTM

PR #7476, opened 2025-04-29
Fix docstring paths for non-posix platforms

beroda:
### Overview

Some doctests are failing in local on a Windows machine.
This is due to file separation here for example.
https://github.com/pyvista/pyvista/blob/9ccc3ff5029d73fa6eab9f798b4615303f65b315/pyvista/core/utilities/reader.py#L197-L203

This PR fixes them.


tkoyama010 (approved):
LGTM

MatthewFlamm:
Not for this PR, but should we be running the doctests on a windows runner on the CI?  Image regression is done through documentation building, so it would only be done to see if the code can run without error.

user27182:
> Not for this PR, but should we be running the doctests on a windows runner on the CI? Image regression is done through documentation building, so it would only be done to see if the code can run without error.

+1, can maybe partially address #7378 too by running them on python 3.9

MatthewFlamm (approved):

tkoyama010 (approved):

PR #7082, opened 2025-01-11
Improve colormap error messages when `colorcet` or `cmocean` are not installed

user27182:
### Overview

Currently, if I don't have the right colormap package installed, I get this error, e.g:

``` python
import pyvista as pv
pv.get_cmap_safe('glasbey')
# ValueError: Invalid colormap "glasbey"
```

This message isn't ideal because , it _is_ a valid colormap, I just don't have the right package installed.

This PR improves the error messages.

CASE: Neither `colorcet` or `cmocean` installed:
``` python
pv.get_cmap_safe('topo')
# ValueError: Colormap 'topo' is not recognized but may be a valid `colorcet` or `cmocean` colormap.
# Install `colorcet` or `cmocean` and try again.
```

CASE: Only `colorcet` installed:
``` python
pv.get_cmap_safe('topo')
# ValueError: Colormap 'topo' is not recognized but may be a valid `cmocean` colormap.
# Install `cmocean` and try again.
```

CASE: Only `cmocean` installed
``` python
pv.get_cmap_safe('glasbey')
# ValueError: Colormap 'glasbey' is not recognized but may be a valid `colorcet` colormap.
# Install `colorcet` and try again.
```

This has been tested locally but no tests are included because this requires dynamically modifying the test environment (or creating a workflow with different environments)

PR #7477, opened 2025-04-29
Improve cmap error handling

user27182:
### Overview

Follow-up to #7082, and fix error messaging for `color_labels`, see https://github.com/pyvista/pyvista/issues/7448#issuecomment-2837450543

#7082 improved error handling for when a cmap is valid but the module is not installed. However, it's not possible to report if a `ValueError` should be raised (invalid) or if an `ImportError` should be raised.

This PR adds lists with known valid cmaps for `colorcet` and `cmocean` so that we can distinguish these cases and properly raise an import error


tkoyama010 (approved):
LGTM

PR #7478, opened 2025-04-29
Remove memory profiling from testing dependencies

beroda:
### Overview

When test fails on CI, it can take a while to scroll until the failures details due to the `pytest-memprof` (long) summary.
Moreover, the package does not seem maintained anymore ([last release 2019](https://pypi.org/project/pytest-memprof/#history)) and there is no documentation.

This PR removes it.

I wanted to replace by [memray](https://pytest-memray.readthedocs.io/en/latest/index.html) but it's only available on posix platforms 😢


tkoyama010 (approved):
LGTM

PR #7479, opened 2025-04-30
Fix flaky download test using `retry-requests`

beroda:
### Overview

Testing flakiness was due to url requests being temporary unavailable (or something else) and consequently, the `AttributeError` was triggered in the `pytest.fail` message.

This PR fixes flakiness by:
- retrying multiple times the request 
- fixing the `AttributeError` in the test


user27182:
These tests are also pretty slow, see #7158. Not sure if there's a better way to test the downloads...

tkoyama010 (approved):
LGTM

PR #7486, opened 2025-05-03
Add type hints for setting colormaps

user27182:
### Overview

Add type hints for all named colormaps (and their aliases) available from `colorcet`, `cmocean`, and `matplotlib`.


pyvista/plotting/_typing.py
```diff
@@ -17,8 +17,16 @@
...
     from .charts import ChartPie as ChartPie
+    from .colors import _CMOCEAN_CMAPS_LITERAL
+    from .colors import _COLORCET_CMAPS_LITERAL
+    from .colors import _MATPLOTLIB_CMAPS_LITERAL
     from .colors import Color as Color
 
+
+ColormapOptions = Union[
+    '_MATPLOTLIB_CMAPS_LITERAL', '_CMOCEAN_CMAPS_LITERAL', '_COLORCET_CMAPS_LITERAL'
+]
```
user27182, line 28: Should we define these conditionally based on the actual installed colormaps? I am slightly in favor of keeping it as-is and raising an import error if trying to use one that isn't available. This allows users to discover new cmaps they may not have known were available to use, and it's easy enough to install the missing package or choose a different cmap

tkoyama010: > I am slightly in favor of keeping it as-is and raising an import error if trying to use one that isn't available

+1

tkoyama010 (approved):
LGTM

PR #7489, opened 2025-05-05
Improve StructuredGrid example

akaszynski:
PyVista's [StructuredGrid docstring example](https://docs.pyvista.org/api/core/_autosummary/pyvista.structuredgrid) uses both `np.arange` and `ij` indexing:

```py
>>> xrng = np.arange(-10, 10, 2, dtype=np.float32)
>>> yrng = np.arange(-10, 10, 5, dtype=np.float32)
>>> zrng = np.arange(-10, 10, 1, dtype=np.float32)
>>> x, y, z = np.meshgrid(xrng, yrng, zrng, indexing='ij')
>>> grid = pv.StructuredGrid(x, y, z)
>>> grid
StructuredGrid (...)
  N Cells:      513
  N Points:     800
  X Bounds:     -1.000e+01, 8.000e+00
  Y Bounds:     -1.000e+01, 5.000e+00
  Z Bounds:     -1.000e+01, 9.000e+00
  Dimensions:   10, 4, 20
  N Arrays:     0
```

~First, for consistency with [other examples](https://docs.pyvista.org/examples/00-load/create_structured_surface.html), we should be using `xy` indexing. Even though it's possible to use `ij` indexing here, there's really no reason to since this evaluates to `True` regardless if we use `xy` or `ij`.~

> Nevermind, I'm completely wrong about this and we must have `ij` indexing. I'm going to specifically call that out. In fact, it's the example that's wrong here.

Secondly, the usage of `np.arange` causes the right endbounds to be non-inclusive, which is unintuitive to new users. For example, the resulting range for `x` above evaluates to `(-10, 8)`. Instead, we should use `np.linspace` with clear bounds and then be explicit how the dimensionality lines up. Here's the improved example:

```py
>>> xrng = np.linspace(-5, 5, 10)
>>> yrng = np.linspace(-8, 8, 4)
>>> zrng = np.linspace(-7, 4, 20)
>>> x, y, z = np.meshgrid(xrng, yrng, zrng, indexing='ij')
>>> grid = pv.StructuredGrid(x, y, z)
>>> grid
StructuredGrid (...)
  N Cells:      513
  N Points:     800
  X Bounds:     -5.000e+00, 5.000e+00
  Y Bounds:     -8.000e+00, 8.000e+00
  Z Bounds:     -7.000e+00, 4.000e+00
  Dimensions:   10, 4, 20
  N Arrays:     0

Note how the grid dimensions match the shape of the input arrays.

>>> (xrng.size, yrng.size, zrng.size)
(10, 4, 20)
```


tkoyama010 (approved):
LGTM

PR #7490, opened 2025-05-05
Fix the magnetic coil example

akaszynski:
Fix the URL in the [Plot a Magnetic Field](https://docs.pyvista.org/examples/99-advanced/magnetic_fields#sphx-glr-examples-99-advanced-magnetic-fields-py) and fix the opacity when plotting the magnetic field strength as @banesullivan's fixes in #7358 seems to have affected this example.

### `main`
(Way too dark)
image

---

### This branch `docs/fix-magnetic-coil-example`

image


banesullivan (approved):

banesullivan:
I don't think #7358 caused this as it was cached as this dark image a a good while before that work. See https://github.com/pyvista/pyvista/blob/main/tests/doc/doc_image_cache/sphx_glr_magnetic_fields_003.jpg

Also you may need to reset the doc image cache for this

akaszynski:
> I don't think #7358 caused this as it was cached as this dark image a a good while before that work. See https://github.com/pyvista/pyvista/blob/main/tests/doc/doc_image_cache/sphx_glr_magnetic_fields_003.jpg
> 
> Also you may need to reset the doc image cache for this

You're right, my bad. It's just too dark.

akaszynski:
> It's just too dark.

Actually, does anyone else agree with that?

MatthewFlamm:
> > It's just too dark.
> 
> Actually, does anyone else agree with that?

In a binary choice, I like the darker version better.  The yellow in the hot colormap is perceptually closer to white in value, so those yellow regions pop slightly better in the darker version in my opinion. But maybe something in between would be optimal.

Colormaps are typically designed to look good without opacity curves.  It follows that colormaps should  be changed when using opacity curves to optimize the rendering.  I find that the effect is better when the dynamic range of the colormap is shifted so that it lies within the region of non-transparent part of the opacity curve.  However, in my experience this only works if you aren't showing a scalar colorbar, as people will get confused why more than half of the scalar colorbar mapping is red.

akaszynski:
@MatthewFlamm, please let me know if this works for you. It's a bit more in-between.

image

MatthewFlamm:
> @MatthewFlamm, please let me know if this works for you. It's a bit more in-between.


This one looks much better thanks!

MatthewFlamm (approved):
Significantly better now!

banesullivan:
Looks soo much better, nice job

PR #7493, opened 2025-05-06
Fix "Suggest edit" button and remove button for `.rst` and `.pdf` downloads

user27182:
### Overview

The gallery examples, e.g.: https://dev.pyvista.org/examples/00-load/create_draped_surface

Have an "edit this page button" which yield a 404 error, e.g.: https://github.com/pyvista/pyvista/edit/main/doc/source/examples/00-load/create_draped_surface.rst

We can see this URL is for the generated `.rst` file in `doc/source/examples`, but we want to point to the `.py` file instead: https://github.com/pyvista/pyvista/edit/main/examples/00-load/create_draped_surface.py

This PR fixes this, along with the download button, which was broken the same way.

### Edit button

The earlier attempts patched `doc/source/_templates/edit-this-page.html`, which is never rendered — it belongs to `pydata-sphinx-theme` and has been dead since we moved to `sphinx-book-theme` in #6369, so it is removed here. The button we actually render is built in Python by `sphinx_book_theme`'s `add_source_buttons`, so it is patched in `context["header_buttons"]` instead, from a `html-page-context` handler at priority 502 (above the theme's 501).

For API pages, the edit link now also lands on the same lines as the `[source]` button, in GitHub's edit UI, instead of the top of the file.

### Download button

`sphinx-book-theme` adds a `.rst`/PDF download button whenever a page has a source suffix, without checking `sourcename` — which Sphinx leaves empty when `html_copy_source` is disabled, as it is for us. So every page in the docs linked to a bare `_sources/` that is never written (https://docs.pyvista.org/_sources/ → 404). Reported in https://github.com/pyvista/pyvista/issues/8871

Rather than repoint it, it's dropped entirely: there's nothing worth downloading — "Print to PDF" is a browser feature, not a download. Gallery examples keep their own `.py`/`.ipynb`/`.zip` downloads via sphinx-gallery's own footer links, untouched.

### Also

`linkcode_resolve` now emits `https://` rather than `http://`, affecting every `[source]` link.


banesullivan:
Good catch! Updating this brach to get the preview and see if this works

user27182:
I recall I wasn’t able to get this to work and gave up trying

banesullivan:
Still not working!

user27182:
Edit this page links are working 🎉

akaszynski (approved):

PR #7496, opened 2025-05-06
Add `splat_fast` method to `pyvista.DataSetFilter`

tkoyama010:
### Overview

Add `splat_fast` method to `pyvista.DataSetFilter`.

### Details

- None

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -7544,6 +7544,41 @@ class DataSetFilters(DataObjectFilters):
 
         return output_volume
 
+    def splat_fast(
+        self: DataSet,
+        *,
+        progress_bar: bool = False,
+    ) -> pyvista.PolyData:
+        """Splat optimized for splatting single kernels.
+
+        This filter is a wrapper around `vtkFastSplatter
+        <https://vtk.org/doc/nightly/html/classvtkFastSplatter.html>`_.
+
+        Parameters
+        ----------
+        progress_bar : bool, default: False
+            Display a progress bar to indicate progress.
+
+        Returns
+        -------
+        pyvista.ImageData
+            The splat meshes.
+
+        Examples
+        --------
+        Splat a mesh.
+
+        >>> from pyvista import examples
+        >>> mesh = examples.download_bunny_coarse()
+        >>> splat_meshes = mesh.splat_fast()
+        >>> splat_meshes.plot()
+
+        """
+        alg = _vtk.vtkFastSplatter()
+        alg.SetInputData(self)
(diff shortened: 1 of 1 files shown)


akaszynski:
~Ideally, we should use a point cloud as an example mesh.~

Actually, the Stanford bunny is a great dataset for this.

pyvista/core/filters/data_set.py
```diff
@@ -7923,6 +7924,89 @@ def _preprocess_polydata(poly_in):
...
+        """
+        # build kernel image
+        x = np.linspace(-1, 1, kernel_dimensions[0])
+        y = np.linspace(-1, 1, kernel_dimensions[1])
+        z = np.linspace(-1, 1, kernel_dimensions[2])
+        xv, yv, zv = np.meshgrid(x, y, z, indexing='ij')
+        kernel = (1 - np.abs(xv)) * (1 - np.abs(yv)) * (1 - np.abs(zv))
+        kernel = np.clip(kernel, 0, 1)
+
+        kernel_image = pyvista.ImageData(dimensions=kernel_dimensions)
```
akaszynski:
```suggestion
        kernel_image = pv.ImageData(dimensions=kernel_dimensions)
```

pyvista/core/filters/data_set.py
```diff
@@ -7923,6 +7924,89 @@ def _preprocess_polydata(poly_in):
 
         return output_volume
 
+    def fast_splat(  # type: ignore[misc]
+        self: DataSet,
+        bounds: BoundsTuple | None = None,
+        kernel_dimensions: tuple[int, int, int] = (10, 10, 10),
+        splat_dimensions: tuple[int, int, int] = (100, 100, 100),
+        *,
```
user27182, line 7932:
```suggestion
    def fast_splatting(  # type: ignore[misc]
        self: DataSet,
        bounds: VectorLike[float] | None = None,
        splat_dimensions: VectorLike[int] = (100, 100, 100),
        *,
        kernel_dimensions: VectorLike[int] = None,
```

- We already have `gaussian_splatting`, should we similarly use splatting in the name?
- BoundsTuple is too restrictive as an input type, since it does not allow for regular tuple
- consider allowing more general VectorLike inputs (though tuple is OK for now for all these args 
)
- I'm not a huge fan of allowing two dimension positional args, IMO it will make it unclear which is the kernel and which is the output dimensions
- 100, 100, 100 seems okay as a default splat dimensions, though this is (50, 50, 50) for Gaussian splatting. Should they be the same? Or maybe more dimensions makes sense since it's fast?
- is it typical for kernel dimensions to be 1/10 of output dimensions ? Should we make this default to None, and set it to 1/10 of splat dimensions by default?

Issue #7485, opened 2025-05-02
InteractorStyle resets unexpectedly after rapidly two left clicks (VTK + pyvistaqt)


skyui1997:
### Describe the bug, what's wrong, and what you expected.

Hello, I found a bug!

The interactor style is consistently changed to an instance of a class called CustomStyle when double left clicked.

I verified this by:

Observing that InteractorStyle is correctly set at first (my own subclass of vtkInteractorStyleJoystickCamera).
When double left clicked in the plotter it gets replaced by a dynamically generated class called CustomStyle, located in pyvista.plotting.render_window_interactor.\_style_factory, inheriting from vtkInteractorStyleTrackballCamera. 
This behavior was observed using both PySide6 and PyQt6 bindings.

This means that the CustomStyle class is created dynamically by PyVista within its render_window_interactor module and inherits from vtkInteractorStyleTrackballCamera.

### Steps to reproduce the bug.

```python


import pyvista as pv
from pyvistaqt import BackgroundPlotter
from PyQt6 import QtWidgets
from PyQt6.QtCore import QTimer, QTime
import vtk
import sys


class HeightmapViewer(QtWidgets.QMainWindow):
    def __init__(self):
        super().__init__()
        self._style_message_set = False
        self.setWindowTitle("VTK JoystickCamera Style Demo - DEBUGGING v2")
        self.setGeometry(100, 100, 900, 700)
        self.central_widget = QtWidgets.QWidget()
        self.setCentralWidget(self.central_widget)
        self.main_layout = QtWidgets.QHBoxLayout(self.central_widget)

        self.plotter = BackgroundPlotter(off_screen=False, show=False, toolbar=False)
        self.vtk_widget = self.plotter.interactor
        self.main_layout.addWidget(self.vtk_widget)

        mesh = pv.Sphere(radius=0.5)
        self.plotter.add_mesh(mesh, color='lightblue')
        self.plotter.reset_camera()

        print("INFO: Setting vtkInteractorStyleJoystickCamera...")
        joystick_style = vtk.vtkInteractorStyleJoystickCamera()
        # Sanity check right after setting
        print(f"INFO: Style immediately after creation: {type(joystick_style).__name__} ({hex(id(joystick_style))})")
        self.plotter.interactor.SetInteractorStyle(joystick_style)
        # Sanity check right after setting it on the interactor
        style_after_set = self.plotter.interactor.GetInteractorStyle()
[...]

Issue #6320, opened 2024-06-30
`plot_directive` generates build image even if `plot()` or `show()` is not called


user27182:
### Describe the bug, what's wrong, and what you expected.

The following docstring from #6318 creates a `Plotter` instance but does not show it (the Plotter is only used to create an actor).
https://github.com/pyvista/pyvista/blob/2cc37cc8fbc306d2e95ff3d608bc9112eb6a6468/pyvista/plotting/prop3d.py#L403-L420

But the build fails because a build image is unexpectedly generated. See failure here:
https://github.com/pyvista/pyvista/actions/runs/9729484955/job/26851429802#step:14:91

The workaround is to add the generated build image to `tests/doc/doc_image_cache`, but this should not be necessary and I think is bloat for the cache/repository.

### Steps to reproduce the bug.

See docstring above

### System Information

```shell
See CI build info
```

user27182:
This happened again with the new `Prop3D.size` property in #7092, an image is generated even though nothing is plotted

MatthewFlamm:
Is it possible to pass the `nofigs` option to the `pyvista-plot`` directive in the example blocks?

PR #7497, opened 2025-05-06
Add ability to use custom interactor styles

MatthewFlamm:
### Overview

This PR refactors the way InteractorStyle's are created and used.  This will ease implementation of, although is not required for, #7359.  An added bonus is that it now becomes easier for users to implement custom interactors.  See #7485.

This PR also will make it easier to address #7500.

### Details

Instead of using a style factory to generate wrapped vtk classes, use an inheritance strategy.  This enables users to pass in a subclass of `vtk.vtkInteractorStyle` and hook into the machinery in pyvista's `RenderWindowInteractor`

See also https://github.com/pyvista/pyvista/pull/4798.

Interactors are hard to write good tests for, it would useful to live test this.

### TODO
- [x] add tests for built in interactor styles
  - some may already be covered by existing tests, but more comprehensive
- [x] add tests for new setter functionality
- [x] add test for custom interactor styles
- [x] add better documentation for setting custom style
  - I don't fully understand the reasoning behind the events that are added for the built-in functionality, but we may want to warn users about this in some form.
- [x] run manual/live tests using built-in interactors
- [x] run manual/live tests using custom interactor styles (not tested, but built-in interactors use same machinery)
- [x] run manual/live tests using charts interaction (adjusts interactor style)
- [x] run manual/live tests using affine widget (adjusts interactor style)


pyvista/plotting/render_window_interactor.py
```diff
@@ -433,26 +437,31 @@ def key_press_event(self, *args):
...
             The current interactor style.
 
         """
-        if self._style_class is None:
-            self.update_style()
         return self._style_class
 
+    @style.setter
+    def style(self, style: _vtk.vtkInteractorStyle):
+        # TODO If a plain vtk class, wrap it in pyvista class???
```
MatthewFlamm, line 460: Always wrapping with custom pyvista functionality might lower flexibility for users?  Additionally, context style interactor functionality currently uses the private `_style_class` attribute, but if this changes in the future to use this setter, we would need to not modify the style passed in here.

pyvista/plotting/render_window_interactor.py
```diff
@@ -1540,63 +1529,81 @@ def close(self):
         self._timer_event = None
 
 
-def _style_factory(klass):
-    """Create a subclass with capturing ability, return it."""
-    # We have to use a custom subclass for this because the default ones
-    # swallow the release events
-    # http://vtk.1045678.n5.nabble.com/Mouse-button-release-event-is-still-broken-in-VTK-6-0-0-td5724762.html
+class _CustomInteractorStyle:
```
MatthewFlamm, line 1532: The functionality of `CustomInteractorStyle` is refactored from the old `CustomStyle` inside `_style_factory`

pyvista/plotting/render_window_interactor.py
```diff
@@ -1528,75 +1525,94 @@ def close(self):
...
-    # swallow the release events
-    # http://vtk.1045678.n5.nabble.com/Mouse-button-release-event-is-still-broken-in-VTK-6-0-0-td5724762.html
+class _CustomInteractorStyle(_vtk.vtkInteractorStyle):
+    def __init__(self, render_window_interactor: RenderWindowInteractor):
+        super().__init__()
+        self._parent = weakref.ref(render_window_interactor)
+
+        self._observers = []
+        self._observers.append(
+            self.AddObserver('LeftButtonPressEvent', partial(try_callback, self._press)),
```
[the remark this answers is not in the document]
MatthewFlamm: The second overload has the first parameter as `char*`.  It seems likely to me that this is some sort of vtk python issue.  I'll ignore here with a comment.

pyvista/plotting/render_window_interactor.py
```diff
@@ -433,26 +441,35 @@ def key_press_event(self, *args):
...
             The current interactor style.
 
         """
-        if self._style_class is None:
-            self.update_style()
         return self._style_class
 
+    @style.setter
+    def style(self, style: Optional[Union[_vtk.vtkInteractorStyle, _CustomInteractorStyle]]):
+        # TODO If a plain vtk class, wrap it in pyvista class???
```
MatthewFlamm, line 468:
```suggestion
```

MatthewFlamm:
I live tested all the interactor styles using the below code.  I am not familiar with all the interactor styles, but it seemed like they all worked as intended.  I want to do a few more live tests as in the description, but this is ready for review now.

```py
import pyvista as pv
pl = pv.Plotter()
styles = [attr for attr in dir(pl) if attr.startswith('enable_') and attr.endswith('_style')]
for style in styles:
    pl = pv.Plotter()
    pl.add_mesh(pv.Cone(center=(0, 0, 0)), name='Cone')
    pl.add_mesh(pv.Cube(center=(1, 0, 0)), name='Cube')
    pl.add_mesh(pv.Sphere(center=(1, 1, 0)), name='Sphere')
    pl.add_mesh(pv.Cylinder(center=(0, 1, 0)), name='Cylinder')
    pl.add_text(style)
    getattr(pl, style)()
    pl.show()
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -433,26 +441,58 @@ def key_press_event(self, *args):
...
+    def style(
+        self,
+    ) -> Optional[
+        Union[_vtk.vtkContextInteractorStyle, _vtk.vtkInteractorStyle, _CustomInteractorStyle]
+    ]:
+        """Get/set the current interactor style.
+
+        .. warning::
+
+            Setting an interactor style needs careful control of events handling.
```
MatthewFlamm, line 460: We could also consider exposing `_CustomInteractorStyle` as public and put in the docs explicitly.  Users could then subclass from it to get the recommended PyVista functionality for interactor styles.  Looking for feedback on this.

MatthewFlamm:
Tested chart interaction and scene interaction interoperability using this example https://docs.pyvista.org/examples/02-plot/chart_overlays

AffineWidget3D also seems to work as the code is written, but I uncovered a bug in testing it (even on the main branch).  This is in https://github.com/pyvista/pyvista/issues/7500.

user27182:
A blank image is (falsely) generated by the plot directive for the `style` docstring... this is the same as issue #6320

MatthewFlamm:
Using an explicit `.. pyvista-plot::` directive with `:nofigs:` option was successful.  Docs for `style` look good to me.  If #7502 is successful and merged, we can revert https://github.com/pyvista/pyvista/pull/7497/commits/89f6ed8f20e3b4bd1db1ed113ac2b06a617a847b.

pyvista/plotting/render_window_interactor.py
```diff
@@ -1528,75 +1551,96 @@ def close(self):
...
+
+    def add_observer(self, event, callback):
+        self._observers.append(self.AddObserver(event, callback))
+
+    def remove_observers(self):
+        for obs in self._observers:
+            self.RemoveObserver(obs)
+
+
+class InteractorStyleImage(_CustomInteractorStyle, _vtk.vtkInteractorStyleImage):
```
MatthewFlamm, line 1611: I assume all of these also have to inherit from `DisableVtkSnakeCase`?

user27182: Probably. In theory the `test_snake_case_api` tests should be picking this up automatically to enforce this... might be an issue with the tests

user27182: Only classes found in these modules are subject to the test for conformance (inheritance) with disabling vtk's snake case API:
https://github.com/pyvista/pyvista/blob/54d8532a008ab7afee41d583f9fe3f0728c2ee81/tests/test_snake_case_api.py#L29-L30

Exposing `_CustomInteractorStyle` as a public class as suggested in https://github.com/pyvista/pyvista/pull/7497#discussion_r2077974809 would cause it to be subject to this test automatically.

Should we update the test to search for any/all classes defined everywhere in all `.py` files? Clearly, it's possible that there are classes defined elsewhere (either private or public) that are not covered by this test.

MatthewFlamm: > Exposing `_CustomInteractorStyle` as a public class as suggested in [#7497 (comment)](https://github.com/pyvista/pyvista/pull/7497#discussion_r2077974809) would cause it to be subject to this test automatically.

Let's do this for this PR.  The style classes could be added to the documentation too.  I will mark as draft in the meantime as I likely won't get to it for a few days (or more)

MatthewFlamm: > Should we update the test to search for any/all classes defined everywhere in all .py files? Clearly, it's possible that there are classes defined elsewhere (either private or public) that are not covered by this test.

I believe the intent of this class is to override all objects inherited from vtk, so it would make sense to me to test for all classes.

doc/source/api/plotting/index.rst
```diff
@@ -35,7 +35,17 @@ all plotting functionality in PyVista.
    CameraPosition
    CornerAnnotation
    CubeAxesActor
+   CustomInteractorStyle
```
MatthewFlamm, line 38: This doesn't work presumably because these aren't in the `pyvista` namespace.  I'm thinking we should use a separate heading here like `Interactor Style API` to include these.

pyvista/plotting/render_window_interactor.py
MatthewFlamm: I made these public, but not in the main pyvista namespace.  So #7509 was needed to check this in the end.  The use case for users using these classes is narrow.  The API documentation now suggests users to utilize the `enable_*_style` methods unless customizing the styles.

pyvista/plotting/render_window_interactor.py
```diff
@@ -433,26 +442,61 @@ def key_press_event(self, *args):
...
+        self,
+    ) -> Optional[
+        Union[_vtk.vtkContextInteractorStyle, _vtk.vtkInteractorStyle, CaptureInteractorStyle]
+    ]:
+        """Get/set the current interactor style.
+
+        .. warning::
+
+            Setting an interactor style needs careful control of events handling.
+            See implementation of interactor styles in source code for more details.
```
MatthewFlamm, line 462:
```suggestion
            See :class:`CaptureInteractorStyle` and its implementation as an example.
```

pyvista/plotting/render_window_interactor.py
MatthewFlamm: This is now public and explicitly referenced in this warning.

MatthewFlamm:
We should wait for #7502 to merge this, if it looks good, to simplify the docstring here.  Only the invocation of `pyvista-plot` will change, so this PR is ready for review again.

pyvista/plotting/render_window_interactor.py
```diff
@@ -1528,75 +1553,176 @@ def close(self):
...
+
+    """
+
+
+class InteractorStyleJoystickActor(CaptureInteractorStyle, _vtk.vtkInteractorStyleJoystickActor):
+    """Joystick actor interactor style.
+
+    See Also
+    --------
+    pyvista.plotting.render_window_interactor.RenderWindowInteractor.enable_joystick_actor_style
```
user27182, line 1652: When I search for this in the current docs https://dev.pyvista.org/search.html?q=enable_joystick, I see results for
[pyvista.Plotter.enable_joystick_actor_style](https://dev.pyvista.org/api/plotting/_autosummary/pyvista.Plotter.enable_joystick_actor_style.html#pyvista.Plotter.enable_joystick_actor_style) and [pyvista.RenderWindowInteractor.enable_joystick_actor_style](https://dev.pyvista.org/api/plotting/_autosummary/pyvista.RenderWindowInteractor.enable_joystick_actor_style.html#pyvista.RenderWindowInteractor.enable_joystick_actor_style), so I think the links need to be one of these, e.g.:
```suggestion
    pyvista.RenderWindowInteractor.enable_joystick_actor_style
```

or make it shorter to simply show it as "enable_joystick_actor_style" if desired:

```suggestion
    :class:`~pyvista.RenderWindowInteractor.enable_joystick_actor_style`
```

user27182: Whoops, I think :meth: is needed instead of :class: ...

pyvista/plotting/render_window_interactor.py
```diff
@@ -433,26 +443,61 @@ def key_press_event(self, *args):
...
+        self,
+    ) -> Optional[
+        Union[_vtk.vtkContextInteractorStyle, _vtk.vtkInteractorStyle, CaptureInteractorStyle]
+    ]:
+        """Get/set the current interactor style.
+
+        .. warning::
+
+            Setting an interactor style needs careful control of events handling.
+            See :class:`CaptureInteractorStyle` and its implementation as an example.
```
user27182, line 463:
```suggestion
            See :class:`~plotting.render_window_interactor.CaptureInteractorStyle` and its implementation as an example.
```

pyvista/plotting/render_window_interactor.py
MatthewFlamm: I switched to using :meth:. I also got rid of `~` now that the resolving path is shorter.

MatthewFlamm:
The docs look good to me now with proper linking.  Thanks for the help.  The only thing that looks off is the Returns section of style.

image

I notice this happens in several places in the documentation, e.g.:
https://docs.pyvista.org/api/core/helpers#pyvista.core.utilities.arrays.convert_array

What is the best style here?

user27182:
> The docs look good to me now with proper linking. Thanks for the help. The only thing that looks off is the Returns section of style.
> 
> image
> 
> I notice this happens in several places in the documentation, e.g.: https://docs.pyvista.org/api/core/helpers#pyvista.core.utilities.arrays.convert_array
> 
> What is the best style here?

Looks like sphinx removes the whitespace? Might be better to use `|` instead, like https://docs.pyvista.org/api/core/_autosummary/pyvista.dataobjectfilters.clip#pyvista.DataObjectFilters.clip

pyvista/plotting/render_window_interactor.py
```diff
@@ -108,8 +111,9 @@ def __init__(self, plotter, desired_update_rate=30, light_follow_camera=True, in
         self._MAX_CLICK_DELTA = 40  # squared => ~6 pixels
 
         # Set default style
-        self._style = 'RubberBandPick'
-        self._style_class = None
+        self._style_class: Optional[_vtk.vtkInteractorStyle] = None
```
user27182, line 114: Use PEP 604 syntax
```suggestion
        self._style_class: _vtk.vtkInteractorStyle | None = None
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -108,8 +111,9 @@ def __init__(self, plotter, desired_update_rate=30, light_follow_camera=True, in
...
+        self._style: Literal['Interactor', 'Context'] = 'Interactor'
+        self.style = InteractorStyleRubberBandPick(self)
```
user27182, line 116: I am not familiar with the old API. Was it possible to use a string to set this before? E.g. `'RubberBandPick'` ? Should we allow doing `self.style = 'RubberBandPick'` ?

pyvista/plotting/render_window_interactor.py
```diff
@@ -433,26 +443,58 @@ def key_press_event(self, *args):
...
+        >>> plotter.iren.style
+        A custom interactor style.
+
         """
-        if self._style_class is None:
-            self.update_style()
         return self._style_class
 
+    @style.setter
+    def style(self, style: Optional[Union[_vtk.vtkInteractorStyle, CaptureInteractorStyle]]):
```
user27182, line 493:
```suggestion
    def style(self, style: _vtk.vtkInteractorStyle | CaptureInteractorStyle | None):
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -433,26 +443,58 @@ def key_press_event(self, *args):
...
+        self.interactor.SetInteractorStyle(self.style)
 
     @property
-    def style(self):
-        """Return the current interactor style.
+    def style(
+        self,
+    ) -> Optional[
+        Union[_vtk.vtkContextInteractorStyle, _vtk.vtkInteractorStyle, CaptureInteractorStyle]
+    ]:
```
user27182, line 457:
```suggestion
    ) -> _vtk.vtkContextInteractorStyle | _vtk.vtkInteractorStyle | CaptureInteractorStyle | None:
```

pyvista/plotting/render_window_interactor.py
```diff
@@ -433,26 +443,58 @@ def key_press_event(self, *args):
...
+        """
+        self.interactor.SetInteractorStyle(self.style)
 
     @property
-    def style(self):
-        """Return the current interactor style.
+    def style(
+        self,
+    ) -> Optional[
+        Union[_vtk.vtkContextInteractorStyle, _vtk.vtkInteractorStyle, CaptureInteractorStyle]
```
user27182, line 456: I see the VTK classes have a convention where
- abstract classes use the convention `<name>InteractorStyle`
- concrete classes use the convention `InteractorStyle<name>`

And it looks like this convention is followed for PyVista classes. I find the reversal of names a bit odd... should we adopt a single, unified convention for these names?

user27182 (changes requested):
I am not familiar with the interactor style API but these changes look good. Just a few comments.

pyvista/plotting/render_window_interactor.py
MatthewFlamm: I'll commit these changes but will likely need to clean up things like imports, but maybe ruff can autocorrect it

pyvista/plotting/render_window_interactor.py
MatthewFlamm: It was only possible if the user used private attributes and knew the class naming structure in vtk AND knew to call `update_style` method.  `self._style` previously held the name of the variable part of the vtk class, e.g. `RubberBandPick` for `vtkInteractorStyleRubbetBandPick`.  So this was not part of public API.

I tried to keep the changes limited to the user selected styles.  `self._style` is no longer needed for the user selected styles, so it is now just labeled 'Interactor' in this case. Maybe a better name should be used?  But it is a private attribute after all.  It is still used for keeping track of when the context interactor style is used for chart interaction.  This could be a future PR to simplify this. 

I think setting via string could be handled with #7359.  This PR would enable this in easier manner.

user27182: Ok thanks for the info, it's fine as-is

pyvista/plotting/render_window_interactor.py
MatthewFlamm: What about being explicit like `InteractorStyleCaptureMixin`?

user27182: I like explicit, It was otherwise unclear which ones were concrete classes that I might want to try to use/ set the interactor to

MatthewFlamm:
pre-commit.ci autofix

MatthewFlamm:
Also put in links to the vtk classes for the interactor styles for each wrapped class.

user27182 (approved):
LGTM

PR #7498, opened 2025-05-06
Add `cmcrameri` scientific colormaps

user27182:
### Overview

Add scientific colormaps by Fabio Crameri https://www.fabiocrameri.ch/colourmaps/ which are packaged as `cmcrameri` https://github.com/callumrollo/cmcrameri

These colormaps are perceptually uniform and very good for scientific visualizations. In particular, this package supplements the existing colormaps from `colorcet`, `cmocean`, and `matplotlib` by addition addtional multi-sequential cmaps as well as categorical cmaps, of which there is currently very limited selection.

This PR:
- adds `cmcrameri` to the `pyvista[colormaps]` install
- searches for these cmaps with  `get_cmap_safe`
- adds type hints for these cmaps
- adds the cmaps to the `Named Colormaps` page

Since there are limited colors to use with the badges in the docs, the red theme color (`danger`) is used for the new `cmc` instead of for the `NPU` badge. Instead, the `PU` and `NPU` badges now use icons.


user27182:
This is ready. There is room for improvement in the ordering of the new colormaps, i.e. some should probably be re-ordered to place similar colormaps together (e.g. group blues with other blues).

I am thinking that I will add automatic grouping for all colormaps based on a hue progress and overall lightness, similar to how the `Named Colors` are automatically sorted. But this will be done in a separate PR.

tkoyama010 (approved):

PR #7501, opened 2025-05-07
Fix downloads tests

beroda:
### Overview

While working on https://github.com/pyvista/pyvista/pull/7441, I realized that `pyvista.examples.downloads` tests, which are disabled by default on CI (absence of `--test_downloads` in the command line) are failing locally (in Windows).

This PR is an attempt to fix those tests by including them by default on CI.
I could not find any motivation on the original PR https://github.com/pyvista/pyvista/pull/3218 to exclude these tests by default on the CI.


beroda:
Quite interestingly, I'm also experiencing sporadic seg faults for the `biplane` example. See there https://github.com/pyvista/pyvista/pull/7441/checks#step:9:7134 

Though it might be a vtk issue since the following MRE is flaky:
```python
from pyvista import examples
from vtk import vtkExodusIIReader

filename = examples.download_biplane(load=False)

reader = vtkExodusIIReader()
reader.SetFileName(str(filename))
reader.UpdateInformation()
reader.Update()
```
Might be only related to Windows ...

MatthewFlamm:
Given that the issue is either in vtk or the dataset itself, there isn't much we can do.  We can use a different dataset in use when possible.  Otherwise, we need to skip in windows.  Since we cannot use a different dataset for the download function itself, we should skip it on windows.

user27182:
> Given that the issue is either in vtk or the dataset itself, there isn't much we can do.  We can use a different dataset in use when possible.  Otherwise, we need to skip in windows.  Since we cannot use a different dataset for the download function itself, we should skip it on windows.

I can update the MultiBlock docs to use a different dataset. All datasets are also loaded by default when building the docs (for generating the Dataset Gallery in `make_tables.py`), so some logic may be needed there to skip loading this dataset for windows builds

beroda:
> Given that the issue is either in vtk or the dataset itself, there isn't much we can do. We can use a different dataset in use when possible. Otherwise, we need to skip in windows. Since we cannot use a different dataset for the download function itself, we should skip it on windows.

Actually, I managed to resolve the flakiness by opening the original file with paraview and saving it back without some data (see screenshot and attached file).

We could replace the file in `pyvista/vtk-data` without having to refactor all examples as @user27182 suggested.

image
[test.zip](https://github.com/user-attachments/files/20104334/test.zip)

beroda:
The issue might be caused by the `read_exodus` function, which is not used in  https://github.com/pyvista/pyvista/pull/7048 that introduced the biplane example.

user27182:
It sounds like the root cause is then still not known?

Based on the MRE from https://github.com/pyvista/pyvista/pull/7501#issuecomment-2860358181 it seems to me that both PyVista's exodus reader class and `read_exodus` should be flaky (and possibly Paraview too?). But if it does open in Paraview and saving a new file fixes this then we should just replace the dataset in vtk-data. 

Or, if only `read_exodus` is the issue then we should properly deprecate this and fix the logic to always use the new Exodus reader.

Might be good to do both (replace the file and deprecate `read_exodus`)

tests/examples/test_download_files.py
```diff
@@ -21,7 +21,6 @@
 if 'TEST_DOWNLOADS' in os.environ:
     warnings.warn('"TEST_DOWNLOADS" has been deprecated. Use `pytest --test_downloads`')
```
user27182, line 22: Should this also be removed?
```suggestion
```

And there is mention of ``--test_downloads`` in the module docstring which should be removed.

beroda:
> Sorry for not making this clearer, my previous comments were more aimed at trying to get this PR merged sooner.  Missing testing to a large module is not a good state and will allow more errors to creep in. It would be good to address this problem, but in a separate issue/PR in my opinion.

Agree. Will skip for Windows, fix coverage and we're good for review.

MatthewFlamm:
There is a good reason to not have this run by pytest by default.  When I run on my local machine, I don't want to have to download all the datasets.  The development environment with pre-commit and so on already takes up an enormous amount of storage.

Would it be better to add the `--test-download` flag into the CI options?

beroda:
> There is a good reason to not have this run by pytest by default. When I run on my local machine, I don't want to have to download all the datasets. The development environment with pre-commit and so on already takes up an enormous amount of storage.
> 
> Would it be better to add the `--test-download` flag into the CI options?

Agree. Done in https://github.com/pyvista/pyvista/pull/7501/commits/cf3d66d12894dbd0c3cbb7411736d66e326d1deb

mattfalcone1997:
Some perspective on the biplane example and how blocks and sets work in the `ExodusIIReader`. The only set/block arrays that are explicitly enabled by default is `element_blocks` which is done through `self.enable_all_cell_arrays()`. From a VTK perspective, this is just a call to `vtkExodusIIReader.SetElementResultArrayStatus`. The nodal arrays are also activated through `self.enable_all_point_arrays` but these are not associated with a block. 

It should be noted that these methods don't explicitly enable the element block but I have not needed to do this previously. This could be done using `self.element_blocks.enable_all()`. Similarly, the other blocks could be explicitly disabled e.g. using `self.face_blocks.disable_all()`. Not sure if this is even the cause of the issue, but it could be tested.

mattfalcone1997:
Something else that is interesting to note is that in the test suite where `download_biplane` failed, `test_exodus_blocks` passed. [This test calls `download_biplane`, albeit with `load=False`.](https://github.com/pyvista/pyvista/blob/bc5108f63b2540a263c2867626521fa341deda2a/tests/core/test_reader.py#L1580)

MatthewFlamm:
> Some perspective on the biplane example and how blocks and sets work in the `ExodusIIReader`. The only set/block arrays that are explicitly enabled by default is `element_blocks` which is done through `self.enable_all_cell_arrays()`. From a VTK perspective, this is just a call to `vtkExodusIIReader.SetElementResultArrayStatus`. The nodal arrays are also activated through `self.enable_all_point_arrays` but these are not associated with a block.
> 
> It should be noted that these methods don't explicitly enable the element block but I have not needed to do this previously. This could be done using `self.element_blocks.enable_all()`. Similarly, the other blocks could be explicitly disabled e.g. using `self.face_blocks.disable_all()`. Not sure if this is even the cause of the issue, but it could be tested.

My comment here is separate from this PR, but PyVista has precedent to enable all data arrays even if the vtk readers do not have that as the default.  I would probably look first to what is the standard reading of data for that kind of dataset, and if not clear, it makes sense to me to default to read in all data.

mattfalcone1997:
> > Some perspective on the biplane example and how blocks and sets work in the `ExodusIIReader`. The only set/block arrays that are explicitly enabled by default is `element_blocks` which is done through `self.enable_all_cell_arrays()`. From a VTK perspective, this is just a call to `vtkExodusIIReader.SetElementResultArrayStatus`. The nodal arrays are also activated through `self.enable_all_point_arrays` but these are not associated with a block.
> > It should be noted that these methods don't explicitly enable the element block but I have not needed to do this previously. This could be done using `self.element_blocks.enable_all()`. Similarly, the other blocks could be explicitly disabled e.g. using `self.face_blocks.disable_all()`. Not sure if this is even the cause of the issue, but it could be tested.
> 
> My comment here is separate from this PR, but PyVista has precedent to enable all data arrays even if the vtk readers do not have that as the default. I would probably look first to what is the standard reading of data for that kind of dataset, and if not clear, it makes sense to me to default to read in all data.

The main reason for only enabling the element block and its results, is that that is what Paraview does. Also, the other blocks and sets are comparatively rare. Side sets are the next most common but activating all of these would hide the element blocks so it only really makes sense to show one or the other.

pyvista/examples/downloads.py
```diff
@@ -5777,19 +5777,20 @@ def download_can(partial=False, load=True):  # pragma: no cover
...
-    if pyvista.vtk_version_info > (9, 1):
-        msg = (
-            'This example file is deprecated for VTK v9.2.0 and newer. '
-            'Use `download_can_crushed_hdf` instead.'
-        )
-        raise VTKVersionError(msg)
+def _dataset_can_files_func():
     can_0 = _SingleFileDownloadableDatasetLoader('hdf/can_0.hdf')
     can_1 = _SingleFileDownloadableDatasetLoader('hdf/can_1.hdf')
     can_2 = _SingleFileDownloadableDatasetLoader('hdf/can_2.hdf')
```
beroda, line 5796: Restoring to the original implementation where both partial and full examples were deprecated.
See change introduced here https://github.com/pyvista/pyvista/pull/5787/files?diff=split&w=0#diff-e0e0a0466d5d721fff8649e50a23fcf0a068ba6979c2692c379822ca56038038

user27182: This change seems to be causing failures. I think something like this is what we need:
``` python
def _dataset_can_partial_read_func(path: str):
    if pyvista.vtk_version_info > (9, 1):
        msg = (
            'This example file is deprecated for VTK v9.2.0 and newer. '
            'Use `download_can_crushed_hdf` instead.'
        )
        raise VTKVersionError(msg)
    return pyvista.read(path)


def _dataset_can_files_func():
    can_0 = __can_partial
    can_1 = _SingleFileDownloadableDatasetLoader('hdf/can_1.hdf')
    can_2 = _SingleFileDownloadableDatasetLoader('hdf/can_2.hdf')
    return can_0, can_1, can_2


_dataset_can = _MultiFileDownloadableDatasetLoader(
    files_func=_dataset_can_files_func,
    load_func=_load_and_merge,
)
__can_partial = _SingleFileDownloadableDatasetLoader(
    'hdf/can_0.hdf', read_func=_dataset_can_partial_read_func
)
```

beroda: The failing tests concern `vtk == 9.0.3` since the `vtkHDFReader` was introduced in `vtk == 9.1` (see the comment in the original implementation https://github.com/pyvista/pyvista/pull/2352#issuecomment-1072865042).

I'll update accordingly.

user27182 (commented):
LGTM except for minor formatting. We should also update the VTK Master workflow.

.github/workflows/testing-and-deployment.yml
```diff
@@ -264,7 +264,7 @@ jobs:
             Examples-1-
 
       - name: Test Core API
-        run: python -m pytest -v --ignore=tests/plotting
+        run: python -m pytest -v --ignore=tests/plotting --test_downloads
```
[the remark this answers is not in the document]
beroda: Done in 7e5c32fcbb13b84ba33b7b3f0e0ed64dd8facb4e

user27182 (approved):
Thanks @beroda !

tkoyama010 (dismissed):

banesullivan (dismissed):

beroda:
The GH rate limit issue already discussed [here](https://github.com/pyvista/pyvista/pull/7525#pullrequestreview-2846278906) is also happening here.
Weird that we did not see it in previous workflow runs ... 

I plan to add the `@flaky` marker to the whole `tests/examples/test_download_files.py` module to solve that.

tests/examples/test_download_files.py
```diff
@@ -10,18 +10,24 @@
...
 import pyvista as pv
 from pyvista import examples
 
 if 'TEST_DOWNLOADS' in os.environ:
     warnings.warn('"TEST_DOWNLOADS" has been deprecated. Use `pytest --test_downloads`')
 
-pytestmark = pytest.mark.needs_download
+pytestmark = [
+    pytest.mark.needs_download,
+    pytest.mark.flaky(retries=5, delay=1, only_on=[HTTPError]),  # due to Github rate limit
```
user27182, line 29: Retrying here seems more like a workaround than a solution, since this problem is also causing docstring test and docs build failures. This workaround is OK if it means we only need to retry those two jobs, and no longer also need to retry the unit tests. But it would be nice to somehow avoid hitting the rate limit to begin with.

beroda: I agree this is a quick workaround.
To provide a more robust download, we might need to use a custom `pooch` downloader that retries sending the request if reaching the rate limit.

We could get inspiration from the existing pooch [`HTTPDownloader`](https://www.fatiando.org/pooch/latest/api/generated/pooch.HTTPDownloader.html)

user27182: The issue with rate limiting seems to have gone away on its own... do we still want to keep this change? I am still unsure if adding more retries to address a rate limiting issue is a good idea, since the extra tries could be making the rate limiting issue worse....

I am partial towards not including this change. If rate limiting is truly an issue, I think it should be addressed more directly based on these [Tips for avoiding the API rate limit](https://github.com/orgs/community/discussions/77255). For retries specifically, it recommends setting retry times based on the header, and not necessarily to a static value like `delay=1` set here. And this change should be more comprehensive to also address the docstring tests and docs build workflows.

MatthewFlamm: See https://github.com/pyvista/pyvista/pull/6186 as an alternative solution.  There we package up vtk-data as an artifact and reuse it in the workflows as a local vtk repo. See https://docs.pyvista.org/extras/vtk_data.  This would remove the vast majority of API calls.  It could be extended to provide data for the documentation too.

I think the issues with windows and/or macos were related to the environment variable setting for vtk-data, but my memory is hazy.  I'm not sure how complex that would be to address.

user27182: +1 for fixing the http errors with caching. Though, it seems like the caching mechanism still needs work.

To move this PR forward, I think fixing the caching should be addressed in a separate PR, and in the interim, we can remove the --test_downloads flag and omit these from CI for now. Otherwise, the changes from this PR are still good, and the downloads test have been fixed, even if we won't run them yet.

edabor: Caching needs rework indeed.

However, if it is fine for you, I plan to enable the `vtk-data` cache for unit tests on `MacOS` and `Linux` to harmonize with `Windows`. I don't want the merge of this PR to cause later issues with rate limit on subsequent PRs.

I agree that we can fix the documentation cache (and potentially other jobs) in a later PR.

user27182: Ok! I mostly didn't want the caching rework to necessarily hold up this PR if it has relatively large scope. But if there are easy fixes for harmonizing windows/linux/macos caching then that's fine with me. 

So then if the unit test caching does work (after harmonizing), the `--test_downloads` flag will be kept after all, right?

edabor: Yes, the flag will be kept if everything works fine.

.github/workflows/testing-and-deployment.yml
```diff
@@ -31,11 +31,45 @@ env:
...
+          path: ${{ env.CACHE_FOLDER_NAME }}
+
+      - name: Create the example data cache
+        if: steps.restore-cache-vtk-data.outputs.cache-hit != 'true'
+        id: save-cache-vtk-data
+        uses: actions/cache/save@v4
+        with:
+          path: ${{ env.CACHE_FOLDER_NAME }}
+          key: vtk-data-${{ env.VTK_DATA_COMMIT }}
+          enableCrossOsArchive: true
```
edabor: This is inspired by https://github.com/pyvista/pyvista/pull/6186

The `pyvista/vtk-data` repo is cached using the commit hash of the repo as cache key, meaning that cache renewal will only occur if the repo changes. This is done in order to reduce CI footprint.

The cache is then restored in all unit tests to prevent reaching the GH rate limit when downloading examples.

.github/workflows/testing-and-deployment.yml
```diff
@@ -31,11 +31,45 @@ env:
     !startsWith(github.ref, 'refs/heads/release/') &&
     !startsWith(github.ref, 'refs/heads/main')
     }}
+  CACHE_FOLDER_NAME: vtk-data
+  PYVISTA_USERDATA_PATH: ${{ github.workspace }}/vtk-data
```
user27182, line 35:
```suggestion
  PYVISTA_USERDATA_PATH: ${{ github.workspace }}/${{ env.CACHE_FOLDER_NAME }}
```

.github/workflows/testing-and-deployment.yml
```diff
@@ -295,8 +342,10 @@ jobs:
   windows:
     name: Windows Unit Testing
     runs-on: windows-latest
+    needs: cache-vtk-data
     env:
       CI_WINDOWS: true
+      PYVISTA_USERDATA_PATH: ${{ github.workspace }}\vtk-data
```
user27182: This is redefining the env var to use `\` instead of `/`. Is it possible to move this change into `downloads.py` so that the code normalizes the path instead of needing to redefine it here?  Maybe just try removing it and see if it works as-is?
```suggestion
```

.github/workflows/testing-and-deployment.yml
```diff
@@ -31,11 +31,45 @@ env:
...
     !startsWith(github.ref, 'refs/heads/main')
     }}
+  CACHE_FOLDER_NAME: vtk-data
+  PYVISTA_USERDATA_PATH: ${{ github.workspace }}/vtk-data
 
 permissions:
   id-token: none
 
 jobs:
+  cache-vtk-data:
```
user27182, line 52: Nit: The other job names use title case. It might also be good to explicitly mention that the _repository_ is being cached
```suggestion
  cache-vtk-data:
    name: Cache vtk-data Repo 
```

tests/plotting/test_conftest.py
```diff
@@ -55,7 +55,7 @@ def test_no_plotting():
     mock.return_value = support_plotting
 
     p = pytester.makepyfile(tests)
-    results = pytester.runpytest(p)
+    results = pytester.runpytest(p, '-p no:pytest-retry')
```
user27182: Is this leftover from when `pytest-retry` was added a dependency? Is it still needed?
```suggestion
    results = pytester.runpytest(p)
```

tests/test_conftest.py
```diff
@@ -74,13 +74,16 @@ def results_parser(monkeypatch: pytest.MonkeyPatch):
...
     pytester: pytest.Pytester,
+    monkeypatch: pytest.MonkeyPatch,
 ):
     with (pytestconfig.rootpath / 'pyproject.toml').open('r') as file:
         toml = pytester.makepyprojecttoml(file.read())
 
     with (pytestconfig.rootpath / 'tests/conftest.py').open('r') as file:
         conftest = pytester.makeconftest(file.read())
 
+    monkeypatch.setenv('PYTEST_ADDOPTS', '-p no:pytest-retry', prepend=' ')
```
user27182: Same here, is pytest-retry still needed?

edabor: Done in d2c1f6b8d87c8b3025163e770aca6825fba52020

tests/plotting/test_conftest.py
edabor: Done in d2c1f6b8d87c8b3025163e770aca6825fba52020

.github/workflows/testing-and-deployment.yml
edabor: Done in d2c1f6b8d87c8b3025163e770aca6825fba52020

.github/workflows/testing-and-deployment.yml
edabor: Done in d2c1f6b8d87c8b3025163e770aca6825fba52020 It was because the cache path was incorrect before centralizing to a single job.

.github/workflows/testing-and-deployment.yml
edabor: Done in d2c1f6b8d87c8b3025163e770aca6825fba52020

user27182: I don't think the GH Actions like this (self referencing env inside env definitions)

.github/workflows/testing-and-deployment.yml
```diff
@@ -31,11 +31,46 @@ env:
     !startsWith(github.ref, 'refs/heads/release/') &&
     !startsWith(github.ref, 'refs/heads/main')
     }}
+  CACHE_FOLDER_NAME: vtk-data
+  PYVISTA_USERDATA_PATH: ${{ github.workspace }}/${{ env.CACHE_FOLDER_NAME }}
```
user27182:
```suggestion
  PYVISTA_USERDATA_PATH: ${{ github.workspace }}/vtk-data
```

user27182 (dismissed):
LGTM, thanks!

.github/workflows/testing-and-deployment.yml
```diff
@@ -72,6 +108,14 @@ jobs:
           python-version: ${{ matrix.config.python-version }}
           cache: ${{ env.USE_CACHE == 'true' && 'pip' || '' }}
 
+      - name: Restore example data cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ env.CACHE_FOLDER_NAME }}
+          key: ${{ needs.cache-vtk-data.outputs.key }}
+          enableCrossOsArchive: true
+          fail-on-cache-miss: true
```
user27182, line 117: I retried the unit test workflow for #7620 [here](https://github.com/pyvista/pyvista/actions/runs/15690024798?pr=7620) that previously had green CI a few days ago,, but for some reason half the jobs failed from a cache miss when I retried all of them... not sure why that would happen, is there is a way to prevent this?

edabor: Seems to be related to https://github.com/actions/cache/issues/1621

PR #7502, opened 2025-05-08
Require explicit calls to `plot` or `show` methods with the `plot-directive`

### LINKED ISSUE #6320: `plot_directive` generates build image even if `plot()` or `show()` is not called

user27182:
Fix #6320

Currently, the plot directive always generates a figure when there is a `Plotter` instance available. However, creating a `Plotter` doesn't necessarily mean that we want to generate a figure, as it's often used in examples just for setting or getting properties for `Plotter`, `Actor`, `Camera`, `Mapper` classes, etc.

This PR adds a condition to the plot-directive logic to only generate a figure when an explicit call is made to `plot` or `show`. Related methods like `plot_cell`, `plot_datasets`, `plot_normals`  are also supported (i.e. `plot_<anything>`), as is the use of `plot=True` keywords.

This PR also refactors the splitting logic to use the same regex so that splitting now properly occurs for methods like `plot_normals`. Previously, calling `plot_normals()` twice would result in the docs showing as `code1 code2 fig1 fig2`, whereas now it's properly ordered as `code1 fig1 code2 fig2`.

Note: I reviewed every image manually and looked at each docstring to see if there was an intention or need to display a figure. In cases where it seemed helpful to have a figure, I added an explicit call to `show`, otherwise the image is removed.


user27182:
The diff on GitHub isn't great for showing the images removed by this PR. For proper review, it may be best to checkout the branch locally and look at the diff with an IDE. Or, have a look at the build artifact from [8b4649b](https://github.com/pyvista/pyvista/pull/7502/commits/8b4649b3873abbd71e8fc91f6f1355c89cea184a):
https://github.com/pyvista/pyvista/actions/runs/14921013418/artifacts/3091579260... almost all of the `from_cache` images have been removed.

MatthewFlamm (approved):
The changes look good to me.  I did not have time to check the images changes.

PR #7504, opened 2025-05-08
Check `sys.meta_path` in `DisableVtkSnakeCase.__getattribute__`

### LINKED ISSUE #7503: Exception ignored in <function PolyData.__del__ ...>

Seeing this after my closed source project unit tests complete now in `pyvista==0.45.0`:

```
==================== 397 passed, 6 skipped, 9 warnings, 1 rerun in 16.96s ====================Exception ignored in: <function PolyData.__del__ at 0x77733a7f2480>
  File "/home/user/.venv312/lib/python3.12/site-packages/pyvista/core/pointset.py", line 1712, in __del__
  File "/home/user/.venv312/lib/python3.12/site-packages/pyvista/core/_vtk_core.py", line 626, in __getattribute__
ImportError: sys.meta_path is None, Python is likely shutting down
Exception ignored in: <function PolyData.__del__ at 0x77733a7f2480>
  File "/home/user/.venv312/lib/python3.12/site-packages/pyvista/core/pointset.py", line 1712, in __del__
  File "/home/user/.venv312/lib/python3.12/site-packages/pyvista/core/_vtk_core.py", line 626, in __getattribute__
ImportError: sys.meta_path is None, Python is likely shutting down
Exception ignored in: <function PolyData.__del__ at 0x77733a7f2480>
  File "/home/user/.venv312/lib/python3.12/site-packages/pyvista/core/pointset.py", line 1712, in __del__
  File "/home/user/.venv312/lib/python3.12/site-packages/pyvista/core/_vtk_core.py", line 626, in __getattribute__
ImportError: sys.meta_path is None, Python is likely shutting down
Exception ignored in: <function PolyData.__del__ at 0x77733a7f2480>
  File "/home/user/.venv312/lib/python3.12/site-packages/pyvista/core/pointset.py", line 1712, in __del__
  File "/home/user/.venv312/lib/python3.12/site-packages/pyvista/core/_vtk_core.py", line 626, in __getattribute__
ImportError: sys.meta_path is None, Python is likely shutting down
```

Can't generate a MWE just yet.

### System Information

```shell
  Date: Thu May 08 10:52:21 2025 MDT

                  OS : Linux (Ubuntu 24.04)
              CPU(s) : 32
             Machine : x86_64
        Architecture : 64bit
                 RAM : 188.5 GiB
         Environment : Python
         File system : ext4
        GPU Renderer : NVIDIA GeForce RTX 4090/PCIe/SSE2
         GPU Version : 4.5.0 NVIDIA 550.120
    MathText Support : True

  Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0]

             pyvista : 0.45.0
                 vtk : 9.4.2
               numpy : 2.2.5
          matplotlib : 3.10.1
              scooby : 0.10.1
               pooch : 1.8.2
              pillow : 11.2.1
             imageio : 2.35.1
           pyvistaqt : 0.11.1
               PyQt5 : 5.15.10
             IPython : 9.1.0
            colorcet : 3.1.0
             cmocean : 4.0.3
          ipywidgets : 8.1.3
               scipy : 1.14.1
                tqdm : 4.66.4
              meshio : 5.3.5
          jupyterlab : 4.2.5
      pytest_pyvista : 0.1.8
               trame : 3.6.5
        trame_client : 3.1.0
        trame_server : 3.0.1
           trame_vtk : 2.8.10
       trame_vuetify : 2.7.1
jupyter_server_proxy : 4.4.0
        nest_asyncio : 1.6.0
```

user27182:
### Overview

Fix #7503
Dynamic imports in dunder methods like `__getattribute__` can cause issues when Python is trying to shut down. This PR fixes this.

### DIFF
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -10,7 +10,6 @@ the entire library.
 from __future__ import annotations
 
 import contextlib
-import sys
 from typing import NamedTuple
 import warnings
 
@@ -632,12 +631,15 @@ class DisableVtkSnakeCase:
             # We normally can import pyvista dynamically, but this can't be done in
             # some cases, e.g. when the python interpreter is shutting down,
             # in which case the meta_path is None and we set the state directly
-            if sys.meta_path is not None:
-                import pyvista as pv
+            # if sys.meta_path is not None:
+            #     import pyvista as pv
+            #
+            #     state = pv._VTK_SNAKE_CASE_STATE
+            # else:
+            #     state = 'error'
+            import pyvista as pv
 
-                state = pv._VTK_SNAKE_CASE_STATE
-            else:
-                state = 'error'
+            state = pv._VTK_SNAKE_CASE_STATE
 
             if state != 'allow':
                 if (
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -628,16 +628,14 @@ class DisableVtkSnakeCase:
         if vtk_version_info >= (9, 4):
             # Raise error if accessing attributes from VTK's pythonic snake_case API
 
-            # Get the current value of the snake case state variable
             if sys.meta_path is None:  # pragma: no cover
                 # Python is likely shutting down, so we avoid any dynamic imports
-                # and simply set a default value
-                state = 'error'  # type: ignore[unreachable]
-            else:  # pragma: no branch
-                import pyvista as pv
+                # and simply return None
+                return None
 
-                state = pv._VTK_SNAKE_CASE_STATE
+            import pyvista as pv
 
+            state = pv._VTK_SNAKE_CASE_STATE
             if state != 'allow':
                 if (
                     attr not in ['__class__', '__init__']
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -631,7 +631,7 @@ class DisableVtkSnakeCase:
             if sys.meta_path is None:  # pragma: no cover
                 # Python is likely shutting down, so we avoid any dynamic imports
                 # and simply return None
-                return None
+                return None  # type: ignore[unreachable]
 
             import pyvista as pv
 
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -632,8 +632,8 @@ class DisableVtkSnakeCase:
             if sys.meta_path is None:
                 # Python is likely shutting down, so we avoid any dynamic imports
                 # and simply set a default value
-                state = 'error'
-            else:
+                state = 'error'  # pragma: no cover
+            else:  # pragma: no branch
                 import pyvista as pv
 
                 state = pv._VTK_SNAKE_CASE_STATE
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -10,6 +10,7 @@ the entire library.
 from __future__ import annotations
 
 import contextlib
+import sys
 from typing import NamedTuple
 import warnings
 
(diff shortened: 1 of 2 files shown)


pyvista/core/_vtk_core.py
```diff
@@ -626,9 +627,17 @@ class DisableVtkSnakeCase:
...
         if vtk_version_info >= (9, 4):
             # Raise error if accessing attributes from VTK's pythonic snake_case API
-            import pyvista as pv
 
-            state = pv._VTK_SNAKE_CASE_STATE
+            # Get the current value of the snake case state variable
+            if sys.meta_path is None:  # pragma: no cover
+                # Python is likely shutting down, so we avoid any dynamic imports
+                # and simply set a default value
+                state = 'error'  # type: ignore[unreachable]
```
MatthewFlamm, line 635: In this case, we don't `import pyvista as pv`, but then later in the function we might try to access the object `pv`.  I am not really clear on the sequence of events that leads to this branch.  But it seems the current logic could lead to cases where we get a `NameError` that will also be confusing.  Some thoughts:

- If we set the state to `allow` it will then pass the vtk attribute.  This might be okay?
- Or we short circuit and `return None` here?

MatthewFlamm: I'm not sure if the sequence I'm proposing is reacahable in practice since `pv.PolyData().this_var_is_from_vtk` might not follow the same shutdown logic that leads to the error.  But it is worth considering to make sure there aren't other confusing errors that pop up in some other edge case.

user27182: Yes you're right later we access pv.PyVistaAttributeError for the 'error' state, so we should set it to something else. 

The issue I think is fundamentally from the dynamic import, which is needed here to avoid a circular import because the class is defined in _vtk_core. But we can alternatively try moving this class elsewhere to avoid the local import altogether ?

MatthewFlamm: Is there a downside to returning `None` here?

user27182: > Is there a downside to returning `None` here?

This is probably the best option.

akaszynski (approved):
Perfect. Thanks!

akaszynski:
I'd appreciate if we target a v0.45.1 patch at the end of May and added this (and a few other patches) it.

MatthewFlamm (approved):

user27182:
I'm not sure if returning `None` was the right solution, we're seeing an odd attribute error here during object deletion:
https://github.com/pyvista/pyvista/actions/runs/14949342452/job/41996755160#step:10:199
``` python
  File "/opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/pyvista/core/pointset.py", line 1717, in __del__
AttributeError: 'PolyData' object has no attribute '_obbTree'
Exception ignored in: <function PolyData.__del__ at 0x7f5bab3bae80>
  File "/opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/pyvista/core/pointset.py", line 1717, in __del__
AttributeError: 'PolyData' object has no attribute '_obbTree'
Exception ignored in: <function PolyData.__del__ at 0x7f5bab3bae80>
  File "/opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/pyvista/core/pointset.py", line 1717, in __del__
AttributeError: 'PolyData' object has no attribute '_obbTree'
Exception ignored in: <function PolyData.__del__ at 0x7f5bab3bae80>
  File "/opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/pyvista/core/pointset.py", line 1717, in __del__
AttributeError: 'PolyData' object has no attribute '_obbTree'
Exception ignored in: <function PolyData.__del__ at 0x7f5bab3bae80>
```

This is not the root cause of the failure though (the error is related to VTK version), and so this misleading error is similar to what's reported it #7503. Plus, the code is supposed to execute only if it does have the attribute:
https://github.com/pyvista/pyvista/blob/d4f136d790eeb95a9566982edd7a301ce091f67a/pyvista/core/pointset.py#L1716-L1717

So I think Python is shutting down here, calling `__del__` on a mesh, and the `hasattr` check succeeds, but then `__getattribute__` is called, and `None` is then returned because of the changes from this PR...

Issue #7508, opened 2025-05-09
PyVista 0.45.1 patch release


user27182:
### Describe what maintenance you would like added.

This issue is to track PRs for possibly cherry-picking as part of a patch release, initially suggested by @akaszynski here https://github.com/pyvista/pyvista/pull/7504#issuecomment-2867237801 

Fixes directly related to the 0.45.0 release
- #7446 
- #7455
- #7456
- #7475 
- #7477
- #7483
- #7504

Some more general fixes we might want to consider including:
- #7340
- #7458
- #7476
- #7479
- #7490
- #7492
- #7499

VTK 9.5 will be released soon, may be good to also include:
- #7433 

On the fence, may be more of a feature?
- #7452
- #7459

The above are just ideas/ suggestions. More PRs likely means more work for whoever is issuing the release so I'll leave it to others to chime in on what's important to include here.

tkoyama010:
:rocket: https://github.com/pyvista/pyvista/releases/tag/v0.45.1

user27182:
> 🚀 https://github.com/pyvista/pyvista/releases/tag/v0.45.1

Wow, so fast, thank you! 👏

I'm not familiar with the patch release process, is CI run against the changes or no? Some of the PRs were strictly related to the docs, does this get updated too?

tkoyama010:
I tried it but it is failing. If you can make PR to [release/v0.45](https://github.com/pyvista/pyvista/tree/release/0.45) branch. I can  release it 🙇 
https://github.com/pyvista/pyvista/actions/runs/14941331459/job/41978809570

user27182:
> I tried it but it is failing. If you can make PR to [release/v0.45](https://github.com/pyvista/pyvista/tree/release/0.45) branch. I can  release it 🙇 
> https://github.com/pyvista/pyvista/actions/runs/14941331459/job/41978809570

#7510 includes the fix for the docs. 

Separately, we may also want to consider #7509 for a patch

tkoyama010:
Thanks! I will release 0.45.2 after #7509 is merged.

user27182:
#7512 should also be patched into 0.45

user27182:
Thanks @tkoyama010 !

PR #7510, opened 2025-05-10
Fix docs reference in release `0.45.1`

user27182:
### Overview

See https://github.com/pyvista/pyvista/issues/7508#issuecomment-2868616421


tkoyama010 (approved):
Thanks!

PR #7511, opened 2025-05-11
Add lightness and color difference image fields to Named Colormaps

user27182:
### Overview

Add reference figures to better communicate the perceptual uniformity of the colormaps. 

1. Add lightness figures, similar to the ones shown in Matplotlib docs https://matplotlib.org/stable/users/explain/colors/colormaps.html#lightness-of-matplotlib-colormaps
2. Add cumulative delta E figures, similar to the ones shown in Figure 3c and 3d from https://www.nature.com/articles/s41467-020-19160-7

From these figures, it is possible to objectively compute if a colormap is perceptually uniform or not. Therefore, this PR also removes the manual and subjective `PU` labels, and instead applies the label consistently based on a computation.

E.g., before this PR, the `gouldian` colormap was labeled (manually) as perceptually uniform simply because it's from `colorcet`. But now with this PR, it is no longer considered perceptually uniform, and we can see from the cumulative delta E plot that there a small kink in the plot (not quite a straight line).

#### Before this PR
before

#### After this PR
after


user27182:
after

NIT: The swatch has a different aspect ratio than the other images, will push an update later to fix this.

tkoyama010 (approved):
LGTM

PR #7512, opened 2025-05-11
Move check for `sys.meta_path` in `DisableVtkSnakeCase`

user27182:
### Overview

The fix in #7504 isn't complete since side-effect errors are still being generated when python is shutting down, e.g. https://github.com/pyvista/pyvista/pull/7504#issuecomment-2869215934. This PR allows returning `super().__getattribute__` instead of `None` when `sys.meta_path` is `None`.


tkoyama010 (approved):
LGTM

PR #7522, opened 2025-05-15
Update imports and pyvista namespace in `make_tables.py`

user27182:
### Overview

The `StrEnum` definition currently splits the imports into two, causing the import sorting to also be split. This PR moves this definition so all imports are grouped and sorted together. `pyvista` and `pyvista as pv` are also both imported, so this PR removes the `pyvista` import.


tkoyama010 (approved):
LGTM

PR #7537, opened 2025-05-21
Add vtkkhdf writer

user27182:
### Overview

Simplified version of #7396 to troubleshoot seg faults. @krishanbhasin-px this seems to be working for me locally, might be helpful in case you're still interested in working through #7396

### DIFF
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -482,6 +482,10 @@ from vtkmodules.vtkIOGeometry import vtkIVWriter as vtkIVWriter
 from vtkmodules.vtkIOGeometry import vtkOBJWriter as vtkOBJWriter
 from vtkmodules.vtkIOGeometry import vtkProStarReader as vtkProStarReader
 from vtkmodules.vtkIOGeometry import vtkSTLWriter as vtkSTLWriter
+
+with contextlib.suppress(ImportError):  # Introduced VTK v9.4.0
+    from vtkmodules.vtkIOHDF import vtkHDFWriter as vtkHDFWriter
+
 from vtkmodules.vtkIOInfovis import vtkDelimitedTextReader as vtkDelimitedTextReader
 from vtkmodules.vtkIOLegacy import vtkDataReader as vtkDataReader
 from vtkmodules.vtkIOLegacy import vtkDataSetReader as vtkDataSetReader
--- a/pyvista/core/composite.py
+++ b/pyvista/core/composite.py
@@ -132,6 +132,8 @@ class MultiBlock(
     plot = pyvista._plot.plot
 
     _WRITERS = dict.fromkeys(['.vtm', '.vtmb'], _vtk.vtkXMLMultiBlockDataWriter)
+    if _vtk.vtk_version_info >= (9, 4):
+        _WRITERS.update({'.vtkhdf': _vtk.vtkHDFWriter})
 
     def __init__(self: MultiBlock, *args, **kwargs) -> None:
         """Initialize multi block."""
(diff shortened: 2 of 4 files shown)


user27182:
Still getting seg faults... might be specific to CI environment

krishanbhasin-px:
> Still getting seg faults... might be specific to CI environment

I was hoping that, but I think I got them last time I ran the tests locally I got the segfaults too. And there _was_ a period where it was all running cleanly on CI and my local, but sadly I didn't keep track of which commit from `main` I was branching off from.

I just pulled `main` down and reran the tests and am now getting a "proper" test failure instead of segfault! I'll find time to investigate later and see if I can push a fix

user27182:
> Still getting seg faults... might be specific to CI environment

> I just pulled this branch and ran locally and can confirm that I _do_ still get segfaults with it on my Macbook (M3).

Thanks for checking/confirming. The seg fault appears to be happening with
```
tests/core/test_composite.py::test_multi_block_io[True-.vtkhdf-True] 
```

But the new tests from this PR are all passing:

tests


So it must be an issue specific to `test_multi_block_io`. Just did a bit of digging and I have found that that `None` blocks are the cause of the seg faults. Adding `clean` fixes the seg faults:

https://github.com/pyvista/pyvista/blob/5b684890cbed4d75e090f05f33c0b379024fcc24/tests/core/test_composite.py#L440-L441
``` diff
multi = multiblock_all_with_nested_and_none 
+ multi.clean()
```

But the tests still fail, e.g.

``` sh
  #009: /opt/glr/vtk/vtk-ci-ext/0/ThirdParty/hdf5/vtkhdf5/src/H5Gloc.c line 396 in H5G__loc_find_cb(): object 'NumberOfPoints' doesn't exist
    major: Symbol table
    minor: Object not found
2025-05-21 08:51:39.256 (   3.331s) [          689403]    vtkHDFUtilities.cxx:772    ERR| (nullptr): Cannot open NumberOfPoints

```
The difference between the new tests from this PR and that one is that `test_multi_block_io` uses a nested `MultiBlock` fixture with `None` blocks.
And so it appears likely that there are (at least) two bugs with `vtkHDFWriter` with `MultiBlock` meshes:

- `None` blocks are not supported (seg fault)
- Nested `MultiBlock` meshes are not supported.

So, for moving https://github.com/pyvista/pyvista/pull/7396 forward, I think we either need to completely drop support for saving `MultiBlock` as `.vtkhdf`, OR provide limited support for simple cases. This means we should raise an error when saving nested multiblocks (i.e. check `if MultiBlock.is_nested == True`), or if there are `None` blocks, i.e. `if any(block is None for block in MultiBlock.recursive_iterator()`. We could probably add these checks here:

https://github.com/pyvista/pyvista/blob/5b684890cbed4d75e090f05f33c0b379024fcc24/pyvista/core/dataobject.py#L221-L225

user27182:
> So, for moving #7396 forward, I think we either need to completely drop support for saving `MultiBlock` as `.vtkhdf`, OR provide limited support for simple cases.

I just checked with `vtk==9.5.0.rc1` installed, and `test_multi_block_io[True-.vtkhdf-True]` succeeds without issue.

@krishanbhasin-px I'll leave it up to you if you want to support simple `MultiBlock` cases for `vtk==9.4.2` (no None blocks, no nested blocks). If not, then the easiest route is probably to only support saving `UnstructuredGrid` and `PolyData` for now as part of #7396, and add support for `MultiBlock` later for `vtk>=9.5`

user27182:
> @krishanbhasin-px I'll leave it up to you if you want to support simple `MultiBlock` cases for `vtk==9.4.2` (no None blocks, no nested blocks). If not, then the easiest route is probably to only support saving `UnstructuredGrid` and `PolyData` for now as part of #7396, and add support for `MultiBlock` later for `vtk>=9.5`

This support is added for both vtk 9.4 and 9.5 in #7396

PR #7538, opened 2025-05-21
Fix test failure for numpy 2.3

user27182:
### Overview

This test is failing for pre-release numpy (2.3.0dev):
https://github.com/pyvista/pyvista/actions/runs/15154194437/job/42643266844?pr=7536#step:12:8557
``` 
FAILED tests/core/test_composite.py::test_move_nested_field_data_to_root[prepend-//-data-Block-00//data-None] - assert False
 +  where False = <function array_equal at 0x7fd6dd963bf0>(pyvista_ndarray([1]), [42])
 +    where <function array_equal at 0x7fd6dd963bf0> = np.array_equal
```

Locally, I do not get `pyvista_ndarray([1]), [42]` as a comparison, but rather `pyvista_ndarray([44636345150720]), [42])`
Which  to me suggests that there's some sort of dangling pointer or invalid reference to the array. This particular test for `move_nested_field_data_to_root` is testing behavior similar to:
``` python
import pyvista as pv
import numpy as np
m1 = pv.MultiBlock()
m2 = pv.MultiBlock([m1])

m1.field_data['data'] = [42]
m2.field_data['data'] = m1.field_data['data']
assert np.array_equal(m1.field_data['data'], m2.field_data['data'])
```
where the nested field data is moved to the root multiblock without copying or removing the nested field data. However, the example above works, so it doesn't quite replicate the internals of `move_nested_field_data_to_root`.

I think this may be related to garbage collection for how array data is stored with `pyvista_ndarray` and/or `VTKArray`... not really sure.

Casting the input as an array "fixes" this, but it's just a workaround for the test, and this could still be a bug waiting to break something in the future....

### DIFF
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -1161,7 +1161,7 @@ def test_recursive_iterator_order(nested_fixture, order, expected_ids, expected_
     [('prepend', '//', 'data', 'Block-00//data'), ('preserve', '::', 'data', 'data')],
 )
 def test_move_nested_field_data_to_root(copy, field_data_mode, separator, name_in, name_out):
-    value = [42]
+    value = np.array([42])
     multi = pv.MultiBlock()
     multi.field_data[name_in] = value
     root = pv.MultiBlock([multi])
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -1161,6 +1161,7 @@ def test_recursive_iterator_order(nested_fixture, order, expected_ids, expected_
     [('prepend', '//', 'data', 'Block-00//data'), ('preserve', '::', 'data', 'data')],
 )
 def test_move_nested_field_data_to_root(copy, field_data_mode, separator, name_in, name_out):
+    # https://github.com/pyvista/pyvista/pull/7538
     value = np.array([42])
     multi = pv.MultiBlock()
     multi.field_data[name_in] = value


tests/core/test_composite.py
```diff
@@ -1161,7 +1161,7 @@ def test_recursive_iterator_order(nested_fixture, order, expected_ids, expected_
     [('prepend', '//', 'data', 'Block-00//data'), ('preserve', '::', 'data', 'data')],
 )
 def test_move_nested_field_data_to_root(copy, field_data_mode, separator, name_in, name_out):
-    value = [42]
+    value = np.array([42])
```
tkoyama010, line 1164:
```suggestion
    # https://github.com/pyvista/pyvista/pull/7538
    value = np.array([42])
```

tkoyama010 (changes requested):

tests/core/test_composite.py
user27182: +1

Note that CI will fail if you commit because of #7539, so this fix will need to be forced and merged with failures.

tkoyama010 (approved):
LGTM

user27182:
When the field data is moved, you can set `copy=False` to move it to root without making a copy. This means it's the same array object, but now it's at the root instead of being nested. If you set `copy=True` it's the same as above, but the moved array is a copy of the original array. In both cases, the array is moved, meaning it no longer exists with the nested MultiBlock. 

But for the `None` case, this like the `False` case, but the original array is `not` removed. So in this case we are left with two references to the same array. And with the `numpy` update the second reference fails or is somehow lost.

(In terms of the API, it probably would have been better to have this as a dedicated keyword like `keep_nested_array`. In the default case, this is `False`, and for the `copy=None` case, this would be like setting `keep_nested_array=True`)

user27182:
> When the field data is moved, you can set `copy=False` to move it to root without making a copy. This means it's the same array object, but now it's at the root instead of being nested. If you set `copy=True` it's the same as above, but the moved array is a copy of the original array. In both cases, the array is moved, meaning it no longer exists with the nested MultiBlock.
> 
> But for the `None` case, this is like the `False` case, but the original array is _not_ removed. So in this case we are left with two references to the same array. And with the `numpy` update the second reference fails or is somehow lost.
> 
> (In terms of the API, it probably would have been better to have this as a dedicated keyword like `keep_nested_array`. In the default case, this is `False`, and for the `copy=None` case, this would be like setting `keep_nested_array=True`)

What I describe above is not correct. This is the documentation:
https://github.com/pyvista/pyvista/blob/ee441b93c3d09b7889ea78959011e5e0173b1031/pyvista/core/composite.py#L749-L755

and this is the important bit of code:
https://github.com/pyvista/pyvista/blob/ee441b93c3d09b7889ea78959011e5e0173b1031/pyvista/core/composite.py#L966-L969

We can see that the root field data is updated, where `bool(copy)` means that data is _not_ copied in the `None` case. But then is cleared from the nested data afterward. So we don't have two references (as I incorrectly state above), but instead we only have a single reference. And this single reference is now lost or garbage collected somehow, such that its value is now incorrect.

MatthewFlamm:
There seems to be a conflation of copy and retain here.  

Is there a way to detect if the only reference to the object is internal to PyVista in this operation?

MatthewFlamm:
> There seems to be a conflation of copy and retain here.

Thinking about this more, it might be ill-defined to allow don't copy and don't retain.  This failure illustrates the problem with allowing this combination.  The datastructure might only exist internally in pyvista and the data handling is too complicated.  A better fix could be to add another param `retain` and, for now, don't allow both of them to be `False`.  Then if someone needs this functionality (unlikely?), they can try to weed through the reference pointing to fix it.

user27182:
> There seems to be a conflation of copy and retain here.

Yes, setting `copy=False` tries to retain the original array. Maybe it should be more like:

- `copy='deep'` (or `True`) -> deep copy
- `copy='retain'` -> keep original array 
- `copy=None` (or `False`) -> no copy

Though it's possible  in this case that the array setters and wrappers make shallow copies instead of truly retaining the original array, such that `array1 is not array2`  but yet they still share memory. So with more testing to verify this, maybe it should be `copy='shallow'` instead of `'retain'`.

This issue exists elsewhere in PyVista, where `copy` is ambiguous and is sometimes a deep copy, sometimes a shallow copy, or sometimes retained as-is. And often with only a binary choice, instead of providing all options.
- [points_to_cells](https://dev.pyvista.org/api/core/_autosummary/pyvista.imagedatafilters.points_to_cells#pyvista.ImageDataFilters.points_to_cells) has `copy=True` for deep copy, and `copy=False` for retain (no option to shallow copy)
- [as_polydata_blocks](https://dev.pyvista.org/api/core/_autosummary/pyvista.multiblock.as_polydata_blocks#pyvista.MultiBlock.as_polydata_blocks) has `copy=True` for shallow copy, and `copy=False` for retain (no option to deep copy)

> Is there a way to detect if the only reference to the object is internal to PyVista in this operation?

I am not sure. I think detecting references is tricky because of all the wrapping that's going on between `pyvista_ndarray`, `ndarray`, and `VTKArray`.

Internally, this method creates a temporary object to store the field data. I think this is a key cause of the issue, where the internal object may be garbage collected (along with its field data) despite the fact that the returned object now stores a reference to this temp object's field data array(s).

PR #7539, opened 2025-05-21
Add `filelock` to MNE test dependencies

user27182:
### Overview

Tests are failing, `filelock` is now required, see https://github.com/mne-tools/mne-python/pull/13241

@larsoner should we change this to install from extras, e.g. `pip install .[full]`, so that dependencies are always up to date?

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -51,7 +51,7 @@ jobs:
       - run: git clone --depth=1 https://github.com/mne-tools/mne-python.git --branch main --single-branch
       - run: ./mne-python/tools/setup_xvfb.sh
       - name: Install MNE dependencies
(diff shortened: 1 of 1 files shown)


MatthewFlamm:
> ### Overview
> Tests are failing, `filelock` is now required, see [mne-tools/mne-python#13241](https://github.com/mne-tools/mne-python/pull/13241)
> 
> @larsoner should we change this to install from extras, e.g. `pip install .[full]`, so that dependencies are always up to date?

This seems like a good approach to me.  We might need to adjust subsequent lines to their the appropriate `main` branch versions are being installed correctly.

larsoner:
It shouldn't be required, it should be optional -- I'll push a fix to MNE

larsoner:
Should be better now even without this PR hopefully!

user27182:
Thanks @larsoner ! This PR is no longer needed, MNE tests are successful, e.g. https://github.com/pyvista/pyvista/actions/runs/15190148151/job/42720567217

PR #7540, opened 2025-05-23
Revert #7538 'Fix test failure for numpy 2.3'

user27182:
### Overview

This change was originally needed because of a numpy bug https://github.com/numpy/numpy/issues/29038 which has now been resolved.

Will leave as draft until the numpy nightly wheels are updated (and CI passes)

### DIFF
--- a/tests/core/test_composite.py
+++ b/tests/core/test_composite.py
@@ -1161,8 +1161,7 @@ def test_recursive_iterator_order(nested_fixture, order, expected_ids, expected_
     [('prepend', '//', 'data', 'Block-00//data'), ('preserve', '::', 'data', 'data')],
 )
 def test_move_nested_field_data_to_root(copy, field_data_mode, separator, name_in, name_out):
-    # https://github.com/pyvista/pyvista/pull/7538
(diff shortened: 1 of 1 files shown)


seberg:
I'll say it here again, sorry for repeating:  The NumPy change wasn't a bug, it just made certain bugs more likely to be triggered/visible -- in this case probably otherwise uninitialized data (and was otherwise a missed optimization).  The bug may well be here and not in NumPy (to me that seems slightly more likely, but I have nothing to go on).

user27182:
> I'll say it here again, sorry for repeating: The NumPy change wasn't a bug, it just made certain bugs more likely to be triggered/visible -- in this case probably otherwise uninitialized data (and was otherwise a missed optimization). The bug may well be here and not in NumPy (to me that seems slightly more likely, but I have nothing to go on).

Thanks for clarifying! I have updated the wording here accordingly, please let me know if this new wording is still incorrect or does not reflect the actual situation.

Bug or no bug, based on local testing the commit that caused our CI to initially fail seems to have been addressed by your recent commit, will wait for our CI to confirm this though.

PR #7550, opened 2025-05-28
Add `BoundsTuple.__repr__` to align values with decimals

user27182:
### Overview

Sometimes the bounds are very long when there are non-round float values. This can make the docstring very long width-wise. This PR adds a repr method to put each value on its own line and align them with the decimal to make the bounds more legible.


tkoyama010 (approved):
LGTM

PR #7551, opened 2025-05-28
Enforce max line length of 99 characters

user27182:
### Overview

Enforce ruff [E501](https://docs.astral.sh/ruff/rules/line-too-long/), "line too long" with max 99 characters as per [PEP8](https://peps.python.org/pep-0008/#maximum-line-length)

Depends on #7550 to make some docstrings compliant


user27182:
> After merging, please add to https://github.com/pyvista/pyvista/blob/main/.git-blame-ignore-revs

There has been lots of discussion about git blame ignores in the past, e.g.: https://github.com/pyvista/pyvista/pull/6664#issuecomment-2364590326

While this PR has some auto-formatting (e.g. long lines with nested brackets were auto-expanded vertically by ruff), most of this PR was manual changes (e.g. pretty much all strings). Based on the linked comment above, it seems that git blame  ignore should be used purely for auto-formatted changes, and therefore this PR is not a good candidate for this. 
Nevertheless, I can still add this commit to the blame ignore if you prefer, please confirm @banesullivan

PR #7558, opened 2025-05-28
Remove unused ruff ignores for E131, E203, and E266

user27182:
### Overview

These rules are unnecessarily ignored:

- E131
- E203
- E266


tkoyama010 (approved):
LGTM

PR #7562, opened 2025-05-29
Add ARG rule to ruff

user27182:
### Overview

See https://docs.astral.sh/ruff/rules/#flake8-unused-arguments-arg


tkoyama010 (approved):
LGTM

PR #7564, opened 2025-05-29
Allow using `dict` inputs with `Plotter.add_legend`

user27182:


user27182:
Will need to update the doc image cache

PR #7572, opened 2025-05-30
Select all TRY rules with ruff

user27182:
### Overview

See https://docs.astral.sh/ruff/rules/#tryceratops-try


tkoyama010 (approved):
LGTM

PR #297, opened 2019-07-07
Opacity by scalar array or custom transfer function

banesullivan:
These changes enable a user to specify all kinds of options for the opacity:

- a scalar array on the mesh be used for opacity mapping to enable us to see two arrays at once 
- predefined transfer functions by string name
- custom built transfer functions
- single value for global opacity

### Why?

Maybe you have an uncertainty array - these changes make it so that you could make regions with high values of uncertainty be transparent. Or maybe you have a mesh with a modeled scalar array on another array of your modeling variance. This would allow you to control the opacity of the mesh by the variance.

### Example

Here I have a mesh with a modelled scalar array from Kriging and an associated kriging variance array which I'd like to use to make my mesh transparent:

```py
temp_grid = pv.read('temp_grid.vtk')
temp_grid
```
<table><tr><th>Header</th><th>Data Arrays</th></tr><tr><td><table><tr><th>UnstructuredGrid</th><th>Information</th></tr><tr><td>N Cells</td><td>2841790</td></tr><tr><td>N Points</td><td>551724</td></tr><tr><td>X Bounds</td><td>3.299e+05, 3.442e+05</td></tr><tr><td>Y Bounds</td><td>4.253e+06, 4.271e+06</td></tr><tr><td>Z Bounds</td><td>-2.700e+03, 2.300e+03</td></tr><tr><td>N Arrays</td><td>2</td></tr></table></td><td><table><tr><th>Name</th><th>Field</th><th>Type</th><th>N Comp</th><th>Min</th><th>Max</th></tr><tr><td><b>Temperature</b></td><td>Points</td><td>float64</td><td>1</td><td>6.729e+00</td><td>2.738e+02</td></tr><tr><td>Temperature_var</td><td>Points</td><td>float64</td><td>1</td><td>2.214e+02</td><td>1.107e+04</td></tr></table></td></tr> </table>

```py
# Make some contours
cntrs = temp_grid.contour(10, scalars='Temperature')

p = pv.Plotter(notebook=0)
p.add_mesh(cntrs, scalars='Temperature', 
           opacity='Temperature_var', 
           flip_opacity=True, cmap='bwr')
p.show()
```

Screen Shot 2019-07-07 at 4 09 40 PM

Which looks a bit more informative than just the scalar array alone:

Screen Shot 2019-07-07 at 4 05 30 PM

### Todo:

- [x] fix issue where the scalar bar is not appearing
[...]

PR #7574, opened 2025-05-30
Use linear opacity interpolation by default

### LINKED ISSUE #7570: Zero opacity is interpolated with non-zero values

Why are "blue", low level values, rendered here with `opacity = [0,0,0,1]`?

Code and data set are attached

[data.zip](https://github.com/user-attachments/files/20523525/data.zip)

```
import pandas as pd
import pyvista as pv
import numpy as np

data = pd.read_hdf('data.h5', key="df", mode="r")

shape = [len(np.unique(v)) for v in zip(*data.index)]
values = data.values.reshape(shape)

values = np.fft.fftshift(np.fft.ifftn(np.fft.fftshift(values)))
values = 20*np.log10(np.abs(values))

mesh = pv.ImageData(dimensions=shape)
mesh["scalars"] = values.flatten(order='F')
plotter = pv.Plotter()
plotter.add_volume(mesh, cmap="jet", opacity=[0,0,0,1], show_scalar_bar=True, mapper='smart')
plotter.show()
```

### System Information

```shell
  Date: Fri May 30 11:15:20 2025 Central Daylight Time

                OS : Windows (11 10.0.26120 SP0 Multiprocessor Free)
            CPU(s) : 24
           Machine : AMD64
      Architecture : 64bit
               RAM : 63.7 GiB
       Environment : Python
        GPU Vendor : Intel
      GPU Renderer : Intel(R) UHD Graphics 770
       GPU Version : 4.5.0 - Build 32.0.101.6793
     Render Window : vtkWin32OpenGLRenderWindow
  MathText Support : False

  Python 3.12.9 (tags/v3.12.9:fdb8142, Feb  4 2025, 15:27:58) [MSC v.1942 64
  bit (AMD64)]

           pyvista : 0.45.2
               vtk : 9.3.1
             numpy : 2.2.4
        matplotlib : 3.10.1
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 11.1.0
           imageio : 2.37.0
         pyvistaqt : 0.11.2
           IPython : 9.0.2
        ipywidgets : 7.8.5
             scipy : 1.15.2
              tqdm : 4.67.1
            meshio : 5.3.5
        jupyterlab : 4.3.6
             trame : 3.8.1
      trame_client : 3.6.1
      trame_server : 3.4.0
         trame_vtk : 2.8.15
     trame_vuetify : 2.9.0
      nest_asyncio : 1.6.0
```

### Screenshots

Image

user27182:
### Overview

Fix #7570

Currently, opacity rendering produces different results depending on if `scipy` is installed or not. I think this is bad practice, and the same result should be produced by default regardless if optional dependencies are installed.  In addition, using a non-linear mapping produces unexpected results in some cases, (see #7570), and should not be the default. Linear interpolation is the safer default choice.

EDIT: Looks like `'quadratic'` was set as the default initially way back as part of https://github.com/pyvista/pyvista/pull/297


pyvista/plotting/tools.py
```diff
@@ -537,7 +537,7 @@ def normalize(x, minimum=None, maximum=None):
     return (x - minimum) / (maximum - minimum)
 
 
-def opacity_transfer_function(mapping, n_colors, interpolate: bool = True, kind='quadratic'):
+def opacity_transfer_function(mapping, n_colors, interpolate: bool = True, kind='linear'):
```
tkoyama010, line 540: Please add [versionchanged](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged) to docstring.

user27182: added in [58316e8](https://github.com/pyvista/pyvista/pull/7574/commits/58316e8371b9d5b04f05837c6842d61e1dfcdd70)

tkoyama010 (approved):
LGTM

PR #7581, opened 2025-06-01
Deprecate and restrict the use of positional args

### LINKED ISSUE #5506: It would have been better to make `progress_bar` a keyword only argument everywhere

user27182:
### Overview

Fix #5506. 
This PR adds a new `_deprecate_positional_args` decorator to facilitate deprecating positional arguments across the entire repo. The decorator is inspired by a similar decorator from sklearn, as suggested here https://github.com/pyvista/pyvista/issues/5506#issuecomment-1906266250

With very few exceptions, the decorator is applied to:

-  any function with more than three positional args. This is enforced with [PLR0917](https://docs.astral.sh/ruff/rules/too-many-positional-arguments/).
- any function which has `bool` type positional arguments. This is enforced with [FBT001 and FBT002](https://docs.astral.sh/ruff/rules/#flake8-boolean-trap-fbt).

For now, these functions are littered with `# noqa` because they currently violate these rules. After the deprecation period, each function will need to have `*,` inserted into the signature to make the deprecated positional arguments keyword-only (see [PEP3102](https://peps.python.org/pep-3102/)). `ruff` will then automatically clean up all the `noqa` comments, and the decorator can be removed.

The deprecation is set to be in force for version `0.50`.

Not that the PLR0917 rule is in preview, and is not considered stable, though I didn't encounter any issue using it. But, enabling this did have some other linting/formatting side-effects, see (https://github.com/pyvista/pyvista/pull/7581/commits/36898d5dfb273e7f54413eb64b701f0355519569). These seemingly non-related changes are also included with this PR. 

A mostly complete list of functions that affected by this decorator is included below, along with their new signatures _after_ deprecation:
<details>

``` python
    plot(var_item, *, off_screen, full_screen, screenshot, ...)
    copy_vtk_array(array, *, deep)
    convert_array(arr, name, *, deep, array_type)
    get_array(mesh, name, *, preference, err)
    get_array_association(mesh, name, *, preference, err)
    numpy_to_idarr(ind, *, deep, return_ind)
    axis_angle_rotation(axis, angle, *, point, deg)
    apply_transformation_to_points(transformation, points, *, inplace)
    VtkErrorCatcher.__init__(self, *, raise_errors, send_to_logging)
    Observer.__init__(self, event_type, *, log, store_history)
    Observer.get_message(self, *, etc)
    set_vtkwriter_mode(vtk_writer, *, use_binary)
    read(filename, *, force_ext, file_format, progress_bar)
    read_texture(filename, *, progress_bar)
    read_exodus(filename, *, animate_mode_shapes, apply_displacements, displacement_magnitude, ...)
    read_grdecl(filename, *, elevation, other_keywords)
    axis_rotation(points, angle, *, inplace, deg, axis)
    voxelize(mesh, *, density, check_surface, enclosed, ...)
    voxelize_volume(mesh, *, density, check_surface, enclosed, ...)
    transform_vectors_sph_to_cart(*, theta, phi, r, ...)
    merge(datasets, *, merge_points, main_has_priority, progress_bar)
    sample_function(function, *, bounds, dim, compute_normals, ...)
    ConeSource.__init__(self, *, center, direction, ...)
    CylinderSource.__init__(self, *, center, direction, ...)
    Text3DSource.__init__(self, string, *, depth, width, ...)
    CubeSource.__init__(self, *, center, x_length, ...)
    DiscSource.__init__(self, *, center, inner, ...)
    SphereSource.__init__(self, *, radius, center, ...)
    PolygonSource.__init__(self, *, center, radius, ...)
    PlaneSource.__init__(self, *, i_resolution, j_resolution, ...)
    ArrowSource.__init__(self, *, tip_length, tip_radius, ...)
    BoxSource.__init__(self, bounds, *, level, quads)
    SuperquadricSource.__init__(self, *, center, scale, ...)
    Capsule(*, center, direction, radius, ...)
    Cylinder(*, center, direction, radius, ...)
    CylinderStructured(*, radius, height, center, ...)
    Arrow(*, start, direction, tip_length, ...)
    Sphere(*, radius, center, direction, ...)
    SolidSphere(*, outer_radius, inner_radius, radius_resolution, ...)
    SolidSphereGeneric(*, radius, theta, phi, ...)
    Plane(*, center, direction, i_size, ...)
    Line(pointa, pointb, *, resolution)
    Tube(*, pointa, pointb, resolution, ...)
    Cube(*, center, x_length, y_length, ...)
    Box(bounds, *, level, quads)
    Cone(*, center, direction, height, ...)
    Polygon(*, center, radius, normal, ...)
    Disc(*, center, inner, outer, ...)
    Text3D(string, *, depth, width, height, ...)
    Wavelet(*, extent, center, maximum, ...)
    CircularArc(*, pointa, pointb, center, ...)
    CircularArcFromNormal(*, center, resolution, normal, ...)
    Circle(*, radius, resolution)
    Ellipse(semi_major_axis, semi_minor_axis, *, resolution)
    Superquadric(*, center, scale, size, ...)
    PlatonicSolid(kind, *, radius, center)
    ImageNoiseSource.__init__(self, *, whole_extent, minimum, ...)
    ImageSinusoidSource.__init__(self, *, whole_extent, direction, ...)
    ImageGaussianSource.__init__(self, *, center, whole_extent, ...)
    KochanekSpline(points, *, tension, bias, continuity, ...)
    ParametricRandomHills(*, numberofhills, hillxvariance, hillyvariance, ...)
    ParametricSuperEllipsoid(xradius, yradius, zradius, *, n1, n2, kwargs)
    ParametricSuperToroid(*, ringradius, crosssectionradius, xradius, ...)
    parametric_keywords(parametric_function, *, min_u, max_u, min_v, ...)
    surface_from_para(parametric_function, *, u_res, v_res, w_res, ...)
    vtk_points(points, *, deep, force_float, allow_empty)
    lines_from_points(points, *, close)
    fit_plane_to_points(points, *, return_meta, resolution, init_normal)
    GaussianCubeReader.read(self, *, grid)
    DataSetAttributes.set_array(self, data, name, *, deep_copy)
    DataSetAttributes.set_scalars(self, scalars, name, *, deep_copy)
    DataSetAttributes.set_vectors(self, vectors, name, *, deep_copy)
    DataSetAttributes.update(self, array_dict, *, copy)
    DataObject.save(self, filename, *, binary, texture)
    DataObject.head(self, *, display, html)
    DataObject.copy(self, *, deep)
    DataObject.add_field_data(self, array, name, *, deep)
    DataObjectFilters.transform(self, trans, *, transform_all_input_vectors, inplace, ...)
    DataObjectFilters.reflect(self, normal, *, point, inplace, ...)
    DataObjectFilters.rotate_x(self, angle, *, point, transform_all_input_vectors, ...)
    DataObjectFilters.rotate_y(self, angle, *, point, transform_all_input_vectors, ...)
    DataObjectFilters.rotate_z(self, angle, *, point, transform_all_input_vectors, ...)
    DataObjectFilters.rotate_vector(self, vector, angle, *, point, transform_all_input_vectors, ...)
    DataObjectFilters.rotate(self, rotation, *, point, transform_all_input_vectors, ...)
    DataObjectFilters.scale(self, xyz, *, transform_all_input_vectors, inplace, ...)
    DataObjectFilters.flip_x(self, *, point, transform_all_input_vectors, ...)
    DataObjectFilters.flip_y(self, *, point, transform_all_input_vectors, ...)
    DataObjectFilters.flip_z(self, *, point, transform_all_input_vectors, ...)
    DataObjectFilters.flip_normal(self, normal, *, point, transform_all_input_vectors, ...)
    DataObjectFilters.clip(self, normal, *, origin, invert, ...)
    DataObjectFilters.clip_box(self, bounds, *, invert, factor, ...)
    DataObjectFilters.slice_implicit(self, implicit_function, *, generate_triangles, contour, ...)
    DataObjectFilters.slice(self, normal, *, origin, generate_triangles, ...)
    DataObjectFilters.slice_orthogonal(self, *, x, y, ...)
    DataObjectFilters.slice_along_axis(self, n, axis, *, tolerance, generate_triangles, ...)
    DataObjectFilters.slice_along_line(self, line, *, generate_triangles, contour, ...)
    DataObjectFilters.extract_all_edges(self, *, use_all_points, clear_data, ...)
    DataObjectFilters.elevation(self, *, low_point, high_point, ...)
    DataObjectFilters.compute_cell_sizes(self, *, length, area, ...)
    DataObjectFilters.cell_centers(self, *, vertex, progress_bar)
    DataObjectFilters.cell_data_to_point_data(self, *, pass_cell_data, progress_bar)
    DataObjectFilters.ctp(self, *, pass_cell_data, progress_bar, ...)
    DataObjectFilters.point_data_to_cell_data(self, *, pass_point_data, categorical, ...)
    DataObjectFilters.ptc(self, *, pass_point_data, progress_bar, ...)
    DataObjectFilters.triangulate(self, *, inplace, progress_bar)
    DataObjectFilters.sample(self, target, *, tolerance, pass_cell_data, ...)
    DataSetFilters.align(self, target, *, max_landmarks, max_mean_distance, ...)
    DataSetFilters.compute_implicit_distance(self, surface, *, inplace)
    DataSetFilters.clip_scalar(self, *, scalars, invert, ...)
    DataSetFilters.clip_surface(self, surface, *, invert, value, ...)
    DataSetFilters.threshold(self, value, *, scalars, invert, ...)
    DataSetFilters.threshold_percent(self, percent, *, scalars, invert, ...)
    DataSetFilters.outline(self, *, generate_faces, progress_bar)
    DataSetFilters.outline_corners(self, *, factor, progress_bar)
    DataSetFilters.extract_geometry(self, *, extent, progress_bar)
    DataSetFilters.contour(self, isosurfaces, scalars, *, compute_normals, compute_gradients, ...)
    DataSetFilters.texture_map_to_plane(self, *, origin, point_u, ...)
    DataSetFilters.texture_map_to_sphere(self, *, center, prevent_seam, ...)
    DataSetFilters.glyph(self, *, orient, scale, ...)
    DataSetFilters.connectivity(self, extraction_mode, variable_input, *, scalar_range, scalars, ...)
    DataSetFilters.extract_largest(self, *, inplace, progress_bar)
    DataSetFilters.split_bodies(self, *, label, progress_bar)
    DataSetFilters.warp_by_scalar(self, scalars, *, factor, normal, ...)
    DataSetFilters.warp_by_vector(self, vectors, *, factor, inplace, ...)
    DataSetFilters.delaunay_3d(self, *, alpha, tol, ...)
    DataSetFilters.select_enclosed_points(self, surface, *, tolerance, inside_out, ...)
    DataSetFilters.interpolate(self, target, *, sharpness, radius, ...)
    DataSetFilters.streamlines(self, vectors, *, source_center, source_radius, ...)
    DataSetFilters.streamlines_from_source(self, source, vectors, *, integrator_type, integration_direction, ...)
    DataSetFilters.streamlines_evenly_spaced_2D(self, vectors, *, start_position, integrator_type, ...)
    DataSetFilters.decimate_boundary(self, target_reduction, *, progress_bar)
    DataSetFilters.sample_over_line(self, pointa, pointb, *, resolution, tolerance, ...)
    DataSetFilters.plot_over_line(self, pointa, pointb, *, resolution, scalars, ...)
    DataSetFilters.sample_over_multiple_lines(self, points, *, tolerance, progress_bar)
    DataSetFilters.sample_over_circular_arc(self, *, pointa, pointb, ...)
    DataSetFilters.sample_over_circular_arc_normal(self, *, center, resolution, ...)
    DataSetFilters.plot_over_circular_arc(self, *, pointa, pointb, ...)
    DataSetFilters.plot_over_circular_arc_normal(self, *, center, resolution, ...)
    DataSetFilters.extract_cells(self, ind, *, invert, progress_bar)
    DataSetFilters.extract_points(self, ind, *, adjacent_cells, include_cells, ...)
    DataSetFilters.extract_surface(self, *, pass_pointid, pass_cellid, ...)
    DataSetFilters.surface_indices(self, *, progress_bar)
    DataSetFilters.extract_feature_edges(self, feature_angle, *, boundary_edges, non_manifold_edges, ...)
    DataSetFilters.merge_points(self, *, tolerance, inplace, ...)
    DataSetFilters.merge(self, grid, *, merge_points, tolerance, ...)
    DataSetFilters.compute_cell_quality(self, quality_measure, *, null_value, progress_bar)
    DataSetFilters.compute_derivative(self, scalars, *, gradient, divergence, ...)
    DataSetFilters.shrink(self, shrink_factor, *, progress_bar)
    DataSetFilters.tessellate(self, *, max_n_subdivide, merge_points, ...)
    DataSetFilters.integrate_data(self, *, progress_bar)
    DataSetFilters.partition(self, n_partitions, *, generate_global_id, as_composite)
    DataSetFilters.extract_cells_by_type(self, cell_types, *, progress_bar)
    DataSetFilters.sort_labels(self, scalars, *, preference, output_scalars, ...)
    DataSetFilters.pack_labels(self, *, sort, scalars, ...)
    CompositeFilters.combine(self, *, merge_points, tolerance)
    CompositeFilters.outline(self, *, generate_faces, nested, ...)
    CompositeFilters.outline_corners(self, *, factor, nested, ...)
    CompositeFilters._compute_normals(self, *, cell_normals, point_normals, ...)
    ImageDataFilters.gaussian_smooth(self, *, radius_factor, std_dev, ...)
    ImageDataFilters.median_smooth(self, *, kernel_size, scalars, ...)
    ImageDataFilters.extract_subset(self, voi, rate, *, boundary, progress_bar)
    ImageDataFilters.image_dilate_erode(self, dilate_value, erode_value, *, kernel_size, scalars, ...)
    ImageDataFilters.image_threshold(self, threshold, *, in_value, out_value, ...)
    ImageDataFilters.fft(self, *, output_scalars_name, progress_bar)
    ImageDataFilters.rfft(self, *, output_scalars_name, progress_bar)
    ImageDataFilters.low_pass(self, x_cutoff, y_cutoff, z_cutoff, *, order, output_scalars_name, ...)
    ImageDataFilters.high_pass(self, x_cutoff, y_cutoff, z_cutoff, *, order, output_scalars_name, ...)
    ImageDataFilters.contour_labeled(self, *, n_labels, smoothing, ...)
    PolyDataFilters.edge_mask(self, angle, *, progress_bar)
    PolyDataFilters.boolean_union(self, other_mesh, *, tolerance, progress_bar)
    PolyDataFilters.boolean_intersection(self, other_mesh, *, tolerance, progress_bar)
    PolyDataFilters.boolean_difference(self, other_mesh, *, tolerance, progress_bar)
    PolyDataFilters.merge(self, dataset, *, merge_points, tolerance, ...)
    PolyDataFilters.intersection(self, mesh, *, split_first, split_second, ...)
    PolyDataFilters.curvature(self, curv_type, *, progress_bar)
    PolyDataFilters.smooth(self, *, n_iter, relaxation_factor, ...)
    PolyDataFilters.smooth_taubin(self, *, n_iter, pass_band, ...)
    PolyDataFilters.decimate_pro(self, reduction, *, feature_angle, split_angle, ...)
    PolyDataFilters.tube(self, *, radius, scalars, ...)
    PolyDataFilters.subdivide(self, nsub, subfilter, *, inplace, progress_bar)
    PolyDataFilters.subdivide_adaptive(self, *, max_edge_len, max_tri_area, ...)
    PolyDataFilters.decimate(self, target_reduction, *, volume_preservation, attribute_error, ...)
    PolyDataFilters.compute_normals(self, *, cell_normals, point_normals, ...)
    PolyDataFilters.clip_closed_surface(self, normal, *, origin, tolerance, ...)
    PolyDataFilters.fill_holes(self, hole_size, *, inplace, progress_bar)
    PolyDataFilters.clean(self, *, point_merging, tolerance, ...)
    PolyDataFilters.geodesic(self, start_vertex, end_vertex, *, inplace, keep_order, ...)
    PolyDataFilters.geodesic_distance(self, start_vertex, end_vertex, *, use_scalar_weights, progress_bar)
    PolyDataFilters.ray_trace(self, origin, end_point, *, first_point, plot, ...)
    PolyDataFilters.multi_ray_trace(self, origins, directions, *, first_point, retry)
    PolyDataFilters.plot_boundaries(self, *, edge_color, line_width, ...)
    PolyDataFilters.plot_normals(self, *, show_mesh, mag, ...)
    PolyDataFilters.remove_points(self, remove, mode, *, keep_scalars, inplace)
    PolyDataFilters.delaunay_2d(self, *, tol, alpha, ...)
    PolyDataFilters.compute_arc_length(self, *, progress_bar)
    PolyDataFilters.project_points_to_plane(self, *, origin, normal, ...)
    PolyDataFilters.ribbon(self, *, width, scalars, ...)
    PolyDataFilters.extrude(self, vector, *, capping, inplace, ...)
    PolyDataFilters.extrude_rotate(self, *, resolution, inplace, ...)
    PolyDataFilters.extrude_trim(self, direction, trim_surface, *, extrusion, capping, ...)
    PolyDataFilters.strip(self, *, join, max_length, ...)
    PolyDataFilters.collision(self, other_mesh, *, contact_mode, box_tolerance, ...)
    PolyDataFilters.contour_banded(self, n_contours, *, rng, scalars, ...)
    PolyDataFilters.reconstruct_surface(self, *, nbr_sz, sample_spacing, ...)
    PolyDataFilters.triangulate_contours(self, *, display_errors, progress_bar)
    PolyDataFilters.protein_ribbon(self, *, progress_bar)
    RectilinearGridFilters.to_tetrahedra(self, tetra_per_cell, *, mixed, pass_cell_ids, ...)
    StructuredGridFilters.extract_subset(self, voi, rate, *, boundary)
    UnstructuredGridFilters.clean(self, *, tolerance, remove_unused_points, ...)
    DataSet.copy_meta_from(self, ido, *, deep)
    DataSet.copy_from(self, mesh, *, deep)
    DataSet.cast_to_pointset(self, *, pass_cell_data)
    DataSet.cast_to_poly_points(self, *, pass_cell_data)
    DataSet.find_closest_cell(self, point, *, return_closest_point)
    MultiBlock.get_data_range(self, name, *, allow_missing, preference)
    MultiBlock.clean(self, *, empty)
    MultiBlock.copy_meta_from(self, ido, *, deep)
    MultiBlock.copy(self, *, deep)
    MultiBlock.shallow_copy(self, to_copy, *, recursive)
    MultiBlock.set_active_scalars(self, name, *, preference, allow_missing)
    MultiBlock.as_polydata_blocks(self, *, copy)
    MultiBlock.as_unstructured_grid_blocks(self, *, copy)
    ImageData.__init__(self, uinput, *, dimensions, spacing, ...)
    Cell.__init__(self, vtk_cell, cell_type, *, deep)
    Cell.copy(self, *, deep)
    CellArray.__init__(self, cells, *, n_cells, deep)
    CellArray.from_arrays(offsets, connectivity, *, deep)
    CellArray.from_regular_cells(cls, cells, *, deep)
    _PointSet.center_of_mass(self, *, scalars_weight)
    _PointSet.remove_cells(self, ind, *, inplace)
    _PointSet.translate(self, xyz, *, transform_all_input_vectors, inplace)
    PointSet.__init__(self, var_inp, *, deep, force_float)
    PointSet.cast_to_polydata(self, *, deep)
    PolyData.__init__(self, var_inp, faces, *, n_faces, lines, ...)
    PolyData.from_regular_faces(cls, points, faces, *, deep)
    PolyData.save(self, filename, *, binary, texture, ...)
    UnstructuredGrid.linear_copy(self, *, deep)
    StructuredGrid.hide_cells(self, ind, *, inplace)
    ExplicitStructuredGrid.clean(self, *, tolerance, remove_unused_points, ...)
    ExplicitStructuredGrid.save(self, filename, *, binary, texture)
    ExplicitStructuredGrid.hide_cells(self, ind, *, inplace)
    ExplicitStructuredGrid.show_cells(self, *, inplace)
    ExplicitStructuredGrid.compute_connectivity(self, *, inplace)
    ExplicitStructuredGrid.compute_connections(self, *, inplace)
    PartitionedDataSet.copy(self, *, deep)
    Report.__init__(self, *, additional, ncol, ...)
```

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -70,8 +70,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.config.python-version }}
-          if: env.USE_CACHE == 'true'
-          cache: "pip"
+          cache: ${{ env.USE_CACHE == 'true' && 'pip' || '' }}
 
       - name: Install PyVista test dependencies
         run: |


pyvista/core/utilities/misc.py
```diff
@@ -379,3 +384,87 @@ def name(self, value: str) -> None:
...
+    """
+
+    def _inner_deprecate_positional_args(f):  # noqa: ANN001, ANN202
+        sig = signature(f)
+        param_names = list(sig.parameters)
+
+        @wraps(f)
+        def inner_f(*args, **kwargs):  # noqa: ANN202
+            if allowed:
+                max_allowed = 5
```
edabor: How about getting the setting value from `pyproject.toml` ? https://docs.astral.sh/ruff/settings/#lint_pylint_max-positional-args 

We might need to ship `pyproject.toml` in the `pyvista` package though ...

pyvista/core/utilities/misc.py
```diff
@@ -379,3 +384,87 @@ def name(self, value: str) -> None:
...
+                        f'Got {n_allowed}:\n{allowed}'
+                    )
+                    raise RuntimeError(msg)
+
+            # Map args to parameter names
+            passed_positional_names = param_names[: len(args)]
+
+            # Exclude allowed ones
+            if allowed:
+                offending_args = [name for name in passed_positional_names if name not in allowed]
```
edabor: We could check that the `allowed` arguments are effectively part of the function signature before filtering them, ie. when returning `inner_f`

pyvista/core/utilities/misc.py
user27182: Yes this would be better. We can probably still hard-code the value as a global variable in the code, and add a separate test to compare the global var to the one in pyproject to make sure they're in sync. Then we don't need to directly reference pyproject.toml at runtime.

edabor: Great idea ! +1

pyvista/core/utilities/misc.py
user27182: Can you elaborate or provide an example of this? Can we also handle to cls and self variables this way? (These should always be allowed, but at the same time should not be mentioned in user-facing error messages)

edabor: Here is a patch proposal [params.patch](https://github.com/user-attachments/files/20555024/params.patch) to check if the allowed parameters are in the function signature.

For the `self` and `cls`, not sure if we can auto detect them ...

pyvista/core/utilities/misc.py
user27182: Done in [a00e71b](https://github.com/pyvista/pyvista/pull/7581/commits/a00e71bc065e032dc0b03f67ef03e6895603390f)

pyvista/core/utilities/misc.py
user27182: Ah now I understand. Had a brain fart... yes this is effectively validating the `allowed` input values. We should also check order of the args, so that we can provide a message for what the expected signature should be post-deprecation. 

E.g. we can provide a dev-specific (not user-facing) error message that says something like 
```
Change function signature to have the form `foo(arg1, arg2, *, ...)` 
and remove the `_deprecate_positional_args` decorator
```

where the decorator has `allowed=['arg1', 'arg2']`.

edabor: I agree that we should issue a warning only for pyvista devs describing how to refactor the function.

The warning should be triggered at decoration time, such that even if the function is not called in the test suite, the module import will still issue the warning.
A warning is to be preferred to an error, since raising an error will prevent discovering later deprecation issues.

user27182: > I agree that we should issue a warning only for pyvista devs describing how to refactor the function.
> 
> The warning should be triggered at decoration time, such that even if the function is not called in the test suite, the module import will still issue the warning.

This would be ideal, but when I try this I get circular import errors when trying to import the pyvista version info. So it may be necessary to only include this at call time. If needed, we can add a separate test to check for the presence of the decorator anywhere in any of the source ccde using regex, and raise an error (post-deprecation) if the decorator is used anywhere that way.

> A warning is to be preferred to an error, since raising an error will prevent discovering later deprecation issues.

It may be the case that warnings are treated as errors by pytest for CI, so in that regard a warning is not so different from an error. But, I have found that there are uncaught warnings issued in some of the Sphinx Gallery examples (e.g. the `glyph_table.py` example currently emits a UserWarning). Unless we can guarantee that these will be caught and dealt with, I think it's safer to raise an error to ensure it causes a CI failure post-deprecation so that the examples can be properly updated.

edabor: > This would be ideal, but when I try this I get circular import errors when trying to import the pyvista version info. So it may be necessary to only include this at call time. If needed, we can add a separate test to check for the presence of the decorator anywhere in any of the source ccde using regex, and raise an error (post-deprecation) if the decorator is used anywhere that way.

When defined in `_deprecate_positional_args` in local, the following **does** raise a circular import error:
```python
from pyvista import version_info
```
but the following does not:
```python
from pyvista._version import version_info
```
Can you confirm ?

pyvista/_deprecate_positional_args.py
```diff
@@ -0,0 +1,189 @@
...
+                    error_positional_args()
+
+            # Raise error post-deprecation
+            if version_info >= version:
+                # Construct expected positional args and signature
+                positional_args = ['self'] if has_self else ['cls'] if has_cls else []
+                if allowed is not None:
+                    for name in allowed:
+                        positional_args.append(name)  # noqa: PERF402
+                new_signature = f'{qualified_name()}({", ".join(positional_args)}, *, ...)'
```
edabor: You can use the `inspect.Signature` class to build the new signature string:
```python
from inspect import Parameter
from inspect import Signature
from typing import *

sig = Signature(
    [
        Parameter(
            'foo',
            kind=Parameter.POSITIONAL_OR_KEYWORD,
        ),
        Parameter(
            'bar',
            kind=Parameter.KEYWORD_ONLY,
            default=None,
        ),
        Parameter(
            'baz',
            kind=Parameter.KEYWORD_ONLY,
            default=None,
        ),
    ]
)
print(sig)
```
```python
(foo, *, bar=None, baz=None)
```

pyvista/core/utilities/misc.py
user27182: > > This would be ideal, but when I try this I get circular import errors when trying to import the pyvista version info. So it may be necessary to only include this at call time. If needed, we can add a separate test to check for the presence of the decorator anywhere in any of the source ccde using regex, and raise an error (post-deprecation) if the decorator is used anywhere that way.
> 
> When defined in `_deprecate_positional_args` in local, the following **does** raise a circular import error:
> 
> ```python
> from pyvista import version_info
> ```
> 
> but the following does not:
> 
> ```python
> from pyvista._version import version_info
> ```
> 
> Can you confirm ?

Confirmed that this import works. We can move the post-deprecation error up, thanks.

Both these import statements were tried earlier and were failing when the decorator was defined in `pyvista.core.utiltiies.misc`. Moving it to a standalone module in https://github.com/pyvista/pyvista/pull/7581/commits/e06cb3b6620f466478fb598cb3f094c0df0ec4b2 must have fixed this (this move was needed to fix similar circular import errors caused by `plotting._plot.py` being imported for use by `DataSet.plot`).

pyvista/_deprecate_positional_args.py
user27182: > You can use the `inspect.Signature` class to build the new signature string:
> 
> ```python
> from inspect import Parameter
> from inspect import Signature
> from typing import *
> 
> sig = Signature(
>     [
>         Parameter(
>             'foo',
>             kind=Parameter.POSITIONAL_OR_KEYWORD,
>         ),
>         Parameter(
>             'bar',
>             kind=Parameter.KEYWORD_ONLY,
>             default=None,
>         ),
>         Parameter(
>             'baz',
>             kind=Parameter.KEYWORD_ONLY,
>             default=None,
>         ),
>     ]
> )
> print(sig)
> ```
> 
>  ```python
> (foo, *, bar=None, baz=None)
> ```

Thanks. Do we want to include the keyword-only args in the message though? Post-deprecation, the main thing we care about is inserting `*`. Including the keywords will make this message very long in some cases (e.g. `add_mesh`). Maybe we can cap it to show a max of 6 args total, followed by `...` if there are more?

I'm also thinking that, although this message is helpful, there are hundreds of function signatures to modify. Perhaps including a helper script to automating this clean up (i.e. insert `*` and remove the decorator) would be  best... in which case this message doesn't matter too much.

edabor: > Thanks. Do we want to include the keyword-only args in the message though? Post-deprecation, the main thing we care about is inserting `*`. Including the keywords will make this message very long in some cases (e.g. `add_mesh`). Maybe we can cap it to show a max of 6 args total, followed by `...` if there are more?

I agree, the error message can be very long for some functions. Capping is good idea.

> I'm also thinking that, although this message is helpful, there are hundreds of function signatures to modify. Perhaps including a helper script to automating this clean up (i.e. insert `*` and remove the decorator) would be best... in which case this message doesn't matter too much.

Unfortunately, ruff `PLR0917` does not provide a fix to rewrite with kw-only arguments.
However, we can get inspiration from https://github.com/Kludex/kwonly-transformer/tree/main and implement our own parser that will look for decorated functions only.

user27182: > You can use the `inspect.Signature` class to build the new signature string:

This is done in [da7d8f3](https://github.com/pyvista/pyvista/pull/7581/commits/da7d8f3f7e9fcb090ec68b652e625e079574fd5d)

> > I'm also thinking that, although this message is helpful, there are hundreds of function signatures to modify. Perhaps including a helper script to automating this clean up (i.e. insert `*` and remove the decorator) would be best... in which case this message doesn't matter too much.
> 
> Unfortunately, ruff `PLR0917` does not provide a fix to rewrite with kw-only arguments. However, we can get inspiration from https://github.com/Kludex/kwonly-transformer/tree/main and implement our own parser that will look for decorated functions only.

The decorator is working well as-is with individual post-deprecation messages. I think we can look into automating the changes later when it comes time to make the change.

user27182:
This is ready for review. This PR is pretty big, I'd say the key areas to review is the decorator code itself, as well as the new "allowed" positional args... for this I have included a list of most affected functions and their effective new signatures. Deciding which args should be allowed to be positional and which should be forced as keyword-only is pretty subjective, so any input on this list would be helpful. 

E.g. should `extract_feature_edges` allow `feature_angle` to be positional? Should `slice_orthogonal` allow `x`, `y`, and `z`, keywords to be positional? Etc.

akaszynski:
I'm fully behind this.

I've started to get in the habit of using kwargs even when it's not required as it results in "self documenting" code. This also helps (as in the case of #7608 and others) to ensure backwards compatibility when adding additional kwargs.

Sorry the delay in reviewing these PRs, feel free to @ me (or ping me on slack). We've got a ton of PRs and I'm happy to review more of them.

akaszynski (dismissed):

user27182:
> I'm fully behind this.
> 
> I've started to get in the habit of using kwargs even when it's not required as it results in "self documenting" code.

Thanks for the review! Let me know if you think we should be more strict here, and enforce 100% keyword-only for more filters (see full list in the top comment). E.g. doing `extract_feature_edges(30)` is still possible with this PR, but maybe it's not obvious that `30` is an angle, so should we enforce `extract_feature_edges(feature_angle=30)`? There are a few cases like this... We can also make these smaller fine-tuning adjustments in a follow-up PR.

akaszynski:
> We can also make these smaller fine-tuning adjustments in a follow-up PR.

Let's go with incremental. I think PLR0917 is a good starting point without becoming too draconian.

akaszynski (approved):

akaszynski:
Fantastic work!

pyvista/plotting/texture.py
```diff
@@ -9,6 +9,8 @@
 import numpy as np
 
 import pyvista
+
+pyvista.OFF_SCREEN = True
```
larsoner, line 13: This looks like a mistake / debugging cruft, right? Opening a PR...

Issue #7582, opened 2025-06-01
Detect test images that are not used in the tests


tkoyama010:
### Describe what maintenance you would like added.

I noticed in #7565 that when a test image is no longer used in the tests, there is currently no mechanism to detect it. We need a way to notify developers when test images become unused.

### Links to source code.

None

### Pseudocode or Screenshots

None

user27182:
Is that what the `--fail_extra_image_cache` flag is used for with https://github.com/pyvista/pytest-pyvista?

Maybe this flag just needs to be added to CI

MatthewFlamm:
> Is that what the `--fail_extra_image_cache` flag is used for with https://github.com/pyvista/pytest-pyvista?
> 
> Maybe this flag just needs to be added to CI

This should be correct.

edabor:
> Is that what the `--fail_extra_image_cache` flag is used for with https://github.com/pyvista/pytest-pyvista?
> 
> Maybe this flag just needs to be added to CI

This flag is already added to CI.
https://github.com/pyvista/pyvista/blob/382d91050d0583cec91329532fcaccd1bc4e0a04/.github/workflows/testing-and-deployment.yml#L78

However, the doc in `pytest_pyvista` README describes the usage you target:
> - When using --fail_extra_image_cache if there is an extra image in the cache, it will report as an error.


but the implementation suggests that the flag is used to raise an error if the cache file is not found:
https://github.com/pyvista/pytest-pyvista/blob/821e35555e23694cdd8420c0c43ee55cb04038b4/pytest_pyvista/pytest_pyvista.py#L200-L204

In other words, I think there is a mismatch between the documentation and the implementation.

MatthewFlamm:
You are right, my memory is bad. 

https://github.com/pyvista/pytest-pyvista/issues/28

edabor:
> You are right, my memory is bad.
> 
> [pyvista/pytest-pyvista#28](https://github.com/pyvista/pytest-pyvista/issues/28)

See also the PR https://github.com/pyvista/pytest-pyvista/pull/92

I think it should be merged to clarify the intent of the CLI flag. 
Then, we can make the `--fail_extra_image_cache` to work as documented.

PR #7584, opened 2025-06-02
Remove unused cached plotting images

user27182:
### Overview

Related to #7582

I downloaded one the generated images artifacts (Linux 3.13 nightly), and compared them to the cache with:

``` python
import os
cached = set(os.listdir('./tests/plotting/image_cache'))
generated = set(os.listdir(artifact))
print(cached - generated)
```

and reviewed all images listed in the result, removing any which did not have a test with the same name as the image.


user27182:
I keep seeing this build error across different PRs, not sure what is causing it:
https://github.com/pyvista/pyvista/actions/runs/15400062719/job/43330275296?pr=7584#step:13:553

```
ERROR tests/doc/tst_doc_build.py::test_static_images[AeroSandbox] - ExceptionGroup: 
multiple unraisable exception warnings (57 sub-exceptions)
```

PR #7588, opened 2025-06-04
Remove Kitware from CODEOWNERS

banesullivan:
They don't seem interested, so removing the restriction

This will unblock PRs modifying these files since they wait on my stand-in approval


tkoyama010 (approved):
Thanks!

PR #7590, opened 2025-06-05
Fix PyVista installation for VTK dev testing and Codespaces

user27182:
### Overview

Follow-up to PR #7459 and  #7430

Use the new dependency groups for VTK dev testing, and fix this error when running codespaces: 
"This codespace is currently running in recovery mode due to a configuration error. Please review the creation logs, update your dev container configuration as needed, and run the "Rebuild Container" command to rectify."


user27182:
I confirmed that the new VTK dev test runs properly with the label applied. And Codespaces starts up without issue.

banesullivan (approved):

PR #7426, opened 2025-04-15
Deprecate `main_has_priority` with `merge` filter

user27182:
### Overview

The behavior of `vtkAppendFilter` has recently changed (from latest vtk pre-release) such that it now effectively sets `main_has_priority` by default. This fixes a bug with `vtkAppendFilter` where the _last_ appended mesh had priority, whereas it makes more sense for the _first_ appended mesh to have priority. See https://gitlab.kitware.com/vtk/vtk/-/issues/19656

IMO, it seems the keyword `main_has_priority` for `merge` was effectively a workaround for this bug, and is more of a fix than a feature. I don't think we need to keep it for use with future vtk versions.

PR #7594, opened 2025-06-06
Fix merging field data with `main_has_priority`

user27182:
### Overview

`main_has_priority` ensures that the arrays from the "main" mesh are prioritized. This was not the case for field data arrays. This PR fixes this.

### DIFF
--- a/tests/core/test_dataobject.py
+++ b/tests/core/test_dataobject.py
@@ -276,7 +276,7 @@ def test_user_dict_persists_with_merge_filter():
     sphere2.user_dict['name'] = 'sphere2'
 
     merged = sphere1 + sphere2
-    assert merged.user_dict['name'] == 'sphere2'
+    assert merged.user_dict['name'] == 'sphere1'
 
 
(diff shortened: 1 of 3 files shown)


tests/core/test_dataobject.py
```diff
@@ -276,7 +276,7 @@ def test_user_dict_persists_with_merge_filter():
     sphere2.user_dict['name'] = 'sphere2'
 
     merged = sphere1 + sphere2
-    assert merged.user_dict['name'] == 'sphere2'
+    assert merged.user_dict['name'] == 'sphere1'
```
tkoyama010, line 279: [ask] So, is this a bug in the test code?

tkoyama010 (approved):
LGTM

tests/core/test_dataobject.py
user27182: Yes. Main here is sphere 1, so its name should have priority

(This bug with the test is causing issues with #7426 and is what led me to investigate this)

PR #7596, opened 2025-06-07
Automatically animate trame viewer

banesullivan:
Sets an async task with trame to automatically update the scene.

This has been a long desired feature and makes it seamless to update data in a background thread or anywhere and have the trame view automatically update (something that has previously had to be explicitly updated in application level code)

### DIFF
--- a/pyvista/trame/jupyter.py
+++ b/pyvista/trame/jupyter.py
@@ -254,6 +254,7 @@ def initialize(
     mode=None,
     default_server_rendering=True,
     collapse_menu=False,
+    animate=True,
     **kwargs,
 ):  # numpydoc ignore=PR01,RT01
     """Generate the UI for a given plotter."""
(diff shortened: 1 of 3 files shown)


banesullivan:
This seems to finally solve issues like https://github.com/pyvista/pyvista/discussions/4006

I'm curious if there is a cost to this and if we should disable this by default. @jourdain, I'm curious if you have any guidance here?

banesullivan:
ah. well as soon as we have more than one plot in a jupyter notebook this freezes up. This is going to need some work...

edit: this is probably because we only use one trame server when in a jupyter notebook

banesullivan:
Closing due to staleness and lack of interest. Trame has some fundamental limitations that make it illsuited to animations. While we could run multiple servers to work around the issues I ran into in this PR, trame has all kinds of issues keeping the renderer in sync and triggering double renders. 

Plus I built a whole new notebook backend from the ground up focused on animation support that I plan to open source later 😎

PR #7603, opened 2025-06-10
Troubleshoot VTK 9.5 plotting test seg faults

user27182:


user27182:
We get seg faults with these versions:

          - vtk-version: "9.4.20250331.dev0"
          - vtk-version: "9.4.20250403.dev0"
          - vtk-version: "9.4.20250405.dev0"
          - vtk-version: "9.4.20250503.dev0"
          
vtk 9.4.2 was released 2025/03/28, so these failures are happening close to the 9.4.2 release...

After a bit of investigating I think that the seg faults are the same ones we saw with #6879, which was tracked down to a VTK issue and patched for `vtk==9.4.2`. But I don't think this patch made its way into the master branch, and so we're seeing seg faults still 

See https://gitlab.kitware.com/vtk/vtk/-/issues/19698

PR #7607, opened 2025-06-11
Skip GC checks for parametrized color tests

user27182:
### Overview

The `check_gc` fixture is expensive (eats up approx 12 mins of test time, see https://github.com/pyvista/pyvista/pull/7441#issuecomment-2960618239). I think we can disable this for ~some test modules~ the parametrized color tests.


tkoyama010 (dismissed):
LGTM

edabor:
Just for more context: the `check_gc` fixture takes more time when all tests (core+plotting) are run at the same time.

When only plotting tests are considered, it takes around 5 minutes.
See https://github.com/pyvista/pyvista/actions/runs/15560420084/job/43811050635#step:10:1810

I think that the fixture is still useful to catch subtle errors. Why disabling-it ?

user27182:
The primary motivation is to reduce test times. 

My understanding is the fixture is used for VTK gc checks, which means the fixture is useless if no vtk objects are involved. So for the test_colors tests, which are testing the `Color` class, I don't think this check is useful.

Similarly, some of the other tests like test_demos are doing more high-level testing, since the demos don't directly interact with vtk objects. And so I think this fixture is better used in the tests for the more primitive PyVista class definitions.

In that sense, I do think this fixture is useful, and should perhaps be enabled more broadly to include some core modules. But, given that we know it's a costly fixture and that test times are a bit long as is, I think we need to be a bit more judicious with where we use it and not blindly enable it everywhere.

tests/plotting/test_colors.py
```diff
@@ -36,6 +36,11 @@
     COLORMAPS.append('batlow')
 
 
+@pytest.fixture(autouse=True)
+def skip_check_gc(skip_check_gc):
+    """GC check not needed."""
```
edabor: Nit suggestion to shorten.
```suggestion
pytestmark = pytest.mark.usefixtures('skip_check_gc')
```

edabor:
Garbage collection still occurs when setting up the test. Adding `check_gc.skip = True` only skip the tear down collection.
https://github.com/pyvista/pyvista/blob/0711dfaaf97bae074d618a367db7f6460c8bd45f/tests/plotting/conftest.py#L48-L68

MatthewFlamm:
I'm not opposed to the goal of tightening scope to reduce testing times, but in general it will be more robust to have this check as wide as possible. Just quickly comparing to recent PRs, the test times are a few minutes faster.  Is it possible rather to optimize the `check_gc` fixture?  Is the time spent in the `gc.check_objects` calls?  Or in the for loop?  If the latter, there may be ways to optimize it to either further reduce time or make it more widely used.

MatthewFlamm:
> Garbage collection still occurs when setting up the test. Adding `check_gc.skip = True` only skip the tear down collection.

Assuming it is the call to `gc.collect` or `gc.collect_objects` that is expensive, it makes sense to prevent set up and tear down.  The answers to this [stack overflow question](https://stackoverflow.com/q/38748257/11423283) have some suggestions to avoid this:

1) use `pytest.mark` and `request`.  This one seems elegant to me.
2) since all the cases in this PR, skip all tests in a module, we can define a new fixture in the module with the same name.

edabor:
+1 for option 1, more elegant indeed

user27182:
> I'm not opposed to the goal of tightening scope to reduce testing times, but in general it will be more robust to have this check as wide as possible. Just quickly comparing to recent PRs, the test times are a few minutes faster. 

This PR mostly stems from looking at the test setup durations linked in https://github.com/pyvista/pyvista/pull/7441#issuecomment-2960618239:

```
  ================================== test setup duration top ==================================
  total          name                                        num   avg            min           
  0:01:06.434628                            test_css4_colors   148 0:00:00.443159 0:00:00.415114
  0:00:32.683528                      test_direction_objects    70 0:00:00.464522 0:00:00.447222
  0:00:27.858993                             test_vtk_colors    62 0:00:00.442998 0:00:00.417892
  0:00:09.414953                             test_theme_parm    20 0:00:00.468429 0:00:00.452478
  0:00:09.251169                test_plot_show_bounds_params    20 0:00:00.458454 0:00:00.443200
  0:00:09.148164                         test_color_synonyms  
```

There are similar numbers for teardown as well. So we can shed almost 4 minutes of test time if we skip gc checks for the parametrized color tests alone. I'm fine with restricting the scope of this PR to only skip gc for these. We can maybe skip `test_theme_parm` as well.

user27182:
> 1. use `pytest.mark` and `request`.  This one seems elegant to me.

This is done in #7611. This PR uses the new marker, and so it's marked as a draft for now until that merges.

PR #7608, opened 2025-06-12
Add pass_cell_data kwarg to cell_centers

akaszynski:
Add `pass_cell_data` to `cell_centers` filter. Usage:

# default (current) behavior

```py
>>> import pyvista as pv
>>> cube = pv.Cube()
>>> cube.cell_centers().cell_data
>>> cube.cell_centers().cell_data
pyvista DataSetAttributes
Association     : CELL
Active Scalars  : None
Active Vectors  : None
Active Texture  : None
Active Normals  : None
Contains arrays :
    FaceIndex               int64      (6,)
```

Now, with `pass_cell_data` disabled:

```py
>>> cube.cell_centers(pass_cell_data=False).cell_data
pyvista DataSetAttributes
Association     : CELL
Active Scalars  : None
Active Vectors  : None
Active Texture  : None
Active Normals  : None
Contains arrays : None
```

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -2319,7 +2319,10 @@ class DataObjectFilters:
         return _get_output(alg)
 
     def cell_centers(  # type: ignore[misc]
-        self: _DataSetOrMultiBlockType, vertex: bool = True, progress_bar: bool = False
+        self: _DataSetOrMultiBlockType,
+        vertex: bool = True,
+        pass_cell_data: bool = True,
+        progress_bar: bool = False,
     ):
         """Generate points at the center of the cells in this dataset.
 
@@ -2329,6 +2332,9 @@ class DataObjectFilters:
         ----------
         vertex : bool, default: True
             Enable or disable the generation of vertex cells.
+
+        pass_cell_data : bool, default: True
+            If enabled, pass the input cell data through to the output.
 
         progress_bar : bool, default: False
             Display a progress bar to indicate progress.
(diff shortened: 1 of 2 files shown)


pyvista/core/filters/data_object.py
```diff
@@ -2319,7 +2319,10 @@ def compute_cell_sizes(  # type: ignore[misc]
         return _get_output(alg)
 
     def cell_centers(  # type: ignore[misc]
-        self: _DataSetOrMultiBlockType, vertex: bool = True, progress_bar: bool = False
+        self: _DataSetOrMultiBlockType,
+        vertex: bool = True,
+        pass_cell_data: bool = True,
```
user27182, line 2324: These args are positional so this is technically a breaking change, though we've approved similar PRs before. Please consider reviewing or commenting on #7581 so we can make these keyword-only once and for all.

akaszynski: Reviewed and commented. Feel free to ping me anytime! I get a ton of github notifications and I'll prioritize @.

akaszynski: I don't think many use `progress_bar`, so while it's a breaking change I'm less than concerned. #7581 will help in that regard.

user27182: > I don't think many use `progress_bar`, so while it's a breaking change I'm less than concerned.

Agreed, it's even less likely that it's used like `cell_centers(True, True)`, so not a concern.

MatthewFlamm (approved):

user27182:
Looks like the VTK dev test has been marked as required. This won't execute without the `vtk-dev-testing` label, and even if it does, the tests will fail, so this shouldn't be required yet IMO if we want to merge with green CI

akaszynski:
> ...this shouldn't be required yet IMO if we want to merge with green CI

image
image

tkoyama010:
> Looks like the VTK dev test has been marked as required. This won't execute without the `vtk-dev-testing` label, and even if it does, the tests will fail, so this shouldn't be required yet IMO if we want to merge with green CI

Removed from required.

PR #7609, opened 2025-06-12
Add annotations to `cells.py`

user27182:


tests/core/test_grid.py
```diff
@@ -345,7 +345,7 @@ def test_cells_dict_variable_length():
 
 def test_cells_dict_empty_grid():
     grid = pv.UnstructuredGrid()
-    assert grid.cells_dict is None
+    assert grid.cells_dict == {}
```
user27182: Does this change need to be documented? This technically breaks `if grid.cells_dict is None:` checks, but doesn't break more general bool checks `if grid.cells_dict:`

We can also revert this and keep `None`, but then we need to add `| None` everywhere and do explicit `if grid.cells_dict is not None:` checks to do type-narrowing

MatthewFlamm: It is good to simplify typing overall.  Would be good to document in some way so that if someone runs into the issue they can find the solution.  I would put in a breaking change label here with a description of the change in the PR description so it is findable.  Also consider a `.. versionchanged` in documentation.

user27182: This change is made explicit in #7626

MatthewFlamm: Good idea to separate out the breaking change the this can be only typing

PR #7610, opened 2025-06-12
Run `test_charts`.py in a separate process and increase `xvfb` max clients

user27182:
### Overview

Duplicate of #7214 as an interim workaround to https://gitlab.kitware.com/vtk/vtk/-/issues/19698 to get our tests passing for VTK 9.5


user27182:
Seg faults are gone, we are finally seeing proper plotting test failures

PR #7611, opened 2025-06-12
Replace `skip_check_gc` fixture with `skip_check_gc` marker

user27182:
### Overview

In https://github.com/pyvista/pyvista/pull/7607#issuecomment-2965580101 it was found that `skip_check_gc` only skips checking gc during teardown, and does not skip gc during test setup.

This PR fixes this by using a marker instead of a fixture.


MatthewFlamm (approved):
LGTM

user27182:
The trame tests needed more changes than I expected because it defined `skip_check_gc` in its own conftest

MatthewFlamm (approved):

PR #7613, opened 2025-06-13
Fix an invalid `if` key causing warnings during workflow execution

kurtmckee:
### Overview

This PR resolves warnings in GitHub workflow runs that are getting thrown because the `if` key is invalid in the location it's being used [[recent example](https://github.com/pyvista/pyvista/actions/runs/15077885943)]:

> image

The strategy used here is a ternary operator to set the caching to "pip" if `USE_CACHE` is set to "true".

_Previously opened in #7530_

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -59,8 +59,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.config.python-version }}
-          if: env.USE_CACHE == 'true'
-          cache: "pip"
+          cache: ${{ env.USE_CACHE == 'true' && 'pip' || '' }}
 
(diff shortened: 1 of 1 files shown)


user27182 (approved):
Thanks!

Maintainers: let's hold off merging with main until we have all required approvals to reduce notification noise

banesullivan (approved):
Okay, for some odd reason I didn't understand this the first time I looked at it.

Thanks for fixing/cleaning this up and sorry for my delay here considering workflow PRs are blocked by my review

kurtmckee:
No worries about delays -- I also contribute to and maintain a number of projects. It's not my day job, and for me it's just fun! I simply get what feels like a lot of notifications and sometimes take action to reduce the quantity of those notifications.

Thank you all for your work on pyvista!

PR #7617, opened 2025-06-15
Enable `allow_useless_fixture` option for `pytest-pyvista`

user27182:
### Overview

It's easy to forget to call `show` or `plot` when using the `verify_image_cache` fixture. Enabling this option will fail a test if the fixture is used but no images were generated, unless `verify_image_cache.expect_plot=False` is explicitly set.

See https://github.com/pyvista/pytest-pyvista/pull/170


user27182:
Superseded by #7659

PR #7621, opened 2025-06-16
Fix `RuntimeError` raised when disabling ssao on a closed plotter

user27182:
### Overview

Related to https://github.com/pyvista/pytest-pyvista/issues/172, this PR removes the use of `auto_close`.
And fixes the expected behavior of raising a RuntimeError when disabling ssao on an already-closed plotter.


tkoyama010 (dismissed):
This is a nice idea! Thanks.

user27182:
Original image was passing on MacOS, failing on Linux. And now we have the reverse. The error is approx 1200, just outside the 1000 limit for flagging this as variance. Should we keep the new image and skip tests for Mac ?

PR #7616, opened 2025-06-14
Update tests to remove unused cache images with `pytest-pyvista`

user27182:
### Overview

Fix #7582

PR #7622, opened 2025-06-16
Remove test skips for `set_environment_texture_cubemap`

user27182:
### Overview

As part of #7616 and #7617, it was found that this test has:

1. An unused cached image (`set_environment_texture_cubemap.png`)
2. Does not generate any images

The cause of this is #7357, which added a skip for `vtk>=9.1` without realizing that there was already a separate skip for `vtk<9.1` in the test body. This made it so that this test is always skipped.

This PR re-enables the test and updates it for more recent vtk versions.


tests/plotting/image_cache/set_environment_texture_cubemap[False].png
Comment (tkoyama010) [inline tests/plotting/image_cache/set_environment_texture_cubemap[False].png:1]: This may not be directly related to this PR, but the True and False images look identical to me. If there is a difference, I think it would be helpful to make it more visually distinguishable.

Comment (user27182) [reply tests/plotting/image_cache/set_environment_texture_cubemap[False].png:1]: Yes the change is very subtle. Will need to use a different texture I think.

tests/plotting/test_plotting.py
```diff
@@ -302,22 +302,21 @@ def test_pbr(sphere, verify_image_cache):
...
+@pytest.mark.needs_vtk_version(9, 2)
+def test_set_environment_texture_cubemap(resample, verify_image_cache):
     """Test set_environment_texture with a cubemap."""
-    verify_image_cache.high_variance_test = True
-
-    texture = examples.download_sky_box_cube_map()
+    if resample:
+        # Skip due to large variance
+        verify_image_cache.windows_skip_image_cache = True
+        verify_image_cache.macos_skip_image_cache = True
```
tkoyama010: Since `resample` is always True, the `if` block can be removed.
```suggestion
    # Skip due to large variance
    verify_image_cache.windows_skip_image_cache = True
    verify_image_cache.macos_skip_image_cache = True
```

tkoyama010 (changes requested):
Thanks for improving the test!

tkoyama010 (approved):
LGTM

PR #7623, opened 2025-06-17
Move local import in `deprecate_positional_args` to avoid erroneous errors

user27182:
### Overview
When running something like `pytest -k set_environment_texture` where the test image is missing from the cache, I get `FileNotFound` errors as expected, but I get errors like these, which are not related to the actual errors:

``` python
  File "pyvista/pyvista/plotting/plotter.py", line 6627, in __del__
  File "pyvista/pyvista/plotting/plotter.py", line 5141, in deep_clean
  File "pyvista/pyvista/plotting/picking.py", line 765, in disable_picking
  File "pyvista/pyvista/_deprecate_positional_args.py", line 191, in inner_f
ImportError: sys.meta_path is None, Python is likely shutting down
Exception ignored in: <function Renderer.__del__ at 0x13e0f84a0>
  File "pyvista/pyvista/plotting/renderer.py", line 3882, in __del__
  File "pyvista/pyvista/_deprecate_positional_args.py", line 191, in inner_f
ImportError: sys.meta_path is None, Python is likely shutting down
Exception ignored in: <function Renderer.__del__ at 0x13e0f84a0>
  File "pyvista/pyvista/plotting/renderer.py", line 3882, in __del__
  File "pyvista/pyvista/_deprecate_positional_args.py", line 191, in inner_f
ImportError: sys.meta_path is None, Python is likely shutting down
```

These errors are similar to those from https://github.com/pyvista/pyvista/issues/7503. This PR fixes this by moving the local import which is causing this.


tkoyama010 (approved):
LGTM

PR #7624, opened 2025-06-17
Use `pytest-xdist` in unit testing

user27182:
### Overview

`pytest-xdist` is included in the test dependencies, but it's not currently used. This PR enables its use.

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -35,7 +35,7 @@ env:
   PYVISTA_USERDATA_PATH: ${{ github.workspace }}/vtk-data
   LINUX_PLOTTING_ARGS: >
     -v --cov=pyvista --cov-append --cov-report=xml --cov-branch
-    --fail_extra_image_cache --generated_image_dir debug_images
+    --fail_extra_image_cache --generated_image_dir debug_images -n auto
   PLOTTING_TEST_MODULES_CHARTS: "tests/plotting/test_charts.py"
   PLOTTING_TEST_MODULES_NO_CHARTS: >
     --ignore=tests/core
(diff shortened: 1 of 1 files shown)


MatthewFlamm:
Looks like there are problems with trying to simultaneously handle the same file when downloading it or transferring it from the local vtk-data repo clone.  We could consider using `--dist loadgroup` and manually marking all tests that use download data to be in the same group (or split them up into multiple groups).  Could be annoying to maintain, but the speedups look promising.

Also getting errors about the naming/ordering of tests being consistent in some cases.

user27182:
> Looks like there are problems with trying to simultaneously handle the same file when downloading it or transferring it from the local vtk-data repo clone.  We could consider using `--dist loadgroup` and manually marking all tests that use download data to be in the same group (or split them up into multiple groups).  Could be annoying to maintain, but the speedups look promising.

The speedups do look promising. We could also try directly adding a hook to the dataset loader class, and always make a temp copy of the files before loading and use a CI-specific flag to control this.


> Also getting errors about the naming/ordering of tests being consistent in some cases.

In the docs I did see that using parametrize with sets can be problematic, might need to look over some of these and use lists instead

user27182:
We're getting some mysterious crashes:
https://github.com/pyvista/pyvista/actions/runs/15746300270/job/44383463654?pr=7624#step:15:3765
``` bash
[gw0] linux -- Python 3.9.22 /opt/hostedtoolcache/Python/3.9.22/x64/bin/python
worker 'gw0' crashed while running 'tests/plotting/jupyter/test_trame.py::test_trame_plotter_ui[vue2]'
```

https://github.com/pyvista/pyvista/actions/runs/15746300270/job/44383463660?pr=7624#step:15:3807
``` bash
[gw0] linux -- Python 3.9.22 /opt/hostedtoolcache/Python/3.9.22/x64/bin/python
worker 'gw0' crashed while running 'tests/plotting/test_plotting.py::test_plot_depth_peeling'
```
Not sure if these are flaky or what the cause is. `test_plot_depth_peeling` uses `download_airplane` but `test_trame_plotter_ui[vue2]` does not appear to download anything

user27182:
> Not sure if these are flaky or what the cause is. `test_plot_depth_peeling` uses `download_airplane` but `test_trame_plotter_ui[vue2]` does not appear to download anything

These are flaky, tests pass after restarting. Will try `--dist loadgroup` for some test modules.

user27182:
Plotting tests seem to be working now without issue... will enable core tests and add xdist groups

user27182:
To use `pytest-xdist`, it seems we'll need to refactor the test fixtures, e.g. https://github.com/pyvista/pyvista/pull/7633. If there is limited appetite or availability to support making these kinds of changes then we can probably close this PR.

PR #7626, opened 2025-06-19
Return `{}` instead of `None` for empty `cells_dict` and `get_mixed_cells`

user27182:
### Overview

This change simplifies the typing from #7609, and I think it makes sense that an empty mesh returns an empty dict here.


MatthewFlamm (approved):

tkoyama010 (approved):
LGTM

PR #7628, opened 2025-06-19
Mark `contour_labels` strict external test as flaky

user27182:
### Overview

This test sometimes fails with CI since it is more of a performance test and the execution times can vary.


tkoyama010 (approved):
LGTM

PR #7633, opened 2025-06-20
Use fixtures with `test_meshio.py`

user27182:
### Overview

Using fixtures provides better test isolation from test setup and will assist with parallel processing for #7624


tkoyama010 (approved):
LGTM

PR #7638, opened 2025-06-21
Update zizmor pre-commit

bjlittle:
### Overview

This pull-request updates the `zizmor` pre-commit hook to the latest available version and repo URL.

It also addresses subsequent non-compliant behaviour reported by the latest version, all of which relate to using versions rather than commit SHAs for third-party GHAs. For further information see,

- [Using third-party actions](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions?learn=getting_started#using-third-party-actions)
- `zizmor` audit rule [unpinned-uses](https://docs.zizmor.sh/audits/#unpinned-uses)

:question: Also, is there a specific reason why `pyvista` isn't using `dependabot` to automatically update the `github-actions` package ecosystem? This would result in similar updates from `dependabot` akin to the `pip` package ecosystem which `pyvista` already has configured. I'd be happy to do that, if you wanted?


.github/zizmor.yml
```diff
@@ -0,0 +1,5 @@
+rules:
+  unpinned-uses:
+    ignore:
+      - style_docstr.yml:78:9
```
bjlittle: This exception ignores the following audit rule error:

```
error[unpinned-uses]: unpinned action reference
  --> .github/workflows/style_docstr.yml:78:9
   |
78 |         uses: errata-ai/vale-action@reviewdog
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
   |
   = note: audit confidence → High
```

The offending [errata-ai/vale-action](https://github.com/errata-ai/vale-action) GHA appears to be referencing the `reviewdog` branch rather than a released version or tag ... so didn't quite know what to do about this. The alternative would be to pin to a commit SHA on that branch?

.github/workflows/integration-tests.yml
```diff
@@ -85,7 +85,7 @@ jobs:
         with:
           python-version: "3.12"
       - run: git clone https://github.com/pyvista/pyvistaqt.git --single-branch
-      - uses: pyvista/setup-headless-display-action@main
+      - uses: pyvista/setup-headless-display-action@7d84ae825e6d9297a8e99bdbbae20d1b919a0b19
```
bjlittle, line 88: Were these GHAs intentionally referencing `main` for bleeding edge and unreleased commits?

bjlittle:
pre-commit.ci autofix

.github/zizmor.yml
MatthewFlamm: `reviewdog` is the default branch, AKA it is equivalent to `main`  So we should follow the same practice to pin to commit.

bjlittle: @MatthewFlamm Done :+1: 

I've pinned this to the latest commit SHA on the `reviewdog` branch.

MatthewFlamm:
> ❓ Also, is there a specific reason why `pyvista` isn't using `dependabot` to automatically update the `github-actions` package ecosystem?

I don't know a reason not to do this, but it should be a separate PR either way.

bjlittle:
> > ❓ Also, is there a specific reason why `pyvista` isn't using `dependabot` to automatically update the `github-actions` package ecosystem?
> 
> I don't know a reason not to do this, but it should be a separate PR either way.

Awesome, thanks.

Once this PR lands I'll follow it up with the `dependabot` support in a separate PR 👍

tkoyama010 (approved):

Issue #7634, opened 2025-06-20
Loaded dataset returns different mesh when read multiple times


user27182:
### Describe the bug, what's wrong, and what you expected.

Loading the same dataset multiple times should yield the same dataset every time. But this is not the case for some datasets, e.g. with `download_biplane`:
``` python
import pyvista as pv
biplane1 = pv.examples.download_biplane()
biplane2 = pv.examples.download_biplane()
biplane3 = pv.examples.download_biplane()
print(biplane1 == biplane2) # False
print(biplane2 == biplane3) # False
print(biplane1 == biplane3) # False
```
I thought it might be caused by the new exodus reader, so I tried the "old" one, and same issue:

``` python
exo = pv.examples.download_biplane(load=False)
biplane1 = pv.read_exodus(exo)
biplane2 = pv.read_exodus(exo)
biplane3 = pv.read_exodus(exo)

print(biplane1 == biplane2) # False
print(biplane2 == biplane3) # False
print(biplane1 == biplane3) # False
```

Or with the `3gqp` dataset, every time I load it max value of one of its arrays is different, ranging from 140-255:

``` python
import pyvista as pv;
mesh1 = pv.examples.download_3gqp()
mesh2 = pv.examples.download_3gqp()
mesh3 = pv.examples.download_3gqp()
mesh4 = pv.examples.download_3gqp()

key = 'secondary_structures_begin'
print(mesh1[key].max())  # 255
print(mesh2[key].max())  # 158
print(mesh3[key].max())  # 146
print(mesh4[key].max())  # 254
```

This issue seems to exists for:
- `download_biplane`
- `download_damavand_volcano`
- `download_embryo`
- `download_notch_stress`
- `download_particles`

Could be a VTK bug with the respective readers...

### Steps to reproduce the bug.

See above

### System Information

```shell
OS : Darwin (macOS 14.4.1)
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
               RAM : 16.0 GiB
       Environment : IPython
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M2
       GPU Version : 4.1 Metal - 88
     Render Window : vtkCocoaRenderWindow
  MathText Support : True
  Python 3.12.2 (v3.12.2:6abddd9f6a, Feb  6 2024, 17:02:06) [Clang 13.0.0
  (clang-1300.0.29.30)]
           pyvista : 0.46.dev0
               vtk : 9.4.2
[...]

PR #7641, opened 2025-06-23
Allow NaN values with data object equality checks

user27182:
### Overview

Partial fix for #7634 

The equality check failures from `download_damavand_volcano` and `download_notch_stress` are due to `nan` values in arrays, and the failure from `download_particles` is caused by `nan` values in the points.

### DIFF
--- a/pyvista/core/datasetattributes.py
+++ b/pyvista/core/datasetattributes.py
@@ -1329,6 +1329,14 @@ class DataSetAttributes(_vtk.VTKObjectWrapper):
 
     def __eq__(self: Self, other: object) -> bool:
         """Test dict-like equivalency."""
+
+        def array_equal_nan(array1: npt.ArrayLike, array2: npt.ArrayLike) -> bool:
+            return (
+                np.issubdtype(np.asanyarray(array1).dtype, np.floating)
+                and np.issubdtype(np.asanyarray(array2).dtype, np.floating)
+                and np.array_equal(array1, array2, equal_nan=True)
+            )
+
         # here we check if other is the same class or a subclass of self.
         if not isinstance(other, type(self)):
             return False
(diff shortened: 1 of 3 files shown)


pyvista/core/datasetattributes.py
```diff
@@ -1338,7 +1346,7 @@ def __eq__(self: Self, other: object) -> bool:
 
         # verify the value of the arrays
         for key, value in other.items():
-            if not np.array_equal(value, self[key]):
+            if not np.array_equal(value, self[key]) and not array_equal_nan(value, self[key]):
```
MatthewFlamm, line 1351: Why can we not use the same form as in `dataobject` here?

```suggestion
            if not np.array_equal(value, self[key], equal_nan=True):
```

user27182: I tried that at first but this was failing for non-numeric arrays. As in, numpy doesn't seem to handle string arrays gracefully with the equal_nan option. I can test this again locally though.

MatthewFlamm: Then I will ask the inverse question :), why do we not need to use this form in `dataobject`?

user27182: One comparison is for point/ cell/field data arrays, where we might expect string arrays. The other comparison is for points, cells, and other numeric attributes, where a string array doesn't make sense.

user27182: BTW the issue with strings and `equal_nan` is tracked here https://github.com/numpy/numpy/issues/16377

MatthewFlamm: Makes a lot of sense.  Thanks!  After expanding the code in `dataobject` it is limited to only a few attrs which are all expected to be numeric.

MatthewFlamm (approved):
LGTM

pyvista/core/datasetattributes.py
MatthewFlamm: I had already approved, but this might be worth commenting in the code so we know why we have this complex form.  Maybe linked to the numpy issue.

MatthewFlamm (approved):
Thanks!

PR #7642, opened 2025-06-26
Fix `Plotter.add_scalar_bar` wrapper args and kwargs

### LINKED ISSUE #7632: add_scalar_bar not using mapper as expected <- documentation issue

Not sure it fully qualifies as a bug, in the end a got where the problem comes from and its maybe more about the documentation.

I was having trouble with add_scalar_bar not using the mapper I was passing, but using the last object instead.
(Note: I know add_scalar_bar is not needed if adding scalar directly as .point_data["something"] = , but I am not using this option on purpose, cf. [this](https://github.com/pyvista/pyvista/discussions/7173#discussioncomment-12364164) )
In fact, it seems to be just because I was not passing the mapper as a keyword parameter but just as the second unnamed parameter.

I know it's probably a better idea to use keyword parameters, but the documentation was giving no clue that this would break the behaviour. In the documentation the parameters are namely listed in the signature, but the actual signature is `def add_scalar_bar(self, *args, **kwargs)` and the mapper is fetched by `mapper = kwargs.get('mapper', None)`

I don't know how you'd want to handle this, but I thought I should still report it in case someone falls into similar trouble.

import pyvista as pv
import numpy as np

# create object and plotter
obj = pv.Sphere()
plotter = pv.Plotter()
actor = plotter.add_mesh(obj)
plotter.show()

# create a scalar array
# It is intentionnal to create it this way to avoid showing it at this stage
scalar_name = "test"
obj.point_data.set_array(np.random.rand(obj.n_points), scalar_name)

# something else would happen in the code including user interaction
# until then, if I show the scalar field, everything is as expected

# create and add a new object
obj2 = pv.Rectangle()
obj2.translate((1,0,0))
actor2 = plotter.add_mesh(obj2)

# After adding a new object, when I activate the scalar field,
# the colorbar is the on of the added object and not the selected one (ie., obj)
scalar_values = obj.point_data[scalar_name] 
vmin = np.min(scalar_values)
vmax = np.max(scalar_values)
cmap = "viridis"
obj.set_active_scalars(scalar_name)
mapper = actor.mapper
mapper.scalar_visibility = True
mapper.array_name = scalar_name
mapper.scalar_range = vmin, vmax
mapper.lookup_table.cmap = cmap

plotter.add_scalar_bar(scalar_name, 
                actor.mapper,  # <- here, giving the mapper this way just ignores it
                # mapper= actor.mapper,   # <- this fixes the problem
                interactive= True
                )

### System Information

```shell
  Date: Fri Jun 20 11:30:46 2025 Paris, Madrid (heure d’été)

                OS : Windows (10 10.0.19045 SP0 Multiprocessor Free)
            CPU(s) : 8
           Machine : AMD64
      Architecture : 64bit
               RAM : 15.9 GiB
       Environment : Jupyter
        GPU Vendor : Intel
      GPU Renderer : Intel(R) UHD Graphics 620
       GPU Version : 4.5.0 - Build 30.0.101.1338
  MathText Support : False

  Python 3.11.9 | packaged by conda-forge | (main, Apr 19 2024, 18:27:10) [MSC
  v.1938 64 bit (AMD64)]

           pyvista : 0.44.0
               vtk : 9.3.1
             numpy : 2.0.0
        matplotlib : 3.9.1
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 10.4.0
           IPython : 8.26.0
          colorcet : 3.1.0
        ipywidgets : 8.1.3
             scipy : 1.14.0
            meshio : 5.3.5
        jupyterlab : 4.2.3
             trame : 3.6.3
      trame_client : 3.2.1
      trame_server : 3.0.3
         trame_vtk : 2.8.9
     trame_vuetify : 2.6.1
      nest_asyncio : 1.6.0
```

### Screenshots

Image

user27182:
### Overview

Fix #7632

This wrapper does not follow the typical wrapping pattern:
``` python
@wraps(ScalarBars.add_scalar_bar)
def add_scalar_bar(*args, **kwargs):
    return self.scalar_bars.add_scalar_bar(*args, **kwargs)
```

and instead does a bunch of custom handling of `*args` and `**kwargs`. But it assumes that the only arg possible is `title`, creating a silent bug where any other args are simply ignored if they are use positionally (e.g. `mapper`).

Instead of a silent bug, this PR makes it so that an error is raised instead if anything other than `title` is used positionally. It was also observed that the `interactive` kwarg is also not forwarded at all (and is also ignored/unused), so this PR fixes this too.

This would normally be a breaking change, but since this function is already broken, I wouldn't consider it as such, and instead I think this is a simple bug fix.

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -4912,7 +4912,7 @@ class BasePlotter(PickingHelper, WidgetHelper):
 
     @wraps(ScalarBars.add_scalar_bar)
     def add_scalar_bar(
-        self, *args, **kwargs
+        self, title: str = '', **kwargs
     ) -> _vtk.vtkScalarBarActor:  # numpydoc ignore=PR01,RT01
         """Wrap for ``ScalarBars.add_scalar_bar``."""
         # only render when the plotter has already been shown
@@ -4928,23 +4928,18 @@ class BasePlotter(PickingHelper, WidgetHelper):
                 raise AttributeError(msg)
             kwargs['mapper'] = self.mapper
 
-        # title can be the first and only arg
-        title = args[0] if args else kwargs.get('title', '')
-        if title is None:
-            title = ''
-        kwargs['title'] = title
-
         interactive = kwargs.get('interactive')
         if interactive is None:
             interactive = self._theme.interactive
             if self.shape != (1, 1):
                 interactive = False
+            kwargs['interactive'] = interactive
         elif interactive and self.shape != (1, 1):
             msg = 'Interactive scalar bars disabled for multi-renderer plots'
             raise ValueError(msg)
         # by default, use the plotter local theme
         kwargs.setdefault('theme', self._theme)
-        return self.scalar_bars.add_scalar_bar(**kwargs)
+        return self.scalar_bars.add_scalar_bar(title, **kwargs)
 
     @_deprecate_positional_args(allowed=['scalars'])
     def update_scalars(self, scalars, mesh=None, render: bool = True) -> None:  # noqa: ANN001, FBT001, FBT002


MatthewFlamm (approved):
Since we are already deprecating many positional args, I agree with the decision to more rapidly discontinue it's use completely here to fix this bug.  I added the breaking change label although I'm not 100% sure whether someone could be using positional args right now.

PR #7645, opened 2025-06-28
Add `crop` filter for `ImageData`

user27182:


pyvista/core/filters/image_data.py
```diff
@@ -272,17 +285,379 @@ def extract_subset(  # noqa: PLR0917
...
-        fixed.point_data.update(result.point_data)
-        fixed.cell_data.update(result.cell_data)
-        fixed.field_data.update(result.field_data)
-        fixed.copy_meta_from(result, deep=True)
-        return fixed
+        if modify_geometry:
+            # Adjust for the confusing issue with the extents
+            #   see https://gitlab.kitware.com/vtk/vtk/-/issues/17938
+            result.origin = result.bounds[::2]
+            result.offset = (0, 0, 0)
```
user27182: The crop filter requires vtkExtractVOI, so I opted to reuse the implementation from `extract_subset`. However, I do not agree with this "fix" (it's more of a misunderstanding/preference) and i think it's better not to modify the geometry as part of the crop filter. And so, this option is added here so it can be set to False when cropping. 

I can create a separate PR for this.

pyvista/core/filters/image_data.py
```diff
@@ -2504,6 +2840,7 @@ def _get_num_components(array_):
                     f"match the number components ({num_input_components}) in array '{scalars}'."
                 )
                 raise ValueError(msg)
+            val = np.broadcast_to(val, (num_input_components,))
```
user27182: This is a separate bug fix discovered while writing the examples for crop. I can make a separate PR for this.

pyvista/core/filters/image_data.py
user27182: Use of this keyword is avoided in [9d11557](https://github.com/pyvista/pyvista/pull/7645/commits/9d115571ccadd6dbef9e11ccac0cb9797d069c94), so this PR no longer depends on #7647

pyvista/core/filters/image_data.py
```diff
@@ -284,6 +280,507 @@ def extract_subset(  # noqa: PLR0917
...
+        >>> gray_image = examples.download_yinyang()
+        >>> gray_image.dimensions
+        (512, 342, 1)
+
+        Define a custom plotting helper to show the image as pixel cells.
+
+        >>> def image_plotter(image):
+        ...     pixel_cells = image.points_to_cells()
+        ...
+        ...     pl = pv.Plotter(off_screen=False)
```
user27182:
```suggestion
        ...     pl = pv.Plotter()
```

tkoyama010 (approved):
LGTM

PR #7646, opened 2025-06-29
Fix padding RGB image with non-RGB value

user27182:
### Overview

Add fix for bug mentioned here https://github.com/pyvista/pyvista/pull/7645#discussion_r2173792486

To reproduce:
``` python
from pyvista import examples
im = examples.download_beach()
im.pad_image(0)
```
```
  File "pyvista/pyvista/core/filters/image_data.py", line 2561, in _get_padded_output
    alg.SetConstant(val[component])  # type: ignore[attr-defined]
                    ~~~^^^^^^^^^^^
IndexError: index 1 is out of bounds for axis 0 with size 1
```


tkoyama010 (approved):
LGTM

PR #7647, opened 2025-06-29
Add `reset_origin` option to `extract_subset` filter

user27182:
### Overview

Add option to keep original values  from the input when using `extract_subset`. See https://github.com/pyvista/pyvista/pull/7645#discussion_r2173791956

### DIFF
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -224,6 +224,7 @@ class ImageDataFilters(DataSetFilters):
         voi,
         rate=(1, 1, 1),
         boundary: bool = False,  # noqa: FBT001, FBT002
+        reset_origin: bool = True,  # noqa: FBT001, FBT002
         progress_bar: bool = False,  # noqa: FBT001, FBT002
     ):
         """Select piece (e.g., volume of interest).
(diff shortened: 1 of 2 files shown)


pyvista/core/filters/image_data.py
```diff
@@ -256,6 +257,18 @@ def extract_subset(  # noqa: PLR0917
             even multiple of the grid dimensions. By default this is
             disabled.
 
+        reset_origin : bool, default: True
```
user27182: Maybe call it `move_origin`? Changing the origin is like applying a translation in xyz space, and removing the offset is like applying the inverse in ijk space. So the physical bounds should be the same either way. 

Or we call it `mode`, with the option for `absolute` (status quo with the "fix" applied) and 
`relative` (where no changes are made to the output)

user27182: Or call it `coordinate_frame` and use `physical` and `index` as the options. This would be consistent with properties like `index_to_physical_matrix` https://docs.pyvista.org/api/core/_autosummary/pyvista.imagedata.index_to_physical_matrix#pyvista.ImageData.index_to_physical_matrix

tkoyama010 (approved):
LGTM

PR #7652, opened 2025-07-02
Add `slice_index` filter and `__getitem__` method for indexing `ImageData`

user27182:
### Overview

E.g. allow:
``` py
import pyvista as pv
pv.Wavelet().slice_index(z=10)
# Or, equivalently:
pv.Wavelet()[:, :, 10]
```

to extract an XY slice at index 10 along the z axis


pyvista/core/filters/image_data.py
```diff
@@ -264,14 +412,12 @@ def extract_subset(  # noqa: PLR0917
...
+
         """
-        alg = _vtk.vtkExtractVOI()
-        alg.SetVOI(voi)
-        alg.SetInputDataObject(self)
-        alg.SetSampleRate(rate)
-        alg.SetIncludeBoundary(boundary)
-        _update_alg(alg, progress_bar=progress_bar, message='Extracting Subset')
-        result = _get_output(alg)
+        result = self._extract_voi(voi, rate=rate, boundary=boundary, progress_bar=progress_bar)
```
user27182: The `slice_index` filter (this PR) and `crop` filter (from #7645) both make use of `vtkExtractVOI` from the `extract_subset` filter. It would probably be best to resolve #7647 first, and then we can add a `rebase_coordinates` keyword to both `slice_index` and `crop` and forward it to `extract_subset`.

tkoyama010 (approved):
LGTM. Awesome feature!

PR #7653, opened 2025-07-04
Fix crash with `contour_labels` when there is no foreground

user27182:
### Overview

While working on an example for #7652 I encountered a bug where `contour_labels` crashes python in some cases where there is no foreground to contour.

``` py
from pyvista import examples

mesh = examples.load_frog_tissues()
subset = mesh.extract_subset((10, 100, 20, 200, 20, 80))
surface = subset.contour_labels()
```

### DIFF
--- a/tests/core/test_imagedata_filters.py
+++ b/tests/core/test_imagedata_filters.py
@@ -6,6 +6,7 @@ import time
 
 import numpy as np
 import pytest
+import vtk
 
 import pyvista as pv
 from pyvista import examples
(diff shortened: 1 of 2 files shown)


tests/core/test_imagedata_filters.py
```diff
@@ -414,6 +414,15 @@ def test_contour_labels_raises_vtkversionerror():
         pv.ImageData().contour_labels()
 
 
+@pytest.mark.needs_vtk_version(9, 3, 0)
+def test_contour_labels_empty_input(frog_tissues):
+    voi = frog_tissues.extract_subset((10, 100, 20, 200, 20, 80))
+    background_value = 0
+    assert np.all(voi.active_scalars == background_value)
```
tkoyama010: Question: Do we need to use `allclose` here to account for floating-point errors?

user27182: Might be best to use all close. The typical use case is to contour labeled data, which is normally integers, but the filter should work with floats too. Will push an update.

tkoyama010 (approved):
LGTM

PR #7656, opened 2025-07-04
Fix `copy_vtk_array` with `vtkImplicitArray` inputs

### LINKED ISSUE #7655: copy_vtk_array yielding a TypeError

Looks like there is an issue adding certain types of arrays leading to a type error on copy. Don't have a lot of time to dig into this

```python
>>> import pyvista as pv
>>> pine_roots = examples.download_pine_roots()
>>> conn = pine_roots.connectivity()
>>> pine_roots["test"] = conn["RegionId"]
  File "/Users/banesullivan/Software/pyvista/pyvista/pyvista/core/dataset.py", line 1543, in __setitem__
    self.cell_data[name] = scalars
    ~~~~~~~~~~~~~~^^^^^^
  File "/Users/banesullivan/Software/pyvista/pyvista/pyvista/core/datasetattributes.py", line 260, in __setitem__
    self.set_array(value, name=key)
  File "/Users/banesullivan/Software/pyvista/pyvista/pyvista/_deprecate_positional_args.py", line 241, in inner_f
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/Users/banesullivan/Software/pyvista/pyvista/pyvista/core/datasetattributes.py", line 631, in set_array
    vtk_arr = self._prepare_array(data=data, name=name, deep_copy=deep_copy)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/banesullivan/Software/pyvista/pyvista/pyvista/core/datasetattributes.py", line 835, in _prepare_array
    vtk_arr = copy_vtk_array(data.VTKObject, deep=deep_copy)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/banesullivan/Software/pyvista/pyvista/pyvista/_deprecate_positional_args.py", line 241, in inner_f
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/Users/banesullivan/Software/pyvista/pyvista/pyvista/core/utilities/arrays.py", line 205, in copy_vtk_array
    new_array = type(array)()
                ^^^^^^^^^^^^^
TypeError: this is an abstract class and cannot be instantiated
>>> 
```

### System Information

```shell
I think this works on VTK 9.3 but fails on VTK 9.4 for `main`
```

### Screenshots

_No response_

user27182:
### Overview

Fix #7655

It is possible for a VTK array to have type `vtkDataArray`, which is normally an abstract class, and yet not actually be abstract. This can happen when an array is dynamically allocated using `vtkImplicitArray`. However, the current copy mechanism assumes that the array type can be instantiated directly, which is not the case for `vtkDataArray` since it is abstract. This PR adds a dict lookup instead to init a new array based on the `GetArrayType` property instead of using the class type itself.

### DIFF
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -106,10 +106,25 @@ from vtkmodules.vtkCommonComputationalGeometry import (
 )
 from vtkmodules.vtkCommonComputationalGeometry import vtkParametricTorus as vtkParametricTorus
 from vtkmodules.vtkCommonCore import VTK_ARIAL as VTK_ARIAL
+from vtkmodules.vtkCommonCore import VTK_BIT as VTK_BIT
+from vtkmodules.vtkCommonCore import VTK_CHAR as VTK_CHAR
 from vtkmodules.vtkCommonCore import VTK_COURIER as VTK_COURIER
+from vtkmodules.vtkCommonCore import VTK_DOUBLE as VTK_DOUBLE
+from vtkmodules.vtkCommonCore import VTK_FLOAT as VTK_FLOAT
 from vtkmodules.vtkCommonCore import VTK_FONT_FILE as VTK_FONT_FILE
+from vtkmodules.vtkCommonCore import VTK_ID_TYPE as VTK_ID_TYPE
+from vtkmodules.vtkCommonCore import VTK_INT as VTK_INT
+from vtkmodules.vtkCommonCore import VTK_LONG as VTK_LONG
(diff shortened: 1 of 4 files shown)


banesullivan (approved):
Thanks for jumping on this!

PR #7662, opened 2025-07-07
Use `vtkOrientPolyData` filter with contour labels

user27182:
### Overview

Use `vtkOrientPolyData` filter internally for orientating the faces instead of `vtkPolyDataNormals`. This is more efficient because the actual normals aren't needed, just the re-orientation.


tkoyama010 (approved):
LGTM

PR #7663, opened 2025-07-07
Bump NumPy pin to <2.4.0

user27182:
### Overview

Not sure if dependabot is working properly since https://github.com/pyvista/pyvista/pull/7473... but NumPy can be bumped.


tkoyama010 (approved):
LGTM

PR #7668, opened 2025-07-07
Add automatic Jupyter display feature for Plotter objects

tkoyama010:
## Summary
This PR adds automatic display support for PyVista Plotter objects in Jupyter notebooks without requiring the `show()` method call. The feature is implemented using the `_repr_html_` method that Jupyter automatically calls for rich display of objects.

## Changes
- Added `_repr_html_` method to the `Plotter` class
- Automatically detects Jupyter environment and uses configured backend
- Supports all existing backends (static, trame, html, etc.)
- Gracefully handles exceptions and falls back to None
- Respects `jupyter_backend='none'` setting to disable auto-display
- Maintains full backward compatibility with existing `show()` method

## Usage
```python
# Before (still works):
plotter = pv.Plotter()
plotter.add_mesh(pv.Sphere())
plotter.show()

# After (new automatic display):
plotter = pv.Plotter()
plotter.add_mesh(pv.Sphere())
plotter  # Automatically displays in Jupyter\!
```

## Testing
- Added comprehensive test suite in `tests/plotting/jupyter/test_auto_display.py`
- Created Jupyter notebook for manual testing
- Tests cover various scenarios including backend switching and error handling
- All tests pass successfully

## Benefits
- More intuitive workflow for Jupyter users
- Consistent with other plotting libraries like matplotlib
- No breaking changes to existing code
- Optional - can be disabled with `jupyter_backend='none'`

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -7492,6 +7492,65 @@ class Plotter(BasePlotter):
             if hasattr(actor, 'mapper') and hasattr(actor.mapper, 'dataset')
         ]
 
+    def _repr_html_(self) -> str | None:
+        """Return HTML representation for Jupyter notebooks.
+
+        This method is called automatically by Jupyter when a Plotter
+        instance is the last expression in a cell, enabling automatic
+        display without calling show().
+
+        Returns
+        -------
+        str | None
+            HTML representation of the plotter, or None if not in Jupyter.
+
+        """
+        # Check if we're in a Jupyter environment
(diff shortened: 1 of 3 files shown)


banesullivan:
Brilliant! I've wanted this for a long while too. However this mechanism is going to be a little more complicated than this as we have the jupyter widget backend producing the widget and we need to persist that on the plotter for reuse. I'm going to open a new issue for this and I'll put some time to think through a robust mechanism for this

PR #7670, opened 2025-07-08
Revert "Limit dependabot PRs to two and update weekly instead of monthly"

tkoyama010:
It seems this frequency is too low.
Reverts pyvista/pyvista#7473


user27182:
> maybe we just need to add a catch-all case for the non-grouped `pip` ecosystem, e.g.
> 
> ``` yaml
> 
>   other:
> 
>     patterns:
> 
>       - "*"
> 
> ```

This shouldn't be necessary though... From the [docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--)
"Any outdated dependencies that do not match a rule are updated in individual pull requests."

tkoyama010:
Thank you! I will think more carefully about the policy.

user27182:
> Thank you! I will think more carefully about the policy.

Thanks, I think it does need to be updated since it doesn't seem like single dependencies (like `numpy`) are being updated properly, I'm just not sure if we need to revert back to the old policy. We can also try increasing the limit from 2->5, but keep weekly updates, and see if that works?

user27182:
I think `rebase-strategy: disabled` is what we need to use in the config to stop dependabot from draining our CI resources:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#rebase-strategy--

PR #7686, opened 2025-07-08
Fix ImageData.cast_to_rectilinear_grid to support offset and direction

banesullivan:
cast_to_rectilinear_grid() for ImageData was not accounting for the offset of the ImageData. `pv.Wavelet()` is a great example of where `offset` is used.


pyvista/core/grid.py
```diff
@@ -918,7 +918,7 @@ def _generate_rectilinear_coords(
         # Use linspace to avoid rounding error accumulation
         dims = self.dimensions
         spacing = self.spacing
-        origin = self.origin
+        origin = np.array(self.origin) + np.array(self.offset)
         return [
             (np.linspace(0, (dims[i] - 1) * spacing[i], dims[i]) + origin[i]) for i in range(3)
         ]
```
user27182: Offset is in IJK coordinates, so to convert to XYZ coords you must include the spacing.
```suggestion
        origin = self.origin
        offset = self.offset
        return [
            (np.linspace(offset[i] * spacing[i], (offset[i] + dims[i] - 1) * spacing[i], dims[i]) + origin[i]) for i in range(3)
        ]
```

And if we're including offset, we should also include the direction matrix. I think we can break this up into several statements:
``` suggestion
        origin = self.origin
        offset = self.offset
        direction = self.direction_matrix

        # Grid IJK indices
        ijk = [np.linspace(offset[i], offset[i] + dims[i] - 1, dims[i]) for i in range(3)]

        # Apply the transformation: coord = origin + direction @ (index * spacing)
        return [sum(direction[axis, j] * ijk[j] for j in range(3)) + origin[axis] for axis in range(3)]
```

tests/core/test_grid.py
```diff
@@ -1041,7 +1041,7 @@ def test_cast_uniform_to_structured():
 
 
 def test_cast_uniform_to_rectilinear():
-    grid = examples.load_uniform()
+    grid = pv.Wavelet()  # Uses an offset
```
user27182: The tests should include these explicitly
```suggestion
    grid.offset = (1, 2, 3)
    grid.direction_matrix = pv.Transform().rotate_x(30).rotate_y(40).rotate_z(50).matrix[:3,:3]
    grid.spacing = (1.1, 2.2, 3.3)
```

user27182 (changes requested):

user27182 (approved):

pyvista/core/grid.py
```diff
@@ -915,13 +915,22 @@ def _generate_rectilinear_coords(
...
-        ]
+        offset = self.offset
+        direction = self.direction_matrix
+        if not np.allclose(np.abs(self.direction_matrix), np.eye(3)):
+            msg = (
+                'Direction matrix must be a diagonal matrix when casting to RectilinearGrid; all'
+                '\noff-diagonal values are ignored. Consider casting to StructuredGrid instead.'
+            )
+            warnings.warn(msg, RuntimeWarning)
+        direction = np.diagonal(direction)
```
user27182: This isn't quite right. If we have a proper rotation with off-diagonal values, then the diagonals themselves will be smaller than 1, which means that taking the diagonals only will inadvertently result in a scaling factor being applied. We should instead just take the _sign_ of the diagonals.

user27182: This is fixed by [c01365c](https://github.com/pyvista/pyvista/pull/7686/commits/c01365c9721e55f21ef7c205770d626532018a06). We don't use the diagonals at all, we now just ignore direction completely if it's not a diagonal matrix.

banesullivan:
@user27182, thanks for fixing the outstanding issues here!!

PR #7690, opened 2025-07-10
Release v0.46

tkoyama010:
## Summary
- Bump version from 0.46.dev0 to 0.47.dev0
- Prepare for v0.46 release

## Release Notes
Release notes will be automatically generated based on merged PRs since the last release.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #7702, opened 2025-07-12
Remove deprecated active_t_coords property

tkoyama010:
## Summary
- Remove deprecated `active_t_coords` property and setter from `DataSet` class
- Remove deprecated `active_t_coords` and `active_t_coords_name` properties from `DataSetAttributes` class  
- Remove associated deprecation tests and update documentation references

## Test plan
- [x] Verify all references to `active_t_coords` have been removed
- [x] Run pre-commit hooks and ensure all checks pass
- [x] Run test suite to ensure no regressions

### DIFF
--- a/tests/core/test_dataset.py
+++ b/tests/core/test_dataset.py
@@ -1539,19 +1539,6 @@ def mesh():
     return examples.load_globe()
 
 
-def test_active_t_coords_deprecated(mesh):
-    with pytest.warns(PyVistaDeprecationWarning, match='texture_coordinates'):
-        t_coords = mesh.active_t_coords
-    if pv._version.version_info[:2] > (0, 46):
-        msg = 'Remove this deprecated property'
-        raise RuntimeError(msg)
-    with pytest.warns(PyVistaDeprecationWarning, match='texture_coordinates'):
-        mesh.active_t_coords = t_coords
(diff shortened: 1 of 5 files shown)


tests/core/test_dataset.py
```diff
@@ -1539,19 +1539,6 @@ def mesh():
     return examples.load_globe()
 
 
-def test_active_t_coords_deprecated(mesh):
-    with pytest.warns(PyVistaDeprecationWarning, match='texture_coordinates'):
-        t_coords = mesh.active_t_coords
-    if pv._version.version_info[:2] > (0, 46):
```
user27182, line 1545: This line suggests this should be removed _after_ 0.46 is released. We should wait until #7690 is merged then?

tkoyama010: That's a good question, but no. We should merge this before version 0.46 is merged, because we need to merge version 0.47.dev first. This is necessary since 0.47.dev is greater than (0, 46), and that's why we added this if block to handle the case where we want to remove a feature in v0.46.

user27182 (approved):

tests/core/test_dataset.py
user27182: I see, deprecated 0.43, so there were three revision (0.43, 0.44, 0.45), and can be removed now.

PR #7705, opened 2025-07-13
Remove deprecated download_frog_tissue function

tkoyama010:
## Summary
- Remove `download_frog_tissue` function that was deprecated in v0.44.0 and scheduled for removal in v0.47.0
- Remove associated helper functions and dataset loaders (`_frog_tissue_files_func`, `_dataset_frog_tissue`)
- Remove test case `test_download_frog_tissue` from test suite

## Background
This function was deprecated because it does not load correctly on some systems. Users should use `load_frog_tissues` instead.

## Test plan
- [x] Verify no remaining references to `download_frog_tissue` in the codebase
- [x] Ensure all tests pass after removal
- [x] Confirm pre-commit hooks pass

### DIFF
--- a/pyvista/examples/downloads.py
+++ b/pyvista/examples/downloads.py
@@ -2261,62 +2261,6 @@ _dataset_frog = _MultiFileDownloadableDatasetLoader(_frog_files_func)
 
 
 @_deprecate_positional_args
-def download_frog_tissue(load=True):  # noqa: FBT002
-    """Download frog tissue dataset.
-
-    This dataset contains tissue segmentation labels for the frog dataset.
-
-    .. deprecated:: 0.44.0
-
-        This example does not load correctly on some systems and has been deprecated.
-        Use :func:`~pyvista.examples.examples.load_frog_tissues` instead.
-
-    Parameters
-    ----------
-    load : bool, default: True
-        Load the dataset after downloading it when ``True``.  Set this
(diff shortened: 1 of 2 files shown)


tkoyama010:
@pyvista-bot LGTM

user27182:
Probably need to also remove line 96 from doc/source/make_tables.py

pyvista/examples/downloads.py
```diff
@@ -2260,62 +2260,6 @@ def _frog_files_func():
...
-            Browse other medical datasets.
-
-    """
-    # Deprecated on v0.44.0, estimated removal on v0.47.0
-    warnings.warn(
-        'This example is deprecated and will be removed in v0.47.0. '
-        'Use `load_frog_tissues` instead.',
-        PyVistaDeprecationWarning,
-    )
-    if pyvista._version.version_info >= (0, 47):
```
user27182, line 2302: If deprecated in 0.44, and the normal deprecation period is three versions, then shouldn't this technically still be valid for 0.44, 0.45, and 0.46? And therefore only be removed after 0.46 is released? Might be an off-by-one error, and need to make the equality check strictly greater than 0.46


Then again, maybe it's just an estimated removal, and need not have exactly 3 versions as the deprecation period.

user27182 (approved):

tkoyama010:
Thanks!

PR #7706, opened 2025-07-13
Remove deprecated update_coordinates method

tkoyama010:
## Summary
- Removes the deprecated `update_coordinates` method from the Plotter class
- Removes associated test for the deprecated method
- Removes unused `PyVistaDeprecationWarning` import from test file

## Background
The `update_coordinates` method was deprecated in v0.43.0 with a note that it would be removed in a future version. The method was functionally equivalent to directly modifying the points of a mesh in-place:

```python
# Instead of:
plotter.update_coordinates(points)

# Users should:
mesh.points = points
plotter.render()  # if needed
```

## Test plan
- [x] Pre-commit hooks pass
- [x] CI tests should pass (no functionality is being changed, only deprecated code removed)

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -5017,58 +5017,6 @@ class BasePlotter(PickingHelper, WidgetHelper):
         if render:
             self.render()
 
-    @_deprecate_positional_args(allowed=['points'])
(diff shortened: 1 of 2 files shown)


tkoyama010:
@pyvista-bot LGTM

PR #7710, opened 2025-07-14
Bugfix: Error loading .obj without `uv`

imontesino:
### Overview

When loading an .obj without `uv` properties the loader unproperly checks this condition leading to the following error: 
```python-traceback
    mesh = pv.wrap(tm)
  File "/home/username/.local/lib/python3.10/site-packages/pyvista/core/utilities/helpers.py", line 244, in wrap
    polydata.active_texture_coordinates = np.asarray(dataset.visual.uv)
  File "/home/username/.local/lib/python3.10/site-packages/pyvista/core/dataset.py", line 2931, in active_texture_coordinates
    self.point_data.active_texture_coordinates = texture_coordinates  # type: ignore[assignment]
  File "/home/username/.local/lib/python3.10/site-packages/pyvista/core/datasetattributes.py", line 1521, in active_texture_coordinates
    raise ValueError(msg)
ValueError: Texture coordinates must be a 2-dimensional array
```  

### Details

- Check if `dataset.visual.uv is not None` after checking if the attribute exists
- Fixes loading as checked with this [.obj file](https://github.com/bulletphysics/bullet3/blob/master/data/kuka_iiwa/meshes/link_0.obj)

### DIFF
--- a/pyvista/core/utilities/helpers.py
+++ b/pyvista/core/utilities/helpers.py
@@ -243,7 +243,7 @@ def wrap(  # noqa: PLR0911
             faces=dataset.faces,
         )
         # If the Trimesh object has uv, pass them to the PolyData
-        if hasattr(dataset.visual, 'uv'):
+        if hasattr(dataset.visual, 'uv') and dataset.visual.uv is not None:
             polydata.active_texture_coordinates = np.asarray(dataset.visual.uv)
         return polydata
 


MatthewFlamm:
trimesh documentation states it can be `None` so this fix looks correct to me. https://trimesh.org/trimesh.visual.texture.html#trimesh.visual.texture.TextureVisuals.uv

 Thanks!

MatthewFlamm (approved):

user27182:
This fix is OK, but it seems more like an issue with the underlying `active_texture_coordinates` API. I think `wrap` here should support setting `active_texture_coordinates = None` directly, without first calling `np.array` on the input. Checking for None and casting to an array should be handled by the setter, not by `wrap` itself.

MatthewFlamm:
> This fix is OK, but it seems more like an issue with the underlying `active_texture_coordinates` API. I think `wrap` here should support setting `active_texture_coordinates = None` directly, without first calling `np.array` on the input. Checking for None and casting to an array should be handled by the setter, not by `wrap` itself.

I agree with this in general.  We also need to check that `active_texture_coordinates` API can accept `ArrayLike` because of this line `polydata.active_texture_coordinates = np.asarray(dataset.visual.uv)`.

user27182:
> > This fix is OK, but it seems more like an issue with the underlying `active_texture_coordinates` API. I think `wrap` here should support setting `active_texture_coordinates = None` directly, without first calling `np.array` on the input. Checking for None and casting to an array should be handled by the setter, not by `wrap` itself.
> 
> I agree with this in general. We also need to check that `active_texture_coordinates` API can accept `ArrayLike` because of this line `polydata.active_texture_coordinates = np.asarray(dataset.visual.uv)`.

Yes the issue is that the [`active_texture_coordinates`](https://dev.pyvista.org/api/core/_autosummary/pyvista.dataset.active_texture_coordinates#pyvista.DataSet.active_texture_coordinates) setter currently requires the input to be an instance of np.ndarray, and does not call `np.asarray` or allow setting `None`:

https://github.com/pyvista/pyvista/blob/f4ed65307d3444c041140a540fbbd9d9a2b1e323/pyvista/core/datasetattributes.py#L1459-L1463

But, the getter can return `None`. So, `None` is a valid value, but we're currently not allowed to set it as such. So IMO the proper fix would be to:

1. Not check for `None` in and not call `np.asarray` in `wrap`, i.e.:
``` diff
-        if hasattr(dataset.visual, 'uv') and dataset.visual.uv is not None:
-            polydata.active_texture_coordinates = np.asarray(dataset.visual.uv)
+        if hasattr(dataset.visual, 'uv'):
+            polydata.active_texture_coordinates = dataset.visual.uv
```
2. Fix up the input validation for `DataSetAttributes.active_texture_coordinates` to check for None and cast with `np.asarray` as needed

user27182:
>  2. Fix up the input validation for `DataSetAttributes.active_texture_coordinates` to check for None and cast with `np.asarray` as needed

I reviewed other similar setters:

1. Setting `active_scalars` or `active_vectors` directly isn't even allowed. `set_scalars` or `set_vectors` must be used instead, and those functions don't allow setting `None` as the array input.
2. Setting `active_normals` _is_ allowed, but this also does not support setting `None` as the input.

I guess the issue is that there is an array name tied to this array as well, so setting the _name_ to `None` seems to be how our API works for removing active arrays, and NOT by setting the array itself. E.g. do `active_texture_coordinates_name = None`, not `active_texture_coordinates = None`.

It seems somewhat awkward not to allow the array setter to accept None, but then allowing the getter to return None. But this _is_ consistent with the current API. So this PR can be merged as-is after all I think.

user27182:
Thanks @imontesino !

PR #7714, opened 2025-07-14
Use `image_threshold` internally by `select_values` to improve performance

user27182:
### Overview
`select_values` provides identical functionality to `image_threshold` in specific cases. And `image_threshold` can be much faster in those cases. This PR therefore uses `image_threshold` internally as part of `select_values` for these cases.

E.g.: 

``` python
import pyvista as pv
ct = pv.examples.download_whole_body_ct_male()['ct']
bone_range = [150, float('inf')]
fill_value = -1000
%timeit bone_image = ct.select_values(ranges=bone_range, fill_value=fill_value)
%timeit bone_image = ct.image_threshold(bone_range, in_value=fill_value)

```

This branch:

789 μs ± 234 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
538 μs ± 5.59 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

`main` branch:

20.2 ms ± 1.57 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
625 μs ± 88 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)


banesullivan:
This looks pretty clean to me, should we update the branch and try to land this?

user27182:
Yes I think I didn’t follow through because I was concerned about missing coverage, where some functionally of `select_values` would be accidentally broken by this change. But yea the change should otherwise be Ok. Worth reviving because the performance difference is substantial, and covers the more common use cases for this filter.

banesullivan (approved):

PR #7715, opened 2025-07-15
Do not fail unit testing if `vtk-data` cache restore fails

user27182:
### Overview

The `vtk-data` cache and restore from #7501 is nice but the action is a bit flaky, and sometimes fails to restore. This currently fails the job, which isn't ideal, but is made worse by the fact that simply retrying the job often doesn't fix the issue. This makes it so that the PR must be closed/reopened to retry the job, or a new commit must be pushed, and _all_ CI must be retried instead of just the one job.

I think it's fine to continue the job without the cache. If the restore fails, the `vtk-data` dir will simply not be created. The downloads module already checks that the USER_DATA_CACHE dir exists, and falls back to using the default dir in this case. So the job should not fail, and simply download directly from the repo as usual.

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -105,7 +105,6 @@ jobs:
           path: ${{ env.CACHE_FOLDER_NAME }}
           key: ${{ needs.cache-vtk-data.outputs.key }}
           enableCrossOsArchive: true
-          fail-on-cache-miss: true
 
       - name: Install PyVista test dependencies
         run: |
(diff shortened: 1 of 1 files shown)


banesullivan (approved):

user27182:
This didn't quite work. A warning is emitted because the cached data path is invalid, which is normally OK, but we have strict warnings enabled which causes a pytest failure:
https://github.com/pyvista/pyvista/actions/runs/16338794504/job/46304630882?pr=7037#step:7:31
``` py
E   UserWarning: Ignoring invalid {PYVISTA_USERDATA_PATH
```

user27182:
Even after #7739, we're _still_ having issues when there is a cache miss, in this case with the tinypages tests:

https://github.com/pyvista/pyvista/actions/runs/16522947724/job/46728920625#step:15:1

IIRC this test runs a subprocess, so maybe the pytest filter warnings don't get forwarded in this case? 🤔

Issue #3806, opened 2023-01-10
Deprecate `BasePlotter.mesh`, `BasePlotter.mapper`, and `BasePlotter.volume`


banesullivan:
We assign `BasePlotter.mesh` and `BasePlotter.mapper` during plotting due to some legacy design choices for the methods `update_scalars()`, `update_coordinates()`, and `add_scalar_bar()`. Removing these assignments will clean up some of the code in `add_mesh()` and `add_volume()` as well as improve GC on the `BasePlotter`. For instance we have to make sure these are cleaned up here:

https://github.com/pyvista/pyvista/blob/61d16855746ad5aa9d63fd2e531bc8b880558cb0/pyvista/plotting/plotting.py#L3870-L3871

and here:

https://github.com/pyvista/pyvista/blob/61d16855746ad5aa9d63fd2e531bc8b880558cb0/pyvista/plotting/plotting.py#L4134-L4135

and here:

https://github.com/pyvista/pyvista/blob/61d16855746ad5aa9d63fd2e531bc8b880558cb0/pyvista/plotting/plotting.py#L4175-L4177

Specifically, `update_scalars()` and `update_coordinates()` are for updating a mesh that was previously added to the scene. IMO, these methods should not be on the Plotter, and any change to the mesh object should be registered by the rendering pipeline (already possible but improved by #3318). These methods allow for their `mesh` argument to be `None`, defaulting to the last added mesh (`self.mesh`):

https://github.com/pyvista/pyvista/blob/61d16855746ad5aa9d63fd2e531bc8b880558cb0/pyvista/plotting/plotting.py#L4031=L4032

https://github.com/pyvista/pyvista/blob/61d16855746ad5aa9d63fd2e531bc8b880558cb0/pyvista/plotting/plotting.py#L4089-L4090

On the other hand, `self.mapper` is only used by `add_scalar_bar()` in a specific scenario:

https://github.com/pyvista/pyvista/blob/61d16855746ad5aa9d63fd2e531bc8b880558cb0/pyvista/plotting/plotting.py#L3989-L3994

which may be easier to handle/deprecate.

## Proposed Deprecation to clean up `self.mesh`

[...]

PR #7716, opened 2025-07-15
Do not allow setting new attributes on PyVista classes 

### LINKED ISSUE #5216: Use `@no_new_attr` decorator for data model classes

### LINKED ISSUE #7713: Texture.interpolate example is incorrect

This property is setting property that doesn't exist:

https://github.com/pyvista/pyvista/blob/916a4fb7c6410e357b461580ee840a6d61761e80/pyvista/plotting/texture.py#L164-L184

I think we need to add the `no_new_attr` decorator to this class which intends to prevent issues like this

The example is setting `texture.interpolation = False` but `interpolation` is not a valid property, it's supposed to be `interpolate`

```python
Insert code here
```

### System Information

```shell
main
```

### Screenshots

_No response_

user27182:
### Overview

- Replace `no_new_attr` decorator with `NoNewAttrMixin` class
- Make all concrete classes inherit from this class
- Move dynamic properties into `__init__` methods
- Update typing and attribute checks to work with these change
- Add `pv.set_new_attribute` function for setting new attributes in case users actually want to do this
- Many actor classes were given a new `name` attribute by `add_actor`; these classes now inherit from `NameMixin` to formally add this attribute
- Many abstract classes were not marked as such, which would cause test failure from the new tests that check for new attributes. These classes are now properly marked as abstract
- Fix #7713 
- Partial fix for #7723 
- Resolve #5216

Unlike using `__slots__` or the `no_new_attr` decorator, this implementation does not require maintaining a list of allowed attributes. Instead, a flag is set in `__init__` to control this. The flag is set automatically by using a metaclass.


pyvista/core/utilities/misc.py
```diff
@@ -280,28 +281,53 @@ def _check_range(value: float, rng: Sequence[float], parm_name: str) -> None:
...
+        if isinstance(obj, _NoNewAttrMixin):
+            obj._no_new_attributes(cls)
+        return obj
+
+
+class _NoNewAttrMixin:
+    """Mixin to prevent adding new attributes.
+
+    Prevents additional attributes via xxx.attribute = "something" after a
+    call to `self._freeze()`. Mainly used to prevent the user from using
```
adeak: Where is `_freeze()`?

user27182: The docstring still needs to be updated. 

This was the method name left over from the pandas implementation. It was renamed to `_no_new_attributes` for clarity. There are also two versions of this mixin class. The "Auto" version uses a metaclass to automatically call this.

adeak: Ah, OK, I didn't notice this was a draft, sorry.

user27182:
This is ready, MNE tests are currently failing but should be fixed by https://github.com/mne-tools/mne-python/pull/13330 downstream

pyvista/core/utilities/misc.py
```diff
@@ -280,28 +282,86 @@ def _check_range(value: float, rng: Sequence[float], parm_name: str) -> None:
         raise ValueError(msg)
 
 
-def no_new_attr(cls):  # noqa: ANN001, ANN201 # numpydoc ignore=RT01
```
user27182, line 283: Looks like at least one downstream package is using this decorator 

https://github.com/ansys/pymapdl/blob/f021ab206446e3fe34822d69ca5e820d3649c0b0/src/ansys/mapdl/core/plotting/plotting_defaults.py#L35

So maybe it shouldn't be removed? I thought this was undocumented though and for internal use only, so we should be free to remove it. 

This is a separate issue, but we should consider renaming private implementation modules, functions and classes to have a leading underscore to make it clear that these are not public.

doc/source/api/utilities/utilities.rst
```diff
@@ -135,6 +135,7 @@ Miscellaneous
 .. autosummary::
    :toctree: _autosummary
 
+   set_new_attribute
```
user27182, line 138: Instead of a new helper function, we _could_ add this directly to `_NoNewAttrMixin` as a method. Would make it slightly easier to add new attributes:

``` diff
- pv.set_new_attribute(mesh, 'foo', 42)
+ mesh.set_new_attribute('foo', 42)
```

`pv.set_new_attribute` matches the signature for `setattr` though, and having it as a separate function prevents it from being suggested as a method by IDEs, both of which are good things IMO.

user27182:
CI is green, thank you @larsoner 🙏

pyvista/core/utilities/misc.py
banesullivan: I think here we should engage with the pymapdl developers and let them know ahead of time that we are changing this. cc @germa89 and @RobPasMue

I am also guilty of using this decorator in some of my own code and will have to migrate... which makes me think @akaszynski probably has too... which makes me think this sort of thing really ought to just exist on its own.

Moving forward: yeah, we should be more explicit with private functions/classes to avoid these scenarios.

doc/source/api/utilities/utilities.rst
banesullivan: Ooo, I like the idea of having this as a part of the `_NoNewAttrMixin` API however:

> having it as a separate function prevents it from being suggested as a method by IDEs

This is a big pro of it being a dedicated method... I'm imaging AI coding tools seeing this method and suggesting all kinds of silly ways to use it

banesullivan: Is `pv.set_new_attribute` well documented/used? if not, i'd be +1 for having it as a private method: `_NoNewAttrMixin._set_new_attribute`

pyvista/core/pointset.py
```diff
@@ -1686,12 +1684,10 @@ def obbTree(self):  # noqa: N802  # numpydoc ignore=RT01
...
-        if not hasattr(self, '_obbTree'):
-            self._obbTree = _vtk.vtkOBBTree()
-            self._obbTree.SetDataSet(self)
-            self._obbTree.BuildLocator()
-
-        return self._obbTree
+        obb_tree = _vtk.vtkOBBTree()
+        obb_tree.SetDataSet(self)
+        obb_tree.BuildLocator()
+        return obb_tree
```
banesullivan, line 1690: This may have performance costs... Would `cached_property` be appropriate here? Overall I'm okay with this change but I want to be mindful of downstream usage of this property now taking a performance hit on repeated access since it has to build the locator each time

doc/source/api/utilities/utilities.rst
user27182: It is documented, here's the current docs preview:
https://687d16c24eebc96a06ca3c04--pyvista-dev.netlify.app/api/utilities/_autosummary/pyvista.set_new_attribute.html#pyvista.set_new_attribute

In terms of current usage where setting new attributes is expected:

- there is a docstring for `save_pickle` that uses this (a reason to save pickled meshes is precisely because any new attributes added will also be saved)
- there is a test in `test_scaper.py` where setting `plotter.app` is a part of the test (and `app` is a new attribute)

So, there is some existing use of this in the code. And if downstream users develop apps where they expect to be able to do this (it _is_ a Python feature and part of the language after all), then at least they have an easy way to do this. Currently, if trying to set a new attribute the "normal" way a suggestion to use `pv.set_new_attribute` explicitly is included in the error message.

banesullivan (approved):
I really love what you've done here! This is a much needed improvement and goes a long way to clean up what feels like cruft to me (e.g. `_new_attr_exceptions` and lots of internal assignments).

Nicely done, just a few comments but nothing to block going ahead and merging

---

I'm hopeful that one day we will handle the `mesh`, `mapper`, etc attributes of the `BasePlotter` class per https://github.com/pyvista/pyvista/issues/3806. Adding `_NoNewAttrMixin` to `Plotter` helps us towards this! 🎉

doc/source/api/utilities/utilities.rst
banesullivan: >Currently, if trying to set a new attribute the "normal" way a suggestion to use pv.set_new_attribute explicitly is included in the error message.

Let's keep it this way with `pv.set_new_attribute` and a clear error message -- I think its the best overall and I really want to avoid rogue use of `mesh.set_new_attribute()`

pyvista/core/pointset.py
user27182: Indeed this appears to be intended to be a caching mechanism, and in our code at least is primarily used by the `ray_trace` filter. IMO `PolyData` is not a static object, and its geometry _can_ change, so I am not sure if caching this globally like this is a good idea. If I do `mesh.scale(2, inplace=True)` then `self._obbTree` is now stale and should be invalidated. Instead, for things like `ray_trace`, this should be refactored to keep a local cache, and create an internal `_ray_trace` helper that requires passing the pre-built OBB as a parameter.

EDIT: Or maybe I'm wrong here, and `BuildLocator()` does _not_ need to be called any time the geometry changes?

banesullivan: I believe that the locator does need to be built if the geometry changes 

I think having a stance that the object is not static and thus the obbtree is rebuilt each time its accessed is appropriate and a "good for most people". If someone is using this, they should access that obbtree and assign it to a var in order to not recompute in for-loops for example

tests/test_attributes.py
```diff
@@ -181,3 +241,61 @@ def test_vtk_snake_case_api_is_disabled(vtk_subclass):
...
+        assert match in repr(e)  # noqa: PT017
+    except AttributeError as e:
+        if 'dict' in repr(e):
+            pytest.skip('Skip dict classes')
+    else:
+        if not issubclass(pyvista_class, _NoNewAttrMixin):
+            msg = (
+                f'The class {pyvista_class.__name__!r} in {pyvista_class.__module__!r}'
+                f'must inherit from \n{_NoNewAttrMixin.__name__!r} or '
+                f'{_NoNewAttrMixin.__name__!r} in {_NoNewAttrMixin.__module__!r}'
```
user27182: This is left over from an earlier implementation where there were multiple versions of the Mixin class
```suggestion
                f'must inherit from \n{_NoNewAttrMixin.__name__!r} '
                f'in {_NoNewAttrMixin.__module__!r}'
```

user27182:
> I'm hopeful that one day we will handle the `mesh`, `mapper`, etc attributes of the `BasePlotter` class per #3806. Adding `_NoNewAttrMixin` to `Plotter` helps us towards this! 🎉

For full transparency, I'll note that this PR specifically avoids having `BasePlotter` inherit from the new `_NoNewAttrMixin` class.  The integration tests fail if you do that because PyVistaQt's plotter inherits from a class that _also_ defines its own metaclass, and hence creates an error due to a metaclass conflict.

The `_NoNewAttrMixin` defines its own metaclass to make the whole freezing attributes process seamless: just inherit and that's it, works like magic. If you do want no new attributes for the `BasePlotter` in the future, then you can create two `NoNewAttrMixin` classes: one which does not have a metaclass (and hence avoids any metaclass conflicts), and one which does.

If you don't use the metaclass, then you need to explicitly call `obj._no_new_attributes()` in init to "freeze" all current attributes. But otherwise these classes would be identical.

pyvista/core/utilities/misc.py
germa89: I am not 100% sure, but I believe PyMAPDL will require minimal changes to adapt to this PR.

Pinging @AlejandroFernandezLuces for confirmation since he did implemented the visualizer in PyMAPDL.

AlejandroFernandezLuces: That part of the code comes from the legacy plotter, but I can take care of the migration regardless. It doesn't seem much work indeed.

user27182:
> The `_NoNewAttrMixin` defines its own metaclass to make the whole freezing attributes process seamless: just inherit and that's it, works like magic. If you do want no new attributes for the `BasePlotter` in the future, then you can create two `NoNewAttrMixin` classes: one which does not have a metaclass (and hence avoids any metaclass conflicts), and one which does.
> 
> If you don't use the metaclass, then you need to explicitly call `obj._no_new_attributes()` in init to "freeze" all current attributes. But otherwise these classes would be identical.

@banesullivan let me know if you'd prefer that I spilt this as part of this PR, e.g. add a `_NoNewAttrMixinNoMeta` class, and make `BasePlotter` inherit from this class instead of having `Plotter` inherit from `_NoNewAttrMixin`. Might be easier than refactoring later.

banesullivan:
Let's refactor later when it makes sense. I'm eager to see the next release with VTK 9.5 support

banesullivan (approved):

kalingibbons:
This really should be listed as a breaking change...

kalingibbons:
While I'm not yet able to articulate a succinct issue title, I'm trying to figure out how this class:
image

Tripped the `PyVistaAttributeError`

image

while

image

Is there any additonal documentation for the internal workings of `pyvista._wrappers` besides this example?

[https://docs.pyvista.org/examples/99-advanced/extending_pyvista.html](https://docs.pyvista.org/examples/99-advanced/extending_pyvista.html)

user27182:
A somewhat ugly but backwards-compatible way to set a new attribute is to use `object.__setattr__(obj, name, value)`, e.g. `object.__setattr__(self, '_transform_operation_counter', 0)` in this case. This should bypass the pyvista checks that would otherwise raise an error.

We could also relax the requirement to use `pv.set_new_attribute`, and only restrict setting new _public_ attributes. But if it's private (i.e. prefixed with an underscore), then don't raise an error... maybe issue a warning instead, or maybe do nothing at all?

user27182:
> While I'm not yet able to articulate a succinct issue title, I'm trying to figure out how this class: image
> 
> Tripped the `PyVistaAttributeError`
> 
> image
> while
> 
> image
> Is there any additonal documentation for the internal workings of `pyvista._wrappers` besides this example?
> 
> https://docs.pyvista.org/examples/99-advanced/extending_pyvista.html

@kalingibbons Can you provide a minimal example for this error?

Here are a few currently valid ways to set new attributes on subclasses:

``` py
import pyvista as pv

def add_foo(cls):
    cls.foo = 42
    return cls

def add_bar(cls):
    original_init = cls.__init__

    def new_init(self, *args, **kwargs):
        original_init(self, *args, **kwargs)
        self.bar = 43

    cls.__init__ = new_init
    return cls

@add_foo
@add_bar
class Poly(pv.PolyData):
    def __init__(self):
        super().__init__()
        self.baz = 44
        
    @property
    def qux(self):
        return 45

Poly().foo  # 42
Poly().bar  # 43
Poly().baz  # 44
Poly().qux  # 45
```

Would something like this be helpful to add to the "Extending PyVista" docs?

user27182:
@kalingibbons this has been patched in `0.46.3`. Your decorator should now continue to work without any breakage.
https://github.com/pyvista/pyvista/releases/tag/v0.46.3

PR #7717, opened 2025-07-15
Add PEP 723 inline script metadata to example files

tkoyama010:
## Summary
- Added PEP 723 inline script metadata to all 181 Python files in `examples/` and `examples_trame/` directories
- Each file now includes proper dependency declarations and Python version requirements
- Enables running examples as standalone scripts with PEP 723-compatible tools like `uv`

## Changes Made
- Added metadata blocks with `requires-python = ">=3.9"` matching project requirements  
- Automatically detected and listed third-party dependencies for each script (pyvista, numpy, matplotlib, scipy, trame)
- Placed metadata after docstrings but before imports as per PEP 723 specification
- Excluded standard library imports from dependency lists

## Test plan
- [ ] Verify examples can be run with `uv run <script.py>` 
- [ ] Check that existing functionality is unchanged
- [ ] Confirm metadata format follows PEP 723 specification

### DIFF
--- a/examples/00-load/create_draped_surface.py
+++ b/examples/00-load/create_draped_surface.py
@@ -33,6 +33,15 @@ profile in 2D with the coordinates associated to the top of each column in your
 2D array.
 """
 
+# /// script
+# requires-python = ">=3.9"
+# dependencies = [
+#   "matplotlib",
+#   "numpy",
+#   "pyvista",
+# ]
+# ///
+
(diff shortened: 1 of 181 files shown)


banesullivan:
Closing due to staleness. I'd like to have a broader discussion in an issue on if we actually want to adopt this or other standards from the UV ecosystem since we have mostly leaned into using UV

PR #7520, opened 2025-05-14
Disable vtk's snake case API for `pyvista_ndarray`

user27182:
### Overview

Fix #7519 

This PR refactors the `DisableVtkSnakeCase` class to support composition cases where vtk classes are not directly inherited.  This is done by separating the attribute check (as a standalone static method) from the __getattribute__ call, allowing `pyvista_ndarray` to disable the snake case API.

PR #7722, opened 2025-07-16
Fix snake case API for `DataSetAttributes`

user27182:
### Overview

It is currently possible to do:
``` py
import pyvista as pv
pv.Plane().point_data.object_name
```

But `object_name` is defined by VTK's snake case API, and not by PyVista, and should normally raise an AttributeError. This PR fixes this.

This fix is identical to fix to `pyvista_ndarray` from #7520 in that it's necessary to insert the check inside of `__getattr__`. To achieve this, the `VTKObjectWrapper` class is redefined in `_vtk_core`, and `pyvista_ndarray` is refactored to use the same method as the wrapper class.


tests/test_snake_case_api.py
```diff
@@ -57,7 +58,7 @@ def get_vtk_subclasses_through_inheritance() -> dict[str, type]:
 
             def inherits_from_vtk(klass):
                 bases = klass.__mro__[1:]
-                return any(base.__name__.startswith('vtk') for base in bases)
+                return any(base.__name__[:3].lower() == 'vtk' for base in bases)
```
tkoyama010: [ask] May I ask why this is necessary? I recall that 'vtk' is always written in lowercase.

user27182: Checking lower case vtk only is what caused the tests to miss collecting `DataSetAttributes`, which inherits from `VTKObjectWrapper`. So, checking upper case VTK allows us to auto-collect cases like this, otherwise they get missed by the tests and hence can create bugs like the one this PR resolves.

tkoyama010 (approved):
LGTM

PR #7725, opened 2025-07-18
Rename `test_snake_case_api` file to `test_attributes`

user27182:
### Overview

Most of the helpers/utility functions for generating the snake case API tests are expanded upon and reused as part of #7716. As such, we can generalize the test file name.


tkoyama010 (approved):
LGTM

PR #7718, opened 2025-07-15
Add `resize` filter for dataset coordinate rescaling

tkoyama010:
## Summary
- Adds a new `rescale` method to `DataSetFilters` for convenient coordinate rescaling
- Supports two modes: rescaling to fit within specified bounds or by scale factor(s)
- Provides comprehensive testing and documentation

## Key Features
- **Bounds rescaling**: Scale and translate dataset to fit exactly within target bounds
- **Scale factor rescaling**: Apply uniform or non-uniform scaling with optional custom center
- **In-place modification**: Optional in-place operation for memory efficiency
- **Data preservation**: All point and cell data is preserved during rescaling
- **Error handling**: Comprehensive validation and clear error messages

## Example Usage
```python
import pyvista as pv

# Rescale to unit bounds
mesh = pv.Sphere(radius=100)
unit_mesh = mesh.rescale(bounds=[-1, 1, -1, 1, -1, 1])

# Scale by factor
scaled = mesh.rescale(scale_factor=2.0)

# Non-uniform scaling
stretched = mesh.rescale(scale_factor=[2, 1, 0.5])
```

## Use Cases
This method addresses common needs for:
- Normalizing datasets to standard coordinate ranges
- Converting between different units
- Preparing data for visualization or analysis
- Fitting multiple datasets into comparable scales

## Test Coverage
Added comprehensive tests covering:
- Bounds-based rescaling
- Scale factor rescaling (uniform and non-uniform)
- Custom center scaling
- In-place modification
- Error handling
- Data preservation across different dataset types

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #7726, opened 2025-07-18
Rework `rescale` filter as `resize` filter

user27182:
### Overview

Update #7718 so that users can set the `size` of a mesh directly. (The size property is added in #7092)

Most of the AI slop and redundancy has been removed/reworked.

fix #7718


tkoyama010 (approved):
Thanks a lot!

Issue #7728, opened 2025-07-19
Inconsistent version of trimesh


riku-sakamoto:
Hello! First of all, thank you for creating and maintaining such a super useful tool.

## Overview

When I clone this repo and try to install dependencies by `uv`, I encounter the following error.
So far, when we follow to the instruction in [Contribution Docs](https://docs.pyvista.org/extras/developer_notes.html), we can ignore this inconsistency by using `pip install`. But, it may causes issues in the future.

```
  × No solution found when resolving dependencies:
  ╰─▶ Because pyvista:docs depends on trimesh==4.6.13 and pyvista:dev depends on trimesh<4.6.0, we can conclude that pyvista:dev and pyvista:docs are incompatible.
      And because your project requires pyvista:dev and pyvista:docs, we can conclude that your project's requirements are unsatisfiable.
```

## Rationale

According to `pyproject.toml`, the version number for trimesh is not compatible.

* dependency for typing : 
https://github.com/pyvista/pyvista/blob/c60756425893b39d61c8f1c1a73e50c76a251082/pyproject.toml#L68
 
* dependency for docs
https://github.com/pyvista/pyvista/blob/c60756425893b39d61c8f1c1a73e50c76a251082/pyproject.toml#L149

## Suggestion

As for dependency of typing, `trimesh<4.7.0` seems to be appropriate, which is the same as `test` dependency.
If it sounds reasonable, I will create PR.

## Pseudocode or Screenshots

```bash
uv venv
uv sync --group dev
```

user27182:
Thank you for reporting, yes this conflict between dev and docs is a known issue: see #6858

When we have the same dependency listed in multiple groups, they have a tendency to get out of sync. 
A PR to resolve this for Trimesh would be welcome, I'll happily review it. Though, unfortunately the dev and docs may become incompatible once again in the future. 

If you know how to resolve this core issue please let us know. One way to keep typing and tests in sync at least is to use subgroups, e.g. create a `trimesh` group, and include that group as part of `typing` and `test` groups. That way the requirement only gets listed once. But our `docs` group has stricter version requirements, so we can't re-use the group there unless we relax those to match the `test` and `typing` groups.

tkoyama010:
Thanks! Yes it is really annoying.

riku-sakamoto:
@user27182 
Thank you for your comment !

[...]

PR #7729, opened 2025-07-19
Fix incompatible version for trimesh

riku-sakamoto:
### Overview

* Update `trimesh` in `typing` dependency. (See #7728 )

### Details

- Set `trimesh<4.7.0` to ensure compatibility with the `docs` dependency.

### Pending Topics

* This is not a fundamental solution. It may be resolved to use `uv` as discussed in  #7728 and #7441
* The `docs` dependency includes `scipy==1.16.0`, which is not compatible with `requires-python = '>=3.9'`. This needs to be addressed to ensure consistent dependency resolution across all groups.


tkoyama010 (approved):
LGTM

PR #7732, opened 2025-07-19
Add comment for necessary noqa PTH109

tkoyama010:
## Summary
- ~Removed unnecessary `noqa PTH109` comments from `_deprecate_positional_args.py`~
- ~Fixed PTH109 violations by replacing `os.getcwd()` with `Path.cwd()`~

## Test plan
- [x] Ran ruff check to verify PTH109 violations are resolved
- [x] Confirmed pre-commit hooks pass

### DIFF
--- a/pyvista/_deprecate_positional_args.py
+++ b/pyvista/_deprecate_positional_args.py
@@ -174,7 +174,7 @@ def _deprecate_positional_args(
 
             # Get source file and line number
             file = Path(
-                os.path.relpath(inspect.getfile(f), start=os.getcwd()) # noqa: PTH109  # https://github.com/pyvista/pyvista/pull/7732
+                os.path.relpath(inspect.getfile(f), start=os.getcwd())  # noqa: PTH109  # https://github.com/pyvista/pyvista/pull/7732
             ).as_posix()
             lineno = inspect.getsourcelines(f)[1]
             location = f'{file}:{lineno}'
@@ -224,7 +224,7 @@ def _deprecate_positional_args(
                         # Get location where the function is called
                         frame = inspect.stack()[stack_level]
                         file = Path(
-                            os.path.relpath(frame.filename, start=os.getcwd()) # noqa: PTH109  # https://github.com/pyvista/pyvista/pull/7732
+                            os.path.relpath(frame.filename, start=os.getcwd())  # noqa: PTH109  # https://github.com/pyvista/pyvista/pull/7732
                         ).as_posix()
                         return f'{file}:{frame.lineno}'
 
--- a/pyvista/_deprecate_positional_args.py
+++ b/pyvista/_deprecate_positional_args.py
@@ -172,7 +172,7 @@ def _deprecate_positional_args(
                 signature_string = f'{signature_string[:-1]}, ...)'
 
             # Get source file and line number
-            file = Path(inspect.getfile(f)).relative_to(Path.cwd()).as_posix()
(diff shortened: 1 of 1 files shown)


pyvista/_deprecate_positional_args.py
```diff
@@ -173,7 +173,7 @@ def qualified_name() -> str:
                 signature_string = f'{signature_string[:-1]}, ...)'
 
             # Get source file and line number
-            file = Path(os.path.relpath(inspect.getfile(f), start=os.getcwd())).as_posix()  # noqa: PTH109
+            file = Path(os.path.relpath(inspect.getfile(f), start=Path.cwd())).as_posix()
```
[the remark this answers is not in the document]
user27182: We'll see with CI but I recall this causes errors with pathlib

user27182 (commented):
An attempt was made to use pathlib here originally but it caused issues. And the call site function was tested locally to ensure there is a proper clickable link generated to the call site. There isn't good coverage for this and using pathlib could break this. 

I am generally very wary of these linting changes to use pathlib instead of os since these often cause untested breakages, and think these need to be made very cautiously

pyvista/_deprecate_positional_args.py
```diff
@@ -173,7 +172,7 @@ def qualified_name() -> str:
                 signature_string = f'{signature_string[:-1]}, ...)'
 
             # Get source file and line number
-            file = Path(os.path.relpath(inspect.getfile(f), start=os.getcwd())).as_posix()  # noqa: PTH109
```
user27182, line 176: Can we revert this and add a comment that this is necessary to use os.path so that it doesn't get changed in the future

pyvista/_deprecate_positional_args.py
```diff
@@ -221,7 +220,7 @@ def inner_f(*args: P.args, **kwargs: P.kwargs) -> T:
                     def call_site() -> str:
                         # Get location where the function is called
                         frame = inspect.stack()[stack_level]
-                        file = Path(os.path.relpath(frame.filename, start=os.getcwd())).as_posix()  # noqa: PTH109
```
user27182, line 224: Same here, we should revert this change and add a comment

user27182 (changes requested):

pyvista/_deprecate_positional_args.py
```diff
@@ -173,7 +172,7 @@ def qualified_name() -> str:
                 signature_string = f'{signature_string[:-1]}, ...)'
 
             # Get source file and line number
-            file = Path(os.path.relpath(inspect.getfile(f), start=os.getcwd())).as_posix()  # noqa: PTH109
+            file = Path(inspect.getfile(f)).relative_to(Path.cwd()).as_posix()
```
tkoyama010:
```suggestion
            file = Path(os.path.relpath(inspect.getfile(f), start=os.getcwd())).as_posix()  # noqa: PTH109 # https://github.com/pyvista/pyvista/pull/7732
```

pyvista/_deprecate_positional_args.py
```diff
@@ -221,7 +220,7 @@ def inner_f(*args: P.args, **kwargs: P.kwargs) -> T:
                     def call_site() -> str:
                         # Get location where the function is called
                         frame = inspect.stack()[stack_level]
-                        file = Path(os.path.relpath(frame.filename, start=os.getcwd())).as_posix()  # noqa: PTH109
+                        file = Path(frame.filename).relative_to(Path.cwd()).as_posix()
```
tkoyama010:
```suggestion
                        file = Path(os.path.relpath(frame.filename, start=os.getcwd())).as_posix()  # noqa: PTH109 # https://github.com/pyvista/pyvista/pull/7732
```

user27182 (approved):

PR #7739, opened 2025-07-21
Fix and filter invalid `PYVISTA_USERDATA_PATH` warning

user27182:
### Overview

When an invalid `PYVISTA_USERDATA_PATH` variable is given, the warning message is not formatted correctly, e.g.:
https://github.com/pyvista/pyvista/actions/runs/16338794504/job/46304630882?pr=7037#step:7:30

``` py
UserWarning: Ignoring invalid {PYVISTA_USERDATA_PATH
```

And for our CI the warning is treated as an error, which we can ignore so that CI can proceed in case there is a cache miss. This will fix the issue from https://github.com/pyvista/pyvista/pull/7715#issuecomment-3091931844


tkoyama010 (approved):
Thanks!

Issue #5066, opened 2023-10-12
ImageData.points doesn't handle DirectionMatrix


alexiscltrn:
### Describe the bug, what's wrong, and what you expected.

It seems that the ImageData.points property is not handling the image DirectionMatrix correctly. In fact, it has been ignored.

### Steps to reproduce the bug.

```python

import pyvista as pv
import numpy as np

img = pv.ImageData(dimensions=(2,3,4), spacing=(0.5,1.0,1.5), origin=(0., 0.5, 1.0))
img.SetDirectionMatrix((-1, 0, 0, 0, 1, 0, 0, 0, 1))

img.points
#array([[0. , 0.5, 1. ],
#       [0.5, 0.5, 1. ],
#       [0. , 1.5, 1. ],
#       [0.5, 1.5, 1. ],
#       [0. , 2.5, 1. ],
#       [0.5, 2.5, 1. ],
#       [0. , 0.5, 2.5],
#       [0.5, 0.5, 2.5],
#       [0. , 1.5, 2.5],
#       [0.5, 1.5, 2.5],
#       [0. , 2.5, 2.5],
#       [0.5, 2.5, 2.5],
#       [0. , 0.5, 4. ],
#       [0.5, 0.5, 4. ],
#       [0. , 1.5, 4. ],
#       [0.5, 1.5, 4. ],
#       [0. , 2.5, 4. ],
#       [0.5, 2.5, 4. ],
#       [0. , 0.5, 5.5],
#       [0.5, 0.5, 5.5],
#       [0. , 1.5, 5.5],
#       [0.5, 1.5, 5.5],
#       [0. , 2.5, 5.5],
#       [0.5, 2.5, 5.5]])

np.array([img.GetPoint(i) for i in range(img.number_of_points)])
#array([[ 0. ,  0.5,  1. ],
#       [-0.5,  0.5,  1. ],
#       [ 0. ,  1.5,  1. ],
#       [-0.5,  1.5,  1. ],
#       [ 0. ,  2.5,  1. ],
#       [-0.5,  2.5,  1. ],
#       [ 0. ,  0.5,  2.5],
#       [-0.5,  0.5,  2.5],
#       [ 0. ,  1.5,  2.5],
#       [-0.5,  1.5,  2.5],
#       [ 0. ,  2.5,  2.5],
#       [-0.5,  2.5,  2.5],
#       [ 0. ,  0.5,  4. ],
#       [-0.5,  0.5,  4. ],
[...]

PR #7741, opened 2025-07-21
Get `points` directly from `vtkImageData` and `vtkRectilinearGrid`

user27182:
### Overview

`GetPoints()` was added to `vtkImageData` in vtk 9.4, let's use it!
See https://github.com/pyvista/pyvista/issues/5066#issuecomment-1784039050 for some context


user27182:
There are already tests for this:
https://github.com/pyvista/pyvista/blob/0aeea96cb811bbe7e1237d6b78724ca50a20c417/tests/core/test_grid.py#L1220-L1221
Including for empty ImageData init:
https://github.com/pyvista/pyvista/blob/0aeea96cb811bbe7e1237d6b78724ca50a20c417/tests/core/test_grid.py#L996

Plus, use of `points` is so ubiquitous in the test code, that there is plenty of coverage elsewhere, e.g. any/all tests that use `__eq__` is implicitly also checking that the `points` are correct, and there are other checks for correct points such as with `DataObjectFilters.transform` tests:
https://github.com/pyvista/pyvista/blob/0aeea96cb811bbe7e1237d6b78724ca50a20c417/tests/core/test_dataobject_filters.py#L660-L662

Is that sufficient? Or are there any other specific tests you have in mind that we should add?

PR #7744, opened 2025-07-23
Add module name prefixes to `image_cache` images

user27182:
### Overview

Use ``--image_prefix_name module`` feature from https://github.com/pyvista/pytest-pyvista/pull/187


pyproject.toml
```diff
@@ -124,9 +124,9 @@ docs = [
   'mypy==1.15.0',
   'numpydoc==1.8.0',
   'osmnx==2.0.2',
-  'pydata-sphinx-theme==0.15.4',     # Do not upgrade, see https://github.com/pyvista/pyvista/pull/7095
+  'pydata-sphinx-theme==0.15.4',                                                               # Do not upgrade, see https://github.com/pyvista/pyvista/pull/7095
```
tkoyama010:
```suggestion
  'pydata-sphinx-theme==0.15.4',     # Do not upgrade, see https://github.com/pyvista/pyvista/pull/7095
```

pyproject.toml
```diff
@@ -73,7 +73,7 @@ test = [
   'cmcrameri<1.10.0',
   'cmocean<4.0.4',
   'colorcet<3.2.0',
-  'embreex<2.17.8; sys_platform != "darwin" or platform_machine != "arm64"', # Does not work with arm-based macs
+  'embreex<2.17.8; sys_platform != "darwin" or platform_machine != "arm64"',                   # Does not work with arm-based macs
```
tkoyama010:
```suggestion
  'embreex<2.17.8; sys_platform != "darwin" or platform_machine != "arm64"', # Does not work with arm-based macs
```

tkoyama010 (changes requested):
@user27182 Should we wait until https://github.com/pyvista/pytest-pyvista/pull/187 is merged?

pyproject.toml
user27182: This formatting is enforced by `taplo`. I would prefer it didn't align comments like this but I don't think there's an option to change this.

user27182:
> @user27182 Should we wait until https://github.com/pyvista/pytest-pyvista/pull/187 is merged?

That's probably best. Though it will need to be merged _and_ a new version released if we want to have stable version here. Otherwise we'll still need to install from the `main` branch.

pyproject.toml
user27182: Turns out there is an option: `align_comments`

https://taplo.tamasfe.dev/configuration/formatter-options.html#formatter-options

PR #7745, opened 2025-07-24
Add `remove_unused_points` filters

user27182:
### Overview

Use `extract_cells` to remove unused points _without_ point merging.


tkoyama010 (dismissed):
LGTM

pyvista/core/filters/unstructured_grid.py
```diff
@@ -157,3 +170,115 @@ def clean(  # noqa: PLR0917
...
+                [
+                    mesh.celltypes,
+                    np.full(n_unused, CellType.VERTEX, dtype=np.uint8),
+                ]
+            )
+
+            # Convert to vtk arrays and set the new cells
+            vtk_cells = CellArray(all_cells)
+            vtk_celltypes = _vtk.numpy_to_vtk(all_celltypes)
+            mesh.SetCells(vtk_celltypes, vtk_cells)
```
user27182: All this nonsense was needed as a workaround for #7750.

BUT, I have found that setting `include_cells=False` is a much better workaround, as this DOES extract only the requested IDs, so we can remove this. Will push an update
```suggestion
```

user27182: This is all removed in https://github.com/pyvista/pyvista/pull/7745/commits/bc702eaf83878bf408c90b9363c00802455d87a6

tkoyama010 (approved):

pyvista/core/filters/unstructured_grid.py
```diff
@@ -157,3 +166,72 @@ def clean(  # noqa: PLR0917
...
+        """
+        # Must not include cells here else unused points will also be extracted
+        new_grid = self.extract_points(self.cell_connectivity, include_cells=False)
+        if (name := 'vtkOriginalPointIds') in (data := new_grid.point_data):
+            del data[name]
+
+        out = self if inplace else self.copy()
+        out.points = new_grid.points
+        out.point_data.update(new_grid.point_data, copy=not inplace)
+        return out
```
user27182: I think we need to use the original cell ids here and update the cells and cell data accordingly. The tests are probably only passing because it `appends` an unused point. But if the unused point was in the middle of the array instead of at the end, I think this method would fail as-is

user27182: This is fixed now.

PR #7746, opened 2025-07-25
BUG: Fix reference cycle regression

larsoner:
@user27182 this PR fixes a regression from https://github.com/pyvista/pyvista/pull/7716/files#diff-74eba4a52c2594069f09d6ca675e50579aa1cdf7bb4d2955f44a46482ebb395aL1748 . In MNE-Python we run a ton of examples building our docs, garbage collect after each one, and [ensure no `vtk` objects linger](https://app.circleci.com/pipelines/github/mne-tools/mne-python/28164/workflows/d33ebd65-7d1d-4d7b-963a-144c2be17bec/jobs/74694). After #7716 some `vtkPolyData` stuck around:
```
      File "/home/larsoner/python/mne-python/doc/sphinxext/mne_doc_utils.py", line 115, in _assert_no_instances
        raise ExtensionError(str(exc)) from None
    sphinx.errors.ExtensionError: 
    3 vtkPolyData @ mne/conf.py:Resetter.__call__:after:70_fnirs_processing.py:
    list[0]: len=1, 4 referrers: cell / dict['_glyph_geom'] / list[1] / list[743402]
    list[0]: len=1, 4 referrers: cell / dict['_glyph_geom'] / list[1] / list[743406]
    list[0]: len=1, 4 referrers: cell / dict['_glyph_geom'] / list[1] / list[743409]
```
It's fixed by restoring the setting-to-None in `__del__` that was (accidentally perhaps?) removed.

### DIFF
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -1737,6 +1737,11 @@ class PolyData(_PointSet, PolyDataFilters, _vtk.vtkPolyData):
         """
         return self.n_open_edges == 0
 
+    def __del__(self) -> None:
+        """Delete the object."""
+        # avoid a reference cycle that can't be resolved with vtkPolyData
+        self._glyph_geom = None
+
 
 @abstract_class
 class PointGrid(_PointSet):


larsoner:
FYI in case you want to try adding a test at some point, this *might* be the call from MNE that eventually leads to the failure / ref cycle:
```
        sphere = vtkSphereSource()
        sphere.SetThetaResolution(resolution)
        sphere.SetPhiResolution(resolution)
        sphere.SetRadius(radius)
        sphere.Update()
        geom = sphere.GetOutput()
        mesh = PolyData(center)
        glyph = mesh.glyph(orient=False, scale=False, factor=factor, geom=geom)
```
maybe you know right where to put it and want to try :smile:

user27182:
Thanks, yes calling `glyph` sets the `_glyph_geom` attr internally... this is presumably done as a cache mechanism for later, but is only useful if adding a legend at some point to show the glyph as a symbol.

But IMO this seems like premature optimization for a somewhat niche use case and a bad design pattern instead instead of doing a proper refactoring or just recomputing the glyph, similar to my comment for removing `_obbTree`
https://github.com/pyvista/pyvista/pull/7716#discussion_r2218095056

We may want to consider removing this attribute altogether... but that's for another PR

PR #7752, opened 2025-07-26
Do not align comments in yaml files

user27182:
### Overview

Set `align_comments` to false. See https://taplo.tamasfe.dev/configuration/formatter-options.html#formatter-options


tkoyama010 (approved):
LGTM

PR #7753, opened 2025-07-26
Add concrete `is_empty` and `get_data_range` methods for `PartitionedDataSet`

user27182:
### Overview

When type-checking with `PartitionedDataSet`, `mypy` complains about abstract methods `is_empty` and `get_data_range`.

This PR only adds proper implementation for `is_empty`, `get_data_range` just calls super.

### DIFF
--- a/pyvista/core/partitioned.py
+++ b/pyvista/core/partitioned.py
@@ -83,7 +83,7 @@ class PartitionedDataSet(DataObject, MutableSequence, _vtk.vtkPartitionedDataSet
     def __getitem__(self, index):
         """Get a partition by its index."""
         if isinstance(index, slice):
-            return PartitionedDataSet([self[i] for i in range(self.n_partitions)[index]])  # type: ignore[abstract]
+            return PartitionedDataSet([self[i] for i in range(self.n_partitions)[index]])
         else:
             if index < -self.n_partitions or index >= self.n_partitions:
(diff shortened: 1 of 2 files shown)


pyvista/core/partitioned.py
```diff
@@ -244,6 +247,26 @@ def n_partitions(self, n) -> None:
...
+    @property
+    def is_empty(self) -> bool:  # numpydoc ignore=RT01
+        """Return ``True`` if there are no blocks.
+
+        .. versionadded:: 0.46
+
+        Examples
+        --------
+        >>> import pyvista as pv
+        >>> mesh = pv.MultiBlock()
```
tkoyama010:
```suggestion
        >>> mesh = pv.PartitionedDataSet()
```

tkoyama010 (changes requested):

pyvista/core/partitioned.py
```diff
@@ -244,6 +247,26 @@ def n_partitions(self, n) -> None:
         self.SetNumberOfPartitions(n)
         self.Modified()
 
+    @property
+    def is_empty(self) -> bool:  # numpydoc ignore=RT01
+        """Return ``True`` if there are no blocks.
```
user27182:
```suggestion
        """Return ``True`` if there are no partitions.
```

tkoyama010 (approved):
LGTM

PR #7760, opened 2025-07-30
Use linear interpolation with anti-aliasing when resampling environment texture

user27182:
### Overview

`resample` uses nearest interpolation by default. But for textures with continuous data it makes sense to use linear interpolation instead.


banesullivan:
This should absolutely be fixed. Marking the PR ready for review so I can come back to it and try to revive this

user27182:
This one should be relatively easy to land.

banesullivan (dismissed):

banesullivan (approved):

PR #7762, opened 2025-07-30
Consolidate `cell_type_helper.py` into `celltype.py`

user27182:
### Overview

`cell_type_helper.py` creates a mapping from `pv.CellType` values to the number of points in each corresponding cell. But, this type of query is already made in `celltype.py`, and is therefore largely redundant.

A few mappings for Langrange and Bezier cells from `cell_type_helper.py` were missing from `celltype.py`, so these have been moved over, but otherwise `cell_type_helper.py` is no longer needed.


user27182:
Docs have been updated with new vtk links:

docs

PR #7763, opened 2025-07-30
Enable linting for bare except (E722 and BLE001)

user27182:
### Overview

Using bare except is bad because it also catches keyboard interrupts and system exits. Using `except Exception` is better than a bare except, though it's best not to use this either and to only include minimal exceptions for the expected errors.

### DIFF
--- a/pyvista/core/filters/poly_data.py
+++ b/pyvista/core/filters/poly_data.py
@@ -2926,7 +2926,7 @@ class PolyDataFilters(DataSetFilters):
             for key in self.cell_data:  # type: ignore[attr-defined]
                 try:
                     newmesh.cell_data[key] = self.cell_data[key][fmask]  # type: ignore[attr-defined]
-                except (ValueError, TypeError, KeyError):
+                except (ValueError, TypeError, KeyError):  # pragma: no cover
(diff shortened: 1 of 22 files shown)


tkoyama010 (dismissed):
LGTM

pyvista/core/filters/poly_data.py
```diff
@@ -2926,7 +2926,7 @@ def remove_points(  # noqa: PLR0917
             for key in self.cell_data:  # type: ignore[attr-defined]
                 try:
                     newmesh.cell_data[key] = self.cell_data[key][fmask]  # type: ignore[attr-defined]
-                except:
+                except (ValueError, TypeError, KeyError):
```
tkoyama010:
```suggestion
                except (ValueError, TypeError, KeyError):  # pragma: no cover
```

pyvista/core/utilities/misc.py
```diff
@@ -214,7 +214,7 @@ def try_callback(func, *args) -> None:  # noqa: ANN001
     """
     try:
         func(*args)
-    except Exception:
+    except Exception:  # noqa: BLE001
```
tkoyama010:
```suggestion
    except Exception:  # noqa: BLE001  # pragma: no cover
```

PR #7693, opened 2025-07-10
Drop support for Python 3.9 and VTK 9.0.3

tkoyama010:
## Summary
- Remove Python 3.9 from project classifiers and CI test matrix
- Update minimum Python version requirement to 3.10
- Drop VTK version 9.0.3 support and set minimum to 9.1.0
- Update documentation to reflect new minimum requirements
- Clean up compatibility code for dropped versions

### Python 3.9 Removal
- Updated `pyproject.toml` to remove Python 3.9 classifier and set minimum to 3.10
- Removed Python 3.9 from GitHub Actions CI test matrix
- Fixed ruff typing violations (Union to | syntax, removed unused imports)

### VTK 9.0.3 Removal  
- Updated minimum VTK version check from 9.0.0 to 9.1.0 in `__init__.py`
- Added `vtk>=9.1.0` requirement to `pyproject.toml` dependencies
- Removed VTK 9.0.3 compatibility code from composite mapper
- Cleaned up VTK 9.0.3 references and test skips
- Updated CI configuration to remove 9.0.3 from version matrix

## Test plan
- [x] Verify CI passes with updated matrix
- [x] Confirm PyPI deployment works with Python 3.10+
- [x] Check that documentation builds correctly
- [x] Verify all pre-commit hooks pass
- [x] Ensure VTK version requirements are properly enforced

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #7785, opened 2025-08-06
Revert "Drop support for Python 3.9 and VTK 9.0.3"

banesullivan:
Reverts pyvista/pyvista#7693

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -79,6 +79,7 @@ jobs:
       fail-fast: false
       matrix:
         config:
+          - { os: macos-13, python-version: "3.9" }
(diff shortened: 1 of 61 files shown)


MatthewFlamm (approved):

user27182 (approved):

tkoyama010 (approved):
Sorry about this 🙇

user27182:
Commenting here so I don't forget, but I realized that with vtk 9.0.3 dropped, we can drop conditional tests for PointSet which requires vtk>=9.1. We can also now include a `PointSet` instance as part of the `datasets` fixture and get much better coverage for that class... I suspect a few tests will break when we do this and reveal some bugs for some filters with PointSet inputs.

PR #7787, opened 2025-08-06
Allow using matplotlib cmap objects with `get_cmap_safe` and global theme

### LINKED ISSUE #7786: PyVista 0.46 regression: `global_theme.cmap` does not allow `matplotlib.colors.Colormap` instances

The global theme should allow setting matplotlib cmap. This was possible before 0.46, but 0.46 changed this and now only allows string cmaps or custom cmaps

``` py
import pyvista as pv
pv.global_theme.cmap = 'viridis'  # OK
pv.global_theme.cmap = pv.get_cmap_safe('viridis')  # Error
```

user27182:
### Overview

Fix #7786

This fix should also be patched into pyvista 0.46

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
   'pillow',
   'pooch',
   'scooby>=0.5.1',
-  'typing-extensions',
+  'typing-extensions>=4.10',
   'vtk!=9.4.0',
   'vtk!=9.4.1',
   'vtk<9.6.0',


tests/plotting/test_theme.py
```diff
@@ -364,13 +365,22 @@ def test_cmap(default_theme):
     default_theme.cmap = cmap
     assert default_theme.cmap == cmap
 
+    cmap_obj = pv.get_cmap_safe(cmap)
```
MatthewFlamm:
```suggestion
    assert isinstance(cmap_obj, mpl.colors.Colormap)
    cmap_obj = pv.get_cmap_safe(cmap)
```

user27182: Do you mean:
```py
assert isinstance(cmap, str)
```
Because this property doesn't do any type conversion, so `str` in, `str` out, `mpl.colors.Colormap` in, `mpl.colors.Colormap` out.

MatthewFlamm: > Because this property doesn't do any type conversion, so str in, str out, mpl.colors.Colormap in, mpl.colors.Colormap out.

This was my understanding.  The current line 368 outputs a `mpl.colors.Colormap`, so this means that there was a `mpl.colors.Colormap` coming in, right?  Maybe I misunderstand, but it would be good to be explicit about the input-output relationship here given that it was an actual regression that we want to prevent in the future.

user27182: Oh, yes there is already an instance check for the colormap object, so I thought you were referring to the previous lines.

I'll add an assertion check for the existing str input test.

MatthewFlamm: I think I was wrong.  `cmap` is a string here, but `cmap_obj` is `mpl.colors.Colormap`. I don't see a test for when the input is `mpl.colors.Colormap`.  If there is I can't tell and an assert should be used to make it explicit.

user27182: This is currently the test for `mpl.colors.Colormap` inputs.
``` py
    cmap_obj = pv.get_cmap_safe(cmap)    
    assert isinstance(cmap_obj, mpl.colors.Colormap)
    default_theme.cmap = cmap_obj
    assert default_theme.cmap == cmap_obj
```

Is this sufficient or did you have something else in mind

MatthewFlamm: Is this the test because `defualt_theme.cmap` uses `get_cmap_safe` under the hood?  It would be good to test `get_cmap_safe` directly as that is where the regression fix was.

user27182: I understand your point now. `get_cmap_safe` tests have been updated

MatthewFlamm: Thanks.  Looks better to me now.

MatthewFlamm (approved):

tkoyama010 (approved):

PR #7788, opened 2025-08-06
Fix camera position typing to allow nested sequence

user27182:
### Overview

Fix incomplete type hints for camera position

This should be backported to v0.46

### DIFF
--- a/pyvista/plotting/_typing.py
+++ b/pyvista/plotting/_typing.py
@@ -76,7 +76,10 @@ OpacityOptions = Literal[
 CullingOptions = Literal['front', 'back', 'frontface', 'backface', 'f', 'b']
 StyleOptions = Literal['surface', 'wireframe', 'points', 'points_gaussian']
 LightingOptions = Literal['light kit', 'three lights', 'none']
-CameraOptions = Literal['xy', 'xz', 'yz', 'yx', 'zx', 'zy', 'iso']
+CameraOptions = Union[
+    Literal['xy', 'xz', 'yz', 'yx', 'zx', 'zy', 'iso'],
+    Sequence[Union[list[float], tuple[float, float, float]]],
+]
 
 
 class BackfaceArgs(TypedDict, total=False):
--- a/pyvista/demos/demos.py
+++ b/pyvista/demos/demos.py
@@ -270,7 +270,7 @@ def plot_wave(fps=30, frequency=1, wavetime=3, notebook=None):  # noqa: PLR0917
     # Start a plotter object and set the scalars to the Z height
     plotter = pyvista.Plotter(notebook=notebook)
     plotter.add_mesh(mesh, scalars='Height', show_scalar_bar=False, smooth_shading=True)
-    plotter.camera_position = cpos  # type: ignore[assignment]
+    plotter.camera_position = cpos
     plotter.show(
         title='Wave Example',
         window_size=[800, 600],
@@ -432,7 +432,7 @@ def plot_beam(notebook=None):
         rng=[-d.max(), d.max()],
         cmap=cmap,  # type: ignore[arg-type]
     )
-    plotter.camera_position = cpos  # type: ignore[assignment]
+    plotter.camera_position = cpos
     plotter.add_text('Static Beam Example')
     plotter.show()
 


pyvista/plotting/_typing.py
```diff
@@ -76,7 +76,10 @@
 CullingOptions = Literal['front', 'back', 'frontface', 'backface', 'f', 'b']
 StyleOptions = Literal['surface', 'wireframe', 'points', 'points_gaussian']
 LightingOptions = Literal['light kit', 'three lights', 'none']
-CameraOptions = Literal['xy', 'xz', 'yz', 'yx', 'zx', 'zy', 'iso']
+CameraOptions = Union[
+    Literal['xy', 'xz', 'yz', 'yx', 'zx', 'zy', 'iso'],
+    Sequence[Union[list[float], tuple[float, float, float]]],
```
MatthewFlamm: Renderer.camera_position and Plotter.camera position both support more than this.  For example, this modified example from the `camera_position` docstring.  Note the `ndarray`s being passed.

```py
>>> import numpy as np
>>> import pyvista as pv
>>> mesh = examples.download_bunny_coarse()
>>> pl = pv.Plotter()
>>> _ = pl.add_mesh(mesh, show_edges=True, reset_camera=True)
>>> pl.camera_position
[(0.02430, 0.0336, 0.9446),
 (0.02430, 0.0336, -0.02225),
 (0.0, 1.0, 0.0)]
>>> pl.camera_position = [
...     np.array((0.3914, 0.4542, 0.7670)),
...     np.array((0.0243, 0.0336, -0.0222)),
...     np.array((-0.2148, 0.8998, -0.3796)),
... ]
>>> pl.show()
```

user27182: We should use `Sequence[VectorLike[float]]` then? And I think there is a CameraPosition class we should add too, assuming it also works at runtime

MatthewFlamm: > We should use `Sequence[VectorLike[float]]` then? And I think there is a CameraPosition class we should add too, assuming it also works at runtime

I hope it works otherwise `pl.camera_position = pl.camera_position` will fail. :)

user27182: > We should use `Sequence[VectorLike[float]]` then? And I think there is a CameraPosition class we should add too, assuming it also works at runtime

Turns out a 2D array is also supported, so we should use `MatrixLike[float]` here. This is also valid:

```  py
import numpy as np
import pyvista as pv
from pyvista import examples
mesh = examples.download_bunny_coarse()
pl = pv.Plotter()
pl.add_mesh(mesh, show_edges=True, reset_camera=True)
pl.camera_position = np.array((
     (0.3914, 0.4542, 0.7670),
     (0.0243, 0.0336, -0.0222),
     (-0.2148, 0.8998, -0.3796),
))
pl.show()

Comment (MatthewFlamm) [reply pyvista/plotting/_typing.py]: Just about to write the same thing.  I agree, let's just make it `MatrixLike[float]`

Comment (user27182) [reply pyvista/plotting/_typing.py]: Indeed `CameraPosition` also works.
``` py
import numpy as np
import pyvista as pv
from pyvista import examples
mesh = examples.download_bunny_coarse()
pl = pv.Plotter()
pl.add_mesh(mesh, show_edges=True, reset_camera=True)
pl.camera_position = pv.CameraPosition(
    position=(0.3914, 0.4542, 0.7670), 
    focal_point=(0.0243, 0.0336, -0.0222), 
    viewup=(-0.2148, 0.8998, -0.3796)
)
pl.show()
```

Will update this to include `MatrixLike[float] | CameraPosition`

user27182: Oh boy I didn't realize there were so many options supported by this, a single vector is also allowed, so we should include `VectorLike[float]` on its own too
``` py
import numpy as np
import pyvista as pv
from pyvista import examples
mesh = examples.download_bunny_coarse()
pl = pv.Plotter()
pl.add_mesh(mesh, show_edges=True, reset_camera=True)
pl.camera_position = (0.3914, 0.4542, 0.7670)
pl.show()
```

MatthewFlamm: I never liked this API design.  But it is probably too baked to make a large change now?

MatthewFlamm: We also aren't going to change it in a PR meant as a backport to a release either way.

user27182: I _always_ get confused with camera position. I never even knew that an actual `CameraPosition` object was returned, so in this regard I think changing `CameraPosition.__repr__` to show the class name and actual attribute names instead of just a list of tuples would go a long way.

I think a key issue is also that `CameraPosition` itself has a `position` attribute. So we have some confusing redundancies:

``` py
import pyvista as pv
pl = pv.Plotter()
pl.camera.position
pl.camera_position.position
```

So sometimes camera position is just a position (i.e. length-3 vector), sometimes it's all of position and focal point and view up.

MatthewFlamm: To add to the confusion, functions like `show` use the parameter `cpos`, which is used to set `camera_position` attribute.  `cpos`  is heavily ingrained in the examples and documentation, and it likely used in a large number of downstream codes.  This gives me pause to deprecate or remove it.

https://github.com/pyvista/pyvista/blob/d987297b012fe90ddfff9efcc524628cd53b4d25/pyvista/_plot.py#L80-L81

MatthewFlamm (approved):
LGTM thanks!

doc/source/conf.py
```diff
@@ -228,6 +228,7 @@
     (r'py:.*', 'BackfaceArgs'),
     (r'py:.*', 'CullingOptions'),
     (r'py:.*', 'OpacityOptions'),
+    (r'py:.*', 'CameraPosition'),
```
user27182: I think this needs to include `Options`
```suggestion
    (r'py:.*', 'CameraPositionOptions'),
```

Issue #6511, opened 2024-08-15
`clip` filter with `return_clipped=True` returns mesh with same bounds as input


user27182:
### Describe the bug, what's wrong, and what you expected.

A clipped mesh is expected to have smaller bounds than the input from which it was clipped. But, using `return_clipped` returns a mesh with the same bounds as the input (see `x_max` bound below). The plot still looks fine either way though, which made this hard to detect. This was discovered while trying to create a new mesh from the clipped bounds (e.g. `pv.Box(bounds=mesh.bounds)`)

### Steps to reproduce the bug.

``` python
import pyvista as pv
cube = pv.Cube().triangulate().subdivide(3)

clipped_cube = cube.clip(return_clipped=False)
print(clipped_cube.bounds)
# Clipped in half, so x_max bound is 0.0 not 0.5 as expected
# BoundsTuple(x_min=-0.5, x_max=0.0, y_min=-0.5, y_max=0.5, z_min=-0.5, z_max=0.5)

clipped_cube, _ = cube.clip(return_clipped=True)
print(clipped_cube.bounds)
# Clipped in half, yet x_max bound is still 0.5, same as input
# BoundsTuple(x_min=-0.5, x_max=0.5, y_min=-0.5, y_max=0.5, z_min=-0.5, z_max=0.5)
```

### System Information

```shell
OS : Darwin (macOS 14.4.1)
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : Apple
      GPU Renderer : Apple M2
       GPU Version : 4.1 Metal - 88
  MathText Support : True

  Python 3.9.13 (v3.9.13:6de2ca5339, May 17 2022, 11:37:23)  [Clang 13.0.0
  (clang-1300.0.29.30)]

           pyvista : 0.45.dev0
               vtk : 9.3.1
             numpy : 2.0.0
        matplotlib : 3.9.0
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 10.4.0
           IPython : 8.18.1
          colorcet : 3.1.0
             scipy : 1.13.1
              tqdm : 4.66.4
            meshio : 5.3.5
    pytest_pyvista : 0.1.9
```

MatthewFlamm:
[...]

Issue #6098, opened 2024-05-17
Crinkle Clip returns UnstructuredGrid instead of PolyData


kalingibbons:
### Describe the bug, what's wrong, and what you expected.

Using the `crinkle=True` keyword in the `clip` filter returns an `UnstructuredGrid` object. I expected the `clip` filter to always return a `PolyData` object.

### Steps to reproduce the bug.

```python
dataset = pv.examples.download_action_figure()
dataset.clip('x', crinkle=False)  # returns PolyData
dataset.clip('x', crinkle=True)  # returns UnstructuredGrid)
```
![image](https://github.com/pyvista/pyvista/assets/11838956/a3d1a1e7-6973-4fb9-abc3-5d7976df42f7)

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Fri May 17 15:35:58 2024 Mountain Daylight Time

                OS : Windows (10 10.0.19045 SP0 Multiprocessor Free)
            CPU(s) : 8
           Machine : AMD64
      Architecture : 64bit
               RAM : 39.9 GiB
       Environment : Python
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : NVIDIA GeForce RTX 2070 SUPER/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 536.99
  MathText Support : False

  Python 3.11.9 | packaged by conda-forge | (main, Apr 19 2024, 18:27:10) [MSC
  v.1938 64 bit (AMD64)]

           pyvista : 0.43.8
               vtk : 9.2.6
             numpy : 1.26.4
        matplotlib : 3.8.4
            scooby : 0.10.0
             pooch : 1.8.1
            pillow : 10.0.0
           imageio : 2.34.1
             PyQt5 : 5.15.9
           IPython : 8.24.0
        ipywidgets : 8.1.1
             scipy : 1.13.0
              tqdm : 4.66.4
            meshio : 5.3.5
        jupyterlab : 4.2.0
      nest_asyncio : 1.6.0
--------------------------------------------------------------------------------
```

user27182:
#6060 was recently patched into version `0.43.8`, not sure if that affected the output.  @kalingibbons do you get this same bug with `0.43.7` ?

kalingibbons:
I was actually on 0.43.3 when I first ran into it, then updated to see if it was fixed before submitting the issue.

user27182:
Ah ok thanks for checking, I guess this bug has been there for a while then.

[...]

PR #7789, opened 2025-08-06
Remove unused points after using `clip` or `clip_box`

### LINKED ISSUE #7738: clip box not working as expected

I'm using [clip box](https://docs.pyvista.org/api/core/_autosummary/pyvista.dataobjectfilters.clip_box#pyvista.DataObjectFilters.clip_box) and noticing that the bounds of the clipped object that is returned are not updated to match the clip, even though the visualization looks correct. Am I misunderstanding how clip_box should be used? I'm trying to create a thick slice along the z-axis.

In the cube example, the bounds are those of the original object (-0.5, 0.5) and not of the clipped surface (0.2, 0.4).

```python
import pyvista as pv
test = pv.Cube()

clipped = test.clip_box(bounds=(test.bounds.x_min, test.bounds.x_max, test.bounds.y_min, test.bounds.y_max, 0.2, 0.4), invert=False)

clipped.bounds
```

### System Information

```shell
  Date: Sun Jul 20 15:37:12 2025 PDT

                OS : Darwin (macOS 15.5)
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
               RAM : 24.0 GiB
       Environment : Jupyter
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M2
       GPU Version : 4.1 Metal - 89.4
  MathText Support : True

  Python 3.13.5 (main, Jun 11 2025, 15:36:57) [Clang 17.0.0
  (clang-1700.0.13.3)]

           pyvista : 0.45.3
               vtk : 9.4.2
             numpy : 2.3.1
        matplotlib : 3.10.3
            scooby : 0.10.1
             pooch : 1.8.2
            pillow : 11.3.0
           IPython : 9.4.0
        ipywidgets : 8.1.7
             trame : 3.10.2
      trame_client : 3.9.1
      trame_server : 3.5.0
         trame_vtk : 2.9.0
     trame_vuetify : 3.0.1
      nest_asyncio : 1.6.0
```

### Screenshots

Image

Image

user27182:
### Overview

Fix https://github.com/pyvista/pyvista/issues/6511 and fix #7738


tests/core/test_dataobject_filters.py
```diff
@@ -162,6 +171,22 @@ def test_clip_box(datasets):
...
+def test_clip_box_no_unused_points(as_composite):
+    mesh = pv.Cube()
+    mesh = pv.MultiBlock([mesh]) if as_composite else mesh
+    new_bounds = (
+        mesh.bounds.x_min,
+        mesh.bounds.x_max,
+        mesh.bounds.y_min,
+        mesh.bounds.y_max,
+        0.2,
+        0.4,
```
tkoyama010: I would like to clearly define what these magic number means.
```suggestion
        mesh.bounds.z_min + (mesh.bounds.z_max - mesh.bounds.z_min) * 7 / 10,
        mesh.bounds.z_min + (mesh.bounds.z_max - mesh.bounds.z_min) * 8 / 10,
```

tkoyama010 (changes requested):

tkoyama010 (dismissed):
LGTM

MatthewFlamm:
The current fix here would modify meshes that intentionally have unused points.  The current PR does not clean PointSet, but you can have unused points in PolyData too for example.

```py
a = pv.PolyData()
a.points = ((0., 0., 0.),)
a
```

```txt
 PolyData (0x21240b12260)
  N Cells:    0
  N Points:   1
  N Strips:   0
  X Bounds:   0.000e+00, 0.000e+00
  Y Bounds:   0.000e+00, 0.000e+00
  Z Bounds:   0.000e+00, 0.000e+00
  N Arrays:   0
```

MatthewFlamm:
> The current fix here would modify meshes that intentionally have unused points. The current PR does not clean PointSet, but you can have unused points in PolyData too for example.
> 

If this is indeed an issue, we could add in optional parameters to clean unused points with the default to True.

user27182:
Adding a keyword for this would be better. Will update

user27182:
> > > The current fix here would modify meshes that intentionally have unused points. The current PR does not clean PointSet, but you can have unused points in PolyData too for example.
> > 
> > If this is indeed an issue, we could add in optional parameters to clean unused points with the default to True.

> Adding a keyword for this would be better. Will update

I tried adding this option, but in initial testing, I found that unused points were removed even if `remove_unused_points=False`.  It looks like the `clip` filter _already_ removes unused points sometimes, e.g. from `main`:
``` py
import pyvista as pv

poly = pv.examples.cells.Triangle().extract_geometry()

# Triangle is clipped and a (smaller) triangle is returned
assert poly.n_cells == 1
assert poly.n_points == 3
clipped = poly.clip()
assert clipped.n_cells == 1
assert clipped.n_points == 3

# One of three vertex cells are removed
poly_points = poly.cast_to_poly_points()
assert poly_points.n_cells == 3
assert poly_points.n_points == 3
clipped = poly_points.clip()
assert clipped.n_cells == 2
assert clipped.n_points == 2

# All unused points are removed
poly_unused_points = pv.PolyData()
poly_unused_points.points = poly.points
assert poly_unused_points.n_cells == 0
assert poly_unused_points.n_points == 3
clipped = poly_unused_points.clip()
assert clipped.n_cells == 0
assert clipped.n_points == 0
```
Note how all three unused points get removed in the last example, but based on the second `poly_points` examples, if the `clip` filter actually respected unused points, we should expect only one unused point to be removed.

Based on this, I don't think we can reliably expect this filter to keep any unused points, and using `remove_unused_points` internally in post-processing (without a keyword option) is really just a patch to fix buggy output from the clip to ensure the filter _actually_ removes all unused points all the time, instead of only sometimes.

user27182:
For completeness, I tried with `PointSet` as well:
``` py
pset = poly.cast_to_pointset()
assert poly_unused_points.n_cells == 0
assert poly_unused_points.n_points == 3
clipped = poly_unused_points.clip()
assert clipped.n_cells == 0
assert clipped.n_points == 0
```

And we can see that it just clips everything. So this filter is currently useless for `PointSet`. Based on this, I'll remove the instance check for this and add a test to confirm the PointSet output is empty.

MatthewFlamm:
Great investigation, maybe the solution is to document this behavior as a feature along with the changes in this PR.

user27182:
> For completeness, I tried with `PointSet` as well:
> 
> ```python
> pset = poly.cast_to_pointset()
> assert poly_unused_points.n_cells == 0
> assert poly_unused_points.n_points == 3
> clipped = poly_unused_points.clip()
> assert clipped.n_cells == 0
> assert clipped.n_points == 0
> ```
> 
> And we can see that it just clips everything. So this filter is currently useless for `PointSet`. Based on this, I'll remove the instance check for this and add a test to confirm the PointSet output is empty.

There is a bug in this code! We clipped `poly_unused_points`, not `pset`


Clipping `PointSet` _does_ work after all:
``` py
import pyvista as pv
pointset = pv.examples.load_ant().cast_to_pointset()
assert pointset.n_points == 486
clipped = pointset.clip()
assert isinstance(clipped, pv.PointSet)
assert clipped.n_points == 242
```
But notice that the output type is `PointSet`, and _not_ `UnstructuredGrid`, which is what's otherwise documented. And similar to #6098, using `crinkle=True` breaks this filter and a `KeyError` is raised


In terms of removing unused points, we should definitely avoid removing them for `PointSet`, but otherwise we should always remove them, even for PolyData, since the filter sometimes removes them anyway.

user27182:
> But notice that the output type is `PointSet`, and _not_ `UnstructuredGrid`, which is what's otherwise documented. And similar to #6098, using `crinkle=True` breaks this filter and a `KeyError` is raised

While writing tests, I also found that `vtkTableBasedClipDataSet` has a `PointSet`  -> `PointSet` mapping, but `MultiBlock[PointSet]` -> `MultiBlock[UnstructuredGrid]` mapping. This means we have to cast the output type... and if we're doing that we might as well do this generally and include casting for `PolyData` -> `PolyData` and therefore also fix #6098. I'll open a separate PR with just the type casting fix

tkoyama010 (approved):
LGTM

PR #7794, opened 2025-08-07
Make `CameraPosition.__repr__` explicit

user27182:
### Overview

Originally mentioned here https://github.com/pyvista/pyvista/pull/7788#discussion_r2261087580
This PR makes it obvious that we're dealing with an object with three attributes

### DIFF
--- a/pyvista/plotting/renderer.py
+++ b/pyvista/plotting/renderer.py
@@ -258,7 +258,11 @@ class CameraPosition(_NoNewAttrMixin):
 
     def __repr__(self) -> str:
         """List representation method."""
-        return '[{},\n {},\n {}]'.format(*self.to_list())
+        return (
+            f'{CameraPosition.__name__}(position={self._position},\n'
+            f'               focal_point={self._focal_point},\n'
+            f'               viewup={self._viewup})'
+        )
 
     def __getitem__(self, index):
         """Fetch a component by index location like a list."""


MatthewFlamm (approved):

MatthewFlamm:
Does this have a knock on effect for the way that `return_cpos=True` from `Plotter.show` is sometimes used?  For example, a common use case, at least for me is to use `return_cpos=True` and then copy and paste the value into my code to use as the starting point.  I _think_ this change would remove the ability of quick copy-paste?

user27182:
> Does this have a knock on effect for the way that `return_cpos=True` from `Plotter.show` is sometimes used? For example, a common use case, at least for me is to use `return_cpos=True` and then copy and paste the value into my code to use as the starting point. I _think_ this change would remove the ability of quick copy-paste?

Yes you're right, this would change that. You should still be able to copy/paste it as-is and use it, since the repr should be useable to init a new `CameraPosition` object but will need to prepend `pv.`

We could keep the old behavior with `__str__`. Using `print(cpos)` will call `__str__`, so you can keep the old representation that way.

user27182:
Both `str` and `repr` are implemented now. The output is  consistent with what NumPy does in the sense that `array()` is included in the `repr`, but not `str` output

NDArray:
``` py
repr(np.array((1,)))  # 'array([1])'
str(np.array((1,)))  # '[1]'
```
CameraPosition
``` py
# repr
CameraPosition(position=(0.3914, 0.4542, 0.767),
               focal_point=(0.0243, 0.0336, -0.0222),
               viewup=(-0.2148, 0.8998, -0.3796))

# str
[(0.3914, 0.4542, 0.767),
 (0.0243, 0.0336, -0.0222),
 (-0.2148, 0.8998, -0.3796)]
```

MatthewFlamm:
I'm +0.2 in favor of this (out of -1 to +1). It is now easier to know what the numbers refer to.  One hesitation is that, I assume that, there are a decent number of people who are using environments that show the repr, and that they are using `return_cpos` to find the final camera position.  For example `ipython` and our own documentation building.  One way to soften this blow would be to support this with better documentation.  An example is that current `show` has this in the example.  There are likely more of this throughout the documentation.

https://github.com/pyvista/pyvista/blob/9b3b97a98242c5de8c48d24b84e809c444481daa/pyvista/plotting/plotter.py#L7155-L7158

user27182:
That's a good find, I wonder why that doctest is skipped.

Will do a search for cpos and try to update those examples. We can show both use cases, and maybe use `print` instead of `str` in the docs as that may also be more intuitive.

Users _can_ still copy/paste and reuse the output as new input for a cpos keyword, so that functionality isn't really lost. And now if they need to make minor tweaks to `CameraPosition.position`, it is immediately obvious to users that (1) camera position is more than just a position vector, and (2) it's clear which of the three vectors needs to be changed.

user27182:
> That's a good find, I wonder why that doctest is skipped.

It's skipped because it raises a RuntimeError otherwise:
https://github.com/pyvista/pyvista/actions/runs/16918236508/job/47937498711#step:10:2145
``` py
RuntimeError: This plotter has been closed and cannot be shown. [docutils]
```

user27182:
> One hesitation is that, I assume that, there are a decent number of people who are using environments that show the repr, and that they are using return_cpos to find the final camera position. For example ipython and our own documentation building. One way to soften this blow would be to support this with better documentation. An example is that current show has this in the example. There are likely more of this throughout the documentation.

`CameraPosition` is now used explicitly in all examples to set cpos instead of a list of tuples. @MatthewFlamm does this address your concerns?

user27182:
This PR has two main updates: one is to replace cpos inputs in the docs to a use a more explicit "CameraPosition" object. This is not a breaking change at all, and is strictly a documentation change. 

The second change is to modify `CameraPosition.__str__`. It seems that this is why a "breaking change" label was added?

What if we remove _all_ custom __str__ and __repr__ methods, and we make the CameraPosition class inherit from NamedTuple instead? The string and repr outputs would then automatically be defined by NamesTuple. Would that help move this PR forward? @tkoyama010 @MatthewFlamm?

Shall I split off the less controversial change to the docs as a separate PR?

user27182:
> This PR has two main updates: one is to replace cpos inputs in the docs to a use a more explicit "CameraPosition" object. This is not a breaking change at all, and is strictly a documentation change.

> Shall I split off the less controversial change to the docs as a separate PR?

See #8045 for the docs-only changes

pyvista/plotting/plotter.py
```diff
@@ -1883,17 +1883,31 @@ def camera_position(self) -> CameraPosition:  # numpydoc ignore=RT01
...
+        >>> cpos = pl.camera_position
+
+        Show the camera position.
+
+        >>> cpos
+        CameraPosition(position=(0.02430, 0.0336, 0.9446),
+                       focal_point=(0.02430, 0.0336, -0.02225),
+                       viewup=(0.0, 1.0, 0.0))
+
+        Use ``print`` or ``str`` to obtain a list representation instead.
```
MatthewFlamm, line 1916: It would be useful to show one example before this `str` section to use `CameraPosition` to set the attribute.

MatthewFlamm:
Sorry for losing the thread here.  I'm +1 here.  Users only need to prepend `pv.` which significantly reduces my concern.  I added in one comment, and I'm +1 to move this ahead.  I also don't think it is a true breaking change, but also fine keeping it.

pyvista/plotting/plotter.py
user27182: Is this the kind of change you had in mind? https://github.com/pyvista/pyvista/pull/7794/commits/2e5cb6e2d3d15e6c25e6cfdfe90240755d5b609d

akaszynski (approved):
Agree with the changes here. It's always been a bit confusing for me when returning a camera position that `repr`s just like a list of tuples but is really a `CameraPosition`.

pyvista/plotting/plotter.py
akaszynski: Calling this resolved.

PR #7799, opened 2025-08-08
Use PEP 673 Self type for methods returning self instances

tkoyama010:
## Summary
- Improves type annotations by using [PEP 673](https://peps.python.org/pep-0673/) `Self` type for methods that return their own instance
- Provides better type safety and more precise type inference for subclasses
- Uses existing `typing_extensions.Self` import for Python 3.9+ compatibility

## Compatibility
Supports Python 3.9+ using `typing_extensions.Self` as already imported throughout the codebase.

### DIFF
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -186,7 +186,7 @@ class _PointSet(DataSet):
         target.RemoveGhostCells()
         return target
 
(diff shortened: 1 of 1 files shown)


tkoyama010:
@pyvista-bot LGTM

Issue #6141, opened 2024-05-29
Inconsistent rendering of ChartMPL plots


user27182:
### Describe the bug, what's wrong, and what you expected.

There is an issue with how `pyvista.ChartMPL` plots are rendered. E.g. in the [ChartMPL docs](https://docs.pyvista.org/version/stable/api/plotting/charts/_autosummary/pyvista.ChartMPL.html#pyvista.ChartMPL), this is the image displayed. Immediately it's clear there is an issue with the plot -- everything seems a bit zoomed in.

<img src="https://d33wubrfki0l68.cloudfront.net/32a06ec070f89b4d5d37cfc0bc8bf3e6e9fbd909/11061/_images/pyvista-chartmpl-2_00_00.png" width="400">

The plot above is from an older (stable) version of the docs for 0.43. However, this [dev version](https://github.com/pyvista/pyvista/blob/197649a5920f0b7e5378ca8f0ba92427d2d65b67/tests/doc/doc_image_cache/pyvista-ChartMPL-2_00_00.jpg) of the plot was recently committed to the `doc_image_cache` as part of #5718. This plot looks "correct".
![image](https://github.com/pyvista/pyvista/assets/89109579/5b68e07e-935f-41fc-8cdb-c7d4b666c4e0)

This was discovered when reviewing recent failed tests from #5712 (see the build failure [here](https://github.com/pyvista/pyvista/actions/runs/9278295702)). In the [doc-debug-images-failed artifact (download)](https://github.com/pyvista/pyvista/actions/runs/9278295702/artifacts/1546716153), the build image is once again appearing incorrectly as the "zoomed in" version.

So, the bug is that the generated plot is not consistent. Sometimes it gets zoomed in (incorrectly), sometimes it plots just fine. Not sure what the cause is or how to reproduce it.

The same bug applies for other ChartMPL plots, e.g. this is the one currently showing in the stable version of the docs for [active_border_color](https://docs.pyvista.org/version/stable/api/plotting/charts/_autosummary/pyvista.chartmpl.active_border_color#pyvista.ChartMPL.active_border_color) method (also with "correct" versions committed to the cache, and also failing in the linked CI documentation build test)
![image](https://github.com/pyvista/pyvista/assets/89109579/b898818b-fd69-44e1-a413-20ca8b6a798d)

### Steps to reproduce the bug.

Not sure how to reproduce.

### System Information

```shell
See software report for CI Build Documentation.
```

user27182:
[...]

PR #7800, opened 2025-08-09
Enable parallel doc build

akaszynski:
This PR makes parallel builds possible.

The main blocker to parallel builds is the sequential ordering of our images in:

https://github.com/pyvista/pyvista/blob/9b3b97a98242c5de8c48d24b84e809c444481daa/pyvista/ext/plot_directive.py#L593-L594

This is not thread safe as it's possible for images to be overwritten since each document can be fed into multiple workers.

The fix is to make the base name deterministic at runtime by hashing the source of the code used in the plot directive. This has a side benefit of making our image tests a bit less brittle since images are named based on the content of the source rather than the location within a rst file.

---

This change required some refactoring of `make_tables` since it's dependent on filenames based on indexing rather than hashes.


akaszynski:
Played around with the gallery quite a bit with this PR, nice work @user27182!

A quick question regarding `make_tables.py`: Is it necessary to download and load each dataset when creating the cards? It seems like it's only necessary to create the cards since the datasets will be downloaded (or use the cache) when the screenshots of the datasets are generated. This way we can make the script much lighter and take advantage of the parallel build when downloading and loading in the datasets.

user27182:
> A quick question regarding `make_tables.py`: Is it necessary to download and load each dataset when creating the cards? It seems like it's only necessary to create the cards since the datasets will be downloaded (or use the cache) when the screenshots of the datasets are generated. This way we can make the script much lighter and take advantage of the parallel build when downloading and loading in the datasets.

Yes they're only downloaded and loaded to create the cards, which includes creating the screenshot and also collecting file and dataset metadata.

Doesn't matter to me exactly when they're downloaded/loaded as long as everything still works.

MatthewFlamm:
Hashing the source code means that any small change to an example will require an image name change.  Not sure there is an easy way to fix this, and it is a minor inconvenience compared to shortening the long build times, assuming it indeed does this.

It is possible that we could have duplicate source code in some cases. I'm thinking about a common pattern like below, which could create unique images.  In these cases, it might be useful to distinguish via comment, but it looks like the current PR might discard comments?

```py
pl = pv.Plotter()
pl.add_mesh(mesh) 
pl.show()
```

akaszynski:
> Hashing the source code means that any small change to an example will require an image name change. Not sure there is an easy way to fix this, and it is a minor inconvenience compared to shortening the long build times, assuming it indeed does this.

Quite right. However, changing the position of a plot directive (or adding one) within a single document invalidates the order of every following example image name. Also, changes to an example might (but are not guaranteed) to change the image output as well, implying that the image cache needs to be changed anyway.

> It is possible that we could have duplicate source code in some cases. I'm thinking about a common pattern like below, which could create unique images. In these cases, it might be useful to distinguish via comment, but it looks like the current PR might discard comments?
> 
> ```python
> pl = pv.Plotter()
> pl.add_mesh(mesh) 
> pl.show()
> ```

This may be a "feature" and not a bug where identical source generates identical images, leading to reuse. Given the previous comment, we can reduce the needless change of hashes by eliminating comments (and doctest output). Furthermore, collisions are unlikely since each image is prepend by its module and function if generated from an autosummary, or the source file name if generated from a rst.

---

Alternatively, we might consider hashing the content of the image itself. This is likely to also be brittle as there are minor changes depending on the backend (e.g. `VTK_DEFAULT_OPENGL_WINDOW=vtkEGLRenderWindow`). One possible approach would be to down-sample both the image resolution and color.

MatthewFlamm:
Since this PR allows the previous method, users could revert back if it is a real issue.

user27182:
> This change required some refactoring of `make_tables` since it's dependent on filenames based on indexing rather than hashes.


The docs for `CellType` are also dependent on image names (the docs reuse images from examples.Cells). Not sure if this will be an issue but as a workaround we could generate the images twice instead of reusing them.

akaszynski:
There's a side effect to this PR in the ChartMPL docs in that it resets the scaling of the MPL figures:

### `main`
image

### This branch
image

Are we fine with this change? This resolves https://github.com/pyvista/pyvista/issues/6141

akaszynski:
Also, good news, build time seems to work in parallel on GH infrastructure, with the "Build Documentation" step going from **1h 39m 21s** to **1h 14m 41s**.

Locally, I can now build fresh docs in 8 minutes, and I'm hoping to move the doc build to a self-hosted agent (https://github.com/pyvista/pyvista-sh/pull/2). We can discuss the ramifications of that down the line, but it's an option to reduce build times more and lighten the load from our public agents.

user27182:
> There's a side effect to our ChartMPL docs in that it resets the scaling of the MPL figures:

> Are we fine with this change? This resolves #6141

This issue is still unresolved... these tests are flaky,  sometimes they're zoomed in, sometimes not. This PR should update the filenames in `tests/doc/flaky_tests` instead.

IMO the images in `doc_image_cache` should remain as the non-zoomed in version, and the flaky ones should be the zoomed in versions

user27182:
> > There's a side effect to our ChartMPL docs in that it resets the scaling of the MPL figures:
> 
> > Are we fine with this change? This resolves #6141
> 
> This issue is still unresolved... these tests are flaky, sometimes they're zoomed in, sometimes not. This PR should update the filenames in `tests/doc/flaky_tests` instead.
> 
> IMO the images in `doc_image_cache` should remain as the non-zoomed in version, and the flaky ones should be the zoomed in versions


Or, if this PR does indeed fix this and the tests are no longer flaky, we should maybe address that separately after removing the flaky test images completely and confirming with multiple successful build runs. Was the zoomed-in image used in the cache (i.e. that one that caused a build failure) previously copied from the github build artifact of generated images, or was the zoomed-in image generated locally ?

akaszynski:
> Was the zoomed-in image used in the cache (i.e. that one that caused a build failure) previously copied from the github build artifact of generated images, or was the zoomed-in image generated locally ?

Generated remotely on GitHub. I was surprised too, running it again with the cached images from github...

user27182:
This is maybe more of a nice-to-have, but is it possible to only rename image files in the cache instead of replacing them completely? Not sure if that's really feasible, might have to create a temporary internal mapping from counter-style names to hash-based names...
From a review perspective, this will give us much more confidence that nothing changed unexpectedly if all we see is file renames instead of files being deleted and re-added.

Or if it's too much work to do rename-only, maybe we can split any changes that affect any example or docstring code into a separate PR? E.g. maybe updates to ChartMPL examples or warp-by-vector-eigenmodes should be made separately to isolate any actual changes to the images from pseudo changes caused by using the hash filenames.

akaszynski:
> This is maybe more of a nice-to-have, but is it possible to only rename image files in the cache instead of replacing them completely? Not sure if that's really feasible, might have to create a temporary internal mapping from counter-style names to hash-based names... From a review perspective, this will give us much more confidence that nothing changed unexpectedly if all we see is file renames instead of files being deleted and re-added.

It's a reasonable point. I think it's feasible, it might have to be two step where we first rename and then modify.

> Or if it's too much work to do rename-only, maybe we can split any changes that affect any example or docstring code into a separate PR? E.g. maybe updates to ChartMPL examples or warp-by-vector-eigenmodes should be made separately to isolate any actual changes to the images from pseudo changes caused by using the hash filenames.

Also a good idea regardless of the outcome of the image cache rename. This PR has gotten more complex than I had expected.

.github/workflows/docs.yml
```diff
@@ -136,7 +140,7 @@ jobs:
       - name: Build I18N Documentation
         if: startsWith(github.ref, 'refs/heads/main') || startsWith(github.ref, 'refs/tags/v')
         run: |
-          make -C doc mini18n-html
+          make -C doc mini18n-html SPHINXOPTS="-j2" # Public GitHub runners have 2 real cores
```
user27182: Does this also need to invoke `gen-external-examples` and `gen-dataset-gallery`? Or maybe not since the RST files are already there from before and don't need to be re-generated? 
If we use `pyvista-sh` and update the number of cores, how can we ensure _both_ args `SPHINXOPTS="-j2"` get updated?

Having two commands with similar arguments that _should_ be kept in-sync but which don't use some sort of way to keep them in sync (e.g. define the options exactly once and reuse them) is probably going to result in out-of-sync commands and docs builds in the future.

There are a lot of instances of this in our CI (e.g. vtk master testing workflow is easily overlooked when updating testing and development workflow). My preference would be to consider something like #7441 to centralize everything, including doc build commands. 

In the interim though, maybe we can define a global variable and reuse them for both the regular doc build and the `mini18n-html` build?

akaszynski: > Does this also need to invoke gen-external-examples and gen-dataset-gallery? Or maybe not since the RST files are already there from before and don't need to be re-generated?

Would be best to run it anyway. At worst it will say the directory already exists and it will proceed without changing anything.

> Having two commands with similar arguments that should be kept in-sync but which don't use some sort of way to keep them in sync (e.g. define the options exactly once and reuse them) is probably going to result in out-of-sync commands and docs builds in the future.

Good point. One way to do this is to follow this approach:

https://github.com/pyvista/pyvista-sh/blob/3820a6d0d1b276f2065a1354d477e0271136179b/.github/workflows/docs-sh.yml#L56-L64

This way the number of cores is both deterministic based on environment and the configuration is shared across workflow steps.

user27182:
For the doc image test failure with `Capusle`, that test is flaky, so probably just need to rename the files in `tests/doc/flaky_tests`

doc/source/make_tables.py
```diff
@@ -98,6 +98,8 @@ def __str__(self) -> str:
     'dual_sphere_animation': '.gif',
 }
 
+PLACEHOLDER = 'GALLERY-URI-PLACEHOLDER'
```
user27182: Something isn't quite working yet with the placeholder images, the dataset gallery images are not loading
image

https://689bf4cb557e63fdc2a936d3--pyvista-dev.netlify.app/api/examples/dataset_gallery

akaszynski: Thanks for checking that. I'll fix it.

user27182: Looks like it's fixed!
img

pyvista/ext/plot_directive.py
```diff
@@ -123,11 +123,25 @@
...
 
+
+**Directive Configuration Settings**
+
+Globally, you can set if the file names should be either:
+
+* Deterministic, based on directive source hash:
+  ``<BASENAME>-<HASH>_<INDEX>_<SUBINDEX>.<EXT>`` (Default)
+* Indexed, based on location in document:
+  ``<BASENAME>-<DOC-INDEX>_<INDEX>_<SUBINDEX>.<EXT>``
```
user27182, line 134: I think hash is good, CI is green so this clearly works, though I wonder if we could have it both ordered _and_ deterministic by using the first line number of the docstring's code instead. The line number would ensure order is kept in when sorting the image names, while also being independent of the execution order from the workers.

Just an idea, don't need to act on it, but something to consider in case we do need to know image order.

pyvista/ext/plot_directive.py
```diff
@@ -123,11 +123,25 @@
...
+
+Enable indexed naming this by setting ``pyvista_plot_use_counter=True``. Note
+that indexed is incompatible with parallel builds due to race conditions.
+
```
user27182, line 138:
```suggestion

.. versionchanged:: 0.47
    Hash-based image naming is now used by default.
```

pyvista/ext/plot_directive.py
akaszynski: It's a good idea. However, adding or removing lines to the rst document before the plot directive would result in a change of the image name.

user27182: True, I think this could maybe be accounted for, i.e. first parse the whole docstring and remove all lines irrelevant to the plot directive, i.e. only keep sections with `>>>` that include `show` or `plot`, and count the number of sections. (I think the plot directive currently does this anyway, but only sequentially and not all at once).

So then instead of using the actual line number, we use our own internal section number instead.
Basically, the docstring is static and we should be able to know how many images will be plotted before-hand, and should therefore also be able to determine which section is currently being processed (and hence still provide a deterministic section number to it).

Not a big deal though, might only matter in very few cases (e.g. if multiple images are generated for an example in `downloads.py`, we currently can't choose which image will be used for the dataset gallery).

akaszynski: > first parse the whole docstring and remove all lines irrelevant to the plot directive, i.e. only keep sections with >>> that include show or plot, and count the number of sections. (I think the plot directive currently does this anyway, but only sequentially and not all at once).

Yes, but what happens if we:

```py
import pyvista as pv

pl = pv.Plotter()
act = pl.add_mesh(pv.Sphere())
act.prop.color='blue'
pl.show()
```

I wish there was a way to make this deterministic based on only on lines critical to plotting, but it would be quite challenging.

. So then instead of using the actual line number, we use our own internal section number instead.
Basically, the docstring is static and we should be able to know how many images will be plotted before-hand, and should therefore also be able to determine which section is currently being processed (and hence still provide a deterministic section number to it).

Alternatively, we could write out the line numbers with each file first with ``<BASENAME>-<LINE-NUMBER>_<INDEX>_<SUBINDEX>.<EXT>`` and then in a second pass renumber all the images based on their relative order. This would be backwards compatible, but we'd still have the issue where adding or removing a previous image would upset the order, requiring renumbering all the proceeding images.

I don't think there's an ideal solution here, only tradeoffs. The sequential approach allows for the seamless change of a single image, whereas the hash approach is insensitive to relative position.

user27182: Yea it's a lot of additional work for little benefit. The speedup in build time is most critical, so that "pro" outweighs some of these other minor considerations.

As long as we primarily only show a single image for the examples in the `examples` module, we won't have any issues with needing to choose which image to show in the dataset gallery, or which image to use for the cells in the CellType docs.

akaszynski:
> Might need to change this line and also use a placeholder, similar to the dataset gallery?

Exactly. 8b2052a should fix just that.

user27182:
> > Might need to change this line and also use a placeholder, similar to the dataset gallery?
> 
> Exactly. [8b2052a](https://github.com/pyvista/pyvista/commit/8b2052afd8e59974853d8379587713571f854bc1) should fix just that.

It worked! https://689cd439bb11fc202481ffef--pyvista-dev.netlify.app/api/utilities/_autosummary/pyvista.celltype#pyvista.CellType.VERTEX

user27182 (approved):

banesullivan (approved):

PR #7804, opened 2025-08-10
Improve number_time_points property in EnsightReader

### LINKED ISSUE #7801: Make `time_values` and `number_time_points` safe for non time-series data

Hi! I propose an improvement for pyvista's Reader class around time handling.
If it makes sense to you, I am happy to submit PR.

## Current Behavior

When we call `time_values` or `number_time_points` method for non time-series data, they raise an error like below.
This behavior seems inconvenient when we write code that should work both for time-series and non time-series dataset.

```
  File "/home/riku/repos/personals/pyvista/workspace/issue1/sample.py", line 11, in <module>
    print(reader.time_values)
          ^^^^^^^^^^^^^^^^^^
  File "/home/riku/repos/personals/pyvista/pyvista/core/utilities/reader.py", line 684, in time_values
    return [self.time_point_value(idx) for idx in range(self.number_time_points)]
                                                        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/riku/repos/personals/pyvista/pyvista/core/utilities/reader.py", line 891, in number_time_points
    return self.reader.GetTimeSets().GetItem(self.active_time_set).GetSize()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'GetSize'
```

## Proposed behaviour

When the underlying dataset has no time:

* time_values returns an empty sequence (e.g., () or []).
* number_time_points returns 0 (e.g., len(time_values)).

Prepare sample dataset

* example.case

```text:example.case
FORMAT
type: ensight gold

GEOMETRY
model: example.geo
```

* example.geo

```text:example.geo
Ensight Gold

node id assign
element id assign

part
1
Simple Block

coordinates
4
0.0 1.0 2.0 3.0
0.0 0.0 0.0 0.0
0.0 0.0 0.0 0.0

hexa8
1
1 2 3 4 1 2 3 4
```

Python script

```python
import pyvista as pv
reader: pv.EnSightReader = pv.get_reader("example.case")
print(reader.number_time_points)
```

### System Information

```shell
  Date: Sat Aug 09 18:29:47 2025 JST

                OS : Linux (Pop!_OS 22.04)
            CPU(s) : 14
           Machine : x86_64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : Intel
       GPU Version : 4.6 (Core Profile) Mesa 24.2.8-1~bpo12+1pop1~1744225826~22.04~b077665
  MathText Support : True

  Python 3.11.9 (main, May 28 2024, 22:26:52) [GCC 11.4.0]

           pyvista : 0.47.dev0
               vtk : 9.5.0
             numpy : 2.3.1
        matplotlib : 3.10.3
            scooby : 0.10.1
             pooch : 1.8.2
            pillow : 11.3.0
```

### Screenshots

_No response_

riku-sakamoto:
### Overview

Return `0` when non-time series data is input in EnsightGold.

Resolves #7801

### Details

- Modified the method to return `0` instead of raising an error for non-time-series files.
- This makes it easier to write code that works for both time-series and non-time-series datasets.


pyvista/core/utilities/reader.py
```diff
@@ -888,7 +888,9 @@ def _set_filename(self, filename) -> None:
 
     @property
     def number_time_points(self):  # noqa: D102
-        return self.reader.GetTimeSets().GetItem(self.active_time_set).GetSize()
+        if (item := self.reader.GetTimeSets().GetItem(self.active_time_set)) is None:
```
MatthewFlamm: Minor style comment:  While this is valid code, the expectation for the walrus operator is that the assignment is used inside the block.  One way to address this would be to use an `else` statement, however using an `else` statement here will create a linting error about returns. Not sure how to 'fix' this other than assigning it before the if block.

pyvista/core/utilities/reader.py
```diff
@@ -888,7 +888,9 @@ def _set_filename(self, filename) -> None:
...
+            return 0
```
MatthewFlamm, line 893: This would be a breaking change as now anyone checking for `is None` will have code that breaks.  But I'm still in favor of this since it simplifies typing and expectation here.  If we merge this, we should implement for other readers to get a consistent behavior.

MatthewFlamm (commented):
Would like to get others opinion on changing this, but I'm overall in favor.

MatthewFlamm:
Coverage check is failing.  The `download_cylinder_crossflow` case has no time points and can be used for testing.  The tests for time points in ensight reader are here:  https://github.com/pyvista/pyvista/blob/4903251bdc96e90f81f75579213d768c1c0716bb/tests/core/test_reader.py#L265


Let me know if you need any help on adding tests.

pyvista/core/utilities/reader.py
riku-sakamoto: @MatthewFlamm 
Thank you for your feedback! 
As you suggested, I agree walrus operator is not appropriate here.
I rewrited it more simple way.

riku-sakamoto:
> Coverage check is failing. The `download_cylinder_crossflow` case has no time points and can be used for testing. The tests for time points in ensight reader are here:
> 
> https://github.com/pyvista/pyvista/blob/4903251bdc96e90f81f75579213d768c1c0716bb/tests/core/test_reader.py#L265
> 
> Let me know if you need any help on adding tests.

@MatthewFlamm 

Thank you for your kind help!  
I added a test case by generating a simple case in the unit test.  
Is this approach suitable for this repo, or should I add an example as a downloadable case instead?

akaszynski:
> Is this approach suitable for this repo, or should I add an example as a downloadable case instead?

I'd add it here:
https://github.com/pyvista/vtk-data

MatthewFlamm (approved):
Looks good to me.  Thanks for the great PR.

MatthewFlamm:
After this PR, we should refactor `TimeReader` to handle this logic for all readers.

riku-sakamoto:
> > Is this approach suitable for this repo, or should I add an example as a downloadable case instead?
> 
> I'd add it here: https://github.com/pyvista/vtk-data

@akaszynski 

Thank you for the instruction! I've updated the test to download the files.

PR #7809, opened 2025-08-11
Add remove_existing_actor parameter to add_mesh

### LINKED ISSUE #7808: Allow passing `remove_existing_actor` flag when calling `add_mesh`

Hello everyone,

I've encountered an issue while working on my PyVista application. I've noticed that the application opens slowly, and after profiling, I discovered something interesting.

When calling `add_mesh`, it subsequently calls `add_actor` and `remove_actor`. Since this is during the initialization of the view, I don't expect `remove_actor` to be invoked. Given the context, I know it's impossible for the actor to already exist, so this call seems unnecessary and adds overhead.

Upon reviewing the implementation of `add_mesh`, I found that providing a name determines whether `remove_actor` is called. This is concerning because it forces me to avoid naming my meshes to prevent the overhead. As a result, I lose some of PyVista's user-friendly features and have to manage the actors manually.

I would like to propose the ability to disable the `remove_actor` call when naming an actor created by `add_mesh`, using the existing `remove_existing_actor flag` available in `add_actor`.

_No response_

### Pseudocode or Screenshots

Image

tkoyama010:
## Summary
Implements the missing functionality for the `remove_existing_actor` parameter in the `add_mesh` method. The parameter was already added to the method signature but was not actually implemented in the underlying renderer logic.

Closes #7808


tests/plotting/test_plotting.py
```diff
@@ -3181,10 +3192,10 @@ def test_add_text_font_file():
     plotter = pv.Plotter()
     font_file = str(Path(__file__).parent / 'fonts/Mplus2-Regular.ttf')
     plotter.add_text(
-        '左上', position='upper_left', font_size=25, color='blue', font_file=font_file
+        '????', position='upper_left', font_size=25, color='blue', font_file=font_file
```
tkoyama010:
```suggestion
        '左上', position='upper_left', font_size=25, color='blue', font_file=font_file
```

tests/plotting/test_plotting.py
```diff
@@ -3181,10 +3192,10 @@ def test_add_text_font_file():
...
     )
     plotter.add_text(
-        '中央',
+        '????',
```
tkoyama010:
```suggestion
        '中央',
```

tkoyama010:
@pyvista-bot LGTM

PR #7815, opened 2025-08-12
Update GitHub Actions workflows to use uv

tkoyama010:
## Summary
Updates all GitHub Actions workflows to use uv instead of pip for faster dependency resolution, improved caching, and consistent dependency management.

## Changes Made
- **Replaced pip with uv**: All workflows now use `astral-sh/setup-uv@caf0cab7` (pinned for security)
- **Dependency management**: Migrated to `uv sync --group <group>` using dependency groups defined in pyproject.toml
- **Command execution**: Updated Python/pytest commands to use `uv run`
- **Build tools**: Changed to `uv tool install/run` for build utilities like `build` and `twine`
- **Dependency consolidation**: Moved conflicting dependency versions to centralized `pinned` group in pyproject.toml

## Benefits
- **Faster CI/CD**: uv is significantly faster than pip for dependency resolution
- **Better caching**: Built-in dependency caching improves subsequent runs
- **Consistency**: All workflows use the same dependency manager as defined in pyproject.toml
- **Simplified management**: Centralized dependency versioning through dependency groups


user27182:
+1 for using `uv`. This has a lot of overlap with #7441 . I wonder if we should consider that PR instead, unless there is a preference _not_ to also use `tox`.

akaszynski:
> +1 for using `uv`. This has a lot of overlap with #7441 . I wonder if we should consider that PR instead, unless there is a preference _not_ to also use `tox`.

Being able to run gh workflows locally would be great. Another +1 for `uv` and #7441

tkoyama010:
The purpose of this pull request is to address the fact that `uv` is currently unavailable. I apologize for not noticing #7441 earlier. Let’s add `uv` in this pull request, and then add `tox` in that one.

pyproject.toml
```diff
@@ -53,21 +53,22 @@ jupyter = [
 
 [dependency-groups]
 pinned = [ # Pinned versions of core dependencies
+  'colour-science<0.5.0',
   'matplotlib<3.10.2',
+  'mypy==1.17.1',
   'numpy<2.4.0',
   'pillow<11.4.0',
   'pooch<1.9.0',
+  'scipy<1.14.0',
```
tkoyama010:
```suggestion
  'scipy<1.17.0',
```

.github/workflows/docs.yml
```diff
@@ -99,7 +104,7 @@ jobs:
 
       - name: Build Documentation
         run: |
-          make -C doc \
+          uv run make \
```
tkoyama010:
```suggestion
          uv run make -C doc \
```

PR #7817, opened 2025-08-12
ci/use self hosted testing

akaszynski:
Use self hosted for testing.


akaszynski:
Sorry! CLI defaults to pyvista/pyvista when I mean pyvista/pyvista-sh

Issue #7811, opened 2025-08-11
HTTP error in CI when downloading from local cache


user27182:
### Describe the bug, what's wrong, and what you expected.

A job was added to CI in https://github.com/pyvista/pyvista/pull/7501 to cache data from the `vtk-data` repo to avoid flaky HTTP errors caused by all of the downloads. But we are still getting HTTP errors, e.g.:
https://github.com/pyvista/pyvista/actions/runs/16887760846/job/47840347309?pr=7807#step:8:88
``` shell
ERROR tests/plotting/test_plotter.py - requests.exceptions.HTTPError: 503 Server Error: Service Unavailable for url: https://github.com/pyvista/vtk-data/raw/master/Data/doorman/t_doorMan_teeth_n.png
```
But, in a previous step, we see that the cache was properly restored:
https://github.com/pyvista/pyvista/actions/runs/16887760846/job/47840347309?pr=7807#step:4:54
``` shell
Cache restored successfully
Cache restored from key: vtk-data-5bfe097b025ef52977af723a993f318ae30172bc
```

So if the data is cached locally and we aren't downloading anything from servers, how is it that we are still getting HTTP errors? Is CI still somehow downloading from GitHub despite having a local cache available?

### Steps to reproduce the bug.

See CI above

### System Information

```shell
See CI above
```

edabor:
After a quick code inspection, it appears that `PYVISTA_VTK_DATA` env variable must be set to enable local cache. In ci, only `PYVISTA_USERDATA_PATH` is set, which is the directory where downloaded files go to.
https://github.com/pyvista/pyvista/blob/main/pyvista%2Fexamples%2Fdownloads.py#L67

user27182:
> After a quick code inspection, it appears that `PYVISTA_VTK_DATA` env variable must be set to enable local cache. In ci, only `PYVISTA_USERDATA_PATH` is set, which is the directory where downloaded files go to. https://github.com/pyvista/pyvista/blob/main/pyvista%2Fexamples%2Fdownloads.py#L67

Seems like we just need to modify this line:
https://github.com/pyvista/pyvista/blob/37b66623dc8faabc1e2cc7b57e29f6286da1b005/.github/workflows/testing-and-deployment.yml#L35
``` diff
- PYVISTA_USERDATA_PATH: ${{ github.workspace }}/vtk-data 
+ PYVISTA_VTK_DATA: ${{ github.workspace }}/vtk-data 
```

[...]

PR #7820, opened 2025-08-13
add flaky test decorator to dataset_loader

akaszynski:
### Overview

Just had a failure here:
https://github.com/pyvista/pyvista/actions/runs/16941506418/attempts/1?pr=7813

Seems like we need to allow for retries for `test_dataset_loader_source_url_blob`.

Also, did we consider https://github.com/pytest-dev/pytest-rerunfailures when creating our own? Seems like we could enable a global option with `pytest-rerunfailures`, which might be nice, but it's not necessary.

### DIFF
--- a/tests/examples/test_downloads.py
+++ b/tests/examples/test_downloads.py
@@ -11,6 +11,7 @@ from retry_requests import retry
 import pyvista as pv
 from pyvista import examples
 from pyvista.examples import downloads
+from tests.conftest import flaky_test
 from tests.examples.test_dataset_loader import DatasetLoaderTestCase
 from tests.examples.test_dataset_loader import _generate_dataset_loader_test_cases_from_module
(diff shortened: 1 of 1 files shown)


user27182:
There are already retries enabled for this test:

https://github.com/pyvista/pyvista/blob/6466e1c0b4919b1b9361c0b3408034f250db9f6f/tests/examples/test_downloads.py#L41-L46

IMO we should fix #7811 instead, to ensure local cache is used and avoid HTTP requests altogether.

akaszynski:
Agreed. I was looking for a cheap solution (retry of a retry), but you're right that it's better to attack the root problem.

However, this checks if URLs are valid, not if we have the file available or not. This is an easy way to check if the URL is correct without downloading it. It's still an expensive test and there probably is a better way. Maybe fail when caching instead?

PR #7822, opened 2025-08-13
Add test coverage for custom VTK_DATA and USERDATA_PATH env vars

user27182:
An `is_dir()` check is also added for the `PYVISTA_VTK_DATA` env var so that if the var isn't valid, the default source is used as a fallback (similar to what's already done for `PYVISTA_USERDATA_PATH`)


pyvista/examples/downloads.py
```diff
@@ -60,49 +60,77 @@
...
+        msg = f'Unable to access path: {path}\n{msg}'
+        warnings.warn(msg)
+
+
+SOURCE, _FILE_CACHE = _get_vtk_data_source()
+USER_DATA_PATH = _get_user_data_path()
+
+_user_data_path_warn_msg = (
+    f'Manually specify the PyVista examples cache with the '
+    f'{_USERDATA_PATH_VARNAME} environment variable.'
```
akaszynski, line 131: Unlikely, but we might consider renaming this to prefix it with `PYVISTA` as it might collide with another env var from another library.

akaszynski: Nope, read that wrong, we're still prepending with PYVISTA. We're good.

akaszynski (approved):
I see no issues with this here. LGTM, thanks for adding coverage.

akaszynski:
Merging so we can test #7825.

user27182:
I think the change in filterwarnings from this PR needs adjustment, the filter is supposed to prevents errors like this in case the cache action fails:

https://github.com/pyvista/pyvista/actions/runs/16982976583/job/48175176674#step:8:1

PR #7825, opened 2025-08-14
Fix caching of `vtk-data` in CI unit tests

### LINKED ISSUE #7811: HTTP error in CI when downloading from local cache

user27182:
### Overview

Fix #7811 

Depends on #7822

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -220,7 +220,7 @@ filterwarnings = [
   "ignore:'mode' parameter is deprecated and will be removed in Pillow 13:DeprecationWarning", # matplotlib toolbar icon creation with newer Pillow
   'ignore:.*A NumPy version .* is required for this version of SciPy.*:UserWarning',
   'ignore:.*Given trait value dtype "float64":UserWarning', # bogus numpy ABI warning (see numpy/#432)
-  'ignore:.*PYVISTA_USERDATA_PATH is not a valid directory.*:UserWarning',
+  'ignore:.*PYVISTA_VTK_DATA is not a valid directory.*:UserWarning',
   'ignore:.*The NumPy module was reloaded*:UserWarning', # bogus numpy ABI warning (see numpy/#432)
   'ignore:.*numpy.dtype size changed.*:RuntimeWarning', # bogus numpy ABI warning (see numpy/#432)
   'ignore:.*numpy.ufunc size changed.*:RuntimeWarning', # bogus numpy ABI warning (see numpy/#432)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -220,10 +220,10 @@ filterwarnings = [
   "ignore:'mode' parameter is deprecated and will be removed in Pillow 13:DeprecationWarning", # matplotlib toolbar icon creation with newer Pillow
   'ignore:.*A NumPy version .* is required for this version of SciPy.*:UserWarning',
   'ignore:.*Given trait value dtype "float64":UserWarning', # bogus numpy ABI warning (see numpy/#432)
+  'ignore:.*PYVISTA_USERDATA_PATH is not a valid directory.*:UserWarning',
   'ignore:.*The NumPy module was reloaded*:UserWarning', # bogus numpy ABI warning (see numpy/#432)
   'ignore:.*numpy.dtype size changed.*:RuntimeWarning', # bogus numpy ABI warning (see numpy/#432)
   'ignore:.*numpy.ufunc size changed.*:RuntimeWarning', # bogus numpy ABI warning (see numpy/#432)
-  'ignore:Ignoring invalid PYVISTA_USERDATA_PATH.*:UserWarning',
   'ignore:Passing .{1}N.{1} to ListedColormap is deprecated since:DeprecationWarning', # https://github.com/matplotlib/cmocean/pull/114
   'ignore:The .*interactive_bk.* attribute was deprecated in Matplotlib 3\.9.*:matplotlib.MatplotlibDeprecationWarning', # https://github.com/microsoft/debugpy/issues/1623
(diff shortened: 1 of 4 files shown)


akaszynski (approved):
This will improve test stability. Thanks!

pyproject.toml
```diff
@@ -221,7 +221,7 @@ filterwarnings = [
   "ignore:'mode' parameter is deprecated and will be removed in Pillow 13:DeprecationWarning", # matplotlib toolbar icon creation with newer Pillow
   'ignore:.*A NumPy version .* is required for this version of SciPy.*:UserWarning',
   'ignore:.*Given trait value dtype "float64":UserWarning', # bogus numpy ABI warning (see numpy/#432)
-  'ignore:.*PYVISTA_USERDATA_PATH is not a valid directory.*:UserWarning',
+  'ignore:.*PYVISTA_VTK_DATA is not a valid directory.*:UserWarning',
```
user27182, line 224: Need to fix regex for multiline as this filter was broken by #7822, see https://github.com/pyvista/pyvista/pull/7822#issuecomment-3191965627
```suggestion
  'ignore:(?s).*PYVISTA_VTK_DATA is not a valid directory.*:\n.*:UserWarning',
```

tests/examples/test_dataset_loader.py
```diff
@@ -40,6 +40,12 @@ class DatasetLoaderTestCase:
     dataset_loader: tuple[str, _DatasetLoader]
 
 
+@pytest.fixture(autouse=True)
+def ignore_local_vtk_data_cache(monkeypatch):
```
edabor, line 44: Could you detail why this monkey patching was needed here ? It seems that htpp requests are emmited when doing so ?

user27182: It's because assertions like these fail otherwise

https://github.com/pyvista/pyvista/blob/main/tests/examples/test_dataset_loader.py#L214-L217

I initially tried modifying one or two tests to accommodate the local path but ran into some issues patching one of the tests, and there were multiple test failures so decided it was easier to monkeypatch the  module instead. Plus, the concept of raw vs. blob doesn't exist for local files, so figured it may be better to keep this kind of web-specific test coverage.

Yes, some http requests _may_ be emitted, but isn't it also possible for no http requests to be made since the download already exists locally? And even if requests are made, I have personally never seen these particular tests demonstrate flaky behavior.

user27182: I disabled wifi and ran all tests from this module locally with the `PYVISTA_VTK_DATA` env var set to point to a local cache of `vtk-data`. The tests all pass, even without an internet connection. So I think it's OK to monkeypatch this here.

akaszynski: > never seen these particular tests demonstrate flaky behavior

I've seen this at least once.

user27182: > > never seen these particular tests demonstrate flaky behavior
> 
> 
> 
> I've seen this at least once.


From these specific dataset loader module tests?

akaszynski: > From these specific dataset loader module tests?

Not quite, from here when attempting to query GitHub:
https://github.com/pyvista/pyvista/blob/cefe1fb05e9c379e7f4b74f9f77f6f4905445980/tests/examples/test_downloads.py#L46-L57

user27182: Yes indeed we have tests that are flaky from too many http requests... those mostly come from test_downloads and test_download_files. My point was specifically concerning the use of monkeypatch for this test module (test_dataset_loader), which doesn't really test the downloads and hence isn't really the cause of all the http requests, which is why I don't recall every seeing http tests failures from this specific module (and hence the monkeypatch isn't an issue).

edabor: > I disabled wifi and ran all tests from this module locally with the `PYVISTA_VTK_DATA` env var set to point to a local cache of `vtk-data`. The tests all pass, even without an internet connection. So I think it's OK to monkeypatch this here.

Thanks for testing. Indeed I think that the pooch fetcher is not affected by the monkey patch change when downloading files

edabor: > Yes indeed we have tests that are flaky from too many http requests... those mostly come from test_downloads and test_download_files. My point was specifically concerning the use of monkeypatch for this test module (test_dataset_loader), which doesn't really test the downloads and hence isn't really the cause of all the http requests, which is why I don't recall every seeing http tests failures from this specific module (and hence the monkeypatch isn't an issue). 

The local cache was specifically designed to reduce these flakiness

PR #7796, opened 2025-08-08
Drop support for Python 3.9 and VTK 9.0.3

tkoyama010:
## Summary
This PR reapplies the changes from #7693 which was initially merged but then reverted due to timing concerns for the 0.46 release. Now that we're on 0.47.dev0, we can safely apply these changes.

- Drop support for Python 3.9 (minimum now Python 3.10)
- Drop support for VTK 9.0.3 (minimum now VTK 9.1.0)
- Update CI configuration, documentation, and type annotations accordingly

## Changes
- **Python version**: Remove Python 3.9 from project classifiers and CI test matrix, update minimum to 3.10
- **VTK version**: Drop VTK 9.0.3 support and set minimum to 9.1.0
- **Type annotations**: Clean up compatibility code and update type hints for Python 3.10+
- **Documentation**: Update installation docs to reflect new minimum requirements
- **CI/CD**: Remove Python 3.9 from test matrix and update deployment to use Python 3.10

## References
- Original PR: #7693
- Revert PR: #7785

## Test plan
- [ ] CI tests pass with Python 3.10+ only
- [ ] Documentation builds correctly
- [ ] Type checking passes
- [ ] All examples run successfully

PR #7827, opened 2025-08-14
Bump `numpy` min version to 1.26 and keep `Union` for documented type aliases

user27182:
Address review comments from #7796


pyproject.toml
```diff
@@ -18,7 +18,7 @@ classifiers = [
 ]
 dependencies = [
   'matplotlib>=3.0.1',
-  'numpy>=1.21.0', # minimum typing support
```
user27182, line 21: We can maybe keep 1.21 , but we should use this version in the tests for vtk 9.2.2 (i.e. oldest vtk supported)

tkoyama010 (approved):
Thanks!

PR #7828, opened 2025-08-14
Use self-hosted MacOS

akaszynski:
### Self-Hosting Apple Silicon Runners

We're limited to a mere 5 MacOS runners, which means that our CI takes absolutely forever in the current config. We should still test on both self-hosted and GitHub-hosted MacOS, but at least half can be on self-hosted.

The approach to self-hosted runners will be to keep identical or near identical workflows and change `runs-on`. Conditional logic ~might be~ is needed, but we should keep it to a minimum.

Luckily, the only conditional step needed is for setting up Python. This approach lets us avoid the usage of `sudo` in this workflow.

For anyone who's worried about self-hosted runners being hard to maintain, we can always revert to using GitHub hosted runners by changing the labels back to GitHub's labels. Nothing else in the workflow needs to be changed.

There were, however, a few changes needed to tests and the library, two of which have been pulled out into separate PRs.
- [x] https://github.com/pyvista/pyvista/pull/7830
- [x] https://github.com/pyvista/pyvista/pull/7831

---

### Speedup and Stability

Overall, the CI for the new workflows takes less than half the time (16 minutes on this branch vs. 38 minutes on `main`). Tests have been stable for the past three runs with no failures or restarts.

---
### Testing Changes in this PR

- `FAILED tests/core/test_utilities.py::test_apply_transformation_to_points - RuntimeWarning: divide by zero encountered in matmul`: Edge case discovered when using M4 hardware. See https://github.com/numpy/numpy/issues/28687, and straightforward to fix (already done so in `numpy==2.3.0`).
- Issues on parallel docbuilds on MacOS Metal. Disabling for now.

---

### Immediately before Merging

Change branch protections to match `MacOS Unit Testing (Python 3.9)` through `MacOS Unit Testing (Python 3.13)`. This will match for both self-hosted and GitHub-hosted runners (see remarks in this branch's [CONTRIBUTING.rst](https://github.com/pyvista/pyvista/blob/ci/macos-self-hosted/CONTRIBUTING.rst).


tests/examples/test_downloads.py
```diff
@@ -150,6 +150,7 @@ def test_local_file_cache(tmp_path: Path):
     try:
         downloads.FETCHER.base_url = dirname + '/'
         downloads.FETCHER.registry[basename] = None
+        downloads.FETCHER.path = tmp_path
```
akaszynski:
```suggestion
```

.github/workflows/testing-and-deployment.yml
```diff
@@ -74,31 +74,50 @@ jobs:
...
-          - { os: macos-latest, python-version: "3.11" }
-          - { os: macos-latest, python-version: "3.12" }
-          - { os: macos-latest, python-version: "3.13" }
-
-    runs-on: ${{ matrix.config.os }}
-    needs: cache-vtk-data
-
+        include:
+          # GitHub-hosted runner configuration
+          - name: MacOS Unit Testing (Python 3.9)
```
user27182: Looks like the new names don't match the old names, so the required tests probably need to be updated (in admin repo settings??)
image

Nit: All other CI jobs also _exclude_ "Python" from the name
```suggestion
          - name: MacOS Unit Testing (3.9)
```

akaszynski: I prefer it with Python as it makes it clear what 3.9 is referring to but you're right, it's inconsistent with the other CI.

akaszynski: Resolved in [901225d](https://github.com/pyvista/pyvista/pull/7828/commits/901225d104d4e552ad51c299f12795d8a8a95249).

akaszynski: > Looks like the new names don't match the old names, so the required tests probably need to be updated (in admin repo settings??)

Correct, right before we merge I'll modify it so we don't block other workflows.

user27182: The other issue with including "Python" explicitly is that it can make the UI less useful since the name may be truncated, e.g. see https://github.com/pyvista/pyvista/pull/7441#discussion_r2143826025
As a middle ground, in #7441 "py3.#" is used instead of "Python 3.#" 
ci

user27182:
This PR made a huge difference. Jobs used to take 1hr and were always queued for a long time. Now they're all done within 15 mins

akaszynski:
> This PR made a huge difference. Jobs used to take 1hr and were always queued for a long time. Now they're all done within 15 mins

Next step, the Linux runners! We get 20 concurrent runners and just took 5 off the hook with our own self-hosted macOS runners.

akaszynski:
Tracking post merge failures here:

```
FAILED tests/examples/test_dataset_loader.py::test_single_file_loader[_SingleFileDatasetLoader-False] - AssertionError: assert False
 +  where False = <function isfile at 0x10077b740>('/Users/runner/Library/Caches/pyvista_3/pyvista_logo.png')
 +    where <function isfile at 0x10077b740> = <module 'posixpath' (frozen)>.isfile
 +      where <module 'posixpath' (frozen)> = os.path
 +    and   '/Users/runner/Library/Caches/pyvista_3/pyvista_logo.png' = <pyvista.examples._dataset_loader._SingleFileDatasetLoader object at 0x12e628410>.path
= 1 failed, 8638 passed, 121 skipped, 9 xfailed, 1 warning in 292.11s (0:04:52) =
```

user27182:
> Tracking post merge failures here:
> 
> ```
> FAILED tests/examples/test_dataset_loader.py::test_single_file_loader[_SingleFileDatasetLoader-False] - AssertionError: assert False
>  +  where False = <function isfile at 0x10077b740>('/Users/runner/Library/Caches/pyvista_3/pyvista_logo.png')
>  +    where <function isfile at 0x10077b740> = <module 'posixpath' (frozen)>.isfile
>  +      where <module 'posixpath' (frozen)> = os.path
>  +    and   '/Users/runner/Library/Caches/pyvista_3/pyvista_logo.png' = <pyvista.examples._dataset_loader._SingleFileDatasetLoader object at 0x12e628410>.path
> = 1 failed, 8638 passed, 121 skipped, 9 xfailed, 1 warning in 292.11s (0:04:52) =
> ```

I think is the same error reported here https://github.com/pyvista/pyvista/pull/7834#discussion_r2286825770
This error started showing up after that PR merged. Seems to only happen when there is a vtk-data cache failure

PR #7831, opened 2025-08-16
Avoid download fetcher test side-effects

akaszynski:
Edge case, but it's possible for multiple instances of python to be accessing the same cache and deleting the cached file simultaneously (unlikely, but it happened in https://github.com/pyvista/pyvista/pull/7828). It's also bad practice to be modifying files outside temporary directories anyway.


tkoyama010 (approved):
LGTM

PR #7833, opened 2025-08-17
Improve morphological filters for ImageData

### LINKED ISSUE #7660: Improve morphological filters for `ImageData`

We should deprecate [image_dilate_erode](https://dev.pyvista.org/api/core/_autosummary/pyvista.imagedatafilters.image_dilate_erode#pyvista.ImageDataFilters.image_dilate_erode) and replace it with dedicated `dilate` and `erode` filters, and extend these to also support:

- dilation of continuous data with https://vtk.org/doc/nightly/html/classvtkImageContinuousDilate3D.html#details
- erosion of continuous data with https://vtk.org/doc/nightly/html/classvtkImageContinuousErode3D.html

And include dedicated `open` and `close` filters too with https://vtk.org/doc/nightly/html/classvtkImageOpenClose3D.html

_No response_

### Pseudocode or Screenshots

_No response_

tkoyama010:
Resolves #7660

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
       - id: blackdoc
         args: ["--skip-string-normalization"]
         types_or: [rst]
+        additional_dependencies: ["black==25.1.0"]
 
   - repo: https://github.com/codespell-project/codespell
     rev: v2.4.1
@@ -47,7 +48,7 @@ repos:
         args: [--branch, main]
       - id: requirements-txt-fixer
       - id: trailing-whitespace
-        exclude: ^(pyvista/examples/.*\.ply|pyvista/examples/.*\.vtk|doc/intersphinx/.*\.inv)
+        exclude: ^(pyvista/examples/.*\.ply|pyvista/examples/.*\.vtk|doc/intersphinx/.*\.inv|tests/test_cli\.py)
       - id: check-docstring-first
       - id: end-of-file-fixer
         exclude: ^(pyvista/examples/.*\.ply|pyvista/examples/.*\.vtk|doc/intersphinx/.*\.inv)
--- a/doc/source/api/core/typing.rst
+++ b/doc/source/api/core/typing.rst
@@ -82,6 +82,20 @@ pyvista.InteractionEventType
 
 .. autodata:: InteractionEventType
 
+pyvista.CameraPositionOptions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. currentmodule:: pyvista
+
+.. autodata:: CameraPositionOptions
+
+pyvista.JupyterBackendOptions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. currentmodule:: pyvista
+
+.. autodata:: JupyterBackendOptions
+
 
 Data Object Types
 -----------------
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -122,6 +122,8 @@ extensions = [
 # Configuration for sphinx.ext.autodoc
 # Do not expand following type aliases when generating the docs
 autodoc_type_aliases = {
+    'CameraPositionOptions': 'pyvista.CameraPositionOptions',
+    'JupyterBackendOptions': 'pyvista.JupyterBackendOptions',
     'Chart': 'pyvista.Chart',
     'ColorLike': 'pyvista.ColorLike',
     'ArrayLike': 'pyvista.ArrayLike',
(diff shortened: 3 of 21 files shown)


pyvista/core/filters/data_set.py
```diff
@@ -7845,7 +7845,7 @@ def _preprocess_polydata(poly_in):
             if background_value == 0
             else np.ones(scalars_shape, dtype=scalars_dtype) * background_value
         )
-        binary_mask['mask'] = scalars  # type: ignore[assignment]
+        binary_mask['mask'] = scalars.astype(float)
```
tkoyama010:
```suggestion
        binary_mask['mask'] = scalars  # type: ignore[assignment]
```

pyvista/demos/logo.py
```diff
@@ -121,7 +121,9 @@ def logo_letters(merge=False, depth=0.3):  # noqa: FBT002
         the keys are the letters and the values are the respective meshes.
 
     """
-    mesh_letters = pyvista.PolyData() if merge else {}  # type: ignore[var-annotated]
+    mesh_letters: pyvista.PolyData | dict[str, pyvista.PolyData] = (
+        pyvista.PolyData() if merge else {}
+    )
```
tkoyama010:
```suggestion
    mesh_letters = pyvista.PolyData() if merge else {}  # type: ignore[var-annotated]
```

user27182 (commented):
Haven't reviewed closely yet but it's not clear to me if `continous` should even be needed as a keyword, as I'm not exactly sure what the difference in output would be for some common use cases, it may be be that the continuous filters do everything the binary ones do, plus more, i.e. perhaps they are more general.

The use of additional examples would help clear this up. 
In addition, the current examples have a binary image input and binary output, but the plots use interpolation to color the cells. Instead, visualizing the point data as cell data using something `points_to_cells` before plotting would help make the plot actually appear binary.

akaszynski:
> The use of additional examples would help clear this up.

Agreed, would be nice to have some additional examples for this.

user27182:
> Haven't reviewed closely yet but it's not clear to me if `continous` should even be needed as a keyword, as I'm not exactly sure what the difference in output would be for some common use cases, it may be be that the continuous filters do everything the binary ones do, plus more, i.e. perhaps they are more general.

Scipy distinguishes these cases as, for example, `binary_erosion` and `grey_erosion` 
https://docs.scipy.org/doc/scipy/tutorial/ndimage.html#binary-morphology
https://docs.scipy.org/doc/scipy/tutorial/ndimage.html#grey-scale-morphology
We don't have to split it up like this or use their naming/terminology but it's a helpful comparison

user27182:
> > Haven't reviewed closely yet but it's not clear to me if `continous` should even be needed as a keyword, as I'm not exactly sure what the difference in output would be for some common use cases, it may be be that the continuous filters do everything the binary ones do, plus more, i.e. perhaps they are more general.
> 
> 
> 
> Scipy distinguishes these cases as, for example, `binary_erosion` and `grey_erosion` 
> 
> https://docs.scipy.org/doc/scipy/tutorial/ndimage.html#binary-morphology
> 
> https://docs.scipy.org/doc/scipy/tutorial/ndimage.html#grey-scale-morphology
> 
> We don't have to split it up like this or use their naming/terminology but it's a helpful comparison

There are also similar filters in sklearn: https://scikit-image.org/docs/stable/api/skimage.morphology.html

There, the binary versions are reported as generating identical output as the grey scale versions but are simply faster.

For vtk, using the binary filters and setting dilate and erode values explicitly  is useful for restricting the dilation/erosion to specific values for multi-label inputs. E.g. a segmentation mask with multiple labels where you only want to operate on a single label and leave the others unaffected. This is similar to the scipy filters where they allow specifying an input mask to restrict the operation to the masked region.

And so, it's useful to keep both variations. But I think we should default to using the continuous filters, and make using the binary versions optional (i.e. for advanced use).

user27182:
@tkoyama010 https://github.com/pyvista/pyvista/pull/7996 is feature-complete and effectively ready for review.

We could:

1. Merge this PR (mostly) as-is with main, then merge [#7996](https://github.com/pyvista/pyvista/pull/7996) separately with main
2. Rebase [#7996](https://github.com/pyvista/pyvista/pull/7996) and merge it into this branch, then merge this PR with main
3. Close this PR and merge [#7996](https://github.com/pyvista/pyvista/pull/7996) directly with main

Option 1 is nice because I think both this PR and [#7996](https://github.com/pyvista/pyvista/pull/7996) required substantial effort, and the commit history will record both @tkoyama010 and myself as primary authors separately. This will create a bit of churn though, as the filters from this PR have effectively been completely rewritten in  [#7996](https://github.com/pyvista/pyvista/pull/7996), and only a few tests from this PR remain untouched by that PR.

tkoyama010:
Thanks! Let's do option 1.

PR #7834, opened 2025-08-18
Fix cache vtk-data

edabor:
### Overview
Fix https://github.com/pyvista/pyvista/issues/7811

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ typing = [
 ]
 
 test = [
+  'PyObjc; sys_platform == "darwin"', # see pyvista/pulls/7828
   'cmcrameri<1.10.0',
   'cmocean<4.0.4',
   'colorcet<3.2.0',
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -11,7 +11,6 @@ import contextlib
 from contextlib import contextmanager
 from contextlib import suppress
 from copy import deepcopy
-import ctypes
 from functools import wraps
 import io
 from itertools import cycle
@@ -138,16 +137,11 @@ if TYPE_CHECKING:
 
 SUPPORTED_FORMATS = ['.png', '.jpeg', '.jpg', '.bmp', '.tif', '.tiff']
 
-# EXPERIMENTAL: permit pyvista to kill the render window
-KILL_DISPLAY = platform.system() == 'Linux' and os.environ.get('PYVISTA_KILL_DISPLAY')
-if KILL_DISPLAY:  # pragma: no cover
-    # this won't work under wayland
-    try:
-        X11 = ctypes.CDLL('libX11.so')
-        X11.XCloseDisplay.argtypes = [ctypes.c_void_p]
-    except OSError:
-        warnings.warn('PYVISTA_KILL_DISPLAY: Unable to load X11.\nProbably using wayland')
-        KILL_DISPLAY = False
+if os.environ.get('PYVISTA_KILL_DISPLAY'):  # pragma: no cover
+    from pyvista.core.errors import DeprecationError
+
+    msg = 'PYVISTA_KILL_DISPLAY has been deprecated'
+    DeprecationError(msg)
 
 
 def close_all() -> bool:
@@ -5037,7 +5031,15 @@ class BasePlotter(_BoundsSizeMixin, PickingHelper, WidgetHelper):
         """Clear the render window."""
         # Not using `render_window` property here to enforce clean up
         if hasattr(self, 'ren_win'):
-            self.ren_win.Finalize()
+            apple_silicon = platform.system() == 'Darwin' and platform.machine() == 'arm64'
+            if not apple_silicon:  # pragma: no cover
+                # Up to vtk==9.5.0, render windows aren't closed on MacOS,
+                # so the resources are not freed making this unnecessary. Also,
+                # we need this disabled so we can use NSAutoreleasePool in unit
+                # testing.
+                # see https://gitlab.kitware.com/vtk/vtk/-/issues/18713
+                self.ren_win.Finalize()
+
             del self.ren_win
 
     def close(self) -> None:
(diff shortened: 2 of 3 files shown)


tests/examples/test_dataset_loader.py
```diff
@@ -164,25 +164,16 @@ def examples_local_repository_tmp_dir(tmp_path):
...
-    # make sure any "downloaded" files (moved from repo -> cache) are cleared
-    cached_paths = [os.path.join(downloads.FETCHER.path, base) for base in downloadable_basenames]
-    [os.remove(file) for file in cached_paths if os.path.isfile(file)]
+    monkeypatch.setattr(downloads, 'FETCHER', FETCHER)
+    monkeypatch.setattr(downloads, '_FILE_CACHE', True)
 
-    yield repository_path
-
-    # teardown
-    downloads.FETCHER.base_url = 'https://github.com/pyvista/vtk-data/raw/master/Data/'
```
edabor, line 180: Original issue was due to this side-effect. The fetcher was reset to a different state compared to before executing the test.

This PR fixes it by using monkeypatching which handles teardown and ensures tests isolation.

pyproject.toml
```diff
@@ -229,6 +228,7 @@ filterwarnings = [
   'ignore:The .*interactive_bk.* attribute was deprecated in Matplotlib 3\.9.*:matplotlib.MatplotlibDeprecationWarning', # https://github.com/microsoft/debugpy/issues/1623
   'ignore:pyvista test \w+ image dir.* does not yet exist.  Creating dir.:UserWarning',
 
+  'always:(?s).*PYVISTA_VTK_DATA is not a valid directory.*:UserWarning',
```
edabor, line 232: Modified since it can be useful for CI logs inspection to see where is the cache source.

edabor:
@user27182 Finally we should be done with https requests flakiness once for all 😆

tests/examples/test_download_files.py
```diff
@@ -10,22 +10,55 @@
...
+
+
+@pytest.fixture(scope='module', autouse=True)
+def check_cache_on_ci():
+    if not _on_ci():
+        return
+
+    assert examples.downloads._FILE_CACHE, (
+        f'Expected `_FILE_CACHE` to be True on CI. Source is set to {examples.downloads.SOURCE}'
+    )
```
user27182: If there is a cache miss, the PYVISTA_VTK_DATA env var will point to an invalid dir, and _FILE_CACHE will be False. This assertion will cause the tests to fail in this case, but we want to allow downloads to continue as normal and download from the vtk-data repo

tests/examples/test_download_files.py
```diff
@@ -10,22 +10,55 @@
...
+    """Mock the requests.get method to make sure HTTP requests are not emitted on CI,
+    since can cause flakiness dut to GH rate limits.
+    """
+    if not _on_ci():
+        yield
+        return
+
+    spy = mocker.spy(requests, 'get')
+    yield
+    assert spy.call_count == 0, spy.mock_calls
```
user27182, line 73: Similar to _FILE_CACHE, we want to allow http requests in case the cache action fails and there is a cache miss

edabor:
> I think these changes will cause CI to fail if there is a cache miss. Should we keep track of this, and allow http requests in this case? E.g. allow http requests `if PYVISTA_VTK_DATA in os.environ and not Path(os.environ[PYVISTA_VTK_DATA]).is_dir()`

Agreed. I implemented your suggestion in https://github.com/pyvista/pyvista/pull/7834/commits/89bcfb09fa245e142e8aea3e5bbec080f17960ec and 591c3613f4b6388bc387d927d312620e351b0a37

Let's see how it behaves when the cache is missing, as intentionally done in https://github.com/pyvista/pyvista/pull/7834/commits/905b1d651618d1b0a009ecc97768fa1fecb6d458

edabor:
After a cache miss, we correctly warns that the cache directory is missing but downloads tests are correctly executed using http requests. https://github.com/pyvista/pyvista/actions/runs/17052244098/job/48342808915?pr=7834#step:12:8811

Reverting https://github.com/pyvista/pyvista/pull/7834/commits/905b1d651618d1b0a009ecc97768fa1fecb6d458

akaszynski (approved):
> After a cache miss, we correctly warns that the cache directory is missing but downloads tests are correctly executed using http requests. https://github.com/pyvista/pyvista/actions/runs/17052244098/job/48342808915?pr=7834#step:12:8811

Calling this good. Thanks for your work!

pyproject.toml
user27182: It is possible that using always instead of ignore is causing this assertion failure, since the message is now showing in the output?

https://github.com/pyvista/pyvista/actions/runs/17081545177/job/48437122529#step:15:1

user27182: It seems the issues we have with this warning may be isolated to tinypages tests. Maybe we can fully ignore the warning for this module only instead of using always?

akaszynski: Alternatively, we could ignore the warning within the tinypages tests. Easy to do within the `conf.py`.

user27182: I guess it's not isolated to `tinypages`, here's another test from the `test_dataset_loader` module with a cache miss and an error raised from this warning
https://github.com/pyvista/pyvista/actions/runs/17086020256/job/48450172182#step:8:8925

edabor: > I guess it's not isolated to `tinypages`, here's another test from the `test_dataset_loader` module with a cache miss and an error raised from this warning https://github.com/pyvista/pyvista/actions/runs/17086020256/job/48450172182#step:8:8925

The failure does not seem to be caused by the warning but is rather a true failure. My guess is that it comes from the test setup `examples_local_repository_tmp_dir` that removes files at teardown.

edabor: I guess we did not see the tinypages tests failure when intentionally creating a cache miss in https://github.com/pyvista/pyvista/commit/905b1d651618d1b0a009ecc97768fa1fecb6d458 since plotting tests were interrupted https://github.com/pyvista/pyvista/actions/runs/17052244098/job/48342808915?pr=7834#step:15:1333

And when testing in local with Windows, the tests are skipped so could not see it ... 
https://github.com/pyvista/pyvista/blob/cfeff3769318953d6eab21cc4f998f728c5800d9/tests/plotting/test_tinypages.py#L26

I'll fix with another PR. Sorry about that

PR #7840, opened 2025-08-19
Fix `BoundsTuple.__repr__` with scientific notation

user27182:
### Overview

The current repr for `BoundsTuple` assumes there is a decimal in the stringified float(s), but this is not always the case. E.g. this example shows a mesh where the bounds use scientific notation:
``` py
import pyvista as pv
grid = pv.UnstructuredGrid().extract_cells(0)
repr(grid)
# UnstructuredGrid (0x1112236a0)
#   N Cells:    0
#   N Points:   0
#   X Bounds:   1.000e+299, -1.000e+299
#   Y Bounds:   1.000e+299, -1.000e+299
#   Z Bounds:   1.000e+299, -1.000e+299
#   N Arrays:   0
```

but then trying to show the bounds directly results in an error
``` py
repr(grid.bounds)
```
``` py
  File "/pyvista/core/_typing_core/_aliases.py", line 98, in __repr__
    left, right = parts
    ^^^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 1)
```

This PR fixes this.

### DIFF
--- a/pyvista/core/_typing_core/_aliases.py
+++ b/pyvista/core/_typing_core/_aliases.py
@@ -84,7 +84,9 @@ class BoundsTuple(NamedTuple):
     def __repr__(self) -> str:
         # Split bounds at decimal and compute padding needed to the left of it
         dot = '.'
-        split_strings = [str(float(val)).split(dot) for val in self]
+        strings = [str(float(val)) for val in self]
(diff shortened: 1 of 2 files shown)


tkoyama010 (approved):
LGTM

PR #7845, opened 2025-08-19
Improve test coverage and support for `PointSet`

user27182:
Swept through every test/filter touched by adding a `PointSet` to the shared `datasets` fixture, fixing real bugs along the way, not just test expectations.

- The `datasets` test fixture now includes a `PointSet` alongside the other dataset types (`ImageData`, `RectilinearGrid`, `UnstructuredGrid`, `PolyData`, `StructuredGrid`); the old cell-based-only set is preserved separately as `datasets_no_pointset` for filters that genuinely need cells
- `glyph()` no longer collapses its output to a cell-less `PointSet` when called on `PointSet` input — glyphs always produce real cell geometry (e.g. arrows), so the output stays `PolyData`
- `PointSet` now explicitly overrides `extract_all_edges`, `compute_cell_sizes`, and `cell_quality` to raise `PointSetCellOperationError`, matching every other unsupported cell-based filter
- `cell_quality`'s composite (`MultiBlock`) dispatch now routes through the block instance rather than a hardcoded base-class function, so a block-level override (like `PointSet`'s) is actually honored
- Fixed a real segfault: `cell_data_to_point_data`, `point_data_to_cell_data`, `extract_all_edges`, `compute_cell_sizes`, and `elevation` all handed a `MultiBlock` straight to a raw VTK algorithm; on VTK 9.3.x this crashes the interpreter if any block is a cell-less `PointSet`. Each now raises a clean Python error first instead (version-gated for `elevation`, which is otherwise valid on `PointSet` and works fine on VTK ≥ 9.4)
- `clip()` and `clip_scalar()` on a `PointSet` correctly return another `PointSet`, and both are `xfail`ed on VTK 9.4.x for a separate known VTK bug where the clip result comes back empty
- Fixed two `NameError` bugs (`pyvista.X` instead of `pv.X`) in filter code that were only reachable when a `PointSet` participated in a composite operation


banesullivan:
I'd like to revive this PR if we can since it discovered gaps in the PointSet's filters API that needed error handling

user27182:
> I'd like to revive this PR if we can since it discovered gaps in the PointSet's filters API that needed error handling

Yes the issue I had with this PR is that enabling the coverage found many errors. And since this coverage was added by changing fixtures that are used everywhere in the tests, _all_ uses of the fixtures needed to be fixed. And so the scope of this PR exploded, and I didn’t have the bandwidth to fix it all.

banesullivan:
Thanks for highlighting this and getting it started! I may try finish it out

user27182:
@banesullivan this is ready to go if you'd like to have a look

akaszynski (approved):

PR #7849, opened 2025-08-20
Raise error when comparing `vtk_version_info` with an unsupported VTK version

user27182:
### Overview

Raise errors to ensure that all vtk version checks in the code are relevant. If checking against old versions, an error is raised so that the check can be removed whenever there is a bump to the min supported vtk version.

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -1402,13 +1402,20 @@ class DataObjectFilters:
             crinkle=crinkle,
         )
 
+        # Post-process clip to fix output type and remove unused points
+        input_bounds = self.bounds
         if isinstance(result, tuple):
             result = (
                 _cast_output_to_match_input_type(result[0], self),
                 _cast_output_to_match_input_type(result[1], self),
             )
+            result = (
+                _remove_unused_points_post_clip(result[0], input_bounds),
+                _remove_unused_points_post_clip(result[1], input_bounds),
+            )
         else:
             result = _cast_output_to_match_input_type(result, self)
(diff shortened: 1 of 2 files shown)


tkoyama010 (dismissed):
LGTM

pyvista/core/_vtk_core.py
```diff
@@ -607,13 +611,45 @@ def VTKVersionInfo():  # noqa: N802
...
+        self._check_min_supported(other)
+        return super().__gt__(other)
+
+    def __ge__(self, other):
+        self._check_min_supported(other)
+        return super().__ge__(other)
+
+
+vtk_version_info = VTKVersionInfo(*_get_vtk_version())
+_MIN_SUPPORTED_VTK_VERSION = (9, 2, 0)
```
edabor: Maybe we could add a test that verifies that the minimum version is compatible with the one in pyproject.toml ?

user27182: This is a good idea. They're already out of sync: must be >=9.2.0 in code, but must be >=9.2.2 in pyproject

tkoyama010 (dismissed):
Thanks!

user27182:
I think the bump may cause new test failures which will need to be fixed, e.g. a quick search for "(9, 2, 2)" found:

https://github.com/pyvista/pyvista/blob/a9e1d5ca11884482eb4b0bd2a34831dd4fbd65fc/pyvista/report.py#L61-L66

EDIT: This might be OK, since an error is only raised if checking strictly less than min supported. Will keep as draft until we have green CI

edabor:
@tkoyama010 LGTM. Thanks for the ping

PR #7853, opened 2025-08-20
Auto-cancel Windows jobs after 60 minutes

user27182:
### Overview

Windows jobs occasionally hang for reasons unknown, and the job will be stuck in limbo until it auto-cancels after a default of 6 hours. Instead, this PR makes it timeout after 1 hr (they usually finish in under 30 minutes).

Might want to consider auto-retry with https://github.com/nick-fields/retry in the future, or somehow fix the hang-up itself if we can figure out the cause.

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -369,6 +369,7 @@ jobs:
   windows:
     name: Windows Unit Testing
     runs-on: windows-latest
+    timeout-minutes: 60
     needs: cache-vtk-data
     env:
       CI_WINDOWS: true


akaszynski (approved):

user27182:
Here's one example, got stuck at 98%
``` py
tests/plotting/test_widgets.py::test_widget_radio_button_multiple_switch PASSED [ 98%]
```
https://github.com/pyvista/pyvista/actions/runs/17133604410/job/48604157210?pr=7855#step:9:1763

akaszynski:
> Here's one example, got stuck at 98%
> 
> ```python
> tests/plotting/test_widgets.py::test_widget_radio_button_multiple_switch PASSED [ 98%]
> ```
> 
> https://github.com/pyvista/pyvista/actions/runs/17133604410/job/48604157210?pr=7855#step:9:1763

Don't tempt me into trying to self host windows runners too...

PR #7848, opened 2025-08-20
Use self-hosted runner for docs

akaszynski:
Use our self-hosted Linux runner for the docs with 8 cores.

Resolves #5118, depth peeling works.

No cache runtime is 34m vs 1h 18m on `main`. This can be improved as well by making our table builder multi-threaded and using local rather than remote caching using https://github.com/falcondev-oss/github-actions-cache-server.

Cached runtime is 23m vs 58m on `main`.

Issue #5118, opened 2023-10-24
Depth peeling with OSMesa might be broken on v0.42


banesullivan:
### Describe the bug, what's wrong, and what you expected.

The depth peeling example in the docs for v0.42 is not working: https://docs.pyvista.org/version/stable/examples/02-plot/depth-peeling.html

![sphx_glr_depth-peeling_001](https://github.com/pyvista/pyvista/assets/22067021/5ea5c678-f88c-4fb7-bf99-7206b1822138)

It was working for v0.41: https://docs.pyvista.org/version/0.41/examples/02-plot/depth-peeling.html

I'm not sure if this issue is because of PyVista or a difference in the VTK or OSmesa versions across those documentation builds. Figured I would track here regardless

### Steps to reproduce the bug.

```python
Insert code here
```

### System Information

```shell
N/A
```

PR #7861, opened 2025-08-21
Use vtkEGLRenderWindow for docs

akaszynski:
Initial results from with https://github.com/pyvista/pyvista/pull/7848 show that using `vtkEGLRenderWindow` results in correctly rendered depth peeling, resolving #5118.

Instead of using two separate backends (one for GPU self-hosted, one for CPU GitHub-hosted), this PR switches our default backend to EGL on the documentation build workflow.

`vtkOSOpenGLRenderWindow`
sphx_glr_depth_peeling_001

`vtkEGLRenderWindow`
sphx_glr_depth_peeling_001

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -36,7 +36,7 @@ env:
   PYDEVD_DISABLE_FILE_VALIDATION: "1"
   PYTEST_ADDOPTS: "--color=yes"
   FORCE_COLOR: "True"
-  VTK_DEFAULT_OPENGL_WINDOW: "vtkOSOpenGLRenderWindow"
+  VTK_DEFAULT_OPENGL_WINDOW: "vtkEGLRenderWindow"
 
 permissions:
   id-token: none


akaszynski:
There's still an issue with depth peeling:

image

```
      GPU Renderer : llvmpipe (LLVM 15.0.7, 256 bits)
       GPU Version : 4.5 (Compatibility Profile) Mesa 23.2.1-1ubuntu3.1~22.04.3
     Render Window : vtkEGLRenderWindow
  MathText Support : True
 ```
 
 Seems the real issue is using `llvmpipe` with Mesa. GPU rendering doesn't have these issues.

That being said, I still think we should merge this just to consistently use `vtkEGLRenderWindow`.

user27182:
> That being said, I still think we should merge this just to consistently use `vtkEGLRenderWindow`.

Ideally, we could configure a test variable to define this in exactly one spot and reuse the variable in both the doctest and doc build jobs. I _think_ using `tox` would enable this?

akaszynski:
> Ideally, we could configure a test variable to define this in exactly one spot and reuse the variable in both the doctest and doc build jobs. I _think_ using `tox` would enable this?

Agreed. There's still quite a bit of refactoring to do though, because now using EGL will mean different results depending on GPU or non-gpu tests. We'll need to test at runtime for the GPU.

PR #7766, opened 2025-07-31
Catch vtk errors in tests

user27182:
### Overview

Partially resolve #7749 (does not apply to Sphinx Gallery)

PR #7863, opened 2025-08-21
Enable `VtkErrorCatcher` to catch VTK warnings

user27182:
### Overview

Part of a broader effort to catch errors and warnings during testing #7766 

This PR:
- catches warnings in a similar manner to how errors were already being caught
- adds dedicated vtk warning and error classes
- fixes some of the error message parsing

### DIFF
--- a/FUNDING.yml
+++ b/FUNDING.yml
@@ -1,0 +1,1 @@
+github: [banesullivan, tkoyama010, numfocus]
--- a/README.rst
+++ b/README.rst
@@ -300,17 +300,12 @@ BibTex:
 Professional Support
 ====================
 
-While PyVista is an Open Source project with a big community, you might be looking for professional support.
-This section aims to list companies with VTK/PyVista expertise who can help you with your software project.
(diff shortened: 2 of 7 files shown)


akaszynski:
Sorry, killed the runner process while testing on MacOS, that's my fault.

pyvista/core/utilities/observers.py
```diff
@@ -10,12 +10,19 @@
...
 import threading
 import traceback
+from typing import TYPE_CHECKING
 from typing import NamedTuple
 
 from pyvista._deprecate_positional_args import _deprecate_positional_args
+from pyvista._warn_external import warn_external
 from pyvista.core import _vtk_core as _vtk
+from pyvista.core.errors import VTKOutputMessageError
+from pyvista.core.errors import VTKOutputMessageWarning
```
user27182: Maybe this is too verbose? Should we rename them?
```suggestion
from pyvista.core.errors import VTKError
from pyvista.core.errors import VTKWarning
```

user27182: I renamed these to `VTKExecutionError` and `VTKExecutionWarning`

PR #7874, opened 2025-08-26
Replace `doc_image_cache` images with fixed legend symbols

user27182:
### Overview

Replace incorrect images in the `doc_image_cache`. The legend symbols were fixed by https://github.com/pyvista/pyvista/pull/7565 but were not replaced, and this difference is causing issues in https://github.com/pyvista/pyvista/pull/7848#issuecomment-3225146994


user27182:
Of the four images that were generating errors before, 3 have been downgraded to warnings, and one (`sphx_glr_antarctica_005.jpg`) is still causing errors.

So, `sphx_glr_antarctica_005.jpg` may be considered "flaky" as it has high variance between machines with an error of approx 750

akaszynski (approved):
LGTM. Thanks!

PR #7875, opened 2025-08-26
Update cached flaky doc images

user27182:
### Overview

Update images for #7848 based on #7874

For the new antarctica image, the exact same one from #7874 is used, and so should not create a merge conflict.


akaszynski:
Looks like we're exceeding the error threshold. My guess is we're just getting different images (NVIDIA vs. Mesa). At least it fails in 25 minutes and not 60...

user27182:
yes the failures are expected until #7874 is merged. I can add the other new images from that PR as well if you want, or we wait and merge main after that merges

PR #7876, opened 2025-08-26
Update vtk string formatting spec for latest vtk dev wheels (vtk 9.6)

user27182:
### Overview

The vtk dev tests are failing again. In part this due to string format changes from https://github.com/Kitware/VTK/commit/672e2596a28dbe1dc5e21ea933946251285a1cbc


pyvista/core/filters/data_set.py
```diff
@@ -143,7 +143,7 @@ def align(  # type: ignore[misc]  # noqa: PLR0917
         ...     scalars=dist,
         ...     scalar_bar_args={
         ...         'title': 'Distance to Source',
-        ...         'fmt': '%.1E',
+        ...         'fmt': '{:.1E}',
```
user27182: This change is correct, but requires the docs to be built with the latest vtk dev wheel. But currently, only the unit tests use the latest dev wheels. So this will need to be updated when the vtk version of the docs build is updated.

akaszynski (approved):

PR #7890, opened 2025-09-01
Only upload docs when not cancelled

akaszynski:
Documentation upload steps should not run when canceled. GitHub recommends using `!cancelled()` instead of `always()` in [Workflows and Actions - always](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always). Using this unblocks canceled documentation builds due to concurrency.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -113,26 +113,26 @@ jobs:
           html SPHINXOPTS="${{ env.SPHINX_JOBS }}"
 
       - name: Dump Sphinx Warnings and Errors
-        if: always()
+        if: ${{ !cancelled() }}
         run: if [ -e doc/sphinx_warnings.txt ]; then cat doc/sphinx_warnings.txt; fi
 
       - name: Dump VTK Warnings and Errors
-        if: always()
+        if: ${{ !cancelled() }}
         run: if [ -e doc/errors.txt ]; then cat doc/errors.txt; fi
 
       - name: Test Documentation
-        if: always()
+        if: ${{ !cancelled() }}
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -113,26 +113,26 @@ jobs:
           html SPHINXOPTS="${{ env.SPHINX_JOBS }}"
 
       - name: Dump Sphinx Warnings and Errors
-        if: always()
+        if: ${{ !cancelled() }}
```
user27182: I've seen the doc build "hang" a few times. In this case, the cancellation is related to a failure, and not simply due to a new commit being pushed. Would it still be useful to dump errors in this case for debugging? 

Same goes for uploading the test images artifacts... unlike the full docs uploads, these artifacts are very small and only take a few seconds to execute.

Maybe we keep most of these as always, and only change the full docs upload and non-interactive docs upload to only run if not cancelled?

.github/workflows/docs.yml
```diff
@@ -121,11 +121,11 @@ jobs:
         run: if [ -e doc/errors.txt ]; then cat doc/errors.txt; fi
 
       - name: Test Documentation
-        if: always()
+        if: ${{ !cancelled() }}
```
user27182, line 124: Earlier I mentioned to always upload images, but these images won't exist if this test doesn't run. 

I think it's fine to not run the tests after all when cancelled. And so you can add back the change to the Upload Debug Images step to only run if not cancelled. Sorry for the confusion !

user27182 (approved):

PR #7907, opened 2025-09-02
Add catch-all dependabot group

akaszynski:
We already have several matching groups for the `pip` ecosystem for `dependabot`, but we don't have a catchall and it's leading to this:

sc003

This PR adds a `all-else` group which will hopefully match the remainder of the dependencies.

See: [Dependabot options reference](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,6 +19,9 @@ updates:
         patterns:
           - "pytest"
           - "pytest-*"
+      all-else:
+        patterns:
+          - "*"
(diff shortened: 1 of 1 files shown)


akaszynski:
I closed the following PRs:
- #7899 
- #7900 
- #7901 
- #7902 
- #7903 

Once this PR is merged, I'll rerun dependabot manually and verify this group works.

PR #7911, opened 2025-09-02
Ignore `pydata-sphinx-theme` dependabot updates

user27182:
### Overview

This should not be updated, see https://github.com/pyvista/pyvista/pull/7095


tkoyama010 (approved):
LGTM

PR #7682, opened 2025-07-08
Update CubeAxesActor and PlanesAssembly text scaling for latest VTK dev wheels

user27182:
### Overview

Fix plotting errors for latest vtk dev:

https://github.com/pyvista/pyvista/actions/runs/16124354772/job/45497719131#step:10:10958
``` py
FAILED tests/plotting/test_plotting.py::test_xyz_assembly[Planes-default] - RuntimeError: test_xyz_assembly[Planes-default] Exceeded image regression error of 1000.0 with an image error equal to: 10892.196078431318
FAILED tests/plotting/test_plotting.py::test_xyz_assembly[Planes-position] - RuntimeError: test_xyz_assembly[Planes-position] Exceeded image regression error of 1000.0 with an image error equal to: 10898.52941176465
FAILED tests/plotting/test_plotting.py::test_xyz_assembly[Planes-orientation] - RuntimeError: test_xyz_assembly[Planes-orientation] Exceeded image regression error of 1000.0 with an image error equal to: 5807.2875816992555
FAILED tests/plotting/test_plotting.py::test_xyz_assembly[Planes-scale] - RuntimeError: test_xyz_assembly[Planes-scale] Exceeded image regression error of 1000.0 with an image error equal to: 2674.53071895428
FAILED tests/plotting/test_plotting.py::test_xyz_assembly[Planes-origin] - RuntimeError: test_xyz_assembly[Planes-origin] Exceeded image regression error of 1000.0 with an image error equal to: 5806.284967320171
FAILED tests/plotting/test_plotting.py::test_xyz_assembly[Planes-user_matrix] - RuntimeError: test_xyz_assembly[Planes-user_matrix] Exceeded image regression error of 1000.0 with an image error equal to: 1614.1281045751734
FAILED tests/plotting/test_plotting.py::test_planes_assembly - RuntimeError: test_planes_assembly Exceeded image regression error of 500.0 with an image error equal to: 3943.5098039215054
FAILED tests/plotting/test_plotting.py::test_planes_assembly_label_position[3D-40-x_label-yz-0.05] - RuntimeError: test_planes_assembly_label_position[3D-40-x_label-yz-0.05] Exceeded image regression error of 500.0 with an image error equal to: 8733.749019608196
[...]

PR #7914, opened 2025-09-02
Allow multiple cache images for `CubeAxesActor` tests

user27182:
### Overview

Use feature from https://github.com/pyvista/pytest-pyvista/pull/211 to store multiple cached images for a single test, and add back CubeAxesActor images for vtk < 9.6 that were removed in #7682

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,7 +87,7 @@ test = [
   'pyobjc-framework-Cocoa; sys_platform == "darwin"', # see pyvista/pulls/7828
   'pytest-cov<6.3.0',
   'pytest-mypy-plugins<3.3.0',
-  'pytest-pyvista==0.2.0',
+  'pytest-pyvista @ git+https://github.com/pyvista/pytest-pyvista.git@feat/doc',
   'pytest-xdist<3.9.0',
   'pytest<8.5.0',
   'pytest_cases<3.10.0',
--- a/tests/plotting/test_plotting.py
+++ b/tests/plotting/test_plotting.py
@@ -366,10 +366,6 @@ def test_plot(sphere, tmpdir, verify_image_cache, anti_aliasing):
     verify_image_cache.high_variance_test = True
     verify_image_cache.macos_skip_image_cache = True
     verify_image_cache.windows_skip_image_cache = True
-    # TODO: Change this to (9, 6, 0) when VTK 9.6 is released
-    if pyvista.vtk_version_info < (9, 5, 99):
-        # Axis labels changed substantially in VTK 9.6
-        verify_image_cache.skip = True
 
     tmp_dir = tmpdir.mkdir('tmpdir2')
     filename = str(tmp_dir.join('tmp.png'))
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -74,8 +74,9 @@ cov_flags =
 
 test_downloads = --test_downloads
 disallow_unused_cache = --disallow_unused_cache
+generated_image_name = --generated_image_name {env_name}
```
user27182: This PR will save cached images inside a subdir (with the same name as the test), and the image name inside the subdir can be anything. I want to make the image names meaningful so you can somewhat trace back its origins. (e.g. if multiple cached images are saved, the different names should help understand where the images came from)

Currently `env_name` is used, which will be something like: `py3.13-numpy_nightly-vtk_latest-cov-plotting`
A similar var is the failed image artifact name from the test matrix, which will be something like:  `failed_test_images-Linux-3.13-latest-nightly-ubuntu-22.04-self-hosted`

but neither of these names are what quite I'm looking for. I'd like to name the images something like `Linux-3.13-pyvista-0.47dev0-vtk-9.5.20250830.dev0`, i.e. capture the os, actual pyvista and vtk runtime versions used to generated the images (maybe include py version too). I don't want to include `'vtk_latest'`.

@beroda do you know if it's possible to store these somehow with tox into the `--generated_image_name` value? I tried using various approaches with `setenv` and `command_pre` to get runtime values (e.g.  by running `python -c 'import vtk; print(vtk.__version__)'` but have not been successful

edabor: I guess you need something like https://github.com/equinor/tox-runtime-env/blob/main/README.md but the project is archived, might not be ideal.

Or you can create a `GENERATED_IMAGE_NAME` env variable outside of `tox` using GitHub variables and get it with:
```ini
generated_image_name = --generated_image_name {env:GENERATED_IMAGE_NAME:default}
```
Note that a default value must be specified if you want to run tox locally without needing to set any env variable

user27182: Thanks. Using env vars like this seems a bit too complex. I have found it's easier to set this directly from within `pytest-pyvista` as a default config value, so I'll just do that.

edabor:
BTW, `vtk_dev` job does not seem to be correctly pulling the latest vtk.
https://github.com/pyvista/pyvista/actions/runs/18485826814/job/52669082517#step:7:33

It should fail, see https://github.com/pyvista/pyvista/pull/8003

Maybe due to self-hosted pypi ?

user27182:
Maybe it's due to package caching? Might need to use `--no-cache` ?

user27182:
> BTW, `vtk_dev` job does not seem to be correctly pulling the latest vtk. https://github.com/pyvista/pyvista/actions/runs/18485826814/job/52669082517#step:7:33
> 
> It should fail, see #8003
> 
> Maybe due to self-hosted pypi ?

Is this a UV or tox thing? Even the numpy nightlies are not being installed:

https://github.com/pyvista/pyvista/actions/runs/18485826814/job/52669082524#step:7:48

akaszynski:
> Maybe due to self-hosted pypi ?

That's correct. I'll resolve that. In the mean time, should we delay merging this PR?

user27182:
This PR adds tests to ensure VTK _less than_ 9.6 has coverage for CubeAxesActor. And those tests are executing and passing. So I think this is good to merge now.

PR #7915, opened 2025-09-02
Use `tox` for mypy

beroda:
### Overview

Part of https://github.com/pyvista/pyvista/pull/7441


akaszynski:
No idea why labeler is failing now...

PR #7892, opened 2025-09-01
Suppress joblib related warnings

akaszynski:
Suppress joblib related warnings in our docbuild:
```
Cannot register /tmp/joblib_memmapping_folder_1225_18016e78c6144e609b889a496fb733f7_57b282146cbd427abf4b447d20e2e8a7 for automatic cleanup: unknown resource type folder
```

~Potentially~ Address #7877 by disabling vestigial `self.get_image_depth()`. Seems to serve no purpose and running `run_image_filter` back to back seems to be unstable. We'll have to run the doc build several times to see if it's stable.

PR #7918, opened 2025-09-03
Complete double rendering patch

akaszynski:
#7892 broke the build on `main` as it relied on cached gallery examples before removing the vestigial `get_image_depth()`. Turns out, the side effect of calling `get_image_depth` is to double render, which in a handful of gallery examples noticeably changes the rendered image. As this does not match what a user would see when first rendering the plotter, it's would be ideal to keep it removed.

Most of the changed images are benign, but critically the `01-filters/rotate.py` example sets the `camera` attribute of `Plotter` using a `Camera` object, which results in poor clipping ranges on the first render. This is most in the rotated cow example:

https://github.com/pyvista/pyvista/blob/eff2016618978354b4114df429e15809d1c2092b/examples/01-filter/rotate.py#L61-L71

sphx_glr_rotate_002

To fix this, this PR replaces `Camera` with its equivalent camera position tuple in the `rotate.py` example.

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -300,6 +300,13 @@ jobs:
         if: ${{ !cancelled() }}
         run: xvfb-run tox run -e py${{ matrix.python-version }}-plotting-vtk_dev
 
+      - name: Upload Images for Failed Tests
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: failed_test_images-${{ github.job }}-${{ join(matrix.* , '-') }}
+          path: _failed_test_images
+
       - name: Upload Generated Images
         if: always()
         uses: actions/upload-artifact@v4


user27182:
> broke the build on `main` as it relied on cached gallery examples

This has happened many times in the past. Now that build times are relatively fast, maybe we only enable caching for draft PRs? Disable it otherwise, which should help catch these errors before being merged into main.

akaszynski:
> > broke the build on `main` as it relied on cached gallery examples
> 
> 
> 
> This has happened many times in the past. Now that build times are relatively fast, maybe we only enable caching for draft PRs? Disable it otherwise, which should help catch these errors before being merged into main.

Yes. I'll change that here in this PR.

.github/workflows/docs.yml
```diff
@@ -84,13 +84,6 @@ jobs:
           echo PYVISTA_EXAMPLE_DATA_PATH=$(python -c "from pyvista import examples; print(examples.USER_DATA_PATH)") >> $GITHUB_ENV
           pip list
 
-      - name: Cache Sphinx-Gallery Examples
-        uses: actions/cache@v4
-        if: env.USE_CACHE == 'true' && (! github.event_name == 'pull_request' || ! contains(github.event.pull_request.labels.*.name, 'no-gallery-cache'))
-        with:
-          path: doc/source/examples/
-          key: doc-examples-${{ hashFiles('pyvista/_version.py') }}
-
```
user27182: In my experience, the gallery cache is the most impactful cache. The other caches "Cache example data" and "Cache Sphinx build directory" seem to not really provide any substantial speedup (this can be tested using the various `no-cache` labels). So if we're removing gallery caching, then maybe we should remove all build caching altogether?

This would greatly simplify the action logic if we did. The `no-cache` labels would also need to be removed, and `CONTRIBUTING` also makes mention of these which would need to be updated. Due to the larger scope, maybe removing all caches should be a separate PR?

akaszynski:
> Due to the larger scope, maybe removing all caches should be a separate PR?

Yes, I'd prefer that. We can land this PR and make the CI a bit faster in the meantime.

@user27182, I'll leave that issue and/or followup PR to your discretion.

user27182:
> > Due to the larger scope, maybe removing all caches should be a separate PR?
> 
> Yes, I'd prefer that. We can land this PR and make the CI a bit faster in the meantime.
> 
> @user27182, I'll leave that issue and/or followup PR to your discretion.

If that's the case then I don't think we should remove the gallery caching in this PR, since the current changes made w.r.t. to that are incomplete in this PR (e.g. removing it makes the `no-gallery-cache` label useless, and its use is mentioned in contributing). We should fully remove it properly at once or leave it alone for now. Keep this PR to doc image/example changes only.

.github/workflows/docs.yml
```diff
@@ -84,13 +84,6 @@ jobs:
           echo PYVISTA_EXAMPLE_DATA_PATH=$(python -c "from pyvista import examples; print(examples.USER_DATA_PATH)") >> $GITHUB_ENV
           pip list
```
akaszynski:
```suggestion

      - name: Cache Sphinx-Gallery Examples
        uses: actions/cache@v4
        if: env.USE_CACHE == 'true' && (! github.event_name == 'pull_request' || ! contains(github.event.pull_request.labels.*.name, 'no-gallery-cache'))
        with:
          path: doc/source/examples/
          key: doc-examples-${{ hashFiles('pyvista/_version.py') }}

```

akaszynski:
> If that's the case then I don't think we should remove the gallery caching in this PR, since the current changes made w.r.t. to that are incomplete in this PR (e.g. removing it makes the `no-gallery-cache` label useless, and its use is mentioned in contributing). We should fully remove it properly at once or leave it alone for now. Keep this PR to doc image/example changes only.

Done in [e728a1c](https://github.com/pyvista/pyvista/pull/7918/commits/e728a1c60ff52d783a32fcca4a438083301bed15).

Issue #7923, opened 2025-09-03
Interactive examples broken


beroda:
> I think interactive rendering of triangle strips won't work due to vtk.js serialization. We might consider disabling the interactive views for those examples.

This is not related to triangle strips. See the examples below.

- In PolyData documentation:

<img width="897" height="749" alt="image" src="https://github.com/user-attachments/assets/90962e04-5183-4e3d-a787-3ba6fd32aa4b" />

- In https://docs.pyvista.org/examples/00-load/create_poly#create-poly-example

<img width="819" height="926" alt="image" src="https://github.com/user-attachments/assets/a68d7ddb-6a20-4edd-aa59-deeb8bdebe94" />

My guess is that it is due to the scalars being integer data (`np.arange`), as the error pointed out by @echedey-ls is `Uncaught (in promise) TypeError: can't convert BigInt to number`

_Originally posted by @beroda in https://github.com/pyvista/pyvista/issues/7922#issuecomment-3249865344_

PR #7924, opened 2025-09-03
Fix interactive plots using `np.arange`

beroda:
### Overview

Closes https://github.com/pyvista/pyvista/issues/7923


MatthewFlamm:
@akaszynski A big downside of the change to hash based image naming in the testing/doc image cache is that it becomes very difficult to compare image changes in review.  I cannot figure out which images should be compared in this PR.  I'd argue that the previous state was more preferable.

akaszynski:
> @akaszynski A big downside of the change to hash based image naming in the testing/doc image cache is that it becomes very difficult to compare image changes in review.  I cannot figure out which images should be compared in this PR.  I'd argue that the previous state was more preferable.

A few of my recent PRs that involved changing the cache were quite tedious and I agree with you. I'll see if I can create a workaround.

user27182:
Would it help with the diff if each image was stored in a subdirectory (with a name that does _not_ include the hash), and the hash was only applied to the image name inside the subdirectory? E.g. would the `--generate_subdirs` feature from https://github.com/pyvista/pytest-pyvista/pull/213 fix this?

akaszynski:
> Would it help with the diff if each image was stored in a subdirectory (with a name that does _not_ include the hash), and the hash was only applied to the image name inside the subdirectory? E.g. would the `--generate_subdirs` feature from [pyvista/pytest-pyvista#213](https://github.com/pyvista/pytest-pyvista/pull/213) fix this?

I think the main annoyance (at least for me) was finding the right file to rename, especially when the hash changed. This might be a bit easier if there were several images under a single directory, which wouldn't be affected by race conditions.

user27182:
> I think the main annoyance (at least for me) was finding the right file to rename, especially when the hash changed.

This may be because the hash is placed before the plot number. Appending the hash to the very end of the file name would help with this since plot 0 with one hash should be next to the other plot 0 with a different hash, if they were all placed in the same folder.

akaszynski:
Sorry about the file names @beroda. It's absolutely maddening getting the names to line up... I'll have to fix it.

edabor:
Here are a summary of fixed scenes:

| This PR | The official documentation |
|---------|-------------------------------|
https://68bb05815292099032b56734--pyvista-dev.netlify.app/user-guide/data_model#cell-data | https://docs.pyvista.org/user-guide/data_model#cell-data|
|https://68bb05815292099032b56734--pyvista-dev.netlify.app/user-guide/what-is-a-mesh#point-data|https://docs.pyvista.org/user-guide/what-is-a-mesh#point-data|
|https://68bb05815292099032b56734--pyvista-dev.netlify.app/user-guide/what-is-a-mesh#cell-data|https://docs.pyvista.org/user-guide/what-is-a-mesh#cell-data|
|https://68bb05815292099032b56734--pyvista-dev.netlify.app/examples/00-load/create_poly.html#|https://docs.pyvista.org/examples/00-load/create_poly#|
|https://68bb05815292099032b56734--pyvista-dev.netlify.app/examples/00-load/create_spline|https://docs.pyvista.org/examples/00-load/create_spline
|https://68bb05815292099032b56734--pyvista-dev.netlify.app/api/core/_autosummary/pyvista.polydatafilters.ribbon#pyvista.PolyDataFilters.ribbon|https://docs.pyvista.org/api/core/_autosummary/pyvista.polydatafilters.ribbon#pyvista.PolyDataFilters.ribbon|
|https://68bb05815292099032b56734--pyvista-dev.netlify.app/api/core/_autosummary/pyvista.polydata#pyvista.PolyData|https://docs.pyvista.org/api/core/_autosummary/pyvista.polydata#pyvista.PolyData|

user27182 (commented):
It's good to fix the interactive plots, though it seems like this is more of a workaround to deal with an upstream vtk js issue.

If I'm reading these examples without any knowledge of this interactive plot bug, I think it's a bit odd to see integer values cast to float type for no apparent reason. Should we add comments about why float dtype is being used for integer values? 

Is there an issue to track and fix this upstream with vtk? I think it's likely this will recur with future examples / interactive plots otherwise. Should we keep #7923 open or open a separate issue to continue tracking this so that these dtype changes can possibly be reverted once integers are better supported in interactive plots?

edabor:
> It's good to fix the interactive plots, though it seems like this is more of a workaround to deal with an upstream vtk js issue.

Agreed, more a workaround than a real bug fixing. We should produce a MWE within an issue  for vtk.js to let them know about it.

> If I'm reading these examples without any knowledge of this interactive plot bug, I think it's a bit odd to see integer values cast to float type for no apparent reason. Should we add comments about why float dtype is being used for integer values?

I don't think adding comments to examples is worth it, since it would add complexity to documentation that is intended to be straightforward.

> Is there an issue to track and fix this upstream with vtk? I think it's likely this will recur with future examples / interactive plots otherwise. Should we keep #7923 open or open a separate issue to continue tracking this so that these dtype changes can possibly be reverted once integers are better supported in interactive plots?

I think nothing has to be fixed on vtk, more on vtk.js side.
Let's open an issue on `trame` and close #7923 ?

edabor:
BTW, I think we should find a way to also test documentation serialized `.vtksz` files to check they are readable without any issues.
We might find some hidden interactive documentation issues ...

Will investigate and see if something is doable.

akaszynski:
> this is more of a workaround to deal with an upstream vtk js issue.

Hopefully, switching to VTK WASM would fix this.

akaszynski:
https://68bb05815292099032b56734--pyvista-dev.netlify.app/api/core/_autosummary/pyvista.polydatafilters.ribbon#pyvista.PolyDataFilters.ribbon still isn't rendering:

```
static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10 Uncaught (in promise) TypeError: Cannot mix BigInt and other types, use explicit conversions
    at static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:165647
    at t.mapScalarsToTexture (static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:165769)
    at e.mapScalars (static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:163283)
    at e.buildBufferObjects (static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:300598)
    at e.updateBufferObjects (static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:300290)
    at e.renderPieceStart (static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:298663)
    at e.renderPiece (static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:300149)
    at e.render (static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:271501)
    at e.opaquePass (static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:271235)
    at e.apply (static_viewer.html?fileURL=../_images/plot_directive/api/core/_autosummary/pyvista-PolyDataFilters-ribbon-c6f300fc28e9f118_00_00.vtksz:10:44908)
```

This might be a stale deployment as the the ribbon example doesn't match the changes in this PR.

akaszynski (approved):

akaszynski:
I didn't see any issues from the downloaded HTML. Seems that netlify was stale.

PR #7926, opened 2025-09-03
Report local VTK data cache usage

akaszynski:
Add additional debug info to the report to indicate local VTK data cache usage (or lack thereof).


tox.ini
```diff
@@ -86,7 +86,7 @@ all_testing_cmdline =
     !cov-!plotting-!core: pytest {[testenv]all_flags}  {posargs}
 
 software_report_cmdline=
-    python -c "import pyvista;print(pyvista.Report(gpu=False));from pyvista import examples;print('User data path:', examples.USER_DATA_PATH)"
+    python -c "import pyvista;print(pyvista.Report(gpu=False));from pyvista import examples;print('User data path:', examples.USER_DATA_PATH);print('Using local VTK data cache:', examples.downloads._FILE_CACHE);print('VTK data source:', examples.downloads.SOURCE)"
```
user27182: Instead of handling imports and prints with python code, I suggest adding this as a formal feature of `Report`. This is done in https://github.com/pyvista/pyvista/pull/7927. Then we can do:
```suggestion
    python -c "import pyvista;print(pyvista.Report(gpu=False, downloads=True)"
```

tox.ini
```diff
@@ -86,7 +86,7 @@ all_testing_cmdline =
     !cov-!plotting-!core: pytest {[testenv]all_flags}  {posargs}
 
 software_report_cmdline=
-    python -c "import pyvista;print(pyvista.Report(gpu=False));from pyvista import examples;print('User data path:', examples.USER_DATA_PATH)"
+    python -c "import pyvista;print(pyvista.Report(gpu=False, downloads=True)"
```
akaszynski:
```suggestion
    python -c "import pyvista;print(pyvista.Report(gpu=False, downloads=True))"
```

PR #7930, opened 2025-09-03
Remove caching from docs workflow

user27182:
### Overview

The cache was created to speed up the docs build. But since #7848 was merged, full doc builds take approx 35 mins, and caching can shave approx 10 minutes off. But caching also creates issues if any cached items are modified, because those changes will not be reflected in the docs build (the stale cache is used instead). This sometimes leads to PRs being merged with stale changes that then cause breakages to main, and also causes issues for other PRs since they now also stale caches that need to be updated to match the changes in main.

I propose removing the cache altogether to simplify things and ensure no stale caches are used by PRs. See also #7918

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -355,11 +355,21 @@ jobs:
         run: pip install tox-uv
 
       - name: Core Testing (no GL)
-        run: tox run -e py${{ matrix.python-version }}-core
+        uses: nick-fields/retry@ce71cc2ab81d554ebbe88c79ab5975992d79ba08
+        with:
+          timeout_minutes: 12
+          max_attempts: 3
+          retry_on: timeout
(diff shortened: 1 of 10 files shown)


user27182:
Instead of removing it, we could make the cache opt-in rather than opt-out. And maybe simplify the labels to use a single one to opt in to all caching instead of using 3 separate labels

EDIT: my preference is to remove it , was just giving alternatives in case that's preferred by others

akaszynski:
> Instead of removing it, we could make the cache opt-in rather than opt-out. And maybe simplify the labels to use a single one to opt in to all caching instead of using 3 separate labels

The doc build at 30 minutes isn't even the longest CI step at the moment... most of the time it's either waiting for the Apple runners or Linux (40 minutes). Let's keep as-is and keep it simple.

user27182:
> Should we remove the labels from the project settings once the PR is merged ?

I have removed the three `no_cache` labels from the project

PR #7916, opened 2025-09-02
Use `tox` to build documentation

beroda:
### Overview

Part of #7441

PR #7945, opened 2025-09-10
Set VTK_DEFAULT_OPENGL_WINDOW for specific tox jobs

edabor:
### Overview

`VTK_DEFAULT_OPENGL_WINDOW` env variable is set on the `tox` config. It prevents users from changing its value outside of tox.
This PR fixes that by setting the env variable specifically when the `CI` env variable is true.

Moreover, `VTK_DEFAULT_OPENGL_WINDOW ` is also explicitely specified for `doctest-modules` job.

### DIFF
--- a/pyvista/core/utilities/points.py
+++ b/pyvista/core/utilities/points.py
@@ -795,7 +795,7 @@ def principal_axes(
 
     Show the standard deviation along each axis.
 
-    >>> std
+    >>> std  # doctest:+SKIP
     array([3.0149 , 1.5074 , 0.7035], dtype=float32)
 
     Compare this to using :meth:`numpy.std` for the computation.
@@ -810,7 +810,7 @@ def principal_axes(
 
     Convert the values to proportions for analysis.
 
-    >>> std / sum(std)
+    >>> std / sum(std)  # doctest:+SKIP
     array([0.5769149 , 0.28845742, 0.1346276 ], dtype=float32)
 
     From this result, we can determine that the axes explain approximately


.github/workflows/docs.yml
```diff
@@ -15,6 +15,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  VTK_DEFAULT_OPENGL_WINDOW: "vtkEGLRenderWindow"
```
user27182: Is there a way to define `"vtkEGLRenderWindow"` as a variable in tox for CI use only, such that it can be reused, e.g. for the docstring/doctest workflow? I.e. address this comment https://github.com/pyvista/pyvista/pull/7861#issuecomment-3212292384

Maybe define `CI_VTK_DEFAULT_OPENGL_WINDOW = "vtkEGLRenderWindow"` as the variable in tox, and do something like:
```suggestion
  VTK_DEFAULT_OPENGL_WINDOW: CI_VTK_DEFAULT_OPENGL_WINDOW
```

user27182: alternatively, define a pre-command that checks if the env var `CI` is true, and if it is then export `VTK_DEFAULT_OPENGL_WINDOW = "vtkEGLRenderWindow"`? This way we wouldn't need to define it as an env var inside the workflow itself, and would be self-contained inside the tox config.

user27182: We can also address this separately with a future PR that adds tox to the docstring tests

edabor: > Is there a way to define `"vtkEGLRenderWindow"` as a variable in tox for CI use only, such that it can be reused, e.g. for the docstring/doctest workflow? I.e. address this comment [#7861 (comment)](https://github.com/pyvista/pyvista/pull/7861#issuecomment-3212292384)
> 

Don't think it's possible to set and `env` variable in `tox` depending on the value of another one (ie. CI)

> alternatively, define a pre-command that checks if the env var `CI` is true, and if it is then export `VTK_DEFAULT_OPENGL_WINDOW = "vtkEGLRenderWindow"`? This way we wouldn't need to define it as an env var inside the workflow itself, and would be self-contained inside the tox config.

commands output cannot be captured in subsequent commands, so I don't think this is possible either.

Maybe through a dedicated homemade plugin ...

edabor: @user27182 does https://github.com/pyvista/pyvista/pull/7945/commits/f7c8862ba784ad93fb0f2870b0b201cf5f9c6ad4 is what you intended ?

If yes, do you prefer cherry-picking https://github.com/pyvista/pyvista/pull/7945/commits/f7c8862ba784ad93fb0f2870b0b201cf5f9c6ad4 inside another dedicated PR ?

user27182: Maybe we set a new env var `PYVISTA_CI=true` in tox, and set the opengl window var in pyvista's init before any vtk imports?

``` py
if os.environ.get("PYVISTA_CI", "").lower() == "true":
    os.environ["VTK_DEFAULT_OPENGL_WINDOW"] = "vtkEGLRenderWindow"
```
But for now it's OK, we can try this (or not) in the docstring workflow PR when there's an actual duplication of variables

user27182: > @user27182 does [f7c8862](https://github.com/pyvista/pyvista/commit/f7c8862ba784ad93fb0f2870b0b201cf5f9c6ad4) is what you intended ?
> 
> If yes, do you prefer cherry-picking [f7c8862](https://github.com/pyvista/pyvista/commit/f7c8862ba784ad93fb0f2870b0b201cf5f9c6ad4) inside another dedicated PR ?

I think this works? Not at all familiar with tox. If it lets us define it exactly once in tox and reuse it in other workflows as needed then that's goal achieved.

I have no preference for adding this as part of this PR or as part of a future docstring workflow tox PR.

edabor: Currently, it only works for the `docs-build` environment but we can extend it to other environments (ie. doctest) if needed.

https://github.com/pyvista/pyvista/blob/f7c8862ba784ad93fb0f2870b0b201cf5f9c6ad4/toxfile.py#L14-L16

> I have no preference for adding this as part of this PR or as part of a future docstring workflow tox PR.

docstring are already tested with tox
https://github.com/pyvista/pyvista/blob/f7c8862ba784ad93fb0f2870b0b201cf5f9c6ad4/tox.ini#L209-L220

user27182: > docstring are already tested with tox

I see, then I guess we should include `doctest-modules` in the list of cases where we want to set `VTK_DEFAULT_OPENGL_WINDOW`? Is that correct, @akaszynski, we want these two workflows to use this consistently?

And I went to verify it's working by checking the software report, but I guess there is none?? Or at least I couldn't find it. Maybe it was excluded with #7916 ?

And as an additional unrelated note, is it an easy change to _exclude_ `-v` from the documentation image tests? I find the extra 5000 lines in the log for all the test cases to be unnecessary for this. (`-v` was not used for this prior to #7916)

akaszynski: > I see, then I guess we should include `doctest-modules` in the list of cases where we want to set `VTK_DEFAULT_OPENGL_WINDOW`? Is that correct, @akaszynski, we want these two workflows to use this consistently?

I'd say so. It's better to be explicit IMO to try to set the render window, and it's a bit of future proofing to ensure that should we decide to use a GPU in one of our workflows.

akaszynski: > And I went to verify it's working by checking the software report, but I guess there is none?? Or at least I couldn't find it. Maybe it was excluded with #7916 ?

Looks like we need to add that back in. We can open an issue or simply add it in this PR.

edabor: > And I went to verify it's working by checking the software report, but I guess there is none?? Or at least I couldn't find it. Maybe it was excluded with #7916 ?

Fixed in 1a0e117d2302f643cb38497bd315ecb83eff3908


> > I see, then I guess we should include `doctest-modules` in the list of cases where we want to set `VTK_DEFAULT_OPENGL_WINDOW`? Is that correct, @akaszynski, we want these two workflows to use this consistently?
> 
> I'd say so. It's better to be explicit IMO to try to set the render window, and it's a bit of future proofing to ensure that should we decide to use a GPU in one of our workflows.

Done in d79dc4172a419fc3b024c307c076204981b29f60

> And as an additional unrelated note, is it an easy change to _exclude_ `-v` from the documentation image tests? I find the extra 5000 lines in the log for all the test cases to be unnecessary for this. (`-v` was not used for this prior to #7916)

Done in a834b4336094a759c7bed1a6600a8f31348f0d41

user27182: Thanks @beroda ! I was expecting the report to include "vtkEGLRenderWindow" as the render window, but instead it's just a generic "vtkRenderWindow".
https://github.com/pyvista/pyvista/actions/runs/17657586560/job/50184056166#step:7:40

This does not appear to be specific to this workflow, it seems the unit tests, type checking, docstring tests also do not appear to be reporting a concrete render window class.

user27182: > Thanks @beroda ! I was expecting the report to include "vtkEGLRenderWindow" as the render window, but instead it's just a generic "vtkRenderWindow". https://github.com/pyvista/pyvista/actions/runs/17657586560/job/50184056166#step:7:40
> 
> This does not appear to be specific to this workflow, it seems the unit tests, type checking, docstring tests also do not appear to be reporting a concrete render window class.

This is reported/tracked in #7948

edabor:
Per https://github.com/pyvista/pyvista/pull/7945#discussion_r2342573960, waiting for https://github.com/pyvista/pyvista/issues/7948 to be closed before continuing this PR.

.github/workflows/docs.yml
user27182: With the fix from #7949 , on Windows unit tests we have `Render Window : vtkOSOpenGLRenderWindow`
https://github.com/pyvista/pyvista/actions/runs/17685334349/job/50268816978?pr=7949#step:7:57

This differs from the default of EGL set for the docs. Perhaps we want to use the same render window for _all_ CI jobs, not just docs/docstrings? Not sure if that will cause any generated images will differ and introduce regressions... but if not perhaps we should consider doing so for consistency?

user27182: > I was expecting the report to include "vtkEGLRenderWindow" as the render window

This is now working, and "vtkEGLRenderWindow" is reported by the build
https://github.com/pyvista/pyvista/actions/runs/17700376046/job/50305304861?pr=7945#step:7:41
and the docstring tests
https://github.com/pyvista/pyvista/actions/runs/17700376033/job/50305304850?pr=7945#step:7:41

Just need to determine whether we want this render window set just for these two jobs or if we want this for all CI jobs, including unit tests.

edabor:
Let's continue the thread here

> Just need to determine whether we want this render window set just for these two jobs or if we want this for all CI jobs, including unit tests.

@akaszynski what's your feeling about it ? I'm +1 with the idea of specifying the value of `VTK_DEFAULT_OPENGL_WINDOW` in CI only for all jobs that run plotting tests for consistency.

user27182:
I think there was a previous PR where @akaszynski also had issues with these docstrings. We could probably just skip the doctest for these, clean up the output and not use numpy printoptions. These functions have plenty of test coverage for numerical error / sign differences in the unit tests and don't need the coverage from the docstrings.

tox.ini
```diff
@@ -216,7 +216,7 @@ setenv =
     PYVISTA_OFF_SCREEN = true
     PYTEST_ADDOPTS = --color=yes -v {env:PARALLEL:}
 commands =
-	modules: pytest --doctest-modules {env_site_packages_dir}{/}pyvista {posargs}
+	modules: pytest --doctest-modules --doctest-continue-on-failure {env_site_packages_dir}{/}pyvista {posargs}
```
edabor, line 219: Found convenient to use `--doctest-continue-on-failure` since it enables to see all docstring issues and not only the first one.

pyvista/core/utilities/points.py
```diff
@@ -333,9 +333,8 @@ def fit_plane_to_points(  # noqa: PLR0917
 
     Note that it is pointing in the positive z-direction.
 
-    >>> with np.printoptions(precision=6, suppress=True):  # for readability
-    ...     normal
-    pyvista_ndarray([0., 0., 1.], dtype=float32)
+    >>> normal  # doctest:+SKIP
+    pyvista_ndarray([0.0, 0.0, 1.0], dtype=float32)
```
akaszynski, line 337: Ran into rounding errors with this. The workaround was to round, but was quite verbose. Alternative would be converting to `np.float64`

akaszynski:
> @akaszynski what's your feeling about it ? I'm +1 with the idea of specifying the value of `VTK_DEFAULT_OPENGL_WINDOW` in CI only for all jobs that run plotting tests for consistency.

The render window only needs to be set for plotting tests, though it shouldn't matter outside of the plotting jobs.

edabor:
> > @akaszynski what's your feeling about it ? I'm +1 with the idea of specifying the value of `VTK_DEFAULT_OPENGL_WINDOW` in CI only for all jobs that run plotting tests for consistency.
> 
> The render window only needs to be set for plotting tests, though it shouldn't matter outside of the plotting jobs.

Let's open a new PR to explicitly set VTK_DEFAULT_OPENGL_WINDOW for all tests.

PR #7951, opened 2025-09-13
Check polyhedron_faces when determining equivalency

akaszynski:
Check polyhedron faces when comparing the equivalency of two `UnstructuredGrids`.

### DIFF
--- a/pyvista/report.py
+++ b/pyvista/report.py
@@ -10,22 +10,55 @@ import scooby
 
 from pyvista._deprecate_positional_args import _deprecate_positional_args
 
-_cmd = """\
-import pyvista; \
-plotter = pyvista.Plotter(notebook=False, off_screen=True); \
-plotter.add_mesh(pyvista.Sphere()); \
-plotter.show(auto_close=False); \
-gpu_info = plotter.render_window.ReportCapabilities(); \
-print(gpu_info); \
-plotter.close()\
+_cmd_render_window_info = """
+from vtkmodules.vtkRenderingCore import vtkRenderer, vtkRenderWindow
+import vtkmodules.vtkRenderingOpenGL2
+ren = vtkRenderer()
+win = vtkRenderWindow()
+win.OffScreenRenderingOn()
+win.AddRenderer(ren)
+win.Render()
+print(win.ReportCapabilities())
+print('vtkRenderWindow class name: ', win.GetClassName())
 """
+
+_cmd_math_text = """
+import vtkmodules.vtkRenderingFreeType
+import vtkmodules.vtkRenderingMatplotlib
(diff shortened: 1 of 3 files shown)


MatthewFlamm:
Why this instead of the cell array to cover all cell types?

akaszynski:
> Why this instead of the cell array to cover all cell types?

At least in the versions I've tested (9.4 and 9.5.X), polyhedron cells are stored in two separate [vtkCellArray](https://vtk.org/doc/nightly/html/classvtkCellArray.html)s.
- https://vtk.org/doc/nightly/html/classvtkUnstructuredGrid.html#a4c5241068181cb067d046e9a92e58bf6
- https://vtk.org/doc/nightly/html/classvtkUnstructuredGrid.html#a874acf928c41c24dc0e8ee391f2d718f

There's a discussion on why and how this was changed:
https://discourse.vtk.org/t/polyhedral-cells-storing-faces-as-vtkcellarray/5656

Given the failures for earlier versions, we're going to have to include backwards compatibility for this check.

tests/core/test_dataobject.py
```diff
@@ -108,6 +108,23 @@ def test_unstructured_grid_eq(hexbeam):
...
+    if pv.vtk_version_info < (9, 4):
+        poly_faces = poly.GetFaces()
+    else:
+        poly_faces = poly_copy.GetPolyhedronFaces().GetConnectivityArray()
+    pv.convert_array(poly_faces)[2] += 1
+    assert poly != poly_copy
+
+    # sanity check: ensure that modifying polyfaces doesn't change the
+    # underlying cell connectivity
+    assert np.allclose(poly.cell_connectivity, poly_copy.cell_connectivity)
```
akaszynski, line 126: @MatthewFlamm, I added this so we're sure that we're detecting changes in polyfaces and not the `cell_connectivity`.

akaszynski:
> Should we also be checking the cell array is equal for other cell types?

I don't think so. We only included it here to make sure that we're genuinely changing the polyhedrons and not the underlying cell array. It's helpful for this test since VTK has been changing their API and if it's changed again, we can catch it here.

MatthewFlamm:
> I don't think so. We only included it here to make sure that we're genuinely changing the polyhedrons and not the underlying cell array. It's helpful for this test since VTK has been changing their API and if it's changed again, we can catch it here.

Sorry I wasnt clear.  I didn't mean due this test here.  I meant to check the cell arrays during the equivalency check?  I don't see this currently.

akaszynski:
> > I don't think so. We only included it here to make sure that we're genuinely changing the polyhedrons and not the underlying cell array. It's helpful for this test since VTK has been changing their API and if it's changed again, we can catch it here.
> 
> Sorry I wasnt clear. I didn't mean due this test here. I meant to check the cell arrays during the equivalency check? I don't see this currently.

It we don't we should.

PR #7953, opened 2025-09-14
Do not always get last image depth

akaszynski:
Getting the last image depth results in a double render which makes our CI unstable (see https://github.com/pyvista/pyvista/actions/runs/17712293267/job/50332593824?pr=7951). It's also unnecessary unless you wish to retrieve the last image depth after rendering.

This PR disables it by default unless it's explicitly enabled within `show`.

A side effect of this change is rendering is around 28% faster (on my machine):

```bash
$ git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
$ python ~/tmp/tmp.py
100 renders in 2.95 seconds
```

```bash
$ git checkout fix/no-last-image-depth
Switched to branch 'fix/no-last-image-depth'
Your branch is up to date with 'origin/fix/no-last-image-depth'.
$ python ~/tmp/tmp.py
100 renders in 2.13 seconds
```

Timing script:

```py
import pyvista as pv
import timeit

mesh = pv.Sphere()


def render():
    mesh.plot(off_screen=True)


num = 100
ttot = timeit.timeit(render, number=num)
print(f"{num} renders in {ttot:.3} seconds")
```

### Bonus

- This PR changes several `p` --> `pl` for consistency. We use `plotting`, `pl`, and `p` throughout our examples, docs, and tests and it would be great to be consistent.
- Includes minor changes to the plot_directive to ensure proper error reporting.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -83,6 +83,10 @@ jobs:
           find doc/_build/mini18n-html -mindepth 1 -maxdepth 1 -type d -exec cp -rf {} doc/_build/html/ \;
           rm -rf doc/_build/mini18n-html
 
+      - name: Dump Sphinx Build Warnings and Errors
+        if: always()
+        run: if [ -e doc/sphinx_warnings.txt ]; then cat doc/sphinx_warnings.txt; fi
+
       - name: Upload HTML documentation
         if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v5


tox.ini
```diff
@@ -233,6 +233,7 @@ setenv =
     PYVISTA_OFF_SCREEN = true
     ALLOW_PLOTTING = true
     SHELLOPTS = errexit:pipefail
+    SPHINX_DEBUG = 1
```
akaszynski:
```suggestion
```

akaszynski:
@user27182, I think I've finally found the issue that's been plaguing the doc build and determined why it's so sporadic. After fixing the `plot_directive` error reporting:

```
reading sources... [100%] tags/widgets .. user-guide/what-is-a-mesh
/home/runner/_work/pyvista/pyvista/doc/source/user-guide/data_model.rst:470: WARNING: Exception occurred in plotting data_model-f9ad1bc032b72627
 from /home/runner/_work/pyvista/pyvista/doc/source/user-guide/data_model.rst:
  File "/home/runner/_work/pyvista/pyvista/.tox/docs-build/lib/python3.13/site-packages/pyvista/ext/plot_directive.py", line 480, in _run_code
    exec(code, ns)
    ~~~~^^^^^^^^^^
  File "<string>", line 1, in <module>
NameError: name 'np' is not defined [docutils]
looking for now-outdated files... none found
pickling environment... done
```

Looks like we're not importing `numpy` before its first use in `data_model.rst`, which is correct:

```

.. pyvista-plot::
   :context:
   :include-source: False

   >>> import numpy as np  # missing
   >>> ugrid.cell_data['my-data'] = np.arange(ugrid.n_cells, dtype=float)

```

`:context:` might be quite leaky and we've simply been getting lucky based on the number of threads to have `numpy` imported somewhere. This is why it's been getting less "stable" when increasing the number of threads or simply by adding additional plotting sections; it's been splitting up the context more to such a point that `numpy` no longer gets imported.

I'd like to resolve the `:context:` issue in a followup, but for now the build works with 8 threads.

pyvista/plotting/plotter.py
```diff
@@ -7019,6 +7029,7 @@ def show(  # noqa: PLR0917
         return_viewer: bool = False,  # noqa: FBT001, FBT002
         return_cpos: bool | None = None,  # noqa: FBT001
         before_close_callback: Callable[[Plotter], None] | None = None,
+        store_image_depth: bool = False,  # noqa: FBT001, FBT002
```
user27182, line 7034: Was this always True before? Does this mean it's a breaking change if users were relying on get_image_depth to work without this keyword?

pyvista/plotting/plotter.py
```diff
@@ -7117,6 +7128,10 @@ def show(  # noqa: PLR0917
                 def fun(plotter):
                     plotter.screenshot('file.png')
 
+        store_image_depth : bool, default: False
```
user27182, line 7143: Should this be part of the theme config?

pyvista/plotting/plotter.py
```diff
@@ -7305,7 +7320,8 @@ def fun(plotter):
                 self.last_image = self.screenshot(filename, return_img=True)
             else:
                 self.last_image = self.screenshot(screenshot, return_img=True)
-            self.last_image_depth = self.get_image_depth()
+            if store_image_depth:
+                self.last_image_depth = self.get_image_depth()
```
user27182, line 7336: Is this previously initialized to None? If so, maybe we should also check if this is None in the getter and let users know that store_image_depth=True is required (print warning or maybe even raise an error)

user27182 (commented):
Seems like this could be a breaking change? Do we need to handle this more gracefully? Or maybe just add the breaking change label and document it?

pyvista/plotting/plotter.py
akaszynski: Maybe? Honestly, I don't think it's used often enough to warrant adding it to the theme config.

akaszynski:
> Seems like this could be a breaking change? Do we need to handle this more gracefully? Or maybe just add the breaking change label and document it?

It's a breaking change. I don't think image depth is used all that often. Even for this repository only one example used it. Still, if we've broken something critical, I'm totally fine yanking the release. Rendering speed and doc build stability seems worth it. Even so, we at least indicate to the user which parameter to add to `show` to enable storing the last image depth.

I've added a note in 5edc0d3813e5983c0e5dbc1aee6a66b1d2b9781d to indicate that we've changed the default behavior.

pyvista/plotting/plotter.py
akaszynski: It's a new kwarg, so I'm not concerned about the `None` case.

pyvista/plotting/plotter.py
akaszynski: Correct, and It will break some workflows I expect. We do, somewhat helpfully I hope, tell which kwarg to use when `get_image_depth` is called without setting this kwarg to `True`.

pyvista/plotting/plotter.py
```diff
@@ -7035,6 +7048,16 @@ def show(  # noqa: PLR0917
     ):
         """Display the plotting window.
 
+        .. versionchanged:: 0.47
+            The last image depth is no longer autoatically stored. You must
+            enable ``store_image_depth=True`` within :meth:`Plotter.show` to
```
user27182:
```suggestion
            The last image depth is no longer automatically stored. You must
            enable ``store_image_depth=True`` within :meth:`show` to
```

pyvista/plotting/plotter.py
```diff
@@ -7035,6 +7048,16 @@ def show(  # noqa: PLR0917
...
+            obtain the image depth after the :class:`pyvista.Plotter` has been
+            closed.
+
+        .. versionadded:: 0.47
+           Added the ``store_image_depth`` parameter to allow storing image depth
+           for use with :meth:`Plotter.get_image_depth`.
```
user27182:
```suggestion
           for use with :meth:`get_image_depth`.
```

pyvista/plotting/plotter.py
```diff
@@ -7117,6 +7140,10 @@ def show(  # noqa: PLR0917
                 def fun(plotter):
                     plotter.screenshot('file.png')
 
+        store_image_depth : bool, default: False
+            Store image depth before closing. Necessary when calling
+            :meth:`Plotter.get_image_depth` after rendering.
```
user27182:
```suggestion
            :meth:`get_image_depth` after rendering.
```

user27182 (commented):
Not sure if the links will resolve, might need full links (with `pyvista` prefix), or just the method name will work when referencing within the class

user27182:
> > Seems like this could be a breaking change? Do we need to handle this more gracefully? Or maybe just add the breaking change label and document it?
> 
> 
> 
> It's a breaking change. I don't think image depth is used all that often. Even for this repository only one example used it. Still, if we've broken something critical, I'm totally fine yanking the release. Rendering speed and doc build stability seems worth it. Even so, we at least indicate to the user which parameter to add to `show` to enable storing the last image depth.
> 
> 
> 
> I've added a note in 5edc0d3813e5983c0e5dbc1aee6a66b1d2b9781d to indicate that we've changed the default behavior.

Good enough for me, thanks!

pyvista/plotting/plotter.py
akaszynski: Where's the add suggestion to batch?!

Image

akaszynski: Seems it's part of the new "experience":

sc016

akaszynski:
> Not sure if the links will resolve, might need full links (with `pyvista` prefix), or just the method name will work when referencing within the class

We'll find out. For now I've applied the suggestions and we can verify the links in the doc build. Plus, we get to test build stability.

user27182 (approved):

akaszynski:
Links work:

sc017

Issue #7956, opened 2025-09-16
Timer class triggering callback incorrectly Bug


DominicBen:
### Describe the bug, what's wrong, and what you expected.

Timer() class has an issue where the callback is called max_steps on the first timer event, rather than each timer step. simple fix is to replace. This causes duration to be ignore in  pl.add_timer_event(max_steps=steps, duration=1000, callback=callback)

```python
    def execute(self, obj, _event):  # pragma: no cover # numpydoc ignore=PR01,RT01
        """Execute Timer."""
        # https://github.com/pyvista/pyvista/pull/5618
        iren = obj

        while self.step < self.max_steps:
            self.callback(self.step)
            iren.GetRenderWindow().Render()
            self.step += 1
        if self.id:
            iren.DestroyTimer(self.id)
```
With
```python
    def execute(self, obj, _event):  # pragma: no cover # numpydoc ignore=PR01,RT01
        """Execute Timer."""
        # https://github.com/pyvista/pyvista/pull/5618
        iren = obj

        if self.step < self.max_steps:
            self.callback(self.step)
            iren.GetRenderWindow().Render()
            self.step += 1
        elif self.id:
            iren.DestroyTimer(self.id)
```

### Steps to reproduce the bug.

```python
import pyvista as pv
import time

# Create a sphere mesh and a plotter
sphere = pv.Sphere()
pl = pv.Plotter()
actor = pl.add_mesh(sphere)

# Define constants for the animation
MAX_STEPS = 2000  # Total number of animation steps

# Variables to track time and animation state
current_step = 0
last_update_time = 0


# Define the callback function
def callback(step):
    print(step)
    actor.position = [step / (100.0 * MAX_STEPS), step / (100.0 * MAX_STEPS), 0]




pl.iren.initialize()
pl.add_timer_event(max_steps=MAX_STEPS, duration=1000, callback=callback)

# Run the plotter
pl.show()

```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Tue Sep 16 10:58:26 2025 Eastern Daylight Time

                OS : Windows 
            CPU(s) : 
           Machine : AMD64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : Intel
      GPU Renderer : Intel(R) Graphics
       GPU Version : 4.5.0 
     Render Window : vtkWin32OpenGLRenderWindow
  MathText Support : True

[...]

PR #7958, opened 2025-09-16
Fixed timer duration only activating on initial start

DominicBen:
### Overview
Fix https://github.com/pyvista/pyvista/issues/7956

### DIFF
--- a/pyvista/plotting/render_window_interactor.py
+++ b/pyvista/plotting/render_window_interactor.py
@@ -56,11 +56,11 @@ class Timer(_NoNewAttrMixin):
         # https://github.com/pyvista/pyvista/pull/5618
         iren = obj
 
-        while self.step < self.max_steps:
+        if self.step < self.max_steps:
             self.callback(self.step)
             iren.GetRenderWindow().Render()
             self.step += 1
-        if self.id:
+        elif self.id:
             iren.DestroyTimer(self.id)
 
 


user27182:
Thanks for the PR! Is there a way to test for this? E.g. check the actor's position has updated properly after the callback using the example code from https://github.com/pyvista/pyvista/issues/7956 ?

And I am not familiar with the timers so would like another maintainer to review this as well.

MatthewFlamm:
I'm not familiar with the timers either, but I can confirm that this fix adds in a delay for each call to the callback rather than just at the beginning.  I do think that the documentation could be more helpful in what to expect here

> Thanks for the PR! Is there a way to test for this? E.g. check the actor's position has updated properly after the callback using the example code from https://github.com/pyvista/pyvista/issues/7956 ?

We already have a test that tests the callback action, but it doesn't really confirm the actor has moved.  We should add to the image cache and make it clear that the actor moves, e.g. a before and after with a stationary actor.

https://github.com/pyvista/pyvista/blob/c556a05d10074ecae199166ec3c3b411fb83b078/tests/plotting/test_render_window_interactor.py#L256-L268

To test the duration, I quickly tried this, which works using an interactive plotter, but it doesn't work inside pytest, presumably due to off screen plotting?
```py
def test_add_timer_event():
    sphere = pv.Sphere()

    pl = pv.Plotter()
    actor = pl.add_mesh(sphere)

    def callback(step):
        actor.position = [step / 100.0, step / 100.0, 0]

    pl.iren.initialize()
    pl.add_timer_event(max_steps=200, duration=50, callback=callback)

    cpos = [(0.0, 0.0, 10.0), (0.0, 0.0, 0.0), (0.0, 1.0, 0.0)]
    start_time = time.time()
    pl.show(cpos=cpos)
    end_time = time.time()
    duration = end_time - start_time
    assert duration >= 200*0.05 * 0.9  # for some wiggle room
```

MatthewFlamm:
Having to use `pl.iren.initialize()` before creating timers seems undocumented, but it looks to be required by vtk

https://vtk.org/doc/nightly/html/classvtkRenderWindowInteractor.html#a7ef06e3e7c8cff5767c3b0d328934f55

We should document this in `create_timer`, and/or warn the user otherwise.

Sorry for going on some tangents here, but this issue and fix has uncovered other related things that need to be addressed.  From my perspective, this PR is good to go pending on whether we can figure out tests, and pending any eyes that are more knowledgeable about the timers.

MatthewFlamm:
This example from vtk confirms that the timer code should work like this PR:

https://examples.vtk.org/site/Cxx/Utilities/Timer/

edabor:
> Having to use `pl.iren.initialize()` before creating timers seems undocumented, but it looks to be required by vtk
> 
> https://vtk.org/doc/nightly/html/classvtkRenderWindowInteractor.html#a7ef06e3e7c8cff5767c3b0d328934f55
> 
> We should document this in `create_timer`, and/or warn the user otherwise.

+1 for both user warning **and** documentation, maybe inside another PR ?

MatthewFlamm (approved):
I'm approving this because it looks right, tests to work interactively, and I cannot think of a way to adequately test it.

MatthewFlamm:
The Timer functionality was added by @tkoyama010 in #4839, so maybe they can take a look at these changes before merging.

tkoyama010 (approved):
LGTM. Thanks!

PR #7964, opened 2025-09-19
Add `border_mode` to `resample` filter

user27182:
### Overview

Partially resolve #7960. A custom `constant` mode is _not_ implemented in this PR.

PR #7965, opened 2025-09-19
Add `bspline` interpolator to `resample` filter

user27182:
### Overview

See https://vtk.org/doc/nightly/html/classvtkImageBSplineInterpolator.html#details

### DIFF
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -3727,7 +3727,7 @@ class ImageDataFilters(DataSetFilters):
         self: ImageData,
         sample_rate: float | VectorLike[float] | None = None,
         interpolation: Literal[
-            'nearest', 'linear', 'cubic', 'lanczos', 'hamming', 'blackman'
+            'nearest', 'linear', 'cubic', 'bspline', 'lanczos', 'hamming', 'blackman'
         ] = 'nearest',
         *,
         reference_image: ImageData | None = None,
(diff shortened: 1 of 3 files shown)


akaszynski (dismissed):

pyvista/core/filters/image_data.py
```diff
@@ -4226,6 +4237,15 @@ def resample(  # type: ignore[misc]
         elif interpolation == 'cubic':
             interpolator = _vtk.vtkImageInterpolator()
             interpolator.SetInterpolationModeToCubic()
+        elif interpolation == 'bspline':
+            interpolator = _vtk.vtkImageBSplineInterpolator()
+            # Need to pre-compute coefficients
+            coefficients = _vtk.vtkImageBSplineCoefficients()
```
user27182: Pretty sure we also need to set the border mode for computing the coefficients. Will need to update this after #7964 is merged.

user27182:
Marking as draft since there will be merge conflicts after #7964 merges, and it's better to use `clim` in the examples (added in https://github.com/pyvista/pyvista/pull/7965/commits/6679d15c34e9cd00f996d1dca266d3eed5a9613d) to compare apples to apples

tests/doc/doc_image_cache/pyvista-ImageDataFilters-resample-16d3d78a2b56cdb7_03_00.jpg
user27182, line 1: Changing the clim revealed that this image now appears too dark. This is a VTK bug: https://gitlab.kitware.com/vtk/vtk/-/issues/19788

Hopefully it will be fixed for the next VTK release. I don't think we necessarily need to add comments or a warning here about this.

akaszynski (approved):
No issues with this. Thanks for adding an example.

PR #7968, opened 2025-09-24
Move documentation testing to separate dedicated job

user27182:
Testing is separate from building so should be a separate job. This is also preparation for updates to `pytest-pyvista` which will enable vtksz file testing. This will add several minutes of test time and would otherwise delay the html upload if kept inside the build job

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -117,6 +117,7 @@ jobs:
     name: Test Documentation
     runs-on: ubuntu-22.04
     needs: doc
+    if: ${{ !cancelled() }}
     env:
       TOX_COLORED: "yes"
     steps:


akaszynski (commented):
LGTM. Should be kept on GitHub hosted runners due to the artifact size vs. the length of the CI job.

akaszynski (approved):

user27182:
> Should be kept on GitHub hosted runners due to the artifact size vs. the length of the CI job.

Moreover, for the vtksz tests, installing `playwright` on GitHub hosted runners is straightforward (`playwright install`). But on self-hosted, a bunch of extra steps and libraries are needed, which took ~6 mins to download/install.

https://github.com/pyvista/pyvista/blob/1e53b96e040837002b2888802d4ede8b23fe4435/.github/workflows/docs.yml#L49-L63

We do lose the ability to use 4 cores instead of 2 for preprocessing and test executing, but that may only be the difference between 8mins and 10mins for a full test run.

akaszynski:
> Moreover, for the vtksz tests, installing `playwright` on GitHub hosted runners is straightforward (`playwright install`). But on self-hosted, a bunch of extra steps and libraries are needed, which took ~6 mins to download/install.

This can be added to the self-hosted docker image, but I'm still of the opinion that we can use GitHub hosted runners unless we end up saturating those runners.

PR #7969, opened 2025-09-25
Fix argument mismatch in picking documentation

beroda:
### Overview

The `start` argument in some picking method documentation mismatches the real implementation.

### DIFF
--- a/pyvista/plotting/picking.py
+++ b/pyvista/plotting/picking.py
@@ -592,7 +592,7 @@ class PickingInterface:  # numpydoc ignore=PR01
         font_size : int, default: 18
             Sets the font size of the message.
 
-        start : bool, default: True
+        start : bool, default: False
             Automatically start the cell selection tool.
 
         show_frustum : bool, default: False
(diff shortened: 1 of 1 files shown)


edabor:
Note that other places in the code base could exhibit such out-of-sync between documentation and implementation. I have not checked everywhere.

~We could use [`pydoclint`](https://jsh9.github.io/pydoclint/) to catch such errors (and many more) but unfortunately, it does not check for default values https://jsh9.github.io/pydoclint/#273-pitfall-default-values-of-arguments and outputs an error if one is found, which is the case in the code-base https://github.com/jsh9/pydoclint/issues/179~

We could use [`pydoclint`](https://jsh9.github.io/pydoclint/) but it would mean to rewrite all our defaults from `default: value` to `default=value`
https://jsh9.github.io/pydoclint/style_deviations.html#12-default-values
Might worth-it 🤔

user27182:
> Note that other places in the code base could exhibit such out-of-sync between documentation and implementation. I have not checked everywhere.

This is indeed a broader issue, see also #5665 for a solution which replaces types in docs with the actual implementation though realistically I don't think we'll get there.

> We could use [`pydoclint`](https://jsh9.github.io/pydoclint/) but it would mean to rewrite all our defaults from `default: value` to `default=value`
> 
> https://jsh9.github.io/pydoclint/style_deviations.html#12-default-values
> 
> Might worth-it 🤔 

Haven't looked at this package but if it helps resolve the out of sync issue then it seems worth it, especially if it can be addressed with a simple replace all.

akaszynski (approved):

PR #7970, opened 2025-09-27
Add `plot` command to pyvista CLI

user27182:
### Overview

Enable `pyvista plot <mesh_file>` to quickly plot a mesh from file.

Depends on #7463

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -56,7 +56,7 @@ jobs:
 
       - name: "Report Failure"
         if: steps.lychee.outputs.exit_code != 0
-        uses: peter-evans/create-issue-from-file@e8ef132d6df98ed982188e460ebb3b5d4ef3a9cd
+        uses: peter-evans/create-issue-from-file@fca9117c27cdc29c6c4db3b86c48e4115a786710
         with:
           title: "chore: linkcheck failure"
           content-filepath: ./lychee/out.md
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,13 +85,13 @@ test = [
   'pandas<=2.4.0',
   'pyanalyze<=0.13.1',
   'pyobjc-framework-Cocoa; sys_platform == "darwin"', # see pyvista/pulls/7828
-  'pytest-cov<6.3.0',
+  'pytest-cov<7.1.0',
   'pytest-mypy-plugins<3.3.0',
   'pytest-pyvista==0.2.0',
   'pytest-xdist<3.9.0',
   'pytest<8.5.0',
   'pytest_cases<3.10.0',
-  'pytest_mock<3.15.0',
+  'pytest_mock<3.16.0',
   'retry-requests<2.1.0',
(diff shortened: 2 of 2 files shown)


tests/plotting/test_plotter.py
```diff
@@ -107,21 +107,21 @@ def test_plotter_add_mesh_multiblock_algo_raises(mocker: MockerFixture):
     from pyvista.plotting import plotter
 
     m = mocker.patch.object(plotter, 'algorithm_to_mesh_handler')
-    m.return_value = pv.MultiBlock(), 'foo'
+    m.return_value = pv.MultiBlock(), {}
```
tkoyama010, line 110: I wasn’t quite sure about the intent behind this change. Would you mind clarifying it for me?

tkoyama010 (changes requested):

tests/plotting/test_plotter.py
user27182: The test is checking an error is raised for invalid input. A string was used, which previously worked because strings _were_ invalid. But now, a string is valid input, since a file path is accepted. But, the test still fails because 'foo' isn't a valid file (and the failure happens inside `read`, not `add_mesh`)


So this change is needed because a dict is now used as the invalid input

pyvista/__main__.py
```diff
@@ -15,21 +15,43 @@
...
     'report': pyvista.Report,
+    'plot': pyvista.plot,
+}
+COMMANDS_DISPLAY = {
+    'report': 'pyvista.Report()',
+    'plot': 'pyvista.plot()',
+}
+COMMANDS_URL = {
+    'report': 'https://docs.pyvista.org/api/utilities/_autosummary/pyvista.report',
+    'plot': 'https://dev.pyvista.org/api/plotting/_autosummary/pyvista.plot.html',
```
user27182:
```suggestion
    'plot': 'https://docs.pyvista.org/api/plotting/_autosummary/pyvista.plot.html',
```

tests/plotting/test_plotter.py
user27182: If preferred, I can create a separate PR to isolate the feature of supporting file inputs with `add_mesh`

tkoyama010: Thanks! Let's do it that way so it's easier for people who see it later.

user27182: See #7976

edabor:
@user27182 Do you mind if I try using the `cyclopts` CLI library inside a PR targeting the current branch ? It provides great features and greatly simplifies the code.

https://cyclopts.readthedocs.io/en/latest/index.html

user27182:
> @user27182 Do you mind if I try using the `cyclopts` CLI library inside a PR targeting the current branch ? It provides great features and greatly simplifies the code.
> 
> 
> 
> https://cyclopts.readthedocs.io/en/latest/index.html

Sure go ahead, could even target main directly as a full alternative to this PR

PR #7971, opened 2025-09-27
Include report with integration tests

user27182:
Include software report. Some key test dependencies might be missing from the report but it should still be helpful.


akaszynski (approved):

akaszynski:
Doc build https://github.com/pyvista/pyvista/actions/runs/18053070029/job/51378390080 failed due to https://github.com/pyvista/pyvista/pull/7953.

PR #7973, opened 2025-09-27
Separate tox env for building and testing docs

user27182:
### Overview

Follow-up to #7968 to split up the envs

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -89,7 +89,7 @@ all_testing_cmdline =
     !cov-!plotting-!core: pytest {[testenv]all_flags}  {posargs}
 
 software_report_cmdline=
-    python -c "import pyvista;print(pyvista.Report(gpu=True, downloads=True))"
+    pyvista report gpu=True downloads=True
 
 # ================= INTEGRATION ENVIRONMENTS =================
 [testenv:integration-{trame,geovista,mne,pyvistaqt}]
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -34,12 +34,21 @@ body:
       label: System Information
       placeholder: The log will automatically be formatted! No need to type backticks.
       description: |
-        Please run the following code wherever you are experiencing the bug, and paste the output below.
+        Please run the following wherever you are experiencing the bug, and paste the output below.
         This report helps us track down bugs and will be critical to addressing your bug.
+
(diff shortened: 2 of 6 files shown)


tkoyama010 (approved):
LGTM

tox.ini
```diff
@@ -220,21 +220,17 @@ commands =
...
-[testenv:docs-{build,test}]
-description =
-    build: Building documentation
-    test: Testing documentation build (needs already built documentation)
+[testenv:docs-build]
+description = Building documentation
 basepython = py3.13
 dependency_groups =
     build: docs
-    test: test
```
edabor: You can remove the `build` factor here

tox.ini
```diff
@@ -245,8 +241,17 @@ setenv =
...
+    python -c "import subprocess, os;subprocess.run([r'{envpython}','make_tables.py'], cwd='{env:_SPHINX_SOURCE_DIR}', check=True)"
+    python -c "import subprocess, os;subprocess.run([r'{envpython}','make_external_gallery.py'], cwd='{env:_SPHINX_SOURCE_DIR}', check=True)"
+commands =
+    python -m sphinx -M {posargs:html} {env:_SPHINX_SOURCE_DIR} {env:_SPHINX_BUILD_DIR} {env:_SPHINX_DEFAULTS} {env:_SPHINX_JOBS}
+
+[testenv:docs-test]
+description = Test documentation build (needs already built docs)
+basepython = py3.13
+dependency_groups =
+    test: test-docs
```
edabor: The `test` factor here can be removed

beroda (changes requested):
Some factors can be removed. Otherwise LGTM

tox.ini
```diff
@@ -246,8 +242,17 @@ setenv =
 change_dir = doc
 commands_pre =
     {[testenv]software_report_cmdline}
```
user27182, line 243: This technically works for both build and test, but is a bit confusing because it seems like magic that this would affect the test env even though it's listed in the build env. 

Should we make this explicit? 

``` suggestion
    software_report_cmdline
```

And also include 
```
commands_pre =
    software_report_cmdline
```
explicitly in the test env?

tkoyama010 (approved):

akaszynski (approved):
LGTM. Thanks @beroda for the edits.

I'll be out for a week hiking with limited internet access and will be delayed reviewing during that time.

Issue #5870, opened 2024-04-05
Allow docs preview to be generated even if docs build fails


user27182:
### Describe what maintenance you would like added.

Previewing the docs with `@pyvista-bot preview` is great, but a preview is currently only generated if the build succeeds. However, even if warnings or errors are generated by the build (and the build "fails"), the docs are usually still in a viewable state, but can't be viewed simply because the failure causes the bot not to generated the preview.

Further, with build times taking 2hrs recently (see #5868), it can be painful to wait 2hrs for a preview only for it not to be generated because of a simple rst formatting warning.

As a workaround, draft PRs can temporarily remove the ` -W --keep-going ` option from the makefile:
https://github.com/pyvista/pyvista/blob/a36f649c1a9ea9274939089d8d726ae620bef8b6/doc/Makefile#L8

in order to allow the build to succeed with warnings, thus allowing the bot to generate the preview. But, it would be nice not to have to use a workaround like this.

PR #7978, opened 2025-09-30
Add annotations to `plot`

edabor:
### Overview

Add annotations of parameters to `plot`


pyvista/_plot.py
```diff
@@ -8,45 +8,54 @@
 
 """
 
+# ruff: noqa: FBT001,FBT002
```
tkoyama010: To make it easier to identify which parameter is causing the error, please set it to asis.

tkoyama010 (changes requested):

pyvista/_plot.py
```diff
@@ -77,10 +85,10 @@ def plot(  # noqa: PLR0917
     interactive : bool, default: :attr:`pyvista.plotting.themes.Theme.interactive`
         Allows user to pan and move figure.
 
-    cpos : list, optional
+    cpos : CameraPositionOptions, optional
```
user27182, line 97: CameraPostionOptions isn't documented, so I don't think this hint will be too helpful as-is. Should we add it to the typing docs? Or maybe expand the description to indicate that strings like 'xy' are acceptable?

edabor: Added in the documentation along with `JupyterBackendOptions` in ~fdf7f403757471b79113e39af02387b98cae5690~ https://github.com/pyvista/pyvista/pull/7978/commits/5c448c76a9de3f03075b850ceb14fb41424c8bce

doc/source/conf.py
```diff
@@ -242,12 +244,10 @@
     (r'py:.*', 'BackfaceArgs'),
     (r'py:.*', 'CullingOptions'),
     (r'py:.*', 'OpacityOptions'),
-    (r'py:.*', 'CameraPositionOptions'),
```
user27182, line 245: These need to stay in the ignore.

Ref errors are generated for both refs in the docstring and refs in the signature. If you register it as an alias, it will fix the ref in the docstring but the ref in the signature will remain broken. This is a known sphinx bug.

We'll need to manually confirm that the ref inside the docstring links to the typing docs correctly.

edabor: Thanks for the help ! I started to struggle 😅

user27182: Separate issue, but we could add the "if not cancelled" condition to the docs preview (and resolve #5870) so that we can still check the links in the preview even if the build fails due to bad refs

edabor: That would be very helpful indeed. Maybe we could allow a draft PR to preview the documentation even if it fails building ? And when ready to be reviewed, preview should only be published if the build is successful.

doc/source/api/core/typing.rst
```diff
@@ -82,6 +82,20 @@ pyvista.InteractionEventType
 
 .. autodata:: InteractionEventType
 
+pyvista.CameraPositionOptions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
edabor:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

doc/source/api/core/typing.rst
```diff
@@ -82,6 +82,20 @@ pyvista.InteractionEventType
...
+
+.. currentmodule:: pyvista
+
+.. autodata:: CameraPositionOptions
+
+pyvista.JupyterBackendOptions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
edabor:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

user27182 (dismissed):

pyvista/_plot.py
```diff
@@ -11,42 +11,56 @@
...
+if TYPE_CHECKING:
+    from pyvista.core._typing_core._array_like import NumpyArray
+    from pyvista.jupyter import JupyterBackendOptions
+    from pyvista.plotting._typing import CameraPositionOptions
+    from pyvista.plotting._typing import ColorLike
+    from pyvista.plotting.themes import Theme
+
+    PlottableType: TypeAlias = (
+        NumpyArray[float] | pyvista.DataSet | pyvista.MultiBlock | str | Path
+    )
```
user27182: My understanding is that the first argument to `plot` has the same input type(s) as `add_mesh`. So `PlottableType` should probably be defined in plotter.py and used for `add_mesh` and then imported and reused here.

edabor: Done in 132fc0ae11f2c71a56bf584deca8e6

edabor: Maybe we should document `PlottableType` ? It's a rather frequent entry point of pyvista plotting API

tkoyama010 (dismissed):
LGTM

pyvista/_plot.py
user27182: Yes, it's a good idea to document these more complex and frequently used types. 

The docs for var_item in add mesh states any wrappable type is also accepted . This includes Trimesh and Meshio meshes and 3D arrays (ie not just 2D matrix-like arrays). 

So there should probably also be a "WrappableType" for pv.wraps, and the typing for PlottableType should include "WrappableType" in the union, and also include ArrayLike somewhere as a more general array type replacement. 

This can all be done in a separate PR though.

beroda:
@tkoyama010 waiting for https://github.com/pyvista/pyvista/pull/7978#discussion_r2408749569 to be addressed

pyvista/_plot.py
akaszynski: Let's do this in a separate PR.

akaszynski: Opened a new issue for this in #7995.

Issue #7873, opened 2025-08-26
Interactive subplots broken in documentation


edabor:
### Describe the bug, what's wrong, and what you expected.

Some interactive subplots are broken in the documentation (non exhaustive list):

- https://docs.pyvista.org/examples/01-filter/gradients.html

<img width="910" height="544" alt="Image" src="https://github.com/user-attachments/assets/7549b03c-65d8-43d0-83dd-5d64939a8359" />

- https://docs.pyvista.org/examples/02-plot/depth_peeling.html

<img width="938" height="559" alt="Image" src="https://github.com/user-attachments/assets/b8ff85b3-f9e7-4fe2-bb3b-6ba2e631d082" />

Tested on both Firefox and Chrome.

### Steps to reproduce the bug.

Go to documentation

### System Information

```shell
None
```

beroda:
Managed to get a pure `trame` + `vtk` MWE that demonstrates the issue. The example runs fine locally, the only issue occurs when loading the exported scene file `scene.vtksz` into https://kitware.github.io/vtk-js/examples/OfflineLocalView.html

@jourdain have you seen that before ?

```python
from __future__ import annotations

from pathlib import Path

from trame.app import get_server
from trame.ui.vuetify import SinglePageLayout
from trame.widgets import vtk as vtk_widgets
from trame.widgets import vuetify
from vtkmodules.vtkFiltersSources import vtkConeSource
from vtkmodules.vtkFiltersSources import vtkCubeSource
from vtkmodules.vtkFiltersSources import vtkCylinderSource
from vtkmodules.vtkFiltersSources import vtkSphereSource
from vtkmodules.vtkRenderingCore import vtkActor
from vtkmodules.vtkRenderingCore import vtkPolyDataMapper
from vtkmodules.vtkRenderingCore import vtkRenderer
from vtkmodules.vtkRenderingCore import vtkRenderWindow
from vtkmodules.vtkRenderingCore import vtkRenderWindowInteractor


def build_render_window() -> vtkRenderWindow:
    rw = vtkRenderWindow()
    iren = vtkRenderWindowInteractor()
    iren.SetRenderWindow(rw)

    # Define viewport ranges.
    xmins = [0, 0.5, 0, 0.5]
    xmaxs = [0.5, 1, 0.5, 1]
    ymins = [0, 0, 0.5, 0.5]
    ymaxs = [0.5, 0.5, 1, 1]

[...]

Issue #7721, opened 2025-07-16
trame-vtk behaviour regression for subplot within interactive scene


bjlittle:
### Describe the bug, what's wrong, and what you expected.

Playing forward notification of a regression in behaviour for `trame-vtk` v2.9.0 which I identified over on `geovista`.

For further details please reference https://github.com/bjlittle/geovista/pull/1561.

This was only uncovered by chance when browsing the documentation interactive examples, as we don't have any GHA CI visual test coverage of interactive `trame` scenes 😬 ... not quite sure that's even possible? 🤷

Isolated the issue down to `trame-vtk`. I suspect this regression will also be applicable to `pyvista` too, although I've not had time to confirm that, sorry.

@tkoyama010 asked me to raise this issue for awareness 👍 

Summary of `trame-vtk` regression:

| Version | Release Data | Status |
| ------- | ------------ | ------ |
| 2.9.0 | 23 Jun '25 | Regression ❌ |
| 2.8.17 | 03 Jun '25 | Regression ❌ |
| 2.8.16 | 02 Jun '25 | Broken due to unbundled `static_viewer.html` ❌ |
| 2.8.15 | 09 Feb '25 | Functional ✅ |

As a result I've pinned `geovista` to`trame-vtk` 2.8.15, see https://github.com/bjlittle/geovista/pull/1562.

### Steps to reproduce the bug.

See https://github.com/bjlittle/geovista/pull/1561.
 
In particular this RTD build of the `sphinx-gallery` [Panel Extraction](https://geovista--1553.org.readthedocs.build/en/1553/generated/gallery/extraction/panel_manifold.html#sphx-glr-generated-gallery-extraction-panel-manifold-py) `geovista` example "**Interactive Scene**" demonstrates this behaviour regression for subplots.

### System Information

```shell
----------------------------------------------------------------------------------------
  Date: Wed Jul 16 13:57:59 2025 BST

                  OS : Linux (Red Hat Enterprise Linux 9.6)
              CPU(s) : 4
             Machine : x86_64
        Architecture : 64bit
                 RAM : 15.3 GiB
         Environment : Python
         File system : xfs
          GPU Vendor : Mesa
        GPU Renderer : llvmpipe (LLVM 19.1.7, 256 bits)
         GPU Version : 4.5 (Core Profile) Mesa 24.2.8

  Python 3.12.11 | packaged by conda-forge | (main, Jun  4 2025, 14:45:31) [GCC
  13.3.0]

[...]

PR #7878, opened 2025-08-28
Move doc image tests to `pytest-pyvista`

user27182:
### Overview

Use feature from https://github.com/pyvista/pytest-pyvista/pull/198 to do the documentation image testing.

PR #7979, opened 2025-09-30
Pin `trame-vtk` for documentation building

beroda:
### Overview

Fix https://github.com/pyvista/pyvista/issues/7721
Fix https://github.com/pyvista/pyvista/issues/7873

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -144,7 +144,7 @@ docs = [
   'sphinxcontrib-websupport==2.0.0',
   'sphinxext-opengraph==0.13.0',
   'sympy==1.14.0',
-  'trame-vtk==2.9.1',
+  'trame-vtk==2.8.15', # Do not upgrade, see https://github.com/pyvista/pyvista/issues/7721
   'trame-vuetify==3.0.2',
   'trame==3.12.0',
   'trimesh==4.8.0',


user27182:
Is it worth it to wait for a new `pytest-pyvista` release so we can see a diff of the changes? 

It would roughly go like this:

- new pytest-pyvista release
- merge pyvista/pyvista#7878
- create separate PR where `--include_vtksz` option is enabled to set a baseline for the images
- merge this PR with diff showing the fix

edabor:
> Is it worth it to wait for a new `pytest-pyvista` release so we can see a diff of the changes?
> 
> It would roughly go like this:
> 
>     * new pytest-pyvista release
> 
>     * merge [Move doc image tests to `pytest-pyvista` pyvista/pyvista#7878](https://github.com/pyvista/pyvista/pull/7878)
> 
>     * create separate PR where `--include_vtksz` option is enabled to set a baseline for the images
> 
>     * merge this PR with diff showing the fix

Totally agree.

edabor:
The subplots with opacity are fixed https://68dc5ce016aabd0ad6ab52a8--pyvista-dev.netlify.app/examples/01-filter/gradients.html

user27182:
Waiting on pyvista/pyvista#8008

edabor:
@user27182 Don't know what's happening here but the doc tests are unexpectedly passing, even though the gradients interactive example [here](https://68dc5ce016aabd0ad6ab52a8--pyvista-dev.netlify.app/examples/01-filter/gradients.html) are different to the static image in `doc_generated_test_images` 

Any idea why ?

sphx_glr_gradients_001_vtksz

user27182:
Not sure, will need to investigate.

user27182:
I think the issue that we need to also pin trame for the tests, as part of the `docs-test` dependency group

pyproject.toml
```diff
@@ -145,15 +145,19 @@ docs = [
...
   'trimesh==4.8.3',
   'vtk-xref==0.1.1',
   { include-group = 'pinned' },
 ]
 
-docs-test = ['pytest-pyvista[vtksz]==0.3.1', 'pytest-xdist<3.9.0']
+docs-test = [
+  'pytest-pyvista[vtksz]==0.3.1',
+  'pytest-xdist<3.9.0',
+  { include-group = 'docs' },
```
user27182: This seems a bit heavy to include _all_ docs dependencies, when we really only need to pin `trame-vtk`, and none of the other deps are used by pytest-pyvista. Can we create a new `docs-trame-vtk` group which only contains the single pinned `trame-vtk` dep (with comment), and include this group in the `docs` and `docs-test` groups?

edabor: Done in aaa2298652fa7c0812e7eaeb5206f8359d194ba6

edabor:
@user27182 Love this new `include_vtksz` feature, this went so smoothly in this PR 😍 
Our docs are going to be rock solid 🚀

PR #7991, opened 2025-10-04
Use `cyclopts` as CLI library

beroda:
### Overview

The [`cyclopts`](https://cyclopts.readthedocs.io/en/latest/index.html) package provides great CLI features out of the box.

Pros:
- get arguments from docstring
- convert CLI using type annotations
- rich terminal output
- flexible

Cons:
- supplementary dependency

See a visual comparison below:

| main                                                                                                                               | This PR                                                                                                                            |
|------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|
| image | image |
| image | image |
| image  | image  |
| image | image  |

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ classifiers = [
   'Topic :: Scientific/Engineering :: Information Analysis',
 ]
 dependencies = [
-  'cyclopts>3.24.0',
+  'cyclopts>=3.24.0',
   'matplotlib>=3.0.1',
   'numpy>=1.21.0',
   'pillow',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ classifiers = [
   'Topic :: Scientific/Engineering :: Information Analysis',
 ]
 dependencies = [
+  'cyclopts>3.24.0',
   'matplotlib>=3.0.1',
   'numpy>=1.21.0',
   'pillow',
--- a/pyvista/__main__.py
+++ b/pyvista/__main__.py
@@ -2,8 +2,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-from typing import Any
+from functools import wraps
 from typing import get_type_hints
 import warnings
 
@@ -13,29 +12,24 @@ import pyvista
 from pyvista import Report
 from pyvista.core.errors import PyVistaDeprecationWarning
 
-if TYPE_CHECKING:
-    from collections.abc import Callable
-
-
 # Assign annotations to be able to use the Report class using
 # cyclopts when __future__ annotations are enabled. See https://github.com/BrianPugh/cyclopts/issues/570
 Report.__init__.__annotations__ = get_type_hints(Report.__init__)
 
 
-COMMANDS: dict[str, Callable[..., Any]] = {
-    'report': Report,
-}
-
 # help format needs to be `plaintext` due to unsupported `versionadded` sphinx directive in rich.
 # Change to `rst` when cyclopts v4 is out. See https://github.com/BrianPugh/cyclopts/issues/568
 app = App(
+    name=pyvista.__name__,
     help_format='plaintext',
     version=f'{pyvista.__name__} {pyvista.__version__}',
 )
 
 
-for name, func in COMMANDS.items():
-    app.command(func, name=name)
+@app.command
+@wraps(Report)
+def report(*args, **kwargs):  # noqa: ANN201, D103
+    return Report(*args, **kwargs)
 
 
 def main(argv: list[str] | str | None = None) -> None:
--- a/pyvista/__main__.py
+++ b/pyvista/__main__.py
@@ -35,7 +35,6 @@ app = App(
 
 
 for name, func in COMMANDS.items():
-    # print(Report.__init__.__annotations__)
     app.command(func, name=name)
 
 
--- a/pyvista/__main__.py
+++ b/pyvista/__main__.py
@@ -2,93 +2,51 @@
 
 from __future__ import annotations
 
-import argparse
-import ast
-import sys
 from typing import TYPE_CHECKING
 from typing import Any
+from typing import get_type_hints
+import warnings
+
+from cyclopts import App
 
 import pyvista
+from pyvista import Report
+from pyvista.core.errors import PyVistaDeprecationWarning
 
 if TYPE_CHECKING:
     from collections.abc import Callable
 
+
+# Assign annotations to be able to use the Report class using
+# cyclopts when __future__ annotations are enabled. See https://github.com/BrianPugh/cyclopts/issues/570
+Report.__init__.__annotations__ = get_type_hints(Report.__init__)
+
+
 COMMANDS: dict[str, Callable[..., Any]] = {
-    'report': pyvista.Report,
+    'report': Report,
 }
-COMMANDS_DISPLAY = {'report': 'pyvista.Report()'}
-COMMANDS_URL = {'report': 'https://docs.pyvista.org/api/utilities/_autosummary/pyvista.report'}
+
+# help format needs to be `plaintext` due to unsupported `versionadded` sphinx directive in rich.
+# Change to `rst` when cyclopts v4 is out. See https://github.com/BrianPugh/cyclopts/issues/568
+app = App(
+    help_format='plaintext',
+    version=f"f'{pyvista.__name__} {pyvista.__version__}'",
+)
 
 
-def _parse_kwargs(args: list[str]) -> dict[str, Any]:
-    """Parse CLI args of form key=value into a dict.
-
-    Try literal_eval first, fallback to string.
-    """
-    kwargs = {}
-    for arg in args:
-        if '=' not in arg:
-            msg = f'Invalid kwarg format: {arg!r}, expected key=value'
-            raise ValueError(msg)
-        key, value = arg.split('=', 1)
-
-        # Convert bool-like strings into literal Python bools
-        lower = value.lower()
-        if lower in ['true', 'y', 'yes']:
-            value = 'True'
-        elif lower in ['false', 'n', 'no']:
-            value = 'False'
-
-        try:
-            # Try Python literal first
-            kwargs[key] = ast.literal_eval(value)
-        except (ValueError, SyntaxError):
-            # Fallback: treat as string
-            kwargs[key] = value
-    return kwargs
+for name, func in COMMANDS.items():
+    # print(Report.__init__.__annotations__)
+    app.command(func, name=name)
 
 
-def main(argv: list[str] | None = None) -> None:
+def main(argv: list[str] | str | None = None) -> None:
     """PyVista Command-Line Interface entry point."""
-    # Use sys.argv if no arguments were explicitly passed
-    if argv is None:
-        argv = sys.argv[1:]
-
-    # Create top-level parser with --version option
-    parser = argparse.ArgumentParser(prog='pyvista')
-    parser.add_argument(
-        '--version',
-        action='version',
-        version=f'pyvista {pyvista.__version__}',
-        help='show PyVista version and exit',
-    )
-
-    # Create a generic keyword subparser for each command
-    subparsers = parser.add_subparsers(dest='subcommand', required=True)
-    for name in COMMANDS:
-        url = COMMANDS_URL[name]
-        display = COMMANDS_DISPLAY[name]
-        subparser = subparsers.add_parser(
-            name,
-            help=f'run {display!r} with optional key=value kwargs',
-            usage='%(prog)s [key=value] ...',
-            epilog=f'See documentation for available keywords and more info:\n{url}',
-            formatter_class=argparse.RawDescriptionHelpFormatter,
-        )
-        subparser.add_argument(
-            'kwargs', nargs='*', help='optional keyword arguments in key=value form'
-        )
-
-    # Parse primary args
-    if not argv:
-        parser.print_help()
-        sys.exit(1)
-    args = parser.parse_args(argv)
-
-    # Parse remaining args as a Python kwargs dict and execute command as a function
-    kwargs = _parse_kwargs(getattr(args, 'kwargs', []))
-    func = COMMANDS[args.subcommand]
-    result = func(**kwargs)
+    # Ignore warnings emitted because arguments are passed positionally by the
+    # inspect module. See https://docs.python.org/3/library/inspect.html#inspect.BoundArguments.kwargs
+    # and https://github.com/BrianPugh/cyclopts/issues/567
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore', category=PyVistaDeprecationWarning)
+        result = app(tokens=argv)
 
     if result is not None:
         print(result)  # noqa: T201
--- a/pyvista/__main__.py
+++ b/pyvista/__main__.py
@@ -30,7 +30,7 @@ COMMANDS: dict[str, Callable[..., Any]] = {
 # Change to `rst` when cyclopts v4 is out. See https://github.com/BrianPugh/cyclopts/issues/568
 app = App(
     help_format='plaintext',
-    version=f"f'{pyvista.__name__} {pyvista.__version__}'",
+    version=f'{pyvista.__name__} {pyvista.__version__}',
 )
 
 
--- a/pyvista/report.py
+++ b/pyvista/report.py
@@ -190,7 +190,9 @@ class Report(scooby.Report):
 
         .. code-block:: shell
 
-            pyvista report [key=value] ...
+            pyvista report --sort ...
+
+        Run `pyvista report --help` for more details on available parameters.
 
 
     Parameters
--- a/pyvista/report.py
+++ b/pyvista/report.py
@@ -180,6 +180,7 @@ class GPUInfo:
         return content
 
 
+@_deprecate_positional_args
 class Report(scooby.Report):
     """Generate a PyVista software environment report.
 
@@ -256,7 +257,6 @@ class Report(scooby.Report):
 
     """
 
-    @_deprecate_positional_args
     def __init__(  # noqa: PLR0917
         self,
         additional=None,
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -4,12 +4,25 @@ import contextlib
 import io
 import subprocess
 import sys
+import textwrap
 
 import pytest
+from rich.console import Console
 
 import pyvista as pv
-from pyvista.__main__ import COMMANDS_URL
+from pyvista.__main__ import app
 from pyvista.__main__ import main
+
+
+@pytest.fixture
+def console():
+    return Console(
+        width=70,
+        force_terminal=True,
+        highlight=False,
+        color_system=None,
+        legacy_windows=False,
+    )
 
 
 def pop_second_line(text: str) -> str:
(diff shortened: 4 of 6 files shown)


pyproject.toml
```diff
@@ -17,6 +17,7 @@ classifiers = [
   'Topic :: Scientific/Engineering :: Information Analysis',
 ]
 dependencies = [
+  'cyclopts>=3.24.0',
```
user27182, line 20: Can we add this to Report so it's included in the printed report deps?

pyvista/report.py
```diff
@@ -189,7 +190,9 @@ class Report(scooby.Report):
 
         .. code-block:: shell
 
-            pyvista report [key=value] ...
+            pyvista report --sort ...
+
+        Run `pyvista report --help` for more details on available parameters.
```
user27182:
```suggestion
        Run ``pyvista report --help`` for more details on available parameters.
```

tests/test_cli.py
```diff
@@ -1,115 +1,155 @@
...
+    assert e.value.code == 1
+    assert expected == capsys.readouterr().out
+
+
+@pytest.fixture
+def mock_report(mocker: MockerFixture):
+    return mocker.patch.object(pv.__main__, 'Report')
+
+
+class Cases_Report:  # noqa: N801
```
user27182:
```suggestion
class CasesReport:
```

tox.ini
```diff
@@ -89,7 +89,7 @@ all_testing_cmdline =
     !cov-!plotting-!core: pytest {[testenv]all_flags}  {posargs}
 
 software_report_cmdline=
-    pyvista report gpu=True downloads=True
+    pyvista report --gpu=True --downloads
```
user27182: Is `--gpu` not also valid? Or `--downloads=True` ? Why mix the two syntaxes here?

tests/test_cli.py
```diff
@@ -1,115 +1,155 @@
...
 
-def pop_second_line(text: str) -> str:
-    """Return text with the second line removed."""
-    lines = text.splitlines()
-    if len(lines) > 1:
-        lines.pop(1)
-    return '\n'.join(lines)
-
-
-def _run_pyvista(argv: list[str] | None = None, *, as_script: bool = False):
```
user27182, line 24: Looks like all the tests have been changed to call `main` exclusively, rather than running a subprocess. This is generally Ok, but we should test the subprocess invocation of `pyvista` at least once. This is important because there is a script entry point defined in pyproject.toml that enables doing `pyvista report` rather than `python -m pyvista report` (though both should be possible).

There is technically indirect coverage of this from CI / tox (which invokes `pyvista report`) but I'd prefer to have an explicit test for this.

tests/test_cli.py
```diff
@@ -1,115 +1,155 @@
...
+    assert e.value.code == 1
+    assert expected == capsys.readouterr().out
+
+
+@pytest.mark.usefixtures('patch_app_console')
+def test_bad_kwarg(capsys: pytest.CaptureFixture):
+    expected = textwrap.dedent(
+        """\
+    ╭─ Error ────────────────────────────────────────────────────────────╮
+    │ Unknown option: "--foo=1".                                         │
```
user27182, line 92: Is cycleopts smart enough to suggest what the available options are (from the annotations?) if so, can we include that output here as part of the test? The previous version included a link to the docs to try to address this...

tests/test_cli.py
```diff
@@ -1,115 +1,155 @@
...
+    def case_no_kw(self):
+        return '', {}
+
+    @parametrize(downloads=['True', 'yes', 'y', 'true'])
+    @parametrize(sort=['True', 'yes', 'y', 'true'])
+    def case_kw_bool(self, downloads, sort):
+        return f'--downloads={downloads} --sort={sort}', dict(downloads=True, sort=True)
+
+    @parametrize(downloads=['False', 'no', 'n', 'false'])
+    @parametrize(sort=['False', 'no', 'n', 'false'])
```
user27182, line 117: The original tests used gpu=False to reduce test times. Can we include that here too?

edabor: Test time is not an issue here since the class is patched

tox.ini
edabor: Done in b773d694596c84241078f62949278fb70ac9a960

pyproject.toml
edabor: Done in b71ad6772391e0292a0a332d560a6

tests/test_cli.py
edabor: Done in 12c092b2d4d9e2ad4c

tests/test_cli.py
edabor: Done in 46e94d9f2053a5606e93432

pyvista/__main__.py
```diff
@@ -2,116 +2,45 @@
...
+import warnings
 
-import pyvista
-
-if TYPE_CHECKING:
-    from collections.abc import Callable
-
-COMMANDS: dict[str, Callable[..., Any]] = {
-    'report': pyvista.Report,
-    'plot': pyvista.plot,
```
user27182, line 18: We should keep the `pyvista plot` command from #7970 (that PR created many merge conflicts, sorry!)

edabor: No worries, I'll include it.

edabor: We might need https://github.com/pyvista/pyvista/pull/7978 before for cyclopts to coerce arguments for `plot` though ...

akaszynski:
@beroda, you OK if #7989 is merged first?

edabor:
@akaszynski Yup

akaszynski (dismissed):
Aside from adding an additional dependency, fully agree `cyclopts` is worth it for CLI output formatting.

LGTM, relying on @user27182's through feedback (thanks!).

pyvista/report.py
```diff
@@ -291,6 +294,7 @@ def __init__(  # noqa: PLR0917
             'trame_vuetify',
             'jupyter_server_proxy',
             'nest_asyncio',
+            'cyclopts',
```
user27182: Is this a required dependency? There is a "core" list of report deps, should we add it there instead?

akaszynski: It's a required dependency, so yes, we should move that.

edabor: Done in d73604eb95dee762dfeffff1e6a7e0f1e

user27182:
Maybe we should move this `StrEnum` import to core misc utilities for reuse by this PR?

https://github.com/pyvista/pyvista/blob/789247fe438fd18752bafb7d26f76e106daf06ca/doc/source/make_tables.py#L57-L64

edabor:
> Maybe we should move this `StrEnum` import to core misc utilities for reuse by this PR?
> 
> https://github.com/pyvista/pyvista/blob/789247fe438fd18752bafb7d26f76e106daf06ca/doc/source/make_tables.py#L57-L64

Done in b4bd8bc33f5f535a1dcae90f6413c77ebd3c257c

doc/source/user-guide/simple.rst
```diff
@@ -221,6 +221,20 @@ your use case:
...
 .. _pyvistaqt: https://qtdocs.pyvista.org/
 
+Command Line
+------------
+
+Starting from version ``0.47``, meshes can be plotted from one (or more) files using a command line tool such that:
+
+.. code-block:: shell
+    pyvista plot <files> [OPTIONS]
+    python -m pyvista plot <files> [OPTIONS]
```
edabor, line 233:
```suggestion
.. code-block:: shell

    pyvista plot <files> [OPTIONS]
    python -m pyvista plot <files> [OPTIONS]
```

doc/source/make_tables.py
```diff
@@ -39,6 +38,7 @@
 from pyvista.core.celltype import _CELL_TYPE_INFO
 from pyvista.core.celltype import PLACEHOLDER
 from pyvista.core.filters.data_object import _get_cell_quality_measures
+from pyvista.core.utilities import StrEnum
```
user27182:
```suggestion
from pyvista.core.utilities.misc import StrEnum
```

doc/source/user-guide/simple.rst
```diff
@@ -221,6 +221,21 @@ your use case:
 
 .. _pyvistaqt: https://qtdocs.pyvista.org/
 
+Command Line
+------------
+
+Starting from version ``0.47``, meshes can be plotted from one (or more) files using a command line tool such that:
```
user27182:
```suggestion
Starting from version ``0.47``, meshes can be plotted from one (or more)
files using a command line tool such that:
```

doc/source/user-guide/simple.rst
```diff
@@ -221,6 +221,21 @@ your use case:
...
+
+.. code-block:: shell
+
+    pyvista plot <files> [OPTIONS]
+    python -m pyvista plot <files> [OPTIONS]
+
+Most options available in :func:`pyvista.plot` are exposed by the command line.
+Run the ``--help`` command to see all options:
+
+.. program-output:: pyvista plot --help
```
user27182: The output has color codes and these are literally included here, not sure if there are directive options for this

image

pyvista/__main__.py
```diff
@@ -2,116 +2,230 @@
...
+            return token.value
+    return None
+
+
+_HELP_KWARGS = """\
+Additional keyword arguments passed to ``Plotter.add_mesh`` or ``Plotter.add_volume``.
+See documentation for more details https://docs.pyvista.org/api/plotting/_autosummary/pyvista.plotter.add_mesh
+and https://docs.pyvista.org/api/plotting/_autosummary/pyvista.plotter.add_volume
+
+Note that contrary to other arguments, hyphens CANNOT not be used (ie. use ``--show_edges=True`` instead of ``--show-edges=True``).
```
user27182:
```suggestion
Note that contrary to other arguments, hyphens CANNOT be used (ie. use ``--show_edges=True`` instead of ``--show-edges=True``).
```

Can you comment on why hyphens don't work? It's just for `plot`?

pyvista/__main__.py
```diff
@@ -2,116 +2,230 @@
...
+@wraps(Report)
+def report(*args, **kwargs):  # noqa: ANN201, D103
+    return Report(*args, **kwargs)
 
-    return py_args, py_kwargs
 
+def _validator_window_size(type_: type, value: list[int] | None) -> None:  # noqa: ARG001
+    if value is not None and len(value) != 2:
+        msg = 'Window size must be a list of two integers.'
+        raise ValueError(msg)
```
user27182, line 58: Are multiple window sizes needed in case you plot multiple files, and each file gets its own window? Does N files need to match N window sizes?

Maybe this can be better documented in terms of what it means to plot multiple files.

I can see it being useful to also plot all files in a single window too, not sure which use case is being targeted here.

pyvista/__main__.py
```diff
@@ -2,116 +2,230 @@
...
+        Parameter(
+            consume_multiple=True,
+            help='File(s) to plot. Must be readable with ``pv.read``. If nothing is provided, show an empty window.',  # noqa: E501
+            validator=_validator_files,
+            group=Groups.IN,
+        ),
+    ] = None,
+    *,
+    off_screen: Annotated[bool | None, Parameter(group=Groups.PLOTTER)] = None,
+    full_screen: Annotated[bool | None, Parameter(group=Groups.RENDERING)] = None,
```
user27182, line 144: Is there a sync mechanism to ensure these annotations match `pyvista.plot`? E.g. what happens if we change `list[int]` to `Sequence[int]` for `window_size`?

pyvista/__main__.py
```diff
@@ -2,116 +2,230 @@
...
+    show_bounds: Annotated[bool, Parameter(group=Groups.RENDERING)] = False,
+    show_axes: Annotated[bool | None, Parameter(group=Groups.RENDERING)] = None,
+    background: Annotated[str | None, Parameter(group=Groups.RENDERING)] = None,
+    text: Annotated[str, Parameter(group=Groups.RENDERING)] = '',
+    eye_dome_lighting: Annotated[bool, Parameter(group=Groups.RENDERING)] = False,
+    volume: Annotated[bool, Parameter(group=Groups.RENDERING)] = False,
+    parallel_projection: Annotated[bool, Parameter(group=Groups.RENDERING)] = False,
+    return_cpos: Annotated[bool, Parameter(group=Groups.RETURN)] = False,
+    anti_aliasing: Annotated[
+        Literal['ssaa', 'msaa', 'fxaa'] | None, Parameter(group=Groups.RENDERING)
```
user27182, line 164:
```suggestion
        Literal['ssaa', 'msaa', 'fxaa'] | None, Parameter(group=Groups.RENDERING)
```
Maybe we can define this Literal once and reuse it?

tests/test_cli.py
```diff
@@ -1,143 +1,586 @@
...
+
+        Run ``pyvista report --help`` for more details on available 
+    parameters.
+
+    ╭─ Parameters ───────────────────────────────────────────────────────╮
+    │ ADDITIONAL --additional  List of packages or package names to add  │
+    │   --empty-additional     to output information.                    │
+    │ NCOL --ncol              Number of package-columns in html table;  │
+    │                          only has effect if                        │
+    │                          ``mode='HTML'`` or ``mode='html'``.       │
```
user27182: Should we include a similar test for `plot`? I'm not really sure what the `Group` does, do these show up in the help?

tests/test_cli.py
```diff
@@ -1,143 +1,586 @@
...
+    return mock_plotter().add_volume
+
+
+@pytest.fixture
+def mock_files_validator(mocker: MockerFixture):
+    mocker.patch.object(pv.__main__, '_validator_files')
+
+
+@fixture
+def default_plot_kwargs():
```
user27182: Do these defaults need to be updated and synced with `pyvista.plot`? E.g. if a new option is added or a default is changed

doc/source/user-guide/simple.rst
edabor: According to https://github.com/OpenNTI/sphinxcontrib-programoutput/issues/49#issuecomment-809313914, it does not seem to be possible. I think I'll remove the `sphinxcontrib-programoutput` extension.

One option would be to control colors activation via an env variable, but I don't think it is worth it ...

pyvista/__main__.py
edabor: This validator only checks that the following commands will not work because it does not make sense
to use a window size different than a length 2 iterable:
```shell
pyvista plot --window-size="[100,200,300]"
pyvista plot --window-size=100
```

Currently, all N files are opened inside a single window, and this parameter controls the window size.
I've updated the documentation for multiple files in 970bf5969f9355d7253d7d04cb3321494653bad5

pyvista/__main__.py
edabor: Strictly speaking, a hyphen is totally valid when providing supplementary keyword arguments.

However, contrary to other arguments defined in the function signature, they are not converted to underscores and the **kwargs dictionary is filled with keys left intact.

For example:
```shell
pyvista plot file.vtp --show-edges=True
```
would call `plotter.add_mesh` with `kwargs = {"show-edges": True}`.

However,
```shell
pyvista plot file.vtp --show_edges=True
```
would call `plotter.add_mesh` with `show_edges = True`, which is the intended usage.

Don't know if I'm clear though 😅 

I suspected that this could confuse users when seeing errors such that `"show-edges" is an invalid keyword argument for "_common_arg_parser"`

Though, I agree that raising an error might be too brutal, so I changed to print a warning panel and continue the execution in b002e0456b5e60ef2bd32880e3840f4cfc30450c

tests/test_cli.py
edabor: I was a bit hesitant whether to test the full help string, because it can be very brittle. What's your feeling about it ?

Regarding the `Group`, yes they show up in the help page.
I found it useful to distinguish the usage of each parameter, see below:

```
Usage: pyvista plot file (file2) [OPTIONS]

Plot a PyVista, numpy, or vtk object.

.. versionadded:: 0.47

    ``plot`` can be invoked with the shell command:

    .. code-block:: shell

        pyvista plot <files> --screenshot output.png --off-screen

    Run ``pyvista plot --help`` for more details on available parameters.

    .. note::
        Providing multiple files renders them inside the same window.

╭─ Inputs ─────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ FILES --files --empty-files  File(s) to plot. Must be readable with ``pv.read``. If nothing is provided,     │
│                              show an empty window.                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Plotter init ───────────────────────────────────────────────────────────────────────────────────────────────╮
│ --off-screen --no-off-screen       Plots off screen when ``True``.  Helpful for saving                       │
│                                    screenshots without a window popping up.  Defaults to the                 │
│                                    global setting ``pyvista.OFF_SCREEN``.                                    │
│ --screenshot                       Saves screenshot to file when enabled.  See:                              │
│                                    :func:`Plotter.screenshot() <pyvista.Plotter.screenshot>`.                │
│                                    Default ``False``.                                                        │
│                                                                                                              │
│                                    When ``True``, takes screenshot and returns ``numpy`` array of            │
│                                    image.                                                                    │
│ --interactive --no-interactive     Allows user to pan and move figure. [default: True]                       │
│ --window-size --empty-window-size  Window size in pixels.                                                    │
│ --border --no-border               Draw a border around each render window. [default: False]                 │
│ --border-color                     Either a string, rgb list, or hex color string.  For example:             │
│                                                                                                              │
│                                        * ``color='white'``                                                   │
│                                        * ``color='w'``                                                       │
│                                        * ``color=[1.0, 1.0, 1.0]``                                           │
│                                        * ``color='#FFFFFF'`` [default: k]                                    │
│ --border-width                     Width of the border in pixels when enabled. [default: 2.0]                │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Rendering ──────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --full-screen --no-full-screen  Opens window in full screen.  When enabled, ignores                          │
│                                 ``window_size``.                                                             │
│ --show-bounds --no-show-bounds  Shows mesh bounds when ``True``. [default: False]                            │
│ --show-axes --no-show-axes      Shows a vtk axes widget.                                                     │
│ --background                    Color of the background.                                                     │
│ --text                          Adds text at the bottom of the plot. [default: ]                             │
│ --eye-dome-lighting             Enables eye dome lighting. [default: False]                                  │
│   --no-eye-dome-lighting                                                                                     │
│ --volume --no-volume            Use the :func:`Plotter.add_volume()                                          │
│                                 <pyvista.Plotter.add_volume>` method for volume rendering. [default: False]  │
│ --parallel-projection           Enable parallel projection. [default: False]                                 │
│   --no-parallel-projection                                                                                   │
│ --anti-aliasing                 Enable or disable anti-aliasing. If ``True``, uses ``"msaa"``. If False,     │
│                                 disables anti_aliasing. If a string, should be one of the following:         │
│                                                                                                              │
│                                 * ``"ssaa"`` - Super-Sample Anti-Aliasing                                    │
│                                 * ``"msaa"`` - Multi-Sample Anti-Aliasing                                    │
│                                 * ``"fxaa"`` - Fast Approximate Anti-Aliasing                                │
│                                                                                                              │
│                                 Defaults to :attr:`pyvista.plotting.themes.Theme.anti_aliasing` [choices:    │
│                                 ssaa, msaa, fxaa]                                                            │
│ --zoom                          Camera zoom.  Either ``'tight'`` or a float. A value greater than 1          │
│                                 is a zoom-in, a value less than 1 is a zoom-out.  Must be greater            │
│                                 than 0.                                                                      │
│ --ssao --no-ssao                Enable surface space ambient occlusion (SSAO). See                           │
│                                 :func:`Plotter.enable_ssao` for more details. [default: False]               │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Return ─────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --return-cpos --no-return-cpos  Return the last camera position from the render window                       │
│                                 when enabled.  Defaults to value in theme settings. [default: False]         │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Supplementary ──────────────────────────────────────────────────────────────────────────────────────────────╮
│ --[KEYWORD]  Additional keyword arguments passed to ``Plotter.add_mesh`` or ``Plotter.add_volume``.          │
│              See the documentation for more details at                                                       │
│              https://docs.pyvista.org/api/plotting/_autosummary/pyvista.plotter.add_mesh                     │
│              and https://docs.pyvista.org/api/plotting/_autosummary/pyvista.plotter.add_volume               │
│                                                                                                              │
│              Note that contrary to other CLI arguments, hyphens ``-`` are not converted to underscores ``_`` │
│              before being passed                                                                             │
│              to the corresponding plotter method. For example, you need to use ``--show_edges=True`` instead │
│              of ``--show-edges=True`` to                                                                     │
│              show mesh edges in the plotting window.                                                         │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

edabor: I decided to only test the bottom of the help page in 4d650f29589a090fe6d66b09ed862a2438635b1d

doc/source/user-guide/simple.rst
edabor: I removed the output in 970bf5969f9355d7253d7d04cb3321494653bad5

tests/test_cli.py
edabor: I've consolidated the tests in 1252ba7f3ae0949 such that it will fail if:

- a new argument in `pv.plot` is not available in `pv.__main__._plot` (you can still exclude some in the `missing_plot_arguments` fixture)
- a default value is changed in `pv.plot` but is not reflected in `pv.__main__._plot`

pyvista/__main__.py
edabor: I guess we should rely on `mypy` to help us catch that.

Note that in an attempt to help `cyclopts` correctly coerce command line arguments strings to the expected types, I've slightly restricted the type hints. Look at `background` arg for example: it is typed as `ColorLike` in `pv.plot` but is typed as `str` in `pv.__main__._plot`.

I can try to add some tests for types similar to what's been done for default values in 1252ba7f3ae0949bbea25f7d59ef85d8c1028cb0 by excluding several arguments.

doc/source/user-guide/simple.rst
user27182: I think the color codes may be enabled by the same env var used to color the sphinx console output. Maybe we can change this for pushes to main only, so that published docs don't forcefully enable console colors?

We can try to re-enable this in a follow up PR

pyvista/__main__.py
edabor: Tried something in 11a99698d3a83c975a7329341c199f8508b03bd2 Feels a bit hacky but works. What's your opinion ?

pyvista/__main__.py
user27182: Raising an error might have been ok, I was asking because of thought I saw tests where `--off-screen` was used, so it seemed like hyphens work sometimes but not always. And I saw kwargs validation which seem to try to normalize `-` into `_` so wasn't sure exactly what works and what doesn't

pyvista/__main__.py
user27182: Yea there are some important and necessary deviations for CLI. Probably Ok to rely on mypy and cyclopts auto-parsing docstrings
 And kwargs should at least implicitly capture any new options added, even if they don't get properly updated for the --help

pyvista/__main__.py
user27182: This can be a separate PR

tests/test_cli.py
user27182: It is a bit verbose and brittle. So this is fine.

pyvista/__main__.py
edabor: Currently, a warning panel is emitted for users to be aware that the hyphen will not be converted.
image

Moreover, the error is printed at the end.
image

PR #7993, opened 2025-10-06
Add ONNX Runtime support for ML inference

tkoyama010:
Add ONNX Runtime support for machine learning inference in PyVista.

## Changes
- New `infer_scalars_with_onnx()` filter method for running ONNX models on datasets
- Example demonstrating ONNX inference workflow and use cases
- Requires VTK 9.6.0+ with ONNX Runtime support

## Benefits
- 100-2000x speedup vs traditional simulations
- Enables surrogate modeling and parameter studies
- Works with PyTorch, TensorFlow, and other ML frameworks

Based on: https://www.kitware.com/enhance-your-paraview-and-vtk-pipelines-with-artificial-neural-networks/

### DIFF
--- a/examples/01-filter/onnx_inference.py
+++ b/examples/01-filter/onnx_inference.py
@@ -1,0 +1,353 @@
+"""
+.. _onnx_inference_example:
+
+ONNX Model Inference for Surrogate Modeling
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Demonstrate ONNX Runtime integration for machine learning inference in PyVista.
+
+This example shows how to use ONNX (Open Neural Network eXchange) models for
+fast inference directly within PyVista pipelines. This feature enables surrogate
+modeling, parameter studies, and inverse problems in scientific visualization.
+
+ONNX Runtime support was introduced in VTK 9.6.0 and allows integration of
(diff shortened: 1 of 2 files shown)


examples/01-filter/onnx_inference.py
```diff
@@ -0,0 +1,355 @@
...
+# ~~~~~~~~~~~~~~~~~
+#
+# Plot the stress distribution on the mesh
+
+result.plot(
+    scalars='von_mises_stress',
+    cmap='jet',
+    show_edges=True,
+    scalar_bar_args={'title': 'Von Mises Stress (Pa)'},
+    cpos='xy',
```
tkoyama010: @copilot 

```suggestion
```

banesullivan:
This is interesting! This is actually an opportunity for a much bigger discussion on how to handle VTK capabilities that are not exposed in the default VTK build. As I understand it, you'd need a custom build of VTK for this functionality? If so we can't actually test and validate what we're wrapping in PyVista and for that reason, I'd consider this out of scope for pyvista-core and better suited for a new pyvista-onnx package. However, I see you want the convenience of this being a simple filter call bound to datasets. What we should do is implement a generalized mechanism to register addition filtes per dataset type much like we now have the reader/writer registries for third party packages. I will open a more detailed feature request issue for this and in the meantime, closing this PR

PR #7994, opened 2025-10-06
chore: update ruff pre-commit hook to v0.13.3

tkoyama010:
Update ruff-pre-commit from v0.12.11 to v0.13.3 and fix new linting errors.

Separated from #7992.

### DIFF
--- a/pyvista/core/utilities/fileio.py
+++ b/pyvista/core/utilities/fileio.py
@@ -41,7 +41,7 @@ if TYPE_CHECKING:
     from pyvista.core.utilities.reader import BaseReader
     from pyvista.plotting.texture import Texture
 
-_CompressionOptions = Literal['zlib', 'lz4', 'lzma', None]
+_CompressionOptions = Literal['zlib', 'lz4', 'lzma'] | None
 PathStrSeq = str | Path | Sequence['PathStrSeq']
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


pyvista/core/utilities/fileio.py
```diff
@@ -41,7 +41,7 @@
     from pyvista.core.utilities.reader import BaseReader
     from pyvista.plotting.texture import Texture
 
-_CompressionOptions = Literal['zlib', 'lz4', 'lzma', None]
+_CompressionOptions = Literal['zlib', 'lz4', 'lzma'] | None
```
user27182: This auto-fix is not safe, we should revert it:
```suggestion
_CompressionOptions = Literal['zlib', 'lz4', 'lzma', None]  # noqa: PYI061
```
Issue raised here: https://github.com/astral-sh/ruff/issues/20729

tkoyama010:
pre-commit.ci autofix

pyvista/core/utilities/fileio.py
user27182: This fix was reverted previously https://github.com/pyvista/pyvista/pull/7994#discussion_r2408961853 but looks like it got re-applied

```suggestion
_CompressionOptions = Literal['zlib', 'lz4', 'lzma', None]
```

user27182:
```suggestion
_CompressionOptions = Literal['zlib', 'lz4', 'lzma', None]  # noqa: PYI061
```

Issue #8004, opened 2025-10-10
`pyvista plot` with invalid args does not warn or error


user27182:
### Describe the bug, what's wrong, and what you expected.

Running `pyvista plot --foo=1` opens an empty plot without errors or warnings. Should we warn if no files are passed? Should we error if kwargs are invalid?

### Steps to reproduce the bug.

`pyvista plot --foo=1`

### System Information

```shell
main branch
```

edabor:
For the unknown kwargs, you will get an error if the a file is passed to the CLI, since `pl.add_mesh` of `pl.add_volume` are called and `_common_arg_parser` is called.

<img width="638" height="152" alt="Image" src="https://github.com/user-attachments/assets/6662e7a3-c859-4bb7-a059-da485f831f22" />

Maybe we shouldn't allow 0 files to be passed such that supplementary kwargs are checked ...

PR #8005, opened 2025-10-10
Add `convert` CLI command for converting mesh file format

user27182:
### Overview

Enable `pyvista convert <input> <output>` via CLI


pyvista/__main__.py
```diff
@@ -52,6 +54,107 @@ def report(*args, **kwargs):  # noqa: ANN201, D103
...
+    out_spec_path = Path(out_spec)
+    in_stem = in_path.stem
+    in_suffix = in_path.suffix
+
+    # Disallow pure directory targets
+    if out_spec_path.is_dir() or str(out_spec_path).endswith('/'):
+        console_error(
+            f'Invalid value for output: {out_spec_path!r}\n'
+            "Specify a filename or extension (e.g. 'bar/*.xyz' or 'bar/foo.xyz')."
+        )
```
edabor: For these kind of Path argument checks, I recommend you using the `cyclopts` validators https://cyclopts.readthedocs.io/en/latest/parameter_validators.html#path

tests/test_cli.py
```diff
@@ -85,7 +89,8 @@ def test_invalid_command(capsys: pytest.CaptureFixture):
 
 
 @pytest.mark.usefixtures('patch_app_console')
-def test_bad_kwarg_report(capsys: pytest.CaptureFixture):
+@pytest.mark.parametrize('command', ['report', 'convert'])  # 'plot'
```
user27182, line 92: IMO plot should also pass this test (i.e. fail with bad kwargs) but doing so doesn't currently raise an error, see #8004 

```suggestion
@pytest.mark.parametrize('command', ['report', 'convert', 'plot'])
```

pyvista/__main__.py
user27182: Done in [5800547](https://github.com/pyvista/pyvista/pull/8005/commits/58005478eafdc8c1ae2dbc12b345602af100d42e) for the inputs (still need a validator for the output though)

I reused the existing file validator for `plot`, but I noticed it reads the mesh and doesn't store the output. This means the mesh is read twice: once for validation, and again when used by the command. Can we return and keep the output from validators for re-use?

user27182: Output validator also used here [4432533](https://github.com/pyvista/pyvista/pull/8005/commits/443253329d020956f75c5672a045f3e5c62a000b)

edabor: We could try raising an error inside a converter to only read the mesh once https://cyclopts.readthedocs.io/en/latest/parameters.html#converters

user27182: @edabor I used the converter in https://github.com/pyvista/pyvista/pull/8005/commits/2357aeadfaea686083b4e03cb2a6c6c7fa07e731. This works, and all the `convert` tests pass. But the `plot` tests all fail because converter is mocked.

I'm not sure if a mocker should continue to be used in the test because we now need to mimic the converter logic in the mocker to ensure the dummy files are properly converted to dummy meshes...

Should we use actual files for the plotting tests instead, and remove the converter mocker?

user27182: > Should we use actual files for the plotting tests instead, and remove the converter mocker?

I definitely think the converter should not be mocked as it's important to test the actual converter is working. I tried doing so in [fc3b413](https://github.com/pyvista/pyvista/pull/8005/commits/fc3b413339db8a8b586f3b6e1d0d25d78cdd7847), and I managed to get basic plot tests to pass, but I think I'm running into issues where other things in the tests are mocked/patched since the converter doesn't to seem to be called in some cases.

I am finding the tests hard to debug with all the mocking, and am thinking of simply removing all of it for easier debugging and maintenance.

Alternatively, I could leave the exising `_validator_files` in place for `pyvista plot`, and only use `_converter_file` for `pyvista convert`. Then maybe you can submit a follow-up PR which updates `pyvista plot` to use the same `_converter_file` and then also update all the tests to make it work properly.

user27182: > Alternatively, I could leave the exising `_validator_files` in place for `pyvista plot`, and only use `_converter_file` for `pyvista convert`. Then maybe you can submit a follow-up PR which updates `pyvista plot` to use the same `_converter_file` and then also update all the tests to make it work properly.

I have reverted all changes to `plot`. Will leave the switch from `_validator_files` -> `_converter_files` with `plot` to a separate PR

edabor: Sorry for the excessive mocking usage :angel:
I can handle converting the validator to converter in https://github.com/pyvista/pyvista/pull/8011 and try to reduce the test maintenance cost.

tests/test_cli.py
edabor: On going in https://github.com/pyvista/pyvista/pull/8011

pyvista/__main__.py
```diff
@@ -46,18 +49,125 @@
...
 
 
+def _converter_files(
+    type_: type,  # noqa: ARG001
+    tokens: Sequence[Token],
+) -> tuple[list[DataObject], list[Path]]:
+    values: list[str] = [t.value for t in tokens]
+    if len(values) == 0:
+        return [], []
+
```
user27182: This branch is missing coverage. It was originally included to handle the no files case for `plot`, so we can remove it.

```suggestion
```

pyvista/__main__.py
```diff
@@ -46,18 +49,125 @@
...
+    # Parse output specification
+    out_spec_path = Path(file_out)
+    out_dir = out_spec_path.parent
+    out_suffix = out_spec_path.suffix
+    if '*' in (spec_stem := str(out_spec_path.stem)):
+        # Pattern like "*.stl" or "bar/*.stl"
+        out_stem = spec_stem.replace('*', path_in.stem, 1)  # type: ignore[attr-defined]
+    else:
+        # Explicit filename with extension
+        out_stem = out_spec_path.stem
```
edabor: IMO, a wildcard has a very specific meaning (ie. all matches) which is not compatible with the usage here (ie. match previous file stem).

Maybe a placeholder such as "{}.stl" could fit ? I agree though that it requires more typing, but it could allow customizing the output file using the input such as `pyvista convert file.vtp {}_converted.stl`.

pyvista/__main__.py
user27182: > Sorry for the excessive mocking usage :angel:
> 
> I can handle converting the validator to converter in https://github.com/pyvista/pyvista/pull/8011 and try to reduce the test maintenance cost.

Okay thanks! Feel free to reuse the converter code from this PR. It currently returns the meshes _and_ paths since the paths are needed for `convert`, but you can return meshes only for #8011 and I'll modify to include paths as part of this PR

user27182: My previous comment assumed #8011 would be merged first. But we can alternatively merge this first as-is, then remove _validator_files and use _converter_files as part of #8011 instead

pyvista/__main__.py
```diff
@@ -46,18 +49,122 @@
...
+      $ pyvista convert foo.abc bar.xyz
+      Saved: bar.xyz
+
+      $ pyvista convert foo.abc '*.xyz'
+      Saved: foo.xyz
+
+    """
+    # get input mesh and input path from file_in str token
+    # which was converted to tuple[meshes, paths]
+    mesh_in, path_in = file_in[0][0], file_in[1][0]
```
edabor: Maybe a list of dedicated `dataclass` or `namedtuple` as return type of `_converter_files` could be more easier to understand ?

pyvista/__main__.py
user27182: > IMO, a wildcard has a very specific meaning (ie. all matches) which is not compatible with the usage here (ie. match previous file stem).


Agreed, in addition if you don't use quotes either on the output path the shell will find all *.ext files and (confusingly) include them all as extra tokens and raise a somewhat confusing too many tokens error. 


> Maybe a placeholder such as "{}.stl" could fit ? I agree though that it requires more typing, but it could allow customizing the output file using the input such as `pyvista convert file.vtp {}_converted.stl`.

The current behavior is already such that the single character "*" is replaced with the input stem, so `_converted.stl` is already appended (though I don't think there is a test case for this, which I should add)

Can we use some other single char placeholder? Maybe "%"?

edabor: Maybe we should stick to Zen "Explicit is better than implicit." and only allow the full file path ... :thinking:

pyvista/__main__.py
user27182: Part of the issue may be with `cyclopts` (or maybe I don't know how to use it properly), because I want to Annotate it as `str` so that the help shows `str` as input type, but the converter then maps `str` -> `tuple[list[mesh], list[mesh]]`, so this results in the type hint being incorrect, where mypy thinks we're dealing with a single string but we're not.

But I'll otherwise update this to return a list of named tuples instead

pyvista/__main__.py
user27182: Okay, how about we allow `pyvista convert foo.stl .ply`, where the output must be an extension (and only an extension), so that no wildcards are needed, and this saves `foo.ply`. And otherwise, you must supply the full path.

edabor: I also thought about it, but dot also has a specific usage for hidden files :sweat_smile: But still better than wildcard though. +1 for this option

pyvista/__main__.py
edabor: The help do not show the type hints (at lease on my side).

user27182: > Maybe a list of dedicated `dataclass` or `namedtuple` as return type of `_converter_files` could be more easier to understand ?

This is done in [271e78c](https://github.com/pyvista/pyvista/pull/8005/commits/271e78c6b7210523c1479fa52037c15ef52e6e59)

we now have:

```py
    mesh_in = file_in[0].mesh
    path_in = file_in[0].path 
```

But I had to keep the annotation as `str`, or at least it cannot be `list[_MeshAndPath]` (i.e. the actual converted type), because if I make this single change to the Annotated type, I get a strange error:

``` 
/Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/typing.py:1081: in _evaluate
    eval(self.__forward_code__, globalns, localns),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   NameError: name 'Self' is not defined

<string>:1: NameError
```

Which to me seems like a bug with `cycleopts` (it makes a call to `get_type_hints` somewhere up the stack). Even if we could get the type hint correct, the name of the parameter is still `file_in` which doesn't really match the `list[tuple]` type. And the parameter name is definitely used in the help (I checked this time)


Since there is already a bit of a mismatch between var name and type, I think it's maybe OK to also keep the `str` type annotation too.

edabor: Yes, `get_type_hints` is limited for types which are not in the locals() and globals(). I might fill and issue to `cyclopts` any time soon with that, because it can get annoying for custom types.

Let's keep the `str` type hint, it is fine for me.

> And the parameter name is definitely used in the help (I checked this time)

You're right, I thought you meant parameter type.

.pre-commit-config.yaml
```diff
@@ -90,10 +90,10 @@ repos:
...
       - id: rst-inline-touching-normal
 
-  - repo: https://github.com/scop/pre-commit-shfmt
-    rev: v3.12.0-2
-    hooks:
-      - id: shfmt
+  #  - repo: https://github.com/scop/pre-commit-shfmt
+  #    rev: v3.12.0-2
+  #    hooks:
+  #      - id: shfmt
```
edabor: Why this change ?

user27182: I'll revert. It's really annoying, this hook doesn't work for me and break my local pre-commit. So I'm constantly commenting it out for dev, and sometimes forget to uncomment

user27182: If you know how to fix this let me know. I get this error when I run pre-commit:

```
    Failed to build shfmt_py
stderr:
      error: subprocess-exited-with-error
      
      × Building wheel for shfmt_py (pyproject.toml) did not run successfully.
      │ exit code: 1
      ╰─> [20 lines of output]
          /private/var/folders/2v/bv4k0q852z7dv3qlx1zwf2h80000gn/T/pip-build-env-eu6wsl4s/overlay/lib/python3.13/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
          !!
          
                  ********************************************************************************
                  Please consider removing the following classifiers in favor of a SPDX license expression:
          
                  License :: OSI Approved :: Apache Software License
                  License :: OSI Approved :: BSD License
                  License :: OSI Approved :: MIT License
          
                  See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
                  ********************************************************************************
          
          !!
            self._finalize_license_expression()
          running bdist_wheel
          running build
          running setuptools_download
          => downloading shfmt...
          error: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1028)>
          [end of output]
      
      note: This error originates from a subprocess, and is likely not a problem with pip.
      ERROR: Failed building wheel for shfmt_py
    error: failed-wheel-build-for-install
    
    × Failed to build installable wheels for some pyproject.toml based projects
    ╰─> shfmt_py

PR #8008, opened 2025-10-10
Include interactive `vtksz` files in documentation tests

user27182:
### Overview

Using `--include_vtksz` option from `pytest-pyvista`

### DIFF
--- a/pyvista/__main__.py
+++ b/pyvista/__main__.py
@@ -140,7 +140,7 @@ def _validator_has_extension(type_: type, value: str) -> None:  # noqa: ARG001
     has_suffix = bool(path.suffix)
     is_suffix = not path.suffix and path.stem.startswith('.')
     if not (has_suffix or is_suffix):
-        msg = 'Output file must have a file extension.'
+        msg = '\nOutput file must have a file extension.'
         raise ValueError(msg)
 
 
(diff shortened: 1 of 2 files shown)


user27182:
> LGTM and thanks for adding this.
> 
> 
> 
> One note: Some of these images are empty (e.g. `pyvista-examples-downloads-download_sky_box_cube_map-e7a895fcb7f3f564_00_00_vtksz.jpg`), and the corresponding interactive doc example also shows an empty example (as expected). We might want to flag these during testing and raise an error in a followup PR and disable the interactive examples if possible.

Agreed we should not include empty images or blank/broken interactive plots. A first manual review will be good to address this, though adding a feature like https://github.com/pyvista/pytest-pyvista/issues/238 is probably needed to catch future cases too

edabor:
Should we modify these lines ? It appears to me that we test only a subset of `vtksz` files

https://github.com/pyvista/pyvista/blob/4769c2192dc7781142c63d5f0335788b24259668/.github/workflows/docs.yml#L82-L90

user27182:
> Should we modify these lines ? It appears to me that we test only a subset of `vtksz` files
> 
> 
> 
> https://github.com/pyvista/pyvista/blob/4769c2192dc7781142c63d5f0335788b24259668/.github/workflows/docs.yml#L82-L90

These other files are left over from the build process. Basically there is an inefficiency with the build where the plot directive files are first generated in one dir, and then copied to _images when the build completes. Only files in _images are shown in the docs (and included in tests). Any other vtksz files elsewhere are not needed and not used by the docs.

If you look at the blame and track down the PR where this change was made there is a comment I made about this somewhere.

edabor:
> These other files are left over from the build process. Basically there is an inefficiency with the build where the plot directive files are first generated in one dir, and then copied to _images when the build completes. Only files in _images are shown in the docs (and included in tests). Any other vtksz files elsewhere are not needed and not used by the docs.
> 
> If you look at the blame and track down the PR where this change was made there is a comment I made about this somewhere.

Sorry for the miss. Great !

PR #8011, opened 2025-10-11
Disallow no files for `pyvista plot` CLI

edabor:
### Overview

Fixes https://github.com/pyvista/pyvista/issues/8004

- [x] Disallow 0 files for `pyvista plot`
- [x] Simplify test suite by limiting mocking (per https://github.com/pyvista/pyvista/pull/8005#discussion_r2423056098)
- [x] Use converter from https://github.com/pyvista/pyvista/pull/8005 to prevent calling `pv.read` twice (at validation and inside `pl.add_mesh`)

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -89,7 +89,7 @@ def test_invalid_command(capsys: pytest.CaptureFixture):
 
 
 @pytest.mark.usefixtures('patch_app_console')
-@pytest.mark.parametrize('command', ['report', 'convert'])  # 'plot'
+@pytest.mark.parametrize('command', ['report', 'convert'])
 def test_bad_kwarg_command(capsys: pytest.CaptureFixture, command):
     expected = textwrap.dedent(
         """\


tests/test_cli.py
```diff
@@ -495,148 +476,278 @@ def case_anti_aliasing(self, default_plot_kwargs: dict, anti_aliasing: str):
...
+
+    @case(tags='raises')
+    def case_kw_unknown(self):
+        """Test when a supplementary keyword argument does not exists"""
+        return '--foo_bar bar', [
+            '⚠ The following exception has been raised when calling pv.plot: ',
+            '"foo_bar" is an invalid keyword argument for  ',
+            '`_common_arg_parser`  ',
+            'Please check the provided arguments.',
+        ]
```
edabor, line 549: > Thanks. Can we also uncomment `plot` here and include it in this test?
> 
> `@pytest.mark.parametrize('command', ['report', 'convert']) # 'plot'`

@user27182 In fact, this is already covered by this test case.

If `plot` were to be added to the `test_bad_kwarg_command`, that would mean rewriting the test as follows, which IMO is a bit too much.
```python
@pytest.mark.usefixtures('patch_app_console')
@pytest.mark.parametrize('command', ['report', 'convert', 'plot'])
def test_bad_kwarg_command(capsys: pytest.CaptureFixture, command):
    expected = textwrap.dedent(
        """\
    ╭─ Error ────────────────────────────────────────────────────────────╮
    │ Unknown option: "--foo=1".                                         │
    ╰────────────────────────────────────────────────────────────────────╯
    """
    )
    n = 4
    if command == 'plot':
        expected = textwrap.dedent(
            """\
        ╭─ PyVista Error ────────────────────────────────────────────────────╮
        │ ⚠ The following exception has been raised when calling pv.plot:    │
        │                                                                    │
        │ ╭─ TypeError ────────────────────────────────────────────────────╮ │
        │ │ "foo" is an invalid keyword argument for `_common_arg_parser`  │ │
        │ ╰────────────────────────────────────────────────────────────────╯ │
        │                                                                    │
        │ Please check the provided arguments.                               │
        ╰────────────────────────────────────────────────────────────────────╯
        """
        )
        n = 10

    command = (
        command
        if command != 'plot'
        else f'{command} --files={Path(pv.examples.antfile).as_posix()}'
    )
    with pytest.raises(SystemExit) as e:
        main(f'{command} --foo=1')
    assert e.value.code == 1
    assert expected == '\n'.join(capsys.readouterr().out.split('\n')[-n:])
```

user27182: Ah ok, then we can just remove the comment for `# 'plot'` then

PR #8013, opened 2025-10-11
Add image `concatenate` filter

user27182:
### Overview

Uses [vtkImageAppend](https://vtk.org/doc/nightly/html/classvtkImageAppend.html#details) internally.


pyvista/core/filters/image_data.py
```diff
@@ -5360,6 +5373,462 @@ def _select_values(  # type: ignore[misc]
...
+              concatenating.
+            - ``'match'``: Cast all array dtypes to match the input's dtype. This casting is
+              unsafe as it may downcast values and lose precision.
+
+        component_policy : 'strict' | 'promote', default: 'strict'
+            - ``'strict'``: Do not modify the number of components of any scalars. All images being
+              concatenated must have the number of components, else a ``ValueError`` is raised.
+            - ``'promote'``: Increase the number of components if necessary. Grayscale scalars
+              with one component may be promoted to RGB or RGBA scalars by duplicating values,
+              and RGB scalars may be promoted to RGBA scalars by including an opacity component.
```
user27182: Not sure if `promote` is the right term to use for this. It's named this way to draw parallels to `promote` for dtype 

EDIT: maybe `increase` is a better term?

akaszynski: Maybe `promote_rgba`? Neither `promote` or `increase` is self documenting.

tests/core/test_imagedata_filters.py
```diff
@@ -2117,3 +2118,311 @@ def test_morphological_filters_custom_kernel_size():
...
+    array_b = np.array([2, 3])
+
+    image_a = pv.ImageData(dimensions=(1, 2, 1))
+    image_a['A'] = array_a
+    image_b = pv.ImageData(dimensions=(2, 1, 1))
+    image_b['B'] = array_b
+
+    match = (
+        'Image 0 dimensions (2, 1, 1) must match the input dimensions (1, 2, 1) along axis 0.\n'
+        'Use the `mode` keyword to allow concatenation with mismatched dimensions.'
```
user27182: This message is a bit unclear, all dimensions _except_ the axis dimension should match.

pyvista/core/filters/image_data.py
user27182: > Maybe `promote_rgba`?

Updated in [63a546b](https://github.com/pyvista/pyvista/pull/8013/commits/63a546bc47ea7fe5669f0c7e43eba06936a3e46f). Do you think this is also a good filter name? Would it be useful to have a dedicated filter for this?

pyvista/core/filters/image_data.py
```diff
@@ -5360,6 +5373,466 @@ def _select_values(  # type: ignore[misc]
...
+            modes operate along the specified ``axis`` except for ``'preserve-extents'``.
+            Specify one of:
+
+            - ``'strict'``: all images must have identical dimensions except along the specified
+              ``axis``.
+            - ``'resample-off-axis'``: :meth:`resample` off-axis dimensions of concatenated images
+              to match the input. The on-axis dimension is `not` resampled.
+            - ``'resample-proportional'``: :meth:`resample` concatenated images proportionally to
+              preserve their aspect ratio(s). For 3D cases, this may not be possible, and a
+              ``ValueError`` may be raised.
```
user27182: It would be nice if the filter just worked for these cases. The issue is that _two_ off-axis dimensions must match for 3D cases. We could resample such that we make one dimension match exactly, and then pad the non-matching dimension with zeros to force it to match.

user27182: I might try to address this in a follow-up PR, maybe add a new keyword to allow choosing the background value that is also used for the other modes

pyvista/core/filters/image_data.py
```diff
@@ -5360,6 +5373,466 @@ def _select_values(  # type: ignore[misc]
...
+        component_policy : 'strict' | 'promote_rgba', default: 'strict'
+            - ``'strict'``: Do not modify the number of components of any scalars. All images being
+              concatenated must have the number of components, else a ``ValueError`` is raised.
+            - ``'promote_rgba'``: Increase the number of components if necessary. Grayscale scalars
+              with one component may be promoted to RGB or RGBA scalars by duplicating values,
+              and RGB scalars may be promoted to RGBA scalars by including an opacity component.
+
+        resample_kwargs : dict, optional
+            Keyword arguments passed to :meth:`resample` when using ``'resample-off-axis'`` or
+            ``'reample-proportional'`` modes. Specify ``interpolation``, ``border_mode`` or
```
user27182:
```suggestion
            ``'resample-proportional'`` modes. Specify ``interpolation``, ``border_mode`` or
```

pyvista/core/filters/image_data.py
```diff
@@ -5360,6 +5373,466 @@ def _select_values(  # type: ignore[misc]
...
+              have the same dtype, else a ``TypeError`` is raised.
+            - ``'promote'``: Use :func:`numpy.result_type` to compute the dtype of the output
+              image scalars. This option safely casts all input arrays to a common dtype before
+              concatenating.
+            - ``'match'``: Cast all array dtypes to match the input's dtype. This casting is
+              unsafe as it may downcast values and lose precision.
+
+        component_policy : 'strict' | 'promote_rgba', default: 'strict'
+            - ``'strict'``: Do not modify the number of components of any scalars. All images being
+              concatenated must have the number of components, else a ``ValueError`` is raised.
```
user27182:
```suggestion
              concatenated must have the same number of components, else a ``ValueError`` is raised.
```

pyvista/core/filters/image_data.py
```diff
@@ -5360,6 +5373,466 @@ def _select_values(  # type: ignore[misc]
...
+            - ``'promote_rgba'``: Increase the number of components if necessary. Grayscale scalars
+              with one component may be promoted to RGB or RGBA scalars by duplicating values,
+              and RGB scalars may be promoted to RGBA scalars by including an opacity component.
```
user27182:
```suggestion
              with one component may be promoted to RGB scalars by duplicating values,
              and RGB scalars may be promoted to RGBA scalars by including an opacity component.
              For integer dtypes, the opacity is set to the max int representable by the dtype;
              for floats it is set to ``1.0``.
```

pyvista/core/filters/image_data.py
```diff
@@ -5360,6 +5373,466 @@ def _select_values(  # type: ignore[misc]
...
+
+            >>> concatenated = beach.concatenate(bird, mode='crop-off-axis')
+            >>> concatenated.plot(**plot_kwargs)
+
+            Reverse the concatenation order.
+
+            >>> concatenated = bird.concatenate(beach, mode='crop-off-axis')
+            >>> concatenated.plot(**plot_kwargs)
+
+            Use ``'crop-center'`` to center-crop the images to match the input's
```
user27182: I think maybe 'crop-same-size' is a better name.
Could also add a new complementary 'resample-same-size' mode.

pyvista/core/filters/image_data.py
akaszynski: Yes, perhaps in a follow-up PR, but not strictly necessary.

Regardless, I do think it's important to have here if the images have a different number of components.

pyvista/core/filters/image_data.py
akaszynski: That's an annoying edge case. This PR is complicated enough as is and a followup might be warranted if we see this edge case in the wild.

pyvista/core/filters/image_data.py
akaszynski: Agreed, specifically if we're:
```
img_shallow_copy.crop(dimensions=self_dimensions)
```

We're simply using the dimensions of the current image.

pyvista/core/filters/image_data.py
user27182: Yea I'll address this in a follow up. I think it's a useful case to support, but the ValueError is good enough for now.

pyvista/core/filters/image_data.py
user27182: Yes, I like the term "crop-same-size". One issue with this though is that the option "crop-extents" _also_ crops a region the same size, it's just a difference of center-crop vs off-center. 

Another issue is that size is a bit ambiguous, do we mean same physical size or same dimensions?

At one point, this option was called "crop-dimensions" instead of "crop-center". 

TBH I don't really like the "crop-extents" option. There are so many way to crop and/or resample, I think we should maybe only support a few simple cases for convenience. If you really need specific extents, just crop it manually yourself _before_ using concatenate, and don't both using `mode`.

Despite the possible ambiguity of "size", I'm thinking we should:

- remove "crop-extents"
- rename "crop-center" to "crop-same-size"
- add a similar "resample-same-size" mode

user27182: > Despite the possible ambiguity of "size", I'm thinking we should:
> 
> - remove "crop-extents"
> - rename "crop-center" to "crop-same-size"
> - add a similar "resample-same-size" mode


Could also name them:

- "resample-match-dimensions"
- "crop-match-dimensions"

Or use "dims" instead of "dimensions" if this seems a bit verbose

user27182: > > Despite the possible ambiguity of "size", I'm thinking we should:
> > 
> > * remove "crop-extents"
> > * rename "crop-center" to "crop-same-size"
> > * add a similar "resample-same-size" mode
> 
> Could also name them:
> 
> * "resample-match-dimensions"
> * "crop-match-dimensions"
> 
> Or use "dims" instead of "dimensions" if this seems a bit verbose

"crop-extents" has been removed, and I settled on "resample-match" and "crop-match". I consider this resolved.

pyvista/core/filters/image_data.py
user27182: See #8025 

This commit seems to do the job to fix this https://github.com/pyvista/pyvista/pull/8025/commits/401aa40fd1a4647025d0914b346d5dc18d9db7ea

It's a simple patch but still needs tests and an example, so will patch it separately.

user27182: @akaszynski the fix from #8025 is ready, would you rather I just patch it into this PR instead?

akaszynski: Yes, if it's easier for you to do it here.

PR #8023, opened 2025-10-15
feat: Add support for creating Rectangle from two diagonal points

tkoyama010:
### Overview

This PR adds the ability to specify a  using two diagonally opposite corner points, making it more intuitive to create axis-aligned rectangles.

### Details

- Add two-point constructor for  where the two points define diagonally opposite corners of the rectangle
- Maintain full backward compatibility with the existing three-point orthogonal vector method
- Update docstring with clear examples demonstrating both use cases
- Add comprehensive tests for the new two-point functionality including edge cases
- Update error messages to reflect that 2 or 3 points are now accepted

### Example Usage

```python
import pyvista as pv

# Create rectangle from two diagonal points (new feature)
pointa = [0.0, 0.0, 0.0]
pointb = [1.0, 1.0, 0.0]
rectangle = pv.Rectangle([pointa, pointb])

# Create rectangle from three orthogonal points (existing feature)
pointa = [1.0, 0.0, 0.0]
pointb = [1.0, 1.0, 0.0]
pointc = [0.0, 1.0, 0.0]
rectangle = pv.Rectangle([pointa, pointb, pointc])
```


MatthewFlamm:
There are an infinite number of rectangles that share two corner points.  It isn't clear to me why this form in the PR was chosen.

tkoyama010:
> There are an infinite number of rectangles that share two corner points. It isn't clear to me why this form in the PR was chosen.

I thought I had a genius idea during my late-night brainwave, but I didn’t realize until I saw this message that PyVista is actually 3D, not 2D :bow:

PR #8024, opened 2025-10-15
Update tox.ini for geovista integration test fix

bjlittle:
Never quite realised that you guys were now explicitly taking control of the `--image_cache_dir` flag within `pyvista` to run the `geovista` integration tests.

This seems like a very fragile contract as I can now easily break you, as per https://github.com/bjlittle/geovista/pull/1754.

TBH this directory name change should be stable. It only changed as I'm currently teeing-up the use of `--doc_mode` with `pytest-pyvista` within `geovista`.

Note that I configure the image cache directory using the `image_cache_dir` INI option within the `pyproject.toml` of `geovista`, so perhaps your use of it is unnecessary? Dunno ...

Anyways, I've realigned the image cache directory name change here in this pull-request.

Apologies for the fallout 🙄

Ping @user27182 @tkoyama010

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -135,7 +135,7 @@ setenv =
 
     _GEOVISTA_ARGS = {env:GEOVISTA_HOME}{/}tests \
         --rootdir={env:GEOVISTA_HOME} \
-        --image_cache_dir={env:GEOVISTA_HOME}{/}tests{/}plotting{/}image_cache
+        --image_cache_dir={env:GEOVISTA_HOME}{/}tests{/}plotting{/}unit_image_cache
 
     _MNE_ARGS = {env:MNE_HOME}{/}mne{/}viz{/}_brain \
         {env:MNE_HOME}{/}mne{/}viz{/}tests{/}test_3d.py \


akaszynski:
> Apologies for the fallout 🙄

I've been saying that for the past month due to the self-hosted runners...

user27182:
> Never quite realised that you guys were now explicitly taking control of the `--image_cache_dir` within `pyvista` to run the `geovista` integration tests.
> 
> 
> 
> This seems like a very fragile contract as I can now easily break you, as per https://github.com/bjlittle/geovista/pull/1754.
> 
> 
> 
> TBH this directory name change should be stable. It only changed it as I'm currently teeing-up the use of `--doc_mode` with `pytest-pyvista` within `geovista`.
> 
> 
> 
> Note that I configure the image cache directory using the `image_cache_dir` INI option within the `pyproject.toml` of `geovista`, so perhaps your use of it is unnecessary? Dunno ...

Thanks for the fix! 

I think ideally you could just remove the CLI flag in tox and have it just work. Maybe we should try this. But maybe it won't work as it may use PyVista's pyproject.toml instead of GeoVista's, depending on how they're installed. 🤷‍♂️

edabor:
> I think ideally you could just remove the CLI flag in tox and have it just work. Maybe we should try this. But maybe it won't work as it may use PyVista's pyproject.toml instead of GeoVista's, depending on how they're installed. 🤷‍♂️

We should do that to prevent future breakage.
And pytest shouldn't be confused to find the `pyproject.toml` since we explicitly pass the `rootdir`:
https://github.com/pyvista/pyvista/blob/b17f378fcdcfbbefe8d1079017869b6b28d39347/tox.ini#L137

PR #8025, opened 2025-10-16
Enable proportional image concatenation with padding

user27182:
### Overview

Follow-up to https://github.com/pyvista/pyvista/pull/8013#discussion_r2430906232


user27182:
This is patched into #8013

PR #8029, opened 2025-10-18
Add `connectivity` region assignment mode

edabor:
### Overview

Add option to specify the region assignment mode for `connectivity` filter.

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -1994,11 +1994,17 @@ class DataSetFilters(_BoundsSizeMixin, DataObjectFilters):
             Strategy used to assign connected region IDs if ``label_regions`` is True.
             Can be either:
 
-            - ``ascending``: IDs are sorted by increasing order of cell count
-            - ``descending``: IDs are sorted by decreasing order of cell counts
-            - ``unspecified``: no particular order
+            - ``"ascending"``: IDs are sorted by increasing order of cell count
+            - ``"descending"``: IDs are sorted by decreasing order of cell counts
+            - ``"unspecified"``: no particular order
 
             .. versionadded:: 0.47
+
+            .. admonition:: ParaView compatibility
+                :class: note dropdown
+
+                The default value ``"descending"`` differs from ParaView's, which
+                is set to ``"unspecified"`` (verified for 5.11 and 6.0 versions).
 
         region_ids : sequence[int], optional
             Region ids to extract. Only used if ``extraction_mode`` is


tests/core/test_dataset_filters.py
```diff
@@ -1130,45 +1136,53 @@ def test_connectivity_scalar_range(
     assert len(conn_with_full_range.array_names) == 3
 
 
-def test_connectivity_all(foot_bones):
+@pytest.mark.parametrize('region_assignement_mode', ['ascending', 'descending', 'unspecified'])
```
user27182:
```suggestion
@pytest.mark.parametrize('region_assignment_mode', ['ascending', 'descending', 'unspecified'])
```

edabor: Added tests for `extraction_mode=specified` in 12654bd77d354da62341d20263df62d358b7d4db and a warning if using `specified` and `unspecified` in d5be20da8c71ed1ba1091f376d6b712d927e236b, since it does not really mean something to use the method using this way. But it still produces a result, so we should allow it.

pyvista/core/filters/data_set.py
```diff
@@ -1989,6 +1990,14 @@ def connectivity(  # type: ignore[misc]  # noqa: PLR0917
             assigned by region cell count in descending order (i.e. the largest
             region has ID ``0``).
 
+        region_assignment_mode : str, default: "descending"
```
edabor, line 1993: BTW, I found that the default value is different than paraview (which is `Unspecified`).

We might need to add such note in the doc ? Because for most (if not all) filters, our defaults are following paraview.

tests/core/test_dataset_filters.py
user27182: Thanks, can you also fix the assignement->assignment typo for the tests? And add versionadded to the docs? LGTM otherwise

pyvista/core/filters/data_set.py
```diff
@@ -1989,6 +1990,14 @@ def connectivity(  # type: ignore[misc]  # noqa: PLR0917
...
+            Strategy used to assign connected region IDs if `label_regions` is True.
```
user27182:
```suggestion
            Strategy used to assign connected region IDs if ``label_regions`` is True.
```

tests/core/test_dataset_filters.py
edabor: Typo was fixed in 606d94f59e4a4a254fef106a51aef41ca0d79015 and versionadded is in a5d099efffbc4ad578d51a504694f5f0fc5198eb

pyvista/core/filters/data_set.py
edabor: Added comment in 7ff837eade2851b651c06cb76febef558eb69e1c

edabor:
> LGTM. If you want to add an extra comment in the docs about the different default compared to ParaView that's also OK.

Done in 7ff837eade2851b651c06cb76febef558eb69e1c

user27182 (approved):

PR #8030, opened 2025-10-18
Set theme as string and add registry

edabor:
### Overview

Currently, `pv.set_plot_theme` accepts strings as theme but not `Plotter` initialization.
This PR allows setting a theme as a string.

In addition to that, it enables registering custom themes for simpler reuse via a `register_theme` function.

### Usage

```python
import pyvista as pv

from pyvista import themes

# Register from instance
theme = themes.DarkTheme()
theme.edge_color = 'red'
theme.show_edges = True

themes.register_theme('my_theme_1', theme)


# Register from class using decorator
@themes.register_theme('my_theme_2')
class MyTheme(themes.DocumentTheme):  # noqa: D101
    def __init__(self) -> None:
        """Initialize the theme."""
        super().__init__()
        self.background = 'red'
        self.show_edges = True


pl = pv.Plotter(theme='my_theme_1')
pl.theme = theme
pl.add_mesh(pv.Sphere())
pl.show()

pl = pv.Plotter(theme='my_theme_2')
pl.add_mesh(pv.Sphere())
pl.show()

pl = pv.Plotter(theme='paraview')
pl.add_mesh(pv.Sphere())
pl.show()
```
<details>
<summary>Show plots</summary>
image
image
image
</details>

### DIFF
--- a/pyvista/plotting/__init__.py
+++ b/pyvista/plotting/__init__.py
@@ -66,8 +66,12 @@ from .texture import image_to_texture as image_to_texture
 from .texture import numpy_to_texture as numpy_to_texture
 from .themes import DocumentTheme as _GlobalTheme
 from .themes import _set_plot_theme_from_env
+from .themes import list_registered_themes as list_registered_themes
 from .themes import load_theme as load_theme
+from .themes import register_theme as register_theme
 from .themes import set_plot_theme as set_plot_theme
+from .themes import unregister_all_themes as unregister_all_themes
+from .themes import unregister_theme as unregister_theme
 from .tools import FONTS as FONTS
 from .tools import check_math_text_support as check_math_text_support
 from .tools import check_matplotlib_vtk_compatibility as check_matplotlib_vtk_compatibility
--- a/pyvista/plotting/_property.py
+++ b/pyvista/plotting/_property.py
@@ -11,6 +11,7 @@ from pyvista.core.utilities.misc import _NoNewAttrMixin
 from . import _vtk
 from .colors import Color
 from .opts import InterpolationType
+from .themes import _get_theme
 
 
 class Property(_NoNewAttrMixin, _vtk.DisableVtkSnakeCase, _vtk.vtkProperty):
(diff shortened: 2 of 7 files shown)


MatthewFlamm:
Have you taken a look at #5119?  This was meant to add part of this functionality.  I lost steam to finish it, and I don't remember what needed to be finished there.

edabor:
> Have you taken a look at #5119? This was meant to add part of this functionality. I lost steam to finish it, and I don't remember what needed to be finished there.

Wow, very sorry, I missed #5119 when starting this PR 🫢
Indeed, #5119 and this PR are overlapping for registering custom theme. The only difference here is that I kept the original class architecture.

To be honest, this PR started from realizing that you could set the plot theme globally with a string via `set_plot_theme("document")` but not at `Plotter` initialization. Which is why I started to implement a private `registry` of themes.

Then, I realized that this `registry` could be useful for users to register their own themes for simpler reuse, thereby making the `registry` public.

Do you see any reason for favoring #5119 over the current PR ? I think it directly addresses https://github.com/pyvista/pyvista/issues/5549 without breaking downstream.

MatthewFlamm:
I have not read this PR in detail.  

We should decide if the idea behind #5119 should move forward, i.e. to make the default themes class methods of `Theme`.  If so, that would likely have breaking changes with respect to the PR here.  Since that one has lingered so long, we can also decide to scrap it and move forward with this.

edabor:
> I have not read this PR in detail.

Sorry, the PR description was not explicit. It is updated now.

edabor:
> We should decide if the idea behind #5119 should move forward, i.e. to make the default themes class methods of `Theme`. If so, that would likely have breaking changes with respect to the PR here. Since that one has lingered so long, we can also decide to scrap it and move forward with this.

I have not used themes a lot in my personal and professional usage of `pyvista`, so I don't feel having sufficient background to tell if #5119 should be moved forward or not.

What could be valuable though would be user's feedback on themes and discussions of which theme feature could be of interest to them. Maybe we should create an open discussion, tagging `geovista` folks ?

user27182:
> > We should decide if the idea behind #5119 should move forward, i.e. to make the default themes class methods of `Theme`. If so, that would likely have breaking changes with respect to the PR here. Since that one has lingered so long, we can also decide to scrap it and move forward with this.
> 
> 
> 
> I have not used themes a lot in my personal and professional usage of `pyvista`, so I don't feel having sufficient background to tell if #5119 should be moved forward or not.
> 
> 
> 
> What could be valuable though would be user's feedback on themes and discussions of which theme feature could be of interest to them. Maybe we should create an open discussion, tagging `geovista` folks ?

I think #5119 got stuck in review, after some API / naming discussions. Though old, if it was updated to resolve merge conflicts I think that PR still has merit and is worth considering.

I am also not a power user of themes though, and have not considered the relative merits of #5119 vs this PR, and agree having additional input is worthwhile. Maybe @bjlittle can comment on this?

MatthewFlamm:
BTW I do like the decorator approach in this PR if we decide to not implement the classmethod approach.

user27182:
Closing since registration was added in #8568 and using string for `Plotter(theme=...)` was added in #8877

PR #8032, opened 2025-10-20
Deprecate setting `theme` attribute to `Plotter`

edabor:
### Overview

While working in #8030, I realized that setting the `theme` attribute to a `Plotter` instance does not have the same effect of setting it at init, especially for global appearance such as `background`. See example below

~This PR raises a warning and updates the doc accordingly.~
This PR deprecates such behavior.

<details>
<summary>Example showing issue</summary>

```python
import pyvista as pv

from pyvista import themes

theme = themes.DarkTheme()
theme.edge_color = 'red'

pl = pv.Plotter()
pl.theme = theme
actor = pl.add_mesh(pv.Sphere(), show_edges=True)
pl.show()

```
image

```python
pl = pv.Plotter(theme=theme)
actor = pl.add_mesh(pv.Sphere(), show_edges=True)
pl.show()
```
image

</details>

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -6245,7 +6245,7 @@ class BasePlotter(_BoundsSizeMixin, PickingHelper, WidgetHelper):
             raise ValueError(msg)
         writer.CompressOff()
         writer.SetFilePrefix(filepath.with_suffix(''))  # type: ignore[arg-type]
-        writer.SetInput(self.render_window)
+        writer.SetRenderWindow(self.render_window)
         modes[extension]()
         writer.SetTitle(title)
         writer.SetWrite3DPropsAsRasterImage(raster)
--- a/.github/workflows/vtk-pre-test.yml
+++ b/.github/workflows/vtk-pre-test.yml
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.9, "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
     steps:
       - uses: actions/checkout@v5
         with:
--- a/pyvista/core/cell.py
+++ b/pyvista/core/cell.py
@@ -699,7 +699,9 @@ class CellArray(
         cells = np.asarray(cells)
         vtk_idarr = numpy_to_idarr(cells, deep=False, return_ind=False)
         self.ImportLegacyFormat(vtk_idarr)
-        imported_size = self.GetNumberOfConnectivityEntries()
+
+        self.ExportLegacyFormat(idarr := _vtk.vtkIdTypeArray())
+        imported_size = _vtk.vtk_to_numpy(idarr).size
 
         # https://github.com/pyvista/pyvista/pull/5404
         if imported_size != cells.size:
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -915,7 +915,8 @@ class PolyData(_PointSet, PolyDataFilters, _vtk.vtkPolyData):
         ... )
 
         """
-        return _vtk.vtk_to_numpy(self.GetVerts().GetData())
+        self.GetVerts().ExportLegacyFormat(arr := _vtk.vtkIdTypeArray())
+        return _vtk.vtk_to_numpy(arr)
 
     @verts.setter
     def verts(self, verts: CellArrayLike) -> None:
@@ -942,7 +943,8 @@ class PolyData(_PointSet, PolyDataFilters, _vtk.vtkPolyData):
         array([10,  0,  1,  2,  3,  4,  5,  6,  7,  8,  9])
 
         """
-        return _vtk.vtk_to_numpy(self.GetLines().GetData()).ravel()
+        self.GetLines().ExportLegacyFormat(arr := _vtk.vtkIdTypeArray())
+        return _vtk.vtk_to_numpy(arr).ravel()
 
     @lines.setter
     def lines(self, lines: CellArrayLike) -> None:
@@ -1014,7 +1016,8 @@ class PolyData(_PointSet, PolyDataFilters, _vtk.vtkPolyData):
         array([3, 0, 1, 2, 3, 3, 2, 1])
 
         """
-        array = _vtk.vtk_to_numpy(self.GetPolys().GetData())
+        self.GetPolys().ExportLegacyFormat(arr := _vtk.vtkIdTypeArray())
+        array = _vtk.vtk_to_numpy(arr)
         # Flag this array as read only to ensure users do not attempt to write to it.
         array.flags['WRITEABLE'] = False
         return array
@@ -1210,7 +1213,8 @@ class PolyData(_PointSet, PolyDataFilters, _vtk.vtkPolyData):
         array([4, 0, 1, 4, 5, 4, 1, 2, 5, 6, 4, 2, 3, 6, 7, 4, 3, 0, 7, 4])
 
         """
-        return _vtk.vtk_to_numpy(self.GetStrips().GetData())
+        self.GetStrips().ExportLegacyFormat(arr := _vtk.vtkIdTypeArray())
+        return _vtk.vtk_to_numpy(arr)
 
     @strips.setter
     def strips(self, strips: CellArrayLike) -> None:
@@ -2069,7 +2073,8 @@ class UnstructuredGrid(PointGrid, UnstructuredGridFilters, _vtk.vtkUnstructuredG
 
         """
         # Flag this array as read only to ensure users do not attempt to write to it.
-        array = _vtk.vtk_to_numpy(self._get_cells().GetData())
+        self._get_cells().ExportLegacyFormat(arr := _vtk.vtkIdTypeArray())
+        array = _vtk.vtk_to_numpy(arr)
         array.flags['WRITEABLE'] = False
         return array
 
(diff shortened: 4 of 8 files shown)


user27182:
Is this a bug? Is the expectation that the background color (and other theme settings) should also be updated when setting the theme? Should we be more explicit with the warning to highlight _all_ theme settings that are affected by this (anti aliasing, parallel projection, depth peeling...)

edabor:
> Is this a bug? Is the expectation that the background color (and other theme settings) should also be updated when setting the theme? 

Hard to tell, because the majority documentation examples for `theme` are:

- mostly modifying the `global_theme` either with `pv.set_plot_theme` or directly by updating it
- showing how to give the theme at `Plotter` init.

But from a user perspective, I guess that if you are passing an object that has the `theme` name at init and modify/reset the  `theme` attribute later, I (but that's only my personal coding experience speaking) would expect to see changes applied globally.

I'd like to hear @akaszynski opinion on that topic since he first introduced `Theme`.

> Should we be more explicit with the warning to highlight _all_ theme settings that are affected by this (anti aliasing, parallel projection, depth peeling...)

Yes, we should, but that requires careful reviewing.

The best option would be to reset global appearances from the newly `theme`. Requires careful testing but doable.
I agree though that this PR is just interim.

MatthewFlamm:
What is the use case of setting the theme after initialization? There will be so many edge and corner cases to document.  Unless there is a good use case, I'd vote for deprecating it, or just raise an error directly.

user27182:
> What is the use case of setting the theme after initialization? There will be so many edge and corner cases to document.  Unless there is a good use case, I'd vote for deprecating it, or just raise an error directly.

Maybe adding two meshes in a loop, e.g. two simulation results, for multiple simulations? Where you want different actor properties for each? Not my use case, but seems reasonable to me.

If it's as simple as refactoring init so that everything is moved to the theme setter, and all we need to do is call the setter in init, it may be fine to support. Or if it's too complex, maybe the setter should just be renamed to reflect what it actually does, maybe `actor_theme`, if that's all it really affects currently?

For now I think adding a warning is fine just to document the status quo.

edabor:
I am also in favor of deprecation, since it can be tedious to maintain a feature that seems niche to me.

Maybe we can issue a new `PendingPyVistaDeprecationWarning` to see if we get feedback from users setting the `theme` attribute, wait one (or more) releases and decide if we go fully deprecating (thereby issuing a `PyVistaDeprecationWarning`).

MatthewFlamm:
A warning is also fine with me, if that is what is preferred.  If we are going to deprecate, I wouldn't complicate it by a pending deprecation.  This adds more developer churn.

user27182:
If there is a PR to deprecate the theme property (and only allow it in plotter init) I would not oppose this. But a warning in the docs until then is fine by me.

edabor:
> If there is a PR to deprecate the theme property (and only allow it in plotter init) I would not oppose this. But a warning in the docs until then is fine by me.

I turned this PR to deprecation accordingly.

tests/plotting/test_theme.py
```diff
@@ -492,16 +493,38 @@ def test_theme_eq():
...
+    match = (
+        'Assigning a theme for a plotter instance is deprecated '
+        'and will removed in a future version of PyVista. '
+        'Set the theme when initializing the plotter instance instead.'
+    )
+
+    with pytest.warns(PyVistaDeprecationWarning, match=match):
+        pl.theme = my_theme
+
+    if pyvista.version_info >= (0, 48):
```
MatthewFlamm: I'm not sure I still understand why, but our contributing guide says that we should use `>` for deprecation version matching if you intended to raise an error in version 48 and remove in version 49.

https://docs.pyvista.org/extras/developer_notes.html#deprecating-features-or-other-backwards-breaking-changes

tests/plotting/test_theme.py
```diff
@@ -492,16 +493,38 @@ def test_theme_eq():
...
+    """Test that the plotter theme is set to the new theme"""
+
     my_theme = pv.plotting.themes.Theme()
     my_theme.color = [1.0, 0.0, 0.0]
     pl = pv.Plotter(theme=my_theme)
     assert pl.theme.color == my_theme.color
     assert pv.global_theme.color != pl.theme.color
 
+
+def test_plotter_set_theme_to_instance():
```
MatthewFlamm: Naming question/suggestion for this test.  This is really about testing the attribute setter than the fact that it is an instance?

examples/02-plot/themes.py
```diff
@@ -166,10 +166,28 @@ def plot_example():
 # Alternatively, set the theme of an instance of ``Plotter``.
 
 pl = pv.Plotter(theme=my_theme)
-# pl.theme = my_theme  # alternatively use the setter
 pl.add_mesh(pv.Cube())
 cpos = pl.show()
 
+# %%
```
MatthewFlamm: The added lines below the one here are referencing a now deprecated functionality.  I would argue we should not add an example for a deprecated feature.

pyvista/plotting/plotter.py
```diff
@@ -459,6 +459,20 @@ def render_window(self) -> _vtk.vtkRenderWindow | None:  # numpydoc ignore=RT01
...
+        .. version-deprecated:: 0.47
+            Assigning the ``theme`` attribute to a plotter object does not affect global appearance
+            settings such as ``background``, which are set at instantiation.
+            To this respect, you need to set the theme such that:
+
+            .. code-block:: python
+
+                import pyvista as pv
+
+                pl = pv.Plotter(theme=theme)
```
MatthewFlamm, line 474: I suggest adding in the old method and the suggested new method.  Something like the below, although I'm not sure it is the clearest format.

```suggestion

                pl = pv.Plotter()
                pl.theme = theme
                # change above lines to
                pl = pv.Plotter(theme=theme)
```

tests/plotting/test_theme.py
MatthewFlamm: If we keep these version numbers for raising error (48) and for removal (49).  We'd issue a warning for only1 version.  Should we make it raise on 49 and remove on 50 to get 2 versions with warnings?

edabor: I guess the number of versions with warnings depends on the method usage frequency.
Hard to tell, might be low since documentation mostly focuses on setting the theme to a plotter at init or globally.

But let's be safe and raise warnings for 2 versions (ie 47, 48), raise error on 49 and remove in 50.

examples/02-plot/themes.py
edabor: Agreed 👍

tests/plotting/test_theme.py
edabor: Feel free to propose another test name 😄

MatthewFlamm: `test_plotter_theme_attribute_setter`?

tests/plotting/test_theme.py
MatthewFlamm: I think 2 versions is a good all-purpose number unless others feel otherwise.

examples/02-plot/themes.py
edabor: Done in f7659c7a580d9a6eb7e855d4943deb1e16c2864f and 613727f336bc7143c2bf5aac22418871c26703bd

tests/plotting/test_theme.py
edabor: Done in ffc669e32a6c9f5054025ff2cfc580f6aa96492c

tests/plotting/test_theme.py
edabor: Done in ddc4dc559eb027b495aca9bc1a17b1365ae8e818

pyvista/plotting/plotter.py
```diff
@@ -471,7 +488,7 @@ def theme(self) -> Theme:  # numpydoc ignore=RT01
         >>> import pyvista as pv
         >>> from pyvista import themes
         >>> pl = pv.Plotter()
-        >>> pl.theme = themes.DarkTheme()
+        >>> pl.theme = themes.DarkTheme()  # doctest: +IGNORE_WARNINGS
```
MatthewFlamm: `pytest-doctestplus` isn't working because it isn't enabled in the pytest options (or invoked as a flag on the invocation, etc.).  But, adding this will fundamentally change our doc test for the whole package.  This would have potential unforeseen issues.

I have two ideas to fix this issue.  Since it is deprecated and we have tests for that, it isn't that crucial to have an additional test in the documentation.
1) remove the example entirely.
2) skip the example using `+SKIP`

MatthewFlamm: I would likely do the latter

edabor: Done in dfb66c76d0946c48193d5ce6f1b092bfb71bd7de

tests/plotting/test_theme.py
```diff
@@ -492,16 +493,38 @@ def test_theme_eq():
...
+    match = (
+        'Assigning a theme for a plotter instance is deprecated '
+        'and will removed in a future version of PyVista. '
+        'Set the theme when initializing the plotter instance instead.'
+    )
+
+    with pytest.warns(PyVistaDeprecationWarning, match=match):
+        pl.theme = my_theme
+
+    if pyvista.version_info >= (0, 49):
```
MatthewFlamm, line 520: @tkoyama010 I think the requirements for this check are driven by the release process.  Do we need to have a `>` here, or is the `>=` OK?

akaszynski: Above we have "convert to error in v0.49", so `>=` should be ok.

PR #8018, opened 2025-10-13
Update `cyclopts` version to v4

edabor:
### Overview

~Anticipating `cyclopts` v4 by testing the beta
https://github.com/BrianPugh/cyclopts/releases/tag/v4.0.0b1~

Updating to released `cyclopts` v4

PR #8033, opened 2025-10-20
Pin `cyclopts` for dev

user27182:
### Overview

Cyclopts 4.0 was released today and it broke our CLI, including `pyvista report`. This PR pins it for dev purposes so that `--group test` still works.

cc @edabor

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,9 +161,8 @@ docs = [
 docs-test = [
   'pytest-pyvista[vtksz]==0.3.1',
   'pytest-xdist<3.9.0',
-
   { include-group = "docs-trame-vtk" },
-
+  { include-group = 'pinned' },
 ]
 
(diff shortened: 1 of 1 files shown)


edabor:
I'll try to finish https://github.com/pyvista/pyvista/pull/8018 asap

edabor:
https://github.com/pyvista/pyvista/pull/8018 is ready, we might not need the current PR

user27182:
> https://github.com/pyvista/pyvista/pull/8018 is ready, we might not need the current PR

Maybe not, though I still think it's good to pin all dependencies for test purposes to so that updates to 3rd-party packages don't break CI unexpectedly (this is partly the reason why we have a pinned group at all).

Much better to have dependabot push a PR with a bump and see any failures there instead.

edabor (dismissed):

PR #8039, opened 2025-10-22
Fix obb tree cache

user27182:
### Overview

Fix #8035 

Should we backport this to 0.46?

PR #8041, opened 2025-10-26
Fix obb tree cache (#8039) backport to v0.46

user27182:
Cherry-pick #8039 into release/0.46 branch

### DIFF
--- a/pyvista/_version.py
+++ b/pyvista/_version.py
@@ -16,7 +16,7 @@ Denotes the first release candidate.
 # major, minor, patch
 from __future__ import annotations
 
-version_info = 0, 46, 3
+version_info = 0, 46, 4
 
 # Nice string for the version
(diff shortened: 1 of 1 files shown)


user27182:
The pre-commit failure can be ignored, this was an issue with `blackdoc` that was patched in main

akaszynski:
> The pre-commit failure can be ignored, this was an issue with `blackdoc` that was patched in main

Checked and confirmed: Release step will still run since it only needs `Linux, windows, macOS` workflows to complete.

akaszynski:
@user27182, do we want to release a patch now?

user27182:
Yes, should be good to go

akaszynski:
> Yes, should be good to go

Released https://pypi.org/project/pyvista/0.46.4/

Thanks @user27182!

PR #8045, opened 2025-10-31
Use `CameraPosition` explicitly in docs

user27182:
### Overview

Split from #7794. 

Camera position can be confusing to users, since using a list with 3 vectors is ambiguous as it's not immediately obvious which vector is which. This PR replaces all uses of a `list[tuple]` format in the docs with an explicit `CameraPosition` object instead. This PR does _not_ modify `CameraPosition` itself in any way.


user27182:
This PR highlights the downside to using a hash of the docstring text to name the images generated by the plot directive. None of the actual images are affected by this PR, but since the docstrings are modified, I have to manually rename 170 files to update the hashes ☹️

tkoyama010 (approved):
LGTM

PR #8051, opened 2025-11-01
Add `Transform` component properties

user27182:
### Overview
Add new `Transform` properties:

* ~position~ translation
* rotation_axis_angle
* rotation_matrix
* reflection
* scale_factors
* shear_matrix
* has_translation
* has_rotation
* has_reflection
* has_scale
* has_shear

Many of these properties rely on `decompose`, so the output from that method is also cached internally with this PR


pyvista/core/utilities/transform.py
```diff
@@ -2336,7 +2388,9 @@ def _get_point_translations(
             point_array = _validation.validate_array3(point, dtype_out=float, name='point')
             translate_away = Transform().translate(-point_array)
             translate_toward = Transform().translate(point_array)
-            if multiply_mode == 'post' or self._multiply_mode == 'post':
+            if multiply_mode == 'post' or (
+                multiply_mode is None and self._multiply_mode == 'post'
+            ):
```
user27182, line 2394: Unrelated to this PR, but seems like a bug not to check for `None`

pyvista/core/utilities/transform.py
```diff
@@ -2359,6 +2413,158 @@ def check_finite(self: Transform) -> bool:  # numpydoc ignore: RT01
     def check_finite(self: Transform, value: bool) -> None:
         self._check_finite = bool(value)
 
+    @property
+    def position(self) -> tuple[float, float, float]:
```
user27182: The name `position` is used to follow the vtk convention from `GetPosition`. But this creates some inconsistency with `translate` method and `has_translation` property. Should we rename this to `translation`?

.github/workflows/docs.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Set environment for self-hosted
         if: runner.environment == 'self-hosted'
         run: |
-          echo "PARALLEL=-j4" >> $GITHUB_ENV
+          echo "PARALLEL=-j2" >> $GITHUB_ENV
```
akaszynski: Instability?

pyvista/core/utilities/transform.py
```diff
@@ -2359,6 +2414,302 @@ def check_finite(self: Transform) -> bool:  # numpydoc ignore: RT01
...
+        --------
+        has_translation, translate, decompose
+
+        Examples
+        --------
+        Compose a translation and get the position.
+
+        >>> import pyvista as pv
+        >>> trans = pv.Transform() + (1, 2, 3)
+        >>> trans.position
```
akaszynski: Yep, translation sounds better here. I don't like how this is exposing `GetPosition` through the python property `translation`, but it makes more sense here and the VTK getter/setter should be `GetTranslation` / `SetTranslation`.

user27182: Ok, will rename it to `translation`. For the implementation, we can easily get it with `return self.matrix[:3, 3]` instead if you prefer that over `GetPosition` ? `GetPosition` would be _marginally_ faster since we avoid `array_from_vtkmatrix` calls...

or similarly, get it with:
``` py
T, _, _, _, _ = self.decompose()
return T
```

.github/workflows/docs.yml
user27182: Yes, the build was successful just a few hours ago after the first commit: [e2ae396](https://github.com/pyvista/pyvista/pull/8051/commits/e2ae3967ecbd01b74c8fa79e34adc5ac983f341a)

But the last few commits have failed, e.g. this build was stalled for over 1hr:
https://github.com/pyvista/pyvista/actions/runs/19005306467/job/54277878850?pr=8051

Maybe this is caused by adding examples? [c23d271](https://github.com/pyvista/pyvista/pull/8051/commits/c23d27196e1102eedd37e0bb51a99a6bea78c269)

akaszynski: The docbuild taking 55 minutes isn't great, but a working build is better than no build. I'm fine with merging this as-is, but we'll have to debug the docbuild and try to get up to at least four threads.

user27182: It's a bit of a baffling/unhelpful error too, hard to figure out what the issue is. 

I can try debugging by (1) removing the examples sections, and (2) removing see also sections?

Not sure if it's the plot directive getting tripped up by the extra examples, or if there are more general scalability issues with sphinx and not able to handle the docs generally growing in volume?

user27182: This has been reverted. Looks like j4 is working after all.

akaszynski: My guess it's the double rendering that's still happening when getting image depth.

pyvista/core/utilities/transform.py
```diff
@@ -2359,6 +2414,302 @@ def check_finite(self: Transform) -> bool:  # numpydoc ignore: RT01
...
+        --------
+        Compose a translation and get the translation component.
+
+        >>> import pyvista as pv
+        >>> trans = pv.Transform() + (1, 2, 3)
+        >>> trans.translation
+        (1.0, 2.0, 3.0)
+
+        Compose a second translation and get the component again.
+        >>> trans += (4, 5, 6)
```
akaszynski, line 2438: Example isn't rendering correctly. See:
https://6907fbe893c29a2ed866e61f--pyvista-dev.netlify.app/api/utilities/_autosummary/pyvista.transform.translation

Needs a new line:
```suggestion
        Compose a second translation and get the component again.

        >>> trans += (4, 5, 6)
```

akaszynski (approved):
One minor nitpick, otherwise LGTM.

PR #8055, opened 2025-11-03
Add extra software reports for integration tests

user27182:
### Overview

The integration tests have different dependencies from `pyvista`. It's sometimes helpful to see the specific versions of those dependencies. This PR uses `scooby` to generate extra reports.


user27182:
the mne report doesn't work, it generates something completely different:

https://mne.tools/stable/generated/mne.Report.html#mne.Report

`scooby` should be updated to only print `Report` if it's an instance of `scooby.Report`

user27182:
> the mne report doesn't work, it generates something completely different:
> 
> https://mne.tools/stable/generated/mne.Report.html#mne.Report
> 
> `scooby` should be updated to only print `Report` if it's an instance of `scooby.Report`

This is fixed in https://github.com/banesullivan/scooby/pull/139. For now we can just use `mne sys_info` which is similar enough to a Scooby Report.

PR #8056, opened 2025-11-03
Remove reference and use of vtk-mesa

banesullivan:
I'm not 100% this is exactly how we want to handle offscreen use of VTK, but vtk-mesa on Kitware's custom index has been deprecated and it ought to be removed now that OSMesa is better supported in VTK 9.5+


akaszynski:
> I'm not 100% this is exactly how we want to handle offscreen use of VTK

EGL works out of the box, no?

banesullivan:
Yes, EGL also works out of the box with the PyPI 9.5 wheel (assuming you have the right system libraries installed)

akaszynski:
> Yes, EGL also works out of the box with the PyPI 9.5 wheel (assuming you have the right system libraries installed)

The error from VTK is actually pretty good. Let's just add the docs about usage with EGL and call it a day.

user27182:
We should change 9.5 to 9.4 in this PR, no? 

From https://docs.vtk.org/en/latest/advanced/available_python_wheels.html :

> Changed in version 9.4: As of VTK 9.4, OSMesa and EGL support are included by default in the vtk wheels and can be selected at runtime. The vtk-osmesa wheels are no longer provided, and it is no longer necessary to install vtk-osmesa.

user27182:
And maybe provide a reference to the vtk docs here:

https://docs.vtk.org/en/latest/advanced/runtime_settings.html

So that users know about the `VTK_DEFAULT_OPENGL_WINDOW` env var for selecting the backend

MatthewFlamm:
I'm +1 for the change.  Does the changes to Docker files, fix the security alert?  I'm not sure how this works

docker/jupyter.Dockerfile
```diff
@@ -10,9 +10,6 @@ WORKDIR /build-context/
 
 RUN pip install --no-cache-dir pyvista*.tar.gz
 RUN pip install -r requirements.txt
-# Install vtk-osmesa wheel
-RUN pip uninstall vtk -y
-RUN pip install --no-cache-dir --extra-index-url https://wheels.vtk.org vtk-osmesa
```
user27182, line 15: Removing this line should also fix the failures we are seeing on main branch:
https://github.com/pyvista/pyvista/actions/runs/19233531541/job/54977499595

akaszynski: @banesullivan, if it's fine with you I'd like to get this PR read for review to fix the build on `main`.

user27182: Turns out the failures are caused by new vtk website security introduced last week https://discourse.vtk.org/t/trouble-in-pip-installing-vtk-osmesa/14890/13

hakostra:
Just want to mention that as of the next release, VTK 9.6, the X dependencies are also dynamically loaded (if available), allowing users to use off-screen rendering in environments without any X libraries, without having to compile VTK from sources. The former vtk-osmesa package did not have X dependencies either, so VTK 9.6 will be the first VTK package where you can drop it in "any" environment (software/hardware rendering and with and without X) and expect it to work.

This can be tested in the VTK nightly builds already.

akaszynski:
> ...Does the changes to Docker files, fix the security alert?

It should. If not I'll resolve it in a separate PR.

PR #8059, opened 2025-11-04
Update `lychee` config to support local use

user27182:
### Overview

- Move `lychee` args from the workflow to `lychee.toml` so that running `lychee .` locally works exactly the same as it does in CI
- Remove old references to the sphinx linkcheck build
- Use the `extensions` option to search for files, rather than specify files manually as inputs
- Fix broken links and remove OpenFOAM example from the external gallery (link no longer works and I could not find a working alternative)

### DIFF
--- a/lychee.toml
+++ b/lychee.toml
@@ -14,6 +14,5 @@ exclude = [
 exclude_all_private = true # exclude 'localhost', '127.0.0.1', etc.
 exclude_path = ['pyvista/core/utilities/docs.py']
 extensions = ['md', 'py', 'rst', 'txt']
-fail = false
 user_agent = 'curl/7.83.1'
 verbose = 'info'


user27182:
The linkcheck workflow runs on a schedule, so I triggered it manually from this branch, and it's passing:
https://github.com/pyvista/pyvista/actions/runs/19085944125/job/54526049361

CONTRIBUTING.rst
```diff
@@ -1525,16 +1525,14 @@ created the following will occur:
...
-    no results are cached.
+#.  Locally test and build the documentation. Be sure to run ``make clean``
+    to ensure no results are cached.
 
     .. code-block:: bash
 
        cd doc
        make clean  # deletes the sphinx-gallery cache
        tox run -e doctest-modules
-       make html -b linkcheck
```
user27182, line 1537: We need to keep this line. Maybe replace with `make html` ?

akaszynski: Yep, just `make html`, or better yet, use tox in `tox -e docs-build`.

PR #8064, opened 2025-11-06
Add Methods and Attributes headings to class docs

user27182:
### Overview

`.. rubric::` is currently used for adding Methods and Attributes sections to class docs. This is fine, but the sections are not hyperlinked, which means they don't show up in the Contents side bar.

This PR changes the rubric to a heading instead so that it's hyperlinked.

#### `main` branch

image

#### This PR

image

One downside of this change is that headings cannot be indented, and so cannot be nested inside of the `.. autoclass::` directive. This makes is so that `.. autosummary::` is no longer tied directly to the class, and hence references to class attributes such as ``` :attr:`matrix` ``` do not resolve anymore. The fix is to explicitly reference the class if you want to link to class attributes or methods directly in the docstring's summary line

### DIFF
--- a/pyvista/plotting/prop3d.py
+++ b/pyvista/plotting/prop3d.py
@@ -397,7 +397,7 @@ class Prop3D(
         *,
         inplace: bool = False,
     ):
-        """Apply a transformation to this object's :attr:`user_matrix`.
+        """Apply a transformation to this object's :attr:`~Prop3D.user_matrix`.
 
         .. note::
 


pyvista/core/utilities/transform.py
```diff
@@ -2398,25 +2399,27 @@ def _get_point_translations(
         return None, None
 
     @property
-    def check_finite(self: Transform) -> bool:  # numpydoc ignore: RT01
-        """Check that the :attr:`matrix` and :attr:`inverse_matrix` have finite values.
+    def check_finite(self: Transform) -> bool:  # numpydoc ignore: RT01,SS06
+        """Check that the :attr:`~Transform.matrix` and :attr:`~Transform.inverse_matrix` have
+         finite values.
```
user27182: This PR breaks refs in summary lines, and :attr:`matrix` must be replaced with :attr:`~Transform.matrix` for autosummary to work. But adding the extra characters pushes many summary lines over the maximum allowed line length (ruff E501 error). 

We could ignore E501, but that would ignore it for the entire docstring, not just one line. Instead, I think making the summary line two lines is better, and then we need to ignore numpydoc SS06 and ruff D205 instead, while keeping E501 intact for the rest of the docstring.

For these summary lines I felt that it was important to keep the ref to the `matrix` attribute so that it's not confused with any other general transformation matrix.

edabor: Do you see that on your side ?
https://690ce0c021f8f8c0bbbb73c0--pyvista-dev.netlify.app/api/utilities/_autosummary/pyvista.transform.check_finite#pyvista.Transform.check_finite
image

user27182: yes. That looks awful, I guess it _must_ be one line to render correctly. The autosummary links work, but it's also cut-off:

image

I guess it's better to ignore E501 then, or remove the ref / shorten the docstring?

edabor: +1 for ignoring E501

edabor: Though it is weird that other ignored `# noqa: D205` docstrings are rendering fine ... 🤔

user27182: Fixed now:
https://690d2b0aa2f26a76a954662a--pyvista-dev.netlify.app/api/utilities/_autosummary/pyvista.transform#attributes

edabor (approved):
LGTM

Issue #1066, opened 2020-12-21
Better spline interpolation


sampotter:
From what I can tell, to interpolate samples from a smooth curve, there are two options: `PolyLine` and `Spline`. It's not clear from the documentation what `Spline` is doing. It would be nice to have more control over spline interpolation. For example, if I have a sequence of points and tangent vectors, I'd like to be able to construct piecewise Hermite spline using something like: `spline = pv.Spline(points, tangents, interp='cubic')`.

whophil:
Right now, pyvista's `Spline` is limited to what `vtkParametricSpline` can do. (https://vtk.org/doc/nightly/html/classvtkParametricSpline.html) The ability to specify tangents at the ends is available but is currently not exposed - it certainly could be. `vtkParametricSpline` provides no option to specify internal tangents.

It is easy enough to use SciPy splines, evaluate them at given parameter values, and store the coordinates into a `PolyData` - this is what `pv.Spline()` does anyway.

banesullivan:
Thanks for chiming in, @whophil!

I'm going to expose those derivative values in the `pv.Spline()` function and call this closed. With the solution being to use scipy for this sort of spline interpolation

banesullivan:
On second look, there are quite a few methods of `vtkParametricSpline` we should wrap in that function

```
'SetClockwiseOrdering',
 'SetDerivativesAvailable',
 'SetJoinU',
 'SetJoinV',
 'SetJoinW',
 'SetLeftConstraint',
 'SetLeftValue',
 'SetMaximumU',
 'SetMaximumV',
 'SetMaximumW',
 'SetMinimumU',
 'SetMinimumV',
 'SetMinimumW',
 'SetParameterizeByLength',
 'SetRightConstraint',
 'SetRightValue',
 'SetTwistU',
 'SetTwistV',
 'SetTwistW',
 'SetXSpline',
 'SetYSpline',
 'SetZSpline',
```

tletut0ur:
Hello, 
I do not know if this is not too old of an issue to fix, if so please let me know. I am currently attempting to add those methods to the `Spline` method of `parametric_objects`. Closing the spline and changing constraint types at left and right side seem to work.
For methods: 
`'SetClockwiseOrdering',
 'SetDerivativesAvailable',
 'SetJoinU',
 'SetJoinV',
 'SetJoinW',
 'SetMaximumU',
 'SetMaximumV',
 'SetMaximumW',
 'SetMinimumU',
[...]

PR #8072, opened 2025-11-07
Add more arguments to spline

tletut0ur:
### Overview

Expose more arguments of Spline. Related to #1066 

### Details

- Closing of spline is now possible.
- Closing and derivative types are exposed.
-  X, Y, and Z Cardinal spline support should (I think) be discussed more and are not implemented here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ jupyter = [
 
 [dependency-groups]
 pinned = [ # Pinned versions of core dependencies
-  'cyclopts<4.3.0',
+  'cyclopts<4.4.0',
   'matplotlib<3.10.8',
   'numpy<2.4.0',
   'pillow<12.1.0',
@@ -67,7 +67,7 @@ typing = [
   'mypy<1.19.0',
   'npt-promote==0.2',
   'numpy>=2.0.0',
-  'trimesh<4.10.0',
+  'trimesh<4.11.0',
   { include-group = 'pinned' },
 ]
 
@@ -76,14 +76,14 @@ test = [
   'cmocean<4.0.4',
   'colorcet<3.2.0',
   'embreex<2.17.8; sys_platform != "darwin" or platform_machine != "arm64"', # Does not work with arm-based macs
-  'hypothesis<6.142.6',
+  'hypothesis<6.148.3',
   'imageio-ffmpeg<0.7.0',
   'imageio<2.38.0',
   'ipython<10.0.0',
   'ipywidgets<9.0.0',
   'libcst<1.8.7',
   'meshio<5.4.0',
-  'nest-asyncio2<1.7.0',
+  'nest-asyncio2<1.8.0',
   'numpydoc<1.10.0',
   'pandas<=2.4.0',
   'pyobjc-framework-Cocoa; sys_platform == "darwin"', # see pyvista/pulls/7828
@@ -105,7 +105,7 @@ test = [
   'trame-vtk>=2.5.8,<2.9.1',
   'trame-vuetify>=2.3.1,<3.2.0',
   'trame>=2.5.2,<3.13.0',
-  'trimesh<4.10.0',
+  'trimesh<4.11.0',
   'typing-extensions<4.16.0',
   { include-group = 'pinned' },
 ]
@@ -122,15 +122,15 @@ docs = [
   'colour-science==0.4.6',
   'enum-tools==0.13.0',
   'imageio-ffmpeg==0.6.0',
-  'imageio==2.37.0',
+  'imageio==2.37.2',
   'joblib==1.5.2',
   'jupyter-sphinx==0.5.3',
-  'jupyterlab==4.4.10',
+  'jupyterlab==4.5.0',
   'lxml==6.0.2',
   'meshio==5.3.5',
   'mypy-extensions==1.1.0',
   'mypy==1.18.2',
-  'nest-asyncio2==1.6.3',
+  'nest-asyncio2==1.7.1',
   'numpydoc==1.9.0',
   'pydata-sphinx-theme==0.15.4', # Do not upgrade, see https://github.com/pyvista/pyvista/pull/7095
   'pypandoc==1.16.2',
@@ -153,7 +153,7 @@ docs = [
   'sympy==1.14.0',
   'trame-vuetify==3.1.0',
   'trame==3.12.0',
-  'trimesh==4.9.0',
+  'trimesh==4.10.0',
   'vtk-xref==0.1.1',
   { include-group = "docs-trame-vtk" },
   { include-group = 'pinned' },


akaszynski:
Looks good so far. You should consider adding additional examples and test coverage for this.

akaszynski:
Also, feel free ignoring `PLR0917` with a `noqa`, but also consider using keyword only arguments for the remainder of the new boolean arguments. See https://peps.python.org/pep-3102/.

tletut0ur:
I think that exposing `**kwargs` to `Spline`  also adds the possibility of a `direction` and `normal` argument, which get picked up by the `test_plotting.py` unit test, particularly in `_generate_direction_object_functions`. Is it okay for me to add an exception to this unit test by removing Spline from the list of collected parametric functions ?
Thanks

user27182:
> I think that exposing `**kwargs` to `Spline` also adds the possibility of a `direction` and `normal` argument, which get picked up by the `test_plotting.py` unit test, particularly in `_generate_direction_object_functions`. Is it okay for me to add an exception to this unit test by removing Spline from the list of collected parametric functions ? Thanks

Yes `Spline` should not be included in this test suite. `surface_from_para` doesn't have a direction or normal arg, so this is odd. If you try `Spline(point, direction=(0,0,1))` what error do you get? I am thinking the test parametrization assumes that the object has `direction` or `normal`, _unless_ there is a `TypeError` with specific message. I am guessing that the `Spline` case produces an error, but the error doesn't quite match what's expected by the test suite, and so is being erroneously included in the list.

Effectively, I think this is a bug with how the tests are generated. If you can find a fix such that `Spline` is excluded please go ahead, otherwise I will look into it.

tletut0ur:
I get `TypeError: surface_from_para() got an unexpected keyword argument 'direction'`. The issue, I think, could come from function `_has_param` of the same file.

user27182:
The vtksz image errors are unrelated to this PR, it's a bug with pytest pyvista

user27182:
> The vtksz image errors are unrelated to this PR, it's a bug with pytest pyvista 

The errors also can't be ignored and need to be addressed. 

I'm not exactly sure what the cause is, but the actual dimensions of generated static images aren't matching up with the corresponding vtksz images. Seems to maybe depend on the file hash changing, might affect how the static image file names are matched up to the corresponding interactive image file names

As a workaround we can also update the affected images to make the tests pass until this is fixed

tletut0ur:
@user27182 Thanks for the insights. I am building the doc on windows. Could this cause any rendering issues that do not line up with what would be expected by building the doc on linux ? 
Thanks,

akaszynski:
> @user27182 Thanks for the insights. I am building the doc on windows. Could this cause any rendering issues that do not line up with what would be expected by building the doc on linux ?
> Thanks,

Yes, a simple way to ensure you have the right figures (build from linux) is to simply pull down the files generated from the workflow. For example, the artifact "[doc-failed-test-images](https://github.com/pyvista/pyvista/actions/runs/19390610460/artifacts/4577336043) from:
https://github.com/pyvista/pyvista/actions/runs/19390610460

Then, simply move the generated images into the documentation cache. This way you don't even have to build the docs locally.

pyvista/core/utilities/parametric_objects.py
```diff
@@ -21,19 +21,61 @@
...
+
+    left_derivative_value : float, optional
+        Value of derivative on left side.
+
+    right_constraint_type : int, optional
+        Derivative constraint type at the right side. Must be 0, 1, 2, or
+        3. See ``left_constraint_type`` description.
+
+    right_derivative_value : float, optional
+        Value of derivative on left side.
```
user27182: IMO this API is confusing.

- using magic int values for the constraint type is non-obvious, e.g. setting `left_constraint_type=0` is meaningless on its own and requires looking at the docs to understand. I think we should deviate from the VTK API and use named literal strings. e.g. `'finite_difference'`, `'clamped'`, `'second'`, `'scaled_second` (feel free to come up with better names).
- There is a disconnect between `constraint_type` and `derivative_value` parameter names. These appear directly linked, so IMO these have the same name, either both use `derivative` or both use `constraint`.
- The `0` constraint type does not use any value, but the other types do. This should be documented, and possibly raise an error if the value is not None when type is `0`. And if the other types _require_ a value to work, then we should also raise an error if the type is `1`, `2`, or `3` but the value is `None`. Basically, it looks like it's currently possible to set type/value combinations which have no effect on the output. We should make it so that only valid combinations are allowed.
- Is it common to set _both_ right and left? Or is it typically desirable to set one but not the other? If they're usually both set, then perhaps we should consider a unified approach, where there is no separate left and right, but instead you do, e.g.  `constraint_type` and `constraint_value`, and allow setting a single value (in which case both left and right have this value), or allow setting _two_ values as a tuple to set left and right separately.

In summary, we should:

1. Use named strings, not integers
2. Ensure a value is set (or not) depending on the constraint type used
3. Consider unifying both left and right into a single parameter

pyvista/core/utilities/parametric_objects.py
```diff
@@ -21,19 +21,61 @@
     from pyvista.core._typing_core import VectorLike
 
 
-def Spline(points: VectorLike[float] | MatrixLike[float], n_points: int | None = None) -> PolyData:
+def Spline(
+    points: VectorLike[float] | MatrixLike[float],
+    n_points: int | None = None,
+    *,
+    closed: bool = False,
+    parameterize_by_length: bool | None = None,
```
user27182: Does `None` here do something different depending on other parameters? If not, we should just set it to `True` or `False` (whatever the current VTK default is).

Looking at the docstring "Parametrize by length rather than point index", it seems that instead of a bool, it may be more appropriate to use a string, e.g. `parametrize_by` and have `'length'` and `'index'` as the two options.

pyvista/core/utilities/parametric_objects.py
akaszynski: Strings, or enumerators. There should also be a default.

pyvista/core/utilities/parametric_objects.py
akaszynski: I think that there's an implicit default within VTK, and we should be explicit about this with a string.

pyvista/core/utilities/parametric_objects.py
tletut0ur: @user27182 @akaszynski thank you for your feedback. 
> Strings, or enumerators. There should also be a default.

For the default value of constraint_type, `'finite_difference'` (0) would be backward compatible and correspond to the default of VTK I think. 

> There is a disconnect between constraint_type and derivative_value parameter names. These appear directly linked, so IMO these have the same name, either both use derivative or both use constraint.

True, but isn't `derivative_value` a bit vague in the context of splines ? How about something like `boundary_constraint_type`, `boundary_value` ? 

>  Consider unifying both left and right into a single parameter 

How about we implement three arguments: `constraint_type`,  `left_constraint_type`,  `right_constraint_type`  (similar to Pandas `merge` function with `on`, `left_on`, `right_on` ) This allows for global setting of constraint types and values but keeps the door open for individual setting.  This would also applies to the boundary derivative values.

user27182: > > There is a disconnect between constraint_type and derivative_value parameter names. These appear directly linked, so IMO these have the same name, either both use derivative or both use constraint.
> 
> 
> 
> True, but isn't `derivative_value` a bit vague in the context of splines ? How about something like `boundary_constraint_type`, `boundary_value` ? 

I like `boundary`, makes things explicit. Could even drop `type`, just have `boundary_constraint` and `boundary_value`. 


> >  Consider unifying both left and right into a single parameter 
> 
> 
> 
> How about we implement three arguments: `constraint_type`,  `left_constraint_type`,  `right_constraint_type`  (similar to Pandas `merge` function with `on`, `left_on`, `right_on` ) This allows for global setting of constraint types and values but keeps the door open for individual setting.  This would also applies to the boundary derivative values.

There are similar APIs in PyVista, e.g. `labels` vs `x_label`, `y_label`, `z_label` for setting all XYZ axis labels at once vs setting them individually. I think this OK, but we should make sure users can't set both at the same time.

pyvista/core/utilities/parametric_objects.py
```diff
@@ -21,19 +21,55 @@
     from pyvista.core._typing_core import VectorLike
 
 
-def Spline(points: VectorLike[float] | MatrixLike[float], n_points: int | None = None) -> PolyData:
+def Spline(
+    points: VectorLike[float] | MatrixLike[float],
+    n_points: int | None = None,
+    *,
+    closed: bool = False,
+    parameterize_by: str = 'length',
```
user27182: NIT: I think there are multiple valid spellings of `parameterize`, but our tests use a different spelling `parametrize`, e.g. `@pytest.mark.parametrize`, so I suggest using the same spelling for consistency within PyVista

pyvista/core/utilities/parametric_objects.py
tletut0ur: After a few implementation attempts I think that the tuples is the way to go. I'd like for users to be able to set values implicitly with a single value but it seems to produce some errors with type checking. Tuples it is for now.

user27182: > After a few implementation attempts I think that the tuples is the way to go. I'd like for users to be able to set values implicitly with a single value but it seems to produce some errors with type checking. Tuples it is for now.

I was able to modify the code and make it work with single inputs for the constraint and/or value. The default value is now also `None`, in the sense that it's unset, and defaults to `(None, None)` for 'finite_difference' and `(0.0, 0.0)` otherwise.


You can use this for the function definition:

<details>

``` py
from collections.abc import Sequence
from typing import Literal
from typing import get_args

_ParametrizeByOptions = Literal['length', 'index']
_BoundaryConstraintOptions = Literal['finite_difference', 'clamped', 'second', 'scaled_second']


def Spline(
    points: VectorLike[float] | MatrixLike[float],
    n_points: int | None = None,
    *,
    closed: bool = False,
    parametrize_by: _ParametrizeByOptions = 'length',
    boundary_values: float | tuple[float | None, float | None] | None = None,
    boundary_constraints: _BoundaryConstraintOptions
    | tuple[_BoundaryConstraintOptions, _BoundaryConstraintOptions] = 'clamped',
    **kwargs,
) -> PolyData:

```

</details>

And update the function body with this code instead:

<details>

``` py
    # Validate inputs
    def check_constraint(value: str) -> None:
        _validation.check_contains(
            get_args(_BoundaryConstraintOptions), must_contain=value, name='boundary_constraints'
        )

    points_ = _validation.validate_arrayNx3(points, name='points')
    _validation.check_contains(
        get_args(_ParametrizeByOptions), must_contain=parametrize_by, name='parametrize_by'
    )

    # Ensure we have valid constraint, value pairs
    _validation.check_instance(boundary_constraints, Sequence, name='boundary_constraints')
    if isinstance(boundary_constraints, str):
        check_constraint(boundary_constraints)
        constraints_pair = (boundary_constraints, boundary_constraints)
    else:
        _validation.check_length(boundary_constraints, exact_length=2, name='boundary_constraints')
        check_constraint(boundary_constraints[0])
        check_constraint(boundary_constraints[1])
        constraints_pair = boundary_constraints

    if boundary_values is None:
        values_list: list[float | None] = []
        for constraint in constraints_pair:
            values_list.append(None if constraint == 'finite_difference' else 0.0)
        values_pair: Sequence[float | None] = values_list
    else:
        _validation.check_instance(
            boundary_values, (Sequence, float, type(None)), name='boundary_values'
        )
        if isinstance(boundary_values, Sequence):
            _validation.check_length(boundary_values, exact_length=2, name='boundary_values')
            values_pair = boundary_values
        else:
            values_pair = (boundary_values, boundary_values)

    spline_function = _vtk.vtkParametricSpline()
    spline_function.SetPoints(pv.vtk_points(points_, deep=False))

    if closed:
        spline_function.ClosedOn()
    else:
        spline_function.ClosedOff()

    if parametrize_by == 'length':
        spline_function.ParameterizeByLengthOn()
    else:
        spline_function.ParameterizeByLengthOff()

    _boundary_types_dict = {
        'finite_difference': 0,
        'clamped': 1,
        'second': 2,
        'scaled_second': 3,
    }
    for left_right, constraint, value in zip(
        ('Left', 'Right'), constraints_pair, values_pair, strict=True
    ):
        method = f'Set{left_right}Constraint'
        getattr(spline_function, method)(_boundary_types_dict[constraint])

        if value is not None:
            if constraint == 'finite_difference':  # pragma: no cover
                msg = f'finite difference not compatible with boundary value {value} (should be None)'
                raise ValueError(msg)
            method = f'Set{left_right}Value'
            getattr(spline_function, method)(value)
```

</details>

I tested this locally and `mypy` seems happy with it, and the following code snippet works now. Note the single constraint, and no need to explicitly set the value to `(None, None)`

``` py
import pyvista as pv
pv.Spline(points=[[0,0,0],[1,1,1]],boundary_constraints='finite_difference')
```

user27182: I took the liberty of making this change directly in https://github.com/pyvista/pyvista/pull/8072/commits/10e1791408ba27bf89246072d360fc5bc3368580, @tletut0ur please review, not sure if any of the examples can be updated to use a single constraint/value input instead of tuple inputs

tletut0ur:
@user27182 thank you for the great help !

pyvista/core/utilities/parametric_objects.py
tletut0ur: @user27182 sorry for the delayed reply I did not see that last message because the convo was marked as outdated and resolved. The code looks good to me and I like that we can use single arguments. Realistically it might be what most users are interested in. I see you've updated the examples as well. Maybe one of the examples (I'm thinking the boundary constraints section) should have a tuple just to show how it could be done. Will do it soon.

akaszynski (approved):
LGTM. Thanks all!

user27182:
Should we make a patch-release of `pytest-pyvista` and bump the version in `pyvista` before merging this? We may be able to avoid the unnecessary churn on the vtksz image file changes. I'm not sure what the patch release process is for `pytest-pyvista` though...

akaszynski:
> Should we make a patch-release of `pytest-pyvista` and bump the version in `pyvista` before merging this? We may be able to avoid the unnecessary churn on the vtksz image file changes. I'm not sure what the patch release process is for `pytest-pyvista` though...

Yes, doing now. See https://github.com/pyvista/pytest-pyvista/pull/256

akaszynski:
@user27182, `pytest-pyvista==0.3.2` has been released.

user27182:
> @user27182, `pytest-pyvista==0.3.2` has been released.

Thanks! Working through #8124 to fixup all the images

user27182 (approved):

PR #8074, opened 2025-11-08
Dynamically generate readers table

user27182:
### Overview

Generate the table in the [get_reader](https://dev.pyvista.org/api/readers/_autosummary/pyvista.get_reader.html) docs using `make_tables.py`

The extension->reader mapping is also swapped to show a reader->extension mapping instead.


akaszynski:
Looks good in the branch docs. I also prefer the new ordering of the file reader class and extension columns.

sc000

PR #8076, opened 2025-11-08
Standardize `import pyvista as pv` and `pl = pv.Plotter()`

### LINKED ISSUE #2909: Converge on usage conventions
Throughout the documentation, we have examples that differ between `import pyvista` and `import pyvista as pv`

I am _very_ biased towards the latter and think that all examples and documentation should conform to `import pyvista as pv` much like how `import numpy as np` has become ubiquitous.

My reasoning is that often, I and likely others, will copy/paste snippets from throughout the documentation and then parts of the code expect the PyVista namespace to be `pv` and other parts `pyvista`. We should have a consistent pattern for usage in our documentation to prevent this mismatch and to demonstrate the thought, engineering, and design that has been put into the rest of the API.

For example: 
- `import pyvista`: https://docs.pyvista.org/examples/02-plot/silhouette.html and https://docs.pyvista.org/examples/99-advanced/openfoam-example.html
- `import pyvista as pv`: most other examples

-----

There are other usage patterns that are nit-picky and less important, but one other I think is still worth discussing:

Should we have a consistent pattern for naming `Plotter` objects? I always use `p` (not very PEP8 🤷🏻 ) and others use `plotter` while others use `pl`. I think the docs should converge on a pattern and then add a numerical suffix if working with multiple plotters.

-----

Thoughts? Opinions? Strong feelings on these?

user27182:
### Overview

Close #2909


akaszynski:
This is probably a canidate for `.git-blame-ignore-revs`. LGTM.

akaszynski:
Also, it only took us 3 years to resolve the issue. Thanks @user27182!

PR #8079, opened 2025-11-10
Redirect redundant anchors generated by Sphinx

user27182:
### Overview

Redundant links are generated by autodoc for each class/method/attribute page.
There is a link for the page itself, e.g.:
https://dev.pyvista.org/api/plotting/_autosummary/pyvista.prop3d
And also a link for the class/method/attribute, e.g.
https://dev.pyvista.org/api/plotting/_autosummary/pyvista.prop3d#pyvista.Prop3D

This makes it so that the search produces redundant results, i.e. basically the same result showing up twice.
https://dev.pyvista.org/search.html?q=Prop3D
search

This PR fixes this by ~removing~ redirecting any links where the anchor is the same as the last component of the base URL.


user27182:
This fix is working. Now, when you click on any class or member link, the link still points to pyvista.org/.../pyvista.name#pyvista.Name, but is redirected to pyvista.org/.../pyvista.name instead. 

This way, navigating to the object page is effectively the same as navigating to the object itself.

The issue of duplicate results in the search still exists though. But now they both link to the same place.

akaszynski:
> The issue of duplicate results in the search still exists though. But now they both link to the same place.

Would be lovely to get rid of that. BTW, There's a way to get immediate interactive search (though I think this requires a service) that I've seen in https://mapdl.docs.pyansys.com/. It's not perfect, but I've been a bit spoiled by immediate search feedback and switching over to an improved search service, especially one that doesn't require a backend, would be a great improvement over the current search experience.

Nonetheless, thanks for the PR. Progress is progress.

akaszynski (approved):

PR #8080, opened 2025-11-10
Standardize `Import pyvista as pv` and `pl = pv.Plotter()`

user27182:
### Overview

Part 2 of #8076 

I included a test this time to ensure there are no regular `import pyvista` statements, and no `p = pv.Plotter` or `plot = pv.Plotter`


doc/source/api/utilities/utilities.rst
```diff
@@ -148,7 +148,7 @@ The PyVista library provides a way of getting the version installed in your
 environment.
 
 >>> # Output the version of PyVista.
->>> import pv
+>>> import pyvista as pv
```
akaszynski, line 151: Good catch!

PR #8081, opened 2025-11-10
Fix return type in documentation

edabor:
### Overview

When returning multiple types, documentation is collapsed.
https://docs.pyvista.org/api/utilities/_autosummary/pyvista.get_array.html#pyvista.get_array
image

This PR is an attempt to fix that by explicitly specifying the return name, as done in `numpy` documentation.
One downside is that `RT02` has to be ignored from `numpydoc` rules.

<details><summary>Regex used to search for occurrences</summary>

* `Returns\n {8}-------\n {8}(.*), (\w*)`
* `Returns\n {8}-------\n {8}(.*) or (.*)`
* `Returns\n {8}-------\n {8}(.*) \| (.*)`
* `Returns\n {4}-------\n {4}(.*), (.*)`
* `Returns\n {4}-------\n {4}(.*) or (.*)`
* `Returns\n {4}-------\n {4}(.*) \| (.*)`
</details>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,6 +68,17 @@ repos:
     hooks:
       - id: check-github-workflows
 
+  - repo: local
+    hooks:
+      - id: warn_external
+        name: Convert warnings to warn_external
+        language: python
+        entry: python -m libcst.tool codemod -x hooks.warnings.ConvertWarningsToExternal --no-format
+        additional_dependencies: [libcst]
+        files: ^pyvista/
+        types: [file, python]
+        exclude: pyvista/_warn_external.py
+
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.14.3
     hooks:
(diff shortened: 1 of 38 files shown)


akaszynski:
@user27182, Mac CI is down for the moment. I'm investigating.

user27182:
> RT02 isn't vital, but it's a bit annoying that it won't render correctly without naming the return. 

Can we ignore it on a per docstring basis instead of globally? The multiple return case seems rare enough. IMO the fix to use `output :` and ignore RT02 is highly non obvious, should this be documented in CONTRIBUTING?

akaszynski:
> Can we ignore it on a per docstring basis instead of globally? The multiple return case seems rare enough. IMO the fix to use `output :` and ignore RT02 is highly non obvious, should this be documented in CONTRIBUTING?

That sounds like a compromise. How about we ignore on a per method/function basis and enforce RT02 for the rest? This should be in the CONTRIBUTING since I also agree it's not obvious.

user27182:
I guess maybe it's more common than expected? E.g. the entire downloads module has this issue, with `mesh | str` as the return type

https://dev.pyvista.org/api/examples/_autosummary/pyvista.examples.downloads.download_meshio_xdmf

edabor:
Opened an issue upstream in `numpydoc` to see if there is a workaround https://github.com/numpy/numpydoc/issues/659

akaszynski:
LGTM, thanks for your work here @edabor.

PR #8083, opened 2025-11-10
Invert dev banner text color

user27182:
### Overview

The text on the dev banner has somewhat low contrast, since we have white font on light color (dark theme), or black font on dark color (light theme). This PR inverts the font color so that we have better contrast.

This should now match the font color of the "Hide Search Matches" button.

From main:
dev banner dark
dev banner light

This branch:
image
image


akaszynski:
Contrast looks better. Consider a PR for numpydoc.

PR #8084, opened 2025-11-11
Add png, jpg, tiff, nifti, bmp, pnm image writers

user27182:
### Overview

Add new image data writers to complement the readers we already have for these file formats.

This PR also fixes a bug with the orientation of TIFF images with the TIFFReader.

### DIFF
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -74,6 +74,8 @@ from vtkmodules.vtkCommonComputationalGeometry import (
 from vtkmodules.vtkCommonComputationalGeometry import vtkParametricTorus as vtkParametricTorus
 from vtkmodules.vtkCommonCore import VTK_ARIAL as VTK_ARIAL
 from vtkmodules.vtkCommonCore import VTK_COURIER as VTK_COURIER
+from vtkmodules.vtkCommonCore import VTK_DOUBLE_MAX as VTK_DOUBLE_MAX
+from vtkmodules.vtkCommonCore import VTK_DOUBLE_MIN as VTK_DOUBLE_MIN
 from vtkmodules.vtkCommonCore import VTK_FONT_FILE as VTK_FONT_FILE
 from vtkmodules.vtkCommonCore import VTK_TIMES as VTK_TIMES
 from vtkmodules.vtkCommonCore import VTK_UNSIGNED_CHAR as VTK_UNSIGNED_CHAR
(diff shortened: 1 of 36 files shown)


pyvista/core/dataobject.py
```diff
@@ -255,6 +255,15 @@ def _write_vtk(mesh_: DataObject) -> None:
                 # enable alpha channel if applicable
                 if mesh_[array_name].shape[-1] == 4:
                     writer.SetEnableAlpha(True)
+            elif isinstance(writer, _vtk.vtkTIFFWriter):
+                # Need to flip y axis tp ensure read/save round trip works
+                flip = _vtk.vtkImageFlip()
+                flip.SetInputData(mesh_)
+                flip.SetFilteredAxis(1)
+                flip.Update()
+                mesh_ = flip.GetOutput()
```
user27182: Seems like an issue with the reader https://gitlab.kitware.com/vtk/vtk/-/issues/19587

So we should not modify orientation for the writer here. Should probably only add the flip for the test

pyvista/core/utilities/reader.py
```diff
@@ -2476,6 +2476,12 @@ class TIFFReader(BaseReader):
     _vtk_module_name = 'vtkIOImage'
     _vtk_class_name = 'vtkTIFFReader'
 
+    def _set_defaults_post(self) -> None:
+        """Set defaults on reader post setting file, if needed."""
+        # Fix bug with default TIFF orientation https://gitlab.kitware.com/vtk/vtk/-/issues/19587
+        ORIENTATION_BOTLEFT = 4
+        self.reader.SetOrientationType(ORIENTATION_BOTLEFT)
```
user27182, line 2483: Can confirm that this is a bug with PyVista, e.g. see this image from docs, the text is clearly upside down:

https://docs.pyvista.org/api/examples/_autosummary/pyvista.examples.downloads.download_crater_imagery.html#pyvista.examples.downloads.download_crater_imagery


Image

tests/core/test_reader.py
```diff
@@ -940,6 +968,20 @@ def test_tiff_reader():
     mesh = reader.read()
     assert all([mesh.n_points, mesh.n_cells])
 
+    new_filename = tmp_path / 'new.tif'
+    mesh.save(new_filename)
+    old = pv.read(filename)
+    new = pv.read(new_filename)
+    assert np.allclose(old.active_scalars, new.active_scalars)
+    assert np.allclose(old.index_to_physical_matrix, new.index_to_physical_matrix)
```
user27182, line 978:
```suggestion
    # We should be able to do `assert new == old` but the equality check is too strict since
    # there is floating point error associated with the spacing, so use `np.allclose` instead
    assert np.allclose(old.active_scalars, new.active_scalars)
    assert np.allclose(old.index_to_physical_matrix, new.index_to_physical_matrix)
```

pyvista/core/utilities/reader.py
akaszynski: This has been an issue for quite some time and there's likely an issue open for this.

akaszynski:
Good catch with:
https://github.com/pyvista/pyvista/pull/8084/files#diff-3160d105e3be60a33d8feaf13f3288e83dbada162449a0cf7a506052b38928bf

Terrain is indeed flipped in the implementation on `main`.

user27182:
If you prefer we can make the bug fix for upside-down tiffs as a separate PR

akaszynski:
> If you prefer we can make the bug fix for upside-down tiffs as a separate PR

Nope, it fits in well enough here.

PR #8088, opened 2025-11-11
Do not allow importing vtk directly in tests

### LINKED ISSUE #8087: Do not use `import vtk` in tests

VTK has magic imports which are triggered by using `import vtk`. We should avoid this, and always import directly from modules in tests to ensure that magic imports are only triggered when importing PyVista.

This should help our tests reflect actual usage of PyVista where `import vtk` is _not_ used and help ensure proper test coverage for things like LaTeX rendering, e.g. https://github.com/pyvista/pyvista/pull/8086

### Links to source code.

_No response_

### Pseudocode or Screenshots

_No response_

user27182:
### Overview

Close #8087 

EDIT: After removing all `import vtk` statements from tests, I tested the fix from https://github.com/pyvista/pyvista/pull/8086, and confirmed that the `test_add_text_latex` now fails (as expected) _without_ the magic imports from #8086, but the test succeeds _with_ the magic imports.

### DIFF
--- a/tests/plotting/test_plotting.py
+++ b/tests/plotting/test_plotting.py
@@ -4210,7 +4210,7 @@ def test_plot_cubemap_alone(cubemap, verify_image_cache):
     cubemap.plot()
 
 
-@pytest.mark.skip_egl(reason='Render window will be current with offscreen builds of _vtk.')
+@pytest.mark.skip_egl(reason='Render window will be current with offscreen builds of VTK.')
 def test_not_current(verify_image_cache):
     verify_image_cache.skip = True
 


tests/plotting/test_plotting.py
```diff
@@ -3147,17 +3146,9 @@ def test_add_text():
     pl.show()
 
 
-@pytest.mark.skipif(
-    not check_math_text_support(),
-    reason='VTK and Matplotlib version incompatibility. For VTK<=9.2.2, '
-    'MathText requires matplotlib<3.6',
-)
-@pytest.mark.usefixtures('recwarn')
+@pytest.mark.needs_vtk_version(9, 4, 0)
```
user27182, line 3149: This is the only change not directly related to imports, it's more related to #8086. Doesn't seem to be working for older vtk, and it may not have previously worked either since the test was otherwise always skipped if no math text support was detected.

akaszynski (dismissed):
Fully agree with this change. It's better that we're dogfooding our `_vtk` module. On the plus side, test will collect just slightly faster (assuming no other library does a full import of `vtk`).

akaszynski:
Prefer merging this first before others to avoid rewriting this PR.

tests/plotting/test_plotting.py
```diff
@@ -4219,7 +4210,7 @@ def test_plot_cubemap_alone(cubemap, verify_image_cache):
     cubemap.plot()
 
 
-@pytest.mark.skip_egl(reason='Render window will be current with offscreen builds of VTK.')
+@pytest.mark.skip_egl(reason='Render window will be current with offscreen builds of _vtk.')
```
akaszynski:
```suggestion
@pytest.mark.skip_egl(reason='Render window will be current with offscreen builds of VTK.')
```

akaszynski (approved):

PR #8091, opened 2025-11-13
Improve documentation for supported file formats

user27182:
### Overview
The mesh IO formats are not currently well documented. We support a bunch of VTK readers, some non-VTK readers, and a number of VTK writers. The readers currently do not document the return type(s), and the writers are completely undocumented. This makes it impossible to easily answer the question "which format can I use to read _and_ write `ImageData` (or any other mesh type)?"

This PR addresses this by:

1. Adding a `_CLASS_READER_RETURN_TYPE` mapping to `reader.py`. This allows for the reader return type to be tested and documented. A return type must be defined for every reader.
2. Documenting the return type in the table shown for `get_reader` (new third column): 
image

3. Documenting mesh-specific formats supported by `pyvista.read` and `DataObject.save`:
image
 

### Limitations

- Only formats supported by `reader.py` or `DataObject._WRITERS` are currently defined in the tables. A follow-up PR should refactor code for other supported formats that are supported by read/save (e.g. `.grdecl`, `.pkl`), but for which there is no VTK reader/writer, so they they too can be included in the tables.
- The "File Format" column of the read/save tables is currently auto-generated by parsing the reader/writer name. It would probably be better to add a new reader/writer property to better control the name, and provide a link to the technical specification on the format

### DIFF
--- a/doc/source/_static/redirect_fragments.js
+++ b/doc/source/_static/redirect_fragments.js
@@ -11,10 +11,12 @@
     return s.toLowerCase().replace(/[._-]/g, "");
   }
 
-  const tail = path.substring(path.lastIndexOf("/") + 1);
+  // Strip trailing .html for comparison and redirect target
+  const cleanPath = path.replace(/\.html$/, "");
+  const tail = cleanPath.substring(cleanPath.lastIndexOf("/") + 1);
 
   if (norm(tail) === norm(frag)) {
-    // Replace URL without reloading the full page twice
-    history.replaceState({}, "", path);
+    // Replace URL without reloading the page
+    history.replaceState({}, "", cleanPath);
   }
 })();


user27182:
Links to updated docs:
https://69160cbf8bbdb7189223baac--pyvista-dev.netlify.app/api/readers/_autosummary/pyvista.get_reader.html
https://69160cbf8bbdb7189223baac--pyvista-dev.netlify.app/api/utilities/_autosummary/pyvista.read.html
https://69160cbf8bbdb7189223baac--pyvista-dev.netlify.app/api/core/_autosummary/pyvista.dataobject.save

pyvista/core/utilities/reader.py
```diff
@@ -3749,3 +3751,80 @@ def array_status(self, name):
...
+    XMLMultiBlockDataReader: 'MultiBlock',
+    XMLPartitionedDataSetReader: 'PartitionedDataSet',
+    XMLPolyDataReader: 'PolyData',
+    XMLPRectilinearGridReader: 'RectilinearGrid',
+    XMLPUnstructuredGridReader: 'UnstructuredGrid',
+    XMLRectilinearGridReader: 'RectilinearGrid',
+    XMLStructuredGridReader: 'StructuredGrid',
+    XMLUnstructuredGridReader: 'UnstructuredGrid',
+    XMLPImageDataReader: 'ImageData',
+}
```
akaszynski, line 3830: Ideally this should be dynamically generated from the return types of each reader. Is that possible?

user27182: Agreed, this would be better. It is possible to analyze the annotations at runtime and get the return type. But then we need something to analyze, which means adding a new `read` method to each reader. I think they would all need to have concrete implementations, and would all basically just call `super().read()`. Seems like a lot of boilerplate (40 ish readers, 4 LOC each including whitespace...) not sure if that solution is more attractive.

user27182:
> Other than one comment, this is good to go. Might need a follow-up for maintainability, but as-is it's much more informative than our existing reader documentation (and way more helpful for new users).

For maintainability I think a key part is to ensure we define the same extensions for reading as we do for writing.

I am indirectly working on this with #8094. Once we have proper Reader _and_ Writer classes, I think it will be easier to manage how file extensions are defined and referenced by both the reader and writer.

pyvista/core/utilities/reader.py
user27182: > Agreed, this would be better. It is possible to analyze the annotations at runtime and get the return type. But then we need something to analyze, which means adding a new `read` method to each reader. I think they would all need to have concrete implementations, and would all basically just call `super().read()`. Seems like a lot of boilerplate (40 ish readers, 4 LOC each including whitespace...) not sure if that solution is more attractive.

We _could_ define new classes instead, one for each return type? And then inherit accordingly depending on what's returned. E.g. `_ReturnsPolyData`, `_ReturnsImageData`,  etc. A few would need special casing though since some can return more than one mesh type

akaszynski: > Seems like a lot of boilerplate (40 ish readers, 4 LOC each including whitespace...) not sure if that solution is more attractive.

As boilerplatey as it seems, the lack of typing for existing methods isn't great. This issue forces us to add better typing, something we really should add in the near future.

akaszynski: How about we merge this PR as-is and then as a follow up we can add the return types? It's something that I can do.

akaszynski:
> For maintainability I think a key part is to ensure we define the same extensions for reading as we do for writing.

In that case, let's merge this PR as-is, then once #8094 is merged, we can add in the necessary typing.

user27182:
> > For maintainability I think a key part is to ensure we define the same extensions for reading as we do for writing.
> 
> In that case, let's merge this PR as-is, then once #8094 is merged, we can add in the necessary typing.

Sounds good. #8094 is ready for review.

PR #8093, opened 2025-11-13
Dynamic stacklevel for warnings

edabor:
### Overview

This PR dynamically specifies the stacklevel of warnings to be _outside of pyvista_.
This will help users correctly identify where the warning was emitted in their code.

```python
import pyvista as pv
from pyvista.examples import download_foot_bones

download_foot_bones().connectivity(largest=True)
```
Before this PR:
```shell
C:\Users\User\local\pyvista\pyvista\_deprecate_positional_args.py:248: PyVistaDeprecationWarning: Use of `largest=True` is deprecated. Use 'largest' or `extraction_mode='largest'` instead.
  return f(*args, **kwargs)
```
After this PR:
```shell
C:\Users\User\test.py:6: PyVistaDeprecationWarning: Use of `largest=True` is deprecated. Use 'largest' or `extraction_mode='largest'` instead.
  download_foot_bones().connectivity(largest=True)
```

This convention is enforced by a `pre-commit` hook located in the `hooks` folder.


edabor:
> At a quick glance, this looks like a good candidate to include tests in `tests/test_naming_conventions.py`, which includes a bunch of regex searches to exclude from the project. Could add `warnings.warn(` as something to disallow there.

This is enforced via a local pre-commit in 4098c2de61c7675652dff964f632143d4902feb7 that uses `libcst` to parse/modify the syntax tree.

I think we could do the same for other conventions such as `import pyvista as pv` or `pl = pv.Plotter()`. I am currently exploring this strategy.

user27182:
> > At a quick glance, this looks like a good candidate to include tests in `tests/test_naming_conventions.py`, which includes a bunch of regex searches to exclude from the project. Could add `warnings.warn(` as something to disallow there.
> 
> 
> 
> This is enforced via a local pre-commit in 4098c2de61c7675652dff964f632143d4902feb7 that uses `libcst` to parse/modify the syntax tree.
> 
> 
> 
> I think we could do the same for other conventions such as `import pyvista as pv` or `pl = pv.Plotter()`. I am currently exploring this strategy.

This seems like a better place for these checks. Though some of the other tests were also useful for checking python code in RST files.

edabor:
> This seems like a better place for these checks. Though some of the other tests were also useful for checking python code in RST files.

Indeed, `.rst` files are not parsed by libcst. I'll think about it when trying :)

akaszynski:
I use `pyvista` for a closed source project and tracking down what's emitting the warning usually involves some detective work. I'll certainly use this and I think others will too. Thanks for adding this.

akaszynski (approved):

PR #8094, opened 2025-11-14
Add writer classes

user27182:
### Overview

New `writer.py` module with new API wrappers for all the vtk writers use by PyVista.

This was loosely modelled off the existing `reader.py` module. Most classes do nothing special, except for the refactor of setting `compression` and `binary`, and `texture` keywords. The actual writer objects now handle this, rather than having writer-specific logic inside of `DataObject.save`.

- `texture` keyword code is moved to `PLYWriter.texture`
- `compression` keyword code is moved to `_XMLWriter.compression`
- `binary` keyword code is moved to `_DataModeMixin.data_format`, and inherited by the relevant classes
- writer-specific warnings are moved out of `save`

### DIFF
--- a/tests/namespace/namespace-top.txt
+++ b/tests/namespace/namespace-top.txt
@@ -317,7 +317,6 @@ set_error_output_file, <function set_error_output_file at 0x163f438b0>
 set_jupyter_backend, <function set_jupyter_backend at 0x1645b2f70>
 set_pickle_format, <function set_pickle_format at 0x163f3e3a0>
 set_plot_theme, <function set_plot_theme at 0x1647a4670>
-set_vtkwriter_mode, <function set_vtkwriter_mode at 0x163f3e5e0>
 # sphinx_gallery, <module 'pyvista.utilities.sphinx_gallery' from 'pyvista/pyvista/utilities/sphinx_gallery.py'>
 start_xvfb, <function start_xvfb at 0x1641c8ca0>
 surface_from_para, <function surface_from_para at 0x1641c8670>


pyvista/core/dataobject.py
```diff
@@ -239,24 +237,12 @@ def _warn_imagedata_direction_matrix(mesh: pv.ImageData) -> None:
                 )
 
         def _write_vtk(mesh_: DataObject) -> None:
-            writer = mesh_._WRITERS[file_ext]()
-            set_vtkwriter_mode(vtk_writer=writer, use_binary=binary, compression=compression)
```
user27182, line 245: `set_vtkwriter_mode` seems to me like it's more of an internal helper method. Its only use in the code was here, and it's no longer needed. Should it be deprecated and (eventually) removed?

tests/core/test_polydata.py
```diff
@@ -713,9 +713,13 @@ def test_save(sphere, extension, binary, tmpdir):
         if extension == '.vtp':
             with Path(filename).open() as f:
                 assert 'binary' in f.read(1000)
+        elif extension in ('.geo', '.obj', '.iv'):
+            # Binary is not supported
+            assert not is_binary(filename)
```
user27182, line 718: This test highlights an issue with our API. Users can set `binary=True` for formats that are ASCII-only, and a non-binary file is generated without a warning/error.

pyvista/core/utilities/writer.py
```diff
@@ -0,0 +1,589 @@
...
+    def write(self) -> None:
+        """Write data to path."""
+        self._execute_before_write()
+        self.writer.Write()
+
+    def _apply_kwargs_safely(self, **kwargs) -> None:
+        """Try to set property keyword values and ignore attribute errors."""
+        for name, value in kwargs.items():
+            with contextlib.suppress(AttributeError):
+                setattr(self, name, value)
```
user27182, line 153: This is needed to preserve existing behavior where keywords that don't apply to a reader are simply ignored. As a follow-up, I think we should change this, and explicitly warn or raise an error if this option isn't available for a particular writer. We can easily do `hasattr` checks on the new readers to check for this.

We could add a keyword to toggle this, and enable it by default.

This would help address the issue highlighted here https://github.com/pyvista/pyvista/pull/8094#discussion_r2530046982

pyvista/core/dataobject.py
```diff
@@ -881,6 +794,14 @@ def _serialize_pyvista_pickle_format(self: Self) -> dict[str, Any]:
             preferred since it supports more objects (e.g. MultiBlock).
 
         """
+        from vtkmodules.vtkIOLegacy import vtkDataSetWriter
+        from vtkmodules.vtkIOXML import vtkXMLImageDataWriter
+        from vtkmodules.vtkIOXML import vtkXMLPolyDataWriter
+        from vtkmodules.vtkIOXML import vtkXMLRectilinearGridWriter
+        from vtkmodules.vtkIOXML import vtkXMLStructuredGridWriter
+        from vtkmodules.vtkIOXML import vtkXMLTableWriter
+        from vtkmodules.vtkIOXML import vtkXMLUnstructuredGridWriter
```
akaszynski, line 803: I'm not at all opposed to dynamically importing, and in fact I think we should move in the direction of dynamically importing as much as possible to reduce memory overhead and import time (e.g. filters). There are even cases where a project might use `pyinstaller` and VTK libraries often take up a significant portion of the frozen application.

pyvista/core/utilities/writer.py
akaszynski: Fully agree to raise or warn rather than silently ignore. Doesn't even have to be optional.

akaszynski: However, I agree that it should be a follow-up.

tests/core/test_polydata.py
akaszynski: Should be resolved when we check kwargs, right?

akaszynski (approved):
I don't see any issues with the overall implementation, and concrete wrapped classes have been needed for a while.

Is the next step to associate readers with writers and then add typing? There may be cases where we don't have an associated writer (i.e. `meshio`).

pyvista/core/dataobject.py
user27182: Yes, and these specific imports were added because they're only needed for legacy pickling , so aren't used by default and will rarely be used anyway, and would kind of "undo" the effort of all other reader/writer classes which try to import lazily

user27182: > I think we should move in the direction of dynamically importing as much as possible to reduce memory overhead and import time (e.g. filters).

PEP 810 was recently approved to add lazy imports natively to Python 3.15, could also try using the Scientific Python lazy loader until then.

pyvista/core/dataobject.py
user27182: I'm +1 for outright removal. It's in the main PyVista namespace with `pv.set_vtkwriter_mode`, so wasn't sure about external use cases.

tests/core/test_polydata.py
user27182: Yes, can be resolved in a follow up

pyvista/core/dataobject.py
akaszynski: Excited about [PEP 810](https://peps.python.org/pep-0810/), though it's going to be at least a year until we can use it (and years more until it's mainstream). The alternative is to place all imports within methods, which, given our degree of testing, I think is equally good at the moment.

PR #8096, opened 2025-11-15
Test vtk dev with python 3.14

user27182:
### Overview

Dev wheels for python 3.14 are now being uploaded. Let's test with them. https://gitlab.kitware.com/vtk/vtk/-/issues/19845

### DIFF
--- a/tests/plotting/test_picking.py
+++ b/tests/plotting/test_picking.py
@@ -606,7 +606,7 @@ def test_block_picking(multiblock_poly):
 
 
 @pytest.mark.parametrize('mode', ['mesh', 'cell', 'face', 'edge', 'point'])
-@pytest.mark.skip_check_gc
+@pytest.mark.skip_check_gc("vtkDataArray not gc'd with 'point' mode on Python 3.14 vtk dev wheels")
 def test_element_picking(mode):
     class Tracker:
         def __init__(self):
--- a/tests/plotting/test_plotting.py
+++ b/tests/plotting/test_plotting.py
@@ -2037,7 +2037,7 @@ def test_volume_rendering_from_plotter(uniform):
 
 
 @skip_windows_mesa  # due to opacity
-@pytest.mark.skip_check_gc
+@pytest.mark.skip_check_gc("vtkWeakReference not gc'd on Python 3.14 vtk dev wheels")
 def test_volume_rendering_rectilinear(uniform):
     grid = uniform.cast_to_rectilinear_grid()
 
@@ -3697,7 +3697,7 @@ def test_plotter_lookup_table(sphere, verify_image_cache):
 
 
 @skip_windows_mesa  # due to opacity
-@pytest.mark.skip_check_gc
+@pytest.mark.skip_check_gc("vtkTypeUInt8Array not gc'd on Python 3.14 vtk dev wheels")
 def test_plotter_volume_lookup_table(uniform):
     uniform.set_active_scalars('Spatial Point Data')
 


toxfile.py
```diff
@@ -14,6 +14,16 @@
...
+    if 'vtk_dev' in env_conf.env_name:
+        # Remove vtk from deps since it's installed separately
+        deps_obj = env_conf['deps']
+        deps = deps_obj.lines()
+
+        filtered = [d for d in deps if not d.strip().startswith('vtk')]
+
+        # Mutate deps
+        deps_obj.raw = '\n'.join(filtered)
+
```
user27182: We need to remove vtk from the project deps otherwise uv can't resolve them since there's no stable vtk for Python 3.14.

This change to the toxfile config tries to do that by modifying the deps used by tox in-place, but it doesn't work.

We can't use the uv pre-release and extra index flags here either since we need a custom install command to make the dev install work.

@edabor do you know of any tox hacks or tricks to make this work? Or somehow "force" the initial PyVista install to continue without vtk?

edabor: As an interim solution, I tried in https://github.com/pyvista/pyvista/pull/8096/commits/ba4c5da7328594cba5e4d083b0e8055cf8af476d to specify that vtk should be installed for py < 3.14 in the package dependencies. I don't see any other simple solution, but it haven't either investigated further since the above was working.

From my understanding of `tox`, the deps are installed following the order:
- `deps` field in `tox.ini`
- `dependency_group` in `tox.ini`
- package deps found in `pyproject.toml`

The fact that dependencies versions specified in the `deps` of `tox.ini` (ie. `vtk`) are kept when installing package deps is only due to the fact that `uv` does not override an already installed satisfying dependency version.
See https://github.com/pyvista/pyvista/actions/runs/19487423844/job/55772535339#step:7:22

edabor: BTW, should we constrain the supported python version below 3.14 for `pyvista` while waiting for `vtk` to publish their official wheel ? I see that https://github.com/pyvista/pyvista/pull/8102 is also related to `3.14` compatibility.

user27182: Thanks for the fix. Didn't realize you could pin deps to python version like that.

> BTW, should we constrain the supported python version below 3.14 for `pyvista` while waiting for `vtk` to publish their official wheel ? I see that https://github.com/pyvista/pyvista/pull/8102 is also related to `3.14` compatibility.

Indeed PyVista currently fails on import for Python 3.14. But users will only see this if they can even get to the import stage. But since vtk is already pinned, the vtk dep blocks PyVista from being installed on 3.14 at all for current users of v0.46.

So yes #8102 is critical for Python 3.14 support, but I don't think we necessarily need to pin python for the purpose of patching 0.46 to avoid compatibility issues.

tests/plotting/test_picking.py
```diff
@@ -606,6 +606,7 @@ def turn_blue(index, dataset):  # noqa: ARG001
 
 
 @pytest.mark.parametrize('mode', ['mesh', 'cell', 'face', 'edge', 'point'])
+@pytest.mark.skip_check_gc
```
user27182: Seems there are some new memory leaks? I haven't investigated, these are just skipped so CI passes

pyproject.toml
```diff
@@ -24,10 +24,10 @@ dependencies = [
...
   'scooby>=0.5.1',
   'typing-extensions>=4.10',
-  'vtk!=9.4.0',
-  'vtk!=9.4.1',
-  'vtk<9.6.0',
-  'vtk>=9.2.2',
+  'vtk!=9.4.0; python_version < "3.14"',
+  'vtk!=9.4.1; python_version < "3.14"',
+  'vtk<9.6.0; python_version < "3.14"',
+  'vtk>=9.2.2; python_version < "3.14"',
```
edabor, line 30: Is this change going to be in the next release ? If so, that would mean that `pyvista` could be installed on `python 3.14` without `vtk`

user27182: I think the next PyVista release will probably support vtk 9.6 (and Python 3.14), and hence this change should be reverted before the next release, though we should probably track this in a new issue.

Ideally, we should be able to install vtk dev wheels for Python 3.14, or any new and yet-supported Python version,  without needing to make this change (which we then need to remember to revert before release).
Getting the latest vtk to install for dev purposes only was one of the challenges with this PR.

PR #8098, opened 2025-11-17
Refactor common API for readers/writers and include new `extensions` property

user27182:
### Overview

The readers and writers have a few things in common. This PR unifies their API with a new `_FileIOBase` class.

### DIFF
--- a/pyvista/core/utilities/fileio.py
+++ b/pyvista/core/utilities/fileio.py
@@ -59,7 +59,7 @@ class _FileIOBase(ABC, _NoNewAttrMixin):
     _vtk_class_name: str = ''
 
     def __repr__(self) -> str:
-        """Representation of a writer object."""
+        """Representation of a FileIO object."""
         return f'{self.__class__.__name__}({self.path!r})'
 
     @property


tests/core/test_utilities.py
```diff
@@ -2983,9 +2987,20 @@ def test_writer_data_mode_mixin(writer_cls):
     assert obj.data_format == 'binary'
 
 
+@pytest.mark.parametrize('cls', [*READER_CLASSES, *WRITER_CLASSES])
+def test_fileio_extensions(cls):
+    if cls in [pv.OpenFOAMReader, pv.MultiBlockPlot3DReader]:
+        # These classes are not associated with any extensions
+        pytest.xfail()
```
user27182, line 2996: `.foam` is already mapped to `pv.POpenFOAMReader`, so `pv.OpenFOAMReader` is not used. I think this is OK

and `pv.MultiBlockPlot3DReader` should probably be mapped to `.xyz` (that's what Paraview lists as the extension for PLOT3D files), but we have `.xyz` already mapped to  `SimplePointsWriter` for saving `PointSet`, which would create a disconnected between read and write extensions.

akaszynski: I'm not sure why we've mapped `.xyz` to `PointSet` and have always been a bit confused by that. If there's no rational, we should probably deprecate it.

akaszynski (approved):
LGTM with one follow-up.

PR #8102, opened 2025-11-18
Python 3.14: Remove custom docstrings from type aliases

user27182:
Python 3.14 updated `Union` to disallow custom `__doc__` properties, so we can no longer set custom docstrings.

`UP007` was previously ignored as part of #7796 to prevent `Union` aliases from being upgraded to `TypeAlias` aliases, since type aliases also do not allow setting `__doc__`. It seems this workaround is shortlived, however, due to the updates to 3.14.

And so, it seems the Python universe is insisting that we should not be allowed to have custom docstrings for our type aliases. This is a bit of a bummer as it makes our typing docs way less understandable.

EDIT: as a follow-up, we could try to use proper sphinx directives to document these

### DIFF
--- a/pyvista/core/_typing_core/_aliases.py
+++ b/pyvista/core/_typing_core/_aliases.py
@@ -6,6 +6,7 @@ import os
 from typing import TYPE_CHECKING
 from typing import Literal
 from typing import NamedTuple
+from typing import TypeAlias
 
 from pyvista.core import _vtk_core as _vtk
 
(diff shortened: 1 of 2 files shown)


akaszynski:
> This is a bit of a bummer as it makes our typing docs way less understandable.

https://691cf0567c957885977b83ad--pyvista-dev.netlify.app/api/core/typing.html

A bit. As a follow-up we can try to use sphinx directives as you suggested. For now, getting it to work on Python 3.14 seems like a prerequisite to Python 3.14 VTK dev testing.

user27182:
> > This is a bit of a bummer as it makes our typing docs way less understandable.
> 
> https://691cf0567c957885977b83ad--pyvista-dev.netlify.app/api/core/typing.html
> 
> A bit. As a follow-up we can try to use sphinx directives as you suggested. For now, getting it to work on Python 3.14 seems like a prerequisite to Python 3.14 VTK dev testing.

For completeness, here's what this change looks like:
## main branch
before

## this branch
after

akaszynski:
That's quite messy. Let's get it working for the moment and clean it up in another PR.

PR #8109, opened 2025-11-20
Use `nest-asyncio2`

### LINKED ISSUE #7938: nest_asyncio dependency is no longer maintained
The [`nest_asyncio` repo](https://github.com/erdewit/nest_asyncio) has been archived as of March 2024 and the [latest pypi](https://pypi.org/project/nest-asyncio/#history) release was in Jan 2024. It is already behind, as its monkey-patched `asyncio.run` doesn't support the same interface as `asyncio.run` as of python 3.12 (missing the [`loop_factory` argument](https://docs.python.org/3.12/library/asyncio-runner.html#asyncio.run)). The dependency seems like a liability.

user27182:
### Overview

Resolve #7938

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,6 +130,7 @@ docs = [
   'meshio==5.3.5',
   'mypy-extensions==1.1.0',
   'mypy==1.18.2',
+  'nest-asyncio2==1.6.3',
   'numpydoc==1.9.0',
   'pydata-sphinx-theme==0.15.4', # Do not upgrade, see https://github.com/pyvista/pyvista/pull/7095
   'pypandoc==1.15',


larsoner:
FYI this led to a cryptic CI failure in MNE-Python with notebooks:

https://github.com/mne-tools/mne-python/actions/runs/19581891303/job/56081777294?pr=13503

I suspect other people will hit this since almost all end users will have `nest-asyncio` since it's required by `ipykernel`:
```
$ pip show nest-asyncio | grep Required
Required-by: ipykernel, nbclassic
```
I'll work around it by installing `nest-asyncio2`. But maybe in the meantime `pyvista` should fall back to `nest-asyncio` if `nest-asyncio2` isn't installed?

user27182:
> I suspect other people will hit this since almost all end users will have `nest-asyncio` since it's required by `ipykernel`:
> 
> ```
> 
> $ pip show nest-asyncio | grep Required
> 
> Required-by: ipykernel, nbclassic
> 
> ```

Since `nest-asyncio` is archived, those projects are likely also going to want to stop using it soon. There will probably a pain period during this switch.

Maybe @Chaoses-Ib can help these packages update to `nest-asyncio2`?

EDIT: replied-to here https://github.com/pyvista/pyvista/issues/8115#issuecomment-3565353856

PR #8111, opened 2025-11-20
Add warping spheres example and enable plotting `PartitionedDataSet`

user27182:
### Overview

New `PartitionedDataSet` example. Turns out there's no plotting support for this class yet, so minimal plot support is added too. As a follow-up, `PartitionedDataSet` should probably be moved to `composite.py`, and there should be a base class for both `MultiBlock` and `PartitionedDataSet`. Pretty sure  `PartitionedDataSet` can share most if not all of the same methods and filters.


pyvista/examples/downloads.py
```diff
@@ -8701,3 +8701,47 @@ def download_yinyang(*, load=True):
...
+
+    Parameters
+    ----------
+    load : bool, default: True
+        Load the dataset after downloading it when ``True``.  Set this
+        to ``False`` and only the filename will be returned.
+
+    Returns
+    -------
+    PartitionedDataSet | str
```
edabor:
```suggestion
    output : PartitionedDataSet | str
```

user27182:
The documentation tests fail due to https://github.com/pyvista/pytest-pyvista/issues/252 , which should be fixed with a patch release

PR #8114, opened 2025-11-21
Move typing docstrings to `doc` rst files

user27182:
### Overview

Reverts #8102 and uses ~[Sphinx doc comments](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings) `#:` instead of relying on `__doc__`~

The Sphinx doc comments don't work. Instead, the docstrings are moved to the `doc` folder.
I was able to restore all docstrings before #8102 merged, except for `ColorLike`, since that alias has its own page generated by autodoc. Also, using `typing.Union` renders way better than `typing.TypeAlias` from #8102, so let's keep those.


user27182:
Docs are mostly restored to what they were before #8102

image

PR #7996, opened 2025-10-07
Use binary filter dynamically with morphological operators

user27182:
### Overview

Follow-up to #7833:
- this PR uses a `binary` keyword instead of `continuous`, and overloads it so that users can specify the actual binary dilation/erosion values
- this PR dynamically selects the vtk filter: the binary vtk filter is used for performance where possible, and the more general continuous vtk filter is otherwise used when the input isn't actually binary
- this PR includes detailed and relevant examples which clearly illustrates how the filters work and how to use them
- this PR separates the various processing steps into standalone functions (e.g. validate scalars, compute binary values, configure alg) so that setup can be done exactly once _before_ doing any alg updates. For the open/close, filters, this avoids redundant input validation calls (where slow np.unique calls may be made) but is also necessary since we want to dilate/erode with the same values for _both_ calls (one dilate call, one erode call) without re-computing these values between update calls

Tests are also moved from dataset to imagedata filters.

PR #8118, opened 2025-11-23
Fix empty interactive examples in documentation

edabor:
### Overview

Potentially reverts:
https://github.com/pyvista/pyvista/pull/7924
https://github.com/pyvista/pyvista/pull/7979
https://github.com/pyvista/pyvista/pull/8107


examples/01-filter/slice.py
```diff
@@ -28,6 +28,11 @@
 # volume. First, load up the volume and preview it:
 
 mesh = examples.load_channels()
+
+# convert to int32 for interactive documentation
+name = mesh.active_scalars_info.name
+mesh.cell_data[name] = mesh.cell_data[name].astype(np.int32)
```
user27182: Is the need to use `int32` the root of most of these problems? Is this an upstream issue with vtkjs? Other than impacting our docs, it seems this would also impact users that "forget" to cast to `int32` for their own work, no? Is there a way to auto-patch this in `PyVista`, or issue a warning somewhere, so that we don't always need to remember to do this (and avoid ugly "for interactive documentation only" code) ?

edabor: > Is the need to use `int32` the root of most of these problems?

Yes

> Is this an upstream issue with vtkjs? 

I guess so. I plan to open an issue there.

> Other than impacting our docs, it seems this would also impact users that "forget" to cast to `int32` for their own work, no? Is there a way to auto-patch this in `PyVista`, or issue a warning somewhere, so that we don't always need to remember to do this (and avoid ugly "for interactive documentation only" code) ?

I agree this is ugly. However, I guess we won't forget it since for new examples, code review of the doc cache would exhibit empty `vtksz` images.

user27182: > > Is the need to use `int32` the root of most of these problems?
> 
> Yes
> 
> > Is this an upstream issue with vtkjs?
> 
> I guess so. I plan to open an issue there.
> 
> > Other than impacting our docs, it seems this would also impact users that "forget" to cast to `int32` for their own work, no? Is there a way to auto-patch this in `PyVista`, or issue a warning somewhere, so that we don't always need to remember to do this (and avoid ugly "for interactive documentation only" code) ?
> 
> I agree this is ugly. However, I guess we won't forget it since for new examples, code review of the doc cache would exhibit empty `vtksz` images.

More than just ugly, this could also result in blank renderings for users that are unaware of the need to cast to int32. Could we not add a check inside of `export_vtksz`, and auto-cast any `int64` active scalars to `int32` and print a warning? Or forgo the auto-cast, and just print the warning (either way, on our end we can't just ignore the warning so we'd need to explicitly cast beforehand anyway, so the ugly fix in the docs would probably need to stay).

Also, with the status quo, instead of applying the fix and casting to fix the interactive plots, devs may end up just forcing static plots since IMO the casting fix is not obvious, e.g. this happened here https://github.com/pyvista/pyvista/pull/7996#issuecomment-3386114473. Since I now know it's a dtype issue, I might try that instead, but I don't think it's safe to assume that the casting fix will be applied by devs in the future.

edabor: I agree with your points, this is goes beyond code ugliness.
I think we should try to solve the issue upstream in `trame-vtk` rather than trying to accommodate.

From local testing, I've seen that `trame-vtk == 2.10.0` is working fine with `int64` data. Let's try and see how it goes.
If that works, we'll have to revert https://github.com/pyvista/pyvista/pull/7979 and force to static for the subplots with opacity, since https://github.com/Kitware/trame-vtk/issues/94 is not fixed.

edabor:
According to https://github.com/pyvista/pyvista/pull/8124#issuecomment-3582563415 , waiting for `pytest-pyvista` patch release to enable latest `trame-vtk` and fix image sizes mismatches.

PR #8128, opened 2025-11-28
Bump sphinx to <9.1

user27182:
### Overview

Currently in pre-release


user27182:
Looks like other deps have Sphinx pinned < 9

PR #8148, opened 2025-12-02
Use `uv lock` in CI workflow to check for dependency conflicts

### LINKED ISSUE #7728: Inconsistent version of trimesh

user27182:
Resolve #7728 
Dependabot regularly updates dependencies in one group but not for all groups. This leads to unresolvable dependencies if trying to install multiple groups. See https://github.com/pyvista/pyvista/issues/7728#issuecomment-3091947859

EDIT: Reviewers, please add this job to the list of required jobs that must pass before merging

### DIFF
--- a/.github/workflows/docker-package.yml
+++ b/.github/workflows/docker-package.yml
@@ -30,7 +30,7 @@ jobs:
       startsWith(github.event.pull_request.head.ref, 'docker')
       )
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           persist-credentials: false
 
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,7 +26,7 @@ jobs:
       PARALLEL: "-j2" # Public GitHub runners have 2 real cores
       TOX_COLORED: "yes"
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           persist-credentials: false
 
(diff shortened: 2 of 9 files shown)


user27182:
The new job successfully identified an issue with `trimesh`:
https://github.com/pyvista/pyvista/actions/runs/19873302783/job/56954383988?pr=8148

user27182:
Reviewers : please add this job to the list of required jobs that must pass before merging

akaszynski:
> Reviewers, please add this job to the list of required jobs that must pass before merging

sc014

PR #8151, opened 2025-12-04
Allow using `Plane` with plane-based filters

user27182:
### Overview

Replacement for #6576.

Add new `plane` parameter to the following filters:
- `clip`
- `slice`
- `clip_closed_surface`
- `project_points_to_plane`

This allows using an actual plane to implicitly define the `normal` and `origin` parameters instead of passing these explicitly.

The "NORMALS" dict is also removed from the top PyVista namespace and is now private. This is more of an implementation detail, this may technically be a breaking change though I don't think it's likely to be in use.


banesullivan (approved):
Nice change! Big +1 for this

pyvista/core/utilities/helpers.py
```diff
@@ -288,13 +299,51 @@ def generate_plane(normal: VectorLike[float], origin: VectorLike[float]):
...
+
+        if (dimensionality := plane_.dimensionality) != 2:
+            msg = (
+                f'The plane mesh must be planar. Got a non-planar mesh with dimensionality of '
+                f'{dimensionality}.'
+            )
+            raise ValueError(msg)
+        origin = plane_.points.mean(axis=0)
+        normal = plane_.point_normals.mean(axis=0)
+        return origin, normal
```
banesullivan, line 329: This will work for any PolyData surface, right? even a curvilinear surface? Hoenstly I think that's a plus being able to slice along any polydata by the mean normal/origin

user27182: There is an instance check to make sure we have `PolyData`, and a dimensionality check which ensures the input is 2D only (using matrix rank of the points). So on this basis, no, curvilinear should not work (though I don't explicitly test for this).

But, there is an (undocumented) quirk with the dimensionality implementation, where for a structured grid, dimensionality is defined by `dimensions`. So a curvilinear `StructuredGrid` can have dimensionality of 2, but if you convert to `PolyData`, it would have dimensionality of 3.

We can remove the instance check for PolyData and allow `any` dataset with dimensionality of 2, including curvilinear. This would also enable slicing with 2D `ImageData`.

EDIT: I can update this and add examples in a follow-up PR

PR #8155, opened 2025-12-04
Use `pl = pv.Plotter` in docstrings

user27182:
### Overview

Follow-up to #8076 and #8080. The regex is updated to search docstring prompts and I fixed the few cases that were found.


banesullivan (approved):
Thank you!

PR #8160, opened 2025-12-07
Add `UnstructuredGrid.distinct_cell_types`

user27182:


banesullivan:
I wonder if we should add a backup method of `set([pv.CellType(v) for v in np.unique(mesh.celltypes)])` for VTK<9.5 so that we can have other parts of the API use this property?

user27182:
> I wonder if we should add a backup method of `set([pv.CellType(v) for v in np.unique(mesh.celltypes)])` for VTK<9.5 so that we can have other parts of the API use this property?

This PR already supports this for older VTK.

banesullivan:
🤦🏻‍♂️

PR #8162, opened 2025-12-07
Add range value to clip_scalar (isovolume)

### LINKED ISSUE #7256: isovolume clip filter

ParaView has an "isovolume" filter that is essentially two `clip_scalar` fitlers chained together. We should implement this as well!

See https://github.com/pyvista/pyvista/discussions/3754#discussioncomment-12346008

cc @kamakshidasan

Ref source code in ParaView: https://github.com/Kitware/ParaView/blob/584e66a54f7902ded3ce05a39a0abef73673f86b/VTKExtensions/FiltersGeneral/vtkIsoVolume.cxx#L54

### Pseudocode or Screenshots

```python
import pyvista as pv

def isovolume(mesh: pv.DataSet, vmin: float, vmax: float, scalars: str | None = None):
    """Perform an isovolume clip."""
    return mesh.clip_scalar(
        scalars=scalars, value=vmin, invert=False,
    ).clip_scalar(
        scalars=scalars, value=vmax, invert=True,
    )

# Example
mesh = pv.Wavelet()
vol = isovolume(mesh, 125, 200)
vol.plot(clim=mesh.get_data_range("RTData"))
```

Image

tletut0ur:
### Overview

resolves #7256

I am currently checking which of the other arguments of `clip_scalar` enter in conflict with this definition.
I can also draft an example that would compare the filters as suggested by @MatthewFlamm, but let me know if you would rather have this as a separate PR.


MatthewFlamm:
Recently my time to dedicate to working on this project has dwindled, but my two cents are that this PR should focus on one thing.  The example comparing multiple filters is related, but really separate to this PR.  A separate PR for that purpose would be very welcome.

tletut0ur:
I wanted to implement the inplace argument. Interestingly, when I run the snippet below on the code that is on main branch (meaning no isovolume type stuff)

```python
import pyvista as pv

mesh = pv.examples.download_whole_body_ct_male()['ct']
mesh.clip_scalar(scalars = "NIFTI", 
                 value=150.0, 
                 inplace=True)
mesh.plot()
```


I get 
`TypeError: The Input DataSet type <class 'pyvista.core.pointset.UnstructuredGrid'> must be compatible with the one being overwritten <class 'pyvista.core.grid.ImageData'>`
Totally fair, but it makes me question whether the `inplace` argument should be kept in its current state. I'll add some check before that says inplace has to be False if the input is ImageData.

pyvista/core/filters/data_set.py
```diff
@@ -571,9 +572,19 @@ def clip_scalar(  # type: ignore[misc]  # noqa: PLR0917
             alg: _vtk.vtkClipPolyData | _vtk.vtkTableBasedClipDataSet = _vtk.vtkClipPolyData()  # type: ignore[unreachable]
         else:
             alg = _vtk.vtkTableBasedClipDataSet()
-
+        is_single_value = isinstance(value, (float, int))
+        if is_single_value:
```
user27182: This _might_ fix the mypy error:
```suggestion
        if (is_single_value := isinstance(value, (float, int))):
```

PR #8164, opened 2025-12-08
Add mesh validation methods

### LINKED ISSUE #8132: Add mesh validation methods

We should add methods to validate mesh arrays, mesh connectivity, and/or  mesh cells.
We should then add a new keyword to `read`, `save`, and `PolyData`/`UnstructuredGrid` init methods to auto-call these methods by default and emit warnings when invalid arrays/connectivity/cells are detected. This can help eliminate or prevent a large number of bugs.

- Validating mesh arrays should check all point and cell array lengths and ensure they match number of points and cells, respectively. This should help prevent issues like [#6222](https://github.com/pyvista/pyvista/issues/6222) and [#8101](https://github.com/pyvista/pyvista/issues/8101). This should be pretty straightforward to implement (just compare lengths against n_points, n_cells)

- Validating connectivity should ensure that cells do not have point ids that exceed the actual number of points, and would help prevent issues like [#1097](https://github.com/pyvista/pyvista/issues/1097), [#2124](https://github.com/pyvista/pyvista/issues/2124), [#7987](https://github.com/pyvista/pyvista/issues/7987), [#4735](https://github.com/pyvista/pyvista/issues/4735). We could also warn if there are unused points (not referenced by any cells?). This should be somewhat straightfoward to implement, e.g.  do:

``` py
def warn_connectivity(poly):
    if np.max(poly._connectivity_array) >= poly.n_points:
        warnings.warn(f'\nThe faces/polygons reference point ids that do not exist. The connectivity\n'
                      f'array should not have values that exceed the number of points ({poly.n_points}).\n'
                      f'This can cause Python to crash with some filters.')
```
as suggested here https://github.com/pyvista/pyvista/issues/6115#issuecomment-2130232536

- Validating cells should help prevent issues like [#8129](https://github.com/pyvista/pyvista/issues/8129), and could be implemented using `vtkCellValidator`. We can possibly re-use the implementation from [ansys/pydpf-core/#1836](https://github.com/ansys/pydpf-core/pull/1836)

_No response_

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

user27182:
Resolve #8132 

- Add new filter `DataSetFilters.cell_validator` which uses `vtkCellValidator`
- Add a new method `DataSet.validate_mesh` which builds upon this and adds additional validation for data arrays, invalid cell references, and invalid points
- `validate_mesh` can be used to emit warnings, raise errors, or for users to analyze the results by inspecting the returned `ValidationReport` dataclass
- Add new "advanced" sphinx gallery example demonstrating some of the issues which are validated by `validate_mesh`
- Tests are also added to ensure all cell examples are valid cells. The `Polygon` cell was found to be concave and is reported as invalid, so it's been changed to be convex.

### DIFF
--- a/tests/examples/test_cell_examples.py
+++ b/tests/examples/test_cell_examples.py
@@ -50,8 +50,8 @@ def test_area_and_volume(cell_example):
 @parametrize('cell_example', cell_example_functions)
 def test_cell_is_valid(cell_example):
     mesh = cell_example()
-    issues = mesh.validate_mesh().invalid_fields
-    assert not issues
+    invalid_fields = mesh.validate_mesh().invalid_fields
+    assert not invalid_fields
 
 
 def test_empty():


pyvista/core/dataset.py
```diff
@@ -62,6 +68,392 @@
...
+
+    @property
+    def is_valid(self) -> bool:  # numpydoc ignore=RT01
+        """Return ``True`` if the mesh is valid."""
+        return not self.issues
+
+    @property
+    def issues(self) -> tuple[str, ...]:  # numpydoc ignore=RT01
+        """Return any field names which have values."""
+        return tuple(f.name for f in fields(self) if getattr(self, f.name))
```
user27182, line 426: Maybe we should call this `invalid_fields` instead of `issues`?

akaszynski: Yes, `invalid_fields` is more descriptive and `issues` could mean anything.

pyvista/core/dataset.py
```diff
@@ -3031,3 +3452,258 @@ def dimensionality(self: Self) -> Literal[0, 1, 2, 3]:
...
+            fields are included by default. Specify individual fields by name, or use group name(s)
+            to include multiple related validation fields:
+
+            - ``'data'`` to include all data fields
+            - ``'cells'`` to include all cell fields
+            - ``'points'`` to include all point fields
+            - ``'unsafe'`` to include all fields that, if invalid, may cause a
+              segmentation fault and crash Python. This option includes
+              ``point_data_wrong_length``, ``cell_data_wrong_length``, and
+              ``invalid_point_references``.
```
user27182: Might be better to use a different term other than `'unsafe'`?

E.g. `validate_mesh('unsafe')` doesn't seem very obvious... maybe `'memory_safe'` ?

akaszynski: Yes, or `"segfault"`, but `"memory_safe"` is probably better.

pyvista/core/dataset.py
user27182: This is done in [ed3c433](https://github.com/pyvista/pyvista/pull/8164/commits/ed3c43385423115eff53fb79973c8951d3b52322)

pyvista/core/dataset.py
user27182: Yea I think `validate_mesh("memory_safe")` probably reads better? "Validate the mesh so it has safe memory access". I think `"segfault"` is too much like `"unsafe"` in that the word has negative implications, whereas `"memory_safe"` conveys the opposite.

akaszynski (approved):
Reviewed PR and generated docs at https://69459f4ade88d26299a65d71--pyvista-dev.netlify.app/api/core/_autosummary/pyvista.dataset.validate_mesh and found no issues. Docs are great for this PR.

PR #8169, opened 2025-12-11
Fix face orientations of cell examples

user27182:
### Overview

Faces should be defined such that point traversal is in the counter-clockwise direction. Otherwise, the cell is not valid as determined by `vtkCellValidator`.

This PR:
- adds a new `show_normals` option to `plot_cell` so that we can visually see the face orientations. There is special handling for Polyhedron cells to ensure the normals that are shown reflect the actual face orientation of the input
- fixes some 3D cells, e.g. `Polyhedron` and `QuadraticWedge` which had incorrect orientations. 
- fixes some 2D cells (e.g. `Triangle`, `TriangleStrip`) which have the correct face orientation, but the normal direction was `-z`. Most cells have a `+z` direction, so these cells were modified such that all 2D cell examples have a `+z` direction.
 - adds better support for plotting grids with multiple cells

### DIFF
--- a/pyvista/core/celltype.py
+++ b/pyvista/core/celltype.py
@@ -15,20 +15,11 @@ _CELL_TYPE_TO_NUM_POINTS: dict[np.uint8, int] = {}
 
 PLACEHOLDER = 'IMAGE-HASH-PLACEHOLDER'
 
-_DROPDOWN_TEMPLATE = """
-.. dropdown:: More info
-    :icon: info
-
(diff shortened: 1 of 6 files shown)


user27182:
The figures from the tests are bit distorted. Here's how the normals look when plotted normally:

``` py
from pyvista import examples
examples.plot_cell(examples.cells.Wedge(), show_normals=True)
```
image

PR #8170, opened 2025-12-11
Only download minimum geovista assets during integration test

bjlittle:
Rather than downloading **all** `geovista` assets (which will only grow over time in number and footprint) prior to its integration test, leave `geovista` to download and unpack only the assets that it needs as it needs them (via `pooch`) during test execution.

Maintenance-wise this seems like a more scalable and efficient approach moving forwards.


bjlittle:
Ohhhh, interesting ... are these documentation CI failures a symptom of this change? 

Or something separate?

user27182:
> Ohhhh, interesting ... are these documentation CI failures a symptom of this change?
> 
> Or something separate?

Probably a github thing. I am seeing similar 503 HTTP errors in other PRs.

akaszynski (approved):

PR #8174, opened 2025-12-14
Enable setting dataset `center`

user27182:
### Overview

Use `translate` to set the center. Using a transformation like this is safe because it doesn't affect any vectors (unlike using `scale`, which may or may not need to use `transform_all_input_vectors` depending on use case).

### DIFF
--- a/pyvista/core/composite.py
+++ b/pyvista/core/composite.py
@@ -1300,7 +1300,7 @@ class MultiBlock(
 
     @property
     def center(self: MultiBlock) -> tuple[float, float, float]:
-        """Set or eturn the center of the bounding box.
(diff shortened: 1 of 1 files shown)


pyvista/core/composite.py
```diff
@@ -1299,7 +1300,10 @@ def bounds(self: MultiBlock) -> BoundsTuple:
 
     @property
     def center(self: MultiBlock) -> tuple[float, float, float]:
-        """Return the center of the bounding box.
+        """Set or eturn the center of the bounding box.
```
akaszynski:
```suggestion
        """Set or return the center of the bounding box.
```

akaszynski (approved):

PR #8175, opened 2025-12-14
Add `length` option to `resize` filter

user27182:
### Overview

We can currently resize a mesh's `bounds` and `bounds_size`. I think it makes sense to extend this to allow resizing `length` directly.

### DIFF
--- a/pyvista/core/celltype.py
+++ b/pyvista/core/celltype.py
@@ -15,20 +15,11 @@ _CELL_TYPE_TO_NUM_POINTS: dict[np.uint8, int] = {}
 
 PLACEHOLDER = 'IMAGE-HASH-PLACEHOLDER'
 
-_DROPDOWN_TEMPLATE = """
-.. dropdown:: More info
-    :icon: info
-
-{}
-"""
-
 _GRID_TEMPLATE_NO_IMAGE = """
 .. grid:: 1
     :margin: 1
 
     .. grid-item::
-
-{}
 
 {}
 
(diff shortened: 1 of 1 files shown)


akaszynski:
Actions runner is failing as it's an older version. Upgrading...

pyvista/core/filters/data_object.py
```diff
@@ -863,8 +864,14 @@ def resize(  # type: ignore[misc]
         This filter rescales and translates the mesh to fit specified bounds. This is useful for
         normalizing datasets, changing units, or fitting datasets into specific coordinate ranges.
 
-        Use ``bounds`` to set the mesh's :attr:`~pyvista.DataSet.bounds` directly or use
-        ``bounds_size`` and ``center`` to implicitly set the new bounds.
+        It has three independent use cases:
+
+        #. Use ``bounds`` to set the mesh's :attr:`~pyvista.DataSet.bounds` directly
```
akaszynski: Ultra-nitpick for consistency.

```suggestion
        #. Use ``bounds`` to set the mesh's :attr:`~pyvista.DataSet.bounds` directly.
```

pyvista/core/filters/data_object.py
```diff
@@ -962,15 +985,28 @@ def resize(  # type: ignore[misc]
                     z_max =  0.5)
 
         """
+        if self.is_empty:
+            return self.copy()
+        bounds_set = bounds is not None
+        length_set = length is not None
+        bounds_size_set = bounds_size is not None
+        n_set = np.count_nonzero((bounds_set, length_set, bounds_size_set))
```
akaszynski: Another nitpick: I feel like `sum` would be fine here, or even:

``` suggestion
        n_set = bounds_set + length_set + bounds_size_set
```

akaszynski (approved):

PR #8176, opened 2025-12-15
Add `pv.to_trimesh` and `pv.from_trimesh` functions

user27182:
### Overview

Converting to Trimesh is a common task. There is a discussion about it here: https://github.com/pyvista/pyvista/discussions/2268

This PR adds `pv.to_trimesh` to facilitate the conversion. `pv.from_trimesh` is also added, which is adapted from the existing code used in `wrap`. `from_trimesh` extends the existing functionality by including data arrays in the conversion.

These new functions mirror existing `pv.to_meshio` and `pv.from_meshio` functions.

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -87,8 +87,8 @@ class DataObjectFilters:
         .. warning::
             Shear transformations are not supported for :class:`~pyvista.ImageData` or
             :class:`~pyvista.RectilinearGrid`, and rotations are not supported for
-            :class:`~pyvista.RectilinearGrid`. If present, these component(s) are removed by the
-            filter. To fully support these transformations, the input should be cast to
+            :class:`~pyvista.RectilinearGrid`. If present, a ``ValueError`` is raised.
+            To fully support these transformations, the input should be cast to
             :class:`~pyvista.StructuredGrid` `before` applying this filter.
 
         .. note::
@@ -108,6 +108,10 @@ class DataObjectFilters:
         .. versionchanged:: 0.46.0
             Transforming :class:`~pyvista.RectilinearGrid` now returns ``RectilinearGrid``.
             Previously, :class:`~pyvista.StructuredGrid` was returned.
+
+        .. versionchanged:: 0.47.0
+            An error is now raised instead of a warning if a transformation cannot be
+            applied.
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


pyvista/core/utilities/fileio.py
```diff
@@ -1340,3 +1348,242 @@ def _try_imageio_imread(filename: str | Path) -> imageio.core.util.Array:
...
+        trimesh_attrs = {
+            'vertices',
+            'faces',
+            'vertex_attributes',
+            'face_attributes',
+            'metadata',
+        }
+        if not all(hasattr(mesh, attr) for attr in trimesh_attrs):
+            msg = f'Mesh must be a Trimesh-like object. Got {type(mesh)} instead.'
+            raise TypeError(msg)
```
user27182: Not sure if this kind of duck-type checking is needed to avoid importing trimesh for the instance check. If you're converting from Trimesh object, you probably have trimesh installed, right?

doc/source/api/utilities/utilities.rst
```diff
@@ -19,10 +19,15 @@ Object Conversions or Wrapping
...
    cubemap_from_filenames
+   from_meshio
+   from_trimesh
    image_to_texture
+   is_meshio_mesh
    is_pyvista_dataset
    numpy_to_texture
    pyvista_ndarray
+   to_meshio
+   to_trimesh
```
user27182, line 31: These to/from methods are currently inside fileio.py. Might make more sense to put them inside helpers.py next to `wrap` since these really have nothing to do with reading or writing files.

pyvista/core/utilities/fileio.py
akaszynski: IMO it's sufficient to merely import `trimesh` for this function. These same attributes could exist in similar objects.

doc/source/api/utilities/utilities.rst
akaszynski: Let's do this in a follow-up.

pyvista/core/utilities/fileio.py
user27182: Ok. I'll remove the duck type check. In `wrap`, we check `__class__.__name__ == Trimesh`, which might also be fine.

user27182: I copied what's done with wrapping meshio objects and added an explicit `is_trimesh_mesh` function in [8cfb809](https://github.com/pyvista/pyvista/pull/8176/commits/8cfb80950125aaaaf9d5f89fbe37a99e41ebf9c8) and do proper instance checks now.

examples/00-load/wrap_trimesh.py
```diff
@@ -51,6 +56,48 @@
...
+# Use :func:`~pyvista.from_trimesh` to convert it back to a :class:`~pyvista.PolyData` mesh.
+# This is the same as using :func:`~pyvista.wrap`.
+pvmesh = pv.from_trimesh(tmesh)
+
+# The trimesh data is recovered as point data, cell data, and field data, including field data
+# stored in the :attr:`~pyvista.DataObject.user_dict`.
+print(pvmesh.point_data.keys())
+print(pvmesh.cell_data.keys())
+print(pvmesh.field_data.keys())
+print(pvmesh.user_dict)
```
user27182, line 104: While writing this example I noticed that `print` calls `vtkStringArray.__str__` instead of `_SerializedDictArray.__repr__` when called inside of a script. This differs from what's printed inside a docstring/doctest and `IPython`.

So I added `__str__` to `_SerializedDictArray`. This is a somewhat noisy change as it required updates to other tests. I can make this change in a separate PR if that's preferred to keep the diff a bit cleaner.

user27182: This was reverted and is addressed in #8177

akaszynski (dismissed):

user27182:
Here's the ugly user dict string from `print` fixed by #8177 
image

akaszynski (approved):

PR #8177, opened 2025-12-15
Implement `_SeralizedDictArray.__str__` for printing user dict

user27182:
### Overview

In some contexts (e.g. inside a script), `print` will call `vtkStringArray.__str__` instead of `_SerializedDictArray.__repr__`. See https://github.com/pyvista/pyvista/pull/8176#discussion_r2620510991 

This PR defines `str` to print the JSON formatted string, and, ~`repr` now prints a more debug-friendly representation to include the class name~ `repr` is unchanged.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -395,6 +395,12 @@ todo_include_todos = False
 from sphinx_gallery.sorting import FileNameSortKey
 
 
+def _filter_sphinx_gallery_warnings():
+    import warnings
+
+    warnings.simplefilter('error')
+
+
 class ResetPyVista:
     """Reset pyvista module to default settings."""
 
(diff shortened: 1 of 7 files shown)


pyvista/core/composite.py
```diff
@@ -873,7 +873,7 @@ def move_nested_field_data_to_root(
         dict is updated with the nested dict.
 
         >>> root.user_dict
-        {"foo": "bar"}
+        _SerializedDictArray({"foo": "bar"})
```
user27182: I'm not sure if I like this. Sure, it's more explicit, but it kind of exposes implementation details. Maybe it's better to make `__repr__` and `__str__` return the same JSON string? I.e. revert changes to `__repr__`

The main issue was to fix `print(mesh.user_dict)` after all...

user27182: repr and str and now equivalent in https://github.com/pyvista/pyvista/pull/8177/commits/67d9e660ca8815f7972a70ea3f57a665bd0e000b

akaszynski (approved):

PR #8178, opened 2025-12-15
Raise errors instead of warnings for unsupported transformations

user27182:
### Overview

See https://github.com/pyvista/pyvista/pull/7697#issuecomment-3657513853.


adeak (approved):
Thanks for the quick fix @user27182, great work! LGTM.

Issue #8105, opened 2025-11-19
Align reader extensions with ParaView


user27182:
### Describe what maintenance you would like added.

Our readers and extensions don't necessarily line up with ParaView. We should try to make them match (up to a point, since there are "ambiguous" formats that use the same extension for different readers).

I used ChatGPT to come up with a list of readers and extensions based on the ParaView source code, see full script:

<details>

``` py
import xml.etree.ElementTree as ET
import requests

# ----------------------------------------------------
# Configuration
# ----------------------------------------------------
COMMIT = "6b34f220bfc1213728de909a805d5c1d83dd149c"
REPO_API = f"https://api.github.com/repos/Kitware/ParaView/contents/Remoting/Application/Resources?ref={COMMIT}"
RAW_BASE = f"https://raw.githubusercontent.com/Kitware/ParaView/{COMMIT}/Remoting/Application/Resources/"

# ----------------------------------------------------
# 1. Get list of all readers*.xml files
# ----------------------------------------------------
resp = requests.get(REPO_API)
resp.raise_for_status()
files = resp.json()

reader_files = [
    item["name"]
    for item in files
    if item["name"].startswith("readers") and item["name"].endswith(".xml")
]

print(f"Found {len(reader_files)} reader XML files:")
for f in reader_files:
    print("  ", f)

# ----------------------------------------------------
# 2. Parse each XML and build the reader map
# ----------------------------------------------------
reader_map = {}

for fname in reader_files:
    url = RAW_BASE + fname
    print(f"Processing {fname} ...")
    xml_text = requests.get(url).text
    root = ET.fromstring(xml_text)

    # Build proxy-name → class map from <Proxy> nodes
    proxy_class = {}
    for p in root.findall(".//Proxy"):
        name = p.get("name")
        cls  = p.get("class")
        if name and cls:
            proxy_class[name] = cls

    # Parse <SourceProxy> nodes for ReaderFactory hints
    for sp in root.findall(".//SourceProxy"):
        hints = sp.find("Hints")
        if hints is None:
            continue
        rf = hints.find("ReaderFactory")
        if rf is None:
            continue

        exts = rf.get("extensions", "").strip()
        desc = rf.get("file_description", "").strip()

        # If SourceProxy wraps a core Proxy via SubProxy
        core = None
        sub = sp.find(".//SubProxy/Proxy")
        if sub is not None:
            pname = sub.get("proxyname")
            core  = proxy_class.get(pname)

[...]

PR #8184, opened 2025-12-20
Add SeriesReader for `.series` file

### LINKED ISSUE #8179: Support `.series` File Extension
## Summary

Hi! I propose adding support for the `.series` file extension in `pyvista.get_reader` function.

(Related to #8105)

## Background

ParaView supports a meta file format based on JSON (`*.series`) for specifying time values. A typical file looks like:

{
  "file-series-version": "1.0",
  "files": [
    { "name": "foo_0000.vtu", "time": 0.0 },
    { "name": "foo_0001.vtu", "time": 0.1 }
  ]
}

Ref: https://docs.paraview.org/en/latest/UsersGuide/dataIngestion.html#handling-temporal-file-series

Currently, `pyvista.get_reader("case.vtu.series")` does not recognize the `.series` extension, so users cannot load such time-series datasets through PyVista.

## Direction

Parview defines `vtkFileSeriesReader` which is a extension class to handle `.series` files (See below). So, it seems we also have to define the new reader class to handle `.series` files.

https://github.com/Kitware/ParaView/blob/6a405b6709ff4181778551120826dd0b0c196510/VTKExtensions/IOCore/vtkFileSeriesReader.cxx#L45

If it is acceptable, I am happly to open PR to implement this feature.

## Current Behaviour

When attempting to read a `.series` file using `pyvista.get_reader`, the following error is raised:

```python
import pyvista as pv

filename = "path/to/your/mesh.vtu.series"
```

output:

```
  File "/home/riku/repos/personals/pyvista/workspace/issue20251216/sample.py", line 15, in <module>
    reader = pv.get_reader("workspace/issue20251216/data/mesh.vtu.series")
  File "/home/riku/repos/personals/pyvista/pyvista/core/utilities/reader.py", line 94, in get_reader
    raise ValueError(msg)
ValueError: `pyvista.get_reader` does not support a file with the .series extension
```

_No response_

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

riku-sakamoto:
### Overview

Closes #8179 

* Add a new reader for `.series` files

### Details

* Following to @MatthewFlamm 's instruction, `SeriesReader` is implemented with a structure similar to `PVDReader`, reusing the same design pattern.
* As discussed in #8179, a `Generic` type is introduced in `SeriesReader` to distinguish the return value of the `read` function.

### Future work

* As reported in #8105, several file extensions are supported by the `.series` format. To minimize the scope of this PR, only tests for `.vtu.series` are prepared. To validate additional extensions, we need to add examples and tests. (Updated 2025/12/23)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -400,6 +400,7 @@ extend-select = [
   'RET',
   'RSE',
   'RUF',
+  'S101',
   'SIM',
   'T10',
   'T20',
@@ -460,6 +461,7 @@ allow-dict-calls-with-keyword-arguments = true
   'PTH113',
   'PTH118',
   'PTH123',
+  'S101',
   'T20',
 ]
 'examples*' = ['ANN', 'INP', 'PLC0415']
@@ -473,12 +475,13 @@ allow-dict-calls-with-keyword-arguments = true
   'D400',
   'D415',
   'PLC0415', # https://github.com/pyvista/pyvista/pull/7676
+  'S101',
   'SIM115',
   'T20',
 ]
 'examples/99-advanced/warp_by_vector_eigenmodes.py' = ['N802', 'N803']
 'examples_trame/*' = ['D100', 'D103', 'T20']
-'examples_trame/tests/*' = ['D']
+'examples_trame/tests/*' = ['D', 'S101']
 'hooks/*' = ['INP001']
 'pyvista/__init__.py' = ['ANN']
 'pyvista/conftest.py' = ['ANN']
@@ -512,6 +515,7 @@ allow-dict-calls-with-keyword-arguments = true
   'PTH117',
   'PTH118',
   'PTH202',
+  'S101',
 ]
 'pyvista/examples/cells.py' = ['N802']
 'pyvista/ext/*' = ['ANN', 'D100', 'D101', 'D102', 'D103']
@@ -559,7 +563,7 @@ allow-dict-calls-with-keyword-arguments = true
 'pyvista/report.py' = ['ANN']
 'pyvista/trame/*' = ['ANN', 'FBT', 'PLR0917']
 'pyvista/utilities/*' = ['ANN', 'F401'] # deprecated modules
-'tests/*' = ['ANN', 'D', 'FBT', 'INP', 'PLC0415', 'PLR0917', 'T20']
+'tests/*' = ['ANN', 'D', 'FBT', 'INP', 'PLC0415', 'PLR0917', 'S101', 'T20']
 'tests/examples/test_dataset_loader.py' = [
   'PTH102',
   'PTH107',
--- a/pyvista/core/pointset.py
+++ b/pyvista/core/pointset.py
@@ -27,6 +27,8 @@ from .cell import _get_offset_array
 from .cell import _get_regular_cells
 from .celltype import CellType
 from .dataset import DataSet
+from .dataset import _MeshValidationReport
+from .dataset import _MeshValidator
 from .errors import CellSizeError
 from .errors import PointSetCellOperationError
 from .errors import PointSetDimensionReductionError
@@ -512,6 +514,29 @@ class PointSet(_PointSet, _vtk.vtkPointSet):
     def extract_geometry(self, *args, **kwargs):  # noqa: ARG002
         """Raise extract geometry are not supported."""
         raise PointSetCellOperationError
+
+    def cell_validator(self, *args, **kwargs):  # noqa: ARG002
+        """Raise cell operations are not supported."""
+        raise PointSetCellOperationError
+
+    @wraps(DataSet.validate_mesh)
+    def validate_mesh(  # type: ignore[override]  # numpydoc ignore=RT01
+        self,
+        validation_fields: _MeshValidator._AllValidationOptions
+        | Sequence[_MeshValidator._AllValidationOptions]
+        | None = None,
+        *args,
+        **kwargs,
+    ) -> _MeshValidationReport:
+        """Wrap validate_mesh with cell-related fields removed."""
+        if validation_fields is None:
+            fields: list[_MeshValidator._AllValidationOptions] = [
+                *_MeshValidator._allowed_data_fields,
+                *_MeshValidator._allowed_point_fields,
+            ]
+            fields.remove('unused_points')
+            return DataSet.validate_mesh(self, fields, *args, **kwargs)
+        return DataSet.validate_mesh(self, validation_fields, *args, **kwargs)
 
 
 class PolyData(_PointSet, PolyDataFilters, _vtk.vtkPolyData):
--- a/pyvista/jupyter/notebook.py
+++ b/pyvista/jupyter/notebook.py
@@ -12,8 +12,7 @@ Includes:
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
-
-import numpy as np
+from typing import cast
 
 from pyvista._warn_external import warn_external
 
@@ -24,6 +23,7 @@ if TYPE_CHECKING:
     from IPython.lib.display import IFrame
     from PIL.Image import Image
 
+    from pyvista import pyvista_ndarray
     from pyvista.jupyter import JupyterBackendOptions
     from pyvista.plotting.plotter import Plotter
     from pyvista.trame.jupyter import EmbeddableWidget
@@ -72,5 +72,5 @@ def show_static_image(
         # Must render here, otherwise plotter will segfault.
         plotter.render()
         plotter.last_image = plotter.screenshot(screenshot, return_img=True)
-    assert isinstance(plotter.last_image, np.ndarray)
-    return PIL.Image.fromarray(plotter.last_image)
+    last_image = cast('pyvista_ndarray', plotter.last_image)
+    return PIL.Image.fromarray(last_image)
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -4462,7 +4462,7 @@ class BasePlotter(_BoundsSizeMixin, PickingHelper, WidgetHelper):
             #       Also, place all data on the nodes as issues arise when
             #       volume rendering on the cells.
             volume = volume.cell_data_to_point_data()
-        assert isinstance(volume, (pv.DataSet, pv.MultiBlock))
+        assert isinstance(volume, (pv.DataSet, pv.MultiBlock))  # noqa: S101
 
         if name is None:
             name = f'{type(volume).__name__}({volume.memory_address})'
--- a/tests/core/test_dataset.py
+++ b/tests/core/test_dataset.py
@@ -1801,6 +1801,53 @@ def test_validate_mesh_str_filtered():
     assert actual == expected
 
 
+def test_validate_mesh_pointset(ant):
+    pset = ant.cast_to_pointset()
+    report = pset.validate_mesh()
+    actual = str(report)
+    expected = (
+        'Mesh Validation Report\n'
+        '----------------------\n'
+        'Mesh:\n'
+        '    Type                     : PointSet\n'
+        '    N Points                 : 486\n'
+        '    N Cells                  : 0\n'
+        'Report summary:\n'
+        '    Is valid                 : True\n'
+        '    Invalid fields           : ()\n'
+        'Invalid data arrays:\n'
+        '    Point data wrong length  : []\n'
+        '    Cell data wrong length   : []\n'
+        'Invalid point ids:\n'
+        '    Non-finite points        : []'
+    )
+    assert actual == expected
+
+    report = pset.validate_mesh('data')
+    actual = str(report)
+    expected = (
+        'Mesh Validation Report\n'
+        '----------------------\n'
+        'Mesh:\n'
+        '    Type                     : PointSet\n'
+        '    N Points                 : 486\n'
+        '    N Cells                  : 0\n'
+        'Report summary:\n'
+        '    Is valid                 : True\n'
+        '    Invalid fields           : ()\n'
+        'Invalid data arrays:\n'
+        '    Point data wrong length  : []\n'
+        '    Cell data wrong length   : []'
+    )
+    assert actual == expected
+
+
+def test_cell_validator_pointset_raises():
+    match = 'Cell operations are not supported'
+    with pytest.raises(pv.PointSetCellOperationError, match=match):
+        pv.PointSet().cell_validator()
+
+
 def test_cell_validator():
     validator_array_names = list(_CELL_VALIDATOR_BIT_FIELD.keys())
     sphere = pv.Sphere()


riku-sakamoto:
I am going to add test data to pyvista/vtk_data, so this will be a WIP for a while.

riku-sakamoto:
I've finished adding the methods and tests, so I've marked this PR as ready.

@MatthewFlamm The `SeriesReader` class is implemented with almost the same structure as `PVDReader`, following your initial suggestion. However, if you would prefer moving to a functional version, I'm happy to do that.

pyvista/core/utilities/reader.py
```diff
@@ -3825,4 +4007,5 @@ def array_status(self, name):
     XMLStructuredGridReader: 'StructuredGrid',
     XMLUnstructuredGridReader: 'UnstructuredGrid',
     XMLPImageDataReader: 'ImageData',
+    SeriesReader[XMLUnstructuredGridReader]: ('UnstructuredGrid', 'MultiBlock'),
```
user27182: Is this correct? The reader can return a single `UnstructuredGrid` _or_ a `MultiBlock`?

tests/core/test_utilities.py
```diff
@@ -3058,7 +3058,7 @@ def test_writer_data_mode_mixin(writer_cls):
 def test_fileio_extensions(cls):
     if cls is pv.HDFWriter and pv.vtk_version_info < (9, 4, 0):
         pytest.xfail('Needs vtk 9.4')
-    if cls in [pv.OpenFOAMReader, pv.MultiBlockPlot3DReader]:
+    if cls in [pv.OpenFOAMReader, pv.MultiBlockPlot3DReader, pv.SeriesReader]:
         # These classes are not associated with any extensions
```
user27182: These exceptions are more for legacy readers. For new readers, we should associate them with a class, e.g. add it ~`MultiBlock._WRITERS`~ `CLASS_READERS`

EDIT: Is it not possible to add a `.series` mapping to `CLASS_READERS`, or is it not because of the wildcard `*.series` extension format?

EDIT2: Could be an issue with the test too, e.g. the `*READER_CLASSES` collection for the test may be making assumptions about the file extensions that don't apply to the wildcard extension format. Or maybe the generic typing of `SeriesReader[XMLUnstructuredGridReader]`is throwing off the test. But the basically the test should pass if there is a mapping from file extension to reader in `CLASS_READERS`

pyvista/core/utilities/reader.py
riku-sakamoto: Thanks for the review.  You're right, that's not correct. I'll fix it.

In the current implementation, `SeriesReader` can have multiple readers per time step (similar to `PVDReader`). However, after checking ParaView’s behavior, only one dataset per time step is actually used; if multiple datasets are present, the additional ones are ignored.

Accordingly, I’ll update `SeriesReader[XMLUnstructuredGridReader].read()` to return an `UnstructuredGrid` rather than a `MultiBlock`.

tests/core/test_utilities.py
riku-sakamoto: > Is it not possible to add a .series mapping to CLASS_READERS, or is it not because of the wildcard *.series extension format?

It is possible. I chose this approach because it allows us to express more precise type hints per concrete series reader, which could be useful for future typing improvements.

For example, this allows us to distinguish:

```
SeriesReader[XMLUnstructuredGridReader]: UnstructuredGrid
SeriesReader[VTKDataSetReader]: get_args(_legacy_dataset_types)
```

instead of a single, more generic mapping like:

```
SeriesReader: (UnstructuredGrid, *get_args(_legacy_dataset_types))
```

If you think this extra typing precision is unnecessary, I’m happy to squash this into a simpler mapping and amend the tests too.

user27182: I think a generic typing is fine, but this test isn't really about the type hints, but more about ensuring the mapping can be used to generate documentation. E.g. have a look at the table for `pyvista.read` https://dev.pyvista.org/api/utilities/_autosummary/pyvista.read.html

The entries are generated based on the mapping from `CLASS_READERS`.

I have not reviewed the docs (there should be an HTML preview artifact generated by the CI build), but the new reader should show up somewhere in that table. If this test is failing then I don't think it'll be part of the table.

riku-sakamoto: Thanks for the explanation! I'll switch to a simpler mapping.

pyvista/core/utilities/reader.py
riku-sakamoto: I’ve completed updating it.

tests/core/test_utilities.py
riku-sakamoto: I’ve completed updating it too.

pyvista/core/utilities/reader.py
user27182: > In the current implementation, `SeriesReader` can have multiple readers per time step (similar to `PVDReader`). However, after checking ParaView’s behavior, only one dataset per time step is actually used; if multiple datasets are present, the additional ones are ignored.
> 
> Accordingly, I’ll update `SeriesReader[XMLUnstructuredGridReader].read()` to return an `UnstructuredGrid` rather than a `MultiBlock`.

I checked out this branch locally and tried `pv.read()` on the series file and compare it to `pv.read()` with the similar PVD file.
I noticed that reading PVD file returns a `MultiBlock` with a single dataset, whereas reading the series file returns a single `UnstructuredGrid`:

``` py
import pyvista as pv
from pyvista import examples

series_file = examples.download_file('vtu_series/wavy.zip')[0]
pvd_file = examples.download_wavy(load=False)

assert series_file.endswith('.series')
assert pvd_file.endswith('.pvd')

assert isinstance(pv.read(series_file), pv.UnstructuredGrid)
assert isinstance(pv.read(pvd_file), pv.MultiBlock)
```

When I read these files in ParaView, I do not get a `MultiBlock` explicitly, but get a separate dataset for each time step. With the series file, these are individual `UnstructuredGrid` files. With the pvd file, these are curvilinear `StructuredGrid` files.

So, the default PyVista behavior is currently:
- for pvd files, read only a single active time step by default as a `MultiBlock`
- for series files, read only a single active time step as the mesh itself, _not_ as a `MultiBlock`

This means these two similar readers currently have dissimilar behaviors. Should they be more aligned, and allow reading multiple active time steps into a `MultiBlock`, like the PVD reader? This would make the `CLASS_READERS` mapping as `SeriesReader: 'MultiBlock'`.

To make this even _more_ confusing, it seems to me that the default should actually be to read _all_ timesteps into a single `MultiBlock` by default, no? So that all data is loaded and available by default without the need to explicitly use the reader? I haven't worked with time series data in the real world though, so am not really sure how these kinds of files generally behave in the wild.

Maybe @MatthewFlamm can comment on what the expected behavior should be. Should a `MultiBlock` always be returned? Should it only include one block by default?

user27182: > To make this even _more_ confusing, it seems to me that the default should actually be to read _all_ timesteps into a single `MultiBlock` by default, no? So that all data is loaded and available by default without the need to explicitly use the reader? I haven't worked with time series data in the real world though, so am not really sure how these kinds of files generally behave in the wild.
> 
> Should a `MultiBlock` always be returned? Should it only include one block by default?

[vtkHDFReader]( https://vtk.org/doc/nightly/html/classvtkHDFReader.html#details) supports time series data, and by default it appears to only read a single time step. E.g. loading [download_warping_spheres](https://dev.pyvista.org/api/examples/_autosummary/pyvista.examples.downloads.download_warping_spheres) returns `PartitionedDataSet` with two partitions (there are two "warped" spheres in the dataset). And so only reading a single time step by default appears to be the "correct" default, and using the reader is necessary to access other time steps.


EDIT: But in this case, the single time step mesh is loaded on its own, it's not loaded inside a separate composite dataset, where each leaf is a separate time step (which is what the pvd reader does)

And so it seems that loading time steps into a `MultiBlock` is a PyVista construct, which is inconsistent with the `vtkHDFReader`. If we keep the current `SeriesReader` mapping of `pv.read('wavy.vtu.series')` -> `UnstructuredGrid`, this is consistent with `vtkHDFReader`, but inconsistent with `PVDReader`.

riku-sakamoto: @user27182 
Thank you for taking a closer look! 

* Regarding the inconsistency between SeriesReader and PVDReader:

It might be beneficial if return type of SeriesReader depends on the extention specified in `{ext}.series`. For example, if the file is `case.vtu.series`, SeriesReader could return an `UnstructuredGrid`, whereas `case.vtm.series` could return a `MultiBlock`.
So far, I have prepared an example for `vtu.series` only, so return type is fixed to `UnstructuredGrid` in the current implementation.

That said, if PyVista prefers to keep the behavior simple and consistent, I am happy to modify the implementation so that SeriesReader always returns MultiBlock, regardless of the extensions inside the `.series` file.

I also think PVDReader is a bit special reader because it can contain multiple types of datasets in one file. So, it might be reasonable that PVDReader always returns MultiBlock.


If we decide to seperate return type of SeriesReader based on the extension inside `.series` file, would it be better to rename `SeriesReader` to more general name like `_GenericSeriesReader` and create new `VTUSeriesReader` ?

```python

class VTUSeriesReader(BaseReader, TimeReader, _GenericSeriesReader[XMLUnstructuredGridReader]):
    ...


CLASS_READERS = {
    '.vtu.series': VTUSeriesReader,
}

_CLASS_READER_RETURN_TYPE = {
    VTUSeriesReader: 'UnstructuredGrid',
}
```

MatthewFlamm: I have not had time to read the changes here on detail yet.

The reader should return a single time step of data.  This is how Paraview works and the pipeline in vtk typically works (although PyVista doesn't really use this model directly).

On return type, I would lean on the behavior in ParaView. I suspect, although cannot confirm right now, that ParaView returns Multiblock with the PVD file and a single mesh like UnstructuredGrid with the other formats.  The PVD file can include much more complexity in structure than the series format.  I would be surprised if it changes type based on the special case of 1 file per time step, but this needs to be confirmed. 

I had suggested a simple functional version initially thinking that the implementation of the extension type to return type matching would be difficult.  Since it seems to be implemented here, this is a nice result.

user27182: > The reader should return a single time step of data. This is how Paraview works and the pipeline in vtk typically works (although PyVista doesn't really use this model directly).

Indeed this is currently the case currently for the `.series` and `.pvd` readers for PyVista and Paraview. No problems here.

> On return type, I would lean on the behavior in ParaView. I suspect, although cannot confirm right now, that ParaView returns Multiblock with the PVD file and a single mesh like UnstructuredGrid with the other formats. The PVD file can include much more complexity in structure than the series format. I would be surprised if it changes type based on the special case of 1 file per time step, but this needs to be confirmed.

I confirmed this behavior earlier. Maybe I am not looking in the right place, but this was my assessment:

> When I read these files in ParaView, I do not get a MultiBlock explicitly, but get a separate dataset for each time step. With the series file, these are individual UnstructuredGrid files. With the pvd file, these are curvilinear StructuredGrid files.
>
> So, the default PyVista behavior is currently:
>
> - for pvd files, read only a single active time step by default as a MultiBlock
> - for series files, read only a single active time step as the mesh itself, not as a MultiBlock

Returning a `MultiBlock` for the PVD reader is a PyVista construct, resulting from this code, where the PVD reader _always_ stores `self._data_object` as a `MultiBlock`:

https://github.com/pyvista/pyvista/blob/8cfe9b69637ad140376768526d208f18d967cc7b/pyvista/core/utilities/reader.py#L1945-L1947

and `reader.read` returns `self._data_object` via wrapped calls to `GetOutputDataObject` here:

https://github.com/pyvista/pyvista/blob/8cfe9b69637ad140376768526d208f18d967cc7b/pyvista/core/utilities/reader.py#L139-L141

https://github.com/pyvista/pyvista/blob/8cfe9b69637ad140376768526d208f18d967cc7b/pyvista/core/utilities/reader.py#L291

I believe this may have been done because the PyVista PVD reader maybe allows setting multiple active time steps? So that calling `.read` _may_ return multiple time steps, though only one is returned by default.

> I had suggested a simple functional version initially thinking that the implementation of the extension type to return type matching would be difficult. Since it seems to be implemented here, this is a nice result.

I agree it's nice that `.vtu.series` returns  a single `UnstructuredGrid` time step, `.vti.series` returns a single `ImageData` time step ,`.vtm.series` returns `MultiBlock`, etc.. This appears to be consistent with ParaView. The question is though, do we want the PyVista `.series` reader to do what ParaView does, or do we want the PyVista `.series` reader to do what the PyVista `.pvd` reader does, which allows reading multiple time steps at once, and therefore _always_ returns a `MultiBlock`. (In the case of reading a `.vtm.series` file, this would return a _nested_ `MultiBlock`, i.e. a `MultiBlock` of time steps, where each time step is a `MultiBlock`).

MatthewFlamm: PVD reader does not return a MultiBlock because it returns multiple time steps, but because it _can_ return multiple meshes in a single time point

MatthewFlamm: The example download `wavy` for PVD just happens to have 1 mesh per time point.

MatthewFlamm: So I think everything is consistent with the current PR state.  A `series` file always has 1 file/mesh per time step, thus it returns the base dataset type.  A `pvd` file generally has more than 1 file/mesh per time step, thus it must return a MultiBlock.  For `pvd` in ParaView, I'm not sure whether it special cases when there is only 1 file/mesh  per time step, but it seems unlikely to me.  If it does, then we can consider fixing that reader.

user27182: I see, so the `.series` format differs in this manner, and can only ever have one mesh per time point? If this is the case, then the status quo with `.vtu.series` returning `UnstructuredGrid` is totally fine. But then we should probably update the class mappings so that _all_ possible mesh types are included, e.g.:

e.g. 
``` py
CLASS_READERS = {
    '.series': SeriesReader,
}

_CLASS_READER_RETURN_TYPE = {
    SeriesReader: get_args(_mesh_types)
```

This should make it so that `SeriesReader` shows up in every dataset type's table in the `pv.read` docs https://dev.pyvista.org/api/utilities/_autosummary/pyvista.read

As  suggested by @riku-sakamoto earlier, we _could_ alternatively make separate subclasses:
``` py
class VTUSeriesReader(BaseReader, TimeReader, _GenericSeriesReader[XMLUnstructuredGridReader]):
    ...

CLASS_READERS = {
    '.vtu.series': VTUSeriesReader,
}

_CLASS_READER_RETURN_TYPE = {
    VTUSeriesReader: 'UnstructuredGrid',
}
```

But, IMO I think lumping all `.series` files together with the single reader is fine for now and covers all cases. Otherwise, we would need to explicitly cover cases like `.stl.series` for basically every file extension.

So, in terms of resolving this thread and moving this PR forward, I think we just need to update this line in `_CLASS_READER_RETURN_TYPE`, since _any_ mesh type may be returned based on the file extension preceding the `.series` extension.

``` diff
- SeriesReader: 'UnstructuredGrid',
+ SeriesReader: get_args(_mesh_types),
```

user27182: > A `pvd` file generally has more than 1 file/mesh per time step, thus it must return a MultiBlock. For `pvd` in ParaView, I'm not sure whether it special cases when there is only 1 file/mesh per time step, but it seems unlikely to me. If it does, then we can consider fixing that reader.

ParaView does have special branching for the single-mesh time step case. I have documented this here https://github.com/pyvista/pyvista/issues/8193

pyvista/core/utilities/reader.py
```diff
@@ -3678,6 +3682,179 @@ def array_status(self, name):
...
+        self._reader_type = self._deterimine_reader_type()
+        self._time_values = sorted({dataset.time for dataset in self._datasets})
+        self._time_mapping: dict[float, SeriesDataSet] = {
+            dataset.time: dataset for dataset in self._datasets
+        }
+        self._SetActiveTime(self._time_values[0])
+
+    def Update(self) -> None:
+        """Read data and store it."""
+        assert self._active_reader is not None
```
user27182: We should avoid assertions in code. This is enforced in #8194

tests/test_attributes.py
```diff
@@ -265,7 +266,11 @@ def skip_test_for_some_classes():
             pytest.skip('Test fails without proper dataset files.')
         elif pyvista_class is pv.core.dataset.ActiveArrayInfo:
             pytest.skip('Deprecated.')
-        elif pyvista_class in (pv.PVDDataSet, pv.core.utilities.cell_quality.CellQualityInfo):
+        elif pyvista_class in (
+            pv.SeriesDataSet,
+            pv.PVDDataSet,
+            pv.core.utilities.cell_quality.CellQualityInfo,
+        ):
```
user27182: We can add this the module imports
``` py
from dataclasses import is_dataclass
```
and then just do this check instead:

```suggestion
        elif is_dataclass(pyvista_class):
```

pyvista/core/utilities/reader.py
riku-sakamoto: @user27182 Thanks. I've updated `_CLASS_READER_RETURN_TYPE` as you suggested.

pyvista/core/utilities/reader.py
riku-sakamoto: Those assert statements were only for satisfying mypy, so I’ve replaced them with using `cast` function instead.

tests/test_attributes.py
riku-sakamoto: Thanks! I've updated it too.

riku-sakamoto:
Thank you for the review and merge! @user27182 @MatthewFlamm @akaszynski

PR #8190, opened 2025-12-25
Add file_version property to XML writers

tkoyama010:
### Overview

This PR adds a `file_version` property to XML writers that allows users to get and set the VTK XML file format version (major and minor).

Different VTK XML file versions may support different features or compression methods. This property provides a convenient way to control the file version when writing XML files.

### Details

- Add `file_version` property (getter and setter) to `_XMLWriter` base class
- Validate input: must be a tuple of two integers (major, minor)
- Gracefully handle VTK versions that don't support these methods (returns `(None, None)` for getter, raises `AttributeError` for setter)
- Add comprehensive tests for both normal usage and validation of invalid inputs


pyvista/core/utilities/writer.py
```diff
@@ -498,6 +498,67 @@ def compression(self, compression: _CompressionOptions) -> None:
...
             self.writer.SetCompressorTypeToLZMA()
 
+    @property
+    def file_version(self) -> tuple[int, int]:
+        """Get or set the file version for VTK XML files.
+
+        The file version determines the VTK XML file format version.
+        Different versions may support different features or compression methods.
+
+        .. versionadded:: 0.48.0
```
tkoyama010, line 512:
```suggestion
        .. versionadded:: 0.47.0
```

pyvista/core/utilities/writer.py
```diff
@@ -11,6 +11,7 @@
 
 import numpy as np
 
+import pyvista
```
tkoyama010:
```suggestion
import pyvista as pv
```

pyvista/core/utilities/writer.py
```diff
@@ -498,6 +499,70 @@ def compression(self, compression: _CompressionOptions) -> None:
...
+        _validation.check_instance(major, int, name='major version')
+        _validation.check_instance(minor, int, name='minor version')
+
+        if not (
+            hasattr(self.writer, 'SetDataSetMajorVersion')
+            and hasattr(self.writer, 'SetDataSetMinorVersion')
+        ):
+            msg = (
+                'Setting file version requires VTK 9.3 or later. '
+                f'Current VTK version is {pyvista.vtk_version_info}.'
```
tkoyama010:
```suggestion
                f'Current VTK version is {pv.vtk_version_info}.'
```

pyvista/core/utilities/writer.py
```diff
@@ -501,6 +503,70 @@ def compression(self, compression: _CompressionOptions) -> None:
...
             self.writer.SetCompressorTypeToLZMA()  # type: ignore[attr-defined]
 
+    @property
+    def file_version(self) -> tuple[int, int] | tuple[None, None]:
+        """Get or set the file version for VTK XML files.
+
+        The file version determines the VTK XML file format version.
+        Different versions may support different features or compression methods.
+
+        .. versionadded:: 0.47.0
```
tkoyama010:
```suggestion
        .. versionadded:: 0.48.0
```

pyvista/core/utilities/writer.py
```diff
@@ -501,6 +503,70 @@ def compression(self, compression: _CompressionOptions) -> None:
...
+        >>> writer = pv.XMLPolyDataWriter('sphere.vtp', mesh)
+        >>> writer.file_version = (2, 2)  # doctest:+SKIP
+        >>> writer.file_version  # doctest:+SKIP
+        (2, 2)
+
+        """
+        if not (
+            hasattr(self.writer, 'GetDataSetMajorVersion')
+            and hasattr(self.writer, 'GetDataSetMinorVersion')
+        ):
```
tkoyama010:
```suggestion
        ):  # pragma: no cover
```

pyvista/core/utilities/writer.py
```diff
@@ -501,6 +503,70 @@ def compression(self, compression: _CompressionOptions) -> None:
...
+            msg = 'file_version must be a tuple of two integers (major, minor)'
+            raise ValueError(msg)
+        major, minor = version
+        _validation.check_instance(major, int, name='major version')
+        _validation.check_instance(minor, int, name='minor version')
+
+        if not (
+            hasattr(self.writer, 'SetDataSetMajorVersion')
+            and hasattr(self.writer, 'SetDataSetMinorVersion')
+        ):
```
tkoyama010:
```suggestion
        ):  # pragma: no cover
```

pyvista/core/utilities/writer.py
```diff
@@ -11,6 +11,8 @@
 
 import numpy as np
 
+import pyvista as pv
+from pyvista._warn_external import warn_external
```
tkoyama010:
```suggestion
```

pyvista/core/utilities/writer.py
```diff
@@ -501,6 +502,70 @@ def compression(self, compression: _CompressionOptions) -> None:
...
+
+        Examples
+        --------
+        Set the file version for an XML writer.
+
+        >>> import pyvista as pv
+        >>> mesh = pv.Sphere()
+        >>> writer = pv.XMLPolyDataWriter('sphere.vtp', mesh)
+        >>> writer.file_version = (2, 2)  # doctest:+SKIP
+        >>> writer.file_version  # doctest:+SKIP
```
user27182, line 533: why skip? When I run this example locally with VTK 9.6, I get `(None, None)`, seems something isn't working?
``` py
import pyvista as pv
writer = pv.XMLPolyDataWriter('sphere.vtp', pv.Sphere())
print(writer.file_version)  # (None, None)
writer.writer.GetDataSetMajorVersion()  # AttributeError
```

pyvista/core/utilities/writer.py
```diff
@@ -501,6 +502,70 @@ def compression(self, compression: _CompressionOptions) -> None:
...
+            raise ValueError(msg)
+        major, minor = version
+        _validation.check_instance(major, int, name='major version')
+        _validation.check_instance(minor, int, name='minor version')
+
+        if not (
+            hasattr(self.writer, 'SetDataSetMajorVersion')
+            and hasattr(self.writer, 'SetDataSetMinorVersion')
+        ):  # pragma: no cover
+            msg = (
```
user27182, line 560: Why no cover? We should have coverage for this automatically for both older vtk (< 9.3) and newer if it's only missing for < 9.3.

Recommend also using `pv.vtk_version_info < 9.3` instead, since that will automatically raise an error when < 9.3 is no longer support and hence will help us remove this version branch later.

pyvista/core/utilities/writer.py
```diff
@@ -501,6 +502,70 @@ def compression(self, compression: _CompressionOptions) -> None:
...
+            hasattr(self.writer, 'GetDataSetMajorVersion')
+            and hasattr(self.writer, 'GetDataSetMinorVersion')
+        ):  # pragma: no cover
+            return (None, None)
+        major = self.writer.GetDataSetMajorVersion()
+        minor = self.writer.GetDataSetMinorVersion()
+        return (major, minor)
+
+    @file_version.setter
+    def file_version(self, version: tuple[int, ...]) -> None:
```
user27182, line 547: Seems odd to me that users can set this to any arbitrary value. Surely this should be restricted to specific values supported by VTK? Otherwise it seems users can set version to `(42, 99)` if they want?

Moreover, looking at the class members for "S" https://vtk.org/doc/nightly/html/functions_s.html#index_s, we can see that only the hyper tree grid writer supports this, so this setter basically doesn't work at all for any of our writers.

Image

tests/core/test_utilities.py
```diff
@@ -3121,3 +3121,35 @@ def test_ply_writer(sphere, tmp_path):
...
+    writer = pv.XMLPolyDataWriter(path, sphere)
+
+    # Test normal case: setting valid file_version values
+    try:
+        writer.file_version = (2, 2)
+        assert writer.file_version == (2, 2)
+        writer.file_version = (1, 0)
+        assert writer.file_version == (1, 0)
+    except AttributeError:
+        # Skip if VTK version doesn't support file_version setter
```
user27182, line 3150: When I run this test locally with VTK 9.6, everything is skipped. This would explain the missing coverage.
Image


We should not unconditionally skip this test on attribute error. We should instead use `@pytest.mark.needs_vtk_version(9,3,0)`, and then test the attribute error being raised with a separate test.

Or, include vtk version-specific branches inside the test.

user27182 (changes requested):
Seems like this feature is a bit broken and isn't currently working as expected.

PR #6901, opened 2024-11-30
Document `ImageData` inherited members `dimensions` and `dimensionality`

user27182:
### Overview

`ImageData.dimensions` and `ImageData.dimensionality` is not documented. This is because it inherits this from `Grid`, but `Grid` is not documented either. This PR fixes this by showing all inherited members for `ImageData`.

PR #8192, opened 2025-12-26
Do not document inherited `ImageData` members

user27182:
### Overview

Documenting all `ImageData` members was initially added in https://github.com/pyvista/pyvista/pull/6901. This was done because inherited members are not documented, so any abstract members are simply not documented unless the abstract parent class itself is documented. This means that things like `Grid.dimensionality` were not documented, because the `Grid` class is not documented.

However, `dimensionality` is now generalized, and is a documented as [`DataSet.dimensionality`](https://dev.pyvista.org/api/core/_autosummary/pyvista.dataset.dimensionality). So there is no longer a need to document `ImageData.dimensionality` specifically.

In addition, there is also an undesired side-effect where _all_ `ImageData` vtk methods are also currently inadvertently documented, e.g. `min_spatial_dimension` is pure VTK property
`https://dev.pyvista.org/api/core/_autosummary/pyvista.imagedata.min_spatial_dimension` which results in an error:

``` py
import pyvista as pv
pv.ImageData().min_spatial_dimension
```
```
pyvista.core.errors.PyVistaAttributeError: The attribute 'min_spatial_dimension' is defined by VTK and is not part of the PyVista API. Did you mean: 'max_spatial_dimension'?
```

This reinforces that we should not document _all_ members like this.


banesullivan:
Just checked the dev docs and indeed still an issue. Updating branch to see if we can land this

user27182:
> Just checked the dev docs and indeed still an issue. Updating branch to see if we can land this

the broader issue is that our docs do not document parent properties or methods for child classes, which is a problem when thr parent class isn't documented.  There is an existing issue about this I believe @MatthewFlamm created some time ago.

and so if we were simply able to document the abstract dimensionality property without also exposing the private _PointGrid class, that would help. Else we just duplicate the property manually across all documented classes.

there are other such properties where this is an issue, e.g. private mixins such as the `bounds_size` property for meshes, or `extensions` property for readers / writers. None of these public API are documented, but should be.

PR #8194, opened 2025-12-27
Do not allow assertions in code

user27182:
### Overview

See https://docs.astral.sh/ruff/rules/assert/#assert-s101
Exceptions are made for cases like tests or documentation code.

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -4456,12 +4456,13 @@ class BasePlotter(_BoundsSizeMixin, PickingHelper, WidgetHelper):
                     msg = 'Invalid resolution dimensions.'
                     raise ValueError(msg)
                 volume.spacing = resolution
+            _validation.check_instance(volume, (pv.DataSet, pv.MultiBlock), name='volume')
         else:
             # HACK: Make a copy so the original object is not altered.
             #       Also, place all data on the nodes as issues arise when
             #       volume rendering on the cells.
             volume = volume.cell_data_to_point_data()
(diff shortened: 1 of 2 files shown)


pyvista/plotting/plotter.py
```diff
@@ -4462,7 +4462,7 @@ def add_volume(  # noqa: PLR0917
             #       Also, place all data on the nodes as issues arise when
             #       volume rendering on the cells.
             volume = volume.cell_data_to_point_data()
-        assert isinstance(volume, (pv.DataSet, pv.MultiBlock))
+        assert isinstance(volume, (pv.DataSet, pv.MultiBlock))  # noqa: S101
```
tkoyama010, line 4465: Let's raise an error here!

tkoyama010 (changes requested):

pyvista/plotting/plotter.py
user27182: this check is done a few lines earlier with `_validation.check_instance` on the `not is_pyvista_dataset(volume)` branch:
https://github.com/pyvista/pyvista/blob/8cfe9b69637ad140376768526d208f18d967cc7b/pyvista/plotting/plotter.py#L4450-L4459

So we already raise a TypeError here. But I think the issue is `mypy` doesn't realize we have already narrowed the type, so it thinks the `volume` variable can still be a non-mesh type. If we refactor a bit to make `mypy` happy, we can remove the need for this assertion. But `volume` is a heavily used variable so might not be so straightforward

tkoyama010: Make sense thanks!

tkoyama010 (approved):
LGTM

PR #8201, opened 2025-12-29
Validate data array lengths when wrapping vtk meshes

### LINKED ISSUE #8101: Incorrectly loaded `.cas.h5` file

I am trying to load a [`.cas.h5`](https://drive.google.com/file/d/1858RhMiTDfNpt4um3PEvscQ0HUUw9mib/view?usp=sharing) file, but I am getting `cell_data` fields with incorrect shapes. If this is a VTK issue I apologize and will raise the issue there. Thanks a lot for you help!

The first element of the `MultiBlock` is loaded correctly
```python
>>> import pyvista as pv
>>> result = pv.read("one_way_fsi_probe_report.cas.h5")
>>> result[0]
UnstructuredGrid (0x114fe4ac0)
  N Cells:    48478
  N Points:   83135
  X Bounds:   -7.500e-02, 1.250e-01
  Y Bounds:   -5.000e-02, 5.000e-02
  Z Bounds:   -4.999e-02, 6.661e-18
  N Arrays:   17

>>> result[0].cell_data
pyvista DataSetAttributes
Association     : CELL
Active Scalars  : None
Active Vectors  : None
Active Texture  : None
Active Normals  : None
Contains arrays :
    SV_DENSITY              float64    (48478,)
    SV_K                    float64    (48478,)
    SV_MU_LAM               float64    (48478,)
    SV_MU_T                 float64    (48478,)
    SV_O                    float64    (48478,)
    SV_P                    float64    (48478,)
    SV_SIGMA_XX             float64    (48478,)
    SV_SIGMA_XY             float64    (48478,)
    SV_SIGMA_XZ             float64    (48478,)
    SV_SIGMA_YY             float64    (48478,)
    SV_SIGMA_YZ             float64    (48478,)
    SV_SIGMA_ZZ             float64    (48478,)
    SV_U                    float64    (48478,)
    SV_V                    float64    (48478,)
    SV_W                    float64    (48478,)
    SV_WALL_DIST            float64    (48478,)
    SV_BF_V                 float64    (48478, 2)
```
while the second contains some incorrect shapes
```python
>>> result[1]
UnstructuredGrid (0x114fe4400)
  N Cells:    290985
  N Points:   83135
  X Bounds:   -7.500e-02, 1.250e-01
  Y Bounds:   -5.000e-02, 5.000e-02
  Z Bounds:   -4.999e-02, 6.661e-18
  N Arrays:   17

>>> result[1].cell_data
pyvista DataSetAttributes
Association     : CELL
Active Scalars  : None
Active Vectors  : None
Active Texture  : None
Active Normals  : None
Contains arrays :
    SV_DENSITY              float64    (290985,)
    SV_K                    float64    (242507,)
    SV_MU_LAM               float64    (242507,)
    SV_MU_T                 float64    (242507,)
    SV_O                    float64    (242507,)
    SV_P                    float64    (242507,)
    SV_SIGMA_XX             float64    (0,)
    SV_SIGMA_XY             float64    (0,)
    SV_SIGMA_XZ             float64    (0,)
    SV_SIGMA_YY             float64    (0,)
    SV_SIGMA_YZ             float64    (0,)
    SV_SIGMA_ZZ             float64    (0,)
    SV_U                    float64    (242507,)
    SV_V                    float64    (242507,)
    SV_W                    float64    (242507,)
    SV_WALL_DIST            float64    (290985,)
    SV_BF_V                 float64    (242507, 2)
```
There are 290,985 cells in the first block and 48,478 in the second. Subtracting the two results in the 242,507 seen above. My guess is that the reader assumes that the cell data is padded, which is not the case when looking at the H5 file via `h5py`:
- `results/1/phase-1/cells/SV_BF_V/1`:  `shape=(290985, 3)`
- `results/1/phase-1/cells/SV_DENSITY/1`:  `shape=(339463,)`
- `results/1/phase-1/cells/SV_K/1`:  `shape=(290985,)`
- `results/1/phase-1/cells/SV_MU_LAM/1`:  `shape=(290985,)`
- `results/1/phase-1/cells/SV_MU_T/1`:  `shape=(290985,)`
- `results/1/phase-1/cells/SV_O/1`:  `shape=(290985,)`
- `results/1/phase-1/cells/SV_P/1`:  `shape=(290985,)`
- `results/1/phase-1/cells/SV_SIGMA_XX/1`:  `shape=(48478,)`
- `results/1/phase-1/cells/SV_SIGMA_XY/1`:  `shape=(48478,)`
- `results/1/phase-1/cells/SV_SIGMA_XZ/1`:  `shape=(48478,)`
- `results/1/phase-1/cells/SV_SIGMA_YY/1`:  `shape=(48478,)`
- `results/1/phase-1/cells/SV_SIGMA_YZ/1`:  `shape=(48478,)`
- `results/1/phase-1/cells/SV_SIGMA_ZZ/1`:  `shape=(48478,)`
- `results/1/phase-1/cells/SV_U/1`:  `shape=(290985,)`
- `results/1/phase-1/cells/SV_V/1`:  `shape=(290985,)`
- `results/1/phase-1/cells/SV_W/1`:  `shape=(290985,)`
- `results/1/phase-1/cells/SV_WALL_DIST/1`:  `shape=(339463,)`

### System Information

```shell
  Date: Tue Nov 18 16:12:51 2025 UTC

                OS : Darwin (macOS 26.1)
            CPU(s) : 11
           Machine : arm64
      Architecture : 64bit
               RAM : 18.0 GiB
       Environment : IPython
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M3 Pro
       GPU Version : 4.1 Metal - 90.5
  MathText Support : True

  Python 3.12.7 (main, Oct 16 2024, 07:12:08) [Clang 18.1.8 ]

           pyvista : 0.46.3
               vtk : 9.3.1
             numpy : 2.2.6
        matplotlib : 3.10.6
            scooby : 0.10.2
             pooch : 1.8.2
            pillow : 11.3.0
           IPython : 9.5.0
             scipy : 1.16.2
              tqdm : 4.67.1
```

### Screenshots

_No response_

user27182:
### Overview

Resolve #8101. This PR emits a warning if the array lengths of a wrapped vtk mesh are invalid. This is similar to how an error is currently raised if trying to set a new array with invalid size. A warning is emitted instead of raising an error here though because this is mostly aimed at validating new meshes read from file, and I think it's easier to let users fix the issue using the PyVista API _after_ wrapping, whereas an error would leave users without a clear way to even read the VTK file or fix the problem.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License
 
-Copyright (c) 2017-2025 The PyVista Developers
+Copyright (c) 2017-2026 The PyVista Developers
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/doc/source/pyvista-binder-config.yml
+++ b/doc/source/pyvista-binder-config.yml
@@ -3,4 +3,4 @@ default_context:
   github_account: "pyvista"
   description: "Quick examples demonstrating what PyVista can do! Simply launch this Binder project and test drive PyVista"
   author_name: "PyVista Developers"
-  year: "2025"
+  year: "2026"


pyvista/core/filters/data_set.py
```diff
@@ -2399,6 +2408,9 @@ def _post_process_extract_values(before_extraction, extracted):
                 pass
             else:
                 return self
+        if 'pytest' in sys.modules:
+            # For CI, ensure buggy output arrays have been fixed
+            assert output.validate_mesh('data').is_valid  # noqa: S101
```
user27182: This is a bit unusual, I can move it into the test modules, or we can keep it but remove the branching based on pytest.

Since the warning is ignored earlier, we don't have coverage for the check here. So that's why I include the validation explicitly.

akaszynski: I'd prefer not having this here. Can we mock instead? I'm fine with leaving it here if there's no other option, but I'd prefer leaving testing related code out of runtime code if at all possible.

user27182: Yes I'll remove this from the run time code. We don't really _need_ to do these kinds of array checks for our filters, anyway since most behave well and return properly sized data arrays. I wonder if we should only restrict these checks to inside `pv.read`, rather than inside `pv.wrap`? The problem though is that this could miss legitimate use cases, e.g. we'd probably also need to add checks inside of `pv.get_reader(mesh).read()` and maybe other read functions. Whereas adding the check inside `pv.wrap` directly covers all cases.

akaszynski: I'd go with `wrap` or implement a separate validation that's shared by both.

user27182: I checked the `Reader.read` implementation, and it uses `wrap` internally post-read, and so adding the validation to `wrap` covers the `pv.read` and `get_reader.read` cases too.

This means the "init from file" cases are also covered, e.g `pv.PolyData(path)`, since this calls `_from_file` which itself calls `read`, which then uses `wrap`. And so I think applying validation inside of `wrap` is the most general, centralized location.

PR #8203, opened 2025-12-30
Add generic typing for mesh validation reports

user27182:
### Overview

Generic typing is added to the mesh validation reports so that the hinted type of the report's mesh matches the input type, i.e. `type(mesh) == type(mesh.validate_mesh().mesh)`


doc/source/api/core/typing.rst
```diff
@@ -117,50 +117,3 @@ Jupyter backend to use.
 .. currentmodule:: pyvista
 
 .. autodata:: JupyterBackendOptions
-
-
-Data Object Types
------------------
-
-pyvista._GridType
-~~~~~~~~~~~~~~~~~
```
user27182, line 126: The typing from this PR inherently has circular imports, since we have `DataSet` and `MultiBlock` references included inside of the `DataObjectFilters` module. This is not an issue from a typing perspective, because the imports are guarded behind `if TYPE_CHECKING` blocks and stringified, and hence there are only circular imports if trying to define these types at run time.

The issue is that Sphinx needs to resolve these classes at runtime in order to document them. So, we can't simultaneous use these classes in a forward-reference manner for actual typing, _and_ document these at run time. In order to enable the typing use cases, these classes are removed from the docs.

These are internal typing classes only, so I don't think they're actually being used by third party code anyway and should be safe to remove from the docs.

akaszynski: > These are internal typing classes only, so I don't think they're actually being used by third party code anyway and should be safe to remove from the docs.

Internal and shouldn't be used. There's no implicit expectation that anything with a leading `_` will be consistent.

akaszynski (approved):

Issue #7076, opened 2025-01-08
Array is unexpectedly broadcasted instead of raising an error


user27182:
### Describe the bug, what's wrong, and what you expected.

Setting an array with size one is unexpectedly broadcasted:
``` python
import pyvista as pv
import numpy as np
mesh = pv.examples.load_uniform()
array = np.ones(())
print(array.size) # size is 1 
mesh.cell_data['data'] = array
print(mesh['data'].size) # size is 729
```
This is only when setting `point_data` or `cell_data` explicitly. If we instead do:
``` python
mesh['data'] = array
```
We get the expected error:
```ValueError: Number of scalars (1) must match either the number of points (1000) or the number of cells (729).```

### Steps to reproduce the bug.

See above.

### System Information

```shell
OS : Darwin (macOS 14.4.1)
            CPU(s) : 8
           Machine : arm64
      Architecture : 64bit
               RAM : 16.0 GiB
       Environment : IPython
       File system : apfs
        GPU Vendor : Apple
      GPU Renderer : Apple M2
       GPU Version : 4.1 Metal - 88
  MathText Support : True
  Python 3.12.2 (v3.12.2:6abddd9f6a, Feb  6 2024, 17:02:06) [Clang 13.0.0
  (clang-1300.0.29.30)]
           pyvista : 0.45.dev0
               vtk : 9.3.1
             numpy : 2.1.0
        matplotlib : 3.9.3
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 11.0.0
           imageio : 2.36.1
           IPython : 8.30.0
          colorcet : 3.1.0
           cmocean : 4.0.3
        ipywidgets : 8.1.5
             scipy : 1.14.1
              tqdm : 4.67.1
            meshio : 5.3.5
        jupyterlab : 4.3.1
    pytest_pyvista : 0.1.8
             trame : 3.7.0
      trame_client : 3.5.0
      trame_server : 3.2.3
         trame_vtk : 2.8.11
     trame_vuetify : 2.7.1
      nest_asyncio : 1.6.0
```

user27182:
It is easy/convenient, but is it documented and tested? In your example it's easy to see that a scalar value is assigned to an array container, so as a reader _maybe_ it's safe to assume the value is broadcasted here?

[...]

PR #8205, opened 2025-12-31
Constant array data typing

edabor:
### Overview

This PR fixes a typing issue encountered when setting a constant value as `point_data` or `cell_data`.

image

### DIFF
--- a/doc/source/user-guide/data_model.rst
+++ b/doc/source/user-guide/data_model.rst
@@ -528,6 +528,14 @@ and the name of the active scalars array can be accessed or set with
    >>> ugrid.cell_data.active_scalars_name = 'my-cell-data'
    >>> ugrid.cell_data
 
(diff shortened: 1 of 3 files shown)


user27182 (changes requested):
Thanks for documenting this, but I am -1 on this change.

This PR effectively resolves #7076 by formally documenting this "feature" and expanding the type hints accordingly to allow scalars. I think this kind of magical array broadcasting is more of a bug than a feature. Instead, I think we should keep the current type hints and raise a `ValueError` for scalars instead (or maybe even `TypeError`).

Perhaps the only case where assigning a scalar should be allowed to work is when `n_points == 1`, but even then we should probably require users to do `(value,)` and pad out the dimensions explicitly. I think users should use `np.broadcast`, `np.zeros`, `np.full`, etc. explicitly if they want this behavior.

PR #8206, opened 2026-01-01
Raise `ValueError` when accessing `obbTree` with empty `PolyData`

### LINKED ISSUE #8204: Errors reported by vtkOBBTree when accessing obbTree from empty pyvista.PolyData
### Problem

Consider the following code example:

```python
import pyvista as pv

pd = pv.PolyData()
obbTree = pd.obbTree
```

Running this code with PyVista 0.46.4 results in a log message from VTK containing:

`vtkOBBTree.cxx:1078   ERR| vtkOBBTree (0x600000cd00e0): Can't build OBB tree - no data available!`

This is caused by the unconditional building of the OBB tree on line 1684https://github.com/pyvista/pyvista/blob/5325d9167c355d89e313ed7999c0cc09322a692f/pyvista/core/pointset.py#L1683-L1685

For better or worse, VTK reports an error condition when it is asked to build the OBB tree on empty datal: https://gitlab.kitware.com/vtk/vtk/-/blob/master/Filters/General/vtkOBBTree.cxx#L1059

I expect no error messages from VTK when running this code.

### Suggested fix

PyVista should check that the `PolyData` has points and cells prior to calling `obb_tree.BuildLocator()`, e.g.

```python
obb_tree.SetDataSet(self)
if (self.n_points > 0 and self.n_cells > 0):
  obb_tree.BuildLocator()
```

```python
import pyvista as pv

pd = pv.PolyData()
obbTree = pd.obbTree
```

### System Information

```shell
  Date: Tue Dec 30 16:34:22 2025 UTC

                OS : Darwin (macOS 15.7.2)
            CPU(s) : 10
           Machine : arm64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : Apple
      GPU Renderer : Apple M4
       GPU Version : 4.1 Metal - 89.4
  MathText Support : True

  Python 3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 16.0.0
  (clang-1600.0.26.6)]

           pyvista : 0.46.4
               vtk : 9.5.2
             numpy : 2.4.0
        matplotlib : 3.10.8
            scooby : 0.11.0
             pooch : 1.8.2
            pillow : 12.0.0
```

### Screenshots

_No response_

user27182:
### Overview

Fix #8204

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -373,6 +373,8 @@ class _MeshValidator(Generic[_DataSetOrMultiBlockType]):
 
     @staticmethod
     def _invalid_cell_msg(name: str, array: list[int]) -> str:
+        if not array:
+            return ''
         name_norm = _MeshValidator._normalize_field_name(name)
         # Need to write name either before of after the word "cell"
         if name == 'non_convex':
@@ -401,6 +403,8 @@ class _MeshValidator(Generic[_DataSetOrMultiBlockType]):
         def _invalid_array_length_msg(
             invalid_arrays: dict[str, int], kind: str, expected: int
         ) -> str:
+            if not invalid_arrays:
+                return ''
             n_arrays = len(invalid_arrays)
             s = 's' if n_arrays > 1 else ''
             msg_template = (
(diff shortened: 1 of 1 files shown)


akaszynski (approved):

pyvista/core/pointset.py
```diff
@@ -1736,7 +1739,8 @@ def __del__(self) -> None:
         """Delete the object."""
         # avoid a reference cycle that can't be resolved with vtkPolyData
         self._glyph_geom = None
-        self.obbTree = None  # type: ignore[assignment]
+        with contextlib.suppress(ValueError):
+            self.obbTree = None  # type: ignore[assignment]
```
user27182: Pretty sure part of the issue is that accessing `self.obbTree` calls the getter and hence always computes the obbTree on deletion if not cached already. This is why the VTK error was printing when only `del` is called here https://github.com/ansys/pyansys-heart/issues/1286. This is why we suppress the ValueError currently. But we can avoid calling the getter if we delete it from the object's dict directly:
``` suggestion
        with contextlib.suppress(KeyError):
            del self.__dict__['obbTree']
```

PR #8212, opened 2026-01-01
Do not format messages for valid mesh validation fields

user27182:
### Overview

When validating meshes, we only need to format error messages if there is an error to report. This PR skips message formatting when the field is valid.


pyvista/core/filters/data_object.py
```diff
@@ -328,7 +328,7 @@ def _validate_cells(
             validated_mesh = mesh.cell_validator()
             for name in mutable_validation_fields:
                 array = validated_mesh.field_data[name].tolist()
-                msg = _MeshValidator._invalid_cell_msg(name, array)
+                msg = '' if not array else _MeshValidator._invalid_cell_msg(name, array)
```
tkoyama010: Applying the fix to `_invalid_cell_msg` within the method would be a better approach to handle the other lines consistently.

pyvista/core/filters/data_object.py
```diff
@@ -422,8 +422,12 @@ def _validate_array_lengths(arrays: DataSetAttributes, expected: int) -> dict[st
...
             if name in validation_fields:
                 invalid_arrays: dict[str, int] = _validate_array_lengths(data, expected_n)
-                message = _invalid_array_length_msg(
-                    invalid_arrays=invalid_arrays, kind=kind, expected=expected_n
+                message = (
+                    ''
+                    if not invalid_arrays
+                    else _invalid_array_length_msg(
+                        invalid_arrays=invalid_arrays, kind=kind, expected=expected_n
+                    )
```
tkoyama010: Same with https://github.com/pyvista/pyvista/pull/8212/changes#r2656700310 .

user27182: Thanks, should be fixed with [2ceb885](https://github.com/pyvista/pyvista/pull/8212/commits/2ceb885fa6e85c51526a7ddf7f6c056b095b6671)

tkoyama010 (approved):
LGTM

PR #8213, opened 2026-01-01
Allow deleting all data arrays

edabor:
### Overview

This PR allows deleting all arrays from dataset attributes (point, cell and field data), as follow:

```python
del mesh.point_data[:]
del mesh.cell_data[:]
del mesh.field_data[:]
```

### DIFF
--- a/pyvista/core/datasetattributes.py
+++ b/pyvista/core/datasetattributes.py
@@ -1030,6 +1030,14 @@ class DataSetAttributes(
         >>> len(mesh.point_data)
         0
 
+        Alternatively, the ``del`` keyword can be used.
+        >>> mesh.point_data['my_data'] = range(mesh.n_points)
+        >>> len(mesh.point_data)
+        1
+        >>> del mesh.point_data[:]
+        >>> len(mesh.point_data)
+        0
+
         """
         for array_name in self.keys():
             self.remove(key=array_name)
--- a/pyvista/core/dataobject.py
+++ b/pyvista/core/dataobject.py
@@ -521,6 +521,15 @@ class DataObject(_NoNewAttrMixin, _vtk.DisableVtkSnakeCase, _vtk.vtkPyVistaOverr
         >>> len(mesh.field_data)
         0
 
+        Alternatively, the ``del`` keyword can be used.
+        >>> mesh = pv.Sphere()
+        >>> mesh.field_data['my-field-data'] = range(10)
+        >>> len(mesh.field_data)
+        1
+        >>> del mesh.field_data[:]
+        >>> len(mesh.field_data)
+        0
+
         """
         if not hasattr(self, 'field_data'):
             msg = f'`{type(self)}` does not support field data'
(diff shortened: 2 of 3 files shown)


pyvista/core/datasetattributes.py
```diff
@@ -276,13 +289,25 @@ def __setitem__(
...
+        if isinstance(key, str):
+            self.remove(key)
+            return
+
+        if key != slice(None, None, None):
+            msg = 'Slicing deletion is only allowed with all values (ie. with [:]).'
+            raise ValueError(msg)
+
+        for k in self.keys():
+            self.remove(key=k)
```
user27182:
```suggestion
        self.clear()
```

edabor:
> Can we add better documentation or cross references to similar deletion methods? E.g. add an example to `DataSetAttributes.clear()` to show this new alternative.

Done in https://github.com/pyvista/pyvista/pull/8213/commits/37a1303a81aef9729e8df7059199c06c8ba585fc

akaszynski:
> Seems a bit redundant since we can already do `mesh.clear_point_data()` or `mesh.point_data.clear()` to do the same thing, though allowing a third method with `del mesh.point_data[:]` does seem to be consistent with `del mesh.point_data['array']`.
> 
> Can we add better documentation or cross references to similar deletion methods? E.g. add an example to `DataSetAttributes.clear()` to show this new alternative.
> 
> Or maybe there should be a separate doc PR for this? I am thinking that maybe it's just a matter of letting users know these other methods exist.

Looking at this I think we should really implement this if there's a genuine need for it. Plus, I thought we were trying to make our `DataSetAttributes` "dict-like", which doesn't intuitively support slices:

```py
$ python
Python 3.12.11 (main, Jun  3 2025, 15:41:47) [GCC 14.2.1 20250322] on linux
>>> mydict = {'a': 1, 'b': 2}
>>> del mydict[:]
KeyError: slice(None, None, None)
```

You can add a `slice` as a key:

```py
>>> mydict[slice(None, None,None)] = 'apparently python supports slices as keys'
>>> mydict[:]
'apparently python supports slices as keys'
```

However, clearly a dict doesn't support a `slice` "operation", only as a key.

user27182:
I can see that this is perhaps a bit unintuitive, since it appears that we are asking to delete the slice key. But only strings are allowed as data keys, so there is no actual slice key, and instead deleting this key will remove _all_ keys.

If someone sees this code without intimate knowledge about the API, I'm not sure if it's obvious that this removes all keys. But `data.clear()` is much more explicit.

edabor:
> I can see that this is perhaps a bit unintuitive, since it appears that we are asking to delete the slice key. But only strings are allowed as data keys, so there is no actual slice key, and instead deleting this key will remove _all_ keys.
> 
> If someone sees this code without intimate knowledge about the API, I'm not sure if it's obvious that this removes all keys. But `data.clear()` is much more explicit.

I agree, this PR might not be needed after all. Let's stick to the zen `There should be one-- and preferably only one --obvious way to do it.`

PR #8215, opened 2026-01-02
Expand docs for PolyData verts, lines, faces, strips

user27182:
### Overview

Add clarity, consistency, and cross-references to the four PolyData connectivity array docs: `verts`, `lines`, `faces`, and `strips`


akaszynski:
Thanks for adding to the docs. The difference between vertices and points is a bit confusing in our current documentation; this clears it up.

Issue #7742, opened 2025-07-22
Sources like SphereSource need double precision


banesullivan:
### Describe the bug, what's wrong, and what you expected.

We really need to be using douple precision when generating points for basic geometries like `pv.Sphere()`

I regularly work in coordinate systems (e.g. UTM) where I loose the necessary precision under float32 with common XY values being `(500_000.01,  5_000_000.01)`

We ought to set all of our sources to use double precision by defualt by setting `vtkAlgorithm.SetOutputPointsPrecision(vtkAlgorithm.DOUBLE_PRECISION)`

### Steps to reproduce the bug.

Current behavior:

```python
>>> import pyvista as pv
>>> a = pv.Sphere(center=(500_000.01,  5_000_000.01, 0.0), radius=3)
>>> a.points.dtype
dtype('float32')
>>> a.plot()
```

<img width="1330" height="600" alt="Image" src="https://github.com/user-attachments/assets/db71d3c5-3e95-4d15-96a1-de622eddcbd1" />

We can set the points to use double precision to avoid this

```py
algo = pv.SphereSource(center=(500_000.01,  5_000_000.01, 0.0), radius=3)
algo.SetOutputPointsPrecision(algo.DOUBLE_PRECISION)
algo.output.plot()
```

<img width="1330" height="600" alt="Image" src="https://github.com/user-attachments/assets/978a174a-51cd-4fe3-9a01-4a2277535107" />

### System Information

```shell
latest pyvista
```

tkoyama010:
relate to https://github.com/pyvista/pyvista/pull/6397

PR #8216, opened 2026-01-02
Add optional `validate` keyword to PolyData and UnstructuredGrid constructors

### LINKED ISSUE #2124: Crash when `compute_normals`

When calculating normals, the interpreter silently crashes.. No error's or output, not even with the `VTKErrorCatcher`. can anyone please enlighten me what happens in this minimal example?

-----

**To Reproduce**

Please include a code snippet to reproduce the bug in the block below:

```py
import pyvista
import numpy as np

xyz = np.random.random(102*3).reshape(-1,3)
faces = np.arange(102*3)
np.random.shuffle(faces)
faces = faces.reshape(-1,3)
faces = np.column_stack((np.ones(len(faces),) * 3, faces)).astype(int)

mesh = pyvista.PolyData()
mesh.points = xyz
mesh.faces = faces
print(mesh.compute_normals(inplace=False))
```

------
**System Information:**

```py
# Get system info
import pyvista as pv
print(pv.Report())
```

```txt
  Date: Wed Feb 02 16:36:41 2022 W. Europe Standard Time

                OS : Windows
            CPU(s) : 8
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.9 GiB
       Environment : Python
       File system : unknown
      GPU Renderer : Quadro M1000M/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 441.66

  Python 3.9.0 (tags/v3.9.0:9cf6752, Oct  5 2020, 15:34:40) [MSC v.1927 64 bit
  (AMD64)]

           pyvista : 0.33.2
               vtk : 9.1.0
             numpy : 1.22.1
           imageio : 2.14.1
           appdirs : 1.4.4
            scooby : 0.5.11
        matplotlib : 3.5.1
         pyvistaqt : 0.7.0
             PyQt5 : 5.15.6
             scipy : 1.7.3


```

user27182:
### Overview

Fix #2124 by adding optional validation during init.

The check for `invalid_point_references` from the `validate_mesh` method specifically is what will raise an error when invalid point ids are given, such as those in #2124.

Initially tried adding a check by default, similar to https://github.com/pyvista/pyvista/pull/5404, but there was a performance cost for computing max and min of the connectivity array which does not seem reasonable to incur for the default case.

### DIFF
--- a/tests/core/test_dataobject_filters.py
+++ b/tests/core/test_dataobject_filters.py
@@ -2125,10 +2125,13 @@ def test_init_invalid_mesh(invalid_random_polydata, tmp_path, as_grid, validate)
         vtk_mesh = alg.GetOutput()
         mesh = pv.UnstructuredGrid()
         mesh.ShallowCopy(vtk_mesh)
+        array_args = mesh.cells, mesh.celltypes, mesh.points
     else:
         mesh = invalid_random_polydata
         vtk_mesh = _vtk.vtkPolyData()
         vtk_mesh.ShallowCopy(mesh)
+        array_args = mesh.points, mesh.faces
+    mesh_type = type(mesh)
 
     filepath = tmp_path / 'invalid.vtk'
     mesh.save(filepath)
@@ -2137,12 +2140,16 @@ def test_init_invalid_mesh(invalid_random_polydata, tmp_path, as_grid, validate)
 
     # Init from file
     with pytest.raises(pv.InvalidMeshError, match=match):
-        mesh.__class__(filepath, validate=validate)
+        mesh_type(filepath, validate=validate)
 
     # Init from unwrapped VTK mesh
     with pytest.raises(pv.InvalidMeshError, match=match):
-        mesh.__class__(vtk_mesh, validate=validate)
+        mesh_type(vtk_mesh, validate=validate)
 
     # Init from PyVista mesh
     with pytest.raises(pv.InvalidMeshError, match=match):
-        mesh.__class__(mesh, validate=validate)
+        mesh_type(mesh, validate=validate)
+
+    # Init from arrays
+    with pytest.raises(pv.InvalidMeshError, match=match):
+        mesh_type(*array_args, validate=validate)


pyvista/core/pointset.py
```diff
@@ -815,6 +817,8 @@ def __init__(  # noqa: PLR0917
                 self.deep_copy(var_inp)
             else:
                 self.shallow_copy(var_inp)  # type: ignore[arg-type]
+            # Validate connectivity
+            self._raise_invalid_point_references(as_warning=True)
```
user27182: The `vtkPolyData` -> `PolyData` case is very common internally when wrapping alg outputs. I don't think we necessarily need to run this check all the time for these cases. The same is probably true to some extent for the array validation added in #8201.

I'll make a follow-up PR to add a `validate` keyword to `DataSet.__init__`, and probably `read` and `wrap` too. This would effectively enable minimal "memory_safe" mesh validation by default, but allow users to toggle it off, or even make it more strict. We could disable it by default for internal alg wraps, but leave it on for external calls to wrap from users.

pyvista/core/pointset.py
```diff
@@ -872,6 +876,55 @@ def _post_file_load_processing(self) -> None:
...
+
+    def _check_invalid_point_references(
+        self, attr: Literal['verts', 'lines', 'faces', 'strips']
+    ) -> str | None:
+        """Verify that a connectivity array does not reference invalid points."""
+        vtkattr = {'verts': 'Verts', 'lines': 'Lines', 'faces': 'Polys', 'strips': 'Strips'}[attr]
+        if getattr(self, f'GetNumberOf{vtkattr}')():
+            n_points = self.n_points
+            connectivity = _get_connectivity_array(getattr(self, f'Get{vtkattr}')())
+            if connectivity.size and connectivity.max() >= n_points:
```
user27182: Should probably also check the values are non-negative, which would match the implementation inside `validate_mesh`

user27182:
I'm going to generalize this check to work with UnstructuredGrid too

tests/core/test_polydata.py
```diff
@@ -229,6 +231,85 @@ def test_invalid_connectivity_arrays(arr: str, value: list | np.ndarray, expecte
...
+        'The connectivity of `UnstructuredGrid.cells` includes references to\n'
+        'point ids that do not exist. The point ids must be non-negative and strictly less '
+        'than the\nnumber of points (1).'
+    )
+    with pytest.raises(pv.InvalidMeshError, match=re.escape(match)):
+        pv.UnstructuredGrid(cells, celltypes, points)
+
+
+@pytest.mark.parametrize('as_grid', [True, False])
+def test_init_invalid_connectivity_warns(invalid_random_polydata, tmp_path, as_grid):  # noqa: F811
```
user27182: There are three tests: one tests only `PolyData`, another tests only `UnstructuredGrid`, and a third tests both. But this is all inside `test_polydata`. We could leave this here, (there are similar examples of this kind of misfit of tests elsewhere), or maybe move it all to `test_dataobject.py`? The errors are tightly coupled and we should expect similar behavior for both types, so I think we should keep them all together.

pyvista/core/pointset.py
user27182: This could also be premature optimization. Checking the length of tens of arrays should not have a high cost. And the connectivity array traversal for computing min, max might have an impact but it's not clear how much unless we do some profiling.

user27182: There is a noticeable performance impact. Not sure what the best way to measure it is, but this code wraps a filter's output on a large mesh with 1 milliion points, and the execution time approximately doubles when compared to `main`

```py
import time
import pyvista as pv
import vtk

mesh = pv.examples.download_antarctica_velocity()

def construct():
    vtkappend = vtk.vtkAppendFilter()
    vtkappend.AddInputData(mesh)
    vtkappend.Update()
    pv.UnstructuredGrid(vtkappend.GetOutput())

N = 1000
start = time.perf_counter()
for _ in range(N):
    construct()
elapsed = time.perf_counter() - start

print(f"Total time: {elapsed:.6f} s")
```

And so I've disabled running the check when wrapping existing meshes. The check now only runs when using arrays explicitly to create the mesh. This change was done in https://github.com/pyvista/pyvista/pull/8216/commits/d8a85286246c4ea664e8aa1f6496c064ab9b19bb

We should still allow users to enable/disable this check optionally though, maybe guarded behind a `validate` keyword implemented as a separate PR.

akaszynski: > We should still allow users to enable/disable this check optionally though, maybe guarded behind a `validate` keyword implemented as a separate PR.

Agreed, and implement it when wrapping or during initialization.

tests/core/test_polydata.py
akaszynski: Recommending moving them to `test_dataobject.py`.

pyvista/core/pointset.py
akaszynski: I'm seeing the same slowdown. Simple test:

```py
import time
import pyvista as pv


def make_cube(check):
    cube = pv.ImageData(dimensions=(50, 50, 50)).cast_to_unstructured_grid()
    if check:
        rng = cube.GetCells().GetConnectivityArray().GetRange()


num = 1000

for check in (False, True):
    start = time.perf_counter()
    for _ in range(num):
        make_cube(check)
    elapsed = time.perf_counter() - start
    print(f"Check {str(check):7s} time: {elapsed:.6f} seconds")
```

```
Check False   time: 2.635287 seconds
Check True    time: 3.008780 seconds
```

However, while investigating this it turns out that `numpy` is faster in this case:

```py
import time
import pyvista as pv


def make_cube(check):
    cube = pv.ImageData(dimensions=(50, 50, 50)).cast_to_unstructured_grid()
    if check:
        cells = cube.cell_connectivity
        rng = cells.min(), cells.max()


num = 1000

for check in (False, True):
    start = time.perf_counter()
    for _ in range(num):
        make_cube(check)
    elapsed = time.perf_counter() - start
    print(f"Check {str(check):7s} time: {elapsed:.6f} seconds")

```


```
Check False   time: 2.645587 seconds
Check True    time: 2.895733 seconds
```

If we wanted to go crazy, there's also https://github.com/nomonosound/numpy-minmax, but I'm loath to add additional dependencies.

pyvista/core/filters/data_object.py
```diff
@@ -510,6 +521,29 @@ def _create_message_header(obj: object) -> str:
...
+        mesh: PolyData | UnstructuredGrid,
+        cell_array: _vtk.vtkCellArray,
+        attr: Literal['verts', 'lines', 'faces', 'strips', 'cells'],
+        *,
+        action: _MeshValidator._ActionOptions,
+    ) -> None:
+        """Raise error if the connectivity array has invalid point references."""
+        if cell_array.GetNumberOfCells() > 0:
+            n_points = mesh.n_points
+            conn_min, conn_max = cell_array.GetConnectivityArray().GetRange()
```
akaszynski: Consider optimizing using `numpy`. See https://github.com/pyvista/pyvista/pull/8216#discussion_r2662993036

pyvista/core/pointset.py
user27182: Okay, thanks for confirming the performance hit. Indeed this should be disabled by default then. 

The PR in its current form was an attempt at adding this check with as small of footprint as possible. This meant avoiding using `validate_mesh` directly and avoiding casting to unstructured grid. I also falsely assumed that VTK's GetRange() would be faster than numpy min/max (one operation vs two). 


But if this will be opt-in, then we should just use a single call to `valdiate_mesh(action='error')` and greatly simplify the implementation. It already uses the faster numpy max/min for the check.

user27182: > The PR in its current form was an attempt at adding this check with as small of footprint as possible. This meant avoiding using `validate_mesh` directly and avoiding casting to unstructured grid.

Another reason for having a distinct implementation for PolyData was so that it could be cell-array specific and check verts, lines, faces, and strips separately. Otherwise, we lose this distinction after casting. The current "CellSizeError" makes this distinction, for example.

Maybe we can customize the MeshValidator messages for PolyData, so that cell errors are reported separately for each array? But this also has issues because we print the cell ids... should users expect global cell ids like the ones extracted by `extract_cells`? Or do users expect fine-grained ids for each PolyData connectivity array?

I'm thinking the easiest solution is to just do the casting for now, and report all cell ids as global ids. That way, we avoid all the complexity of customizing the implementation and reporting.

user27182:
Based on this thread: https://github.com/pyvista/pyvista/pull/8216#discussion_r2658880898, I have reworked the PR to optionally enable using `validate_mesh` directly when initializing PolyData or UnstructuredGrid. All review comments should now be resolved.

pyvista/core/pointset.py
akaszynski: > I also falsely assumed that VTK's GetRange() would be faster than numpy min/max (one operation vs two).

I'd assume that too. It should be faster to compute the min/max in a single pass than calling `numpy.min` and `numpy.max`.

> should users expect global cell ids like the ones extracted by `extract_cells`? 

I think so. Implement this as cheaply as possible and if/when users have issues about global IDs we can resolve that (or think about the trade offs again).

user27182:
> LGTM. Let's follow up with setting a global theme for those who want to enable/disable always.

Yes, and probably add the keyword to the other mesh inits, along with `wrap`. 

We can add it for now to `global_theme`, but I think we should keep theme for global _plotting_ configuration, and we should consider adding a new `pv.global_config` attribute to consolidate _core_ based global variable like `vtk_verbosity`, `vtk_snake_case`, `pickle_format`, `allow_new_attributes`, etc. or maybe we just put everything inside of global theme?

Edit: this could be extended to included things like `force_float` for casting points to floats, `points_dtype` if we ever add support for 64bit points with meshes, etc.

akaszynski:
> We can add it for now to `global_theme`, but I think we should keep theme for global _plotting_ configuration, and we should consider adding a new `pv.global_config` attribute to consolidate _core_ based global variable like `vtk_verbosity`, `vtk_snake_case`, `pickle_format`, `allow_new_attributes`, etc. or maybe we just put everything inside of global theme?

Good point. I'm tempted to keep this out of the `global_theme` for the moment and simply implement this as-is and have a follow-up to add it to `wrap` and `pv.read`. We can add `global_config` for the next minor release.

akaszynski:
> Edit: this could be extended to included things like `force_float` for casting points to floats, `points_dtype` if we ever add support for 64bit points with meshes, etc.

That would be great for resolving https://github.com/pyvista/pyvista/issues/7742.

Issue #8130, opened 2025-11-28
Revert `python_version < "3.14"` before next release


user27182:
> ``` py
>  'vtk!=9.4.0; python_version < "3.14"',
>  'vtk!=9.4.1; python_version < "3.14"',
>  'vtk<9.6.0; python_version < "3.14"',
>  'vtk>=9.2.2; python_version < "3.14"',
> ```
Is this change going to be in the next release ? If so, that would mean that `pyvista` could be installed on `python 3.14` without `vtk`

_Originally posted by @edabor in https://github.com/pyvista/pyvista/pull/8096#discussion_r2572133817_

PR #8222, opened 2026-01-07
Add validate keyword to mesh constructors

user27182:
### Overview

Follow-up to #8216

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -483,11 +483,12 @@ class _MeshValidator(Generic[_DataSetOrMultiBlockType]):
             )
 
         summaries: list[_MeshValidator._FieldSummary] = []
-        for name, point_ids, info in [
-            ('unused_points', get_unused_point_ids(), ' not referenced by any cell(s)'),
-            ('non_finite_points', get_non_finite_point_ids(), ''),
+        for name, func, info in [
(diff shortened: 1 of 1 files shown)


akaszynski:
Aside from this PR, which others are needed before the next minor release?

user27182:
> Aside from this PR, which others are needed before the next minor release?

#8130 should also be fixed before release

user27182:
There's a lot of complexity trying to get this to work with `pv.wrap` and `pv.read`. The code is very fragmented with various functions like `read_meshio`, `read_trimesh`, `read_grdecl`, etc., along with a  `_post_file_load_processing` method, plus all the different VTK readers with `BaseReader.read`. We may want to add individual `validate` keywords to these too. But it's kind of a mess, and it's really unclear exactly where or what should actually make the final, single call to `validate_mesh`.

Might be easier to manage this with a global config variable later, and only support validating with mesh constructors explicitly for now.

akaszynski:
> Might be easier to manage this with a global config variable later, and only support validating with mesh constructors explicitly for now.

Works for me. Let's not change too much now if it requires a substantial change.

user27182:
Looks like this test is flaky on Windows with xdist:
https://github.com/pyvista/pyvista/actions/runs/20860570074/job/59938871583#step:7:19059
Will need to follow-up to skip for windows

PR #8227, opened 2026-01-08
Include cell type info with mesh validation report

user27182:

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -413,7 +413,7 @@ class _MeshValidator(Generic[_DataSetOrMultiBlockType]):
         name: str,
         array: list[int] | tuple[list[int]],
         cell_type: CellType | list[CellType] | None = None,
-    ) -> str | list[str]:
+    ) -> _NestedStrings:
         if not array:
             return ''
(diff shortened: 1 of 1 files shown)


tests/core/test_dataobject_filters.py
```diff
@@ -2113,6 +2134,38 @@ def _format_composite(match):
...
+        'PolyData mesh is not valid due to the following problems:\n'
+        ' - Mesh has 1 TRIANGLE cell with invalid point references. Invalid cell id: [0]'
+    )
+    assert message == expected
+
+    message = mixed_2d_cells_invalid_point_references.validate_mesh().message
+    expected = (
+        'PolyData mesh is not valid due to the following problems:\n'
+        ' - Mesh has 1 TRIANGLE cell with invalid point references. Invalid cell id: [0]\n'
+        ' - Mesh has 1 QUAD cell with invalid point references. Invalid cell id: [1]'
```
user27182, line 2156: One minor downside to this PR is that although it makes the error message more explicit, it's now inconsistent with the report string/summary. E.g. the report summary here would just say 
"Invalid point references: [0, 1]"

And so the message gives you a bit more info that the report itself doesn't include, and users can't easily access the separate TRIANGLE and / or QUAD ids.

PR #8228, opened 2026-01-08
Support VTK 9.6

### LINKED ISSUE #8130: Revert `python_version < "3.14"` before next release

user27182:
### Overview

Resolve #8130. This should allow us to publish another minor release. 
It's currently possible to install PyVista _without_ vtk on Python 3.14. This was done for `main` only so that we could install the latest vtk dev wheels with Python 3.14, but we won't need this anymore once the vtk 9.6 release candidate is released.

Update: vtk 9.6.0rc2 wheels are now on PyPI, so this PR tests against that version of vtk now


akaszynski:
> ... once the vtk 9.6 release candidate is released.

Let's wait until the release candidate is out to merge this. Tracking here:
https://discourse.vtk.org/t/vtk-9-6-0-release-planning/15760

I'm fine with waiting for our minor release if it's done by late January. If not, we can generate a patch release.

user27182:
Just to clarify, this change affects dev only, and effectively reverts #8096. The 0.46 release already cannot be installed without vtk, so we don't need any patch releases.

user27182:
Draft until the VTK 9.6 rc is released on PyPI. Our CI vtk dev testing will break if we merge this PR before then.

user27182:
There is a docs build failure with `resample`:
https://github.com/pyvista/pyvista/actions/runs/21107859731/job/60701962469?pr=8228#step:7:539

This is a VTK regression: https://gitlab.kitware.com/vtk/vtk/-/issues/19909

examples/99-advanced/mesh_validation.py
```diff
@@ -87,7 +87,7 @@
 
 # %%
 # If we review the invalid fields, we see that `two` are reported instead of only one.
```
akaszynski: We're going to have to correct this text since it's no longer "two".

akaszynski:
Looks like there are several regressions. One of which is the lack of axis labels in `show_grid`:

`dataobjectfilters.scale` - `vtk<9.6`:
https://docs.pyvista.org/api/core/_autosummary/pyvista.dataobjectfilters.scale#pyvista.DataObjectFilters.scale
image

`dataobjectfilters.scale` - `vtk==9.6.0rc2`:
https://696d27960e305a9ec5a3a162--pyvista-dev.netlify.app/api/core/_autosummary/pyvista.dataobjectfilters.scale
image

examples/99-advanced/mesh_validation.py
user27182: Fixed in fd59399a179c2ed01569fdc94156ec4d82503bbe

user27182:
> Looks like there are several regressions. One of which is the lack of axis labels in `show_grid`:

This is a known VTK regression: https://gitlab.kitware.com/vtk/vtk/-/issues/19729. It doesn't seem likely that this will be fixed upstream. We can try to fix this and use the 2D labels instead? I.e. set `use_3d_text=False` for vtk>=9.6. This would technically be a breaking change, but it would actually be a fix to address the VTK issue.

Looks like `slice_orthogonal` is also very broken:

image

akaszynski:
Confirmed that `use_3d_text=False` "fixes" the regression for vtk=9.6. Maybe set the default to `None` and note that the default is `False` for `vtk>=9.6`?

akaszynski:
Looks like the issue is with `slice`:
```py
import pyvista as pv
from pyvista import examples

mesh = examples.download_embryo()

slc = mesh.slice(normal='z')
slc.plot()
```


sc003

akaszynski:
Uncovered the issue with `slice_orthoginal`. `spacing` from the embryo example is incorrectly loaded as `(1.0, 1.0, 0.0)` with `vtk==9.6.0rc2`

```py
>>> mesh = examples.download_embryo()
>>> mesh.spacing
(1.0, 1.0, 0.0)
```

Simply setting `spacing` to `(1.0, 1.0, 1.0)` fixes it.
```py
import pyvista as pv
from pyvista import examples

mesh = examples.download_embryo()
mesh.spacing = (1.0, 1.0, 1.0)
mesh.slice(normal="z").plot()
```
sc004

user27182:
> Uncovered the issue with `slice_orthoginal`. `spacing` from the embryo example is incorrectly loaded as `(1.0, 1.0, 0.0)` with `vtk==9.6.0rc2`
> 
> ```python
> >>> from pyvista import examples
> >>> mesh = examples.download_embryo()
> >>> mesh.spacing
> (1.0, 1.0, 0.0)
> ```
> 
> Simply setting `spacing` to `(1.0, 1.0, 1.0)` fixes it.


Maybe this is the same issue with `sample` then from https://github.com/pyvista/pyvista/pull/8228#issuecomment-3765085991? This breakage _also_ uses `embryo`. I'm thinking maybe the root cause of both issues stems from `SLCReader` instead?

user27182:
> Confirmed that `use_3d_text=False` "fixes" the regression for vtk=9.6. Maybe set the default to `None` and note that the default is `False` for `vtk>=9.6`?

Done in #8248

user27182:
The issues we're seeing with SLC files relating to `download_bolt_nut`, `download_knee_full`, and `download_embryo`, should all be fixed upstream into the 9.6 release with https://gitlab.kitware.com/vtk/vtk/-/merge_requests/12824.

The other major issues I'm seeing are with `connectivity`, where the `RegionId` array is wrong:


connectivity

And with environment textures. Not sure yet if this one is a feature/bug fix or if it's a regression:

Screenshot 2026-01-19 at 7 17 09 AM

And one of the point labels here is not rendered:

Screenshot 2026-01-19 at 7 17 52 AM

There is also a change to `ExplicitStructuredGrid`, though this may be a VTK fix to ghost cells? CC @keurfonluu 
Screenshot 2026-01-18 at 12 51 58 PM

user27182:
> The other major issues I'm seeing are with `connectivity`, where the `RegionId` array is wrong:

The issue appears not to be with the `connectivity` filter itself or the `RegionId` array, but I think it's an issue with the mapper when using `categories=True`. Using the `color_labels` filter instead is a workaround for this, and is done in https://github.com/pyvista/pyvista/pull/8228/commits/c2991ef2a5a43ef1027e8ada18e4e5e06cb89f83

user27182:
> > The other major issues I'm seeing are with `connectivity`, where the `RegionId` array is wrong:
> 
> The issue appears not to be with the `connectivity` filter itself or the `RegionId` array, but I think it's an issue with the mapper when using `categories=True`. Using the `color_labels` filter instead is a workaround for this, and is done in [c2991ef](https://github.com/pyvista/pyvista/commit/c2991ef2a5a43ef1027e8ada18e4e5e06cb89f83)

Actually, it's not related to `categories=True`, but I still think it's a mapper-related issue. `connectivity` adds _both_ point data and cell data arrays named `RegionId`. If we clear the point data specifically, the issue is fixed, e.g.

This is broken:
``` py
import pyvista as pv
large = pv.Sphere(
    center=(-4, 0, 0), phi_resolution=40, theta_resolution=40
)
medium = pv.Sphere(
    center=(-2, 0, 0), phi_resolution=15, theta_resolution=15
)
small = pv.Sphere(center=(0, 0, 0), phi_resolution=7, theta_resolution=7)
mesh = large + medium + small
conn = mesh.connectivity('all')
conn.plot(cmap=['red', 'green', 'blue'], show_edges=True)
```
But if we do `conn.point_data.clear()` before plotting, then it renders as expected.

user27182:
UPDATE: I've marked all comments as resolved. This PR is effectively ready for review, and the current diff for the doc images all appear correct to me. 

But there are still doc image test failures due to VTK regressions, and we need to wait for the next VTK release (either RC3 or the proper 9.6.0 release) to fix these two issues:
- SLCReader spacing https://gitlab.kitware.com/vtk/vtk/-/issues/19909
- environment texture rotations https://gitlab.kitware.com/vtk/vtk/-/issues/19910

pyvista/core/utilities/reader.py
```diff
@@ -297,6 +297,10 @@ def read(self):
             msg = 'File reader failed to read and/or produced no output.'
             raise RuntimeError(msg)
         data._post_file_load_processing()
+        if getattr(self, '_filename', None) and self._filename.endswith('.slc'):  # type: ignore[union-attr]
+            # SLCReader spacing is broken in VTK 9.6 https://gitlab.kitware.com/vtk/vtk/-/issues/19909
+            # Set explicitly as a workaround
+            data.spacing = (*data.spacing[0:2], 1.0)
```
user27182: This is a temp fix for the SLCReader regression, and can be removed once the next 9.6 release is made.
```suggestion
```

akaszynski:
> we need to wait for the next VTK release (either RC3 or the proper 9.6.0 release) to fix these two issues:

Let's wait to review then. Looking at [VTK Milestones - 9.6.0](https://gitlab.kitware.com/vtk/vtk/-/milestones/27#tab-issues) it should be soon.

One troubling difference/regression is the apparent inversion of the visible faces when casting to an `UnstructuredGrid`:

sc009

Is this just a behavior change or is this a regression? Other than that, all image diffs are acceptable.

user27182:
> > we need to wait for the next VTK release (either RC3 or the proper 9.6.0 release) to fix these two issues:
> 
> Let's wait to review then. Looking at [VTK Milestones - 9.6.0](https://gitlab.kitware.com/vtk/vtk/-/milestones/27#tab-issues) it should be soon.
> 
> One troubling difference/regression is the apparent inversion of the visible faces when casting to an `UnstructuredGrid`:
> 
> sc009
> Is this just a behavior change or is this a regression? Other than that, all image diffs are acceptable.

This does seem like a regression.

I believe this is related to fixes/updates to ghost cells with VTK. For this specific example of `ExplicitStructuredGrid.cast_to_unstructured_grid`, there is a call to `hide_cells`: https://github.com/pyvista/pyvista/blob/0c47ed5f4631dff1e201ab021ffc46033f6bca96/pyvista/core/pointset.py#L3359-L3360

The `hide_cells` image is fine, but the broken image is when the `ExplicitStructuredGrid` _with hidden cells_ is then cast to `UnstructuredGrid`. So it may be an issue related to `vtkAppendFilter` and/or representations of ghost cells in `UnstructuredGrid` ?

But in the example, the `UnstructuredGrid` is then _re-cast_ back to `ExplicitStructuredGrid`, and the plot appears correct again. So the round-trip conversion seems to work still, but there seems to be a problem in between.

See the development preview docs:
https://696fd97e11d478536c2d0dd7--pyvista-dev.netlify.app/api/core/_autosummary/pyvista.explicitstructuredgrid.cast_to_unstructured_grid

I am very much unfamiliar with ghost cells and explicit structured grids. Maybe @keurfonluu knows more about this and what's expected?

user27182:
The example works with `vtk==9.5.20250802.dev0` but is broken with `vtk==9.5.20250809.dev0`
I tried looking at the diff https://github.com/Kitware/VTK/compare/6ac70ce2dc4520624d3bc398097aa8e264302368...5cda859050d00a943eb1007d995fc6fdf37eec24 between the Aug 2 timestamp https://github.com/Kitware/VTK/commit/6ac70ce2dc4520624d3bc398097aa8e264302368 and the aug 9 timestamp https://github.com/Kitware/VTK/commit/5cda859050d00a943eb1007d995fc6fdf37eec24 but nothing is jumping out in terms of changes to ghost cells affecting unstructured grids...
Might need to bisect locally between these commits and test VTK builds

user27182:
> The example works with `vtk==9.5.20250802.dev0` but is broken with `vtk==9.5.20250809.dev0` I tried looking at the diff [Kitware/VTK@6ac70ce...5cda859](https://github.com/Kitware/VTK/compare/6ac70ce2dc4520624d3bc398097aa8e264302368...5cda859050d00a943eb1007d995fc6fdf37eec24) between the Aug 2 timestamp [Kitware/VTK@6ac70ce](https://github.com/Kitware/VTK/commit/6ac70ce2dc4520624d3bc398097aa8e264302368) and the aug 9 timestamp [Kitware/VTK@5cda859](https://github.com/Kitware/VTK/commit/5cda859050d00a943eb1007d995fc6fdf37eec24) but nothing is jumping out in terms of changes to ghost cells affecting unstructured grids... Might need to bisect locally between these commits and test VTK builds

I posted in the VTK discourse: https://discourse.vtk.org/t/possible-vtk-9-6-0rc2-regression-with-ghost-cell-rendering/16251

akaszynski:
@user27182, I'll need to restart the cluster, NVIDIA drivers updated. Expect it (CI) to be unstable for the next few hours while I iron out the issues.

user27182:
The issue with ghost cell rendering is tracked here: https://gitlab.kitware.com/vtk/vtk/-/issues/19922
But it doesn't look like it'll be fixed for 9.6.0.

I've added a workaround here though to use `vtkDataSetSurfaceFilter` explicitly for these specific cases
https://github.com/pyvista/pyvista/pull/8228/commits/dd7c2d4b2bd984e68bc14a676cfc6f95dd8a2c9e

I'm going to mark earlier comments related to this regression as resolved.

akaszynski:
> But it doesn't look like it'll be fixed for 9.6.0.

Nope. Recommend addressing it in pyvista until it is. Thanks for doing so.

user27182:
I've reviewed the failed doc images and they all look like SLC datasets and/or cubemap plots, both of which should be fixed in 9.6.0rc3
Seems like we just need to wait for 9.6.0rc3 to be released.

akaszynski:
> I've reviewed the failed doc images and they all look like SLC datasets and/or cubemap plots, both of which should be fixed in 9.6.0rc3 Seems like we just need to wait for 9.6.0rc3 to be released.

Looks to be soon (30 January):
https://discourse.vtk.org/t/vtk-9-6-0-release-planning/15760/7

user27182:
RC3 released: https://pypi.org/project/vtk/9.6.0rc3/

user27182:
> I've reviewed the failed doc images and they all look like SLC datasets and/or cubemap plots, both of which should be fixed in 9.6.0rc3

These have all been fixed. All images are now up-to-date.

The only issue now is that there is a failure with the `resampling.py` gallery example.

akaszynski:
> The only issue now is that there is a failure with the `resampling.py` gallery example.

Issue seems to be with how we treat ghost cells in:
https://github.com/pyvista/pyvista/blob/fb16ef20635f62014c33209287dd7196b02739e1/pyvista/plotting/plotter.py#L3675-L3688

Specifically this line:
```py
            not_hidden.points = np.append(not_hidden.points, bounds_points, axis=0)
```
```
(Pdb) mesh
UnstructuredGrid (0x7ff58994bdc0)
  N Cells:    16616
  N Points:   26483
  X Bounds:   3.791e+01, 2.033e+02
  Y Bounds:   7.926e+01, 2.068e+02
  Z Bounds:   2.068e+01, 2.205e+02
  N Arrays:   4
(Pdb) not_hidden.points = np.append(not_hidden.points, bounds_points, axis=0)
(Pdb) mesh
UnstructuredGrid (0x7ff58994bdc0)
  N Cells:    16616
  N Points:   26491
  X Bounds:   3.791e+01, 2.033e+02
  Y Bounds:   7.926e+01, 2.068e+02
  Z Bounds:   2.068e+01, 2.205e+02
  N Arrays:   4
(Pdb) 
```

Running it changes the number of points shown in `mesh` (and then not_hidden), without changing the number of points in the active scalars within `point_data`.

Also note that adding points to `not_hidden` also changes the points in `mesh`, mutating the input.

```
-> not_hidden.points = np.append(not_hidden.points, bounds_points, axis=0)
(Pdb) not_hidden.points
pyvista_ndarray([[130.94595 , 124.054054,  20.675676],
                 [134.39189 , 124.054054,  20.675676],
                 [137.83784 , 124.054054,  20.675676],
                 ...,
                 [124.054054, 144.72974 , 220.54054 ],
                 [127.5     , 144.72974 , 220.54054 ],
                 [130.94595 , 144.72974 , 220.54054 ]],
                shape=(26483, 3), dtype=float32)
(Pdb) mesh.points
pyvista_ndarray([[130.94595 , 124.054054,  20.675676],
                 [134.39189 , 124.054054,  20.675676],
                 [137.83784 , 124.054054,  20.675676],
                 ...,
                 [124.054054, 144.72974 , 220.54054 ],
                 [127.5     , 144.72974 , 220.54054 ],
                 [130.94595 , 144.72974 , 220.54054 ]],
                shape=(26483, 3), dtype=float32)
(Pdb) np.may_share_memory(mesh.points, not_hidden.points)
True
(Pdb) 
```

```
(Pdb) not_hidden.points = np.append(not_hidden.points, bounds_points, axis=0)
(Pdb) np.may_share_memory(mesh.points, not_hidden.points)
True
```

user27182:
> > The only issue now is that there is a failure with the `resampling.py` gallery example.
> 
> Issue seems to be with how we treat ghost cells in:
>
> Running it changes the number of points shown in `mesh` (and then not_hidden), without changing the number of points in the active scalars within `point_data`.
> 

Thanks for identifying this Do you think there's a better way to implement the workaround code here? Or you're fine with it, just need to update the point data too?

> Also note that adding points to `not_hidden` also changes the points in `mesh`, mutating the input.

This is surprising, because I thought `extract_cells` would have created a deep copy. And not just as a PyVista thing (preference for deep over shallow copies), but necessarily by the vtk filter since the points of extracted cells should differ from the input... unless there are no hidden cells here, but we still have `vtkGhostType` ? Indeed I just checked, and `sample` generates exactly that... it includes a `vtkGhostType` array with all zeros. So this is what triggers the branch to execute. And I recall that the vtk code for extract cells has a fast path that returns the input as-is when n_cells being extracted == n_cells of the input. I think this is a bug with `extract_cells` that we should fix (separately).

keurfonluu:
> This is surprising, because I thought `extract_cells` would have created a deep copy. And not just as a PyVista thing (preference for deep over shallow copies), but necessarily by the vtk filter since the points of extracted cells should differ from the input... unless there are no hidden cells here, but we still have `vtkGhostType` ? Indeed I just checked, and `sample` generates exactly that... it includes a `vtkGhostType` array with all zeros. So this is what triggers the branch to execute. And I recall that the vtk code for extract cells has a fast path that returns the input as-is when n_cells being extracted == n_cells of the input. I think this is a bug with `extract_cells` that we should fix (separately).

Any chance it's related to https://gitlab.kitware.com/vtk/vtk/-/issues/19723?
Not sure why `vtkExtractSelection()` is used instead of `vtkExtractCells()` (which doesn't remove ghost cells, per my own test): https://github.com/pyvista/pyvista/blob/f4a5296c02e1573e2a6112629167f9aad2206076/pyvista/core/filters/data_set.py#L4524

user27182:
> > This is surprising, because I thought `extract_cells` would have created a deep copy. And not just as a PyVista thing (preference for deep over shallow copies), but necessarily by the vtk filter since the points of extracted cells should differ from the input... unless there are no hidden cells here, but we still have `vtkGhostType` ? Indeed I just checked, and `sample` generates exactly that... it includes a `vtkGhostType` array with all zeros. So this is what triggers the branch to execute. And I recall that the vtk code for extract cells has a fast path that returns the input as-is when n_cells being extracted == n_cells of the input. I think this is a bug with `extract_cells` that we should fix (separately).
> 
> 
> 
> Any chance it's related to https://gitlab.kitware.com/vtk/vtk/-/issues/19723?
> 
> Not sure why `vtkExtractSelection()` is used instead of `vtkExtractCells()` (which doesn't remove ghost cells, per my own test): https://github.com/pyvista/pyvista/blob/f4a5296c02e1573e2a6112629167f9aad2206076/pyvista/core/filters/data_set.py#L4524
> 
> 

Nah the code here isn't trying to keep ghost cells, the issue is the ghost cells don't render properly, so we're explicitly extracting non-hidden cells to _remove_ ghost cells.

The fast execution path I'm referring to is mentioned here https://gitlab.kitware.com/vtk/vtk/-/issues?sort=created_date&state=opened&search=vtkExtractSelection&first_page_size=20&show=eyJpaWQiOiIxOTc0NyIsImZ1bGxfcGF0aCI6InZ0ay92dGsiLCJpZCI6NTMxMDEyfQ%3D%3D

user27182:
> The issue with ghost cell rendering is tracked here: https://gitlab.kitware.com/vtk/vtk/-/issues/19922 But it doesn't look like it'll be fixed for 9.6.0.
> 
> I've added a workaround here though to use `vtkDataSetSurfaceFilter` explicitly for these specific cases [dd7c2d4](https://github.com/pyvista/pyvista/commit/dd7c2d4b2bd984e68bc14a676cfc6f95dd8a2c9e)
> 
> I'm going to mark earlier comments related to this regression as resolved.

Might be fixed for 9.6 after all: https://gitlab.kitware.com/vtk/vtk/-/merge_requests/12877
We'll need to revert the rendering workaround if that's the case. And we'll need to wait for the 9.6 release to land first before we can test.

EDIT: Nope, the fix is targeted for VTK 9.6.1

Issue #8234, opened 2026-01-09
UnstructuredGrid construction with polyhedron failing


antoinecollet5:
### Describe the bug, what's wrong, and what you expected.

First of all, thanks for this great tool ;)

I try to create complex unstructured grids with POLYHEDRON cell types. I follow https://docs.pyvista.org/examples/00-load/create_polyhedron#polyhedron-connectivity-array but it does not seem to work as expected, or maybe I am doing something wrong... but I cannot figure what ? Maybe the order of points in the faces definition but it makes no sense to me... ???

### Steps to reproduce the bug.

Here is a simple test with one cell. I will first create a polydata to demonstrate that all seems fine with the data, and then move to the unstructured grid approach where I fail to get what I want.

```python
import numpy as np
import pyvista as pv

test_faces = [
    [21, 20, 0, 1, 4],
    [14, 15, 16],
    [14, 15, 4, 1, 7, 8, 3],
    [8, 7, 13, 11],
    [3, 8, 11, 10],
    [7, 1, 0, 2, 12, 13],
    [12, 9, 10, 11, 13],
    [9, 6, 18, 17, 16, 14, 3, 10],
    [2, 5, 6, 9, 12],
    [17, 21, 4, 15, 16],
    [0, 20, 19, 5, 2],
    [20, 21, 17, 18, 19],
    [19, 18, 6, 5],
]

test_vertices = [
    [63.89493459, 267.15312404, 74.80558535],
    [110.54145631, 273.59466265, 67.64924053],
    [59.34378268, 265.31658383, 86.60516832],
    [118.44905499, 245.36697435, 86.55446217],
    [89.8413722, 229.64676298, 50.67136423],
    [59.42347354, 256.18052528, 84.48475809],
    [67.99803334, 238.6987468, 85.93195597],
    [114.53452566, 273.46459276, 73.299068],
    [118.98591774, 264.82710911, 81.94037802],
    [82.47579797, 247.16451464, 97.4075909],
    [112.91306495, 249.73127377, 95.27243464],
    [112.01588852, 264.15615259, 94.08529455],
    [73.94351127, 266.03204737, 96.31754568],
    [105.23478398, 270.08285036, 94.00075211],
    [92.78958104, 224.56440319, 56.21849071],
    [89.93824247, 227.86629306, 51.29877142],
    [87.58433904, 223.22143463, 55.02498018],
    [70.39638055, 227.97247969, 66.92129774],
[...]

PR #8236, opened 2026-01-11
Add `font_size`, `point_size`, `line_width` and `normals_scale` keywords to `plot_cell`

user27182:
### Overview

https://github.com/pyvista/pyvista/pull/6567 increased the values of `font_size`, `point_size` and `line_width` for `plot_cell` so that documentation images for plotted cells have nicer proportions. This "fixed" the documentation renderings, but it "broke" renderings when using `plot_cell` normally, since the scale is now much too large.

This change is not ideal, since users may now avoid using `plot_cell` and instead implement custom scripts to get the scaling correct, e.g. see https://github.com/pyvista/pyvista/issues/8234#issue-3798464754.

This PR fixes this, and effectively reverts changes from https://github.com/pyvista/pyvista/pull/6567 to restore the original values for `font_size`, `point_size` and `line_width`, and only uses the newer values when building documentation.

New keywords are also added to customize these values directly, along with the option to customize the size of any cell normals.


pyproject.toml
```diff
@@ -249,6 +249,7 @@ filterwarnings = [
   'ignore:.*numpy.dtype size changed.*:RuntimeWarning', # bogus numpy ABI warning (see numpy/#432)
   'ignore:.*numpy.ufunc size changed.*:RuntimeWarning', # bogus numpy ABI warning (see numpy/#432)
   'ignore:Passing .{1}N.{1} to ListedColormap is deprecated since:DeprecationWarning', # https://github.com/matplotlib/cmocean/pull/114
+  'ignore:Setting the shape on a NumPy array has been deprecated in NumPy 2.5.', # https://numpy.org/devdocs/release/2.5.0-notes.html#setting-the-shape-attribute-is-deprecated
```
user27182, line 252: This is also fixed separately by #8236

pyvista/plotting/themes.py
```diff
@@ -200,6 +200,8 @@ def to_dict(self) -> dict[str, Any]:
         for key in self._all__slots__():
             value = getattr(self, key)
             key_ = key[1:]
+            if key_ == 'plot_cell':  # private config values
+                continue
```
user27182, line 204: I don't think we should expose the new theme config for `plot_cell` as part of the public API, the theme is strictly being used for internal config. And so ignoring this key is necessary for theme save/load tests to pass.

akaszynski: Agreed.

PR #8241, opened 2026-01-14
PyVista patch `0.46.5`

user27182:
### Overview

Fix #8112, and backport a fix when deleting obbTree, see https://github.com/pyvista/pyvista/issues/8204#issuecomment-3711475421

PR #8007, opened 2025-10-10
Use trusted publishing for PyPI

akaszynski:
In support of https://github.com/pyvista/pyvista/pull/7992 and as part of a growing [best practice](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/), this PR updates our testing and deployment workflow to use PyPI's [Trusted Publishing](https://docs.pypi.org/trusted-publishers/) to upload releases to PyPI.

We've already created the trusted publisher connection at [pyvista publishing](https://pypi.org/manage/project/pyvista/settings/publishing):
sc004

Once this PR is merged we can safely delete the `TWINE_PASSWORD` GitHub secret.

Note: This PR also bumps the Python version used for release to 3.13, giving us a few more years to update it.

PR #8243, opened 2026-01-14
Use trusted publishing for release `0.46.5`

user27182:
### Overview

Similar to #8007 , but for the `release/0.46` branch to address PyPI upload failure https://github.com/pyvista/pyvista/pull/8241#issuecomment-3752068711

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -167,6 +167,13 @@ jobs:
           - python-version: "3.13"
             vtk-version: "latest"
             numpy-version: "nightly"
+
+    environment:
+      name: pypi
+      url: https://pypi.org/p/pyvista
+    permissions:
+      id-token: write # this permission is mandatory for trusted publishing
+
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


akaszynski:
Confirming that trusted publishing is set up in PyPI:

sc002

.github/workflows/testing-and-deployment.yml
```diff
@@ -271,13 +278,8 @@ jobs:
           twine check --strict dist/*
 
       - name: Upload to PyPi
```
akaszynski: Nitpick out of the scope of this PR.
```suggestion
      - name: Upload to PyPI
```

akaszynski:
Ah shoot, this is for the release branch. Ignore the comments.

akaszynski:
Tracking release workflow here: https://github.com/pyvista/pyvista/actions/runs/21013930087

akaszynski:
> Tracking release workflow here: https://github.com/pyvista/pyvista/actions/runs/21013930087

Worked, though we do have a potentially bogus failure:
https://github.com/pyvista/pyvista/actions/runs/21013930087/job/60415312810

Issue might be due to duplicate wheels.

user27182:
We should consider separating unit testing from deployment. And only run the deployment job once after the rest of CI passes

user27182:
> We should consider separating unit testing from deployment. And only run the deployment job once after the rest of CI passes

This is already done in main:
https://github.com/pyvista/pyvista/blob/94788df30cc1e567e6ec954544d2331b80037e67/.github/workflows/testing-and-deployment.yml#L382

PR #8250, opened 2026-01-19
Rotate 180 for env texture for vtk>=9.6

akaszynski:
This PR addresses the behavior change pointed out in #8228 where the enviornment texture is flipped about the Y axis by 180 degrees in VTK v9.6.

I think the base behavior change by VTK is acceptable, but our plot should flip the view direction of the environment as the sphere reflects the enviornment back, which is not the case anymore on `main` for `vtk==9.6.0rc2`. Turning down the roughness you can see that the default behavior is to directly map the environment onto the skybox:

```py
from pyvista import examples

cube_map = examples.download_sky_box_cube_map()
cube_map.plot(roughness=0.5)
```
sc008

This PR reverts this behavior by simply flipping the enviornment rotation about the Y axis which results in:

sc006

### DIFF
--- a/pyvista/plotting/texture.py
+++ b/pyvista/plotting/texture.py
@@ -562,7 +562,8 @@ class Texture(DataObject, _vtk.vtkTexture):
 
         # maintain backwards compatibility with VTK <9.6
         if pv.vtk_version_info >= (9, 6, 0):
-            vmat = pv.vtkmatrix_from_array(pv.Transform().rotate_y(180).rotation_matrix)
+            rotate_y_180 = np.array([[-1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, -1.0]])
+            vmat = pv.vtkmatrix_from_array(rotate_y_180)
             pl.renderer.SetEnvironmentRotationMatrix(vmat)
 
         pl.add_mesh(
--- a/pyvista/plotting/texture.py
+++ b/pyvista/plotting/texture.py
@@ -559,7 +559,13 @@ class Texture(DataObject, _vtk.vtkTexture):
         pl = pv.Plotter(lighting=lighting)
         pl.add_actor(self.to_skybox())
         pl.set_environment_texture(self, is_srgb=True)
-        pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.5, metallic=1.0)
+
+        # maintain backwards compatibility with VTK <9.6
+        if pv.vtk_version_info >= (9, 6, 0):
+            vmat = pv.vtkmatrix_from_array(pv.Transform().rotate_y(180).as_rotation().as_matrix())
+            pl.renderer.SetEnvironmentRotationMatrix(vmat)
+
+        pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.1, metallic=1.0)
         pl.camera_position = cpos
         pl.camera.zoom(zoom)
         if show_axes:
--- a/pyvista/plotting/texture.py
+++ b/pyvista/plotting/texture.py
@@ -561,8 +561,8 @@ class Texture(DataObject, _vtk.vtkTexture):
         pl.set_environment_texture(self, is_srgb=True)
 
         # maintain backwards compatibility with VTK <9.6
-        if pv.vtk_version_info >= (9, 6, 0):  # pragma: no cover
-            vmat = pv.vtkmatrix_from_array(pv.Transform().rotate_y(180).as_rotation().as_matrix())
+        if pv.vtk_version_info >= (9, 6, 0):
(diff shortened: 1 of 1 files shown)


pyvista/plotting/texture.py
```diff
@@ -559,7 +559,18 @@ def _plot_skybox(self, **kwargs):
         pl = pv.Plotter(lighting=lighting)
         pl.add_actor(self.to_skybox())
         pl.set_environment_texture(self, is_srgb=True)
-        pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.5, metallic=1.0)
+
+        # maintain backwards compatibility with VTK <9.6
+        if pv.vtk_version_info >= (9, 6, 0):  # pragma: no cover
+            vmat = pv.vtkmatrix_from_array(pv.Transform().rotate_y(180).as_rotation().as_matrix())
+            pl.renderer.SetEnvironmentRotationMatrix(vmat)
```
user27182:
```suggestion
        if pv.vtk_version_info >= (9, 6, 0):
            vmat = pv.vtkmatrix_from_array(pv.Transform().rotate_y(180).rotation_matrix)
            pl.renderer.SetEnvironmentRotationMatrix(vmat)
```

We should use `vtk-dev-testing` label to get coverage. 

And `as_rotation` needs Scipy.

EDIT: might be better to just copy the 3x3 array directly instead of using Transform to compute it directly.

user27182:
@akaszynski this fixes `_plot_skybox`, which is called by `Texture.plot`, but it doesn't fix the general `Plotter` case, e.g. this is still broken (from `examples.download_cubemap_park`):
``` python

    >>> import pyvista as pv
    >>> pl = pv.Plotter(lighting=None)
    >>> dataset = examples.download_cubemap_park()
    >>> _ = pl.add_actor(dataset.to_skybox())
    >>> pl.set_environment_texture(dataset, is_srgb=True)
    >>> pl.camera_position = 'xy'
    >>> pl.camera.zoom(0.4)
    >>> _ = pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.1, metallic=0.5)
    >>> pl.show()
```

I think a more general fix like this commit https://github.com/pyvista/pyvista/pull/8228/commits/baf929e8b51a74d61de61b0f97e2b1dd0067c458 from #8228 might work better?

user27182:
> @akaszynski this fixes `_plot_skybox`, which is called by `Texture.plot`, but it doesn't fix the general `Plotter` case, e.g. this is still broken (from `examples.download_cubemap_park`):
> 
> ```python
>     >>> from pyvista import examples
>     >>> import pyvista as pv
>     >>> pl = pv.Plotter(lighting=None)
>     >>> dataset = examples.download_cubemap_park()
>     >>> _ = pl.add_actor(dataset.to_skybox())
>     >>> pl.set_environment_texture(dataset, is_srgb=True)
>     >>> pl.camera_position = 'xy'
>     >>> pl.camera.zoom(0.4)
>     >>> _ = pl.add_mesh(pv.Sphere(), pbr=True, roughness=0.1, metallic=0.5)
>     >>> pl.show()
> ```
> 
> I think a more general fix like this commit [baf929e](https://github.com/pyvista/pyvista/commit/baf929e8b51a74d61de61b0f97e2b1dd0067c458) from #8228 might work better?

That change is also not acceptable since it results in other failures: https://github.com/pyvista/pyvista/actions/runs/21153690923/job/60834673976?pr=8228 

> This PR addresses the behavior change pointed out in #8228 where the enviornment texture is flipped about the Y axis by 180 degrees in VTK v9.6.
> 
> I think the base behavior change by VTK is acceptable

Maybe this is a VTK regression after all, and the new environment texture rotations have the wrong defaults? We could open a discussion with VTK folks and confirm if this is this change is intentional.

We could alternatively flip the cube map texture itself on init so that it renders properly. Or we could expose the environment rotation as a new keyword for `set_environment_texture`, but then we'd need to add this keyword to every cubemap example, and only add the keyword conditionally (based on VTK version).

akaszynski:
> We could alternatively flip the cube map texture itself on init so that it renders properly. Or we could expose the environment rotation as a new keyword for `set_environment_texture`, but then we'd need to add this keyword to every cubemap example, and only add the keyword conditionally (based on VTK version).

Let's open up a discussion first. We'd have to expose each axis in the kwargs, probably better to set the environment rotation transformation either in Plotter or in the Renderer class.

user27182:
I started a discussion: https://discourse.vtk.org/t/possible-9-6-0-regression-with-environment-texture-rendering/16247

akaszynski:
Looks like this will be fixed in `vtk`. We should expose enviornment rotation eventually, but for now we can drop this PR. Thanks for your work investigating the root cause.

Issue #5350, opened 2023-12-15
Add support for CalculiX results file [frd]


efirvida:
### Describe the feature you would like to be added.

[Calculix](https://www.calculix.de/) Is opensourcegeneral propose FEM Solver, it will be nice to have native support in PyVista. I could not find the frd documentación, but here I link the code of two scripts to convert the frd to vtk.

https://github.com/calculix/ccx2paraview

https://gitlab.kitware.com/francois.mazen/calculix-to-vtk

### Pseudocode or Screenshots

```python
Insert code here
```
Or drag your screenshot here!

tkoyama010:
[pyvista.read](https://docs.pyvista.org/version/stable/api/utilities/_autosummary/pyvista.read.html) use meshio to read and there is a PR of adding CalculiX results file.
https://github.com/nschloe/meshio/pull/1365
After release this feature, we can read frd files.

3rav:
I am in favor of native support for this format in pyvista.

user27182:
The meshio PR https://github.com/nschloe/meshio/pull/1365 has been closed, it seems development has all but stopped for that repo. There is precedent in PyVista for supporting custom readers, e.g. the addition of [`read_grdecl`](https://docs.pyvista.org/api/utilities/_autosummary/pyvista.read_grdecl.html) from https://github.com/pyvista/pyvista/pull/6516,  so I think this is something we can support. But we need someone to do the work to create the reader and submit a PR.

In terms of the implementation, the [`ccx2paraview` reader uses GPL3](https://github.com/calculix/ccx2paraview?tab=GPL-3.0-1-ov-file#readme), which I don't think is compatible with MIT. And [calculix-to-vtk](https://gitlab.kitware.com/francois.mazen/calculix-to-vtk) doesn't have any licensing details. Maybe we can use @Krande's implementation from https://github.com/nschloe/meshio/pull/1365 (with their permission)?

3rav:
See this (python, MIT):
https://github.com/wr1/frd2vtu

or (Visual, ?):
https://forum.freecad.org/viewtopic.php?p=122348&sid=890c97b4adbb9657d5ef6c766f57a98f#p122348

Krande:
Hey, feel free to use what I did for https://github.com/nschloe/meshio/pull/1365. You probably want to use it primarily as a starting point and write your own implementation from it. (I think there is a lot of things to improve on what I did back then)

QiangF:
[...]

PR #8255, opened 2026-01-22
Add the ability to read CalculiX .frd files

3rav:
Test:
1D:
1D

2D:
2D

3D C3D8:
C3D8

3D C3D20:
C3D20

3D Tet10:
image

Closes https://github.com/pyvista/pyvista/issues/5350

### DIFF
--- a/tests/core/test_frd.py
+++ b/tests/core/test_frd.py
@@ -175,19 +175,19 @@ def test_frd_reader_time_steps(mock_frd_file):
 def test_frd_reader_set_active_time_value(mock_frd_file):
     reader = pv.FRDReader(mock_frd_file)
 
-    # Exact time match is required. 
+    # Exact time match is required.
     # The mock_frd_file has time steps: 0.1, 0.2, 0.3.
-    
+
     # 1. Check setting an invalid time (e.g., 0.18 instead of 0.2)
-    with pytest.raises(ValueError, match="Not a valid time"):
+    with pytest.raises(ValueError, match='Not a valid time'):
         reader.set_active_time_value(0.18)
-        
+
     # 2. Check setting a valid time using the method
     reader.set_active_time_value(0.2)
     assert reader.active_time_value == 0.2
 
     # 3. Check setting an invalid time via the setter property
-    with pytest.raises(ValueError, match="Not a valid time"):
+    with pytest.raises(ValueError, match='Not a valid time'):
         reader.active_time_value = 0.29
 
     # 4. Check setting a valid time via the setter property


MatthewFlamm:
Thank you for this PR!   I don't think we have good documentation on adding readers.  I'm torn on adding in denovo readers that do not exist in vtk or ParaView.  I believe all existing readers are currently either in vtk or ParaView.  My main concern is maintainability.  However, adding readers for more file types makes PyVista more widely usable.

Assuming we want to take on adding de novo readers, I think it is important that we have test files.  We currently store those in pyvista/vtk-data.  Please see examples of how this is used in the current reader tests.  Some tests use prepackaged download functions while some cherry pick files using `pyvista.examples.downloads.download_file`.

This reader should be in `pyvista/core/utilities/reader.py` and inherit from at least `BaseReader`.  The other readers that are de novo create a vtk-like reader that inherits from `BaseVTKReader` and implements the main logic.  This class is wrapped by the one that inherits from `BaseReader`.  See recently added `SeriesReader` as an example, or any that inherit from `BaseVTKReader`.

Finally, the implementation should look forward to implementing support for time series and cell and point data selection (see the mixin classes in the same module as above) if the file format allows this.

pyvista/core/utilities/fileio.py
```diff
@@ -310,6 +310,10 @@ def read(  # noqa: PLR0911, PLR0917
     ext = _get_ext_force(filename, force_ext)
     if ext in ['.e', '.exo']:
         return read_exodus(filename)
+    if ext == '.frd':
+        from .frd import FRDReader  # noqa: PLC0415
+
+        return FRDReader(filename).read()
```
user27182: FRDReader should be defined in `reader.py` and `.frd` should be added to `CLASS_READERS`. These lines can then be removed.

pyvista/core/utilities/_frd.py
```diff
@@ -0,0 +1,483 @@
...
+from pyvista.core.utilities.reader import BaseReader
+from pyvista.core.utilities.reader import BaseVTKReader
+from pyvista.core.utilities.reader import TimeReader
+
+# ---------------------------------------------------------------------------
+# Low-level, VTK-style reader
+# ---------------------------------------------------------------------------
+
+
+class _FRDVTKReader(BaseVTKReader):
```
user27182: I think we can move the FRD classes into `reader.py`. The reader methods should be moved there, but you can keep any of the FRD-specific definitions and parsing methods (e.g. static methods) in `frd.py` to keep some separation.

(There are tests that run automatically on reader classes, but only if they're defined in reader.py)

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+        4: 25,  # he20 -> VTK_QUADRATIC_HEXAHEDRON
+        5: 13,  # pe15 -> VTK_WEDGE (degraded; pe15 not native to base VTK)
+        6: 24,  # te10 -> VTK_QUADRATIC_TETRA
+        7: 5,  # tr3  -> VTK_TRIANGLE
+        8: 22,  # tr6  -> VTK_QUADRATIC_TRIANGLE
+        9: 9,  # qu4  -> VTK_QUAD
+        10: 23,  # qu8  -> VTK_QUADRATIC_QUAD
+        11: 3,  # be2  -> VTK_LINE
+        12: 21,  # be3  -> VTK_QUADRATIC_EDGE
+    }
```
user27182: Please use enums for this. You can use `pv.CellType` for the VTK cell type mapping.

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+        4: 20,
+        5: 15,
+        6: 10,
+        7: 3,
+        8: 6,
+        9: 4,
+        10: 8,
+        11: 2,
+        12: 3,
+    }
```
user27182: Use enums here too. For the LHS you can use a new FRD enum, and for the RHS you can import `_CELL_TYPE_TO_NUM_POINTS` from `pyvista.core.celltype` and use that to lookup the number of nodes.

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+                node_ids[17],
+                node_ids[18],
+                node_ids[19],
+                node_ids[12],
+                node_ids[13],
+                node_ids[14],
+                node_ids[15],
+            ]
+
+        if etype == 7 and len(node_ids) == 3:
```
user27182: Avoid magic numbers in the code, use enum instead.

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+    # ------------------------------------------------------------------
+    # Block parsers (Memory efficient iterators)
+    # ------------------------------------------------------------------
+
+    def _parse_lines(self, fh: Any) -> None:
+        """Dispatch lines to block-specific parsers."""
+        for line in fh:
+            s = line.strip()
+            if s.startswith('2C'):
+                self._parse_nodes(fh)
```
user27182: I don't know anything about this format... what is 2C and 3C and 100? Should this also use an enum for better readability/maintenance?

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+                self._parse_elements(fh)
+            elif s.startswith('100'):
+                _step_id, step_time = self._parse_100cl_header(s)
+                self._parse_results(fh, step_time)
+
+    def _parse_nodes(self, fh: Any) -> None:
+        """Parse 2C node block."""
+        for line in fh:
+            s = line.strip()
+            if s.startswith('-3'):
```
user27182: What does -3 mean?

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+        node_ids: list[int] = []
+        etype: int | None = None
+        vtk_type: int | None = None
+
+        for line in fh:
+            s = line.strip()
+            if s.startswith('-3'):
+                return
+
+            if s.startswith('-1'):
```
user27182: What does -1 mean?

akaszynski:
Nice work! Can we add in a more substantial example file to https://github.com/pyvista/vtk-data for the purpose of adding it to [pyvista.examples.downloads](https://docs.pyvista.org/api/examples/_autosummary/pyvista.examples.downloads.html)?

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+            if s.startswith('-3'):
+                return
+            try:
+                parts = self._fix_scientific(s).split()
+                nid = int(parts[1])
+                self._nodes[nid] = [float(parts[2]), float(parts[3]), float(parts[4])]
+            except (ValueError, IndexError):
+                pass
+
+    def _parse_elements(self, fh: Any) -> None:
```
user27182: Can we use more descriptive variables? Is this a file header?

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+                # Pass the first data line and iterator down to collect data
+                self._parse_result_data(s, fh, name, step_bucket)
+                return
+            elif s.startswith('-3'):
+                return
+
+    def _parse_result_data(  # noqa: PLR0917
+        self, first_line: str, fh: Any, name: str, step_bucket: dict[str, dict[int, list[float]]]
+    ) -> None:
+        """Parse -1 records until -3 sentinel is hit."""
```
user27182: I'm thinking an enum for all these '-#' records may be a good idea.

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+            msg = 'No nodes found in FRD file -- cannot build grid.'
+            raise ValueError(msg)
+
+        sorted_ids = sorted(self._nodes)
+        node_map = {nid: idx for idx, nid in enumerate(sorted_ids)}
+        points = np.array([self._nodes[n] for n in sorted_ids], dtype=float)
+
+        cells: list[int] = []
+        types: list[int] = []
+        for conn, ctype in zip(self._elements, self._cell_types, strict=False):
```
user27182: Why `strict=False`? Should there not always be a 1:1 mapping between an element and its cell type?

tests/test_frd.py
```diff
@@ -0,0 +1,229 @@
+from __future__ import annotations
+
+import pytest
+
+import pyvista as pv
+from pyvista.core.utilities.frd import FRDReader
```
user27182: I don't think should we should import FRDReader here. Instead, use `pv.FRDReader`. This will implicitly test that it's available in the public root `pyvista` namespace.

tests/core/test_frd.py
```diff
@@ -0,0 +1,229 @@
...
+
+    reader = FRDReader(comprehensive_frd_file)
+    mesh = reader.read()
+
+    assert mesh.n_points == 4
+    assert reader.number_time_points == 3
+
+    # Check if derived stress fields were computed correctly
+    assert 'STRESS_vMises' in mesh.point_data
+    assert 'STRESS_sgMises' in mesh.point_data
```
user27182, line 271: This isn't really testing that it's computed _correctly_. Do we have known expected values we can assert?

tests/test_frd.py
```diff
@@ -0,0 +1,229 @@
...
+    return str(file_path)
+
+
+def test_frd_reader_init_and_read(mock_frd_file):
+    reader = FRDReader(mock_frd_file)
+    mesh = reader.read()
+
+    assert mesh.n_points == 8
+    assert mesh.n_cells == 1
+    assert 'Original_Node_ID' in mesh.point_data
```
user27182: Can we add a similar test using `pv.read` directly?

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,483 @@
...
+            cells.extend(vtk_ids)
+            types.append(ctype)
+
+        grid = UnstructuredGrid(
+            np.array(cells, dtype=np.int64),
+            np.array(types, dtype=np.uint8),
+            points,
+        )
+
+        grid.point_data['Original_Node_ID'] = np.array([str(nid) for nid in sorted_ids])
```
user27182: Is this a standard array naming convention for this file format? If not I'd prefer all lower case convention, e.g. `original_node_id`, or use all caps like the other arrays that are saved.

EDIT: And maybe make it plural? `original_node_ids`

user27182:
> Nice work! Can we add in a more substantial example file to https://github.com/pyvista/vtk-data for the purpose of adding it to [pyvista.examples.downloads](https://docs.pyvista.org/api/examples/_autosummary/pyvista.examples.downloads.html)?

+1

It looks like lots of arrays are computed and added by the reader as well. Should these arrays be documented in the docstring? Or at the very least, show the `point_data` array names and `cell_data` array names in either the FRDReader docstring (after reading/loading the example dataset), or in the new example's docstring in downloads.

3rav:
Feel free to make whatever changes you see fit. As long as the .frd file reader is functional, I'm happy.

user27182:
Unfortunately the process here is not as simple as "Make CI green and maintainers will do the rest of the work". We still need more input from you, both in terms of a real dataset uploaded to https://github.com/pyvista/vtk-data, and in terms of resolving review comments. For example, I do not know what '-1' means in the code, and need you to document this to make this clear (e.g. by using an enum with self-explanatory member names).

@MatthewFlamm mentioned earlier that the main concern with this addition is maintainability. That is also my main concern. If maintainers are already expected to make substantial changes on their own _before_ this PR even lands, then I worry about the potential long-term burden this will have on maintainers, and am less inclined to approve this contribution.

Please try to address some of the comments. If you need help or clarification, let us know.

akaszynski:
> Make CI green and maintainers will do the rest of the work

Nonetheless, I do appreciate how far you've gotten just to get the CI green (which is generally 80% of the work considering our CI). We just need to make sure we're able to maintain this in the future.

Alternatively, you can always feel free to create a new module to read in *.frd files as a new python library. I've made several just for reading in files (e.g. https://github.com/pyvista/pyminiply). I'm not opposed to adding in a file reader into PyVista, but the key to software maintenance is testing, documentation, and code quality. I think we're mostly there, we just need a few additional files and fixes.

3rav:
Source of information about the format: https://www.dhondt.de/cgx_2.23.pdf  (11 Result Format, page 144):

```
> 1Ctest !’’1C’’ defines a new calc of name ’’test’’
> 1UDATE 26.01.2000 !’’1U’’ stores user job-informations
> 2C !’’2C’’ starts a block of node coordinates
> -1 1 0.00000E+00 0.00000E+00 0.00000E+00 ! 1. node
> -1 2 0.10000E+01 0.00000E+00 0.00000E+00 ! 2. node ....
> -3 !end of the current block
> 3C !’’3C’’ starts a block of elem definitions
> -1 1 4 0 0!first elem, type of that elem is 4 (he20)
> -2 1 2 3 4 13 14 15 16 5 6 7 8 ..
> -2 12 17 18 19 20 !twenty nodes defining that element
> -1 ...
> -2 ...
> -2 ...
> -3 !end of the current block
> 1PHID 10 !defines a parameter HID value 10
> 100CL101 !’’100C’’ starts a user defined result block
> -4 DISP 3 1 !Attribute Header Record (Dataset)
> -5 D1 1 2 1 0 !Component Def. Record (Entity)
> -5 D2 1 2 2 0
> -5 D3 1 2 3 0
> -1 1 0.00000E+00 1.00000E+00 1.00000E+00 !Nodal Values
> -1 2 1.00000E+00 0.00000E+00 0.00000E+00
> -3 !end of the current block
> 9999 !end of data
```

user27182:
> Source of information about the format: https://www.dhondt.de/cgx_2.23.pdf (11 Result Format, page 144):

Great! Please include this link for reference as part of the enum definition, something like:

``` py
class CGXFormat(IntEnum):
    # Source https://www.dhondt.de/cgx_2.23.pdf (11 Result Format, page 144)
    NODAL_VALUES = -1
    END_OF_BLOCK = -3
    ATTRIBUTE_HEADER = -4 
    COMPONENT_DEFINITION = -5

**Comment by user27182** [comment]
To address the docs build failure the new class should be added to the docs here:
https://github.com/pyvista/pyvista/blob/main/doc/source/api/readers/index.rst

File: pyvista/core/utilities/_frd.py [hunk 16]
```diff
@@ -0,0 +1,475 @@
...
+                try:
+                    etype_val = int(parts[2])
+                    etype = FRDElementType(etype_val)
+                except (ValueError, IndexError):
+                    etype = None
+                    continue
+
+                if etype not in self.CCX_TO_VTK_TYPE:
+                    etype = None
+                    continue
```
user27182, line 249: This seems redundant ? We are checking `FRDElementType(etype_val)` to see if it's defined by the enum, but then checking `in self.CCX_TO_VTK_TYPE` is checking `CCX_TO_VTK_TYPE.keys()`, which is basically the same as checking the enum values.

It's only redundant because they're both setting etype=None and doing nothing else. But, I think we should separate the conversion to `int` to the checking of `etype`. 

If we're parsing an element node that's not convertible to int, then it seems OK to continue. But if it is an int, but it's not defined in the `self.CCX_TO_VTK_TYPE` mapping, then I think this means that the file has a cell type that isn't handled by this reader. In this case, we should emit a warning to let the user know that an unsupported cell type was encountered and will be missing from the result.

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,475 @@
...
+    def _parse_100cl_header(line: str) -> tuple[int, float]:
+        parts = line.split()
+        try:
+            return int(parts[1]), float(parts[2])
+        except (IndexError, ValueError):
+            return -1, 0.0
+
+    def _permute_nodes(self, node_ids: list[int], etype: FRDElementType) -> list[int]:
+        """Reorder node IDs from CalculiX to VTK conventions."""
+        if etype == FRDElementType.HE20 and len(node_ids) == 20:
```
user27182:
```suggestion
        if etype == FRDElementType.HE20 and len(node_ids) == _FRDVTKReader.NODES_PER_ELEM[FRDElementType.HE20]:
```
Should use a lookup here.
Also, what happens if we have `etype == FRDElementType.HE20` but the wrong number of node ids? Should we keep the (invalid) element but emit a warning?

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,475 @@
...
+        return self._output
+
+    # ------------------------------------------------------------------
+    # Parsing helpers
+    # ------------------------------------------------------------------
+
+    @classmethod
+    def _fix_scientific(cls, line: str) -> str:
+        """Insert a space before a bare ``-`` sign adjacent to a number."""
+        return cls._SCIENTIFIC_RE.sub(' -', line)
```
user27182:
```suggestion
    @staticmethod
    def _fix_scientific(line: str) -> str:
        """Insert a space before a bare ``-`` sign adjacent to a number."""
        return _FRDVTKReader._SCIENTIFIC_RE.sub(' -', line)
```

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,475 @@
...
+                node_ids[17],
+                node_ids[18],
+                node_ids[19],
+                node_ids[12],
+                node_ids[13],
+                node_ids[14],
+                node_ids[15],
+            ]
+
+        if etype == FRDElementType.TR3 and len(node_ids) == 3:
```
user27182:
```suggestion
        if etype == FRDElementType.TR3 and len(node_ids) == _FRDVTKReader.NODES_PER_ELEM[FRDElementType.TR3]:
```

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,475 @@
...
+            area = (
+                p[0][0] * (p[1][1] - p[2][1])
+                + p[1][0] * (p[2][1] - p[0][1])
+                + p[2][0] * (p[0][1] - p[1][1])
+            )
+            if area < 0:
+                return [node_ids[0], node_ids[2], node_ids[1]]
+            return node_ids
+
+        if etype in (FRDElementType.QU4, FRDElementType.QU8):
```
user27182: We checked `len(node_ids)` for the other elem types, why not also check here? If the elem has the wrong number of nodes then the following code will probably raise an index error.

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,475 @@
...
+            corners = node_ids[:4]
+            mids = node_ids[4:]
+            p = [self._nodes[n] for n in corners]
+            area = sum(p[i][0] * p[(i + 1) % 4][1] - p[(i + 1) % 4][0] * p[i][1] for i in range(4))
+            if area < 0:
+                corners = [corners[0], corners[3], corners[2], corners[1]]
+                if mids:
+                    mids = [mids[0], mids[3], mids[2], mids[1]]
+            return corners + mids
```
user27182: I see there are a few checks to ensure the area is positive. I'm not sure if I saw tests where the elements were explicitly defined incorrectly to ensure this fix is working.

Is this fix just a kind of bonus check? Is it not possible that the point order is wrong, but wrong in a different way than is assumed by this fix?

Basically, I am not so sure if we should be doing this kind of mutation upon read. Users can already check for negative area/ negative volume using the `validate_mesh` method. If there is a need to fix the orientation this should be done in post-processing by the user I think.

tests/test_frd.py
```diff
@@ -0,0 +1,253 @@
...
+    mesh = reader.read()
+
+    assert 'STRESS' in mesh.point_data
+
+    # Check calculated stress derived fields for mock_frd_file
+    # values: xx=10, yy=20, zz=30, xy=yz=zx=0
+    # vMises for this state is ~17.3205
+    expected_vmises = np.sqrt(300.0)
+
+    np.testing.assert_allclose(mesh.point_data['STRESS_vMises'], expected_vmises)
```
user27182: is `vMises` standard? The Calculix user manual omits the "v" and uses `Mises`. What is the array name when using Calculix?
https://www.dhondt.de/cgx_2.23.pdf

Image

pyvista/core/utilities/reader.py
```diff
@@ -3505,6 +3505,75 @@ def set_active_time_point(self, time_point):
         self.reader.SetTimeStep(time_point)
 
 
+from pyvista.core.utilities.frd import _FRDVTKReader
+
+
+class FRDReader(BaseReader, TimeReader):
+    """Reader for CalculiX FRD ASCII result files (``.frd``)."""
+
```
user27182: Can we add an example to the docstring here? Will need to upload an frd file https://github.com/pyvista/vtk-data, and add it to `downloads.py`.

e.g.

``` py
    >>> import pyvista as pv
    >>> from pathlib import Path
    >>> filename = examples.download_frd(load=False)
    >>> Path(filename).name
    'mesh.frd'
    >>> reader = pv.get_reader(filename)
    >>> mesh = reader.read()
    >>> mesh.plot()
```

Should probably also document:
1. Which cell types are currently supported (I believe this PR only supports a subset)
2. Mention that PE15 nodes are supported, but are degraded to a simple Wedge

Mention that derived arrays are pre-computed and included, e.g. von Mises, principal stress/strain.

pyvista/core/utilities/frd.py
```diff
@@ -0,0 +1,475 @@
...
+    # Compiled regex to fix scientific notation formatting issues
+    _SCIENTIFIC_RE = re.compile(r'(?<![EeDd])-')
+
+    # CalculiX element type -> VTK cell type
+    CCX_TO_VTK_TYPE: ClassVar[dict[FRDElementType, CellType]] = {
+        FRDElementType.HE8: CellType.HEXAHEDRON,
+        FRDElementType.PE6: CellType.WEDGE,
+        FRDElementType.TE4: CellType.TETRA,
+        FRDElementType.HE20: CellType.QUADRATIC_HEXAHEDRON,
+        FRDElementType.PE15: CellType.WEDGE,  # degraded; pe15 not native to base VTK
```
user27182: Can FRDElementType.PE15 be mapped to a quadratic wedge?

https://dev.pyvista.org/api/utilities/_autosummary/pyvista.celltype#pyvista.CellType.QUADRATIC_WEDGE

Image

Seems like the node ids will need to be mapped accordingly.

pyvista/core/utilities/_frd.py
```diff
@@ -0,0 +1,475 @@
...
+
+    @staticmethod
+    def _parse_100cl_header(line: str) -> tuple[int, float]:
+        parts = line.split()
+        try:
+            return int(parts[1]), float(parts[2])
+        except (IndexError, ValueError):
+            return -1, 0.0
+
+    def _permute_nodes(self, node_ids: list[int], etype: FRDElementType) -> list[int]:
```
user27182: I only see special handling for some element types:
- FRDElementType.QU4
- FRDElementType.QU8
- FRDElementType.HE20
- FRDElementType.TR3

And there is no special handling for other supported types:
- FRDElementType.HE8
- FRDElementType.PE6
- FRDElementType.TE4
- FRDElementType.HE20
- FRDElementType.PE15
- FRDElementType.TE10
- FRDElementType.TR6
- FRDElementType.BE2
- FRDElementType.BE3

I haven;t reviewed the node ordering for all of these, but based on my findings for `PE15` (Quadratic Wedge), it seems likely that the point ordering can differ between the FRD cell and the VTK cell definition.

And so, I think it's extremely important to ensure every single supported cell type has a test file to ensure the node order is properly handled. You can assert `mesh.volume` > 0 in each test after reading the file.

If this is too much work, then we should drop support for untested cell types. IMO it's better not to read the cell at all than to potentially read it incorrectly.

pyvista/examples/downloads.py
```diff
@@ -2732,6 +2732,35 @@ def download_brain(load=True):  # noqa: FBT002
...
+    >>> dataset.plot()
+
+    """
+    filename = _download_file('mesh.frd')
+    if not load:
+        return filename
+
+    import pyvista as pv  # noqa: PLC0415
+
+    return pv.read(filename)
```
user27182: You'll probably want something more like this:

```suggestion
    .. seealso::

        :ref:`Frd Dataset <frd_dataset>`
            See this dataset in the Dataset Gallery for more info.

    """
    return _download_dataset(_dataset_frd, load=load)


_dataset_frd = _SingleFileDownloadableDatasetLoader('mesh.frd',)
```

user27182:
I checked this out locally and ran:

``` py
>>> import pyvista as pv 
>>> mesh = pv.read(pv.examples.download_file('mesh.frd'))
>>> mesh.distinct_cell_types
{<CellType.LINE: 3>,
 <CellType.TRIANGLE: 5>,
 <CellType.QUAD: 9>,
 <CellType.TETRA: 10>,
 <CellType.HEXAHEDRON: 12>,
 <CellType.WEDGE: 13>,
 <CellType.QUADRATIC_EDGE: 21>,
 <CellType.QUADRATIC_TRIANGLE: 22>,
 <CellType.QUADRATIC_QUAD: 23>,
 <CellType.QUADRATIC_TETRA: 24>,
 <CellType.QUADRATIC_HEXAHEDRON: 25>}

>>> mesh.array_names
['original_node_ids']
```

Looks like a variety of cells are included which is great for testing. Though the computed arrays are not there, presumably because no simulation ran, and hence there are no STRESS and STRAIN arrays?

user27182:
Also, looks like there is currently an issue with WEDGE?
``` py
>>> import pyvista as pv
>>> mesh = pv.read(pv.examples.download_file('mesh.frd')).validate_mesh(action='error')

pyvista.core.errors.InvalidMeshError: UnstructuredGrid mesh is not valid:
 - Mesh has 2 WEDGE cells with wrong number of points. Invalid cell ids: [11, 12]
```

We'll want to add an assertion for this e.g.

``` py
mesh = examples.download_frd()
assert mesh.validate_mesh().is_valid
```

3rav:
File with results (DISP, STRESS, TEMP, STRAIN, FORC, ERROR)

[mesh.zip](https://github.com/user-attachments/files/25725857/mesh.zip)

3rav:
Generated mesh files for (wedge) PE6 and PE15:

```
*NODE, NSET=NC3D6
     127,1.400000000000e+02,0.000000000000e+00,0.000000000000e+00
     128,1.500000000000e+02,0.000000000000e+00,0.000000000000e+00
     129,1.400000000000e+02,1.000000000000e+01,0.000000000000e+00
     130,1.500000000000e+02,1.000000000000e+01,0.000000000000e+00
     242,1.400000000000e+02,0.000000000000e+00,1.000000000000e+01
     243,1.500000000000e+02,0.000000000000e+00,1.000000000000e+01
     244,1.400000000000e+02,1.000000000000e+01,1.000000000000e+01
     245,1.500000000000e+02,1.000000000000e+01,1.000000000000e+01
*ELEMENT, TYPE=C3D6, ELSET=EC3D6
    82,   127,   128,   129,   242,   243,   244
    83,   129,   128,   130,   244,   243,   245
```

```
*NODE, NSET=NC3D15
     131,2.000000000000e+02,0.000000000000e+00,0.000000000000e+00
     132,2.100000000000e+02,0.000000000000e+00,0.000000000000e+00
     133,2.000000000000e+02,1.000000000000e+01,0.000000000000e+00
     134,2.050000000000e+02,0.000000000000e+00,0.000000000000e+00
     135,2.050000000000e+02,5.000000000000e+00,0.000000000000e+00
     136,2.000000000000e+02,5.000000000000e+00,0.000000000000e+00
     137,2.100000000000e+02,1.000000000000e+01,0.000000000000e+00
     138,2.100000000000e+02,5.000000000000e+00,0.000000000000e+00
     139,2.050000000000e+02,1.000000000000e+01,0.000000000000e+00
     253,2.000000000000e+02,0.000000000000e+00,1.000000000000e+01
     254,2.100000000000e+02,0.000000000000e+00,1.000000000000e+01
     255,2.000000000000e+02,1.000000000000e+01,1.000000000000e+01
     256,2.050000000000e+02,0.000000000000e+00,1.000000000000e+01
     257,2.050000000000e+02,5.000000000000e+00,1.000000000000e+01
     258,2.000000000000e+02,5.000000000000e+00,1.000000000000e+01
     259,2.100000000000e+02,1.000000000000e+01,1.000000000000e+01
     260,2.100000000000e+02,5.000000000000e+00,1.000000000000e+01
     261,2.050000000000e+02,1.000000000000e+01,1.000000000000e+01
     262,2.000000000000e+02,0.000000000000e+00,5.000000000000e+00
     263,2.100000000000e+02,0.000000000000e+00,5.000000000000e+00
     264,2.000000000000e+02,1.000000000000e+01,5.000000000000e+00
     268,2.100000000000e+02,1.000000000000e+01,5.000000000000e+00
*ELEMENT, TYPE=C3D15, ELSET=EC3D15
    84,   131,   132,   133,   253,   254,   255,   134,   135,   136,
         256,   257,   258,   262,   263,   264
    85,   133,   132,   137,   255,   254,   259,   135,   138,   139,
         257,   260,   261,   264,   263,   268
```

hcpy_3

user27182:
> File with results (DISP, STRESS, TEMP, STRAIN, FORC, ERROR)
> 
> [mesh.zip](https://github.com/user-attachments/files/25725857/mesh.zip)

looks like this

```
pyvista plot mesh.frd
```

image

but still seeing issues:
```
pyvista validate mesh.frd
```
```
UnstructuredGrid mesh 'mesh.frd' is not valid:
 ▪ Mesh has 2 WEDGE cells with wrong number of points. Invalid cell ids: [11, 12]
```

3rav:
Change from (in frd.py):
`FRDElementType.PE15: CellType.WEDGE,  # degraded; pe15 not native to base VTK`

to?:

`FRDElementType.PE15: CellType.QUADRATIC_WEDGE,`

user27182:
> Change from (in frd.py): `FRDElementType.PE15: CellType.WEDGE, # degraded; pe15 not native to base VTK`
> 
> to?:
> 
> `FRDElementType.PE15: CellType.QUADRATIC_WEDGE,`

Fixed in https://github.com/pyvista/pyvista/pull/8255/commits/90f2df42378301df4f50d179a6734afb507e941c

we still have this error:
```
UnstructuredGrid mesh 'mesh.frd' is not valid:
 ▪ Mesh has 2 QUADRATIC_WEDGE cells with negative volume. Invalid cell ids: [11, 12]
```
but this is a VTK bug https://gitlab.kitware.com/vtk/vtk/-/issues/19639

user27182:
Looking pretty good, the docs build will fail, images need to be updated. Otherwise might just need a few more tests to address some missing test coverage.

user27182:
@3rav are you able to add a test to trigger the warning about wrong number of points? Otherwise I think we're good for test coverage.

user27182:
Can we keep the match removed by 0aacc43124d7be81e3836bd0d08cd3ac9c514277? so we can ensure the message is correct

tests/test_frd.py
```diff
@@ -0,0 +1,425 @@
...
+    assert pv.CellType.HEXAHEDRON in low_level_reader._has_wrong_number_of_points
+
+    # 2. Test that Update() raises the warning based on the internal state
+    reader = pv.FRDReader(empty_frd_file)
+    reader.reader._has_wrong_number_of_points.add(pv.CellType.HEXAHEDRON)
+
+    with pytest.warns(
+        pv.InvalidMeshWarning, match='Cell types with wrong number of points detected'
+    ):
+        reader.read()
```
user27182: While this may be useful as a test, I'd prefer to test using an actual (invalid) test file. Is it not possible to re-use one of the test file fixtures? And simply add or remove a node from one of the valid elements to make it invalid? Maybe just modify `comprehensive_frd_file(tmp_path)` to include the invalid element with too many points and/or too few?

pyvista/core/utilities/_frd.py
```diff
@@ -0,0 +1,486 @@
...
+                    etype = None
+                    continue
+
+                needed = NODES_PER_ELEM[etype]
+                vtk_type = CCX_TO_VTK_TYPE[etype]
+                node_ids = []
+
+            elif s.startswith(elem_faces) and etype is not None and vtk_type is not None:
+                node_ids.extend(int(x) for x in s.split()[1:])
+                if len(node_ids) >= needed:
```
user27182: With the new invalid elements, this logic catches the triangle element with _too many_ points, since we have`len(node_ids) >= needed`. But the wedge element with _too few_ points is skipped completely since it has `len(node_ids) < needed`. And if we remove this line, we get index error when parsing, which makes sense, since there aren't enough points to define the new re-ordered element.

Not sure how to handle this? Should we still parse it, but avoid index error when re-ordering the points? Or maybe add an `else` branch here, and add the invalid etype to the same `_has_wrong_number_of_points` set? But it's different since this element is now skipped completely, wherase with too many points we still keep it...

pyvista/core/utilities/_frd.py
```diff
@@ -0,0 +1,486 @@
...
+            warn_external(msg, InvalidMeshWarning)
+        if elements := self._has_unsupported_element:
+            msg = (
+                f'Unknown element type code(s) encountered {elements}. These elements are skipped.'
+            )
+            warn_external(msg, InvalidMeshWarning)
+
+        self._time_steps = sorted(self._results_by_step.keys())
+
+    def GetOutput(self) -> UnstructuredGrid:  # noqa: N802
```
MatthewFlamm: It has been awhile since I've looked at this part of vtk, but our current pattern doesn't seem to use `GetOutput`

https://github.com/pyvista/pyvista/blob/574b7fe6b46e3cf796c2cab221beda25d32ed505/pyvista/core/utilities/reader.py#L136-L145

pyvista/core/utilities/_frd.py
```diff
@@ -0,0 +1,486 @@
...
+        self._time_steps: list[float] = []
+        self._active_time_point: int = 0
+        self._output_time: object = object()
+        self._output: UnstructuredGrid | None = None
+
+    def SetFileName(self, filename: str) -> None:  # noqa: N802
+        self._filename = filename
+
+    def UpdateInformation(self) -> None:  # noqa: N802
+        pass
```
MatthewFlamm: Can we set the available time steps here?  If we don't, then you user will only be able to inspect time points _after_ reading a mesh.

pyvista/core/utilities/reader.py
```diff
@@ -3505,6 +3505,99 @@ def set_active_time_point(self, time_point):
...
+    """
+
+    _class_reader = _FRDVTKReader
+
+    def __init__(self, path: str) -> None:
+        """Initialize the reader."""
+        super().__init__(path)
+        self._reader: _FRDVTKReader = self._class_reader()
+        self._reader.SetFileName(self.path)
+        self._reader.Update()
```
MatthewFlamm: Do we need to do this still if logic is moved into `UpdateInformation` in `_FRDVTKReader`?  Can we get rid of this entire `__init__` method?

MatthewFlamm (commented):
I am not that familiar with this filetype.  This is looking close from a structural perspective, but it needs reorganizing.  Overall, the structure should look like this:

1) All information about the data should be parsed in `UpdateInformation`.  For example, which time points are available.  In the case where the point and cell data can be selected on and off, this might also happen in that method.
2) The mesh should be constructed in the `Update` call and set to `self._data_object`.
3) The `BaseVTKReader` already handles the output using the above structure.

tests/test_frd.py
```diff
@@ -0,0 +1,409 @@
...
+@pytest.mark.parametrize('generic_element_frd', ['BE2', 'BE3'], indirect=True)
+def test_frd_1d_element_lengths(generic_element_frd):
+    filepath, elem_name = generic_element_frd
+    mesh = pv.FRDReader(filepath).read()
+
+    # Calculate cell length
+    length = mesh.compute_cell_sizes().cell_data['Length'][0]
+    assert length > 0.0, (
+        f'Element {elem_name} generated non-positive length ({length}). Bad node ordering!'
+    )
```
user27182: The cell types for these tests are hard-coded, which is an issue if future cell types are added/supported, since the tests won't pick these new elements up automatically.

I think these tests can be condensed into a single test instead, and then we can parametrize over _all_ elements in 'generic_element_frd'.
For the 1d/2d/3d cases, just add logic to assert length or area or volume based on the cell dimension. You can map from frd element to vtk element using the CCX_TO_VTK_TYPE dict, and then check dimension by checking celltype membership in either _CELL_TYPES_1D,  _CELL_TYPES_2D, or _CELL_TYPES_3D defined in `celltype.py`.

MatthewFlamm:
From a structural point of view, this is looking good.  Thank you for addressing my comments.  I unfortunately do not have time to look at detail at the file format and mesh building part of this PR.

tests/test_frd.py
```diff
@@ -0,0 +1,413 @@
...
+        msg = (
+            'VTK bug with negative volume for quadratic wedge '
+            'https://gitlab.kitware.com/vtk/vtk/-/issues/19639'
+        )
+        pytest.xfail(msg)
+
+    # 3. Oblicz rozmiary komórek
+    sizes = mesh.compute_cell_sizes().cell_data
+
+    # 4. Sprawdź wymiar dynamicznie i zrób odpowiednią asercję
```
user27182: Can you translate these to english?

pyvista/core/utilities/_frd.py
akaszynski: Honestly, if the file's been written incorrectly we just should report it and attempt to read the remaining elements. While we should be able to handle improperly written files to some degree, the responsibility to adhere to the standard CalculiX FRD falls primarily with the writer.

user27182 (changes requested):
I think I can fully separate the VTK reader class from the FRD file parsing. This should eliminate the circular import issues, so that `reader.py` can import the parser from `_frd.py` without issue, and we only define one `FRDReader` class in `reader.py` (i.e. no need to define an intermediate  `_FRDVTKReader` class inside of `_frd.py`.)

I will work on this and push changes shortly.

pyvista/core/utilities/_frd.py
user27182: I updated the reader to warn about too few points.

But this now results in invalid cells being flagged in the new `mesh.frd` file, e.g.
``` py
from pyvista import examples
examples.download_frd()
```
``` 
InvalidMeshWarning: 1 cell with too few points detected. These elements are skipped:
  line 30, element type 4 (HE20), num nodes 15 (expected 20)
```

@3rav do you know why this is happening? Is there is an issue with the file? Or with the reader?

user27182: These are the problem lines
https://github.com/pyvista/vtk-data/blob/126ee906165a8b2f984f08259a4a2d32cf3b3409/Data/mesh.frd#L227-L237

```
 -1        84    5    0    1
 -2       131       132       133       253       254       255       134       135       136       262
 -2       263       264       256       257       258
 -1        85    5    0    1
 -2       133       132       137       255       254       259       135       138       139       264
 -2       263       268       257       260       261
 -1        86    1    0    1
 -2       140       141       142       143       144       145       146       147
 -1        87    4    0    1
 -2       148       149       150       151       152       153       154       155       156       157
 -2       158       159       160       161       162       163       164       165       166       167
```
The issue is that we have _multiple_ `-2` lines, and only the first line is read, so the other node ids are missing and hence too few points.

Is this a valid file? Or should `-2` only span a single line, e.g.:

``` diff
 -1        84    5    0    1
- -2       131       132       133       253       254       255       134       135       136       262
- -2       263       264       256       257       258
+ -2       131       132       133       253       254       255       134       135       136       262       263       264       256       257       258
 -1        85    5    0    1
- -2       133       132       137       255       254       259       135       138       139       264
- -2       263       268       257       260       261
+ -2       133       132       137       255       254       259       135       138       139       264       263       268       257       260       261
 -1        86    1    0    1
 -2       140       141       142       143       144       145       146       147
 -1        87    4    0    1
- -2       148       149       150       151       152       153       154       155       156       157
- -2       158       159       160       161       162       163       164       165       166       167
+ -2       148       149       150       151       152       153       154       155       156       157       158       159       160       161       162       163       164       165       166       167
```

When apply this diff, the warnings go away.
But, confusingly, the loaded mesh has 61 cells both times. I would have expected the cells to be 58 before (3 invalid cells skipps), and 61 after the fix.

user27182:
> I think I can fully separate the VTK reader class from the FRD file parsing. This should eliminate the circular import issues, so that `reader.py` can import the parser from `_frd.py` without issue, and we only define one `FRDReader` class in `reader.py` (i.e. no need to define an intermediate `_FRDVTKReader` class inside of `_frd.py`.)
> 
> I will work on this and push changes shortly.



> I think I can fully separate the VTK reader class from the FRD file parsing. This should eliminate the circular import issues, so that `reader.py` can import the parser from `_frd.py` without issue, and we only define one `FRDReader` class in `reader.py` (i.e. no need to define an intermediate `_FRDVTKReader` class inside of `_frd.py`.)
> 
> I will work on this and push changes shortly.

This change has been made. Still need to resolve https://github.com/pyvista/pyvista/pull/8255#discussion_r2886235751

pyvista/core/utilities/_frd.py
user27182: @3rav , I have assumed that the file is correct, which means we should expect element definitions to span multiple lines. I have updated the reader accordingly. It now parses `mesh.frd` without warning.

Is it expected that this mesh should have exactly 61 cells?

Please confirm that the reader is now reading this file correctly.

user27182 (approved):

pyvista/core/utilities/_frd.py
3rav: CalculiX ccx (solver) saves the results in this format, divides the nodes of the HE20 element into 2 lines of 10 nodes each (PE15 first 10 secod 5 nodes):

Example division 1:
PE_HE
image

image


Example division 4:
hcpy_1
image

Example from the official repository: https://github.com/Dhondtguido/CalculiX/blob/master/test/achtel29.frd.ref

pyvista/core/utilities/reader.py
```diff
@@ -3505,6 +3510,128 @@ def set_active_time_point(self, time_point):
...
+            raise IndexError(msg)
+        self.reader._active_time_point = time_point
+
+    def set_active_time_value(self, time_value: float) -> None:
+        """Set the active time value."""
+        steps = self.reader._time_steps
+        if not steps:
+            msg = 'No time steps found in the FRD file.'
+            raise RuntimeError(msg)
+        idx = int(np.argmin(np.abs(np.array(steps) - time_value)))
```
MatthewFlamm: This is likely my last comment.  Sorry for interleaving comments over multiple days here.  In vtk, the call to set the time point is by the value.  I'm not actually sure what this means in practice.  I had assumed it was by exact value, but I have not tested it.  It would be good in the future to mirror vtk behavior, and/or provide some user option like `exact_value=True`.

https://github.com/pyvista/pyvista/blob/7bdd50d6940b4378207cff311cde324e86b79848/pyvista/core/utilities/reader.py#L803-L810

In another instance, we are only allowing exact values for time.

https://github.com/pyvista/pyvista/blob/7bdd50d6940b4378207cff311cde324e86b79848/pyvista/core/utilities/reader.py#L3763-L3768

Unless we know how vtk works, I would prefer to keep consistency with the in-built reader to require an exact time value match for the time being.  Then we could align on what a fuzzy match for time should look like for all readers together.  Otherwise, we will build up a haphazard set of logic.

pyvista/core/utilities/_frd.py
user27182: Thanks for confirming! Indeed the reader now handles this correctly and will read from multiple lines as needed for the node ids.

pyvista/core/utilities/reader.py
3rav: I hope that's what you meant https://github.com/pyvista/pyvista/pull/8255/commits/0e3c7e497e221cedea83e2958224af492626dcb1

MatthewFlamm: yes thanks!

user27182:
> I have only reviewed the structure of the reader code, but I haven't reviewed the logic of parsing and building the mesh.

I did the opposite, and reviewed the file parser.

akaszynski:
> > I have only reviewed the structure of the reader code, but I haven't reviewed the logic of parsing and building the mesh.
> 
> I did the opposite, and reviewed the file parser.

Guess that leaves me as the reviewer of the reviewers.

Thanks all!

MatthewFlamm:
Great work @3rav

3rav:
Thanks for the help and your valuable feedback @user27182 @MatthewFlamm @akaszynski . I also really appreciate your patience throughout the process!

PR #8259, opened 2026-01-23
Deprecate `extract_geometry` in favor of `extract_surface`

### LINKED ISSUE #6620: Unify `extract_surface` and `extract_geometry` into a single filter

These two filters perform very similar things. I think having two functions do similar things makes it confusing, especially when there is little or no documentation in the PyVista docs highlighting the differences. When should we use one vs. the other?

I propose deprecating one of the two so that there is only one cell-boundary extraction function in the `PyVista` library. But which one should be deprecated?

### VTK Filter:`vtkGeometryFilter` vs `vtkDataSetSurfaceFilter`

Internally, I think the default vtk filter to use should be `vtkGeometryFilter`. See rationale below.

On the one hand, the docs for [`extract_geometry`](https://dev.pyvista.org/api/core/_autosummary/pyvista.datasetfilters.extract_geometry#pyvista.DataSetFilters.extract_geometry), which uses [`vtkGeometryFilter`](https://vtk.org/doc/nightly/html/classvtkGeometryFilter.html#details), notes that it may be less performant than [`extract_surface`](https://dev.pyvista.org/api/core/_autosummary/pyvista.datasetfilters.extract_surface#pyvista.DataSetFilters.extract_surface), which uses [`vtkDataSetSurfaceFilter`](https://vtk.org/doc/nightly/html/classvtkDataSetSurfaceFilter.html#details). This may have been true in the past, but the VTK post linked below suggests this is no longer the case, and shows that `vtkGeometryFilter` is much more performant. `vtkDataSetSurfaceFilter` is even labeled as _evil_ here.
https://discourse.vtk.org/t/vtkgeometryfilter-and-its-evil-spawn-vtkdatasetsurfacefilter/4083

The `vtkGeometryFilter` docs also highlight some key differences between the methods, but these differences only seem to suggest that `vtkGeometryFilter` does everything `vtkDataSetSurfaceFilter` and more.

### PyVista Filter Name

I think we should keep the name `extract_surface` and deprecate `extract_geometry`. See rationale below.

Intuitively, I think `extract_surface` is a better choice. The name `extract_geometry` leaves me wondering... "which geometry, exactly?" Plus there is an unrelated vtk filter named `vtkExtractGeometry` which has nothing to do with PyVista's `extract_geometry`, so it may be best to avoid using this name.

### Deprecation and Backwards compatibility

Since the two vtk methods can produce different outputs in some cases, it would be a breaking change to fully deprecate one or the other. Plus, based on the above, I think we should use the `extract_surface` filter name but use `vtkGeometryFilter` as the default method under the hood. This naturally leads to a solution where `extract_geometry` and `extract_surface` are combined into a single filter: `extract_surface`.

This could be done as follows:

1. Add a new `method` parameter to `extract_surface` which is `None` by default, but is optionally set to either `geometry` to use `vtkGeometryFilter` or `surface` to use `vtkDataSetSurfaceFilter`. 
2. If `method` is left as `None`, `surface` will be used as the default (for now), but we issue a `PyVistaFuture` warning stating that the default will be changed to use `geometry` in a future version.
3. Long term, once `geometry` is set as the default, we either leave `surface` as a permanent option (for legacy reasons and avoid breaking changes internally), or explore deprecating/removing it altogether. 

### Links to source code.

_No response_

### Pseudocode or Screenshots

_No response_

### LINKED ISSUE #8258: `extract_geometry` extent is buggy

This output is not correct:

``` py
import pyvista as pv
surf = pv.examples.load_uniform().extract_geometry(extent=(0, 5, 0, 100, 0, 100))
surf.plot()
```

image

Better to remove this, and use `clip_box` separately instead.

```py
import pyvista as pv
surf = pv.examples.load_uniform().extract_geometry().clip_box(bounds=(0, 5, 0, 100, 0, 100), invert=False)
surf.plot()
```

Image

-

### System Information

```shell
-
```

### Screenshots

_No response_

user27182:
### Overview

Resolve #6620. `vtkGeometryFilter` is now used internally for all cases where possible except for non-linear cells.

In addition to `vtkDataSetSurfaceFilter` being "evil" (see #6620), VTK 9.6 is also using `vtkGeometryFilter` now for the rendering pipeline and "should be a drop in replacement": https://discourse.vtk.org/t/use-of-vtkgeometryfilter-in-vtkdatasetmapper/15740

Resolve #8258 too by excluding the `extent` keyword from `extract_surface`, which is buggy anyway and isn't an option with `vtkDataSetSurfaceFilter` anyway.


pyvista/core/filters/data_object.py
```diff
@@ -3259,6 +3259,198 @@ def extract_all_edges(  # type: ignore[misc]
             output.clear_data()
         return output
 
+    @_deprecate_positional_args
+    def extract_surface(  # type: ignore[misc]  # noqa: PLR0917
+        self: DataSet | MultiBlock,
+        pass_pointid: bool = True,  # noqa: FBT001, FBT002
+        pass_cellid: bool = True,  # noqa: FBT001, FBT002
+        nonlinear_subdivision: int = 1,
+        algorithm: Literal['geometry', 'dataset_surface'] | None = None,
```
user27182: let me know if someone can think of a better keyword and options to use. Some ideas:
- Use `algorithm: 'geometry' | 'dataset_surface'`. This is a bit technical and very literal, but is also explicit
- Use `method: 'default' | 'legacy'`. This is a bit better and aligns with the deprecation intent, where DataSetSurfaceFilter  is the "legacy" alg being deprecated, but `legacy` also implies that the new alg is a superset of the old alg, but DataSetSurfaceFilter can do things GeometryFilter can't, i.e. non-linear subdivision, so can't be fully replaced
- Use `mode: 'linear' | 'nonlinear'`. This makes sense in terms of the internal logic, where 'nonlinear' is delegated to DataSetSurfaceFilter for non-linear subdivision. But _both_ algs can handle non-linear cells, so this doesn't quite fit.

examples/99-advanced/mesh_validation.py
```diff
@@ -187,7 +187,7 @@
 
 # %%
 # This time, the point is properly processed by the filter and is retained.
-poly = grid.extract_geometry()
+poly = grid.extract_surface(algorithm='geometry')
```
user27182: > ```poly = grid.extract_surface(algorithm='geometry')```

I find this usage with an `algorithm` keyword a bit ugly, and would prefer not to expose the internal algorithm to users. This usage is only temporarily needed during the transition period.

The reason this keyword is used at all is to avoid a breaking change, since we're trying to (1) keep the `extract_surface` name, but use `vtkGeometryFilter` by default instead of `vtkDataSetSurfaceFilter` internally. The two algs generate the same surface, but sometimes with different point order, and I found even the number of points can differ slightly depending on the input (I believe these are duplicate points, not new ones).

But, the problem is that, by introducing a new keyword, we are now committing to supporting backwards-compatibility by enshrining both algorithms as plausible options. Ideally, there should be no options at all, and the implementation should simply choose the best alg automatically without user input.

I'm thinking that a clean breaking change in this case may be an acceptable alternative. The break is minimal (different point ordering), and will avoid any confusion for users trying to choose `algorithm` now, only for us to then deprecate `algorithm` in the future.

pyvista/core/filters/data_object.py
akaszynski: Top option: (`'geometry' | 'dataset_surface'`). Most people will use defaults, but those that do want to understand what's going on "under the hood" will be familiar (or soon will be familiar) with VTK. While we do abstract out VTK, we don't want to obfuscate it.

examples/99-advanced/mesh_validation.py
akaszynski: Alternatively, we can just use `**kwargs`. That way, under the hood we can change it as needed and `**kwargs` (in general) prescribes a mutable API.

user27182: Yes hiding it behind `**kwargs` would be better. While generalizing the filter though as `DataObjectFilters.extract_surface`, I found another issue: performance. `vtkCompositeGeometryFilter` (used in `MultiBlock.extract_geometry()`) doesn't have any pass-ids flags, but `extract_surface` includes original IDs by default. We can still provide original ids for MultiBlock, but it means we have to use `generic_filter` and manually iterate over each block to extract the surface (similar to `MultiBlock.as_polydata_blocks()`). But, I found that using `vtkCompositeGeometryFilter` directly was approx 20x faster than using `generic_filter`.

And so for performance, we should have `pass_cell_ids` and `pass_point_ids` as False by default. But this means we need to then do a lot of simultaneous deprecations: change the default internal alg _and_ change the default value of `pass_cell_ids`. I'm thinking the deprecation period will be particularly ugly and confusing for users.

Maybe it's better to keep these two filters separate after all? And just add better docs to guide users?

We can still generalize the two filters, i.e. have `DataObjectFilters.extract_surface` and `DataObjectFilters.extract_geometry`

akaszynski: I still don't like having both `extract_surface` and `extract_geometry`, but I know from experience that I've investaged using both and found that in some cases `extract_surface` has better performance (for my use cases on `UnstructuredGrid` datasets). Maybe disabling passing cell IDs would help, though I need point IDs.

We should ideally have one method to extract the surface and expose (as a kwarg) the option to use the legacy algorithm so users can revert. We can even use the legacy by default and deprecate `extract_geometry` for this minor release. We can also set passing point and cell IDs to None and warn the user that if unset, the default will become `False`. This worked great for the `inplace` transition.

user27182: I agree long term we should really only have one method for this. And I really do think `extract_surface` is the name we should use.

For the internal alg performance, the VTK posts I've read strongly suggest that `vtkGeometryFilter` is more performant, with recent hash-based internals to speed things up. Maybe we should benchmark this ourselves, possibly with some of the datasets you use, using the "raw" vtk algs (not wrapped by PyVista). Note that both algs have passing ids disabled by default. We should then make our own determination of what the default should be based on that.

Making things more complicated though, is the fact that `vtkCompositeGeometryFilter` actually uses `vtkDataSetSurfaceFilter` on each of the blocks internally. So if we do make `vtkGeometryFilter` the default alg, it would only be for `DataSet` types, since the default for MultiBlock should be to use `vtkCompositeGeometryFilter`. So maybe for consistency sake, there's an argument to be make that `vtkDataSetSurfaceFilter` should actually be the default. 🤷‍♂️


In terms of disabling passing ids by default, we can actually rename the kwargs here too:

`pass_cellid` -> `pass_cell_ids`
`pass_pointid` -> `pass_point_ids`

The new terms match other PyVista filters (e.g. extract_cells). The defaults would still be `True` for a few releases, but would eventually be `False`.

akaszynski: > Maybe we should benchmark this ourselves, possibly with some of the datasets you use, using the "raw" vtk algs (not wrapped by PyVista).

Let me run a few benchmarks.

akaszynski: Adhoc benchmarks show the two filters are about equivalent in terms of performance. Tested several unstructured grids and got similar results as to this one. I think we're good to go with the refactor.

```py
import timeit

import vtk
import pyvista as pv


grid = fem.grid # or fem.quadgrid for quadratic grid
# from 
# FEM (0x7fd4af645480)
#   N Cells:    426012
#   N Points:   1541264
#   X Bounds:   -7.416e-01, 3.712e+00
#   Y Bounds:   -1.266e+00, 1.344e+00
#   Z Bounds:   -4.948e-04, 7.508e+00

pass_ids = False


def ds_filter():
    dss = vtk.vtkDataSetSurfaceFilter()
    dss.SetInputData(grid.copy())
    dss.SetPassThroughCellIds(pass_ids)
    dss.SetPassThroughPointIds(pass_ids)
    dss.Update()
    return dss.GetOutput()


def geo_filter():
    gf = vtk.vtkGeometryFilter()
    gf.SetInputData(grid.copy())
    gf.SetPassThroughCellIds(pass_ids)
    gf.SetPassThroughPointIds(pass_ids)
    gf.Update()
    return gf.GetOutput()


ds_mesh = pv.wrap(ds_filter())
geo_mesh = pv.wrap(geo_filter())
assert ds_mesh.n_points == geo_mesh.n_points

# warmup
n = 20
geo_time = timeit.timeit(geo_filter, number=n) / n
ds_time = timeit.timeit(ds_filter, number=n) / n

geo_time = timeit.timeit(geo_filter, number=n) / n
ds_time = timeit.timeit(ds_filter, number=n) / n

print(f"ds_filter: {ds_time:.6f} s")
print(f"geo_filter: {geo_time:.6f} s")
```
Results:
```
# quadratic cells

# pass_ids=False
# ds_filter: 0.189164 s
# geo_filter: 0.188722 s

# pass_ids=True
# ds_filter: 0.196480 s
# geo_filter: 0.198679 s

# linear cells
# pass_ids=False
# ds_filter: 0.069665 s
# geo_filter: 0.067970 s

# pass_ids=True
# ds_filter: 0.071791 s
# geo_filter: 0.066970 s
```

user27182: Okay, so what do we want for the default alg internally? Keeping it as `vtkDataSetSurfaceFilter` is much easier deprecation-wise, and is more consistent in terms of how MultiBlock would be filtered by default. And it also properly handles the non-linear subdivision cases.


And do we want new `pass_point_ids` and `pass_cell_ids` keywords, with the intention of making the default `False` in the future (for compatibility with `vtkCompositeGeometryFilter`)?

akaszynski: Yes for both. Keep using `vtkDataSetSurfaceFilter`, but allow the user to change the alg. Maybe when the dust settles we can change the default alg (or not).

user27182: Here are two compelling reason to use `vtkGeometryFilter` by default:

(1) It preserves input geometry:

``` py
import pyvista as pv
poly = pv.Sphere()
ugrid = poly.cast_to_unstructured_grid()
poly == ugrid.extract_geometry()  # True
poly == ugrid.extract_surface(pass_pointid=False, pass_cellid=False)  # False
```

(2)  `vtkDataSetSurfaceFilter` generates open surfaces in cases where we'd expect a closed surface.

 
``` py
import pyvista as pv
import vtk
def is_closed(surface):
    return bool(vtk.vtkSelectEnclosedPoints.IsSurfaceClosed(surface))

mesh = pv.ImageData(dimensions=(10,10,10))
is_closed(mesh.extract_geometry())  # True
is_closed(mesh.extract_surface())  # False
```

EDIT: Fix first example

akaszynski: Didn't realize `vtkDataSetSurfaceFilter` generated open surfaces. In that case. I'm fine with defaulting to `vtkGeometryFilter`, especially as benchmarks indicate it's parity. As long as we give users the option to use the other alg.

user27182: Yes I also think it weakens the case for changing the default `pass_pointid` and `pass_cellid` to `False`. A key motivating factor was to provide a default fast branch for `MultiBlock` to use `vtkCompositeGeometryFilter`. But since that uses `vtkDataSetSurfaceFilter` internally, and hence may generate open surfaces, we should make the default path to use `MutliBlock.generic_filter` instead. It's likely slower, but the trade-off would be that it's self-consistent (all types use `vtkGeometryFilter` by default), and "more correct". We can make a note about the more performant branch and keywords to use in case users would rather opt-in to that behavior.

So, I'm fine with leaving `pass_pointid` and `pass_cellid`. I still prefer the names like `pass_cell_ids` instead of `pass_cellid`, but we can also change that another time.

akaszynski: > So, I'm fine with leaving `pass_pointid` and `pass_cellid`. I still prefer the names like `pass_cell_ids` instead of `pass_cellid`, but we can also change that another time.

That migth be one deprecation too many, but I'm also fine with pulling off the bandaid now just as long as we allow users to transition. We're pretty good about backwards compatability and Python libraries are known for changing things (looking at you `np.in1d`).

user27182: I think we can skip the kwarg rename for now. Might be able to group it with a "bulk rename" PR in the future, where a bunch of inconsistencies like this are all addressed at once.

akaszynski:
Let's finalize and merge this PR and make a release. Any others @user27182?

user27182:
> I'm aware that this is a draft, but I'm satisfied about the refactor as-is.

Thanks. Still draft because I wanted to double-check the default kwargs for some function calls, e.g. make sure the point/cell ids are passed (or not) for cases where the alg used may have changed.

And I want to include some documentation comparing the two algs and highlighting the different outputs and use-cases.

> I think they've enabled multi-threaded geometry extraction, but it's unclear if the Python wheels are built with TBB as I don't see multiple threads being used with the geometry filter. I'll check again with the latest release. 

I don't think VTK is built with multi-threading by default, but it would be helpful to confirm this (and document it).

user27182:
> Let's finalize and merge this PR and make a release. Any others @user27182?

No other pending PRs for me, no.

akaszynski:
> I don't think VTK is built with multi-threading by default, but it would be helpful to confirm this (and document it).

https://discourse.vtk.org/t/vtk-9-6-0-release-planning/15760/9

akaszynski:
Update on `vtkDataSetSurfaceFilter` vs `vtkGeometryFilter`. When building with ``VTK_SMP_IMPLEMENTATION_TYPE=TBB``, I'm seeing a 4x improvement when using `vtkGeometryFilter` as it's able to utilize multiple threads better than `vtkDataSetSurfaceFilter`.

<details>
<summary>VTK build script</summary>

```bash
#!/usr/bin/env bash
set -e

mkdir -p build
cd build

PYBIN=$(which python)

# yes, python paths are hard coded. I have no idea what I'm doing on nix-os
cmake -GNinja \
      -DCMAKE_BUILD_TYPE=Release \
      -DVTK_BUILD_TESTING=OFF \
      -DVTK_BUILD_DOCUMENTATION=OFF \
      -DVTK_BUILD_EXAMPLES=OFF \
      -DVTK_DATA_EXCLUDE_FROM_ALL:BOOL=ON \
      -DVTK_MODULE_ENABLE_VTK_PythonInterpreter:STRING=NO \
      -DVTK_MODULE_ENABLE_VTK_WebCore:STRING=NO \
      -DVTK_MODULE_ENABLE_VTK_WebGLExporter:STRING=NO \
      -DVTK_MODULE_ENABLE_VTK_WebPython:STRING=NO \
      -DVTK_WHEEL_BUILD=ON \
      -DVTK_SMP_IMPLEMENTATION_TYPE=TBB \
      -DPython3_EXECUTABLE=/nix/store/8v5s2s5ixpn5hn5pivv7ybsg27sinlv7-python3-3.11.14/bin/python3 \
      -DPython3_INCLUDE_DIR=/nix/store/8v5s2s5ixpn5hn5pivv7ybsg27sinlv7-python3-3.11.14/include/python3.11 \
      -DPython3_LIBRARY=/nix/store/8v5s2s5ixpn5hn5pivv7ybsg27sinlv7-python3-3.11.14/lib/libpython3.11.so \
   ../
      # -DPython3_EXECUTABLE=$PYBIN ../
ninja

# buld wheel
python setup.py bdist_wheel
```

</details>

<details>
<summary>Benchmark code</summary>

```py
import timeit
import vtk

r = vtk.vtkUnstructuredGridReader()
r.SetFileName("/tmp/grid.vtk")
r.Update()

grid = r.GetOutput()

pass_ids = False
ugrid = r.GetOutput()


def ds_filter():
    dss = vtk.vtkDataSetSurfaceFilter()
    dss.SetInputData(grid)
    dss.SetPassThroughCellIds(pass_ids)
    dss.SetPassThroughPointIds(pass_ids)
    dss.Update()
    return dss.GetOutput()


def geo_filter():
    gf = vtk.vtkGeometryFilter()
    gf.SetInputData(grid)
    gf.SetPassThroughCellIds(pass_ids)
    gf.SetPassThroughPointIds(pass_ids)
    gf.Update()
    return gf.GetOutput()


# warmup
n = 20
geo_time = timeit.timeit(geo_filter, number=n) / n
ds_time = timeit.timeit(ds_filter, number=n) / n

geo_time = timeit.timeit(geo_filter, number=n) / n
ds_time = timeit.timeit(ds_filter, number=n) / n

print(f"ds_filter: {ds_time:.6f} s")
print(f"geo_filter: {geo_time:.6f} s")
```

</details>

Benchmark results on a 426,012 cell and 1,541,264 `UnstructuredGrid` (linear, hex-dominant):

```
[nix-shell:~/source/VTK]$ python script.py
ds_filter: 0.080470 s
geo_filter: 0.020019 s
```

I think this is another reason to move to `vtkGeometryFilter`. However, note that by default, VTK wheels are built with `VTK_SMP_IMPLEMENTATION_TYPE=SEQUENTIAL`, which means that we won't see the performance benefits from this unless Kitware builds their wheels with multi-threaded support.

pyvista/core/filters/data_object.py
```diff
@@ -3259,6 +3259,188 @@ def extract_all_edges(  # type: ignore[misc]
...
+            raise ValueError(msg)
+
+        _validation.check_contains(
+            ('geometry', 'dataset_surface', None), must_contain=algorithm, name='algorithm'
+        )
+        if algorithm is None and nonlinear_subdivision == 1:
+            # Warn about future change in default alg when the default
+            # 'nonlinear_subdivision' is used
+            warn_future()
+            algorithm = 'dataset_surface'  # The old default behavior
```
user27182: This check isn't quite correct. The default will in fact _always_ be `'dataset_surface'` if the input has non-linear cells, and not just if `nonlinear_subdivision == 1` . This is because `vtkGeometryFilter` delegates to `vtkDataSetSurfaceFilter` internally by default if there are _any_ non-linear cells. It doesn't actually matter what the value of `nonlinear_subdivision` is for determining the delegation. See https://github.com/Kitware/VTK/blob/4d4939b6f716ae8d418d7cbf66f8dd508d83ec9e/Filters/Geometry/vtkGeometryFilter.cxx#L2889-L2891

So this means that users can currently set `'geometry'` as the desired alg, but if the mesh has non-linear cells, then the resulting surface is actually generated by `vtkDataSetSurfaceFilter`.  But if a user explicitly asks for `'geometry'`, then I think we should consider _not_ delegating to `vtkDataSetSurfaceFilter`. This would require that we first check for the presence of non-linear cells, (e.g. using `distinct_cell_types` and checking the linearity of each cell type), and raise an error if trying to use `'geometry'` with non-linear cells.

This to me suggests that we need a _third_ option for the `algorithm` keyword: `'default'`.
The options would then be:
- `'geometry'`: Use `vtkGeometryFilter` _without_ allowing delegation for non-linear cells
- `'dataset_surface'`: Use `vtkDataSetSurfaceFilter`
- `'default'`: Use `vtkGeometryFilter`, but allow delegation for non-linear cells

Since `vtkGeometryFilter` handles this delegation automatically, the main difference between `'geometry'` and `'default'` in terms of PyVista code would effectively be to check for non-linear cells (and raise error) for the `'geometry'` case.

akaszynski: That works. Note that checking for non-linear cells might be expensive (running unique or `set` on potentially millions of cells).

user27182: Yes, and users will generally be encouraged to use `default` anyway to avoid this. This is mostly a protection in case users actually care about the output having specific ordering or handling from the different filters, or if they rely on the mesh being closed by default (which it may not be if delegated to DataSetSurface)

And from what I can tell the vtk code already does the IsLinear check by default as well? Otherwise it can't know to delegate. Maybe they're using a fast path though... and for 9.6 users the distinct cell types code uses VTK methods, not np.unique, not sure if it's actually faster though.

user27182: I just ran this, the mesh has 1.7 million cells
``` py
import pyvista as pv
import timeit
# Load the example mesh
mesh = pv.examples.download_electronics_cooling()
# Function to access distinct_cell_types
def get_distinct_cell_types():
    return mesh[1].distinct_cell_types
# Time it using timeit
times = timeit.repeat(
    stmt="get_distinct_cell_types()",
    setup="from __main__ import get_distinct_cell_types",
    repeat=5,  # number of trials
    number=10  # number of executions per trial
)
print("Execution times (s) per 10 calls:", times)
print("Average time per single call (s):", sum(times) / len(times) / 10)
```
With `vtk==9.6.0rc3`, we get:
```
Average time per single call (s): 0.0001900116799515672
```
And with `vtk==9.5.2`:
```
Average time per single call (s): 0.40515386334009235
```
And so this check is like 2000 times faster with VTK 9.6

akaszynski: 1.9 ms is pretty trivial. I'm good with including this check when alg is set to `"geometry"`. Let's not bother with a branch per version of VTK.

akaszynski:
+1 for `“auto”`

pyvista/core/filters/data_object.py
```diff
@@ -3259,6 +3261,190 @@ def extract_all_edges(  # type: ignore[misc]
...
+            # Despite its name, it uses `vtkDataSetSurfaceFilter` internally:
+            # https://github.com/Kitware/VTK/blob/e75f54db867f82ab50ad887e5ac36bf82425ee69/Filters/Geometry/vtkCompositeDataGeometryFilter.cxx#L81
+            if (
+                algorithm == 'dataset_surface'
+                and not pass_cellid
+                and not pass_pointid
+                and not progress_bar
+                and nonlinear_subdivision == 1
+            ):
+                return mesh._composite_geometry_filter()
```
user27182: I have previously mentioned that we should keep this branch since it's faster than manually iterating over each block of the mesh with `MultiBlock.generic_filter`. But I ran this benchmark:

``` py
import pyvista as pv
import timeit

mesh = pv.examples.download_electronics_cooling()

def extract_surface():
    return mesh.extract_surface(algorithm='dataset_surface', pass_cellid=True, pass_pointid=False)

times = timeit.repeat(
    stmt="extract_surface()",
    setup="from __main__ import extract_surface",
    repeat=5,  # number of trials
    number=10  # number of executions per trial
)

print("Execution times (s) per 10 calls:", times)
print("Average time per single call (s):", sum(times) / len(times) / 10)
```

Using `pass_cellid=True`, which uses the `MultiBlock.generic_filter` branch, I get:
```Average time per single call (s): 0.45728222666017243```

And using `pass_cellid=False` to trigger the branch which uses `vtkCompositeGeometryFilter`, I get: 
```Average time per single call (s): 0.4670622850002838```

And so, there is no performance gain in this case. I think where I saw the improvement last time was with a small mesh, so the time difference was probably due to python function call overhead rather than algorithm execution.

Therefore, I think we should remove this special case which will greatly simplify the implementation.

Comment (user27182) [reply pyvista/core/filters/data_object.py]: Also, as a side note, using `vtkGeometryFilter` with:
`mesh.extract_surface(algorithm='geometry', pass_cellid=False, pass_pointid=False)`
produces:
```Average time per single call (s): 0.33407516082050276```

And so `'geometry'` (0.33s) is faster than `'dataset_surface'` (0.46s), which reaffirms claims that `vtkGeometryFilter` is more performant, and should indeed be the default alg.

File: tests/core/test_dataset_filters.py [hunk 5]
```diff
@@ -2709,7 +2715,18 @@ def test_decimate_boundary():
...
+        geom = dataset.extract_surface(algorithm='auto', progress_bar=True)
+        assert geom is not None
+        assert isinstance(geom, pv.PolyData)
+    # Now test composite data structures
+    output = multiblock_all.extract_surface(algorithm='auto')
+    assert isinstance(output, pv.PolyData)
+
+
+@pytest.mark.parametrize('as_multiblock', [True, False])
+def test_extract_surface_nonlinear(as_multiblock):
```
user27182, line 2729: We can move all `extract_surface` tests into `test_dataobject_filters`. They're left as-is for review to see the diff, but can move then as part of this PR or as a follow-up.

pyvista/core/filters/data_set.py
```diff
@@ -5432,105 +5470,68 @@ def _extract_values(  # type: ignore[misc]
...
+        if algorithm in ('auto', 'geometry'):
+            # Default case: use vtkGeometryFilter. This will automatically delegate to
+            # vtkDataSetSurfaceFilter internally as needed for non-linear cells
+            if algorithm == 'geometry' and isinstance(self, pv.UnstructuredGrid):
+                is_linear = (
+                    _vtk.vtkCellTypeUtilities.IsLinear
+                    if pv.vtk_version_info >= (9, 6, 0)
+                    else _vtk.vtkCellTypes.IsLinear
+                )
+                if not all(is_linear(celltype) for celltype in self.distinct_cell_types):
```
user27182: We should consider adding a new property: `UnstructuredGrid.is_linear` to encapsulate this check. Can do it as part of this PR or a separate PR for better visibility.

user27182: This is done in https://github.com/pyvista/pyvista/pull/8286

user27182:
> I think this is another reason to move to `vtkGeometryFilter`. However, note that by default, VTK wheels are built with `VTK_SMP_IMPLEMENTATION_TYPE=SEQUENTIAL`, which means that we won't see the performance benefits from this unless Kitware builds their wheels with multi-threaded support.

I have asked about enabling this here:
https://discourse.vtk.org/t/build-python-wheels-with-smp-tools-for-runtime-backend-selection/16268

pyvista/core/filters/data_set.py
```diff
@@ -1294,13 +1298,42 @@ def extract_geometry(  # type: ignore[misc]
...
+
+    def _geometry_filter(  # type: ignore[misc]
+        self: _DataSetType,
+        *,
+        pass_pointid: bool,
+        pass_cellid: bool,
+        extent: VectorLike[float] | None,
+        nonlinear_subdivision: int,
+        progress_bar: bool,
+        message: str = 'Extracting Geometry',
```
user27182, line 1325: This message keyword is really only used for backwards compatibility to ensure that the old calls to `extract_geometry` still show `'Extracting Geometry'` when progress bar is enabled. We can simplify the functions and remove this though, and always show the newer text `'Extracting Surface'` instead.
I really don't think anyone will notice this change if they're still using `extract_geometry`.

pyvista/core/filters/data_set.py
user27182:
```suggestion
            if algorithm == 'geometry' and self.has_nonlinear_cells:
```

user27182:
> Agree with the changes. `algorithm="auto"` makes more sense than `"default"` as it shows that there's some logic under-the-hood. We might want to fully deprecate it in the future, but that can be after several releases and if users don't have issues with using `vtkGeometryFilter` by default.

We're using `None` to mean unselected, so that we know if the value is unset so that we can issue the future warning. And then we tell users to use `'auto'` by default. But then as you say we may want to deprecate this later, because we don't actually need `'auto'`, it's only used for the future warning.

So perhaps, if what we really want is for the default to be `None`, what if we swap these two around? I.e. make `'auto'` the default, and make it deprecated out-of-the-box, and tell users to use `algorithm=None` instead? That way, `'auto'` is strictly used for the future warning, AND it's immediately deprecated, such that after the deprecation we're left with `algorithm=None` as the default, and only have `'geometry'` and `'dataset_surface'` as the two alg options.

user27182:
> > Agree with the changes. `algorithm="auto"` makes more sense than `"default"` as it shows that there's some logic under-the-hood. We might want to fully deprecate it in the future, but that can be after several releases and if users don't have issues with using `vtkGeometryFilter` by default.
> 
> 
> 
> We're using `None` to mean unselected, so that we know if the value is unset so that we can issue the future warning. And then we tell users to use `'auto'` by default. But then as you say we may want to deprecate this later, because we don't actually need `'auto'`, it's only used for the future warning.
> 
> 
> 
> So perhaps, if what we really want is for the default to be `None`, what if we swap these two around? I.e. make `'auto'` the default, and make it deprecated out-of-the-box, and tell users to use `algorithm=None` instead? That way, `'auto'` is strictly used for the future warning, AND it's immediately deprecated, such that after the deprecation we're left with `algorithm=None` as the default, and only have `'geometry'` and `'dataset_surface'` as the two alg options.
> 
> 

I guess we can't do that, because the default isn't actually currently auto, but is 'dataset_surface'. Still though, we could introduce our own temporary sentinel value, e.g. "UNSET", and use that to indicate that no value has been provided. 


So in the signature we set `algorithm=UNSET` and warn users that `algorithm=None` will be the future default, and then go on to set the current default `algorithm='dataset_surface`.

akaszynski:
> So in the signature we set `algorithm=UNSET` and warn users that `algorithm=None` will be the future default, and then go on to set the current default `algorithm='dataset_surface`.

That probably works best for the long run as `None` will equate to our current "auto" behavior. If we deprecate over three minor releases (probably a year if we're honest), that should give users enough time to refactor. Our usual pattern is to set `None` if there's some underlying logic rather than a neutral "smart" value like "auto".

This has been a long enough PR, and the current implementation is also fine, but this might be a bit more optimal if you don't mind extending it a bit further.

user27182:
I can make changes in a follow-up if you prefer, just to get the current revisions in. But yea I think it will be annoying to force users to use "auto" now, only to force users to stop using "auto" in a few releases and use None instead. So I think we should try to make it so that None will be the default before making a release.

user27182:
> I can make changes in a follow-up if you prefer, just to get the current revisions in. But yea I think it will be annoying to force users to use "auto" now, only to force users to stop using "auto" in a few releases and use None instead. So I think we should try to make it so that None will be the default before making a release.

This is done in #8289. That PR is based on this PR's branch, so that #8289 can be merged into this branch first, then we merge this PR into main.

PR #8274, opened 2026-01-29
Enable customizing `AxesAssembly` geometry and add anti-distortion scaling mode

user27182:
### Overview

When scaling `AxesAssembly`, the axes are distorted when using non-uniform scaling, e.g:
``` py
pv.AxesAssembly(scale=(0.4, 1.0, 2.5))
```

axes_assembly_scale_mode uniform

With this PR:
``` py
pv.AxesAssembly(scale=(0.4, 1.0, 2.5), scale_mode='anti_distortion')
```

axes_assembly_scale_mode normalize_shape

### Details

- Make `AxesGeometrySource.shaft_radius` and `AxesGeometrySource.tip_radius` store a separate radius for each axis. This is technically a breaking change, since the getter returns a tuple instead of float, but setting a float still works.
- Expose `AxesGeometry` properties directly as `AxesAssembly` properties.
- Add new `scale_mode` keyword with anti-distortion option when using non-uniform scaling.

### DIFF
--- a/pyvista/core/utilities/geometric_sources.py
+++ b/pyvista/core/utilities/geometric_sources.py
@@ -3228,6 +3228,8 @@ class AxesGeometrySource(_NoNewAttrMixin):
         # Set flags
         self._symmetric = symmetric
         self._symmetric_bounds = symmetric_bounds
+
+        # Used by AxesAssembly for scale_mode='anti_distortion'
         self._anti_distortion_factor: NumpyArray[float] = np.ones(shape=(3,), dtype=float)
 
     def __repr__(self: AxesGeometrySource) -> str:


user27182:
Should be ready except for doc images, looks like the docs build step is queued, not sure if there's anything I can do get it working @akaszynski

akaszynski:
> Should be ready except for doc images, looks like the docs build step is queued, not sure if there's anything I can do get it working @akaszynski

Looking into it.

akaszynski (approved):
Looks like CI is back to normal.

I've not tested it independently, but functionally (from doc cache) looks like it works.

LGTM.

PR #8291, opened 2026-02-05
Raise OSError for invalid file paths when saving

### LINKED ISSUE #8290: Silent failure to save data when output directory does not exist

When attempting to save data to a path whose parent directory does not exist, VTK prints an error (e.g., "Unable to open file ..."), but the failure is not surfaced to Python as an exception such as `FileNotFoundError` / `OSError`.

As a result, the Python process continues execution as if the save operation had succeeded. This may lead to silent data loss or incorrect downstream behavior.

Suggested behavior

- If the output directory does not exist, raise `FileNotFoundError` (or a related `OSError`) from `mesh.save(...)`.

This would be consistent with typical Python file I/O APIs (e.g., `open()`, NumPy’s `numpy.save`).

Ideally, we would like to check the write operation result from VTK API, but it seems it does not return any error code. Therefore, I suggest this approach.

```python
import pathlib
import pyvista as pv


if __name__ == "__main__":
    cylinder = pv.Cylinder(center=(0, 0, 0), direction=(0, 0, 1))

    try:
        here = pathlib.Path(__file__).parent / "not_existing_directory"
        cylinder.save(here / "cylinder.vtk")
    except Exception as e:
        # Cannot catch the core dumped error from vtk
        print("No exception raised; execution continues (unexpected).")
        raise e
    
    # This line can be reached !
    print("Finished")
```

Output

```
2026-02-05 14:58:55.515 (   0.168s) [    7C723A044740]      vtkDataWriter.cxx:176    ERR| vtkPolyDataWriter (0x2e747120): Unable to open file: /home/riku/repos/personals/pyvista/workspace/20260205/not_existing_directory/cylinder.vtk
ERROR:root:Unable to open file: /home/riku/repos/personals/pyvista/workspace/20260205/not_existing_directory/cylinder.vtk
Finished
```

### System Information

```shell
  Date: Thu Feb 05 05:51:49 2026 UTC

                OS : Linux (Pop!_OS 22.04)
            CPU(s) : 14
           Machine : x86_64
      Architecture : 64bit
               RAM : 38.8 GiB
       Environment : Python
       File system : ext4
        GPU Vendor : Intel
       GPU Version : 4.6 (Core Profile) Mesa 24.2.8-1~bpo12+1pop1~1744225826~22.04~b077665
  MathText Support : True

  Python 3.13.5 (main, Jun 12 2025, 12:40:22) [Clang 20.1.4 ]

           pyvista : 0.47.dev0
               vtk : 9.5.2
             numpy : 2.3.5
        matplotlib : 3.10.7
            scooby : 0.11.0
             pooch : 1.8.2
            pillow : 12.0.0
 typing-extensions : 4.15.0
          cyclopts : 4.3.0
    pytest-pyvista : 0.3.2
             PyQt5 : Version unknown
           IPython : 9.8.0
             scipy : 1.16.0
              tqdm : 4.67.1
           imageio : 2.37.2
            meshio : 5.3.5
         cmcrameri : 1.9
           cmocean : 4.0.3
          colorcet : 3.1.0
             trame : 3.12.0
      trame-client : 3.11.2
      trame-server : 3.9.0
         trame-vtk : 2.10.2
     trame-vuetify : 3.2.0
        ipywidgets : 8.1.8
        jupyterlab : 4.5.0
     nest-asyncio2 : 1.7.1
```

### Screenshots

_No response_

riku-sakamoto:
### Overview

Closes #8290

(Updated)
This PR implements guard `if` statement before and after `Write` operation.
As a result, PyVista now raises `FileNotFoundError` or `OSError` when a VTK writer fails.

~~This PR uses the `pv.VtkErrorCatcher` context manager to capture internal VTK errors
during write operations. As a result, PyVista now raises `pv.VTKExecutionError`
when a VTK writer fails.~~

#### About Error Code

We can get error code from vtk writer using `writer.GetErrorCode()`.
So, if we want to emit more specific Python exceptions like `FileNotFoundError`, we can do so by checking the returned code. I am happy to change it if you think it is better.

#### `.Output()` API

Following to @user27182 suggestion, I looked into  the VTK discussion and merge request below. It seems that using `.Update()` API enables us to check error status in future VTK versions.

* https://discourse.vtk.org/t/changing-vtkalgorithm-update-return-type-from-void-to-bool/16164
* https://gitlab.kitware.com/vtk/vtk/-/merge_requests/12654

However, in current VTK version used in PyVista, `.Update()` still return `None`. Also, `GetExecutive().Update()` returns `1` even when writing process is failed.

So, I did not use this API in this PR. 
I've added sample code below to demonstrate the behavior of these APIs in `pyvista/core/dataobject.py`

```python

writer.write()

ret = writer.writer.GetExecutive().Update()
print(ret)
# Output: 1

ret = writer.writer.Update()
print(ret)
# Output: None
```


user27182:
Thanks for the fix! My main concern with using `pv.VtkErrorCatcher` by default is that I think it can be a bit slow (based on my experience trying to catch VTK errors in our tests (#7766)). Maybe we should run some benchmarks on some different examples with both small files and large files to see if it affects performance? E.g. [`download_electronics_cooling`](https://dev.pyvista.org/api/examples/dataset_gallery#electronics-cooling-dataset) is probably a good example of a "large" file.

To avoid using the error catcher, I think maybe checking the error code directly first may be good, and then raise an error based on the error code.

akaszynski:
> To avoid using the error catcher, I think maybe checking the error code directly first may be good, and then raise an error based on the error code.

Alternatively, you could try twice: one with VTK's error handling and without. We could also add a kwarg to enable VTK's error catcher, letting the user know that it potentially reduces performance.

Agreed regarding the issue of performance. I'd like to see a simple benchmark before making this default (though I'm fine with it being optional).

user27182:
> Following to @user27182 suggestion, I looked into the VTK discussion and merge request below. It seems that using `.Update()` API enables us to check error status in future VTK versions.
> 
> * https://discourse.vtk.org/t/changing-vtkalgorithm-update-return-type-from-void-to-bool/16164
> * https://gitlab.kitware.com/vtk/vtk/-/merge_requests/12654
> 
> However, in current VTK version used in PyVista, `.Update()` still return `None`. Also, `GetExecutive().Update()` returns `1` even when writing process is failed.
> 
> So, I did not use this API in this PR. I've added sample code below to demonstrate the behavior of these APIs in `pyvista/core/dataobject.py`
> 
> ```python
> writer.write()
> 
> ret = writer.writer.GetExecutive().Update()
> print(ret)
> # Output: 1
> 
> ret = writer.writer.Update()
> print(ret)
> # Output: None
> ```

I had a closer look at the VTK discourse thread, and for `vtkWriter` specifically, the status is reported with a call to `Write()`, and not `Update()`. But `Write()` doesn't report the status correctly, which is bug with VTK: https://gitlab.kitware.com/vtk/vtk/-/issues/19934

I think the correct approach here is to check the VTK status first, and then if the status reports a failure, we retry a second time using `VtkErrorCatcher` to catch and raise as a proper python error. This would be similar to what's done with `_update_alg` in #8292.

So I think we should wait for https://gitlab.kitware.com/vtk/vtk/-/issues/19934 to be fixed before implementing this for calls to `Write()`, and hence we should not use `VtkErrorCatcher` by default and I don't think we should add a keyword to make this optional. For now, to address https://github.com/pyvista/pyvista/issues/8290 specifically, I think maybe just using `pathlib` to check if the parent dir exists before writing is sufficient.

riku-sakamoto:
Thanks for the review!

### About the performannce

As you pointed out, I've confirmed that using context manager have some performance overhead.

I did a simple benchmark to compare the writing time with and without context manager (`pv.VtkErrorCatcher`).  I measured the time taken to save three datasets, and calculated the average time over 10 runs for each case. (average and standard deviation are reported.)

The overhead depends on the dataset size. For larger datasets, it is relatively small (about 5%), whereas for smaller datasets it is more pronounced (about 35%).

Here are the results:

| Dataset | File size | without context manager | with context manager | ratio (with/without) |
|---------|--------|--------|-----------|--------|
| download_electronics_cooling |67.9 MB | 2.372 s +/- 0.03444 s | 2.483 s +/- 0.1171 s | 1.0467 |
|download_aero_bracket | 14.7 MB |0.402 s +/- 0.009196 s| 0.4215 s +/- 0.00624 s | 1.0485 |
| download_bird | 31.9 KB | 0.4612 ms +/- 0.1194 ms | 0.6228 ms +/- 0.5318 ms | 1.3506 |


### `.Write()` API

Thank you for letting me know about the issue of `.Write()` API.
I agree that using return value of `.Write()` is a straightforward way to check the writing status.
So, as a temporary solution, I have added a check before and after the write operation.


### Alternative approach using `GetErrorCode()`

I considered using `GetErrorCode()` to check for errors after the write operation like below. However, I found that `GetErrorCode()` does not always provide reliable error information. So, I decided not to use this approach.

Maybe, if the issue https://gitlab.kitware.com/vtk/vtk/-/issues/19934 is solved, we can use this approach in future.


```python
writer.write()
error_code = writer.writer.GetErrorCode()
if error_code != vtk.vtkErrorCode.NoError:
    with with pv.VtkErrorCatcher(raise_errors=True):
        writer.write()
```

Just to note, here are several examples to check output of `GetErrorCode()`.

* when failed to save `vtu` file, the `GetErrorCode()` returns `0` (NoError).

```python
cylinder = pv.Cylinder(center=(0, 0, 0), direction=(0, 0, 1))
here = pathlib.Path(__file__).parent / "not_existing_directory"
cylinder.cast_to_unstructured_grid().save(here / "cylinder.vtu")
```

Output
```
2026-02-06 09:04:16.076 (   0.333s) [    7B9BF29D2740]       vtkXMLWriter.cxx:633    ERR| vtkXMLUnstructuredGridWriter (0x1f1c6aa0): Error opening output file "/home/riku/repos/personals/pyvista/workspace/20260205/not_existing_directory/cylinder.vtu"
ERROR:root:Error opening output file "/home/riku/repos/personals/pyvista/workspace/20260205/not_existing_directory/cylinder.vtu"
2026-02-06 09:04:16.076 (   0.333s) [    7B9BF29D2740]       vtkXMLWriter.cxx:635    ERR| vtkXMLUnstructuredGridWriter (0x1f1c6aa0): Error code "Success"
ERROR:root:Error code "Success"
```

* when failed to save `vtk` file, the `GetErrorCode()` returns `20002`.

```python
cylinder = pv.Cylinder(center=(0, 0, 0), direction=(0, 0, 1))
here = pathlib.Path(__file__).parent / "not_existing_directory"
cylinder..save(here / "cylinder.vtk")
```

Output
```
2026-02-06 09:07:20.543 (   0.315s) [    72D2B0DF8740]      vtkDataWriter.cxx:176    ERR| vtkPolyDataWriter (0x40731680): Unable to open file: /home/riku/repos/personals/pyvista/workspace/20260205/not_existing_directory/cylinder.vtk
ERROR:root:Unable to open file: /home/riku/repos/personals/pyvista/workspace/20260205/not_existing_directory/cylinder.vtk
```

akaszynski:
@riku-sakamoto, with #8292 merged can you check the behavior of pyvista now and see if this PR is still needed? I use PyVista on a performance critical closed source library and 5-35% overhead is less than ideal.

Regardless, thanks for the through benchmarks and the PR. It inspired #8292, which should cover most edge cases. If reader errors still crop up I'm still open to this PR.

user27182:
> @riku-sakamoto, with #8292 merged can you check the behavior of pyvista now and see if this PR is still needed? I use PyVista on a performance critical closed source library and 5-35% overhead is less than ideal.
> 
> 
> 
> Regardless, thanks for the through benchmarks and the PR. It inspired #8292, which should cover most edge cases. If reader errors still crop up I'm still open to this PR.

#8292 works for readers since they use `Update()`. But for this PR the issue is with writers which use `Write()`, so the issue is still valid. And pathlib is now used instead of VtkErrorCatcher so there should no longer be any performance overhead for the checks in this PR.

PR #8292, opened 2026-02-05
Catch VTK errors when updating algorithms

user27182:
### Overview

It's possible to check the status when calling `Update`. This PR adds a status check, and if the update is known to have failed, we catch and raise the error using `pyvista.VtkErrorCatcher`.


akaszynski (approved):
Only issue I see with this is very long running algorithms might take nearly twice as long before failure, but I can’t see a way around this. Perhaps warning that it failed and we’re getting the status, but IMO it’s sufficient as is. At least the user can obtain more information about what caused their filter/reader to fail.

Unless we can work around this, which I think we cant’t, this is good to go.

user27182:
> Only issue I see with this is very long running algorithms might take nearly twice as long before failure, but I can’t see a way around this. Perhaps warning that it failed and we’re getting the status, but IMO it’s sufficient as is. At least the user can obtain more information about what caused their filter/reader to fail.
> 
> Unless we can work around this, which I think we cant’t, this is good to go. 

Indeed the double execution time is unfortunate. This also means users probably shouldn't `try...except...pass` on VTKExecutionError either since the code would be super inefficient.

IMO this error catching is more for rare edge cases though to prevent silent errors from leaking through our filters which would be hard to debug.

Ideally these errors would only show up during development when creating a wrapper for the alg, and any VTK errors would be prevented from occurring in the first place with proper python checks and errors raised before the alg is ever updated.

akaszynski:
> IMO this error catching is more for rare edge cases though to prevent silent errors from leaking through our filters which would be hard to debug.

Agreed, which is why I'm more than happy to merge this. Unknown VTK errors are a bit frustrating and I prefer explicit errors. Running twice on edge cases is better than having it always on.

PR #8295, opened 2026-02-05
Limit Sphinx Gallery examples line length to 90 characters

user27182:
### Overview

The project has a default line length of 99. This is a problem for Sphinx Gallery examples, since it requires horizontal scrolling, e.g. for https://dev.pyvista.org/examples/01-filter/extract_surface#compare-surface-extract-algorithms
Screenshot 2026-02-05 at 1 19 35 PM

This PR fixes this by enforcing a max line length of 90 instead for files inside `examples`. A new `ruff.toml` config file is added to specify a different line length than the rest of the project. See https://docs.astral.sh/ruff/configuration/#config-file-discovery


examples/ruff.toml
tkoyama010, line 1: Nice! I didn't know we could use this.

tkoyama010 (approved):

PR #8301, opened 2026-02-08
Fix mesh_validation.py assertion error

tkoyama010:
### Overview

Fixes the AssertionError in `examples/99-advanced/mesh_validation.py` where the validation method now detects both 'non_convex' and 'inverted_faces' issues in the invalid polyhedron test case, but the assertion only checked for 'inverted_faces'.

This addresses the Sphinx Gallery build failure:
```
AssertionError at line 87:
assert report.invalid_fields == ('inverted_faces',)
```

### Details

- Updated assertion on line 87 to expect both `'non_convex'` and `'inverted_faces'` in the `invalid_fields` tuple, matching the actual validation output


user27182:
The assertion depends both on OS version and VTK version. I suggest putting the assertion behind an OS and/or VTK version check instead. It currently requires Linux and VTK 9.6 to work (I.e. our current build CI). 

That example has a narrative, and it's important that the assertion remains as-is for the entire example to make sense. E.g. it had to be reworked for the VTK 9.6 update in #8228

user27182:
> The assertion depends both on OS version and VTK version. I suggest putting the assertion behind an OS and/or VTK version check instead. It currently requires Linux and VTK 9.6 to work (I.e. our current build CI). 

I'm thinking we can add a helper "assert" function that does the proper version/os checks and only conditionally asserts

user27182:
And maybe so that user don't see them, we can also ignore the assertions in the output. 

``` py 

# sphinx_gallery_start_ignore
assert True
# sphinx_gallery_end_ignore

PR #8307, opened 2026-02-09
Add docstring test for windows and older python

user27182:
### Overview

Fix https://github.com/pyvista/pyvista/issues/7378

New Python syntax checks are also included to check syntax of Sphinx Gallery examples.


.github/workflows/python-syntax.yml
```diff
@@ -0,0 +1,33 @@
...
+permissions:
+  id-token: none
+
+jobs:
+  compile:
+    name: Compile Python Files
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        python-version: [3.10, 3.11, 3.12, 3.13, 3.14]
```
akaszynski:
```suggestion
        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```

Otherwise 3.10 evaluates to 3.1

Quotes around other versions added for consistency.

.github/workflows/python-syntax.yml
```diff
@@ -0,0 +1,34 @@
+name: Check Python Syntax
```
user27182, line 1: Should consider making these checks required before allowing merge

.github/workflows/style-docstring.yml
```diff
@@ -33,8 +43,8 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: 3.13
-          cache: "pip"
+          python-version: ${{ matrix.python-version }}
+          cache: pip
```
edabor, line 47: Specifying the python version here has no impact on `tox`, since it overrides the installed python to create a virtual environment with the required version specified by `base_python` in the `tox.ini` config.

See for example:
https://github.com/pyvista/pyvista/actions/runs/21817626730/job/62942788787#step:7:17
https://github.com/pyvista/pyvista/actions/runs/21817626730/job/62942788770?pr=8307#step:7:18

You need to remove the following line in the config such that `tox` uses the python version from which the command was launched.

https://github.com/pyvista/pyvista/blob/e9c7f023b022f606ee826d2695c1faa33f747200/tox.ini#L226

.github/workflows/python-syntax.yml
```diff
@@ -0,0 +1,34 @@
...
+  id-token: none
+
+jobs:
+  compile:
+    name: Compile Python Files
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```
edabor, line 20: Could we store in the repo variables the supported versions to ease python EOL and new versions ?

https://github.com/pyvista/pyvista/settings/variables/actions

```suggestion
        python-version: ${{ fromJSON(vars.PYTHON_VERSIONS) }}
```

It could also be used elsewhere 
https://github.com/pyvista/pyvista/blob/e9c7f023b022f606ee826d2695c1faa33f747200/.github/workflows/testing-and-deployment.yml#L318

I can handle it inside another PR if you think it's worth it.

edabor: Unfortunately, repository variables are not accessible from forks so we cannot use it here 😢

https://github.com/pyvista/pyvista/actions/runs/21827486371/job/62976202037#step:4:1

.github/workflows/style-docstring.yml
user27182: Thanks. This is done in [15c463e](https://github.com/pyvista/pyvista/pull/8307/commits/15c463e148a2f4d7df1313f8d00ef736d74b6b53). 

Contributing usage for tox says

```tox run -e doctest-modules```

does this need to be updated too?

File: .github/workflows/python-syntax.yml (hunk 2 above)
Comment (user27182) [reply .github/workflows/python-syntax.yml:1]: Should also make the new Windows docstring check required

File: .github/workflows/style-docstring.yml (hunk 3 above)
Comment (edabor) [reply .github/workflows/style-docstring.yml:47]: No need since `base_python` applies for both `doctest-modules` and `doctest-local` environments.

More specifically, `tox` is run in the CI with `tox run -f doctest`, which stands for `Launch all environments that contain the **doctest** factor.`.

A tox environment is a combination of factors separated by a `-`

File: .github/workflows/style_docstr.yml [hunk 5]
```diff
@@ -33,8 +43,8 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: 3.13
-          cache: "pip"
+          python-version: ${{ matrix.python-version }}
+          cache: pip
 
       - name: Setup headless display
         uses: pyvista/setup-headless-display-action@7d84ae825e6d9297a8e99bdbbae20d1b919a0b19
```
akaszynski: Is this still necessary with vtk 9.5+?

akaszynski: Not needed on Linux. See:

https://github.com/pyvista/pyvista/actions/runs/21831346277/job/62990378975?pr=8315

```suggestion
      - name: Setup headless display on Windows
        if: runner.os == 'Windows'
        uses: pyvista/setup-headless-display-action@7d84ae825e6d9297a8e99bdbbae20d1b919a0b19
```

user27182:
Based on work from #8303 I _thought_ I understood how the caching works: 
1. only need to run the caching workflow _once_ with _any_ github-hosted runner to enable caching for _all_ github runners (e.g. we run the cache with ubuntu for the Windows unit testing)
2. it's necessary to run the workflow _separately_ for _each_ self-hosted runner (e.g. we run the cache workflow separately for the linux and macos unit tests which are self-hosted)


But apparently there's more to it than that? In this PR, I did _not_ add a separate cache job for the new Windows docstring job, so I would have expected the Windows job to download `vtk-data` directly from GitHub. But it seems that the cache is working as-is anyway, and is being restored correctly for both the self-hosted Linux docstring check and the github-hosted windows check, even though initially caching is only done for the Linux runner. 

I think it's fine to leave it as-is even though I don't fully understand it, and only revisit this later if/when we start getting HTTP failures again that need to be investigated.

EDIT: might have to do with the fact that we still run the cache restore step for both workflows. And I guess the cache would already have been created by the unit-testing jobs. So the cache is created by another, unrelated job, and then gets properly restored by the docstring job. Still, it might be best to explicitly add the cache job for Windows docstrings here.

akaszynski:
> I think it's fine to leave it as-is even though I don't fully understand it, and only revisit this later if/when we start getting HTTP failures again that need to be investigated.

Note that we're caching using a local cache for linux that's on the kubernetes cluster, but both windows and mac os must use GitHub's caching approach. Honestly, now that I know that we can use node local caches, we can actually skip pulling caches from a remote for our linux runs an incorporate something that only requires network access when the cache updates. This would effectively be as easy as providing a local path for the cache and making sure that before use the cache is updated. This would really reduce writes on the SSD as for each and every workflow we have to write 2.4 GB to disk.

user27182:
> > I think it's fine to leave it as-is even though I don't fully understand it, and only revisit this later if/when we start getting HTTP failures again that need to be investigated.
> 
> Note that we're caching using a local cache for linux that's on the kubernetes cluster, but both windows and mac os must use GitHub's caching approach. Honestly, now that I know that we can use node local caches, we can actually skip pulling caches from a remote for our linux runs an incorporate something that only requires network access when the cache updates. This would effectively be as easy as providing a local path for the cache and making sure that before use the cache is updated. This would really reduce writes on the SSD as for each and every workflow we have to write 2.4 GB to disk.

Okay! Sounds like there's room for improvement with the caching. I'll leave those updates for a separate PR.

For now the docstring jobs are (finally) passing, so I think this is ready as-is.

.github/workflows/python-syntax.yml
akaszynski: Done.

akaszynski (approved):

PR #8314, opened 2026-02-09
Update tox and CI base python to 3.14

user27182:
## Summary

Updates remaining CI workflows to use Python 3.14 as the base/default version, following the original `tox` basepython bump.

All workflows updated from 3.13 → 3.14

## Dependency bumps

- `colour-science`: `0.4.6` only supports up to Python 3.13, which broke the docs job on 3.14. Split the pin so 3.10–3.13 stay on `0.4.6` and 3.14 uses `0.4.7` (first version with 3.14 support, however they dropped 3.10 support so we have to have a double pin in the pyproject.toml now)
- `scipy`: bumped from `1.16.0` → `1.16.3` (still gated on `python_version >= "3.11"`). `1.16.0` has no cp314 wheels and the docs job tried to build from source on 3.14, failing on missing OpenBLAS.

## Docker

The Docker images (`docker/Dockerfile`, `docker/README.md`) stay on Python 3.13. The jupyter stage is pinned to `quay.io/jupyter/base-notebook:python-3.13` because upstream only publishes a single current tag, and the slim stage is kept on 3.13 so both published images (`latest` and `latest-slim`) share the same Python version.


user27182:
Looks like we'll either need to enable installing pre-releases, or wait for VTK 9.6 to be released (it should be released today).

banesullivan (dismissed):

banesullivan (dismissed):

akaszynski (approved):

PR #8318, opened 2026-02-10
Augment `cell_validator` with additional checks

### LINKED ISSUE #8276: Add `cell_quality` measures to `validate_mesh`

Filters or users may sometimes generate cells with negative volume, e.g. #6254.
We should allow users to check for this. I think we can add a `cell_quality` keyword to `validate_mesh`, and include `volume` and `area` measures by default, with the option to validate other measures based on known invalid values defined by `CellQualityInfo`.

We might need to be careful in how we implement and test this though since some cells have volume incorrectly computed, and we don't want to generate false positives.

_No response_

### Pseudocode or Screenshots

_No response_

user27182:
### Overview

`vtkCellValidator` has a few issues:
1. Cells with negative volume / incorrect orientation are not currently reported as such (https://gitlab.kitware.com/vtk/vtk/-/issues/19938)
2. Degenerate cells are not currently reported as such (https://gitlab.kitware.com/vtk/vtk/-/issues/19885)

This issue is particularly problematic because the validity states and descriptions for `vtkCellValidator` make it seem like these cases are covered and properly flagged, but they're not, giving users a false impression that cells may be valid when they may in fact not be.

This PR addresses this by:
1. Add a new status `negative_size` for checking for negative length (1D cells), negative area (2D cells) or negative volume (3D cells). This is mostly for detecting negative volume... I don't think length or area will even technically be negative but perhaps in some cases it could be.
2. Add a new status `zero_size` for checking for zero length/area/volume. This effectively indirectly checks for degenerate cells and/or for cells with coincident points.

In addition, the separate `invalid_point_references` check that was done with `validate_mesh` has been moved into `cell_validator` to simplify things.

A new `CellStatus` enum has also been added to better document and organize the cell statuses, and to provide users with a public API for it.

Resolves #8276

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ jupyter = [
   'nest-asyncio2',
   'trame-client>=2.12.7',
   'trame-server >= 2.11.7, != 3.7.*, != 3.8.0',
-  'trame-vtk>=2.5.8',
+  'trame-vtk>=2.5.8, !=2.10.3', # Exclude 2.10.3 due to https://github.com/Kitware/trame-vtk/issues/101
   'trame-vuetify>=2.3.1',
   'trame>=2.5.2',
 ]


pyvista/core/filters/data_object.py
```diff
@@ -4548,3 +4763,25 @@ def add_cell_ids(dataset: DataSet | MultiBlock):
                 block.n_cells, dtype=_Crinkler.INT_DTYPE
             )
         return active_scalars_info
+
+
+def _cell_status_docs_insert():
+    """Format CellStatus enum info for inserting into a docstring."""
+    return '\n' + '\n'.join(
+        [f'- :attr:`~pyvista.CellStatus.{status.name}`: {status.__doc__}' for status in CellStatus]
+    )
```
akaszynski, line 4772: This is a great way to programmatically generate a long description.

Since it's programmatic, would sorting improve readability?

Image

user27182:
Looks like the tests are flaky:

https://github.com/pyvista/pyvista/actions/runs/21951704608/job/63404270696#step:7:1
Will update the test to only assert on the invalid point refs part.

pyvista/core/filters/data_object.py
user27182: Yes we can sort it, I presume you mean by name? (It's already sorted by int value)

Maybe we can keep `VALID` first but sort the rest?

user27182:
> Looks like the tests are flaky:
> 
> https://github.com/pyvista/pyvista/actions/runs/21951704608/job/63404270696#step:7:1 Will update the test to only assert on the invalid point refs part.

Hm, for this test, I think it's important to verify the full report. For now we can simply retry the test to make CI green. I would prefer to do a follow-up and add a new `exclude_field` keyword, so we can do `validate_mesh(exclude_field='negative_size')` for this test to address this.

pyvista/core/filters/data_object.py
user27182: Should we extend the sorting to the `validate_mesh` report too? I was thinking of also moving the Point Fields in the report to show before the Cell Fields. Can do both in a follow-up.

akaszynski: It's up to you. For me, it's easier to look for something when lists are sorted. We could keep valid above the rest.

user27182: Will submit a new PR shortly. Will also sort the field names in the mesh validation report so that everything is sorted and consistent across `validate_mesh` and `cell_validator`

user27182: Done in #8324 
That PR is currently based on this PR's branch to see the diff. But we should _not_ merge into this branch. Instead, I think #8324 should be merged into main as a separate PR/commit.

user27182:
> I would prefer to do a follow-up and add a new `exclude_field` keyword, so we can do `validate_mesh(exclude_field='negative_size')` for this test to address this.

I'll wait for this PR and #8324 to be merged first to avoid merge conflicts.

pyvista/core/filters/data_object.py
user27182: New docs from #8324 
image

PR #8321, opened 2026-02-12
Remove deprecated parameters

user27182:
### Overview
Remove:
-  `max_time` from `streamlines_from_source` 
- `n_verts`, `n_lines`, `n_faces`, `n_strips` from `PolyData.__init__`


user27182:
Seems these parameters already have tests in place to enforce a deprecation schedule. But the docstrings don't match the tests... which is why I thought these were ready for removal.

Issue #8283, opened 2026-02-02
local html export: trame-vtk 2.10.3 causes 404 page to be written on pv.Plotter.export_html


MarkusZimmerDLR:
### Describe the bug, what's wrong, and what you expected.

On the most recent version of trame-vtk (2.10.3), the export of html-plots is compromised.
Exporting >local< html plots with the script below only exports a github 404 page (?!)

Downgrading to trame-vtk 2.10.1 solves this problem.

### Steps to reproduce the bug.

```python
import pyvista as pv
s = pv.Sphere()
p = pv.Plotter()
_ = p.add_mesh(s)
p.export_html()
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Mon Feb 02 14:33:02 2026 UTC

                  OS : Linux (Red Hat Enterprise Linux 8.10)
              CPU(s) : 16
             Machine : x86_64
        Architecture : 64bit
                 RAM : 251.1 GiB
         Environment : IPython
         File system : xfs
          GPU Vendor : NVIDIA Corporation
        GPU Renderer : Quadro RTX 4000/PCIe/SSE2
         GPU Version : 4.5.0 NVIDIA 575.57.08
       Render Window : vtkXOpenGLRenderWindow
    MathText Support : True

  Python 3.13.11 | packaged by conda-forge | (main, Jan 26 2026, 23:57:06)
  [GCC 14.3.0]

             pyvista : 0.46.5
                 vtk : 9.5.2
               numpy : 2.4.2
          matplotlib : 3.10.8
              scooby : 0.11.0
               pooch : 1.9.0
              pillow : 12.1.0
             imageio : 2.37.2
           pyvistaqt : 0.11.3
               PyQt5 : 5.15.11
             IPython : 9.10.0
          ipywidgets : 8.1.8
               scipy : 1.17.0
                tqdm : 4.67.2
              meshio : 5.3.5
               trame : 3.12.0
        trame_client : 3.11.2
        trame_server : 3.10.0
           trame_vtk : 2.10.3
       trame_vuetify : 3.2.0
jupyter_server_proxy : 4.4.0
        nest_asyncio : 1.6.0
--------------------------------------------------------------------------------
```

### Screenshots

<img width="917" height="682" alt="Image" src="https://github.com/user-attachments/assets/f4f98c2e-9817-4e0d-83b8-deaf1d14e41f" />

edabor:
[...]

PR #8322, opened 2026-02-12
Exclude `trame-vtk` 2.10.3 from deps

edabor:
### Overview

Fix https://github.com/pyvista/pyvista/issues/8283


edabor:
We might need to exclude `2.11` also ... https://github.com/Kitware/trame-vtk/issues/101#issuecomment-3891547333

edabor:
... or maybe set an upper limit to `< 2.10.3` as an interim solution

Issue #8323, opened 2026-02-12
`DynamicScraper` fails to create the interactive scene with `trame-vtk>=2.11.0'`.


AlejandroFernandezLuces:
### Describe the bug, what's wrong, and what you expected.

When using the `DynamicScraper` to generate the gallery with `Sphinx`, it generates the static image, but the interactive scene is not generated. In previous versions of `trame-vtk` this is still working well. See for instance

https://visualization-interface.tools.docs.pyansys.com/version/dev/examples/00-basic-pyvista-examples/plain_usage.html

I see that `trame-vtk` version is limited in the doc dependencies, is this a known issue?

### Steps to reproduce the bug.

To replicate, you can simply build the project docs following this guide and see the results: 

https://visualization-interface.tools.docs.pyansys.com/version/stable/contributing.html#build-the-documentation

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Thu Feb 12 15:32:14 2026 UTC

                OS : Windows (11 10.0.26100 SP0 Multiprocessor Free)
            CPU(s) : 28
           Machine : AMD64
      Architecture : 64bit
               RAM : 63.7 GiB
       Environment : Python
        GPU Vendor : Intel
      GPU Renderer : Intel(R) RaptorLake-S Mobile Graphics Controller
       GPU Version : 4.5.0 - Build 32.0.101.6556
     Render Window : vtkWin32OpenGLRenderWindow
  MathText Support : True

  Python 3.13.7 (tags/v3.13.7:bcee1c3, Aug 14 2025, 14:15:11) [MSC v.1944 64
  bit (AMD64)]

           pyvista : 0.46.5
               vtk : 9.5.2
             numpy : 2.4.2
        matplotlib : 3.10.8
            scooby : 0.11.0
             pooch : 1.9.0
            pillow : 12.1.1
           IPython : 9.10.0
        ipywidgets : 8.1.8
              tqdm : 4.67.3
             trame : 3.12.0
      trame_client : 3.11.2
      trame_server : 3.10.0
         trame_vtk : 2.11.0
     trame_vuetify : 3.2.1
      nest_asyncio : 1.6.0
--------------------------------------------------------------------------------
```

### Screenshots

<img width="1061" height="597" alt="Image" src="https://github.com/user-attachments/assets/c3cae4e3-8ba6-40c7-8a9d-bf21fb2b8083" />

PR #8325, opened 2026-02-12
Upper pin `trame-vtk`

edabor:
### Overview

Placing an upper-pin on `trame-vtk` while https://github.com/Kitware/trame-vtk/issues/101 is not fixed.

Some motivations: 
https://github.com/pyvista/pyvista/issues/8323#issuecomment-3892903253
https://github.com/Kitware/trame-vtk/issues/101#issuecomment-3891773983

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ jupyter = [
   'nest-asyncio2',
   'trame-client>=2.12.7',
   'trame-server >= 2.11.7, != 3.7.*, != 3.8.0',
+  'trame-vtk != 2.10.3, != 2.11.*',
   'trame-vtk>=2.5.8, <2.10.3', # Unpin when https://github.com/Kitware/trame-vtk/issues/101 is closed
   'trame-vuetify>=2.3.1',
   'trame>=2.5.2',


pyproject.toml
```diff
@@ -48,7 +48,7 @@ jupyter = [
   'nest-asyncio2',
   'trame-client>=2.12.7',
   'trame-server >= 2.11.7, != 3.7.*, != 3.8.0',
-  'trame-vtk>=2.5.8, !=2.10.3', # Exclude 2.10.3 due to https://github.com/Kitware/trame-vtk/issues/101
+  'trame-vtk>=2.5.8, <2.10.3', # Unpin when https://github.com/Kitware/trame-vtk/issues/101 is closed
```
user27182, line 52: I'm not sure if we should unpin this later. Breakage from trame-vtk updates does not seem like a one-time irregular thing, but has happened before too (e.g. with subplots).

I'm inclined to pin this similar to how `vtk` is pinned... assume there will be breakage, and PyVista must act as a gatekeeper to prevent downstream users from being exposed to the breakage as well.

edabor: Agreed. Done in a7de4e2ec8ef91d1f633bf726b636803701a0f09

user27182: Thanks. I like the explicit skips for known broken versions. 

I'm still thinking we should always keep this pinned though, and not unpin when upstream issues are resolved. Could break again with a 2.12 release...

I guess what I'm saying is that I think dependencies with a high risk of breakage should probably be pinned indefinitely until there's long-term stability demonstrated across versions. The fact that a patch release (2.10.3) completely broke interactive plots for PyVista to me suggests there's some fragility with this dependency, and any future version compatibility should be dependent upon our interactive plot CI passing first.

PR #8327, opened 2026-02-13
Add `exclude_fields` option to `validate_mesh`

user27182:
The current `validation_fields` option is additive, which means it can be annoying if you want all fields _except_ for one or two. This PR adds `exclude_fields` as a subtractive alternative.

See also https://github.com/pyvista/pyvista/pull/8318#issuecomment-3892000014


user27182:
docstring annotations are a bit excessive now and not really user-friendly... I wonder if we can shrink this down

image

PR #8328, opened 2026-02-13
Remove assertions from `mesh_validation` example

user27182:
### Overview

Closes #8301

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -81,7 +81,7 @@ repos:
         exclude: pyvista/_warn_external.py
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.14
+    rev: v0.15.0
     hooks:
       - id: ruff-format
       - id: ruff-check
--- a/doc/source/make_tables.py
+++ b/doc/source/make_tables.py
@@ -3296,8 +3296,8 @@ class PointCloudCarousel(DatasetGalleryCarousel):
     @classmethod
     def fetch_dataset_names(cls):
         pointset_names = DatasetCardFetcher.fetch_dataset_names_by_datatype(pv.PointSet)
-        vertex_polydata_filter = (
-            lambda poly: isinstance(poly, pv.PolyData) and poly.n_verts == poly.n_cells
+        vertex_polydata_filter = lambda poly: (
+            isinstance(poly, pv.PolyData) and poly.n_verts == poly.n_cells
         )
         vertex_polydata_names = DatasetCardFetcher.fetch_and_filter(vertex_polydata_filter)
         return sorted(list(pointset_names) + list(vertex_polydata_names))
@@ -3312,9 +3312,8 @@ class SurfaceMeshCarousel(DatasetGalleryCarousel):
 
     @classmethod
     def fetch_dataset_names(cls):
-        surface_polydata_filter = (
-            lambda poly: isinstance(poly, pv.PolyData)
-            and (poly.n_cells - poly.n_verts - poly.n_lines) > 0
+        surface_polydata_filter = lambda poly: (
+            isinstance(poly, pv.PolyData) and (poly.n_cells - poly.n_verts - poly.n_lines) > 0
         )
         surface_polydata_names = DatasetCardFetcher.fetch_and_filter(surface_polydata_filter)
         return sorted(surface_polydata_names)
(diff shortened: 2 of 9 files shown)


akaszynski (commented):
Didn't even realize we can utilize:
```
# sphinx_gallery_start_ignore
```

Normally I prefer to keep testing and examples separate and not including logic that triggers only within CI. Still, it renders correctly at https://698f880b3ea9ca35af90a848--pyvista-dev.netlify.app/examples/99-advanced/mesh_validation

Ideally it would be possible to pull down the script without the ignore sections (the pulled script still contains the ignore sections), and even their docs https://sphinx-gallery.github.io/stable/configuration.html#hiding-lines-of-code state that it should be used sparingly.

Is it necessary to include asserts at all in the example? If we need to test for behavior, shouldn't it be within a unit test?

As an aside, I looked over the example again with fresh eyes. It's great. I'll be using this for mesh validation in a few of my other projects.

user27182:
> Didn't even realize we can utilize:
> 
> ```
> # sphinx_gallery_start_ignore
> ```
> 
> Normally I prefer to keep testing and examples separate and not including logic that triggers only within CI. Still, it renders correctly at https://698f880b3ea9ca35af90a848--pyvista-dev.netlify.app/examples/99-advanced/mesh_validation
> 
> Ideally it would be possible to pull down the script with the ignore sections (the pulled script still contains the ignore sections), and even their docs https://sphinx-gallery.github.io/stable/configuration.html#hiding-lines-of-code state that it should be used sparingly.
>
> I suppose is it necessary to include asserts at all in the example? If we need to test for behavior, shouldn't it be within a unit test?


Most of the assertions aren't necessary (e.g. `n_points == 1`), so we can probably get rid of those. The only ones that really matter are the ones for `report.invalid_fields`. This is because, as #8301 shows, the invalid fields reported differ across VTK versions, and even OS. And the example directly speaks to, and tries to explain, the exact errors that are reported. So if the invalid fields change... then the example stops making sense. This is why substantial changes had to be made to the example as part of #8228. And, if the assertions weren't there, the example would have silently "broken" when upgrading VTK.

And so the assertions aren't for behavior, we already have tests for this. It's really just to closely link the text to the code output. Ideally this example would simply be a regular docstring such that we could make use of doctest...

---

How about I get rid of the ignores and custom CI assertion, and get rid of most of the unnecessary asserts. And just keep one or two of the important `invalid_fields` asserts. To address https://github.com/pyvista/pyvista/pull/8301, we can just add VTK version and OS version guards to avoid assertion errors if users run it on older VTK. We can add comments so users are aware their output may differ and sort of explain why the assert is there (instead of trying to hide it).

This is probably OK, but I have grown to kind of distrust the output from `vtkCellValidator`, and I am still concerned there may be some kind of undetected "breakage" (in terms of the narrative, not in terms of code) if we remove some asserts... though if there's any meaningful change to `vtkCellValidator`, I'm fairly confident our tests will catch it. But whoever fixes/updates the tests will need to make a mental note to also check the `mesh_validation` example.

---

> As an aside, I looked over the example again with fresh eyes. It's great. I'll be using this for mesh validation in a few of my other projects.

I'm glad it's useful!

akaszynski:
> To address #8301, we can just add VTK version and OS version guards to avoid assertion errors if users run it on older VTK. We can add comments so users are aware their output may differ and sort of explain why the assert is there (instead of trying to hide it).

IMO, that's ideal. It makes it clear that there's variation in output due to the VTK version and OS. It's not ideal, better would be for `vtkCellValidator` to just work, but that's an upstream issue and we can't resolve everything unless we want to write a new cell validation library from scratch.

user27182:
> > To address #8301, we can just add VTK version and OS version guards to avoid assertion errors if users run it on older VTK. We can add comments so users are aware their output may differ and sort of explain why the assert is there (instead of trying to hide it).
> 
> IMO, that's ideal. It makes it clear that there's variation in output due to the VTK version and OS. It's not ideal, better would be for `vtkCellValidator` to just work, but that's an upstream issue and we can't resolve everything unless we want to write a new cell validation library from scratch.

Okay. I've decided to just remove all assertions and simply add a note.

akaszynski (approved):
LGTM. Thanks for adding the note. It's best to point out that VTK's cell validation is sometimes incorrect rather than obscuring it behind conditional checks.

PR #8343, opened 2026-02-19
Fix tox factors

edabor:
### Overview

From local testing, I saw that tox  `>=4.39` is causing those issues.
I suspect https://github.com/tox-dev/tox/pull/3753 being the root cause.

It does not occur for Linux CI since it uses a local pypi https://github.com/pyvista/pyvista/actions/runs/22188242201/job/64168341258?pr=8195#step:6:21

I'm investigating, but as an interim solution we can upper-pin `tox`

_Originally posted by @edabor in https://github.com/pyvista/pyvista/issues/8195#issuecomment-3928326284_

This PR is an attempt to fix compatibility with latest `tox`

### DIFF
--- a/pyvista/core/filters/data_set.py
+++ b/pyvista/core/filters/data_set.py
@@ -1593,6 +1593,12 @@ class DataSetFilters(_BoundsSizeMixin, DataObjectFilters):
         if origin is None or point_u is None or point_v is None:
             alg.SetAutomaticPlaneGeneration(True)
         else:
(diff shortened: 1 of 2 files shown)


edabor:
This has been fixed in https://github.com/tox-dev/tox/pull/3753 so this PR is not needed once latest tox is released

user27182:
Thanks @edabor. This is fixed with v4.42 https://github.com/tox-dev/tox/releases/tag/4.42.0

Issue #8340, opened 2026-02-17
Error using os.path.relpath across different drives on Windows,


jcapriot:
### Describe the bug, what's wrong, and what you expected.

In the _deprecate_positional_args decorator, you are using "os.path.relpath" to find the relative location between the calling site and the current working directory, which fails if the two locations are on different drives in Windows. This needs to be refactored to work regardless of the drive difference on Windows, or just remove this logic and use something closer to what `sklearn` has.

### Steps to reproduce the bug.

Install pyvista in an environment on one drive, say C:

Then startup a python console on another drive, say E:

```python

>>> import pyvista as pv
>>> pv.Cylinder([0, 0, -375], [0, 0, 1], radius=10, height=750)
ValueError ...
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Tue Feb 17 15:42:31 2026 Mountain Standard Time

                OS : Windows (11 10.0.26200 SP0 Multiprocessor Free)
            CPU(s) : 32
           Machine : AMD64
      Architecture : 64bit
               RAM : 127.7 GiB
       Environment : Jupyter
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : NVIDIA GeForce RTX 4080 SUPER/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 591.74
     Render Window : vtkWin32OpenGLRenderWindow
  MathText Support : False

  Python 3.12.6 | packaged by conda-forge | (main, Sep 22 2024, 14:01:26) [MSC
  v.1941 64 bit (AMD64)]

           pyvista : 0.47.0
               vtk : 9.3.1
             numpy : None
        matplotlib : 3.9.2
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 10.4.0
 typing-extensions : 4.12.2
          cyclopts : 4.5.2
           IPython : 8.27.0
             scipy : 1.14.1
           imageio : 2.36.0
             trame : 3.6.5
      trame-client : 3.3.2
      trame-server : 3.2.3
         trame-vtk : 2.8.10
     trame-vuetify : 2.7.1
        ipywidgets : 8.1.5
        jupyterlab : 4.2.5
--------------------------------------------------------------------------------
```

user27182:
Thanks for the report! I think formatting the warning is not so important, so I think an easy fix is to `try...except OSError` (or whatever error you're seeing) around the offending line.
If you can patch it and test this locally and submit a PR we can include the fix in a patch release.

PR #8345, opened 2026-02-20
Remove `relpath` from `deprecate_positional_args` decorator

jcapriot:
### Overview
Fix for #8340 

### Details
- `os.relpath` will fail on Windows when the two paths are on different drives.
- Changes the message to print out the full path to the offending file instead.

### DIFF
--- a/pyvista/_deprecate_positional_args.py
+++ b/pyvista/_deprecate_positional_args.py
@@ -4,7 +4,6 @@ from functools import wraps
 import inspect
 from inspect import Parameter
 from inspect import Signature
-import os
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import TypeVar


user27182:
Thanks for the fix! For reference I made this comment here https://github.com/pyvista/pyvista/pull/7732#pullrequestreview-3035399289 specifically to _keep_ the call to `os.path.relpath`. I recall having a hard time making this work correctly so that the call site was properly linked by IDEs and easy for users to update their code.

But I just tested the fix from this PR locally with a simple file containing `import pyvista as pv; pv.Sphere(0.5)` and the file path in the deprecation warning is correctly linked and working.

I think most of these issues were fixed by #8093, so much of the call site code may be redundant now. Not 100% sure though, and the positional arg deprecation decorator is only temporary anyway so not much sense in trying to update or improve it.

jcapriot:
Yeah, I think the rough part was just trying to make it relative to the current working directory.

user27182 (approved):

PR #8347, opened 2026-02-21
Update paraview named colors

user27182:
### Overview

ParaView grays were officially added to `vtkNamedColors` here: https://gitlab.kitware.com/vtk/vtk/-/merge_requests/12804

This PR:
- updates the current "paraview_background" color to match the new, official blue gray color
- adds a new "paraview_background_warm" color to match the new, official warm gray color

PR #8350, opened 2026-02-21
Use delimited color names

user27182:
### Overview

Color names are currently formatted so that there are no delimiters, e.g. `aliceblue` instead of `alice_blue`. But I find this style hard to read, especially for longer names like `darkslategray`, or the recently added `paraviewbackgroundwarm` from https://github.com/pyvista/pyvista/pull/8347.

I believe the lack of delimiter stems from copying the convention from `matplotlib` color names, which are adopted from a CSS standard https://www.w3.org/TR/css-color-4/#named-colors. Although this convention is "standard", the more important thing about the standard IMO is the actual colors used, and not necessarily strict adherence to the name convention. For example, [vtkNamedColors](https://htmlpreview.github.io/?https://github.com/Kitware/vtk-examples/blob/gh-pages/VTKNamedColorPatches.htm) uses the standard colors, but the names use CamelCase _and_ snake_case aliases.

This PR changes all official names of our named colors so that they now use underscore delimiters.
- `pv.hexcolors` is deprecated and replaced with `pv.hex_colors`.
- The output from `pv.Color.name` now returns a name with delimiters. This is a breaking change for _some_ colors, i.e. any colors that use more than one word
- The colors in our "Named Colors" documentation is updated to include delimiters

Importantly, this PR does _not_ change the allowed named color inputs. Named colors may still contain capital letters and/or delimiters or no delimiters at all. The only breaking change is with the value returned by the `Color.name` getter.

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -179,6 +179,7 @@ _ExtractSurfaceOptions = Literal['geometry', 'dataset_surface', None]  # noqa: P
 _NestedStrings = str | Sequence['_NestedStrings']
 
 _ActionOptions = Literal['warn', 'error']
+_ReportBodyOptions = Literal['fields', 'message']
 _DataFields = Literal[
     'cell_data_wrong_length',
     'point_data_wrong_length',
@@ -402,6 +403,7 @@ class _MeshValidator(Generic[_DataSetOrMultiBlockType]):
             _mesh=validated_mesh,
             _message=message_structure,
             _subreports=None,
+            _report_body=None,
             **dataclass_fields,  # type: ignore[arg-type]
         )
 
@@ -459,6 +461,7 @@ class _MeshValidator(Generic[_DataSetOrMultiBlockType]):
             _mesh=validated_mesh,
             _message=message,
             _subreports=tuple(reports),
+            _report_body=None,
             **dataclass_fields,  # type: ignore[arg-type]
         )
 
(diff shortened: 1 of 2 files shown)


user27182:
Annotations from #8349 is also ready but depends on this PR.

akaszynski:
Main issue I see with this is we will now deviate from `matplotlib`'s named colors (see https://matplotlib.org/stable/gallery/color/named_colors.html). I agree that `"light coral"` reads better than `"lightcoral"`, but I'd prefer to allow both. Honestly, I'm not sure why it's not allowed in matplotlib (PR maybe?).

Would we be open to allowing both?

user27182:
> Main issue I see with this is we will now deviate from `matplotlib`'s named colors (see https://matplotlib.org/stable/gallery/color/named_colors.html). I agree that `"light coral"` reads better than `"lightcoral"`, but I'd prefer to allow both. Honestly, I'm not sure why it's not allowed in matplotlib (PR maybe?).
> 
> Would we be open to allowing both?

We already allow both. We allow `'LightCoral'`, `'lightcoral'`, `'light_coral'`, even `'light-coral'` and `'light coral'`. We format the inputs with this function:
https://github.com/pyvista/pyvista/blob/1ccf9efef77630800bd7a1ca69a34089321c97a0/pyvista/plotting/colors.py#L52-L58

The main change from this PR is the value _returned_ by `Color.name`. The name returned now contains underscores between words.

akaszynski:
Sorry, didn't read:

> Importantly, this PR does not change the allowed named color inputs.

I'm fine with this change. Since we already mutate the `Color.name`:

```py
>>> import pyvista as pv

>>> pv.Color('light blue')
Color(name='lightblue', hex='#add8e6ff', opacity=255)
```

we might as well mutate it to something more readable.

PR #8353, opened 2026-02-22
Add `report_body` keyword to `validate_mesh`

user27182:
### Overview

Select contents to show inside the body of the validation report.

The new option to show the message in the report body is superior IMO as it's more compact _and_ more informative. I will make a follow-up PR #8354 to change the default.


user27182:
Both #8354 and #8355 are also ready, but depend on this PR.

akaszynski (approved):

PR #8344, opened 2026-02-20
Add `validate` command to PyVista CLI

user27182:
### Overview

CLI version of `validate_mesh` filter.

PR #8355, opened 2026-02-22
Add `--report` option to `validate` CLI command

user27182:
### Overview

Follow-up to #8353 , and make CLI output minimal by default (see comment #8344).


tests/test_cli.py
```diff
@@ -296,6 +307,11 @@ def test_convert_help(capsys: pytest.CaptureFixture):
 def test_validate(tmp_ant_file: Path, capsys: pytest.CaptureFixture):
     main(f'validate {str(tmp_ant_file)!r}')
     out = capsys.readouterr().out
+    expected = "PolyData mesh 'ant.ply' is valid!\n"
+    assert out == expected
+
+    main(f'validate {str(tmp_ant_file)!r} -r')
+    out = capsys.readouterr().out
```
user27182, line 314: I am thinking of changing this API

This PR makes it so that simple output is the default ("mesh is valid!" if valid, or just the validation error message if invalid), and users need to use `--report` explicitly to see more details. But more detail probably makes by default, especially with #8354 which makes the output much less verbose. This is primarily a diagnostic tool after all...

So perhaps we should add `--quiet` instead for the simple case?

user27182: After thinking about this a bit more, I like the status quo. If running `pyvista validate file` and file is valid, we get a green "Mesh is valid" message. Very simple and clear. No need to clutter it with mesh meta data and a validation summary that only tells you that it's valid (i.e. no new info).

And if the validation fails, we already get a helpful message, which by itself _already_ summarizes the report status with any/all invalid fields. The mesh metadata is mostly just a bonus diagnostic.

Issue #8252, opened 2026-01-20
VTK 9.6 regression with plotting ambiguous scalars


user27182:
Related to: https://github.com/pyvista/pyvista/issues/8228#issuecomment-3768968171

With VTK 9.6, rendered colors may be incorrect when there are both point and cell scalars with the same name. Not sure if it's a mapper issue, could be a VTK regression or an issue with PyVista, but I haven't determined exactly what the cause is.

For example, the connectivity filter generates both point and cell data `RegionId` scalars, and with vtk 9.6.0rc2 this is now broken:
``` py
import pyvista as pv
large = pv.Sphere(
    center=(-4, 0, 0), phi_resolution=40, theta_resolution=40
)
medium = pv.Sphere(
    center=(-2, 0, 0), phi_resolution=15, theta_resolution=15
)
small = pv.Sphere(center=(0, 0, 0), phi_resolution=7, theta_resolution=7)
mesh = large + medium + small
conn = mesh.connectivity('all')
conn.plot(cmap=['red', 'green', 'blue'], show_edges=True)
```

<img width="800" height="870" alt="Image" src="https://github.com/user-attachments/assets/4e29a112-c225-4fc1-af18-1ed01cec604b" />

But if we do `conn.point_data.clear()` before plotting, then it renders as expected. Similarly, using the keyword `preference='cell'` with `plot` also fixes this.

As a workaround, the docs for connectivity were modified to use `color_labels` for plotting in https://github.com/pyvista/pyvista/pull/8251, but we should still try to identify/fix the underlying issue.

cquammen:
I could not reproduce with vtk==9.6.0 and pyvista 0.47.0.

cquammen:
However, with vtk==9.6.0-RC2, I _was_ able to reproduce the bug. So it appears to be fixed in the vtk 9.6.0 final release.

cquammen:
And just for fun, it's also working fine with vtk==9.6.0-RC3. Looking through the commits between VTK 9.6.0-RC2 and VTK 9.6.0-RC3, I don't see anything obvious that would change the behavior. But I suggest closing this issue once you confirm that it works as expected with VTK 9.6.0.

user27182:
Thanks for checking! Confirmed it's no longer an issue.

PR #8356, opened 2026-02-22
Add connectivity cmap test

user27182:
### Overview

Add test for failure described in https://github.com/pyvista/pyvista/issues/8252

Since we modified our docs example in #8251, this case is now no longer tested, which means there is no test coverage for it in either PyVista or VTK. So it seems wise to include it here, especially since the exact source of this bug is unknown. `connectivity` uses implicit data arrays to compress the output dtype though, so I suspect it may be related to this.

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -179,6 +179,7 @@ _ExtractSurfaceOptions = Literal['geometry', 'dataset_surface', None]  # noqa: P
 _NestedStrings = str | Sequence['_NestedStrings']
 
 _ActionOptions = Literal['warn', 'error']
+_ReportBodyOptions = Literal['fields', 'message']
 _DataFields = Literal[
     'cell_data_wrong_length',
     'point_data_wrong_length',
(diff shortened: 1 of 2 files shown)


tests/plotting/test_plotting.py
```diff
@@ -5376,3 +5376,14 @@ def test_hide_cells_no_scalars():
...
+
+@pytest.mark.skip_check_gc
+def test_connectivity_cmap():
+    # Test case described in https://github.com/pyvista/pyvista/issues/8252
+    large = pv.Sphere(center=(-4, 0, 0), phi_resolution=40, theta_resolution=40)
+    medium = pv.Sphere(center=(-2, 0, 0), phi_resolution=15, theta_resolution=15)
+    small = pv.Sphere(center=(0, 0, 0), phi_resolution=7, theta_resolution=7)
+    mesh = large + medium + small
+    conn = mesh.connectivity('all')
+    conn.plot(cmap=['red', 'green', 'blue'], show_edges=True)
```
tkoyama010:
```suggestion
    connected = mesh.connectivity('all')
    connected.plot(cmap=['red', 'green', 'blue'], show_edges=True)
```

PR #8357, opened 2026-02-23
Enable pretty printing with mesh validation report

user27182:
### Overview
TOP: Regular report
BOTTOM: Pretty report
image


user27182:
This seems way too complex with lots of branching and produces different reports with different formatting/spacing. But I think simple styling with color/bold is sufficient.

And `cyclopts` _already_ enables some simple formatting just by using `app.console.print` instead of python's `print`

image

I think we can instead try to replace headings with formatted text _after_ generating the report, rather than formatting it pre-emptively with `rich` inside of `validate_mesh`.

PR #8364, opened 2026-02-24
Allow excluding fields from a subset of mesh validation fields

user27182:
### Overview

Currently on `main`, this is not valid:
``` sh
pyvista validate cow.vtp cells -e non_convex
```
but should be, because `non_convex` is a subset of `cells`. This PR fix this.

### DIFF
--- a/tests/core/test_dataobject_filters.py
+++ b/tests/core/test_dataobject_filters.py
@@ -1622,7 +1622,7 @@ def test_validate_mesh_exclude_fields_subset(invalid_tetra_negative_volume):
     expected = (
         'Mesh Validation Report\n'
         '----------------------\n'
-        'Mesh:\n'
+        'Mesh info:\n'
         '    Type                     : UnstructuredGrid\n'
         '    N Points                 : 4\n'
         '    N Cells                  : 1\n'


tests/core/test_dataobject_filters.py
```diff
@@ -1572,10 +1572,66 @@ def test_validate_mesh_exclude_fields():
...
+    )
+    assert expected in report
+    assert 'Negative size' not in report
+
+    fields = ['data', 'points', NEGATIVE_SIZE]
+    report = str(mesh.validate_mesh(fields, exclude_fields=fields, report_body='fields'))
+    expected = (
+        'Mesh Validation Report\n'
+        '----------------------\n'
+        'Mesh:\n'
```
user27182: This PR is ready for review but this will likely break if #8358 is merged first.




```suggestion
        'Mesh info:\n'
```

akaszynski: Saw this only now.

user27182: Fixed in [b6e6737](https://github.com/pyvista/pyvista/pull/8364/commits/b6e67372ada21333dc7f19b7f2d247295e029f29)

Issue #5449, opened 2024-01-10
Inexact clipping with clip_surface() for polydata surfaces


jesshe:
### Describe the bug, what's wrong, and what you expected.

I've been trying to clip a dipping surface where it intersects with another vertical surface (red surface in the example below). However, the clipped surface doesn't appear to exactly match the intersection between the original and clipping surfaces. As shown in the screenshot,  Wondering if there's a way to get an exact clip between these two surfaces?

![image](https://github.com/pyvista/pyvista/assets/44711989/177ceeb1-8779-4933-b602-5ec943c79d80)

### Steps to reproduce the bug.

```
import pyvista as pv

pts1 = [
        [6.8426375e+05, 5.6135830e+06, 0.0000000e+00],
        [6.8454338e+05, 5.6133550e+06, 0.0000000e+00],
        [6.8526319e+05, 5.6129250e+06, 0.0000000e+00],
        [6.8621606e+05, 5.6125785e+06, 0.0000000e+00],
        [6.8708600e+05, 5.6123970e+06, 0.0000000e+00],
        [6.8426375e+05, 5.6135830e+06, 1.2000000e+03],
        [6.8454338e+05, 5.6133550e+06, 1.2000000e+03],
        [6.8526319e+05, 5.6129250e+06, 1.2000000e+03],
        [6.8621606e+05, 5.6125785e+06, 1.2000000e+03],
        [6.8708600e+05, 5.6123970e+06, 1.2000000e+03]]
surf1 = pv.PolyData(pts1)
surf1 = surf1.delaunay_2d()

pts2 = [
        [6.84100040e+05, 5.61519935e+06, 1.19900000e+03],
        [6.85188690e+05, 5.61625085e+06, 1.00000000e+00],
        [6.84811730e+05, 5.61400653e+06, 1.19900000e+03],
        [6.85900380e+05, 5.61505802e+06, 1.00000000e+00],
        [6.85000390e+05, 5.61378254e+06, 1.19900000e+03],
        [6.86089040e+05, 5.61483401e+06, 1.00000000e+00],
        [6.85051610e+05, 5.61365417e+06, 1.19900000e+03],
        [6.86140260e+05, 5.61470564e+06, 1.00000000e+00],
        [6.85079010e+05, 5.61327472e+06, 1.19900000e+03],
        [6.86167650e+05, 5.61432621e+06, 1.00000000e+00],
        [6.85073290e+05, 5.61343298e+06, 1.19900000e+03],
[...]

PR #8368, opened 2026-02-26
Add `points_dtype` to control the dtype filters produce

### LINKED ISSUE #3161: Translate not handling float to the required precision
In the given example the trans.center should be 2.0, 1.0, 2,0

```python
import pyvista
mesh = pyvista.Sphere()
mesh.center
trans = mesh.translate((2, 1, 2), inplace=True)
trans.center

```

The result from the example is:
`[1.9999999403953552, 1.0000000298023224, 2.0]`

I came across this when trying to relocate some surfaces which must perfectly align, I translated them so their centre was at the zero point, that worked, then I translated them to `trans = trans.translate((991.0000000000000, 262.2000000000000, 193.2000000000000), inplace=True)` and it gives the new center as `[990.8000183105469, 262.260009765625, 193.0800018310547]`

### LINKED ISSUE #5411: Precision issues with big number using MultipleLines

I have a line with big coordinates. Then I tried to refine it with more nodes - say 100 nodes. Then I found a strange thing that the refined line is not straight if I created the line with `pyvista.MultipleLines`. 
However, if I created the line with` pyvista.PolyData`, everything is OK. I compaired the points of 2 lines, I think the bug is that when creating a line by `pyvista.MultipleLines`, there is some precision issue when using big coordinates.

Here is a sample code
```python
import pyvista as pv
import numpy as np

def refine_line(start_point, end_point, number_of_nodes:int) -> list:
    """Used to refine the line with 2 points."""
    end_point = end_point
    start_point = start_point
    length = np.sqrt((end_point[0] - start_point[0])**2 + (end_point[1] - start_point[1])**2 + (end_point[2] - start_point[2])**2)
    segment_length = length/(number_of_nodes-1)
    new_points = [start_point]
    total_delta_x = end_point[0] - start_point[0]
    total_delta_y = end_point[1] - start_point[1]
    total_delta_z = end_point[2] - start_point[2]
    for i in range(number_of_nodes):
        if i == 0:
            continue
        delta_x = (i*segment_length)/length*total_delta_x
        delta_y = (i*segment_length)/length*total_delta_y
        delta_z = (i*segment_length)/length*total_delta_z
        new_point = [start_point[0]+delta_x, start_point[1]+delta_y, start_point[2]+delta_z]
        new_points.append(new_point)
    new_points.append(end_point)
    return new_points

start_point = [630567.4265653,4272091.976505,50]
end_point = [630649.9007899,4272053.7930221,50]

# Original line
line = pv.Line(start_point, end_point)
line.plot()

# Refine the line
new_points = refine_line(start_point, end_point, 100)
print(np.array(new_points))

# Create by MultipleLines
refined_line = pv.MultipleLines(new_points)
print(refined_line.points)
refined_line.plot()

# Create by PolyData
lines = np.arange(len(new_points))
lines = np.hstack([len(new_points), lines])
poly_refined_line = pv.PolyData(new_points, lines=lines)
print(refined_line.points)
poly_refined_line.plot()
```

### System Information

```shell
OS : Windows
            CPU(s) : 16
           Machine : AMD64
      Architecture : 64bit
               RAM : 15.8 GiB
       Environment : Python
        GPU Vendor : Intel
       GPU Version : 4.5.0 - Build 31.0.101.1999
  MathText Support : False

  Python 3.11.5 | packaged by conda-forge | (main, Aug 27 2023, 03:23:48) [MSC
  v.1936 64 bit (AMD64)]

           pyvista : 0.42.3
               vtk : 9.2.6
             numpy : 1.24.3
        matplotlib : 3.7.1
            scooby : 0.7.2
             pooch : v1.7.0
            pillow : 10.0.0
           imageio : 2.31.2
           IPython : 8.14.0
             scipy : 1.11.1
              tqdm : 4.66.1
            meshio : 5.3.4
      nest_asyncio : 1.5.6

  Intel(R) oneAPI Math Kernel Library Version 2023.1-Product Build 20230303
```

### Screenshots

Original line with 2 points
image
Refined line with 100 points created by MultipleLine
image
Refined line with 100 points created by PolyData
image

### LINKED ISSUE #5610: Add `precision` param for geometric objects to support 64bit float point meshes

The default precision for vtk source objects is 32-bit. However, floats in python and numpy are 64-bit by default. This can create many precision related-issues, and the greater precision is likely desirable for computations. 

See related issues #5608, #5411, #3161

This feature could leverage the `SetOutputPrecision` method (e.g. see docs for [vtkPlaneSource](https://vtk.org/doc/nightly/html/classvtkPlaneSource.html#a9c2ad37502fd2e0fc0a132403c25fdd8))

### Pseudocode or Screenshots

_No response_

### LINKED ISSUE #5690: Problem rendering points with glyphs (dtype precision issue?)

When rendering one or more points with x,y, or z > ~1e6 using a glyph, the result is severely distorted (in the case of a sphere, it's definitely not a sphere, as you can see in the example below! I've also tried using an arrow, which also results in distortion). Checking the data type of the corresponding point(s) reveals that they're float32 rather than the expected float64, and this loss of precision might explain the distortion for such large numbers. Note that this holds even when running `points_to_double()`. 

```python
import pyvista as pv
m = pv.PolyData([[1e7, 1, 1]])
sphere=pv.Sphere(radius=1)
m = m.glyph(geom=sphere, orient=False, scale=False)
m.plot()
```
^Experiment with the above by replacing 1e7 with values between ~1e5 and 1e8.
```python
m = pv.PolyData([[1e7, 1, 1]])
sphere=pv.Sphere(radius=1)
sphere.points_to_double()
m = m.glyph(geom=sphere, orient=False, scale=False)
print(m.points.dtype)
```

### System Information

Screenshot 2024-02-26 at 10 41 12 PM

### Screenshots

Screenshot 2024-02-26 at 10 39 03 PM

### LINKED ISSUE #6675: Accuracy issue with `PolyDataFilters.extrude`

`PolyDataFilters.extrude` can return a mesh with a significant error on the bounds after extrusion, depending on the extrusion length.

It looks like the error comes from the output of `PolyDataFilters.extrude` (maybe it's more of a VTK issue). 

```python
micro=pv.Rectangle().extrude((0, 0, 1.5), capping=False)
print( micro.bounds[-1])
```
> 1.5

```python
micro=pv.Rectangle().extrude((0, 0, 1.6), capping=False)
print( micro.bounds[-1])
```
> 1.600000023841858

### System Information

```shell
  Date: Tue Sep 24 21:59:56 2024 CEST

                OS : Linux (Ubuntu 24.04)
            CPU(s) : 12
           Machine : x86_64
      Architecture : 64bit
               RAM : 15.1 GiB
       Environment : Jupyter
       File system : ext4
        GPU Vendor : Intel
      GPU Renderer : Mesa Intel(R) Graphics (ADL GT2)
       GPU Version : 4.6 (Core Profile) Mesa 24.3
  MathText Support : True

  Python 3.12.3 (main, Sep 11 2024, 14:17:37) [GCC 13.2.0]

           pyvista : 0.45.dev0
               vtk : 9.3.1
             numpy : 1.26.4
        matplotlib : 3.6.3
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 10.2.0
           imageio : 2.35.1
             PyQt5 : 5.15.10
           IPython : 8.27.0
          colorcet : 3.1.0
           cmocean : 4.0.3
        ipywidgets : 8.1.5
             scipy : 1.14.1
              tqdm : 4.66.5
            meshio : 5.3.5
        jupyterlab : 4.2.5
    pytest_pyvista : 0.1.8
             trame : 3.6.3
      trame_client : 3.2.5
      trame_server : 3.1.2
         trame_vtk : 2.8.10
     trame_vuetify : 2.7.0
      nest_asyncio : 1.6.0
```

### Screenshots

_No response_

### LINKED ISSUE #6685: Sphere loses detail after coordinates exceed 2^16

I'm planning to use pyvista for visualizing 3d objects which may have very large coordinates due to the coordinate system I am using. I found that when shapes are too far from the origin (approx beyond (2^16,2^16,2^16)), the geometries start losing detail. I suspect its because of floating point precision? What is the prefered way of handling large coordinates?

```python
import pyvista as pv
import numpy as np
n = 16
# Create the sphere with double-precision points
sphere = pv.Sphere(radius=1.0, center=(2**n, 2**n, 2**n))
# this does nothing
sphere.points = sphere.points.astype(np.float64)

# Create the plotter and add the sphere
plotter = pv.Plotter()
plotter.add_mesh(sphere, color='red', opacity=0.6, show_edges=True, label='Sphere')

# Show the plot
plotter.show()

```

### System Information

```shell
  Date: Wed Oct 02 12:56:20 2024 HKT

                OS : Linux
            CPU(s) : 24
           Machine : x86_64
      Architecture : 64bit
               RAM : 31.1 GiB
       Environment : Python
       File system : ext4
      GPU Renderer : NVIDIA GeForce RTX 3080/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 535.183.01
  MathText Support : False

  Python 3.8.10 (default, Sep 11 2024, 16:02:53)  [GCC 9.4.0]

           pyvista : 0.44.1
               vtk : 9.3.1
             numpy : 1.24.4
        matplotlib : 3.1.2
            scooby : 0.10.0
             pooch : 1.8.2
            pillow : 10.0.1
           imageio : 2.31.3
             PyQt5 : 5.14.1
           IPython : 8.12.2
        ipywidgets : 8.1.0
             scipy : 1.10.1
              tqdm : 4.66.1
            meshio : 5.3.5
      nest_asyncio : 1.5.6
```

### Screenshots

image
image
image
image
and at `n=24` it just degenerates to a cube
image

### LINKED ISSUE #7742: Sources like SphereSource need double precision

### LINKED ISSUE #8166: Shrink filter reduce the precision of point array

Hy,

I discover that shrink method reduce the precision of point array.
The following simplified test case reproduce the behavior observed on my laptop.
I expect that the grid.point array generated by shrink operation would have the same level of precision than the input grid which is not the case.
Is it a bug ?
If not is it possible to add an additional parameter to triggered it?

Best regards,

Image

```python
import pyvista as pv

grid = pv.examples.cells.Hexahedron()
print(grid.points.dtype)
new_grid = grid.shrink(shrink_factor=1.)
print(new_grid.points.dtype)
```

float64
float32

### System Information

```shell
  Date: Tue Dec 09 23:05:06 2025 CET

                OS : Linux (Ubuntu 24.04)
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 15.5 GiB
       Environment : Python
       File system : ext4
        GPU Vendor : Intel
      GPU Renderer : Mesa Intel(R) UHD Graphics 620 (WHL GT2)
       GPU Version : 4.6 (Core Profile) Mesa 25.0.7-0ubuntu0.24.04.2
  MathText Support : True

  Python 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0]

           pyvista : 0.46.4
               vtk : 9.4.2
             numpy : 2.2.5
        matplotlib : 3.10.3
            scooby : 0.10.1
             pooch : 1.8.2
            pillow : 11.2.1
             PyQt5 : 5.15.11
           IPython : 9.4.0
        ipywidgets : 8.1.7
             scipy : 1.15.3
            meshio : 5.3.5
        jupyterlab : 4.4.4
      nest_asyncio : 1.6.0
```

### Screenshots

_No response_

user27182:
Resolves #8166, resolves #5610, resolves #3161, resolves #5411, resolves #5690, resolves #6685, closes #7742, and closes #6675.

VTK algorithms decide the `points` dtype for themselves and disagree about it: on a double-precision mesh 27 filters hand back single precision, and on a single-precision mesh 4 hand back double. `pyvista.global_config.points_dtype` lets a session decide instead, taking `'preserve'` to keep whatever dtype went in, or `'float32'`/`'float64'` to fix one everywhere, sources included. It is enforced where PyVista wraps an algorithm's output, so no filter grows a precision keyword.

The default is `None`, which changes nothing. That is deliberate: a dtype can always be delivered and the precision behind it cannot, so every widening cast warns with a new `PrecisionWarning` naming whatever generated the points, and `warnings.filterwarnings('error', ...)` turns that into an exception. Narrowing does not warn.

- adds `points_to_single`, the counterpart to `points_to_double`
- `Cube`, `CubeSource` and `CubeFacesSource` take `points_dtype`; the old `point_dtype` spelling is deprecated
- `'preserve'` is constrained only by meshes that store their points, so `ImageData` and `RectilinearGrid` filters are unchanged
- rides along: the doctest-artifact `.gitignore` entry and the `test_to_tetrahedral*` renames

Changes drafted by Claude Opus 5 but fully understood by me.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)


user27182:
The handling of dtype precision by filters is definitely a bit of a mess. Not only does it depend on the filter itself to support it, but even that's not a guarantee. I am finding that different mesh types are also handled differently, and some with have a double -> double mapping , whereas others (StructuredGrid) will get converted to single. There are also edge cases for empty inputs.

This means it'll be super important _not_ to ignore this error inside any filter implementations. Instead, we'll need to handle this before calling the filter, so that warnings can be properly emitted for users to see.

akaszynski:
> The handling of dtype precision by filters is definitely a bit of a mess.

The handling of precision by VTK in general is a mess...

user27182:
One function is worth a closer look than the rest: `_points_dtype` in `pyvista/core/filters/__init__.py`. It decides the whole policy, and two defects of the same shape have already come out of it, both from reading `mesh.points` and trusting whatever came back.

- `ImageData.points` is a lazy structured array on VTK 9.4+, so reading it just to learn a dtype materialized it. That OOM'd on a `voxelize_binary_mask` reference volume trying to allocate 25.8 GiB of meshgrid ([job](https://github.com/pyvista/pyvista/actions/runs/33142503492/job/98756288110)).
- `pv.PolyData(..., force_float=False)` keeps integer points, and echoing that dtype truncated a filter's interpolated output back onto them. `ruled_surface` collapsed its surface onto the four integer corners of its input ([job](https://github.com/pyvista/pyvista/actions/runs/33146499845/job/98771657603)). Its unit test asserted only `n_cells` and passed; the docs image test caught it.

Both are fixed, and the function now answers from a closed set rather than from whatever the input reports: `vtkPointSet` inputs only, `float32`/`float64` only, everything else left to VTK. I do not see a third case, but the previous two were not obvious from the diff either, so a second pair of eyes on that one function is worth more than on the rest of the change put together.

user27182:
I went looking for the other precision issues this touches. Six are fixed, and I ran each reporter's own reproducer to check rather than going by the title. All figures are with `pyvista.global_config.points_dtype = 'float64'`.

| Issue | Reported | Now |
| --- | --- | --- |
| #8166 | `shrink` turns double points single | dtype preserved; `'float64'` reports that `vtkShrinkFilter` cannot compute in double |
| #5610 | want 64-bit points from geometric objects | every source honors the setting |
| #3161 | `translate((2, 1, 2))` gives `(1.9999999404, 1.0000000298, 2.0)` | exactly `(2.0, 1.0, 2.0)` |
| #5411 | `MultipleLines` bends with UTM-scale coordinates | straightness deviation `3.4e-01` → `2.3e-10`, matching the `PolyData` reference in the report |
| #5690 | glyph at `x=1e7` is visibly not a sphere | max radial error `3.2e-01` → `8.2e-10` |
| #6685 | sphere loses detail past `2**16` | max radial error `4.5e-03` → `9.9e-12` |

Two more are closed here without being fixed in the usual sense, because this is as far as PyVista can take them:

- #7742 asked for sources to be double *by default*. The default stays `None`, so nothing changes for existing code, but `pyvista.global_config.points_dtype = 'float64'` now gets exactly the result it wanted -- the UTM sphere's max radial error goes `2.3e-01` → `4.1e-10` -- and it can be set once for a session.
- #6675 is the same VTK limitation as #8166, just measured differently. `vtkLinearExtrusionFilter` has no `SetOutputPointsPrecision`, so it emits a float32 points array even when handed float64 input, exactly as `vtkShrinkFilter` does. PyVista restores the dtype and cannot restore the digits, so `extrude` still returns `1.600000023841858`, which is `float32(1.6)` exactly. The extrusion vector is not the problem -- `GetVector()` gives back `1.6` unchanged. #8166 reported the dtype changing, which is fixed; #6675 reported the value being wrong, which needs the filter to gain the setter upstream. What does change here is that it is no longer silent: opting in warns that this filter could not deliver the precision asked for.

Left open: #5449. Inexact `clip_surface` measures the same 1350.9 m off the clip surface under every setting, so it is algorithmic rather than a precision problem.

akaszynski (approved):
I think this is a great approach to addressing the dtype problem. I like how it addressed the (what would be) silent `float64` --> `float32` --> `float64` loss of precision, which this PR had me worried about.

akaszynski:
Weird:
```
FAILED tests/core/test_utilities.py::test_sample_function_raises - NotImplementedError: cannot instantiate 'WindowsPath' on your system
```
on: https://github.com/pyvista/pyvista/actions/runs/34043027187/job/101513080944

Seen this before @user27182?

user27182:
Fixed in #9124

PR #8370, opened 2026-02-27
Use `getattr_static` instead of dynamic `getattr` lookup for checking new attributes

user27182:
### Overview
The `_NoNewAttrMixin` currently has undesirable side effects for child classes.
For example, this code is only _setting_ an attribute's value (`obj.foo = 42`), but it triggers a dynamic call to `getattr` (via `hasattr`), which means setting an attribute also triggers _getting_ the same attribute. Getting when setting is the undesirable side effect.

``` py
from pyvista.core.utilities.misc import _NoNewAttrMixin
class Parent(_NoNewAttrMixin):
    @property
    def foo(self):
        raise RuntimeError
    @foo.setter
    def foo(self, val):
        ...

class Child(Parent):
    ...

obj = Child()
obj.foo = 42

```

The issue is with this line:
https://github.com/pyvista/pyvista/blob/f1441b03ebc478f6b1f31d69700cf74ed08ea5c8/pyvista/core/utilities/misc.py#L344
And it only triggers for child classes because the first condition is `True` for parent classes, so `hasattr` (and hence `getattr`) is never called.

This side effect is also what created issues for subclasses emitting obbTree warnings upon deletion (see https://github.com/pyvista/pyvista/issues/8204).

This PR fixes this by using a static lookup instead.


tests/core/test_utilities.py
```diff
@@ -1498,6 +1498,22 @@ def test_no_new_attr_mixin(no_new_attributes_mixin_subclass):
...
+            raise RuntimeError
+
+        @foo.setter
+        def foo(self, val): ...
+
+    class Child(Parent): ...
+
+    # Test that setting attributes on child classes does not trigger a call to the getter
+    obj = Child()
+    obj.foo = 42
```
akaszynski, line 1520: I suppose you could have used a sentinel value rather than a `RuntimeError` to track if it's called to be more explicit (something like `assert not_called[0]`). Still, this works.

akaszynski (approved):
PR is good to go as is.

It's a good catch. I didn't think about the side effect that `getattr` executes properties.

pyvista/core/utilities/misc.py
```diff
@@ -305,6 +306,15 @@ class _AutoFreezeABCMeta(_AutoFreezeMeta, ABCMeta):
     """Metaclass to combine automatic attribute freezing with ABC support."""
 
 
+def _hasattr_static(obj: Any, attr: str) -> bool:
+    """Replicate behavior of hasattr using static lookup."""
+    try:
+        inspect.getattr_static(obj, attr)
```
user27182: Since this called so many times maybe we should micro optimize this and import from inspect instead to save a lookup 
```suggestion
        getattr_static(obj, attr)
```

PR #8382, opened 2026-03-06
Add optional `'Size'` array to `compute_cell_sizes`

user27182:
### Overview

New `'Size'` array generated to combine output from all other size arrays.

This helps simplify the logic in the `cell_validator` filter.


pyvista/core/filters/data_object.py
```diff
@@ -4110,26 +4058,46 @@ def compute_cell_sizes(  # type: ignore[misc]  # noqa: PLR0917
         volume: bool = True,  # noqa: FBT001, FBT002
         progress_bar: bool = False,  # noqa: FBT001, FBT002
         vertex_count: bool = False,  # noqa: FBT001, FBT002
+        size: bool = False,  # noqa: FBT001, FBT002
```
user27182, line 4061: I can remove this from the public cell sizes API if preferred. And just compute the size as an internal array inside of cell validator.

This change would effectively make this PR a refactor instead of a new feature.

user27182:
Superseded by #8438

Issue #1260, opened 2021-04-12
pyvista doesn't show the plot the 2nd time it is run in a script


maxcapodi78:
When running a python script which plot different  3D models pyvista plots it correctly the first time you run the code. the 2nd time the image quickly shows and then close it automatically

-----

**To Reproduce**

```py
import pyvista as pv
def show():
    sphere = pv.Sphere()
    plotter = pv.Plotter()
    plotter.add_mesh(sphere)
    plotter.show()
show()
show()
```

tkoyama010:
Could you run the following code wherever you are experiencing the bug and paste the output below. This report helps us track down bugs and it is critical to addressing your bug:

```py
# Get system info
import pyvista as pv
print(pv.Report())
```

akaszynski:
@maxcapodi78, I'll get on this today.  It's been a busy week already, but I'll try to get this out.  I've been able to duplicate this locally on windows as well.

akaszynski:
See #186, #1018 and #1078 for other issues related to this one.  This is a reoccurring issue and is cited in `plotter.py`

akaszynski:
Fixed in #1266.  Release 0.30.0 might not be out by the end of the week, so we're releasing 0.29.1 since this makes `pyvista` unusable beyond one plot on windows.

u-anurag:
I still have issues with the plot window. The plot shows up and the window closes instantly. Here is the example,


```python
import pyvista as pv
p = pv.Plotter()
box = pv.Cube(bounds=(-3,3,-5,6,50,75))
p.add_mesh(box, color=True,show_edges=True)
p.show()
```


I have already updated the installation to v0.29.1. Here is the warning,

![image](https://user-images.githubusercontent.com/35711550/116119772-054cb200-a67c-11eb-94ce-780b79a056fe.png)

Here is the report outcome using

```python
import pyvista as pv
print(pv.Report())
```

![image](https://user-images.githubusercontent.com/35711550/116119935-34fbba00-a67c-11eb-906a-34b07b7b4687.png)

The window works fine when I comment out lines 4390 and 4391 in the `plotting.py` file.

akaszynski:
This might be an issue with `vtk==8.1.2`.  I'm assuming you're using the one available from anaconda that's not up on pypi.  Can you check on 8.1.2 or 9.0?  I'll know how to attack this going forward.

Opening until this is resolved.

u-anurag:
[...]

PR #8384, opened 2026-03-07
Fix Windows interactor unresponsiveness after focus changes

### LINKED ISSUE #8383: Windows: Plotter.show() window becomes unresponsive after backgrounding (focus loss), even with minimal sphere repro
### Summary
On Windows, [pyvista.Plotter.show()](vscode-file://vscode-app/c:/Users/jason/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) opens and renders normally, but the render window becomes unresponsive after it is sent to background (loses focus) and then brought back to foreground.

This reproduces with a minimal single-sphere script and also with structured-grid terrain meshes (small and large).

### Minimal Reproduction

```
import pyvista as pv

p = pv.Plotter(window_size=[800, 600])
p.add_mesh(pv.Sphere())
p.show()
```

Steps to ### Reproduce

1. Run the script above.
2. Confirm the plot window appears.
3. Alt-tab or click another app so the PyVista window is in background.
4. Wait until the window becomes unresponsive.  eg. slight color change on window and close icon, mousing over yields blue circle of futility.
5. Return focus to the PyVista window.

### Expected Behavior

1. Window remains interactive after refocus:
2. mouse drag rotates/pans
3. scroll zoom works
4. keyboard close (q/Esc) works

### Actual Behavior

1. After backgrounding/refocus, the window is often unresponsive:
2. mouse/keyboard interaction stops working
3. closing can fail/hang
4. sometimes the window self recovers after a long/short wait
5. terminal Ctrl+C is often needed to recover

### Repro Scope
Reproduces with:

- minimal sphere example
- small structured grid terrain
- larger structured grid terrain
- Plain Qt6 and Tkinter windows are responsive on the same machine.

Environment ([pv.Report()]
```

```

Potentially Related Issues
#1260
#1018
#1078
#186
PR #1289

```python
import pyvista as pv

p = pv.Plotter(window_size=[800, 600])
p.add_mesh(pv.Sphere())
p.show()

```

1. Run the script above.
2. Confirm the plot window appears.
3. Alt-tab or click another app so the PyVista window is in background.
4. Wait until the window becomes unresponsive.  eg. slight color change on window and close icon, mousing over yields blue circle of futility.
5. Return focus to the PyVista window.

### System Information

```shell
  Date: Sat Mar 07 09:32:26 2026 UTC

                OS : Windows (10 10.0.26200 SP0 Multiprocessor Free)
            CPU(s) : 16
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.9 GiB
       Environment : Python
      GPU Renderer : NVIDIA GeForce RTX 3080 Ti/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 595.71
     Render Window : vtkWin32OpenGLRenderWindow
  MathText Support : True

  Python 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)]

           pyvista : 0.47.1
               vtk : 9.6.0
             numpy : 2.2.6
        matplotlib : 3.10.8
            scooby : 0.11.0
             pooch : 1.9.0
            pillow : 12.1.1
 typing-extensions : 4.15.0
          cyclopts : 4.6.0
           IPython : 8.38.0
             scipy : 1.15.3
        ipywidgets : 8.1.8
```

### Dataset Information

### Screenshots

Barebones app in foreground.  working normally
Image

Barebones app in background just after context switch
Image

Barebones app after 4-15 seconds in background note close button turns red, it is now unresponsive
Image

Barebones app in foreground again reporting unresponsiveness more clearly on the title bar
Image

Zetarcos:
Resolves #8383

Replace VTK's iren.start() with a custom event loop on Windows to properly handle window focus changes. The standard VTK interactor.Start() method can become unresponsive after backgrounding and foregrounding the window multiple times on Windows systems.

The fix uses ProcessEvents() in a loop with explicit Done checking, which maintains proper Windows message processing throughout the window lifecycle.

### Overview

On Windows systems, PyVista interactive plot windows become unresponsive after cycling between background and foreground focus multiple times (typically 3-4 cycles). Once unresponsive, the window no longer accepts mouse or keyboard input, requiring the user to force-close the window or terminate the Python process.

This issue is timing-dependent and reliably reproducible by:
1. Opening an interactive PyVista plot
2. Alt-tabbing away (background the window)
3. Waiting 10-20 seconds
4. Alt-tabbing back (foreground the window)
5. Repeating steps 2-4 several times

- resolves #8383
- Supersedes previous workaround from #1260

#### Root Cause
VTK's `RenderWindowInteractor.Start()` method does not properly handle Windows message processing after the window loses and regains focus. The method enters an internal event loop that can fail to process Windows messages correctly when the window state changes, leading to a hung interactor.

#### Solution
Replace the standard `iren.start()` call with a custom event loop on Windows systems that:
- Calls `ProcessEvents()` explicitly to handle Windows messages
- Checks `GetDone()` to detect window closure
- Renders at a consistent frame rate (~60 FPS)
- Yields CPU time appropriately

This approach mirrors successful implementations used in other parts of the PyVista ecosystem and ensures proper Windows message handling throughout the window lifecycle.

Non-Windows systems continue to use the standard `iren.start()` path unchanged.

#### Changes
- Modified `pyvista/plotting/plotter.py` to implement Windows-specific event loop
- Removed obsolete `process_events()` workaround for issue #1260
- Added clear code comments linking to issue #8383

#### Testing
Tested on:
- Windows 10.0.26200
- PyVista 0.47.1
- VTK 9.6.0
- Python 3.10.5

**Test procedure:**
1. Open interactive plot with sphere
2. Perform 10+ background/foreground cycles
3. Verify window remains responsive to mouse/keyboard input
4. Verify window closes cleanly with 'q' key

**Results:**
- ✅ Window remains fully responsive after many cycles
- ✅ No hanging or unresponsive behavior observed
- ✅ Clean window closure
- ✅ Non-Windows behavior unchanged (Linux/macOS use original code path)

#### Breaking Changes
None. This is a Windows-specific fix that does not affect the API or behavior on other platforms.

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -11,12 +11,12 @@ import contextlib
 from contextlib import contextmanager
 from contextlib import suppress
 from copy import deepcopy
+import ctypes
 from functools import wraps
 import io
 from itertools import cycle
 import logging
 import os
-import ctypes
 from pathlib import Path
 import platform
 import sys


pyvista/plotting/plotter.py
```diff
@@ -7375,10 +7375,18 @@ def fun(plotter):
...
                     if os.name == 'nt':  # pragma: no cover
-                        self.iren.process_events()  # type: ignore[union-attr]
-                    self.iren.start()  # type: ignore[union-attr]
+                        vtk_iren = self.iren.interactor  # type: ignore[union-attr]
+                        while True:
+                            vtk_iren.ProcessEvents()
+                            if vtk_iren.GetDone():
+                                break
+                            self.render_window.Render()  # type: ignore[union-attr]
+                            time.sleep(0.016)
```
akaszynski: I'm assuming this is for a ~60 fps. Would `TSLEEP_60FPS = 1/60` at the top of the module work instead?

akaszynski: Actually, a better approach is to:

```py
telap = time.time() - tstart_frame
need_tsleep = FPS_1_OVER_60 - telap
if need_tsleep > 0:
    time.sleep(need_tsleep)
```

This way we vary the time to sleep based on the time it takes to render the render window.

akaszynski:
Since it's not possible to test this within our unit testing I went ahead and manually tested it. I wasn't able to reproduce the issue in https://github.com/pyvista/pyvista/issues/8383, but I was able to verify that the render window does work as expected with this fix. As an unexpected benefit of this PR, I noticed that the CPU usage dropped off from fully saturating a core to only using around 16% of a core (probably because we're limiting to 60 FPS).

Might be due to Windows 11. Here's my system report:

```
  Date: Sun Mar 08 19:53:28 2026 UTC

                OS : Windows (11 10.0.26200 SP0 Multiprocessor Free)
            CPU(s) : 16
           Machine : AMD64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : Intel
       GPU Version : 4.5.0 - Build 32.0.101.7082
     Render Window : vtkWin32OpenGLRenderWindow
  MathText Support : True

  Python 3.13.12 (main, Mar  3 2026, 15:00:12) [MSC v.1944 64 bit (AMD64)]

           pyvista : 0.48.dev0
               vtk : 9.6.0
             numpy : 2.3.5
        matplotlib : 3.10.8
            scooby : 0.11.0
             pooch : 1.9.0
            pillow : 12.1.1
 typing-extensions : 4.15.0
          cyclopts : 4.5.4
    pytest-pyvista : 0.3.2
           IPython : 9.11.0
             scipy : 1.16.3
              tqdm : 4.67.3
           imageio : 2.37.2
            meshio : 5.3.5
         cmcrameri : 1.9
           cmocean : 4.0.3
          colorcet : 3.1.0
             trame : 3.12.0
      trame-client : 3.11.3
      trame-server : 3.10.0
         trame-vtk : 2.10.2
     trame-vuetify : 3.2.1
        ipywidgets : 8.1.8
     nest-asyncio2 : 1.7.2
```

akaszynski:
@Zetarcos, can you please verify that my changes still fix #8383 in your env?

akaszynski:
Actually (and maybe it's because I've been a bit more patient), I've experienced #8383 in both Python 3.10 and Python 3.13 on Windows 11. Note that the window does briefly show as non-responsive (with a red X in the upper right), but it's brief and it rapidly recovers even when trying to interact with it when it's "non-responsive". Still seems like there should be a way to ensure that it remains responsive at all times, but this PR is already an improvement.

akaszynski:
Turns out that the real trick is to use Win32's [MsgWaitForMultipleObjects](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-msgwaitformultipleobjects). As far as I can tell, even with the approach initially implemented you can't eliminate the unresponsive window behavior unless you're pumping messages and telling Windows that the application isn't stuck.

Don't ask me why a simple render isn't enough to tell Windows that the application is still alive, but with this approach we both maintain ~60 FPS and tell windows that the application isn't unresponsive.

Background and inspiration:
[Paradice Software - Sleep vs WaitMessage vs MsgWaitForMultipleObjects](https://www.paradicesoftware.com/blog/2014/02/sleep-waitmessage-msgwaitformultipleobjects/)

pyvista/plotting/plotter.py
Zetarcos: Ok I will try this and verify.  Thx!

Zetarcos: > Actually, a better approach is to:
> 
> telap = time.time() - tstart_frame
> need_tsleep = FPS_1_OVER_60 - telap
> if need_tsleep > 0:
>     time.sleep(need_tsleep)
> This way we vary the time to sleep based on the time it takes to render the render window.

Ok this is the snippet I tested.  minor change to capture the frame start.
Seems robust on my system.  No unresponsiveness in the background noted after 5-10 tries over multiple runs.

```Python
# get frame start
tstart_frame = time.time()
FPS_1_OVER_60 = 1 / 60

vtk_iren.ProcessEvents()
if vtk_iren.GetDone():
    break
self.render_window.Render()  # type: ignore[union-attr]

# get time to process and render a frame
telap = time.time() - tstart_frame
# get remaining frame time for 60 FPS
need_tsleep = FPS_1_OVER_60 - telap
# sleep if we have time remaining in the frame, otherwise just continue to the next frame
if need_tsleep > 0:
    time.sleep(need_tsleep)
```

I didn't see a good place for definitions already in use in this module but the FPS time could go at the top.

akaszynski: I've changed it quite a bit an we're using https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-msgwaitformultipleobjects

Please verify via the current branch.

Zetarcos:
> @Zetarcos, can you please verify that my changes still fix #8383 in your env?

I tested your proposed change with a minor mod to define the framestart.  Seems to work for me.  
See in the inline code review comments.

BTW: When I originally encounter #8383 sometimes the responsiveness would sometimes return immediately when brought to the foreground.  Other attempts would sometimes take 10-15 seconds to recover, maybe as Windows started to intervene.

Zetarcos:
> Turns out that the real trick is to use Win32's [MsgWaitForMultipleObjects](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-msgwaitformultipleobjects). As far as I can tell, even with the approach initially implemented you can't eliminate the unresponsive window behavior unless you're pumping messages and telling Windows that the application isn't stuck.
> 
> Don't ask me why a simple render isn't enough to tell Windows that the application is still alive, but with this approach we both maintain ~60 FPS and tell windows that the application isn't unresponsive.
> 
> Background and inspiration: [Paradice Software - Sleep vs WaitMessage vs MsgWaitForMultipleObjects](https://www.paradicesoftware.com/blog/2014/02/sleep-waitmessage-msgwaitformultipleobjects/)

Yep.  After continuing to test.  I see the red X pop on and off after running for about 30 min when I noticed.
So this PR #8384 is just a work around as you discovered.   Should I try to integrate the messages?

pyvista/plotting/plotter.py
```diff
@@ -7370,17 +7372,40 @@ def fun(plotter):
...
+                                # unresponsive
+                                ctypes.windll.user32.MsgWaitForMultipleObjects(  # type: ignore[attr-defined]
+                                    0,
+                                    None,
+                                    False,
+                                    sleep_ms,
+                                    0x04FF,  # QS_ALLINPUT
+                                )
+                    else:
+                        self.iren.start()  # type: ignore[union-attr]
```
akaszynski, line 7406: > Yep. After continuing to test. I see the red X pop on and off after running for about 30 min when I noticed.
So this PR #8384 is just a work around as you discovered. Should I try to integrate the messages?

@Zetarcos, I've pushed to your branch, can you see if this new implementation works?

Zetarcos: looks good.  

minor change suggested to conditional import ctypes.

will continue testing.

pyvista/plotting/plotter.py
```diff
@@ -16,6 +16,9 @@
 from itertools import cycle
 import logging
 import os
+
+if os.name == 'nt':
+    import ctypes
```
akaszynski:
```suggestion
import ctypes
```

akaszynski: We generally don't conditionally import. We've also imported ctypes already, so it's only a few nanoseconds to add it to this module.

akaszynski (approved):

pyvista/plotting/plotter.py
Zetarcos: Make sense and Sounds good.  
Thanks for looking at my change proposal!

PR #8388, opened 2026-03-09
ENH: Add top/bottom colors to north arrow widget

T4mmi:
This PR provides a different geometry for the default north arrow widget.  
The current one is identical for top/bottom view so it is sometime hard to evaluate whether the scene is viewed from under/over. The proposal bumps the arrow from under so one can quickly evaluate the up direction (z axis).

### Overview

before:
image

after:
image
image


user27182:
IMO the new look isn't great aesthetically.
image

And at a quick glance, it may not be readily obvious which side is supposed to be the top vs bottom since both top and bottom kind of look the same from some angles (e.g. view from directly above or directly below).

What about keeping all the points the same, but just splitting the bottom face into two? That way, the geometry stays the same and the top face continues to look "clean", but the bottom is now distinguishable by the presence of a center line.

Or, we have other tricks we can employ, like use RGB scalars to color the faces, and color the bottom face differently (e.g. always black, or make it the inverse of the selected color). Or even computing normals, flipping the normals of the bottom face,  and adding backface culling...

T4mmi:
Any idea is ok to me !  
just beware that the face splitting is only "visible" if you show edges or use differentiated colors for faces (wich is a solid option too for up/down distinction).

T4mmi:
I tried to match the original geometry, adding the z bump.  
Coloring faces might work but requires changing the method adding it to the renderer [here](https://github.com/T4mmi/pyvista/blob/feat/3d-north-arrow/pyvista/plotting/renderer.py#L1399) since the actor will try and set a plain color for all faces 😞

user27182:
> Coloring faces might work but requires changing the method adding it to the renderer [here](https://github.com/T4mmi/pyvista/blob/feat/3d-north-arrow/pyvista/plotting/renderer.py#L1399) since the actor will try and set a plain color for all faces 😞

Indeed we can't set the actor's color for this. Instead, we need to use RGB scalars to map the color to each face.

e.g. color the top red, bottom blue, and sides green:

``` py
import pyvista as pv
from pyvista.plotting.tools import create_north_arrow
arrow = create_north_arrow()
arrow['index'] = range(arrow.n_cells)
top_color = 'red'
bottom_color = 'blue'
side_color = 'green'
colors = [side_color, side_color,side_color,side_color, bottom_color,top_color]
colored_arrow = arrow.color_labels(colors)
colored_arrow.plot()
```

image


I would support keeping the default arrow as-is, no changes to geometry or its color, but add new keyword args ``top_color`` and ``bottom_color``. When these are `None` (default), no change. But if not None, then use `color_labels` as above.

T4mmi:
seems like a good option to me, should just be aware of color vision deficiency (mixing green and red is usually not a great idea)

user27182:
The example was more about _how_ to color the different faces, the colors used were just RGB for simplicity. But you’re correct, if this feature is implemented, there should be a docstring example included, and the colors in the example should be selected more carefully.

banesullivan (dismissed):

PR #8389, opened 2026-03-09
Support registration of custom notebook backends

banesullivan:
PyVista currently only supports a fixed set of Jupyter backends (`static`, `trame`, `server`, `client`, `html`, `none`). Third-party visualization tools that want to integrate with PyVista's notebook rendering pipeline have no extension point. This change introduces a public registration API and an entry point discovery mechanism so that external packages can plug in custom rendering strategies (e.g., alternative widget frameworks) without forking or monkey-patching PyVista.

- Add `pv.register_jupyter_backend()` to let third-party packages register custom notebook display handlers without modifying PyVista internals
- Support automatic discovery of backends via the `pyvista.jupyter.backends` entry point group, so packages can register handlers through their `pyproject.toml` alone
- Custom backend handlers receive `(plotter, screenshot=..., **kwargs)` and return any IPython-displayable object
- Built-in backend names are protected from collisions, and error messages now list both built-in and registered custom backends
- Includes full test coverage for registration, collision detection, case normalization, handler dispatch, and entry point discovery

**Programmatic registration:**
```python
import pyvista as pv

def my_handler(plotter, screenshot=None, **kwargs):
    # Return any IPython-displayable object
    return MyWidget(plotter)

pv.register_jupyter_backend('my_backend', my_handler)
pv.set_jupyter_backend('my_backend')
```

**Entry point discovery** (in a third-party package's `pyproject.toml`):
```toml
[project.entry-points."pyvista.jupyter.backends"]
my_backend = "my_package.jupyter:handler"
```

Entry points are discovered lazily on first lookup, so there is no startup cost when custom backends are not in use.

**The question you're probably asking**
Is there seriously another Jupyter notebook backend? Yes. Yes there is 😎  However, I can't open source it just yet but want to have this mechanism in place for when I do. Here's a preview:

https://github.com/user-attachments/assets/e0114dac-81ca-4146-a57d-96a2b08df7d8

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -117,6 +117,7 @@ if TYPE_CHECKING:
     from pyvista.core._typing_core import VectorLike
     from pyvista.core.utilities.arrays import CellLiteral
     from pyvista.core.utilities.arrays import PointLiteral
+    from pyvista.jupyter import JupyterBackendOptions
     from pyvista.plotting._typing import BackfaceArgs
     from pyvista.plotting._typing import CameraPositionOptions
     from pyvista.plotting._typing import Chart
@@ -7116,7 +7117,7 @@ class Plotter(_NoNewAttrMixin, BasePlotter):
         screenshot: str | Path | io.BytesIO | bool = False,  # noqa: FBT001, FBT002
         return_img: bool = False,  # noqa: FBT001, FBT002
         cpos: CameraPositionOptions | None = None,
-        jupyter_backend: str | None = None,
+        jupyter_backend: JupyterBackendOptions | str | None = None,
         return_viewer: bool = False,  # noqa: FBT001, FBT002
         return_cpos: bool | None = None,  # noqa: FBT001
         before_close_callback: Callable[[Plotter], None] | None = None,


pyvista/jupyter/__init__.py
```diff
@@ -11,9 +13,87 @@
...
+    from collections.abc import Callable
+
 JupyterBackendOptions = Literal['static', 'client', 'server', 'trame', 'html', 'none']
 ALLOWED_BACKENDS = get_args(JupyterBackendOptions)
 
+_custom_backends: dict[str, Callable[..., object]] = {}
+_entry_points_loaded: bool = False
+
+
+def register_jupyter_backend(name: str, handler: Callable[..., object]) -> None:
```
user27182, line 26: This should be documented, no? Maybe add a new autofunction here:
https://github.com/pyvista/pyvista/blob/main/doc/source/user-guide/jupyter/index.rst?plain=1

And then the doc fixes can be reverted so that there is a proper link to the new funciton

user27182: added in https://github.com/pyvista/pyvista/pull/8389/commits/7622462eb0bdb91334d17afc5e02c44588d12209

pyvista/jupyter/__init__.py
```diff
@@ -37,25 +117,29 @@ def _validate_jupyter_backend(
...
+    all_backends = list(ALLOWED_BACKENDS) + sorted(_custom_backends.keys())
+    backend_list_str = ', '.join([f'"{item}"' for item in all_backends])
+    msg = (
+        f'Invalid Jupyter notebook plotting backend "{backend}".\n'
+        f'Use one of the following:\n{backend_list_str}'
+    )
+    raise ValueError(msg)
+
+
+def set_jupyter_backend(backend: JupyterBackendOptions | str, name=None, **kwargs):  # noqa: ARG001
```
user27182, line 142: I guess this is a separate issue, but name and kwargs here are not used and do nothing.

tests/plotting/jupyter/test_custom_backends.py
```diff
@@ -0,0 +1,105 @@
...
+import pytest
+
+import pyvista as pv
+from pyvista.jupyter import _custom_backends
+from pyvista.jupyter import _get_custom_backend_handler
+from pyvista.jupyter import _validate_jupyter_backend
+from pyvista.jupyter import register_jupyter_backend
+
+has_ipython = bool(importlib.util.find_spec('IPython'))
+skip_no_ipython = pytest.mark.skipif(not has_ipython, reason='Requires IPython package')
```
user27182, line 18: I am not a fan of skipping tests like this. If `IPython` is somehow removed from test dependencies, these tests will never run and we won't know about it. If running tests, is it not reasonable to assume all test dependencies are installed, or that they _should_ be installed?

I think it's OK for this PR, because these kinds of skips also exist elsewhere, but we should clean these up and either remove them (i.e. require that all test deps _must_ be installed), or if we really want to enable skips, then add additional `'CI' in os.environ` checks, to ensure that these are never skipped for CI.

user27182:
@banesullivan I pushed changes to document the new function, and updated the typing to keep the existing `JupyterBackendOptions` type annotations for public functions that accept this as user-input. This will enable IDEs to continue hinting and auto-completing the existing backends, but also allow any generic `str` inputs for custom backends.

pyvista/jupyter/__init__.py
banesullivan: Thanks for handling this!

pyvista/jupyter/__init__.py
banesullivan: Yeah this is for sure left over from a previous implementation where the `name` was passed to trame when spinning up the server. This should be cleaned up, I just didn't want to tackle it in this PR

tests/plotting/jupyter/test_custom_backends.py
banesullivan: I think you're right here, these tests shouldn't skip and we should always have all test dependencies installed. Otherwise we end up with issues where people `pip install pyvista[jupyter]` and its missing a needed dependency (there was an issue from this not too long ago). Should we handle that in this PR? Should we go ahead and remove the skips for this test module?

banesullivan:
Thanks for cleaning up that documentation issue and for preserving the type hinting!

pyvista/jupyter/__init__.py
user27182: Created #8391 to track this

tests/plotting/jupyter/test_custom_backends.py
user27182: If you agree then yes maybe remove the skips from this PR. I still created #8392 though to keep track of the need to remove these skips from any other test modules though.

user27182:
New docs: https://69b05e21502e8196eb8cbfe2--pyvista-dev.netlify.app/user-guide/jupyter/index.html#pyvista.register_jupyter_backend

PR #8393, opened 2026-03-11
Add `CellType` properties

user27182:
### Overview

VTK has `vtkCellTypeUtilities` to make it easy to determine things like cell type linearity or dimension. This PR adds a similar utility in the form of new `pv.CellType` enum properties:
   - `vtk_class`
   - `dimension`
   - `is_linear`
   - ~`is_primary`~`is_composite`
   - `n_points`
   - `n_edges`
   - `n_faces`

There are also internal changes and refactoring:
- Use `vtkCellTypeUtilities` to get the corresponding class, dimension, and linearity
- Remove private lookup dicts inside `celltype.py` and use new `CellType` enum properties instead
- Convert the internal points/edges/faces overrides into a boolean

### DIFF
--- a/pyvista/core/celltype.py
+++ b/pyvista/core/celltype.py
@@ -901,7 +901,7 @@ class CellType(IntEnum):
 
         See Also
         --------
-        is_primary
+        is_composite
         pyvista.Cell.is_linear
 
         Examples
--- a/doc/source/_templates/autosummary/enum.rst
+++ b/doc/source/_templates/autosummary/enum.rst
@@ -14,7 +14,7 @@
 
    CellType.dimension
    CellType.is_linear
-   CellType.is_primary
+   CellType.is_composite
    CellType.n_points
    CellType.n_edges
    CellType.n_faces


user27182:
New attributes 
https://69b312670b71903910318ef9--pyvista-dev.netlify.app/api/utilities/_autosummary/pyvista.celltype#attributes

image

pyvista/core/celltype.py
```diff
@@ -773,85 +728,364 @@ def __new__(  # noqa: PYI034
...
+
+        >>> import pyvista as pv
+        >>> pv.CellType.LINE.n_points
+        2
+
+        :attr:`POLY_LINE` has a variable number of points and raises a ValueError.
+
+        >>> try:
+        ...     pv.CellType.POLY_LINE.n_points
+        ... except ValueError:
```
akaszynski: I'd prefer printing out the error here and in other examples. Also "!" should be avoided, even in print statements. For example:

```py
        ... except ValueError as error:
            print(error)

File: pyvista/core/celltype.py [hunk 2]
```diff
@@ -773,85 +728,364 @@ def __new__(  # noqa: PYI034
...
+
+        >>> pv.CellType.QUADRATIC_HEXAHEDRON.is_linear
+        False
+
+        """
+        return self._is_linear
+
+    @property
+    def is_primary(self) -> bool:  # numpydoc ignore=RT01
+        """Return ``True`` if this cell type is primary.
```
akaszynski: Perhaps a description of "primary" might be helpful.

user27182: Maybe we should inver this logic, and call this `is_composite` instead of `is_primary`? While trying to write the description, it seems less intuitive to say "A primary cell cannot be decomposed further into sub-cells". Seems more easily understood if we say "A composite cell is a composed of multiple smaller cells."

pyvista/core/celltype.py
user27182: Done in https://github.com/pyvista/pyvista/pull/8393/commits/d5daf97813cbb1959237fef2ff32c36191fd883f

pyvista/core/celltype.py
user27182: And hence we could write the docstring simply as:

"""Return ``True`` if the cell type is composed of multiple primary cells."""

And if there's still confusion, I think the examples make this clear (VERTEX vs POLY_VERTEX)

akaszynski: Composite is more intuitive, let's go with that.

akaszynski:
Clusters has a deprecated runner, upgrading, Linux runs will be disabled for a bit:

```
Runner version v2.330.0 is deprecated and cannot receive messages
```

~Fixing~ Fixed in https://github.com/pyvista/arc-runners/pull/14

PR #8396, opened 2026-03-11
Include named colormap type hints for `color_labels` filters

user27182:
Enable IDEs to hint the colormaps e.g.

image


user27182:
The failure https://github.com/pyvista/pyvista/actions/runs/22985125660/job/66733404404?pr=8396 is unrelated.

PR #8402, opened 2026-03-14
DOC: Add CalculiX cgx custom colormap example to colormap.py

3rav:


user27182:
Whoops, sorry @akaszynski , I thought the simple Cylinder didn't do the colormap justice, so I updated the example to use a mesh with actual stress scalars. I hope that's OK @3rav

user27182:
image

akaszynski:
I prefer the notched tensile specimen dataset (even though that was generated from MAPDL). Up to you if you want to use that or another dataset @3rav.

3rav:
Even better

PR #8404, opened 2026-03-15
Add `generate_cell_blocks` to generate mesh with arbitrary cell type(s)

user27182:


examples/00-load/linear_cells.py
```diff
@@ -273,5 +273,59 @@ def add_cell_helper(pl, *, text, grid, subplot, cpos=None):
...
+
+# %%
+# Remove special linear cell types not used earlier.
+linear_cell_types.remove(pv.CellType.EMPTY_CELL)
+linear_cell_types.remove(pv.CellType.POLYHEDRON)
+linear_cell_types.remove(pv.CellType.CONVEX_POINT_SET)
+
+# %%
+# Generate a :class:`~pyvista.MultiBlock` with all 16 linear cells.
+linear_cells = generate_cell_blocks(linear_cell_types)
```
user27182, line 297: Maybe forego all this added complexity and just use range?
```suggestion
# Generate a :class:`~pyvista.MultiBlock` with all 16 linear cell types.
linear_cells = generate_cell_blocks(range(1, 17))
```

akaszynski: It's less explicit and relies on VTK keeping their cell types consistent. I'm fine with either approach.

user27182: Okay let’s keep as-is then. Omitting 0 in particular (EMPTY_CELL) is needed to avoid empty mesh plotting errors. And that’s not immediately obvious.

akaszynski:
Looking at the online docs I really wish `vtkjs` had full parity with VTK... the points are too large when not rendered as spheres. Still, the native VTK plots work and the docs look great.

PR #8405, opened 2026-03-16
Fixes for latest VTK dev wheels (VTK 9.7)

user27182:
- `vtkImageThreshold` is deprecated, use `vtkBinaryImageThreshold` instead
- `vtkDataArray.GetSize` is deprecated, use `GetCapacity` instead
- `vtkWedge` point ordering has changed to fix negative volume issues
- `vtkCellTypeUtilities.GetClassNameFromTypeId` now returns `UnknownClass` for classes without a concrete `vtkCell` implementation

### DIFF
--- a/pyvista/core/filters/image_data.py
+++ b/pyvista/core/filters/image_data.py
@@ -1850,7 +1850,7 @@ class ImageDataFilters(DataSetFilters):
             field = self.get_array_association(scalars, preference=preference)
 
         # For some systems integer scalars won't threshold
-        # correctly. Cast to float to be robust.
+        # correctly. Cast to float to be robust. See https://gitlab.kitware.com/vtk/vtk/-/work_items/20019
         cast_dtype = (array_dtype := self.active_scalars.dtype) == np.int64  # type: ignore[union-attr]
         if cast_dtype:
             alg_input = self.copy(deep=False)


tests/core/test_geometric_objects.py
```diff
@@ -113,9 +113,9 @@ def test_solid_sphere():
 
     # make sure cell creation gives positive volume.
     for cell in sphere.cell:
-        assert cell.cast_to_unstructured_grid().volume > 0
+        assert cell.cast_to_unstructured_grid().volume > -1e-4
```
user27182: FYI @MatthewFlamm there seems to be changes to cell volume calcs with latest VTK dev wheels.

Looks like SolidSphere generates Wedges

``` py
>>> import pyvista as pv
>>> pv.SolidSphere().distinct_cell_types
{<CellType.TETRA: 10>,
 <CellType.HEXAHEDRON: 12>,
 <CellType.WEDGE: 13>,
 <CellType.PYRAMID: 14>}
```

The point ordering may need to change fix negative volume issues

https://discourse.vtk.org/t/vtk-wedge-cell-types-fix-point-ordering-triangulation-and-volume-correctness/16322

user27182:
Will also need to wait for upstream changes to resolve numpy wrapper issues.
- https://gitlab.kitware.com/vtk/vtk/-/merge_requests/13030
- https://gitlab.kitware.com/vtk/vtk/-/merge_requests/13016

tests/core/test_geometric_objects.py
MatthewFlamm: Yes there are wedge types.  I'm unclear on that MR whether it is only the documentation is changing or whether the code is changing.  If the code, do we need to swap the ordering here based on version?  This will be seriously annoying for backwards compatibility.

MatthewFlamm: I looked at the changes, it seems that we will have to order points according to whatever vtk version is being used.  We will want to call this out in the documentation of that function.

user27182: yes, our other tests confirms this, e.g. the new FRDReader does this to reorder wedges:
``` py
        if etype == FRDElementType.PE6:
            return [node_ids[0], node_ids[2], node_ids[1], node_ids[3], node_ids[5], node_ids[4]]
```
This produced a wedge with positive volume for VTK 9.6. But this fails with 9.7. Removing these lines (i.e. _not_ reordering the points) fixes this.

MatthewFlamm: It seems likely I put in this part of the test because I couldn't figure out the correct cell point ordering to give the correct volume in the latter part of the test.

user27182: I added a point swap and reverted changes to the tests in https://github.com/pyvista/pyvista/pull/8405/commits/001e133c07438ee63372d5720723d2fa3059af19

pyvista/core/_vtk_core.py
```diff
@@ -11,6 +11,19 @@
...
+# There are hundreds of array classes that have overrides, so instead of polluting the
+# `_vtk_core` imports and manually overriding each class, we remove the array overrides wrapper
+# This must be done before importing classes from vtkmodules
+with contextlib.suppress(ImportError):
+    from vtkmodules import MODULE_MAPPER
+
+    core_overrides: list[str] | None = MODULE_MAPPER.pop('vtkCommonCore', None)
+    array_overrides = 'vtkmodules.numpy_interface.array_overrides'
+    if core_overrides is not None and array_overrides in core_overrides:
+        core_overrides.remove(array_overrides)
```
user27182: Just want to highlight this: VTK has new array overrides to give `vtkDataArray` a NumPy-like API. I think this is great, and would enable us to clean up a lot of ugly internals for things like `convert_array`. But we can't use them due to a breaking change with `trame-vtk`, so they're disabled here. See https://gitlab.kitware.com/vtk/vtk/-/issues/19996

user27182:
FYI @larsoner this PR should resolve https://gitlab.kitware.com/vtk/vtk/-/issues/19996 for MNE-Python. Will probably be another week or so before our CI is green and this gets merged though, so you may want to temporarily install PyVista from this branch if having green CI is important for MNE-Python

user27182:
> Will also need to wait for upstream changes to resolve numpy wrapper issues.
> 
> * https://gitlab.kitware.com/vtk/vtk/-/merge_requests/13030
> * https://gitlab.kitware.com/vtk/vtk/-/merge_requests/13016

Still waiting for these updates. Looks like the latest dev wheels for `9.6.20260321` didn't get published https://gitlab.kitware.com/vtk/vtk/-/packages. Will wait for `9.6.20260328` this weekend.

pyvista/core/_vtk_core.py
user27182: We can keep the array overloads after all since #8410 fixed the trame-vtk issue


but we still need to check vtk version and trame-vtk version and tell users using vtk 9.7 to update to the latest trame-vtk (>=2.11.5)

tests/plotting/jupyter/test_trame.py
```diff
@@ -48,6 +48,11 @@
         r'\(Use ExportLegacyFormat, or GetOffsetsArray/GetConnectivityArray instead\.\)'
         ':DeprecationWarning:trame_vtk'
     ),
+    pytest.mark.filterwarnings(
+        r'ignore:Call to deprecated method GetSize\. '
+        r'\(Use GetCapacity\(\) instead\)'
+        r':DeprecationWarning:trame_vtk'
+    ),
```
user27182, line 55: This silences the warning for our tests, but users will still get nuisance warnings. These should be filtered in the source module (e.g. `jupyter.py`). This can be addressed separately, since there are other existing warnings that are similarly silenced in tests, but should be silenced in the user-facing code.

tests/plotting/test_gc.py
```diff
@@ -22,7 +22,7 @@ def test_leak_vtk() -> None:
 def test_leak_pv(sphere) -> None:
     """A VTK leak within a pyvista object with a simple self-reference."""
     points = sphere.points
-    points.VTKObject._ref = points.VTKObject
+    points.VTKObject._ref = points
```
user27182, line 25: Not really sure why this was needed, it has to do with the new VTK array overrides, maybe someone can confirm this change is OK

tests/plotting/test_plotting.py
```diff
@@ -3493,12 +3493,14 @@ def test_tight_square(noise_2d):
     )
 
 
+@pytest.mark.skip_check_gc  # Remove once resolved https://gitlab.kitware.com/vtk/vtk/-/work_items/20018
```
user27182, line 3496: These check gc skips need to be removed before next release. There are leaks _everywhere_ not just in plotting (we don't check gc for core tests, but I confirmed they're there too).

IMO if this doesn't get fixed, we should just disable the new VTK array overrides.

user27182 (commented):
Finally ready for review. Note that there are a few follow-ups needed:

- Some cell types still need to be deprecated https://gitlab.kitware.com/vtk/vtk/-/work_items/20016
- The new VTK array overrides work, but require `trame-vtk>=2.11.6`. But this version of trame-vtk also broke interactive plots (see https://github.com/pyvista/pyvista/issues/8442), so IMO the new array overrides should be disabled before release if that issue cannot be resolved, since we would need to pin `trame-vtk<=2.10.3` (see https://github.com/pyvista/pyvista/pull/8405#discussion_r2967030383 for how these can be disabled)
- A VTK version check and  `trame-vtk` version check should ideally be perfomed wherever `trame-vtk` is imported, and if the VTK version is 9.7 or later, then we need to ensure the `trame-vtk` version is also >=2.11.6. We can wait until #8442 is resolved first before commit to this.
- There are also gc issues with the VTK array overrides which needs to be resolved upstream https://gitlab.kitware.com/vtk/vtk/-/work_items/20018

tests/plotting/test_gc.py
akaszynski: This was added to ensure we cause a self-reference. How we do this is irrelevant as long as we create the reference. Changing this is fine.

akaszynski:
> new array overrides should be disabled before release if that issue cannot be resolved

Seems there are multiple issues with this anyway. Let's see what VTK 9.7 looks like when they release it and adjust accordingly.

pyvista/core/_vtk_core.py
akaszynski: And wait for them to fix GC...

user27182:
Looks like the new merge queue stuff is not compatible with vtk dev checks. Codecov is failing precisely because the dev checks are not running as part of the merge. @banesullivan

user27182:
it’s possible the merge queue will still fail, even if vtk dev checks are enabled with the merge queue. This is because Codecov fails _temporarily_ when the other Codecov reports are initially uploaded… it’s not until all reports are uploaded that codecov will report succeeds.  So, we may need to delay uploading all reports until after all the Linux checks are finished. 

Note that this isn’t specific to dev checks either. If a new feature is added that requires VTK 9.6, and any other Linux checks finish first, then the Codecov with also temporarily fail in this case.

banesullivan:
Proposed a fix for this in https://github.com/pyvista/pyvista/pull/8420#issuecomment-4204165580

but turns out it was this and also a genuine coverage miss there. need to pull that change out and land as a sperate PR

akaszynski:
@user27182, this good to merge?

user27182:
Yes

user27182:
> Looks like the new merge queue stuff is not compatible with vtk dev checks. Codecov is failing precisely because the dev checks are not running as part of the merge.

This is still an issue, so merging will fail. 

Plus, the Codecov upload race condition still exists due to
https://github.com/pyvista/pyvista/issues/8486

EDIT:
Seems I don’t understand the merge queue since the merge succeeded somehow

PR #8406, opened 2026-03-18
Add unique_bar keyword to manage scalar bars in subplots

### LINKED ISSUE #8398: scalar bar is not shown when using multiple subplots
### scalar bar is not shown when using multiple subplots

Hi!

When using multiple subplots, scalar bar is shown only for the first subplot. 
I've attached a minimal example code and screenshot of the issue.

Investigating PyVista's source code, it seems that the issue is related to the `return` statement below.
If my understanding is correct, we might need to check `render` parameter before returning. 

https://github.com/pyvista/pyvista/blob/577bb09c271ef96f247219160dad62e6fea16b18/pyvista/plotting/scalar_bars.py#L408-L409

If this is not intentional behavior, I would be happy to submit a PR. Thanks!

```python
import pyvista as pv

mesh = pv.Sphere()
random_scalars = mesh.points[:, 2] + 5
mesh["random_scalars"] = random_scalars

plotter = pv.Plotter(
    shape=(2, 2)
)
positions = [(0, 0), (1, 0), (0, 1), (1, 1)]
for pos in positions:
    plotter.subplot(pos[0], pos[1])
    plotter.add_mesh(
        mesh.copy(),
        scalars="random_scalars",
        clim=[0, 10],
        show_scalar_bar=True,
    )
plotter.show()
```

### System Information

```shell
  Date: Thu Mar 12 23:37:09 2026 UTC

                OS : Linux (Pop!_OS 22.04)
            CPU(s) : 14
           Machine : x86_64
      Architecture : 64bit
               RAM : 38.8 GiB
       Environment : Python
       File system : ext4
        GPU Vendor : Intel
       GPU Version : 4.6 (Core Profile) Mesa 25.1.5-1pop0~1756399231~22.04~b84bab8
  MathText Support : True

  Python 3.13.5 (main, Jun 12 2025, 12:40:22) [Clang 20.1.4 ]

           pyvista : 0.48.dev0
               vtk : 9.5.2
             numpy : 2.3.5
        matplotlib : 3.10.7
            scooby : 0.11.0
             pooch : 1.8.2
            pillow : 12.0.0
 typing-extensions : 4.15.0
          cyclopts : 4.3.0
    pytest-pyvista : 0.3.2
             PyQt5 : Version unknown
           IPython : 9.8.0
             scipy : 1.16.0
              tqdm : 4.67.1
           imageio : 2.37.2
            meshio : 5.3.5
         cmcrameri : 1.9
           cmocean : 4.0.3
          colorcet : 3.1.0
             trame : 3.12.0
      trame-client : 3.11.2
      trame-server : 3.9.0
         trame-vtk : 2.10.2
     trame-vuetify : 3.2.0
        ipywidgets : 8.1.8
        jupyterlab : 4.5.0
     nest-asyncio2 : 1.7.2
```

### Dataset Information

_No response_

### Screenshots

Image

riku-sakamoto:
### Overview

Resolves #8398 

### Details

- Add `unqiue_bar` option to control whether a scalar bar is registered as an actor to each subplot.


riku-sakamoto:
@MatthewFlamm 

Thank you for the comment!

You're right. The naming is a bit confusing.
I intended to show scalar bars in each subplot, but they are not truly independent.
So, I should have named it `per_subplot=True` instead of `unique_bar=False`.

That said, scalar bars are currently managed by a dictionary of which key is scalar bars' name. So, if we change the way of managing scalar bars, it seems possible to make them independent in each subplot.
If preferred, I would try to do so.

MatthewFlamm:
In the current PyVista code, can we get unique scalar bar by providing a unique scalar bar name, even if the scalar name is the same?  If so, one thought is that if the user supplies `unique_bar=True`, then could mangle the name with a integer or something like this to get unique name.

Alternatively, we could document this use case instead of adding functionality.  I do see the value in having an option here, as it seems easier to find this way.

riku-sakamoto:
@MatthewFlamm 

Sorry for the delayed response, and thank you for the suggestion.
I agree your idea and I've updated the code to create a key which is unique to each subplot.

It is slightly possible that those keys are accidentally conflicted, but I think it is a rare case. 
Ideally, we should manage scalar bars in a way that they are independent in each subplot, but it seems to require a large change. So, I think this is a good compromise for now.

pyvista/plotting/scalar_bars.py
```diff
@@ -326,6 +327,13 @@ def add_scalar_bar(  # noqa: PLR0917
 
             .. versionadded:: 0.44.0
 
+        unique_bar : bool, default: False
+            Whether to create a scalar bar which is unique to the subplot.
+            If ``True``, the scalar bar will be created with a unique key
```
MatthewFlamm, line 397: Should we document what the unique key behavior will be?

MatthewFlamm (commented):
This is looking good to me, but one last comment from me.

pyvista/plotting/scalar_bars.py
riku-sakamoto: Thank you for the comment!
I've updated the docstring and added the behavior as a note.

pyvista/plotting/scalar_bars.py
```diff
@@ -326,6 +327,24 @@ def add_scalar_bar(  # noqa: PLR0917
...
+            which is not shared with other subplots, even if the input title is the same.
+
+            .. note::
+                Scalar bars are managed by a dictionary with the title
```
MatthewFlamm, line 402:
```suggestion

                Scalar bars are managed by a dictionary with the title
```

banesullivan:
Sorry about the CI failures. We changed the deployments of some build artifacts and forgot to cleanly handle external PRs. We will need to merge in `main` here after https://github.com/pyvista/pyvista/pull/8512 lands

Issue #8395, opened 2026-03-11
AttributeError when closing plotter in key callback


titorau:
### Describe the bug, what's wrong, and what you expected.

**Use case**
I want to close the plotter when I press `A` on my keyboard. 

**Expected behavior**
The window closes silently.

**Current behavior**
My script crashes with the following error.
```
Traceback (most recent call last):
  File "/home/titorau/Documents/issues/pyvista_close/close.py", line 6, in <module>
    plotter.show()
  File "/home/titorau/.venv/lib/python3.12/site-packages/pyvista/_deprecate_positional_args.py", line 243, in inner_f
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/home/titorau/.venv/lib/python3.12/site-packages/pyvista/plotting/plotter.py", line 7397, in show
    _is_current = self.render_window.IsCurrent()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'IsCurrent'
```

### Steps to reproduce the bug.

```python
import pyvista as pv
plotter = pv.Plotter()
plotter.add_key_event("a", plotter.close)
plotter.add_mesh(pv.Cube())
plotter.show()
```

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Wed Mar 11 09:37:52 2026 UTC

                OS : Linux (Ubuntu 24.04)
            CPU(s) : 6
           Machine : x86_64
      Architecture : 64bit
               RAM : 31.2 GiB
       Environment : Python
       File system : ext4
        GPU Vendor : Intel
      GPU Renderer : Mesa Intel(R) UHD Graphics 630 (CFL GT2)
       GPU Version : 4.6 (Core Profile) Mesa 25.2.8-0ubuntu0.24.04.1
     Render Window : vtkXOpenGLRenderWindow
  MathText Support : True

  Python 3.12.3 (main, Mar  3 2026, 12:15:18) [GCC 13.3.0]

           pyvista : 0.47.1
               vtk : 9.5.2
             numpy : 2.3.4
        matplotlib : 3.10.7
            scooby : 0.11.0
             pooch : 1.9.0
            pillow : 12.0.0
 typing-extensions : 4.15.0
          cyclopts : 4.8.0
         pyvistaqt : 0.11.3
           IPython : 8.12.3
              tqdm : 4.67.1
        ipywidgets : 8.1.8
        jupyterlab : 4.5.1
[...]

PR #8413, opened 2026-03-21
fix: handle closed render window in show() after key callback close

saschabuehrle:
## Bug
[#8395](https://github.com/pyvista/pyvista/issues/8395) — calling `plotter.close()` from an `add_key_event` callback causes an `AttributeError` because `show()` unconditionally accesses `self.render_window.IsCurrent()` after the event loop exits, but `close()` already deleted the render window.

## Fix
Cache the `render_window` reference before checking `IsCurrent()`. When it is `None` (already cleaned up by a prior `close()` call, e.g. from a key callback), skip the stale-window cleanup and screenshot branches entirely.

This mirrors the guard already present in `_check_has_render_window()` (line 2071) which checks `self.render_window is None` before accessing `IsCurrent()`.

## Testing
Verified with the reproduction script from the issue:
```python
import pyvista as pv
plotter = pv.Plotter()
plotter.add_key_event("a", plotter.close)
plotter.add_mesh(pv.Cube())
plotter.show()
```
Before: `AttributeError: 'NoneType' object has no attribute 'IsCurrent'`
After: clean exit, no error.

Happy to address any feedback.

Greetings, saschabuehrle


akaszynski:
Reproduced the error and verified solution proposed in this PR works. `mypy` error is a red herring (would love to use `ty` assuming it's still around post-acquisition).

akaszynski (dismissed):

user27182:
> Reproduced the error and verified solution proposed in this PR works. `mypy` error is a red herring

The `mypy` error is likely valid based on the current typing for `self.render_window`. It probably doesn’t include `| None`, hence doing `self.render_window is not None` is considered redundant since it cannot be None. But clearly the runtime behavior suggests that it can be None. So the proper fix would be to update `self.render_window` to include `None` in the annotations.

akaszynski:
> The `mypy` error is likely valid based on the current typing for `self.render_window`. It probably doesn’t include `| None`, hence doing `self.render_window is not None` is considered redundant since it cannot be None. But clearly the runtime behavior suggests that it can be None. So the proper fix would be to update `self.render_window` to include `None` in the annotations.

Seems like it includes `None`:

https://github.com/pyvista/pyvista/blob/8b6727924a3ef8ba3c0d598d9dfb7a108faa1d5d/pyvista/plotting/plotter.py#L441-L459

The real fix would be to use `_render_window` as an attribute and then set it to None when we delete the window (instead of `del` which seems like a bad practice).

user27182:
> > The `mypy` error is likely valid based on the current typing for `self.render_window`. It probably doesn’t include `| None`, hence doing `self.render_window is not None` is considered redundant since it cannot be None. But clearly the runtime behavior suggests that it can be None. So the proper fix would be to update `self.render_window` to include `None` in the annotations.
> 
> Seems like it includes `None`:
> 
> https://github.com/pyvista/pyvista/blob/8b6727924a3ef8ba3c0d598d9dfb7a108faa1d5d/pyvista/plotting/plotter.py#L441-L459

Hm, that `mypy` error does seem odd then. 

> The real fix would be to use `_render_window` as an attribute and then set it to None when we delete the window (instead of `del` which seems like a bad practice).

+1 

When I first added the NoNewAttrMixin class this pattern of setting attributes dynamically and using `del` later was very prevalent, and led to bugs (unexpected AttributeError), but also was terrible for type hints. Setting private attributes in init (with annotations) and setting None is way better for type checkers.

Issue #8379, opened 2026-03-03
Notch stress example dataset has invalid cells, leading to segfaults


adeak:
### Describe the bug, what's wrong, and what you expected.

Formally this is a bug report but I think the issue is with a prominent example mesh.

The immediate problem is that calling `clip()` on the notch stress example dataset leads to a segfault. With @banesullivan we've covered Windows, Mac and Linux, all show the same behaviour. @banesullivan tells me `clip()` isn't typically known to segfault.

While opening this issue I've become aware of `mesh.validate_mesh()` (neat!), and indeed the mesh is quite invalid:
```python
from pyvista import examples

mesh = examples.download_notch_stress()
validation = mesh.validate_mesh()
print(mesh.n_cells)  # 2192
print(len(validation.wrong_number_of_points))  # 2192
print(len(validation.zero_size))  # 1220
```

Every cell has the wrong number of points and more than half the cells have zero size. E.g. the first cell:
```
>>> next(mesh.cell)
Cell (0x7f9eaa2faa40)
  Type:        <CellType.HEXAHEDRON: 12>
  Linear:      True
  Dimension:   3
  N Points:    20
  N Faces:     6
  N Edges:     12
  X Bounds:    1.766e-01, 1.812e-01
  Y Bounds:    6.163e-02, 6.597e-02
  Z Bounds:    0.000e+00, 5.000e-03
```
It's a hexahedron with 20 points.

Is there any way we can clean up this mesh? Since it's on the front page it would be nice to make it more robust.

### Steps to reproduce the bug.

Segfault reproducer:
```python
from pyvista import examples

mesh = examples.download_notch_stress()
mesh.points -= mesh.center

print('before clip')
clipped = mesh.clip()  # <-- segfault
print('clip OK')
```

### System Information

uv-installed env on Debian with PyVista `main`:
```shell
--------------------------------------------------------------------------------
  Date: Tue Mar 03 21:14:52 2026 UTC

                OS : Linux (Debian GNU/Linux 12)
            CPU(s) : 12
           Machine : x86_64
      Architecture : 64bit
       Environment : Python
        GPU Vendor : AMD
      GPU Renderer : AMD Radeon Graphics (renoir, LLVM 15.0.6, DRM 3.49, 6.1.0-40-amd64)
       GPU Version : 4.6 (Core Profile) Mesa 22.3.6
     Render Window : vtkXOpenGLRenderWindow
  MathText Support : True

[...]

PR #8420, opened 2026-03-26
Add `cell_validator` tolerance keywords

user27182:
### Overview

In https://github.com/pyvista/pyvista/issues/8379 it was highlighted that cell validator currently generates false positives for meshes with small volumes. The small volume is valid, but the tolerance is currently too high such that it's incorrectly flagged as "zero volume".

Part of the issue is that the default tolerance used by `vtkCellValidator` uses eps for `float32` dtypes, whereas computing eps dynamically based on the actual points dtype makes more sense.

This PR fixes the dynamic eps issue, and also adds `tolerance`, `planarity_tolerance`, and `size_tolerance` for better user control over these values.

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -1618,14 +1618,19 @@ class DataObjectFilters:
         # Use single-precision eps by default (even if points have double precision)
         tol: float = tolerance if tolerance is not None else np.finfo(np.float32).eps
 
+        if planarity_tolerance is not None and pv.vtk_version_info < (9, 6, 0):
+            msg = 'Planarity tolerance requires VTK 9.6 or later.'
+            raise pv.VTKVersionError(msg)
+
         cell_validator = _vtk.vtkCellValidator()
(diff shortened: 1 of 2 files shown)


banesullivan (dismissed):

banesullivan:
Okay this seems to be reliably failing once it hits the queue but other PRs have all gone through fine. I'm digging into it...

banesullivan:
This took a lot of digging but I'm pretty sure its a race condition with codecov where codecov initially reports codecov/patch as failure when partial coverage data arrives (since the test has a VTK requires condition it skips on some jobs and thus no coverage), and the merge queue  immediately dequeues the PR. Codecov then updates to success moments later when the remaining coverage uploads arrive... but it's too late and its dequeued.

The easy fix is to remove codecov from the required checks (but this would remove it from the required PR checks)

There's also an `after_n_builds` option for the codecove config that may help with this... going to push that change to this branch directly to try it out.

banesullivan:
Or maybe it was just a genuine codecov patch failure!

PR #8421, opened 2026-03-27
Set `cell_validator` tolerance dynamically based on points dtype

user27182:
### Overview

Split from #8420 to separate this fix from new features.

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -1424,14 +1424,16 @@ class DataObjectFilters:
         .. versionadded:: 0.47
 
         .. versionchanged:: 0.48
-            Include cell status checks for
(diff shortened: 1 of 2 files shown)


user27182:
The test failures are very odd and unexpected. I have narrowed this down to a vtk issue where `vtkCellValidator` reports false positives: https://gitlab.kitware.com/vtk/vtk/-/work_items/20011

This isn't really a big deal, since the mesh is _already_ invalid due to having unused points, so the false positives are mostly noise.

But, to address the test failures I think it's better to explicitly set the tolerance values for some tests to avoid the false positives. But this requires new tolerance args implemented in https://github.com/pyvista/pyvista/pull/8420, so we can't split that PR after all.

PR #8423, opened 2026-03-29
Skip NSWindow for off-screen rendering and solve the zombie window on macOS 

### LINKED ISSUE #1096: Plotter window hangs around in an inoperable state after close on macOS
On MacOS, if I open up IPython and run:

```py
import pyvista as pv; pv.Sphere().plot()
```

The plotter opens up as expected, and I'm able to interact with it. When I try to close the window, the window doesn't disappear but instead hangs (beach balls).

Weirdly, IPython _is_ still interactive, and the call to `pv.Sphere().plot()` actually returns. E.g., after X-ing out the plotter window:

```py
In [6]: pv.Sphere().plot()
Out[6]:
[(1.7773465342683825, 2.4356984953049423, 1.428995981058186),
 (0.0, 0.0, 0.0),
 (0.4815608961128151, -0.6775433723938147, 0.5559083394411702)]
```

In fact, if I call `pv.Sphere().plot()` again, the X'd out window disappears, having been replaced by a new, fully interactive plotter. I can repeat this process ad infinitum.

I'm not sure if this is relevant, but I tried this with the latest VTK 9 install and also after downgrading to VTK 8.1.2. The problem is the same in both cases.

```
  Date: Thu Jan 21 07:58:22 2021 EST

                OS : Darwin
            CPU(s) : 8
           Machine : x86_64
      Architecture : 64bit
               RAM : 16.0 GB
       Environment : IPython
        GPU Vendor : Intel Inc.
      GPU Renderer : Intel Iris Pro OpenGL Engine
       GPU Version : 4.1 INTEL-14.7.11

  Python 3.7.9 (default, Sep  6 2020, 13:20:25)  [Clang 11.0.3
  (clang-1103.0.32.62)]

           pyvista : 0.27.4
               vtk : 8.1.2
             numpy : 1.19.5
           imageio : 2.9.0
           appdirs : 1.4.4
            scooby : 0.5.6
            meshio : 4.3.8
        matplotlib : 3.3.3
         pyvistaqt : 0.3.0
             PyQt5 : 5.15.2
           IPython : 7.19.0
          colorcet : 1.0.0
             scipy : 1.6.0
--------------------------------------------------------------------------------```
```

### LINKED ISSUE #8422: Skip NSWindow for off-screen rendering on macOS

When `off_screen=True` on macOS, PyVista should call `render_window.SetConnectContextToNSView(False)` to create a standalone CGL context without an NSWindow.

Currently, `vtkCocoaRenderWindow` creates an NSWindow even for off-screen rendering. This has two consequences:

1. **Background thread rendering is impossible.** Cocoa requires NSWindow creation on the main thread, so `pv.Plotter(off_screen=True)` crashes with `NSInternalInconsistencyException` when called from any thread other than main.
2. **A dock icon appears** every time an off-screen plotter is created, even though there's no window to display. This has been a minor annoyance for years (see screenshot below)

Image

With `SetConnectContextToNSView(False)`, VTK creates a raw CGL context with no Cocoa dependency. The plotter works on any thread, no dock icon appears, and rendering output is identical.

I've verified this works with PyVista on macOS Apple Silicon (VTK 9.6.0) including concurrent multi-thread rendering with separate plotters.

- [`vtkCocoaRenderWindow` source](https://github.com/Kitware/VTK/blob/master/Rendering/OpenGL2/vtkCocoaRenderWindow.mm) (see `CreateAWindow` and `SetConnectContextToNSView`)
- [Apple: Concurrency and OpenGL](https://developer.apple.com/library/archive/documentation/GraphicsImaging/Conceptual/OpenGL-MacProgGuide/opengl_threading/opengl_threading.html)

### Pseudocode or Screenshots

This crashes on macOS (NSWindow created on background thread):

```python
import threading
import pyvista as pv

pv.OFF_SCREEN = True

def render_bg():
    pl = pv.Plotter(off_screen=True)
    pl.add_mesh(pv.Sphere())
    pl.screenshot("bg.png")
    pl.close()

t = threading.Thread(target=render_bg)
t.start()
t.join()
# NSInternalInconsistencyException: NSWindow should only be instantiated on the main thread!
```

This works, including concurrent rendering on multiple threads:

```python
import threading
import pyvista as pv

pv.OFF_SCREEN = True

def render_bg(name, mesh, filename):
    pl = pv.Plotter(off_screen=True)
    pl.render_window.SetConnectContextToNSView(False)  # standalone CGL, no NSWindow
    pl.add_mesh(mesh, color="steelblue")
    pl.screenshot(filename)
    pl.close()
    print(f"[{name}] done")

# Two background threads rendering concurrently
t1 = threading.Thread(target=render_bg, args=("A", pv.Cone(), "/tmp/a.png"))
t2 = threading.Thread(target=render_bg, args=("B", pv.Sphere(), "/tmp/b.png"))
t1.start()
t2.start()
t1.join()
t2.join()
# Both produce valid screenshots. No crash, no dock icon.
```

The proposed change in PyVista would be something like this in `Plotter.__init__`:

```python
import sys

if sys.platform == "darwin" and self._off_screen:
    self.render_window.SetConnectContextToNSView(False)
```

This should be a safe default for all off-screen usage on macOS. On-screen plotters would continue using the NSWindow path as before.

banesullivan:
Fixes #8422
Fixes #1096

Two macOS-specific fixes for `vtkCocoaRenderWindow` behavior:

1. **Off-screen rendering no longer creates an NSWindow.** When `off_screen=True`, call `SetConnectContextToNSView(False)` to create a standalone CGL context instead of going through Cocoa's NSWindow path. This eliminates the dock icon flash and enables background-thread rendering (previously crashed with `NSInternalInconsistencyException`).

2. **On-screen plotter windows no longer leave a frozen "zombie" window after closing.** Always call `Finalize()` on the render window during cleanup (removing the previous Apple Silicon guard that skipped it), and flush pending Cocoa events via `ProcessEvents()` so the macOS window server actually dismisses the window. Previously, VTK's `DestroyWindow()` called `[window close]` which queued a hide message, but without processing the Cocoa event queue that message was never delivered thus leaving a frozen, unresponsive 🧟 window on screen.

## Changes

- **`pyvista/plotting/plotter.py`**:
  - Set `SetConnectContextToNSView(False)` in `Plotter.__init__` right after `SetOffScreenRendering(1)` for off-screen plotters.
  - In `_clear_ren_win()`: always call `Finalize()` on all platforms (removes the Apple Silicon guard), then flush pending Cocoa events via `ProcessEvents()` on macOS so the window server dismisses the window.
- **`pyvista/report.py`**: Set the flag in the `_cmd_render_window_info` subprocess script used by `pv.GPUInfo()`. Without this, every `GPUInfo()` call spawns a subprocess that briefly flashes a dock icon.
- **`pyvista/plotting/utilities/gl_checks.py`**: Set the flag in `check_depth_peeling()` which creates a temporary off-screen render window.
- **`pyvista/plotting/tools.py`**: Set the flag in `_supports_opengl()` which creates a temporary render window to check GL capabilities.

All `SetConnectContextToNSView` calls are guarded by `hasattr(render_window, 'SetConnectContextToNSView')` so they're no-ops on Linux/EGL where `vtkCocoaRenderWindow` doesn't exist. On-screen plotters are unaffected by the off-screen changes.

## Tests

- Added `test_off_screen_background_thread_rendering` which creates a plotter on a background thread, renders a sphere, and captures a screenshot. Without the off-screen fix, this crashes with `SIGBUS` on macOS.
- Verified on macOS Apple Silicon (VTK 9.6.0): 
    - concurrent multi-thread rendering works, `pytest -n auto tests/plotting/` produces zero dock icons
    - closing on-screen plotters no longer leaves zombie windows.


banesullivan:
After this lands it should be safe to remove both the `pyobjc-framework-Cocoa` dependency and the `NSAutoreleasePool` test fixture. Going to do that separately though out of an abundance of caution and to isolate anything that may need to be reverted if CI starts failing again due to needing that fixture.

user27182:
Tested locally and confirmed that #1096 is fixed (the window now closes properly and no longer hangs)

banesullivan:
>this has been bothering me for years!

Same. I got so frustrated with it the other day that I just had to dig into it

banesullivan:
@akaszynski and I are thinking of issuing a patch release for this change. Any thoughts on that?

user27182:
Patch release is OK since it is a bug fix.

I'd also like to add a separate fix to address VTK 9.6.1 failures that are skipped from #8429, since some distros (e.g. NixOS) will likely support the latest PyVista along with VTK 9.6.1, and their test suite will fail without a patch for 9.6.1.

But, I also don't want to hold anything up, so no need to wait for this other fix.

user27182:
> I'd also like to add a separate fix to address VTK 9.6.1 failures that are skipped from #8429, since some distros (e.g. NixOS) will likely support the latest PyVista along with VTK 9.6.1, and their test suite will fail without a patch for 9.6.1.

The fixes are added here: https://github.com/pyvista/pyvista/pull/8440

So I would recommend cherry-picking and backporting the commit to main from this PR, as well as the commit to main after #8440 is merged.

banesullivan:
Done. Cherry-picked 062693283bd0acb50e45ade38d9b2de7ba62b154 and 90f02faeb3d1ccf1294ebe0d8c9599b2c03ede69 onto `release/0.47`

bumped the version to `0.47.2` and pushed a tag. Should be live momentarily

PR #8425, opened 2026-03-30
Fix trame servers overlapping in tests

edabor:
### Overview

Fix https://github.com/pyvista/pyvista/issues/8419 by clearing the `trame` servers registry at teardown.

See https://github.com/pyvista/pytest-pyvista/pull/278#issuecomment-4154601075 for more details.

PR #8429, opened 2026-03-31
Pin VTK 9.5.2 job and limit VTK 9.6.1

banesullivan:
VTK 9.6.1 is breaking CI across all PRs. Two issues

- closes #8426 The `vtk_9.5.2` tox factor was never actually pinned in `tox.ini`, so that CI job silently resolved to 9.6.1 instead of 9.5.2
- closes #8427: VTK 9.6.1 itself has rendering regressions that break our image tests. The ghost cell workaround we added for 9.6.0 ([VTK#19922](https://gitlab.kitware.com/vtk/vtk/-/issues/19922)) conflicts with changes in 9.6.1 ([VTK MR 12877](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/12877))

Failures show up as image regression errors in `test_set_environment_texture_cubemap` and all the ExplicitStructuredGrid doc examples.

Per @user27182's [comment](https://github.com/pyvista/pyvista/pull/8425#discussion_r3010879455), this is low impact and we should temp pin until we have a proper fix.

I've combined @edabor's PRs #8426 and #8427 into one to attempt to get CI green across the board.

Additionally, we are seeing Bezier/Lagrange cell failures now... I'm marking those as skipped to get CI back online and those will need to be dealt with later

- **`tox.ini`**: Add the missing `vtk_9.5.2: vtk == 9.5.2` dep so the CI job actually tests what it says it tests
- **`pyproject.toml`**: Add `vtk!=9.6.1` to block the broken release for CI and users
- **`tests/examples/test_cell_examples.py`**: Skip `test_cell_vtk_class` and `test_cell_n_edges` for Bezier/Lagrange Triangle and Quadrilateral cell types on VTK 9.5.2. Now that the tox pin actually works, these tests reveal a ??preexisting?? VTK 9.5.2 bug where these cell types are swapped (e.g. `BezierTriangle` returns a `vtkBezierQuadrilateral`). I think we didn't notice because the "9.5.2" CI job was silently running 9.6.1 which this PR fixes.
- **`tests/plotting/test_plotting.py`**: Skip `test_chart_matplotlib_plot` and `test_add_text_latex` on macOS. The macOS CI runners are missing the DejaVu Sans font, causing consistent failures.

The "latest" CI job will fall back to 9.6.0 which works fine with our existing workaround.

### DIFF
--- a/tests/plotting/test_plotting.py
+++ b/tests/plotting/test_plotting.py
@@ -2840,6 +2840,7 @@ def test_chart_plot():
     pl.show()
 
 
+@pytest.mark.skip_mac('DejaVu Sans font missing on macOS CI runners')
 def test_chart_matplotlib_plot(verify_image_cache):
     """Test integration with matplotlib"""
     # Seeing CI failures for Conda job that need to be addressed
@@ -3133,6 +3134,7 @@ def test_add_text():
     pl.show()
 
 
+@pytest.mark.skip_mac('DejaVu Sans font missing on macOS CI runners')
 @pytest.mark.needs_vtk_version(9, 4, 0)
 def test_add_text_latex():
     """Test LaTeX symbols."""


akaszynski (approved):

banesullivan:
Merged with haste to bring CI back online and unblock other PRs

Issue #6049, opened 2024-05-07
Allow reading of files from remote file stores such as S3 [willing to contribute a PR!]


krishanbhasin-px:
### Describe the feature you would like to be added.

I would like to be able to easily read files direct from an remote filestore, such as S3.

### Links to VTK Documentation, Examples, or Class Definitions.

Currently, the [definition of `read()`](https://github.com/pyvista/pyvista/blob/eb93e5aabbfad9aa0efc946e03b154f8abb0fb6f/pyvista/core/utilities/fileio.py#L109) makes [strong assumptions](https://github.com/pyvista/pyvista/blob/eb93e5aabbfad9aa0efc946e03b154f8abb0fb6f/pyvista/core/utilities/fileio.py#L172) that the objects to be loaded live on a local filesystem.

### Pseudocode or Screenshots

I'd love to be able to either:
1. pass in a string that starts with `s3://` and have pyvista know to use [fsspec](https://filesystem-spec.readthedocs.io/en/latest/developer.html#clients-of-the-library)/[s3fs](https://github.com/s3fs-fuse/s3fs-fuse)
2. construct an fsspec filesystem object, reference a file from it, and then pass that object directly to the `read` method
3. Any other suggestions for making this work!

I'd be super happy to work on bringing a PR to do this, if you would be open to merging this kind of a change in

krishanbhasin-px:
Hey @tkoyama010, I saw your 👍 on the issue and just wanted to check; can I take that as an endorsement of the idea/you being open to merging a PR that implements this?

Sorry for the direct tag; I just want to be sure before I spend any time working on making this happen.

Thanks!

MatthewFlamm:
This seems like a good idea, but my experience is that some (many?) VTK readers are not happy with non-string based paths or direct data being passed in binary/string form.  If PyVista can transform user input to what VTK expects, it makes sense to me, particularly if we do not have to add any dependencies.

banesullivan:
This is high up there on my wish list and I'm happy to help you make this happen in pyvsita!

@MatthewFlamm makes a great point that we are mostly limited by what the upstream VTK readers can handle.

Some native VTK readers support the `ReadFromInputStringOn` option, specifically the XML VTK formats. Here is a routine that will read those files from S3 by fetching the file contents and passing along to the reader directly:

```py
def read_xml_from_s3(uri):
    import pyvista as pv
    import fsspec, s3fs
    from vtkmodules import vtkIOXML
    readers = {
[...]

PR #228, opened 2019-05-20
Use/Test Sphinx Gallery Scraper Agreement

banesullivan:
Implements a software contract between PyVista and Sphinx-Gallery per https://github.com/sphinx-gallery/sphinx-gallery/pull/494

Note that this was actually implemented in these three (fa897774ca3fa09773988ad14a1292a2aa608cb9, 0a814e8e35991825268e591678db5d3128b6e25d, and d474c94cd90514e24f9f631d98eb8d4c015a52a7) commits on the master branch - this PR tests that it's working

PR #8437, opened 2026-04-02
Add pluggable reader registry for pv.read()

### LINKED ISSUE #8436: Pluggable IO Registry for `pv.read()`
`pv.read()` already handles a wide range of formats through VTK readers and meshio fallback. But third-party libraries that support additional formats or storage backends (cloud URIs, databases, custom binary formats) have no way to plug into `pv.read()`. Users end up needing to know which library handles which format instead of having a single entry point.

This is related to #6049 (reading from S3/cloud URIs). A registry mechanism would let external packages provide that kind of support without adding more dependencies to PyVista or maintenance burden of additional formats.

## Prior Art

This pattern is well-established in the scientific Python ecosystem.

**xarray** uses a [`BackendEntrypoint`](https://docs.xarray.dev/en/stable/internals/how-to-add-new-backend.html) system where third-party packages register storage backends via the `xarray.backends` entry point group. Packages like `cfgrib`, `h5netcdf`, and `zarr` all plug into `xr.open_dataset(engine=...)` this way. Each backend implements `open_dataset()` and a `guess_can_open()` method for automatic format detection. This is probably the closest analogue to what we'd want for PyVista.

**pandas** has a [PDEP-9 proposal](https://pandas.pydata.org/pdeps/0009-io-extensions.html) for extensible IO through the `dataframe.io` entry point group. Third-party packages register readers and writers that become available as `pd.read_<name>()` and `df.to_<name>()`. pandas also already uses entry points for plotting backends (`pandas_plotting_backends` group), which is similar in spirit to our `pyvista.jupyter.backends` system.

Both projects use `importlib.metadata.entry_points()` for discovery, lazy loading to avoid import overhead, and a simple callable interface for the registered handlers. We'd follow the same conventions.

## Proposed Solution

Add a lightweight reader registry, following the same pattern we already use for Jupyter backends (`pv.register_jupyter_backend()`). Third-party packages can register reader callables keyed by file extension, URI scheme, or format name. `pv.read()` checks this registry before falling back to the existing VTK/meshio pipeline.

### Public API

```python
import pyvista as pv

# Register a reader for a custom extension
pv.register_reader('.myformat', my_custom_reader)

# Register a reader for URI schemes
pv.register_reader('s3://', my_cloud_reader)

# Then just use pv.read() as usual
mesh = pv.read('model.myformat')
mesh = pv.read('s3://bucket/data.vtu')
```

The registered callable should have a simple signature:

```python
def my_reader(path: str, **kwargs) -> pv.DataSet:
    """Read a file and return a PyVista dataset."""
    ...
```

### Entry Point Discovery

Just like the Jupyter backend system, support automatic discovery via Python entry points so third-party packages can register readers at install time without requiring users to call `register_reader()` explicitly:

```toml
# In a third-party package's pyproject.toml
[project.entry-points."pyvista.readers"]
".myformat" = "my_package:my_reader"
"s3://" = "my_package:s3_reader"
```

Entry points are discovered lazily on first lookup (not at import time), matching our existing pattern in `pyvista.jupyter`.

### Resolution Order

When `pv.read(filename)` is called:

1. Check if `filename` matches a registered URI scheme (e.g., `s3://`, `gs://`)
2. Check if the file extension matches a registered custom reader
3. Fall back to the existing `CLASS_READERS` / VTK reader pipeline
4. Fall back to meshio

Custom readers take priority over meshio but not over built-in VTK readers unless explicitly overriding. We could add an `override=True` parameter to `register_reader()` for cases where a third-party reader should replace a built-in one.

### Collision Protection

Same pattern as `register_jupyter_backend()`: by default, raise `ValueError` if the extension is already claimed by a built-in reader. The `override` flag opts into replacing built-ins when that's the intent.

## Use Cases

- **Cloud/remote IO**: Register handlers for `s3://`, `gs://`, `https://` URIs that download or stream data, then pass it through the existing pipeline or a custom reader.
- **New mesh formats**: Libraries that support formats VTK doesn't (domain-specific binary formats, database-backed meshes, custom stores) can plug directly into `pv.read()`.
- **Format upgrades**: A third-party package could register an improved reader for an existing format (e.g., faster STL parsing) with `override=True`.
- **Compressed formats**: Projects like [`zvtk`](https://github.com/pyvista/zvtk) that add zstandard compression to VTK files could register a reader for `.zvtk` so `pv.read('mesh.zvtk')` just works.

## Example: Third-Party Package Integration

A third-party library could register itself on import:

```python
# my_format_library/__init__.py
import pyvista as pv

def read_my_format(path, **kwargs):
    # Custom reading logic
    ...
    return pv.wrap(data)

pv.register_reader('.myformat', read_my_format)
```

Or automatically via entry points (no user action required):

```toml
[project.entry-points."pyvista.readers"]
".myformat" = "my_format_library:read_my_format"
```

## Pseudocode Implementation

The core implementation is minimal. Based on the existing Jupyter backend pattern in `pyvista/jupyter/__init__.py`:

```python
# pyvista/core/utilities/fileio.py (or a new registry module)

_custom_readers: dict[str, Callable[..., DataObject]] = {}
_reader_entry_points_loaded: bool = False

def register_reader(
    key: str,
    handler: Callable[..., DataObject],
    override: bool = False,
) -> None:
    """Register a custom reader for a file extension or URI scheme."""
    key = key.lower()
    if not override and key in CLASS_READERS:
        msg = f'Cannot register reader for "{key}": collides with built-in reader.'
        raise ValueError(msg)
    _custom_readers[key] = handler

def _get_custom_reader(filename: str) -> Callable[..., DataObject] | None:
    """Look up a custom reader by URI scheme or extension."""
    # Check URI schemes first
    for scheme in _custom_readers:
        if '://' in scheme and filename.lower().startswith(scheme):
            return _custom_readers[scheme]
    # Check file extensions
    ext = get_ext(filename)
    handler = _custom_readers.get(ext)
    if handler is not None:
        return handler
    _discover_reader_entry_points()
    return _custom_readers.get(ext)

def _discover_reader_entry_points() -> None:
    """Scan pyvista.readers entry point group."""
    global _reader_entry_points_loaded
    if _reader_entry_points_loaded:
        return
    _reader_entry_points_loaded = True
    from importlib.metadata import entry_points
    eps = entry_points(group='pyvista.readers')
    for ep in eps:
        key = ep.name.lower()
        if key not in CLASS_READERS and key not in _custom_readers:
            with contextlib.suppress(Exception):
                _custom_readers[key] = ep.load()
```

Then in `pv.read()`, insert a check before the existing reader pipeline:

```python
def read(filename, ...):
    ...
    # Check custom readers first (URI schemes + extensions)
    custom_reader = _get_custom_reader(str(filename))
    if custom_reader is not None:
        return custom_reader(str(filename))

    # Existing pipeline continues below
    filename = Path(filename).expanduser().resolve()
    ...
```

## Scope

This proposal covers reading only. A parallel `register_writer()` mechanism for `DataObject.save()` could follow the same pattern but I think we should focus on handling reading first.

banesullivan:
Resolves #8436 and addresses #6049 (but doesn't close it since this is read only and only niavely downloads the full file to disk)

- Add `pv.register_reader()` to let third-party packages register custom file format readers that plug into `pv.read()`. Supports both explicit registration and a `@pv.register_reader('.ext')` decorator form. Third-party packages can also register via `pyvista.readers` entry points in `pyproject.toml` for zero-config discovery at install time.
- Add remote URI support to `pv.read()`. Passing `https://`, `s3://`, or other URIs downloads the file to a temp path automatically using `fsspec` (optional, in `pyvista[io]`) or `pooch` (required dep) as fallback for HTTP(S). Custom readers receive the raw URI first, giving them the chance to handle cloud paths natively.
- Add `LocalFileRequiredError` exception that custom readers can raise to signal they need a local path, triggering automatic download-and-retry.
- Add `has_scheme()` utility for downstream packages to detect URI strings.

Companion PR in zvtk showcasing how this can be used: https://github.com/pyvista/zvtk/pull/15

### DIFF
--- a/tests/core/test_reader_registry.py
+++ b/tests/core/test_reader_registry.py
@@ -3,21 +3,19 @@
 from __future__ import annotations
 
 import contextlib
-import importlib
 import importlib.util
 import io
 from pathlib import Path
 import re
+import subprocess
+import sys
 import types
 from unittest.mock import MagicMock
-from unittest.mock import call
 from unittest.mock import patch
 
 import pytest
 
 import pyvista as pv
-import pyvista.core.utilities as pv_utilities
-from pyvista.core.utilities import fileio as _fileio_mod
 from pyvista.core.utilities import reader_registry as _reg_mod
 
 
@@ -41,15 +39,19 @@ def _mock_reader(_path, **__):
     return pv.PolyData()
 
 
-def test_module_import_registers_cleanup_handler():
-    module_path = Path(_reg_mod.__file__)
-    spec = importlib.util.spec_from_file_location('reader_registry_test_copy', module_path)
+def _load_module_copy(module_name: str, module_path: str | Path):
+    spec = importlib.util.spec_from_file_location(module_name, module_path)
     assert spec is not None
     assert spec.loader is not None
 
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_module_import_registers_cleanup_handler():
     with patch('atexit.register') as register:
-        module = importlib.util.module_from_spec(spec)
-        spec.loader.exec_module(module)
+        module = _load_module_copy('reader_registry_test_copy', _reg_mod.__file__)
 
     register.assert_called_once_with(module._cleanup_temp_files)
     assert module._custom_ext_readers == {}
@@ -335,29 +337,17 @@ def test_temp_file_cleanup(tmp_path):
     assert len(_reg_mod._temp_files) > initial
 
 
-def test_top_level_exports_survive_reload(tmp_path):
-    importlib.reload(_fileio_mod)
-    utilities_module = importlib.reload(pv_utilities)
-    pyvista_module = importlib.reload(pv)
-
-    assert pyvista_module.register_reader is _reg_mod.register_reader
-    assert pyvista_module.has_scheme is _reg_mod.has_scheme
-    assert pyvista_module.LocalFileRequiredError is _reg_mod.LocalFileRequiredError
-    assert utilities_module.register_reader is _reg_mod.register_reader
-    assert utilities_module.has_scheme is _reg_mod.has_scheme
-    assert utilities_module.LocalFileRequiredError is _reg_mod.LocalFileRequiredError
-
-    test_file = tmp_path / 'data.reloadext'
-    test_file.touch()
-    mock = MagicMock(return_value=pyvista_module.PolyData())
-    pyvista_module.register_reader('.reloadext', mock)
-
-    pyvista_result = pyvista_module.read(str(test_file))
-    utilities_result = utilities_module.read(str(test_file))
-
-    assert isinstance(pyvista_result, pyvista_module.PolyData)
-    assert isinstance(utilities_result, pyvista_module.PolyData)
-    assert mock.call_args_list == [
-        call(str(test_file.resolve())),
-        call(str(test_file.resolve())),
-    ]
+def test_top_level_exports_available_in_fresh_python_process():
+    command = (
+        'import pyvista as pv; '
+        'import pyvista.core.utilities as utilities; '
+        'assert pv.register_reader is utilities.register_reader; '
+        'assert pv.has_scheme is utilities.has_scheme; '
+        'assert pv.LocalFileRequiredError is utilities.LocalFileRequiredError'
+    )
+    subprocess.run(
+        [sys.executable, '-c', command],
+        check=True,
+        capture_output=True,
+        text=True,
+    )


user27182:
Thanks for this. Users of DICOM and NIFTI formats will also be happy with this since the native vtk readers tend to be insufficient. I can add a follow-up example to demonstrate this (I suppose it would be similar to the zvtk example)

Is there a plan to add a similar registry for writers as well? To work with `mesh.save()`

banesullivan:
Yes, I think we should add a similar registry for writers but think we should handle readers first and see what issues come up. Plus writers felt a little more complicated since we have to map a pyvista dataset type to the writers so I figured they should be separate chunks of work

user27182:
Adding `fsspec` to this list should resolve test failures:
https://github.com/pyvista/pyvista/blob/1c51c169fe119c207ee81e4784dff6b6c9c7d1da/pyvista/report.py#L289

user27182:
> ## [Codecov](https://app.codecov.io/gh/pyvista/pyvista/pull/8437?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pyvista) Report
> ❌ Patch coverage is `80.48780%` with `24 lines` in your changes missing coverage. Please review. ✅ Project coverage is 96.49%. Comparing base ([`0626932`](https://app.codecov.io/gh/pyvista/pyvista/commit/062693283bd0acb50e45ade38d9b2de7ba62b154?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pyvista)) to head ([`2e7f9e1`](https://app.codecov.io/gh/pyvista/pyvista/commit/2e7f9e138071c296b811c9c092e7c66ce87eea45?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pyvista)). ⚠️ Report is 8 commits behind head on main.
> 
> Additional details and impacted files
> ```diff
> @@            Coverage Diff             @@
> ##             main    #8437      +/-   ##
> ==========================================
> - Coverage   96.63%   96.49%   -0.15%     
> ==========================================
>   Files         162      163       +1     
>   Lines       35512    35604      +92     
>   Branches     4503     4512       +9     
> ==========================================
> + Hits        34318    34356      +38     
> - Misses        575      623      +48     
> - Partials      619      625       +6     
> ```

Is it just me, or is it no longer possible to view coverage report details for which lines are missing?

I can see this summary:
image

But when I try to drill down and review the affected files, I get an error
image

banesullivan:
It's not just you. I noticed this on my other recent PRs too

banesullivan:
Um. And I just got a notification on this PR from codecov https://github.com/pyvista/pyvista/pull/228#issuecomment-4188964011

banesullivan:
I have removed codecov from the required checks for now

user27182:
> I have removed codecov from the required checks for now

I'm not sure if this is a good thing... 24 lines missing coverage seems like a lot, I don't think we should ignore this. IMO we should resolve the issue with the codecov report, or maybe upload the reports as an artifact so we can download and manually review to see which lines are missing coverage, and add or update tests accordingly.

banesullivan:
Re-enabled the required check. Will see what I can do to track down the uncovered changes and resolve here

Otherwise, yeah we may just have to upload the coverage report to the build artifacts

PR #24, opened 2018-12-19
Add wrappers for vtkImageData/vtkRectilinearGrids and a whole lot more

banesullivan:
Overall, this pull request is adding wrappers for `vtkImageData` and `vtkRectilinearGrids` as well as  a few features I found useful:

## `vtki.wrap`

A new wrapping method under the `utilities` module. This allows users to quickly wrap any VTK dataset they have to its appropriate `vtki` object:

```py
import vtk, vtki
stuff = vtk.vtkPolyData()
better = vtki.wrap(stuff)
```

## `vtki.Common._repr_html_` and a few other methods

A new way to represent the object in Jupyter notebooks that might be a bit more insightful that printing the class name and memory address. See below:

screen shot 2018-12-18 at 9 54 47 pm

## `vtki.get_scalar`

A helper to search both the Point and Cell data for an array by its name. This allows users to now pass an array name to the `plot`/`add_mesh` method so that it will find the Cell or Point data with that name and then color the dataset by that array.

screen shot 2018-12-18 at 9 52 41 pm

screen shot 2018-12-18 at 9 52 30 pm

## Other updates

- I fixed the ReadTheDocs badge and cleaned up the README

### Jupyter Plotting

I really, REALLY want to get the plotting code working for interactive plots in Jupyter notebooks. I think I can get some traction on this from both myself and a few other open-source developers using VTK. 

What is needed to implement interactive 3D plots in Jupyter notebooks? Can we just open the rendering window into a frame in the notebook?

### Other Wrappers

I have a branch started for adding a `vtkMultiBlockDataSet` wrapper. This will be necessary for me to use `vtki` with my project [**`PVGeo`**](https://github.com/OpenGeoVis/PVGeo) so hopefully, we can collaborate to make a wrapper for multiblock datasets.

### Wrapping VTK algorithms

I have been thinking about wrapping all of the VTK algorithms/filters into a common API such that users can call the filters/ writers in a more Pythonic manner. **Is this something that might fit into `vtki`?** For example, we could wrap the `vtkThreshold` to go from:

```py
import vtk
alg = vtkThreshold()
alg.SetInputDataObject(pdi)
alg.ThresholdByLower(99.0)
alg.Update()
pdo = alg.GetOutputDataObject(0)
```

to this easier API:

```py
import vtki
pdo = vtki.algs.Threshold(pdi, 99.0, option='lower')
```

## Why am I doing this?

[...]

PR #8443, opened 2026-04-05
Modernize HTML repr for Jupyter notebooks

banesullivan:
The Jupyter notebook HTML representation for PyVista objects has looked largely the same since I first introduced it in #24 over 7 years ago. It was a plain `<table>` with inline styles, no interactivity, and no visual hierarchy. Meanwhile, libraries like xarray have set a higher bar for what a modern notebook repr looks like. It's time to bring PyVista's repr into the same era.

Screenshot 2026-04-05 at 1 40 06 AM

- Replace the plain HTML table repr across `DataSet`, `MultiBlock`, `PartitionedDataSet`, `Table`, and `CameraPosition` with a modern, theme-aware design featuring collapsible data array sections, header badges (n_points, n_cells, memory size), a metadata block for bounds and grid properties, and SVG mesh-type icons. The expand/collapse mechanism is pure CSS (hidden checkboxes, no JavaScript) following the pattern xarray established. The only JS is a minimal `navigator.clipboard.writeText` for copy-to-clipboard buttons.
- Add a shared `pyvista/core/formatting_html.py` module that all repr implementations use, keeping each class's `_repr_html_` method lean. All HTML content is escaped via `html.escape()` for XSS safety, and a `<pre>` text fallback is included for non-HTML environments like nbviewer.
- Create hand-crafted SVG icons for each mesh type (PolyData, UnstructuredGrid, ImageData, RectilinearGrid, StructuredGrid, MultiBlock) displayed in the repr header. These are tiny (1-2 KB each) isometric illustrations that give immediate visual recognition of the dataset type. A "PyVista" wordmark SVG sits in the top-right corner too (our primary SVG logo is way too large to use here).
- Add a theme-aware CSS stylesheet (`pyvista/core/static/style.css`) with CSS custom properties that automatically adapt to JupyterLab, VS Code, and generic dark/light modes. Badge colors distinguish active scalars, vectors, normals, and texture coordinates at a glance. Active scalar array names are bold.
- Show richer metadata everywhere: PolyData displays cell-type breakdown (faces, lines, strips, verts), grid types show dimensions and spacing, MultiBlock children list per-block point/cell counts and memory, nested MultiBlocks show block count and memory, field data arrays display their full shape since they're arbitrary length, and bounds are formatted per-axis with a copy button for the full tuple.

### DIFF
--- a/pyvista/core/dataset.py
+++ b/pyvista/core/dataset.py
@@ -1681,7 +1681,7 @@ class DataSet(DataSetFilters, DataObject):
                 ncomp = coerced.shape[1] if coerced.ndim > 1 else 1
                 shape = str(tuple(coerced.shape)) if show_shape else ''
                 range_str = ''
-                if show_range and np.issubdtype(coerced.dtype, np.number):
+                if show_range and coerced.size > 0 and np.issubdtype(coerced.dtype, np.number):
                     lo = fmt.format(np.nanmin(coerced))
                     hi = fmt.format(np.nanmax(coerced))
                     range_str = f'[{lo}, {hi}]'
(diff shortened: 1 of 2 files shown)


banesullivan:
Going ahead with a merge on this to validate the new merge queue. If this breaks or requires changes I'll be sure to quickly address in follow on PRs

PR #8444, opened 2026-04-05
Add browser interactivity test for interactive plots

### LINKED ISSUE #8442: Interactive plots are not responsive
_Originally posted by @bjlittle in [#7721](https://github.com/pyvista/pyvista/issues/7721#issuecomment-4187478727)_

Compare the interactive plot from dev docs:
https://dev.pyvista.org/api/utilities/_autosummary/pyvista.sphere

To the one from stable docs:
https://docs.pyvista.org/api/utilities/_autosummary/pyvista.sphere

The plot from dev docs is no longer interactive and does not respond to user input.

user27182:
### Overview

Add interactivity tests and resolve pyvista/trame-pyvista#29


user27182:
New test fails as expected with `trame-vtk==2.11.6`:
https://github.com/pyvista/pyvista/actions/runs/24000528138/job/69995721864?pr=8444#step:8:4880

user27182:
and new test passes with `trame-vtk==2.10.2` https://github.com/pyvista/pyvista/actions/runs/24000916650/job/69996692669?pr=8444

bjlittle:
@user27182 This is lovely 💯

akaszynski (dismissed):
That's a smart way of doing it. Just for a sanity check, would it be smart to take two screenshots in a row with no interactivity (maybe a click and that's it or another no-op user interaction) and then interact with it.

Otherwise this is good to go. Up to you if you want to implement that.

user27182:
Feel free to push any changes to improve the test!

tests/plotting/test_tinypages.py
```diff
@@ -176,3 +181,74 @@ def test_parallel(tmp_path: Path) -> None:
...
+        text=True,
+        check=False,
+    )
+
+    assert result.returncode == 0, result.stderr
+
+    old_cwd = Path.cwd()
+    os.chdir(html_dir)
+
+    server = ThreadingHTTPServer(('127.0.0.1', 8000), SimpleHTTPRequestHandler)
```
user27182: Not sure if we need to modify this to avoid this “address already in use” error
https://github.com/pyvista/pyvista/actions/runs/24363360612/job/71148638734?pr=8444#step:8:2624


Or maybe add a marker to ensure the tiny pages tests run serially on the same worker

``` py
pytestmark = pytest.mark.xdist_group("tinypages")
```

Or maybe we just add the `@flaky_test` decorator?

banesullivan: Oh yes we definitely should let it choose a port automatically. I think just setting `0` for the port will do it?

tox.ini
```diff
@@ -67,6 +67,7 @@ commands_pre =
     {[testenv]software_report_cmdline}
 
 commands =
+    playwright install --with-deps chromium
```
banesullivan: ooof, I'm used to playwright take a bit to install but looking at CI logs doesn't seem to add much time if I'm reading this right.

banesullivan (commented):
I'm not terribly enthusiastic about this one... playwright tests like this are notoriously brittle in my experience. However, if this proves to be a stable test, then great! But if not, it's probably not worth the maintenance overhead from CI flakiness and I'd be quick to suggest 🪓'ing it.

To alleviate my concerns, could we:
1) isolate this playwright test/workflow to its own job and 
2) that job **not** be a **required check** for PRs

Feeling stronger abut 1 than 2, but if it proves flaky then its easy to go in and disable 2 on the fly to bring CI back online.

In an ideal world, I'd say lets implement better CI testing in trame-vtk itself... I'm just not sure how feasible/realistic/welcomed that is

user27182:
> I'm not terribly enthusiastic about this one... playwright tests like this are notoriously brittle in my experience. However, if this proves to be a stable test, then great! But if not, it's probably not worth the maintenance overhead from CI flakiness and I'd be quick to suggest 🪓'ing it. 

To be fair I have only seen one OSError so far in terms of flakiness. But indeed there are other ways this could fail with future updates, and while generating this test I found it essential to use `headless=False` for debugging…. So if there was a future breakage for some reason this kind of debugging would definitely be a maintenance burden. 


> To alleviate my concerns, could we:
> 
> 1. isolate this playwright test/workflow to its own job and
> 2. that job **not** be a **required check** for PRs

We should probably do both. This is mostly only needed to prevent breakage from `trame-vtk` updates. But the functionality is essential, and I would be hesitant to remove this test completely unless/until we are confident this kind of breakage won’t happen again. 

> In an ideal world, I'd say lets implement better CI testing in trame-vtk itself... I'm just not sure how feasible/realistic/welcomed that is


There appears to be a desire for this to be supported upstream  https://github.com/Kitware/trame-vtk/issues/109

I think the main blocker is probably creating a minimal example to reproduce the failure without PyVista. I did initially try making a somewhat minimal one here (still depends on PyVista):

https://github.com/pyvista/pyvista/issues/8442#issuecomment-4188129654


But there’s something specific to the `tinypages` tests that was needed to reproduce. I don’t really have the expertise or bandwidth to dig into this much further. 

EDIT: fix quotes

tox.ini
user27182: It would be better not to need to install playwright. The main thing with this I think is the need to download the browser, which isn’t too bad but is an issue for offline tests. And using `tox` is a must due to the need for install, so regular pytest invocations in a clean environment will break. 

And for our self-hosted runners `—with-deps` is also needed, which may add extra overhead. Though like you say it doesn’t appear to be substantial.

banesullivan:
Thanks for putting this together and helping dig into these trame-vtk failures. Indeed we **need** to be sure a failure like this doesn't happen again and for that reason I'm happy to see this PR land. I say lets isolate the test in its own job and keep an eye on it for trame upgrades

user27182:
> I say lets isolate the test in its own job

@banesullivan this has been isolated to `tox run -e playwright`. It currently does not run at all. How/when do you want this to run for CI? When a label is applied...? On pushes to main? Should we follow what's done with integration tests? Maybe we can also auto-run when any dependencies change? Feel free to push any changes accordingly.

banesullivan:
I view this as a bit of an integration test, so I think we should follow the same pattern as the integration workflow or place it in that workflow directly

user27182:
I can no longer push changes to workflows locally. This is understandable from a security perspective, but it is annoying since I must now use the GitHub GUI to make changes and push commits (I had already resolved conflicts and had changes pending but these have been rejected).

user27182:
FYI @edabor, please review tox config

user27182:
New `playwright` integration tests are passing
https://github.com/pyvista/pyvista/actions/runs/24624262243/job/72000332265?pr=8444

tox.ini
```diff
@@ -113,6 +113,29 @@ description =
...
+# ================= PLAYWRIGHT TESTING ENVIRONMENT =================
+# This environment is specialized to run PyVista tests that require
+# playwright. It is treated like an integration test, but it is not
+# a true integration test in the sense that a 3rd party library is
+# not being tested. The "integration" part is mainly relevant for CI
+# since this test is treated just like an integration test and is
+# run in the same manner and with the same frequency.
+[testenv:integration-playwright]
+description = Run tests that use playwright
+basepython = py3.13
```
banesullivan: Should this upgrade to py3.14? geovista is only pinned to py3.13 temporarily and then this would be the odd one out

banesullivan (dismissed):
This looks clean to me. I feel good merging as is but will defer on the tox config conventions. py3.13 vs py3.14 isnt too important and we can update when we upgrade geovista

tox.ini
user27182: Yes this was copied from before #8314 landed

user27182:
```suggestion
basepython = py3.14
```

tox.ini
```diff
@@ -113,6 +113,29 @@ description =
...
+
+setenv =
+    PYTEST_ADDOPTS = --color=yes -v
```
edabor:
```suggestion
```
Already inherited from base `testenv`

edabor (commented):
Maybe nitpick, but we could move this test to the `examples_trame/tests`.
This would remove the need to add pytest specific marker + CLI option in the `pyvista` test suite.

user27182:
> Maybe nitpick, but we could move this test to the `examples_trame/tests`. This would remove the need to add pytest specific marker + CLI option in the `pyvista` test suite.

It depends on how closely we want to couple this test with `trame-vtk`. If we ever switch to using web assembly, we should arguably continue testing interactivity, but moving the test to `examples_trame` would cause it to no longer test our interactive docs.

Plus, the initial docs build process for this test was copy/pasted from the tinypages test directly above this one, so in that sense it makes sense to keep this in `test_tinypages.py`.

edabor:
> Plus, the initial docs build process for this test was copy/pasted from the tinypages test directly above this one, so in that sense it makes sense to keep this in `test_tinypages.py`.

Indeed, it makes more sense to put it there.

My concern was that we add a specific marker + CLI option for one test only.
But this is trivial, considering the benefits of interactively testing the plots. Perfect is the enemy of good.

edabor (approved):

banesullivan (approved):

Issue #8428, opened 2026-03-31
Inline Terminal Rendering for PyVista


banesullivan:
### Describe the feature you would like to be added.

PyVista should be able to render meshes inline in the terminal, the same way matplotlib can display plots in IPython/Jupyter terminal sessions. When a user calls `mesh.plot()` or uses the PyVista CLI in a capable terminal (iTerm2 or any Sixel-capable terminal), the rendered image should appear directly in the terminal output instead of opening a GUI window.

This is useful for:

- **SSH sessions** where X11 forwarding is unavailable or slow
- **CI/CD pipelines** where you want visual verification in logs
- **IPython console** workflows where popping a GUI window is disruptive

The implementation is small. The core is ~80 lines of code: render off-screen with `Plotter(off_screen=True)`, capture the framebuffer as a PIL Image, and emit it to the terminal using the appropriate image protocol (iTerm2 inline images, Sixel, or halfblock fallback via `textual-image`).

```python
import pyvista as pv
from pyvista import examples

pv.OFF_SCREEN = True
mesh = examples.download_fea_bracket()
mesh.plot(scalars="Equivalent (von-Mises) Stress (psi)")

# Or via the existing CLI
# pyvista render mesh.vtk --scalars temperature --cmap viridis --off-screen
```

PyVista will detect if `off_screen=True`, `notebook=False`, and no screenshot was requested so the user must want to display the output inline in the terminal (otherwise the plot is silently disposed). Similarly, the CLI would have an `--off-screen` parameter for this. I am concerned that this could incorrectly fire during `pv.GPUInfo()` or during tests, so we may want to have an explicit opt-in or opt-out flag... to be determined...

New dependencies: `textual-image` for terminal protocol detection. The iTerm2 path shouldn't need additional dependencies beyond the standard library.

### FYI `pyvista-tui`

I recently made [`pyvista-tui`](https://github.com/pyvista/pyvista-tui) as an interactive terminal-based 3D viewer with keyboard controls, themes, animations, and a full TUI. That's a lot of machinery for "I just want to see my mesh in the terminal." Honestly, `pyvista-tui` started as a bit of a joke wanting to render meshes to ASCII characters but I have also wanted to be able to static render without leaving the command line or having a popup window for years and so the project was born. Theres a LOT of cruft in there that PyVista does not need.

The core rendering-to-terminal capability should live in PyVista itself because:

1. It's easy (see code is below)
[...]

PR #8445, opened 2026-04-05
Show failed images from unit tests in CI summary

user27182:
### Overview

Inspired by https://github.com/pyvista/pyvista/issues/8428#issuecomment-4166880143, except this PR uses a markdown table and does not use inline terminal

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -184,11 +184,11 @@ jobs:
       - name: Show Images for Failed Tests
         if: always()
         run: |
-          {
-            echo '| File | From cache | From test |'
-            echo '|------|------------|-----------|'
-            echo '| cat | ![](https://upload.wikimedia.org/wikipedia/commons/3/3a/Gustav_chocolate.jpg) | ![](https://upload.wikimedia.org/wikipedia/commons/3/3a/Gustav_chocolate.jpg) |'
-          } >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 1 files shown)


banesullivan:
What if we created a static HTML page with the test info, error message if any, baseline image and the failed image for all failed tests and uploaded that to netlify like we do for the PR preview documentation? (see comment above)

Having it in a markdown comment would be nice, I'm just afraid it may crowd the discussion here

user27182:
Yea I don't think it's going to work to try to show the images in CI directly with HTML, markdown, or even with inline rendering from #8428, I'm thinking this just isn't well supported by the GitHub CI shell

Including it in Preview docs would work very well, Sphinx Gallery already publishes a similar table with execution times, e.g. https://69d27a8f34e32ce82f27674f--pyvista-dev.netlify.app/sg_execution_times.html. This would pretty much give us exactly what we want. The main downsides I see with this though is that it requires a successful docs build to work, otherwise the preview step doesn't run. And docs preview isn't supported for PRs from forks either.

It is possible to publish to a _separate_ HTML page, separate from the docs? Something we can run at the end of each unit testing job?

banesullivan:
We can create a second netlify site and deploy the static HTML there seperately from the docs build

banesullivan:
This should be pretty easy to set up. I'll see if I can prototype this on another branch quickly today: I'll create the new netlify site and upload a dummy HTML page from CI with a separate comment.

PR #8446, opened 2026-04-05
Add interactor style registry and theme entry

banesullivan:
I'm building a downstream package with custom interactor styles that I plan to open source. This PR adds the registry infrastructure in PyVista so that downstream packages can register styles and have them work seamlessly with the theme system and plotter API.

closes #7359

- Add a named interactor style registry (`pyvista.register_interactor_style`) that lets downstream packages register custom interactor styles programmatically or via the `pyvista.interactor_styles` entry-point group
- Add `Theme.interactor_style` property so themes control the default interactor style by name (defaults to `'trackball_style'`, preserving current behavior)
- Add `enable_interactor_style(name)` to `RenderWindowInteractor` and `BasePlotter` to activate any built-in or registered style by name
- Accept strings in the `RenderWindowInteractor.style` setter, resolving them through the registry (e.g., `pl.iren.style = 'terrain_style'`)
- Plotter init now uses `enable_interactor_style()` which reads the theme, replacing the hardcoded `enable_trackball_style()` call
- Entry point discovery is lazy and cached, with warnings for load failures, name collisions, and duplicate providers
- Built-in style names mirror the existing `enable_*_style` methods and are protected from collision
- Comprehensive test coverage for registration, validation, entry point discovery (including all warning paths), theme serialization, and plotter integration

### Usage

A downstream package registers its styles at import time or via entry points:

```py
import pyvista as pv

def my_custom_style(interactor):
    """Factory that returns a vtkInteractorStyle instance."""
    ...

pv.register_interactor_style('my_custom_style', my_custom_style)
```

Users can then set the style through the theme, the plotter, or the interactor:

```py
# Via the theme (applies to all new plotters)
pv.global_theme.interactor_style = 'my_custom_style'

# Via the plotter method
pl = pv.Plotter()
pl.enable_interactor_style('my_custom_style')

# Via the interactor style setter
pl.iren.style = 'my_custom_style'
```

Packages can also use the `pyvista.interactor_styles` entry-point group in `pyproject.toml` so styles are discovered automatically when the package is installed. The entry point should resolve to a dict mapping style names to handler callables:

```toml
[project.entry-points."pyvista.interactor_styles"]
my_package_styles = "my_package:INTERACTOR_STYLES"
```

```py
# my_package/__init__.py
INTERACTOR_STYLES = {
    'my_custom_style': custom_style_factory,
    'my_other_style': other_style_factory,
}
```

A single callable entry point is also supported, in which case the entry point name is used as the style name

### DIFF
--- a/.github/path-filters.yml
+++ b/.github/path-filters.yml
@@ -1,0 +1,83 @@
+# Shared path filters for CI skip conditions
+# Used by dorny/paths-filter in docs.yml, testing-and-deployment.yml, and integration-tests.yml
+#
+# DENYLIST approach with predicate-quantifier: 'every'
+# A file triggers the filter only if it matches ALL patterns (i.e., is not excluded by any).
+# This means any new/unknown file triggers CI by default — the safe choice.
+#
+# To exclude a path, add a negation pattern: "!path/**"
+# Do NOT add positive patterns (like "**") — they break 'every' logic.
+
+tests:
+  # Directories that don't affect unit tests
+  - "!doc/**"
+  - "!joss/**"
+  - "!logo/**"
+  - "!assets/**"
+  - "!docker/**"
+  - "!.devcontainer/**"
+  # Root files that don't affect unit tests
+  - "!README.rst"
+  - "!CONTRIBUTING.rst"
+  - "!AUTHORS.rst"
+  - "!CODE_OF_CONDUCT.md"
+  - "!SECURITY.md"
+  - "!FUNDING.yml"
+  - "!CITATION.cff"
+  - "!CITATION.rst"
+  - "!LICENSE"
+  - "!codecov.yml"
+  - "!lychee.toml"
+  - "!.pre-commit-config.yaml"
+  - "!.mailmap"
+  - "!.gitignore"
+  - "!.gitattributes"
+
+integration:
+  # Directories that don't affect integration tests
+  - "!doc/**"
+  - "!joss/**"
+  - "!logo/**"
+  - "!assets/**"
+  - "!docker/**"
+  - "!.devcontainer/**"
+  # Root files that don't affect integration tests
+  - "!README.rst"
+  - "!CONTRIBUTING.rst"
+  - "!AUTHORS.rst"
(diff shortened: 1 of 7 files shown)


MatthewFlamm:
Have you looked at #5119 ?  There seems to be overlap here.  I no longer have time to get that PR finished.  But it might be worth taking a look at the discussion there, or the implementation there to get some ideas (or reject them).

banesullivan:
I had forgotten about https://github.com/pyvista/pyvista/pull/5119, thanks for reminding me @MatthewFlamm. I'll review it again and see we can incorporate some of the same idea

MatthewFlamm:
Oh wait this is interactor style registering.  I latched onto the theme part.  I will modify my comment that #5119 proposed a way to register custom themes.  Having a consistent method to register custom items for downstream packages should be considered, whatever that method is.

akaszynski (dismissed):

pyvista/plotting/interactor_style_registry.py
```diff
@@ -0,0 +1,265 @@
...
+from typing import TYPE_CHECKING
+from typing import TypedDict
+
+from pyvista._warn_external import warn_external
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+    from typing import Any
+
+    InteractorStyleFactory = Callable[..., Any]
```
MatthewFlamm: What is the need for a factory here?  Why can we not pass in directly?  After #7497, it should be now possible to use an object/class.

banesullivan: This new registry adds the ability to reference styles *by name*, which is needed for themes (including serialized JSON themes) and for entry-point discovery from downstream packages. To clarify that this isn't really a factory anymore, I renamed `InteractorStyleFactory` to `InteractorStyleHandler` and updated the docs to make it clearer that you just register a class directly (or a name) and no factory wrapper is needed.

MatthewFlamm: Thanks.  I don't have time to do an in depth review, but making it clear that the class can be registered is what I was getting to.  I'm good with that plan.

PR #8448, opened 2026-04-05
Refactor FRD parsing to handle multiple formats

### LINKED ISSUE #8447: Extension of the reading of the frd format (shrot and long)
There are two standard length for the node and element numbers: 5 and 10, now pyvista only reads long versions correctly: 
Image

cgx 2.23 manual (page 144, https://www.dhondt.de/cgx_2.23.pdf): "The ascii format is able to store element- and node-numbers up to ’99999’ in the short form or up to ’9999999999’ in the long form."

nodes:
Image

elements:
Image

**the same mesh but in short notation:**
Image

nodes:
Image

elements:
Image

frd from ccx (endfile):
Image

frd from cgx (endfile):
Image

ccx.frd output from ccx-solver (long version) and mesh.frd output from cgx-prepost (only mesh for preview, short version):
[short_and_long.zip](https://github.com/user-attachments/files/26490411/short_and_long.zip)

3rav:
Closes https://github.com/pyvista/pyvista/issues/8447


3rav:
image


By the way, maybe you've seen something like this before (this kind of problem, probably not related to frd):
https://github.com/Kitware/trame/discussions/868

pyvista/core/utilities/_frd.py
```diff
@@ -247,14 +272,48 @@ def _parse_elements(file_stream: Any, frd_data: _FRDData) -> None:
...
+                    node_ids.extend(new_nodes)
+                else:
+                    # Fallback to space-separated values for PyVista's mock test suite files
+                    for p in data_str.split():
+                        with contextlib.suppress(ValueError):
+                            node_ids.append(int(p))
 
                 if (n_nodes := len(node_ids)) < needed:
-                    # Element has too few points, and might be invalid, or maybe more points
-                    # are defined on the next line
```
akaszynski, line 254: Any reason why these comments were removed? As a non-expert in this file type I think they'r equite helpful.

tests/core/test_frd.py
```diff
@@ -122,7 +122,7 @@ def comprehensive_frd_file(tmp_path):
 @pytest.fixture
 def no_steps_frd_file(tmp_path):
     """Fixture to trigger empty step checks."""
-    content = '1C\n2C\n -1 1 0.0 0.0 0.0\n -3\n'
+    content = '1C\n2C\n -1    1 0.0 0.0 0.0\n -3\n'
```
akaszynski, line 125: Any reason why we've added these whitespace changes (here and all over the place in the PR). Would prefer reverting if it's just cosmetic.

pyvista/core/utilities/_frd.py
3rav: You're absolutely right, it's a very helpful piece of context! It was accidentally dropped during the parser loop refactor. I'll add it right back.

tests/core/test_frd.py
3rav: These changes actually aren't just cosmetic. Real CalculiX FRD files don't use arbitrary spaces; they use a strict fixed-width format (either 5 or 10 characters per column). The refactored parser now explicitly detects and chunks the data based on this fixed width to prevent bugs with glued numbers (e.g., -1123.0 instead of -1 123.0).

I updated the test fixtures with this authentic spacing so our tests run against realistic FRD output formats, rather than forcing the parser to constantly fall back to the space-separated fail-safe logic designed for older mock tests.

akaszynski: Ah, that's a real pain then. Thanks for letting me know and I'd add a note in the file (unless I missed it).

3rav: I can mark this as "resolve"?

akaszynski: Yep!

pyvista/core/utilities/_frd.py
```diff
@@ -247,14 +272,50 @@ def _parse_elements(file_stream: Any, frd_data: _FRDData) -> None:
...
+                    for p in data_str.split():
+                        with contextlib.suppress(ValueError):
+                            node_ids.append(int(p))
 
                 if (n_nodes := len(node_ids)) < needed:
                     # Element has too few points, and might be invalid, or maybe more points
                     # are defined on the next line
                     continue
+
                 if n_nodes > needed:
```
akaszynski, line 318:
```suggestion
                if n_nodes > needed:
                # Keep track of elements with too many points, but don't skip
```

akaszynski:
@MatthewFlamm, @user27182, this looks fine to me, but I'm not an expert with this reader. Mind chiming in?

user27182:
I also don’t know the format but will review in a few days

pyvista/core/utilities/_frd.py
```diff
@@ -133,6 +134,10 @@ class _FRDData:
     _has_too_few_points: list[_InvalidElement] = field(default_factory=list)
     _has_unsupported_element: list[_InvalidElement] = field(default_factory=list)
 
+    # Format detection
+    is_long_format: bool = False
```
user27182:
```suggestion
    _is_long_format: bool = False
```

pyvista/core/utilities/_frd.py
```diff
@@ -189,12 +194,24 @@ def _parse_nodes(file_stream: Any, frd_data: _FRDData) -> None:
             s = line.strip()
             if s.startswith(end_block):
                 return
-            try:
-                parts = _FRDParser._fix_scientific(s).split()
-                nid = int(parts[1])
-                frd_data.nodes[nid] = [float(parts[2]), float(parts[3]), float(parts[4])]
-            except (ValueError, IndexError):
-                pass
+            if not s.startswith('-1'):
```
user27182: Use enum value instead of a magic number `-1`.

pyvista/core/utilities/_frd.py
```diff
@@ -189,12 +194,24 @@ def _parse_nodes(file_stream: Any, frd_data: _FRDData) -> None:
...
+                continue
+
+            idx = line.find('-1')
```
user27182: Use enum value

pyvista/core/utilities/_frd.py
```diff
@@ -189,12 +194,24 @@ def _parse_nodes(file_stream: Any, frd_data: _FRDData) -> None:
...
+            if idx == -1:  # pragma: no cover
```
user27182: Use enum value. And why no cover?

pyvista/core/utilities/_frd.py
```diff
@@ -189,12 +194,24 @@ def _parse_nodes(file_stream: Any, frd_data: _FRDData) -> None:
...
+            # split() is completely immune to varying space widths in both tests and real files
+            data_str = _FRDParser._fix_scientific(line[idx + 2 :].rstrip('\n\r'))
+            parts = data_str.split()
+
+            if len(parts) >= 4:
+                try:
+                    nid = int(parts[0])
+                    frd_data.nodes[nid] = [float(parts[1]), float(parts[2]), float(parts[3])]
+                except ValueError:
+                    pass
```
user27182, line 218: What is happening here? What does it mean to have 4 parts? And can you add a comment after ValueError to indicate what this means, and why it's OK to `pass` here.

pyvista/core/utilities/_frd.py
```diff
@@ -210,26 +227,34 @@ def _parse_elements(file_stream: Any, frd_data: _FRDData) -> None:
...
             if s.startswith(elem_def):
                 elem_line_number = file_stream.line_number
-                parts = s.split()
+
+                # Key: remove "-1" before split so that a glued ID does not break the parser
+                idx = line.find('-1')
+                if idx == -1:  # pragma: no cover
+                    continue
+                data_str = line[idx + 2 :]
+                parts = data_str.split()
```
user27182: Spacing seems important, and `split` seems to be causing a number of issues. Is there a better way to parse the data without `split` so that whitespace is preserved? If the widths are fixed, can we just use regular indexing to grab each value (whitespace included), or maybe use regex to find/capture the values with whitespace?

Also, use enum instead of magic numbers, and why no cover? It's OK if there is some missing coverage reported for branches that exist for rare edge cases

user27182 (changes requested):
This long / short format detection seems important, so I'd like to see this PR land, but I have concerns that the implementation is using `split`, which is probably the wrong function to use since it does not preserve whitespace. Because of this, it looks like there are a lot of workarounds needed during parsing to accommodate for this. This makes it hard to understand the code and seems to add a bunch of branches, many of which seem to not have coverage.

Instead, I think the use of `split` should be reconsidered, and another tool used instead to preserve whitespace. The use of magic numbers should also be avoided (use enums instead).

pyvista/core/utilities/_frd.py
3rav: ` # A valid node definition contains at least 4 components: Node ID, X, Y, Z`

pyvista/core/utilities/_frd.py
```diff
@@ -247,12 +274,34 @@ def _parse_elements(file_stream: Any, frd_data: _FRDData) -> None:
...
+                try:
+                    chunks = [data_str[i : i + width] for i in range(0, len(data_str), width)]
+                    new_nodes = [int(c) for c in chunks if c.strip()]
+                    node_ids.extend(new_nodes)
+                except ValueError:
+                    # Fallback for unstructured PyVista mock test files
+                    # where standard fixed-width indexing naturally fails.
+                    for p in data_str.split():
+                        with contextlib.suppress(ValueError):
+                            node_ids.append(int(p))
```
user27182: This comment suggests that this branch is only needed for test purposes. Is this coded not needed otherwise? If the test case isn't representative of real-world data, we should remove this code and the test. Or, if this code is still useful for potential real-world edge cases, then we should keep it, but update the comment to remove the test-specific language.

tests/core/test_frd.py
```diff
@@ -435,3 +440,41 @@ def test_frd_element_sizes(generic_element_frd):
...
+    """Tests parsing edge cases to ensure proper error handling and coverage."""
+    # Invalid elements will raise warnings, which we catch here
+    with pytest.warns(pv.InvalidMeshWarning):
+        reader = pv.FRDReader(coverage_edge_cases_frd)
+
+    mesh = reader.read()
+
+    # Validate that the bad node (Node 1) was successfully skipped
+    assert '1' not in mesh.point_data['original_node_ids']
+    assert '2' in mesh.point_data['original_node_ids']
```
user27182: Can we use the FRD enum values here (I can't recall what Node 1 means)

tests/core/test_frd.py
```diff
@@ -435,3 +440,41 @@ def test_frd_element_sizes(generic_element_frd):
...
+"""
+    file_path = tmp_path / 'cov_edge_cases.frd'
+    file_path.write_text(content, encoding='utf-8')
+    return str(file_path)
+
+
+def test_frd_reader_coverage_edge_cases(coverage_edge_cases_frd):
+    """Tests parsing edge cases to ensure proper error handling and coverage."""
+    # Invalid elements will raise warnings, which we catch here
+    with pytest.warns(pv.InvalidMeshWarning):
```
user27182: Can we include the expected warning message?

```suggestion
    match = '<insert message here>'
    with pytest.warns(pv.InvalidMeshWarning, match=match):
```

user27182 (changes requested):
Thanks for your patience @3rav. I just have a few minor comments, otherwise looks good.

user27182 (approved):

PR #8449, opened 2026-04-05
Add 'cmap', 'clim', and 'lookup_table' parameters to Plotter.add_scalar_bar

### LINKED ISSUE #8424: Convenient way to initialize scalar bar from colormap and range

Recently I ran into a situation where I wanted to make a static scalar bar using a colormap to display during an animation. I am adding and removing several objects with this mapping over the course of the animation so I didn't want to use the scalar bars created by pv.add_mesh, but creating a static scalar bar before adding data uses the default mapper instead of the colorbar that I would like. Maybe there's a way to this that I missed? If not, here's my proposed solution:

Currently, Plotter.add_scalar_bar only accepts a mapper argument (vtkMapper subclass instance) to customize the colors of the bar. It is possible to create a mapper from a colorbar, but it requires creating a lookup table first, requiring three extra steps. I propose two shortcuts to make this process easier: a new _BaseMapper subclass initialized directly from a colormap and the other arguments to a LookupTable called `ColormapMapper`, and a way to pass a colormap & range to Plotter.add_scalar_bar so users don't have to worry about mappers or lookup tables etc.

Unfortunately, unlike existing Mappers in pyvista, there would be no existing vtk implementation of ColormapMapper. In my testing, deriving from vtkDataSetMapper works perfectly, though I suspect there might be some unnecessary overhead of doing so. I wonder if a minimal implementation of the vtkAbstractMapper methods would be easy, though I was unable to parse the C++ documentation well enough to see which methods those would be.

_No response_

### Pseudocode or Screenshots

my ColorbarMapper implementation which has been working for my use:
```python
class ColormapMapper(pv.mapper._BaseMapper,vtk.vtkDataSetMapper):
    #ideally the kwargs would be `wraps`ed to match pv.LookupTable
    def __init__(self, cmap:str, theme=None, **kwargs) -> None:
        super().__init__(theme=theme)
        self.lookup_table = pv.LookupTable(cmap,**kwargs)
```

The add_scalar_bar semantics might be more of a reach.

minerharry:
### Overview

Adds convenience parameters to Plotter.add_scalar_bar to take either a colormap or a LookupTable object as alternatives to a Mapper object, thus allowing the creation of scalar bars independently from mesh data and adding greater flexibility to scalar bar customization without needing the user to instantiate a Mapper object. Resolves #8424.

### Details

Adds 'lookup_table', 'cmap', and 'clim' as keyword arguments to Plotter.add_scalar_bar and ScalarBars.add_scalar_bar
- In Plotter.add_scalar_bar, defaults to the existing Plotter's self.mapper if it exists like original behavior
- In ScalarBars.add_scalar_bar, requires exactly one of 'cmap', 'lookup_table', or 'mapper' to be non-None
- Docstring updated and small example included. TODO: Include a more complex example involving customizing a lookup table in the scalar bar example demo?


akaszynski:
I agree that it would be great to add this; it should be possible to independently add scalar bars to the `pv.Plotter` without relying on an existing mapper.

akaszynski:
Good for review. 3/4 failing due to fork, 1/4 failing due to latest numpy.

pyvista/plotting/scalar_bars.py
MatthewFlamm: `clim` parameter is already used internally

pyvista/plotting/scalar_bars.py
```diff
@@ -197,16 +200,35 @@ def add_scalar_bar(  # noqa: PLR0917
...
+
+            .. versionadded:: 0.48.0
+
+        cmap : str | list, optional
+            Colormap used for the scalar bar.
+
+            .. versionadded:: 0.48.0
+
+        clim : sequence[float], optional
+            Two item range for the scalar bar.
```
MatthewFlamm: I think we need to expand on how to use these optional parameters.  For example, in the current PR, `clim` only is used in conjunction with a `cmap`.  Also, using `cmap` will override the use of `lookup_table`, etc.  Any of these uses will override any `mapper` passed in.  It isn't clear to me without going to the source code how all of this works together.  Even then I might not have a clear understanding.

pyvista/plotting/scalar_bars.py
MatthewFlamm, line 517: What do we expect should happen if the scalar bar title already exists like in this block and one of these new items is passed in?  Does any logic need to be updated or added here?

minerharry:
thanks @akaszynski for picking this up, sorry I let this slip

akaszynski:
> thanks @akaszynski for picking this up, sorry I let this slip

No prob! Still working on it...

pyvista/plotting/scalar_bars.py
minerharry: There are currently two separate checks validating cmap, mapper, and lookup_table, one in Plotter.add_scalar_bar and one in ScalarBars.add_scalar bar.

In ScalarBars.add_scalar_bar, a ValueError is raised if more than one of 'cmap', 'lookup_table', and 'mapper' are specified - very crude, but it avoids having to document the override order. Added this to the add_scalar_bar docstring in several places.

At the moment, the validation check in Plotter.add_scalar_bar is only concerned with passing in the saved mapper if there's no argument specified. We could also copy the logic from ScalarBars.add_scalar_bar so it's obvious from reading the code, but for now I changed the comment to make it more explicit that the check for more than one happens in ScalarBars.add_scalar_bar.

pyvista/plotting/scalar_bars.py
minerharry: updated internal name to _clim

minerharry: I don't see why it should be different; cmap and lookup_table are convenience parameters to create a mapper, which is then used as normal. If the logic worked for passing in a custom mapper, it should work with a cmap or a lookup_table too.

minerharry:
> > thanks @akaszynski for picking this up, sorry I let this slip
> 
> No prob! Still working on it...

oh whoops I did not see this. hopefully nothing conflicts, please nuke my changes if it does

akaszynski:
Just meant I'm not quite done, but I don't have any local changes. Thanks for addressing the outstanding comments.

pyvista/plotting/scalar_bars.py
MatthewFlamm: The documentation additions are good.  Thanks!

pyvista/plotting/scalar_bars.py
MatthewFlamm: I think there may be some edge cases here.  But those might exist without this PR.  So I'm going to resolve this comment for now.  We can always address after this PR, if it is indeed faulty logic in this block.

pyvista/plotting/scalar_bars.py
```diff
@@ -446,14 +480,37 @@ def add_scalar_bar(  # noqa: PLR0917
...
+            )
+            raise ValueError(msg)
+
+        if cmap is not None:
+            lookup_table = pv.LookupTable(cmap=cmap, scalar_range=clim)
+
+        if lookup_table is not None:
+            mapper = pv.DataSetMapper(theme=theme)
+            mapper.lookup_table = lookup_table
+
```
user27182, line 521:
```suggestion

        if cmap is None and clim is not None:
            msg = “`cmap` must be specified when `clim` is provided.”
            raise ValueError(msg)

```

tests/plotting/test_plotting.py
```diff
@@ -1036,6 +1036,40 @@ def test_plot_add_scalar_bar(sphere, verify_image_cache):
...
 
+def test_plot_add_scalar_bar_cmap(verify_image_cache):
+    verify_image_cache.windows_skip_image_cache = True
+
+    pl = pv.Plotter()
+
+    ltable = pv.LookupTable()
+    with pytest.raises(ValueError, match='Exactly one of'):
+        pl.add_scalar_bar(cmap='jet', lookup_table=ltable)
+
```
user27182, line 1064:
```suggestion
    match = “`cmap` must be specified when `clim` is provided.”
    with pytest.raises(ValueError, match=re.escape(match)):
        pl.add_scalar_bar(clim=(0, 1), lookup_table=ltable)

```

pyvista/plotting/scalar_bars.py
user27182: This might not be correct, seems that clim applies to other options too. But reading the docs I did get the impression that these are tightly coupled.

minerharry:
Sorry this has been stagnant. good suggestions

akaszynski (approved):

PR #8451, opened 2026-04-06
Regression image test report on CI

banesullivan:
## Summary

When image regression tests fail in CI, you currently have to download artifact zips and manually compare PNGs to figure out what changed. This PR adds an HTML report that shows baseline vs. generated images side-by-side, deployed to Netlify on every PR with image differences.

The report groups results by CI job and severity (errors, errors as warnings, warnings) and links back to the PR and branch.

Inspired by @user27182's https://github.com/pyvista/pyvista/pull/8445

## Local usage

Run your tests, then generate the report:

```bash
uv run pytest tests/plotting/ -v
tox run -e image-report
open _image_report/index.html
```

Custom input/output paths work too:

```bash
tox run -e image-report -- /path/to/failed_images /path/to/output
```

## CI

A new `image-report` job runs after all test jobs on PRs. If any image differences exist, it generates the HTML report, deploys it to a separate Netlify site, and the bot comments with a preview link. No image differences means no deploy, no comment.

I have intentionally broken some tests here to demonstrate the preview and validate that it works

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -210,13 +210,23 @@ jobs:
           NETLIFY_SITE_ID: ${{ secrets.NETLIFY_DEV_SITE_ID }} # DEV site
         timeout-minutes: 10
 
+      - name: Find existing comment
+        if: github.event_name == 'pull_request'
+        uses: peter-evans/find-comment@3eae4d37986fb5a8592848f6a574fdf654e61f9e
+        id: find-comment
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body-includes: <!-- docs-preview -->
+          token: ${{ secrets.PYVISTA_BOT_TOKEN }}
+
       - name: Comment on PR
         if: github.event_name == 'pull_request'
         uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043
         with:
+          comment-id: ${{ steps.find-comment.outputs.comment-id }}
           issue-number: ${{ github.event.pull_request.number }}
           token: ${{ secrets.PYVISTA_BOT_TOKEN }}
-          body-includes: <!-- docs-preview -->
+          edit-mode: replace
           body: |
             <!-- docs-preview -->
             **Documentation Preview:** ${{ steps.netlify.outputs.deploy-url }}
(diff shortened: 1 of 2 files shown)


tests/plotting/test_plotting.py
banesullivan, line 1: revert all of these changes before merging

banesullivan:
Honestly, this script should be ported to pytest-pyvista so any project can use it

banesullivan:
wow. we have a lot of warnings

Untitled_Export_V2

(meme credit to @akaszynski)

banesullivan:
Got a little carried away with this as its going to be useful for a number of projects of mine with image regression testing. Let's get this ironed out here in pyvista and then transfer it upstream to pytest-pyvista down the line

banesullivan:
Hm.... gotta figure out how to get the CI action to update its original comment

user27182:
> wow. we have a lot of warnings

they are basically just noise at this point. Don’t think the warnings are terribly useful at the moment. Allowing users to customize the warning and error thresholds would help. This would be a new pytest-pyvista feature. 

For our docs image tests where we only build with Linux and latest VTK, the warning threshold should just be the same as the error threshold, which should effectively disable warnings. And for docs maybe 300 should be the error threshold, again since there should be little variance between runs.

user27182:
Some possible inspiration for the HTML report: https://pytest-mpl.readthedocs.io/en/stable/usage.html#html-summary-reports. Seems that matplotlib uses circle CI to host the report

user27182:
Not sure if this is considered as part of the implementation, but there are edges cases:

- Some tests have multiple cached images, stored in a subdir. Which image is shown in this case?
- Sometimes there is a test image but no cache image (or vise-versa). Is this case currently handled?

Also, I think we should include the docs test images in the summary as well. This can be a follow-up, but hopefully it should be straightforward to extend.

banesullivan:
Yeah there are a lot of edge cases and features we should follow up with but I think its worth going ahead and merging this so we can do that in parallel and have this utility available on PRs

user27182:
Can we add this for the docs image tests too? Since they're separate workflows I think we'll need to duplicate most of the deployment steps. But then we'll end up with yet another pyvista bot comment... Might be a bit spammy to get 4 comments on every PR (codecov, docs preview, unit test images, docs test images), not sure if there's a better way to do it.

banesullivan:
Yeah we should definitely enable this for the doc tests as well. It will definitely be spammy or annoying to have three separate comments with link previews. Maybe we can start with that, get it implemented, get the value of having these comparisons easily accessible, and then later try to figure out a way to either make the preview deployment URLs predictable rather than the random hash that they currently use. That way we could have one comment with 3 links

banesullivan:
Scratch that. I have a better solution with CI statuses that will remove the comment. PR incoming

PR #8455, opened 2026-04-06
Skip docs build for unrelated changes

banesullivan:
## Summary

This needs careful review

- Add `paths-ignore` filters to the `pull_request` trigger in three expensive CI workflows so builds are skipped when a PR only touches unrelated files
- Filtered workflows:
  - **docs.yml** — skips for changes to tests, CI workflows, Docker, trame examples, repo metadata, etc.
  - **testing-and-deployment.yml** — skips for changes to docs, CI workflows, repo metadata, etc.
  - **integration-tests.yml** — same exclusions as testing-and-deployment
- All other triggers (push to main, tags, merge queue, schedule, manual dispatch) are unchanged and always run

## Motivation

These three workflows are the most expensive in our CI:

| Workflow | Jobs per PR | Runners |
|---|---|---|
| **docs.yml** | 1 build + test + preview | Self-hosted GPU |
| **testing-and-deployment.yml** | ~16 matrix jobs (5 macOS + 6 Linux + 5 Windows) | Self-hosted + GitHub-hosted |
| **integration-tests.yml** | 4 downstream projects (mne, trame, pyvistaqt, geovista) | Self-hosted |

Currently all three run on **every PR** regardless of what changed. Many PRs only touch files that have no impact on a subset of these workflows:

- A **docs-only PR** doesn't need unit tests or integration tests
- A **test-only PR** doesn't need the docs build
- A **CI config PR** (editing other workflows, dependabot, labeler, etc.) doesn't need any of them
- **Repo metadata changes** (README, CONTRIBUTING, CODE_OF_CONDUCT, SECURITY, etc.) don't need tests

### Design decisions

- **`paths-ignore` (not `paths`)** — the safer, allowlist-based approach. Any new file or directory added to the repo triggers builds by default; you only skip what's explicitly excluded.
- **Only `pull_request` is filtered** — pushes to `main`, tag pushes, scheduled runs, and manual dispatch always run unconditionally.
- **Merge queue is unfiltered** — the `merge_group` trigger has no `paths-ignore`, so the full suite always runs in the merge queue before a PR is actually merged. This is the safety net ensuring nothing slips through even when PR-level checks were skipped.
- **Each workflow excludes its own negation pattern** (e.g., `!.github/workflows/docs.yml`) so changes to a workflow's own config still trigger it.
- **`CONTRIBUTING.rst`, `AUTHORS.rst`, `CITATION.rst`, and `LICENSE` are kept as docs triggers** because they are `.. include::`'d into the Sphinx build via rst directives.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -27,7 +27,6 @@ on: # zizmor: ignore[cache-poisoning]
       - ".mailmap"
       - ".gitignore"
       - ".gitattributes"
-  merge_group:
   workflow_dispatch:
   schedule:
     - cron: "0 4 * * *"


.github/workflows/integration-tests.yml
```diff
@@ -3,6 +3,31 @@ name: Integration Tests
...
+      - "CITATION.cff"
+      - "CITATION.rst"
+      - "LICENSE"
+      - "codecov.yml"
+      - "lychee.toml"
+      - ".pre-commit-config.yaml"
+      - ".mailmap"
+      - ".gitignore"
+      - ".gitattributes"
+  merge_group:
```
banesullivan:
```suggestion
```

banesullivan:
The only way we could do that is to actually run the job and have it read a file and determine if the rest of the job could be skipped. I don't like this though because then we get a false positive for CI status rather than it being marked as skipped

banesullivan:
I'd rather have these hardcoded like this and deal with the maintenance burden than have confusing or misleading status checks

PR #8456, opened 2026-04-06
Add ability to update scalar bar title

banesullivan:
Closing out some of my old PRs and noticed #5998 never landed.

I noticed the `mesh_scalars` trame example wasn't fully working, changing the selected scalars updated the mapper but didn't call `set_active_scalars` on the mesh or update the scalar bar title. I had a stale draft PR (#5998) that partially addressed this with a TODO for the title update, but there was no PyVista API for it. The only option was dropping down to VTK's `SetTitle` directly, which isn't great.

`ScalarBars.update_title(old_title, new_title)` handles all the internal bookkeeping so users don't have to think about it.

- Add `ScalarBars.update_title` method to rename an existing scalar bar, re-keying all internal tracking dicts (actors, ranges, mappers, widgets, slot lookup) atomically
- Fix the `mesh_scalars` trame example: set active scalars on the mesh and update the scalar bar title when the user changes the selected array

### DIFF
--- a/pyvista/plotting/scalar_bars.py
+++ b/pyvista/plotting/scalar_bars.py
@@ -175,6 +175,7 @@ class ScalarBars(_NoNewAttrMixin):
         """Update the title of an existing scalar bar.
 
         .. versionadded:: 0.48.0
+
         Parameters
         ----------
         old_title : str


pyvista/plotting/scalar_bars.py
```diff
@@ -165,6 +165,71 @@ def __contains__(self, key) -> bool:
...
+
+        >>> import pyvista as pv
+        >>> mesh = pv.Sphere()
+        >>> mesh['Data'] = mesh.points[:, 2]
+        >>> pl = pv.Plotter()
+        >>> _ = pl.add_mesh(mesh, scalars='Data')
+        >>> pl.scalar_bars.update_title('Data', 'Elevation')
+        >>> pl.show()
+
+        .. versionadded:: 0.48.0
```
akaszynski: Nitpick, but we should have `versionadded` right after the long description:

```suggestion
```

pyvista/plotting/scalar_bars.py
```diff
@@ -165,6 +165,71 @@ def __contains__(self, key) -> bool:
...
 
+    def update_title(
+        self,
+        old_title: str,
+        new_title: str,
+        *,
+        render: bool = False,
+    ) -> None:
+        """Update the title of an existing scalar bar.
+
```
akaszynski, line 176: See examples section comment below.

```suggestion

        .. versionadded:: 0.48.0
```

PR #8458, opened 2026-04-06
Remove select jobs from the merge queue

banesullivan:
Removes `merge_group` trigger from VTK master testing, Docker image build, and integration test workflows. These jobs were running unconditionally in the merge queue because their `if` conditions only gated on `pull_request` events, letting `merge_group` slip through. The merge queue should only run core CI checks, not nightly, label-gated, or downstream integration tests

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -3,7 +3,6 @@ name: Integration Tests
 # zizmor ignore note: All caching for pushes to main should be disabled with the `USE_CACHE` env var
 on: # zizmor: ignore[cache-poisoning]
   pull_request:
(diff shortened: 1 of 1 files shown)


user27182:
Seems these jobs are blocking merge:

image

banesullivan:
That really doesn't make sense... something else must've happened because these aren't required checks and the merge queue merged several PRs before this one with those VTK dev tests failing. I'm going to just pop it on the queue again and see if its happy. Then hopefully whatever happened doesn't happen again

banesullivan:
Ah it was because the docker build job was marked as required and it timed out waiting for it. I'm disabling that from the required checks

PR #8459, opened 2026-04-07
Actor shader customizations for custom point sprites and max intensity projection

### LINKED ISSUE #1212: Feature Request: Manual Z index ordering for Point Cloud Maximum Intensity Projection
I'd like to replicate the Maximum Intensity Projection ('enhance high values') feature of Leapfrog, which brings sparse points with high scalar values to the foreground. According to Cowan (2014) this is 'easily' achieved by using the z-index ordering capabilities of OpenGL, reordering de z index according to grade value instead of distance from the point of view, even when interacting with the scene. ¿Can z-ordering be accessed through the vtkOpenGLActor class, or is there a better way to achieve the same results? 

Below are some figures that describe what I want to achieve.
Thanks!
From Cowan (2014):
image
Dense point cloud: not enhanced, low values block higher values at depth:
image
Dense point cloud: enhanced (MIP), higher values are brought to the foreground:
image

reference:
http://www.orefind.com/docs/default-source/orefind-research-papers-in-pdf/cowan_2014_xray.pdf

banesullivan:
Adds shader customization support to the `Actor` class, implementing two features that have been requested for a while:

1. **Maximum Intensity Projection (MIP)** for point clouds, where vertices with higher scalar values render in front regardless of camera distance. This picks up the work from #6709 and the original request in #1212, with a key fix: scalar values are normalized to OpenGL clip space so points don't vanish when values fall outside `[-1, 1]`.

2. **Custom point sprite shapes** via fragment shaders, replacing VTK's default square points with circles, triangles, hexagons, diamonds, asterisks, or stars. This wraps the approach demonstrated in #7681 into a clean public API.

Both features target different shader stages (vertex and fragment), so they compose on the same actor without conflict.

The core addition is a general-purpose `add_shader_replacement` / `clear_shader_replacements` API on `Actor` that wraps VTK's `GetShaderProperty().AddShaderReplacement()` with a registry for conflict tracking. Each built-in feature (MIP, point sprites) registers under its own key, so clearing one doesn't nuke the other. The registry also prevents two features from silently stomping on the same shader tag.

On top of that:

- `enable_maximum_intensity_projection(clim=None)` / `disable_maximum_intensity_projection()` handle the vertex shader replacement with scalar normalization
- `set_point_sprite_shape(shape)` / `clear_point_sprite_shape()` handle the fragment shader replacement with six built-in shapes
- `point_shape` parameter on `add_mesh` and the global theme for convenient access without touching the actor directly
- `ShaderType` and `PointSpriteShape` Literal types keep the parameter space constrained

The low-level API is also public for advanced users who want to write their own GLSL replacements.

- Resolves #1212
- Supersedes #6709
- Implements the shader approach from #7681

Screenshot 2026-04-06 at 9 57 21 PM

Screenshot 2026-04-06 at 9 57 25 PM


banesullivan:
> **Test Image Report:** 

Already proving SO USEFUL!

PR #8468, opened 2026-04-07
Lint rule: disallow bare vtk imports

banesullivan:
## Summary

Add a ruff lint rule (`TID251`) to ban bare `import vtk` / `from vtk import ...` across the PyVista library, enforcing the use of targeted `vtkmodules` imports instead.

## Motivation

PyVista's core API is designed to work without pulling in VTK's full graphics stack. When code uses `import vtk` or `from vtk import ...`, it triggers the import of the **entire** VTK package (rendering, OpenGL, and GUI modules) even if only a single data-model class is needed causing:

- **Unintended GPU/graphics dependencies in headless environments.** importing `pyvista.core` for data processing on CI servers can hit `libGL` errors or segfaults because a stray `import vtk` pulled in the rendering backend.
- **Slower import times.** Importing the monolithic `vtk` package is significantly slower than importing individual `vtkmodules` subpackages.
- **Defeats the core/plotting separation.** PyVista deliberately splits VTK imports into `_vtk_core.py` (no GL) and `_vtk.py`/`_vtk_gl.py` (GL-dependent). A bare `import vtk` anywhere in core silently breaks that boundary.

PyVista already follows the `vtkmodules` pattern by convention, but without a lint rule it's easy for bare `vtk` imports to slip back in, especially via `TYPE_CHECKING` blocks or `except ImportError` fallbacks that look harmless but still resolve to the monolithic package.

I think some of these `from vtk imort <thing>` imports were added to deal with different import paths across VTK versions. We should instead have those import paths ironed out in our `_vtk_core.py` or `_vtk_gl.py` modules instead.

## Changes

This adds `[tool.ruff.lint.flake8-tidy-imports.banned-api]` rule banning `vtk` imports with a descriptive message. Added `TID251` to the `examples*` per-file-ignores since example scripts are user-facing and may legitimately demonstrate raw VTK interop.

I suspect this branch will error across versions of VTK so will use it as a place to iron this out

### DIFF
--- a/pyvista/plotting/lights.py
+++ b/pyvista/plotting/lights.py
@@ -7,16 +7,14 @@ from typing import TYPE_CHECKING
 
 import numpy as np
 
-# imports here rather than in _vtk to avoid circular imports
-from vtkmodules.vtkCommonMath import vtkMatrix4x4
-from vtkmodules.vtkRenderingCore import vtkLight
-from vtkmodules.vtkRenderingCore import vtkLightActor
-
 from pyvista._deprecate_positional_args import _deprecate_positional_args
 from pyvista.core import _validation
+from pyvista.core._vtk_core import vtkMatrix4x4
 from pyvista.core._vtk_utilities import DisableVtkSnakeCase
 from pyvista.core.utilities.arrays import vtkmatrix_from_array
 from pyvista.core.utilities.misc import _NoNewAttrMixin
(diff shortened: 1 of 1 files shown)


akaszynski (approved):
> I suspect this branch will error across versions of VTK so will use it as a place to iron this out

Turns out it only needed a handful of modifications. Good to go as-is. Thanks for catching the `from vtk import vtkWriter`.

PR #8471, opened 2026-04-08
Reduce noise in image regression warnings

banesullivan:
The new image report introduced in #8451 had too much noise and shows ~50+ warnings that it honestly shouldn't be. 

In a few places we already have the right cached image, but we still warned because the wrong baseline was checked on CI. In others, the rendered result was effectively stable and just barely over the warning threshold.

There's a bigger issue that needs to be addressed in https://github.com/pyvista/pytest-pyvista/issues/286 but this PR should at least clean up a few of the warnings.

This PR adjusts the `warning_value` overrides for the near-match single-cache cases that are visually stable but still drift a bit across CI. That covers `bounding_box[frame]`, `plot_above_below_color`, `show_bounds_no_labels`, `show_bounds_n_labels`, `create_axes_orientation_box`, `user_matrix_silhouette`, `hide_cells[cell]`, and `hide_cells_no_scalars`.

I left `plot_bounds_axes_with_no_data` and `multi_renderers` alone for now. Those still line up with alternate cached renders and look like real cross-render differences?

### DIFF
--- a/tests/plotting/test_plotting.py
+++ b/tests/plotting/test_plotting.py
@@ -2360,7 +2360,9 @@ def test_user_matrix_mesh(sphere):
         pl.add_mesh(sphere, user_matrix='invalid')
 
 
-def test_user_matrix_silhouette(airplane):
+def test_user_matrix_silhouette(airplane, verify_image_cache):
+    verify_image_cache.warning_value = 400
+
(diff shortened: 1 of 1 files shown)


banesullivan:
Modest improvement. Think this is fine to merge as is and will have to wait on https://github.com/pyvista/pytest-pyvista/issues/286 for any addition improvements

Issue #3180, opened 2022-08-18
Side Effects (Array Mutations) when Accessing Dataset Attributes


adam-grant-hendry:
### Describe the bug, what's wrong, and what you expected.

As revealed in [PR #3170](https://github.com/pyvista/pyvista/pull/3170), accessing certain dataset attributes in function scopes causes side effects: mutations of arrays. Specifically, we saw:

1. Accessing `volume.active_scalars` in `plotting.add_volume` 

```python
2918:            scalars = volume.active_scalars
```

led to `volume['Spatial Point Data']` from the `examples.load_uniform()` dataset getting mutated (it was rescaled to the `0-255` range)

![img_diff](https://user-images.githubusercontent.com/59346180/184715785-36b22abd-26fd-40b8-ac07-1d24e16d9322.png)

and attempting to loop through 

```python
for name in volume.array_names:
    ...
```

led `name` to maintain a reference to the last element in `volume.array_names`, causing the last item of `test_plot_helper_two_volumes` to not be plotted:

![Scope_Leak](https://user-images.githubusercontent.com/59346180/184936427-d6e918bf-772e-4421-bd57-c5470d144f8a.png)

### Steps to reproduce the bug.

Please see the problem description.

### System Information

```shell
--------------------------------------------------------------------------------
  Date: Wed Aug 17 17:09:20 2022 Pacific Daylight Time

                OS : Windows
            CPU(s) : 96
           Machine : AMD64
      Architecture : 64bit
               RAM : 190.7 GiB
       Environment : Python
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : Quadro RTX 8000/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 516.94

  Python 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64    
  bit (AMD64)]

           pyvista : 0.36.1
               vtk : 9.1.0
             numpy : 1.23.2
           imageio : 2.21.1
           appdirs : 1.4.4
            scooby : 0.5.12
        matplotlib : 3.5.3
         pyvistaqt : 0.9.0
           IPython : 8.4.0
              tqdm : 4.64.0
            meshio : 5.3.4
--------------------------------------------------------------------------------
```

### Screenshots

Please see the problem description.

adam-grant-hendry:
Adding this here so I don't forget: I wonder if we use `weakref` anywhere in the code base...?

Issue #1463, opened 2021-06-28
clip_scalar influenced by global active scalar state? Wrong Filter Result


flutefreak7:
**Describe the bug, what's wrong, and what you expect:**

Multiple successive clip_scalar operations are giving the wrong resulting geometry. It appears that the wrong scalar is being used in some operations.

-----

**To Reproduce**

```py
import pyvista as pv

sphere = pv.Sphere()
sphere['x'] = sphere.points[:, 0]
sphere['y'] = sphere.points[:, 1]
sphere['z'] = sphere.points[:, 2]
plotter = pv.Plotter()
plotter.add_mesh(sphere.clip_scalar(scalars='x', value=0), color='red', opacity=0.1)
plotter.add_mesh(sphere.clip_scalar(scalars='y', value=0.1), color='green', opacity=0.1)
plotter.add_mesh(sphere.clip_scalar(scalars='z', value=0.2), color='blue', opacity=0.1)
plotter.show_axes()
plotter.show()
```

**Screenshots**
![image](https://user-images.githubusercontent.com/2966015/123675704-a71d8600-d808-11eb-8fc6-61235bd01d79.png)

-----

**System Information:**
 Please run the following code wherever you are experiencing the bug and paste the output below. This report helps us track down bugs and it is critical to addressing your bug:

```py
# Get system info
import pyvista as pv
print(pv.Report())
```

```txt
# Paste system info here

--------------------------------------------------------------------------------
  Date: Mon Jun 28 12:02:19 2021 Central Daylight Time

                OS : Windows
            CPU(s) : 8
           Machine : AMD64
      Architecture : 64bit
               RAM : 31.9 GB
       Environment : Jupyter
        GPU Vendor : NVIDIA Corporation
      GPU Renderer : Quadro P3000/PCIe/SSE2
       GPU Version : 4.5.0 NVIDIA 432.06

  Python 3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 18:58:18) [MSC v.1900 64
  bit (AMD64)]

           pyvista : 0.29.0
               vtk : 9.0.1
             numpy : 1.19.4
           imageio : 2.9.0
           appdirs : 1.4.4
            scooby : 0.5.6
            meshio : 4.3.6
        matplotlib : 3.3.3
         pyvistaqt : 0.2.0
             PyQt5 : 5.15.2
           IPython : 7.19.0
          colorcet : 1.0.0
     ipyvtk_simple : 0.1.4
[...]

PR #8472, opened 2026-04-08
Fix shared mesh active scalars mutation in subplots

### LINKED ISSUE #542: Unexpected colormapping result using subplots

banesullivan:
## Summary

Resolves #542 

When the same mesh object is added to multiple subplots with different scalar arrays, VTK's single-active-scalars-per-dataset limitation causes each `add_mesh` call to overwrite the active scalars set by the previous one. The result: all subplots render the last scalar array added, and `clim` arguments are silently ignored for earlier subplots. This is especially bad if trying to plot the same mesh in multiple, separate plotters in a jupyter notebook and invalidates your previous, still active plots.

This has been an open issue since 2019. The workaround was to `.copy()` the mesh before adding it to each subplot, but that breaks a common interactive workflow where users modify mesh data and expect changes to propagate across all renderers.

`add_volume` is not affected by this bug. It already copies the input via `cell_data_to_point_data()` before touching scalars, so each volume mapper operates on its own dataset. The shared-mutation problem only arises in `add_mesh` where multiple `_DataSetMapper` instances can reference the same mesh object.

## Approach

Each `_DataSetMapper` now inserts an `ActiveScalarsAlgorithm` into its VTK pipeline. The algorithm produces a shallow copy of the input with the desired scalars set as active, so each mapper independently controls which array is active without mutating the original dataset. This is the same pattern ParaView uses internally and I originally outlined in https://github.com/pyvista/pyvista/issues/542#issuecomment-1407545025. I've actually been using this exact pattern in all of my work for the last few years stably.

Key changes:

- **`_DataSetMapper._configure_scalars_mode`** now routes through `_update_active_scalars_algo` instead of setting `dataset.active_scalars_name` directly
- **`_DataSetMapper.dataset`** property returns the original input mesh (not the algorithm output), so user code that inspects `mapper.dataset` still gets the object they passed in
- **Plotter**: removed the `elif` branch in `add_mesh` that explicitly mutated mesh active scalars when no algorithm was present

## Performance

The `ActiveScalarsAlgorithm` is now created for every `add_mesh(..., scalars=...)` call, not just the shared-mesh case. The cost is a `ShallowCopy` (pointer copy of data arrays, no memory allocation for array data) per pipeline update. Previously, `add_mesh` mutated the dataset directly, which was cheaper but incorrect for shared meshes. The overhead is negligible compared to the rendering cost.

## Usage

```python
import pyvista as pv
import numpy as np

mesh = pv.ParametricMobius()
mesh["z"] = mesh.points[:, 2]
mesh["x"] = mesh.points[:, 0]

pl = pv.Plotter(shape=(1, 2))
pl.subplot(0, 0)
pl.add_mesh(mesh, scalars="z")
pl.subplot(0, 1)
pl.add_mesh(mesh, scalars="x")
pl.link_views()
pl.show()

# The original mesh is never mutated
assert mesh.active_scalars_name == "elevation"
```
Screenshot 2026-04-08 at 12 22 40 PM

or in a jupyter notebook

Screenshot 2026-04-08 at 12 41 07 PM

### Documentation image updates

Several documentation examples were silently rendering incorrectly due to this bug. Most notably, the `compute_boundary_mesh_quality` docstring example plots the same mesh `cqual` across three subplots with different scalar arrays. Previously, all three subplots rendered the last scalar array set due to the active scalars mutation. With this fix, each subplot now correctly displays its own scalar field. Baselines have been regenerated to reflect the corrected output.

Other affected examples (`perlin_noise_2d`, `pump_bracket`) relied on the side effect of `add_mesh` mutating the mesh's active scalars. Their second plot call omitted the `scalars=` argument, expecting the prior `add_mesh` to have set it. These are not broken by this PR, they just render with whatever active scalars the mesh originally had, which may differ from what the mutation previously left behind.

---

cross linking #3293, #1463, #3180 for posterity as I think this is relevant to those as well

### DIFF
--- a/pyvista/plotting/mapper.py
+++ b/pyvista/plotting/mapper.py
@@ -107,17 +107,6 @@ class _BaseMapper(_NoNewAttrMixin, _BoundsSizeMixin, DisableVtkSnakeCase, _vtk.v
         new_mapper.ShallowCopy(self)
         if hasattr(self, 'dataset'):
             new_mapper.dataset = self.dataset
-        if getattr(self, '_input_dataset', None) is not None:
-            new_mapper._input_dataset = self._input_dataset
-        # Reproduce the active-scalars pipeline so the copy maps the same
-        # array as the original.  Pipeline reconnection can clear VTK-level
-        # properties set by ShallowCopy, so restore array_name afterwards.
-        algo = getattr(self, '_active_scalars_algo', None)
-        if algo is not None:
-            new_mapper._update_active_scalars_algo(algo.scalars_name, algo.preference)
-        array_name = self.GetArrayName()
-        if array_name:
-            new_mapper.array_name = array_name
         return new_mapper
 
     @property
@@ -450,8 +439,9 @@ class _DataSetMapper(_BaseMapper):
     def array_name(self, name: str) -> None:
         self.SetArrayName(name)
         if self._active_scalars_algo is not None:
+            # The algorithm's setter calls Modified() on change, so the
+            # next render re-runs the pipeline against the new array.
             self._active_scalars_algo.scalars_name = name
-            self._active_scalars_algo.Modified()
 
     @property
     def _mapped_scalars(self) -> pv.pyvista_ndarray | None:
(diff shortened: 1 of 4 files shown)


akaszynski:
This has been an issue for a while as well (usually addressed by copying via `mesh.copy()` for each call of `add_mesh`).

banesullivan:
I'm very eager to merge this but will give others time to review or provide drive by feedback

banesullivan:
Honestly, I'd like a minor release just for this feature/bugfix

user27182:
> Honestly, I'd like a minor release just for this feature/bugfix

We should address https://github.com/pyvista/pyvista/issues/8442 before minor release. Easiest would probably be to create new 0.48 release branch then pin `trame-vtk<2.10.3`. 

Effectively revert https://github.com/pyvista/pyvista/pull/8417 and https://github.com/pyvista/pyvista/pull/8431. I think we should be ready for release after that.

MatthewFlamm:
Really nice.  I have been waiting for this feature to reduce memory requirements for very large datasets.

pyvista/plotting/mapper.py
```diff
@@ -397,20 +411,60 @@ def __init__(
...
+        self._input_dataset: DataSet | None = None
         if dataset is not None:
             self.dataset = dataset
 
     @property
     def dataset(self) -> DataSet | None:  # numpydoc ignore=RT01
         """Return or set the dataset assigned to this mapper."""
+        if self._input_dataset is not None:
+            return self._input_dataset
         return cast('pv.DataSet | None', wrap(_mapper_get_data_set_input(self)))
```
MatthewFlamm, line 415: I'm assuming that since this method can return `None` that we could simplify this to `return None`?  But I'm not exactly sure given that `_mapper_get_data_set_input` does not use `_input_dataset` attribute.

https://github.com/pyvista/pyvista/blob/cb91b786daed9a20844c4d8d1e6cbf311396973d/pyvista/plotting/mapper.py#L1242-L1252

banesullivan: The fallback is needed for situations where you set an algorithm as the input. When you do `mapper.dataset = some_algorithm`, `_input_dataset` stays `None` (the setter only assigns it when `obj` is a `DataSet`), but the mapper still has a real dataset flowing through `GetDataSetInput()`. So `return None` would break that case.

Also, I realized there is a bug while relooking at this. The setter never clears `_input_dataset` when an algorithm is assigned, so the sequence:

```python
mapper.dataset = my_dataset
mapper.dataset = my_algorithm
```

leaves `_input_dataset` pointing at the stale `my_dataset`, and the property returns that instead of the algorithm's output. I'll fix the setter to reset `_input_dataset = None` in the algorithm branch, add a regression test, and a comment here

banesullivan: Handled in 52eef0f

MatthewFlamm: The comment is helpful.  Now I understand that `None` likely refers to an algorithm usage.  Much clearer intent now

banesullivan: Pushing one last change now and then I think this is ready to ship

PR #8482, opened 2026-04-09
Configure Codecov to wait for all CI jobs

banesullivan:
Configure codecove to wait for all Linux tests to complete before reporting the coverage status to prevent false negatives during the merge queue.

Ref https://github.com/pyvista/pyvista/pull/8420#issuecomment-4204165580

Ref https://github.com/pyvista/pyvista/pull/8405#issuecomment-4211668670

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -488,6 +488,12 @@ jobs:
     needs: [Linux, changes]
     if: always() && github.event_name == 'pull_request' && needs.changes.outputs.tests == 'true'
     steps:
+      # Needs to checkout only for codecov to match file names
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 2
+          persist-credentials: false
+
       - name: Download coverage reports
         uses: actions/download-artifact@v8
         with:
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -256,12 +256,13 @@ jobs:
           name: generated_test_images-${{ github.job }}-${{ join(matrix.* , '-') }}
           path: _generated_test_images
 
-      - name: Upload coverage to CodeCov
-        uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de
+      - name: Upload coverage reports
+        if: always()
+        uses: actions/upload-artifact@v7
         with:
-          fail_ci_if_error: true
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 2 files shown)


codecov.yml
```diff
@@ -13,10 +13,16 @@ coverage:
         if_not_found: success
         if_ci_failed: error
         if_no_uploads: error
+        # Wait for all 6 Linux CI jobs to upload coverage before reporting
+        # status. Without this, Codecov reports failure on partial data which
+        # causes the merge queue to dequeue the PR before the final status
+        # flips to success.
+        after_n_builds: 6
```
edabor: IMO, we could easily forget this parameter and when adding an item in the Linux matrix (ie. new python/vtk), forget to update it.

One option could be to transfer coverage reports as artifacts to a single dedicated codecov upload job and remove per-matrix item codecov upload.

edabor (dismissed):

codecov.yml
banesullivan: That would be more robust. we will definitely forget to update this if the CI jobs change

edabor: If needed, I can chime in this PR

akaszynski (approved):

codecov.yml
banesullivan: Thanks for taking this over!

.github/workflows/testing-and-deployment.yml
```diff
@@ -481,6 +482,31 @@ jobs:
           echo "### Test Image Report" >> $GITHUB_STEP_SUMMARY
           echo "$DEPLOY_URL" >> $GITHUB_STEP_SUMMARY
 
+  upload-coverage-:
+    name: Upload coverage reports
+    runs-on: ubuntu-22.04
+    needs: [Linux, changes]
```
user27182, line 488: Should also include VTK-dev in needs

banesullivan: Do we base coverage on that job?

banesullivan: I also don't want a failure with VTK dev to cause this to fail... may not matter with an `if: always()` condition though

user27182: Coverage is enabled and uploaded when the VTK-dev job runs. This is necessary, otherwise we get false negative coverage failures when new branches are added that are only hit with the dev wheels. 

And we have the same race conditions with the VTK-dev job as the regular unit tests. I _thought_ this would have prevented #8405 from merging (https://github.com/pyvista/pyvista/pull/8405#issuecomment-4225836033), but it somehow wasn’t an issue. Still think it should be uploaded together with the other reports. Should be able to make it work somehow even without the dev testing label applied.

Issue #8484, opened 2026-04-10
Deprecated setting `dtype` for latest NumPy dev


larsoner:
See [this](https://github.com/mne-tools/mne-python/actions/runs/24248318019/job/70800795772?pr=13833) for example
```
  File "/home/runner/work/mne-python/mne-python/mne/surface.py", line 813, in _call_pyvista
    out = meth(self.pdata, check_surface=False)
  File "/opt/hostedtoolcache/Python/3.14.3/x64/lib/python3.14/site-packages/pyvista/core/filters/data_set.py", line 3090, in select_interior_points
    out['selected_points'] = bools
    ~~~^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.14.3/x64/lib/python3.14/site-packages/pyvista/core/dataset.py", line 1531, in __setitem__
    self.point_data[name] = scalars
    ~~~~~~~~~~~~~~~^^^^^^
  File "/opt/hostedtoolcache/Python/3.14.3/x64/lib/python3.14/site-packages/pyvista/core/datasetattributes.py", line 261, in __setitem__
    self.set_array(value, name=key)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.14.3/x64/lib/python3.14/site-packages/pyvista/_deprecate_positional_args.py", line 243, in inner_f
    return f(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.14.3/x64/lib/python3.14/site-packages/pyvista/core/datasetattributes.py", line 552, in set_array
    vtk_arr = self._prepare_array(data=data, name=name, deep_copy=deep_copy)
  File "/opt/hostedtoolcache/Python/3.14.3/x64/lib/python3.14/site-packages/pyvista/core/datasetattributes.py", line 769, in _prepare_array
    data = data.view(np.uint8)
  File "/opt/hostedtoolcache/Python/3.14.3/x64/lib/python3.14/site-packages/pyvista/core/utilities/misc.py", line 369, in __setattr__
    object.__setattr__(self, key, value)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
DeprecationWarning: Setting the dtype on a NumPy array has been deprecated in NumPy 2.4.
Instead of changing the dtype on an array x, create a new array with x.view(new_dtype)
[...]

PR #8487, opened 2026-04-10
Add SMP tools helper

banesullivan:
## Summary

- Adds `pv.enable_smp_tools(backend='stdthread', n_threads=None)` as a public helper that switches VTK's runtime SMP backend and (optionally) sets the max thread count for any PyVista filter that delegates to `vtkSMPTools`.
- No behavior changes for existing code. The helper is opt-in; users who never call it continue to run VTK's default (sequential) backend.
- Validates `backend` against a fixed `Literal['stdthread', 'tbb', 'openmp', 'sequential']` set, validates `n_threads >= 1`, raises a clear `RuntimeError` on VTK builds that lack runtime backend selection, and restores prior state on a failed `SetBackend`.

## Why

VTK's Python wheels default to the sequential SMP backend, so PyVista filters that would otherwise be parallelized via `vtkSMPTools` run single-threaded out of the box. VTK exposes a runtime `vtkSMPTools.SetBackend()` API on recent builds, but most users don't know it exists or how to drive it from Python. This helper surfaces that capability with input validation, safe failure modes, and examples.

Typical usage:

```py
import pyvista as pv
pv.enable_smp_tools()  # STDThread, VTK's default thread count
# or
pv.enable_smp_tools(n_threads=8)
```

## Performance

Benchmarked with the exact filter calls shown below (1 warmup + 7 timed runs, median reported, alternating backend order per trial). Each row shows the best STDThread speedup observed across my Mac and one of my intel linux machines (will run on AMD later). The "Machine" column indicates which one produced the peak number.

| Dataset / call | Sequential (ms) | STDThread (ms) | Speedup | Machine |
| --- | ---: | ---: | ---: | --- |
| `knee_full.contour([40,80,120], method='flying_edges')` | 57.06 | 13.43 | **4.25x** | Mac |
| `fea_bracket.linear_copy().extract_surface(algorithm='geometry')` | 13.16 | 3.78 | **3.48x** | Linux |
| `head.contour([40,80,120], method='flying_edges')` | 3.19 | 1.51 | **2.12x** | Linux |
| `fea_bracket.cell_quality('scaled_jacobian')` | 4.07 | 1.97 | **2.06x** | Linux |
| `channels.contour_labels(smoothing=False)` | 109.44 | 71.42 | 1.53x | Mac (tied) |
| `bunny.compute_normals()` | 23.72 | 17.54 | 1.35x | Linux |
| `dragon.compute_normals()` | 368.26 | 277.71 | 1.33x | Linux |

### Benchmark environments

| Item | Mac | Linux |
| --- | --- | --- |
| OS | macOS 26.3 arm64 | Ubuntu 24.04 x86_64 |
| CPU threads | 10 | 20 |
| Python | 3.13.12 | 3.13.12 |
| VTK wheel | 9.6.0 | 9.6.1 |
| Helper default backend | `STDThread` | `STDThread` |

Neither wheel currently ships `TBB` or `OpenMP`, so those code paths exist in `enable_smp_tools` but weren't validated. A VTK build-from-source with `-DVTK_SMP_IMPLEMENTATION_TYPE=TBB` would be required to benchmark TBB directly.

### DIFF
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -105,10 +105,6 @@ from vtkmodules.vtkCommonCore import vtkTypeUInt32Array as vtkTypeUInt32Array
 from vtkmodules.vtkCommonCore import vtkUnsignedCharArray as vtkUnsignedCharArray
 from vtkmodules.vtkCommonCore import vtkVersion as vtkVersion
 from vtkmodules.vtkCommonCore import vtkWeakReference as vtkWeakReference
-
-with contextlib.suppress(ImportError):
-    from vtkmodules.vtkCommonCore import vtkSMPTools as vtkSMPTools
-
 from vtkmodules.vtkCommonDataModel import VTK_BEZIER_CURVE as VTK_BEZIER_CURVE
 from vtkmodules.vtkCommonDataModel import VTK_BEZIER_HEXAHEDRON as VTK_BEZIER_HEXAHEDRON
 from vtkmodules.vtkCommonDataModel import VTK_BEZIER_PYRAMID as VTK_BEZIER_PYRAMID
@@ -547,3 +543,6 @@ except ImportError:
     from vtkmodules.vtkCommonDataModel import (  # type:ignore[assignment]
         vtkCellTypes as vtkCellTypeUtilities,  # noqa: F401
     )
+
+with contextlib.suppress(ImportError):
+    from vtkmodules.vtkCommonCore import vtkSMPTools as vtkSMPTools
--- a/pyvista/core/_vtk_core.py
+++ b/pyvista/core/_vtk_core.py
@@ -105,6 +105,10 @@ from vtkmodules.vtkCommonCore import vtkTypeUInt32Array as vtkTypeUInt32Array
 from vtkmodules.vtkCommonCore import vtkUnsignedCharArray as vtkUnsignedCharArray
 from vtkmodules.vtkCommonCore import vtkVersion as vtkVersion
 from vtkmodules.vtkCommonCore import vtkWeakReference as vtkWeakReference
+
+with contextlib.suppress(ImportError):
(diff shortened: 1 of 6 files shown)


akaszynski:
I'm getting 8x speed up using [vtkFlyingEdges3D](https://vtk.org/doc/nightly/html/classvtkFlyingEdges3D.html) via [contour](https://docs.pyvista.org/api/core/_autosummary/pyvista.datasetfilters.contour) on an i9-14900 with 24 threads.

There's a good thread on Kitware's discourse for those wanting to read more about this:
https://discourse.vtk.org/t/build-python-wheels-with-smp-tools-for-runtime-backend-selection/16268/20

pyvista/core/_vtk_core.py
```diff
@@ -105,6 +105,10 @@
 from vtkmodules.vtkCommonCore import vtkUnsignedCharArray as vtkUnsignedCharArray
 from vtkmodules.vtkCommonCore import vtkVersion as vtkVersion
 from vtkmodules.vtkCommonCore import vtkWeakReference as vtkWeakReference
+
+with contextlib.suppress(ImportError):
```
user27182: Placing this here disrupts import sorting (imports above are now sorted independently of imports below). Recommend moving to end of file with all other suppressed imports

banesullivan: Fixed in 0129f70ea841342be2290fe21efa175b2314e1a1

akaszynski (approved):
Tracking latest `numpy` failure in https://github.com/pyvista/pyvista/issues/8484

Note: Encountered several unit test segfaults when testing with `pyvista-xdist` locally with a similar implementation using several theads with `STDThread`. We might want to specifically highlight which filters benefit from this in our docs rather than suggesting that it gets enabled always.

banesullivan:
In hindsight as I'm still experimenting with this, I'd really love to have a context manager for this. That way its less of an all or nothing

PR #8494, opened 2026-04-11
Use interior seam borders for subplots

banesullivan:
## Summary

Converts PyVista's multi-subplot borders from per-renderer boxes into shared interior seams, and moves their defaults onto the theme so they can be styled globally.

With `border=True` (implicit for any multi-subplot layout), every renderer drew a full rectangle in its own normalized viewport. Visually that produced:

- A heavy outer frame hugging the window edges, which is redundant — the window already has an edge.
- Doubled-up interior lines at each seam (both neighbors drew their own edge), making the separators look blockier than intended.
- A thick 2px black default that was loud on white/document-style backgrounds.

The goal is for multi-subplot layouts to look like a clean grid: thin, subtle separators *between* neighbors and nothing on the outside.

## Changes

**Interior-only borders**

- `Renderer.add_border()` now takes an optional `edges=` kwarg (any subset of `('top', 'left', 'bottom', 'right')`); default still draws all four, so direct callers are unaffected.
- New `Renderer._refresh_interior_border()` helper inspects the renderer's viewport after it has been positioned and rebuilds the border actor, keeping only edges whose viewport coordinate is strictly inside `(0, 1)`.
- `Renderers.__init__` calls the helper on each renderer after viewports are assigned, gated on `len(self._renderers) > 1`, so 1×1 plotters that opt in with `border=True` still get the full rectangle.

**Theme-backed defaults**

- `Theme` gains `border_color` (default `'grey'`) and `border_width` (default `1.0`) fields with the usual property/setter pair, slots entry, and `__repr__` line. Serialization rides the existing slot-driven `to_dict` / `from_dict`, so save/load round-trips cleanly.
- `BasePlotter`, `Plotter`, `Renderers`, `_plot`, and the `_cli.plot` entrypoint switch `border_color`/`border_width` kwargs to `None` sentinels and pull from the theme when unset. Users passing explicit values get the same behavior as before.
- Docstrings updated to reference the theme instead of hardcoded `'k'` / `2.0`.

## Visual before/after

Before: each subplot had a full rectangle, outer frame + doubled seams. After: a single thin grey seam between neighbors, nothing on the outside.

A 2×2 document-theme plotter now renders with just a clean `+` between the four cells. Grouped layouts (`shape=(3, 3)` + `groups=...`) trace their seams correctly around non-trivial neighbor boundaries.

## Backwards compatibility

- `add_border()` and `Plotter(..., border_color=..., border_width=...)` keep their existing signatures; only the defaults move.
- Themes without the new fields still work via slot-driven serialization. Users loading old saved themes will pick up the new defaults automatically.
- Single-plotter users who explicitly set `border=True` still see a full rectangle — the interior-only rewrite is gated on multi-subplot layouts.

77f24788-0793-4b09-ad52-25e4e94b5b14

742a7f08-fc03-4863-be1f-14c8362999cc

Going to pull the failed image regressions from CI as the new baselines


user27182:
Neat! Just seeing this now. I also noticed the hard-coded border color: it's themed in #8876. Mostly noticed this border issue while working on `plot_compare` from #8862.

I like the thinner and more subtle border profiles from this PR.

user27182:
@banesullivan I pushed changes to this PR to roll things along. It looked good overall and I really like the change. But one issue I had is that the implementation effectively was overloading the term "border", where `border=True` would still draw an actual outer border for a single plot, but `border=True` would now _only_ draw interior seams for subplots. Thus, the ability to actually draw a proper outer border around a multi-plot window was lost.

Instead of overloading `border` this way, I opted instead to make this more of a breaking change to redefine `border` to strictly mean _outer_ border of the entire window. This way, the feature isn't lost, and an outer border is always drawn with this option enabled. Instead, a new themed config option `subplot_seams` is added to control the _interior_ borders. But this is a breaking change since previously `border=False` would remove _all_ borders, interior end exterior. But now, `border=False` no longer removes the interior ones, so the seams still show up, and users need to set `subplot_seams=False` as well to remove everything.

banesullivan:
Excellent, thanks for getting this over the finish line @user27182!

user27182 (dismissed):

user27182 (approved):

PR #8495, opened 2026-04-12
Return a context manager with pv.enable_smp_tools()

banesullivan:
- `pv.enable_smp_tools()` now returns a context manager that restores the previous VTK SMP backend and thread count on exit, so `with pv.enable_smp_tools(n_threads=8):` cleanly scopes the change to a block.
- Plain calls keep their existing behavior: the backend is applied immediately and the returned value can be discarded when the change should apply for the remainder of the process.
- Restoration runs even when the `with` block raises, and nested `with` blocks unwind in LIFO order.
- Added three tests covering normal exit, exception propagation, and nested usage.
- Updated the docstring with a `with`-block example.

```py
mesh = ...
with pv.enable_smp_tools(n_threads=8):
    mesh.contour(method='flying_edges')
```

### DIFF
--- a/pyvista/core/utilities/misc.py
+++ b/pyvista/core/utilities/misc.py
@@ -221,15 +221,35 @@ def has_module(module_name: str) -> bool:
     return module_spec is not None
 
 
+class _SMPToolsContext:
+    """Context manager that restores VTK SMP backend state on exit."""
+
+    def __init__(self, original_backend: str, original_threads: int) -> None:
+        self._original_backend = original_backend
+        self._original_threads = original_threads
+
+    def __enter__(self) -> Self:
+        return self
+
+    def __exit__(self, *exc_info: object) -> None:
+        _vtk.vtkSMPTools.SetBackend(self._original_backend)
+        _vtk.vtkSMPTools.Initialize(self._original_threads)
+
+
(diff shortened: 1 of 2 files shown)


banesullivan:
For posterity, the question of if garbage collection would trigger the `__exit__` and thus disable SMP if `pv.enable_smp_tools()` is called but never assigned to a variable. In short, no this is not an issue because `__exit__` is not called on `__del__`.

Python's context manager protocol (`__enter__`/`__exit__`) is only invoked by the `with` statement. Garbage collection does not call `__exit__`, only `__del__` if one exists, and `_SMPToolsContext` here doesn't define one.

`pv.enable_smp_tools()` works like this:
1. It imperatively switches VTK's SMP backend (calls `vtkSMPTools.SetBackend` + `Initialize`). This side effect is permanent for the process.
2. It returns an `_SMPToolsContext` object that captures the prior state.
3. If you use `with pv.enable_smp_tools():`, Python calls `__enter__` on entry and `__exit__` on exit, and `__exit__` restores the prior state.
4. If you discard the return value, the object is eventually garbage-collected with no side effects and the backend change stays in place.

akaszynski (approved):

PR #8496, opened 2026-04-12
Add `CellType.dimension_map` class property

banesullivan:
Add `CellType.with_dimension(dim)` classmethod that returns a `frozenset` of all cell types with the given topological dimension (0, 1, 2, or 3). Groupings are derived dynamically from each member's existing `CellType.dimension`, which comes from `vtkCellTypeUtilities.GetDimension`, so the sets stay in sync with VTK as new cell types are added instead of requiring hand-maintained constants.

This helps me replaces hard-coded `CELLS_0D/1D/2D/3D` sets and enables one-liner checks such as `mesh.distinct_cell_types <= pv.CellType.with_dimension(2)` for "all 2D cells".


pyvista/core/celltype.py
```diff
@@ -1113,6 +1114,60 @@ def n_faces(self) -> int:  # numpydoc ignore=RT01
             raise ValueError(msg)
         return n_faces
 
+    @classmethod
```
user27182: Auto enum members need to be added to the template explicitly to be documented. May also need to add a methods heading. 
https://github.com/pyvista/pyvista/blob/main/doc/source/_templates/autosummary/enum.rst

pyvista/core/celltype.py
```diff
@@ -1113,6 +1114,60 @@ def n_faces(self) -> int:  # numpydoc ignore=RT01
...
+
+        Check whether a mesh contains only 2D cells by combining this with
+        :attr:`~pyvista.DataSet.distinct_cell_types`.
+
+        >>> mesh = pv.Sphere()
+        >>> mesh.distinct_cell_types <= pv.CellType.with_dimension(2)
+        True
+
+        """
+        if dimension not in (0, 1, 2, 3):
```
user27182: Use 

``` py
from pyvista import _validation
validation.check_contains([0, 1, 2, 3], dimension, name=‘dimension’)
```
(No branching required)

pyvista/core/celltype.py
```diff
@@ -1113,6 +1114,60 @@ def n_faces(self) -> int:  # numpydoc ignore=RT01
...
+            msg = f'`dimension` must be 0, 1, 2, or 3, got {dimension!r}.'
+            raise ValueError(msg)
+        return frozenset(member for member in cls if member.dimension == dimension)
```
user27182: Since these don’t change, consider adding private cached versions of these sets instead of iterating every time this is called

(Can use functools @cached_property)

user27182: You can also build these sets directly inside` __new__` and store as private properties

user27182:
@banesullivan do you mind if I push changes directly to resolve my suggestions? And what do you think about defining a `dimension_map` peroperty instead of `with_dimension` function?
``` py
@property
def dimension_map(self) -> dict[Literal[0, 1, 2, 3], frozenset[CellType]]:
```

banesullivan:
Yes go ahead!

banesullivan:
Let's go with your `dimension_map` implementation

user27182:
Sigh... `autoenum` has so many problems with rendering anything other than plain enum members. It's currently rendering `repr(CellType.dimension_map)` instead of its docstring.

image

pyvista/core/celltype.py
```diff
@@ -608,7 +612,145 @@ class _CellTypeTuple(NamedTuple):
 )
 
 
-class CellType(IntEnum):
+class _CellTypeMeta(EnumMeta):
+    """Metaclass to enable class property definitions for CellType."""
```
user27182, line 616: Alternatively, we could use `@_classproperty` from `core.misc.utilities`. But this creates a circular import cycle that would require moving this helper to a different module. Plus, that helper does not guard against mutation, whereas using a metaclass with `@property` gives that protection.

user27182 (commented):
This is effectively ready for review. Unfortunately, the docs currently do not render correctly, since the `autoenum` extension renders all non-instance attributes using `repr`:

https://github.com/domdfcoding/enum_tools/blob/05201f19f7028d8b270082b252aff91543cf28dc/enum_tools/autoenum.py#L524-L529

(this makes sense for rendering actual enum members, but does not for any other case)

Although the docs are important, I'm not sure if this should block this PR from landing. There are other issues with the `autoenum` extension as well, like the fact that we need a custom template file to document instance properties using autosummary, as well as the fact that this extension has not been updated to support the latest version of Sphinx.

I think we can maybe implement our own custom enum extension instead as a follow-up. This is covered by https://www.sphinx-doc.org/en/master/development/tutorials/autodoc_ext.html

(Bonus: this tutorial also covers the case for rendering hex values, which would be nice for improving the `CellStatus` enum docs too)

user27182 (approved):

banesullivan:
Thanks for getting this over the finish line, @user27182!

PR #8502, opened 2026-04-12
Streamline local developer workflow

banesullivan:
## Summary

Streamlines the local developer workflow by wrapping the most common tasks (install, lint, typecheck, test) as top-level `make` targets, and documents them as the recommended entry point in `CONTRIBUTING.rst`. 

I'm contributing to PyVista across multiple machines (mac, linux+AMD, linux+NVIDIA to test out different hardware for my work) and it means I'm repeatedly looking up or copy/pasting the same handful of commands. A single `make <target>` interface is easier to remember, easier to document, and gives us one place to evolve the developer workflow if the underlying tools change. It also lowers the barrier for new contributors, who can now run the full local check suite with a few quick commands instead of looking up the underlying tool to invoke for each task (`uv sync`, `pre-commit run`, `tox run -e mypy`, `uv run pytest tests/`), and the exact incantations were scattered across different sections of the contributing guide. This PR gives everyone a single, memorable set of commands while leaving the underlying tools fully accessible.

### `Makefile`

Adds four new targets, all thin wrappers around the project's existing tooling:

- **`make sync`**: installs dev dependencies via `uv sync --group dev` and installs `tox` (with `tox-uv`) as a `uv` tool.
- **`make lint`**: runs `uv run pre-commit run --all-files`. This is the same pre-commit gate documented in the existing "Style Checking" section and enforced in CI.
- **`make typecheck`**: runs `tox run -e mypy`, matching the mypy environment used in CI.
- **`make test`**: runs `uv run pytest -vv tests/`. Extra arguments can be forwarded to pytest via the `ARGS` variable, e.g. `make test ARGS="-n 10"` or `make test ARGS="-k filters -x"`. Parallelism is intentionally *not* hardcoded so developers can choose a worker count appropriate to their machine.

Also tweaks the existing `docstyle` target to drop the stray `./*.rst` argument, making the local invocation match what CI actually runs:

https://github.com/pyvista/pyvista/blob/5d35056009ee1faac2b82020d140e843f8d47aaf/.github/workflows/style-docstring.yml#L87-L91

### `CONTRIBUTING.rst`

Adds a new "Quick Development Commands" subsection at the top of the "Development Practices" section. This subsection lists the four `make` targets as the recommended entry point for day-to-day development, shows examples of forwarding arguments via `ARGS`, and links down to the existing "Unit Testing", "Style Checking", and "Building the Documentation" sections for contributors who need more control (e.g., running against a specific VTK or NumPy version, using tox factors, or building docs).

The existing detailed sections on pytest/tox/pre-commit are left untouched as they remain the authoritative reference for advanced usage.


banesullivan:
One thing that may need to be addressed here is that CI pins vale to `2.29.5` and running a a bare `uv tool install vale && make docstyle` fails:

```
Running vale
E100 [vocab] Runtime error

'config/vocabularies/pyvista' directory does not exist

Execution stopped with code 1.
make: *** [Makefile:28: docstyle] Error 2
```

We should probably patch this on our end for the latest vale?

banesullivan:
And also. We should probably add `uv.lock` to the `.gitignore`?

banesullivan:
@user27182 and @edabor, do you have thoughts on this as I think you two have mostly centralized/streamlined our tooling here and I don't want to add a conflict or additional way of doing things

akaszynski:
Really like the idea of a quickstart `make` for setting up a dev env, but I'd also like to hear other opinions.

banesullivan:
Also finding that I need to update the `doc/Makefile` as well to use `uv`

user27182:
Can you add any of the deps used by the make commands to the dev group in py pyproject.toml? E.g. tox, vale, uv, and make itself? Such that all commands work out of the box.


RE: Vale pin, I recall there was an issue with reviewdog when the diff had more than 300 file changes (it would fail).  I _think_ this may be related to the Vale pin.  Not sure if this is still relevant. Maybe a separate PR bumping this dep (and testing adding/removing 300+ files) is needed.

banesullivan:
Good call, I added `tox` and `tox-uv` to the `dev` group so `make sync` sets everything up in one shot, and updated `make typecheck` to use `uv run tox` so it picks up the project-installed tox.

The other three are trickier and I left them out:

- `vale` is a Go binary, not a Python package, so it can't live in a dependency group. I pinned the recommended install to `uv tool install vale@2.29.5` in the `docstyle` target's comment to match the CI pin
- `uv` and `make` are both a bit of a chicken/egg situation... you need it to have something installed on the system first to call the install command

Makefile
```diff
@@ -22,7 +22,26 @@ coverage-docs:
...
+	@uv run pre-commit run --all-files
+
+typecheck:
+	@echo "Running mypy"
+	@uv run tox run -e mypy
+
+# Extra pytest args can be passed via ARGS, e.g. `make test ARGS="-n 10 -k filters"`
+test:
+	@echo "Running tests"
+	@uv run pytest -vv tests/ $(ARGS)
```
edabor:
```suggestion
	@uv run pytest tests/ $(ARGS)
```
Maybe nitpick, but I prefer not enforcing verbose by default since it can show very long tracebacks.

Makefile
```diff
@@ -22,7 +22,26 @@ coverage-docs:
...
-# Install vale first with `pip install vale`
+# Vale is pinned to match CI (.github/workflows/style-docstring.yml).
+# Install with: `uv tool install vale@2.29.5`
+# Newer vale versions currently fail on the pyvista vocab config.
 docstyle:
 	@echo "Running vale"
-	@vale --config doc/.vale.ini doc pyvista examples ./*.rst --glob='!*{_build,AUTHORS.rst,_autosummary,source/examples}*'
+	@vale --config doc/.vale.ini doc pyvista examples
+
+sync:
```
edabor:
```suggestion
sync-deps:
```

Maybe more clear that dependencies are installed in this target ? `CONTRIBUTING.rst` should be modified accordingly.

Makefile
edabor, line 19: Maybe we could add a default target to allow shorter `make` command ?
Currently, it points to `coverage` target.

The `test` target is just a suggestion, we need to decide on the most appropriate one.

```suggestion
.DEFAULT_GOAL := test
```

CONTRIBUTING.rst
```diff
@@ -151,6 +151,36 @@ This section provides a guide to how we conduct development in the
 PyVista repository. Please follow the practices outlined here when
 contributing directly to this repository.
 
+Quick Development Commands
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For convenience, the most common developer tasks are wrapped as ``make``
+targets in the repository's top-level ``Makefile``. These are the
+recommended entry points for day-to-day development:
```
edabor: Maybe mention that `uv` must be installed on the machine in order to use most `make` targets.

edabor (changes requested):
Overall, I agree that we should lower the entry barrier to new contributors. This was the intent when switching CI to `tox`.

Just a few notes (that might be very subjective based on my own dev experience) regarding the `make` commands.

For non-POSIX users who are not experienced with `make`, you first need to install it. This could be a bit tedious, although modern package managers such as `scoop` or `chocolatey` are very helpful.

I think it is safe to assume that new contributors mostly use python and are more familiar with `virtualenv`, `uv` ... than `Makefile`. Although I agree that LLMS can quickly help you understand its syntax and how to use it.

Moreover, I think that CI should either use the `make` targets or test them to make sure they are up-to-date in order to reduce maintenance burden.

banesullivan:
@edabor, Thanks for the review, I think I've addressed all feedback

All test/doc/typecheck/doctest commands are now defined once in `tox.ini` and consumed by both CI and the Makefile. `make test-core` runs the exact same invocation as `tox run -e pyX.Y-core` in CI via the existing `core`/`plotting` factor logic, so there's a single source of truth for filters and flags.

**New Makefile targets matching CI:**
- `make test` / `test-core` / `test-plotting` (mirrors `py*-core` / `py*-plotting` jobs)
- `make doctest` (mirrors `tox -f doctest`)
- `make docs` / `docs-test` (mirrors `docs-build` / `docs-test` jobs)
- `make integration PROJECT=<name>` for trame/geovista/mne/pyvistaqt
- `make typecheck` routes through `tox -e mypy`

`doc/Makefile` also got `ci-build` and `ci-test` targets that delegate to the same tox envs

user27182:
Can we add `make` tabs to the docs as well? We currently have `pytest` and `tox` tabs... e.g. https://dev.pyvista.org/extras/developer_notes#unit-testing.

If the `make` commands are alternatives to using those, we should document this as a new tab.

edabor (changes requested):
Thanks a lot for addressing my review.

In order to make sure the Makefile is up-to-date with underlying commands syntax, maybe we should have a dedicated CI workflow that runs `make` commands in a dry-run fashion ?

For example, with ARGS="--collect-only" for pytest related jobs ?

.github/workflows/validate-makefile.yml
```diff
@@ -0,0 +1,92 @@
...
+  merge_group:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  id-token: none
+
```
edabor:
```suggestion
env:
  TOX_COLORED: "yes"
```
To help visualize CI

CONTRIBUTING.rst
```diff
@@ -1097,6 +1196,13 @@ To test all the images, run tests using either ``pytest`` or ``tox`` such that:
             tox run -e py3.11 -- tests/doc/tst_doc_build.py::test_static_images
             tox run -e docs-test -- -k test_static_images
 
+    .. tab-item:: make
```
edabor, line 1198: This section is not up-to-date. Fix is in https://github.com/pyvista/pyvista/pull/8555

banesullivan: Updated this branch, can you double check?

akaszynski (approved):

PR #8504, opened 2026-04-12
Add remove_nan_cells helper filter

banesullivan:
## Summary

Adds `DataSetFilters.remove_nan_cells`, a discoverable convenience filter for dropping cells whose scalar values are NaN. Works for both point and cell data via a single call: `mesh.remove_nan_cells(scalars='foo')`.

Previously, you had to call `point_data_to_cell_data().threshold()` pipeline with manual `cell_ids` tracking and `vtkOriginalPointIds` / `vtkOriginalCellIds` cleanup to handle NaN point data. This new filter replaces that recipe with a more robust mechanism.

## Why a new filter instead of extending `threshold`

The underlying machinery already exists as `threshold(scalars=..., preference='point', all_scalars=True)` with the default auto-range produces the same result, because:

- `get_data_range` uses `np.nanmin` / `np.nanmax`, so the bounds exclude NaN.
- With `all_scalars=True`, every point in a cell must satisfy the threshold. NaN comparisons return `False`, so any cell containing a NaN point is dropped. 

1. **Discoverability.** I've pointed people to a bare `threshold()` filter to remove NaNs for years and am always met with suprise. Users searching for "remove NaN" will not think to reach for `threshold()`.
2. **`threshold` already has a lot going on.** Adding an additional argument there that would change semantics (forcing `all_scalars=True`, overriding the `preference='cell'` default) would be a hidden footgun.
3. **Different defaults.** `threshold` defaults to `preference='cell'` and `all_scalars=False` because it's a cell-wise operation over cell data. `remove_nan_cells` defaults to `preference='point'` and `all_scalars=True` because the common NaN-cleanup case is point data.

### DIFF
--- a/examples/01-filter/using_filters.py
+++ b/examples/01-filter/using_filters.py
@@ -96,13 +96,14 @@ pl.show()
 # filtering pipeline through a chain; attaching each filter to the last filter.
 # In the following example, several filters are chained together:
 #
-# 1. First, and empty ``threshold`` filter to clean out any ``NaN`` values.
+# 1. First, use :func:`~pyvista.DataSetFilters.remove_nan_cells` to drop any
+#    cells whose scalar values are ``NaN``.
 # 2. Use an ``elevation`` filter to generate scalar values corresponding to height.
 # 3. Use the ``clip`` filter to cut the dataset in half.
 # 4. Create three slices along each axial plane using the ``slice_orthogonal`` filter.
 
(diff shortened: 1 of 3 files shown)


akaszynski (approved):
Didn't even realize you could use threshold for this.

banesullivan:
Well that emphasizes the need more than anything then!

PR #8511, opened 2026-04-13
chore: update pre-commit hooks, zizmor config, ignore uv.lock

mwtoews:
Update several pre-commit versions.

Add a `.github/zizmor.yml` [configuration file](https://docs.zizmor.sh/configuration/#rulesiddisable) <s>to ignore `template-injection` and `unpinned-uses` rules. The latter rule is new with recent zizmor versions, and would require changing e.g. `actions/checkout@v6` to use a hash instead of `v6` tags. I feel this is an overreach of security paranoia, and I would much prefer to see version tags. Also replace inline `template-injection` comments with a blanket ignore rule, if this is preferred as a default</s>.

Lastly, add uv.lock to .gitignore (I personally use uv for many projects locally, and I see there use tox-uv in use for CI).

### DIFF
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -236,11 +236,11 @@ jobs:
         run: pip install tox-uv
 
       - name: Core Testing (no GL)
-        run: tox run -e ${{env.TOX_FACTOR}}-cov-core # zizmor: ignore[template-injection]
+        run: tox run -e ${{env.TOX_FACTOR}}-cov-core
 
       - name: Plotting Testing (uses GL)
         if: ${{ !cancelled() }}
-        run: xvfb-run tox run -e ${{env.TOX_FACTOR}}-cov-plotting # zizmor: ignore[template-injection]
+        run: xvfb-run tox run -e ${{env.TOX_FACTOR}}-cov-plotting
 
       - name: Upload Images for Failed Tests
         if: always()
--- a/.github/workflows/testing-and-deployment.yml
+++ b/.github/workflows/testing-and-deployment.yml
@@ -236,11 +236,15 @@ jobs:
         run: pip install tox-uv
 
       - name: Core Testing (no GL)
-        run: tox run -e ${{env.TOX_FACTOR}}-cov-core
+        env:
+          TOX_ENV: ${{ env.TOX_FACTOR }}-cov-core
+        run: tox run -e "$TOX_ENV"
 
       - name: Plotting Testing (uses GL)
         if: ${{ !cancelled() }}
-        run: xvfb-run tox run -e ${{env.TOX_FACTOR}}-cov-plotting
+        env:
+          TOX_ENV: ${{ env.TOX_FACTOR }}-cov-plotting
+        run: xvfb-run tox run -e "$TOX_ENV"
 
       - name: Upload Images for Failed Tests
         if: always()
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,0 +1,5 @@
+rules:
+  template-injection:
+    disable: true
+  unpinned-uses:
+    disable: true
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,5 +1,6 @@
 rules:
-  template-injection:
-    disable: true
   unpinned-uses:
-    disable: true
+    config:
+      policies:
+        actions/*: ref-pin
+        "*": hash-pin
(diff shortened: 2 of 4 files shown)


banesullivan:
Sorry about the CI failures. We changed the deployments of some build artifacts and forgot to cleanly handle external PRs. We will need to merge in `main` here after https://github.com/pyvista/pyvista/pull/8512 lands

.gitignore
```diff
@@ -116,6 +116,9 @@ flycheck*
 # PyCharm
 .idea/
 
+# uv environments
+uv.lock
+
```
banesullivan, line 121: Thank you for adding this. I was wondering if we should in https://github.com/pyvista/pyvista/pull/8502#issuecomment-4232720263

.github/workflows/testing-and-deployment.yml
```diff
@@ -236,11 +236,11 @@ jobs:
         run: pip install tox-uv
 
       - name: Core Testing (no GL)
-        run: tox run -e ${{env.TOX_FACTOR}}-cov-core # zizmor: ignore[template-injection]
+        run: tox run -e ${{env.TOX_FACTOR}}-cov-core
 
       - name: Plotting Testing (uses GL)
         if: ${{ !cancelled() }}
-        run: xvfb-run tox run -e ${{env.TOX_FACTOR}}-cov-plotting # zizmor: ignore[template-injection]
+        run: xvfb-run tox run -e ${{env.TOX_FACTOR}}-cov-plotting
```
banesullivan: You're right that `${{env.TOX_FACTOR}}` is safe today since it's built from the matrix, so the current lint is noisy for this specific case. But disabling the rule repo-wide means we lose coverage against the actually-dangerous pattern: someone later adding a `run:` that interpolates `${{ github.event.* }}` or `${{ github.head_ref }}`.

Could we instead refactor those two steps to the env-var passthrough pattern? That makes zizmor happy without an ignore (I can't do a suggestion outside of the changed lines, so you may have to copy past this)

```yaml
      - name: Core Testing (no GL)
        env:
          TOX_ENV: ${{ env.TOX_FACTOR }}-cov-core
        run: tox run -e "$TOX_ENV"

        if: ${{ !cancelled() }}
        env:
          TOX_ENV: ${{ env.TOX_FACTOR }}-cov-plotting
        run: xvfb-run tox run -e "$TOX_ENV"
```

.github/zizmor.yml
```diff
@@ -0,0 +1,5 @@
+rules:
+  template-injection:
+    disable: true
```
banesullivan: Rejecting this (see comment above to solve the issue)

.github/zizmor.yml
```diff
@@ -0,0 +1,5 @@
...
+  unpinned-uses:
+    disable: true
```
banesullivan: I'm even more hesitant here. We SHA-pin every third-party action in the repo. Only the GitHub-maintained `actions/*` use tag pins. That is a deliberate policy on our side.

A blanket disable means the next time someone adds a third-party action with a tag pin, we won't catch it in review. I'd rather encode our actual policy in the config, something like:

```yaml
rules:
  unpinned-uses:
    config:
      policies:
        actions/*: ref-pin
        "*": hash-pin
```

That keeps the ergonomics you want for `actions/checkout@v6` while still enforcing SHA pinning on everything else... at least I think it does...

banesullivan (changes requested):
Thanks for the pre-commit bumps and the `uv.lock` gitignore! I want to push back on the `zizmor.yml` disables though (see comments inline) as we use self hosted runners and want to have a bit more aggressive/paranoid security posture.

.github/zizmor.yml
akaszynski: +1 for keeing it pinned. It's annoying... but better safe than sorry IMO.

.gitignore
mwtoews: Absolutely, it's a minor unrelated contribution in this PR; much more suited in that PR to streamline local developer workflows. I'll watch this one to coordinate where it lands.

.github/zizmor.yml
mwtoews: Thanks for the suggested configuration @banesullivan this passes unpinned-uses cases without needing to modify each GHA.

.github/workflows/testing-and-deployment.yml
mwtoews: Thanks again for the suggestion, works perfectly!

.gitignore
banesullivan: Let's go ahead and land it here

banesullivan (approved):

banesullivan:
Thanks for addressing the feedback!

PR #8514, opened 2026-04-14
Route VTK-dev coverage through central upload job

### LINKED ISSUE #8486: Include VTK-dev in Codecov upload step
> Should also include VTK-dev in needs 

 _Originally posted by @user27182 in [#8482](https://github.com/pyvista/pyvista/pull/8482/changes#r3066007606)_

banesullivan:
Address https://github.com/pyvista/pyvista/pull/8482#discussion_r3066007606 and close #8486 

- Replace the direct CodeCov upload in the `VTK-dev` workflow job with a `coverage_reports-*` artifact upload matching the `Linux` job, so VTK-dev coverage flows through the same pipeline.
- Add `VTK-dev` to the `upload-coverage-` job's `needs` so its coverage gets bundled into the single CodeCov upload when the `vtk-dev-testing` label is applied.
- Keep `if: always()` on `upload-coverage-` so the job still runs when `VTK-dev` is skipped (label not applied) and regular Linux coverage still uploads.


.github/workflows/testing-and-deployment.yml
banesullivan, line 429: @akaszynski upload happens here

user27182 (dismissed):
Thanks!

akaszynski (approved):

banesullivan:
@user27182, I'm pretty sure we didn't need to update the branch to benefit from #8526 as the merge queue automatically applies this to main

banesullivan:
So if we have situations like this in the future, just readded to the merge queue after the needed changes are on main is sufficient

banesullivan:
Ah the actual run on this PR was failing due to the flaky test. nevermind, the merge was likely warranted

user27182:
Indeed I had seen a similar comment made about generally not needing to merge with main. But yea I merged in this case because CI had previously failed from the flaky test.

Maybe we need something like a “Merge Queue 101” section for contributors to explain this.

banesullivan:
Yeah, lets let the dust settle on all these CI changes and see what works and doesn't work then write up details in the contributing guide

PR #8515, opened 2026-04-14
Fix bug with reference cycle

larsoner:
In MNE-Python [this CI run](https://app.circleci.com/pipelines/gh/mne-tools/mne-python/31198/workflows/992ba417-0441-4527-8f1b-b3d4651d18bf/jobs/81154) failed. It (1) runs each example, (2) does `gc.collect()` each time, and (3) makes sure there are no lingering `vtk*` objects around and errors if they do:
```
      File "/home/circleci/python_env/lib/python3.12/site-packages/joblib/parallel.py", line 773, in _return_or_raise
        raise self._result
    sphinx.errors.ExtensionError: 
    1 vtkPolyData @ mne/conf.py:Resetter.__call__:after:30_reading_fnirs_data.py:
    dict['_input_dataset'] with len=5 | 3 referrers: dict['r'] / list[1] / list[529835]
```
Based on `git bisect` I isolated it to:

https://github.com/pyvista/pyvista/pull/8472/changes#diff-c7905aa57849d51967437c53e63d699f9e2b68a878e311b113cbae3b6ddec7b3R425 

Changing these to `weakref.ref` fixes the issue for me.

Not sure why this wasn't caught by a test at the VTK end, I see you have `check_gc` mechanisms similar to those of MNE-Python :shrug:

### DIFF
--- a/pyvista/plotting/mapper.py
+++ b/pyvista/plotting/mapper.py
@@ -526,7 +526,7 @@ class _DataSetMapper(_BaseMapper):
             if input_conn is not None:
                 self._active_scalars_algo.SetInputConnection(0, input_conn)
             if self._input_dataset is not None:
-                set_algorithm_input(self, self._input_dataset)
+                set_algorithm_input(self._active_scalars_algo, self._input_dataset)
             self.SetInputConnection(0, self._active_scalars_algo.GetOutputPort())
         else:
             self._active_scalars_algo.scalars_name = name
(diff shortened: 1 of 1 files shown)


banesullivan (dismissed):
I'm also confused why our GC validations didn't pick up on this. Thanks for digging into it and cleaning this up

I'd like to try to get a test here in PyVista to capture the issue and prevent a regression, but let's land this fix first

banesullivan (approved):

PR #8516, opened 2026-04-14
Isolate netlify deployments to unblock forked PRs

banesullivan:
## The problem

Netlify deploys (docs preview, doc image report, test image report, release docs) were failing for PRs opened from forks. GitHub does not expose repository secrets to workflows triggered by `pull_request` from a fork, so the Netlify token was always empty and deploy steps either soft-failed or cascaded into downstream step failures and the recent addition of the commit status step in #8465 is what started triggering the failures. 

My previous attempt in #8512 to gate the deploy jobs behind GitHub Actions environments with required reviewers broke because GitHub's behavior for fork PRs is that **any job referencing a gated environment causes the entire workflow run to pause pending approval, not just the gated job**. I didn't catch this until the `doc` build, the `test` jobs, and the `preview` deploy would all sit in "Waiting" state together. Worse, the "Review deployments" button is only reachable by clicking into the workflow run detail page and not surfaced on the PR page, not attached to any status check, and not linked from anywhere a sane person would naturally look.

## The fix

Split the deployments out of the parent CI workflows into a single unified deploy workflow triggered by `workflow_run`:

- **`docs.yml`** (parent): builds the docs and runs the doc tests, uploads artifacts, and now also uploads a `pr-info` artifact containing PR number, head SHA, and fork status. No `environment:` references anywhere. Fork PR CI runs unblocked.
- **`testing-and-deployment.yml`** (parent): same pattern. Added a `pr-info` job; removed the `image-report` job. All existing tests run unblocked.
- **`netlify-deploy.yml`** (new): triggered on `workflow_run` completion of either parent workflow. Contains `setup`, `preview`, `doc-image-report`, `test-image-report`, and `release` jobs. Each job filters itself via `github.event.workflow_run.name` to match the correct parent. Carries the environment gating.

The `pr-info` artifact pattern is needed because `workflow_run` events don't populate `pull_requests` for fork PRs and the child workflow has no other way to discover the PR number or fork status of the build it's processing.

### Details

`workflow_run` is evaluated in the context of the base repository's default branch. GitHub always loads the child workflow file from `main` regardless of what's in the PR, so fork modifications to `netlify-deploy.yml` are ignored. The child runs with full secret access because it's effectively a base-repo workflow, not a fork workflow. This means:

- The parent CI (all the expensive builds, tests, doc generation) runs completely unblocked for fork PRs with approval only being gated to run the actions themselves which **is a button on the PR page.**
- The deploy workflow runs as an independent, secret-bearing run that can be gated or not, and can sit pending indefinitely without affecting any of the PR's required status checks. External contributors are unblocked this way and can ping @pyvista/maintainers for approval if they need to see the artifacts. 
- These deploys can be left unapproved altogether and not prevent status checks or merging PRs

Security is maintained because the child workflow **never checks out or executes any code from the fork**. It only consumes downloaded artifacts as data (the HTML from `docs-build` is uploaded to Netlify as static files, and the `*_failed_test_images` are passed to pyvista's own `tox run -e image-report` tool which runs from a fresh checkout of main). An explicit comment in the child workflow warns future maintainers against adding any checkout of the PR head. 

Deploy jobs also run exclusively on `ubuntu-22.04` GitHub-hosted runners, not self-hosted hardware.

Zizmor flags `workflow_run` as a dangerous trigger by default (it's the same broad heuristic that flags `pull_request_target`). The suppression is documented inline with the safety reasoning, following the same pattern we use for `label.yml`.

### Internal PRs

Everything runs automatically with zero friction. The parent workflows build and test; the deploy workflow's jobs run through the `internal-deploys` environment, which has no required reviewers. Commit statuses link to the Netlify URLs.

### External PRs (from forks)

The parent workflows run unblocked (pending approval to run CI from the PR page).

The deploy workflow runs as a separate workflow run that **pauses pending maintainer approval** for the (new) `fork-deploys` environment. Because all three fork-gated jobs (`preview`, `doc-image-report`, `test-image-report`) share a single environment, a maintainer only clicks approve once per parent workflow completion to unblock all deploys from that parent: **two clicks total for a PR to get all deployment previews**

If nobody approves, the deploy simply never happens and the PR is unaffected and can still be merged based on the parent workflow's required checks.

To approve: Actions → find the pending "Deploy" run → click "Review deployments" → approve the `fork-deploys` environment.

## Limitations and things to know

- **Fork PR deploy requests are not shown on the PR page.** If a maintainer doesn't proactively check the Actions tab, they won't see that a deploy is available for approval. The PR can still be reviewed and merged without the deploy, but if the image diff is genuinely useful for reviewing a visual regression, contributors can request maintainers to approve it.
- **2 approval clicks is the floor for fork PRs.** One per parent workflow completion (`Build Documentation`, `Unit Testing and Deployment`), since `workflow_run` fires per-triggering-workflow. Getting to 1 click would require merging the two parent workflows into one, which is not a good idea.
- **Child workflow does not appear as a PR status check by default.** The `preview`, `doc-image-report`, and `test-image-report` jobs post commit statuses with deploy URLs when they succeed, which do appear on the PR.
- **Fork modifications to `netlify-deploy.yml` are ignored.** GitHub loads the file from the default branch. This is the security property that makes `workflow_run` safe; any future maintainer editing this file should be aware that testing changes requires merging to main (or temporarily pointing the trigger at a test branch).
- **Deploy jobs run on GitHub-hosted runners only**, never on self-hosted hardware.


akaszynski:
> One per parent workflow completion (Build Documentation, Unit Testing and Deployment), since workflow_run fires per-triggering-workflow. Getting to 1 click would require merging the two parent workflows into one, which is not a good idea.

That's two clicks too many IMO, but if it's the least we can do for forks, that's fine. Guess we're going to be adding more maintainers :)

banesullivan:
While it is indeed annoying and too many clicks, most external contributions do not need these deployments and they can be requested on an as needed basis

akaszynski (approved):
Fork worked. LGTM.

MatthewFlamm:
This needs a sticky somewhere with a summary of most common workflow up top.  In the discussions?  

> To approve: Actions → find the pending "Deploy" run → click "Review deployments" → approve the fork-deploys environment.

This bit needs to be highlighted up top as it takes focused reading to find it.

banesullivan:
I'm going to see how this whole set up goes over the next few days and figure out what the best maintainer workflow is and see what I can do to streamline it or best document it

PR #8517, opened 2026-04-14
[Do not merge] Validate that forked PRs run

banesullivan:
This should validate CI runs from forked PRs  for https://github.com/pyvista/pyvista/pull/8516 (note base branch is that PR)

The deployments should skip because the they are not on main yet


banesullivan:
It worked!

PR #8535, opened 2026-04-15
Prevent numpy nightly job from failing entire workflow

banesullivan:
The numpy nightly job of the Linux matrix has been failing red on `main` and dragging the overall workflow status down with it. Same with the overall status of PRs.

This adds `continue-on-error: ${{ matrix.numpy-version == 'nightly' }}` to the `Linux` job. The nightly entry still runs and still reports its own status, but a failure no longer fails the parent check or the workflow. Every other matrix entry is unaffected because the expression evaluates to `false` for them.

This will allow for the overall status of the commit to report green while keeping the job as a failure in the full list


akaszynski (dismissed):

banesullivan:
This didn't work out how I hoped. Closing since #8544 is landing

PR #8536, opened 2026-04-15
Remove timing assertion from test_contour_labels_strict_external

### LINKED ISSUE #8529: test_contour_labels_strict_external flaky CI failure
```
FAILED tests/core/test_imagedata_filters.py::test_contour_labels_strict_external - assert 1.1244114842265844 < (1.4501817040145397 / 1.5)
```

Seen this fail on CI a few times in the merge queue

banesullivan:
Fixes #8529. `test_contour_labels_strict_external` was asserting a 1.5x wall-clock speedup between `contour_labels('external')` and `contour_labels('strict_external')` from single-sample `perf_counter` deltas. On shared CI runners that margin is narrower than the noise, so the test flaked in the merge queue. The reported failure was a 1.29x ratio, i.e. `strict_external` was faster, just not by enough to clear the threshold.

The timing assertion was proving a runtime property with the wrong signal. The thing we actually want to guarantee (that `'strict_external'` takes the short path and produces a correctly simplified output) is already covered by the remaining assertions plus the `select_inputs` / `select_outputs` `TypeError` checks below it.

This drops the timing block entirely, removes the now-unused `@flaky_test` decorator, and cleans up the `time` and `flaky_test` imports that were only used by that test.

### DIFF
--- a/tests/core/test_imagedata_filters.py
+++ b/tests/core/test_imagedata_filters.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import operator
 import re
-import time
 from typing import get_args
 
 import numpy as np
@@ -16,7 +15,6 @@ from pyvista.core.errors import DeprecationError
 from pyvista.core.errors import PyVistaDeprecationWarning
 from pyvista.core.filters.image_data import _InterpolationOptions
 from tests.conftest import NUMPY_VERSION_INFO
-from tests.conftest import flaky_test
 
 BOUNDARY_LABELS = 'boundary_labels'
 MORPHOLOGICAL_MAX_VAL = 42.0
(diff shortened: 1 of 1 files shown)


user27182:
> The timing assertion was proving a runtime property with the wrong signal. The thing we actually want to guarantee (that `'strict_external'` takes the short path and produces a correctly simplified output) is already covered by the remaining assertions plus the `select_inputs` / `select_outputs` `TypeError` checks below it.

I disagree here, we do care about the performance of the short path. Taking the short path is only a proxy for the performance, the thing we actually care about _is_ the performance delta, i.e. `strict_external` should definitely be more performant than `external`, whatever internal branching or code paths are taken.

But, the issue is this is indeed a performance check, and current CI is ill-suited to these kinds of checks. This was less of an issue before merge queue because it was easy to retry a single job, but now the entire CI suite must be re-run to address failures.

I'm fine with removing this, but we should open an issue to track that this important performance check is missing and should be included as part of a hypothetical future benchmarking/profiling CI job.

user27182 (approved):

akaszynski (approved):

banesullivan:
I agree we need some mechanism to handle performance tracking for these cases. I guess I missed the importance of the performance delta here. I'll ponder how we could validate this for the issue you raised in https://github.com/pyvista/pyvista/issues/8537

PR #8538, opened 2026-04-16
Remove future annotations from examples gallery

banesullivan:
Our example gallery scripts all start with `from __future__ import annotations`, which sphinx-gallery renders into every example page on the docs site. It's noisy, confusing, and an eyesore. The import does nothing useful here since PyVista requires Python 3.10+, and the examples have no type annotations. 

This PR strips the line from all example scripts and adds a lint config to the `examples/` directory so it can't creep back in.

- Dropped `TID251` from the `'examples*'` entry in `[tool.ruff.lint.per-file-ignores]` so the `banned-api` rule actually applies to the gallery.
- Added two overrides to `examples/ruff.toml`: `required-imports = []` to stop isort from auto-inserting the future import, and a `flake8-tidy-imports.banned-api` entry that bans `__future__.annotations` with an explanatory message.
- Removed `from __future__ import annotations` (and its trailing blank line) from all example files.

PyVista's minimum supported Python is 3.10, where `list[int]`, `X | None`, and the rest of the typing syntax already work at runtime without the future import. Mypy and pyright don't need it either. The examples directory has zero `TYPE_CHECKING` blocks and only one file with any function annotations at all (`99-advanced/anatomical_groups.py`, which uses `list[str]` (works natively on 3.10+)

This change is scoped to `examples/` only though as we do need `from __future__ import annotations` and because of the `if TYPE_CHECKING:` blocks defering imports like `from pyvista import Color`, and without lazy annotations those references would raise `NameError` at runtime.


examples/02-plot/moving_cmap.py
user27182, line 1: Strange that this diff is so huge even though only `from __future__ import annotations` is removed

user27182 (dismissed):
Thanks, these were indeed ugly and not necessary.

examples/02-plot/moving_cmap.py
banesullivan: Yeah this is strange... is it a github rendering issue? I don't think my ide did anythign to this file?

examples/ruff.toml
```diff
@@ -3,3 +3,11 @@ extend = "../pyproject.toml"
...
 # But use a different line length.
 line-length = 90
+
+[lint.isort]
+# Examples don't need `from __future__ import annotations` (Python >=3.10),
+# and it clutters the rendered sphinx-gallery output.
+required-imports = []
+
+[lint.flake8-tidy-imports.banned-api]
+"__future__.annotations" = { msg = "Not needed on Python >=3.10; omitted from examples for cleaner sphinx-gallery output." }
```
user27182: Nit: this isn't really a Python 3.10 thing. It's for resolving forward references, e.g. it's needed here since `Foo` is used in the type hints before it's defined. Without `from __future__`, we get `NameError: name 'Foo' is not defined`

``` py

foo: Foo
class Foo:
    ...
```

banesullivan: All I know about type annotations in Python is that I don't understand anything about type annotations in Python 🫠

banesullivan: Been cargo-culting `from __future__ import annotations` around in my own code bases to be honest without even bothering to ask why until today

user27182 (approved):

PR #8539, opened 2026-04-16
Forbid using `pltr` as `Plotter` variable name

user27182:
### Overview

Discovered while reviewing #8538


banesullivan:
Great minds think alike! I didn't think of `pltr` but I did a grep for `p = pv.Plotter` and didn't see any so I didn't add a linting rule around that

banesullivan (approved):

banesullivan:
Ergh. That weird diff in my PR may conflict with this.

user27182:
> Great minds think alike! I didn't think of `pltr` but I did a grep for `p = pv.Plotter` and didn't see any so I didn't add a linting rule around that

I also had not considered `pltr`! The `p` case was already handled in our tests, but I didn't know we could use a linter for this... might be able to completely remove the greps in `test_naming_conventions.py` then.

banesullivan:
👀 `test_naming_conventions.py`... yes, we should clean that out and rely on either a lint rule or a custom precommit hook.

Here's an example I added to pyvista-zstd to ban/reserve a name/alias for a project I'm working on that will open source later: https://github.com/pyvista/pyvista-zstd/pull/19/changes

Issue #8256, opened 2026-01-22
Possible memory bug with cutting operations performance


PathosV:
### Describe the bug, what's wrong, and what you expected.

I had made a [discussion topic](https://github.com/pyvista/pyvista/discussions/7598) on the behavior I observed with slice a while back but from my new findings I think this deserves to be upgraded to an actual bug. 

To recap, I noticed certain cutting operations were being slowed down in ways that did not make sense. In the discussion topic I noted that clip_scalar speed slowed with number of total cells in the mesh, not the number of cells being cut.

My new tests are on vtk9.6.0rc2 and have no polyhedral cells, I saw similar behavior on the previous vtk version. Paraview also does not appear to have the issues I'm seeing.

### Steps to reproduce the bug.

The new test I've made focuses on base slice method on a mesh that has ~80,000,000 cells.

```python
clip1 = grid.clip(
    normal=[0.571135, -0.804376, -0.163659], origin=[1.8256976403692484, -3.1154205039654324, 28.45077343224965]
)
# time to finish: 3.9s
clip2 = grid.slice(
    normal=[0.571135, -0.804376, -0.163659], origin=[1.8256976403692484, -3.1154205039654324, 28.45077343224965]
)
# time to finish 4m 1s
```
Thankfully, unlike the behavior for clip_scalar, base clip is still very performative. Slice, while making the same cuts is about 60x slower.

I went searching for method's of speeding up slice, starting out with a crinkle cut. We don't have an option for slices to give crinkle cuts so I made my own using two threshold calls:

```python

def crinkleSlice(
    grid: pv.UnstructuredGrid,
    normal: VectorLike[float],
    origin: VectorLike[float],
    progress_bar: bool = False,
) -> pv.UnstructuredGrid:
    normal = np.array(normal)
    origin = np.array(origin)
    normal_normalized = normal / np.linalg.norm(normal)
    grid["plane_distance"] = np.dot(grid.points - origin, normal_normalized)
    crinkle = grid.threshold(0, scalars="plane_distance", progress_bar=progress_bar).threshold(
        0, scalars="plane_distance", progress_bar=progress_bar, continuous=True
    )
    crinkle.point_data.remove("plane_distance")
    return crinkle
```
Using this is decently fast:
```python
crinkleclip = pylon.crinkleSlice(
    grid, normal=[0.571135, -0.804376, -0.163659], origin=[1.8256976403692484, -3.1154205039654324, 28.45077343224965]
)
[...]

PR #8554, opened 2026-04-17
Array and filter optimizations

banesullivan:
A few internal code path optimizations that I uncovered while benchmarking one of my projects.

### Speed up string array conversions

Two Python loops in `arrays.py` flagged with `OPTIMIZE` markers were burning per-element Python/C crossings.

- `convert_string_array(numpy -> vtk)`: pre-allocate via `SetNumberOfValues(n)` and iterate `arr.tolist()` to skip `numpy.str_` boxing. ~1.4x.
- `_SerializedDictArray._string` setter: store the JSON string as one value instead of one per character. The getter still joins so legacy char-per-value files keep loading. ~85x on a 50-key dict.

### Speed up has array check

`'name' in mesh.point_data` materialized the full key list before the membership test. Replaced with the VTK object's `HasArray`, with a fallback to `keys()` for an edge case for empty-names still resolves.

~76x on `'a5' in point_data` (20 arrays). This adds up when we look at how commonly `if 'foo' in mesh.point_data` occurs.

### DOC: Add performance note about slice `generate_triangles`

See https://github.com/pyvista/pyvista/issues/8256#issuecomment-4265932183

The slice filters defaults to `generate_triangles=False`, which selects `vtkCutter`'s slow per-cell path. `vtkCutter` defaults to `True`:

```text
slice(generate_triangles=False)   118 ms ± 2.3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
slice(generate_triangles=True)    31.9 ms ± 264 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### Code quality

These aren't really performance gains but just code quality things I noticed while benchmarking mesh creation

### `numpy_to_idarr` ravel

`numpy_to_idarr` always called `ind.ravel()`, even for 1D contiguous input where it only allocates view metadata. Skip it in the 1D case as this is widely used in `vtkCellArray` / `PolyData` / `UnstructuredGrid` construction. 

### Reduce redundant wrappers in `linear_copy`

The quad fix-up loop read `lgrid.cell_connectivity` 4-7 times per branch. Each read allocates a fresh numpy view so changed this to cache it to a local variable. Probably has perf gains for smaller meshes

### DIFF
--- a/pyvista/core/utilities/arrays.py
+++ b/pyvista/core/utilities/arrays.py
@@ -728,15 +728,21 @@ def convert_string_array(
             # setting the object name
             _set_string_scalar_object_name(vtkarr)
 
-        # OPTIMIZE ###########
-        for val in arr:
-            vtkarr.InsertNextValue(val)
-        ################################
+        # Pre-allocate the underlying storage instead of growing it via
+        # ``InsertNextValue`` per element. Iterating over ``arr.tolist()``
+        # avoids the per-element ``numpy.str_`` scalar boxing cost that
+        # iterating a numpy string array pays.
+        flat_list = arr.reshape(-1).tolist()
+        vtkarr.SetNumberOfValues(len(flat_list))
+        for i, val in enumerate(flat_list):
+            vtkarr.SetValue(i, val)
         if isinstance(name, str):
             vtkarr.SetName(name)
         return vtkarr
-    # Otherwise it is a vtk array and needs to be converted back to numpy
-    # OPTIMIZE ###############
+    # Otherwise it is a vtk array and needs to be converted back to numpy.
+    # ``np.array(list, dtype='|U')`` auto-sizes the unicode width to the
+    # longest value; passing dtype='|U' to np.empty defaults to width 1
+    # which truncates strings.
     nvalues = arr.GetNumberOfValues()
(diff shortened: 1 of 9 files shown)


pyvista/core/utilities/arrays.py
```diff
@@ -973,14 +978,24 @@ class _SerializedDictArray(DisableVtkSnakeCase, UserDict, _vtk.vtkStringArray):
...
-        self.SetNumberOfValues(0)  # Clear string
-        for char in str_:  # Populate string
-            self.InsertNextValue(char)
+        # Store the entire string as a single value rather than one
+        # character per value. This avoids O(len(str_)) Python<->C
+        # crossings and matches how every other VTK string field stores
+        # text. Reading still works for the legacy char-per-value format
+        # because the getter joins all values.
+        self.SetNumberOfValues(1)
+        self.SetValue(0, str_)
```
user27182, line 998: Good catch. Sounds like the original implementation is the result of my naive understanding of how VTK string arrays work.

The docs have a bunch of warnings about `user_dict` being inefficient. Do you think these warnings can be removed?

banesullivan: I changed the docs around them

user27182 (approved):

PR #8556, opened 2026-04-17
Move contributing guide to top level

edabor:
### Overview

IMO, the `Contributing` guide should be at the top level in the documentation, as done in many scientific projects (`matplotlib`, `scipy`).

Moreover, I think we should reorganize some sections to prevent high number of levels as follows
image


user27182:
It's good to make this more visible, but I think the mallet icon is already intended to be the top-level link to the contributing guide:

image

Though one issue with that is it seems links to raw .rst file, rather than the rendered docs.

akaszynski:
Mallet icon isn't obvious (enough) for me, though maybe I'm text sensitive. Agree with making this more prioritized.

user27182:
New Contributing section is there and works:

image

But the mallet icon link is broken.

I think keeping both is perhaps a bit redundant? I agree that the mallet icon is kind of non-obvious for what it should point to. Maybe we should remove the icon, and just keep the new Contributing section?

WDYT @edabor @akaszynski

akaszynski:
Remove the mallet, keep the new contributing section.

user27182 (dismissed):
Looks good, contributing is top level and no more gavel.
https://6a77baa38c5929a078bea0c9--pyvista-dev.netlify.app/contributing

user27182 (approved):

PR #8557, opened 2026-04-17
Move filtered warnings from tests to `pyvista/__init__.py`

user27182:
### Overview

We have a number of deprecation warning filters for our unit tests. Many of these are needed due to the warning being emitted by a third party lib which we have no control over.

This filter is great for testing, but it also means that users will still see the deprecation warning in actual use. E.g. 5 NumPy deprecation warnings are emitted when creating and saving a sphere, see https://github.com/pyvista/pyvista/issues/8484#issuecomment-4264419224
We should therefore also silence the warnings for users.

Due to the way that pytest filters warnings, I found it necessary to filter using an auto-use fixture. Filtering during `__init__` alone is not sufficient, nor is filtering at pytest session start. Using `scope='session'` to only filter once during testing is also insufficient.


user27182:
Coverage is failing because the changes require VTK dev testing to succeed. But these tests are failing due to upstream issues
https://gitlab.kitware.com/vtk/vtk/-/work_items/20034
https://gitlab.kitware.com/vtk/vtk/-/work_items/20018

But this PR is otherwise ready. Admin will likely need to temp disable coverage as a required check to get this merged.

banesullivan:
Coverage is disabled as a required check pending codecov proves to be stable for more than a week

pyvista/__init__.py
```diff
@@ -40,8 +44,57 @@
...
+                category=DeprecationWarning,
+            )
+        if vtk_version_info >= (9, 6, 99):  # >= (9, 7, 0)  # pragma: no branch
+            warnings.filterwarnings(  # Filter added in https://github.com/pyvista/pyvista/pull/8405
+                'ignore',
+                message=r'Call to deprecated method GetSize.*',
+                category=DeprecationWarning,
+            )
+
+    numpy_version_info = tuple(map(int, np.__version__.split('.')[:3]))
```
banesullivan, line 74: I think this will error on any pre-release or annotated release

```py
>>> tuple(map(int, '2.5.0rc1'.split('.')[:3]))
    tuple(map(int, '2.5.0rc1'.split('.')[:3]))
    ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: '0rc1'
>>> 
```

pyvista/__init__.py
```diff
@@ -40,8 +44,57 @@
...
+def _filter_warnings():
+    """Filter warnings emitted by 3rd party modules which PyVista has no control over.
+
+    These are filtered both for user-facing code, and for PyVista unit tests.
+    """
+    warnings.filterwarnings(  # Filter added in https://github.com/pyvista/pyvista/pull/7009
+        'ignore',
+        message=r'It is recommended to use web\.AppKey instances for keys.*',
+        category=Warning,
+    )
```
banesullivan, line 57: I think we need to use the `module=` scope with filter warnings otherwise this is process-global and any of these warnings coming from calls outside pyvista get suppressed. Same for other `filterwarnings` call sites. Otherwise the only way to recover from this is a hard `warnings.resetwarnings() ` which effects all `filterwarnings` everywhere in the environment

pyvista/__init__.py
```diff
@@ -40,8 +44,57 @@
...
+
+    if importlib.util.find_spec('trame_vtk'):  # pragma: no branch
+        # Filter vtk deprecations that were fixed in PyVista, but not in trame-vtk
+        if vtk_version_info >= (9, 5, 0):
+            warnings.filterwarnings(  # Filter added in https://github.com/pyvista/pyvista/pull/8003
+                'ignore',
+                message=r'Call to deprecated method GetData.*',
+                category=DeprecationWarning,
```
banesullivan, line 65: I'm pretty sure that Python hides `DeprecationWarning` by default and these only appears in a pytest run. We should avoid calling this in our package init and move it to our conftest.

pyvista/__init__.py
```diff
@@ -40,8 +44,57 @@
...
+        # These warnings are triggered by pyvista_ndarray.view() and VTK's C++ code
+        # See https://github.com/pyvista/pyvista/issues/8484
+        warnings.filterwarnings(
+            'ignore',
+            message=r'Setting the dtype on a NumPy array has been deprecated.*',
+            category=DeprecationWarning,
+        )
+
+
+_filter_warnings()
```
banesullivan, line 97: I'm not huge on this being forced on all downstream consumers on import.

banesullivan (changes requested):
I have a few concerns here, see comments

pyvista/__init__.py
user27182: I think I tried this at one point but couldn’t get it to work. Can try again.

pyvista/__init__.py
user27182: I’d have to try a fresh install of my IDE but these are surfaced in my local version of PyCharm, though it could be my config. I see the NumPy dtype deprecation warnings in IPython as well as the trame GetData deprecation warnings in a Jupyter notebook.

pyvista/__init__.py
user27182: This is kind of the whole point. Users can’t control the deprecation warnings emitted by trame-vtk, nor the NumPy dtype warnings emitted by PyVista calls to `pyvista_ndarray.view()`. If this kind of filtering isn’t desirable then we should close this PR, let the deprecation warnings (maybe?) surface for users, and let them decided if they want to do something about it. Though, they’ll likely initially think it’s something they should act upon in terms of code changes, only to realize it’s just noise emitted by PyVista that they need to ignore.

pyvista/__init__.py
banesullivan: Works fine on my machine! 😄 I don't ever see any of these warnings... maybe my environment config is overly "YOLO vibez, warnings don't matter!" though

I haven't been using trame in a while so these very likely are surfacing for users and I'm just not a trame user

pyvista/__init__.py
banesullivan: The numpy dtype warnings I get. But the trame-vtk warnings should be handled in trame-vtk, right?

banesullivan:
The NumPy side of this makes sense to me. The `dtype`/`shape` deprecations fire from our own code path (`pyvista_ndarray.view()` in `datasetattributes.py`), every user will hit them once NumPy 2.5 ships, and users have no way to silence them on their end. I agree we should handle at import-time.

I'm not thrilled on the trame_vtk and aiohttp filters in here though. Those warnings only trigger when trame_vtk is installed and a trame backend is active, so they're already scoped to a narrow slice of users. Silencing them at `import pyvista` mutates `warnings.filters` globally for every process that imports pyvista, including environments that never touch trame. This could/shoulde be mitigated with a `module=` scope, so a that if a downstream library calls `vtkCellArray.GetData()` directly in their own code, they don't loose that deprecation notice (which some of my code does)

I'd rather see issues (or PRs) filed against `Kitware/trame-vtk` for each of these deprecations, but if we need to mask them in the meantime, do it with `warnings.catch_warnings()` scoped to the trame submodules or call sites so only users who actually invoke trame are affected.

I'm not on board with trying to provide support or guaruntees on compatibility for multiple versions of trame or maintaining compatibility between trame and different versions of VTK. This is too complex and too out of our control. Our usage of trame should only be supported for the latest version of VTK that trame supports and issues with that compatibility should not be handled in pyvista directly.

Honestly, I'm looking at all of this and thinking we should migrate trame out of here and create a new `trame-pyvista` package for all of this. It can use our new jupyter notebook backend registration mechanism and have its own support/release cycle for these kinds of issues.

banesullivan:
My two cents: scope this down to just numpy and try to handle the rest upstream in trame. I'll volunteer to do so

pyvista/__init__.py
user27182: Confirmed using bare `Terminal` on a macbook, I see the NumPy deprecation warning:

``` shell
% python3 -m pip list
Package Version
------- -------
pip     25.2
% python3 -m pip install --pre --upgrade numpy \
  -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
Looking in indexes: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
Collecting numpy
  Downloading https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/numpy/2.5.0.dev0/numpy-2.5.0.dev0-cp314-cp314-macosx_14_0_arm64.whl (5.3 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 5.3/5.3 MB 659.6 kB/s  0:00:08
Installing collected packages: numpy
Successfully installed numpy-2.5.0.dev0
% python3 -i
Python 3.14.0 (v3.14.0:ebf955df7a8, Oct  7 2025, 08:20:14) [Clang 16.0.0 (clang-1600.0.26.6)] on darwin
>>> import numpy as np
>>> array = np.array([1])
>>> array.dtype = int
<python-input-2>:1: DeprecationWarning: Setting the dtype on a NumPy array has been deprecated in NumPy 2.5.
Instead of changing the dtype on an array x, create a new array with x.view(new_dtype)
```

tests/plotting/jupyter/test_trame.py
```diff
@@ -39,20 +39,6 @@
     pytest.mark.skipif(not has_trame, reason='Requires trame'),
     pytest.mark.skip_plotting,
     pytest.mark.skip_check_gc,
-    pytest.mark.filterwarnings(
-        r'ignore:It is recommended to use web\.AppKey instances for '
-        r'keys:aiohttp.web_exceptions.NotAppKeyWarning'
-    ),
```
banesullivan, line 45: I think this was fixed long ago in https://github.com/Kitware/wslink/commit/3a313f9418498a6f9de66f45ecb7ca691ace45e5 and we can just remove it?

banesullivan:
I opened https://github.com/Kitware/trame-vtk/issues/110 and https://github.com/Kitware/trame-vtk/issues/111

user27182:
> My two cents: scope this down to just numpy and try to handle the rest upstream in trame.

I can no longer push changes from remote, and the requested changes are substantial, so I am considering closing this PR and creating a new one from my fork instead.

banesullivan:
>I can no longer push changes from remote, 

Did something break? Do you not have access to push to the repository anymore?

banesullivan:
Ah saw your comment in https://github.com/pyvista/admin/issues/5#issuecomment-4301786705 on it

user27182:
Superseded by https://github.com/pyvista/pyvista/pull/8598

PR #8561, opened 2026-04-18
Replace pyvista-bot PAT with GitHub App and adopt Codecov OIDC

banesullivan:
- Retire `PYVISTA_BOT_TOKEN` across all workflows. Org-wide 2FA enforcement breaks the `pyvista-bot` user account, and App installation tokens are scoped per-workflow, rotated hourly, and not tied to any human user.
- `label.yml` drops the PAT entirely for all 9 labeler calls. `pull_request_target` runs in base-repo context, so the built-in `GITHUB_TOKEN` with `pull-requests: write` is sufficient.
- `update-pr-branch.yml`, `intersphinx-update-pull-request.yml`, `netlify-deploy.yml` (6 uses across preview + two image-report jobs), and `docs.yml` switch to `actions/create-github-app-token`. Each call requests only the repositories and permissions that one job actually needs.
- `docs.yml` mints its cross-repo token scoped to `pyvista/pyvista-examples` with `contents: write`, so the release-tag notebook publish stops needing a cross-repo PAT.
- `testing-and-deployment.yml` switches `codecov-action` to OIDC: adds `id-token: write` on the upload job, sets `use_oidc: true`, removes the `CODECOV_TOKEN` env so the long-lived Codecov secret can also go.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -240,12 +240,25 @@ jobs:
           cp -r doc/source/examples/ ./pyvista-examples/
           ls -l ./pyvista-examples/
 
+      - name: Get app installation token (pyvista-examples)
+        id: app-token
+        if: startsWith(github.ref, 'refs/tags/v')
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3
+        with:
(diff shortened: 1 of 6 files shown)


banesullivan:
The pull request labeler will not pass until this branch hits main

akaszynski (approved):

PR #8562, opened 2026-04-18
Add a `zero()` filter that centers a dataset at the origin

banesullivan:
- Adds `DataObjectFilters.zero()`, a thin wrapper over `translate(-self.center)` that moves a dataset's bounding-box center to `(0, 0, 0)`.
- Works on every dataset type (PolyData, UnstructuredGrid, StructuredGrid, ImageData, RectilinearGrid, MultiBlock) without relying on `.points`, so implicit-geometry types like ImageData and RectilinearGrid are handled correctly.
- Replaces the common `mesh.points -= mesh.center` idiom, which silently breaks for ImageData and RectilinearGrid.
- Boolean parameters are keyword-only (`*`), and the implementation reuses the existing `transform()` pipeline so behavior matches `translate()` exactly.

### DIFF
--- a/pyvista/core/filters/data_object.py
+++ b/pyvista/core/filters/data_object.py
@@ -2431,6 +2431,63 @@ class DataObjectFilters:
             inplace=inplace,
         )
 
+    def zero(  # type: ignore[misc]
+        self: _MeshType_co,
+        *,
+        transform_all_input_vectors: bool = False,
+        inplace: bool = False,
+    ) -> _MeshType_co:
+        """Center the dataset at the origin.
+
+        Translates the dataset so that its bounding box center is at
+        ``(0, 0, 0)``. Equivalent to ``mesh.translate(-np.array(mesh.center))``
+        and works for all dataset types, including those with implicit
+        geometry such as :class:`~pyvista.ImageData` and
+        :class:`~pyvista.RectilinearGrid`.
+
+        .. note::
+            See also the notes at :func:`transform` which is used by this filter
+            under the hood.
+
+        Parameters
+        ----------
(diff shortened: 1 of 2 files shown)


user27182 (commented):
IMO this filter seems redundant. There is already a setter for `center`, so users can already do `mesh.center = (0, 0, 0)`. 

Furthermore, transforming all input vectors does nothing for translations, so this keyword is kind of useless. The only functionality this filter adds that the center setter can’t do is use the `inplace` keyword (the center setter is always an inplace operation). But users can just call `copy` if they want this. 

Also, the name `zero` sounds a bit ambiguous… my initial thought is this filter has something to do with arrays, not geometry. Perhaps `center_at_origin` is a better name.

banesullivan:
My reasoning for this was to be able to chain filters together: mesh.zero().rotate_x(90) . setting the center property isn’t very ergonomic in that situation however, I agree this filter is pretty redundant and just calling the translate filter works for what I need to do

user27182:
> My reasoning for this was to be able to chain filters together: mesh.zero().rotate_x(90) . setting the center property isn’t very ergonomic in that situation however, I agree this filter is pretty redundant and just calling the translate filter works for what I need to do

I see the value in being able to chain these methods. I had not considered that use case. We can still keep this filter then, though I still think `center_at_origin` is a better name. I would recommend adding crossrefs with `center` and `translate` in this case. 


Also it sounds to me like using the `Transform` class maybe better suited to what you need. It supports automatic translations to/from origin for applying rotations and/or scale transformations about the mesh’s center. I believe there is also a similar `point` keyword available with the rotation filters.

Issue #8523, opened 2026-04-15
Third-Party Filter Registration/Extension


banesullivan:
## Describe the feature you would like to be added.

PyVista should expose a public mechanism for third-party packages to register custom filter methods against specific dataset types, so that ecosystem tools can extend the fluent filter API on `pyvista.DataSet` subclasses without monkey-patching or subclassing. Beyond the immediate ergonomic win, this keeps PyVista's core API lean and prevents feature bloat by delegating niche algorithms to downstream packages.

### Motivation

Several PyVista ecosystem packages implement domain-specific operations that are most naturally expressed as filters bound to a dataset instance:

- [pymeshfix](https://github.com/pyvista/pymeshfix): mesh repair on `PolyData`
- [tetgen](https://github.com/pyvista/tetgen): tetrahedralization from `PolyData` to `UnstructuredGrid`
- Filters that wrap capabilities only available in custom VTK builds (e.g. the ONNX filter discussed in https://github.com/pyvista/pyvista/pull/7993), which cannot live in pyvista-core because the default VTK wheels do not expose them and we cannot test or validate them here.

Today, each of these packages either (a) ships a standalone function the user must import and call, breaking the chained-filter ergonomics PyVista users expect, or (b) subclasses `PolyData` / `UnstructuredGrid`, which does not compose well. A first-class registration API would let third-party libraries plug in cleanly and would give the core project a principled way to say "this belongs in a plugin, not pyvista-core" when a VTK capability is not in the default build.

### Prior art outside PyVista: pandas and xarray accessors

The accessor pattern was pioneered by [pandas](https://pandas.pydata.org/docs/development/extending.html#registering-custom-accessors) and popularized further by xarray, which borrowed the design almost verbatim. Both are strong success stories and together cover nearly every extensibility case PyVista is likely to hit.

**pandas** exposes `pd.api.extensions.register_dataframe_accessor("name")`, `register_series_accessor("name")`, and `register_index_accessor("name")`. The accessor is a class whose `__init__` receives the pandas object, is cached per-instance on first access, and owns a single top-level namespace. pandas' own built-in `.str`, `.dt`, `.cat`, and `.sparse` accessors use exactly this mechanism internally, so the pattern is not just sanctioned for third parties; it is how pandas organizes its own domain-specific method surface. The pandas docs explicitly recommend accessors as the preferred alternative to subclassing.

[...]

PR #8568, opened 2026-04-19
Add theme registry and entrypoints

banesullivan:
# Theme registration

Makes `Theme` subclasses and plugin themes discoverable by name. Extends `set_plot_theme()` and `PYVISTA_PLOT_THEME` to accept registered names and `package.module:ClassName` dotted paths, and adds `pv.registered_themes()` for inspection.

Two registration paths, one lookup.

## For subclass authors

Declare a class-level `_default_name` and the theme is registered automatically:

```python
from typing import ClassVar
from pyvista.plotting.themes import DocumentTheme

class MyTheme(DocumentTheme):
    _default_name: ClassVar[str] = 'mine'

pv.set_plot_theme('mine')
```

Ad-hoc `class X(Theme): ...` subclasses without `_default_name` are ignored, so this is opt-in. This path requires the defining module to be imported before the name resolves, so it is useful for scripts, notebooks, testing, and local development.

## For plugin packages

Declare a `pyvista.themes` entry point in `pyproject.toml`:

```toml
[project.entry-points.'pyvista.themes']
my_theme = 'my_package.theme:MyTheme'
```

PyVista discovers it lazily on first lookup. `PYVISTA_PLOT_THEME=my_theme` picks it up on import without requiring the user to import the plugin first. This mirrors the existing `pyvista.interactor_styles`, `pyvista.readers`, and `pyvista.writers` entry-point groups.

## Dotted paths

`PYVISTA_PLOT_THEME=pkg.mod:ClassName` and `pv.set_plot_theme('pkg.mod:ClassName')` both work without the package shipping an entry point.

## Discoverability

`pv.registered_themes()` returns a `tuple[ThemeRegistration, ...]` describing how each name was registered. `ThemeRegistration` is a `NamedTuple(name, kind, source)` with `kind` in `{'subclass', 'entry_point', 'alias'}`:

```python
>>> next(r for r in pv.registered_themes() if r.name == 'dark')
ThemeRegistration(name='dark', kind='subclass', source='pyvista.plotting.themes.DarkTheme')
>>> next(r for r in pv.registered_themes() if r.name == 'default').kind
'alias'
```

## Alignment with #8595

#8595 unified collision, source-attribution, and lazy-loading conventions across PyVista's plugin registries. This PR adopts them for themes:

- Last-wins collision: registering a name that's already taken now warns and replaces, matching the accessor, reader, writer, jupyter-backend, and interactor-style registries.
- `registered_themes()` returns `tuple[ThemeRegistration, ...]` to match `registered_accessors()` / `registered_readers()` / `registered_writers()` / `registered_jupyter_backends()`.
- Source attribution uses the shared `handler_source` helper from `pyvista/core/utilities/_registry_helpers.py`.
- Per-name lazy entry-point loading: `set_plot_theme('document')` no longer imports any installed `pyvista.themes` plugins; only the plugin claiming the requested name (or `registered_themes()`, which forces full discovery) does.

## Context

#8030 (@edabor) added `register_theme` as a combined function and decorator. Earlier revisions of this PR preserved `register_theme()` for pre-configured instances, but @MatthewFlamm pointed out that instance registration requires the same user import as plain subclassing and does not add discoverability beyond what entry points and subclasses already provide. Removing it trims ~230 lines and leaves two clearly differentiated paths.

#5119 (@MatthewFlamm) proposed replacing the subclass API with classmethods and deprecating the `Theme` subclasses. Downstream projects depend on subclassing, so this PR preserves it.

## Backward compatibility

Fully preserved. `_name` stays in `Theme.__slots__`, the `name` property still writes to it, `_NATIVE_THEMES` is unchanged, and `'default'`/`'vtk'` still resolve to `DocumentTheme`/`Theme` via explicit aliases.

### DIFF
--- a/doc/source/api/plotting/conv_func.rst
+++ b/doc/source/api/plotting/conv_func.rst
@@ -11,7 +11,6 @@ routines in PyVista.
    plot
    plot_arrows
    set_plot_theme
-   register_theme
    registered_themes
    ThemeRegistration
    create_axes_orientation_box
--- a/doc/source/api/plotting/conv_func.rst
+++ b/doc/source/api/plotting/conv_func.rst
@@ -13,6 +13,7 @@ routines in PyVista.
    set_plot_theme
    register_theme
    registered_themes
+   ThemeRegistration
    create_axes_orientation_box
    create_axes_marker
    opacity_transfer_function
--- a/doc/source/api/plotting/conv_func.rst
+++ b/doc/source/api/plotting/conv_func.rst
@@ -11,6 +11,8 @@ routines in PyVista.
    plot
    plot_arrows
    set_plot_theme
+   register_theme
+   registered_themes
    create_axes_orientation_box
    create_axes_marker
    opacity_transfer_function
--- a/pyvista/plotting/theme_registry.py
+++ b/pyvista/plotting/theme_registry.py
@@ -6,6 +6,8 @@ from collections.abc import Mapping
 from importlib import import_module
 from importlib.metadata import entry_points
 from typing import TYPE_CHECKING
+from typing import Literal
+from typing import NamedTuple
 from typing import TypedDict
 
 from pyvista._warn_external import warn_external
@@ -17,11 +19,42 @@ if TYPE_CHECKING:
 THEME_ENTRY_POINT_GROUP = 'pyvista.themes'
 
 
+ThemeRegistrationKind = Literal['subclass', 'instance', 'entry_point', 'alias']
+
+
+class ThemeRegistration(NamedTuple):
+    """Describe how a theme name became registered.
+
+    Returned by :func:`~pyvista.registered_themes`. ``kind`` distinguishes
+    the registration path: a :class:`~pyvista.plotting.themes.Theme`
+    subclass with a class-level ``_default_name`` (``'subclass'``), an
+    instance passed to :func:`~pyvista.register_theme` (``'instance'``),
+    a ``pyvista.themes`` entry point from an installed plugin
+    (``'entry_point'``), or a built-in legacy alias such as ``'default'``
+    or ``'vtk'`` (``'alias'``).
+
+    Attributes
+    ----------
+    name : str
+        The registered (normalized) theme name.
+    kind : str
+        One of ``'subclass'``, ``'instance'``, ``'entry_point'``, ``'alias'``.
+    source : str
+        Human-readable origin (e.g. ``'my_package.theme.MyTheme'``).
+
+    """
+
+    name: str
+    kind: ThemeRegistrationKind
+    source: str
+
+
 class _ThemeRegistryState(TypedDict):
     """Stored registry state used by tests."""
 
     classes: dict[str, type[Theme]]
     classes_sources: dict[str, str]
+    aliases: set[str]
     instances: dict[str, Theme]
     instances_sources: dict[str, str]
     discovered: dict[str, type[Theme] | Theme]
@@ -31,6 +64,7 @@ class _ThemeRegistryState(TypedDict):
 
 _registered_theme_classes: dict[str, type[Theme]] = {}
 _registered_theme_classes_sources: dict[str, str] = {}
+_registered_theme_aliases: set[str] = set()
 _registered_theme_instances: dict[str, Theme] = {}
 _registered_theme_instances_sources: dict[str, str] = {}
 _discovered_entry_point_themes: dict[str, type[Theme] | Theme] = {}
@@ -48,6 +82,7 @@ def _save_registry_state() -> _ThemeRegistryState:
     return {
         'classes': _registered_theme_classes.copy(),
         'classes_sources': _registered_theme_classes_sources.copy(),
+        'aliases': _registered_theme_aliases.copy(),
         'instances': _registered_theme_instances.copy(),
         'instances_sources': _registered_theme_instances_sources.copy(),
         'discovered': _discovered_entry_point_themes.copy(),
@@ -63,6 +98,8 @@ def _restore_registry_state(state: _ThemeRegistryState) -> None:
     _registered_theme_classes.update(state['classes'])
     _registered_theme_classes_sources.clear()
     _registered_theme_classes_sources.update(state['classes_sources'])
+    _registered_theme_aliases.clear()
+    _registered_theme_aliases.update(state['aliases'])
     _registered_theme_instances.clear()
     _registered_theme_instances.update(state['instances'])
     _registered_theme_instances_sources.clear()
@@ -140,7 +177,8 @@ def _register_alias(name: str, cls: type[Theme]) -> None:
     """
     normalized = _normalize_theme_name(name)
     _registered_theme_classes[normalized] = cls
-    _registered_theme_classes_sources[normalized] = f'{cls.__module__}.{cls.__qualname__} (alias)'
+    _registered_theme_classes_sources[normalized] = f'{cls.__module__}.{cls.__qualname__}'
+    _registered_theme_aliases.add(normalized)
 
 
 def _lookup_explicit(normalized: str) -> Theme | None:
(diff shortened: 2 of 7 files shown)


MatthewFlamm:
While I still believe the classmethod approach is much more sane from a design perspective, practically there have been few to no issues about this.  One problem is the messiness around inspection of type with the way `global_theme` works.  Moving this PR forward would more or less cement the current structure.  I'm okay with this given the theoretical nature of the issue. 

The other current issue is around `slider_style`, which in practice works like one config but has two default styles.  It is inconsistent with all other configs.  #5119 created a structure such that users could set default configs for any 'sub-theme'.  This is one of the more powerful additions of that PR.  Beyond giving more options for downstream hooks, it allows us to clean up the inconsistency of `slider_style`.  It was naturally solved by the classmethod registration approach.  It isn't clear to me how to solve this issue with the class based approach, but that doesn't mean it isn't possible.

pyvista/plotting/themes.py
MatthewFlamm:
```suggestion
    """Set plotting parameters to a predefined theme using a string or ``Theme``.
```

MatthewFlamm: The docstring is no longer fully accurate, gave a suggestion, but not 100% happy with it.

pyvista/plotting/themes.py
```diff
@@ -120,6 +129,14 @@ def set_plot_theme(theme):
         - ``'testing'`` and
         - ``'vtk'``.
 
+        Any :class:`~pyvista.plotting.themes.Theme` subclass that declares
```
MatthewFlamm: This paragraph, particularly the part about `_default_name`, is aimed at downstream package developers.  End users only need to know that any downstream packages can hook in a name here.  I would prefer the documentation of this function to list how users can discover the names.  Instructions for downstream maintainers should be separate IMO, but there could be a link here.  See other comment about making public the list of registered themes.

pyvista/plotting/theme_registry.py
```diff
@@ -0,0 +1,295 @@
...
+    if resolved is not None:
+        return resolved
+
+    discovered = _discovered_entry_point_themes.get(normalized)
+    if isinstance(discovered, type):
+        return discovered()
+    return discovered
+
+
+def _available_theme_names() -> tuple[str, ...]:
```
MatthewFlamm, line 218: This should be made public in some fashion.  We should decorate that return with how the theme was registered.

MatthewFlamm (commented):
I have not done a deep dive into the mechanics, but some comments on structure of both exposed functions and documentation to start.

pyvista/plotting/themes.py
banesullivan: Agreed, the `_default_name` detail doesn't belong in the user-facing parameter description. I reworked `set_plot_theme` to show the three input shapes (registered name, dotted path, `Theme` instance) and point users at the new `registered_themes()` function for discovery. 

Moved the `__init_subclass__`, `_default_name`, and entry-point guidance to the `Theme` class docstring under a Notes section explicitly scoped to theme authors.

pyvista/plotting/theme_registry.py
banesullivan: Added `pv.registered_themes()` which returns `dict[str, ThemeRegistration]` (`NamedTuple(name, kind, source)`):

```python
>>> pv.registered_themes()['dark']
ThemeRegistration(name='dark', kind='subclass', source='pyvista.plotting.themes.DarkTheme')
>>> pv.registered_themes()['default'].kind
'alias'
```

banesullivan:
Pushed some changes to address everything you raised, @MatthewFlamm 

On the classmethod vs class-based and `slider_style` inconsistency: I kept the subclass path here because downstream projects already rely on it, but I agree the structure in #5119 handles the `slider_style` dual-default case more naturally than this does. I think we could still pick up the sub-theme config change as a separate follow-up once this lands if there is desire (layered on top of the class-based API rather than replacing it)

tests/plotting/test_theme_registry.py
```diff
@@ -0,0 +1,506 @@
...
+from pyvista.plotting.themes import _set_plot_theme_from_env
+
+
+def test_init_subclass_auto_registers():
+    class FooTheme(Theme):
+        _default_name: ClassVar[str] = 'foo_theme'
+
+    assert _reg_mod._resolve_theme('foo_theme') is not None
+    assert 'foo_theme' in _reg_mod._available_theme_names()
+
```
MatthewFlamm, line 28: Check that it is in public function for available theme names too.

tests/plotting/test_theme_registry.py
```diff
@@ -0,0 +1,506 @@
...
+def test_init_subclass_missing_default_name_is_silent():
+    """Ad-hoc subclasses without ``_default_name`` should register silently."""
+    with warnings.catch_warnings():
+        warnings.simplefilter('error')
+
+        class NoName(Theme):
+            pass
+
+    assert 'noname' not in _reg_mod._available_theme_names()
```
MatthewFlamm, line 66: We should also test that the registered themes are the same before and after.  This would ensure that it doesn't register as some other name.

tests/plotting/test_theme_registry.py
```diff
@@ -0,0 +1,506 @@
...
+    resolved = _reg_mod._resolve_theme('derived_doc')
+    assert isinstance(resolved, Derived)
+
+    # DocumentTheme still resolves to DocumentTheme, not Derived
+    doc = _reg_mod._resolve_theme('document')
+    assert type(doc) is DocumentTheme
+
+
+def test_init_subclass_missing_default_name_is_silent():
+    """Ad-hoc subclasses without ``_default_name`` should register silently."""
```
MatthewFlamm: Does this mean the intent is to 'register silently', or it is 'not registered silently'?  Based on the test and my expectation, it should not register.

tests/plotting/test_theme_registry.py
```diff
@@ -0,0 +1,506 @@
...
+
+def test_register_theme_instance():
+    custom = DarkTheme()
+    custom.show_edges = True
+
+    returned = pv.register_theme('my_custom', custom)
+    assert returned is custom
+
+    pv.set_plot_theme('my_custom')
+    assert pv.global_theme.show_edges is True
```
MatthewFlamm: Also test that setting plot theme to 'dark' maintains the expected behavior, i.e. it doesn't corrupt the Class theme via instance registration.

doc/source/api/plotting/conv_func.rst
```diff
@@ -11,6 +11,8 @@ routines in PyVista.
    plot
    plot_arrows
    set_plot_theme
+   register_theme
+   registered_themes
```
MatthewFlamm, line 14: We need to add `ThemeRegistration` if it is referenced in `registered_themes`.

pyvista/plotting/theme_registry.py
```diff
@@ -0,0 +1,395 @@
...
+    """Return all registered themes, keyed by name.
+
+    Use this to discover which names can be passed to
+    :func:`~pyvista.set_plot_theme` or the ``PYVISTA_PLOT_THEME``
+    environment variable. Entry-point plugins are loaded on the first
+    call so they appear in the result.
+
+    Returns
+    -------
+    dict[str, ThemeRegistration]
```
MatthewFlamm: If this is kept, then we need to document `ThemeRegistration` as user documentation.

MatthewFlamm (commented):
One overarching comment.  What is the use case for allowing the class based and instance based registration overall?  For end users, they can simply create a sub-class and then load it into the global theme.  For downstream packages, it is also not ideal as I think it requires the end-user to `import package_containing_theme` for the registration to occur.  If there is no compelling case, I would either remove this functionality for easier maintenance, or document it as being for ease of testing/implementation for downstream packages but we recommend the entrypoint.

MatthewFlamm (commented):
This is looking good to me.  

I read #8523 in more detail in the meantime.  In this PR I had suggested greedy loading of theme registration via entrypoint as the suggested method.  However, in #8523 it is stated that

> Explicit opt-in. Users are never surprised to find new methods on their datasets because some unrelated package happens to be installed. The namespace only exists if the user asked for it by importing the plugin.

Registrations that result in attributes being added is more explicitly obvious to users, while registrations that result in string keys being available to use in functions are less obvious to users.  The registration mechanism in this PR is still opt-in in the sense that users have to explicitly request the key from the function.  However, it might lead to confusion for users, e.g. 'when do I need to `import pyvista` to get the registration functionality?'

The edge case I can come up with is if a user installs two packages with conflicting theme registration names.  They will no longer be able to control which theme gets registered.  To avoid this edge case, we could also use something like a namespace for the entrypoint registered names, e.g. pv.set_plot_theme(`my_package.themes.my_awesome_theme`).  Not really ideal either.  Requiring an import would allow users to choose which package to import first to register the theme.

This is more thinking out loud for alignment than a specific proposal.

banesullivan:
Yeah, that's a good call. I dropped `register_theme()` entirely in the latest push... it indeed felt redundant and not really solving a real problem

This PR provides two registration paths now:

- Entry points via the `pyvista.themes` group: the only mechanism that gives import-free discovery, so the recommended path for third party packages.
- Subclass `_default_name`: auto-registers on class definition and requires the module to be imported first. Useful for scripts, notebooks, testing, and local dev.

I kept the subclass path I've seen a number of downstream projects follow this pattern and entry points can point at a class to reuse the same registration machinery.

akaszynski (approved):

PR #8577, opened 2026-04-20
typing: narrow pyvista.read() return via cls overload

banesullivan:
## Overview

Two static-typing ergonomics wins for `pv.read()` and the `BaseReader` hierarchy, plus a refactor of the reader → output-type metadata that was previously hand-maintained.

### `pv.read(..., cls=...)` — explicit return-type narrowing

`pv.read()` returns `DataSet | MultiBlock`, so callers who know the concrete type had to `cast(...)`, `assert isinstance(...)`, or eat type-checker errors. This PR adds a `cls=` keyword with two `@overload` stubs:

```python
mesh = pv.read('file.vtu', cls=pv.UnstructuredGrid)  # inferred as UnstructuredGrid
mesh = pv.read('file.vtu')                            # still DataSet | MultiBlock
```

At runtime, `isinstance(result, cls)` is enforced and a `TypeError` is raised on mismatch.

### `BaseReader` is now `Generic[_T_Output_co]`

Every concrete reader is parameterized with its output type: `XMLPolyDataReader(BaseReader['PolyData'])`, `DICOMReader(BaseReader['ImageData'])`, etc. `BaseReader.read()` returns `_T_Output_co`, so `pv.XMLPolyDataReader(path).read()` is statically known to be `PolyData` without a `cast`. One base `read()` implementation; no per-subclass overrides required.

### Reader output-type mapping is derived, not duplicated

Addresses @user27182's review comment. The pre-existing `_CLASS_READER_RETURN_TYPE` dict (added in #8091 for docs/tests) was a 50-entry hand-maintained duplicate of information the type system now carries. It's now a dict comprehension over `CLASS_READERS`:

- Single-output readers (the common case): derived from the `BaseReader[X]` generic parameter via `__orig_bases__`. No manual bookkeeping.
- Multi-output readers (7 of them — `HDFReader`, `XdmfReader`, `SegYReader`, `GaussianCubeReader`, `VTKDataSetReader`, `VTKPDataSetReader`, `SeriesReader`): declare `_output_types: ClassVar[tuple[_mesh_types, ...] | None]` next to `_vtk_class_name`. The `_mesh_types` Literal bound means mypy catches typos at the class-definition site.

## Ripple effects

Several `# type: ignore` comments and manual `isinstance` checks fall out:

- `pv.compare_images`: 7-line "read → isinstance → raise TypeError" block → `read(img, cls=ImageData)`.
- `BackgroundRenderer`, `Texture._from_file`, `_cubemap_from_paths`, `WidgetHelper.add_logo_widget`: all use `cls=ImageData` instead of the wrap-and-check pattern.
- Dozens of `# type: ignore[arg-type]` in `pyvista/examples/downloads.py` / `examples/examples.py` are gone because `BaseReader[Any]` now accepts the callable types cleanly.
- Corrects a latent annotation bug in `_dataset_loader.py`: `Callable[[str], DatasetType]` (returns a type object?) → `Callable[[str], DatasetObject]` (returns a dataset instance).

## Tests

New coverage in `tests/core/test_reader.py`:

- `test_read_cls_narrows`, `test_read_cls_mismatch_raises`, `test_read_cls_none_behaves_like_default` — runtime behavior of the `cls=` parameter.
- `test_reader_output_type_derived_from_generic`, `test_reader_output_type_override_for_multi_output` — guard the single-output / multi-output paths of the new derivation.
- `test_extract_base_reader_generic_arg_*`, `test_derive_reader_output_types_*` — direct unit tests for the two derivation helpers.


user27182:
I know this is draft but wanted to highlight that the type mapping between reader and output type is already defined here:
https://github.com/pyvista/pyvista/blob/a79cdc06087971cc7c53affb14d664a260657812/pyvista/core/utilities/reader.py#L4083

this was added in https://github.com/pyvista/pyvista/pull/8091, but is only used for the docs and tests at the moment. It seems to me that with this PR, we should (maybe?) be able to auto-generate this mapping based on the new reader type. Or perhaps in some way refactor this to avoid duplicating this info.

pyvista/core/filters/data_set.py
```diff
@@ -8230,7 +8230,7 @@ def _preprocess_polydata(poly_in):
             if background_value == 0
             else np.ones(scalars_shape, dtype=scalars_dtype) * background_value
         )
-        binary_mask['mask'] = scalars  # type: ignore[type-var]
+        binary_mask['mask'] = scalars
```
banesullivan: Whoops. This keeps popping up locally for me

banesullivan:
```suggestion
        binary_mask['mask'] = scalars  # type: ignore[type-var]
```

akaszynski (approved):

PR #8578, opened 2026-04-21
Support glob patterns in CLI for bulk file conversions

banesullivan:
## Summary

Extend `pyvista convert` to accept multiple input files and glob patterns so meshes can be bulk-converted in one invocation. Previously the command took exactly one input and one output, so batch conversions (e.g. `*.vtu` → `.pv`) required a shell loop.

## Details

- CLI signature is now `FILE-IN [FILE-IN...] FILE-OUT`. The existing single-input + named-output form still works.
- Glob characters (`*`, `?`, `[...]`) are expanded inside the converter, so both `pyvista convert *.vtu .pv` (shell-expanded) and `pyvista convert '*.vtu' .pv` (quoted) work. Patterns with no matches surface as a `file not found` error.
- Output spec semantics for multi-input runs:
  - Bare extension `.pv` writes next to each input, so `pyvista convert sub/*.vtu .pv` drops `.pv` files into `sub/` alongside the sources.
  - Explicit parent `out/.pv` writes into that directory.
  - A named output alongside multiple inputs is rejected as ambiguous.
- Streaming read+save: each mesh is read, saved, and released per file. Hundreds of files now convert without holding all meshes in memory and without the silent upfront read phase that caused the "hangs with no output" experience.
- Rich progress bar for multi-input runs shows current filename, N/M counter, elapsed time, and ETA. Single-input runs still print the one-shot `Saved: <path>` line. Final summary reports the target directory (or `across K directories` when outputs span several paths).
- Extracted `_check_paths_exist` and refactored `_load_paths` in `pyvista/_cli/utils.py` so `convert` can validate upfront without eagerly reading meshes. `_converter_files` stays a thin wrapper, and `plot`/`validate` pick up glob expansion for free.
- Moved output-extension validation from the cyclopts validator into the handler so all errors render through `_console_error` (the same `PyVista Error` panel used for file-not-found and save failures).

## Tests

Added coverage for multi-input + extension output, multi-input + subdir output, shell-expanded and quoted globs, no-match glob, named-output-with-many-inputs rejection, too-few-args rejection, and the subdir-adjacent output behavior. Existing tests updated for the new help text and error surface.

-----

Screenshot 2026-04-20 at 8 14 20 PM


user27182 (commented):
Been meaning to test this locally but haven’t yet had time to check this out. Here are my main concerns:

> `plot/validate` pick up glob expansion for free

Is this also documented / tested? Maybe even just a quick sanity test locally 

Also, does this feature work properly with `.mb` files? IIRC this format uses a file to store metadata, but also uses a folder structure to store nested meshes as separate files. Might be a risk that both the MultiBlock file is converted as a whole, but also child meshes are separately converted, resulting in a single MultiBlock input being incorrectly converted as multiple output files. 

Another edge case is with DICOM stacks. These are usually read as a dir, rather than reading individual files. It seems this PR would probably convert the individual slice files, rather than the full stack, which is probably OK, though I wonder how we could support the use case of “I have a directory with multiple DICOM stacks, and want to convert all stacks as individual volume files” 

E.g. if I have a dir with 5 DICOM stacks, each a directory with 10 `.dcm` files, this PR probably creates 5 directories with 10 converted files, i.e. 50 files. But for use case above, 5 files should be returned (one for each stack). 

maybe a follow-up feature is to support an `—image-stack` option? In cases where multiple 2D images are found with the same extension in a common dir, first try reading the dir itself (reading a dir directly is supported by DICOMReader), otherwise for a dir with PNG files only, read them separately but use `concatenate` filter to combine sliced into a single volume.

user27182:
> Also, does this feature work properly with `.mb` files? IIRC this format uses a file to store metadata, but also uses a folder structure to store nested meshes as separate files. Might be a risk that both the MultiBlock file is converted as a whole, but also child meshes are separately converted, resulting in a single MultiBlock input being incorrectly converted as multiple output files.

I'm specifically thinking of nested MultiBlocks, e.g. child `.mb` files inside a parent `.mb`. How is the child `.mb` file handled?

EDIT: No need to necessarily change code for this, and `glob` should probably work as expected. But if there is a risk that child  nested meshes are duplicated, maybe just document it as such with a warning? 

Except that a key use case for this tool would be to convert existing `.mb` files into new `.pv` files. So perhaps we should handle this correctly after all in terms what the user expects as output, e.g. `pyvista convert *.mb .pv` should always have a 1:1 mapping of input files to output files, regardless of nested `.mb` files (i.e. nested `.mb` need to be removed from the `glob` output before converting)

akaszynski:
> maybe just document it as such with a warning?

+1

banesullivan:
Thanks for taking a look, @user27182. I've pushed a follow-up that addresses the MultiBlock concerns

Added a new `_filter_multiblock_children` helper in `pyvista/_cli/utils.py` that drops paths that live inside the sibling sidecar directory of a `.vtm`/`.vtmb` parent already in the input list, so a recursive glob like `convert **/* .pv` (or any input list that mixes a parent and its children) still produces a 1:1 input/output mapping. When files are dropped, the converter prints a warning `Skipping N file(s) inside MultiBlock sidecar directories..` so the you can see what was filtered. Two new tests cover the parent+children case and the orphan-children case

I went with filter-and-warn rather than warn-only

Re: nested `.vtm` inside a parent `.vtm`, VTK's `vtkXMLMultiBlockDataWriter` flattens these on save, so the on-disk structure is always one parent + flat sidecar children. The filter handles that case. If anyone hits a hand-crafted nested case in the wild we can revisit.

```py
>>> import pyvista as pv
>>> nested = pv.MultiBlock([pv.Cone(), pv.Cube()])
>>> top = pv.MultiBlock([pv.Sphere(), nested])
>>> top.save("foo.vtm")
>>> exit()
```
```bash
❯ ls -a foo/
foo_0.vtp  foo_1.vtp  foo_2.vtp
```

I added help text on both `plot`/`validate` to document glob support. Validate will validate all meshes passed via the glob now too

user27182 (approved):

PR #8583, opened 2026-04-22
Add PNG repr to plotter

banesullivan:
Adds a simple fallback PNG repr for the Plotter for data science IDEs and the like that support `_repr_png_`

### DIFF
--- a/pyvista/plotting/plotter.py
+++ b/pyvista/plotting/plotter.py
@@ -30,7 +30,6 @@ import warnings
 import weakref
 
 import numpy as np
-import PIL.Image
 import scooby
 
 import pyvista as pv
@@ -6633,6 +6632,8 @@ class BasePlotter(_BoundsSizeMixin, PickingHelper, WidgetHelper):
             the plotter has not been rendered.
 
         """
+        import PIL.Image  # noqa: PLC0415
+
         if self.last_image is None:
             return None
         buf = io.BytesIO()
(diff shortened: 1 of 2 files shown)


pyvista/plotting/plotter.py
```diff
@@ -30,6 +30,7 @@
 import weakref
 
 import numpy as np
+import PIL.Image
```
banesullivan: Should I move this import to `_repr_png_` with a `# noqa: PLC0415`? We do that in a few other places with pillow

pyvista/plotting/plotter.py
```diff
@@ -6615,6 +6616,29 @@ def screenshot(  # noqa: PLR0917
...
+        captured yet; this method never triggers a render, so IDE variable
+        inspectors can probe a plotter without side effects.
+
+        Returns
+        -------
+        bytes or None
+            PNG-encoded image of the most recent screenshot, or ``None`` if
+            the plotter has not been rendered.
+
+        """
```
banesullivan, line 6634:
```suggestion
        """
        import PIL.Image  # noqa: PLC0415

```

pyvista/plotting/plotter.py
banesullivan:
```suggestion
```

user27182: > Should I move this import to `_repr_png_` with a `# noqa: PLC0415`? We do that in a few other places with pillow

this should probably be enforced by adding Pillow to this test:
https://github.com/pyvista/pyvista/blob/3c80baff8c590d902fa46641ab081a3cf2d3b583/tests/test_init.py#L32

akaszynski (approved):

PR #8584, opened 2026-04-22
Filter out VTK methods/attributes in __dir__

### LINKED ISSUE #8582: Make PyVista data objects pleasant to inspect in data-science IDEs
### Motivation

When I open a PyVista mesh in Positron's Variables pane (same thing happens in VS Code's Jupyter extension, etc), I get over 400 entries... a lot of them are VTK methods inherited from the inherited base class: `GetBounds`, `AddObserver`, `BreakOnError`, `DebugOn`, `DeepCopy`, `GetAttributesToInterpolate`, and so on. Finding the PyVista-defined properties (`points`, `bounds`, `point_data`, `cell_data`, `n_points`, etc.) is... not working.

Compare the same workflow with xarray. `Dataset` objects browse cleanly in Positron because the public surface is curated. PyVista has great `_repr_html_` for notebook cell output, but the variables pane is a separate inspection path that goes through `dir()`, and `dir()` on a PyVista mesh is overwhelming.

Tab-completion has the same problem. `mesh.<TAB>` in IPython/Jupyter floods you with VTK names in addition to PyVista properties and methods.

### Proposal

Override `__dir__` on `DataObject` which will propagate to all datatypes (`DataSet`, `PolyData`, `UnstructuredGrid`, `StructuredGrid`, `ImageData`, `RectilinearGrid`, `MultiBlock`, `PartitionedDataSet`, and `Table`) to filter out attributes that are defined on the VTK base classes

The filter can reuse the `is_vtk_attribute` helper that already exists, so there's no new machinery to maintain or implement really

https://github.com/pyvista/pyvista/blob/3c80baff8c590d902fa46641ab081a3cf2d3b583/pyvista/core/_vtk_utilities.py#L148-L169

```python
def __dir__(self):
    return sorted(
        attr for attr in super().__dir__()
        if not is_vtk_attribute(self, attr)
    )
```

VTK methods stay fully callable, only the enumeration surface changes.

### Breaking change caveat?

This is technically breaking for anyone who:

- Relies on `dir(mesh)` enumerating VTK methods (introspection scripts, docs generators).
- Uses tab-completion to discover the raw VTK API (`GetBounds`, `DeepCopy`, `AddObserver`, etc.).

so this raises some questions:

- Do we want this as a hard filter, or opt-in via something like `pyvista.set_options(show_vtk_api=True)` so VTK users/developers can get the full surface back?
- Should we also do this to`Renderer`/`Actor`/`Property`? I think not on this first go as my intention is to improve the API surface to data-science workflows.
- Any known downstream workflows (docs tooling?) that would break and need a heads-up/refactor?

banesullivan:
Resolves #8582 

Makes tab completion so much more sane and improves the UX in data science IDEs like positron

| Before | After |
|---|---|
| ~622 on `pv.Sphere()` | ~331 on `pv.Sphere()`|
| Screenshot 2026-04-21 at 6 33 12 PM | Screenshot 2026-04-21 at 6 33 24 PM |

PyVista objects inherit hundreds of VTK methods (`GetBounds`, `DeepCopy`, `AddObserver`, ...) that swamp `dir()` and tab-completion. `dir(pv.Sphere())` returns 622 entries before this change. The curated PyVista API (`points`, `bounds`, `point_data`, `n_points`, ...) is buried under inherited VTK machinery, which makes PyVista datasets painful to browse in Positron's Variables pane, VS Code's Jupyter extension, and IPython.

This PR filters VTK-inherited attributes out of `__dir__` by default. `dir(pv.Sphere())` drops to 333 entries. VTK methods stay fully callable; only their enumeration in `dir` is suppressed.

## New config: `pv.global_config.show_vtk_api`

A boolean on `pv.global_config` (default `False`) controls whether CamelCase VTK attributes appear in `dir`. Set it to `True` to restore the full VTK surface for introspection, docs generators, or any workflow that enumerates VTK methods.

```python
import pyvista as pv

pv.global_config.show_vtk_api = True   # VTK API visible in dir/tab-completion
pv.global_config.show_vtk_api = False  # default: VTK API hidden
```

## Where the filter lives

`__dir__` moved to the `DisableVtkSnakeCase` mixin, so the filter applies to every PyVista class that wraps a VTK type: `DataObject` and all subclasses, `Renderer`, `Actor`, `Property`, `Camera`, `Light`, chart classes, source classes, `DataSetAttributes`, and more. One override, broad coverage.

## Snake_case VTK aliases are tied to `pv.vtk_snake_case()`

VTK 9.4+ ships snake_case aliases for its methods (`information`, `get_bounds`, `deep_copy`, ...). PyVista already blocks access to those via `pv.vtk_snake_case('error')` by default, so surfacing them in `dir` would only mislead users into calling attributes that raise `PyVistaAttributeError`. The filter handles snake_case independently: hidden unless `pv.vtk_snake_case('allow')` is active. `show_vtk_api` only controls the CamelCase surface, so the two states toggle independently.


banesullivan:
Blocked by https://github.com/pyvista/pyvista/pull/8493

user27182:
> Blocked by #8493

Feel free to separate the global config aspect of that PR into its own PR to unblock this one. (I currently have limited availability/access to a computer for review, but will definitely review PRs where my review is requested in due course, within a week or so)

pyvista/core/dataobject.py
```diff
@@ -104,6 +105,19 @@ def __getattr__(self: Self, item: str) -> Any:
         """Get attribute from base class if not found."""
         return super().__getattribute__(item)
 
+    def __dir__(self: Self) -> list[str]:
```
user27182: Consider moving this to the `DisableVtkSnakeCase`  class to apply this to all classes that inherit from VTK classes. And possibly rename the class to `FilterVtkAtrributes` to reflect the new and more general behavior. 

Further, I think all snake-case attributes should be filtered conditionally based on `pv._VTK_SNAKE_CASE_STATE`.

And the non-snake-case attributes should be similarly filtered conditionally based on a separate global config value.

tests/core/test_dataobject.py
user27182, line 1: Consider moving attribute-related tests to `test_attributes.py`
https://github.com/pyvista/pyvista/blob/main/tests/test_attributes.py

doc/source/conf.py
```diff
@@ -124,6 +128,30 @@
...
+def _resilient_vtk_xref_get(url, *args, **kwargs):
+    kwargs.setdefault('timeout', 15)
+    last_exc = None
+    for attempt in range(3):
+        try:
+            return requests.get(url, *args, **kwargs)
+        except requests.RequestException as exc:
+            last_exc = exc
+            time.sleep(attempt + 1)
+    raise last_exc  # type: ignore[misc]
```
banesullivan: @user27182, should it be thirty seconds upstream? Not three? 

https://github.com/pyvista/vtk-xref/blob/6cf55a93e05ce22f8e8735ab0c9e9360d1521047/vtk_xref/__init__.py#L77

banesullivan: This snippet here is not meant to stay; I just put it in here to get CI to pass. Should change this upstream.

user27182: Yea we should change this upstream. It's related to https://github.com/pyvista/vtk-xref/issues/7. Your fix seems easier to implement. This has been an ongoing flaky issue for a while.

user27182:
Just tested this locally with PyCharm using IPython and also within the file editor, and it doesn't seem to work 😔:

image

I think PyCharm gets its type hints from the vtk `.pyi` files and not `__dir__`.

But I can confirm this does work with VSCode:

image

And so maybe we should add a warning that this does not work with some IDEs/editors, e.g. PyCharm

pyvista/core/_vtk_utilities.py
```diff
@@ -144,6 +145,39 @@ def __getattribute__(self, item):
...
+        VTK-inherited names are hidden by default so PyVista objects present
+        a curated public surface in data-science IDEs (Positron Variables
+        pane, VS Code Jupyter extension, ...) and in IPython / Jupyter
+        tab-completion. VTK methods remain fully callable; only their
+        enumeration is suppressed.
+
+        - CamelCase VTK attributes (``GetBounds``, ``DeepCopy``, ...) are
+          hidden unless :attr:`pyvista.global_config.show_vtk_api` is
+          ``True``.
+        - snake_case VTK aliases (``get_bounds``, ``deep_copy``, ...) are
```
user27182: VTK's snake case API removes the `Get` prefix, suggest using a different attr that PyVista doesn't override.
```suggestion
        - CamelCase VTK attributes (``GetNumberOfPoints``, ``DeepCopy``, ...) are
          hidden unless :attr:`pyvista.global_config.show_vtk_api` is
          ``True``.
        - snake_case VTK aliases (``number_of_points``, ``deep_copy``, ...) are
```

pyvista/core/config.py
```diff
@@ -243,5 +249,52 @@ def validate_on_wrap(self, value: bool) -> None:
...
+        Jupyter tab-completion. VTK methods remain fully callable regardless
+        of this setting.
+
+        When ``True``, the full VTK API is enumerated alongside the PyVista
+        API, which is useful for VTK developers who want to discover the raw
+        VTK method surface via introspection.
+
+        Notes
+        -----
+        The snake_case VTK aliases (``get_bounds``, ``deep_copy``, ...) are
```
user27182:
```suggestion
        The snake_case VTK aliases (``number_of_points``, ``deep_copy``, ...) are
```

pyvista/core/config.py
```diff
@@ -243,5 +249,52 @@ def validate_on_wrap(self, value: bool) -> None:
...
+        :class:`~pyvista.Actor`, :class:`~pyvista.Property`, etc.). This keeps
+        the public surface curated for data-science IDEs such as Positron's
+        Variables pane and VS Code's Jupyter extension, and for IPython /
+        Jupyter tab-completion. VTK methods remain fully callable regardless
+        of this setting.
+
+        When ``True``, the full VTK API is enumerated alongside the PyVista
+        API, which is useful for VTK developers who want to discover the raw
+        VTK method surface via introspection.
+
```
user27182, line 268:
```suggestion

        .. warning::

            This option requires runtime inspection and does not work with all developer
            tools, e.g. it has no effect when using PyCharm. This is because it relies on
            calling the object's ``__dir__`` method for generating auto-completion
            suggestions. Tools like PyCharm that only use static analysis for
            auto-completion are therefore unaffected.

```

tests/core/test_dataobject.py
```diff
@@ -553,3 +553,123 @@ def test_raise_error_when_writing_is_failed(tmp_path):
...
+        pv.StructuredGrid(),
+        pv.ImageData(),
+        pv.RectilinearGrid(),
+        pv.MultiBlock(),
+        pv.PartitionedDataSet(),
+        pv.Table(),
+        pv.PointSet(),
+    ],
+)
+def test_dir_hides_vtk_inherited_attributes(data_object):
```
user27182: Maybe add some plotting classes too to get a better sample of classes this should apply to? Alternatively, there are tests in `test_attributes.py` which auto-collect _all_ PyVista classes that inherit from VTK classes, such that we can test all classes to ensure this behavior is correct.

I.e. consider moving this test there.

user27182:
I also confirmed that VTK's snake case attributes are _hidden_ by default, e.g. VTK's API has an `inverse` method which was super confusing since PyVista uses `invert` instead:

image

user27182 (commented):
Looks great, I just have some minor suggestions

doc/source/conf.py
banesullivan: Staged https://github.com/pyvista/vtk-xref/pull/8

akaszynski (dismissed):

akaszynski (approved):

